Issue #1500, opened 2021-02-02
Epic: Consistent font/size, spacing scale, color, etc for components


jrowlingson:
### Description
Enable consistency throughout the component library of font/size, spacing scale, color, shadow, focus states, etc.

Now that @asangma 's tailwind configuration refactor has landed (https://github.com/Esri/calcite-components/pull/1415), we should be in a good spot to begin the larger component refactor to use said configuration. I'd suggest grabbing a range below (e.g. a-b) if you are interested in helping out.

### Acceptance Criteria 
- Consistency with UI Kit(s)

#### Which Component

- [ ] `calcite-accordion`, (#1803 )
- [ ] `calcite-accordion-item`, (#1803)
- [ ] `calcite-action` (https://github.com/Esri/calcite-components/pull/1501)
- [ ] `calcite-action-bar`(#1805)
- [ ] `calcite-action-group` (#1814)
- [ ] `calcite-action-pad`
- [ ] `calcite-alert` (#1558)
- [ ] `calcite-avatar` (https://github.com/Esri/calcite-components/pull/1559)
- [ ] `calcite-block` (#1561)
- [ ] `calcite-block-section` (#1561)
- [ ] `calcite-button`
- [ ] `calcite-card` (#1778)
- [ ] `calcite-checkbox` (#1780)
- [ ] `calcite-chip` (#1781)
- [ ] `calcite-color-picker` (#1782)
- [ ] `calcite-color-hex-input` (#1784)
- [ ] `calcite-color-swatch` (#1787)
- [ ] `calcite-combobox` (#1789)
- [ ] `calcite-combobox-item` (#1792)
- [ ] `calcite-date-picker` (#1793)
- [ ] `calcite-date-picker-day` (#1794)
- [ ] `calcite-date-picker-month` (#1795)
- [ ] `calcite-date-picker-month-header` (#1799)
- [ ] `calcite-dropdown` (#1800)
- [ ] `calcite-dropdown-group` (#1801)
- [ ] `calcite-dropdown-item` (#1802)
- [ ] `calcite-fab` (#1654)
- [ ] `calcite-filter` (#1658)
- [ ] `calcite-flow`(#1660)
- [ ] `calcite-graph` (#1662)
- [ ] `calcite-handle`(#1664)
- [ ] `calcite-icon`(#1666)
- [ ] `calcite-inline-editable` (#1668)
- [ ] `calcite-input`(#1668)
- [ ] `calcite-input-date-picker` (#1668)
[...]

PR #1738, opened 2021-03-11
refactor(tooltip): tailwindify

julio8a:
Suppports #1500

### DIFF
--- a/src/components/calcite-tooltip/calcite-tooltip.scss
+++ b/src/components/calcite-tooltip/calcite-tooltip.scss
@@ -2,16 +2,16 @@
 @include popperArrow();
 
 .container {
-  position: relative;
-  background: var(--calcite-ui-foreground-1);
+  @apply relative 
+    bg-foreground-1 
+    flex 
+    justify-start 
+    flex-col 
+    font-medium 
+    py-3 
+    px-4 
+    overflow-hidden 
+    text--3-wrap;
   max-width: 300px;
   max-height: 300px;
-  display: flex;
-  justify-content: flex-start;
-  flex-direction: column;
-  font-weight: 500;
-  color: var(--calcite-ui-text-1);
-  padding: 12px 16px;
-  overflow: hidden;
-  @include font-size(-3);
 }
--- a/src/components/calcite-tooltip/calcite-tooltip.scss
+++ b/src/components/calcite-tooltip/calcite-tooltip.scss
@@ -17,7 +17,7 @@
     border-solid
     border-color-3;
   @include font-size(-3);
-  max-width: 20rem
+  max-width: 20rem;
   max-height: 20rem;
   border-radius: var(--calcite-border-radius);
 }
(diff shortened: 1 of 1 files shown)


src/components/calcite-tooltip/calcite-tooltip.scss
```diff
@@ -2,16 +2,16 @@
...
+    flex 
+    justify-start 
+    flex-col 
+    font-medium 
+    py-3 
+    px-4 
+    overflow-hidden 
+    text--3-wrap;
   max-width: 300px;
   max-height: 300px;
```
jrowlingson, line 16: Could we use `rem` units here?

jrowlingson (approved):

src/components/calcite-tooltip/calcite-tooltip.scss
```diff
@@ -2,16 +2,16 @@
...
+    flex 
+    justify-start 
+    flex-col 
+    font-medium 
+    py-3 
+    px-4 
+    overflow-hidden 
+    text--3-wrap;
+  max-width: 18.75rem;
+  max-height: 18.75rem;
```
jrowlingson, line 16: bonus points if we can snap this to an even whole number e.g. 18 or 20

julio8a: What would I use? spacing only goes up to 16rem

jrowlingson: It looks like extrapolating would give us something like `max-width: 20rem`. This would be ~`w-72`. Not suggesting we modify the config here - just hard code it.

julio8a: I hard codded to 20rem, this is what you mean right?

jrowlingson (approved):

julio8a:
didn't have border-radius either, just added it. 

@jrowlingson, since we use this border-radius on a few components (alerts, cards, modal, etc) Should we add this as a tailwind value?

bstifle:
yes should totally be a var for border radius!

julio8a:
I added a list item at the bottom of the list:
https://github.com/Esri/calcite-components/issues/1500

src/components/calcite-tooltip/calcite-tooltip.scss
```diff
@@ -2,16 +2,22 @@
...
+    font-medium
+    text-color-1 
+    py-3 
+    px-4 
+    overflow-hidden 
+    text--1-wrap
+    border
+    border-solid
+    border-color-3;
   @include font-size(-3);
```
jrowlingson, line 19: Intended? this is overriding text--1-wrap

src/components/calcite-tooltip/calcite-tooltip.scss
```diff
@@ -2,16 +2,22 @@
...
+  max-width: 20rem;
+  max-height: 20rem;
+  border-radius: var(--calcite-border-radius);
```
jrowlingson, line 22: I thought we decided we would use the [defaults](https://v1.tailwindcss.com/docs/border-radius) for `border-radius`? 

We could set vars for those values if we must.. but either way, we should be using a util here.

src/components/calcite-tooltip/calcite-tooltip.scss
julio8a: nope!

PR #1743, opened 2021-03-11
fix(input, label): Removing disabled prop should also updated slotted elements.

driskull:
**Related Issue:** NA

## Summary

fix(input, label): Removing disabled prop should also updated slotted elements.

- If you add a disabled prop to an input or label, it adds disabled to the slotted input.
  - However, if you then remove that disabled prop, it does nothing.
  - I think it should update it.

### DIFF
--- a/src/components/calcite-input/calcite-input.tsx
+++ b/src/components/calcite-input/calcite-input.tsx
@@ -450,10 +450,11 @@ export class CalciteInput {
     if (!this.slottedActionEl) {
       return;
     }
+    const slottedActionEl = this.slottedActionEl as HTMLElement;
 
     this.disabled
-      ? ((this.slottedActionEl as HTMLElement) as HTMLElement).setAttribute("disabled", "")
-      : ((this.slottedActionEl as HTMLElement) as HTMLElement).removeAttribute("disabled");
+      ? slottedActionEl.setAttribute("disabled", "")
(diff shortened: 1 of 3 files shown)


src/components/calcite-input/calcite-input.tsx
```diff
@@ -447,7 +447,13 @@ export class CalciteInput {
...
 
   private setDisabledAction(): void {
-    if (this.slottedActionEl) (this.slottedActionEl as HTMLElement).setAttribute("disabled", "");
+    if (!this.slottedActionEl) {
+      return;
+    }
+
+    this.disabled
+      ? ((this.slottedActionEl as HTMLElement) as HTMLElement).setAttribute("disabled", "")
+      : ((this.slottedActionEl as HTMLElement) as HTMLElement).removeAttribute("disabled");
```
eriklharper: Doesn't Stencil take care of adding/removing the `disabled` attribute for us if this is reflected?

eriklharper (approved):

src/components/calcite-input/calcite-input.tsx
driskull: This isn't on the host element.

eriklharper: Woops!  Totally get it now :)

macandcheese (approved):

PR #1745, opened 2021-03-12
refactor(tab-title): use tailwind

caripizza:
**Related Issue:** #1500

## Summary
Refactors tab-title styles to use the tailwind config. Changes 3px border widths to 2px (match calcite-notice).

### DIFF
--- a/src/components/calcite-tab-title/calcite-tab-title.scss
+++ b/src/components/calcite-tab-title/calcite-tab-title.scss
@@ -12,8 +12,8 @@
 :host([position="below"]) a {
   @apply border-b-0
     border-t-2
-    border-t-solid
     border-t-color-transparent;
+  border-top-style: solid;
 }
 
 :host a {
@@ -50,19 +50,18 @@ span {
     appearance-none
     cursor-pointer
     truncate
-    border-b-3
+    border-b-2
     border-b-color-transparent
-    border-b-solid
     w-full
     h-full
     flex
     justify-center
     text-color-3
-    text--1
-    transition-default
     text--1h
+    transition-default
     px-0
-    py-3;
+    py-2;
+  border-bottom-style: solid;
 }
 
 span {
--- a/src/components/calcite-tab-title/calcite-tab-title.scss
+++ b/src/components/calcite-tab-title/calcite-tab-title.scss
@@ -1,12 +1,12 @@
 :host {
-  @apply flex-initial outline-none mx-5;
+  @apply flex-initial outline-none mr-5;
   margin-inline-start: theme("margin.0");
   margin-inline-end: theme("margin.5");
 }
 
 :host([layout="center"]) {
   @apply text-center my-0 mx-5;
-  flex-basis: 200px;
+  flex-basis: theme("spacing.48");
 }
 
 :host([position="below"]) a {
--- a/src/components/calcite-tab-title/calcite-tab-title.scss
+++ b/src/components/calcite-tab-title/calcite-tab-title.scss
@@ -1,124 +1,114 @@
-$tab-margin: 1.25rem;
-
 :host {
-  flex: 0 1 auto;
-  outline: none;
-  margin-right: $tab-margin;
-  margin-inline-start: 0;
-  margin-inline-end: $tab-margin;
+  @apply flex-initial outline-none mx-5;
+  margin-inline-start: theme("margin.0");
+  margin-inline-end: theme("margin.5");
 }
 
 :host([layout="center"]) {
+  @apply text-center my-0 mx-5;
   flex-basis: 200px;
-  text-align: center;
-  margin: 0 $tab-margin;
 }
 
 :host([position="below"]) a {
-  border-bottom: 0;
-  border-top: 3px solid transparent;
+  @apply border-b-0
+    border-t-2
+    border-t-solid
+    border-t-color-transparent;
 }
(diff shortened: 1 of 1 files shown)


src/components/calcite-tab-title/calcite-tab-title.scss
```diff
@@ -1,124 +1,114 @@
...
+    truncate
+    border-b-3
+    border-b-color-transparent
+    border-b-solid
+    w-full
+    h-full
+    flex
+    justify-center
+    text-color-3
+    text--1
```
caripizza, line 61: todo: `text--1`, or `text--1h`?

jrowlingson: text--1h

src/components/calcite-tab-title/calcite-tab-title.scss
```diff
@@ -1,124 +1,113 @@
-$tab-margin: 1.25rem;
-
 :host {
-  flex: 0 1 auto;
-  outline: none;
-  margin-right: $tab-margin;
-  margin-inline-start: 0;
-  margin-inline-end: $tab-margin;
+  @apply flex-initial outline-none mx-5;
```
jrowlingson, line 2: mx-5 intended or should this be mr-5?

src/components/calcite-tab-title/calcite-tab-title.scss
```diff
@@ -1,124 +1,113 @@
...
+  margin-inline-start: theme("margin.0");
+  margin-inline-end: theme("margin.5");
 }
 
 :host([layout="center"]) {
+  @apply text-center my-0 mx-5;
   flex-basis: 200px;
```
jrowlingson, line 9: Perhaps a design follow up - but it would be cool if we could update this to an even int rem.

src/components/calcite-tab-title/calcite-tab-title.scss
```diff
@@ -1,124 +1,113 @@
...
-  margin: 0 $tab-margin;
 }
 
 :host([position="below"]) a {
-  border-bottom: 0;
-  border-top: 3px solid transparent;
+  @apply border-b-0
+    border-t-2
+    border-t-color-transparent;
+  border-top-style: solid;
```
jrowlingson, line 16: would `border-solid` work here?

src/components/calcite-tab-title/calcite-tab-title.scss
```diff
@@ -1,124 +1,113 @@
...
+    border-b-color-transparent
+    w-full
+    h-full
+    flex
+    justify-center
+    text-color-3
+    text--1h
+    transition-default
+    px-0
+    py-2;
```
jrowlingson, line 63: p-0

src/components/calcite-tab-title/calcite-tab-title.scss
```diff
@@ -1,124 +1,113 @@
...
+  border-bottom-style: solid;
```
jrowlingson, line 64: would `border-solid` work here?

src/components/calcite-tab-title/calcite-tab-title.scss
caripizza: nope good catch - I'll use `mr-5` here

src/components/calcite-tab-title/calcite-tab-title.scss
caripizza: how about 12rem? `flex-basis: theme("spacing.48");`

src/components/calcite-tab-title/calcite-tab-title.scss
caripizza: Nope, if I remove `border-top-style: solid` here, and instead use `border-solid` in this apply directive, then when the calcite-tab-title's `position="below"`, the gray hover effect appears on all sides except for bottom

src/components/calcite-tab-title/calcite-tab-title.scss
caripizza: But padding wasn't applied to all sides of the link tag previously, and wasn't 0:
```scss
  padding: $baseline/2 0; // ie., `0.75rem 0;`
```

Initially, I used
```scss
 px-0
 py-3 // 0.75rem
```
But @bstifle told me to use spacing 2 from the theme (0.5rem), thus I went with `py-2` (0.5rem).
I can change this back to `py-3` though - if we prefer more spacing around the tab title.

src/components/calcite-tab-title/calcite-tab-title.scss
caripizza: Here, `border-solid` adds the gray border effect around all sides of the hovered tab-title link; I would need to override the unwanted sides' borders with:

```
    border-b-2
    border-b-color-transparent
    border-r-0
    border-l-0
    border-t-0
    border-solid
```

src/components/calcite-tab-title/calcite-tab-title.scss
jrowlingson: Oh, fail on my part :facepalm:  i read that as px-0 py-0

jrowlingson (approved):

caripizza:
I accepted all 8 Screener diffs, since they were tied to the new 2px padding height for tab-title.

PR #1754, opened 2021-03-12
fix(shell): simplifies content render removes center-row conditional

asangma:
Also adds async test demo

**Related Issue:** #

## Summary
* Simpler render
* removes conditional render of center-row slot

cc @kevindoshier

### DIFF
--- a/src/components/calcite-shell/calcite-shell.tsx
+++ b/src/components/calcite-shell/calcite-shell.tsx
@@ -53,37 +53,27 @@ export class CalciteShell {
     return hasHeader ? <slot name={SLOTS.header} /> : null;
   }
 
-  renderContent(): VNode {
-    return !!this.contentBehind ? this.renderContentBehind() : this.renderContentInline();
-  }
+  renderContent(): VNode[] {
+    const content = !!this.contentBehind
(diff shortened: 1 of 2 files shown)


src/components/calcite-shell/calcite-shell.tsx
```diff
@@ -53,37 +53,27 @@ export class CalciteShell {
...
-  }
-
-  renderCenterRow(): VNode {
-    const hasCenterRow = !!getSlotted(this.el, SLOTS.centerRow);
-
-    return hasCenterRow ? <slot name={SLOTS.centerRow} /> : null;
+  renderContent(): VNode[] {
+    const content = !!this.contentBehind
+      ? [
+          <div
```
driskull, line 59: Do we not need the key on here anymore?

driskull (approved):
👍

src/components/calcite-shell/calcite-shell.tsx
asangma: It doesn't seem like it was making a difference.

asangma:
Thanks @driskull.

Issue #1749, opened 2021-03-12
Bug: calcite-panel - disabled:true makes panel scroll to top


AdelheidF:
### Summary
We want to disable the panel if the user opens a popover overlapping it. 
If you disable the panel while you're scrolled to the bottom it will automatically scroll up. So the user has to scroll back down after closing the popover.

### Reproduction Steps

1.

AdelheidF:
I noticed a second thing.

On the left is my panel without the popover. On the right is the panel in disabled state (I'm hiding the popover to show the issue better). Notice where the footer buttons are.

<img src=https://user-images.githubusercontent.com/44481292/111010740-37b08300-834c-11eb-8ff1-c52fba9309f7.png width=210/><img src=https://user-images.githubusercontent.com/44481292/111010786-63336d80-834c-11eb-8c4f-031a91d1e34b.png width=200/>

driskull:
Installed.

AdelheidF:
looks good with next.123

PR #1757, opened 2021-03-15
fix(filter): Place scrim over filter when loading or disabled.

driskull:
**Related Issue:** #1749

## Summary

- fix(filter): Place scrim over filter when loading or disabled.
- refactor(block, panel, pick-list, value-list): Refactor scrim usage in components.

Places scrim above components and doesn't wrap any content.

We don't want to wrap any content and set overflow:hidden because that will kill the scroll bar any components use and users will lose their place in the scroll position.

### DIFF
--- a/src/components/calcite-filter/calcite-filter.scss
+++ b/src/components/calcite-filter/calcite-filter.scss
@@ -4,20 +4,20 @@
 }
 
 label {
-  @apply flex 
-    items-center 
-    overflow-hidden 
+  @apply flex
+    items-center
+    overflow-hidden
     relative w-full
     mx-1
     my-0;
 }
 
 input[type="text"] {
-  @apply bg-transparent 
+  @apply bg-transparent
     border-none
-    font-inherit 
-    text--2h 
-    mb-1 
+    font-inherit
+    text--2h
+    mb-1
     w-full
     py-1
     pr-1
@@ -29,9 +29,9 @@ input[type="text"] {
 }
 
 .search-icon {
-  @apply flex 
-    left-0 
-    absolute 
+  @apply flex
+    left-0
+    absolute
     text-color-2;
   transition: left $transition, right $transition, opacity $transition;
 }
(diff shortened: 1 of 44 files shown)


src/components/calcite-filter/calcite-filter.scss
```diff
@@ -62,12 +62,23 @@ input[type="text"]:focus {
...
+    border-0
     cursor-pointer;
   &:hover,
   &:focus {
     @apply text-color-1;
   }
 }
+
+calcite-scrim {
+  margin: -0.5rem;
```
driskull, line 76: @julio8a is there a better way to set a negative margin of `p-2`?

jcfranco (approved):
👍

src/components/calcite-filter/calcite-filter.scss
asangma: Instead of the negative margin, which might be a bit brittle, we could try top, left, right, bottom positionings of zero.

driskull: Can't, the padding is on the host.

driskull:

@julio8a can you review the css?

src/components/calcite-scrim/calcite-scrim.scss
```diff
@@ -1,5 +1,15 @@
...
+  absolute
+  z-50
+  flex-col
+  items-stretch
+  w-full
+  h-full
+  top-0
+  left-0
+  bottom-0
+  right-0;
```
julio8a, line 12: Instead of top, left, bottom, right, use `inset-0`.

julio8a (approved):

driskull:
@jcfranco this good to merge?

src/components/calcite-flow/calcite-flow.stories.ts
```diff
@@ -66,6 +66,10 @@ const createFlowItemAttributes: (group: string) => Attributes = (group) => {
     {
       name: "intl-close",
       value: text("intlClose", TEXT.close, group)
+    },
+    {
+      name: "style",
```
jcfranco, line 71: Is this intentional? If it's needed for the story to render properly, let's move it to the story creation logic.

Applies to other component stories.

jcfranco (approved):
👍

PR #1762, opened 2021-03-16
feat(filter): clear text on Escape

jcfranco:
**Related Issue:** N/A

## Summary

Allows the filter text term to be cleared by pressing the Escape key.


driskull (approved):
Yus. escape it all

PR #1763, opened 2021-03-16
chore: fine-tune codeowners

jcfranco:
**Related Issue:** N/A

## Summary

Updates codeowners file to use reviewer group as catch-all.

### Questions

Do we want to also set up granular ownership? Something like this would work:

```
# component-specific owners
calcite-block/                    @jcfranco
calcite-block-section/            @jcfranco
calcite-color-picker/             @jcfranco
calcite-color-picker-hex-input/   @jcfranco
calcite-color-picker-swatch/      @jcfranco
calcite-icon/                     @jcfranco
calcite-option/                   @jcfranco
calcite-option-group/             @jcfranco
calcite-pick-list/                @jcfranco
calcite-select/                   @jcfranco
calcite-value-list/               @jcfranco
```

Anything not specified above would be owned by the reviewers group (see [codeowners syntax](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax)).

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,8 +1,2 @@
 # These owners will be the default owners for everything in the repo.
-*         @esri/calcite-components
-
-# docs:
-/docs/**  @esri/all-esri-members
-
-# Template and assets
-*.md      @esri/all-esri-members
+*     @esri/calcite-components-reviewers


driskull:
> Do we want to also set up granular ownership? Something like this would work:

Maybe that might be a little too fine grained? I'm ok with it though.

CODEOWNERS
```diff
@@ -1,8 +1,2 @@
 # These owners will be the default owners for everything in the repo.
-*         @esri/calcite-components
-
-# docs:
-/docs/**  @esri/all-esri-members
-
-# Template and assets
-*.md      @esri/all-esri-members
+*     @esri/calcite-components-reviewers
```
driskull, line 2: @julio8a @jcfranco  can we rename this group to `calcite-codeowners`?

driskull (approved):
This is a good start to slim down notification 👍

paulcpederson:
> Do we want to also set up granular ownership?

I think this would be great in the future. This is at least an improvement, though.

CODEOWNERS
jcfranco: I'm fine either way. Something that applies to reviews and code ownership works for me.

jcfranco:
Merging to slim down notifications. Will hold off on granular ownership for now.

@julio8a If you rename the reviewers group, can you update this file as well? 🙇‍♂️

julio8a:
okay, this looks good to me, but it's good to know this needs to change too :D

PR #1765, opened 2021-03-16
julio8a/refactor-tooltip

julio8a:
Supports #1500

- uses rounded-sm for border radius. 2px (not 3px) @bstifle is this okay? 3px is not in the tailwind config
- Fixed the pointer position

cc'ing @bstifle @driskull

### DIFF
--- a/src/assets/styles/_popper.scss
+++ b/src/assets/styles/_popper.scss
@@ -111,7 +111,7 @@ $pointer_offset: -$pointer_size/2;
     position: absolute;
     width: $pointer_size;
     height: $pointer_size;
-    z-index: -1;
+    z-index: 1;
   }
 
   .arrow::before {
--- a/src/components/calcite-tooltip/calcite-tooltip.scss
+++ b/src/components/calcite-tooltip/calcite-tooltip.scss
@@ -16,7 +16,7 @@
     border
     border-solid
     border-color-3
-    rounded-sm;
+    rounded;
   max-width: 20rem;
   max-height: 20rem;
 }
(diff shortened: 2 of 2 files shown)


src/assets/styles/_popper.scss
```diff
@@ -111,7 +111,7 @@ $pointer_offset: -$pointer_size/2;
     position: absolute;
     width: $pointer_size;
     height: $pointer_size;
-    z-index: -1;
+    z-index: 1;
```
driskull, line 114: shouldn't the tail be under?

julio8a: It was, but the border went over the pointer and looks disconnected. 
image

julio8a: The tooltip didn't have a border before, which is why it wasn't noticeable, but from the design, it does need a border.

bstifle:
> Supports #1500
> 
> * uses rounded-sm for border radius. 2px (not 3px) @bstifle is this okay? 3px is not in the tailwind config
> * Fixed the pointer position
> 
> cc'ing @bstifle @driskull

global border radius should be 4px on anything using a border radius

julio8a:
Great! there is 4px, just pushed an update!

julio8a:
@bstifle, can you review screener changes

bstifle:
looks good. just curious, what font size are you using? @julio8a

julio8a:
Whoops, looks like I had --1h when I should have had --2h (what's in Figma), I pushed up an update.

driskull (approved):
👍

Issue #1622, opened 2021-02-25
Enhancement: Add a method to set focus on the calcite-filter for a pick list with filter enabled


ethanbdev:
### Description
It would be nice to have a method on the pick list component to set focus on the filter when enabled. Right now the setFocus focuses on the pick list itself. This way we can programmatically take the user directly to the search box to immediately start typing. 
### Acceptance Criteria 
A method callable on the pick list to focus the underlying calcite-filter instead of the pick list items.
#### Which Component
calcite-pick-list / calcite-filter
#### Example Use Case
We use a pick-list as a field select in a popover. Feedback we have received is to immediately be able to start typing in the filter on the popup opening, saving the user a click.
>I expected that I could push the drop down > opens the pop over > I can type and it would search. It seems like I need to explicitly click in the search and then type.

jcfranco:
@ethanbdev Could you help verify this one? 🙇🏼‍♂️

ethanbdev:
Looks good to me!
https://codepen.io/eborgen/pen/VwPaMVN

My only comment would be to document that "filter" is an option for the `focusId` argument.

jcfranco:
> My only comment would be to document that "filter" is an option for the focusId argument.

Noted. 

--- 

Closing as verified, thanks!

PR #1766, opened 2021-03-16
feat(pick-list, value-list): Add option on setFocus method to focus the filter component. #1622

driskull:
**Related Issue:** #1622

## Summary

feat(pick-list, value-list): Add option on setFocus method to focus the filter component. #1622


jcfranco (approved):
Awesome! 🌮 🎉  

Before merging, could you move the tests into `shared-list-tests.ts` and consolidate them?

PR #1773, opened 2021-03-17
refactor(tip-group): tailwindify

julio8a:
Supports #1500


src/components/calcite-tip-group/calcite-tip-group.scss
```diff
@@ -1,10 +1,9 @@
 :host {
   @extend %component-host;
-  @extend %component-spacing;
   --calcite-tip-max-width: 540px;
```
jrowlingson, line 3: Is the tip group ever used outside of the tip manager? Could we update this to become an even int rem?

src/components/calcite-tip-group/calcite-tip-group.scss
```diff
@@ -1,10 +1,8 @@
...
   @extend %component-host;
-  @extend %component-spacing;
-  --calcite-tip-max-width: 540px;
 }
 
 ::slotted(calcite-tip) {
-  border: none;
-  max-width: var(--calcite-tip-max-width);
+  @apply border-none;
+  max-width: 33.75rem;
```
jrowlingson, line 7: Change to 34? or 32?

jrowlingson (approved):

PR #1774, opened 2021-03-17
refactor(tip-manager): tailwindify

julio8a:
Supports #1500

### DIFF
--- a/src/components/calcite-tip-manager/calcite-tip-manager.scss
+++ b/src/components/calcite-tip-manager/calcite-tip-manager.scss
@@ -46,7 +46,6 @@
 
 ::slotted(calcite-tip) {
   @apply border-none;
-  border: none;
   max-width: var(--calcite-tip-max-width);
 }
 
--- a/src/components/calcite-tip-manager/calcite-tip-manager.scss
+++ b/src/components/calcite-tip-manager/calcite-tip-manager.scss
@@ -1,50 +1,51 @@
 :host {
   @extend %component-host;
-  @extend %component-spacing;
-  display: block;
+  @apply block;
   --calcite-tip-manager-height: 18vh;
   --calcite-tip-max-width: 540px;
 }
 
 :host([closed]) {
-  display: none;
+  @apply hidden;
 }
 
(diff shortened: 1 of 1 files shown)


julio8a:
Not sure why this is showing tooltip, which didn't change.

src/components/calcite-tip-manager/calcite-tip-manager.scss
```diff
@@ -1,50 +1,51 @@
...
-  @include focus-style-base();
   &:focus {
-    @include focus-style-outset();
+    @apply focus-outset;
   }
 }
 
 ::slotted(calcite-tip) {
+  @apply border-none;
   border: none;
```
eriklharper, line 49: ☝️ remove `border: none`

eriklharper (changes requested):

src/components/calcite-tip-manager/calcite-tip-manager.scss
julio8a: Good catch!

eriklharper (approved):

src/components/calcite-tip-manager/calcite-tip-manager.scss
```diff
@@ -59,35 +59,31 @@
...
-  align-items: center;
-  justify-content: center;
-  padding: var(--calcite-spacing-quarter) 0;
+  @apply flex items-center justify-center py-1 px-0;
 }
 
 .page-position {
-  @include font-size(-4);
-  margin: 0 var(--calcite-spacing-half);
+  @apply text--2h my-0 mx-2;
```
paulcpederson, line 66: I think you replaced `-4` with `-2` here, just confirming that is intentional

paulcpederson (approved):

src/components/calcite-tip-manager/calcite-tip-manager.scss
julio8a: Yes, this is what is on the Figma design. Thanks for checking. Good eyes Paul, good eyes. You still got it.

PR #1777, opened 2021-03-17
feat(combobox): add constant prop to combobox item

paulcpederson:
## Summary

This allows devs to keep certain combobox items visible even when filtering. The use case here is some sort of "new" option:

Screen Shot 2021-03-17 at 11 52 00 AM

However, this could also be used to put together an async combobox UI. Imagine you want to have a selection UI for members, but you have to hit a paginated API. You could use the filter event from the box, hit the API, and render the results. If you use this prop on all of them, you'll be able to show them even though text is entered. Basically it allows devs to take care of list filtering themselves if they want to as an opt-in.

I haven't added a storybook entry for this. The combobox storybook needs a rewrite to work more like accordions where there are multiple pages of knobs that are grouped for combobox, combobox-item.

### DIFF
--- a/src/components/calcite-combobox-item/calcite-combobox-item.tsx
+++ b/src/components/calcite-combobox-item/calcite-combobox-item.tsx
@@ -58,6 +58,9 @@ export class CalciteComboboxItem {
   /** A unique value used to identify this item - similar to the value attribute on an <input>. */
   @Prop({ reflect: true }) value!: string;
 
+  /** Don't filter this item based on the search text */
+  @Prop({ reflect: true }) alwaysVisible: boolean;
+
   // --------------------------------------------------------------------------
   //
   //  Private Properties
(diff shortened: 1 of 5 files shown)


src/components/calcite-combobox-item/calcite-combobox-item.tsx
```diff
@@ -58,6 +58,9 @@ export class CalciteComboboxItem {
   /** A unique value used to identify this item - similar to the value attribute on an <input>. */
   @Prop({ reflect: true }) value!: string;
 
+  /** Don't filter this item based on the search text */
+  @Prop({ reflect: true }) alwaysVisible: boolean;
```
driskull, line 62: nitpick: maybe we can think of a better name? How about `constant`, `keep`, `hold`?

driskull (approved):
👍

src/components/calcite-combobox-item/calcite-combobox-item.tsx
paulcpederson: `constant` is pretty good

PR #1780, opened 2021-03-17
refactor(checkbox): tailwindify

julio8a:
Supports #1500

### DIFF
--- a/src/components/calcite-checkbox/calcite-checkbox.scss
+++ b/src/components/calcite-checkbox/calcite-checkbox.scss
@@ -17,21 +17,22 @@
     select-none;
   -webkit-tap-highlight-color: transparent;
   .check-svg {
+    @apply overflow-hidden
+      inline-block
+      bg-foreground-1
+      pointer-events-none
+      box-border
+      transition-all
+      duration-150
+      ease-in-out;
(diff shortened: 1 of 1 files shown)


src/components/calcite-checkbox/calcite-checkbox.scss
```diff
@@ -58,9 +58,7 @@
   }
```
jrowlingson, line 58: Possibly a question for @asangma - it looks like this mixin is only used once, should we kill? Create a util for this?

jrowlingson:
The checkbox size expressed in terms of `fontSize` seems a little weird.. there is also two different patterns in here for setting properties based on the `scale` attribute. I have'nt played around with this but wonder if we could revisit later on.. perhaps remove the reliance on grid?

eriklharper:
> perhaps remove the reliance on grid?

Since we don't have to worry about IE anymore, I think we can still take advantage of css grid here, but we could definitely refactor how the size of the checkbox is expressed.  These two issues are mutually exclusive I'd say.

julio8a:
Okay, can I merge this and focus those mentioned updates later?

jrowlingson (approved):

Issue #1750, opened 2021-03-12
Bug:  calcite-shell-panel - header slot messes up fab button


AdelheidF:
### Summary

Might have started with beta.51.

The "header" in the calcite-shell-panel that the layerSwitcher is in, it pushes everything down.
<img src=https://user-images.githubusercontent.com/44481292/111005284-607e4b80-833f-11eb-977a-6f9c8c572b05.png width=250/>

### Reproduction Steps

1.

asangma:
@AdelheidF A simple fix I found was to set `overflow:hidden` on your component, e.g. on `arcgis-filter`.

I can add a style for a slotted `calcite-panel` or `calcite-flow`, but I would be reluctant to do a wildcard `*` selector.
I could maybe do a `[class^="arcgis-"]` to target our micro apps...?

cc @driskull @jcfranco @paulcpederson

driskull:
> I could maybe do a [class^="arcgis-"] to target our micro apps...?

Probably not a good idea.

AdelheidF:
So, we should add `overflow:hidden` to all panel-components that show a fab button and could be used with a layer switcher?

asangma:
I think apply `overflow:hidden` to any of the `arcgis-[your component name]` elements.
This would apply to Popup config, etc. as well. (Can't seem to get Dhrumil's handle to appear here).

Charts would need to do something similar, but there seems to be other things happening with that mico-app.

asangma:
I'll keep trying other things in the meantime.

asangma:
Update: hold off on adding the overflow:hidden. I'll try to find another way that doesn't require that. If I can't soon, you'll have to add it.

asangma:
@AdelheidF Installed.

You can ignore the overflow suggestion earlier.
If you see a lack of a scrollbar where there should be one, try setting `height:100%` on things.

If a Panel or Flow is placed directly into the default slot, it should work as-is.
For anything else, you'll probably need to put `height: 100%` on it. 
This includes a Panel inside something else, e.g. `arcgis-[some thing]` > `calcite-panel`.

AdelheidF:
looks good with next.123

PR #1791, opened 2021-03-18
fix(shell-panel): adds divs to provide accurate height calculations for slotted content

asangma:
**Related Issue:** # (#1750)

## Summary
This PR lets a component in ShellPanel's default slot more accurately get the height of its container.

The issues from #1750 appear to be happening because a component in ShellPanel's default slot will read the height of the entire `.content` div. So if there's a header, the calculated value on `height: 100%` is no longer any good.

Note that unless a Flow or Panel is placed directly into the default slot, e.g. anything else like a div or an outside component, it will still need to set `height: 100%` on itself and probably any Panel inside it.

cc @AdelheidF @kevindoshier

### DIFF
--- a/src/components/calcite-shell-panel/calcite-shell-panel.scss
+++ b/src/components/calcite-shell-panel/calcite-shell-panel.scss
@@ -29,6 +29,22 @@
   transition: max-height $transition, max-width $transition;
 }
 
+.content__header,
+.content__body {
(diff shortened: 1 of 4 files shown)


driskull (approved):
👍

asangma:
Thanks @driskull!

PR #1797, opened 2021-03-18
refactor: FocusId types were clashing

driskull:
refactor: FocusId types were clashing

### DIFF
--- a/src/components/calcite-popover/resources.ts
+++ b/src/components/calcite-popover/resources.ts
@@ -13,3 +13,5 @@ export const TEXT = {
 export const POPOVER_REFERENCE = "data-calcite-popover-reference";
 export const ARIA_CONTROLS = "aria-controls";
 export const ARIA_EXPANDED = "aria-expanded";
+
+export type PopoverFocusId = "close-button";
--- a/src/components/calcite-pick-list/calcite-pick-list.tsx
+++ b/src/components/calcite-pick-list/calcite-pick-list.tsx
@@ -12,7 +12,7 @@ import {
 } from "@stencil/core";
 import { ICON_TYPES } from "./resources";
 import {
-  FocusId,
+  ListFocusId,
   calciteListItemChangeHandler,
   calciteListItemValueChangeHandler,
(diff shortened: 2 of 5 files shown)


src/components/calcite-popover/resources.ts
```diff
@@ -13,3 +13,5 @@ export const TEXT = {
 export const POPOVER_REFERENCE = "data-calcite-popover-reference";
 export const ARIA_CONTROLS = "aria-controls";
 export const ARIA_EXPANDED = "aria-expanded";
+
+export type PopoverFocusId = "close-button";
```
jcfranco, line 17: Can we drop the component from all focus ID type names? It's already implied by the module.

jcfranco (approved):
Just had a comment about simplifying the type names. Other than that, LGTM!

src/components/calcite-popover/resources.ts
driskull: @jcfranco it was like that but the typings were renaming one of them to Focus1.

jcfranco: > the typings were renaming one of them to FocusId

I didn't follow that. Can you elaborate?

driskull: Yeah, when I have two exported FocusId interfaces, it does this:

```

import { FocusId } from "./components/calcite-pick-list/shared-list-logic";
--
31 | import { ICON_TYPES } from "./components/calcite-pick-list/resources";
32 | import { FocusId as FocusId1 } from "./components/calcite-popover/calcite-popover";


```

https://unpkg.com/browse/@esri/calcite-components@1.0.0-next.122/dist/types/components.d.ts

Then it causes issues with using the Popover cus the typings are wrong.

jcfranco: Updating based on our side conversation, this seems like a Stencil compiler issue. Looks like named imports do not get assigned properly for all cases. Renaming the imports is a good workaround for now. Once fixed, we can revisit simplifying the type name.

Issue #1752, opened 2021-03-12
Bug: calcite-panel - issue with scrollbars


AdelheidF:
### Summary

from Dhrumil:

I think this was after the latest calcite update.
I see a few panels where the scroll ability is gone, and in all of them I use the calcite-panel directly without calcite-flow.
Places where calcite-panel is wrapped with calcite-flow, scrolling is not an issue.

Is wrapping calcite-panel within calcite-flow a new requirement? If not, can this be fixed for 9.1?

beta.51:
<img src=https://user-images.githubusercontent.com/44481292/111007644-30857700-8344-11eb-907b-dc88ea66751e.png width=300/>

production (beta.49 or 48?):
<img src=https://user-images.githubusercontent.com/44481292/111007673-45620a80-8344-11eb-80a1-2a3c447551c0.png width=300/>

### Reproduction Steps

1.

AdelheidF:
SUPER HIGH PRIORITY for 9.1!

AdelheidF:
Looks good with next.123. I see scrollbars again.

<img src=https://user-images.githubusercontent.com/44481292/111810869-b600ae00-8893-11eb-9564-34e4ada60173.png width=200/>

Issue #1671, opened 2021-03-04
Question: Do we need the image node in Popover?


asangma:
### Question
Propose removing the image slot from Popover.

Right now we have an [image node](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-popover/calcite-popover.tsx#L346-L352) in Popover.

The image slot seems like an unnecessary and unused feature.
Supporting it also causes unnecessary nesting of nodes when there's just content.

### Helpful Details 
Divs with single divs inside them have a tendency to cause unforeseen UI bugs.

julio8a:
This sounds good to me

asangma:
Thanks @julio8a!

asangma:
@julio8a Mind if I assign this one to you?

PR #1798, opened 2021-03-18
fix(popover, panel)!: image slot removed and sizing styles updated

asangma:
**Related Issue:** #1752

## Summary
Includes some clean-up on Popover as discussed previously with @julio8a .

Also updates demo app for better demo apping.

### DIFF
--- a/src/components/calcite-popover/calcite-popover.scss
+++ b/src/components/calcite-popover/calcite-popover.scss
@@ -1,5 +1,3 @@
-$image_max_height: 200px;
-
 @include popperHost();
 @include popperArrow();
 
@@ -11,23 +9,28 @@ $image_max_height: 200px;
   pointer-events: initial;
 }
 
+.calcite-popper-anim {
+  @apply h-full;
+}
+
 .container {
   @apply bg-foreground-1 
     relative 
     flex 
     overflow-hidden 
     flex-no-wrap 
-    flex-col 
+    flex-row 
+    h-full
     text-color-1;
 }
 
 .content {
-  @apply flex 
-    flex-row 
-    justify-between 
-    justify-start;
-  line-height: theme("spacing.6");
+  @apply flex
+  flex-col
+  flex-no-wrap
+  h-full;
 }
+
 // focus styles
 .close-button {
   @apply focus-base;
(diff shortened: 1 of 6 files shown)


src/components/calcite-popover/calcite-popover.tsx
```diff
@@ -343,14 +343,6 @@ export class CalcitePopover {
   //
   // --------------------------------------------------------------------------
 
-  renderImage(): VNode {
```
driskull, line 346: how come this was removed? This would be a breaking change.

src/components/calcite-popover/calcite-popover.scss
```diff
@@ -60,20 +63,6 @@ $image_max_height: 200px;
   }
 }
 
-.image-container {
```
driskull, line 63: Why was the image slot removed?

src/components/calcite-popover/calcite-popover.tsx
driskull: The doc for the slot would have to be removed too.

```
/**
 * @slot image - A slot for adding an image. The image will appear above the other slot content.
 */
```

The commit message would need to change to specify the breaking change.

src/demos/shell/block-configurations.html
```diff
@@ -45,7 +45,7 @@
         const view = new MapView({
           container: "viewDiv",
           map: webmap,
-          padding: { left: 49, right: 385 }
+          // padding: { left: 49, right: 385 }
```
driskull, line 48: we can remove this padding.

driskull (commented):
Just need clarification on the breaking change.

src/components/calcite-popover/calcite-popover.tsx
asangma: I can update the doc and the PR to reflect the breaking change.
This image slot was added when the component was created and was never used.
Checked with @julio8a on this and got the go-ahead to remove it.

src/demos/shell/block-configurations.html
asangma: copy that.

driskull (approved):
👍

bstifle:
@asangma was never used, period? i thought one of the main points of making the popover component was for images

driskull:
Guess it wasn't :)

I never saw any use cases for showing an image in a popover.

asangma:
Forgot to mention the related issue #1671

PR #1804, opened 2021-03-19
refactor(accordion-item): tailwindify

julio8a:
Supports #1500

### DIFF
--- a/src/components/calcite-accordion-item/calcite-accordion-item.scss
+++ b/src/components/calcite-accordion-item/calcite-accordion-item.scss
@@ -1,26 +1,3 @@
-// host-content() is being used to support IE11 styling in this component.
-// scale variants for child
-:host-context([scale="s"]) {
-  --calcite-accordion-item-spacing-unit: theme("spacing.2");
-  --calcite-accordion-item-padding: var(--calcite-accordion-item-spacing-unit)
-    var(--calcite-accordion-item-spacing-unit);
-  @apply text--2h;
-}
-
-:host-context([scale="m"]) {
-  --calcite-accordion-item-spacing-unit: theme("spacing.3");
-  --calcite-accordion-item-padding: var(--calcite-accordion-item-spacing-unit)
-    var(--calcite-accordion-item-spacing-unit);
-  @apply text--1h;
-}
-
-:host-context([scale="l"]) {
-  --calcite-accordion-item-spacing-unit: theme("spacing.6");
-  --calcite-accordion-item-padding: var(--calcite-accordion-item-spacing-unit)
-    var(--calcite-accordion-item-spacing-unit);
-  @apply text-1h;
-}
-
 // icon position variants for child
 :host([icon-position="start"]) {
   --calcite-accordion-item-flex-direction: row-reverse;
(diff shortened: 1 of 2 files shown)


paulcpederson:
screener is being really weird, it's highlighting tabs, which we didn't change? Also it is not highlighting the actual change I just made which is fixing accordion header height when there is no subtitle:

| before | after |
| --- | --- |
| Screen Shot 2021-03-23 at 10 36 49 AM | Screen Shot 2021-03-23 at 10 36 59 AM |

eriklharper:
> screener is being really weird, it's highlighting tabs, which we didn't change? Also it is not highlighting the actual change I just made which is fixing accordion header height when there is no subtitle:
> 
> before	after
> Screen Shot 2021-03-23 at 10 36 49 AM	Screen Shot 2021-03-23 at 10 36 59 AM

The height of each row and the entire container does appear different in the after shot.  The after shot shows the text appears more centered in the row.  Is that not intended?

paulcpederson:
@eriklharper no, that's the intentional change. screener doesn't think they're different, but they clearly are.

macandcheese:
giphy-downsized

src/components/calcite-accordion-item/calcite-accordion-item.scss
```diff
@@ -1,92 +1,88 @@
 // host-content() is being used to support IE11 styling in this component.
```
jrowlingson, line 1: Can we remove these IE hacks?

eriklharper:
Weird... I wonder if our pixel diffing threshold needs to be adjusted.  Can you include the link that shows the problem screenshot?  I'd like to pass that onto Loyal to see what he thinks.

src/components/calcite-accordion-item/calcite-accordion-item.scss
```diff
@@ -157,63 +152,61 @@
...
+  @apply flex-col flex-grow my-0 py-0;
   margin-right: auto;
-  flex-direction: column;
-  flex: 1;
   text-align: initial;
 }
 
 :host([dir="rtl"]) .accordion-item-header-text {
+  @apply mr-0;
   margin-left: auto;
```
jrowlingson, line 162: ml-auto

src/components/calcite-accordion-item/calcite-accordion-item.scss
```diff
@@ -157,63 +152,61 @@
 
 // accordion item title
 :host .accordion-item-header-text {
+  @apply flex-col flex-grow my-0 py-0;
   margin-right: auto;
```
jrowlingson, line 156: mr-auto

src/components/calcite-accordion-item/calcite-accordion-item.scss
```diff
@@ -157,63 +152,61 @@
...
+  @apply text-color-2 font-medium;
 }
 :host .accordion-item-subtitle {
-  color: var(--calcite-ui-text-3);
+  @apply text-color-3 mt-1 leading-snug;
 }
 :host([dir="rtl"]) .accordion-item-title {
-  margin-right: 0;
+  @apply mr-0;
   margin-left: auto;
```
jrowlingson, line 178: ml-auto et al

julio8a:
merged this into accordion

Issue #1506, opened 2021-02-03
Enhancement: Make option for icon-only buttons responsive to container width


mitc7862:
### Description
I was able to make a grid of action buttons that fill the width of a panel, but the icon's doesn't appear centered because the calcite icon button width is restricted. 

What we want it to look like:
![image](https://user-images.githubusercontent.com/12564905/106818060-9d626e80-662c-11eb-9e09-01dd4aa47fea.png)

As close as we could get:
![image](https://user-images.githubusercontent.com/12564905/106817890-5d02f080-662c-11eb-9e7a-e10ff0eed519.png)

### Acceptance Criteria 

- [ ] Icon-only buttons no longer restrict max width.

#### Which Component

- Calcite Icon button
- Calcite Action

#### Example Use Case

See screenshot above.

macandcheese:
I think restricting to a square (or a rectangle when icon-start and icon-end are populated), should still be available somehow.

mitc7862:
Agree, I would suggest that still be the default behavior… but I think all components should have a responsive state for varying-width containers.

asangma:
I think you could use Action as a starting place.

That being said, I think this could be handled by using a layout attribute or some layout attributes.

If we wanted to start small to get a feel for it, we could add a layout attribute to ActionGroup to help accomplish that design. I have a couple other designs that would benefit from that.

Also, Action already has a `text-enabled` attribute that can be set to `false` for the icon only option.

asangma:
Also, this would allow users to have an ActionPad with a grid layout.

asangma:
Looks like Action might also need a layout property.

asangma:
@mitc7862 Do you mind if I rename this and update the first comment to be more about making layout options specific to the ActionGroup (with a note about this having the potential of becoming a broader pattern)?

macandcheese:
I think a regular button still needs to have this option, since it’s basically styling users might want to just update a prop on a button. 

Not that it would preclude action bar from also having a layout like “grid” in addition to horizontal or vertical that could be used like this.

asangma:
@macandcheese Seems like the Button should be a bit more prescriptive considering it's usage, but perhaps I'm missing something. Let's connect on this.

asangma:
I'll make a new issue specific to Action and ActionGroup so if we wanna do something similar in Button, we can do that too.

AdelheidF:
@dhrumil83 verified this.

macandcheese:
Valid

PR #1811, opened 2021-03-19
feat(action-group, action): add grid layout to action-group. add center layout to action

asangma:
**Related Issue:** #1506

## Summary
Adds a grid layout option to ActionGroup.
Also adds a related center layout option to Action.
Note: centered layout in Action only applies when text-enabled is false.

Enables layouts like this:
image


asangma:
I know we wanted to stop using attribute selectors for our styling and instead apply classes based on those attributes.
I will need to add a node to receive those classes and would prefer to do that in another pass so as to reduce the risk of regressions at this time.

src/components/calcite-action-group/calcite-action-group.tsx
```diff
@@ -24,6 +25,16 @@ export class CalciteActionGroup {
...
 
+  /**
+   * Indicates the horizontal, vertical, or grid layout of the component.
+   */
+  @Prop({ reflect: true}) layout?: Layout = "vertical";
+
+  /**
+   * Indicates number of columns.
+   */
+  @Prop({ reflect: true}) columnCount?: ColumnCount;
```
driskull, line 36: nitpick: maybe just `columns`?

src/components/calcite-action-group/calcite-action-group.tsx
```diff
@@ -24,6 +25,16 @@ export class CalciteActionGroup {
    */
   @Prop({ reflect: true }) expanded = false;
 
+  /**
+   * Indicates the horizontal, vertical, or grid layout of the component.
+   */
+  @Prop({ reflect: true}) layout?: Layout = "vertical";
```
driskull, line 31: nitpick: The property is optional but has a default value so the optional isn't needed.

driskull (approved):
Looks good. Can we get some tests first? or a follow up issue for tests?

src/components/calcite-action-group/calcite-action-group.tsx
asangma: I like it. Thanks!

src/components/calcite-action-group/calcite-action-group.tsx
asangma: Ah...good catch.

PR #1812, opened 2021-03-19
fix(tip): Hide tip when dismissed.

driskull:
**Related Issue:** None

## Summary

fix(tip): Hide tip when dismissed.

Currently, closing the tip still shows the close button and border.

https://esri.github.io/calcite-components/?path=/story/components-app-tip--basic


driskull:
For things like this, it might be best if we didn't apply styling on the host. Since those styles can be modified by the user, a user could override the display:none styling.

driskull:
I can haz merge?

driskull:
@jcfranco can you review?

jcfranco (approved):
Awesome!

Issue #1822, opened 2021-03-22
Bug: Panel Actions styles are not as expected


subgan82:
### Summary

The chart actions on the top are not in correct position should be on right 

![2021-03-22_15-42-59](https://user-images.githubusercontent.com/1017712/112057767-b74a0900-8b27-11eb-8ea5-7cce28b4d513.png)

### Actual Behavior

The chart actions on the top are in left

### Expected Behavior
The chart actions on the top  should be on right

### Reproduction Steps

https://devtesting.mapsdevext.arcgis.com/apps/mapviewer/index.html?webmap=44dfa082b7cd4104a2bada8e093305c7

Open charts from charts list or using layer charts right action. Open the histogram chart, the issue can be seen

1.

asangma:
@subgan82 FYI, this is an issue with Panel.

asangma:
@subgan82 Installed.

subgan82:
@asangma will this be in v1.0.0-beta.52

asangma:
@subgan82 I'm thinking it can be. Would be good to test it out in `@next` yeah?

PR #1825, opened 2021-03-22
fix(panel): retain header node even without content to retain action position

asangma:
**Related Issue:** (#1822)

## Summary
Alternatively, I could add conditional styles for there not being a header, but this seemed cleaner.


driskull (approved):
👍

asangma:
Dah...looks like I'll have to revert this and do fancy styles to deal with no header since disallowing a Panel without a header is a breaking change.

asangma:
Reverted conditional rendering of header content node.
Updated style instead.

Issue #1819, opened 2021-03-22
Enhancement: improve calcite-action-bar overflow menu handling


AdelheidF:
### Description
There was a new slot added to action-group for the overflow-menu.

```html
           <calcite-action-group>
                <calcite-action text="Save" icon="save" indicator>
                </calcite-action>
                <calcite-action text-enabled icon="map" text="New" slot="menu-actions">
                </calcite-action>
                <calcite-action text-enabled icon="collection" text="Open" slot="menu-actions">
                </calcite-action>
            </calcite-action-group>
```

But it requires the app to count all actions and groups in the action bar and guess their height to know how many actions need to be added to the slot in order to avoid scrollbars. Instead this logic should be handled by calcite, especially since it knows the height of action items and groups.

Maybe each action item in a group could have a property that indicates that this item could be moved into the overflow if there is not enough space for all. E.g. bottom most action item with that property goes first into the overflow list. If there are several groups with those action items, it could start with the bottom most group first.
Or just the group has a new property and it would automatically move action items into the overflow starting with the bottom one. 

Adding Alan here since I'm not clear what the expected order of things moving into the overflow should be, especially since the left action bar in map viewer beta has 4 groups.

https://github.com/Esri/calcite-components/pull/1510
https://github.com/Esri/calcite-components/issues/1466

cc @asangma

asangma:
This seems like it could be a generalized utility or something like that.

**OverflowManager**
* wraps the main list of things
* has a limit property : [ n | "auto" ]*

**OverflowContainer**
* container that receives the overflowing items
* maybe this is a wrapper for Popover?
* probably also needs to be able to insert items into a nested component
  *  e.g. if Actions need to get added to an ActionGroup inside of the OverflowContainer


Maybe it's something like this?
```
<calcite-overflow-manager limit="auto" overflow-target="overflow-thing">
  <calcite-action-group>
    <calcite-action text="cool"></calcite-action>
    <calcite-action text="cool"></calcite-action>
    <calcite-action text="cool"></calcite-action>
    <calcite-action text="cool" overflow-me></calcite-action>
    <calcite-action text="cool" overflow-me></calcite-action>
  </calcite-action-group>
</calcite-overflow-manager>
```

Elsewhere

```
<calcite-overflow-container id="overflow-thing" target-child="target-list">
  <calcite-action-group id="target-list">
  </calcite-action-group>
</calcite-overflow-container>
```

*Not really sure how "auto" would work.

[...]

PR #1828, opened 2021-03-22
feat(action-bar): Automatically overflow actions into a menu depending on available height

driskull:
**Related Issue:** #1819

## Summary

Enhancement: improve calcite-action-bar overflow menu handling #1819

### Breaking Changes

- fix(action-menu)!: Replace 'intlOptions' property with 'label'. // BREAKING
- fix(action-group)!: Replace 'intlOptions' property with 'intlMore'. // BREAKING

### Features

- feat(action-bar): Automatically overflow actions into a menu depending on available height.
- feat(action-menu): Add 'tooltip' slot for adding an tooltip for the menu.
- feat(action-bar): Add overflowActionsDisabled property. Allows scrolling again and disables resizeObserver.

### Notes

- Adds "@types/resize-observer-browser": "0.1.5" for resizeObserver typescript typings
- Removes scrolling on action-bar
- Adds label prop for `calcite-action-menu` since `intlOptions` does not make sense since each group contains a different set of actions that need different labeling. This is a breaking change.
- `calcite-action-group` has an `intlMore` prop for changing the text for the menu.
- Update demos

### Questions

- Should we do anything to `calcite-action-pad` to remove it's scrolling support? I'm not sure nested menus make sense for good UX.

Mar-24-2021 12-11-27


asangma (changes requested):
Sorry if I missed this.
We should have a default string for the overflow menu.
Suggest that string be "More".

image

asangma:
Another thing. 
Right now, it appears that the actions are moved to the overflow from the top action-group first. It seems like we'd want the bottom (last) action-group to start pushing actions to overflow first. And the top (first) action-group should show all actions for as long as possible.

The first shall be last and the last shall be first.

driskull:
> Right now, it appears that the actions are moved to the overflow from the top action-group first.

Not true, it moves from largest group first, then bottom up. The group must also have at least 3 actions to overflow.

driskull:
Removed sorting by group, but that example will still show the same.

asangma:
@driskull Ah...I see.

Do you think there's a reasonable way to make it so that (excluding the bottom-actions slot), the automatic overflow would prioritize as described above?

This is based on the assumption that the highest priority action-group in an app will likely be first in an action-bar, and that action-group will probably often have the most actions in it.

asangma (approved):
Removing my block on this PR.
`But I think prioritizing action-groups based on their order in the DOM would be a good idea.~

Other than that, this is pretty siiiick!

asangma:
Updated after more testing.
Sorry for the noise.

Looks rad!

driskull:
@jcfranco can you review?

jcfranco:
@driskull Before reviewing, are all of the different fixes and features meant to be squashed? Multiple, independent commits in a single PR don't mesh well with our auto changelog tooling.

driskull:
@jcfranco how should I separate or combine them?

jcfranco:
> how should I separate or combine them?

If you want to keep them separate and isolated, you'll need some git-fu to break off commits/code chunks from the history and submit PRs for each one. Otherwise, you'll have to squash all of the changes and use the best commit type and message to describe all of the changes, making sure to include breaking change info in the commit body (if applicable). 

I know it's more work, but the first option seems the most ideal to me in that it adheres to conventional commits, keeps PRs focused and the changelog will have an entry for each relevant change.

driskull:
Opened:

- #1869
- #1870
- #1871 
- #1872 
- #1873

Issue #936, opened 2020-09-02
Bug: calcite-input type=number / up/down arrows create rounding issues


AdelheidF:
### Summary

```
        <calcite-input
          type="number"
          value={`${value}`}
          scale="s"
          onCalciteInputInput={this.handleValueChange}
        ></calcite-input>
```

If I start e.g. with value 10.123 and click the up arrow a few times the number displays with too many decimals. Adding `step=1` doesn't change that.

<img src=https://user-images.githubusercontent.com/44481292/92045735-200d9780-ed36-11ea-81e4-01d18cae4a36.png width=200/>
<img src=https://user-images.githubusercontent.com/44481292/92045747-2865d280-ed36-11ea-904a-b26b939dfdd1.png width=200/>
<img src=https://user-images.githubusercontent.com/44481292/92045759-31ef3a80-ed36-11ea-89ce-68c7428772d3.png width=200/>

### Reproduction Steps

1.

AdelheidF:
I still see this happening...

AdelheidF:
This is still happening.

AdelheidF:
This is still happening.

paulcpederson:
@AdelheidF I've opened a pr to fix this because I can't stand to watch you battle the github bot any longer

paulcpederson:
@AdelheidF I think this made it into beta.52

AdelheidF:
yes, this is good now.

PR #1830, opened 2021-03-23
fix(input): solve bug with decimal precision in number input (#936)

paulcpederson:
**Related Issue:** #936

## Summary

Fixes bug in input where decimal precision got all wonky if you clicked the input a whole bunch of times. I think this bug is due to javascript being... bad at math. Setting a fixed decimal length which is the same as the current value solves the issue.


macandcheese (approved):
💣

Issue #1121, opened 2020-10-13
calcite-radio-button: shows with both buttons unselected


AdelheidF:
I have a new issue that I'm only seeing with `@next`. It works fine on devext, it's using `beta.40` there. In both cases I use `calciteRadioButtonGroupChange`.

The first time I open the panel my correct radio button is selected. When I open the panel again, without making any changes, none of the radio buttons is selected. 
```
          <calcite-radio-button-group
            class={CSS.variableFieldRadio}
            name="variableType"
            scale="s"
            afterCreate={(node: HTMLElement) => {
              node.addEventListener("calciteRadioButtonGroupChange", (event: CustomEvent) => {
                this.handleTypeChange(event.detail);
              });
            }}
          >
            <calcite-radio-button checked={isField} value="field">
              {i18n.ui.field}
            </calcite-radio-button>
            <calcite-radio-button checked={!isField} value="date">
              {i18n.ui.customDate}
            </calcite-radio-button>
         </calcite-radio-button-group>
 ```

<img src=https://user-images.githubusercontent.com/44481292/95920999-52fb7000-0d65-11eb-9b1d-981bb4e49f89.png width=200/><img src=https://user-images.githubusercontent.com/44481292/95921031-627ab900-0d65-11eb-919a-0360a119164f.png width=200/>

I see this in both places (Age + Rotation) where I use radio buttons. 

first time:
<img src=https://user-images.githubusercontent.com/44481292/95921349-fc426600-0d65-11eb-9108-0e191bf0b18f.png width=250/>

second time:
<img src=https://user-images.githubusercontent.com/44481292/95921272-d6b55c80-0d65-11eb-81bc-cea9be8b484c.png width=250/>

eriklharper:
It looks like it is still that lingering issue where the component isn't initializing completely the second time around, evidenced by the lack of a `calcite-label` in the dom.

AdelheidF:
I had marked this as high, but it didn't make it into Map Viewer Beta 10 tonight. 

This workflow is working fine currently on production, but not on devext. 

https://www.arcgis.com/apps/mapviewer/index.html?url=https://maps.iema.state.il.us/arcgis/rest/services/NPP/WIND/FeatureServer/1

https://devext.arcgis.com/apps/mapviewer/index.html?url=https://maps.iema.state.il.us/arcgis/rest/services/NPP/WIND/FeatureServer/1

[...]

PR #1832, opened 2021-03-23
build(deps): bump versions

driskull:
build(deps): bump versions


jcfranco (approved):
🌮 🎉

paulcpederson:
@eriklharper  will this fix #1121 ?

paulcpederson (approved):

eriklharper:
> @eriklharper will this fix #1121 ?

Doubt it.  Ionic still has not addressed this underlying bug with relocating default slotted content in scoped components: https://github.com/ionic-team/stencil/issues/2801.

Based on recent discussions I've been having about labelling form components, I think we should deprecate built-in label support in `calcite-radio-button` (and any other component that is currently doing this like `calcite-checkbox`) and have users supply their own label instead.  Because both `calcite-radio-button` and `calcite-label` components are `scoped` and use default `<slot />`s, this issue could still persist inside `calcite-label` even after the built-in labelling support is removed from said components, but this would simplify them and remove the need for a fix on Ionic's end.

As far as how to handle the Stencil bug that is still affecting `calcite-label`, a possible solution would be to turn `calcite-label` back into a shadowed component, but then require the user to pass their own `<label>` tag into the default slot so that `calcite-label` can style it and keep the label available for form and accessibility on the document.

The crux of this Stencil bug is that it is difficult to replicate slot functionality for non-shadow DOM components because manual dom manipulation is required, which interferes with VDOM tools, which are designed to strictly control the DOM structure.

PR #1834, opened 2021-03-23
docs(modal): fix 'none' option values for modal color attribute

jcfranco:
**Related Issue:** N/A

## Summary

Uses attribute value to unset the color attribute. Stories use attributes instead of props, so `null` would get stringified when rendered:

```html
<calcite-modal
  active=""
  color="null"
  background-color="white"
  scale="m"
  width="s"
  intl-close="Close"
>/*...*/</calcite-modal>
```


driskull (approved):
👍

PR #1835, opened 2021-03-23
fix(panel): header-actions--end layout fix when header-content is not rendered

asangma:
**Related Issue:** #1822

## Summary
Updates styles to actually fix the layout issue.
Actions should all be in the right place whether or not header-content is rendered.

cc @subgan82


driskull (approved):
👍

asangma:
Thanks @driskull!

Issue #1527, opened 2021-02-09
Bug: calcite-pick-list - can't enter list via keyboard after search if nothing is selected


AdelheidF:
### Summary
Initially nothing is selected in my list
<img src=https://user-images.githubusercontent.com/44481292/107304412-d9833e00-6a35-11eb-9759-c201b7ec5c18.png width=200/>

Then I start a search
<img src=https://user-images.githubusercontent.com/44481292/107304460-ed2ea480-6a35-11eb-9130-b0910f33bea4.png width=200/>

Now I can't use the keyboard to get to the only visible pick-list-item. When using the tab key it jumps from the 'delete search' icon to the 'Add' button.

(2) A second issue I just noticed is that if I only use the keyboard and click on the 'delete search' x button I lose the focus from the pick-list.

### Reproduction Steps

1.

jcfranco:
Installed. @AdelheidF could you help verify in the upcoming @next update? 🙇‍♂️

AdelheidF:
The first issue is fixed, but the second isn't.

> (2) A second issue I just noticed is that if I only use the keyboard and click on the 'delete search' x button I loose the focus from the pick-list

jcfranco:
@AdelheidF (2) should be fixed now. 

Could you help verify w/ @next when you get a chance? Thanks!

AdelheidF:
This works fine now, tested locally with next.129. Thanks.

PR #1837, opened 2021-03-24
fix(filter): keep focus on filter after clearing text via mouse

jcfranco:
**Related Issue:** #1527

## Summary

This addresses lost focus for anyone clearing the filter text by clicking the clear button.

cc @AdelheidF


driskull (approved):
👍

Issue #1831, opened 2021-03-23
Bug: Remove setting `dir=` on components


driskull:
### Summary

Bug: Remove setting `dir=` on the host of components

### Actual Behavior

Components are setting the `dir` attribute on their host and child components. When they do so, it makes it difficult to switch direction in an application because the author has to remove all of these `dir` attributes everywhere.

### Expected Behavior

Instead, we should be using the utility class or only setting `dir` on shadowed elements.

https://github.com/Esri/calcite-components/blob/master/src/utils/resources.ts#L2

### Relevant Info 

62 results - 40 files

- [x] src/components/calcite-accordion-item/calcite-accordion-item.tsx: @driskull 
- [x] src/components/calcite-alert/calcite-alert.tsx: 
- [x] src/components/calcite-avatar/calcite-avatar.tsx: @driskull 
- [x] src/components/calcite-button/calcite-button.tsx: @driskull 
- [x] src/components/calcite-card/calcite-card.tsx: @driskull 
- [x] src/components/calcite-checkbox/calcite-checkbox.tsx: @driskull 
- [x] src/components/calcite-chip/calcite-chip.tsx:
- [x] ~~src/components/calcite-color-picker/calcite-color-picker.tsx:~~ Not needed.
- [x] ~~src/components/calcite-color-picker-hex-input/calcite-color-picker-hex-input.tsx:~~ Not needed.
- [x] src/components/calcite-combobox/calcite-combobox.tsx: @driskull 
- [x] src/components/calcite-combobox-item/calcite-combobox-item.tsx: 
- [x] src/components/calcite-combobox-item-group/calcite-combobox-item-group.tsx: @driskull 
- [x] src/components/calcite-date-picker/calcite-date-picker.tsx:
- [x] src/components/calcite-date-picker-day/calcite-date-picker-day.tsx:
- [x] ~~src/components/calcite-date-picker-month/calcite-date-picker-month.tsx:~~ Not needed.
- [x] src/components/calcite-date-picker-month-header/calcite-date-picker-month-header.tsx:
- [x] src/components/calcite-dropdown/calcite-dropdown.tsx: @driskull 
- [x] src/components/calcite-dropdown-group/calcite-dropdown-group.tsx: @driskull 
- [x] src/components/calcite-dropdown-item/calcite-dropdown-item.tsx:
- [x] ~~src/components/calcite-fab/calcite-fab.tsx:~~ Not needed.
[...]

PR #1838, opened 2021-03-24
fix(chip): stop setting dir on component, use class

caripizza:
**Related Issue:** #1831 

## Summary
This pr removes the dir attribute from `calcite-chip`, and refactors its styles to use the rtl class instead.

Also fixes the demo html's:
- dismissible event listener script
- removes erroneous `active` attribute
- puts the rtl demo chips inside a parent `section` element (which has `dir="rtl"` applied)

### DIFF
--- a/src/components/calcite-chip/calcite-chip.tsx
+++ b/src/components/calcite-chip/calcite-chip.tsx
@@ -11,6 +11,7 @@ import {
 } from "@stencil/core";
 import { getElementDir } from "../../utils/dom";
 import { guid } from "../../utils/guid";
+import { CSS_UTILITY } from "../../utils/resources";
 import { CSS, TEXT } from "./resources";
 import { ChipColor } from "./interfaces";
 import { Appearance, Scale, Theme } from "../interfaces";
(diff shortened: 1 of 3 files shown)


src/components/calcite-chip/calcite-chip.tsx
```diff
@@ -129,7 +130,7 @@ export class CalciteChip {
     );
 
     return (
-      <Host dir={dir}>
+      <Host class={{ [CSS_UTILITY.rtl]: rtl }}>
```
driskull, line 133: @caripizza Can we add a container div and apply the class to that instead of the host?

That way the styles are internal to the component (shadow DOM) and can't be affected or removed by the end user.

src/components/calcite-chip/calcite-chip.tsx
```diff
@@ -104,12 +105,12 @@ export class CalciteChip {
...
   render(): VNode {
     const dir = getElementDir(this.el);
+    const rtl = dir === "rtl";
     const iconScale = this.scale !== "l" ? "s" : "m";
 
     const iconEl = (
       <calcite-icon
-        class="calcite-chip--icon"
-        dir={dir}
+        class={`calcite-chip--icon ${rtl && CSS_UTILITY.rtl}`}
```
driskull, line 113: nitpick: 
```jsx
class={{["calcite-chip--icon"]: true, [CSS_UTILITY.rtl]: true}}
```

driskull (changes requested):
Looks good @caripizza 🍕 👍

caripizza:
Closing for now to focus on Tailwind refactors with @julio8a - will open if/when I can implement a better approach on this one.

driskull:
@caripizza do you still want to handle this one?

caripizza:
@driskull I wasn't able to make much progress after you had reviewed this one, but now that docs site is up I'm ready to hop back in! I'll reopen it to get new feedback from you as soon as I can this week.

PR #1840, opened 2021-03-24
fix(action): reset line-height

asangma:
## Summary
Resets line-height back to pre-refactor.
Update was causing descenders to get clipped.

cc @driskull 

Fixes this.
Screen Shot 2021-03-24 at 1 38 25 PM


driskull (approved):
👍

PR #1843, opened 2021-03-25
fix(input): only emit input event on user interaction

jcfranco:
**Related Issue:** N/A

## Summary

This brings `<calcite-input>` closer to `<input>` by only emitting events on user interaction. 

This also helps w/ VDOM integrations where value comes from a state model and is set on the input and not from interaction, so no need to emit in that case. Otherwise, you get two events fired.

cc @AdelheidF


macandcheese (approved):
Noice

eriklharper (approved):
Perfect!  Now I don't need to implement this fix on my time-picker branch.  I was noticing change event firing twice when it shouldn't have.

AdelheidF:
This works fine now, tested locally with next.129. Thanks.

PR #1851, opened 2021-03-26
fix(tooltip): improve tooltip border, tooltip usage in card story, card/link demo html

caripizza:
**Related Issue:** n/a

## Summary
This pr aims to fix a few style issues with `calcite-tooltip`'s Tailwind refactors (pr #1738, pr #1765).

The goal here is to a). ensure the tooltip tail doesn't appear **over** the content area, and b). ensure the tooltip's new border styles are visible.

I also noticed the tooltip wasn't working in a few places (card stories, card/link demo html), so I tried to fix those as well.

### DIFF
--- a/src/assets/styles/_popper.scss
+++ b/src/assets/styles/_popper.scss
@@ -17,7 +17,7 @@ $popper-default-z-index: 900;
     transition-property: transform, visibility, opacity;
     opacity: 0;
     box-shadow: $shadow-2;
-    border-radius: var(--calcite-border-radius);
+    @apply rounded;
   }
 }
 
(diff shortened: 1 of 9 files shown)


caripizza:
@driskull let me know if there's a better to achieve what you were looking for 🌞

driskull (changes requested):
It's looking much better @caripizza!

I'm still seeing an offset on the drop shadow for the arrow though...

Screen Shot 2021-03-26 at 1 03 04 PM


Also, it seems like the tooltip has a border but popover doesn't. Should we be consistent here?

Screen Shot 2021-03-26 at 1 02 17 PM


Zooming in with osx makes it easy to see the differences.

driskull:
I think the popover is missing the rounded corners too.

driskull:
Looking good @caripizza. I see the consistent border now

image

There seems to be a little bit of background bleeding through. What do you think? I'm ok with it if you guys are :)

bstifle:
dang it it's so close!

bstifle:
Tooltip and popover looking 100% perfect on my end. no bleeds

driskull (approved):
👍

PR #1853, opened 2021-03-26
chore(tile-select-group): update story to make sure docs visible

caripizza:
**Related Issue:** # n/a

## Summary
I noticed Storybook's showing a blank page for tile-select-group's docs, so I've added it in here.


driskull (approved):
👍

Issue #1575, opened 2021-02-18
Bug: Changing `disabled` attribute of calcite-switch visually removes disabled, but the toggle will not work.


ethanbdev:
### Summary
![12F8FA83-DBD9-47C7-A6B2-DEF7D32BE3F6](https://user-images.githubusercontent.com/52869490/108392504-a3795300-71c7-11eb-98c7-ba7559964b75.GIF)
It seems like the input in the shadow dom has disabled persisted even though disabled has been removed from the switch itself.
![image](https://user-images.githubusercontent.com/52869490/108391962-12a27780-71c7-11eb-8695-ed7f60fd9c39.png)
Couldn't reproduce using straight JS in a codepen, but can using Stencil.
### Actual Behavior
Switch is not disabled but cannot be toggled
### Expected Behavior
Switch should be "switchable"
### Reproduction Steps
See this component:
https://github.com/ethanbdev/calcite-switch-disabled-bug/tree/main/src/components/my-component
(clone, npm install, npm start)

driskull:
It seems like the `disabled` property needs to be watched for changes and update the input. @macandcheese @paulcpederson @eriklharper who can take this one?

eriklharper:
I can take a look at this one.  Would be good while I'm in there to remove the proxy input stuff too.

ethanbdev:
> I can take a look at this one. Would be good while I'm in there to remove the proxy input stuff too.

Thanks Erik, if you do remove the proxy input can that be in the same release as this fix? I am using that to workaround, like so:
```ts
<calcite-switch switched={this.enableLogarithmic === true && this.minBound > 0} disabled={this.minBound <= 0} onCalciteSwitchChange={this.enableLogScale}>
    <input disabled={this.minBound <= 0} type="checkbox" />
</calcite-switch>
```
Thanks for taking a look!

eriklharper:
Sure thing Ethan!  I've done a proxy input removal before, and it is pretty straightforward, so definitely doable in the same release 👍

MikeTschudi:
I bumped into this in parallel. What I saw was if the `calcite-switch` is initially disabled, it stays permanently disabled because the "disabled" attribute is added to the `<input>` tag when the component is created (in addition to the `<calcite-switch>` tag). If initially enabled, it's possible to change its disable state because "disabled" attribute changes only affect the `<calcite-switch>`.

Here's a web page that reproduces the behavior outside of Stencil:
```html
<!doctype html>
<html>
[...]

PR #1854, opened 2021-03-26
fix(calcite-switch)!: behavior no longer breaks when changing disabled.  BREAKING CHANGE: removes proxy input support

eriklharper:
**Related Issue:** #1575

This is a BREAKING CHANGE.  It removes proxy input support for `calcite-switch` and takes full ownership over the native `<input>` element used to respond to toggle changes.


src/components/calcite-switch/calcite-switch.tsx
```diff
@@ -21,34 +22,107 @@ import { Scale, Theme } from "../interfaces";
...
+  @Watch("disabled")
+  disabledWatcher(newDisabled: boolean): void {
+    this.inputEl.disabled = newDisabled;
+  }
+
+  /** The id attribute of the switch.  When omitted, a globally unique identifier is used. */
+  @Prop({ reflect: true, mutable: true }) guid: string;
+
+  /** The name of the switch input */
+  @Prop({ reflect: true }) name?: string = "";
```
driskull, line 51: Does the name need to be an empty string by default? Just thinking that the input shouldn't have a name attribute unless its defined.

src/components/calcite-switch/calcite-switch.tsx
```diff
@@ -21,34 +22,107 @@ import { Scale, Theme } from "../interfaces";
...
 
-  /** The name of the checkbox input */
-  @Prop({ reflect: true, mutable: true }) name?: string = "";
+  @Watch("disabled")
+  disabledWatcher(newDisabled: boolean): void {
+    this.inputEl.disabled = newDisabled;
+  }
+
+  /** The id attribute of the switch.  When omitted, a globally unique identifier is used. */
+  @Prop({ reflect: true, mutable: true }) guid: string;
```
driskull, line 48: Shouldn't this be an internal state instead? We don't want users to set guid, correct?

src/components/calcite-switch/calcite-switch.tsx
```diff
@@ -21,34 +22,107 @@ import { Scale, Theme } from "../interfaces";
...
+    focusElement(this.inputEl);
+  }
+
+  //--------------------------------------------------------------------------
+  //
+  //  Private Methods
+  //
+  //--------------------------------------------------------------------------
+
+  private updateSwitch(): void {
```
driskull, line 100: nitpick: rename to `toggle()` or `toggleSwitch()`?

src/components/calcite-switch/calcite-switch.tsx
```diff
@@ -60,109 +134,76 @@ export class CalciteSwitch {
...
-  componentWillRender(): void {
-    this.syncProxyInputToThis();
+  // --------------------------------------------------------------------------
+  //
+  //  Render Methods
+  //
+  // --------------------------------------------------------------------------
+
+  private renderInput(): void {
+    this.inputEl = document.createElement("input");
```
driskull, line 175: This input is created each time the `connectedCallback` is called which could be if the component is moved and would lead to multiple input elements being created, is that correct? I think it should only create and place the stored input once.

src/components/calcite-switch/calcite-switch.tsx
```diff
@@ -60,109 +134,76 @@ export class CalciteSwitch {
...
   render(): VNode {
     const dir = getElementDir(this.el);
-
     return (
-      <Host
-        aria-checked={this.switched.toString()}
-        dir={dir}
-        tabIndex={this.disabled ? -1 : this.tabIndex}
-      >
+      <Host aria-checked={this.switched.toString()} dir={dir} tabindex={this.disabled ? -1 : 0}>
```
driskull, line 202: nitpick: can we add these props on an internal node so they are encapsulated instead of in light DOM?

src/components/calcite-switch/calcite-switch.tsx
```diff
@@ -21,34 +22,107 @@ import { Scale, Theme } from "../interfaces";
...
-  @Prop({ reflect: true, mutable: true }) value?: string = "";
+  @Prop({ reflect: true }) value?: string;
+
+  //--------------------------------------------------------------------------
+  //
+  //  Private Properties
+  //
+  //--------------------------------------------------------------------------
+
+  private inputEl: HTMLInputElement;
```
driskull, line 81: nitpick: can you just create the element here?

driskull (commented):
Looking good @eriklharper. Just some comments

src/components/calcite-switch/calcite-switch.tsx
eriklharper: Yes, they would set the `id` attribute only.  Good catch.

src/components/calcite-switch/calcite-switch.tsx
```diff
@@ -60,109 +141,85 @@ export class CalciteSwitch {
...
-  private inputProxy: HTMLInputElement;
+  //--------------------------------------------------------------------------
+  //
+  //  Lifecycle
+  //
+  //--------------------------------------------------------------------------
 
   connectedCallback(): void {
-    this.setupProxyInput();
+    this.guid = this.el.id || `calcite-switch-${guid()}`;
```
driskull, line 171: I'm not sure if this is a big deal, but the guid will change if a component is reconnected.  Maybe it should be in `componentWillLoad`? 

https://stenciljs.com/docs/component-lifecycle

src/components/calcite-switch/calcite-switch.tsx
```diff
@@ -60,109 +141,85 @@ export class CalciteSwitch {
...
 
-  componentWillRender(): void {
-    this.syncProxyInputToThis();
+  // --------------------------------------------------------------------------
+  //
+  //  Render Methods
+  //
+  // --------------------------------------------------------------------------
+
+  private renderInput(): void {
```
driskull, line 184: nitpick: I think the `render` methods are for returning `VNode`s. Should we rename and move it into private methods?

src/components/calcite-switch/calcite-switch.tsx
```diff
@@ -60,109 +141,85 @@ export class CalciteSwitch {
...
+    this.switched && this.inputEl.setAttribute("checked", "");
+    this.inputEl.disabled = this.disabled;
+    this.inputEl.id = `${this.guid}-input`;
+    this.inputEl.name = this.name;
+    this.inputEl.style.setProperty("bottom", "0", "important");
```
driskull, line 189: nitpick: Might be nice to just set the whole css incase there are any other styles that might be problematic?

```
.style.cssText = `
bottom: 0 !important;
...
`
```

Do you think this would be useful for other components as a utility?

src/components/calcite-switch/calcite-switch.tsx
```diff
@@ -60,109 +141,85 @@ export class CalciteSwitch {
...
-        dir={dir}
-        tabIndex={this.disabled ? -1 : this.tabIndex}
-      >
-        <div class="track">
-          <div class="handle" />
+      <Host>
+        <div
+          aria-checked={this.switched.toString()}
+          class="container"
+          dir={dir}
```
driskull, line 215: Can we use a class instead of the dir attribute here? I don't think this CSS is working currently since the styles in the CSS are specified for host direction.

driskull:
```
import { CSS_UTILITY } from "../../utils/resources";

```

```
.calcite--rtl calcite-switch {
  margin-left: 0;
  margin-right: var(--calcite-spacing-half);
}
```

src/components/calcite-switch/calcite-switch.tsx
eriklharper: It would be a nice utility, definitely.  i'm using this same thing in at least two other components, radio and checkbox.

driskull (approved):
👍 Nice 🎉 🌮

caripizza:
Hi @eriklharper - is there a separate issue related to this one that will update the docs and demo html? We're using those a ton with the Tailwind refactors.

Also, I looked at the linked issue but it's closed, was that intentional?

eriklharper:
> Hi @eriklharper - is there a separate issue related to this one that will update the docs and demo html? We're using those a ton with the Tailwind refactors.
> 
> Also, I looked at the linked issue but it's closed, was that intentional?

I'll put up another PR linked to this same original issue to update those docs.  Thanks for spotting that!

eriklharper:
@caripizza PR for docs update: https://github.com/Esri/calcite-components/pull/1895

Issue #1852, opened 2021-03-26
Bug: Cannot type in `color-picker` hex input in Firefox


ethanbdev:
### Summary
Cannot type in the hex input in the color picker component if the user highlights the text and types, and if the user backspaces then types, it works.

Happens in Firefox, not chrome, have not tested other browsers.

### Reproduction Steps
**update**
1. Highlight all of the text in the hex input, then try to type
2. Typing will not input any values, you have to use the backspace and enter.

https://codepen.io/eborgen/pen/OJRmmPO?editors=1000

### Relevant Info 
Firefox 78.9.0esr (64-bit)

bstifle:
getting the same bugs in Brave

jcfranco:
Installed.

jcfranco:
@ethanbdev could you help verify this w/ `@next`? Thanks! 🙇🏼‍♂️

ethanbdev:
Looks good, thanks @jcfranco for the quick fix 👏

PR #1855, opened 2021-03-29
fix(color-picker): ensure consistent handling of hex input text selection

jcfranco:
**Related Issue:** #1852 

## Summary

This refactors hex-input's character input logic to avoid `window.getSelection`, which doesn't work on Firefox when text selection is on a form control: https://bugzilla.mozilla.org/show_bug.cgi?id=85686


driskull (approved):
👍

PR #1858, opened 2021-03-29
fix(button, fab, pick-list, value-list, action-bar, action-pad): Remove internallly setting 'dir' attribute

driskull:
**Related Issue:** #1831

## Summary

fix(button, fab, pick-list, value-list, action-bar, action-pad): Remove internallly setting 'dir' attribute. #1831


driskull:
I'll probably just split these into separate PRs.

Issue #1849, opened 2021-03-26
Expose a CSS var for Panel a base width.


asangma:
### Question
Currently the CSS vars for Panel widths are defined in the shadow.
It could be useful to users to be able to adjust these for their purposes.

I propose we expose these and allow users to overwrite them.
It would probably mean having a var for each width-scale.

Currently it looks like this inside of Panel
```
:host([width-scale="s"]) {
  --calcite-panel-width: 12vw;
  --calcite-panel-max-width: 300px;
  --calcite-panel-min-width: 150px;
}

:host([width-scale="m"]) {
  --calcite-panel-width: 20vw;
  --calcite-panel-max-width: 420px;
  --calcite-panel-min-width: 240px;
}

:host([width-scale="l"]) {
  --calcite-panel-width: 45vw;
  --calcite-panel-max-width: 680px;
  --calcite-panel-min-width: 340px;
}
```

We will offer a base panel-width var at the global level.

```
:root {
  --calcite-panel-width-base: 12vw;
}
```

Existing vars in Panel (width-scale width, max-width, and min-width) for each scale will be be calculated (`calc`) based on that base var to produce the same value that are currently being used.

asangma:
cc @kevindoshier

macandcheese:
I've run up to this frequently recently. 

I'd also encourage allowing override of `flex-direction` in some way for the `.main` div in the shadow root of `calcite-shell`.

Allowing this to be user-configured would mean that I can stack my panels for certain screen sizes as I need to. Currently I need to access that `.main` div via JS and manually change that `flex-direction` property:

![Screen Shot 2021-03-26 at 11 34 31 AM](https://user-images.githubusercontent.com/4733155/112677831-6f78f980-8e27-11eb-94af-dbc4c765352d.png)
![Screen Shot 2021-03-26 at 11 34 20 AM](https://user-images.githubusercontent.com/4733155/112677835-70119000-8e27-11eb-9eb3-1eb9f4e486dd.png)


Combined with setting 100% width for `shell-panel` and `panel`, this allows me to "fake" mobile pretty well, but allowing it as in the example you provide would be way better.

asangma:
@macandcheese I could see that.

I would suggest utilizing our Layout interface and call the property `layout`.

That being said, I think with your mobile stuff, there's the additional complexity of `content-behind` in Shell.

driskull:
I'm ok with exposing them. Maybe we don't document them for a while though incase we need to refactor them?

[...]

PR #1861, opened 2021-03-29
feat(panel, shell-panel): add width base unit at root level

asangma:
**Related Issue:** #1849

## Summary
This adds a base width unit at the :root level that is then used when calculating the width, min-width, and max-width for Panel and ShellPanel at each scale.

It should not affect the UI or any existing usages of Shell and ShellPanel.
The `width-scale` property should still behave as it has been on both components.

### DIFF
--- a/src/assets/styles/global.scss
+++ b/src/assets/styles/global.scss
@@ -19,6 +19,8 @@
   --calcite-border-radius: 3px;
   --calcite-ui-opacity-disabled: 0.5;
 
+  --calcite-ui-panel-width-unit: 1;
+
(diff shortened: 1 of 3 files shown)


asangma:
If anybody sees a cleaner way to do this, I'm open.

driskull (approved):
This looks good to me. So we would just expose that "unit" multiplier. I just wonder if there's a better name for it? Like modifier or multiplier.

asangma:
"multiplier" is more descriptive. I'm cool with that.

PR #1864, opened 2021-03-30
refactor(panel, shell-panel): rename CSS var to remove -ui- and add testing

asangma:
**Related Issue:** #

## Summary
Remove "ui" from var name.

cc @kevindoshier

### DIFF
--- a/src/assets/styles/global.scss
+++ b/src/assets/styles/global.scss
@@ -19,7 +19,7 @@
   --calcite-border-radius: 3px;
   --calcite-ui-opacity-disabled: 0.5;
 
-  --calcite-ui-panel-width-multiplier: 1;
+  --calcite-panel-width-multiplier: 1;
 
   @apply font-sans;
 }
--- a/src/components/calcite-panel/calcite-panel.e2e.ts
+++ b/src/components/calcite-panel/calcite-panel.e2e.ts
@@ -280,4 +280,45 @@ describe("calcite-panel", () => {
 
     expect(footer).toBeNull();
   });
+
+  it("should update width based on the multipier CSS variable", async () => {
+
+    const multipier = 2;
+
+    const page = await newE2EPage();
+
+    await page.setContent(`
+      <calcite-panel width-scale="m">
+        test
+      </calcite-panel>
+    `);
+
+    await page.waitForChanges();
+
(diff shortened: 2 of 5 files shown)


driskull (commented):
@asangma code changes look good 👍

Do you think we should update this story: https://esri.github.io/calcite-components/?path=/story/overview-custom-theme--interactive-example

Also, I we should probably setup a test for this multiplier to make sure it's working. 

Related to https://github.com/Esri/calcite-components/issues/1537

driskull:
How are we documenting these variables? Should we add them to the component tsx file somehow?

asangma:
Ah...I thought we weren't gonna doc them for now.

asangma:
I could use some help on writing the test.
For now, I was leaning on the updates not breaking Screener tests.

asangma:
> How are we documenting these variables?
The vars inside the Shadow probz wouldn't be doc-ed.
I could see doc-ing the multiplier var at some point.

> Should we add them to the component tsx file somehow?

Not sure what you mean. 🤔

driskull:
For the test, we should be able to use `getComputedStyle()`.

Like this:

```
page = await newE2EPage({
              html: `<div style="text-align: center;">
              <calcite-label dir="ltr">Label text
              <calcite-input></calcite-input>
              </calcite-label>
              </div>`
            });
            element = await page.find("calcite-label label");
            style = await element.getComputedStyle();
            expect(style["textAlign"]).toEqual("left");
```

So we would just test the computed style and then set the variable to something else and test the computed style again for the changes.

driskull:
> The vars inside the Shadow probz wouldn't be doc-ed.
I could see doc-ing the multiplier var at some point.

Ok, lets just not document the multiplier for now. I was just wondering how we would document it in the future. We don't document the css variables anywhere but that storybook, right?

asangma:
> Ok, lets just not document the multiplier for now.

Cool. I do think we should doc it, but I foresee potential for some changes to how the width var might work, so it'd be good to get some internal feedback on how this works and make some tweaks before adding it to doc.

asangma:
@driskull Updated. Awaiting critique.

driskull (approved):
👍 💯

asangma:
yew!

PR #1868, opened 2021-03-31
fix(color-picker): prevent invalid chars from being entered when hex input text is selected

jcfranco:
**Related Issue:** N/A

## Summary

This tweaks up the input processing logic to prevent users from entering invalid characters when the the input has a selection. 

invalid-hex-input-char


driskull (approved):
Looks good! 🎉

PR #1871, opened 2021-03-31
feat(action-group): Add 'menu-tooltip' slot for adding a tooltip to the overflow menu button.

driskull:
**Related Issue:** #1819

## Summary

feat(action-group): Add 'menu-tooltip' slot for adding a tooltip to the overflow menu button.

### DIFF
--- a/src/components/calcite-action-group/resources.ts
+++ b/src/components/calcite-action-group/resources.ts
@@ -1,3 +1,8 @@
 export const SLOTS = {
-  menuActions: "menu-actions"
+  menuActions: "menu-actions",
+  menuTooltip: "menu-tooltip"
+};
+
+export const TEXT = {
+  more: "More"
 };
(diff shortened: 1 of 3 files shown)


src/components/calcite-action-group/resources.ts
```diff
@@ -1,3 +1,8 @@
 export const SLOTS = {
-  menuActions: "menu-actions"
+  menuActions: "menu-actions",
+  menuTooltip: "menu-tooltip"
+};
+
+export const TEXT = {
```
jcfranco, line 6: Is this used in another PR? If not, this can be dropped.

jcfranco (approved):
Can you add a test before merging? Other than that, this looks kewl-ah! 🆒

src/components/calcite-action-group/resources.ts
driskull: Yes it is.

PR #1873, opened 2021-03-31
feat(action-menu): Add 'tooltip' slot for adding an tooltip for the menu.

driskull:
**Related Issue:** #1819

## Summary

feat(action-menu): Add 'tooltip' slot for adding an tooltip for the menu.


jcfranco (approved):
Can you add a test before merging? Other than that, this looks gewd-ah! 👍🏼

Issue #1409, opened 2020-12-16
Bug: Select menu blank after Ember re-render


nwhittaker:
### Summary
Possibly stems from a quirk with how Ember templates are updated, but re-rendering the `CalciteSelect` component causes the menu to appear blank.

### Actual Behavior
After re-render, opening the calcite select menu reveals a list of blank options.

### Expected Behavior
After a re-render, opening the calcite select menu reveals its list of options.

### Reproduction Steps
1. Visit https://jsbin.com/gedivigoqo/1/edit?html,js,output and wait for the **Output** column to render a calcite select component.
2. Open the calcite select menu and notice the two options.
3. Close the menu and click the **hide** button. Click the button again to cause a re-render of the calcite select component.
4. Open the calcite select menu and notice the two options are now blank.

### Relevant Info 
Seen in latest macOS versions of Chrome, Firefox, and Safari. Additionally, the following console error appears:
```
TypeError: undefined is not an object (evaluating 'this.selectEl.append')
```

One possible workaround is to populate the `calcite-option` elements' `label` attributes. This appears to have the side-effect of keeping the option values populated.

patrickarlt:
This is also affecting us when using React it looks like the mutation events MIGHT trigger before the select is rendered hence why `this.selectEl` is undefined.

jcfranco:
Installed. 

@nwhittaker, @patrickarlt could you help verify this with `@esri/calcite-components@next`?

nwhittaker:
>@nwhittaker, @patrickarlt could you help verify this with @esri/calcite-components@next?

I updated the js bin to reference the `@next` Calcite tag[1] and I no longer see the console error, but the menu is still blank for me after the second render. Per https://github.com/Esri/calcite-components/issues/1409#issuecomment-751835804, I think the underlying error is that for the second render, the `mutationObserver` is firing before `storeSelectRef()` is called.

For the first render, `mutationObserver` fires after the call to `storeSelectRef()` -- which would explain why we don't see a problem until subsequent renders.

[1] https://jsbin.com/cosalesiya/2/edit?html,js,console,output

jcfranco:
> I think the underlying error is that for the second render, the mutationObserver is firing before storeSelectRef() is called.

`storeSelectRef` executes the method from the mutation observer callback, so I think there's another issue here. I'll take another look.

Thanks for testing!

nwhittaker:
[...]

PR #1878, opened 2021-03-31
fix(option): fix mutation observer options used to track content changes

jcfranco:
**Related Issue:** #1409 

## Summary

This tweaks the mutation observer options to actually track when its text content changes. `characterData` was not working because of the node being observed was not a text node. Dropped `subtree` since option is not expected to have a subtree as content.


driskull (approved):
👏

Issue #947, opened 2020-09-04
Tooling: Please add license and version to distribution


MikeTschudi:
## Summary 
Trying to keep track of our app's dependencies

## Measure of Success 
https://www.npmjs.com/package/@esri/calcite-components/v/1.0.0-beta.38 shows that the distribution is licensed and the calcite.js file shows the license and version number.

currently in npm:
![image](https://user-images.githubusercontent.com/2125181/92278886-fe81ec80-eeaa-11ea-8e50-48c98120d41f.png)

currently in file:
```javascript
import{p as plt,w as win,d as doc,N as NAMESPACE,a as promiseRe...
```

macandcheese:
cc @julio8a

julio8a:
Thanks for bringing this up. This is something we’re working on. Will post updates here.

macandcheese:
Valid

julio8a:
This is still in the works.

julio8a:
Calcite is under the Esri MLA and not open source. The repository and NPM point to the license file in the repository.

Please let me know if there are any questions about this.

MikeTschudi:
@julio8a, I see the license referral in npm, but the license is missing from the distribution:
![image](https://user-images.githubusercontent.com/2125181/123012722-a489df00-d377-11eb-8f88-d5948687f31b.png)

From Keith Ryden:
> Anything we distribute should have a license – different licenses are applicable to different uses.  So if you’re seeing something without a license – whether it’s code, executable binaries, whatever – we need to get a license on it.

jcfranco:
Blocked by https://github.com/ionic-team/stencil/issues/2169. Fortunately, there is a [PR](https://github.com/ionic-team/stencil/pull/2791) for it.

macandcheese:
Might need to ping Stencil as that PR is still open.

jcfranco:
@benelan Moving this to the current sprint as Stencil 2.9.0 fixed the preamble functionality.

PR #1881, opened 2021-03-31
docs(license): update file name and npm license text

julio8a:
**Related Issue:** #947

## Summary

follow same license files and name as jsapi. 

Also adds package json license (point to copyright.txt)

@paulcpederson, you had mentioned that `UNLICENSED` is what we should have, so not sure if this is still applicable.


driskull (approved):
👍

PR #1883, opened 2021-03-31
fix(combobox): Update scrollable menu height for 'maxItems' 

driskull:
**Related Issue:** None

## Summary

fix(combobox): Update scrollable menu height for 'maxItems' when maxItems or active changes.


driskull:
Previously, it was not setting the height correctly because it can't calculate the height until the menu is active and the dom actually has heights to calculate.

paulcpederson (approved):

Issue #1406, opened 2020-12-15
A11y:  Can't use keyboard to navigate color areas in color picker


capeoples:
### Summary
In the color picker, it is currently not possible to tab to the controls that operate the color field and slider (highlighted below).
<img width="319" alt="Screen Shot 2020-12-15 at 2 12 00 PM" src="https://user-images.githubusercontent.com/30473086/102279559-74e9ae00-3ee0-11eb-9b7f-917955c2035d.png">

### Expected Behavior
To make this accessible, a user should be able to tab to the color area buttons and then use the arrow keys to control where that button is in the color field or slider. I've found a color selector component that support this keyboard behavior here: https://app.contrast-finder.org/?lang=en

Note, a non-calcite color picker was flagged for having similar a11y issues in the Level Access accessibility audit done for Enterprise in Sept 2020.

bstifle:
@jcfranco do you have an eta on this?

jcfranco:
@bstifle No, sir. I do not.

paulcpederson:
@bstifle eta is whenever #1885 lands 😄

jcfranco:
Installed. Thanks @paulcpederson for tackling this one! 🎉 

@capeoples could you help verify? It should be available for testing on `@esri/calcite-components@next` or when 1.0.0-beta.52 ships.

jcfranco:
Verified w/ beta.52 update of https://esri.github.io/calcite-components/?path=/story/components-controls-colorpicker--simple.

PR #1885, opened 2021-04-01
fix(color-picker): add keyboard support to color field and hue slider

paulcpederson:
**Related Issue:** #1406

## Summary

I kept getting issues assigned to me about lacking keyboard support in this component, so thought I'd just take a crack at it. I've chosen to model this as multiple `slider` components in terms of accessibility. Because multi-axis sliders are not supported via the `slider` role, when you are moving in the vertical direction, we treat this as a 0-100 value slider, while when you're moving horizontally we treat it as a 0-100 saturation slider.

### DIFF
--- a/src/components/calcite-color-picker/calcite-color-picker.e2e.ts
+++ b/src/components/calcite-color-picker/calcite-color-picker.e2e.ts
@@ -1082,28 +1082,51 @@ describe("calcite-color-picker", () => {
     await assertHiddenSection([]);
   });
 
-  it("allows editing handles via keyboard", async () => {
-    const page = await newE2EPage({
-      html: `<calcite-color-picker value="#beefee"></calcite-color-picker>`
+  describe("scope keyboard interaction", () => {
+    it("allows editing color field via keyboard", async () => {
+      const page = await newE2EPage({
+        html: `<calcite-color-picker allow-empty value=""></calcite-color-picker>`
+      });
+
+      const picker = await page.find("calcite-color-picker");
+      const scope = await page.find(`calcite-color-picker >>> .${CSS.scope}`);
+
+      await scope.press("Tab");
+      expect(await picker.getProperty("value")).toBeFalsy();
+      await scope.press("ArrowDown");
+      expect(await picker.getProperty("value")).toBe("#ffffff");
+      await scope.press("ArrowDown");
+      expect(await picker.getProperty("value")).toBe("#ededed");
+      await scope.press("ArrowDown");
+      expect(await picker.getProperty("value")).toBe("#dbdbdb");
+      await scope.press("ArrowUp");
+      expect(await picker.getProperty("value")).toBe("#ededed");
+      await scope.press("ArrowRight");
+      expect(await picker.getProperty("value")).toBe("#e4eaed");
+      await scope.press("ArrowLeft");
+      expect(await picker.getProperty("value")).toBe("#ededed");
     });
+    it("allows editing hue slider via keyboard", async () => {
+      const page = await newE2EPage({
+        html: `<calcite-color-picker allow-empty value=""></calcite-color-picker>`
+      });
 
-    const picker = await page.find("calcite-color-picker");
-    const spy = await picker.spyOnEvent("calciteColorPickerChange");
+      const picker = await page.find("calcite-color-picker");
+      const scopes = await page.findAll(`calcite-color-picker >>> .${CSS.scope}`);
 
-    await page.keyboard.press("Tab");
-    await page.keyboard.press("ArrowUp");
-    expect(spy).toHaveReceivedEventTimes(1);
-    await page.keyboard.press("ArrowDown");
-    expect(spy).toHaveReceivedEventTimes(2);
-    await page.keyboard.press("ArrowLeft");
-    expect(spy).toHaveReceivedEventTimes(3);
-    await page.keyboard.press("ArrowRight");
-    expect(spy).toHaveReceivedEventTimes(4);
+      await scopes[0].press("Tab");
+      await scopes[1].press("ArrowDown");
+      expect(await picker.getProperty("value")).toBe("#007ec2");
+      await scopes[1].press("ArrowRight");
+      expect(await picker.getProperty("value")).toBe("#007bc2");
+      await scopes[1].press("ArrowLeft");
+      expect(await picker.getProperty("value")).toBe("#007ec2");
 
-    await page.keyboard.press("Tab");
-    await page.keyboard.press("ArrowLeft");
-    expect(spy).toHaveReceivedEventTimes(5);
-    await page.keyboard.press("ArrowRight");
-    expect(spy).toHaveReceivedEventTimes(6);
+      await page.keyboard.press("Shift");
+      await scopes[1].press("ArrowDown");
+      expect(await picker.getProperty("value")).toBe("#0081c2");
+      await scopes[1].press("ArrowUp");
+      expect(await picker.getProperty("value")).toBe("#007ec2");
+    });
   });
 });
--- a/src/components/calcite-color-picker/calcite-color-picker.e2e.ts
+++ b/src/components/calcite-color-picker/calcite-color-picker.e2e.ts
@@ -1081,4 +1081,29 @@ describe("calcite-color-picker", () => {
     await assertHiddenSection(["channels"]);
     await assertHiddenSection([]);
   });
+
+  it("allows editing handles via keyboard", async () => {
+    const page = await newE2EPage({
+      html: `<calcite-color-picker value="#beefee"></calcite-color-picker>`
+    });
+
+    const picker = await page.find("calcite-color-picker");
+    const spy = await picker.spyOnEvent("calciteColorPickerChange");
+
+    await page.keyboard.press("Tab");
+    await page.keyboard.press("ArrowUp");
+    expect(spy).toHaveReceivedEventTimes(1);
+    await page.keyboard.press("ArrowDown");
+    expect(spy).toHaveReceivedEventTimes(2);
+    await page.keyboard.press("ArrowLeft");
+    expect(spy).toHaveReceivedEventTimes(3);
+    await page.keyboard.press("ArrowRight");
+    expect(spy).toHaveReceivedEventTimes(4);
+
+    await page.keyboard.press("Tab");
+    await page.keyboard.press("ArrowLeft");
+    expect(spy).toHaveReceivedEventTimes(5);
+    await page.keyboard.press("ArrowRight");
+    expect(spy).toHaveReceivedEventTimes(6);
+  });
 });
--- a/src/components/calcite-color-picker/calcite-color-picker.scss
+++ b/src/components/calcite-color-picker/calcite-color-picker.scss
@@ -91,7 +91,32 @@ $gap--large: 12px;
 .container {
   display: inline-block;
   border: 1px solid var(--calcite-ui-border-1);
-  background-color: var(--calcite-ui-foreground-1);
+}
+
+.color-field-and-slider-wrap {
+  position: relative;
+  overflow: hidden;
+}
+
+.scope {
+  @apply text--1
+    p-0
+    mb-0
+    mr-0
+    focus-base
+    w-2
+    h-2
+    absolute
+    pointer-events-none;
+
+  margin-top: -0.25rem;
+  margin-left: -0.25rem;
+  outline-offset: 10px;
+
+  &:focus {
+    @apply focus-outset;
+    outline-offset: 12px;
+  }
 }
 
 .color-field-and-slider {
--- a/src/components/calcite-color-picker/calcite-color-picker.scss
+++ b/src/components/calcite-color-picker/calcite-color-picker.scss
@@ -96,7 +96,6 @@ $gap--large: 12px;
 
 .color-field-and-slider-wrap {
   position: relative;
-  overflow: hidden;
 }
 
 .scope {
--- a/src/components/calcite-color-picker/calcite-color-picker.scss
+++ b/src/components/calcite-color-picker/calcite-color-picker.scss
@@ -89,6 +89,7 @@ $gap--large: 12px;
 }
 
 .container {
+  @apply bg-foreground-1;
   display: inline-block;
   border: 1px solid var(--calcite-ui-border-1);
 }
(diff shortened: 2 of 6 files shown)


src/components/calcite-color-picker/calcite-color-picker.scss
```diff
@@ -91,7 +91,32 @@ $gap--large: 12px;
...
+    mr-0
+    focus-base
+    w-2
+    h-2
+    absolute
+    pointer-events-none;
+
+  margin-top: -0.25rem;
+  margin-left: -0.25rem;
+  outline-offset: 10px;
```
paulcpederson, line 114: large outline here allows us to keep the element small. When the element overflowed it's container, on focus it was forcing a repaint of the canvas and causing a gap below the canvas element. When the size is accomplished outline there are no issues. Actual size isn't important as this element has no mouse events and is only keyboard accessible.

src/components/calcite-color-picker/calcite-color-picker.tsx
```diff
@@ -995,6 +1070,11 @@ export class CalciteColorPicker {
     const x = hsvColor.saturationv() / (HSV_LIMITS.s / width);
     const y = height - hsvColor.value() / (HSV_LIMITS.v / height);
 
+    window.requestAnimationFrame(() => {
```
paulcpederson, line 1073: Setting these state values via animation frame allows the render to complete, avoiding repaint bugs as this code is run as part of render.

src/components/calcite-color-picker/calcite-color-picker.tsx
```diff
@@ -295,6 +305,40 @@ export class CalciteColorPicker {
     this.updateChannelsFromColor(this.color);
   };
 
+  private handleScopeKeydown = (event: KeyboardEvent): void => {
```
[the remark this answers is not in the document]
jcfranco: Nitpick: For consistency, `ColorFieldScope` vs `Scope`

src/components/calcite-color-picker/calcite-color-picker.tsx
```diff
@@ -462,22 +506,53 @@ export class CalciteColorPicker {
...
+              [CSS.colorFieldAndSliderInteractive]: colorFieldAndSliderInteractive
+            }}
+            onMouseEnter={this.handleColorFieldAndSliderMouseEnterOrMove}
+            onMouseLeave={this.handleColorFieldAndSliderMouseLeave}
+            onMouseMove={this.handleColorFieldAndSliderMouseEnterOrMove}
+            ref={this.initColorFieldAndSlider}
+          />
+          <div
+            aria-label={vertical ? this.intlValue : this.intlSaturation}
+            aria-valuemax="100"
```
jcfranco, line 540: Use `HSV_LIMITS` const to set min/max.

src/components/calcite-color-picker/calcite-color-picker.tsx
```diff
@@ -462,22 +506,53 @@ export class CalciteColorPicker {
...
+            aria-valuenow={color?.hsv().array()[vertical ? 2 : 1] || "0"}
+            class={CSS.scope}
+            onKeyDown={this.handleScopeKeydown}
+            role="slider"
+            style={{ top: `${scopeTop}px`, left: `${scopeLeft}px` }}
+            tabindex="0"
+          />
+          <div
+            aria-label={this.intlHue}
+            aria-valuemax="360"
```
jcfranco, line 551: Use `HSV_LIMITS` const to set min/max.

src/components/calcite-color-picker/calcite-color-picker.tsx
```diff
@@ -462,22 +506,53 @@ export class CalciteColorPicker {
...
+            aria-valuemin="0"
+            aria-valuenow={color?.hue()?.toFixed(0) || "0"}
```
jcfranco, line 553: For the no-color case, I think we need to remove `aria-valuenow` and possibly use `aria-valuetext` to provide a description:

From https://www.w3.org/TR/wai-aria/#aria-valuenow:

> If the current value is not known (for example, an indeterminate progress bar), the author SHOULD NOT set the aria-valuenow attribute. If the aria-valuenow attribute is absent, no information is implied about the current value.

> When the rendered value cannot be accurately represented as a number, authors SHOULD use the aria-valuetext attribute in conjunction with aria-valuenow to provide a user-friendly representation of the range's current value.

jcfranco, line 553: I think you can use `color?.round().hue()` here.

bstifle (approved):
Very clever, Paul. Ship it

src/components/calcite-color-picker/calcite-color-picker.tsx
```diff
@@ -462,22 +506,53 @@ export class CalciteColorPicker {
...
+            onMouseEnter={this.handleColorFieldAndSliderMouseEnterOrMove}
+            onMouseLeave={this.handleColorFieldAndSliderMouseLeave}
+            onMouseMove={this.handleColorFieldAndSliderMouseEnterOrMove}
+            ref={this.initColorFieldAndSlider}
+          />
+          <div
+            aria-label={vertical ? this.intlValue : this.intlSaturation}
+            aria-valuemax="100"
+            aria-valuemin="0"
+            aria-valuenow={color?.hsv().array()[vertical ? 2 : 1] || "0"}
```
jcfranco, line 542: This would skip the array creation: `!color ? "0" : vertical ? color.saturationv() : color.value()`?

src/components/calcite-color-picker/calcite-color-picker.tsx
```diff
@@ -462,22 +506,53 @@ export class CalciteColorPicker {
...
+              [CSS.colorFieldAndSlider]: true,
+              [CSS.colorFieldAndSliderInteractive]: colorFieldAndSliderInteractive
+            }}
+            onMouseEnter={this.handleColorFieldAndSliderMouseEnterOrMove}
+            onMouseLeave={this.handleColorFieldAndSliderMouseLeave}
+            onMouseMove={this.handleColorFieldAndSliderMouseEnterOrMove}
+            ref={this.initColorFieldAndSlider}
+          />
+          <div
+            aria-label={vertical ? this.intlValue : this.intlSaturation}
```
jcfranco, line 539: Should this slider's label be for value **and** saturation? I'm wondering what the best approach would be here since the screenreader would give no context about the value type changing based on direction.

src/components/calcite-color-picker/calcite-color-picker.e2e.ts
```diff
@@ -1081,4 +1081,29 @@ describe("calcite-color-picker", () => {
...
+    const page = await newE2EPage({
+      html: `<calcite-color-picker value="#beefee"></calcite-color-picker>`
+    });
+
+    const picker = await page.find("calcite-color-picker");
+    const spy = await picker.spyOnEvent("calciteColorPickerChange");
+
+    await page.keyboard.press("Tab");
+    await page.keyboard.press("ArrowUp");
+    expect(spy).toHaveReceivedEventTimes(1);
```
jcfranco, line 1095: This test should also assert the color to make sure the arrow keys (and shift modifier) update values in the expected direction.

src/components/calcite-color-picker/calcite-color-picker.tsx
```diff
@@ -275,6 +275,16 @@ export class CalciteColorPicker {
 
   @State() savedColors: string[] = [];
 
+  @State() scopeTop: number;
```
jcfranco, line 278: Nitpick: For consistency, `colorFieldScope` vs `scope`

src/components/calcite-color-picker/calcite-color-picker.tsx
```diff
@@ -275,6 +275,16 @@ export class CalciteColorPicker {
...
+
+  @State() scopeLeft: number;
+
+  @State() scopeOrientation: "vertical" | "horizontal";
+
+  @State() hueScopeLeft: number;
```
jcfranco, line 284: Nitpick: left and top props seem to be updated and used at the same time. Maybe these could be consolidated.

src/components/calcite-color-picker/calcite-color-picker.tsx
```diff
@@ -1044,6 +1124,11 @@ export class CalciteColorPicker {
     const x = hsvColor.hue() / (360 / width);
     const y = height / 2 + colorFieldHeight;
 
+    window.requestAnimationFrame(() => {
```
jcfranco, line 1132: You can omit `window`.

src/components/calcite-color-picker/calcite-color-picker.tsx
```diff
@@ -323,6 +323,8 @@ export class CalciteColorPicker {
   };
 
   private handleHueScopeKeydown = (event: KeyboardEvent): void => {
+    const { shiftKey } = event;
```
jcfranco, line 326: Nitpick: you can also move `key` to this destructuring line.

jcfranco (changes requested):
@paulcpederson This is looking good. Thanks for helping out with this one! 

I had a few comments and nitpicks. There's some stuff that can be broken out into follow-up issues. Let me know, which ones I can help out with.

src/components/calcite-color-picker/calcite-color-picker.tsx
```diff
@@ -295,6 +305,40 @@ export class CalciteColorPicker {
...
+  private handleScopeKeydown = (event: KeyboardEvent): void => {
+    const key = getKey(event.key);
+    const cardinals = {
+      ArrowUp: { x: 0, y: -10 },
+      ArrowRight: { x: 10, y: 0 },
+      ArrowDown: { x: 0, y: 10 },
+      ArrowLeft: { x: -10, y: 0 }
+    };
+
+    if (this.color && Object.keys(cardinals).indexOf(key) > -1) {
```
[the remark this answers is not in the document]
paulcpederson: is that polyfilled for ie11?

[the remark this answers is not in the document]
paulcpederson: @jcfranco how do I recreate a totally empty state via the demo page?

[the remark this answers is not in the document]
jcfranco: I believe so: https://github.com/ionic-team/stencil/commit/b1071cb2c62fdbae64c6100a27f31b9542425c18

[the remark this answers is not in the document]
jcfranco:
```html
<calcite-color-picker allow-empty value=""></calcite-color-picker>
```

src/components/calcite-color-picker/calcite-color-picker.tsx
paulcpederson: We are in uncharted waters here. The spec doesn't have a two-dimensional slider. Most of the responses I found on forums basically said "just ignore this for screen readers", which I can't do because the whole point is that it's in the tab order, so if you can tab to it you have to allow screen reader access. 

Probably creating a component that controls two independent sliders would be better but man I have no idea how you'd even do that and there are 0 examples of that kind of thing...

src/components/calcite-color-picker/calcite-color-picker.tsx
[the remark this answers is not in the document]
paulcpederson: ty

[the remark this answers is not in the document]
paulcpederson: actually I still get a blue even with this prop...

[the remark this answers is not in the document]
jcfranco: > I still get a blue

For the `value`/`color` props? The color field will still render the default blue in this case.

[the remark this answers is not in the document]
paulcpederson: yeah like a hex is selected  in the hex input, etc

src/components/calcite-color-picker/calcite-color-picker.tsx
paulcpederson: I was yelled at by the a11y tools when I didn't provide a now value. But tbh, there is a color field, and it has a scope, even when it's empty. So to me, value now should reflect the position of the scope.

src/components/calcite-color-picker/calcite-color-picker.tsx
paulcpederson: I think we can make the accessibility of the labeling better in a future pr.

src/components/calcite-color-picker/calcite-color-picker.tsx
jcfranco: This is not correct though. Having no hue is not the same as a hue of 0. We can pick this up in that follow-up a11y issue you proposed.

paulcpederson: Yeah, you could remove the fallback to `"0"`. Again, I have not been able to get that condition to ever happen, though.

src/components/calcite-color-picker/calcite-color-picker.tsx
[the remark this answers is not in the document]
jcfranco: Could you share the steps to get into this state? The values are initially null: https://codepen.io/jcfranco-the-scripter/pen/LYxWWWX?editors=1000. The same holds when the color is cleared.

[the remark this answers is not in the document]
paulcpederson: ok, I see. So you have to set `allow-empty` and explicitly set `value` to an empty string. I assumed if I used `allow-empty` it should be empty when I don't provide a value...

src/components/calcite-color-picker/calcite-color-picker.e2e.ts
paulcpederson: working on this right now, fyi

src/components/calcite-color-picker/calcite-color-picker.tsx
paulcpederson: So you can't omit `now`. Our accessibility check inside the test suite fails you if it's omitted. Is there a way to tell the a11y check to ignore certain rules in certain checks?

paulcpederson: looks like the axe core folks recommend using "0" if it's unknown... 

https://github.com/dequelabs/axe-core/issues/2639#issuecomment-728210415

jcfranco: Thanks for digging into this. It's really useful info. Seems like that and `aria-valuetext` is the winning combo. 💥 👊🏼 💥

paulcpederson:
sigh, that test doesn't fail locally, so I don't know why it fails on travis...

PR #1894, opened 2021-04-01
fix(action-bar): Only allow one overflow menu open at a time.

driskull:
**Related Issue:** #1819

## Summary

fix(action-bar): Only allow one overflow menu open at a time.


driskull:
Passing tests so merging. If updates are needed let me know.

PR #1711, opened 2021-03-09
refactor(radio-group-item): tailwindify

julio8a:
Supports #1708 #1500

PR #1903, opened 2021-04-02
fix(radio-group): ensure group's outline is included in new total heights

caripizza:
**Related Issue:** n/a - relates to Tailwind refactor from pr #1711, issue #1500

cc @julio8a @bstifle 

## Summary
Ensures the gray outline around `calcite-radio-group` stays inside the new total heights for each scale. Also freshens up its demo html a bit.


julio8a (approved):
Awesome! thx @caripizza

Issue #1896, opened 2021-04-01
calcite-action-menu: when opening menu set focus on first action


AdelheidF:
### Summary
If the popover opens the focus is not inside the popover. 

### Reproduction Steps

1.

AdelheidF:
Tested locally with next.142.

It now sets the focus on the first action in the menu after opening the menu, but when using the up and down arrows the focus doesn't move, unlike calcite-dropdown. Also, hitting space doesn't execute the action. As a keyboard only user it's impossible to select the focused action or move down to another one.

driskull:
Ah, forgot this was using activedescendent. This will need some more work then 👍

AdelheidF:
Using next.146...

When I open the menu I now can move up and down using the arrow keys and hitting enter/space selects the grey action. But when I open the menu I no longer see the typical blue focus box.

![image](https://user-images.githubusercontent.com/44481292/113637844-6ad3f280-962a-11eb-961b-bd78d07f1441.png)

driskull:
@AdelheidF yes that's by design. It's using `activedescendant` so that the button doesn't lose focus.

https://www.w3.org/TR/wai-aria-practices-1.1/examples/menu-button/menu-button-actions-active-descendant.html

PR #1904, opened 2021-04-02
feat(action-menu): Adds setFocus method and focus on first action when open or menu button when closed.

driskull:
**Related Issue:** #1896

## Summary

feat(action-menu): Adds setFocus method and focus on first action when open or menu button when closed.

### DIFF
--- a/src/components/calcite-action-menu/calcite-action-menu.tsx
+++ b/src/components/calcite-action-menu/calcite-action-menu.tsx
@@ -7,6 +7,7 @@ import {
   EventEmitter,
   Prop,
   Watch,
+  Method,
   State
 } from "@stencil/core";
 import { CSS, ICONS, SLOTS } from "./resources";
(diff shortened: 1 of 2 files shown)


src/components/calcite-action-menu/calcite-action-menu.tsx
```diff
@@ -142,6 +143,17 @@ export class CalciteActionMenu {
...
 
+  // --------------------------------------------------------------------------
+  //
+  //  Methods
+  //
+  // --------------------------------------------------------------------------
+
+  @Method()
+  async setFocus(): Promise<void> {
+    focusElement(this.open ? this.actionElements[0] || this.menuEl : this.menuButtonEl);
```
jcfranco, line 154: Would there be cases where we want a `focusId` for the menu button?

src/components/calcite-action-menu/calcite-action-menu.tsx
```diff
@@ -274,11 +285,14 @@ export class CalciteActionMenu {
...
-      (this.el
-        .querySelector("slot")
-        ?.assignedElements({ flatten: true })
-        .filter((el) => el.tagName === "CALCITE-ACTION") as HTMLCalciteActionElement[]) || [];
+    const { el } = this;
+
+    const assignedElements = el
+      .querySelector("slot")
+      ?.assignedElements({ flatten: true })
+      .filter((element) => element?.tagName === "CALCITE-ACTION") as HTMLCalciteActionElement[];
```
jcfranco, line 293: `element?.tagName` shouldn't need optional chaining.

jcfranco (approved):
👍🏼

src/components/calcite-action-menu/calcite-action-menu.tsx
driskull: I don't think so. It depends on if it's open or not so we can go with that for now.

Issue #1914, opened 2021-04-05
Bug: Panel and ShellPanel need flex grow, shrink, and basis defined


asangma:
### Summary
The height is not calculated correctly in Safari due to a lack of flex definitions.

### Reproduction Steps

1.

asangma:
Flow also needs to be updated.

asangma:
@benelan I know this is old, but I created the issue and made the fix. I'm OK to close it.

PR #1915, opened 2021-04-05
fix(panel, shell-panel): flex grow, shrink, basis definitions 

asangma:
**Related Issue:** (#1914)

## Summary
Adds explicit flex grow, shrink, and basis definitions to make Safari play nice.

cc @AdelheidF 
Also added to Panel at the :host level to ease integration.


driskull (approved):
Looks good. Do we need a test?

asangma:
@driskull I don't think so since it's not property dependent and mostly just makes an explicit CSS style for what Chrome does by default.

asangma:
Unless you mean a Screener test?

driskull (approved):
👍

PR #1918, opened 2021-04-06
fix(panel, shell-panel, flow): fixes height and scrolling issues with Safari

asangma:
… Safari (#1914)

**Related Issue:** #1914 

## Summary
Adds additional styles to play nice with Safari.

Custom elements, i.e. other components inside Panel and Flow will still need these styles applied to them:

```
display: flex;
flex: 1 1 auto;
overflow: hidden;
```

cc @AdelheidF @kevindoshier @dhrumil83 @subgan82
If you know Ganesh and Dhrumil's handles, please mention them here.

### DIFF
--- a/src/components/calcite-flow/calcite-flow.scss
+++ b/src/components/calcite-flow/calcite-flow.scss
@@ -1,15 +1,16 @@
 :host {
   @extend %component-host;
   @apply flex 
+    flex-auto
     items-stretch 
     bg-transparent 
-    w-full 
-    h-full 
+    w-full
     overflow-hidden 
     relative;
 
   .frame {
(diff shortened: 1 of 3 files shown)


asangma:
@driskull I'm going to try and do a test for the heights.
I'll try adding them to this PR if I can get them together soon.

driskull:
Yeah a test sounds good so we can catch these. We should be able to check the scrolling area to make sure it's scrollHeight is the same as it's clientHeight.

At least a screenshot test would be nice :)

driskull (approved):
👍

asangma:
Issue to add test created #1919

Issue #1917, opened 2021-04-05
A11y: The aria-expanded in calcite-block should be on the button not the article element 


kellyhutchins:
### Summary
This issue was reported during an accessibility audit of the configurable apps. Here's how they reported the issue: 

The info panel's points of interest expandable/collapsible controls contain invalid ARIA attributes and are not marked up correctly:
-Invalid: the aria-expanded attribute is on the article element
-The button is missing aria-expanded
The aria-expanded attribute should be set on the <button> elements.

jcfranco:
Installed.

kellyhutchins:
Tested locally and looks great. Thanks for the quick fix @jcfranco

PR #1922, opened 2021-04-07
fix(block): place aria-expanded on block toggle button

jcfranco:
**Related Issue:** #1917

## Summary

Moves aria attribute to the proper toggling element.


driskull (approved):
👍

PR #1923, opened 2021-04-07
fix(date-picker-day): restore out-of-range day styling

paulcpederson:
## Summary

Fixing the broken out of range day styling. Also updated the date-picker story to include out of range days so that screener will verify future updates.


driskull (approved):
👍

driskull:
Can we merge?

paulcpederson:
Yeah let's get it in as it fixes something that's currently broken

Issue #1810, opened 2021-03-19
Bug: calcite-input (number) has no locale support


AdelheidF:
### Summary
In French e.g. the decimal separator needs to be a comma and date separator needs to be a dot. I can't find a way to make this happen. 

![image](https://user-images.githubusercontent.com/44481292/111825089-49da7600-88a4-11eb-934b-48ff4044925e.png)

<img src=https://user-images.githubusercontent.com/44481292/111824707-c751b680-88a3-11eb-9837-8f9e5ae44a8f.png width=200/>

Outside of those input boxes...
<img src=https://user-images.githubusercontent.com/44481292/111824582-a2f5da00-88a3-11eb-89b7-f52bbe24ab2f.png width=200/>

Also, where does that (English) text come from when I hover over the input? I'm not adding a step property and what it says is not true.

![image](https://user-images.githubusercontent.com/44481292/111826211-a5593380-88a5-11eb-89ff-2a7c2536651b.png)
![image](https://user-images.githubusercontent.com/44481292/111825502-c2d9cd80-88a4-11eb-97bf-97719e244b47.png)

cc @jcfranco @macandcheese 

### Reproduction Steps

1.

ethanbdev:
You can set step to "any" to get rid of that hover tip @AdelheidF , if it is not set it defaults to 1.

AdelheidF:
> You can set step to "any" 

Thanks, that works.

AdelheidF:
What is the plan for a French UI that uses calcite-input with step:1 when the user enters a decimal value? I assume the message won't be translated.

AdelheidF:
@paulcpederson @macandcheese is this doable for beta.52?

paulcpederson:
@AdelheidF date picker can already do this. For example, try `fr-CH` as lang. Not sure why the `fr` locale doesn't use the dot separator, but would be easy to just remove fr and replace it with `fr-CH`. Maybe @raje9276 can comment on the distinction between these two regions.

AdelheidF:
fr-CH is French spoken in Switzerland. 

de or es is not any better:

<img src=https://user-images.githubusercontent.com/44481292/112067995-40f8d700-8b26-11eb-811d-56ea982f33a0.png width=250/><img src=https://user-images.githubusercontent.com/44481292/112068063-6554b380-8b26-11eb-90a8-a10c6b15cca8.png width=250/>


> would be easy to just remove fr and replace it with fr-CH

Who would do this?

[...]

PR #1924, opened 2021-04-07
fix(calcite-input): number locale support

eriklharper:
**Related Issue:** #1810

image


AdelheidF:
Could you try ar and he with the page being in RTL?

jcfranco (approved):
noice

This is looking nooice, @eriklharper! Had a few comments and suggestions for maintainability, but the tests and code look solid. All that's left I think is confirming how `ar` + `dir=rtl` behaves.

src/components/calcite-input/calcite-input.e2e.ts
```diff
@@ -651,4 +652,45 @@ describe("calcite-input", () => {
     expect(await getInputValidity()).toBe(true);
     expect(await input.getProperty("value")).toBe("123");
   });
+
+  describe("number locale support", () => {
+    const localesWithIssues = ["ar", "bs", "de-CH", "it-CH", "mk"];
+    locales
+      .filter((locale) => (localesWithIssues.includes(locale) ? false : true))
```
jcfranco, line 659: Simplify to `.filter((locale) => (!localesWithIssues.includes(locale))`

AdelheidF (approved):

src/utils/number.ts
```diff
@@ -0,0 +1 @@
+export const numberKeys = ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9"];
```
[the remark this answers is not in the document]
eriklharper: I'll stick this in the `key.ts` file.

eriklharper:
> Could you try ar and he with the page being in RTL?

image

Should it be `65.432,1` instead?

AdelheidF:
ar looks good, this is correct. What about he, Hebrew.

Get Outlook for Android<https://aka.ms/AAb9ysg>

________________________________

eriklharper:
> ar looks good, this is correct. What about he, Hebrew.

image

AdelheidF:
Is today the happy day when this gets installed? :)

eriklharper:
> Is today the happy day when this gets installed? :)

I've hit just one snag, which is arrowup/arrowdown support is broken because number inputs now have to use a plain text input under the hood to allow the formatting to be typed and displayed.  I just need to refactor input a bit to restore up/down keyboard support, which is already built into the component, just need to extract that functionality into standalone functions that can be called on keypress in addition to the up/down arrow buttons that display in the UI.

zaramatheson:
@eriklharper Just wanted to check the status of this one today? Thanks!

eriklharper:
> @eriklharper Just wanted to check the status of this one today? Thanks!

Hi @zaramatheson !  Good timing, I just pushed up a major fix that was breaking several tests in the color-picker component.  Now that that is resolved, there is just one more small tweak that needs to be made, which @jcfranco is patching up now.  Things are looking good to get this merged today!

src/components/calcite-input/calcite-input.tsx
```diff
@@ -183,27 +240,270 @@ export class CalciteInput {
...
-    this.stepString = this.step?.toString();
+    this.minString = this.min?.toString();
+    this.requestedIcon = setRequestedIcon(INPUT_TYPE_ICONS, this.icon, this.type);
   }
 
   componentDidLoad(): void {
     this.slottedActionEl = this.el.querySelector("[slot=input-action]");
     this.setDisabledAction();
+    if (this.shouldFormatNumberByLocale()) {
+      this.childEl.style.cssText = hiddenInputStyle;
```
jcfranco, line 254: Does this need to happen in another lifecycle hook? `componentDidLoad` only happens once, so this could be problematic if the component is reused.

src/components/calcite-input/calcite-input.tsx
```diff
@@ -183,27 +240,270 @@ export class CalciteInput {
...
+  ): CustomEvent => {
+    return this.calciteInputInput.emit({
+      element: this.childEl,
+      nativeEvent,
+      value
+    });
+  };
+
+  private incrementNumberValue = (nativeEvent: KeyboardEvent | MouseEvent): void => {
+    if (this.type === "number") {
```
jcfranco, line 355: Consolidate the increment/decrement value methods. There's a good chunk of reusable code between them.

src/components/calcite-input/calcite-input.tsx
```diff
@@ -183,27 +240,270 @@ export class CalciteInput {
...
+  }
+
+  //--------------------------------------------------------------------------
+  //
+  //  Private Methods
+  //
+  //--------------------------------------------------------------------------
+
+  private clearInputValue = (event: KeyboardEvent | MouseEvent): void => {
+    const calciteInputInputEvent = this.emitInputFromUserInteraction("", event);
```
jcfranco, line 316: Looks like the emit => defaultPrevented? => setValue logic could also be extracted into a method.

src/components/calcite-input/calcite-input.tsx
```diff
@@ -183,27 +240,270 @@ export class CalciteInput {
...
+
+  /**
+   * @internal
+   */
+  @Event() calciteInputBlur: EventEmitter;
+
+  /**
+   * This event fires when the value of the input changes.
+   */
+  @Event({ eventName: "calciteInputInput", cancelable: true }) calciteInputInput: EventEmitter;
```
jcfranco, line 277: Just to confirm, the `cancelable` is still relevant to this PR after the internal input refactor, right?

jcfranco (approved):
Awesome work, @eriklharper ! Aside from a few comments, this is looking good to merge!

src/components/calcite-input/calcite-input.tsx
eriklharper: Correct!  The cancelability is still relevant even though the color-picker isn't using that custom event directly, its still just relying on the native input event.

src/components/calcite-input/calcite-input.tsx
eriklharper: Unfortunately, `this.childEl` isn't accessible until the first `render` cycle, so this operation can't be performed inside `connectedCallback` or `componentWillLoad` or `componentWillRender` (on the first render cycle) so the only place that made sense to me to do this was in `componentDidLoad`.

src/components/calcite-color-picker/calcite-color-picker.tsx
```diff
@@ -730,6 +730,7 @@ export class CalciteColorPicker {
       aria-label={ariaLabel}
       class={CSS.channel}
       data-channel-index={index}
+      localeFormat={false}
```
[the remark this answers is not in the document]
eriklharper: I needed a way to enable an escape hatch so that color-picker can just resume its normal functionality, so that's what this prop is for.

src/components/calcite-input/calcite-input.tsx
jcfranco: Just an idea (for later), but If `hiddenInputStyle` was an object, we could set it directly on the `style` prop where childEl is rendered. This would eliminate the need to get a reference to the node and set it.

src/components/calcite-input/calcite-input.tsx
eriklharper: Follow up issue?  I'd like to also maybe add some more tests too, but I'd like to get this merged in today.

julio8a:
🎉 Thanks for the hard work on this @eriklharper!

src/components/calcite-input/calcite-input.tsx
jcfranco: 👍🏼

Issue #1927, opened 2021-04-07
Enhancement: Add CSS utility class to ShellPanel to help match flex container heights


asangma:
### Description
Add a class that helps users correctly size their own elements when putting them inside of a container component (e.g. ShellPanel).

### Acceptance Criteria 
* works with ShellPanel
* works with non component tags, e.g. `div` or a custom element.
* works with nested tags, e.g. 
```
<div>
  <some-custom-element>
    <calcite-panel>
      ...
    </calcite-panel>
  </some-custom-element>
</div>
``` 
cc @AdelheidF @kevindoshier @driskull

asangma:
Would someone mind verifying this.
I can provide details.

driskull:
Closing. reopen if still an issue.

PR #1929, opened 2021-04-07
feat(shell-panel): adds utility class for properly setting the height of external elements and adds related demo apps

asangma:
**Related Issue:** #1927

## Summary
Adds a utility class **to ShellPanel** for setting proper heights and scrolling for content that is not an internal component.
This is an opt-in class.

It should help with multiple levels of nested nodes.
See the nesting-testing demos.

cc @AdelheidF @kevindoshier

### DIFF
--- a/src/assets/styles/global.scss
+++ b/src/assets/styles/global.scss
@@ -45,13 +45,6 @@
   overflow: hidden;
 }
 
-.calcite-match-height {
-  @apply flex
-  flex-auto
-  flex-col
-  overflow-hidden;
-}
-
 [calcite-hydrated-hidden] {
   visibility: hidden;
   pointer-events: none;
--- a/src/assets/styles/global.scss
+++ b/src/assets/styles/global.scss
@@ -45,6 +45,13 @@
   overflow: hidden;
 }
 
+.calcite-match-height {
+  @apply flex
+  flex-auto
+  flex-col
+  overflow-hidden;
+}
+
 [calcite-hydrated-hidden] {
   visibility: hidden;
   pointer-events: none;
--- a/src/components/calcite-shell-panel/calcite-shell-panel.scss
+++ b/src/components/calcite-shell-panel/calcite-shell-panel.scss
@@ -14,6 +14,13 @@
   max-width: unset;
 }
 
+::slotted(.calcite-match-height) {
+  @apply flex
+  flex-auto
+  flex-col
+  overflow-hidden;
+}
+
 .content {
   @apply items-stretch 
     self-stretch 
--- a/src/components/calcite-shell-panel/calcite-shell-panel.scss
+++ b/src/components/calcite-shell-panel/calcite-shell-panel.scss
@@ -17,7 +17,6 @@
 ::slotted(.calcite-match-height) {
   @apply flex
   flex-auto
-  flex-col
   overflow-hidden;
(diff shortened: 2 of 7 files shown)


src/assets/styles/global.scss
```diff
@@ -45,6 +45,13 @@
   overflow: hidden;
 }
 
+.calcite-match-height {
```
driskull, line 48: Just wondering, Is there a way in tailwind to define a utility class? Just wondering if we should be using tailwind here at all since I don't really know it well :)

driskull: Also, how are we going to document these utility classes? Should we have a `.mdx` page for storybook we can use to document them. At least a `markdown` file somewhere?

driskull (commented):
Comments :)

src/assets/styles/global.scss
asangma: re: doc
I think that's a good idea.

re: tailwind
I was thinking that, but so far it seems like the Tailwind utility classes aren't accessible if you're just using the components. But mayhapz I'm missing something.

paulcpederson: IMO `global.scss` should not store an entire utility class set. It's really more about global things that all components need, ie. theme, hydration, etc.

asangma: Where could I put this so that it can be used in the way demonstrated in the [demo app](https://github.com/Esri/calcite-components/pull/1929/files#diff-cad01ac18aa67c8ff5e4a590929fee4211f25f2476e955c46cb2c6aa490195c8R117-R120)?
Assuming that there's not something weird about it.

asangma: Also, yes...the demo app seems kinna weird, but it's a reality.

driskull: Maybe we should discuss all the options at a meeting?

- component to do this?
- utility class? 
- Just have them apply the needed styles?

asangma:
@paulcpederson @driskull 
I tried the `::slotted(.calcite-match-height)` approach. 
This works, but only for one-level deep.

```
<calcite-shell-panel>
  <something class="calcite-match-height">
    <calcite-panel>
      ...
    </calcite-panel>
  </something>
</calcite-shell-panel>
```

It seems that if there is any nesting, it no longer applies.
```
<calcite-shell-panel>
  <something class="calcite-match-height">
    <div class="calcite-match-height">
      <calcite-panel>
        ...
      </calcite-panel>
    </div>
  </something>
</calcite-shell-panel>
```

In this case `div.calcite-match-height` doesn't receive the styles.
This may actually be fine since the previous example seems like a more reasonable node structure.

asangma:
Aight! Updated and ready for your re-review.

driskull (approved):
👍

AdelheidF:
Will this wait for after beta.53?

asangma:
@AdelheidF We can wait, but it should have no affect since we're adding just styles to a new selector.

AdelheidF:
> We can wait, but it should have no affect since we're just adding styles to a new selector.

OK, thanks. I prefer waiting though.

asangma:
@AdelheidF Copy that.

asangma:
Converting to draft until after beta.53.

Issue #1919, opened 2021-04-06
Enhancement: ShellPanel - add tests for Panel and Flow heights and scrolling


asangma:
### Description
With updates to how height is handled in ShellPanel, Panel, and Flow, it will be helpful to add tests to ensure that Panels and Flows (with Panels) have heights that equal the height of ShellPanel's `content__body` node.

It would also be helpful to ensure proper scrolling behavior.

Related PR #1918

PR #1935, opened 2021-04-08
test(shell-panel, panel): adds test to ensure panel height matches shell-panel content node

asangma:
**Related Issue:**  (#1919)

## Summary
Adds test to ensure Panel height matches the content node of ShellPanel.


driskull (approved):
👍 💥

PR #1940, opened 2021-04-08
fix(combobox): open combobox in the correct direction

paulcpederson:
## Summary

We were running into issues where at first, a combobox would _always_ open down. Any scrolling or recalculation would make it move to the top (correct), but initially it was wrong. 

Digging in, this is because we were setting the height to `0` as part of https://github.com/Esri/calcite-components/pull/1577

So this pr, we first show the items (giving the menu height), then reposition before showing the menu itself. Also removed the `dir` prop on the host component as I think we're getting rid of those.


driskull (approved):
👍 nice!!

Issue #1809, opened 2021-03-19
Bug: Tree item does not show if it starts expanded


driskull:
### Summary

Bug: Tree item does not show if it starts expanded

### Actual Behavior

https://codepen.io/driskull/pen/KKNOKpN?editors=1000

![image](https://user-images.githubusercontent.com/1231455/111826014-6b882d00-88a5-11eb-98d2-343a3747c0ab.png)

### Expected Behavior

The tree item children should be visible.

![image](https://user-images.githubusercontent.com/1231455/111826050-75119500-88a5-11eb-9343-413169f9ba45.png)

### Reproduction Steps

1. https://codepen.io/driskull/pen/KKNOKpN?editors=1000

### Relevant Info 

N/A

paulcpederson:
@driskull sorry this was fixed a while back I totally spaced reassigning it.

driskull:
👍 awesome

PR #1943, opened 2021-04-09
fix(tree-item): allow trree to begin expanded #1809

paulcpederson:
**Related Issue:** #1809

## Summary

Tree items should be able to start expanded.


driskull (approved):
👍

Issue #1930, opened 2021-04-08
Bug: `calcite-action-bar` expand tooltip not opening on hover when used inside web component


ethanbdev:
Using the new slot, the tooltip does not open on hover. The action bar is being used inside a web component, that could be related to the issue. I cannot reproduce in a code pen. Sent @driskull a url where it is reproducible.
![98781FEF-2192-40EB-AC1F-4999C16B4EE8](https://user-images.githubusercontent.com/52869490/114039413-8fe98080-9837-11eb-90f5-6af01afe8923.GIF)

driskull:
Yeah, it's because it's put into another shadowDom element and it's trying to find the tooltip with `docoment.getElementById()`. 

We will have to introduce some logic to query the shadow dom instead.

ethanbdev:
This looks good, thanks @driskull 👍

PR #1956, opened 2021-04-13
fix(tooltip, tooltip-manager, popover, popover-manager): Work correctly within shadowRoot elements.

driskull:
**Related Issue:** #1930

## Summary

fix(tooltip, tooltip-manager, popover, popover-manager): Work correctly within shadowRoot elements.


jcfranco (approved):
Noice!

Issue #1710, opened 2021-03-09
Enhancement: Auto-adjust headingLevel of child elements


jcfranco:
### Description

https://github.com/Esri/calcite-components/pull/1639/ added `headingLevel` to the block, panel, pick-list-group and tip-manager components. The level prop needs to be set on the parent and child elements (except for tip & tip-manager). As an enhancement, the remaining components should follow this pattern to make it easier for users.

### Acceptance Criteria 

Root elements with a `headingLevel` prop should automatically adjust for child elements (similar to tip-manager & tip).

#### Which Component

- `calcite-block`
- `calcite-pick-list`
- `calcite-panel`

driskull:
@jcfranco how does this affect block or panel? I think it's just the `pick-list` and `value-list` that need to modify the `pick-list-group`.

jcfranco:
> how does this affect block or panel? I think it's just the pick-list and value-list that need to modify the pick-list-group.

Sorry for the delay. I can't remember why I added block or panel. For panel, maybe because I was thinking about having it adjust the heading level of any child content, but that may be too complex/intrusive.

jcfranco:
Verified locally!

PR #1960, opened 2021-04-13
fix(block, date, panel, pick-list, pick-list-group, tip, tip-manager)!: Remove default 'headingLevel' value.  #1710

driskull:
**Related Issue:** #1710

## Summary

fix(block, date, panel, pick-list, pick-list-group, tip, tip-manager)!: Remove default 'headingLevel' value. 

Removes the default headingLevel value and child elements will calculate the value relative to their parent components.

### DIFF
--- a/src/components/calcite-pick-list-group/calcite-pick-list-group.tsx
+++ b/src/components/calcite-pick-list-group/calcite-pick-list-group.tsx
@@ -3,7 +3,7 @@ import { CSS, SLOTS } from "./resources";
 import { CSS_UTILITY } from "../../utils/resources";
 import { HEADING_LEVEL } from "./resources";
 import { getElementDir, getSlotted } from "../../utils/dom";
-import { HeadingLevel, CalciteHeading, ConstrainHeadingLevel } from "../functional/CalciteHeading";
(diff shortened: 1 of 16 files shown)


src/components/calcite-pick-list-group/calcite-pick-list-group.tsx
```diff
@@ -56,11 +56,14 @@ export class CalcitePickListGroup {
     };
 
     const title = groupTitle;
+    const parentLevel = el.closest("calcite-pick-list")?.headingLevel;
+    const relativeLevel = parentLevel ? ConstrainHeadingLevel(parentLevel + 1) : null;
```
jcfranco, line 60: Unrelated: Since this is a utility function and not a functional component, should this be `constrainHeadingLevel`?

jcfranco (approved):
Nice!

I saw you have some additional info in the commit message. Before merging, could you preface it in the commit message body w/ `BREAKING CHANGE:` so it gets pulled into the changelog?

```
BREAKING CHANGE: Removes the default headingLevel value and child elements will calculate the value relative to their parent components.
```

Issue #1145, opened 2020-10-20
Refactor: API/nomenclature review


jcfranco:
Initial review notes: https://github.com/Esri/calcite-components/wiki/calcite-components-API-review

## Action items

**Note**: PRs for each item must include an update to our conventions doc.

~* [ ] ensure consistent boolean names (@jcfranco)~
  ~* prefer adjectives (e.g., collapsible, moveable, etc), if an appropriate one does not exist, `-disabled`/`-enabled` suffixes can be used to describe the prop.~
~* avoid initializing booleans to true~
☝🏼  extracted to https://github.com/Esri/calcite-components/issues/2115
* [x] wherever applicable, update `value` prop types to accept the most flexible type allowed. (https://github.com/Esri/calcite-components/pull/1961
  * widening calcite-input's value was extracted to #758 
* [x] make sure that all public events have doc (@jcfranco) https://github.com/Esri/calcite-components/pull/1517
* [x] mark registration-related events (for parent-child components) as internal to be excluded from the doc (@jcfranco) https://github.com/Esri/calcite-components/pull/1517
* [x] for any component with focusable content, the `setFocus` method must be implemented (@jcfranco) https://github.com/Esri/calcite-components/pull/1522
* [x] remove outdated sections from readmes and move any examples to a usage folder (@jcfranco) https://github.com/Esri/calcite-components/pull/1521
* [x] prefer `calciteComponentChange` name for events that notify about component updates (@jcfranco) https://github.com/Esri/calcite-components/pull/1519
* [x] mark reflected props used for styling as internal to be excluded from the doc (@jcfranco) https://github.com/Esri/calcite-components/pull/1523

julio8a:
Hey guys, checking in on this one since it's a required item for the 1.0

paulcpederson:
Below is a summary of all the components and their props. This may help us spot inconsistencies.

---

#### calcite-accordion
appearance, iconPosition, iconType, scale, selectionMode, theme

#### calcite-accordion-item
active, icon, itemSubtitle, itemTitle

#### calcite-action
active, appearance, compact, disabled, icon, indicator, intlLoading, label, loading, scale, text, textEnabled, theme

#### calcite-action-bar
expand, expanded, intlCollapse, intlExpand, position, theme, tooltipExpand

#### calcite-action-pad
expand, expanded, intlCollapse, intlExpand, layout, position, theme, tooltipExpand

#### calcite-alert
[...]

PR #1961, opened 2021-04-14
feat(checkbox, chip, combobox, pick-list, radio-button, switch, tile-select, value-list): widen value prop type

jcfranco:
**Related Issue:** #1145 

## Summary

This PR allows the `value` prop of many components to accept any type. This is greatly helpful w/ VDOM use cases where a value object may be passed directly instead of having to map them to element attributes.

### Notes

* Setting the `value` **prop** programmatically no longer reflects. I don't _think_ this was commonly used, but it is easy to replace by using another attribute for querying/styling.
* For any components using an internal input for forms, the value will be cast as a string.
* Did not update the types for date-related components as they have a separate `valueAsDate` props.
* Holding off on updating `calcite-input` (value as string and number) it has some incoming changes to it.


paulcpederson (approved):

bstifle:
@jcfranco is this valid?

jcfranco:
@bstifle Yes, still valid. Reopening.

driskull (approved):
👍

PR #1962, opened 2021-04-14
fix(color-picker): fix color field thumb nudging when saturation change does not change the color's RGB value

jcfranco:
**Related Issue:** N/A

## Summary

This fixes an issue saturation/value cannot be nudged if it would produce the same RGB value. This is reproducible by setting the color picker to 000 and then trying to modify saturation/value with the keyboard.


paulcpederson (approved):

paulcpederson:
nice, thanks @jcfranco

PR #1965, opened 2021-04-14
fix: fix transform value typo from form style util

jcfranco:
**Related Issue:** N/A

## Summary

🔠 ✨ 🔨

### DIFF
--- a/src/utils/form.ts
+++ b/src/utils/form.ts
@@ -8,7 +8,7 @@ export const hiddenInputStyle = `
   position: absolute !important;
   right: 0 !important;
   top: 0 !important;
-  transform: noone !important;
+  transform: none !important;
   -webkit-appearance: none !important;
   z-index: -1 !important;
 `;


src/utils/form.ts
```diff
@@ -8,7 +8,7 @@ export const hiddenInputStyle = `
   position: absolute !important;
   right: 0 !important;
   top: 0 !important;
-  transform: noone !important;
```
driskull, line 11: noooooooooooone :)

driskull (approved):
👍 

Any idea how we can prevent this in the future? Should we have some way to lint this?

paulcpederson (approved):

src/utils/form.ts
jcfranco: nooo

jcfranco:
> Any idea how we can prevent this in the future? Should we have some way to lint this?

Maybe we could do something w/ stylelint, but I'm not entirely sure. Did a quick search but couldn't find any tooling for linting embedded CSS. 😞

eriklharper:
Thanks for catching that, I blame my keyboard!  The flatter keys on these MacBooks are always getting stuck and double-tapping for me.

PR #1969, opened 2021-04-14
refactor(calcite-input): making locale-format false by default

eriklharper:
**Related Issue:** #1810

This changes `localeFormat` prop to be off by default.


jcfranco (approved):
💥

PR #1973, opened 2021-04-14
fix(button): Remove setting 'dir' attribute in light DOM elements. #1831

driskull:
fix(button): Remove setting 'dir' attribute in light DOM elements. #1831

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@esri/calcite-components",
-  "version": "1.0.0-next.160",
+  "version": "1.0.0-next.161",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


src/components/calcite-button/calcite-button.scss
```diff
@@ -75,14 +75,14 @@
     margin-right: unset;
   }
 }
-:host([icon-start][dir="rtl"]:not([icon-end])) {
-  .content {
+:host([icon-start]:not([icon-end])) {
+  .calcite--rtl .content {
     margin-left: unset;
```
paulcpederson, line 80: If we are dropping IE support you could just use [logical properties](https://caniuse.com/css-logical-props) here and drop the rtl logic all together.

paulcpederson (approved):

src/components/calcite-button/calcite-button.scss
driskull: Yeah we should do this at some point. This probably is something we could do for all components

PR #1975, opened 2021-04-14
fix(combobox-item-group): Remove setting 'dir' attribute in light DOM elements. #1831

driskull:
fix(combobox-item-group): Remove setting 'dir' attribute in light DOM elements. #1831


src/components/calcite-combobox-item-group/calcite-combobox-item-group.scss
```diff
@@ -29,7 +29,7 @@
   --calcite-combobox-item-indent-end-2: unset;
 }
 
-:host([dir="rtl"]) {
+:host .calcite--rtl {
```
paulcpederson, line 32: I don't think you need `:host` here, right?

jcfranco (approved):
👍🏼

PR #1984, opened 2021-04-15
fix(custom-elements): ensure root components get a util bundle generated

jcfranco:
**Related Issue:** N/A

## Summary

`createBundles.ts` uses `collection-manifest.json` to generate the util bundles for the custom-elements output target (https://github.com/Esri/calcite-components/pull/1499). This would work for most cases except if a bundle in `stencil.config.ts` had component names where sorting them alphabetically moved the 1st item (the Stencil build sorts them when generating the manifest file).

For example, the following would work fine since the root component will remain first after sorting.

```ts
{
  components: [
    "calcite-action",
    "calcite-action-group",
    "calcite-action-menu",
    "calcite-action-bar",
    "calcite-action-pad"
  ]
},
```

By contrast, the following would generate a bundle for `calcite-option` and not `calcite-select` (root) after sorting.

```ts
{
  components: [
    "calcite-select", 
    "calcite-option", 
    "calcite-option-group"
  ]
},
```

### DIFF
--- a/conventions/README.md
+++ b/conventions/README.md
@@ -471,6 +471,8 @@ As a best practice we should follow [Ionic's configuration](https://github.com/i
 
 **Note:** This is highly likely to change as we move closer to our first release and as Stencil improves their documentation around their specific methods and build processes.
 
+Each root component should have a corresponding bundle entry in `stencil.config.ts`. It is important that the root component be listed first (the `util:build-custom-elements-output-utils` NPM script depends on this).
+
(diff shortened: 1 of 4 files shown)


driskull (approved):
Awesome 👏. Thank you for digging into this Franco!

paulcpederson (approved):

paulcpederson:
Nice work, @jcfranco 👍🏻

Issue #1982, opened 2021-04-14
Bug: calcite-input type=number can't update value by changing it internally


AdelheidF:
### Summary
Using next.161.

If a value shows up or it's empty and I then change the value through another way, e.g. by picking a unique value or clicking on the histogram, the value that shows or doesn't show in the input doesn't change. This worked before so I'm pretty sure my component sets it correctly. Another proof is that if I click up/down it changes to the correct value just one higher/lower than what was set internally.

Example
https://devext.arcgis.com/apps/mapviewer/index.html?url=https://services1.arcgis.com/4yjifSiIG17X0gW4/arcgis/rest/services/FatalAccidents2017/FeatureServer/0

I do not have a test case for this in arcgis-components.

Click layer to select it, open Filter tool on right. Click Add expression, change field to Fatalities. Click up arrow so you see the value 2. Click on the dropdown button on the right of the input and select value 5. The value in the input should change to 5, but it stays at 2. The UI re-renders after picking 5, so the value in the input should change too.

https://devtopia.esri.com/WebGIS/arcgis-app-components/blob/master/src/components/arcgis-filter/arcgis-filter-condition/arcgis-filter-condition.tsx#L589

I think this issue is related to https://github.com/Esri/calcite-components/issues/1972

### Reproduction Steps

1.

AdelheidF:
This looks good now with `localeFormat` set to true. Tested with next.164

PR #1985, opened 2021-04-15
fix(calcite-input): localized number value now correctly updates when value is set programmatically

eriklharper:
**Related Issue:** #1982


jcfranco (approved):
👍🏼

Issue #1983, opened 2021-04-14
Bug: calcite-input type=number inside calciteInputInput event the value on the node is not updated


AdelheidF:
### Summary

In the `calciteInputInput` event the input changes visually but node.value has the old value. The event detail looks right.

Example
https://devext.arcgis.com/apps/mapviewer/index.html?url=https://services1.arcgis.com/4yjifSiIG17X0gW4/arcgis/rest/services/FatalAccidents2017/FeatureServer/0

Click layer to select it, open Styles tool on right. Click Add field and select Fatalities. Click on Counts and Amounts (color), click on the tile again, toggle on Classify data. Click up arrow to see value 5 in input. The 4 dots below that input should change to 5 dots, but it stays at 4. If you click up again the number changes to 6 and the dots count changes to 5. It's always one behind.

https://devtopia.esri.com/WebGIS/arcgis-components/blob/master/src/smartMapping/support/Classify/index.tsx#L174
https://devtopia.esri.com/WebGIS/arcgis-components/blob/master/src/smartMapping/support/Classify/index.tsx#L566

### Reproduction Steps

1.

ethanbdev:
I think a fix went into next for this. It seems better, but now it seems to be stripping the last digit when I get the value, even though it looks correct in the display.
![229F0A7E-3442-4DB0-AEC4-A90D262F9EFB](https://user-images.githubusercontent.com/52869490/114925277-c0a35a00-9de3-11eb-9bd7-a64eb4d35fb8.GIF)

AdelheidF:
I can still repo it with next.164 in that original test case.

Only up clicks...
![image](https://user-images.githubusercontent.com/44481292/114927573-7bccf280-9de6-11eb-89b1-4443d7a45fa8.png)

You can also see that sometimes I have to click twice 'up' to get a new value in the event.

AdelheidF:
This looks good now.

ethanbdev:
Looks okay to me too 👍

PR #1992, opened 2021-04-16
fix(calcite-input): inputInput event now correctly broadcasts the target element's value (event.target.value) to match event.detail.value

eriklharper:
**Related Issue:** #1983 

@driskull This also fixes console.warn messages from appearing in tests and in the browser when entering number values ending in a decimal.  Numbers in the input that end with a decimal are allowed to be typed, but the stored value doesn't update until another number is typed.  The `calciteInputInput` event also won't emit the value if it ends in a decimal, it waits to emit until a number is added to the right of the decimal.


jcfranco (approved):
👍🏼

Issue #1988, opened 2021-04-15
Enhancement: calcite-input type="number" should always display localized decimal separator and opt-in to group separator


eriklharper:
### Description

`calcite-input type="number"` should always display localized decimal separator by default and the group separator should always be handled when parsing, but for display it is based on a prop that is false by default.

### Acceptance Criteria 

- [x] `calcite-input type="number"` should always display localized decimal characters based on the `locale` prop.
- [x] `calcite-input type="number"` should enable opt-in support for displaying the group separator with a `display-group-separator` prop set to false by default.
- [x] `calcite-input type="number"` should allow typing or pasting values that contain the locale's group separator regardless of the `display-group-separator` setting.
- [x] `calcite-input type="number"` should continue to store the unformatted value as before

#### Which Component

`calcite-input`

#### Example Use Case

```
<calcite-input type="number" locale="fr" value="1234.56"></calcite-input> <!-- should display "1234,56" -->
<calcite-input type="number" locale="fr" value="1234.56" display-group-separator></calcite-input> <!-- should display "1 234,56" -->
```

PR #1995, opened 2021-04-19
feat(calcite-input): display localized decimals by default, don't display group separator character by default, but allow opt-in.  Deprecates locale-format prop

eriklharper:
**Related Issue:** #1988 

image

### DIFF
--- a/src/components/calcite-input/calcite-input.tsx
+++ b/src/components/calcite-input/calcite-input.tsx
@@ -466,13 +466,11 @@ export class CalciteInput {
   }
 
   private setLocalizedValue = (unlocalizedValue: string): void => {
-    console.log("setLocalizedValue before", this.localizedValue);
     this.localizedValue = localizeNumberString(
       this.sanitizeNumberString(unlocalizedValue),
       this.locale,
       this.displayGroupSeparator
     );
-    console.log("setLocalizedValue after", this.localizedValue);
   };
 
   private setValue = (value: string, nativeEvent): void => {
@@ -590,7 +588,6 @@ export class CalciteInput {
 
     const suffixText = <div class="calcite-input-suffix">{this.suffixText}</div>;
 
-    console.log("render", this.localizedValue);
     const localeNumberInput = this.shouldFormatNumberByLocale() ? (
       <input
         autofocus={this.autofocus ? true : null}
--- a/src/components/calcite-input/calcite-input.tsx
+++ b/src/components/calcite-input/calcite-input.tsx
@@ -22,7 +22,8 @@ import {
   getDecimalSeparator,
   delocalizeNumberString,
   localizeNumberString,
-  getGroupSeparator
+  getGroupSeparator,
+  sanitizeDecimalString
 } from "../../utils/locale";
 import { numberKeys } from "../../utils/key";
 import { hiddenInputStyle } from "../../utils/form";
@@ -442,10 +443,6 @@ export class CalciteInput {
     this.setValue(this.defaultValue, event);
   };
 
-  private sanitizeNumberString(value: string) {
-    return value.endsWith(".") ? value.replace(".", "") : value;
-  }
-
   private setChildElRef = (el) => {
     this.childEl = el;
   };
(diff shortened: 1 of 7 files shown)


src/components/calcite-input/calcite-input.tsx
```diff
@@ -455,10 +466,13 @@ export class CalciteInput {
   }
 
   private setLocalizedValue = (unlocalizedValue: string): void => {
+    console.log("setLocalizedValue before", this.localizedValue);
```
AdelheidF, line 469: want to leave those 3 console.log messages in?

ethanbdev:
Quick question, will the `display-group-separator` toggle displaying a thousands seperator _or_ toggle displaying a localized thousands seperator?

By default, there is no seperator at all, will that continue to be the default?

eriklharper:
> Quick question, will the `display-group-separator` toggle displaying a thousands seperator _or_ toggle displaying a localized thousands seperator?

Everything is displayed localized by default, including the thousands (group) separator when it is toggled on.  This new prop toggles the group separator on or off, which always displays localized (there's no way to override the localized display).

> By default, there is no seperator at all, will that continue to be the default?

Yes, there is no group separator on by default, this matches native behavior for number inputs too and there was consensus that it is not common for people to type the group character when entering numbers anyway so it makes sense to have this be the default setting for those two reasons.

ethanbdev:
> > Quick question, will the `display-group-separator` toggle displaying a thousands seperator _or_ toggle displaying a localized thousands seperator?
> 
> Everything is displayed localized by default, including the thousands (group) separator when it is toggled on. This new prop toggles the group separator on or off, which always displays localized (there's no way to override the localized display).
> 
> > By default, there is no seperator at all, will that continue to be the default?
> 
> Yes, there is no group separator on by default, this matches native behavior for number inputs too and there was consensus that it is not common for people to type the group character when entering numbers anyway so it makes sense to have this be the default setting for those two reasons.

Nice, thank you! That sounds great.

src/components/calcite-input/calcite-input.tsx
```diff
@@ -70,6 +71,9 @@ export class CalciteInput {
     this.setDisabledAction();
   }
 
+  /** for number values, displays the locale's group separator */
+  @Prop({ reflect: true }) displayGroupSeparator = false;
```
jcfranco, line 75: This prop isn't used for styling, right? If not, I think we can disable reflecting.

jcfranco (approved):
Awesome! ✨ 🏆 ✨

src/components/calcite-input/calcite-input.tsx
```diff
@@ -70,6 +71,9 @@ export class CalciteInput {
     this.setDisabledAction();
   }
 
+  /** for number values, displays the locale's group separator */
+  @Prop() displayGroupSeparator = false;
```
driskull, line 75: Maybe we can rename this to just be `separate` or `separated`  to be more consistent with HTML attributes?

AdelheidF: I feel the name group is necessary here.

eriklharper: +1 for the word "group" just because that is the name of the specific character in the Intl.NumberFormat API that is controlled by this prop:  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/formatToParts

image

AdelheidF:
I assume everything is installed here. Could we get it all reviewed quickly? Thanks!

jcfranco:
The latest test runs have been failing because of color-picker tests timing out. I'll skip them here and create a follow-up issue to restore. They pass locally just fine.

src/components/calcite-input/calcite-input.tsx
```diff
@@ -70,6 +72,9 @@ export class CalciteInput {
     this.setDisabledAction();
   }
 
+  /** for number values, displays the locale's group separator */
+  @Prop() groupSeparator = false;
```
driskull, line 76: nitpick: This could be an optional prop like: `@Prop() groupSeparator?: boolean;`

driskull (approved):
👍

Issue #1999, opened 2021-04-20
Bug: calcite-input type=number ctrl+v pasting does not work on windows


AdelheidF:
### Summary

On a Windows machine I can paste fine by right-click and then choosing Paste. But ctrl-v does not work. ctrl-v does work fine with a calcite-input type=text.

Tested with next.166

### Reproduction Steps

1.

AdelheidF:
This works fine now, tested with next.167

PR #2000, opened 2021-04-20
 fix(calcite-input): allowing ctrl+v (and other modifier keys) for pasting numbers on windows machines

eriklharper:
**Related Issue:** #1999


driskull (approved):
👍

jcfranco (approved):
Looks good. I think we need a test for this too. Let's tackle this in a follow-up issue since we're pressed for time.

PR #2011, opened 2021-04-21
refactor(pick-list-item): Use 'calciteActionClick' event for removing item.

driskull:
fix(pick-list-item): Use 'calciteActionClick' event for removing item.


jcfranco (approved):
LGTM!

Nitpick: this should use the `test` commit type since it is addressing warning messages in test runs.

driskull:
> Nitpick: this should use the test commit type since it is addressing warning messages in test runs.

Yeah, but its making changes to the component so I think fix might be better. If it was just editing a test, then probably test would be good.

jcfranco:
Good point, refactor makes more sense then. It's changing code without affecting implementation and it ends up cleaning up the tests.

Issue #2001, opened 2021-04-20
Bug: Setting `value` property of `calcite-input` to the string "undefined" renders the string itself


ethanbdev:
### Description
If the input is `type=number`, it should handle the string "undefined" as non-numeric characters and render a blank input.
### Acceptance Criteria 
`calcite-input` of type number should only render numeric values.

PR #2013, opened 2021-04-22
fix(calcite-input): setting initial value for number input to "undefined" doesn't display in the input

eriklharper:
**Related Issue:** #2001

### DIFF
--- a/src/components/calcite-input/calcite-input.tsx
+++ b/src/components/calcite-input/calcite-input.tsx
@@ -27,6 +27,7 @@ import {
 } from "../../utils/locale";
 import { numberKeys } from "../../utils/key";
 import { hiddenInputStyle } from "../../utils/form";
+import { isValidNumber } from "../../utils/number";
 
 type NumberNudgeDirection = "up" | "down";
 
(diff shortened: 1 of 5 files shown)


src/utils/number.ts
```diff
@@ -0,0 +1,3 @@
+export function isValidNumber(numberString: string): boolean {
+  return !numberString || isNaN(Number(numberString)) ? false : true;
```
driskull, line 2: Would this work?

`!numberString || !isNaN(Number(numberString));`

src/components/calcite-input/calcite-input.tsx
```diff
@@ -257,6 +258,9 @@ export class CalciteInput {
     this.scale = getElementProp(this.el, "scale", this.scale);
     this.status = getElementProp(this.el, "status", this.status);
     this.step = !this.step && this.shouldFormatNumberByLocale() ? "any" : this.step;
+    if (this.type === "number" && !isValidNumber(this.value)) {
+      this.value = undefined;
```
driskull, line 262: Should this maybe be `null` instead?

driskull (approved):
👍

src/utils/number.ts
eriklharper: That won't work because the function needs to return `false` when the result of calling `isNaN` is true, so its the exact reverse.

PR #2014, opened 2021-04-22
fix(calcite-input): number input localization no longer restricted to static list of locales

eriklharper:
**Related Issue:** #2015


jcfranco (approved):
👍🏼

PR #2021, opened 2021-04-22
fix(modal): render correctly with no footer slots

paulcpederson:
## Summary

Working on a design right now where there are actions in the footer in some steps and not in others. The current implementation is not "live". So if you have slotted buttons the footer shows, but when you remove them it still shows the empty footer with border, etc:

Screen Shot 2021-04-22 at 1 09 41 PM

This pr adds a `MutationObserver` which checks if you have slotted content in the footer when your dom changes. That way you can flip back and forth between having a footer and not having a footer.

### DIFF
--- a/src/components/calcite-modal/calcite-modal.tsx
+++ b/src/components/calcite-modal/calcite-modal.tsx
@@ -1,4 +1,5 @@
 import {
+  Build,
   Component,
   Element,
   Event,
@@ -100,8 +101,19 @@ export class CalciteModal {
     }
   }
 
+  componentDidLoad(): void {
+    this.observer?.observe(this.el, { childList: true, subtree: true });
+  }
+
+  connectedCallback(): void {
+    if (Build.isBrowser) {
+      this.observer = new MutationObserver(this.updateFooterVisibility);
+    }
+  }
+
   disconnectedCallback(): void {
     this.removeOverflowHiddenClass();
+    this.observer?.disconnect();
   }
 
   render(): VNode {
@@ -121,7 +133,8 @@ export class CalciteModal {
           <div
             class={{
               content: true,
-              "content--spaced": !this.noPadding
+              "content--spaced": !this.noPadding,
+              "content--no-footer": !this.hasFooter
             }}
             ref={(el) => (this.modalContent = el)}
           >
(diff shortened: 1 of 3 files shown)


src/components/calcite-modal/calcite-modal.tsx
```diff
@@ -100,8 +101,19 @@ export class CalciteModal {
     }
   }
 
+  componentDidLoad(): void {
+    this.observer?.observe(this.el, { childList: true, subtree: true });
+  }
+
+  connectedCallback(): void {
+    if (Build.isBrowser) {
```
driskull, line 109: nitpick: Just curious, why the build check? Otherwise, you could just create the observer when you create the private property.

```
private observer: MutationObserver = new MutationObserver(this.updateFooterVisibility);
```

driskull (approved):
👍

src/components/calcite-modal/calcite-modal.tsx
paulcpederson: honestly I copied this pattern from elsewhere and assumed there was good reason for it in terms of some hydrated error but if not I would love to get rid of it.

driskull: Yeah i'm not sure when/why it started. My guess is that it was causing an error running in tests?

driskull: @jcfranco do you know? Someone must know :)

driskull: I see this: https://github.com/Esri/calcite-components/blob/master/conventions/README.md#prerendering-and-ssr

But it was from a couple years ago. I'm sure we have many places we're using browser APIs and not checking for the build flag.

jcfranco: It's needed for server-side-rendering: https://stenciljs.com/docs/static-site-generation-basics#things-to-watch-for. I know it was a main focus for components in the beginning, but not sure how many people are using it in that way.

> I'm sure we have many places we're using browser APIs and not checking for the build flag.

True. I think anything in the index page will go through prerendering, so that could be a way to catch problems.

driskull: Is there a way we can test for this? I know there are many places where we are setting up observers in componentConnected and we don't check the build flag. What should we do?

PR #2027, opened 2021-04-23
fix(modal): solve issue with custom widths in prod builds

paulcpederson:
## Summary

In production builds, stencil uses `new CSSStyleSheet()` with `shadowRoot.adoptedStyleSheets` to style components usign shadow dom when browsers support it. This actually overrides the `<style>` tag we have in the render function. To get around this I've used the `!important` sledgehammer. I tried to construct a stylesheet and add it to adoptedStyleSheets but there is very little visibility into the stencil build process and the normal host styles were still winning. This at least will consistently solve the issue.

### DIFF
--- a/src/components/calcite-modal/calcite-modal.tsx
+++ b/src/components/calcite-modal/calcite-modal.tsx
@@ -183,20 +183,20 @@ export class CalciteModal {
       <style>
         {`
         .modal {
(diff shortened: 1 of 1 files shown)


driskull (approved):
I don't like important, but I understand the situation :\

PR #2028, opened 2021-04-23
feat(shell-panel): adds utility class for properly setting the height of external elements and adds related demo apps

asangma:
**Related Issue:** #1927

## Summary
Adds a utility class **to ShellPanel** for setting proper heights and scrolling for content that is not an internal component.
This is an opt-in class.

It should help with multiple levels of nested nodes.
See the nesting-testing demos.


driskull (approved):
👍

PR #2031, opened 2021-04-23
fix(modal): restore footer slotted content

paulcpederson:
## Summary

Earlier pr today incorrectly removed slotted content from the footer of modals. This restores it and corrects the footer display with no slotted content. 

I'm not totally sure why screener didn't pick the regression up? Seems like out modal visual testing is probably not very good...


driskull (approved):
👍

Issue #2002, opened 2021-04-20
Bug: setting the value of a `calcite-input` to `undefined` gives errors in the console


ethanbdev:
### Summary
Rendering `calcite-input` in stencil (could happen in other jsx/vdom) and setting the `value` property to undefined gives errors in the console.

![image](https://user-images.githubusercontent.com/52869490/115462888-0abf7d80-a1e0-11eb-8145-a2a5e58a3b0f.png)

noahmulfinger:
I think I have the same issue in a slightly different context. When a form reset happens before a calcite-input is finished loading into the page, it throws the above error. I created a codepen for it: https://codepen.io/noahmulfinger/pen/yLgGJwj?editors=1010. Seems like it should get fixed by 4021b90, but if not I can open a separate issue.

eriklharper:
Fixed by #2036

PR #2036, opened 2021-04-24
fix(calcite-input): setting value of number inputs restricted to valid numbers or no value

eriklharper:
**Related Issue:** #2002

### DIFF
--- a/src/components/calcite-input/calcite-input.e2e.ts
+++ b/src/components/calcite-input/calcite-input.e2e.ts
@@ -827,7 +827,7 @@ describe("calcite-input", () => {
       expect(await input.getProperty("value")).toBeFalsy();
     });
 
-    it(`disallows setting value to null after initial load.`, async () => {
+    it(`allows setting value to null after initial load.`, async () => {
       const initialValue = "1234";
       const page = await newE2EPage({
         html: `<calcite-input type="number" value="${initialValue}"></calcite-input>`
@@ -841,8 +841,8 @@ describe("calcite-input", () => {
       calciteInput.setProperty("value", null);
       await page.waitForChanges();
 
-      expect(await calciteInput.getProperty("value")).toBe(initialValue);
-      expect(await input.getProperty("value")).toBe(initialValue);
+      expect(await calciteInput.getProperty("value")).toBeFalsy();
+      expect(await input.getProperty("value")).toBeFalsy();
     });
 
     it(`disallows setting text value after initial load.`, async () => {
--- a/src/components/calcite-input/calcite-input.e2e.ts
+++ b/src/components/calcite-input/calcite-input.e2e.ts
@@ -827,7 +827,7 @@ describe("calcite-input", () => {
       expect(await input.getProperty("value")).toBeFalsy();
     });
 
-    it(`disallows setting value to "undefined" after initial load.`, async () => {
+    it(`disallows setting value to null after initial load.`, async () => {
       const initialValue = "1234";
       const page = await newE2EPage({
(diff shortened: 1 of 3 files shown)


src/components/calcite-input/calcite-input.e2e.ts
```diff
@@ -808,5 +808,77 @@ describe("calcite-input", () => {
...
+      const page = await newE2EPage({
+        html: `<calcite-input type="number" value="${initialValue}"></calcite-input>`
+      });
+      const calciteInput = await page.find("calcite-input");
+      const input = await page.find("input");
+
+      expect(await calciteInput.getProperty("value")).toBe(initialValue);
+      expect(await input.getProperty("value")).toBe(initialValue);
+
+      calciteInput.setProperty("value", "null");
```
paulcpederson, line 859: is this actually setting to `null` or the string `"null"`?

eriklharper: the string "null".  These tests are kinda redundant, but wanted to include them anyway.  There's an actual test for the `null` value also in there, same for undefined.

src/components/calcite-input/calcite-input.tsx
```diff
@@ -278,11 +278,19 @@ export class CalciteInput {
...
-    if (this.shouldFormatNumberByLocale()) {
+    if (this.type === "number") {
       this.childEl.style.cssText = hiddenInputStyle;
     }
   }
 
+  componentShouldUpdate(newValue: string, oldValue: string, property: string): boolean {
+    if (this.type === "number" && property === "value" && newValue && !isValidNumber(newValue)) {
+      this.value = oldValue;
+      return false;
```
jcfranco, line 289: Nitpick: Would it be possible to store the conditional expression into an expressive var and return that instead? Totally optional.

src/demos/calcite-input-number.html
```diff
@@ -101,14 +101,29 @@ <h2>Locales</h2>
...
+      console.log("calciteInputBlur", event.target.getAttribute("name"), event.target.value);
     });
+
+    const undefinedInput = document.querySelector("calcite-input[name=set-to-undefined]");
+    document.getElementById("undefined").addEventListener("click", () => {
+      undefinedInput.value = undefined;
+      console.log("undefined", undefinedInput.value);
+    });
+    document.getElementById("undefined-string").addEventListener("click", () => {
+      undefinedInput.value = "undefined";
```
jcfranco, line 119: Similar to the my test comment. This would be the same as testing an invalid number string, right? If so, is it needed?

src/components/calcite-input/calcite-input.e2e.ts
jcfranco: What is this testing exactly? From the component's perspective, it's just receiving an invalid number string, right? If so, I'd suggest removing it since there's another test that would catch this case.

src/components/calcite-input/calcite-input.tsx
```diff
@@ -278,11 +278,19 @@ export class CalciteInput {
   componentDidLoad(): void {
     this.slottedActionEl = this.el.querySelector("[slot=input-action]");
     this.setDisabledAction();
-    if (this.shouldFormatNumberByLocale()) {
+    if (this.type === "number") {
       this.childEl.style.cssText = hiddenInputStyle;
     }
   }
 
+  componentShouldUpdate(newValue: string, oldValue: string, property: string): boolean {
```
jcfranco, line 286: Type-wise, this is not correct as not all prop types are strings.

jcfranco (approved):
👍🏼

PR #2017, opened 2021-04-22
test(color-picker): restore tests

jcfranco:
**Related Issue:** #1998 

## Summary

Restores color-picker test suite.

Issue #1998, opened 2021-04-20
Unstable Test: color-picker tests timing out


jcfranco:
## Summary

**Which component?:** calcite-color-picker

**Which test(s)?:** 

Varied between runs, but some of these are: 

**Test error:**

```
FAIL src/components/calcite-color-picker/calcite-color-picker.e2e.ts (336.948 s)
  ● calcite-color-picker › color inputs › color gets propagated to support inputs › when no-color › restores previous color value when a nudge key is pressed
    : Timeout - Async callback was not invoked within the 30000 ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 30000 ms timeout specified by jest.setTimeout.Error:
      850 |         });
      851 | 
    > 852 |         it("restores previous color value when a nudge key is pressed", async () => {
          |         ^
      853 |           const consistentRgbHsvChannelValue = "0";
      854 |           const initialValue = "#".padEnd(7, consistentRgbHsvChannelValue);
      855 | 
      at new Spec (node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
      at Suite.<anonymous> (src/components/calcite-color-picker/calcite-color-picker.e2e.ts:852:9)
  ● calcite-color-picker › color storage › allows saving unique colors
    The page was already closed
      506 | 
      507 |       await inputOrHexInput.type(value);
    > 508 |       await page.waitForChanges();
          |                  ^
      509 | 
      510 |       await page.keyboard.press("Enter");
      511 |       await page.waitForChanges();
      at Page.r (node_modules/@stencil/core/testing/index.js:3793:11)
      at clearAndEnterValue (src/components/calcite-color-picker/calcite-color-picker.e2e.ts:508:18)
      at assertChannelValueNudge (src/components/calcite-color-picker/calcite-color-picker.e2e.ts:872:13)
      at Object.<anonymous> (src/components/calcite-color-picker/calcite-color-picker.e2e.ts:911:11)
  ● calcite-color-picker › color storage › loads saved colors
    The page was already closed
      506 | 
      507 |       await inputOrHexInput.type(value);
    > 508 |       await page.waitForChanges();
          |                  ^
      509 | 
      510 |       await page.keyboard.press("Enter");
      511 |       await page.waitForChanges();
      at Page.r (node_modules/@stencil/core/testing/index.js:3793:11)
      at clearAndEnterValue (src/components/calcite-color-picker/calcite-color-picker.e2e.ts:508:18)
[...]

PR #2037, opened 2021-04-26
test(color-picker): fix timeout of color picker tests

paulcpederson:
**Related Issue:** #1998

## Summary

Was looking at some of these skipped tests on an ubuntu box and it seems like you can just remove some of the `waitForChanges` calls. This brings the test time down by ~33%. Feel free to close if #2017 works better but thought I'd open a pr JIC before I move on.

### DIFF
--- a/src/components/calcite-color-picker/calcite-color-picker.e2e.ts
+++ b/src/components/calcite-color-picker/calcite-color-picker.e2e.ts
@@ -5,7 +5,7 @@ import { E2EElement, E2EPage, EventSpy, newE2EPage } from "@stencil/core/testing
 import { ColorValue } from "./interfaces";
 import SpyInstance = jest.SpyInstance;
 
-describe.skip("calcite-color-picker", () => {
+describe("calcite-color-picker", () => {
   let consoleSpy: SpyInstance;
 
   beforeEach(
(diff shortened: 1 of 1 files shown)


eriklharper (approved):

jcfranco:
Thanks for this PR, @paulcpederson. I did an overhaul of `page.waitForChanges` in the tests in #2017 after checking this out. I think changing the input assertions is wrong. It should be "", but it's null for some reason in the test. @eriklharper do you know of a case where value would be null? Could it be it needs time to set up?

paulcpederson:
Yeah I saw the null value, ubuntu will use firefox, not chrome headless for the puppeteer run I think. Checking a value attr of a input where value was never set in firefox yields null

eriklharper:
> Thanks for this PR, @paulcpederson. I did an overhaul of `page.waitForChanges` in the tests in #2017 after checking this out. I think changing the channel input assertions is wrong. It should be "", but it's null for some reason in the test. @eriklharper do you know of a case where value would be null? Could it be it needs time to set up?

Yes.  I set the `value` to null here https://github.com/Esri/calcite-components/blob/master/src/components/calcite-input/calcite-input.tsx#L262 when an invalid number is set by default on initial load.  I originally had `undefined` here but I recall a review comment by @driskull that called for using `null` instead.  

In my testing I've noticed that when the value is null internally, it still manifests as empty string when a form is submitted containing the input, so it effectively has the same result?  I remember also having to update some tests to use `toBeFalsy` so perhaps the null setting is the reason why?

jcfranco:
Closing in favor of #1998

jcfranco:
@eriklharper I noticed some inconsistencies w/ setting the input's value to an empty string. Created  https://github.com/Esri/calcite-components/issues/2060 for this.

Issue #2020, opened 2021-04-22
Enhancement: Add Change event to calcite-tile-select-group


paulcpederson:
### Description

Currently the way you listen to changes to a calcite-tile-select-group is to listen for bubbled up input/checkbox events like so:

```ts
const group = document.querySelector("calcite-tile-select-group");

group.addEventListener("calciteRadioButtonChange", (e) => {
  console.log(e.target)
});
```

This is an internal detail and, we should make a new event while stopping propagation of the internal event.

### Acceptance Criteria 

```ts
const group = document.querySelector("calcite-tile-select-group");

group.addEventListener("calciteTileSelectGroupChange", (e) => {
  console.log(e.target.selectedTiles) // => HTMLCalciteTileSelectElement[]
});
```

There is currently no property set on a `calcite-tile-select-group` which allows you to know which are selected. So this would mean a new prop `selectedTiles`. 

If the tile select group uses radio buttons, it may make sense to emit the selected value as part of the detail?

```js
group.addEventListener("calciteTileSelectGroupChange", (e) => {
  console.log(e.detail) // => string
});
```

### Relevant Info

Similar Events:

- [calciteRadioGroupChange](https://github.com/Esri/calcite-components/tree/master/src/components/calcite-radio-group#events)
- [calciteDropdownSelect](https://github.com/Esri/calcite-components/tree/master/src/components/calcite-dropdown#events)
- [calciteLookupChange](https://github.com/Esri/calcite-components/tree/master/src/components/calcite-combobox#events)
- [calciteCheckboxChange](https://github.com/Esri/calcite-components/tree/master/src/components/calcite-checkbox#events)
- [calciteSelectChange](https://github.com/Esri/calcite-components/tree/master/src/components/calcite-select#events)
- [calciteTabChange](https://github.com/Esri/calcite-components/tree/master/src/components/calcite-tab-nav#events)

#### Which Component

calcite-tile-select-group

#### Example Use Case

I'd like to be able to respond to changes to a tile select. This is very common.

/cc @jcfranco @driskull for API feedback

PR #2038, opened 2021-04-26
feat(calcite-tile-select): add calciteTileSelectGroupChange event #2020

### LINKED ISSUE #1020: Enhancement: Tile Select Group Docs - show example of vertically stacked tiles
### Description
The tile select group is capable of being displayed vertically / stacked - but browsing the component Storybook that's not apparent. 

### Acceptance Criteria 
Add a story that shows how that could work.

paulcpederson:
**Related Issue:** #2020

## Summary

Creates a new `calciteTileSelectGroupChange` event. This allows devs to listen for changes to tiles directly on the group. Previously, we used the `calciteRadioButtonChange` and `calciteCheckboxChange` events from the internal input components. Stopping propagation of those events, and documenting this new event.

Also resolves #1020 documenting full width tile select groups by adding knob to the group story.

### DIFF
--- a/src/components/calcite-tile-select-group/calcite-tile-select-group.e2e.ts
+++ b/src/components/calcite-tile-select-group/calcite-tile-select-group.e2e.ts
@@ -23,15 +23,18 @@ describe("calcite-tile-select-group", () => {
     });
     const internalEventSpy = await page.spyOnEvent("calciteCheckboxChange");
     const eventSpy = await page.spyOnEvent("calciteTileSelectGroupChange");
+    const group = await page.find("calcite-tile-select-group");
     const items = await page.findAll("calcite-tile-select");
     await items[0].click();
     expect(eventSpy).toHaveReceivedEventTimes(1);
     expect(internalEventSpy).toHaveReceivedEventTimes(0);
-    expect(await eventSpy.lastEvent.detail.length).toBe(1);
+    let selected = await group.getProperty("selectedItems");
+    expect(selected.length).toBe(1);
     await items[1].click();
     expect(eventSpy).toHaveReceivedEventTimes(2);
     expect(internalEventSpy).toHaveReceivedEventTimes(0);
-    expect(await eventSpy.lastEvent.detail.length).toBe(2);
+    selected = await group.getProperty("selectedItems");
+    expect(selected.length).toBe(2);
   });
 
   it("correctly emits on change event for radio tiles", async () => {
@@ -45,14 +48,14 @@ describe("calcite-tile-select-group", () => {
     });
     const internalEventSpy = await page.spyOnEvent("calciteRadioButtonChange");
     const eventSpy = await page.spyOnEvent("calciteTileSelectGroupChange");
+    const group = await page.find("calcite-tile-select-group");
     const items = await page.findAll("calcite-tile-select");
+    let selected = await group.getProperty("selectedItem");
+    expect(selected).toBeFalsy();
     await items[0].click();
     expect(eventSpy).toHaveReceivedEventTimes(1);
     expect(internalEventSpy).toHaveReceivedEventTimes(0);
-    expect(await eventSpy.lastEvent.detail.length).toBe(1);
-    await items[1].click();
-    expect(eventSpy).toHaveReceivedEventTimes(2);
-    expect(internalEventSpy).toHaveReceivedEventTimes(0);
-    expect(await eventSpy.lastEvent.detail.length).toBe(1);
+    selected = await group.getProperty("selectedItem");
+    expect(selected.value).toBe("one");
   });
 });
--- a/src/components/calcite-tile-select-group/calcite-tile-select-group.e2e.ts
+++ b/src/components/calcite-tile-select-group/calcite-tile-select-group.e2e.ts
@@ -1,4 +1,6 @@
+import { newE2EPage } from "@stencil/core/testing";
 import { accessible, defaults, reflects, renders } from "../../tests/commonTests";
+import { html } from "../../tests/utils";
 
 describe("calcite-tile-select-group", () => {
   it("renders", async () => renders("calcite-tile-select-group"));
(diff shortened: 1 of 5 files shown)


src/components/calcite-tile-select-group/calcite-tile-select-group.tsx
```diff
@@ -15,6 +28,30 @@ export class CalciteTileSelectGroup {
...
+  checkboxChangeHandler(event: CustomEvent): void {
+    this.emitChange(event);
+  }
+
+  private emitChange(event: CustomEvent): void {
+    event.stopPropagation();
+    const selected = nodeListToArray(this.el.querySelectorAll("calcite-tile-select")).filter(
+      (el: HTMLCalciteTileSelectElement) => el.checked
+    );
+    this.calciteTileSelectGroupChange.emit(selected);
```
eriklharper, line 52: In the case of radio-buttons, it would only need to emit the newly checked single value as opposed to an array.  Do you think we should accommodate that so this new custom event only emits the single radio button's information?

src/components/calcite-tile-select-group/calcite-tile-select-group.tsx
```diff
@@ -15,6 +28,30 @@ export class CalciteTileSelectGroup {
...
+  }
+
+  @Listen("calciteCheckboxChange")
+  checkboxChangeHandler(event: CustomEvent): void {
+    this.emitChange(event);
+  }
+
+  private emitChange(event: CustomEvent): void {
+    event.stopPropagation();
+    const selected = nodeListToArray(this.el.querySelectorAll("calcite-tile-select")).filter(
```
eriklharper, line 49: Rather than use `querySelectorAll` here we should be able to setup the radio and checkbox custom events to emit the new values.  Perhaps that might be a more streamlined way of obtaining the new checked items?

eriklharper (approved):

src/components/calcite-tile-select-group/calcite-tile-select-group.tsx
jcfranco: Nitpick: I think we could get rid of `nodeListToArray` and use `Array.from` by relying on the types.

src/components/calcite-tile-select-group/calcite-tile-select-group.tsx
jcfranco: For future-proofing, I think the custom event detail should be an object with a prop for the selected items. I just checked and we're not consistent in this matter, so I guess we should discuss this separately.

jcfranco: WRT Erik's comment, what would be the most useful approach for most use cases? 

1. a convenience prop reflecting the active selection (a la HTMLSelectElement#selectedOptions)
2. a mapping of current selection in the event's payload (a la HTMLCalcite(Pick|Value)ListElement)
3. a mix of 1 & 2 where the event payload has the most recently selected item, which can't be easily determined via a prop
4. no event detail and have users query the DOM for selected items? 👎🏼

jcfranco: To move forward and avoid breaking changes, would it be acceptable to not doc the event detail until we decide on the above?

eriklharper: @jcfranco Having trouble grokking what you mean by # 2 above.  Could you elaborate?

eriklharper: I would lean towards # 1 because that is the pattern in other components.

paulcpederson: I know I'm in the minority, but I like useful event details (2)  because they are really nice to use in react/preact. You don't have to store a `ref` to the element and then programmatically get the prop storing the data, it just comes over the wire. 

Assuming the `target` is going to be a particular element is not always safe because [events in shadow DOM are pretty complicated](https://pm.dartus.fr/blog/a-complete-guide-on-shadow-dom-and-event-propagation/). I've run in to several situations where I'm making a component using shadow dom and the target becomes my shadow dom component rather than the actual emitting component.

eriklharper: I like event details too, I think both event detail and setting the value on the element reference should be supported.  The only question in my mind is which method we should encourage people to use in our documentation.... that part I haven't quite made up my mind about.

driskull: I'm not a fan of the event details containing duplicate information. It's inconsistent with how native HTML elements work with events and can cause double the breaking changes if we decide to rename a prop. It's much cleaner with once source of information (truth).

paulcpederson: The thing is, when you use shadow dom in your app, you don't have access to the actual event target so it's not really "duplicating information". Here is a small stencil example with two components: https://github.com/paulcpederson/stencil-events-example

The only way for me to get the actual event target is to use something like `composedPath()[0]` and hope that's correct, which does not work in ie11, and also does not work in legacy edge. Because of that, what you end up doing is saving a ref to the target:


```ts
private selectMenu: HTMLCalciteSelectMenu;

render() {
  <calcite-select 
    ref={(el)=> this.selectMenu = el} 
    onCalciteSelectChange={(e) => { 
      const value = this.selectMenu.selectedOption?.value;
      // do things with value
    }}
  >
    <calcite-option value="1">1</calcite-option>
    <calcite-option value="1">1</calcite-option>
  </calcite-select>
}
```

This is fine, but when you have useful event details it's much easier:

```ts
render() {
  <calcite-select 
    onCalciteSelectChange={(e) => { 
      // do things with e.detail
    }}
  >
    <calcite-option value="1">1</calcite-option>
    <calcite-option value="1">1</calcite-option>
  </calcite-select>
}
```

Why would we make somebody add refs or reach into `composedPath` to get information from an event when `CustomEvents` allow you to just pass useful info in the event itself? `detail` to me is a perfect description of that data ie. "here's useful details about the event". 

That's my main reason I like the event detail. It's always available. In the UI I'm building I've had to ref looooots of times and it just feels like cruft.

driskull: I think we can make an exception for when the target is in shadow DOM that it can be added to the event.detail. If it's not accessible in light DOM then it's reasonable to have the information in the event.detail.

In many cases, those 'calcite-item's would be slotted though.

As a side note, we aren't really supporting IE11 or LegacyEdge anymore right?

paulcpederson: > I think we can make an exception for when the target is in shadow DOM

All of our components can be in shadow dom when you are building an application using shadow dom... The constraint is not related to shadow usage in our application, but in the consuming application. The code above is app code, not CC code.

jcfranco: @paulcpederson in your example case is the problem the targets when `my-component` is used in an app or another component? I'm having trouble seeing the issue as I can get to either component through the existing event handler.

paulcpederson: think of `my-component` as an application using shadow dom. When I use `my-nested-component`, I need to save a reference to the `my-nested-component` element. The `target` and `srcElement` will always be `my-component` (my app) because it uses shadow dom. Unless I compose the path and reference an element inside it the event does not include `my-nested-component`. Unless I'm doing something wrong?

So in the UI I'm building, we're using shadow dom, so the event target is never `<calcite-select>`. I have to save references to it.

jcfranco: I can't reproduce that. I've updated your sample to use `calcite-select` and the following works:

```ts
// my-component.tsx

<div>
  Click to trigger nested event:
  <calcite-select label="test"
    onCalciteSelectChange={(e) => {
      console.log((e.currentTarget as any).selectedOption.value);
    }}
  >
    <calcite-option>1</calcite-option>
    <calcite-option>2</calcite-option>
  </calcite-select>
</div>
```

Is there something I'm missing?

paulcpederson: hmmm, you are right it's there when accessing the property, but when you log the event it's null?

Screen Shot 2021-04-28 at 11 10 09 AM

paulcpederson: I mean, if it works it works. I'm weirded out it's not in the actual event log, but I just tried it in the other UI we're building and it works fine? IDK there is probably some event propagation magic messing with the currentTarget prop by the time it gets logged or something.  

So assuming that is a thing you can always do, in this pr I should just add an unsettable prop called `selectedTiles`?

src/components/calcite-tile-select-group/calcite-tile-select-group.e2e.ts
```diff
@@ -9,4 +11,51 @@ describe("calcite-tile-select-group", () => {
...
+    const items = await page.findAll("calcite-tile-select");
+    await items[0].click();
+    expect(eventSpy).toHaveReceivedEventTimes(1);
+    expect(internalEventSpy).toHaveReceivedEventTimes(0);
+    let selected = await group.getProperty("selectedItems");
+    expect(selected.length).toBe(1);
+    await items[1].click();
+    expect(eventSpy).toHaveReceivedEventTimes(2);
+    expect(internalEventSpy).toHaveReceivedEventTimes(0);
+    selected = await group.getProperty("selectedItems");
```
paulcpederson, line 36: @jcfranco @driskull do either of you know why the property here in the test environment would be undefined? In the demo site/normal internet it works great. I've never tested these "mutable properties as event detail" components...

paulcpederson:
Moved to read only properties on the tile group rather than event details. Have a problem with the tests, though.

src/components/calcite-tile-select-group/calcite-tile-select-group.e2e.ts
eriklharper: I've seen this too when trying to access state variables in the test environment.  In that particular case I asserted dom changes that come as a result of state changes, rather than the state itself, but it would be kinda nice to be able to assert the private properties of the component too.

paulcpederson: This is a public prop, I thought we could access those?

eriklharper: Yeah, we can, that's weird... 🤔

PR #2047, opened 2021-04-26
fix(tab-title): Remove setting 'dir' attribute in light DOM elements #1831

driskull:
**Related Issue:** #1831

## Summary

fix(tab-title): Remove setting 'dir' attribute in light DOM elements #1831


jcfranco (approved):
👍

PR #2049, opened 2021-04-26
fix(pagination): Remove setting 'dir' attribute in light DOM elements #1831

driskull:
**Related Issue:** #1831

## Summary

fix(pagination): Remove setting 'dir' attribute in light DOM elements #1831


jcfranco (approved):
👍

Issue #2034, opened 2021-04-23
Enhancement: simplify slot names


jcfranco:
### Description

Some components have the component name in some of their slot names (e.g., `notice-title`). These should be simplified to not include the component name.

### Acceptance Criteria 

Slot names are no longer prefixed by the component name.

#### Which Component

alert, chip, input, notice, shell

driskull:
Closing. reopen if still an issue.

PR #2057, opened 2021-04-26
refactor(alert, chip, input, notice, shell)!: simplify slot names

jcfranco:
**Related Issue:** #2034

## Summary

This drops the component-name prefix from some slots for a simpler API.

* calcite-alert
  * alert-title => title
  * alert-message => message
  * alert-link => link
* calcite-chip
  * chip-action => action
* calcite-input
  * input-action => action
* calcite-notice
  * notice-title => title
  * notice-message => message
  * notice-link => link
* calcite-shell
  * shell-header => header
  * shell-footer => footer

### Question

* Should `tab-nav` be simplified to `nav`? (https://github.com/Esri/calcite-components/blob/master/src/components/calcite-tabs/calcite-tabs.tsx#L49)


driskull (approved):
Nice!!!! 🎉 moar consistency. We should let the map viewer peeps know this will be coming. Breaking changez ahead.

Issue #2059, opened 2021-04-26
Bug: Refactoring host element inconsistencies


driskull:
### Actual Behavior

Some of the components are setting attributes on the host element and they maybe shouldn't be. Lets discuss what to do in these scenarios.

### Expected Behavior

Host element should only have needed props and attributes applied to it. Things like aria roles, tabIndex, id and event listeners.

### Reproduction Steps or Sample

1. accordion-item: `<Host aria-expanded={this.active.toString()} icon-position={this.iconPosition} tabindex="0">` Why iconPosition? Should it be a prop? iconPosition is a private variable but gets set on the host element.
1. alert: ` <Host active={active} aria-hidden={hidden.toString()} aria-label={this.label} calcite-hydrated-hidden={hidden} queued={this.queued} role={role}>` Why queued and active?
1. color-picker-swatch: `<Host aria-label={hex} title={hex}>` Why title? Should it be internal?
1. combobox-item: `<Host aria-hidden disabled={this.disabled} scale={scale} tabIndex={-1}>` Why disabled? Scale as well? These are already props.
1. combobox-item-group: `<Host scale={scale}>` Why scale here?
1. dropdown-group: `<Host role="menu" scale={scale} title={this.groupTitle}>` Why title & scale?
1. dropdown-item: `<Host aria-checked={itemAria} dir={dir} isLink={this.href} role={itemRole} scale={scale} selection-mode={this.selectionMode} tabindex="0">` Why isLink, scale, selectionMode?
1. input-message: `<Host calcite-hydrated-hidden={hidden} theme={this.theme}>`Why theme?
1. modal: `<Host aria-modal="true" is-active={this.isActive} role="dialog">` Why is-active?
1. notice: `<Host active={this.active} dir={dir}>` Why active?
1. radio-button: `<Host labeled={!!this.el.textContent}>` Why labeled? what is that for?
1. radio-group-item: `<Host appearance={appearance} aria-checked={checked.toString()} layout={layout} role="radio" scale={scale}>` Why appearnace, layout, scale?
1. slider: `<Host id={id} is-range={this.isRange}>` Why isRange?
1. tab-title: `<Host aria-controls={this.controls} aria-expanded={this.active.toString()}hasText={this.hasText} id={id} role="tab" tabindex={this.disabled ? "-1" : "0"}>` Why hasText?
1. value-list-item: `<Host data-id={this.guid}>` Why data-id?

### Relevant Info

- [x] accordion-item (@jcfranco)
[...]

PR #2064, opened 2021-04-27
fix(radio-group-item): Remove undocumented properties from the Host element. #2059

driskull:
**Related Issue:** #2059

## Summary

fix(radio-group-item): Remove undocumented properties from the Host element. #2059


jcfranco (approved):
👍

PR #2066, opened 2021-04-27
fix(tab-title): Remove undocumented properties from the Host element. #2059

driskull:
**Related Issue:** #2059

## Summary

fix(tab-title): Remove undocumented properties from the Host element. #2059


jcfranco (approved):
👍

Issue #2071, opened 2021-04-28
Bug: [calcite-input] Missing name attribute on generated input element


mpriour:
### Actual Behavior

`calcite-input` components with a `name` attribute do not set that `name` attribute on the generated `input` on initial render. However, once the input is typed into (not just tabbed through), the `input` element now has the reflected `name` attribute on it.

### Expected Behavior

`calcite-input` components with a `name` attribute should reflect that attribute down to the generated `input` element without any user interaction required.

### Reproduction Steps or Sample

I've made a simple example using `calcite-input` components with and without labels and with and without prefix / suffix text. Click on the submit button without typing into the fields and look at the console output. Then type something into one or more fields and click submit. You'll see that the actual input fields now have a `name` attribute.

https://codesandbox.io/s/busy-waterfall-rmpzz

**Here are the relevant parts of that sample**

```html
<form onSubmit={reportNames}>
          First <calcite-input name="firstName" />
          <calcite-label scale="l">Last
            <calcite-input name="lastName" />
          </calcite-label>
          Website <calcite-input prefix-text="https://" name="website" />
          Email <calcite-input suffix-text="@esri.com" name="email" />
          <button>Submit</button>
</form>
```

```js
function reportNames(evt) {
  evt.preventDefault();
  const namedInputs = document.querySelectorAll("input[name]");
  console.log(`Named inputs count is ${namedInputs.length} | Expected 4`);
  const form = evt.target;
  const { lastName, firstName, website, email } = form.elements;
  console.log(`First name is ${firstName?.value}`);
  console.log(`Last name is ${lastName?.value}`);
  console.log(`Website is ${website?.value}`);
  console.log(`Email is ${email?.value}`);
}
```

Before typing into the fields, `reportNames` logs "Named input count is 0" and all the fields destructured from `form.elements` are undefined. After typing into all the fields, the function logs "Named input count is 4" and all the fields destructured from the form are real `input` elements with a value.

### Relevant Info

tested on Chrome 92, Edge 91, Safari 13.2.1 and Firefox 88 on Mac

_Version_: `@esri/calcite-components@1.0.0_beta_54`

- [ ] CDN
- [X] NPM package

eriklharper:
Strange, I can't seem to reproduce this locally with the same preact version you're using.  Regardless I'll explicitly set the name attribute and hope that resolves it.

driskull:
related to #1579

PR #2073, opened 2021-04-28
fix(calcite-input): setting name explicitly on internal input

eriklharper:
**Related Issue:** #2071

I couldn't reproduce this locally, but this fix is still something we want to do not just with the `name` attribute, but with all others that we're currently spreading onto the internal input.


driskull (approved):
👍

Issue #1602, opened 2021-02-23
Research: How to best handle notification of animation complete


driskull:
### Background 

https://github.com/Esri/calcite-components/pull/1510#discussion_r580651509

A lot of our components are using setTimeout to wait for animations to end and then calling setFocus(). This is quite brittle. We should expose the animationend/transitionend event to notify when internal animations are completed.

We need to do some prototype work.

https://github.com/Esri/calcite-components/search?l=TSX&q=setTimeout%28

### Desired Outcome 

setTimeout isn't used to focus on nodes.

paulcpederson:
This is a cool idea!

jcfranco:
What's the priority of this one?

driskull:
I'd say medium to low. I think all of these things work now with the timeouts but they are subject to breaking.

driskull:
I'm wondering if we should use the existing events or add additional events for the transitionEnd?

One issue is that it doesn't seem the `.e2e` tests get notified of transitionEnd.

jcfranco:
Verified via https://codepen.io/jcfranco/pen/gOWzapR?editors=1010

PR #2074, opened 2021-04-28
refactor: Replace setTimeouts for animations with transition events

driskull:
**Related Issue:** #1602

## Summary

refactor: Replace setTimeouts for animations with transition events.

A lot of our components are using setTimeout to wait for animations to end and then calling setFocus(). This is quite brittle. We should expose the animationend/transitionend event to notify when internal animations are completed.


driskull:
@jcfranco @paulcpederson @macandcheese took a stab at this for dropdown and action-menu. Can you give me some feedback?

src/components/calcite-popover/calcite-popover.tsx
```diff
@@ -180,6 +180,9 @@ export class CalcitePopover {
   /** Fired when the popover is opened */
   @Event() calcitePopoverOpen: EventEmitter;
 
+  /** Fired when the popover is opened and the transition has ended */
+  @Event() calcitePopoverOpenTransitionEnd: EventEmitter;
```
driskull, line 184: Could maybe be simplified to `calcitePopoverOpenEnd`?

src/components/calcite-dropdown/calcite-dropdown.tsx
```diff
@@ -334,14 +339,18 @@ export class CalciteDropdown {
...
-  private dropdownFocusTimeout: number;
-
   //--------------------------------------------------------------------------
   //
   //  Private Methods
   //
   //--------------------------------------------------------------------------
 
+  transitionEnd = (event: TransitionEvent): void => {
+    if (event.propertyName === "opacity") {
```
eriklharper, line 349: Does `transitionEnd` fire once for each transitioned property?  Is there a way to listen for when all of the applied transitions end so that you don't have to check specifically for `opacity` here?

src/components/calcite-inline-editable/calcite-inline-editable.tsx
```diff
@@ -222,14 +223,18 @@ export class CalciteInlineEditable {
...
-  private editingFocusTimeout: number;
-
   //--------------------------------------------------------------------------
   //
   //  Private Methods
   //
   //--------------------------------------------------------------------------
 
+  transitionEnd = (event: TransitionEvent): void => {
+    if (event.propertyName === "padding-left" || event.propertyName === "padding-right") {
```
eriklharper, line 233: ☝️ suggestion: perhaps we should define a pattern for how we store and access transitionable properties so that we don't have to arbitrarily access them like this and could instead do something like `this.transitionableProperties.includes(event.propertyName)`.

src/components/calcite-dropdown/calcite-dropdown.tsx
driskull: Yes it fires for each property, there's not a way to know when all are done.

src/components/calcite-inline-editable/calcite-inline-editable.tsx
driskull: Agreed

macandcheese (dismissed):
Great addition, will sure up a lot of those flakey animations!

driskull:
Gotta clean this one up. some failing tests. I think I'll add separate events.

driskull:
Please review once more :)

eriklharper (dismissed):
Excellent work :)

driskull:
@jcfranco can you review?

jcfranco:
@driskull Sorry, I haven't had time to review this yet. I can take a look next week.

driskull:
Is it ok to merge it? @macandcheese and @eriklharper have approved at some time.

driskull:
@julio8a @jcfranco @caripizza what milestone can this be apart of?

jcfranco:
@driskull I think it could go into the next one based on triaging.

julio8a:
Yup, next sprint.

driskull:
@jcfranco can you review? 

If we want, we can make these new events private.

jcfranco:
@driskull Yeah, we should introduce these as internal helpers and then revisit if we want to make them public. I'll prioritize this for the 7/5 – 7/16 milestone and start reviewing today.

driskull:
Updated to make the new events internal.

driskull:
@jcfranco I can haz merge? :)

src/components/calcite-combobox/calcite-combobox.tsx
```diff
@@ -251,6 +250,30 @@ export class CalciteCombobox {
...
+   *  Fired when the combobox is closed
+   * @internal
+   */
+  @Event() calciteComboboxClose: EventEmitter;
+
+  /**
+   *  Fired when the combobox is closed and the transition has ended
+   * @internal
+   */
+  @Event() calciteComboboxCloseEnd: EventEmitter;
```
jcfranco, line 275: Since these prop events have animation tied to them, can they be merged and have them fire after the animation/transition is complete (e.g., only `close` vs `close` + `closeEnd`)? In this particular case, the combobox is truly closed after the animation is done.

jcfranco (changes requested):
Nice work, @driskull! This is looking way better than relying on timeouts. I only had one comment about having the prop-related events fire after animation instead of adding animation-specific events.

---

In the long run (and not for this PR), I'd like to have a pattern and/or utils to make this simpler to adopt. Here are some thoughts:

* provide a util to fire transition end callbacks without having to maintain the listener (e.g., it could use the `once` event listener 
  
  ```tsx
  <div class=close-button onTransitionEnd={emitOnEnd(this.onClose, this.transitionEndProps)} ... />
  ```

  or 

  ```tsx
  <div class=close-button data-transition-end-props="opacity", onTransitionEnd={emitOnEnd(this.onClose)} ... />
  ```
     
  This util could additionally:

  * use the event listener + the `once` arg to auto-remove on invocation. 
* design the pattern/util to be generic enough to cover both transition and animation events

This is of course just a brain dump and by no means final. 🧠 💩

Let me know what you think about this approach. We can also talk about this separately.

src/components/calcite-combobox/calcite-combobox.tsx
driskull: @jcfranco in this case, since both are internal we could do this. However, in some of the other cases there are already events that emit immediately and we can't change them to fire after animation without it being a breaking change that requires updates to tests. 

For example, the following is expected to happen immediately in the tests.

```
/** Fired when the popover is opened */
  @Event() calcitePopoverOpen: EventEmitter;
```

Do we want to change that?

driskull:
I like the first approach.

```
<div class=close-button onTransitionEnd={emitOnEnd(this.onClose, this.transitionEndProps)} ... />
```

src/components/calcite-combobox/calcite-combobox.tsx
driskull: Same with `calciteInlineEditableEditingCancel`

jcfranco: > breaking change that requires updates to tests.

I don't see this is as a breaking change. Currently, when this event fires the popover isn't truly open until the transition is done, right?

driskull: The tests would need to be updated to wait for the transition then.

driskull:
@jcfranco ready for review again.

src/components/calcite-inline-editable/resources.ts
```diff
@@ -4,4 +4,4 @@ export const TEXT = {
   intlConfirmChanges: "Save"
 };
 
-export const transitionProperties = ["padding-left", "padding-right"];
+export const transitionProperty = "border-top-color";
```
jcfranco, line 7: Nitpick: I think we will need a name that communicates how this prop is used. Currently, it's hard to associate that this prop is tied to the end of the open/close transition.

jcfranco (approved):
👍🏼

src/components/calcite-inline-editable/resources.ts
driskull: Agreed. Should it be named according to the event? Like `editingCancelTransitionProp`?

driskull: Maybe we can refactor this part when we have a utility for it?

jcfranco: Sure, no rush on this.

PR #2076, opened 2021-04-29
fix(chip): Remove setting 'dir' attribute in light DOM elements. #1831

driskull:
**Related Issue:** #1831

## Summary

fix(chip): Remove setting 'dir' attribute in light DOM elements. #1831

### DIFF
--- a/src/components/calcite-chip/calcite-chip.scss
+++ b/src/components/calcite-chip/calcite-chip.scss
@@ -63,8 +63,7 @@
 }
 
 .container {
-  @apply inline-flex
-    items-center;
+  @apply inline-flex items-center h-full;
 }
 
 //focus
@@ -106,8 +105,9 @@
 
 :host button {
   @apply m-0
-    p-0
     inline-flex
+    min-h-full
+    max-h-full
     self-stretch
     items-center
     border-none
--- a/src/components/calcite-chip/calcite-chip.scss
+++ b/src/components/calcite-chip/calcite-chip.scss
@@ -18,7 +18,7 @@
   @apply h-5 text--3;
   --calcite-chip-spacing-unit-l: theme("spacing.2");
   --calcite-chip-spacing-unit-s: theme("spacing.1");
-  & slot[name="chip-image"]::slotted(*) {
+  .chip-image-container {
     @apply h-5 w-5;
   }
 }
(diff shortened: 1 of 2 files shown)


src/components/calcite-chip/calcite-chip.scss
```diff
@@ -62,6 +62,11 @@
   --calcite-chip-button-border-radius: 0 50px 50px 0;
 }
 
+.container {
+  @apply inline-flex
+    items-center;
```
caripizza, line 67: It looks like if we add `h-full` here, the close button when in focus will extend the full height of the chip. What do you think?
```css
@apply inline-flex items-center h-full;
```

src/components/calcite-chip/calcite-chip.scss
```diff
@@ -94,18 +99,18 @@
...
 
 :host button {
-  @apply m-0 
-    p-0 
-    inline-flex 
-    self-stretch 
-    items-center 
-    border-none 
+  @apply m-0
+    p-0
```
caripizza, line 109: Should we remove `p-0` here, since padding is added on line 121?

It also looks like we can get the small size Chip's close button to not extend taller than the container by adding min/max height at 100%. What do you think?
```css
@apply m-0
    inline-flex
    self-stretch
    min-h-full /* new */
    max-h-full /* new */
    items-center
    border-none
    cursor-pointer
    bg-transparent
    duration-150
    ease-in-out
    text-color-1;
```

caripizza (commented):
Thanks for this @driskull 🙏  
(I went overboard when I attempted my refactor of Chip, your approach is cleaner, I like it.)

Left a couple comments that aren't crucial, but I think may help match the existing close button/small chip styles.

src/components/calcite-chip/calcite-chip.scss
driskull: Sounds good to me. 👍

src/components/calcite-chip/calcite-chip.scss
driskull: Sounds good to me. 👍  You guys are the design consistency experts :)

caripizza (approved):
Nice Matt 👍

jcfranco (approved):
👍🏼

Issue #1986, opened 2021-04-15
Docs: Update README to remove `nomodule` script tag


dasa:
### Description 

https://github.com/Esri/calcite-components/#script-tag

Can this be updated to remove the `nomodule` script tag and associated text now that older browsers are not supported?

https://github.com/Esri/calcite-components/#browser-support

jcfranco:
Installed. 

@dasa could you help verify?

dasa:
The text below the scripts needs to be updated too.

jcfranco:
Thanks! I can't believe I missed that. 😅  

PR submitted! https://github.com/Esri/calcite-components/pull/2083/

jcfranco:
@dasa Installed!

jcfranco:
Just checked and `GETTING_STARTED` is getting the 🪓 .

jcfranco:
Closing as verified!

PR #2081, opened 2021-04-29
docs(readme): drop nomodule script snippets

jcfranco:
**Related Issue:** #1986

## Summary

Based on our current [browser support](https://github.com/Esri/calcite-components#browser-support), we should no longer include the `nomodule` script in the doc. 

We can still keep it in our test pages until we no longer need to support it (end of 2021?).

cc @julio8a


driskull (approved):
👍

Issue #2003, opened 2021-04-20
Bug: localized formatting inconsistency when pasting numbers repeatedly in number inputs


eriklharper:
### Summary

Create a number input and do not set group-separator. Then paste in number 1,234,567.89, the number will change and remove the group separators. Good. Now paste in the number 1,234,567.89 again. Now the group separators stay.

### Actual Behavior

Create a number input and do not set group-separator. Then paste in number 1,234,567.89, the number will change and remove the group separators. Good. Now paste in the number 1,234,567.89 again. Now the group separators stay.

### Expected Behavior

Pasting in numbers with group separators when `group-separator` property is false should always persist the number without the separators.

### Reproduction Steps

1. Create a number input and do not set group-separator
2. Then paste in number 1,234,567.89, the number will change and remove the group separators
3. Paste in the number 1,234,567.89 again. Now the group separators stay.

PR #2089, opened 2021-05-03
fix(calcite-input): pasting localized and non-localized numbers works as expected for default locale

eriklharper:
**Related Issue:** #2003

### DIFF
--- a/src/components/calcite-input/calcite-input.e2e.ts
+++ b/src/components/calcite-input/calcite-input.e2e.ts
@@ -808,5 +808,59 @@ describe("calcite-input", () => {
       expect(await calciteInput.getProperty("value")).toBeFalsy();
       expect(await input.getProperty("value")).toBeFalsy();
     });
+
+    it(`allows setting value to undefined after initial load.`, async () => {
+      const initialValue = "1234";
+      const page = await newE2EPage({
+        html: `<calcite-input type="number" value="${initialValue}"></calcite-input>`
+      });
+      const calciteInput = await page.find("calcite-input");
+      const input = await page.find("input");
+
+      expect(await calciteInput.getProperty("value")).toBe(initialValue);
+      expect(await input.getProperty("value")).toBe(initialValue);
+
+      calciteInput.setProperty("value", undefined);
+      await page.waitForChanges();
+
+      expect(await calciteInput.getProperty("value")).toBeFalsy();
+      expect(await input.getProperty("value")).toBeFalsy();
+    });
+
+    it(`allows setting value to null after initial load.`, async () => {
+      const initialValue = "1234";
+      const page = await newE2EPage({
+        html: `<calcite-input type="number" value="${initialValue}"></calcite-input>`
+      });
+      const calciteInput = await page.find("calcite-input");
+      const input = await page.find("input");
+
+      expect(await calciteInput.getProperty("value")).toBe(initialValue);
+      expect(await input.getProperty("value")).toBe(initialValue);
+
+      calciteInput.setProperty("value", null);
+      await page.waitForChanges();
+
+      expect(await calciteInput.getProperty("value")).toBeFalsy();
+      expect(await input.getProperty("value")).toBeFalsy();
+    });
+
+    it(`disallows setting text value after initial load.`, async () => {
+      const initialValue = "1234";
+      const page = await newE2EPage({
+        html: `<calcite-input type="number" value="${initialValue}"></calcite-input>`
+      });
+      const calciteInput = await page.find("calcite-input");
+      const input = await page.find("input");
+
+      expect(await calciteInput.getProperty("value")).toBe(initialValue);
+      expect(await input.getProperty("value")).toBe(initialValue);
+
+      calciteInput.setProperty("value", "i am a text value");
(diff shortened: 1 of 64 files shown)


src/utils/number.spec.ts
```diff
@@ -17,3 +17,28 @@ describe("isValidNumber", () => {
...
+describe("parseNumberString", () => {
+  it("returns null for string values that can't compute to a number", () => {
+    expect(parseNumberString("undefined")).toBe(null);
+    expect(parseNumberString()).toBe(null);
+    expect(parseNumberString("")).toBe(null);
+    expect(parseNumberString(undefined)).toBe(null);
+    expect(parseNumberString(null)).toBe(null);
+    expect(parseNumberString("null")).toBe(null);
+    expect(parseNumberString("not a number")).toBe(null);
+    expect(parseNumberString("kjas;lkjwo;aiej(*&,asd;flkj-")).toBe(null);
```
jcfranco, line 30: If this and the next assertion are testing for specific characters, can you either extract a variable that describes what the string is testing or update the string itself with this info if it's simpler.

This also applies to the test below.

src/components/calcite-input/calcite-input.e2e.ts
```diff
@@ -862,5 +862,111 @@ describe("calcite-input", () => {
       expect(await calciteInput.getProperty("value")).toBe(initialValue);
       expect(await input.getProperty("value")).toBe(initialValue);
     });
+
+    it(`disallows pasting just text characters with no initial value`, async () => {
+      const page = await newE2EPage({
```
jcfranco, line 867: When pasting, if internally the component is the same when it starts with an initial value in the HTML or not, I think you can merge the with-initial/without-initial set of tests.

src/components/calcite-input/calcite-input.tsx
```diff
@@ -394,6 +397,20 @@ export class CalciteInput {
...
 
+  private inputNumberInputHandler = (nativeEvent: InputEvent): void => {
+    const value = (nativeEvent.target as HTMLInputElement).value;
+    const delocalizedValue = delocalizeNumberString(value, this.locale);
+    if (nativeEvent.inputType === "insertFromPaste") {
+      if (!isValidNumber(delocalizedValue)) {
+        nativeEvent.preventDefault();
+      }
+      this.setValue(parseNumberString(delocalizedValue), nativeEvent);
+      this.childNumberEl.value = this.localizedValue;
```
jcfranco, line 408: Doesn't `childNumberEl` display `localizedValue` without setting it?

src/components/calcite-input/calcite-input.tsx
```diff
@@ -425,14 +442,14 @@ export class CalciteInput {
...
-      event.key == getDecimalSeparator(this.locale) &&
-      this.localizedValue.indexOf(event.key) === -1
-    ) {
-      return;
+    const decimalSeparator = getDecimalSeparator(this.locale);
+    if (event.key === decimalSeparator) {
+      if (!this.value && !this.childNumberEl.value) {
+        return;
+      }
+      if (this.value && this.childNumberEl.value?.indexOf(decimalSeparator) === -1) {
```
jcfranco, line 450: Should this be checking `localizedValue` instead of going through `childNumberEl.value`?

src/components/calcite-input/calcite-input.tsx
eriklharper: It's kinda complicated.  `localizedValue` is essentially a computed value that almost always is reflected verbatim in `childNumberEl.value` but because the input needs to allow typing decimals which results in a momentary invalid number (because numbers can't end with a period) the typing in the input is less restricted than it might otherwise need to be, and because of this the render method isn't always called when a user types into the input.

src/components/calcite-input/calcite-input.tsx
eriklharper: Nope.  It needs to check `childNumberEl.value` because of the case when a user types in `1234.` which will not update `this.value` to `1234.` because that is an invalid number, so in order to allow typing decimal characters we have to check the actual value in the input instead of `localizedValue` or `value`.  This logic also prevents the user from typing more than 1 decimal character in the input.

src/components/calcite-input/calcite-input.tsx
jcfranco: Thanks for the info. Not sure if will help here, but you can force a render if necessary (w/ `forceUpdate`).

src/components/calcite-input/calcite-input.tsx
jcfranco: Gotcha.

One more thing I spotted here: `value` is guaranteed here: `this.childNumberEl.value`

jcfranco (approved):
👍🏼

src/components/calcite-input/calcite-input.tsx
eriklharper: `forceUpdate` doesn't cut it, tried that.  The only thing that actually works is directly setting the value like this.

src/utils/number.spec.ts
eriklharper: done.

PR #2100, opened 2021-05-04
fix(dropdown-group): Remove undocumented properties from the Host element. #2059

driskull:
**Related Issue:** #2059

## Summary

fix(dropdown-group): Remove undocumented properties from the Host element. #2059

### DIFF
--- a/src/components/calcite-dropdown-group/calcite-dropdown-group.tsx
+++ b/src/components/calcite-dropdown-group/calcite-dropdown-group.tsx
@@ -12,6 +12,8 @@ import {
 import { getElementDir, getElementProp } from "../../utils/dom";
 import { GroupRegistration, ItemRegistration } from "../calcite-dropdown/interfaces";
 import { SelectionMode } from "./interfaces";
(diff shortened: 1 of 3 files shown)


src/components/calcite-dropdown-group/calcite-dropdown-group.tsx
```diff
@@ -104,8 +111,17 @@ export class CalciteDropdownGroup {
       ) : null;
 
     return (
-      <Host role="menu" scale={scale} title={this.groupTitle}>
-        <div dir={dir}>
+      <Host role="menu">
+        <div
+          class={{
+            container: true,
```
caripizza, line 117: Can you clarify why the additional container class on this line? (`class="container container--s"`)

driskull: Its there to be used in the CSS incase needed. The other classes are modifiers in BEM.

caripizza (approved):
👍

Issue #1737, opened 2021-03-11
Enhancement:  Remove styles that directly hook into a theme attribute


driskull:
### Description

I think we'll need to remove styles that directly hook into a theme attribute as well, as that will no longer happen. A quick search reveals the following:

```
src/components/calcite-alert/calcite-alert.scss:
  7  // dark theme
  8: :host([theme="dark"]) {
  9    --calcite-alert-dismiss-progress-background: #{rgba($blk-200, 0.8)};

src/components/calcite-button/calcite-button.scss:
  11  // dark theme
  12: :host([theme="dark"]) {
  13    --calcite-button-transparent-hover: rgba(255, 255, 255, 0.05);

src/components/calcite-chip/calcite-chip.scss:
    9  // dark theme
   10: :host([theme="dark"]) {
   11    --calcite-chip-transparent-hover: rgba(255, 255, 255, 0.05);

  173    // handle dark theme text accessibility for solid
  174:   :host([theme="dark"][color="#{$name}"][appearance="solid"]:not([color="grey"])),
  175:   :host([theme="dark"][color="#{$name}"][appearance="solid"]:not([color="grey"])) button {
  176      color: var(--calcite-ui-background);

  187  
  188:   :host([theme="dark"][color="#{$name}"][appearance="clear"]) {
  189      border-color: $bgColorClearAppearanceDark;

src/components/calcite-inline-editable/calcite-inline-editable.scss:
  66  :host(:not([editing-enabled])) {
  67:   &:not([theme="dark"]) {
  68      &::slotted(*) {

src/components/calcite-input-message/calcite-input-message.scss:
  20  
  21: :host([theme="dark"]) {
  22    --calcite-input-message-floating-background: #{rgba($ui-foreground-1-dark, 0.96)};

src/components/calcite-link/calcite-link.scss:
  8  // dark theme
  9: :host([theme="dark"]) {
  10    --calcite-link-blue-underline: #{rgba($d-bb-420, 0.4)};

src/components/calcite-scrim/calcite-scrim.scss:
  8  // dark theme
  9: :host([theme="dark"]) {
  10    --calcite-scrim-background: #{rgba($blk-240, 0.75)};
```

These vars will need to be moved to the actual global theme variables, I think.

### Acceptance Criteria 

No more variables hooking into theme attribute

### Relevant Info 

https://github.com/Esri/calcite-components/pull/1735#issuecomment-796907826

#### Which Component

All

#### Example Use Case

[...]

PR #2101, opened 2021-05-05
fix(input-message): Remove undocumented properties from the Host element. #2059

driskull:
**Related Issue:** #2059

## Summary

fix(input-message): Remove undocumented properties from the Host element. #2059


caripizza (commented):
Can I remove this theme attribute in #1737 instead? That way we won't need this todo comment and we won't break the input-message's dark theme styles before the var is moved to global.scss.

driskull:
Yes, i'll assign this one to you then.

Issue #2103, opened 2021-05-05
Bug: calcite-action-group - using menu-tooltip slot doesnt show the tooltip on hover of the overflow action


kevindoshier:
Im placing a calcite-tooltip into the "menu-tooltip" slot of a calcite-action-group, but when hovering on the overflow action the tooltip is not being shown

![image](https://user-images.githubusercontent.com/11093161/117165685-8f7ecf80-ad7a-11eb-9756-2203b0d56da5.png)

### Expected Behavior

A tooltip for the overflow action should show up when hovering similiar to how the expand action works

### Reproduction Steps or Sample

1.

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

driskull:
@AdelheidF @kevindoshier installed. Can you let me know once 'next' is built?

AdelheidF:
I'm using `next.194` and I still see no tooltips, in all 3 different ways I tried before.

driskull:
@AdelheidF do you have a sample I can look up? Maybe this is a separate issue to open.

AdelheidF:
```
  private renderTest() {
    return (
      <div
        id="content"
        ref={
          () => {
            console.log(
              this.hostElement.shadowRoot.getElementById("program-button"),
              this.hostElement.shadowRoot.getElementById("body-button")
            );
            // -> null, null

            const tooltip2: HTMLCalciteTooltipElement = document.createElement("calcite-tooltip");
            tooltip2.placement = "auto";
            tooltip2.label = "body";
            tooltip2.referenceElement = document.getElementById("body-button");
            tooltip2.innerHTML = "body";
            document.body.appendChild(tooltip2);
          }
        }
      >
        <calcite-tooltip-manager>
          <calcite-button
            id="program-button"
            ref={(node: HTMLElement) => {
              console.log("button created", this.hostElement.shadowRoot.getElementById("content"));
              // -> null

              setTimeout(() => {
                console.log("waited", this.hostElement.shadowRoot.getElementById("content"));
                // -> returns node

                const tooltip: HTMLCalciteTooltipElement = document.createElement(
                  "calcite-tooltip"
                );
                tooltip.placement = "auto";
                tooltip.label = "program";
                tooltip.referenceElement = node;
                tooltip.innerHTML = "program";
                this.hostElement.shadowRoot.getElementById("content")?.appendChild(tooltip);
[...]

PR #2107, opened 2021-05-05
fix(action-group): 'menu-tooltip' slot to show the tooltip.

driskull:
**Related Issue:** #2103

## Summary

fix(action-group): 'menu-tooltip' slot to show the tooltip.

@jcfranco can you review?

When components are used within a shadow dom, we should first query the shadow dom and then the document. If querying for multiple, both should be queried.


src/utils/dom.ts
```diff
@@ -20,6 +20,66 @@ export function getElementProp(el: Element, prop: string, fallbackValue: any, cr
   return closest ? closest.getAttribute(prop) : fallbackValue;
 }
 
+function getRootNode(el: Element): HTMLDocument | ShadowRoot {
+  return el.getRootNode() as HTMLDocument | ShadowRoot;
+}
+
+export function queryElementsCrossShadowBoundary<T extends Element = Element>(
```
driskull, line 27: @jcfranco I'm not sure if having separate functions for these dom queries is good or if we want something like `getSlotted` that allows `all`?

We need to be able to do the following queries:

- querySelector
- querySelectorAll
- closest
- Potentially `getElementById` if we feel the slight performance increase is worth it over using `querySelector(#id)`

All of these will query the shadowDom first then go up the chain using `getRootNode()`

For `querySelectorAll`, we may want an option to just return the first set of results or return all of the results from each shadowDom and document.

driskull: Let me know what you are thinking about this...

Its rather complicated so I'm not sure about brining it up in a sync meeting.

driskull (commented):
@jcfranco can you review and provide feedback?

src/components/calcite-tooltip/calcite-tooltip.tsx
```diff
@@ -74,7 +74,7 @@ export class CalciteTooltip {
   /**
    * Reference HTMLElement used to position this component.
    */
-  @Prop() referenceElement!: HTMLElement | string;
+  @Prop() referenceElement: HTMLElement | string;
```
driskull, line 77: Changing doc to not make this required since a slotted tooltip doesn't need a referenceElement.

driskull:
@jcfranco can you review?

src/utils/dom.e2e.ts
```diff
@@ -0,0 +1,83 @@
...
+
+      return resultEl?.textContent;
+    });
+
+    expect(text).toBe("Outside Host");
+  });
+
+  it("queryElementsRelativeTo: should query multiple elements", async () => {
+    const results = await page.evaluate((): string[] => {
+      (window as any).setupTestComponent();
```
jcfranco, line 70: Should this be called in `beforeEach` too?

jcfranco (approved):
Pretty rad! 🏄🏼

src/utils/dom.ts
```diff
@@ -20,6 +20,67 @@ export function getElementProp(el: Element, prop: string, fallbackValue: any, cr
...
+export function getRootNode(el: Element): HTMLDocument | ShadowRoot {
+  return el.getRootNode() as HTMLDocument | ShadowRoot;
+}
+
+export function getHost(root: HTMLDocument | ShadowRoot): Element | null {
+  return (root as ShadowRoot).host || null;
+}
+
+// based on https://stackoverflow.com/q/54520554/194216
+export function queryElementsRelativeTo<T extends Element = Element>(selector: string, element: Element = this): T[] {
```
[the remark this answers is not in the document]
driskull: Yeah. `closestElementCrossShadowBoundary` doesn't have it that way though.

src/utils/dom.ts
```diff
@@ -20,6 +20,67 @@ export function getElementProp(el: Element, prop: string, fallbackValue: any, cr
...
+  function queryFromAll<T extends Element = Element>(el: Element, allResults: T[]): T[] {
+    if (!el) {
+      return allResults;
+    }
+
+    if ((el as Slottable).assignedSlot) {
+      el = (el as Slottable).assignedSlot;
+    }
+
+    const rootNode = getRootNode(el);
```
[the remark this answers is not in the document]
driskull: I can't here, it needs to always call both.

src/utils/dom.ts
```diff
@@ -20,6 +20,67 @@ export function getElementProp(el: Element, prop: string, fallbackValue: any, cr
...
+    allResults = [...allResults, ...uniqueResults];
+
+    return host ? queryFromAll(host, allResults) : allResults;
+  }
+
+  return queryFromAll(element, []);
+}
+
+// based on https://stackoverflow.com/q/54520554/194216
+export function queryElementRelativeTo<T extends Element = Element>(
```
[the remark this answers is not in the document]
driskull: I considered that, but it makes the function more complicated with return types depending on input and going with separate methods is more consistent with native DOM methods like querySelector and querySelectorAll.

src/utils/dom.ts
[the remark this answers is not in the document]
driskull: This is what it does...

### Queries an element's rootNode and any ancestor rootNodes
Gets the rootNode (shadowRoot or document) of an element and if the rootNode is a shadowRoot, it queries the element's shadowRoot.host and any ancestor rootNodes, else queries the rootNode.

Maybe:

- queryElementRoots / queryElementsRoots
- queryElementFromRoot / queryElementsFromRoot
- queryElementFromRootNodes / queryElementsFromRootNodes
- rootQueryElement / rootQueryElements

What do you think @jcfranco ?

src/utils/dom.e2e.ts
```diff
@@ -0,0 +1,83 @@
...
+      }
+
+      customElements.define("test-component", TestComponent);
+
+      const testComponent = document.createElement("test-component");
+      testComponent.innerHTML = "<button>Inside Host</button>";
+      document.body.appendChild(testComponent);
+    }
+
+    await page.addScriptTag({
```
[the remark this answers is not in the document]
driskull: I tried but when I use the whole module it doesn't run it in browser.

src/utils/dom.ts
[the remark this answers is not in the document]
jcfranco: I'm not concerned about DOM API consistency for internal utils. Also, unlike browsers, any duplicate code ends up being downloaded by the user.

Also, different return types based on input should be fine as long as the types are correct.

src/utils/dom.ts
jcfranco: Sorry, I missed this comment. I like the idea of having a single util to would allow us to call DOM APIs across shadow boundary. Maybe something like (very rough):

```ts
domQuery<T extends HTMLElement>(
  method, // `selector`, `closest`, others?
  options // `all`, others?
): Element | Elements | null; // we can fine tune results based on input arguments
```

We can iterate on this if you think it'd be useful.

src/utils/dom.ts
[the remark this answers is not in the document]
jcfranco: I like queryElementRoot(s).

src/utils/dom.ts
driskull: Yeah let's iterate on it.  I'd like to get this in so that the issue is fixed.

Since this is all internal, lets proceed with the current name and we can revisit with a complete dom query API.

src/utils/dom.e2e.ts
driskull: I don't think so, because it needs to modify the document.

src/utils/dom.ts
jcfranco: Sounds good.

src/utils/dom.ts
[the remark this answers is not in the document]
jcfranco: I must've been tired when I reviewed since I can clearly see now that you need both. 😅

src/utils/dom.ts
driskull: Cool i'll create a new issue after merge.

Issue #1537, opened 2021-02-11
Testing: Add test that checks if CSS vars successfully updated values for theming


asangma:
## Summary 
Our Tailwind config references CSS vars which should allow users to theme component colors and fonts.
We need to check to ensure that those CSS vars are working in that way.

## Measure of Success 

- Add unit test set to make sure that css variables are defined.
- Add unit test set to make sure that overwriting each calcite css variable does in fact render the override.
- The tests can use the  getComputedStyle and see if it is valid.

driskull:
@caripizza can you take this one?

caripizza:
Update: I added a bunch of e2e tests for this issue in pr #2114, after we removed the `theme` attribute from component .scss files. However, there are still 26 component stylesheets referencing the global theming CSS vars, so per this issue they'll need e2e tests as well:

```
# Query: --calcite-ui-
# Including: src/components/**/*.scss

298 results - 26 files

src/components/calcite-alert/calcite-alert.scss:
   59:     background-color: var(--calcite-ui-foreground-1);
  109:   color: var(--calcite-ui-text-1);
  120:   color: var(--calcite-ui-text-2);
  194:   color: var(--calcite-ui-text-3);
  200:     background-color: var(--calcite-ui-foreground-2);
  204:     background-color: var(--calcite-ui-foreground-3);
  226:   color: var(--calcite-ui-text-2);
  239:     border-left: 1px solid var(--calcite-ui-border-3);
  240:     border-right: 1px solid var(--calcite-ui-border-3);
  284: $alertColors: "blue" var(--calcite-ui-brand), "red" var(--calcite-ui-danger), "yellow" var(--calcite-ui-warning),
  285:   "green" var(--calcite-ui-success);

src/components/calcite-block-section/calcite-block-section.scss:
   8:   border-bottom: 1px solid var(--calcite-ui-border-3);
  18:   color: var(--calcite-ui-text-2);
  38:     color: var(--calcite-ui-text-1);

src/components/calcite-button/calcite-button.scss:
  205:     opacity: var(--calcite-ui-opacity-disabled);
  259:       var(--calcite-ui-brand),
  260:       var(--calcite-ui-brand-hover),
  261:       var(--calcite-ui-brand-press),
  262:       var(--calcite-ui-foreground-1)
  270:       var(--calcite-ui-danger),
  271:       var(--calcite-ui-danger-hover),
  272:       var(--calcite-ui-danger-press),
[...]

PR #2114, opened 2021-05-07
feat(alert, button, chip, inline-editable, input-message, link, scrim): move theme attribute styles to global.scss

caripizza:
**Related Issues:** #1737, #1537, #1500

## Summary
This pr moves remaining components' styles which hooked directly into the `theme` attribute over to the global stylesheet instead. It adds e2e tests for their theme styles, and also updates calcite-chip's styles to match what I see in Figma (and removes that Sass loop for the colors).


src/assets/styles/global.scss
```diff
@@ -20,6 +20,17 @@
...
+  --calcite-button-transparent-hover: rgba(0, 0, 0, 0.05);
+  --calcite-button-transparent-press: rgba(0, 0, 0, 0.08);
+  --calcite-chip-solid-bg-light: #{rgba($ui-text-inverse-light, 0.1)};
+  --calcite-chip-solid-bg-dark: #{rgba($ui-text-inverse-dark, 0.1)};
+  --calcite-chip-font-color: #{$ui-text-1-light};
+  --calcite-chip-transparent-hover: rgba(0, 0, 0, 0.05);
+  --calcite-chip-transparent-press: rgba(0, 0, 0, 0.08);
+  --calcite-input-message-floating-background: #{rgba($ui-foreground-1-light, 0.96)};
+  --calcite-link-blue-underline: #{rgba($h-bb-070, 0.4)};
+  --calcite-scrim-background: #{rgba($blk-000, 0.75)};
```
caripizza, line 33: chip and button use the same transparent colors here, should we combine?

bstifle: chips should not have any transparency

bstifle:
> @bstifle can you let me know how chip's new styles look here based on Figma?

There shouldn't be any chips with transparency

caripizza:
calcite-modal renders a calcite-scrim with [`theme="dark"`](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-modal/calcite-modal.tsx#L125), so Screener picked up this style change because the `--calcite-scrim-background` CSS property is now nested inside calcite-modal that has the default non-dark theme (ie., light).

Is this intended? Or should calcite-scrim just have the dark theme background style in general?

paulcpederson:
scrim for modal should always be dark, a white scrim in dark mode doesn't make sense, IMO

driskull:
Nice work @caripizza! Looks good to me.

> Are you saying the close button shouldn't have transparency when you hover or focus it?

It's safer to not use transparency here since it could affect the a11y readability when hovered or focused. Could we use a text-color/background change instead?

src/assets/styles/global.scss
bstifle: yep my bad

bstifle:
> Nice work @caripizza! Looks good to me.
> 
> > Are you saying the close button shouldn't have transparency when you hover or focus it?
> 
> It's safer to not use transparency here since it could affect the a11y readability when hovered or focused. Could we use a text-color/background change instead?

Thats not accurate. Hover states don't actually need to pass A11y, even so it is being applied behind the X, and graphical elements need only be 3:1

bstifle:
> Could we use a text-color/background change instead?


We could use a hex value. but we dont have any variables for lighter/darker greens/yellows anymore

driskull:
> Thats not accurate. Hover states don't actually need to pass A11y, even so it is being applied behind the X, and graphical elements need only be 3:1

My bad!

src/assets/styles/global.scss
```diff
@@ -20,6 +20,12 @@
   --calcite-ui-opacity-disabled: 0.5;
 
   --calcite-panel-width-multiplier: 1;
+  --calcite-alert-dismiss-progress-background: #{rgba($blk-000, 0.8)};
```
paulcpederson, line 23: @bstifle I wonder if in the future we should just add a sequence of transparent blacks to the palette? Seems like most of these are just various amounts of semi-transparent black

paulcpederson (approved):

Issue #2102, opened 2021-05-05
Bug: calcite-combobox- updating the selected item doesnt update the "title" of the combobox


kevindoshier:
If I update the selected item for the combobox, the title of the combobox doesnt seem to reflect the change, it remains to be the previously selected item

https://github.com/Esri/calcite-components/issues/2097

### Reproduction Steps or Sample

1.

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

driskull:
@paulcpederson this is because the 'calciteComboboxItemChange' doesn't fire unless a combobox item is clicked. Do you know why?

driskull:
Related to the discussion in #2033

In the components i've worked on i've just had a watcher that emits so that its always 1:1 (all prop changes are emitted no matter how they changed).

So we need a pattern defined I guess.

paulcpederson:
Is this in a particular selection mode? https://codepen.io/paulcp/pen/xxqxqOG?editors=1010 changing selected on an item should update the displayed chips, etc

driskull:
Like this:
https://codepen.io/driskull/pen/NWpWpZm?editors=1010

driskull:
If you try the codepen I provided you see it's all messed up :\

paulcpederson:
Oh of course, right. Yeah I think the combobox item's change event is not propagated outside the tree, so it's probably safe to  emit that on changes to the property.

julio8a:
Verified with latest on Codepen.

PR #2118, opened 2021-05-07
fix(combobox, combobox-item): Toggling 'selected' on combobox-item should update the combobox parent. #2102

driskull:
**Related Issue:** #2102

## Summary

fix(combobox, combobox-item): Toggling 'selected' on combobox-item should update the combobox parent. #2102

### DIFF
--- a/src/components/calcite-combobox/calcite-combobox.e2e.ts
+++ b/src/components/calcite-combobox/calcite-combobox.e2e.ts
@@ -311,11 +311,11 @@ describe("calcite-combobox", () => {
 
       await input.press("Enter");
       expect(await item1.getProperty("selected")).toBe(true);
-      expect(eventSpy).toHaveReceivedEventTimes(1);
+      expect(eventSpy).toHaveReceivedEventTimes(2);
 
       await input.press("Enter");
       expect(await item1.getProperty("selected")).toBe(false);
-      expect(eventSpy).toHaveReceivedEventTimes(2);
+      expect(eventSpy).toHaveReceivedEventTimes(4);
     });
 
     it("should delete last chip on Delete", async () => {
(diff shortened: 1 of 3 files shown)


driskull:
Still active

driskull:
@jcfranco can you review? The PR is stale :)

src/components/calcite-combobox/calcite-combobox.tsx
```diff
@@ -137,7 +137,7 @@ export class CalciteCombobox {
 
   @Listen("calciteComboboxItemChange")
   calciteComboboxItemChangeHandler(event: CustomEvent<HTMLCalciteComboboxItemElement>): void {
-    this.toggleSelection(event.detail);
+    this.toggleSelection(event.detail, event.detail.selected);
```
jcfranco, line 140: I think we should drop `selected` and go through the event's target.

src/components/calcite-combobox/calcite-combobox.e2e.ts
```diff
@@ -311,11 +311,11 @@ describe("calcite-combobox", () => {
 
       await input.press("Enter");
       expect(await item1.getProperty("selected")).toBe(true);
-      expect(eventSpy).toHaveReceivedEventTimes(1);
+      expect(eventSpy).toHaveReceivedEventTimes(2);
 
       await input.press("Enter");
       expect(await item1.getProperty("selected")).toBe(false);
-      expect(eventSpy).toHaveReceivedEventTimes(2);
+      expect(eventSpy).toHaveReceivedEventTimes(4);
```
jcfranco, line 318: The lookup event should only occur once on `Enter`, shouldn't it?

driskull: Well, if there's already one selected then that one becoming unselected also fires an event.

jcfranco: Technically, yes, but should it? It's one operation firing 2 change events. I think the pick/value list tackled something similar some time ago by debouncing the emitter.

driskull: Gotcha. yeah i'll update that 👍

jcfranco (approved):
Awesome! 💥

Issue #2127, opened 2021-05-11
Bug: <calcite-input type="number"> allows non-numeric characters


patrickarlt:
### Actual Behavior

`<calcite type="number">` allows all non-numeric non-alphabet characters.

### Expected Behavior

`<calcite type="number">` should not even allow you to enter non numeric characters potential related to https://github.com/Esri/calcite-components/issues/2009

### Reproduction Steps or Sample

![2021-05-11_14-32-54 (1)](https://user-images.githubusercontent.com/378557/117889070-cae03900-b267-11eb-8f53-2813cd1341b2.png)

1. Configure a numeric input on https://esri.github.io/calcite-components/?path=/story/components-controls-input--with-label
2. Enter any combination of non `1-9a-zA-Z` characters.

### Relevant Info

Any

_Version_: `@esri/calcite-components@v1.0.0-beta.54`

- [x] CDN
- [x] NPM package

macandcheese:
What browser was the above screenshot from? 

AFAIK each browser treats the number input field differently: only Chrome limits it to numbers (and "e") - calcite-input behaves this way in Chrome. 

In Safari and Firefox, native inputs are NOT limited to just numbers - you can enter anything, and calcite-input behaves this way too.

Related: https://github.com/Esri/calcite-components/issues/961#issuecomment-736720527

@eriklharper I see the addition in the above PR  - it could be a "feature" of our calcite-input in type="number" to behave in that way - but this is a departure from (bespoke) browser functionality.

eriklharper:
Thanks for that added context @macandcheese .  I didn't realize the breadth of differences between browsers and this behavior.  Sounds like there is a big difference in opinion too when it comes to what a number input UI/UX _should_ be.  Honestly, I would love to remove all this keydown validation to prevent users from typing anything but valid number characters, but I might be in the minority on a decision like that.  It looks like Firefox's number input doesn't restrict you to typing a just single decimal, nor does it restrict typing a comma like Chromium does, and we have that kind of validation in there from my recent locale changes.

I'm super curious what others think about this issue?  @Esri/calcite-components-reviewers

bstifle:
what about copy/pasting numbers into that field? if it has a comma, will it jack it up?

eriklharper:
In Firefox the number `1,234.56` pastes verbatim in a number input.  In Chromium it strips the comma out, resulting in `1234.56`.

eriklharper:
[...]

PR #2128, opened 2021-05-11
fix(calcite-input): disallowing typing any key with shift modifier down inside a number input

eriklharper:
**Related Issue:** #2127

### DIFF
--- a/src/utils/key.ts
+++ b/src/utils/key.ts
@@ -25,3 +25,31 @@ export function getKey(key: string, dir?: "rtl" | "ltr"): string {
 }
 
 export const numberKeys = ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9"];
+export const letterKeys = [
+  "a",
+  "b",
+  "c",
+  "d",
+  "e",
+  "f",
+  "g",
+  "h",
+  "i",
+  "j",
+  "k",
+  "l",
+  "m",
+  "n",
+  "o",
+  "p",
+  "q",
+  "r",
+  "s",
+  "t",
+  "u",
+  "v",
+  "w",
+  "x",
+  "y",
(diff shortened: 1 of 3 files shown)


src/utils/key.ts
```diff
@@ -25,3 +25,31 @@ export function getKey(key: string, dir?: "rtl" | "ltr"): string {
 }
 
 export const numberKeys = ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9"];
+export const letterKeys = [
```
macandcheese, line 28: I noted it here as well - https://github.com/Esri/calcite-components/issues/2127#issuecomment-839392889, but this doesn't match all browser's native behavior with native input - only Chrome limits this (and it allows the entry of a single "e"). 

I think it would be a reasonable feature, so long as we document that its a departure from native functionality - even though that native behavior is inconsistent (and maybe unexpected) across browsers! 

Thread on Bugzilla : https://bugzilla.mozilla.org/show_bug.cgi?id=1398528

eriklharper: Didn't think about the Euler's constant, thanks for bringing that up!  I'll get some other folks' take on this.

macandcheese: Cool - Just to be clear I’m not against having this baked in at all - most of the time it seems reasonable / expected -  so long as it’s documented / functionally doesn’t prevent folks from using the number input in a way they might expect a native one to!

jcfranco: ☝🏼 Agreed. We can add support for additional characters (e.g., `e`, `+`, `-`) in a follow-up PR.

jcfranco (approved):
👍🏼  #️⃣ 1️⃣

Issue #2108, opened 2021-05-06
Docs: add RTL stories for Storybook docs/Screener tests


caripizza:
## Summary 
Now that #1831 Epic has landed, ~we need e2e tests which confirm the component can inherit `dir="rtl"` from further up the DOM tree.~

Update per [this comment](https://github.com/Esri/calcite-components/issues/2108#issuecomment-846188037):
>I'll remove the rtl tests since they're just testing the browser default and we have the new ESLint rule, which checks for dir attribute on the host.

~(I'm thinking the tests can use `.getComputedStyle()` to check that the component's `dir` CSS property value is `rtl`.)~

## Measure of Success 
~All components have an e2e test which asserts that the component can inherit its rtl direction from an ancestor.~
All components have an RTL story for Screener test/Storybook docs.

### Resources 
This is a follow-on that we're hoping will help us verify all changes from Epic #1831.

driskull:
@jcfranco can you take this one? I'm not sure how it will work because testing 'dir' would pass if the 'dir' tag is on it.

jcfranco:
![10-4](https://user-images.githubusercontent.com/197440/117330439-8b21e780-ae4a-11eb-8995-c4f4cdf2d5e8.gif)

driskull:
Yeah that should work. 👍 

It should also make sure that the `dir` attribute is not applied onto the component tag in the first scenario.

jcfranco:
Do we need to test for something internally? With the new `@calcite-components/ban-props-on-host` ESLint rule, `dir` should no longer be allowed on the host, so the above snippet is testing default browser behavior and nothing from the components themselves.

caripizza:
@jcfranco should I rework the commonTests then? (I only added them to tile, tile-select, and tile-select-group to start)
https://github.com/Esri/calcite-components/blob/master/src/components/calcite-tile-select/calcite-tile-select.e2e.ts#L132

jcfranco:
Not unless we can find a way to generalize the screenshot diffing you've got there. Maybe we can rely on Screener for this and have each storybook have an RTL story?

caripizza:
Works for me! I almost created the issue yesterday but ran out of time. Here's the notes I took:

Based on a quick audit of master, these components that were touched in 1831 don't have RTL stories yet and should:
- Button
- Card
- Checkbox
- Date Picker
- Input (story should include Input Message too)
- Input Date Picker
- Label
- Link
- Modal
[...]

Issue #1230, opened 2020-11-05
Enhancement: Add RTL support for tile-select


macandcheese:
### Description

Tile select needs to support `dir="rtl"`:

<img width="1547" alt="Screen Shot 2020-11-05 at 11 08 36 AM" src="https://user-images.githubusercontent.com/4733155/98285605-a4a7ac80-1f57-11eb-9554-a11e60055cc3.png">

~We should also change the nomenclature for "show-input" from "left/right" to "start/end"~ (this is already implemented - Cari)

caripizza:
Adding an update that Tile Select's rtl UI appears properly mirrored on page render. It's just using default browser behavior when text direction is rtl. In pr #2129 I added RTL stories to all Tile components which shows the UIs on render:

![image](https://user-images.githubusercontent.com/42423180/119165596-821c4300-ba12-11eb-9148-ce018ebd67c3.png)
![image](https://user-images.githubusercontent.com/42423180/119165641-906a5f00-ba12-11eb-91d0-1b08e43a38ca.png)
![image](https://user-images.githubusercontent.com/42423180/119165659-952f1300-ba12-11eb-99fc-dcb3c433a805.png)

As a part of #2108, Franco and I are removing the rtl commonTests I had added, and instead focusing on rtl stories for Screener.

Marking this one verified, since the RTL stories now exist on master.

PR #2129, opened 2021-05-12
feat(tile, tile-select, tile-select-group): adds RTL support docs, tests, screenshots 

caripizza:
**Related Issues:** #1230, #2108, #1020

## Summary
This pr updates the test npm script to run screenshot tests, and checks the screenshots into github. I'm not sure why but these were gitignored and weren't even running since we didn't pass the `--screenshot` flag. We don't have to use this feature (it's still in "experimental status" from Stencil), but I think it could be helpful since we can interact with the tests using Puppeteer, and potentially use `.click()` before taking any screenshots.)

Based on what I can see, Tile Select supports RTL. It just didn't have docs or html showing it worked on initial render. I have added the stories and updated the e2e tests per 2108. I also updated the tile/tile-select demo html so that each page shows the examples in LTR and RTL on initial render, and removed the light/dark classes since we can use the new "Theme" toggle.


src/components/calcite-tile-select-group/calcite-tile-select-group.e2e.ts
```diff
@@ -9,4 +10,64 @@ describe("calcite-tile-select-group", () => {
...
+        expect(await (await tileSelectGroupEl.getComputedStyle()).getPropertyValue("direction")).toEqual("ltr");
+        expect(tileSelectGroupEl.getAttribute("dir")).toBeNull();
+        // 1: screenshot diff for LTR
+        const results = await page.compareScreenshot();
+        expect(results).toMatchScreenshot({ allowableMismatchedPixels: 100 });
+      });
+    });
+
+    describe("when inheriting direction from further up the DOM tree", () => {
+      it("should honor ancestor's `dir` attribute, and not have its own `dir` attribute", async () => {
```
driskull, line 43: Should we make this a common test? So it can be used in every file?

src/components/calcite-tile-select-group/calcite-tile-select-group.e2e.ts
```diff
@@ -9,4 +10,64 @@ describe("calcite-tile-select-group", () => {
...
+        `
+        });
+        const tileSelectGroupEl = await page.find("calcite-tile-select-group");
+        expect(await (await tileSelectGroupEl.getComputedStyle()).getPropertyValue("direction")).toEqual("rtl");
+        expect(tileSelectGroupEl.getAttribute("dir")).toBeNull();
+      });
+    });
+
+    describe("when it has its own `dir` attribute", () => {
+      it("should render with text direction based on `dir` value", async () => {
```
driskull, line 60: Should we make this a common test? So it can be used in every file?

driskull (approved):
Looks great @caripizza 🍕 

Should we have an issue to convert these tests into the common file or do it in this PR? So we can use them elsewhere for testing RTL :)

caripizza:
@driskull good call - I'll rework those as common tests in this pr!

driskull:
Looks good!

Issue #1740, opened 2021-03-11
Bug: remove calcite-label from internals of checkbox and radio


driskull:
### Summary

Bug: remove calcite-label from internals of checkbox and radio

### Actual Behavior

these two components automatically add a calcite-label internally

### Expected Behavior

User is expected to add their own calcite-label externally

driskull:
@julio8a should this be a pre-1.0 breaking change?

julio8a:
Yes, likely. I want to spend some time reviewing all the `p-high` labels to see what else we have that might be breaking needed for 1.0

PR #2148, opened 2021-05-13
fix(calcite-radio-button): removing internal label

eriklharper:
**Related Issue:** #1740


src/components/calcite-radio-button/calcite-radio-button.tsx
```diff
@@ -333,56 +333,35 @@ export class CalciteRadioButton {
...
-            hidden={this.hidden}
-            hovered={this.hovered}
-            ref={(el) => (this.radio = el)}
-            scale={this.scale}
-            theme={this.theme}
-          />
-          {this.renderLabel()}
-        </div>
+        <input
+          aria-label={value || this.guid}
```
driskull, line 342: question: why is a guid be using as as label?

driskull, line 342: Do we need an aria-label here? Shouldn't the radio button be wrapped in a label or assigned an aria-label itself?

driskull: If we do need a label, would a separate prop be best? I'm not sure 'value' would always make sense. If the value is some ID, it's not very screen reader friendly.

driskull (commented):
Looks good, just had a few questions. 👍

src/components/calcite-radio-button/calcite-radio-button.tsx
eriklharper: Good question, I'm wondering if I can just do `this.el.getAttribute("aria-label")` and not even define a specific prop here, but like you said the user should provide the label either as a wrapper or with the for attribute.

eriklharper: I'm thinking since this input is internal, it shouldn't need this attribute because the label would need to be on the radio-button element instead.  I'll remove it and see if accessibility tests still pass.

driskull (approved):
👍

jcfranco (commented):
Nice! 💥

Issue #2098, opened 2021-05-04
Bug: ValueListItem drag-handle is the wrong color


asangma:
### Actual Behavior
The icon in a value-list-item is incorrect.

### Expected Behavior
The icon should be 

| state | color |
|-----|-------|
| idle | border-color-3 |
| hover | text-color-1 |
| activated | text-color-1 |

cc @bstifle 

### Reproduction Steps or Sample

1. view `demos/value-list/advanced.html`

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

asangma:
Mind verifying this one?

asangma:
@julio8a Mind verifying this one?

bstifle:
@asangma should be border-1 #CACACA on idle 😇

julio8a:
Based on Bryan's comment, should I wait to verify then?

asangma:
See new PR.

asangma:
Updated with border-1

asangma:
Thanks for catching that @bstifle!

julio8a:
Nice! thank @asangma. Verified on master.

PR #2155, opened 2021-05-17
fix(value-list-item): updates handle colors to match CalciteHandle

asangma:
**Related Issue:** #2098

## Summary
Styles internal "handle" element to match CalciteHandle.

Refactoring ValueListItem to use Handle would be cool.
We are going to have a general ListItem, and when ValueListItem is refactored to use that component, we can take care of this then.


driskull (approved):
👍

Issue #1906, opened 2021-04-02
Bug: calcite-action-menu styling


AdelheidF:
### Summary
In the filter component I switched from a calcite-dropdown to a calcite-action-menu.

dropdown:
<img src=https://user-images.githubusercontent.com/44481292/113449318-6ac9bd80-93b2-11eb-81b6-edf763eea4f5.png width=200/><img src=https://user-images.githubusercontent.com/44481292/113449425-a82e4b00-93b2-11eb-9ad4-57597b09cf1a.png width=200/>

action-menu:
<img src=https://user-images.githubusercontent.com/44481292/113449220-35bd6b00-93b2-11eb-812c-18e54531d40d.png
 width=200/><img src=https://user-images.githubusercontent.com/44481292/113449601-09561e80-93b3-11eb-9ca3-6c0255f992ee.png width=200/>

(1)
To me the new action menu now looks more like a tooltip/more info than a dropdown menu with actions, especially in regards to the triangle pointer. Also, this seems to be the only component that I use that has round corners. Everything else that I use has sharp corners.

(2)
With `1.0.0-beta.54` the blue focus box went missing
![image](https://user-images.githubusercontent.com/44481292/116929706-a9ea6900-ac13-11eb-9fe8-35636dd9fe99.png)

If this is all purpose, please close this issue.

cc @driskull 

### Reproduction Steps

1.

driskull:
@asangma we should decide if the action menu should have a pointer or not.

Since both the `panel` and `action-bar` use the `calcite-action-menu`, we should probably be consistent right?

TheBlueDog:
Following. Agreed. The mix of radius and non-radius seems odd.

macandcheese:
Could those actions just render as dropdown-items of "selection-mode:none" in a dropdown which would still not have a tail but have the correct 4px border radius?

AdelheidF:
With 1.0.0-beta.54 the blue focus box went missing
![image](https://user-images.githubusercontent.com/44481292/116929706-a9ea6900-ac13-11eb-9fe8-35636dd9fe99.png)

asangma:
@driskull Started this but could use your help on the keyboard event handling See related PR.

jcfranco:
@AdelheidF This is installed. Can you help verify this one? 🙇🏼‍♂️

AdelheidF:
tested with beta.56

[...]

PR #2157, opened 2021-05-17
feat(action-menu): update styling and keyboard behavior

asangma:
**Related Issue:** #1906

## Summary
@driskull would you mind updating the keyboard handlers on this one? I took a whack at it, but realized I don't know what I'm doing.

### DIFF
--- a/src/components/calcite-action-menu/calcite-action-menu.scss
+++ b/src/components/calcite-action-menu/calcite-action-menu.scss
@@ -1,23 +1,23 @@
 :host {
-  @extend %component-host;
-  @extend %component-spacing;
-  display: flex;
-  flex-direction: column;
+  @apply flex
+  flex-col
+  box-border
+  bg-foreground-1
+  text-color-2
+  text-1;
 }
 
 ::slotted(calcite-action) {
-  display: flex;
-  width: 100%;
+  @apply flex w-full;
 }
 
 .menu-button {
-  align-self: stretch;
-  flex: 0 1 auto;
-  height: 100%;
-  position: relative;
+  @apply self-stretch
+  flex-initial
+  h-full
+  relative;
 }
 
 .menu {
-  min-width: var(--calcite-menu-min-width);
-  flex-flow: column nowrap;
+  @apply flex-col flex-no-wrap;
 }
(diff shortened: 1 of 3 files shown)


driskull:
@asangma I don't think we need to change the keyboard interaction at all. We just need to add a focus outline when an action is `active` on these.

driskull (commented):
Can we just add styling for slotted actions that are active to have a focus outline?

asangma:
@driskull 

> Can we just add styling for slotted actions that are active to have a focus outline?

Does this mean I should remove all the keyboard event handling?
Seems like we'd maybe want to trap focus inside the list of actions until either a selection is made or the escape key is hit?
Also, we'd probably want the menu to close when something is clicked yeah?

driskull:
https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-actions-active-descendant.html

asangma:
bump

src/components/calcite-action-menu/calcite-action-menu.tsx
```diff
@@ -218,6 +218,7 @@ export class CalciteActionMenu {
 
     return (
       <calcite-popover
+        disablePointer
```
driskull, line 221: Are we ok with this? This will remove the pointer on submenus in the action-bar opening to the right/left. This could be a noticeable difference for users.

driskull:
@asangma can you open an issue for this one:

> Also, we'd probably want the menu to close when something is clicked yeah?

driskull:
@asangma does this PR do this:

> With 1.0.0-beta.54 the blue focus box went missing

asangma:
Updated and ready for review.

src/components/calcite-action-menu/calcite-action-menu.tsx
asangma: I think it's good since we're emulating a drop-down.

driskull: cc @AdelheidF

driskull (approved):
👍

asangma:
Thanks @driskull.

src/components/calcite-action-menu/calcite-action-menu.scss
```diff
@@ -1,23 +1,30 @@
 :host {
-  @extend %component-host;
```
driskull, line 2: Is the host/spacing no longer needed here? i know it has some box sizing stuff.

asangma: I added the styles from %component-host via tailwind, so that should all be good.

PR #2160, opened 2021-05-18
fix(value-list-item): updated handle color 

asangma:
**Related Issue:** (#2098)

## Summary
Updated handle idle color.
cc @bstifle @julio8a


driskull (approved):
🌈

Issue #1497, opened 2021-02-01
Bug: Slider value should move on MouseDown instead of MouseUp


driskull:
### Summary

Bug: Slider value should move on MouseDown instead of MouseUp

### Actual Behavior

Slider value changes on mouse up. 🐁 

### Expected Behavior

Slider value changes on mouse down. 🐭 

### Reproduction Steps

1. View  a calcite-slider
2. Click on a spot on the range
3. Notice it won't change if you press and hold on a location.
4. The native input slider will change immediately.

driskull:
Also, if I press and drag and then unpress outside of the slider the mouse still moves the slider

![Uploading Feb-01-2021 11-06-33.gif…]()

driskull:
Another nice improvement would be to multiply the step value if the user is holding the shift key while using arrow keys.

driskull:
@eriklharper do you think you can address this one?

eriklharper:
My pleasure!

PR #2165, opened 2021-05-18
fix(calcite-slider): slider handle position changes on mousedown instead of click/mouseup

eriklharper:
**Related Issue:** #1497

### DIFF
--- a/src/components/calcite-slider/calcite-slider.e2e.ts
+++ b/src/components/calcite-slider/calcite-slider.e2e.ts
@@ -186,7 +186,7 @@ describe("calcite-slider", () => {
       expect(changeEvent).toHaveReceivedEventTimes(1);
     });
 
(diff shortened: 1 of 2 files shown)


jcfranco (approved):
Nice! 

I know dragging in puppeteer is a drag (badumtss :drum:), but is there a way to add a test for this?

driskull (approved):
Nice! Slider working well. This is a sweet UX improvement.

eriklharper:
> Nice!
> 
> I know dragging in puppeteer is a drag (badumtss 🥁), but is there a way to add a test for this?

Sure would be, yes.  I'll look into that next.  pptr.dev does have some docs on mousedown events, that could perhaps work.

eriklharper:
Added a bunch of tests for single and range values, both single-clicking the track and clicking and dragging the track for both.

PR #2166, opened 2021-05-18
refactor: remove unused host elements from components.

driskull:
**Related Issue:** N/A

## Summary

refactor: remove unused host elements from components.

Uses <Fragment> when necessary. Also runs prettier.


eriklharper (approved):

src/components/calcite-combobox/calcite-combobox.tsx
```diff
@@ -885,7 +885,7 @@ export class CalciteCombobox {
     const single = this.selectionMode === "single";
     const labelId = `${guid}-label`;
     return (
-      <Host>
+      <Fragment>
```
jcfranco, line 888: Should we keep `Host` when there's more than one child? Using `Fragment` doesn't reduce nesting and I'm not sure what we gain here.

driskull: Yeah, I guess we just protect from modifying the host.

jcfranco: Cool. We've got the new ESLint rules for that too, so I think the combo will help keep the host clean.

jcfranco (approved):
👍🏼

Issue #1957, opened 2021-04-13
Bug: Label - Remove use of document.getElementById()


driskull:
### Summary

Remove use of document.getElementById since it won't work if the component is within another shadow dom component. It seems like other teams are wrapping our components in their shadow dom components so we need to be mindful of this.

The solution is to use this.el.getRootNode() and then query either the rootNode (document) or rootNode.host (shadow dom) element.

### Actual Behavior

```jsx
// 2. htmlFor matches a calcite component
const inputForThisLabel = document.getElementById(this.for);
```

### Expected Behavior

```jsx
import { getElementById, getRootNode } from "../../utils/dom";

// 2. htmlFor matches a calcite component
    const rootNode = getRootNode(this.el);
    const inputForThisLabel = getElementById(rootNode, this.for);
```

### Reproduction Steps

1. Use a `calcite-label` inside the shadow dom of another component
2. Notice that the `for` property is not working as expected.

### Relevant Info 

https://github.com/Esri/calcite-components/pull/1956

jcfranco:
@eriklharper Was this closed intentionally?

eriklharper:
Yes, it was closed as per the linked PR, is that not correct?

jcfranco:
We shouldn't be closing via PR/commits, since we want someone to verify an issue is fixed before closing it and to set the appropriate labels. cc @caripizza

eriklharper:
Makes sense, I guess I did actually close it, since I assume its not getting closed automatically anymore on merged linked PRs?

PR #2167, opened 2021-05-19
fix(calcite-label): label for clicks work properly inside shadowRoots

eriklharper:
**Related Issue:** #1957

### DIFF
--- a/src/demos/calcite-label.html
+++ b/src/demos/calcite-label.html
@@ -17,6 +17,20 @@
 <script src="./_assets/head.js"></script>
   <script src="_assets/validateInput.js"></script>
   <script type="module" src="./_assets/demo-form.js"></script>
+  <script>
+    class ShadowComponent extends HTMLElement {
+      constructor() {
+        super();
+        const shadow = this.attachShadow({ mode: "open" });
+        shadow.innerHTML = `
+                <calcite-label for="input">Label and input inside a shadow component</calcite-label>
+                <calcite-input id="input"></calcite-input>
+              `;
+      }
+    }
+
+    customElements.define("shadow-component", ShadowComponent);
+  </script>
 </head>
 
 <body>
@@ -25,6 +39,8 @@
 
   <demo-form>
     <form name="calcite-label-demo-form">
+
+      <shadow-component></shadow-component>
 
       <calcite-label layout="inline" disable-spacing>
         <span style="margin-right: 0.75rem">Manual spacing for browsers that don't support flex gap</span>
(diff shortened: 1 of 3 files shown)


eriklharper:
Keeping this in draft while I whip up a test for this.

eriklharper:
For some reason I can't get this test to pass, so I'm pasting it here for reference just to keep a copy of the code.

```
it("clicking sibling label focuses input when both are inside a shadowRoot", async () => {
  await page.evaluate(() => {
    class ShadowComponent extends HTMLElement {
      constructor() {
        super();
        const shadow = this.attachShadow({ mode: "open" });
        shadow.innerHTML = `
          <calcite-label for="input">Label</calcite-label>
          <calcite-input id="input"></calcite-input>
        `;
      }
    }
    customElements.define("shadow-component", ShadowComponent);
    const shadowComponent = document.createElement("shadow-component");
    document.body.appendChild(shadowComponent);
  });
  const inputFocusEvent = await page.spyOnEvent("calciteInputFocus");
  const calciteLabel = await page.find("shadow-component >>> calcite-label");
  expect(inputFocusEvent).toHaveReceivedEventTimes(0);
  await calciteLabel.click();
  await page.waitForChanges();
  expect(inputFocusEvent).toHaveReceivedEventTimes(1);
});
```

driskull:
I'm not sure, but can you try running the click inside evaluate?

jcfranco:
Just checked this out. You need to initialize the page w/ a dummy calcite-component to bootstrap the components. Otherwise, they'll never upgrade. Similar to this: https://github.com/Esri/calcite-components/blob/55b6affe737d9c40fe907c16554eccf326604edc/src/components/calcite-date-picker-month-header/calcite-date-picker-month-header.e2e.ts#L35-L38

eriklharper:
I got the test working using the whole `window.eventDetail` trick that other tests do.  Also moving the click into the page.evaluate worked, but I think it would also work outside too, it was just the issue of listening to that event that was the biggest hurdle.  Should be good now!

src/demos/calcite-label.html
```diff
@@ -17,6 +17,20 @@
 <script src="./_assets/head.js"></script>
   <script src="_assets/validateInput.js"></script>
   <script type="module" src="./_assets/demo-form.js"></script>
+  <script>
+    class ShadowComponent extends HTMLElement {
```
jcfranco, line 21: I don't think we need this since it's covered by the unit test.

jcfranco (approved):
👍🏼

Issue #2171, opened 2021-05-19
Bug: tooltip flashes on mobile


isaacviel:
### Actual Behavior

On iOS a button label appears after clicking button 

### Expected Behavior

When clicking on this button, I should not see the hover/label 

### Reproduction Steps or Sample

On a mobile device/simulator:
1. Navigate to https://master.sites.afd.arcgis.com/pricing/tile-estimator/ (Currently on dev only, prod after 5/20)
2. Click the `<<` icon in the top right to collapse the sidebar
3. Click the filter icon to open the sidebar
4. Notice the `Collapse` hover label appear in the top right then jumps down the `<<` location then disappears. This label should not show it self at all on this tap let alone in the top corner then jump. 

![calcite](https://user-images.githubusercontent.com/3312529/118840231-0f795f00-b87c-11eb-8b38-4127378329b8.gif)

### Relevant Info

iOS simuator

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

driskull:
It seems like this is the tooltip on the expand/collapse. It appears to be showing delayed because the simulator is running very slow.

I'm not sure what we can do to address this.

isaacviel:
Thanks @driskull 

That's actually just the gif. The conversion from mov to gif really slowed it down which helps see the issue but misrepresents the speed. Even in real-time, the issue is quite obvious as a flash.

driskull:
I think the issue is that when clicking it, it focuses it which triggers the tooltip to show.

isaacviel:
Yeah, it seems like that's the issue for sure. Thanks @driskull

jcfranco:
@isaacviel This was installed last week. Can you help verify w/ `next`?

isaacviel:
You bet. Where exactly should I be testing?

jcfranco:
If possible, could you update the tile-estimator app to use the latest release?

isaacviel:
Thanks @jcfranco. We're hoping to be upgraded to the lasted version this week. I'll test as soon as we have an environment up.

cc @noahmulfinger

jcfranco:
Sounds good. I'll go ahead and close this as self-verified then since our sprint is done. Please submit a new issue if you find anything else to address.

jcfranco:
Closing as self-verified.

isaacviel:
@jcfranco this is still happening only the label is white now. 

Here's a test from our Calcite Components v55 preview site:


[...]

PR #2176, opened 2021-05-19
fix(tooltip-manager): Don't show tooltip on click of referenceElement

driskull:
**Related Issue:** #2171

## Summary

fix(tooltip-manager): Don't show tooltip on click of referenceElement. #2171

### DIFF
--- a/src/components/calcite-tooltip-manager/calcite-tooltip-manager.tsx
+++ b/src/components/calcite-tooltip-manager/calcite-tooltip-manager.tsx
@@ -155,6 +155,7 @@ export class CalciteTooltipManager {
     const tooltip = this.queryTooltip(event.target as HTMLElement);
 
     if (!tooltip || tooltip === this.clickedTooltip) {
+      this.clickedTooltip = null;
       return;
     }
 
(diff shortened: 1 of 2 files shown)


src/components/calcite-tooltip-manager/calcite-tooltip-manager.tsx
```diff
@@ -152,7 +154,8 @@ export class CalciteTooltipManager {
   focusEvent = (event: FocusEvent, value: boolean): void => {
     const tooltip = this.queryTooltip(event.target as HTMLElement);
 
-    if (!tooltip) {
+    if (!tooltip || tooltip === this.clickedTooltip) {
+      this.clickedTooltip = null;
```
jcfranco, line 158: Just curious if this can use `composedPath`, since it already gives you the elements the event traversed, instead of capturing and storing the reference?

jcfranco (approved):
Can you add a test for this? Otherwise, this LGTM!

src/components/calcite-tooltip-manager/calcite-tooltip-manager.tsx
driskull: I tried that but it doesn't give me the original event. It only gives me the focus event and I need to know if it came from a click.

jcfranco: I see. Never mind then. :)

PR #2182, opened 2021-05-20
docs(custom-theme): update to latest CSS color vars

caripizza:
**Related Issue:** n/a

## Summary
I noticed the custom-theme story shows some outdated CSS properties so this pr updates them according to global.scss.


driskull (approved):
👍 nice catch

Issue #1734, opened 2021-03-11
Bug: calcite-pick-list - keyboard user selection issue


AdelheidF:
### Summary
I'm a keyboard user only and I open a pick-list inside a popover with radio button selection, one selected item at a time. Let's say item 2 is initially selected. So I tab my way to it. Now I want to select the item number 5. I plan to use the arrow down key 3 times to go there and then select it with 'space'.  Same way it works for multiple selection. But as soon as I use the arrow key once the calciteListChange event fires, which takes the newly selected value and closes the popover in my workflow. How can I move 3 items down and then only select item number 5.

<img src=https://user-images.githubusercontent.com/44481292/110717400-d195f580-81bd-11eb-8dcf-43683a1b1bcd.png width=200/>

calcite-dropdown works differently. I can move up and down, only when I hit space does it execute.

cc @asangma 

### Reproduction Steps

1.

jcfranco:
@asangma WDYT? Splitting up keyboard navigation and selection would also make the experience consistent for both multiple and single selection modes.

asangma:
@jcfranco I agree. I think we were mimicking the select UX. But looking at it now, it doesn't quite work with pick-list.

jcfranco:
Installed! 

@AdelheidF Can you help verify this one w/ `next` (deploying soon)?

AdelheidF:
Cool, that works now.

A side note, it's kind of weird to me that I can use the down arrows on the title, but not on the (focused) icon at the end. When I use the down arrows on the icon the list just scrolls and my focus scrolls out of view eventually.

jcfranco:
@AdelheidF Thanks for verifying!

> A side note, it's kind of weird to me that I can use the down arrows on the title, but not on the (focused) icon at the end. When I use the down arrows on the icon the list just scrolls and my focus scrolls out of view eventually.

@asangma ☝🏼

AdelheidF:
> But as soon as I use the arrow key once the calciteListChange event fires

I do think this change here we made is correct and we need it (e.g for selecting a unique value in the filter component), but it seems that in some scenarios it is useful that a move with the arrow buttons also selects the new row and fires the change event (e.g. when trying out the different layer blending modes). 

Would it be possible to keep the current behavior as the default and then add a new property for the behavior to have move and selection together? Jeremy thinks it's a step backward for the blending scenario.

Could we possibly squeeze this in for the 9.2 release of Map Viewer?

test case: 
[...]

PR #2190, opened 2021-05-20
feat(pick-list, value-list): improve keyboard interaction for single-item selection

jcfranco:
**Related Issue:** #1734 

## Summary

This tweaks pick list's keyboard interaction to allow an item to be selected with the spacebar and allow items to be navigated via up/down arrows (before it would alter the selection when navigating through items).

This makes it consistent to when the pick list has the `multiple` mode enabled.

pick-list-keyboard


driskull (approved):
Looks good!

PR #2181, opened 2021-05-20
test(input-message): skip unstable test

caripizza:
**Related Issue:** n/a

## Summary
Passes locally fine but seems brittle in Travis CI, so skipping. (All these tests may go away anyways, due to #2172)

Here are few Travis build failures showing the issue:
 
[Travis fail 1](https://travis-ci.com/github/Esri/calcite-components/builds/226248957):
```
FAIL src/components/calcite-input-message/calcite-input-message.e2e.ts (28.516 s)
  ● calcite-input-message › CSS properties for light/dark themes › should allow the CSS custom property to be overridden
    expect(received).toEqual(expected) // deep equality
    Expected: "rgba(0, 0, 0, 0.4)"
    Received: "rgba(0, 0, 0, 0.094)"
      217 |       inputMessage = await page.find("calcite-input-message");
      218 |       inputMessageStyles = await inputMessage.getComputedStyle();
    > 219 |       expect(await inputMessageStyles.getPropertyValue("background-color")).toEqual(overrideStyle);
          |                                                                             ^
      220 |     });
      221 |   });
      222 | });
      at Object.<anonymous> (src/components/calcite-input-message/calcite-input-message.e2e.ts:219:77)
          at runMicrotasks (<anonymous>)
```

[Travis fail 2](https://travis-ci.com/github/Esri/calcite-components/builds/226260821):
```
FAIL src/components/calcite-input-message/calcite-input-message.e2e.ts (20.528 s)
  ● calcite-input-message › CSS properties for light/dark themes › should allow the CSS custom property to be overridden
    expect(received).toEqual(expected) // deep equality
    Expected: "rgba(0, 0, 0, 0.4)"
    Received: "rgba(0, 0, 0, 0.24)"
      217 |       inputMessage = await page.find("calcite-input-message");
      218 |       inputMessageStyles = await inputMessage.getComputedStyle();
    > 219 |       expect(await inputMessageStyles.getPropertyValue("background-color")).toEqual(overrideStyle);
          |                                                                             ^
      220 |     });
      221 |   });
      222 | });
      at Object.<anonymous> (src/components/calcite-input-message/calcite-input-message.e2e.ts:219:77):
          at runMicrotasks (<anonymous>)
```
[Travis fail 3](https://travis-ci.com/github/Esri/calcite-components/builds/226432895):
```
FAIL src/components/calcite-input-message/calcite-input-message.e2e.ts (23.557 s)
[...]

PR #2191, opened 2021-05-21
test(input-message): remove frail CSS var theme tests

caripizza:
**Related Issue:** n/a

## Summary
I added these tests as a part of #1737 but they're clearly unstable. Already skipped one yesterday in pr #2181. Since we're planning to [remove this "floating" variant](https://github.com/Esri/calcite-components/issues/2172) in the near future, the tests have very little value at this point (which is why I removed them entirely instead of skipping.)


driskull (approved):
👍

Issue #2189, opened 2021-05-20
Bug: deprecate usage of the non-standard theme attribute


eriklharper:
### Actual Behavior

```
<div theme="dark"><!-- insert dark theme components here --></div>
```

### Expected Behavior

```
<div class="calcite-theme--dark"><!-- insert dark theme components here --></div>
```

We should be using a CSS class to set the theme variables instead of the non-standard HTML attribute "theme".  We will probably also need to set a `--calcite-theme-name` or similarly named CSS variable so components know what the current theme is set to.

### Reproduction Steps or Sample

1.

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

@driskull

lpaniccia-crown:
If this moves to CSS, I'd love to have the option to match the user's system automatically with the prefers-color-scheme media query. Possibly with something like this: 
`<div class="calcite-theme--auto"><!-- insert components here. Theme will match the user's preferred color scheme.--></div>`

driskull:
yeah that would be possible as well.

macandcheese:
> If this moves to CSS, I'd love to have the option to match the user's system automatically with the prefers-color-scheme media query.`

+1, this would be a great addition !

eriklharper:
@driskull can this issue be closed now?

driskull:
Assigned to @caripizza . Don't want to verify my own issue.

PR #2194, opened 2021-05-21
refactor!: Remove 'theme' props from components

driskull:
**Related Issue:** #2189

## Summary

This PR is the first step of removing the use of the theme attribute selector.

- Removes all 'theme' properties from components
- Removes all use of the theme properties in tests
- Adds css class selectors for light/dark themes

The next steps would be to

- Replace theme attribute selector with classes
- Update stories to replace 'theme' properties with classes.
- Update test apps to replace 'theme' properties with classes.
- Document the theme classes somehow

Please review

### DIFF
--- a/src/utils/dom.ts
+++ b/src/utils/dom.ts
@@ -1,4 +1,5 @@
 import { Theme } from "../components/interfaces";
+import { CSS_UTILITY } from "./resources";
 
 export const themeNameCSSVariable = "--calcite-theme-name";
 
@@ -15,7 +16,7 @@ export function getAttributes(el: HTMLElement, blockList: string[]): Record<stri
 }
 
 export function getThemeName(el: HTMLElement): Theme {
-  return getElementProp(el, "theme", "light", true) as Theme;
+  return closestElementCrossShadowBoundary(`.${CSS_UTILITY.darkTheme}, [theme=dark]`, el) ? "dark" : "light";
 }
 
 export function getElementDir(el: HTMLElement): Direction {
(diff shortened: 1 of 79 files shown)


src/utils/dom.ts
```diff
@@ -10,6 +14,12 @@ export function getAttributes(el: HTMLElement, blockList: string[]): Record<stri
     .reduce((acc, { name, value }) => ({ ...acc, [name]: value }), {});
 }
 
+export function getThemeName(el: HTMLElement): Theme {
+  return getComputedStyle(el)
```
paulcpederson, line 18: this causes reflow of an element, which may cause reflow to elements around it. If you're calling this as part of a render loop it may be good to stress test it and do some memory auditing

paulcpederson:
nice! tons of work, thanks @driskull . I don't think we should be using `getComputedStyle` unless we absolutely have to, but other than that looks good to me

src/utils/dom.ts
driskull: Updated to use closest() and cross shadow.

driskull:
For now, closest will look for either the class or attribute. Once we remove all uses of the attribute we can update the utility to remove that part.

src/utils/dom.ts
```diff
@@ -1,3 +1,8 @@
+import { Theme } from "../components/interfaces";
+import { CSS_UTILITY } from "./resources";
+
+export const themeNameCSSVariable = "--calcite-theme-name";
```
paulcpederson, line 4: do you need this still?

driskull: No but I left it there incase there is a need. I can remove it.

paulcpederson:
I think you may be able to remove the theme name vars altogether? 

LGTM, though

paulcpederson (approved):

PR #2195, opened 2021-05-21
test: removes text directionality e2e commonTests, screenshot test functionality

caripizza:
**Related Issue:** #2108 (See [this comment](https://github.com/Esri/calcite-components/issues/2108#issuecomment-844266960) specifically for background)

## Summary
This is a follow-up to our verification of #1831, and removes:
-  `--screenshot` flag from npm test script (relying on Screener instead)
- pngs from screenshot/ folder
- text directionality tests that were added to commonTests (just testing browser defaults, new ESlint rule covers this check)

For rtl tests moving forward, I'm thinking we should just emulate this test and how it looks for the class if we really need to assert some RTL styles:
https://github.com/Esri/calcite-components/blob/master/src/components/calcite-action-pad/calcite-action-pad.e2e.ts#L117


driskull (approved):
👍

PR #2197, opened 2021-05-21
feat: Add auto theme class

driskull:
feat: Add auto theme class. #2189

Adds a class `calcite-theme-auto` to automatically use the theme color a user prefers. The `prefers-color-scheme` CSS media feature is used to detect if the user has requested a light or dark color theme.


src/assets/styles/global.scss
```diff
@@ -47,20 +34,39 @@
...
-  --calcite-alert-dismiss-progress-background: #{rgba($blk-000, 0.8)};
-  --calcite-button-transparent-hover: #{rgba($blk-240, 0.05)};
-  --calcite-button-transparent-press: #{rgba($blk-240, 0.08)};
-  --calcite-input-message-floating-background: #{rgba($blk-000, 0.96)};
-  --calcite-link-blue-underline: #{rgba($h-bb-070, 0.4)};
-  --calcite-scrim-background: #{rgba($blk-000, 0.75)};
+  @include calcite-theme-light-extended();
+}
+
+.calcite-theme--auto {
```
driskull, line 62: We could also call this `calcite-theme--preferred`?

src/assets/styles/global.scss
```diff
@@ -47,20 +34,39 @@
...
+
+  --calcite-border-radius: 3px;
+  --calcite-ui-opacity-disabled: 0.5;
+  --calcite-panel-width-multiplier: 1;
+
+  @apply font-sans;
+}
+
+.calcite-theme--dark,
+[theme="dark"] {
```
eriklharper, line 53: ☝️  Are we keeping this here for backwards compatibility so that this doesn't introduce a breaking change?

driskull: Yes. we are.

driskull:
Can one of the reviewers review this?`

src/assets/styles/global.scss
jcfranco: `auto` seems descriptive and concise enough to me.

jcfranco (approved):
Can you add a demo page snippet or test for this? Also, can you add brief summary to this PR's description? 

Other than that, this LGTM!

PR #2198, opened 2021-05-24
docs(changelog): add extra info about removal of `theme` prop and include note on removal of internal label from calcite-radio-button

jcfranco:
**Related Issue:** N/A

## Summary

Relevant breaking change doc updates: 

> - Remove 'theme' props from components (#2194)
>  The following CSS classes can be used to set the component's theme: `.calcite-theme--light`, `.calcite-theme--dark`. These are inherited, so they can be set on the topmost-themed element or on individual elements/components. This was done to avoid using the non-standard `theme` attribute/property on non-calcite elements.

> - **calcite-radio-button:** removing internal label (#2148)

The rest is part of the automated formatting on commit.


jcfranco:
Merging to have updated release doc available.

Issue #2041, opened 2021-04-26
Enhancement: Update Large component sizes


bstifle:
Update large component sizes use 16px font (font-size-0) and 16px icons

padding and spacing remain the same

### Affected components

- [x] Accordion
- [x] Avatar
- [x] Button (pr #2533) 
- [x] Checkbox
- [x] Chip
- [x] Combobox
- [x] Dropdown
- [x] FAB (pr #2533) 
- [x] Input
- [x] Inline Editable (pr #2533) 
- [x] Pagination
- [x] Radio Button
- [x] Radio Button Group
- [x] Radio Group (pr #2613)
- [x] Rating
- [x] Select
- [x] Split Button (pr #2533) 
- [x] Tab
- [x] Tree (pr #2803)

link to Figma designs
https://www.figma.com/file/E3SB0i5wPy7AagB7KcjF05/Calcite-Components

caripizza:
@bstifle Tabs doesn't have a large size yet (or `scale` property for that matter). It only has 1 size right now. Are you wanting to implement the S, M, L scale feature for Tabs from issue #803 here as well? Seems like that would also allow us to close issue #1610 too?

bstifle:
Correct!!!! I am speccing those out for you today

bpatterson88:
Taking on avatar

bpatterson88:
The avatar component renders at 40x40 in the codebase, but is 44x44 in Figma. Assuming the Figma version has the correct dimensions, should that be addressed as part of this issue?

caripizza:
Adding a note that I'm taking on the tabs portion of this issue as a part of 803

caripizza:
Taking on button from this list (from #2456)

bpatterson88:
Taking on input

julio8a:
Woah, nice Ben! That's a big one. Feel free to reach out if you need any help. @jcfranco was going to work on input this sprint, but go ahead and take it. Will re-assign the issue to Ben.

julio8a:
@justinhough will work on:
* Radio button
* ration button group
* rating

julio8a:
Looks like this one is complete, changing to installed.

julio8a:
@macandcheese, can you verify this one?

macandcheese:
Spot checking this looks good!

PR #2214, opened 2021-05-26
fix(calcite-accordion): updated large accordion to match figma

bpatterson88:
**Related Issue:** #2041

## Summary
- Updated large accordion header font size to 0 (16px)
- Removed logic for decorative header icon sizing so that all accordion sizes use small icons


bstifle:
@bpatterson88 looks good! i think the subtitles for all scales need to use the text-h lineheight. This may have been existing.... but can we shove this change in there too?

- small subtext should be text--2h
- medium subtext should be text--1h
- large subtext should be text--0h

bpatterson88:
@bstifle Fixed

bstifle (approved):
Lookin good!!!!

Issue #2130, opened 2021-05-12
Bug: calcite-popover - close button in weird position


AdelheidF:
It sounds like the current location of the close icon of a popover is by design. But it does look weird if the popover is a bit longer. 
If you still think the current behavior is how it should be, please close this issue...

This is a popover containing a div

![image](https://user-images.githubusercontent.com/44481292/117897152-ff0f2600-b276-11eb-89ca-96d0be884353.png)

This is a popover containing a panel

![image](https://user-images.githubusercontent.com/44481292/117896981-988a0800-b276-11eb-834b-54844b0f0b5f.png)

This is a popover containing a tile

<img src=https://user-images.githubusercontent.com/44481292/118016509-55797480-b30a-11eb-901a-02c5b23d3b58.png width=350/>

It would be nice if it were closer to the top-right like this design.

![image](https://user-images.githubusercontent.com/44481292/117897248-3b428680-b277-11eb-9750-bcfa7d101047.png)

To get this look currently I have to add a flow and a panel inside the popover, even for this small case here. E.g. similar to this one.

<img src=https://user-images.githubusercontent.com/44481292/118010858-4263a600-b304-11eb-9c0e-64065b5f670a.png width=450/>

### Reproduction Steps or Sample

1.

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

bstifle:
Oh weird, yeah that close button should always be in the right-top corner

asangma:
Thanks for the correction @bstifle.
Should take much to fix it.

bstifle:
ok so after talking with julio and adam, i think we did that deliberately........ so maybe its not a mistake. just cant remember when we did that with the X haha.

mitc7862:
@bstifle @julio8a @macandcheese 

This looks bad when the popover has more than one or two lines of text in it. Having an X in the corner of a dialog is a convention that's been around since GUI computers have existed… seems odd to decide to do it differently for this, doesn't _look_ very deliberate to me. 

Seems like the popover is used more like a dialog (modal, panel etc) and less like an alert or an in-line notice. Maybe it needs its own little header element to constrain the X, that way it can be consistent and not look bad when there's more text or images in the popover.

bstifle:
![image](https://user-images.githubusercontent.com/14083530/118186977-fa2bad00-b403-11eb-87ba-aac3c681f960.png)

u mean like this? we'd need to require a header at that point

[...]

PR #2221, opened 2021-05-26
feat(popover): adds heading prop and header node for when heading is supplied

asangma:
**Related Issue:** #2130

## Summary
Adds a `heading` property
Adds a `headingLevel` property (to optionally set heading level)

When `heading` is supplied:
- a **header** node is added
- **header** node contains a heading
- close button (if enabled) is "moved" to **header** node

image
Note: content of this example has inline styles since the default slot is unstyled.


asangma:
@bstifle just realized you dropped some designs in the issue. I'll make updates to match.

bstifle (approved):
looks good! 

would be nice to add a couple more examples! (with image no header) (with header and body copy, no image)

asangma:
Thanks @bstifle. Re: examples. You suggesting that for the demo app or Storybook?

PR #2227, opened 2021-05-27
refactor(label): L size text size down to 16px

julio8a:
**Related Issue:** #1500 and #2041 

- Large label text size down to 16px


bstifle (approved):
Large looks good! ✅🚢📦

PR #2229, opened 2021-05-28
fix(calcite-chip): updated scale to match figma

bpatterson88:
**Related Issue:** #2041

## Summary
- Changed font sizes to 12px for S, 14px for M, and 16px for L

### DIFF
--- a/src/components/calcite-chip/calcite-chip.scss
+++ b/src/components/calcite-chip/calcite-chip.scss
@@ -79,6 +79,8 @@
 
 :host button {
   @apply m-0
+    bg-transparent
+    focus-base
     inline-flex
     min-h-full
     max-h-full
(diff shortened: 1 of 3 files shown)


bstifle:
@bpatterson88 

Looks like small is good to go at 24px height

Medium and large are weirding me out.......

image
image

div is showind 42 and 30...

but visually, under a microscope it IS correct at 44px and 32px. so something's off?



 - Also, Large icons should  be 16, and X should be 16 per #2041

bpatterson88:
@bstifle L icon is now 16x16

I am seeing the correct sizes for S/M/L when I inspect `calcite-chip`. It is a little odd that the container inside isn't quite the same height, likely related to the borders rendering on `calcite-chip`.

Screen Shot 2021-05-28 at 9 45 27 AM
Screen Shot 2021-05-28 at 9 45 33 AM

bstifle:
Oh yeah you're right, just the container.

Looks good!

bstifle (approved):
Perfect, nice work man thank you!

PR #2246, opened 2021-06-01
chore: use node 14 for update-doc action

jcfranco:
**Related Issue:** N/A

## Summary

The Stencil build step in this action fails because the setup-node action was installing Node v10 (instead of v12+). This updates it to use v14.


jcfranco:
Merging for a test run today.

PR #2247, opened 2021-06-02
fix(rating): focus typo fix - outset

julio8a:
## Summary

A bug caused by the refactor. This brings back the focus animation.


jcfranco (approved):
LGTM! 

Since the commit messages get pulled into the changelog, before doing the squash merge, can you rephrase the commit message to convey what the fix does (e.g., restore outset focus)?

PR #2257, opened 2021-06-03
build(deps): bump versions

driskull:
build(deps): bump versions


driskull:
@jcfranco for some reason, this bump caused all the stories to disappear from storybook. Do you know how stories are added? Is there some config we need to update?

jcfranco:
@driskull Is there an error of some sort showing up? I don't see why they would disappear from this PR. There are a couple of dependencies that went through a major version bump, so maybe those could be causing issues. If that still fails, testing w/o bumping the storybook deps could shed some light.

driskull:
It seems the typescript bump to 4.3.2 is causing the problem. 

Maybe something here: https://devblogs.microsoft.com/typescript/announcing-typescript-4-3/#breaking-changes

I'll skip that bump for now.

jcfranco:
Sounds good. Thanks for looking at this! Can you add `typescript` as an exception to the [`deps:update` script](https://github.com/Esri/calcite-components/blob/master/package.json#L22)? It should probably be manually updated along w/ Stencil's other dependencies.

driskull:
I found this on storybook issues: https://github.com/storybookjs/storybook/issues/15067

Seems like a potential solution is to do this:
https://github.com/storybookjs/storybook/issues/15067#issuecomment-851959296


```
typescript: {
    reactDocgen: 'none',
}
```

Would we like to do this instead?

jcfranco:
Let's try if there's no fix by the time Stencil bumps to 4.3.2.

driskull:
Tried that fix and it didn't work.

driskull:
Can someone approve the bump?

jcfranco (approved):
👍🏼

PR #2239, opened 2021-05-28
feat(combobox): apply entered text on tab for custom values combobox

paulcpederson:
## Summary
 
We had people typing into a custom values combobox and tabbing away. Because they didn't "enter" the text (and create a chip) their value was thrown away, which was the source of a lot of confusion. One user said "that's weird it's not styled like a tag" and tabbed away anyways.

This change essentially adds a chip for them if they've entered text when they tab away.

PR #2263, opened 2021-06-04
fix(combobox):  treat entered text as chip in custom-value comboboxes on blur

paulcpederson:
## Summary

Related to the pr I opened the other day: #2239

Previously we used a tab key event to enter the text, this moves to using a blur event so that the text appears correctly if the user clicks away as well.


driskull (approved):
Does this need a test?

paulcpederson:
@driskull IMO this is covered by the test I wrote for the prior pr as it's the same feature  https://github.com/Esri/calcite-components/blob/master/src/components/calcite-combobox/calcite-combobox.e2e.ts#L411-L431

Issue #1113, opened 2020-10-12
Docs: how to listen for change events on tile-select


eriklharper:
### Description 

Since `tile-select` uses `checkbox` and `radio-button`, the way to listen for change events is to use their respective change events, but the documentation in the README lacks this info.  Let's add some docs above the auto generated line in the `tile-select` readme which explains this and provides a code sample.

### Which Component

`tile-select`/`tile-select-group`

macandcheese:
Still valid.

eriklharper:
Closing in favor of https://github.com/Esri/calcite-components/issues/2266

PR #2265, opened 2021-06-04
docs(calcite-tile-select): adding instructions for listening to change events

eriklharper:
**Related Issue:** #1113

### DIFF
--- a/src/components/calcite-tile-select-group/readme.md
+++ b/src/components/calcite-tile-select-group/readme.md
@@ -1,5 +1,37 @@
 # calcite-tile-select-group
 
+## How to listen for change events
+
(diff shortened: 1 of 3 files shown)


jcfranco:
I don't think we want to do this based on a previous discussion. If I recall correctly, we had decided to not expose the internal events and instead provide component-specific events.

eriklharper:
> I don't think we want to do this based on a previous discussion. If I recall correctly, we had decided to not expose the internal events and instead provide component-specific events.

I recall this discussion as well.  Do you know if an issue has been created to add those component-specific events?  I searched and couldn't find one.

jcfranco:
I couldn't find anything either, so I'm guessing we need one. Can you create an issue for this?

eriklharper:
Will do 👍

jcfranco:
Thanks! 🙇🏼‍♂️

Issue #2269, opened 2021-06-07
Bug: Calcite Flow not found in dist of calcite in JS API


previnWong:
### Actual Behavior

In the calcite distributable in node modules in JS API, calcite flow is missing.  See screenshot below
![image](https://user-images.githubusercontent.com/11183838/121062509-ff1d1b80-c779-11eb-9c2e-775ce65072b7.png)

### Expected Behavior

A js file should exists in this folder for FLOW.

### Reproduction Steps or Sample

1.  Look under this directory and Calcite Flow is missing.
<webroot>\arcgis-js-api\node_modules\@esri\calcite-components\dist\custom-elements\bundles

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [x] NPM package

jcfranco:
Installed. This will be available in `@next` and today's (later) release.

jcfranco:
Closing as verified based on https://unpkg.com/browse/@esri/calcite-components@1.0.0-next.219/dist/custom-elements/bundles/.

PR #2270, opened 2021-06-07
fix(custom-elements): add missing flow bundle

jcfranco:
**Related Issue:** #2269

## Summary

`calcite-panel` can be used independently, so this splits up the generated panel + flow bundle.


driskull (approved):
👍

Issue #2262, opened 2021-06-04
Docs: remove `theme` attribute from demo html and stories, use classes


caripizza:
### Description 
This is a follow on to #2189. Now that pr #2194 is in, and we are deprecating the usage of `theme="light"` and `theme="dark"` attributes for hooking into a theme, we need to update our resources:

1. remove theme attribute from demo html pages
2. remove theme attribute from component stories

Refactors should instead use the new `.calcite-theme-light` or `.calcite-theme-dark` classes for setting deliberate themes on components.

(Once we have these resources updated, we'll then add documentation on how to use the light/dark themes via the new classes.)

### Which Component
- All component demo html pages
- All component stories

### Resources 

cc @driskull

driskull:
For the stories, how do we want to proceed? Do we want a knob for the theme class? Or do we just want each story to have another "dark" theme story?

caripizza:
I vote actual dark theme story for Screener screenshot tests if we can (like the RTL ones)

caripizza:
Do you think it'd be good to keep the existing theme knobs in place too?

driskull:
I'm not sure about the knobs. We would have to use a class which is kind of an odd knob.

driskull:
I'm wondering if we should just remove both knobs for `theme` and `dir` and just have something to automatically generate `dark` and `RTL` stories for every story we create? This would give screener better coverage and be more automatic so that we only have to create unique stories. It would have to automatically export a const for dark/RTL

What do you think @jcfranco?

PR #2271, opened 2021-06-07
feat!: Use HTML classes instead of custom 'theme' attributes for theming components

driskull:
**Related Issue:** #2262 #2189

## Summary

feat!: Use HTML classes instead of custom 'theme' attributes for theming components #2262 #2189

BREAKING CHANGE: Using a 'theme' attribute is no longer supported to theme a component. Instead, a HTML class attribute of 'calcite-theme-light', 'calcite-theme-dark', or 'calcite-theme-auto' should be set on a component or one of its parent elements. Example: `<calcite-button theme="light" />` now becomes `<calcite-button class="calcite-theme-light" />`. The `calcite-theme-auto` class will defer to the browser's CSS 'prefers-color-scheme' media query to decide whether the light or dark theme will be used.


eriklharper (approved):

jcfranco:
@driskull Can you hold off on merging this until we release today?

jcfranco:
@driskull Actually, you can merge when ready. 

Also, the last release had a, premature, note on [breaking changes related to this](https://github.com/Esri/calcite-components/blob/master/CHANGELOG.md#-breaking-changes). Should we edit the changelog and remove that note in favor of this one?

driskull:
> BTW, the last release had a, premature, note on breaking changes related to this. Should we edit the changelog and remove that note in favor of this one?

Maybe we should leave both? The last one removed the props but the css selectors were still there working.

jcfranco:
Sounds good. Leaving for now. 👍🏼

PR #2272, opened 2021-06-07
fix(popover): add forceUpdate to ensure Action is available for setFocus

asangma:
**Related Issue:** #2130

## Summary
Fixes setFocus method of Popover.


driskull (approved):
👍

asangma:
Thanks for your halp @driskull!

Issue #2117, opened 2021-05-07
New Component: CalciteList & CalciteListItem - general purpose list and list-item


asangma:
### Description
A general purpose list that enables users to construct list items that conform to Calcite styling.
The list will also replace the currently used functional component and act as a base for several other lists, e.g.

* ValueList
* ValueListItem
* PickList
* PickListItem
* Proposed FlowMenu & FlowMenuItem

#### User Stories
Any developer for whom our pre-baked list-item doesn't work.

### Acceptance Criteria 
* Allow non-breaking refactoring of ValueList, ValueListItem, PickList and PickListItem.
* Act as a base for building the proposed FlowMenu and FlowMenuItem

### Questions
* Should it include `filterEnabled`?
* Should it include `dragEnabled`?

### Helpful Details
Things like selected states can be handled at the app level (or internally) by utilizing the "content-start" and/or "content-end" slots. See below.

**Item structure**
```
<calcite-list-item label="candy" description="kingdom">

  <calcite-action icon="some-icon" label="finn" slot="actions-start" />
  <calcite-icon icon="some-icon" slot="content-start"  />
  <img src="http://www.fillmurray.com/g/140/100" slot="content-start" />

  <calcite-icon icon="some-icon" slot="content-end" />  
  <calcite-action icon="some-icon" label="jake" slot="actions-end" />

</calcite-list-item>
```

RENDERS THIS LAYOUT

![image](https://user-images.githubusercontent.com/12503298/117511922-7412f080-af43-11eb-97fd-13185f2ff6ec.png)

### Additional thoughts on UI and scale
We may want to add a **scale** property and adjust internal spacing and font size according to our existing scales. Maybe this is a later enhancement as it will probably require additional UI design cycles.

At least for V1, we can mimic what we currently have in PickList and ValueList.

asangma:
@driskull Leaving the milestone up to you.

driskull:
Moving this to next milestone

asangma:
@julio8a @caripizza Could use this sooner rather than later. :)

julio8a:
@jcfranco and I will be triaging issues tomorrow.

driskull:
@asangma I'll start prioritizing it! Should start on it end of this week or early next week. Sorry for the delays.

asangma:
Yeet!

driskull:
@asangma about the "hit area". What should hitting the hit area do? Should each list item automatically have a hit area (button)? or should it depend on a property?

[...]

PR #2274, opened 2021-06-08
feat(list, list-item): Adds CalciteList & CalciteListItem, A general purpose list and list-item.

driskull:
**Related Issue:** #2117

## Summary

feat(list, list-item): Adds CalciteList & CalciteListItem, A general purpose list and list-item.

cc @asangma


driskull:
@asangma do you think you could start styling?

TheBlueDog:
Adding here per @macandcheese ... In Field Maps we need to use a value-list to list out all the fields on a layer, but we need to group them similar to pick-list. This is a MUST HAVE in order for us to convert our custom code to a component

asangma:
@driskull Styles are in.

@bstifle Let's review the UI of this thing.

driskull:
Ready for review

jcfranco:
@driskull Awesome! I'll make sure to review before the end of today.

bstifle (approved):
Lookin real nice love it, etc

src/components/calcite-list/calcite-list.e2e.ts
```diff
@@ -0,0 +1,21 @@
...
+      <calcite-icon icon="banana" slot="content-start" />
+      <img src="http://www.fillmurray.com/g/140/100" slot="content-start" />
+      <calcite-icon icon="banana" slot="content-end" />
+      <calcite-action icon="banana" label="jake" slot="actions-end" />
+    </calcite-list-item>
+    <calcite-list-item label="test" button-disabled description="hello world"></calcite-list-item>
+    <calcite-list-item label="test" description="hello world"></calcite-list-item>
+  </calcite-list>`);
+  });
+});
```
asangma, line 21: I will add a test to check that the content node is *not* added when no `title` or `description` is added.

asangma (approved):
yay

src/components/calcite-list-item/calcite-list-item.tsx
```diff
@@ -0,0 +1,152 @@
...
+        {content}
+      </button>
+    ) : (
+      <div class={CSS.contentContainer}>{content}</div>
+    );
+  }
+
+  render(): VNode {
+    return (
+      <div class={CSS.container}>
```
jcfranco, line 145: The list's host should have a role of `list`, no?

src/components/calcite-list-item/calcite-list-item.tsx
```diff
@@ -0,0 +1,152 @@
...
+  // --------------------------------------------------------------------------
+  //
+  //  Properties
+  //
+  // --------------------------------------------------------------------------
+
+  /**
+   * @todo document.
+   */
+  @Prop({ reflect: true }) buttonDisabled = false;
```
jcfranco, line 26: Is there another name to convey what this mode does? I think `button` is an implementation detail.

src/components/calcite-list-item/calcite-list-item.tsx
```diff
@@ -0,0 +1,152 @@
+import { Component, Element, Event, EventEmitter, Prop, h, VNode } from "@stencil/core";
+import { SLOTS, CSS } from "./resources";
+import { getSlotted } from "../../utils/dom";
+
+/**
+ * @slot actions-start - todo document.
```
jcfranco, line 6: JSDoc should be filled out before merging since we have automated doc-update PRs enabled.

src/components/calcite-list-item/calcite-list-item.tsx
```diff
@@ -0,0 +1,152 @@
...
+   */
+  @Prop() label: string;
+
+  // --------------------------------------------------------------------------
+  //
+  //  Events
+  //
+  // --------------------------------------------------------------------------
+
+  @Event() calciteListItemClick: EventEmitter;
```
jcfranco, line 49: Can you type the payload as `void`? Otherwise, the type and doc will show `any`.

src/components/calcite-list-item/readme.md
```diff
@@ -0,0 +1,3 @@
+# calcite-list-item
+
```
jcfranco, line 2: This will need some base info on the component too. 

I think we can drop the item's usage file since the list one would cover it.

src/components/calcite-block/calcite-block.tsx
```diff
@@ -135,13 +135,13 @@ export class CalciteBlock {
...
     ) : (
       <calcite-icon
-        icon={icon}
-        scale="m"
         class={{
           [CSS.statusIcon]: true,
           [CSS.valid]: status == "valid",
           [CSS.invalid]: status == "invalid"
         }}
+        icon={icon}
```
jcfranco, line 143: Optional: you can leave this as it was to keep the commit/PR focused and we can run linting across all components to avoid inconsistencies across PRs. Up to you.

src/components/calcite-list/calcite-list.tsx
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * A general purpose list that enables users to construct list items that conform to Calcite styling.
+ * @slot - todo document.
+ */
+@Component({
+  tag: "calcite-list",
+  styleUrl: "calcite-list.scss",
+  shadow: true
+})
+export class CalciteList {
```
jcfranco, line 13: Can we make the list manage tabbing among its items when navigating via keyboard? This would help with accessibility and help keyboard users easily tab through the list without having to go through all items.

jcfranco (changes requested):
This is looking really nice and clean! I had a few comments.

The requirements specify these components being used to refactor/simplify pick/value list + items, but didn't see any changes related to that. Did you get a chance to prototype how these would be used in the other list + items?

src/components/calcite-list/calcite-list.tsx
driskull: I dont think the list role would do that.

asangma:
> The requirements specify these components being used to refactor/simplify pick/value list + items, but didn't see any changes related to that.

If I understand the concern here, value-list-item and pick-list-item will be able to route content from "actions-end" and "actions-end" to the relevant slots in list-item.

Content stuff like the check or dot in pick-list-item will utilize list-item's "content-start" slot.

value-list-item will do similar content and action routing.

Hope that addresses the concern. 😬

src/components/calcite-list-item/calcite-list-item.tsx
driskull: We had discussed this in the issue. https://github.com/Esri/calcite-components/issues/2117

Do you have a better name in mind?

src/components/calcite-list/calcite-list.tsx
jcfranco: Right, this goes against my comment to use that role. I guess my suggestion is based on how close this and the other lists are in name and me assuming they would behave similarly. Clear doc and renaming pick/value list would help others avoid my assumption, right?

jcfranco:
@asangma It sure helps, but I was also asking if it has been prototyped.

src/components/calcite-list-item/calcite-list-item.tsx
jcfranco: I'd have to mull for a bit, but off the top of my head `interactive`?

asangma:
@jcfranco I can do dat.

src/components/calcite-list/calcite-list.tsx
driskull: Yes, I think we can evaluate pick-list and value-list at a later time. They should probably conform to more useful names that are common in UI libraries.

src/components/calcite-list-item/calcite-list-item.tsx
driskull: So `non-interactive` maybe? Because the default would be interactive unless we want to opt into it.

src/components/calcite-list/calcite-list.tsx
jcfranco: Sounds good. That should help clear things up for sure. Until then, clear doc would be the key.

asangma:
Initial prototype of pick-list-item looks good.
A refactor will cause a breaking change for PickListItem for sure.
ValueListItem initially seems like it might come through without any breaking changes.

asangma:
@driskull PickList and ValueList use the shared logic for selection, multi vs single select, etc. Should we add that logic as part of CalciteList?

driskull:
> @driskull PickList and ValueList use the shared logic for selection, multi vs single select, etc. Should we add that logic as part of CalciteList?

I don't think so, that would unnecessarily complicate the list.

asangma:
> I don't think so, that would unnecessarily complicate the list.

Copy that.

driskull:
@jcfranco ready for another review (minus the event removal discussion). :)

driskull:
@asangma do we need/want the disabled property for the list item?

jcfranco (approved):
Awesome! Thanks for working on this, @driskull!

jcfranco:
> Initial prototype of pick-list-item looks good.
> A refactor will cause a breaking change for PickListItem for sure.
> ValueListItem initially seems like it might come through without any breaking changes.

@asangma Thanks for prototyping. Can you write up your findings (especially the breaking changes in pick-list)? Let's see if we can refactor to avoid breaking changes.

src/components/calcite-list/calcite-list.e2e.ts
driskull: @asangma can you create a follow up issue for this? Going to merge PR.

driskull:
> PickList and ValueList use the shared logic for selection, multi vs single select, etc. Should we add that logic as part of CalciteList?

My assumption would be that ValueList/PickList internally will use list/list-item and further extend them by adding roles to them of listbox & option as well as props needed like 'selected', etc.

src/components/calcite-list/calcite-list.e2e.ts
driskull: Actually, i already see a test for this. So not needed?

PR #2276, opened 2021-06-08
fix(input): restore ability to tab backwards (shift+tab) out of number inputs

kstinson14:
**Related Issue:** N/A

## Summary
No issue for this one, but we noticed a bug in Field Maps where we weren't able to reverse tab out of calcite number inputs.  It seems like the restrictions on using "Shift" in number inputs was blocking the valid "Shift" + "Tab" interaction.


caripizza:
Thanks @kstinson14! I reproduced the bug on input's demo html page just now. Can you add a test for this fix?

caripizza (approved):
Thanks @kstinson14 👍

Issue #2230, opened 2021-05-28
A11y: [calcite-color-picker] - "ghost" focus ring when hue is set to 0 or 360


hccampos:
### Actual Behavior

When setting the hue to 0 or 360, the component ends up with a focus ring in the middle, as shown in the following image:
![Screen Shot 2021-05-28 at 14 44 15](https://user-images.githubusercontent.com/3759114/119993868-5974ea00-bfcc-11eb-93c2-2d6305a79253.jpg)

Could be related to https://github.com/Esri/calcite-components/issues/1297

### Expected Behavior

The focus ring should stay over the hue thumb.

### Reproduction Steps

1. Add a `<calcite-color-picker>` to an app.
2. Drag the hue thumb all the way to the left.
3. Tab into the component and notice that there is a ghost focus ring in the middle of the hue gradient.

hccampos:
@jcfranco this now works 👍  However, noticed a slightly different but related issue:
1. Initialize the `calcite-color-picker` with pure white.
2. Tab to the hue handle - notice you can't move the handle with the keyboard ⚠️  
4. Tab to the color handle and move to a different color - works 👍 
5. Tab back to the hue handle and try to move it with the keyboard - color gets reset to pure white.

This may be related to https://github.com/Esri/calcite-components/issues/1297

jcfranco:
Awesome, thanks for the breakdown! I see the above issues in the previous release, so I'll create a follow-up issue.

jcfranco:
Closing as verified based on the above.

jcfranco:
Created issue: https://github.com/Esri/calcite-components/issues/2357

PR #2277, opened 2021-06-08
fix(color-picker): handle hue scope when color is 000.

jcfranco:
**Related Issue:** #2230 

## Summary

Tweaks fallback logic for positioning the hue slider scopes.


driskull (approved):
👍

Issue #2223, opened 2021-05-27
Enhancement: Popover "close-button" property should be "dismissible"


asangma:
### Description
Popover uses the property `close-button` to turn on its close button. 
Notice and Panel use `dismissible`.

Suggest deprecating `close-button` on Popover and using `dismissible`.

### Acceptance Criteria 
Don't break stuff.

#### Which Component
Popover

#### Example Use Case
"I want to use Calcite Components and use the same prop to do the same thing in different components."

julio8a:
Verified in master

AdelheidF:
`setFocus("close-button")` still uses the old name. Is this expected?

https://github.com/Esri/calcite-components/blob/master/src/components/calcite-popover/calcite-popover.tsx#L228

driskull:
Yes, that wasn't changed.

AdelheidF:
How can I find out what focusIds are supported without looking at the code?

driskull:
Right now the code is the best way. The interface isn't being generated into doc. I think we would need to get stencil to fix it.

driskull:
Can you open a doc issue? We can probably just doc it in the method.

PR #2280, opened 2021-06-08
refactor(popover)!: Renames 'closeButton' property to be 'dismissible'. #2223

driskull:
**Related Issue:** #2223

## Summary

refactor(popover)!: Renames 'closeButton' property to be 'dismissible'. #2223

### DIFF
--- a/src/components/calcite-popover/calcite-popover.e2e.ts
+++ b/src/components/calcite-popover/calcite-popover.e2e.ts
@@ -66,7 +66,7 @@ describe("calcite-popover", () => {
         defaultValue: false
       },
       {
-        propertyName: "closeButton",
+        propertyName: "dismissible",
         defaultValue: false
       },
       {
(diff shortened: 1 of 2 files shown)


jcfranco (approved):
LGTM! Can you keep `close-button` and mark it as deprecated? We'll drop it at 1.0.0 or at a later release to drop renamed props in batch.

driskull:
@jcfranco I thought we were still allowing for breaking changes? Has that changed?

I figure we can start deprecating once we have released 1.0.

jcfranco:
Yes, we're still allowing for breaking changes, but if possible let's make it easier for users since we're releasing more frequently. I think we can skip this for anything that would introduce a lot of maintenance issues.

driskull:
Noted. Is it ok if this one goes in like this since it already has breaking changes and commit message reflecting it.

jcfranco:
If you push another commit, GH will use the PR's title for the squashed commit, so there's still a chance to amend.

PR #2281, opened 2021-06-08
refactor(calcite-combobox): scale L updates

bpatterson88:
**Related Issue:** #2041 

## Summary
- Updated padding to match Figma
- Changed icons to S when combobox scale is L
- Added some examples to the demo page


bpatterson88:
- Updated single select sizes to 24/32/44
- Updated select menus to better match size of inputs. Note that the select menu item heights aren't perfect (35px instead of 32px on M, 46px instead of 44px on L) because I'm using the wrapping line-height since the labels could potentially wrap.

caripizza (approved):
These changes look great to me based on Figma, and thanks for adding the additional demos @bpatterson88. 👍 
(I know you've been working with @bstifle so I'll make sure he has eyes on this too on Monday when he's back.)

macandcheese:
Chatted with @bpatterson88 on this - let’s make sure these items can line wrap without issue, that will happen often with long string values, especially combined with an icon.

bstifle:
this is looking really good ben!!!!

bstifle (approved):

bstifle:
Ok! looks like single select has incorrect left/right padding.

Small should be 8, Medium 12, large 16

image
image
image

bpatterson88:
@bstifle Updated the left/right padding

PR #2283, opened 2021-06-08
fix(combobox): fix custom chip selection on blur

paulcpederson:
## Summary

Fixing an issue with the combobox when you're using custom values, enter text, then select an item from the tags list. Also updated the test to behave more like an actual blur and used text that matches an existing option to cover this better.


driskull (approved):
👍

PR #2292, opened 2021-06-09
fix(combobox): emit lookup change when custom tag is added

paulcpederson:
## Summary

Combobox wasn't firing any event when custom values were added.


driskull (approved):
👍

Issue #2142, opened 2021-05-13
Enhancement: calcite-tile add a disabled state option


AdelheidF:
### Description

design
![image](https://user-images.githubusercontent.com/44481292/118170530-0bae8e00-b3df-11eb-9f0e-a77e5767ae4b.png)![image](https://user-images.githubusercontent.com/44481292/118170661-3567b500-b3df-11eb-9eb3-c4ae3cd55e57.png)

This is not possible currently.

AdelheidF:
Nice

![image](https://user-images.githubusercontent.com/44481292/122306570-888cc600-cebd-11eb-9ab5-6d1dc850ae4c.png)

PR #2298, opened 2021-06-09
feat(tile): add disabled property

jcfranco:
**Related Issue:** #2142 

## Summary

Adds `disabled` prop to prevent user interaction.


driskull (approved):
Looks good!

Do we need any changes for when a tile is both active and disabled?

jcfranco:
I don't think so since `active` only seems to affect the text color, but @macandcheese can confirm.

macandcheese:
Yep, that makes sense.

PR #2299, opened 2021-06-09
fix(tooltip-manager): Close hovered or focused tooltip when the reference element is clicked.

driskull:
**Related Issue:** N/A

## Summary

fix(tooltip-manager): Close hovered or focused tooltip when the reference element is clicked.

Closes a hovered or focused tooltip when its reference element is clicked. This gets the tooltip out of the way once an action is executed.

Jun-09-2021 16-46-38


driskull:
plz review

src/components/calcite-tooltip-manager/calcite-tooltip-manager.e2e.ts
```diff
@@ -321,4 +321,39 @@ describe("calcite-tooltip-manager", () => {
 
     expect(await hoverTip.getProperty("open")).toBe(true);
   });
+
+  it("should close hovered tooltip when clicked", async () => {
+    const page = await newE2EPage();
```
jcfranco, line 326: Can you pass the HTML as a parameter when the creating the page and use the `html` helper to automate formatting?

jcfranco (approved):
🎉

src/components/calcite-tooltip-manager/calcite-tooltip-manager.e2e.ts
jcfranco: Muchas thanks por updating las otras tests! 🥇

PR #2300, opened 2021-06-10
refactor(action-group, action-menu, panel): Adds a new slot 'trigger' to 'calcite-action-menu' to supply a custom trigger action.

driskull:
**Related Issue:** N/A

## Summary

refactor(action-group, action-menu, panel): Adds a new slot 'trigger' to 'calcite-action-menu' to supply a custom trigger action.

The action to open the calcite-action-menu can now be slotted and further customized.

This is consistent with the way that the 'calcite-dropdown' is authored and used.


jcfranco:
Is `calcite-action-menu` considered public? I don't see a story nor demo for it. If it's **not public**, I think we can make the change w/o considering it breaking. If it **is public**, can we allow the trigger be optional and have a default? This component makes it really quick and easy to set up by passing actions and forcing a trigger seems unnecessary if the `...` action is a good default for the trigger.

> This is consistent with the way that the 'calcite-dropdown' is authored and used.

The main difference I see here is that dropdown needs some text for the trigger. For action-menu, the default `...` seems to be an accepted pattern. cc @bstifle @macandcheese

driskull:
> Is calcite-action-menu considered public?

I'm not sure on this one.

> can we allow the trigger be optional and have a default?

I think that might be an odd experience to have the trigger sometimes be in shadow dom or not in shadow dom.

> This component makes it really quick and easy to set up by passing actions and forcing a trigger seems unnecessary if the ... action is a good default for the trigger.

The problem is that more and more the `...` action needs to be customized via size, icon, text, etc. So we would just end up having to pass all action props to it. So it works best as a slot.

jcfranco:
> I think that might be an odd experience to have the trigger sometimes be in shadow dom or not in shadow dom.

Can you elaborate on that? 

> The problem is that more and more the ... action needs to be customized via size, icon, text, etc. So we would just end up having to pass all action props to it. So it works best as a slot.

That is 100% a fair point. If the most common use case uses the default `...` trigger, having both would be the sweet spot, IMO.

driskull:
> Can you elaborate on that?

NVM. I misunderstood. So we check to see if a trigger is slotted else internally use a ... action?

jcfranco:
> So we check to see if a trigger is slotted else internally use a ... action?

I was thinking having the `...` action as the default content for the trigger slot, then you could listen for the calcite action click at a container level. Would this work?

driskull:
> Would this work?

I think so. if we can identify that it's the trigger action by its slot.

driskull:
Changed so its not a breaking change now.

jcfranco (approved):
Awesome! Thanks for making the non-breaking change. 🏆 🌮 🎉 

Sidebar, do you think the connecting/disconnecting logic is generic enough for a util? I think a few components have a similar setup, right?

driskull:
> do you think the connecting/disconnecting logic is generic enough for a util? I think a few components have a similar setup, right?

Yes that would be a nice util to have to make sure that we are properly setting up and removing events.

PR #2304, opened 2021-06-10
fix(action-menu): Focus on the menu's trigger when a slotted action is clicked and the menu is closed

driskull:
**Related Issue:** N/A

## Summary

fix(action-menu): Focus on the menu's trigger when a slotted action is clicked and the menu is closed.

When an action in the `calcite-action-menu` is clicked, the menu is closed but now nothing is focused. This restores focus to the button that opened the menu when an action has been clicked and the menu is now closed.


jcfranco (approved):
Noice! Can you add a test for this? I think the `focusable` test util can help.

Issue #2185, opened 2021-05-20
Bug: Allow updating tab programatically


paulcpederson:
### Actual Behavior

Setting `.active` on a tab title doesn't update the blue active underline position

### Expected Behavior

Setting active to a new tab should mark that tab as active and move the underline. 

### Reproduction Steps or Sample

https://codepen.io/paulcp/pen/WNpoJdW 

### Relevant Info

I think a watcher on the active prop of a tab should update location of underline.

_Version_: all versions

- [ ] CDN
- [ ] NPM package

caripizza:
I'm going to take a stab at this one alongside #803

julio8a:
verified on master. closing.

Issue #803, opened 2020-07-31
Enhancement: Add S M L Tabs and add new prop


bstifle:
1. Add SML sizes for tabs
2. ~Add `default` prop~ (update: this is not needed - Cari)
3. 2. ~Add `default-top` prop~ (update: this is not needed - Cari)
4. 2. Add `bordered` prop

![image](https://user-images.githubusercontent.com/14083530/89087390-f6cda600-d348-11ea-8b81-add4557c6a69.png)

macandcheese:
Related https://github.com/Esri/calcite-components/issues/186 cc @patrickarlt

macandcheese:
cc @rmstinson

capeoples:
+1 for this. I'm working on a project now with @joshtaylor0404 that could benefit from the medium scale size.

macandcheese:
Suggestion: change "minimal" to default, and choose a new name for what's shown as "default" above. Because that would be a breaking change, we should keep what's currently the design as default. Maybe "bordered" for the new style cc @bstifle

bstifle:
@macandcheese Agreed!! bordered makes sense

macandcheese:
Valid

julio8a:
This issue says in development, who is working on it?

macandcheese:
Not being worked on right now.

bstifle:
![image](https://user-images.githubusercontent.com/14083530/109027371-52190a00-7686-11eb-8192-37c3ce493627.png)

most recent designs and prop names

julio8a:
This is still needed

bstifle:
![image](https://user-images.githubusercontent.com/14083530/116442532-36fa8000-a818-11eb-9415-ac6c3e2daff1.png)
specs

bstifle:
![image](https://user-images.githubusercontent.com/14083530/116614544-ac8d4b80-a8ff-11eb-86a1-743078879ab2.png)

TheBlueDog:
Is there an update on this? Field Maps is in need of the sizes of tab to fix a UI defect.

caripizza:
Adding the latest mockups from Bryan:
![MicrosoftTeams-image (5)](https://user-images.githubusercontent.com/42423180/121594684-431a5580-c9f2-11eb-8372-f68813795363.png)
[...]

Issue #2177, opened 2021-05-19
Tooling: Set up Screener to test against demo pages


jcfranco:
## Summary 

With https://github.com/Esri/calcite-components/issues/2174, we'll want to include those pages into Screener's testing routine to improve testing of visual regressions.

We could use this as a first step towards https://github.com/Esri/calcite-components/issues/2019.

## Measure of Success 

Demo pages are included as part of the Screener screenshot test suite.

### Resources 

N/A

driskull:
From the little research I've done it seems like we would need the CI to run our local build and serve the demos so that the screener-runner can access the local URL and begin running visual tests.

I'm not sure we we can use the built in stencil server or if we need to copy the files and use something else? If we can use the stencil server, we need to be able to hook into when it is ready and in a separate process start running the visual tests.

Maybe the screener team can help us here.

eriklharper:
Here's the response I got back from Loyal that guides us through the steps to try to make this work:

> Hi Erik,

> Screener-runner is used under the hood of the screener-storybook module, so it should be able to do what you need.
> As you've run this locally, I'm assuming you used screener-runner's built-in tunnel option [github.com] - this can be used to > connect to a web server running in a CI build job as well.

> The process you've outlined at the end sounds like it should work:
> In your CI Build script, fire up your web server with demo pages. Let's assume this web server then becomes accessible in > > the CI job via http://localhost:3000 [localhost]
> Ensure your screener-runner config has the "tunnel" option set to host "localhost:3000"
> Then in your CI script run screener-runner using npm
> Hope that helps, thanks,

> Loyal

driskull:
Yep that makes sense.

I guess the part I'm unsure of is if the web server can just be the stencil build with the `--serve` argument. If so, what are the hooks we need to listen for to be able to start running the screener-runner.

eriklharper:
As I recall, this was a stumbling block I ran into as well.  I remember putting in a wait somewhere so that my local server was running properly before the runner started.

driskull:
Just playing devils advocate here... 😈 

Storybook-screener solves these issues already and has a method to use the steps API in screener. Do we really want to move off of that? Would it be easier to just update the stories to have a single story that then does all the steps to toggle sizing, theme, rtl, etc and take screenshots after each step?

[...]

PR #2306, opened 2021-06-10
feat(tabs): add `scale` and `bordered` properties, styles (WIP)

caripizza:
**Related Issue:** #803, #2185, #2041

## Summary
This pr adds SML scales to calcite-tabs, and a new `bordered` property per Bryan's mockups in 803. (When `calcite-tabs` has `bordered` on it and an "inline" `layout`, it now can appear like a folder menu with new colors/padding/borders. And, the blue active indicator swaps from top to bottom on the `active` tab-title.)

It also does the following:
- adds a usage example for `calcite-tab-nav` as a parent  (we provide this option in our developers docs currently, and I decided that rather than require folks to use `calcite-tabs` as a parent only, that `calcite-tab-title` can derive its scale and position values from a parent `calcite-tab-nav`.
- updates tab-title to use the rtl utility class, instead of just `.rtl`
- adds a watcher on `calcite-tab-title`'s `active` property to update the location of the blue active indicator (ie., 2185. See "Programatically update active tab" demo html example).


bstifle:
looking really really good! nice work cari

just these couple of spacing things:

image

bstifle:
AMAZING, TRULY.

bstifle (approved):

caripizza:
@driskull or @paulcpederson when you have a moment can one of you review this one? Thanks for your help!

src/components/calcite-color-picker/calcite-color-picker.scss
```diff
@@ -75,6 +76,10 @@ $gap--large: 12px;
 
   .color-hex-options {
     flex-shrink: 1;
+    min-height: 90px;
+    display: flex;
```
paulcpederson, line 80: I think we have tailwind-y classes for these guys

src/components/calcite-color-picker/calcite-color-picker.scss
```diff
@@ -141,6 +146,10 @@ $gap--large: 12px;
   }
 }
 
+.color-hex-options {
+  margin-bottom: theme("spacing.4");
```
paulcpederson, line 150: `@apply mb-4`?

src/components/calcite-tab/calcite-tab.e2e.ts
```diff
@@ -1,12 +1,65 @@
...
+
+  it("has block display when active", async () => {
+    const page = await newE2EPage({ html: "<calcite-tab active></calcite-tab>" });
+    const element = await page.find("calcite-tab");
+    const styles = await element.getComputedStyle();
+    expect(styles["display"]).toEqual("block");
+  });
+
+  describe("when nested within calcite-tabs component", () => {
+    it("should render with medium scale", async () => {
```
paulcpederson, line 31: I'm not entirely sure checking computed styles is necessary. IMO that's sort of what the visual regression testing is for.

src/components/calcite-tabs/calcite-tabs.scss
```diff
@@ -2,6 +2,39 @@
...
+:host([bordered][position="below"]) {
+  box-shadow: inset 0 1px 0 var(--calcite-ui-border-1), inset 0 -1px 0 var(--calcite-ui-border-1);
+}
+
+:host([bordered]) section {
+  border: 1px solid var(--calcite-ui-border-1);
+}
+
+:host([bordered][scale="s"]) section {
+  padding: 12px;
```
paulcpederson, line 27: should probably use the tailwind dudes for these as well

paulcpederson:
I think looks good, but maybe a pinch more tailwind to bring out a little more flavor.

src/components/calcite-tab/calcite-tab.e2e.ts
caripizza: @paulcpederson wouldn't we would a need a story for the tabs in each scale for that to work?

eriklharper: We could in theory accomplish this kind of testing with Screener Runner on demo pages, which would allow us to just render all the scales on a page that Screener can photograph and keep a record of.

caripizza: That would be awesome @eriklharper, you're referring to #2177 right?

eriklharper: For sure, yep!

Issue #2268, opened 2021-06-07
Bug: calcite-dropdown throws errors on open (bad items array)


hccampos:
### Actual Behavior

When opening a dropdown, an error is thrown:
```
utils.js:12293 Uncaught TypeError: Cannot read property 'isLink' of undefined
```
As a consequence, no item is focused and other things like the `max-items` attribute fail.

### Expected Behavior

No errors should be thrown when opening the dropdown. Furthermore, the list of items should be rebuilt when needed, rather than only when the component gets loaded.

### Reproduction Steps or Sample

Could not reproduce this in the Storybook demos but this is happening in one of our JS API widgets, using Maquette.

Digging in I noticed this happens because the `calcite-dropdown` component tries to access `item.attributes.isLink` but the item doesn't exist. This is because `componentDidLoad` runs before any of the calls to `registerCalciteDropdownGroup`, which means that when the dropdown is opened, the `items` array isn't correct.

### Relevant Info

- Happens when using Calcite Components through Maquette.js in an ArcGIS JS API widget.
- Issue exists in both `1.0.0-beta.55` and `1.0.0-next.215`.

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [x] NPM package

caripizza:
Related: #2026

jcfranco:
Installed.

jcfranco:
@hccampos @gavinr Can either of you help verify this one? FYI, this will be in today's release. Thanks!

jcfranco:
Closing as self-verified. Please submit a new issue if something else needs to be addressed.

Issue #2026, opened 2021-04-23
Bug:  `calcite-dropdown` error with Rollup


gavinr:
### Summary

Error `Uncaught TypeError: Cannot read property 'isLink' of undefined` when clicking on a `calcite-dropdown` when using with Rollup.

### Actual Behavior

Error:
![image](https://user-images.githubusercontent.com/209355/115819444-5ba4c100-a3ee-11eb-88a5-7fe3c3156c6e.png)

### Expected Behavior

No error.

### Reproduction Steps

1. Open https://gavinr.github.io/calcite-components-svelte-example/
2. In browser, click on the "Open Dropdown" dropdown.

OR:

1. Clone https://github.com/gavinr/calcite-components-svelte-example
2. `npm install`
3. `npm run dev`
4. In browser, click on the "Open Dropdown" dropdown.

### Relevant Info

- Error above is in Chrome. In Firefox, I get a similar but different error: `Uncaught TypeError: item.attributes is undefined`

- calcite-components v1.0.0-beta.54

Originally mentioned [here](https://github.com/Esri/calcite-components-examples/pull/7#issuecomment-666069317).

julio8a:
Thanks for opening this issue. We'll review and prioritize.

jcfranco:
Installed.

jcfranco:
Closing as self-verified. Please submit a new issue if something else needs to be addressed.

PR #2308, opened 2021-06-11
fix(dropdown): fix processing of groups and items

jcfranco:
**Related Issue:** #2268 #2026

## Summary

This refactors how items and groups are processed by the dropdown. It previously relied on events to register these elements, but these shouldn't be necessary because of [Stencil's lifecycle hierarchy](https://stenciljs.com/docs/component-lifecycle#lifecycle-hierarchy). 

The tests pass after refactoring and I also tested with the test case provided in #2026 (no more console error).

### DIFF
--- a/src/components/calcite-dropdown-group/calcite-dropdown-group.tsx
+++ b/src/components/calcite-dropdown-group/calcite-dropdown-group.tsx
@@ -10,7 +10,6 @@ import {
   VNode
 } from "@stencil/core";
 import { getElementDir, getElementProp } from "../../utils/dom";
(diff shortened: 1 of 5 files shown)


driskull (approved):
Nice cleanup!!

paulcpederson:
@jcfranco this is awesome. So basically parent components won't run until children are fully hydrated so you're able to just count on them being there from the top-most level?

jcfranco:
> So basically parent components won't run until children are fully hydrated so you're able to just count on them being there from the top-most level?

From what I understand, the Stencil lifecycle hooks will honor the hierarchy. `connectedCallback`/`disconnectedCallback` follow DOM rules and don't wait for components to load/initialize.

PR #2312, opened 2021-06-11
fix(combobox): in-progress custom values should get commited on blur or tab

paulcpederson:
## Summary

OK, last pr for this I swear. Was improving this test to cover both tab and click away blur and it revealed a bug where using tab key no longer entered the chip so I've addressed it here. 

FYI @driskull you were correct when you asked "Does this need a test?" in https://github.com/Esri/calcite-components/pull/2263 as the test only covered one of the two ways to tab away form the component.

Related prs: #2263 #2283


driskull (approved):
😄 👍

Issue #2232, opened 2021-05-28
Enhancement: Refactor BlockSection design and spacing


asangma:
### Description
Refactor BlockSection to use Label.

### Acceptance Criteria 
Don't break stuff.

### Relevant Info 
Current implementation allows switch and text to be misaligned.
![image](https://user-images.githubusercontent.com/12503298/120004292-6d4c2c00-bf8b-11eb-9d26-09dc73e89c13.png)

#### Which Component
BlockSection

#### Example Use Case
All of them.

jcfranco:
@asangma Do you have cycles to work on this for the next sprint? Otherwise, I can label this as 'help wanted'.

jcfranco:
The original misalignment is no longer present in the latest release (see https://codepen.io/jcfranco/pen/jOXreMj?editors=1100).

![Screenshot 2023-09-01 at 11 18 30 AM](https://github.com/Esri/calcite-design-system/assets/197440/9b091f17-5c09-4057-be40-28ad165242c5)

Closing based on the above.

PR #2317, opened 2021-06-14
refactor(block-section): use CalciteLabel for toggle header

asangma:
**Related Issue:** #2232

## Summary
Refactor to use our label component.
Includes a fix for Switch's vertical alignment issue.


driskull (approved):
👍

src/components/calcite-block-section/calcite-block-section.e2e.ts
```diff
@@ -131,12 +135,14 @@ describe("calcite-block-section", () => {
     expect(toggle.getAttribute("aria-label")).toBe(TEXT.expand);
 
     await toggle.click();
+    await page.waitForChanges();
```
asangma, line 138: @driskull Added this but am getting the error from the next check. Note that if I comment out that check, the check for "open" also fails.

Issue #2096, opened 2021-05-04
Bug: calcite-combobox:  long strings arent formatted correctly


kevindoshier:
### Actual Behavior

calcite-combobox:  long strings arent formatted correctly

![image](https://user-images.githubusercontent.com/11093161/117064249-c8fffe00-acda-11eb-8a4a-b1dc2c9dce05.png)

### Expected Behavior

Shouldnt mess up the formatting of the combobox

### Reproduction Steps or Sample

1.

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

driskull:
@julio8a who would this be for?

julio8a:
Does this look like this might be part of refactoring? @caripizza 

If it is, we should add it to that refactor epic.

driskull:
I think we should elevate this, mapviewer can't use this for layer selector until this is fixed. @julio8a @jcfranco @caripizza

jcfranco:
@driskull @kevindoshier Does this codepen represent the issue? https://codepen.io/jcfranco/pen/YzZOXNp?editors=1000

driskull:
@jcfranco yes but also when the selection mode is single

jcfranco:
Got it. Thanks for the info. Adding it to this sprint as `help wanted`. cc @caripizza

driskull:
I think a designer needs to take this issue since it's mostly a design decision on what to do and then updating the SCSS.

caripizza:
@kevindoshier @driskull Can you please add the code snippet that shows the issue? I'm having trouble replicating

kevindoshier:
Doesnt this one reproduce it?

https://github.com/Esri/calcite-components/issues/2096#issuecomment-858050015

caripizza:
Thanks @kevindoshier, I missed that comment. Can you add repro steps to the Description to help make it more clear? Or is this just a question of what it should look like when it's refactored to wrap?

kevindoshier:
Its just about needing it to be styled correctly

bstifle:
Couple of options for this:

![image](https://user-images.githubusercontent.com/14083530/122080636-09679700-cdc4-11eb-81ad-bbcb2943e58e.png)

caripizza:
@driskull - I chatted with @bstifle about this just now. What if we just truncate the selected item's text label with ellipses ("...") if it's over 25 characters long?

```js
[...]

PR #2319, opened 2021-06-15
fix(combobox): Truncate long strings correctly

driskull:
**Related Issue:** #2096

## Summary

fix(combobox): Truncate long strings correctly. #2096

### DIFF
--- a/src/components/calcite-combobox-item/calcite-combobox-item.scss
+++ b/src/components/calcite-combobox-item/calcite-combobox-item.scss
@@ -83,6 +83,7 @@ ul {
 }
 
 .title {
+  @apply truncate;
   padding: 0 var(--calcite-combobox-item-spacing-unit-s);
 }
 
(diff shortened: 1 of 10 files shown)


src/components/calcite-combobox-item/calcite-combobox-item.scss
```diff
@@ -83,6 +83,7 @@ ul {
 }
 
 .title {
+  @apply truncate;
```
caripizza, line 86: I don't think we should truncate the list items. I like that you can view the full text there, especially for mobile devices that can't rely on `title` attribute hover behavior. Can this line be removed? 

@bstifle do you agree or what should we do here? If it's removed, we get this:
image

Instead of this:
image

driskull: Fixed! makes sense.

caripizza (approved):
Looks great to me @driskull! 👍

src/components/calcite-combobox-item/calcite-combobox-item.scss
bstifle: image
this is more ideal!

bstifle:
Looks good guys! nice work

Issue #2259, opened 2021-06-03
A11y: No label for calcite-tile-select-group in screen reader


paulcpederson:
### Summary

When using a `calcite-tile-select-group` the text present in the tile is not read by the screen reader.

### Actual Behavior

Screen reader reads "radio button, 1 of 4"

### Expected Behavior

Screen reader should announce "TEXT OF TILE, radio button, 1 of 4"

### Reproduction Steps

1. Go to calcite select group story
2. turn on voiceOver

### Relevant Info 

This presents in Apple VoiceOver, but assume NVDA or JAWS would also behave poorly.

### Suggestion

Add a hidden label component connected to the radio button with the text of the tile's heading

driskull:
@paulcpederson Should the `calcite-label` be inside the component or should it wrap the `calcite-tile`? 

It would be nice if it used the `calcite-tile.heading`, however, it seems that the heading is optional and we would need it to be required if the tile has `input-enabled`.

driskull:
This can be fixed by just adding an `aria-label` on the `radio/switch`.

![image](https://user-images.githubusercontent.com/1231455/122126360-77bd5100-cde6-11eb-9a33-6277b607b8f3.png)

jcfranco:
Verified using https://esri.github.io/calcite-components/?path=/story/components-tiles-tile-select-group--light + VoicerOver.

PR #2320, opened 2021-06-15
fix(tile-select): Add aria-label for the form element used

driskull:
**Related Issue:** #2259

## Summary

fix(tile-select): Add aria-label for the form element used. #2259

Also adds guid for the input's id.

### DIFF
--- a/src/components/calcite-tile-select/calcite-tile-select.tsx
+++ b/src/components/calcite-tile-select/calcite-tile-select.tsx
@@ -71,12 +71,6 @@ export class CalciteTileSelect {
   /** specify the width of the tile, defaults to auto */
   @Prop({ reflect: true }) width: Extract<"auto" | "full", Width> = "auto";
 
-  @Watch("heading")
-  @Watch("name")
-  updateInputLabel(): void {
-    this.inputLabel = this.heading || this.name || "";
-  }
-
   //--------------------------------------------------------------------------
   //
   //  Private Properties
@@ -86,8 +80,6 @@ export class CalciteTileSelect {
   private input: HTMLCalciteCheckboxElement | HTMLCalciteRadioButtonElement;
 
   guid = `calcite-tile-select-${guid()}`;
-
-  inputLabel = "";
 
   //--------------------------------------------------------------------------
   //
@@ -184,7 +176,6 @@ export class CalciteTileSelect {
   //--------------------------------------------------------------------------
 
   connectedCallback(): void {
-    this.updateInputLabel();
     this.renderInput();
   }
 
@@ -206,7 +197,7 @@ export class CalciteTileSelect {
     this.input.disabled = this.disabled;
     this.input.hidden = this.hidden;
     this.input.id = this.guid;
-    this.input.label = this.inputLabel;
+    this.input.label = this.heading || this.name || "";
 
     if (this.name) {
       this.input.name = this.name;
--- a/src/components/calcite-tile-select/calcite-tile-select.tsx
+++ b/src/components/calcite-tile-select/calcite-tile-select.tsx
@@ -3,6 +3,7 @@ import { Alignment, Width } from "../interfaces";
 import { TileSelectType } from "./interfaces";
 import { getElementDir } from "../../utils/dom";
 import { CSS_UTILITY } from "../../utils/resources";
+import { guid } from "../../utils/guid";
 
 @Component({
   tag: "calcite-tile-select",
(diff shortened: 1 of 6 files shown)


src/components/calcite-tile-select/calcite-tile-select.tsx
```diff
@@ -193,13 +196,17 @@ export class CalciteTileSelect {
     this.input.checked = this.checked;
     this.input.disabled = this.disabled;
     this.input.hidden = this.hidden;
-    this.input.id = this.el.id;
```
driskull, line 196: @eriklharper do you know why this was being set to the element's id? All ids should be unique so I'm not sure why this was happening.

driskull:
While working on this PR, I was trying to figure out "How can I get the id of the input used within the component".

For example, If I have a `calcite-label` that is not surrounding a `calcite-input` and I set the `for` property on the `calcite-label`, how would I know what to set it to? It's not exposed anywhere right?

Do we want to support this:

```tsx
<calcite-label for="something">My Label</calcite-label> // the for here is invalid since the real input's id is not exposed.
<calcite-input id="something">Do Something?</calcite-input>
```

driskull:
cc @eriklharper ^

driskull:
We may need to have another prop on every input component that we can reference to know what the real ID of the input is going to be.

src/components/calcite-tile-select/calcite-tile-select.tsx
eriklharper: Definitely was a mistake to set that on this component originally.  IIRC it was to allow native label support so that a wrapping `<label>` would properly focus the input.

driskull:
I see that the `calcite-label` searches for the `id` using the `for` property to handle focusing, however since the id of the actual input doesn't match the `for` attribute, the screen readers can't link the two together.

eriklharper:
> Do we want to support this:

```
<calcite-label for="something">My Label</calcite-label>
// the for above is invalid since the real input's id is not exposed.
<calcite-input id="something">Do Something?</calcite-input>
```

My inclination is to say yes this should be supported, and I believe this currently is based on the logic in this function: https://github.com/Esri/calcite-components/blob/47433942db7655b65627b557881feb879df290b8/src/components/calcite-label/calcite-label.tsx#L82

eriklharper:
@driskull @jcfranco until browsers ship form-associated custom elements (which may very well never fully happen) we should probably just convert all our scoped components to shadow so that we can take full control over labeling behavior and screen reader behavior.

Seems like the experiment of making form components without shadow DOM probably has too many downsides than it is worth based on this and other issues we've seen with VDOM (I'm mainly thinking of the Maquette and Stencil bug where slotted content in a scoped component isn't relocated properly in some instances).

driskull:
> My inclination is to say yes this should be supported, and I believe this currently works based on the logic in this function:

It's not really functioning correctly because the native label's `for` is different than the native `input` id.

driskull:
> Seems like the experiment of making form components without shadow DOM probably has too many downsides than it is worth based on this and other issues we've seen with VDOM (I'm mainly thinking of the Maquette and Stencil bug where slotted content in a scoped component isn't relocated properly in some instances).

Yeah we will need to discuss this and make a call.

eriklharper:
Sounds good I'll put it on the agenda for Thursday.

eriklharper (approved):

driskull:
I do agree that it's odd that some form controls we have are scoped (input) and some are shadow (checkbox).

@eriklharper  Can't we just make them all shadow and have slots for the native inputs to be put into so that we can style them? It would also simplify some of the API since we currently have to duplicate props like 'disabled' and 'required'. Those could just be set on the slotted native element.

eriklharper:
> I do agree that it's odd that some form controls we have are scoped (input) and some are shadow (checkbox).
> 
> @eriklharper Can't we just make them all shadow and have slots for the native inputs to be put into so that we can style them? It would also simplify some of the API since we currently have to duplicate props like 'disabled' and 'required'. Those could just be set on the slotted native element.

Eventually I think we should convert all existing scoped to components to shadow, without supporting any slotted native inputs.  I think all the form components should just be able to be used standalone without any expectation the user would have to provide any other html elements.

driskull:
> I think all the form components should just be able to be used standalone without any expectation the user would have to provide any other html elements.

The downside to that is that we have to add props for things like disabled, aria-label, etc and just pass them to the internal form element. Its a bit redundant.

eriklharper (approved):
LGTM 👍

Issue #2005, opened 2021-04-21
Bug: calcite-input type=number - can enter weird numbers


AdelheidF:
### Summary

- calciteInputInput returns 000000, should probably not display this
![image](https://user-images.githubusercontent.com/44481292/115477956-eec8d580-a1f9-11eb-9bec-f394e57dd3a8.png)

- calciteInputInput returns valid numbers 1 and 5, but it should not allow a display like this
![image](https://user-images.githubusercontent.com/44481292/115478060-1f107400-a1fa-11eb-8c3c-4be73c4c8f2a.png)

- calciteInputInput returns 03
Start with a value of 0. Then type in a 3 after the 0. The input box correctly updates to just 3.
![image](https://user-images.githubusercontent.com/44481292/115478140-567f2080-a1fa-11eb-8f8e-775e37aa5cf8.png)

### Reproduction Steps

1.

eriklharper:
@AdelheidF @jcfranco Should we diverge from how the native `<input type="number">` element works by truncating `00000` to just `0`?  In a native number input, it allows multiple zeros, both in the display and the stored value at form submission time:

![image](https://user-images.githubusercontent.com/821864/122275598-be6b8380-ce98-11eb-946b-05087c94f3c8.png)

AdelheidF:
I don't see where this would ever be useful for an input field of type number.

AdelheidF:
Tested with the latest next.226 (from 11 days ago!)

I can no longer enter 1.... or 1,,,,, so that part is good.
Waiting on a newer next release to test 0000.

AdelheidF:
Tested now with next.227 and 0000 is good too.

PR #2325, opened 2021-06-16
fix(calcite-input): preventing number input from allowing more than one zero

eriklharper:
**Related Issue:** #2005

### DIFF
--- a/src/components/calcite-input/calcite-input.tsx
+++ b/src/components/calcite-input/calcite-input.tsx
@@ -500,6 +500,10 @@ export class CalciteInput {
     this.setValue(this.defaultValue, nativeEvent);
   };
 
+  private sanitizeNumberString(value: string): string {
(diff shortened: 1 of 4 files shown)


src/components/calcite-input/calcite-input.tsx
```diff
@@ -495,6 +500,10 @@ export class CalciteInput {
     this.setValue(this.defaultValue, nativeEvent);
   };
 
+  private sanitizeNumberString(value: string): string {
```
driskull, line 503: Should this go into a utility file somewhere? It seems like a useful utility.

driskull (approved):
👍

src/components/calcite-input/calcite-input.tsx
eriklharper: Just added it :)

Issue #2337, opened 2021-06-18
Enhancement: Allow for greater nesting of calcite-combobox-items within calcite-combobox


phenderson00:
### Description
Per Paul Pederson, there is currently a depth limit of 3 for nesting calcite-combobox-items within a calcite-combobox.  My use case requires a depth of 4, so while functionally everything is working correctly, visually it does not appear as it should.

### Acceptance Criteria 
Ability to nest calcite-combobox-items within a calcite-combobox to a depth of at least 4.

#### Which Component
calcite-combobox/calcite-combobox-item

#### Example Use Case
```
<calcite-combobox selection-mode="ancestors" placeholder="Select Map Theme">
    <calcite-combobox-item text-label="Biodiversity" value="">

        <calcite-combobox-item text-label="Species Biodiversity" value="SpBioRnkEco">

            <calcite-combobox-item text-label="Terrestrial Biodiversity" value="BioRankEco">
                <calcite-combobox-item text-label="Native Species Richness" value="NtvRankEco"></calcite-combobox-item>
                <calcite-combobox-item text-label="Rarity Richness" value="RarRankEco"></calcite-combobox-item>
                <calcite-combobox-item text-label="Irreplaceability" value="RwiRankEco"></calcite-combobox-item>
            </calcite-combobox-item>

            <calcite-combobox-item text-label="Aquatic Biodiversity" value="BioAqRankSW"></calcite-combobox-item>

        </calcite-combobox-item>

    </calcite-combobox-item>
</calcite-combobox>
```

Renders as shown in the screenshot below

![image](https://user-images.githubusercontent.com/48392427/122571811-bd5e6180-d012-11eb-8ac7-a124eadab647.png)

The expected output is that the three items with red arrows next to them would nested under Terrestrial Biodiversity (screenshot below).

![image](https://user-images.githubusercontent.com/48392427/122571984-ee3e9680-d012-11eb-9652-a7a1a72e49b3.png)

julio8a:
verified on beta.59

PR #2340, opened 2021-06-18
feat(combobox): Allow for greater nesting of calcite-combobox-items within calcite-combobox

driskull:
**Related Issue:** #2337

## Summary

feat(combobox); Allow for greater nesting of calcite-combobox-items within calcite-combobox #2337

### DIFF
--- a/src/components/calcite-combobox/utils.ts
+++ b/src/components/calcite-combobox/utils.ts
@@ -1,6 +1,7 @@
 import { nodeListToArray } from "../../utils/dom";
 import { ComboboxChildElement } from "./interfaces";
 import { ComboboxChildSelector } from "./resources";
+import { Build } from "@stencil/core";
 
 export function getAncestors(element: HTMLElement): ComboboxChildElement[] {
   const parent: ComboboxChildElement = element.parentElement?.closest(ComboboxChildSelector);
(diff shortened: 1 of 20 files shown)


src/components/calcite-combobox/utils.ts
```diff
@@ -24,12 +25,17 @@ export function hasActiveChildren(node: HTMLCalciteComboboxItemElement): boolean
...
-  if (!parent) {
+  if (!Build.isBrowser) {
     return 0;
   }
-  if (!grandparent) {
-    return 1;
-  }
-  return 2;
+
+  const result = document.evaluate(
```
paulcpederson, line 32: wow this is an advanced maneuver 🕶️ ⚡

paulcpederson (approved):

src/components/calcite-combobox/utils.ts
driskull: Is xpath use ok @jcfranco ?

jcfranco: TBH, I need a refresher on it. The last time I dealt w/ xpath was in the good ol' Flex/ActionScript days. 😂  

How does it deal in terms of performance and does it work across shadow DOM boundaries?

driskull: I'm not sure on the performance but it seems to work fine.

It shouldn't need to work across shadow for this use case but I think it does.

Issue #2060, opened 2021-04-27
Bug: [calcite-input] inconsistent value when setting ""


jcfranco:
### Actual Behavior

`<calcite-input>` behaves inconsistently depending on when value is set to "".

### Expected Behavior

When initially set to "", value will be null. Later on, if you set the value prop to "", value will be "" instead of null (as it was in the beginning).

### Reproduction Steps or Sample

https://codepen.io/jcfranco/pen/rNjRKPO?editors=1000

### Relevant Info

_Version_: `@esri/calcite-components@next.175`

- [x] CDN
- [ ] NPM package

eriklharper:
@jcfranco should a falsy value always be `""` or `null'?

jcfranco:
IMO, it should be `null` meaning there's no number set.

jcfranco:
Verified with issue test case.

PR #2342, opened 2021-06-18
fix(calcite-input): initially setting value to empty string results in value being "" instead of null

eriklharper:
**Related Issue:** #2060

### DIFF
--- a/src/components/calcite-color-picker/calcite-color-picker.e2e.ts
+++ b/src/components/calcite-color-picker/calcite-color-picker.e2e.ts
@@ -799,15 +799,15 @@ describe("calcite-color-picker", () => {
 
           await rgbModeButton.click();
 
-          expect(await rInput.getProperty("value")).toBeFalsy();
-          expect(await gInput.getProperty("value")).toBeFalsy();
-          expect(await bInput.getProperty("value")).toBeFalsy();
+          expect(await rInput.getProperty("value")).toBeUndefined();
+          expect(await gInput.getProperty("value")).toBeUndefined();
+          expect(await bInput.getProperty("value")).toBeUndefined();
 
           await hsvModeButton.click();
 
-          expect(await hInput.getProperty("value")).toBeFalsy();
-          expect(await sInput.getProperty("value")).toBeFalsy();
-          expect(await vInput.getProperty("value")).toBeFalsy();
+          expect(await hInput.getProperty("value")).toBeUndefined();
+          expect(await sInput.getProperty("value")).toBeUndefined();
+          expect(await vInput.getProperty("value")).toBeUndefined();
         });
 
         describe("clearing color via supporting inputs", () => {
@@ -839,10 +839,10 @@ describe("calcite-color-picker", () => {
             await clearAndEnterValue(page, rInput, "");
 
             // clearing one clears the rest
-            expect(await gInput.getProperty("value")).toBeFalsy();
-            expect(await bInput.getProperty("value")).toBeFalsy();
+            expect(await gInput.getProperty("value")).toBeUndefined();
+            expect(await bInput.getProperty("value")).toBeUndefined();
 
-            expect(await picker.getProperty("value")).toBeFalsy();
+            expect(await picker.getProperty("value")).toBeNull();
           });
 
           it("clears color via HSV channel inputs", async () => {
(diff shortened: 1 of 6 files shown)


src/components/calcite-input/calcite-input.tsx
```diff
@@ -176,15 +176,17 @@ export class CalciteInput {
     | "week" = "text";
 
   /** input value */
-  @Prop({ mutable: true, reflect: true }) value?: string = "";
+  @Prop({ mutable: true, reflect: true }) value?: string = null;
```
driskull, line 179: Does this prop need to be reflected?

I think it should be undefined by default. Null is an explicit value and the optional '?' does nothing here.

```
@Prop({ mutable: true }) value?: string;
```

driskull (changes requested):
Comments ☁️

src/components/calcite-input/calcite-input.tsx
```diff
@@ -176,7 +176,7 @@ export class CalciteInput {
     | "week" = "text";
 
   /** input value */
-  @Prop({ mutable: true, reflect: true }) value?: string = "";
+  @Prop({ mutable: true }) value: string;
```
driskull, line 179: nitpick: Is this optional or required?

`@Prop({ mutable: true }) value?: string;`

or

`@Prop({ mutable: true }) value!: string;`

driskull (approved):
👍 💯

src/components/calcite-input/calcite-input.tsx
eriklharper: Optional.

src/components/calcite-input/calcite-input.e2e.ts
```diff
@@ -670,7 +670,7 @@ describe("calcite-input", () => {
 
   it("allows restricting input length", async () => {
     const page = await newE2EPage({
-      html: `<calcite-input minLength="2" maxLength="3"></calcite-input>`
+      html: `<calcite-input minLength="2" maxLength="3" value=""></calcite-input>`
```
jcfranco, line 673: Is this needed for the test to pass? Or I should say, does the initial value affect the input length restriction?

src/components/calcite-color-picker/calcite-color-picker.e2e.ts
```diff
@@ -799,15 +799,15 @@ describe("calcite-color-picker", () => {
 
           await rgbModeButton.click();
 
-          expect(await rInput.getProperty("value")).toBeNull();
-          expect(await gInput.getProperty("value")).toBeNull();
-          expect(await bInput.getProperty("value")).toBeNull();
+          expect(await rInput.getProperty("value")).toBeFalsy();
```
jcfranco, line 802: Can you assert against the explicitly expected type? Currently, it would allow `false`, `0`, `''`, `null`, `undefined`, and `NaN`.

src/demos/calcite-input-number.html
```diff
@@ -144,5 +146,21 @@ <h2>Locales</h2>
...
       undefined;
     </script>
+
+    <script>
+      (async () => {
+        await customElements.whenDefined("calcite-input");
+        const input = document.querySelector("calcite-input");
+        await input.componentOnReady();
+
+        console.log(`when initially set to "", value is ${input.value} (${typeof input.value})`);
```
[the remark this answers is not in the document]
eriklharper: ah, thank you, forgot to do that.  Been working on this one too long...

src/components/calcite-input/calcite-input.e2e.ts
eriklharper: Yes, it is needed for the test to pass for some reason.  I have been struggling with getting that test to pass for days, and for some reason doing that makes it pass even though in all the browsers I tested this use case with it all worked as expected without having to set the initial value like that.

jcfranco: Sounds like regression based on that. Do you know exactly at which point the test fails?

eriklharper: It fails only when it is checking validity when the value is set to `1` with a `minlength` of 2.  It passes on the subsequent assertions.

PR #2345, opened 2021-06-21
feat(combobox): tab complete of custom values comboboxes

paulcpederson:
## Summary

Slight teak to keyboard handling for comboboxes. Tab will enter the current custom entry if available and leave focus on the combobox input. If there is no text tab acts normally.

/cc @capeoples

### DIFF
--- a/src/components/calcite-combobox/calcite-combobox.tsx
+++ b/src/components/calcite-combobox/calcite-combobox.tsx
@@ -146,9 +146,12 @@ export class CalciteCombobox {
       case "Tab":
         this.activeChipIndex = -1;
         this.activeItemIndex = -1;
-        this.active = false;
         if (this.allowCustomValues && this.text) {
           this.addCustomChip(this.text, true);
+          event.preventDefault();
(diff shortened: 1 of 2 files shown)


src/components/calcite-combobox/calcite-combobox.tsx
```diff
@@ -146,9 +146,12 @@ export class CalciteCombobox {
       case "Tab":
         this.activeChipIndex = -1;
         this.activeItemIndex = -1;
-        this.active = false;
         if (this.allowCustomValues && this.text) {
           this.addCustomChip(this.text, true);
+          event.preventDefault();
+          event.stopPropagation();
```
jcfranco, line 152: Can this be done w/o stopping propagation? For example, a dev may be listening for all global key presses and this would break that logic.

paulcpederson: mm, yes I think I only really need prevent default here, will update

jcfranco (approved):
👍🏼

PR #2354, opened 2021-06-22
fix(combobox): fix maxItems after selection, fix maxItems calculation when nested

paulcpederson:
## Summary

This solves a problem with the height being incorrect on the combobox in a couple different scenarios:

1. When items are nested in a particular way the amount of items shown was always one less than the prop
2. After entering a custom value, only two items where shown, regardless of maxItems setting


driskull (approved):
👍

PR #2363, opened 2021-06-22
fix(combobox): use normal pointer for input of single select combobox

paulcpederson:
## Summary

Text input for single select combobox was using a pointer cursor. This is inconsistent with all the other inputs, comboboxes, etc. This was probably a mistake when I added this mode.


driskull (approved):
👍

PR #2366, opened 2021-06-22
docs(value-list): Update group property description.

driskull:
docs(value-list): Update group property description.

### DIFF
--- a/src/components/calcite-value-list/calcite-value-list.tsx
+++ b/src/components/calcite-value-list/calcite-value-list.tsx
@@ -72,6 +72,8 @@ export class CalciteValueList<
   @Prop({ reflect: true }) filterPlaceholder: string;
 
   /**
+   * The list's group identifier.
+   *
    * To drag elements from one list into another, both lists must have the same group value.
    */
   @Prop() group?: string;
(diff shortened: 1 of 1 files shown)


src/components/calcite-value-list/calcite-value-list.tsx
```diff
@@ -72,9 +72,9 @@ export class CalciteValueList<
   @Prop({ reflect: true }) filterPlaceholder: string;
 
   /**
-   * If this is set and drag is enabled, items can be dropped between lists of the same group.
+   * To drag elements from one list into another, both lists must have the same group value.
```
jcfranco, line 75: I like your phrasing better. Can we also add some text to describe the prop? Something like:

```
/**
 * The list's group name. 
 * 
 *  To drag items between lists, both must be draggable and have the same group value.
 */
```

?

driskull: I think its more of an id than a name, right? does name imply that it will be rendered?

jcfranco: Either would work. I remember seeing "name" used more in the group prop context.

Issue #2316, opened 2021-06-14
A11y: calcite-slider sets a zero value when hitting Enter/space


RalucaNicola:
### Summary

We use the slider to set the opacity of a color. Hitting Enter/space when the thumb is selected sets the slider value to zero:
![Jun-14-2021 14-38-48](https://user-images.githubusercontent.com/8091967/121893469-4711e480-cd1e-11eb-87f5-cf55eb99fc51.gif)

### Actual Behavior
Currently when hitting Enter/space with the slider thumb on focus, the value of the slider is set to zero.

### Expected Behavior

Hitting Enter/space with the slider thumb on focus should keep the selected value in place.

### Reproduction Steps

1. Open the slider demo: https://esri.github.io/calcite-components/?path=/story/components-controls-slider--single-value
2. Set the focus on the slider thumb.
3. Hit Enter and notice that the thumb goes to zero.

### Relevant Info 
macos chrome

jcfranco:
Installed.

jcfranco:
@RalucaNicola You can use the following codepen after `@next` deploys (in ~10 mins) to verify. https://codepen.io/jcfranco/pen/rNyEYPp?editors=1000

RalucaNicola:
Tested with your app Franco and it looks good, Space and Enter now keep the thumb value.

PR #2373, opened 2021-06-23
fix(slider): prevent Enter/Space from setting value to 0

jcfranco:
**Related Issue:** #2316 

## Summary

This fixes an issue where pressing the Enter/Space key would fire a click event on the button element used for the thumb and would set the value to 0.


driskull (approved):
Prevent zeee default!

PR #2376, opened 2021-06-23
docs(alert): use correct slot names in usage ex, demo util

caripizza:
**Related Issue:** # n/a

## Summary
Updates outdated slot names for alert in its usage example and demo util.


driskull (approved):
👍

Issue #2364, opened 2021-06-22
Docs: calcite-popover should list supported focusId values for setFocus()


AdelheidF:
### Description 

`calcite-popover` supports `setFocus("close-button")`. It would be nice to document this somewhere for the component. 
There are probably other components with the same issue.

Side note: I would think "dismiss-button" would make more sense now.

### Which Component

calcite-popover

AdelheidF:
Is good now. I still think "dismiss-button" would be better.

PR #2379, opened 2021-06-23
docs(popover): List supported focusId values for setFocus()

driskull:
**Related Issue:** #2364

## Summary

docs(popover): List supported focusId values for setFocus(). #2364


jcfranco (approved):
💥

PR #2387, opened 2021-06-23
docs(action-pad, action-bar): add group layout prop to story

asangma:
**Related Issue:** #none

## Summary
Nobody asked for it, but here it is.
Added the grid layout capability to the relevant storybooks.
Also added general layout props/values to ActionBar and ActionPad.
Includes a couple fixes for weird UI stuffs.


jcfranco:
@asangma Thanks for submitting this! Based on the description, can you split the fixes into a separate commit? Otherwise, they won't show up under the docs commit type when the changelog is generated.

PR #2392, opened 2021-06-24
chore: Allow typescript to import markdown files.

driskull:
**Related Issue:** None

## Summary

chore: Allow typescript to import markdown files.

This mutes the error about importing markdown files in storybook files.

### DIFF
--- a/globals.d.ts
+++ b/globals.d.ts
@@ -1,0 +1,1 @@
+declare module "*.md";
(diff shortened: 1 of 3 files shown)


jcfranco (commented):
`*.stories.ts` use a different tsconfig. 

These are excluded from the main tsconfig, but I still see the highlighted errors: https://github.com/Esri/calcite-components/blob/5c28273d5b03740e92df82e0c8ab907ec735696a/tsconfig.json#L4

Issue #2144, opened 2021-05-13
Enhancement: calcite-notice - add option to change the close icon to a different action


AdelheidF:
### Description

From the story book
![image](https://user-images.githubusercontent.com/44481292/118171050-b58e1a80-b3df-11eb-86e9-e501e734ac37.png)

Design
![image](https://user-images.githubusercontent.com/44481292/118171495-203f5600-b3e0-11eb-98a5-4707893274ca.png)

macandcheese:
There was a similar request for alert - in my mind a good addition to both notice and alert : https://github.com/Esri/calcite-components/issues/2012#issuecomment-827792704

jcfranco:
@macandcheese Would users expect the info icon to close it?

AdelheidF:
The info icon opens a popover. It will never be closed in this use case.

macandcheese:
I don't think so... it would also likely be any Calcite UI Icon so somewhat up to the implementer?

jcfranco:
@AdelheidF Thanks for clarifying. 

@macandcheese Would there be a use case for a trailing icon next to the close button? If so, we could add an icon-end slot and still allow it to be dismissed if necessary.

macandcheese:
That was the ask in the issue linked above https://github.com/Esri/calcite-components/issues/2012#issuecomment-827792704 - slot makes sense for having a tooltip, if we can style the border to and limit size of slotted icon to match that of the appened "x"

jcfranco:
Awesome. I think that will work. To recap, this should:

* add an icon-end slot for `calcite-action`s
  * placed before the close icon
  * close icon can be independently toggled via `dismissed`
  * styling should match the designs from the comment you linked above.

Does the above seem correct? Did I miss anything?

macandcheese:
Seems good to me!

bstifle:
![Uploading shia best part.png…]()
do it

bstifle:
<img width="432" alt="shia best part" src="https://user-images.githubusercontent.com/14083530/120852403-c691fd80-c53f-11eb-9377-57fd2726923e.png">
do it

jcfranco:
Installed. 

@AdelheidF Can you help verify once you use the latest next or the next beta release (beta 58)? 🙇🏼‍♂️

AdelheidF:
I'm using next.231.

Do we really need these vertical lines?

Without adjusting the position of the icon...

[...]

PR #2393, opened 2021-06-24
fix(notice): fix setFocus

jcfranco:
**Related Issue:** N/A

## Summary

This came up during #2144 and fixes a bug that would prevent `setFocus` from working.


driskull (approved):
👍 🥇

PR #2394, opened 2021-06-24
feat(notice): allow slotting multiple actions

jcfranco:
**Related Issue:** #2144 

## Summary

Based on a recent discussion, this changes the action slot name to allow for multiple actions and simplifies styling.

notices

### DIFF
--- a/src/components/calcite-notice/calcite-notice.tsx
+++ b/src/components/calcite-notice/calcite-notice.tsx
@@ -26,7 +26,7 @@ import { CSS_UTILITY } from "../../utils/resources";
  * @slot title - Title of the notice (optional)
  * @slot message - Main text of the notice
  * @slot link - Optional action to take from the notice (undo, try again, link to page, etc.)
- * @slot action-end - Allows adding a `calcite-action` at the end of the notice. It is recommended to use 2 or less actions.
+ * @slot actions-end - Allows adding a `calcite-action` at the end of the notice. It is recommended to use 2 or less actions.
  */
 
 @Component({
--- a/src/components/calcite-notice/calcite-notice.tsx
+++ b/src/components/calcite-notice/calcite-notice.tsx
@@ -26,7 +26,7 @@ import { CSS_UTILITY } from "../../utils/resources";
  * @slot title - Title of the notice (optional)
  * @slot message - Main text of the notice
  * @slot link - Optional action to take from the notice (undo, try again, link to page, etc.)
- * @slot action-end - Allows adding a `calcite-action` at the end of the notice
+ * @slot action-end - Allows adding a `calcite-action` at the end of the notice. It is recommended to use 2 or less actions.
  */
 
 @Component({
@@ -105,7 +105,7 @@ export class CalciteNotice {
       </button>
     );
 
-    const hasActionEnd = getSlotted(el, SLOTS.actionEnd);
+    const hasActionEnd = getSlotted(el, SLOTS.actionsEnd);
 
     return (
       <div class={{ [CSS.container]: true, [CSS_UTILITY.rtl]: dir === "rtl" }}>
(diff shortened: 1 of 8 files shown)


jcfranco:
As long as the slotted actions have a matching scale, their width comes very close to the dismiss button. 

@macandcheese @bstifle WDYT about the slight width difference? Is it triggering?

207qh0

Would tweaking the dismiss button area be an option?

src/components/calcite-notice/calcite-notice.tsx
```diff
@@ -26,7 +26,7 @@ import { CSS_UTILITY } from "../../utils/resources";
  * @slot title - Title of the notice (optional)
  * @slot message - Main text of the notice
  * @slot link - Optional action to take from the notice (undo, try again, link to page, etc.)
- * @slot action-end - Allows adding a `calcite-action` at the end of the notice
+ * @slot action-end - Allows adding a `calcite-action` at the end of the notice. It is recommended to use 2 or less actions.
```
driskull, line 29: Doesn't this need to change to `actions-end`?

driskull: Maybe we need to think about checking generated doc in tests? :)

driskull (changes requested):
Wouldn't this be a breaking change if the slot name has changed?

src/components/calcite-notice/calcite-notice.tsx
jcfranco: Thanks for catching that. 😅 

I'm treating this as a new feature since this change was omitted from changelog and the slot's name wasn't matching either. 

> Maybe we need to think about checking generated doc in tests? :)

That'd be tricky, but definitely a good idea. I'll create a research issue for this.

driskull (approved):
👍

src/components/calcite-notice/calcite-notice.tsx
driskull: If we build stories from the json-doc then this would be caught there most likely.

bstifle:
@jcfranco def super triggering!

let's get them to the width of the X :)

Large and medium should use 24px icons only (M)

jcfranco:
> let's get them to the width of the X :)

Right now, calcite-action will fill up the height if given an explicit one, but not the height. Is it OK to update it to fill the available space if given an explicit width/height? 

### scales + 100x100 class
Screen Shot 2021-06-29 at 8 55 15 AM

If this doesn't align with the design system, the easiest course of action would be to tweak the close button's size.

> Large and medium should use 24px icons only

I can update the demo page, but we can't really force the size since they're slotted/provided by users.

bstifle:
>Right now, calcite-action will fill up the height if given an explicit one, but not the height

assuming width there 😉, but yeah can we make those match the x?

image

jcfranco:
> assuming width there 😉, but yeah can we make those match the x?

lol correct. Updated.

OK, just for clarity, you're suggesting to update calcite-action's scale spacing to match the above design. Correct?

bstifle:
Correct! thanks man

jcfranco:
notice-sizing

I was able to implement the above design, but had to change some additional things:

* `calcite-notice`'s close button was using sizing vars that gave it a width of 72 (for large) – **changed it to be 64**.
* `calcite-action` was forcing the icon's scale in the following way:
  * when given "l" => forced it to "m" – **no change**
  * when given "m" => forced it to "s" – **changed to honor "m"**
  * honors "s" – **no change**

Are the above changes correct? `calcite-action`'s icon sizing would have the biggest impact.

> I can update the demo page, but we can't really force the size since they're slotted/provided by users.

I take this back. Fortunately, calcite-action can normalize the scale used for the icon. 🏆

bstifle:
@jcfranco yep 24px (M) for Large and Medium notice/alerts

16px (S) for Small notice/alerts

bstifle:
@jcfranco lets get the hover state of the X to match action. the icon should go from ui-text-3 to ui-text-1 on hover

jcfranco:
It wasn't clear to me from the previous comment, but are we OK with changing the way that calcite-action maps scale to its internal icon? This change would affect anything using this component and not only notice/alert.

jcfranco:
@bstifle Updating per our conversation:

* I'll revert the calcite-action changes in favor of updating the sizes of internal icons used by notice/alert + scale=m (https://github.com/Esri/calcite-components/issues/2417).
* Will update demo page to include all sizes for this feature.

Issue #2284, opened 2021-06-09
Docs: Documentation issues at beta 56


kellyhutchins:
### Description 
I've noticed a few places where the doc needs to be updated to reflect the changes at beta 55/56

- [x]  Sample code on Alert page uses the old slot names (alert-message, alert-link)

https://github.com/Esri/calcite-components/blob/v1.0.0-beta.56/src/components/calcite-alert/readme.md

- [x] In calcite-panel the code sample shows a slot with header-leading-content and header-trailing-content but I don't see those doc'd in the slot section https://github.com/Esri/calcite-components/tree/v1.0.0-beta.56/src/components/calcite-panel

jcfranco:
Installed! 

@kellyhutchins Thanks for reporting this. Can you help verify this one? 🙇🏼‍♂️

kellyhutchins:
@jcfranco  the alert one looks good but I still see "header-trailing-content" and "header-leading-content" here: 
https://github.com/Esri/calcite-components/tree/master/src/components/calcite-panel

jcfranco:
@kellyhutchins I updated the usage files, but the readme's haven't been updated yet. We have a GH action that submits a PR for readme updates every day at 5:45p. The readmes should be up to date by tonight. Sorry for the hassle!

kellyhutchins:
@jcfranco  got it thanks!

kellyhutchins:
Looks good!

PR #2396, opened 2021-06-25
docs(panel): use proper slot names for leading/trailing actions

jcfranco:
**Related Issue:** #2284

## Summary

Fixes outdated slot names used in panel usage examples.


driskull (approved):
👏

Issue #2344, opened 2021-06-21
calcite-pick-list - single selection with 2 change event options


AdelheidF:
### Description

Read https://github.com/Esri/calcite-components/issues/1734 first...

> But as soon as I use the arrow key once the calciteListChange event fires

I do think this change here we made is correct and we need it (e.g for selecting a unique value in the filter component), but it seems that in some scenarios it is useful that a move with the arrow buttons also selects the new row and fires the change event (e.g. when trying out the different layer blending modes). 

Would it be possible to keep the current behavior as the default and then add a new property for the behavior to have move and selection together? Jeremy thinks it's a step backward for the blending scenario.

Could we possibly squeeze this in for the 9.2 release of Map Viewer?

test case: 
Click layer to select it, go to Properties panel and look for Blending and open dropdown.

requires arrow + space key
https://devext.arcgis.com/apps/mapviewer/index.html?url=https://services1.arcgis.com/4yjifSiIG17X0gW4/arcgis/rest/services/FatalAccidents2017/FeatureServer/0

only requires arrow key
https://www.arcgis.com/apps/mapviewer/index.html?url=https://services1.arcgis.com/4yjifSiIG17X0gW4/arcgis/rest/services/FatalAccidents2017/FeatureServer/0

jbartley:
@jcfranco @asangma @julio8a can you comment on this request?  It is a change in behavior for how the new map viewer allows users to arrow down through different blend modes.  

Here is how it works on production:
![caclcite list keyboard navigation production](https://user-images.githubusercontent.com/2277787/123301482-8c79a300-d4d0-11eb-98fa-eb7ca3b926ad.gif)

Here is how it works on dev:
![caclcite list keyboard navigation](https://user-images.githubusercontent.com/2277787/123301520-94d1de00-d4d0-11eb-9401-d45fb60911f1.gif)

I have a feeling people will complain on this as it was quite handy to cycle through the list with arrow keys.  Now it is arrow + spacebar or explicit mouse clicks.  

I understand that multi-select requires this combination of arrow keys + space bar to be accessible.  For single selection maybe we can have the arrow key also trigger a selection?

jcfranco:
From an implementation-perspective, totally doable. I think an opt-in prop would work well. The name would just have to indicate that it only applies for single-select.

From the UX-side of things, it makes sense to me to have these two modes, but I'll defer to @asangma's UXpertise™.

asangma:
This make sense to me too.
[...]

PR #2401, opened 2021-06-25
feat(pick-list, value-list): add prop to enable changing selection via keyboard in single-selection mode

jcfranco:
**Related Issue:** #2344 

## Summary

This adds a `selectionFollowsFocus` (default false) prop to both pick and value lists to allow users to change the selected item as they navigate items with the ⬆  ⬇️  keys.


driskull (approved):
👍

jcfranco:
I'm having some test failures related to changing the list's role to `listbox` that need further investigation. I'll revert the role change and create a separate issue.

Issue #2234, opened 2021-05-28
Enhancement: Option to make Modal closable by clicking outside the dialog


andreasulmer:
### Description

Most modal web dialogs have the option to be closed by clicking outside the dialog, with the same behavior as clicking ✖, or pressing Esc. Calcite Modal does not provide this functionality.
It would ideally be configurable, the boolean could be called `disable-outside-close` (similar to `disable-escape`).

### Acceptance Criteria 
Dialog closes when user clicks (or touches) outside the dialog.

#### Which Component
Modal

#### Example Use Case
No specific use case, but following common web UX patterns.

bstifle:
**100% yes**

andreasulmer:
Tested locally w/ npm @1.0.0-next.237
Tested on Win 11 Chrome, Firefox, Edge, Chrome w/ mobile touch emulation, Android Chrome.

✔ by default, modal now closes on outside click
✔ `<calcite-modal disable-outside-close...>` disables the outside close
✔ behaves as expected on drag-clicks, reliably triggers close on outside mouse-up
❌ One minor inconsistency with drag-release: 
  * Mouse (non-touch) device: click-drag-release_in_outside_area triggers close ✔
  * Touch device: tapAndHold-drag-release_in_outside_area does not trigger close ❌
-> Sloppy taps prevent modal from being closed

IMHO can still be closed as done if ❌ is considered too minor.

PR #2409, opened 2021-06-28
feat(modal): Closes modal when the outside is clicked.

driskull:
**Related Issue:** #2234

## Summary

feat(modal): Closes modal when the outside is clicked and adds property disableOutsideClose to disable it. #2234


jcfranco (approved):
Code LGTM!

This does change the default behavior to now close on outside clicks. @bstifle Can you confirm this is OK? This isn't specified in the original issue.

jcfranco:
> @bstifle Can you confirm this is OK? This isn't specified in the original issue.

Actually, I was wrong, the proposed boolean in the original issue does hint at changing the default behavior. I guess there wasn't much discussion on that, so just wanted to confirm.

driskull:
It seems like the preferred default would be to close on click. Ideally, it should do that, correct?

bstifle:
by default yes @driskull

bstifle:
@jcfranco  yep close when clicked outside by default!

Issue #1579, opened 2021-02-19
refactor: Remove getAttributes() utility helper and no longer spread attributes.


driskull:
### Description

Remove getAttributes() utility helper and no longer spread attributes. #1579

### Acceptance Criteria 

We only set defined properties on internal elements

#### Which Component

All components using getAttributes() helper utility:

- [x] `calcite-button` #2410
- [x] `calcite-dropdown-item` #2427
- [x] `calcite-input` #2443 
- [x] `calcite-label` #2503 
- [x] `calcite-link` #2509

PR info: https://github.com/Esri/calcite-components/pull/1567

Related to #1538

driskull:
This isn't pre 1.0 but is something we want to refactor soon since it allows a user to affect the shadow dom from the host element.

eriklharper:
Resolved by #2410, #2427, #2443, #2503, #2509

PR #2410, opened 2021-06-28
refactor(calcite-button)!: stop spreading attributes  BREAKING CHANGE: to set an aria-label on the button use the label attribute

eriklharper:
**Related Issue:** #1579

### DIFF
--- a/src/components/calcite-button/calcite-button.tsx
+++ b/src/components/calcite-button/calcite-button.tsx
@@ -36,8 +36,8 @@ export class CalciteButton {
   /** specify the appearance style of the button, defaults to solid. */
   @Prop({ reflect: true }) appearance: ButtonAppearance = "solid";
 
-  /** The aria-label attribute to apply to the button or hyperlink */
-  @Prop() ariaLabel?: string;
+  /** Applies to the aria-label attribute on the button or hyperlink */
+  @Prop() label?: string;
 
   /** specify the color of the button, defaults to blue */
   @Prop({ reflect: true }) color: ButtonColor = "blue";
@@ -82,7 +82,7 @@ export class CalciteButton {
   @Prop() target?: string;
 
   /** The type attribute to apply to the button */
-  @Prop() type?: string;
+  @Prop({ mutable: true }) type?: string;
 
   /** specify the width of the button, defaults to auto */
   @Prop({ reflect: true }) width: Width = "auto";
@@ -149,7 +149,7 @@ export class CalciteButton {
 
     return (
       <Tag
-        aria-label={this.ariaLabel}
+        aria-label={this.label}
         class={{ [CSS_UTILITY.rtl]: dir === "rtl", [CSS.contentSlotted]: this.hasContent }}
         disabled={this.disabled}
         href={this.childElType === "a" && this.href}
--- a/src/components/calcite-button/calcite-button.tsx
+++ b/src/components/calcite-button/calcite-button.tsx
@@ -1,6 +1,6 @@
 import { Component, Element, h, Method, Prop, Build, State, VNode } from "@stencil/core";
 import { CSS, TEXT } from "./resources";
-import { getAttributes, getElementDir } from "../../utils/dom";
+import { getElementDir } from "../../utils/dom";
 import { ButtonAlignment, ButtonAppearance, ButtonColor } from "./interfaces";
 import { FlipContext, Scale, Width } from "../interfaces";
 import { CSS_UTILITY } from "../../utils/resources";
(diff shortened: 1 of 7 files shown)


src/components/calcite-button/calcite-button.tsx
```diff
@@ -150,12 +134,19 @@ export class CalciteButton {
 
     return (
       <Tag
-        {...attributes}
+        aria-label={this.el.getAttribute("aria-label")}
```
driskull, line 137: These are still just being passed from the host to the internal tag as the same attribute. Can we add props for things like 'label', 'href', 'name', 'rel' etc? Then they can be documented with descriptions on how and when they can be set.

src/components/calcite-button/calcite-button.tsx
```diff
@@ -150,12 +134,19 @@ export class CalciteButton {
...
         class={{ [CSS_UTILITY.rtl]: dir === "rtl", [CSS.contentSlotted]: this.hasContent }}
         disabled={this.disabled}
+        href={this.childElType === "a" && this.href}
+        name={this.childElType === "button" && this.el.getAttribute("name")}
         onClick={this.handleClick}
         ref={(el) => (this.childEl = el)}
+        rel={this.childElType === "a" && this.el.getAttribute("rel")}
+        role={this.el.getAttribute("role")}
```
driskull, line 145: We should set the role for the user as a convenience I think.

src/components/calcite-button/calcite-button.tsx
```diff
@@ -150,12 +134,19 @@ export class CalciteButton {
...
         tabIndex={this.disabled ? -1 : null}
+        target={this.childElType === "a" && this.el.getAttribute("target")}
+        title={this.el.getAttribute("title")}
```
driskull, line 148: I don't think this is needed to be passed. title is a native attribute that can just be set on the host if needed.

driskull (changes requested):
I think these need to be documented props for usage.

src/components/calcite-button/calcite-button.tsx
eriklharper: What should it be?

src/components/calcite-button/calcite-button.tsx
eriklharper: I'll try removing that and see if the accessibility tests still pass, I thought I saw one fail because the title wasn't being added, but I could be wrong.

src/components/calcite-button/calcite-button.tsx
driskull: Is it even needed? A button shouldn't need a role of button.

src/components/calcite-button/calcite-button.tsx
eriklharper: Yep, this is in fact needed by `calcite-fab`, which passes a `title` attribute to button.  Should I make `title` an official prop also?

driskull: @eriklharper no, title is already a global property. It shouldn't need to be passed at all. Passing would make the title tooltip display on both nodes redundantly.

eriklharper: Does this test need changed or does calcite-fab need to be changed as a result of this change then, because this test definitely isn't passing as a result of removing `title` 
image

eriklharper: I'm wondering what the difference is between `calcite-fab`'s usage of `aria-label` versus `title`?

driskull: title wouldn't break a11y. Its probably how the aria-label is being set.

eriklharper: Looks like title is breaking a11y in this case, otherwise it wouldn't say "Element has no title attribute" as one of the possible issues.  I'll try and see how the aria-label is being set to see if there's a way to fix that instead of adding the title.

src/components/calcite-button/calcite-button.tsx
```diff
@@ -31,9 +30,15 @@ export class CalciteButton {
...
   //--------------------------------------------------------------------------
 
+  /** optionally specify alignment of button elements. */
+  @Prop({ reflect: true }) alignment?: ButtonAlignment = "center";
+
   /** specify the appearance style of the button, defaults to solid. */
   @Prop({ reflect: true }) appearance: ButtonAppearance = "solid";
 
+  /** The aria-label attribute to apply to the button or hyperlink */
+  @Prop() ariaLabel?: string;
```
driskull, line 40: Can you rename to just 'label'? `aria-label` is already a global property and we don't want it to be set on the host.

driskull (commented):
Comment

src/components/calcite-button/calcite-button.tsx
eriklharper: Just updated the test to pass both a label and a text attribute, that was the issue all along it appears.

driskull (approved):
👍 looks good if it passes.

Issue #2267, opened 2021-06-07
A11y: [calcite-color-picker] - focus should be set on the color field


hccampos:
### Summary
Currently, calling `setFocus` on a `<calcite-color-picker>` sets the focus on the `hexInputNode` but that the first focusable element in the component is the color field. 

### Actual Behavior
- Calling `setFocus` on a `<calcite-color-picker>` sets the focus on the `hexInputNode`.
- Furthermore, if `.setFocus()` is called immediately after creating the color picker, the hex input won't be initialized and `.setFocus()` will be no op.

### Expected Behavior
- The first focusable element in the `<calcite-color-picker>` is the main color field handle, so it should be focused when calling `.setFocus()`.
- If `.setFocus()` is called before the internal elements are present, the color picker should remember this and set focus when possible. Alternatively, the component could provide a `whenFocusable()` method (or similar).

### Reproduction Steps

1. Create a `<calcite-color-picker>`.
2. Call `element.setFocus()`.
3. Note that the focus is not set right away because the hex input isn't ready immediately.
4. Wait for the hex input to be ready and call `.setFocus()` again.
5. Note that the focus is set, but on the hex input which isn't the first focusable element of the component.

jcfranco:
Closing as self-verified with the following: https://codepen.io/jcfranco/pen/YzVZdLj?editors=1000

PR #2413, opened 2021-06-29
fix(color-picker): setFocus() should focus on the first focusable element. #2267

driskull:
**Related Issue:** #2267

## Summary

fix(color-picker): setFocus() should focus on the first focusable element. #2267


src/components/calcite-color-picker/calcite-color-picker.tsx
```diff
@@ -606,8 +616,11 @@ export class CalciteColorPicker {
 
   /** Sets focus on the component. */
   @Method()
-  async setFocus(): Promise<void> {
-    focusElement(this.hexInputNode);
+  async setFocus(focusId?: "hex-input"): Promise<void> {
+    const elementToFocus =
+      focusId === "hex-input" ? this.hexInputNode : getFocusableElements(this.el.shadowRoot)[0];
```
jcfranco, line 621: I think you can simplify this by checking which parts are hidden and focusing on the relevant refs since there's no user-provided content:

```
// pseudocode
if (focusIdHexInputAndNotHidden) 
  focusElement(hexInputNode)
else 
  focusElement(scopeNode)
```

src/components/calcite-color-picker/calcite-color-picker.tsx
```diff
@@ -700,6 +713,7 @@ export class CalciteColorPicker {
             aria-valuemin="0"
             aria-valuenow={(vertical ? color?.saturationv() : color?.value()) || "0"}
             class={CSS.scope}
+            data-test-id="color-field-scope"
```
jcfranco, line 716: Can you use the 1st element w/ the scope class or additional class for now? I see the value in this, but would like to spend some time on guidelines on when/how to use it to avoid having test IDs all over.

src/utils/dom.ts
```diff
@@ -121,15 +121,19 @@ function closestElementCrossShadowBoundary<E extends Element = Element>(
 }
 
 export interface CalciteFocusableElement extends HTMLElement {
-  setFocus?: () => void;
+  setFocus?: () => Promise<void>;
+}
+
+export function hasSetFocus(el: CalciteFocusableElement): boolean {
```
jcfranco, line 127: Naming suggestion: `isCalciteFocusable`?

driskull:
I tried to move `getFocusableElements()` to `utils/dom` but the way that the `@ally` dependency is setup it wouldn't work.

```
 Test suite failed to run
    Jest encountered an unexpected token
    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.
    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".
    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/en/ecmascript-modules for how to enable it.
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.
    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html
    Details:
    /home/travis/build/Esri/calcite-components/node_modules/@a11y/focus-trap/shadow.js:10
    export function queryShadowRoot(root, skipNode, isMatch, maxDepth = 20, depth = 0) {
    ^^^^^^
    SyntaxError: Unexpected token 'export'
      1 | import { CSS_UTILITY } from "./resources";
      2 | import { guid } from "./guid";
    > 3 | import { queryShadowRoot } from "@a11y/focus-trap/shadow";
        | ^
      4 | import { isFocusable, isHidden } from "@a11y/focus-trap/focusable";
      5 | 
      6 | /**
      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1350:14)
      at Object.<anonymous> (src/utils/dom.ts:3:1)
```

src/components/calcite-color-picker/calcite-color-picker.e2e.ts
```diff
@@ -18,6 +18,17 @@ describe("calcite-color-picker", () => {
 
   afterEach(() => consoleSpy.mockClear());
 
+  it("should focus scope", async () =>
```
jcfranco, line 21: Please group related focus tests in a `describe("is focusable (or similar)")` block.

jcfranco (changes requested):
Looking good, man!

src/components/calcite-color-picker/calcite-color-picker.tsx
driskull: I thought we already accepted this approach in a meeting way back? There are a couple other uses like this if you search for data-test-id.

jcfranco: I remember we discussed it, but don't remember adopting it. Feel free to leave this one based on that, but we should revisit since this ends up being included in production code and our build unfortunately doesn't have a way to automatically strip these out.

driskull: The other uses are in tree/tree-item.

driskull: Can you create an issue for discussion? If we don't adopt it we can remove both.

Its easier to identify with a data-attribute but I can add an additional class that serves no purpose if needed.

src/components/calcite-color-picker/calcite-color-picker.tsx
driskull: @jcfranco are you saying to not use `getFocusableElements`? Because it does that work already.

If we want to simplify, i can just get rid of the focusId argument and not allow a user to choose what to focus on.

driskull:

src/components/calcite-color-picker/resources.ts
```diff
@@ -22,6 +22,8 @@ export const CSS = {
   colorFieldAndSliderInteractive: "color-field-and-slider--interactive",
   colorFieldAndSliderWrap: "color-field-and-slider-wrap",
   scope: "scope",
+  hueScope: "scope--hue",
```
jcfranco, line 25: Thanks for adding these specific ones.

src/components/calcite-color-picker/calcite-color-picker.tsx
```diff
@@ -607,7 +609,9 @@ export class CalciteColorPicker {
   /** Sets focus on the component. */
   @Method()
   async setFocus(): Promise<void> {
-    focusElement(this.hexInputNode);
+    const elementToFocus = getFocusableElements(this.el.shadowRoot)[0] || this.el;
```
jcfranco, line 612: No need to query, let's just focus on the scope ref node. 

Sidebar, looks like `focusElement(host)` will cause an infinite loop based on the logic. Do you think renaming the parameter would help? Otherwise, we may need to refactor to `focusElement(host, target)`. We can tackle in a separate issue.

jcfranco (changes requested):
One more change and I think this is good to murge!

src/components/calcite-color-picker/calcite-color-picker.tsx
driskull: ah, yeah good catch. I'll remove.

> No need to query, let's just focus on the scope ref node.

Ok. We can revisit after we get feedback from level access.

driskull:
Ready for re-review.

src/components/calcite-color-picker/calcite-color-picker.tsx
```diff
@@ -607,7 +596,13 @@ export class CalciteColorPicker {
...
   @Method()
   async setFocus(): Promise<void> {
-    focusElement(this.hexInputNode);
+    const { colorFieldScopeNode, el } = this;
+
+    if (colorFieldScopeNode) {
+      return focusElement(colorFieldScopeNode);
+    }
+
+    el.focus();
```
jcfranco, line 605: Focusing on the host element here has the same effect as not focusing since it has the default tab-index.

I think this should be: 

```ts
  return focusElement(this.colorFieldScopeNode);
}
```

driskull: fixed

jcfranco (approved):
Awesome! 🎉 🌮

Issue #2241, opened 2021-06-01
Docs: Enhance auto-generated documentation content


macandcheese:
### Description 
To support https://github.com/ArcGIS/afd-calcite-documentation/issues/165, we need to improve / add to the auto-generated documentation per component. These improvements will also help fill out the readme files on GH.

There are some holes in the auto-generated documentation, mostly when the documentation is referencing a typed interface or content from a resource file. Some examples:

- [x] For events, we should return the content of the detail, not just the event name.
    - Example: Pagination https://github.com/Esri/calcite-components/tree/master/src/components/calcite-pagination#events - this should also include the detail of CalcitePaginationDetail - https://github.com/Esri/calcite-components/blob/781a7a7b8ae7b34b0138739ce1c61114e1200e3d/src/components/calcite-pagination/calcite-pagination.tsx#L18 (pr #2497)

- [x] For props with defined default text strings in an external resource file, we should return the string, not the name of the referenced variable (pr #2496)
    - Example: Pagination https://github.com/Esri/calcite-components/tree/master/src/components/calcite-pagination#properties - this should include the default string in resources.ts (ie, show "next" in the documentation, not "TEXT.nextLabel")
    - This also occurs when referencing vars from elsewhere in the project - ie defaultOffsetDistance on popper components.
- When accepted prop values come from an interface - just return the human readable strings
  - Example: TreeSelectionMode in tree https://github.com/Esri/calcite-components/tree/master/src/components/calcite-tree#properties (pr #2497)

- [x] Document and display component-specific CSS vars (I think this can be documented similarly to slots at the top of a jsx file) (pr #2404)

I'm sure more items will come up, feel free to add more examples here.

There's a lot of inconsistency around how we write the documentation for props - sometimes we indicate optionally in the definition, other times we just add it to the prop annotation. It will be easier for us to go through and address the voice / tone differences once the auto-generated documentation is visible.

caripizza:
Related issue: #1097

@macandcheese when you have a moment, can you add in the docs-json data you're referring to here for clarity on each bullet point?

macandcheese:
Sure, here are some current data shapes from that file. As long as the content is present in the file we can display it consistently in the docs site.

Events without detail example (pagination):

```
 "events": [
        {
          "event": "calcitePaginationChange",
          "detail": "CalcitePaginationDetail",
          "bubbles": true,
          "cancelable": true,
          "composed": true,
          "docs": "Emitted whenever the selected page changes.",
[...]

PR #2414, opened 2021-06-29
docs: troubleshooting type info that comes through in readmes (TESTING)

caripizza:
**Related Issue:** #2433 (#2241)

## Summary
WIP - I'm testing the types that come through when readme content is auto-generated from `npm run build`. I'm seeing the types correctly appear in the docs-json. However, for some reason the Storybook readme content misses these types, and just uses `any`.


caripizza:
@jcfranco these are the differences I noticed in the auto-generated readmes while working on #2241. Do you know why the readmes/auto-update task isn't printing these correct types into the markdown? Should we just update them manually by merging this one in?

jcfranco:
It could be because the auto-doc-update runs a Stencil doc build (`stencil build --docs`) and not `npm run build`.

caripizza:
gotcha, @jcfranco can we update the action to use `npm run build` instead, so these types appear like they do in the bundle output? (I can convert issue 2433 to request that work if so.)

caripizza:
Closing this for now, we can prioritize in the related issue since this is pretty low-hanging fruit (#2433).

jcfranco:
@caripizza Sounds good. I still haven't been able to repro the different doc scenario. These are the steps I'm following:

1. check out local test branch (based on latest `master`)
2. run `npm run build`
3. commit readme changes
4. run `npx stencil build --docs`
5. commit read me changes

Step 5 fails after the commit is auto-linted as it ends up being identical to the existing code and bails to avoid merging an empty commit. 

On the Storybook-side of things, that doc is only updated when we release, so the auto-updating action wouldn't have any effect on it between releases. 

Let me know if I missed a step. We can also move this discussion to #2433.

Issue #2290, opened 2021-06-09
Bug: calciteTreeSelect event doesn't fire on spacebar keypress


joeyHarig:
### Actual Behavior
The event `calciteTreeSelect` fires when the user clicks on a tree item or presses enter while focused on the tree-item. When the user is focused on the `calcite-tree-item` and presses the spacebar key, the `calcite-tree-item` appears to be selected, but the event doesn't fire.

### Expected Behavior
The event `calciteTreeSelect` should fire on click, enter keypress and spacebar keypress.

### Reproduction Steps or Sample

1. console.log the event returned from `calciteTreeSelect` 
2. Press the enter key while focussed on a `calcite-tree-item`
3. Notice that the event is logged
2. Press the spacebar while focussed on a `calcite-tree-item`
3. Notice that the event is not logged despite the tree item appearing selected

### Relevant Info

_Browser_: Chrome, Firefox
_Tree Accessibility Guides_: https://www.w3.org/TR/wai-aria-practices-1.1/#TreeView
_Version_: `@esri/calcite-components@1.0.0-beta.51`

- [ ] CDN
- [ ] NPM package

jcfranco:
@joeyHarig Can you confirm if this is still an issue? The following codepen uses the latest and the event seems to fire on click and Enter/Space keydown: https://codepen.io/jcfranco/pen/gOmNByj

Sidebar, I did notice that when a nested item is selected, two events are fired. @paulcpederson Do you know if this is the expected behavior? If not, I'll create a separate issue for this.

paulcpederson:
@jcfranco it's the multiple trees, each one emits a change event, so when you select one of the nested ones you get n events based on your depth.

paulcpederson:
I don't know if it's _expected_, though...

joeyHarig:
@paulcpederson Let me update our CC dependency and I'll get back to you.

jcfranco:
@joeyHarig Did you get a chance to test this with the latest?

joeyHarig:
@jcfranco I have not yet, but I will have time tomorrow and I'll get back to you by EOBD

jcfranco:
Thanks! I truly appreciate it. 😄

joeyHarig:
@jcfranco I created a code pen where I have reproduced the bug: https://codepen.io/joeyHarig/pen/eYWmpbm?editors=1111

[...]

PR #2416, opened 2021-06-30
fix(tree): allow selecting children with Enter/Space when selection-mode=children

jcfranco:
**Related Issue:** #2290 

## Summary

This changes the `forceToggle` option in tree-item to be the same as when the item is clicked based on the premise that both of these events would trigger a selection event.

@paulcpederson Can you take a look at this? I'm not sure how to manually test for the ancestors-selection scenario.


driskull (approved):
👍

paulcpederson:
Anscestors selection mode is easiest to see with input-enabled. Basically when you select children the parent should then appear as an indeterminate checkbox. When all children are selected it should appear checked.

This change looks good to me!

paulcpederson (approved):

Issue #683, opened 2020-06-24
Accessibility: calcite-tree traps focus, preventing backwards navigation out of it


MikeTschudi:
The following function in calcite-tree.tsx prevents the focus from navigating backwards out of the tree:
```tsx
  @Listen("focus") onFocus() {
    if (this.root) {
      const selectedNode = this.el.querySelector(
        "calcite-tree-item[selected]"
      ) as HTMLCalciteTreeItemElement;
      const firstNode = this.el.querySelector(
        "calcite-tree-item"
      ) as HTMLCalciteTreeItemElement;

      (selectedNode || firstNode).focus();
    }
  }
```

Without the ability to go backwards, a user has to go forwards through many elements to cycle around back to the beginning of the page.

Can this function be removed?

[shift + tab](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Accessibility)

MikeTschudi:
Removing the function is not appropriate by itself because there are bigger accessibility problems.

Our version traps the backwards focus and causes the screen reader to say more than is needed. E.g., after clicking on the first tree item ("Child 1") in the [tree example](https://esri.github.io/calcite-components/?path=/story/tree--simple) and tabbing through the list, one hears NVDA say
```
Child 1
clickable  •  clickable  Child 1  clickable   
clickable     clickable  Child 2  clickable  •  clickable  Grandchild 1  clickable     clickable     clickable  Grandchild 2  clickable      
clickable  •  clickable  Grandchild 1  clickable   
clickable     clickable  Grandchild 2  clickable   
clickable     clickable  Child 3  clickable  
```
The arrow keys work, but don't seem to because they stop on each of the extra, hidden "clickables"--items with focus. It's awkward, and doesn't seem to meet the [WAI-ARIA Authoring Practices 1.1 for a TreeView](https://www.w3.org/TR/wai-aria-practices-1.1/#TreeView).

http://www.oaa-accessibility.org/examplep/treeview1/ handles it the way that the Practices recommends. Running this example in a screen reader makes sense, and it doesn't trap the focus. E.g., stepping to the first tree item ("Fruits"), one hears
```
Foods  tree view
Group collapsed Fruits  collapsed  1 of 2  level 1
expanded
level 2  Oranges  1 of 5
Pineapples  2 of 5  level 2
Group expanded Apples  expanded  3 of 5  level 2
Pineapples  2 of 5  level 2
Oranges  1 of 5  level 2
[...]

PR #2419, opened 2021-06-30
fix(tree): allow focusing out of tree via shift + tab

jcfranco:
**Related Issue:** #683 

## Summary

This fixes an issue where tree would keep focusing on the first tabbable item when you tried focusing out of it. This updates the component to make the root only tabbable if focus has been moved outside.

This also has a test for this case, but it's weird to have a test for browser element tabbing. 

I looked at adding a `setFocus` method, but will create a separate issue for it instead since I'm not sure if we need special considerations for focusing based on the selection mode.


driskull (approved):
👍

PR #2427, opened 2021-07-01
refactor(calcite-dropdown-item): stop spreading attributes

eriklharper:
**Related Issue:** #1579


src/components/calcite-dropdown-item/calcite-dropdown-item.tsx
```diff
@@ -142,7 +133,14 @@ export class CalciteDropdownItem {
...
       slottedContent
     ) : (
-      <a {...attributes} class="dropdown-link" ref={(el) => (this.childLink = el)}>
+      <a
+        class="dropdown-link"
+        href={this.href}
+        ref={(el) => (this.childLink = el)}
+        rel={this.el.getAttribute("rel")}
+        target={this.el.getAttribute("target")}
+        title={this.el.getAttribute("title")}
```
eriklharper, line 142: @driskull should we make explicit props for these as well?

driskull: Yes 👍 

For props that already exist at global level like 'title' we will need to use label or something.

driskull: explicit props means that they are documented and supported. Unlike before :)

eriklharper: I matched what we just did with `calcite-button` and created a label prop to apply to `aria-label` on the `<a>` and ditch setting the `title` attribute.

driskull (approved):
👍 Sounds good. If we need a way to set a title (tooltip) we can do that later.

Issue #2091, opened 2021-05-03
Refactor: calcite-input tests show 'The state/prop "value" changed during rendering' warnings 


driskull:
## Summary

Test runs show the following warnings:

```
PASS src/components/calcite-input/calcite-input.e2e.ts (329.206 s)
376 ● Console
377
378 console.warn
379 STENCIL: The state/prop "value" changed during rendering. This can potentially lead to infinite-loops and other bugs. 
380 Element JSHandle@node 
381 New value 
382 Old value 
383 Location: http://localhost:3333/build/index-09c7248a.js:2710:70
384
394 console.warn
395 STENCIL: The state/prop "localizedValue" changed during rendering. This can potentially lead to infinite-loops and other bugs. 
396 Element JSHandle@node 
397 New value 
398 Old value 
399 Location: http://localhost:3333/build/index-09c7248a.js:2710:70
```

## Additional Info 

probably related to this? https://github.com/Esri/calcite-components/pull/2036/files#diff-d4e9e38d573f92a6608141863d1a3ce63b4cc9c7bc0b97585e8a91c48ac9a705R286

eriklharper:
This is caused anytime a number input's `.value` is set on the input programmatically to an invalid number, which resets the number back to the previous value, which one of the tests is doing.  I don't know why Stencil complains about this and if there is another way to implement this without triggering the warning, but I can look into it.

jcfranco:
Changed the title, milestone and priority of this one since it's not an unstable test.

jcfranco:
Warnings are no longer observed with the latest codebase. Closing!

Issue #1485, opened 2021-01-29
Enhancement: Add ability to set "any" on the step property of calcite-input


ethanbdev:
### Description
Currently, the step property can only be set to a number. The native HTML input has an option for "any" to be set. This option would leave the ability to use the arrows to increment by 1 but remove the step being considered during validation, removing tooltips like this:
![image](https://user-images.githubusercontent.com/52869490/106213892-86320580-6182-11eb-86e6-764fdc4412fa.png)

### Acceptance Criteria 
Can set the step property to "any" as well as a number without changing any other behavior.
### Relevant Info 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#htmlattrdefstep
#### Which Component
calcite-input
#### Example Use Case
There is a minimum/max value calculated that the use can also change. The calculated value cannot be rounded to match the `step` precision as this would be inaccurate. Setting to "any" removes this extra validation that is not needed.

eriklharper:
Looks like this issue has been resolved.  Will go ahead and close it.

PR #2428, opened 2021-07-02
fix(input): Fix step property rendering errors.

driskull:
**Related Issue:** #2091

## Summary

fix(input): Fix step property rendering errors.

### DIFF
--- a/src/components/calcite-input/calcite-input.e2e.ts
+++ b/src/components/calcite-input/calcite-input.e2e.ts
@@ -252,7 +252,7 @@ describe("calcite-input", () => {
   it("correctly increments and decrements value by one when any is set for step", async () => {
     const page = await newE2EPage();
     await page.setContent(`
-    <calcite-input type="number" step="any" value="5.5"></calcite-input>
+    <calcite-input type="number" value="5.5"></calcite-input>
     `);
 
     const element = await page.find("calcite-input");
@@ -311,7 +311,7 @@ describe("calcite-input", () => {
     expect(await element.getProperty("value")).toBe("12");
     expect(await element.getProperty("min")).toBe(10);
     expect(await element.getProperty("max")).toBe(15);
-    expect(await element.getProperty("step")).toBe("1");
+    expect(await element.getProperty("step")).toBe(1);
 
     element.setProperty("min", null);
     element.setProperty("max", null);
--- a/src/components/calcite-input/calcite-input.tsx
+++ b/src/components/calcite-input/calcite-input.tsx
@@ -147,7 +147,7 @@ export class CalciteInput {
   @Prop({ mutable: true, reflect: true }) status: Status = "idle";
 
   /** input step */
-  @Prop({ mutable: true, reflect: true }) step?: number | "any";
+  @Prop({ reflect: true }) step?: number | "any";
 
   /** optionally add suffix  **/
   @Prop() suffixText?: string;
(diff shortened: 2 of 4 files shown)


driskull:
@eriklharper can we make 'step' to be only a number so that those stencil warnings are no longer reported? 

It seems like `step: any` is just another way for step to be `1` when the type is a number. It seems like we can do that internally, right?

Also, what is the reasoning for `maxString` and `minString`? Seems like we could just pass the number to the input instead of maxString/minString.

src/components/calcite-input/calcite-input.e2e.ts
```diff
@@ -311,7 +311,7 @@ describe("calcite-input", () => {
     expect(await element.getProperty("value")).toBe("12");
     expect(await element.getProperty("min")).toBe(10);
     expect(await element.getProperty("max")).toBe(15);
-    expect(await element.getProperty("step")).toBe("1");
```
driskull, line 314: This was kind of weird to be a string. If min/max are numbers why can't step be?

eriklharper:
> @eriklharper can we make 'step' to be only a number so that those stencil warnings are no longer reported?
> 
> It seems like `step: any` is just another way for step to be `1` when the type is a number. It seems like we can do that internally, right?
> 
> Also, what is the reasoning for `maxString` and `minString`? Seems like we could just pass the number to the input instead of maxString/minString.

@ethanbdev reported [this issue](https://github.com/Esri/calcite-components/issues/1485) which is why support for `any` is in place.  We could add a custom boolean prop like `stepAny` that could be used to both satisfy the requirements in #1485 and get around this annoying console issue, but ultimately, [this is a Stencil issue](https://github.com/ionic-team/stencil/issues/2814) that they just need to fix already, IMHO.

src/components/calcite-input/calcite-input.e2e.ts
eriklharper: Yeah, if its checking the property, it should be a number, not a string.  Only the attribute would be set to a string.

eriklharper:
> Also, what is the reasoning for `maxString` and `minString`? Seems like we could just pass the number to the input instead of maxString/minString.

No idea why `maxString` and `minString` have to exist.  Stencil/JSX should be able to handle passing the number value to the input.

src/components/calcite-input/calcite-input.e2e.ts
```diff
@@ -252,7 +252,7 @@ describe("calcite-input", () => {
   it("correctly increments and decrements value by one when any is set for step", async () => {
     const page = await newE2EPage();
     await page.setContent(`
-    <calcite-input type="number" step="any" value="5.5"></calcite-input>
+    <calcite-input type="number" value="5.5"></calcite-input>
```
eriklharper, line 255: If we're really gonna drop support for `any` this test should just get deleted.

src/components/calcite-input/calcite-input.tsx
```diff
@@ -244,6 +244,16 @@ export class CalciteInput {
...
   @State() localizedValue: string;
 
+  @State() computedStep: number;
+
+  @Watch("type")
+  @Watch("step")
+  setComputedStep(): void {
+    const { step, type } = this;
+
+    this.computedStep = step ? Math.abs(step) : type === "number" ? 1 : null;
```
eriklharper, line 254: ☝️ `step` supports floats in addition to integers, so we probably shouldn't do `Math.abs` here.  This is particularly relevant for making sure up and down arrows increment properly when decimals are expected in the input and the user wants a finer-grained control over the stepped values.

eriklharper (changes requested):
I think rather than totally ditching `step="any"` support we should provide a prop that allows setting that special value for the use cases where it is needed.  I can envision something like when a `stepAny` prop is set to `true` it will bypass the `this.step` property entirely to achieve its intended behavior, but by default it will stay `false` and we can permanently convert it to be a number all the time.

Issue #2412, opened 2021-06-29
Bug: ColorPicker display is inline by default which makes the height only 19px.


driskull:
### Actual Behavior

Bug: ColorPicker display is inline by default which makes the height only 19px.

![MicrosoftTeams-image](https://user-images.githubusercontent.com/1231455/123864430-9be95980-d8df-11eb-88e2-1e537f93dc93.png)

### Expected Behavior

```
:host {
 display: inline-block;
}
```

### Reproduction Steps or Sample

Can be seen on storybook as well. https://esri.github.io/calcite-components/?path=/story/components-controls-colorpicker--simple

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [x] NPM package

driskull:
cc @julio8a @caripizza We should make sure that all components have the correct display type by default so that their dimensions are correct.

caripizza:
@driskull sorry can you clarify? Was a defect introduced on one of our related refactor prs?

driskull:
@caripizza nope I don't think so, it seems like it was always like this.

I just wanted to make an announcement to the peeps who work on the SCSS files.

We might need to do an audit to make sure that the host has the correct display type for all components.

jcfranco:
Installed!

PR #2432, opened 2021-07-02
fix(color-picker): ensure color picker dimensions match its content

jcfranco:
**Related Issue:** #2412 

## Summary

The host was using the default `display` value, causing the component's box model to not match the component (visually).


driskull (approved):
💯

Issue #2196, opened 2021-05-21
Bug: Calcite Tree in input-enabled mode fires select twice when label is clicked


paulcpederson:
### Actual Behavior

calciteTreeSelect event is fired twice

### Expected Behavior

calciteTreeSelect event should be fired once

### Reproduction Steps or Sample

1. go to https://codepen.io/paulcp/pen/GRWWoLx?editors=1111
2. click the _text_ of the first tree option to select it
3. notice the console has two logs that are identical

jcfranco:
Installed!

julio8a:
verified

PR #2438, opened 2021-07-03
fix(tree): prevent emitting selection event twice when a tree-item's checkbox label is clicked

jcfranco:
**Related Issue:** #2196 

## Summary

This PR fixes an issue caused the default behavior from the tree item's checkbox label. Clicking it would cause a 2 clicks to be fired (from the host and checkbox+label). The internal label + checkbox are no longer interactive since the item itself handles this already.

### DIFF
--- a/src/components/calcite-tree-item/calcite-tree-item.scss
+++ b/src/components/calcite-tree-item/calcite-tree-item.scss
@@ -124,6 +124,7 @@
 .calcite-tree-label {
   display: flex;
   align-items: center;
+  pointer-events: none;
 }
 
 .calcite-tree-children {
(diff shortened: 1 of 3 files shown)


driskull (approved):
Nice fix 💯

paulcpederson:
@jcfranco this solves part of the issue, but leaves the duplicate events. If you click an item that's nested within multiple trees, all the trees fire events. Maybe this is fine? If that's the desired behavior we should offer some guidance on how you receive just one event or disambiguate between the events in some way.

jcfranco:
@paulcpederson I think the duplicate event per nested tree is a separate issue. Do you know what the use case is for the nested tree events? IMO, there should only be one selection event fired at the tree root.

paulcpederson:
The other option would be for the tree to handle events from nested trees inside of itself and stop them from leaking out.

jcfranco:
Yeah. I need to double check, but I think it's supposed to stop the nested tree selection events from propagating. I can create a separate issue for this since I couldn't find one. WDYT?

paulcpederson:
> IMO, there should only be one selection event fired at the tree root.

Agree. This would make my life a lot easier.

> I can create a separate issue for this since I couldn't find one. WDYT?

Sounds good 👍🏻

paulcpederson (approved):

Issue #720, opened 2020-07-10
Enhancement: [Slider] support RTL / Mirrored


macandcheese:
While working on the video component, I discovered we don't currently support flipping the slider orientation for `dir=rtl`

After discussion in https://github.com/Esri/calcite-components/issues/721 - maybe this is really a `mirrored` prop instead?

macandcheese:
Valid

dhatcher:
Is there any update on when this might be able to be implemented/looked into?

jcfranco:
Hi @dhatcher. This is scheduled for the current sprint, so we're looking at having it land by 6/7.

jcfranco:
Apologies, I didn't get a chance to wrap this up by today. I'll push it to the next spring (starts today) and will shoot for this week.

jcfranco:
FYI, I've got PR coming up soon. A couple of questions popped up from this:

* does `mirrored` only affect the component visually or does it also invert keyboard interaction? 
* how should we deal with the histogram + slider use case? Based on https://material.io/design/usability/bidirectionality.html#mirroring-layout, talking to some teams that work w/ data visualization + sliders and some good ol' fashioned googling, charts/graphs shouldn't be mirrored, which makes the histogram use case incompatible w/ this property. Some options for dealign with the slider + histogram use case are:
a. ignore `mirrored`
b. hide the histogram
c. mirror it and leave to the user

  For each of these, we could additionally display a warning message or throw an error if we want to prevent this scenario.

@Esri/calcite-design-representatives WDYT?

caripizza:
I'll defer to @macandcheese and team here since I'm newer to the conversation. To me, slider's UI in rtl means `minValue` => `maxValue` appears as `maxValue` <= `minValue` on the scale (for single and range use cases):
![image](https://user-images.githubusercontent.com/42423180/124336800-597c8280-db54-11eb-94f4-efbad1b17a8e.png)

So yes, I think of the keyboard interaction being inverted. (ie., decrementing/incrementing slider value up or down, left/right arrow keys.)

And `a.` seems good to me - to ignore `mirrored` property with `histogram` use case (but I admit I'm probably missing some context there).

bstifle:
@caripizza that makes sense to me. maxValue <= minValue in RTL

jcfranco:
Moving to the next milestone. This should land this week.

jcfranco:
Installed. 

[...]

PR #2484, opened 2021-07-07
feat(slider): add support for mirroring slider

jcfranco:
**Related Issue:** #720

## Summary

This adds a `mirrored` property to `calcite-slider` (default false). When set to true, it will invert the slider unless there is a histogram associated. In this case, `mirrored` has no effect. This decision is based on https://material.io/design/usability/bidirectionality.html#mirroring-layout, talking to some teams that work w/ data visualization + sliders and some good ol' fashioned googling.

**Extras**

* sorts props alphabetically
* fixed casing of an internal type
* refactored method arguments/vars to make them value-agnostic - these are now left/right-based
* adds local vars to store bounding client rects to avoid duplicate computations
* adds another column for testing mirrored along with some edge-positioning cases
    Screen Shot 2021-07-07 at 2 33 32 PM

### DIFF
--- a/src/components/calcite-slider/calcite-slider.tsx
+++ b/src/components/calcite-slider/calcite-slider.tsx
@@ -1,13 +1,13 @@
 import {
   Component,
   Element,
-  Prop,
-  Host,
   Event,
   EventEmitter,
+  h,
+  Host,
   Listen,
   Method,
-  h,
+  Prop,
   State,
   VNode,
   Watch
@@ -15,9 +15,10 @@ import {
 import { guid } from "../../utils/guid";
 import { getKey } from "../../utils/key";
 import { DataSeries } from "../calcite-graph/interfaces";
-import { hasLabel } from "../../utils/dom";
+import { hasLabel, intersects } from "../../utils/dom";
+import { clamp } from "../../utils/math";
 
-type activeSliderProperty = "minValue" | "maxValue" | "value" | "minMaxValue";
+type ActiveSliderProperty = "minValue" | "maxValue" | "value" | "minMaxValue";
 
 @Component({
   tag: "calcite-slider",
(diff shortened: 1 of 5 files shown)


src/components/calcite-slider/calcite-slider.tsx
```diff
@@ -1158,33 +1178,32 @@ export class CalciteSlider {
    */
   private getHostOffset(leftBounds: number, rightBounds: number): number {
     const hostBounds = this.el.getBoundingClientRect();
-    if (leftBounds + 7 < hostBounds.left) {
-      const offset = hostBounds.left - leftBounds - 7;
-      return offset;
+    const buffer = 7;
```
driskull, line 1181: nitpick. Should we store this in the resources file?

src/components/calcite-slider/calcite-slider.tsx
```diff
@@ -41,57 +42,64 @@ export class CalciteSlider {
...
+
+  @Watch("histogram") histogramWatcher(newHistogram: DataSeries): void {
+    this.hasHistogram = !!newHistogram;
+  }
+
+  /** Label handles with their numeric value */
+  @Prop({ reflect: true }) labelHandles?: boolean;
+
+  /** Label tick marks with their numeric value. */
+  @Prop({ reflect: true }) labelTicks?: boolean;
```
driskull, line 59: nitpick: Do we have guidelines for when boolean props should be initialized to false vs undefined?

For example, hasHistagram = false but labelTicks = undefined. mirrored = false;

Its not a big deal just a consistency thing :)

src/utils/dom.ts
```diff
@@ -215,3 +215,12 @@ export function setRequestedIcon(
     return iconObject[matchedValue];
   }
 }
+
+export function intersects(rect1: DOMRect, rect2: DOMRect): boolean {
```
driskull, line 219: 👏

driskull (approved):
LGTM!

Nice cleanup as well!

src/components/calcite-slider/calcite-slider.tsx
driskull: I know we have the same inconsistencies with strings. Some are undefined by default and others are "";

jcfranco: We don't have any guidelines at the moment for defaults. Whenever possible, I think we should initialize to a sensible value (similar to native elements). This use case is weird to me if we don't:

```ts
const slider = document.createElement("calcite-slider");
console.log(slider.labelTicks); // undefined even though the doc types it as boolean
```

There could be some exceptions. For example, if we treat `intl` strings as overrides instead of defaults we could skip initializing to keep the defaults internal and to also promote reuse between components (e.g., [`calcite-time-picker#intlHour`](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-time-picker/calcite-time-picker.tsx#L59) vs. [`calcite-input-time-picker#intlHour`](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-input-time-picker/calcite-input-time-picker.tsx#L49) )

src/utils/dom.ts
eriklharper: this is awesome!  Now it can be used for more than just the x direction.

eriklharper (approved):
Excellent work!  Love the more semantic variable names that apply to the left/right versus min/max.  Makes the code so much easier to read and understand.

bstifle:
This has some weird problems with keyboard nav. Left goes right and right goes left... cant those stay the same for RTL? west isnt east in arabic ;)

jcfranco:
@bstifle Fixed left/right keyboard issue.

PR #2487, opened 2021-07-08
fix(combobox-item): Fix indent style for nested items.

driskull:
**Related Issue:** #2490

## Summary

fix(combobox-item): Fix indent style for nested items. #2490

Moves some CSS variables from the host element to the internal container element.


jcfranco (approved):
👍🏼  Can you update one of the stories, so Screener can keep an eye on this? 👁️

PR #2496, opened 2021-07-08
docs: add JSDoc tag to document non-simple default prop values for docs-json

caripizza:
**Related Issue:** #2241

## Summary
This pr targets the default text strings portion of the issue. Currently, defaults sourced from variables and resource objects don't get their values documented. To address, I added the `@default` JSDoc tag for documenting default prop values stored in variables and resource objects.

The new tag and its value now get added in the prop's `docsTags` array, like so (for display in future dev docs site property tables):
```js
        {
          "name": "intlLoading",
          "type": "string",
          "mutable": false,
          "attr": "intl-loading",
          "reflectToAttr": false,
          "docs": "string to override English loading text",
          "docsTags": [
            {
              "text": "\"Loading\"", // NEW
              "name": "default" // NEW
            }
          ],
          "default": "TEXT.loading",
          "values": [
            {
              "type": "string"
            }
          ],
          "optional": true,
          "required": false
        }
```

The other option to fix this documentation gap is to just store the default values themselves, and not source from the resource objects (like instead of  `= TEXT.loading` we'd say `= "Loading"`, and just use the object properties elsewhere in code. This would make it easier for display on docs site, since we could use the existing `prop.default` docs-json value).

To-do:
- [x] determine whether all non-empty default string values should be capitalized (ex: `intlClose` default value appears as both "Close" and "close" in different components) (Issue #2534 created to address)
- [ ] ~determine whether or not to store the `TEXT` (for aria labels) string values in a more re-usable place (and source from there instead)~ (will evaluate this need at a later time)


caripizza:
@macandcheese is this going to work for you, or how should I proceed?

macandcheese (approved):
This would work for me on the afd site.

PR #2497, opened 2021-07-08
docs: add JSDoc tag to document links to interface definitions in docs-json

caripizza:
**Related Issue:** #2241

## Summary
This pr targets the event details content and accepted prop values from an interface portions of the issue.

#### Events
Currently, when a component's event detail content is a custom interface (ie., not `any` or `void`), the contents aren't documented unless they're inlined, like `SomeEvent<{ a: string }>` instead of `SomeEvent<SomeEventDetail>`.

To address this doc limitation (since we'd like to display at very least a link to the interface definition in docs), I added `@see` JSDoc tags with markdown links to the definitions on master. (On docs site, we can then replace `master` with the cc version, so they appear with a version number instead, like `https://github.com/Esri/calcite-components/blob/v1.0.0-beta.59/src/components/calcite-tree/interfaces.ts#L1`.)

The resulting docs-json looks like so:
```js
      "events": [
        {
          "event": "calcitePaginationChange",
          "detail": "CalcitePaginationDetail",
          "bubbles": true,
          "cancelable": true,
          "composed": true,
          "docs": "Emitted whenever the selected page changes.",
          "docsTags": [
            {
              "text": " [CalcitePaginationDetail](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-pagination/calcite-pagination.tsx#L18)", // NEW
              "name": "see" // NEW
            }
          ]
        },
  ...
```

(If we prefer not to go this route, we can just inline doc the interface event details instead. This would make it easier on docs site since we can reference `event.detail` string instead of the `docsTags` array.)

#### Props
I also found a few props that utilize complex types (ex `placement`, `selectionMode`, etc), so I added `@see` JSDoc tags for them as well, with links to their interface definitions.


driskull:
Looks good @caripizza!

Maybe we can use the custom docs to replace some kind of [versionPath] variable? https://stenciljs.com/docs/docs-custom

That way the replacement doesn't have to happen only on docs side?

I'm thinking maybe we could have a variable for `https://github.com/Esri/calcite-components/tree/<package.json.version>`?

src/components/calcite-action-menu/calcite-action-menu.tsx
```diff
@@ -95,6 +95,7 @@ export class CalciteActionMenu {
 
   /**
    * Determines where the component will be positioned relative to the referenceElement.
+   * @see [PopperPlacement](https://github.com/Esri/calcite-components/blob/master/src/utils/popper.ts#L25)
```
driskull, line 98: Maybe these should just be relative paths like `../../utils/popper.ts#L25`? Then the consuming app (API pages) can just add the full url and version? Using `https://github.com/Esri/calcite-components/tree/` + <package.json.version> + `/...`?

driskull (approved):
👍

caripizza:
Thanks @driskull! Yep we'll use the `docs-custom` output target to generate our own json file to consume in the near future (and mutate/transform everything there instead of on consumer-end).

At that point we'll be able to use your suggestion, or even `process.env.npm_package_version` to get the cc version number in the markdown links. For now since we're short on time, I just wanted to match the way Storybook links to the readmes on master. (And the JSDoc tags appear in the built `components.d.ts` types file as well.)

Issue #2216, opened 2021-05-26
Bug: X icons in close buttons should hover/press to ui-text-1 color


bstifle:
When hovering over close buttons in CC, the X icon should change color to ui-text-1

Popover, Modal, Alert, Notice, Input (clear X) ((would affect inline-editable))

![image](https://user-images.githubusercontent.com/14083530/119736508-6b775100-be43-11eb-8080-7e33d90ccbe5.png)

julio8a:
assigning to @bstifle for verification.

bstifle:
Modal X is not ui-text-3 on idle

Input close X and other controls (spinners, chevrons, etc)  are not ui-text-3 on idle

Popover, Alert, Notice are good ✅

julio8a:
@bstifle, can you verify this one on master?

bstifle:
dang it. input and inline editable dont have this fix

bstifle:
I mentioned this in the comment above, before merge

https://github.com/Esri/calcite-components/issues/2216#issuecomment-873238055

julio8a:
Thanks, Bryan! Bringing this back to in dev

jcfranco:
Moving to the next sprint. @julio8a Feel free to adjust the sprint if needed.

julio8a:
Self-verified input buttons idle/hover
I'm going to close this one and add notes to the inline editable issue: #2468

PR #2501, opened 2021-07-09
refactor(components-x): components need consistent `x` idle, hover, etc

julio8a:
**Related Issue:** #2216

## Summary

- Modal X is not ui-text-3 on idle
- Input close X and other controls (spinners, chevrons, etc) are not ui-text-3 on idle

CC @bstifle


bstifle (approved):
lookin good!

Issue #2454, opened 2021-07-07
refactor(Pick List): match Figma and utilize Tailwind


julio8a:
Description
This component needs to match Figma designs.

Acceptance Criteria
The component appears as it does in Figma.

Related issues:
- #1547

Related prs:
- pr #1682
- pr #1684
- pr #1686

Design:
- Needs full refactor, see figma spacing
- https://www.figma.com/file/E3SB0i5wPy7AagB7KcjF05/?node-id=1183%3A7816

6/24: Mike Stinson is working on this refactor - CP

asangma:
A note on the future refactor using Calcite List.
As mentioned in the comment [here](https://github.com/Esri/calcite-components/pull/2502#issuecomment-883610499), refactoring the PickListItem to user ListItem will allow the dot or checkmark to become part of the main hit area and thus the hover area.

This is true. However, any action added will be to the left (in LTR) of that check or dot since the check or dot will be in the content area.

julio8a:
verified on master, local build demo pages.

PR #2502, opened 2021-07-09
refactor(pick-list, pick-list-group, pick-list-item): Design specs for pick list matched

rmstinson:
**Related Issue:** #2454

## Summary

Matched the visual specs for pick list from Figma, one behavioral issue pending. Hover state is an issue on the action slot. See below:

Picklist Behavior


caripizza (changes requested):
This is looking good @rmstinson, but can you clean it up a bit - remove whitespace changes and commented out code?

jcfranco:
> remove whitespace changes

Formatting should be taken care of automatically when committing. Can you run `npm install && npm run util:prep-autolint`?

bstifle:
@rmstinson looks like a couple visual bugs were introduced on hover:
image

image

image

hovering the x/end action should only hover the X hit area bg:
image
image

rmstinson:
@bstifle Yep, the hover bug was mentioned up top. Since the pick-list is full-width, and the action buttons are nested in the pick-list, there doesn't seem to be a way to prevent the hover state from being incorrectly triggered. I think that the host element needs to get reworked in order to get the hover state to work correctly. (this applies to the Value List too, which leverages the Pick List as like most of itself.) Syncing with @asangma tomorrow to address.

jcfranco:
Moving to the next sprint.

rmstinson (commented):
Made requested changes

bstifle:
i have a finesse ask.

this makes sense, because the active indicator is separated by the action:
image

Can we make some logic to where if no action-start, this does not happen?:
image

would make more sense if the active indicator and item hit areas were married in this case

bstifle:
that being said.... does the active indicator area even need its own affordance? its acting like an invisible checkbox

rmstinson:
@bstifle Good point, what you're describing is the desired behavior once we refactor the component itself and not just the styling. Because of the way the host is built, the current hover behavior is a little inaccurate to the designs. The icon slot and the label slots can't have active hover states at the same time, without also being active when the right-action slot is hovered on.

action_hoverstate

Once we're able to refactor the actual architecture of the component, it'll be super-duper easy to fix.

pick_hoverstate

bstifle:
dope dope dope

bstifle:
@rmstinson hey for top and bottom padding... can we get those back to p2 (8px)

Alan and I had worked on those to get them a little tighter

rmstinson:
@bstifle Changes made! 
Screen Shot 2021-07-20 at 3 47 05 PM

bstifle (approved):
Lookin good mr. mike

julio8a:
@rmstinson, can you merge master into this?

asangma (changes requested):
A couple minor things. 
Looks good otherwise!

rmstinson (commented):
Question about min-height attribute for Header-only, no right-action pick list

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
```diff
@@ -1,74 +1,84 @@
...
     flex
     text-color-1
     shadow-border-bottom
     m-0
     mb-px;
-  transition: background-color $transition;
-  animation: calcite-fade-in $transition;
+	transition: background-color $transition;
+	animation: calcite-fade-in $transition;
+	min-height: 3rem;
```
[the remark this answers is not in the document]
asangma: It's not ideal, but it should be left up to the user and the content.
In most cases, a list should consist of homogenous content.

bstifle:
Looks good Mike, let's ship this to the moon

julio8a:
We need a code review

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
```diff
@@ -1,74 +1,81 @@
 :host {
-  @extend %component-host;
-  @apply items-stretch
+	@extend %component-host;
```
asangma, line 2: We should probably replace this with any relevant styles from this partial.

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
```diff
@@ -1,74 +1,81 @@
...
+	@apply items-stretch
     flex
     text-color-1
     shadow-border-bottom
     m-0
     mb-px;
-  transition: background-color $transition;
-  animation: calcite-fade-in $transition;
+	transition: background-color $transition;
```
asangma, line 9: Are we able to use values from the Tailwind config here?

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
```diff
@@ -1,74 +1,81 @@
...
+	animation: calcite-fade-in $transition;
```
asangma, line 10: Are we able to use values from the Tailwind config here?

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
```diff
@@ -1,74 +1,81 @@
...
 }
 
-.label {
-  @apply bg-transparent flex flex-auto px-3 py-2 items-center cursor-pointer focus-base;
-  &:focus {
-    @apply focus-inset;
-  }
+:host([selected]) .icon-dot:before,
+:host([selected]) .icon calcite-icon {
+	transition: opacity $transition;
```
asangma, line 47: Are we able to use values from the Tailwind config here?

asangma (dismissed):
One change request.

The Sass var comments might need to be handled in a separate PR if a value needs to be added to the Tailwind config.

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
asangma: Host partial is [here](https://github.com/Esri/calcite-components/blob/master/src/assets/styles/_host.scss).

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
rmstinson: For the transition?

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
rmstinson: Handling in a separate PR.

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
rmstinson: Handling in a separate PR.

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
rmstinson: Handling in a separate PR.

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
rmstinson: Handling in a separate PR.

PR #2503, opened 2021-07-09
refactor(calcite-label): stop spreading attributes to the label element

eriklharper:
**Related Issue:** #1579

### DIFF
--- a/src/components/calcite-checkbox/calcite-checkbox.e2e.ts
+++ b/src/components/calcite-checkbox/calcite-checkbox.e2e.ts
@@ -209,16 +209,13 @@ describe("calcite-checkbox", () => {
     const input = await page.find("input");
     const label = await page.find("label");
 
-    expect(calciteCheckbox).not.toHaveAttribute("checked");
     expect(await calciteCheckbox.getProperty("checked")).toBe(false);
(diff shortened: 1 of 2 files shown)


src/components/calcite-checkbox/calcite-checkbox.e2e.ts
```diff
@@ -209,16 +209,13 @@ describe("calcite-checkbox", () => {
...
-    expect(calciteCheckbox).not.toHaveAttribute("checked");
     expect(await calciteCheckbox.getProperty("checked")).toBe(false);
-    expect(input).not.toHaveAttribute("checked");
     expect(await input.getProperty("checked")).toBe(false);
 
     await label.click();
 
     await page.waitForChanges();
 
-    expect(calciteCheckbox).toHaveAttribute("checked");
```
eriklharper, line 221: These lines were causing problems, saying the checked attribute wasn't getting applied, so I removed them, because we want to check the property rather than the attribute anyway.

eriklharper: Also verified that the behavior was correct in the browser with this change.

driskull (approved):
👍

Issue #2411, opened 2021-06-29
Enhancement: Animation options


tjameswhite:
### Description
Update _animation.scss to provide reusable animations for elements. 

Add timing to `root` to allow users to customize timing.
Add classes for animation trigger, and named animations. 

```
:root {
  --calcite-animation-timing: 1s;
}

// allows animation to be trigger via JS by adding class to element
.calcite-animate { 
  opacity: 0;
  animation-fill-mode: both;
  animation-duration: var(--calcite-animation-timing);
}

// specify which animation
.calcite-animate__in {
  animation-name: in;
}

.calcite-animate__in-down {
  animation-name: in-down;
}

.calcite-animate__in-up {
  animation-name: in-up;
}

.calcite-animate__in-scale {
  animation-name: in-scale;
}
```

### Sample Use
Animate card in-up on page load: 
```
<div class="card calcite-animate calcite-animate__in-up" >
        <h3>Hello world</h3>
        <p>Far far away, behind the word mountains, far from the countries Vokalia and Consonantia, there live the blind
            texts.</p>
</div>
```

caripizza:
Installed, assigning to @tjameswhite for verification - the classes and global CSS var now appear in the latest release on master (`1.0.0-next.258`).

PR #2504, opened 2021-07-09
feat: Add animation classes.

tjameswhite:
**Related Issue:** #2411 

## Summary
Animation keyframes exist in Calcite, but they are not available via any classes. 

Add `.calcite-animate` and related classes
Add `:root {--calcite-animation-timing: 0.6s;}` 

Remainder of implementation is up to end user.

### DIFF
--- a/src/assets/styles/_animation.scss
+++ b/src/assets/styles/_animation.scss
@@ -39,3 +39,31 @@
     transform: scale3D(1, 1, 1);
   }
 }
+
+:root {
+  --calcite-animation-timing: 0.6s;
+}
+
+// allows animation trigger vis JS by adding class to element
+.calcite-animate {
+  opacity: 0;
+  animation-fill-mode: both;
+  animation-duration: var(--calcite-animation-timing);
+}
+
+// specify animation
+.calcite-animate__in {
+  animation-name: in;
+}
+
+.calcite-animate__in-down {
+  animation-name: in-down;
+}
+
+.calcite-animate__in-up {
+  animation-name: in-up;
+}
+
+.calcite-animate__in-scale {
+  animation-name: in-scale;
+}


driskull (approved):
This looks good to me. @julio8a @asangma what do you think?

julio8a:
@macandcheese, can you take a look?

macandcheese:
I don’t mind the idea of opt in classes, I’ll  defer to @bstifle on the actual transitions and speeds.

I guess my only thought is that this would perhaps sit better adjacent to the tailwind classes generated in the calcite-styles repo than here. That’s still not available so maybe here is fine in the interim, but as of now no components use these internally - I think long term these belong elsewhere and CC can just consume them. That way folks can use these as they would a design token or tw class.

driskull:
>  perhaps sit better adjacent to the tailwind classes

Can we move these to tailwind? That would make more sense if we have a tokens repo eventually.

macandcheese:
> > perhaps sit better adjacent to the tailwind classes
> 
> Can we move these to tailwind? That would make more sense if we have a tokens repo eventually.

Yeah that makes more sense to me. Either in the tailwind config (that eventually gets generated as class names for folks to use w/o CC), or in that tokens / styles repo.

tjameswhite:
For what its worth -- timing was set by @cebender for our needs on esri.com. Most components in AEM have an animation on them, so that is our use case. We need to complete an audit of current animations, but the plan is to move to CC animation soonish. (I was actually looking to other css animation libraries when I found this already in CC.) One reason I used the css custom property for timing was to provide flexibility if others want to adjust.

caripizza:
Hi @tjameswhite - just following up from my Teams message. Julio clarified to me yesterday that yes, we would like to add this work to calcite-components. (It will be some time before we get the Tailwind npm package ready to be consumed in calcite-components, so this is a better home for the time being.)

@julio8a can you add Tim to the [GitHub team](https://github.com/orgs/Esri/teams/calcite-components/members), so he can clone the repo (non-fork) and open a new pr?

julio8a:
Tim has write access to the repo.

caripizza:
Closing in favor of pr #2607

PR #2509, opened 2021-07-10
refactor(calcite-link): stop spreading attributes to child a or span tag

eriklharper:
**Related Issue:** #1579


src/components/calcite-link/calcite-link.tsx
```diff
@@ -75,21 +81,21 @@ export class CalciteLink {
...
         <Tag
-          {...attributes}
           class={{ [CSS_UTILITY.rtl]: dir === "rtl" }}
           dir={dir}
           href={Tag === "a" && this.href}
           ref={this.storeTagRef}
+          rel={Tag === "a" && this.rel}
           role={role}
           tabIndex={tabIndex}
+          target={this.target}
```
driskull, line 98: Should target only get passed to "a" tags?

driskull (approved):
👍

src/components/calcite-link/calcite-link.tsx
eriklharper: thanks, good catch!

Issue #2505, opened 2021-07-09
Enhancement: calcite-modal - update demo html to achieve parity with Figma designs


ManahariDahal:
### Description
This issue aims to achieve parity between the calcite-checkbox following the similar layout on the Figma HTML Demo page and any usage snippets we provide.

### Acceptance Criteria 
- Modal design should encompass the Figma HTML demo page

### Relevant Info
- The calcite-modal demo page should have all the variants

#### Which Component
calcite-modal

ManahariDahal:
Ready to verify!

julio8a:
verified on master

PR #2516, opened 2021-07-12
test(modal): update demo html to match Figma

ManahariDahal:
**Related Issue:** #2505

## Summary
- updated _calcite-modal_ encompassing the Figma design
- this version includes _small_, _medium_, and _large_ **width & scale**, which the previous version failed to include

Screen Shot 2021-07-12 at 4 38 12 PM


julio8a:
We might need to update the layout a bit to avoid confusion with the width scales and component scales.

This grouping just looks and feels confusing since the medium and large are far from the column header (width):
image

Maybe something like this, just using negative space for grouping:
image


Thoughts @bstifle?

bstifle:
@ManahariDahal 

modal is a weird one since it uses SML for two properties. Might be more clear if we spell it out like this.

Additionally, i think showing SML scale is more beneficial than SML widths for the second section

image

julio8a:
Nice, I like the specific label with `Width x` 

That makes it much easier to read and understand the props

ManahariDahal:
> @ManahariDahal
> 
> modal is a weird one since it uses SML for two properties. Might be more clear if we spell it out like this.
> 
> Additionally, i think showing SML scale is more beneficial than SML widths for the second section
> 
> image

This design is way better than the one I had. I basically have to rewrite the code for this, haha. I will work on it later today.

bstifle:
@ManahariDahal  looks like the modals being invoked dont match the labels

let's get on a call and iron this out!

bstifle (approved):
WELL DONE 🍔

julio8a:
LGTM!

ManahariDahal:
> Looks like you have some failing checks. Can you try merging master into your branch?

All checks have passed.

julio8a:
LGTM!

PR #2518, opened 2021-07-12
chore(deps): set up automated dependency updates

jcfranco:
**Related Issue:** N/A

## Summary

This sets up a custom GitHub action to generate a PR with updated NPM dependencies. It is currently scheduled to run twice a month, but we can tweak this as needed. Not sure if we need more frequent updates than that.


driskull (approved):
👍 💯

Issue #2508, opened 2021-07-09
Enhancement: calcite-checkbox - update demo html to achieve parity with Figma


ManahariDahal:
### Description
This issue aims to achieve parity between the calcite-checkbox following the similar layout on the Figma HTML Demo page and any usage snippets we provide.

### Acceptance Criteria 
- Checkbox design should encompass the Figma HTML demo page to achieve parity

### Relevant Info 
- The calcite-checkbox demo page should have all the variants

#### Which Component
calcite-checkbox

ManahariDahal:
Ready to verify!

julio8a:
verified on master

PR #2528, opened 2021-07-13
test(checkbox): update demo html to match Figma

ManahariDahal:
**Related Issue:** #2508

## Summary
- updated calcite-checkbox encompassing the Figma design
- this version includes small, medium, and large _wrapped calcite label_ variants


julio8a (dismissed):
Focus states: I don't think we need the `focused` variations as they are not actually focused. i.e. unchecked focused is the same as unchecked. The way to check focus is just to focus on the unchecked version. Same for others like checked focused, indeterminate unchecked focused, etc.

Aside from ^, this looks good to me!

eriklharper:
> Focus states: I don't think we need the `focused` variations as they are not actually focused. i.e. unchecked focused is the same as unchecked. The way to check focus is just to focus on the unchecked version. Same for others like checked focused, indeterminate unchecked focused, etc.
> 
> Aside from ^, this looks good to me!

Why wouldn't we want to capture what focusing looks like?  IMO this would be a useful thing for Screener to capture once we get it running against demo pages.

julio8a:
Maybe I missed something? @ManahariDahal, is there a difference between the focused ones? on load, visually, they look the same as the ones without the focused label.

eriklharper:
> Maybe I missed something? @ManahariDahal, is there a difference between the focused ones? on load, visually, they look the same as the ones without the focused label.

My bad.  I'm forgetting that we converted the `focused` prop/attribute to a private state variable so applying `focused` doesn't do anything anymore.  Those examples should probably just get removed from the demo HTML and down the road when we get Screener setup to test against demo pages, we'll use the Screener Steps API to programmatically focus all the different variants and take screenshots of each in order to capture them properly.

The focused state for most if not all of the calcite designs is a blue focus ring 
image

julio8a:
That sounds great!

ManahariDahal:
> Focus states: I don't think we need the `focused` variations as they are not actually focused. i.e. unchecked focused is the same as unchecked. The way to check focus is just to focus on the unchecked version. Same for others like checked focused, indeterminate unchecked focused, etc.
> 
> Aside from ^, this looks good to me!

removed the focused states

julio8a (approved):
LGTM!

src/demos/calcite-checkbox.html
```diff
@@ -32,846 +33,750 @@
...
+      **************************************************
+    -->
+    <div class="parent">
+      <div class="child right-aligned-text">wrapped in native label with for</div>
+
+      <!-- small -->
+      <div class="child">
+        <label for="light-wrapped-in-label-with-for">
+          <calcite-checkbox
+            id="light-wrapped-in-label-with-for"
```
caripizza, line 716: Can you make the checkbox ids and (wrapping label's `for` attributes) unique to each checkbox?

src/demos/calcite-checkbox.html
```diff
@@ -32,846 +33,750 @@
...
+      **************************************************
+    -->
+    <div class="parent">
+      <div class="child right-aligned-text">sibling calcite-label</div>
+
+      <!-- small -->
+      <div class="child">
+        <calcite-label for="light-sibling-calcite-label"
+          >sibling calcite-label
+          <calcite-checkbox id="light-sibling-calcite-label" name="sibling-calcite-label" scale="s"></calcite-checkbox>
```
caripizza, line 650: Can you  make the ids unique here as well?

bstifle:
1.  let's remove the `hovered` versions from the demo page, as we can just see those when interacting with them
2. ship it!

julio8a:
I think leaving the hover will be good for Screener to check it as part of the automated testing.

eriklharper:
> I think leaving the hover will be good for Screener to check it as part of the automated testing.

Hover testing is also something that Screeners Step API supports doing in tests which will be a more robust way of testing hover states since it will be testing against an actual browser hover event instead of just applying a custom `hovered` attribute.  

image

https://github.com/screener-io/screener-runner#testing-interactions

julio8a:
Okay, let's remove the hover states and leverage screener

bstifle (approved):
nice work man

ManahariDahal:
sweet! should I go ahead and merge then? @julio8a

julio8a:
Go for it!

Issue #2040, opened 2021-04-26
Enhancement: pick/value-list and pick/value-list-item add nonInteractive prop


asangma:
### Description
Add a disableSelect property which disables select.
There are cases where value-list-items benefit from the ability to sort and quickly remove but where the UX suffers from an irrelevant selected state.

#### Which Component
ValueList and ValueListItem

cc @driskull

TheBlueDog:
Field Maps has a use for this. We have a list of fields in the layer that will never need a selected state.

asangma:
@jcfranco not sure who to assign to for verfee since I created the issue. :)

jcfranco:
No prob. I can help verify. 😄

jcfranco:
Verified!

PR #2539, opened 2021-07-14
feat(value-list-item, pick-list-item): add disableSelect prop

asangma:
**Related Issue:** #2040

## Summary
Adds a disableSelect prop that disables selection of the item

The issue only calls for an enhancement to ValueListItem but in order to add the feature, it appeared that the update needed to also be done on PickListItem.


src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -49,6 +49,11 @@ export class CalcitePickListItem {
    * When false, the item cannot be deselected by user interaction.
    */
   @Prop() disableDeselect = false;
+  
+  /**
+   * When true, the item cannot be selected by user interaction.
+   */
+  @Prop({ reflect: true }) disableSelect = false;
```
driskull, line 56: The name of this prop is a bit confusing. Right above it we have `disableDeselect` and this is `disableSelect`. It sounds like they would conflict? Any way these can be merged or named differently?

I thought for naming we had proposed names like `selectDisabled`?

driskull (commented):
Comments

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
asangma: Actually, thinking about it now, we should probably use what we used in ListItem: `nonInteractive`. Yeah?

driskull: yeah

asangma: @driskull Updated.

driskull (approved):
This looks good to me. 🌈

asangma:
Thanks @driskull.

Issue #2327, opened 2021-06-17
Bug: calcite-input-time-picker - input box should support 12 hour format


AdelheidF:
### Actual Behavior

`hourDisplayFormat` seems to only work for the little picker popup, not the main input box. This is very confusing for people that are not used to the 24 hour format. 

![image](https://user-images.githubusercontent.com/44481292/122468810-08279d00-cf71-11eb-8a87-19ee8a6c1e3f.png)
 
When I see this I would definitely think it’s 5:55 in the morning.

![image](https://user-images.githubusercontent.com/44481292/122468834-107fd800-cf71-11eb-9c98-284e5da825b8.png)

This is probably a showstopper for using it in Map Viewer.

### Expected Behavior

The input box should also show the 12 hour format with AM/PM if `hourDisplayFormat` is set to 12.

### Reproduction Steps or Sample

1.

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

AdelheidF:
Looks good, tested with next.245.

PR #2540, opened 2021-07-14
fix(input-time-picker): input box supports 12-hour display

eriklharper:
**Related Issue:** #2327 

This updates the text input to display the full time in correct 12-hour format including the meridiem.  You can still type in 24-hour format when using the 12-hour mode, but on input blur, it will format it to the correct 12-hour version.  Typing `00` in the `time-picker` popup's hour input when using the 12-hour format displays `12` instead of `00` now.


jcfranco:
@eriklharper Awesome! Before reviewing, can you add more info to the description on what this PR encompasses?

jcfranco (approved):
🎉 🌮 🕐

PR #2543, opened 2021-07-15
fix(dropdown): Fix dropdown display issues.

driskull:
**Related Issue:** N/A

## Summary

fix(dropdown): Fix dropdown display issues.

### DIFF
--- a/src/components/calcite-dropdown/calcite-dropdown.tsx
+++ b/src/components/calcite-dropdown/calcite-dropdown.tsx
@@ -125,6 +125,8 @@ export class CalciteDropdown {
     this.items = Array.from(
       this.el.querySelectorAll<HTMLCalciteDropdownItemElement>("calcite-dropdown-item")
     );
+
+    this.reposition();
   }
 
   disconnectedCallback(): void {
--- a/src/components/calcite-dropdown/calcite-dropdown.tsx
+++ b/src/components/calcite-dropdown/calcite-dropdown.tsx
@@ -223,9 +223,13 @@ export class CalciteDropdown {
 
   @Listen("calciteDropdownOpen", { target: "window" })
   closeCalciteDropdownOnOpenEvent(e: Event): void {
-    if (e.target !== this.el) {
-      this.active = false;
+    const composedPath = e.composedPath();
+
+    if (composedPath.includes(this.el)) {
+      return;
     }
+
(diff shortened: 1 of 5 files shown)


src/components/calcite-dropdown/calcite-dropdown.tsx
```diff
@@ -223,9 +223,13 @@ export class CalciteDropdown {
 
   @Listen("calciteDropdownOpen", { target: "window" })
   closeCalciteDropdownOnOpenEvent(e: Event): void {
-    if (e.target !== this.el) {
-      this.active = false;
+    const composedPath = e.composedPath();
+
+    if (composedPath.includes(this.el)) {
```
jcfranco, line 228: Nitpick: `e.composedPath().includes(this.el)` since the path is only referenced once.

jcfranco (approved):
👍🏼

caripizza:
Thanks @driskull for fixing that split-button dropdown issue so quickly! Do you know of any way we can add test coverage for the fix or what are your thoughts on that?

driskull:
I think we should have a story that shows the dropdown opened when its set to auto width. I think that would have caught this.

driskull:
@jcfranco can you review again? I'd like to update the stories to have at least one dropdown open by default. This means adding a prop to the split-button.

src/components/calcite-split-button/calcite-split-button.tsx
```diff
@@ -21,6 +21,12 @@ export class CalciteSplitButton {
   /** is the control disabled  */
   @Prop({ reflect: true }) disabled?: boolean;
 
+  /**
+   * Is the dropdown currently active or not
+   * @internal
```
driskull, line 26: I can make this public in a separate PR or we could do it at a later time.

jcfranco: 👍  to revisit making it public later.

jcfranco (approved):
👍🏼

jcfranco:
> Do you know of any way we can add test coverage for the fix or what are your thoughts on that?

For anything that specifically needs to test `composedPath` in event handling, we'll need some sort of shadow DOM wrapper util. We can bring back the e2e test page setup util from `calcite-app-components` to make it easy to use.

Issue #2357, opened 2021-06-22
Bug: color-picker hue thumb cannot be nudged via keyboard when color is #000 or #fff


jcfranco:
### Actual Behavior

When the color value is #fff or #000, tabbing to the hue handle will not move when the arrow keys are pressed. This works as expected when the color is another value.

### Expected Behavior

The hue handle can be nudged regardless of color value.

### Reproduction Steps or Sample

1. Initialize the calcite-color-picker with pure white.
2. Tab to the hue handle 
3. Press left/right or up/down and notice you can't move the handle

### Relevant Info

Stems from https://github.com/Esri/calcite-components/issues/2230#issuecomment-860439982

_Version_: `@esri/calcite-components@1.0.0-beta.56`

- [ ] CDN
- [x] NPM package

jcfranco:
Installed!

jcfranco:
@julio8a Can you help verify this one? You can use https://codepen.io/jcfranco/pen/poPPwQq?editors=1000

julio8a:
verified!

PR #2551, opened 2021-07-16
fix(color-picker): fix nudging hue slider with left/right arrows when color is #000 or #fff

jcfranco:
**Related Issue:** #2357 

## Summary

This addresses an issue caused by the scope key-nudging logic never updating the current hue value when 0. Additionally includes some minor cleanup.


driskull (approved):
👍 🏅

jcfranco:
Screener failure is unrelated. Merging...

Issue #2437, opened 2021-07-02
Bug: input-enabled calcite-tree not removing items from selected


joeyHarig:
### Actual Behavior
Items are not removed from the selected array (returned from `calciteTreeSelect` -> `event.detail.selected`) once they have been unchecked.

### Expected Behavior
When `input-enabled` is true and items are unselected, the items should be removed from the array: event.detail.selected.

### Reproduction Steps or Sample
https://codepen.io/joeyHarig/pen/eYWmpbm?editors=1111

### Relevant Info

_Version_: latest

jcfranco:
@joeyHarig I started looking into this and noticed that it works correctly if you press the shift key when clicking items. You can use this as a workaround in the meantime. Also, note that the selection mode should be `multi` and not `multiple`.

@paulcpederson Since users have an expectation of how checkboxes should behave, should we always toggle the value when clicking an input-enabled tree item regardless of whether the shift key was pressed or not?

https://codepen.io/jcfranco/pen/PompyrW?editors=1111 (based off the original test case)

paulcpederson:
@jcfranco this is a bug. This worked fine in beta.58 https://codepen.io/paulcp/pen/mdmWQRW 

I'm not sure what changed but they should work like checkboxes.

jcfranco:
Thanks for the info! The regression is probably related to my previous changes. It's worth noting that the event payload selection issue is present in beta.58 too. Will fix! 💪🏼 🛠️

jcfranco:
Installed!

jcfranco:
@joeyHarig Can you help verify once this gets deployed to `@next`? 🙇🏼‍♂️

joeyHarig:
@jcfranco Just verified that the issue has been resolved.

jcfranco:
Awesome, thanks!

PR #2555, opened 2021-07-16
fix(tree): fix item selection when multi + input-enabled

jcfranco:
**Related Issue:** #2437 

## Summary

This fixes an issue where clicking a tree-item's checkbox would not toggle the selection when a tree is in multi selection mode with input-enabled.

@paulcpederson Can you help test this one? I'm not clear if other selection modes + input-enabled need to work differently. The ancestor selection-mode + input-enabled from the demo page appear to work fine. I added an E2E test to start keeping track of this.

### DIFF
--- a/src/components/calcite-tree-item/calcite-tree-item.tsx
+++ b/src/components/calcite-tree-item/calcite-tree-item.tsx
@@ -191,7 +191,7 @@ export class CalciteTreeItem {
     }
     this.expanded = !this.expanded;
     this.calciteTreeItemSelect.emit({
-      modifyCurrentSelection: (e as any).shiftKey,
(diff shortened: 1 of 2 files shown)


paulcpederson:
@jcfranco I have never understood the other selection modes or what their purpose is. Checked locally, though, and ancestors mode looks great 👍🏻

paulcpederson (approved):

jcfranco:
> I have never understood the other selection modes or what their purpose is. 

😭  I thought you would have all the secrets! Thanks for testing! 

@patrickarlt Checked the blame view and saw you introduced the prop. Could you share some info on the different selection modes? It would greatly help with doc and setting up tests.

Issue #2459, opened 2021-07-07
Bug: Slotting a calcite-action-menu inside a calcite-action-bar or calcite-action-pad display issues


driskull:
### Actual Behavior

Slotting a calcite-action-menu inside a calcite-action-bar or calcite-action-pad display issues.

The slotted action has a small margin which adds a "gutter" to one side of the action pad.

### Expected Behavior

The slotted action of the action-menu doesn't have a gutter to one side of the action-bar/pad.

### Reproduction Steps or Sample

```html
<calcite-action-bar slot="action-bar" class="calcite-theme-dark">
              <calcite-action-group>
                <calcite-action text="Add">
                  <calcite-icon icon="plus" scale="s"></calcite-icon>
                </calcite-action>
                <calcite-action text="Save" disabled>
                  <calcite-icon icon="save" scale="s"></calcite-icon>
                </calcite-action>
                <calcite-action text="Layers" active>
                  <calcite-icon icon="layers" scale="s"></calcite-icon>
                </calcite-action>
                <calcite-action-menu>
                  <calcite-action slot="trigger" text="Add" icon="banana"></calcite-action>
                  <calcite-action text="Add" icon="plus" text-enabled></calcite-action>
                  <calcite-action text="Add" icon="plus" text-enabled></calcite-action>
                  <calcite-action text="Add" icon="plus" text-enabled></calcite-action>
                </calcite-action-menu>
              </calcite-action-group>
              <calcite-action-group>
                <calcite-action text="An unbelievably long title">
                  <calcite-icon icon="plus" scale="s"></calcite-icon>
                </calcite-action>
                <calcite-action text="Save" disabled>
                  <calcite-icon icon="save" scale="s"></calcite-icon>
                </calcite-action>
                <calcite-action text="Layers">
                  <calcite-icon icon="layers" scale="s"></calcite-icon>
                </calcite-action>
              </calcite-action-group>
            </calcite-action-bar>
```

### Relevant Info

![image](https://user-images.githubusercontent.com/1231455/124806705-d96a6a00-df11-11eb-8796-cc548f507297.png)

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

asangma:
@caripizza @jcfranco This would be a minor fix but the bug causes a pretty noticeable UI bug.

jcfranco:
Added to the next sprint. Will move to this one if the queue clears up.

asangma:
[...]

PR #2557, opened 2021-07-16
fix(action-menu): target 1px margin more accurately to actions in menu

asangma:
**Related Issue:** #2459

## Summary
More accurately targets the 1px margin style to the actions in side the menu.
The 1px margin is to accommodate the focus ring.


julio8a (approved):
Nice!

asangma:
Thanks @julio8a !

Issue #834, opened 2020-08-10
Enhancement: Slider Histogram Color Stops


gpbmike:
### Description

Site Scan is replacing in house UI components with Calcite where possible. The slider with histogram is one we would like to use, but our implementation allows for adding color stops (see example use case).

### Acceptance Criteria 

See example use case.

### Relevant Info 

Currently we use the `<AreaChart>` component from Recharts to provide this visualization and sync it to our slider. We can continue to do that with the Calcite slider, but thought it would look better if provided by calcite.

#### Which Component

Slider with histogram

#### Example Use Case

Elevation data using the "turbo" palette.

![image](https://user-images.githubusercontent.com/8754/89807659-ce367080-daed-11ea-958f-697d93eb86b6.png)

Thermal data using the "inferno" palette. 

![image](https://user-images.githubusercontent.com/8754/89807697-ddb5b980-daed-11ea-8561-ec6dae63d186.png)

Color stops update to reflect changes in slider positions.

![image](https://user-images.githubusercontent.com/8754/89807737-ef975c80-daed-11ea-85e3-cc2e90e0bd8b.png)

julio8a:
@gpbmike, do you think this is something you can help us implement?

gpbmike:
> @gpbmike, do you think this is something you can help us implement?

Sure I can give it a go.

julio8a:
I'll add it to the coming sprint and assign to you.

gpbmike:
This work is ready for review across two PRs (#2562, #2568). I'm blocked by lack of permissions in the calcite-components repository. If someone with permission would like to use my work, that's A-OK with me.

julio8a:
@gpbmike, you should be good on the permissions now, can you check?

jcfranco:
Installed!

julio8a:
verified on master

PR #2562, opened 2021-07-16
feat(graph): Add support for color stops

gpbmike:
**Related Issue:** #834

## Summary

Adds support for color stops for the `<calcite-graph>` component. A subsequent PR will add color stop support to the `<calcite-slider>` component.

```tsx
/**
 * Array of values describing a single color stop ([offset, color, opacity])
 * These color stops should be sorted by offset value
 */
@Prop() colorStops: [number, string, number][];
```
Color stops are added via a `<linearGradient>` element and apply along the x-axis.

image

- [x] feature or fix has a corresponding test
- [ ] changes have been tested with demo page in Edge

## Discussion question
This is pretty rudimentary support for adding a linear gradient with color stops. It solves the use case in the related issue, but are there other, more complex, use cases that we want to be prepared to support in the future?


gpbmike:
Replaced by #2572

Issue #2009, opened 2021-04-21
Enhancement: calcite-input type=number - only allow integer inputs


AdelheidF:
### Description

We have places where we use calcite-input and only allow integer values. It would be nice if calcite-input took care of this and not allow decimal separators. Otherwise we have to check the value in the event and 'fix' it accordingly everywhere we use it.

It would be great if it would stop you from doing that, even though once you hit enter it becomes a rounded number again.
![image](https://user-images.githubusercontent.com/44481292/115594986-3ba8bc80-a28b-11eb-822d-6b1a89b8b8b6.png)

jcfranco:
Verified with https://codepen.io/jcfranco/pen/qBmYONO

@eriklharper Please avoid closing issues without verification.

eriklharper:
> Verified with https://codepen.io/jcfranco/pen/qBmYONO
> 
> @eriklharper Please avoid closing issues without verification.

Did I miss the memo on that?  Should we be creating codepens for every resolved issue now?

AdelheidF:
Verfied as well with next.261, looks good.

> Please avoid closing issues without verification.

Yes, please don't close issues that I open before I (myself) verified them. I'm happy to do the verification. 
I was just out of office for a bit this time.

ethanbdev:
@eriklharper I am seeing that with `step="any"` a decimal also cannot be inputted. This doesn't seem right, what is intended?

https://codepen.io/mpayson/pen/XWRxXBr

AdelheidF:
We can't upgrade to beta.61 because of this.

babakbolour:
@AdelheidF just surfing the repo and saw this one!  the screenshot at the beginning comment shows RGB values, but I assume that  is just one application of the component. Can Calcite-input accept numbers bigger than 999? and if yes, then what is the behavior of this component, when an international user enters a number like this: "12 345" or "12'345"? in other words does the component accept different 'thousand separators symbols'? and hope there is not crash, if they do enter such number.

AdelheidF:
> does the component accept different 'thousand separators symbols'?

yes, we tested that

eriklharper:
@babakbolour correct.  The input component accepts the locale-appropriate thousand separators.

eriklharper:
> @eriklharper I am seeing that with `step="any"` a decimal also cannot be inputted. This doesn't seem right, what is intended?
> 
> https://codepen.io/mpayson/pen/XWRxXBr

[...]

PR #2563, opened 2021-07-17
feat(calcite-input): allow integer only input when step is a whole number

eriklharper:
**Related Issue:** #2009

This PR adds support for allowing only integer values to be entered when `step` is a whole number.


src/utils/number.ts
```diff
@@ -1,5 +1,9 @@
 import { numberKeys } from "./key";
 
+export function isValidDecimal(value: number): boolean {
+  return Number(value) % 1 != 0;
```
jcfranco, line 4: I think this can be simplified to `!Number.isInteger(value)`.

jcfranco (approved):
Can you add a test before merging? Other than that, this LGTM!

src/utils/number.ts
eriklharper: Man you really do know all the nooks and crannies of the browser API!  Thanks! :)

Issue #2422, opened 2021-07-01
Bug: Conflicting class names in scoped components


joeyHarig:
### Actual Behavior
The `<div/>` rendered inside of `calcite-radio-button` has a class of "container". This is a very common class name in many apps and may cause conflicting style declarations. This is the case in ArcGIS Monitor. 

Can we change the class name to something more specific: "calcite-radio-button-container"?

<img width="500" alt="Screen Shot 2021-07-01 at 9 27 22 AM" src="https://user-images.githubusercontent.com/55154968/124151959-50a28880-da50-11eb-84a6-7a300c2637da.png">

**Updated by @jcfranco**: This should namespace classes used by all scoped components:

* `calcite-inline-editable`
* `calcite-input`
* ~`calcite-label`~ – only uses RTL util class
* `calcite-radio-button`

### Relevant Info

_Version_: `@esri/calcite-components@1.0.0-beta.58`

driskull:
@eriklharper more reason to make all of our components shadowed :D

eriklharper:
I'm surprised Stencil's scoped feature doesn't BEM-ify or randomize that class name.  We can totally remove that classname and just target the `div` directly in the CSS as a fast fix.

jcfranco:
This issue should tackle namespacing CSS classes used by all scoped components. Fortunately, there's only 4 at the moment. I've updated the description.

We'll tackle moving these to shadow DOM in a separate issue.

driskull:
Maybe we can shadowDom them next sprint?

jcfranco:
Installed! 

@joeyHarig Can you help verify this one once `@next` updates?

joeyHarig:
@jcfranco Verified! Thanks for your help

jcfranco:
No problem and thank **you** for verifying. 😄

PR #2564, opened 2021-07-17
fix(inline-editable, input, radio-button): namespace scoped CSS classes

jcfranco:
**Related Issue:** #2422

## Summary

This PR namespaces classes (following [BEM](http://getbem.com/)) of currently scoped components since these are not transformed by the Stencil build and are available in the light DOM. 

This is a stopgap solution until all scoped components use shadow DOM.


caripizza (approved):
lgtm! 👍

driskull (approved):
👍

Issue #1958, opened 2021-04-13
Bug: Button - Remove use of document.getElementsByName()


driskull:
### Summary

Remove use of document.getElementById since it won't work if the component is within another shadow dom component. It seems like other teams are wrapping our components in their shadow dom components so we need to be mindful of this.

The solution is to use this.el.getRootNode() and then query either the rootNode (document) or rootNode.host (shadow dom) element.

### Actual Behavior

```jsx
const targetForm = requestedForm
        ? (document.getElementsByName(`${requestedForm}`)[0] as HTMLFormElement)
        : (this.el.closest("form") as HTMLFormElement);
```

### Expected Behavior

```jsx
import { getElementById, getRootNode } from "../../utils/dom";

const targetForm =  rootNode instanceof ShadowRoot
          ? rootNode.host.querySelector(`[name=${requestedForm}]`)
          : document.getElementsByName(`${requestedForm}`)[0] as HTMLFormElement
```

### Reproduction Steps

1. Use a `calcite-button` with a form inside the shadow dom of another component
2. Notice that the `form` functionality is not working as expected.

### Relevant Info 

https://github.com/Esri/calcite-components/pull/1956

driskull:
Can we get this issue elevated so that we use the `queryElementsRoots` in dom utils?

What do you think @jcfranco Should this be a priority since it uses document and would fail inside a shadowRoot.

jcfranco:
Added to the next sprint. Will move to this one if the queue clears up.

driskull:
Think this should be pre 1.0

caripizza:
I can verify this one - @driskull can you suggest a code snippet for me to confirm the fix?

driskull:
@caripizza you will have to setup a page with a shadowed component that uses the button.

```html
<script>
      class DemoShadow extends HTMLElement {
        constructor() {
          super();
          this.attachShadow({ mode: "open" });
          this.shadowRoot.innerHTML = `
          <form id="myform">
            <calcite-button form="myform"></calcite-button>
          </form>
        `;
        }
      }
      customElements.define("demo-shadow", DemoShadow);
    </script>
```

```html
<demo-shadow></demo-shadow>
```

caripizza:
Verified the fix locally with this code on master branch (closest prior release is 1.0.0.next-259):
```html
    <script>
      class DemoShadow extends HTMLElement {
        constructor() {
          super();
          this.attachShadow({ mode: "open" });
[...]

PR #2574, opened 2021-07-20
fix(button): Remove use of document.getElementsByName()

driskull:
**Related Issue:** #1958

## Summary

fix(button): Remove use of document.getElementsByName() #1958


src/utils/dom.ts
```diff
@@ -32,7 +32,7 @@ export function getElementDir(el: HTMLElement): Direction {
 
 export function getElementProp(el: Element, prop: string, fallbackValue: any, crossShadowBoundary = false): any {
   const selector = `[${prop}]`;
-  const closest = crossShadowBoundary ? closestElementCrossShadowBoundary(selector, el) : el.closest(selector);
+  const closest = crossShadowBoundary ? closestElementCrossShadowBoundary(el, selector) : el.closest(selector);
```
driskull, line 35: @jcfranco I'm not seeing crossShadowBoundary being used in any components. Maybe we can remove this here?

Just to note: `getElementProp` doesn't cross shadow boundary. This is probably ok since I see it used in child components only so they shouldn't cross the shadowDom.

However, I see a lot of use of `this.el.closest()` in things like checkbox and input so it seems like we should update those to use `closestElementCrossShadowBoundary` Do we want an issue to prioritize this?

driskull: https://github.com/Esri/calcite-components/blob/19882020664c4354f3110367712cb496b5192253/src/components/calcite-input/calcite-input.tsx

https://github.com/Esri/calcite-components/blob/9e5c810532eec100fcb13e232a7949d61721bf11/src/components/calcite-checkbox/calcite-checkbox.tsx#L191

driskull: https://github.com/Esri/calcite-components/blob/19882020664c4354f3110367712cb496b5192253/src/components/calcite-input/calcite-input.tsx#L263

jcfranco (approved):
LGTM!

jcfranco:
> I'm not seeing crossShadowBoundary being used in any components. Maybe we can remove this here?

That was added to make the switch at some point to have the utils always look up across shadow boundaries. I'm guessing it wasn't a high priority since it didn't come up later. I think you brought up making this behavior the default too, right?

driskull:
> I think you brought up making this behavior the default too, right?

Yes, but now that I look how its being used I'm not sure that would be useful. This is used a lot to just get a parent item from within a child. I don't think we would want it to go cross shadow for these instances.

jcfranco:
Agreed. Probably why it wasn't enabled by default. We can revisit this later if we need it back.

PR #2575, opened 2021-07-20
fix(button, checkbox, input, radio-button): Support native forms across shadow boundary.

driskull:
**Related Issue:** N/A

## Summary

fix(button, checkbox, input, radio-button): Support native forms across shadow boundary.


eriklharper (approved):
Other than adding some tests, in particular one that tests the form submit functionality from within another shadowRoot, this looks good to me!  With this closestElementCrossShadowBoundary util we can now look at implementing full native form support whilst moving away from the rest of the scoped form components we have.

driskull:
This would submit a form across shadowRoot but i'm not sure if the form would be able to access the elements within the shadowRoot.

eriklharper:
> This would submit a form across shadowRoot but i'm not sure if the form would be able to access the elements within the shadowRoot.

That is true, the form will not find the native input elements inside shadowRoots, I was forgetting about that part.  Going full shadow on all our components would necessitate a `calcite-form` component.

Issue #2576, opened 2021-07-20
Enhancement: calcite-color-picker demo html update to achieve parity with Figma


ManahariDahal:
### Description
This issue aims to achieve parity between the calcite-color-picker following the similar layout on the Figma HTML Demo page and any usage snippets we provide.

### Acceptance Criteria 
- Color-picker design should encompass the Figma HTML demo page

### Relevant Info 
- The calcite-color-picker demo page should have all the variants

#### Which Component
_calcite-color-picker_

ManahariDahal:
ready to be verified @julio8a

julio8a:
verified on master

PR #2578, opened 2021-07-20
test(color-picker): updated calcite-color-picker demo page

ManahariDahal:
**Related Issue:** #2576 

## Summary
- updated calcite-color-picker encompassing the Figma design
- this version includes _small_, _medium_, and _large_ scale, which the previous version failed to include

### DIFF
--- a/src/demos/calcite-color-picker.html
+++ b/src/demos/calcite-color-picker.html
@@ -6,59 +6,174 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <title>Calcite ColorPicker</title>
     <style>
-      .demo-background-dark {
-        background: #202020;
-        color: white;
-        padding: 1rem;
+      .parent {
+        display: flex;
+        width: 60%;
+        padding: 25px 0;
+      }
+
+      .child {
+        flex: 1 0 25%;
+        margin: 0 25px;
+        color: var(--calcite-ui-text-3);
+        font-family: var(--calcite-sans-family);
+        font-size: var(--calcite-font-size-0);
+        font-weight: var(--calcite-font-weight-medium);
+      }
+
+      .right-aligned-text {
+        text-align: right;
+      }
+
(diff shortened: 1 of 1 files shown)


bstifle (approved):
NICE! perf

julio8a:
is this a bug with the component?
image
Should this one be empty (no color selected) based on the prop ^

Also, I don't see a visual difference between default, default (no color), with initial color (there than a different color selected), and store saved colors. 

Therefore, do we need these and just depend on default?

cc'ing @bstifle @jcfranco

jcfranco:
> Also, I don't see a visual difference between default, default (no color), with initial color (there than a different color selected), and store saved colors.

The no-color example requires users to clear the number to see a visual difference in the UI. We could update the code so that the color picker is empty by default: 

```html
<calcite-color-picker allow-empty value=""></calcite-color-picker>
```

The store-saved-colors example requires saving colors and reloading the page to see a visual difference.

ManahariDahal:
@julio8a this is ready to be reviewed.

julio8a:
Thanks! I don't think we need these:

* With initial color
  * The default starts with initial color
* Stored saved color
  * There are no stored colors on load

@jcfranco, is there a way to have this component start up some stored colors?

julio8a:
@jcfranco, any thoughts on the stored colors on load?

jcfranco:
I think it can be removed for now. The demo page would possibly have something for this use case once https://github.com/Esri/calcite-components/issues/748 is implemented. In the meantime, there is an E2E test that will cover it.

julio8a (changes requested):
@ManahariDahal, please remove `with initial color` and `stored saved color`

ManahariDahal:
> @ManahariDahal, please remove `with initial color` and `stored saved color`

oops, forgot to push the latest changes.

julio8a (approved):
Nice!

Issue #2531, opened 2021-07-13
Docs: calcite-block - description for slot=icon is wrong


AdelheidF:
### Description 

![image](https://user-images.githubusercontent.com/44481292/125530132-179b0c25-e73f-4cb6-91ff-02f5d9c9c199.png)

Though icon seems to appear at the leading side.
![image](https://user-images.githubusercontent.com/44481292/125530192-ed44cf92-b6b9-475b-8f40-4188376d172c.png)

julio8a:
installed. @AdelheidF, would you mind verifying this one?

julio8a:
self-verified on master

AdelheidF:
looks good, thanks

![image](https://user-images.githubusercontent.com/44481292/128051766-45ed908c-a82c-4dfa-b650-7e9bb0d54bfb.png)

PR #2586, opened 2021-07-21
docs(action-block): slot icon position is wrong

julio8a:
**Related Issue:** #2531

## Summary

Updated doc to leading, not trailing. This was likely caused by a typo.


asangma (approved):
LGTM

Issue #2580, opened 2021-07-20
Enhancement: calcite-date-picker demo html update to achieve parity with Figma


ManahariDahal:
### Description
This issue aims to achieve parity between the calcite-date-picker following the similar layout on the Figma HTML Demo page and any usage snippets we provide.

### Acceptance Criteria 
- Date-picker design should encompass the Figma HTML demo page

### Relevant Info 
- The calcite-date-picker demo page should have all the variants

#### Which Component
_calcite-date-picker_

ManahariDahal:
Ready to verify!

julio8a:
verified on master

PR #2594, opened 2021-07-21
test(date-picker): updated date-picker demo according to Figma

ManahariDahal:
**Related Issue:** #2580 

## Summary
- updated calcite-date-picker encompassing the Figma design
- this version includes _small_, _medium_, and _large_ scale, which the previous version failed to include

### DIFF
--- a/src/demos/calcite-date-picker.html
+++ b/src/demos/calcite-date-picker.html
@@ -6,187 +6,199 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <title>Calcite Date Picker</title>
     <style>
-      .demo-background-dark {
-        background: #202020;
-        color: white;
-        padding: 1rem 1rem 3rem 1rem;
+      .parent {
+        display: flex;
+        width: 60%;
+        padding: 25px 0;
       }
-      h2,
-      h3 {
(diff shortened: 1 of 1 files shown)


bstifle (approved):
LOOKS GOOD! nice job

bstifle:
@ManahariDahal  looks good, ship it!

julio8a:
@bstifle @caripizza. Do we need all of these? i.e. always open

bstifle:
@julio8a One should be as a dropdown, always open should be when just static on a page

julio8a:
ah, okay. Does it seem most of these need to have a button or something to trigger the calendar?

Are you guys okay with just not having the trigger (button, anchor, etc)? since we really want to see the component. This would mean removing the always open variant. (since the trigger, i.e. button is not part of this component)

bstifle:
@julio8a yeah if its not in dropdown it should just have that "always open" styling (border, no shadow)

julio8a:
So this one needs a few buttons to trigger the floating date picker?

ManahariDahal:
yes, I agree with @julio8a . I will get rid off the _always open_ state.

julio8a (approved):
Looks good!

PR #2596, opened 2021-07-21
fix(x-close input and inline): x hover ui-text-1

julio8a:
**Related Issue:** #2216

## Summary

`x` on input and inline editible hover state updated to ui-text-1


bstifle:
@julio8a inline editable is now ui-text-1 in all states.

should be ui-text-3 on idle.




input X is correct!

bstifle:
@julio8a 
image
the chevrons are kinda jacked up too

julio8a:
Bryan, as we discussed on teams, the `X` on the inline will be done in a separate branch since it needs a component change. Anything else on this one needed?

bstifle (approved):
lookin kln!

Issue #2605, opened 2021-07-22
Enhancement: calcite-scrim demo html update to achieve parity with Figma


ManahariDahal:
### Description
This issue aims to achieve parity between the calcite-date-picker following the similar layout on the Figma HTML Demo page and any usage snippets we provide.

### Acceptance Criteria 
- calcite-scrim design should encompass the Figma HTML demo page

### Relevant Info 
- It should have all the variants

#### Which Component
_calcite-scrim_

julio8a:
verified on master

PR #2606, opened 2021-07-22
test(scrim): updated demo design using flexbox with two rows

ManahariDahal:
**Related Issue:** #2605 

## Summary
- updated demo page
- used flexbox with two rows


julio8a (approved):

bstifle:
@ManahariDahal 
image


im seeing 2 of everything.... i prefer the presentation on top

julio8a (approved):
Nice, we might just need to one without the loading indicator, as Screener will always see this as a change when it takes a screenshot, but leave the loading there, for now, we can easily remove it later if needed.

ManahariDahal:
> Nice, we might just need to one without the loading indicator, as Screener will always see this as a change when it takes a screenshot, but leave the loading there, for now, we can easily remove it later if needed.

The screener test passed, but Travis CI didn't. Should I trigger Travis CI and we can leave the loading indicator if the screener passes again. What do you think?

julio8a:
Yes

bstifle (approved):
skrimtiful!

ManahariDahal:
Ready to merge!

PR #2607, opened 2021-07-22
feat(animation): add global animation classes, duration CSS Custom Property

caripizza:
**Related Issue:** #2411

## Summary
Adds the animation classes/CSS Var from @tjameswhite pr #2504. (He asked that I open it up from non-fork pr for expediency sake. For the CSS Var, I used the `$transition-medium` value from calcite-base - 300ms - so we can use the var in the Tailwind config.)

Also adds a new animation.html demo page and a new e2e test file for the global animation classes/CSS var.


driskull (approved):
👍 🎉

Issue #2582, opened 2021-07-20
Bug: disabled calcite-input + type=number allows interaction with the up/down buttons. 


jcfranco:
### Actual Behavior

The up/down buttons can be still be clicked when the input component is disabled. 

### Expected Behavior

Value cannot be adjusted when clicking up/down buttons.

### Reproduction Steps or Sample

1. Open https://codepen.io/jcfranco/pen/XWRaqYW
2. Click on up/down buttons

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [x] CDN
- [ ] NPM package

caripizza:
Related: #2378

julio8a:
verified on Codepen.

PR #2609, opened 2021-07-22
fix(input): disabled calcite-input + type=number allows interaction with the up/down buttons

driskull:
**Related Issue:** #2582

## Summary

fix(input): disabled calcite-input + type=number allows interaction with the up/down buttons. #2582


jcfranco (approved):
Awesome! Can you add a test for this?

eriklharper (approved):

Issue #2468, opened 2021-07-07
refactor(Inline Editable): match Figma and utilize Tailwind


julio8a:
Description
This component needs to match Figma designs.

Acceptance Criteria
The component appears as it does in Figma.

Related issues:
- #2041

Design:
- Issues above^
- full refactor (spacing, color, and non-wrapping text classes)
- close `X` idle (ui-text-3) and hover (ui-text-1) need to be updated
- https://www.figma.com/file/E3SB0i5wPy7AagB7KcjF05/?node-id=1183%3A22445

julio8a:
- The heights for S (24px), M (32px), and L(48px) should already be taken care of from the input refactor.
- the close `X`. Currently, this is using calcite-button, but will need to be switches to a native button. Similar to Alert: https://github.com/Esri/calcite-components/blob/master/src/components/calcite-alert/calcite-alert.tsx#L118
- Spacing and general color will need a review against what's in Figma

bstifle:
only thing there is L should be 44px total height

julio8a:
Verified on master demo pages.

PR #2619, opened 2021-07-23
refactor(inline-editable): tailwindify/match figma

julio8a:
**Related Issue:** #2468 #2041

## Summary

- s, m, l size adjustments 24, 32, 44 (this was actually handled by Ben with the input :) thx @bpatterson88 
- The text color when there's a value is ui-text-1 which matches Figma, placeholder text is browser default for placheolder (I believe)
- inline with controls: These are calcite-buttons, which means that they use the s, m, and l assigned icons. Figma has 16px icon for all sizes. This means L uses the 24px icons for `x` and check (due to L calcite-button).
- `x` idle/hover states: I tried using calcite-action, however, the scale is off because calcite-action has S 32px, M 48px, L 64px, which is different than input S 24px, M 32px, L 44px. other option would be to not use calcite-button and just use button, but we should use the CC as much as possible throughout components

Looking for some guidance on these items
@bstifle @caripizza @macandcheese

### DIFF
--- a/src/components/calcite-inline-editable/calcite-inline-editable.scss
+++ b/src/components/calcite-inline-editable/calcite-inline-editable.scss
@@ -19,7 +19,7 @@
 :host(:not([editing-enabled])) {
   .calcite-inline-editable__wrapper {
     &:hover {
-      @apply bg-foreground-2;
+      @apply bg-foreground-3;
     }
   }
 }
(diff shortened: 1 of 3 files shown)


bstifle:
@julio8a 

1. The X's Top/Bottom borders need to match the new border-input color:
image

2. The pencil should not have it's own separate hover area:
image

Issue #2623, opened 2021-07-26
Docs: Installation CDN


gpbmike:
### Description 

The [installation instructions](https://github.com/Esri/calcite-components#installation) use the unpkg CDN. *.arcgis.com should probably be the recommendation like [ArcGIS for JavaScript](https://developers.arcgis.com/javascript/latest/get-started/)

```html
<script type="module" src="https://unpkg.com/@esri/calcite-components/dist/calcite/calcite.esm.js"></script>
<link rel="stylesheet" type="text/css" href="https://unpkg.com/@esri/calcite-components/dist/calcite/calcite.css" />
```

driskull:
@jcfranco do you think we could have a URL on the CDN that redirects to the latest version like unpgk does?

Otherwise, we probably need to have some kind of instruction changes like this:

**Replace `<VERSION>` with the correct version tag before adding to your document.** eg. 1.0.0-beta.60

```html
<script type="module" src="https://js.arcgis.com/calcite-components/<VERSION>/calcite.esm.js"></script>
<script nomodule="" src="https://js.arcgis.com/calcite-components/<VERSION>/calcite.js"></script>
```

jcfranco:
> do you think we could have a URL on the CDN that redirects to the latest version like unpgk does?

I'll see if we can do the redirect option.

driskull:
I'm second guessing that actually. As a best practice, we wouldn't want users to use the redirect since it could automatically break their apps. Maybe its best we have something that automatically updates the readme's?

jcfranco:
True. Would there be any other case where we want the redirect-to-latest behavior from unpkg? 

Going back to the original request, I'll look at automating this on each release.

driskull:
> Would there be any other case where we want the redirect-to-latest behavior from unpkg?

I'm not sure. I would say probably not.

benelan:
Doesn't look quite right, I'm assigning it back to you @jcfranco
```css
<script
  type="module"
  src="https://unpkg.com/@esri/calcite-components@1.0.0-next.280/dist/calcite/calcite.esm.js"
></script>
<link
  rel="stylesheet"
  type="text/css"
  href="https://unpkg.com/@esri/calcite-components@1.0.0-next.280/dist/calcite/calcite.css"
/>
```

jcfranco:
Installed!

benelan:
verified in the readme after beta.64 released

PR #2644, opened 2021-07-26
docs: Update CDN usage instructions in getting started and readme documentation

driskull:
**Related Issue:** #2623

## Summary

<!--

docs: Update CDN usage instructions in getting started and readme documentation. #2623

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -2,14 +2,20 @@
 
 ## Installation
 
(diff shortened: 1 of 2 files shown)


jcfranco (changes requested):
Not all versions are deployed to production (at the moment), so the updated instructions won't work consistently.

Also, this is not what we had agreed to do in the issue: https://github.com/Esri/calcite-components/issues/2623#issuecomment-886912402

driskull:
Ok ill close for now.

Issue #2642, opened 2021-07-26
Enhancement: calcite-tooltip demo html update to achieve parity with Figma


ManahariDahal:
### Description
This issue aims to achieve parity between the calcite-tooltip following the similar layout on the Figma HTML Demo page and any usage snippets we provide.

### Acceptance Criteria 
- calcite-tooltip design should encompass the Figma HTML demo page

### Relevant Info 
- It should have all the variants

#### Which Component
_calcite-tooltip_

ManahariDahal:
Ready to be verified!

julio8a:
verified on master

PR #2647, opened 2021-07-26
test(tooltip): updated tooltip demo page

ManahariDahal:
**Related Issue:** #2642 

## Summary
- updated tooltip demo page

### DIFF
--- a/src/demos/calcite-tooltip.html
+++ b/src/demos/calcite-tooltip.html
@@ -8,18 +8,23 @@
     <style>
       .parent {
         display: flex;
+        align-items: center;
+        margin-left: 9%;
         color: var(--calcite-ui-text-3);
         font-family: var(--calcite-sans-family);
         font-size: var(--calcite-font-size-0);
         font-weight: var(--calcite-font-weight-medium);
-        margin-left: 10%;
       }
 
       .child {
         display: inline-flex;
         flex-direction: column;
-        width: 20%;
-        padding: 10px;
+        width: 10rem;
+        padding: 25px;
+      }
+
+      .right-aligned-text {
+        text-align: right;
       }
     </style>
     <script src="./_assets/head.js"></script>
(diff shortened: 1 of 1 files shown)


julio8a:
For this one, I'm assuming we'll use Screener's interactive tool to activate and take a screenshot of the tooltips @eriklharper?

julio8a:
on the first column, the left shows up on the right
image

Leading shows up as trailing
image

@caripizza @jcfranco @bstifle do you guys think we need all these variations?

bstifle:
@ManahariDahal 

lets make some layout adjustments!

image

1. labels on the left.
2. put buttons in columns like Alert does. the full width buttons are distracting
3. add vertical space between each row (56px?)

ManahariDahal:
> on the first column, the left shows up on the right
> image
> 
> Leading shows up as trailing
> image
> 
> @caripizza @jcfranco @bstifle do you guys think we need all these variations?

Is it a problem with the component itself? I double-checked, and the placements have the correct name, but it's not showing up how it should.

ManahariDahal:
> on the first column, the left shows up on the right
> image
> 
> Leading shows up as trailing
> image
> 
> @caripizza @jcfranco @bstifle do you guys think we need all these variations?

I was finally able to find what was wrong with the _left_ and _leading_. If the size of the tooltip is greater than the size of the remaining space, it will change its position accordingly. 
For instance, if there is a button on the left side of the window and there's not much space for the tooltip, it will reverse its position, same with the right side. Hope it makes sense.

bstifle:
>do you guys think we need all these variations?

yes, definitely!

bstifle (approved):

julio8a (approved):
Nice!

julio8a:
Might want to merge laster to see if we can get Travis to be nice.

PR #2649, opened 2021-07-26
test(loader): loader demo updated

ManahariDahal:
**Related Issue:** #2633 

## Summary
- updated loader demo page


bstifle (approved):
lookin good

julio8a (approved):
Looooooooks good. Same as indeterminate progress, we need to think of the best way auto test these without having Screener call out it's different on every PR.

eriklharper:
Screener has the ability to ignore regions.  In my experience there is no good way to automate visual testing of animations.  We'll just have to rely on good ol' manual human labor for those...

julio8a:
Okey, let's proceed with this one

PR #2650, opened 2021-07-26
test: Replace demo toggles with buttons in nav

driskull:
**Related Issue:** N/A

## Summary

test: Replace demo toggles with buttons in nav #2650

- Gets rid of floating toggle panel
- Replace toggles template with nav
- Use template file for common styles and toggling page properties
- Add HTML5 semantic elements to pass aXe

image


src/demos/_assets/nav.html
```diff
@@ -0,0 +1,46 @@
+<template>
```
driskull, line 1: We could call this file `demo-template.html` and `demo-template.ts` if that makes more sense.

driskull:
If this is approved, we should update each demo to make sure it's passing aXe and will therefore ensure more accurate accessibility testing results. cc @jcfranco @julio8a

julio8a (changes requested):
If we're going to move forward with something like this @bstifle had a nice design for the toggles and the header section for all demo pages:

image


image


image

However, @caripizza mentioned that having the `dir` toggle doesn't work as well. I don't understand the technical detail, but @caripizza, you can chime in here.

driskull:
For the top bar, we only probably need to toggle the dark theme and RTL. Toggling everything `loading` or `disabled` doesn't make sense.

asangma:
This is great! I'd be happy to lose the floating annoyance.
I agree that we probz don't need an app-level toggle for disabled and loading. That seems like something the test app can add per component.

re: toggles
It might be nice to have the buttons be switches labelled "Dark theme" and "RTL". And have them be off by default.

driskull:
I'll update the HTML to use the `calcite-switch`'s

asangma:
> mentioned that having the dir toggle doesn't work as well

@driskull Will it work to set the `dir` on `main` so that the toggles don't move?

asangma:
@julio8a Is the full design needed for this update? If so, do you know the components for that design?

julio8a:
No, full design not needed. Manahari has been working with Bryan on the component design for all the demo pages.

driskull:
Updated header with a rough design. Let me know if thats a good enough starting point @asangma

bstifle:
> For the top bar, we only probably need to toggle the dark theme and RTL. Toggling everything `loading` or `disabled` doesn't make sense.

agreed on this.

I think leaving them switches is best. more clear ux. 


Also, that header section should be sticky to the top. currently you have to scroll all the way to the top of the page to go home. (would also keep the switches readily accessible)

driskull:
If someone could take over the styling part that would be awesome. 🌈

julio8a:
@ManahariDahal can take over this one and work with @bstifle.

asangma:
@julio8a I can take on the styling of the demo header.

julio8a:
Okay, please work with @bstifle if needed.

asangma:
@driskull Aside from setting `display: none` on the header, is there an easy way we can turn off the demo header when need be?

The shell demos (full page demos) suffer from having the demo header in there.

driskull:
@asangma what did you have in mind? Do you want a url param, escape key to close, close button on the header?

driskull:
A close button on the header is probably the best solution.

bstifle:
Looking super good pals. love it! 💯

asangma:
@driskull Hmmm...a URL param might be nice so that one doesn't have to hit a close button every time. Could also just be a js var.

driskull:
Can we add a close button to the UI and I can add url param support after merge?

asangma:
@driskull Yeah...I think a close button and/or URL param can wait. For my purposes, I can just CSS hide the header in those full page apps for now.

driskull:
What do you want the url param to be? `headless`, `no-header`, `header-disabled`?

asangma:
`header-disabled` seems to match our pattern.

driskull:
DONE

asangma:
Yew! Thanks @driskull!

driskull:
@julio8a can you approve?

asangma:
@bstifle and @rmstinson were seeing a UI bug that @driskull and I are not seeing. I say we approve and murgle and see if the UI bug persists after murgling.

asangma:
FYI. UI bug that @bstifle reported is not specific to this PR.

bstifle:
Head is cool. but im not seeing the lines on dark theme:

image

image

asangma:
@bstifle That issue is not specific to this PR. Suggest creating a new issue for what you're seeing.

bstifle:
@caripizza @driskull related to this: https://github.com/Esri/calcite-components/pull/2772

still in the works?

bstifle:
image
just noticed this too haha! Z-axis issue?

asangma:
re: z-index issue.
I'll have a look at that.

driskull:
Create separate issues for those. Lets not hold this up

asangma:
@driskull Done.

driskull:
The components that are going over the header need to have a `position:relative` on their host most likely. Their zIndex shouldn't be leaking out.

julio8a (approved):

Issue #2303, opened 2021-06-10
Enhancement: calcite-color-picker (calciteColorPickerChange)


rslibed:
### Description
When using calcite-color-picker, the `calciteColorPickerChange` event fires while the thumb is dragged. 

### Acceptance Criteria 
Detect a value change only when the user completes the thumb drag. 

#### Which Component
calcite-color-picker

#### Example Use Case
Configuration app makes network requests when input events are fired and has very chatty behavior with the current behavior of the `calciteColorPickerChange`event. It would be ideal if a value is written once when a user completes a thumb drag rather than writing multiple values from a series of `calciteColorPickerChange` event fires during the thumb drag.

jcfranco:
Did some research on this and found a few libs that do this via 2 events (similar to `<input>`'s `input` and `change` events):

- https://material-ui.com/api/slider/ (`change` + `changeCommitted`)
- https://ant.design/components/slider/#components-slider-demo-event (`change` + `afterChange`)
- https://opensource.adobe.com/spectrum-web-components/components/slider (`input` + `change`)
- https://sapui5.hana.ondemand.com/#/api/sap.m.Slider%23events/Summary (`liveChange` + `change`)
- https://github.com/material-components/material-components-web-components/blob/master/packages/slider/README.md (`input` + `change`)

I think following `input` + `change` would be the simplest option. These names align with existing events with the same semantics and we could apply this to other components as well.

@driskull @eriklharper @paulcpederson WDYT?

driskull:
Agreed. input + change.

eriklharper:
`input` + `change` feels right, its the most idiomatic.

jcfranco:
Installed! 

@rslibed can you help verify this one?

rslibed:
Verified 🥇 
 `calciteColorPickerChange` event is fired only when thumb drag is complete.

jcfranco:
@rslibed Thanks for verifying!

PR #2653, opened 2021-07-27
feat(color-picker)!: introduce input event and modify change event to no longer fire as hue slider/color field thumbs are dragged

jcfranco:
**Related Issue:** #2303

## Summary

This introduces an input event that fires as the color field or hue slider thumb is dragged. The "minor" breaking change is that the change event will no longer fire when dragging and only when the thumb is released and final value is committed.


driskull (approved):
👍

PR #2656, opened 2021-07-27
test(select): updated demo page

ManahariDahal:
**Related Issue:** #2637 

## Summary
- updated select demo page

### DIFF
--- a/src/demos/calcite-select.html
+++ b/src/demos/calcite-select.html
@@ -6,6 +6,25 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <title>Calcite Select</title>
     <style>
+     .parent {
+        display: flex;
+        align-items: center;
+        padding: 25px 0;
+      }
+
+      .child {
+        flex: 0 0 15%;
+        margin: 0 25px;
+        color: var(--calcite-ui-text-3);
+        font-family: var(--calcite-sans-family);
+        font-size: var(--calcite-font-size-0);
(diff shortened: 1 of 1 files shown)


bstifle:
@ManahariDahal i believe we need a row with calcite-label applied

and it would be cool to have another row with an error state/message

src/demos/calcite-select.html
```diff
@@ -6,6 +6,25 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <title>Calcite Select</title>
     <style>
+     .parent {
```
asangma, line 9: Looks like we're repeating this pattern across multiple demo pages.
We should add this to demo.css.
Also it might be good to make a use class names that are semantically more grid oriented...or use consider using table layout.

asangma (changes requested):

src/demos/calcite-select.html
julio8a: Will be best to get through all the demo pages component variants, then go through and update them all at once.

asangma: @ManahariDahal I can help with centralizing the styles if you want a hand. Should be a pretty straightforward find and replace in most cases.

asangma: Ah...I see. Looks like there are a lot of one-off styles per demo page.

asangma (approved):

julio8a (approved):
Looks good!

bstifle (approved):

Issue #2630, opened 2021-07-26
Enhancement: calcite-inline-editable demo html update to achieve parity with Figma


ManahariDahal:
### Description
This issue aims to achieve parity between the calcite-inline-editable following the similar layout on the Figma HTML Demo page and any usage snippets we provide.

### Acceptance Criteria 
- calcite-inline-editable design should encompass the Figma HTML demo page

### Relevant Info 
- It should have all the variants

#### Which Component
_calcite-inline-editable_

PR #2667, opened 2021-07-28
test(inline-editable): updated inline-editable demo page

ManahariDahal:
**Related Issue:** #2630 

## Summary
- updated `calcite-inline-editable` demo page


bstifle:
Inline editable needs to broken down into S M L columns

Not in accordions.

Rows: No controls, With controls, RTL, Disabled

julio8a:
We shouldn't have the inline editable inputs inside accordions, this hides them from screener. We should get rid of other input types, RTL and disabled should be displayed out with the scales. 

This one might need a design from you @bstifle

julio8a:
This one has RTL and some others don't. For now, I would say let's keep it constant with no RTL until we figure out how we're approaching that. A toggle switch, on the same page, RTL on a different page.

ManahariDahal:
@julio8a removed the RTLs, let me know if this looks good.

bstifle (approved):
Looking fantastic! 🥇

Issue #2513, opened 2021-07-12
Tooling: Make sure that components are honoring a correct default display value.


driskull:
## Summary 

I've noticed a few components that are by default displayed as inline but should probably be displayed as block by default.

For example, a new component `calcite-list` is displayed as `inline` by default. cc @asangma 

`calcite-alert` is also displayed as `inline` by default.

It seems this is an issue with more than a few components.

## Measure of Success 

All components are displayed correctly by default.

### Resources 

Maybe we can add a common test to help make sure that the components are honoring a default display?

```ts
export async function displayedAs(componentTagOrHTML: TagOrHTML, display: string): Promise<void> {
  const page = await simplePageSetup(componentTagOrHTML);
  const element = await page.find(getTag(componentTagOrHTML));

  expect((await element.getComputedStyle()).display).toBe(display);
}
```

cc @jcfranco @julio8a can you please prioritize?

driskull:
Other components with issues:

- dropdown-group
- fab
- graph?
- inline editable?
- label?
- link
- notice (this is inline by default but then flex when [active]). I think there are a few like this.

We should probably just have a test for every component I think.

jcfranco:
A few questions on this one. 

* Do we have a default/common display value?
* Do we need to support more than `inline` | `inline-block`? 

In any case, I think we could enhance the `renders` test helper to test the display.

```ts
renders("calcite-foo", { displayAs: "block" });
```

driskull:
> Do we have a default/common display value?

No, I don't think so. It depends on the component.

> Do we need to support more than inline | inline-block?

Yes, I know that some of our components are displayed as block, flex, or inline-flex by default. We should just keep them consistent once we decide and probably document their default display as well.

The problem I'm seeing at this time is that some do not set a display at all and by default that makes them `inline`. However, they have heights set and should not be set to display inline.

driskull:
@jcfranco I can take this one if you want.

jcfranco:
Thanks for tackling this one, @driskull!

I had asked if we had a default display based on our common component structure (I think most use an internal wrapper) to possibly define a default on the global CSS and define needed overrides in individual component stylesheets. If applicable, this could slightly reduce the file size while simplifying the pattern and test util. WDYT?

driskull:
Gotcha. I don't think we're supposed to define `:host {` styles on the global stylesheet right? We would have to target all the `calcite-` prefixed components.

[...]

PR #2669, opened 2021-07-29
test: Make sure that components are honoring a correct default display value

driskull:
**Related Issue:** #2513

## Summary

test: Make sure that components are honoring a correct default display value.

### DIFF
--- a/src/components/calcite-radio-button/calcite-radio-button.tsx
+++ b/src/components/calcite-radio-button/calcite-radio-button.tsx
@@ -143,8 +143,6 @@ export class CalciteRadioButton {
 
   private inputEl: HTMLInputElement;
 
-  private radio: HTMLCalciteRadioElement;
-
   private rootNode: HTMLElement;
 
   //--------------------------------------------------------------------------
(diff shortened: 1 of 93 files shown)


src/components/calcite-radio-button/calcite-radio-button.tsx
```diff
@@ -257,7 +255,7 @@ export class CalciteRadioButton {
   @Listen("click")
   check(event: MouseEvent | FocusEvent): void {
     // Prevent parent label from clicking the first radio when calcite-radio-button is clicked
-    if (this.el.closest("label") && (event.target === this.el || event.target === this.radio)) {
+    if (this.el.closest("label") && event.composedPath().includes(this.el)) {
```
jcfranco, line 258: This is unrelated. Please separate this out. This should be a `fix`.

jcfranco:
@driskull Thanks for helping out with this! Can you extract out the radio-button fix before merging?

jcfranco (approved):
👍🏼

src/components/calcite-radio-button/calcite-radio-button.tsx
driskull: @jcfranco I had to do this because the test was failing since the radio switched from inline to inline-block.

driskull: I can skip the test for now if that is what you would like.

driskull: These two tests are failing from the change from inline to inline-block.

https://github.com/Esri/calcite-components/blob/master/src/components/calcite-label/calcite-label.e2e.ts#L610-L664

jcfranco: If this was needed because of the display changes then it's fine. Ignore my comment.

Issue #2640, opened 2021-07-26
Enhancement: calcite-tabs demo html update to achieve parity with Figma


maxdahal:
### Description
This issue aims to achieve parity between the calcite-tabs following the similar layout on the Figma HTML Demo page and any usage snippets we provide.

### Acceptance Criteria 
- calcite-tabs design should encompass the Figma HTML demo page

### Relevant Info 
- It should have all the variants

#### Which Component
_calcite-tabs_

ManahariDahal:
ready to verify!

PR #2674, opened 2021-07-29
test(tabs): updated tabs demo page

ManahariDahal:
**Related Issue:** #2640 

## Summary
- updated `calcite-tabs` demo page
- small, medium and large variants included


bstifle:
pretty good!

couple things:
image

1. can we expand the column widths for these so they dont scroll?
image
2. We need a bordered position bottom version
3. Would be nice to have 3 items in each example, instead of 2

julio8a (approved):

caripizza:
Why are we removing the cookie and programmatically changing the active tab examples? Are these already covered in e2e tests?

bstifle:
@ManahariDahal can you add the position bottom version of this?
image

bstifle (approved):
Nice one! 1️⃣

Issue #2335, opened 2021-06-18
Bug: calcite-color-picker - color indicator doesn't follow pointer in Safari


hccampos:
### Actual Behavior

![color-picker-safari](https://user-images.githubusercontent.com/3759114/122550528-125e9f80-d034-11eb-9c87-7565bf5e3336.gif)

As can be seen above, when moving the mouse pointer in Safari, the color indicator doesn't match the pointer, like it does in Chrome and Firefox.

### Expected Behavior

The color indicator should always match the pointer like it does in Chrome and Firefox.

### Reproduction Steps or Sample

1. Open the [color picker sample](https://esri.github.io/calcite-components/?path=/story/components-controls-colorpicker--simple) in Safari.
2. Drag the color indicator at the top of the component.
3. Note that the color indicator starts "lagging" behind the mouse pointer.

### Relevant Info

_Browser_: Safari 14.1.1
_OS_: MacOS Catalina
_Version_: `@esri/calcite-components@1.0.0-beta.56`

- [ ] CDN
- [x] NPM package

CC @RalucaNicola

jcfranco:
Installed! 

@hccampos Can you help verify this one?

hccampos:
Now works in Safari, Chrome and Firefox. Looking good 👍

jcfranco:
@hccampos Thanks for verifying! 💪🏼

PR #2676, opened 2021-07-30
fix(color-picker): improve mouse tracking when dragging color field/hue slider thumb

jcfranco:
**Related Issue:** #2335 

## Summary

This revisits mouse tracking when the color field/hue slider thumb is dragged. The previous implementation could get out of sync in some cases (e.g, `MouseEvent#movement(X|Y)` would emit non-zero values when the cursor would move past the edge, Chrome and Firefox emit 0). This approach is simpler and slightly more efficient since it won't have to calculate the canvas region when the mouse is dragged outside of its bounds.


driskull (approved):
👍

Issue #2675, opened 2021-07-30
Bug: calcite-input with clearable + "enter" key = no value


gpbmike:
### Actual Behavior

- Using the `clearable` attribute (or `type="search"`) will result in no value when using the "enter" key to submit a form.
- Input is cleared

### Expected Behavior

- Input value is submitted
- Input not cleared

### Reproduction Steps or Sample

Examples: https://codepen.io/gpbmike/pen/gOWzMOW

```html
<form>
  <calcite-input clearable />
  <button>Submit</button>
</form>
```

1. Add `clearable` or `type="search"` to `calcite-input`.
2. Submit form with "enter" key
3. Input value will be "".

### Relevant Info

I think the enter key is triggering the clearable button.

_Version_: `@esri/calcite-components@1.0.0-beta.60`

Example uses https://unpkg.com/@esri/calcite-components... which is both CDN and NPM package...

- [x] CDN
- [x] NPM package

driskull:
This seems to be because the clear button is submitting the form. Adding `type="button"` to it fixes it. Probably need to do that for all buttons in this component.

Wouldn't be an issue if it were a shadowed component :)

gpbmike:
This looks fixed to me on the codepen example (points to beta 61). I don't know what the process is for closing out issues.

PR #2679, opened 2021-07-30
fix(input): Internal buttons should not be able to submit forms

driskull:
**Related Issue:** #2675

## Summary

fix(input): Internal buttons should not be able to submit forms. #2675

This sets all internal buttons to be a type of "button" so that they do not submit the form when an enter key is pressed on the search input element.

### DIFF
--- a/src/components/calcite-input/calcite-input.tsx
+++ b/src/components/calcite-input/calcite-input.tsx
@@ -30,7 +30,12 @@ import {
 } from "../../utils/locale";
 import { numberKeys } from "../../utils/key";
 import { hiddenInputStyle } from "../../utils/form";
-import { isValidDecimal, isValidNumber, parseNumberString, sanitizeNumberString } from "../../utils/number";
+import {
+  isValidDecimal,
+  isValidNumber,
+  parseNumberString,
+  sanitizeNumberString
+} from "../../utils/number";
 import { CSS_UTILITY, TEXT } from "../../utils/resources";
 
 type NumberNudgeDirection = "up" | "down";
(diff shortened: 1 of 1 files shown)


jcfranco:
@driskull Can you revisit the title to make the fix clearer in the changelog for users? If it sums up the PR that's an added bonus. Otherwise, can you add some context or summarize the PR? Without this, reviewers need to infer this by looking at all of the changes and reading up on the original issue. I think these steps are great if reviewers want some additional info behind the PR. 

Also, be mindful of our process. The linked issue is not triaged.

driskull:
> Also, be mindful of our process. The linked issue is not triaged.

Sorry :(

I had no other issues and none were left unassigned for the milestone. With that in mind, we should be flexible.

driskull:
Updated descriptions

jcfranco:
> With that in mind, we should be flexible.

Totally agree with you there, but please keep us in the loop in order to keep sprints and labels in sync.

jcfranco (approved):
:boom:

src/components/calcite-input/calcite-input.tsx
```diff
@@ -586,6 +594,7 @@ export class CalciteInput {
         disabled={this.disabled}
         onClick={this.clearInputValue}
         tabIndex={this.disabled ? -1 : 0}
+        type="button"
```
jcfranco, line 597: Do you think we'll remember to remove this once `calcite-input` is shadowified™?

driskull: Its probably a best practice for most of internal buttons to set the type anyway so that they don't submit a form. Its ok if it stays as button type. It won't cause any harm at all.

Issue #1863, opened 2021-03-30
Bug: calcite-date-picker - selecting day activates previous day in certain cases


ChristopheSuter:
### Summary
Selects previous day when click on Calendar. Screenshot click on March the 24th selects March the 23rd

### Reproduction Steps

1.

### Relevant Info 
![image](https://user-images.githubusercontent.com/78811924/113018515-ee9c6380-9180-11eb-8e7d-08a38217c995.png)

caripizza:
Hi @ChristopheSuter - do you have any information? Please provide your snippet of code, the calcite-components npm package version you're using, your browser version - anything to help us investigate this issue.

ChristopheSuter:
Hi, 
Here is some code
`<!DOCTYPE html>
<html>
<head>
     <meta charset="utf-8">
     
     <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no">
     <meta name="mobile-web-app-capable" content="yes">
     <meta name="apple-mobile-web-app-capable" content="yes">
     <meta name="apple-mobile-web-app-status-bar-style" content="default">
     <script type="module" src="https://unpkg.com/@esri/calcite-components/dist/calcite/calcite.esm.js"></script>
     <script nomodule="" src="https://unpkg.com/@esri/calcite-components/dist/calcite/calcite.js"></script>
     <link rel="stylesheet" type="text/css" href="https://unpkg.com/@esri/calcite-components/dist/calcite/calcite.css" />
     
     <link rel="stylesheet" href="https://js.arcgis.com/4.18/esri/themes/light/main.css" />
     
</head>
<body>
     <div style="width: 400px">
          <calcite-date-picker scale="m" value="2020-12-12" min="2016-08-09" max="2023-12-18" locale="en"
               intl-next-month="Next month" intl-prev-month="Previous month"></calcite-date-picker>
     </div>
</body>
</html>`

ChristopheSuter:
Hi,
Here is a codepen, I am using latest version of Chrome 
https://codepen.io/christophesuter/pen/JjEEbYd

caripizza:
@ChristopheSuter I don't see the issue in your CodePen. How can I reproduce?

ChristopheSuter:
I have the same behavior in Chrome, Edge and FF, could this be linked to my locale (fr-ch) ?

ChristopheSuter:
[...]

PR #2680, opened 2021-07-30
fix(date-picker): Selecting a day no longer activates previous day in certain time zones.

driskull:
**Related Issue:** #1863

## Summary

fix(date-picker): Selecting a day no longer activates previous day in certain time zones. 

Before converting a date to ISO format, `date.getTimezoneOffset() * 60000` is called to prevent this error from happening.


jcfranco (approved):
LGTM! Before merging, can you add a test for this and revisit the commit title? The current text is describing the broken behavior.

driskull:
@jcfranco I tried to add a test but failed. The test is dependent on setting the browser timezone in order to run correctly.

driskull:
Updated title of PR.

jcfranco:
> The test is dependent on setting the browser timezone in order to run correctly.

I see. If you haven't already, could you try [`page#emulateTimezone`](https://github.com/puppeteer/puppeteer/blob/v5.4.0/docs/api.md#pageemulatetimezonetimezoneid)? If this doesn't work, let's create an issue to see if we can come up with a consistent way of handling these tests.

driskull:
Yeah that doesn't seem to work. It only seems to work if I set my system timezone. Emulating has no effect on the value.

Seems like i was testing it correctly  as seen here: https://github.com/puppeteer/puppeteer/pull/4949/files#diff-ffaaf2938a0892d9da34f14f80532475e9d3a5a337225c9d5a51e674c8302c41R174

Maybe they don't emulate `getTimezoneOffset`.

driskull:
I think @paulcpederson should look at this one before we merge it. Can it wait till he's back?

jcfranco:
@driskull Sure, let's hold off on this one then. BTW, thanks for making all of the changes and trying to get the test in! 💪🏼

driskull:
@paulcpederson can you review?

driskull:
@paulcpederson Will merge. Let me know if this needs further changes.

paulcpederson:
Looks good, no changes needed.

Issue #2697, opened 2021-08-02
Tooling: error on fresh repo clone


ptmdmusique:
### Actual Behavior

I cloned the repo, then got error about `Plugin` after `npm install` then `npm start`. 
![image](https://user-images.githubusercontent.com/37349324/127921442-07f605bd-d089-4593-b0b4-e667af13410b.png)

I tried `npm run build` then it starts throwing Tailwind `@apply` error
![image](https://user-images.githubusercontent.com/37349324/127921465-2bb05ff8-b55c-48fe-9d99-17566e918eb4.png)

### Expected Behavior

No error, `npm start` should run correctly after `npm install` 

### Relevant Info

I'm using Windows

cc @paulcpederson

caripizza:
Related: #1941

@ptmdmusique can you provide a) your system details and b) the version of node you used when you attempted the installs?

ptmdmusique:
Hey @caripizza 

Here is my spec:
- Windows 10
- Node v14.17.0
- npm 6.14.13

driskull:
Autoprefixer has its own types here: https://unpkg.com/browse/autoprefixer@10.3.1/lib/autoprefixer.d.ts

However, we still have a npm package for `@types/autoprefixer` which is deprecated. 

https://www.npmjs.com/package/@types/autoprefixer

Seems like we also need to update `postcss` or lock it to a version that works with `autoprefixer`. Currently, we just have `@stencil/postcss` installed. Ill submit a PR.

@jcfranco can I take this one? AKA triage this thang

jcfranco:
@driskull Thang triaged, thanks!

driskull:
I'm also seeing this on build:

```
[WARNING] calcite-filter is not referenced in stencil.config.ts and will not have a bundle generated
[WARNING] calcite-handle is not referenced in stencil.config.ts and will not have a bundle generated
```

Do we want to do these bundles @jcfranco ?

jcfranco:
No, those aren't standalone components, so no bundle. 🚫 📦

caripizza:
Thanks @jcfranco and @driskull! @ptmdmusique can you test drive beta.61 to see if this issue still occurs?

caripizza:
Oops - I just noticed Matt's pr was merged after we released beta.61 last night, so this one will need to wait for 1.0.0-next.262 to test the fix.

caripizza:
[...]

PR #2705, opened 2021-08-03
build: Remove @types/autoprefixer and add postcss. #2697

driskull:
**Related Issue:** #2697

## Summary

build: Remove @types/autoprefixer and add postcss. #2697

Update build to get latest postcss version compatible with autoprefixer.


driskull:
Maybe this will fix the repo switching error we have been seeing. 🤷

jcfranco:
From what I understand, I _think_ we only need to remove the deprecated `@types/autoprefixer` package (good catch, BTW). We only reference `postcss` from `@stencil/postcss` and the latest `autoprefixer` includes the types.

driskull:
I tried that but the two weren't in sync. I think we need both.

jcfranco:
> I tried that but the two weren't in sync. 

Can you expand on that? Did you get a warning or an error?

driskull:
Yeah the issue was that the auto prefixer typings were not compatible with the version of postcss that stencil/postcss defines.

If we set the latest of both it works fine. That's why the plug-in error is shown

jcfranco:
I think I see what's going on. `autoprefixer` has `postcss@^8.1.0` as a peer dependency. Since we don't have it as an explicit dependency, the one at the root of `node_modules` is the one used to dedupe the common version across other dependencies (`postcss@7.0.35`).

Thanks for fixing this one!

jcfranco (approved):
👍🏼

jcfranco:
On a related note, we should probably look at adding other peer dependencies based on the post `npm install` warnings. NPM 7 is supposed to handle this automatically, but we can't upgrade until Node 15 gets into Active LTS status.

driskull:
> On a related note, we should probably look at adding other peer dependencies based on the post npm install warnings. NPM 7 is supposed to handle this automatically, but we can't upgrade until Node 15 gets into Active LTS status.

👍 yeah

Issue #2470, opened 2021-07-07
refactor(Tile and Tile Select): match Figma and utilize Tailwind


julio8a:
Description
This component needs to match Figma designs.

Acceptance Criteria
The component appears as it does in Figma.

Related issues:
- #1230

Related prs:
- pr #1769
- pr #1770
- pr #1771

Design:
- https://www.figma.com/file/E3SB0i5wPy7AagB7KcjF05/?node-id=1183%3A19780

julio8a:
@bstifle, merged the PR for this one. Can you verify it on master?

bstifle:
lookin good!

PR #2710, opened 2021-08-03
refactor(Tile and Tile Select): match Figma and utilize Tailwind

justinhough:
**Related Issue:** #2470 

## Summary

The alignment on calcite-tile-select for title and inputs was caused by the `calcite-checkbox` and `calcite-radio` using `inline-block` rather than `block`.

I checked the following components to verify that the change did not adversely affect anything else.

- calcite-radio-button
- calcite-radio-group
- calcite-tree


justinhough:
@julio8a This PR won't pass since the tests for both `calcite-checkbox` and `calcite-radio` are checking against the element being `inline` in the build and the change needed for Tile Select need those two elements to be `block` in order to layout properly. Let me know how you want to proceed on this. Thanks.

julio8a:
@driskull or @jcfranco since we need to change from inline to block to get the visual we need, should we update the tests?

driskull:
Yeah update the tests 👍

bstifle (approved):
screener had some weird stuff but verified that they were fine.

lookin good!

julio8a (approved):
Nice! Thanks for updating the test!

julio8a:
@justinhough, can you resolve conflicts in this PR?

julio8a:
Screener is showing visual changes on popover, I don't think it fully loaded when it took the screenshot because it's on the top left.

julio8a:
@jcfranco, in this case, should be approve screener and merge the PR? or merge without approving screener. 

My understanding is that once we approve it, the approved version (top left popover, which is incorrect) becomes the new norm, is this correct?

driskull:
@julio8a I think it's better to `ignore` it here:

image

src/tests/commonTests.ts
```diff
@@ -65,8 +65,8 @@ export async function renders(
   const element = await page.find(getTag(componentTagOrHTML));
 
   expect(element).toHaveAttribute(HYDRATED_ATTR);
-  expect(await element.isVisible()).toBe(options?.visible ?? true);
-  expect((await element.getComputedStyle()).display).toBe(options?.display ?? "inline");
+  expect(await element.isVisible()).toBe(options.visible ?? true);
```
driskull, line 68: @julio8a these two lines should be reverted back. Can you revert both of these changes?

src/tests/commonTests.ts
```diff
@@ -65,8 +65,8 @@ export async function renders(
...
+  expect((await element.getComputedStyle()).display).toBe(options.display ?? "block");
```
driskull, line 69: fallback display needs to be "inline". The default display for a custom element is "inline". Please revert this.

src/tests/commonTests.ts
driskull: options are optional so the `?` needs to be there.

PR #2711, opened 2021-08-03
test(slider): updated slider demo page

ManahariDahal:
**Related Issue:** #2638 

## Summary
- updated _calcite-slider_ demo page


bstifle (approved):
perfection 🏆

julio8a:
We don't need the mirrored, we should remove these.

jcfranco:
> We don't need the mirrored, we should remove these.

Will there be a replacement for this? `mirrored` is different from RTL and the existing set up helps catch regressions on  the different slider configurations + mirrored.

julio8a:
Maybe I got confused with some that seem like duplicates, is there a difference between these two columns?
image

jcfranco:
The 2nd column does not not look mirrored.

julio8a:
Let's mirror the right column all the way through. @ManahariDahal

ManahariDahal:
@jcfranco does it have to do with the histogram specifically? It's weird that it's not mirroring it. `mirrored` attribute is working for every other apart from the histogram.

jcfranco:
> `mirrored` attribute is working for every other apart from the histogram.

I didn't notice that we were talking specifically about the histogram portion of the demo page. That is by design. `mirrored` will not kick in if there's a histogram involved. Sorry about the confusion.

ManahariDahal:
> > `mirrored` attribute is working for every other apart from the histogram.
> 
> I didn't notice that we were talking specifically about the histogram portion of the demo page. That is by design. `mirrored` will not kick in if there's a histogram involved. Sorry about the confusion.

yeah, sorry for the confusion on my part too. I will remove the mirrored part in the histogram then.

julio8a (changes requested):
Remove mirrored for histogram section

julio8a (approved):
Thanks for the updates!

Issue #2692, opened 2021-08-02
Enhancement: calcite-tip-manager demo html update to achieve parity with Figma #2643


ManahariDahal:
### Description
This issue aims to achieve parity between the calcite-tip-manager following the similar layout on the Figma HTML Demo page and any usage snippets we provide.

### Acceptance Criteria 
- calcite-tip-manager design should encompass the Figma HTML demo page

#### Which Component
_calcite-tip-manager_

PR #2723, opened 2021-08-04
test(tip-manager): updated tip and tip-manager into one demo page

ManahariDahal:
**Related Issue:** #2692 

## Summary
- combined _calcite-tip_ and _calcite-tip-manager_ into one demo


bstifle (approved):
nailed it! 🔨

PR #2725, opened 2021-08-04
refactor(tooltip, popover): Log warning when reference element if it is not found and rename internal variables.

driskull:
**Related Issue:** #2446

## Summary

refactor(tooltip, popover): Log warning when reference element if it is not found and rename internal variables. #2446


src/components/calcite-popover/calcite-popover.tsx
```diff
@@ -250,6 +242,26 @@ export class CalcitePopover {
   //
   // --------------------------------------------------------------------------
 
+  setUpReferenceElement = async (): Promise<void> => {
+    this.removeReferences();
+    this._referenceElement = this.getReferenceElement();
+
+    if (!this._referenceElement) {
+      await new Promise((resolve) => setTimeout(resolve, 250));
```
jcfranco, line 250: Where does the 250 come from?

src/components/calcite-popover/calcite-popover.tsx
```diff
@@ -165,7 +158,7 @@ export class CalcitePopover {
 
   @Element() el: HTMLCalcitePopoverElement;
 
-  @State() _referenceElement: HTMLElement = this.getReferenceElement();
+  @State() _referenceElement: HTMLElement;
```
jcfranco, line 161: Nitpick: To avoid confusion with the prop, maybe this could be renamed to `effectiveReferenceElement` or similar?

driskull: Yeah I like that.

src/components/calcite-popover/calcite-popover.tsx
driskull: It's just an arbitrary number. That's the thing. If the reference element isn't in the DOM, how long should we wait to try again?

driskull: Maybe we should just remove this and have the warning so at least a user can see something happened?

jcfranco: Yeah, that sounds like a good idea. There's no guarantee it'll be available after the timeout either. 😞  

I would've thought that resolving the reference element would be reliable by the time `open` is set to true. Does this problem show up only when the component is created w/ `open=true` and the ref element may not be there yet or is this problem occurring regardless of when the tooltip/popover is opened?

driskull: Waiting for `open` isn't a since the `calcite-tooltip-manager` queries by `data-calcite-tooltip-reference` which doesn't get set unless a reference element is found. So it can't access the tooltip to tell it to find it's reference element because it identifies it by the data attribute.

jcfranco: I see. Looks like we would need to refactor the manager + tooltip to cover this case. Is it recommended to have all tooltips after the tooltip manager? If so, maybe the warning could have some info about this. 

Another thing to try would be to resolve the reference element in the next animation frame to ensure the DOM is in sync. WDYT?

driskull: > Another thing to try would be to resolve the reference element in the next animation frame to ensure the DOM is in sync. WDYT?

How would we do this?

driskull: > Is it recommended to have all tooltips after the tooltip manager? If so, maybe the warning could have some info about this.

We could recommend that.

driskull: > Looks like we would need to refactor the manager + tooltip to cover this case. 

We could refactor the entire tooltip/popover so that instead of using a referenceElement, the user slots the "trigger". However, that would mean that the tooltip surrounds the trigger and we lose the flexibility of having the tooltip and trigger be in completely different areas of the DOM.

Maybe we can just offer a slot as another way to set the reference element?

- Add `trigger` slot
- rename `reference-element` to `trigger-id`? Maybe a better name than `trigger`?

driskull: @jcfranco Can we merge this so there's a warning at least? Then we can discuss the possible refactoring later?

jcfranco: Sounds good. WRT the commit type/message, looks like there was more done than showing a warning when the ref is not found. If this is the case, can you revisit the PR title?

driskull: Done

caripizza:
Should Tooltip/Popover stories be updated with this one? I'm testing now and it looks like when the component has `open` to start, the arrow is not placed correctly:

image
image
Tooltip demo page, first example with additional `open` attribute:
image

driskull:
@caripizza i'm not seeing that on the deployed storybook. Are you?

driskull:
@caripizza can you create a new issue for this?

caripizza:
@driskull Yep I'll create an issue (screenshots are from my local builds on master branch, currently at `c211b04`).

caripizza:
^ #2831

Issue #2686, opened 2021-08-02
Enhancement: calcite-action demo html update to achieve parity with Figma


ManahariDahal:
### Description
This issue aims to achieve parity between the calcite-action following the similar layout on the Figma HTML Demo page and any usage snippets we provide.

### Acceptance Criteria 
- calcite-action design should encompass the Figma HTML demo page

#### Which Component
_calcite-action_

julio8a:
verified on demo pages

PR #2733, opened 2021-08-04
Manahari/2686 calcite action demo update

ManahariDahal:
**Related Issue:** #2686 

## Summary
- updated _calcite-action_, _calcite-action-bar_, and _calcite-action-pad_ into one demo page


bstifle (commented):
image

not a big fan of how the action bar stuff is layed out... would probably be better to do something like this:

image

Issue #2740, opened 2021-08-05
Bug: Block shows toggle icon when header-menu-actions has content


asangma:
### Actual Behavior
Block show toggle icon when there's content in the `header-menu-actions` slot.

### Expected Behavior
Block shouldn't show toggle icon when there's content in the `header-menu-actions` slot.

### Reproduction Steps or Sample

1. See Block demo.

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

asangma:
@caripizza @jcfranco Another quick one punch-out-able.

PR #2742, opened 2021-08-05
fix(block): stops rendering collapsibleIcon when there are header-menu-actions

asangma:
**Related Issue:** (#2740)

## Summary
- adds logic to not render collapsibleIcon when `header-menu-actions` has content
- adds explicit test for this
- adds explicit test to not render collapsibleIcon when `control` has content

cc @caripizza @jcfranco


driskull (approved):
👍

Issue #2735, opened 2021-08-04
Bug: Block heading color is wrong on idle


bstifle:
### Actual Behavior
heading text is ui-text-3 at idle
![image](https://user-images.githubusercontent.com/14083530/128267384-17c586c5-51a4-4f74-8849-4506fa6cf2cd.png)

### Expected Behavior
heading text should be ui-text-1 at all times
![image](https://user-images.githubusercontent.com/14083530/128267334-818f069a-7cba-4c6f-acc5-925a062c651b.png)

julio8a:
Verified on master

PR #2754, opened 2021-08-06
refactor(Block heading): match Figma and utilize Tailwind

justinhough:
**Related Issue:** #2735

## Summary

Heading text adjusted from `ui-text-3` to `ui-text-1` by default.

### DIFF
--- a/src/components/calcite-block/calcite-block.scss
+++ b/src/components/calcite-block/calcite-block.scss
@@ -79,7 +79,7 @@ calcite-handle {
 .header .title .heading {
   @apply p-0
     text--1
-    text-color-1
+    text-color-2
     font-medium
     word-break
     leading-tight
--- a/src/components/calcite-block/calcite-block.scss
+++ b/src/components/calcite-block/calcite-block.scss
@@ -79,7 +79,7 @@ calcite-handle {
 .header .title .heading {
   @apply p-0
     text--1
-    text-color-3
+    text-color-1
     font-medium
     word-break
     leading-tight


bstifle (approved):
Nice! Ship it 📦

src/components/calcite-block/calcite-block.scss
```diff
@@ -79,7 +79,7 @@ calcite-handle {
 .header .title .heading {
   @apply p-0
     text--1
-    text-color-3
+    text-color-1
```
asangma, line 82: This is the default color for the heading and as per @bstifle's suggestion, we made it text-1 [when the Block is open](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-block/calcite-block.scss#L152).

asangma (changes requested):

src/components/calcite-block/calcite-block.scss
bstifle: hey i think we are both wrong. We should match what Accordion is doing:

ui-text-2 idle and closed

ui-text-1 hover and open

justinhough: @bstifle. Looks like the open state is already there, so you just need to the default state to change to `text-color-2`, correct?

bstifle: Yeah. That would keep the open/close accordion type components cohesive. You good with that @asangma ?

bstifle:
hmm. ok. digging into accordion again. lets get block 1:1 with accordion.

Looks like summary gets a hover/active change too:

### Idle:
Heading: ui-text-2
Summary: ui-text-3

### Idle Hover:
Heading: ui-text-1
Summary: ui-text-2

### Active (hover and idle):
Heading: ui-text-1
Summary: ui-text-2

src/components/calcite-block/calcite-block.scss
asangma: I think the difference between 1 and 2 is too subtle...especially in more complex interfaces.
Might be fine on an awesome monitor with good eyes, but not so great in other cases.

asangma: I'll defer, but here's a comparison.

image

bstifle: it's pretty weird to have the heading and summary the same color. we dont do that anywhere else

asangma: Copy that.

asangma (approved):
G2G

Issue #2688, opened 2021-08-02
Enhancement: calcite-flow demo html update to achieve parity with Figma


ManahariDahal:
### Description
This issue aims to achieve parity between the calcite-flow following the similar layout on the Figma HTML Demo page and any usage snippets we provide.

### Acceptance Criteria 
- calcite-flow design should encompass the Figma HTML demo page

#### Which Component
_calcite-flow_

julio8a:
verified on master

PR #2755, opened 2021-08-06
test(flow): updated flow demo page

ManahariDahal:
**Related Issue:** #2688 

## Summary
- update `calcite-flow` demo update


bstifle (approved):
flow looks good! nice! ✨

julio8a (changes requested):
For this one, it should just take to you to the component page and not have a single selection intermediate page
image


Also, within the component, we should probably use calcite-buttons and not native default buttons.

julio8a (approved):

Issue #2547, opened 2021-07-15
Bug: calcite-input date range picker changes the wrong input


noahmulfinger:
### Actual Behavior

In certain cases, when I am choosing a new value for the first input in the date range picker, it will instead change the second input.

### Expected Behavior

My expectation is that only the first date picker would be updated when I modify the first input. Same applies to the second input.

### Reproduction Steps or Sample

1. Go to https://esri.github.io/calcite-components/?path=/story/components-controls-inputdatepicker--range
1. Click on the first date input and select 12/13 in the date picker. It should correctly update the first input to 12/13 and open the second date picker
1. Refresh the page to reset.
1. Click on the first date input and select 12/14 in the date picker. It will unexpectedly set the *second* input to 12/14 and open the second date picker
1. You can repeat that with any date higher than 12/14 and it will do the same behavior

### Relevant Info

Tested in Firefox and Chrome

_Version_: `@esri/calcite-components@latest`

- [ ] CDN
- [ ] NPM package
- [x] Storybook build

cc @isaacviel

julio8a:
Verified on master, demo pages.

noahmulfinger:
I also just took another look at this (on the Storybook), and its seems partially fixed, but the behavior still seems unexpected to me. When I choose 12/14 as in the steps I listed above, it still updates the second input. I would expect that if I click the first input, any action I take would only update the first date. 

Happy to open a second issue if this requires more discussion of the expected behavior.

PR #2756, opened 2021-08-06
fix(date-picker, input-date-picker): Fix setting date for the correct range input

driskull:
**Related Issue:** #2547

## Summary

fix(date-picker, input-date-picker): Fix setting date for the correct range input. #2547

- Uses activeRange to determine which date is being set.
- getDaysDiff() was returning odd results.
- Fixes mutable prop warnings.
- Keeps component active when setting range between start and end.


driskull:
@paulcpederson should the range allow the same date as shown below? I think maybe this could be another setting on the range?

Screen Shot 2021-08-06 at 12 27 22 PM

driskull:
This may also break `proximitySelectionDisabled`. Do we need to support that?

julio8a:
Yeah, range should allow you to select the same day.

driskull:
Ok. I thought so. currently it doesn't let you do that but I think its kind of messed up.

driskull:
@paulcpederson what are your thoughts on removing `proximitySelectionDisabled`?

julio8a:
gotcha, yea, we should fix that to allow same date selection.

driskull:
I'll have to meet with @paulcpederson about this. The current UX in the `calcite-date-picker` is a bit odd. It's hard to select a range between two dates next to each other and impossible to select the same date as a range.

paulcpederson:
@driskull I think @rpanichakit9541 may know more about the original requirements for range selection as he implemented that peice.

rpanichakit9541:
@driskull - I believe the "proximitySelectionDisabled" flag spawned from this comment here https://github.com/Esri/calcite-components/pull/1136#issuecomment-712330139 .  But if we prefer to remove it and just one one range selection mode, that sounds good to me.  If you want to meet or if I can assist with anything let me know.  Also @bstifle might be able to speak to things as well since he contributed a lot to the design and PR feedback.

Here's the original issue https://github.com/Esri/calcite-components/issues/541

bstifle:
@rpanichakit9541 what precisely do you need feedback on visually? (being able to select the same day in a range seems acceptable)

rpanichakit9541:
@bstifle - I agree, I think the not being to select the same day is a bug.  I just added you to the thread in case there were other questions regarding original requirements.

driskull:
Besides not being able to set the same day, its really tough just to get the range correct. Especially since when you hover a range and click, it doesn't set to what it is telling you. For example:

test

I would expect the range to be what the blue hover is set to but it does the opposite.

I think outside of this issue we should fix:

- Setting same range date
- Get hover and setting date range in sync (as seen in gif)

For the `calcite-input-date-picker`, this PR now limits the range to only before or after dates.

input

src/components/calcite-date-picker/calcite-date-picker.tsx
```diff
@@ -40,10 +40,10 @@ export class CalciteDatePicker {
   //
   //--------------------------------------------------------------------------
   /** Active range */
-  @Prop() activeRange?: "start" | "end" = "start";
+  @Prop() activeRange?: "start" | "end";
```
driskull, line 43: For the default date-picker without an input, we don't want it to be in "start" or "end". We could have a new mode for both or just leave it empty.

bstifle:
yeah its trying to find the day nearest to one of the selections.

but definitely if you focus on the second input, it should ONLY be choosing the end date

then with that, if focusing on the first input, only selecting the start date

driskull:
@paulcpederson @rpanichakit9541 Can you review the PR?

src/components/calcite-date-picker/calcite-date-picker.tsx
```diff
@@ -441,35 +466,30 @@ export class CalciteDatePicker {
...
       }
-      this.start = dateToISO(date);
-      this.setStartAsDate(date, true);
-      this.activeStartDate = date;
+      this.setStartDate(date);
     } else if (!this.endAsDate) {
-      this.end = dateToISO(date);
-      this.setEndAsDate(date, true);
-      this.activeEndDate = date;
+      this.setEndDate(date);
```
rpanichakit9541, line 473: Nice refactor! 👍

paulcpederson (approved):

paulcpederson:
LGTM

rpanichakit9541 (approved):
LGTM 👍

PR #2766, opened 2021-08-09
docs(action): Deprecate calciteActionClick.

driskull:
**Related Issue:** N/A

## Summary

Use onClick instead.

docs(action): Deprecate calciteActionClick.

### DIFF
--- a/src/components/calcite-action/calcite-action.e2e.ts
+++ b/src/components/calcite-action/calcite-action.e2e.ts
@@ -133,40 +133,4 @@ describe("calcite-action", () => {
 
     expect(clickSpy).toHaveReceivedEventTimes(0);
   });
-
-  it("should emit 'calciteActionClick' event", async () => {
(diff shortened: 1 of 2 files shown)


src/components/calcite-action/calcite-action.e2e.ts
```diff
@@ -133,40 +133,4 @@ describe("calcite-action", () => {
 
     expect(clickSpy).toHaveReceivedEventTimes(0);
   });
-
-  it("should emit 'calciteActionClick' event", async () => {
```
jcfranco, line 137: Should these be removed when `calciteActionClick` is dropped?

jcfranco (approved):
👍🏼

src/components/calcite-action/calcite-action.e2e.ts
driskull: I think its safe enough to remove now.

Issue #2631, opened 2021-07-26
Enhancement: calcite-label demo html update to achieve parity with Figma


ManahariDahal:
### Description
This issue aims to achieve parity between the calcite-label following the similar layout on the Figma HTML Demo page and any usage snippets we provide.

### Acceptance Criteria 
- calcite-label design should encompass the Figma HTML demo page

### Relevant Info 
- It should have all the variants

#### Which Component
_calcite-label_

julio8a:
verified on demo pages

PR #2768, opened 2021-08-09
test(label): updated calcite-label demo

ManahariDahal:
**Related Issue:** #2631 

## Summary
- updated _calcite-label_ demo

### DIFF
--- a/src/demos/calcite-label.html
+++ b/src/demos/calcite-label.html
@@ -7,10 +7,41 @@
     <title>Calcite Label</title>
 
     <style>
-      .demo-background-dark {
-        background: #202020;
-        color: white;
-        padding: 1rem;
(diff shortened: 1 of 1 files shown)


julio8a:
Do we need all the labels with the native elements (button, input, etc)?

ManahariDahal:
> Do we need all the labels with the native elements (button, input, etc)?

I think it's a good idea to show the difference between native and calcite. But, I am fine either way.

julio8a:
for now, let's remove those.

bstifle (commented):
Calcite Label Wrapping Calcite Radio Group
and
Calcite Label Wrapping Calcite Radio Group With For

Are using native.... need to use calcite radio button and group
image

Issue #2769, opened 2021-08-09
Enhancement: demo html update to achieve parity with Figma


ManahariDahal:
### Description
This issue aims to achieve parity between the calcite-radio following the similar layout on the Figma HTML Demo page and any usage snippets we provide.

### Acceptance Criteria 
- _calcite-radio_, _calcite-radio-button(rtl)_, _calcite-radio-button-group_, _calcite-radio-button-group(rtl)_, and _calcite-radio-group_ should be combined into one _calcite-radio_ page

#### Which Component
_calcite-radio_

ManahariDahal:
ready to verify!

PR #2771, opened 2021-08-09
test(radio-button): calcite-radio-button demo update

ManahariDahal:
**Related Issue:** #2769 

## Summary
- ~~`checked` attribute not working as of now~~
- `calcite-radio` is isolated to `calcite-radio-button` to make it easy to test


ManahariDahal:
@caripizza this is what I have, something's going on with `checked` or I am missing something.

caripizza:
@ManahariDahal can we make this pr only for calcite-radio-button's demo page? I see the group being added here. (I think you'll want to use the group when all radio-buttons have the same `name` like that for `checked` to work on each. Otherwise consider each label + radio-button its own instance.)

Also if possible, can you add `for` attributes to the labels - that match their nested radio-button's `id` attribute? ([MDN Example](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio))

src/demos/calcite-radio-button.html
```diff
@@ -6,6 +6,44 @@
...
+        font-weight: var(--calcite-font-weight-medium);
+
+      }
+
+      .parent-flex {
+        font-size: var(--calcite-font-size-0);
+      }
+
+      .header {
+        font-size: var(--calcite-font-size-4);
```
caripizza, line 25: @ManahariDahal when are you planning to start combining repeated classes into src/demos/_assets/demos.css for re-use?

ManahariDahal: Even though the `css` code looks similar, there are few minor differences. Once every demo is deployed, I would work on a single branch combining all the similar classes.

eriklharper:
> @caripizza this is what I have, something's going on with `checked` or I am missing something.

I took a look at your changes and at first I was super confused as well, but the reason it looks like it's broken is because the entire group of buttons here, for example have the same name of `s` and because the last checked radio in the html order wins, that's why the last group has a checked option.  You'll need to make each group of 3 radios have a unique `name` for the checked behavior to show properly.

image

src/demos/calcite-radio-button.html
caripizza: Okay cool that works for me

ManahariDahal:
> > @caripizza this is what I have, something's going on with `checked` or I am missing something.
> 
> I took a look at your changes and at first I was super confused as well, but the reason it looks like it's broken is because the entire group of buttons here, for example have the same name of `s` and because the last checked radio in the html order wins, that's why the last group has a checked option. You'll need to make each group of 3 radios have a unique `name` for the checked behavior to show properly.
> 
> image

Sweet! thanks for the feedback.

bstifle (approved):
Lookin Good! 👁

Issue #2507, opened 2021-07-09
Bug: calcite-input-date-picker range event not firing


joeyHarig:
### Actual Behavior
When the user types into the range inputs on the `calcite-input-date-picker`, rather than using the date picker popover ui, the event `calciteDatePickerRangeChange` doesn't fire.

### Expected Behavior
The event `calciteDatePickerRangeChange` should fire when a valid date is typed into either of the rage text inputs, similar to how event `calciteDatePickerChange` fires when a valid date is typed into the `calcite-input-date-picker` when `range="false"`.

### Reproduction Steps or Sample
https://codepen.io/joeyHarig/pen/MWmjWVb?editors=1111

1. Go to the codepen
2. Type a valid date into both range text inputs and notice that no events are logged
3. Use the date picker popover ui to select the start/end dates and notice that the event are logged

### Relevant Info

_Version_: latest

PR #2779, opened 2021-08-10
fix(input-date-picker): When range is true, emit 'calciteDatePickerRangeChange' on input.

driskull:
**Related Issue:** #2507

## Summary

fix(input-date-picker): When range is true, emit 'calciteDatePickerRangeChange' on input. #2507


jcfranco (approved):
💥📅💥

Issue #2726, opened 2021-08-04
Enhancement: add `readonly` property to calcite-input


caripizza:
### Description
Add `readonly` property to calcite-input, ensure it renders with accompanying styles. Per @macandcheese:

>we used to spread any attributes on the input to the rendered native input, so adding readonly would work and the referenced styles would be applied. But I think that was removed.

See [Discussion in Teams](https://teams.microsoft.com/l/message/19:fd15b51dacd24e70895ec1218a54ae06@thread.skype/1628092970609?tenantId=aee6e3c9-711e-4c7c-bd27-04f2307db20d&groupId=56fae21a-9407-4943-859f-a9bfcf0bbad3&parentMessageId=1628092970609&teamName=Calcite%20Design%20System&channelName=Calcite%20Components&createdTime=1628092970609) for additional info/context.

### Acceptance Criteria 
When calcite-input renders with a `readonly` attribute, it should appear with the corresponding "Read-only" styles per Figma.

### Relevant Info 
Ionic reference: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/input/input.tsx#L160
MDN docs: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-readonly

#### Which Component
calcite-input

#### Example Use Case
Per Figma designs:
![image](https://user-images.githubusercontent.com/42423180/128226076-dc2a88b8-fb94-4166-99e4-a9d7dda9b46f.png)

caripizza:
@macandcheese please add anything I missed here. @julio8a @jcfranco we may need to add this one to the current sprint.

macandcheese:
Thanks for adding. Probably a larger discussion around other native attributes we don't explicitly accept should happen - pattern, inputmode, etc could be useful and currently not possible to use.

jcfranco:
@caripizza Added! 
@macandcheese Agreed.

driskull:
> Probably a larger discussion around other native attributes we don't explicitly accept should happen - pattern, inputmode, etc could be useful and currently not possible to use.

Agreed.

This is why I was thinking we could have a `input` slot on `calcite-input` where a user can just slot a native input with all these attributes. Then we could still have `calcite-input` with shadowRoot and maintain native form support as well.

macandcheese:
[...]

PR #2780, opened 2021-08-10
fix(input): Prevent keyboard events from changing values when readOnly is true

driskull:
**Related Issue:** #2726

## Summary

fix(input): Prevent keyboard events from changing values when readOnly is true. #2726


caripizza (approved):
Thanks @driskull LGTM 🥇 . Can you add a couple tests before merging?

Issue #2458, opened 2021-07-07
refactor(Action): match Figma and utilize Tailwind


julio8a:
Description
This component needs to match Figma designs.

Acceptance Criteria
The component appears as it does in Figma.

5/19 QA: Work that needs to be addressed: 
-  add Clear variant to Figma (Bryan)

Related prs:
- pr #1501

Design:
- needs small and large variants
- https://www.figma.com/file/E3SB0i5wPy7AagB7KcjF05/?node-id=1184%3A3828

julio8a:
@bstifle, this one already has s, m, and l. 

Does it need anything else?

bstifle:
i believe the Large size type is wrong, and the spacing between the type and icons are off on small and large

bstifle:
also, there are no examples of action pad or action bar in small or large sizes, would be dope to update the demo there

bstifle:
@asangma @julio8a @macandcheese yo guys. Action type. Currently Small text size is text--2h, Medium is text--2h, and Large is text-0h.

Would like to stay consistent with the global font scaling and get medium to use text--1h

caripizza:
Blocked by #2736

asangma:
I'll have a look at how that affects things.
@bstifle Per your comment, it looks like you're suggesting

| text size | scale |
|:-------:|:-----:|
| --2h? | s |
| --1h | m |
| -0h | l |

Is that correct?

asangma:
@bstifle 
Here's how it would look in an interface.
Thoughts?

![image](https://user-images.githubusercontent.com/12503298/129425822-cfe8aa49-645a-4ed5-ad62-9d65bf450161.png)

asangma:
Broke out the font-size stuff into it's own issue.

julio8a:
https://github.com/Esri/calcite-components/issues/2864

julio8a:
Closing this one.

Issue #2736, opened 2021-08-04
Bug: calcite-input no longer has the same height as calcite-action


AdelheidF:
With next.261. It was fine with beta.60.

`calcite-input` and `calcite-action` have both `scale=s`.

![image](https://user-images.githubusercontent.com/44481292/128270046-9c3f2712-6b03-4e3e-8444-72c68753673e.png)

### Reproduction Steps or Sample

1.

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

bstifle:
@asangma currently the actions are at 32px, 48px, 64px.

Would you be down with getting actions down to 24, 32, 44 like buttons, inputs, selects, etc?

julio8a:
Why would they not follow the height scale from all other components?

julio8a:
If there are no objections with updating the heights of `calcite-action`: S (24px) , M (32px), and L (44px) we can have this one updated this sprint by prioritizing `calcite-action` Figma/tailwind refactor. Will wait until 8/9 for feedback/comments on this.

asangma:
@bstifle will need to see how that affects..kinda everything.

It would have broad effects in existing implementations.

As a note, when Action is not showing text and is just the icon, medium would be 32x32 which I believe is smaller than suggested mobile touch size.

It's also pretty small hit areas for widely separated interfaces like left and right action-bars.

bstifle:
@asangma agreed. pretty much all the layout components utilize action.

another way around this would be to not use actions in tandem with inputs, per the OG issue. use a native button and style accordingly

AdelheidF:
> another way around this would be to not use actions in tandem with inputs, per the OG issue. use a native button and style accordingly

This is a breaking change for a bunch of our components.

AdelheidF:
What about adding a size XS for `calcite-action` that would then match calcite-input? Still a breaking change, but at least it's easy to fix.

macandcheese:
Couldn't actions that are slotted in those locations just use scale="l", just a 4px change from current?

asangma:
I see two things going on here.

1. We could add styles in Input to target a slotted Action or Button.
  - we do this is other places effectively
2. Input currently sets heights based on scale. We should consider letting the height be defined by the font sizes and padding.
3. We can also update Input styling to be more responsive to its slotted content.

asangma:
[...]

PR #2786, opened 2021-08-11
refactor(action): large component needs to match figma design

justinhough:
**Related Issue:** #2458

## Summary

updated large calcite-action to match Figma for sizing and spacing from icon.


julio8a:
For the semantic PR, you'll need to get a new commit to pass that check.
Initial commit is missing `:`
> refactor(action) large component

justinhough:
All the tests passed this time.

caripizza:
This one is blocked by #2736

asangma:
@caripizza created an issue aimed specifically at the font-size that this PR might address.

julio8a:
https://github.com/Esri/calcite-components/issues/2864

src/components/calcite-action/calcite-action.scss
```diff
@@ -64,16 +64,31 @@
...
 
-:host([scale="s"]) .button {
-  @apply p-2;
+:host([scale="s"]) {
+  .button {
+    @apply p-2 text--2h;
+  }
+  .button--text-visible .icon-container {
+    @apply mr-2;
+  }
```
asangma, line 73: If this is set here, we should probably also remove margin styles on `icon-container` I think around line 124.

src/components/calcite-action/calcite-action.scss
```diff
@@ -64,16 +64,31 @@
   }
 }
 
-:host([scale="s"]) .button {
-  @apply p-2;
+:host([scale="s"]) {
+  .button {
+    @apply p-2 text--2h;
```
asangma, line 69: We should set the font-weight to "normal".
"medium" seems pretty heavy here.

asangma (changes requested):
Couple of small things.

src/components/calcite-action/calcite-action.scss
justinhough: Sounds good. I will leave only keep `m-0` on line 124.

src/components/calcite-action/calcite-action.scss
justinhough: Do you figure this would be good only on "small" scale or for all instances of button?

src/components/calcite-action/calcite-action.scss
justinhough: @asangma Updated.

src/components/calcite-action/calcite-action.scss
justinhough: @asangma Updated the small scale only on this one.

asangma: Sorry. I meant for all scales.

src/components/calcite-action/calcite-action.scss
asangma: 🧹

src/components/calcite-action/calcite-action.scss
justinhough: Sounds good. Will make the correction.

justinhough: @asangma Updated for all scales.

asangma (changes requested):
Sorry @justinhough.
Just realized we need to add margin styles for each scale when it's RTL (`calcite--rtl`). 

There's one style there, but it looks like it's for scale M and not specific to that scale.

justinhough:
@asangma Just added the update. Let me know.

asangma:
👀

asangma (approved):
yay

Issue #2764, opened 2021-08-09
Bug: calcite-input type="number" step="any" doesn't support decimals


eriklharper:
### Actual Behavior

Decimals cannot be entered when a number calcite-input's step is set to "any".

### Expected Behavior

Decimals are supported with the "any" step in the native number input, so calcite-input support the same.

### Reproduction Steps or Sample

```
<calcite-input type="number" step="any"></calcite-input>
```
1. Type `123.45`
2. Result: decimal is ignored

### Relevant Info

@eriklharper I am seeing that with `step="any"` a decimal also cannot be inputted. This doesn't seem right, what is intended?

https://codepen.io/mpayson/pen/XWRxXBr

_Originally posted by @ethanbdev in https://github.com/Esri/calcite-components/issues/2009#issuecomment-893678716_

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

AdelheidF:
In what sprint will this be? Map Viewer is waiting on it because we can't upgrade from beta.60 without it.

cc @jcfranco @julio8a

jcfranco:
@eriklharper What's your estimated level of effort for this?

eriklharper:
> @eriklharper What's your estimated level of effort for this?

That should be a fairly quick turnaround, maybe a day or two?

jcfranco:
@eriklharper Thanks! Added to the current sprint, although it may roll over to the next one. Can you update the issue title and description to match our templates?

ethanbdev:
LGTM on next, https://codepen.io/mpayson/pen/XWRxXBr

jcfranco:
@eriklharper Thanks for tackling this and thanks to @ethanbdev for verifying!

PR #2804, opened 2021-08-13
fix(calcite-input): number inputs with step="any" should allow decimals

eriklharper:
**Related Issue:** #2764

This pr allows number inputs with `step="any"` to input decimal values akin to the native number input.

### DIFF
--- a/src/components/calcite-input/calcite-input.tsx
+++ b/src/components/calcite-input/calcite-input.tsx
@@ -483,7 +483,8 @@ export class CalciteInput {
     const decimalSeparator = getDecimalSeparator(this.locale);
     if (
       event.key === decimalSeparator &&
-      (this.step === "any" ? true : isValidDecimal(this.step))
+      (this.step === "any" || isValidDecimal(this.step))
+      // (this.step === "any" ? true : isValidDecimal(this.step))
     ) {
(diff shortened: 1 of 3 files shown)


eriklharper:
test is forthcoming, stay tuned!

src/components/calcite-input/calcite-input.tsx
```diff
@@ -484,7 +483,7 @@ export class CalciteInput {
     const decimalSeparator = getDecimalSeparator(this.locale);
     if (
       event.key === decimalSeparator &&
-      isValidDecimal(this.step === "any" ? 1 : (this.step as number))
+      (this.step === "any" ? true : isValidDecimal(this.step))
```
jcfranco, line 486: Nitpick: This can be simplified to `this.step === "any" || isValidDecimal(this.step)`

jcfranco (approved):
✨🥇 ✨

src/components/calcite-input/calcite-input.tsx
eriklharper: Yeah for some reason I had it that way and it was allowing decimals in integer-only mode, but now that I add it back in I'm seeing it work.  Must have been a dev server problem where the changes weren't loading right or something.

Issue #2019, opened 2021-04-22
Research: explore using additional Screener APIs to improve screenshot testing


eriklharper:
### Background 

Screener provides additional tools to perform visual testing on more complicated components, such as dropdowns or modals that require interaction to display additional UI.  

### Desired Outcome 

- [x] Explore how the Screener steps API would change our current testing (https://github.com/Esri/calcite-components/pull/3141, https://github.com/Esri/calcite-components/pull/3140)
- [x] If we do incorporate the steps API, 
  - [x] add documentation for usage guidelines:
    - background usage guidelines
    - when to use it (pros and cons)
    - how to use it
  - [x] if applicable, create component-specific issues to use the steps API

driskull:
I merged storybook changes for popover/tooltip

eriklharper:
Wiki Doc Page created here: https://github.com/Esri/calcite-components/wiki/Screenshot-Testing

PR #2805, opened 2021-08-13
test(tooltip, popover): Refactor storybook screenshot diffs

driskull:
**Related Issue:** #2019

## Summary

test(tooltip, popover): Refactor storybook screenshot diffs

### How this works

This PR uses [screener-storybook](https://github.com/screener-io/screener-storybook) to test for visual changes to or storybook stories and introduces the use of the [screener storybook Steps](https://screener.io/v2/docs/test-interactions_ to test for interaction changes.

The Steps API allows you to interact or call methods before taking a snapshot of the story.

This will allow us to replace all the RTL and Dark theme stories by calling a method to change to RTL or Dark theme in the same story and take a snapshot. 

For example, instead of setting a knob to open a component, we can click the component's button to open it.

### This PR:

Uses screener-storybook steps: https://screener.io/v2/docs/test-interactions within storybook to take screenshots after interactions as well as changes for dark theme and RTL.

- Replaces dark and RTL stories for the Popover and Tooltip with Screener steps
- Introduces a storybook helper to setup tests for a story
- Exposes a couple strings of JS scripts for toggling theme and setting knobs on a screener run

### The plan

- Do this for all other storybook stories.
- Replace all "Dark" theme and "RTL" stories with screener steps
- Remove any stories where content doesn't change
- Use screener interactions (click/hover) where applicable instead of separate stories or setting knob values.

### Proposed plan for testing

- In order to test the changes, you have to look at the screener build.
- You should see new screenshots with the name of the snapshots that are provided in the steps code.

### Resources

Screener-runner executeScript types: https://github.com/screener-io/screener-runner/blob/ad615d12691c52ac13d069df960f204725486cfd/src/steps.js


driskull:
@eriklharper I can't seem to get it working. Do you have any ideas?

eriklharper:
> @eriklharper I can't seem to get it working. Do you have any ideas?

Good question, I was following your changes there for a bit.  Not sure, but based on this error, it seems the story is not being exported correctly. 

image

I'm wondering, do you think it has to do with the fact that the template literal exported here isn't tagged with `html` like other stories are?  https://github.com/Esri/calcite-components/blob/4b367be6103d1f7704f64cfe5764e13d49465e7d/src/components/calcite-tooltip/calcite-tooltip.stories.ts#L41

src/components/calcite-tooltip/calcite-tooltip.stories.ts
```diff
@@ -37,23 +39,29 @@ export default {
...
+      offset-distance="${number("offset-distance", 6)}"
+      offset-skidding="${number("offset-skidding", 0)}"
+      ${boolean("open", true)}
+    >
+      ${contentHTML}
+    </calcite-tooltip>
+  </div>
+`;
+
+Simple.decorators = [
```
driskull, line 58: Good news, this works. Maybe we can create a utility to handle this because it seems the `typings` for Story aren't there and creating the fragment isn't very pretty. We can just have a utility that we pass the steps and it handles setting the decorators on the story, What do you think @jcfranco and @eriklharper?

jcfranco: Nice! 

Your suggestion seems doable. Any idea on what the API & usage would look like?

eriklharper: Awesome dude, thanks for following up on Loyal's response.

eriklharper: We'll definitely want to write a utility to make doing the steps easier and reduce boilerplate.

driskull:
@eriklharper @jcfranco can you review?

src/components/calcite-popover/calcite-popover.stories.ts
```diff
@@ -56,70 +63,34 @@ export default {
...
+        offset-skidding="${number("offset-skidding", 0)}"
+        ${boolean("open", true)}
+        text-close="${text("text-close", "Close")}"
+      >
+        ${contentHTML}
+      </calcite-popover>
+    </div>
+  `,
+  new Steps()
+    .wait(calciteHydratedSelector)
```
eriklharper, line 86: ☝️ I'm wondering if this step can be included in the `stepStory` decorator since we'll want to wait for the hydrated class to appear before doing any other step in all components?

src/components/calcite-popover/calcite-popover.stories.ts
```diff
@@ -56,70 +63,34 @@ export default {
...
+    .wait(popperTimeoutMS)
+    .snapshot("Default")
+    .executeScript('window.location.href = "?path=/story/components-popover--simple&knob-dismissible=true"')
+    .snapshot("dismissible")
+    .rtl()
+    .snapshot("Rtl")
+    .ltr()
```
eriklharper, line 93: Since `ltr` is the default, could we add the `rtl` step last so that we don't have to call `ltr` again?

eriklharper (commented):
This is a great proof-of-concept for using Steps, but I think instead of rewriting the basic stories "simple", "rtl" and "dark" with the new `stepStory` decorator (since Screener already is capturing these since they are separate stories), I think we should only use `stepStory` when there is a need to perform an interaction before taking a screenshot, such as clicking or keyboard events.  What do you all think?  Should we continue writing separate stories or just have every single component have a basic story and then use steps to interact with it where necessary?

src/components/calcite-popover/calcite-popover.stories.ts
driskull: hmm that would test dark in RTL though. I think this needs to be flexible.

src/components/calcite-popover/calcite-popover.stories.ts
driskull: I'm not sure we can add it to the front there.

driskull:
> This is a great proof-of-concept for using Steps, but I think instead of rewriting the basic stories "simple", "rtl" and "dark" with the new stepStory decorator (since Screener already is capturing these since they are separate stories), I think we should only use stepStory when there is a need to perform an interaction before taking a screenshot, such as clicking or keyboard events. What do you all think? Should we continue writing separate stories or just have every single component have a basic story and then use steps to interact with it where necessary?

I think this depends on the point of the storybook. Is it a developer testing tool, a public testing tool, or a diff tool? Or some combination of the above. I still think we need to define its purpose.

driskull:
We can always add a knob for `darkTheme` and `rtl` instead of having separate stories. The stories here are exactly the same code with one attribute changed (dir/class)

eriklharper:
> We can always add a knob for `darkTheme` and `rtl` instead of having separate stories. The stories here are exactly the same code with one attribute changed (dir/class)

I would prefer this approach.  It feels redundant to have multiple stories that render the exact same HTML just to have a way to display RTL and Dark theme or other variants.  That is the purpose of knobs and other features built-in to Storybook itself, we just need to take advantage of them (if and where we can).

driskull:
Actually, I don't think a knob is right for `dir` or `theme`. These should really be applied at the storybook level since we don't want to add them on the component themselves.

https://storybook.js.org/addons/storybook-addon-themes
https://storybook.js.org/addons/storybook-dark-mode

https://storybook.js.org/addons/@pxblue/storybook-rtl-addon

I'll remove the knobs for rtl/theme in this PR>

eriklharper:
> Actually, I don't think a knob is right for `dir` or `theme`. These should really be applied at the storybook level since we don't want to add them on the component themselves.
> 
> https://storybook.js.org/addons/storybook-addon-themes
> https://storybook.js.org/addons/storybook-dark-mode
> 
> https://storybook.js.org/addons/@pxblue/storybook-rtl-addon

Yeah we can just toggle the URL with `&globals=backgrounds.value:!hex(202020)` for dark and `&globals=backgrounds.value:transparent` for light.

eriklharper:
> I think this depends on the point of the storybook. Is it a developer testing tool, a public testing tool, or a diff tool? Or some combination of the above. I still think we need to define its purpose.

I guess this kind of depends on how useful people find the Storybook versus the official documentation site, but I think because we have a separate doc site, that frees up Storybook to be setup how we need it to be to accomplish our testing goals.

driskull:
If this is approved, we can setup a bulk issue to update all existing component stories to use steps and remove their stories that are for dark theme and RTL. (RTL and DarkTheme const stories).

driskull:
@jcfranco @eriklharper can you review this one? I'd like to get it in so we can proceed updating other components.

.storybook/helpers.ts
```diff
@@ -14,3 +17,37 @@ export const boolean = (prop, value, standalone = true) => {
   const attrValue = standalone ? "" : `="${knob}"`;
   return `${propValue}${attrValue}`;
 };
+
+export interface Story {
+  (): string;
+  decorators?: ((Story: Story) => DocumentFragment)[];
+}
+
+export const setKnobScript = ({ story, knob, value }: { story: string; knob: string; value: string }) => {
```
eriklharper, line 26: nitpick: perhaps the name could be `setKnobURL`.  The word `script` makes it sound like it adds a `<script>` tag to the page.

.storybook/helpers.ts
```diff
@@ -14,3 +17,37 @@ export const boolean = (prop, value, standalone = true) => {
...
+  return `window.location.href = "?path=/story/${story}&knob-${knob}=${value}"`;
+};
+
+export const setThemeScript = (value: ThemeName) => `${THEMES.map(function (theme) {
```
eriklharper, line 30: Could also drop the word `script` from the name and just make it `setTheme`.

eriklharper (approved):
This is really cool, I can't wait to start using these new story utilities!

src/components/calcite-popover/calcite-popover.stories.ts
```diff
@@ -56,70 +56,35 @@ export default {
...
+      >
+        ${contentHTML}
+      </calcite-popover>
+    </div>
+  `,
+  createSteps("calcite-popover")
+    .snapshot("Default")
+    .click("#reference-element")
+    .snapshot("Open")
+    .executeScript(setKnobScript({ story: "components-popover--simple", knob: "dismissible", value: "true" }))
```
jcfranco, line 82: What's the story (badumtsss 🥁 ) for cases that need rich data and won't be able to set value via an attribute? Would we add a specific story for that use case that may or may not have steps associated with it?

jcfranco (commented):
This is looking really nice! 

One question before proceeding further, have you had a chance to test and compare run times before and after the steps API change? Maybe we can do some local, scoped test runs. cc @eriklharper

jcfranco:
This PR is related to https://github.com/Esri/calcite-components/issues/2019, right?

src/components/calcite-popover/calcite-popover.stories.ts
```diff
@@ -56,70 +56,35 @@ export default {
...
-          ${contentHTML}
-        </calcite-popover>
-      </div>
-    `;
-};
-
-DarkMode.story = {
-  parameters: { themes: themesDarkDefault }
-};
+export const Simple = stepStory(
```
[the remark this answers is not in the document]
macandcheese: Correct - the docs on developers.arcgis.com do not use stories, safe to change here in that regard. 👍

.storybook/helpers.ts
```diff
@@ -14,3 +17,37 @@ export const boolean = (prop, value, standalone = true) => {
...
   return `${propValue}${attrValue}`;
 };
+
+export interface Story {
+  (): string;
+  decorators?: ((Story: Story) => DocumentFragment)[];
+}
+
+export const setKnobScript = ({ story, knob, value }: { story: string; knob: string; value: string }) => {
+  return `window.location.href = "?path=/story/${story}&knob-${knob}=${value}"`;
```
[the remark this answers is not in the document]
driskull: Correct. We could enhance this to set multiple knobs at once if needed. Ideally, only one is modified at a time though.

.storybook/helpers.ts
```diff
@@ -14,3 +17,37 @@ export const boolean = (prop, value, standalone = true) => {
...
+`;
+
+export const createSteps = (componentSelector: string): Steps => {
+  return new Steps().wait(`${componentSelector}[calcite-hydrated]`);
+};
+
+export const stepStory = (story: Story, steps: Steps): Story => {
+  const stepsDecorator = (Story: Story) => {
+    const node = document.createRange().createContextualFragment(Story());
+    (node as any).steps = steps.end();
```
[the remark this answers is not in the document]
driskull: Not that I could find. Seems screener doesn't have good typings.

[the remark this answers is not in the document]
driskull: This is what was provided from them for a sample.

src/components/calcite-popover/calcite-popover.stories.ts
```diff
@@ -56,70 +56,35 @@ export default {
...
+        offset-skidding="${number("offset-skidding", 0)}"
+        ${boolean("open", false)}
+        text-close="${text("text-close", "Close")}"
+      >
+        ${contentHTML}
+      </calcite-popover>
+    </div>
+  `,
+  createSteps("calcite-popover")
+    .snapshot("Default")
```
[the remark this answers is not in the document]
driskull: I think once we approve this method, I can create some docs and guidelines.

src/components/calcite-popover/calcite-popover.stories.ts
driskull: If the data needs to change, then a separate story is the way to go. We don't want screener to be handling changing of data/content.

driskull:
> One question before proceeding further, have you had a chance to test and compare run times before and after the steps API change? 

I haven't, but it may be faster since we are removing RTL/Dark theme stories. Can someone test?

driskull:
Updated PR content.

eriklharper:
> This PR is related to #2019, right?

Yes, this PR is an implementation of the ideas in #2019.

PR #2809, opened 2021-08-13
Revert "refactor(inline-editable): match Figma and utilize Tailwind"

julio8a:
Reverts Esri/calcite-components#2776


caripizza:
@julio8a I think we're okay and this isn't needed any more, it's just that Screener flipped out.

julio8a:
Thanks for checking @caripizza

PR #2822, opened 2021-08-14
build: transpile numeric separators from color package

jcfranco:
**Related Issue:** N/A

## Summary

The latest `color` package is on the bleeding-edge and may cause issues for clients that have a build process that does not target ES2021. This PR tweaks the build to transpile numeric separators from the package. 

cc @dasa


driskull (approved):
👍

Issue #2659, opened 2021-07-27
Enhancement: radio button white background when unchecked


COV-GIS:
### Description

Unchecked radio buttons would have better contrast and IMHO look better with a white (or corresponding in dark theme) background like checkbox does.

![image](https://user-images.githubusercontent.com/14297903/127220040-3062b361-312d-4dfc-a210-fdbfc3fbf2c9.png)

Visually doesn't seem to match other controls.

#### Which Component

radio button

jcfranco:
@Esri/calcite-design-representatives Can you chime in on this one? Seems like a low-effort one for this sprint.

caripizza:
Agreed. I think we just need to add `bg-foreground-1` to the radio [on this line](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-radio/calcite-radio.scss#L4). Should we assign this one to Justin @julio8a?

benelan:
I can't tell if the unchecked colors are the same in this sample:
https://codepen.io/benesri/pen/PojJbpw?editors=1000

@julio8a can you take a quick look? It may just be my eyes playing tricks on me due to the different outlines.

julio8a:
I can verify this one, thanks Ben.

julio8a:
Verified on master, demo pages.

PR #2829, opened 2021-08-16
refactor(radio): background color when unchecked

justinhough:
**Related Issue:** #2659

## Summary

Added `bg-foreground-1` to radio when unchecked for better visibility on light and dark themes.


bstifle (approved):
lookin good! 🔘

PR #2830, opened 2021-08-16
fix(popover, tooltip): Reposition open popover and tooltip on initial load. #2446

driskull:
**Related Issue:** #2446 #2831

## Summary

fix(popover, tooltip): Reposition open popover and tooltip on initial load. #2446

cc @caripizza


caripizza (approved):
Thanks @driskull 👍  🥇

Issue #2577, opened 2021-07-20
Bug: calcite-block-section and calcite-label+calcite-switch toggle buttons not aligned vertically


AdelheidF:
### Actual Behavior

Beta.60

![image](https://user-images.githubusercontent.com/44481292/126394532-5342c347-0677-452e-9726-dbeff84d5075.png)

Top and bottom elements are `calcite-block-section`. Center is a `calcite-label` with a `calcite-switch`.

1. if you look at the three toggle buttons they're not aligned vertically
2. if you look at the very first toggle button it's not horizontally aligned with the text.

I can add a margin to the center element. Not sure if this is expected though.

The switch inside the block-section uses a margin:
![image](https://user-images.githubusercontent.com/44481292/126394982-d048e26f-2f92-4db6-bb32-4e45605e2fa1.png)

### Reproduction Steps or Sample

1.

### Relevant Info

_Version_: `@esri/calcite-components@1.0.0-beta.60`

- [ ] CDN
- [ ] NPM package

jcfranco:
Looks like label or block-section needs to tweak the trailing space. @Esri/calcite-design-representatives Do you think this is the case or are there other things to consider?

caripizza:
Which version? This might be fixed in latest `next` [per my comment in Justin's pr](https://github.com/Esri/calcite-components/pull/2681#issuecomment-892215276)

jcfranco:
@caripizza Thanks for the update! 

@AdelheidF Can you update the description with the version and also check if this is fixed in `next`?

AdelheidF:
Tested with next.265.

Horizontal alignment is correct now. Vertical is not.

![image](https://user-images.githubusercontent.com/44481292/128778991-9e746dda-13d3-418d-8b1b-f0990489c729.png)

caripizza:
Where is the code snippet so we can reproduce the issue?

AdelheidF:
The test case is part of Map Viewer.

Can you not reproduce it with a `calcite-block-section` and a `calcite-label` with a `calcite-switch`?

caripizza:
@AdelheidF we have various devs working on style issues like this, and some don't have access to what you're looking at. Any code you can give us will help us fix the issue faster, thanks.

julio8a:
CodePen example:
https://codepen.io/julio8a/pen/gOWJwvQ

justinhough:
[...]

PR #2839, opened 2021-08-17
refactor(calcite block and switch) not aligned in label

justinhough:
**Related Issue:** #2577

## Summary

Fixing horizontal alignment when calcite-switch is used within a label using `layout="inline-space-between"` in context with `calcite-block-section`.


asangma:
@julio8a I think we may still need to fix this on our end.

julio8a:
Let's follow up on this

julio8a:
closing per Alan's comment https://github.com/ArcGIS/calcite-presentation-icons

PR #2843, opened 2021-08-17
ci(screener): enable alwaysAcceptBaseBranch

jcfranco:
**Related Issue:** N/A

## Summary

`screener-runner` has [an option](https://github.com/screener-io/screener-runner/#:~:text=alwaysAcceptBaseBranch) that sounds like it would help with some visual diff inconsistencies we've noticed between PRs:

> alwaysAcceptBaseBranch: Option to automatically accept new and changed states in base branch. Assumes base branch should always be correct.

This PR enables this option for testing. 

If this ends up working the way we need it, we can schedule a date to ensure all PRs have the latest version of `master` merged into them. In theory, the base for all screenshot diffs would be the same and the only visual diffs we get should be related to the PR.

@eriklharper @caripizza For us to test this, would we need to accept some invalid changes in master and see how Screener behaves with a PR with small or no visual changes?

cc @driskull @julio8a

### DIFF
--- a/screener.config.js
+++ b/screener.config.js
@@ -1,4 +1,5 @@
 module.exports = {
+  alwaysAcceptBaseBranch: true,
   projectRepo: 'Esri/calcite-components',
   storybookConfigDir: '.storybook',
   storybookStaticDir: './__docs-temp__',


eriklharper:
> For us to test this, would we need to accept some invalid changes in master and see how Screener behaves with a PR with small or no visual changes?

That sounds like a reasonable way to vet this change, but we can also just turn it on and then observe PRs that come in after it has been merged into master, and also have a human do a once-over on the baseline accepted screenshots to ensure that they look good as well.

eriklharper (approved):

driskull (approved):
do it

jcfranco:
> We should also just quality check the screener builds on the master branch like this one for example: https://screener.io/v2/dashboard/1leJJn.Esri-calcite-components/master

100% agreed. 

> we can also just turn it on and then observe PRs that come in after it has been merged into master, and also have a human do a once-over on the baseline accepted screenshots (ones that are not included in any diff) to ensure that they look good as well. 

Does this mean you're volunteering? 😂  j/k That's a lot of potential PRs for us to look out for (both before and after this lands). How about we schedule a time slot to let the team know about this to get volunteers and to pause installs, select a couple of PRs to install and compare results? Maybe this Friday?

eriklharper:
> Does this mean you're volunteering? 😂 j/k That's a lot of potential PRs for us to look out for (both before and after this lands). How about we schedule a time slot to let the team know about this to get volunteers and to pause installs, select a couple of PRs to install and compare results? Maybe this Friday?

Sounds good.  I think we won't really know what we're dealing with until we start seeing it in action, at which point we can adjust our plans appropriately.

jcfranco:
Agreed. This is all based on the prop's description since I didn't find any other resources about it. Hope it works as expected. 🤞🏼 

I'll communicate this to the team.

Issue #2801, opened 2021-08-12
Enhancement: Scrim - render default slot


asangma:
### Description
Render default slot in Scrim.
This will allow users more flexibility for loading, success, error, etc. states and messages.

```
<calcite-scrim>
  <calcite-notice color="red">
    <h4 slot="title">Something went wrong.</h4>
    <p slot="message">The vuvuzella did not honk to the required volume.</p>
    <a href="vuvuzella-technique.com" slot="link">Learn how to make it loud.</a>
  </calcite-notice>
</calcite-scrim>
```

### Acceptance Criteria 
* Render the default slot.
* Add CSS to retain some padding around anything that is added to the default slot
* Possibly also add some CSS for some reasonable/expected content, e.g. Notice or "modal" like use-cases

### Relevant Info 
Sometimes the Alert is not the right solution. Especially since it has a pre-defined (and appropriate) app-level message.

#### Which Component
Scrim

#### Example Use Case
Using a determinate loader in a scrim.

![image](https://user-images.githubusercontent.com/12503298/129755441-7b0938c3-7e95-4f25-bffd-74b54c723633.png)

asangma:
Another example use case would be

1. either set `loading` to a Panel or manually add Scrim to the Panel
2. when the process is done, show status (error or success), remove `loading` prop from Panel or Scrim
3. add Notice to Scrim
4. based on timing set in app, remove Scrim.

asangma:
@jcfranco @caripizza This might be needed/helpful relatively soon and the update is low effort and low risk.

jcfranco:
Added to the following sprint.

asangma:
Thanks @jcfranco!

asangma:
Do you want me to handle the PR?

jcfranco:
That'd be awesome. Feel free to grab it and remove the `help wanted` label. :)

asangma:
Will do.

julio8a:
Before moving forward, we should have a design discussion around this.

jcfranco:
@julio8a Sounds good. Sorry for overlooking that step. 😅

asangma:
@julio8a There's not really a design since it's an empty slot.

But for a design review, I would dig syncing with @bstifle on default padding for when a user drops in a known component.

julio8a:
Can a user drop anything in there? Seems like we're opening up some bad UX practices on this. Can they drop an accordion in there with content or any other component?

[...]

PR #2846, opened 2021-08-17
feat(scrim): render default slot

asangma:
**Related Issue:** #2801

## Summary
* updates render to include default slot
* updates tests
* updates scrim bg color per @bstifle

### DIFF
--- a/src/components/calcite-scrim/calcite-scrim.scss
+++ b/src/components/calcite-scrim/calcite-scrim.scss
@@ -21,7 +21,6 @@
 .scrim {
   @apply flex
   absolute
-  select-none
   overflow-hidden
   inset-0
   content-center
--- a/src/components/calcite-scrim/calcite-scrim.scss
+++ b/src/components/calcite-scrim/calcite-scrim.scss
@@ -25,7 +25,13 @@
   overflow-hidden
   inset-0
   content-center
-  items-center;
+  items-center
+  flex-col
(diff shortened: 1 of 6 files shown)


src/components/calcite-scrim/calcite-scrim.scss
```diff
@@ -25,7 +25,13 @@
   overflow-hidden
```
driskull, line 25: Should the `select-none` be removed from scrim if we allow things inside it?

driskull (approved):

src/components/calcite-scrim/calcite-scrim.scss
asangma: That's a good point.
Right now, content in the default slot is still interact-able (in Chrome) and anything "under" the scrim remains non-interact-able.

I can add some tests.

asangma: @driskull I added a couple tests to check for click. 

And wow...I totally misunderstood your suggestion.
Yes...we should remove `user-select: none`. Will do.

asangma:
@ManahariDahal Would you mind having a look at the Scrim demo page and make sure I didn't break too much stuff?

ManahariDahal:
> @ManahariDahal Would you mind having a look at the Scrim demo page and make sure I didn't break too much stuff?

looks good!

Issue #2492, opened 2021-07-08
Docs: ColorPicker missing Usage section


dasa:
### Description 

Most controls seem to have a Usage section, like [here](https://esri.github.io/calcite-components/?path=/docs/components-controls-checkbox--simple), but ColorPicker [does not](https://esri.github.io/calcite-components/?path=/docs/components-controls-colorpicker--simple).

### Which Component

ColorPicker

benelan:
Hi @dasa, I added Color Picker [usage examples](https://github.com/Esri/calcite-components/tree/master/src/components/calcite-color-picker/usage), do you want to verify before I close the issue? I also created #2860 to add usage examples to the rest of the components that don't have any.

jcfranco:
Verified locally!

PR #2855, opened 2021-08-18
docs(color-picker): add usage examples

benelan:
**Related Issue:** #
- #2492 

## Summary
Add usage examples for the calcite-color-picker documentation.

### DIFF
--- a/src/components/calcite-color-picker/readme.md
+++ b/src/components/calcite-color-picker/readme.md
@@ -7,7 +7,7 @@
 ### Basic
 
 ```html
-<calcite-color-picker dir="ltr" scale="m" value="#b33f33" appearance="default" />
+<calcite-color-picker dir="ltr" scale="m" value="#b33f33" appearance="default"></calcite-color-picker>
 ```
 
 ### Minimal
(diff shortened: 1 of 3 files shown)


src/components/calcite-color-picker/readme.md
```diff
@@ -2,6 +2,29 @@
...
+```
+
+### Minimal
+
+For a minimal design, you can hide unused color formats and options:
+
+```html
+<calcite-color-picker
+  appearance="minimal"
+  allow-empty=""
```
driskull, line 20: These boolean attributes shouldn't have values assigned to them right?

benelan: That was my thought as well. However in [storybook](https://esri.github.io/calcite-components/?path=/story/components-controls-colorpicker--simple) if you check the 'hide-saved' box and go to the HTML snippet it has the empty string. So that's what I went with. I can change it and log an issue for the storybook if it shouldn't have a value.

driskull: Yeah, that's kind of a known storybook limitation but lets remove it from our docs.

benelan: Fixed it. Thanks for clarifying, storybook had me a bit confused there.

src/components/calcite-color-picker/readme.md
```diff
@@ -2,6 +2,22 @@
 
 <!-- Auto Generated Below -->
 
+## Usage
+
+### Basic
+
+```html
+<calcite-color-picker dir="ltr" scale="m" value="#b33f33" appearance="default" />
```
driskull, line 10: Can we add the closing tag? The closing tag is required for custom elements. I think this will only work in JSX.

driskull (approved):
Looks good after closing tags added

Issue #2854, opened 2021-08-18
Bug: Add types to public events EventEmitter


samMatenaer:
Many components are either missing or have incomplete event types which makes causes a headache with handling events in typescript. 

For example, here is an excerpt from my code using the calcite-components-react package
```ts
onCalciteInputInput={(evt) => {
    // evt currently typed as CustomEvent<any>
    const typedEvent = evt as CustomEvent<{ value: string }>;
    const newValue = typedEvent.detail.value;
}}
```
You can see the return type on calciteInputInput is currently `any`, when it should have a specific type.

This is because the [EventEmitter has no type provided](https://github.com/Esri/calcite-components/blob/c87f343651f55dc4ab0031b9f35af74da70dca49/src/components/calcite-input/calcite-input.tsx#L336), so it auto-casts to `any`

### Actual Behavior
Current calciteInputInput definition:
```ts
 @Event({ cancelable: true }) calciteInputInput: EventEmitter;
```

### Expected Behavior

Proposed definition:
```ts
  @Event({ cancelable: true }) calciteInputInput: EventEmitter<{
    element: HTMLInputElement | HTMLTextAreaElement;
    value: string;
    nativeEvent: any;
  }>;
```

I'd like to take this task on if it's something others think would be worth doing

jcfranco:
Thanks for bringing this up. We need to do an audit of our custom events as we will be removing any custom event payload that is not already on the component itself. 

For the `calciteInputInput` case, you can to through the event's `target`/`currentTarget` to get the value: https://codepen.io/jcfranco/pen/eYRmJej?editors=0010

cc @benelan

driskull:
It seems all events have types now. @geospatialem can you confirm?

geospatialem:
Verified on the `master` branch. 👍🏻

PR #2858, opened 2021-08-19
refactor: add types to EventEmitter definitions

samMatenaer:
**Related Issue:** #2854 

## Summary
Previously, event emitters were not defined with a type argument, which caused them to autocast to `EventEmitter<any>`. This PR contains EventEmitter definitions with strict types to maintain types through the emit() function.

All types are based on the arguments currently being passed to emit(). There are no changes in functionality, just passing along types through the events instead of using `any`

### DIFF
--- a/BOILERPLATE_COMPONENT.md
+++ b/BOILERPLATE_COMPONENT.md
@@ -70,7 +70,7 @@ export class CalciteExample {
   // --------------------------------------------------------------------------
 
   @Event()
-  calciteExampleEvent: EventEmitter;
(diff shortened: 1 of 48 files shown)


driskull (approved):
@samMatenaer this looks good. Not sure why the build is failing though.

samMatenaer:
@driskull 
i believe it's because i made the PR in a fork of the repo, as i do not have write access to create a branch here. the issue seems to be the apiKey is not set when pulling from my fork

Issue #2706, opened 2021-08-03
Bug: calcite-slider histogram not taking host width into account


jcfranco:
### Actual Behavior

The slider histogram has a fixed width and is not placed correctly based on the slider width.

### Expected Behavior

The histogram should align with the data values regardless of the slider's size.

### Reproduction Steps or Sample

1. Open https://codepen.io/jcfranco/pen/jOmpMLR?editors=1000
2. Notice how the histograms of the 3 differently-sized sliders have identical widths.

### Relevant Info

_Version_: `@esri/calcite-components@1.0.0-beta.60`

- [x] CDN
- [ ] NPM package

This is a regression from beta 59, possibly from https://github.com/Esri/calcite-components/commit/9dd3812dc0f1294163e46b2f608272e27f6fa0d5.

caripizza:
I happened upon this issue while investigating #2457. It looks like this started happening in 1.0.0-next.236, after setting `shadow: true` in [calcite-graph.tsx](https://github.com/Esri/calcite-components/pull/2423/files#diff-a8e20ed09e8a382be826d7c02796c4d0f4eb642b59d466e2375b63cf02214967R9).

The [`svg` slider selector style override](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-slider/calcite-slider.scss#L352) is no longer applying due to it being unreachable in graph's shadow dom. 

Should [graph's svg width](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-graph/calcite-graph.tsx#L85) be `"100%"` instead, so it automatically stretches the width of the container it's nested inside?

Or, can we add a new optional prop like `matchHostWidth` to graph, so slider can pass it as a truthy prop to the component? (in graph we would do `width={!this.matchHostWidth ? width : "100%"}`)

caripizza:
Related: #2948

jcfranco:
Verified w/ repro case!

Issue #2457, opened 2021-07-07
refactor(Slider): match Figma and utilize Tailwind


julio8a:
Description
This component needs to match Figma designs.

Acceptance Criteria
The component appears as it does in Figma.

Linked issue:
-#2224

Related issues:
- #720

Design:
- Medium is good. Small and Large need to be added
- https://www.figma.com/file/E3SB0i5wPy7AagB7KcjF05/?node-id=0%3A15235

caripizza:
Self-assigning and adding to next sprint, per conversation with @julio8a.

caripizza:
@bstifle There are no histogram variants in Figma for slider, does it need to look the same as it does now?

bstifle:
@caripizza 
![image](https://user-images.githubusercontent.com/14083530/129968702-d7064038-7c43-44b1-b996-6dde0d20419a.png)

caripizza:
@bstifle thanks, should it look like that with the color stops in action too? (ie., solid b-bb-30 for the highlighted range?)

caripizza:
@julio8a should we create an issue in the ui-kits repo to add Slider's histogram variant(s) to Figma?

julio8a:
Good catch, yes, I'll create and link to this issue.

benelan:
@macandcheese can you please help me verify this design issue?

macandcheese:
@benelan done - thanks!

Issue #913, opened 2020-08-28
Enhancement: Slider: Align with edge of container


macandcheese:
### Description
Currently a slider at min / max extends past the edge of the container, which causes some misalignment when presented in a form. 

### Acceptance Criteria 
Compensate for the extra widht of the slider handle and make it align with the edge of container.

|   |   |
|---|---|
|  Current | Desired  |
| <img width="512" alt="Screen Shot 2020-08-28 at 10 54 31 AM" src="https://user-images.githubusercontent.com/4733155/91600619-1db5d280-e91d-11ea-89e9-e212f5d985a8.png">  |  <img width="512" alt="Screen Shot 2020-08-28 at 10 54 20 AM" src="https://user-images.githubusercontent.com/4733155/91600625-23131d00-e91d-11ea-853c-41932e2174e2.png"> |

#### Which Component
Slider

julio8a:
Is this one still valid? If so, what's the status? Maybe this is something I can work on or @bstifle 

Seems like a low priority one though and might get fixed with the spacing scale that gets implemented to all the compnents?

macandcheese:
Wouldn't get fixed with spacing, it's about where the slider rests at the high / low bounds. You can see here: https://esri.github.io/calcite-components/?path=/story/components-label--wrapping-components-other-than-input

Low priority but should keep open so we can address it at some point.

julio8a:
Thanks @macandcheese

jcfranco:
Verified w/ latest storybook: https://esri.github.io/calcite-components/?path=/story/components-label--wrapping-components-other-than-input

![Screen Shot 2021-09-09 at 1 27 05 PM](https://user-images.githubusercontent.com/197440/132757918-1f2ffde5-c72a-4a45-8782-732c02e4b14b.png)

PR #2867, opened 2021-08-19
fix(slider): align slider handle to container, ensure histogram stretches full width

caripizza:
**Related Issue:** #913, #2706

## Summary
Before implementing the Figma scales in slider (#2457), I'd like to address these issues:
- ensure track handle is x-aligned to edge of container
- make slider's histogram fill its host container's width. To do so, I added `data-style` tag to slider's calcite-graph, and moved its svg `width: 100%;` style override into calcite-graph's stylesheet so it works again. (Thought about using [an optional prop](https://github.com/Esri/calcite-components/issues/2706#issuecomment-900718476) in graph, but this seemed easier.) 
- changed slider's histogram color to `--calcite-ui-foreground-3` (I find this color more visible than `--calcite-ui-foreground-2` for the unselected region, especially in its light theme but I can change this back if we want.)

#### Screenshots
**_before_**
image
image

**_after_**
image
image


src/components/calcite-graph/calcite-graph.scss
```diff
@@ -7,3 +7,12 @@
   stroke: transparent;
   @apply m-0 p-0 block;
 }
+
+:host([data-style="slider-histogram"]) svg {
+  @apply w-full;
```
jcfranco, line 12: Should this be baked into the graph or do we have use cases where we don't want the SVG to occupy the full width and the highlight color set?

jcfranco (commented):
This is looking great, @caripizza! 🎉 🍕

src/components/calcite-graph/calcite-graph.scss
caripizza: I wondered about this too. At first I added an optional `matchHostWidth` prop as a flag to graph for it to stretch the full width of its container if desired. (I went this route instead mainly because it seemed to be the way slider's histogram was originally styled.)

jcfranco: @paulcpederson as the historical parent of `calcite-graph`, can you chime on this? 📈 👶🏼

paulcpederson: the width is always full in the slider, which was the original use case.

jcfranco: Sorry for the delay, but is there a use case where the SVG wouldn't need to fill the whole width? If not, shouldn't it always be full width and up to the host component to modify its width?

jcfranco:
@caripizza Let's move forward with this for now since the graph sizing is handled internally. Based on https://github.com/Esri/calcite-components/pull/2867#discussion_r698616607, can you create a follow-up issue to revisit graph's default SVG width?

jcfranco (approved):
🎉 🌮 🦖

Issue #2483, opened 2021-07-07
Enhancement: calcite-input - update demo html to match Figma designs


ManahariDahal:
### Description
This issue aims to achieve parity between the calcite-input Figma design, its demo HTML page, and any usage snippets we provide.

### Acceptance Criteria 
- Variants illustrated in the Figma designs match the variants illustrated on the demo HTML page.

### Relevant Info 
- Figma: calcite-input needs to match the design in Figma.

#### Which Component
calcite-input

julio8a:
verified on demo pages

PR #2871, opened 2021-08-20
Manahari/2483 calcite input update

ManahariDahal:
**Related Issue:** #2483 

## Summary
- updated `calcite-input` demo page merging all other `input` components into one


julio8a:
These are not consistent across scales:
image

The calcite-input-message here should reflect medium and large:
image

Same for the error states label scales for M and L

Nice work on this! very extensive!

Looks like the spinners  and left/right arrow buttons are broken when it comes to high alignment with scales. Will create an issue to fix on the component.

julio8a:
Looks like those are safari specific bugs

julio8a:
The read-only need to match (from my previous comment)

bstifle (commented):
Needs to be yellow, warning:
image

Not consistent, please use same formatting that M and L are using for small:
image

Left-Missaligned header and button (probably need to mess with the grid on the inputs of this section):
image
image

Issue #2515, opened 2021-07-12
Bug: Combobox input's focus style


ptmdmusique:
### Actual Behavior

After the dropdown is opened, pressing Esc will make the box loses its focus style. We can still type inside the input and I _think_ it's still focused, but it loses its focused style for the border

### Expected Behavior

The input should not lose the focus style as it's still focused

### Reproduction Steps or Sample

https://esri.github.io/calcite-components/?path=/story/components-controls-combobox--simple

1. Focus on the combobox through click or tab -> dropdown should appear
2. Press Escape -> dropdown disappears and focus style is lost but we can still type in it

cc @paulcpederson

ptmdmusique:
verified!

PR #2875, opened 2021-08-20
fix(combobox): Set focus style correctly

driskull:
**Related Issue:** #2515

## Summary

fix(combobox): Set focus style correctly. #2515


jcfranco (approved):
💥

Issue #2853, opened 2021-08-18
Bug: `calcite-color-picker` value attribute stays null when the format property is "rgb"


ethanbdev:
### Actual Behavior

The value of the color picker element does not change from null when the format is "rgb"
start from null -> change the color -> color updates visually but value attribute is null

### Expected Behavior

start from null -> change the color -> element value attribute should match the visual color

### Reproduction Steps or Sample

https://codepen.io/eborgen/pen/LYRwjEm?editors=1001

### Relevant Info
It 🚫 some workflows for charts in the map viewer that will need to be fixed for 9.3 since it is a regression from 9.2.

Looks similar to https://github.com/Esri/calcite-components/issues/1468, could be similar fix

_Version_: `@esri/calcite-components@beta.62`

- [x] CDN
- [x] NPM package

PR #2886, opened 2021-08-24
fix(color-picker): fix value for empty color-picker with different initial format

jcfranco:
**Related Issue:** #2853 

## Summary

This fixes a bug where the internal mode was not set properly when a user set an initial format, which would prevent the `value` prop from being updated.


driskull (approved):
👍

Issue #2807, opened 2021-08-13
Docs: Consistent slot documentation


samMatenaer:
### Description 
All the available slots in the Panel component are documented in a Slots list in the [documentation](https://esri.github.io/calcite-components/?path=/docs/components-panel--basic).

However, for components like Card [don't have a similar list](https://esri.github.io/calcite-components/?path=/docs/components-card--simple), making it difficult to find available slots

List of components with slots documentation:
- [Alert](https://esri.github.io/calcite-components/?path=/docs/components-alert--title-message-link)
- [Flow](https://esri.github.io/calcite-components/?path=/docs/components-flow--basic)
- [Shell](https://esri.github.io/calcite-components/?path=/docs/components-shell--basic)
- [Tip](https://esri.github.io/calcite-components/?path=/docs/components-tips-tip--basic)
- [Popover](https://esri.github.io/calcite-components/?path=/docs/components-popover--simple)
- [Notice](https://esri.github.io/calcite-components/?path=/docs/components-notice--simple)

List of components without (complete) slots documentation
- [Card](https://esri.github.io/calcite-components/?path=/docs/components-card--simple)
- [Modal](https://esri.github.io/calcite-components/?path=/docs/components-modal--simple)
- [List Item](https://esri.github.io/calcite-components/?path=/docs/components-list--simple)
- [Pick List Item](https://esri.github.io/calcite-components/?path=/docs/components-pick-list--basic)
- [Value List Item](https://esri.github.io/calcite-components/?path=/docs/components-value-list--basic)
- [Tree](https://esri.github.io/calcite-components/?path=/docs/components-tree--simple)
- [Tabs](https://esri.github.io/calcite-components/?path=/docs/components-tabs--simple)

benelan:
List Item, Pick List Item, and Value List Item all have the proper jsdoc slot tags in the code. The issue with those components is that they are not documented at all in storybook under their parents. I created a different issue to fix that: #2887. The rest of the missing slots are added in #2888.

jcfranco:
@benelan Aside from `calcite-tree` and `calcite-tree-item`, everything looks good!

* `calcite-tree` – has a default slot, but doc has `children` slot
* `calcite-tree-item` – has default and `children` slot, but doc has none

benelan:
Thanks @jcfranco, should I be documenting default slots by the way? Mat was wondering that too cc @driskull

jcfranco:
[...]

PR #2888, opened 2021-08-24
docs(card, chip, link modal, tree, tabs): add missing slot doc

benelan:
**Related Issue:** #2807 

## Summary
Added slot documentation for the components where it was missing.


src/components/calcite-link/readme.md
```diff
@@ -32,6 +32,12 @@ You can programmatically focus a `calcite-link` with the `setFocus()` method:
 
 Type: `Promise<void>`
 
+## Slots
+
+| Slot                                | Description |
+| ----------------------------------- | ----------- |
+| `"default text slot for link text"` |             |
```
driskull, line 39: @benelan this one needs a name or dash I think.

src/components/calcite-modal/calcite-modal.tsx
```diff
@@ -36,6 +36,14 @@ const getFocusableElements = (el: HTMLElement | ShadowRoot): HTMLElement[] => {
   return queryShadowRoot(el, isHidden, isFocusableExtended);
 };
 
+/**
+ * @slot header - a slot for adding a modal header
```
driskull, line 40: @jcfranco Should we always doc the default slot?

driskull (changes requested):
Some slot feedback

src/components/calcite-link/readme.md
benelan: This one I didn't add myself, I just moved it. It was just in the wrong location so the slot wasn't being added to the generated html. It sounds like it is a default slot, should it be named "default"?

driskull: Gotcha. yeah it was already incorrect. Can we fix it since it moved? I think for default slots we usually just add a dash.

Like this:

```
 * @slot - A slot for adding `calcite-action`s that will appear at the top of the action bar.
 ```

driskull (approved):
Approved after slot fix. We should discuss if we always should document default slot. My opinion is that we should document it and what a user should slot there.

src/components/calcite-link/readme.md
benelan: > Approved after slot fix. We should discuss if we always should document default slot. My opinion is that we should document it and what a user should slot there.

Got it thanks. I'll take a yay/nay count for documenting default slots in our next team meeting.

Issue #2113, opened 2021-05-07
Enhancement: Allow icon size and position customizations for Tiles


mitc7862:
### Description
In order to launch the item details page/REST endpoint of a layer from within Map Viewer, I need the ability to add a launch indicator at the end of the tile

Existing component design:
![image](https://user-images.githubusercontent.com/12564905/117382911-60f51780-ae94-11eb-9c14-490492904306.png)

Proposed variant:
![image](https://user-images.githubusercontent.com/12564905/117382835-3f942b80-ae94-11eb-8ea1-f94db3e9f4ba.png)

Two possible resolutions could be:
- Add a slot at the end of the tile where we can add an indicator
- Allow resizing or repositioning of the existing tile icon (currently it always shows with size L)

### Acceptance Criteria 
- Ability to add an indicator (icon) at the end of a tile
- Allow an icon to be added inline within the description body

#### Which Component
Tile

#### Example Use Case
![image](https://user-images.githubusercontent.com/12564905/117382808-2c815b80-ae94-11eb-8c35-5623d22f9bd8.png)

bstifle:
Cool! propose calling this variant `icon-end` and the original layout 'icon-start`

jcfranco:
I find `icon-start`/`icon-end` a bit misleading, since `start` is more of on top. Would there be cases where you'd need both the top and end icons?

jcfranco:
@bstifle @mitc7862 Could you confirm whether ☝🏼  is possible with the proposed changes?

bstifle:
> I find `icon-start`/`icon-end` a bit misleading, since `start` is more of on top. Would there be cases where you'd need both the top and end icons?

No just one or the other. 

> Could you confirm whether ☝🏼 is possible with the proposed changes?

Just icon-top or icon-end, not at the same time :)

jcfranco:
Moving to next sprint. Draft PR should be up today.

jcfranco:
> Draft PR should be up today.

![starting-starting-now](https://user-images.githubusercontent.com/197440/129816084-b8216af8-c42f-43bc-b764-2b311b3ed0a2.gif)

jcfranco:
Actually, going to put this on hold until we decide on the rules for when to use slots vs props for icons. cc @asangma @macandcheese

asangma:
Per our recent sync, the proposal is to add a slot that can be used for that icon.
@mitc7862 This should enable your proposed design.

jcfranco:
[...]

PR #2890, opened 2021-08-24
feat(tile): add content-start/content-end slots

jcfranco:
**Related Issue:** #2113 

## Summary

This updates `calcite-tile` to start following the standardization of slots (content) per our [prop vs slot discussion](https://github.com/Esri/calcite-components/wiki/Property-vs-Slot) to support slotting icons.

Screen Shot 2021-08-24 at 9 37 28 AM

@asangma Can you review the structure & spacing?

### Questions

* Story or demo page update for slots? Both? 
* Do we need to update the doc on guidance on using the slots? E.g., The `icon` prop should not be used along with a slotted icon (per https://github.com/Esri/calcite-components/issues/2113#issuecomment-866047377).

### DIFF
--- a/src/components/calcite-tile/calcite-tile.e2e.ts
+++ b/src/components/calcite-tile/calcite-tile.e2e.ts
@@ -1,6 +1,7 @@
 import { newE2EPage } from "@stencil/core/testing";
-import { accessible, defaults, hidden, reflects, renders } from "../../tests/commonTests";
+import { accessible, defaults, hidden, reflects, renders, slots } from "../../tests/commonTests";
 import { html } from "../../tests/utils";
+import { SLOTS } from "./resources";
 
 describe("calcite-tile", () => {
   it("renders", async () => renders("calcite-tile", { display: "inline-block" }));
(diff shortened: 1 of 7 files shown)


src/components/calcite-tile/calcite-tile.e2e.ts
```diff
@@ -15,6 +16,8 @@ describe("calcite-tile", () => {
       { propertyName: "hidden", defaultValue: false }
     ]));
 
+  it("has slots", async () => slots("calcite-tile", SLOTS));
```
driskull, line 19: nice

src/components/calcite-tile/calcite-tile.tsx
```diff
@@ -1,5 +1,10 @@
 import { Component, Fragment, h, Prop, VNode } from "@stencil/core";
+import { SLOTS } from "./resources";
 
+/**
+ * @slot content-start - A slot for adding non-actionable elements before the tile content.
```
driskull, line 5: Should we doc default slot information?

driskull (approved):
💯

src/components/calcite-tile/calcite-tile.tsx
jcfranco: For the content, tile uses the `heading`/`description` props, so there's no default slot at the moment. Or were you suggesting a default slot that has the heading and description as its default content?

driskull: Oh interesting. Is it odd if there isn't a default slot but there are named slots?

driskull: I think its probably fine. Nvm then!

eriklharper (approved):

PR #2891, opened 2021-08-24
docs(avatar, date-picker): add closing tags

benelan:
**Related Issue:** ##2859

## Summary
Custom Elements should not be self closing, so I added closing tags to the usage examples that are.


driskull (approved):
👍 nice

Issue #2440, opened 2021-07-06
Bug: Tree with nested tree elements fires duplicate events


jcfranco:
### Actual Behavior

A tree with nested tree children (as items) will emit duplicate events (one per nested tree in the hierarchy).

### Expected Behavior

A single tree select event should fire at the root of the tree.

### Reproduction Steps or Sample

With the following: https://codepen.io/jcfranco/pen/eYWZYwL?editors=1001

1. Click parents until you reach the E item
2. Click on the E item
3. Notice how 3 `calciteTreeSelect` events are emitted

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [x] CDN
- [ ] NPM package

Related comments:
* https://github.com/Esri/calcite-components/issues/2290#issuecomment-867235506
* https://github.com/Esri/calcite-components/pull/2438#issuecomment-874922008

jcfranco:
Verified w/ repro case.

PR #2893, opened 2021-08-24
fix(tree): Tree with nested tree elements should fire only one event

driskull:
**Related Issue:** #2440

## Summary

fix(tree): Tree with nested tree elements should fire only one event. #2440

### DIFF
--- a/src/components/calcite-date-picker/calcite-date-picker.e2e.ts
+++ b/src/components/calcite-date-picker/calcite-date-picker.e2e.ts
@@ -121,7 +121,7 @@ describe("calcite-date-picker", () => {
   });
 
   describe("when the locale is set to Slovak calendar", () => {
-    it("should start the week on Monday", async () => {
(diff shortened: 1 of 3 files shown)


src/components/calcite-date-picker/calcite-date-picker.e2e.ts
```diff
@@ -121,7 +121,7 @@ describe("calcite-date-picker", () => {
   });
 
   describe("when the locale is set to Slovak calendar", () => {
-    it("should start the week on Monday", async () => {
+    it.skip("should start the week on Monday", async () => {
```
jcfranco, line 124: I submitted https://github.com/Esri/calcite-components/pull/2899 to keep the test-skip separate. The reason is if we had to revert or do something with the squashed commit for the fix, it would affect an unrelated test.

jcfranco (approved):
🎉 🌮

Issue #2082, opened 2021-04-29
Enhancement: Selection visuals for Tree


bstifle:
Remove left-hand bar single selection active state style:
![image](https://user-images.githubusercontent.com/14083530/116617741-d183bd80-a903-11eb-82ac-d0fbaa14df23.png)

- Replace with circle (radio) visual
- Have ability to have multi select with checkmarks
- have ability to style multi select using `calcite-checkbox` including all available states such as idle, hover, selected, and indeterminate. Checkbox should give the ability to select an entire parent/group and its children
![image](https://user-images.githubusercontent.com/14083530/116618761-19571480-a905-11eb-8894-a9f52817488d.png)

bstifle:
![Uploading Tree Specs.png…]()

bstifle:
![Tree Specs](https://user-images.githubusercontent.com/14083530/116725789-0b0e0480-a9a8-11eb-83ca-5b2e06e5d479.png)

macandcheese:
Comps look good, great alignment

🔲🔲
🔲🔲


It seems like the "checkbox" design functions differently than "check icon" - are parent-level items not selectable in that mode?

bstifle:
to my understanding no, which is kind of a neat differentiator imo

macandcheese:
Ok, it might be worth looking how these three layouts map to all 5 of the selection-modes, and how it interacts with the show-input prop (if they need to both still remain)?

If we can make that decision for them based on mode, we'll have less of a tree-mode-potpourri across our apps. Just props for theme, lines, scale is nice.

evanmosby:
Looking forward to these options with ArcGIS Monitor! Especially the center option (checks w/o box) for facet filter multi-select!

caripizza:
Tree scales were installed in pr #2803.

Next I'll fine-tune the selection visuals with Bryan in Figma. We'll need to support all 5 of tree's selection-modes, which were implemented according to the specs here:
- https://www.w3.org/TR/wai-aria-practices-1.1/#:~:text=Selection%20in%20multi-select%20trees
- https://www.w3.org/TR/wai-aria-practices-1.1/#:~:text=Alternative%20selection%20model%20--%20Moving%20focus%20without%20holding%20the%20Shift%20or%20Control%20modifier%20unselects

(Descriptions and screenshot examples for our current tree behavior in each selection-mode can be found [in the markdown table here](https://github.com/Esri/calcite-components/pull/2803#issue-711740128).)

[...]

PR #2906, opened 2021-08-25
feat(tree-item): updates selection visuals to match Figma

caripizza:
**Related Issue:** #2082 (this is a follow on to deprecating the input-enabled prop pr)

## Summary
Updates selected tree-item styles to match Figma, for each tree selection-mode:
- checkmark icon: multi/multi-children
- bullet point icon: single/children
- checkbox input: ancestors

Latest designs from issue:
no lines
lines


src/components/calcite-tree-item/calcite-tree-item.tsx
```diff
@@ -149,6 +149,30 @@ export class CalciteTreeItem {
           <slot />
         </label>
       ) : null;
+    const bulletIcon =
+      this.selectionMode === TreeSelectionMode.Single ||
```
jcfranco, line 153: Since `selectionMode` is referenced a few times, can you store it in a local var?

jcfranco: I think this can be merged too:

```tsx
const icon = selectionMode === TreeSelectionMode.Children ? 
  "bullet-point" : selectionMode === TreeSelectionMode.Multi || selectionMode === treeSelectionMode.MultiChildren ? 
  "check" : null;
//...
const bulletOrCheckIcon = icon ?
<calcite-icon  
  class={{
    [CSS.bulletPointIcon]: icon === "bullet-point",
    [CSS.checkmarkIcon]: icon === "check",
    [CSS_UTILITY.rtl]: rtl
  }}
  icon={icon}
  scale="s"
/>
: null
```

src/components/calcite-tree-item/resources.ts
```diff
@@ -3,7 +3,9 @@ export const CSS = {
   checkbox: "calcite-tree-checkbox",
   chevron: "calcite-tree-chevron",
   nodeContainer: "calcite-tree-node",
-  childrenContainer: "calcite-tree-children"
+  childrenContainer: "calcite-tree-children",
```
jcfranco, line 6: These don't need the `calcite-tree` prefix since it's all encapsulated.

jcfranco (approved):
Had a few comments, but code changes LGTM! I'll let someone else chime on the styles.

macandcheese (approved):
Apologies for not looking at the original comps earlier... 

The choice to align the vertical lines with the right extent of the chevron vs. the center seems a bit odd, as a result these end up looking misaligned when multiple children are open...

However, the PR implements as shown in the original comps, so I think we should approve this and follow up with any small design tweaks in a new issue. Nice work on the stylesheet clean up here!

caripizza:
Closing this pr because Screener is not showing the changes it should be. Moving forward in pr #3001 instead.

Issue #2874, opened 2021-08-20
Bug: Safari Input buttons 


julio8a:
### Actual Behavior

Safari: buttons don't align with the input.

<img width="455" alt="image" src="https://user-images.githubusercontent.com/6144462/130290701-de46dea3-f42d-4e56-865d-d71b0c9f7fd6.png">

### Expected Behavior
On Chrome:
<img width="1331" alt="image" src="https://user-images.githubusercontent.com/6144462/130290724-3dfa3b17-cbdf-466a-95a7-83bb47cbcaed.png">

### Reproduction Steps or Sample
Use safari, v14.1.2
https://codepen.io/julio8a/pen/PomMrxG

### Relevant Info
Safari Mac 4.1.2

beta.61

- [X] CDN
- [X] NPM package

julio8a:
verified on demo pages

PR #2929, opened 2021-08-27
fix(input): ensure clearable button has focus outline

caripizza:
**Related Issue:** #2874 (follow-on to pr #2904)

## Summary
Prevents added margins to calcite-input's clearable button, and adds `focus-inset` so it matches the input's focus style.


jcfranco (approved):
👍🏼

julio8a (approved):

Issue #2926, opened 2021-08-27
Bug: Inline editable `disabled` has a hover state


julio8a:
### Actual Behavior

The disabled inline editable should not have a hover state. Moreover, I think the idle text should be a lighter gray to appear disabled.

![image](https://user-images.githubusercontent.com/6144462/131074593-ab063317-69aa-4e8d-8ab0-f2de3fe1a46a.png)

### Expected Behavior

No hover state and lighter text to appear disabled. 

### Reproduction Steps or Sample

Hover over the disabled inline-editable here (beta.62):
https://codepen.io/julio8a/pen/MWowVVK

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [X] CDN
- [X] NPM package

julio8a:
verified on master local build demo pages

PR #2931, opened 2021-08-27
fix(inline-editable): prevent button :hover style when disabled, center in container

caripizza:
**Related Issue:** #2926

## Summary
Ensures disabled inline-editable doesn't change color on :hover, centers button within its container, and adds back missing script to demo html page.


jcfranco (approved):
💥

PR #2932, opened 2021-08-27
feat(block): add margin between heading and summary

benelan:
**Related Issue:** ##2494

## Summary
Added 2px margin between Block's heading and summary. The issue also mentioned to add the margin to List Item but it is already there: https://github.com/Esri/calcite-components/blob/master/src/components/calcite-list-item/calcite-list-item.scss#L71


benelan:
@eriklharper I was expecting the screener check to fail since I added margin. Do you have any ideas about why it passed here?

julio8a:
I've noticed this before too. When there's a spacing or font change and Screener doesn't see it.

asangma (approved):
Noice.

Issue #2918, opened 2021-08-26
Bug: calcite-input step:0.1 issues with arrows


AdelheidF:
(1)
If step is 0.1 the arrow-up only goes up to 1. Once it reaches 1 the arrows don't do anything anymore. Same when going down to 1. Same with 0 and 2. I can type in a value fine, it's just an issue with the arrows. 

(2)
Using the arrow up/down keys on my keyboard when the input has the focus just moves the cursor around inside the input. No value change happening. Apparently the value should go up/down according to its step setting.

![image](https://user-images.githubusercontent.com/44481292/131050083-318ee929-c38a-4393-9182-089239e1c707.png)

I need this 0.1 stepper for the new effects component which goes out with 9.3 (in 3 weeks)

@eriklharper @julio8a 

### Reproduction Steps or Sample

1.

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

AdelheidF:
Nice, fixed with beta.63.

I'll close issues that I opened after verifying...

PR #2939, opened 2021-08-27
fix(calcite-input): number values properly nudge when step is set to a decimal or "any"

eriklharper:
**Related Issue:** #2918

## Summary

This fixes an issue where nudging number values when step is set to a decimal was setting the value to 1 and then further nudging wouldn't have any effect.

### DIFF
--- a/src/components/calcite-input/calcite-input.tsx
+++ b/src/components/calcite-input/calcite-input.tsx
@@ -502,17 +502,13 @@ export class CalciteInput {
     if (this.type !== "number") {
       return;
     }
+    const value = this.value;
     const decimals = this.step?.toString().split(".")[1]?.length || 0;
     const inputMax = this.maxString ? parseFloat(this.maxString) : null;
     const inputMin = this.minString ? parseFloat(this.minString) : null;
     const inputStep = this.step === "any" ? 1 : Math.abs(this.step || 1);
-    const inputVal =
-      this.value && this.value !== ""
-        ? decimals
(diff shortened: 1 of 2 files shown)


src/components/calcite-input/calcite-input.tsx
```diff
@@ -502,19 +502,24 @@ export class CalciteInput {
...
       return;
     }
-    const decimals = this.value?.split(".")[1]?.length || 0;
+    const decimals = this.step?.toString().split(".")[1]?.length || 0;
     const inputMax = this.maxString ? parseFloat(this.maxString) : null;
     const inputMin = this.minString ? parseFloat(this.minString) : null;
     const inputStep = this.step === "any" ? 1 : Math.abs(this.step || 1);
-    let inputVal = this.value && this.value !== "" ? parseFloat(this.value) : 0;
+    const inputVal =
+      this.value && this.value !== ""
```
jcfranco, line 510: Nitpick: `this.value` is referenced a bit, so can you store it in a local variable?

jcfranco: This line can be simplified to: `const inputVal = this.value ? ...`

```js
// when "1" => ("1") && ("1" !== "" /* will never be true */);
// when null/undefined => (null/undefined) && (/* not reached */ null/undefined !== "");
// when "" => ("") && (/* not reached */ null !== "");
```

jcfranco (approved):
🌮 🎉

PR #2954, opened 2021-08-31
chore: add gh action to label prs

benelan:
## Summary
Added a GitHub Action to automatically label PRs based on the conventional commit title. 

Tested here: https://github.com/benelan/test-actions/pull/1


driskull (approved):
👏

.github/labeler.yml
```diff
@@ -0,0 +1,16 @@
+ version: 1
+ labels:
+  - label: "enhancement"
+    title: "^[fF][eE][aA][tT].*"
```
jcfranco, line 4: I was thinking if we needed to have the pattern strictly match the conventional commit type (`type: message` or `type(scope): message`), but the Semantic PR check should catch that.

jcfranco (approved):
Awesome! :sparkles:🤖 🏷️ :sparkles:

.github/labeler.yml
benelan: Yeah it seems like overkill to check if the title strictly matches the conventional commit since we have the Semantic PR check like you said. As long as this action matches titles both with/without scope (which it does) it should be enough in my opinion.

Issue #1926, opened 2021-04-07
Enhancement: Add border for Action Bars


TheBlueDog:
### Description

Currently when an action bar is used, there is no separation between it and the rest of the content on the page without adding customization.

Example:

![Action-Bar Default](https://user-images.githubusercontent.com/5419306/113941125-91ca2a00-97cc-11eb-9a96-b78814c7d30a.png)

![Screen Shot 2021-04-07 at 4 43 19 PM](https://user-images.githubusercontent.com/5419306/113932572-c4225a00-97c1-11eb-8f54-f8c0e39af2fe.png)

Since there is configuration for `start` and `end` positions, the appropriate border can be added as part of the component.

Example:

![Action-Bar Borders](https://user-images.githubusercontent.com/5419306/113941214-b0302580-97cc-11eb-97d5-512d71c717d9.png)

![Screen Shot 2021-04-07 at 4 55 21 PM](https://user-images.githubusercontent.com/5419306/113933426-8ffb6900-97c2-11eb-9c60-96c73cad8076.png)

### Acceptance Criteria 

- [ ] If an action-bar is used in the `start` position, add `border-right` with `ui-border-1`.
- [ ] If an action-bar is used in the `end` position, add `border-left` with `ui-border-1`.

#### Which Component
Calcite Action Bar

#### Example Use Case
Map View Beta and Field Maps are currently using the actionbars with added local customizations to apply a border and border-color.

bstifle:
i think border-1 is too much. suggest using border-2 #dfdfdf to match whats already used in action bar and action pad

justinhough:
@TheBlueDog Do you have a working example of this that I can take a look at? 

I don't have access to the "Map View Beta and Field Maps" and not sure if this would be a local project fix or something in relation to action bar within calcite. 

Let me know. Thanks.

cc @julio8a @caripizza

justinhough:
@TheBlueDog I added a demo for action bar to show when position is used and used the color recommendation from @bstifle

justinhough:
cc @julio8a

TheBlueDog:
@justinhough You might want to check with @asangma to see what the relationship of this is when used inside vs outside the app-shell.

TheBlueDog:
@julio8a This Figma library needs to be update to use a boarder left/right

julio8a:
[...]

PR #2960, opened 2021-09-01
refactor(action bar): border added when position start or end used

justinhough:
**Related Issue:** #1926

## Summary

Updated border on action bar to show borders when start (border-right) or end (border-left) used. Update demo page to showcase this feature update.


TheBlueDog (changes requested):
Per @asangma, this change is not needed in the components

justinhough:
Closing PR since the change is not needed

Issue #2597, opened 2021-07-21
Bug: Accordion chevron direction doesn't match Block


asangma:
### Actual Behavior
Note: this only applies when Accordion is `icon-position`: `end` and `icon-type`: `chevron` or `caret`.

**Block**
In Block the chevron indicates the direction of the resultant action. Down when collapsed...
![image](https://user-images.githubusercontent.com/12503298/126569967-f39827bc-30ea-4f15-a95c-53043fd707f6.png)

...up when opened.
![image](https://user-images.githubusercontent.com/12503298/126570020-7176aae4-81b0-4bb0-bd2b-9f2b796d9cf7.png)

**Accordion**
Accordion uses the chevron to indicated state. Up when collapsed...
![image](https://user-images.githubusercontent.com/12503298/126570081-9f13edbb-fd30-4640-825f-76afbc0c0305.png)

...down when open.
![image](https://user-images.githubusercontent.com/12503298/126570117-de07de1c-d4a5-4ccb-be6d-b402305bdbd5.png)\

### Expected Behavior
Align with Block chevron direction: down when closed & up when open.

### Reproduction Steps or Sample

1. See Accordion demo and Block demo.

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

cc @bstifle

bstifle:
yes! this is the way it was intended. thanks alan

asangma:
Note, this reverts the changes discussed [here](https://github.com/Esri/calcite-components/issues/289#issuecomment-630419597).

But we have an established pattern in Block which is being used a fair amount.

bstifle:
Yeah, the chevron direction paradigm should be different depending on start/end placement.

Start closed: pointing right
Start open: pointing down

End closed: pointing down
End open: pointing up

asangma:
Agreed.
This is currently what we use in Block (right/end) and BlockSection (left/start).

benelan:
This has been installed, @asangma can you please verify?

asangma:
@anveshmekala Sorry. Looks like we missed the `icon-position--start` case. With the update, the open icon should be pointing down.
![Screen Shot 2021-09-03 at 4 34 31 PM](https://user-images.githubusercontent.com/12503298/132074462-bdd31089-0bb5-4bd3-843f-843eb71ce8de.png)

The animation should also go the other way. Right now, the icon spins the wrong way.

anveshmekala:
[...]

PR #2963, opened 2021-09-01
fix(accordion): match accordion icon directions with block

anveshmekala:
Related Issue: [#2597](https://github.com/Esri/calcite-components/issues/2597)

## Summary
This fixes an issue in `calcite-accordion-item` where chevron and caret directions pointing opposite of `calcite-block` .

### DIFF
--- a/src/components/calcite-accordion-item/calcite-accordion-item.tsx
+++ b/src/components/calcite-accordion-item/calcite-accordion-item.tsx
@@ -126,7 +126,6 @@ export class CalciteAccordionItem {
                   : this.active
                   ? "minus"
                   : "plus"
-                //this.getIcon()
               }
               scale="s"
             />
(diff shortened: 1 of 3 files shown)


julio8a (approved):
Visually, looks good. Thank you!

src/components/calcite-accordion-item/calcite-accordion-item.tsx
```diff
@@ -239,4 +240,12 @@ export class CalciteAccordionItem {
       this.el
     );
   }
+
+  // private getIcon(): string {
```
julio8a, line 244: Do we need this commented-out code?

anveshmekala: Nope. Will remove the commented out part.

asangma (approved):
Looks good!

asangma:
@anveshmekala Not required, but if you have a moment, it'd be nice to add a couple examples in the demo that use the caret icon. Right now the demo page only demonstrates the chevron and +/- icons.

anveshmekala:
@asangma Added an example for `icon-position` prop with caret. Please review it.

asangma:
Awesome! Looks solid. Thanks @anveshmekala!

asangma:
@anveshmekala Sorry. Looks like we missed the `icon-position--start` case. With the update, the open icon should be pointing down.
Screen Shot 2021-09-03 at 4 34 31 PM

The animation should also go the other way. Right now, the icon spins the wrong way.

PR #2967, opened 2021-09-02
fix(block-section): style update to align better with label and switch combo.

asangma:
**Related Issue:** #2577

## Summary
* cleans up styles that were causing misalignments with label+switch
* adds example in a demo

### DIFF
--- a/src/components/calcite-block-section/calcite-block-section.scss
+++ b/src/components/calcite-block-section/calcite-block-section.scss
@@ -22,7 +22,7 @@
     border-0
     text-color-2
     font-sans
-    font-normal
+    font-medium
     w-full;
 }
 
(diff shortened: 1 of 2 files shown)


julio8a (approved):
Visually, looks good on demo pages! Thx for the fix @asangma

julio8a:
thx for the font-weight update too, it matches Figma now.

asangma:
Thanks @julio8a.

caripizza:
@asangma looks good! Is there ever a case where block-section would need to display with status + icon? (Just curious if that would play into this alignment issue.)

asangma:
@caripizza Ah! Good catch. I'll test that.

caripizza (approved):

asangma:
Thanks @caripizza !

Issue #2887, opened 2021-08-24
Docs: Add child components that are not included in the parent storybook page


benelan:
### Description 
Most of the components with direct children have those children documented on the same storybook page, e.g. [Tabs](https://esri.github.io/calcite-components/?path=/docs/components-tabs--simple), [Flow](https://esri.github.io/calcite-components/?path=/docs/components-flow--basic), [Shell](https://esri.github.io/calcite-components/?path=/docs/components-shell--basic). etc. 

The components listed below do not have their child components documented in storybook and should have them added.

### Which Component
[List](https://esri.github.io/calcite-components/?path=/docs/components-list--simple)
[Pick List](https://esri.github.io/calcite-components/?path=/docs/components-pick-list--basic)
[Value List](https://esri.github.io/calcite-components/?path=/docs/components-value-list--basic)
[Radio Button](https://esri.github.io/calcite-components/?path=/docs/components-controls-radio-radio-button--light-theme)

benelan:
Installed, can one of you please verify @jcfranco @julio8a?

jcfranco:
I'll verify.

benelan:
@jcfranco can you verify this one?

jcfranco:
Reopening to verify.

jcfranco:
Verified via internal storybook! 🎉

PR #2970, opened 2021-09-02
docs: add missing storybook readmes

benelan:
**Related Issue:** #2887

## Summary
Some of the child components did not have their readmes on storybook, such as `calcite-value-list-item`. I added the missing readmes under the parent components, similar to how [calcite-stepper](https://esri.github.io/calcite-components/?path=/docs/components-stepper--simple) is [implemented](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-stepper/calcite-stepper.stories.ts#L12).


driskull (approved):
👍 🎉 Nice!

Issue #2851, opened 2021-08-18
Refactor: Consistent use of resource files for slots


benelan:
### Description
For some components slots are hardcoded into the .tsx file rather than being imported from a resource file. Here are the components with hardcoded slots (I searched ``<slot name="`` to find these):

```
calcite-modal
calcite-chip
shared-list-render
calcite-radio-group-actions
calcite-shell
calcite-tabs
calcite-tree-item
```

Examples: 
https://github.com/Esri/calcite-components/blob/master/src/components/calcite-modal/calcite-modal.tsx#L171
https://github.com/Esri/calcite-components/blob/master/src/components/calcite-shell/calcite-shell.tsx#L66

Additionally, some components have ``resources.ts`` files and other haves ``{component-name}.resources.ts`` files. If there is not a reason for the inconsistency should we change the former to the latter? Any thoughts?

Examples: 
https://github.com/Esri/calcite-components/blob/master/src/components/calcite-shell/resources.ts
https://github.com/Esri/calcite-components/blob/master/src/components/calcite-alert/calcite-alert.resources.ts

### Related Issues
- #2807

jcfranco:
I found a straggler in dropdown: https://github.com/Esri/calcite-components/blob/0950e3d62c019f7396aa8716bc91acdae06cde84/src/components/calcite-dropdown/calcite-dropdown.tsx#L149

Aside from that, looks great!

benelan:
@jcfranco, @julio8a, ready for verification!

jcfranco:
Verified locally!

PR #2973, opened 2021-09-02
refactor(dropdown): use resource file for slot

benelan:
**Related Issue:** #2851

## Summary
Missed one hardcoded slot from calcite-dropdown.


jcfranco (approved):
💥

PR #2984, opened 2021-09-04
refactor: add util to create observers consistently

jcfranco:
**Related Issue:** N/A

## Summary

Introduces a utility to consistently create observers (mutation, intersection, resize) for browser contexts. This will help with prerendering.


src/components/calcite-tip-manager/calcite-tip-manager.tsx
```diff
@@ -105,11 +106,11 @@ export class CalciteTipManager {
 
   connectedCallback(): void {
     this.setUpTips();
-    this.observer.observe(this.el, { childList: true, subtree: true });
+    this.mutationObserver?.observe(this.el, { childList: true, subtree: true });
```
driskull, line 109: If we want to enforce using `mutationObserver` and `resizeObserver` variables, we can create interfaces for classes to implement as well.

driskull (approved):
👍 Nice cleanup 🎉 🌮

Issue #2947, opened 2021-08-30
Bug: Slider handle not centered vertically on Safari


samMatenaer:
The center of the handle on the slider component does not fall on the line in Safari

### Actual Behavior
Safari:
<img width="275" alt="image" src="https://user-images.githubusercontent.com/50883428/131365572-b0916e46-8b9f-4ab6-97de-0035f7f95439.png">
<img width="273" alt="image" src="https://user-images.githubusercontent.com/50883428/131366118-5ebf823e-5de4-4959-a086-f8e2cc64b2f4.png">

### Expected Behavior
Firefox
<img width="286" alt="image" src="https://user-images.githubusercontent.com/50883428/131365607-8c1c58dd-3b6c-4538-acaa-5cf0a4ff89bf.png">

<img width="279" alt="image" src="https://user-images.githubusercontent.com/50883428/131366089-edc2466e-1ed9-45c1-b689-63f735151837.png">

### Reproduction Steps or Sample
https://esri.github.io/calcite-components/?path=/story/components-controls-slider--single-value

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

julio8a:
@caripizza, can you fix this one as part of the slider work you are working on?

julio8a:
verified on master branch local build demo page.

PR #2987, opened 2021-09-07
fix(slider): prevent added handle margins in Safari

caripizza:
**Related Issue:** #2947

## Summary
Adds `margin: 0` to calcite-slider's `.thumb` button so the handle stays aligned with the track.


macandcheese (approved):
Nice cleanup 👍

PR #2990, opened 2021-09-07
fix: default boolean properties to be false instead of undefined.

driskull:
**Related Issue:** None

## Summary

fix: default boolean properties to be false instead of undefined.


jcfranco (approved):
🎉 🌮 🦖

Issue #2880, opened 2021-08-23
Bug: Uncheck and unfocused calcite-check is smaller than checked and unfocused


yannikmesserli:
### Actual Behavior

The calcite-check, when checked and unfocused is 16x16px:
![Screenshot 2021-08-23 at 19 04 04](https://user-images.githubusercontent.com/1910449/130488162-7c03eb41-7a68-4795-9bec-ab7cdd6a46a6.png)
When it is focused, it is shrunk to allow space for the usability focused visuals:
![Screenshot 2021-08-23 at 19 04 48](https://user-images.githubusercontent.com/1910449/130488326-c124ac4c-06f7-435a-9998-a60375eb4b70.png)
Which is consistant when the check box is unchecked:
![Screenshot 2021-08-23 at 19 04 23](https://user-images.githubusercontent.com/1910449/130488385-3ae072e6-30ce-407f-ac42-aaf29845782d.png)
However, when it is unfocused, the checkbox stays shrunk, which is not aligned with checked checkbox:
![Screenshot 2021-08-23 at 19 04 43](https://user-images.githubusercontent.com/1910449/130488453-3e50e0c0-f20e-4a3a-ab7b-0e19ddf1d81a.png)

This is particularly visible when having a lit of checkboxes:
![Screenshot 2021-08-23 at 19 10 25](https://user-images.githubusercontent.com/1910449/130488565-1e270ba1-de22-472f-b078-79f69e2604f6.png)

### Expected Behavior

The unchecked and unfocused checkbox takes up 16px by 16px, border included.

### Reproduction Steps or Sample

Using the storybook  as an example.

Additional repro case: https://codepen.io/jcfranco/pen/jOwraaZ?editors=1010

### Relevant Info

_Version_: `@esri/calcite-components@1.0.0-beta.62`

- [ ] CDN
- [x] NPM package

julio8a:
I took screenshots in the HTML demo pages and placed them side by side. 

Widths looks good:
![image](https://user-images.githubusercontent.com/6144462/134415796-0e993fa7-22de-4705-b156-0e0603c4b1b5.png)

However, the height of the checked non-focused checkbox is 1px smaller than the rest.
![image](https://user-images.githubusercontent.com/6144462/134415684-44defeb1-8149-4d99-8773-5a5f08701983.png)

anveshmekala:
@julio8a Unable to reproduce the issue. Can you provide the reproduction steps or a sample repo would be helpful too.

[...]

PR #2995, opened 2021-09-08
fix(checkbox): resize calcite-checkbox when unchecked & focusout

anveshmekala:
**Related Issue:** #2880 

## Summary
This fixes the `calcite-checbox` size shrink when it is unchecked & focusout and matches with size of checkbox when checked set to true

### DIFF
--- a/src/components/calcite-checkbox/calcite-checkbox.scss
+++ b/src/components/calcite-checkbox/calcite-checkbox.scss
@@ -50,6 +50,7 @@
 :host([indeterminate]) {
   .check-svg {
     @apply bg-brand;
+    box-shadow: inset 0 0 0 1px var(--calcite-ui-brand);
   }
 }
 :host {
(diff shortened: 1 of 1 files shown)


macandcheese (changes requested):
The sizing change here is good, but we need to make sure the outline is the same color as the background here:

Screen Shot 2021-09-10 at 4 09 56 PM

Applying `box-shadow: inset 0 0 0 1px var(--calcite-ui-brand)` for the`[checked]` and `[indeterminate]` states should work.

julio8a:
@Esri/calcite-components-reviewers should screener fail on this one?

anveshmekala:
@julio8a  It should have . We removed `shadow-outline-active` for the `[checked]`state which shrinks the border by 1px.

caripizza:
@anveshmekala can you re-merge in master to see if Screener catches the change and fails? 

If it doesn't, I'm wondering maybe if it's due to the fact that this pr branch was created before the Screener config update.

anveshmekala:
@caripizza sounds good.  Will re-megre and see if the screener catches the issues.

anveshmekala:
@caripizza  Screener test has passed and don't see any failure. Any thoughts?

caripizza (approved):
👍  @anveshmekala I'm not sure what's up with Screener, but we can debug that 1px diff issue in a subsequent pr.

Issue #2940, opened 2021-08-27
Bug: Color input does not display the selected color


benelan:
### Actual Behavior
In Chrome an input with type="color" does not consistently display the selected color. It appears gray no matter which color is selected. It does work sometimes but most of the time it does not. It consistently works in Firefox for me. @julio8a mentioned that the input is completely empty in Safari.

### Expected Behavior
The input is supposed to display the selected color by changing the color of the gray line.

### Reproduction Steps or Sample
https://codepen.io/benesri/pen/qBjOdbG

Chrome:
![image](https://user-images.githubusercontent.com/10986395/131197506-c9c112dd-f899-4c0d-baae-76883cdcb84d.png)

Firefox:
![image](https://user-images.githubusercontent.com/10986395/131197547-e87f8d06-9789-45eb-b3d1-596c2fb35b4f.png)

### Relevant Info
Chrome Version 92.0.4515.159
Windows 10

_Version_: `@esri/calcite-components/1.0.0-beta.62`

- [x] CDN
- [ ] NPM package

### Related Issue
- #1934

benelan:
Installed, @jcfranco or @julio8a can you verify using the sample in the issue post?

jcfranco:
Verified with repro case!

PR #2997, opened 2021-09-09
fix(input): change color input padding so the selected color appears

benelan:
**Related Issue:** #2940 

## Summary
The color would remain gray in chrome and safari instead of showing the selected color. Changing the padding resolved this issue. Here is a [sample](https://codepen.io/benesri/pen/qBjOdbG) showing how the change looks. You can remove the style to see the issue.


caripizza (approved):
Nice @benelan 👍 , can you merge in latest master too?

PR #2978, opened 2021-09-03
test: Screener story changes

caripizza:
**Related Issue:** n/a

## Summary
Using this pr branch to test Screener changes getting picked up (TESTING ONLY)

Notes:
- first commit: Screener recorded no visual changes even though I deleted a story (should have failed)
- second commit: Screener recorded no visual changes when I commented out styles in switch's stylesheet
- third commit: Screener recorded no visual changes when I changed the height of a default/medium switch

PR #2998, opened 2021-09-09
ci(screener): disables alwaysAcceptBaseBranch setting

caripizza:
**Related Issue:** n/a

## Summary
Disables the `alwaysAcceptBaseBranch` Screener setting (from pr #2843).

**Why?**
We don't want undesirable changes that inadvertently get merged into master to be assumed to be correct. These were leading to false positives on subsequent pr builds.

According to Screener, this is how the setting functions:

> Yes, the "alwaysAcceptBaseBranch" is making an assumption that everything in your base branch is correct, and will automatically accept all states - Loyal

For additional insight, see pr #2978 for screenshot examples with the diff behavior Screener failed to pick up with this setting enabled.

**What else did we learn?**
- removal of stories does not impact test results (ie., this won't fail Screener)
- Screener's default settings ignore small layout position/dimension differences. According to Loyal at Screener:

> This can be made stricter using the diffOptions setting in your screener.config.js:
>```
>diffOptions: {
>  minLayoutPosition: 1,
>  minLayoutDimension: 1
>}
>```
>The above sample options will set the minimum threshold to trigger a layout change on > 1px

### DIFF
--- a/screener.config.js
+++ b/screener.config.js
@@ -1,13 +1,13 @@
 module.exports = {
-  projectRepo: "Esri/calcite-components",
-  storybookConfigDir: ".storybook",
-  storybookStaticDir: "./__docs-temp__",
+  projectRepo: 'Esri/calcite-components',
+  storybookConfigDir: '.storybook',
+  storybookStaticDir: './__docs-temp__',
   apiKey: process.env.SCREENER_API_KEY,
-  resolution: "1024x768",
-  baseBranch: "master",
+  resolution: '1024x768',
+  baseBranch: 'master',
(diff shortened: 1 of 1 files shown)


driskull (approved):

eriklharper (approved):

jcfranco (approved):
Awesome! 

Can you add some context to the PR description and possibly remove the formatting changes from this commit?

caripizza:
@jcfranco sure, should I use the `--no-verify` flag in my commit so it doesn't auto-lint the file?

Issue #2652, opened 2021-07-27
Enhancement: Slider should support input and change events


jcfranco:
### Description

Similar to `HTMLInputElement`, an input and change events would allow users to distinguish between actively changing and the final value when the user drags any thumb.

### Acceptance Criteria 

The slider has the following events: 

* `calciteSliderInput` – fires on other interactions that update `value` and as the thumb is dragged (on move)
*  `calciteSliderChange` – fires on other interactions that update `value` and during a drag, it only fires on thumb release

#### Which Component

* `calcite-slider` 

#### Example Use Case

Any event handler listening on the existing `calciteSliderChange` needs to 

#### Additional info

Currently, the change event is being throttled as it fires during thumb drag. This could be removed and have the input event fire as the thumb is dragged.

gavinr:
If using "steps", it would be nice to have an event _only_ when the slider pops into one of the steps (`calciteSliderStep` event, or something like that)

anveshmekala:
@gavinr that would be great add-on if using steps. Would it be expected to just emit `calciteSliderStep` ignoring `calciteSliderChange` when the slider pops into one of the steps ?

jcfranco:
@gavinr Thanks for the suggestion. Since that would introduce a new event, can you create a separate issue for that?

gavinr:
@jcfranco sure - what name would you suggest for the event?

jcfranco:
Thanks! Any name would work for now, I just wanted a separate issue for discussion and to flesh out the requirements if we decide to proceed.

PR #3004, opened 2021-09-09
BREAKING: rename calciteSliderChange event to calciteSliderInput

anveshmekala:
**Related Issue:** #2652 

## Summary

This is breaking change to replace the `calciteSliderChange` event with `calciteSliderInput` and let the `calicateSliderChange` event to emit value on thumb release

- `calciteSliderInput` – fires on other interactions that update value and as the thumb is dragged (on move)
- `calciteSliderChange` – fires on other interactions that update value and during a drag, it only fires on thumb release


src/components/calcite-slider/calcite-slider.tsx
```diff
@@ -758,14 +758,21 @@ export class CalciteSlider {
    * expensive operations consider using a debounce or throttle to avoid
    * locking up the main thread.
    */
+  @Event() calciteSliderInput: EventEmitter;
+
+  /**
+   * Fires on when the thumb is released on slider
```
jcfranco, line 764: The change event should also fire when the value is updated via keyboard. The only difference between this and the input event is that change event does not fire between the thumb being pressed and released: https://codepen.io/jcfranco/pen/VwWMjBP

jcfranco (changes requested):
Looking good, but the the event emitting logic needs a little bit of tweaking.

src/components/calcite-slider/calcite-slider.tsx
```diff
@@ -709,14 +709,14 @@ export class CalciteSlider {
...
-
     if (isNaN(adjustment)) {
       return;
     }
-
     event.preventDefault();
     this[activeProp] = this.clamp(adjustment, activeProp);
-    this.emitChange();
+    if (this[activeProp] === adjustment) {
+      this.emitInput();
```
jcfranco, line 718: This is not correct. Both input and change events should be emitted in this case.

src/components/calcite-slider/calcite-slider.tsx
```diff
@@ -810,6 +822,7 @@ export class CalciteSlider {
 
   @State() private tickValues: number[] = [];
 
+  @State() private isValueChanged: boolean;
```
jcfranco, line 825: If `isValueChanged` is not used for rendering it doesn't need the `@State` decorator.

jcfranco: This prop may be tricky to maintain since there are 2 possible values that can change (value for single-thumb and minValue/maxValue for 2-thumb). Is there another way this can be tracked? I think you could do this without a variable if you always check the values before updating them.

jcfranco (changes requested):
Getting close! 😎

src/components/calcite-slider/calcite-slider.tsx
anveshmekala: I misunderstood that we don't need to emit `change` event for keyboard interactions.

src/components/calcite-slider/calcite-slider.e2e.ts
```diff
@@ -148,7 +148,7 @@ describe("calcite-slider", () => {
     expect(ticks.length).toBe(11);
   });
 
-  it("fires calciteSliderChange event on changes", async () => {
+  it("keyboard interaction", async () => {
```
eriklharper, line 151: nitpick: the title "keyboard interaction" sounds more appropriate for a `describe("keyboard interaction")` section in the tests.  Could you rename this test to make it more meaningful and if it is helpful to group all the keyboard tests together, wrap them in a `describe("keyboard interaction")`.

src/components/calcite-slider/calcite-slider.tsx
```diff
@@ -747,8 +744,12 @@ export class CalciteSlider implements LabelableComponent {
         prop = closerToMax || position > this.maxValue ? "maxValue" : "minValue";
       }
     }
-    this[prop] = this.clamp(position, prop);
+    this.lastDragPropValue = this[prop];
     this.dragStart(prop);
+    const thumbHovered = !!this.el.shadowRoot.querySelector(".thumb:hover");
```
driskull, line 749: Can you store the hovered state on the instance using the mouseenter/mouseexit event on the thumb? Querying through the shadowRoot each time isn't very efficient.

eriklharper: I was going to write a comment to request this as well.  It also allows easier discoverability as to where the event listener is wired up to respond to mouse events on the thumb.

anveshmekala: Changed the query to look for `:active `state rather than hover keeping mobile in mind. 
cc: @driskull , @eriklharper

src/components/calcite-slider/calcite-slider.tsx
```diff
@@ -747,8 +744,12 @@ export class CalciteSlider implements LabelableComponent {
         prop = closerToMax || position > this.maxValue ? "maxValue" : "minValue";
       }
     }
-    this[prop] = this.clamp(position, prop);
+    this.lastDragPropValue = this[prop];
     this.dragStart(prop);
+    const isThumbActive = this.el.shadowRoot.querySelector(".thumb:active");
```
driskull, line 749: @anveshmekala this still needs to change to not query the shadowRoot every time a render occurs.

Instead, store the active state value in the class.

driskull (changes requested):
Getting closer!

src/components/calcite-slider/calcite-slider.tsx
jcfranco: @driskull This query only happens on `pointerdown`, so it doesn't have a noticeable hit on performance.

driskull: Ok. I think as a best practice we should avoid querying the DOM/shadowRoot when possible though.

jcfranco: Agreed, but I think slider needs a bigger refactor before doing so (e.g., consolidating handle rendering, improving/lessening queries on post render).

driskull: Sounds good. Ignore this feedback @anveshmekala :)

src/components/calcite-slider/calcite-slider.e2e.ts
```diff
@@ -148,7 +151,7 @@ describe("calcite-slider", () => {
     expect(ticks.length).toBe(11);
   });
 
-  it("fires calciteSliderChange event on changes", async () => {
+  it("keyboard interaction : key press should change the value and emit input and change events", async () => {
```
jcfranco, line 154: Nitpick: for consistency, can you make sure the `:` goes right after `interaction`?

Sidebar (not for this PR), I think anything `<category>: <test description>` should be using a `describe` block. The `:` is an indicator that the test is part of a group:

```ts
describe("keyboard interaction", () => {
  it("should change the value and emit input and change events", ...);
})
```

src/components/calcite-slider/calcite-slider.e2e.ts
```diff
@@ -162,34 +165,35 @@ describe("calcite-slider", () => {
...
-          <style>body { margin: 0; }</style>
-          <div style="width:100px">
-            <calcite-slider snap></calcite-slider>
-          </div>
-        `
+        html: `<calcite-slider snap style="width:${sliderWidthFor1To1PixelValueTrack}"></calcite-slider>`
       });
+      await page.waitForChanges();
       const slider = await page.find("calcite-slider");
       const changeEvent = await slider.spyOnEvent("calciteSliderChange");
```
jcfranco, line 186: Can you make sure all tests that assert events spy on both the change and input events?

jcfranco (approved):
Other than a few comments about the tests, this LGTM! 🎉🌮

src/components/calcite-slider/calcite-slider.e2e.ts
anveshmekala: added assertion for both input and change events.

Issue #2701, opened 2021-08-02
Docs: document undocumented props/methods/events


caripizza:
### Description 
This is a follow-on to #2241. This issue should add descriptions for all undocumented items in .tsx files.

You can find these easily when you add the strict flag to the `docs-readme` output target in stencil.config.ts. Check out master and run a local build; you'll see the following warnings:

```
[ WARN  ]  Property "active" of "calcite-accordion-item" is not documented.
           ./src/components/calcite-accordion-item/calcite-accordion-item.tsx


[ WARN  ]  Method "setFocus" of "calcite-action" is not documented. ./src/components/calcite-action/calcite-action.tsx


[ WARN  ]  Method "setFocus" of "calcite-action-bar" is not documented.
           ./src/components/calcite-action-bar/calcite-action-bar.tsx


[ WARN  ]  Method "setFocus" of "calcite-action-menu" is not documented.
           ./src/components/calcite-action-menu/calcite-action-menu.tsx


[ WARN  ]  Method "setFocus" of "calcite-action-pad" is not documented.
           ./src/components/calcite-action-pad/calcite-action-pad.tsx


[ WARN  ]  Method "setFocus" of "calcite-button" is not documented. ./src/components/calcite-button/calcite-button.tsx


[ WARN  ]  Method "setFocus" of "calcite-checkbox" is not documented.
           ./src/components/calcite-checkbox/calcite-checkbox.tsx


[ WARN  ]  Property "value" of "calcite-chip" is not documented. ./src/components/calcite-chip/calcite-chip.tsx


[ WARN  ]  Method "setFocus" of "calcite-chip" is not documented. ./src/components/calcite-chip/calcite-chip.tsx


[ WARN  ]  Method "reposition" of "calcite-combobox" is not documented.
           ./src/components/calcite-combobox/calcite-combobox.tsx


[ WARN  ]  Method "setFocus" of "calcite-combobox" is not documented.
           ./src/components/calcite-combobox/calcite-combobox.tsx


[ WARN  ]  Property "rangeHover" of "calcite-date-picker-day" is not documented.
           ./src/components/calcite-date-picker-day/calcite-date-picker-day.tsx


[ WARN  ]  Property "hoverRange" of "calcite-date-picker-month" is not documented.
           ./src/components/calcite-date-picker-month/calcite-date-picker-month.tsx


[ WARN  ]  Property "active" of "calcite-dropdown" is not documented.
[...]

PR #3006, opened 2021-09-10
docs: Add documentation for undocumented properties, methods, and events

driskull:
**Related Issue:** #2701

## Summary

docs: Add documentation for undocumented properties, methods, and events. #2701

### DIFF
--- a/src/components/calcite-date-picker-day/calcite-date-picker-day.tsx
+++ b/src/components/calcite-date-picker-day/calcite-date-picker-day.tsx
@@ -59,6 +59,7 @@ export class CalciteDatePickerDay {
   /** Date is the end of date range */
   @Prop({ reflect: true }) endOfRange = false;
 
+  /** Date is being hovered and within the set range */
   @Prop({ reflect: true }) rangeHover = false;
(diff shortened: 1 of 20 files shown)


src/components/calcite-date-picker-day/calcite-date-picker-day.tsx
```diff
@@ -59,6 +59,7 @@ export class CalciteDatePickerDay {
   /** Date is the end of date range */
   @Prop({ reflect: true }) endOfRange = false;
 
+  /** Date is being hovered and within the set range */
```
driskull, line 62: @paulcpederson @jcfranco I wonder if we should set all these props to be internal since a user would never create this component? Same with the other internal date-picker components. Can we just opt a whole file out to be internal??

jcfranco: I think it's fine as long as support components don't get pulled into the SDK pages. @benelan @julio8a WDYT?

Separate discussion, but I imagine a few of these can be converted into functional components too.

PR #3010, opened 2021-09-10
chore: fix CDN-updating logic

jcfranco:
**Related Issue:** #2623

## Summary

Fixes the CDN-URL-updating code:

* update regexps to handle formatted script/link tags – previously would not match after formatting
* use releaseAs version to generate CDN URLs – previously would use pre-bump/outdated version


driskull (approved):
💥

Issue #2994, opened 2021-09-08
Bug: calcite-color-picker ignores initial value


AdelheidF:
beta.63

```
<calcite-color-picker
            scale="m"
            value="#FF0000"
            format="hex"
            appearance="default"
            hideSaved="true"
          ></calcite-color-picker>
```

any hex color value displays blue swatch
<img src=https://user-images.githubusercontent.com/44481292/132595662-a213ad95-5635-4999-a581-d46231a49545.png width=300/>

I tried to 're-set' the value again in ref, but this doesn't help.

If I remove `format` it works fine.

Repro case: https://codepen.io/jcfranco/pen/dyRvVpX?editors=1000

jcfranco:
Installed.

AdelheidF:
Tested with beta.64, all good now. Thanks.

PR #3020, opened 2021-09-10
fix(color-picker): render current value when set initially with custom format

jcfranco:
**Related Issue:** #2994 

## Summary

The way the initial value and default were set would sometimes skip setting the internal color and mess up rendering if there was a custom value and format set by the user.


driskull (approved):
👍

Issue #2889, opened 2021-08-24
Bug: calcite-sortable-list around calcite-block changes block layout


AdelheidF:
When I add `calcite-sortable-list` around `calcite-block` elements that I want to sort they suddenly appear all in one line.
I have to add `flex-flow: column nowrap;` to `calcite-sortable-list` to make them appear properly.

<img src=https://user-images.githubusercontent.com/44481292/130658923-d3c1af07-99a3-46f3-bf78-85b39068412f.png width=250/>

![image](https://user-images.githubusercontent.com/44481292/130658962-d9fa7201-34a2-4906-b3ca-1661372aab41.png)

@asangma suggested to add a layout property to SortableList;

### Reproduction Steps or Sample

1.

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

asangma:
Simple solution for this would be to add a default style that lays content out in a column.

A more involve (though not much more involved) solution would be to add a `layout` property that defaults to "vertical".

asangma:
I can do either.

AdelheidF:
not yet fixed with beta.65 https://codepen.io/afreitag/pen/jOwKGmK?editors=1010

benelan:
verified in beta.66 using the sample above

PR #3031, opened 2021-09-13
fix(sortable-list): updated to use column layout

asangma:
**Related Issue:** #2889

## Summary
* Updates to use flex column layout 
* Adds `layout` property that defaults to "vertical"

### DIFF
--- a/src/components/calcite-sortable-list/calcite-sortable-list.tsx
+++ b/src/components/calcite-sortable-list/calcite-sortable-list.tsx
@@ -49,7 +49,7 @@ export class CalciteSortableList {
   /**
    * Indicates the horizontal or vertical orientation of the component.
    */
-  @Prop({ reflect: true }) layout: Layout = "vertical";
+  @Prop() layout: Layout = "vertical";
 
   /**
    * When true, disabled prevents interaction. This state shows items with lower opacity/grayed.
--- a/src/components/calcite-sortable-list/calcite-sortable-list.tsx
+++ b/src/components/calcite-sortable-list/calcite-sortable-list.tsx
@@ -208,14 +208,14 @@ export class CalciteSortableList {
 
   render(): VNode {
     const { layout } = this;
-    const layoutProp = layout == "vertical" || layout == "horizontal" ? layout : "vertical";
-    const layoutClass = layoutProp === "vertical" ? CSS.containerVertical : CSS.containerHorizontal;
+    const horizontal = layout === "horizontal" || false;
 
     return (
       <div
         class={{
           [CSS.container]: true,
-          [layoutClass]: true
+          [CSS.containerVertical]: !horizontal,
+          [CSS.containerHorizontal]: horizontal
         }}
       >
         <slot />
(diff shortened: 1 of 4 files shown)


src/components/calcite-sortable-list/calcite-sortable-list.tsx
```diff
@@ -200,6 +207,19 @@ export class CalciteSortableList {
   // --------------------------------------------------------------------------
 
   render(): VNode {
-    return <slot />;
+    const { layout } = this;
+    const layoutProp = layout == "vertical" || layout == "horizontal" ? layout : "vertical";
+    const layoutClass = layoutProp === "vertical" ? CSS.containerVertical : CSS.containerHorizontal;
```
asangma, line 212: @driskull If you have a suggestion for a cleaner way to do this, I'd love to know it. :)

driskull: @asangma how about this?

```ts
const layoutClass =(layout ?? "vertical") === "vertical" ? CSS.containerVertical : CSS.containerHorizontal;
```

or 

```ts
const layoutProp = layout  ?? "vertical";
```

```ts
class={{
          [CSS.container]: true,
          [CSS.containerVertical]: layoutProp === "vertical",
          [CSS. containerHorizontal]: layoutProp === "horizontal"
        }}
```

driskull (changes requested):
Just some code cleanup

src/components/calcite-sortable-list/calcite-sortable-list.tsx
asangma: Ah...that'd work if checking for a null value yeah?
I think our default value should handle that.

I'm concerned about an invalid value, e.g. "foo".

I guess I could redo the style to be column and add a row rule when the horizontal CSS class is added?

driskull: Yeah that would be better. make the fallback a little easier.

```ts
class={{
          [CSS.container]: true,
          [CSS.containerVertical]: effectiveLayout !== "horizontal",
          [CSS. containerHorizontal]: effectiveLayout === "horizontal"
        }}
```

driskull:
```
const horizontal = layout === "horizontal"  ||  false;

class={{
          [CSS.container]: true,
          [CSS.containerVertical]: !horizontal,
          [CSS. containerHorizontal]: horizontal
        }}
```

asangma: Thanks @driskull.

asangma:
Thanks @driskull. Your suggestion is kleen.

src/components/calcite-sortable-list/calcite-sortable-list.tsx
```diff
@@ -44,6 +46,11 @@ export class CalciteSortableList {
    */
   @Prop() handleSelector = "calcite-handle";
 
+  /**
+   * Indicates the horizontal or vertical orientation of the component.
+   */
+  @Prop({ reflect: true }) layout: Layout = "vertical";
```
driskull, line 52: Does it need to be reflected? Seems like it doesn't.

driskull (approved):
👍

src/components/calcite-sortable-list/calcite-sortable-list.tsx
asangma: Makes sense. Will update.

Issue #2962, opened 2021-09-01
Tooling: create GitHub Action to make sure new bug issues have samples


benelan:
## Summary 
We will now be requiring sample code to reproduce bug issues. The issue template will say this, but I will make a GH Action just in case. The action should: 
1. check for no link to sample code (codepen/jsbin/jsfiddle/etc)
2. automate comment / add needs more info label if there is no sample
3. if no new comment with URL after 5 days, close issue with comment
4. if follow up comment has url meeting criteria, remove needs more info, don't close.

## Measure of Success 
A working GH action for the workflow above

benelan:
Installed! @julio8a or @jcfranco, can I get a verifier

julio8a:
🔥 verified on master

PR #3036, opened 2021-09-14
chore: new bugs require a repro sample

benelan:
**Related Issue:** #2962 

## Summary
Repro samples and the full issue template format will now be required when logging bugs. 

- Updated the contributing doc and bug issue template to mention the new requirement
- Added a new [github action](https://github.com/benelan/need-info-action) to automatically request more info and close the issue if none are provided
- If an issue has a `need more info` label it will be closed in three days unless info is provided, so keep that in mind when adding the label manually


driskull (approved):
Nice! 🥇 

These are super helpful @benelan

Issue #2808, opened 2021-08-13
Bug: Stepper needs to use -h text sizing instead of -wrap


bstifle:
Stepper needs to use -h text sizing instead of -wrap in all scales and sizes, headers and summaries, and the numbers

### Actual Behavior
Stepper type using text-wrap

### Expected Behavior
Stepper type using text-h

julio8a:
Verified on master demo pages.

PR #3038, opened 2021-09-14
refactor(stepper): change text sizing to use -h

justinhough:
**Related Issue:** #2808

## Summary

Stepper needs to use -h text sizing instead of -wrap in all scales and sizes, headers and summaries, and the numbers


julio8a (approved):
Thanks!

Issue #3035, opened 2021-09-14
Bug: [color-picker] dragging the color field thumb gets offset by content above when placed in a component's shadow DOM


jcfranco:
### Actual Behavior

When the color-picker is placed in a shadow DOM, any content above will mess up the positioning used when dragging the color field around.

### Expected Behavior

The color-picker's color field drag interaction should behave consistently regardless of where it is placed (light or shadow DOM).

### Reproduction Steps or Sample

1. Open https://codepen.io/jcfranco/pen/xxrXEPr?editors=0010
2. Click on the color field, press and move the pointer around
3. Notice how the position is offset by the container above

### Relevant Info

_Version_: `@esri/calcite-components@1.0.0-beta.61`

- [x] CDN
- [ ] NPM package

jcfranco:
Installed

benelan:
Verified in beta.65 using the codepen in the issue.

PR #3041, opened 2021-09-14
fix(color-picker): fix mouse tracking logic when moved within another component's shadow DOM

jcfranco:
**Related Issue:** #3035

## Summary

Fixes an issue where the X/Y coordinates were sampled incorrectly while dragging the color field thumb when the component was placed within another shadow DOM root. 

**Note**: I'm looking at adding a specific test to catch this that doesn't involve an additional test suite run within shadow DOM.


driskull (approved):
👍

Issue #3025, opened 2021-09-13
Bug: calcite-action-bar:  we dont see a scrollbar anymore when the window size is small


kevindoshier:
### Actual Behavior

When the window size is shrunk vertically we no longer see the scrollbar

### Expected Behavior

We should see a scrollbar so we can access the other actions
![scrollbar](https://user-images.githubusercontent.com/11093161/133150518-852a2239-14ba-49e0-b3cb-7195bdf15eb9.png)

### Reproduction Steps or Sample

1.

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

driskull:
@jcfranco @julio8a can we triage this for this sprint?

jcfranco:
Added to current sprint! I assigned it to you in the meantime. cc @benelan

driskull:
@jcfranco @kevindoshier I'm unable to locally reproduce this so I think it might be a timing issue. However, I have some improvements that may fix it. PR coming.

zaramatheson:
@kevindoshier @driskull Is it expected that this should be in the MV today? We are still seeing an issue on 1280x1024 monitor and smaller window sizes

![image](https://user-images.githubusercontent.com/50747646/133675803-8498eb95-7f19-47fb-95a4-1316a037ba75.png)

benelan:
@zaramatheson it should be in beta.65, is that the version being used?

kevindoshier:
Todays build is using beta.65,  also Im able to repro here if I shrink the window really small

https://esri.github.io/calcite-components/?path=/story/components-shell--basic

driskull:
@kevindoshier there's not enough icons to reproduce it in that sample.

kevindoshier:
@driskull - thats true, will this work?
https://codepen.io/kevindoshier/pen/rNwpKVe

driskull:
Seems like its working correctly there if you set the height to 400px in the code instead of 800.

@kevindoshier Can you tell me why this shell container has a height of 0? seems like that might be an issue.

![image](https://user-images.githubusercontent.com/1231455/133679068-058a284e-8067-4981-b8f5-006f04ce58c4.png)

Also, seems like the body of the page should have a margin and padding of 0 set.

kevindoshier:
I set it to 400 and still dont see a scrollbar on the actionBar

[...]

PR #3042, opened 2021-09-14
fix(action-bar): Better handling of overflow actions on smaller screens

driskull:
**Related Issue:** #3025

## Summary

fix(action-bar): Better handling of overflow actions on smaller screens #3042

## This PR

- Modifies the mutationObserver to listen to `subtree` changes.
- Calls resize on `connectedCallback`
- Checks resize for an actual height greater than 0 before continuing.
- Cleans up code for resizing based on the host element clientHeight.
- Cleans up code for a slot name.

### DIFF
--- a/src/components/calcite-action-bar/calcite-action-bar.e2e.ts
+++ b/src/components/calcite-action-bar/calcite-action-bar.e2e.ts
@@ -301,4 +301,87 @@ describe("calcite-action-bar", () => {
 
     expect(await button.getProperty("scale")).toBe("l");
   });
+
+  describe("overflow actions", () => {
+    it("should slot 'menu-actions' on sublist changes", async () => {
+      const page = await newE2EPage({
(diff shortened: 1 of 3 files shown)


driskull:
@jcfranco can we maybe get this into the release for MV?

jcfranco (approved):
LGTM! Before merging, can you add a test for this? It's labeled as fix, but it has no repro case nor info on how to reproduce.

driskull:
@jcfranco I'll change it to `feat` since it may not be a fix. However, the change is to start listening for sublist changes on the mutation observer. I'm not sure if we need a test for that part.

jcfranco:
> However, the change is to start listening for sublist changes on the mutation observer. 

That's a change of behavior to me (feat) and would warrant a test.

jcfranco (approved):
Noice! 🎉 🌮 :t-rex:

Issue #2826, opened 2021-08-16
Bug: calcite-tile's "large visual" variant doesn't center the content


eriklharper:
### Actual Behavior

The large icon in `calcite-tile`'s "large visual" option displays on the left in LTR and on the right in RTL modes.

![image](https://user-images.githubusercontent.com/821864/129614965-8de60fc3-9ba8-4a88-9125-5929bbf9110a.png)

![image](https://user-images.githubusercontent.com/821864/129615038-caa3d1c5-82e9-4d86-aac7-b2f4096ffa86.png)

### Expected Behavior

The icon and description should always be displayed in the center in either reading mode.

### Reproduction Steps or Sample

```
<calcite-tile icon="layers" heading="Tile title lorem ipsum"></calcite-tile>
```

### Relevant Info

All browsers/OS

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

benelan:
Verified on beta.66
https://codepen.io/benesri/pen/KKqJPod

PR #3044, opened 2021-09-15
refactor(calcite-tile): center content on large variant

justinhough:
**Related Issue:** #2826

## Summary

Centering icon and content for large visual `calcite-tile` when description is not present on both LTR and RTL.


jcfranco (approved):
Code changes LGTM! 

@caripizza Can you review Screener changes after @justinhough merges the latest from master? 🙇🏼‍♂️

src/components/calcite-tile/calcite-tile.scss
```diff
@@ -21,6 +21,14 @@
...
   @apply p-0;
 }
+:host(:not([input-alignment]):not([description])) .large-visual {
+  .icon {
+    @apply flex justify-center;
+  }
+  .content {
+    text-align: center;
+  }
+}
```
caripizza, line 31: I think all that is needed to center the icon in the "large visual" variant of Tile (ie., when there's no description passed in, only heading + icon) is this portion:

```scss
:host(:not([description])) .large-visual {
  .icon {
    @apply flex justify-center;
  }
}
```

(`input-alignment` refers to the Tile Select prop, not a prop within the Tile component.)

justinhough: @caripizza I added the text alignment for this scenario, but not sure if this would really be a use case in the real world. 
CleanShot 2021-09-17 at 16 41 38

caripizza: Ah didn't realize that, thanks for clarifying for me. In that case it seems like this may be better?

```scss
:host([icon][heading]:not([description])) {
  .large-visual {
    @apply text-center;
  }
}
```

caripizza: I'm not seeing the text-center style?

justinhough: Thought I had committed it up already, but I guess I removed it.

caripizza (approved):

Issue #2860, opened 2021-08-19
Docs: Add basic usage example to all components


benelan:
### Description 
There are a lot of components without usage examples. I will add at least a 'Basic' example to all of them.

macandcheese:
We'd like to eventually consume these samples in the developer site, where we can generate component samples from the provided snippets.

It would be great if each example could have explicit data points for title, description, and then the code sample. Perhaps having individual files for each sample?

Right now, it's hard to parse the returned files since it's often a mix of inline code, text, other markdown formatting, etc., cc @caripizza

benelan:
@macandcheese Do you want to set up a meeting next week to discuss this? I can do it however works best for you and then go through and change the rest of them too.

benelan:
Components that still have missing usage examples:
- [ ] card
- [ ] input-date-picker
- [ ] input-time-picker
- [ ] calcite-radio-group
- [ ] calcite-rating (needs another)
- [ ] link
- [ ] pagination (needs rework)
- [ ] progress (has a TODO header??)
- [ ] tile

benelan:
Installed, @julio8a or @jcfranco, can I get a verification please.

julio8a:
Verified on master

PR #3050, opened 2021-09-15
docs(usage): add missing examples

benelan:
**Related Issue:** #2860 

## Summary
Added a basic usage example to the components that did not have any examples. The child components such as `calcite-*-item` still do not have usage examples because they are under their parent on storybook which do have examples. If whoever reviews this thinks I should add them to _every_ component I can add that to this PR.


src/components/calcite-input-date-picker/usage/Basic.md
```diff
@@ -0,0 +1,26 @@
+```html
+<div style="width: 400px">
+  <calcite-label
+    layout="inline"
+    class="sc-calcite-label-h sc-calcite-label-s"
```
driskull, line 5: @benelan can we remove the class here? I think these are added by stencil because they are scoped components currently. Applies to other usage snippets.

driskull (approved):
Looks good! Lets just get rid of those stencil generated classes.

The modifications to the non usage readme's will probably get overriden by the next docs generation so those files don't need to be checked in next time.

src/components/calcite-input-date-picker/usage/Basic.md
benelan: Nice catch, I removed those classes from most of the snippets but missed that one. I searched all of the usage examples, the rest are fine

PR #3054, opened 2021-09-16
refactor(action-bar): Refactor action-bar overflowActions.

driskull:
**Related Issue:** #3025

## Summary

refactor(action-bar): Refactor action-bar overflowActions.

- Adds public method to call `overflowActions` a user can use if needed to trigger a resize manually
- Refactors resize logic to debounce instead of tracking last values


driskull:
@jcfranco I ran this a couple times and it passed. Can you give it a look? 👀

src/components/calcite-action-bar/calcite-action-bar.tsx
```diff
@@ -166,6 +166,15 @@ export class CalciteActionBar {
   //
   // --------------------------------------------------------------------------
 
+  /**
+   * Overflows actions that won't fit into menus.
+   * @internal
```
driskull, line 171: we could make this public as well.

jcfranco: Is this method solely for testing?

driskull: As of now, yes.

driskull: I'll make it public. It can be useful for situations where a user may need to trigger a resize manually.

jcfranco: For now, let's make it internal for testing. We can keep this PR as refactor that way too.

jcfranco (approved):
Good to merge once `overflowActions` is made hidden via `@internal`.

jcfranco:
@driskull Do you have any ideas on what's causing the automatic overflow logic to not work in some cases? We can investigate this further in a separate issue. It'd be nice to handle overflow automatically all the time.

driskull:
> Do you have any ideas on what's causing the automatic overflow logic to not work in some cases? 

Yes, Its this bad logic here: https://github.com/Esri/calcite-components/pull/3054/files#diff-b26c11cd8c459dae40710c10802a373e61af78ee544f63b4688e3bb3054a85a0L233-L244

Debouncing instead of doing this fixes the issue.

Issue #2481, opened 2021-07-07
Bug: slider's max value thumb can't be moved via pointer if thumbs overlap at min edge


jcfranco:
### Actual Behavior

The max thumb can't be grabbed when thumbs overlap at the min edge. Users can tab into the thumb and move it via the keyboard, but this seems broken.

### Expected Behavior

When both thumbs overlap at the edge, click/tap should grab the max value thumb

### Reproduction Steps or Sample

https://codepen.io/jcfranco/pen/eYWZxxz?editors=1000

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [x] CDN
- [ ] NPM package

caripizza:
Installed via pr #3078

benelan:
verified on beta.66

Issue #2480, opened 2021-07-07
Bug: slider's min value thumb not placed correctly when range is -100 – 0 and both min and max are 0


jcfranco:
### Actual Behavior

For a range of -100 – 0, sliders are placed at opposite ends when both min-value/max-value are 0

### Expected Behavior

Both thumbs to overlap at 0

### Reproduction Steps or Sample

https://codepen.io/jcfranco/pen/LYyNqmj?editors=1000

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [x] CDN
- [ ] NPM package

caripizza:
Happened upon this one while looking into slider issues with #2457. 

Beyond the thumb button placement issue mentioned here, the `min` label handle also displays `0` when it should show `-100`:

![image](https://user-images.githubusercontent.com/42423180/132380560-31b397bf-1940-4be7-9276-6d4ab504328e.png)

caripizza:
I'm stuck on this issue. The main problem to me seems to be that when slider has a negative `min` value in its range, it disallows a `minValue` of `0`; the handle get's `left: 0%` positioning added to it, which forces it to appear to the far left of the track.

It's like slider's range was designed to assume that the minimum value would be 0, not a negative number.

caripizza:
Installed via pr #3083

caripizza:
@benelan it looks like this also happens in a histogram range - I'll open a pr for that edge case fix today since I missed that before.

caripizza:
Additional histogram fix installed via pr #3092.

benelan:
verified on beta.66

PR #3060, opened 2021-09-16
fix(slider): ensure range thumbs are placed correctly when min/max share the same value

caripizza:
**Related Issue:** #2480, #2481

## Summary
Adds the following fixes for calcite-slider:

1. Fixes an issue where slider's minValue thumb is not placed correctly when both min and max are `0` (2480). (Repro with [Storybook iframe url](https://esri.github.io/calcite-components/iframe.html?id=components-controls-slider--range&knob-min=-100&knob-min-label=Temperature,%20lower%20bound&knob-min-value=0&knob-max=0&knob-max-label=Temperature,%20upper%20bound&knob-max-value=0&knob-step=1&knob-ticks=20&knob-label-ticks=true&knob-label-handles=false&viewMode=story))

2. Fixes a related issue where slider's maxValue thumb can't be moved via pointer if thumbs overlap at min edge (2481). (Repro with [Storybook iframe url](https://esri.github.io/calcite-components/iframe.html?id=components-controls-slider--range&args=&knob-min=5&knob-min-label=Temperature,%20lower%20bound&knob-min-value=5&knob-max=100&knob-max-label=Temperature,%20upper%20bound&knob-max-value=5&knob-step=10&knob-ticks=10&knob-snap=true&knob-label-handles=true&knob-label-ticks=true&viewMode=story))

3. Cleans up some of the .tsx logic by adding and referencing the resources.ts and interfaces.ts files per convention.


caripizza:
One thing I'm trying to do on this pr branch is better handle slider's tick calculations tied to precision loss with decimals (ex: `console.log(.1 + .2);  // 0.30000000000000004`). 

When the ticks get generated, I'm converting the numbers to localeString() (like the labels), to get the right calculations for determining the tick offset percentages. (I can pull this out if it's not the way we want to handle generating the tick values/if we plan to normalize the number values in some other way. I realize the `Number(from offset string)` mutation isn't ideal. :/ )

src/components/calcite-slider/calcite-slider.tsx
```diff
@@ -143,11 +145,20 @@ export class CalciteSlider {
...
     const maxProp = this.isRange ? "maxValue" : "value";
     const value = this[maxProp];
+    const valueForLabeledHandle = value && value.toLocaleString();
+    const minValueForLabeledHandle = this.minValue && this.minValue.toLocaleString();
     const minInterval = this.getUnitInterval(min) * 100;
     const maxInterval = this.getUnitInterval(max) * 100;
     const mirror = this.shouldMirror();
-    const leftThumbOffset = `${mirror ? 100 - minInterval : minInterval}%`;
     const rightThumbOffset = `${mirror ? maxInterval : 100 - maxInterval}%`;
+    this.minValueZeroFlag = !!(this.isRange && !this.hasHistogram && this.minValue === 0);
```
eriklharper, line 154: ☝️ This looks like state is being modified in the render method.  Does stencil give you a warning when doing this, since it is not encouraged?

caripizza: @eriklharper I don't see any warning nope but I'll move this out, good call

src/components/calcite-slider/calcite-slider.tsx
```diff
@@ -825,7 +911,13 @@ export class CalciteSlider {
     let current = this.min;
     while (this.ticks && current < this.max + this.ticks) {
       ticks.push(current);
-      current = current + this.ticks;
+      // match tick label formatting; handle binary floating point numbers not mapping correctly to decimals (precision loss with `number` type)
+      // https://v8.dev/features/bigint
+      const currentWithTick = (current + this.ticks).toLocaleString();
```
eriklharper, line 916: Assuming `this.ticks` always contains non-localized numbers, then we probably shouldn't convert ☝️ this value to a locale string to immediately construct it back to a `Number` here 👇 .  The locale string versions of the numbers should only be used for display.

caripizza: Got it, I'll remove 👍

caripizza:
Closing this one, I'll open a new pr for each individual issue cc @jcfranco

PR #3073, opened 2021-09-19
chore(actions): need info edit event removes label

benelan:
## Summary
- When an issue is marked with the `need more info` label and the user edits the issue body to add all of the required items, the label is removed
- This came up in issue 3071 (I don't want to actually link the issue since this isn't a fix for it)

## Changes
- add edit event to gh action
- bump action version to correctly handle edit events


jcfranco (approved):
👍🏼

PR #3085, opened 2021-09-21
chore: format yml files using prettier

benelan:
## Summary
Use prettier to format yml files the same way we do for other file types


jcfranco (approved):
LGTM! Before merging, could you do the following?

1. update [`.lintstagedrc.json`](https://github.com/Esri/calcite-components/blob/master/.lintstagedrc.json) to pick up `yml` files 
1. split this up into a PR with ☝🏼  + the NPM linting script and a separate one for formatting?

jcfranco (approved):
👍🏼

PR #3088, opened 2021-09-21
chore(actions): fix milestone action to stop failing on dependabot prs

benelan:
**Related Issue:** #3077

## Summary
The milestone action is failing on PRs by dependabot. This is by design for security reasons, as described [here](https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/).

I [changed](https://github.com/benelan/milestone-action/blob/main/src/main.ts#L13-L15) the action to skip PRs by dependabot for the time being until I think of a better solution.

### DIFF
--- a/.github/workflows/pr-bot.yml
+++ b/.github/workflows/pr-bot.yml
@@ -17,4 +17,4 @@ jobs:
   milestone:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


jcfranco (approved):
💥

PR #3089, opened 2021-09-21
chore(actions): no self assign for verification

benelan:
**Related Issue:** 2881

## Summary
When I am the one who installed an issue it should not assign me for verification. It now assigns franco and julio to verify my issues.


.github/workflows/assign-for-verification.yml
```diff
@@ -1,26 +1,28 @@
...
+  assign:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v4
+        with:
+          script: |
+            const { label, sender } = context.payload;
+            if (sender && label && label.name === "3 - installed") {
+              const assignees =
+                sender.login !== "benelan" ? ["benelan"] : ["jcfranco", "julio8a"];
```
jcfranco, line 15: Not for this PR and more of an idea, but it'd be useful to have some sort of `VERIFIERS` variable for this and possibly shared between actions. Maybe as GH ssssssssssecrets. 🤫

jcfranco (approved):
💥 💥 🪃

.github/workflows/assign-for-verification.yml
benelan: @jcfranco Good idea, if I end up needing to use our usernames in another action I'll make that change.

PR #2777, opened 2021-08-10
refactor: convert scoped components to shadowed

driskull:
**Related Issue:** #2570

## Summary

This PR refactors any remaining scoped components to be shadowed. Form integration is also included in this work as making components shadowed had an impact on existing usage.

## Notable changes

- to ease setup, a utility module with helpers (`src/utils/form.tsx`) was added along with a test helper `src/tests/commonTests#formAssociated`). 
- component-specific code for form integration has been removed
- for consistency, a `value` prop was added to all form components
  - not reflected as not all component usage is form-related and may be rich-data in some cases
  - `start`/`end`/`startAsDate`/`endAsDate` deprecated in both `calcite-input-date-picker` & `calcite-date-picker`
- for form-submitting, `disabled`, `name` and `required` props have been added
  - any new `required` props on components will be internal at the moment to allow for proper design
- more components are now form-associated, this is the current list:
  - `calcite-checkbox` 
  - `calcite-combobox` 
  - `calcite-input` 
  - `calcite-input-date-picker` 
  - `calcite-input-time-picker` 
  - `calcite-radio-button` 
  - `calcite-radio-button-group` 
  - `calcite-rating` 
  - `calcite-select` 
  - `calcite-slider` 
  - `calcite-switch`

### `FormComponent` pattern
    
This pattern enables components to participate in form submission, resetting and reporting validity errors. 

In order for a component to work with an HTML `<form>`, a component must:

1. Implement the `FormComponent` interface
2. Use the `connectForm`/`disconnectForm` utils in the component's `connectedCallback`/`disconnectedCallback` lifecycle methods. 
3. Use the `HiddenFormInputSlot` functional component in the component's `render` method. It must be placed at the bottom of the component's root container.
4. Include the `hidden-form-input` mixin in the component's `scss` file 
5. **If the default value cannot be determined in `connectedCallback`**, the component can use the `afterConnectDefaultValueSet` helper in a more appropriate lifecycle method. 
6. **If the component needs special handling of its associated hidden input**, the component can implement a `syncHiddenFormInput` method with custom input handling logic. 
7. **If the component needs special handling resetting the value to the default**, the component can implement a `onFormReset` method with custom resetting logic.
7. Add an appropriate test by using the `formAssociated` common test helper.
                                                                            
For more details, you can look at the doc in the interface as well as the implementation on any of the form components from this PR.

## Pending

* [x] Handle form-associated component candidates that don't have a `value` prop:
  * `calcite-combobox`
[...]

Issue #2570, opened 2021-07-19
Epic: Shadow DOM for all components


jcfranco:
Components should avoid using scoped styles as they are not ideal for the following reasons:

1. VDOM integration – VDOMs have their representation of the DOM tree that can potentially clash with any elements within the scoped component's 'internal' HTML structure
2. Breaks encapsulation 
   a. Component internal elements are exposed in light DOM queries   
   b. Scoped CSS is prone to being overridden with high-specificity selectors
   c. Scoped CSS can clash with external ones unless the component's selectors are namespaced (#2422)
3. Light DOM HTML structure differs before & after hydration (related to 1) – this is currently affecting our Storybook preview tab https://github.com/Esri/calcite-components/issues/2362

The biggest impact this would have is on form usage since it means that native form controls are no longer discoverable by light DOM forms. 

Fortunately, we only have a handful of scoped components:

* `calcite-inline-editable`
* `calcite-input`
* `calcite-label`
* `calcite-radio-button`

### Plan of action:

- [x] namespaced classes to help w/ 2.c - https://github.com/Esri/calcite-components/issues/2422  
- [x] assess impact on current form usage ([survey](https://survey123.arcgis.com/share/99cfa715e01843cf9bcd6ef060596228))
- [x] use shadow DOM on current scoped components (needs to happen early and tested early)
  - [x] `<calcite-label>` #3091
- [ ] add custom ESLint rule to disallow scope usage (out of scope)

Related: #2949 https://github.com/Esri/calcite-components/issues/1741

driskull:
@jcfranco should we have issues to convert the components? I can start taking one of them.

jcfranco:
@driskull To keep things simple, I think we can use this issue and have multiple PRs link to tasks/subtasks. cc @julio8a @caripizza 

On the conversion part, we're still waiting on results from the form-usage survey, but it'd be great if you can kick start a draft PR (shadow all the things!). 🥇

driskull:
I think the biggest changes are going to be the logic for handling clicks on `calcite-label` elements.

If the native `label` and native `input` are now in shadow, clicking on a label won't focus or change the input.

In components like `calcite-radio-button`, we could listen for the event `calciteLabelFocus` and handle it. But what about native elements like `<calcite-label><input type="checkbox" /> test</calcite-label>`? The native label and native checkbox can't interact. Any ideas here?

jcfranco:
[...]

PR #3091, opened 2021-09-22
feat!: improve integration with calcite-label

jcfranco:
**Related Issue:** #2570 

## Summary

**Breaking change**: components no longer support working with `<label>`. To fix, existing code should swap `<label>` for `<calcite-label>`.

This PR makes the following components `calcite-label` compatible (following the same pattern):

* button
* checkbox
* combobox
* inline-editable
* input-date-picker
* input-time-picker
* input
* radio-button
* radio-group
* rating
* select
* slider
* switch

### Notable changes

* enables shadow DOM on `calcite-label` (this contains the the label-specific work from #2777)
* adds utilities and accompanying interface for any component to be able to implement and test `calcite-label` compatibility
* Replaces redundant label tests in favor of `labelable` test util
* Drops duplicate label demo page entry
* Drops `dom#hasLabel` util.

### `LabelableComponent` pattern

In order for a component to work with `calcite-label`, a component must:

1. implement the `LabelableComponent` interface and use the `connectLabel`/`disconnectLabel` utils in the component's `connectedCallback`/`disconnectedCallback` lifecycle methods. A `labelable` test helper was introduced to facilitate testing across components. 
2. Use the `getLabelText` util to get the appropriate label to use on the component's `aria-label` property. 

For more details, you can look at the doc in the interface as well as the implementation on any of the labelable components from this PR.


src/tests/commonTests.ts
```diff
@@ -220,3 +221,108 @@ export async function slots(componentTagOrHTML: TagOrHTML, slots: Record<string,
...
+  if (propertyToToggle) {
+    component = await page.find(componentTag);
+    initialPropertyValue = await component.getProperty(propertyToToggle);
+  }
+
+  const label = await page.find("calcite-label");
+  await label.callMethod("click"); // we call the method to avoid clicking the child element
+  await page.waitForChanges();
+
+  console.log(html, propertyToToggle, focusTargetSelector);
```
driskull, line 252: got some left over logs in this file.

src/tests/commonTests.ts
```diff
@@ -220,3 +221,108 @@ export async function slots(componentTagOrHTML: TagOrHTML, slots: Record<string,
...
+      await page.$eval(
+        componentTag,
+        (element: HTMLElement, selector: string) => element.shadowRoot.activeElement.matches(selector),
+        shadowFocusTargetSelector
+      )
+    ).toBe(true);
+  }
+
+  if (propertyToToggle) {
+    console.log(component.outerHTML, initialPropertyValue);
```
driskull, line 272: got some left over logs in this file.

src/tests/commonTests.ts
```diff
@@ -220,3 +221,108 @@ export async function slots(componentTagOrHTML: TagOrHTML, slots: Record<string,
...
+  propertyToToggle?: string;
+}
+
+/**
+ * Helper for asserting label clicking functionality works.
+ *
+ * @param componentTagOrHtml - The component tag or HTML used to test label support.
+ * @param propertyToToggle - The component's property that should be toggled when it's calcite-label is clicked.
+ */
+export async function labelable(componentTagOrHtml: TagOrHTML, options?: LabelableOptions): Promise<void> {
```
driskull, line 291: nice :)

src/components/calcite-switch/calcite-switch.tsx
```diff
@@ -97,7 +104,7 @@ export class CalciteSwitch {
   /** Sets focus on the component. */
   @Method()
   async setFocus(): Promise<void> {
-    focusElement(this.inputEl);
+    this.el.focus();
```
driskull, line 107: For this one, i changed it to focus on the host because the input is hidden so it has no visual state for being focused and can't be tabbed to. We should probably just get rid of this hidden input and use aria roles here.

src/components/calcite-input/resources.ts
```diff
@@ -1,6 +1,8 @@
 export const CSS = {
   loader: "calcite-input__loader",
   clearButton: "calcite-input__clear-button",
+  editingEnabled: "calcite-input__editing-enabled",
```
driskull, line 4: I think these are leftover from the work done to change inline-editable to not affect the input. Seems like these aren't being used though in this PR.

driskull (approved):
Looking pretty 🔥! Just found some minor stuff

src/components/calcite-switch/calcite-switch.tsx
jcfranco: > We should probably just get rid of this hidden input and use aria roles here.

Agreed. I updated `Host` accordingly. I'll submit an issue to clean this (and others?) up since we're [not supposed to override user-set global attributes](https://developers.google.com/web/fundamentals/web-components/best-practices#do-not-override-author-set,-global-attributes.). Regarding the hidden input, let's remove it when we implement `FormComponent` across the board.

src/tests/commonTests.ts
jcfranco: This is what happens when you fix locally, but don't push. 😂

src/components/calcite-input/resources.ts
```diff
@@ -1,6 +1,7 @@
 export const CSS = {
   loader: "calcite-input__loader",
   clearButton: "calcite-input__clear-button",
+  inlineChild: "calcite-input__inline-child",
```
driskull, line 4: Does this still need to be here?

jcfranco: Nope. Removed!

Issue #2914, opened 2021-08-26
Enhancement: support highlighting histogram range when slider's min value is negative


caripizza:
### Description
Add support for highlighting a histogram range when calcite-slider's `min` prop value is a negative number.

While working on #2457, I'm noticing that slider's track range handles the negative `min` value just fine:

![image](https://user-images.githubusercontent.com/42423180/131000372-7e5b6186-9ab3-48e8-b8ab-6f1f4e231b78.png)

However, when the slider includes histogram data, the highlighted region from calcite-graph doesn't yet align vertically:

![image](https://user-images.githubusercontent.com/42423180/131001709-52783c95-eee8-486a-b813-0ba7e98233c1.png)

### Acceptance Criteria 
When calcite-slider's histogram range has a negative `min` prop value, the highlighted track (ie., in between the `minValue` and `maxValue` data points) aligns with the highlighted histogram region.

### Relevant Info 
[...]

PR #3092, opened 2021-09-22
fix(slider): align histogram minValue thumb when both minValue/maxValue are zero

caripizza:
**Related Issue:** #2480 (follow-on to pr #3083)

## Summary
Fixes an issue where a histogram range slider's minValue thumb is not placed correctly when both minValue and maxValue are 0. ([Repro with Storybook iframe url](https://esri.github.io/calcite-components/iframe.html?id=components-controls-slider--histogram&args=&knob-min=-10&knob-min-value=0&knob-max=1&knob-max-value=0&knob-ticks=1&knob-label-handles=true&knob-label-ticks=true&viewMode=story))

(The highlighted histogram range will still need to be fixed in #2914.)

### DIFF
--- a/src/components/calcite-slider/calcite-slider.tsx
+++ b/src/components/calcite-slider/calcite-slider.tsx
@@ -825,7 +825,12 @@ export class CalciteSlider {
   }
 
   private shouldUseMinValue(): boolean {
-    return this.isRange && !this.hasHistogram && this.minValue === 0;
+    if (!this.isRange) {
+      return;
+    }
+    return (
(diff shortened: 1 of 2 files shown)


src/components/calcite-slider/calcite-slider.tsx
```diff
@@ -825,7 +825,12 @@ export class CalciteSlider {
   }
 
   private shouldUseMinValue(): boolean {
-    return this.isRange && !this.hasHistogram && this.minValue === 0;
+    if (!this.isRange) {
+      return;
```
jcfranco, line 829: `return false;`

jcfranco (approved):
👍🏼

Issue #2728, opened 2021-08-04
Bug: calcite-font-weight-bold is the wrong weight


bstifle:
calcite-font-weight-bold is the wrong weight

```
   🦠 --calcite-font-weight-bold: 700🦠
```

should be:

```
   ✨ --calcite-font-weight-bold: 600✨
```

benelan:
verified on beta.66

PR #3093, opened 2021-09-22
refactor (calcite-font-weight-bold): updated weight to 600

justinhough:
**Related Issue:** #2728

## Summary

Updated `--calcite-font-weight-bold` from `700` to `600` and updated styles README to reflect that change.


caripizza (approved):
👍

PR #3095, opened 2021-09-22
chore: use github secrets for verifier usernames

benelan:
## Summary
- Created two new github secrets:
  - PRIMARY_CONTACT: a single username that will be the primary point of contact used in github actions. Right now the value is `benelan`
  - SECONDARY_CONTACTS: A comma separated list of secondary contacts. It can be as many as needed. Right now the value is `jcfranco, julio8a`
- Changed all gh actions with hardcoded usernames to use the new gh secrets
- Changed the gh action that assigns verifiers for installed issues to randomly select a secondary contact if it is the primary contact's issue

### DIFF
--- a/.github/workflows/assign-for-verification.yml
+++ b/.github/workflows/assign-for-verification.yml
@@ -7,22 +7,30 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/github-script@v4
+        env:
+          PRIMARY_CONTACT: ${{secrets.PRIMARY_CONTACT}}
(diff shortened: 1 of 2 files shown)


.github/workflows/assign-for-verification.yml
```diff
@@ -7,22 +7,30 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/github-script@v4
+        env:
+          PRIMARY_CONTACT: ${{secrets.PRIMARY_CONTACT}}
```
jcfranco, line 11: Nitpick: Is there another name besides `contact` that could express its usage?

jcfranco (approved):
Awesome!

Issue #2974, opened 2021-09-02
Docs: add missing default slots


benelan:
### Description 
The following components have undocumented default slots that need to be documented. We decided that the default slot should be the first one so I will go through all of the components to fix that in this issue as well.

### Which Component
```
calcite-accordion
calcite-accordion-item
calcite-button
calcite-checkbox
calcite-chip
calcite-combobox
calcite-combobox-item
calcite-combobox-item-group
calcite-dropdown
calcite-dropdown-group
calcite-dropdown-item
calcite-input-message
calcite-label
calcite-option-group
calcite-panel
calcite-popover
calcite-radio-button-group
calcite-radio-group
calcite-scrim
calcite-select
calcite-split-button
calcite-stepper
calcite-stepper-item
calcite-tab
calcite-tav-nav
calcite-tab-title
calcite-tabs
calcite-tile-select
calcite-tile-select-group
calcite-tip
calcite-tip-group
calcite-tooltip
```

### Resources 
**Related Issue:**: #2807

macandcheese:
Following up re: Teams chat about ‘action’ - we should also make sure the verbiage used for default slot documentation across the all components is generalized and made consistent.

benelan:
@macandcheese yeah this issue is part of an upcoming epic already planned out for syntax/verbiage consistency throughout usage, prop, and slot doc

jcfranco:
`calcite-checkbox` is missing its default slot. The rest look good! I'll create a follow-up issue for checkbox.

jcfranco:
Actually, after looking at the implementation and stories, checkbox's slot appears to be internal and should not be doc'd.

jcfranco:
@benelan Can you confirm? ☝🏼

benelan:
@jcfranco yeah I saw the slot but I don't think we want users putting anything in there so I left it out of the doc.

PR #3098, opened 2021-09-23
docs: add missing default slots

benelan:
**Related Issue:** #2974

## Summary
Added documentation for missing default slots


src/components/calcite-label/calcite-label.tsx
```diff
@@ -4,6 +4,9 @@ import { FocusRequest } from "./interfaces";
 import { Alignment, Scale, Status } from "../interfaces";
 import { CSS_UTILITY } from "../../utils/resources";
 
+/**
+ * @slot - A slot for adding text.
```
driskull, line 8: I think we should say text and a labelable component. We might need to document the labelable components here as well. thoughts @jcfranco ?

driskull (approved):
👍

src/components/calcite-label/calcite-label.tsx
benelan: @driskull do you mean a native label?

driskull: No, one of our components that supports a calcite-label.

PR #3110, opened 2021-09-25
docs: cleanup and reorganize onboarding information

benelan:
## Summary
When I was first onboarding I was a bit confused by having 4+ docs to read. There was also some conflicting information since its hard to keep multiple docs with the same info in sync. After this change we will only have Contributing and Conventions as the onboarding docs.

### Changes
- The getting started doc is already in the main readme word for word, so I removed it
- The travis webhook instructions are not something that most people need to see, so I moved it to the [release wiki](https://github.com/Esri/calcite-components/wiki/Releasing)
- The project guidelines was mostly a word for word mix between sections in the contributing doc and the conventions doc
  - I moved the non-repeat items into the contributing/convention docs according to the thought process below

### Contributing vs Conventions
- Conventions are specific to how we develop the web components
- Contributing is a more general "this our process and how to get setup" document. This includes how to log bugs, issue management, commit messages, branch names, setting up for local development, etc.


CONTRIBUTING.md
```diff
@@ -50,7 +59,16 @@ Milestones are used to manage sprints, which are two weeks long. Sprint mileston
 - **13:** Issues that are more complicated and need some workflow or design planning. These issues usually need additional unit tests written.
 - **40:** If an issue is this complicated it should be converted into an epic.
 
-### Getting a development environment set up
+## Code base
+
+Our code base is written entirely in TypeScript and we should aim to have it that way. We have configured the project to help adhere to some conventions and formatting. Here are additional items to follow:
+
+1. Avoid using any as much as possible.
```
driskull, line 66: Maybe rephrase as: "Avoid setting types as `any`."?

CONTRIBUTING.md
```diff
@@ -50,7 +59,16 @@ Milestones are used to manage sprints, which are two weeks long. Sprint mileston
 - **13:** Issues that are more complicated and need some workflow or design planning. These issues usually need additional unit tests written.
 - **40:** If an issue is this complicated it should be converted into an epic.
 
-### Getting a development environment set up
+## Code base
+
+Our code base is written entirely in TypeScript and we should aim to have it that way. We have configured the project to help adhere to some conventions and formatting. Here are additional items to follow:
```
driskull, line 64: > Our code base is written entirely in TypeScript and we should strive to keep it that way in order to benefit from type protection. We have configured the project to help adhere to some conventions and formatting. Here are additional items to follow:

?

CONTRIBUTING.md
```diff
@@ -16,13 +18,20 @@ If you aren't familiar with the basics of Web Components and Shadow DOM, please
...
 
-If something isn't working the way you expect, please take a look at the [existing issues](https://github.com/Esri/calcite-components/issues) before logging a new one. Have you found a new bug? Want to request a new feature? We'd love to hear from you! Please make sure to provide all of the requested info from the appropriate [issue template](https://github.com/Esri/calcite-components/issues/new/choose) so we can work on resolving the issue as soon as possible. A sample that reproduces the issue is required for logging bugs, we created templates in [codepen](https://codepen.io/pen?template=RwgrjEx), [codesandbox](https://codesandbox.io/s/calcite-template-p95kp?file=/src/App.js), and [jsbin](https://jsbin.com/lopumatiru/edit?html,output) (with the ArcGIS API for JavaScript) to help get started. Alternatively, a [documentation](https://developers.arcgis.com/calcite-design-system/components/) sample can be used if the issue is reproducible.
+- Use a clear and descriptive title
+- What is happening now vs what should happen?
```
driskull, line 26: Detail what is happening now vs what should happen?

driskull:
> The getting started doc is already in the main readme word for word, so I removed it

The reason for the separation was to have different "pages" on storybook. So if we combine these, we probably need to update the storybook mdx files.

https://github.com/Esri/calcite-components/blob/master/src/2-getting-started.stories.mdx

Should we just remove all these mdx files?

driskull (commented):
Comments on content. Looks good @benelan !

benelan:
> Should we just remove all these mdx files?

I think the getting started one should be removed and the conventions one should be added. Does that make sense to you @driskull?

driskull:
Yeah that makes sense.

CONTRIBUTING.md
benelan: All of the stuff in these 3 comments were taken directly from the project guidelines, e.g. https://github.com/Esri/calcite-components/blob/master/PROJECT_GUIDELINES.md#code-base

This PR is just about removing repeat info without loosing any content. Next I will make a PR to fix wording. Is that okay?

benelan: I'll make sure to fix the three things you mentioned in the next PR

driskull: Yeah thats fine 👍

driskull (approved):
👍

Issue #2112, opened 2021-05-06
Bug: calcite-tree ancestry mode - setting `selected` on nested `calcite-tree-item`s does not update lineage


stevenalbers:
Clicking on a nested tree item correctly manages the selected/indeterminate state of its parents:

<img width="400" alt="Screen Shot 2021-05-05 at 5 20 25 PM" src="https://user-images.githubusercontent.com/5317415/117369993-8f65f900-ae7a-11eb-90c3-1064ea54abb0.png">

But if you're generating a `calcite-tree` and you initialize it by setting `selected` on a nested item, its ancestry isn't being updated (In this case `Regional Deterministic Prediction System (RDPS) [10 km]` should be shown as indeterminate:

<img width="625" alt="Screen Shot 2021-05-05 at 4 51 49 PM" src="https://user-images.githubusercontent.com/5317415/117370001-90972600-ae7a-11eb-9f24-c6692b3bd3fc.png">

What's interesting is that `Regional Deterministic Prediction System (RDPS) [10 km]` knows it has selected children (the text is bold), but the checkbox didn't get the memo.

@paulcpederson

jcfranco:
@paulcpederson @caripizza Can either of you look at this for the incoming sprint?

caripizza:
@jcfranco yep I can!

benelan:
verified locally on master in the `calcite-tree-scales` demo page. @caripizza, should that page be added to the demo `index` file? If so I can do that.

benelan:
Or at least linked to in the tree demo page?

caripizza:
> should that page be added to the demo index file?

@benelan yep we can handle that in #2622 if that sounds good to you?

benelan:
@caripizza sounds great, thanks!

PR #3111, opened 2021-09-25
fix(tree-item): update `indeterminate` on ancestor tree-items when `selected` on load

caripizza:
**Related Issue:** #2112

## Summary
Fixes an issue where a tree-item with "ancestors" selection-mode would appear as selected, but its ancestor tree-items would not load with an `indeterminate` checkbox. Now ancestors will render with their indeterminate checkboxes on initial page load.

(The fix borrows the logic from calcite-tree's private [`updateAncestorTree`](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-tree/calcite-tree.tsx#L231-L236) method.)

### DIFF
--- a/src/components/calcite-tree-item/calcite-tree-item.tsx
+++ b/src/components/calcite-tree-item/calcite-tree-item.tsx
@@ -125,6 +125,10 @@ export class CalciteTreeItem {
     }
   }
 
+  componentDidLoad(): void {
+    this.updateAncestorTree();
+  }
+
   render(): VNode {
     const rtl = getElementDir(this.el) === "rtl";
     const showBulletPoint =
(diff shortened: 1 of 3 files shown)


src/components/calcite-tree-item/calcite-tree-item.tsx
```diff
@@ -125,6 +125,10 @@ export class CalciteTreeItem {
     }
   }
 
+  componentDidLoad(): void {
+    this.updateAncestorTree();
```
driskull, line 129: Should this maybe happen at connectedCallback? I think componentDidLoad is only called once per initialization.

driskull (commented):
Looks good @caripizza Just had one question

src/components/calcite-tree-item/calcite-tree-item.tsx
caripizza: Good question - I tried moving it into connectedCallback but couldn't get it working. It seems like tree-item needs to have its selectionMode set first [within componentWillRender](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-tree-item/calcite-tree-item.tsx#L112).

I went with this route since it seems like it only needs to be called once after the first render (and calcite-tree handles the indeterminate updates after that).

driskull: Ok sounds good

driskull (approved):

Issue #3059, opened 2021-09-16
Bug: [Combobox] - Incorrect display of long strings and single select value


macandcheese:
### Actual Behavior
There is a scrolling behavior that currently occurs inside a combobox input when the string is "too long". However, it appears that width is not as wide as the input.

For `selection-mode=single`, the above is not an issue but there is a gap between the border of the input and where the text is placed...

### Expected Behavior
The input should be as wide as the container allows - only becoming narrower to pad a chevron or "selected chips".
For `selection-mode=single`, the selected item should appear in the same location the search string does.

### Reproduction Steps and Sample

https://user-images.githubusercontent.com/4733155/133702023-5a27f24e-fc05-4b12-a144-d2c01523ce29.mp4

_Sample_:

1. https://jsbin.com/hibafawiyu/1/edit?html,output

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

macandcheese:
It looks like the "selection-mode=single" alignment issue is caused by the combobox inheriting the `text-align:center` from the jsbin sample code... Probably similar to https://github.com/Esri/calcite-components/pull/3039#issuecomment-920208922

I am of the opinion that for both of those instances, the combobox should not inherit either text-align:center or the font-weight in the above comment. However that would probably stop happening once `calcite-input` is transitioned to use shadowed css, correct?

benelan:
verified on master

PR #3114, opened 2021-09-27
fix(combobox): Correct display issues with single select and long strings. #3059

driskull:
**Related Issue:** #3059

## Summary

fix(combobox): Correct display issues with single select and long strings. #3059


macandcheese (approved):
Wonderful 🧹

Issue #2965, opened 2021-09-02
Bug: tab-nav z-index causes tabs to appear above active popover


COV-GIS:
### Actual Behavior

The z-index applied here https://github.com/Esri/calcite-components/blob/master/src/components/calcite-tab-nav/calcite-tab-nav.scss#L2 causes tabs to appear over active popover.

![image](https://user-images.githubusercontent.com/14297903/131864722-434a0ed6-a1d9-4a31-b32a-fda4eec1a857.png)

![image](https://user-images.githubusercontent.com/14297903/131866173-da025204-0d2e-48b3-bb64-0c6a49a39082.png)

A z-index of 10 or less fixes this issue.

### Expected Behavior

Popover should be above everything.

### Reproduction Steps or Sample

1.

### Relevant Info

_Version_: `@esri/calcite-components@1.0.0-beta.63`

- [x] CDN
- [x] NPM package

caripizza:
@COV-GIS can you move your calcite-popover to be outside of the tabs markup? I think that should fix this issue.

COV-GIS:
@caripizza I tried that but the reference-element is in the tab and doesn't work otherwise. The reference-element ended up being in a different position and with placement the popover being over the tab-nav can be avoided.

benelan:
Verified on master in the popover demo page

PR #3115, opened 2021-09-27
fix(tab, tab-nav): remove z-index styles so popover appears on top

caripizza:
**Related Issue:** #2965

## Summary
Removes the z-index styles set on calcite-tab-nav and calcite-tab, so a calcite-popover nested within a tab will appear above everything else (ie., use default stacking order).


driskull (approved):
👍 Looks good

Issue #3071, opened 2021-09-18
Bug: calcite-dropdown displays max-items correctly but adds scrolling


dhrumil83:
### Actual Behavior
beta.65 you can scroll as in the gif. If you click on the dropdown first, cannot scroll.
Note: This only occurs if you have many items in the dropdown, and the display is controlled via max-items.

![draw](https://user-images.githubusercontent.com/14285436/133895758-c269b5e2-1e2b-40b3-99f9-b055a4249f3b.gif)

### Expected Behavior
Change to beta.59 and things works as expected, and cannot scroll based on contents of the dropdown.

### Reproduction Steps and Sample
https://codepen.io/dhrumil83/pen/poJBbEW?editors=1100

### Relevant Info

_Version_: `@esri/calcite-components@65`

- [x] CDN
- [ ] NPM package

benelan:
verified in beta.66

PR #3116, opened 2021-09-27
fix(dropdown): Set scroller height to zero when not active. #3071

driskull:
**Related Issue:** #3071

## Summary

fix(dropdown): Set scroller height to zero when not active. #3071


caripizza (approved):
👍

Issue #3125, opened 2021-09-28
Bug: calcite-action alignment="center" doesn't work


eriklharper:
### Actual Behavior

```
<calcite-action
    scale="s"
    indicator
    active
    text="click-me"
    label="hello world"
    text-enabled
    icon="configure-popup"
    alignment="center"
  >
  </calcite-action>
```
Results in:

![image](https://user-images.githubusercontent.com/821864/135174878-f4e1316b-f318-4684-9474-15b64c21c857.png)

### Expected Behavior

The above code should produce this instead:

![image](https://user-images.githubusercontent.com/821864/135174908-cd82024a-e23a-43da-9d7a-a64ce1976d7d.png)

### Reproduction Steps and Sample

1. Start the dev server
2. Open `src/demos/calcite-action.html`
3. Add `alignment="center"` to the first `<calcite-action>` instance.
4. Refresh the page

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

PR #3128, opened 2021-09-29
fix(calcite-action): fixing center alignment appearance

eriklharper:
**Related Issue:** #3125

## Summary

This pull request aligns the content of the calcite-action button properly in the center when `alignment="center"` is used.


caripizza:
This looks great to me! Thanks for adding the variants to the demo page too. Only question I have is should we prevent the center alignment button from stretching full width when there's no text as well, so it matches the other alignments?
https://github.com/Esri/calcite-components/blob/master/src/components/calcite-action/calcite-action.scss#L95

eriklharper:
> This looks great to me! Thanks for adding the variants to the demo page too. Only question I have is should we prevent the center alignment button from stretching full width when there's no text as well, so it matches the other alignments? https://github.com/Esri/calcite-components/blob/master/src/components/calcite-action/calcite-action.scss#L95

Good call, it looks like `w-full` is being applied when the text is visible anyway, so this is actually a redundant rule.  I'll remove.

src/components/calcite-action/calcite-action.tsx
```diff
@@ -43,7 +43,7 @@ export class CalciteAction {
   @Prop({ reflect: true }) active = false;
 
   /**
-   * Indicates the alignment when text-enabled is false.
+   * Sets the horizontal alignment of the button's content.
```
caripizza, line 46: Nitpick: since non-text actions are square, can we clarify this targets the text specifically? Something like "Optionally specify the horizontal alignment of button elements with text content."

caripizza (approved):
LGTM 👍  Just one small nitpick about the JSDoc, otherwise seems good!

PR #3129, opened 2021-09-29
chore(action): make need info action work with a11y issues

benelan:
**Related Issue:** Discovered issue on 3123

## Summary
The a11y/bug issue template headers now match so that the `need more info` action will work for both of them. The link to the issue templates is generalized instead of directing to the bug template.


driskull (approved):
Nice!

Issue #3101, opened 2021-09-23
Bug: a range slider's minValue thumb should appear on bottom near max edge when handles collide


caripizza:
### Actual Behavior
In a `mirrored` range slider with `label-handles` ~

#### A.
When the minValue thumb is moved via keyboard or mouse towards the max edge, it's positioned **above** the maxValue thumb on the range with the maxValue thumb below (bug):

![image](https://user-images.githubusercontent.com/42423180/134575979-bb03d780-3789-409e-9405-f3962b3d0e7c.png)

#### B.
The `.hyphen` class gets added to the value handle label, instead of the minValue handle label.
The hyphenation that appears when the labeled handles collide on max edge currently targets the [`:after`](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-slider/calcite-slider.scss#L88) pseudo-element.

Here's how both the non-mirrored (left) and mirrored (right) range sliders appear which shows the UI bug:

![image](https://user-images.githubusercontent.com/42423180/134576694-ac018415-ce6d-4d11-ac74-b68f22c8713e.png)

(The above screenshots have a semi-transparent red color to articulate the minValue thumb.)

### Expected Behavior

In a `mirrored` range slider with `label-handles` ~

#### A.
When the minValue thumb is moved via keyboard or mouse towards the max edge, it's positioned **below** the maxValue thumb on the range with the maxValue thumb above (correct, ie., match non-mirrored behavior).

#### B.
The `.hyphen` class gets added to the minValue handle label.
The hyphenation that appears when the labeled handle thumbs collide on max edge targets the `:before` pseudo-element.

### Reproduction Steps and Sample

_Sample_:

1. https://codepen.io/caripizza-the-animator/pen/eYRLEXG

### Relevant Info

_Version_: `@esri/calcite-components@beta.65`

- [x] CDN
- [ ] NPM package

caripizza:
I think I might have found an easy fix for this while working on #2457, so I'm going to take a stab at it.

anveshmekala:
Currently, the hyphenation in slider is aligned vertically below the `minHandle` label which looks like a regression bug. Correct me if am wrong @asangma & @macandcheese 

Guessing it should look like below attached hyphenation for both mirrored and non-mirrored slider. Correct me if am wrong. 

[...]

PR #3137, opened 2021-09-29
fix(slider): use non-breaking hyphen for colliding handle labels

caripizza:
**Related Issue:** #3101

## Summary
Fixes an issue where a slider range with `label-handles` would wrap its minValue handle label's hyphen when both thumbs overlap near the max edge.

_Before_

image

_After_

image

### DIFF
--- a/src/components/calcite-slider/calcite-slider.scss
+++ b/src/components/calcite-slider/calcite-slider.scss
@@ -86,9 +86,9 @@ $tick-height: 4px;
       bottom: 0;
     }
     &.hyphen::after {
-      content: "\2014";
-      display: inline-block;
-      width: 1em;
+      content: "\2011";
+      display: inline;
+      margin-left: 0.1875em;
     }
   }
 


src/components/calcite-slider/calcite-slider.scss
```diff
@@ -86,9 +86,9 @@ $tick-height: 4px;
       bottom: 0;
     }
     &.hyphen::after {
-      content: "\2014";
-      display: inline-block;
-      width: 1em;
+      content: "\2011";
+      display: inline;
+      margin-left: 0.1875em;
```
driskull, line 91: Is this margin going to be affected by RTL?

caripizza: Great question. It looks like the hyphen appears in the wrong place when it's on an RTL page currently, even when `mirrored` due to the `:after` pseudo-element targeted style. 🤔 

We can fix this if we `getElementDir` inside `hyphenateCollidingRangeHandleLabels`, then add the rtl utility class when the transformed labels overlap:

```tsx
if (rtl) {
  hyphenLabel.classList.add(CSS_UTILITY.rtl)
}
```

```scss
   ...
   &:not(.calcite--rtl).hyphen::after {
     content: "\2011";
     display: inline;
     margin-left: 0.1875em;
   }
   &.calcite--rtl.hyphen::before {
     content: "\2011";
     display: inline;
     margin-left: 0.1875em;
   }
```

which produces this for me on the demo page:

image

driskull: Can you use a css logical property instead? We don't want to use getElementDir anymore

caripizza: Hm. Without the rtl check + utility class, how can I know whether the `.hyphen` class should have a `:before` or `:after` pseudo-element style?

driskull: I think if you display it as inline-block it shouldn't need the ::before.

caripizza: I'm noticing even negative numbers appear flipped in slider on an rtl page. Can we do fixes for rtl sliders down the road instead of in this pr?

caripizza:
Sounds like this isn't going to work so I'll go ahead and close this one. I'll create a new issue for the rtl tick/handle label bugs with negative numbers, then we can discuss a plan forward there instead.

caripizza:
(Issue #3167 created for the above rtl bugs I noticed here.)

PR #3148, opened 2021-10-01
docs(faq): add section about slots

benelan:
## Summary
- Added a brief explanation about slots to the faq doc
- Added the faq page to storybook since I will continue adding stuff here


driskull (approved):
Nice!

Issue #3075, opened 2021-09-20
Enhancement: Block loading state when not open


asangma:
### Description
Add a way to see if Block is `loading` when its not `open`.

### Acceptance Criteria 
* Design should use existing slots

#### Which Component
Block

#### Example Use Case
Have a consistent design for a Block that is in a loading state but is not open.

#### Proposed design
![image](https://user-images.githubusercontent.com/12503298/134081542-8a48c08d-03db-491a-b853-942c51e93fe6.png)

* Use the same slot that other statuses use.
* Loading indicator rotates
* If Block has `loading` prop, `status` is ignored.

benelan:
verified on `beta.68` in the doc
https://developers.arcgis.com/calcite-design-system/components/block/#sample

PR #3158, opened 2021-10-05
feat(block): add loading status to block header

anveshmekala:
**Related Issue:** #3075 

## Summary

This will add loading status to the header of the calcite-block when the `loading` prop is set to `:true` and  not open. Infinite animation is added to the refresh icon with 2s  duration.


driskull:
This looks good to me :)

I'll approve when its out of draft.

Issue #3146, opened 2021-10-01
Bug: first radio button always selected when clicking other buttons in a radio-button-group


jcfranco:
### Actual Behavior

Clicking a `calcite-radio-button` in a `calcite-radio-button-group` results in the first one always being selected.

### Expected Behavior

The clicked `calcite-radio-button` should be selected.

### Reproduction Steps and Sample

_Sample_: https://codepen.io/jcfranco/pen/RwgORaY

1. Click on any radio button except the first
2. Notice how the first option is always selected

### Relevant Info

_Version_: `@esri/calcite-components@1.0.0-beta.66`

- [x] CDN
- [ ] NPM package

**Workaround**: use the `for` attribute in label (see codepen for example)

driskull:
@jcfranco this might be the problem. I had this previously...

<img width="515" alt="Screen Shot 2021-10-01 at 9 48 51 AM" src="https://user-images.githubusercontent.com/1231455/135658311-2d8e4d5f-99b1-4c87-a910-5e761ff42841.png">

driskull:
Seems like it should be querying within the label instead of the rootNode.

jcfranco:
Installed.

PR #3161, opened 2021-10-06
fix(label): fix issue where clicking on a wrapped labelable component would not update its value correctly

jcfranco:
**Related Issue:** #3146 

## Summary

This PR fixes an issue that affected wrapped labelable components.

### Notable changes

* Updates `calcite-checkbox` to handle clicks on the host (otherwise, it could fail in [certain scenarios](https://codepen.io/jcfranco/pen/rNwbGNb))
* Updates label utils to avoid emitting the label click event if the user clicks directly on the labelable component – this ends up simplifying code on `LabelableComponent` implementors
* Fix logic for querying `calcite-radio-button` when wrapped vs referenced with `for`
* Updates `labelable` test util to also cover clicking on the component directly

### Next steps

* Will update `calcite-radio-group.e2e.ts` tests to use `calcite-label` to be closer to the intended usage.
* Will tidy up `LabelableComponent` implementors to no longer use bound methods, as it's no longer required

### DIFF
--- a/src/components/calcite-radio-button/calcite-radio-button.tsx
+++ b/src/components/calcite-radio-button/calcite-radio-button.tsx
@@ -182,10 +182,6 @@ export class CalciteRadioButton implements LabelableComponent {
   };
 
   private clickHandler = (): void => {
-    if (this.labelEl) {
-      return;
-    }
-
     this.toggle();
   };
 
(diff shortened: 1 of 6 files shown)


src/components/calcite-radio-button/calcite-radio-button.tsx
```diff
@@ -182,26 +182,24 @@ export class CalciteRadioButton implements LabelableComponent {
...
-      if (firstButton) {
-        firstButton.checked = true;
-        firstButton.focused = true;
+      const radioButton = label.for
+        ? this.rootNode.querySelector<HTMLCalciteRadioButtonElement>(
+            `calcite-radio-button[id="${label.for}"]`
+          )
+        : label.querySelector<HTMLCalciteRadioButtonElement>(
+            `calcite-radio-button[name="${this.name}"]`
+          );
```
eriklharper, line 198: Even though its probably safe to assume that if a `for` attribute isn't present on the label, that implies that the label contains the radio-button, but in the case that a `for` attribute is supplied to a radio that it is also wrapping (which is probably an anti-pattern I know) it is likely to break in that particular case.  I remember handling that specifically so that it worked regardless if someone did that or not, but its up to you if you think that's a case worth worrying about.

PR #3162, opened 2021-10-06
refactor(switch): Add checked property and deprecate switched property

driskull:
**Related Issue:** N/A

## Summary

refactor(switch): Add checked property and deprecate switched property

### DIFF
--- a/src/components/calcite-switch/calcite-switch.tsx
+++ b/src/components/calcite-switch/calcite-switch.tsx
@@ -64,12 +64,21 @@ export class CalciteSwitch implements LabelableComponent {
   /** The scale of the switch */
   @Prop({ reflect: true }) scale: Scale = "m";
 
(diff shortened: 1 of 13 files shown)


src/components/calcite-switch/calcite-switch.tsx
```diff
@@ -64,12 +64,21 @@ export class CalciteSwitch implements LabelableComponent {
...
   @Prop({ reflect: true }) scale: Scale = "m";
 
+  /** True if the switch is initially on
+   * @deprecated use 'checked' instead.
+   */
+  @Prop() switched = false;
+
+  switchedWatcher(switched: boolean): void {
+    this.checked = switched;
+  }
```
eriklharper, line 74: Where is `switchedWatcher` called without the `@Watch` decorating it?

PR #3178, opened 2021-10-08
refactor(filter): avoid stateful regexp used for filtering

jcfranco:
**Related Issue:** N/A

## Summary

Removes the global flag to keep the regexp stateless.

✨ 🧹 ✨


jcfranco:
@benelan Gotcha. You are correct that we are not using `lastIndex` anywhere, but for testing patterns we want to avoid the global flag unless necessary. Otherwise, it may mess up subsequent matching: https://codepen.io/jcfranco/pen/yLoBWeg?editors=0010

benelan:
Ah gotcha, thanks for the explanation!

Issue #3144, opened 2021-09-30
Bug: combobox-item height should match dropdown-item height (24/32/44)


caripizza:
### Actual Behavior
calcite-combobox-item non-wrapping text heights are 32/40/52 but should match calcite-dropdown-item heights and be 24px for small, 32px for medium, and 44px high for large scale items.

### Expected Behavior
A small combobox-item total height is 24px, a medium total height is 32px, and a large total item height is 44px.

### Reproduction Steps and Sample
Each scale of combobox-item with dropdown-item beside it is illustrated here, showing the height diffs: https://codepen.io/caripizza-the-animator/pen/oNwVPrx

_Sample_:

1. `npm start` on master > calcite-combobox.html > notice the item heights are larger than 24/32/44

### Relevant Info

_Version_: `@esri/calcite-components@1.0.0-beta.66`

- [x] CDN
- [ ] NPM package

caripizza:
@macandcheese should we add combobox-item to the "Combobox" Figma page so it matches Dropdown?

macandcheese:
Yeah, there should probably be versions in Figma of the various indentation levels and states for combobox items.

PR #3197, opened 2021-10-11
fix(combox-item): adjust height of the item to be consistent respective to its scale size (#3144)

y0n4:
**Related Issue:** #3144 

## Summary
The combobox-item total height is now 24px, 32px, 44px by the scale size s, m, l respectively.


caripizza:
> 28, 32, 44px

24 right? 😉

Issue #2153, opened 2021-05-14
Bug: `calcite-link` not linked if `href` defined later


nwhittaker:
### Actual Behavior
If a `<calcite-link>` element's `href` attribute is not immediately known (i.e. waiting on resolving the link through a URL shortening service like Bitly), setting the `href` value later has no effect.

### Expected Behavior
After setting the link's `href` attribute, clicking the link works as expected.

### Reproduction Steps or Sample
```html
<calcite-link id="link" href="">link</calcite-link>
<button onclick="document.getElementById('link').href = 'https://www.esri.com';">Set href</button>
```

1. Run the above snippet.
2. Wait ~1sec, then click the _Set href_ button.
3. Click the _link_ text.

### Relevant Info
_Version_: `@esri/calcite-components@1.0.0-next.192`

PR #3222, opened 2021-10-13
fix(link): Setting the href property after init should update rendering

driskull:
**Related Issue:** #2153

## Summary

fix(link): Setting the href property after init should update rendering. #2153


jcfranco:
Same as #3248, can you open a follow-up refactor issue to get rid of `childElType`? I think we can simplify the component by computing it in `render` and checking `type` & `href` in the click handler.

PR #3228, opened 2021-10-13
refactor: use the Sass migrator to automatically update your stylesheets to use math.div() and list.slash()

driskull:
## Summary

refactor: use the Sass migrator to automatically update your stylesheets to use math.div() and list.slash()


driskull:
This part is kind of tricky...

```
[ ERROR ]  sass error: src/components/calcite-loader/calcite-loader.scss:1:88
           @use rules must be written before any other rules.

      L1:  @use "sass:math"; 
```

Issue #3223, opened 2021-10-13
Enhancement: [calcite-tree-item] limit padding-y CSS var to top/bottom spacing, remove end indent


caripizza:
### Description
Pr #3001 refactored tree-item to match latest Figma designs. In doing so, I repurposed a `--calcite-tree-padding-y` CSS var for the item's [margin-inline](https://github.com/Esri/calcite-components/pull/3001/files#diff-2e52adbf1fa3a525a8c3f837ff0826ffe6e472e9fac455b37fb21d50e68b0bebR114) styles as well. However, it came up recently with the Dashboards team that they'd like to fine-tune the top/bottom item spacing independently of the left/right spacing. cc @ellenupp

This issue should:
- refactor tree-item to derive it's margin-inline styles from the TW config instead of the padding-y var, so consumers can set the padding-y CSS var value to something different without affecting the left/right margins on a tree-item.
- remove the margin-inline-end style from the children-container's tree-items.

### Acceptance Criteria 
a ) Updating the `--calcite-tree-padding-y` CSS var value on calcite-tree-item only changes its top/bottom spacing
b) There are no more margin-inline-end styles (end-indentation) on tree-items

### Relevant Info 
n/a

#### Which Component
calcite-tree-item

#### Example Use Case
See issues #2210 and #3127 for more details on the related designs driving this request.

benelan:
@macandcheese can you please help me verify this installed design issue?

macandcheese:
LGTM although I don't believe that css variable should be customizable, teams can just adjust padding themselves with divs they slot. Oh well, it's fine for now and we can re-visit when these get refactored with the list-item slot paradigm.

PR #3237, opened 2021-10-13
refactor(tree-item): limit padding-y CSS var to top/bottom spacing, remove end indentation

caripizza:
**Related Issue:** #3223

## Summary
Refactors calcite-tree-item styles to only apply the `--calcite-tree-padding-y` CSS var to top/bottom spacing.

Also adds the following stylesheet changes:
- removes the margin-inline-end styles set on tree-items (no more end indentation, only start)
- refactors out CSS vars that should not be exposed for style overrides on the host


src/components/calcite-tree-item/calcite-tree-item.e2e.ts
```diff
@@ -174,4 +174,26 @@ describe("calcite-tree-item", () => {
       ancestors.forEach(async (node) => expect(await node.getProperty("indeterminate")).toBe(true));
     });
   });
+
+  describe("CSS Custom Properties for style overrides", () => {
```
jcfranco, line 178: This is great. I think we could add a test util for this (outside of this PR). Ideally a spec test since we don't need interaction.

jcfranco (approved):
Code LGTM! 

I'll defer to @macandcheese on adding `--calcite-tree-padding-y`.

macandcheese:
Should this css variable be internal? I don't see a reason for it to be public.

caripizza:
@jcfranco can you please approve Screener changes on this one?

Issue #3134, opened 2021-09-29
Bug: [calcite-tree] Visibility of child elements can be out of sync with parent's expanded state


ellenupp:
### Actual Behavior

In some scenarios the visibility of a child element can be out of sync with its parent's expanded state, i.e. the parent is set to expanded but the child is hidden. I've encountered this is in two scenarios:
1. When a tree item is set to expanded by default and the tree is hidden and re-rendered
2. When an item is dynamically added to the tree while it is in an expanded state

This was first encountered in EmberJS but the second scenario can easily be reproduced in vanilla JS. 

### Expected Behavior

Whenever a tree item is expanded, all its children should be visible. 

### Reproduction Steps and Sample

_Sample_: Conditionally rendered tree in EmberJS: https://codepen.io/ellupp/pen/LYLqXao?editors=1010

1. Click "Hide tree" to hide the tree
2. Click "Show tree" to render the tree again. Notice that the first item is expanded but its children are not visible. 

![image](https://user-images.githubusercontent.com/14304581/135325799-8e4b35be-f2ff-4851-9551-dd58f2a7eaca.png)
Initial state

![image](https://user-images.githubusercontent.com/14304581/135325837-a601c4ef-cb7e-4c2d-afe9-79cedd452b80.png)
After re-render

_Sample 2_: Dynamically adding item to tree: https://codepen.io/ellupp/pen/powGqOq?editors=1010

1. Click "Add item to tree". Note that the tree item is created but is not visible. It has the attribute `calcite-hydrated-hidden`. When collapsing and re-expanding "Element 1" all children will show up as expected. 

![image](https://user-images.githubusercontent.com/14304581/135326284-81357a3a-1a4e-4fa8-b64c-301cb13a3c45.png)

### Relevant Info

_Version_: `@esri/calcite-components@beta.65`

- [x] CDN
- [x] NPM package

cc @vijayendranj @julio8a

caripizza:
@ellenupp can you help me replicate Sample 1 in vanilla js? I can see the issue clearly in Sample 2, but it's less clear to me in that first sample you provided.

ellenupp:
Hi @caripizza, I'm not sure how to accomplish that in vanilla js since you can't do conditional rendering like that. The closest thing I could think to try is removing the tree from the dom then creating and adding it back again but that works fine...

caripizza:
Gotcha, yep I noticed the same behavior as well. Thanks @ellenupp

benelan:
[...]

PR #3266, opened 2021-10-15
fix(tree-item): ensure items are visible when appended to expanded parent item

caripizza:
**Related Issue:** #3134

## Summary
Ensures that newly appended tree-items are visible when they're appended into an expanded parent tree-item.


src/components/calcite-tree-item/calcite-tree-item.e2e.ts
```diff
@@ -174,4 +174,37 @@ describe("calcite-tree-item", () => {
...
+        <calcite-tree>
+          <calcite-tree-item expanded><span>Element 1</span>
+            <calcite-tree slot='children' id='target-tree'>
+              <calcite-tree-item>Child 1</calcite-tree-item>
+            </calcite-tree>
+          </calcite-tree-item>
+          <calcite-tree-item>Element 2</calcite-tree-item>
+        </calcite-tree>
+      </calcite-panel>`);
+      await page.addScriptTag({
```
jcfranco, line 192: Can you use `page.evaluate` instead of adding a script tag?

src/components/calcite-tree-item/calcite-tree-item.tsx
```diff
@@ -50,16 +50,22 @@ export class CalciteTreeItem {
 
   @Watch("expanded")
   expandedHandler(newValue: boolean): void {
-    const items = getSlotted<HTMLCalciteTreeItemElement>(this.el, "children", {
-      all: true,
-      selector: "calcite-tree-item"
-    });
+    // handle newly appended items on parents
+    if (!newValue && this.parentTreeItem) {
+      const { expanded } = this.parentTreeItem;
```
jcfranco, line 55: Does this logic apply outside of newly appended tree items? If not, maybe this could be moved into `connectedCallback` to clearly separate this.

src/components/calcite-tree-item/calcite-tree-item.tsx
```diff
@@ -50,16 +50,22 @@ export class CalciteTreeItem {
...
+      });
 
-    items.forEach((item) => (item.parentExpanded = newValue));
+      items.forEach((item) => (item.parentExpanded = newValue));
+    }
   }
 
   /** @internal Is the parent of this item expanded? */
-  @Prop() parentExpanded = false;
+  @Prop({ mutable: true }) parentExpanded = false;
```
jcfranco, line 68: Sidebar: Not for this PR, but I think it'd be cleaner to have each tree item observe their children and update this property directly instead of having children be aware of their parents.

jcfranco (approved):
Aside from a couple of comments, this LGTM! 👍🏼

src/components/calcite-tree-item/calcite-tree-item.tsx
caripizza: Good call Franco, I'll take this approach.

Issue #3029, opened 2021-09-13
Bug: calcite-list-item styles is overlapping with panel


ffaubry:
### Actual Behavior

![image](https://user-images.githubusercontent.com/3506166/133157043-e5935676-455e-4641-a87d-c10f15989d1a.png)

As you can see the action-end styling is overlapping with the panel focus.

https://codepen.io/faubry/pen/JjJyXWz?editors=100

### Expected Behavior

The panel blue line for the focus is not overlapped.

### Reproduction Steps or Sample

1. Create a panel
2. Add a list
3. Add two list items with actions-end actions
4. set focus on the panel
5. check the ficus line of the panel around the actions

Issue #1359, opened 2020-12-04
Enhancement: Only show focus outlines on keyboard interaction


patrickarlt:
### Description

Most Calcite components show focus rings when clicked such as in the following case with `<calcite-button>`:

![image](https://user-images.githubusercontent.com/378557/101210509-7090db80-362a-11eb-9cd9-d3450a41846c.png)

These focus states are almost entirely unnecessary for mouse interactions and are a common complaint of people I see using Calcite Components. With Edge using Chromium now almost all browsers except Safari support `:focus-within` which only triggers on keyboard interactions so you can do this https://codepen.io/patrickarlt/pen/poEgXmv?editors=0100

```css
/* enable a default focus state for legacy browsers */
button:focus {
  outline: 2px solid hotpink;
  outline-offset: 2px;
}

/* 
 * disable the focus if it is has focus but NOT focus-visible
 * non supporting browsers will ignore this rule because they
 * don't understand focus-visible
 */
button:focus:not(:focus-visible) {
  outline: none;
}

/* now add a focus + focus visible state for keyboard focus only */
button:focus:focus-visible {
  outline: 2px solid hotpink;
  outline-offset: 2px;
}
```

Although Safari doesn't support `focus-within` for whatever reason it also doesn't seem to apply focus styles on pointer interactions by default.

The CodePen demo also shows that elements given programmatic focus still show styles with `focus-within`.

### Acceptance Criteria 

Only show focus styles on programmatic focus for explicit keyboard interaction

### Relevant Info 

* CanIUse: https://caniuse.com/css-focus-visible
* CodePen Demo: https://codepen.io/patrickarlt/pen/poEgXmv
* Google Dev article: https://blog.chromium.org/2020/09/giving-users-and-developers-more.html

#### Which Component

All components with `:focus` styles.

#### Example Use Case

![image](https://user-images.githubusercontent.com/378557/101210509-7090db80-362a-11eb-9cd9-d3450a41846c.png)

patrickarlt:
I thought IE 11 supported `@supports` but it doesn't https://caniuse.com/?search=%40supports. You can still do this like:

```css

button:focus {
  outline: 2px solid hotpink;
  outline-offset: 2px;
}

@supports (:focus-within) {
  button:focus {
    outline: none;
  }
  button:focus-within {
    outline: 2px solid hotpink;
    outline-offset: 2px;
  }
}
```

patrickarlt:
[...]

PR #3294, opened 2021-10-18
fix(panel): outline no longer overlap with action button

anveshmekala:
**Related Issue:** #3029 

## Summary

This fix will make sure the `outline` of `calcite-panel` don't overlap with action button of `calcite-list-item`

### DIFF
--- a/src/components/calcite-action/calcite-action.scss
+++ b/src/components/calcite-action/calcite-action.scss
@@ -10,14 +10,13 @@
 .button {
   @apply bg-foreground-1
     border-none
-    box-border
     cursor-pointer
     fill-color-3
     flex
     focus-base
     items-center
     justify-start
-    m-1
+    m-0
     relative
     text--2h
     font-medium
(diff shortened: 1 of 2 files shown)


caripizza:
@anveshmekala I don't think we should modify action styles for this fix.

Since calcite-panel's `section` is keyboard-navigable (ie., has `tabindex="0"` set on it [here](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-panel/calcite-panel.tsx#L394)), we just need to add some corresponding focus styles [to it](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-panel/calcite-panel.scss#L161), along with maybe a small margin to accommodate a focus offset. (Both list-item and action have inset focus styles, so we'll need to have an outline that doesn't interfere with their inset outlines.)

Here's an example of an approach you can take within that class in panel's stylesheet to a) apply base transparent focus color to the section and b) an outset focus style when it's in focus:
```scss
  ...
  flex-col
  focus-base
  m-1;
  &:focus-visible,
  &:focus {
    @apply focus-outset;
  }
```

@asangma @macandcheese do you guys have any opinion on how the panel `section`'s outline needs to appear when it's tabbed to and in focus? (Related #1359.)

Issue #3142, opened 2021-09-30
Bug: [slider] odd slider value range


jcfranco:
### Actual Behavior

Configuring a slider the following way will show the highest value tick outside of the max-allowed range:

```html
  <calcite-slider min="5"
                  max="100"
                  step="10"
                  ticks="10"
                  label-handles
                  label-ticks
                  snap></calcite-slider>
```

![Screen Shot 2021-09-30 at 10 50 55 AM](https://user-images.githubusercontent.com/197440/135506051-fe2642d2-7adc-4716-b3e8-8ef51373851f.png)

### Expected Behavior

Max tick value is capped at the specified value (100).

### Reproduction Steps and Sample

_Sample_: https://codepen.io/jcfranco/pen/zYzbWPp?editors=1000

1. Load the sample above and notice the last tick value (see screenshot above)

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [x] CDN
- [ ] NPM package

caripizza:
This bug can also be reproduced when using a decimal for ticks like "0.1"; it overlays an additional tick on top of the max tick: ([Storybook iframe url](https://esri.github.io/calcite-components/iframe.html?id=components-controls-slider--range&args=&knob-min=0&knob-min-label=Temperature,%20lower%20bound&knob-min-value=0&knob-max=1&knob-max-label=Temperature,%20upper%20bound&knob-max-value=0&knob-step=0.1&knob-ticks=0.1&knob-snap=true&knob-label-ticks=true&knob-label-handles=true&viewMode=story))

If you inspect the element, you can see that the last 2 ticks get the same offset value:

![image](https://user-images.githubusercontent.com/42423180/135507061-106e9ce2-c7e3-46c5-84d0-af10049427aa.png)

Based on testing, it seems like the while loop in `generateTickValues` needs some modifications. (See https://github.com/Esri/calcite-components/issues/2928#issuecomment-926181759)

anveshmekala:
After this fix , the labels might overlap on smaller screens . We might need to move the max value tick label to the `right ` by 1% so that `labels` won't overlap. Any thoughts?

 @jcfranco @caripizza 
 
![227BAD91-C8B6-469A-9563-FC891A9E5FD1_4_5005_c](https://user-images.githubusercontent.com/88453586/137998553-d765da1e-4bc4-439d-b43a-0e907d37ac9c.jpeg)
[The above image is simulating ipad screen ]

[...]

PR #3302, opened 2021-10-19
test(slider): add e2e test to verify maxValue capped to max prop

anveshmekala:
**Related Issue:** #3142

## Summary

This PR adds an `e2e` test to verify if the max value of the` calcite-slider` is capped to the provided `max` prop value

### DIFF
--- a/src/components/calcite-slider/calcite-slider.e2e.ts
+++ b/src/components/calcite-slider/calcite-slider.e2e.ts
@@ -151,7 +151,7 @@ describe("calcite-slider", () => {
     expect(ticks.length).toBe(11);
   });
 
-  it("max value should be capped to the provided max prop #3142", async () => {
+  it("should cap the rendered last tick label to the slider's provided max #3142", async () => {
     const page = await newE2EPage();
     await page.setContent(`
       <calcite-slider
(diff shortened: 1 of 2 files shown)


caripizza (changes requested):
@anveshmekala thanks for trying to tackle the root cause. But for now, I think this pr should only be a test since that is the pr title and it was specifically requested here: https://github.com/Esri/calcite-components/pull/3262#issuecomment-946072101

Can you please assert the logic you added in your prior pr on this one?

Then, open a new GitHub issue + pr for the additional "root cause" fix you're adding here. Please let me know if you have questions about this approach.

anveshmekala:
> @anveshmekala thanks for trying to tackle the root cause. But for now, I think this pr should only be a test since that is the pr title and it was specifically requested here: [#3262 (comment)](https://github.com/Esri/calcite-components/pull/3262#issuecomment-946072101)
> 
> Can you please assert the logic you added in your prior pr on this one?
> 
> Then, open a new GitHub issue + pr for the additional "root cause" fix you're adding here. Please let me know if you have questions about this approach.
@caripizza  I think we should open a new issue to fix the bug introduced with the PR #3262 which doesn't append the classname `.thumb--maxValue`. It can be solved by fixing the root cause . We can close this PR and have the test added in the new issue. Any thoughts on this approach ?

caripizza:
@anveshmekala it's not the thumb class that fails to render, it's the max tick label class.

On this pr, you can use the original selector in the e2e test from my first comment before I edited it (`const maxTickLabel = await page.find("calcite-slider >>> .tick:nth-of-type(11)");`.)

You can give the new issue a title like "Bug: [calcite-slider]: last tick value generated can be greater than slider max" (or something similar). That issue will document the behavior you noticed here and propose a possible fix.

src/components/calcite-slider/calcite-slider.e2e.ts
```diff
@@ -151,6 +151,24 @@ describe("calcite-slider", () => {
     expect(ticks.length).toBe(11);
   });
 
+  it("should cap the rendered last tick label to the slider's provided max #3142", async () => {
```
caripizza, line 154: nitpick: can you remove the issue number from the test title?

caripizza (approved):
Thanks @anveshmekala, just a tiny last comment 👍

Issue #3186, opened 2021-10-08
Bug: block header and block content x-spacing don't match


bpatterson88:
### Actual Behavior
![Screen Shot 2021-10-08 at 2 39 32 PM](https://user-images.githubusercontent.com/15875886/136628175-7f3684d7-e7ec-4159-a46d-b4c70c5b3386.png)

### Expected Behavior

The left edge of the header label should align with the left edge of the labels below

### Reproduction Steps and Sample

Codepen: https://codepen.io/benpatterson/pen/abyoNrx

### Relevant Info

_Version_: `@esri/calcite-components@1.0.0-beta.63`

- [ ] CDN
- [ ] NPM package

lejara:
hello,
Its cool if i can work on this?

benelan:
Sure, we appreciate the help! Can you quickly read through the [contributing documentation](https://github.com/Esri/calcite-components/blob/master/CONTRIBUTING.md) first? For now you can fork the repo and create a PR. The checks on the PR won't pass unless you use conventional commits as [mentioned here](https://github.com/Esri/calcite-components/blob/master/CONTRIBUTING.md#commit-message-format) Thanks!

PR #3316, opened 2021-10-20
fix(block): block title and content now match on the x-axis #3186

jcfranco:
**Related Issue:** #3186

## Summary
The left and right padding of the `.title` class now matches the `.content` class

cc @lejara

### DIFF
--- a/src/components/calcite-block/calcite-block.scss
+++ b/src/components/calcite-block/calcite-block.scss
@@ -73,7 +73,7 @@ calcite-handle {
 }
 
 .title {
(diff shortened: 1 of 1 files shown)


jcfranco:
@caripizza @macandcheese Could either of you take a look at the Screener changes? Thanks!

lejara:
hello! @caripizza

Trying to understand what you see;
`(Screener shows the chevron is now not aligned with the block title)`
I am having trouble seeing how the chevron is not aligned with the block title. 
Could you please provide an example?

Issue #3249, opened 2021-10-14
Refactor: Migrate Sass to be 2.0.0-friendly


jcfranco:
### Description 

The latest dependency bump led to the following build warnings:

```cli
DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.
Recommendation: math.div(-$pointer-size, 2)
```

We should migrate our style files to be Sass 2.0.0 ready. The following are some changes that will be needed:

* [ ] use `math.div` instead of `/`
* [ ] use `@use` (and `@forward`) instead of `@import`

### Proposed Advantages 

Prevent moving to the latest dependencies when they move to Sass 2.0.0.

### Relevant Info 

Sass provides a migrating tool to help with this: https://sass-lang.com/documentation/cli/migrator

Dependencies that use Sass will likely also need to be updated (e.g., [`calcite-base`](https://github.com/Esri/calcite-base)).

#### Which Component

All style files that meet the aforementioned change criteria.

jcfranco:
1\. was mitigated by #3268 and 2\. will be addressed by #10583 and #10585.

Closing based on ☝️. cc @geospatialem @DitwanP

PR #3320, opened 2021-10-20
refactor(slider, loader): prevent Sass deprecation warnings from being thrown on start

caripizza:
**Related Issue:** n/a (the root of the actual issue is tied to #3249)

## Summary
Removes usage of the calcite-base `font-size` mixin in calcite-slider, and adds the `calc` CSS function to the calcite-loader mixin so the Sass deprecation warnings no longer appear on `npm start`.


jcfranco (approved):
👍

Issue #3344, opened 2021-10-25
Bug: Can't type in input with slotted action


mpayson:
### Actual Behavior

When clicking on an input with a slotted action to type a value, the focus automatically goes to the action, preventing text from being entered.

### Expected Behavior

The focus stays on the input so the user can type.

### Reproduction Steps and Sample

[Current documentation](https://developers.arcgis.com/calcite-design-system/components/input/) under "With slotted action".

### Relevant Info

_Browser_: Firefox
_Reproducible version_: `@esri/calcite-components@1.0.0-beta.67`

- [x] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-beta.66`

jcfranco:
Installed.

benelan:
verified on beta.69

PR #3350, opened 2021-10-25
fix(input): resolve focus issue that prevents typing into wrapped input with slotted action

caripizza:
**Related Issue:** #3344

## Summary
Adds 2 failing input tests for debugging the bug fix in this pr - WIP, opening this up per Teams convo with Franco so we can fix the issue via an alternative to calling `event.stopPropagation()`.


jcfranco:
Thanks for pushing the PR, @caripizza. 💪

I added a spec test on the label util in #3424 to cover the problematic scenario instead.

Issue #1475, opened 2021-01-26
Enhancement: l10n support for input-time-picker and time-picker components


eriklharper:
### Description

Adds locale-specific formatting to `calcite-time-picker` and `calcite-input-time-picker` based on `Intl.DateTimeFormat`.

### Acceptance Criteria 

Formats time appropriately based on locale in both the popup time-picker UI and in the corresponding input.

### Relevant Info 

Consult `calcite-date` for implementation details, which will be similar.

#### Which Component

`calcite-time-picker`, `calcite-input-time-picker`

### Resources

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/formatToParts
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat

AdelheidF:
From Erik

It looks promising to see that the Intl.DateTimeFormat api supports locale formatting through this similar formatToParts API that I used to add number locale support https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/formatToParts so it looks like we would do something quite similar for time formats as well.  We would simply extract the time components based on the provided locale and arrange the components as appropriate in the UI.

julio8a:
@babakbolour can you take a look at this one

babakbolour:
@julio8a ok. will have @raje9276 look at this and comment back.

raje9276:
I18n Considerations
1. 12hour vs 24 hour clock. In some locales default is 24 hour clock
2. AM/PM should be translated in some locales
3. AM/PM location may be after or before the time. Example: 10.00 **PM** can be written as **오후** 10시
4. Native numbers should be considered for Arabic.
5.

jcfranco:
Adjusting milestone to install this (breaking change) after the upcoming app release.

babakbolour:
@jcfranco just wondering why the bot is labeling this a stale, if there is a milestone (of 126/ - 12/7) assigned to it?

benelan:
The bot marks issues as stale when there isn't any activity for a month. However, you're right this issue shouldn't be marked as stale since it has the `in development` label. I'll fix the github action to exclude issues that are in development.

AdelheidF:
According to Olga the current format of ss:mm:hh is not correct for Arabic

[...]

PR #3354, opened 2021-10-26
feat(input-time-picker)!: localization support for input-time-picker

eriklharper:
**Related Issue:** #1475 

## Summary

Adds support for localization in both the `input-time-picker` and `time-picker` components through the `lang` attribute.  

BREAKING CHANGE: `lang` replaces the now deprecated `hour-display-format` prop.


src/components/calcite-input-time-picker/calcite-input-time-picker.tsx
```diff
@@ -89,6 +80,14 @@ export class CalciteInputTimePicker implements LabelableComponent {
   /** aria-label for the second up button */
   @Prop() intlSecondUp?: string;
 
+  /** BCP 47 language tag for desired language and country format */
+  @Prop() lang: string = document.documentElement.lang || navigator.language;
```
driskull, line 84: I don't think we should set this prop by default. We can have a `@State` variable that looks up the lang if this isn't set.

driskull (commented):
Comment on lang

src/components/calcite-input-time-picker/calcite-input-time-picker.tsx
eriklharper: Agreed.  This gives a Stencil compile warning anyway because its already defined on the prototype.  My approach I think at this point is to convert `locale` into a State variable that syncs with the `lang` attribute.

src/components/calcite-input-time-picker/calcite-input-time-picker.tsx
```diff
@@ -81,6 +80,15 @@ export class CalciteInputTimePicker implements LabelableComponent {
   /** aria-label for the second up button */
   @Prop() intlSecondUp?: string;
 
+  /** BCP 47 language tag for desired language and country format */
+  @Prop({ attribute: "lang", mutable: true }) locale: string =
```
driskull, line 84: I'm not sure that this is the right approach either because now the property and name won't match at all. Seems like this might just be a way to skirt around the warning/error. I think we should consult with Stencil on the best approach for listening to changes to native attributes on a component.

> `document.documentElement.lang || navigator.language || "en";`

I think it would be best if we don't set a default for this property. We can fallback to querying for the closest lang => navigator.language => "en".

eriklharper: > I think we should consult with Stencil on the best approach for listening to changes to native attributes on a component.

I did just that.  Here's the [Stencil Slack thread](https://stencil-worldwide.slack.com/archives/C79EANFL7/p1636133062272300):

image

src/utils/time.ts
```diff
@@ -12,45 +22,93 @@ export interface Time {
...
+  return hourAsNumber >= 0 && hourAsNumber <= 11 ? "AM" : "PM";
+}
+
+export function isValidTime(value: string): boolean {
+  if (!value || value.startsWith(":") || value.endsWith(":")) {
+    return false;
+  }
   const splitValue = value.split(":");
-  if (splitValue.length > 1) {
+  if (splitValue.length > 1 && splitValue.length < 4) {
```
jcfranco, line 111: Suggestion: For readability, consider inverting the condition so that the shorter block is first. In this case, this also allows us to reduce nesting. You can also return the value of the last conditional instead of having it fall through:

```ts
const validLength = splitValue.length > 1 && splitValue.length < 4;
if (!validLength) {
  return false;
}

const [hour, minute, second] = splitValue;
// ...
return (hourValid && minuteValid && !second) || (hourValid && minuteValid && secondValid);
```

src/utils/time.ts
```diff
@@ -59,17 +117,95 @@ export function formatTimeString(value: string): string {
...
+
+function isValidTimePart(value: string, part: TimePart): boolean {
+  if (part === "meridiem") {
+    return value === "AM" || value === "PM";
+  }
+  if (!isValidNumber(value)) {
+    return false;
+  }
+  const valueAsNumber = Number(value);
+  switch (part) {
```
jcfranco, line 134: Would an if statement be simpler?

```ts
if (part === "hour") {
  return valueAsNumber >= 0 && valueAsNumber < 24;
} else {
  return valueAsNumber >= 0 && valueAsNumber < 60;
}
```

src/utils/time.ts
```diff
@@ -59,17 +117,95 @@ export function formatTimeString(value: string): string {
...
+export function localizeTimePart(value: string, part: TimePart, locale: string): string {
+  if (!isValidTimePart(value, part)) {
+    return;
+  }
+  const date = new Date(
+    Date.UTC(
+      0,
+      0,
+      0,
+      part === "hour" ? parseInt(value) : part === "meridiem" ? (value === "AM" ? 0 : 12) : 0,
```
jcfranco, line 152: Can you store the computed value and reuse?

src/components/calcite-time-picker/calcite-time-picker.tsx
```diff
@@ -17,18 +16,25 @@ import { getKey, isActivationKey, numberKeys } from "../../utils/key";
...
+  isValidTime,
+  localizeTimeStringToParts,
+  parseTimeString,
+  localizeTimePart,
+  Meridiem,
+  getLocaleHourCycle
 } from "../../utils/time";
 import { CSS, TEXT } from "./resources";
 
+function capitalize(str: string): string {
```
jcfranco, line 34: Not for this PR, but this would be good util for a `string` util.

src/components/calcite-time-picker/calcite-time-picker.tsx
```diff
@@ -280,14 +298,24 @@ export class CalciteTimePicker {
   //
   // --------------------------------------------------------------------------
 
+  private buttonActivated(event: KeyboardEvent): boolean {
+    const key = getKey(event.key);
+
+    if (key === " ") {
```
jcfranco, line 304: Nitpick: If it's an activation key, I think you can call `preventDefault` regardless of the key.

src/components/calcite-time-picker/calcite-time-picker.tsx
```diff
@@ -570,47 +570,90 @@ export class CalciteTimePicker {
...
         }
-        break;
+        this.localizedHour = localizeTimePart(this.hour, "hour", this.locale);
+      }
+    } else {
+      this[key] = typeof value === "number" ? formatTimePart(value) : value;
+      this[`localized${capitalize(key)}`] = localizeTimePart(this[key], key, this.locale);
+    }
+    if (this.hour && this.minute) {
+      this.value =
```
jcfranco, line 645: Nitpick: to DRY this up a bit you could move the conditional that affects seconds within its corresponding expression:

```ts
const showSeconds = this.second && this.showSecond;
this.value = `${this.hour}:${this.minute}:${showSeconds ? this.second : "00"}`;
```

src/components/calcite-input-time-picker/calcite-input-time-picker.tsx
jcfranco: Can we use `@internal` to avoid documenting the prop?

jcfranco (changes requested):
This is looking great. Nicely job, @eriklharper! 😎

src/components/calcite-input-time-picker/calcite-input-time-picker.e2e.ts
```diff
@@ -326,4 +212,13 @@ describe("calcite-input-time-picker", () => {
...
     expect(await inputTimePicker.getProperty("value")).toBe("14:59");
   });
+
+  it("sets initial value to undefined when it is not a valid time value", async () => {
+    const page = await newE2EPage({
+      html: `<calcite-input-time-picker value="invalid"></calcite-input-time-picker>`
+    });
+    const inputTimePicker = await page.find("calcite-input-time-picker");
+
+    expect(await inputTimePicker.getProperty("value")).toBeUndefined();
```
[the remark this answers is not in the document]
eriklharper: Looks like when you clear the input after inputting a valid value, it sets it to an empty string:

image

src/components/calcite-input-time-picker/calcite-input-time-picker.tsx
```diff
@@ -220,24 +233,23 @@ export class CalciteInputTimePicker implements LabelableComponent {
     this.setFocus();
   }
 
+  private shouldIncludeSeconds = (): boolean => {
```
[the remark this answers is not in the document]
eriklharper: I'm confused, isn't it already a class method?

src/components/calcite-input-time-picker/calcite-input-time-picker.tsx
```diff
@@ -296,15 +315,13 @@ export class CalciteInputTimePicker implements LabelableComponent {
...
+      this.setValue({ value: isValidTime(this.value) ? this.value : undefined, origin: "loading" });
     }
     connectLabel(this);
   }
 
   componentDidLoad() {
-    if (this.calciteInputEl.value !== this.value) {
-      this.setInputValue(this.value);
-    }
+    this.setInputValue(this.localizedValue);
```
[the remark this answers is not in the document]
eriklharper: I think so.  This behavior of having to set the input value will likely change completely as a result of https://github.com/Esri/calcite-components/issues/2709 which will move away from a free-form text input in favor of input masking the time value.

src/components/calcite-tile-select/calcite-tile-select.tsx
```diff
@@ -145,7 +145,8 @@ export class CalciteTileSelect {
   }
 
   @Listen("calciteInternalCheckboxFocus")
-  checkboxFocusHandler(event: CustomEvent): void {
+  @Listen("calciteInternalCheckboxBlur")
+  checkboxFocusBlurHandler(event: CustomEvent): void {
```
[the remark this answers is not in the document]
eriklharper: #3516

driskull:
Seems like we should implement a DOM utility from this Gist to be able to watch native attributes. What do you guys think?

https://gist.github.com/willmartian/b4dd6b57d71dd0438fb9e7c6f4048578

src/components/calcite-time-picker/calcite-time-picker.e2e.ts
```diff
@@ -35,25 +35,14 @@ const letterKeys = [
 describe("calcite-time-picker", () => {
   it("renders", async () => renders("calcite-time-picker", { display: "inline-block" }));
 
-  it("is accessible", async () => accessible(`<calcite-time-picker hour="00" minute="00"></calcite-time-picker>`));
+  it("is accessible", async () => accessible(`<calcite-time-picker step="1" value="00:00:00"></calcite-time-picker>`));
```
jcfranco, line 38: Nitpick: shouldn't this be accessible without configuring?

src/demos/_assets/locales.ts
```diff
@@ -0,0 +1,223 @@
+const locales = [
+  { name: "Afrikaans", locale: "af" },
+  { name: "Albanian", locale: "sq" },
```
jcfranco, line 3: Please check with the localization team to confirm the list of locales we need to support. The `locales` list should reflect that (separate from this PR). I'd be useful to doc this list as well. cc @benelan

src/components/calcite-time-picker/calcite-time-picker.tsx
```diff
@@ -642,212 +676,220 @@ export class CalciteTimePicker {
...
+            aria-valuemin="1"
+            aria-valuenow={minuteIsNumber && parseInt(this.minute)}
+            aria-valuetext={this.minute}
+            class={{
+              [CSS.input]: true,
+              [CSS.minute]: true
+            }}
+            onFocus={this.focusHandler}
+            onKeyDown={this.minuteKeyDownHandler}
+            ref={(el) => (this.minuteEl = el)}
```
jcfranco, line 768: Can you add class methods for storing the references?

jcfranco (approved):
Great work, @eriklharper! 🎉 🌮🦃 🦖

src/demos/_assets/locales.ts
eriklharper: https://github.com/Esri/calcite-components/issues/3660

Issue #3215, opened 2021-10-12
Bug: calcite-tile doesn't handle long heading correctly


AdelheidF:
![image](https://user-images.githubusercontent.com/44481292/137012689-2f84b52b-05fa-41a8-b8e5-14c8b2bf50d5.png)

### Actual Behavior

text moves beyond tile

### Expected Behavior

text ends with ellipses

### Reproduction Steps and Sample

https://codepen.io/afreitag/pen/qBXdEoN?editors=1000

_Sample_:

1.

### Relevant Info

_Version_: `@esri/calcite-components@1.0.0-beta.65`

- [ ] CDN
- [ ] NPM package

caripizza:
Can we remove the [`10%` width](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-tile/calcite-tile.scss#L50) I see set on the .content class? It's ignored since that element already has `flex-auto` set on it (which includes flex-basis for width).

benelan:
Assigning back to @y0n4 based on Cari's comment

benelan:
verified on next

PR #3361, opened 2021-10-26
fix(tile): adjust the heading styling to be word wrapped including long string texts #3215

y0n4:
**Related Issue:** #3215

## Summary
`calcite-tile` heading text will word wrap long string texts. It will also break the word halfway if the a single string is too long for the width

Screen Shot 2021-11-04 at 10 19 21 AM


src/components/calcite-tile/calcite-tile.scss
```diff
@@ -77,6 +79,9 @@
     ease-in-out
     font-medium
     pointer-events-none;
+  white-space: nowrap;
```
driskull, line 82: Can we use tailwind CSS where possible?

driskull (commented):
just a comment :)

driskull (approved):
Looks good to me! Is a screener test needed to make sure this doesn't break?

y0n4:
Ah, good point thanks. I will adjust the story for the screener test to catch that

Issue #2145, opened 2021-05-13
Bug: calcite loader's `no-padding` attribute has no visible effect


nwhittaker:
### Actual Behavior
The position of the visible loader elements remains unchanged when toggling the `<calcite-loader>` element's `no-padding` attribute.

Inside a box of the same dimensions as the loader:

| With `no-padding` | Without `no-padding` |
| - | - |
| <img width="66" alt="Screen Shot 2021-05-13 at 3 47 18 PM" src="https://user-images.githubusercontent.com/421496/118179157-ec295c80-b402-11eb-96fd-ef7520053975.png"> | <img width="66" alt="Screen Shot 2021-05-13 at 3 47 54 PM" src="https://user-images.githubusercontent.com/421496/118179376-37436f80-b403-11eb-8f44-24e93f432033.png"> |

### Expected Behavior
With `no-padding`, the position of the visible loader elements are no longer positioned `4rem` from the top.

Inside a box of the same dimensions:

<img width="66" alt="Screen Shot 2021-05-13 at 3 53 05 PM" src="https://user-images.githubusercontent.com/421496/118179655-92756200-b403-11eb-8a4d-5cb79369e43e.png">

### Reproduction Steps or Sample
```html
<div style="border: 1px solid black; width: 64px; height: 64px;">
  <calcite-loader label="loading" active no-padding></calcite-loader>
</div>
```

### Relevant Info
_Version_: `@esri/calcite-components@v1.0.0-next.191`

jcfranco:
`no-padding` does seem to remove the top/bottom padding, but it looks like the loader isn't positioned correctly within the host.

https://codepen.io/jcfranco/pen/vYJLxYo?editors=1000

![Screen Shot 2021-10-15 at 5 33 36 PM](https://user-images.githubusercontent.com/197440/137566774-dd5a6c3c-7e32-45a1-9b79-013a7734b78d.png)

FWIW, `no-padding` got introduced by https://github.com/Esri/calcite-components/issues/240.

benelan:
verified on `next`

PR #3370, opened 2021-10-28
fix(loader): alter loader styling to prevent no padding prop from overlapping #2145

y0n4:
**Related Issue:** #2145

## Summary
If the loader has the "determinate" prop passed. The value centered in the loader animation will be consistent as well

Before
Screen Shot 2021-10-27 at 4 25 53 PM
After
Screen Shot 2021-10-27 at 4 07 13 PM


jcfranco (approved):
If there isn't one already, can you add a story for this? Other than that, this LGTM!

Issue #3048, opened 2021-09-15
Bug: Dropdown wrapper causes UI bug when combined with a Panel in a PickList


asangma:
### Actual Behavior
Because the `calcite-dropdown-wrapper` node still has a height when its hidden, it can cause height calculation issues in small containers.

### Expected Behavior
Dropdown item container shouldn't have a height when hidden.

### Reproduction Steps and Sample

_Sample_: https://codepen.io/mtb-alan/pen/wvePErY

1. Use filter to search for "lisa"
2. Observe as the items disappear.
3. Note that a scrollbar appears but there doesn't appear to be content.
4. Inspect the sorting action and search in the browser's inspector for "calcite-dropdown-wrapper"

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

asangma:
@jcfranco @julio8a  @benelan This could use some priority. The combo of PickList in a Panel (often in a Popover) with actions is a regular occurrence in existing and upcoming interfaces.

benelan:
@asangma This sprint is already quite hefty, I'll put it in the next one

asangma:
Thanks @benelan!

anveshmekala:
issue  #3071  resolved this bug.

benelan:
verified on beta.75

PR #3378, opened 2021-10-28
fix(dropdown): dropdown wrapper no longer overflows inside a panel

anveshmekala:
**Related Issue:** #3048 

## Summary

This fixes  `overflow` caused by dropdown wrapper when wrapped inside a `panel`  with `filter` in a `pick-list`


src/components/calcite-dropdown/calcite-dropdown.tsx
```diff
@@ -161,7 +161,9 @@ export class CalciteDropdown {
             onTransitionEnd={this.transitionEnd}
             ref={this.setScrollerEl}
```
driskull, line 162: This seems to work for me, I don't think we need another div element.

```
<div
            class={{
              ["calcite-dropdown-content"]: true,
              [PopperCSS.animation]: true,
              [PopperCSS.animationActive]: active
            }}
            hidden={!active}
            onTransitionEnd={this.transitionEnd}
            ref={this.setScrollerEl}
          >
```

anveshmekala: The issue with this approach is that `calciteDropOpen` event isn't emitted as the element is hidden.

driskull: Gotcha. makes sense

driskull (approved):
LGTM.

Lets get another reviewer since I contributed.

asangma:
Thanks @anveshmekala!
It's a better situation.

asangma (approved):
Whoop whoop!

src/components/calcite-dropdown/calcite-dropdown.e2e.ts
```diff
@@ -1031,4 +1031,35 @@ describe("calcite-dropdown", () => {
       })
     ).toBe(false);
   });
+
+  it("dropdown wrapper shouldnt' have height when filter results empty and combined with a PickList in Panel  #3048", async () => {
```
asangma, line 1035: typo: "shouldn't"

src/components/calcite-dropdown/calcite-dropdown.tsx
```diff
@@ -316,6 +318,8 @@ export class CalciteDropdown {
 
   mutationObserver = createObserver("mutation", () => this.updateItems());
 
+  resizeObserver = createObserver("resize", () => this.setMaxScrollerHeight());
```
driskull, line 321: @anveshmekala we might just be able to call `reposition()` along with `this.setMaxScrollerHeight()` in order to fix https://github.com/Esri/calcite-components/issues/3234

driskull: I verified that calling `this.reposition();` inside `setMaxScrollerHeight` fixes the issue.

@anveshmekala

Issue #2949, opened 2021-08-31
Enhancement: Input - support "inputmode"


macandcheese:
### Description
As a follow up to https://github.com/Esri/calcite-components/issues/2726, we should add support for inputmode: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode

### Acceptance Criteria 
inputmode is a usable attribute on the input component

#### Which Component
Input

#### Example Use Case
Implementer should have control over the type of inputmode 

#### Related

#2949

jcfranco:
Verified with https://codepen.io/jcfranco/pen/wvqXrrb?editors=1000 on an iOS device.

PR #3382, opened 2021-10-29
feat(input): add inputmode property to specify mobile keyboard display

benelan:
**Related Issue:** #2949

## Summary
Added `inputmode` property to specify the mobile keyboard style
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode


jcfranco:
This PR is no longer needed because of https://github.com/Esri/calcite-components/pull/3396, right?

benelan:
> This PR is no longer needed because of #3396, right?

This is a different prop (`inputMode`), but I'm going to make sure I get `enterKeyHint` working first before installing this one

eriklharper:
Should we be creating non-standard attribute, in this case `input-mode` as opposed to `inputmode` or should we just pass the HTML standard attribute down to the native shadowed input with `inputMode={this.el.inputMode}` or if that's not accessible, then `inputMode={this.el.getAttribute("inputmode")}`?

benelan:
@eriklharper I'm using the one from the base element, I tested and it worked for `enterKeyHint`

eriklharper (approved):

Issue #3383, opened 2021-10-29
Enhancement: Input - add `enterkeyhint`


macandcheese:
### Description
Similar to the recently added `inputmode` - this is a useful native attribute that can help devs provide a better UX to end users: 

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/enterkeyhint#values

### Acceptance Criteria 
`enterkeyhint` is available as an attribute.

### Relevant Info 

Should we just audit all native attributes and add them - there are others we don't support, `pattern`, etc., that may be useful (or, wait until we "break apart" and add them to their related input types?)

jcfranco:
@benelan Can we refactor the prop this sprint to follow prop-naming conventions (`enterKeyHint`)? For consistency, we want to avoid having mixed casing in our APIs. This should also apply to https://github.com/Esri/calcite-components/issues/2949 cc @macandcheese

jcfranco:
Verified with https://codepen.io/jcfranco/pen/wvqXrrb?editors=1000 on an iOS device.

PR #3384, opened 2021-10-29
feat(input): add enterkeyhint property to specify mobile enter key type

benelan:
**Related Issue:** #3383

## Summary
Added `enterkeyhint` property to input which specifies which type of `enter` key should be provided on a mobile keyboard


macandcheese (approved):
Fastest Issue -> PR Turnaround I've seen 🎈 

Question - do we need to list the values in the hint above the typed values? I think this would be shown twice on the developer site.

benelan:
> Question - do we need to list the values in the hint above the typed values? I think this would be shown twice on the developer site.

Good point, I'll remove the list on the comment for both PRs.

PR #3385, opened 2021-10-29
build(ci): try to fix screener github bug

benelan:
**Related Issue:** NA

## Summary
Maybe the screener check is showing the old one since I named it the same thing? So I changed the new one from `screener` to `screen` lets so if that gets rid of this weird bug

### DIFF
--- a/.github/workflows/pr-screener.yml
+++ b/.github/workflows/pr-screener.yml
@@ -33,6 +33,6 @@ jobs:
           authToken: ${{secrets.GITHUB_TOKEN}}
           context: Screener Report
           description: Click details to view the report on Screener
-          state: ${{ job.status }}
+          state: success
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-screener.yml
```diff
@@ -3,7 +3,7 @@ on:
   pull_request:
     branches: [master]
 jobs:
-  screener:
+  screen:
```
jcfranco, line 6: Nitpick: this could be renamed to express what the check does, no? `screenshot_tests` or something like that.

jcfranco (approved):
💥

.github/workflows/pr-screener.yml
benelan: sure

Issue #3279, opened 2021-10-15
Bug: TabNav causes scrolling issue when container has set width


asangma:
### Actual Behavior
When in a container with a set width, TabNav causes a horizontal scrollbar.
It seems to be related to the combo of [the padding and negative margin](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-tab-nav/calcite-tab-nav.scss#L23-L24).

### Expected Behavior
Not have horizontal scrollbars.

### Reproduction Steps and Sample

_Sample_: https://codepen.io/mtb-alan/pen/GRvorRY

1. In demo, change tabs
2. observe the horizontal scrollbar.
3. note that it doesn't matter how wide the container is.

### Relevant Info
Fixed width container are common in the JS API.

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

benelan:
verified on beta.72

PR #3387, opened 2021-10-29
fix(tab): remove the scroll that pops up regardless of any width size (#3279)

y0n4:
**Related Issue:** #3279

## Summary
It seems like from what @asangma shared, the tab titles were affecting the width scroll. Now with set width or no width, the horizontal scroll will not show

Before
Screen Shot 2021-12-15 at 1 46 11 PM

After
Screen Shot 2021-12-15 at 1 45 25 PM

I also added a one-line fix from issue #3541


jcfranco (approved):
Code LGTM! I'll let @caripizza chime in on the visual changes. 🎨  👀

caripizza:
I think removing the negative margin messes up more than it's fixing here, since those e2e tests are failing along with the stories. (Tab-nav's active indicator styles seem to rely on this margin for placement currently.)

@asangma if you add `width: 100%; overflow-x: hidden;` on the parent calcite-tabs, it seems to fix the issue/prevent the horizontal scrollbars in your CodePen. Can we do something like that instead here?

```html
<calcite-tabs layout="center" position="above"  style="width: 100%; overflow-x: hidden;">
...
```

asangma:
@y0n4 Hmmm...I'm seeing the same scrolling issue in compact spaces. I think it's worth it to fix this in the component. IMO, negative margins are probably not a thing we should do.

I'll pull the branch and have a look.

jcfranco:
Any updates on this one @asangma?

PR #3393, opened 2021-10-29
chore: remove title from issue templates

benelan:
**Related Issue:** NA

## Summary
Sometimes people forget to fill out the title which is possible since we already prefill some of it. If we remove the prefilled part and people don't add a title then the issue won't submit. All of the titles are duplicates of the labels that are added to the issue so having both is redundant anyway


jcfranco (approved):
💥

Issue #3389, opened 2021-10-29
Bug: calcite-tab will not show horizontal scroll if the content width gets too large


y0n4:
### Actual Behavior
When the width is already set and wrapped around the tabs component, and the tab content becomes larger than the width length, there is no scroll to view the rest of the content

### Expected Behavior
Have a scroll shown when the content gets larger than the given height/width

### Reproduction Steps and Sample

_Sample_: https://codepen.io/yonachos/pen/GRvvGwd
1. view content at first tab "tab one"
2. observe that you cannot view the full image since it is cut off

### Relevant Info
May have to add an overflow styling in the tab component

_Reproducible version_: `@esri/calcite-components@1.0.0-<version>`

- [ ] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-<version>`

benelan:
verified on `next`

PR #3395, opened 2021-11-01
fix(tab): add scroll to tab content #3389

y0n4:
**Related Issue:** #3389 

## Summary
Previously tab contents do not have a autoscroll if the content itself expands larger than than the container width/height

Before
Screen Shot 2021-10-29 at 1 21 20 PM
After
Screen Shot 2021-10-29 at 1 21 45 PM


driskull (approved):
Can you use the tailwind variables instead? Otherwise, it looks good. Do we need to add a test for this? Screenshot or e2e?

src/components/calcite-tabs/calcite-tabs.stories.ts
```diff
@@ -28,7 +28,13 @@ export const Simple = (): string => html`
       <calcite-tab-title>Tab 4 Title</calcite-tab-title>
     </calcite-tab-nav>
 
-    <calcite-tab active><p>Tab 1 Content</p></calcite-tab>
+    <calcite-tab active>
```
jcfranco, line 31: @caripizza @eriklharper how would you suggest we make it clear what each story covers visually? Would a simple comment suffice here or do we need a separate test? A separate one seems clearer to me, but may be overkill.

jcfranco (approved):
Code LGTM! 

@Esri/calcite-design-representatives can some one take a look at the screener changes? 🙇🏻

macandcheese:
@jcfranco I approved the changes in Screener - do we want to keep the "Simple" story as it was, and add another separately for this demonstration instead?

src/components/calcite-tabs/calcite-tabs.stories.ts
eriklharper: Good question.  I'd say that if we were to revamp the stories for this component with the steps utilities, we could simply use the snapshot name to label what each story is visually testing, so for this particular change we would name the snapshot something like "horizontal overflow scroll" or something, but doing so will necessitate the small bit of refactoring to turn the Simple story into a stepStory with that additional screenshot in there.  We could make the horizontal overflow scroll snapshot show up in Tab 2 and just add a step that clicks Tab 2 and then takes a snapshot of the resulting UI to capture it under that name.

eriklharper: Might be out of scope for this particular PR, but if @y0n4 wants to try and tackle that and learn the new steps utils at the same time, I'm totally available and glad to help her out on that!

PR #3396, opened 2021-11-01
refactor(input): enterKeyHint prop naming convention

benelan:
**Related Issue:** #3383

## Summary
enterKeyHint prop naming convention


jcfranco (approved):
👍

eriklharper:
Isn't this considered an anti-pattern to declare global attributes as Props in a stencil component? 
image

benelan:
> Isn't this considered an anti-pattern to declare global attributes as Props in a stencil component?

I followed what ionic did
https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/input/input.tsx#L108

eriklharper:
> > Isn't this considered an anti-pattern to declare global attributes as Props in a stencil component?
> 
> I followed what ionic did https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/input/input.tsx#L108

I think that's why you're getting those errors, because this property is already defined on the inherited interface that all Stencil components use.  You shouldn't need to redefine this in the component because it appears to already be on the prototype chain.

paulcpederson:
@eriklharper  I don't think that error is related to this pr? He's updating `enterKeyHint`. I also don't see a `lang` prop anywhere in the codebase on `master`.

@benelan the tests I think are failing because `HTMLElement` inherits `ElementContentEditable` which has `enterKeyHint` declared as a required attribute. So we're setting this optionally on our comopnents and typescript thinks it's required. Maybe you already fixed that, though?

eriklharper:
The `lang` screenshot I showed is just an example on a WIP change I made to add a lang prop to input-time-picker.  You're correct that this is not used currently anywhere in the codebase.

paulcpederson:
@eriklharper oooooooooooooooooooh
over-my-head-no

eriklharper:
@paulcpederson can you show us where this inheritance is defined?  `HTMLElement inherits ElementContentEditable`

paulcpederson:
it's deep dwithin the bowels of typescript in  `lib.dom.d.ts`.

Theoretically you should be able to remove this prop or mark it as required to get around this? or liberally sprinkle `@tsignores`...

Issue #2938, opened 2021-08-27
Bug: calcite-color-picker calls calciteColorPickerInput on initial setup


AdelheidF:
I noticed that the initial setup of `calcite-color-picker` already calls the `calciteColorPickerInput` event. 
That's an issue because by just creating this popover it sets an effect on the layer.

```
          <calcite-color-picker
            dir="ltr"
            scale="m"
            value="#000000"
            format="hex"
            appearance="default"
            intlBlue={blue}
            intlGreen={green}
            intlHex={hex}
            intlRed={red}
            intlSaved={saved}
            hideSaved={true}
            onCalciteColorPickerInput={this.onPropChange}
          ></calcite-color-picker>
```

### Reproduction Steps or Sample

1.

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

AdelheidF:
FYI, the other day I noticed that `calcite-radio-group` has the same issue.

benelan:
verified on next

PR #3399, opened 2021-11-02
fix(color-picker): avoid input/change events from firing when initializing color/value

jcfranco:
**Related Issue:** #2938

## Summary

This fixes an issue where the code to set the internal color from the initial value on `connectedCallback` was causing `change` and `input` event to unintentionally fire.


src/components/calcite-color-picker/calcite-color-picker.tsx
```diff
@@ -294,6 +297,8 @@ export class CalciteColorPicker {
 
   private hueThumbState: "idle" | "hover" | "drag" = "idle";
 
+  private internalColorUpdateContext: "internal" | "initial" | null = null;
```
driskull, line 300: 👍

src/components/calcite-color-picker/calcite-color-picker.tsx
```diff
@@ -696,15 +701,17 @@ export class CalciteColorPicker {
 
     const valueIsNoColor = allowEmpty && !value;
     const parsedMode = parseMode(value);
-    const valueIsCompatible = (format === "auto" && parsedMode) || format === parsedMode;
+    const valueIsCompatible =
+      valueIsNoColor || (format === "auto" && parsedMode) || format === parsedMode;
```
driskull, line 705: nitpick. Maybe const could be `valueNotAColor` or `colorlessValue`? no biggie. Just a weird name `valueIsNoColor`

driskull (approved):
👍 🎉

src/components/calcite-color-picker/calcite-color-picker.tsx
jcfranco: Good idea. Changed it to `willSetNoColor` to better express the intent.

PR #3410, opened 2021-11-02
build(deps): remove unnecessary stylelint-scss as it is already bundled by stylelint-config-recommended-scss

jcfranco:
**Related Issue:** N/A

## Summary

This removes a dependency that is already bundled by https://www.npmjs.com/package/stylelint-config-recommended-scss:

```
This config:

extends the `stylelint-config-recommended` shared config and configures its rules for SCSS
bundles the `stylelint-scss` plugin pack and turns on its rules that check for possible errors
bundles the `postcss-scss` custom syntax and configures it
```


driskull (approved):
Nice. 🧹 cleaning up 🧼

Issue #3236, opened 2021-10-13
Bug: calcite-fab in calcite-panel slot overlaps content in Firefox


ellenupp:
### Actual Behavior

When a calcite-fab is used in the `fab` slot of a calcite-panel, and the content takes up the entire vertical height of the panel, the fab overlaps the panel even when scrolling to the bottom (in firefox). 
![image](https://user-images.githubusercontent.com/14304581/137018297-d0b93f2a-b43b-4734-838f-783fa8f629f7.png)

### Expected Behavior

The fab should be sticky'd to the bottom of the panel, but once the scroll reaches the bottom it should no longer overlap.
In chrome, if you scroll to the bottom of the panel, all content is visible with the fab positioned below it. 
![image](https://user-images.githubusercontent.com/14304581/137018549-cf045cc7-a6fa-4606-9144-9fb97d2beb0a.png)

### Reproduction Steps and Sample

_Sample_: https://codepen.io/ellupp/pen/qBXdEdB?editors=1000

1. Open sample in firefox, scroll to bottom of panel. 

### Relevant Info

Browser: Firefox (not reproducible in Chrome)

_Version_: `@esri/calcite-components@beta.65`

benelan:
reproducible in beta.67 too using the sample ^

PR #3413, opened 2021-11-02
fix(panel): get the fab slot to have more spacing on firefox #3236

y0n4:
**Related Issue:** #3236

## Summary
Previously the fab slot button would cover the text as the scroll gets to the bottom of the panel when browsing on Firefox. After the adjustment, it should be able to operate smoothly like Chrome and would not interfere with other input panel attributes


src/components/calcite-panel/calcite-panel.tsx
```diff
@@ -391,10 +391,12 @@ export class CalcitePanel {
 
   renderContent(): VNode {
     return (
-      <section class={CSS.contentContainer} onScroll={this.panelScrollHandler} tabIndex={0}>
-        <slot />
+      <div class={CSS.contentWrapper}>
+        <section class={CSS.contentContainer} onScroll={this.panelScrollHandler} tabIndex={0}>
```
driskull, line 395: @y0n4 can you verify that the tabIndex doesn't nee to be moved to the node that has overflow:auto?

driskull (approved):
@y0n4 The code looks good. Can you verify that a user can still scroll using a keyboard since the overflow auto was moved and the tabindex wasn't moved.

src/components/calcite-panel/calcite-panel.tsx
y0n4: User is able to scroll using keyboard and navigate with tab keyboard. Moved those properties to the container that has overflow

PR #3415, opened 2021-11-02
test(action): adding a default story so that it shows up in the deployed Storybook

eriklharper:
**Related Issue:** n/a

## Summary

This pr adds a `Default` story for `calcite-action` so that it appears on the deployed Storybook website.


jcfranco (approved):
👍🏼

Issue #3349, opened 2021-10-25
Bug: Styling inconsistencies in dropdown items (link vs non-link, RTL)


ellenupp:
### Actual Behavior

There are some inconsistencies in the styling of dropdown items:
- horizontal padding is different in RTL vs LTR
- vertical padding is different between links and non-links (extra padding on links makes them 40px in height while non-links are 32px at medium scale)
- focus and hover states are different between links and non-links (text color does not change for links)

![image](https://user-images.githubusercontent.com/14304581/138755281-71872876-50f2-4f86-8690-5924b47ea419.png)

### Expected Behavior

Padding should be consistent between LTR and RTL. Not sure if it is by design that there should be differences between links and non-links but I would expect them to have the same height at least. 

### Reproduction Steps and Sample

_Sample_: https://codepen.io/ellupp/pen/YzxZaQz?editors=1000

1. Create a calcite-dropdown that includes both link (has href attribute) and non-link calcite-dropdown-items. Observe the differences. View same dropdown in RTL and observe differences.

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-beta67`

- [x] CDN
- [x] NPM package

benelan:
verified on `beta.69`

PR #3422, opened 2021-11-03
fix(dropdown-item): match padding for RTL vs LTR and links vs non-links, consistent focus and hover states for all

Elijbet:
**Related Issue:** #3349

## Summary

Dropdown items now have consistent padding for RTL/LTR as well as links/non-links.
Focus and Hover states are now consistent for both RTL/LTR and links/non-links.

Observe the behavior on demo Codepen before:
https://codepen.io/ellupp/pen/YzxZaQz?editors=1000

Current

https://user-images.githubusercontent.com/19231036/140199954-25bb7354-5879-4839-be77-d837c4fe329e.mov

### DIFF
--- a/src/components/calcite-dropdown-item/calcite-dropdown-item.scss
+++ b/src/components/calcite-dropdown-item/calcite-dropdown-item.scss
@@ -55,13 +55,16 @@
   @apply ml-auto mr-0;
 }
 
+.calcite--rtl a.dropdown-link {
+  @apply pr-0 pl-0;
+}
+
 //focus
 :host,
 .container--link a {
   @apply focus-base;
 }
-:host(:focus),
-.container--link a:focus {
+:host(:focus) {
   @apply focus-inset;
 }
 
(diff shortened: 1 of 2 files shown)


src/components/calcite-input/calcite-input.tsx
```diff
@@ -512,7 +512,6 @@ export class CalciteInput implements LabelableComponent {
     if (direction === "up" && ((!inputMax && inputMax !== 0) || inputVal < inputMax)) {
       newValue = (inputVal + inputStep).toString();
     }
-
```
jcfranco, line 515: Please revert this as it is not related to the fix. 

Sidebar: IMO, newlines between if statements help with readability (easier to scan vertically).

src/components/calcite-dropdown-item/calcite-dropdown-item.scss
```diff
@@ -111,7 +119,8 @@
   }
 }
 
-:host([active]) .container:not(.container--none-selection) {
+:host([active]) .container:not(.container--none-selection),
+:host([active]) .container--link a {
```
jcfranco, line 123: Can you target the anchor's class (`dropdown-link`) instead of the element? This helps avoid cases where the style may break if the element structure is updated.

This also applies to the selectors above.

src/components/calcite-dropdown-item/calcite-dropdown-item.scss
```diff
@@ -55,13 +55,16 @@
   @apply ml-auto mr-0;
 }
 
+.calcite--rtl a.dropdown-link {
```
jcfranco, line 58: It think you can make the selector less specific and remove the `a` by moving this to the `.calcite-rtl` selector below: https://github.com/Esri/calcite-components/blob/9cc831345bdf987eb14cd86ddb90bbb1d385ee2f/src/components/calcite-dropdown-item/calcite-dropdown-item.scss#L115-L120

src/components/calcite-dropdown-item/calcite-dropdown-item.scss
```diff
@@ -122,19 +131,6 @@
   }
 }
 
-// none selection mode and scales
```
jcfranco, line 125: Nice! ✨ :broom:✨

jcfranco (changes requested):
This is looking great! Just had a few comments and this will be good to merge. 

Also, can you update the PR title to follow Conventional Commits?

jcfranco (approved):
Awesome! 🤩

Great, descriptive PR title by the way.

asangma:
🚀

PR #3424, opened 2021-11-03
fix(label): avoid associating labels to nested labelable components

jcfranco:
**Related Issue:** #3344 

## Summary

This fixes an issue where components with nested labelable components (either in shadow DOM or slotted) could potentially be associated with the label instead of the owning component. For example, in the following snippet, the label gets associated with the slotted button.

```html
<calcite-label>
  Labeled input (wrapped) ⛔
  <calcite-input>
    <calcite-button id="action-slot" slot="action">Go</calcite-button>
  </calcite-input>
</calcite-label>
```

Note that this is not an issue with for-associated labels.

```html
<calcite-label for=input>
  Labeled input (for) ✅
</calcite-label>
<calcite-input id=input>
  <calcite-button id="action-slot" slot="action">Go</calcite-button>
</calcite-input>
```

The label lookup occurs in component's `connectedCallback`, which is not guaranteed to be called in hierarchical order. This is because each component's `connectedCallback` method gets called in the order that they are defined in the browser. For consumers of the lazy-loaded bundle, this order will vary.

This PR changes the rules for labelable components and will no longer associate a labelable component if it is a child of another custom element.

Just as an idea, if we needed to support nested labelables, we could refactor the label util to handle the nested lookup to be async and resolve when `calcite-label` has finished loading or to resolve the lookup when clicked.

### DIFF
--- a/src/utils/label.ts
+++ b/src/utils/label.ts
@@ -54,12 +54,13 @@ const findLabelForComponent = (componentEl: HTMLElement): HTMLCalciteLabelElemen
 };
 
 function hasAncestorCustomElements(label: HTMLCalciteLabelElement, componentEl: HTMLElement): boolean {
-  let composedPath: HTMLElement[];
+  let traversedElements: HTMLElement[];
   const customElementAncestorCheckEventType = "custom-element-ancestor-check";
 
   const listener = (event) => {
     event.stopImmediatePropagation();
-    composedPath = event.composedPath() as HTMLElement[];
+    const composedPath = event.composedPath() as HTMLElement[];
+    traversedElements = composedPath.slice(composedPath.indexOf(label), composedPath.indexOf(componentEl));
   };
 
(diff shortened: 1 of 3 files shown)


src/utils/label.ts
```diff
@@ -22,18 +22,59 @@ export interface LabelableComponent {
...
+    // labelable components within other custom elements are not considered labelable
+    hasAncestorCustomElements(parentLabel, componentEl)
+  ) {
+    return null;
+  }
+
+  return parentLabel;
 };
 
+function hasAncestorCustomElements(label: HTMLCalciteLabelElement, componentEl: HTMLElement): boolean {
```
driskull, line 56: Looks good. We can always tweak this if necessary.

driskull (approved):
🎉 🌮

jcfranco:
Screener failure is unrelated. Merging.

Issue #3138, opened 2021-09-29
Calcite Filter Enhancement: 


sam11927:
### Description
Should have some way of knowing if the calcite-filter has text currently typed into the input, or knowing what the text is. 

### Acceptance Criteria 
When text is entered into the input element of the calcite-filter, an outside component should be able to read the entered text and know if it is empty or not.

### Relevant Info 
It would be nice to have both a public `isEmpty` and a `Prop` for the input elements current text value.

#### Which Component
`calcite-filter`

#### Example Use Case
Say someone wants to render different elements based on whether or not the calcite-filter has any text entered. They should be able to read the value or check if it is empty.

driskull:
@jcfranco Seems like combobox should also have this ability, right? I think we should add a value prop to both of these components.

driskull:
I wonder if we should also add a property for filtered items? Like "selectedItems" or "filteredItems".

driskull:
Will be addressed by making the component public via #2208.

benelan:
verified on `next`! https://codepen.io/benesri/pen/wvrMRyE?editors=0010

Issue #2208, opened 2021-05-25
Enhancement: Expose calcite-filter or equivalent with improvements


ffaubry:
### Description

value-list implies that the data to be filtered are the value-list-items.
My issue is that my collection of value-list-items are coming from a service.
As the user types a word in the value-list filter component (calcite-filter), I would like to use the word and query the service and then update the value-list-items based on the result.

In that sense, it is not really about 'filtering' but rather about 're-fetching' data for the value-list.
This is why, my solution was to simply disable the filter on the value list and provide my own UI above the value-list and control myself the whole workflow.
I can achieve this by using calcite-input. However, the look and feel of calcite-input is not matching the look and feel and animations of the calcite-filter.

### Acceptance Criteria 

A control that expose the same UX as calcite-filter (animations, etc) and an event when the text changed.

jcfranco:
Installed! 

Recap:

* `calcite-filter` is now public
* added `items` and `filteredItems` props
* `data` and event payload is deprecated and will be removed in the next sprint
* I will create follow-up issues for
  * dropping deprecated props in the next sprint 
  * adding support for custom filter function

jcfranco:
[Belated] Installed!

benelan:
verified on `next`! https://codepen.io/benesri/pen/wvrMRyE?editors=0010

PR #3432, opened 2021-11-04
refactor(filter): Add value property to set a filter.

driskull:
**Related Issue:** #3138

## Summary

refactor(filter): Add value property to set a filter.


jcfranco (approved):
Nice enhancement. 

Since `calcite-filter` isn't public, can we install this as a refactor and address #3138 by making the component public via #2208. I think the API is good for release. WDYT?

Issue #3436, opened 2021-11-04
Bug[slider]: range slider maxValue label is missing CSS class


anveshmekala:
### Actual Behavior

The range slider maxValue label is missing `.tick__label--max` 

### Expected Behavior

maxValue label  in range slider should have `.tick__label--max`

### Reproduction Steps and Sample

Inspect the maxValue tick and observe the <span> element doesn't have class `.tick__label--max`

_Sample_:

1.https://codepen.io/ANVESHMEKALA/pen/wvqyJYw

### Relevant Info

The bug is introduced with  #3142

_Reproducible version_: `@esri/calcite-components@1.0.0-beta.68`

- [ ] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-beta.68`

benelan:
Hi @anveshmekala, I'm unable to reproduce this issue with the sample you provided. Can you edit it and make the issue reproducible please?

anveshmekala:
> Hi @anveshmekala, I'm unable to reproduce this issue with the sample you provided. Can you edit it and make the issue reproducible please?

PR #3439, opened 2021-11-05
fix(slider): range slider maxValue label no longer overlaps with preceding labels

anveshmekala:
**Related Issue:** #3436 

## Summary

This fix add backs the missing CSS class `tick__label--max`  for `maxValue` label in range slider


src/components/calcite-slider/calcite-slider.e2e.ts
```diff
@@ -196,6 +196,17 @@ describe("calcite-slider", () => {
     expect(changeEvent).toHaveReceivedEventTimes(1);
   });
 
+  it("maxValue label in range slider should have tick__label--max class", async () => {
```
jcfranco, line 199: The missing class affected the behavior of overlapping labels, right? Could we add stories to cover this instead of asserting on the class being there? 

This commit could be helpful as it points to a version of the demo page with configurations for all of the overlapping scenarios: https://github.com/Esri/calcite-components/blob/cd16695ee0546b4d89f220d3f5f0aabee4e6c06b/src/demos/calcite-slider.html#L246

src/components/calcite-slider/calcite-slider.stories.ts
```diff
@@ -48,6 +48,22 @@ export const Range = (): string => html`
   ></calcite-slider>
 `;
 
+export const RangeLabeledTicks = (): string => html`
```
jcfranco, line 51: I think we need more stories for Screener to catch the rest of the cases where labels may overlap. I had shared in a comment a link to a previous demo page with each case.

jcfranco (changes requested):

src/components/calcite-slider/calcite-slider.stories.ts
anveshmekala: I think this story should catch the expected behavior. This issue focus on tick labels irrespective of the thumb handles  positions. One other example i can think of is the mirrored range slider.

jcfranco: If you look at [the following](https://codepen.io/jcfranco/pen/mdMoKJv?editors=1000), you can see the difference between the additional stories I suggested and the ones in this PR: 

Screen Shot 2021-11-19 at 6 42 32 PM

With the additional stories, we have more Screener coverage on visual changes to labels and thumbs at the edges. If I remember correctly, the entries from the demo page covered a different code path of the label positioning. I don't believe this has changed, so that is why I was suggesting adding these.

anveshmekala: Make sense. The current demo's follow a different path.

jcfranco (approved):
Awesome, thanks for adding the stories!

src/components/calcite-slider/calcite-slider.stories.ts
```diff
@@ -48,6 +48,103 @@ export const Range = (): string => html`
...
+  <calcite-slider
+    min="${number("min", 5)}"
+    min-label="${text("min-label", "Temperature, lower bound")}"
+    min-value="${number("min-value", 95)}"
+    max="${number("max", 100)}"
+    max-label="${text("max-label", "Temperature, upper bound")}"
+    max-value="${number("max-value", 100)}"
+    step="${number("step", 10)}"
+    ${boolean("label-handles", false)}
```
jcfranco, line 60: I just noticed this after looking at the Screener diffs. Can you also enable the label handles before merging?

PR #3445, opened 2021-11-06
refactor(filter): apply filter when data changes

jcfranco:
**Related Issue:** #3138 #2208 

## Summary

Also updates test to cover this case.

cc @sam11927


src/components/calcite-filter/calcite-filter.tsx
```diff
@@ -35,6 +35,11 @@ export class CalciteFilter {
    */
   @Prop() data: object[];
 
+  @Watch("data")
+  watchDataHandler(): void {
+    this.filter(this.value);
```
driskull, line 40: nice

driskull (approved):

PR #3449, opened 2021-11-08
feat(filter): add filter component

jcfranco:
**Related Issue:** #2208 

## Summary

This PR makes some tweaks to the existing, internal, filter component for its premiere:

```ts
interface HTMLCalciteFilterElement {
  data: T[]; // deprecated
  items!: T[]; // ✨ new✨
  filteredItems: T[]; // T = the item type ✨ new✨
}
```

### Notes

* The event detail will be deprecated in favor of this new property.
* `items` will be doc'd as required, but enforced via types next release
* `data` is deprecated and will be removed in next release as it was not publicly doc'd

### Questions

* Would renaming data => `items` and `filteredData` => `filteredItems` or `filtered` be a better fit? If so, for this release, we can add them and deprecate `data` and remove in the next sprint since the filter was never publicly doc'd
* Do we need a story + SDK entry for this? Filter is useful for a consistent filtering experience, but not sure what the examples/stories need to cover. cc @caripizza @macandcheese @benelan 

### Miscellaneous

* This commit is being treated as a new component being added, so I also tidied up a few tests and resource files. Let me know if you want me to break these out.

### DIFF
--- a/src/components/calcite-filter/calcite-filter.e2e.ts
+++ b/src/components/calcite-filter/calcite-filter.e2e.ts
@@ -30,7 +30,7 @@ describe("calcite-filter", () => {
       await page.setContent("<calcite-filter></calcite-filter>");
       await page.evaluate(() => {
         const filter = document.querySelector("calcite-filter");
-        filter.data = [{ foo: "bar" }];
+        filter.items = [{ foo: "bar" }];
       });
     });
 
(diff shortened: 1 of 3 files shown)


driskull (approved):
👍

benelan:
> Would renaming data => `items` and `filteredData` => `filteredItems` or `filtered` be a better fit? If so, for this release, we can add them and deprecate `data` and remove in the next sprint since the filter was never publicly doc'd

I do like `items` and  `filteredItems` or `filtered` better

> Do we need a story + SDK entry for this? Filter is useful for a consistent filtering experience, but not sure what the examples/stories need to cover

Yeah we should add a component page for Filter. But as you mentioned, since it has never been publicly doc'd it doesn't need to happen right away

jcfranco:
Deprecated `data` in favor of new `items` and `filteredItems` props.

PR #3087, opened 2021-09-21
refactor(slider): add scale property, update styles to match Figma

caripizza:
**Related Issue:** #2457

## Summary
Adds `scale` property to calcite-slider per convention, and refactors styles for each scale to match Figma.

PR #3450, opened 2021-11-08
test(slider): check if latest Screener story actually checked in (TESTING)

caripizza:
**Related Issue:** n/a (testing the Screener failure that pr #3087 is currently showing, doesn't seem accurate)

## Summary
When I pull down master and check out the stories locally, I see that the slider story appears like so @ 1024 x 768px in Chrome:

image

However this is being picked up as a width change/[failure on my slider scales refactor pr](https://screener.io/v2/states/1leJJn.Esri-calcite-components/refs%2Fpull%2F3087%2Fmerge/1024x768/Chrome/componentscontrolsslider-histogram-0), trying to figure out why/if it's a false positive. (The only visual change to that story should be the tick color, not width.)


caripizza:
Closing, the updated slider story is definitely there based on this Screener pass.

Issue #3372, opened 2021-10-28
Bug: calcite-popover causes notice not to be full width


AdelheidF:
If there is a notice with width=full inside a popover it's not as wide as the popover.

![image](https://user-images.githubusercontent.com/44481292/139304579-0100095f-4d27-45d3-a80d-36cf65fcc7f0.png)

From Alan: popover's content node needs to have something like `flex: 1 0 auto` defined on it.

### Actual Behavior

notice with width=full is not taking full width of popover

### Expected Behavior

notice takes full width of popover

### Reproduction Steps and Sample

https://codepen.io/afreitag/pen/oNewOXR?editors=1000

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-68`

- [ ] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-<version>`

benelan:
verified on `next`

PR #3451, opened 2021-11-08
Elijbet/3372 fix popover to allow full width content

Elijbet:
**Related Issue:** #3372

## Summary

Adds a flex modification to the content on popover to allow for notice to be set to width= "full".

Before:
https://codepen.io/afreitag/pen/oNewOXR?editors=1000

After:
notice covering popover

### DIFF
--- a/src/components/calcite-popover/calcite-popover.scss
+++ b/src/components/calcite-popover/calcite-popover.scss
@@ -63,11 +63,11 @@
 }
 
 .content {
-  @apply flex
-  flex-col
+  @apply flex-col
   flex-no-wrap
   self-center
   h-full;
+  flex: 1 0 auto;
 }
 
 .calcite--rtl {


src/components/calcite-popover/calcite-popover.scss
```diff
@@ -63,11 +63,11 @@
 }
 
 .content {
-  @apply flex
-  flex-col
+  @apply flex-col
   flex-no-wrap
   self-center
   h-full;
+  flex: 1 0 auto;
```
asangma, line 70: What do you think about using `w-full` instead?

Elijbet: That works as well, is there a benefit to using one or the other?

asangma (approved):
Woot!

PR #3452, opened 2021-11-08
fix(slider): set the current min/max to the provided values to keep histogram range updated #2914

y0n4:
**Related Issue:** #2914

## Summary
  The histogram still gets misaligned when the `min` / `max` values are in the lesser range of the histogram data values. Whenever the `max` value is less than highest histogram data, the highlighted range would not match vertically. This also includes when the `min` value is higher than the lowest histogram data. This change can also handle min and max values that go over the histogram data as well

Before (top- misaligned max value, bottom- misaligned min value)
Screen Shot 2021-11-08 at 1 16 00 PM
Screen Shot 2021-11-08 at 1 16 30 PM
After
Screen Shot 2021-11-08 at 1 19 56 PM

Regarding to the comment from #2914 addresses histogram data to show at the right range place
Before
Screen Shot 2021-11-08 at 3 57 42 PM
After
Screen Shot 2021-11-08 at 4 00 04 PM


caripizza:
@y0n4 I'm unclear on how this is fixing the issue pointed out in https://github.com/Esri/calcite-components/issues/2914#issuecomment-957926448. Can you clarify for me, or add a test that asserts the fix?

src/components/calcite-slider/calcite-slider.tsx
```diff
@@ -53,7 +53,7 @@ export class CalciteSlider implements LabelableComponent, FormComponent {
   /** Indicates if a histogram is present */
   @Prop({ reflect: true, mutable: true }) hasHistogram = false;
 
-  /** Display a histogram above the slider */
+  /** A list of [x-axis, y-axis] points that illustrates the histogram chart within the slider min and max range */
```
caripizza, line 56: This definitely helps better describe the prop, thanks @y0n4. I have a small change request after looking at this again:
- slight verbiage change for clarity
- add `@see` JSDoc tag which we use to add markdown links to the type definition over in Calcite's Developer docs

Does this sound good to you @macandcheese and @benelan?

```tsx
  /**
   * List of x,y coordinates within the slider's min and max, displays above the slider track.
   * @see [DataSeries](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-graph/interfaces.ts#L5)
  */
```

caripizza (approved):

src/components/calcite-slider/calcite-slider.tsx
macandcheese: SGTM

PR #3470, opened 2021-11-10
chore(ci): log tags and add noVerify to commit on next release script

benelan:
**Related Issue:** NA

## Summary
Last run tried to publish on the current next release and we noticed that it was linting on the commit:
https://github.com/Esri/calcite-components/runs/4159041103?check_suite_focus=true#step:5:60


jcfranco (approved):
🔍 :shipit:

Issue #3443, opened 2021-11-05
Bug: [calcite-input] incrementing/decrementing number input with a decimal step sometimes results in a very long precision decimal


Elijbet:
### Actual Behavior

The input is giving a very long precision on decimal steps. Seems to be something that can happen when you’re adding and subtracting (using math) in JS. 

### Expected Behavior

The stepping should match the precision of the provided step. 

### Reproduction Steps and Sample

Increment/Decrement several times to observe the behavior here: https://codepen.io/elijbet/pen/gOxebyE

https://user-images.githubusercontent.com/19231036/140584264-2f524f5d-c500-4758-811d-d69554e9db0c.mov

### Relevant Info

Need to add a utility function that determines decimal places and set it so that stepping matches the precision of the provided step.  We just need to run toFixed in the calculated stepped value before setting it. 

The existing click test for checking incrementing/decrementing with a decimal value doesn’t catch this long value, as it only occasionally pops up and goes away. 

The existing step test can be updated to use mouse down/up or key down/up to catch the error on long hold.

benelan:
verified on `next`

PR #3473, opened 2021-11-10
fix(input): stepping precision to match the provided step

Elijbet:
**Related Issue:** #3443

## Summary

The stepping precision now matches the given step.  

Provided a utility function to determine decimal places. 

Since the long precision decimal occurs after multiple clicks, modified an existing step test to catch this on long hold.


driskull (approved):
🎉 🌮

PR #3478, opened 2021-11-10
chore(ci): fix github action push issue on protected branch

benelan:
**Related Issue:** NA

## Summary
- Added a new `ADMIN_TOKEN` secret with a PAT I generated. An admin token is needed in order to push to a protected branch
- Made the push `atomic` so it either all works or all fails


jcfranco (approved):
⚡🤘🏼 🤖:🤘🏼⚡

Issue #3081, opened 2021-09-21
Enhancement: allow customizing the animation/transition duration of components


jcfranco:
### Description

Allowing users to customize the animation/transition duration will help with:

1. Accessibility – end-users can modify the duration based on https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion
2. Testing – for screenshot testing,  it is crucial to be able to set the animation/transition duration to 0 to avoid unrelated failures

For now, this issue should focus on making sure all components refer to the same CSS prop to define their animation duration both directly (setting explicit value) and indirectly (via [Tailwind util class](https://github.com/Esri/calcite-components/blob/master/tailwind.config.js#L240-L245)).

cc @caripizza 

### Acceptance Criteria 

All component animations/transitions use `--calcite-animation-timing` to determine duration.

### Relevant Info 

https://github.com/Esri/calcite-components/pull/2607/ introduced the `--calcite-animation-timing` CSS var, but this is currently not used by our components. 

#### Which Component

All of the things! ✨

##### Using explicit duration value

* [ ] [calcite-tree-item.scss](https://github.com/Esri/calcite-components/blob/d811d1c8c6fd475afd3c1cfaab31415b7a135da7/src/components/calcite-tree-item/calcite-tree-item.scss#L85)
* [ ] [calcite-tip-manager.scss](https://github.com/Esri/calcite-components/blob/d7a42e59c4d8e216d497d5bdaf73055000485e61/src/components/calcite-tip-manager/calcite-tip-manager.scss#L66)
* [ ] [calcite-split-button.scss](https://github.com/Esri/calcite-components/blob/f5270f8085b8a08955e65cdc9fd2a1a39c7e5e0b/src/components/calcite-split-button/calcite-split-button.scss#L83)
* [ ] [calcite-slider.scss](https://github.com/Esri/calcite-components/blob/13cbc7c27d3ca088877b7e7f32b8abe8d819f520/src/components/calcite-slider/calcite-slider.scss#L335)
* [ ] [calcite-slider.scss](https://github.com/Esri/calcite-components/blob/13cbc7c27d3ca088877b7e7f32b8abe8d819f520/src/components/calcite-slider/calcite-slider.scss#L327)
* [ ] [calcite-slider.scss](https://github.com/Esri/calcite-components/blob/13cbc7c27d3ca088877b7e7f32b8abe8d819f520/src/components/calcite-slider/calcite-slider.scss#L103)
[...]

PR #3479, opened 2021-11-10
feat(alert): allow disabling animation and customizing animation/transition duration

anveshmekala:
**Related Issue:** #3081 

## Summary

This will allow to disable animations for `calcite-alert` when `--calcite-animation-timing-factor` is set to 0 and customize the duration of animation timing

### DIFF
--- a/src/components/calcite-alert/calcite-alert.scss
+++ b/src/components/calcite-alert/calcite-alert.scss
@@ -76,7 +76,6 @@
 
 :host {
   @apply block;
-  --calcite-animation-timing: 300ms;
   .container {
     @apply shadow-2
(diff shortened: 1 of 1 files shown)


macandcheese:
To clarify - this doesn't actually change the duration of the alert - it allows for changing the speed at which the alert animates in and out of view?

anveshmekala:
> To clarify - this doesn't actually change the duration of the alert - it allows for changing the speed at which the alert animates in and out of view?

Yup. This shouldn't affect the current animation/transition duration of the alert.

jcfranco (approved):
👍

Issue #3362, opened 2021-10-27
Enhancement: ListItem - add padding when no description


asangma:
### Description
When there is no description, list items are very small.
In order to retain good hit area, we should add some additional padding when there is no description.

### Acceptance Criteria 
Add 0.25rem top and bottom padding to `.content` or `.label` when `.label` is the only child.

### Relevant Info 
See demo example here: https://codepen.io/mtb-alan/pen/XWagXRG

#### Which Component
CalciteListItem

#### Example Use Case
Any time a list-item is used with only the label.

asangma:
@benelan I'd like to bust this one out in this sprint. Low risk, better UI.

benelan:
@asangma sure, I'm adding it to this sprint

asangma:
Whoop!

benelan:
verified on next

PR #3481, opened 2021-11-10
feat(list-item): add padding when there is only a label or only a description

asangma:
**Related Issue:** #3362

## Summary
Keeps list-item from being too small when there's only a label or description.

Accomplished by adding padding around `label` or `description` when they are `:only-child`.


driskull (approved):
Looks good. Is this covered by a screenshot test?

asangma:
Oh...good question. I'm not sure how to add that.

driskull:
Can you add a list-item with only a child to one of the list item stories?

asangma:
@driskull Does that look right?

PR #3482, opened 2021-11-10
feat(slider): allow disabling & customizing animation/transition duration

anveshmekala:
**Related Issue:** #3081 

## Summary
This will allow to disable animations for `calcite-slider` when `--calcite-animation-timing-factor` is set to `0` and customize the duration of animation timing


src/components/calcite-slider/calcite-slider.scss
```diff
@@ -35,6 +35,7 @@
 }
 
 :host {
+  --calcite-slider-internal-animation-timing: calc(250ms * var(--calcite-animation-timing-factor));
```
jcfranco, line 38: Since there's only a 50ms difference between the internal global vars, could this use either the `medium` or `slow` var? cc @macandcheese

src/components/calcite-slider/calcite-slider.scss
```diff
@@ -268,11 +271,11 @@
 }
 
 .tick__label--min {
-  transition: opacity 150ms;
+  transition: opacity var(--calcite-internal-animation-timing);
 }
 
 .tick__label--max {
-  transition: opacity 50ms;
+  transition: opacity calc(50ms * var(--calcite-animation-timing-factor));
```
jcfranco, line 278: Since there's only a 50ms difference between the internal global vars, could this use either the `quick` var? cc @macandcheese

jcfranco (approved):
Code looks good. Just wondering if we could use the internal vars if possible. If not, can you create a follow-up issue to update these? I'm assuming all durations will reference an internal CSS var.

src/components/calcite-slider/calcite-slider.scss
macandcheese: Yeah let's standardize to use the established ones here.

src/components/calcite-slider/calcite-slider.scss
macandcheese: Yeah let's standardize to use the established ones here - medium would be the suggestion here.

PR #3483, opened 2021-11-10
feat(scrim): allow disabling animation & customizing animation/transition duration

anveshmekala:
**Related Issue:** #3081 

## Summary

This will allow to disable animations for `calcite-scrim` when `--calcite-animation-timing-factor` is set to `0` and customize the duration of  animation timing


driskull (commented):
Looks good to me 👍

src/components/calcite-scrim/calcite-scrim.scss
```diff
@@ -27,7 +27,7 @@
   items-center
   flex-col
   justify-center;
-  animation: calcite-scrim-fade-in 250ms ease-in-out;
+  animation: calcite-scrim-fade-in calc(250ms * var(--calcite-animation-timing-factor)) ease-in-out;
```
jcfranco, line 30: Since there's only a 50ms difference between the internal global vars, could this use either the `medium` or `slow` var? cc @macandcheese

jcfranco (approved):
Code looks good. Just wondering if we could use the internal vars if possible. If not, can you create a follow-up issue to update these? I'm assuming all durations will reference an internal CSS var.

src/components/calcite-scrim/calcite-scrim.scss
macandcheese: Yeah let's standardize any outliers... would suggest slow since scrim can cover a large visual area but happy to adjust after we standardize if needed.

PR #3484, opened 2021-11-10
feat(radio-button-group-item): allow disabling & customizing animation/transition duration

anveshmekala:
**Related Issue:** #3081 

## Summary
This will allow to disable animations for `calcite-radio-button-group-item` when `--calcite-animation-timing-factor` is set to `0` and customize the duration of  animation timing


jcfranco (approved):
👍

PR #3485, opened 2021-11-10
feat(notice): allow disabling animation & customizing animation/transition duration

anveshmekala:
**Related Issue:** #3081 

## Summary

This will allow to disable animations for `calcite-notice` when `--calcite-animation-timing-factor` is set to `0` and customize the duration of animation/transition timing


jcfranco (approved):
👍🏼

PR #3487, opened 2021-11-10
feat(loader): allow disabling animation and customizing animation/transition duration

anveshmekala:
**Related Issue:** #3081

## Summary
This will allow to disable animations for `calcite-loader` when `--calcite-animation-timing-factor` is set to `0`  and customize the duration of animation timing


driskull (approved):
👍

Issue #3346, opened 2021-10-25
Bug: Block heading and summary don't seem to align correctly in RTL


asangma:
### Actual Behavior
In RTL, Block heading and summary don't seem to align correctly.
This seems to only happen when the Block is `collapsible`

![Screen Shot 2021-10-25 at 10 47 30 AM](https://user-images.githubusercontent.com/12503298/138744958-8f08db15-404f-4910-a6e8-de29600cf91e.png)

### Expected Behavior
Align correctly in RTL.

### Reproduction Steps and Sample

_Sample_: https://codepen.io/mtb-alan/pen/porepam?editors=1000

1. View the difference between the collapsible and not collapsible Blocks.

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-<version>`

- [ ] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-<version>`

benelan:
verified on next

PR #3490, opened 2021-11-11
fix(block): heading and summary now align in RTL

Elijbet:
**Related Issue:** #3346

## Summary

Heading and summary now align in RTL for when the Block is collapsible.

align-block-heading-and-summary-RTL


driskull (approved):
Looks good @Elijbet!

If you want bonus points, you could remove the `calcite--rtl` class and use logical properties instead :)

<sub>** bonus points are fictional</sub>

Elijbet:
> Looks good @Elijbet!
> 
> If you want bonus points, you could remove the `calcite--rtl` class and use logical properties instead :)
> 
> ** bonus points are fictional

Oh man, can I at least get a sticker?

src/components/calcite-block/calcite-block.scss
```diff
@@ -160,6 +160,9 @@ calcite-action-menu {
   .icon {
     @apply ml-0 mr-3;
   }
+  .toggle {
+    @apply text-right;
```
asangma, line 164: Hmmm...I think as @driskull it might be better to take this opportunity to refactor with logical properties for the margins.

For the text, we could change the `text-align` rule [here](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-block/calcite-block.scss#L54) to be `initial`.

asangma (commented):
Comment about the RTL selector and text-align.
I'm not blocking because I'll be OOO for a few days and don't want to hold you up.

driskull (approved):
Looks good! Just needs the RTL class and getElementDir utility removed from the block.tsx. Otherwise Good to merge!

money-nintendo

Issue #3491, opened 2021-11-11
Enhacement: Match animation timing global variable value with calcite base transition  & add scales for animation timing variable


anveshmekala:
### Description

Global transition variable is added in #2607  which allows to customize the duration time for animation/transitions which is currently set to `300ms` . Match `--calcite-animation-timing` global variable value with `calcite-base` [transition](https://github.com/Esri/calcite-base/blob/master/dist/_variables.scss#L2) value which is `150ms` . With the [audit](https://github.com/Esri/calcite-components/wiki/Animation-&-Transition-duration-Audit) of animation duration, it is clear that component have different values and we need to have different levels of animation duration variables.

### Acceptance Criteria 

1. `--calcite-animation-timing` value should be set to `150ms`
2. Make sure any existing components utilizing this var should be updated
3.  Add multiple levels of animation timing variable to support component across CC repo.
4.  List of tailwind UtilClasses using `--calcite-animation-timing`  :
- animate-in
- animate-in-down
- animate-in-up
- animate-in-scale
5. Update components utilizing above mentioned tail wind classes

### Relevant Info 
Refer to #3081 

#### Which Component
List of components utilizing tailwind util classes with `--calcite-animation-timing`:

- [ ]  [calcite-action-pad](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-action-pad/calcite-action-pad.scss#L3)
- [ ] [calcite-block](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-block/calcite-block.scss#L144)
- [ ] [calcite-shell](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-shell/calcite-shell.scss#L91)
- [ ] [calcite-shell-center-row](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-shell-center-row/calcite-shell-center-row.scss#L27)

benelan:
@macandcheese can you please help verify this installed design issue?

macandcheese:
Looks great - thanks for the work on this @anveshmekala. cc @asangma for reference.

benelan:
verified on next!

PR #3492, opened 2021-11-11
feat(animation): add animation timing factor global var and match animation timing with calcite base value

anveshmekala:
**Related Issue:** #3491 

## Summary

This will make sure global animation timing class `--calcite-animation-timing` value matches with the `calcite-base` [$transition](https://github.com/Esri/calcite-base/blob/master/dist/_variables.scss#L2) variable value and brings uniformity among components in calcite repo.

### DIFF
--- a/src/assets/styles/_animation.scss
+++ b/src/assets/styles/_animation.scss
@@ -41,7 +41,7 @@
 }
 
 :root {
-  --calcite-animation-timing: 300ms;
+  --calcite-animation-timing: 150ms;
 }
 
 // allows animation trigger via JS by adding class to element
--- a/src/assets/styles/_animation.scss
+++ b/src/assets/styles/_animation.scss
@@ -40,8 +40,9 @@
   }
 }
 
-:root {
-  --calcite-animation-timing: 150ms;
+* {
+  --calcite-animation-factor: 1;
+  --calcite-animation-timing: calc(var(--calcite-animation-factor) * 150ms);
 }
 
 // allows animation trigger via JS by adding class to element
--- a/src/assets/styles/_animation.scss
+++ b/src/assets/styles/_animation.scss
@@ -43,9 +43,9 @@
 :root {
   --calcite-animation-timing-factor: var(--animation-timing-factor, 1);
   --calcite-animation-timing: calc(150ms * var(--calcite-animation-timing-factor));
-  //   --calcite-animation-timing-100 : calc(100ms *var(--calcite-animation-timing-factor) );
-  //   --calcite-animation-timing-200 : calc(200ms * var(--calcite-animation-timing-factor));
-  //   --calcite-animation-timing-300 : calc(300ms *var(--calcite-animation-timing-factor));
+  --calcite-animation-timing-quick: calc(100ms * var(--calcite-animation-timing-factor));
+  --calcite-animation-timing-medium: calc(200ms * var(--calcite-animation-timing-factor));
+  --calcite-animation-timing-slow: calc(300ms * var(--calcite-animation-timing-factor));
 }
 
 // allows animation trigger via JS by adding class to element
--- a/src/assets/styles/_animation.scss
+++ b/src/assets/styles/_animation.scss
@@ -41,16 +41,11 @@
 }
 
 :root {
-  // --calcite-animation-timing:  150ms;
-  // --calcite-animation-timing-50  : calc(var(--calcite-animation-timing)* 0.34)
-  // --calcite-animation-timing-100 : calc(var(--calcite-animation-timing) * 0.67)
-  // --calcite-animation-timing-180 : calc(var(--calcite-animation-timing)* 1.2)
-  // --calcite-animation-timing-200 : calc(var(--calcite-animation-timing)* 1.34)
-  // --calcite-animation-timing-300 : calc(var(--calcite-animation-timing) * 2);
-}
-
-@mixin calciteAnimationTiming($animationFactor: var(--calcite-animation-factor, 1)) {
-  animation-duration: calc(var(--calcite-animation-timing, 150ms) * $animationFactor);
+  --calcite-animation-timing-factor: var(--animation-timing-factor, 1);
+  --calcite-animation-timing: calc(150ms * var(--calcite-animation-timing-factor));
+  //   --calcite-animation-timing-100 : calc(100ms *var(--calcite-animation-timing-factor) );
+  //   --calcite-animation-timing-200 : calc(200ms * var(--calcite-animation-timing-factor));
+  //   --calcite-animation-timing-300 : calc(300ms *var(--calcite-animation-timing-factor));
 }
 
 // allows animation trigger via JS by adding class to element
--- a/src/assets/styles/_animation.scss
+++ b/src/assets/styles/_animation.scss
@@ -40,9 +40,17 @@
   }
 }
 
-* {
-  --calcite-animation-factor: 1;
-  --calcite-animation-timing: calc(var(--calcite-animation-factor) * 150ms);
+:root {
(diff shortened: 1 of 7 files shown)


benelan:
> /merge/master

It's `/merge master`

src/tests/globalStyles.e2e.ts
```diff
@@ -44,7 +44,26 @@ describe("global styles", () => {
...
           return window.getComputedStyle(document.body).getPropertyValue("--calcite-animation-timing");
         })
-      ).toBe(" 300ms");
+      ).toBe(" 150ms");
+    });
+
+    it("should set animation duration to 0ms", async () => {
+      const page = await newE2EPage({
+        html: snippet
+      });
```
jcfranco, line 53: For test stability, can you add a `await page.waitForChanges()` call here?

src/tests/globalStyles.e2e.ts
```diff
@@ -32,7 +32,7 @@ describe("global styles", () => {
             opacity: opacity
           };
```
jcfranco, line 33: @caripizza Do these tests need to use `notice`? I originally thought they were testing component-specific animation, but it looks like the tests focus on asserting values on the global classes.

src/components/calcite-action-pad/calcite-action-pad.scss
```diff
@@ -1,5 +1,6 @@
 :host {
   @extend %component-host;
+  --calcite-animation-timing: 300ms;
```
jcfranco, line 3: Aside from focus timing that will go away, I thought all components would rely on the same global timing var. Am I missing something? cc @caripizza @macandcheese

caripizza: @jcfranco no I think you have it right. It's just that a few components which reference these global animation utility classes (ex `animate-in-up`) seem like they need the longer 300ms speed for their animations (not the new 150ms speed from updated the global var value). That's why I recommended that Anvesh just reset their values directly above the animation class as he's doing here.

jcfranco: >  It's just that a few components which reference these global animation utility classes (ex animate-in-up) seem like they need the longer 300ms speed for their animations

I see. What are the downsides to having all of them use the global var? Would some animations look off? 

We need a way to allow users to override the global variable to tweak or even disable animations. Having component-specific overrides does defeats the purpose. I think @anveshmekala brought this up before. Looping in @asangma + @macandcheese for suggestions.

caripizza: These utility classes are using the global var already - is that your concern?
https://github.com/Esri/calcite-components/blob/master/tailwind.config.js#L130

src/tests/globalStyles.e2e.ts
caripizza: Nope, please remove or improve them (these were just tests I added with the [original demo page](https://github.com/Esri/calcite-components/blob/master/src/demos/animation/animation.html)).

It seems more important to test that animations can be disabled globally by setting the timing var to `0s`.

src/components/calcite-action-pad/calcite-action-pad.scss
jcfranco: My concern having a mixed usage of the global var and some components overriding the duration var. I'd like to have users override a single global var for all animation/transition duration. Let me know if this makes sense. We can have a quick chat later today if I'm misunderstanding the approach.

src/tests/globalStyles.e2e.ts
```diff
@@ -32,19 +31,78 @@ describe("global styles", () => {
...
     });
 
-    it("should have initial --calcite-animation-timing CSS Custom Property value", async () => {
-      const page = await newE2EPage({ html: snippet });
-      expect(
-        await page.evaluate(() => {
-          return window.getComputedStyle(document.body).getPropertyValue("--calcite-animation-timing");
-        })
-      ).toBe(" 300ms");
+    it("should set animation duration to 0ms", async () => {
```
jcfranco, line 40: I'm not sure what this test is covering. `--calcite-animation-timing` is internal, so no need to test it, right? Maybe I'm missing something.

src/tests/globalStyles.e2e.ts
```diff
@@ -32,19 +31,78 @@ describe("global styles", () => {
...
+      const page = await newE2EPage({
+        html: snippet
+      });
+      await page.waitForChanges();
+      await page.$eval("calcite-notice", (element: any) => {
```
jcfranco, line 45: Type the element and avoid using `any` types whenever possible. This applies to similar functions.

src/assets/styles/_animation.scss
```diff
@@ -41,7 +41,11 @@
 }
 
 :root {
-  --calcite-animation-timing: 300ms;
+  --calcite-animation-timing-factor: var(--animation-timing-factor, 1);
```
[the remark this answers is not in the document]
anveshmekala: Make sense. Will swap them.

src/assets/styles/_animation.scss
```diff
@@ -41,7 +41,11 @@
...
+  --calcite-animation-timing: calc(150ms * var(--calcite-animation-timing-factor));
```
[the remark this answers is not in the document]
anveshmekala: These are going to be internal for now. We let user's control the animation duration with `--calcite-animation-timing-factor` .

src/tests/globalStyles.e2e.ts
anveshmekala: This test was added prior to the audit. Might have overlooked and  this test makes more sense now considering `--calcite-animation-factor` will be the public.

anveshmekala: never mind. This test should be removed.

src/assets/styles/_animation.scss
```diff
@@ -41,7 +41,11 @@
 }
 
 :root {
-  --calcite-animation-timing: 300ms;
+  --animation-timing-factor: var(--calcite-animation-timing-factor, 1);
```
jcfranco, line 44: Per our conversation w/ @macandcheese, all root internal CSS props should be namespaced w/ `--calcite-internal-<prop>`. Please rename the following:

* `--animation-timing-factor` -> `--calcite-internal-animation-timing-factor`
* `--calcite-animation-timing-quick` -> `--calcite-internal-animation-timing-quick`
* `--calcite-animation-timing-medium` -> `--calcite-internal-animation-timing-medium`
* `--calcite-animation-timing-slow` -> `--calcite-internal-animation-timing-slow`

Note: `--calcite-animation-timing` stays the same as it is public.

src/tests/globalStyles.e2e.ts
```diff
@@ -32,19 +31,58 @@ describe("global styles", () => {
...
     });
 
-    it("should have initial --calcite-animation-timing CSS Custom Property value", async () => {
-      const page = await newE2EPage({ html: snippet });
-      expect(
-        await page.evaluate(() => {
-          return window.getComputedStyle(document.body).getPropertyValue("--calcite-animation-timing");
-        })
-      ).toBe(" 300ms");
+    it("should set animation duration to 0ms when --animation-timing-factor set to zero", async () => {
```
jcfranco, line 40: For completeness, can you add a test that asserts on the default transition duration? You already cover disabling animation and not allowing component-level overrides.

jcfranco (approved):
Once the remaining items are covered, this is good to merge. Nice stuff, @anveshmekala! 🎉

jcfranco:
@anveshmekala I think the PR title may need updating since this also introduces the animation timing factor variable.

PR #3507, opened 2021-11-13
chore(ci): use admin token to fix stalling pr checks

benelan:
**Related Issue:** NA

## Summary
After some troubleshooting, it appears that gh actions won't run on events triggered by bots. As a workaround I changed the default (bot) token to a PAT to trick the actions into thinking an actual user triggered the event.


jcfranco (approved):
💥

Issue #3212, opened 2021-10-12
Bug: Nested Calcite Accordions incorrectly toggle parent


Soreon:
### Actual Behavior

When you nest two accordion components with the parent being in `selection-mode="single"`, clicking on one of the items of the child accordion toggle the parent `<calcite-accordion-item>` instead of toggling the child.

If you remove `selection-mode="single"` on the parent accordion, this issue dissapear but the child accordion is not clickable.

### Expected Behavior

Clicking on the child accordion should not toggle the parent.

### Reproduction Steps and Sample
https://codepen.io/jcfranco/pen/KKvRbYO?editors=1000

### Relevant Info

This issue seems to have been fixed at some point : https://github.com/Esri/calcite-components/issues/330

_Version_: `@esri/calcite-components@1.0.0-beta.66`

- [x] CDN
- [ ] NPM package

Elijbet:
@Soreon Could you please provide steps on how to reproduce the issue? Eg which accordion-item do I need to click to observe the unwanted behavior?

jcfranco:
@Elijbet I updated the repro case to make it clearer to see. If you expand one accordion group and then expand the other one, the first one closes.

benelan:
verified on next

PR #3511, opened 2021-11-15
fix(accordion-item): clicking on an accordion with selection-mode=single no longer toggles unrelated accordions with the same selection mode

Elijbet:
**Related Issue:** #3212

## Summary

Clicking on the nested child accordion-items does not toggle the parent when selection-mode="single".
The accordion-item now checks if calciteAccordionChange event was emitted from its own parent.


src/components/calcite-accordion/calcite-accordion.e2e.ts
```diff
@@ -130,6 +130,41 @@ describe("calcite-accordion", () => {
     expect(await item3Content.isVisible()).toBe(true);
   });
 
+  it("clicking on the nested child accordion-item does not toggle the parent in single selection mode", async () => {
+    const page = await newE2EPage();
+    await page.setContent(`
```
jcfranco, line 135: Can you use the [`html` helper](https://github.com/Esri/calcite-components/blob/master/src/tests/utils.ts#L119-L137) to have Prettier handle formatting of embedded JS?

```ts
await page.setContent(html`
// ...
```

src/components/calcite-accordion/calcite-accordion.e2e.ts
```diff
@@ -130,6 +130,41 @@ describe("calcite-accordion", () => {
...
+    <calcite-accordion selection-mode="single" id="first">
+    ${accordionContent}
+    </calcite-accordion>
+    <calcite-accordion selection-mode="single" id="second">
+    ${accordionContent}
+    </calcite-accordion>`);
+    const firstAccordion = await page.find("calcite-accordion[id='first']");
+    expect(firstAccordion).toEqualAttribute("selection-mode", "single");
```
jcfranco, line 143: The test set up implies this, so we don't need to assert on this.

src/components/calcite-accordion/calcite-accordion.e2e.ts
```diff
@@ -130,6 +130,41 @@ describe("calcite-accordion", () => {
...
+
+    const item1FirstAccordion = await firstAccordion.find("calcite-accordion-item[id='1']");
+    const item1FirstAccordionContent = await firstAccordion.find(
+      "calcite-accordion[id='first'] > calcite-accordion-item[id='1']"
+    );
+    await item1FirstAccordion.click();
+    expect(item1FirstAccordion).toHaveAttribute("active");
+    expect(await item1FirstAccordionContent.isVisible()).toBe(true);
```
jcfranco, line 151: Can you assert on the `active` property instead of visibility? We already have tests that cover visibility based on the `active` prop.

src/components/calcite-accordion/calcite-accordion.e2e.ts
```diff
@@ -130,6 +130,41 @@ describe("calcite-accordion", () => {
...
+    </calcite-accordion>`);
+    const firstAccordion = await page.find("calcite-accordion[id='first']");
+    expect(firstAccordion).toEqualAttribute("selection-mode", "single");
+
+    const item1FirstAccordion = await firstAccordion.find("calcite-accordion-item[id='1']");
+    const item1FirstAccordionContent = await firstAccordion.find(
+      "calcite-accordion[id='first'] > calcite-accordion-item[id='1']"
+    );
+    await item1FirstAccordion.click();
+    expect(item1FirstAccordion).toHaveAttribute("active");
```
jcfranco, line 150: Can you assert on the property vs the attribute?

jcfranco (changes requested):
Awesome! 👏🏼 

I just had a few comments related to the tests.

src/components/calcite-accordion-item/calcite-accordion-item.tsx
```diff
@@ -174,6 +174,9 @@ export class CalciteAccordionItem {
   updateActiveItemOnChange(event: CustomEvent): void {
     this.requestedAccordionItem = event.detail
       .requestedAccordionItem as HTMLCalciteAccordionItemElement;
+    if (this.el.parentNode !== this.requestedAccordionItem.parentNode) {
```
driskull, line 177: Is there any chance `el.parentNode` or `requestedAccordionItem` could be undefined? I wonder if we need `this.requestedAccordionItem?.parentNode`.

driskull (approved):
Code changes and test look good to me!

src/components/calcite-accordion/calcite-accordion.e2e.ts
```diff
@@ -130,6 +131,26 @@ describe("calcite-accordion", () => {
     expect(await item3Content.isVisible()).toBe(true);
   });
 
+  it("clicking on the nested child accordion-item does not toggle the parent in single selection mode", async () => {
+    const page = await newE2EPage({
+      html: html`<calcite-accordion selection-mode="single" id="first"> ${accordionContent} </calcite-accordion>
+        <calcite-accordion selection-mode="single" id="second"> ${accordionContent} </calcite-accordion>`
+    });
```
jcfranco, line 138: For test stability, can you add a `await page.waitForChanges()` after creating the page? By contrast, when you call `page.setContent()` it calls `waitForChanges` internally.

jcfranco (approved):
Aside from one test comment, this looks good to merge! 🎉🌮:t-rex:

src/components/calcite-accordion-item/calcite-accordion-item.tsx
jcfranco: Good catch.

jcfranco: Actually, looking at the event emitting, it looks like the item would always be there.

src/components/calcite-accordion/calcite-accordion.e2e.ts
Elijbet: Thanks for these tips, these are great!

src/components/calcite-accordion-item/calcite-accordion-item.tsx
```diff
@@ -174,6 +174,9 @@ export class CalciteAccordionItem {
   updateActiveItemOnChange(event: CustomEvent): void {
     this.requestedAccordionItem = event.detail
       .requestedAccordionItem as HTMLCalciteAccordionItemElement;
+    if (this.el.parentNode !== this.requestedAccordionItem?.parentNode) {
```
jcfranco, line 177: @Elijbet Can you double check if there's a scenario where `requestedAccordionItem` could be undefined? I'd like to avoid unnecessary checks if possible.

Elijbet: Sure.

src/components/calcite-accordion-item/calcite-accordion-item.tsx
Elijbet: I looked at the emitter like you mentioned, and the requestedAccordion item is the el, which is the host, correct? That wouldn't be undefined.

PR #3512, opened 2021-11-15
chore(ci): bump node versions to latest lts in gh actions

benelan:
**Related Issue:** #3509

## Summary
Some of the gh actions were hardcoded as node version 14. I changed them to `lts/*` which uses the latest LTS version, in this case v16


jcfranco (approved):
💥

PR #3514, opened 2021-11-15
chore(ci): allow users to run merge action in prs created by bot

benelan:
**Related Issue:** NA

## Summary
I had previously restricted the action to only allow a user to run the merge action on their own PRs. However this means no one could merge the PRs created by github actions, so I added a case for that


jcfranco (approved):
👍

Issue #3366, opened 2021-10-27
Bug: calcite-alert stays open beyond animation end when another alert is added


noahmulfinger:
### Actual Behavior

When an auto-dismiss calcite alert is open and second alert is is added, there seems to be a delay between the progress bar animation ending and the first alert closing. You can see in this video:

https://user-images.githubusercontent.com/6539297/139159791-544070bc-0eec-456f-888c-2fb36a02bdf7.mov

### Expected Behavior

I would expect the alert close behavior to be the same regardless of when a new alert has been added.

### Reproduction Steps and Sample

_Sample_:

1. Go to https://codepen.io/noahmulfinger/pen/rNzwLRK?editors=1010
2. Click on the `Add alerts` button and you will see the same behavior as in the above video.

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-68`

- [x] CDN
- [x] NPM package

benelan:
@macandcheese can you take a look at this one for design? I don't think there should be a close animation if there is another alert already queued. There was no close animation before this fix, we just needed to remove the delay before the progress bar started moving. I used Noah's sample and changed the version to `next`.

macandcheese:
So, this is kind of a weird one. Because an individual alert animates in, the alert "object" should animate out before another one can replace it. If we didn't do that, we'd be replacing the content of the alert vs. replacing the alert itself, which the animation infers.

That said - it's not a great interaction. Ideally when we handle multiple queued alerts with this update: https://github.com/Esri/calcite-components/issues/2835, we can have queued alerts animate in "behind" and then the animation for the queued alert becoming active will be more of a "move to front", with the dismissed alert animating out.

I realize that probably doesn't make sense, the image in the linked issue may help. For now I think we can leave it until that's implemented.

benelan:
Thanks for looking Adam. This is verified on `next` then.

PR #3525, opened 2021-11-16
fix(alert): remove clearTimeOut that delays the alert to close on autoDismiss mode #3366

y0n4:
**Related Issue:** #3366

## Summary
The current alert was delayed in closing itself out if there was another alert on queue. I believe it was the clearTimeOut clearing the previous timer which was never able to get to the callback to mark the current alert to close on time

This also addresses whenever there's a change of autoDismissDuration which will sync up with the animation of the speed matching at the duration seconds while in the middle of showing the alert on autoDismiss.

How to test:
select alerts with autoDismiss valid prop and add them on queue


src/components/calcite-alert/calcite-alert.tsx
```diff
@@ -275,12 +273,8 @@ export class CalciteAlert {
   private determineActiveAlert(): void {
     if (this.queue?.[0] === this.el) {
       this.openAlert();
-      clearTimeout(this.autoDismissTimeout);
```
driskull, line 278: I think there should be some cases where we still need to clear the timeout.

- When the component is destroyed
- If the autoDismissDuration is changed?

driskull (changes requested):
Looking good @y0n4. 🏆 

Can we add a test for this? 🤖

src/components/calcite-alert/calcite-alert.tsx
```diff
@@ -86,12 +86,31 @@ export class CalciteAlert {
   /** specify the scale of the button, defaults to m */
   @Prop({ reflect: true }) scale: Scale = "m";
 
+  @Prop() id: string;
```
driskull, line 89: I don't think we want to add an id prop here. Can we use the element itself as a lookup in a map? See the tooltip manager for implementation details.

```
hoverTimeouts: WeakMap<HTMLCalciteTooltipElement, number> = new WeakMap();
```

```
if (hoverTimeouts.has(tooltip)) {
      window.clearTimeout(hoverTimeouts.get(tooltip));
      hoverTimeouts.delete(tooltip);
    }
```

src/components/calcite-alert/calcite-alert.tsx
```diff
@@ -211,8 +176,10 @@ export class CalciteAlert {
   @Listen("calciteAlertRegister", { target: "window" })
   alertRegister(): void {
     if (this.active && !this.queue.includes(this.el as HTMLCalciteAlertElement)) {
+      console.log("id:", this.id);
```
driskull, line 179: remove console log here :)

src/components/calcite-alert/calcite-alert.tsx
```diff
@@ -273,8 +248,16 @@ export class CalciteAlert {
...
-      if (this.autoDismiss) {
-        window.setTimeout(() => this.closeAlert(), DURATIONS[this.autoDismissDuration]);
+      if (this.autoDismiss && this.queueIds.includes(this.el.id)) {
+        if (!this.timeoutIds[this.queueIds.indexOf(this.el.id)]) {
+          this.trackTimer = new Date();
+          this.autoDismissTimeout = window.setTimeout(
+            () => this.closeAlert(),
+            DURATIONS[this.autoDismissDuration]
+          );
+          console.log("set timer...", this.autoDismissTimeout);
```
driskull, line 258: remove console log here :)

src/components/calcite-alert/calcite-alert.tsx
```diff
@@ -283,6 +266,7 @@ export class CalciteAlert {
 
   /** close and emit the closed alert and the queue */
   private closeAlert = (): void => {
+    console.log("closing...", this.autoDismissTimeout);
```
driskull, line 269: remove console log here :)

src/components/calcite-alert/calcite-alert.tsx
```diff
@@ -112,63 +131,9 @@ export class CalciteAlert {
...
-            <slot name={SLOTS.link} />
-          </div>
-          {queueCount}
-          {!autoDismiss ? closeButton : null}
-          {active && !queued && autoDismiss ? progress : null}
-        </div>
-      </Host>
-    );
+  disconnectedCallback(): void {
+    // clear all timeout ids in array
```
driskull, line 135: remove console log here :)

Also needs to clear all timeouts.

driskull (changes requested):
Getting there @y0n4. Can we use a map to lookup the timeout instead of an id? The id is already a global attribute so we shouldn't add a property. We can lookup using a map and the element.

y0n4:
Sorry I should've put this in draft mode as I knew I was planning to push additional changes, all of these are resolved now. I scrapped the id idea.

src/components/calcite-alert/calcite-alert.tsx
y0n4: There will only be one timeout id stored now

src/components/calcite-alert/calcite-alert.tsx
```diff
@@ -92,6 +92,19 @@ export class CalciteAlert {
     this.requestedIcon = setRequestedIcon(StatusIcons, this.icon, this.color);
   }
 
+  @Watch("autoDismissDuration")
+  updateDuration(newDuration: string, prevDuration: string): void {
+    if (this.autoDismiss && prevDuration !== newDuration) {
+      if (this.autoDismissTimeout) {
```
driskull, line 98: Can you combine these if statements? For better legibility, we can avoid nested if statements.

driskull (approved):
👍 Looks good @y0n4 !

src/components/calcite-alert/calcite-alert.tsx
```diff
@@ -92,6 +92,17 @@ export class CalciteAlert {
     this.requestedIcon = setRequestedIcon(StatusIcons, this.icon, this.color);
   }
 
+  @Watch("autoDismissDuration")
+  updateDuration(newDuration: string, prevDuration: string): void {
+    if (this.autoDismiss && prevDuration !== newDuration && this.autoDismissTimeout) {
```
jcfranco, line 97: You can drop `prevDuration !== newDuration` since it will always be true. Watchers only get invoked when the new and old values are different.

src/components/calcite-alert/calcite-alert.e2e.ts
```diff
@@ -70,6 +70,38 @@ describe("calcite-alert", () => {
     expect(icon).not.toBeNull();
   });
 
+  it("closes on time based on alert duration", async () => {
+    const page = await newE2EPage();
+    await page.setContent(`
+    <div>
```
jcfranco, line 76: Please use the [`html` helper](https://github.com/Esri/calcite-components/blob/master/src/tests/utils.ts#L119-L137) to have Prettier format embedded HTML:

```js
await page.setContent(html`
    <div>
      ...
`);
```

src/components/calcite-alert/calcite-alert.tsx
```diff
@@ -275,8 +294,8 @@ export class CalciteAlert {
   private determineActiveAlert(): void {
     if (this.queue?.[0] === this.el) {
       this.openAlert();
-      clearTimeout(this.autoDismissTimeout);
-      if (this.autoDismiss) {
+      if (this.autoDismiss && !this.autoDismissTimeout) {
+        this.trackTimer = new Date();
```
jcfranco, line 298: To simplify, you can use `Date.now()` instead of creating a dates + calling `valueOf`.

src/components/calcite-alert/calcite-alert.tsx
```diff
@@ -289,6 +308,9 @@ export class CalciteAlert {
 
   /** close and emit the closed alert and the queue */
   private closeAlert = (): void => {
+    if (this.autoDismissTimeout) {
+      this.autoDismissTimeout = null;
```
jcfranco, line 312: Unless I'm missing something, you could always clear the timeout here without checking.

src/components/calcite-alert/calcite-alert.tsx
```diff
@@ -257,10 +274,12 @@ export class CalciteAlert {
   /** the slotted alert link child element  */
   private alertLinkEl?: HTMLCalciteLinkElement;
 
-  private autoDismissTimeout: number;
+  private autoDismissTimeout: number = null;
```
jcfranco, line 277: For clarity, please rename this to `autoDismissTimeoutId`.

jcfranco (commented):
Apologies for the late review. Most items can be tackled in a refactor PR since they wouldn't affect the behavior.

src/components/calcite-alert/calcite-alert.tsx
```diff
@@ -112,6 +123,12 @@ export class CalciteAlert {
     this.alertLinkEl = this.el.querySelectorAll("calcite-link")[0] as HTMLCalciteLinkElement;
   }
 
+  disconnectedCallback(): void {
+    if (this.autoDismissTimeout) {
+      window.clearTimeout(this.autoDismissTimeout);
```
jcfranco, line 128: I missed one more, you can clear the timeout without checking if it exists. This will end in a noop if there is no associated timeout ID.

src/components/calcite-alert/calcite-alert.tsx
driskull: nice catch 🏈

src/components/calcite-alert/calcite-alert.tsx
driskull: nice catch 🏈

src/components/calcite-alert/calcite-alert.tsx
driskull: nice catch 🏈

src/components/calcite-alert/calcite-alert.tsx
driskull: nice catch 🏈

Issue #782, opened 2020-07-29
Enhancement: calcite-input type:number, holding arrow should speed up the counter


AdelheidF:
### Description
If I use the HTML input element and hold down the mouse on one of the arrow keys the number keeps changing faster and faster. With calcite-input I have to click for every step increase.

In my test case I have values 0 to 100 for a percentage value and it's annoying to have to keep clicking.

<img src=https://user-images.githubusercontent.com/44481292/88857316-e8e21e80-d1aa-11ea-8f04-c13bce3cc84f.png width=200/>

<img src=https://user-images.githubusercontent.com/44481292/88857454-28a90600-d1ab-11ea-9812-fa082bf6b695.png width=200/>

macandcheese:
This should also occur when holding down arrow up / down keys, in addition to holding down mouse on the increment / decrement ui buttons.

driskull:
This issue getting stale. Who can we assign it to?

benelan:
In beta.69 the number increases but at a slow rate. Should we increase it? The native input's rate of increase is much faster.
https://codepen.io/benesri/pen/ExvRoYJ?editors=1000

cc @macandcheese

macandcheese:
@benelan Agreed - the speed of incrementing / decrementing with the arrow keys or the buttons should match a native input element where incrementing / decrementing "accelerates". 

The codepen above also has a strange interaction where the cursor occasionally flips to the start of the input.

benelan:
@Elijbet I'm assigning the issue back to you, can you try to get the rate of increase/decrease to match the [native input element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number). Thanks!

Elijbet:
Cool, I'll work on the cursor and speed issue next.

Elijbet:
> @Elijbet I'm assigning the issue back to you, can you try to get the rate of increase/decrease to match the [native input element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number). Thanks!

I researched and didn't see an easy way of finding out the exact rate of change, but I took a video and tracked the time to be about 100ms. Would an estimate like that work?

AdelheidF:
Looks good to me now.

benelan:
verified on `next`

PR #3527, opened 2021-11-17
fix(input): prevent default behavior of cursor jumping to the beginning of input, match rate of native input type=number (100ms)

Elijbet:
**Related Issue:** #782

## Summary

Cursor behavior: 
The default behavior for input is to move the cursor to the beginning of the input when pressing Arrow Up. This results in cursor jumping back and forth as the value changes. The code now prevents this default behavior. 

Rate to match native input:
The input behavior now imitates the native HTML input type=“number”. The rate of increment/decrement is matched to be 100ms.
Now that the interval is shorter, pressing down too hard would result in incrementing/decrementing twice, which is why a local var is used as a flag to ignore the first interval call. It should start only if the user really intends to hold. 

e2e
This is followed with an update in count of emitted events in previous e2e tests.

### DIFF
--- a/src/components/calcite-input/calcite-input.tsx
+++ b/src/components/calcite-input/calcite-input.tsx
@@ -260,8 +260,6 @@ export class CalciteInput implements LabelableComponent, FormComponent {
 
   private nudgeNumberValueIntervalId;
 
-  private nudgeNumberValueIntervalCount: number;
-
   //--------------------------------------------------------------------------
   //
   //  State
@@ -525,12 +523,12 @@ export class CalciteInput implements LabelableComponent, FormComponent {
     const valueNudgeDelayInMs = 100;
 
     this.incrementOrDecrementNumberValue(direction, inputMax, inputMin, nativeEvent);
-    this.nudgeNumberValueIntervalCount = 0;
+    let nudgeNumberValueIntervalCount = 0;
     this.nudgeNumberValueIntervalId = setInterval(() => {
-      this.nudgeNumberValueIntervalCount !== 0
+      nudgeNumberValueIntervalCount !== 0
         ? this.incrementOrDecrementNumberValue(direction, inputMax, inputMin, nativeEvent)
         : null;
-      this.nudgeNumberValueIntervalCount++;
+      nudgeNumberValueIntervalCount++;
     }, valueNudgeDelayInMs);
   };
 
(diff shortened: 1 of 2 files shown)


src/components/calcite-input/calcite-input.tsx
```diff
@@ -365,6 +367,9 @@ export class CalciteInput implements LabelableComponent, FormComponent {
   //--------------------------------------------------------------------------
 
   keyDownHandler = (event: KeyboardEvent): void => {
+    if (event.key === "ArrowUp") {
```
jcfranco, line 370: For readability, could you add a small comment on why this only is needed for the up key?

src/components/calcite-input/calcite-input.tsx
```diff
@@ -516,17 +521,20 @@ export class CalciteInput implements LabelableComponent, FormComponent {
...
+    const valueNudgeDelayInMs = 100;
 
     this.incrementOrDecrementNumberValue(direction, inputMax, inputMin, nativeEvent);
 
-    this.nudgeNumberValueIntervalId = setInterval(() => {
-      this.incrementOrDecrementNumberValue(direction, inputMax, inputMin, nativeEvent);
-    }, valueNudgeDelayInMs);
+    this.nudgeNumberValueTimeoutId = setTimeout(() => {
+      this.nudgeNumberValueIntervalId = setInterval(() => {
+        this.incrementOrDecrementNumberValue(direction, inputMax, inputMin, nativeEvent);
```
jcfranco, line 530: Instead of the wrapping timeout, would it be simpler to use a local var as a flag to ignore the first interval call?

src/components/calcite-input/calcite-input.tsx
```diff
@@ -516,17 +521,20 @@ export class CalciteInput implements LabelableComponent, FormComponent {
...
+      }, valueNudgeDelayInMs);
+    }, 100);
```
jcfranco, line 532: Ignore this comment if the timeout gets removed, but can you reuse the delay variable here? We want to avoid magic numbers whenever possible.

jcfranco (approved):
Had one suggestion for simplifying the timeout behavior. Aside from that, this LGTM!

driskull (approved):
Looks good to me! 🎉

Issue #2770, opened 2021-08-09
Enhancement: Resize calcite-panel with click-and-drag


bpatterson88:
### Description

Allow users to resize calcite-panel with click and drag.

### Acceptance Criteria 

- User can click and drag to resize calcite-panel width

### Relevant Info 

Changing the width of the panel through width-scale property (M->L) doesn't leave enough room for the dashboard preview.

#### Which Component

`calcite-panel`

#### Example Use Case

ArcGIS Dashboards is planning to place calcite-tree in calcite-panel. Users will be able to add items to calcite-tree, potentially creating a tree that is too wide for calcite-panel, causing a breakdown in the UI. It would be very nice if users could resize calcite-panel by clicking and dragging the right border. 

![Screen Shot 2021-08-09 at 3 35 05 PM](https://user-images.githubusercontent.com/15875886/128782796-d48e4934-1058-48b7-9296-fb1a6385295e.png)

This interaction is pretty common in similar UIs:

![figma-panel-resize](https://user-images.githubusercontent.com/15875886/128783592-1bde3a47-1b8d-4773-9650-010a5466ee66.gif)

![Untitled](https://user-images.githubusercontent.com/15875886/128783649-9b266354-9fb2-4aee-aa3a-31e8b9e09832.gif)

bstifle:
This would be so dope! What are your thoughts on the hover/press/drag states?

Just simple like figma, or do you want to design something more like vscode?

driskull:
It would probably be good to add a `resizable` prop to where we need it. Maybe just the `calcite-shell-panel` for now? We would also use a tailwind property for setting css resize.

bpatterson88:
> What are your thoughts on the hover/press/drag states? Just simple like figma, or do you want to design something more like vscode?

- Simple like Figma, using either the `ew-resize` or `col-resize` cursors.
- Min width would match `width-scale=s`, max-width would match `width-scale=l`
- Closing and reopening panel would retain custom width if set.

bstifle:
cool. what about behaviors like double clicking to expand to longest item?

bpatterson88:
Double click to expand would be pretty cool to have, but probably not a requirement.

Another wrinkle is handling items with right aligned actions. I suppose they would hug the right edge of the panel until they collide with the left edge of the item at which point they would overflow the panel. (The reality is probably more complex than that but I think that'd be the target behavior.)

TheBlueDog:
[...]

PR #3535, opened 2021-11-18
feat(shell-panel): Add resizable property to allow resizing of content. #2770

driskull:
**Related Issue:** #2770

## Summary

feat(shell-panel): Add resizable property to allow resizing of content. 

- Adds `resizable` property to turn this feature on.
- Adds `intlResize` property for resize aria label.
- Sets up pointer events for handling resize
- Uses aria role of "separator"
  -  https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/roles/separator_role
- max/min width from CSS still restricts bounds of resizing. Uses JS to get min/max values from the CSS variable.
- Keyboard accessible as well

## Todo

Styling


driskull:
@asangma can you review the DOM structure and advise on styling?

@jcfranco can you review zeee codez?

macandcheese:
This is awesome - really like the attention paid to home / end / shift + nudges. Want me to take a crack at styling?

driskull:
> This is awesome - really like the attention paid to home / end / shift + nudges. Want me to take a crack at styling?

Yes plz!

macandcheese:
Do you think adding an event that emits the px width at interaction end would be useful? If an app wants to save a user's last width, for example.

src/components/calcite-shell-panel/calcite-shell-panel.tsx
```diff
@@ -50,13 +63,56 @@ export class CalciteShellPanel {
    */
   @Prop({ reflect: true }) position: Position;
 
+  /** Accessible label for resize separator.
```
jcfranco, line 66: Nitpick: Can you start the doc text in the next line?

driskull:
Yeah, i was considering an event. I think we could add it if we think its beneficial. However, nowadays users can just use a resize observer too.

src/components/calcite-shell-panel/calcite-shell-panel.tsx
```diff
@@ -86,25 +142,222 @@ export class CalciteShellPanel {
...
+    }
+
+    forceUpdate(this);
+  };
+
+  storeContentEl = (contentEl: HTMLDivElement): void => {
+    this.contentEl = contentEl;
+  };
+
+  getKeyvalue = (event: KeyboardEvent): number => {
```
jcfranco, line 246: Nitpick: `getKeyValue`

src/components/calcite-shell-panel/calcite-shell-panel.tsx
```diff
@@ -86,25 +142,222 @@ export class CalciteShellPanel {
...
+      "ArrowRight",
+      "Home",
+      "End",
+      "PageUp",
+      "PageDown"
+    ];
+
+    if (MOVEMENT_KEYS.indexOf(key) > -1) {
+      event.preventDefault();
+      event.stopPropagation();
```
jcfranco, line 271: Do you need to stop propagation of the event? It's not internal and it would break end-user expectations.

src/components/calcite-shell-panel/calcite-shell-panel.tsx
```diff
@@ -86,25 +142,222 @@ export class CalciteShellPanel {
...
+
+    const rounded = Math.round(position === "end" ? width : width);
+
+    this.contentWidth =
+      typeof contentWidthMax === "number" && typeof contentWidthMin === "number"
+        ? clamp(rounded, contentWidthMin, contentWidthMax)
+        : rounded;
+  }
+
+  updateAriaValues = (): void => {
```
jcfranco, line 215: Let's avoid assigning bound functions if not needed.

src/components/calcite-shell-panel/calcite-shell-panel.tsx
jcfranco: Nitpick: The event is only passed for the key, so maybe it could be passed directly?

src/components/calcite-shell-panel/calcite-shell-panel.tsx
```diff
@@ -86,25 +142,222 @@ export class CalciteShellPanel {
...
+
+    forceUpdate(this);
+  };
+
+  storeContentEl = (contentEl: HTMLDivElement): void => {
+    this.contentEl = contentEl;
+  };
+
+  getKeyvalue = (event: KeyboardEvent): number => {
+    const key = getKey(event.key);
```
jcfranco, line 247: `getKey` should no longer be needed for our current browser support.

src/components/calcite-shell-panel/calcite-shell-panel.tsx
```diff
@@ -86,25 +142,222 @@ export class CalciteShellPanel {
...
+
+    if (key === "PageDown") {
+      return initialContentWidth - multipliedStep;
+    }
+
+    if (key === "PageUp") {
+      return initialContentWidth + multipliedStep;
+    }
+
+    return null;
```
jcfranco, line 308: Can you update the return type to include `null`?  Or maybe this should return `contentWidth` as a fallback.

src/components/calcite-shell-panel/calcite-shell-panel.tsx
```diff
@@ -86,25 +142,222 @@ export class CalciteShellPanel {
...
+  };
+
+  separatorKeyDown = (event: KeyboardEvent): void => {
+    this.setInitialContentWidth();
+    const width = this.getKeyvalue(event);
```
jcfranco, line 313: Nitpick: Can you rename this to better express its intent? Something like `getKeyAdjustedWidth` or along those lines? 

On my first scan I thought this method returned a value based on the key.

src/components/calcite-shell-panel/calcite-shell-panel.tsx
```diff
@@ -86,25 +142,222 @@ export class CalciteShellPanel {
...
+  // --------------------------------------------------------------------------
+  //
+  //  private Methods
+  //
+  // --------------------------------------------------------------------------
+
+  setContentWidth(width: number): void {
+    const { contentWidthMax, contentWidthMin, position } = this;
+
+    const rounded = Math.round(position === "end" ? width : width);
```
jcfranco, line 207: This is rounding the same value (`width`). Is there another value that should be used for rounding in the ternary?

src/components/calcite-shell-panel/calcite-shell-panel.e2e.ts
```diff
@@ -224,4 +233,139 @@ describe("calcite-shell-panel", () => {
...
+        </calcite-shell>
+      </div>
+    `);
+
+    await page.waitForChanges();
+
+    const separator: E2EElement = await page.find(`calcite-shell-panel >>> .${CSS.separator}`);
+
+    expect(separator).toBeDefined();
+    expect(separator.getAttribute("aria-valuenow")).toBe("320");
```
jcfranco, line 360: Can these tests also assert on the actual width?

jcfranco (approved):
Awesome! Just had some suggestions and nitpicks. Looks like a great enhancement! I'll defer to @Esri/calcite-design-representatives for the styling-side of things.

src/components/calcite-shell-panel/calcite-shell-panel.tsx
driskull: I think this is just a 🧠 💨 

:)

src/components/calcite-shell-panel/calcite-shell-panel.tsx
driskull: Its passed to preventDefault() as well.

driskull:
@jcfranco checked in fixes.

src/components/calcite-shell-panel/calcite-shell-panel.tsx
jcfranco: Ah, never mind then. 👍🏼

jcfranco (approved):
🎉 🌮 🦖

macandcheese:
@asangma did you have any thoughts on design? I think it's tough to use any kind of handle affordance because of potential overlap with content. 

Below shows the separator while hovering anywhere in the panel, to increase visibility of feature a bit. I looked at many examples of this functionality and most do not provide anything except the hover state while over the 1px separator. Below is 2px, feel free to provide feedback here!

https://jmp.sh/6gcuI7X

driskull:
> I think it's tough to use any kind of handle affordance because of potential overlap with content.

Maybe we should just have an invisible div that is positioned above the content on the side?

macandcheese:
> Maybe we should just have an invisible div that is positioned above the content on the side?

We could do that but I kind of like just using the `.separator` div. I've struggled to find an example of a draggable panel that does have some "extra handle" or anything like that, even in native OS it is usually a cursor change and border color. Open to using different color values for that cc @asangma

asangma:
Hmmm...my initial thought is to have the "grab-able" area fatten up on focus/hover. But I'll pull and build and have a look.

asangma:
As far as the `.separator` node, it has some issues when `content-behind="false"`.

driskull:
@asangma `content-behind="false"` shouldn't be a thing. Do you mean `content-behind`?

macandcheese:
Let me know if I should create a branch with the above video styles from this one. I think the 2px / hover and focus affordance while over shell panel is a good middle ground, I worry about jumpiness if that pixel width increases further on hover.

driskull:
@macandcheese just check them in here. I don't think we need a separate branch.

macandcheese:
Added styles if ya want to take another look. The `content-behind` issue @asangma mentioned is when a panel is detached - the separator is still full-height currently. It's kind of an edgy case - and is true to the actual space occupied by the shell panel container - maybe not the worst thing but open to thoughts there:
Screen Shot 2021-11-29 at 10 42 43 AM

driskull:
Do we just want to disable resizing when a panel is detached?

asangma:
> Do we just want to disable resizing when a panel is detached?

yeah...I think so.

asangma:
> content-behind="false" shouldn't be a thing. Do you mean content-behind?

I meant when `content-behind` is not set.

asangma (approved):
image
YODA MAN.

Issue #3120, opened 2021-09-27
Bug: calcite-radio-group emits multiple calciteRadioGroupChange events


jmanke:
### Actual Behavior
The calcite-radio-group emits multiple calciteRadioGroupChange events, with the event's detail alternating between the previous and the selected value. Note that this bug was originally found in EmberJs.

### Expected Behavior
Only one calciteRadioGroupChange should be fired when making a selection to be consistent with how this event is usually emitted.

### Reproduction Steps and Sample

_Setup:_
1. Create a scenario where a calcite-radio-group is conditionally rendered.
2. Add two calcite-radio-group-items.

_Repro steps:_
1. Select the second calcite-radio-group-item. For example, if your options are [1, 2] then select 2.
2. Toggle the calcite-radio-group off then on again by conditionally rendering it.
3. Select the first calcite-radio-group-item. For example, if your options are [1, 2] then select 1.
4. Observe the calciteRadioGroupChange event fires multiple times.

The [codepen](https://codepen.io/jmanke/pen/zYzyxXK) contains a more detailed set of repro steps specific to that sample. 

_Sample_:

1.  EmberJs - [codepen](https://codepen.io/jmanke/pen/zYzyxXK)

### Relevant Info

_Browser_: Chrome, Firefox, Safari
_Version_: @esri/calcite-components@<1.0.0-beta.65>
_Framework_: EmberJs v3.28.0
- [x] CDN
- [x] NPM package

cc @vijayendranj @julio8a

jgrayson-apl:
+1

anveshmekala:
> I was able to repro in vanilla JS, triaging to next sprint https://codepen.io/benesri/pen/eYENJBX?editors=0010
> 
> cc @jcfranco

Couldn't reproduce this in local on master thought its repro in codepen & jsfiddle

benelan:
I can repro locally
```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8" />
    <meta
      name="viewport"
      content="initial-scale=1, maximum-scale=1, user-scalable=no"
    />
    <script
      type="module"
      src="https://unpkg.com/@esri/calcite-components@1.0.0-beta.69/dist/calcite/calcite.esm.js"
    ></script>
    <link
      rel="stylesheet"
      type="text/css"
      href="https://unpkg.com/@esri/calcite-components@1.0.0-beta.69/dist/calcite/calcite.css"
    />
  </head>
  <style>
    calcite-button {
[...]

PR #3565, opened 2021-11-24
fix(radio-group): no longer emit calciteRadioGroupChangemultiple times

anveshmekala:
**Related Issue:** #3120 

## Summary

This fix will avoid emitting `calciteRadioGroupChange` event multiple when the user switch the `calciteRadioGroupItem`


anveshmekala:
The issue is resolved in #3632

PR #3568, opened 2021-11-24
fix(input): fix inconsistent value stepping

jcfranco:
**Related Issue:** N/A

## Summary

This PR fixes a regression introduced by https://github.com/Esri/calcite-components/issues/3443 ([repro case](https://codepen.io/jcfranco/pen/MWvMQWM)). 

cc @benelan

### DIFF
--- a/src/components/calcite-input/calcite-input.tsx
+++ b/src/components/calcite-input/calcite-input.tsx
@@ -383,22 +383,26 @@ export class CalciteInput implements LabelableComponent, FormComponent {
 
   incrementOrDecrementNumberValue(
     direction: NumberNudgeDirection,
-    inputMax: number,
-    inputMin: number,
+    inputMax: number | null,
+    inputMin: number | null,
     nativeEvent: KeyboardEvent | MouseEvent
   ): void {
-    const value = this.value;
+    const { value } = this;
     const inputStep = this.step === "any" ? 1 : Math.abs(this.step || 1);
     const inputVal = value && value !== "" ? parseFloat(value) : 0;
-    let newValue = value;
 
-    if (direction === "up" && ((!inputMax && inputMax !== 0) || inputVal < inputMax)) {
-      newValue = (inputVal + inputStep).toFixed(decimalPlaces(inputStep)).toString();
-    }
-    if (direction === "down" && ((!inputMin && inputMin !== 0) || inputVal > inputMin)) {
-      newValue = (inputVal - inputStep).toFixed(decimalPlaces(inputStep)).toString();
-    }
-    this.setValue(newValue, nativeEvent, true);
+    const adjustment = direction === "up" ? 1 : -1;
+    const nudgedValue = inputVal + inputStep * adjustment;
+    const finalValue =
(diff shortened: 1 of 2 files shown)


benelan:
woops, nice catch

benelan (approved):
nice catch

src/components/calcite-input/calcite-input.tsx
```diff
@@ -383,22 +383,26 @@ export class CalciteInput implements LabelableComponent, FormComponent {
...
-    if (direction === "down" && ((!inputMin && inputMin !== 0) || inputVal > inputMin)) {
-      newValue = (inputVal - inputStep).toFixed(decimalPlaces(inputStep)).toString();
-    }
-    this.setValue(newValue, nativeEvent, true);
+    const adjustment = direction === "up" ? 1 : -1;
+    const nudgedValue = inputVal + inputStep * adjustment;
+    const finalValue =
+      (typeof inputMin === "number" && !isNaN(inputMin) && nudgedValue < inputMin) ||
+      (typeof inputMax === "number" && !isNaN(inputMax) && nudgedValue > inputMax)
+        ? inputVal
```
benelan, line 399: Question: in terms of functionality, should it be bumped to the min/max value instead of staying at the the input value in this case? Since the user sets min/max I think it should cap out at those numbers nudging rather than a step away

eriklharper (approved):
Looks awesome

src/components/calcite-input/calcite-input.tsx
eriklharper: This might not matter too much considering we plan on allowing more flexibility with what is allowed to be typed, leaning more on externally-provided validation, but this does bring up an interesting question.  I'm wondering what the native input does in this case.

benelan: Can you think of a use case where someone might want a step that would not align with their min/max values? Like a step of 5 and a max of 44. The user could still type 44 but using the + button they would only get to 40. I guess we could just keep it the way it is so that step always does what people expect and then its up to the user to chose min/max values that align with their step value.

jcfranco: Great question. I tested out `<input type="number" value="1.008" max=7 step=5>` on our supported browsers they all behave as this PR except for Safari. 

Looking at [the spec](https://html.spec.whatwg.org/multipage/input.html#:~:text=If%20the%20element%20has%20a%20minimum%2C%20and,is%20less%20than%20or%20equal%20to%20maximum.), I _think_ this is the correct behavior:

> If the element has a minimum, and value is less than that minimum, then set value to the smallest value that, when subtracted from the step base, is an integral multiple of the allowed value step, and that is more than or equal to minimum.
> 
> If the element has a maximum, and value is greater than that maximum, then set value to the largest value that, when subtracted from the step base, is an integral multiple of the allowed value step, and that is less than or equal to maximum.

jcfranco: > Can you think of a use case where someone might want a step that would not align with their min/max values?

The only case I can think of is being able to undo a series of value updates with the same amount of steps and without having to type in a value.

Issue #3373, opened 2021-10-28
Bug: calcite-value-list-item's label takes up space even if there is no content specified


ptmdmusique:
### Actual Behavior

calcite-value-list-item's label (which is inside calcite-pick-list-item) takes up space even if no label is passed as prop

### Expected Behavior

the label should not take up space and we can stretch the `actions-end` or `actions-start`

### Reproduction Steps and Sample

https://codepen.io/ptmdmusique/pen/YzxZaZK

cc @paulcpederson @abi11500

### Relevant Info

---

driskull:
We might just need to improve documentation and make `label` required.  Label is like a title and description is like a subtitle in this case. You can't have just a subtitle without a title.

@benelan @jcfranco Any suggestions on how we can catch these things before assignment? I don't want to waste dev time. Or we can communicate that issues require further investigation before proceeding with development.

jcfranco:
`calcite-value-list-item#label` [is marked as required](https://developers.arcgis.com/calcite-design-system/components/value-list/#component-api-properties-label). Sidebar, `calcite-pick-list-item#label` [is not](https://developers.arcgis.com/calcite-design-system/components/value-list/#component-api-properties-label). cc @benelan 

@driskull Let me assure you that nobody wants to waste dev time, this just slipped through the cracks during triaging.

benelan:
Woops the mention must have gotten buried in my emails, thanks for taking care of this Matt. [Verified on master](https://github.com/Esri/calcite-components/blob/master/src/components/pick-list-item/readme.md)

PR #3579, opened 2021-11-29
feat(value-list-item): add new slot labelled as content-center to replace label/description (#3373)

y0n4:
**Related Issue:** #3373

## Summary

In the case of when the `label` and `description` prop is not utilized, there is a custom `content` slot that can replace the area of the label/description. This way there won't be a big gap in the middle of the component and no need to squish more content to the other slots like `actions-start` or `actions-end`. Label/description or content slot must be provided while using the `calcite-pick-list-item` component

Content slot will take precedence over label/description if provided.
When the list item is interactive and has content slot, the item can be selected if user clicks outside of the content slot.

Before: you can see that there is an empty gap where the label/description is supposed to be and the user wants to include more content
Screen Shot 2021-12-13 at 11 09 49 AM 2
After: now label/description can be replaced with custom content and there is no longer that empty gap
Screen Shot 2021-12-03 at 1 52 06 PM 2


driskull:
@y0n4 can you add a test for this? Maybe a screener update to show an item without a title?

jcfranco:
@asangma Can you confirm if having no content still fits with the standardized item layout pattern

`[action-start][content-start][content][content-end][action-end])`

?

asangma:
If I understand correctly, you want to know if no `label` or  `description` is added, the `actions-start` and `actions-end` slots should fill the available space in, yes?

As far as that layout pattern, that' pretty much it, though the full pattern wraps the content slots and node in a button.
`[actions-start] [ button [content-start][content][content-end] ] [actions-end]`

y0n4:
@asangma yes that is correct. Should the component require a label or can it be empty and the slots will stretch the rest of the space? There is that whole empty gap around the label area because it is expecting a label. The e2e test is also expecting a label as well so if it's not required, the test will fail

It is a bit confusing since label prop is required in `calcite-value-list-item` (which is utilized in this issue use case) but not required in `calcite-pick-list-item` which is basically the entire structure of `calcite-value-list-item` component. The consistency is a bit off

asangma:
Seems like we should just require it in PickListItem.

asangma:
Because the request is focused on allowing custom content in the main content area, I suggest that we 

- do *not* require `label`
- we render the default slot in the content area

This way a user can do like

```html
<calcite-pick-list-item>
  <calcite-action slot="actions-end"></calcite-action>
  <calcite-label>
    <calcite-input />
  </calcite-label>
  <!-- or whatever -->
</calcite-pick-list-item>
```

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -190,6 +193,7 @@ export class CalcitePickListItem {
   @Method()
   async setFocus(): Promise<void> {
     this.focusEl?.focus();
+    this.focusDiv?.focus();
```
driskull, line 196: I'm not sure it will ever focus the `focusEl` in this case. It will always focus the div. Is that correct? I'm not sure we can focus one and then the other.

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -217,6 +221,10 @@ export class CalcitePickListItem {
     }
   };
 
+  contentSlotClickHandler = (event: MouseEvent): void => {
+    event.stopPropagation();
```
driskull, line 225: What is the reasoning for stopping propagation here?

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
driskull: maybe it should be written so that it's clear only one of these is going to get focused?

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -270,6 +278,43 @@ export class CalcitePickListItem {
...
+      <label
+        aria-label={label}
+        class={CSS.label}
+        onClick={this.pickListClickHandler}
+        onKeyDown={this.pickListKeyDownHandler}
+        ref={(focusEl): HTMLLabelElement => (this.focusEl = focusEl)}
+        tabIndex={0}
+      >
+        <div
+          aria-checked={this.selected.toString()}
```
driskull, line 307: Is this `role` and `aria-checked` no longer needed when the content slot is used? Should they be moved onto the div surrounding the content slot?

driskull (commented):
Just some comments/questions.

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
y0n4: When label/description text content is clicked, the item container gets focused. In this case when the slot has input and when input is clicked, the item container gets focused instead of the input. So this was to prevent the slot content focus from bubbling up to the outer div. I tried to word it in the pr summary for that

> When the list item is interactive and has content slot, the item can be selected if user clicks outside of the content slot.

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
```diff
@@ -91,9 +91,46 @@
 }
 
 .actions {
-  @apply items-stretch flex justify-end m-0 flex-initial;
+  @apply items-stretch flex m-0 flex-initial;
 }
 
 .actions--start ~ .label {
+  @apply pl-1;
```
driskull, line 98: Can you use `padding-inline-start` instead of left? Then we can remove the padding-right and RTL classes.

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
```diff
@@ -91,9 +91,46 @@
...
+  @apply cursor-default;
+  &:focus {
+    @apply focus-base;
+  }
+  &:hover {
+    @apply bg-transparent;
+  }
+}
+
+.calcite--rtl .actions--start ~ .label {
```
driskull, line 132: remove this block of code.

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -129,6 +130,8 @@ export class CalcitePickListItem {
 
   private focusEl: HTMLLabelElement;
```
driskull, line 131: maybe this could be `private focusEl: HTMLLabelElement | HTMLDivElement;`?;

driskull: And then the same private property can be shared?

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
driskull: hmmm I'm not sure we want to do this. Maybe that should be up to the author to prevent bubbling on the input? This doesn't seem right to stop propagation here.

@jcfranco what do you think?

driskull: What's the use case for the input? This is already essentially a role of checkbox so having an input inside of it might be problematic.

driskull (commented):
Code looks good, I just have concerns/questions about stopping the propagation on the content container as well as what we are expecting in that slot.

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
y0n4: Ah, yea the input was an example that the creator for this issue was presenting.

@asangma might be able to speak more about the case in which content slot is used and what happens if content slot is clicked

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
y0n4: This styling was a change that was included when I merged master into this branch. I didn't want to alter it and accepted this current change when resolving merge conflicts

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
y0n4: I ended up not using `stopPropagation` and used a styling where container rests on top of the item list so the focus won't bubble up to the current item

src/components/calcite-value-list/calcite-value-list.stories.ts
```diff
@@ -105,6 +105,23 @@ export const basic = (): string =>
...
+  create(
+    "calcite-value-list",
+    createAttributes(),
+    html`
+      <calcite-value-list-item label="Default label and description" description="used as content" value="dogs">
+        ${action}
+      </calcite-value-list-item>
+      <calcite-value-list-item value="cats">
+        <div slot="content">
+          <calcite-input placeholder="slot content"></calcite-input>
```
driskull, line 118: I think this will be problematic, the role of the list-item is a checkbox. The checkbox can't contain an input inside of it I don't think. The DOM structure and roles would need to be reconfigured.

Can we make sure that this HTML passes aXe?

driskull (changes requested):
The code looks good @y0n4 but I think we need to check accessibility and make sure that its passing aXe. Also, we should document if the content of the slot can be form elements. If so, we should restructure the code so that there are not nested form element roles.

cc @asangma

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
```diff
@@ -91,9 +92,51 @@
...
+  items-center;
+}
+
+.content--select {
+  @apply bg-transparent
+    cursor-pointer
+    focus-base
+    w-full
+    h-full
+    absolute;
```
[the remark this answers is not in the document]
y0n4: I did absolute position for the selected portion of the item in relative to the current content. That way users can have the choice to focus on the item itself or the content within. This is to escape the events from bubbling up and an alternative to the stopPropagation method.

asangma:
@driskull I'm thinking that if the `content` slot is used, we just render whatever they put in there and don't do anything with label props or click handlers. Basically, if you wanna take control of that, you have control over behavior.

driskull:
@asangma I think some more research needs to happen here. The role of a pick-list-item is a checkbox.

If a user slots an input inside of a pick-list-item, then its basically like putting a input inside a checkbox.

If we remove the checkbox role, then it's no longer able to "pick".

The purpose of this slot needs to be clarified and limited otherwise we are setting users up for a11y failures and breaking changes down the road for us.

driskull:
If a input needs to be slotted, it needs to go next to the checkbox role, not inside it.

driskull:
@benelan @jcfranco we might need to require more usage of the `accessible` test helper. So that when a new slot is added its a11y is tested. Probably a convention update here.

jcfranco:
Apologies for not chiming in sooner. I'll review first thing tomorrow.

asangma:
@jcfranco @driskull I think we need to revisit the design and implementation of this. Chatted with @y0n4.

asangma:
Here is the basic concept. I'd like to run the UI (see below) through a design review.
image

I have some initial UI stuff [here](https://www.figma.com/file/WAAX5rFvNQu3PDlvajMd93/List-with-content-slot?node-id=9%3A1342).

driskull:
Maybe two separate item types since the UI will be different depending on a slot being used? I'm not sure where we draw the line and say its a separate item component rather than it renders differently when things are slotted.

asangma:
I considered that. However, one of the workflows/UX in question is being able to do some kind of inline editing, e.g. temporarily replace the standard content with an input.

asangma:
I'd be willing to remove the select/deselect button when "content" slot is used.
That'd be the most straightforward approach I think.

macandcheese:
At what point is a regular list item more apt?

asangma:
@macandcheese A dev could do that, but ListItem doesn't come with all the single-select, multi-select and prescribed icons (check vs radio).

asangma:
Really, this is a pattern we use in a couple other places, i.e. having a slot that overrides a main content area.

That being said, I'm open to other solutions for the initial request to have inline editable text.

jcfranco:
Blocking and freezing this one until we flesh out the details for the action/content slot pattern epic. @y0n4 @asangma @driskull @macandcheese Thanks for the time and energy on this, but the implementation of this brought up some items that needed to be well-defined first.

Issue #1813, opened 2021-03-19
Enhancement: Disable spinner and horizontal spinners for Read Only Inputs


rmstinson:
### Description
Current spinner inputs keep their spinner enabled when input is read-only. Proposing that we disable the spinner buttons along with the input field.
### Acceptance Criteria 
If those spinners are disabled, we got a stew goin'.
### Relevant Info 
NA
#### Which Component
Input
#### Example Use Case
![Read Only Disabled Spinners](https://user-images.githubusercontent.com/12650705/111851891-38a85e00-88d2-11eb-9238-83a71119c21f.png)

macandcheese:
For input of type number, should we hide the number increment / decrement buttons (either in horizontal or vertical) altogether when in readonly?

bstifle:
yeah that makes sense to just hide them if we can!!

asangma:
@Elijbet Hopefully [this codepen](https://codepen.io/mtb-alan/pen/QWMYxMp) will help.

When the input is set to `read-only`, the increment buttons and input remain interact-able.

I believe @rmstinson is suggesting that `read-only` should respond to mouse and keyboard the same way as `disabled`.

@rmstinson I think we should also update the `disabled` appearance no?

Elijbet:
@rmstinson If you could also provide styles suggestions for disabled vs read-only, that'd be awesome. Thanks!

rmstinson:
So as far as the visual design goes, I'm proposing we follow this design, with the disabled inputs set to .5 opacity, following the disabled state of the calcite button's example.

In addition, we need to add the appropriate aria tag so that even is a component is disabled, screen readers & keyboard navigators will still be able to focus on the component and communicate the info and status of the component:

![Input_states](https://user-images.githubusercontent.com/12650705/142936782-48a54f1f-fd76-4af9-9ea2-19d93df259b8.png)

Link to the design doc: https://www.figma.com/file/eN2aaw0QPZCfLHA3lErhae/input_states?node-id=0%3A1

The application of the aria tag is pending further review, but the visual design should be pretty close to set if the team agrees.

asangma:
@rmstinson @Elijbet If they're not already there, we may need to add an `intlDisabled` property to allow users to localize that string for that aria tag.

Mike, I'm assuming that that aria tag would default to "disabled"?

rmstinson:
[...]

PR #3582, opened 2021-11-30
fix(input): style read-only and disabled inputs per new design reqs and hide spinners for read-only inputs

Elijbet:
**Related Issue:** #1813

## Summary

- for input of type number, hide the number increment / decrement buttons when in readonly
- incorporated styles suggestions for disabled vs read-only with lower opacity for disabled inputs and bolder font for read-only
- set up demos for comparison of filled, disabled, read-only


driskull:
> use aria-disabled attribute instead of disabled, so that element can be assigned focus

Should it be focusable if its disabled? Seems like native doesn't behave like that.

Elijbet:
> > use aria-disabled attribute instead of disabled, so that element can be assigned focus
> 
> Should it be focusable if its disabled? Seems like native doesn't behave like that.

I followed the direction in the issue: 
"In addition, we need to add the appropriate aria tag so that even is a component is disabled, screen readers & keyboard navigators will still be able to focus on the component and communicate the info and status of the component"

I might have misunderstood what it's asking for?

driskull:
I think a disabled component shouldn't receive focus.

src/components/calcite-input/calcite-input.tsx
```diff
@@ -718,6 +718,7 @@ export class CalciteInput implements LabelableComponent, FormComponent {
     const localeNumberInput =
       this.type === "number" ? (
         <input
+          aria-disabled={this.disabled ? true : null}
```
driskull, line 721: I think we can remove this since it already has disabled attribute.

src/components/calcite-input/calcite-input.tsx
```diff
@@ -788,7 +788,7 @@ export class CalciteInput implements LabelableComponent, FormComponent {
     return (
       <Host onClick={this.inputFocusHandler} onKeyDown={this.keyDownHandler}>
         <div class={{ [CSS.inputWrapper]: true, [CSS_UTILITY.rtl]: dir === "rtl" }} dir={dir}>
-          {this.type === "number" && this.numberButtonType === "horizontal"
+          {this.type === "number" && this.numberButtonType === "horizontal" && !this.readOnly
```
driskull, line 791: Can we add an e2e test for this readOnly usage?

src/components/calcite-input/calcite-input.tsx
```diff
@@ -802,11 +802,11 @@ export class CalciteInput implements LabelableComponent, FormComponent {
           <div class={CSS.actionWrapper}>
             <slot name={SLOTS.action} />
           </div>
-          {this.type === "number" && this.numberButtonType === "vertical"
+          {this.type === "number" && this.numberButtonType === "vertical" && !this.readOnly
```
driskull, line 805: Can we add an e2e test for this readOnly usage?

src/components/calcite-input/calcite-input.tsx
```diff
@@ -802,11 +802,11 @@ export class CalciteInput implements LabelableComponent, FormComponent {
...
             ? numberButtonsVertical
             : null}
           {this.suffixText ? suffixText : null}
-          {this.type === "number" && this.numberButtonType === "horizontal"
+          {this.type === "number" && this.numberButtonType === "horizontal" && !this.readOnly
```
driskull, line 809: Can we add an e2e test for this readOnly usage?

driskull (changes requested):
Looks good @Elijbet, can we just add some e2e tests for the `readOnly` changes.

driskull (approved):
Looks good! 🎉 🌮

PR #3596, opened 2021-11-30
feat: Add utility to watch global attributes of a component.

driskull:
**Related Issue:** N/A

## Summary

feat: Add utility to watch global attributes of a component.

- Adds a `attributes` utility file. This will allow us to listen for changes to global attributes on a component like "lang" without having to create a property on the component.
- File contains two methods...
  -  watchAttributes: watches global attributes for changes and re-renders with them saved to an object for use within render methods.
  - unwatchAttributes: removes watcher.

@jcfranco what do you think?


src/utils/attributes.ts
```diff
@@ -0,0 +1,68 @@
...
+export function watchAttributes(component: WatchedComponent, attributes: string[]): void {
+  const attributeObject: AttributeObject = {};
+  const { el } = component;
+
+  const updateAttributesObject = () => {
+    attributes.forEach((attr) => {
+      if (el.hasAttribute(attr)) {
+        const value = el.getAttribute(attr);
+        if (value !== null) {
+          attributeObject[attr] = el.getAttribute(attr);
```
jcfranco, line 46: `attributeObject[attr] = value;`?

src/utils/attributes.ts
```diff
@@ -0,0 +1,68 @@
...
+        }
+      }
+    });
+
+    component.inheritedAttributes = { ...attributeObject };
+  };
+
+  updateAttributesObject();
+
+  component.inheritedAttributesObserver = createObserver("mutation", () => updateAttributesObject());
```
jcfranco, line 56: Can this utility use a shared observer? It would make it simpler on implementors since they wouldn't need to store anything.

src/utils/attributes.ts
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Helper to set up listening for changes to global attributes.
+ *
+ * render() {
+ *   const lang = this.inheritedAttributes['lang'] ?? 'en';
+ *   return <div>My lang is {lang}</div>;
+ * }
+ */
+export function watchAttributes(component: WatchedComponent, attributes: string[]): void {
```
jcfranco, line 37: To help keep this strictly for inherited attribute usage, should we add an attribute allowlist? As an extra step, we can also use some TS magic to avoid watching anything that matches a prop from a component.

src/utils/attributes.ts
```diff
@@ -0,0 +1,68 @@
...
+  const attributeObject: AttributeObject = {};
+  const { el } = component;
+
+  const updateAttributesObject = () => {
+    attributes.forEach((attr) => {
+      if (el.hasAttribute(attr)) {
+        const value = el.getAttribute(attr);
+        if (value !== null) {
```
jcfranco, line 45: Can you confirm that boolean attributes get toggled appropriately?

jcfranco (commented):
This is looking great! 

Unless this PR will update components to start inheriting attributes, this should probably use the `refactor` commit type.

src/utils/attributes.ts
driskull: Yeah, that sounds good. Should we just start with `["lang"]`?

src/utils/attributes.ts
driskull: Will do once testing proceeds.

src/utils/attributes.ts
driskull: Can you explain this one? I'm not seeing it.

eriklharper: Shared meaning you pass in your own observer to the `watchGlobalAttributes` function?

eriklharper:
This is a great start!  I'm wondering if we should expand this to also observe global property changes with something like [Proxies](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy)?  This would be very useful as well since developers will probably expect to be able to mutate the property and not just the attribute.

This also feels like this kind of ability should be baked into Stencil Core itself.  Have you filed an issue or asked a question about this with the Stencil Core team?  Might be worth it to at least start a conversation to see if they're thinking about this problem based on user feedback.

src/utils/attributes.ts
jcfranco: I meant having the utility use and manage the observer itself. This would eliminate implementors having to add a `globalAttributesObserver` property on a component.

src/utils/attributes.ts
jcfranco: Sounds like a good start.

src/utils/globalAttributes.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  updateAttributesObject();
+
+  const mutationObserver = createObserver("mutation", () => updateAttributesObject());
+
+  mutationObserver.observe(el, {
+    attributeFilter: attributes
+  });
+
+  mutationObserverMap.set(component, mutationObserver);
```
driskull, line 62: @jcfranco is this what you had in mind?

src/utils/attributes.ts
driskull: The only bummer about that is if the component forgets to disconnect the observer it stays in the utility class.

src/utils/globalAttributes.ts
jcfranco: Very close. Instead of a `component-mutation` observer map, I was thinking an `element-[component, attributes]` map (or 2 separate maps to avoid arrays) that a single observer and update function could use.

src/utils/attributes.ts
jcfranco: True, but this would be a problem for any utils/patterns we have.

src/utils/attributes.ts
jcfranco: Just to clarify, will this PR cover that testing?

src/utils/globalAttributes.ts
driskull: @jcfranco is that because you think the `attributes` may change after initialization? I don't think that's a very likely case. Maybe we could defer doing that?

jcfranco: > is that because you think the attributes may change after initialization?

No, it's to be able to use a single observer and associate each element to their watched attributes.

driskull:
@jcfranco @eriklharper this util may not be necessary if we are going to use browser intl, is that correct?

src/utils/globalAttributes.ts
driskull: @jcfranco I'm not following how you can use a single observer. Can you take that part? I'm not seeing how it can setup a single observer to call a generic method like that.

jcfranco: Apologies for the delay on this. I've updated the PR with what I was envisioning and also added a spec test for the util. Can you take a look? 👀

eriklharper (approved):
Thanks for adding the spec test!  I suppose a follow-up issue would be appropriate to refactor time picker to use this new utility, replacing the Prop decorator.

jcfranco:
> I suppose a follow-up issue would be appropriate to refactor time picker to use this new utility, replacing the Prop decorator.

Yeah, that'd be a great. Can you create an issue for this?

eriklharper:
Done! https://github.com/Esri/calcite-components/issues/3939

src/utils/attributes.ts
jcfranco: I'll create a separate issue for this.

PR #3611, opened 2021-12-01
feat(tooltip): simplify tooltip usage by no longer requiring tooltip manager

jcfranco:
**Related Issue:** N/A

## Summary

Simplifies tooltip usage by no longer requiring the manager.

### Pending

- Replacement for `selector` – this could be replaced by adding a tooltip `group` property or similar
- Updating tests


src/components/calcite-tooltip/calcite-tooltip.tsx
```diff
@@ -11,6 +11,207 @@ import {
...
+
+  registerElement(element: HTMLElement, tooltip: HTMLCalciteTooltipElement): void {
+    this.registeredElements.set(element, tooltip);
+
+    element.addEventListener("keyup", this.keyUpHandler);
+    element.addEventListener("mouseover", this.mouseEnterShow);
+    element.addEventListener("mouseout", this.mouseLeaveHide);
+    element.addEventListener("click", this.clickHandler);
+    element.addEventListener("focus", this.focusShow);
+    element.addEventListener("blur", this.blurHide);
```
driskull, line 82: The only downside i see is that every tooltip needs all these events. The reason for the manager, was to have delegated events so that the manager only has these events and not every tooltip instance.

driskull:
@jcfranco whats the plan for this? Do we want to try and do this?

jcfranco:
It's on my TODO list. 😄  I need to make some tweaks and we could do some testing.

src/components/tooltip/tooltip.tsx
```diff
@@ -129,17 +362,20 @@ export class Tooltip {
   /** Updates the position of the component. */
   @Method()
   async reposition(): Promise<void> {
-    const { popper, el, placement } = this;
+    const { popper, el, open, placement } = this;
+
+    if (!open || !popper) {
```
driskull, line 367: We need to position/create popper before we open so that we can do the animation. I'll revert this part.

src/components/tooltip/tooltip.tsx
```diff
@@ -57,8 +286,12 @@ export class Tooltip {
   @Prop({ reflect: true }) open = false;
 
   @Watch("open")
-  openHandler(): void {
-    this.reposition();
+  openHandler(value: boolean): void {
+    if (value) {
```
driskull, line 290: We need to position/create popper before we open so that we can do the animation. I'll revert this part.

src/components/tooltip/TooltipManager.ts
```diff
@@ -0,0 +1,199 @@
...
+
+  private removeClosedListeners(referenceEl: HTMLElement): void {
+    referenceEl.removeEventListener("click", this.clickHandler);
+    referenceEl.removeEventListener("focus", this.focusShow);
+    referenceEl.removeEventListener("mouseover", this.mouseEnterShow);
+  }
+
+  private removeOpenListeners(referenceEl: HTMLElement): void {
+    referenceEl.removeEventListener("keydown", this.keyDownHandler);
+    referenceEl.removeEventListener("mouseout", this.mouseLeaveHide);
```
driskull, line 88: I think we need to add an event listener to the tooltip mouseout too to close it if the tooltip is hovered off.

driskull:
@jcfranco seems like the only thing not working is hovering out of a tooltip doesn't trigger the timeout to restart and close the tooltip.

driskull:
@jcfranco I see that in the PR there are events being set up on every referenceElement.

What if instead, we just setup the events on the document instead? Currently, we setup the events on the tooltip-manager element but we can probably just move these events to the document instead. What do you think?

jcfranco:
Superseded by https://github.com/Esri/calcite-components/pull/4422. 🎉

jcfranco:
Thanks, @driskull! 💥

PR #3612, opened 2021-12-01
fix(dropdown): stop calciteDropdownClose event from firing twice when closing the dropdown

Elijbet:
**Related Issue:** ##1688

## Summary

When calciteDropdownItemSelect closes the dropdown, transitionEnd fires for each property transitioned.
this.calciteDropdownClose.emit() fires on the 'opacity' property when this.active changes to false, but opacity then is transitioned twice (I think for icon and popper repositioning). Changed it to a property 'visibility' that fires only once when closing.


driskull (approved):
👍

PR #3614, opened 2021-12-01
fix(shell-panel): Correct resizing of panels in RTL direction. #2770

driskull:
**Related Issue:** #2770

## Summary

fix(shell-panel): Correct resizing of panels in RTL direction. #2770

- Fixes logic to handle resizing via mouse and keyboard when in RTL
- Deprecates `getElementDir`
  - The problem with this utility method is that it only considers the dir attribute. Someone could set CSS `direction: rtl` and this wouldn't find it.
  - The other reason for deprecating this is to make sure that components that use it evaluate if they actually need to. Many components can just use CSS logical properties instead.
- Adds `getElementStyleDir` which looks for the CSS direction value.
  - This is only necessary when needing to know the direction to do something in JS.

cc @jcfranco

### DIFF
--- a/src/utils/dom.ts
+++ b/src/utils/dom.ts
@@ -26,6 +26,13 @@ export function getThemeName(el: HTMLElement): "light" | "dark" {
   return closestElementCrossShadowBoundary(el, `.${CSS_UTILITY.darkTheme}`) ? "dark" : "light";
 }
 
+export function getElementStyleDir(el: HTMLElement): Direction {
+  return (window.getComputedStyle(el).direction as Direction) || "ltr";
+}
+
+/**
+ * @deprecated use getElementStyleDir instead.
+ */
 export function getElementDir(el: HTMLElement): Direction {
   const prop = "dir";
   const selector = `[${prop}]`;
--- a/src/components/calcite-shell-panel/calcite-shell-panel.tsx
+++ b/src/components/calcite-shell-panel/calcite-shell-panel.tsx
@@ -276,7 +276,7 @@ export class CalciteShellPanel {
     const dir = getElementStyleDir(el);
 
     const directionKeys = ["ArrowLeft", "ArrowRight"];
-    const directionFactor = dir === "rtl" ? -1 : 1;
+    const directionFactor = dir === "rtl" && directionKeys.includes(key) ? -1 : 1;
 
     const increaseKeys =
       key === "ArrowUp" ||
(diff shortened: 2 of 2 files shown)


src/utils/dom.ts
```diff
@@ -26,6 +26,13 @@ export function getThemeName(el: HTMLElement): "light" | "dark" {
   return closestElementCrossShadowBoundary(el, `.${CSS_UTILITY.darkTheme}`) ? "dark" : "light";
 }
 
+export function getElementStyleDir(el: HTMLElement): Direction {
```
driskull, line 29: I could also just replace `getElementDir` with this logic. But we should refactor all uses of getElementDir to see where we can remove its use and use CSS logical properties instead.

jcfranco (approved):
LGTM! 

The one suggestion I'll make is to refactor the code to avoid associating RTL with inverting the left and right arrow keys. Here's one idea:

```diff
diff --git c/src/components/calcite-shell-panel/calcite-shell-panel.tsx i/src/components/calcite-shell-panel/calcite-shell-panel.tsx
index 1bedc03cf..b7c618515 100755
--- c/src/components/calcite-shell-panel/calcite-shell-panel.tsx
+++ i/src/components/calcite-shell-panel/calcite-shell-panel.tsx
@@ -276,10 +276,7 @@ export class CalciteShellPanel {
     const dir = getElementStyleDir(el);

     const directionKeys = ["ArrowLeft", "ArrowRight"];
-
-    if (dir === "rtl") {
-      directionKeys.reverse();
-    }
+    const directionFactor = dir === "rtl" ? -1 : 1;

     const increaseKeys =
       key === "ArrowUp" ||
@@ -288,7 +285,7 @@ export class CalciteShellPanel {
     if (increaseKeys) {
       const stepValue = event.shiftKey ? multipliedStep : step;

-      return initialContentWidth + stepValue;
+      return initialContentWidth + directionFactor * stepValue;
     }

     const decreaseKeys =
@@ -298,7 +295,7 @@ export class CalciteShellPanel {
     if (decreaseKeys) {
       const stepValue = event.shiftKey ? multipliedStep : step;

-      return initialContentWidth - stepValue;
+      return initialContentWidth - directionFactor * stepValue;
     }

     if (typeof contentWidthMin === "number" && key === "Home") {
@@ -336,13 +333,12 @@ export class CalciteShellPanel {
     const offset = event.clientX - initialClientX;
     const dir = getElementStyleDir(el);

-    const directionValues = [initialContentWidth + offset, initialContentWidth - offset];
-
-    if (dir === "rtl") {
-      directionValues.reverse();
-    }
+    const adjustmentDirection = dir === "rtl" ? -1 : 1;
+    const adjustedOffset =
+      position === "end" ? -adjustmentDirection * offset : adjustmentDirection * offset;
+    const width = initialContentWidth + adjustedOffset;

-    this.setContentWidth(position === "end" ? directionValues[1] : directionValues[0]);
+    this.setContentWidth(width);
   };

   separatorPointerUp = (event: PointerEvent): void => {
```

Issue #3580, opened 2021-11-29
calcite-panel height issues 


AdelheidF:
When using calcite-panel the height is no longer correct.

With beta.69
![image](https://user-images.githubusercontent.com/44481292/143956435-6f28cba9-bc74-4c79-9f86-94ab0502c384.png)
![image](https://user-images.githubusercontent.com/44481292/143956582-caef9d94-b433-43f5-9d9a-4e0ab3ac75ea.png)

With beta.70
![image](https://user-images.githubusercontent.com/44481292/143956479-578441ff-2353-417c-a986-cc5fdfdb5b97.png)
![image](https://user-images.githubusercontent.com/44481292/143956539-f47a59d7-91b9-4a61-b9d4-b49cf36bcad3.png)

There is a new div element now. The div has a correct height of 100%, but the section in it has height 2px. The height of the section element when it was working was 252px (in my case).

Might have been the fix for this issue: https://github.com/Esri/calcite-components/issues/3236
https://github.com/Esri/calcite-components/commit/18f1e1d6b5d11a200ec4b7fb734ac5c7d91f1643#diff-fdbe551fa08339a23b3f85cd7119b1d75fa4cc666cd4d36369a8ce37355397edL394

### Actual Behavior

panel content shows in UI

### Expected Behavior

panel content does not show in UI

### Reproduction Steps and Sample

https://codepen.io/afreitag/pen/XWeWgOM

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-beta.70`

- [ ] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-beta.69`

jcfranco:
@y0n4 Can you take a look at this one?

jcfranco:
@AdelheidF Could you update the repro case? It looks the same using both versions (69 & 70).

AdelheidF:
@y0n4 Hi, what's the status on this one?

y0n4:
Hi @AdelheidF , I've been looking into it all day and trying to form a solution that can keep this issue and the other fixed issue on a happy medium but it's been clashing. Reverting the changes is possibly up there and I might consider to readdress the other fixed issue at a later time. I will try to wrap this up by tomorrow

AdelheidF:
Hi, any updates today?

y0n4:
@AdelheidF I opened a pr, hoping to merge today

AdelheidF:
[...]

PR #3616, opened 2021-12-02
fix(panel): add conditional rendering for slots with fab and no fab

y0n4:
**Related Issue:** #3580 

## Summary
The updated fix cannot cater equally to content with fab icon and content with unset height since both cases have contradicting structure, I have taken the previous (.69 version) code for this use case and opted for a conditional rendering to differ between content + fab and just regular content

Before
Screen Shot 2021-12-02 at 9 06 59 AM
After
Screen Shot 2021-12-02 at 9 06 38 AM


jcfranco (approved):
The code looks good to me, but this needs stories for screenshot testing.

Issue #1707, opened 2021-03-08
Bug: Toggling a modal's close button changes height of modal header


nwhittaker:
### Summary
Sometimes it is necessary to dynamically change whether or not a `<calcite-modal>` element's close button is enabled. For example, a delete confirmation initially has the close button enabled, but once the user chooses to do the delete, the modal could stick around in a _waiting_ state to ensure the delete succeeds before it dismisses. While in this state, the close button becomes disabled.

### Actual Behavior
The act of toggling the modal's `disableCloseButton` property actually add/removes the button entirely which can cause the height of the modal's `<header>` to fluctuate if the height of the `header` slot's content is different from the height of the close button.

### Expected Behavior
The height of the `<header>` element does not fluctuate when toggling the `disableCloseButton` property.

### Reproduction Steps
```html
<calcite-modal active>
    <h3 slot="header">Modal title</h3>
    <button slot="primary" onclick="this.parentElement.disableCloseButton = !this.parentElement.disableCloseButton">
        Toggle close button
    </button>
</calcite-modal>
```

benelan:
It looks like the height is still changing on `next`. @Elijbet can you please take another look at this one.

https://codepen.io/benesri/pen/MWEjWeb?editors=1000

Elijbet:
> It looks like the height is still changing on `next`. @Elijbet can you please take another look at this one.
> 
> https://codepen.io/benesri/pen/MWEjWeb?editors=1000

Hmm... can't seem to reproduce it on my end. I looked at it on a larger screen, but don't see the height change.

benelan:
Weird, @nwhittaker can you repro with the sample above? Maybe I'm just seeing things.

nwhittaker:
>Weird, @nwhittaker can you repro with the sample above? Maybe I'm just seeing things.

I can't repro the issue in https://codepen.io/benesri/pen/MWEjWeb?editors=1000 in the latest versions of Chrome, Edge, Firefox, and Safari on macOS. Specifically, I don't see the height change anymore when the calcite-modal's `scale` attribute is a valid value.

I do see a hight change when the `scale` attribute is set to an invalid or empty value, but I'm not sure that's in scope or what the expected behavior would be anyway.

benelan:
![cc_issue_1707](https://user-images.githubusercontent.com/10986395/145622107-d895e991-0778-4751-a5fa-12f55b9eb2fe.gif)
[...]

PR #3619, opened 2021-12-02
fix(modal): modal now doesn’t shrink when close button disabled

Elijbet:
**Related Issue:** #1707

## Summary

The error was occurring because icon is watching scale prop from modal, where “m” is set as a default. But the check for the scale on the icon was not accounting for medium scale, resulting in a large icon (and button) for a medium scale modal. Which made it shrink and expand when removing/adding the close button. The change in check now makes the modal scale reflect on icon for a corresponding scale.

https://user-images.githubusercontent.com/19231036/144491453-7321e9a1-e354-441f-8200-9d34a781c88d.mov


driskull (approved):
Nice 👍

PR #3620, opened 2021-12-03
chore(ci): prepush hook checks for a conventional commit on the branch

benelan:
**Related Issue:** NA

## Summary
- You are allowed to push if there is at least one conventional commit on a branch
- pushes to master are exempt, but they still prompt to make sure its on purpose (version releases are non conventional commits)

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,11 +47,11 @@
     "util:copy-icons": "cpy ./node_modules/@esri/calcite-ui-icons/js/*.json ./src/components/calcite-icon/assets/calcite-icon/",
     "util:deploy-next-from-existing-build": "npm run util:prep-next-from-existing-build && npm run util:push-tags && npm run util:publish-next",
     "util:deploy-next-from-ci": "ts-node --project ./tsconfig-node-scripts.json support/deployNextFromCI.ts",
+    "util:check-squash-mergeable-branch": "ts-node --project ./tsconfig-node-scripts.json support/checkSquashMergeableBranch.ts",
     "util:prep-next-from-existing-build": "ts-node --project ./tsconfig-node-scripts.json support/prepReleaseCommit.ts --next",
     "util:publish-next": "npm publish --tag next",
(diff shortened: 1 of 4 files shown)


benelan:
This actually isn't ready to install I could use some help @jcfranco. The regex I have in this PR isn't great, the scope can't have dashes or commas. I tried a bunch of other stuff that works in online editors but not in the bash script, eg https://regex101.com/r/1kURw0/1

Bash uses Extended Regular Expression syntax right? Or am I missing something

package.json
```diff
@@ -50,7 +50,8 @@
     "util:prep-next-from-existing-build": "ts-node --project ./tsconfig-node-scripts.json support/prepReleaseCommit.ts --next",
     "util:publish-next": "npm publish --tag next",
     "util:push-tags": "git push --follow-tags",
-    "util:run-tests": "npm run util:copy-icons && stencil test --no-docs --spec --e2e"
+    "util:run-tests": "npm run util:copy-icons && stencil test --no-docs --spec --e2e",
+    "util:conventional-commit": "ts-node --project ./tsconfig-node-scripts.json support/matchConventionalCommit.ts"
```
jcfranco, line 54: Nitpick: place in alphabetical order

jcfranco (approved):
Looks great! Can the util and script be renamed to indicate that it only checks/runs on single-commit branches to ensure a conventional commit message when squash-merged? Now that's a mouthful. 😂

benelan:
> Can the util and script be renamed to indicate that it only checks/runs on single-commit branches to ensure a conventional commit message when squash-merged? Now that's a mouthful. 😂

what about something like `isOneCommitBranchConventional.ts` and `util:is-one-commit-branch-conventional`?

jcfranco:
> Edit: Something like isBranchSquashMergeable might be better because it also passes if there are 2+ commits so its not specific to single-commit branches.

I like where the name is going. One more suggestion: `checkSquashMergeableBranch`?

Issue #3381, opened 2021-10-29
Bug: Inconsistent event detail returned from `calciteInputInput` when using `clearable`


macandcheese:
### Actual Behavior
Using the `clearable` affordances (`esc` key, "x button") on an input causes inconsistent value to be set: `null` instead of `""`.

### Expected Behavior

Consistently value - `""` for both use cases.

### Reproduction Steps and Sample

_Sample_:

JSBIN: https://jsbin.com/puwuvuxofa/1/edit?html,js,console,output
Step 1: enter text into input
Step 2: delete text with "delete" key.
Step 3: notice that upon removal of all text, `""` is the input's value.
Step 4: enter text into input.
Step 5: clear text with `esc` key or by clicking the "x" affordance
Step 6: notice that the input value is now `null` - different than `""` above.
### Relevant Info

_Reproducible version_: `1.0.0-beta.68`

- [ ] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-<version>`

Related issue: #2060

jcfranco:
Possibly related to https://github.com/Esri/calcite-components/issues/1856

PR #3627, opened 2021-12-03
fix(input): ensure input's value is an empty string when cleared or set to null/undefined

jcfranco:
**Related Issue:** #3381

## Summary

This adjusts the approach taken for https://github.com/Esri/calcite-components/issues/2060 where value would support being cleared with `null`/`undefined`. `value` will now be a string to make it easier for end-users and align with expectations when compared to `<input>`.

### DIFF
--- a/src/components/calcite-input/calcite-input.tsx
+++ b/src/components/calcite-input/calcite-input.tsx
@@ -195,10 +195,15 @@ export class CalciteInput implements LabelableComponent, FormComponent {
     | "week" = "text";
 
   /** input value */
-  @Prop({ mutable: true }) value: string;
+  @Prop({ mutable: true }) value = "";
 
   @Watch("value")
   valueWatcher(newValue: string): void {
+    if (newValue == null) {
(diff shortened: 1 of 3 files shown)


driskull (approved):
👍 nice work

src/components/calcite-input/calcite-input.tsx
```diff
@@ -195,10 +195,15 @@ export class CalciteInput implements LabelableComponent, FormComponent {
...
-  @Prop({ mutable: true }) value: string;
+  @Prop({ mutable: true }) value = "";
 
   @Watch("value")
   valueWatcher(newValue: string): void {
+    if (newValue == null) {
+      this.value = "";
+      return;
+    }
+
```
eriklharper, line 206: Does this statement need to handle `undefined` as well?

jcfranco: It already handles it since this isn't using strict equality.

Issue #3317, opened 2021-10-20
Bug: calcite-input-date-picker min and max not updating


joeyHarig:
### Actual Behavior

When the properties `min` or `max` are  updated on a `calcite-input-date-picker`, the min and max values within the date picker are not updated.

### Expected Behavior

The date picker update min and max values when they are updated on the `calcite-input-date-picker`

### Reproduction Steps and Sample
The following code pen includes steps to reproduce this bug.
https://codepen.io/joeyHarig/pen/oNezexE?editors=1001

### Relevant Info

_Reproducible version_: `@esri/calcite-components@v1.0.0-beta.67`

Elijbet:
I think there might be a misunderstanding of how the date-picker works. These are 2 independent fields, one with min attribute and one without. I wouldn’t expect picking a date on the one without to set the minimum attribute on the other. Am I misunderstanding the issue?

joeyHarig:
@Elijbet In the codepen I'm setting the min for input-2 after a date select form input-1. 

The issue is that the 'min' value being passed to the date-picker from the input isn't being updated when the 'min' value on the input is updated.

jcfranco:
@Elijbet It looks like the issue is that the date-picker isn't updating internally when `min` is updated after initialization.

benelan:
verified on next

PR #3630, opened 2021-12-04
fix(input-date-picker): date-picker is now updating internally when min is updated after initialization

Elijbet:
**Related Issue:** #3317

## Summary

The min/max prop changes weren’t reflected on the input-date-picker after initialization because there was no watcher on the prop. I added one and also reflected it back to attribute.

See the description of the initial issue and fixed input-date-picker below:

https://user-images.githubusercontent.com/19231036/144690189-122b3e78-13c6-43a0-8823-754e7d9141b3.mov


jcfranco (changes requested):
Looking good, but this needs a test.

jcfranco (approved):
Awesome! 🎉

PR #3632, opened 2021-12-06
fix(radio-group): fix issue caused by radio group emitting change events from internal/programmatic changes

jcfranco:
**Related Issue:** #3120 

## Summary

This fixes an issue where multiple  `calcite-radio-group` change events would fire from internal changes.

### DIFF
--- a/src/components/calcite-radio-group/calcite-radio-group.tsx
+++ b/src/components/calcite-radio-group/calcite-radio-group.tsx
@@ -111,7 +111,7 @@ export class CalciteRadioGroup implements LabelableComponent, FormComponent {
   //
   //--------------------------------------------------------------------------
 
-  connectedCallback(): void {
+  componentWillLoad(): void {
     const items = this.getItems();
     const lastChecked = Array.from(items)
       .filter((item) => item.checked)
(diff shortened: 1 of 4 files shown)


driskull (approved):
Looks good! 👍 🦖

jcfranco:
Whoops, forgot to add a test for this. Coming up! 🚧

src/components/calcite-radio-group/calcite-radio-group.tsx
```diff
@@ -112,17 +111,12 @@ export class CalciteRadioGroup implements LabelableComponent, FormComponent {
...
   //--------------------------------------------------------------------------
 
-  connectedCallback(): void {
-    const items = this.getItems();
-    const lastChecked = Array.from(items)
-      .filter((item) => item.checked)
-      .pop();
+  componentWillLoad(): void {
+    this.selectInitialItem();
+  }
```
eriklharper, line 116: Is this part of the behavior spec for this component?  This would mean that there is no way to make this component optional in a form, and it diverts from native radio behavior.

jcfranco: The name is a bit misleading, it will only select an item if it has the checked property. Otherwise, it makes the first option tabbable.

driskull:
@jcfranco this broke the defaultValue in beta 71. The defaultValue needs to be set on connectedCallback

Issue #2720, opened 2021-08-03
Bug: calcite-popover close-button partially hidden


AdelheidF:
### Actual Behavior

wrong:
![image](https://user-images.githubusercontent.com/44481292/128097680-edaccb5c-f38a-4378-8725-92d2a915ebc8.png)

correct:
![image](https://user-images.githubusercontent.com/44481292/128097707-b16f7c84-8e90-49ff-ac8d-276878112ed8.png)

![image](https://user-images.githubusercontent.com/44481292/128097912-a55a4ad8-26df-4b1a-9fac-602a6d2dbaae.png)

![image](https://user-images.githubusercontent.com/44481292/128097950-ef180682-6221-47b8-b4c8-868c230f56b8.png)

I think it has something to do with the exact length of the title string 'Verwendung von A mit B vergleichen'.

Reproducible in Chrome, FF, Safari, Edge Chromium

------------------
from @driskull 
putting an overflow:hidden on the shadowRoot .content div seems to fix it

### Reproduction Steps or Sample

https://codepen.io/afreitag/pen/RwLrvKe?editors=1000

### Relevant Info

_Version_: `@esri/calcite-components@next.261`

- [ ] CDN
- [ ] NPM package

driskull:
Note:

[4:11 PM] Matt Driscoll
    putting an overflow:hidden on the shadowRoot .content div seems to fix it
​[4:11 PM] Matt Driscoll
    Seems like a component issue. Its missing some styles to handle a bunch of text content I think

benelan:
Hi @AdelheidF, I have been [unable to repro](https://codepen.io/benesri/pen/GRMoYgY?editors=1000) this issue in codepen, can you give it a try? It may be related to something else in your code besides the `heading` length.

AdelheidF:
https://codepen.io/afreitag/pen/RwLrvKe?editors=1000

jcfranco:
Verified with [updated codepen](https://codepen.io/jcfranco/pen/MWEVygX?editors=1000)!

PR #3635, opened 2021-12-06
fix(popover): close button no longer partially hidden

benelan:
**Related Issue:** #2720 

## Summary
I was unable to reproduce this issue locally or on a codepen so I'm not sure if this resolves the issue. @driskull how did you repro and find this solution?

### DIFF
--- a/src/components/calcite-popover/calcite-popover.scss
+++ b/src/components/calcite-popover/calcite-popover.scss
@@ -67,6 +67,7 @@
   flex-col
   flex-no-wrap
   self-center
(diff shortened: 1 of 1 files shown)


driskull:
Just a long title string I think

benelan:
I was able to verify that the fix works using the [sample](https://codepen.io/afreitag/pen/RwLrvKe?editors=1000) Adelheid provided in the issue. This PR is ready for review.

driskull (approved):

Issue #2079, opened 2021-04-29
Use CSS Logical Properties, remove dir usage


caripizza:
**Summary**
Refactor component styles to use CSS Logical Properties and Values, perhaps via Tailwind v2 plugin
> Start using flow-relative CSS values like start/end (This would get rid of 99% of > the use for `getElementDir()`)

**Relevant Info**
- Background discussion: https://github.com/Esri/calcite-components/issues/1831#issuecomment-828786269
- MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties
- Tailwind plugins research: https://github.com/stevecochrane/tailwindcss-logical, https://github.com/stevecochrane/tailwindcss-logical#flow-relative-margins

## Components

- [x] calcite-accordion-item/calcite-accordion-item.tsx
- [x] calcite-action-pad/calcite-action-pad.tsx
- [x] calcite-alert/calcite-alert.tsx
- [x] calcite-avatar/calcite-avatar.tsx
- [x] calcite-button/calcite-button.tsx
- [x] calcite-card/calcite-card.tsx
- [x] calcite-chip/calcite-chip.tsx
- [x] calcite-color-picker/calcite-color-picker.tsx
- [x] calcite-color-picker-hex-input/calcite-color-picker-hex-input.tsx
- [x] calcite-combobox/calcite-combobox.tsx
- [x] calcite-combobox-item/calcite-combobox-item.tsx
- [x] calcite-combobox-item-group/calcite-combobox-item-group.tsx
- [x] calcite-date-picker/calcite-date-picker.tsx
- [x] calcite-date-picker-day/calcite-date-picker-day.tsx
- [x] calcite-date-picker-month/calcite-date-picker-month.tsx
- [x] calcite-date-picker-month-header/calcite-date-picker-month-header.tsx
- [x] calcite-dropdown/calcite-dropdown.tsx
- [x] calcite-dropdown-group/calcite-dropdown-group.tsx
- [x] calcite-dropdown-item/calcite-dropdown-item.tsx
- [x] calcite-fab/calcite-fab.tsx
- [x] calcite-filter/calcite-filter.tsx
- [x] calcite-icon/calcite-icon.tsx
- [x] calcite-input/calcite-input.tsx
- [x] calcite-input-date-picker/calcite-input-date-picker.tsx
- [x] calcite-input-message/calcite-input-message.tsx
- [x] calcite-label/calcite-label.tsx
[...]

PR #3636, opened 2021-12-06
refactor: Remove setting 'dir' attribute within components

driskull:
**Related Issue:** #2079

## Summary

refactor: Remove setting 'dir' attribute within components.

This removes `dir=` within components and removes `getElementDir` usage if that was the only use of the utility.

The next step is to individually update components that use the `getElementDir` to apply a RTL class. This will involve using CSS logical properties in place of that.


jcfranco (commented):
Changes look good. 

Let's do the logical prop changes in this PR as well to avoid breakage between next releases.

driskull:
Some things still need the utility because there's not a CSS logical property for them yet or no way to handle it. It's mostly animations. For example...

- The `calcite-link` underline animation uses CSS `background-position` which has no alternative
- The `accordion-item` uses CSS transform rotation that goes opposite direction in RTL.

For now, i've kept the utility there but we could always say that animations won't change in RTL? @jcfranco

paulcpederson:
wow this really cleans stuff up, nice work @driskull 🥂

src/components/calcite-select/calcite-select.scss
```diff
@@ -19,8 +19,7 @@
 :host([scale="s"]) {
   @apply h-6;
   --calcite-select-font-size: theme("fontSize.-2h");
-  --calcite-select-spacing: 0 theme("spacing.8") 0 theme("spacing.2");
-  --calcite-select-spacing-rtl: 0 theme("spacing.2") 0 theme("spacing.8");
+  --calcite-select-spacing-inline: theme("spacing.8") theme("spacing.2");
```
driskull, line 22: I think these are backwards :)

src/components/calcite-select/calcite-select.scss
```diff
@@ -104,23 +96,15 @@ select:disabled {
...
     inset-y-0;
-}
-
-.icon-container.calcite--rtl {
-  @apply border-l
-    border-r-0 
-    left-0;
-  right: unset;
+  inset-inline-end: 0;
+  border-inline-width: 0 1px;
```
driskull, line 107: I think these are backwards too? Should we use `theme("borderWidth.default")` here?

src/components/calcite-select/calcite-select.tsx
```diff
@@ -330,29 +329,27 @@ export class CalciteSelect implements LabelableComponent, FormComponent {
   //
   //--------------------------------------------------------------------------
 
-  renderChevron(dir: Direction): VNode {
+  renderChevron(): VNode {
     return (
-      <div class={{ [CSS.iconContainer]: true, [CSS_UTILITY.rtl]: dir === "rtl" }}>
+      <div class={{ [CSS.iconContainer]: true }}>
```
driskull, line 334: should we change to `class={CSS.iconContainer]}`

src/components/calcite-shell-center-row/calcite-shell-center-row.scss
```diff
@@ -56,17 +56,10 @@
...
-}
-
-.calcite--rtl {
-  ::slotted(calcite-action-bar) {
-    @apply border-r-0 border-l border-solid border-color-3;
-  }
-  ::slotted(calcite-action-bar[position="end"]) {
-    @apply border-l-0 border-r border-solid border-color-3;
-  }
+  border-inline-width: theme("borderWidth.0") theme("borderWidth.default");
```
driskull, line 64: reversed I think

src/components/calcite-stepper-item/calcite-stepper-item.scss
```diff
@@ -73,20 +71,11 @@
...
-    text-left
-    duration-150
-    ease-in-out;
-}
-
-.calcite--rtl .stepper-item-content,
-.calcite--rtl .stepper-item-header {
-  @apply pr-0
-    text-right
+  @apply text-left
```
driskull, line 74: should this be `text-align:start`?

src/components/calcite-stepper-item/calcite-stepper-item.scss
```diff
@@ -73,20 +71,11 @@
...
     duration-150
     ease-in-out;
-  padding-left: initial;
+  padding-block: var(--calcite-stepper-item-spacing-unit-l);
+  padding-inline: 0 var(--calcite-stepper-item-spacing-unit-m);
```
driskull, line 78: reversed I think

src/components/calcite-stepper-item/calcite-stepper-item.scss
```diff
@@ -229,17 +202,17 @@
     border-solid
     border-color-3
     py-0
-    pr-0
     mx-0
     mt-0;
-  padding-left: var(--calcite-stepper-item-spacing-unit-l);
+  padding-inline: var(--calcite-stepper-item-spacing-unit-l) 0;
```
driskull, line 207: this one is correct I think.

src/components/calcite-select/calcite-select.scss
```diff
@@ -75,7 +71,8 @@
     box-border
     truncate;
   font-size: var(--calcite-select-font-size);
-  padding: var(--calcite-select-spacing);
+  padding-inline: var(--calcite-select-spacing-inline);
```
driskull, line 74: @jcfranco I think this needs to change. padding `!==` padding-inline. `padding-inline` doesn't set top/bottom whereas padding does. Currently, this isn't right because the previous spacing variable set top/bottom/left/right.

jcfranco: I only set `padding-inline` and  left out the top/bottom padding because it was being set to 0 (the default).

src/components/calcite-notice/calcite-notice.scss
```diff
@@ -132,11 +127,11 @@
   @include notice-element-base;
   @apply flex flex-col min-w-0 break-words;
   flex: 1 1 0;
-  padding: var(--calcite-notice-spacing-token-small) var(--calcite-notice-spacing-token-small)
-    var(--calcite-notice-spacing-token-small) 0;
+  padding: var(--calcite-notice-spacing-token-small) 0;
```
jcfranco, line 130: Since we are setting the padding-inline below, this should be:

```scss
padding-block: var(--calcite-notice-spacing-token-small);
```

src/components/calcite-panel/calcite-panel.scss
```diff
@@ -131,11 +127,11 @@
 }
 
 .header-actions--end {
-  @apply ml-auto;
+  margin-inline-start: auto;
 }
 
 .menu-container:only-child {
-  @apply ml-auto;
+  margin-inline-start: auto;
```
jcfranco, line 134: Nitpick: this could be merged into the above selector.

src/components/calcite-panel/calcite-panel.tsx
```diff
@@ -424,16 +423,13 @@ export class CalcitePanel {
...
   render(): VNode {
-    const { dismissed, disabled, dismissible, el, loading, panelKeyDownHandler } = this;
-
-    const rtl = getElementDir(el) === "rtl";
+    const { dismissed, disabled, dismissible, loading, panelKeyDownHandler } = this;
 
     const panelNode = (
       <article
         aria-busy={loading.toString()}
         class={{
```
jcfranco, line 431: This can be simplified.

src/components/calcite-rating/calcite-rating.tsx
```diff
@@ -177,12 +175,11 @@ export class CalciteRating implements LabelableComponent, FormComponent {
 
   render() {
     const { intlRating, showChip, scale, count, average } = this;
-    const dir = getElementDir(this.el);
 
     return (
       <Fragment>
         <fieldset
-          class={{ fieldset: true, [CSS_UTILITY.rtl]: dir === "rtl" }}
+          class={{ fieldset: true }}
```
jcfranco, line 182: This can be simplified.

src/utils/dom.ts
```diff
@@ -27,17 +27,7 @@ export function getThemeName(el: HTMLElement): "light" | "dark" {
 }
 
 export function getElementStyleDir(el: HTMLElement): Direction {
```
jcfranco, line 29: Sidebar: [not for this PR] 2 things regarding this util,

1. I think this should still go through the property. [From MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/direction#:~:text=Note%20that%20text%20direction%20is%20usually%20defined%20within%20a%20document%20(e.g.%2C%20with%20HTML%27s%20dir%20attribute)%20rather%20than%20through%20direct%20use%20of%20the%20direction%20property.):

> Note that text direction is usually defined within a document (e.g., with HTML's dir attribute) rather than through direct use of the direction property.

An additional reason would be that it may also be more performant since the browser doesn't have to compute the styles (including unrelated ones to the direction).

2. Since we're going to stick with one util, the name can be `getElementDir` as the origin of the direction is an implementation detail.

jcfranco (approved):
Other than a few comments, this LGTM!

Truly awesome stuff here! This will simplify the codebase and make styles leaner.

src/utils/dom.ts
driskull: 1. I think we should use the computed style. Otherwise, if someone does set the style we wouldn't know. I see it as equivalent to the `hidden` property and `display:none`. Even though MDN advises to set the style via an attribute, it says "usually" not "should".

2. I agree that we should rename to `getElementDir`. I can do that...

jcfranco: WRT 1. I get your point, but how common is setting direction via HTML vs CSS? I think this is what MDN is hinting at. The W3C also [recommends against setting direction via CSS](https://www.w3.org/International/questions/qa-html-dir#no_css):

> ### Don't use CSS for direction!
> Do _not_ use CSS to apply base direction in HTML pages.
> 
> Basically, this is because you want the directional information to be available even when the CSS is not. The directional information can affect the semantics of your content, and so should be part of the markup. (See a [longer explanation](https://www.w3.org/International/questions/qa-bidi-css-markup)).
> 
> Both the CSS and HTML specs echo this same admonition.

src/components/calcite-accordion-item/calcite-accordion-item.scss
```diff
@@ -83,14 +83,10 @@
     text-color-3
     duration-150
     ease-in-out;
-  margin-right: var(--calcite-accordion-item-icon-spacing-start);
-  margin-left: var(--calcite-accordion-item-icon-spacing-end);
+  margin-inline-end: var(--calcite-accordion-item-icon-spacing-start);
```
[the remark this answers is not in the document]
driskull: I chose to do these like this because the variable says start/end but its being used for the opposite sometimes. I think this could be refactored more in the future.

src/components/calcite-chip/calcite-chip.scss
```diff
@@ -95,7 +74,10 @@
     transition-default
     text-color-1;
   -webkit-appearance: none;
-  border-radius: var(--calcite-chip-button-border-radius);
+  border-start-start-radius: 0;
+  border-start-end-radius: 50px;
```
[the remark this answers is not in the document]
driskull: I removed this internal variable, but I can put it back...

`--calcite-chip-button-border-radius: 0 50px 50px 0;`

src/components/calcite-label/calcite-label.e2e.ts
```diff
@@ -65,7 +65,7 @@ describe("calcite-label", () => {
             });
             element = await page.find("calcite-label >>> .container");
             style = await element.getComputedStyle();
-            expect(style["textAlign"]).toEqual("left");
+            expect(style["textAlign"]).toEqual("start");
```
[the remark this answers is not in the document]
eriklharper: I'd be happy to help on that.

src/components/calcite-chip/calcite-chip.scss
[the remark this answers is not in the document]
driskull: actually, I can't because there's not a shorthand equivalent. I'll leave for now.

src/components/calcite-combobox-item-group/calcite-combobox-item-group.scss
```diff
@@ -46,13 +32,11 @@
 }
 
 .label--indent-1 {
-  padding-left: var(--calcite-combobox-item-indent-start-1);
-  padding-right: var(--calcite-combobox-item-indent-end-1);
+  padding-inline-start: var(--calcite-combobox-item-spacing-indent-1);
```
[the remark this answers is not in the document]
driskull: No, because it was just `unset` in this case. Its because of the weird use case here using a CSS var

src/components/calcite-dropdown/calcite-dropdown.scss
```diff
@@ -43,25 +43,3 @@
 :host([width="l"]) {
   --calcite-dropdown-width: theme("spacing.64");
 }
-// scale
-:host([scale="s"]) {
-  --calcite-dropdown-group-padding: #{theme("spacing.2") 0};
```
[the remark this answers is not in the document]
driskull: Removed the variable `--calcite-dropdown-item-padding`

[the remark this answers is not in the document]
driskull: OH, no these weren't used anywhere

src/components/calcite-panel/calcite-panel.scss
eriklharper: Tailwind doesn't provide an equivalent selector for `margin-inline-start` including many if not all of the other CSS logical properties.

src/utils/dom.ts
driskull: I get your point too :)

But, previously, we had to pierce multiple shadow DOMs to look for a `[dir]` attribute, I'm not sure how performant that is anyway. (using .closest() crossing shadow).

It seems like we would be better off just seeing if the host has a direction style IMO.

eriklharper: I guess the case of when CSS wouldn't be available would be screen readers.  Is there any other case where CSS wouldn't be available?

src/components/calcite-chip/calcite-chip.scss
```diff
@@ -128,16 +110,14 @@
...
-  margin-left: var(--calcite-chip-spacing-unit-l);
-  border-radius: var(--calcite-chip-button-border-radius);
-}
-
-.calcite--rtl .calcite-chip--icon {
-  @apply my-0 ml-0;
-  margin-right: var(--calcite-chip-spacing-unit-l);
+  margin-inline-end: 0;
+  margin-inline-start: var(--calcite-chip-spacing-unit-l);
+  border-start-start-radius: 0;
```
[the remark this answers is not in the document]
driskull: border-radius doesn't have a CSS logical shorthand so it needs to be this way.

src/utils/dom.ts
jcfranco: > But, previously, we had to pierce multiple shadow DOMs to look for a [dir] attribute, I'm not sure how performant that is anyway. (using .closest() crossing shadow).

One key difference is that computing the style of any element will cause [layout thrashing](https://gist.github.com/paulirish/5d52fb081b3570c81e3a), traversing the DOM will not.

> I guess the case of when CSS wouldn't be available would be screen readers. 

I may have misunderstood your comment, but I don't think this is correct. CSS does affect what screen readers interpret. For example, any readable element targeted by a selector with `display: none` won't be read by screen readers. Let me know if I missed something.

driskull: @jcfranco would you like me to put the previous function back? Its not as big of deal now that its not used as much.

jcfranco: Yes, please.

eriklharper: That makes sense about display:none with screen readers.  I was mostly asking that out of curiosity.  I was wondering what would be the typical use cases of concern with regards to not having CSS available.  I suppose the main benefit is that there wouldn't be any layout thrashing or FOUC (flash of unstyled content) or something similar when you don't rely on computed styles to set the UI based on direction.

Issue #3495, opened 2021-11-12
calcite-action-bar inside shell-center-row expand tooltip doesn't work at first


AdelheidF:
When adding the action-bar programmatically the expand tooltip doesn't work initially. Once you expand and collapse the action-bar the expand tooltip starts working.

![image](https://user-images.githubusercontent.com/44481292/141387420-4f57b38f-26b1-47cc-b7da-0ced1c35dace.png)

cc @driskull 

### Actual Behavior

expand tooltip doesn't work initially

### Expected Behavior

expand tooltip should work from the start

### Reproduction Steps and Sample

https://codepen.io/afreitag/pen/RwZBxKa

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-beta.69`

- [ ] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-<version>`

driskull:
This is due to the use of the getSlotted utility...

The utility dom.getSlotted() is great, but one problem is that if a user slots an element after initialization using JS, the getSlotted() method doesn't know about the newly slotted element until a re-render occurs. 
 
We might want to setup a mutation listener to re-render for those scenarios.

>>> Yeah, that makes sense. There is also a slot event when its content changes that we could listen for to rerender.

driskull:
@benelan @jcfranco Can you triage this? I think we need a utility refactor first.

benelan:
This issue is currently in the next sprint starting 12/20. Do you think it needs to be in the current sprint starting today?

driskull:
I think that's fine.

driskull:
@jcfranco Is this the approach we want to take?

```tsx
<slot name={SLOTS.expandTooltip} onSlotchange={() => forceUpdate(this)} />
```

It would look something like this...

```tsx
<calcite-action-group class={CSS.actionGroupBottom} scale={scale}>
        <slot name={SLOTS.bottomActions} />
        <slot name={SLOTS.expandTooltip} onSlotchange={this.slotChanged} />
        {expandToggleNode}
      </calcite-action-group>
```

```tsx
slotChanged = (): void => {
    forceUpdate(this);
  };
```

driskull:
This would also involve going through any component that uses `getSlotted` and adding an `onSlotchange` listener to call forceUpdate.

jcfranco:
[...]

PR #3642, opened 2021-12-07
fix(action-bar): Allow slotting tooltip after initialization

driskull:
**Related Issue:** #3495

## Summary

fix(action-bar): Allow slotting tooltip after initialization.  #3495

- Creates utility to connect and disconnect the slotchange event on a component
- The utility will watch for slotchange events and re-render.


eriklharper (approved):
Might be good to add a visual screenshot test to verify this works after setting a delay perhaps?

jcfranco (approved):
Looks great! Can you add a test before merging?

driskull:
@jcfranco @benelan Can we get an issue/epic to update other components that conditionally use getSlotted() to add this?

benelan:
> @jcfranco @benelan Can we get an issue/epic to update other components that conditionally use getSlotted() to add this?

Sorry, just saw this while verifying. Yeah I'll log an issue.

PR #3644, opened 2021-12-07
refactor(radio-group-item): remove input proxy

jcfranco:
**Related Issue:** N/A

## Summary

Removes the now unsupported input proxy from `calcite-radio-group-item`.


driskull (approved):
👍

Issue #3500, opened 2021-11-12
Panel is dismissed on escape keyup rather than keydown


ellenupp:
### Actual Behavior

A dismissible panel is closed with 'escape' on keyup, while other components like dropdown and combobox are closed on escape keydown. 

### Expected Behavior

Panel should also be dismissed on escape keydown for consistency. Using the same event would make it simpler to stop propagation of that event in the case you have for example a dropdown inside a dismissible panel, where closing the dropdown using escape should not also close the panel. 

### Reproduction Steps and Sample

_Sample_: https://codepen.io/ellupp/pen/vYJzQWQ?editors=1000

1. Open dropdown. Press and hold escape key. Dropdown closes on keydown. Release escape key. Panel is dismissed on keyup.

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-beta.69`

benelan:
verified on next

PR #3657, opened 2021-12-08
fix(panel): the panel is now dismissed on escape keydown and not keyu…

Elijbet:
…p for consistency

**Related Issue:** #3500

## Summary

Since other components like dropdown and combobox are closed on escape keydown, this needs to follow the same behavior for consistency and to make it simpler to stop propagation of keyup event in case we have a nested component that closes on keydown.


driskull (approved):
👍 Looks good! Thanks @Elijbet 🎉 🌮

Issue #3597, opened 2021-11-30
Combobox: Focus styles remain after clicking another combobox


paulcpederson:
### Actual Behavior

Clicking on labels for multiple comboboxes leaves the focus style on all of them.

### Expected Behavior

Clicking the label of any combobox or form control should  remove focus style from now unfocused combobox.

### Reproduction Steps and Sample

_Sample_: https://codepen.io/paulcp/pen/qBPEEao?editors=1000

1. Click the label text of each combobox
2. Observe all three have focus styles

![Screen Shot 2021-11-30 at 11 10 59 AM](https://user-images.githubusercontent.com/1031758/144112189-6144cb7e-8637-407b-85d3-b92354c510f2.png)

### Relevant Info

This exists in both 66 and 70 and probably others. It may have always been broken like this.

_Reproducible version_: `@esri/calcite-components@1.0.0-70`

- [x] CDN
- [x] NPM package

Regression?: no

paulcpederson:
@Elijbet are you working on this already? I think I have a fix for it I could pr

benelan:
verified on next

PR #3661, opened 2021-12-08
fix(combobox): fix focus behavior inside calcite label (#3597)

paulcpederson:
**Related Issue:** #3597

## Summary

Fix bug where inside calcite label the combobox would retain focus style after blur.


driskull (approved):
👍

jcfranco:
Thanks for the fix, @paulcpederson!

Issue #3543, opened 2021-11-19
Deprecate calcite-radio by moving markup and styles into calcite-radio-button


eriklharper:
### Description 

Now that `calcite-radio-button` is shadowed, the `calcite-radio` component, which originally existed purely to encapsulate the Esri radio branding behind a shadow root so that `calcite-radio-button` could expose its child native input as a scoped component (for native form integration), is no longer needed.

### Proposed Advantages 

This ticket will deprecate the separate internal `calcite-radio` component, removing an unnecessary separate element, and add the markup and styling for the UI it produces into `calcite-radio-button`.

### Relevant Info 

During this process would also be a good opportunity to refactor the CSS to move away from attribute selectors into CSS classes, since `calcite-radio` relies heavily on these.

#### Which Component

`calcite-radio`, `calcite-radio-button`

benelan:
verified on next. `calcite-radio-button` doesn't look or behave differently than before

PR #3662, opened 2021-12-08
refactor(radio-button): remove internal calcite-radio component

eriklharper:
**Related Issue:** #3543

## Summary

This PR replaces the content of the internal `calcite-radio` component directly into `calcite-radio-button` itself.  This component existed prior to `radio-button`'s conversion to use shadow dom so that the radio styling could be encapsulated within a shadowRoot, so now that radio-button is shadowed, this component is redundant.


driskull:
I think `deprecate` is the wrong word. Shouldn't it be `remove`? :)

jcfranco (approved):
🎉 🌮 🦖

Issue #2451, opened 2021-07-07
Remove IE specific workarounds


paulcpederson:
### Background 

We have several workarounds for ie11 in the codebase, we can safely remove them now.

caripizza:
How can this one be verified?

julio8a:
Following up on Cari's question. Also adding @jcfranco and @driskull

jcfranco:
@driskull helped out by removing some IE11 chunks (🎉), but the following items remain:

1. https://github.com/Esri/calcite-components/blob/5cf0ccf4461c261bc74e99012cc0214950d70aba/src/utils/key.ts#L1-L4
2. https://github.com/Esri/calcite-components/blob/0a709cff3f4a10ca8e9cbd093f418e53c2b519ee/src/components/calcite-input/calcite-input.scss#L544-L550
3. https://github.com/Esri/calcite-components/blob/f86d27a89b25d2111d6194623db51c18356c3ff3/src/components/calcite-input/calcite-input.tsx#L524-L530
4. https://github.com/Esri/calcite-components/blob/857528c9ea9b0685d52aafd08aa83f5d899b7a33/src/components/calcite-radio-group-item/calcite-radio-group-item.tsx#L87-L92

I think we can mark this as verified once the above items are addressed. There may still be a few things left over, but we can tackle them in refactor PRs moving forward.

I'm gonna set this back to `assigned` and move it to the back-burner.

benelan:
@jcfranco can you please help verify this one? The code you commented snippets above were removed but I'm not sure if there's anything else that needs to be done.

jcfranco:
Verified locally! There is [one possible item](https://github.com/Esri/calcite-components/blob/28e122a5b2af171134b7cbcf2f1a4f9b0599c29c/src/components/calcite-stepper-item/calcite-stepper-item.tsx#L246) left, but I'll tackle that when I [refactor stepper](https://github.com/Esri/calcite-components/issues/1258).

PR #3664, opened 2021-12-08
refactor(input): remove IE specific functionality

y0n4:
**Related Issue:** #2451

## Summary
Removed IE specific functionality addressed code area at https://github.com/Esri/calcite-components/issues/2451#issuecomment-898780240. 1 has already been addressed


driskull (approved):
👏 🎉

jcfranco (changes requested):
> 1 and 3 has already been addressed

3 has not been addressed, but it looks like `stepUp` & `stepDown` are not supported in Firefox yet. Let's remove the comment and create a refactor issue instead.

y0n4:
oops, good catch thanks

Issue #3533, opened 2021-11-17
CalciteButton - incorrect layout with space-between


asangma:
### Actual Behavior
When Button has `alignment="space-between"` and `width=[ "half" | "full" ]`, the layout has extra space in the start area.
This seems to be caused by the loader still taking up space.

This may also cause issues with `icon-start-space-between` and `icon-end-space-between`.

### Expected Behavior
The loader shouldn't take up space if it's not visible.

### Reproduction Steps and Sample

_Sample_: https://codepen.io/mtb-alan/pen/oNemXBO

1. Set CalciteButton width to "half"
2. Set Calcite Button alignment to "space-between"
3. observe

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-<version>`

- [ ] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-<version>`

benelan:
verified on next

PR #3666, opened 2021-12-09
fix(button): no longer render loader when width set to half or full

anveshmekala:
**Related Issue:** #3533 

## Summary

This fixes the spaced occupied by `<div>` wrapper of `calcite-loader` in button when width is set to `half | full` and `alignment = space-between | icon-start-space-between | icon-end-space-between`

### DIFF
--- a/src/components/calcite-button/calcite-button.tsx
+++ b/src/components/calcite-button/calcite-button.tsx
@@ -151,7 +151,7 @@ export class CalciteButton implements LabelableComponent {
     const dir = getElementDir(this.el);
     const Tag = this.childElType;
 
-    const loader = this.hasLoader ? (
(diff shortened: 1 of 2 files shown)


src/components/calcite-button/calcite-button.tsx
```diff
@@ -151,19 +151,17 @@ export class CalciteButton implements LabelableComponent {
     const dir = getElementDir(this.el);
     const Tag = this.childElType;
 
-    const loader = (
+    const loader = this.hasLoader ? (
```
driskull, line 154: nitpick: can we name this const to `loaderNode`. Looks good!

driskull (approved):

Issue #3529, opened 2021-11-17
Calcite-button: when having empty icon-end or icon-start, it adds padding


yannikmesserli:
### Actual Behavior

```typescript
<calcite-button icon-end>Test</calcite-button>
```

Has extra padding at the end. 

### Expected Behavior

No padding

### Reproduction Steps and Sample

A sample code: https://codepen.io/yannikm/pen/VwzqEzZ?editors=100

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-<version>`

- [x] CDN
- [x] NPM package

Not a regression

benelan:
verified on next

PR #3677, opened 2021-12-09
fix(button): remove extra margin when icon is empty

anveshmekala:
**Related Issue:** #3529 

## Summary

This fix removes the `margin` added to the `.content `class when  `calcite-button` is passed with empty `icon-start | icon-end` prop

Ex:
`<calcite-button icon-end>Button</calcite-button>`

### DIFF
--- a/src/components/calcite-button/calcite-button.scss
+++ b/src/components/calcite-button/calcite-button.scss
@@ -62,12 +62,12 @@
 
 .icon-start-empty {
   .content {
-    margin-left: unset;
+    margin-inline-start: unset;
   }
 }
 .icon-end-empty {
   .content {
-    margin-right: unset;
+    margin-inline-end: unset;
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/components/calcite-button/calcite-button.scss
```diff
@@ -60,6 +60,17 @@
   margin-inline-end: var(--calcite-button-content-margin);
 }
 
+.icon-start-empty {
+  .content {
+    margin-left: unset;
```
driskull, line 65: @anveshmekala can you unset the CSS logical property here? not left/right values.

driskull (changes requested):

driskull (approved):
Looks good! Do we need a test or will screener cover this?

anveshmekala:
> Looks good! Do we need a test or will screener cover this?

Screener test should be a good I guess.

Issue #2928, opened 2021-08-27
Bug: calcite-slider with step:0.1 returns long decimals


AdelheidF:
I have a `calcite-slider` with step of 0.1, min/max = 0/5, snap=true.
`CalciteSliderUpdate` returns numbers like 1.9000000000000001 and 2.4000000000000004 in `event.target.value`.

### Reproduction Steps or Sample

1.

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

caripizza:
Adding guidance from @driskull when I asked about this issue in Teams yesterday:

> It seems like we should probably honor the precision defined in the step IMO

caripizza:
This doesn't just apply to the value coming through in the event, it also seems to apply to the tick values getting generated in slider's [`while` loop](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-slider/calcite-slider.tsx#L839).

While working on #2457, I noticed that when `ticks` is a decimal (ex: 0.1), some of the generated tick values end up like `-1.3877787807814457e-16`. Then, the tick value renders as `-0` after it's converted to a locale string. (Repro: https://codepen.io/caripizza-the-animator/pen/eYRLGrL)

It's like the slider's calculations are off due to the issue described here: https://basarat.gitbook.io/typescript/recap/number#decimal

driskull:
@caripizza we may need to try something like...

```js
const countDecimals = (value): number {
if(Math.floor(value) === value) return 0;
return value.toString().split(".")[1].length || 0;
}

current = (current + this.ticks).toFixed(countDecimals(this.step));
```

I think we should have a `@State stepDecimals: number;` that is generated when steps are set.

benelan:
What about using a library, e.g. https://github.com/MikeMcl/big.js/#bigjs

jcfranco:
@Elijbet Since you fixed a similar `calcite-input` bug, do you want to take a stab at this one?

Elijbet:
> @Elijbet Since you fixed a similar `calcite-input` bug, do you want to take a stab at this one?

Yep, that'd be great. Thanks!

benelan:
verified on next

PR #3678, opened 2021-12-09
fix(slider): slider now takes the precision of the step if it's provided or no decimal points if it isn’t; added demos for each study case

Elijbet:
**Related Issue:** #2928

## Summary

Fixes slider decimal steps returning long precision decimal. Slider now takes the precision of the step if it's provided or no decimal points if it isn’t; added demos for each study case.

https://user-images.githubusercontent.com/19231036/145484530-4c8c4968-08b4-4d90-a4fe-9e07971145d2.mov


driskull (approved):
Looks good @Elijbet ! Can we add a test for this? Some kind of e2e test?

driskull (approved):
👍

PR #3688, opened 2021-12-11
build: use postcss to ensure CSS logical prop usage

jcfranco:
**Related Issue:** N/A

## Summary

This wires up postcss + stylelint + https://www.npmjs.com/package/stylelint-use-logical-spec to ensure the CSS output uses [logical props](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties). 

Note that this is using the default configuration and will transform all properties that have a logical equivalent. We can tweak this if we need to prevent some properties from being transformed.

### DIFF
--- a/stencil.config.ts
+++ b/stencil.config.ts
@@ -5,6 +5,7 @@ import babel from "@rollup/plugin-babel";
 import autoprefixer from "autoprefixer";
 import tailwind from "tailwindcss";
 import { generatePreactTypes } from "./support/preact";
+import stylelint from "stylelint";
 
 export const create: () => Config = () => ({
   buildEs5: "prod",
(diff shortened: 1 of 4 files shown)


stencil.config.ts
```diff
@@ -99,7 +100,14 @@ export const create: () => Config = () => ({
       injectGlobalPaths: ["src/assets/styles/includes.scss"]
     }),
     postcss({
-      plugins: [tailwind(), autoprefixer()]
+      plugins: [
+        tailwind(),
+        autoprefixer(),
+        stylelint({
```
driskull, line 106: Dope 😎

driskull (approved):

jcfranco:
Failing test is unrelated. Merging!

PR #3690, opened 2021-12-13
fix(dropdown-item): align items with heading in selection-mode: none

Elijbet:
**Related Issue:** #3690

## Summary

Align items with heading in selection-mode:none.


driskull (approved):
🎉 :nice! Looks like screener test is good enough.

PR #3691, opened 2021-12-13
ci(action): issues in development do not become stale

benelan:
**Related Issue:** #1475

## Summary
Issues that are in development should not be marked as stale


driskull (approved):
👍

PR #3692, opened 2021-12-13
ci(e2e): skip e2e tests on markdown only prs

benelan:
**Related Issue:** NA

## Summary
- Implementing the same logic that is working for screener
- Unlike screener, the e2e will only be skipped on the PR and will still run on pushes to master after installing. 
    - This is because of the way the next deployments are set up, [e2e must pass in order to deploy next](https://github.com/Esri/calcite-components/blob/master/.github/workflows/deploy-next.yml#L22-L34). 
    - The deployment script is cancelled if a new install comes in so that multiple, identical deployments don't happen (and to batch multiple deployable installs into a single next release). 
    - Therefore, a next release can happen on a markdown only install because a previous deployable install was cancelled in favor of the new one.


.github/workflows/pr-e2e.yml
```diff
@@ -7,8 +7,27 @@ jobs:
...
+        id: one
+        run: |
+          current_branch=$(git rev-parse --abbrev-ref HEAD)
+          # diff of branch excluding md
+          testable_changes=$(git diff --name-only "$current_branch" $(git merge-base "$current_branch" origin/master) -- . ':(exclude)*.md')
+          echo "changed files: $testable_changes"
+          # skip if there are only md changes
+          if [ -z "$testable_changes" ]; then
+            echo "skip e2e"
+            echo "::set-output name=skip::skip"
```
jcfranco, line 22: Nitpick: does this output need a value set here if it is not used?

.github/workflows/pr-e2e.yml
```diff
@@ -7,8 +7,27 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: skip for markdown only prs
+        id: one
```
jcfranco, line 13: Nitpick: can this step have a more descriptive name?

jcfranco (approved):
💥

.github/workflows/pr-e2e.yml
benelan: God question, I'll look into that. I was just being safe because I didn't know what would happen to the conditional below if the variable was undefined

Issue #3687, opened 2021-12-10
CSS selector syntax in `id` attribute breaks `<calcite-tooltip>`


nwhittaker:
### Actual Behavior
If the tooltip's `reference-element` is an element `id` string that includes certain characters (e.g. periods, commas, etc...), the reference element is not found and a warning appears in the console.

### Expected Behavior
All valid `id` strings are supported and the reference element is found because the `reference-element` attribute is documented to support `id` strings, not a CSS selector string.

### Reproduction Steps and Sample

_Sample_: https://codepen.io/nwhittaker/pen/bGowLdM?editors=1000

1. Hover linked text and notice tooltip does not appear.

### Relevant Info
Consider passing the `id` string through `document.getElementById()` instead of `document.querySelector()`.

_Reproducible version_: `@esri/calcite-components@1.0.0-<next.322>`

- [ ] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-<version>`

benelan:
verified on `beta.74`

PR #3696, opened 2021-12-13
fix(popover, popover-manager, tooltip, tooltip-manager): Correctly query by ID within the DOM.

driskull:
**Related Issue:** #3687

## Summary

fix(popover, popover-manager, tooltip, tooltip-manager): Correctly query by ID within the DOM.

- Updates `dom.queryElementRoots` to accept either a selector or id.
- `id` will use `rootNode.getElementById(id)`


src/utils/dom.ts
```diff
@@ -78,7 +78,16 @@ export function queryElementsRoots<T extends Element = Element>(element: Element
...
 // Queries an element's rootNode and any ancestor rootNodes.
 // based on https://stackoverflow.com/q/54520554/194216
-export function queryElementRoots<T extends Element = Element>(element: Element, selector: string): T | null {
+export function queryElementRoots<T extends Element = Element>(
+  element: Element,
+  {
+    selector,
+    id
+  }: {
+    selector?: string;
```
jcfranco, line 87: Can you doc that `id` takes precedence if both are specified? It might not be clear from the signature alone.

jcfranco (approved):
🎉 🌮 🦖

Issue #3494, opened 2021-11-12
calcite-action-bar inside shell-center-row has thick gray border


AdelheidF:
This thick gay border should not be on the top, left, bottom of the action-bar
![image](https://user-images.githubusercontent.com/44481292/141386901-0dd5ca64-7661-495a-8f05-1d85c8d2c742.png)

cc @asangma 

### Actual Behavior

thick gray border

### Expected Behavior

no border

### Reproduction Steps and Sample

https://codepen.io/afreitag/pen/ZEJjaRa?editors=1000

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-beta.69`

- [ ] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-<version>`

AdelheidF:
@asangma

PR #3698, opened 2021-12-14
fix(action-bar): removed thick border around action-bar when nested in the shell-center-row

Elijbet:
**Related Issue:** #3494

## Summary

Removed unnecessary border from action bar inside shell center row.

action-bar-border


Elijbet:
@macandcheese  Removing the thick border on top, left, and bottom reveals that article header does not line up with action bar.

driskull (approved):
👍 Nice @Elijbet

src/components/calcite-shell-center-row/calcite-shell-center-row.scss
```diff
@@ -56,8 +56,8 @@
 }
 
 ::slotted(calcite-action-bar) {
-  @apply border-solid border-color-3;
-  border-inline-end-width: theme("borderWidth.default");
+  border-inline-end: 1px solid;
+  @apply border-color-3;
 }
 ::slotted(calcite-action-bar[position="end"]) {
   @apply border-solid border-color-3;
```
asangma, line 63: You'll need to do a similar style for `position="end"`.

driskull (approved):
👍

Issue #3425, opened 2021-11-03
A11y Audit: [calcite-date-month-header] - `.year` input is missing associated label describing its purpose


caripizza:
### Issue Summary
The [.year input](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-date-picker-month-header/calcite-date-picker-month-header.tsx#L144) in Calcite Date Month Header needs an associated aria-label.

From the Violation Instance:
[Issue]
There are form fields that do not provide a programmatically associated label that describes the purpose of each field. 

[User Impact]
Users may not be able to fill out form fields correctly if the visible label is not programmatically associated with its field.

[Code Reference]
`<input class="year" inputmode="numeric" maxlength="4" minlength="1" pattern="\d*" type="text">`

### Related Standards
https://amp.levelaccess.net/public/standards/view_best_practice.php?violation_id=338&report_id=526424

### Reproduction Steps
Create a new Date Picker or Input Date Picker and inspect the input with class .year, notice there is no related aria-label describing the purpose of the input field.

```html
<calcite-date-picker value="2020-03-27" min="2020-02-01" max="2021-01-01"></calcite-date-picker>

<div style="width: 400px">
  <calcite-label layout="inline">
    Date
    <calcite-input-date-picker
      min="2016-08-09"
      max="2023-12-18"
      locale="en"
      intl-next-month="Next month"
      intl-prev-month="Previous month"
      role="application"
      layout="horizontal"
    ></calcite-input-date-picker>
  </calcite-label>
</div>
```

### Expected Behavior
The .year input has an associated label that describes its purpose.

[Recommendation]
Developer should provide "aria-label" such as aria-label="year" in the source code of the input field.

[Compliant Code Example]
`<input aria-label="year" class="year" inputmode="numeric" maxlength="4" minlength="1" pattern="\d*" type="text">`

### Remediation Steps
n/a

### Screenshots/Images
![image](https://user-images.githubusercontent.com/42423180/140232296-8099b180-18da-48c3-b0cd-de37f969fcbc.png)

### Additional Context
This component is used by Calcite Date Picker and Calcite Input Date Picker.

### Issue Source

- Instance: https://amp.levelaccess.net/public/reporting/view_instance.php?instance_id=1899531371
- Instance: https://amp.levelaccess.net/public/reporting/view_instance.php?instance_id=1899531372

benelan:
verified on `beta.74`

PR #3707, opened 2021-12-14
fix(date-picker-month-header): add a corresponding aria-label to the year input

Elijbet:
**Related Issue:** #3425

## Summary

The .year input in Calcite Date Month Header now has an associated aria-label.

## Screenshots/Images

aria-label year


src/components/calcite-date-picker-month-header/calcite-date-picker-month-header.tsx
```diff
@@ -134,6 +134,7 @@ export class CalciteDatePickerMonthHeader {
           </CalciteHeading>
           <span class="year-wrap">
             <input
+              aria-label="year"
```
driskull, line 137: @Elijbet this needs a `intlYear` prop so that a user can localize and set this value.

Elijbet: Ok, I'll take care of that.

driskull (commented):
I think we need to set the intlYear on the date-picker and input-date-picker as well so that the value can be assigned by a user. See the following code...

https://github.com/Esri/calcite-components/blob/c21e31e4a8d459841947f79deb3d395ac48ce662/src/components/calcite-date-picker/calcite-date-picker.tsx#L113

driskull (approved):
🎉 🌮 👍

Issue #3574, opened 2021-11-26
When disabling Radio Button and Radio Button group can be selected


arxit-nis:
### Actual Behavior

When setting the disabled property to a Radio Button or Radio Button group, the component can be still selected.

### Expected Behavior

When the property 'disabled' is set to the component, it should not be selectable.

### Reproduction Steps and Sample

_Sample_: https://developers.arcgis.com/calcite-design-system/components/radio-button/ 

1. From the properties, uncheck the 'check' property
2. Check the 'disabled' property
![image](https://user-images.githubusercontent.com/95091581/143601390-6f3406c2-ef3e-4c2c-a15c-7f1f5fbc654a.png)
3. Let's see that you can select the Radio button 

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-69`

- [ ] CDN
- [ ] NPM package

Regression? Known working version_: `@esri/calcite-components@1.0.0-52`

benelan:
In version `beta-70` it is not selectable, however we still need to remove the focus when clicking a disabled radio button, eg
https://codepen.io/benesri/pen/vYeORvJ

jcfranco:
I this can be fixed by making the following changes to https://github.com/Esri/calcite-components/blob/master/src/utils/label.ts:

1. Adding a `disabled` prop to the `LabelableComponent` interface
2. Update `onLabelClick` to return early if the labelable component is disabled

benelan:
verified on `beta.74`

PR #3718, opened 2021-12-14
fix(label): Prevents selecting disabled labeled element. #3574

driskull:
**Related Issue:** #3574

## Summary

fix(label): Prevents selecting disabled labeled element. #3574


benelan (approved):
Can you add an e2e test for this some where? Otherwise looks good!

src/components/calcite-input/calcite-input.tsx
```diff
@@ -283,7 +283,9 @@ export class CalciteInput implements LabelableComponent, FormComponent {
     this.scale = getElementProp(this.el, "scale", this.scale);
     this.status = getElementProp(this.el, "status", this.status);
     this.inlineEditableEl = this.el.closest("calcite-inline-editable");
-    this.editingEnabled = this.inlineEditableEl?.editingEnabled;
```
driskull, line 286: cc @eriklharper this was messing up tests. It was setting editingEnabled to undefined always which was undoing the default false value. Lets be careful with connectedCallback stuff.

Issue #2538, opened 2021-07-14
Bug: Number input event not fired when removing remaining decimal value


mpayson:
### Actual Behavior

When I type "1.2" into a number input then delete "2", the component only emits events for "1" and "1.2" not "1", but the final input displays "1".

### Expected Behavior
The component emits an event for the final "1" as well.

### Reproduction Steps or Sample

Codepen: https://codepen.io/mpayson/pen/BaRpXvy

Type "1.2" into the input then delete "2" and check the console logs, there should only be two events
### Relevant Info

Firefox, latest (89)

_Version_

- [x] CDN (1.0.0-beta.59) 
- [x] NPM package (1.0.0-beta.57)

benelan:
Relevant code
https://github.com/Esri/calcite-components/blob/14a8f75f99750e0c0f6d9ca7d39a3b641d46757e/src/components/calcite-input/calcite-input.tsx#L613-L621

jcfranco:
Verified with repro case!

PR #3719, opened 2021-12-14
fix(input): input event fires when number ends with a decimal

benelan:
**Related Issue:** #2538

## Summary
- `calciteInputInput` wasn't firing when the last character was a decimal.

Edit:
With these changes the functionality will be

Type "1" -> event fires w/ "1"
Type "." -> event fires w/ "1"
Type "2" -> event fires w/ "1.2"
Delete "2" -> event fires w/ "1"


driskull (approved):
Nice catch 👍 ✨ 🎆

Issue #3694, opened 2021-12-13
Bug: browser-specific switch focus border regressions


benelan:
### Actual Behavior

~~There is a black border around a calcite-switch when clicking on the text of a parent calcite-label.~~
> Anvesh fixed the black border on Chrome, thanks!

Even with the black border fix, `calcite-switch`'s focus border still has regressions. It looks different in FF and Chrome, but both of them are not correct. 

-  In Chrome there is no focus border when clicking on the switch itself. When clicking on the label the border is rounded.
- In FF there is a dotted border inside of the expected border when clicking on the label. Clicking on the switch works as expected in FF.
- I don't have Safari but it should be tested as well.

The correct focus border is the one in `beta.69`, which works on both browsers.

### Expected Behavior
The focus border should look like it does in `beta.69` on all browsers.

### Reproduction Steps and Sample

_Sample_: https://codepen.io/benesri/pen/ZEXLYqx?editors=1100

1. In all the browsers, use the switch (clicking both on the label and on the switch itself).
2. Repeat in `beta.69`, which is the expected behavior in both browsers

### Relevant Info
[Screenshots](#issuecomment-996211600)

_Reproducible version_: `@esri/calcite-components@1.0.0-beta.71`

- [x] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-beta.69`

macandcheese:
Yeah it looks like the focus outline is no longer being applied - previously:
<img width="347" alt="Screen Shot 2021-12-13 at 12 48 56 PM" src="https://user-images.githubusercontent.com/4733155/145886600-b467cacb-0240-4623-951c-46b9513cfb70.png">

benelan:
Yup the focus outline is present in `beta.69`. I didn't know if removing that was by design but I assumed that ugly black border was not.

anveshmekala:
@benelan I don't see a black border while toggling the switch with label. Visual difference i see between `beta.69` and `.71` is the shrinked `focus-outline`.

benelan:
you don't see this?
![image](https://user-images.githubusercontent.com/10986395/146091518-5886b073-1f01-47a2-84f6-f468c888325a.png)

anveshmekala:
nope. i am using chrome. The black border is visible in edge though.

benelan:
That screenshot is from Chrome, looks like it's not as obvious in FF

benelan:
[...]

PR #3721, opened 2021-12-15
fix(switch): switch no longer has black outline when toggled with label

anveshmekala:
**Related Issue:** #3694 

## Summary

Fixes `outline` of `calcite-switch` when toggled  with `calcite-label`  which currently is set to `black` by Chrome and edge browsers.

### DIFF
--- a/src/components/calcite-label/calcite-label.stories.ts
+++ b/src/components/calcite-label/calcite-label.stories.ts
@@ -1,6 +1,7 @@
 import { themesDarkDefault } from "../../../.storybook/utils";
 import readme from "./readme.md";
 import { html } from "../../tests/utils";
+import { createSteps } from "../../../.storybook/helpers";
 
 export default {
   title: "Components/Label",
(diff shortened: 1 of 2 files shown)


driskull (commented):
Can we add an e2e test for this?

eriklharper:
> Can we add an e2e test for this?

This would be even better as a Screener snapshot, as long as this is triggerable using `setFocus`.

anveshmekala:
> > Can we add an e2e test for this?
> 
> This would be even better as a Screener snapshot, as long as this is triggerable using `setFocus`.

The `outline` turns black only when you toggle the switch with label. If we can replicate this in screener that would be ideal.

eriklharper:
Cool, so yeah, you should be able to click the switch with a `.click()` step with our screener utils documented here:  https://github.com/Esri/calcite-components/wiki/Screenshot-Testing  Let me know if you have any questions or run into any issues!

eriklharper (approved):
new snapshots look awesome!

Issue #3053, opened 2021-09-16
Bug: calcite-input type:number allows multiple dashes


AdelheidF:
https://developers.arcgis.com/calcite-design-system/components/input/
![image](https://user-images.githubusercontent.com/44481292/133652330-df000f96-f1e0-49d2-8fee-ab9a0e8a170c.png)

or 5-5-5-5

### Reproduction Steps and Sample

_Sample_:

1.

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

benelan:
As mentioned, this issue is reproducible in the doc so the request for additional information is not applicable, thanks!

AdelheidF:
Could we please give this issue a higher priority? We get bug reports logged about it.

jcfranco:
Verified with https://codepen.io/jcfranco/pen/BawrLPz. Input is not restricted, but the value is properly parsed.

AdelheidF:
I don't think it's fixed if it still allows users to enter values like '5-----'

benelan:
We decided to not strict the input, but rather parse the value correctly. So even though it displays `5-----` , the value is `5`. Typing another five will remove the invalid dashes and it becomes `55`.

AdelheidF:
Ok, so you do automatically undo user's characters if the user enters the second 5. Why not do it right away?

benelan:
The native number input allows you to type the two hyphens and keeps them there. This is because `-5e-5` is a valid number. However with the native input you can also type `5-55-5`. And it doesn't parse out invalid hyphens. Would you restricting the number of dashes to two like native does be sufficient?

eriklharper:
The native number input allows you to type up to 2 hyphens in the input so as to accommodate values like `-5e-5`.  We can't restrict the user from typing a hyphen at the end of the value because it would prevent them from being able to enter a value like the one above.  We can set this up so that it restricts the input to a maximum of 2 hyphens.

AdelheidF:
Idea from Paul B. : it should fix itself when it looses focus

So we kind of expect users to type something like this? Mhh...
-
--
-5-
-5-5
-5e-5

Yeah I don't know what the best behavior would be. We get a bunch of issues from people being able to enter values that are not allowed or just weird.

AdelheidF:
Maybe it would help having this one fixed too: 
https://github.com/Esri/calcite-components/issues/3003 (when I programmatically update values in the input they don't show)

[...]

PR #3726, opened 2021-12-16
fix(input): sanitize extra dashes and leading zeros from number input value

benelan:
**Related Issue:** #3053

## Summary
- Sanitize leading zeros, e.g. `000000` and extra dashes, e.g. `-22----4-` which were being emitted and used for the value.
    - The zero example is now `0`, however it will still display all of the zeros in the input until a different number is entered. We are trying not to restrict the ability to type stuff (cc @eriklharper)
    - The extra dashes example is now `-224`. Same idea, you can type them but they are sanitized out of the value. Then typing another number removes the extra dashes from the displayed number as well.
- Removed a skipped test which is no longer valid after these changes
- Added a test for the zeros and a test for the dashes

### DIFF
--- a/src/utils/number.ts
+++ b/src/utils/number.ts
@@ -29,8 +29,18 @@ export function sanitizeDecimalString(decimalString: string): string {
   return decimalString?.endsWith(".") ? decimalString.replace(".", "") : decimalString;
 }
 
+export function sanitizeNegativeString(negativeString: string): string {
+  return negativeString.charAt(0) + negativeString.substring(1).replace(/-/g, "");
+}
+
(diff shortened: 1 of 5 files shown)


src/utils/number.ts
```diff
@@ -29,8 +29,18 @@ export function sanitizeDecimalString(decimalString: string): string {
   return decimalString?.endsWith(".") ? decimalString.replace(".", "") : decimalString;
 }
 
+export function sanitizeNegativeString(negativeString: string): string {
```
driskull, line 32: maybe a `utils.spec.ts` test for these number utils?

driskull (approved):
👍

Issue #3684, opened 2021-12-10
Remove header padding in calcite-block when there is no heading and it's open


jmanke:
### Description
Currently, calcite-block has extra padding in the header when heading is unset and the calcite-block is open `<calcite-block open>`. We have a requirement for some calcite-blocks to be configured like this in our setup.

### Acceptance Criteria 
When a calcite-block is configured without a heading and is open, remove the header padding.

### Relevant Info 
![Screen Shot 2021-12-10 at 10 47 19 AM](https://user-images.githubusercontent.com/19629809/145625902-fe600425-73a1-4a1f-918a-3e56ed73d513.png)

#### Which Component
calcite-block

#### Example Use Case
In our setup we have calcite-blocks containing configuration settings for widgets. For many widget configurations, the top-most calcite-block contains default configurations which doesn't need a heading, nor does it need to be collapsible.

jcfranco:
I'll let @asangma chime in on this, but the header is required by design. See https://github.com/Esri/calcite-components/issues/2824#issuecomment-905065990. cc @benelan

macandcheese:
@jmanke - could you share a few screenshots of the larger workflow - it could help us understand the full context and need (feel free to ping on Teams if you don't want to share publicly here) - thanks!

PR #3728, opened 2021-12-16
fix(Block): only render title node when heading or summary is used

asangma:
**Related Issue:** #3684

## Summary
Only render `.title` node is **heading** or **summary** is used.
Also adds some styles for some edge cases, e.g. for 
- collapsible but without heading and summary
- icon without heading and summary


driskull (approved):
👍

PR #3729, opened 2021-12-16
ci(release): don't update cdn versions in readme on next releases

benelan:
**Related Issue:** NA

## Summary
Right now the README is updated with `next` versions. From a product standpoint, the `next` versions should not be the recommended version in our project README. The people who need to use `next` will know how to do so, but the general public should use the `beta` versions.


driskull (approved):
👍

PR #3731, opened 2021-12-16
ci(release): ensure tags are pushed when releasing

benelan:
**Related Issue:** NA

## Summary
The last couple releases, the new tag was not pushed with the rest of the changes (package.json/readme/changelog). Adding `--atomic` to ensure the tag pushes, and if it doesn't it will give me a chance to troubleshoot.


driskull (approved):
👍

Issue #3117, opened 2021-09-27
Bug: calcite-tree selection-mode of multi


joeyHarig:
### Actual Behavior
When `selection-mode` for `calcite-tree` is set to `multi` or `multi-children`, the user is still unable to select multiple `calcite-tree-items`.

### Expected Behavior
A `selection-mode` of `multi` or `multi-children` should allow the user to select multiple `calcite-tree-items`. The event `calciteTreeSelect` should also return an array of all selected items, not just the most recently selected item. 

### Reproduction Steps and Sample
Here is a code pen that uses a `multi` select `calcite-tree`. Notice that only one item can be selected and that the event only returns an array of 1 selected items (this is being console logged). 
https://codepen.io/joeyHarig/pen/BaZqWRM?editors=1111

### Relevant Info

_Version_: `@esri/calcite-components@1.0.0-beta.65`

caripizza:
@joeyHarig as far as I know, the "shift" key is required when selecting multiple tree-items with either "multi" or "multi-children" selection-modes. (I made a run-down of the behavior I noticed [here](https://github.com/Esri/calcite-components/pull/2803#issue-969712281) for reference.)

caripizza:
I had the same thought myself @joeyHarig. AFAIK, tree's multi-select behavior was added via these specs (which mention the Shift key to select multiple list items design pattern):
-  https://www.w3.org/TR/wai-aria-practices-1.1/#:~:text=Selection%20in%20multi-select%20trees
- https://www.w3.org/TR/wai-aria-practices-1.1/#:~:text=Alternative%20selection%20model%20--%20Moving%20focus%20without%20holding%20the%20Shift%20or%20Control%20modifier%20unselects

That said, myself and @macandcheese agree -  it seems like at the very least, selection-mode "multi" shouldn't require the Shift key. It should allow you to click to select multiple items (but also support the shift key selection when used).

I'll keep this issue open so we can gather feedback from the team on potentially changing the Shift key requirement.

macandcheese:
Agree with above - having the shift key there seems a little unexpected, even if it's technically within spec. We should allow click / key interaction to select multiple. At which point I'm not sure there would be any support needed for the shift key at all?

macandcheese:
Here are some examples of multi-select trees that allow users to select with mouse / keyboard (not shift):

SAP: https://sap.github.io/ui5-webcomponents/playground/components/Tree/
[...]

PR #3733, opened 2021-12-17
feat(tree):  multi-selection no longer requires holding shift key

anveshmekala:
**Related Issue:** #3117 

## Summary

This will enable to select multiple `calcite-tree-items` when mode is set to `multi | multi-children` without holding `shift` key.  Additionally, `calciteTreeSelect` will emit the array of selected items instead of one.


src/components/calcite-tree/calcite-tree.tsx
```diff
@@ -188,34 +188,42 @@ export class CalciteTree {
...
         window.getSelection().removeAllRanges();
       }
-
       if (
         (shouldModifyToCurrentSelection && target.selected) ||
         (shouldSelectChildren && e.detail.forceToggle)
       ) {
-        targetItems.forEach((treeItem) => {
-          treeItem.selected = false;
+        this.selectedItems.forEach((treeItem, i) => {
```
jcfranco, line 208: Can you simplify this to

```ts
if (this.selectedItems.includes(target)) {
  // deselect `target` and update selected items
}
```
?

This applies to the following `else` statement.

jcfranco (changes requested):
This is looking good, @anveshmekala!

`calcite-tree` has a few selection modes, so it'd be good to get some additional manual testing to drive new E2E tests since the current E2E test suite doesn't cover all of them. @benelan Can you help with this?

jcfranco:
@anveshmekala Should this use the `feat` type? I believe multi selection was possible before with the shift key. Also, can you update the PR title to convey this (multi-selection no longer requires holding down the shift key).

src/components/calcite-tree/calcite-tree.tsx
```diff
@@ -225,6 +233,19 @@ export class CalciteTree {
     });
   }
 
+  deselectSiblings(target: HTMLCalciteTreeItemElement): void {
```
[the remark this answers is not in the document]
anveshmekala: how about `selectedTreeItem` ?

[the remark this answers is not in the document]
jcfranco: 👍🏼

src/components/calcite-tree-item/calcite-tree-item.tsx
```diff
@@ -418,4 +418,11 @@ export class CalciteTreeItem {
       return;
     }
   };
+
+  private isSelectionMulti(): boolean {
```
anveshmekala, line 422: Looking for a better name for this method. We can use `isSelectionMultiOrMultiChildren` which is a bit long .Any suggestions?

jcfranco: `isSelectionMultiLike`? Your suggestion also works since it's descriptive enough. It's private too, so we can refactor if necessary.

benelan:
Tested the branch locally, LGTM! One thing I noticed is that `calciteTreeSelect` emits when opening/closing child trees, even though nothing was selected. This happens on `master` so unrelated to this PR, but is that expected behavior?

cc @jcfranco

jcfranco:
> This happens on master so unrelated to this PR, but is that expected behavior?

I don't _think_ so. That event should be exclusively tied to selection which is different from the visual state of the tree. [The doc](https://github.com/Esri/calcite-components/blob/7469960b57a13d4df1d5a453b4d42b6fd92216d2/src/components/calcite-tree/calcite-tree.tsx#L277-L280) seems to indicate this as well.

src/components/calcite-tree-item/calcite-tree-item.tsx
```diff
@@ -429,4 +429,11 @@ export class CalciteTreeItem implements ConditionalSlotComponent {
...
+    return (
+      this.selectionMode === TreeSelectionMode.Multi ||
```
jcfranco, line 435: To simplify, can you introduce an internal `isSelectionMulti` variable that gets computed whenever `selectionMode` is updated? The latter is an internal prop already, so you should be able to watch for changes.

src/components/calcite-tree/calcite-tree.e2e.ts
```diff
@@ -254,7 +254,6 @@ describe("calcite-tree", () => {
         await item1.click();
 
         expect(await getSelectedIds()).toEqual(["1"]);
```
jcfranco, line 256: To improve our current coverage, can you add a test case for the `multi-children` selection mode?

jcfranco (approved):
Aside from a couple of comments, this LGTM! ✨🌳✨

Issue #2823, opened 2021-08-16
Bug: Panel heading and summary should be allowed to wrap


asangma:
### Actual Behavior
The update [here](https://github.com/Esri/calcite-components/commit/8085fbe207682d9d0a3a66442f78cd9b3cb6a5f7#diff-b21f80b9fd202502f92995d56073caa7d05c12dd8c9ef8a3622115ea048e8d50) was a regression.
Long strings get truncated.

### Expected Behavior
Panel heading and summary should be allowed to wrap.

### Reproduction Steps or Sample

1. View a Panel with a long heading or summary string.

### Relevant Info
Panel headings and summaries in our interfaces are potentially long and cutting off the string can cause problems for end users.

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

asangma:
This example shows a regular case where multiple layers that all have similar names, and the distinctive part of those names are at the end of the string.

* Popular Demographics in the United States - Country
* Popular Demographics in the United States - State
* Popular Demographics in the United States - County
* Popular Demographics in the United States - Zip Code
* Popular Demographics in the United States - Tract
* Popular Demographics in the United States - Block Group

![image](https://user-images.githubusercontent.com/12503298/129595996-8fd4444f-d789-47ff-9728-7cf2d952a44c.png)

asangma:
cc @mitc7862 Could use your input on this.

As I recall, truncating strings was an issue, but it have not been as relevant in Panel heading.
Seems at least relevant in summary.

macandcheese:
@asangma Is this still an issue? It might be good to tackle truncation across all components - some options might be to reveal the full title on hover in a native label / tooltip, or truncate the middle portion to reveal end, etc.

asangma:
Still an issue.
I think that revealing on hover would be a bit taxing especially considering cases where there are a bunch of potential strings that are identical except for the last part of the string, e.g. a Demographics layer.

I like the middle truncation idea, but it can run into similar issues.

I know that multi-line headings and summaries aren't pretty, but it's really useful to be able to read the full names of things like layers or data fields. Put another way, hiding any portion of those strings can hinder users.

macandcheese:
Agreed for the scope of this issue - revealing all is the safest. Apps can truncate strings themselves if they need for now.

y0n4:
[reproduced](https://codepen.io/yonacorn/pen/wvrdELB)

asangma:
Noice!

PR #3736, opened 2021-12-17
fix(panel): remove truncating for summary and heading prop

y0n4:
**Related Issue:** #2823 

## Summary
Changed any truncated titles/summaries to word wrap around. For now it was in component panel. I wasn't able to find other components that had any text that gets truncated. 

Before
Screen Shot 2021-12-17 at 1 56 18 PM

After
Screen Shot 2021-12-17 at 1 55 59 PM


src/components/calcite-panel/calcite-panel.scss
```diff
@@ -96,10 +96,8 @@
   .summary {
     @apply p-0
       block
-      overflow-hidden
-      whitespace-no-wrap
-      w-full;
-    text-overflow: ellipsis;
+      pointer-events-none
```
driskull, line 99: why no pointer events here?

driskull (commented):
Just a comment.

src/components/calcite-panel/calcite-panel.scss
y0n4: oh! You're right. Thanks for catching that

driskull (approved):
👍

Issue #3185, opened 2021-10-08
Enhancement: allow exponential numbers for input type="number"


benelan:
### Description
We should allow exponential numbers to be entered in `calcite-input` type="number". For example: `2e25` or `5e-4`

### Acceptance Criteria 
- Exponential numbers can be entered into the number input
- The exponential number string is converted into an actual number, which is used for the value

#### Which Component
`calcite-input`
#### Example Use Case
Inputting numbers with a lot of zeros gives me a headache

cc @jcfranco @eriklharper

jcfranco:
Verified via https://codepen.io/jcfranco/pen/zYELrbd?editors=1000

PR #3738, opened 2021-12-18
feat(input): add support for exponential number format

benelan:
**Related Issue:** #3185

## Summary
Adds support for exponential number format. If someone can checkout this branch and see if it works as you would expect that would be great!

- Only allowed to type one `e`
- Sanitizes on typing/pasting by splitting valid exponential numbers in half, validating each side with the preexisting functions, and then joins the two sides back together.
- Stored as an exponential format string. Since the `value` prop is a string, even for the number type, providing the exponential format allows the user to decide how they want to display/use it. They can easily call `Number()` to convert it to long form, e.g.  `Number(1e2) === 100`
- Nudging works, and converts the exponential string into long form.
- Handled localization the same way as sanitization (splitting). @eriklharper do you think that splitting workflow is sufficient or do I need to look into how other languages handle exponential format?

Also moved some other e2e I made this week to the correct `describe` scope.


benelan:
Based on @macandcheese's suggestions, I made a couple changes to match the way the native input sanitizes exponential strings:
- When pasting, keep the first `e` and remove the rest
- Do not allow decimals after the `e`

eriklharper (approved):
👍 lookin awesome

Issue #3610, opened 2021-12-01
Extra left padding in dropdown item in selection-mode:none


ellenupp:
### Actual Behavior

There is extra left padding in dropdown items in a group with selection-mode='none'. 

![image](https://user-images.githubusercontent.com/14304581/144287513-726e0899-5a59-42e7-83f9-d0d5b37cd226.png)

### Expected Behavior

In none selection mode, the dropdown items should be left-aligned with the header as per this issue: https://github.com/Esri/calcite-components/issues/1921#issuecomment-833633878

![image](https://user-images.githubusercontent.com/14304581/144287622-95585505-01b2-45af-994a-1cbd1d4ff58e.png)

### Reproduction Steps and Sample

_Sample_: https://codepen.io/ellupp/pen/OJxPaPx?editors=1000

1. See sample, applies to links and non-links

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-beta.70`

- [x] CDN
- [x] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-beta.68` (related issue #3349)

ellenupp:
Also I noticed there are CSS variables for padding, but they don't seem to be used. So unfortunately can't use that as a quick fix.
![image](https://user-images.githubusercontent.com/14304581/144332381-14e27081-1ce1-42d5-b47f-952b44c3f638.png)

macandcheese:
Yes this should be updated to the previous functionality. @benelan we should probably also remove those css variables since it seems like a component-specific variable here.

benelan:
@Elijbet can you take another look at this one, on `next` the "Link" item still has the extra padding. 
![image](https://user-images.githubusercontent.com/10986395/146098288-9317fa57-cde3-49ee-8b3d-140f15768cd6.png)

This is the correct padding in `1.0.0-beta.68` before the regression:
![image](https://user-images.githubusercontent.com/10986395/146098225-0aa2dda7-ecf5-47ec-a2d1-0674e316ab35.png)

benelan:
verified on `beta.74`

PR #3742, opened 2021-12-20
fix(dropdown-item): align items with heading in selection-mode: none

Elijbet:
**Related Issue:** #3610

## Summary

In none-selection mode, the dropdown items (links and non links) are now aligned with the header using logical properties.


driskull (approved):
Looks good @Elijbet. Does screener catch this or do we need a test for it?

PR #3746, opened 2021-12-21
refactor(3491): rename animation duration variables and revert base variab…

anveshmekala:
**Related Issue:** #3491 

## Summary

- This will rename the `--calcite-animation-timing-factor` to `--calcite-duration-factor` be more concise and avoids ambiguity of whether it is used for animations only or transitions included.  
- Additionally, reverted the change introduced in #3492 where the base global animation timing variable `--calcite-animation-timing` is moved to  `@internal`. With this PR, `--calcite-animation-timing` will be a global variable
- Rename the  internal animation timing scale `quick` introduced in #3492 to `fast`

### DIFF
--- a/src/assets/styles/_animation.scss
+++ b/src/assets/styles/_animation.scss
@@ -41,9 +41,9 @@
 }
 
 :root {
-  --calcite-internal-duration-factor: var(--calcite-duration-factor, 1);
   --calcite-animation-timing: calc(150ms * var(--calcite-internal-duration-factor));
-  --calcite-internal-animation-timing-quick: calc(100ms * var(--calcite-internal-duration-factor));
+  --calcite-internal-duration-factor: var(--calcite-duration-factor, 1);
+  --calcite-internal-animation-timing-fast: calc(100ms * var(--calcite-internal-duration-factor));
   --calcite-internal-animation-timing-medium: calc(200ms * var(--calcite-internal-duration-factor));
   --calcite-internal-animation-timing-slow: calc(300ms * var(--calcite-internal-duration-factor));
 }
--- a/src/assets/styles/_animation.scss
+++ b/src/assets/styles/_animation.scss
@@ -41,18 +41,18 @@
 }
 
 :root {
-  --calcite-internal-animation-timing-factor: var(--calcite-animation-timing-factor, 1);
-  --calcite-internal-animation-timing: calc(150ms * var(--calcite-internal-animation-timing-factor));
-  --calcite-internal-animation-timing-quick: calc(100ms * var(--calcite-internal-animation-timing-factor));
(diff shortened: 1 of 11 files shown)


src/tests/globalStyles.e2e.ts
```diff
@@ -38,17 +38,17 @@ describe("global styles", () => {
...
 
-    it("should set animation duration to 0ms when --animation-timing-factor set to zero", async () => {
+    it("should set animation duration to 0ms when --calcite-duration-factor set to zero", async () => {
       const page = await newE2EPage({
         html: html`
           <html>
             <style>
               html {
-                --calcite-animation-timing-factor: 0;
+                --calcite-duration: 0;
```
jcfranco, line 47: This should be `--calcite-duration-factor`, right? This also applies to the next test.

src/assets/styles/_animation.scss
```diff
@@ -41,18 +41,18 @@
 }
 
 :root {
-  --calcite-internal-animation-timing-factor: var(--calcite-animation-timing-factor, 1);
-  --calcite-internal-animation-timing: calc(150ms * var(--calcite-internal-animation-timing-factor));
-  --calcite-internal-animation-timing-quick: calc(100ms * var(--calcite-internal-animation-timing-factor));
-  --calcite-internal-animation-timing-medium: calc(200ms * var(--calcite-internal-animation-timing-factor));
-  --calcite-internal-animation-timing-slow: calc(300ms * var(--calcite-internal-animation-timing-factor));
+  --calcite-internal-duration-factor: var(--calcite-duration-factor, 1);
+  --calcite-animation-timing: calc(150ms * var(--calcite-internal-duration-factor));
```
jcfranco, line 45: Nitpick: can you move this one to be first so it's easier to distinguish from the internal ones?

jcfranco (approved):
Once the test comments are fixed, this is good to merge.

src/assets/styles/_animation.scss
```diff
@@ -41,18 +41,18 @@
 }
 
 :root {
-  --calcite-internal-animation-timing-factor: var(--calcite-animation-timing-factor, 1);
-  --calcite-internal-animation-timing: calc(150ms * var(--calcite-internal-animation-timing-factor));
-  --calcite-internal-animation-timing-quick: calc(100ms * var(--calcite-internal-animation-timing-factor));
```
macandcheese, line 46: Can we use 'fast' here? Nitpick but we already use that terminology on `alert`.

src/tests/globalStyles.e2e.ts
anveshmekala: Yup.

src/assets/styles/_animation.scss
anveshmekala: change the name to `fast` . Good catch @macandcheese

PR #3765, opened 2021-12-21
build: use postcss to ensure logical prop usage

jcfranco:
**Related Issue:** N/A

## Summary

Resubmitting this because [earlier install](https://github.com/Esri/calcite-components/pull/3688) introduced visual regressions and was reverted.

---

This wires up postcss + stylelint + https://www.npmjs.com/package/stylelint-use-logical-spec to ensure the CSS output uses [logical props](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties). 

Note that this is using the default configuration and will transform all properties that have a logical equivalent. We can tweak this if we need to prevent some properties from being transformed.


driskull:
Should we close this stale PR?

jcfranco:
@driskull I updated this branch and tests are now passing. Right now, we'll have to set `direction` for cases where we want left/right to mean just that. If we want to keep top/left/right/bottom as they are to keep this simple, I can add an exception for these props.

driskull (approved):
This looks good!

Issue #3675, opened 2021-12-09
calcite-link has no default display for it's host


driskull:
### Actual Behavior

calcite-link has no default display for it's host

### Expected Behavior

calcite-link should be displayed as inline-block or inline-flex?

### Reproduction Steps and Sample

_Sample_:

1. https://codepen.io/driskull/pen/LYzZooe?editors=1010

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-<version>`

- [x] CDN
- [x] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-<version>`

driskull:
@benelan @jcfranco plz triage

macandcheese:
I believe this needs to be set as `inline` to keep it displaying as it does now - because of how the link may wrap with longer strings, see example:
https://jsbin.com/tapecupitu/1/edit?html,js,output

<img width="325" alt="Screen Shot 2021-12-15 at 3 10 04 PM" src="https://user-images.githubusercontent.com/4733155/146279119-b190a3a9-b0a5-4ed3-9a7e-acdc35cb1623.png">

driskull:
Makes sense, maybe we should set it to inline in the host and add a test as well?

It also looks weird with RTL since it doesn't position the icons correctly:

https://jsbin.com/tuzupuqere/1/edit?html,js,output

benelan:
verified on `beta.74`

PR #3768, opened 2021-12-22
fix(link): Set host display to inline. #3675

driskull:
**Related Issue:** #3675

## Summary

fix(link): Set host display to inline. #3675


jcfranco (approved):
💥

Issue #3200, opened 2021-10-11
Enhancement: support customizing calcite-icon's color via new global CSS variable


caripizza:
### Description
Refactor calcite-icon to utilize a new optional CSS Custom Property for its color (ex, `--calcite-ui-icon-color` or similar).

This change should allow for calcite-icon - when used inside a calcite-component - to derive its color value either from the new CSS var, or from its [`currentColor`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#currentcolor_keyword) property.

(Franco)
> case 1: calcite-icon uses --calcite-ui-text-1 (or a more fitting var) for the icon color 
> case 2: calcite-icon uses --calcite-ui-icon (new) for the icon color
> Both of these options will help set the icon color in cases where the icon component is internal to another component.

(Paul)
> I could see giving it a variable. IMO we should still set this var's value in the calcite-icon styles to be currentColor as that functionality is really convenient in many situations.

### Acceptance Criteria 
A new global CSS Var is provided and documented which allows for overriding the color of calcite-icon.

### Relevant Info 
(See internal Teams Core discussion from 10/7, no share link unfortunately.)

#### Which Component
calcite-icon, and all components that use calcite-icon

#### Example Use Case
```html
<style>calcite-chip { --calcite-ui-icon-color: #6d0b5d; }</style>
<calcite-chip value="Global" icon="globe" appearance="clear" color="green">Global</calcite-chip>
```
![image](https://user-images.githubusercontent.com/42423180/136862747-27d9cd95-7c57-4118-ae37-2fac337bb677.png)

caripizza:
(Added discussion label to this issue since I think it may benefit from a bit more fine-tuned AC before assignment.)

One question I have is - should we prevent changing the color of calcite-icon when it's used as a dismissible "x" icon, like in chip?

benelan:
verified on `beta.74` 🌈 
https://codepen.io/benesri/pen/gOGZRvV

PR #3770, opened 2021-12-22
feat(icon): Add new global CSS variable to customize icon's color. #3200

driskull:
**Related Issue:** #3200

## Summary

feat(icon): Add new global CSS variable to customize icon's color. #3200


benelan:
Is there a reason why this can't be a component-specific css var? We have a few of those floating around, eg https://developers.arcgis.com/calcite-design-system/components/alert/#component-api-styles

macandcheese:
If we did that we’d need to add it to each component that could render an icon internally via prop right?

jcfranco (approved):
Awesome! 🏆

Issue #3430, opened 2021-11-04
Icon: components with icons are not aligned with ones that do not


benelan:
### Actual Behavior
The Avatars in the documentation are not aligned properly

### Expected Behavior
All of the avatars are aligned properly, regardless of the content

### Reproduction Steps and Sample

_Sample_: https://developers.arcgis.com/calcite-design-system/components/avatar/#sample

1. Reproducible in the "Variety of Avatars" documentation sample

### Relevant Info
from holistic feedback

_Reproducible version_: `@esri/calcite-components@1.0.0-beta.68`

- [x] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-<version>`

macandcheese:
This occurs when using the `calcite-icon` component inside chip as well - probably independent of `avatar` : https://jsbin.com/miyowogisa/1/edit?html,output

driskull:
Seems like some flex alignment is needed here.

Elijbet:
@benelan Do you think you could put together a codepen for this outside of the documentation?

benelan:
Thanks for fixing the avatar issue! However, I am still able to repro this issue in [Adam's example on `next`](https://codepen.io/benesri/pen/GRMMeYo). Sorry for moving the goal post, but I'd like to continue using this issue in case there is a fix that works for all affected components (maybe in `calcite-icon`?). I edited the title and description of the issue.

Elijbet:
> Thanks for fixing the avatar issue! However, I am still able to repro this issue in [Adam's example on `next`](https://codepen.io/benesri/pen/GRMMeYo). Sorry for moving the goal post, but I'd like to continue using this issue in case there is a fix that works for all affected components (maybe in `calcite-icon`?). I edited the title and description of the issue.

Sorry I missed the other affected coponents. I'll get back to that in a bit.

Elijbet:
@benelan: I’ve discussed this with @asangma and he thinks it might not be something that can be fixed from within our components. Seems like that's just what DIV does with images and SVGs. It just likes to add a few pixels to the bottom. He thinks we're up against default browser behavior in this case.

He is suggesting to design and build a ChipGroup (or Chips or ChipList) component that helps the layout and spacing between the chips. Like we do for PickList and PickListItem, etc.

What do you think @macandcheese?

driskull:
If I add `vertical-align: top;` to the first chip it seems to fix it.

[...]

PR #3774, opened 2021-12-22
fix(avatar): avatars containing text vs icon vs image are now lined up when placed next to each other

Elijbet:
**Related Issue:** #3430

## Summary

Avatars containing text vs icon vs image are now line up when placed next to each other.


driskull (approved):
@Elijbet can you revert the changes to package-lock? Otherwise this looks good 🦖 👍

PR #3775, opened 2021-12-22
fix(input): set value to empty string instead of null or undefined

benelan:
**Related Issue:** #3627

## Summary
- The PR linked above moved to use an empty string for the value instead of `null`. I made the same change for pasting invalid numbers.

### DIFF
--- a/src/components/calcite-input/calcite-input.tsx
+++ b/src/components/calcite-input/calcite-input.tsx
@@ -288,7 +288,7 @@ export class CalciteInput implements LabelableComponent, FormComponent {
       if (isValidNumber(this.value)) {
         this.localizedValue = localizeNumberString(this.value, this.locale, this.groupSeparator);
       } else {
-        this.value = undefined;
+        this.value = "";
       }
     }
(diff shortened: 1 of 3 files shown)


benelan:
@jcfranco / @eriklharper, should this one be an empty string too?
https://github.com/Esri/calcite-components/blob/6a9756c4f3a0840c0c22195b2287aa487c8f1c1f/src/components/calcite-input/calcite-input.tsx#L291

eriklharper:
> @jcfranco / @eriklharper, should this one be an empty string too?
> 
> https://github.com/Esri/calcite-components/blob/6a9756c4f3a0840c0c22195b2287aa487c8f1c1f/src/components/calcite-input/calcite-input.tsx#L291

I would say yes, since the intention with this PR is to use "" as the empty value.  This is how the native input behaves.

eriklharper (approved):
LGTM once you change the line you mentioned from undefined to empty string.

jcfranco (commented):
Thanks for fixing this! :muscle:

Issue #3767, opened 2021-12-22
Bug: hyphen is removed when typing negative decimals in number input


benelan:
### Actual Behavior
You can't type negative decimal numbers, eg `-0.5`. The hyphen is removed when typing the `0`. If you type `0.5` then go back and add the hyphen it works fine. 

### Expected Behavior
Typing negative decimal values does not remove the hyphen

### Reproduction Steps and Sample

_Sample_: https://codepen.io/benesri/pen/bGorQYW

1. Type `-0.5` and notice the hyphen is removed after entering the zero
   - You have to type the hyphen twice due to #3766)
2. Switch to `beta.56` and it works fine

Also reproducible in the doc: https://developers.arcgis.com/calcite-design-system/components/input/#sample
### Relevant Info
Looks related to #2325

_Reproducible version_: `@esri/calcite-components@1.0.0-beta.72`

- [x] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-beta.56`

jcfranco:
Verified via repro case!

Issue #3766, opened 2021-12-22
Bug: the first hyphen entered into number input is removed


benelan:
### Actual Behavior
When typing `-` in number input, you have to type it twice in order for the hyphen to appear

### Expected Behavior
Only have to type a hyphen once

### Reproduction Steps and Sample

_Sample_: https://codepen.io/benesri/pen/bGorQYW

1. Type `-` and it won't add the character. Type it again and it will
2. Switch to `beta.70` and it works fine

Also reproducible in the doc: https://developers.arcgis.com/calcite-design-system/components/input/#sample
### Relevant Info
The only `input` PRs in the `beta.71` release notes are #3582 and #3627 

_Reproducible version_: `@esri/calcite-components@1.0.0-beta.72`

- [x] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-beta.70`

---
#### Edits
This is [the commit](https://github.com/Esri/calcite-components/commit/6ef5d4f338cdc29a4096f7f92efa48ebca8107cc#diff-d4e9e38d573f92a6608141863d1a3ce63b4cc9c7bc0b97585e8a91c48ac9a705L197-L244) that causes the regression

jcfranco:
Verified via repro case!

PR #3776, opened 2021-12-22
fix(input): allow typing negative decimal numbers

benelan:
**Related Issue:** #3767 and #3766

## Summary
- The sanitization function currently sanitizes the value, converts it to a number, and then back to string. I changed it to only use the sanitized value if it is a valid number. Otherwise use the non-sanitized value. Thanks @eriklharper for that suggestion.
- `Number("-0.000") === 0` which causes issues when trying to type numbers like `-0.0005`. So I added regex to make sure typing negative decimal values does not get converted `string->number->string` until there is a non-zero number.

Its worth considering not converting `string->number->string` at all once we have a validation story. However, it does clean stuff up like `Number("0.0") === 0`, which comes up with decimal steps.


src/utils/number.ts
```diff
@@ -26,20 +26,25 @@ export function parseNumberString(numberString?: string): string {
...
 }
 
 export function sanitizeNegativeString(negativeString: string): string {
-  return negativeString.charAt(0) + negativeString.substring(1).replace(/-/g, "");
+  return negativeString.replace(/(?!^-)-/g, "");
 }
 
 export function sanitizeLeadingZeroString(zeroString: string): string {
-  return zeroString.replace(/^0+$/, "0");
+  return zeroString.replace(/(?=^[-0])^([-0])0+(?=\d)/, `$1`);
```
jcfranco, line 37: Does the updated version also sanitize the non-negative case? Wondering if it needs to be renamed. Also, could the pattern be stored in a local var with an expressive name to convey what it will capture?

src/utils/number.ts
```diff
@@ -26,20 +26,25 @@ export function parseNumberString(numberString?: string): string {
 }
 
 export function sanitizeDecimalString(decimalString: string): string {
-  return decimalString?.endsWith(".") ? decimalString.replace(".", "") : decimalString;
+  return decimalString.replace(/\.$/, "");
```
jcfranco, line 29: Do you need to handle `decimalString` being `null`? Not sure if the previous version did as an extra precaution or an actual scenario. cc @eriklharper

src/utils/number.ts
```diff
@@ -26,20 +26,25 @@ export function parseNumberString(numberString?: string): string {
...
 }
 
 export function sanitizeNegativeString(negativeString: string): string {
-  return negativeString.charAt(0) + negativeString.substring(1).replace(/-/g, "");
+  return negativeString.replace(/(?!^-)-/g, "");
```
jcfranco, line 33: For readability, can the pattern be stored in an expressive local var?

jcfranco (approved):
Other than a few comments regarding maintainability, this LGTM! 🎉

src/utils/number.ts
benelan: No because of #3627

src/utils/number.ts
benelan: It does capture non-zero as well. `[-0]` is hyphen or zero. I will move all of these to a local var for maintainability

src/utils/number.ts
jcfranco: Noice. 👍🏼

src/utils/number.ts
jcfranco: Thanks for clarifying. I always need a double take on RegExps.

confused-marky

Issue #3427, opened 2021-11-04
Dropdown group is displayed as inline but should be a block or flex displayed element


driskull:
### Actual Behavior

Dropdown group is displayed as inline

### Expected Behavior

Dropdown group should be a block or flex displayed element

### Reproduction Steps and Sample

N/A

_Sample_:

1. https://codepen.io/driskull/pen/WNEMGgv

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-<version>`

- [x] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-<version>`

benelan:
reproducible in the doc sample: https://developers.arcgis.com/calcite-design-system/components/dropdown/#sample

benelan:
verified on `beta.74`

PR #3777, opened 2021-12-22
fix(dropdown-group): Set display as block. #3427

driskull:
**Related Issue:** #3427

## Summary

fix(dropdown-group): Set display as block. #3427


jcfranco (approved):
💥

PR #3778, opened 2021-12-22
refactor: Update tests to use common test and apply correct default display.

driskull:
**Related Issue:** N/A

## Summary

refactor: Update tests to use common test and apply correct default display.

- Uses `renders` common test util in place of one off renders tests.
- Adds default display values where necessary. These will need to be verified.


src/components/calcite-inline-editable/calcite-inline-editable.scss
```diff
@@ -1,3 +1,7 @@
+:host {
+  @apply inline-block;
```
driskull: @eriklharper what display should this component have? Should it be "block" like input?

src/components/calcite-label/calcite-label.scss
```diff
@@ -1,3 +1,7 @@
+:host {
+  @apply inline;
```
driskull, line 2: Native label is inline.

src/components/calcite-loader/calcite-loader.scss
```diff
@@ -45,6 +45,10 @@ $loader-circumference: ($loader-scale - (2 * $stroke-width)) * 3.14159;
   @apply py-0;
 }
 
+:host {
+  @apply hidden;
```
driskull, line 49: loader should be hidden by default and displayed when active.

src/components/calcite-split-button/calcite-split-button.scss
```diff
@@ -1,3 +1,7 @@
+:host {
+  @apply inline-block w-auto align-middle;
```
driskull, line 2: Copied this from `calcite-button`. is this correct @eriklharper?

src/components/calcite-input/calcite-input.scss
```diff
@@ -1,3 +1,7 @@
+:host {
+  @apply block;
```
driskull, line 2: @eriklharper what display should this component have? Should it be "block"? Seems like its already block now but I'm not sure why.

src/components/calcite-inline-editable/calcite-inline-editable.scss
```diff
@@ -1,3 +1,7 @@
+:host {
+  @apply block;
```
driskull, line 2: @eriklharper is this correct? seems input is block.

driskull (commented):
Questions

src/components/calcite-split-button/calcite-split-button.scss
jcfranco: Should the non-display styles be set on the container instead?

src/components/calcite-tab/calcite-tab.e2e.ts
```diff
@@ -1,17 +1,12 @@
...
-    const page = await newE2EPage({ html: tabHtml });
-    const element = await page.find("calcite-tab");
-    const styles = await element.getComputedStyle();
-    expect(element).toHaveAttribute(HYDRATED_ATTR);
-    expect(styles["display"]).toEqual("none");
-    expect(styles["visibility"]).toEqual("visible");
-  });
+  it("renders", () => renders("calcite-tab", { display: "none", visible: false }));
+
+  it("renders: active", () => renders("<calcite-tab active></calcite-tab>", { display: "block", visible: true }));
```
jcfranco, line 9: Let's try to group related tests whenever possible w/ `describe`.

src/components/calcite-inline-editable/calcite-inline-editable.e2e.ts
```diff
@@ -1,10 +1,20 @@
...
+import { accessible, labelable, renders } from "../../tests/commonTests";
 import { E2EPage } from "@stencil/core/testing";
 import { newE2EPage } from "@stencil/core/testing";
 import { CSS } from "./resources";
 import { html } from "../../tests/utils";
 
 describe("calcite-inline-editable", () => {
+  it("renders", () =>
+    renders(
+      `
```
jcfranco, line 10: Can you use the `html` helper on multi-line HTML for formatting?

src/components/calcite-loader/calcite-loader.scss
jcfranco: Sort of a separate discussion, but maybe we should use `hidden` for this? Looks like it's only used for visibility. cc @macandcheese

src/components/calcite-input/calcite-input.scss
jcfranco: cc @macandcheese

jcfranco (commented):
This is looking great. Thanks for setting this up.

src/components/calcite-split-button/calcite-split-button.scss
eriklharper: That looks right since the native `<button>` is also `inline-block` by default.

eriklharper: > Should the non-display styles be set on the container instead?

Good question.  I think the display should probably be set on the host element, but as far as the other properties, there's probably a lot you can move to a container div, unless the host needs like a position property or some other parent-child css property that enables styles to work for elements underneath the host element, like flex for example.

src/components/calcite-input/calcite-input.scss
eriklharper: `inline-block` is what the native input does, so that's probably the more appropriate one to use for this.

src/components/calcite-inline-editable/calcite-inline-editable.scss
eriklharper: I would advocate `inline-block` for this too.

src/components/calcite-loader/calcite-loader.scss
driskull: Its a separate discussion but by default the loader is not shown until it has an active attribute/property. We can evaluate changing this so that the active property is not necessary and just remove it.

src/components/calcite-split-button/calcite-split-button.scss
driskull: I'll remove the width and align stuff.

src/components/calcite-input/calcite-input.scss
driskull: If I change this to `inline-block` the screener tests will change due to width differences. I will leave as now. If you think it should be changed I think a different issue is needed.

src/components/calcite-inline-editable/calcite-inline-editable.scss
driskull: If I change this to `inline-block` the screener tests will change due to width differences. I will leave as now. If you think it should be changed I think a different issue is needed.

driskull:
@jcfranco can you review?

src/tests/commonTests.ts
```diff
@@ -60,7 +60,7 @@ export async function renders(
   componentTagOrHTML: TagOrHTML,
   options?: {
     visible?: boolean;
-    display?: string;
+    display: string;
```
jcfranco, line 63: 🏆

src/components/calcite-loader/calcite-loader.scss
jcfranco: https://github.com/Esri/calcite-components/issues/3821

jcfranco (approved):
LGTM!

@eriklharper @macandcheese I second @driskull's suggestion to open a new issue if any of the defaults need to change.

src/components/calcite-input/calcite-input.scss
driskull: I do agree that both of these should probably be inline-block like their native counterparts. However, I didn't want to break screener tests and user apps. So this would likely be a breaking change.

Issue #3783, opened 2021-12-23
Disabled checkboxes and radios should not be focusable


eriklharper:
### Actual Behavior

Disabled checkboxes and radio-buttons are focusable and display focus styling even when disabled.

### Expected Behavior

Neither checkboxes nor radio-buttons should be focusable or show visual indicators of focus state when disabled.

### Reproduction Steps and Sample

https://codepen.io/eriklharper/pen/jOGGRwp

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-next.343`

- [ ] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-<version>`

PR #3784, opened 2021-12-23
fix(checkbox,radio-button): prevents disabled checkboxes and radio-buttons from receiving keyboard or mouse focus

eriklharper:
**Related Issue:** #3783

## Summary

This pr fixes checkbox and radio-button so that they no longer allow focusing when disabled.  Testing for this change will be covered with Screener visual snapshots in this PR: https://github.com/Esri/calcite-components/pull/3571


jcfranco (approved):
Awesome! Thanks for splitting this up.

Before merging, can you:

1. Add a note in the description about testing being covered in the linked PR?
2. Rewrite the commit message before squash merging to convey what the commit fixes instead of what is not working?

Issue #3160, opened 2021-10-05
Bug: calcite-color-picker-swatch border does not respect class="calcite-theme-light"


jmanke:
### Actual Behavior

The calcite-color-picker-swatch's border is transparent when a parent element has class="calcite-theme-dark" and the calcite-color-picker-swatch has class="calcite-theme-light".

### Expected Behavior

The calcite-color-picker-swatch's border should be the visible when class="calcite-theme-light", regardless if there is a class="calcite-theme-dark" in a parent element. Otherwise, other calcite components will be light themed while the calcite-color-picker-swatch will have no border (colors close to white will just blend into the background).

### Reproduction Steps and Sample

1. Create a `<div class="calcite-theme-dark">`
2. Create a `<calcite-color-picker-swatch class="calcite-theme-light" color="#ffffff" >` as a child of the div created in step 1
3. Observe a transparent border for the calcite-color-picker-swatch

_Sample_:

1. [codepen sample](https://codepen.io/jmanke/pen/PojrLRB?editors=1010)

### Relevant Info

Browsers: Chrome, Firefox, Safari

_Version_: `@esri/calcite-components@<1.0.0-beta.66>`

- [x] CDN
- [x] NPM package

benelan:
For this one either double the alpha value or use the same as disabled for `borderLight`

https://github.com/Esri/calcite-components/blob/master/src/components/calcite-color-picker-swatch/resources.ts#L7-L8

jcfranco:
@Elijbet Looks like this got fixed a while back. Can you confirm if anything else is pending?

Elijbet:
> @Elijbet Looks like this got fixed a while back. Can you confirm if anything else is pending?

I believe it was finalized and merged.

jcfranco:
Thanks, updating labels!

jcfranco:
Never mind, you had already done so. 😅

benelan:
verified on beta.76

PR #3785, opened 2021-12-24
fix(color-picker-swatch, avatar): in case where theme is set on multiple levels, these components now take theme of the element or the closest parent

Elijbet:
**Related Issue:** #3160

## Summary
The closest() method traverses the element and its parents until it finds a node that matches the provided selector string. The getThemeName function in the utils to do this is set to use darkTheme as a selector. This means it will look for darkTheme up the parent tree even if the immediate parent is set to light and will set it to dark whenever it finds it. 

BorderColor in the swatch is set in the svg rect element based on the theme provided by this util function. Light theme provides a border.

Looks like initially this was written assuming there will only be one theme set. To account for cases where theme is set on multiple levels, the swatch, and other components using this util function will now take the theme of the closest parent, since now it looks for both dark and light themes and takes whichever is the closest.


driskull (approved):
Nice catch! I wonder if we should add a couple tests to `dom.e2e.ts`?

jcfranco (approved):
🎉🌮🦖

Minor thing regarding the commit message:

> take theme of the closest parent

The util should also return the theme if it's set on the element itself and does not necessarily need a parent.

Issue #1954, opened 2021-04-13
Bug: pick list retains a pick list item in its `selectedItems` when it is removed while selected


ethanbdev:
### Summary
When a pick list item is selected, then removed while selected, calling `getSelectedItems` after a new selection will return a map containing the **old, removed** selected item _and_ the new one.

### Reproduction Steps

1. Create a pick list with a few items
2. Select a pick list item
3. Remove the above selected item
4. Select a new item
5. Observe the result from both the `calciteListChange` event and `getSelectedItems()` method.

https://codepen.io/eborgen/pen/gOxJpey?editors=1011

benelan:
@ethanbdev Can you provide a codepen to repro this issue?

ethanbdev:
> @ethanbdev Can you provide a codepen to repro this issue?

Yeah, let me work on one.. Maybe I am missing something easy but how does the `getSelectedItems` method work in the codepen? https://codepen.io/eborgen/pen/gOxJpey?editors=1011

benelan:
`getSelectedItems()` returns a Promise that is resolved with a Map, so log something like `console.log(...selectedItems.entries());`

ethanbdev:
> `getSelectedItems()` returns a Promise that is resolved with a [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map), so log something like `console.log([...selectedItems.entries()]);`

Right 🤦 

Here you go on a repro, let me know if it makes sense:
https://codepen.io/eborgen/pen/gOxJpey?editors=1011

benelan:
Makes sense to me, thanks for the repro sample!

ethanbdev:
Looks good in my repro!

PR #3787, opened 2021-12-27
fix(pick-list, value-list): deselect values when it gets removed from the list (#1954)

y0n4:
**Related Issue:** #1954 

## Summary
Alerts and updates the `pick-list` with the correct stored selected values whenever a selected `pick-list-item` gets removed (externally / not by click) and it will get unselected.


driskull (commented):
Code looks good @y0n4. Can we add a test for this?

src/components/calcite-pick-list/calcite-pick-list.tsx
```diff
@@ -104,7 +104,10 @@ export class CalcitePickList<
 
   lastSelectedItem: ItemElement = null;
 
-  mutationObserver = createObserver("mutation", mutationObserverCallback.bind(this));
+  mutationObserver = createObserver("mutation", () => {
+    this.setUpItems();
```
jcfranco, line 108: This is not needed since [`mutationObserverCallback` calls this method](https://github.com/Esri/calcite-components/blob/15410a72c52a7ceb5e8031facccb763a5f8749de/src/components/calcite-pick-list/shared-list-logic.ts#L14).

src/components/calcite-pick-list/shared-list-logic.ts
```diff
@@ -228,6 +228,16 @@ export function setUpItems<T extends Lists>(
     }
   });
 
+  // remove items from selectedValues if the item doesn't exist in the list
```
jcfranco, line 231: Can you move this near the end of the method to keep the item-tabbing logic grouped? Bonus points if you extract this into a private method with an expressive name that eliminates the comment. ⭐

jcfranco (changes requested):
Looking good, @y0n4! 😎  

Please add a test. We want to make sure to improve test coverage as we fix/enhance things.

jcfranco:
@y0n4 Can you also revisit the commit title? It doesn't convey what got fixed.

jcfranco (approved):
🎉 🌮 🦖

jcfranco:
@y0n4 FYI, I added `value-list` to the PR's title since it also benefited from this change.

y0n4:
Thanks for including it in!

PR #3788, opened 2021-12-28
refactor(dom): Update getSlotted utility to handle querying for item(s) in default slot

driskull:
**Related Issue:** None

## Summary

refactor(dom): Update getSlotted utility to handle querying for item(s) in default slot

- Update dom utility `getSlotted` to query for an element in the default slot.
- make `slotName` param optional

This will allow us to replace `this.hasDefaultSlot = this.el.querySelector(":not([slot])") !== null;` taken from `combobox-item`.


src/utils/dom.ts
```diff
@@ -185,7 +187,7 @@ function querySingle<T extends Element = Element>(
   match = options && options.direct === false ? match : match?.parentElement === element ? match : null;
 
   const selector = options?.selector;
-  return selector ? match.querySelector<T>(selector) : match;
+  return selector ? match?.querySelector<T>(selector) : match;
```
jcfranco, line 190: Since the default slot selector will also return children in a custom element without a `<slot>`, this should also make sure that `assignedSlot` from the default slot selector result is not null.

src/utils/dom.spec.ts
```diff
@@ -115,17 +115,23 @@ describe("dom", () => {
           <span>🙂</span>
         </h2>
         <h2 slot=${testSlotName}><span>😂</span></h2>
+        <div><p>🙂</p></div>
       </slot-test>
     `;
     });
 
     describe("single slotted", () => {
```
jcfranco, line 123: The `multiple slotted` group also needs to be updated.

src/utils/dom.ts
```diff
@@ -135,22 +135,24 @@ interface GetSlottedOptions {
...
 export function getSlotted<T extends Element = Element>(
   element: Element,
-  slotName: string,
+  slotName?: string,
   options?: GetSlottedOptions
 ): T | null;
 export function getSlotted<T extends Element = Element>(
   element: Element,
-  slotName: string,
+  slotName?: string,
```
jcfranco, line 152: Can you update the signature to support both

```ts
getSlotted(el, slotName, options);
getSlotted(el, options);
```

?

src/utils/dom.spec.ts
```diff
@@ -98,7 +98,7 @@ describe("dom", () => {
         }
 
         connectedCallback(): void {
-          this.shadowRoot.innerHTML = `<slot name="${testSlotName}"></slot>`;
+          this.shadowRoot.innerHTML = `<slot name="${testSlotName}"></slot><slot></slot>`;
```
jcfranco, line 101: For readability, can the default slot be placed in a separate line below?

jcfranco (changes requested):
This is looking great! This will be a nice update to `getSlotted`. 🎉

src/utils/dom.ts
driskull: I don't think the `assignedSlot` will work on the spec test. I think we would have to setup an e2e test?

src/utils/dom.ts
driskull: Yeah, or we can just update the slotName to be apart of options? What do you think?

Do you have an example of making the second param be identified as the third?

driskull:

src/utils/dom.ts
```diff
@@ -196,6 +205,10 @@ function queryMultiple<T extends Element = Element>(
   let matches = Array.from(element.querySelectorAll<T>(slotSelector));
   matches = options && options.direct === false ? matches : matches.filter((el) => el.parentElement === element);
 
+  if (slotSelector === defaultSlotSelector) {
+    matches = matches.filter((match) => match?.assignedSlot);
```
driskull, line 209: @jcfranco It seems like this would work but we would need to move the spec tests to e2e for assignedSlot to work I think.

src/components/calcite-dropdown/calcite-dropdown.e2e.ts
```diff
@@ -794,9 +794,9 @@ describe("calcite-dropdown", () => {
 
     it("opens when dropdown-trigger is an action", async () => {
       const page = await newE2EPage();
-      await page.setContent(html` 
+      await page.setContent(html`
         <calcite-dropdown>
-          <calcite-action slot="dropdown-trigger">Open dropdown</calcite-button>
+          <calcite-action slot="dropdown-trigger">Open dropdown</calcite-action>
```
driskull, line 799: I missed this in review :( cc @Elijbet

src/utils/dom.spec.ts
```diff
@@ -119,11 +123,17 @@ describe("dom", () => {
         </h2>
         <h2 slot=${testSlotName}><span>😂</span></h2>
         <h3 slot=${testSlotName2}><span>😂</span></h3>
+        <div id="default-slot-el"><p>🙂</p></div>
       </slot-test>
     `;
+
+      const assignedSlot = document.querySelector("slot-test").shadowRoot.querySelector(`slot:not([name])`);
```
jcfranco, line 130: Is this a workaround for `Node.assignedSlot` missing? If so, just wanted to check if it's working or not based on your comments to move to an e2e test.

src/utils/dom.ts
```diff
@@ -162,24 +162,33 @@ interface GetSlottedOptions {
   selector?: string;
 }
 
+const defaultSlotSelector = "> :not([slot])";
+
 export function getSlotted<T extends Element = Element>(
```
jcfranco, line 167: I think this should be typed like so:

```ts
export function getSlotted<T extends Element = Element>(element: Element, options: GetSlottedOptions & { all: true }): T[];
export function getSlotted<T extends Element = Element>(element: Element, options?: GetSlottedOptions): T[];
export function getSlotted<T extends Element = Element>(element: Element, slotName: string | string[], options: GetSlottedOptions & { all: true }): T[];
export function getSlotted<T extends Element = Element>(element: Element, slotName: string | string[], options?: GetSlottedOptions): T[];
export function getSlotted<T extends Element = Element>(
  element: Element,
  slotNameOrOptions: string | string[] | GetSlottedOptions,
  optionsOrUndefined?: GetSlottedOptions): (T | null) | T[] { /* ... */ }
```

Can you create a separate refactor issue this one and assign it to me?

jcfranco (approved):
Code changes LGTM! 

For the remaining items I brought up (1. `getSlotted` signature, 2. `assignedSlot` check), can you create follow up issues?

src/utils/dom.spec.ts
driskull: Yes, this is a workaround.

driskull: Without this, the assignedNode won't be there. So we would have to run the getSlotted() method in the browser. Which means placing the whole utility in the browser so that all methods can be accessed.

Issue #1004, opened 2020-09-17
Tooling: @types cause Node context to leak into source


jcfranco:
### Background 

Node types are leaking into component source. For example:

```ts
    const foo: number = setTimeout(() => {}, 1000);
```

Will throw an error because of incompatible types (`NodeJS.Timeout` vs `number`).

### Desired Outcome 

No Node types in component source.

jcfranco:
https://github.com/ionic-team/stencil/commit/1a907f784f58463cfe52f87518f1a1d38a2908e5 may fix this.

julio8a:
Is this issue still valid?

jcfranco:
Still valid.

jcfranco:
Installed! Now, global interval and timeout methods resolve to the right type as they are called from their corresponding context object. The original issue will still arise whenever the global methods are called directly as TS will resolve them using Node's global declarations (pulled in by `jest` dependencies) (see https://github.com/microsoft/TypeScript/issues/842#issuecomment-301228428).

benelan:
verified on master

PR #3798, opened 2021-12-28
refactor: use global or window when calling global methods

jcfranco:
**Related Issue:** #1004

## Summary

This PR eliminates the ambiguity of global timeout and interval methods. This is needed for to avoid mismatching types since `@types/node/globals.d.ts` is getting pulled in automatically from `jest` dependencies.


jcfranco:
Resubmitted via https://github.com/Esri/calcite-components/pull/3798

PR #3799, opened 2021-12-28
refactor: use window or global when calling global interval or timeout methods

jcfranco:
**Related Issue:** #1004

## Summary

This PR eliminates the ambiguity of global timeout and interval methods. This is needed for to avoid mismatching types since `@types/node/globals.d.ts` is getting pulled in automatically from `jest` dependencies.

### DIFF
--- a/src/components/calcite-alert/calcite-alert.tsx
+++ b/src/components/calcite-alert/calcite-alert.tsx
@@ -316,7 +316,7 @@ export class CalciteAlert {
 
   /** emit the opened alert and the queue */
   private openAlert(): void {
-    clearTimeout(this.queueTimeout);
+    window.clearTimeout(this.queueTimeout);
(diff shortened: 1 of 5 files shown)


driskull (approved):

eriklharper (approved):
Just curious, why `window` in some cases and `global` in others?

benelan:
> Just curious, why `window` in some cases and `global` in others?

Browser uses `window` and node uses `global`. So `global is used for the e2e stuff and `window` is used for the component.

benelan:
Alternatively, [globalThis](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) is newish and works on both environments.

jcfranco:
`globalThis` won't work since it has the same problem as using the global directly. TS will use `node`'s global declaration and have the wrong return type.

Issue #2800, opened 2021-08-12
Enhancement: Alert - update autoDismissDuration to accept CSS time values


asangma:
### Description
Update Alert's autoDismissDuration property to also accept valid CSS time values: `ms` and `s`.

### Acceptance Criteria 
Valid CSS duration values are accepted.
Existing strings continue to be accepted.

### Relevant Info 
Sometimes "fast" is too slow.

#### Which Component
Alert

#### Example Use Case
Showing a success message in a blocked ui.

asangma:
Alternatively, make "fast" faster.
Or add "turbo".

julio8a:
Tweaking values would be the best approach here. @asangma and @macandcheese can you guys work on this to find the optimal timing updates.

macandcheese:
Will be hard to find values that please everyone, can we consider having a single default that is overridable (and has a minimum speed set and documented to prevent folks putting an alert up for .005 ms or whatever)

jcfranco:
Let's see if we can fine-tune the existing values first. Granted, it may not please everyone, but at least we'll have a consistent way of defining durations across components.

macandcheese:
I'm happy to defer to whatever values @asangma wants here.

anveshmekala:
@asangma please update the  expected values of `fast| turbo` for `calcite-alert` in the description

macandcheese:
It does seem a bit weird to pass either a css value or a string as a value. Could we just update the underlying variables used and only accept slow/medium/fast? 

Otherwise just removing the duration as a prop, and accepting a css variable to override a single default seems like another option. 

I’m comfortable just making fast… faster if needed.

jcfranco:
> It does seem a bit weird to pass either a css value or a string as a value. Could we just update the underlying variables used and only accept slow/medium/fast?

👍

anveshmekala:
Here is the [prototype](https://github.com/Esri/calcite-components/runs/4654039785?check_suite_focus=true) of how it may look like if we allow users to control the `autoDimissDuration`.

cc: @macandcheese , @asangma , @jcfranco

macandcheese:
Yeah, I guess I don’t see a reason to let people set their own value here if we are already providing three speeds. 

There is an open issue to pause the timer when hovering or focusing over the alert which would solve for the “dismisses too fast while trying to interact” case- but generally even having three seems overkill, imo. 

[...]

PR #3803, opened 2021-12-28
feat(alert): allow time values in autoDimissDuration

anveshmekala:
**Related Issue:** #2800 

## Summary

This allows users to parse time values in `autoDismissDuration` prop for `calcite-alert` and default value is set to `5000ms`


jcfranco:
Thanks for prototyping this, @anveshmekala. Per the discussion on #2800, we'll be tweaking the named duration values instead. Closing based on this.

Issue #3699, opened 2021-12-14
Bug: action-bar and panel don't line up when nested side by side. 


Elijbet:
### Actual Behavior

Action-bar and panel don't line up when nested side by side. 

### Expected Behavior

Feedback from @macandcheese:
On one hand it would be hard to ensure they are always the same height (since you can slot anything into header-actions of a panel), but it would be nice for the simple case of title + action bar to line up (you can see this same issue live in Map Viewer).

### Reproduction Steps and Sample

<img width="462" alt="action-bar-border" src="https://user-images.githubusercontent.com/19231036/145911779-f544f6a7-d43e-407f-8618-5f6bc120534f.png">

### Relevant Info

Reproducible version: @Esri/calcite-components@v1.0.0-beta.71

AdelheidF:
The padding probably needs to be fixed first (https://github.com/Esri/calcite-components/issues/3494). 
https://codepen.io/afreitag/pen/ZEJjaRa

asangma:
I think this would be a low-risk easy adjustment, i.e. reducing the top/bottom padding to 0.75rem.

AdelheidF:
this looks good since beta.73

PR #3805, opened 2021-12-29
fix(Panel): match header height and border color to Action

asangma:
**Related Issue:** #3699

## Summary
Updates how header height is achieved in order to match the height to an Action when there's only a heading in the Panel's header.

Also updates border colors to match.

Of course, if there's a summary or long heading text, the alignment no longer happens.

Screen Shot 2021-12-29 at 9 15 22 AM


driskull (approved):
LGTM

PR #3812, opened 2021-12-29
docs(accordion): nest accordion-items not accordions

benelan:
**Related Issue:** NA

## Summary
Question came up on the Teams channel, so adding doc for clarity


driskull (approved):
👏 🎬 👍

PR #3813, opened 2021-12-30
fix(slider): maxValue thumb positioned above minValue when overlapped…

anveshmekala:
**Related Issue:** #3101 

## Summary

In mirrored `calcite-slider` when overlapped at edge, `maxValue` handle should be above `minValue` and `hyphen` position matches the non-mirrored slider.


src/components/calcite-slider/calcite-slider.tsx
```diff
@@ -1150,7 +1150,7 @@ export class CalciteSlider implements LabelableComponent, FormComponent {
     const labelOffset = labelFontSize / 2;
 
     if (labelTransformedOverlap > 0) {
-      hyphenLabel.classList.add("hyphen");
+      hyphenLabel.classList.add("hyphen", "handle__label--flex");
```
driskull, line 1153: Does this class need to be removed at some point in runtime?

src/components/calcite-slider/calcite-slider.scss
```diff
@@ -34,6 +34,10 @@
   font-size: var(--calcite-slider-container-font-size);
 }
 
+.handle__label--flex {
```
driskull, line 37: Could we name this something better to reflect its purpose?

driskull (commented):
Comments! 💬

src/components/calcite-slider/calcite-slider.tsx
anveshmekala: It can be removed considering when hyphenation is not required.

src/components/calcite-slider/calcite-slider.scss
anveshmekala: may be  `hyphen--wrap` ?

driskull (approved):
Looks good @anveshmekala.

is there a way we can have a test for this? Either screener or e2e?

anveshmekala:
> Looks good @anveshmekala.
> 
> is there a way we can have a test for this? Either screener or e2e?

We have screener test for edge positioning of labels. For some reason, screener is not using the current master branch to run the test. Any thoughts? @eriklharper

Issue #3685, opened 2021-12-10
calcite-value-list doesn't emit the calciteListOrderChange event when using a keyboard


jmanke:
### Actual Behavior

When sorting a `calcite-value-list` with a keyboard, no `calciteListOrderChange` event is emitted.

### Expected Behavior

When sorting a `calcite-value-list` with a keyboard, the `calciteListOrderChange` event should be emitted since the list is being re-ordered.

### Reproduction Steps and Sample

_Sample_: [codepen](https://codepen.io/jmanke/pen/YzrGeqZ?editors=1111)

1. Create a `calcite-value-list` with `drag-enabled` attribute and at least two `calcite-value-list-item`s
2. Listen for the `calciteListOrderChange` event on the `calcite-value-list`
3. Use a keyboard to sort the list
4. Observe no `calciteListOrderChange` event is emitted

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-beta.71`

- [x] CDN
- [ ] NPM package

Regression? No

jmanke:
Noticed on @next that the event detail is different when using a keyboard. When using a keyboard the event detail is the element itself. When using a mouse the event detail is the value assigned to the `calcite-value-list-item`.

jcfranco:
My bad. I'll submit a fix in a few.

benelan:
verified on `beta.74`

PR #3816, opened 2021-12-30
fix(value-list): emit list order change event when value list items are reordered via the keyboard

jcfranco:
**Related Issue:** #3685

## Summary

The value list was only emitting order change events when reordered with the mouse (via SortableJS). This PR adds the eventing logic when reordered via keyboard. 

Note that the event be emitted as each item is moved with the up/down keys.

### DIFF
--- a/src/components/calcite-value-list/calcite-value-list.e2e.ts
+++ b/src/components/calcite-value-list/calcite-value-list.e2e.ts
@@ -85,6 +85,7 @@ describe("calcite-value-list", () => {
 
     it("works using a mouse", async () => {
       const page = await createSimpleValueList();
+      const listOrderChangeSpy = await page.spyOnEvent("calciteListOrderChange");
 
       await dragAndDrop(
         page,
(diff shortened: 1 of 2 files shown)


driskull (approved):
so clean, so fresh. 🧼 ✨ 🛀

jcfranco:
Failing E2E test is unrelated (for reals this time, 😅):

```
Summary of all failing tests
FAIL src/components/calcite-input/calcite-input.e2e.ts (341.271 s)
  ● calcite-input › number type › allows typing negative decimal values

    expect(received).toBe(expected) // Object.is equality

    Expected: "-0.001"
    Received: "-1"

      948 |       await page.keyboard.type("-0.001");
      949 |       await page.waitForChanges();
    > 950 |       expect(await element.getProperty("value")).toBe("-0.001");
          |                                                  ^
      951 |     });
      952 |
      953 |     it("disallows typing any letter or number with shift modifier key down", async () => {

      at Object.<anonymous> (src/components/calcite-input/calcite-input.e2e.ts:950:50)
```

All other checks look fine. Merging!

Issue #2948, opened 2021-08-30
Research: revisit calcite-graph's default SVG width


caripizza:
### Background 
I'm opening up this issue [per Franco's request](https://github.com/Esri/calcite-components/pull/2867#issuecomment-908473509) in my slider style update pr:

> Is there a use case where the SVG wouldn't need to fill the whole width? If not, shouldn't it always be full width and up to the host component to modify its width?

### Desired Outcome 
Answer the above questions and determine path forward with calcite-graph (and calcite-slider's histogram).

jcfranco:
Discussed with @benelan and @macandcheese during triage and decided we should refactor `calcite-graph`'s styling so that it is full-width by default.

benelan:
verified on `beta.75` https://codepen.io/benesri/pen/GROgvRJ

PR #3819, opened 2021-12-30
refactor(graph): updated graph default width to be 100%

jcfranco:
**Related Issue:** #2948

## Summary

This PR ensures the graph SVG is full width by default. Before, it would require a special data property for the chart to be full width. This has been removed.


driskull (approved):
Code changes look good. Seems like the calcite-graph still has a width property set to 300. It seemed like the issue was more about removing those width/height properties and just making them 100% of container?

jcfranco:
The original issue was about removing the slider-specific option for full width, but you brought up a good point. If the chart SVG occupies the full width of its container, there's no need for the `width` and `height` props, right?

jcfranco:
@macandcheese Can you review the visual diffs? The latest changes made the histogram selection dividers narrower and now share the same width and location as the handles, which seems like an improvement, at least to my 👀 . 

basic-histogram-diff

color-histogram-diff

driskull (approved):
AWESOME SAUCE 🎊

macandcheese:
These changes look good to me - cc @rmstinson @asangma for Figma update.

jcfranco:
Current E2E failure is unrelated:

```
Summary of all failing tests
FAIL src/components/calcite-dropdown/calcite-dropdown.e2e.ts (69.154 s)
  ● calcite-dropdown › opens the dropdown with click, enter, or space › opens when dropdown-trigger is an action

    expected event "calciteDropdownOpen" to have been called 1 times, but was called 2 times

      828 |       await page.waitForChanges();
      829 |       expect(await dropdownWrapper.isVisible()).toBe(true);
    > 830 |       expect(calciteDropdownOpen).toHaveReceivedEventTimes(1);
          |                                   ^
      831 |     });
      832 |   });
      833 |

      at Object.<anonymous> (src/components/calcite-dropdown/calcite-dropdown.e2e.ts:830:35)
```

The rest of the tests and checks are ✅. Merging!

Issue #3209, opened 2021-10-12
A11y: Value list item with non-interactive prop should not read in screen reader or be in the tab order


paulcpederson:
### Summary

The `non-interactive` prop should make the value list item completely non-interactive. While currently it prevents clicking via mouse, it's still interactive to keyboard users and is read as an interactive menu item to users using a screen reader.

### Actual Behavior

1. NVDA reads "Clickable _[label]_, _[label]_ menu item, not checked"
2. Label is still in the tab order and has a focus state 

### Expected Behavior

Nothing should be read by the screen reader and the item should not be a menu item. When non-interactive all interactive roles should be removed and element should be removed from the tab order. 

### Reproduction Steps and Sample

_Sample_: https://codepen.io/paulcp/pen/JjyoymE

1. Tab to the value list, note the focus outline
2. Turn on a screen reader, note when tabbed to the label it still has an interactive set of roles and is read aloud as if it's interactive

### Relevant Info

_Version_: `@esri/calcite-components@beta.66`

eriklharper:
What is the difference between `non-interactive` and `disabled`?

geospatialem:
Blocked due to re-work on the `list` component. 

Will revisit, and determine feasibility, or if the `non-interactive` prop will proceed with the updated component structure.

geospatialem:
> Blocked due to re-work on the `list` component.
> 
> Will revisit, and determine feasibility, or if the `non-interactive` prop will proceed with the updated component structure.

@driskull If memory serves, were we thinking of dropping the `nonInteractive` prop from list-based components? Thinking we could either close this out, or revisit post 1.0 release depending on next steps.

driskull:
Yes, its already been removed

geospatialem:
Closing as this is no longer applicable with the removal of the `nonInteractive` prop.

PR #3824, opened 2021-12-30
fix(value-list-item, pick-list-item): make item unable to be tabbed when in non interactive state

y0n4:
**Related Issue:** #3209

## Summary
When there is a non-interactive `value-list-item` and `pick-list-item`. Those items cannot be selected when pressing the tab or arrow keys on the label content area


src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -282,30 +282,50 @@ export class CalcitePickListItem {
...
-          onKeyDown={this.pickListKeyDownHandler}
-          ref={(focusEl): HTMLLabelElement => (this.focusEl = focusEl)}
-          tabIndex={0}
-        >
-          <div
-            aria-checked={this.selected.toString()}
-            class={CSS.textContainer}
-            role="menuitemcheckbox"
+        {this.nonInteractive ? (
+          <label
```
driskull, line 308: If this item is non-interactive, I don't think it needs the `aria-label`, `onclick`, `onkeydown`, or `<label>` element. Is that right? It may not need to be focused either. I'm not sure on that one.

The content wouldn't need a checkbox role or aria-checked, right? I think the role would just be `menuitem` and not `menuitemcheckbox` or maybe not even be a menuitem if it doesn't need to be focused at all.

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -282,30 +282,50 @@ export class CalcitePickListItem {
...
 
-  render(): VNode {
+  renderLabelContent(): VNode {
     const { description, label } = this;
 
+    return (
+      <div
+        aria-checked={this.selected.toString()}
+        class={CSS.textContainer}
+        role="menuitemcheckbox"
```
driskull, line 292: If this is non-interactive, does it need this role?

driskull (commented):
Some comments. I'm not really sure how this should behave. Can you help here @asangma ?

y0n4:
@driskull @asangma 

I agree it doesn't need `aria-label`. I found this at the [mdn](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute) page. Please see the first bullet point
Screen Shot 2022-01-05 at 2 56 04 PM

For `label`, I also checked at the [mdn](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label) page. It refers to label being used when data is being entered so maybe don't need it? Also if `aria-label` is removed there is no label to read out loud to screen reader users.
Screen Shot 2022-01-05 at 3 05 47 PM

I agree it doesn't need `aria-checked`. This [mdn](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role) page says this is used for interactive purposes (keyboard, tab, focus).
Screen Shot 2022-01-05 at 3 36 10 PM

For `role`, I also think it is not needed if we do not want to give any interactivity. Also `role = "menuitem"` is considered as an [option](https://www.w3.org/TR/wai-aria-1.1/#menuitem)

driskull:
@y0n4 I think we might need to meet on this one with @asangma to discuss the correct approach. 

I'm not really sure of the purpose of `nonInteractive`. Like why we need an item that can't be interacted with. Is that not the same as `disabled`??

```ts
/**
   * @internal When true, the item cannot be selected by user interaction.
   */
```

I do know that just removing the tabIndex isn't the right approach. It will still have the role and click handlers, etc.

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
asangma: I think you're right on that @driskull. But we can discuss in a quick chat...?

PR #3825, opened 2021-12-30
refactor(pick-list, value-list): add proper types to list change event payload

jcfranco:
**Related Issue:** N/A

## Summary

Types the list change event payload for both pick and value lists. I had to use `any` for the shared implementation since it requires some advanced TS-fu. I'll create a follow-up issue for that.

### DIFF
--- a/src/components/calcite-pick-list/shared-list-logic.ts
+++ b/src/components/calcite-pick-list/shared-list-logic.ts
@@ -151,7 +151,7 @@ export function keyDownHandler<T extends Lists>(this: List<T>, event: KeyboardEv
 }
 
 export function internalCalciteListChangeEvent<T extends Lists>(this: List<T>): void {
-  this.calciteListChange.emit(this.selectedValues);
+  this.calciteListChange.emit(this.selectedValues as any);
 }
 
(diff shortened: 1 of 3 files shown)


src/components/calcite-pick-list/shared-list-logic.ts
```diff
@@ -151,7 +151,7 @@ export function keyDownHandler<T extends Lists>(this: List<T>, event: KeyboardEv
 }
 
 export function internalCalciteListChangeEvent<T extends Lists>(this: List<T>): void {
-  this.calciteListChange.emit(this.selectedValues);
+  this.calciteListChange.emit(this.selectedValues as any);
```
driskull, line 154: Can this be typed more or no? Just wondering

driskull (approved):
👍

src/components/calcite-pick-list/shared-list-logic.ts
jcfranco: I'll do that in the follow-up issue I mentioned. I couldn't get TS to stop complaining by using the conditional types.

Issue #3629, opened 2021-12-03
ListItem is missing line-height definition


asangma:
### Actual Behavior
ListItem's label and description are missing line-height definitions, e.g. it's using `text--1`.

### Expected Behavior
ListItem's label and description should have line-height defined, e.g. `text--1h`.
Could probably be set at the :host or `.container` level.

### Reproduction Steps and Sample

_Sample_: https://codepen.io/mtb-alan/pen/GRMpjqd

1. Inspect the label and/or description.

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-<version>`

- [ ] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-<version>`

benelan:
@asangma can you verify this one? The line height is on content rather than host/container like suggested, so I just want to make sure that's what you're looking for.

asangma:
Looking good!

PR #3832, opened 2021-12-31
fix(list-item): apply line-height to label and description

Elijbet:
**Related Issue:** #3629

## Summary

Added tailwind text-n2-wrap to the content class. This defines line-height at 1.375 rem.


driskull (approved):
👍

Issue #3659, opened 2021-12-08
Bug: calciteDatePickerChange always get triggered twice


fendiaoxiaoshuazi:
### Actual Behavior

The calciteDatePickerChange event of calcite-input-date-picker always gets triggered twice when the date picker is changed.
![calcite date picker](https://user-images.githubusercontent.com/26078106/145198751-a296adf2-4cff-406c-af35-71b03e1fc46d.gif)

### Expected Behavior

The calciteDatePickerChange event should only be triggered once when the date picker is changed
### Reproduction Steps and Sample

_Sample_: https://codepen.io/nan9266/pen/VwzQdYq?editors=1011

1. Choose a value for the first date picker
2. Observe in the console, "Test picker changed" has been logged twice
3. Choose a value for the second date picker
4. Observe in the console, "NewInputPicker changed." has been logged twice

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-beta.69>`

- [ ] CDN
- [x] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-<version>`

benelan:
verified on beta.75

PR #3833, opened 2021-12-31
fix(date-picker): ensure change event gets emitted on user interaction

jcfranco:
**Related Issue:** #3659 

## Summary

This updates `calcite-date-picker` to only emit the change event when clicked or selected via keyboard.


driskull (approved):
👍 💯

Issue #3835, opened 2022-01-01
Refactor: replace popper.js with floating-ui


driskull:
### Description 

Refactor: replace popper.js with floating-ui

### Proposed Advantages 

Future support

### Relevant Info 

https://github.com/floating-ui/floating-ui
https://floating-ui.com/
https://floating-ui.com/docs/getting-started

#### Which Component

All that use popper.js

benelan:
This has been installed for two weeks and I've been using/testing/verifying floating-ui related components with no issues. I'm marking this as verified, awesome work Matt!! :trophy:

PR #3836, opened 2022-01-01
refactor: replace popper.js with floating-ui

driskull:
**Related Issue:** #3835

## Summary

Popper is now Floating UI! For Popper v2, visit [its dedicated branch.](https://github.com/floating-ui/floating-ui/tree/v2.x)

https://github.com/floating-ui/floating-ui

- Renames references to 'popper' to be 'floating-ui'
  - Renames files from popper to floating-ui.
  - Updates spec test file to match new name.
- Updates dependencies 
  - Removes popper.js dependency
  - Adds floating-ui dependency
- CSS remains unchanged
- Updates interface names to match floating-ui
- Adds utility method `positionFloatingUI` to position the floating ui
- Adds utility methods `connectFloatingUI` and `disconnectFloatingUI` to connect scrollParents for both the floating element and reference element.
- Moves modifiers from each component into the `positionFloatingUI` utility. There are 3 types of "middleware" groups that we handle currently. 
  - `type UIType = "menu" | "tooltip" | "popover";` 
  - Just setting one of these types is easier and more consistent for components rather than them each having a 'getModifiers()' method.
- Creates an interface `FloatingUIComponent` that components with floating counterparts can extend.

## Interface

```ts
export interface FloatingUIComponent {
  /**
   * Describes the type of positioning to use for the overlaid content. If your element is in a fixed container, use the 'fixed' value.
   */
  overlayPositioning: OverlayPositioning;

  /**
   * Determines where the component will be positioned relative to the referenceElement.
   *
   * Possible values: "auto", "auto-start", "auto-end", "top", "right", "bottom", "left","top-start", "top-end", "right-start", "right-end", "bottom-start", "bottom-end", "left-start", "left-end", "leading-start", "leading", "leading-end", "trailing-end", "trailing", "trailing-start", "leading-leading", "leading-trailing", "trailing-leading", "trailing-trailing", "top-leading", "top-trailing", "bottom-leading", "bottom-trailing", "right-leading", "right-trailing", "left-leading", or "left-trailing".
   *
   */
  placement: LogicalPlacement;

  /**
   * Updates the position of the component.
   */
  reposition(): Promise<void>;
}
```

## References

https://github.com/floating-ui/floating-ui
https://floating-ui.com/
https://floating-ui.com/docs/getting-started

### Scrolling support

https://floating-ui.com/docs/computePosition#scrollparents


driskull:
@jcfranco when you have a chance, can you give this a look over? 👀

jcfranco:
Nice! This looks like a great start, but I think we should wait until the package reaches 1.0.0 before moving it over.

jcfranco (changes requested):
As stated previously, this is a great start, but we should wait until the replacement is in a more mature state. There is no immediate need to switch over at this time, is there?

driskull:
> wait until the replacement is in a more mature state. There is no immediate need to switch over at this time, is there?

Nope, no need. We can switch over whenever.

src/utils/floating-ui.ts
```diff
@@ -0,0 +1,392 @@
...
+  "top-start" | "top" | "top-end" | "bottom-start" | "bottom" | "bottom-end"
+>;
+
+export const defaultMenuPlacement: MenuPlacement = "bottom-start";
+
+export interface FloatingUIComponent {
+  /**
+   * Whether the component is opened.
+   */
+  active: boolean;
```
jcfranco, line 113: I believe this should be `open` to align with the recent API consistency changes.

src/utils/floating-ui.ts
```diff
@@ -0,0 +1,392 @@
...
+*/
+export function getEffectivePlacement(floatingEl: HTMLElement, placement: LogicalPlacement): EffectivePlacement {
+  const placements = ["left", "right"];
+
+  if (getElementDir(floatingEl) === "rtl") {
+    placements.reverse();
+  }
+
+  return placement
+    .replace(/-leading/gi, "-start") // todo: remove. Has already been removed in typings.
```
jcfranco, line 228: Suggestion: to avoid unnecessary TS errors, you can add the deprecated values in a `DeprecatedPlacement` type or similar.

src/utils/floating-ui.ts
```diff
@@ -0,0 +1,392 @@
...
+  referenceEl: ReferenceElement;
+  floatingEl: HTMLElement;
+  overlayPositioning: Strategy;
+  placement: LogicalPlacement;
+  disableFlip?: boolean;
+  flipPlacements?: EffectivePlacement[];
+  offsetDistance?: number;
+  offsetSkidding?: number;
+  arrowEl?: HTMLElement;
+  type: UIType;
```
jcfranco, line 270: I'm really digging being able to pass the type.

src/utils/floating-ui.ts
```diff
@@ -0,0 +1,392 @@
...
+
+  if (cleanup) {
+    cleanup();
+  }
+
+  cleanupMap.delete(component);
+}
+
+export function hypotenuse(sideA: number, sideB: number): number {
+  return Math.sqrt(Math.pow(sideA, 2) + Math.pow(sideB, 2));
```
jcfranco, line 382: Use [`Math.hypot()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/hypot) instead.

src/components/combobox/combobox.tsx
```diff
@@ -640,50 +656,9 @@ export class Combobox
...
-    const flipModifier: Partial<StrictModifiers> = {
-      name: "flip",
-      enabled: true
-    };
-
-    flipModifier.options = {
-      fallbackPlacements: this.filteredFlipPlacements || popperMenuComputedPlacements
-    };
-
-    const eventListenerModifier: Partial<StrictModifiers> = {
```
jcfranco, line 655: Do we need to take care of this or does `floating-ui ` handle this better? I remember we had a perf issue before. Awesomesauce if it's the latter. 🎉🥫

jcfranco (approved):
Awesome work, @driskull! Thanks for staying on top of this. Had a couple of comments, but this looks great!

src/assets/styles/includes.scss
```diff
@@ -6,7 +6,7 @@
 @import "focus";
 @import "host";
 @import "spacing";
-@import "popper";
+@import "floating-ui";
```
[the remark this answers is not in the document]
driskull: Lets think of a better file name incase it changes again. Maybe something on the backburner if not that important.

src/utils/floating-ui.ts
driskull: Nice catch. I'll update

src/utils/floating-ui.ts
driskull: nice. thanks

src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -169,6 +174,11 @@ export class InputDatePicker
     }
   }
 
+  /**
+   * @internal
+   */
+  @Prop({ mutable: true, reflect: true }) open = false;
```
driskull, line 180: @jcfranco `open` doesn't exist on `input-date-picker` yet. I added this prop for now.

driskull: created: https://github.com/Esri/calcite-components/pull/4972

src/assets/styles/includes.scss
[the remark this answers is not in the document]
jcfranco: Back-burner sounds good as long as we don't have any public-facing APIs affected.

src/components/input-date-picker/input-date-picker.tsx
jcfranco: Awesome, thanks!

driskull:
@jcfranco good to merge?

PR #3854, opened 2022-01-04
ci(build): move commit from npm script to action

benelan:
**Related Issue:** #3782

## Summary
The fix I made in the PR linked above worked, and the action ran the update and audit successfully. However the action still failed and it doesn't say why. 
https://github.com/Esri/calcite-components/runs/4676569957?check_suite_focus=true#step:5:108163

It may be an issue with the commit, so I moved it from the npm script to the gh action. This is how its done in both the doc and browserlist actions.
https://github.com/Esri/calcite-components/blob/master/.github/workflows/update-browserslist-db.yml#L19
https://github.com/Esri/calcite-components/blob/master/.github/workflows/update-doc.yml#L25


jcfranco (approved):
👍🏼

PR #3855, opened 2022-01-04
Sam11927/pick list content slot

benelan:
**Related Issue:** #

### DIFF
--- a/src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
+++ b/src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
@@ -282,6 +282,17 @@ export class CalcitePickListItem {
     ) : null;
   }
 
+  renderCenterContent(): VNode {
+    const { el } = this;
+    const hasCenterContent = getSlotted(el, SLOTS.centerContent);
+
+    return hasCenterContent ? (
+      <div>
+        <slot name={SLOTS.centerContent} />
+      </div>
+    ) : null;
+  }
+
   render(): VNode {
     const { description, label } = this;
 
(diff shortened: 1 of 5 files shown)


ethanbdev:
> Could you fix the conflicting files and get the e2e tests passing @sam11927?

The e2e test still needs to be updated with the new slot added

asangma:
I think we should hold off on this for now.
There are some good things here that aligns with our plans for PickListItem as well as our other list components.
However, we will want whatever happens with PickListItem to align with our broader patterns in all of our lists.

ethanbdev:
> I think we should hold off on this for now. There are some good things here that aligns with our plans for PickListItem as well as our other list components. However, we will want whatever happens with PickListItem to align with our broader patterns in all of our lists.

Okay, yeah I definitely defer to you on this. Is there any set plans for that? I can share the use case we have if it would be helpful

jcfranco:
> Is there any set plans for that? 

We're still working out through some details for list/item patterns. We can share this once the plans are more concrete. 

> I can share the use case we have if it would be helpful

Yes, please. This could be useful.

jcfranco:
I'm going to go ahead and close this since we're looking to tackle it separately.

jcfranco:
@sam11927 @ethanbdev Thanks for working on this, btw. We do appreciate the effort that went into this.

ethanbdev:
> Yes, please. This could be useful.

Here you go, I can put this on an issue too..

Basically the idea is we have a selection scenario where the user should select a field, and the statistics to calculate on that field. Wanted to streamline this by putting the statistic selection in the same place as the field selection, which is using a pick list. Here is a capture of the design:
image

Ideally the chips that are shown being used for selection would fit into a content slot of a pick list item, so we get the nice selection logic from it.

PR #3856, opened 2022-01-04
feat(alert): add support for alert position

macandcheese:
**Related Issue:** https://github.com/Esri/calcite-components/issues/3033

## Summary
Adds the ability to set the `placement` of the `calcite-alert` component:

https://user-images.githubusercontent.com/4733155/148124651-4b8461ad-b7a3-4207-967a-b85876fc76c2.mov

Accepts the following values: `bottom`, `bottom-end`, `bottom-start`, `top`, `top-end`, `top-start`, and defaults to `bottom` to retain current behavior.

Note that at small screen sizes, only the "top" and "bottom" difference will be noticeable, as the alert adheres to a max-width and will appear centered.

Design fyi : @mitc7862 @rmstinson @asangma

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -48,7 +48,7 @@
         "@types/sortablejs": "1.10.7",
         "@types/standard-version": "7.0.1",
         "@types/yargs": "17.0.8",
-        "@typescript-eslint/eslint-plugin": "5.8.0",
+        "@typescript-eslint/eslint-plugin": "5.9.0",
         "@typescript-eslint/parser": "5.9.0",
         "@whitespace/storybook-addon-html": "5.0.0",
         "autoprefixer": "10.4.1",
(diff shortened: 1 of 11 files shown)


driskull:
I wonder if we should be using `popper.js`/`floating-ui` to position the alert instead of doing it on our own? It would sync up the available positions and interfaces used for that.

macandcheese:
> I wonder if we should be using `popper.js`/`floating-ui` to position the alert instead of doing it on our own? It would sync up the available positions and interfaces used for that.

Makes sense - right now both Alert and Modal are similar in that they are both positioned around the page as fixed elements  (modal has `docked`, `fullscreen`, etc.), not anchored to an element absolutely.

When we migrate to `floating-ui`, maybe that is a good time refactor these components? Or should we hold off on this PR until then?

macandcheese:
> I wonder if we should be using `popper.js`/`floating-ui` to position the alert instead of doing it on our own? It would sync up the available positions and interfaces used for that.

Makes sense - right now Alert and Modal are similar in that they are both positioned around the page as fixed elements  (modal has `docked`, `fullscreen`, etc.), not anchored to an element absolutely.

When we migrate to `floating-ui`, maybe that is a good time refactor these components? Or should we hold off on this PR until then?

driskull:
We can refactor after and the PR can continue I think.

I think we can just anchor to the body or window when we refactor. (use body as reference element).

macandcheese:
@driskull I updated the prop name and values here to match Popper-based components, so any refactor to Alert will hopefully be additive and non-breaking (or at least consistent with other Popper -> floating-ui updates).

driskull (approved):
👍

Issue #3720, opened 2021-12-14
[Action] When used as dropdown trigger, can't activate dropdown with space key


macandcheese:
### Actual Behavior
When used in the `dropdown-trigger` slot of `calcite-dropdown`, it is hard as a keyboard user to interact with the `calcite-action` component to open the dropdown. A user needs to hold down `space` key _until the animation of the dropdown opening completes_, otherwise the dropdown won't open completely and revert to a closed state.

### Expected Behavior
As a keyboard user, I expect the dropdown to open immediately when interacting with a slotted `calcite-action`.

### Reproduction Steps and Sample

_Sample_:

1. View sample: https://jsbin.com/qipezarebo/edit?html,output
2. Interact with the `calcite-button` with the `space` key - notice the dropdown opens no matter how long you press `space`
3. Try the same with the `calcite-action` - notice that the animation will cancel if the `space` key isn't held until animation completion.

### Relevant Info
This might affect assistive tech that replicates a keyboard interaction as well.

_Reproducible version_: beta.71

- [x] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-<version>`

ellenupp:
In dashboards we are using a dropdown triggered by a calcite-fab, which suffers from the same problem. 
https://codepen.io/ellupp/pen/LYyKQwq?editors=1000

PR #3857, opened 2022-01-04
fix(dropdown): exclude calcite action from keyDownHandler 

Elijbet:
**Related Issue:** #3720

When button is selected with space or enter, the browser default behavior is that it triggers the click handler. If we call the click handler separately with keyboard, we end up calling it twice which then toggles it on, and immediately off, resulting in the described issue.

Here is a demo of the issue that was occurring.
demo of calling click handler twice on keyboard

Update: Trigger slot should support any slotted content, eg link, div, calcite-link, etc. Removing call for click handler on keyboard interaction removed that ability, so I added it back. Instead, I added calcite-action to a list of ignored node types in keyDownHandler. This isn’t great so there will be a follow up issue to refactor.


driskull:
Nice catch. The enter key should already work with button. It was probably a div or something at one point.

src/components/calcite-dropdown/calcite-dropdown.e2e.ts
```diff
@@ -753,6 +753,60 @@ describe("calcite-dropdown", () => {
     expect(await dropdownWrapper.isVisible()).toBe(false);
   });
 
+  describe("opens the dropdown with click, enter, or space", () => {
+    it("opens when dropdown-trigger is a button", async () => {
+      const page = await newE2EPage();
```
driskull, line 758: nitpick: You can use `newE2EPage({html: "<html>"})` and then you won't have to call `setContent`.

src/components/calcite-dropdown/calcite-dropdown.tsx
```diff
@@ -409,10 +409,6 @@ export class CalciteDropdown {
     ) {
       if (target.nodeName !== "BUTTON" && target.nodeName !== "CALCITE-BUTTON") {
```
macandcheese, line 410: Can we scope this fix to the calcite-button? As it stands this removes the ability for any other element (`calcite-link`, a `div`, etc.) to trigger the dropdown via keyboard.

src/components/calcite-dropdown/calcite-dropdown.e2e.ts
```diff
@@ -753,6 +753,60 @@ describe("calcite-dropdown", () => {
...
+        </calcite-dropdown> 
+      `);
+      const element = await page.find("calcite-dropdown");
+      const trigger = await element.find("calcite-button[slot='dropdown-trigger']");
+      const dropdownWrapper = await page.find("calcite-dropdown >>> .calcite-dropdown-wrapper");
+
+      expect(await dropdownWrapper.isVisible()).toBe(false);
+      await trigger.click();
+      await page.waitForChanges();
+      expect(await dropdownWrapper.isVisible()).toBe(true);
```
driskull, line 780: Should we also test a space/enter key here? Maybe close it and try again with space and enter key. The test name can be a bit deceiving.

driskull: Maybe also test that the event is only called once each time?

driskull (changes requested):
Just need to beef up the tests. 🐄

src/components/calcite-dropdown/calcite-dropdown.tsx
driskull: I think this is probably out of the scope of this PR but I agree, this code looks suspicious. We shouldn't be checking node types here.

driskull: This one also: `const itemToFocus = target.nodeName !== "A" ? target : target.parentNode;`

src/components/calcite-dropdown/calcite-dropdown.e2e.ts
Elijbet: Added space/enter to tests + check for how many times the event has been called. Adam pointed out that trigger slot should support any slotted content, eg link, div, calcite-link, etc. Removing call for click handler on keyboard interaction removed that ability, so I added it back. Instead, I added a check for target.nodeName !== "CALCITE-ACTION” to exclude that along with the button. Matt was suggesting to get rid of checking node types, as these very feeble and can easily break if the nodeNames change. But that this can be handled in another pr. I added some demos so I can see it in action, let me know if I should remove these.

src/demos/calcite-dropdown.html
```diff
@@ -37,6 +37,10 @@
         margin: 25px 0;
         border-top: 1px solid var(--calcite-ui-border-2);
       }
+
+      *:focus {
```
driskull, line 41: can you remove this styling please :)

src/components/calcite-dropdown/calcite-dropdown.tsx
```diff
@@ -407,7 +407,11 @@ export class CalciteDropdown {
       this.triggers.includes(target) ||
       this.triggers.some((trigger) => trigger.contains(target))
     ) {
-      if (target.nodeName !== "BUTTON" && target.nodeName !== "CALCITE-BUTTON") {
```
driskull, line 410: My advice would be to not modify this as part of this PR> It seems to be out of scope. The fix isn't a feat to support action elements here.

driskull (commented):
Final comments

src/components/calcite-dropdown/calcite-dropdown.tsx
driskull: Ok. we can leave this code as is but can you create a follow up issue to refactor this code?

We shouldn't have to keep blacklisting elements here when a different button type of element is slotted.

We can discuss how to clean it up in the new issue.

driskull (approved):
👍 

Can you just remove that focus styling?

Elijbet:
> 👍
> 
> Can you just remove that focus styling?

Yep, and modify the PR title as discussed.

Issue #3646, opened 2021-12-07
Closing and Opening of Calcite Tree Children does not work as expected


moosetraveller:
### Description

I was not sure if that would be a bug or an enhancement. In my opinion it's a bug. However, I was not sure if the behaviour is as intended by design. Therefore, I submit an enhancement issue. Please change labels if applicable/needed.

Click on the expand/collapse icon, checkbox or label of the 'group' item does not work as expected when `selectionMode="ancestors"`:

![Example Tree 2](https://user-images.githubusercontent.com/3456293/145121516-5f3b0580-1ea9-44e1-b928-681b7173844d.gif)

### Acceptance Criteria 

- click on the checkbox should only select/deselect all items within the 'group' and not close the 'group' (if open) 
    - reason: a user may want to deselect some children (if there are 10 children and 9 of them should be selected, a user is faster to select the whole group and deselect one child)
- click on label of a 'group' should not change the value of a the checkbox
    - reason: if clicked by accident (not expecting the behaviour like me) then it can be a lot of work to re-select the desired children
    - it may be another matter in case of a child (personally I think, even then, it should not change the checkbox' value (if within a tree))
- click on 'expand/collapse' icon should only open and close the 'group' (it can be finicky not to hit the margin/padding of the label)
    - reason: a user wants to explore the children

### Relevant Info 
see description

#### Which Component
Tree

benelan:
verified on `beta.74` using the doc sample with `selection-mode="ancestors"` https://developers.arcgis.com/calcite-design-system/components/tree/#sample

PR #3865, opened 2022-01-05
fix(tree-item): toggle behavior

Elijbet:
**Related Issue:** #3646

Adjusted toggle behavior:

- Click on the chevron icon only opens and closes the 'group’, does not select/deselect.
- Clicking on parent checkbox should only selects/deselects all items within the 'group' and not expand/collapse.
- Clicking the label of the checkbox, I think by default should do the same as checkbox, select/deselect.

https://user-images.githubusercontent.com/19231036/148296610-c796aeb1-c8f9-4c82-a598-e593f4b7a3f2.mov

- There is an existing test to make sure chevron icon expands/collapses, modified to ensure it doesn't select/deselect group.
- Added another one to confirm clicking on node-container, label, or checkbox selects/deselects, but does not expand/collapse children.


driskull (approved):
This looks good @Elijbet 🎉 Awesome

PR #3870, opened 2022-01-06
fix(input): restrict typing more than two hyphens

benelan:
**Related Issue:** #3053

## Summary
Restricts typing more than two hyphens, which is how native input does it.


driskull (approved):
👍

Issue #3686, opened 2021-12-10
Update relevant components to use ConditionalSlotComponent


benelan:
### Description 
Any components that conditionally use `getSlotted` should be updated to use the `ConditionalSlotComponent` interface.

### Proposed Advantages 
Fixes an issue where `getSlotted` doesn't know about elements slotted after initialization using JS, until a re-render occurs.

### Relevant Info 
The interface is defined in the  [`conditionalSlots.ts`](https://github.com/Esri/calcite-components/blob/master/src/utils/conditionalSlot.ts) util.

#### Which Component
All components that conditionally use `getSlotted`:

- [x] calcite-action-bar
- [x] calcite-action-group
- [x] calcite-action-menu
- [x] calcite-action-pad
- [x] calcite-block
- [x] calcite-chip
- [x] calcite-list-item
- [x] calcite-modal
- [x] calcite-notice
- [x] calcite-panel
- [x] calcite-pick-list-group
- [x] calcite-pick-list-item
- [x] calcite-shell
- [x] calcite-shell-center-row
- [x] calcite-shell-panel
- [x] calcite-tile
- [x] calcite-tip
- [x] calcite-tree-item
- [x] calcite-value-list-item

cc @driskull 

### Round II

Components using `this.el.querySelector` which should use `getSlotted()` and implement `ConditionalSlotComponent`

- [x] modal: Already using mutation observer
- [x] dropdown: Already using mutation observer
- [x] action-menu: needs testing.
- [x] calcite-alert
- [x] calcite-card
- [x] calcite-combobox-item
- [x] calcite-inline-editable: We should discuss deprecating this component and moving functionality.
- [x] calcite-input: Uses mutation observer & getSlotted
- [x] calcite-notice
- [x] calcite-radio-button-group: Uses mutation observer & getSlotted
- [x] calcite-radio-group
- [x] ~~calcite-radio-group-item~~ Nothing to change here.
- [x] ~~calcite-stepper-item~~ This component just needs to be refactored with a mutation observer and IE code removed.
- [x] ~~calcite-tree~~ This component just needs to be refactored with a mutation observer
- [x] ~~calcite-tree-item~~ This component just needs to be refactored with a mutation observer

driskull:
@benelan @jcfranco can we increase priority on this one and divvy up the work?

```
60 results - 19 files

src/components/calcite-action-bar/calcite-action-bar.tsx:
src/components/calcite-action-group/calcite-action-group.tsx:
[...]

PR #3871, opened 2022-01-06
fix: rerender components with conditional slots when slotting elements after initialization

jcfranco:
**Related Issue:** #3686 

## Summary

This PR updates components that were already using `getSlotted` for conditionally rendering their named slots to use the interface and utils from the `conditionalSlot.ts` util.

### Notable changes

1. Refactored the conditional slot helpers to ensure components rerender when elements are appended after initialization
2. Enhanced the `slots` test helper to help test conditional rendering
3. Refactored `calcite-panel`'s footer slot rendering to render both slots†

† **Reasoning for 3**:

* Both footer and footer-action slot rendering are identical except for the named slot (see https://github.com/Esri/calcite-components/blob/5b4793b7aae448e785fdaa06ad6824eb10cc6715/src/components/calcite-panel/calcite-panel.tsx#L387-L409)
* The usage of ☝️ (see https://github.com/Esri/calcite-components/blob/5b4793b7aae448e785fdaa06ad6824eb10cc6715/src/components/calcite-panel/calcite-panel.tsx#L459) indicates that it is invalid to slot both footer and footer-action elements, so we can render both and rely on the doc (sidebar: if both slots are identical, can we drop one in favor of the other?)
* It makes it easier to set up and use the test helper, otherwise we need a way to let the test helper know how slots are related to each other (I'd like to avoid this if possible)

### DIFF
--- a/src/components/calcite-panel/calcite-panel.tsx
+++ b/src/components/calcite-panel/calcite-panel.tsx
@@ -282,7 +282,7 @@ export class CalcitePanel implements ConditionalSlotComponent {
    */
   renderHeaderSlottedContent(): VNode {
     return (
-      <div class={CSS.headerContent} key="header-content">
+      <div class={CSS.headerContent} key="slotted-header-content">
         <slot name={SLOTS.headerContent} />
       </div>
     );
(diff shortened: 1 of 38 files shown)


driskull (approved):
Nice!

src/components/calcite-panel/calcite-panel.tsx
```diff
@@ -361,29 +381,16 @@ export class CalcitePanel {
...
-    const { el } = this;
-
     const hasFooterActions = getSlotted(el, SLOTS.footerActions);
 
-    return hasFooterActions ? (
-      <footer class={CSS.footer}>
-        <slot name={SLOTS.footerActions} />
+    return hasFooterSlottedContent || hasFooterActions ? (
+      <footer class={CSS.footer} key="footer">
+        {hasFooterSlottedContent ? <slot key="footer-slot" name={SLOTS.footer} /> : null}
```
asangma, line 392: Looks like this removes the option to use the "footer" slot.
Is that intended?

driskull (approved):
Looks good!

jcfranco:
Existing test failure is unrelated:

```ts
Summary of all failing tests
FAIL src/components/calcite-dropdown/calcite-dropdown.e2e.ts (69.154 s)
  ● calcite-dropdown › opens the dropdown with click, enter, or space › opens when dropdown-trigger is an action

    expected event "calciteDropdownOpen" to have been called 1 times, but was called 2 times

      828 |       await page.waitForChanges();
      829 |       expect(await dropdownWrapper.isVisible()).toBe(true);
    > 830 |       expect(calciteDropdownOpen).toHaveReceivedEventTimes(1);
          |                                   ^
      831 |     });
      832 |   });
      833 |

      at Object.<anonymous> (src/components/calcite-dropdown/calcite-dropdown.e2e.ts:830:35)
```

cc @Elijbet 

All other checks are ✅. Merging!

PR #3872, opened 2022-01-06
feat(inline-editable): change internal events to public

y0n4:
**Related Issue:** #3557  

## Summary
Removed `internal` tag for 2 existing events to be public. Previously known as `calciteInlineEditableChangesConfirm` and `calciteInlineEditableEditingCancel`


jcfranco (changes requested):
Looking good, @y0n4! Can you update the tests too?

jcfranco (approved):
Sorry for not bringing this up sooner, but looking at some other events, `calciteInlineEditableEdit<Action>` would be more consistent with naming.

Once the event is renamed, this is good to merge!

Issue #3875, opened 2022-01-07
A11Y: Radio buttons don't show checked state in high contrast mode


mpriour:
### Summary
High contrast mode on Windows renders webpages with forced colors and removes box-shadows and background colors. Calcite-component's radio buttons use box-shadows to indicate checked state. This doesn't work on high contrast and makes it impossible to determine which radio button is selected.

### Actual Behavior
<img width="434" alt="Screen Shot 2022-01-06 at 7 52 39 PM" src="https://user-images.githubusercontent.com/142636/148478515-7f3b9b40-87b6-4b6d-97f2-9cea73cf266d.png">

Layers is selected, but shows nothing.

### Expected Behavior

Something like this

<img width="299" alt="Screen Shot 2022-01-06 at 8 52 12 PM" src="https://user-images.githubusercontent.com/142636/148483830-cbcdf7fd-aacf-4b85-934a-5b6283184311.png">

### Reproduction Steps 

1. Open windows VM
2. Go to settings > High Contrast and enable
3. Visit https://developers.arcgis.com/calcite-design-system/components/radio-button-group/

### Relevant Info

All browsers on Windows. MacOS doesn't provide a true high contrast option. 

_Version_: `@esri/calcite-components@beta.73`

- [ ] CDN
- [X] NPM package

### Working W3C Example/Tutorial 
Possible to use `forced-colors` media query to add elements or styling specifically for this issue.
https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors#examples
This media query is supported on Chrome 89, Edge 79, and Firefox 89

mpriour:
Closely related to #2355

macandcheese:
It may be good to audit all of our components. I know you can provide user-defined high contrast colors in MS as well so could be something to test against. Cc @benelan

jcfranco:
Agreed. Created https://github.com/Esri/calcite-components/issues/3887 as a starting point. Feel free to modify it. cc @benelan

mpriour:
Radio buttons are visible in high contrast now. However, there is a slight but obvious overlap and misalignment of the selection indicator for small and medium scales. It needs to be scaled by the radio button size and not a set size for all scales.

PR #3876, opened 2022-01-07
fix(radio-button): show selection in high contrast (#3875)

mpriour:
* Use 'forced-colors' media query to detect high-contrast mode and draw checked radio state using ::after pseudo-element

**Related Issue:** #3875 

## Summary

Produces following UI on browsers with OS in high contrast mode.

Firefox 95
Screen Shot 2022-01-07 at 8 10 37 AM

Chrome 97
Screen Shot 2022-01-07 at 8 12 34 AM


benelan:
closing in favor of #3886

PR #3879, opened 2022-01-07
feat(input-date-picker): add calciteInputDatePickerChange event

jcfranco:
**Related Issue:** N/A

## Summary

Stems from https://github.com/Esri/calcite-components/issues/3659.

This introduces an `calciteInputDatePickerChange` event and deprecates the use of `calciteDatePickerChange` and `calciteDatePickerRangeChange` from that component since they should be considered internal.


driskull (approved):
👍

Issue #2138, opened 2021-05-13
Enhancement: Manage bottom margin on labels with inline layout


nwhittaker:
### Description
Given a `<calcite-label>` element with an `inline` layout, it is currently difficult to maintain a clean vertical rhythm. The vertical spacing between either itself and the next sibling, or between itself and the bottom of the parent element, becomes disjointed and inconsistent.

Given a grid layout with a vertical gap of `24px`:

```html
<form style="display: grid; row-gap: 24px; padding: 24px;">
  <calcite-label layout="inline">
    Label one
    <calcite-input />
  </calcite-label>
  <calcite-label layout="inline">
    Label two
    <calcite-input />
  </calcite-label>
</form>
```

Instead of the expected `24px` between `<calcite-label>` elements, it's `36px`. Same thing between the last `<calcite-label>` and the bottom of the parent's border-box.

---
There are several ways to workaround this problem in user-land:

#### Disjointed styling
One alternative to this specific example is to style the parent with `style="display: grid; row-gap: 12px; padding: 24px; padding-bottom: 12px;"`. However, it obfuscates what the real vertical rhythm is, and could become incorrect if the label's default margin changes.

#### Set `disable-spacing`
Another alternative is setting `disable-spacing`. This does address the immediate problem, but at the expense of losing the horizontal spacing between the label and the input. To address this, the docs cite a workaround:

```html
<calcite-label layout="inline" disable-spacing>
  <span style="margin-right: 0.75rem">Text leading inline</span>
  <calcite-checkbox></calcite-checkbox>
</calcite-label>
```

However, this has a few problems:
1. The length of the horizontal space could fall out of date w/ the default Calcite value.
2. The space will be on the wrong side in RTL layouts.
3. Adds verbosity to the code, uses an inline style, and is more to maintain.

#### Override `label` styles
Add styles to the user-land stylesheet to override the `label.sc-calcite-label` element's bottom-margin. Necessitating this workaround is an indicator that the `<calcite-label>` element is perhaps too opinionated in terms of what layouts it fits into. It's also at odds with the trend towards styling via Tailwind.

---
Hopefully the above illustrates how the component's current implementation poses a dilemma for the consuming app:
1. Choose clean mark-up at the expense of a consistent vertical rhythm;
2. Or choose a consistent vertical rhythm at the expense of clean mark-up.

This dilemma stems from the component assuming its bottom margin is always needed.

### Acceptance Criteria 
[...]

PR #3888, opened 2022-01-10
feat(label): add new custom prop for label w/o default margin spacing…

y0n4:
**Related Issue:** #2138 

## Summary
There is a custom prop `disable-vertical-spacing` which will remove the margin space below any label component so that user has more free will to style. This prop is used as a css specific variable class

Before
Screen Shot 2022-01-10 at 10 20 56 AM

After
Screen Shot 2022-01-10 at 10 19 28 AM

It can now be used like this
```
<form style="display: grid; row-gap: 24px; padding: 24px">
  <calcite-label layout="inline" disable-vertical-spacing>
    Label one
    <calcite-input />
  </calcite-label>
  <calcite-label layout="inline" disable-vertical-spacing>
    Label two
    <calcite-input />
  </calcite-label>
</form>
```


driskull:
I think we need to resolve this one first: https://github.com/Esri/calcite-components/issues/3860. I don't think we want to add another property to disable spacing.

PR #3889, opened 2022-01-10
refactor: Update relevant components to handle slotting after init

driskull:
**Related Issue:** #3686

## Summary

refactor: Update relevant components to use ConditionalSlotComponent

Also spawned: https://github.com/Esri/calcite-components/issues/3881


src/utils/dom.ts
```diff
@@ -159,10 +159,11 @@ export async function focusElement(el: CalciteFocusableElement): Promise<void> {
 interface GetSlottedOptions {
   all?: boolean;
   direct?: boolean;
+  matches?: string;
```
driskull, line 162: @jcfranco what do you think? Should we have a way to do `getSlotted(this.el, { matches: "calcite-link"});`

src/utils/dom.ts
```diff
@@ -159,10 +159,11 @@ export async function focusElement(el: CalciteFocusableElement): Promise<void> {
...
   selector?: string;
 }
 
-const defaultSlotSelector = "> :not([slot])";
+const defaultSlotSelector = ":not([slot])";
```
driskull, line 166: I'm not sure why I had the `> ` but this might be something we need??

src/components/calcite-action-menu/calcite-action-menu.tsx
```diff
@@ -363,15 +363,7 @@ export class CalciteActionMenu {
...
-    const assignedActions = this.getAssignedElements().filter(
-      (element) => element.tagName === "CALCITE-ACTION" && element.slot !== "trigger"
-    ) as HTMLCalciteActionElement[];
-
-    const actionElements = assignedActions.length
-      ? assignedActions
-      : (Array.from(
-          el.querySelectorAll("calcite-action:not([slot=trigger])")
-        ) as HTMLCalciteActionElement[]);
+    const actionElements = getSlotted(el, { all: true, matches: "calcite-action" }) as any;
```
driskull, line 366: Need to test this cleanup...

src/components/calcite-alert/calcite-alert.tsx
```diff
@@ -121,10 +121,6 @@ export class CalciteAlert {
     this.requestedIcon = setRequestedIcon(StatusIcons, this.icon, this.color);
   }
 
-  componentDidLoad(): void {
-    this.alertLinkEl = this.el.querySelectorAll("calcite-link")[0] as HTMLCalciteLinkElement;
```
driskull, line 125: moved to query on focus instead of load incase the element is added later.

src/components/calcite-inline-editable/calcite-inline-editable.tsx
```diff
@@ -92,18 +97,15 @@ export class CalciteInlineEditable implements LabelableComponent {
...
+    this.scale =
+      this.scale || this.inputElement?.scale || getElementProp(this.el, "scale", undefined);
   }
 
   disconnectedCallback() {
     disconnectLabel(this);
-  }
-
-  componentWillLoad() {
-    this.inputElement = this.el.querySelector("calcite-input") as HTMLCalciteInputElement;
```
driskull, line 102: moved to observer so that the input doesn't need to be added before init.

src/components/calcite-input/calcite-input.tsx
```diff
@@ -309,11 +312,6 @@ export class CalciteInput implements LabelableComponent, FormComponent {
     this.requestedIcon = setRequestedIcon(INPUT_TYPE_ICONS, this.icon, this.type);
   }
 
-  componentDidLoad(): void {
-    this.slottedActionEl = this.el.querySelector("[slot=action]");
-    this.setDisabledAction();
```
driskull, line 314: Moved to observer so slot isn't needed on init.

src/components/calcite-notice/calcite-notice.tsx
```diff
@@ -88,10 +88,6 @@ export class CalciteNotice {
     this.requestedIcon = setRequestedIcon(StatusIcons, this.icon, this.color);
   }
 
-  componentDidLoad(): void {
-    this.noticeLinkEl = this.el.querySelector("calcite-link") as HTMLCalciteLinkElement;
```
driskull, line 92: Moved to focus method so slot isn't needed on init.

src/components/calcite-radio-button-group/calcite-radio-button-group.tsx
```diff
@@ -73,6 +74,14 @@ export class CalciteRadioButtonGroup {
     this.passPropsToRadioButtons();
   }
 
+  // --------------------------------------------------------------------------
+  //
+  //  Private Properties
+  //
+  // --------------------------------------------------------------------------
+
+  mutationObserver = createObserver("mutation", () => this.passPropsToRadioButtons());
```
driskull, line 83: `passPropsToRadioButtons` was only being called on connectedCallback.

driskull (commented):
@jcfranco I think this needs your mutation observer fixes for conditionalSlot. Some getSlotted are not working correctly. I'll wait for that to get in.

jcfranco:
#3871 just landed, so you the util should work more reliably and you now have access to the test helper (for named slots) if needed.

src/utils/dom.ts
jcfranco: I think `selector` should work here. Let me know if it doesn't _match_ your needs (badumtss🥁) .

driskull: @jcfranco selector queries the element for children, matches tests the element.

driskull: the matches is more of a filter.

src/utils/dom.ts
driskull: Ok, so can't do direct selector because of the scope. Instead, have to filter child nodes first.

I tried using `:scope > ` but that wasn't working either.
https://developer.mozilla.org/en-US/docs/Web/CSS/:scope

src/utils/dom.ts
jcfranco: Gotcha. Looks like you found a hole in this util. 🍩  I think queries using `selector` should include the slotted element. WDYT? Fortunately, there are only 2 selector usages so far:

1. https://github.com/Esri/calcite-components/blob/2ec984876f33939d06c7ad0b47c9040f1a7d0a9d/src/components/calcite-tree-item/calcite-tree-item.tsx#L414
2. https://github.com/Esri/calcite-components/blob/5e0f9f259d160f7826dbf1be56b05ad52b0d0add/src/components/calcite-alert/calcite-alert.tsx#L241

I'll create an issue for the above and can tackle this separately.

src/utils/dom.ts
jcfranco: Interesting. Do the current spec tests cover this? Wonder why it didn't pick this up.

src/components/calcite-combobox-item/calcite-combobox-item.tsx
```diff
@@ -171,14 +175,11 @@ export class CalciteComboboxItem {
...
+    const defaultSlotNode = <slot />;
+    if (!getSlotted(this.el)) {
+      return defaultSlotNode;
     }
-    return (
-      <ul>
-        <slot />
-      </ul>
-    );
+    return <ul>{defaultSlotNode}</ul>;
```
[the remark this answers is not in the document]
driskull: No, I just did this for clarity so its apparent that its the same VNode being used.

src/components/calcite-combobox/calcite-combobox.e2e.ts
```diff
@@ -20,15 +20,15 @@ describe("calcite-combobox", () => {
   it("is accessible", async () =>
     accessible(`
       <calcite-combobox label="Trees" value="Trees">
-        <calcite-combobox-item value="Pine" text-label="Pine"></calcite-combobox-item>
+        <calcite-combobox-item guid="i1" value="Pine" text-label="Pine"></calcite-combobox-item>
```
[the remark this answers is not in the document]
driskull: It should, but it wasn't doing that in the tests. I don't think the `<calcite-combobox-item>` is getting constructed in the test and so the guid was not there. Maybe it needs waitForChanges or something? @jcfranco

src/utils/dom.ts
driskull: Yeah we can either update how selector works or matches.

src/utils/dom.ts
driskull: > Do the current spec tests cover this? Wonder why it didn't pick this up.

We weren't using a selector like this for all default slot queries.

src/components/calcite-radio-button/calcite-radio-button.tsx
```diff
@@ -89,12 +89,6 @@ export class CalciteRadioButton implements LabelableComponent, CheckableFormComp
   @Watch("name")
   nameChanged(): void {
     this.checkLastRadioButton();
-    const currentValue: HTMLInputElement = this.rootNode.querySelector(
-      `input[name="${this.name}"]:checked`
-    );
-    if (!currentValue?.value) {
-      this.uncheckAllRadioButtonsInGroup();
```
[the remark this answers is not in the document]
driskull: It seems to work fine for me. Can you confirm @eriklharper?

src/components/calcite-action-menu/calcite-action-menu.tsx
driskull: I think this issue needs to happen first.... https://github.com/Esri/calcite-components/issues/3881

driskull: > Test successful?

Yes this seems to be working.

src/components/calcite-combobox-item/calcite-combobox-item.tsx
```diff
@@ -10,12 +10,17 @@ import {
...
 } from "@stencil/core";
-import { getElementProp } from "../../utils/dom";
+import { getElementProp, getSlotted } from "../../utils/dom";
 import { CSS } from "./resources";
 import { guid } from "../../utils/guid";
 import { ComboboxChildElement } from "../calcite-combobox/interfaces";
 import { getAncestors, getDepth } from "../calcite-combobox/utils";
 import { Scale } from "../interfaces";
+import {
+  connectConditionalSlotComponent,
```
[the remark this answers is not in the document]
driskull: Split it up! 🍌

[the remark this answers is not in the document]
driskull: https://github.com/Esri/calcite-components/pull/3959

src/components/calcite-radio-button/calcite-radio-button.tsx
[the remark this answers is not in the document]
eriklharper: Sure thing, I'll try that locally real quick.  If there's not a test yet for that I'll add one.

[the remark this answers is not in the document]
eriklharper: Just out of curiosity, why is this being removed?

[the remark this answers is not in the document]
driskull: It was failing tests when left in with the changes herre.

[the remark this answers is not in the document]
eriklharper: This should be fine to remove.  I'm blanking on the original purpose of this code, but it seems like we don't need to do this.  Just because the name changes doesn't mean we have to change the checked state of the group.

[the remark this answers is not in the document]
eriklharper: LGTM! 👍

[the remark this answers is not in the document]
eriklharper: Cool, yeah I don't see a problem removing this.

driskull:
@jcfranco anything else I need to do to get this one merged?

src/components/calcite-radio-button-group/calcite-radio-button-group.tsx
```diff
@@ -81,6 +90,11 @@ export class CalciteRadioButtonGroup {
 
   connectedCallback(): void {
     this.passPropsToRadioButtons();
+    this.mutationObserver?.observe(this.el, { childList: true, subtree: true });
```
jcfranco, line 93: Do you need to watch for changes to radio-button children? I don't think it supports this, right?

jcfranco (approved):
Just had one suggestion, but this LGTM! 🎉 

Anything else that came up will be tackled separately.

src/components/calcite-radio-button-group/calcite-radio-button-group.tsx
driskull: @jcfranco I don't think so. The props should only go down the dom right?

jcfranco: > The props should only go down the dom right?

If I understand correctly, only one level, yes.

Issue #3878, opened 2022-01-07
Number input: Value changes by itself indefinitely


jmanke:
### Actual Behavior

In certain situations the `<calcite-input type='number'>` will auto increment/decrement by itself indefinitely.

![chrome-capture](https://user-images.githubusercontent.com/19629809/148585073-df518eb7-6b09-4feb-ab2b-b83c7b02a8fd.gif)

### Expected Behavior

The number input should not change its value in unexpected ways.

### Reproduction Steps and Sample

_Sample_: [codepen](https://codepen.io/jmanke/pen/mdBjOEY?editors=1010)

1. Press the 'Up Arrow' and 'Down Arrow' keys in quick succession (or at the same time). If that doesn't work, try alternating between the 'Up Arrow' and 'Down Arrow' keys very quickly.
2. Observe the number input starts incrementing or decrementing the value by itself

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-beta.73`

- [x] CDN
- [ ] NPM package

jcfranco:
Yikes, this is a gnarly one. Thanks for reporting this.

This regression got introduced in beta 69. Probably from https://github.com/Esri/calcite-components/pull/3340. 

@Elijbet Do you think you can take a look at this one for this sprint?

Elijbet:
> Yikes, this is a gnarly one. Thanks for reporting this.
> 
> This regression got introduced in beta 69. Probably from #3340.
> 
> @Elijbet Do you think you can take a look at this one for this sprint?

Yeah, I'll look at it. Sorry, missed that somehow.

benelan:
verified on beta.75

PR #3908, opened 2022-01-11
fix(input): number input no longer increments by itself indefinitely when up/down buttons are toggled fast or pressed at the same time;

Elijbet:
**Related Issue:** #3878

## Summary

It appears that if you toggle up/down fast, the first interval ID doesn't clear. This means if you click up and then down in sequence fast, it'll keep incrementing by itself indefinitely. Now the previous is cleared, if present, before new one is started. 

This is an attempt to demo keyboard interaction with the number input. 

https://user-images.githubusercontent.com/19231036/149009988-b4981c53-c92c-41d2-9bcc-f80db0b9ee2f.mov

UPDATE: Initially disabled incrementing altogether when up/down are pressed at the same time. Changed to follow the native input's behavior where the most recently pressed key takes over.


jcfranco (changes requested):
Nice starting point. Can you update this to follow the native input's behavior where the most recently pressed key takes over? It looks like the current version will not nudge at all when both keys are pressed. https://codepen.io/jcfranco/pen/rNGQdYp

jcfranco (approved):
👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍
👍✅👍👍👍✅👍👍✅✅👍👍✅✅✅👍👍✅✅✅👍✅✅✅✅👍✅👍
👍✅✅👍👍✅👍✅👍👍✅👍👍✅👍👍✅👍👍👍👍✅👍👍👍👍✅👍
👍✅👍✅👍✅👍✅👍👍✅👍👍✅👍👍✅👍👍👍👍✅✅✅👍👍✅👍
👍✅👍👍✅✅👍✅👍👍✅👍👍✅👍👍✅👍👍👍👍✅👍👍👍👍👍👍
👍✅👍👍👍✅👍👍✅✅👍👍✅✅✅👍👍✅✅✅👍✅✅✅✅👍✅👍
👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍

Issue #3420, opened 2021-11-03
calcite-tree interferes with normal tab flow in Firefox


MikeTschudi:
### Summary

When one tabs into a calcite-tree in Firefox, tabbing out--forwards or backwards--jumps to unexpected locations.

### Actual Behavior

One can tab into a calcite-tree, but shift-tabbing back out doesn't go to previous element. Tabbing out the bottom doesn't go to the next element.

### Expected Behavior

Expect to tab into and through a calcite-tree and back again

### Reproduction Steps and Sample

https://codepen.io/miketschudi/pen/wvqpXRq

### Relevant Info

OK in Chrome 95
problem in Firefox 93

_Version_: `@esri/calcite-components@1.0.0-beta.68`

- [x] CDN
- [ ] NPM package

jcfranco:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/calcite-5f7d10460071bf0024e66277/issues/Esri/calcite-components/3420?planning-poker) with ZenHub @caripizza

caripizza:
@MikeTschudi are on Mac, and does this happen with other components in Firefox? It almost looks to me like this issue instead: https://css-tricks.com/proper-tabbing-to-interactive-elements-in-firefox-on-macos/

MikeTschudi:
@caripizza, it's on Windows 10 and we've only seen tabbing into/out of issues with calcite-tree. Backwards nav used to be a problem in Chrome, but that was fixed via #683.

y0n4:
Hi could you provide an example of an expected order of a tab key forward/backward and shift+tab sequence for this code sample? It could be an instruction by text or by video. Thanks!

MikeTschudi:
@y0n4, I modified the https://codepen.io/miketschudi/pen/wvqpXRq to make visiting the neighbor before and after more obvious, and I switched it from *beta.68 to *beta.74.

Open the pen in Chrome or Chromium Edge:
1. Start the pen
2. Click in the demo area on "Neighbor before"
3. Tab: you enter list with "Child 1"
4. Keep tabbing past "Child 3": you land on "Neighbor after"
5. Shift-tab: you re-enter list with "Child 3"
6. Keep shift-tabbing past "Child 1": you land on "Neighbor before"
7. Repeat steps 3 through 6

[...]

PR #3914, opened 2022-01-12
fix(tree): change tab index when tabbing in/out for first element

y0n4:
**Related Issue:** #3420

## Summary
Previously was unable to tab in/out of the `calcite-tree` to other elements specifically in Firefox browsers. Now tabbing out by `tab` key or `shift + tab` key will jump to other elements in proper sequence. This change does not affect in Chrome browsers and can still tab like the way it did previously


jcfranco (approved):
For maintainability, can you add a comment about why the attribute is modified instead of the property?

src/components/calcite-tree/calcite-tree.tsx
```diff
@@ -108,7 +108,7 @@ export class CalciteTree {
       event.relatedTarget === this.el || !this.el.contains(event.relatedTarget as HTMLElement);
 
     if (focusedFromRootOrOutsideTree) {
-      this.el.tabIndex = -1;
+      this.el.removeAttribute("tabindex");
```
jcfranco, line 111: For maintainability, can you add a comment about why the attribute is modified instead of the property?

Issue #3754, opened 2021-12-21
[Combobox] - Internally rendered input has incorrect id assigned


macandcheese:
### Actual Behavior
`calcite-combobox` is assigning the `guid` *function* as a string, not the result of the function, to the internal input.

<img width="1528" alt="Screen Shot 2021-12-21 at 11 27 47 AM" src="https://user-images.githubusercontent.com/4733155/146986703-0a5db498-cb88-4f7a-927d-279cd9f578e7.png">

https://github.com/Esri/calcite-components/blob/c24ae4e242ab6da7a516b96ea29554ec8c411b33/src/components/calcite-combobox/calcite-combobox.tsx#L950

### Expected Behavior
A guid is correctly assigned and associated with label. Perhaps this could use a calcite-input or calcite-label?

### Reproduction Steps and Sample
Inspect shadow DOM of combobox

_Sample_:

1.

### Relevant Info

_Reproducible version_: `71`

https://developers.arcgis.com/calcite-design-system/components/combobox/

- [ ] CDN
- [ ] NPM package

benelan:
verified in the documentation sample

PR #3919, opened 2022-01-12
fix(combobox): Properly set internal identifiers. #3754

driskull:
**Related Issue:** #3754

## Summary

fix(combobox): Properly set internal identifiers. #3754

### DIFF
--- a/src/components/calcite-combobox/calcite-combobox.tsx
+++ b/src/components/calcite-combobox/calcite-combobox.tsx
@@ -49,6 +49,12 @@ interface ItemData {
 const isGroup = (el: ComboboxChildElement): el is HTMLCalciteComboboxItemGroupElement =>
   el.tagName === ComboboxItemGroup;
 
+const itemUidPrefix = "combobox-item-";
+const chipUidPrefix = "combobox-chip-";
+const labelUidPrefix = "combobox-label-";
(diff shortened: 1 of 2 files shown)


src/components/calcite-combobox/calcite-combobox.tsx
```diff
@@ -891,12 +897,11 @@ export class CalciteCombobox implements LabelableComponent, FormComponent {
       const label = selectionMode !== "ancestors" ? item.textLabel : pathLabel.join(" / ");
       return (
         <calcite-chip
-          aria-label={label}
           class={chipClasses}
           dismissLabel={"remove tag"}
```
driskull, line 901: @jcfranco @benelan Need an issue for this string. Do we need/want a issue template for localization? Or would this be categorized as a bug? or feat cus it needs a new prop?

benelan: I'd call that a bug since it is a hardcoded string. And then just link the doc site for the repro sample. I don't think we need a new issue template for localization, we can just use the existing labels. I'll make the issue for this.

driskull: I think its more of a feature. Nothing is broken, we are just adding the ability to set it via a property.

PR #3921, opened 2022-01-12
chore(template): remove value from enhancement field

benelan:
**Related Issue:** NA

## Summary
An unneeded `value` property was left in during my copy/paste rampage


driskull (approved):
👍

Issue #3758, opened 2021-12-21
Add support for `download` attribute to `calcite-link`


nwhittaker:
### Description
Native anchor elements provide a `download` attribute for prompting a resource/data download. Since `<calcite-link>` wraps an `<a>` element, it'd be nice to expose this part of the native API.

### Acceptance Criteria 
A `download` attribute on the `<calcite-link>` element is passed through to the underlying `<a>` element.

### Relevant Info 
[`download` attribute doc][1]

[1]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-download

#### Which Component
`CalciteLink`

#### Example Use Case
```html
<calcite-link download href="data:...">
  Download QR code
</calcite-link>
```

macandcheese:
We may want to add this for `calcite-button` - when used with a `href` that renders an `<a>` tag internally.

PR #3922, opened 2022-01-12
feat(calcite-link): Add support for download attribute. #3758

driskull:
**Related Issue:** #3758

## Summary

feat(calcite-link): Add support for download attribute. #3758

### DIFF
--- a/src/components/calcite-link/calcite-link.e2e.ts
+++ b/src/components/calcite-link/calcite-link.e2e.ts
@@ -33,7 +33,7 @@ describe("calcite-link", () => {
     element.setProperty("download", true);
     await page.waitForChanges();
 
-    expect(await elementAsLink.getProperty("download")).toBe(true);
+    expect(await elementAsLink.getProperty("download")).toBe("");
     expect(elementAsLink).toHaveAttribute("download");
     expect(elementAsLink.getAttribute("download")).toBe("");
 
--- a/src/components/calcite-link/calcite-link.e2e.ts
+++ b/src/components/calcite-link/calcite-link.e2e.ts
@@ -29,11 +29,11 @@ describe("calcite-link", () => {
 
     const element = await page.find("calcite-link");
 
-    element.setAttribute("download", true);
+    element.setAttribute("download", "");
     await page.waitForChanges();
 
     expect(elementAsLink).toHaveAttribute("download");
-    expect(elementAsLink.getAttribute("download")).toBe("true");
+    expect(elementAsLink.getAttribute("download")).toBe("");
 
     const newFilename = "my-cool-file.jpg";
     element.setAttribute("download", newFilename);
--- a/src/components/calcite-link/calcite-link.e2e.ts
+++ b/src/components/calcite-link/calcite-link.e2e.ts
@@ -29,14 +29,14 @@ describe("calcite-link", () => {
 
     const element = await page.find("calcite-link");
 
-    element.setAttribute("download", "");
+    element.setProperty("download", true);
     await page.waitForChanges();
 
     expect(elementAsLink).toHaveAttribute("download");
     expect(elementAsLink.getAttribute("download")).toBe("");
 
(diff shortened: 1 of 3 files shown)


src/components/calcite-link/calcite-link.tsx
```diff
@@ -93,6 +100,7 @@ export class CalciteLink {
       <Host role="presentation">
         <Tag
           class={{ [CSS_UTILITY.rtl]: dir === "rtl" }}
+          download={Tag === "a" && (download === "" || download) ? download : null}
```
driskull, line 103: @jcfranco @eriklharper have you run into the following before...

- If a property is `boolean` or `string` and the property is set to `true`, the value is `""`?
- I had to work around this, I'm not sure if its as intended or stencil issue is needed?
- I had expected the value of `download` to be `true` if set to `true` via a property but it was a `""` empty string instead.
- Basically, I have to treat the value of `""` as `true` which seems counterintuitive and wrong 🤷

driskull: When setting via a prop we also get this lovely warning.

```
$0.download = true;
true
index-07dce265.js:2915 stencil The state/prop "download" changed during rendering. This can potentially lead to infinite-loops and other bugs. 
Element <calcite-link href=​"calcite-link.html" role=​"presentation" calcite-hydrated download>​…​</calcite-link>​ 
New value  
Old value true
```

I guess stencil just doesn't like mixed boolean string types?

src/components/calcite-link/calcite-link.e2e.ts
```diff
@@ -1,15 +1,48 @@
...
   });
 
+  it("sets download attribute on internal anchor", async () => {
+    const page = await newE2EPage();
+    await page.setContent(`<calcite-link href="file.jpg">Continue</calcite-link>`);
+
+    const elementAsLink = await page.find("calcite-link >>> a");
+
+    expect(elementAsLink).not.toBeNull();
+    expect(elementAsLink).not.toHaveAttribute("download");
```
jcfranco, line 28: For consistency, can these assertions use props vs attributes?

src/components/calcite-link/calcite-link.tsx
jcfranco: Wow, interesting find. Can you add a comment there for posterity? 

I'll create a repro case and send it over to the Stencil team.

jcfranco (approved):
💥

src/components/calcite-link/calcite-link.e2e.ts
driskull: Yes, but if I use the prop it will emit that console warning.

```
The state/prop "download" changed during rendering. This can potentially lead to infinite-loops and other bugs. 
```

driskull:
@jcfranco here's a repro case: https://github.com/driskull/stencil-component-starter/tree/boolean-and-string-property

```tsx
import { Component, Prop, h } from "@stencil/core";

@Component({
  tag: "my-component",
  styleUrl: "my-component.css",
  shadow: true,
})
export class MyComponent {
  /** Prompts the user to save the linked URL instead of navigating to it. Can be used with or without a value:
   * Without a value, the browser will suggest a filename/extension
   * See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-download
   */
  @Prop({ reflect: true }) download: string | boolean = false;

  /**
   * text
   */
  @Prop() text: string;

  render() {
    const { download, text } = this;

    console.log({ download, text });

    return (
      <div>
        <a
          // Why is a boolean value of true converted to ""?
          download={download === "" || download ? download : null}
          href="/"
        >
          {text}
        </a>
        <span> | value: {JSON.stringify(download)}</span>
      </div>
    );
  }
}

```

```html
<body>
    <my-component text="1: default"></my-component>

    <my-component download text="2: download set without value"></my-component>

    <my-component
      download="my-file.html"
      text="3: download value set"
    ></my-component>
  </body>
```

Issue #3196, opened 2021-10-11
Enhancement: placement for large popovers on top of shell-panels


AdelheidF:
In the Map Viewer we often display larger popovers on top of the right or left shell-panel, e.g. when users have to select a field. These popovers cover about half the vertical size of the panel and the reference element, e.g. button, that opens this popover can be anywhere on the panel. 

The problem we encounter is that the popup is not always placed well and might even have its top cut off. Is there a way to have a fixed placement for the popover over a panel?

<img src=https://user-images.githubusercontent.com/44481292/136856265-e199da2d-32a5-4d24-badb-3cfb0e3b19bd.png width=250/>![image](https://user-images.githubusercontent.com/44481292/136859284-b4dd4e20-adba-4412-949e-c1565e727204.png)

One idea was to add a slot for the popover and have it always use the same internal reference element for placement.

benelan:
Hey @AdelheidF, do you have a sample that reproduces this issue?

AdelheidF:
Could the panel header node be exposed somehow? Then we could anchor the popover below the header.

AdelheidF:
We get several bug reports about popovers being partially hidden in several of our components all due to calcite-panel or its content not having a good reference element that doesn't scroll out of view or scroll into positions where popper places them weird.

Could we give this issue a higher priority, please?

benelan:
Moving this to the next sprint starting 12/20

driskull:
It looks like the `calcite-shell-panel` has a slot named `header`. Can you throw a div in that and use that div as the reference element for the popover?

I don't think this slot is documented but I'm not sure why not? @asangma?

AdelheidF:
> It looks like the calcite-shell-panel has a slot named header. Can you throw a div in that and use that div as the reference element for the popover?

Most of our app components are just calcite-panels that the app places into the shell-panel. This would require all apps to create that header div and pass a reference of it to the app components. This works but is a bit complicated. Could each panel have that kind of slot/header instead?

driskull:
@AdelheidF so if I understand correctly...

- You need a way to attach a popover to the header of a panel
- So that the popover doesn't scroll with the content of a panel
- Attaching to the panel itself won't work because the target would be too large.

So we can either...

- Create a slot for supplying a reference element
  -  Not ideal because it wouldn't be anything visible.
- Create a read only property that is the header of the panel for use as a reference element.
  - Tricky because I don't think stencil allows readOnly props.
[...]

PR #3926, opened 2022-01-13
feat(panel): Allow slotting popover to attach to the header. #3196

driskull:
**Related Issue:** #3196

## Summary

feat(panel): Allow slotting popover to attach to the header. #3196


src/components/calcite-panel/calcite-panel.scss
```diff
@@ -10,7 +10,7 @@
 
 :host {
   @extend %component-host;
-  @apply flex flex-auto overflow-hidden relative w-full;
```
driskull, line 13: cant cut off popover

src/components/calcite-panel/calcite-panel.scss
```diff
@@ -22,7 +22,8 @@
 @import "../../assets/styles/header";
 
 .container {
-  @apply flex flex-auto flex-col items-stretch bg-background w-full p-0 m-0;
+  @apply flex flex-auto flex-col items-stretch bg-background w-full p-0 m-0 relative;
+  z-index: 1;
```
driskull, line 26: contains z-index of header.

driskull:
@jcfranco can you review this one?

jcfranco:
Apologies for the delayed review and thanks for taking a stab at this. This works for this specific use case, but I do wonder about the scalability and how it affects our other components.

I'd like to propose revisiting and to pivot towards improving the popover placement, so that it isn't cut off as described in the issue description.

Here are some reasons why I think we should reconsider:

* Popover placement is entirely dynamic, so even if we support the header as a reference element, there may still be inconsistent/undesirable placement.
* This would set precedent for adding slots for internal parts of components – this adds complexity to components that could be handled elsewhere.

Some ideas to try:

* Adding a way for users to provide modifiers/middleware to fine-tune positioning (we need to keep in mind the future move to `floating-ui` and how it could impact this).
* Introduce a custom modifier and component pattern for smarter placement (would need to sync with the design team to define this).

driskull:
Sounds good. I'll close the PR

Issue #3577, opened 2021-11-29
calcite-action alignment=center issues when inside calcite-action-group with layout=grid


AdelheidF:
With beta.66

![image](https://user-images.githubusercontent.com/44481292/143918389-f79f2215-fa5d-4f65-a022-c86a1d773fd2.png)

with beta.67

![image](https://user-images.githubusercontent.com/44481292/143918476-225c6f13-d158-4d61-b3e0-925ff0b673ad.png)

In the release notes for beta.67 I see this (@eriklharper):

![image](https://user-images.githubusercontent.com/44481292/143918612-db4ca3d2-ac66-4fee-94d2-5f74e58380d0.png)

### Actual Behavior

alignment=center doesn't seem to work anymore

### Expected Behavior

actions being centered in grid

### Reproduction Steps and Sample

https://codepen.io/afreitag/pen/WNZNGdo?editors=1010

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-beta.67`

- [ ] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-beta.66`

AdelheidF:
Is there a workaround for this? It's part of a new component in the map viewer.

jcfranco:
Here's a workaround in the meantime: https://codepen.io/jcfranco/pen/xxXxqGY?editors=1000

jcfranco:
@eriklharper Can you look at this for the upcoming sprint?

benelan:
verified on beta.75

PR #3929, opened 2022-01-13
fix(action, action-group): alignment center when enclosed within group (grid layout, text-enabled vs no text)

Elijbet:
**Related Issue: #3577**

Specified flex-grow on the button to fill the container, added test and demos.


driskull (approved):
👍

asangma:
This is looking good. It appears though that the screenshots that are **not** `text-enabled` and `indicator` are looking off. Do you know if those screener tests are setting a width on the container?

asangma (approved):
Whoop!

PR #3930, opened 2022-01-13
ci(screener): updating to Chrome latest

eriklharper:
**Related Issue:** n/a

### DIFF
--- a/screener.config.js
+++ b/screener.config.js
@@ -9,7 +9,7 @@ module.exports = {
   browsers: [
     {
       browserName: "chrome",
-      version: "latest"
+      version: "97"
     }
   ],
   diffOptions: {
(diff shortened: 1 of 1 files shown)


driskull (approved):

driskull:
We may want to bump this up even more. I'm not sure.

We could setup something to maybe automate bumping it every so often so we keep up with our supported browsers

benelan:
> We may want to bump this up even more. I'm not sure.
> 
> We could setup something to maybe automate bumping it every so often so we keep up with our supported browsers

We could scrape the latest version here https://chromedriver.storage.googleapis.com/LATEST_RELEASE

eriklharper:
Closing in favor of #3940

PR #3934, opened 2022-01-13
docs(readme): remove stencil webpack plugin

benelan:
**Related Issue:** NA

## Summary
Removing stencil's webpack plugin from our doc since it doesn't work in v5. @jcfranco put in a PR but the repo hasn't been touched in 4 years https://github.com/ionic-team/stencil-webpack/pull/9

PR #3935, opened 2022-01-13
chore(action): install cached packages on all check runs

benelan:
**Related Issue:** #3934

## Summary
The e2e and screener checks are set up to skip installing deps and running the tests on markdown-only PRs. I then ended up caching the npm deps for all actions. Apparently if you cache the deps and the action run doesn't install any, it throws an error on clean up: https://github.com/actions/setup-node/issues/317#issuecomment-912336228

- I moved the install outside of the conditional. 
- Performance on all PRs that have non-markdown changes won't change
- Markdown only PRs shouldn't be too bad since the packages are cached

### DIFF
--- a/.github/workflows/pr-e2e.yml
+++ b/.github/workflows/pr-e2e.yml
@@ -27,7 +27,6 @@ jobs:
         with:
           node-version: lts/*
           cache: npm
+      - run: npm install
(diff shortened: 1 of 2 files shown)


jcfranco (approved):
Nice sleuthing! 🔍:shipit:

Issue #1225, opened 2020-11-04
Bug: Accordion doesn't correctly apply transparent prop in FireFox


evanmosby:
### Summary
In FireFox, an Accordion component set to `appearance: transparent` appears white

### Actual Behavior
Component background is white

### Expected Behavior
Component background should be transparent

### Reproduction Steps

1. go to https://esri.github.io/calcite-components/?path=/story/components-accordion--simple
2. Set appearance to transparent
3. Observe background is still white

### Relevant Info 

Latest Firefox

macandcheese:
Ah, looking at the css, there are styles set with `host-context` that are based on the parent `calcite-accordion`'s props - so we will also see this issue with the `appearance=minimal` setting, and in IE and Safari as well (I think they both also have trouble with `host-context`). Also might need to check out setting CSS `calcite-accordion` on the parent that pass through.

Looks like we'll need to grab those props and set them as attributes on the child `calcite-accordion-item` and then style that way.

macandcheese:
Valid

benelan:
verified using the doc sample https://developers.arcgis.com/calcite-design-system/components/accordion/#sample

PR #3954, opened 2022-01-18
fix(accordion): no longer has white background when set to transparent

anveshmekala:
**Related Issue:** #1225 

## Summary

This fix will set the `background-color` of the `calcite-accordion` to `transparent` in FireFox & Safari.

### DIFF
--- a/src/components/calcite-accordion-item/calcite-accordion-item.scss
+++ b/src/components/calcite-accordion-item/calcite-accordion-item.scss
@@ -29,26 +29,13 @@
 }
 
 :host {
-  --calcite-accordion-item-background: theme("backgroundColor.foreground.1");
-}
-
-:host([appearance="minimal"]) {
-  --calcite-accordion-item-padding: var(--calcite-accordion-item-spacing-unit) 0;
-}
-:host([appearance="transparent"]) {
-  --calcite-accordion-item-border: transparent;
-  --calcite-accordion-item-background: transparent;
-}
-
-:host {
   @apply flex
     flex-col
     no-underline
     outline-none
     relative
     text-color-3;
-  background-color: var(--calcite-accordion-item-background);
-  --calcite-accordion-item-border: theme("borderColor.color.2");
+  background-color: var(--calcite-accordion-item-background, theme("backgroundColor.foreground.1"));
 }
 
 // focus styles
@@ -90,7 +77,7 @@
 :host .accordion-item-content,
 :host .accordion-item-header {
   padding: var(--calcite-accordion-item-padding);
-  border-bottom: 1px solid var(--calcite-accordion-item-border);
+  border-bottom: 1px solid var(--calcite-accordion-item-border, theme("borderColor.color.2"));
 }
 
 :host .accordion-item-header * {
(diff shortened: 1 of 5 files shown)


src/components/calcite-accordion-item/calcite-accordion-item.tsx
```diff
@@ -49,6 +50,11 @@ export class CalciteAccordionItem {
   /** optionally pass an icon to display - accepts Calcite UI icon names  */
   @Prop({ reflect: true }) icon?: string;
 
+  /**
+   * @internal
+   */
+  @Prop({ reflect: true })
+  appearance: AccordionAppearance;
```
driskull, line 57: Can we use a CSS class on the internal element instead of a private property?

src/components/calcite-accordion-item/calcite-accordion-item.scss
```diff
@@ -32,10 +32,10 @@
   --calcite-accordion-item-background: theme("backgroundColor.foreground.1");
 }
 
-:host-context([appearance="minimal"]) {
```
driskull, line 35: Lets change this to a CSS class.

src/components/calcite-accordion-item/calcite-accordion-item.scss
```diff
@@ -32,10 +32,10 @@
...
+:host([appearance="minimal"]) {
   --calcite-accordion-item-padding: var(--calcite-accordion-item-spacing-unit) 0;
 }
-:host-context([appearance="transparent"]) {
+:host([appearance="transparent"]) {
   --calcite-accordion-item-border: transparent;
```
driskull, line 39: these CSS properties could be set on the parent item (accordion) instead of here.

driskull (changes requested):
Instead of adding properties that are undocumented, lets keep the CSS internal and use the CSS properties and set them on the parent item.

src/components/calcite-accordion-item/calcite-accordion-item.tsx
anveshmekala: > Instead of adding properties that are undocumented, lets keep the CSS internal and use the CSS properties and set them on the parent item.

If we need to apply CSS classes on parent element `calcite-accordion` we might need to use `::slotted`  . Is it suggestible ?

anveshmekala:
> Instead of adding properties that are undocumented, lets keep the CSS internal and use the CSS properties and set them on the parent item.

If we need to apply CSS classes on parent element `calcite-accordion` we might need to use `::slotted`  . Is it suggestible ?

driskull:
> If we need to apply CSS classes on parent element calcite-accordion we might need to use ::slotted . Is it suggestible ?

No, not on the parent host, inside the parent host. Somewhere surrounding the default slot.

driskull (approved):
👍 🎉 🌮 Looks good to me!

paulcpederson:
I think this pr broke styling for accordions. Transparent now have a border and I think minimal is also incorrect now.

macandcheese:
76: https://jsbin.com/teparewaxa/1/edit?html,output
75: https://jsbin.com/fixeqawebu/1/edit?html,output

The regressions look to be:
- border on three sides of "transparent" (should be removed)
- borders on "minimal" - should just be the bottom line under each item.

macandcheese:
cc @benelan

paulcpederson:
https://github.com/Esri/calcite-components/issues/4120

Issue #3725, opened 2021-12-15
PickList and ValueList - add position definition for slotted Dropdown


asangma:
### Actual Behavior
In Firefox and Safari, when a Dropdown is added to the `menu-actions` slot of PickList (and ValueList), the Dropdown's position doesn't respect the `sticky` position of the header.

![image](https://user-images.githubusercontent.com/44481292/149390995-e13d0cd4-3e17-461a-82dc-a846143bb2b5.png)

### Expected Behavior
Dropdown should stay sticky with the header.
Solution appears to be simply adding `position: relative` to a Dropdown that is in the `menu-actions` slot.

### Reproduction Steps and Sample

_Sample_: https://codepen.io/afreitag/pen/vYegbbg?editors=1000

Note: we're not able to repro in a Codepen, but it's happening in Feature Specific effects in Map Viewer. See attached gif.

1. In FF or Safari, open the filter next to the search field
2. scroll

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-<version>`

- [ ] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-<version>`

asangma:
@benelan Mind if we push up the priority on this?

benelan:
I put it in the next sprint @asangma

y0n4:
Hi, could you include an attached gif? Thanks!

jcfranco:
This is blocked by #3815.

geospatialem:
There are a few blockers associated with this fix, including (Feb 18): 

> This is blocked by #3815.

Additional R&D has determined fixing for picklist and valuelist, creates a break for others.

AdelheidF:
I would like to point out here how big of a problem this is in just the Map Viewer application. (I probably didn't catch all scenarios.)

Field picker used in Effects (shown here), Table, Styles, Filter, Labels, FeatureDisplayOrder components
![picklist](https://user-images.githubusercontent.com/44481292/166306789-02244383-e119-4462-9e0b-e5b122329c75.gif)

Unique values picker in Filter component, which is also used in Effects, Labels, Clustering component
![picklist2](https://user-images.githubusercontent.com/44481292/166306806-22db59af-0dcb-43d6-a081-1ce7ffd7357f.gif)

Symbol picker in Sketch and Styles components
[...]

Issue #3007, opened 2021-09-10
Bug: calcite-dropdown cut off when expanded inside of a small calcite-tab


noahmulfinger:
### Actual Behavior

When a `calcite-dropdown` is triggered/expanded inside of a `calcite-tab` that has a height smaller than the dropdown, the dropdown is cut off and some items are not clickable. 

### Expected Behavior

I would expect the dropdown to expand outside of the `calcite-tab`, similar to the behavior of a popover or tooltip. 

### Reproduction Steps or Sample

1. Go to https://codepen.io/noahmulfinger/pen/PojpdJN
2. Click on the "Dropdown" button and you will only the see the first option visible, with the rest cut off.
3. Click on the "Popover" button and it will expand outside of the `calcite-tab` with no issue.

### Relevant Info

Tested on latest Firefox and Chrome

_Version_: `@esri/calcite-components@63`

- [x] CDN
- [x] NPM package

driskull:
Seems like the `position:relative` needs to be removed on the dropdown host.

julio8a:
maybe related #3071

benelan:
Dashboard is having issues due to the PR for this issue. Can you take another look? 

cc @ellenupp

anveshmekala:
I did tried to simulate the issue in code-pen and the dropdown worked as expected in chrome,Firefox and edge. 
https://codepen.io/ANVESHMEKALA/pen/gOxpPyz

The reason behind removing the `position:relative` on host is to give the end users the flexibility to set the position. Would recommend setting the `position:relative` on host element in the application and verify if the issue persists.

cc @ellenupp

benelan:
It looks like it has something to do with `placement`. In @anveshmekala's sample, creating the dropdown with a placement prop also has weird issues:
` <calcite-dropdown placement='bottom'>`

benelan:
@ellenupp yeah a new issue for that would be great, thanks!

Issue #3815, opened 2021-12-30
Research: nested popper is contained within the component instead of overflowing


Elijbet:
**Related Issues** 
#3099
#3724
#3725 
#3667 

### Background 

We have an issue with components using nested popper.js. Instead of overflowing, these are contained within the component.

eg. if placed inside a modal, a combobox is contained within that modal by default.

### Desired Outcome 

In this example the calcite-combobox list container needs to overflow like it does with calcite-select.

Sample: 
calcite-select: https://codepen.io/benesri/pen/NWgLgJJ
 calcite-combobox: https://codepen.io/benesri/pen/eYRLWej

It was suggested that we look at other component libs and how they handle nested popovers. That would require looking into libs that have some sort of custom modal, dropdown, or similar component using popper.

Also need to look into accessibility issues around the solution.

Please let me know if I'm missing anything from the issue.

driskull:
It seems like carbon does it the same way. An issue I found says that its just up to the container to overflow: visible.
https://github.com/carbon-design-system/carbon/issues/2573

Spectrum components have the same issue It appears.
https://webcomponents.dev/edit/collection/fO75441E1Q5ZlI0e9pgq/EYQn9T6wOnieZbv4xnPj/src/index.stories.js?p=stories

Paper has same issue:
https://www.webcomponents.org/element/@polymer/paper-dropdown-menu/demo/demo/index.html

Looks like vaadin appends an overlay to the body.
https://vaadin.com/docs/latest/ds/components/combo-box

It seems like the easiest thing to do is to add an option to modal to either scroll or overflow-visible for these things.

jcfranco:
https://github.com/floating-ui/floating-ui/pull/1827 should help address some issues related to using the `fixed` positioning strategy. Tested locally on some of our reported issues and floating-UI placement seems to work as expected.

If additional testing looks good, I believe we can consider this fixed once ☝️, #4372 and #4355 land.

jcfranco:
The floating UI PRs were installed and are included in the latest release. Local testing of referenced issues looks good! @benelan Per our convo, assigning to you for follow up once the doc is done.

geospatialem:
Verified with the `overlay-positioning` attribute when the value is set to "fixed". 👍🏻

PR #3958, opened 2022-01-19
fix(picklist,valuelist): dropdown menu will be postioned sticky

anveshmekala:
**Related Issue:** #3725 

## Summary

Dropdown menu slotted in `menu-actions` of `calcite-pick-list` and `calcite-value-list` will be positioned `sticky` in FF and safari. 

With this fix,  [#3007](https://github.com/Esri/calcite-components/issues/3007) will be re-introduced and  the work around would be making sure the tab has enough `height` to avoid overflow of dropdown.

### DIFF
--- a/src/components/calcite-pick-list/calcite-pick-list.scss
+++ b/src/components/calcite-pick-list/calcite-pick-list.scss
@@ -35,3 +35,7 @@ calcite-filter {
 :host([loading][disabled]) {
   min-height: 2rem;
 }
+
+::slotted(calcite-dropdown) {
+  position: relative;
+}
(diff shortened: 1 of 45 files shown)


src/components/calcite-pick-list/calcite-pick-list.scss
```diff
@@ -35,3 +35,7 @@ calcite-filter {
 :host([loading][disabled]) {
   min-height: 2rem;
 }
+
+::slotted(calcite-dropdown) {
```
jcfranco, line 39: Is there a way this can be tackled in a generic way? Having component-specific fixes will not scale well.

@driskull Will setting `position:relative` on the shadow DOM root container improve positioning or does it have to go on the host?

jcfranco (changes requested):

src/components/calcite-pick-list/calcite-pick-list.scss
driskull: Shadow dom root container should work and is preferred 👍

anveshmekala:
holding off . related to #3815

PR #3963, opened 2022-01-20
ci(screener): adding Firefox 96

eriklharper:
**Related Issue:** n/a


eriklharper:
We need to get Chrome 97 stable before we can add additional browsers, so I'm going to close this while we work out a full resolution.

PR #3964, opened 2022-01-20
chore: pin node 16 and npm 8 for volta

benelan:
**Related Issue:** NA

## Summary
Pinning node16 / npm8 for volta users


driskull (approved):
👏

Issue #3667, opened 2021-12-09
Calcite Dropdown - Dropdown positioning is incorrect when nested in shadow DOM - Only in Browsers other than Chrome (confirmed in Firefox & Safari)


Csmith246:
### Actual Behavior

Issue can be seen in following codepen, when using FF or safari (it behaves properly in Chrome)(https://codepen.io/csmith55/pen/NWardEd?editors=1000)

Dropdown opens further to the right than it should. This appears to be a cumulative effect

![image](https://user-images.githubusercontent.com/22039422/145312230-17f7b4fc-6351-42a3-bd14-cd59b378f21c.png)

### Expected Behavior

Dropdowns to open attached to button, as usual

### Reproduction Steps and Sample

Open Codepen with FF, and press dropdown button

_Sample_:

1. provided

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-<version>` - all

- [x] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-<version>` - all

driskull:
Seems to be fixed if the `calcite-dropdown` is positioned as relative within the shadowDom.

jcfranco:
This is blocked by https://github.com/Esri/calcite-components/issues/3815.

jcfranco:
Local testing w/ latest version of `floating-ui` (see https://github.com/Esri/calcite-components/issues/3815#issuecomment-1201698447 for more info) seems to work as expected. Marking as installed!

geospatialem:
Verified placement in Firefox with next.557 via https://codepen.io/geospatialem/pen/VwXBGNv.

![verified dropdown positioning](https://user-images.githubusercontent.com/5023024/183940467-4776652a-0d7a-476f-8fea-a1385cd1e166.png)

PR #3966, opened 2022-01-20
fix(card):  shift position of card component for child dropdown menu to align consistently

y0n4:
**Related Issue:** #3667 

## Summary
Specifically in Firefox browser:
The dropdown menu in `calcite-card` fell out of range from the dropdown button due to the current positioning of the component. There was also a similar issue found with with the dropdown demo page as well

Before
Screen Shot 2022-01-20 at 1 32 39 PM
After
Screen Shot 2022-01-20 at 1 34 48 PM

Before
Screen Shot 2022-01-26 at 4 17 04 PM
After
Screen Shot 2022-01-26 at 4 10 05 PM


src/components/calcite-dropdown/calcite-dropdown.scss
```diff
@@ -7,7 +7,7 @@
  */
 
 :host {
-  @apply inline-flex flex-initial;
+  @apply inline-flex flex-initial relative;
```
driskull, line 10: Looks good, I only wonder if we should add another container internally to apply this so that it can't be messed with by the end user. Something to contain both the trigger container and the dropdown menu.

driskull (commented):
@jcfranco what are your thoughts here?

src/components/calcite-dropdown/calcite-dropdown.scss
jcfranco: Yeah, that's a good idea. Otherwise, this will come up again if changed by the user.

y0n4: It works great but I realized, doing this ended up making the dropdown list change its positioning when it is a child in a tab component

driskull: > It works great but I realized, doing this ended up making the dropdown list change its positioning when it is a child in a tab component

I think that might be a separate issue.

src/components/calcite-card/calcite-card.scss
```diff
@@ -8,7 +8,6 @@
       bg-foreground-1
       duration-150
       ease-in-out
-      relative
```
driskull, line 11: @y0n4 the PR title and commit message is for dropdown but there are no changes dropdown, only card.

Can we update the PR title/description?

Does this need a test of some kind?

driskull (changes requested):

src/components/calcite-card/calcite-card.scss
```diff
@@ -29,7 +28,11 @@
   @apply flex-auto flex flex-col;
 }
 
-:host([loading]) .calcite-card-container *:not(calcite-loader):not(.calcite-card-loader-container) {
+:host([loading]) .calcite-card-container {
+  @apply relative;
```
driskull, line 32: @y0n4 can you comment on why we still need this container to be relative when loading?

Do you know why the relative messes up positioning after loading?

driskull (commented):
Code changes look good, just want to understand why the position relative is messing things up.

src/components/calcite-card/calcite-card.scss
y0n4: I'm not sure for it to be just a Firefox browser issue. I'm thinking the relative positioning was affecting the following child's element with relative positioning even though it should be set to the boundaries of the dropdown position

driskull: Interesting, I would expect the relative positioning to be a good thing.

Maybe this is something we can fix inside the dropdown component instead of having to fix it in other components?

Issue #3965, opened 2022-01-20
input-date-picker/date-picker no longer looks up the base language file if the specific locale code is not supported


jcfranco:
### Actual Behavior

Setting `pt-BR` (no matching locale file) as the date-picker's `locale` no longer loads the `pt` locale.

### Expected Behavior

When an unsupported locale value with language and region is provided (e.g., `pt-BR`), the language file should be used as the locale before falling back to English.

### Reproduction Sample

https://codepen.io/jcfranco/pen/NWaZbbN?editors=1000

### Reproduction Steps

1. Click on the input and notice the date-picker shows in `pt` (since `pt-BR` is not found)
2. Change the version to beta.75 
3. Repeat step 1 and notice how the date-picker displays in English.

### Reproduction Version

@esri/calcite-components@1.0.0-beta.75

### Relevant Info

Regression? Last working version: `@esri/calcite-components@1.0.0-beta.74`

This was introduced by https://github.com/Esri/calcite-components/issues/3925

### Source

- [X] CDN
- [ ] NPM package

jcfranco:
Triaged to this milestone. cc @benelan 

@y0n4 Can you look at this one?

AdelheidF:
Works with beta.74, doesn't work with beta.75, works with next.

benelan:
verified by Adelheid on next

PR #3968, opened 2022-01-20
fix(date-picker): update utils locale to get the lang code if regional code is not found

y0n4:
**Related Issue:** #3965 

## Summary
When a specific file cannot be found if a locale with regional code does not exist, then the date picker will seek for just the basic lang code instead.
For example: `pt-BR` -> `pt`
Before
Screen Shot 2022-01-20 at 3 19 40 PM
After
Screen Shot 2022-01-20 at 3 21 21 PM

### DIFF
--- a/src/components/calcite-date-picker/utils.spec.ts
+++ b/src/components/calcite-date-picker/utils.spec.ts
@@ -28,6 +28,13 @@ describe("utils", () => {
       expect(requestCache).toHaveProperty("en");
     });
 
+    it("falls to lang code locale if regional code is not found", async () => {
+      const locale = "pt-BR";
+
+      await getLocaleData(locale);
+      expect(requestCache).toHaveProperty("pt");
+    });
+
     it("fetches locale with conventional-cased lang code", async () => {
       const locale = "es";
 
--- a/src/components/calcite-date-picker/utils.spec.ts
+++ b/src/components/calcite-date-picker/utils.spec.ts
@@ -29,7 +29,7 @@ describe("utils", () => {
     });
 
     it("falls to lang code locale if regional code is not found", async () => {
-      const locale = "pt-BR";
+      const locale = "pt-UnsupportedRegion";
 
       await getLocaleData(locale);
       expect(requestCache).toHaveProperty("pt");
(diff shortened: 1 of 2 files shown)


src/components/calcite-date-picker/utils.ts
```diff
@@ -42,6 +42,9 @@ function getSupportedLocale(lang = "") {
 
   if (lang.includes("-")) {
     lang = lang.replace(/(\w+)-(\w+)/, (_match, language, region) => `${language}-${region.toUpperCase()}`);
+    if (!locales.includes(lang)) {
+      return locales.includes(lang.split("-")[0]) ? lang.split("-")[0] : "en";
```
driskull, line 46: Can you save `lang.split("-")[0])` as a const so that you don't have to split it twice here? Otherwise looks good! 👍

driskull (approved):
👍

src/components/calcite-date-picker/utils.spec.ts
```diff
@@ -28,6 +28,13 @@ describe("utils", () => {
       expect(requestCache).toHaveProperty("en");
     });
 
+    it("falls to lang code locale if regional code is not found", async () => {
+      const locale = "pt-BR";
```
jcfranco, line 32: Heads up, we may add `pt-BR` (depending on https://github.com/Esri/calcite-components/issues/3925#issuecomment-1018026093), so could you use another region to prevent this test from breaking? Maybe `pt-UnsupportedRegion` to make the test clearer and less brittle?

src/components/calcite-date-picker/utils.ts
```diff
@@ -42,6 +42,9 @@ function getSupportedLocale(lang = "") {
 
   if (lang.includes("-")) {
     lang = lang.replace(/(\w+)-(\w+)/, (_match, language, region) => `${language}-${region.toUpperCase()}`);
+    if (!locales.includes(lang)) {
```
jcfranco, line 45: Nitpick: for readability, can you add a newline before the if statement?

src/components/calcite-date-picker/utils.ts
jcfranco, line 46: You could use the original return statement if you assign the language part to `lang`.

```ts
lang = lang.split("-")[0]; // the rest is handled by the existing return statement
```

jcfranco (approved):
Very nice! Other than a few suggestions, this LGTM!

PR #3970, opened 2022-01-21
docs(template): provide context for the repro samples

benelan:
**Related Issue:** NA

## Summary
- Provide context for the repro samples we provide in the issue templates
- Change the JSBin to use `next`, it was still at `beta.64` 
    - Should it be `next` or `latest`? I think having it as `next` can help prevent unnecessary issues by having people test the unreleased changes.


driskull (approved):

jcfranco (approved):
Nice!

jcfranco:
> Should it be `next` or `latest`

Let's stick to `latest` since `next` may not be stable in some instances.

PR #3980, opened 2022-01-21
fix(date-picker, input-date-picker): update pt-BR and pt-PT localization files

jcfranco:
**Related Issue:** N/A

## Summary

Stemmed from https://github.com/Esri/calcite-components/issues/3925#issuecomment-1018028342. This PR adds `pt-BR.json` and updates some entries `pt-PT.json` per the CLDR locale dataset.

One thing I did notice was that `unitOrder` and `placeholder` from our files are different from the CLDR data. 

### Current

```json
{
    "default-calendar": "gregorian",
    "separator": "/",
    "unitOrder": "DD/MM/YYYY",
    "weekStart": 7,
    "placeholder": "DD/MM/YYYY",
    ...
}
```

### CLDR

```json
{
    "default-calendar": "gregorian",
    "separator": "/",
    "unitOrder": "dd/MM/yy",
    "weekStart": 7,
    "placeholder": "dd/MM/yy",
    ...
}
```

Worth noting that aside from Bulgaria (bg), all other locale files use the `MM`, `DD` and `YYYY` tokens. 

I kept the previous versions since that's what is being used by the input date picker component. Having said this, which version should we be using from the input date picker perspective? If it's the `MM`, `DD`, `YYYY` version, is this something that we need to maintain here or should this be coming from the source localization files? cc @raje9276 @paulcpederson

### DIFF
--- a/src/components/calcite-date-picker/assets/calcite-date-picker/nls/pt-BR.json
+++ b/src/components/calcite-date-picker/assets/calcite-date-picker/nls/pt-BR.json
@@ -5,7 +5,7 @@
   "weekStart": 7,
   "placeholder": "DD/MM/YYYY",
   "days": {
-    "abbreviated": ["domingo", "segunda", "terça", "quarta", "quinta", "sexta", "sábado"],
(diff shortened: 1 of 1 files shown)


driskull (approved):
👍

Issue #3755, opened 2021-12-21
[Input Message / Input] Remove usage of `getElementProp` to set `status`


benelan:
### Description 
`calcite-input-message` and `calcite-input` should not use `getElementProp` to set `status`.

### Proposed Advantages 
This is not a great pattern for inheriting properties - as of now it only detects and sets on load. Users should manage `status`  on each component individually.

### Relevant Info 
The verbiage on `calcite-label` here: (https://github.com/Esri/calcite-components/tree/master/src/components/calcite-label#basic) should also be removed.

#### Which Component
`calcite-input-message`
`calcite-input`
`calcite-label`

cc @driskull

macandcheese:
To clarify (I think), this property should be kept on `calcite-input-message` and `calcite-input`, but their use of `getElementProp` to get and set the prop from `calcite-label` should be stopped.

The verbiage on `calcite-label` here: (https://github.com/Esri/calcite-components/tree/master/src/components/calcite-label#basic) should also be removed.

y0n4:
Is it also recommended to remove the verbiage or change the code demo by adding additional status prop in the components for [calcite-input ](https://github.com/Esri/calcite-components/tree/master/src/components/calcite-input#with-label) and [calcite-input-mesage](https://github.com/Esri/calcite-components/tree/master/src/components/calcite-input-message#basic)?

y0n4:
Does `calcite-label` still need the `status` property other then the purpose of having it propagate to `calcite-input`/`calcite-input-message`?

driskull:
No, we should deprecate it or remove it.

benelan:
verified on beta.76

PR #3984, opened 2022-01-22
feat(label): deprecate status prop

y0n4:
**Related Issue:** #3755

## Summary
Originally user can set status indirectly by declaring status prop in `calcite-input`
```
<calcite-label status="invalid">
  <calcite-input></calcite-input>
  <calcite-input-message>Invalid</calcite-input-message>
</calcite-label>
```
However it is not ideal so to set status in `calcite-input` and `calcite-input-message`, the prop must be declared directly within those components. The status in `calcite-label` will be deprecated.
```
<calcite-label>
  <calcite-input status="invalid"></calcite-input>
  <calcite-input-message status="invalid">Invalid</calcite-input-message>
</calcite-label>
```


driskull (commented):
@y0n4 what about removing or deprecating the status property on the label?

Would this be a breaking change if we remove the getElementProp support on the inputs?

src/components/calcite-input/calcite-input.e2e.ts
```diff
@@ -75,14 +75,13 @@ describe("calcite-input", () => {
...
     await page.setContent(html`
-      <calcite-label status="invalid" scale="s">
+      <calcite-label scale="s">
         Label text
         <calcite-input></calcite-input>
       </calcite-label>
     `);
 
     const element = await page.find("calcite-input");
-    expect(await element.getProperty("status")).toEqual("invalid");
```
jcfranco, line 85: I think you can drop this since the test is about inherited props, which `status` no longer participates in. We have another issue to tackle `scale`, but we'll get to that separately.

src/components/calcite-label/calcite-label.tsx
```diff
@@ -28,7 +28,10 @@ export class CalciteLabel {
   /** specify the text alignment of the label */
   @Prop({ reflect: true }) alignment: Alignment = "start";
 
-  /** specify the status of the label and any child input / input messages */
+  /**
+   * specify the status of the label and any child input / input messages
+   * @deprecated status is no longer supported, set status directly in child input / input messages
```
jcfranco, line 33: Pinging @benelan for editorial review. :)

benelan: Can you please change the `@deprecated` message to `set directly on child element instead` to match the cadence of our other deprecation notices.

jcfranco (commented):
Sorry for not catching this sooner, but since we are deprecating the label prop, we should still have it work. We will just have the examples and doc not promote the deprecated way. That means this will no longer be a breaking change, so no need for the `!` in the PR title nor the BREAKING CHANGES footer when merging.

So, can you do the following?

1. Restore the input and message (**not label**) e2e tests and update the test name with a `(deprecated)` or add a comment to the assertions (if the test title isn't specific to `status`) to indicate that it is deprecated. 
2. Restore the `this.status = getElementProp(this.el, "status", this.status);` lines.

I'll approve after these changes are made.

driskull:
Thanks for chiming in! I wasn't sure if we wanted to deprecate or if we were ok removing it and allowing a breaking change.

I guess the issue should have clarified a deprecation instead of removal.

src/components/calcite-input-message/calcite-input-message.e2e.ts
```diff
@@ -37,8 +37,8 @@ describe("calcite-input-message", () => {
     </calcite-label>
     `);
 
-    const element = await page.find("calcite-input-message");
-    expect(element).toEqualAttribute("status", "invalid");
+    const deprecatedLabelStatusElement = await page.find("calcite-input-message");
```
jcfranco, line 40: Nice!

jcfranco (approved):
👍🏼 

I renamed the PR title according to [this thread](https://github.com/openedx/open-edx-proposals/pull/189#discussion_r591111224).

src/components/calcite-label/calcite-label.tsx
```diff
@@ -28,7 +28,10 @@ export class CalciteLabel {
   /** specify the text alignment of the label */
   @Prop({ reflect: true }) alignment: Alignment = "start";
 
-  /** specify the status of the label and any child input / input messages */
+  /**
+   * specify the status of the label and any child input / input messages
+   * @deprecated set directly on child element instead
```
jcfranco, line 33: +@benelan for editorial glance. 👀

benelan: Looks good, thanks for changing it Yona!

benelan: Although the prop description is kinda confusing to me but that was already there lol. The new deprecation notice is 👨‍🍳 👌

Issue #3986, opened 2022-01-23
Progress component is fully invisible in high contrast mode


mpriour:
### Summary

The progress component is fully invisible in high contrast mode. Both the track and the bar use `background-color` values to make the elements show. The values used are changed to `none` by the forced colors rendering process that occurs in high contrast, making them not show at all.

This issue was discovered as part of #3887 - [High contrast audit](https://github.com/Esri/calcite-components/wiki/High-Contrast-Audit#progress) and identified as a severity level 4 out of 4 issue.

### Actual Behavior

Totally invisible (from the development demos page)
<img alt="progress, high contrast" height="240" src="https://user-images.githubusercontent.com/142636/149631329-0876ba46-a08a-4c58-b5eb-b28cceaf9c1a.png" />

### Expected Behavior

Something like this, but with high contrast system colors

<img alt="progress, normal" height="240" src="https://user-images.githubusercontent.com/142636/149631319-06384595-789c-45a0-872f-b48489d8bf45.png" />

### Reproduction Sample

https://esri.github.io/calcite-components/?path=/story/components-progress--indeterminate

### Reproduction Steps

1.  Set the theme on a Windows machine or vm instance to high contrast accessibility
2. View the above sample. Nothing is shown.

### Reproduction Version

@esri/calcite-components@1.0.0-75

### Relevant Info

Windows 7, 10, 11 with a high contrast theme enabled.
Any browser (Edge, Chrome, or Firefox)

### Source

- [X] CDN
- [X] NPM package

benelan:
verified on beta.77

PR #3988, opened 2022-01-24
fix(progress): make progress visible in high contrast (#3986)

mpriour:
* Use system colors for background of track and bar elements

**Related Issue:** #3986 

## Summary
Provide system colors for track and bar of the progress component

Looks like this on high contrast in Edge on Windows:

Screen Shot 2022-01-23 at 8 03 22 PM

Uses `HighlightText` for the track, which should always be a light color on dark high contrast themes and `LinkText` for the bar which should always be a highly visible color and different from borders or text.


mpriour:
@macandcheese do you approve of the design used here, or would you like to see something different?

macandcheese (approved):
Looks good to me - thanks @mpriour !

PR #3990, opened 2022-01-24
ci(screener): static storybook build

eriklharper:
**Related Issue:** #

## Summary

This PR turns on Screener's static storybook build option so that it doesn't have to start up the Storybook Dev Server to run the visual tests.  The hope is this will resolve ngrok server failures and also speed up the tests.


eriklharper:
This issue was an attempt to improve Screener performance with Sauce browsers, but because we're abandoning that effort, we can also abandon these changes.

PR #3992, opened 2022-01-24
build(deps): fix and run update deps action

benelan:
**Related Issue:** NA

## Summary
- exclude `ts-jest` from the update script. The rest of the `jest` packages are excluded and it was causing peer deps issues trying to bump that one
- The script now reruns if it failed (up to 10 times). 
    - Before, if all of the dep tests except for the last one passed then `package.json` would be updated with all of the deps except the last. However since the last one failed the whole action would end. This happens sometimes due to unstable tests. 
    - Now, if the last one fails, the action will rerun only the last dep's tests since the `package.json` has the rest of the deps updated.
- The action now sends the PR even if the update step failed, for the same reason as above. If it fails 10 times on the last dep, then it will send a PR with all of the updates except the last dep. If there are no changes then the commit/PR won't be sent.
- Run the action on Tuesdays and move dependabot to Thursdays. This gives us a couple days to merge before individual dependabot PRs come in. Hopefully this will cut down on the PRs we need to merge, while still having dependabot as a backup in case this action fails due to peer dep conflicts, e.g.
    image


.github/workflows/update-deps.yml
```diff
@@ -12,19 +12,19 @@ jobs:
...
+        if: always()
         with:
           base: master
           branch: deps-update
           commit_message: "build(deps): bump versions"
           token: ${{ secrets.ADMIN_TOKEN }}
           title: "build(deps): bump versions"
           body: This PR was automatically generated by the update-deps GitHub action
-          labels: "chore"
+          labels: "dependencies"
```
jcfranco, line 30: WDYT about keeping chore, so it matches the commit type?

jcfranco (approved):
💥

.github/workflows/update-deps.yml
benelan: The commit type is `build` though. And I changed it so it matches the Dependabot PRs which have the `dependencies` label. I can switch it back if you want

jcfranco: Gotcha. Sounds good.

Issue #3883, opened 2022-01-08
calcite-time-picker/calcite-input-time-picker: wrong order of time in RTL


AdelheidF:
According to Olga the current format of ss:mm:hh is not correct for Arabic or Hebrew. The numbers should not switch order between LTR and RTL. 

value="18:46"
![image](https://user-images.githubusercontent.com/44481292/148110669-d34924f6-bf50-4e2c-9c8b-01fd1ebf6b45.png)

![image](https://user-images.githubusercontent.com/44481292/149235717-29449841-0dce-4db8-8fac-e97640323e57.png)

https://codepen.io/afreitag/pen/QWqrRGg

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-beta.73`

- [ ] CDN
- [ ] NPM package

AdelheidF:
looks good to me now

AdelheidF:
Actually, I think the unit needs to stay in front of the number.

This is wrong 
![image](https://user-images.githubusercontent.com/44481292/151268708-6567776f-cd94-4540-8731-b1c11acf5905.png)

benelan:
> Actually, I think the am/pm needs to stay in front of the numbers.

Assigning back to @anveshmekala to look into ^

AdelheidF:
Those 2 might be related, not sure. https://github.com/Esri/calcite-components/issues/1475#issuecomment-1022739244

anveshmekala:
@AdelheidF  Setting meridiem  in front of the time format is applicable for just `rtl` direction ?

cc: @asangma

AdelheidF:
Seems to be also the case for ko and zh, see here: https://github.com/Esri/calcite-components/issues/1475#issuecomment-1023475022

Isn't there a lookup you could use?

I don't know those specifics....

jcfranco:
@anveshmekala brought this up, but is this a duplicate of #1475? They both seem related to the ordering of time parts per locale.

AdelheidF:
> They both seem related to the ordering of time parts per locale.

The original issues addressed separate things, but what's left is kind of similar now. 

I opened this issue here because of you guys not commenting on this: https://github.com/Esri/calcite-components/issues/1475#issuecomment-1005093188

AdelheidF:
To recap:

What's left here is that in RTL languages AR and HE the am/pm is on the wrong side in `calcite-time-picker`. `calcite-input-time-picker` is correct.

[...]

PR #3993, opened 2022-01-24
fix(time-picker): match rtl order with ltr 

anveshmekala:
**Related Issue:** #3883 

## Summary

This fix will match  `rtl`  version of `calcite-time-picker` with `ltr` version.

Earlier, when `dir` attribute is set to `rtl` the `calcite-time-picker` would align the container as below:
5E2CBC0C-58C9-477F-A8EA-E93AC68D09D5

Current change will set the container to match the ltr when the `dir` is set to `rtl`.

81FABAD6-A29C-4869-8422-F1F48A885317_4_5005_c

### DIFF
--- a/src/components/calcite-time-picker/calcite-time-picker.scss
+++ b/src/components/calcite-time-picker/calcite-time-picker.scss
@@ -109,6 +109,6 @@
   }
 }
 
-:host([lang="ar"]) ([lang="he"]) {
+:host([lang="ar"], [lang="he"]) {
   direction: ltr;
 }
--- a/src/components/calcite-time-picker/calcite-time-picker.scss
+++ b/src/components/calcite-time-picker/calcite-time-picker.scss
@@ -108,7 +108,3 @@
     }
   }
 }
-
-:host([lang="ar"], [lang="he"]) {
-  direction: ltr;
-}
--- a/src/components/calcite-time-picker/calcite-time-picker.scss
+++ b/src/components/calcite-time-picker/calcite-time-picker.scss
@@ -108,3 +108,7 @@
     }
   }
 }
+
+:host([lang="ar"]) ([lang="he"]) {
+  direction: ltr;
+}
--- a/src/components/calcite-time-picker/calcite-time-picker.e2e.ts
+++ b/src/components/calcite-time-picker/calcite-time-picker.e2e.ts
@@ -1003,5 +1003,14 @@ describe("calcite-time-picker", () => {
 
       expect(meridiem.textContent).toBe("PM");
     });
+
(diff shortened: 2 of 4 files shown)


src/components/calcite-time-picker/calcite-time-picker.scss
```diff
@@ -108,3 +108,7 @@
     }
   }
 }
+
+:host([lang="ar"], [lang="he"]) {
```
driskull, line 112: @anveshmekala this will only work if the component has a lang attribute on it. What if the lang is set on a parent?

Also, we don't want to set `direction` on the host because it can be overriden by a user.

So, if we need to force direction ltr always, lets set it inside the component where the direction needs to be ltr always.

driskull (changes requested):
Comments! :)

src/components/calcite-time-picker/calcite-time-picker.scss
anveshmekala: If the user doesn't specify the `lang` attribute then the requirement to match the `rtl` with `ltr` will be out of scope.

eriklharper: I guess my question is why isn't the browser correctly setting the direction for `ar` and `he` locales?  Are we trying to override otherwise correct behavior on the part of the browser or is this a browser bug?

eriklharper: And I agree with Matt's comments.  Any kind of CSS solution is brittle and overridable.  If we need to address direction it needs to be in the component itself.  I also would like to see this component go away in the future in favor of the new `calcite-input-time` component which would likely just consist of an input.

driskull: > Any kind of CSS solution is brittle and overridable.

Only if its on the host. ;)

> Are we trying to override otherwise correct behavior on the part of the browser

Thats what the issue wants done. I'm not sure why exactly.

driskull: 1. We would not want to check for specific languages and set styles on any components.
2. Setting styles on the host will expose those styles to the user and allow them to easily override them.
3. Using this code above, the following wouldn't set the `direction: ltr;` css.

```html
<div lang="ar">
<calcite-time-picker />
</div>
```

So to handle this issue, we can:

Set `direction:ltr` on the element that contains the row of time picking inputs.

This would ensure that it is always LTR even if the direction is set to RTL.

driskull (approved):
Can we add a e2e test that checks the direction on the element? 👍

PR #3996, opened 2022-01-24
docs(readme): remove sketch info until it is up to date

benelan:
**Related Issue:** NA

## Summary
Our sketch UI kit is super far behind the components, so I'm removing it from the README until it is updated. People will be really frustrated if they take the time to design something and it is completely different in the component library. 

We may want to add a disclaimer notice in the dev docs until figma/sketch gets caught up. Thoughts @macandcheese @jcfranco ?

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -6,10 +6,6 @@
 
 Shared Web Components for Esri's Calcite Design Systems. To see the components in action, [view the documentation](https://developers.arcgis.com/calcite-design-system/components/).
 
-## Sketch library
-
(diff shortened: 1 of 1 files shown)


jcfranco:
> We may want to add a disclaimer notice in the dev docs until figma/sketch gets caught up. 

Makes sense to me. +@HeathMeyette

jcfranco (approved):
👍🏼

macandcheese:
Agreed.

Issue #3936, opened 2022-01-14
Unstable test: dropdown > opens when dropdown-trigger is an action or a button


Elijbet:
### Which Component

dropdown

### Which Tests(s)

```
    it.skip("opens when dropdown-trigger is a button", async () => {
      const page = await newE2EPage();
      await page.setContent(html`
        <calcite-dropdown>
          <calcite-button slot="dropdown-trigger">Open dropdown</calcite-button>
          <calcite-dropdown-group selection-mode="single">
            <calcite-dropdown-item id="item-1"> Dropdown Item Content </calcite-dropdown-item>
            <calcite-dropdown-item id="item-2" active> Dropdown Item Content </calcite-dropdown-item>
          </calcite-dropdown-group>
        </calcite-dropdown>
      `);
      const element = await page.find("calcite-dropdown");
      const trigger = await element.find("calcite-button[slot='dropdown-trigger']");
      const dropdownWrapper = await page.find(`calcite-dropdown >>> .calcite-dropdown-wrapper`);
      const calciteDropdownOpen = await element.spyOnEvent("calciteDropdownOpen");

      expect(await dropdownWrapper.isVisible()).toBe(false);
      await trigger.click();
      await page.waitForChanges();
      expect(await dropdownWrapper.isVisible()).toBe(true);

      await trigger.focus();
      await page.keyboard.press("Space");
      await page.waitForChanges();
      expect(await dropdownWrapper.isVisible()).toBe(false);
      expect(calciteDropdownOpen).toHaveReceivedEventTimes(1);

      await page.keyboard.press("Enter");
      await page.waitForChanges();
      expect(await dropdownWrapper.isVisible()).toBe(true);
      expect(calciteDropdownOpen).toHaveReceivedEventTimes(1);
    });

    it.skip("opens when dropdown-trigger is an action", async () => {
      const page = await newE2EPage();
      await page.setContent(html`
        <calcite-dropdown>
          <calcite-action slot="dropdown-trigger">Open dropdown</calcite-action>
          <calcite-dropdown-group selection-mode="single">
            <calcite-dropdown-item id="item-1"> Dropdown Item Content</calcite-dropdown-item>
            <calcite-dropdown-item id="item-2" active> Dropdown Item Content</calcite-dropdown-item>
          </calcite-dropdown-group>
        </calcite-dropdown>
      `);
      const element = await page.find("calcite-dropdown");
[...]

PR #3998, opened 2022-01-25
test(dropdown): correct logical errors in the test

Elijbet:
**Related Issue:** #3936

## Summary

This was flaking for 2 reasons:

- the event open/close fires after transition ends, so I needed to check for it in the right sequence and preceded with await page.waitForChanges();
- in the open position it fires calciteDropdownOpen, and in close: calciteDropdownClose; had to check for both

### DIFF
--- a/src/components/calcite-dropdown/calcite-dropdown.e2e.ts
+++ b/src/components/calcite-dropdown/calcite-dropdown.e2e.ts
@@ -640,8 +640,8 @@ describe("calcite-dropdown", () => {
     expect(await dropdownWrapper.isVisible()).toBe(false);
   });
 
-  describe("opens the dropdown with click, enter, or space", () => {
-    it.skip("opens when dropdown-trigger is a button", async () => {
+  describe("toggles the dropdown with click, enter, or space", () => {
+    it("toggles when dropdown-trigger is a button", async () => {
       const page = await newE2EPage();
       await page.setContent(html`
         <calcite-dropdown>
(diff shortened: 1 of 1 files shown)


src/components/calcite-dropdown/calcite-dropdown.e2e.ts
```diff
@@ -656,25 +656,30 @@ describe("calcite-dropdown", () => {
...
       const dropdownWrapper = await page.find(`calcite-dropdown >>> .calcite-dropdown-wrapper`);
       const calciteDropdownOpen = await element.spyOnEvent("calciteDropdownOpen");
+      const calciteDropdownClose = await element.spyOnEvent("calciteDropdownClose");
 
       expect(await dropdownWrapper.isVisible()).toBe(false);
       await trigger.click();
       await page.waitForChanges();
       expect(await dropdownWrapper.isVisible()).toBe(true);
+      await page.waitForChanges();
+      expect(calciteDropdownOpen).toHaveReceivedEventTimes(1);
```
jcfranco, line 666: WDYT about asserting both events after each toggle action? I'll leave this up to you to decide. 😄

src/components/calcite-dropdown/calcite-dropdown.e2e.ts
```diff
@@ -689,22 +694,27 @@ describe("calcite-dropdown", () => {
...
       await page.waitForChanges();
       expect(await dropdownWrapper.isVisible()).toBe(false);
-      expect(calciteDropdownOpen).toHaveReceivedEventTimes(1);
+      await page.waitForChanges();
+      expect(calciteDropdownClose).toHaveReceivedEventTimes(1);
 
       await page.keyboard.press("Enter");
       await page.waitForChanges();
       expect(await dropdownWrapper.isVisible()).toBe(true);
-      expect(calciteDropdownOpen).toHaveReceivedEventTimes(1);
```
jcfranco, line 707: Was this test failing consistently before or was it random? If it's the latter, you may need to use `page.waitForEvent` before doing assertions.

jcfranco (approved):
Awesome! This may need to use `page.waitForEvent` to ensure events are fired before asserting.

src/components/calcite-dropdown/calcite-dropdown.e2e.ts
driskull: I think it was consistent, is that right @Elijbet? Might be good to have in regardless.

driskull (approved):
Looks good. Might want to throw in the waitForEvent just for good measure so that there are no timing issues.

src/components/calcite-dropdown/calcite-dropdown.e2e.ts
Elijbet: I am not sure it was consistent. It was consistently receiving one event locally and passed the tests when I pushed it up. But after that I started seeing 2 events received instead of one.

PR #3915, opened 2022-01-12
fix(input): the visible input value stays in sync with the stored value when it is controlled programmatically

eriklharper:
**Related Issue:** #3003

## Summary

This PR fixes an issue where the visible input value in a number input wasn't reflecting the internally stored value when javascript code was programmatically setting the value of the input.

PR #4008, opened 2022-01-25
refactor(calcite-input): simplifying logic based on comments in #3915

### LINKED ISSUE #3003: Bug: calcite-input type=number not always updating value in box
beta.63

When updating a input box value programmatically the value inside the input box does not always update.
In the test case below I use it for the valid range of values, but it is not specific to that case.

I can reproduce it with SymbolStyler and I built a simple codepen.

The value of that input box should never be greater than 5. If the user types in a number that's bigger I want to set it back to 5. That works the first time, but not after. E.g. enter value 6, you can see the value changing back to 5 and below the input box I display what I then get from the input node as a value. Now enter 6 again or another number greater than 5 and you can see the value in the box stays at 6, but the node.value displayed below is correctly 5.

https://codepen.io/afreitag/pen/vYZxrmx?editors=1010

eriklharper:
**Related Issue:** #3003

## Summary

This PR addresses the comments in #3915 that originally fixed #3003.

### DIFF
--- a/src/components/calcite-input/calcite-input.tsx
+++ b/src/components/calcite-input/calcite-input.tsx
@@ -203,14 +203,7 @@ export class CalciteInput implements LabelableComponent, FormComponent {
     if (!this.internalValueChange) {
       this.setValue({
         origin: "external",
-        value:
-          newValue == null
-            ? ""
-            : this.type === "number"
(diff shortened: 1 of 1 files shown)


jcfranco (approved):
👍🏼

Issue #4005, opened 2022-01-25
a11y Date-picker component is unusable in high contrast


mpriour:
### Summary

In high contrast mode on Windows, the date-picker component is not useable.
* There is no visible indicator of the selected date.
* There are too many interior borders.
* There is no visible date range indicator.
* There is no visible indicator of a keyboard focused date.

This issue was discovered as part of #3887 - [High contrast audit](https://github.com/Esri/calcite-components/wiki/High-Contrast-Audit#datepicker) and identified as a severity 4 out of 4 issue. 

### Actual Behavior

Here is a date-picker with a selected date range in high contrast mode on Windows

<img height="300" alt="date-picker-high-contrast" src="https://user-images.githubusercontent.com/142636/149594591-be01da90-a63c-43ab-b9b9-74b06c92e9dd.png">

### Expected Behavior

It should be the high contrast equivalent to this:

<img height="300" alt="date-picker-normal" src="https://user-images.githubusercontent.com/142636/149594572-6d23e6b4-1510-4b49-a982-885fd43ed8e6.png">

There are some open questions about what the high contrast version of this is, since not all the styles and features directly translate to a high contrast rendering pattern.

### Reproduction Sample

https://esri.github.io/calcite-components/?path=/story/components-controls-datepicker--range

### Reproduction Steps

1.  Set the theme to a high contrast accessibility theme on a Windows machine or VM.
2. Open the above sample.
3. Attempt to interact with it via mouse or keyboard.
4. Observe the issues stated above.

### Reproduction Version

@esri/calcite-components@1.0.0-latest

### Relevant Info

OS: Windows 7, 10, or 11 with a high contrast accessibility theme selected
Browser: any (Chrome, Edge, Firefox)

### Source

- [X] CDN
- [X] NPM package

mpriour:
This issue could use some design help, @macandcheese. It might also benefit from some input by an actual high contrast user like @MehulChoksey.

By moving the `outline-none` tailwind apply rules to the focus state, it solves the major problems of the single date date-picker.
<img width="550" alt="date-picker with better focus rules" src="https://user-images.githubusercontent.com/142636/151053020-1339602d-3227-49b2-8ac0-9b1aa8cd749c.png">

You can see the selected date on the right hand component and the keyboard focused date (in green) on the left hand component in the screenshot above.

However, the date range use case is harder. The number one question would be should the selected range be solid or just outlines. Here are some rough examples (this use case has a lot of classes that potentially need adjustment)

[...]

PR #4011, opened 2022-01-26
fix(date-picker): make date picker work in high-contrast (#4005)

mpriour:
* Move tailwind outline-none rule to focus state, as per tailwind docs. This fixes the date-picker for single date selection

* Add high contrast specific styles for the date range use case. Uses the highlight system color for a filled range indicator.

**Related Issue:** #4005 

### After PR
**Single Date use case**

**Date range use case**


mpriour:
@macandcheese does this look good?

macandcheese (approved):
Apologies for delay here - LGTM, thanks again for your help with these.

cc @asangma

Issue #3572, opened 2021-11-24
Suggest adding configuration for aria-live on calcite-alert


FelFly:
### Summary

While the rendered calcite-alert component does have a `role="alertdialog"` on it, typically I'd expect to use that in conjunction with an [aria-live region](https://www.w3.org/TR/wai-aria-1.1/#aria-live), especially on auto-dismissing alerts. It appears that the component can support this today through code modification, but I didn't see option for configuration on https://developers.arcgis.com/calcite-design-system/components/alert/. At the very least, I might recommend adding notes to the Accessibility section on how this can be manipulated so others don't have to guess.

### Actual Behavior
Out-of-the-box aria-live isn't set as an attribute at all, so the only way to hear the alert in VoiceOver is to navigate to the appropriate place in the DOM (or for the user-agent to move keyboard focus.)

### Expected Behavior
When `role="alert"` or `role="alertdialog"` is present with an aria-live region, it doesn't matter where keyboard focus is in the page, the screen-reader (in my case, VoiceOver) will read out the message either after their current element has been read (eg. "polite") or as an immediate interruption (eg. "assertive").

### Reproduction Steps and Sample

_Sample_: https://codepen.io/FelFly/pen/VwzJXpq

1. Open https://developers.arcgis.com/calcite-design-system/components/alert/ sample codepen
2. Add a line of paragraph text before the calcite-alert
2. Turn on VoiceOver (CMD+F5)
3. Click onto preview pane
4. Ctrl+Option+right-arrow to move to the paragraph text (you should hear it read)
5. Do not move to the next DOM element using keyboard
6. Observed: You will not hear the alert.
7. Open my codepen above, repeat steps 3-5. You should hear the alert read without having to move focus to it. With `assertive` you should hear it right after or in the middle of whatever paragraph text you wrote. With `polite` you may hear additional navigation instructions in between the paragraph text and alert being read.

### Relevant Info

Safari, VoiceOver

_Version_: `@esri/calcite-components@v1.0.0-beta.69`

- [ ] CDN
- [ ] NPM package

macandcheese:
We might also want to address focus trapping in alerts in relation to this.

anveshmekala:
@macandcheese & @asangma  can we default the `aria-live` value to be `polite` and when `autoDimiss` is enabled set it to `assertive` . Any thoughts?

macandcheese:
[...]

PR #4013, opened 2022-01-26
feat(alert): add aria-live support

anveshmekala:
**Related Issue:** #3572 

## Summary

This will add `aria-live` attribute for `calcite-alert` to help the screen readers identify the alert. The default value of aria-live is set to be `polite` and when `autoDismiss` is enabled the value will be `assertive`.


src/components/alert/alert.e2e.ts
```diff
@@ -192,6 +192,19 @@ describe("calcite-alert", () => {
...
+    <calcite-alert placement="top-end" auto-dismiss>
+    ${alertContent}
+    </calcite-alert>`);
+
+    const alert = await page.find("calcite-alert");
+    const priority = alert.getAttribute("aria-live");
+
+    expect(priority).toBe("assertive");
+  });
+
```
driskull, line 207: @anveshmekala can you add another test using the `accessible` test in `commonTests`? 

This will run the HTML in aXe.

anveshmekala: Currently we have a [test](https://github.com/Esri/calcite-components/pull/4013/files#diff-c4f0b7905696c89b9d919db65d83f2fba7f2a25c40b8682fb79ea29fbd19a4a1L14) to check accessibility . Do we need to add another?

driskull: Yeah, that doesn't run the HTML through aXe. I think we need to do that to verify that it is accessible.

anveshmekala:
closing as the issue is non-reproducible in safari.

PR #4017, opened 2022-01-27
docs(tooltip): use the correct component name

benelan:
**Related Issue:** NA

## Summary
Looks like these were copy and pasted from `popover` and the component name wasn't changed


driskull (approved):
😱 👍

Issue #3956, opened 2022-01-19
calcite-input type="number": Incorrect negative values in some locales


jmanke:
### Actual Behavior

When typing a negative number in certain locales (et, ar), the `calciteInputInput` event always has the value set as `-1`. Additionally, when focusing off of the input, the value is reset to `-1`.

### Expected Behavior

The value emitted should reflect the value typed into the `calcite-input` since this would be consistent with other locales such as `en`.

### Reproduction Sample

https://codepen.io/jmanke/pen/mdBgawv

### Reproduction Steps

1. Use Estonian "et" locale
2. Type a negative number such as '-123'
3. Click off of the input (cause input to lose focus)
4. Notice value resets to "-1". The input also only outputs a value of "-1" on emitted events.

### Reproduction Version

@esri/calcite-components@next

### Relevant Info

I didn't try every locale, but 'et' and 'ar' exhibit this behavior.

Last working version: `@esri/calcite-components@1.0.0-beta.72`

### Source

- [X] CDN
- [ ] NPM package

benelan:
In `beta.72` when typing `-123` you can see in the console that the value is `NaN`.  Since then, I made changes to invalid number sanitation, which is a little funky in this case due to an existing bug. I'll look into why `-123` is `NaN` in those locales.

benelan:
- `In beta.55`, `calciteInputChange` started emitting `-1` instead of `-123`. [`beta.54`](https://github.com/Esri/calcite-components/blob/master/CHANGELOG.md#100-beta54-2021-04-22) correctly emits `-123`.
- `In beta.72`, `calciteInputInput` started emitting `NaN` instead of `-123`. [`beta.71`](https://github.com/Esri/calcite-components/blob/master/CHANGELOG.md#100-beta71-2021-12-07) correctly emits `-123`.

jmanke:
@benelan It looks like the specific issue I logged above is fixed; however, there are some other issues with negative numbers with the same locales. Please see this [codepen](https://codepen.io/jmanke/pen/gOXPQZd?editors=1011). I can make another issue if it's not related to this one.

benelan:
This is a different problem, it looks like the number isn't being delocalized correctly before being set as the value.  `e.target.value` should be delocalized to `1.5`. Right now it is `1,5` which is causing issues when converting it to a number. Can you please create a new issue, and I'll take a look?

jmanke:
Sounds good, thanks @benelan 🙂

[...]

PR #4020, opened 2022-01-27
fix(input): correct negative values in ar and et locales

benelan:
**Related Issue:** #3956

## Summary
- `-123` was considered an invalid number in `ar` and `et` locales. 
- The issue started when adding the `componentShouldUpdate` lifecycle method in [this PR](https://github.com/Esri/calcite-components/pull/2036). 
- It looks like there was [no number sanitation at the time of that PR](https://github.com/Esri/calcite-components/blob/c0cc3ed5b6334282e5f7a4b415030331dae5f3c9/src/utils/number.ts). Can you think of any use cases where this method is needed @eriklharper? It seems like the new(ish) number sanitation covers it. 
- The existing tests passed locally and the functionality was the same in the demos. The test I added fails on master but passes on this branch.

### DIFF
--- a/src/components/calcite-input/calcite-input.e2e.ts
+++ b/src/components/calcite-input/calcite-input.e2e.ts
@@ -1084,6 +1084,18 @@ describe("calcite-input", () => {
       });
   });
 
+  it(`allows negative numbers for ar locale`, async () => {
+    const value = "123";
+    const page = await newE2EPage();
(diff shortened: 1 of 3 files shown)


eriklharper (approved):
LGTM!  I tested the programmatic setting of invalid number values and it looks to be working fine.

Issue #4009, opened 2022-01-25
popover's internal hiding of overflow trips up positioning when nested


jcfranco:
### Actual Behavior

Nesting popovers will demonstrate issues with positioning. See the repro case for more details.

### Expected Behavior

Nested popovers to be placed properly.

### Reproduction Sample

https://codepen.io/jcfranco/pen/mdqdKzz?editors=1000

### Reproduction Steps

1. Click on the buttons from the already open parent popover
2. Notice the behavior (each configuration has a description)
3. Once all tested, click on the bottom button that will hack the popover's internal styling (fix suggested by @driskull 🎉)
4. Retest
5. Notice how all configurations work now

### Reproduction Version

@esri/calcite-components@1.0.0-beta.75

### Relevant Info

Regression? Last working version: `@esri/calcite-components@1.0.0-VERSION`

The `overflow: hidden` styles were introduced by the following installs, so we we need to make sure we don't introduce regressions:

https://github.com/Esri/calcite-components/commit/bebd22e41a49e1d29c2bfbc15e9d11f93e6cefc9
https://github.com/Esri/calcite-components/commit/6e4c81de6e560094ca7ff2a5cf68aed78ca2a32c

### Source

- [ ] CDN
- [X] NPM package

driskull:
I can get this to work by using the word break mixin instead of overflow:hidden.

However, the action button used for the close button doesn't support rounded corners. So we would likely have to find some way to support that.

![image](https://user-images.githubusercontent.com/1231455/151459348-89178839-6149-4e58-8b8b-359827bff186.png)

benelan:
I'll be working on #3152 which will create a global css var for border radius. So we will have precedent and the ability to use it here if design approves

driskull:
@jcfranco @benelan @asangma do we even want to allow nested popovers? That seems like not the best UX.

benelan:
Yeah I agree it's a bit weird. @jcfranco do you have a valid use case for it?

AdelheidF:
There are plenty of use cases in the map viewer and they are not weird.
E.g. shell right panel opens a popover for some more complicated setting like symbol styler, filter, .... Then a small popover opens from the first popover for selecting a color, a field, some other single values.

macandcheese:
> @jcfranco @benelan @asangma do we even want to allow nested popovers? That seems like not the best UX.

[...]

PR #4021, opened 2022-01-27
fix(popover): Allow nesting of popovers

driskull:
**Related Issue:** #4009

## Summary

fix(popover): Allow nesting of popovers. #4009

Break words instead of hiding overflows of content.


driskull:
@asangma @jcfranco can you review this one?

src/demos/popover.html
```diff
@@ -103,13 +103,38 @@ <h1 style="margin: 0 auto; text-align: center">Popover</h1>
...
+        <style>
+          .info {
+            padding: 10px;
+          }
+
+          .header {
+            font-weight: bold;
+          }
+        </style>
+        <div class="info">
```
jcfranco, line 122: Can this be converted into a nested-popover story?

jcfranco (approved):
Code LGTM! I'll defer to @asangma on the CSS.

src/demos/popover.html
driskull: Yeah good idea. I'll add one.

src/components/popover/popover.scss
```diff
@@ -67,11 +66,16 @@
   flex-col
   flex-nowrap
   self-center
-  overflow-hidden
+  word-break
   h-full
   w-full;
 }
 
+.close-button {
```
driskull, line 74: @asangma these were added so that the close button doesn't mess with the rounded corners of the popover.

driskull:
Story added, Waiting on CSS feedbackz

src/components/popover/popover.scss
asangma: 👀

asangma: We might be able to remove those margins if we added a div around Action.close-button, gave that wrapper appropriate border-radius (for RTL and LTR) and `overflow: hidden`.

The wrapper would need to be flex and probably `flex: 0 0 auto`

Too much?

image

driskull: I'll give it a shot. If I can't get it to work can I hand it off to you?

asangma (approved):
Beaut!

asangma:
re: Screener
I dunno what's up but it looks like it's not showing the rounded corners on the close buttons.

driskull:
> I dunno what's up but it looks like it's not showing the rounded corners on the close buttons.

Yeah, this is because its not screener 97 so it doesn't support those CSS logical properties. It should be good in local tests.

@jcfranco @eriklharper more reason to get screener bumped.

asangma:
Ah...got it.

Issue #3953, opened 2022-01-18
Action : Update story and Screener test


asangma:
### Description

Update the story for screener to align with the updates in #3929.
Most screenshot tests should use a container without a defined width.

We will probably need defined widths on containers for `text-enabled` examples.

cc @Elijbet 

### Acceptance Criteria

Accurately check for changes in Action.

### Which Component

Action

asangma:
@benelan I discussed this with @Elijbet. She has the understanding to take it on.

asangma:
@Elijbet I updated the first comment.

Elijbet:
I checked if in the latest release for any odd alignment w/ `text-enabled` +`indicator` configurations with or without container width specified for the following scenarios:

- action by default
- action + text-enabled
- action + indicator
- action + text-enabled-indicator

I didn't see any issues, so I think the issue is no longer valid.

geospatialem:
Closing out the above per [Eliza's comment above](https://github.com/Esri/calcite-design-system/issues/3953#issuecomment-1707483546) ☝🏻 .

PR #4022, opened 2022-01-28
fix(action): updated story and screener

Elijbet:
**Related Issue:** #3953

## Summary

Updated story and screener, as some screenshot tests should use a container without a defined width.


driskull:
Should we close this stale PR?

jcfranco:
@Elijbet is this still valid?

PR #4024, opened 2022-01-28
ci(screener): upgrading to Chrome 97

eriklharper:
**Related Issue:** n/a

## Summary

This pr is a second attempt at upgrading Chrome to 97 using the Sauce Labs integration.  This PR will be used as a test case for working with Sauce Support on resolving issues with this setup.

UPDATE: This pr now includes a dual project setup where the existing "legacy" setup that uses the older version of Chrome will continue to be used as the configuration that will fail the check alongside a separate configuration that points to a second Screener project using the upgraded Chrome 97 setup with Sauce Labs integration.

The standard Screener configuration that we have been using will from this point on be referred to as the "Chrome 87" configuration since that is the last reported version that I was told Screener's built-in Chrome browser was set to.  This means that the contents of the old `screener.config.js` file have been moved to `screener-chrome-87.config.js` which points to a new Screener project called `Esri/calcite-components-chrome-87` which we will continue to use to fail the build while we diagnose issues with the newer Sauce setup.

The new Sauce labs configuration will use the existing `screener.config.js` file so that we can eventually migrate without having to change the name of the config file.  This points to the existing Screener project `Esri/calcite-components` that we all know and love.  All the screenshots in this project will need to be accepted as new since it is now running the visual tests on Chrome 97.  This configuration will NOT fail the build in order that test run failures and other issues that might arise don't get in the way of merging existing PRs.  The plan is to eventually remove the `screener-chrome-87.config.js` once this Sauce setup stabilizes and produces consistent results.

Due to this change, every PR build will show two screener checks, both in Github Actions as well as in Screener itself:

image

image

For the time being, we'll review the snapshots in both builds, but the `Esri/calcite-components-chrome-87` will be the only one that will affect the mergeable status of the PR.


eriklharper:
We're going to keep Screener as-is for now.

Issue #3152, opened 2021-10-04
Enhancement: Calcite-Card style options for Rounded corners & shadow


amandathomas-90:
### Description
On the Hub team, we're exploring a redesign of our components - several of which utilize calcite cards. I specifically wanted to alter the appearance of the `<calcite-card-container>` but it seems that we do not have that ability. Is this something that could potentially be doable in the future?

I'm happy to clarify this further if needed.

### Acceptance Criteria 
Ability to edit appearance of `<calcite-card>`

- Optional Rounded corner (boolean or radius)
- Optional Drop Shadow (boolean or enumeration like low,medium,high)

#### Which Component

`<calcite-card>`

#### Example Use Case

![calcite-comp-request](https://user-images.githubusercontent.com/49413021/135888300-4670d69c-f5dd-46e8-9f11-0d76871f3cb3.png)

benelan:
@macandcheese / @asangma - thoughts on the shadow half of this issue?

ajturner:
@benelan can you clarified what _installed_ means in this project context? Is there a PR or documentation we can refer to for starting to use this great enhancement?

benelan:
Installed means that it is available on `next` and will be released in `beta.77` next week! There is no documentation yet, but here is a small sample: https://codepen.io/benesri/pen/JjOJMRB?editors=1100

Please note that this should not be used in Esri products. It is for end users who need to customize the component to meet their own design system's criteria.

eriklharper:
Verified on next

mjuniper:
Thanks for this - it's working great. Have you given any thought to the box-shadow part?

macandcheese:
> Thanks for this - it's working great. Have you given any thought to the box-shadow part?

I think it's something we'll eventually want to support as part of the general theming story but not at this time. cc @asangma

PR #4026, opened 2022-01-29
feat(card): use css variable to change border-radius

benelan:
**Related Issue:** #3152

## Summary
~Use the existing `--calcite-border-radius` css variable in `calcite-card` so users (non-esri) can change it.~

As @macandcheese  suggested, I created a new global `--calcite-border-radius-base` css variable for components that start at our base `0px`, but that we want to allow users to change. This follows the level paradigm we use with other css vars.


src/components/calcite-card/calcite-card.scss
```diff
@@ -1,4 +1,5 @@
 :host {
+  --calcite-border-radius: 0;
```
driskull, line 2: Should this variable be named to reflect its just for calcite-card? Otherwise it seems like a global calcite-border-radius variable that would apply to other components.

driskull: Should we document this variable in the readme? See the `calcite-alert` readme. https://github.com/Esri/calcite-components/blob/master/src/components/calcite-alert/readme.md#css-custom-propertiesw

driskull (commented):
Comments!

src/components/calcite-card/calcite-card.scss
benelan: `--calcite-border-radius` is a preexisting global variable. It's used in modal etc

benelan: Yeah I think it would be nice to document, not sure where though since its global. Did you document the icon color one somewhere?

driskull: Interesting, I thought we documented all of them in the theme story. 

`6-custom-theme.stories.mdx`

All those global ones start with `--calcite-ui-`

Maybe we need to improve this documentation? A readme for global variables and consistent naming.

driskull: It would also be beneficial to users if we let them know what variables are being used in a component. Global and non global.

driskull: @jcfranco what do you think? ^

It seems like we need to document global variables and local variables better. It might also help to standardize the naming between them.

- global: `calcite-ui-`?
- local:  `calcite-<component>-`?

macandcheese: On the developer site, we currently just show component-scoped variables: https://developers.arcgis.com/calcite-design-system/components/modal/#component-api-styles. I have a WIP feature that will let users override these in the component preview as well.

I do think it will be useful to also show "global css variables that can affect this component" so users can tell what effect global overrides will have. Right now it can be a bit of a guessing game as a consumer to determine which border or text color to override.

benelan: I agree with both of you. If a user can use a global css var on a component, it should be documented on that component's page. We just haven't done that yet so I think we should be consistent and do them all at once. 

I will create an issue to audit which components use which global variables. And then we can add them side by side with the component-specific vars, or did you want to create a different doc section on the component pages for the global vars?

macandcheese: > And then we can add them side by side with the component-specific vars, or did you want to create a different doc section on the component pages for the global vars?

It would be nice to assign them some distinguishing name / data so we can group them or otherwise indicate the difference in the docs.

macandcheese: @benelan I totally edited your comment instead of mine - sorry 😆

benelan: So I could refactor this one to the `-ui` pattern used by all of the other global ones. I don't think having them in a completely separate section on the doc page is necessary.

benelan: This is probably something we should discuss in our meeting next week, I'll add it to the agenda

driskull: > It would be nice to assign them some distinguishing name / data so we can group them or otherwise indicate the difference in the docs.

I think it would be super nice to easily know whether a variable is local to a component or global. I don't think we do anything like this right now.

macandcheese (approved):

Issue #3761, opened 2021-12-21
Race condition setting `for` click handler for labelable components


nwhittaker:
### Actual Behavior
If the rendering of a `<calcite-label>` element (or its `for` attribute) occurs after the assoc. input component is rendered, clicking the label does not focus or otherwise affect a "click" event on the input.

### Expected Behavior
Clicking the label should always affect a "click" event on whatever input is referenced via the label's `for` attribute, because it's consistent with native browser behavior.

### Reproduction Steps and Sample

_Sample_: https://ember-twiddle.com/ba0427f707a2227959a2406f1acb5f16

1. Visit the twiddle.
2. Wait for the `<calcite-switch>` element to render.
2. Click the "Show/Hide labels" button.
3. Click the "Calcite label" label and notice the switch is not toggled.

### Relevant Info
In practice, this bug is more likely to occur when using a dynamic templating system (e.g. Ember) where it's harder to control when a given element or attribute might get pushed to the DOM.

_Reproducible version_: `@esri/calcite-components@1.0.0-next.338`

- [ ] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-<version>`

Elijbet:
Did some additional testing and confirmed it also occurs outside of Ember.

https://user-images.githubusercontent.com/19231036/150437490-0d591a7f-6c37-42d5-ad43-cfe20316a5a6.mov

benelan:
@Elijbet can you post a codepen for that repro sample please?

Elijbet:
> @Elijbet can you post a codepen for that repro sample please?

https://codepen.io/elijbet/pen/XWeLvWW?editors=1010

jcfranco:
Verified with repro case. Great job, @Elijbet!

PR #4029, opened 2022-01-31
fix(label, input, label util): clicking the label always triggers a "click" event, whether the label was added before component was added or after

Elijbet:
**Related Issue:** #3761

## Summary

Clicking the label should always trigger a "click" event, whether the label was added before component was added or after. This requires some logic around different scenarios of adding or removing either.  

- set up a listener to document for when new label elements might be initialzed. This is done in label util, which is called when component is initialized or destroyed
- emit event that label has been initialized on connectedCallback in label component
- keep track of the components missing an assigned label as a Set "unlabeledComponents"
- when connectLabel() is called and no label is found, the component is to the Set
- when onCalciteLabelConnected is emitted, it checks the Set to see if the component is in the Set. If it is, then it will try calling connectLabel().
- the event listener is only added to the component if its label can't be found, otherwise the listener is removed
- once a label is found for the component, the component is removed from the Set
- when label is removed, dispatch calciteInternalLabelConnected event to body, so we can listen to it being reattached


src/utils/label.ts
```diff
@@ -87,20 +88,54 @@ function hasAncestorCustomElements(label: HTMLCalciteLabelElement, componentEl:
...
+      labelListenerMap.delete(component);
+    }
+
+    console.log("componentLabelAddedEventListener", componentLabelAddedEventListener);
+  }
+
+  if (!labelEl) {
+    const componentLabelAddedEventListener = labelListenerMap.get(component);
+    if (!componentLabelAddedEventListener) {
+      const labelAddedListener = () => {
```
driskull, line 118: I think we should use binding similar to how the other listener is done above.

```
const boundOnLabelClick = onLabelClick.bind(component);
```

src/utils/label.ts
```diff
@@ -87,20 +88,54 @@ function hasAncestorCustomElements(label: HTMLCalciteLabelElement, componentEl:
...
+        connectLabel(component);
+      };
+      document.addEventListener("labelAdded", labelAddedListener);
+      labelListenerMap.set(component, labelAddedListener);
```
driskull, line 122: I don't think we need to store the listener because we know what the listener function will be.

driskull (commented):
Looking good! Just had some comments and cleanup! 🎉

src/components/label/label.tsx
```diff
@@ -62,6 +62,8 @@ export class Label {
     sourceEvent: MouseEvent;
   }>;
 
+  @Event() calciteInternalLabelConnected: EventEmitter;
```
driskull, line 65: Lets call this `calciteInternalLabelRegister` to be consistent with the other event. 👍

src/components/label/label.tsx
```diff
@@ -80,6 +82,15 @@ export class Label {
   //
   //--------------------------------------------------------------------------
 
+  connectedCallback(): void {
+    this.calciteInternalLabelConnected.emit();
+  }
+
+  disconnectedCallback(): void {
+    // Dispatching to body in order to be listened by other elements that are still connected to the DOM.
+    document.body?.dispatchEvent(new CustomEvent("calciteInternalLabelUnregister"));
```
driskull, line 91: maybe this could import the const `labelDisconnectedEvent` from the `label.ts`?

src/components/label/label.tsx
driskull: Since this event is going to be internal, I think we need to either...

- Mark it as `@internal`
- Use body.dispatchEvent like disconnected event.

Which do you prefer?

src/utils/label.ts
```diff
@@ -87,20 +90,44 @@ function hasAncestorCustomElements(label: HTMLCalciteLabelElement, componentEl:
...
-  onLabelClickMap.set(component.labelEl, boundOnLabelClick);
-  component.labelEl.addEventListener(labelClickEvent, boundOnLabelClick);
+  if (labelEl) {
+    addClickEventListenerToComponentLabel();
+    unlabeledComponents.delete(component);
+    document.removeEventListener(labelConnectedEvent, boundOnLabelConnected);
+    document.body.addEventListener(labelDisconnectedEvent, boundOnLabelDisconnected);
+  }
+
+  if (!labelEl) {
```
driskull, line 114: I think you can use ` else {` here.

src/utils/label.ts
```diff
@@ -87,20 +90,44 @@ function hasAncestorCustomElements(label: HTMLCalciteLabelElement, componentEl:
...
 
-  component.labelEl = labelEl;
-  const boundOnLabelClick = onLabelClick.bind(component);
-  onLabelClickMap.set(component.labelEl, boundOnLabelClick);
-  component.labelEl.addEventListener(labelClickEvent, boundOnLabelClick);
+  if (labelEl) {
+    addClickEventListenerToComponentLabel();
+    unlabeledComponents.delete(component);
+    document.removeEventListener(labelConnectedEvent, boundOnLabelConnected);
+    document.body.addEventListener(labelDisconnectedEvent, boundOnLabelDisconnected);
```
driskull, line 111: Maybe we should use `document.body?` just incase.

src/utils/label.ts
```diff
@@ -87,20 +90,44 @@ function hasAncestorCustomElements(label: HTMLCalciteLabelElement, componentEl:
...
 /**
- * Helper to tear down label interactions on disconnectedCallback.
+ * Helper to tear down label interactions on disconnectedCallback on labelable components.
  */
 export function disconnectLabel(component: LabelableComponent): void {
+  const boundOnLabelConnected = onLabelConnected.bind(component);
+  const boundOnLabelDisconnected = onLabelDisconnected.bind(component);
+  unlabeledComponents.delete(component);
+  document.removeEventListener(labelConnectedEvent, boundOnLabelConnected);
+  document.body.removeEventListener(labelDisconnectedEvent, boundOnLabelDisconnected);
```
driskull, line 129: `document.body?` just incase.

It looks like the above event just uses `document` and not `document.body`. Maybe we should be consistent here?

src/utils/label.ts
```diff
@@ -87,20 +90,44 @@ function hasAncestorCustomElements(label: HTMLCalciteLabelElement, componentEl:
...
-  component.labelEl.addEventListener(labelClickEvent, boundOnLabelClick);
+  if (labelEl) {
+    addClickEventListenerToComponentLabel();
+    unlabeledComponents.delete(component);
+    document.removeEventListener(labelConnectedEvent, boundOnLabelConnected);
+    document.body.addEventListener(labelDisconnectedEvent, boundOnLabelDisconnected);
+  }
+
+  if (!labelEl) {
+    unlabeledComponents.add(component);
```
driskull, line 115: Just some cleanup...

`function onLabelDisconnected(this: LabelableComponent): void {` uses the same code as below. Maybe it could just be called here?

driskull (commented):
Looking great @Elijbet ! I just have some cleanup and suggestions. And a question or two.

src/utils/label.ts
driskull: I'm not sure if we should emit on the document or the body. Maybe the document is ok?

src/components/label/label.tsx
Elijbet: I think I rather mark it as @internal, since body.dispatchEvent only comes in to solve for not being able to emit on disconnectedCallback.

src/utils/label.ts
Elijbet: I think it should be document, unless the listener should be added to a more specific element.

driskull: document seems best 👍

src/components/label/label.tsx
```diff
@@ -83,12 +87,12 @@ export class Label {
...
 
   connectedCallback(): void {
-    this.calciteInternalLabelConnected.emit();
+    this.calciteInternalLabelRegister.emit();
   }
 
   disconnectedCallback(): void {
     // Dispatching to body in order to be listened by other elements that are still connected to the DOM.
-    document.body?.dispatchEvent(new CustomEvent("calciteInternalLabelUnregister"));
+    document?.dispatchEvent(new CustomEvent(labelDisconnectedEvent));
```
driskull, line 95: nitpick> I don't think the `?` is needed for document. We can remove the question mark here and any other uses.

driskull (approved):
This looks good to me. Lets let @jcfranco look at it now :)

I think the next step is to update the label test to test for when a component is removed or a label is added after component is initialized.

Awesome work @Elijbet 🎉 🌮 💯

Elijbet:
> This looks good to me. Lets let @jcfranco look at it now :)
> 
> I think the next step is to update the label test to test for when a component is removed or a label is added after component is initialized.
> 
> Awesome work @Elijbet 🎉 🌮 💯

Thank you!! :)

I would like to test :
    -  Label click works if label is there before input
    -  Label click works if label is added after input
    -  Label click works if label is removed and added back after the input

src/components/label/label.e2e.ts
```diff
@@ -174,4 +175,39 @@ describe("calcite-label", () => {
...
+      );
+      await page.waitForTimeout(100);
+      await page.click("calcite-label", { clickCount: 1 });
+    });
+
+    it("works when label is rendered after input", async () => {
+      const page = await newE2EPage();
+      await page.setContent(html` <calcite-switch id="switch"></calcite-switch> `);
+      await page.waitForTimeout(100);
+      await page.evaluate(() => `<calcite-label for='switch'></calcite-label>`);
```
driskull, line 195: @Elijbet this isn't setting anything on the document. You would need to do something like...

```
await page.evaluate(() => {
const label = document.createElement("calcite-label");
label.setAttribute("for", "switch");
document.body.appendChild(label);
});
```

driskull: Should it also test the clicking after page.evaluate is executed?

src/components/label/label.e2e.ts
```diff
@@ -174,4 +175,39 @@ describe("calcite-label", () => {
...
+        html`
+          <calcite-label for="switch">
+            <calcite-switch id="switch"></calcite-switch>
+          </calcite-label>
+        `
+      );
+      await page.waitForTimeout(100);
+      await page.evaluate(() => document.querySelector("calcite-label").remove());
+      await page.waitForTimeout(100);
+      await page.evaluate(() => `<calcite-label for='switch'></calcite-label>`);
```
driskull, line 210: This isn't doing anything to the document either.

src/components/label/label.e2e.ts
```diff
@@ -174,4 +175,39 @@ describe("calcite-label", () => {
...
+    });
+
+    it("works when label is removed and added back", async () => {
+      const page = await newE2EPage();
+      await page.setContent(
+        html`
+          <calcite-label for="switch">
+            <calcite-switch id="switch"></calcite-switch>
```
driskull, line 203: Should the label already be in the page.setContent? Seems odd to be here and also created below.

src/components/label/label.e2e.ts
```diff
@@ -174,4 +175,39 @@ describe("calcite-label", () => {
...
+        (node) => ((node as HTMLElement).innerHTML = `<calcite-switch></calcite-switch>`)
+      );
+      await page.waitForTimeout(100);
+      await page.click("calcite-label", { clickCount: 1 });
+    });
+
+    it("works when label is rendered after input", async () => {
+      const page = await newE2EPage();
+      await page.setContent(html` <calcite-switch id="switch"></calcite-switch> `);
+      await page.waitForTimeout(100);
```
driskull, line 194: I don't think the timeout is needed here.

src/components/label/label.e2e.ts
driskull: Same as above comment.

driskull (commented):
Comments :)

src/components/label/label.e2e.ts
driskull: Maybe after page.evaluate you can use page.find() to make sure the label is there and has the correct for attribute?

src/tests/commonTests.ts
```diff
@@ -338,16 +336,35 @@ export async function labelable(componentTagOrHtml: TagOrHTML, options?: Labelab
...
+  const label = await componentFirstPage.find("calcite-label");
+  expect(await label.getProperty("for")).toBe(`${id}`);
+
+  await assertLabelable({
+    page: componentFirstPage,
+    componentTag,
+    propertyToToggle,
+    focusTargetSelector,
+    shadowFocusTargetSelector
+  });
```
jcfranco, line 366: Do we need a test for the label-first case? 

How about the component-first, label-first cases for wrapped labelables?

src/components/label/label.tsx
```diff
@@ -80,6 +86,15 @@ export class Label {
   //
   //--------------------------------------------------------------------------
 
+  connectedCallback(): void {
+    this.calciteInternalLabelRegister.emit();
```
jcfranco, line 90: Since you're listening on the document, why not dispatch the event from it too (similar to `disconnectedCallback`)?

src/utils/label.ts
```diff
@@ -87,20 +90,41 @@ function hasAncestorCustomElements(label: HTMLCalciteLabelElement, componentEl:
...
-  const boundOnLabelClick = onLabelClick.bind(component);
-  onLabelClickMap.set(component.labelEl, boundOnLabelClick);
-  component.labelEl.addEventListener(labelClickEvent, boundOnLabelClick);
+  if (labelEl) {
+    addClickEventListenerToComponentLabel();
+    unlabeledComponents.delete(component);
+    document.removeEventListener(labelConnectedEvent, boundOnLabelConnected);
+    document.addEventListener(labelDisconnectedEvent, boundOnLabelDisconnected);
+  } else {
+    boundOnLabelDisconnected();
```
jcfranco, line 113: Does this need to be skipped if the labelable is already in `unlabeledComponents`? Otherwise, I think you may end up with multiple listeners in the following scenario:

```html
<calcite-some-labelable id="foo"></calcite-some-labelable> 
<!-- label not found, listens for connected event -->

<calcite-some-labelable id="bar"></calcite-some-labelable> 
<!-- label not found, listens for connected event -->
```



```html
<calcite-label for="foo">foo</calcite-label> 
<!-- added! fires connected event -->

<calcite-some-labelable id="foo"></calcite-some-labelable>
<!-- label found, gets wired up, cleans up listeners -->

<calcite-some-labelable id="bar"></calcite-some-labelable>
<!-- label not found, listens for connected event **again** -->
```

jcfranco (approved):
@Elijbet This is awesome! Great work. 🎉🌮🦖🦃

🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉
🎉🏆🎉🎉🎉🎉🎉🏆🎉🎉🏆🏆🎉🎉🏆🎉🎉🎉🎉🎉🏆🎉
🎉🏆🎉🎉🎉🎉🎉🏆🎉🏆🎉🎉🏆🎉🏆🎉🎉🎉🎉🎉🏆🎉
🎉🏆🎉🎉🏆🎉🎉🏆🎉🏆🎉🎉🏆🎉🏆🎉🎉🏆🎉🎉🏆🎉
🎉🏆🎉🏆🎉🏆🎉🏆🎉🏆🎉🎉🏆🎉🏆🎉🏆🎉🏆🎉🏆🎉
🎉🎉🏆🎉🎉🎉🏆🎉🎉🎉🏆🏆🎉🎉🎉🏆🎉🎉🎉🏆🎉🎉
🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉

Can you confirm if the additional test scenarios I suggested will be an issue? If not, this is good to merge after the last round of changes.

src/components/label/label.tsx
```diff
@@ -58,18 +59,23 @@ export class Label {
   /**
    * @internal
    */
-  @Event({ bubbles: false }) calciteInternalLabelClick: EventEmitter<{
+  @Event({ bubbles: false }) labelClickEvent: EventEmitter<{
```
jcfranco, line 62: This doesn't need to change, please restore the previous name.

src/components/label/label.tsx
```diff
@@ -58,18 +59,23 @@ export class Label {
...
     sourceEvent: MouseEvent;
   }>;
 
+  /**
+   * @internal
+   */
+  @Event() labelConnectedEvent: EventEmitter;
```
jcfranco, line 69: The label is dispatching the events directly on the document, so this can be removed.

src/tests/commonTests.ts
```diff
@@ -338,16 +336,57 @@ export async function labelable(componentTagOrHtml: TagOrHTML, options?: Labelab
...
     focusTargetSelector,
     shadowFocusTargetSelector
   });
-}
 
+  const labelFirstSiblingPage: E2EPage = await newE2EPage();
+  await labelFirstSiblingPage.setContent(`<calcite-label for="${id}"></calcite-label>`);
+  await labelFirstSiblingPage.waitForChanges();
+  await labelFirstSiblingPage.evaluate((componentHtml: string) => {
+    function htmlToElement(html) {
```
jcfranco, line 354: This function isn't reused, so you can inline its content here.

src/utils/label.ts
```diff
@@ -31,10 +31,13 @@ export interface LabelableComponent {
  * Exported for testing purposes only
  * @internal
  */
-export const labelClickEvent = "calciteInternalLabelClick";
+export const labelClickEvent = "labelClickEvent";
```
jcfranco, line 34: Please follow event name conventions. These should use the `calciteInternal` prefix and do not need to end with `Event`.

src/utils/label.ts
```diff
@@ -87,20 +90,42 @@ function hasAncestorCustomElements(label: HTMLCalciteLabelElement, componentEl:
...
+      component.labelEl = labelEl;
+      const boundOnLabelClick = onLabelClick.bind(component);
+      onLabelClickMap.set(component.labelEl, boundOnLabelClick);
+      component.labelEl.addEventListener(labelClickEvent, boundOnLabelClick);
+    };
+    addClickEventListenerToComponentLabel();
+    unlabeledComponents.delete(component);
+    document.removeEventListener(labelConnectedEvent, boundOnLabelConnected);
+    document.addEventListener(labelDisconnectedEvent, boundOnLabelDisconnected);
+  } else {
```
jcfranco, line 111: Can you merge the `else`, `if` statements into a single block?

src/tests/commonTests.ts
jcfranco: After our last discussion, the wrapped component-first (WCF) scenario might not be common, but the wrapped label-first one is. Can you add coverage for this case and add a comment explaining why WCF is not covered? 

Bonus points if you do cover WCF for completeness. :sparkles:⭐:sparkles:

jcfranco (changes requested):
Looking great, @Elijbet! I think this will be good to merge after this next pass.

jcfranco (approved):
Aside from a few comments, this looks good to merge! 🎉 🌮🦖🦃

Great work, @Elijbet!

src/utils/label.spec.ts
```diff
@@ -154,6 +154,78 @@ describe("label", () => {
         expect(labelable.onLabelClick).toHaveBeenCalledTimes(1);
       });
 
+      it("supports being rendered after labelable", () => {
+        document.body.innerHTML = `
+          <fake-labelable id="renderedFirst"></fake-labelable>
+        `;
+
+        const label = document.createElement("calcite-label");
+        label.setAttribute("for", "renderedFirst");
```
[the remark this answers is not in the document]
Elijbet: It appears that setting label.id = `${id}` sets the prop but does not set the attribute (which fails the test), while setting label.setAttribute("for", `${id}`) sets both attribute and the prop.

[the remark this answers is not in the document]
jcfranco: Oh, that's right. I forgot that our components don't hydrate in this spec context, so setting the prop wouldn't reflect (same as setting an extraneous prop on an existing element).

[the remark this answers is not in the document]
Elijbet: Here is a follow up issue on this: https://github.com/Esri/calcite-components/issues/4210

PR #4030, opened 2022-02-01
fix(time-picker): meridiem order matches with input time picker

anveshmekala:
**Related Issue:** #3883 

Fixes the meridiem order of `time-picker` and matches it with `input-time-picker`

Example
0385AB72-9A90-4C5A-AFF7-FC751D5E767A_4_5005_c


src/components/time-picker/time-picker.tsx
```diff
@@ -666,6 +674,32 @@ export class TimePicker {
...
+    if (this.formatParts) {
+      const index = this.formatParts.findIndex((parts: { type: string; value: string }) => {
+        return parts.value === this.localizedMeridiem;
+      });
+      return index;
+    }
+    return 0;
+  }
+
+  private isDefaultLocaleRTL(newLocale?: string): boolean {
```
driskull, line 695: cc @jcfranco 

I don't think we should be having `locale` influence `dir` within components. Only the `dir` attribute should influence `direction` styling. `dir` and `lang` are two separate attributes in HTML that do not influence each other. Having one modify the other can become complex and inconsistent with how native HTML works. 

I think it's fine if we want to say that an element inside a component should always be styled `direction: ltr;`, but having a specific locale specify the direction styling opens up a can of worms of doing different things for different languages and styling becomes unpredictable.

I would push for not having `lang` influence `direction` styling.

src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -360,6 +380,7 @@ export class InputTimePicker implements LabelableComponent, FormComponent {
           aria-haspopup="dialog"
           aria-label={this.name}
           aria-owns={popoverId}
+          dir={this.calciteInputElDir}
```
driskull, line 383: If we do need to do direction styling by language, could you instead try using the lang psuedo selector?

https://developer.mozilla.org/en-US/docs/Web/CSS/:lang


```scss
div:lang(ar) {
  direction: rtl;
}
```

If this works, we wouldn't need to use getElementDir or wait for another render.

driskull (commented):
@anveshmekala @jcfranco 

I had some comments on this language influenced styling stuff.

src/components/time-picker/time-picker.tsx
jcfranco: Fair points. FWIW, `direction` and `lang` were being set together to avoid incorrect display/behavior when one was set, but not the other. We confirmed w/ the translation team that we can assume users will set both when needed, so I think we can simplify this approach.

anveshmekala: Even with the assumption of setting `dir` for `lang` other than default, the current complexity is allowing page direction  for default `lang` and determining default `lang`. If we can able to ignore the direction irrespective of lang that would cut down the complexity.

src/components/time-picker/time-picker.tsx
```diff
@@ -847,7 +861,11 @@ export class TimePicker {
           <span class={CSS.delimiter}>{this.localizedSecondSuffix}</span>
         )}
         {showMeridiem && (
-          <div class={CSS.column} role="group">
+          <div
+            class={CSS.column}
+            role="group"
+            style={{ order: this.getMeridiemOrder() === 0 ? "-1" : "0" }}
```
driskull, line 867: It may be better to use a class here instead of using inline styles.

I would suggest...

```
class={{
[CSS.column]: true,
[CSS.columnFirst:  meridiemOrder === 0,
[CSS.columnSecond: meridiemOrder !== 0
}}
```

src/utils/time.ts
```diff
@@ -190,6 +190,20 @@ export function localizeTimeStringToParts(value: string, locale = "en"): Localiz
   return null;
 }
 
+export function getformatToParts(value: string, locale = "en"): Intl.DateTimeFormatPart[] {
```
jcfranco, line 193: Suggestion: `getTimeParts`?

src/utils/time.ts
```diff
@@ -190,6 +190,20 @@ export function localizeTimeStringToParts(value: string, locale = "en"): Localiz
...
+  if (!isValidTime(value)) {
+    return null;
+  }
+  const { hour, minute, second = "0" } = parseTimeString(value);
+  const dateFromTimeString = new Date(Date.UTC(0, 0, 0, parseInt(hour), parseInt(minute), parseInt(second)));
```
jcfranco, line 198: Could you use a constant instead of creating a new date every time? We're interested in the part order, so the actual time isn't crucial.

src/components/input-time-picker/input-time-picker.stories.ts
```diff
@@ -37,4 +37,21 @@ export const DarkTheme = (): string => html`
   </calcite-input-time-picker>
 `;
 
+export const KoreanLocale = stepStory(
```
jcfranco, line 40: Nice! Can you also add a story for Arabic or Hebrew + RTL?

src/components/time-picker/time-picker.tsx
```diff
@@ -847,7 +861,10 @@ export class TimePicker {
           <span class={CSS.delimiter}>{this.localizedSecondSuffix}</span>
         )}
         {showMeridiem && (
-          <div class={CSS.column} role="group">
+          <div
+            class={{ [CSS.column]: true, [CSS.meridiemStart]: this.getMeridiemOrder() === 0 }}
```
jcfranco, line 865: Since `getMeridiemOrder` is mostly used to determine if it's at the start or not, why not refactor it to something like `isMeridiemAtStart(): boolean`? Disregard this if you think the order number may be used in the near future.

src/components/time-picker/time-picker.tsx
```diff
@@ -593,6 +596,7 @@ export class TimePicker {
         localizedSecondSuffix,
         localizedMeridiem
       } = localizeTimeStringToParts(value, this.locale);
+      this.formatParts = getformatToParts(value, this.locale);
       this.localizedHour = localizedHour;
```
jcfranco, line 600: Can you compute the meridiem order here instead of on every render? Storing the parts may no longer be needed after this.

jcfranco (approved):

src/components/input-time-picker/input-time-picker.stories.ts
anveshmekala: the direction shouldn't be specified as we don't support page direction right?

src/components/time-picker/time-picker.tsx
anveshmekala: it's better to keep is as `getMeridiemOrder( ) ` considering it might be used in future or may be move it to utils ?

src/utils/time.ts
anveshmekala: we cannot use a constant here as few locales have diff meridiem symbols based on the time of the day.

jcfranco: That's good to know. Thanks for the info.

src/components/time-picker/time-picker.tsx
```diff
@@ -666,6 +671,17 @@ export class TimePicker {
     }
   };
 
+  private getMeridiemOrder(formatParts: Intl.DateTimeFormatPart[]): number {
+    const isRTLkind = this.locale === "ar" || this.locale === "he";
```
jcfranco, line 675: Can you change this to `isRtlKind` or `isRTLKind`?

jcfranco (approved):
:boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom:
:boom::bomb::bomb::bomb::bomb::boom::boom::bomb::bomb::boom::boom::boom::bomb::bomb::boom::boom::bomb::boom::boom::boom::bomb::boom::bomb::boom:
:boom::bomb::boom::boom::bomb::boom::bomb::boom::boom::bomb::boom::bomb::boom::boom::bomb::boom::bomb::bomb::boom::bomb::bomb::boom::bomb::boom:
:boom::bomb::bomb::bomb::boom::boom::bomb::boom::boom::bomb::boom::bomb::boom::boom::bomb::boom::bomb::boom::bomb::boom::bomb::boom::bomb::boom:
:boom::bomb::boom::boom::bomb::boom::bomb::boom::boom::bomb::boom::bomb::boom::boom::bomb::boom::bomb::boom::boom::boom::bomb::boom::boom::boom:
:boom::bomb::bomb::bomb::bomb::boom::boom::bomb::bomb::boom::boom::boom::bomb::bomb::boom::boom::bomb::boom::boom::boom::bomb::boom::bomb::boom:
:boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom:

src/components/time-picker/time-picker.tsx
```diff
@@ -666,6 +671,17 @@ export class TimePicker {
     }
   };
 
+  private getMeridiemOrder(formatParts: Intl.DateTimeFormatPart[]): number {
+    const isRTLKind = this.locale === "ar" || this.locale === "he";
```
driskull, line 675: @anveshmekala @jcfranco this is still setting a visual style based on language and not a direction right?. Are we sure we want to do this?

Would it be more preferable to check the direction of the element instead?

---

Separate discussion, but it seems like locale only works correctly if set on the host or document. Would we want to enhance that in the future? Maybe it should use `closest` or `closestCrossShadowBoundary`?

```
@Prop({ attribute: "lang", mutable: true }) locale: string =
    document.documentElement.lang || navigator.language || "en";
```

jcfranco: > @anveshmekala @jcfranco this is still setting a visual style based on language and not a direction right?. Are we sure we want to do this?
> Would it be more preferable to check the direction of the element instead?

I think we need to do this. `Intl.DateTimeFormat#formatToParts` gives us the parts in the the correct order, but there's no notion or info on RTL/LTR from it. The meridiem part gets ordered correctly in the input by the browser based on `dir` ([example](https://codepen.io/jcfranco/pen/oNooOKd)). The time picker can't rely on this since we render each part individually. 

If you have an idea on how to improve this, we can do that in a follow-up PR.

--- 

> Separate discussion, but it seems like locale only works correctly if set on the host or document. Would we want to enhance that in the future? Maybe it should use closest or closestCrossShadowBoundary?

Yeah, I think we'd want to go this route. This came up once the time picker updated the way it determines `lang`.

driskull: Makes sense. Thanks for confirming! :)

driskull (approved):

Issue #4038, opened 2022-02-01
Expose `scale` attribute of calcite-filter to enable using different size inputs


ethanbdev:
### Description

The underlying `calcite-input` and `calcite-icon` both have a scale prop to control the size/height of the control. Exposing these props up through the `calcite-filter` would allow aligning the size of the filter with other components in a layout.

### Acceptance Criteria

A `scale` prop exists on `calcite-filter` that takes `"s" | "m" | "l"`, with "m" as the default.

### Which Component

calcite-filter

jcfranco:
@ethanbdev Thanks for submitting this. Next time, can you ping @benelan or myself to triage accordingly?

Adding @macandcheese for design review.

ethanbdev:
> @ethanbdev Thanks for submitting this. Next time, can you ping @benelan or myself to triage accordingly?
> 
> Adding @macandcheese for design review.

Yes, sorry, apologies. I figured it was a small change but will wait next time.

benelan:
verified on next https://codepen.io/benesri/pen/JjOENQL

PR #4041, opened 2022-02-02
feat(calcite-filter): add scale prop/attribute

ethanbdev:
**Related Issue:** #4038 

## Summary

Add a prop/attribute to modify the scale of the `calcite-filter`. Simply apply this prop to the underlying input and actions to make it scale to different sizes. Added tests to cover.
image


driskull (approved):
This looks good to me!

PR #4042, opened 2022-02-02
chore: prototype dropdown reference element.

driskull:
**Related Issue:** N/A

## Summary

Trabajo prototipo para sir franco


src/components/dropdown/dropdown.tsx
```diff
@@ -326,6 +357,67 @@ export class Dropdown {
   //
   //--------------------------------------------------------------------------
 
+  getReferenceElement(): HTMLElement {
```
driskull, line 360: This could be a shared utility function I think.

src/components/dropdown/dropdown.tsx
```diff
@@ -34,7 +43,7 @@ import { createObserver } from "../../utils/observers";
   styleUrl: "dropdown.scss",
   shadow: true
 })
-export class Dropdown {
+export class Dropdown implements ConditionalSlotComponent {
```
driskull, line 46: Could maybe have a interface to implement `ReferenceElementComponent`?

driskull: We could also add a `referenceElement` property as part of `FloatingUIComponent`

https://github.com/Esri/calcite-components/pull/3836/files#diff-3c419c81de1cc684c4d1d859881782bc87d089da410675b5af3f6c468b95ef13R43

src/utils/popper.ts
```diff
@@ -97,3 +97,33 @@ export function hypotenuse(sideA: number, sideB: number): number {
...
+export interface ReferenceElementComponent {
+  /**
+   * The host element.
+   */
+  readonly el: HTMLElement;
+
+  /**
+   * Reference HTMLElement used to position this component according to the placement property. As a convenience, a string ID of the reference element can be used. However, setting this property to use an HTMLElement is preferred so that the component does not need to query the DOM for the referenceElement.
+   */
+  referenceElement?: HTMLElement | string;
```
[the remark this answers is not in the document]
jcfranco: This pattern is purely for positioning, right? We would still have an issue if a parent container clips its content with a popper. Is this correct?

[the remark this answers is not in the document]
jcfranco: Have you thought about or tested how this pattern would work in a scenario where there are multiple levels of components?

jcfranco (commented):
Looking good. I had a few questions, but we can discuss them in our meeting today.

Issue #3894, opened 2022-01-11
Tabbing between pick-list item content can prevent slotted actions from being tabbed into


jcfranco:
### Actual Behavior

Tabbing between pick-list item content can prevent slotted actions from being tabbed into.

### Expected Behavior

Consistent tabbing through pick list item content.

cc @asangma 

### Reproduction Steps and Sample

_Sample_: https://codepen.io/jcfranco/pen/eYGPRRr

1. Click first item
2. Press Tab (focus shifts to action)
3. Press Shift + Tab (focus shifts back to item content)
4. Press Tab (focus skips action and moves to the next focusable element)

After 4, clicking on the item and tabbing always skips the action

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-73`

- [x] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-73`

benelan:
verified on next

PR #4053, opened 2022-02-03
fix(pick-list): prevent disabling tabbing into slot while in the curr…

y0n4:
…ent element

**Related Issue:** #3894

## Summary
This issue was occurring in the `calcite-pick-list` when tabbing into a slotted element while user is tabbing in the current item (doesn't have to be selected). Somehow only the slotted elements tabIndex was getting disabled so the tab skipped over the slotted elements


jcfranco (commented):
Looking good! Can you add a test to cover this case?

src/components/pick-list-item/pick-list-item.e2e.ts
```diff
@@ -127,4 +127,30 @@ describe("calcite-pick-list-item", () => {
 
     expect(removeEventSpy).toHaveReceivedEventTimes(1);
   });
+
+  it("slot keyboard navigation", async () => {
+    const page = await newE2EPage({
```
jcfranco, line 132: For test stability, can you add a `page.waitForChanges()` after creating the page? Either that or use the following:

```ts
await page.setContent(/* ... */);`
```

src/components/pick-list-item/pick-list-item.e2e.ts
```diff
@@ -127,4 +127,30 @@ describe("calcite-pick-list-item", () => {
...
+      html: `<calcite-pick-list style="width: 400px">
```
jcfranco, line 133: Can you use the html test util so that embedded code is automatically formatted by Prettier?

jcfranco (approved):
Sweet! 🍬

Issue #3979, opened 2022-01-21
Date-Picker - Unable to edit year input


zaramatheson:
### Actual Behavior

Editing the year on calcite date-picker or input-date-picker does not stick. 
![date](https://user-images.githubusercontent.com/50747646/150597345-d97a0731-38ed-4616-8027-b006e8503e19.gif)

### Expected Behavior

The year you enter into the calendar should stick and the calendar should update to reflect that year. 

cc @fra

### Reproduction Sample

https://esri.github.io/calcite-components/?path=/story/components-controls-datepicker--simple

### Reproduction Steps

1. Type the year into the text box on the calendar
2. Click out of the input box and the year will revert back to the original

### Reproduction Version

@esri/calcite-components@1.0.0-VERSION

### Relevant Info

Regression? Last working version: `@esri/calcite-components@1.0.0-VERSION`

### Source

- [ ] CDN
- [ ] NPM package

AdelheidF:
we had an issue before https://github.com/Esri/calcite-components/issues/1557

jcfranco:
@eriklharper Can you look at this for the next milestone?

eriklharper:
Looks like the reason that editing the year by itself doesn't change the value is that the component requires you to click a day on the resulting month that appears when you change the year.  I see the reasoning behind this UX decision, because when you start typing the year, the expectation is that the month display reflects the given month of the year that is present in the box, but because the new value (which is actually the day) hasn't been selected, this is why the value does not update when you unfocus the year box without clicking a day of the month.

It sounds like the desired behavior is to immediately update the value of the date picker as soon as a new valid year is entered in the year box.  Is this correct?

One thing I am noticing that I would consider a bug (or at least a valid UX question) is the month view does not update when typing a new year unless I hit the enter key or tab to focus the right arrow key.  If I click or unfocus the component completely, that resets the view to the last selected date, which makes sense.  I do expect the month view to change immediately as I type, but I would not necessarily expect the value to change purely as a result of typing a new year, because I have not selected a new date, I have only changed the year.

When you start typing a different year than the current value and press enter or tab to unfocus the year input but keep the focus inside the component, it will highlight the corresponding day of the month in gray instead of blue (if it exists) which should tell the user that this is in fact the last selected nominal "day" value, but because it appears inactive it has not been committed or saved yet:

![image](https://user-images.githubusercontent.com/821864/151892697-8a2feacb-3d76-404c-ae78-4a7edb0ef1a4.png)

[...]

PR #4056, opened 2022-02-03
fix(date-picker): update the calendar instantly as the year changes

eriklharper:
**Related Issue:** #3979 

## Summary

This PR changes the date picker behavior to automatically refresh the calendar as a new year is being entered, either via typing it manually or using the up/down keys to increment or decrement.


jcfranco (approved):
Awesome. Can you update the E2E test before merging this?

eriklharper:
> Awesome. Can you update the E2E test before merging this?

I was actually thinking this would be a good candidate for a Screener snapshot test because we can visually see the month changing when the year changes.  Unless you had a specific test in mind that I would need to update or were just hoping to add a new test that verifies it works on input instead of change.

jcfranco:
This is more of a behavioral fix vs visual, so I think E2E would be more appropriate. Off the top of my head, I think the simplest test would be to check that the date value updates as the input changes but before the value is committed (i.e., enter or blur).

eriklharper:
Oh, sorry, I should have clarified the fix here.  The fix is only to refresh the month view when the year changes, not update the stored date value.  That's why I think it will be better as a visual test.  I can write a step story that changes the year to a leap year which will show an extra day in February in the snapshot.  That would be a good one to visually confirm that the change is happening on input rather than pressing enter first or blurring the year.

eriklharper:
I opted to make this an E2E test instead of a screener snapshot due to the lack of an ability to type in the input.

jcfranco (approved):
Awesome! This is good to merge after simplifying the E2E test.

Issue #3877, opened 2022-01-07
calcite-value-list no longer displays selection outline on items


jcfranco:
### Actual Behavior

Selected items have no outline.

#### Before
![Screen Shot 2022-01-07 at 9 06 58 AM](https://user-images.githubusercontent.com/197440/148580420-a7469289-72d6-4594-bf44-9ce57687fe89.png)

#### After
![Screen Shot 2022-01-07 at 9 08 26 AM](https://user-images.githubusercontent.com/197440/148580423-9e35d91b-6ea4-40b2-ae72-6a20370dd9dd.png)

### Expected Behavior

Selected items should have a visual indicator.

### Reproduction Steps and Sample

_Sample_: https://codepen.io/jcfranco/pen/bGojjPj?editors=1000

For single:

1. Click on an item
2. Click outside of the list
3. Notice there is no indicator on the selected item

For multiple:

1. Click on an item
2. Click on another item
3. Note there is no indicator on the two selected items
2. Click outside of the list
3. Notice there is no indicator on any selected item

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-73`

- [x] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-72`

benelan:
verified on next

PR #4057, opened 2022-02-03
fix(value-list-item): change to older tailwind class for outlining a selected item

y0n4:
**Related Issue:** #3877 

## Summary
The new tailwind class `ring` was created in effort to replace the `shadow` class. However the ring class currently has several new custom tailwind properties (ex, `--tw-ring-inset`, `--tw-ring-offset-width`, `--tw-ring-color`) that are not visible when rendered. 

There seems to be an issue with the ring class specifically. Found [here](https://github.com/tailwindlabs/tailwindcss/discussions/7129) and [here](https://github.com/tailwindlabs/tailwindcss/issues/6328). For now just going to go back to using the shadow class just for outlining a selected item.


jcfranco (approved):
LGTM! 🎉

Can you update the description and link to the Tailwind v3 issues you found? We should also create a follow-up issue to restore `ring` when fixed.

jcfranco:
cc @driskull @macandcheese

driskull:
Just keep in mind this: https://v2.tailwindcss.com/docs/upgrading-to-v2#replace-shadow-outline-and-shadow-xs-with-ring-utilities

driskull (approved):
👍 

Looks good. We just gotta upgrade it at some point.

Issue #3234, opened 2021-10-13
Bug: calcite-dropdown does not flip when there isn't enough space in preferred placement


ellenupp:
### Actual Behavior

When a placement is set on a dropdown, it will open in that placement even if it doesn't fit. 
![image](https://user-images.githubusercontent.com/14304581/137219581-1f77fef3-3263-4323-8e8b-259588f6357b.png)

### Expected Behavior

The dropdown should fallback to another placement where it has enough space to render without being cut off (if one exists). 
![image](https://user-images.githubusercontent.com/14304581/137219603-0bc3230f-bbc0-4fb6-9c48-b26194ca6b75.png)

### Reproduction Steps and Sample

_Sample_: https://codepen.io/ellupp/pen/PoKqvJy?editors=1000

1. Click dropdown button and note that dropdown opens upwards and is cut off, even though it has plenty of space below it

### Relevant Info

_Version_: `@esri/calcite-components@beta66`

Appears to be a regression from beta65

driskull:
Seems like this broke it: https://github.com/Esri/calcite-components/pull/3116

ellenupp:
@benelan This is regressed again in beta75. You can repro using the same sample above. I don't have the power to reopen this, would you prefer a new issue?

benelan:
I can repro with the sample above in `beta.75`. This is a [dropdown PR](https://github.com/Esri/calcite-components/pull/3378) that landed for that release, @anveshmekala can you check and see if its related?

driskull:
We need to setup a screener test for this one so it doesn't break again.

benelan:
We might want to add a story for this if it is a common regression

driskull:
I can take this one @anveshmekala

benelan:
verified on next

PR #4058, opened 2022-02-04
fix(dropdown): Reposition floating menu correctly.

driskull:
**Related Issue:** #3234

## Summary

fix(dropdown): Reposition floating menu correctly. #3234

- Makes sure that when a dropdown's placement is "top" but it doesn't have room to open that way, that it opens "bottom".
- Adds screenshot test via story
- Fixes story placement options


macandcheese:
I checked out this PR and it also resolves https://github.com/Esri/calcite-components/issues/4003 💣

jcfranco (approved):
Noice!

Issue #3868, opened 2022-01-06
Refactor combobox max-items to align with dropdown's max-items.


driskull:
### Description 

Apply the resize observer logic done in this PR to the combobox so that they are similarly structured. 

https://github.com/Esri/calcite-components/pull/3378

### Proposed Advantages 

Consistency

### Relevant Info 

https://github.com/Esri/calcite-components/issues/3048

#### Which Component

Combobox

benelan:
verified on next https://codepen.io/benesri/pen/JjOEOKZ

PR #4059, opened 2022-02-04
fix(combobox): Align with dropdown logic and repositioning

driskull:
**Related Issue:** #3868

## Summary

fix(combobox): Align with dropdown logic and repositioning. #3868

- Makes sure when a combobox is placed at the bottom of a page that when it opens it opens "top".
- Adds screenshot test via story
- Refactors positioning and maxItems logic to be in sync with dropdown. (uses resizeObserver)
- Fixes issue where maxItem height was being applied when items where less than maxItems.


driskull:
@jcfranco can you review this one as well?

jcfranco (approved):
Awesome! 

Is this a fix or a refactor? These changes are based on dropdown's fix, so both should probably align.

driskull:
I'll change to fix. thanks

Issue #3890, opened 2022-01-10
Combobox: chip overflows border when 1 row is visible.


driskull:
### Actual Behavior

Combobox: chip overflows border when 1 row is visible.

<img width="469" alt="Screen Shot 2022-01-10 at 1 52 33 PM" src="https://user-images.githubusercontent.com/1231455/148847417-899ea79f-a682-4160-abaf-05f7884b14ed.png">

### Expected Behavior

Chip should not overflow border

### Reproduction Steps and Sample

_Sample_:

1.https://codepen.io/pen?editors=1000
2. Look

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-<version>`

- [x] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-<version>`

driskull:
Plz add a screener test for this as well.

TheBlueDog:
We are seeing this behavior in Field Maps as well.

![Screen Shot 2022-01-21 at 1 26 59 PM](https://user-images.githubusercontent.com/5419306/150581600-7037bb86-e055-4a59-8b28-9587e4fcb8aa.png)

benelan:
Triaging this to the next sprint cc @jcfranco @macandcheese

y0n4:
Hi, the codepen link takes me to a blank editor. Is it alright if you can update with another codepen link please? Thanks!

jgibson02:
I can file this as a new item, but I thought it might be relevant here due to the adjustments to chip margin: should there be margin between the last chip and the combobox placeholder?

y0n4:
The merged pr #4068 also includes the spacing fix from @jgibson02 suggestion

benelan:
verified on next

PR #4068, opened 2022-02-07
fix(combobox): balance the margin to be equal on vertical chip sides

y0n4:
**Related Issue:** #3890 

## Summary
The margin on the vertical sides of the chip in combobox was not even.

Before
Screen Shot 2022-02-07 at 11 48 34 AM
After
Screen Shot 2022-02-07 at 11 46 46 AM

### DIFF
--- a/src/components/combobox/combobox.scss
+++ b/src/components/combobox/combobox.scss
@@ -173,7 +173,7 @@
 }
 
 .chip {
-  margin-block: calc(var(--calcite-combobox-item-spacing-unit-s) / 2) 0;
+  margin-block: calc(var(--calcite-combobox-item-spacing-unit-s) / 2);
   margin-inline: 0 var(--calcite-combobox-item-spacing-unit-s);
   max-width: 100%;
 }


macandcheese:
This will change the height of the combobox - cc @asangma re: other discussions around height.

asangma:
Ideally, the combobox height would match other inputs.
Perhaps the Chip needs an update?

macandcheese:
The chip ends up being prohibitively small in small scale… I think the height should be the same before interaction but if it changes after interaction that’s less of an issue. 

If it’s a multiple select and wraps to multiple lines that will happen either way.

asangma:
@macandcheese That's a good point.

@y0n4 Did you say that if you used like 1px block margins, it'd retain the height?

y0n4:
Hi, pardon my earlier observation. I was looking at it more carefully again I don't think 1px margins will help retain its height. Because I noticed that the chip height is the same height has the standard scale size height (without interaction) so there won't be any extra space to even add margins or else it will actually expand combobox height...

|   | S | M | L |
| ------------- | ------------- | ------------- | ------------- |
| chip height  | 24px  | 32px  | 44px  |
| standard height  | 24px  | 32px  | 44px  |

I do think margins are needed though because when multiple lines occur, there needs to be proper spacing between each chips lined up

macandcheese:
I don’t think the “with chip” needs to be the same height, for now. At rest they will be the same as other controls. But the current implementation appears broken so imo we should proceed with the pr.

jcfranco:
> I don’t think the “with chip” needs to be the same height, for now. At rest they will be the same as other controls. But the current implementation appears broken so imo we should proceed with the pr.

@asangma @macandcheese Can either of you create a new issue for the `with chip` case?

jcfranco (approved):
💥 

Could you add a story that catches this change? Disregard this if the updated story diff was accepted already.

macandcheese:
@jcfranco @asangma - this looks good now, it also resolves the "chip touching placeholder" issue. The height changing after interaction is not a problem, imo ... and is somewhat expected as most cases will wrap with a few chips anyway. There were some other explorations in Teams of an "append underneath" option but this PR solves the immediate issue. Nice work @y0n4 !

Issue #4007, opened 2022-01-25
FAB inside `<form>` unexpectedly causes form submission


nwhittaker:
### Actual Behavior

Clicking a FAB that's inside a `<form>` causes the form to submit which can lead to an unexpected page refresh.

### Expected Behavior

Clicking the FAB does not trigger form submission.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/LYOYmvR?editors=1001

### Reproduction Steps

1. Visit the codepen and click the FAB.
2. Notice the form's `onsubmit` handler prints "submitted" to the console and the preview area redirects to a "Not found" page.

### Reproduction Version

@esri/calcite-components@1.0.0-VERSION

### Relevant Info

The FAB contains a `<calcite-button>` that wraps a `<button>` with `type="submit"`. Changing things such that this resolves to `type="button"` instead may fix the issue.

### Source

- [ ] CDN
- [ ] NPM package

driskull:
> Changing things such that this resolves to type="button" instead may fix the issue.

I think you are correct.

We may also want to set the type to button in these components as well.

![image](https://user-images.githubusercontent.com/1231455/151070440-d463e3ef-aaad-45b1-8bbb-c45c4b8b1762.png)

PR #4070, opened 2022-02-07
fix(fab, color-picker, inline-editable, split-button): Set calcite button type

driskull:
**Related Issue:** #4007

## Summary

fix(fab, color-picker, inline-editable, split-button): Set calcite button type. #4007

- Make sure button types are "button" so forms don't submit.


jcfranco (approved):
🎉

Issue #2814, opened 2021-08-14
Enhancement: Expose property for fallback placements in calcite-dropdown


ellenupp:
### Description
Similar to `calcite-popover`, it would be great if `calcite-dropdown` exposed a property to set fallback (flip) placements. I am using placement "bottom" but when it flips it goes to "top-start" because that's hardcoded as the first fallback. I'd want it to be mirrored and fallback to "top" (centered over dropdown trigger).

### Acceptance Criteria 
Able to configure fallback/flip placement for dropdown

#### Which Component
`calcite-dropdown`

benelan:
verified on next
https://codepen.io/benesri/pen/popwvWG

PR #4073, opened 2022-02-08
feat(combobox, dropdown, input-date-picker): Add flipPlacements property

driskull:
**Related Issue:** #2814

## Summary

feat(combobox, dropdown, input-date-picker): Add flipPlacements property. #2814

Expose property for fallback placements


eriklharper (approved):
LGTM! 👍

driskull:
@jcfranco can you approveth?

src/utils/popper.ts
```diff
@@ -22,10 +23,96 @@ type VariationRtl =
   | "left-leading"
   | "left-trailing";
 
+export type PopperComputedPlacement = ComputedPlacement;
```
jcfranco, line 26: Can you drop `Popper` from this interface to hide the implementation details? 

We may need to introduce another type to replace `PopperPlacement` too (this one is a separate item).

jcfranco (approved):
Aside from the rename comment, this looks good to me! 🎉

src/utils/popper.ts
driskull: Yeah we can rename all of these. Maybe when we move to floatingUI?

Issue #3928, opened 2022-01-13
Combobox with allowCustomValues and single selectionMode


MukeZ:
### Actual Behavior

1. Set allowCustomValues and selectionMode is single to the Combobox.
2. Select a item from Combobox.
3. Type a new value to the Combobox
4. It automatically selects two items displays nothing in the input while the DropDownList is opening
5. Close the DropDownList, it displays the first item rather than the inputted data

You can compare the behaviours when selectionMode is multi. I input a new item in both of them.

single model
![image](https://user-images.githubusercontent.com/9669996/149261380-354200d6-132d-4412-8379-6165ac27b507.png)

multi model
![image](https://user-images.githubusercontent.com/9669996/149261471-4196c291-bbe4-481b-9772-7a7d9d9e3282.png)

This one is for using the single mode but it selected two items
![image](https://user-images.githubusercontent.com/9669996/149261629-29ab6e04-28c1-442d-8b62-415efd130989.png)

### Expected Behavior

It should show the inputted item after user inputted a new item.
It should select only one item when using the single mode after user inputted a new item.

### Reproduction Sample

https://codepen.io/pen/?&editors=100

### Reproduction Steps

1. Set allowCustomValues and selectionMode is single to the Combobox.
2. Select a item from Combobox.
3. Type a new value to the Combobox
4. It automatically selects two items displays nothing in the input while the DropDownList is opening
5. Close the DropDownList, it displays the first item rather than the inputted data

### Reproduction Version

@esri/calcite-components@1.0.0-VERSION

### Relevant Info

Regression? Last working version: `@esri/calcite-components@1.0.0-VERSION`

### Source

- [X] CDN
- [X] NPM package

macandcheese:
Thanks @MukeZ - good catch on this case.

y0n4:
What if `selection-mode` is still single and nothing is selected. Then you add a custom value... does the custom value get auto selected?

MukeZ:
> What if `selection-mode` is still single and nothing is selected. Then you add a custom value... does the custom value get auto selected?

Yes, it will auto selected the new value

PR #4080, opened 2022-02-09
fix(combobox): limit the auto selection for custom new values at single selection mode #3928

y0n4:
**Related Issue:** #3928

## Summary
Previously if `selection-mode = single` and a new custom value gets inputted, the value gets automatically selected even though an existing one is already selected.

Before - new value gets selected with current value and input field will continue to show current value
Screen Shot 2022-02-10 at 9 35 59 AM
Screen Shot 2022-02-10 at 9 33 58 AM

After - new value gets selected and the input field will be filled with new value when list is not showing
Screen Shot 2022-02-09 at 3 15 29 PM
Screen Shot 2022-02-09 at 3 16 15 PM

`selection-mode = single`
- An item is already selected and a custom value gets inputted. The new value will replace the current value and get auto selected
- An item is not selected and a custom value gets inputted. The new value will get auto selected

`selection-mode = multi`
- The new value will get auto selected no matter what


jcfranco:
> selection-mode = single
> An item is already selected and a custom value gets inputted. The new value will not get auto selected

@macandcheese Can you confirm this is the intended behavior? Just my 2 cents, but as a user I would expect my new custom value to be added and for it to be selected.

y0n4:
If the input field is supposed to be filled in with the new value... then I'm guessing the new custom value is supposed to replace the current value?

Issue #3949, opened 2022-01-18
calcite-tabs component not throwing events when used inside a custom web component


RalucaNicola:
### Actual Behavior

I'm using a calcite-tabs component inside of a custom web component. Clicking on the tab titles does not switch them (the same tab remains active all the time).

### Expected Behavior

I would expect that the current active tab switches based on the tab title that the user clicks on.

### Reproduction Sample

https://codepen.io/ralucanicola/pen/mdBvbem?editors=1010

### Reproduction Steps

1. Click on Eligibility and notice that neither the eligibility tab title, nor the tab itself change.

### Reproduction Version

@esri/calcite-components@1.0.0-beta.74

### Relevant Info

Not a regression. Tested on Chrome, macOS Catalina.

### Source

- [X] CDN
- [ ] NPM package

jcfranco:
Looks like the tab-title's [tab change handler](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-tab-title/calcite-tab-title.tsx#L189) needs to check the event's `composedPath` instead. It doesn't work when wrapped in another custom element because the event is retargeted, so the target is not the tab-title, but the wrapping element.

benelan:
verified on next

PR #4114, opened 2022-02-15
fix(tab-title, tab): tab switching will function even when wrapped wi…

y0n4:
…thin a custom external element

**Related Issue:** #3949 

## Summary
When switching tab titles while the tabs component was wrapped within a custom external container, the tab content was unable to target the respectful tab title.

Before (Hello belongs to the description tab)
Screen Shot 2022-02-14 at 12 53 14 PM
After
Screen Shot 2022-02-14 at 12 52 54 PM


jcfranco (approved):
Awesome!

Before merging, can you add a test similar to [this one](https://github.com/Esri/calcite-components/blob/628e9c2357c57dcef8eb506f2a2b3c3e7a192895/src/components/dropdown/dropdown.e2e.ts#L898) to make sure we have it covered?

PR #4115, opened 2022-02-15
fix(popover, tooltip, combobox, dropdown, input-date-picker): Hide popover when the reference element or popover is out of view.

driskull:
**Related Issue:** None

## Summary

fix(popover, tooltip, combobox, dropdown, input-date-picker): Hide popover that escapes reference element's boundary (and so it appears detached) or when the reference element is fully clipped and hidden from view.

- Hides popover that escapes reference element's boundary (and so it appears detached)
- Hides popover when the reference element is fully clipped and hidden from view.
- Demo: https://popper.js.org/docs/v2/modifiers/hide/#demo

### Hide modifier

https://popper.js.org/docs/v2/modifiers/hide/

data-popper-reference-hidden: This attribute gets applied to the popper when the reference element is fully clipped and hidden from view, which causes the popper to appear to be attached to nothing.

data-popper-escaped: This attribute gets applied when the popper escapes the reference element's boundary (and so it appears detached).

### Gif

Feb-15-2022 13-22-02

### DIFF
--- a/src/assets/styles/_popper.scss
+++ b/src/assets/styles/_popper.scss
@@ -78,7 +78,7 @@ $popper-default-z-index: 900;
 
   :host([data-popper-escaped]),
   :host([data-popper-reference-hidden]) {
-    @apply opacity-0 pointer-events-none;
+    @apply opacity-0 pointer-events-none !important;
   }
 }
 
--- a/src/assets/styles/_popper.scss
+++ b/src/assets/styles/_popper.scss
@@ -76,9 +76,9 @@ $popper-default-z-index: 900;
     @include popperAnimActive();
   }
 
-  :host([data-popper-escaped]),
-  :host([data-popper-reference-hidden]) {
-    @apply opacity-0 pointer-events-none !important;
+  :host([data-popper-placement][data-popper-escaped]),
+  :host([data-popper-placement][data-popper-reference-hidden]) {
+    @apply opacity-0 pointer-events-none;
   }
 }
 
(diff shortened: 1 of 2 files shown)


jcfranco:
Could you add more info about this? When does it occur? Also, can you update the E2E to reflect the problem use case instead of manually adding the data attribute?

driskull:
Updated description.

> Could you add more info about this? When does it occur? Also, can you update the E2E to reflect the problem use case instead of manually adding the data attribute?

I could scroll a div so that the reference element is out of view and verify but at that point I think I would be testing popper.js right?

I figured it was simpler to just make sure that the element is hidden when the attribute is applied. The attribute is applied by popper.js.

src/assets/styles/_popper.scss
```diff
@@ -75,6 +75,11 @@ $popper-default-z-index: 900;
   :host([data-popper-placement]) .calcite-popper-anim--active {
     @include popperAnimActive();
   }
+
+  :host([data-popper-escaped]),
+  :host([data-popper-reference-hidden]) {
+    @apply opacity-0 pointer-events-none;
```
jcfranco, line 81: Might also need to test for pointer events along with visibility.

jcfranco (approved):
> I could scroll a div so that the reference element is out of view and verify but at that point I think I would be testing popper.js right?

This is one of those gray area ones. We want to test that our styles for `data-popper-hidden` hide the element since it's not done by popper, but at the same time we want to provide some context on when that happens. The other tricky part here is that the change is to our popper styles and not just popover.

Can you add a comment to the test with some context on this with a link to the `hide` modifier? I think that hits a good balance of context and testing.

Does the fix scope also need to include other components using popper?

Issue #4075, opened 2022-02-08
Date Picker - First day of the week is not localized


zaramatheson:
### Actual Behavior

This was fixed for Russian via this issue: https://github.com/Esri/calcite-components/issues/2291 but it is still an issue for a number of other languages. 

In CS/DA/DE/EL/ES/ET/FI/FR/HR/IT/LT/LV/NL/NB/PL/RO/SR/SV/TR/VI/BS/SL/CA/HU/UK languages, the first day is still Sunday rather than Monday. 

Here is detailed list of languages/locales and their expected First days of the week
https://olga8614.github.io/calendars/

cc @jcfranco @benelan 

### Expected Behavior

First day of the week should be localized. 

### Reproduction Sample

https://codepen.io/zmatheson/pen/ExbWVKb

### Reproduction Steps

1.  Open https://codepen.io/zmatheson/pen/ExbWVKb and note that first day of the week for uk locale is Sunday not Monday. 

### Reproduction Version

@esri/calcite-components@1.0.0-VERSION

### Relevant Info

Regression? Last working version: `@esri/calcite-components@1.0.0-VERSION`

### Source

- [ ] CDN
- [ ] NPM package

macandcheese:
In addition to updating the defaults above, A nice quality of life feature would be to allow configuration of a "week start day" option. 

This is referenced in this example as `start-weekday`: https://bootstrap-vue.org/docs/components/form-datepicker

jcfranco:
@macandcheese Can you submit a separate enhancement ticket for that? It's out of scope for the original request.

jcfranco:
@zaramatheson @olga8614 We have 2 locale files that are missing from https://olga8614.github.io/calendars/: 

* MK – updated according to https://github.com/unicode-cldr/cldr-core/blob/master/supplemental/weekData.json#L153
* HI – left as is (starting on Sunday) according to https://happydays365.org/days-of-the-week/weekdays-in-hindi/#:~:text=The%20first%20day%20of%20the,%E2%80%9CMangalwar.%E2%80%9D%20The%20fourth%20day

Is the above correct?

Olga8614:
@jcfranco I have compiled my list of First Days of the Week based on our Locale Lookup tool, that has a lot of different locales.
You can find it here: https://i18n.esri.com/website/landing
Please go to i18n Data and Tools -> Tools
[...]

PR #4118, opened 2022-02-15
fix(date-picker): set correct first day of week for all locales

jcfranco:
**Related Issue:** #4075 

## Summary

Sets the first day of the week according to https://olga8614.github.io/calendars/ for the following locales:

BS, CA, CS, DA, DE, EL, ES, ET, FI, FR, HR, HU, IT, LT, LV, NB, NL, PL, RO, SL, SR, SV, TR, UK, VI

MK (missing from issue) was updated following https://github.com/unicode-cldr/cldr-core/blob/master/supplemental/weekData.json#L153


benelan (approved):
🧹 📅 ⛏️

PR #4119, opened 2022-02-15
fix(input): Allow "" to clear a number input

timmorey:
**Related Issue:** # none

## Summary

After the changes in https://github.com/Esri/calcite-components/pull/3915, setting `element.value = ''` will no longer clear a `<calcite-input type="number">` that previously had a numerical value.  Setting `element.value = null` does clear the input, but that gives us TS errors since we build with `strictNullChecks: true`.

I added a failing test for this case, and then fixed it with a small change to the input component.

### DIFF
--- a/src/components/input/input.e2e.ts
+++ b/src/components/input/input.e2e.ts
@@ -986,6 +986,15 @@ describe("calcite-input", () => {
       await page.waitForChanges();
       expect(await input.getProperty("value")).toBe("1.008");
     });
+
+    it("allows clearing value with an empty string", async () => {
+      const page = await newE2EPage();
+      await page.setContent(html`<calcite-input type="number" value="1"></calcite-input>`);
+      const input = await page.find("calcite-input");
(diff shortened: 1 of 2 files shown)


jcfranco (approved):
@timmorey Awesome! Thanks for the PR! 🎉

jcfranco:
Added @eriklharper for another set of 👀. 

I think we could merge this w/o the Screener check since it's mostly a functional change. @benelan WDYT?

eriklharper (approved):
Awesome!

timmorey:
@benelan yes, good to merge from my perspective - no further changes planned.  I do see one failing check (PR BOT/milestone), and I'm not sure what that means.

benelan:
> I do see one failing check (PR BOT/milestone), and I'm not sure what that means.

It just adds the PR to the current milestone. It isn't passing since this PR is from a fork, but it's no big deal.

Issue #4076, opened 2022-02-08
calcite-input: Fails to update number programmatically after blur event


noahmulfinger:
### Actual Behavior

When the input is type `number` and you trigger a blur event, the next time you update the value via JavaScript, it doesn't display the correct new value (it looks like it does update the internal state, just not the displayed value).

cc @MonikaKanduri4525 

### Expected Behavior

I would expect the displayed value to always match the value passed in via JavaScript.

### Reproduction Sample

https://codepen.io/noahmulfinger/pen/dyZvGdY

### Reproduction Steps

1. In the codepen, click `Increment value` a few times. Things should work as expected
2. Click on the input to focus and click away to trigger the blur event.
3. Click `Increment value` again. The input will not update.
4. Click `Increment value` once again. The input should now increment by two.

In the code, it looks like the variable `this.internalValueChange` is set to true during the blur event (because `origin` defaults to "internal" during [the `setValue` call](https://github.com/Esri/calcite-components/blob/master/src/components/input/input.tsx#L427) , so the next time the [`valueWatcher` updates](https://github.com/Esri/calcite-components/blob/master/src/components/input/input.tsx#L202-L203), it doesn't call `setValue`.

### Reproduction Version

@esri/calcite-components@1.0.0-beta.76

### Relevant Info

Regression? Last working version: `@esri/calcite-components@1.0.0-beta.74`. I believe it was introduced in this commit: https://github.com/Esri/calcite-components/commit/7d9e545dc841f476e1bbf8c0010e2ef98f122fb2

### Source

- [X] CDN
- [X] NPM package

benelan:
verified on `next`

PR #4125, opened 2022-02-16
fix(input): update number programmatically after blur event

Elijbet:
**Related Issue:** #4076

## Summary

When you trigger a focus or blur event on input type number, it wouldn't display the correct new value the next time you updated the value via JavaScript. Now it will update number programmatically after focus/blur events.

this.internalValueChange needs to always be false except when 1) the value actually changed AND 2) only when triggered by user input in the box or with the keyboard or up/down arrow buttons. What was causing that bug was it was being set to true even though the value didn't actually change.


src/components/input/input.tsx
```diff
@@ -434,7 +434,10 @@ export class Input implements LabelableComponent, FormComponent {
 
   private inputBlurHandler = () => {
     if (this.type === "number") {
-      this.setValue({ value: this.value });
+      this.setValue({
+        value: this.value,
+        origin: "external"
+      });
```
eriklharper, line 440: Using `origin: "external"` here distorts the meaning and intention behind the `external` label which is intended to only apply to programmatic changes to the value originating from external javascript, and not as a result of a user interacting with the input normally.  Because of this, the only place where `external` can be used is from within the `valueWatcher` and ONLY when `this.internalValueChange` is set to `false`.

eriklharper (changes requested):
We should either try removing the call to `setValue` from within the blurHandler or change this line: https://github.com/Esri/calcite-components/blob/72160b96da7cd851742eb7c06f39069151c46c3d/src/components/input/input.tsx#L695 to include a check for if the value actually changed or not like:

```
this.internalValueChange = origin === "internal" && this.value !== newValue;
```
This way the component isn't fooled into thinking the value changed internally when it actually wasn't.

src/components/input/input.e2e.ts
```diff
@@ -1363,4 +1363,30 @@ describe("calcite-input", () => {
...
+
+      await element.callMethod("setFocus");
+      await page.$eval("calcite-input", (e: HTMLInputElement) => e.blur());
+
+      await page.evaluate(() => {
+        const button = document.getElementById("myButton");
+        button.onclick = function () {
+          const el = <HTMLInputElement>document.getElementById("myInput");
+          const currentValue = parseInt(el.value);
+          el.value = `${currentValue + 2}`;
```
jcfranco, line 1384: Do you need to set this via button click? Couldn't you set the value directly on the input since it's already blurred? 

```ts
element.setProperty("value", "3");
await page.waitForChanges();
```

?

src/demos/input.html
```diff
@@ -1418,6 +1418,21 @@
         <calcite-input type="number" name="step-any" step="any" placeholder='step="any"'></calcite-input>
       </div>
     </div>
+
+    <hr />
+    <!-- update number programmatically after blur event -->
+    <div style="margin: 10% 20% 10% 20%">
```
jcfranco, line 1424: Let's remove this and keep testing on the E2E/spec side.

jcfranco (approved):
Aside from a few comments, this looks good! 🎉

eriklharper (approved):
👍

jcfranco (approved):
🎉 🌮:t-rex:🦃

src/components/input/input.tsx
```diff
@@ -692,7 +692,7 @@ export class Input implements LabelableComponent, FormComponent {
         ? localizeNumberString(newValue, this.locale, this.groupSeparator)
         : "";
 
-    this.internalValueChange = origin === "internal";
+    this.internalValueChange = origin === "internal" && this.value !== newValue;
```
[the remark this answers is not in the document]
eriklharper: It could, yes.  I think I wrote this as a one-liner while I was helping Eliza out, but sure it could be a check at the top of the function that just returns when the value isn't changing.

Issue #3284, opened 2021-10-18
Refactor: remove 'childElType' from button and link components.


driskull:
### Description 

Refactor: remove 'childElType' from button and link components.

### Proposed Advantages 

Calculate the child element type during rendering.

### Relevant Info 

https://github.com/Esri/calcite-components/pull/3248#discussion_r730255124
https://github.com/Esri/calcite-components/pull/3222#discussion_r730255535

#### Which Component

- calcite-button
- calcite-link

benelan:
verified on master

PR #4147, opened 2022-02-18
refactor(button,link): remove childElType as state

anveshmekala:
**Related Issue:** #3284 

### Summary

This will remove the existing `childElType` state in `button` & `link` components and determine the `type` in render method.


macandcheese:
Does this still switch the internally rendered component if a href is added after initial render?

anveshmekala:
> Does this still switch the internally rendered component if a href is added after initial render?

yup. It does switch to an `<a>`

jcfranco (approved):
🎉

Issue #3123, opened 2021-09-28
A11y: [calcite-tree] Can't navigate through all tree items using arrow keys


ellenupp:
### Summary
The arrow key navigation in `calcite-tree` does not function as expected and it is not always possible to reach every item using the keyboard alone.

### Actual Behavior
Left arrow key: works as expected
Right arrow key: with focus on expanded parent, right arrow moves focus up instead of to first child
Up/down arrow keys: work fine for first level of tree, but when focus is on a deeper level item, using up/down moves the focus back out to the top level instead of to the element's next sibling. See repro steps for example. 

### Expected Behavior
All items in the tree should be reachable using the keyboard. Up and down arrow keys should move focus between adjacent tree items. Right arrow key should expand a parent container if collapsed, else move focus to its first child. W3C links below for examples. Please contact me for link to a custom tree implementation in dashboards for another example of expected behaviour for navigating between tree items. 

### Reproduction Steps

Code pen sample: https://codepen.io/ellupp/pen/PojWvmR?editors=1100
1. Click tree item "Rich text 2"
2. Press up arrow key. Focus jumps out two levels instead of moving up to previous sibling "Rich text 1".
2. Click tree item "Rich text 1". Press down arrow key. Focus jumps out of current level and down to parent's sibling instead of going to that item's next sibling "Rich text 2".
3. Press right arrow key (with focus now on last "Column"). Focus jumps up to element's first sibling instead of to its first child "Serial chart". It doesn't seem possible to navigate to the children. 

### Working W3C Example/Tutorial 
https://w3c.github.io/aria-practices/#TreeView
https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-2/treeview-2a.html

cc @vijayendranj @julio8a

caripizza:
> Please contact me for link to a custom tree implementation in dashboards for another example of expected behaviour for navigating between tree items.

Thanks for adding this issue @ellenupp, can you please share this link with me when you have a moment?

Elijbet:
It seems like the navigation doesn’t work because it’s handled on the tree item level and is unaware that the current state of the entire tree (and of the nested tree items). It moves up/down siblings only. Looks like the whole keyboard interaction logic needs to be moved a level up so the arrow up/down can move to next node whether it’s a sibling, a parent, or a child. Also so that I can have a property that checks what I’m focused on currently, which will allow me to check if the item is expandable or not so I can figure out the right arrow behavior based on that. 

[...]

PR #4150, opened 2022-02-18
fix(tree-item): Pressing right arrow key focuses first child when the tree-item contains an expanded subtree

eriklharper:
**Related Issue:** #3123

## Summary

This is a partial fix for #3123 that restores the correct right arrow key behavior.  The fix now focuses the first child in an expanded subtree.

### DIFF
--- a/src/components/tree-item/tree-item.e2e.ts
+++ b/src/components/tree-item/tree-item.e2e.ts
@@ -292,4 +292,59 @@ describe("calcite-tree-item", () => {
     expect(checkbox).not.toHaveAttribute("checked");
     expect(isVisible).toBe(true);
   });
+
+  it("right arrow key expands subtree and left arrow collapses it", async () => {
+    const page = await newE2EPage({
+      html: `
+        <calcite-tree>
+          <calcite-tree-item id="cables">
+            Cables
+            <calcite-tree slot="children">
+              <calcite-tree-item id="xlr">XLR Cable</calcite-tree-item>
+              <calcite-tree-item id="instrument">Instrument Cable</calcite-tree-item>
+            </calcite-tree>
+          </calcite-tree-item>
+        </calcite-tree>
+    `
+    });
+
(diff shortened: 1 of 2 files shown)


src/components/tree-item/tree-item.e2e.ts
```diff
@@ -292,4 +292,59 @@ describe("calcite-tree-item", () => {
...
+              <calcite-tree-item id="instrument">Instrument Cable</calcite-tree-item>
+            </calcite-tree>
+          </calcite-tree-item>
+        </calcite-tree>
+    `
+    });
+
+    await page.keyboard.press("Tab");
+
+    expect(await page.evaluate(() => document.activeElement.id)).toBe("cables");
```
jcfranco, line 313: Nitpick: you could return the values you're asserting on with the same `evaluate` callback.

src/components/tree-item/tree-item.tsx
```diff
@@ -345,7 +345,8 @@ export class TreeItem implements ConditionalSlotComponent {
 
         // When focus is on a open node, moves focus to the first child node.
         if (this.hasChildren && this.expanded) {
-          this.el.querySelector("calcite-tree-item").focus();
+          this.el.querySelector("calcite-tree-item")?.focus();
+          e.stopPropagation();
```
jcfranco, line 349: I see that this is an existing pattern in this implementation, but unless it's an internal event, let's avoid stopping propagation of events. It breaks expectations and could lead to other bugs (e.g., end-user expects to handle key event, but doesn't receive it, so they resort to listening on the capture phase, etc...).

jcfranco (approved):
Awesome! Just had one comment regarding stopping event propagation.

src/components/tree-item/tree-item.tsx
eriklharper: I'll try and see if there is a better way to solve this in the coming iterations on this bug.

Issue #2188, opened 2021-05-20
Enhancement: change calcite-label's layout="inline" to use layout="horizontal"


eriklharper:
### Description

To keep terminology consistent with other components, `calcite-label` should use the term "horizontal" to describe laying out slotted content as opposed to "inline".  This probably should also apply to `inline-space-between`, so the correct wording should be `horizontal-space-between`.

### Acceptance Criteria 

`calcite-label`'s `layout` property should support the values `horizontal` and `horizontal-space-between`.

#### Which Component

`calcite-label`

#### Example Use Case

```
<calcite-label layout="horizontal">
   <calcite-radio-button></calcite-radio-button>
   Label
</calcite-label>
```
☝️ renders a radio and the label to the right of the radio (in LTR mode).

asangma:
Hi. Sorry to do this late, but seems like the "inline" terminology is something we should keep since it's more inline (no pun intended) with CSS terminology and stuff.

OK, maybe the pun was intended.

macandcheese:
@eriklharper is this safe to close? It could be brought back up at a later time as a Discussion, if needed.

eriklharper:
> @eriklharper is this safe to close? It could be brought back up at a later time as a Discussion, if needed.

Yeah, we can let this go for now.  If the need to reconsider comes up again, we can create a discussion.

PR #4159, opened 2022-02-24
feat(label): add additional layout of horizontal to replace deprecate…

y0n4:
…d inline

**Related Issue:** #2188

## Summary
For easier clarity on layout prop- "inline" and "inline-space-between" will be changed to "horizontal" and "horizontal-space-between". I kept "inline"/"inline-space-between" still as an option just in case it's already been utilized and to not cause any problems.


asangma:
Please hold on this. I posted a comment on the issue. :)

Issue #3064, opened 2021-09-17
A11Y: [calcite-modal] Focus trapping issue with disable-close-button


BlakeStearman:
**Issue**
When `disable-close-button` is set on the `calcite-modal`, focus trapping doesn't work and the user is allowed to tab out of the modal to other page or browser elements.

I've attached a simple test case illustrating this with `disable-close-button` active. If you simply remove the `disable-close-button` attribute, focus trapping will work.

Discovered in beta 65.

**Sample**
https://codepen.io/rpanichakit/pen/RwgQjzW

**Attachments**
[calcite-modal-focus-trap-issue.zip](https://github.com/Esri/calcite-components/files/7187515/calcite-modal-focus-trap-issue.zip)

caripizza:
@BlakeStearman modal also has a `disable-outside-close` attribute, would that work for you here?

BlakeStearman:
Keep open

BlakeStearman:
Keep open for validation&hellip;

BlakeStearman:
Bumping this to keep open. Would be nice to fix for proper A11Y support.

cc @benelan @paulcpederson

benelan:
still reproducible in `beta.77` using the included sample

nwhittaker:
Noticing this issue in the field-apps-web project, as well.

geospatialem:
verified on `next`  ✅

PR #4160, opened 2022-02-24
fix(modal): focus traps when close button is disabled

anveshmekala:
**Related Issue:** #3064 

 ### Summary
 
 This fix will trap the focus for modal with `disable-close-button`


jcfranco (changes requested):
Great job adding the test! 

Focus trapping is already set up, and this is adding extra code. Can you rework the solution to leverage [`focusFirstElement`](https://github.com/Esri/calcite-components/blob/master/src/components/modal/modal.tsx#L385-L387). This is where the issue lies.

jcfranco (approved):
🎉 🌮 🦖 🦃

src/components/modal/modal.tsx
```diff
@@ -304,7 +304,6 @@ export class Modal implements ConditionalSlotComponent {
   @Method()
   async setFocus(focusId?: "close-button"): Promise<void> {
     const closeButton = this.closeButtonEl;
-
```
jcfranco, line 307: Sorry, I missed this. Can you restore this? It's an unrelated change.

anveshmekala: ahh, my bad. will restore this.

Issue #4158, opened 2022-02-23
calcite-tree: Allow all keydown events to propagate


eriklharper:
### Description

Currently `keydown` events are suppressed inside `tree-item`'s keydown handlers in order to avoid ancestor tree-items from receiving a bubbled keydown event that do not apply to them.  We need to instead allow keydown events to propagate outside the root calcite-tree component, so we'll need to refactor how the keydown behavior is applied.

### Proposed Advantages

The existing keyboard behavior will stay exactly the same, and improve on the accessibility by fixing the bugs outlined in #3123 while also allowing the native keydown events to bubble outside the root tree for developer usage, and to be consistent with other components and our goals to stop preventing native events from bubbling outside our components.

### Which Component

`calcite-tree`, `calcite-tree-item`

### Relevant Info

This refactor stems from #3123

benelan:
verified on beta.80!
https://codepen.io/benesri/pen/XWVXBjg?editors=1010

PR #4167, opened 2022-02-25
refactor(tree, tree-item): allow all arrow keydown events to propagate outside the root tree element

eriklharper:
**Related Issue:** #4158 

## Summary

This PR refactors `calcite-tree` and `calcite-tree-item` to not only fix the keyboard behavior for focusing within trees and subtrees, but it also allows all of the native "Arrow*" `keydown` events to propagate outside the containing root tree, making it more predictable and consistent with native behavior.


src/components/tree/tree.e2e.ts
```diff
@@ -349,4 +349,284 @@ describe("calcite-tree", () => {
       });
     });
   });
+
+  describe("keyboard support", () => {
```
jcfranco, line 353: Thanks for adding these tests! ⌨️ 👀

src/components/tree/tree.tsx
```diff
@@ -230,6 +230,59 @@ export class Tree {
...
   }
 
+  @Listen("keydown")
+  keyDownHandler(event: KeyboardEvent): void {
+    const root = this.el.closest("calcite-tree:not([child])") as HTMLCalciteTreeElement;
+    const target = event.target as HTMLCalciteTreeItemElement;
+
+    if (root === this.el && target.tagName === "CALCITE-TREE-ITEM" && this.el.contains(target)) {
+      switch (event.key) {
+        case "ArrowDown":
```
jcfranco, line 240: Can you call `preventDefault` wherever the component handles the event?

jcfranco (commented):
Looking good! Would it be possible to move all keyboard event handling to the parent tree and not have the navigation/selection code in two different places?

eriklharper:
> Looking good! Would it be possible to move all keyboard event handling to the parent tree and not have the navigation/selection code in two different places?

Yeah, I can create a follow-up refactor issue for that, unless you think its within the scope of this ticket?

jcfranco:
I think we can keep it within this ticket since the code is related (event propagation and navigation code moving to tree), right?

jcfranco (approved):
Sweet! 🍬

jcfranco:
@benelan @anveshmekala @Elijbet could you help do some testing on this to make sure tree is behaving as expected? We're slowly ramping up automated tests for tree/tree-item, so manual testing would be helpful until we have better coverage.

anveshmekala:
> I think we can keep it within this ticket since the code is related (event propagation and navigation code moving to tree), right?

Testing looks good 🥳

Issue #3995, opened 2022-01-24
Bug: calciteTreeSelect emits when opening/closing child trees


benelan:
### Actual Behavior

`calciteTreeSelect` emits when opening/closing child trees, even though no selection is happening

### Expected Behavior

`calciteTreeSelect` should only emit when an item is selected.

### Reproduction Sample

https://codepen.io/benesri/pen/QWOLJPa

### Reproduction Steps

1. Click on the caret to open/close child 2
2. Notice in the console that clicking on the caret emits `calciteTreeSelect`

### Reproduction Version

@esri/calcite-components@1.0.0-beta.75

### Source

- [X] CDN
- [X] NPM package

benelan:
verified on beta.79

PR #4170, opened 2022-02-28
fix(tree, tree-item): calciteTreeSelect only emits when item is selected and not on toggle parent icon

Elijbet:
**Related Issue:** #3995

## Summary

calciteTreeSelect only emits when item is selected and not on toggle parent icon, and a corresponding test on a tree.


jcfranco (approved):
Awesome! 🎉

Issue #1156, opened 2020-10-21
axe reports a11y issues on calcite-slider


driskull:
### Summary

The following violations are reported by axe on `calcite-slider`:

1. [ ] ARIA role should be appropriate for the element (for both single and double thumbs)
2. [ ] Ensures buttons have discernible text

#### Details

```
ARIA role should be appropriate for the element

Ensures role attribute has an appropriate value for the element

[More info...](https://dequeuniversity.com/rules/axe/4.3/aria-allowed-role?application=axeAPI)
#calcite-slider-3343e91d-32b7-2975-96f1-06dfc91cb47d.thumb

Minor
ARIA role slider is not allowed for given element
cat.aria
best-practice
```

```
Ensures buttons have discernible text

[More info...](https://dequeuniversity.com/rules/axe/4.3/button-name?application=axeAPI)
#calcite-slider-3343e91d-32b7-2975-96f1-06dfc91cb47d.thumb

Critical
Element does not have inner text that is visible to screen readers
Serious
aria-label attribute does not exist or is empty
Serious
aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
Serious
Element has no title attribute
Minor
Element's default semantics were not overridden with role="none" or role="presentation"
```

![Screen Shot 2021-12-28 at 11 11 12 AM](https://user-images.githubusercontent.com/197440/147599494-a889cfef-92ca-4a15-97b3-0bd0d606e56c.png)

### Actual Behavior

The above issues are reported by axe.

### Expected Behavior

No more axe violations reported on the slider.

### Reproduction Steps

1. Go to [storybook](https://esri.github.io/calcite-components/?path=/story/components-controls-slider--single-value)
1. Go to the A11y tab.
1. Click on the "Tests completed" button in the bottom right to re-run tests.
1. Toggle a knob.
1. Repeat clicking on "Tests completed"
1. Do for each story and knob

paulcpederson:
@driskull thanks for writing these up, we'll probably need an additional prop like `intl-year` on the date component so we can get a localized string for this. The card will probably need props for strings as well. I can see app devs needing something like "select item {title}" and "deselect item {title}".

driskull:
These could also be tested by...

- using e2e tests with the `accessible` helper.
- using the aXe toolbar to run on storybook or demo pages.

[...]

PR #4173, opened 2022-02-28
fix(slider): aria-role is assigned to appropriate element

anveshmekala:
**Related Issue:** #1156 

## Summary

This fix will assign the `aria-role=slider` to appropriate element `div` and enable the screen readers to identify slider.

### DIFF
--- a/src/components/slider/slider.tsx
+++ b/src/components/slider/slider.tsx
@@ -221,7 +221,6 @@ export class Slider implements LabelableComponent, FormComponent {
         aria-valuemax={this.max}
         aria-valuemin={this.min}
         aria-valuenow={value}
-        // disabled={this.disabled}
         class={{
           thumb: true,
           "thumb--value": true,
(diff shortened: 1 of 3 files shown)


src/components/slider/slider.tsx
```diff
@@ -189,7 +189,8 @@ export class Slider implements LabelableComponent, FormComponent {
     const rightThumbOffset = `${mirror ? maxInterval : 100 - maxInterval}%`;
 
     const handle = (
-      <button
+      <div
```
jcfranco, line 192: Can you share the reasoning behind the `button` to `div` change? As a best practice, we want to have our components use as much semantic HTML as possible.

anveshmekala: The reason for choosing div other than button is the role compatibility. Assigning slider role to button is not allowed but w3 docs doesn't explain the limitations of `role=slider`.

jcfranco: Gotcha. Thanks for the breakdown.

jcfranco (approved):
💥

Issue #3866, opened 2022-01-05
Bug: Action size is wrong when loading


asangma:
### Actual Behavior
When `scale="l"`, the loading state of Action is wrong.

### Expected Behavior
Should be the same size.

### Reproduction Steps and Sample

_Sample_: https://codepen.io/mtb-alan/pen/NWazzLg

1. view the difference between the two actions.

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-<version>`

- [ ] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-<version>`

benelan:
Looks like the actions are now 100% width since `beta.75`, is that by design or a regression @asangma / @macandcheese ?
https://codepen.io/benesri/pen/mdpeQLd

benelan:
@macandcheese mentioned 100% width makes sense for expanded action bars. verified on `beta.80`

PR #4175, opened 2022-03-01
fix(action, loader): adjust the loadable icon size to match regular icon size

y0n4:
**Related Issue:** #3866

## Summary
The loadable icon size was not matching up with the regular icon size in calcite-action even when they were both on the `scale="l"`.  The loadable icon (inline) is originally on the smaller side.

|   | S | M | L |
| ------------- | ------------- | ------------- | ------------- |
| inline loadable height  | 12px  | 16px  | 20px  |
| icon height  | 16px  | 24px  | 32px  |

I've only changed just the large scale from 20px to 24px. Maybe it is worth considering to have the inline loadable height to be on par with the icon height?

Before (icons are not aligned vertically and the icons are different size)
Screen Shot 2022-03-01 at 9 25 00 AM
Screen Shot 2022-03-01 at 9 48 37 AM

After
Screen Shot 2022-03-01 at 9 27 05 AM

cc @asangma


asangma:
The screengrab in your comment looks good.
Seems like something went wrong with running the Screener test though.

asangma (commented):
Yew!

asangma (approved):
Yew!

Issue #4168, opened 2022-02-26
value list should only leave the selected item as tabbable


jcfranco:
### Actual Behavior

Navigating with the keyboard will leave navigated elements as tabbable.

### Expected Behavior

Only the selected item should be tabbable after the list loses focus.

### Reproduction Sample

https://codepen.io/jcfranco/pen/jOavwbX?editors=1000

### Reproduction Steps

1. tab or click the selected item
2. press ⬆️/⬇️ to go to the last item (don't select)
3. press Tab to move focus to the next element
4. press Shift + Tab to tab back in

### Reproduction Version

beta.78

### Relevant Info

N/A

### Regression?

No

jcfranco:
@y0n4 Can you take a look at this one?

benelan:
verified on beta.79

PR #4177, opened 2022-03-01
fix(pick-list): change unselected items to be untabbable when an item is selected

y0n4:
**Related Issue:** #4168

## Summary
Previously, when an item from `value-list` gets selected, and when you tab into that list, the unselected item gets tabbed into first rather than the selected item.

It's now changed so that when an item gets selected, then the rest of the unselected items on the value-list will be un-tabbable so it can tab straight to the selected item

I don't think this was a regression because I noticed that this issue existed even before  beta.70


jcfranco (changes requested):
The added test does not reproduce the problematic scenario. I tested locally and I still see the issue:

value-list-focusable-item

> I don't think this was a regression because I noticed that this issue existed even before beta.70

Good catch. I think I missed a step when testing w/ beta.76. I'll update the issue and codepen.

src/components/pick-list/shared-list-tests.ts
```diff
@@ -201,6 +201,31 @@ export function keyboardNavigation(listType: ListType): void {
...
     });
+
+    it("resets tabindex to selected item when focusing out of list", async () => {
+      const page = await newE2EPage({
+        html: `
+        <calcite-${listType}-list>
+          <calcite-${listType}-list-item value="one" label="One" selected></calcite-${listType}-list-item>
+          <calcite-${listType}-list-item value="two" label="Two"></calcite-${listType}-list-item>
+        </calcite-${listType}-list>
+        <calcite-button name="next-element">next element</calcite-button>
```
jcfranco, line 212: Can you update the test the following way?

1. create HTML page with list HTML only (no `#next-element`)
2. tab twice, assert on `activeElement` (should be `body`)
3. tab back, assert on first list item.

jcfranco (approved):
Awesome, thanks for the fix! 🎉 🌮 🦖 🦃 

This is good to merge once the remaining changes are made.

PR #4181, opened 2022-03-02
fix: solve regression with preact type generation

paulcpederson:
## Summary

When all the files were renamed without the `calcite` prefix, it caused the generated preact types to be incorrect. This led to a lot of typescript errors where we're using a ref pointing to these types as preact was then thinking the CalciteInput was an Input.


jcfranco (approved):
🎉

Issue #3724, opened 2021-12-15
calcite-input-date-picker partially hidden by panel header


AdelheidF:
Instructions:
- open calendar popover
- notice that the top of the calendar will not show on top of the panel header
- if I add more space above the input box it's fine
- if I remove some space above the input box it's fine too and the calendar shows below

<img src=https://user-images.githubusercontent.com/44481292/146238457-b8d1d42c-146a-48b2-be51-21ea6328b5a7.png width=300/>

Reported by Beyondsoft

### Actual Behavior

calendar partially hidden by panel header

### Expected Behavior

calendar should show on top of panel header

### Reproduction Steps and Sample

https://codepen.io/afreitag/pen/gOGmWeg?editors=1000

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-beta.71`

- [ ] CDN
- [ ] NPM package

macandcheese:
I've encountered this with the internally rendered `.header` div `z-index` declaration when other components are positioned over it like dropdown, etc. Unless there's a reason seems like we shouldn't be setting that internally as it causes users to have to use JS to unset that `z-index`.

driskull:
I don't recall about the zindex. Was that added by me?

jcfranco:
[Added way back in the early days](https://github.com/Esri/calcite-app-components/blame/master/src/components/calcite-panel/calcite-panel.scss#L64). 

The demo pages and screenshot tests should help catch any issues from dropping z-index, right?

driskull:
Yeah it should. It might have been from when the header was over the panel?

jcfranco:
FYI, I've submitted a PR for the z-indexing issues.

driskull:
Just to add some clarity to this issue...

The header needs a higher z-index because a menu within the header should go over the content below it.

Any content within the scrolling panel should go under the header.

We will need to figure out floating menus within the scrolling area of the panel.

depken:
We are having a similar issue with calcite-input-date-picker within calcite-modal.
https://codepen.io/jonasdepelsmaeker/pen/XWZLoJd?editors=100

jcfranco:
> We are having a similar issue with calcite-input-date-picker within calcite-modal.

This was recently fixed via https://github.com/Esri/calcite-components/issues/3099. You'll need to [set `overlay-positioning=fixed` on the input-date-picker though](https://codepen.io/jcfranco/pen/vYRdLxW?editors=1000).

[...]

PR #4182, opened 2022-03-03
fix(panel): drop z-index usage for internal component styles

jcfranco:
**Related Issue:** #3724

## Summary

This PR aims to fix layering issues caused by unnecessary z-index usage.


src/components/panel/panel.scss
```diff
@@ -182,5 +180,4 @@
   block
   p-2;
   width: fit-content;
-  z-index: 1;
```
driskull, line 185: Are you sure this one isn't needed? If slotted user content has a z-index it might appear over this.

driskull (approved):
👍 

Looks good.

I guess if users find their slotted content appearing over this then we can put it back.

I recall having some of these for Firefox, should we give it a check?

src/components/panel/panel.scss
jcfranco: I _think_ so. Currently, It looks like the [slotted content and the FAB containers don't overlap](https://codepen.io/jcfranco/pen/rNpMerg?editors=1010). Do you know if there's another configuration where they do?

driskull: Yes, they can overlap like so:
https://codepen.io/driskull/pen/MWrjxZa?editors=1000

However, seems like a z-index messes it up currently anyway so might not be an issue.

In most cases they will overlap since a height is defined.

src/components/panel/panel.scss
```diff
@@ -70,7 +70,6 @@
 :host([loading]) .container,
 :host([disabled]) .container {
   @apply relative;
-  z-index: 1;
```
driskull, line 73: this might break this case:
https://codepen.io/driskull/pen/yLpawmY?editors=1000

src/components/panel/panel.scss
driskull: I think these were here mainly to make sure that the slotted content is always below the header/fab/scrim/etc.

src/components/panel/panel.scss
jcfranco: Good catch. I'll revisit this one.

src/components/panel/panel.scss
jcfranco: Thanks for testing and the example. I suspected I was missing something.

src/components/panel/panel.scss
jcfranco: Started looking at this one today and I have a question about the codepen above. What is the use case for setting `position: relative` and `z-index` for the elements in the panel content?

driskull: Its to illustrate that users might have content with a z-index set and we don't want their content to appear over the panel.

jcfranco: This will still be an issue regardless of what we do in the component, right? Users can always modify stacking contexts and set `z-index` to a value higher than we expect ([block example](https://codepen.io/jcfranco/pen/ExobaXe?editors=1000)). 

Sidebar: we could introduce a set of tokens to represent different levels of z-indices used in the design system. This could help our components use the same index depending on the context and better inform users on how to override these values. Some resources:

https://www.lightningdesignsystem.com/design-tokens/#category-z-index
https://designsystem.morningstar.com/legacy/v/2.4.0/visual-language/layering.html
https://www.duetds.com/tokens/#z-index
https://github.com/dojo/widgets#z-index-layering
https://www.smashingmagazine.com/2021/02/css-z-index-large-projects/

jcfranco: Based on ☝️, I think we can move forward with this after patches are done. I'll create a separate issue to introduce and use z-index tokens.

Issue #3079, opened 2021-09-20
Bug: calcite-input showing Arabic characters for numbers only, add an option to switch to western numbers


AdelheidF:
Original issue submitted by Olga.

In FF and Safari when language is set to ar numbers in the input boxes appear in Arabic. Chrome doesn't have this problem, it is always showing English numbers. We only want to see English numbers, no matter the user's language.

![image](https://user-images.githubusercontent.com/44481292/134092650-58cda652-dd47-4d9f-8d40-7515e7abd9f2.png)

![image](https://user-images.githubusercontent.com/44481292/134087429-abf2a44d-dd50-4d2a-b7cb-629b1b760375.png)
When dragging the dot of color picker over the colored area the values update with other Arabic numbers.
Same when using up and down arrows.

https://jsbin.com/dutarurohu/edit?html,output

I'm using beta.65, but I also see it with beta.59.

### Actual Behavior

Showing Arabic numbers

### Expected Behavior

only English numbers

### Reproduction Steps and Sample

_Sample_:

Contact me

### Relevant Info

_Version_: `@esri/calcite-components@beta.65`

- [ ] CDN
- [ ] NPM package

benelan:
Reproducible in firefox: https://jsbin.com/yarikuquji/edit?html,output

julio8a:
@AdelheidF, where does the requirement to only support English numbers come from?
@babakbolour mentioned that showing Arabic characters is the best practice, see comment here: https://github.com/Esri/calcite-components/issues/2765#issuecomment-926079026

AdelheidF:
> where does the requirement to only support English numbers come from?

Map Viewer issue 3854

babakbolour:
@AdelheidF here is my perspective on this:  What Map Viewer has implemented or restricted users to (English numbers), is Map Viewer (or even Esri app) app specific. Calcite is not only for internal esri-developed apps, but also for public to use and create solutions/GIS apps. For that, it would be best to not exclude or restrict our users to only use English/western digits.
Also, @raje9276 is saying Chrome has now fixed their bug and now show Arabic western digits.

AdelheidF:
Ok, is there an option to change it to western?

babakbolour:
@AdelheidF let me talk to Olga and get back to you and clarify.

babakbolour:
[...]

PR #4195, opened 2022-03-03
feat(input): allow customizing numbering system

anveshmekala:
**Related Issue:** #3079 

## Summary

Allow users to specify numberingSystem  in  `calcite-input` via  `@Prop( ) numberingSystem` . The default numbering system will be determined by the `Intl` api  for a given `locale`


src/components/input/input.stories.ts
```diff
@@ -270,3 +270,9 @@ export const WithLabelAndInputMessageRTL = (): string => html`
     </calcite-label>
   </div>
 `;
+
+export const WithTibetNumberingSystem = (): string =>
+  html` <calcite-input type="number" numbering-system="tibt" value="123456"></calcite-input>`;
+
+export const ArabicLocaleWithLatinNumberingSystem = (): string =>
+  html` <calcite-input type="number" locale="ar" numbering-system="latn" value="123456"></calcite-input>`;
```
eriklharper, line 278: I would use `ar-EG` (Egypt) here instead of just `ar` or one of the other ones that actually displays arabic numerals by default.  ☝️

src/utils/locale.ts
```diff
@@ -52,11 +51,12 @@ export const locales = [
   "zh-TW"
 ];
 
-function createLocaleNumberFormatter(locale: string): Intl.NumberFormat {
+function createLocaleNumberFormatter(locale: string, numberingSystem?: string): Intl.NumberFormat {
```
eriklharper, line 54: You could also set `latn` as the default option in the `numberingSystem` parameter.

anveshmekala: The browser preference is set as default in the `input` component .

eriklharper: I was thinking of just having the default in here in case this gets used in other places where a default might not exist or is not needed, so it would look like: `function createLocaleNumberFormatter(locale: string, numberingSystem: string = 
 "latn"): Intl.NumberFormat`

src/components/input/input.tsx
```diff
@@ -121,6 +121,11 @@ export class Input implements LabelableComponent, FormComponent, InteractiveComp
   /** BCP 47 language tag for desired language and country format */
   @Prop() locale?: string = document.documentElement.lang || "en";
 
+  /**
+   * standard UniCode numeral system tag for localization
+   */
+  @Prop() numberingSystem?: string = new Intl.NumberFormat().resolvedOptions().numberingSystem;
```
driskull, line 127: If this is an optional property, it shouldn't have a default value set. If it has a default value, then the `?` is ignored since a value is already supplied. I think the above `locale` is incorrect as well.

Maybe for these we shouldn't set a default value on the component? We can compute the value internally if we need to but not set it on the component prop.

What do you think?

driskull: Does `new Intl.NumberFormat().resolvedOptions().numberingSystem` need to be called on every instance of the component?

driskull (commented):
Comments/questions :)

src/components/input/input.tsx
anveshmekala: It makes sense to remove the default value as the i18n team confirmed that CLDR defines a default numbering system for each locale . And it doesn't have to be called on every instance unless there is a change in locale / numbering System prop.

driskull (commented):
This looks good to me 👍 Nice work.

Just leaving a comment since I think someone else should review.

src/utils/locale.ts
```diff
@@ -56,12 +56,14 @@ export const locales = [
...
+const defaultNumberingSystem = new Intl.NumberFormat().resolvedOptions().numberingSystem;
 
-function createLocaleNumberFormatter(locale: string): Intl.NumberFormat {
+function createLocaleNumberFormatter(locale: string, numberingSystem = defaultNumberingSystem): Intl.NumberFormat {
   return new Intl.NumberFormat(locale, {
     minimumFractionDigits: 0,
-    maximumFractionDigits: 20
-  });
+    maximumFractionDigits: 20,
+    numberingSystem: numberingSystem
```
eriklharper, line 65: You should be able to simplify this to just say `numberingSystem,`

PR #4207, opened 2022-03-07
fix(filter): filter method runs on load in case there is a predefined search value prop

Elijbet:
There is a @Watch decorator which triggers filter() when the property value changes, but not when the property is initially set. We need to run filter in componetWillMount lifecycle hook as well, to account for a predefined value prop. Since we're filtering on initialization, we need the data and value to be present when initialized. Added a test for filter behavior with predefined value prop to checked for matched value.


src/components/filter/filter.e2e.ts
```diff
@@ -145,7 +145,7 @@ describe("calcite-filter", () => {
             name: "Matt",
             description: "developer",
             value: "matt",
-            metadata: { haircolor: "black", favoriteBand: "unknown" }
+            metadata: { haircolor: "black", favoriteBand: "Radiohead" }
```
jcfranco, line 148: ⚡ 🤘🏼 :neckbeard: 🤘🏼 ⚡

src/components/filter/filter.e2e.ts
```diff
@@ -226,4 +226,46 @@ describe("calcite-filter", () => {
...
+            name: "Matt",
+            description: "developer",
+            value: "matt",
+            metadata: { haircolor: "black", favoriteBand: "Radiohead" }
+          }
+        ];
+      });
+    });
+
+    function assertMatchingItems(filtered: any[], values: string[]): void {
```
jcfranco, line 255: Can you move this helper one level up so that it can be shared between both `describe` groups instead of having 2 copies?

jcfranco (approved):
💥

src/components/filter/filter.e2e.ts
Elijbet: Had to be done.

driskull: giphy

Issue #3727, opened 2021-12-16
calcite-tooltip performance issues when rendering many instances


rpanichakit9541:
### Actual Behavior

When rendering a bunch of calcite-tooltip (300 in the sample CodePen), I noticed some slow rendering when scrolling in Firefox. In Chrome, the rendering does not blink in and out, but the scrolling itself becomes a bit laggy.

![tooltip_performance](https://user-images.githubusercontent.com/44479773/146433572-b3bcc50f-5546-4ff0-97d6-dd8633ffa5a1.gif)

### Expected Behavior

I would expect the scrolling/rendering to be smooth regardless of how many tooltips are on the page.

### Reproduction Steps and Sample

_Sample_:

1. https://codepen.io/rpanichakit/pen/ZEXKYYQ

### Relevant Info

Firefox 95.0.1 (64-bit) 
Chrome Version 96.0.4664.93 (Official Build) (64-bit)
OS: Windows 10

_Reproducible version_: `@esri/calcite-components@1.0.0-beta.71`

driskull:
I'm not noticing any performance issues on FF/Chrome with beta71 or 74.

rpanichakit9541:
@driskull - I see the blinking issue on my FF 96.0 and Chrome Version 97.0.4692.71 (Official Build) (64-bit)

driskull:
@rpanichakit9541 does this codepen make any difference?

https://codepen.io/driskull/pen/JjOoJaa?editors=0010

rpanichakit9541:
@driskull - Unfortunately, I still see the issue in your codepen

![tooltip_perf](https://user-images.githubusercontent.com/44479773/151249472-c9818da4-1eda-4af3-9f82-441d3b7f1e1c.gif)

driskull:
@rpanichakit9541 I'm not sure we can do anything to fix this at this time.

If you remove setting the referenceElement on the tooltip, I suspect the performance issue will no longer be a problem.

@benelan @jcfranco maybe we can move this to the backburner for now?

The reason for this performance hit is that it updates the placement of the tooltip when scrolling occurs. We need to have the positioning correct before we show the tooltip to handle the slight animation.

rpanichakit9541:
@driskull - Yes I verified that removing `referenceElement` does resolve the issue.

driskull:
I think floating-ui might fix this performance a bit. It would allow us to not reposition any hidden tooltips.

benelan:
[...]

PR #4216, opened 2022-03-09
fix(combobox, dropdown, input-date-picker, popover, tooltip): disable scroll event listeners for hidden poppers

jcfranco:
**Related Issue:** #3727 

## Summary

This PR updates all popper-owning components to toggle the `eventListeners` modifier based on the popper's visibility (as suggested [here](https://popper.js.org/docs/v2/tutorial/#performance)). 

This also cleans up some internals for combobox. It seemed to have multiple state props based on `active`.

### DIFF
--- a/src/components/combobox/combobox.tsx
+++ b/src/components/combobox/combobox.tsx
@@ -446,12 +446,12 @@ export class Combobox implements LabelableComponent, FormComponent, InteractiveC
   };
 
   private toggleCloseEnd = (): void => {
-    this.active = true;
+    this.active = false;
     this.el.removeEventListener("calciteComboboxClose", this.toggleCloseEnd);
   };
 
   private toggleOpenEnd = (): void => {
-    this.active = false;
+    this.active = true;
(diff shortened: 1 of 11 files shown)


driskull:
nice! Thanks for this Franco.

src/components/combobox/combobox.tsx
```diff
@@ -1065,7 +1058,7 @@ export class Combobox implements LabelableComponent, FormComponent, InteractiveC
   }
 
   render(): VNode {
-    const { guid, open, label } = this;
```
driskull, line 1068: Looks like `open` was being used below. Causing a JS error.

driskull (changes requested):

src/components/combobox/combobox.tsx
jcfranco: Thanks for catching this. There were other references that I missed that didn't come up in focused tests.

driskull (approved):
👍 Nice!

Issue #3931, opened 2022-01-13
[Combobox] - Unable to use icon to close dropdown while open


macandcheese:
### Actual Behavior

If the dropdown is open, interaction with the chevron is disabled

https://user-images.githubusercontent.com/4733155/149390361-e9a8269f-1c07-4422-a13c-4a460855b696.mov

.

### Expected Behavior

I would expect to be able to toggle the dropdown by using the icon affordance at all times. It uses `cursor:pointer` which indicates interactivity.

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/combobox/#sample

### Reproduction Steps

1. Open the dropdown. 
2. Try to close the dropdown by clicking on the chevron icon.

### Reproduction Version

beta 74

### Relevant Info

Regression? Last working version: `@esri/calcite-components@1.0.0-VERSION`

### Source

- [ ] CDN
- [ ] NPM package

avezina:
We have also observed this issue in the new [Utility Network Trace widget](https://developers.arcgis.com/javascript/latest/api-reference/esri-widgets-UtilityNetworkTrace.html) which is using this component. It is difficult to "get out" of the Combobox right now. Thank you!
![image](https://user-images.githubusercontent.com/50338557/149427288-2eb744e9-47d3-4e5e-ba95-471a5712d0e0.png)

y0n4:
I noticed that the chevron icon is not focusable/tabbable (this can be addressed by wrapping it in a calcite-button).

Right now when the dropdown is opened, the whole combobox gets focused. In the case when the chevron icon gets clicked and the dropdown opens/closes, which in element gets focused? The whole combobox or the chevron icon button or does it depend on which element gets clicked?

macandcheese:
My vote would be to not make the element itself focusable at all - just keep the visual as an affordance (and change the cursor to the same as the input). Clicks anywhere on the input would focus and open / close and focus. That behavior is replicable with a keyboard Esc press.

Thoughts @asangma ?

asangma:
Basically, the behavior should be the same as CalciteSelect.

geospatialem:
Verified chevron icon 🔼 behavior when dropdown is open in next 👍🏻

avezina:
Is the new behavior available to test here?: https://next.sites.afd.arcgis.com/calcite-design-system/components/combobox/
[...]

Issue #2514, opened 2021-07-12
Combobox: change dropdown focus/open behavior


ptmdmusique:
### Question

Currently the combobox element will open the dropdown whenever it's focused. I'm wondering if we can make it more like a `select` tag where it will only open the dropdown on clicked or on some certain keys pressed instead

cc @paulcpederson

paulcpederson:
@bstifle this came up recently while looking into some accessibility issues. Interested in your thoughts.

paulcpederson:
@jcfranco if you get a second, we should figure out how to prioritize this one

jcfranco:
@paulcpederson Can you share info on the a11y issues and provide an effort estimate on this one? Seems low effort to me.

paulcpederson:
The [spec for combobox](https://www.w3.org/TR/wai-aria-practices-1.1/#combobox) sort of allows for multiple interpretations:

> The popup is hidden by default, and the conditions that trigger its display are specific to each implementation. Some possible popup display conditions include:
> - It is displayed only if a certain number of characters are typed in the textbox and those characters match some portion of one of the suggested values.
> - It is displayed as soon as the textbox is focused, even if the textbox is empty.
> - It is displayed when the Down Arrow key is pressed or the show button is activated, possibly with a dependency on the content of the textbox.
> - It is displayed if the value of the textbox is altered in a way that creates one or more partial matches to a suggested value.

Currently we use `It is displayed as soon as the textbox is focused` but for the single select, because it's more like a select menu, I think we'd like to make it comparable to a select menu, meaning moving to `It is displayed when the Down Arrow key is pressed or the show button is activated`. Obviously typing in characters will work as well.

I'm not entirely sure what effort would be required, but I think it should be relatively straightforward to move the opening behavior behind the down arrow key and off the input focus event. However the combobox is fairly complex so ensuring that this doesn't cause regressions and making sure tests pass successfully may add time.

geospatialem:
@paulcpederson noticed in the list you mentioned a list of items to implement with this issue. While running through a verification it looks like we were able to fulfill one of the requirements. 

Can you verify we're fulfilling the expected behavior, as there is no current functionality for characters entered by the user. _Note that there is functionality for arrow keys (and home/end), spacebar, and on click._

![screenshot of combobox with focus/open behavior](https://user-images.githubusercontent.com/5023024/161626424-4627f69b-bc4f-44f1-9e29-95f24c69133f.png)

[...]

PR #4218, opened 2022-03-09
fix(combobox): can toggle the dropdown using the icon

Elijbet:
**Related Issue:** #3931, #2514

## Summary

When working on the issue, I realized that open/close is based on focus, and there are many instances of calling setFocus. Changed the behavior to only open the dropdown on click or on some certain keys pressed instead. Here is the breakdown of how it behaves now.

-  Tab focuses on the next item, but doesn’t open it; another tab will close if it’s open, next tab moves to next input (shift + tab will go in reverse)
-  Arrow right/left only moves from one chip to another
-  Mac Fn + right/left (home / end) opens dropdown and puts focus on first item/last item; (existing bug described in #4265)
-  Escape closes the dropdown, but remains focused
-  Enter shouldn’t toggle the dropdown open/close, instead it should be left to submit a form; arrow down opens the combobox, esc closes the combobox
-  Clicking the entire area focuses/toggles the listbox open/close and chevron up/down; uses a pointer.
- Allow Space key hit in an empty field to open the dropdown. Once you have begun typing, space key adds a space to the string. 
-  Label click focuses on the input, does not toggle open/close


src/components/combobox/combobox.tsx
```diff
@@ -505,7 +509,7 @@ export class Combobox implements LabelableComponent, FormComponent, InteractiveC
       return;
     }
 
-    if (this.allowCustomValues && this.text) {
+    if (this.allowCustomValues && this.text.length) {
```
driskull, line 512: I wonder if we should check for blank spaces before adding a custom chip? Maybe `&& this.text?.trim().length`?

src/components/combobox/combobox.e2e.ts
```diff
@@ -418,7 +408,7 @@ describe("calcite-combobox", () => {
       expect(await item.getProperty("selected")).toBe(true);
     });
 
-    it("should replace current value to new custom value in single selection mode", async () => {
+    it("should replacee current value to new custom value in single selection mode", async () => {
```
driskull, line 411: typo-eeee :)

src/components/combobox/combobox.e2e.ts
```diff
@@ -481,21 +471,188 @@ describe("calcite-combobox", () => {
...
+      await inputEl.focus();
+      await page.waitForChanges();
+      expect(await page.evaluate(() => document.activeElement.id)).toBe("parentOne");
+
+      await page.keyboard.press("End");
+      await page.waitForChanges();
+      const lastFocusedGroupItem = await page.find("#three >>> .label--active");
+      expect(lastFocusedGroupItem).toBeTruthy();
+
+      // follow up issue on existing bug: the last item in focus won't scroll into view.
```
driskull, line 649: Can we link the issue here? or maybe we can get it working with some kind of timeout call?

driskull (approved):
Looking good @Elijbet 👍 💯

src/components/combobox/combobox.e2e.ts
Elijbet: https://tenor.com/view/ross-eeee-ffiends-gif-11109902

Elijbet: Typooo

src/components/combobox/combobox.e2e.ts
```diff
@@ -213,33 +198,29 @@ describe("calcite-combobox", () => {
...
+          <calcite-combobox-item value="Rocks" text-label="Rocks"></calcite-combobox-item>
+          <calcite-combobox-item value="Insects" text-label="Insects"></calcite-combobox-item>
+          <calcite-combobox-item value="Rivers" text-label="Rivers"></calcite-combobox-item>
+        </calcite-combobox>
+      `
+    );
+    const input = await page.find("calcite-combobox >>> input");
     const openEvent = page.waitForEvent("calciteComboboxOpen");
-    await element.click();
+    await input.click();
```
jcfranco, line 221: Does this need to target the `input`? Clicking on the combobox should work too, no? 

This also applies to any other _new/modified_ tests with a similar setup.

src/components/combobox/combobox.tsx
```diff
@@ -505,7 +509,7 @@ export class Combobox implements LabelableComponent, FormComponent, InteractiveC
       return;
     }
 
-    if (this.allowCustomValues && this.text) {
+    if (this.allowCustomValues && this.text?.trim().length) {
```
jcfranco, line 512: I _think_ there's no need to use conditional chaining here since `text` should be guaranteed (it's initialized/reset to "" and is always updated from an `<input>`'s value).

src/components/combobox/combobox.e2e.ts
```diff
@@ -481,21 +471,188 @@ describe("calcite-combobox", () => {
...
+      await inputEl.focus();
+      await page.waitForChanges();
+      expect(await page.evaluate(() => document.activeElement.id)).toBe("parentOne");
+
+      await page.keyboard.press("End");
+      await page.waitForChanges();
+      const lastFocusedGroupItem = await page.find("#three >>> .label--active");
+      expect(lastFocusedGroupItem).toBeTruthy();
+
+      // follow up issue #4265 on an existing bug: the last item in focus won't scroll into view.
```
jcfranco, line 649: Could you remove this and update the related issue with any relevant info instead?

src/components/combobox/combobox.e2e.ts
```diff
@@ -481,21 +471,188 @@ describe("calcite-combobox", () => {
...
-            <calcite-combobox-item id="one" value="one" text-label="one"></calcite-combobox-item>
-            <calcite-combobox-item id="two" value="two" text-label="two"></calcite-combobox-item>
+          <calcite-combobox id="parentOne">
+            <calcite-combobox-item id="one" value="one" label="one"></calcite-combobox-item>
+            <calcite-combobox-item id="two" value="two" label="two"></calcite-combobox-item>
+            <calcite-combobox-item-group label="Last Item">
+              <calcite-combobox-item id="three" value="three" label="three"></calcite-combobox-item>
+            </calcite-combobox-item-group>
+          </calcite-combobox>
+          <calcite-combobox id="parentTwo">
```
jcfranco, line 540: Can the tests use a single combobox or why is a sibling combobox needed?

src/components/combobox/combobox.e2e.ts
```diff
@@ -481,21 +471,188 @@ describe("calcite-combobox", () => {
...
+  describe("mouse navigation", () => {
+    let page: E2EPage;
+    let label;
+    let popper;
+
+    beforeEach(async () => {
+      page = await newE2EPage();
+      await page.setContent(
+        html`
+          <calcite-label id="label">
```
jcfranco, line 483: Clicking on the label should behave the same as clicking on the component, so I think we could remove the `calcite-label` aspect from these tests. Label integration is already handled by the `labelable` helper.

jcfranco (approved):
Once the last round of changes are made, this is good to merge! 🎉

src/components/combobox/combobox.e2e.ts
Elijbet: To test tabbing behavior and the second combobox items are selected to make it easier to test chip interaction.

jcfranco: Tabbing does not depend on another combobox being present though. The chip interaction is a different test and it would be better for maintainability to separate it.

Issue #3982, opened 2022-01-21
calcite-value-list drag-enabled selection and auditory reordering not available to screen-readers


FelFly:
### Summary

Using a screen-reader with the calcite-value-list, I want to hear inline instruction on what keystrokes to use to interact with the component. I want to be able to "pick up" an item, hear the value label of the item associated with the drag control, move my selection and hear its current position, and "drop" my selection in a new position.

Applicable WCAG Guidelines
[1.3.3 Sensory Characteristics](https://www.w3.org/WAI/WCAG21/Understanding/sensory-characteristics.html)
[2.1.1 Keyboard](https://www.w3.org/TR/UNDERSTANDING-WCAG20/keyboard-operation-keyboard-operable.html)
[2.2.6 Headings and Labels](https://www.w3.org/WAI/WCAG21/Understanding/headings-and-labels.html)

### Actual Behavior

There are three things I observed happening:

1 . In VoiceOver and NVDA, I do not hear the option name (eg. Rivers) associated with the toggle button that I am selecting / deselecting.

2a. Also in VoiceOver, I cannot use the VO keys (Ctrl+Option) with Spacebar as if it is actually the Spacebar, which is how it operates other places on the web. It will not select the drag button and all I hear is "deselect button." Likewise, even though normally the VO keys + arrow keys allow me to move up/down through content on a page, I cannot use the VO keys with the arrow keys to move this particular element. If I use strictly the Spacebar and the arrow keys as if I were a sighted user, then it works, but that's contradictory to how I'd expect to use VoiceOver.

2b. In NVDA, there aren't modify keys like in VoiceOver, so in theory using Spacebar should work. However, it does not. I do not observe the item being selected. In NVDA if I am moving from top to bottom of a page (and not jumping around by links or form elements) then I use the up/down arrow keys to move through content, but when I had my toggle highlighted, the arrow keys did not move the value item.

3 . If I do use the sighted keyboard user controls of Spacebar and the arrow keys, then I still don't hear any awareness of placement. I would expect to hear something along the lines of "current position 2 of 4" and then as I move it down, for example, "current position 3 of 4" etc.

Video of problems 1-3 in VoiceOver: https://esriis-my.sharepoint.com/:v:/g/personal/klar7790_esri_com/EaCXndpv0wZDqcX8Kg-23PMByJ5TV01Xua8ky3F45Lt0Ig?e=4scdVo

[...]

PR #4228, opened 2022-03-10
fix(value-list): enable auditory reordering of items when dragEnabled

anveshmekala:
**Related Issue:** #3982 

## Summary

**prototype . Ref: [Sortable list Accessibility design spec](https://medium.com/salesforce-ux/4-major-patterns-for-accessible-drag-and-drop-1d43f64ebf09#:~:text=4%20Accessible%20Patterns%20for%20Drag%20and%20Drop)


src/components/pick-list/shared-list-render.tsx
```diff
@@ -31,6 +31,9 @@ export const List: FunctionalComponent<{ props: ListProps } & DOMAttributes<any>
   return (
     <Host aria-busy={loading.toString()} role="menu" {...rest}>
       <section>
+        <span aria-live="assertive" class="assistive-text">
+          {" "}
```
driskull, line 35: Is this displayed?

src/components/value-list/resources.ts
```diff
@@ -8,3 +8,7 @@ export enum ICON_TYPES {
 export const TEXT = {
   filterResults: "Filter results"
 };
+
+export const HANDLE_LABEL = "press space and use arrow keys to reorder content";
```
driskull, line 12: I think we need an `intl` property to go along with this so that users can localize it.

src/components/value-list/value-list.tsx
```diff
@@ -279,7 +278,9 @@ export class ValueList<
 
     const { el } = this;
     const nextIndex = moveItemIndex(this, item, event.key === "ArrowUp" ? "up" : "down");
-
+    this.updateLiveText(
+      `New Position ${nextIndex + 1} of ${this.getItems().length}, press space to confirm `
```
driskull, line 282: This needs to be an `intl` property as well so we can have it localized. It might need to be a HTML template so that the numbers can be inserted.

driskull (changes requested):
Looking good. I think the main things are the properties for users to be able to define the text.

src/components/pick-list/shared-list-render.tsx
anveshmekala: nope. this isn't displayed. it's just an assistive text element which has no width or height

anveshmekala: Will remove the empty string.

anveshmekala:
> Looking good. I think the main things are the properties for users to be able to define the text.

Exposing a property for users to customize the text would be out of scope as it may drive the user to parse text that isn't 
perceivable by screen reader. 
Regarding localization ,it would be great to localize the screen reader text as voiceover support multiple languages but that's something we many need to define first considering some of the locales supported by ArcGIS JS api aren't supported by voiceover. (e.g: Bosnian and Bulgarian are not supported by voice over in web)

driskull:
> Exposing a property for users to customize the text would be out of scope as it may drive the user to parse text that isn't
perceivable by screen reader.

@jcfranco what do you think? I would think this text should have a `intl` property for it to be localized.

jcfranco:
I agree w/ @driskull. Until our components have built-in localization, we need props to enable users to provide their own strings (avoiding hardcoded strings). There's already precedent for this in some components (e.g., `intlClose` in `calcite-alert`, `calcite-modal`).

anveshmekala:
> I agree w/ @driskull. Until our components have built-in localization, we need props to enable users to provide their own strings (avoiding hardcoded strings). There's already precedent for this in some components (e.g., `intlClose` in `calcite-alert`, `calcite-modal`).

Is it expected that prop only manages the wording but not the index part where the screen reader readout `current position 2 of 4` ?  or can we safely assume once the user start using prop this feature will be unavailable . Any thoughts? 

@jcfranco , @driskull

jcfranco:
I don't think any of our existing components do this, but we could use a string with placeholders to replace at runtime. Similar to template literals. We'd need to make sure the doc mentions the placeholder tokens.

geospatialem (commented):
### Voiceover and Safari
Overall really good experience. I love how with Safari and Voiceover I'm able to navigate using the arrow keys throughout the `value-list-items` and actions. The drag action is great and the feedback is very clear. Would recommend adding context to the item that's being moved (e.g., drag enabled for 'Food' list item).

https://user-images.githubusercontent.com/5023024/175697474-1b3ede31-7a03-4a8c-ac8b-92cd22206aa8.mp4

### NVDA and JAWS
Both Windows screen readers unfortunately have a few restrictions, which could be a result of the other `value-list`s as well. I'm unable to move through the lists entirely and could only do so by tabbing, and even then it wasn't a great experience.

Related to the drag handle, when selecting the `space` key the focus changes, but am unable to move the list item. 

Was able to receive the assertive text (it works **great**!), but unfortunately none of them items can be moved via keyboard. Maybe this is a result of different key naming conventions between Mac and Windows?


#### JAWS transcript with accompanying video
```
press space and use arrow keys to re-order content. current position 1 of 4 Button 
To move through items press up or down arrow.
action start
Button 
Unlabeled 0
Button 
press space and use arrow keys to re-order content. current position 2 of 4 Button 
Unlabeled 0
Button 
Unlabeled 0 Button 
press space and use arrow keys to re-order content. current position 2 of 4
Button 
action start
Button 
action start Button 
Rent Button 
press space and use arrow keys to re-order content. current position 1 of 2 Button 
One Button 
Unlabeled 0 Button 
One Button 
press space and use arrow keys to re-order content. current position 1 of 2 Button 
Unlabeled 0
Button 
press space and use arrow keys to re-order content. current position 2 of 2 Button
```

https://user-images.githubusercontent.com/5023024/175698661-04fab8e9-5ea0-4aca-bee7-5fcc7192f843.mp4

FelFly:
Drag-and-drop is much improved in VoiceOver, thank you! And Kitty already caught my one suggestion, which would be to associate the drag handle button with its corresponding title text so there's unique context per button in case someone open's VoiceOver's shortcut menu to form controls.

I did notice while navigating down to drag and drop that single selection and multi-selection have the same checked/unchecked issue I opened on the [pick-list](https://github.com/Esri/calcite-components/issues/4434). Not expecting it to be fixed as part of this issue, just wanted to mention it.

src/components/value-list/value-list.tsx
```diff
@@ -335,6 +393,43 @@ export class ValueList<
     return type;
   }
 
+  updateScreenReaderText(assertiveText: string): void {
+    //code to update text to our aria-live span element
+    const screenReaderElement = this.el.shadowRoot.querySelector(".assistive-text");
```
driskull, line 398: We should store this node rather than query for it each time.

src/components/value-list/value-list.tsx
```diff
@@ -335,6 +393,43 @@ export class ValueList<
...
+  getScreenReaderText(item: HTMLCalciteValueListItemElement, status: DragStatus): string {
+    const { items } = this;
+    const total = items.length;
+    const position = getItemIndex(this, item) + 1;
+
+    switch (status) {
+      case "start":
+        return this.intlDragHandleStart
+          ? this.intlDragHandleStart
+          : `press space and use arrow keys to re-order content. current position ${position} of ${total}`;
```
driskull, line 415: Should all these strings be stored in a separate file? Like a resources file?

driskull (changes requested):
Looking good! Just a few suggestions :)

src/components/value-list/value-list.tsx
anveshmekala: The strings contain variables `${position}` , `${total}` which cant be stored in  a resource file.

anveshmekala:
> Drag-and-drop is much improved in VoiceOver, thank you! And Kitty already caught my one suggestion, which would be to associate the drag handle button with its corresponding title text so there's unique context per button in case someone open's VoiceOver's shortcut menu to form controls.
> 
> I did notice while navigating down to drag and drop that single selection and multi-selection have the same checked/unchecked issue I opened on the [pick-list](https://github.com/Esri/calcite-components/issues/4434). Not expecting it to be fixed as part of this issue, just wanted to mention it.

@geospatialem, @FelFly  would it be better if the context is added as a prefix to the drag instructions, or suffix make sense? 

example: 
`Food item. press space and use arrow keys to re-order content. current position 2 of 4 .`
`Food Item re-ordering. current position 1 of 4.`

FelFly (commented):
I think the prefix on the first line makes sense, or at least that's what I encounter the most frequently in VoiceOver.

geospatialem:
> @geospatialem, @FelFly would it be better if the context is added as a prefix to the drag instructions, or suffix make sense?

> I think the prefix on the first line makes sense, or at least that's what I encounter the most frequently in VoiceOver.

Concur with @FelFly. Prefix provides context prior to action. In addition to expected behavior, is helpful to a wider audience in particular cognitive users who rely on screen readers. 

Thanks for all your great work on this effort, @anveshmekala!

src/components/value-list/value-list.tsx
anveshmekala: nvm. gotcha

src/components/value-list-item/value-list-item.tsx
```diff
@@ -189,6 +191,10 @@ export class ValueListItem implements ConditionalSlotComponent, InteractiveCompo
     this.selected = event.detail.selected;
   };
 
+  handleFocus = (event: FocusEvent): void => {
+    this.calciteInternalListItemDragHandleFocused.emit(event);
```
driskull, line 195: Do we need to emit an event nested inside another event? We may want to discuss this pattern if so.

src/components/pick-list/shared-list-logic.ts
```diff
@@ -176,6 +176,12 @@ export function moveItemIndex<T extends Lists>(
   return index;
 }
 
+export function getItemIndex<T extends Lists>(list: List<T>, item: ListItemElement<T>): number {
+  const { items } = list;
+  const currentIndex = (items as ListItemElement<T>[]).indexOf(item);
```
driskull, line 181: nitpick...

```ts
return (items as ListItemElement<T>[]).indexOf(item);
```

src/components/value-list/value-list.tsx
```diff
@@ -134,6 +165,9 @@ export class ValueList<
   connectedCallback(): void {
     initialize.call(this);
     initializeObserver.call(this);
+    if (this.dragEnabled) {
+      this.assitiveTextEl = this.el.shadowRoot.querySelector(".assistive-text");
```
driskull, line 169: For this, can you use `ref={}` to store the element instead of querying it?

src/components/value-list/value-list.tsx
```diff
@@ -192,6 +226,15 @@ export class ValueList<
     event.stopPropagation();
   }
 
+  @Listen("calciteInternalListItemDragHandleFocused")
+  handleDragHandleFocus(event: FocusEvent): void {
+    const item = getItemElement(event);
+    const handleElement = getHandleElement(event.detail as any);
```
driskull, line 232: Maybe this event can just pass the element instead of the event? That would clean up the `any` as well.

driskull (changes requested):
Looking good! Awesome stuff @anveshmekala 🥂

src/components/pick-list/shared-list-render.tsx
```diff
@@ -24,14 +25,16 @@ export const List: FunctionalComponent<{ props: ListProps } & DOMAttributes<any>
...
+    setFilterEl,
+    dragEnabled
   },
   ...rest
 }): VNode => {
   const defaultSlot = <slot />;
   return (
     <Host aria-busy={toAriaBoolean(loading)} role="menu" {...rest}>
       <section>
+        {dragEnabled ? <span aria-live="assertive" class="assistive-text" tabindex="-1" /> : null}
```
geospatialem, line 37: Was there discussion on the usage of `assertive` for `aria-live`? Usually its restricted to time sensitive or critical notifications [[Resource](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions)]. 

If you agree we don't need to dismiss previous verbiage to screen readers, can we change to `aria-live="polite"`?

src/components/value-list/value-list.tsx
```diff
@@ -335,7 +382,17 @@ export class ValueList<
...
   }
 
+  updateScreenReaderText(assertiveText: string): void {
+    //code to update text to our aria-live span element
+    const screenReaderElement = this.el.shadowRoot.querySelector(".assistive-text");
+    screenReaderElement.textContent = assertiveText;
+  }
+
+  updateHandleAriaLabel(handleElement: HTMLSpanElement, assertiveText: string): void {
+    handleElement.ariaLabel = assertiveText;
```
geospatialem, line 392: nitpick (and if changing to `aria-live="polite"`): rename `assertiveText` to `accessibleText`, which could help if we make changes down the road as well.

geospatialem (commented):
Few minor recommendations recommending `aria-live="polite"` for screen readers. Nice work, @anveshmekala! 

_Side note_: `value-list` will only work on Mac (including VoiceOver) or when using mouse or keyboard functionality on Windows (without a screen reader). Unfortunately if either JAWS or NVDA are activated on Windows, navigation of the component isn't great and context is not provided to screen reader users. This has been verified prior to this PR, so we'll focus on the `list` component enhancements to support Windows screen reader users in the future.

src/components/pick-list/shared-list-render.tsx
anveshmekala: `polite=assertive` is required for the assistive text to be able to announce the instructions as it is not the focused element when the dragging /sorting is in happening. if we use `aria-live = polite` then the instructions will be placed in queue and might not be announced as expected.

geospatialem: Ah, that's right, I think we discussed this earlier as well. Sounds like a great use case for `aria-live="assertive"`. Thanks @anveshmekala!

driskull (approved):
LGTM! can we get one more review before merging?

src/components/value-list/value-list.tsx
```diff
@@ -101,6 +102,34 @@ export class ValueList<
    */
   @Prop() selectionFollowsFocus = false;
 
+  /**
+   * localize the screen reader text when drag button is focused
```
jcfranco, line 106: @geospatialem Can you give the doc an editorial pass? Some things I notice:

1. Inconsistent capitalization
2. `place holder` -> `placeholder`
3. placeholder names should be camelCased
4. `indexes` -> `indices`? (I'll defer to you on this one as both are correct)

src/components/value-list/value-list.tsx
```diff
@@ -125,6 +154,8 @@ export class ValueList<
 
   filterEl: HTMLCalciteFilterElement;
 
+  assitiveTextEl: HTMLSpanElement;
```
jcfranco, line 157: `assitiveTextEl` -> `assistiveTextEl`

src/components/value-list/interfaces.ts
```diff
@@ -0,0 +1 @@
+export type DragStatus = "start" | "activated" | "newPosition" | "currentPosition";
```
jcfranco, line 1: Suggestion: `idle | active | change | commit`. I think the drag status types can be named independently from the accessible messages.

src/components/value-list/value-list.tsx
```diff
@@ -335,7 +369,34 @@ export class ValueList<
     return type;
   }
 
+  updateScreenReaderText(assertiveText: string): void {
+    //code to update text to our aria-live span element
```
jcfranco, line 373: I think we can drop this comment since the code conveys this.

src/components/value-list/value-list.tsx
```diff
@@ -335,7 +369,34 @@ export class ValueList<
     return type;
   }
 
+  updateScreenReaderText(assertiveText: string): void {
```
jcfranco, line 372: Should the param name simply be `text`?

src/components/value-list/utils.ts
```diff
@@ -0,0 +1,56 @@
...
+      return intlDragHandleCurrentPosition
+        ? intlDragHandleCurrentPosition
+        : `${item.label} ,current position ${position} of ${total}`;
+    default:
+      break;
+  }
+}
+
+export function getHandleAndItemElement(event: KeyboardEvent | FocusEvent): {
+  handleElement: HTMLCalciteHandleElement;
```
jcfranco, line 42: Nitpick: the interface includes 2 elements, but the naming is inconsistent. My vote is for `{ handle, item }`.

src/components/value-list/utils.ts
```diff
@@ -0,0 +1,56 @@
...
+    intlDragHandleNewPosition,
+    intlDragHandleCurrentPosition
+  } = valueList;
+  const total = items.length;
+  const position = getItemIndex(valueList, item) + 1;
+  switch (status) {
+    case "start":
+      return intlDragHandleStart
+        ? intlDragHandleStart
+        : `${item.label} ,press space and use arrow keys to re-order content. current position ${position} of ${total}`;
```
jcfranco, line 23: Can you try switching this to an if/else statement? Switch statements can add a lot of unnecessary nesting.

Also, IMO, a newline between the statement and const assignment can help improve visually scanning the code.

src/components/value-list/utils.ts
```diff
@@ -0,0 +1,56 @@
...
+        ? intlDragHandleActivated
+        : `Reordering ${item.label} ,current position ${position} of ${total}`;
+    case "newPosition":
+      return intlDragHandleNewPosition
+        ? intlDragHandleNewPosition
+        : `${item.label} , new position ${position} of ${total}. press space to confirm`;
+    case "currentPosition":
+      return intlDragHandleCurrentPosition
+        ? intlDragHandleCurrentPosition
+        : `${item.label} ,current position ${position} of ${total}`;
```
jcfranco, line 35: There are punctuation/capitalization issues in the default strings that need to be fixed:

1. `${item.label}, press space and use arrow keys to reorder content. Current position ${position} of ${total}`
2. `${item.label}, current position ${position} of ${total}`
3. `${item.label}, new position ${position} of ${total}. Press space to confirm`
4. `Reordering ${item.label}, current position ${position} of ${total}`

@geospatialem Can you give this an editorial pass as well?

jcfranco (changes requested):

jcfranco:
I also forgot to mention that this will need a test. Asserting on the text changing after keyboard interactions should suffice.

src/components/value-list/value-list.tsx
anveshmekala: Roger that.

src/components/value-list/interfaces.ts
anveshmekala: Thought of keeping them in sync so that it's easy to understand but the suggested names are more precise and clean.

src/components/value-list/utils.ts
geospatialem: @jcfranco's recommendations are sound to me. Perhaps periods at the end of each statement, but the contents and edits are great.

geospatialem (commented):
Few considerations for API reference styling. Looking forward to this fantastic update, @anveshmekala!

src/components/value-list/value-list.tsx
```diff
@@ -101,6 +102,34 @@ export class ValueList<
    */
   @Prop() selectionFollowsFocus = false;
 
+  /**
+   * localize the screen reader text when drag button is focused
+   *
+   * use ${itemposition} of ${totalitems} as place holder for indexes.
```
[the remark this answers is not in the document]
anveshmekala: Good catch @jcfranco. the placeholder text is now replaced.  
`${itemposition}` is replaced by `${position}`
`${totalitems}` is replaced by `${items}` 

We don't need any code for this we use the same template literals while building the assistive text.

src/components/value-list/value-list.tsx
```diff
@@ -103,30 +103,30 @@ export class ValueList<
   @Prop() selectionFollowsFocus = false;
 
   /**
-   * When "drag-enabled" is true and active, specifies accessible context to the `calcite-value-list-item's` initial position.
+   * When "drag-enabled" is true and active, specifies accessible context to the `calcite-value-list-item`'s initial position.
    *
-   * use ${position} of ${total} as placeholder for indexes.
+   * use ${position} of ${total} as placeholder for displaying indices.
```
geospatialem, line 108: nitpick: Can we capitalize "**U**se" throughout (lines 108, 115, 122, 129)?

anveshmekala: sure thing.

src/components/value-list/utils.ts
```diff
@@ -0,0 +1,63 @@
...
+    intlDragHandleNewPosition,
+    intlDragHandleCurrentPosition
+  } = valueList;
+
+  const total = items.length;
+  const position = getItemIndex(valueList, item) + 1;
+
+  if (status === "idle") {
+    const idleText = intlDragHandleStart
+      ? intlDragHandleStart
```
jcfranco, line 23: You'll have to do a manual replacement for the placeholder tokens on user-provided strings, otherwise they won't get replaced (template literals are interpolated at creation-time). We'll be working on built-in translations soon, so this can be a simple string replacement.

jcfranco (approved):
Once the remaining comments are addressed, this is good to merge! ✨💪✨

Issue #4197, opened 2022-03-04
calcite-popover doesn't show when referenceElement is calcite-panel


AdelheidF:
### Actual Behavior

calcite-popover doesn't show when referenceElement is calcite-panel, `placement="leading-start"`.

I'm guessing it's related to this change: https://github.com/Esri/calcite-components/pull/4115

![image](https://user-images.githubusercontent.com/44481292/156674718-105f44be-816f-479c-920b-73cbd2dd3398.png)

### Expected Behavior

popover shows when it's set to open

### Reproduction Sample

https://codepen.io/afreitag/pen/XWzOjwK

### Reproduction Steps

Popover should show when starting codepen app. Change to version beta.77 and you'll see the popover.

### Reproduction Version

beta.78

### Regression?

beta.77

driskull:
Interesting. For some reason, the popover is getting "escaped". `data-popper-escaped`. https://popper.js.org/docs/v2/modifiers/hide/

driskull:
@jcfranco we can do one of two options pretty easily...

1. Add a property to turn off the "hide" modifier. 
2. Remove hiding the popover when it is "data-popper-escaped". 

What do you think?

driskull:
This demo explains pretty well if you scroll down. https://popper.js.org/docs/v2/modifiers/hide/#demo

jcfranco:
Let's do 2. I think showing when escaped is fine since it means popper couldn't place it ideally. We still want to hide it if the reference is not visible, right?

PR #4239, opened 2022-03-15
fix(popover, tooltip, combobox, dropdown, input-date-picker): display escaped poppers

jcfranco:
**Related Issue:** #4197

## Summary

This restores visibility for escaped poppers. Poppers with hidden reference elements will continue to be hidden.


driskull (approved):
👍 Perfect. Just what I was going to do :)

Issue #4169, opened 2022-02-27
action-bar expanded not respected when expand-disabled


ffaubry:
### Actual Behavior

```
<calcite-action-bar expanded expand-disabled>
    <calcite-action-group>
        <calcite-action text="Add" icon="plus"></calcite-action>
        <calcite-action text="Save" icon="save"></calcite-action>
    </calcite-action-group>
    <calcite-action-group>
        <calcite-action text="Layers" icon="layers"></calcite-action>
        <calcite-action text="Basemaps" icon="layer-basemap"></calcite-action>
    </calcite-action-group>
</calcite-action-bar>
```
The action bar doesn't render expanded. 

### Expected Behavior

My expectation is that expand-disabled only remove the UX at the bottom of the action-bar ">>" and "<< Collapse".
I still want to be able to control if the action-bar is expanded independently of the expand-diabled.

### Reproduction Sample

[See code in "Actual Behavior"](https://codepen.io/ffaubry/pen/LYOXjMd?editors=100)

### Reproduction Steps

See code in "Actual Behavior"

### Reproduction Version

beta.77

PR #4241, opened 2022-03-15
fix(action-bar, action-pad): allow expanding via API when expandDisabled. #4169

driskull:
**Related Issue:** #4169

## Summary

fix(action-bar, action-pad): allow expanding via API when expandDisabled. #4169

Allows action-bar/action-pad to still change `expanded` state even when `expandDisabled` is true. `expandDisabled` only applies to showing the expand/collapse button at the bottom.


jcfranco (approved):
💥

Issue #4066, opened 2022-02-07
Fab: When used as dropdown trigger, can't activate dropdown with space key


benelan:
### Actual Behavior

Same issue as #3720 for `calcite-fab`

### Expected Behavior

The dropdown should stay open when pressing space (same resolution as #3720)

### Reproduction Sample

https://codepen.io/ellupp/pen/LYyKQwq

### Reproduction Steps

1. Focus the fab and press space, notice the dropdown does not stay open
2. Do the same with button/action, and notice that the dropdown stays open

### Reproduction Version

@esri/calcite-components@1.0.0-76

### Relevant Info

First reported by @ellenupp
___
> We shouldn't have to keep blacklisting elements in the keyDownHandler when a different button type of element is slotted. Need to refactor nodeType checking or create a document for which node types are allowed in dropdown trigger. That way we can maintain this list of exclusion nodes.
Related PR: https://github.com/Esri/calcite-components/pull/3857

-- Eliza in #3880

The refactor Eliza described above and this fix can be done together.

driskull:
@benelan I think this should be handled via https://github.com/Esri/calcite-components/issues/3880

We can't keep adding components to the code inside dropdown. This "blocklist" isn't maintainable.

geospatialem:
Verified in next ✔️ 

![screenshot with dropdown open using space bar](https://user-images.githubusercontent.com/5023024/161615855-9d50c263-66ad-4089-9439-d68149cf9563.png)

Issue #3880, opened 2022-01-07
Refactor(dropdown): nodeType checking for elements used in the dropdown-trigger slot of calcite-dropdown


Elijbet:
### Description

We shouldn't have to keep blacklisting elements in the keyDownHandler when a different button type of element is slotted. Need to refactor nodeType checking or create a document for which node types are allowed in dropdown trigger. That way we can maintain this list of exclusion nodes.

Related PR: #3857

benelan:
Closing this issue in favor of bug #4066, which should take this refactor into account when implementing the fix. The fix will show up on the changelog unlike this refactor issue.

PR #4243, opened 2022-03-15
fix(dropdown): trigger keydown handler should only work for trigger element

driskull:
**Related Issue:**  #4066 #3880

## Summary

fix(dropdown): trigger keydown handler should only work for trigger element.

### DIFF
--- a/src/components/dropdown/dropdown.tsx
+++ b/src/components/dropdown/dropdown.tsx
@@ -438,28 +438,27 @@ export class Dropdown implements InteractiveComponent {
   }
 
   private keyDownHandler = (e: KeyboardEvent): void => {
-    const target = e.target as HTMLSlotElement;
+    const target = e.target as HTMLElement;
+
+    if (target !== this.referenceEl) {
+      return;
+    }
+
     const key = e.key;
-    if (
-      this.triggers.includes(target) ||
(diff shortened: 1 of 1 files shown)


driskull:
@jcfranco can you review these changes? What I'm doing seems to work, just wondering why the previous code was there.

src/components/dropdown/dropdown.tsx
```diff
@@ -438,28 +438,27 @@ export class Dropdown implements InteractiveComponent {
...
-          break;
-      }
+
+    if (this.active && (key === "Escape" || (e.shiftKey && key === "Tab"))) {
+      this.closeCalciteDropdown();
+      return;
+    }
+
+    switch (key) {
+      case " ":
```
jcfranco, line 455: Sidebar: this could use the `isActionKey` util if we switch (ba dum tss 🥁 ) to an `if`/`else` statement.

jcfranco (approved):
LGTM! 

We'll assume existing tests will cover this since writing a test for all acceptable triggers would be a pain to maintain. 💥👊💥

I had one refactor suggestion, but that can be tackled separately.

jcfranco:
cc @Elijbet

Issue #4215, opened 2022-03-08
Bug: calcite-dropdown-group attributes not respected when using custom elements bundle


jul11557:
### Actual Behavior

When in multi-select mode, the calcite-dropdown does not show checkmarks and behaves as if it is still in "single" selection mode.

### Expected Behavior

- [ ] The multi-select checkmarks should display if the dropdown is used in a custom widget
- [ ] Users should be allowed to select multiple items from list

### Reproduction Sample
https://github.com/tre11178/jsapi-custom-widget
https://developers.arcgis.com/calcite-design-system/components/dropdown/

### Reproduction Steps

To reproduce the issue from the repository:
1. Download the repository
2. npm install
3. npm run-script build
4. npm run-script serve
5. Connect to localhost:<port number> in browser
6. Click open the Dropdown on the top-right of the Map
7. Select multiple items from the "Multi" fields and observe

When you open it in the browser, use the dropdown and notice behavior with the multi-select

### Reproduction Version

1.0.0-beta.77

jcfranco:
This may be a duplicate of https://github.com/Esri/calcite-components/issues/3368.

jul11557:
@jcfranco it does look to us like it may be the same issue. Thanks for catching, we did a search for duplicates and didn't see that one. Does this issue plus #3368 provide enough info for you all to look into resolving the bug?

jcfranco:
> Does this issue plus https://github.com/Esri/calcite-components/issues/3368 provide enough info for you all to look into resolving the bug?

@jul11557 I believe so, thanks!

geospatialem:
Verified in beta.80: 

![screenshot of multi select dropdown option behavior](https://user-images.githubusercontent.com/5023024/160934907-b495ed74-058a-4a37-aa8f-ab2cc77f5b0d.png)

PR #4244, opened 2022-03-16
fix(dropdown-item): ensure dropdown items are initialized properly

jcfranco:
**Related Issue:** #4215 

## Summary

This fixes an issue that shows up in the custom-elements output target where a dropdown item is initialized before its ancestors are ready. I am not sure if it's possible to cover this in our automated test suite. I'll have to investigate.

I'll create a separate issue to go over our components to identify similar cases.


driskull (approved):
👍

Issue #4208, opened 2022-03-07
Pick List Item in Pick List Group - Removable prop causing entire Group to be removed


apfister:
### Actual Behavior

When using a Pick List Item with its `removable` property in a Pick List Group, when the "X" is clicked to remove the item, the entire Pick List Group is removed.

### Expected Behavior

Only the Pick List item in the group should be removed, not the entire Group and any other items.

### Reproduction Sample

https://codepen.io/apfister/pen/LYOvJKG?editors=1000

### Reproduction Steps

1. Open the repro codepen
2. Click on the "X" next to the only removable item
3. Notice the entire group is removed

### Reproduction Version

beta.77

geospatialem:
Verified in `next` 👍🏻

PR #4245, opened 2022-03-16
fix(pick-list): only remove group if removable item is slotted as parent

jcfranco:
**Related Issue:** #4208 

## Summary

This fixes an issue where grouped pick list items were removed regardless if they were slotted as a group's parent item or not.


driskull (approved):
💯 Sweeet. Knockin out the bugs 🐛

Issue #4174, opened 2022-02-28
Slider does not reset when a form is reset


benelan:
### Actual Behavior

`calcite-slider` does not reset when a form is reset.

### Expected Behavior

I expect `calcite-slider` to reset when a form is reset.

### Reproduction Sample

https://codepen.io/benesri/pen/QWOZPgW?editors=1010

### Reproduction Steps

steps in sample

### Reproduction Version

beta.77

geospatialem:
verified on next 👍🏻

PR #4249, opened 2022-03-16
fix(slider): Properly reset and submit form value when a range is used

driskull:
**Related Issue:** #4174

## Summary

fix(slider): Properly reset and submit form value when a range is used. #4174

- Changes `value` type to allow for an array.
- The change to value is to support form resetting but also form submission. Currently, form submission wouldn't submit values for range sliders.
- We might want to update clamping/stepping for range values later

### DIFF
--- a/src/components/slider/slider.tsx
+++ b/src/components/slider/slider.tsx
@@ -126,7 +126,7 @@ export class Slider implements LabelableComponent, FormComponent, InteractiveCom
   @Prop() ticks?: number;
 
   /** Currently selected number (if single select) */
-  @Prop({ reflect: true, mutable: true }) value: null | number | number[] = null;
+  @Prop({ reflect: true, mutable: true }) value: null | number | number[] = 0;
 
   @Watch("value")
   valueHandler(): void {
@@ -166,11 +166,11 @@ export class Slider implements LabelableComponent, FormComponent, InteractiveCom
   componentWillLoad(): void {
     this.isRange = !!(this.maxValue || this.maxValue === 0);
     this.tickValues = this.generateTickValues();
-    if (typeof this.value === "number") {
+    if (!Array.isArray(this.value)) {
       this.value = this.clamp(this.value);
     }
     afterConnectDefaultValueSet(this, this.value);
-    if (this.snap && typeof this.value === "number") {
+    if (this.snap && !Array.isArray(this.value)) {
       this.value = this.getClosestStep(this.value);
     }
     if (this.histogram) {
@@ -194,10 +194,10 @@ export class Slider implements LabelableComponent, FormComponent, InteractiveCom
 
   render(): VNode {
     const id = this.el.id || this.guid;
+    const maxProp = Array.isArray(this.value) ? "maxValue" : "value";
+    const value = Array.isArray(this.value) ? this.maxValue : this.value;
     const min = this.minValue || this.min;
-    const max = this.maxValue || (typeof this.value === "number" && this.value);
-    const maxProp = this.isRange ? "maxValue" : "value";
-    const value = this[maxProp];
+    const max = this.maxValue || value;
     const useMinValue = this.shouldUseMinValue();
     const minInterval = this.getUnitInterval(useMinValue ? this.minValue : min) * 100;
     const maxInterval = this.getUnitInterval(max) * 100;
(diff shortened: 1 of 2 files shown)


src/components/slider/slider.tsx
```diff
@@ -153,9 +166,11 @@ export class Slider implements LabelableComponent, FormComponent, InteractiveCom
   componentWillLoad(): void {
     this.isRange = !!(this.maxValue || this.maxValue === 0);
     this.tickValues = this.generateTickValues();
-    this.value = this.clamp(this.value);
+    if (!Array.isArray(this.value)) {
+      this.value = this.clamp(this.value);
+    }
```
driskull, line 171: I tried clamping range values but it was messing up some tests.

src/components/slider/slider.tsx
```diff
@@ -153,9 +166,11 @@ export class Slider implements LabelableComponent, FormComponent, InteractiveCom
...
     afterConnectDefaultValueSet(this, this.value);
-    if (this.snap) {
+    if (this.snap && !Array.isArray(this.value)) {
       this.value = this.getClosestStep(this.value);
     }
```
driskull, line 175: I tried doing this for range values but it was messing up some tests.

src/components/slider/slider.tsx
```diff
@@ -179,10 +194,10 @@ export class Slider implements LabelableComponent, FormComponent, InteractiveCom
 
   render(): VNode {
     const id = this.el.id || this.guid;
+    const maxProp = Array.isArray(this.value) ? "maxValue" : "value";
+    const value = Array.isArray(this.value) ? this.maxValue : this.value;
     const min = this.minValue || this.min;
-    const max = this.maxValue || this.value;
-    const maxProp = this.isRange ? "maxValue" : "value";
-    const value = this[maxProp];
+    const max = this.maxValue || value;
```
jcfranco, line 200: With the recent changes, isn't `max` === `value` here? 

```tsx
const value = 
  Array.isArray(this.value) ? 
    this.maxValue : // if we have a range, `value` will be `this.maxValue` (a)
    this.value; // else it's `this.value` (b)

// if we have a range, `this.maxValue` will be defined and assigned to `value` already (a)
// if not, it will be `this.value` (b)
const max = this.maxValue || value;
```

src/components/slider/slider.tsx
```diff
@@ -647,7 +662,7 @@ export class Slider implements LabelableComponent, FormComponent, InteractiveCom
         class="graph"
         colorStops={this.histogramStops}
         data={this.histogram}
-        highlightMax={this.isRange ? this.maxValue : this.value}
+        highlightMax={Array.isArray(this.value) ? this.maxValue : this.value}
```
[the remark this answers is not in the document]
driskull: I used the array check so that it wouldn't think that highlightMax is being set as an array. Type safety. I can see if I can remove it but there might be some typing issue

src/components/slider/slider.tsx
driskull: > Can you share the test failures you noticed? It could be that we need to update the tests.

I was going to go down that route but stopped because its really out of scope. Maybe another issue if we need to clamp or set closestStep.

The failures I saw was that it was that when handles were overlapping it was automatically moving them. 5 => 10.

src/components/slider/slider.tsx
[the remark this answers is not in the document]
driskull: This won't work right...

image

src/components/slider/slider.tsx
driskull: I think max is still used when its not a range. I don't think the logic changed right?

`const max = this.maxValue || this.value;`

src/components/slider/slider.tsx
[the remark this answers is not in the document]
driskull: Problem with `isRange` is that you lose type safety. :\

src/components/slider/slider.tsx
jcfranco: Correct, the logic didn't change, but the assignments could be simplified since I think all of these are equivalent now:

```ts
const max = Array.isArray(this.value)? value[1]: value;
const max = this.maxValue || this.value;
const max = value;
```

jcfranco (approved):
Once the 'is range' typing is updated, this is good to merge! 

Can you create a separate issue if you think we need to look into clamping range values?

driskull:
@jcfranco pushed review fixes

src/components/slider/slider.tsx
```diff
@@ -126,7 +130,18 @@ export class Slider implements LabelableComponent, FormComponent, InteractiveCom
   @Prop() ticks?: number;
 
   /** Currently selected number (if single select) */
```
dasa, line 132: Should this doc be updated too?

Issue #4279, opened 2022-03-21
Calcite-tooltip sometimes throws browser console error


arxit-nis:
### Actual Behavior

Using calcite-tooltip components is sometimes producing a console error (Chrome browser).
The error is : `o.getElementById is not a function` at https://js.arcgis.com/calcite-components/1.0.0-beta.77/p-2e29554c.js

### Expected Behavior

Even if the behavior of the component seems to be still OK, the error should not appear in the console.

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/tooltip/

### Reproduction Steps

1. Open a new tab in your Chrome browser with the devtools opened.
2. Open the official calcite-tooltip documentation page
3.  Update the component properties and / or refresh configuration from the sample section.
4.  The error in the console should appear.
![calcite-tooltip-error](https://user-images.githubusercontent.com/95091581/159251673-194fd3bd-cf2a-4fbd-9ddc-d775536de05f.png)
 

### Reproduction Version

1.0.0-beta.80

### Relevant Info

Chrome : 99.0.4844.74

geospatialem:
Verified on [tooltip documentation site](https://developers.arcgis.com/calcite-design-system/components/tooltip) 👍🏻

PR #4280, opened 2022-03-21
fix(tooltip): Removes console error when querying the tooltip

driskull:
**Related Issue:** #4279

## Summary

fix(tooltip): Removes console error when querying the tooltip (#4279)

### DIFF
--- a/src/utils/dom.ts
+++ b/src/utils/dom.ts
@@ -116,7 +116,11 @@ export function queryElementRoots<T extends Element = Element>(
 
     const found = id
       ? "getElementById" in rootNode
-        ? (rootNode.getElementById(id) as Element as T)
+        ? /*
+          Check to make sure 'getElementById' exists in cases where element is no longer connected to the DOM and getRootNode() returns the element.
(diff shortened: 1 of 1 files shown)


jcfranco (approved):
Can you add a test before merging? Other than that, this LGTM! 🎉

driskull:
@jcfranco the error occurs when typing a value in fast. For example spamming numbers in the `offset-skidding`

The root of the error stems from `el.getRootNode()` for some reason returning the component element instead of its rootNode.

Like this...

image

However, getRootNode should return a shadowRoot or document. https://developer.mozilla.org/en-US/docs/Web/API/Node/getRootNode

It might be some timing issue with the iframe reloading while executing.

So this PR as it is won't fix this. Any ideas on how I can best test for this?

driskull:
Looking into it more, the problem is that sometimes `getRootNode()` returns the tooltipmanager. My guess is that it happens when the element is no longer connected to the DOM.

So when that happens, we can just disregard the query.

I've updated the PR to do that.

Issue #2033, opened 2021-04-23
Bug: Ensure change events are only emitted on user interaction


jcfranco:
### Actual Behavior

Some components will emit a change event when `value` is set programmatically (e.g., color-picker) and also after user interaction. This would be consistent with how `<input>` works.

### Expected Behavior

Change events should only emit after the user interacts with a component and not when setting the changed property programmatically.

This aligns [with the HTML spec on input/change events](https://html.spec.whatwg.org/multipage/input.html#common-input-element-events):

> When the input and change events apply (which is the case for all input controls other than buttons and those with the type attribute in the Hidden state), the events are fired to indicate that the user has interacted with the control. The input event fires whenever the user has modified the data of the control. The change event fires when the value is committed, if that makes sense for the control, or else when the control loses focus. In all cases, the input event comes before the corresponding change event (if any).

### Reproduction Steps

https://codepen.io/jcfranco/pen/yLgGmLK?editors=1000

##### Component/Event candidates

- [x] `block-section.tsx` - already emits on user interaction only (@anveshmekala)
  - [x] `93 this.calciteBlockSectionToggle.emit();`
- [x] `block.tsx` -already emits on user interaction only (@anveshmekala)
  - [ ] `150 this.calciteBlockToggle.emit();`
- [x] `card.tsx` – already emits on user interaction only @driskull 
  - [x] `138 this.calciteCardSelect.emit();`
- [x] `checkbox.tsx` – already emits on user interaction only
  - [x] `127 this.calciteCheckboxChange.emit();`
- [x] `color-picker-hex-input.tsx` (@jcfranco) – https://github.com/Esri/calcite-components/pull/4288
  - [x] `119 this.calciteColorPickerHexInputChange.emit();`
  - [x] `127 this.calciteColorPickerHexInputChange.emit();`
  - [x] `181 this.calciteColorPickerHexInputChange.emit();`
- [x] `color-picker.tsx` (@jcfranco) – https://github.com/Esri/calcite-components/pull/4288
  - [x] `266 this.calciteColorPickerInput.emit();`
  - [x] `268 this.calciteColorPickerChange.emit();`
  - [x] `279 this.calciteColorPickerInput.emit();`
  - [x] `281 this.calciteColorPickerChange.emit();`
  - [x] `553 this.calciteColorPickerChange.emit();`
- [x] `date-picker-day.tsx` – already emits on user interaction only @driskull 
[...]

PR #4288, opened 2022-03-23
fix(color-picker)!: emit color change and input events on user-interaction only

jcfranco:
**Related Issue:** #2033 

## Summary

This updates the color-picker to only emit change/input events on user interaction and not when updated programmatically.


src/components/color-picker-hex-input/color-picker-hex-input.tsx
```diff
@@ -288,6 +242,42 @@ export class ColorPickerHexInput {
   //
   //--------------------------------------------------------------------------
 
+  private internalSetValue(
+    value: string | null,
+    oldValue: string | null,
+    // skipEqual = true,
```
driskull, line 248: wats dis? skipEqual stuff. Can we remove it?

jcfranco: Whoopsie. I'll remove. Copypasta from the picker's internal value setter. 🍝

PR #4302, opened 2022-03-24
ci(screener): integrating sauce labs without build failures

eriklharper:
**Related Issue:** n/a

## Summary

This PR is setup for testing just the Sauce Labs integration with Screener.


eriklharper:
We're going to keep Screener as-is for now.

PR #4303, opened 2022-03-24
build: add version to preamble

benelan:
**Related Issue:** NA

## Summary
Appends the CC version number to the preamble so users can find out which version a product is using. I had to change the `release:prepare` npm script to build after bumping the version number rather than before.

For example, in the browser dev tools for Map Viewer you can't find a version:
image

This PR will display the version in the preamble comment like:
```js
/*!
 * All material copyright ESRI, All Rights Reserved, unless otherwise specified.
 * See https://github.com/Esri/calcite-components/blob/master/LICENSE.md for details.
 * v1.0.0-beta.80
 */
```

### Edit
If we are worried about character/line count, I can replace `master` with the version instead, like:
```js
/*!
 * All material copyright ESRI, All Rights Reserved, unless otherwise specified.
 * See https://github.com/Esri/calcite-components/blob/v1.0.0-beta.80/LICENSE.md for details.
 */
```
However, the link won't work if the GitHub tag isn't created yet, or if there is a typo on the tag. That _shouldn't_ happen since the process is automated. Also, if we ever have to change the license, it will be automatically updated for everyone on `master`. But if we use the version tag, then people on old versions will see an old license.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "prepublishOnly": "ts-node ./support/prepublish.ts",
     "release:docs": "npm run docs && storybook-to-ghpages --existing-output-dir=docs",
     "release:next": "npm run util:clean-tested-build && npm run util:deploy-next-from-existing-build",
-    "release:prepare": "npm run util:clean-tested-build && ts-node --project ./tsconfig-node-scripts.json support/prepReleaseCommit.ts && git add .",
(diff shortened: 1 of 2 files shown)


driskull (approved):

jcfranco (approved):
🎉

Issue #4265, opened 2022-03-17
Bug: [combobox] Mac Fn + right/left (home / end) should toggle corresponding first/last item into focus & scroll into view after opening listbox


Elijbet:
### Actual Behavior

Mac Fn+left (home) initially focuses on the first item in the open listbox, but won't scroll to the last when hitting Fn+right (end). And after you scroll down to it manually, it won't scroll back to home. 

### Expected Behavior

Mac Fn + right/left (home / end) should toggle corresponding first/last item into focus & scroll into view if open or after opening the listbox.

### Reproduction Sample

https://codepen.io/elijbet/pen/jOYWwbW

### Reproduction Steps

1. Click on Combobox to open.
2. Click home/fn+leftArrow and see it focus on the first item.
3. Click end/fn+righArrow and notice it won't scroll to last item.</p>
4. Scroll to last item manually and see that it is highlighted.
5. Click home/fn+leftArrow and notice that it won't scroll back to the highlighted first item.

### Reproduction Version

v1.0.0-beta.80

### Relevant Info

Follow up on PR #4218

Once the last item in focus scrolls into view, update the combobox.e2e.tx test with: 
```
    it(`End opens dropdown and puts focus on last item`, async () => {
      const inputEl = await page.find(`#parentOne >>> input`);
      await inputEl.focus();
      await page.waitForChanges();
      expect(await page.evaluate(() => document.activeElement.id)).toBe("parentOne");

      await page.keyboard.press("End");
      await page.waitForChanges();
      const lastFocusedGroupItem = await page.find("#three >>> .label--active");
      expect(lastFocusedGroupItem).toBeTruthy();
      const visible = await lastFocusedGroupItem.isVisible();
      expect(visible).toBe(true);
    });
```

benelan:
@jcfranco / @macandcheese can one of you verify this one? Kitty and I don't have a Mac

geospatialem:
Was able to test on my personal Mac - verified on next. ✅ Also helps create a better experience for Windows users as well! 👍🏻

PR #4304, opened 2022-03-24
fix(combobox): listbox item focused with home or end now scrolls into view

Elijbet:
**Related Issue:** #4265

## Summary

This discussion prompted several changes for Combobox to be consistent with the Dropdown behavior:

- when the combobox is focused but closed, End/Home (fn arrow left/right) scrolls the entire page to the very top/bottom and won’t keep the focused combobox in view
- Page up/down (fn arrow up/down) scroll up and down as well
- while it’s open it'll scroll the listbox to focused item into view on Home/End or ArrowUp/Down 
- when you open the listbox, the focus jumps to the first item
- in the situation where the listbox is open and overflows the page also scroll on these keystrokes to bring the combobox item into view.

Includes adjustments to related failing tests.


src/components/combobox/combobox.tsx
```diff
@@ -382,6 +382,12 @@ export class Combobox implements LabelableComponent, FormComponent, InteractiveC
 
   keydownHandler = (event: KeyboardEvent): void => {
     const { key } = event;
+    const bounding = this.el.getBoundingClientRect();
+    const inViewport =
```
jcfranco, line 386: Can you refactor the viewport-checking code to only run when needed (i.e., to run when the pressed key is related to item navigation)?

src/components/combobox/combobox.tsx
```diff
@@ -880,6 +883,13 @@ export class Combobox implements LabelableComponent, FormComponent, InteractiveC
     } else if (activeItem.offsetTop < scrollTop) {
       this.listContainerEl.scrollTop = activeItem.offsetTop;
     }
+  };
+
+  shiftActiveItemIndex(delta: number): void {
+    const length = this.visibleItems.length;
```
jcfranco, line 889: Nitpick: you could use destructuring if the assigned variable has the same name:

```
const { length } = this.visibleItems;
```

src/components/combobox/combobox.tsx
```diff
@@ -867,11 +874,7 @@ export class Combobox implements LabelableComponent, FormComponent, InteractiveC
     chip?.setFocus();
   }
 
-  shiftActiveItemIndex(delta: number): void {
-    const length = this.visibleItems.length;
-    const newIndex = (this.activeItemIndex + length + delta) % length;
-    this.updateActiveItemIndex(newIndex);
-    // ensure active item is in view if we have scrolling
+  private activeInView = (): void => {
```
jcfranco, line 877: Could you rename this to better express what it does? Currently, it reads to me like it will return a boolean if the active (item?) is in view.

src/components/combobox/combobox.tsx
```diff
@@ -419,14 +423,16 @@ export class Combobox implements LabelableComponent, FormComponent, InteractiveC
         }
         break;
       case "Home":
-        event.preventDefault();
-        this.active = true;
+        this.active ? event.preventDefault() : null;
         this.updateActiveItemIndex(0);
+        this.activeInView();
+        !inViewport ? this.el.scrollIntoView({ block: "start", inline: "nearest" }) : null;
```
jcfranco, line 429: Unfortunately, `scrollIntoView` in Safari does not support the options param: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView#browser_compatibility.

src/components/combobox/combobox.tsx
```diff
@@ -419,14 +423,16 @@ export class Combobox implements LabelableComponent, FormComponent, InteractiveC
         }
         break;
       case "Home":
-        event.preventDefault();
-        this.active = true;
+        this.active ? event.preventDefault() : null;
```
jcfranco, line 426: Can you use `if` statements instead of using conditionals? The latter is more commonly used for assigning values. This applies to other lines in this method.

src/components/combobox/combobox.tsx
```diff
@@ -382,6 +382,12 @@ export class Combobox implements LabelableComponent, FormComponent, InteractiveC
 
   keydownHandler = (event: KeyboardEvent): void => {
     const { key } = event;
+    const bounding = this.el.getBoundingClientRect();
```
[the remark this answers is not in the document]
Elijbet: I think because of popper repositioning, it's the combobox that often overflows the page. This brings all of it into view. Does that work?

src/components/combobox/combobox.e2e.ts
```diff
@@ -571,6 +573,39 @@ describe("calcite-combobox", () => {
...
+      await combobox.callMethod(`setFocus`);
+      const popper = await page.find("#myCombobox >>> .popper-container--active");
+      expect(popper).toBeNull();
+      expect(await page.evaluate(() => window.scrollY)).toEqual(0);
+
+      await page.keyboard.press("PageDown");
+      await page.waitForChanges();
+      expect(await page.evaluate(() => window.scrollY)).toBeTruthy();
+
+      const previousScrollPosition = await page.evaluate(() => window.scrollY);
```
jcfranco, line 599: Nitpick: you can store the `scrollY` from above and reuse here.

src/components/combobox/combobox.e2e.ts
```diff
@@ -571,6 +573,39 @@ describe("calcite-combobox", () => {
...
+      await page.keyboard.press("PageUp");
+      await page.waitForChanges();
+      expect(
+        await page.evaluate((previousScrollPosition) => {
+          return window.scrollY < previousScrollPosition;
```
jcfranco, line 604: Can you add a comment or store this value in a variable to express what it represents?

src/components/combobox/combobox.e2e.ts
```diff
@@ -473,6 +474,7 @@ describe("calcite-combobox", () => {
 
   describe("keyboard navigation", () => {
     let page: E2EPage;
+    let scrollablePageSizeInPx: number;
```
jcfranco, line 477: This doesn't change, so let's assign the value here.

src/components/combobox/combobox.e2e.ts
```diff
@@ -571,6 +573,39 @@ describe("calcite-combobox", () => {
...
     });
 
+    it("when the combobox is focused & closed, Page up/down (fn arrow up/down) scrolls up and down the page", async () => {
+      const scrollablePageSizeInPx = 2400;
+      await page.addStyleTag({
+        content: `body {
+            height: ${scrollablePageSizeInPx}px;
+            width: ${scrollablePageSizeInPx}px;
+          }
+          html, body {
```
jcfranco, line 583: Can you add a comment explaining why this is used?

jcfranco, line 583: Can you add a comment explaining why this is needed?

src/components/combobox/combobox.tsx
```diff
@@ -397,6 +397,16 @@ export class Combobox implements LabelableComponent, FormComponent, InteractiveC
     this.setFocus();
   };
 
+  inViewport = (): boolean => {
```
jcfranco, line 400: Nitpick: this can be a class method and not an assigned function. Reasoning: no need for a bound function here.

src/components/combobox/combobox.tsx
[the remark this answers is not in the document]
jcfranco: I don't follow. Is `inViewport` meant to detect if the combobox is in view or the item? If it's the latter, this method needs to be updated since it's checking if the combobox (not the items) is in view, right?

src/components/combobox/utils.ts
```diff
@@ -39,3 +40,7 @@ export function getDepth(element: HTMLElement): number {
 
   return result.snapshotLength;
 }
+
+export async function scrollTo(x: number, y: number, page: E2EPage): Promise<void> {
```
jcfranco, line 44: Is this used? I think I had suggested possibly creating a reusable function for tests, not the component. If not needed at the moment, let's remove it.

src/components/combobox/combobox.tsx
```diff
@@ -419,14 +429,16 @@ export class Combobox implements LabelableComponent, FormComponent, InteractiveC
...
-        event.preventDefault();
-        this.active = true;
         this.shiftActiveItemIndex(-1);
+        if (!this.inViewport()) {
+          this.el.scrollIntoView();
+        }
         break;
       case "ArrowDown":
-        event.preventDefault();
-        this.active = true;
```
jcfranco, line 428: I thought ⬇️ would:

1. when closed, open the menu
2. when open, navigate to the next item

jcfranco (commented):
Looking good!

src/components/combobox/combobox.tsx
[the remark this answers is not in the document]
Elijbet: What I was trying to say, is because listbox flips, combobox is the first thing that overflows, so if you bring that in viewport, both combobox and listbox will be in view. This happens when you go up/down on list item. The only case of focusing the combobox and still having the items be out of view would be if popper didn’t flip for some reason. I wanted to make sure combobox is in view every time you interact with listbox is because you'd want to be aware of the input in it. I’ll rename the `inViewport` method so it's a bit more descriptive.

src/components/combobox/combobox.e2e.ts
```diff
@@ -514,7 +516,7 @@ describe("calcite-combobox", () => {
       await page.waitForChanges();
       expect(await page.evaluate(() => document.activeElement.id)).toBe("myCombobox");
 
-      await page.keyboard.press("ArrowDown");
+      await page.keyboard.press("Space");
```
jcfranco, line 519: Do we also need a test for `ArrowDown` opening the menu?

Elijbet: 💡 There is a test below this one that does that.

PR #4307, opened 2022-03-25
ci: disable merge squash prepush due to windows error

benelan:
**Related Issue:** NA

## Summary
Kitty ran into an error pushing due to the prepush hook which ensures conventional commits. 
2a8c01f3-027d-4a7d-9b20-0da2abba3486

Relevant bit:
`Error: Command failed: git log --format=%B geospatialem/test --not master | tr '`

It looks like its an issue due to Windows. I disabled the hook so she can push and I'll work on making it OS agnostic.

PR #4309, opened 2022-03-25
ci: squash mergeable script os agnostic

benelan:
**Related Issue:** #4307

## Summary
Makes the squash-mergeable prepush script work on Windows (Kitty tested and confirmed). It also still works on Unix:
image


jcfranco (approved):
💥💻💥

Issue #4294, opened 2022-03-23
Optionally keep tooltip open when user clicks on reference element


noahmulfinger:
### Description

Currently when a tooltip is attached to a button, the tooltip will open on hover and then close immediately when clicked. This causes a problem in my use-case (a click-to-copy button) where the text of the tooltip changes once the user clicks. The tooltip closes before user can see the change in the text. In previous versions of calcite components (I last checked beta.55), the tooltip stayed open. I would like a way to keep that behavior in the current version.

Alternatively, if this is not a recommended pattern when using tooltips, suggestions for a different setup would be welcome.

### Acceptance Criteria

There is a way to keep the calcite-tooltip open even when the reference element is clicked.

### Which Component

calcite-tooltip, calcite-tooltip-manager

### Example Use Case

Simple click-to-copy button where the text changes to `Copied` once the user clicks the button, but its difficult to see the change as the tooltip is closing.
https://codesandbox.io/s/competent-sutherland-ifzx1w?file=/src/App.js

driskull:
I can revert this back. I thought it might be nice to just get it the tooltip out of the way once a user clicks on the reference element. @jcfranco what do you think?

macandcheese:
I think this would be a nice option. There are cases like this where you may want to persist it - in this case you can't see the "Copied!" success text. We encounter the same issue in the docs site component preview when clicking to copy the code : https://developers.arcgis.com/calcite-design-system/components/notice/#sample

geospatialem:
Verified the tooltip does not close on click in `0.22.0`. 🖱️

macandcheese:
Seems like this could be better an opt-in / out behavior somehow? Some feedback on the functionality -https://github.com/Esri/calcite-components/issues/4399

geospatialem:
@macandcheese Agreed, the reverted behavior doesn't seem useful for all use cases. 

Maybe we can leverage that issue to add in an option, as discussed above, so the default functionality closes on click, but if the option is enabled, would have the functionality disabled. Perhaps also allowing the user to define the timeout.

What are your thoughts on adding this functionality as an option, as part of #4399 @driskull?

PR #4310, opened 2022-03-25
fix(tooltip-manager): Don't close tooltips on click. (#4294)

driskull:
**Related Issue:** #4294

## Summary

fix(tooltip-manager): Don't close tooltips on click. (#4294)


jcfranco (approved):
💥

PR #4311, opened 2022-03-25
chore: move @types/color to devDep

benelan:
**Related Issue:** NA

## Summary
Type packages belong in devDeps


jcfranco:
`@types/color` needs to be a dependency because of https://github.com/Esri/calcite-components/blob/master/src/components/color-picker/color-picker.tsx#L73. See https://github.com/Esri/calcite-components/pull/775.

Issue #4305, opened 2022-03-25
Stepper in custom element - next/previous methods have no visual effect


benelan:
### Actual Behavior

When stepper is in a custom element, the `nextStep` and `prevStep` methods do not visually change position. However, the `calciteStepperItemChange` event does trigger and the position is changed internally.

### Expected Behavior

Stepper works like it does when it is not inside a custom element.

### Reproduction Sample

https://codepen.io/benesri/pen/MWrJqzj

### Reproduction Steps

1. Open the repro sample
2. Open the dev tools console
3. Click the "Previous Step" and "Next Step" buttons
4. Notice that there are no visual changes to the stepper. 
5. In the console the `calciteStepperItemChange` is logged and the `position` is changed internally
6. Repeat in [this codepen](https://codepen.io/benesri/pen/PoEWKXv), which works correctly

### Reproduction Version

beta.80

### Relevant Info

In the repro sample, the `calciteStepperItemChange` event target is `custom-element`. It looks like it isn't visually changing position because the [target element needs to be the parent `calcite-stepper`](https://github.com/Esri/calcite-components/blob/master/src/components/stepper-item/stepper-item.tsx#L185-L188).

geospatialem:
verified in next. 🎉

PR #4313, opened 2022-03-26
fix(stepper): visually change position using component methods when inside a custom element

benelan:
**Related Issue:** #4305

## Summary
The issue which only occurs when stepper is inside a custom element. Any ideas about how to add an e2e test for this?


jcfranco (approved):
Nice! For test inspiration, here's a couple of existing ones:

* https://github.com/Esri/calcite-components/blob/7a8b6dbed71dced31ba9a442afa13067b1c18d1b/src/components/tabs/tabs.e2e.ts#L211
* https://github.com/Esri/calcite-components/blob/19a16577ed55f48ea107bb4955c968869fee8bdf/src/components/dropdown/dropdown.e2e.ts#L894

Issue #4211, opened 2022-03-08
Calling `click()` on switch element does not toggle


jcfranco:
### Actual Behavior

`HTMLCalciteSwitchElement#click()` does not toggle.

via @dbouwman 

> We just bumped from .69 -> .78 and our webdriver.io tests started failing. After digging around, it seems that a regression was introduced to <calcite-switch> at .70. The issue is that calling .click() on a reference to the switch element does nothing at .70, where as before that it toggled the state.
> 
> Beta.69 Example: https://codepen.io/dbouwman/pen/vYWbKKB?editors=1000
> Beta.70 Example: https://codepen.io/dbouwman/pen/gOXqMow?editors=1000
> 
> Behavior still present @ beta.78
> 
> Looking at the code, there is a test which seems to explicitly test this... but it's not working in the codepen or selenium/webdriver... any input appreciated

### Expected Behavior

`HTMLCalciteSwitchElement#click()` toggles the component on/off.

### Reproduction Sample

https://codepen.io/dbouwman/pen/gOXqMow?editors=1000 

### Reproduction Steps

1. get a reference to the switch element
2. call `click()` on it
3. notice there is no change

### Reproduction Version

beta.78

### Regression?

beta.70

benelan:
verified on beta.81

PR #4320, opened 2022-03-28
fix(switch): Calling click() on switch should toggle checked

driskull:
**Related Issue:** #4211

## Summary

fix(switch): Calling click() on switch should toggle checked. (#4211)


jcfranco (approved):
🎉

Issue #4206, opened 2022-03-07
Set value on calcite-input-time-picker works incorrectly when default value exists


fendiaoxiaoshuazi:
### Actual Behavior

When the default value exists on the calcite-input-time-picker, the set value on the input time picker will work incorrectly.

For example, the input time picker has the default value of 11:00, when I try to set the input time picker's value to 14:30 by code, the input time picker will receive an incorrect value 14:00:00. (Note: if the input time picker does not have the default value, set value can work correctly.)

![image](https://user-images.githubusercontent.com/26078106/156959298-cfe766c2-1654-4e8e-aeac-104f8cd80879.png)

### Expected Behavior

The value should be set correctly on the input time picker when the default value exists.

### Reproduction Sample

https://codepen.io/nan9266/pen/XWzzaxM?editors=1111

### Reproduction Steps

1. Open the console
2. Click on the "change time as 14:30" button 
    - Do not manually change the value of the input time picker before clicking the button
3. Observe the UI and console both shows incorrect value

### Reproduction Version

1.0.0-beta.72

fendiaoxiaoshuazi:
The issue was found in the Survey123 JS API. When the time question has the default value, and they want to change the value through the JS function, the value cannot be set correctly.

eriklharper:
Thanks for reporting this and reproducing the issue!  I am the original author of this component and have verified this locally. It appears to be caused by the internal `calciteTimePickerChange` event listener sending an incorrect time value back up to the input component.  I can fix this by making the internal time-picker component respect direct value setting so that it doesn't interfere with direct value setting in the input component.

geospatialem:
Verified on next

![change time fix verification](https://user-images.githubusercontent.com/5023024/170576395-6f412e85-ad4a-4615-a15a-d0bb0ffabd68.png)

PR #4321, opened 2022-03-28
fix(input-time-picker, time-picker): fixing direct value setting issue where the minutes and seconds weren't respected when a default value is supplied

eriklharper:
**Related Issue:** #4206 

## Summary

This PR removes a now-redundant watcher on the `hour` property of the `time-picker` component that was causing an incorrect updating of the `minute` and `second` values before the original programmatically-set value had a chance to propagate to them.  This watcher is a remnant of a previous iteration of the component and is no longer needed because the `meridiem` is updated as part of the normal `setValue` workflow.

### DIFF
--- a/src/components/input-time-picker/input-time-picker.e2e.ts
+++ b/src/components/input-time-picker/input-time-picker.e2e.ts
@@ -192,6 +192,37 @@ describe("calcite-input-time-picker", () => {
     expect(inputTimePickerValue).toBe(expectedValue);
   });
 
+  it("value displays correctly in the input when it is programmatically changed for a 24-hour language when a default value is present", async () => {
+    const lang = "fr";
+    const page = await newE2EPage({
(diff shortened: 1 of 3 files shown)


src/components/input-time-picker/input-time-picker.e2e.ts
```diff
@@ -161,6 +161,37 @@ describe("calcite-input-time-picker", () => {
     expect(inputTimePickerValue).toBe(expectedValue);
   });
 
+  it("value displays correctly in the input when it is programmatically changed for a 12-hour language when a default value is present", async () => {
```
jcfranco, line 164: Does this need a test for the 24-hour format case?

jcfranco (approved):
🎉

src/components/input-time-picker/input-time-picker.e2e.ts
eriklharper: Sure couldn't hurt, I'll add that!

PR #4322, opened 2022-03-29
build(deps): bump versions

benelan:
This PR was automatically generated by the update-deps GitHub action


jcfranco (changes requested):
Blocking until we can get rid of that `typescript` log file.

PR #4325, opened 2022-03-29
ci(update-deps): only commit package files

benelan:
**Related Issue:** #4322

## Summary
Only add `package.json` and `package-lock.json` when committing. This is to prevent a log file from being generated, which happened in the linked PR


jcfranco (approved):
💥

Issue #4238, opened 2022-03-15
RTL - Color picker HSV/RGB fields incorrectly mirrored


jmanke:
### Summary

The RGB and HSV fields are incorrectly RTL mirrored.

### Actual Behavior

![image](https://user-images.githubusercontent.com/19629809/158449379-67431b2e-2aee-4184-ac7a-0563c5669db7.png)

### Expected Behavior

The RGB and HSV fields should not be RTL mirrored, they should be in R,G,B and H,S,V order like on EN language.

### Reproduction Sample

https://codepen.io/jmanke/pen/NWXqBRb

### Reproduction Steps

1. Set dir to 'rtl' in the root element
2. Observe RGB and HSV fields are incorrectly RTL mirrored.

### Reproduction Version

@esri/calcite-components@latest

jcfranco:
@anveshmekala I _think_ the layout within inputs is correct in RTL, it's the order of the inputs that needs to be fixed. cc @macandcheese

jmanke:
@anveshmekala @jcfranco Yes @jcfranco is correct

benelan:
verified on `beta.81`

PR #4327, opened 2022-03-29
fix(color-picker): color channels order no longer mirrors when dir='rtl'

anveshmekala:
**Related Issue:** #4238 

## Summary

Fixes the direction of  RGB and HSV channels when `dir=rtl` in `calcite-color-picker`


src/components/color-picker/color-picker.scss
```diff
@@ -143,6 +143,10 @@ $gap--large: 12px;
   flex-wrap: wrap;
 }
 
+.channels--rtl {
```
jcfranco, line 146: Wouldn't it be simpler to fix the direction of `.channels` since it should follow the same order regardless of page direction?

anveshmekala: fixing direction of `.channels` wouldn't work with the logical props.

driskull: Couldn't you just set the `direction` to always be `ltr`? Then you don't have to do a dom query to get the direction.

anveshmekala: setting the direction to `ltr` would make the `calcite-input` prefix to be `ltr` always .

jcfranco: @anveshmekala will unavailable for a while, so I've gone ahead and pushed a revisited approach. @driskull Can you review?

driskull (approved):
👍 

Looks good. It sucks we have to use getElementDir here but I get it.

jcfranco:
FYI, I accepted the following screenshot diff:

### before
componentscontrolscolorpicker-rtl-0 reference

### after
componentscontrolscolorpicker-rtl-0

Issue #4117, opened 2022-02-15
Change <calcite-button>'s default type to be "button" rather than "submit"


jcfranco:
### Description

As `calcite-button` is used more and more as a building block for other components and not just forms, we should change the default type accordingly. A few design systems/widgets also use "button" as the default type:
 
https://github.com/ing-bank/lion/tree/master/packages/button
https://github.com/shoelace-style/shoelace/blob/next/src/components/button/button.ts#L67-L71
https://github.com/microsoft/fast/tree/master/packages/web-components/fast-foundation/src/button
https://www.carbondesignsystem.com/components/button/code (can't confirm from the source, but the default story has `type=button`)
https://github.com/dojo/widgets/blob/master/src/button/index.tsx

**Note**: this will be a breaking change for form use cases using `calcite-button` w/o a type. These will now need to provide the "submit" type.

### Acceptance Criteria

`HTMLCalciteButtonElement#type` is `"button"` by default.

### Relevant Info

https://github.com/Esri/calcite-components/issues/4007 brought up an issue where components were unintentionally causing the form to submit. 

### Which Component

`calcite-input`

benelan:
verified on `next` 
https://codepen.io/benesri/pen/PoEBKxR?editors=1010

PR #4335, opened 2022-03-30
fix(button)!: Change type property to be "button" by default

driskull:
**Related Issue:** #4117

## Summary

fix(button)!: Change type property to be "button" by default. (#4117)


jcfranco (approved):
Nice!

Can you hold off on installing until release time wraps up? 

Also, right before merging, can you add a `BREAKING CHANGES:` note for users in the commit message body about this (the issue description has one you could repurpose).

Issue #2885, opened 2021-08-24
Bug: active tab "underline" not visible when when tabs contained in initially hidden panel


COV-GIS:
### Actual Behavior

As stated and shown here: https://codepen.io/COV-GIS/pen/ExXxzWj

### Expected Behavior

Active tab should be underlined when panel is shown.

Switching tabs or resizing the window causes the underline to appear.

### Relevant Info

_Version_: `@esri/calcite-components@v1.0.0-beta.62`

- [ ] CDN
- [ ] NPM package

ellenupp:
Encountering this as well. Put together a sample before I noticed this was already logged: https://codepen.io/ellupp/pen/ZEaYeMV?editors=1010

jcfranco:
Installed. 

Simplified repro case: https://codepen.io/jcfranco/pen/jOYGyBb

COV-GIS:
Thanks for working on this @jcfranco and fixing https://github.com/Esri/calcite-components/pull/4351 too.

geospatialem:
Closing this issue, verification provided by @jcfranco 👍🏻

PR #4351, opened 2022-03-31
fix(tabs): update active tab styling on resize

jcfranco:
**Related Issue:** #2885 

## Summary

This modifies `tabs` to use a resize observer instead of relying on the window being resized to recompute the active tab underline. Previously, it would not update the active tab styling when initially hidden.

### DIFF
--- a/src/components/tab-nav/tab-nav.tsx
+++ b/src/components/tab-nav/tab-nav.tsx
@@ -110,7 +110,7 @@ export class TabNav {
   }
 
   disconnectedCallback(): void {
-    this.resizeObserver?.unobserve(this.el);
+    this.resizeObserver?.disconnect();
   }
 
   componentWillLoad(): void {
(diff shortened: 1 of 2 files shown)


src/components/tab-nav/tab-nav.tsx
```diff
@@ -105,6 +106,11 @@ export class TabNav {
 
   connectedCallback(): void {
     this.parentTabsEl = this.el.closest("calcite-tabs");
+    this.resizeObserver?.observe(this.el);
+  }
+
+  disconnectedCallback(): void {
+    this.resizeObserver?.unobserve(this.el);
```
driskull, line 113: I wonder if we should just call `this.resizeObserver?.disconnect()` since we're only observing 1 element anyway. I guess it doesn't matter either way.

driskull (approved):
👍

src/components/tab-nav/tab-nav.tsx
jcfranco: Good point. I'll update it for consistency.

PR #4358, opened 2022-03-31
ci: automatically deploy storybook once a week

benelan:
**Related Issue:** 

## Summary
Automatically deploy storybook at the beginning of the week.

### Disclaimer
I didn't test this in my test repo because it seems like more work than its worth. Its pretty simple and should work, but things won't explode if it doesn't. Worst case scenario it fails and I fix it.


jcfranco (approved):
> Worst case scenario it fails and I fix it.

fix-futurama

Issue #1258, opened 2020-11-10
Bug: calcite-stepper does not work with React DOM manipulation


noahmulfinger:
### Summary

When using calcite-stepper (with layout="horizontal") in a React component, React will throw an error any time you add/remove DOM elements. I'm guessing this is because of the way calcite-stepper moves DOM elements around depending on the active step.

### Actual Behavior

React will throw the error `Node.insertBefore: Child to insert before is not a child of this node` whenever you update the DOM inside a calcite-stepper. Here's a repro case in this codesandbox: https://codesandbox.io/s/nifty-ardinghelli-28ry7?file=/src/App.js

### Expected Behavior

The calcite-stepper should work fine with React. It might require using making steps hidden instead of moving their contents into the active slot.

### Reproduction Steps

1. Open https://codesandbox.io/s/nifty-ardinghelli-28ry7?file=/src/App.js
2. Hit the `Toggle` button
3. You will see the `Node.insertBefore` error.

macandcheese:
To confirm, this only happens in "horizontal" not "vertical" layout, correct?

noahmulfinger:
@macandcheese Interesting, yeah only `horizontal`.

macandcheese:
"vertical" doesn't do anything with slotted content, just shows and hides. "Horizontal" basically functions as tabs and moves slotted content from the stepper item "up" to a single container in the parent calcite-stepper. It seems like this causes more hassle than is needed, there's probably an easier way. We might be able to try and totally re-think that.

driskull:
Looking at the stepper code, it could use quite a bit of cleanup. I see a lot of nodes being created without JSX and some IE specific testing going on.

macandcheese:
Valid

driskull:
@jcfranco @paulcpederson @macandcheese @julio8a  I think we need to revisit this component and it's API.

We could just have a property for `currentPosition` called `step` and get rid of the methods and internal events going on.

The `updateContent()` method seems pretty hacky.

Would we want to do this pre 1.0?

macandcheese:
I think maybe @eriklharper had looked at this (?) - I know it works fine in vertical mode, the bug is in the "tab" functionality in horizontal mode for React since it's (I think?) doing what label does and creating / removing Nodes after it's mounted in React

jcfranco:
> moves slotted content from the stepper item "up" to a single container in the parent calcite-stepper

[...]

PR #4367, opened 2022-04-04
fix(stepper): revisit horizontal layout approach to avoid rearranging the DOM

jcfranco:
**Related Issue:** #1258 

## Summary

This revisit the logic the stepper uses to switch between horizontal and vertical layouts. Instead of moving the content around the DOM, the stepper & item styles now use `display: grid` and `display: contents` (for horizontal layout), respectively, to accomplish the same layout.

### Notes

- `display: contents` takes the host element out of tab order, so some redundant code and styles needs to be applied to the focusable element on items when `layout=horizontal`

### Pending

* [x] update styles to have consistent spacing
* [ ] ~add a11y attributes to `stepper-item` focusable element when `layout=horizontal`~ no longer needed
* [x] update grid definition styles to accommodate any number of items
* [x] update E2E tests
* [x] DRY up CSS


src/components/stepper-item/stepper-item.tsx
```diff
@@ -135,7 +133,13 @@ export class StepperItem implements InteractiveComponent {
     return (
       <Host aria-expanded={this.active.toString()} onClick={() => this.emitRequestedItem()}>
         <div class="container">
-          <div class="stepper-item-header">
+          <div
+            class="stepper-item-header"
+            tabIndex={
+              /* additional tab index logic needed because of display: contents */
```
driskull, line 139: This part is a bit weird...

In the horizontal case, the content isn't within the "button" whereas in the vertical case, the content is within the "button". That seems like kind of a big difference. Can we make them consistent? The content is within the trigger area in one layout and not in the other.

driskull: Might be a breaking change but worth it.

driskull (commented):
Yay! this is like a gazillion times better 👍

src/components/stepper-item/stepper-item.tsx
jcfranco: Sorry for the delayed reply. I agree that both layouts having different structure is odd. In the meantime, I think we should look into addressing the immediate issue and we can discuss redesigning this. cc @macandcheese @ashetland

jcfranco:
@macandcheese @asangma I know this week is a bit hectic with UC going on, but can either of you help look at the styles and layout changes? The reported diffs from Screener seem to be laid out better, at least to me.

macandcheese:
> @macandcheese @asangma I know this week is a bit hectic with UC going on, but can either of you help look at the styles and layout changes? The reported diffs from Screener seem to be laid out better, at least to me.

Yep - agreed the changes look better, the updated css also equalizes the width of step items which is an improvement. Thanks @jcfranco

src/components/stepper-item/stepper-item.tsx
```diff
@@ -191,7 +200,17 @@ export class StepperItem implements InteractiveComponent {
...
+
+  //--------------------------------------------------------------------------
+  //
+  //  Public Methods
+  //
+  //--------------------------------------------------------------------------
+
+  @Method()
+  async setFocus(): Promise<void> {
+    (this.layout === "vertical" ? this.el : this.headerEl).focus();
```
driskull, line 213: Maybe...

```
(this.layout === "vertical" ? this.el : this.headerEl)?.focus();
```

just to be safe.

driskull (approved):
🪄 Awesome 🪄  🌈 🐰

src/components/stepper-item/stepper-item.scss
```diff
@@ -245,6 +303,28 @@
   @apply border-color-danger-hover;
 }
 
+:host([layout="horizontal"][complete]) .stepper-item-header {
+  border-color: rgba($h-bb-060, 0.5);
+}
+:host([layout="horizontal"][complete]:hover:not([disabled]):not([active])) .stepper-item-header,
```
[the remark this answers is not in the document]
jcfranco: I like this idea very much. Maybe stylelint's [selector-max-compound-selectors rule](https://stylelint.io/user-guide/rules/list/selector-max-compound-selectors/) + custom message can help?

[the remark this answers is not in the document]
driskull: Yeah that would be nice.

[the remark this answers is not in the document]
jcfranco: My original suggestion didn't work, but it looks like we could use [`selector-max-specificity`](https://stylelint.io/user-guide/rules/list/selector-max-specificity) instead. We'd just need to come up with a sensible value for the max specificity. I'll submit a PR for this.

Issue #4293, opened 2022-03-23
calcite-split-button: Cannot navigate dropdown with keyboard


jmanke:
### Summary

When opening the dropdown in the split-button, I am unable to navigate the list with arrow keys (up and down).

### Actual Behavior

The dropdown belonging to the split-button does not allow keyboard navigation through its items.

### Expected Behavior

The dropdown belonging to the split-button should allow keyboard navigation (up and down arrow keys) through its items.

### Reproduction Sample

https://codepen.io/jmanke/pen/XWVNOzP

### Reproduction Steps

1. Open split-by dropdown
2. Tab into the dropdown and try navigating through its items via the up/down arrow keys
3. Observe that navigation is not possible

### Reproduction Version

@esri/calcite-components@latest

driskull:
@jcfranco this one is a bit tricky, can we setup some time to go through our options?

The root if the issue is that there are nested default slots in play here and querying for assigned elements gets tricky.

driskull:
The root of the problem is that the slotted dropdown group is in light dom, but the dropdown itself is in shadow dom

they can't communicate easily

We would need a way to tell the dropdown where its items are and where to query

geospatialem:
Verified on next 👍🏻 

In addition to support for ⬆️ and ⬇️ keys also includes `home` and `end` keys.

PR #4368, opened 2022-04-04

driskull:
**Related Issue:** #4293

## Summary

### DIFF
--- a/src/components/dropdown/dropdown.tsx
+++ b/src/components/dropdown/dropdown.tsx
@@ -55,6 +55,11 @@ export class Dropdown implements InteractiveComponent {
   //
   //--------------------------------------------------------------------------
 
+  /**
+   * @internal
+   */
+  @Prop() hostEl: HTMLElement = null;
+
   /** Opens or closes the dropdown */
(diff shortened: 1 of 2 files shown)


driskull:
@jcfranco this seems like the only way to really get around this without having the user pass the dropdown.

IMO, I still think we should be keeping a parent/child component within the same root node so this seems like a hack.

As it stands, the dropdown is in shadowDOM and the dropdown items are in light DOM. Therefore, their relationship is kind of fractured. The host can't query the items without knowing where to query from.

driskull:
> Do you think we could introduce a set of utils (be it helper functions, interfaces, functional components or internal support components) to define a pattern where we can keep the dropdown/menu logic on the same host?

Are you suggesting that we have the same logic applied on the dropdown and split-button? I think that might be difficult to scale and is really internal to just the dropdown.

I think we should just enforce having the parent/child relationship within the same rootNode, be it a shadowRoot or document.

driskull:
We could try updating the `getSlotted()` utility to handle getting assigned elements for the default slot case. That might get this working instead of using querySelectorAll

PR #4369, opened 2022-04-04
ci(next): correct version number in build files preamble

benelan:
**Related Issue:** #4303

## Summary
In the PR linked above I added the version number to the preamble. To do so, I had to move the build step after the script that bumps the version. The next releases have a different build step which I didn't move, so the version is off by one:
https://unpkg.com/@esri/calcite-components@1.0.0-next.432/dist/calcite/calcite.esm.js

This PR ensures that the build happens after the version bump for next releases


jcfranco (approved):
👍

PR #4373, opened 2022-04-05

driskull:
**Related Issue:** #4293

## Summary

### DIFF
--- a/src/components/dropdown/dropdown.tsx
+++ b/src/components/dropdown/dropdown.tsx
@@ -368,14 +368,39 @@ export class Dropdown implements InteractiveComponent {
       : null;
   };
 
+  getGroups = (): HTMLCalciteTipGroupElement[] => {
+    const groups = getSlotted(this.el, {
+      all: true,
+      matches: "calcite-dropdown-group"
+    }) as any;
+
+    return groups;
+  };
+
+  getItems = (): HTMLCalciteDropdownItemElement[] => {
+    const groups = this.getGroups();
+
+    let items: any = [];
+
+    console.log(groups);
+
+    groups.forEach((group) => {
+      const groupItems = Array.from(group?.querySelectorAll("calcite-dropdown-item"));
+      console.log(groupItems);
+      items = items.concat(groupItems);
+    });
+
+    console.log(items);
+
+    return items;
+  };
+
   updateItems = (): void => {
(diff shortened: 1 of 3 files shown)


src/utils/dom.ts
```diff
@@ -162,15 +162,28 @@ export async function focusElement(el: FocusableElement): Promise<void> {
   return isCalciteFocusable(el) ? el.setFocus() : el.focus();
 }
 
+function getAssignedElements<T extends Element = Element>(element: Element, slotName: string | string[]): T[] {
+  const slotSelector = Array.isArray(slotName)
+    ? slotName.map((name) => `slot[name="${name}"]`).join(",")
```
driskull, line 167: I wonder if we should support multiple slot queries at once. We might be able to get rid of this support.

src/utils/dom.ts
```diff
@@ -162,15 +162,28 @@ export async function focusElement(el: FocusableElement): Promise<void> {
...
+  const elements = slots
+    .map((slot) => slot.assignedElements({ flatten: true }) as T[])
+    .reduce((previousValue, currentValue) => [...previousValue, ...currentValue], []);
+
+  return elements;
+}
+
 interface GetSlottedOptions {
   all?: boolean;
-  direct?: boolean;
```
driskull, line 167: Do we need direct? If we use assignedElements() then we can be sure that it is direct. Thats a benefit IMO.

driskull (commented):
This is more in line with what I was thinking @jcfranco but would still need some cleanup/refactor.

The general gist is that we use `slot.assignedElements()` instead of trying to use `querySelectorAll()` with a `slot` attribute.

src/components/dropdown/dropdown.tsx
```diff
@@ -194,7 +194,7 @@ export class Dropdown implements InteractiveComponent {
             ref={this.setScrollerEl}
           >
             <div hidden={!this.active}>
-              <slot />
+              <slot onSlotchange={this.updateItems} />
```
driskull, line 197: Some ideas:

Instead of calling `getSlotted` in `render` where most components do, we could just call it `onSlotChange`. Then we wouldn't have to query for the slot since we would have its reference from the event. It would just be getting its assignedElements and storing them on the instance for usage.

PR #4375, opened 2022-04-06
fix(split-button): Enable keyboard navigation

driskull:
**Related Issue:** #4293

## Summary

### Problem
The root of the problem is that the slotted dropdown group is in light dom, but the dropdown itself is in shadow dom. they can't communicate easily. Querying from the dropdown host won't always work because the items may be in light dom, not in the shadowRoot.

### Solution

Use `slot.assignedElements` to get the elements assigned to the slot instead of trying to query from the host. 

### Blocking issues

~https://github.com/ionic-team/stencil/issues/2830~ No longer an issue after recent changes


driskull:
@jcfranco this is working in e2e tests without playwright.

I got playwright up and running but i had to make modifications to get it to pass. Once I did that, it was working in regular e2e tests. So I'm guessing something else was just wrong 🤷

jcfranco (approved):
This looks great! 😎

Can you add a test before merging?

Issue #4040, opened 2022-02-02
calcite-input type="number": value is not delocalizing in some locales


jmanke:
### Actual Behavior

In locales such as 'et' and 'ar' the value on the calcite-input remains in the original locale, causing issues when parsing the value into a number.

### Expected Behavior

The value should be delocalized so that it can be parsed into a javascript number since this is the behavior for most other locales.

### Reproduction Sample

https://codepen.io/jmanke/pen/gOXPQZd?editors=1011

### Reproduction Steps

1. Use Estonian "et" locale.
2. Attach an event listener to the calcite-input to observe the input's value when changed.
2. Type a negative number with decimal places such as '-1,5'.
3. Click off of the input to trigger the change event.
4. Notice the target value is in its original locale. This prevents me from using `new Number()` or `parseFloat()` on the input's value.

### Reproduction Version

@esri/calcite-components@next

### Source

- [X] CDN
- [X] NPM package

benelan:
Closing this issue as verified based on ☝️ 
This is the issue I requested due to the negative sign not delocalizing. I will keep working on the original issue #3956 for the `ar` locale, which was resolved in #4020 and then reverted for #4063.

PR #4387, opened 2022-04-08
fix(input): correctly de-localize negative numbers in ar and et

benelan:
**Related Issue:** #4040 #3956

## Summary
The negative character is different in some locales, and it was not being delocalized, which caused the number to be invalid.


jcfranco (approved):
💥

PR #4403, opened 2022-04-13
fix(input): correctly delocalize numbers with group separators

benelan:
**Related Issue:** #3956

## Summary
When adding the locale spec tests in the PR above, the group separators were not working properly for a lot of locales (33 failed tests if I remember correctly). I fixed and cleaned up the locale utils, so now everything is passing.

### DIFF
--- a/src/utils/locale.spec.ts
+++ b/src/utils/locale.spec.ts
@@ -33,5 +33,11 @@ describe("localizeNumberString and delocalizeNumberString", () => {
         const delocalizedNumberString = delocalizeNumberString(localizedNumberString, locale);
         expect(delocalizedNumberString).toBe(numberString);
       });
+      it(`numbers with group separators localize and delocalize in "${locale}"`, () => {
+        const numberString = "12,345,678.9";
+        const localizedNumberString = localizeNumberString(numberString, locale, true);
+        const delocalizedNumberString = delocalizeNumberString(localizedNumberString, locale);
+        expect(delocalizedNumberString).toBe("12345678.9");
+      });
     });
 });
--- a/src/utils/locale.spec.ts
+++ b/src/utils/locale.spec.ts
@@ -33,6 +33,7 @@ describe("localizeNumberString and delocalizeNumberString", () => {
         const delocalizedNumberString = delocalizeNumberString(localizedNumberString, locale);
         expect(delocalizedNumberString).toBe(numberString);
       });
+
       it(`numbers with group separators localize and delocalize in "${locale}"`, () => {
         const numberString = "1,234";
         const localizedNumberString = localizeNumberString(numberString, locale, true);
--- a/src/utils/locale.spec.ts
+++ b/src/utils/locale.spec.ts
@@ -21,7 +21,7 @@ describe("localizeNumberString and delocalizeNumberString", () => {
       });
 
       it(`floating point numbers localize and delocalize in "${locale}"`, () => {
-        const numberString = "1.05";
+        const numberString = "4.321";
         const localizedNumberString = localizeNumberString(numberString, locale);
         const delocalizedNumberString = delocalizeNumberString(localizedNumberString, locale);
         expect(delocalizedNumberString).toBe(numberString);
@@ -34,6 +34,13 @@ describe("localizeNumberString and delocalizeNumberString", () => {
         expect(delocalizedNumberString).toBe(numberString);
       });
       it(`numbers with group separators localize and delocalize in "${locale}"`, () => {
+        const numberString = "1,234";
+        const localizedNumberString = localizeNumberString(numberString, locale, true);
+        const delocalizedNumberString = delocalizeNumberString(localizedNumberString, locale);
+        expect(delocalizedNumberString).toBe("1234");
+      });
+
+      it(`floating point numbers with group separators localize and delocalize in "${locale}"`, () => {
         const numberString = "12,345,678.9";
         const localizedNumberString = localizeNumberString(numberString, locale, true);
         const delocalizedNumberString = delocalizeNumberString(localizedNumberString, locale);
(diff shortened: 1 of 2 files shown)


src/utils/locale.ts
```diff
@@ -61,57 +62,44 @@ function createLocaleNumberFormatter(locale: string): Intl.NumberFormat {
...
-            return "";
-          }
-          return value;
-        })
-        .reduce((string, part) => string + part)
-        .replace(decimalSeparator, ".")
-        .replace(minusSign, "-")
-        .replace(/[^0-9\-\.]/g, ""); // remove everything except numbers, "-", and "."
+    const allDecimalsExceptLast = new RegExp(`[.](?=.*[.])`, "g");
+    const everythingExceptNumbersDecimalsAndMinusSigns = new RegExp("[^0-9-.]", "g");
```
jcfranco, line 66: Can you move this and the above regex outside of the function to be module level? They don't seem to be dynamic, so they can be reused instead of recreating on each call.

src/utils/locale.spec.ts
```diff
@@ -33,5 +33,18 @@ describe("localizeNumberString and delocalizeNumberString", () => {
         const delocalizedNumberString = delocalizeNumberString(localizedNumberString, locale);
         expect(delocalizedNumberString).toBe(numberString);
       });
+      it(`numbers with group separators localize and delocalize in "${locale}"`, () => {
```
jcfranco, line 36: Nitpick: can you add a newline here?

src/utils/locale.ts
```diff
@@ -61,57 +62,44 @@ function createLocaleNumberFormatter(locale: string): Intl.NumberFormat {
...
+  const parts = formatter.formatToParts(-9);
+  const value = parts.find((part) => part.type === "minusSign")?.value;
+  return value ? value.trim() : "";
 }
 
 export function localizeNumberString(numberString: string, locale: string, displayGroupSeparator = false): string {
   return sanitizeExponentialNumberString(numberString, (nonExpoNumString: string): string => {
     if (nonExpoNumString) {
-      const number = Number(sanitizeDecimalString(nonExpoNumString));
+      const number = Number(sanitizeDecimalString(nonExpoNumString.replace(/,/g, "")));
```
jcfranco, line 102: Can you also store this regex as a module-level variable?

src/utils/locale.ts
```diff
@@ -61,57 +62,44 @@ function createLocaleNumberFormatter(locale: string): Intl.NumberFormat {
...
 }
 
 export function getGroupSeparator(locale: string): string {
   const formatter = createLocaleNumberFormatter(locale);
-  const parts = formatter.formatToParts(1234567.8);
-  const value = parts.find((part) => part.type === "group").value;
-  return value.trim().length === 0 ? " " : value;
+  const parts = formatter.formatToParts(1234567);
+  const value = parts.find((part) => part.type === "group")?.value;
+  return value ? value.trim() : "";
```
jcfranco, line 82: Suggestion: `value?.trim() ?? ""`

src/utils/locale.ts
```diff
@@ -61,57 +62,44 @@ function createLocaleNumberFormatter(locale: string): Intl.NumberFormat {
...
 }
 
 export function getMinusSign(locale: string): string {
   const formatter = createLocaleNumberFormatter(locale);
-  const parts = formatter.formatToParts(-1234567.8);
-  const value = parts.find((part) => part.type === "minusSign").value;
-  return value.trim().length === 0 ? "-" : value;
+  const parts = formatter.formatToParts(-9);
+  const value = parts.find((part) => part.type === "minusSign")?.value;
```
jcfranco, line 95: Is it likely that the decimal and minus sign (maybe group too) parts won't be found? Seems like this wouldn't happen, but maybe I'm missing something.

benelan: I can't think of a time where the part couldn't be found with the way the functions are now. However, the three functions are a bit redundant, so I'm combining them to

```js
export function getNumberPart(type: string, locale: string): string {
  const formatter = createLocaleNumberFormatter(locale);
  const parts = formatter.formatToParts(-1234567.89);
  return parts.find((part) => part.type === type)?.value || "";
}
```
in which case it could be `undefined` if someone passes an invalid part type. Also, it actually shouldn't be trimmed because some locales group part is `" "`.

benelan: Sorry I just confused myself a bit. Here is an example of why I did that: https://codepen.io/benesri/pen/KKZxLEK?editors=0012

benelan: Okay, I figured out why the `return value.trim().length === 0 ? " " : value;` code existed. There are whitespace group characters that won't render correctly. So that is swapping the non-renderable white space with a renderable one. None of the locales have white space for the decimal, so that must have just been due to copy/paste. I added a comment so we don't forget again. cc @eriklharper

src/utils/locale.ts
```diff
@@ -61,57 +66,41 @@ function createLocaleNumberFormatter(locale: string): Intl.NumberFormat {
...
-    }
-    return nonExpoNumString;
+    return isValidNumber(delocalizedNumberString) ? delocalizedNumberString : nonExpoNumString;
   });
 }
 
 export function getGroupSeparator(locale: string): string {
   const formatter = createLocaleNumberFormatter(locale);
-  const parts = formatter.formatToParts(1234567.8);
+  const parts = formatter.formatToParts(1234567);
```
eriklharper, line 82: Is having the `.8` necessary for this calculation?

eriklharper (approved):

src/utils/locale.ts
benelan: No, which is why I removed it. That function only needs to find the group separator part. Now there will be less parts to iterate through in that function, FWIW.

eriklharper: Oh woops, my bad, I was reading that diff in the opposite way.  I guess that means I'm going color blind?!?! lol

PR #4412, opened 2022-04-14
fix(label): avoid accumulation of event listeners

timmorey:
**Related Issue:** # none

## Summary

When updating from `next.413` to `next.414` (see diffs https://github.com/Esri/calcite-components/compare/v1.0.0-next.413...v1.0.0-next.414), we observed a significant slowdown in the field-maps test suite.  This appears to be caused by an accumulation of event listeners on the `document` that never get properly removed.

To address this, I followed a similar pattern to how we were managing the `calciteInternalLabelClick` event handlers.  When adding a listener, we store it in a `WeakMap`, so that we'll have the exact listener reference on hand when it comes time to remove it.

We have validated these changes in field-maps, and it appears to restore the performance of our test suite without regression.

### DIFF
--- a/src/utils/label.ts
+++ b/src/utils/label.ts
@@ -37,6 +37,8 @@ export const labelDisconnectedEvent = "calciteInternaLabelDisconnected";
 
 const labelTagName = "calcite-label";
 const onLabelClickMap = new WeakMap<HTMLCalciteLabelElement, typeof onLabelClick>();
+const onLabelConnectedMap = new WeakMap<LabelableComponent, typeof onLabelConnected>();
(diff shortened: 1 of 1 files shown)


alexabreu:
Should `v1.0.0-beta.79` get labeled as a breaking change due to the memory leak?

jcfranco:
@timmorey Thanks for submitting this! I'll work on adopting this PR so it gets merged ASAP. @benelan We should probably have a release shortly after this lands since label usage is pretty common.

> Should v1.0.0-beta.79 get labeled as a breaking change due to the memory leak?

I don't think so. This is seems to be a bug and not an intentional change.

timmorey:
Closing this in favor of the adopted pr https://github.com/Esri/calcite-components/pull/4413

PR #4413, opened 2022-04-14
fix(label): avoid accumulation of event listeners

jcfranco:
# Adopted https://github.com/Esri/calcite-components/pull/4412

cc @timmorey 

--- 

**Related Issue:** # none

## Summary

When updating from `next.413` to `next.414` (see diffs https://github.com/Esri/calcite-components/compare/v1.0.0-next.413...v1.0.0-next.414), we observed a significant slowdown in the field-maps test suite.  This appears to be caused by an accumulation of event listeners on the `document` that never get properly removed.

To address this, I followed a similar pattern to how we were managing the `calciteInternalLabelClick` event handlers.  When adding a listener, we store it in a `WeakMap`, so that we'll have the exact listener reference on hand when it comes time to remove it.

We have validated these changes in field-maps, and it appears to restore the performance of our test suite without regression.

### DIFF
--- a/src/utils/label.ts
+++ b/src/utils/label.ts
@@ -37,6 +37,8 @@ export const labelDisconnectedEvent = "calciteInternaLabelDisconnected";
 
 const labelTagName = "calcite-label";
 const onLabelClickMap = new WeakMap<HTMLCalciteLabelElement, typeof onLabelClick>();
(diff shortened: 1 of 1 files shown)


driskull (approved):
👍 nice cleanup

PR #4423, opened 2022-04-19
feat(popover): Simplify popover usage by deprecating popover-manager component

driskull:
**Related Issue:** N/A

## Summary

feat(popover): Simplify popover usage by deprecating popover-manager component


driskull:
> To avoid the breaking change, could the popup manager set autoClose on child popups? 

@jcfranco it could, but it would need to setup a mutation observer to do it anytime the dom below it changes otherwise it might not be correct.

IMO I don't think its worth it but let me know.

src/components/popover-manager/popover-manager.tsx
```diff
@@ -36,4 +59,16 @@ export class PopoverManager {
...
     return <slot />;
   }
+
+  // --------------------------------------------------------------------------
+  //
+  //  Private Methods
+  //
+  // --------------------------------------------------------------------------
+
+  setAutoClose = (): void => {
```
jcfranco, line 69: Nitpick: This can be a class method instead of an assigned bound function.

jcfranco:
> IMO I don't think its worth it but let me know.

Avoiding a breaking change is totally worth it, IMO.

jcfranco (approved):
This is awesome stuff, @driskull! 

Approving per our convo (to use a mutation observer). Can you also add a test before merging?

im-so-freakin-excited-shouting

Issue #3961, opened 2022-01-19
Enter key event should submit form


COV-GIS:
### Description

Enter key event should submit form if any calcite input like component therein is focused.

I noticed this behavior on the desktop which is annoying but easy enough to get around by tabbing to submit button or using the mouse. But this is a problem on mobile devices....

1. The mobile keyboard `Go` button does nothing. The user has to close the keyboard and click the submit button.
2. The keyboard closes but the form doesn't submit. This behavior is the biggest problem as the user thinks they are doing something but nothing is actually happening.

### Acceptance Criteria

Enter key event should submit form if any calcite input like component therein is focused.

### Relevant Info

https://codepen.io/COV-GIS/pen/wvrxQpP

### Which Component

All input components.

gpbmike:
Also of note, I'm still on beta.46 and have a hack to get default form behavior (submit on enter/return) working. The hack involves adding a hidden `<button type="submit">` to the form. As I evaluate beta.77, I see that the default form behavior still does not work (this issue) and that my hack also no longer works. I'm guessing because the input the user interacts with was moved to the shadow DOM.

noahmulfinger:
I also ran into this while updating to beta.79. Definitely a pretty difference in behavior from standard inputs, likely due to the shadow DOM.

driskull:
@jcfranco looks like we need to enhance all our "input" components to call `formEl?.requestSubmit();` when the enter key is pressed. Does that sound correct?

We could probably do this via the connectForm utility, right?

jcfranco:
@driskull I believe so. I think the util would also need to provide a way for implementors to prevent submitting on Enter key presses meant for another workflow (e.g., combobox accepting a custom value).

driskull:
How about in that case...

We have a `onFormEnter` function they could override and do their own logic to submit the form when necessary?

Maybe `onFormEnter` could actually just return a boolean value to determine if a submit should happen or not?

PR #4427, opened 2022-04-20
fix: Enter key event should submit form from form elements. (#3961)

driskull:
**Related Issue:** #3961

## Summary

fix: Enter key event should submit form from form elements. (#3961)

### DIFF
--- a/src/utils/form.tsx
+++ b/src/utils/form.tsx
@@ -56,6 +56,11 @@ export interface FormComponent<T = any> {
   defaultValue: T;
 
   /**
+   * Hook for components to provide custom form enter key behavior.
+   */
+  onFormEnter?(): void;
+
+  /**
    * Hook for components to provide custom form reset behavior.
    */
   onFormReset?(): void;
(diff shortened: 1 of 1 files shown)


driskull:
Opened this PR to get some eyes on it. 👀 

We would need to do some exclusionary stuff for components that have more than 1 element inside of them that would trigger an enter key and we wouldn't want it to submit the form.

WDYT @jcfranco ?

jcfranco:
This is looking good! Do you have an example of how a component would implement `onEnterShouldSubmit`?

driskull:
> This is looking good! Do you have an example of how a component would implement onEnterShouldSubmit?

Not yet. I'm still on the fence about this. I wonder if it might be easier if a component just called `formEl?.requestSubmit()` when it needed to?

The only benefit of having to opt out is that we enforce the enter key functionality.

jcfranco:
Adding to your idea, how about adding a method to the interface for submitting? Another benefit to it is that it keeps all form-related behavior/details in the util. I like the idea of having the util take care of it, but managing complex components will be tricky.

Issue #4350, opened 2022-03-31
calcite-inline-editable keeps emitting events after editing was cancelled or confirmed


ffaubry:
### Actual Behavior

The calcite-inline-editable element will emit events after the editing has been cancelled or confirmed.
The event is emitted on mouse hovering on the edit button.

![image](https://user-images.githubusercontent.com/3506166/161120844-eba4ca10-2100-4db2-a814-3448c480e325.png)

### Expected Behavior

After editing the events should not be emitted.

### Reproduction Sample

https://codepen.io/ffaubry/pen/qBpPRvJ?editors=1001

### Reproduction Steps

1. Open the console view
2. Click on the edit button
3. Click in the cancel button
4. Hover on the edit button
5. You should see events being fired

### Reproduction Version

beta.80

jcfranco:
[For implementor]: https://github.com/Esri/calcite-components/blob/083d283caf719eed91cb1f5bd71b4584864b5e61/src/components/inline-editable/inline-editable.tsx#L278 may be firing because of bubbling transition events.

geospatialem:
verified on next ✔️

PR #4432, opened 2022-04-21
fix(inline-editable): will no longer emit unnecessary calciteInlineEditableEditCancel events

Elijbet:
**Related Issue:** #4350

## Summary

The event emitter fires on transitionEnd (on the condition that editingEnabled = false), which is on the wrapper. The only thing I see transitioning on the wrapper is the background, and this happens on hover/mouseover. So the moment you set editingEnabled to false by clicking on the x button, with every mouse over the wrapper, the background transition also triggers emit along with it. 

Buttons show conditionally on shouldShowControls (this is set to true when clicking on the wrapper). There seems to be no transition/animation that needs to be watched when canceling. If that’s the case, then the simple solution is to move the if condition for emitting the event into the cancelEditing function, so calciteInlineEditableEditCancel is called whenever this function is called, which is when pressing the x button and ESC key.


src/components/inline-editable/inline-editable.e2e.ts
```diff
@@ -233,13 +233,37 @@ describe("calcite-inline-editable", () => {
...
+
+      await element.click();
+      const cancelEditingButton = await page.find(`calcite-inline-editable >>> .${CSS.cancelEditingButton}`);
+
+      await input.type("one");
+      await cancelEditingButton.click();
+      await cancelEvent;
+      expect(cancelEvent).toHaveReceivedEventTimes(1);
+
+      // should not emit on hover after editing is cancelled
```
jcfranco, line 255: Could you add a link to the issue here for reference?

jcfranco (approved):
Nice! 🏆

Issue #2426, opened 2021-07-01
A11y: Card Component Checkbox Should Have a Visible Label


r-watkins:
### Summary
When the card component is selectable the checkbox does not have a visible label.

![Screen Shot 2021-06-30 at 9 50 23 AM](https://user-images.githubusercontent.com/17993027/124170505-5aca8400-da5c-11eb-9e30-a82105299136.png)

### Actual Behavior

There is a Select/Deselect label that appears which I think serves screen readers and will provide context for most sighted users.

![Screen Shot 2021-07-01 at 11 07 26 AM](https://user-images.githubusercontent.com/17993027/124170850-bbf25780-da5c-11eb-9acc-70671e3706ca.png)

### Expected Behavior

It's my understanding that checkboxes should have a visible label that can be clicked to check/uncheck the checkbox to increase the clickable area and make it easier for those with a motor control impairment to perform the selection action. 

This is my typical card design which uses the card title as the label:

![Screen Shot 2021-07-01 at 11 01 10 AM](https://user-images.githubusercontent.com/17993027/124172704-fd840200-da5e-11eb-9c64-684a12d82d9e.png)

I wonder though if we could just make the entire card a clickable area to check/uncheck the checkbox and fulfill the same criteria of an increased selection area?

### Relevant Info 
I am referencing the card design here with the selectable option:
https://esri.github.io/calcite-components/?path=/story/components-card--simple

### Working W3C Example/Tutorial 
I am referencing this W3.org spec:
https://www.w3.org/WAI/WCAG21/Techniques/html/H44

julio8a:
We just had an accessibility audit from a third-party company. Will check with them on the a11y requirements for something like this.

macandcheese:
It seems like the "Actual Behavior" referenced in the OP no longer works - we should ensure that functionality is working.

driskull:
@benelan @jcfranco I think this should go through design to position the label alongside the checkbox.

We don't want to just implement it without consulting the design team and updating figma/etc.

jcfranco:
We did triage this together, but I'll defer to @macandcheese.

macandcheese:
We decided during triage not to address the add a visible label portion of the request. However the “actual behavior” in the OP doesn’t work - we should at least have a native label show up as it used to.

eriklharper:
> I wonder though if we could just make the entire card a clickable area to check/uncheck the checkbox and fulfill the same criteria of an increased selection area?

[...]

PR #4436, opened 2022-04-21
fix(card): Checkbox label should be visible and accessible

driskull:
**Related Issue:** #2426

## Summary

fix(card): Checkbox label should be visible and accessible. (#2426)

- Moves checkbox next to the title and surrounds it with a label for accessibility.
- Current use of "Select"/"Deselect" doesn't meet a11y guidelines as the labels are not useful.

image

image

### DIFF
--- a/src/components/card/card.e2e.ts
+++ b/src/components/card/card.e2e.ts
@@ -69,9 +69,9 @@ describe("calcite-card", () => {
       </calcite-card>
     `);
 
-    const thumbContainer = await page.find(`calcite-card >>> .${CSS.checkboxWrapper}`);
+    const checkbox = await page.find(`calcite-card >>> calcite-checkbox`);
 
-    expect(thumbContainer).not.toBeNull();
+    expect(checkbox).not.toBeNull();
   });
 
   describe("when a card is selectable", () => {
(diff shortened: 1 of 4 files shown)


jcfranco (approved):
Code LGTM! @macandcheese Can you review the visuals?

If missing, can you add stories that represent the screenshots from the description? This would improve our coverage.

macandcheese:
Some questions / thoughts... sorry in advance for the rambling. 

If a user were to slot a title as a link - how would that work with the current construction. Clicking the title (as rendered with label) is nice when the title is not a link, but I could see that being a weird interaction if it were to quickly select before navigating to the intended link.

The calcite-label should not have margin-bottom - the spacing between the "selectable" title slot and description should be the same if the card is selectable or not, IMO.

The card example with "two buttons" should not show a check / "x" - might be confusing as an example - just square placeholder icons could be fine there.

How does this look with icon in subtitle? I know Online and other implementations often have a layer type icon under the title. Having an icon under this checkbox could be a little weird.

Not specific to this PR - but using `<h3>` for the title elements in samples - is that expected?

For voiceover, is using the title sufficient? Do we still need to provide a hidden "select / deselect" string before the title?


This issue is in progress - let's make sure to reconcile differences between this PR and that proposed design: https://github.com/Esri/calcite-components/issues/3181


cc @asangma @eriklharper for some 👀 .

driskull:
Yeah those a good questions. I can close this PR until the design is figured out.

The problem I see with the current approach is that the checkbox doesn't have a useful label.

We could add a separate label property for the checkbox if necessary.

PR #4439, opened 2022-04-22
fix(input, combobox, input-date-picker, input-time-picker): Enter key event should submit form

driskull:
**Related Issue:** #4439

## Summary

fix(input, combobox, input-date-picker, input-time-picker): Enter key event should submit form. (#4439)


driskull:
@jcfranco how about dis?

src/utils/form.tsx
```diff
@@ -126,6 +126,20 @@ function hasRegisteredFormComponentParent(
   return hasRegisteredFormComponentParent;
 }
 
+/**
+ * Helper to submit a form.
+ */
+export function submitForm(formEl: HTMLFormElement): void {
```
jcfranco, line 132: How about passing the form component and we can grab the `formEl` from there? It'd be consistent with existing utils here (e.g., `connectForm`).

jcfranco:
yesss

src/utils/form.tsx
driskull: The problem with that is it won't work for the button case since the button isn't a FormElement, it just submits it :\

jcfranco: Gotcha. Let's leave button out of this then since it's already doing its own thing and the original issue is about form components.

driskull: I think its better now 🎉

jcfranco: Oooh, yusss. 🤩

Last feedback on the draft is maybe renaming `FormSubmitComponent` to `FormOwner` and adding some internal doc on it for guidance.

driskull:
So if a parent component uses a `input` internally, should there be a way to turn off the form submitting on enter?

driskull:
@jcfranco should we be doing this?

```tsx
if (event.key === "Enter") {
      this.calciteInputChange.emit();

      if (!event.defaultPrevented) {
        submitForm(this);
      }
    }
```

src/components/combobox/combobox.tsx
```diff
@@ -261,6 +262,13 @@ export class Combobox implements LabelableComponent, FormComponent, InteractiveC
    */
   @Event() calciteComboboxClose: EventEmitter;
 
+  @Listen("keydown", { target: "body" })
```
driskull, line 265: @jcfranco is this the proper place we should be listening?

src/components/filter/filter.tsx
```diff
@@ -179,10 +179,14 @@ export class Filter implements InteractiveComponent {
     this.filter(target.value, true);
   };
 
-  keyDownHandler = ({ key }: KeyboardEvent): void => {
-    if (key === "Escape") {
+  keyDownHandler = (event: KeyboardEvent): void => {
+    if (event.key === "Escape") {
       this.clear();
```
driskull, line 184: todo: this should be preventable in separate issue/PR

src/components/button/button.tsx
```diff
@@ -266,17 +267,17 @@ export class Button implements LabelableComponent, InteractiveComponent {
...
   private handleClick = (): void => {
-    const { formEl, type } = this;
+    const { type } = this;
 
     if (this.href) {
       return;
     }
 
     // this.type refers to type attribute, not child element type
     if (type === "submit") {
```
driskull, line 277: @jcfranco should we support preventingDefault on click events here or do you think its ok if a user just sets the button type to "button" in those cases?

src/components/color-picker-hex-input/color-picker-hex-input.tsx
```diff
@@ -288,4 +285,10 @@ export class ColorPickerHexInput {
   private nudgeRGBChannels(color: Color, amount: number): Color {
     return Color.rgb(color.array().map((channel) => channel + amount));
   }
+
+  handleKeyDown(event: KeyboardEvent): void {
```
driskull, line 289: This fixes the following from submitting a form.

```html
<form>
        <calcite-color-picker scale="s"></calcite-color-picker>
      </form>
```

src/components/color-picker/color-picker.tsx
```diff
@@ -993,6 +994,12 @@ export class ColorPicker implements InteractiveComponent {
   //
   //--------------------------------------------------------------------------
 
+  handleKeyDown(event: KeyboardEvent): void {
```
driskull, line 997: This fixes the following from submitting a form.

```html
<form>
        <calcite-color-picker scale="s"></calcite-color-picker>
      </form>
```

src/components/filter/filter.tsx
```diff
@@ -179,10 +179,14 @@ export class Filter implements InteractiveComponent {
...
     }
+
+    if (event.key === "Enter") {
```
driskull, line 187: Prevents the following from submitting a form:

```html
<form>
        <calcite-value-list id="one" multiple filter-enabled>
          <calcite-action slot="menu-actions" indicator text="Cool" scale="s" icon="hamburger"> </calcite-action>
          <calcite-value-list-item label="START ACTIONS NESTED SLOT" description="Man's best friend" value="dogs">
            <calcite-action slot="actions-start" text="click-me" onClick="console.log('clicked');" icon="banana">
            </calcite-action>
            <calcite-action slot="actions-start" text="click-me" onClick="console.log('clicked');" icon="banana">
            </calcite-action>
            <calcite-action slot="actions-end" text="click-me" onClick="console.log('clicked');" icon="sliders">
            </calcite-action>
          </calcite-value-list-item>
          <calcite-value-list-item label="Cats" description="Independent and fluffy" value="cats">
            <calcite-action label="action start" slot="actions-start" scale="m">
              <div style="align-self: center; width: 16px; height: 16px; background-color: #56ccf2"></div>
            </calcite-action>
          </calcite-value-list-item>
          <calcite-value-list-item
            label="Fish. But not just any fish, a tiger fish caught live in the Atlantic Ocean while on vacation."
            description="Easy to care for." value="fish">
            <calcite-action label="action start" slot="actions-start" scale="m">
              <div style="align-self: center; width: 16px; height: 16px; background-color: #990000"></div>
            </calcite-action>
          </calcite-value-list-item>
        </calcite-value-list>
      </form>
```

src/components/button/button.tsx
jcfranco: Your suggestion makes sense. I don't think we need to cancel click events.

driskull:
@jcfranco do you think you could update the `formAssociated` test to check for submitting via an enter key? It would just be another optional test we can do for those components that need that support. (input, combobox, input-date-picker, input-time-picker)

driskull:
Yep thats fine 👍

src/tests/commonTests.ts
```diff
@@ -596,6 +603,32 @@ export async function formAssociated(componentTagOrHtml: TagOrHTML, options: For
...
+    });
+
+    const stringifiedTestValue = stringifyTestValue(options.testValue);
+
+    await component.setProperty("value", stringifiedTestValue);
+    await component.callMethod("setFocus");
+    await page.keyboard.press("Enter");
+    const called = await page.evaluate(() => (window as TestWindow).called);
+
+    expect(called).toBe(true);
```
driskull, line 626: awesome 👏

driskull (commented):
wooot!!! Looks good

src/components/filter/filter.tsx
jcfranco: Is this suggestion to be able to prevent the filter from being cleared when the event is canceled?

src/demos/_assets/demo-form.ts
```diff
@@ -20,7 +20,6 @@ class DemoForm extends HTMLElement {
       data[pair[0]] = pair[1];
     }
     console.log(data);
-    alert(JSON.stringify(data, null, 2));
```
jcfranco, line 23: ✨:broom:✨

jcfranco (approved):
LGTM!

@driskull Can you create a refactor issue to revisit the form util to only have the topmost form owner handle all the form logic?

@benelan This adds more content for our form component pattern (internal dev doc).

src/components/filter/filter.tsx
driskull: Yes, correct.

driskull: We could add the event listener on the document using the `@Event` decorator and check to see if it has `event.defaultPrevented` in this case as a solution to keep the logic on this component.

driskull: https://github.com/Esri/calcite-components/issues/4507

Issue #4425, opened 2022-04-20
Refactor: consolidate filter logic


benelan:
### Description

Right now `calcite-combobox` uses the [`filter.ts` util](https://github.com/Esri/calcite-components/blob/master/src/utils/filter.ts). We also have a `calcite-filter` component which has [almost identical filtering logic](https://github.com/Esri/calcite-components/blob/master/src/components/filter/filter.tsx#L143-L174).

If possible, we should use `calcite-filter` in `calcite-combobox`. If it is not possible (maybe due to the chips?), then we should refactor `calcite-filter` to use the `filter.ts` util.

### Proposed Advantages

Consolidate duplicate code that may need to be changed in multiple places for enhancements or bug fixes.

### Which Component

`calcite-filter.tsx` and `filter.ts`

benelan:
fixed by #5012 - thanks @jcfranco  :trophy:

Issue #4348, opened 2022-03-31
Combobox is not filtering certain values


tresh11178:
### Actual Behavior

When inputting certain values, the Combobox dropdown does not properly filter the options as expected.

### Expected Behavior

The expectation is that all values should be filtered properly based on the input from the user.

### Reproduction Sample

https://codepen.io/tre11178/pen/KKZXMYW?editors=100

### Reproduction Steps

1. Click into the Combobox
2. Type "F" into the input.
3. Notice that the dropdown is not filtering all options other than "Fire Safety Team".
4. Type "Fi" into the input.
5. Notice the dropdown is still not filtering.'
6. Type "Fir" into the input.
7. Notice that the dropdown properly filters.

### Reproduction Version

beta.80

benelan:
[Implementer Notes]: This is probably because the `forIn` function "Iterates over own and inherited enumerable properties of an object invoking iteratee for each property."

All of the combobox-items are matching the word "define"

geospatialem:
@Elijbet I'm seeing some filtering issues with a few letters on `next`. Put together an [updated codepen](https://codepen.io/geospatialem/pen/oNpOEwa) for reference. Can you take a peek 👀? 

*Note: Added an additional item to showcase filtering further, but tested across the alphabet.*

Some of the filtering experienced:
- d, For results must type "di"
- e, For results must type "edi"
- f, For results must type "fir"
- n, For results must type "na"
- u, For results must type "ur"

![filter results from 'd'](https://user-images.githubusercontent.com/5023024/164799587-c50d8e18-8aea-4054-ae90-64f9375d22ef.png)

![filter results from 'di'](https://user-images.githubusercontent.com/5023024/164799613-19dae15d-675c-4869-8cb3-2f8adc08b323.png)

Elijbet:
Sure, I'll check now.

benelan:
Yeah that's what I thought. Its not just the guid, its the inherited properties too. The first solution worked IIRC.

benelan:
Or if I'm wrong and the `const inputSubset = (({ label, value }) => ({ label, value }))(input as any);` didn't work, I'd look into [`Object.hasOwn`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn).

Elijbet:
> Yeah that's what I thought. Its not just the guid, its the inherited properties too. The first solution worked IIRC.

[...]

PR #4443, opened 2022-04-23
fix(combobox, filter): only test regex on string properties when filtering

benelan:
**Related Issue:** #4348

## Summary
It was matching the `constant` attribute's `undefined` default value. 
```js
/u/.test(undefined) === true
```
Fixing it this way also works for `calcite-filter`, rather than only filtering based on `label` and `value` properties (relevant for #4425)

### DIFF
--- a/src/components/filter/filter.tsx
+++ b/src/components/filter/filter.tsx
@@ -151,14 +151,14 @@ export class Filter implements InteractiveComponent {
     const find = (input: object, RE: RegExp): any => {
       let found = false;
       forIn(input, (val) => {
-        if (typeof val === "function") {
+        if (typeof val === "function" || val == null /* intentional == to catch undefined */) {
           return;
         }
         if (Array.isArray(val) || (typeof val === "object" && val !== null)) {
           if (find(val, RE)) {
             found = true;
           }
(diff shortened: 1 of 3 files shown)


src/components/filter/filter.tsx
```diff
@@ -158,7 +158,7 @@ export class Filter implements InteractiveComponent {
           if (find(val, RE)) {
             found = true;
           }
-        } else if (RE.test(val)) {
+        } else if (typeof val === "string" && RE.test(val)) {
           found = true;
```
benelan, line 162: @jcfranco should this also accept `number` typed values? Maybe if there is a rating property or something that people want to filter by

jcfranco: The current version changes the filtering behavior, right? I think pick/value list items allow passing metadata objects that go through filtering.

How about returning if the value is `null` or `undefined` instead?

```ts
if (typeof val === "function" || val == null /* intentional == to catch undefined */) {
  return;
}

if (Array.isArray(val) || typeof val === "object") { /* ... */ }
{ else if (RE.test(val)) { /* ... */ }
```

Elijbet (approved):

Issue #4364, opened 2022-04-02
The time picker of the input time picker appears in different positions


fendiaoxiaoshuazi:
### Actual Behavior

The time picker of the input time picker appears in different positions when the value gets cleared.

### Expected Behavior

The time picker should appear on the same position.

### Reproduction Sample
https://codepen.io/nan9266/pen/KKZymWx

### Reproduction Steps

1. Click on the input time picker, observe the position of the time picker
2. Clear the value of the input time picker, observe the position of the time picker changes
3. Click outside, and click the input time picker again, observe the position of the time picker changes

### Reproduction Version

beta.80

benelan:
Hi @fendiaoxiaoshuazi, can you please provide a sample that reproduces the issue?

fendiaoxiaoshuazi:
Hi @benelan , the sample and screen shot has been included in the description.

jcfranco:
[Implementor's note]: The input time picker needs a `reposition` method (similar to [other](https://github.com/Esri/calcite-components/blob/master/src/components/tooltip/tooltip.tsx#L131) [popper](https://github.com/Esri/calcite-components/blob/master/src/components/input-date-picker/input-date-picker.tsx#L293)-[owner](https://github.com/Esri/calcite-components/blob/master/src/components/dropdown/dropdown.tsx#L213) [components](https://github.com/Esri/calcite-components/blob/master/src/components/combobox/combobox.tsx#L205)).

jcfranco:
@fendiaoxiaoshuazi Here's one workaround in the meantime: https://codepen.io/jcfranco/pen/jOYxxNQ?editors=1010

benelan:
verified on `next`

PR #4448, opened 2022-04-25
feat(input-time-picker): Add reposition method

driskull:
**Related Issue:** #4364

## Summary

feat(input-time-picker): Add reposition method. (#4364)


jcfranco (approved):
👍

jcfranco:
I noticed that none of the popper owners have tests for `reposition`. Can you open a test issue to set one up? Maybe we can enhance the `popperOwner` helper to handle this.

driskull:
> I noticed that none of the popper owners have tests for reposition. Can you open a test issue to set one up? Maybe we can enhance the popperOwner helper to handle this.

Maybe we can just setup a "PopperOwner" class for them to inherit to make sure they have one?

Issue #4306, opened 2022-03-25
It is not possible to override the Icon color in the Chip component


BennyBaruba:
### Actual Behavior

When overriding the Icon color using `--calcite-ui-icon-color` for the Chip component, the icon color does not change.  However the text color changes.

### Expected Behavior

 Using `--calcite-ui-icon-color` to override the Icon color for the Chip component, should change the color for the icon and not the color of the text in the Chip.  The workflow should be the same as seen in the Button or Tab component.

### Reproduction Sample

https://codepen.io/NebHtesmar/pen/vYpgWVZ

### Reproduction Steps

1. Add a Chip component to an application
2. Override the Icon color by using `--calcite-ui-icon-color`
3. Notice the color of the Icon does not change, however the text does.

### Reproduction Version

https://codepen.io/NebHtesmar/pen/vYpgWVZ

BennyBaruba:
@macandcheese

geospatialem:
Verified in next

![icon color change in calcite-chip with CSS verified](https://user-images.githubusercontent.com/5023024/167164861-c4bbbc3d-d7b7-4143-9d73-7a939810604d.png)

PR #4450, opened 2022-04-26
fix(chip): apply --calcite-ui-icon-color to chip icon if defined

jcfranco:
**Related Issue:** #4306 

## Summary

This updates chip to apply `--calcite-ui-icon-color` to the chip icon if defined. Otherwise, it will apply the icon color according to the `color`-prop.

Note: only the chip icon will be be affected and not the close icon. 

Screen Shot 2022-04-25 at 5 52 24 PM

Also, the image container CSS class was renamed to follow changes to the icon one. I can revert this if deemed better as a separate refactor PR.

### DIFF
--- a/src/components/chip/chip.scss
+++ b/src/components/chip/chip.scss
@@ -4,7 +4,7 @@
   @apply text-n2 h-6;
   --calcite-chip-spacing-unit-l: theme("spacing.2");
   --calcite-chip-spacing-unit-s: theme("spacing.1");
-  .chip-image-container {
+  .image-container {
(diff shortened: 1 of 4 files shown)


driskull (approved):
Looks good!

> Note: only the chip icon will be be affected and not the close icon.

If users request, we could introduce a chip specific variable to handle the close icon color.

jcfranco:
@macandcheese Could you review the screenshot diffs? https://screener.io/v2/states/1leJJn.Esri-calcite-components/refs%2Fpull%2F4450%2Fmerge

PR #4453, opened 2022-04-26
ci: add current milestone when pr merged

benelan:
**Related Issue:** NA

## Summary
Right now the current milestone is being added when the PR is opened. We need everything in a past-due milestone to be closed before we can close the milestone. Stale PRs are making it harder to close milestones. So it will be better to add PRs to milestones when they are merged, rather than opened.


jcfranco (approved):
🎉 🌮 🦖 🦃

jcfranco:
Worth noting that until it can be automatically handled, PRs without associated issues will have to be independently tracked as they will no longer be tied to a milestone.

PR #4454, opened 2022-04-26
chore: modify test template

geospatialem:
**Related Issue:** N/A

## Summary

Modify the unstable test template to fit all test cases including new, unstable tests, or Screener. 

- Renamed the template to "test.yml" for its new use
- All original content remains, but some fields are now optional
- An additional section was added to the top of the template indicating the type of test being requested


driskull (approved):
👍 💯

Issue #4376, opened 2022-04-06
Bug: Filter clear focus ring is black


benelan:
### Actual Behavior

The focus ring for filter's clear button is black and rounded.
![image](https://user-images.githubusercontent.com/10986395/162091520-69f6da1d-658a-449e-96ed-9390e9710020.png)

### Expected Behavior

Discussed with @macandcheese, and we both think that filter should just be a clearable input.
![image](https://user-images.githubusercontent.com/10986395/162091717-49e2fb3b-7799-4c36-bcf6-bbc8d90a74c6.png)

```html
<calcite-input clearable></calcite-input>
```

### Reproduction Sample

https://codepen.io/benesri/pen/VwyXPaa

### Reproduction Steps

1. Go to the sample and type something into the filter
2. Press tab

### Reproduction Version

beta.80

geospatialem:
Verified on next 👍🏻 

![image](https://user-images.githubusercontent.com/5023024/166320397-43fd619d-52f9-4836-8c41-c1bf1256b66c.png)

PR #4456, opened 2022-04-27
fix(filter): fix inconsistent text-clearing UX

jcfranco:
**Related Issue:** #4376 

## Summary

This fixes a UI inconsistency with `calcite-filter` by refactoring it to use `calcite-input`'s `clearable` option. 

### Noteworthy changes to `calcite-input`

* Added `intlClear` prop
* Added default aria-label text for the clear button in case ☝️ is empty cc @geospatialem


driskull (approved):
💯 👍

Issue #4229, opened 2022-03-12
Event dispatched for programmatic change if input has focus


nwhittaker:
### Actual Behavior

If a `calcite-input` element has focus, and its value is changed programmatically, it will dispatch a `calciteInputChange` event when it's blurred.

### Expected Behavior

A `calciteInputChange` event is not dispatched because the value was not changed due to a user interaction.

### Reproduction Sample

https://ember-twiddle.com/5f817823e90d22b8417ae66304a80c52

Ember sample forked from `calcite-components-examples` https://github.com/eriklharper/calcite-4229-repro

### Reproduction Steps

1. Visit the sample and open the browser's dev tools console
2. Click the "Seed value with focus shift" button
3. After focus shifts to the input, blur the input in some way
4. Observe that `"didChangeSeededInput"` appears in the console indicating that the input's `calciteInputChange` handler was hit

### Reproduction Version

1.0.0-next.415

### Relevant Info

Related/parent issue: https://github.com/Esri/calcite-components/issues/2033.

The example is somewhat contrived, but Ember has a data-down, action-up pattern that makes it easy to craft a scenario where focus can shift to an input before a previous action has finished updating the input's value.

jcfranco:
@eriklharper Do you think you can take a look at this? This block looks suspect: https://github.com/Esri/calcite-components/blob/ffff3d5933d3a00776adc90d78d2cb449ba7920a/src/components/input/input.tsx#L449-L451

eriklharper:
@nwhittaker since this is a contrived example and is discouraged in Ember's own documentation, I'm curious if you are able to workaround this by using a modifier?

I tried something like this which works to resolve this for me locally in our Ember example app:

First install `ember-modifier` with:

`ember install ember-modifier` or `npm i --save-dev ember-modifier`

app/modifiers/focus-calcite-input.js
```
import { modifier } from 'ember-modifier';

export default modifier((element) => element.setFocus());
```

app/controllers/application.js
```
import Controller from '@ember/controller';
import { action } from '@ember/object';
import { tracked } from '@glimmer/tracking';

export default class ApplicationController extends Controller {
  @tracked seededValue

  @action seedValueWithFocusShift() {
    this.seededValue = Math.random();
  }

  @action didChangeSeededInput() {
    console.log('didChangeSeededInput')
  }
}
```

app/templates/application.hbs
```
<p>
  <calcite-button {{on 'click' this.seedValueWithFocusShift}}>
[...]

PR #4463, opened 2022-04-28
fix(input): stop emitting change event on blur when the value is set directly

eriklharper:
**Related Issue:** #4229

## Summary

This fixes an issue where the `calciteInputChange` event is fired when a focused input is blurred after its value is set directly.  This fix also keeps the input from firing change events when hitting the enter key repeatedly or blurring the input repeatedly when the value didn't actually change.


src/components/input/input.tsx
```diff
@@ -697,15 +706,17 @@ export class Input implements LabelableComponent, FormComponent, InteractiveComp
         ? localizeNumberString(newValue, this.locale, this.groupSeparator)
         : "";
 
-    this.internalValueChange = origin === "internal" && this.value !== newValue;
-    this.setPreviousValue(this.value);
+    this.internalValueChange = origin === "user" && this.value !== newValue;
```
jcfranco, line 709: Same as above, please use an if statement for readability.

src/components/input/input.tsx
```diff
@@ -437,6 +441,13 @@ export class Input implements LabelableComponent, FormComponent, InteractiveComp
     });
   };
 
+  private emitChangeEvent = (): void => {
```
jcfranco, line 444: Is there an alternative name we could use for this? It emits conditionally, so it could be misleading.

jcfranco (approved):
Other than some minor comments, this LGTM! 🎉

src/components/input/input.tsx
```diff
@@ -32,6 +32,7 @@ import { createObserver } from "../../utils/observers";
 import { InteractiveComponent, updateHostInteraction } from "../../utils/interactive";
 
 type NumberNudgeDirection = "up" | "down";
+type SetValueOrigin = "initial" | "loading" | "user" | "direct";
```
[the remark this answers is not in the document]
eriklharper: Yes, exactly!  This was my initial stab at naming these origins.  I like `direct` and that is echoed in places like MDN for direct value setting.  Either `user` or `user-interaction` works, but I could even see going so far as making user interaction sources more specific to use words like `keyboard` or `mouse` even, depending on the component.  

Normal typing would be handled by the `keyboard` origin, and I can imagine a use case like the number input where a user clicks the up/down arrows with the mouse and differentiating between `mouse` interaction that way.

src/components/input/input.tsx
eriklharper: Yeah, name for this one was really hard :/  I wanted to put it into a function that can be called everywhere a change event needed to be emitted so that I could include the conditionals all in one place as opposed to spreading that out in all the calling locations.  I'll think on it some more, but that was the best I could come up with in the moment.

src/components/input/input.tsx
jcfranco: Apologies, I misread this line. You can ignore my comment here.

src/components/input/input.tsx
eriklharper: `emitChangeEventIfNecessary()` but seems kinda long.

eriklharper: `handleChangeEvent()`?

src/components/input/input.tsx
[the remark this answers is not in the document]
driskull: Yeah it would be great if these could be cleaned up into one private property and standardize the language.

It seems like these could be grouped..

```ts
  private previousEmittedValue: string;

  private previousValue: string;

  private previousValueOrigin: SetValueOrigin = "initial";

  private userChangedValue = false;
```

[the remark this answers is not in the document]
eriklharper: Perfect, yeah, this is just the kind of feedback I'm hoping for.  I'll refer back to this example in next week's meeting on the topic.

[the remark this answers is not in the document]
driskull: Another thing would be if the arguments on setValue were more descriptive of what it is going to do. Like if it should emit an event, commit the changes, format the value, etc. IMO the origin logic is kind of confusing.

I'm not sure why we are emitting the native event inside of the custom event either. Is that a pattern/practice we want to do?

Issue #4289, opened 2022-03-23
combobox/combobox-item: tabbing out after selecting chips by clicking leaves the listbox open and focused


Elijbet:
### Actual Behavior

If you use arrow down to add chips and tab out of it, it works correctly, but if you click on items to add chips and tab out if it, it'll proceed to focus on the next element, leaving the current combobox open and focused.

https://user-images.githubusercontent.com/19231036/159601566-d08136e0-5fca-4077-9fe3-c2dd545090b0.mov

### Expected Behavior

Tabbing out of combobox after click & select chips should work exactly like tabbing out of combobox after ArrowUp/Down & select chips, which is by closing the listbox, unfocusing and only then focusing on the next element. 
Follow up with a test. 

### Reproduction Sample

https://codepen.io/elijbet/pen/oNpYbMa

### Reproduction Steps

1. Click on Combobox to open.
2. Press ArrowDown key down to Pine and click Enter to add a chip.
3. Tab and another Tab to see it close the listbox and unfocus.
4. Click on Combobox to open.
5. Click on Pine Nested to add another chip.
6. Tab and another Tab to see it move to the next element without closing the listbox or unfocusing the current element.

### Reproduction Version

v1.0.0-beta.80

### Relevant Info

Related to PR #4218 and Issue #4265

Video:

https://user-images.githubusercontent.com/19231036/159603257-fa3c94a8-7eb0-4ded-8ce0-ea136fc0bc9f.mov

driskull:
This seems to fix it but i'm not sure if its the correct approach.

```patch
diff --git a/src/components/combobox/combobox.tsx b/src/components/combobox/combobox.tsx
index 378b4c5f2f27c7fc4f12390baa17bf4f28169f4a..f39cf1b4ba70fe28aeb4300fca43a6aebee0a3b1 100644
--- a/src/components/combobox/combobox.tsx
+++ b/src/components/combobox/combobox.tsx
@@ -748,6 +748,7 @@ export class Combobox implements LabelableComponent, FormComponent, InteractiveC
       this.resetText();
       this.filterItems("");
     }
+    this.setFocus();
   }
 
   updateAncestors(item: HTMLCalciteComboboxItemElement): void {

``

Elijbet:
> This seems to fix it but i'm not sure if its the correct approach.
> 
> ```diff
> diff --git a/src/components/combobox/combobox.tsx b/src/components/combobox/combobox.tsx
[...]

PR #4464, opened 2022-04-28
fix(combobox): when listbox item is selected with a click, "tab" will now close the listbox and unfocus the element before moving to the next

Elijbet:
**Related Issue:** #4289

## Summary

The calciteComboboxItemChange event Listener wasn’t correctly updating the active item index. This was causing problems tabbing out of the Listbox as well as using the keyboard to pick up where we clicked last.


src/components/combobox/combobox.e2e.ts
```diff
@@ -810,6 +810,71 @@ describe("calcite-combobox", () => {
...
+              <calcite-combobox-item value="Black Eyed Susan"></calcite-combobox-item>
+              <calcite-combobox-item value="Nasturtium"></calcite-combobox-item>
+            </calcite-combobox-item>
+          </calcite-combobox>
+        `
+      );
+      element = await page.find("calcite-combobox");
+      await element.click();
+
+      lisboxItem = await page.find("calcite-combobox-item#PineNested");
```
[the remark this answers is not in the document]
jcfranco: I'd suggest renaming to `comboboxItem` for readability.

src/components/combobox/combobox.e2e.ts
```diff
@@ -810,6 +810,71 @@ describe("calcite-combobox", () => {
...
+      await lisboxItem.click();
+      await page.waitForChanges();
+
+      itemNestedLi = await page.find("calcite-combobox-item#PineNested >>> li");
+      closeEvent = page.waitForEvent("calciteComboboxClose");
+    });
+
+    it("clicking on Listbox item focuses on the item and closes out of Listbox with tab", async () => {
+      expect(itemNestedLi as any).toHaveClass("label--active");
```
jcfranco, line 852: Typing the variables above should eliminate the need to use `any`, which we want to avoid whenever possible.

src/components/combobox/combobox.e2e.ts
```diff
@@ -810,6 +810,71 @@ describe("calcite-combobox", () => {
...
+
+      await element.press("Tab");
+      await closeEvent;
+      await element.press("Tab");
+      expect(await page.evaluate(() => document.activeElement.id)).not.toBe("calcite-combobox");
```
jcfranco, line 857: Should this also assert on the expected focused item? This applies to the following test too.

src/components/combobox/combobox.e2e.ts
jcfranco: Sidebar: it'd be good for reusability to start moving combobox's classes to a `CSS` object in the `resources` module.

jcfranco (approved):

Issue #4469, opened 2022-04-28
Calcite-Input cannot use "up" key to move cursor


tyanthony:
### Actual Behavior

In the `input` component, of type="textarea", using the arrow keys should move the cursor. All keys work correctly except for the "up" key, which currently does not move the cursor.

The same is for a number input - the up key does not increment the number.

### Expected Behavior

- [ ] The Up arrow key should move the cursor up.

This can be seen in the docs site as well.

### Reproduction Sample

https://codepen.io/tyleranthony/pen/MWQgegK?editors=100

### Reproduction Steps

1. Create a new Calcite Input of type="textarea"
2. Type in text so that there are multiple lines
3. Use the arrow keys to move the cursor
4. Notice that the "up" does not move the cursor up

### Reproduction Version

beta.81

jcfranco:
@Elijbet Could you look at this one?

geospatialem:
Verified in next 
⌨️ ⬆️ ✔️

PR #4471, opened 2022-04-28
fix(input): up key moves caret to the beginning of text for text input and text area, for number it only increments

Elijbet:
**Related Issue:** #4469

The old fix was supposed to prevent caret from jumping to the beginning of input while incrementing on ArrowUp held down on input type number, but wasn't implemented correctly. This will bring back the correct functionality so that the ArrowUp key moves the caret to the beginning of text within calcite-input for text and text area inputs but not on number input, where it will increment instead.

For a multiline text area input up key jumps directly above the current spot and the next up key jumps to the beginning of the text.


src/components/input/input.tsx
```diff
@@ -511,6 +507,8 @@ export class Input implements LabelableComponent, FormComponent, InteractiveComp
       return;
     }
     if (event.key === "ArrowUp") {
+      /* prevent default behavior of moving cursor to the beginning of the input when holding down ArrowUp */
+      event.preventDefault();
```
eriklharper, line 511: Having these two separate keyDownHandlers threw me off recently too.  Will be nice when we get around to separating out the number component.

src/components/input/input.e2e.ts
```diff
@@ -202,189 +202,386 @@ describe("calcite-input", () => {
       shadowFocusTargetSelector: "input"
     }));
 
-  it("correctly increments and decrements decimal value when number buttons are clicked and the step precision matches the precision of the initial value", async () => {
-    const page = await newE2EPage();
-    await page.setContent(html` <calcite-input type="number" value="3.123" step="0.001"></calcite-input> `);
+  describe("input type number increment/decrement functionality", () => {
+    let page;
```
jcfranco, line 206: Please add type. This applies to other variables below.

src/components/input/input.e2e.ts
```diff
@@ -202,189 +202,386 @@ describe("calcite-input", () => {
...
+    });
 
-    await numberHorizontalItemUp.click();
-    await page.waitForChanges();
-    expect(await element.getProperty("value")).toBe("10");
-    await numberHorizontalItemUp.click();
-    await page.waitForChanges();
-    expect(await element.getProperty("value")).toBe("10");
+    it("should emit event when up or down clicked on input", async () => {
+      await page.setContent(html` <calcite-input type="number" max="0" value="-2"></calcite-input> `);
```
jcfranco, line 390: Nitpick: remove leading and trailing space between HTML and helper. This applies to other snippets.

src/components/input/input.e2e.ts
```diff
@@ -1350,25 +1359,139 @@ describe("calcite-input", () => {
...
+        await page.$eval("calcite-input", (element: HTMLInputElement) => {
+          const textarea = element.shadowRoot.querySelector("textarea");
+          textarea.focus();
+          return textarea.selectionStart === textarea.value.length;
+        })
+      ).toBeTruthy();
+    });
+
+    it("works for type text", async () => {
+      await page.setContent(`<calcite-input id="text-element" type="text" ></calcite-input>`);
```
jcfranco, line 1401: Nitpick: extra space near the end of opening tag.

src/components/input/input.e2e.ts
```diff
@@ -1350,25 +1359,139 @@ describe("calcite-input", () => {
...
+          textarea.focus();
+          return textarea.selectionStart === 0;
+        })
+      ).toBeTruthy();
+
+      await page.keyboard.press("ArrowDown");
+      await page.waitForChanges();
+
+      expect(
+        await page.$eval("calcite-input", (element: HTMLInputElement) => {
```
jcfranco, line 1392: Since a big chunk of this function is reused a bit, can you extract it to a reusable one that returns the caret index and assert on that?

src/components/input/input.e2e.ts
```diff
@@ -1350,25 +1359,139 @@ describe("calcite-input", () => {
...
+      await page.keyboard.type("test");
+      await page.waitForChanges();
+
+      await page.keyboard.press("ArrowUp");
+      await page.waitForChanges();
+
+      expect(
+        await page.$eval("calcite-input", (element: HTMLInputElement) => {
+          const input = element.shadowRoot.querySelector("input");
+          input.focus();
```
jcfranco, line 1414: Do you need to focus here? I think the caret index is available regardless of the input being focused or not.

src/components/input/input.e2e.ts
```diff
@@ -1350,25 +1359,139 @@ describe("calcite-input", () => {
...
-  describe("external/programmatic changes to the value", () => {
-    it("incrementing correctly updates the value after focus and blur events", async () => {
-      const page = await newE2EPage();
-      await page.setContent(html` <calcite-input type="number" value="1"></calcite-input> `);
+    beforeEach(async () => {
+      page = await newE2EPage();
+    });
+
+    it("works for type textarea", async () => {
+      await page.setContent(`<calcite-input id="text-element" type="textarea" ></calcite-input>`);
```
jcfranco, line 1370: If the `id` isn't referenced, let's drop it. This applies to other HTML snippets.

src/components/input/input.e2e.ts
```diff
@@ -1350,25 +1359,139 @@ describe("calcite-input", () => {
     expect(await element.getProperty("value")).toBe("-123");
   });
 
-  describe("is form-associated", () => {
-    it("supports type=text", () => formAssociated("calcite-input", { testValue: "test" }));
-    it("supports type=number", () => formAssociated("<calcite-input type='number'></calcite-input>", { testValue: 5 }));
-  });
+  describe("ArrowUp/ArrowDown function as Home/End within calcite-input", () => {
```
jcfranco, line 1362: To avoid association with `Home`/`End` keys (e.g., scrolling page), can you replace `Home/End` with something that describes the caret behavior?

jcfranco (approved):
Aside from some comments, this LGTM! 🎉🌮🦖🦃

jcfranco:
@Elijbet Also, before merging, can you revisit the PR title to add more context about the fix? It's not clear to me what exactly got fixed from it.

Issue #4187, opened 2022-03-03
calcite-input-date-picker: minAsDate/maxAsDate not respected


ethanbdev:
### Actual Behavior

When setting the `minAsDate` and `maxAsDate`, the min/max is not respected and the user can change the date to something before/after the defined min and max.

### Expected Behavior

The user should not be able to select a date outside the bounds of the min/max.

### Reproduction Sample

https://codepen.io/eborgen/pen/xxPmvBq

### Reproduction Steps

1. Open the date picker
2. Observe the year can be changed to before 1950 / after 2050

### Reproduction Version

beta.77

### Regression?

unknown

ethanbdev:
Possibly related to #4186

benelan:
verified on next

PR #4486, opened 2022-05-04
fix(input-date-picker): no longer allows selection before minAsDate and after maxAsDate

anveshmekala:
**Related Issue:** #4187 

## Summary

 This fix will ensure user cannot select date before `minAsDate` and after `maxAsDate` in `calcite-input-date-picker`


driskull (approved):
LGTM!

Issue #2256, opened 2021-06-03
Refactor: Use calciteInternal prefix for internal events


jcfranco:
### Description

For clarity and to minimize unintentional usage, internal events should:

1. be renamed to use the `calciteInternal` prefix to make it easier to distinguish between internal/external events. Most internal events should be identifiable by the `@internal` JSDoc tag.
2. be stopped from propagating when handled.

Sidebar, an ESLint rule can be authored to catch all internal events that did not have `stopPropagation`/`stopImmediatePropagation` called. This would be handled in https://github.com/Esri/eslint-plugin-calcite-components.

PR #4496, opened 2022-05-06
refactor: use calciteInternal prefix for internal events

Elijbet:
**Related Issue:** #2256

## Summary

Renaming internal events to use calciteInternal prefix and stopping them from propagating when handled (where propagation is not needed).

## List of Internal Events revised:

calciteAccordionItemKeyEvent
calciteAccordionItemSelect
calciteAccordionItemClose
calciteItemRegister
calciteAlertSync
calciteAlertRegister
calciteComboboxOpen
calciteComboboxClose
calciteDayHover
calciteDatePickerHover
calciteDatePickerMouseOut
calciteDropdownItemChange
calciteDropdownItemSelect
calciteDropdownItemKeyEvent
calciteDropdownCloseRequest
calciteInlineEditableEnableEditingChange
stop and check if all green first
calciteInputFocus
calciteInputBlur
calciteInputDatePickerOpen
calciteInputDatePickerClose
calciteInternalLabelClick
calciteOptionChange
calciteOptionGroupChange
calciteListItemPropsChange
calciteListItemValueChange
calciteInternalRadioButtonBlur
calciteInternalRadioButtonCheckedChange
calciteInternalRadioButtonFocus
calciteRadioGroupItemChange
calciteStepperItemChange
calciteStepperItemKeyEvent
calciteStepperItemSelect
calciteStepperItemRegister
calciteTabRegister
calciteInternalTabChange
calciteInternalTabsActivate
calciteTabsFocusNext
calciteTabsFocusPrevious
calciteTabTitleRegister
calciteTabTitleRegister
calciteTimePickerBlur
calciteTimePickerChange
calciteTimePickerFocus
calciteTreeItemSelect


src/components/combobox/combobox.tsx
```diff
@@ -257,13 +257,13 @@ export class Combobox implements LabelableComponent, FormComponent, InteractiveC
    * Fired when the combobox is opened
    * @internal
    */
-  @Event() calciteComboboxOpen: EventEmitter;
+  @Event() calciteInternalComboboxOpen: EventEmitter;
```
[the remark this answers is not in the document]
driskull: https://github.com/Esri/calcite-components/issues/1688

driskull (approved):
Looks good! 💯

src/components/combobox/combobox.tsx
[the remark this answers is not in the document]
Elijbet: Should we be removing @internal from Open/Close events then and making them public? Or did you mean specifically for combobox?

src/components/accordion-item/accordion-item.tsx
```diff
@@ -184,14 +185,15 @@ export class AccordionItem {
...
-  @Listen("calciteAccordionChange", { target: "body" })
+  @Listen("calciteInternalAccordionChange", { target: "body" })
   updateActiveItemOnChange(event: CustomEvent): void {
     this.requestedAccordionItem = event.detail
       .requestedAccordionItem as HTMLCalciteAccordionItemElement;
     if (this.el.parentNode !== this.requestedAccordionItem.parentNode) {
       return;
     }
     this.determineActiveItem();
+    event.stopPropagation();
```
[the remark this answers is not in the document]
Elijbet: I suppose I could remove the stopPropagation from this handler in that case, since it's practically meaningless in this scenario?

src/components/input/input.tsx
```diff
@@ -360,12 +360,12 @@ export class Input implements LabelableComponent, FormComponent, InteractiveComp
...
    * @internal
    */
-  @Event() calciteInputFocus: EventEmitter;
+  @Event() calciteInternalInputFocus: EventEmitter;
 
   /**
    * @internal
    */
-  @Event() calciteInputBlur: EventEmitter;
+  @Event() calciteInternalInputBlur: EventEmitter;
```
eriklharper, line 368: I'm actually trying to remember why these two events are actually labeled `@internal`.  These seem like public events that consumers should have access to.

src/components/combobox/combobox.tsx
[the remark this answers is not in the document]
eriklharper: I concur with @driskull on this one too.  Open/Close events should probably be publicly exposed.  Especially since this is a top-level component that isn't emitting this event for another Calcite component to listen for.  In that case, just removing the `@internal` label on it and restoring the original event name should suffice.

src/components/accordion-item/accordion-item.tsx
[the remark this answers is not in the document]
driskull: Lets just keep it here.

src/components/combobox/combobox.tsx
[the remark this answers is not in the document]
driskull: I'm just questioning why these are internal. We may want to expose them at some point but we can do so later if needed.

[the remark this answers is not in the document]
Elijbet: Should there be a follow-up issue for all top-level components to publicly expose Open/Close events, or does this need to be discussed for all components, including nested?

src/components/input/input.tsx
driskull: I'm not sure about the inputblur. That native event should bubble up.

The open/close ones seem like they should be public.

eriklharper: Should shadowed `<input>` element events bubble up when there are `calciteInput*` event equivalents?  Seems redundant to allow both, but this is probably why these were labeled `@internal` in the first place because we were relying on the native events to bubble.

src/components/combobox/combobox.tsx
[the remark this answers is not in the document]
eriklharper: I like the idea of starting a pattern discussion on open/close events and if places that use them should expose them publicly if they aren't already.  Let's create a discussion for this here: https://github.com/Esri/calcite-components/discussions

src/components/input/input.tsx
driskull: I don't think there can be equivalents because CustomEvents aren't the same as Keydown, Click, Blur, etc.

I think we should let the native events bubble out.

I don't like how we're passing the native events in some Custom events either. Seems kind of odd to do that.

eriklharper: I guess I'm trying to understand the distinction between the native events and the custom events.  What are the custom "Input" and "Change" events really there for if the native ones are already bubbling out and usable?

eriklharper: > I'm actually trying to remember why these two events are actually labeled @internal.

I think I remember now.  It is most likely for other Calcite components that consume Calcite Input in their shadow DOM for styling or other purposes.

eriklharper: > I don't like how we're passing the native events in some Custom events either. Seems kind of odd to do that.

I guess if we chose to stop bubbling the native events, this would be the way to still expose them to listeners via the custom event.

eriklharper: FYI, created a discussion for this here: https://github.com/Esri/calcite-components/discussions/4546

eriklharper: > I don't think there can be equivalents because CustomEvents aren't the same as Keydown, Click, Blur, etc.

Should we then ditch redundant custom events like `calciteInputBlur` and `calciteInputFocus` and any other ones that can rely on native bubbled events?

jcfranco:
> I'm actually trying to remember why these two events are actually labeled @internal.

From https://github.com/Esri/calcite-components/issues/1705:

> Note: focus/blur events were made internal to allow for more general calciteFocus/calciteBlur events.

gpbmike:
Just leaving a note, I think this change should have been called out in the breaking changes section of the release notes.

I was using `calciteInputFocus` and `calciteInputBlur`. 😢

jcfranco:
Those props were marked as internal and omitted from the doc for quite a while now, hence why we didn't mark them as breaking. On the other hand, I understand the need for these events and will look into this in upcoming sprints (see https://github.com/Esri/calcite-components/issues/4611#issuecomment-1194410198).

gpbmike:
> Those props were marked as internal and omitted from the doc for quite a while now

Fair enough. I'm working on upgrading from beta.43 to beta.86. Not sure if the internal tag was there before, if so apologies. Thanks.

Issue #4495, opened 2022-05-05
Calcite Label: Slows more and more each rerender


paulcpederson:
### Actual Behavior

When rendering several labels with inputs inside (may happen with other content in the label, not sure) the time to paint gets slightly slower each go around. By the third or fourth time, there is a noticable delay. 

### Expected Behavior

Ideally, building calcite labels en masse should be as speedy as it was in beta.78

### Reproduction Sample

https://codepen.io/paulcp/pen/WNMQjNx?editors=1011

### Reproduction Steps

1. Click "build inputs"
2. Click "clear inputs"
3. Repeat steps 1 and 2 noting the slightly increased waiting time 

### Reproduction Version

beta.81, v1.0.0-next.457

### Relevant Info

Related pr: https://github.com/Esri/calcite-components/pull/4413

This pr improves performance a lot, but seems like a smaller memory leak may remain?

### Regression?

1.0.0-beta.78

geospatialem:
Verified on next ✔️
🐢 ➡️  🐇

PR #4498, opened 2022-05-06
fix(label): fix memory leak from label util

jcfranco:
**Related Issue:** #4495

## Summary

This updates the label util to:

* remove mapped entries for bound label-connect/disconnect listeners for when a labelable component is removed (no need to track these anymore) 
* reuse bound label-connect listeners for when a label is disconnected


driskull (approved):
Way to fix a leak!

Elijbet (approved):
Can confirm it passes manual testing!

PR #4502, opened 2022-05-06
docs(tab-title): calciteTabsActivate emits tab property or index position

benelan:
**Related Issue:** #

## Summary
Let users know that setting the tab property affects what calciteTabsActivate emits.


driskull (approved):
👍

Issue #4489, opened 2022-05-04
[Bug] dismissed panels have a visible artifact 


benelan:
### Actual Behavior

Dismissed panels should not be visible at all. However, there is a small slice of an artifact where the panel goes when it is not dismissed.
![image](https://user-images.githubusercontent.com/10986395/166831910-e997270c-2aa0-47b2-8177-0697cfecaae4.png)
(The top of the image above)

### Expected Behavior

No visual artifacts when panels are dismissed

### Reproduction Sample

https://codepen.io/benesri/pen/jOZPpzJ?editors=1010

### Reproduction Steps

1. Go to the sample
2. Notice the visual artifact where the top of the panel will be

### Reproduction Version

beta.81

### Relevant Info

Noticed it when answering a [community question](https://community.esri.com/t5/calcite-design-system-questions/close-button-for-basemap-widget/m-p/1170866#M131). The codepen sample I responded with in the thread has a hacky workaround that uses both `dismissed` with the global `hidden` attribute.

### Regression?

beta.69

driskull:
This is actually an issue with the shell-panel. It's styling the slotted panel even when its dismissed.

```
slot[name="action-bar"]::slotted(calcite-action-bar), .content ::slotted(calcite-flow), .content ::slotted(calcite-panel) {
    border-width: 1px;
    border-style: solid;
    border-color: var(--calcite-ui-border-3);
}
```

driskull:
It could also be fixed by setting the host of the calcite-panel to visibility hidden when dismissed.

benelan:
Triaging this to the next sprint since it sounds like an easy fix. cc @jcfranco @macandcheese

PR #4503, opened 2022-05-06
fix(shell-panel): Remove border on slotted panels that are dismissed

driskull:
**Related Issue:** #4489

## Summary

fix(shell-panel): Remove border on slotted panels that are dismissed.  (#4489)


jcfranco (approved):
💥

Issue #4462, opened 2022-04-27
Console error + rendering bug when `calcite-radio-button` does not have `checked` attr


jgibson02:
### Actual Behavior

This error is observed when using a `calcite-radio-button` without a `checked` attribute. I believe this line might be the cause since `this.checked` may be `undefined`.
https://github.com/Esri/calcite-components/blob/083d283caf719eed91cb1f5bd71b4584864b5e61/src/components/radio-button/radio-button.tsx#L409

![image](https://user-images.githubusercontent.com/5069711/165609789-be74980a-6010-424c-ad83-556b2ed98564.png)

### Expected Behavior

No call to `toString()` when undefined, perhaps `aria-checked` can fallback to `"false"`?

### Reproduction Sample

https://codesandbox.io/s/calcite-radio-button-checked-tostring-console-error-rjifui

### Reproduction Steps

1. Inspect the Codesandbox console after the calcite-radio-button renders

### Reproduction Version

beta.81

jgibson02:
It might be particular to `calcite-components-react`, but I also noticed it with the `active` prop on `calcite-accordion-item`. Maybe an issue with all boolean props? Added an example for the Accordion to the codesandbox.

driskull:
@jcfranco should we create a utility for these boolean props to string? Something that does `(!!this.myProp).toString()`?

Or maybe we can patch the JSX or get JSX to implement the above if the property is boolean?

jcfranco:
> should we create a utility for these boolean props to string? Something that does (!!this.myProp).toString()?

Good idea.

@jgibson02 Just curious, did you recently switch to React 17? The React wrappers were recently updated to support v16+ (not just v16), so I'm wondering if this is something new or we just hadn't noticed. cc @benelan

jgibson02:
> > should we create a utility for these boolean props to string? Something that does (!!this.myProp).toString()?
> 
> Good idea.
> 
> @jgibson02 Just curious, did you recently switch to React 17? The React wrappers were recently updated to support v16+ (not just v16), so I'm wondering if this is something new or we just hadn't noticed. cc @benelan

@jcfranco in our actual project we have not, we're on v16.13.1, but the codesandbox sample is on React 17

geospatialem:
Verified with `0.23.0` ✔️ 

![calcite radio button with aria-checked fix](https://user-images.githubusercontent.com/5023024/167885857-89c8b8bd-efd5-484d-97df-6c7984d28046.png)

PR #4504, opened 2022-05-07
fix: Ensure components set boolean aria values properly

driskull:
**Related Issue:** #4462

## Summary

fix: Ensure components set boolean aria values properly. (#4462)

This helper makes sure that boolean aria attributes are properly converted to a string. Otherwise, undefined or null values will cause a JS Error. We need to ensure aria attributes that are boolean should be "true" or "false".

```ts
/**
 * This helper makes sure that boolean aria attributes are properly converted to a string.
 *
 * It should only be used for aria attributes that require a string value of "true" or "false".
 *
 * @returns {string} The string conversion of a boolean value ("true" | "false").
 */
```


eriklharper (approved):
LGTM 👍

src/utils/dom.ts
```diff
@@ -274,3 +274,7 @@ export function intersects(rect1: DOMRect, rect2: DOMRect): boolean {
     rect2.bottom < rect1.top
   );
 }
+
+export function toAriaBoolean(value: boolean): string {
```
jcfranco, line 278: Since this function could apply to non-ARIA attributes, how about `toBooleanAttr`?

jcfranco, line 278: Could you add a comment about why we need this util? It won't be clear from our codebase, methinks.

jcfranco (approved):
Awesome! 🎉

src/utils/dom.ts
driskull: I thought about that but it seems like `aria-` is the **only** place we are using it and where it really should be used.

driskull: Yes

driskull: Most boolean attributes we don't want to be strings. If we do need to rename, I would recommend `toBooleanString()` instead but I actually prefer `toAriaBoolean` because it says what it is for.

jcfranco: Sold! `toAriaBoolean` it is.

Issue #2622, opened 2021-07-26
Docs: update demos/ index page to include links to all pages, remove old styles


caripizza:
### Description
Right now we have some pages linked to on the main index.html page, but there are many other pages you can only get to by manually typing demos/ path into the browser.

### Acceptance Criteria 
1. All demos/ html pages appear as clickable links on the main index.html file.
2. All demos pages source from the same updated styles (ie., no more gray background color pages with "Courier New" font).

caripizza:
@julio8a should this one be worked on in the current sprint or next sprint?

julio8a:
Not this sprint. Will look at the next sprint.

benelan:
There are a couple more demos that are missing from the landing page:
- graph
- input-date-picker
- theme
- animation
- handle
- shell (pick whichever you think is best)

benelan:
verified on master

PR #4513, opened 2022-05-09
docs(demos): finalize demo homepage

geospatialem:
**Related Issue:** #2622

## Summary
1. Added remaining demos to the demo homepage, including:
    *  graph
    * input-date-picker
    * theme
    * animation
    * handle
    * shell
2. Updated demos with shell inclusion
    * shell only spans the full window height
3. Theme switcher added to theme and shell example for the JSAPI
    * map will also switch when theme switch is enabled (light vs. dark)


benelan (approved):
Nice!

Issue #4459, opened 2022-04-27
Focus outline does not always hug `stepper-item` title


nwhittaker:
### Actual Behavior

If a `calcite-stepper` has a `min-height` style, the height of the focus outline for each `stepper-item` title can become quite tall.

<img width="1204" alt="Screen Shot 2022-04-27 at 12 45 31 PM" src="https://user-images.githubusercontent.com/421496/165577910-f055f8a1-2377-4581-9c4e-76ee1b9dc021.png">

### Expected Behavior

The focus outline hugs the `stepper-item` title regardless of the minimum height of the `stepper`.

<img width="1204" alt="Screen Shot 2022-04-27 at 12 45 59 PM" src="https://user-images.githubusercontent.com/421496/165578069-c130197b-22f5-4b99-8c66-4d6f3b48820e.png">

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/qBpzgPP

### Reproduction Steps

1. Visit the sample
2. Click/focus one of the stepper titles

### Reproduction Version

1.0.0-next.451

### Relevant Info

I can produce the expected behavior by adding an `align-items: start` style to the `stepper` element.

geospatialem:
Verified on `next`. ✅ 

![calcite stepper title focus outline](https://user-images.githubusercontent.com/5023024/167878705-2ec77c0e-e3c9-4241-aec4-02d9f5ebd3d6.png)

PR #4515, opened 2022-05-09
fix(stepper-item): focus outline wraps item title

anveshmekala:
**Related Issue:** #4459 

## Summary

This fix will ensure focus outline wraps around `stepper-item` title


driskull (approved):
👍

Issue #4036, opened 2022-02-01
[Design] Remove transition from focus style


macandcheese:
### Description

This is a currently a bullet point in https://github.com/Esri/calcite-components/issues/3345 - but it can be accomplished independently. 

Our transition for focus outline can cause components to appear slow, or laggy. It can also cause a situation where quickly tabbing through a set of components leads to a situation where the focus ring can't keep up.

### Acceptance Criteria

Remove the focus outline transition. I do not think this needs to be customizable.

### Relevant Info

cc @asangma @mitc7862 @TheBlueDog @rmstinson

### Which Component

All the components. Rating, Dropdown, are some that appear visibly sluggish.

### Example Use Case

As a user, sometimes the transition of focus ring can cause the application to appear slow, or if as a keyboard user I quickly tab, the currently focused element may not appear in focus.

geospatialem:
Verified on next

https://user-images.githubusercontent.com/5023024/170577788-7e9aaea0-8334-49e6-baac-e0ff42d4a8bf.mov

PR #4516, opened 2022-05-09
feat: drop focus transitions to look snappier

jcfranco:
**Related Issue:** #4036 

## Summary

🤌💥

This PR modifies the focus utils to no longer transition on outline-related styles, so focus movement is immediate.


src/components/input/input.scss
```diff
@@ -99,7 +99,6 @@
 
 :host textarea,
 :host input {
-  transition: var(--calcite-animation-timing), height 0s;
```
jcfranco, line 102: @macandcheese Can you double check this one? I had to remove this to avoid some flickering going on when the focus outline got displayed. This was introduced [here](https://github.com/Esri/calcite-components/commit/533c270ec2e22ba1db04261189fc29bd3e7bfc82#diff-4185271e827d2b26e19a1a16201fb3dc4a31448b796c61107967bb5cfe2ca722R65).

jcfranco: Disregard the above. The flickering was caused by `outline-offset` being transitioned. I've explicitly prevented it from being animated on focusable elements where the transition property default is `all`.

macandcheese: Yep, IIRC the height was there to prevent the input of type textarea from appearing laggy when resizing vertically, so as long as that doesn’t happen 👍

jcfranco:
@macandcheese @eriklharper Screener reports some tile-select diffs regarding (checkbox/radio) focus states. I have not been able to observe the reported diff locally either through keyboard, mouse and programmatic focus. Would you suggest proceeding with this and keeping an eye out for differences?

**Note**: other visual diffs are expected (e.g., width difference between box-shadow and outline approach).

eriklharper:
> @macandcheese @eriklharper Screener reports some tile-select diffs regarding (checkbox/radio) focus states. I have not been able to observe the reported diff locally either through keyboard, mouse and programmatic focus. Would you suggest proceeding with this and keeping an eye out for differences?
> 
> **Note**: other visual diffs are expected (e.g., width difference between box-shadow and outline approach).

I'd like to open a Sauce support ticket just in case something they know about on their end that changed could be the culprit...

jcfranco:
@macandcheese @eriklharper Merging per our conversation. 🎉

Issue #4518, opened 2022-05-10
[Bug] Graph does not render


benelan:
### Actual Behavior

In beta.74 calcite-graph renders, in beta.75 it does not. There are no errors in the console.

### Expected Behavior

Expect calcite-graph to render

### Reproduction Sample

https://codepen.io/benesri/pen/GROgvRJ

### Reproduction Steps

1. Wait for the graph to render (it takes a couple seconds)
2. Switch the version to beta.75
3. Notice the graph does not render

### Reproduction Version

beta.81

### Relevant Info

Triaging to this sprint - cc @macandcheese @jcfranco 

### Regression?

beta.74

driskull:
@benelan currently, the graph requires a height to be set on it to render. Previously, it had a fixed height set of 100px. The slider sets a height on it so that's why it renders.

We can...

- Set the graph to `100%` height but its container will have to have a height.
- Set the default graph height to `100px` in css so it will render at least.

Let me know what sounds best. cc @jcfranco

benelan:
> Set the graph to 100% height but its container will have to have a height.

I personally like this option in general, its better for responsive designs. cc @macandcheese for design

PR #4530, opened 2022-05-11
fix(graph): Set default height to 100%

driskull:
**Related Issue:** #4518

## Summary

fix(graph): Set default height to 100%. (#4518)

The graph will now take up the available space of its parent.

Adds screener tests too.


jcfranco (approved):
Nice!

Issue #4539, opened 2022-05-12
calcite-popopver: popovers open without being set to open


AdelheidF:
### Actual Behavior

This started with beta.82, beta.81 was fine.

If I don't use popover-manager and create several popovers that I manage myself a seemingly random popover opens automatically when the user clicks on the reference element, which in my case is the calcite-panel.

### Expected Behavior

popover only opens when I tell it to

### Reproduction Sample

https://codepen.io/driskull/pen/WNMGJNO?editors=1010

https://codepen.io/afreitag/pen/ExQgLYO

### Reproduction Steps

use effects, sketch or background color in Map Viewer

### Reproduction Version

beta.82

### Regression?

beta.81

driskull:
The issue is that previously a user could not use the `calcite-popover-manager` and control the popover opening/closing themselves.

When we removed that component, we failed to add a way to opt out of it.

driskull:
@jcfranco can we strategize a release fix plan for this? Likely will be a breaking change.

AdelheidF:
looks good, tested with next.468

AdelheidF:
I would think this is a breaking change for users, not only a feature. Or you could mark the install with beta.82 as a breaking change.

driskull:
I'll defer to @jcfranco here. Thanks Adelheid.

jcfranco:
If standalone popover usage was supported after the manager was introduced then it's a breaking change and we can update the changelog accordingly.

benelan:
If standalone popover usage was supported after the manager was introduced then it's a breaking change and we can update the changelog accordingly.

> Will do

driskull:
> If standalone popover usage was supported after the manager was introduced then it's a breaking change and we can update the changelog accordingly.

I don't think it was supported or advertised so maybe we don't need to mark it as breaking.

PR #4540, opened 2022-05-12
feat(popover): Add triggerDisabled property to disable automatically toggling a popover via mouse or keyboard

driskull:
**Related Issue:** #4539

## Summary

feat(popover): Add triggerDisabled property to disable automatically toggling a popover via mouse or keyboard. (#4539)


jcfranco (commented):
Looking good. A couple of questions:

1. Can you confirm whether standalone popover usage was supported (before we deprecated the manager)? Looks like all examples and tests had the manager set up, so this could be the deciding point between a `feat` or `fix!`.
2. Is there another name to describe the toggling behavior? I find `autoOpenDisabled` a bit misleading since it handles both opening and closing when the reference element is triggered.

If we deem 1 a breaking change, can you use `BREAKING CHANGE` (not plural) in the commit body to [adhere to conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#:~:text=in%20Semantic%20Versioning\).-,BREAKING%20CHANGE,-%3A%20a%20commit)?

driskull:
> Can you confirm whether standalone popover usage was supported (before we deprecated the manager)? Looks like all examples and tests had the manager set up, so this could be the deciding point between a feat or fix!.

I don't think we had any documentation stating that the popover could be used without the manager so maybe we can get away with it being a feat. 

However, the mapviewer peeps have been using it without one since the beginning in some instances.

> Is there another name to describe the toggling behavior? 

We could do `autoToggleDisabled`. Any suggestions?

jcfranco (approved):
💥

Issue #4538, opened 2022-05-12
Popover keyboard navigation regression


nwhittaker:
### Actual Behavior

Focusing a popover's reference element and then pressing the `space` or `enter` key does not open the popover.

### Expected Behavior

Focusing a popover's reference element and then pressing the `space` or `enter` key does open the popover.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/eYVdywq?editors=1000

### Reproduction Steps

1. Visit the code sample
2. Tab to the **Activate Popover** button
3. Press the `space` or `enter` key

### Reproduction Version

1.0.0-beta.82

### Relevant Info

A possibly [related issue][1] is whether or not keyboard navigation should also work if the reference element is not a button.

[1]: https://teams.microsoft.com/l/message/19:fd15b51dacd24e70895ec1218a54ae06@thread.skype/1652372269215?tenantId=aee6e3c9-711e-4c7c-bd27-04f2307db20d&groupId=56fae21a-9407-4943-859f-a9bfcf0bbad3&parentMessageId=1652372269215&teamName=Calcite%20Design%20System&channelName=Calcite%20Components&createdTime=1652372269215

### Regression?

1.0.0-beta.81

driskull:
@benelan can we triage this for this release.

benelan:
@driskull Triage'd

geospatialem:
Also adding in the `a11y` label for tracking purposes. 👣

PR #4541, opened 2022-05-12
fix(popover): Fix toggling popover via keyboard

driskull:
**Related Issue:** #4538

## Summary

fix(popover): Fix toggling popover via keyboard. (#4538)


jcfranco (approved):
🎉🦃🦖🌮

PR #4550, opened 2022-05-13
build(deps): install jsdoc eslint plugin

benelan:
**Related Issue:** #

## Summary
Added [`eslint-plugin-jsdoc`](https://www.npmjs.com/package/eslint-plugin-jsdoc) with the recommended settings, except:
- Turned off [`require-jsdoc`](https://www.npmjs.com/package/eslint-plugin-jsdoc#user-content-eslint-plugin-jsdoc-rules-require-jsdoc) which created empty jsdoc for every single function that wasn't doc'd. It could be worth setting it to `publicOnly` so that all of our exported functions are documented. Thoughts?
- Turned off [`check-tag-names`](https://www.npmjs.com/package/eslint-plugin-jsdoc#user-content-eslint-plugin-jsdoc-rules-check-tag-names) which threw warnings for our `slot` tags.

I have a separate `style` PR where I linted the code: #4551

With the current settings, if a function has a jsdoc (including something like `/** does XYZ */`) it will automatically add `@param` tags when relevant. For example, it added [this](https://github.com/Esri/calcite-components/blob/benelan/lint-code/src/components/date-picker-month/date-picker-month.tsx#L258-L260). the plugin will then throw warnings because the param type/description is missing. Should we turn that off too? - [`require-param`](https://www.npmjs.com/package/eslint-plugin-jsdoc#user-content-eslint-plugin-jsdoc-rules-require-param)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -57,6 +57,8 @@
     "curly": "error",
     "jest/expect-expect": "off",
     "jest/no-export": "warn",
+    "jsdoc/require-jsdoc": "off",
+    "jsdoc/check-tag-names": "off",
     "lines-between-class-members": ["error", "always"],
     "no-multiple-empty-lines": ["error", { "max": 1 }],
     "no-unneeded-ternary": "error",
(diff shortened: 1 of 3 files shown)


driskull:
publicOnly makes sense in theory, we'll have to see in practice :)

driskull (approved):
LGTM!

Issue #4558, opened 2022-05-16
Tabs: High contrast mode removes most visual information from tabs


mpriour:
### Summary

The lack of visual information for tabs was identified as a level 3 severity issue (missing critical information) in the [High Contrast Audit](https://github.com/Esri/calcite-components/wiki/High-Contrast-Audit#tabs-normal)

Problems include:
* Unselected tabs look selected
* The selected tab indicator is not visible
* All tab styles are heavily outlined
* The bordered tabs are confusing since everything is outlined

### Actual Behavior

High Contrast "normal" tabs
<img width="437" alt="high contrast normal tabs" src="https://user-images.githubusercontent.com/142636/168641180-4f64805c-f4e0-433a-9a60-51aa9942d84d.png">

High Contrast bordered tabs
<img width="446" alt="high contrast bordered tabs" src="https://user-images.githubusercontent.com/142636/168641847-5bf4085d-6d19-4fa0-a7ec-7a322feeb6c4.png">

### Expected Behavior

High contrast mode should show equivalent visual information as normal mode

Normal tabs
<img width="443" alt="normal tabs" src="https://user-images.githubusercontent.com/142636/168642332-4ba020b5-fb90-4fb1-a631-d862ac5f532b.png">

Bordered tabs
<img width="446" alt="bordered tabs" src="https://user-images.githubusercontent.com/142636/168642383-3f273e01-2d55-4a33-afa4-131616366d51.png">

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/tabs/

### Reproduction Steps

Windows
1. Set Windows color theme to a high-contrast variant.
2. Open the reproduction sample (https://developers.arcgis.com/calcite-design-system/components/tabs/) in any browser.
3. Observe the issue.

Mac
1. Open the reproduction sample (https://developers.arcgis.com/calcite-design-system/components/tabs/) in Chrome or Edge.
2. Open Dev Tools.
3. Click on Elements tab.
4. Next to the console sub tab, there should be a "Rendering" tab (if not you will need to enable it)
5. On the Rendering tab choose Forced-colors: active from the "Emulate forced colors..." selector
6. Observe the issue.

### Reproduction Version

latest

### Relevant Info

specific High contrast (ie forced-colors) styles should be selectively applied to provide the visual information that is currently missing.

geospatialem:
Verified on next.

[...]

PR #4559, opened 2022-05-16
fix(tabs): apply forced colors styles where needed

mpriour:
* Removes excessive outlining
* Adds obvious selection indicator
* Adds specific outline color for focus

**Related Issue:** #4558 

## Summary
Add forced color (high contrast) styles to tabs

### DIFF
--- a/src/components/tab-nav/tab-nav.scss
+++ b/src/components/tab-nav/tab-nav.scss
@@ -63,3 +63,9 @@
   bottom: 0; // display active blue line below instead of above
   top: unset;
 }
+
+@media (forced-colors: active) {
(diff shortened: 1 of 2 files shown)


macandcheese (approved):
Thanks for this as always @mpriour !

mpriour:
Actually just realized that the bordered tabs are not quite fixed.
Reverting to draft status

mpriour:
@macandcheese Added the correct styles for bordered tabs as well.

Fix produces:

normal tabs
Screen Shot 2022-05-16 at 12 08 14 PM


bordered tabs
Screen Shot 2022-05-16 at 1 07 40 PM

Issue #4564, opened 2022-05-17
Stepper: High contrast stepper has too many borders


mpriour:
### Summary

The lack of visual information for stepper was identified as a level 3 severity issue (missing critical information) in the [High Contrast Audit](https://github.com/Esri/calcite-components/wiki/High-Contrast-Audit#stepper)

The main issue is that everything is outlined and there is no indicator for what step in the stepper is selected.

### Actual Behavior

All stepper "tabs" are outlined with no indicator for selection/active step

<img width="1015" alt="Screen Shot 2022-05-17 at 8 03 04 AM" src="https://user-images.githubusercontent.com/142636/168817145-45c502cd-deb7-40ad-86d4-953cd7fe159e.png">

### Expected Behavior

A high contrast version of the existing "normal" mode design

<img width="1015" alt="Screen Shot 2022-05-17 at 8 05 40 AM" src="https://user-images.githubusercontent.com/142636/168817712-43f27011-48d8-4954-8a5a-baa156aced38.png">

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/stepper/

### Reproduction Steps

Windows
1. Set Windows color theme to a high-contrast variant.
2. Open the reproduction sample (https://developers.arcgis.com/calcite-design-system/components/stepper/) in any browser.
3. Observe the issue.

Mac
1. Open the reproduction sample (https://developers.arcgis.com/calcite-design-system/components/stepper/) in Chrome or Edge.
2. Open Dev Tools.
3. Click on Elements tab.
4. Next to the console sub tab, there should be a "Rendering" tab (if not you will need to enable it)
5. On the Rendering tab choose Forced-colors: active from the "Emulate forced colors..." selector
6. Observe the issue.

### Reproduction Version

latest

### Relevant Info

Excessive visual noise can be eliminated and an obvious selected state achieved with high contrast (forced-colors) specific styling.

benelan:
verified on beta.83

PR #4565, opened 2022-05-17
fix(stepper): Add high contrast specific styles to stepper

mpriour:
**Related Issue:** #4564 

## Summary
* Removes the excessive outlines
* Adds an obvious selection indicator
* Make focus state obvious and work on different layouts

Changes produce this layout:
Screen Shot 2022-05-17 at 7 54 06 AM

Colored states for non-active steps is not something that can be done in the high contrast mode. Additional text or visual clues will need to be used if a design is to work well in both normal and high contrast mode.


geospatialem (approved):
Great addition, @mpriour! 🎆🎇

macandcheese (approved):
Thanks @mpriour !

Issue #4291, opened 2022-03-23
[Input Message] Fix overzealous animation


macandcheese:
### Actual Behavior

When making an input-message active by adding the attribute, currently the animation is rather ... in your face. I think we should tone this down a bit. 

https://user-images.githubusercontent.com/4733155/159779399-1a3c561b-3537-40f4-8964-d46e90f244eb.mov

![bam-cooking-show](https://user-images.githubusercontent.com/4733155/159779330-d38e9a9c-e07c-4648-9c95-06c24b0f8322.gif)

### Expected Behavior

Don't use a scale-in effect to display the input message. 

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/ExoPEBe?editors=1111

### Reproduction Steps

In codepen:
Click "add" in the slotted input button without entering anything in the input. See the wild and wacky animation.

### Reproduction Version

80

### Relevant Info

cc @mitc7862 @asangma

geospatialem:
Verified on next

https://user-images.githubusercontent.com/5023024/170574419-3edff617-cf06-4493-bf10-88ecb10fa716.mov

PR #4572, opened 2022-05-17
fix(input-message): remove the scaling transition effect when displaying an input message

Elijbet:
**Related Issue:** #4291

## Summary

The discrepancy between status active=false and status active=true font size was causing an odd scaling effect when transitioning all. There shouldn’t be any font size transitioning for message display, so the nonactive state shouldn’t have a size that is different from the active one. 

https://user-images.githubusercontent.com/19231036/168929796-b2e37cab-e249-4b8c-b38c-e3053767abd9.mov

### DIFF
--- a/src/components/input-message/input-message.scss
+++ b/src/components/input-message/input-message.scss
@@ -50,19 +50,17 @@
 
 // Validation Text
 :host([status][active]) {
-  @apply text-color-1
-    font-medium;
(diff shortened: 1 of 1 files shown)


jcfranco (approved):
LGTM! @macandcheese can you give this one a look? 👀

macandcheese (approved):
Excellent!

Issue #2664, opened 2021-07-28
Enhancement: Format total pagination count


joeyHarig:
### Description
Add commas to the total pagination count when pages exceed 1,000 to make the count more readable.

#### Which Component
`calcite-pagination`

#### Example Use Case
Representing thousands of items in a table view. The page count below should read "10,752".

<img width="1177" alt="Screen Shot 2021-07-28 at 12 10 26 PM" src="https://user-images.githubusercontent.com/55154968/127374471-e950a5bf-809b-4f74-8fba-5683bca11058.png">

driskull:
I think we need to render all pages using Intl number formatting. That should solve this issue. 👍

geospatialem:
Verified on next ✔️ 
![pagination with commas where thousand separators are used](https://user-images.githubusercontent.com/5023024/170135701-3b91c7a7-7025-49c8-b2c7-b95abb3b8d6a.png)

PR #4574, opened 2022-05-18
feat(pagination): format page numbers with localized group seperators

benelan:
**Related Issue:** #2664

## Summary
- Used the `globalAttributes` util to format page numbers based on `lang` (Thanks for the help Franco)
- Added a "Locales" pagination story which defaults to "fr" (which uses a space for group separators)
- Updated the existing stories to have larger page numbers to display commas for the default "en".

The visual regressions related to page numbers that will be in screener are expected.

### DIFF
--- a/src/components/pagination/pagination.tsx
+++ b/src/components/pagination/pagination.tsx
@@ -63,13 +63,6 @@ export class Pagination implements GlobalAttrComponent {
   /** The scale of the pagination */
   @Prop({ reflect: true }) scale: Scale = "m";
 
-  /**
-   * for number values, displays the locale's group separator
-   *
-   * @internal
-   */
-  @Prop() globalAttributes = {};
-
   // --------------------------------------------------------------------------
   //
   //  Private Properties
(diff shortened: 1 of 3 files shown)


src/components/pagination/pagination.tsx
```diff
@@ -183,6 +212,7 @@ export class Pagination {
   }
 
   renderPage(start: number): VNode {
+    const lang = this.globalAttributes["lang"] || document.documentElement.lang || "en";
```
driskull, line 215: I could see the fallback options for the document and "en" being used a lot. I wonder if we can DRY this up so that every component getting "lang" doesn't have to set these fallbacks. What do you think @jcfranco?

driskull (approved):
👍 💯

src/components/pagination/pagination.tsx
```diff
@@ -64,6 +70,15 @@ export class Pagination {
   // --------------------------------------------------------------------------
   @Element() el: HTMLCalcitePaginationElement;
 
+  //--------------------------------------------------------------------------
+  //
+  //  State
+  //
+  //--------------------------------------------------------------------------
+
+  /** Inherited global attributes */
```
jcfranco, line 79: Nitpick: I think we can drop the comment since the interface has this info.

PR #4580, opened 2022-05-18
ci: fix npm start script

benelan:
**Related Issue:** NA

## Summary
`npm run start` was exiting with `1` because it appends flags to the build script, which was now appending it to the patch instead. I switched it to the prexisting `build:watch-dev` script, which is not getting patched. We don't need the es5 patch locally, it's just needed for releases.

Confirmed that the release is not failing with the patch: https://github.com/Esri/calcite-components/commit/120899649fb991a937c5c524304eff87f50ec084

@jcfranco can you confirm that the code is actually patched here:
https://unpkg.com/browse/@esri/calcite-components@1.0.0-next.473/dist/esm-es5/


driskull (approved):
👍

Issue #4577, opened 2022-05-18
non-visible calcite-tooltip sometimes prevents interaction with reference element


mjuniper:
### Actual Behavior

When used in another component, non-visible calcite-tooltips are sometimes positioned in a way that prevents interaction with the reference element.

### Expected Behavior

Tooltips should not prevent interaction with the reference element.

### Reproduction Sample

https://codepen.io/mjuniper/pen/XWZRXqV

### Reproduction Steps

1. visit the codepen
2. hover over one of the fabs below the card
3. observe that the tooltip does not show and the fabs can't be clicked

If you instead visit [this codepen](https://codepen.io/mjuniper/pen/OJQmMaV) it will work as expected. That is because in the version of hub-components that codepen uses, we have added the following css to the component:

```css
calcite-tooltip[aria-hidden="true"] {
  @apply pointer-events-none;
}
```

### Reproduction Version

beta.81

macandcheese:
Here's a visual to see where it's blocking - we should ensure visibility on this is hidden (or more broadly anywhere `calcite-hydrated-hidden` attribute is used.

![Screen Shot 2022-05-18 at 10 32 12 AM](https://user-images.githubusercontent.com/4733155/169105987-28f9f43b-b330-4fec-91d1-fef2b29805a6.png)
)

driskull:
Nice catch. I think we can probably just set `pointer-events:none` on any tooltip/popover that isn't open.

driskull:
We don't want to use `display:hidden` as it will mess with the animations

driskull:
@mjuniper are you sure you're including the calcite stylesheet? That has the following CSS which seems to be missing...

```css
[calcite-hydrated-hidden] {
  visibility: hidden;
  pointer-events: none;
}
```

https://unpkg.com/@esri/calcite-components@latest/dist/calcite/calcite.css

driskull:
Its not a problem on the storybook either. https://esri.github.io/calcite-components/?path=/story/components-tooltip--simple

benelan:
It could also be an issue that only occurs within another custom element

driskull:
I can move that CSS into a SCSS mixn that we can use in every component that needs it if we have to.

driskull:
Yeah I bet the stylesheet would also need to be included in that shadowRoot. So we should probably move this CSS into a mixin so its included inside every component that needs it.

cc @jcfranco.

We should also move the following style:
[...]

PR #4581, opened 2022-05-18
fix(popover, tooltip, alert, tree-item): Make component invisible and have no pointer events by default

driskull:
**Related Issue:** #4577

## Summary

fix(popover, tooltip, alert, tree-item): Make component invisible and have no pointer events by default. (#4577)


jcfranco (approved):
🎉🌮🦃🦖

Issue #4591, opened 2022-05-19
Input type="time" content not aligned properly in Firefox


tyanthony:
### Actual Behavior

When using the `Input` component of type `time`, the content is not aligned properly in Firefox browser

### Expected Behavior

In Firefox, the alignment of the Input type=time component should be aligned properly (centered vertically). This is correct in Chrome.

### Reproduction Sample

https://codepen.io/tyleranthony/pen/PoQmxvy?editors=100

### Reproduction Steps

1. Add a `calcite-input type="time"` component
2. set a time
3. notice the content (time and "clear" X) are not center-aligned. The icon is correct

### Reproduction Version

beta.82

### Relevant Info

Browser: Firefox

geospatialem:
Verified on `next` with Firefox.

![alignment on firefox fix in next](https://user-images.githubusercontent.com/5023024/170559844-a4b6368a-6d23-42e4-af45-3a7b9528fef4.png)

PR #4601, opened 2022-05-20
fix(input): Align type="time" correctly in Firefox

driskull:
**Related Issue:** #4591

## Summary

fix(input): Align type="time" correctly in Firefox. (#4591)


jcfranco (approved):
💥

Issue #4220, opened 2022-03-10
Text alignment inconsistencies in dropdown items


ellenupp:
### Actual Behavior

When a dropdown is used inside a container with `text-align` set, it is inconsistently applied to the dropdown items as well. 

Here, `text-align: center` is set on a div containing the dropdown. The alignment is applied to the dropdown items.
![image](https://user-images.githubusercontent.com/14304581/157571659-a7ac51e4-5a10-42fc-9053-c8ccf067ccd5.png)

However when there is an icon in the dropdown item, it only appears to affect items where the text wraps.
![image](https://user-images.githubusercontent.com/14304581/157571570-cf82e922-b13d-479b-b75b-5b80743646c0.png)

### Expected Behavior

I wasn't expecting the dropdown items to inherit the alignment, but if it does I'd expect it to be consistent. 

### Reproduction Sample

https://codepen.io/ellupp/pen/VwrovvW?editors=1100

### Reproduction Steps

1. See sample

### Reproduction Version

beta.78

macandcheese:
Good catch, we should ensure we are setting text-align explicitly to prevent this.

geospatialem:
Verified on next

![dropdown alignment verification with icons](https://user-images.githubusercontent.com/5023024/170575305-d10310fa-f422-4d4c-8f64-7a6ba61b7f0e.png)

![dropdown alignment verification without icons](https://user-images.githubusercontent.com/5023024/170575334-a1077be5-09cb-460f-98fa-2321fec5a3ab.png)

PR #4615, opened 2022-05-25
fix(dropdown-item, dropdown-group): text align to "start" by default.

driskull:
**Related Issue:** #4220

## Summary

fix(dropdown-item, dropdown-group): text align to "start" by default.


jcfranco (approved):
💥

Issue #4614, opened 2022-05-25
Dropdown is preventing all keys when opened


driskull:
### Actual Behavior

When a dropdown is opened, you can't open the developer console because the component is calling event.preventDefault for all keys.

### Expected Behavior

A dev can open the console and only specific keys are prevented.

### Reproduction Sample

https://codepen.io/driskull/pen/yLvPjKw?editors=1000

### Reproduction Steps

1. Open the dropdown.
2. Try to open the console via keyboard
3. Notice that it won't open because keys are being prevented.

### Reproduction Version

latest

macandcheese:
Agree- Probably just need to prevent esc, enter / return, space default behavior?

PR #4625, opened 2022-05-26
fix(dropdown-item): Only prevent default event for specific keys.

driskull:
**Related Issue:** #4614

## Summary

fix(dropdown-item): Only prevent default event for specific keys.


macandcheese (approved):
Looks good! - seems unrelated to these changes but I noticed the "Link Trigger" and "Div Trigger" examples on the local dev site do not respond to enter / space keyboard interaction, just click.

Issue #4537, opened 2022-05-12
Height of popover heading inconsistent when `dismissible` enabled


nwhittaker:
### Actual Behavior

For non-dismissible `calcite-popover` elements (`dismissible=false`), the height of its heading content-box is `46px`. For dismissible popovers, the height is `56px`.

### Expected Behavior

For dismissible popovers, the height of the heading content-box remains unchanged.

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/popover/

### Reproduction Steps

1. Visit the doc page
2. Configure the example popover with a heading
3. Activate the popover and use dev tools to inspect the height of its heading
4. Configure the example popover as dismissible
5. Reactivate the popover and use dev tools to inspect the height of its heading

### Reproduction Version

1.0.0-beta.82

### Relevant Info

The padding built into the `m` scaled close action is causing it to stretch the height of the heading area. Adjusting to an `s` scaled action appears to correct the immediate issue.

The `calcite-modal` component exhibited a similar behavior which was (mostly?) corrected in https://github.com/Esri/calcite-components/issues/1707.

Elijbet:
@benelan Do we have any new info on what is happening with the height in Windows for https://github.com/Esri/calcite-components/issues/1707 that I could use for this one? I'll assign it to myself since it's related.

benelan:
Unfortunately I haven't had a chance to look into that yet. If you want to take over #1707 I'd be happy to help you with any Windows related testing

benelan:
verified on beta.83

PR #4627, opened 2022-05-26
fix(popover): height is no longer inconsistent when dismissable enabled

Elijbet:
**Related Issue:** #4537

## Summary

Popover requests adding a medium button which is inconsistent with the constant height of the component. Setting action to scale=s doesn’t change the nested icon without specified scale to scale=s, but brings in default m. Setting both to scale s, otherwise, both default scaling to m. Will be opening a separate issue to add scale on popover and will be addressing scale cascading from parent to nested children.

### DIFF
--- a/src/components/popover/popover.tsx
+++ b/src/components/popover/popover.tsx
@@ -456,9 +456,10 @@ export class Popover {
           class={CSS.closeButton}
           onClick={this.hide}
           ref={(closeButtonEl) => (this.closeButtonEl = closeButtonEl)}
+          scale="s"
           text={intlClose}
         >
-          <calcite-icon icon="x" scale="m" />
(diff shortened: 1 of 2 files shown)


Elijbet:
Noticed an inconsistency in scale prop inheritance from action to icon within the button:

Setting action to scale=s doesn’t change the nested icon without specified scale to scale=s, but brings in default m. Setting both to scale s, otherwise, both default scaling to m. 

Added an item to https://github.com/Esri/calcite-components/wiki/Consistency-Audit to check for ensuring consistent scale cascading from parent to nested children.

driskull (approved):
This looks good @Elijbet! Can we add a screenshot test so we make sure it is tested 👍

Issue #4576, opened 2022-05-18
Refactor globalAttributes utility to handle common fallback options


driskull:
### Description

I could see the fallback options for the document and "en" being used a lot. 

example: `const lang = this.globalAttributes["lang"] || document.documentElement.lang || "en";`

### Proposed Advantages

I wonder if we can DRY this up so that every component getting "lang" doesn't have to set these fallbacks.

The goal would be that any component using `this.globalAttributes["lang"]` has the same fallbacks.

### Which Component

All components

### Relevant Info

https://github.com/Esri/calcite-components/pull/4574#discussion_r876069296

jcfranco:
At least for the `lang` case, there's now a [`getLocale` util](https://github.com/Esri/calcite-components/blob/master/src/utils/locale.ts#L298-L311), which DRYs up this issue's use case. Are there any other cases we would need a fallback for?

geospatialem:
No longer relevant, has been addressed in previous efforts.

PR #4657, opened 2022-05-27
docs(internal): globalAttributes util usage

benelan:
**Related Issue:** NA

## Summary

Documented the globalAttributes util

### DIFF
--- a/conventions/README.md
+++ b/conventions/README.md
@@ -278,7 +278,7 @@ There are utilities for common workflows in [`src/utils`](https://github.com/Esr
 
 ### Global attributes
 
-The [`globalAttributes`](https://github.com/Esri/calcite-components/blob/master/src/utils/globalAttributes.ts) util was specifically made to access the `lang` global attribute when set on a Calcite component. However, it can be extended to allow additional global attributes by adding to the [`allowedGlobalAttributes`](https://github.com/Esri/calcite-components/blob/a33aa0df0c5bf103f91187826e6b12b8ff266d90/src/utils/globalAttributes.ts#L4-L5) array. The util is used in [`calcite-pagination`](https://github.com/Esri/calcite-components/blob/master/src/components/pagination/pagination.tsx), which you can use as a reference.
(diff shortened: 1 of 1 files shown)


conventions/README.md
```diff
@@ -271,3 +271,49 @@ focusMenu(): void => {
...
+
+   disconnectedCallback(): void {
+     unwatchGlobalAttributes(this);
+   }
+   ```
+
+5. Use the state to access `lang` (or another global attribute that may be allowed in the future).
+
+   ```js
+   const lang = this.globalAttributes["lang"] || document.documentElement.lang || "en";
```
driskull, line 318: I wonder if we should have a utility for this part: `document.documentElement.lang || "en"`. I think we want consistency here and it may actually be better as the following:

```ts
const lang = this.globalAttributes["lang"] || this.el.closest("lang") || "en"; // may want to do `closestElementCrossShadowBoundary` instead?
```

benelan: I completely agree, you actually said something similar in my [PR](https://github.com/Esri/calcite-components/pull/4574#discussion_r876069296) for implementing the util in pagination. Should I create a refactor issue for it? In the issue I can mention to change that line of the convention depending on the solution. Or if its a quick refactor I can hold off on merging until it happens.

benelan: Oh nvm, you already made a refactor issue #4576

conventions/README.md
```diff
@@ -271,3 +271,49 @@ focusMenu(): void => {
...
 }
 ```
+
+## Utils
+
+There are utilities for common workflows in [`src/utils`](https://github.com/Esri/calcite-components/tree/master/src/utils).
+
+### Global attributes
+
+The [`globalAttributes`](https://github.com/Esri/calcite-components/blob/master/src/utils/globalAttributes.ts) util was specifically made to access the `lang` global attribute when set on a Calcite component. However, it can be extended to allow additional global attributes by adding to the [`allowedGlobalAttributes`](https://github.com/Esri/calcite-components/blob/a33aa0df0c5bf103f91187826e6b12b8ff266d90/src/utils/globalAttributes.ts#L4-L5) array. The util is used in [`calcite-pagination`](https://github.com/Esri/calcite-components/blob/master/src/components/pagination/pagination.tsx), which you can use as a reference.
```
jcfranco, line 281: Maybe we can link to https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes#list_of_global_attributes here for additional context.

jcfranco (approved):
💥

PR #4665, opened 2022-05-31
feat(popover, tooltip): Support virtual elements for positioning.

driskull:
**Related Issue:** #

## Summary

feat(popover, tooltip): Support virtual elements for positioning.

This gives more control over positioning popovers and tooltips by allowing an author to specify where the reference element is located using a `VirtualElement`. 

https://popper.js.org/docs/v2/virtual-elements/#usage

Example:

```js
function generateGetBoundingClientRect() {
        return () => ({
          width: 0,
          height: 0,
          top: 100,
          right: 100,
          bottom: 100,
          left: 600
        });
      }

      const virtualElement = {
        getBoundingClientRect: generateGetBoundingClientRect()
      };

      const popover = document.getElementById("virtual-popover");
      popover.referenceElement = virtualElement;
```


src/demos/popover.html
```diff
@@ -210,5 +211,29 @@ <h1 style="margin: 0 auto; text-align: center">Popover</h1>
         </calcite-tabs>
       </div>
     </div>
+
+    <!-- virtual el popover -->
+    <calcite-popover id="virtual-popover" open label="right start popover" placement="right" id="popover-right" open>
```
jcfranco, line 216: Can this be moved to a usage file? I think it'd be helpful to show the different ways to set up the reference element. This also applies to the tooltip.

We can tackle this in a separate doc issue. cc @benelan

jcfranco (approved):
I really like this. This will be a great improvement!

WRT testing, could you add a test that asserts on valid reference elements? For example, does not show warning for a matched ID, HTML & virtual elements, but does show for an unmatched ID and a plain object or null or similar?

src/demos/popover.html
driskull: I'll move to usage file 👍

driskull:
> could you add a test that asserts on valid reference elements? For example, does not show warning for a matched ID, HTML & virtual elements, but does show for an unmatched ID and a plain object or null or similar?

Maybe this could be done as followup as it kind of goes beyond scope of this issue? The typings should handle these cases.

driskull:
@benelan can you review the doc changes?

jcfranco:
> Maybe this could be done as followup as it kind of goes beyond scope of this issue? The typings should handle these cases.

My suggestion came off wrong, you can ignore it. We need a test for the new behavior, types alone won't cut it. Popover tests already cover both element ID and node references. We need to assert that the virtual element reference is wired up properly in some way.

driskull:
yeah that's tricky. The virtual element we are just passing to popper.js so it might just be testing what they test.

jcfranco:
We're not testing popper's positioning, we're testing that popover/tooltip accepts a virtual reference element and it gets positioned. Similar to popover's [`should accept referenceElement as string id` test](https://github.com/Esri/calcite-components/blob/master/src/components/popover/popover.e2e.ts#L138).

driskull:
Sounds good. I can work on a test for that.

Issue #4609, opened 2022-05-23
Tooltip on calcite-action-menu


yannikmesserli:
### Actual Behavior

If you open a `calcite-action-menu` from a `calcite-action-bar` for which there is a tooltip, then the tooltip shows up for any element of the menu you over.
![2022-05-23 16 43 40](https://user-images.githubusercontent.com/1910449/169845189-098c4bfe-ce94-4266-a476-b28010e1bde8.gif)

### Expected Behavior

The tooltip shows up only when overing the item in the bar, not in the menu.

### Reproduction Sample

https://codepen.io/yannikm/pen/qBxXGQL

### Reproduction Steps

1. Open the menu by clicking on the "..."
2. Move the mouse somewhere else for the tooltip to disappear
3. Move the mouse over one of the menu's item

### Reproduction Version

beta.82

### Relevant Info

mac osx, firefox

driskull:
@yannikmesserli there's a slot on the action-menu for adding a tooltip for this reason. https://github.com/Esri/calcite-components/blob/master/src/components/action-menu/readme.md

yannikmesserli:
Ok, went through the code to find out that [the slot is actually](https://github.com/Esri/calcite-components/blob/master/src/components/action-menu/action-menu.tsx#L339) `calcite-tooltip`. However, this does not work:
[https://codepen.io/yannikm/pen/qBxXGQL](https://codepen.io/yannikm/pen/qBxXGQL) - the tooltip simply never shows up.

Alternatively, I [tried to create a calcite tooltip](https://codepen.io/yannikm/pen/NWyyVxM) with the slot, but it exhibits the same problem as above.

driskull:
@yannikmesserli here is the correct code: https://codepen.io/driskull/pen/PoQRLxv?editors=1000

yannikmesserli:
Ok, then there is still two problems:
- tooltip does not disappear when menu opens, preventing the user to select the sub menu items
- when closing the sub menu (aka clicking on sub items) the tooltip unexpectedly appears:

![2022-06-01 06 38 08](https://user-images.githubusercontent.com/1910449/171328915-9071c821-6af7-4767-b888-78da59308636.gif)

driskull:
> tooltip does not disappear when menu opens, preventing the user to select the sub menu items

I'll fix this one.

> when closing the sub menu (aka clicking on sub items) the tooltip unexpectedly appears:

I'll see if I can get this one fixed too.

PR #4668, opened 2022-06-01
fix(action-menu): Hide tooltip when open. (#4609)

driskull:
**Related Issue:** #4609

## Summary

fix(action-menu): Hide tooltip when open. (#4609)


jcfranco (approved):
💥

Can you add a test for this before merging?

Issue #4667, opened 2022-06-01
calcite-panel: when using the jsapi directions widget within a panel contained within a flow, the panel gets marked hidden


kevindoshier:
### Actual Behavior

When I add the jsapi directions widget to a calcite-panel. where the calcite-panel is within a calcite-flow, once the directions widget is loaded, the calcite-panel containing the widget gets marked hidden

### Expected Behavior

The calcite-panel should remain visible

### Reproduction Sample

https://codepen.io/kevindoshier/pen/wvyjyvJ

### Reproduction Steps

1. Click the codepen link
2. Observe after signing in using arcgis.com credentials the panel is hidden

### Reproduction Version

beta.82

driskull:
@jcfranco I think we should update the flow component to only use slotted `calcite-panel`s. The change here would be that if a `calcite-panel` isn't a direct child of the parent `flow` then it will no longer work. IMO this is an intended fix and avoids using `calcite-panel`s that may be further down the DOM tree such as with this case where there is a `calcite-panel` used in the directions widget.

benelan:
verified on beta.83

PR #4670, opened 2022-06-01
fix(flow): Only slotted panels should be managed. (#4667)

driskull:
**Related Issue:** #4667

## Summary

fix(flow): Only slotted panels should be managed. (#4667)


jcfranco (approved):
Awesome! 🎉

Can you add a test before merging? It'll make it clearer that only direct panel children will be managed.

Issue #4578, opened 2022-05-18
Number input incorrect rounding behavior for long decimals


ethanbdev:
### Actual Behavior

For the number input, there seems to be a maximum allowable number of decimal places, but the behaviour seems inconsistent depending on the starting number and the decimal value. In some cases it actually replaces what the user would have typed or adds an extra number to the end.

You can continuously enter decimal values <6 without any weird rounding occurring.

### Expected Behavior

It should not round decimals unless the developer specifies the behavior. It should additionally not be adding numbers to the end of the decimal or replacing user input with random numbers in long decimals

### Reproduction Sample

https://codepen.io/eborgen/pen/YzeVwdb?editors=1000

### Reproduction Steps

The codepen has 3 examples, they are labeled in the pen:

1. Continuously adding a decimal value of 9 eventually converts the value to the next largest whole number
2. Continuously adding decimal values >5 eventually causes the digit to increase by 1 for a bit, before continuing to allow that number again
3. Setting a value with many decimal values >5 causes an extra digit to be appended

### Reproduction Version

beta.82

### Regression?

unknown

benelan:
Looks like this is happening when we convert the string into a number because 
```js
Number("4.9999999999999999") === 5 // true
````

amring1:
This doesn't just occur with decimal values. Typing 9999999999 eventually turns into 10000000000000000

benelan:
@ethanbdev I have a PR for this, is there a specific decimal precision you need for your use case? I currently have it at 100 decmial places.

amring1:
> @ethanbdev I have a PR for this, is there a specific decimal precision you need for your use case? I currently have it at 100 decmial places.

I think that should be plenty for us.

geospatialem:
Verified on `next` ✔️ 

![verified decimal fix](https://user-images.githubusercontent.com/5023024/173392100-a00e9c32-effb-4987-9d10-5bd2f7003f6f.png)

PR #4679, opened 2022-06-03
fix(input): support larger than 64-bit floating point numbers without losing precision

benelan:
**Related Issue:** #4578

## Summary

- Added a BigDecimal number util for arbitrary precision arithmetic. I modified a solution that uses BigInt, which I found here
  - https://stackoverflow.com/a/66939244
 
- I had to change the tsconfig target to "es2019" which is when BigInt was introduced. I used `npm link` to test the React and Angular examples locally with the [es5 build](https://developers.arcgis.com/calcite-design-system/get-started/#distribution). Everything is working properly in the latest versions of Chrome, Firefox, and Brave (also chromium).

- I was able to implement a `formatToParts()` method for BigDecimal since the `NumberFormat` prototype method accepts BigInt as a param type. 
  - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/formatToParts#parameters

- I had to make an additional change so that negative fractional numbers (e.g. `-0.001`) do not become positive when typing. `BigInt("-0").toString() === "0"` which ends up stripping the minus sign.


src/utils/number.ts
```diff
@@ -1,4 +1,97 @@
...
+// adopted BigInt solution from https://stackoverflow.com/a/66939244
+export class BigDecimal {
+  value: bigint;
+
+  // BigInt("0") === BigInt("-0") which strips the minus sign when typing numbers like -0.1
+  // https://forum.kirupa.com/t/js-tip-of-the-day-negative-zero/643094
+  isNegative: boolean;
+
+  // Configuration: constants
+  static DECIMALS = 20; // number of decimals on all instances
```
benelan, line 19: How many decimals do we want to allow before rounding? It can be 50 or more if we really want

src/utils/number.ts
```diff
@@ -1,4 +1,98 @@
...
+
+    const s = this.value
+      .toString()
+      .replace(new RegExp("-", "g"), "")
+      .padStart(BigDecimal.DECIMALS + 1, "0");
+
+    const i = s.slice(0, -BigDecimal.DECIMALS);
+    const d = s.slice(-BigDecimal.DECIMALS).replace(/\.?0+$/, "");
+
+    const parts = formatter.formatToParts(BigInt(i));
```
eriklharper, line 69: I'm wondering if this line ☝️ could read like this, that way the next line could be removed:
```
const parts = formatter.formatToParts(BigInt(this.isNegative ? "-" + i : i));
```

src/utils/number.spec.ts
```diff
@@ -97,3 +98,32 @@ describe("sanitizeNumberString", () => {
...
   });
 });
+
+describe("BigDecimal", () => {
+  it("handles precise/large numbers and arbitrary-precision arithmetic", () => {
+    const subtract = new BigDecimal("0.3").subtract("0.1").toString();
+    expect(subtract).toBe("0.2");
+
+    const add = new BigDecimal(Number.MAX_SAFE_INTEGER.toString()).add("1").toString();
+    expect(add).not.toBe(Number.MAX_SAFE_INTEGER.toString());
```
eriklharper, line 108: Could you add an additional assertion that the resulting sum is 1 higher than the original value too?

src/utils/number.ts
eriklharper: Yeah, 100 sounds fine, might as well max it out as large as we think would accommodate all use cases.

src/utils/number.ts
benelan: I tried that, it would cause the minus sign to be removed when typing negative franctional numbers like `-0.1`. There is no such thing as `-0` for `BigInt` like there is for `number`. https://forum.kirupa.com/t/js-tip-of-the-day-negative-zero/643094

eriklharper (approved):
Other than a couple of suggestions, this looks good to me.  Primo work!

Issue #4631, opened 2022-05-27
calcite-tooltip: line breaks added with bold text


AdelheidF:
### Actual Behavior

If the tooltip string contains a word with <b> tags it puts that word on its own line

### Expected Behavior

no line breaks for bold word

### Reproduction Sample

https://codepen.io/afreitag/pen/KKQZmXP

### Reproduction Steps

Open and click on link

### Reproduction Version

beta.82

driskull:
Looks like a CSS issue with the popover's internal container. Its displaying as a flex column.

driskull:
@asangma or @macandcheese can you advise on this one. I'm not sure if flex-col is needed here: https://github.com/Esri/calcite-components/blob/master/src/components/tooltip/tooltip.scss#L10

Its making `<span>` inside the tooltip display as block. Do we need the flex-col?

benelan:
verified on next

PR #4684, opened 2022-06-03
fix(tooltip): Remove internal flex display. (#4631)

driskull:
**Related Issue:** #4631

## Summary

fix(tooltip): Remove internal flex display. (#4631)

The flex display is making `<span />` placed inside the tooltip act like a block element.

Demo can be seen here: https://codepen.io/afreitag/pen/KKQZmXP


Elijbet (approved):
Looks right to me!

benelan (approved):
👍

PR #4692, opened 2022-06-06
ci: add action to move issues to current milestone

benelan:
**Related Issue:** NA

## Summary
Created an action that moves issues from the previous, past due milestone to the current one. Only moves issues with one of these three labels:
- `0 - new`
- `1 - assigned`
- `2 - in development`

The action runs on Mondays. It won't do anything for the week between sprints since all of the movable issues were already moved. It automatically adds the `milestone adjusted` label to the issues it moves.


driskull (approved):
LGTM!

Issue #4689, opened 2022-06-05
Add 'before' events for events that fire after a transition


jcfranco:
### Description

We have some events that are emitted after a transition finishes (e.g., closing modal). Adding an event that fires before the transition kicks off would be a useful hook for developers. 

### Acceptance Criteria

1. `calcite<Component>Before<Event>` events are added for transition-chained events.
2. The conventions guidelines should be updated to describe this pattern

### Which Component

* `calcite-alert`
* `calcite-combobox`
* `calcite-dropdown`
* `calcite-input-date-picker`
* `calcite-modal` 
* `calcite-popover`

Each of the following has an open/close event.

### Example Use Case

```js
alert.addEventListener("calciteAlertBeforeOpen", () => { /* do something on the app or component side that needs to happen before the DOM is updated */});
alert.addEventListener("calciteAlertOpen", () => { /* do something with the updated DOM */});
```

benelan:
This is looking really good, nice work Eliza!

The only issue I found is that `calciteModalOpen` and `calciteModalBeforeOpen` do not emit on page load like the rest of the components. However they do emit correctly when you close and reopen the modal. Here is a codepen with all of the affected components open on page load: 
https://codepen.io/benesri/pen/VwXXJBZ?editors=1111

I'm assigning this back to dev to get `calciteModalOpen` and `calciteModalBeforeOpen` to emit on page load when relevant, such as the codepen above.

Elijbet:
> This is looking really good, nice work Eliza!
> 
> The only issue I found is that `calciteModalOpen` and `calciteModalBeforeOpen` do not emit on page load like the rest of the components. However they do emit correctly when you close and reopen the modal. Here is a codepen with all of the affected components open on page load: https://codepen.io/benesri/pen/VwXXJBZ?editors=1111
> 
> I'm assigning this back to dev to get `calciteModalOpen` and `calciteModalBeforeOpen` to emit on page load when relevant, such as the codepen above.

Thanks Ben, I'll take care of that.

geospatialem:
Verified in beta.93

PR #4693, opened 2022-06-07
refactor(alert): added 'before' events for events that fire after transition

Elijbet:
**Related Issue:** #4689

## Summary

Provide devs with more control of the components through additional hooks. For public events that are emitted after a transition finishes (in this case, opening/closing of the alert), we're now adding events that fire before the transition kicks off as well.

To accomplish this I used the `transitionrun` event, which is marked as experimental on MDN.

- `transitionrun` fires when the transition is created at the start of any delay and is not cancellable once started.
-  if there is no transition delay, both `transitionrun` and `transitionstart` are fired at the same time.


src/components/alert/alert.tsx
```diff
@@ -193,10 +202,16 @@ export class Alert {
   //
   //--------------------------------------------------------------------------
 
-  /** Fired when an alert is closed */
+  /* Fired when an alert is going to be closed and before the transition starts. */
+  @Event() calciteAlertBeforeClose: EventEmitter;
```
macandcheese, line 206: With these added names, I find it hard to determine which is which. In the close instance - does beforeClose mean "after user has requested to close the alert"?

Does "close" mean - has been closed and animation is complete, or the close is about to occur? Maybe we can be more explicit with the naming for those.

src/components/alert/alert.tsx
```diff
@@ -318,22 +333,54 @@ export class Alert {
...
-  /** emit the opened alert and the queue */
+  /* * 
+  - `transitionrun` fires when the transition is created (i.e. at the start of any delay).
+  - `transitionstart` fires when the actual animation has begun (i.e. at the end of any delay).
+  - `transitionrun` will occur even if the transition is canceled before the delay expires. 
+  - if there is no transition delay or if `transition-delay` is negative, both `transitionrun` and `transitionstart` are fired.
+  */
+  private transitionRunEvent = (): void => {
+    document
+      .querySelector("calcite-alert")
```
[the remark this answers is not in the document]
eriklharper: ☝️ In this instance you're going to want to add the event listener on `this.el` instead of `document.querySelector("calcite-alert")`.

src/components/alert/alert.tsx
```diff
@@ -126,8 +126,17 @@ export class Alert {
     this.requestedIcon = setRequestedIcon(StatusIcons, this.icon, this.color);
   }
 
+  componentDidLoad(): void {
+    this.transitionRunEvent();
+  }
```
eriklharper, line 131: This will only wire up the event listener once in the page lifecycle so if the element is disconnected from the dom and relocated, the listener will be destroyed.  You probably want this to be inside `connectedCallback` instead.

src/components/alert/alert.tsx
```diff
@@ -318,22 +333,54 @@ export class Alert {
   };
 
   transitionEnd = (event: TransitionEvent): void => {
+    if (event.propertyName === this.activeTransitionProp) {
+      this.active ? this.openCloseEventEmitter("open") : this.openCloseEventEmitter("close");
+    }
+  };
+
+  transitionRun = (event: TransitionEvent): void => {
```
eriklharper, line 341: In place of `this.transitionRunEvent()` you should be able to just decorate this function with `@Listen("transitionrun")` unless the `transitionrun` event doesn't bubble up to the Host element.  Does that work?

eriklharper: Would be nice to rename that function to `onTransitionRun` to give it more semantic meaning too.

Elijbet: When I use @Listen, it looks like the transitionRun bubbles but is not composed. According to MDN, all UA-dispatched UI events are composed (click/touch/mouseover/copy/paste, etc.). Most other types of events are not composed. This one looks like it bubbles, but is not composed.

src/components/alert/alert.tsx
Elijbet: @macandcheese does this make it more clear?

src/components/alert/alert.e2e.ts
```diff
@@ -269,25 +270,31 @@ describe("calcite-alert", () => {
...
 
     expect(await container.isVisible()).toBe(true);
 
+    const beforeCloseEvent = page.waitForEvent("calciteAlertBeforeClose");
     const closeEvent = page.waitForEvent("calciteAlertClose");
+
     element.removeAttribute("active");
     await page.waitForChanges();
+    await beforeCloseEvent;
     await closeEvent;
```
eriklharper, line 298: Could we add some event spy assertions in this test as well?  It would be cool to spy on all 4 component state events (beforeOpen, open, beforeClosing, closed) and make sure that they each only fire once per interaction.

Elijbet (commented):
To sum up our discussion with Erik on this at this point: 

There was one thing I had doubts about. When there is no delay on animation, transitionRun = transitionStart. I recorded to see in slow motion what that looks like, 
and basically, calciteAlertBeforeOpening will fire at the very beginning of the transition, which means it's sort of already faintly visible. This might be a problem depending on what the hook is used for. 

The event is not cancellable, to begin with, so it wouldn't be an event a dev would use to cancel opening the alert or something similar, even if it had a delay. This kind of implementation would be fine for side effects that a dev might want to respond to, but not control the active state of the alert (eg. display a notification badge in a different part of the UI before the alert closes/opens).

Erik suggested that if we want to expose a custom event that a dev can use to cancel opening or activating the alert, the better thing to do would probably be to respond in the `active` Watcher or create a @Method on the alert component that can be used to toggle activate.

Since the issue originated from research I did to compare handling event emission around transitions in other components, we decided I’d look at the use cases for `before` events, to see if we want to implement the second approach on top of the first, or just use one of those.

eriklharper (approved):
Thanks for adding those additional event assertions!  This is looking good :)

driskull:
My comments:
I'm not sure about the naming of the events. They should be clear and consistent for each property and when they are going to be fired.

I was expecting something like `calciteOpenBefore` and `calciteOpenAfter` / `calciteCloseBefore` and `calciteCloseAfter`. 

It doesn't look like we're using `transitionstart` right? Just `transitionrun`. That makes sense, just want to confirm since the `transitionstart` is mentioned.

driskull:
If we rename `calciteAlertClose` then it would be a breaking change.

We want to try to avoid that so if we do need the name changes, lets deprecate `calciteAlertClose` and add the new event so that this PR doesn't need to have breaking changes.

Elijbet:
> If we rename `calciteAlertClose` then it would be a breaking change.
> 
> We want to try to avoid that so if we do need the name changes, lets deprecate `calciteAlertClose` and add the new event so that this PR doesn't need to have breaking changes.
> 
> Same with `calciteAlertOpen`

Np, I can revert the change.

Elijbet:
I've pulled together **some use cases for `before` events**

**before opening:**

- For dialogs, the event occurs immediately before the menu is displayed, allowing to fetch/load profile information.
- For context menus and submenus, the event occurs immediately before the menu is displayed, allowing menu items to be checked, disabled, etc., according to the current context. 
- For selection box controls, the event occurs immediately before the selection box list is opened, allowing dynamic re-initialization of the list on demand. 
- The element might ensure that it's visible.
- May display alert message before element appears.
- Display a notification badge in a different part of the UI.

**before closing:**

`before` event is fired before this element is being closed upon a user gesture. Cancellation of this event stops the user-initiated action of closing this modal. It May be useful for:
- Running validations.
- Displaying alert message before element disappears.
- Display a notification badge in a different part of the UI.

**before other events:**

**before tab selection**
- Validate the current tab when the user selects another tab/prevent user interaction with the active tab if not passing.
- Give warning before tab changes (“You have unsaved changes”)

Elijbet:
before opening:
For dialogs, the event occurs immediately before the menu is displayed, allowing to fetch/load profile information.
For context menus and submenus, the event occurs immediately before the menu is displayed, allowing menu items to be checked, disabled, etc., according to the current context. 
For selection box controls, the event occurs immediately before the selection box list is opened, allowing dynamic (re-)initialisation of the list on demand. 
The element might ensure that it's visible.
May display alert message before element appears.
Display a notification badge in a different part of the UI.
before closing:
`before` event is fired before this element is being closed upon a user gesture. Cancellation of this event stops the user-initiated action of closing this modal. May be useful for:
Running validations.
Displaying alert message before element disappears.
Display a notification badge in a different part of the UI.
before other events:
tab selection
Validate the current tab when the user selects another tab/prevent user interaction with the active tab if not passing.
Give warning before tab changes (“You have unsaved changes”)

Elijbet:
I pulled some use cases for `before` events:

**before opening:**

- For dialogs, the event occurs immediately before the menu is displayed, allowing to fetch/load profile information.
- For context menus and submenus, the event occurs immediately before the menu is displayed, allowing menu items to be checked, disabled, etc., according to the current context. 
- For selection box controls, the event occurs immediately before the selection box list is opened, allowing dynamic reinitialization of the list on demand. 
- The element might ensure that it's visible.
- May display alert message before element appears.
- Display a notification badge in a different part of the UI.

**before closing:**

- `before` event is fired before this element is being closed upon a user gesture. Cancellation of this event stops the user-initiated action of closing this modal. It May be useful for:
- Running validations.
- Displaying alert message before element disappears.
- Display a notification badge in a different part of the UI.

**before other events:
tab selection**

- Validate the current tab when the user selects another tab/prevent user interaction with the active tab if not passing.
- Give warning before tab changes (“You have unsaved changes”)

src/components/alert/alert.tsx
```diff
@@ -270,6 +278,13 @@ export class Alert {
   /** is the alert queued */
   @State() queued = false;
 
+  private containerDiv: HTMLDivElement;
+
+  private setContainerDiv = (el): void => {
+    this.containerDiv = el;
+    this.containerDiv.addEventListener("transitionrun", this.onTransitionRun);
```
driskull, line 285: @Elijbet does `onTransitionrun` not work on the JSX element?

eriklharper: I think it would if Typescript didn't fail the build.  We couldn't figure out how to get Typescript to not complain about a property it doesn't know about on the div element.  If you know of a way to override types in JSX that might work, but I suggested this method to get around that because I couldn't figure out a way to get past that TS compile issue with the JSX method.

eriklharper: Perhaps an upgrade to the types package might also resolve it?  Not sure...

driskull: thanks! sounds good.

macandcheese (approved):

src/components/alert/alert.tsx
```diff
@@ -128,6 +128,7 @@ export class Alert {
 
   disconnectedCallback(): void {
     window.clearTimeout(this.autoDismissTimeoutId);
+    this.containerDiv.removeEventListener("transitionrun", this.onTransitionRun);
```
driskull, line 131: Maybe check to see if containerDiv is defined first. I think there's a chance it might get disconnected before first render.

driskull (approved):
👍 👏 💯

PR #4694, opened 2022-06-08
fix(label): expose bottom margin value and remove the disable-spacing property

eriklharper:
**Related Issue:** #2138

This fix exposes label's default bottom margin as a css variable and removes the `disable-spacing` property while preserving scaled spacing between the label text and any other sibling elements in either inline or block layout through the use of a `flex-gap` which only applies the space between sibling elements when they are present.


eriklharper:
> I think we need to provide a way to maintain current spacing.

I feel like this is a userland problem, 
> 
> I know we had talked about either/both of:
> 
> * eventually having a label-group, fieldset, etc component that handled spacing of children for vertical rhythm within forms (in addition to other form related behavior)
> * offering the exported TW spacing utilities to let devs provide this spacing themselves.
> 
> Neither of those currently exist, so this will be a large change.
> 
> In the meantime, could we implement the suggestion in the issue to have a css variable that (for now at least) is set by default to the margin bottom removed here?

We could do that, but

macandcheese:
Thanks for adding the css variable! Can we keep the values the same as the current ones for now (for each scale)? Other than that this looks good to me.

eriklharper:
> Thanks for adding the css variable! Can we keep the values the same as the current ones for now (for each scale)? Other than that this looks good to me.

Good catch!  Didn't realize those got out of sync.  Will update that.

macandcheese (approved):
Thanks for adding!

jcfranco:
@eriklharper Could you update the screenshot tests to cover this? More importantly, to avoid the breaking change, could you look into restoring the prop and marking it as deprecated? Correct me if I'm wrong, but I think this could work by keeping the previous styles but making them higher in specificity. In short, existing use cases of `disable-spacing` would still work and would require users to remove it for the custom CSS prop to be honored.

eriklharper:
> @eriklharper Could you update the screenshot tests to cover this? More importantly, to avoid the breaking change, could you look into restoring the prop and marking it as deprecated? Correct me if I'm wrong, but I think this could work by keeping the previous styles but making them higher in specificity. In short, existing use cases of `disable-spacing` would still work and would require users to remove it for the custom CSS prop to be honored.

Here's a follow-up PR implementing the suggestions you lay out here.  I like it! https://github.com/Esri/calcite-components/pull/4889

Issue #2115, opened 2021-05-07
Refactor: consistent boolean names


jcfranco:
Stems from #1145 

## Summary of changes

Booleans must adhere to the following for consistency:

* prefer adjectives (e.g., collapsible, moveable, etc), if an appropriate one does not exist, `-disabled`/`-enabled` suffixes can be used to describe the prop.
* avoid initializing booleans to true

This came up in https://github.com/Esri/calcite-components/pull/2841#discussion_r698597822. The goal for this refactor is have boolean names be expressive and for similar ones to be named consistently. We should be concise whenever possible as long as it is clear how the name will affect the component. A prefix/suffix pattern may help, but may not be ideal in all cases. We should to explore other libs for inspiration before moving forward with renames.

### Relevant Info 
* [x] replace `constant` prop in `combo-box-item` to use something like `disableFilter` which is self explanatory.

geospatialem:
Verified in beta.86 with the `filter-disabled` attribute and in the [docs](https://developers.arcgis.com/calcite-design-system/components/combobox-item/#component-api-properties-filterDisabled).

PR #4709, opened 2022-06-09
refactor(combobox-item): enable filter-disabled prop support 

anveshmekala:
**Related Issue:** #2115

## Summary

 enable` filter-disabled` prop support for `combobox-item` which is equivalent of `constant` prop.  deprecation of `constant` prop will follow.


driskull (approved):
👍

PR #4710, opened 2022-06-10
refactor(combobox): added 'before' events for events that fire after transition

Elijbet:
Related Issue: https://github.com/Esri/calcite-components/issues/4689

Provide devs with more control of the components through additional hooks. For public events that are emitted after a transition finishes (in this case, opening/closing of the combobox), we're now adding events that fire before the transition kicks off as well.


Elijbet:
Should I extract the `private openCloseEventEmitter` method into util?

driskull (approved):
> Should I extract the private openCloseEventEmitter method into util?

Yeah, great idea. More DRY that way.

👍

PR #4711, opened 2022-06-10
fix(tooltip): Fix syntax typo in `TooltipManager`

nwhittaker:
**Related Issue:** #

## Summary
It was preventing the `focusin` handler from being removed in `removeListeners()` because the `capture` flag was obscured when first adding the handler.

Looking to get this fix in quickly to unblock a Calcite upgrade to Field Maps ahead of the EA release next week. Is it worth a follow-on issue to add test coverage for this scenario?


eriklharper (approved):
Wow, thanks for catching that!

PR #4714, opened 2022-06-10
fix(tabs): update style for centered and bordered tab titles

macandcheese:
**Related Issue:** https://github.com/Esri/calcite-components/issues/3682

## Summary
Adjusts styling for long titles in bordered mode, and fix for cut off content in center alignment. 

- Long titles should now display properly in "bordered" mode.
Screen Shot 2022-06-10 at 10 35 19 AM
Screen Shot 2022-06-10 at 10 34 22 AM

- Center alignment should no longer cut off names 
  - if the container is wide enough to display all items, it will appear center-aligned.
  - if the container is not wide enough, the first tab will remain visible.
Screen Shot 2022-06-10 at 10 36 45 AM
Screen Shot 2022-06-10 at 10 36 33 AM

### DIFF
--- a/src/components/tab-nav/tab-nav.scss
+++ b/src/components/tab-nav/tab-nav.scss
@@ -21,10 +21,6 @@
     overflow-auto;
 }
 
-:host([layout="center"]) .tab-nav {
-  @apply justify-center;
-}
-
 // prevent indicator overflow in horizontal scrolling situations
 .tab-nav-active-indicator-container {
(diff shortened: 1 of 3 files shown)


macandcheese:
Note: The component doesn't currently support bordered + centered combination - this doesn't change that, but maybe it's something to look at in the future, it may be unexpected that those combined props do not display the intended design.

Issue #4702, opened 2022-06-09
[Action] Add css var for overriding indicator color


macandcheese:
### Description

Currently, the indicator is using the brand variable. This allows it to be overridden, but it also means that if you do so, the focus outline of the action is updated, which may be undesirable.

### Acceptance Criteria

Suggest adding a local-to-action css var along the lines of "action-indicator-color" to allow this customization. It should not leverage the global icon override color as the user may want the icon and the indicator to be of different colors.

### Which Component

Action

### Example Use Case

The action indicator may have different meanings - the current solution for overriding this also affects the focus outline, which isn't desired.

macandcheese:
This is released - https://codepen.io/mac_and_cheese/pen/oNqqRQM?editors=100

Do we need to keep this open until the change is logged for Figma now? cc @benelan @SkyeSeitz

SkyeSeitz:
Change has been logged in Figma. I can start providing a Figma update when I get tagged in PRs if that is helpful.

jcfranco:
Closing as self-verified. cc @geospatialem

PR #4717, opened 2022-06-10
feat(action): Add css variable to override indicator color

macandcheese:
**Related Issue:** #4702 

## Summary
Adds a css variable to optionally set the indicator color on the action.


driskull (approved):
👍

PR #4720, opened 2022-06-10
ci: stop auto closing bug reports with missing repro samples

benelan:
**Related Issue:** 

## Summary
On a trial basis, bug reports with missing repro samples will not automatically close after 5 days. People understandably find it offputting, and I can see how it might prevent users from contributing. It was a great way to get everyone to start providing samples, and now that we built that habit I think we can tone it back a bit. If there is a notable decrease in bug reports with repro samples I will change it back.


eriklharper (approved):
LGTM 👍

PR #4725, opened 2022-06-13
build(deps): bump packages that are usually excluded

benelan:
**Related Issue:** NA

## Summary

@jcfranco do you remember why jest packages and puppeteer were excluded from the `deps:update` npm script? I was able to bump them up a few versions (not to latest) when I was testing Storyshots as a local alternative to screener.

e2e and screener are passing. `npm start` and `npm run docs:preview` also work.


driskull (approved):
👍

PR #4727, opened 2022-06-13
test(tooltip, popover): Add tests for virtual reference elements

driskull:
**Related Issue:** #3196

## Summary

test(tooltip, popover): Add tests for virtual reference elements. (#3196)


Elijbet (approved):
Neat! Didn't know you could use a virtual reference element to make popper/tooltip follow the mouse cursor.

Issue #4688, opened 2022-06-05
API consistency refactor


jcfranco:
### Description

This issue aims to refactor API members for consistency. 

Individual tasks should be derived from https://github.com/Esri/calcite-components/wiki/Consistency-Audit. Boolean consistency tasks can be moved to https://github.com/Esri/calcite-components/issues/2115 if needed to split up.

The goal is to deprecate as much as possible in favor of consistent names/patterns. 

### Proposed Advantages

Consistent props, methods, event names, patterns.

### Which Component

Any component that matches audit findings.

### Relevant Info

* [x]  deprecate `'clear' | 'default'` values for `appearance` prop in the components applicable. Accepted values are `'transparent'| 'outline' | 'minimal' | 'solid'`. The prop should have no default property value.
* [x] use `open` prop for components that are not visible by default, and then appear when the prop is true. e.g. in `modal`, `popover` & `dropdown`
* [x] Use `expanded` if the component is already visible, and then grows in size when the prop is true. e.g. in `accordion-item`
* [x] use `closed` prop if the component is initially visible, and then disapears when the prop is true. Deprecate `dismissed` and `dismissible` (replace `dismissible` with `closable`). Only add `closable` if the component already has a `dismissible` prop.
* [x] replace usage of `active` prop with `open`, `expanded`, or `closed` based on the definitions above. Make sure to deprecate the `active` prop when replacing it. Ideally the `active` prop should be used to indicate focus behavior, e.g. in  `tab-title` and `action`.
* [x] deprecate `itemTitle` prop in favor of having just one prop `heading`. 
* [x] deprecate `summary `, `itemSubTitle` prop and use `description` when needed.
* [x] replace `above| below` value for `position` prop in `Tab` component with `top| bottom`. 
* [x] use `iconStart` and `iconEnd` props instead of `iconPosition` and `icon`.  Look into accordion and radio-group to transition them away from `iconPosition`.
* [x] Add `closed` prop to chip component which enables user's to hide and display chip irrespective of `closable` prop [4836](https://github.com/Esri/calcite-components/pull/4836)
* [x] Replace `disabled<Property>` with `<property>Disabled`. https://github.com/Esri/calcite-components/pull/5644#discussion_r1029836755

#### Update :

- Accordion will use existing `iconType` and `iconPosition` props to place pre-define indicator icons.
-  `action-*` components will use the `expanded` prop as the trigger element is expanding the screen space rather than
[...]

PR #4729, opened 2022-06-13
feat(tabs,tab-nav,tab-title): add top,bottom values for position prop

anveshrmekala:
**Related Issue:** #4688 

## Summary

This will enable `top`, `bottom` values for `@Prop( ) position` in `tabs`,`tab-nav` & `tab-title` component. current values `above , below` will be deprecated.

### DIFF
--- a/src/components/tab-nav/tab-nav.scss
+++ b/src/components/tab-nav/tab-nav.scss
@@ -45,12 +45,12 @@
     ease-out;
 }
 
-:host([position="below"]) .tab-nav-active-indicator {
+:host([position="below"] [position="bottom"]) .tab-nav-active-indicator {
   bottom: unset;
   @apply top-0;
 }
 
-:host([position="below"]) .tab-nav-active-indicator-container {
+:host([position="below"] [position="bottom"]) .tab-nav-active-indicator-container {
   bottom: unset;
   @apply top-0;
 }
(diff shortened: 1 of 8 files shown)


src/components/tabs/tabs.tsx
```diff
@@ -33,7 +33,8 @@ export class Tabs {
   @Prop({ reflect: true }) layout: TabLayout = "inline";
 
   /**
-   * Display the tabs above (default) or below the tab content
+   * Display the tabs top (default) or bottom of the tab content. above and below are deprecated.
+   *
    */
   @Prop({ reflect: true }) position: TabPosition = "above";
```
driskull, line 39: If above and below are deprecated should we set the default to be "top"?

driskull: I wonder if we should use the @deprecated tag here to note the deprecation?

src/components/tab-nav/tab-nav.scss
```diff
@@ -50,11 +50,21 @@
   @apply top-0;
 }
 
+:host([position="bottom"]) .tab-nav-active-indicator {
+  bottom: unset;
```
driskull, line 54: Do you know if we could construct the CSS so we don't have to use "unset"?

driskull (approved):
👍

src/components/tabs/tabs.tsx
anveshmekala: if we use the @deprecated tag here, it will be applied to the prop rather than values.

anveshmekala: Agree that we need to make `top` the default here for tabs.

src/components/tab-nav/tab-nav.scss
anveshmekala: as discussed in teams, the bottom needs to be unset only in two scenarios and in all other cases it is required to have `bottom-0`.

PR #4734, opened 2022-06-14
fix(color-picker): fixed hue bar click and drag interaction

Elijbet:
**Related Issue:** #

Prevent the selector from snapping, while using a mouse to click and drag past the component boundary. 

Because of the radial nature of HSV representation, the end value of colorFieldWidth is basically the same as 0. Subtracting one from the value prevents going it full circle.


macandcheese:
Looks good! I couldn't replicate the horizontal scrollbar on the screener while testing locally.

macandcheese (approved):
🌈

driskull (approved):
👍

Issue #4731, opened 2022-06-14
component method call error in vue


webmapLee:
### Actual Behavior

when I call components method in vue, throw a error:
![image](https://user-images.githubusercontent.com/20334500/173588366-6aaa90f1-de61-4e6f-b0c7-657634642c3b.png)

my component code:
```typescript
import { defineComponent, onMounted, ref } from 'vue'

export default defineComponent({
  name: 'DemoPage',
  setup() {
    const actionRef = ref()
    const msgRef = ref('hello')
    onMounted(() => {
      console.log('🍎🍎🍎🍎', actionRef.value.setFocus)
      actionRef.value.setFocus.call(actionRef.value)
    })
    setTimeout(() => {
      msgRef.value = 'world'
    }, 2000)
    return () => {
      const textClass = {
        'text-3xl': true,
        'text-orange-700': true,
        'font-bold': true,
        underline: true,
      }
      return (
        <>
          <div>
            <calcite-shell>
              <h1 slot="title" class={textClass} id="header-title">
                Hello World
              </h1>
              <calcite-action
                ref={actionRef}
                text={msgRef.value}
                text-enabled
                alignment="start"
                appearance="solid"
                intl-loading="aa"
                scale="m"
              ></calcite-action>
              <calcite-shell-panel
                slot="contextual-panel"
                position="end"
                detached
                height-scale="l"
              >
                <calcite-action-bar slot="action-bar">
                  <calcite-action-group>
                    <calcite-action
                      text="Add"
                      active
                      icon="plus"
                    ></calcite-action>
                    <calcite-action
                      text="Save"
                      disabled
                      icon="save"
                    ></calcite-action>
                    <calcite-action
                      text="Layers"
                      icon="layers"
                    ></calcite-action>
                  </calcite-action-group>
                  <calcite-action-group>
                    <calcite-action text="Add" icon="plus"></calcite-action>
                    <calcite-action
                      text="Save"
                      disabled
                      icon="save"
                    ></calcite-action>
                    <calcite-action
                      text="Layers"
                      icon="layers"
                    ></calcite-action>
                  </calcite-action-group>
                  <calcite-action
                    slot="bottom-actions"
                    text="Tips"
                    icon="lightbulb"
                  ></calcite-action>
                </calcite-action-bar>
              </calcite-shell-panel>
              <div id="viewDiv"></div>
            </calcite-shell>
          </div>
        </>
      )
    }
  },
})

```

### Expected Behavior

method successfully called

[...]

PR #4739, opened 2022-06-15
fix: Make sure component setFocus methods do not throw error

driskull:
**Related Issue:** #4731

## Summary

fix: Make sure component setFocus methods do not throw error. (#4731)

When calling setFocus before the first render, an error may be thrown. This prevents that.

### DIFF
--- a/src/components/color-picker/color-picker.tsx
+++ b/src/components/color-picker/color-picker.tsx
@@ -556,7 +556,7 @@ export class ColorPicker implements InteractiveComponent {
     } else if (region === "slider") {
       this.sliderThumbState = "drag";
       this.captureHueSliderColor(offsetX);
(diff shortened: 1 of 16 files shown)


src/components/color-picker/color-picker.tsx
```diff
@@ -556,7 +556,7 @@ export class ColorPicker implements InteractiveComponent {
     } else if (region === "slider") {
```
anveshmekala, line 556: this one might return error right ?

anveshmekala (commented):
LGTM! 

This [one ](https://github.com/Esri/calcite-components/blob/master/src/components/color-picker/color-picker.tsx#L555) might return error  in color-picker.

eriklharper (approved):

Issue #4613, opened 2022-05-25
Modal has `firstFocus` property documented but not implemented


noahmulfinger:
### Description

The `firstFocus` property is listed in the [readme docs](https://github.com/Esri/calcite-components/tree/master/src/components/modal#properties) and [as a prop in modal.tsx](https://github.com/Esri/calcite-components/blob/master/src/components/modal/modal.tsx#L88), but it is not used anywhere. Seems like it should either be removed or the functionality should be implemented.

It's also documented in the developers website: https://developers.arcgis.com/calcite-design-system/components/modal/#component-api-properties-firstFocus

### Which Component

calcite-modal

jcfranco:
@noahmulfinger Great catch. 

@benelan @geospatialem @driskull we should probably drop this prop. ✋🎤

driskull:
drop it! Agreeed

benelan:
Verified on `beta.83`

PR #4742, opened 2022-06-15
test(modal): remove test for firstFocus property

benelan:
**Related Issue:** #4613

## Summary
Removing the e2e test for the firstFocus property that was removed in #4671


driskull (approved):
👍

PR #4744, opened 2022-06-15
feat(accordion, action, button, chip, color-picker, split-button): deprecate clear & default appearance types

anveshmekala:
**Related Issue:** #4688 

## Summary

-  deprecated `default`  values for `@Prop( ) appearance` and replaced them with `solid` type appearance. 
-  deprecated `clear` value and replace it with `transparent` and for a component which has `transparent` value listed prior to the change will use `minimal`. 

### Effected components 

- accordion
- action
- button
- chip
- color-picker
- split-button


src/components/action/action.tsx
```diff
@@ -44,8 +44,8 @@ export class Action implements InteractiveComponent {
    */
   @Prop({ reflect: true }) alignment?: Alignment;
 
-  /** Specify the appearance style of the action, defaults to solid. */
-  @Prop({ reflect: true }) appearance: Extract<"solid" | "clear", Appearance> = "solid";
+  /** Specify the appearance style of the action. */
+  @Prop({ reflect: true }) appearance: Extract<"solid" | "clear" | "transparent", Appearance>;
```
driskull, line 48: @benelan @jcfranco I think we need a plan on how we can communicate deprecation of values.

driskull: @anveshmekala do we want to keep "solid" as the default value?

Lets hold off on this one till @jcfranco is back.

driskull (commented):
Lets wait on comments from @jcfranco before proceeding here.

Code changes look good but I think we need a plan on how we can deprecate specific values and if we want to keep "solid" instead of "default".

src/components/action/action.tsx
benelan: Deprecation of values should definitely show up in the changelog. We can always make a PR to add additional info to changelog entries (after the corresponding `next` release when the changelog is updated) if there isn't enough space in the PR title. Eventually we want to have changelog-ish page on the developer doc site like the api has release pages.

anveshmekala: `solid` as default is more meaningful than `default` but it's unnecessary IMO because the expectation of the user is to render the `action` as a whole .  If there is a difference in UX or visual aspect of component when `appearance=solid` vs `appearance=''` then having the `solid`  is understandable.

driskull: I agree with that, but I think there are a lot of instances where we have the value explicitly set when it is something like "solid" and not assumed. I think we need to maybe set it as "solid" here. cc @jcfranco

jcfranco: Let's keep assigning the default value. If the prop wasn't reflected, I'd be OK not assigning to simplify, shed some bits (minimally) and lower selector specificity. Another issue is that the reflected attributes would read as booleans for the unassigned default cases (e.g., `<calcite-button appearance ...>`).

On the plus side, I think we can still keep the simplified CSS (from this PR) by not using the defaults in the selectors. 

> do we want to keep "solid" as the default value?

Great question. I think we should either:

1. always use `default` as the default value
2. change `default` to something more meaningful

src/components/color-picker/color-picker.e2e.ts
```diff
@@ -64,10 +64,6 @@ describe("calcite-color-picker", () => {
         propertyName: "allowEmpty",
         defaultValue: false
       },
-      {
-        propertyName: "appearance",
```
jcfranco, line 68: This won't matter after my recommendation to keep assigning the defaults, but the `color-picker` was not included in the scope portion of the PR title.

jcfranco:
WRT the commit type, I think this should be a feature that deprecates some values. Reasoning: refactors should not affect public APIs.

src/components/action/action.scss
```diff
@@ -145,27 +145,36 @@
   @apply bg-foreground-1;
 }
 
-:host([appearance="clear"]) .button {
+//clear is deprecated. use transparent
```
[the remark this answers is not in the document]
anveshmekala: A deprecation note for the props was added this is a note for the devs to not apply styles specific to the deprecated prop selector.

driskull:
> change default to something more meaningful

I like this idea better. The string value "default" is meaningless. "solid" is more meaningful.

src/components/action/action.tsx
driskull: > Another issue is that the reflected attributes would read as booleans for the unassigned default cases 

Great point. So we need a default value. We just need to change it from "default" to something more meaningful.

anveshmekala:
Are we going to stick with `solid` as appearance type and replace `default` with `solid` wherever required? 

cc : @jcfranco @macandcheese @driskull

jcfranco:
@anveshmekala @driskull Yeah, let's move forward w/ descriptive values.

src/components/chip/chip.scss
```diff
@@ -127,31 +127,31 @@
 }
 
 // solid
-:host([color="blue"]) {
+:host([color="blue"][appearance="solid"]) {
```
jcfranco, line 130: I think these are missing the host + color attribute selector.

src/components/button/button.scss
```diff
@@ -230,15 +230,15 @@
 
 // button styles
 // solid
-:host {
+:host([appearance="solid"]) {
```
jcfranco, line 233: To simplify our styles, let's not add selectors for default styles. No attribute here should mean `appearance=solid`.

This applies to any selectors explicitly using defaults.

jcfranco (changes requested):

jcfranco (approved):
Code LGTM! 🎉 

@macandcheese can you give this a quick look? 👀

macandcheese:
Looks good, screener changes seemed unrelated (tooltip positioning) 👍

AdelheidF:
> button - deprecated clear value and replace it with transparent

This doesn't seem true to me.

beta.86 (minimal was not supported)
image

beta.93
image

I would have expected with 93 transparent having an outline and minimal having none.

Issue #4743, opened 2022-06-15
Regression setting stepper-item's `active` property


nwhittaker:
### Actual Behavior

Setting a stepper-item's `active` property to `true` does not swap in the step's contents.

### Expected Behavior

Setting a stepper-item's `active` property to `true` swaps in the step's contents.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/WNMmbbo

### Reproduction Steps

1. Visit the code sample and notice the last stepper item is active.
2. Click the **Prev step** button and notice the previous step's header becomes active, but the stepper's body content remains unchanged.

### Reproduction Version

beta.82, beta.83

### Relevant Info

Feels similar to https://github.com/Esri/calcite-components/issues/4567.

### Regression?

beta.81

benelan:
@driskull can you take a look at this issue? Looks like a regression from #4490

driskull:
Yep. i'll check it out.

geospatialem:
Verified on next

https://user-images.githubusercontent.com/5023024/175561017-87b3f519-4b0b-4971-97a1-1e322f5d34c0.mp4

PR #4745, opened 2022-06-16
fix(stepper): setting active property should update content

driskull:
**Related Issue:** #4743

## Summary

fix(stepper): setting active property should update content. (#4743)


eriklharper (approved):
LGTM!

Issue #4766, opened 2022-06-16
[Stepper] "Grid" is not valid value for layout property


macandcheese:
### Actual Behavior

Layout property has a valid value of "grid" - but that is not applicable for Stepper.

I think when the local types were updated to be sourced from the new shared interface, we never excluded "grid". 

We should extract just "horizontal" and "vertical" from https://github.com/Esri/calcite-components/blob/6847f57359897972b002bb5f9377fb4c6c3d5a4f/src/components/interfaces.ts#L6

### Expected Behavior

"grid" is not a valid value of "layout".

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/YzegLQm?editors=100

### Reproduction Steps

Codepen.

### Reproduction Version

83

geospatialem:
Verified on beta.90

PR #4767, opened 2022-06-16
fix(stepper): remove invalid layout value (#4766)

macandcheese:
Updates typing of layout prop in stepper and stepper-item

**Related Issue:** #4766 

## Summary
This extracts just the values that are relevant to Stepper from the shared Layout interface.

Note - we should probably move away from styling based on reflected props in a future update.


driskull (approved):
👍

PR #4769, opened 2022-06-16
ci(dependabot): prevent bumping cheerio and chalk

benelan:
**Related Issue:** #4763

## Summary

chalk/cheerio were already excluded from the `update-deps` npm script but not dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,6 +15,8 @@ updates:
       prefix: "build(deps): "
     # Ignore updates to packages below
     ignore:
+      - dependency-name: "chalk"
+      - dependency-name: "cheerio"
       - dependency-name: "typescript"
       - dependency-name: "@types/jest"
       - dependency-name: "jest"


driskull (approved):
👍

driskull:
Do we still need this script? it seems like duplication to me. 

https://github.com/Esri/calcite-components/blob/master/package.json#L23

benelan:
Yeah it is a duplication but there's a reason lol. The npm script runs `npm test` locally and then bumps one package, runs `npm test`, and then goes to the next one if it passes. That works for the most part now that we don't have as many unstable tests. 

However, sometimes we get peer dep errors on the script because it bumps a single storybook dep at a time and some of them need to be updated in unison. Might be the same for jest, I'm not sure.

So best case scenario is everything gets bumped by the script so we aren't spammed with PRs from dependabot (and to keep our screener runs lower). But I set up dependabot as a backup for those situations.

So we can remove the npm script and the associated action, but that's why we currently have both. We can't remove dependabot for the reason above.

benelan:
> However, sometimes we get peer dep errors on the script because it bumps a single storybook dep at a time and some of them need to be updated in unison. Might be the same for jest, I'm not sure.

Now that I type this out, I'm curious why we don't get peer deps errors in our ci when dependabot bumps a single storybook dependency at a time.

PR #4774, opened 2022-06-17
ci: notify leads of new component label

geospatialem:
**Related Issue:** N/A

## Summary
Adds automation  to notify CC design lead(s) when the `new component` label is added to an issue. 🤖

### Features
* The automation adds text to the issue's body `cc @<design-username>`  
* The design lead(s) are stored in a GH secret `CALCITE_DESIGNERS` 🔒 
    * The entries in the secret will be entered as `<design-username>` or `<design-username1>, <design-username2>, ....` 
    * The automation automatically adds in the `@` before the `<design-username>` text.

### DIFF
--- a/.github/workflows/new-component-notify.yml
+++ b/.github/workflows/new-component-notify.yml
@@ -1,30 +1,30 @@
-# If the "new component" label is added to an issue, generates a notification to the CC team lead(s)
-# The secret is formatted like so: lead1, lead2, lead3
-# Note the script automatically adds the "@" character in to notify the leads
-name: "Notify CC Leads of New Component Issue"
+# If the "new component" label is added to an issue, generates a notification to the Calcite designer lead(s)
+# The secret is formatted like so: designer1, designer2, designer3
+# Note the script automatically adds the "@" character in to notify the designer lead(s)
+name: "Notify CC Designer Leads of New Component Issue"
 
 on:
   issues:
     types: [labeled]
 
 jobs:
-  notify-lead-new-component:
+  notify-designer-new-component:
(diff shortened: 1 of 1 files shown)


.github/workflows/new-component-notify.yml
```diff
@@ -0,0 +1,30 @@
...
+            const { designers } = process.env;
+            // Add a "@" character to notify the user
+            const calcite_designers = designers.split(",").map(v => " @" + v.trim());
+
+            // Add a comment to issues with the 'new component' label to notify the designer(s)
+            await github.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: `cc ${calcite_designers}`,
```
benelan, line 29: Won't the comment end up being 
> cc [ @benelan, @geospatialem, ...]

Maybe you could change line 22 to 

```js
 const calcite_designers = designers.split(",").map(v => "@" + v.trim()).join(" ");
```

geospatialem: Got it working as-is in a local private repo so it looks like `cc @benelan` or `cc @benelan, geospatialem`. If it looks cleaner code-wise, can modify it to your recommendation.

benelan: > Got it working as-is in a local private repo so it looks like `cc @benelan` or `cc @benelan, @geospatialem`. If it looks cleaner code-wise, can modify it to your recommendation.

Idk why I thought the square brackets would be there lol. The two spaces after cc isn't really a big deal so LGTM

benelan (approved):

PR #4775, opened 2022-06-20
feat(accordion-item,combobox,dropdown,modal): deprecate active prop for open in collapsible components

anveshmekala:
**Related Issue:** #4688 

## Summary

- Deprecate `active` prop and replace it with  `open` prop for collapsible component .
- Deprecate ` active` prop and replace it with `expanded` prop for `accordion-item` component.

### List of Components with Change

- accordion-item
- combobox
- dropdown
- modal


driskull (commented):
@anveshmekala can you check storybook and HTML demo files as well to make sure they are updated? We might need to go back to any PRs already installed with the renaming deprecations and make sure they are updated there too.

driskull (approved):
LGTM!

PR #4789, opened 2022-06-21
feat(chip,notice,panel,popover): deprecate dismissible & dismissed props and add support for closable & closed props

anveshmekala:
**Related Issue:** #4688 

## Summary

- Deprecate `dismissible` and `dismissable` props across the components
- Components using `dismissible` will use `closable` prop instead  
- Components using `dismissied` will use `closed` prop instead  

### List of Components with Change:

- chip
- notice
- panel
- popover


driskull:
@anveshmekala should we also fix `chip` so that when it is `closed` it no longer renders any UI?

driskull (commented):
I think we should probably update `chip` to no longer render any UI if closed.

If not part of this PR, a followup issue/pr should be linked

anveshmekala:
> I think we should probably update `chip` to no longer render any UI if closed.
> 
> If not part of this PR, a followup issue/pr should be linked

agreed. we need to remove the chip el from DOM when on close.

driskull:
@anveshmekala looks like @kevindoshier found a bug with this one.

https://codepen.io/kevindoshier/pen/abYJXZZ?editors=1000

When the panel is closed by a user it sets both `dismissed` and `closed` to `true`. So a user has to set both to `false` to get it back. Ideally, a user would just have to set `closed` to `false`.

src/components/panel/panel.tsx
```diff
@@ -241,6 +252,7 @@ export class Panel implements ConditionalSlotComponent, InteractiveComponent {
 
   dismiss = (): void => {
     this.dismissed = true;
```
driskull, line 254: I think `this.dismissed = true` needs to be removed here.

Issue #4116, opened 2022-02-15
Add `no-padding` property to `block`


TheBlueDog:
### Description

There are many places where we use calcite-block inside of leading and trailing panels. It's not always desirable to have the built in padding. When using other components, such as lists, inside a block the vertical rhythm and alignment is thrown off.

Modal has a property for `no-padding` that work well in order to remove the padding. The default setting could be `true` as to not break current implementations. Setting it to `false` would remove the paddings.

### Acceptance Criteria

- [ ] calcite-block has a new property added for `no-padding`
- [ ] Setting options are `true` and `false`
- [ ] Default setting of `false` maintains current behavior
- [ ] Setting of `true` removes the padding from all four sides 

### Relevant Info

This component is in use in Field Maps today and the enhancement would be utilized as soon as it is implemented.

### Which Component

Calcite-Block

### Example Use Case

Default layout with spacing:

![Screen Shot 2022-02-15 at 1 58 37 PM](https://user-images.githubusercontent.com/5419306/154134213-a3eedf08-51af-4526-90b2-3359d66c5a67.png)

Layout with spacing removed:

![Screen Shot 2022-02-15 at 1 59 25 PM](https://user-images.githubusercontent.com/5419306/154134311-171ee308-b213-4e49-b488-0bedee5fb595.png)

macandcheese:
I agree with this use case - however I don't think it should be "True by default" - disabling the padding here would be a breaking visual change so having it as opt-in behavior works.

cc @asangma - we may also want to look at adding a reciprocal pattern to panel to provide a consistent spacing when needed.

TheBlueDog:
@macandcheese maintaining current behavior as the default is what I meant. The description has been updated to reflect that. Good catch.

macandcheese:
cc @asangma - any issues with this? @benelan - can we re-triage with accurate priority for an upcoming sprint?

macandcheese:
Bump @benelan @jcfranco - seems like a simple one to add - would be great for some Dev site components where I have to set this manually.

geospatialem:
Verified on `next`

### `noPadding` / `no-padding` is `false` (default)
![no padding default verified](https://user-images.githubusercontent.com/5023024/175563668-dda71d3a-bad9-4e02-b0c6-14180abd5e6b.png)

### `noPadding` / `no-padding` is `true`
[...]

Issue #3860, opened 2022-01-04
deprecate `disablePadding` , `noPadding` in favor of CSS variables


asangma:
### Question
Label currently has a `disableSpacing` property which "eliminates any space around the label".

```
:host([disable-spacing]) {
  .container {
    @apply gap-0 m-0;
  }
}
```

There has been some call to introduce this approach to other components. Should we? Or is this a broader issue?

Some points from a discussion are:
* Components shouldn't have margins at the :host or .container level.
  *  Label currently does this.
* Spacing is core part of a UI's rhythm and readability.
  * It's also important to retain spacing relative to scale.
* If our spacing is causing problems with our own stuff, we should probably fix that.

Some have suggested that a systematic solution for spacing might be the better solution, e.g. a spacing component, utility classes, etc.

Something else to consider is that when "known components" are children, inter-component spacing should be considered as helpful, e.g. **ListItem** inside a **List** or **Action** inside **ActionBar**.

### Update 

- As discussed in teams channel, we will deprecate `disablePadding` , `disableSpacing`  & `noPadding` props across the repo and replace them with CSS vars instead
- Any new requests for customizing spacing will be addressed based on user feedback.

driskull:
> Something else to consider is that when "known components" are children, inter-component spacing should be considered as helpful, e.g. ListItem inside a List or Action inside ActionBar.

Yes, we should space children but not parent components. In this case, it's probably best if the parent handles the spacing and the children don't have spacing on their own.

macandcheese:
I agree that components should not "space themselves". 

I think `calcite-label` is a bit of an outlier because it is so often a part of a larger group of multiple `calcite-labels` . 

If we had a `calcite-label-group` type component, we could offload the spacing responsibility to that parent, and continue to have consistent rhythm in forms, while keeping a single `calcite-label` free of any applied margins.

asangma:
> it's probably best if the parent handles the spacing and the children don't have spacing on their own.

Agreed.

driskull:
> If we had a calcite-label-group type component, we could offload the spacing responsibility to that parent, and continue to have consistent rhythm in forms, while keeping a single calcite-label free of any applied margins.

I think that is warranted :) Some type of group component.

asangma:
The pattern for "[something]-group" sounds good to me.

And I agree about not wanting to do `calcite-form` unless it had the implied logic.

driskull:
At some point we might need a `fieldset` equivalent component.

[...]

PR #4790, opened 2022-06-22
feat(block): support noPadding property

anveshmekala:
**Related Issue:** #4116 

## Summary

This will add a new prop `noPadding` which is defaulted to `false`. When `true`, it removes padding of the content slotted in `calcite-block`

### DIFF
--- a/src/components/block/block.stories.ts
+++ b/src/components/block/block.stories.ts
@@ -176,7 +176,7 @@ export const disabled = (): string => html`<calcite-block heading="heading" summ
 </calcite-block>`;
 
 export const noPadding = (): string => html` <calcite-panel heading="Properties">
-  <calcite-block heading="Example block heading" summary="example summary heading" collapsible open>
+  <calcite-block heading="Example block heading" summary="example summary heading" collapsible open no-padding>
     <div>calcite components ninja</div>
   </calcite-block>
 </calcite-panel>`;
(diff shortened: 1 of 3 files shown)


src/components/block/block.stories.ts
```diff
@@ -174,3 +174,9 @@ export const disabled = (): string => html`<calcite-block heading="heading" summ
     <img alt="demo" src="${placeholderImage({ width: 320, height: 240 })}" />
   </calcite-block-section>
 </calcite-block>`;
+
+export const noPadding = (): string => html` <calcite-panel heading="Properties">
+  <calcite-block heading="Example block heading" summary="example summary heading" collapsible open>
```
eriklharper, line 179: ☝️ shouldn't this have the `no-padding` prop applied?

anveshmekala: It should. Good catch.

eriklharper (approved):

driskull:
@anveshmekala @eriklharper @jcfranco can we discuss this one? I thought we were eliminating properties like this? Similar to https://github.com/Esri/calcite-components/issues/3860

This is similar to the discussion here...

> Components should probably not position themselves.

https://www.gabe.pizza/notes-on-component-libraries/

We may want to revert this one in the meantime?

eriklharper:
> @anveshmekala @eriklharper @jcfranco can we discuss this one? I thought we were eliminating properties like this? Similar to #3860
> 
> This is similar to the discussion here...
> 
> > Components should probably not position themselves.
> 
> https://www.gabe.pizza/notes-on-component-libraries/
> 
> We may want to revert this one in the meantime?

I do recall this discussion, thanks for finding that original issue, I had trouble locating that one for some reason.  We probably should just use a css variable instead.

PR #4791, opened 2022-06-22
refactor(combobox): create openCloseComponent interface in global utils that defines methods for beforeOpen, open, beforeClose and close public events

Elijbet:
**Related Issue:** #4689, #4710

## Summary

Create an `OpenCloseComponent` interface in global utils that defines methods for `beforeOpen`, `open`, `beforeClose`, and `close`  event handlers. Combobox implements the interface to emit these public events at `transitionRun` and `transitionEnd.`


src/components/combobox/combobox.tsx
```diff
@@ -386,6 +389,17 @@ export class Combobox implements LabelableComponent, FormComponent, InteractiveC
...
+    beforeOpen: () => this.calciteComboboxBeforeOpen.emit(this.publicEmitPayload),
+    open: () => this.calciteComboboxOpen.emit(this.publicEmitPayload),
+    beforeClose: () => this.calciteComboboxBeforeClose.emit(this.publicEmitPayload),
+    close: () => this.calciteComboboxClose.emit(this.publicEmitPayload)
+  };
+
+  private publicEmitPayload: any = {
+    element: this
+  };
+
```
Elijbet, line 402: I split the function up to understand which parts are shared to be extracted and the payload seems to vary (eg. `combobox` only emits `el`, but `alert` also emits a `queue`). 

Are we expecting components to vary on `OpenCloseState` as well (eg. would some components maybe have `close` but not the `open` event emitted)?

Elijbet, line 402: I split the function up to understand which parts are shared to be extracted and the payload seems to vary (eg. `combobox` only emits `el`, but `alert` also emits a `queue`).

Are we expecting components to vary on `OpenCloseState` as well (eg. would some components maybe have `close` but not the `open` event emitted)?

src/utils/emitter.ts
```diff
@@ -0,0 +1,16 @@
+export interface OpenCloseState {
```
driskull, line 1: @Elijbet Instead of exporting the interface, I was thinking that the `OpenCloseComponent` interface would have a set of required functions like this:

```tsx
export interface OpenCloseComponent {
  /**
   * The host element.
   */
  readonly el: HTMLElement;

 beforeOpen: () => void;
  open: () => void;
  beforeClose: () => void;
  close: () => void;
}
```

What do you think?

Otherwise, we aren't really forcing the developer to implement the interface as we want because the only requirement right now is the `readonly el: HTMLElement;`

src/components/combobox/combobox.tsx
```diff
@@ -1002,6 +1000,13 @@ export class Combobox implements LabelableComponent, FormComponent, InteractiveC
     this.setInactiveIfNotContained(event);
   };
 
+  private openCloseEventEmitter(componentVisibilityState: string): void {
```
driskull, line 1003: Does this private method still need to be here?

src/components/combobox/combobox.tsx
```diff
@@ -386,6 +389,17 @@ export class Combobox implements LabelableComponent, FormComponent, InteractiveC
...
   private activeTransitionProp = "opacity";
 
+  private OpenCloseState: OpenCloseState = {
+    beforeOpen: () => this.calciteComboboxBeforeOpen.emit(this.publicEmitPayload),
+    open: () => this.calciteComboboxOpen.emit(this.publicEmitPayload),
+    beforeClose: () => this.calciteComboboxBeforeClose.emit(this.publicEmitPayload),
+    close: () => this.calciteComboboxClose.emit(this.publicEmitPayload)
+  };
+
+  private publicEmitPayload: any = {
```
driskull, line 399: Can we type this instead of `any`? Maybe this is a `CustomEvent`?

driskull (commented):
Looking much better!!! Nice work @Elijbet.

src/components/combobox/combobox.tsx
driskull: I don't think we should make these be the same across components. It may depend but from what I see, we really don't need to emit anything. The `target/currentTarget` on the event should be the `el` and the user can get the open state from that.

src/utils/emitter.ts
Elijbet: That makes sense. Thanks for the explanation.

src/components/combobox/combobox.tsx
Elijbet: I feel like there was a PR asking for standardization in emitting a payload which involved ensuring there is always one emitted. But I can't seem to find it. I'll remove it from here and maybe add a discussion topic.

src/utils/emitter.ts
```diff
@@ -0,0 +1,14 @@
+/**
```
driskull, line 1: Should we maybe rename this file to `OpenCloseComponent`?

If Franco approves of this class, lets also add some doc for the methods `onOpen/onBeforeOpen...`

driskull (approved):
🎉 🌮

src/utils/openCloseComponent.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Defines interface for components with open/close public emitter.
+ */
+export interface OpenCloseComponent {
+  /**
+   * The host element.
+   */
+  readonly el: HTMLElement;
+
+  onBeforeOpen: () => void;
```
driskull, line 10: Looks good. Just needs some doc for these methods other devs should implement.

driskull (approved):
✨ 💥 ✨

jcfranco (approved):
Noice! I dig this pattern. Way cleaner. 🏆

For a separate PR, can you work w/ @benelan to doc this pattern in our dev doc?

Elijbet:
> Noice! I dig this pattern. Way cleaner. 🏆
> 
> 
> 
> For a separate PR, can you work w/ @benelan to doc this pattern in our dev doc?

Yep. Sounds good. There is one more change I'll be making per @driskull suggestions. I'll update as soon as it's in.

PR #4797, opened 2022-06-23
refactor(dropdown): added 'before' events for events that fire after transition

Elijbet:
**Related Issue:** #4689

## Summary

Provide devs with more control of the components through additional hooks. For public events that are emitted after a transition finishes (in this case, opening/closing of the dropdown), we are now adding events that fire before the transition kicks off as well.

### DIFF
--- a/src/components/dropdown/dropdown.e2e.ts
+++ b/src/components/dropdown/dropdown.e2e.ts
@@ -540,9 +540,8 @@ describe("calcite-dropdown", () => {
 
     it("control max items displayed", async () => {
       const maxItems = 7;
-      const page = await newE2EPage();
-      await page.setContent(html`
-        <calcite-dropdown max-items="${maxItems}">
+      const page = await newE2EPage({
+        html: html`<calcite-dropdown max-items="${maxItems}">
           <calcite-button slot="dropdown-trigger">Open Dropdown</calcite-button>
           <calcite-dropdown-group group-title="First group">
             <calcite-dropdown-item id="item-1">1</calcite-dropdown-item>
@@ -558,8 +557,8 @@ describe("calcite-dropdown", () => {
             <calcite-dropdown-item id="item-9">9</calcite-dropdown-item>
             <calcite-dropdown-item id="item-10">10</calcite-dropdown-item>
           </calcite-dropdown-group>
-        </calcite-dropdown>
-      `);
+        </calcite-dropdown>`
+      });
 
       const element = await page.find("calcite-dropdown");
       const dropdownOpenEvent = page.waitForEvent("calciteDropdownOpen");
--- a/src/components/dropdown/dropdown.e2e.ts
+++ b/src/components/dropdown/dropdown.e2e.ts
@@ -1024,4 +1024,56 @@ describe("calcite-dropdown", () => {
         shadowPopperSelector: ".calcite-dropdown-wrapper"
       }
     ));
+
+  it("should emit component status for transition-chained events: 'calciteDropdownBeforeOpen', 'calciteDropdownOpen', 'calciteDropdownBeforeClose', 'calciteDropdownClose'", async () => {
+    const page = await newE2EPage();
+    await page.setContent(html`
+      <calcite-dropdown>
+        <calcite-button slot="dropdown-trigger">Open dropdown</calcite-button>
+        <calcite-dropdown-group id="group-1">
+          <calcite-dropdown-item id="item-1"> Dropdown Item Content </calcite-dropdown-item>
+          <calcite-dropdown-item id="item-2" active> Dropdown Item Content </calcite-dropdown-item>
+          <calcite-dropdown-item id="item-3"> Dropdown Item Content </calcite-dropdown-item>
+        </calcite-dropdown-group>
+      </calcite-dropdown>
+    `);
+    const element = await page.find("calcite-dropdown");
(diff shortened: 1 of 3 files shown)


src/components/dropdown/dropdown.tsx
```diff
@@ -439,14 +448,39 @@ export class Dropdown implements InteractiveComponent {
...
     }
   };
 
+  onTransitionRun = (event: TransitionEvent): void => {
+    if (event.propertyName === this.activeTransitionProp) {
+      this.active
+        ? this.openCloseEventEmitter("beforeOpen")
+        : this.openCloseEventEmitter("beforeClose");
+    }
+  };
```
eriklharper, line 466: I would recommend renaming this to `transitionRunHandler` to be more explicit that it is a handler function as opposed to an event.

src/components/dropdown/dropdown.tsx
```diff
@@ -439,14 +448,39 @@ export class Dropdown implements InteractiveComponent {
...
+      beforeOpen: () => this.calciteDropdownBeforeOpen.emit(payload),
+      open: () => this.calciteDropdownOpen.emit(payload),
+      beforeClose: () => this.calciteDropdownBeforeClose.emit(payload),
+      close: () => this.calciteDropdownClose.emit(payload)
+    };
+    (
+      emitComponentState[componentVisibilityState] ||
+      emitComponentState["The component state is unknown."]
+    )();
+  }
```
eriklharper, line 482: In a similar vein, this function would be more appropriately titled `emitOpenCloseEvent` since it emits events, but is not actually an `EventEmitter` object.

src/components/dropdown/dropdown.e2e.ts
```diff
@@ -539,8 +540,9 @@ describe("calcite-dropdown", () => {
 
     it("control max items displayed", async () => {
       const maxItems = 7;
-      const page = await newE2EPage({
-        html: html`<calcite-dropdown max-items="${maxItems}">
+      const page = await newE2EPage();
+      await page.setContent(html`
+        <calcite-dropdown max-items="${maxItems}">
```
eriklharper, line 545: This doesn't look related to the changes in this PR, but more a refactor of how the page content is set.  Can you revert this back to the way it was since the shorthand method is preferred?

src/components/dropdown/dropdown.e2e.ts
```diff
@@ -1024,4 +1026,56 @@ describe("calcite-dropdown", () => {
...
+    await page.setContent(html`
+      <calcite-dropdown>
+        <calcite-button slot="dropdown-trigger">Open dropdown</calcite-button>
+        <calcite-dropdown-group id="group-1">
+          <calcite-dropdown-item id="item-1"> Dropdown Item Content </calcite-dropdown-item>
+          <calcite-dropdown-item id="item-2" active> Dropdown Item Content </calcite-dropdown-item>
+          <calcite-dropdown-item id="item-3"> Dropdown Item Content </calcite-dropdown-item>
+        </calcite-dropdown-group>
+      </calcite-dropdown>
+    `);
```
eriklharper, line 1041: Similarly here, you can simplify this with `await newE2EPage({ html: <html-goes-here> })`.

Elijbet: I believe setting page through `page.setContent` was @jcfranco's preference, as he corrected it a couple of times. From what I understand as opposed to `await newE2EPage()`, `page.setContent()` should have `await page.waitForChanges()` built in. I might be wrong though.

eriklharper: Lol, I recall Franco doing the exact opposite on some of my PRs where he recommended setting it the way I described above.  I honestly don't have a preference either way, just was following what I know I've encountered on reviews in the past.

Elijbet: Ok :) Maybe we should convene at some point and pick one :).

Elijbet: I put it on the agenda for the next time to agree on a preferred method of setting the page content.

src/components/dropdown/dropdown.e2e.ts
```diff
@@ -1024,4 +1026,56 @@ describe("calcite-dropdown", () => {
...
+    const element = await page.find(`calcite-dropdown`);
+    const group = await page.find(`calcite-dropdown >>> .${CSS.calciteDropdownContent}`);
+
+    expect(await group.isVisible()).toBe(false);
+
+    const calciteDropdownBeforeOpenEvent = page.waitForEvent("calciteDropdownBeforeOpen");
+    const calciteDropdownOpenEvent = page.waitForEvent("calciteDropdownOpen");
```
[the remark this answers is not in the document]
Elijbet: It might be the transition timing, but when I don't intentionally `waitForEvent`, I get an error that `expected event "calciteDropdownOpen" to have been called 1 times, but was called 0 time`.

anveshmekala (approved):
:rocket:

src/components/dropdown/dropdown.e2e.ts
eriklharper: Fine to leave as-is for now, shouldn't block merging.

eriklharper (approved):
🎉

Issue #4442, opened 2022-04-22
Define style guide and proof read component API reference


benelan:
### Current Style Guide
So far we have the following conventions for API ref descriptions:

- Use complete sentences with proper grammar, capitalization, and punctuation.  
- No abbreviations, e.g. use "property" instead of "prop".  
- Verbs must be in present tense.  
- When referring to the component that the property/slot/method/css variable belongs to, call it "the component".  
- Use the full tag name when referencing other Calcite Components (prefix with `calcite-`), e.g. `calcite-button` instead of `button`.  
- For plural context, use `calcite-button`s instead of `calcite-button` elements.  
- Use backticks (``` ` ```) for the names of slots, events, properties, CSS variables, and component names.  
    - In line with the JSAPI docs [example](https://developers.arcgis.com/javascript/latest/api-reference/esri-views-View.html#ready)  
    - Docs will incorporate styling with [afd-calcite-documentation issue #490](https://github.com/ArcGIS/afd-calcite-documentation/issues/490)  
- Use double quotes (`"`) for the values of properties/attributes and event details.  
- Only use single quotes (`'`) as apostrophes.  
- No links or URLs allowed in descriptions. If a link is necessary, a [custom JSDoc tag](https://stenciljs.com/docs/docs-json#custom-jsdocs-tags) should be added and parsed in the SDK site.  
- Refrain from using "e.g." or "i.e." references. Leverage "such as" (or similar) where examples are referenced.  
- Use "Accessible" instead of "Aria" language  
- Verify slots don't use the text "optional" in their descriptions  
- Remove optional JSDoc tags for props (`?`)

#### Common slots
- `header` - A slot for adding header text.
 - `content` - A slot for adding the component's content.

#### Events
- For click events, use descriptors such as "Fires", as opposed to "Emits", in line with MDN [mousedown](https://developer.mozilla.org/en-US/docs/Web/API/Element/mousedown_event) and [mouseup](https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseup_event) events.
    - For example: "Fires when the back button has been clicked." 

#### Common properties
- `active` - When true, the component is active.
- `appearance` - Specifies the appearance style of the component.
- `checked` - When true, the component is checked.
- `disabled` -  When true, interaction is prevented and the component is displayed with lower opacity.
- `dismissible` - When true, a close button is added to the component.
- `flipPlacements` - Defines the available placements that can be used when a flip occurs.
[...]

PR #4798, opened 2022-06-23
docs(input): proofread and update api reference

benelan:
**Related Issue:** #4442

## Summary

Proof read and update API ref for `calcite-input`. Also added an `@mdn` jsdoc tag to to link to MDN doc for global and native input attributes. I will be testing if we can parse and display that information using custom tags for doc-json: 
https://stenciljs.com/docs/docs-json#custom-jsdocs-tags


src/components/input/input.tsx
```diff
@@ -200,9 +244,9 @@ export class Input implements LabelableComponent, FormComponent, InteractiveComp
   @Prop({ mutable: true, reflect: true }) editingEnabled = false;
 
   /**
-   * specify the input type
+   * Specifies the input type.
```
geospatialem, line 247: For consistency, should we use _Specifies the ~input~ **component** type_?

src/components/input/input.tsx
```diff
@@ -63,37 +63,47 @@ export class Input implements LabelableComponent, FormComponent, InteractiveComp
...
 
-  /** for number values, displays the locale's group separator */
+  /**
+   * When true, number values are displayed with the locale's group separator.
+   */
   @Prop() groupSeparator = false;
 
-  /** when true, the component will not be visible */
+  /**
+   * when true, the component will not be visible.
```
geospatialem, line 99: Should this be ~when~ **When** true,...? Wondering since this is the "hidden" prop, if we should use language indicating it will hide the component as opposed to it not being visible? (e.g., When true, the component will be hidden and won't be rendered in the browser)?

src/components/input/input.tsx
```diff
@@ -103,26 +113,28 @@ export class Input implements LabelableComponent, FormComponent, InteractiveComp
   @Prop() intlClear?: string;
 
   /**
-   * string to override English loading text
+   * A text label that will appear while loading for screen readers.
```
geospatialem, line 116: For some of the other component consistency we were using "Accessible name that will...". Since screen readers are only one type of user, we could keep it general to ensure we're including more audiences, so maybe something like **Accessible name that will appear while loading**?

src/components/input/input.tsx
```diff
@@ -133,7 +145,11 @@ export class Input implements LabelableComponent, FormComponent, InteractiveComp
...
 
-  /** input max */
+  /**
+   * Specifies the maximum value for type "number".
+   *
+   * @mdn [max](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#max)
+   */
   @Prop({ reflect: true }) max?: number;
 
   /** watcher to update number-to-string for max */
```
geospatialem, line 155: Should the watcher line be included in the PR for consistency?

src/components/input/input.tsx
```diff
@@ -152,46 +172,70 @@ export class Input implements LabelableComponent, FormComponent, InteractiveComp
...
-  /** explicitly whitelist placeholder attribute */
+  /**
+   * Specifies placeholder text for the component.
+   *
+   * @mdn [placeholder](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#placeholder)
+   */
   @Prop() placeholder: string;
 
-  /** optionally add prefix  */
+  /** Adds text to the start of the component.  */
```
geospatialem, line 212: Nitpick: There are two spaces between the text and end comment chars.

src/components/input/input.tsx
```diff
@@ -152,46 +172,70 @@ export class Input implements LabelableComponent, FormComponent, InteractiveComp
...
 
-  /** specify the status of the input field, determines message and icons */
+  /** Specifies the status of the input field, which determines message and icons. */
   @Prop({ mutable: true, reflect: true }) status: Status = "idle";
 
-  /** input step */
+  /**
+   * Specifies the granularity that the component's value must adhere to.
+   *
+   * @mdn [step](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#step)
```
geospatialem, line 234: The MDN url with `../#step` doesn't seem to navigate to the placement in the document. It seems that page doesn't have the capability for the attributes. 

Would `https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/step` suffice in its place?

geospatialem (commented):
Provided some overall feedback on component consistency for review.

src/components/input/input.tsx
benelan: I'm only doing public API ref. So no internal props, watchers, internal events, etc. Do you think we should update all of the comments? Seems unnecessarily time consuming

src/components/input/input.tsx
benelan: I try not to use the word I'm defining in the description. Also setting the `hidden` flag doesn't remove the element from the DOM, so it is technically still rendered. It just isn't visible.

src/components/input/input.tsx
benelan: Weird, it works for me. But sure I can change it

driskull (approved):
👍

PR #4799, opened 2022-06-23
test(dropdown): skip unstable test: scrolling > control max items dis…

Elijbet:
This might be a timing issue.
Screen Shot 2022-06-23 at 2 19 20 PM

I noticed when there is a max item limit and more items, it’ll fail with this message (regardless if there are more items in just one group or spanning 2 groups).


eriklharper (approved):

jcfranco:
@Elijbet Can you create a follow-up issue to restore this test?

Elijbet:
> @Elijbet Can you create a follow-up issue to restore this test?

Will do. Thank you!

PR #4807, opened 2022-06-27
feat(accordion-item,radio-group-item): deprecate iconPosition & icon and add support for iconStart and iconEnd

anveshmekala:
**Related Issue:** #4688 

- This will allow users to add `icon`'s both at the `start` and `end` of `accordion-item` and `radio-group-item` 
- `iconPosition` with `icon` is deprecated in favor of `icon-end` and `icon-start`

Example :
` <calcite-radio-group-item icon-start="car" icon-end="car" value="car">Car</calcite-radio-group-item>`

9F0A5096-2417-45DD-9590-8CE9C5847020_4_5005_c

Note :
- New props always take precedence. 
- Mixing deprecated props with new ones isn't supported. 
- In radio-group-item use either `icon` with `iconPosition` or  `iconEnd` & `iconStart` props  but do not combine them.
- In accordion-item `icon` is an alias for `iconStart`.

Example: 

#### Supported patterns: 
`<calcite-radio-group-item icon-start="car" icon-end="car" > </calcite-radio-group-item>`
OR 
`<calcite-radio-group-item icon="car" icon-position="end" > </calcite-radio-group-item>`


src/components/radio-group-item/radio-group-item.tsx
```diff
@@ -85,9 +117,11 @@ export class RadioGroupItem {
           }}
         >
           {this.icon && this.iconPosition === "start" ? iconEl : null}
+          {this.iconStart ? iconStartEl : null}
           <slot>{value}</slot>
           <slot name={SLOTS.input} />
           {this.icon && this.iconPosition === "end" ? iconEl : null}
+          {this.iconEnd ? iconEndEl : null}
```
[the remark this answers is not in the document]
anveshmekala: Yup the user can set both `icon` and `iconEnd` . I think we need to allow that considering `icon` will be deprecated soon and I checked the case where the user passes both the attribute and the UI doesn't break.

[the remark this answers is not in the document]
anveshmekala: This also adds possibility of using the `icon-end` attribute without removing existing `icon-position =start` with `icon`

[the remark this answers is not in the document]
driskull: I'll defer to @jcfranco here but my initial thoughts were that the new prop should replace the old one and not allow both props to work.

src/components/accordion-item/accordion-item.tsx
```diff
@@ -86,8 +86,23 @@ export class AccordionItem implements ConditionalSlotComponent {
...
+  @Watch("icon")
+  iconHandler(value: string): void {
+    this.iconStart = value;
+  }
+
+  /**Specified an icon to display at the start - accepts Calcite UI icon names. */
+  @Prop({ reflect: true }) iconStart?: string;
+
+  /**Specified an icon to display at the end - accepts Calcite UI icon names. */
+  @Prop({ reflect: true }) iconEnd?: string;
```
driskull, line 105: Should updating iconEnd update icon/iconPosition? Same with iconstart. I see that its updated on connectedCallback but not anytime after.

src/components/accordion-item/accordion-item.tsx
```diff
@@ -131,6 +146,13 @@ export class AccordionItem implements ConditionalSlotComponent {
       this.activeHandler(isExpanded);
       this.expandedHandler(isExpanded);
     }
+
+    if (this.icon) {
```
driskull, line 150: if someone sets both `icon` and `iconStart` I think `iconStart` should take precedence. However, I think this code would honor icon before iconStart. right?

driskull (commented):
comments :D

src/components/accordion-item/accordion-item.tsx
anveshmekala: Will update this.

src/components/accordion-item/accordion-item.tsx
anveshmekala: Either ways we have to make changes to the current implementation.

src/components/radio-group-item/radio-group-item.tsx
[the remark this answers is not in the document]
jcfranco: Agreed. ☝️

When we replace props, new ones take precedence and mixing them will not be supported.

anveshmekala:
@geospatialem @benelan  please suggest what would be the best way to include the notes section of the PR in docs. Thanks.

src/components/accordion-item/accordion-item.tsx
```diff
@@ -172,7 +200,9 @@ export class AccordionItem implements ConditionalSlotComponent {
 
   render(): VNode {
     const dir = getElementDir(this.el);
-    const iconEl = <calcite-icon class={CSS.icon} icon={this.icon} scale="s" />;
+    const iconStartEl = <calcite-icon class={CSS.iconStart} icon={this.iconStart} scale="s" />;
```
driskull, line 203: these elements might need `key` attributes since they are conditional.

src/components/accordion-item/accordion-item.tsx
```diff
@@ -190,12 +220,15 @@ export class AccordionItem implements ConditionalSlotComponent {
...
               tabindex="0"
             >
-              {this.icon ? iconEl : null}
-              <div class={CSS.headerText}>
-                <span class={CSS.heading}>{this.heading || this.itemTitle}</span>
-                {this.itemSubtitle || this.description ? (
-                  <span class={CSS.description}>{this.description || this.itemSubtitle}</span>
-                ) : null}
+              <div class={CSS.headerContainer}>
+                {this.iconStart ? iconStartEl : null}
```
driskull, line 224: nitpick: would be nice to move the conditional part to where the const is being declared.

src/components/accordion-item/accordion-item.tsx
```diff
@@ -190,12 +220,15 @@ export class AccordionItem implements ConditionalSlotComponent {
...
+                <div class={CSS.headerText}>
+                  <span class={CSS.heading}>{this.heading || this.itemTitle}</span>
+                  {this.itemSubtitle || this.description ? (
```
driskull, line 227: Is there a follow up refactor issue to make `description` the source of truth so that these two don't compete?

src/components/radio-group-item/radio-group-item.tsx
```diff
@@ -43,15 +43,29 @@ export class RadioGroupItem {
...
   @Prop({ reflect: true }) icon?: string;
 
   /** flip the icon in rtl */
   @Prop({ reflect: true }) iconFlipRtl = false;
 
-  /** optionally used with icon, select where to position the icon */
+  /**
+   * optionally used with icon, select where to position the icon
+   *
+   * @deprecated use iconStart and iconEnd
```
driskull, line 59: Should this say use `iconStart or iconEnd`?

src/components/radio-group-item/radio-group-item.tsx
```diff
@@ -43,15 +43,29 @@ export class RadioGroupItem {
     this.calciteInternalRadioGroupItemChange.emit();
   }
 
-  /** optionally pass an icon to display - accepts Calcite UI icon names  */
+  /**
+   * optionally pass an icon to display - accepts Calcite UI icon names
+   *
+   * @deprecated use iconStart and iconEnd
```
driskull, line 49: Should this say use `iconStart or iconEnd`?

src/components/radio-group-item/radio-group-item.tsx
```diff
@@ -64,6 +78,24 @@ export class RadioGroupItem {
     const appearance: RadioAppearance = getElementProp(this.el, "appearance", "solid");
     const layout: Layout = getElementProp(this.el, "layout", "horizontal");
 
+    const iconStartEl = (
+      <calcite-icon
```
driskull, line 82: Same comments from the accordion items about the key attribute and conditional logic.

src/components/radio-group-item/radio-group-item.tsx
```diff
@@ -84,10 +116,12 @@ export class RadioGroupItem {
             "label--outline": appearance === "outline"
           }}
         >
-          {this.icon && this.iconPosition === "start" ? iconEl : null}
+          {this.icon && this.iconPosition === "start" && !this.iconStart ? iconEl : null}
```
driskull, line 119: Can you refactor this to just render `iconStart` like you have in the accordion? That way its less confusing here.

src/components/radio-group-item/radio-group-item.tsx
```diff
@@ -84,10 +116,12 @@ export class RadioGroupItem {
...
+          {this.iconStart ? iconStartEl : null}
           <slot>{value}</slot>
           <slot name={SLOTS.input} />
-          {this.icon && this.iconPosition === "end" ? iconEl : null}
+          {this.icon && this.iconPosition === "end" && !this.iconEnd ? iconEl : null}
```
driskull, line 123: Can you refactor this to just render `iconEnd` like you have in the accordion? That way its less confusing here.

driskull (changes requested):
Looking good @anveshmekala! Just had some feedback and refactor suggestions. 👍 💯

benelan:
> @geospatialem @benelan please suggest what would be the best way to include the notes section of the PR in docs. Thanks.

The description following the `@deprecated` jsdoc tag shows up in the doc, so I'd add it there for these two notes:

> In radio-group-item use either icon with iconPosition or iconEnd & iconStart props but do not combine them.
In accordion-item icon is an alias for iconStart.

@macandcheese what do you think about adding a tooltip to the `deprecated` chip that says

> Where relevant, using deprecated properties with their replacement is not supported. However, replacement properties take precedence over deprecated properties when both are set.

Did I interpret that correctly for the first two notes Anvesh?

benelan:
Is the behavior from those notes true throughout CC or just for the changes in this PR? If we added a tooltip like I suggested above it would show up on deprecated properties for all components.

anveshmekala:
> Is the behavior from those notes true throughout CC or just for the changes in this PR? If we added a tooltip like I suggested above it would show up on deprecated properties for all components.

This is related to the changes in PR but not the whole CC repo.

anveshmekala:
> > @geospatialem @benelan please suggest what would be the best way to include the notes section of the PR in docs. Thanks.
> 
> The description following the `@deprecated` jsdoc tag shows up in the doc, so I'd add it there for these two notes:
> 
> > In radio-group-item use either icon with iconPosition or iconEnd & iconStart props but do not combine them.
> > In accordion-item icon is an alias for iconStart.
> 
> @macandcheese what do you think about adding a tooltip to the `deprecated` chip that says
> 
> > Where relevant, using deprecated properties with their replacement is not supported. However, replacement properties take precedence over deprecated properties when both are set.
> 
> Did I interpret that correctly for the first two notes Anvesh?

SPOT ON .

I will add  `In radio-group-item use either icon with iconPosition or iconEnd & iconStart props but do not combine them.`  description following `@deprecated` tag in `radio-group-item` and for `accordion` it will be straightforward.

src/components/radio-group-item/radio-group-item.tsx
anveshmekala: I don't think we can render iconEnd here as we did in accordion, the issue is with the mapping of the icon to `iconEnd` when `iconPosition` is set to `end` and able to mutate either of `iconStart` or `iconEnd`   when `iconPosition` changes but not `icon` value.

src/components/radio-group-item/radio-group-item.tsx
anveshmekala: I don't think we can render iconStart here as we did in accordion, the issue is with the mapping of the icon to `iconStart` when `iconPosition` is set to `start` and able to mutate either of `iconStart` or `iconEnd`   when `iconPosition` changes but not `icon` value.

src/components/accordion-item/accordion-item.tsx
anveshmekala: `description` is a string and doesn't require a follow-up refactor IMO  as we don't append both the attribute to DOM if one of them is true. Even when the user parses both `description` & `itemSubtitle`, the newer prop takes precedence.

driskull: Yeah that makes sense. I was just thinking a const would be nice so that `this.description || this.itemSubtitle` isn't repeated.

src/components/radio-group-item/radio-group-item.tsx
driskull: I think these should be updated into one line then so its clear they both can't be true.

```
{this.iconEnd ? iconEndEl : this.icon && this.iconPosition === "end" ? iconEl : null}
```

anveshmekala: sorry to misunderstand, I thought you were suggesting rendering `iconStart` instead of both `iconStart`  &  conditional `icon` with `iconPosition`

anveshmekala:
@macandcheese  this PR will add `iconEnd` position for accordion-item,  please review the design. Thanks!

cc : @ashetland

ashetland:
> @macandcheese this PR will add `iconEnd` position for accordion-item, please review the design. Thanks!
> 
> cc : @ashetland

@SkyeSeitz Were we already aware of this? I thought I saw this mentioned in our kit....

src/components/accordion-item/accordion-item.tsx
```diff
@@ -172,7 +200,13 @@ export class AccordionItem implements ConditionalSlotComponent {
 
   render(): VNode {
     const dir = getElementDir(this.el);
-    const iconEl = <calcite-icon class={CSS.icon} icon={this.icon} scale="s" />;
+    const iconStartEl = this.iconStart ? (
+      <calcite-icon class={CSS.iconStart} icon={this.iconStart} key={CSS.iconStart} scale="s" />
```
driskull, line 204: Interesting using the CSS class for the key :)

I'm ok with it, just wondering if we should allow it or not? 🤷

src/components/radio-group-item/radio-group-item.tsx
```diff
@@ -84,10 +123,12 @@ export class RadioGroupItem {
             "label--outline": appearance === "outline"
           }}
         >
-          {this.icon && this.iconPosition === "start" ? iconEl : null}
+          {iconAtStart}
```
driskull, line 126: @anveshmekala could there possibly be `{iconAtStart}` and `{this.iconStart ? iconStartEl : null}` both rendered?

I was thinking only one of them would be rendered right? If so, can we move the iconStartEl to the `iconAtStart` const above? Same with the end icons.

To me, it just seems confusing with iconStartEl and iconAtStart. It seems like only one of them should only be possible at a time.

driskull (commented):
comments! 👍

src/components/accordion-item/accordion-item.tsx
anveshmekala: Thought of using `guid` from util classes, but these are unique to the component which could be used a key . I can change it if the pattern is inconsistent.

src/components/radio-group-item/radio-group-item.tsx
anveshmekala: Yup, there is a possibility of both `iconAtStart `and `iconStartEl `being rendered because we are not maintaining a single source of truth for the `radio-group-item` component due to the complexity. Added a note next to the` deprecated` tag advising users to not use both of them but one or the other.

driskull: ok. makes sense then. 👍

src/components/accordion-item/accordion-item.tsx
driskull: Maybe just set the keys to `"icon-start"` since using the classes could mess things up if a class changes.

driskull (approved):
👍 :shipit:

anveshmekala:
> > @macandcheese this PR will add `iconEnd` position for accordion-item, please review the design. Thanks!
> > cc : @ashetland
> 
> @SkyeSeitz Were we already aware of this? I thought I saw this mentioned in our kit....

Wanted to make sure spacing and alignment match Figma.

anveshmekala:
Can I get a design review for this one. Thanks

ashetland:
> Can I get a design review for this one. Thanks

We are looking at this today.

ashetland:
@SkyeSeitz will follow up. Also adding @yelenakreyndel so we can start putting our process in place for design reviews.

benelan:
RE: https://github.com/Esri/calcite-components/issues/4996#issuecomment-1210108567 - maybe we shouldn't doc how unsupported functionality behaves? If we really want to doc it, then it should be consistent. cc @jcfranco

anveshmekala:
> RE: [#4996 (comment)](https://github.com/Esri/calcite-components/issues/4996#issuecomment-1210108567) - maybe we shouldn't doc how unsupported functionality behaves? If we really want to doc it, then it should be consistent. cc @jcfranco

The current implementation in this PR deviates from the pattern of keeping props in sync because of the dependency of one prop with the other ( icon and iconPosition )  whereas new props are independent (`icon-start` , `icon-end`) which increases the complexity of keeping props in sync. To avoid confusion we are documenting the unsupported behavior.

Referring to [#4996](https://github.com/Esri/calcite-components/issues/4996#issuecomment-1210108567) - even if one prop changes the other should change keeping both in sync.

anveshmekala:
any update on this one @SkyeSeitz ?

SkyeSeitz:
> any update on this one @SkyeSeitz ?

@ashetland and I have taken a look and this looks good to go. 🚀

anveshmekala:
> > any update on this one @SkyeSeitz ?
> 
> @ashetland and I have taken a look and this looks good to go. 🚀

before we install, can you review the screener changes.  Thanks

anveshmekala:
> > > any update on this one @SkyeSeitz ?
> > 
> > 
> > @ashetland and I have taken a look and this looks good to go. 🚀
> 
> before we install, can you review the screener changes. Thanks

PR #4808, opened 2022-06-27
refactor(combobox): rename functions that emit open/close events

Elijbet:
**Related Issue:** #4689 

Small follow-up: rename functions to be more explicit that it is a handler function as opposed to an event.


eriklharper (approved):
🎉

PR #4809, opened 2022-06-27
fix(input): set the default numbering system for arab languages to latin

benelan:
**Related Issue:** #3079

## Summary
Sets the default numbering system for arab languages to latin. This is for browser consistency so that Firefox behaves like Chrome and Safari. According to the locale team, this is what our arab users want and it is the way other esri products do it, including the JSAPI team.

Users can still use the arab numbering system like:
```html
<calcite-input numbering-system="arab" type="number" value="5"></calcite-input>
```

### DIFF
--- a/src/components/input/input.stories.ts
+++ b/src/components/input/input.stories.ts
@@ -275,6 +275,6 @@ export const HebrewNumberingSystem = (): string =>
   html` <calcite-input type="number" locale="ar-EG" numbering-system="hebr" value="123456"></calcite-input>`;
 
 export const ArabicLocaleWithLatinNumberingSystem = (): string =>
-  html` <calcite-input type="number" locale="ar-EG" numbering-system="latn" value="123456"></calcite-input>`;
+  html` <calcite-input type="number" locale="ar-EG" value="123456"></calcite-input>`;
 
 export const disabled = (): string => html`<calcite-input disabled value="disabled"></calcite-input>`;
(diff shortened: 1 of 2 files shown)


driskull (approved):
Nice!

eriklharper (approved):

jcfranco (approved):
Awesome! 

Before merging, can you add a simple test for this? Also, will this be doc'd at the SDK-level (and not per component)?

jcfranco:
I'm fine using `numberingSystem` to change it, but the defaults should be consistent across browsers. This wouldn't be the case if we use the browser defaults (and forces users to be aware of this and set it), right?

benelan:
They are currently not consistent across browsers, but this PR would do that. Context starting from [this comment](https://github.com/Esri/calcite-components/issues/3079#issuecomment-1167912096) down.

driskull:
I could go either way on this one. I understand the confusion for users and making it consistent for all browsers.

However, I also see the problem with trying to juggle browser inconsistencies with the i18n function.

If we do try to normalize them, we should track what we have so that we can potentially remove it in the future. If more browsers start to do their own thing this could get messy.

jcfranco:
Let's proceed with this one.

> If we do try to normalize them, we should track what we have so that we can potentially remove it in the future. If more browsers start to do their own thing this could get messy.

Good idea. Maybe the SDK's internationalization page would be a good place for this.

benelan:
> Good idea. Maybe the SDK's internationalization page would be a good place for this.

@geospatialem can you please help document this? Let me know if you have any questions!

Issue #4793, opened 2022-06-23
Enter key does not submit form in Safari


ReneU:
### Actual Behavior

When using a `calcite-input` within a `<form>` element, pressing enter in Safari results in a JS error and the form is not submitted.

![image](https://user-images.githubusercontent.com/5748802/175255376-ecbd8ff2-8f46-4498-8059-2873b4b7aa38.png)

### Expected Behavior

Similar to other browsers, the form should be submitted correctly (i.e. the form's `onsubmit` event should be triggered).

### Reproduction Sample

https://codepen.io/reneu/pen/jOZjXRq?editors=1000

### Reproduction Steps

1. Open linked codepen with Safari browser
2. Focus input element
3. Press enter

### Reproduction Version

v1.0.0-beta.83

### Relevant Info

macOS Menterey Version 12.4
Safari Version 15.5 (17613.2.7.1.8)

It seems like [HTMLFormElement.requestSubmit](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/requestSubmit#browser_compatibility) has only 'preview browser support' in Safari but is being used in: https://github.com/Esri/calcite-components/blob/4bca90679c19a2c7fed7acb8f126bdd223c0e50f/src/utils/form.tsx#L141-L143

geospatialem:
Verified on Safari Version 15.5 (17613.2.7.1.8) on `next`.

ReneU:
@driskull @geospatialem Thank you!

PR #4815, opened 2022-06-28
fix: Enter key should submit forms in Safari

driskull:
**Related Issue:** #4793

## Summary

fix: Enter key should submit forms in Safari (#4793)

### DIFF
--- a/src/utils/form.tsx
+++ b/src/utils/form.tsx
@@ -139,7 +139,13 @@ function hasRegisteredFormComponentParent(
  * @param component
  */
 export function submitForm(component: FormOwner): void {
-  component.formEl?.requestSubmit();
+  const { formEl } = component;
+
+  if (!formEl) {
+    return;
+  }
+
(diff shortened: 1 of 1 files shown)


eriklharper (approved):
Interesting the kind of hacks needed for Safari!  #SafariIsTheNewIE

driskull:
Yep. I thought we checked and that requestSubmit worked in all browsers but I guess not :\

jcfranco:
@driskull @eriklharper @geospatialem do either of you recall if the original issue only had inputs inside a form? It's no longer available. The polyfill we use for Safari was applied in `calcite-button` and could lead to the reported error if not used in a form. If this is the case, I've submitted https://github.com/Esri/calcite-components/pull/6623 to address this.

driskull:
I don't recall. I think it just had buttons for submitting.

eriklharper:
I don't recall either...

Issue #4691, opened 2022-06-06
Accordion buttons should indicate whether panel is expanded or collapsed to screen-readers


FelFly:
### Summary

Auditory indicators of state on whether an accordion panel has been collapsed or expanded are missing in VoiceOver and NVDA.

### Actual Behavior

When navigating an accordion by screen-reader, one hears that the accordion header is a button, but when you select it via (spacebar) there is no auditory indicator that anything has happened.

### Expected Behavior

When selecting an accordion heading by spacebar, I would expect to hear "{accordion heading}, expanded, button" and/or "{accordion heading}, collapsed, button" to communicate the state of the accordion panel.

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/accordion/#sample

### Reproduction Steps

1. Turn on VoiceOver (CMD+F5)
2. Navigate into webpage (Ctrl+Option+Shift+Down Arrow)
3. Navigate to accordion sample container (Ctrl+Option+Right Arrow)
4. Enter into container (Ctrl+Option+Shift+Down Arrow)
5. Navigate to second accordion button
6. Select Accordion (Ctrl+Option+Spacebar)

### Reproduction Version

beta.82

### Working W3C Example/Tutorial

https://design-system.service.gov.uk/components/accordion/

### Relevant Info

VoiceOver (10.15.7) / Safari (15.5) in MacOS Catalina
NVDA (2021.2) / Firefox (100.0) in Windows

benelan:
I don't have VoiceOver, but verified w/ NVDA in beta.86

PR #4822, opened 2022-06-28
fix(accordion-item): Indicate whether accordion is expanded to screen readers.

driskull:
**Related Issue:** #4691

## Summary

fix(accordion-item): Indicate whether accordion is expanded to screen readers.


anveshmekala (approved):
👍

Issue #404, opened 2020-03-20
calcite-dropdown: width as wide as dropdown-trigger ?


AdelheidF:
The design:
<img src=https://user-images.githubusercontent.com/44481292/77197774-fbcd3f80-6aa2-11ea-8518-62adb8667f59.png width=250/>

The component appears inside a shell panel which changes width depending on browser width.
So it's hard to find a width=[s,m,l] that will fit in all cases. 
Really what I would want is a dropdown that is as wide as the `dropdown-trigger` element. Or one that is at least as wide as the `drodown-trigger` element and expands with longer text like the standard HTML dropdown.

<img src=https://user-images.githubusercontent.com/44481292/77198112-9cbbfa80-6aa3-11ea-9eb1-8c96aa86cf16.png width=200/><img src=https://user-images.githubusercontent.com/44481292/77198038-77c78780-6aa3-11ea-82be-025a0f92f138.png width=200/>

Maybe a `width=auto`?

Current implementation with a  small screen (width=s)
<img src=https://user-images.githubusercontent.com/44481292/77198311-fde3ce00-6aa3-11ea-91f5-b01b4f6575a5.png width=200/>

With a little bit bigger screen.
<img src=https://user-images.githubusercontent.com/44481292/77198439-31bef380-6aa4-11ea-8cc5-1d4b0103820c.png width=220/>

macandcheese:
Yeah, it may be worth it to add a few options here... one that was fixed to the size of the trigger element, one that expanded the dropdown (to edge of page) based on length of content, and... maybe some combination of those?

For naming... 
- "constrain to width of trigger" = `inherit` or `trigger`
- "size to width of content (not extending past edge of page)" = `auto`
- "constrain to width of trigger, (dropdown could not be narrower than trigger) , but extend to edge of page if width was wider" - `?????`

Thoughts?

dhrumil83:
Same issue with font dropdown:

![image](https://user-images.githubusercontent.com/14285436/78699037-13d0fb80-78b8-11ea-8692-6a5cf485de9e.png)

macandcheese:
Yeah, will be a good enhancement to make. Would probably just need to get the width of the slotted `dropdown-trigger` element, and check for updates to that width for resize, etc. 

[...]

PR #4823, opened 2022-06-29
fix(dropdown): Restore trigger container to flex-auto

driskull:
**Related Issue:** #404

## Summary

fix(dropdown): Restore trigger container to flex-auto. (#404)


anveshmekala (approved):
LGTM!

Can we add a screener story with dropdown width set to 100% in a custom width container to safeguard.

driskull:
Great idea. added story

Issue #4826, opened 2022-06-29
Components that are closed/dismissed should no longer render any UI


driskull:
### Description

> I think we should probably update `chip` to no longer render any UI if closed.
> 
> If not part of this PR, a followup issue/pr should be linked

https://github.com/Esri/calcite-components/pull/4789#pullrequestreview-1020883311

### Proposed Advantages

We should be consistent with how we render closed/dismissed components. 

Most of the components will no longer render any UI when closed. However, the `chip` component still renders UI and emits an event.

### Which Component

chip

### Relevant Info

https://github.com/Esri/calcite-components/pull/4789

benelan:
verified on next

PR #4836, opened 2022-06-30
feat(chip): add support for closed prop

anveshmekala:
**Related Issue:** #4826 

## Summary

This PR will add support for `closed` prop on `calcite-chip` which hides the chip element when set to true.


anveshmekala:
> **Related Issue:** #4826
> 
> ## Summary
> This will remove the `calcite-chip` element from DOM when the user clears the chip using the close button.

adding `closed` property isn't necessary considering the only time chip can be closed is when the `closable` prop is set to true and the user clicks on the close icon. 

If we have a use case where the user didn't set the `closable` prop and yet wants to hide and display an icon based on some user interaction then yes we may need `closed/open` prop.

benelan:
> adding `closed` property isn't necessary considering the only time chip can be closed is when the `closable` prop is set to true and the user clicks on the close icon.

I disagree, I think we should add a `closed` property for when a chip needs to be displayed again after clicking the "x" on closable elements. Otherwise it has to be completely re-added to the DOM like Matt is saying

driskull:
Ben summed it up! hit the nail on the head. 🔨

benelan:
multi selection comboboxes will need a refactor after adding the `closed` prop

Issue #4685, opened 2022-06-04
color-picker internal canvas not resizing on scale change


jcfranco:
### Actual Behavior

Color field/hue slider layout doesn't respond well to changing scale.

![Screen Shot 2022-06-03 at 4 33 50 PM](https://user-images.githubusercontent.com/197440/171968234-da62a459-a3af-4888-8d4e-2483ba04cde1.png)

### Expected Behavior

Color field/hue slider layout to match according to scale.

### Reproduction Sample

https://codepen.io/jcfranco/pen/MWQXLqq

### Reproduction Steps

1. Open codepen
2. Wait for timeout to change scale
3. Note how the layout is messed up

### Reproduction Version

beta.82

PR #4839, opened 2022-06-30
fix(color-picker): resize color field/hue slider canvas on scale change

jcfranco:
**Related Issue:** #4685 

## Summary

This addresses an issue where the internal canvas (for color field and hue slider) would stay the same size as its initial scale.


driskull (approved):
awesome 👏 👍

Issue #2554, opened 2021-07-16
Unstable Test: dropdown-item - item selection should work when placed inside shadow DOM


driskull:
## Summary

Investigate why the following change is necessary. 

https://github.com/Esri/calcite-components/pull/2485#pullrequestreview-708562921

https://github.com/Esri/calcite-components/blob/deb41815531c0d1282a7357b0e3f6667b8a9fd74/src/components/calcite-dropdown/calcite-dropdown.e2e.ts#L951

**Which component?:** dropdown-item

**Which test(s)?:** 

https://github.com/Esri/calcite-components/pull/2485#pullrequestreview-708562921

**Test error:**

Test requires another request for animation frame change.

**PR that skipped the test:** N/A

benelan:
verified on master

PR #4843, opened 2022-07-01
test(dropdown): split up browser functions to allow rendering to finish before asserting on the selected item

jcfranco:
**Related Issue:** #2554

## Summary

In the previous version, the browser function had to wait on two animation frames after clicking an item for all pending changes to be done. This splits up the browser function to leverage `page.waitForChanges` to make it a bit clearer on what's going on.


driskull (approved):
👍 💥 ✨

Issue #4172, opened 2022-02-28
calcite-input: up/down button hold not working with touch


AdelheidF:
### Actual Behavior

Holding the up/down arrows to speed up the number change works with mouse interaction, but not with touch.
If this is a limitation with touch then it should at least be mentioned in the doc.

### Expected Behavior

Holding up/down arrows with touch keeps incrementing or decrementing the number.

### Reproduction Sample

https://codepen.io/afreitag/pen/jOaeejz

### Reproduction Steps

open the codepen on a touch device

### Reproduction Version

@Esri/calcite-components@1.0.0-beta.77

### Relevant Info

Win 10 on Surface Pro 7, but probably more

PR #4846, opened 2022-07-06
fix(input): update number stepper buttons to work with both mouse/touch inputs

jcfranco:
**Related Issue:** #4172

## Summary

Switches mouse events to pointer events to improve interactivity regardless of input type.


driskull (approved):
👍 nice fix

Issue #4662, opened 2022-05-31
calcite-radio-group does not reset properly in form


COV-GIS:
### Actual Behavior

No calcite-radio-group-item is checked after form reset.

Also...item checked when form reset cannot be reselected until a different item is selected first.

### Expected Behavior

calcite-radio-group should reset to default item checked.

### Reproduction Sample

https://codepen.io/COV-GIS/pen/bGLvZWJ

### Reproduction Steps

1. click `Reset` button
2. click `Layers` item

### Reproduction Version

1.0.0-beta.82

geospatialem:
Verified on beta.86
![verify-clear-select-radio](https://user-images.githubusercontent.com/5023024/179628598-eb98f2e6-8426-4fd8-8ecd-b9a253e8f9f6.gif)

PR #4848, opened 2022-07-06
fix(radio-group): fix form reset for radio groups with initial selected item

jcfranco:
**Related Issue:** #4662 

## Summary

This fixes an issue where the radio group would not assign its default value properly.

### DIFF
--- a/src/components/radio-group/radio-group.tsx
+++ b/src/components/radio-group/radio-group.tsx
@@ -1,22 +1,28 @@
 import {
-  Component,
-  Event,
-  h,
-  EventEmitter,
-  Listen,
-  Element,
-  Prop,
-  Watch,
-  Host,
   Build,
+  Component,
+  Element,
+  Event,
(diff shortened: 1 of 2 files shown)


driskull (approved):
Nice!! 👍

src/components/radio-group/radio-group.tsx
```diff
@@ -124,6 +130,10 @@ export class RadioGroup implements LabelableComponent, FormComponent, Interactiv
     }
   }
 
+  componentDidLoad(): void {
+    afterConnectDefaultValueSet(this, this.value);
```
eriklharper, line 134: I feel like a more intuitive name for this would be `setDefaultValueAfterConnect`

driskull: lol

driskull: After connect default value you set

PR #4852, opened 2022-07-06
refactor(flow, flow-item)!: Adds `open` property to flow-item to control which panel should be displayed.

driskull:
**Related Issue:** #5072

## Summary

refactor(flow, flow-item)!: Adds `open` property to flow-item to control which panel should be displayed.

BREAKING CHANGE: The `open` property must be defined to display a specific `flow-item` within a flow.


driskull:
@jcfranco do you want to discuss this PR soon?

Besides the flow, the only other component that removes an element seems to be the pick list: https://github.com/Esri/calcite-components/blob/master/src/components/pick-list/shared-list-logic.ts#L201-L204

Issue #991, opened 2020-09-15
Bug: flash of default styles due to Stencil bug with hydratedFlag


eriklharper:
### Summary

This bug is due to an issue with how Stencil applies CSS for the `hydratedFlag` config option: https://github.com/ionic-team/stencil/issues/2245. This bug would be a workaround to temporarily fix the flashing while Stencil works on a more permanent solution.

### Actual Behavior

When loading the page and on subsequent reloads, a flash of default styling appears on components in certain scenarios.

### Expected Behavior

The correct styling should appear without any flash of incorrect styling based on the supplied props.

### Reproduction Steps

1. Render a `calcite-label` with a `scale` set to `s` or `l`
2. Reload the page multiple times and notice the flash of the medium font size
3. Observe in the DOM that the `calcite-hydrated` attribute takes a second to apply before the correct font size is shown.

```
<calcite-label scale="s">Test Label</calcite-label>
```

paulcpederson:
I think I may have found a temporary workaround for this. Basically, we can select elements that are `:not` hydrated and hide their contents in the global css file:

```
calcite-accordion:not([calcite-hydrated]) * {
  visibility: hidden;
}
```

Further, this flash only happens when components contain a slot. If a component doesn't have a slot, it won't show anything until js runs anyway, where stencil's hydrated functionality will work fine. This makes the list smaller in our css.

Elements with slots are:

```
calcite-accordion
calcite-accordion-item
calcite-action
calcite-action-bar
calcite-action-group
calcite-action-pad
calcite-alert
calcite-block
calcite-block-section
calcite-button
calcite-card
calcite-checkbox
calcite-chip
calcite-combobox
calcite-combobox-item
calcite-date
calcite-dropdown
calcite-dropdown-group
calcite-dropdown-item
calcite-flow
calcite-input
calcite-input-message
calcite-label
calcite-link
calcite-modal
calcite-notice
calcite-panel
calcite-pick-list/shared-list-render
calcite-pick-list-group
calcite-pick-list-item
calcite-popover
calcite-radio-button
calcite-radio-button-group
calcite-radio-group
calcite-radio-group-item
calcite-scrim
calcite-shell
calcite-shell-center-row
calcite-shell-panel
calcite-sortable-list
calcite-split-button
calcite-stepper
calcite-stepper-item
calcite-tab
calcite-tab-nav
calcite-tab-title
calcite-tabs
calcite-tile-select
calcite-tile-select-group
calcite-tip
[...]

Issue #620, opened 2020-05-22
Gatsby JS SSR with Calcite Components


patrickarlt:
We have been successfully using Calcite Component in a lot of our work on ArcGIS for Developers. However our main blocker remains the fact that there will always be a FOUC with Calcite Component because the JS and CSS are  bundled together and the CSS won't apply until the JavaScript renders leaving items in slots unstyled. This was particularly noticeable on `<calcite-tree>` https://github.com/Esri/calcite-components/issues/609 in cases where we had ~3000+ items like the [ArcGIS Enterprise SDK API Ref](https://developers.arcgis.com/enterprise-sdk/api-reference/net/esriGeometry/Geometry_library/) often taking several seconds to parse and initialize.

Here is a smaller example from the Enterrpise SDK Guide (~100) items:

![72474343-7628cd00-379d-11ea-918d-75df832ba37a](https://user-images.githubusercontent.com/378557/82714961-e4970380-9c45-11ea-8b58-625b76987346.gif)

We eventually re-wrote a sub-set of `<calcite-tree>` in React to avoid this and maintain more control over rendering and performance which is what you see on the site now.

Stencil does support pre-rendering via [hydrate app](https://stenciljs.com/docs/hydrate-app) which I setup for Calcite Components. I then wrote a [Gatsby JS plugin for calcite component prerendering](https://gist.github.com/patrickarlt/0feeab6770563845cb4d3ec2d2f32cd6) . This worked but had a few issues:

1. Poor prerender performance. Since this runs over every page and renders all components on that page our ~3000 tree items prerendered over all 3000 pages  for ~9 million components. This looks several hours even working across multiple cores.
2. Inconsistent prerender results. We noticed that our tree behaved subtly differently when run through hydrate app vs run client side only. We never debugged the because of other issues.

In the end I think it would help to define what the scope of Calcite Components is and how it should be used. Our use cases on ArcGIS for Developers require everything to be prerendered for SEO we want to avoid a FOUC and thus can't wait for Calcite Components to load and style everything. I'm sure esri.com has similar issues. This has lead us to limiting Calcite Component use to things like `<calcite-dropdown>`, `<calcite-tooltip>` `<calcite-modal>` and `<calcite-popover>` which can hide under other pre-rendered elements.

Ideally we could:

1. Improve prerender performance of all our components `<calcite-tree>` is a big one for us but we could also use others.
2. Improve prerender stability of components. In general this might be fine with a few exceptions but no one is testing it.
[...]

Issue #609, opened 2020-05-20
[calcite-tree] Slotted children FOUC on load


macandcheese:
I think generally there are some load-in delays across components, especially when many are used on a page... but there seems to be something *extra* wrong with `calcite-tree` - note this example with just a single rendered item flashing in as an unstyled link before styling is applied (choppy gif alert 🚨):

![Screencast 2020-05-20 at 11 09 40 AM](https://user-images.githubusercontent.com/4733155/82481762-a86b7380-9a8a-11ea-8a4b-3636edcb0347.gif)

It's obviously much more noticeable when many items are rendered. Additionally I think all the css here can be cleaned up... considerably... maybe in conjunction with updates requested in https://github.com/Esri/calcite-components/issues/522 and https://github.com/Esri/calcite-components/issues/529

PR #4856, opened 2022-07-06
fix: css-only hydration to prevent FOUC (#991)

paulcpederson:
**Related Issue:** #991

(see also #620, #609)

## Summary

Because stencil's hydration styles are applied via JavaScript, there is a time before the JavaScript has run where you can see the slotted content flash in before the component is rendered. 

A few releases ago stencil allowed you to turn off their hydrated styles entirely via the `invisiblePrehydration` config flag. This pull request opts out of the injected styles and generates a set of styles for all our components so that they are invisible if they are not hydrated. Because these styles are then in the global css file, they will apply to the page even prior to stencil's loader running at all,  ensuring that the FOUC is removed.

There may be a better way to generate this list of components, but currently just reading the components directory and adding a style for each one.


driskull:
This is cool. Do we want to try to tackle the components that should be `visibility:hidden` by default as well? If not, we could do it as a followup. I'm thinking just a JSON file where we can map the components hidden by default.

https://github.com/Esri/calcite-components/search?q=calcite-hydrated-hidden&type=code

paulcpederson:
@driskull yeah, for sure. I didn't totally understand all the calcite hydrated hidden stuff, so I left it as is, but could absolutely just remove the `:not([calcite-hydrated])` and get hidden by default for those.

driskull:
>  could absolutely just remove the :not([calcite-hydrated]) and get hidden by default for those.

Yep, that would be ideal.

We would just need to maintain a list/map of those components

support/hydrationStyles.ts
```diff
@@ -0,0 +1,15 @@
+const fs = require("fs");
```
driskull, line 1: @paulcpederson can you use...

```
import fs from "fs";
import path from "path";
```

So we get some type safety?

driskull: I think we'd have to update our package here :\. Related to https://github.com/Esri/calcite-components/pull/4700#issuecomment-1171487395

support/hydrationStyles.ts
```diff
@@ -0,0 +1,15 @@
...
+  const components = await fs.promises.readdir(path.normalize(`${__dirname}/../src/components/`));
+  const selectors = components.map((component) => `calcite-${component}:not([calcite-hydrated])`);
+  const fileContents = `// generated by support/hydrationStyles.ts
+${selectors.join(",\n")} {
+  visibility: hidden;
+  pointer-events: none;
+}`;
+  await fs.promises.writeFile(stylePath, fileContents);
+  process.exit(0);
+})();
```
driskull, line 15: @paulcpederson ran this on a mac and I'm seeing some other lines added, such as...

```
// generated by support/hydrationStyles.ts
calcite-.DS_Store:not([calcite-hydrated])
calcite-calcite-accordion:not([calcite-hydrated]),
calcite-calcite-action-menu:not([calcite-hydrated]),
calcite-calcite-alert:not([calcite-hydrated]),
calcite-calcite-block-section:not([calcite-hydrated]),
calcite-calcite-button:not([calcite-hydrated]),
calcite-calcite-chip:not([calcite-hydrated]),
calcite-calcite-color:not([calcite-hydrated]),
calcite-calcite-color-picker:not([calcite-hydrated]),
calcite-calcite-combobox:not([calcite-hydrated]),
calcite-calcite-date-picker:not([calcite-hydrated]),
calcite-calcite-dropdown:not([calcite-hydrated]),
calcite-calcite-dropdown-group:not([calcite-hydrated]),
calcite-calcite-fab:not([calcite-hydrated]),
calcite-calcite-graph:not([calcite-hydrated]),
calcite-calcite-icon:not([calcite-hydrated]),
calcite-calcite-input:not([calcite-hydrated]),
calcite-calcite-input-date-picker:not([calcite-hydrated]),
calcite-calcite-input-message:not([calcite-hydrated]),
calcite-calcite-label:not([calcite-hydrated]),
calcite-calcite-list:not([calcite-hydrated]),
calcite-calcite-list-item:not([calcite-hydrated]),
calcite-calcite-modal:not([calcite-hydrated]),
calcite-calcite-pagination:not([calcite-hydrated]),
calcite-calcite-panel:not([calcite-hydrated]),
calcite-calcite-pick-list:not([calcite-hydrated]),
calcite-calcite-popover:not([calcite-hydrated]),
calcite-calcite-popover-manager:not([calcite-hydrated]),
calcite-calcite-radio-group:not([calcite-hydrated]),
calcite-calcite-shell:not([calcite-hydrated]),
calcite-calcite-shell-panel:not([calcite-hydrated]),
calcite-calcite-stepper:not([calcite-hydrated]),
calcite-calcite-tab:not([calcite-hydrated]),
calcite-calcite-tabs:not([calcite-hydrated]),
calcite-calcite-tile-select:not([calcite-hydrated]),
calcite-calcite-tile-select-group:not([calcite-hydrated]),
calcite-calcite-tooltip-manager:not([calcite-hydrated]),
calcite-calcite-tree:not([calcite-hydrated]),
calcite-calcite-tree-item:not([calcite-hydrated]),
calcite-calcite-value-list:not([calcite-hydrated]),
```

support/hydrationStyles.ts
```diff
@@ -0,0 +1,15 @@
+const fs = require("fs");
+const path = require("path");
+
+(async function () {
+  const stylePath = path.normalize(`${__dirname}/../src/assets/styles/_hydration.scss`);
+  const components = await fs.promises.readdir(path.normalize(`${__dirname}/../src/components/`));
+  const selectors = components.map((component) => `calcite-${component}:not([calcite-hydrated])`);
```
driskull, line 7: Something like this would be nice...

```ts
const hiddenComponents = new Set<string>(["alert", "input-message", "popover", "tooltip", "tree-item"]);

(async function () {
  const stylePath = path.normalize(`${__dirname}/../src/assets/styles/_hydration.scss`);
  const components = await fs.promises.readdir(path.normalize(`${__dirname}/../src/components/`));
  const selectors = components.map((component) => `calcite-${component}:not([calcite-hydrated])`);
  const selectors = components
    .filter((component) => !hiddenComponents.has(component))
    .map((component) => `calcite-${component}:not([calcite-hydrated])`);
```

paulcpederson:
@driskull So after playing with this a bit, even a raw selector like `calcite-alert {}` will override styles you put in the host stylesheet, so the solutions still end up being pretty gross. The calcite-hydrated-hidden attribute is actually looking a lot cleaner than the alternatives so it may be better to leave it.

support/hydrationStyles.ts
paulcpederson: I tried that and got thrown an error about using import outside a module?

support/hydrationStyles.ts
paulcpederson: Yeah, I just noticed a couple other ones, let me look at this a little closer

support/hydrationStyles.ts
paulcpederson: this would flash unstyled content prior to javascript running

support/hydrationStyles.ts
paulcpederson: ok, checking for only directories now, which should be more accurate.

support/hydrationStyles.ts
paulcpederson: ooooh, ok that makes sense, I can make that change once that pr goes in.

src/assets/styles/includes.scss
```diff
@@ -7,6 +7,7 @@
 @import "host";
 @import "spacing";
 @import "popper";
+@import "hydration";
```
jcfranco, line 10: Should this import go in `globals.scss` or are they here so that they get bundled with each component (e.g., so styling applies when wrapped in other components)? Currently, it looks like the styles are built into both the global CSS file and each component.

driskull (commented):
@paulcpederson I'm still seeing some modifications on OSX.

```diff
diff --git a/src/assets/styles/_hydration.scss b/src/assets/styles/_hydration.scss
index 889c140f0cfc3a198e2c44823b304f523d1939c2..8685c9fe6e945e9b8af25fe029255019cdf1605c 100644
--- a/src/assets/styles/_hydration.scss
+++ b/src/assets/styles/_hydration.scss
@@ -11,6 +11,45 @@ calcite-avatar:not([calcite-hydrated]),
 calcite-block:not([calcite-hydrated]),
 calcite-block-section:not([calcite-hydrated]),
 calcite-button:not([calcite-hydrated]),
+calcite-calcite-accordion:not([calcite-hydrated]),
+calcite-calcite-action-menu:not([calcite-hydrated]),
+calcite-calcite-alert:not([calcite-hydrated]),
+calcite-calcite-block-section:not([calcite-hydrated]),
+calcite-calcite-button:not([calcite-hydrated]),
+calcite-calcite-chip:not([calcite-hydrated]),
+calcite-calcite-color:not([calcite-hydrated]),
+calcite-calcite-color-picker:not([calcite-hydrated]),
+calcite-calcite-combobox:not([calcite-hydrated]),
+calcite-calcite-date-picker:not([calcite-hydrated]),
+calcite-calcite-dropdown:not([calcite-hydrated]),
+calcite-calcite-dropdown-group:not([calcite-hydrated]),
+calcite-calcite-fab:not([calcite-hydrated]),
+calcite-calcite-graph:not([calcite-hydrated]),
+calcite-calcite-icon:not([calcite-hydrated]),
+calcite-calcite-input:not([calcite-hydrated]),
+calcite-calcite-input-date-picker:not([calcite-hydrated]),
+calcite-calcite-input-message:not([calcite-hydrated]),
+calcite-calcite-label:not([calcite-hydrated]),
+calcite-calcite-list:not([calcite-hydrated]),
+calcite-calcite-list-item:not([calcite-hydrated]),
+calcite-calcite-modal:not([calcite-hydrated]),
+calcite-calcite-pagination:not([calcite-hydrated]),
+calcite-calcite-panel:not([calcite-hydrated]),
+calcite-calcite-pick-list:not([calcite-hydrated]),
+calcite-calcite-popover:not([calcite-hydrated]),
+calcite-calcite-popover-manager:not([calcite-hydrated]),
+calcite-calcite-radio-group:not([calcite-hydrated]),
+calcite-calcite-shell:not([calcite-hydrated]),
+calcite-calcite-shell-panel:not([calcite-hydrated]),
+calcite-calcite-stepper:not([calcite-hydrated]),
+calcite-calcite-tab:not([calcite-hydrated]),
+calcite-calcite-tabs:not([calcite-hydrated]),
+calcite-calcite-tile-select:not([calcite-hydrated]),
+calcite-calcite-tile-select-group:not([calcite-hydrated]),
+calcite-calcite-tooltip-manager:not([calcite-hydrated]),
+calcite-calcite-tree:not([calcite-hydrated]),
+calcite-calcite-tree-item:not([calcite-hydrated]),
+calcite-calcite-value-list:not([calcite-hydrated]),
 calcite-card:not([calcite-hydrated]),
 calcite-checkbox:not([calcite-hydrated]),
 calcite-chip:not([calcite-hydrated]),
@@ -27,6 +66,7 @@ calcite-date-picker-month-header:not([calcite-hydrated]),
 calcite-dropdown:not([calcite-hydrated]),
 calcite-dropdown-group:not([calcite-hydrated]),
 calcite-dropdown-item:not([calcite-hydrated]),
+calcite-dropdown-menu:not([calcite-hydrated]),
 calcite-fab:not([calcite-hydrated]),
 calcite-filter:not([calcite-hydrated]),
 calcite-flow:not([calcite-hydrated]),
@@ -88,8 +128,9 @@ calcite-tooltip:not([calcite-hydrated]),
 calcite-tooltip-manager:not([calcite-hydrated]),
 calcite-tree:not([calcite-hydrated]),
 calcite-tree-item:not([calcite-hydrated]),
+calcite-utils:not([calcite-hydrated]),
 calcite-value-list:not([calcite-hydrated]),
 calcite-value-list-item:not([calcite-hydrated]) {
   visibility: hidden;
   pointer-events: none;
-}
+}
\ No newline at end of file
```

driskull:
Maybe we could use the docs:json or stats to generate the styles?

https://stenciljs.com/docs/docs-json
https://stenciljs.com/docs/stats

jcfranco:
> Maybe we could use the docs:json or stats to generate the styles?

This would be great, but but it would require 2 builds. Let's proceed with the current approach and we can revisit later if we notice any funkiness. 🎶🎸🎶

src/assets/styles/includes.scss
paulcpederson: no, it should only be needed in the global file, good catch

paulcpederson:
@driskull going to experiment with using a custom target, if that requires two builds I can just add a filter for these old folders

paulcpederson:
Should we gitignore the generated scss file? Seems like the sass files in our project are not distributed as part of a release, so shouldn't be an issue to ignore it as it's generated each build anyways.

jcfranco:
Good question. I think we can ignore it.

jcfranco (approved):
🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘
🤘🎸🎸🎸🎸🤘🎸🤘🤘🤘🤘🤘🎸🤘🎸🎸🎸🎸🤘🤘🎸🎸🎸🤘🤘🎸🎸🤘🤘🎸🤘🤘🤘🎸🤘🎸🎸🎸🎸🤘
🤘🎸🤘🤘🎸🤘🎸🤘🤘🤘🤘🤘🎸🤘🎸🤘🤘🤘🤘🎸🤘🤘🤘🤘🎸🤘🤘🎸🤘🎸🎸🤘🎸🎸🤘🎸🤘🤘🤘🤘
🤘🎸🎸🎸🎸🤘🎸🤘🤘🎸🤘🤘🎸🤘🎸🎸🎸🤘🤘🤘🎸🎸🤘🤘🎸🤘🤘🎸🤘🎸🤘🎸🤘🎸🤘🎸🎸🎸🤘🤘
🤘🎸🤘🤘🎸🤘🎸🤘🎸🤘🎸🤘🎸🤘🎸🤘🤘🤘🤘🤘🤘🤘🎸🤘🎸🤘🤘🎸🤘🎸🤘🤘🤘🎸🤘🎸🤘🤘🤘🤘
🤘🎸🤘🤘🎸🤘🤘🎸🤘🤘🤘🎸🤘🤘🎸🎸🎸🎸🤘🎸🎸🎸🤘🤘🤘🎸🎸🤘🤘🎸🤘🤘🤘🎸🤘🎸🎸🎸🎸🤘
🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘

paulcpederson:
@driskull can you try this one more time. I've added a little heuristic to filter out non-component files and folders, so I think this should be pretty accurate now, even for devs with older cached directories

driskull (approved):
Looks good to me!

Issue #3955, opened 2022-01-19
Action-bar - Arabic text is is cut off in Firefox


zaramatheson:
### Actual Behavior

Arabic text is cut off in Map Viewer in Firefox only. 
![image](https://user-images.githubusercontent.com/50747646/150042299-4a200e0d-88a8-43b9-b773-cbf87eaaf900.png)

### Expected Behavior

Text should not be cut off:
![image](https://user-images.githubusercontent.com/50747646/150042364-2a914776-c2cb-445e-9436-8d96439ce09e.png)

### Reproduction Sample

https://www.arcgis.com/apps/mapviewer/index.html?locale=ar

Codepen sample: https://codepen.io/mtb-alan/pen/LYzvMgg

1. If you don't see the descenders clipping, inspect the Action and look for the `.button` element.
2. Set `font-family` to `revert`

### Reproduction Steps

1. Open https://www.arcgis.com/apps/mapviewer/index.html?locale=ar in Firefox and expand the toolbars to see the text. 

### Reproduction Version

@esri/calcite-components@1.0.0-VERSION

### Relevant Info

Regression? Last working version: `@esri/calcite-components@1.0.0-VERSION`

### Source

- [ ] CDN
- [ ] NPM package

cc @asangma @driskull

asangma:
@benelan I can take this on.
I added a Codepen and some repro steps.

asangma:
@benelan I made a PR but then realized this issue doesn't have a milestone.

benelan:
No worries, it happens.

jcfranco:
@asangma Is this still valid? I didn't see a difference in FF when comparing the codepen between its version (beta.65) and the latest.

zaramatheson:
@jcfranco @asangma This is still an issue in MV using beta.77

AdelheidF:
beta.77 was released Feb 14, 2022. This PR got merged Jan 28. So I assume this is still an issue with the latest beta.80.

AdelheidF:
still an issue with beta.81, @zara checked

jcfranco:
[...]

PR #4857, opened 2022-07-06
fix(action): updates line-height and affected y padding

asangma:
**Related Issue:** (#3955)

## Summary
Updates line-height and reduces y-padding to accommodate increased height of text container.


macandcheese:
LGTM - how can we communicate these changes to Figma design team (not unique to this, but any issue that could affect Figma instance)? 

@asangma we don't need this change on scale L?

asangma:
@macandcheese I think the sizing should be the same in Figma. 

re: scale L
It appears that the increased line-height isn't affecting the height of large actions.
In M and S the text container increases in size, hence the block padding updates for those scales.

asangma:
> Could this issue be reproduced in our storybook? 

@jcfranco I believe the issue would occur for Arabic. So I would add a test with the language set as such yeah?

jcfranco:
> So I would add a test with the language set as such yeah?

Correct. [Here’s](https://github.com/Esri/calcite-components/blob/master/src/components/input-time-picker/input-time-picker.stories.ts#L71) an example.

asangma:
Thanks @jcfranco

eriklharper:
@asangma that's such a miniscule diff that I would ignore that for now and just accept the changes.  From the looks of it, neither screenshot looks ideal since the hyphen appears to close to the left-most number, so it is probably worth taking a look at why that's specifically occurring in slider in a separate issue.  It could also be just present on Chrome 87 and maybe it looks fine in the latest browser versions too.

asangma:
Hurm...the e2e failure seems to be coming from Filter?

```
calcite-filter › filter behavior › updates filtered items after filtering
```

PR #4869, opened 2022-07-07
docs: api style consistency for L-P and some T-Z components

geospatialem:
**Related Issue:** #4442 

## Summary
1. Added api style consistency to: 
    - `option`
    - `option-group`
    - `pagination`
    - `panel`
    - `pick-list`
    - `pick-list-group`
    - `pick-list-item`
    - `popover`
2. Also includes minor consistency modifications to previously addressed components based on team discussion and decision(s):
    - `list`
    - `list-item`
    - `list-item-group`
    - `loader`
    - `modal`
    - `notice`
    - `tooltip`
    - `tree`
    - `tree-item`
    - `value-list`
    - `value-list-item`


src/components/list-item/list-item.tsx
```diff
@@ -28,22 +28,22 @@ export class ListItem implements ConditionalSlotComponent, InteractiveComponent
...
+   * When true, prevents the content of the component from user interaction.
    */
   @Prop({ reflect: true }) nonInteractive = false;
 
   /**
-   * An optional description for the list item. Displays below the label text.
+   * A description for the component. Displays below the label text.
    */
-  @Prop() description: string;
+  @Prop() description?: string;
```
driskull, line 38: nice catch.

src/components/option-group/option-group.tsx
```diff
@@ -1,7 +1,7 @@
 import { Component, h, Prop, VNode, Watch, Event, EventEmitter, Fragment } from "@stencil/core";
 
 /**
- * @slot - A slot for adding `calcite-option`s.
+ * @slot - A slot for adding `calcite-option` elements.
```
driskull, line 4: Should we refer to `calcite-option` as an element, custom element, or component? I wonder what the best wording would be for referencing other components. Is it as elements? In properties, we reference the current element as a component.

src/components/panel/panel.tsx
```diff
@@ -25,10 +25,10 @@ import { createObserver } from "../../utils/observers";
 
 /**
  * @slot - A slot for adding custom content.
- * @slot header-actions-start - A slot for adding actions or content to the start side of the panel header.
- * @slot header-actions-end - A slot for adding actions or content to the end side of the panel header.
+ * @slot header-actions-start - A slot for adding actions or content to the start side of the header.
+ * @slot header-actions-end - A slot for adding actions or content to the end side of the header.
  * @slot header-content - A slot for adding custom content to the header.
- * @slot header-menu-actions - A slot for adding an overflow menu with actions inside a dropdown.
+ * @slot header-menu-actions - A slot for adding an overflow menu with actions inside a `calcite-dropdown`.
```
driskull, line 31: Here we just say `calcite-dropdown` and not element or component.

src/components/panel/panel.tsx
```diff
@@ -180,24 +180,24 @@ export class Panel implements ConditionalSlotComponent, InteractiveComponent {
...
 
   /**
-   * Emitted when the content has been scrolled.
+   * Emits when the content has been scrolled.
    */
   @Event() calcitePanelScroll: EventEmitter;
 
   /**
-   * Emitted when the back button has been clicked.
+   * Emits when the back button has been clicked.
```
driskull, line 200: I'm fine with Emits but just wanted to point out MDN.

> The click event is raised when the user clicks on an element. It fires after the [mousedown](https://developer.mozilla.org/en-US/docs/Web/API/Element/mousedown_event) and [mouseup](https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseup_event) events, in that order.

> change events fire when the user commits a value change to a form control. This may be done, for example, by clicking outside of the control or by using the Tab key to switch to a different control.

Seems like they use "is raised" or "fire/s"


> The keydown event fires when the user presses a keyboard key.

src/components/pick-list-group/pick-list-group.tsx
```diff
@@ -25,13 +25,13 @@ export class PickListGroup implements ConditionalSlotComponent {
   // --------------------------------------------------------------------------
 
   /**
-   * The title used for all nested `calcite-pick-list` rows.
+   * Specifies the title for all nested `calcite-pick-list` rows.
```
driskull, line 28: Maybe "rows" should be "calcite-pick-list-item"s? or just "items"?

src/components/pick-list-item/pick-list-item.tsx
```diff
@@ -111,14 +111,14 @@ export class PickListItem implements ConditionalSlotComponent, InteractiveCompon
...
-   * Used as an accessible label (aria-label) for the "remove item" action. Only applicable if removable is true.
+   * Accessible name for the component's remove button. Only applicable if removable is "true".
    *
    * @default "Remove"
    */
   @Prop({ reflect: true }) intlRemove = TEXT.remove;
 
   /**
-   * The item's associated value.
+   *  The component's value.
```
driskull, line 121: maybe an extra space here.

src/components/pick-list-item/pick-list-item.tsx
```diff
@@ -202,7 +202,7 @@ export class PickListItem implements ConditionalSlotComponent, InteractiveCompon
   // --------------------------------------------------------------------------
 
   /**
-   * Used to toggle the selection state. By default this won't trigger an event.
+   * Toggle's the selection state. By default this won't trigger an event.
```
driskull, line 205: Should this be "Toggles" instead of "Toggle's"?

src/components/pick-list/pick-list.tsx
```diff
@@ -210,7 +210,7 @@ export class PickList<
   //
   // --------------------------------------------------------------------------
 
-  /** Returns the currently selected items */
+  /** Returns the component's selected item(s). */
```
driskull, line 213: Should we say item(s) or `calcite-pick-list-item`s?

driskull (approved):
Awesome work. Consistency!!!!

src/components/option-group/option-group.tsx
driskull: To be consistent with other text, maybe just `calcite-option`s.

geospatialem: Was going back and forth on this one, your recommendation looks best. Going to add this in to our consistency docs as well. 😃

src/components/panel/panel.tsx
geospatialem: That makes sense, going to add some click events language to our consistency docs. Also changed a few other pages in this PR that were related, all to "Fires", it seemed more clear than "Is raised", or "Raises".

src/components/pick-list-item/pick-list-item.tsx
geospatialem: Good 👁️👁️, thanks for spotting that one!

Issue #4817, opened 2022-06-28
flow broken with custom component panel


AdelheidF:
### Actual Behavior

Flow doesn't work if one of the panels is inside a custom component. It instead displays the 2 panels at the same time.

This was fixed for beta.81 (or around that time) and now is broken again in beta.83.

cc @matt

### Expected Behavior

only one panel inside the flow is visible at any time

### Reproduction Sample

https://codepen.io/afreitag/pen/gOoEJza

### Reproduction Steps

1. open test app, click back button. All looks good with beta.82
2. change to beta.83

### Reproduction Version

beta.83

### Regression?

beta.82

driskull:
Hey @AdelheidF,

I think this one is actually expected.

We changed the component to only count panels slotted into the flow in order to fix #4667. So now it only honors directly slotted children. See https://github.com/Esri/calcite-components/issues/4667#issuecomment-1144101986

This is the same relationship that some native elements have like `<ul><li></li></ul>` whereas `<ul><div><li></li></div></ul>` is not allowed.

When we fixed the issue above, I don't think it is a breaking change since since we never documented it to work where other elements could be in-between the flow/panel relationship in any of our samples and doing so may actually cause issues if you didn't intend the panel to be counted as a flow child. @jcfranco or @benelan let me know if you think we should document that change in more detail or mark it as a breaking change.

I think in your use case, you might be able to use a functional component to handle common configuration of a `calcite-panel`. If we allow any number of elements to be placed in-between what we expect, any number of things could go wrong with styling or functionality.

@jcfranco Let me know if you disagree with any of the above.

AdelheidF:
found the issue that we discussed this before in: https://github.com/Esri/calcite-components/issues/2180

jcfranco:
> I don't think the change is a breaking change since since we never documented it to work where other elements could be in-between the calcite-flow/calcite-panel relationship. None of our samples had that and it may actually cause issues if you didn't intend the calcite-panel to be counted as a calcite-flow child.

Agreed.

> let me know if you think we should document that change in more detail 

Let's add more info to the change.

WRT the original request, it would be useful to provide support for custom panel elements (CPEs) as this would be a common scenario. Here's a quick prototype for CPEs implementing an interface that flow could use: https://github.com/Esri/calcite-components/tree/jcfranco/4817-proto-custom-panel-component-support. This approach assumes custom panel elements adhere to whatever we define as a panel, so the flow-panel relationship can be honored. 
[...]

PR #4872, opened 2022-07-07
fix(flow): Do not allow nested panels when allowDescendantPanels is true

driskull:
**Related Issue:** #4817

## Summary

fix(flow): Do not allow nested panels when allowDescendantPanels is true


jcfranco (approved):
💥

PR #4875, opened 2022-07-08
refactor(flow): Use querySelectorAll to query panels.

driskull:
**Related Issue:** #4817

## Summary

refactor(flow): Use querySelectorAll to query panels.


jcfranco (approved):
The changelog appears to be outdated, I don't see

> flow: Do not allow nested panels when allowDescendantPanels is true (https://github.com/Esri/calcite-components/issues/4872) ([840b2f1](https://github.com/Esri/calcite-components/commit/840b2f1a7e522ee515c8d4c64bbd9707d61f0538))

Should it be updated to indicate the latest behavior?

driskull:
I'll remove changelog edits. We can fix it as necessary.

src/components/flow/flow.tsx
```diff
@@ -120,33 +98,13 @@ export class Flow {
...
   handleMutationObserverChange = (): void => {
-    const { el, allowDescendantPanels } = this;
-
-    if (!allowDescendantPanels) {
-      return;
-    }
-
     const newPanels: HTMLCalcitePanelElement[] = Array.from(
-      el.querySelectorAll("calcite-panel:not(calcite-panel calcite-panel)")
+      this.el.querySelectorAll("calcite-panel:not(calcite-panel calcite-panel)")
```
eriklharper, line 102: Will this query still work if you have a 3rd nested `calcite-panel` like `(calcite-panel calcite-panel calcite-panel)` or even beyond that?

driskull: Wouldn't that fall under `:not(calcite-panel calcite-panel)`?

src/components/flow/flow.stories.ts
```diff
@@ -38,8 +38,8 @@ const createFlowItemAttributes: (group: string) => Attributes = (group) => {
       value: boolean("menuOpen", false, group)
     },
     {
-      name: "summary",
-      value: text("summary", "Summary", group)
+      name: "description",
+      value: text("description", "Description", group)
```
driskull, line 42: @anveshmekala I updated the flow story to set the panel to use `description` instead of summary. Any other stories we should check to make sure they aren't using the deprecated props?

src/components/flow/flow.tsx
```diff
@@ -119,44 +97,21 @@ export class Flow {
...
-      return;
-    }
-
-    const newPanels = (event.target as HTMLSlotElement)
-      .assignedElements({
-        flatten: true
-      })
-      .filter((el) => el?.matches("calcite-panel")) as HTMLCalcitePanelElement[];
+      el.querySelectorAll("calcite-panel")
+    ).filter((panel) => !panel.closest("calcite-panel")) as HTMLCalcitePanelElement[];
```
driskull, line 105: @jcfranco storybook kept failing when using `el.querySelectorAll("calcite-panel:not(calcite-panel calcite-panel)")`.

I'm guessing this is because its stuck on a version that doesn't support it?

https://developer.mozilla.org/en-US/docs/Web/CSS/:not#browser_compatibility

I'd rather use the single selector if possible but storybook is holding us back here.

driskull: cc @eriklharper

src/components/flow/flow.tsx
```diff
@@ -119,44 +97,21 @@ export class Flow {
...
-      return;
-    }
-
-    const newPanels = (event.target as HTMLSlotElement)
-      .assignedElements({
-        flatten: true
-      })
-      .filter((el) => el?.matches("calcite-panel")) as HTMLCalcitePanelElement[];
+      el.querySelectorAll("calcite-panel")
+    ).filter((panel) => !panel.matches("calcite-panel calcite-panel")) as HTMLCalcitePanelElement[];
```
driskull, line 105: https://github.com/Esri/calcite-components/pull/4875#discussion_r917120040

src/components/flow/flow.tsx
jcfranco: > I'd rather use the single selector if possible but storybook is holding us back here.

Yeah, that'd be cleaner. For now, let's proceed with this tested version and we should create a refactor issue to make the switch once newer browsers are used for screenshot tests.

src/components/flow/flow.stories.ts
jcfranco: This should be in a separate PR as it's unrelated to the overall changes.

src/components/flow/flow.tsx
```diff
@@ -119,44 +97,21 @@ export class Flow {
...
-    const { panels } = this;
     const oldPanelCount = panels.length;
     const newPanelCount = newPanels.length;
     const activePanel = newPanels[newPanelCount - 1];
     const previousPanel = newPanels[newPanelCount - 2];
 
     if (newPanelCount && activePanel) {
       newPanels.forEach((panelNode) => {
-        panelNode.showBackButton = newPanelCount > 1;
+        panelNode.showBackButton = panelNode === activePanel && newPanelCount > 1;
```
jcfranco, line 114: Is this a separate fix? It seems unrelated to the panel query behavior.

jcfranco (approved):
💥

src/components/flow/flow.stories.ts
driskull: Ok. I'll revert.

src/components/flow/flow.tsx
driskull: yeah ill separate it out.

src/components/flow/flow.tsx
eriklharper: Yeah, unfortunately we'll have to just deal with the outdated Chrome for now until we can make the switch to Chromatic.

Issue #1947, opened 2021-04-12
Enhancement: input-date-picker sates


joeyHarig:
### Description
Add `disabled` and `readonly` states to the `calcite-input-date-picker` component. 

### Acceptance Criteria 
- `disabled` should disable the input and any pointer events
- `readonly` should set `readonly` on the input to true and also disable pointer events on the input

#### Which Component
- `calcite-input-date-picker`

#### Example Use Case
Disabled: When the date-input within a form is waiting on other inputs and is not yet enabled
Readonly: When the date has already been set and can not be changed.

driskull:
@paulcpederson @eriklharper @macandcheese We don't have `readonly` on `calcite-input` so we would need to add that first, correct? Is there a reason we don't have it currently?

paulcpederson:
No, we should probably have `readonly` for inputs to maintain API parity with native HTML.

macandcheese:
Yeah agree.. right now you can add readonly and the styles will correctly apply, and the attribute will render on the child input element... however it’s not defined as a custom prop - it’s just getting spread if present - so it’s not getting documented.

Should add that so it’s on the calcite input and to the point of the issue probably anything that wraps one like a date, or time input.

eriklharper:
I'm in favor of removing spread attributes as a general principle.  Best to always explicitly define the attributes being used.

driskull:
> I'm in favor of removing spread attributes as a general principle. Best to always explicitly define the attributes being used and supported.

Yeah, we shouldn't be passing everything into the internals of the component.

jcfranco:
The disabled prop support was handled by #2655, so I'll repurpose this issue to add read-only support to the input date/time pickers. cc @benelan

jcfranco:
@benelan can we squeeze this one for this release?

benelan:
> @benelan can we squeeze this one for this release?

Sure, triaged!

jcfranco:
@Elijbet Can you look at this one for this release?

Elijbet:
> @Elijbet Can you look at this one for this release?

Sure thing.

PR #4876, opened 2022-07-08
fix(input, input-date-picker): added readOnly support to input-date-picker and fixed related issues

Elijbet:
**Related Issue:** #1947

## Summary

WIP: I will be addressing the feedback while splitting this issue into smaller PRs. 

- fixed disabled buttons on `calcite-input` to stop interacting: to not nudge the number, transition, or respond to pointer events
- modified the documentation to reflect the difference between `disabled` and `readOnly`: `readOnly` input is still focusable but won’t display the controls or allow for changing the value, whereas `disabled` controls can not receive focus and generally do not function as controls until they are enabled, refer to [MDN here]( https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly#:~:text=The%20difference%20between%20disabled%20and%20readonly%20is%20that%20read%2Donly%20controls%20can%20still%20function%20and%20are%20still%20focusable%2C%20whereas%20disabled%20controls%20can%20not%20receive%20focus%20and%20are%20not%20submitted%20with%20the%20form%20and%20generally%20do%20not%20function%20as%20controls%20until%20they%20are%20enabled.)
- fixed `readOnly` to not nudge the input and remove the display of controls, while remaining focusable
- refactor: `<calcite-input-date-picker>` demo example aggregated into its top-level page `demos/calcite-input-date-picker`
-  enabled passing down of `disabled` and `readOnly` attributes from top-level `<calcite-input-date-picker>` to `<calcite-input>` to native `<input>`
- refactor: grouped `input` tests in `describe disabled` and `describe readOnly` blocks 
- suggested documenting the correct format for the attribute as `read-only` and not `readonly` as a helpful reminder


src/components/input/input.e2e.ts
```diff
@@ -76,7 +76,78 @@ describe("calcite-input", () => {
       }
     ]));
 
-  it("can be disabled", () => disabled("calcite-input"));
+  describe("calcite-input when disabled", () => {
+    it("can be disabled", () => disabled("calcite-input"));
+  });
+
+  describe("calcite-input when readOnly", () => {
+    it("sets internals to readOnly or disabled when readOnly is true", async () => {
```
jcfranco, line 84: Prefer testing behavior and not the internals. This will prevent the test from breaking if we ever refactored it. Since you already do this in the following tests, I think this one can be removed.

Elijbet:
Closing this issue, as the parts that are relevant have been pulled into other smaller PRs.

Issue #4873, opened 2022-07-07
Radio Group - Selected item is not selected in High Contrast mode


mpriour:
### Summary

The lack of visual information for selected items was identified as a level 3 severity issue (missing critical information) in the [High Contrast Audit](https://github.com/Esri/calcite-components/wiki/High-Contrast-Audit#radiogroup). The problem is that there is no indication of what item is selected in a radio group component when using High Contrast mode.

### Actual Behavior

High Contrast Radio Group with selected items using both solid and outline form with and without icons

<img height=400 src="https://user-images.githubusercontent.com/142636/149631506-ada043e6-b991-4764-bc4e-0da2eb0dc975.png">

### Expected Behavior

Selected item in the Radio Group should be visually obvious even in High Contrast mode.

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/radio-group/

### Reproduction Steps

Windows
1. Set Windows color theme to a high-contrast variant.
2. Open the reproduction sample (https://developers.arcgis.com/calcite-design-system/components/radio-group/) in any browser.
3. Observe the issue.

Mac
1. Open the reproduction sample (https://developers.arcgis.com/calcite-design-system/components/radio-group/) in Chrome or Edge.
2. Open Dev Tools.
3. Click on Elements tab.
4. Next to the console sub tab, there should be a "Rendering" tab (if not you will need to enable it)
5. On the Rendering tab choose Forced-colors: active from the "Emulate forced colors..." selector
6. Observe the issue.

### Reproduction Version

latest

geospatialem:
Verified on next ✔️ 
![high contrast verification](https://user-images.githubusercontent.com/5023024/178078139-68dfbaf9-94e3-4816-b6ac-d9bb9ae781b7.png)

PR #4877, opened 2022-07-08
fix: (radio group) - apply forced-colors styles

mpriour:
* Adds highlight color to solid selected item
* Adds bold outline to outline selected item
* Modifies icon color on solid selected item for better contrast

**Related Issue:** #4873 

## Summary

Modifies the radio group selections to look like this on High Contrast mode:

Screen Shot of high contrast radio group selection states

Unfortunately, the text *ALWAYS* has its own un-style-able background so it doesn't have the same background as the radio group "button".

### DIFF
--- a/src/components/radio-group-item/radio-group-item.scss
+++ b/src/components/radio-group-item/radio-group-item.scss
@@ -72,6 +72,18 @@
   @apply hidden;
 }
 
+@media (forced-colors: active) {
+  :host([checked]) label {
+    background-color: highlight;
+  }
+  :host([checked]) .label--outline {
+    @apply outline-none;
+  }
+  :host([checked]) label:not([class~="label--outline"]) .radio-group-item-icon {
+    color: highlightText;
+  }
+}
+
 // icon
 .radio-group-item-icon {
   @apply relative


mpriour:
@macandcheese Now that all the tests passed, can you take a look at this to see if you have any further comments or requests on the change for high contrast users.

driskull (approved):
LGTM!

macandcheese:
Thanks @mpriour - is "outline" strong enough? Not sure if using the highlight color for that outline could help.

I guess it's worth considering if this visual difference is needed for high contrast users, or if there are concerns with "outline", display both as "solid" when in this mode?

mpriour:
@macandcheese I can easily increase the outline thickness. It is 2 pixels now but 4 pixels would be more obvious. I will get some feedback from a real high contrast user at least as well.

mpriour:
@macandcheese Shared changes with a real high contrast user and they thought that the 2px outline was perfect. We experimented with 3px and 4px and both were determined to be too thick.

macandcheese:
> 

Awesome, thanks for checking.

PR #4881, opened 2022-07-08
ci(create-milestone): remove label event used for testing

benelan:
**Related Issue:** NA

## Summary
I used the label event to make sure the action worked on my test repo, and then forgot to move it here 😅


benelan:
pushing this through since it is effecting our production CI (making new milestones every time an issue is labeled or unlabeled)
cc @jcfranco

Issue #4879, opened 2022-07-08
[Input] disabled spinner buttons on Input type=number should not be interactive


Elijbet:
### Actual Behavior

Disabled spinner buttons on the Input (type=number) respond to pointer events, nudge the number, and transition.

### Expected Behavior

Disabled spinner buttons on the Input (type=number) should not be interactive.

### Reproduction Sample

https://codepen.io/elijbet/pen/jOzqZjv

### Reproduction Steps

Click on the up/down spinner buttons to see them highlight and nudge the number.

### Reproduction Version

v1.0.0-beta.83

### Relevant Info

Seems to appear on the new Input-number component PR #[4870](https://github.com/Esri/calcite-components/pull/) as well.

geospatialem:
Verified on `beta.86`

![verify-input-disabled-spinbtns-interaction](https://user-images.githubusercontent.com/5023024/179016973-1d4cb1d8-12ed-4fdb-943a-1345d46cd57a.gif)

PR #4870, opened 2022-07-07
feat(input-number): create separate component for input type number

benelan:
**Related Issue:** part of the roadmap item to separate `calcite-input` types
 
## Summary
Created a new `calcite-input-number` component by copying over input and removing all non-number functionality.

PR #4884, opened 2022-07-08
fix(Input): type number spinner buttons will no longer interact when disabled

Elijbet:
**Related Issue:** #4879, PR #4870

## Summary

Added:
- `:disabled` pseudo-class to `.number-button-item`, set to `pointer-events-none` 
- a backup check for the `disabled` prop before the number value can be nudged
- test to make sure `disabled` stepper buttons should not be interactive/nudge the number


driskull (approved):
👍 ☑️ ✅

PR #4888, opened 2022-07-08
fix(input-date-picker): added readOnly prop support

Elijbet:
**Related Issue:** #1947

## Summary

- added `readOnly` prop support
- modified the documentation to reflect the difference between `disabled` and `readOnly`: 

> `readOnly input` is still focusable but won’t display the controls or allow for changing the value, whereas `disabled` controls can not receive focus and generally do not function as controls until they are enabled, refer to [MDN here](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly#:~:text=The%20difference%20between%20disabled%20and%20readonly%20is%20that%20read%2Donly%20controls%20can%20still%20function%20and%20are%20still%20focusable%2C%20whereas%20disabled%20controls%20can%20not%20receive%20focus%20and%20are%20not%20submitted%20with%20the%20form%20and%20generally%20do%20not%20function%20as%20controls%20until%20they%20are%20enabled.)

-  add a test to ensure when set to `readOnly`, the element is still focusable but won’t display the controls or allow for changing the value


jcfranco (approved):

PR #4890, opened 2022-07-08
build: switch project module type to ESM

jcfranco:
**Related Issue:** https://github.com/Esri/calcite-components/pull/4700

## Summary

Changes `package.json`'s type to "module" for ESM support.

Manually tested the following:

* a couple of support scripts
* Stencil production build
* Stencil test run
* Storybook preview (see note below)

This could use additional testing before merging.

cc @driskull @benelan 

### Notes

* There's an [issue w/ Storybook when used inside an ESM project](https://github.com/storybookjs/storybook/issues/9854), so we use a dummy `package.json` to force it to use CJS.
* Updates support scripts to be ESM
* Bumps `pify` and `chalk`
* Removes `chalk` from the dep update block list


driskull (approved):
This looks good to me. 👍 

We can always revert if we run into issues.

PR #4891, opened 2022-07-09
chore(storybook): update static dir option

jcfranco:
**Related Issue:** N/A

## Summary

This fixes the deprecation warning from Storybook:

```
info @storybook/html v6.5.6
info
(node:21979) DeprecationWarning: --static-dir CLI flag is deprecated, see:

https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated---static-dir-cli-flag
```

cc @benelan


benelan (approved):
LGTM!

**Note:** this PR will cause the Storybook deployment action to break because it will need to use the `stencil.storybook.config.ts` file again. I will re-troubleshoot the action the next time it runs. Relevant issue: #4451

Issue #4865, opened 2022-07-07
`popover` content not read back to screen reader users on demand


geospatialem:
### Summary

When accessed via a reference element `popover` content is not read back to screen reader users on demand.

### Actual Behavior

The `popover` content is not read back to screen reader users when accessed via the reference element. 

*However, as a workaround, users can press the down arrow to read back the content, but this is not expected behavior for screen reader users.*

### Expected Behavior

Upon engaging the `popover` from the reference element, content should be provided politely, perhaps using the `aria-role='status'`, to provide context to screen reader users.

Similar to the current functionality, after completion, the focus should return to the reference element.

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/popover/#sample

### Reproduction Steps

1. Enable a screen reader (NVDA, JAWS or VoiceOver)
2. Navigate to the [sample](https://developers.arcgis.com/calcite-design-system/components/popover/#sample)
3. With the screen reader active, navigate via keyboard to focus on the "Activate popover" button to trigger the popover
4. Observe no context is provided to the screen reader initially.
    - However, if using the down arrow key, context will be read back to the user

### Reproduction Version

beta.83

### Relevant Info

- Windows (JAWS and NVDA)
- Mac OS (Safari/VoiceOver)
- related to: #4864

driskull:
@geospatialem popover already has a role of `role="dialog"`. Do you know what should be added exactly?

geospatialem:
> @geospatialem popover already has a role of `role="dialog"`. Do you know what should be added exactly?

@driskull thanks for taking a peek at this issue! This will be the same resolution as the `tooltip` screen reader fix in adding an `aria-live="polite"` for JAWS, NVDA, and VoiceOver.

benelan:
verified on master

PR #4894, opened 2022-07-11
fix(popover): Add aria-live="polite" for screen readers

geospatialem:
**Related Issue:** #4865

## Summary
Fix enables `popover` to receive context for screen readers. 

Prior to the fix screen readers would only receive the popover content when using the down arrow key.

The [`aria-live="polite"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) will provide the popover content politely (once the screen reader reads through all the context)

### DIFF
--- a/src/components/popover/popover.tsx
+++ b/src/components/popover/popover.tsx
@@ -500,6 +500,7 @@ export class Popover {
       <Host
         aria-hidden={toAriaBoolean(hidden)}
         aria-label={label}
+        aria-live="polite"
         calcite-hydrated-hidden={hidden}
         id={this.getId()}
         role="dialog"


jcfranco (approved):
Noice! 

Is there a way we could include this in our automated tests?

driskull (approved):

geospatialem:
> Is there a way we could include this in our automated tests?

@jcfranco Great question! Still trying to grasp what's possible with a11y tests, but I don't think its feasible in this case since its an attribute on the host, similar to `role` and `aria-label`, unless we could add in a general test ensuring `aria-label`, and `aria-live` are present?

PR #4897, opened 2022-07-11
chore: properly filter non-component files when generating hydration styles

jcfranco:
**Related Issue:** https://github.com/Esri/calcite-components/pull/4856

## Summary

The `hydrationStyles.ts` script was generating styles for non-component files because it was using an async function for sync filtering when processing the component files. This ensures filtering is async.


driskull (approved):
👍

PR #4914, opened 2022-07-13
docs(changelog): replacing breaking change from changelog with deprecation notice

eriklharper:
**Related Issue:** #2138 

## Summary

This resubmits the changelog update to remove label's breaking change and add it in as a deprecation instead.


jcfranco (approved):
💥

PR #4916, opened 2022-07-13
revert: css-only hydration to prevent FOUC

jcfranco:
**Related Issue:** #991

## Summary

This rolls back https://github.com/Esri/calcite-components/pull/4856 as it introduced a regression where components from the `dist-custom-elements` bundle do not get displayed since these don't apply the `calcite-hydrated` attribute when hydrated (see https://github.com/Esri/calcite-components/issues/991#issuecomment-1183480664). There is an existing [Stencil issue](https://github.com/ionic-team/stencil/issues/3437) related to this, so we should revisit once that gets resolved.

This reverts 6405b1f224a22c2415145a92f94c6a46dcbd042f, dae95865110c191b1135ebe0af2abb088d983cc8.


.gitignore
```diff
@@ -8,7 +8,6 @@ node_modules/
 support/output-targets/custom-elements/bundles/
 src/**/components.d.ts
 src/**/*.js
-src/assets/styles/_hydration.scss
```
driskull, line 11: Maybe we can just comment all this stuff out until its ready?

src/assets/styles/global.scss
```diff
@@ -4,9 +4,6 @@
 /* CSS vars (@include in global) */
 @import "type";
 
-/* Hydration styles to hide non-hydrated components */
```
driskull, line 7: Maybe we can just comment this out for now?

support/hydrationStyles.ts
```diff
@@ -1,33 +0,0 @@
-type Dirent = import("fs").Dirent;
```
driskull, line 1: Maybe we can leave this file for now even though it won't be used until its fixed?

driskull (approved):
👍 

What are your thoughts on leaving the hydrationStyles file in the codebase for now?

benelan (approved):
LGTM! I'll release when this is merged. Third time's the charm 🤞

I think verting™️ when the stencil issue is resolved would be easier than remembering which lines in which files to uncomment, as well as remembering to change some of the NPM scripts

Issue #4918, opened 2022-07-13
Combobox selection not announced to screen reader users


geospatialem:
### Summary
Selection is not announced to `combobox` in `selection-mode="single"` on-demand or when in focus. Other `selection-mode`s receive announcement when in focus, but not on-demand.

### Actual Behavior

After selecting a `combobox-item` where `selection-mode="single"` context is is not provided to the screen reader, even when the `combobox` is in focus and a selected value is displayed visually.

All `combobox`s do not receive context when a `combobox-item` value is selected.

#### JAWS transcript
```
Favorite fruit 
Choose a fruit
ExtendedSelect ListBox 
Not Selected Apple
To move to an item press the Arrow keys.
Not Selected Banana 
2 of 12
selected
```

### Expected Behavior
Provide context that an item has been selected in the combobox. A [combobox example](https://w3c.github.io/aria-practices/examples/combobox/combobox-select-only.html) that depicts an active value when the combobox is in focus.

- [x] If there is a selection in the combobox, it should be announced when in focused

#### JAWS transcript
```
Favorite Fruit Combo box 
Apple
To change the selection use the Arrow keys.
```
- [x] After selection of a combobox item, it should be announced.

#### JAWS transcript
```
Favorite Fruit Combo box 
Apple
To change the selection use the Arrow keys.
Favorite Fruit List box 
Apple 
2 of 13
To move to an item press the Arrow keys.
Banana 
3 of 13
Favorite Fruit Combo box 
Banana
To change the selection use the Arrow keys.
```

### Reproduction Sample

https://codepen.io/geospatialem/pen/VwXbRmB

### Reproduction Steps

1. Activate JAWS, NVDA, or VoiceOver
2. Navigate to the [sample](https://codepen.io/geospatialem/pen/VwXbRmB)
3. Expand the first `combobox` (single `selection-mode`)
4. Observe there is no announcement to the selected value.
5. Select a `combobox-item` from the list.
6. Observe there is no announcement provided when selecting a new value
7. Expand the second `combobox` (multi `selection-mode`)
8. Observe there is an announcement to the selected value.
9. Select a `combobox-item` from the list.
10. Observe there is no announcement provided when selecting a new value

### Reproduction Version

beta.83

### Relevant Info

Present across screen readers (JAWS, NVDA, VoiceOver)
cc #4388

#### 1. Potential solution to announce selection on-demand (applies to all `selection-mode` types)
- Add `aria-labelledby` to div for single selection
[...]

Issue #4917, opened 2022-07-13
Combobox item selection state is not read back to screen reader users


geospatialem:
### Summary

Unselected states for `combobox-item`s are read back to screen reader users, but selected ones are not. 

**Some of the expected behavior has been mitigated, but additional development is needed.**

### Actual Behavior

If a user selects a `combobox-item`, context is not currently provided to the screen reader that selection has taken place. When (Unselected items are read back as unselected.)

Also, if the user returns to the input, currently context is not provided to the screen reader that an item has been previously selected.

### Expected Behavior

- [x] When selecting a `combobox-item`, screen reader users should be provided context that the item was selected.
- [ ] When navigating the `combobox-item`s, screen reader users should be provided context that the item has been selected.
- [ ] When selections have been made in the `combobox` the number of items selected should be read back to screen readers (e.g., "1 item selected")
    - [Sample: Accessible Multi-Selection Dropdown](https://codepen.io/geospatialem/pen/oNpVLYx) 

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/combobox/#sample

### Reproduction Steps

1. Activate a screen readers (JAWS/NVDA/VoiceOver)
2. Navigate to the [sample](https://developers.arcgis.com/calcite-design-system/components/combobox/#sample)
3. Keyboard navigate the `combobox` and select an item in the list using the arrow keys for navigation and the `Enter` key for selection
    - Observe the behavior of non-selected items (that are provided context)
    - Observe the behavior upon selecting (provided no context)
    - Navigate away from the selected item, and back to it using the arrows keys to observe the behavior of the selected item (provided no context)

### Reproduction Version

beta.83

### Relevant Info

Present across screen readers (JAWS, NVDA, VoiceOver)

cc #4388

driskull:
The combobox already sets `aria-selected` on the item. Is there something missing?

geospatialem:
The second and third bullet points above are likely a result of assistive technology (AT) products. Since there is still some context provided, we are meeting the minimum requirements. Closing this issue for now, and can re-evaluate if a new issue is created in the future.

> 2. When navigating the combobox-items, screen reader users should be provided context that the item has been selected.
 
> 3. When selections have been made in the combobox the number of items selected should be read back to screen readers (e.g., "1 item selected")

cc @SaraLawrence

PR #4931, opened 2022-07-14
fix(accordion-item): add on-demand screen reader feedback to collapsed and expansion interaction

geospatialem:
**Related Issue:** #4691

## Summary
Adds on-demand feedback to screen reader users when expanding or collapsing the `accordion-item`.

Previously screen reader users received context if the `accordion-item` was previously expanded or collapsed when the item was in focus, but were not provided on-demand feedback when interacting with the `accordion-item`.

Tested with JAWS, NVDA, and VoiceOver.

_Similar modifications might also assist with fixes to #4917 and #4918._


anveshmekala (approved):
🚀

src/components/accordion-item/accordion-item.tsx
```diff
@@ -130,7 +130,12 @@ export class AccordionItem {
     const buttonId = `${guid}-button`;
 
     return (
-      <Host tabindex="0">
+      <Host
+        aria-expanded={toAriaBoolean(this.active)}
```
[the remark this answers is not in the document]
geospatialem: > Is it possible to move these new attributes into the div inside the host so that they cannot be modified by an end user?

Unfortunately using the div wouldn't provide on-demand feedback as the user is expanding/collapsing the component. If we leverage the div, collapse/expand is read back, but only when navigating away and then back to the item.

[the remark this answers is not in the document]
driskull: What if the tabindex is moved from the host to the div inside it?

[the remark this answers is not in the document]
geospatialem: > What if the tabindex is moved from the host to the div inside it?

That does it for JAWS and NVDA. 🎉 Will commit it up to check on VoiceOver.

src/components/accordion-item/accordion-item.tsx
```diff
@@ -127,23 +127,22 @@ export class AccordionItem {
...
           class={{
             [`icon-position--${this.iconPosition}`]: true,
             [`icon-type--${this.iconType}`]: true
           }}
+          onClick={this.itemHeaderClickHandler}
+          role="button"
+          tabindex="0"
         >
           <div
             aria-controls={regionId}
```
driskull, line 144: Do we still need this `aria-controls`?

src/components/accordion-item/accordion-item.tsx
```diff
@@ -170,7 +169,6 @@ export class AccordionItem {
           </div>
           <div
             aria-expanded={toAriaBoolean(this.expanded || this.active)}
```
driskull, line 171: Do we still need this `aria-expanded` here?

driskull: As well as the other aria attributes and roles here?

driskull (approved):
Looking good!

src/components/accordion-item/accordion-item.tsx
geospatialem: Originally it was needed, but these changes might have mitigated the issue. Updating to exclude the `aria-expanded` and `role`.

src/components/accordion-item/accordion-item.tsx
geospatialem: With the earlier updates, we don't. Good catch!

Issue #4595, opened 2022-05-19
calciteInternalStepperItemChange event sometimes emits when programmatically changing items


benelan:
### Actual Behavior 
`calciteInternalStepperItemChange` inconsistently emits when programmatically changing stepper items. 

### Expected Behavior
`calciteInternalStepperItemChange` should not ever emit when stepper items are changed programmatically. #4398

### Reproduction Sample

https://codepen.io/benesri/pen/NWyjdxB?editors=1011

### Reproduction Steps

1. Click "Next Step" a couple times and notice the position is logged in the console
2. Click "Previous Step" a couple times and notice the position is not logged in the console
3. Repeat step 1 and nothing is logged in the console

### Reproduction Version

next.473

### Regression?

benelan:
Nevermind, it got renamed to `calciteInternalStepperItemChange`

benelan:
Reopening/repurposing this issue because `calciteInternalStepperItemChange` sometimes emits when programmatically changing stepper items, which shouldn't happen after #4398. I updated the issue body

driskull:
I think this would require a refactor because the `stepper-item` is listening for the event on the body in order to do some updates.

benelan:
The action items for this issue are
-  make sure `calciteInternalStepperItemChange` only emits on user interaction
- rename the event to `calciteInternalStepperItemSelect` and make it public

geospatialem:
Verified on `next.500`

https://user-images.githubusercontent.com/5023024/175562221-3f621f83-9575-4bde-9a5c-07462c8bc895.mp4

`

kevindoshier:
This still doesnt seem correct,  in the sample programatically a nextStep, prevStep are being used, but there is the case where we dont use that and the user will just click one of the stepper positions, for that case we do not get back a position, here is a codePen

https://codepen.io/kevindoshier/pen/wvyNLJV

driskull:
@kevindoshier can you create a new issue for this? I have a fix in progress.

geospatialem:
> This still doesnt seem correct, in the sample programatically a nextStep, prevStep are being used, but there is the case where we dont use that and the user will just click one of the stepper positions, for that case we do not get back a position, here is a codePen

@kevindoshier, For some reason the Codepen sample wasn't providing a value for the second or the last stepper item. 

[...]

PR #4938, opened 2022-07-14
fix(stepper-item): Correctly emit position in calciteStepperItemChange event

driskull:
**Related Issue:** #4595

## Summary

fix(stepper-item): Correctly emit position in calciteStepperItemChange event

Calculates and stores position of the element on `connectedCallback` instead of `onslotchange`. This is needed because some instances of stepper may not set any content and therefore the position would never be calculated and set. If the element moves, its position should be updated with `connectedCallback`.


kevindoshier:
Thanks @driskull! 👍

driskull:
Ready for review

jcfranco (approved):
LGTM!

PR #4939, opened 2022-07-14
feat(modal,block,loader): deprecate disablePadding , noPadding and add support for custom CSS variables

anveshmekala:
**Related Issue:** #3860

This PR will deprecate `noPadding`, `disablePadding` props across the repo and replace them with CSS props instead. List of CSS props added to replace props

- modal : `--calcite-modal-padding`
- loader : `--calcite-loader-padding`
- block : `--calcite-modal-padding`

Moreover, public CSS props ` --calcite-modal-padding-large`,`--calcite-modal-title-text` &`--calcite-modal-content-text` of `modal` are deprecated.


src/components/block/block.tsx
```diff
@@ -107,7 +107,11 @@ export class Block implements ConditionalSlotComponent, InteractiveComponent {
   /**Block description */
   @Prop() description: string;
 
-  /** When true, removes padding for the slotted content */
+  /**
+   * When true, removes padding for the slotted content
+   *
+   * @deprecated Use `--calcite-block-padding`  CSS variable instead.
```
driskull, line 113: Looks like an extra space before `  CSS`.

@geospatialem do we refer to CSS variables as `variables` or `custom properties`? 

Custom properties (--*): CSS variables

src/components/loader/loader.scss
```diff
@@ -13,14 +13,15 @@ $loader-scale: 54;
...
-  @apply relative mx-auto hidden items-center justify-center py-16 opacity-100;
+  @apply relative mx-auto hidden items-center justify-center opacity-100;
   min-block-size: var(--calcite-loader-size);
   font-size: var(--calcite-loader-font-size);
   stroke: var(--calcite-ui-brand);
   stroke-width: $stroke-width;
   fill: none;
   transform: scale(1, 1);
   animation: loader-color-shift 6s alternate-reverse infinite linear;
+  padding-block: var(--calcite-loader-padding, 4rem);
```
driskull, line 24: why the 4rem here?

src/components/modal/modal.tsx
```diff
@@ -115,7 +115,11 @@ export class Modal implements ConditionalSlotComponent, OpenCloseComponent {
   /** Sets the background color of the component's content. */
   @Prop({ reflect: true }) backgroundColor: ModalBackgroundColor = "white";
 
-  /** When true, disables spacing to the content area slot. */
+  /**
+   * When true, disables spacing to the content area slot.
+   *
+   * @deprecated  Use `--calcite-modal-padding`  CSS variable instead.
```
driskull, line 121: spacing issue here too.

driskull (commented):
comments :)

src/components/block/block.tsx
geospatialem: We'd prefer `variables` for consistency, also perhaps better to keep separated from our custom element properties.

src/components/loader/loader.scss
anveshmekala: replacing py-16, spacing.16 would be 4rem in tailwind.

driskull: Can you reference the spacing instead of 4rem? `theme("spacing.16")`?

driskull: The problem with 4rem is that if we change the spacing down the road it won't get updated here.

anveshmekala: agreed. better to refer tailwind utils.

src/components/block/block.scss
```diff
@@ -146,7 +135,8 @@ calcite-handle {
 }
 
 .content--spaced {
-  @apply px-3 py-2;
+  padding-block: var(--calcite-block-padding, 0.5rem);
```
driskull, line 138: These values need to be referenced from tailwind as well.

src/components/block/block.stories.ts
```diff
@@ -175,7 +175,7 @@ export const disabled = (): string => html`<calcite-block heading="heading" summ
   </calcite-block-section>
 </calcite-block>`;
 
-export const noPadding = (): string => html` <calcite-panel heading="Properties">
+export const disablePadding = (): string => html` <calcite-panel heading="Properties">
```
driskull, line 178: nitpick: paddingDisabled

driskull (changes requested):
Looks good just needs those two vars updated to use tailwind values

driskull (approved):
👍 🎉 🌮 w00t

PR #4946, opened 2022-07-15
feat(input-text): create separate component for input type text

benelan:
**Related Issue:** part of the roadmap item to separate `calcite-input` types

## Summary
Created a new `calcite-input-text` component by copying over input and removing all non-text functionality.


src/components/input-text/input-text.tsx
```diff
@@ -0,0 +1,554 @@
...
+  }: {
+    committing?: boolean;
+    nativeEvent?: MouseEvent | KeyboardEvent | InputEvent;
+    origin: SetValueOrigin;
+    previousValue?: string;
+    value: string;
+  }): void => {
+    this.setPreviousValue(previousValue || this.value);
+    this.previousValueOrigin = origin;
+    this.userChangedValue = origin === "user";
```
eriklharper, line 446: ☝️ any reason why the `&& this.value !== newValue` didn't carry over on this line?  I recall adding that because there were sometimes cases when the new value coming into setValue was the same as the previous value so the user couldn't possibly have changed it in that case since it is the same as the previous value.

src/components/input-text/input-text.tsx
```diff
@@ -0,0 +1,554 @@
...
+        disabled={this.disabled ? true : null}
+        enterKeyHint={this.el.enterKeyHint}
+        inputMode={this.el.inputMode}
+        maxLength={this.maxLength}
+        minLength={this.minLength}
+        name={this.name}
+        onBlur={this.inputBlurHandler}
+        onFocus={this.inputFocusHandler}
+        onInput={this.inputInputHandler}
+        onKeyDown={this.inputKeyDownHandler}
```
eriklharper, line 524: I wonder since this are all internal, if they should be named `internalInput*Handler`?

src/components/input-text/input-text.tsx
```diff
@@ -0,0 +1,554 @@
...
+        ref={this.setChildElRef}
+        required={this.required ? true : null}
+        tabIndex={this.disabled || (this.inlineEditableEl && !this.editingEnabled) ? -1 : null}
+        type="text"
+        value={this.value}
+      />
+    );
+
+    return (
+      <Host onClick={this.inputFocusHandler} onKeyDown={this.keyDownHandler}>
```
eriklharper, line 536: ☝️ I'm also thinking these two handlers should be named `host${eventName}Handler`.

src/components/input-text/readme.md
```diff
@@ -0,0 +1,162 @@
+# calcite-input-text
+
+<!-- Auto Generated Below -->
+
+## Usage
+
+### Basic
+
+```html
+<calcite-input value="Entered value" placeholder="My placeholder"></calcite-input>
```
eriklharper, line 10: ☝️ all instances of `calcite-input` in this readme need renamed to `calcite-input-text`.

src/components/input-text/readme.md
```diff
@@ -0,0 +1,162 @@
...
+| `value`            | `value`              | The component's value.                                                                                                                                                                         | `string`                                                                                                                                                                               | `""`                                      |
+
+## Events
+
+| Event                | Description                                         | Type                |
+| -------------------- | --------------------------------------------------- | ------------------- |
+| `calciteInputChange` | Fires each time a new value is typed and committed. | `CustomEvent<void>` |
+| `calciteInputInput`  | Fires each time a new value is typed.               | `CustomEvent<any>`  |
```
eriklharper, line 105: These event names are out of date too.

src/components/input-text/readme.md
```diff
@@ -0,0 +1,162 @@
...
+| `suffixText`       | `suffix-text`        | Adds text to the end of the component.                                                                                                                                                         | `string`                                                                                                                                                                               | `undefined`                               |
+| `type`             | `type`               | Specifies the component type. Note that the following types add type-specific icons by default: "date", "email", "password", "search", "tel", "time".                                          | `"color" \| "date" \| "datetime-local" \| "email" \| "file" \| "image" \| "month" \| "number" \| "password" \| "search" \| "tel" \| "text" \| "textarea" \| "time" \| "url" \| "week"` | `"text"`                                  |
```
eriklharper, line 97: Type prop no longer present.

src/components/input-text/readme.md
```diff
@@ -0,0 +1,162 @@
...
+| `numberButtonType` | `number-button-type` | Specifies the placement of the buttons for type "number".                                                                                                                                      | `"horizontal" \| "none" \| "vertical"`                                                                                                                                                 | `"vertical"`                              |
+| `numberingSystem`  | `numbering-system`   | Specifies the Unicode numeral system used by the component for localization.                                                                                                                   | `string`                                                                                                                                                                               | `undefined`                               |
```
eriklharper, line 88: Number stuff is gone too

src/components/input-text/input-text.scss
```diff
@@ -0,0 +1,327 @@
+:host {
```
eriklharper, line 1: There appear to be several spots where logical props can be used in place of things that specify height, width, padding and margins.  Should we update those spots?

src/components/input-text/readme.md
benelan: those are autogenerated, a GH action will fix the readme

src/components/input-text/readme.md
benelan: autogenerated as well

src/components/input-text/input-text.tsx
benelan: I can't find any other examples of using the `internal` prefix for handlers in our repo, are you suggesting a convention change? I think its unnecessarily verbose

src/components/input-text/input-text.scss
benelan: That should probably go in a separate `refactor` issue/PR

src/components/input-text/input-text.tsx
benelan: Because we don't need `newValue` anymore. For type="text" it looks like `value === newValue`

src/components/input-text/input-text.tsx
eriklharper: Not a huge deal, but perhaps it could be a good convention change.  It would differentiate between internal handlers when doing project-wide searches.  Not a blocker for this pr though.

benelan: Are there any external handlers?

src/components/input-text/input-text.tsx
eriklharper: Right, I forgot that `newValue` was calculated as part of the number input.  I think it would be good to check that the value coming into `setValue` is actually different than the currently set `this.value`, particularly for this state property, so it could read `this.userChangedValue = origin === "user" && value !== this.value`.

src/components/input-text/input-text.tsx
eriklharper: Not necessarily in tsx, but there would be in usage snippets like in tests and docs/html.

src/components/input-text/input-text.tsx
```diff
@@ -0,0 +1,554 @@
...
+
+  /**
+   * @internal
+   */
+  @Event() calciteInternalInputTextBlur: EventEmitter;
+
+  /**
+   * Fires each time a new value is typed.
+   */
+  @Event({ cancelable: true }) calciteInputTextInput: EventEmitter;
```
jcfranco, line 280: Can you type the emitters as `<void>` if there's no payload?

src/components/input-text/input-text.tsx
```diff
@@ -0,0 +1,554 @@
...
+
+/**
+ * @slot action - A slot for positioning a button next to the component.
+ */
+@Component({
+  tag: "calcite-input-text",
+  styleUrl: "input-text.scss",
+  shadow: true
+})
+export class InputText implements LabelableComponent, FormComponent, InteractiveComponent {
```
jcfranco, line 41: One more 'next step' comment, we should start thinking about extracting common behavior, rendering, etc so all input types can leverage.

jcfranco (approved):
LGTM!

src/components/input-text/readme.md
benelan: @jcfranco storybook won't build when I delete the readme. I'll just delete the api ref table hopefully that does it 
image

Issue #4944, opened 2022-07-15
calcite dropdown cannot be navigated with keyboard when containing a disabled item


janett-baresel:
### Summary

When creating a dropdown with a disabled dropdown item, the user cannot reach any elements passed the disabled item.

### Actual Behavior

When tabbing through the dropdown items the dropdown closes when the focus would reach the disabled item.

Note: if I use a dropdown group with the same items, this works as expected.

### Expected Behavior

When tabbing through the dropdown items the dropdown stays open and the focus skips the disabled dropdown item and goes to the next item. (This works as expected for item sin a dropdown group)

### Reproduction Sample

https://codepen.io/Programmine/pen/jOzVvRN?editors=100

### Reproduction Steps

1. create `calcite-dropdown` without `calcite-dropdown-group` with 3 items
2. make the middle item disabled
3. use your keyboard to navigate passed the disabled item
-> dropdown closes and focus moves on somewhere else

### Reproduction Version

latest (coming from codepan) so I assume beta86?

<SUP>SVCH</SUP>

driskull:
@benelan @jcfranco I think dropdown should be navigated via arrow keys instead of using tab. Tab should probably skip out of the dropdown to the next focusable element in the DOM.

Example: https://www.w3.org/WAI/ARIA/apg/patterns/menubutton/
https://www.w3.org/WAI/ARIA/apg/example-index/menu-button/menu-button-actions-active-descendant.html

@geospatialem Is that correct?

Maybe we can use this issue to update it to use arrow keys instead of tab?

benelan:
> I think dropdown should be navigated via arrow keys instead of using tab. Tab should probably skip out of the dropdown to the next focusable element in the DOM.

That sound intuitive to me, but I'll defer to kitty/franco

geospatialem:
> Tab should probably skip out of the dropdown to the next focusable element in the DOM.
> @geospatialem Is that correct?

@driskull Yes, that is the expected behavior, using arrow keys to access the items. 👍🏻 

This only seems to apply when the `dropdown-group` isn't present, as Janett mentioned, so the initial behavior when using `dropdown-group` is performing as intended. Should we account for the missing group? If there is no specified `groupTitle` prop on the group, only the items are shown, like this [example](https://codepen.io/geospatialem/pen/eYMgYbb).

[...]

PR #4949, opened 2022-07-15
fix(dropdown): Fix navigating without a dropdown-group

driskull:
**Related Issue:** #4944

## Summary

fix(dropdown): Fix navigating without a dropdown-group. (#4944)

Also updates aria roles for a11y.

### DIFF
--- a/src/components/dropdown-group/dropdown-group.tsx
+++ b/src/components/dropdown-group/dropdown-group.tsx
@@ -83,7 +83,7 @@ export class DropdownGroup {
       this.groupPosition > 0 ? <div class="dropdown-separator" role="separator" /> : null;
 
     return (
-      <Host role="menu">
+      <Host role="group">
         <div
           class={{
             container: true,
--- a/src/components/dropdown-group/dropdown-group.tsx
+++ b/src/components/dropdown-group/dropdown-group.tsx
@@ -83,7 +83,7 @@ export class DropdownGroup {
       this.groupPosition > 0 ? <div class="dropdown-separator" role="separator" /> : null;
 
     return (
-      <Host role="group">
+      <Host aria-label={this.groupTitle} role="group">
         <div
           class={{
             container: true,
--- a/src/components/dropdown/dropdown.tsx
+++ b/src/components/dropdown/dropdown.tsx
@@ -158,6 +158,7 @@ export class Dropdown implements InteractiveComponent, OpenCloseComponent {
     this.mutationObserver?.observe(this.el, { childList: true, subtree: true });
     this.createPopper();
     this.setFilteredPlacements();
+    this.updateItems();
   }
 
   componentDidLoad(): void {
@@ -212,7 +213,7 @@ export class Dropdown implements InteractiveComponent, OpenCloseComponent {
             role="menu"
           >
             <div hidden={!(open || active)}>
-              <slot onSlotchange={this.updateItems} />
+              <slot />
             </div>
           </div>
         </div>
@@ -407,7 +408,7 @@ export class Dropdown implements InteractiveComponent, OpenCloseComponent {
   };
 
   updateItems = (): void => {
-    this.items = Array.from(this.el?.querySelectorAll("calcite-dropdown-item"));
+    this.items = Array.from(this.el.querySelectorAll("calcite-dropdown-item"));
 
     this.updateSelectedItems();
 
--- a/src/components/dropdown/dropdown.tsx
+++ b/src/components/dropdown/dropdown.tsx
@@ -419,19 +419,19 @@ export class Dropdown implements InteractiveComponent, OpenCloseComponent {
   updateItems = (): void => {
     const { defaultAssignedElements } = this;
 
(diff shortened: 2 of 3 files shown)


src/components/dropdown-group/dropdown-group.tsx
```diff
@@ -83,7 +83,7 @@ export class DropdownGroup {
       this.groupPosition > 0 ? <div class="dropdown-separator" role="separator" /> : null;
 
     return (
-      <Host role="menu">
+      <Host role="group">
```
geospatialem, line 86: Could we add an `aria-label={groupTitle}` here as well? This would announce the title to screen reader users.

src/components/dropdown/dropdown.tsx
```diff
@@ -175,18 +176,20 @@ export class Dropdown implements InteractiveComponent, OpenCloseComponent {
...
+          id={`${guid}-menubutton`}
           onClick={this.openCalciteDropdown}
           onKeyDown={this.keyDownHandler}
           ref={this.setReferenceEl}
         >
           <slot
+            aria-controls={`${guid}-menu`}
             aria-expanded={toAriaBoolean(active || open)}
-            aria-haspopup="true"
+            aria-haspopup="menu"
```
geospatialem, line 192: After testing, it looks like we should use the `listbox` role here to accommodate JAWS in all browsers and NVDA in the Chrome browser (also replaced below).

```html
aria-haspopup="listbox"
```

src/components/dropdown/dropdown.tsx
```diff
@@ -197,16 +200,19 @@ export class Dropdown implements InteractiveComponent, OpenCloseComponent {
...
+            aria-labelledby={`${guid}-menubutton`}
             class={{
               ["calcite-dropdown-content"]: true,
               [PopperCSS.animation]: true,
               [PopperCSS.animationActive]: active || open
             }}
+            id={`${guid}-menu`}
             onTransitionEnd={this.transitionEnd}
             ref={this.setScrollerEl}
+            role="menu"
```
geospatialem, line 212: Similar to above, to accommodate JAWS (all browsers) and NVDA in Chrome:

```html
role="listbox"
```

geospatialem (commented):
Looking good! 👓 

Few considerations encountered while testing. Specifically with JAWS in all browsers, and NVDA in Chrome.

src/components/dropdown/dropdown.tsx
driskull: Listbox isn't the right role for this element. Listbox is a selectable form element whereas this is a menu.

src/components/dropdown/dropdown.tsx
```diff
@@ -401,25 +417,27 @@ export class Dropdown implements InteractiveComponent, OpenCloseComponent {
   };
 
   updateItems = (): void => {
-    this.items = this.groups
-      .map((group) => Array.from(group?.querySelectorAll("calcite-dropdown-item")))
+    const { defaultAssignedElements } = this;
```
driskull, line 420: For the dropdown, we need to continue to use slotted elements because the dropdown is used within the split-button component where slotted elements are passed from the split button to the dropdown.

Previously, we were just expecting `calcite-dropdown-groups` in slotted elements. This change allows both groups and items to be slotted elements.

src/components/dropdown/dropdown.tsx
macandcheese: I think menu only applies when the dropdown group is selection mode: none, as there’s no concept of chosen / selected item, and the dropdown items are actions or links. 

For single / multiple select, it does seem like a different role than menu applies.

driskull: I'd prefer not to change the role based on the selection type. That could get complicated.

If we change it to use listbox then it is really just another `select`/`combobox`

The menu role does have selectable options. Such as...

```
const itemRole = this.href
      ? null
      : this.selectionMode === "single"
      ? "menuitemradio"
      : this.selectionMode === "multi"
      ? "menuitemcheckbox"
      : "menuitem";
```

driskull:
@jcfranco this one is ready for review

src/components/dropdown/dropdown.tsx
```diff
@@ -401,25 +417,27 @@ export class Dropdown implements InteractiveComponent, OpenCloseComponent {
...
+
+    this.items = (
+      defaultAssignedElements.filter((el) =>
+        el?.matches("calcite-dropdown-group, calcite-dropdown-item")
+      ) as (HTMLCalciteDropdownItemElement | HTMLCalciteDropdownGroupElement)[]
+    )
+      .map((el) =>
+        el.matches("calcite-dropdown-group")
+          ? Array.from(
+              (el as HTMLCalciteDropdownGroupElement).querySelectorAll("calcite-dropdown-item")
```
jcfranco, line 430: If dropdown groups can only have items as children, you could skip the query.

jcfranco (approved):

Issue #4902, opened 2022-07-12
calciteModalOpen emitted when changing calcite-input-message active


tom7248:
### Actual Behavior

If a calcite-modal contains a calcite-input-message, changing the active property of the message causes the modal to emit a calciteModalOpen.

### Expected Behavior

calciteModalOpen should only be emitted once

According to the docs: https://developers.arcgis.com/calcite-design-system/components/modal/#component-api-events-calciteModalOpen

> Emits when `calcite-modal` finishes the open animation.

### Reproduction Sample
https://codepen.io/_t_o_m_/pen/mdxrOON

### Reproduction Steps

1. Tap the "Show Modal" button
2. Observe in the console the message "calciteModalOpen" (expected)
3. On the modal, tap the "Show Message" button, the message appears
4. Observe in the console the message "calciteModalOpen" a second time (not expected)

### Reproduction Version

beta.83

driskull:
@jcfranco I think this would fix it. Does this sound reasonable?

```patch
diff --git a/src/components/modal/modal.tsx b/src/components/modal/modal.tsx
index 42a1e408b99d07028470161e432f935734cc93f6..19683057cd7ce46f44dcf8af16ad4b2f576f6641 100644
--- a/src/components/modal/modal.tsx
+++ b/src/components/modal/modal.tsx
@@ -151,7 +151,11 @@ export class Modal implements ConditionalSlotComponent {
       >
         <calcite-scrim class={CSS.scrim} onClick={this.handleOutsideClose} />
         {this.renderStyle()}
-        <div class="modal" onTransitionEnd={this.transitionEnd}>
+        <div
+          class="modal"
+          onTransitionEnd={this.transitionEnd}
+          ref={(el) => (this.containerEl = el)}
+        >
           <div data-focus-fence onFocus={this.focusLastElement} tabindex="0" />
           <div class={CSS.header}>
             {this.renderCloseButton()}
@@ -262,6 +266,8 @@ export class Modal implements ConditionalSlotComponent {
 
   private activeTransitionProp = "opacity";
 
+  private containerEl: HTMLDivElement;
+
   //--------------------------------------------------------------------------
   //
   //  Event Listeners
@@ -346,7 +352,7 @@ export class Modal implements ConditionalSlotComponent {
   //
   //--------------------------------------------------------------------------
   transitionEnd = (event: TransitionEvent): void => {
-    if (event.propertyName === this.activeTransitionProp) {
[...]

PR #4962, opened 2022-07-18
fix(modal, alert, combobox, dropdown, input-date-picker, popover): Scope open and close events to component

driskull:
**Related Issue:** #4902

## Summary

fix(modal, alert, combobox, dropdown, input-date-picker, popover): Scope open and close events to component. (#4902)

Makes sure that transition events are only fired when they occur on the element they are supposed to.

### DIFF
--- a/src/components/alert/alert.tsx
+++ b/src/components/alert/alert.tsx
@@ -350,13 +350,13 @@ export class Alert implements OpenCloseComponent {
   }
 
   transitionStartHandler = (event: TransitionEvent): void => {
-    if (event.propertyName === this.activeTransitionProp) {
+    if (event.propertyName === this.activeTransitionProp && event.target === this.containerEl) {
       this.active ? this.onBeforeOpen() : this.onBeforeClose();
     }
   };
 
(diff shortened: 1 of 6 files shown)


jcfranco (approved):
Nice! 

We should look into adding a common test util for all OpenCloseComponents that covers this in addition to the expected behavior. Can you create an issue for this? 

cc @Elijbet

driskull:
https://github.com/Esri/calcite-components/issues/4968

driskull:
I think we could add `activeTransitionProp` to the `openCloseComponent` interface as well as a new prop like `transitionEl` which would be the element that the transition is allowed to emit on.

Elijbet:
> I think we could add `activeTransitionProp` to the `openCloseComponent` interface as well as a new prop like `transitionEl` which would be the element that the transition is allowed to emit on.

Are you taking this on, or should I? I'll be working on adding the interface to 4 additional components today.

driskull:
Feel free to grab it @Elijbet :) Thanks!!

Issue #4955, opened 2022-07-18
Regression with stepper item `disabled` styling


nwhittaker:
### Actual Behavior

The title of the disabled stepper item is no longer at half opacity:

<img alt="Screen Shot 2022-07-18 at 3 30 56 PM" src="https://user-images.githubusercontent.com/421496/179602611-f301f4fc-98b1-4d6c-84b3-2db2a7862223.png">

### Expected Behavior

The title of the disabled stepper item is at half opacity:

<img alt="Screen Shot 2022-07-18 at 3 31 43 PM" src="https://user-images.githubusercontent.com/421496/179602742-14d14daf-e6b2-4e25-acd6-3458388152e6.png">

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/KKoWMNW

### Reproduction Steps

1. Visit the sample
2. Compare the last stepper item's appearance with the doc

### Reproduction Version

next.533

### Relevant Info

Likely introduced by the switch to `display: contents` in https://github.com/Esri/calcite-components/pull/4367:
https://github.com/Esri/calcite-components/blob/aa74d90ae5061e2691abfe8915dcfeb6c93f27e2/src/components/stepper-item/stepper-item.scss#L246-L251

### Regression?

beta.86

jcfranco:
@nwhittaker Thanks for catching this, #4954 and #4956 and for the excellent bug reports! PRs are up and should land this week.

benelan:
verified on next

PR #4964, opened 2022-07-19
fix(stepper-item): restore disabled styling on horizontal layout stepper items

jcfranco:
**Related Issue:** #4955 

## Summary

The disabled style mixin was no longer applied to the host as the stepper item uses `display:contents`, which basically replaces the host container with its children. This extracts the host styles from that mixin to be applied individually in this case.


driskull (approved):
👍

Issue #4954, opened 2022-07-18
Regression with stepper `cursor` style


nwhittaker:
### Actual Behavior

Hovering stepper content shows the `pointer` style cursor.

### Expected Behavior

Hovering stepper content shows the `default` style cursor.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/abYJNpp

### Reproduction Steps

1. Visit the sample
2. Mouse over the active stepper item's content

### Reproduction Version

next.533

### Relevant Info

Likely introduced by https://github.com/Esri/calcite-components/pull/4367 which keeps the stepper item's content within the `.container` element which is supplying the `cursor-pointer` style.

### Regression?

beta.86

benelan:
verified on next

PR #4965, opened 2022-07-19
fix(stepper-item): avoid cursor-pointer on item content for horizontal layouts

jcfranco:
**Related Issue:** #4954 

## Summary

This fixes a mouse-cursor regression introduced by #4367.

Note: this has no accompanying screenshot test as Screener does not capture the mouse cursor


driskull (approved):
👍

PR #4972, opened 2022-07-19
feat(input-date-picker): deprecate active property in favor of open

driskull:
**Related Issue:** #

## Summary

feat(input-date-picker): deprecate active property in favor of open

### DIFF
--- a/src/components/input-date-picker/input-date-picker.tsx
+++ b/src/components/input-date-picker/input-date-picker.tsx
@@ -87,7 +87,7 @@ export class InputDatePicker
   @Watch("readOnly")
   handleDisabledAndReadOnlyChange(value: boolean): void {
     if (!value) {
-      this.active = false;
+      this.open = false;
     }
   }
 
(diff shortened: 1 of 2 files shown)


Elijbet (approved):

src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -169,17 +169,31 @@ export class InputDatePicker
...
+  activeHandler(value: boolean): void {
+    this.open = value;
+  }
+
+  /** Expand or collapse when calendar does not have input */
+  @Prop({ mutable: true, reflect: true }) open = false;
+
+  @Watch("open")
+  openHandler(value: boolean): void {
+    this.active = value;
```
anveshmekala, line 189: Wouldn't this cause an infinite loop? one prop changing the value of others and vice versa.

src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -347,6 +361,9 @@ export class InputDatePicker
   // --------------------------------------------------------------------------
 
   connectedCallback(): void {
+    this.activeHandler(this.active || this.open);
+    this.openHandler(this.active || this.open);
```
anveshmekala, line 365: it's better to invoke these methods only if the user sets `active || open` to `true `

anveshmekala (commented):
LGTM ! Added a few comments.

jcfranco (approved):
noice

PR #4973, opened 2022-07-19
feat(notice): deprecate active prop and add support for open prop

anveshmekala:
**Related Issue:** #4688 

## Summary

This will deprecate `active` prop in favor of `open`.


driskull (approved):
👍

Issue #4967, opened 2022-07-19
Accordion: Initially expanded accordion items don't toggle


paulcpederson:
### Actual Behavior

If you set an accordion item to be expanded on load, the user will never be able to close it. 

### Expected Behavior

Even if the accordion item is expanded on load, the user should be able to close/open later.

### Reproduction Sample

https://codepen.io/paulcp/pen/NWYpEwr?editors=1000

### Reproduction Steps

1. Click the last accordion item
2. Notice it doesn't open

### Reproduction Version

1.0.0-beta.86

### Relevant Info

Seems like this is related to the deprecation of the `active` property. In beta.83, if you use `active`, this works properly. But in 84, it stops working both with expanded and active.

### Regression?

1.0.0-beta.83

anveshmekala:
Agreed. Ideally, they should be in sync with each other given they have a default value and it's hard to determine whether the value of the prop is set by the user or the default which is required to determine which one is set by the user.

driskull:
I think part of the problem is that there are currently 2 sources of truth for the value. I think one of them (preferably the newer prop) needs to be the source of truth and the other one should just act like an alias for setting it.

benelan:
verified on next

PR #4974, opened 2022-07-20
fix(accordion-item): allows  initially expanded items to toggle

anveshmekala:
**Related Issue:** #4967 

## Summary

This will allow users to toggle items that are initially `expanded`.


driskull (approved):
💥 👍 Nice 👍 💥

Issue #4968, opened 2022-07-19
Add a common test util for all OpenCloseComponents


driskull:
### Description

Add a common test util for all OpenCloseComponents.

### Acceptance Criteria

A common utility to handle https://github.com/Esri/calcite-components/pull/4962#pullrequestreview-1043972984

### Relevant Info

https://github.com/Esri/calcite-components/pull/4962#pullrequestreview-1043972984

### Which Component

All openCloseComponents

geospatialem:
Verified on the `main` branch. ✨

PR #4975, opened 2022-07-20
refactor(openCloseComponent, alert, combobox, dropdown, input-date-picker, modal, popover): reworked openCloseComponent Interface to apply connect pattern for consistency with other components

Elijbet:
**Related Issue:** #4689

## Summary

* reworked `openCloseComponent` interface to apply connect pattern for consistency with other components

* added `openTransitionProp` to the interface to specify the property on which active transition is watched for
* added `transitionEl` to the interface to specify the element that the transition is allowed to emit on
* added the new `open` prop to the interface that replaces deprecated `active` prop

* applied new interface changes to all affected components:

      `calcite-alert`
      `calcite-combobox`
      `calcite-dropdown`
      `calcite-input-date-picker`
      `calcite-modal`
      `calcite-popover`


src/utils/openCloseComponent.ts
```diff
@@ -8,20 +8,57 @@ export interface OpenCloseComponent {
...
   onClose: () => void;
 }
+
+export function transitionStartHandler(this: OpenCloseComponent, event: TransitionEvent): void {
+  if (event.propertyName === this.activeTransitionProp && event.target === this.transitionEl) {
+    this.active || this.open ? this.onBeforeOpen() : this.onBeforeClose();
+  }
+}
+
+export function transitionEnd(this: OpenCloseComponent, event: TransitionEvent): void {
```
driskull, line 60: Can you change this to be...

``` 
export const transitionEnd = (this: OpenCloseComponent, event: TransitionEvent): void => {
```

This way the `.bind()` will not be necessary in the files.

src/utils/openCloseComponent.ts
```diff
@@ -8,20 +8,57 @@ export interface OpenCloseComponent {
    */
   readonly el: HTMLElement;
 
+  /**
+   * When true, the component is active.
+   *
+   * @deprecated use open instead
+   */
+  active?: boolean;
```
driskull, line 16: Can you remove the active property here?

driskull: We don't need to deprecate private interface things

src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -593,6 +593,11 @@ export class InputDatePicker
       : null;
   };
 
+  private setTransitionEl = (el): void => {
+    this.transitionEl = el;
+    this.transitionEl.addEventListener("transitionstart", transitionStartHandler.bind(this));
```
driskull, line 598: remove `.bind()`

src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -169,9 +173,10 @@ export class InputDatePicker
     }
   }
 
-  /** Expand or collapse when calendar does not have input */
```
driskull, line 172: Need to keep doc here.

src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -169,9 +173,10 @@ export class InputDatePicker
...
   @Prop({ mutable: true, reflect: true }) active = false;
 
+  @Prop({ mutable: true, reflect: true }) open = false;
```
driskull, line 178: I have a PR to handle this here: https://github.com/Esri/calcite-components/pull/4972

For now, just add the `open` prop as a private variable.

src/components/alert/alert.tsx
```diff
@@ -49,8 +53,9 @@ export class Alert implements OpenCloseComponent {
   //
   //---------------------------------------------------------------------------
 
-  /** When true, the component is active. */
```
driskull, line 52: Why was the doc removed here?

src/components/alert/alert.tsx
```diff
@@ -169,8 +174,8 @@ export class Alert implements OpenCloseComponent {
             queued,
             [placement]: true
           }}
-          onTransitionEnd={this.transitionEnd}
-          ref={this.setContainerEl}
+          onTransitionEnd={transitionEnd.bind(this)}
```
driskull, line 177: remove bind usage here.

src/components/alert/alert.tsx
```diff
@@ -299,14 +297,21 @@ export class Alert implements OpenCloseComponent {
...
 
   //--------------------------------------------------------------------------
   //
   //  Private Methods
   //
   //--------------------------------------------------------------------------
 
+  private setTransitionEl = (el): void => {
+    this.transitionEl = el;
+    this.transitionEl.addEventListener("transitionstart", transitionStartHandler.bind(this));
```
driskull, line 312: remove bind usage here.

src/components/combobox/combobox.tsx
```diff
@@ -82,12 +86,9 @@ export class Combobox
   //
   //--------------------------------------------------------------------------
 
-  /**
```
driskull, line 85: Why was the doc for this removed?

src/components/combobox/combobox.tsx
```diff
@@ -101,9 +102,6 @@ export class Combobox
     this.setMaxScrollerHeight();
   }
 
-  /**When true, opens the combobox */
```
driskull, line 104: Why was the doc for this removed?

src/components/combobox/combobox.tsx
```diff
@@ -325,7 +323,7 @@ export class Combobox
     this.destroyPopper();
     disconnectLabel(this);
     disconnectForm(this);
-    this.listContainerEl?.removeEventListener("transitionstart", this.transitionStartHandler);
+    this.listContainerEl?.removeEventListener("transitionstart", transitionStartHandler.bind(this));
```
driskull, line 326: remove bind usage here.

src/components/dropdown/dropdown.tsx
```diff
@@ -56,15 +60,9 @@ export class Dropdown implements InteractiveComponent, OpenCloseComponent {
...
 
-  /**
-   * Opens or closes the dropdown
-   *
-   * @deprecated use open instead.
-   */
-  @Prop({ reflect: true, mutable: true }) active = false;
+  @Prop({ mutable: true, reflect: true }) active = false;
 
-  /** When true, opens the dropdown */
```
driskull, line 66: Keep doc

src/components/dropdown/dropdown.tsx
```diff
@@ -370,14 +368,16 @@ export class Dropdown implements InteractiveComponent, OpenCloseComponent {
...
 
-  private activeTransitionProp = "visibility";
-
   private scrollerEl: HTMLDivElement;
 
   mutationObserver = createObserver("mutation", () => this.updateItems());
 
   resizeObserver = createObserver("resize", (entries) => this.resizeObserverCallback(entries));
 
+  activeTransitionProp = "opacity";
```
driskull, line 377: how come this changed from `visibility` to `opacity`?

driskull (changes requested):
Interface looks good. I think the changes for `open` need to happen in separate PRs so that they are deprecated correctly.

src/components/modal/modal.tsx
```diff
@@ -68,14 +68,8 @@ export class Modal implements ConditionalSlotComponent, OpenCloseComponent {
   //
   //--------------------------------------------------------------------------
 
-  /**
```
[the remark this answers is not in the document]
Elijbet: Because I moved the prop to the interface, the doc got moved along with it. I just forgot that JSdoc needs to be in the TSX for the generated MD files. Sorry. Moving it back :)

src/components/alert/alert.tsx
```diff
@@ -299,14 +296,25 @@ export class Alert implements OpenCloseComponent {
   /* @internal */
   @State() requestedIcon?: string;
 
-  private activeTransitionProp = "opacity";
+  activeTransitionProp = "opacity";
+
+  transitionEl: HTMLDivElement;
+
+  private transitionStartHandler = transitionStartHandler.bind(this);
```
driskull, line 303: @Elijbet can we pass the component instead? That way its consistent with the other utils and we don't have to bind these functions in every component.

src/utils/openCloseComponent.ts
```diff
@@ -8,20 +8,50 @@ export interface OpenCloseComponent {
    */
   readonly el: HTMLElement;
 
+  /**
+   * When true, the component opens.
+   */
+  open?: boolean;
```
driskull, line 14: We'll need this to not be optional before we merge. 👍

src/utils/openCloseComponent.ts
```diff
@@ -8,20 +8,50 @@ export interface OpenCloseComponent {
...
+
   /**
    * Defines method for `close` event handler:
    */
   onClose: () => void;
 }
+
+export function transitionStartHandler(event: TransitionEvent): void {
+  if (event.propertyName === this.activeTransitionProp && event.target === this.transitionEl) {
+    this.active || this.open ? this.onBeforeOpen() : this.onBeforeClose();
```
driskull, line 49: Remove `this.active ||` We should just use `open`.

src/utils/openCloseComponent.ts
```diff
@@ -8,20 +8,50 @@ export interface OpenCloseComponent {
...
 
+  /**
+   * When true, the component opens.
+   */
+  open?: boolean;
+
+  /**
+   * Specifies property on which active transition is watched for.
+   */
+  activeTransitionProp: string;
```
driskull, line 19: We should probably rename this to just `transitionProperty` or `openTransitionProperty`?

src/utils/openCloseComponent.ts
```diff
@@ -8,20 +8,50 @@ export interface OpenCloseComponent {
...
+
+export function transitionStartHandler(event: TransitionEvent): void {
+  if (event.propertyName === this.activeTransitionProp && event.target === this.transitionEl) {
+    this.active || this.open ? this.onBeforeOpen() : this.onBeforeClose();
+  }
+}
+
+export function transitionEnd(event: TransitionEvent): void {
+  if (event.propertyName === this.activeTransitionProp && event.target === this.transitionEl) {
+    this.active || this.open ? this.onOpen() : this.onClose();
```
driskull, line 55: Remove `this.active ||` We should just use `open`.

driskull (commented):
comments!

src/components/alert/alert.tsx
```diff
@@ -301,6 +300,10 @@ export class Alert implements OpenCloseComponent {
 
   transitionEl: HTMLDivElement;
 
+  private transitionStartHandler = transitionStartHandler.bind(this);
+
+  private transitionEnd = transitionEnd.bind(this);
```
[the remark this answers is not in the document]
Elijbet: Recapping our offline conversation here: I tried to implement the arrow function and pass in component but arrow function doesn't take this as an argument and event listener signature expects this as the first argument and event as second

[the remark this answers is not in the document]
Elijbet: Recapping our online conversation here: 
I tried to implement the arrow function and pass in the component but the arrow function doesn't take this as an argument and the event listener signature expects this as the first argument and event as the second. 
The approach here was to resolve the memory leak that would happen if I did bind when passing these functions as a callback to event listeners.
Because you brought up that point, he's suggesting I implement the connect pattern instead to make it consistent with other utils.

src/utils/openCloseComponent.ts
Elijbet: Ok, I think `alert` is the component I'm waiting for to deprecate the `active` prop in favor of `open`. And then I'll make `open` mandatory.

src/components/alert/alert.tsx
```diff
@@ -144,9 +148,13 @@ export class Alert implements OpenCloseComponent {
     this.requestedIcon = setRequestedIcon(StatusIcons, this.icon, this.color);
   }
 
+  componentDidLoad(): void {
+    connectOpenCloseComponent(this);
```
driskull, line 152: maybe instead of calling this on `componentDidLoad()` we can call it in `setTransitionEl()` after the element has been set? That way we can ensure the transitionEl exists when this is called.

driskull: This applies to all components using this interface.

src/utils/openCloseComponent.ts
```diff
@@ -8,20 +8,66 @@ export interface OpenCloseComponent {
...
+function transitionEnd(event: TransitionEvent): void {
+  if (event.propertyName === this.openTransitionProp && event.target === this.transitionEl) {
+    this.open ? this.onOpen() : this.onClose();
+  }
+}
+
+let boundOnTransitionStart;
+let boundOnTransitionEnd;
+
+export const connectOpenCloseComponent = (component: OpenCloseComponent): void => {
```
driskull, line 62: awesome 👏 

Changes to this file look great

driskull (changes requested):
Just one suggestion 👍 💯

src/utils/openCloseComponent.ts
```diff
@@ -8,20 +8,66 @@ export interface OpenCloseComponent {
...
+  }
+}
+
+function transitionEnd(event: TransitionEvent): void {
+  if (event.propertyName === this.openTransitionProp && event.target === this.transitionEl) {
+    this.open ? this.onOpen() : this.onClose();
+  }
+}
+
+let boundOnTransitionStart;
```
jcfranco, line 59: Please type.

src/demos/dropdown.html
```diff
@@ -280,9 +280,9 @@
       </div>
     </div>
 
-    <!-- icon-end + multi select -->
+    <!-- icon-end + multi select I AM HERE-->
```
jcfranco, line 283: Remove test comments.

src/utils/openCloseComponent.ts
```diff
@@ -8,20 +8,67 @@ export interface OpenCloseComponent {
...
```
jcfranco, line 59: You need to keep a map of listeners. Otherwise, `disconnectOpenCloseComponent` could end up not removing listeners properly. For example:

1. `openCloseComp1` (oCC1) connects
  * `boundOnTransitionStart/End` is updated and references functions created for oCC1
  * listeners for oCC1 added to oCC1
2. `openCloseComp2` (oCC2) connects
  * `boundOnTransitionStart/End` is updated and references functions created for oCC2
  * listeners for oCC2 added to oCC2
3. `openCloseComp1` disconnects
  * `boundOnTransitionStart/End` **still references functions created for oCC2** ⚠️
  * attempts removing listeners for oCC2 from oCC1 – no listeners get removed

src/components/alert/alert.tsx
jcfranco: Seconded. 

Additionally, `componentDidLoad` runs only once in a component's lifecycle, so if you were to move the component to another element, `disconnectedCallback` would remove the listener and it would no longer be reattached. I think the same could happen if we only set it in `setTransitionEl` (e.g., util is wired up on render, element is removed then reattached but does not re-render). We could additionally introduce a util solely for updating the reference. I suggest this to avoid confusion on having the set up util be called on both `connectedCallback` and `ref` hooks.

Sidebar: We've been using `(dis)connect<Util>` for all of these component utils to guide devs on where to use it. Maybe it's time to change the naming convention to something more generic (e.g., `setUp`/`tearDown`) and rely more on the util doc?

src/utils/openCloseComponent.ts
Elijbet: Yeah, that's a good point. I'll map the listeners.

src/components/alert/alert.tsx
jcfranco: > We could additionally introduce a util solely for updating the reference. I suggest this to avoid confusion on having the set up util be called on both connectedCallback and ref hooks.

Scrap this. Synced up w/ @Elijbet, and there really wouldn't be much difference between the connect method and the ref updating util. Renaming the utils to move away from `connect`/`disconnect` + util doc should help avoid confusion. To recap: the connect util should be called in `setTransitionEl` and `connectedCallback` to guarantee the node is available.

src/utils/openCloseComponent.ts
```diff
@@ -8,20 +8,90 @@ export interface OpenCloseComponent {
...
+  }
+}
+
+/**
+ * Helper to tear down transition listeners on disconnectedCallback on OpenCloseComponent components.
+ *
+ * @param component
+ */
+export function disconnectOpenCloseComponent(component: OpenCloseComponent): void {
+  const [start, end] = componentToTransitionListeners.get(component);
```
driskull, line 92: Should there be a check to make sure the map has the component?

Like....

```
if(!componentToTransitionListeners.has(component)){
return;
}
```

src/utils/openCloseComponent.ts
```diff
@@ -8,20 +8,90 @@ export interface OpenCloseComponent {
...
+  }
+}
+/**
+ * Helper to keep track of transition listeners on setTransitionEl and connectedCallback on OpenCloseComponent components.
+ *
+ * @param component
+ */
+export function connectOpenCloseComponent(component: OpenCloseComponent): void {
+  if (component.transitionEl) {
+    const [start, end] = componentToTransitionListeners.get(component);
```
driskull, line 70: Do you think that we should just call `disconnectOpenCloseComponent(component);` here and save some lines of code?

driskull (approved):
fabulous 💯

src/utils/openCloseComponent.ts
```diff
@@ -66,13 +66,11 @@ function transitionEnd(event: TransitionEvent): void {
  * @param component
  */
 export function connectOpenCloseComponent(component: OpenCloseComponent): void {
+  if (componentToTransitionListeners.has(component)) {
```
driskull, line 69: I don't think you need this check here. We want it to proceed to disconnect and add new entry.

driskull (approved):
🌈

jcfranco (approved):
Aside from a couple of comments, this LGTM! 🎉

Not bringing up testing as there is a follow-up issue for this (#4968).

jcfranco (approved):
🎉🎉🎉

Issue #4969, opened 2022-07-19
calcite-panel:  setting `closed` to false doesnt remove `dismissed`


kevindoshier:
### Actual Behavior

When setting the `closed` attribute on calcite-panel to false, the deprecated `dismissed` attribute is not also removed

### Expected Behavior

`dismissed` should be removed when setting `closed` to false

### Reproduction Sample

https://codepen.io/kevindoshier/pen/abYJXZZ

### Reproduction Steps

1. Open the sample above
2. Click the "X" and wait for panel to toggle
3. Observe that the calcite-panel dismissed flag is still set

### Reproduction Version

beta.86

driskull:
cc @anveshmekala

tannerjt:
I encountered this same issue.  As a temporary fix, I'm setting both `closed` and `dismissed` props on `calcite-panel`.  It looks like the [draft PR](https://github.com/Esri/calcite-components/pull/4977) will resolve the issue once merged.

benelan:
verified on next

PR #4977, opened 2022-07-20
fix(panel): removes dismissed prop on close

anveshmekala:
**Related Issue:** #4969 

## Summary

This fix will remove `dismissed` prop when the user sets the `closed` prop to `true`


src/components/panel/panel.tsx
```diff
@@ -56,8 +56,14 @@ export class Panel implements ConditionalSlotComponent, InteractiveComponent {
...
+  dismissedHandler(value: boolean): void {
+    this.closed = value;
+    this.calcitePanelDismissedChange.emit();
+  }
+
   @Watch("closed")
-  dismissedHandler(): void {
+  closedHandler(value: boolean): void {
+    this.dismissed = value;
     this.calcitePanelDismissedChange.emit();
```
driskull, line 67: Does this still need to be here? Can you make sure the event isnt being fired twice?

src/components/panel/panel.tsx
```diff
@@ -251,8 +275,8 @@ export class Panel implements ConditionalSlotComponent, InteractiveComponent {
   };
 
   dismiss = (): void => {
-    this.dismissed = true;
     this.closed = true;
+    this.closable = false;
```
driskull, line 279: why is closable being set here? closable is independent of closing or dismissing right?

driskull (changes requested):
Some questions here

src/components/panel/panel.tsx
anveshmekala: It is firing twice. limited it to single emit now.

src/components/panel/panel.tsx
anveshmekala: yup, closable is independent of dismissing.

src/components/panel/panel.tsx
driskull: Can we move this to the dismissed watcher?

src/components/panel/panel.tsx
```diff
@@ -251,7 +278,6 @@ export class Panel implements ConditionalSlotComponent, InteractiveComponent {
   };
 
   dismiss = (): void => {
```
driskull, line 280: nitpick: maybe we should rename this to `close`

src/components/panel/panel.tsx
```diff
@@ -381,19 +407,18 @@ export class Panel implements ConditionalSlotComponent, InteractiveComponent {
...
-      closable || dismissible ? (
-        <calcite-action
-          aria-label={text}
-          icon={ICONS.close}
-          onClick={dismiss}
-          ref={this.setDismissRef}
-          text={text}
-        />
-      ) : null;
+    const dismissibleNode = closable ? (
```
driskull, line 413: nitpick: Maybe we should rename this to `closeNode`

driskull (approved):
🎉 Looks good 🎉 
🌮

src/components/panel/panel.tsx
```diff
@@ -196,7 +223,7 @@ export class Panel implements ConditionalSlotComponent, InteractiveComponent {
   @Event() calcitePanelDismiss: EventEmitter;
 
   /**
-   * Fires when the close button is clicked.
+   * Fires when there is a change in dismissed prop value .
```
benelan, line 226: Nitpick:
We try not to abbreviate words in the doc. Also property/slot/event/css-var names should be in back ticks. So:

> Fires when there is a change to the \`dismissed\` property value.

src/components/panel/panel.tsx
anveshmekala: Do we have to change `setDismissRef` to `setCloseRef` and  `dismissButtonEl` to `closeButtonEl` . thoughts?

driskull: That would be nice too.

Issue #2124, opened 2021-05-11
A11y: keyboard navigation in Tabs is counterintuitive


janett-baresel:
### Summary
When navigating through tabs via keyboard it is cumbersome to get from the tab to its content. 

Additionally, both sighted and non-sighted users (screen reader users) are unable to navigate tab content. Screen readers are not provided with any context residing in the `calcite-tab` component.

### Actual Behavior / Reproduction Steps
1. tab through tabs via keyboard - navigation goes from one to the next tab 👍 
2. press enter on one of the tabs - tab gets activated 👍 
3. tab again and the focus jumps to the next tab. Shift tab and the focus jumps to the previous tab. 
4. Shift tab all the way to the very first tab. Shift tab again and the focus jumps to the first element inside the tab container. 👎 

### Expected Behavior
1. tab through tabs via keyboard - navigation goes from one to the next tab
2. press enter on one of the tabs - tab gets activated 
3. tab again and the focus jumps into the tab container to the first element.

### Reproduction Sample
https://developers.arcgis.com/calcite-design-system/components/tabs

### Reproduction Version
beta.83

### Relevant Info
- [x] JAWS and Chrome/Firefox
- [x] NVDA and Firefox
- [x] Mac OS and VoiceOver

### Working W3C Example/Tutorial 
https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-manual.html

#### Code sample
```html
<div class="tabs">
  <div role="tablist" aria-labelledby="tablist-1" class="automatic">
    <button id="tab-1" type="button" role="tab" aria-selected="true" aria-controls="tabpanel-1">
      <span class="focus">Maria Ahlefeldt</span>
    </button>
    <button id="tab-2" type="button" role="tab" aria-selected="false" aria-controls="tabpanel-2" tabindex="-1">
      <span class="focus">Carl Andersen</span>
    </button>
  </div>

  <div id="tabpanel-1" role="tabpanel" tabindex="0" aria-labelledby="tab-1">
    <p>Maria Theresia Ahlefeldt (16 January 1755 – 20 December 1810) was a Danish, (originally German), composer.</p>
  </div>
  <div id="tabpanel-2" role="tabpanel" tabindex="0" aria-labelledby="tab-2" class="is-hidden">
    <p>Carl Joachim Andersen (29 April 1847 – 7 May 1909) was a Danish flutist, conductor and composer born in Copenhagen, son of the flutist Christian Joachim Andersen.</p>
[...]

PR #4979, opened 2022-07-20
fix(tab, tab-title): Fix accessibility and keyboard navigation

driskull:
**Related Issue:** #2124

## Summary

fix(tab, tab-title): Fix accessibility and keyboard navigation. (#2124)

### DIFF
--- a/src/components/tab-title/tab-title.scss
+++ b/src/components/tab-title/tab-title.scss
@@ -9,25 +9,25 @@
   flex-basis: theme("spacing.48");
 }
 
-:host([position="below"]) a {
+:host([position="below"]) .container {
   @apply border-t-color-transparent
     border-b-0
     border-t-2;
   border-block-start-style: solid;
 }
 
-:host([position="bottom"]) a {
+:host([position="bottom"]) .container {
   @apply border-t-color-transparent
     border-b-0
     border-t-2;
   border-block-start-style: solid;
 }
 
-:host a {
+:host .container {
   @apply focus-base;
 }
 
-:host(:focus) a {
+:host(:focus) .container {
   @apply focus-inset;
 }
 
@@ -39,42 +39,37 @@
   }
 }
 
-:host([active]) a {
+:host([active]) .container {
   @apply text-color-1 border-color-transparent;
 }
 
 @include disabled() {
-  span,
-  a {
+  .container {
     @apply pointer-events-none opacity-50;
   }
 }
 
 :host([scale="s"]) {
   margin-inline-end: 1rem;
-  a,
-  span {
+  .container {
     @apply text-n2h py-1;
   }
 }
 
 :host([scale="m"]) {
-  a,
-  span {
+  .container {
     @apply text-n1h py-2;
   }
 }
 
 :host([scale="l"]) {
   margin-inline-end: 1.5rem;
-  a,
-  span {
+  .container {
     @apply text-0h py-3;
   }
 }
 
-a,
-span {
+.container {
   @apply border-b-color-transparent
     text-color-3
     text-n1h
@@ -91,10 +86,6 @@ span {
     px-0
     py-2;
   border-block-end-style: solid;
-}
-
-span {
-  @apply cursor-default;
 }
 
 .calcite-tab-title--icon {
@@ -146,53 +137,50 @@ span {
 :host([bordered]:hover),
 :host([bordered]:focus),
 :host([bordered]:active) {
-  a {
+  .container {
     @apply relative;
   }
 }
 
 :host([bordered]:hover) {
-  a {
+  .container {
     background-color: var(--calcite-button-transparent-hover);
   }
 }
 
-:host([bordered]) a {
+:host([bordered]) .container {
   border-block-end-style: unset;
   border-inline-start: 1px solid transparent;
   border-inline-end: 1px solid transparent;
 }
 
-:host([bordered][position="below"]) a {
+:host([bordered][position="below"]) .container {
   border-block-start-style: unset;
 }
 
-:host([bordered][position="bottom"]) a {
+:host([bordered][position="bottom"]) .container {
   border-block-start-style: unset;
 }
 
-:host([active][bordered]) a {
+:host([active][bordered]) .container {
   border-inline-start-color: var(--calcite-ui-border-1);
   border-inline-end-color: var(--calcite-ui-border-1);
 }
 
 :host([bordered]) {
-  a,
-  span {
+  .container {
     @apply px-3;
   }
 }
 
 :host([bordered][scale="s"]) {
-  a,
-  span {
+  .container {
     @apply px-2;
   }
 }
 
 :host([bordered][scale="l"]) {
-  a,
-  span {
+  .container {
     @apply px-4;
   }
 }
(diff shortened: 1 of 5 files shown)


driskull:
@geospatialem do you think you could review this PR?

src/components/tab/tab.tsx
```diff
@@ -66,10 +66,10 @@ export class Tab {
 
     return (
       <Host
-        aria-expanded={toAriaBoolean(this.active)}
         aria-labelledby={this.labeledBy}
         id={id}
         role="tabpanel"
+        tabIndex={this.active ? 0 : -1}
       >
```
geospatialem, line 73: Beautiful! ✨ 

JAWS had some trouble handling this event until it was placed into a div as a child of the Host. Would it be possible for us to add these attributes to a child `<div>`?

```ts
      <Host>
        <div
          aria-labelledby={this.labeledBy}
          id={id}
          role="tabpanel"
          tabIndex={this.active ? 0 : -1}
        >
          <section>
            <slot />
          </section>
        </div>
      </Host>
```

geospatialem (commented):
Very nice fixes! 🔨 

One small recommendation, if its feasible, to work with JAWS in Chrome and FF.

src/components/tab/tab.tsx
driskull: Does JAWS not work with custom elements? I'm not sure if we could move that. I'd rather not if possible.

geospatialem: I'm not sure what may be causing the issue, but don't think its the custom element since it seems we've got a few use cases in other components?

I put in the change before checking NVDA first, and it also affects NVDA. So between the two of them would account for the majority of screen reader users (over 80% of sr users).

Is there another option we could consider to mitigate for JAWS and NVDA users?

Some background on the tests and the JAWS transcripts:

### Test experience:
tab-test


### JAWS transcript with attrs in the `Host`:
The content is not read back when focused, instead the title is (in this case 'Tab 1 Title' and its role 'Tab').
```
Tab 1 Title Tab selected 
1 of 3
Tab 1 Title 
Tab 
```


### And the `div`:
```
Tab 1 Title Tab selected 
1 of 3
Tab 1 Content
```

driskull: I don't think we can move these:

```
aria-labelledby={this.labeledBy}
          id={id}
```

They won't be able to be found by the tab title element.

geospatialem: Would this work?

```ts
 <Host aria-labelledby={this.labeledBy} id={id}>
    <div role="tabpanel" tabIndex={this.active ? 0 : -1}>
      <section>
        <slot />
      </section>
    </div>
  </Host>
```

This is the JAWS transcript with the above code (Also works with NVDA).
```
Tab 1 Title Tab selected 
1 of 3
Tab 1 Title 
Tab 1 Content
```

driskull: According to W3c, those should all be on the same element. https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-manual.html

driskull: This is one of those things where I start to wonder if we should follow w3c or the screen readers. I tend to defer to w3c so we are at least following the directions and ensuring the proper structure of code.

geospatialem: Hear ya on that front. FWIW, the sample does have a note on it indicating it shouldn't be used in production environments due to support gaps with browsers and AT. Maybe this is one of those use cases?

src/components/tab-title/tab-title.tsx
```diff
@@ -184,7 +185,9 @@ export class TabTitle implements InteractiveComponent {
   }
 
   componentDidRender(): void {
-    updateHostInteraction(this, true);
+    updateHostInteraction(this, () => {
+      return this.active;
```
driskull, line 189: Added this so that tabindex wasn't conflicting with `updateHostInteraction` utility

src/components/tab-title/tab-title.tsx
```diff
@@ -228,13 +231,15 @@ export class TabTitle implements InteractiveComponent {
         e.preventDefault();
         break;
       case "ArrowRight":
+        e.preventDefault();
```
driskull, line 234: Added these so horizontal scrolling does not occur when arrow keys were pressed.

src/components/tab-title/tab-title.scss
```diff
@@ -146,53 +137,50 @@ span {
...
 
-:host([active][bordered]) a {
+:host([active][bordered]) .container {
   border-inline-start-color: var(--calcite-ui-border-1);
   border-inline-end-color: var(--calcite-ui-border-1);
 }
 
 :host([bordered]) {
-  a,
-  span {
```
driskull, line 181: I'm not sure why `span` was here. No span is being rendered in the component.

src/components/tab-title/tab-title.scss
```diff
@@ -93,10 +88,6 @@ span {
   border-block-end-style: solid;
 }
 
-span {
```
driskull, line 96: I'm not sure why `span` was here. No span is being rendered in the component. cc @macandcheese @paulcpederson

driskull:
Ready for a dev to review

src/components/tab/tab.tsx
jcfranco: > This is one of those things where I start to wonder if we should follow w3c or the screen readers. I tend to defer to w3c so we are at least following the directions and ensuring the proper structure of code.

I think we'll have to rely on our best judgement for these cases since ARIA guidance doesn't seem to mention custom elements/shadow DOM (at least to my knowledge) and sometimes screen readers don't do the right thing in this context. I like the idea of being spec-compliant, but it's not much use if screen readers can't use it.

jcfranco (approved):
Code changes LGTM. I'll defer to @geospatialem regarding screen reader changes.

src/components/tab/tab.tsx
driskull: I think my best judgement here would be to not move the `id` and `aria-labeledby` to an element within the shadowRoot since those IDs would not be able to reference the element outside of the shadowRoot.

```jsx
aria-labelledby={this.labeledBy}
          id={id}
```

geospatialem:
> Code changes LGTM. I'll defer to @geospatialem regarding screen reader changes.

JAWS is still unable to access any content contained in the `calcite-tab`. Since JAWS is the [top used screen reader](https://webaim.org/projects/screenreadersurvey9/#primary), hesitant to say we're good to go.

NVDA is able to access and hear the content (there were a few updates last week, so _could_ have mitigated a fix).

> think my best judgement here would be to not move the `id` and `aria-labeledby` to an element within the shadowRoot since those IDs would not be able to reference the element outside of the shadowRoot.

@driskull Is there another route we could try and take to mitigate for JAWS users? Attempted a few other potential options, but wasn't able to find a successful solution outside of the [child element](#discussion_r926077563).

driskull:
@geospatialem we can move them to the child element if we want to. I'm not sure how its possible for it to work but if it does I'm ok with doing that. As long as aXe doesn't complain about it :)

geospatialem:
> we can move them to the child element if we want to. I'm not sure how its possible for it to work but if it does I'm ok with doing that. As long as aXe doesn't complain about it :)

Will put in a PR shortly for review. 👍🏻

Issue #4980, opened 2022-07-20
calcite-dropdown in calcite-card will not extend outside card


mjuniper:
### Actual Behavior

If you use a calcite-dropdown in a calcite-card, the dropdown content is clipped by the card.

### Expected Behavior

calcite-dropdown should work when in a calcite-card.

### Reproduction Sample

https://codepen.io/mjuniper/pen/gOeWRJp?editors=1000

### Reproduction Steps

1. visit https://codepen.io/mjuniper/pen/gOeWRJp?editors=1000
2. click the dropdown trigger
3. observe that the dropdown content is contained by the card.

### Reproduction Version

beta.85

### Relevant Info

I believe this was introduced by https://github.com/Esri/calcite-components/pull/4908 (sorry!)

### Regression?

beta.83

jcfranco:
As a workaround, the dropdown should be able to escape the card by using `overlay-positioning: fixed`. https://codepen.io/jcfranco/pen/eYMWEoo?editors=1000

mjuniper:
Oh, great point, thank you!

mjuniper:
I think that since the [proposed fix](https://github.com/Esri/calcite-components/pull/4981) has [problems](https://github.com/Esri/calcite-components/pull/4981#discussion_r926073889), the workaround is a more than sufficient solution.

PR #4981, opened 2022-07-20
fix(card): Remove overflow:hidden

driskull:
**Related Issue:** #4980

## Summary

fix(card): Remove overflow:hidden. (#4980)


src/components/card/card.scss
```diff
@@ -110,6 +109,8 @@
 
 .thumbnail-wrapper {
   @apply flex;
+  border-start-start-radius: var(--calcite-border-radius-base);
+  border-start-end-radius: var(--calcite-border-radius-base);
```
mjuniper, line 113: This is the obvious solution but I think it breaks down in a case like this: https://codepen.io/mjuniper/pen/qBoNJpv?editors=1100.

driskull: yeah. It does :\

driskull: We might just have to leave it up to the user to apply a radius on their slotted image. Thoughts @jcfranco ?

mjuniper: This case makes it kinda gnarly and I have not been able to find a solution yet.

mjuniper: I thought maybe clip-path might help us here but I have not been able to make that work.

driskull: We might just have to round all corners

jcfranco: I'll defer to @macandcheese on this one.

Issue #4996, opened 2022-07-21
modal,dropdown, combobox : deprecated `active` prop is not acting as alias with `open` 


anveshmekala:
### Actual Behavior

Whenever the user sets either of `active` or `open` prop, the other one should be in sync and deprecated prop should act as an alias. The same applies when the user removes either of the props.

E.g:
`<calcite-combobox open active> </calcite-combobox>` 

When the user removes the `open` prop, then the `active` prop still stays. 

### Expected Behavior

- If the user sets `active` prop via DOM manipulation, parsing as prop or setting it programmatically `open` prop should be added as well and vice-versa. 
- 
- If the user removes `active` prop via DOM manipulation , parsing as prop or setting it programmatically `open` prop should be added as well and vice-versa. 

### Reproduction Sample

https://codepen.io/ANVESHMEKALA/pen/KKomjgd?editors=1010

### Reproduction Steps

1. Click on the combobox to set `open` to `true`
2. Both `open` & `active` attributes will be appended to combobox in DOM.
3. When you remove `open` attribute from DOM , `active` still stays.

### Reproduction Version

beta.86

### Relevant Info

Related to [4775](https://github.com/Esri/calcite-components/pull/4775)

benelan:
I tested this on `next`, you still need to set both `active` and `open` to false in order to close the combobox:
https://codepen.io/benesri/pen/abYYewb?editors=1010

Assigning this back to dev for another look

jcfranco:
I believe this is working as expected. The default for both is false, so we honor whichever is present and can't assume the lack of `open` means closed. To summarize:

1. none specified ➡️ closed
2. any specified (both or either) ➡️ open

The codepen does bring up some subtle behavior related to updating components before hydration. It looks like setting attributes before hydration does populate the property correctly (see https://github.com/Esri/calcite-components/issues/4762#issuecomment-1193099704), but it [doesn't work the other way around](https://codepen.io/jcfranco/pen/poLZeXO?editors=1011). Not sure if this is by design or not. 

LMK if I missed something.

benelan:
> 1. none specified ➡️ closed
> 2. any specified (both or either) ➡️ open

This contradicts what we [discussed in a related PR here](https://github.com/Esri/calcite-components/pull/4807#issuecomment-1202720528), specifically:

> replacement properties take precedence over deprecated properties when both are set.

[...]

PR #4997, opened 2022-07-21
fix(dropdown,modal,combobox): active prop will be in sync with open prop

anveshmekala:
**Related Issue:** #4996 

## Summary

This will make sure `active` and `open` prop will be sync and if user adds/removes one of the them the other will follow and vice-versa.


src/components/combobox/combobox.tsx
```diff
@@ -96,19 +96,29 @@ export class Combobox
   @Prop({ reflect: true, mutable: true }) active = false;
 
   @Watch("active")
-  @Watch("open")
-  activeHandler(): void {
+  activeHandler(value: boolean): void {
     if (this.disabled) {
```
driskull, line 100: I think you can get rid of all the logic here and just have the handler set the open value.

src/components/dropdown/dropdown.tsx
```diff
@@ -63,18 +63,28 @@ export class Dropdown implements InteractiveComponent, OpenCloseComponent, Float
    */
   @Prop({ reflect: true, mutable: true }) active = false;
 
+  @Watch("active")
+  activeHandler(value: boolean): void {
+    if (!this.disabled) {
```
driskull, line 68: I think you can get rid of all the logic here and just have the handler set the open value. Let the open watch handler do the logic.

driskull (approved):
👍

src/components/combobox/combobox.tsx
driskull, line 100: Can you remove this logic so that only `open` handles it?

src/components/combobox/combobox.tsx
```diff
@@ -96,19 +96,29 @@ export class Combobox
...
     }
-
-    this.setMaxScrollerHeight();
+    this.open = value;
   }
 
   /**When true, opens the combobox */
-  @Prop({ reflect: true, mutable: true }) open = false;
+  @Prop({ reflect: true, mutable: true })
+  open = false;
```
driskull, line 110: I think open should be on the same line as the prop.

driskull (commented):
Are there any `.scss` or `.stories.ts` changes that need to happen as well?

anveshmekala:
> Are there any `.scss` or `.stories.ts` changes that need to happen as well?

There aren't any screen stories that require a change. I guess we could remove styles with `[active]` selectors.

src/components/dropdown/dropdown.tsx
```diff
@@ -163,6 +166,11 @@ export class Dropdown implements InteractiveComponent, OpenCloseComponent, Float
     this.mutationObserver?.observe(this.el, { childList: true, subtree: true });
     this.setFilteredPlacements();
     this.reposition();
+    const isOpen = this.active || this.open;
```
jcfranco, line 169: Can you check if the property is set from the attribute here? If not, I think this also needs to look at the user-set attributes.

src/components/combobox/combobox.tsx
```diff
@@ -307,13 +316,18 @@ export class Combobox
 
   connectedCallback(): void {
     this.internalValueChangeFlag = true;
+    const isOpen = this.active || this.open;
```
jcfranco, line 319: Nitpick: this could be moved closer to its usage.

src/components/combobox/combobox.tsx
```diff
@@ -613,15 +620,14 @@ export class Combobox
...
-    this.active = !this.active;
     this.open = !this.open;
     this.updateActiveItemIndex(0);
     this.setFocus();
   };
 
   setInactiveIfNotContained = (event: Event): void => {
     const composedPath = event.composedPath();
-    const isOpen = !(this.open || this.active);
+    const isOpen = !this.open;
```
jcfranco, line 630: Const could be inlined since it's only used once here and is pretty expressive.

jcfranco (approved):

src/components/dropdown/dropdown.tsx
anveshmekala: yup, the property is set from the attribute.

jcfranco:
I noticed there's still an issue with modal. Pressing `Escape` no longer closes the modal.

driskull:
Should we open an issue for that @jcfranco @anveshmekala ?

anveshmekala:
I tried reproducing the error and the only case where the escape key cannot close the modal is when the user sets the `active` or `open` prop as an attribute during the initial render. Moreover, I see the issue persistent with `beta.79` and wanted to confirm if it's not by design.
ref:https://codepen.io/ANVESHMEKALA/pen/BarwRBL?editors=1000.

driskull:
The esc key closes it for me in that codepen. As long as i'm within the iframe somewhere.

anveshmekala:
Isn't it supposed to close while the user is somewhere in the window  or it has to be the iframe of the modal ?

driskull:
I'm not sure what codepen does but I wouldn't rely on it as a good testing place. It might be blocking window events or something.

The event listener is on the window.

Issue #4156, opened 2022-02-22
[Shell / Shell-Panel] Allow a single detached shell-panel at the end of the shell in content-behind mode


macandcheese:
### Actual Behavior

Currently you cannot have a single detached `shell-panel` positioned at the right/end of the `shell` when `content-behind` is set (this works as expected with when not in `content-behind`).  A user needs to "fake" an empty `shell-panel` in the `primary` slot in order for one in the `contextual` slot to appear at the right/end.

### Expected Behavior

I would expect to be able to have a single `shell-panel` detached and positioned at the right/end. I might also expect the slots to be named `start-, end-` instead of `primary-, contextual-` since that may not be applicable to my design. cc @asangma 

### Reproduction Sample

https://codepen.io/mtb-alan/pen/NWwzrex?editors=1000

### Reproduction Steps

View codepen above, see "faked" slotted `shell-panel`. Remove that and observe the `contextual-panel` shift to the `primary-panel` position

### Reproduction Version

all

asangma:
I think this calls for deprecating the primary and contextual panel slot names.
I am seeing enough examples of designs that have a single panel on the end side.

@benelan @jcfranco cool?

geospatialem:
Verified on next

PR #5000, opened 2022-07-21
fix(shell): Allow a single detached shell-panel at the end in content-behind mode

driskull:
**Related Issue:** #4156

## Summary

fix(shell): Allow a single detached shell-panel at the end in content-behind mode (#4156)


src/components/shell/shell.tsx
```diff
@@ -11,8 +11,10 @@ import {
  * @slot - A slot for adding content to the shell. This content will appear between any leading and trailing panels added to the shell. (eg. a map)
  * @slot header - A slot for adding header content. This content will be positioned at the top of the shell.
  * @slot footer - A slot for adding footer content. This content will be positioned at the bottom of the shell.
- * @slot primary-panel - A slot for adding the leading `calcite-shell-panel`.
- * @slot contextual-panel - A slot for adding the trailing `calcite-shell-panel`.
+ * @slot panel-start - A slot for adding the starting `calcite-shell-panel`.
+ * @slot panel-end - A slot for adding the ending `calcite-shell-panel`.
+ * @deprecated @slot primary-panel - A slot for adding the leading `calcite-shell-panel`.
```
driskull, line 16: Is this the correct way to deprecate a slot? cc @benelan

driskull:
@asangma @macandcheese this is ready for review

macandcheese:
Screen Shot 2022-07-22 at 10 01 25 AM

src/demos/shell/demo-app-center-row-action-bar.html
```diff
@@ -53,11 +53,12 @@
       });
     </script>
   </head>
+
   <body>
     <main>
       <div class="shell-container">
         <calcite-shell>
-          <calcite-shell-panel id="primary-panel" slot="primary-panel" position="start" id="primary-panel">
```
macandcheese, line 60: Just demo stuff, but a duplicate `id` here.

macandcheese (approved):
Awesome! Naming is much clearer now. I suppose we could have the slots as `shell-panel-start/end` to be a bit more clear that users should slot `shell-panel`, not `panel`, but will defer to others on that.

src/components/shell/shell.tsx
benelan: cc @macandcheese @geospatialem you both found out the jsdoc tag doesn't work for slots as well as css vars right? IIRC we use `[Deprecated]` and then parse it out and add the chip in the doc.

> @ slot primary-panel - [Deprecated] A slot for adding the leading `calcite-shell-panel`.

(added a space between `@` and `slot` so I didn't tag some random guy)

driskull:
> I suppose we could have the slots as shell-panel-start/end to be a bit more clear that users should slot shell-panel, not panel, but will defer to others on that.

@jcfranco what do you think about the slot names?

src/components/shell/shell.tsx
benelan: @driskull I confirmed that adding `[DEPRECATED]` to the description is the correct way to deprecate slots and css vars. We will get that added to the documentation conventions.

driskull:
I changed the slot names to `shell-panel-start` and `shell-panel-end`. @jcfranco let me know if that should change to something else. Otherwise, I think this is good to merge.

jcfranco:
I think the original names (sans `shell`) are a good balance between concise and meaningful. Most of our slot names are contextual and aren't 100% tied to the exact elements that get slotted.

driskull:
@jcfranco I'll put them back to as they were.

Maybe we should discuss the naming of slots more. Some slots like `actions-start` do suggest some type of content that should be slotted.

In this case, it might help distinguish the `calcite-panel` from the `calcite-shell-panel`.

jcfranco:
Yeah, I can see that, but for slots, users are sort of forced to go through the doc/examples as slots don't come through the type declarations nor can be easily discovered like props /methods in the console. I would lean more on adding `shell` this wasn't the case. Do you recall any confusion w/ `primary/secondary-panel`?

jcfranco (approved):
:boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom:
:boom::bomb::bomb::bomb::bomb::boom::boom::bomb::bomb::boom::boom::boom::bomb::bomb::boom::boom::bomb::boom::boom::boom::bomb::boom::bomb::boom:
:boom::bomb::boom::boom::bomb::boom::bomb::boom::boom::bomb::boom::bomb::boom::boom::bomb::boom::bomb::bomb::boom::bomb::bomb::boom::bomb::boom:
:boom::bomb::bomb::bomb::boom::boom::bomb::boom::boom::bomb::boom::bomb::boom::boom::bomb::boom::bomb::boom::bomb::boom::bomb::boom::bomb::boom:
:boom::bomb::boom::boom::bomb::boom::bomb::boom::boom::bomb::boom::bomb::boom::boom::bomb::boom::bomb::boom::boom::boom::bomb::boom::boom::boom:
:boom::bomb::bomb::bomb::bomb::boom::boom::bomb::bomb::boom::boom::boom::bomb::bomb::boom::boom::bomb::boom::boom::boom::bomb::boom::bomb::boom:
:boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom:

PR #5001, opened 2022-07-22
chore(linting): fix event emitter type linting errors

jcfranco:
**Related Issue:** N/A

## Summary

This fixes linting errors reported by the new `require-event-emitter-type` rule and ensures all event emitters have a type. 

### Notes

* added a type to help us keep track of deprecated event payloads – it just aliases `any`, but it's mainly to provide context
* event payload for internal events was mostly typed and kept intact – we will likely refactor this in the future


src/components/date-picker-month/date-picker-month.tsx
```diff
@@ -73,24 +73,24 @@ export class DatePickerMonth {
...
    * @internal
    */
-  @Event() calciteInternalDatePickerHover: EventEmitter;
+  @Event() calciteInternalDatePickerHover: EventEmitter<void>;
 
   /**
    * Active date for the user keyboard access.
    */
-  @Event() calciteDatePickerActiveDateChange: EventEmitter;
+  @Event() calciteDatePickerActiveDateChange: EventEmitter<void>;
```
anveshmekala, line 88: emit type should be `Date` right?

anveshmekala: considering `dateFromRange( )` return type could be `Date`

anveshmekala: Applies to following events as well.
`calciteInternalDatePickerMouseOut`
`calciteInternalDatePickerHover`
`calciteDatePickerSelect`

driskull: Looks like it. Unless we want to encourage the user getting these values from the target which might be preferred.

src/components/split-button/split-button.tsx
```diff
@@ -82,11 +82,15 @@ export class SplitButton implements InteractiveComponent {
...
+  /**
+   * fired when the primary button is clicked
+   *
+   * **Note:** The event payload is deprecated, please use the evevnt listeners to  property on the component instead
+   */
+  @Event() calciteSplitButtonPrimaryClick: EventEmitter<DeprecatedEventPayload>;
 
   /** fired when the secondary button is clicked */
-  @Event() calciteSplitButtonSecondaryClick: EventEmitter;
+  @Event() calciteSplitButtonSecondaryClick: EventEmitter<void>;
```
anveshmekala, line 93: we are emitting `MouseEvent` type in  `calciteSplitButtonSecondaryClickHandler( )`

anveshmekala (changes requested):
🚀 🚀🚀 .

Have some questions, other than those everything else is 🏆  .

src/components/date-picker-month/date-picker-month.tsx
jcfranco: Good catch. The build didn't complain and I hadn't checked the test results. Looks like I also left out some commits. Will fix. 

Sidebar: it looks like all of the events from date-picker supporting components are really internal since they're not exposed and they are not mapped on the date-picker itself (meaning not in the date-picker type snor doc).

anveshmekala: make sense why they didn't show up in the build error.

src/components/action-group/action-group.tsx
```diff
@@ -1,14 +1,15 @@
-import { Component, h, Prop, Watch, Element } from "@stencil/core";
-import { SLOTS, TEXT, ICONS } from "./resources";
+import { Component, Element, h, Prop, Watch } from "@stencil/core";
+import { ICONS, SLOTS, TEXT } from "./resources";
 import { Fragment, VNode } from "@stencil/core/internal";
 import { getSlotted } from "../../utils/dom";
 import { SLOTS as ACTION_MENU_SLOTS } from "../action-menu/resources";
-import { Columns, Layout, Scale } from "../interfaces";
+import { Columns, DeprecatedEventPayload, Layout, Scale } from "../interfaces";
```
anveshmekala: Nitpick : `DeprecatedEventPayload` is not used

anveshmekala (approved):
🎉

PR #5004, opened 2022-07-22
ci: deploy storybook after next releases

benelan:
**Related Issue:** https://github.com/Esri/calcite-components/pull/4891#pullrequestreview-1033471305

## Summary
Storybook Deployment action is failing [(logs here)](https://github.com/Esri/calcite-components/runs/7373968016?check_suite_focus=true#step:4:86) like I predicted in the comment linked above.

This PR _should_ fix the error. The error was also happening with chromatic, and I tested the fix on my chromatic branch and the build is working there. I'm guessing `concurrently` was causing a problem for some reason.

Additionally, storybook now deploys after every `next` release, rather than once a week. Storybook should be used for `next` testing and snapshot testing, wheras the developer doc should be used for stable API info. I added the storybook deployment step at the end of the workflow, so even if it doesn't work it won't affect `next` deployments.


driskull (approved):
👏

Issue #2380, opened 2021-06-23
Bug: Action pad expanded on mobile device size doesn't work correctly


driskull:
### Actual Behavior

I have a Calcite Action Pad defined as a sibling to a Calcite Shell component which contains a header and my mapView div like this:

```html
<calcite-action-pad style="position:absolute;margin-top:4px;margin-left:4px;visibility:hidden;" expanded layout="vertical" id="hamburgerMenu">
      <calcite-action id="menu1" onclick="window.open('aURL','_blank')" text="Frequently Asked Questions" icon="question-mark"></calcite-action>
      <calcite-action id="menu2" onclick="window.open('aURL','_blank')" text="FWP Buy and Apply" icon="shopping-cart"></calcite-action>
      <calcite-action id="menu3" onclick="window.open('aURL','_blank')" text="TIP-MONT" icon="star-circle"></calcite-action>
      <calcite-action id="menu4" onclick="window.open('aURL','_blank')" text="Submit Feedback" icon="edit-attributes"></calcite-action>
      <calcite-action id="menu5" onclick="document.getElementById('myModal-disclaimer').style.display = 'block'" text="Show Disclaimer" icon="exclamation-mark-circle"></calcite-action>
  </calcite-action-pad>
```

It is not visible by default but is triggered by a click event on an icon. On desktop the labels show up as expected but on mobile they do not show up at all. When expanded the action pad expands by a few pixels but no text appears. Is this a known issue, or something I'm doing wrong here?

### Expected Behavior

action pad expands correctly on mobile device sized screens

### Reproduction Steps or Sample

see above

### Relevant Info

https://community.esri.com/t5/calcite-design-system-questions/action-pad-labels/m-p/1069514#M18

caripizza:
This looks to me like it's from the action-pad's .container class, that has `max-width: 15vw;` style (I think even when expanded):
https://github.com/Esri/calcite-components/blob/master/src/components/calcite-action-pad/calcite-action-pad.scss#L25

Should the width match action-bar and instead be `20vw;` to show more of the truncated text labels?

samMatenaer:
Should the width of this component really change based on screen width? If anything, rather than resizing the entire action pad, it might be better to just auto collapse the pad? For example, in the image below, the text should not be cut off.

<img width="556" alt="image" src="https://user-images.githubusercontent.com/50883428/128533549-d996e5a7-2e00-4aa5-ae91-8a156513fd69.png">

[...]

PR #5006, opened 2022-07-22
fix(action-bar, action-pad): Allow users to define custom max-width with CSS

driskull:
**Related Issue:** #2380

## Summary

fix(action-bar, action-pad): Allow users to define custom max-width with CSS

Users can use the new custom CSS properties to define max widths for action-bar and action-pad

### DIFF
--- a/src/components/action-bar/action-bar.scss
+++ b/src/components/action-bar/action-bar.scss
@@ -3,7 +3,7 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
- * @prop --calcite-action-bar-expanded-max-width: optionally specify the expanded max width of the action bar.
+ * @prop --calcite-action-bar-expanded-max-width: optionally specify the expanded max width of the action bar when in "vertical" layout.
  */
 
 :host {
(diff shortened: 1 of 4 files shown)


driskull:
Can I get a review on this one?

src/components/action-bar/action-bar.scss
```diff
@@ -1,10 +1,18 @@
...
+ */
+
 :host {
   @extend %component-host;
   @apply pointer-events-auto
     inline-flex
     flex-col
     self-stretch;
-  max-inline-size: 15vw;
+  --calcite-action-bar-expanded-max-width: auto;
```
eriklharper, line 15: Is the reason this is being set to `auto` by default so that it has a less opinionated starting value?

driskull: Correct.

driskull: https://github.com/Esri/calcite-components/issues/2380#issuecomment-1192768235

eriklharper (approved):
LGTM 👍

src/components/action-bar/action-bar.scss
jcfranco, line 15: Looking a this name in particular has me thinking if we need them namespaced if they are scoped to the component? If so, we should look into how these are baked into the component styles and whether we need a pattern/util to use a more concise name internally to potentially shave some bits. cc @macandcheese

PR #5011, opened 2022-07-23
ci(next): try to get logs displayed in github action

benelan:
**Related Issue:** #

## Summary
- Fixes a type error in the release prep script
- exports the release prep script as a function to use in the next release. Hopefully that gets us better logging in the GH Action. Currently logs from `deployNextFromCI` show up, and that scripts uses the pify to call the npm script for the prep release script. Now it stays in JS which _might_ make the logs show up? shot in the dark 🤞 
- Removes `echo` calls from the script which was a previous attempt to get logging to work

### DIFF
--- a/support/prepReleaseCommit.ts
+++ b/support/prepReleaseCommit.ts
@@ -3,7 +3,7 @@ import type { Options } from "standard-version";
 import pify from "pify";
 import yargs from "yargs";
 
-(async function prepReleaseCommit(): Promise<void> {
+export async function prepReleaseCommit(): Promise<void> {
   const childProcess = await import("child_process");
   const { promises: fs } = await import("fs");
   const { default: gitSemverTags } = await import("git-semver-tags");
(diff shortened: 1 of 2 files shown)


support/prepReleaseCommit.ts
```diff
@@ -178,4 +172,8 @@ import yargs from "yargs";
 
     await fs.writeFile(readmePath, updatedReadmeContent);
   }
+}
+
+(async function () {
```
jcfranco, line 177: I think this will run the script when you import the function from another module. Can you test?

benelan: I created two files:

```js
// test.js
export function test(string) {
  console.log(string);
}

(function () {
  test("IIFE");
})();
```
and 
```js
// index.js
import { test } from "./test.js";
test("import");
```
I ran `node index.js` and this is what was logged in the console:
```bash
IIFE
import
```
So it doesn't seem like the IIFE runs the script on import, or else it would have been logged twice right?

Issue #4851, opened 2022-07-06
[Filter] Regex error when using special character in search term


macandcheese:
### Actual Behavior

Saw this while using the filter-enabled pick-list, I'm guessing it's at the filter level. Using special characters - [ ] \ - etc., will product a console error:
![Screen Shot 2022-07-06 at 9 07 02 AM](https://user-images.githubusercontent.com/4733155/177595249-32b9dc20-4b03-4c0e-87e3-aa198f70d73b.png)
 

### Expected Behavior

No errors are shown in console with any string in filter field.

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/GRxoMRq?editors=1000

### Reproduction Steps

1 - go to Codepen sample
2 - type any string in filter-enabled pick list and then add `[`  or `]` or `\`
3 - see console log error

### Reproduction Version

83

benelan:
verified on next

PR #5012, opened 2022-07-23
fix(pick-list, value-list): fix error when typing special regex chars

jcfranco:
**Related Issue:** #4851 

## Summary

In addition to fixing the error, this refactors the components to reuse the same filter util.

### DIFF
--- a/src/components/filter/filter.tsx
+++ b/src/components/filter/filter.tsx
@@ -3,18 +3,19 @@ import {
   Element,
   Event,
   EventEmitter,
-  Prop,
-  h,
-  VNode,
-  Method,
   Fragment,
+  h,
+  Method,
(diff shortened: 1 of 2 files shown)


src/components/filter/filter.tsx
```diff
@@ -140,38 +141,10 @@ export class Filter implements InteractiveComponent {
...
-      return found;
-    };
-
-    const result = this.items.filter((item) => {
-      return find(item, regex);
-    });
-
-    this.updateFiltered(result, emit);
-  }, DEBOUNCE_TIMEOUT);
+  filter = debounce(
```
driskull, line 144: beautiful 🌞

driskull (approved):
🌻

Issue #4768, opened 2022-06-16
[Pick List Group] Incorrect display of `heading` when `heading-level` is set


macandcheese:
### Actual Behavior

The Pick List Group is not correctly setting styles for header elements:

The Pick List Group heading styles are scoped to `h3` : https://github.com/Esri/calcite-components/blob/6847f57359897972b002bb5f9377fb4c6c3d5a4f/src/components/pick-list-group/pick-list-group.scss#L20

The List Item Group styles apply to any of the conditionally-rendered heading elements: https://github.com/Esri/calcite-components/blob/6847f57359897972b002bb5f9377fb4c6c3d5a4f/src/components/list-item-group/list-item-group.scss#L5 
<img width="730" alt="Screen Shot 2022-06-16 at 3 25 16 PM" src="https://user-images.githubusercontent.com/4733155/174188349-a652fcba-af2c-4bd4-a1b4-45e16c596bf2.png">

### Expected Behavior

The visual styling of the heading text is not changed when `heading-level` value changes.

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/XWZGqxy?editors=1000

### Reproduction Steps

View style differences on Codepen.

### Reproduction Version

83

macandcheese:
@mitc7862 @asangma - Aside from the bug above for the changing of sizes, none of the group title styles from the pick list matches that of the regular list.

Can we make these more consistent? I don't have a preference but it seems like these could be more aligned - thoughts?

macandcheese:
It seems the styles for nested items also differ between List and Pick List - it would be great to do a design pass to ensure they are consistent, and see if we need a border at all (or an option to enable / disable).

geospatialem:
cc @jcfranco potential linting rule implementation?

driskull:
https://codepen.io/driskull/pen/XWEedNo?editors=1000

geospatialem:
Verified on next.540 ✔️

PR #5014, opened 2022-07-25
fix(pick-list-group, tip, tip-manager, block, panel): headingLevel should not adjust styling

driskull:
**Related Issue:** #4768

## Summary

fix(pick-list-group, tip, tip-manager, block, panel): headingLevel should not adjust styling. (#4768)

### DIFF
--- a/src/assets/styles/_header.scss
+++ b/src/assets/styles/_header.scss
@@ -15,19 +15,3 @@
 .header .heading {
   @apply flex-auto p-2;
 }
-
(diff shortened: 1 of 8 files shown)


macandcheese:
Approved Screener, looks good! There was an unrelated dropdown change show, seems like a case where a dropdown container didn’t match the past screener px for px.

anveshmekala (approved):
LGTM ! 👍

Issue #4772, opened 2022-06-17
calcite-combobox hangs out of the window when used within a JS API widget


gavinr:
### Actual Behavior

When using the `calcite-combox` in a JS API widget, the dropdown options drop outside of the viewable area of the screen

### Expected Behavior

`calcite-combobox` should properly size the dropdown to allow the user to see the entire dropdown no matter the screen size

### Reproduction Sample

https://jsbin.com/maridel/1/edit?html,output

### Reproduction Steps

1. Open the jsbin example (https://jsbin.com/maridel/1/edit?html,output)
2. Click open the widget (underneath the zoom buttons) and click open the combobox.
    - Note all the options that are in the box:
    ![image](https://user-images.githubusercontent.com/209355/174350333-c6b0571a-609e-49f1-8b48-b9c61c41b3ef.png)
3. Close the combobox dropdown.
4. Resize your browser to be smaller
    - Actual: Notice that you cannot see all the options
    - Expected: You should be able to see all the options
    ![image](https://user-images.githubusercontent.com/209355/174350689-243d37bc-ecab-4585-a560-d912a6b6c909.png)

### Reproduction Version

1.0.0-beta.83

### Relevant Info

Is this related to #3815? I'm not totally sure.

gavinr:
I have created a very rough temporary workaround by dynamically modifying the `max-items` property based on screen calculations https://jsbin.com/cibekek/edit?html,output

geospatialem:
Also pertains to `dropdown` component. Resolution is to set the `max-height` to a percentage of the view height.

benelan:
verified on next

PR #5015, opened 2022-07-25
fix(combobox, dropdown): Make sure menus do not take up more than half of available height by default

driskull:
**Related Issue:** #4772

## Summary

fix(combobox, dropdown): Make sure menus do not take up more than half of available height. (#4772)

### DIFF
--- a/src/components/combobox/combobox.scss
+++ b/src/components/combobox/combobox.scss
@@ -166,8 +166,8 @@
 
 .list-container {
   @apply bg-foreground-1
-    max-h-screen
     overflow-y-auto;
+  max-block-size: 50vh;
   inline-size: var(--calcite-dropdown-width);
 }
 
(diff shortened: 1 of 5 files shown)


src/components/combobox/combobox.scss
```diff
@@ -166,8 +166,8 @@
 
 .list-container {
   @apply bg-foreground-1
-    max-h-screen
     overflow-y-auto;
+  max-block-size: 50vh;
```
jcfranco, line 170: This could be a neat util/mixin to add.

jcfranco (approved):
LGTM! Do you think Screener could help keep this in check?

src/components/combobox/combobox.scss
driskull: I'll add a screener and see if we can do a tailwind option for this.

It's still not perfect if you open slightly below half the page and it doesn't open above. We can maybe tweak it to 40%.

PR #5017, opened 2022-07-25
ci: deploy storybook before releasing next

benelan:
**Related Issue:** #

## Summary
The storybook deployment isn't happening because we set up concurrency to batch `next` release commits, which cancels the current run when a new push to master happens. Moving the storybook deployment before the `next` release will fix the issue.

Edits:
- Put the storybook deployment inside of the `deployNextFromCI` script so that storybook only deploys when there are deployable commits
- Make sure that `next` releases even if the storybook deployment fails
    
    image

### DIFF
--- a/.github/workflows/deploy-next.yml
+++ b/.github/workflows/deploy-next.yml
@@ -32,6 +32,8 @@ jobs:
         run: |
           if [ "$NEXT_RELEASE_ENABLED" == "true" ]; then
             npm run util:is-next-deployable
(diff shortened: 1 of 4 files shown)


driskull (approved):
👏

jcfranco (approved):
Thanks for fixing! 

Per our convo, let's look at keeping storybook and NPM deployment scripts separate.

jcfranco (approved):
Awesome! Thanks for splitting this up so quickly!

Issue #4776, opened 2022-06-20
date-picker in pt-PT locale has overlapping day names


BlakeStearman:
### Actual Behavior

When the locale is `pt-PT` the day names for Sunday, Monday, and Tuesday overlap within the calendar.

### Expected Behavior

Users will expect the day names or abbreviations to be clear and legible, not overlapping.

### Reproduction Sample

https://codepen.io/bstearman/pen/jOZomov

### Reproduction Steps

1. Set locale property on the `date-picker` or `input-date-picker` to "pt-PT".
  ```html
  <calcite-date-picker locale="pt-PT"></calcite-date-picker>
  ```
2. Observe the day names overlap between Sunday, Monday, and Tuesday on the date-picker calendar.
  ![image](https://user-images.githubusercontent.com/5124840/174685297-7e21e6df-e0cc-4b53-bdf5-ac03502ece89.png)

### Reproduction Version

v1.0.0-next.499

jcfranco:
@annierm18 Could you advise on this one? Looks like the [abbreviated names](https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-dates-full/main/pt-PT/ca-generic.json#L106-L114) are longer than the [short version](https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-dates-full/main/pt-PT/ca-generic.json#L124-L132) for `pt-PT`.

annierm18:
@jcfranco For this, it would be okay to use the short version instead of the abbreviated version. Once you have a mockup of the component ready with the short version, could you share a screen shot of it in this issue?

jcfranco:
![Screen Shot 2022-07-26 at 12 41 30 PM](https://user-images.githubusercontent.com/197440/181097981-248532bf-bdf7-4a56-86be-e8d17b1d66ec.png)

geospatialem:
Verified on next

Olga8614:
Still observed with PT language on devext 
![image](https://user-images.githubusercontent.com/29614798/183775371-4e970b4c-d1b9-49f9-ac26-d3a3a6648f42.png)

PR #5028, opened 2022-07-26
fix(date-picker): add short day names for pt-PT date picker

jcfranco:
**Related Issue:** #4776 

## Summary

Adds short day entry from CLDR.

cc @annierm18


driskull (approved):
👍

PR #5029, opened 2022-07-26
feat(alert): deprecates active prop and syncs with open

Elijbet:
**Related Issue:** #4688

## Summary

Deprecates `active` prop on `alert` component and syncs with `open`.

### DIFF
--- a/src/components/alert/alert.tsx
+++ b/src/components/alert/alert.tsx
@@ -49,17 +49,31 @@ export class Alert implements OpenCloseComponent {
   //
   //---------------------------------------------------------------------------
 
-  /** When true, the component is active. */
+  /**
+   * When true, opens the combobox
+   *
+   * @deprecated use open instead
+   */
   @Prop({ reflect: true, mutable: true }) active = false;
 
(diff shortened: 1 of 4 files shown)


src/components/alert/alert.tsx
```diff
@@ -118,7 +132,13 @@ export class Alert implements OpenCloseComponent {
   //--------------------------------------------------------------------------
 
   connectedCallback(): void {
-    if (this.active && !this.queued) {
+    if ((this.open || this.active) && !this.queued) {
+      if (this.active) {
+        this.activeHandler(this.active);
```
jcfranco, line 137: I think it'd be clearer and simpler if you use a 'source of truth' var and use it for this whole statement:

```ts
const open = this.open || this.active;
if (open && !this.queued) {
  this.activeHandler(open);
  this.openHandler(open);
}
```

jcfranco (approved):
LGTM!

src/components/alert/alert.tsx
Elijbet: neat! 🧴🧽✨

Issue #1967, opened 2021-04-14
Enhancement: Event on Dropdown Item when SelectionMode is None


tyanthony:
### Description
When the dropdown list has the `selectionMode` set to `none`, the item that is selected is unknown. The `calciteDropdownSelect` event is fired, but we are unable to determine _which_ item was selected. 

### Acceptance Criteria 
Ability to determine _which_ item in the dropdown list was selected when the `selectionMode` is `none`.
Desired Behavior (1 of the following):
- exposed event on the dropdown-item
- detail on the parent event to include the element that was selected
- other?

#### Which Component
[Dropdown](https://esri.github.io/calcite-components/?path=/story/components-buttons-dropdown--groups-and-selection-modes)

macandcheese:
Follow up from Teams - likely have the parent detail include the selected item AND expose the child event as well. Likely applies to other components with parent/child events like this (accordion, tree, stepper, etc...)

geospatialem:
Verified on next.551 with [Codepen](https://codepen.io/geospatialem/pen/RwMQNjM).

![verify-selectnone-dropdown-items](https://user-images.githubusercontent.com/5023024/182239749-f421dc2a-aaf8-4306-8915-69f64808591f.gif)

PR #5030, opened 2022-07-26
feat(dropdown): include selection-triggering item in select event payload

jcfranco:
**Related Issue:** #1967 

## Summary

Adds the most recently selected item to `calciteDropdownSelect`'s payload. This will benefit users in knowing the item that triggered the event, especially when the item group's selection mode is `none`.

### DIFF
--- a/src/components/dropdown/dropdown.tsx
+++ b/src/components/dropdown/dropdown.tsx
@@ -11,7 +11,7 @@ import {
   VNode,
   Watch
 } from "@stencil/core";
-import { ItemKeyboardEvent } from "./interfaces";
+import { ItemKeyboardEvent, Selection } from "./interfaces";
 
 import { focusElement, toAriaBoolean } from "../../utils/dom";
 import {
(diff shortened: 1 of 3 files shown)


src/components/dropdown/dropdown.tsx
```diff
@@ -260,8 +261,12 @@ export class Dropdown implements InteractiveComponent, OpenCloseComponent, Float
   //
   //--------------------------------------------------------------------------
 
-  /** fires when a dropdown item has been selected or deselected */
-  @Event() calciteDropdownSelect: EventEmitter<void>;
+  /**
+   * fires when a dropdown item has been selected or deselected
+   *
```
driskull, line 266: why all the line breaks?

driskull (approved):
👍 👏

src/components/dropdown/dropdown.tsx
jcfranco: vs no line breaks?

why-not-both-why-not

J/k. Force of habit. I find them more readable and easy to add additional notes or tags. I'm sure there's an ESLint rule for this.

jcfranco: Oh, I'm also missing some event info here. That's why there's extra, extra newlines. Fixing.

jcfranco: Removed extra lines and will rely on the type. Spoke w/ @benelan and @macandcheese about introducing a custom JSDoc tag to provide the interface w/o losing the type reference.

PR #5031, opened 2022-07-26
test: fix async issues

jcfranco:
**Related Issue:** N/A

## Summary

* Adds missing `await`s 
* Replaces `Array.forEach` with `for...of` loops to ensure `await` works within loops


driskull (approved):
Nice!

PR #5033, opened 2022-07-27
ci(next): pass when next deployment is not required

benelan:
**Related Issue:** #

## Summary
The next/storybook deployment is working! This just cleans it up so that the build on master doesn't fail if there are no deployable commits. See error:
https://github.com/Esri/calcite-components/runs/7529705317?check_suite_focus=true#step:5:25


jcfranco (approved):
🎉

Issue #4290, opened 2022-03-23
Calcite range slider not working properly on touch laptop


daniellaumer:
### Actual Behavior

When trying to drag a calcite slider on a laptop with touch screen, the dragging of the slider does not work. The slider thumb keeps stopping and you have to re-drag it. See attached screenshot: 
![CalciteSliderDrag](https://user-images.githubusercontent.com/93542103/159752574-ea3fbdef-eb82-4660-b158-4fb5a0c2ee0a.gif)

Note: A common html range slider works fine. 

### Expected Behavior

The expected behaviour is that the slider thumb follows the touch gesture and does not stop. 

### Reproduction Sample

Just any calcite slider

### Reproduction Steps

Open any example of a calcite slider on a laptop with touchscreen, for example this one: https://developers.arcgis.com/calcite-design-system/components/slider/

Not sure though if it happens for all touch laptops, the problem occured on a Dell Precision 5560

### Reproduction Version

beta.2

### Relevant Info

Windows 10, Chrome and Edge, Hardware: Dell Precision 5560

### Regression?

Probably never worked

jcfranco:
@daniellaumer Thanks for reporting this. 

Additional testing on Windows 10 (thanks @bsvensson!) revealed issues with the following:

* Chrome version 99.0.4844.82 (Official Build) (64-bit)
* Chrome Version 100.0.4896.46 (Official Build) beta (64-bit)
* Chrome Version 100.0.4896.56 (Official Build) beta (64-bit) 
* Edge 99
* Firefox 99 (probably the worse, similar to Edge 99)
* Firefox Beta 100 (behaves like Chrome)

benelan:
@bsvensson @daniellaumer can I have some help verifying that this issue is resolved? I don't have a touch laptop. Thanks!

bsvensson:
I can help test @benelan - do you have a codepen or similar to test before/after?

benelan:
Before:
https://codepen.io/benesri/pen/wvmmLJW

After:
https://codepen.io/benesri/pen/NWYYZpb

daniellaumer:
I tested both links on my touch laptop and the bug still exists on both. Not sure if the bug was not fixed correctly or something goes wrong with loading the new version of calcite in the second app?

bsvensson:
No improvement on my Dell Latitude 7320 touch laptop -- running Windows 10 Enterprise (10.0.1042 Build 19042)
[...]

PR #5037, opened 2022-07-27
fix(slider): Focus active handle on pointerdown instead of click

driskull:
**Related Issue:** #4290

## Summary

fix(slider): Focus active handle on pointerdown instead of click. (#4290)


jcfranco (approved):
✨🎚✨

Issue #5036, opened 2022-07-27
modal : no longer close modal on `esc` key when open on page load


anveshmekala:
### Actual Behavior

When the `active` prop is set on the modal on page load ,  `esc` key isn't closing the modal.

### Expected Behavior

`esc` key press should close the modal

### Reproduction Sample

https://codepen.io/ANVESHMEKALA/pen/vYReQgJ

### Reproduction Steps

1. Open the code pen 
2. Try to press `esc` key and observe modal wont close.

### Reproduction Version

 1.0.0-next.542

### Regression?

beta.89

geospatialem:
Verified on next

PR #5038, opened 2022-07-27
fix(modal): escape key closes modal when open on page load

anveshmekala:
**Related Issue:** #5036 

## Summary

This fix will close the modal on `esc` when  `active` or `open` is set to `true` on page load.


src/components/modal/modal.tsx
```diff
@@ -129,7 +129,7 @@ export class Modal implements ConditionalSlotComponent, OpenCloseComponent {
   //--------------------------------------------------------------------------
   componentWillLoad(): void {
     // when modal initially renders, if active was set we need to open as watcher doesn't fire
-    if (this.open) {
+    if (this.open || this.active) {
```
driskull, line 132: I think this file needs some more work so that there is only one source of truth. Inlcuding...

```
if (this.open || this.active) {
```

Another example:

```ts
@Watch("active")
  @Watch("open")
```

```
 private openModal() {
    this.previousActiveElement = document.activeElement as HTMLElement;
    this.el.addEventListener("calciteModalOpen", this.openEnd);
    this.open = true;
    this.active = true;
```

```
close = (): Promise<void> => {
    return this.beforeClose(this.el).then(() => {
      this.open = false;
      this.active = false;
      focusElement(this.previousActiveElement);
      this.removeOverflowHiddenClass();
    });
  };
```

driskull (changes requested):
Hey @anveshmekala, I think we need to do the "one source of truth" work here too. The active/open props shouldn't be both doing their thing.

src/components/modal/modal.tsx
anveshmekala: The current code base is already working as one source of truth. Watchers for both `active` & `open` trigger either `openModal( )` or `closeModal( )` which sync up both active and open values.

anveshmekala: `if (this.open || this.active) {
  this.openModal( )
  }` 
  inside  componentWillLoad( ) will invoke `openModal( )` if either of open or active is true and openModal method sets both active and open to `true`.

driskull: Ok. I was just thinking it might be better to just have `active` as a property that sets `open` in a watcher. Then `active` doesn't need to be used in the methods or render functions and will be easier to remove when we release 1.0 and can remove deprecated properties.

anveshmekala: I was thinking of avoiding the additional `watchers` code which is expensive. Agree with the fact it is easier to clean up in the future.

anveshmekala:
wondering if this[ line](https://github.com/Esri/calcite-components/blob/master/src/components/modal/modal.tsx#L399) is required as the handler always receives new values so the comparison isn't required. Or am I missing something. Thoughts?

src/components/modal/modal.tsx
```diff
@@ -394,9 +400,14 @@ export class Modal implements ConditionalSlotComponent, OpenCloseComponent {
   };
 
   @Watch("active")
+  activeHandler(value: boolean): void {
+    this.open = value;
+  }
+
   @Watch("open")
   async toggleModal(value: boolean, oldValue: boolean): Promise<void> {
     if (value !== oldValue) {
```
driskull, line 409: Yeah i'm not sure if this check is needed. It shouldn't be if the watcher is the only thing calling it.

driskull (approved):
Looks good!

capegreg:
Why would you not want the modal to close on escape? 
"Additionally, for those using a device with a keyboard, the Escape key is commonly expected to close modal dialogs as well."
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#:~:text=Additionally%2C%20for%20those%20using%20a,close%20modal%20dialogs%20as%20well.

macandcheese:
@capegreg - by default, that is the behavior, to allow `esc` to close the Modal. 

We have a set of configuration options - `outside-close-disabled`, `escape-disabled`, `close-button-disabled`, that can be used alone or in tandem to customize the experience, when needed. Use cases for this aren't that common, but are valid, and can include instances where a modal with destructive action is presented, or an application needs a user to make a choice to continue, for instance, where accidentally using `esc` may prematurely close a workflow.

Issue #5043, opened 2022-07-28
popover,notice, chip,panel : setting `closable` prop to false does not remove `dismissible`.


anveshmekala:
### Actual Behavior

When the `dismissible` prop is set to `true ` and in the script, user sets the `closable` prop to `false`, the `dismissible` prop still stays as an attribute, and the `clear (x)`  icon is still visible in the component.

Note: This behavior is observed when the user uses one prop as an attribute (either deprecated or new one) and uses another to mutate the value. 

### Expected Behavior

Setting the `closable` prop to false should remove the `dismissible `prop in DOM with value as false and vice-versa.

### Reproduction Sample

https://codepen.io/ANVESHMEKALA/pen/RwMjGrB?editors=1010

### Reproduction Steps

1. Open the code pen and observe every component has a `dismissible` set to true
2. In the script we set `closable` to `false`.
3. Observe the clear icon still stays for all the components

### Reproduction Version

beta.86

### Relevant Info

Related to [4789](https://github.com/Esri/calcite-components/pull/4789)

benelan:
@anveshmekala when I change the sample to `next` the close button is still there for all of the components. Am I doing something wrong?

```html
<script
  type="module"
  src="https://unpkg.com/@esri/calcite-components@next/dist/calcite/calcite.esm.js"
></script>
<link
  rel="stylesheet"
  type="text/css"
  href="https://unpkg.com/@esri/calcite-components@next/dist/calcite/calcite.css"
/>
```

benelan:
Assigning this one back to dev for another look as well:
https://codepen.io/benesri/pen/bGvvXYM?editors=1010

anveshmekala:
From the initial debugging, what I observed is `<script/>` is executed even before `connectedCallback`  which means `watchers` are not invoked when `dismissible =false` or `closable=false` in the script.

anveshmekala:
This is a similar [issue](https://github.com/Esri/calcite-components/issues/4996#issuecomment-1210005463) where setting attributes changes props but not the other way around prior to hydration.

anveshmekala:
can we close this one @benelan ?

geospatialem:
Closing, as this has similar behavior as mentioned in https://github.com/Esri/calcite-components/issues/4996#issuecomment-1210005463.

PR #5044, opened 2022-07-28
fix(notice,chip,popover): setting closable prop to false will remove dismissible prop

anveshmekala:
**Related Issue:** #5043 

## Summary

This fix will make sure the `dismissible` prop will be removed if `closable=false` and vice-versa.


src/components/popover/popover.tsx
```diff
@@ -77,9 +77,18 @@ export class Popover implements FloatingUIComponent, OpenCloseComponent {
    */
   @Prop({ reflect: true }) dismissible = false;
 
+  handleDismissible(value: boolean): void {
```
driskull, line 80: does this need a watcher?

driskull (changes requested):
I think there's a watch missing. otherwise looks good 👍

src/components/popover/popover.tsx
anveshmekala: This needs the watcher to update `closable` value .

driskull: Is it missing?

anveshmekala: Ahh i see it now, missed the decorator.

anveshmekala: Will add an e2e test to make sure both props are in sync.

driskull (approved):
👏 🌈

Issue #3099, opened 2021-09-23
Bug: combobox inside of modal not overflowing correctly


benelan:
### Actual Behavior
The `calcite-combobox` list container opens inside of the `calcite-modal` content div rather than overflowing.

### Expected Behavior
The `calcite-combobox` list container overflows correctly like it does with `calcite-select`.

### Reproduction Steps and Sample

_Sample_:
`calcite-select`: https://codepen.io/benesri/pen/NWgLgJJ
`calcite-combobox`: https://codepen.io/benesri/pen/eYRLWej

### Relevant Info

https://community.esri.com/t5/calcite-design-system-questions/combobox-vs-select-in-a-modal/m-p/1084169#M29

_Version_: `@esri/calcite-components@1.0.0-beta.65`

- [x] CDN
- [ ] NPM package

benelan:
Changing the modal class to `overflow-visible` has worked for me in the past with bootstrap, but since it currently works for `calcite-select` there may be a different approach?

Elijbet:
> @jcfranco this is an issue with modal, not combobox.
> 
> The native select is apples to oranges as it renders using browser UI.
> 
> If the `.modal` class on the modal removes `overflow:hidden` and the `.content` class in the modal is set to `overflow:visible` then this works fine.
> 
> We probably need an option to set that up?

I tried to explore that option, but I don't see how removing overflow:hidden would work with modals that have both large content and a combobox.  Here is a sample: https://codepen.io/elijbet/pen/JjroqZq

driskull:
I think this is a bigger issue that applies to any component using popper.js. If placed inside a model, it's contained within that modal by default.

If we want it outside of that modal, the popper node would need to be appended to the document.body instead. This would require some discussion and refactoring.

I think we should setup a discussion and push the issue to a different milestone.

jcfranco:
This is blocked by https://github.com/Esri/calcite-components/issues/3815.

jcfranco:
I think I've found a way to move forward with this w/o having to wait for improvements to floating-ui. Will have a test PR up soon.

jcfranco:
Installed 2 PRs that should help fix the main issue as long as the combobox is positioned below (see https://codepen.io/jcfranco/pen/zYWPQey). Otherwise, there is an issue with z-index layering. I've submitted #5058 to fix this and other z-index issues.

jcfranco:
#5058 landed.

benelan:
verified on next

PR #5045, opened 2022-07-28
fix(combobox): set min width of combobox items to match its input

jcfranco:
**Related Issue:** #3099 

## Summary

This updates the combobox to avoid a situation where the option list will stretch to the combobox's parent width (see https://codepen.io/jcfranco/pen/BarmpJV?editors=1000).

This is part 1 of 2 fixes needed for #3099. 

@driskull I tried following the same approach as the dropdown. LMK if I need to make any adjustments. 🙇‍♂️

### DIFF
--- a/src/components/combobox/combobox.tsx
+++ b/src/components/combobox/combobox.tsx
@@ -590,7 +590,7 @@ export class Combobox
   };
 
   setMaxScrollerHeight = async (): Promise<void> => {
-    const { listContainerEl, open } = this;
+    const { listContainerEl, open, referenceEl } = this;
 
     if (!listContainerEl || !open) {
       return;
@@ -599,6 +599,7 @@ export class Combobox
     await this.reposition();
     const maxScrollerHeight = this.getMaxScrollerHeight();
     listContainerEl.style.maxHeight = maxScrollerHeight > 0 ? `${maxScrollerHeight}px` : "";
+    listContainerEl.style.minWidth = `${referenceEl.clientWidth}px`;
     await this.reposition();
   };
 
--- a/src/components/combobox/combobox.scss
+++ b/src/components/combobox/combobox.scss
@@ -142,7 +142,6 @@
 }
 
 .floating-ui-container {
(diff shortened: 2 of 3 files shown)


src/components/combobox/combobox.tsx
```diff
@@ -599,6 +599,7 @@ export class Combobox
     await this.reposition();
     const maxScrollerHeight = this.getMaxScrollerHeight();
     listContainerEl.style.maxHeight = maxScrollerHeight > 0 ? `${maxScrollerHeight}px` : "";
+    listContainerEl.style.minWidth = `${referenceEl.clientWidth}px`;
```
driskull, line 602: do you need to check to see if `referenceEl` exists here?

driskull (approved):
👍

src/components/combobox/combobox.tsx
jcfranco: `referenceEl` exists if `listContainerEl` does, right?

driskull: Not exactly. referenceEl will depend on the user setting the referenceElement.

jcfranco: I think I'm missing something. `referenceEl` is internal, [isn't](https://github.com/Esri/calcite-components/blob/master/src/components/combobox/combobox.tsx#L424) [it](https://github.com/Esri/calcite-components/blob/master/src/components/combobox/combobox.tsx#L1175)?

driskull: Yep. my bad! I thought this was popover :) 🧠 💨

jgrayson-apl:
...just curious and wanting to understand...

Q: with this fix, is it expected behavior for a combobox-item with a long string to extend beyond the width of the combobox?

https://codepen.io/john-grayson/pen/xxWPQJV

jcfranco:
Yes, similar to dropdown (see https://github.com/Esri/calcite-components/pull/4635).

PR #5046, opened 2022-07-28
fix(modal): improve interaction between modal and floating-ui-owning components

jcfranco:
**Related Issue:** #3099 

## Summary

This helps mitigate an issue where `floating-ui` cannot position properly if a parent container has a transform applied (see https://github.com/Esri/calcite-components/issues/3815).

It works by applying its transition transform during the opening/closing sequences. Once open or closed, the transform is removed and will allow floating-ui children to be positioned correctly (assuming there are no parent transforms applied further up the tree).

This is part 2 of 2 fixes needed for #3099.

### Notes

* There is a slight difference in timing in the opening transition (shorter now). I would appreciate some feedback and/or help on tweaking it to match the current transition. cc @macandcheese @paulcpederson 
* This fix will only work as long as modal doesn't use `floating-ui`. Once this happens, we'll need the bug fixed upstream.
* Inspired by https://github.com/AFASSoftware/maquette-css-transitions/


macandcheese (approved):
Looks smooth to me 👍

Issue #4814, opened 2022-06-28
calcite-tree broken navigation when inside a shadow DOM


ffaubry:
### Actual Behavior

If the calcite-tree is inside an another component with shadow root and the tree has sub-trees, the navigation with the right arrow is broken.

If the calcite-tree is not wrapped in a div, we get runtime errors.

### Expected Behavior

Being able to navigate the whole tree with arrow up/down/left/right independently of being in a shadow dom.

### Reproduction Sample

https://github.com/ffaubry/calcite-tree-navigation-issue

### Reproduction Steps

See README.md in the repo.

### Reproduction Version

beta.83

benelan:
Thanks @ffaubry you can ignore the bot, we will take a look.

benelan:
Bumping the RepoRepro™️ sample to `next` fixes the sub tree issue. However removing the parent `div` still causes a runtime error
![image](https://user-images.githubusercontent.com/10986395/182977129-c0de320a-f28f-4ff0-b8b0-762a56145d4d.png)

@jcfranco is this expected behavior, in which case I'll document it, or is more tweaking needed?

jcfranco:
Per our convo, this above error is a separate issue.

benelan:
Verified based on ☝️ 
I created a new issue for the error due to tree assuming there will be a parent node #5333

PR #5057, opened 2022-07-30
fix(tree): fix keyboard navigation when wrapped in shadow DOM

jcfranco:
**Related Issue:** #4814 

## Summary

Tree was using the document's `activeElement` instead of its root node for keyboard navigation. 

Automatically testing our components when wrapped in shadow DOM can get complex. I'll look into an approach with demo pages and possibly storybook to see if we can spot these earlier. It'd be a manual check, but at least something easy to test.


driskull (approved):
🎉 💥

Issue #4781, opened 2022-06-21
Tooltip text is being cut off by action panel header 


Apahadi73:
### Actual Behavior

Tooltip UI is being cut off when the parent element of tooltip is placed right below the action-panel header element.
This issue was first observed in the issue: [WebGIS/arcgis-charts#4756](https://devtopia.esri.com/WebGIS/arcgis-charts/issues/4756)

![tooltip-issue](https://user-images.githubusercontent.com/36856709/174854894-0f45d027-def4-4cd7-8038-89ae5aa287a4.gif)

### Expected Behavior

When the user hovers over eye and trash can icon, the tooltip should appear above the action-panel header.  I have attached codepen link where you can observe both the actual and expected behaviors.

### Reproduction Sample

https://codepen.io/apahadi73/pen/RwQYNEq

### Reproduction Steps

1. Open the link: https://analysis-0.mapsdevext.arcgis.com/
2. Sign in using your test credentials.
3. Open Map Viewer.
4. Add a feature layer to the map.
5. Create a chart using chart config action displayed on the rightmost action panel
6. Select any one of the charts option available.
7. Click on the guides option/menu.
8. Hover over eye or trash icon.
9. Observe that the icon's tooltip is being cut-off by action-panel header.

### Reproduction Version

v1.0.0-next.489

Issue #4682, opened 2022-06-03
calcite-panel: menus that overflow the panel header are cut off


driskull:
### Actual Behavior

menus that overflow the panel header are cut off

### Expected Behavior

Menus are displayed below or above the header of the panel component without being cut off.

### Reproduction Sample

https://codepen.io/COV-GIS/pen/VwyGmZb

### Reproduction Steps

1. Hover over a header action and the tooltip is behind the panel's content.

### Reproduction Version

1.0.0-beta.82

### Relevant Info

https://github.com/Esri/calcite-components/pull/4182

benelan:
verified on beta.83

Issue #4372, opened 2022-04-05
Introduce tokens for z-index management (overlay layers)


jcfranco:
### Description

Stems from https://github.com/Esri/calcite-components/pull/4182#discussion_r840900777

It would be beneficial to introduce a set of tokens to represent different levels of z-indices used in the design system. This could help our components use the same index depending on the context and better inform users on how to override these values. 

### Acceptance Criteria

1. A new set of tokens for z-indices is added (Tailwind for now)
2. These z-index tokens should provide context for their intended use
3. Components that require z-index should use these for consistency instead of hardcoded values

### Relevant Info

https://www.lightningdesignsystem.com/design-tokens/#category-z-index
https://designsystem.morningstar.com/legacy/v/2.4.0/visual-language/layering.html
https://www.duetds.com/tokens/#z-index
https://github.com/dojo/widgets#z-index-layering
https://www.smashingmagazine.com/2021/02/css-z-index-large-projects/

### Which Component

Any component that makes use of overlays.

jcfranco:
All components should now use the [contextual tailwind z-index utils](https://github.com/Esri/calcite-components/blob/master/tailwind.config.ts#L213-L221) when layering is needed. If developers need to tweak layering, they can reference our z-index layers and set z-indices in their apps accordingly.

geospatialem:
Verified on master ✨

PR #5058, opened 2022-07-30
fix: apply layered z-indices across components

jcfranco:
**Related Issue:** #4372 #3099 #4682 #3724 #4781 

## Summary

This fixes a couple of layering issues we had in our components by doing the following:

1. updated tailwind config to introduce contextual z-index layers
2. updated component styles to reference these
3. removed unnecessary z-indices (see 3 below)

### Notes

1. I can break any of these items to separate PRs if deemed appropriate
2. Layers are based on https://www.duetds.com/tokens/#z-index and can be tweaked accordingly
3. Any removed `z-index` can be restored if we spot a regression. Removal was determined by:
    * evaluating the need for a stacking context
    * local testing
4. Repro cases from the referenced issues were tested and work as expected
5. Updated all z-indices to be applied as tailwind utils
6. Skipped adding CSS props for now until the z-index tokens are decided

### DIFF
--- a/src/components/action/action.scss
+++ b/src/components/action/action.scss
@@ -199,7 +199,6 @@
   .button::after {
     content: "";
     @apply absolute
-      z-10
       h-2
       w-2
       rounded-full
--- a/src/components/panel/panel.scss
+++ b/src/components/panel/panel.scss
@@ -67,12 +67,6 @@
   @apply hidden;
 }
 
-:host([loading]) .container,
-:host([disabled]) .container {
-  @apply relative;
-  z-index: 1;
-}
-
 .header {
   border-block-end: 1px solid;
   @apply bg-foreground-1
@@ -81,9 +75,9 @@
     top-0
     w-full
     items-stretch
-    justify-start;
+    justify-start
+    z-header;
   flex: 0 0 auto;
-  z-index: 2;
 }
 
 .header-content {
@@ -155,7 +149,6 @@
   flex-col
   flex-nowrap
   items-stretch;
-  z-index: 1;
 }
 
 .footer {
@@ -181,7 +174,7 @@
   my-0
   mx-auto
   block
-  p-2;
+  p-2
+  z-sticky;
   inline-size: fit-content;
-  z-index: 1;
 }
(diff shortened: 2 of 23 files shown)


macandcheese:
Awesome! @SkyeSeitz @ashetland these layers will be useful to replicate in Figma / tokens / documentation

tailwind.config.ts
```diff
@@ -208,6 +208,17 @@ export default {
       },
       maxHeight: {
         menu: "45vh"
+      },
+      zIndex: {
```
driskull, line 212: this is nice :)

src/components/shell-panel/shell-panel.scss
```diff
@@ -35,7 +35,6 @@
   absolute
   bottom-0
   top-0
-  z-10
```
driskull, line 38: Do we know why some of these are here? Just don't want to cause any regressions. Maybe more testing with the ones we remove?

src/components/panel/panel.scss
```diff
@@ -67,12 +67,6 @@
   @apply hidden;
 }
 
-:host([loading]) .container,
```
driskull, line 70: Did you test this to make sure that the scrim is showing correctly and with scrolling elements below?

src/components/action/action.scss
```diff
@@ -199,7 +199,6 @@
   .button::after {
     content: "";
     @apply absolute
-      z-10
```
driskull, line 202: Not sure why this was here 🤷

driskull (approved):
Great idea 💡

src/components/panel/panel.scss
jcfranco: Yes, but will test again to make sure.

src/components/shell-panel/shell-panel.scss
jcfranco: I couldn't deduce the need for it, and didn't notice any visual issues after removing. 

> Maybe more testing with the ones we remove?

Great idea.

src/components/panel/panel.scss
jcfranco: Local non-exhaustive testing looks good. 🎉

PR #5060, opened 2022-08-01
test: add demo page component to move content between light/shadow DOM

jcfranco:
**Related Issue:** N/A

## Summary

This adds a demo component that can help test our components within shadow DOM. The idea is to have it wrap all demo content, so that it can be easily moved to shadow DOM and tested. 

From https://github.com/Esri/calcite-components/pull/5057:

> Automatically testing our components when wrapped in shadow DOM can get complex. I'll look into an approach with demo pages and possibly storybook to see if we can spot these earlier. It'd be a manual check, but at least something easy to test.

### Notes

* ~Only `accordion` demo page is wired up properly~ All demo pages updated!
* ~Page styles need to be a child of the component so demo page parts are styled consistently.~ Demo page style is automatically injected by the component.


src/demos/_assets/demo-template.html
```diff
@@ -18,6 +18,9 @@ <h1 id="demo-heading" class="demo-heading"></h1>
         <calcite-label class="demo-switches__label" layout="inline" scale="s">
           <calcite-switch scale="s" id="toggle-theme"></calcite-switch> Dark Theme
         </calcite-label>
+        <calcite-label class="demo-switches__label" layout="inline" scale="s">
+          <calcite-switch scale="s" id="toggle-dom"></calcite-switch> Move to shadow DOM
```
driskull, line 22: Maybe simplify the button to `shadowRoot`,  `Shadowed`, or just `Shadow`?

driskull (approved):
💣

eriklharper (commented):
Looks good to me, great idea!

jcfranco:
Found a way to copy and inject all styles into the component, so `<style>` tags can stay where they are. 🎉

PR #5061, opened 2022-08-01
fix(tab): Screen reader support to access tab content

geospatialem:
**Related Issue:** #2124 

## Summary
Adds screen reader support to access the `calcite-tab` content by moving the `role` and `tabindex` to a child element.

Prior to the fix when accessing the tab content, screen reader users were provided with the `tab-title`, but not the `tab` content in particular with JAWS and VoiceOver.

Additional reference: [4979 comment](https://github.com/Esri/calcite-components/pull/4979#discussion_r926069374)

Verified on:
- [x] JAWS and Chrome
- [x] JAWS and Firefox
- [x] NVDA and Chrome
- [x] NVDA and Firefox
- [x] VoiceOver and Safari

### DIFF
--- a/src/components/tab/tab.tsx
+++ b/src/components/tab/tab.tsx
@@ -65,15 +65,12 @@ export class Tab {
     const id = this.el.id || this.guid;
 
     return (
-      <Host
-        aria-labelledby={this.labeledBy}
-        id={id}
-        role="tabpanel"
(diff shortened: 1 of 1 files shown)


driskull (approved):
This looks good. 👍  

Should be fine as logn as the labeledby and id are still on the host.

Issue #4961, opened 2022-07-18
Add built-in translations to components


jcfranco:
### Overview

Currently, our components have placed the responsibility of text translations on the user. Unfortunately, this pattern has lead to some unnecessary boilerplate from teams/projects that have to support more than one locale and do not provide customized strings (e.g., each project needs to set up translation for the same strings and update each component).
The goal for this epic is to make it easier for developers to bring in components w/o having to worry about setting up translations themselves.

### Tasks

- [x] Define pattern for localized components (see notes below) (https://github.com/Esri/calcite-components/pull/5065)
  - [x] Deprecate `intl<Prop>`s as these should no longer be needed #5946 
  - [x] Determine if we need support for overrides at 1.0.0
- [x] Add script to generate types for t9n bundles for type safety (https://github.com/Esri/calcite-components/pull/5065)
- [x] Plan automation of translation-bundles
  <details>

    * make sure to follow supported locales
    * use t9n naming
    * need to consider that designs may need to be updated to accommodate translations (aim to get translations early for this)
    * keep CLDR data-driven strings (e.g., dates, raw number data) outside of bundles
    * special considerations may be needed for RTL
    * error strings will need to be localized
    **action items**
    * Franco + Anvesh will work on a PR with the translation bundle scaffolding in place
    * Once ☝️ is ready, the next steps is to review and test our setup
</details>

### Notes

- No implementation details should be exposed, this will give us total flexibility on translation/loading implementation, we can revisit if we need to adopt a different lib for consistency across components – first iteration can be a simple JSON file fetcher
- String bundles should:
  - be placed under `src/components/<component>/assets/t9n/`
  - have a main bundle (manually updated), and the rest will be automatically generated and named per their locale
    - the structure for the above could be anything as long as it does not unnecessarily add bytes, is easy to maintain and is clear to distinguish the main file from the rest
  - be JSON files – need to determine structure (a flat object of strings may be the best way to support string overrides, if needed
- To support backwards-compatibility:
  - `intl<Prop>` values should no longer be set by default
  - after loading the built-in bundle, any provided `intl<Prop>`s should be merged w/ the built-in strings
  - technically, this would break anyone referencing the strings, but we expect this to not be a common use case 
- requests/bundles must be cached to handle multiple components of the same type loading later or in parallel
- All core t9n behavior should be placed in a module and reused across components

### Questions
              
- Do we need to support runtime locale changes before 1.0.0?
[...]

PR #5065, opened 2022-08-02
feat: add built-in translations

jcfranco:
**Related Issue:** #4961

## Summary

Adds built-in translations for all components:

* added string bundles for all components with default text
* add script to generate bundle types
* wires up color-picker with simple bundle loading pattern

### Pending

- [x] add locale fetching util (potentially introduce `LocalizedComponent` interface for consistency)
  - [x] should support both `nb` and `no` for Norwegian locale
  - [x] cache requests so they are shared across component instances
- [ ] find optimal way to add a common bundle (later)
- [ ] wire up all components (separate PRs)
- [x] update conventions doc with t9n pattern
- [x] add test helper
- [x] add test helper for mocking fetch requests

cc @Esri/calcite-translation


src/utils/localization/fetchLocale.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Get supported locale code from raw user input
+ * Exported for testing purposes.
+ *
+ * @param lang
+ * @private
+ * @returns {string}
+ */
+export function getSupportedLocale(lang = ""): string {
```
jcfranco, line 11: Do we need a default here? I'm not sure in which scenario we wouldn't have a locale defined.

jcfranco, line 11: Please extract this in a way it can be shared between the date-picker utils and here.

src/components/color-picker/color-picker.tsx
```diff
@@ -1087,21 +1090,25 @@ export class ColorPicker implements InteractiveComponent {
...
-      await fetch(getAssetPath(`./assets/color-picker/i18n/${locale}.json`))
-    ).json();
+    this.builtInStrings = await fetchLocaleStrings(locale, "color-picker");
   }
 
   getLocale(): string {
-    const lang = document.documentElement.lang || navigator.language; // need to determine best place to grab locale from
+    // need to determine best place to grab locale from
+    //we can go by order ( locale of component > locale of closests > lang of document > navigator)
+    const lang = document.documentElement.lang || navigator.language;
```
jcfranco, line 1104: Let's include the element's `lang` with higher precedence:

`el.lang || document.documentElement.lang || navigator.language;`

src/components/color-picker/color-picker.tsx
```diff
@@ -1087,21 +1090,25 @@ export class ColorPicker implements InteractiveComponent {
...
+
     return this.normalizeLocale(lang);
   }
 
-  normalizeLocale(locale: string): string {
+  normalizeLocale(lang: string): string {
```
jcfranco, line 1109: This and `getLocale` were just placeholder methods. They can be dropped in favor of the string fetching util.

src/components/color-picker/color-picker.tsx
```diff
@@ -388,7 +390,7 @@ export class ColorPicker implements InteractiveComponent {
 
   @State() dimensions = DIMENSIONS.m;
 
-  @State() mergedStrings: Strings;
+  @State() mergedStrings: Partial<Strings>;
```
jcfranco, line 393: Only the user-provided strings should be partial since they're optional.

src/utils/localization/fetchLocaleData.ts
```diff
@@ -0,0 +1,11 @@
+import { getAssetPath } from "@stencil/core";
+
+export async function fetchLocaleStrings(locale: string, component: string): Promise<any> {
+  const res = fetch(getAssetPath(`./assets/${component}/t9n/${locale}.json`))
```
jcfranco, line 4: Let's use `try`/`await` here.

src/utils/localization/fetchLocaleData.ts
```diff
@@ -0,0 +1,11 @@
...
+    .then((resp) => resp.json())
+    .catch(() => {
+      console.error(`Translations for "${locale}" not found or invalid, falling back to english`);
```
jcfranco, line 7: I don't think we need to warn here since we'll always normalize to a supported locale.

src/utils/localization/fetchLocale.ts
```diff
@@ -0,0 +1,34 @@
+import { locales } from "../locale";
+
+/**
+ * Get supported locale code from raw user input
+ * Exported for testing purposes.
```
jcfranco, line 5: Let's drop this. I think it's expressive enough and it's not being tested yet.

src/components/color-picker/color-picker.tsx
```diff
@@ -791,6 +793,9 @@ export class ColorPicker implements InteractiveComponent {
       this.savedColors = JSON.parse(localStorage.getItem(storageKey));
     }
 
+    //fetch Bundle
```
jcfranco, line 796: This comment isn't needed as the following line is expressive enough.

src/components/color-picker/color-picker.tsx
```diff
@@ -805,7 +810,6 @@ export class ColorPicker implements InteractiveComponent {
         });
     }
 
-    await this.fetchStrings();
```
jcfranco: Let's avoid formatting unrelated sections of code for now.

src/utils/localization/fetchLocale.ts
anveshmekala: nope, we dont.

src/components/color-picker/color-picker.tsx
anveshmekala: It should be `el.locale` which gets the preference.

anveshmekala: atleast for the components with `locale` prop. We may need a util method that determines locale for a given component considering the element's locale, lang, closest elements locale & document.documentElement.lang || navigator.language. Thoughts?

jcfranco: We're going to be dropping `locale` in favor of the global `lang` attribute/prop. 

I agree that we need to think about looking up `lang` if set on a parent vs the component itself, but that's outside the scope of this PR. I'll create an issue for this.

jcfranco: Lookup and resolution of locale is handled by https://github.com/Esri/calcite-components/pull/5338

jcfranco:
Closing in favor of splintered PRs.

Issue #4866, opened 2022-07-07
Provide context to `dropdown` item selections when closing the menu for screen reader users


geospatialem:
### Summary

When a dropdown item is selected screen reader users aren't currently provided with feedback that a selection, or un-selection has been made. In order for them to get this context they must re-open the component in its entirety, and navigate through each item to hear the selection(s).

### Actual Behavior

Screen reader users must navigate the `dropdown` menu completely upon selection/un-selection to identify if a selection, or multiple selections exist in the menu.

### Expected Behavior

Provide context to users if they make item selections, or un-selections.

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/dropdown/#sample

### Reproduction Steps

1. Enable a screen reader (NVDA, JAWS or VoiceOver)
2. Navigate to the [sample](https://developers.arcgis.com/calcite-design-system/components/dropdown/#sample)
3. With the screen reader active, navigate via keyboard to focus on the "Activate dropdown" button to trigger the dropdown
4. Select a dropdown item in the list.
5. Observe no selection context is provided to the screen reader user, and they must navigate the dropdown in its entirety to receive feedback on the selections.

### Reproduction Version

beta.83

### Relevant Info

- Windows (JAWS and NVDA)
- Mac OS (Safari/VoiceOver)
- iOS (Safari/VoiceOver)

driskull:
I tried a few things with voiceover but no luck. I think this one might be related to the shadowRoot/slotting of web components. We might need to experiment or research a bit more.

geospatialem:
> I tried a few things with voiceover but no luck. I think this one might be related to the shadowRoot/slotting of web components. We might need to experiment or research a bit more.

Will dig into some research with this one this week. Thanks for taking an initial look at this, @driskull!

driskull:
@geospatialem how do we provide that context?

geospatialem:
> @geospatialem how do we provide that context?

Not super confident in a good implementation. It seems like prior to closing, perhaps in the [`calciteInternalDropdownItemSelect`](https://github.com/Esri/calcite-components/blob/master/src/components/dropdown/dropdown.tsx#L358-L368), we'd need to ensure the selection is heard back prior to running lines 365-371.

Was able to get it to successfully work after delaying via a `setTimeout` of 0.05 seconds (50 ms), but that changes the expected experience to users. 

@driskull Is there a better method, perhaps a `Promise` prior to running `closeCalciteDropdown`?

[...]

PR #5080, opened 2022-08-03
fix(dropdown): Provide context to dropdown item selections when closing the menu for screen reader users

driskull:
**Related Issue:** #4866

## Summary

fix(dropdown): Provide context to dropdown item selections when closing the menu for screen reader users (#4866)


driskull:
@geospatialem could you test this out?

geospatialem:
> could you test this out?

@driskull It seems to be functioning the same for VO, NVDA, and JAWS with no feedback when the menu closes after selection. ☹️

driskull:
@geospatialem can you test again?

geospatialem:
Its much easier to navigate - now its announcing when its collapsed (which is a fantastic improvement). 

Will take a peek tomorrow to see if the great enhancements you added make it easier to announce the item when its selected and the menu is collapsed.

driskull:
Ok if we need to do this then we need to rethink how the slotted dropdown trigger works. 

If a `button` is slotted into this component and the "menu" is also a button then aXe will complain about nested button roles.

Ideally, anything a user would slot as a trigger would not be a button so that there are no nested button roles. cc @jcfranco @benelan. Whatever the user clicks should only toggle the menu and not do anything else anyway.

driskull:
```
Summary of all failing tests
FAIL src/components/dropdown/dropdown.e2e.ts (86.621 s)
  ● calcite-dropdown › is accessible

    expect(received).toHaveNoViolations(expected)

    Expected the HTML found at $('calcite-dropdown,#calcite-dropdown-e33cd73c-bbaf-faab-d0af-93bfc9decc28-menubutton') to have no violations:

    <div aria-controls="calcite-dropdown-e33cd73c-bbaf-faab-d0af-93bfc9decc28-menu" aria-expanded="false" aria-haspopup="menu" class="calcite-dropdown-trigger-container" id="calcite-dropdown-e33cd73c-bbaf-faab-d0af-93bfc9decc28-menubutton" role="button"><slot name="dropdown-trigger"></slot></div>

    Received:

    "Interactive controls must not be nested (nested-interactive)"

    Fix any of the following:
      Element has focusable descendants

    You can find more information on this issue here: 
    https://dequeuniversity.com/rules/axe/4.4/nested-interactive?application=axeAPI

      65 |       getTag(componentTagOrHTML)
      66 |     )
    > 67 |   ).toHaveNoViolations();
         |     ^
      68 | }
      69 |
      70 | /**

      at accessible (src/tests/commonTests.ts:67:5)
          at runMicrotasks (<anonymous>)
      at Object.<anonymous> (src/components/dropdown/dropdown.e2e.ts:934:5)
```

geospatialem:
This looks to be the expected behavior of the `role="menu"`, since its collapsed, screen readers provide context the menu is now collapsed, rather than an item was selected since its the primary focus. In adding a `setTimeout` function prior to closing the menu, the selection is heard, so my guess is there is an assertive tag associated with the closing of the menu, that ignores that a selection has taken place. For context, [the role only needs to convey elements within the menu](https://a11ysupport.io/tech/aria/menu_role), not necessarily whether they are selected/not selected.

If we modified the roles to `listbox` (and set `aria-multiselectable="true"` for multi selection mode and "false" for the other selection types) and `option` , we'd get the results for the intent back to screen readers, but that's a much more major overhaul.

Some possible options as we move forward:
1. No code changes. Instead provide documentation that to support assistive technologies, we recommend setting the `disableCloseOnSelect` to provide on-demand feedback to assistive technologies on `dropdown-item` select.
2. Change the `disableCloseOnSelect` to be true by default to provide support to assistive technologies. The property could still be set to false, for those seeking to disable it, but the documentation could indicate the lack of support to assistive technologies.
3. Update the roles, possibly to `listbox`, or another alternative role, to provide on-demand feedback to assistive technologies.
4. TBD?

driskull:
I think this issue is blocked until we can resolve the nested button issue.

Issue #4144, opened 2022-02-17
calcite-filter stop escape key propagation to prevent closing parent panels


mpayson:
### Description

Currently if a user presses "escape" within `calcite-filter` that event will propagate which can have unintended consequences like closing a dismissible parent panel. Here's an [example codepen](https://codepen.io/mpayson/pen/NWwXBey) -- type a value into the filter then hit escape, notice the whole panel will close

### Acceptance Criteria

In the above example, pressing the escape key just clears the filter, it does not close the panel

### Relevant Info

NA

### Which Component

`calcite-filter` and consuming components like `calcite-pick-list`, although our use case is just using the calcite filter.

### Example Use Case

See above

geospatialem:
Verified on `next`

https://user-images.githubusercontent.com/5023024/184908850-c59c8461-e630-422c-b44d-5e05cf09a3a5.mov

PR #5083, opened 2022-08-04
fix: ensure handled key events are canceled and ignore canceled-events

jcfranco:
**Related Issue:** #4144 

## Summary

Updates all native keyboard event handling to cancel default behavior if processed by the component and also updates components to ignore canceled events.

### Notes

* Left `date-picker`'s `reset` method as is and doesn't cancel it. It looks like it may be tied to `input-date-picker`'s behavior. 🤔
* Only updated components that accept children or have supporting components that handle keyboard events.
* Added entry to conventions.
* Dropped obsolete `"Spacebar"` checks. https://github.com/Esri/calcite-components/pull/5107
* Used `isActivationKey` where appropriate. https://github.com/Esri/calcite-components/pull/5108
* Restored propagation of some events.
* Simplified by using destructuring of `key`. https://github.com/Esri/calcite-components/pull/5106

**Note:** unrelated refactor changes will be split into separate PRs


driskull (approved):
👍 Nice!!!

jcfranco:
I'd like to install this as soon as beta.91 is released to allow for additional testing on `@next`. cc @benelan

PR #5093, opened 2022-08-04
feat(tab,tab-title): deprecate active and add selected prop

anveshmekala:
**Related Issue:** #4688 

## Summary

This will deprecate `active` prop and adds `selected` prop as a replacement.


driskull (approved):
🎉 Looks good 👍

PR #5097, opened 2022-08-04
feat(stepper-item,dropdown-item): deprecate active and add support for selected prop

anveshmekala:
**Related Issue:** #4688 

## Summary

This PR will deprecate `active` prop in `stepper-item` & `dropdown-item`. Instead of `active` it is suggested to use `selected` prop.


driskull (approved):
LGTM! 👍

PR #5103, opened 2022-08-05
fix(tabs): does not ignore layout="center" attribute

Elijbet:
**Related Issue:** #[4971](https://github.com/Esri/calcite-components/issues/4971)

## Summary

🚧 :construction: Testing solutions


src/components/tabs/tabs.tsx
```diff
@@ -197,4 +208,9 @@ export class Tabs {
     this.tabs.forEach((el) => el.updateAriaInfo(tabIds, titleIds));
     this.titles.forEach((el) => el.updateAriaInfo(tabIds, titleIds));
   }
+
+  private updateParentLayoutChanged = (el: HTMLCalciteTabsElement, layout: TabLayout): void => {
+    const item = getSlotted<HTMLCalciteTabsElement>(el, SLOTS.tabNav);
```
driskull, line 213: ok I think I understand...

When the `layout` property on the `tabs` parent component changes, you want to update the children `tab-nav`s to set their layout. Is that correct?

If so, there are a few ways we could do this but I think this is the most efficient...

```patch
diff --git a/src/components/tabs/tabs.tsx b/src/components/tabs/tabs.tsx
index 34ddd0f4d1ef664b93bca409ed4c22cf2ee83cd0..8cbadcb480155871a7a48a0f0eb2fc40e2d73915 100644
--- a/src/components/tabs/tabs.tsx
+++ b/src/components/tabs/tabs.tsx
@@ -1,4 +1,4 @@
-import { Component, Prop, h, Element, Listen, State, VNode, Fragment } from "@stencil/core";
+import { Component, Prop, h, Element, Listen, State, VNode, Fragment, Watch } from "@stencil/core";
 import { TabLayout, TabPosition } from "./interfaces";
 import { SLOTS } from "./resources";
@@ -32,6 +32,11 @@ export class Tabs {
    */
   @Prop({ reflect: true }) layout: TabLayout = "inline";
 
+  @Watch("layout")
+  layoutHandler(): void {
+    this.updateTabNavLayout();
+  }
+
   /**
    * Display the tabs top (default) or bottom of the tab content. above and below are deprecated.
    *
@@ -63,7 +68,7 @@ export class Tabs {
   render(): VNode {
     return (
       <Fragment>
-        <slot name={SLOTS.tabNav} />
+        <slot name={SLOTS.tabNav} onSlotchange={tabNavSlotChangeHandler} />
         <section>
           <slot />
         </section>
@@ -146,12 +151,33 @@ export class Tabs {
    */
 
+  tabNav: HTMLCalciteTabNavElement;
+
   //
   //  Private Methods
   //
 
+  updateTabNavLayout = (): void => {
+    const { tabNav } = this;
+
+    if (tabNav) {
+      tabNav.layout = this.layout;
+    }
+  };
+
+  tabNavSlotChangeHandler = (event: Event): void => {
+    const tabNavs = (event.target as HTMLSlotElement)
+      .assignedElements({
+        flatten: true
+      })
+      .filter((el) => el?.matches("calcite-tab-nav")) as HTMLCalciteTabNavElement[];
+
+    this.tabNav = tabNavs[0];
+    this.updateTabNavLayout();
+  };
+
   /**
    *
    * Matches up elements from the internal `tabs` and `titles` to automatically

```

Let me know if this is what you were thinking.

Elijbet: Neat! :golf: This would have to be done to tab-title as well though yeah? It has a CSS centering on it that also needs to run when layout="center" at the top level.

driskull: Yep. that makes sense.

Elijbet:
Closing this PR as it solves a slightly different problem than presented by the actual repo case.

PR #5104, opened 2022-08-05
feat: emit calciteThemeChange event when dynamically toggling between light and dark modes

benelan:
**Related Issue:** https://devtopia.esri.com/WebGIS/arcgis-js-api/issues/43744

## Summary
Adds a `calciteThemeChange` event which emits when switching themes using CSS clases __on \<body>__ or through OS preferences.  It will allow the JSAPI and others to change their own UI component's theming to match Calcite. The current theme is specified in the CustomEvent's `detail` property.

```js
document.addEventListener("calciteThemeChange", (event) => {
  console.log(event.detail.theme) // "light" or "dark"
  }
});
```

Tested using the demos, it works when toggling the dark theme class

```js
document.body.classList.toggle("calcite-theme-dark");
```
as well as explicitly swapping light, dark, and/or auto theme classes.

```js
darkMode = !darkMode; // assumes dark mode OS preference for testing purposes
if (darkMode) {
  document.body.classList.remove("calcite-theme-light");
  document.body.classList.add("calcite-theme-auto");
} else {
  document.body.classList.remove("calcite-theme-auto");
  document.body.classList.add("calcite-theme-light");
}
```

When using `calcite-theme-auto`, the calciteThemeChange event is emitted with the appropriate light or dark value depending on the OS preferences.

calciteThemeChange


driskull (approved):
Awesome. 🥳

src/assets/styles/global.scss
```diff
@@ -1,6 +1,23 @@
 @import "./node_modules/@esri/calcite-colors/dist/colors.scss";
 @import "./node_modules/@esri/calcite-base/dist/_index.scss";
 
+@keyframes calciteThemeChangeLight {
+  from {
+    /**/
```
jcfranco, line 6: Can you add comments to these about their use and the intentional noops?

src/assets/styles/global.scss
```diff
@@ -51,22 +68,30 @@
...
+    animation-name: calciteThemeChangeDark;
   }
 
   @media (prefers-color-scheme: light) {
     @include calcite-theme-light-extended();
+    animation-name: calciteThemeChangeLight;
   }
 }
 
+body:not(.calcite-theme-dark):not(.calcite-theme-auto):not(.calcite-theme-light) {
```
jcfranco, line 91: Is this meant exclusively for `body` or anywhere a theme can be applied? I believe `animationstart` bubbles, so this could lead to some boilerplate on end-users.

benelan: Yeah this part is only meant to work on `body`. The main scenario for dynamically changing themes is for the full app. This handles the default scenario when no theme class is set and it defaults to light. 

If people are dynamically changing themes for different parts of their UI - that seems concerning right? But even if there is a valid use case, they can explicitly add remove the dark/light themes and it will work fine. I'm going to document all of that here (until we find a better place):
https://developers.arcgis.com/calcite-design-system/faq/#how-do-i-change-themes

driskull: For the JSAPI use case, we can just listen for the animation change on body right? I don't think we need to be concerned about themes changing on any other elements?

benelan: > For the JSAPI use case, we can just listen for the animation change on body right? I don't think we need to be concerned about themes changing on any other elements?

Correct

benelan:
@jcfranco I abstracted to a `calciteThemeChange` CustomEvent. Is this good for now? With the abstraction we can always look into using a mutation observer in the future w/o impacting our users.

src/utils/theme.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Emits when the theme is dynamically toggled between light and dark on <body>.
+ * This file is imported in Stencil's `globalScript` config option.
+ *
+ * @see {@link https://stenciljs.com/docs/config#globalscript Stencil's globalScript property}
+ */
+export default function (): void {
+  const { classList } = document.body;
+  const prefersDark = window.matchMedia && window.matchMedia("(prefers-color-scheme: dark)").matches;
```
jcfranco, line 11: I think we can drop the check for `matchMedia`.

stencil.config.ts
```diff
@@ -97,6 +97,7 @@ export const create: () => Config = () => ({
     }
   ],
   globalStyle: "src/assets/styles/global.scss",
+  globalScript: "src/utils/theme.ts",
```
jcfranco, line 100: This can be done separately, but can you introduce a global script? It’ll be clearer without looking at the config and allow us to add additional things if needed.

src/utils/theme.ts
```diff
@@ -0,0 +1,39 @@
+import { autoTheme, darkTheme } from "./resources";
+
+/**
+ * Emits when the theme is dynamically toggled between light and dark on <body>.
+ * This file is imported in Stencil's `globalScript` config option.
+ *
+ * @see {@link https://stenciljs.com/docs/config#globalscript Stencil's globalScript property}
+ */
+export default function (): void {
```
jcfranco, line 9: Once we have a global script can you name this function to indicate it sets up theme update changes?

jcfranco: You’ll also need to update this method to return early if not running on the browser, so the SSR build doesn’t go 💥.

src/utils/theme.ts
```diff
@@ -0,0 +1,39 @@
...
+  // emits event on page load
+  emitThemeChange(currentTheme);
+
+  // emits event when changing OS theme preferences
+  window
+    .matchMedia("(prefers-color-scheme: dark)")
+    .addEventListener("change", (event) => themeChangeHandler(event.matches ? "dark" : "light"));
+
+  // emits event when toggling between theme classes on <body>
+  new MutationObserver(() => themeChangeHandler(getTheme())).observe(document.body, {
```
jcfranco, line 35: Is it possible to add a test for this? This would be the first one we try for the global script.

jcfranco (approved):
Awesome! 🌮🎉🦖

src/utils/theme.ts
benelan: I tried creating a new file `src/utils/theme.e2e.ts` as well as adding it to an existing component's e2e tests, but neither worked. I tried w/ and w/o adding `<body>` in the HTML content. I also tried "sleeping" in case it was a timing issue:
`await new Promise((r) => setTimeout(r, 2000));`

@jcfranco / @driskull Do you notice any issues with the code below, or have any other ideas?

<details>

  <summary>Attempt 1</summary>

```ts
import { html } from "../../support/formatting";

it("emits calciteThemeChange when setting calcite-theme-dark class on body", async () => {
  await page.setContent(html`<calcite-button>Test</calcite-button>`);
  const calciteThemeChangeSpy = await page.spyOnEvent("calciteThemeChange");

  document.body.classList.add("calcite-theme-dark");
  await page.waitForChanges();

  expect(document.body.classList.contains("calcite-theme-dark")).toBe(true);
  expect(calciteThemeChangeSpy).toHaveReceivedEventTimes(1);
  expect(calciteThemeChangeSpy.lastEvent.detail.theme).toBe("dark");

  document.body.classList.toggle("calcite-theme-dark");
  await page.waitForChanges();

  expect(document.body.classList.contains("calcite-theme-dark")).toBe(false);
  expect(calciteThemeChangeSpy).toHaveReceivedEventTimes(2);
  expect(calciteThemeChangeSpy.lastEvent.detail.theme).toBe("light");
});

```
image

</details>

<details>

  <summary>Attempt 2</summary>

```ts
import { html } from "../../support/formatting";

it("emits calciteThemeChange when setting calcite-theme-dark class on body", async () => {
  await page.setContent(html`<calcite-button>Test</calcite-button>`);

  const emitted = await page.evaluate(() => {
    const emittedEvents = [];
    document.body.addEventListener("calciteThemeChange", (event) => emittedEvents.push(event));
    document.body.classList.toggle("calcite-theme-dark");
    document.body.classList.remove("calcite-theme-dark");
    return emittedEvents;
  });

  expect(emitted).toHaveLength(2);
  expect(emitted[0].detail.theme).toBe("dark");
  expect(emitted[1].detail.theme).toBe("light");
});

```

image

</details>

stencil.config.ts
benelan: I created `src/utils/index.ts` which is now the file used for the `globalScript` stencil config option. Is that what you meant? 

Is that a good filename and directory? An alternative filename could be `globalScript.ts` and alternative directories could be `support` or `src`.

src/utils/theme.ts
benelan: I logged a ticket with stencil to see if creating an e2e test for code imported through the globalScript is possible before I spend more timing looking into this

benelan: Update from ionic:

> The globalScript should be called as part of the newE2EPage() call, but you cannot do much direct testing of it. Your tests seem more indirect, though, so you should be OK. Looking at your PR, it seems like attempt 1 should have worked. If you can create a simplified version of this and share it, I would be happy to have a look.

I was able to [repro with the stencil starter](https://github.com/benelan/stencil-component-starter) so Ken is looking into it

stencil.config.ts
jcfranco: `utils` makes sense to me. I'd suggest renaming to `global` or `globalScript` to make its use clearer.

src/utils/theme.ts
benelan: Installing this so it makes it into the release I'll be doing shortly. I'll add a test in a separate PR when Stencil gets back to us (if it's possible)

Issue #4971, opened 2022-07-19
`calcite-tabs` ignores `layout="center"` attribute


seanthayer:
### Actual Behavior

When `calcite-tabs` is passed a `layout="center"` attribute, the `calcite-tab-nav` component does not center.

### Expected Behavior

The `calcite-tab-nav` component would center its content along the element, as per documentation.

### Reproduction Sample

https://codepen.io/seanthayer/pen/VwXpgNd

### Reproduction Steps

1. Create `calcite-tabs` with `layout="center"` attribute
2. Slot `calcite-tab-nav` with `slot="tab-nav"` attribute
3. Compose one or many `calcite-tab-title` as children of `calcite-tab-nav`
4. Render and observe incorrect layout

### Reproduction Version

`beta.86`

### Relevant Info

The layout appears to stop working correctly starting at `beta.84`.

**Note**: The component layout works correctly when adjusting via "knobs" in storybook, however in lone implementation or usage in another component it appears to break down.

### Regression?

`beta.83`

Elijbet:
Wasn’t able to replicate this issue working properly at either beta.83 or beta.82. 
Setting layout snapshots at these levels results in the following:

`tabs` does not center
<img width="999" alt="Screen Shot 2022-08-05 at 1 24 57 PM" src="https://user-images.githubusercontent.com/19231036/183158582-2ed8dc42-20c1-40c5-bd82-95e0e8437864.png">

`tab-nav` does center but doesn’t center titles
<img width="1037" alt="Screen Shot 2022-08-05 at 1 25 19 PM" src="https://user-images.githubusercontent.com/19231036/183158608-e68e407f-f5e0-4db1-9840-9d57a7c7d68e.png">

`tab-title` centers


I was suggesting a solution that aligns layout on all nested components if you set it at levels above.
So if you set at top level `tabs`, you’d get this result:
<img width="1034" alt="Screen Shot 2022-08-05 at 1 25 49 PM" src="https://user-images.githubusercontent.com/19231036/183158612-5968cc00-5804-44e5-a12f-d3da73f6ac64.png">

geospatialem:
Verified on next.557

![verify calcite-tabs center layout](https://user-images.githubusercontent.com/5023024/183941301-b076bec1-d355-4ee5-b44e-2ff57c24510c.png)

PR #5112, opened 2022-08-08
fix(tabs): reverts regression that makes tabs ignore layout center

Elijbet:
**Related Issue:** #4971

## Summary

Reverts regression that makes `tabs` ignore `layout="center"`. When `calcite-tabs` is passed a `layout="center"` attribute, the `calcite-tab-nav` component now centers again.


driskull (approved):
💯 🥇

Issue #3121, opened 2021-09-28
Enhancement: [calcite-tree] add selection-mode "none"


caripizza:
### Description
Support a calcite-tree selection-mode of "none", for displaying a hierarchical list of tree-items which display no icons on hover/focus, and do not allow any items to be selected.

### Acceptance Criteria 
A new selection-mode of "none" is added to the [TreeSelectionMode](https://github.com/Esri/calcite-components/blob/master/src/components/calcite-tree/interfaces.ts#L5) type values. When a tree has selection-mode "none", no child tree-items can be selected, and no icons appear on :hover or :focus.

### Relevant Info 
Related issue tied to this request: #2210 (cc @bpatterson88)

#### Which Component
calcite-tree, calcite-tree-item

#### Example Use Case
(See screenshots provided in #2210)

caripizza:
Here's one example I found in the wild of "none" selection-mode with IntegralUI's TreeView component for Angular:
https://www.lidorsystems.com/support/articles/angular/treeview/treeview-multi-select.aspx

macandcheese:
This implementation from SAP also uses "none":  https://sap.github.io/ui5-webcomponents/playground/components/Tree/

ellenupp:
Use case for Dashboards:
We are using a tree to represent the layout of the elements in the dashboard. The tree does not need a selection state. Clicking a tree item should place focus on that item but not otherwise change its state. Tree items don't need bullet point and instead child items have a custom icon. 
![image](https://user-images.githubusercontent.com/14304581/135142106-bee99dcb-0f19-488c-9d85-6ba50112066b.png)

paulcpederson:
We also have a use case for this. Seems pretty straight forward, I could try to add this?

benelan:
Sure, @paulcpederson, we always appreciate your help!

jcfranco:
Stealing this one... 👿

geospatialem:
Verified in beta.92 👍🏻

PR #5128, opened 2022-08-10
feat(tree): add none selection mode

jcfranco:
**Related Issue:** #3121

## Summary

This introduces a new `none` selection mode that will still allow tree navigation, but prevent an item from being selected. The `calciteTreeSelect` event will emit as well after an item is 'selected'.

### DIFF
--- a/src/components/tree-item/tree-item.scss
+++ b/src/components/tree-item/tree-item.scss
@@ -5,6 +5,12 @@
     cursor-pointer;
 }
 
+:host([selection-mode="none"]:not([has-children])) {
+  .node-container {
+    margin-inline: theme("margin.3");
+  }
+}
+
 @include calciteHydratedHidden();
 
 :host([scale="s"]) {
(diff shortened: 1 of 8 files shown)


jcfranco:
@macandcheese Could you test the latest tweaks? 🙇

macandcheese (approved):
✅

src/components/tree-item/tree-item.scss
```diff
@@ -177,7 +183,7 @@
 }
 
 // dropdown expanded and not selected
-:host([has-children][expanded]:not([selected])) > .node-container {
+:host([has-children][expanded]:not([selected]):not([selection-mode="none"])) > .node-container {
```
driskull, line 186: nitpick: some of these selectors are getting pretty complex. Maybe at some point we move this logic to the tsx file?

driskull (approved):
LGTM!

src/components/tree-item/tree-item.scss
jcfranco: Yeah, I went with the simplest option for now. I think we'll have to do a pass on complex selectors and see which ones are better off as dynamic classes vs. their styling equivalent. cc @benelan

Issue #5070, opened 2022-08-02
Calcite-input can lose focus when next to slider in certain scenario


zaramatheson:
### Actual Behavior

To reproduce:
Go to highlight the input value, drag the mouse outside of the input box and release the mouse select button and you see the slider get highlighted and input box loses focus

![HoldnRelease](https://user-images.githubusercontent.com/50747646/182474711-5ad5b7bc-00f0-40b9-a993-4ac95e73ffde.gif)

### Expected Behavior

The focus should remain on the input. 

### Reproduction Sample

Map Viewer Sketch Pane 

https://codepen.io/jcfranco/pen/wvmxJvG?editors=1010

### Reproduction Steps

To reproduce:
Go to highlight the input value, drag the mouse outside of the input box and release the mouse select button and you see the slider get highlighted and input box loses focus

### Reproduction Version

latest

jcfranco:
I believe this is a wrapping-component issue and not specific to Calcite. I should know because I created it. 😂 In any case, I'll dig into it and will close this one my hunch is correct.

jcfranco:
Quick update: After digging into this, I think it's best if we update label to not assume there will only be one labelable. For comparison, browsers only focus on the first labelable and ignore the rest.

Apahadi73:
I noticed the same issue in ArcGIS Charts app. It's similar to this issue. When I try to click on `calcite-input` element, the focus shifts to nearby `calcite-slider` element. Both of these elements are inside `calcite-label` element. Here is the code-pen used for the issue repro - https://codepen.io/apahadi73/pen/JjaELpW.

driskull:
See findings from https://github.com/Esri/calcite-components/issues/7085#issuecomment-1572890663

geospatialem:
Verified in `1.7.0-next.9`.

AdelheidF:
Mhh, I can still reproduce it using the codepen with 1.7.0 and 1.9.2.

Click on the input box and hold the mouse button then move the mouse over the slider line and release the mouse button. Now the slider button is in focus.

geospatialem:
> Mhh, I can still reproduce it using the codepen with 1.7.0 and 1.9.2.
> 
> Click on the input box and hold the mouse button then move the mouse over the slider line and release the mouse button. Now the slider button is in focus.

@AdelheidF Thanks for getting this back on our radar! 

[...]

PR #5134, opened 2022-08-11
fix(label): handle first labelable child only

jcfranco:
**Related Issue:** #5070 

## Summary

Pending: tests! 🚧


jcfranco:
Closing. Will revisit solution.

Issue #4976, opened 2022-07-20
Add `groupSeparator` prop to Slider


geospatialem:
### Actual Behavior

The Slider component displays numbers that are comma separated, but not all situations require a comma, such as years (`1970`).

We've received a few Community posts asking for functionality to support years on:
- [May 14](https://community.esri.com/t5/calcite-design-system-questions/format-labels-values-on-slider/m-p/1173994#M137) and
- [July 17](https://community.esri.com/t5/calcite-design-system-questions/remove-digit-separator-from-slider/m-p/1192723#M179)

### Expected Behavior

Users should be able to use a `groupSeparator` prop to display numbers with, or without separated. 

Where if true, displays the separator (e.g., `1,970`) and if false/not present does not (e.g., `1970`).

### Reproduction Sample

https://codepen.io/geospatialem/pen/wvmdodY

### Reproduction Steps

1. Visit the [sample](https://codepen.io/geospatialem/pen/wvmdodY)
2. Observe the year displayed in the first `slider` example, displaying the year as `1,970` (with a comma separator)
3. Observe the year displayed in the second `slider` example, displaying the year as-expected with a value of `1970`

### Reproduction Version

beta.86

### Relevant Info

Add a `groupSeparator` prop, similar to some other components, such as [Input](https://developers.arcgis.com/calcite-design-system/components/input/#component-api-properties-groupSeparator) for API consistency, and enabling users to display years, in addition to numbers in the `slider` component.

cc https://github.com/Esri/calcite-components/issues/4004

geospatialem:
Verified on `next`.

PR #5138, opened 2022-08-11
feat(slider)!: add a groupSeparator prop to enable toggling display of separator in label

Elijbet:
**Related Issue:** #4976

## Summary

Adds a `groupSeparator` prop to `slider` which displays numbers as separated and localized when true (e.g. 1,970 for 'en') and does not when false/not present (e.g. 1970). `lang` and `numberingSystem` props are added to support `localizeNumberString` util used for handling the localization.

Before this change, `slider` was automatically adding the language's corresponding number separator to the value label.

Adds a number locale support test to check for tick min/max and thumb min/max localized display values based on `groupSeparator` prop.


src/components/slider/slider.tsx
```diff
@@ -150,6 +150,16 @@ export class Slider implements LabelableComponent, FormComponent, InteractiveCom
    */
   @Prop() scale: Scale = "m";
 
+  /**
+   * When true, displays numbers as comma separated (e.g. 1,970) and when false/not present does not (e.g. 1970)
+   */
+  @Prop({ reflect: true, mutable: true }) groupSeparator = false;
+
+  @Watch("groupSeparator")
```
driskull, line 158: @Elijbet I don't think this watcher is necessary as it's just setting the value to what it has already been set as.

src/components/slider/slider.tsx
```diff
@@ -574,13 +584,13 @@ export class Slider implements LabelableComponent, FormComponent, InteractiveCom
...
-          {this.minValue && this.minValue.toLocaleString()}
+          {this.minValue && this.determineGroupSeparator(this.minValue)}
         </span>
         <span aria-hidden="true" class="handle__label handle__label--minValue static">
-          {this.minValue && this.minValue.toLocaleString()}
+          {this.minValue && this.determineGroupSeparator(this.minValue)}
         </span>
         <span aria-hidden="true" class="handle__label handle__label--minValue transformed">
-          {this.minValue && this.minValue.toLocaleString()}
+          {this.minValue && this.determineGroupSeparator(this.minValue)}
```
driskull, line 593: Instead of all the `this.minValue && ` maybe `determineGroupSeparater` could handle when the value is not set?

src/components/slider/slider.tsx
```diff
@@ -1396,4 +1406,13 @@ export class Slider implements LabelableComponent, FormComponent, InteractiveCom
...
     return intersects(maxLabelBounds, handleBounds);
   }
+
+  /**
+   * Returns a string representing the label value that is parsed for comma separator based on whether groupSeparator prop is set to true/false.
+   *
+   * @param input
+   */
+  private determineGroupSeparator = (input: number): string => {
+    return this.groupSeparator ? input.toLocaleString() : input.toLocaleString().replace(",", "");
```
driskull, line 1416: question: Does `input.toLocaleString()` always have commas or could there be other number separators for different languages?

src/components/slider/slider.tsx
```diff
@@ -574,13 +584,13 @@ export class Slider implements LabelableComponent, FormComponent, InteractiveCom
         <div class="handle-extension" />
         <div class="handle" />
         <span aria-hidden="true" class="handle__label handle__label--minValue">
-          {this.minValue && this.minValue.toLocaleString()}
+          {this.minValue && this.determineGroupSeparator(this.minValue)}
```
driskull, line 587: The `minValue` of `0` won't call `determineGroupSeparator` Is this ok?

driskull (commented):
Code looks good! Have some comments/questions regarding a few things.

src/components/slider/slider.e2e.ts
```diff
@@ -673,4 +673,49 @@ describe("calcite-slider", () => {
   it("is form-associated", () => formAssociated("calcite-slider", { testValue: 5 }));
 
   it("is form-associated with range", () => formAssociated("calcite-slider", { testValue: [5, 10] }));
+
+  it("groupSeparator: displays numbers as comma separated when prop is true and does not when false", async () => {
+    const page = await newE2EPage({
```
jcfranco, line 678: For consistency, can you create an empty page and use `setContent` afterwards? No need for `waitForChanges` after doing this.

src/components/slider/slider.e2e.ts
```diff
@@ -673,4 +673,49 @@ describe("calcite-slider", () => {
...
+      const labelVal = (await element.shadowRoot.querySelector("span.handle__label--value")) as HTMLElement;
+
+      const tickMin = (await element.shadowRoot.querySelector("span.tick__label--min")) as HTMLElement;
+      const tickMax = (await element.shadowRoot.querySelector("span.tick__label--max")) as HTMLElement;
+
+      return [labelMinVal.innerText, labelVal.innerText, tickMin.innerText, tickMax.innerText];
+    };
+    await page.exposeFunction("displayedValuesArray", displayedValuesArray);
+
+    const noSeparator = await page.$eval("calcite-slider", async (): Promise<string[]> => {
```
jcfranco, line 707: I think you can get rid of the wrapping anonymous functions and pass `displayedValuesArray` directly.

src/components/slider/slider.e2e.ts
```diff
@@ -673,4 +673,49 @@ describe("calcite-slider", () => {
...
+                label-handles 
+                label-ticks 
+                style="width:${sliderWidthFor1To1PixelValueTrack}">
+            </calcite-slider>`
+    });
+    await page.waitForChanges();
+    const element = await page.find("calcite-slider");
+
+    const displayedValuesArray = async (): Promise<string[]> => {
+      await page.waitForChanges();
```
jcfranco, line 695: I think you can get rid of this since you already wait for changes before calling this method.

src/components/slider/slider.e2e.ts
```diff
@@ -673,4 +673,49 @@ describe("calcite-slider", () => {
...
+    });
+
+    element.setProperty("groupSeparator", true);
+    await page.waitForChanges();
+
+    const withSeparator = await page.$eval("calcite-slider", async (): Promise<string[]> => {
+      return await displayedValuesArray();
+    });
+
+    expect([...noSeparator]).toEqual(["2500", "5000", "1000", "10000"]);
```
jcfranco, line 718: You can pass the arrays directly and not need to spread values.

src/components/slider/slider.tsx
```diff
@@ -257,13 +262,13 @@ export class Slider implements LabelableComponent, FormComponent, InteractiveCom
         tabIndex={0}
       >
         <span aria-hidden="true" class="handle__label handle__label--value">
-          {value ? value.toLocaleString() : value}
+          {value ? this.determineGroupSeparator(value) : value}
```
jcfranco, line 265: Optional: can you compute the value once, store in a local variable and reuse?

src/components/slider/slider.tsx
jcfranco: I agree with the suggestion. Looks like it was guarding against it to avoid calling a number on `null`/`undefined`.

src/components/slider/slider.tsx
jcfranco: Can you use [`localizeNumberString` from `src/utils/locale.ts`](https://github.com/Esri/calcite-components/blob/master/src/utils/locale.ts#L106) instead? It should only be used if `groupSeparator` is true and return the value as is otherwise. +@eriklharper, @benelan for additional hints

src/components/slider/slider.tsx
```diff
@@ -1396,4 +1401,13 @@ export class Slider implements LabelableComponent, FormComponent, InteractiveCom
     const handleBounds = handle.getBoundingClientRect();
     return intersects(maxLabelBounds, handleBounds);
   }
+
+  /**
+   * Returns a string representing the label value that is parsed for comma separator based on whether groupSeparator prop is set to true/false.
+   *
+   * @param input
+   */
+  private determineGroupSeparator = (input: number): string => {
```
jcfranco, line 1410: Can you rename this method to better express what it does? The current name doesn't tell me that it returns a formatted value.

Also, can you rename `input` to `value` or something of that nature?

src/components/slider/slider.e2e.ts
```diff
@@ -673,4 +673,49 @@ describe("calcite-slider", () => {
...
+    expect([...withSeparator]).toEqual(["2,500", "5,000", "1,000", "10,000"]);
+  });
```
jcfranco, line 720: Great test! Can you update it to cover additional locales? Something like https://github.com/Esri/calcite-components/blob/adcbb55b3ab3ab9fca0921b4bbbea1e4bf92907b/src/components/input/input.e2e.ts#L1057 would be good.

jcfranco (changes requested):

src/components/slider/slider.e2e.ts
Elijbet: Oops, sorry.

src/components/slider/slider.tsx
Elijbet: > commas

 Yeah, good point. A period is used as the thousands separator in many non-English speaking countries.

src/components/slider/slider.tsx
```diff
@@ -150,6 +151,21 @@ export class Slider implements LabelableComponent, FormComponent, InteractiveCom
    */
   @Prop() scale: Scale = "m";
 
+  /**
+   * When true, number values are displayed with the locale's group separator.
+   */
+  @Prop({ reflect: true }) groupSeparator = false;
+
+  /** Specifies the BCP 47 language tag for the desired language and country format. */
+  @Prop() locale: string = document.documentElement.lang || "en";
```
driskull, line 160: These might be related to this:

https://github.com/Esri/calcite-components/issues/4576
https://github.com/Esri/calcite-components/issues/4213

Should we start prioritizing these? Ideally, we should just use the global `lang` attribute.

driskull (commented):
Looks good @Elijbet!!!

My only concern is if we should not introduce the `locale` property here since we are planning on deprecating it elsewhere.

src/components/slider/slider.tsx
Elijbet: Ok, I'll make the change to `lang` instead.

jcfranco: These are scheduled for this month according to the roadmap. 😄

jcfranco: In this PR's context, let's move forward w/ `lang`. I forgot `pagination` already was using our global util, so let's use that moving forward and we'll align other components this month.

src/components/slider/slider.tsx
```diff
@@ -150,6 +151,21 @@ export class Slider implements LabelableComponent, FormComponent, InteractiveCom
    */
   @Prop() scale: Scale = "m";
 
+  /**
+   * When true, number values are displayed with the locale's group separator.
+   */
+  @Prop({ reflect: true }) groupSeparator = false;
```
jcfranco, line 157: Can you move this prop so it's sorted alphabetically?

jcfranco (approved):
Aside from a few comments, this LGTM! 🎉🌮🦖🦃

src/components/slider/slider.e2e.ts
```diff
@@ -673,4 +673,112 @@ describe("calcite-slider", () => {
...
+      element.setProperty("locale", "hi");
+      await page.waitForChanges();
+
+      withSeparator = await page.$eval("calcite-slider", async (): Promise<string[]> => {
+        return await getDisplayedValuesArray();
+      });
+      expect(withSeparator).toEqual(["2,500", "5,00,000.5", "1,000", "10,00,000.5"]);
+    });
+
+    it("Spanish", async () => {
```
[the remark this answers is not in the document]
Elijbet: It all gotta Spanish!

src/components/slider/slider.e2e.ts
```diff
@@ -673,4 +673,112 @@ describe("calcite-slider", () => {
...
+      await page.waitForChanges();
+
+      noSeparator = await page.$eval("calcite-slider", async (): Promise<string[]> => {
+        return await getDisplayedValuesArray();
+      });
+      expect(await element.getProperty("groupSeparator")).toBe(false);
+      expect(noSeparator).toEqual(expectedNotSeparatedValueArray);
+    });
+
+    it("English", async () => {
```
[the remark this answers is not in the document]
Elijbet: **### 🧽🧼🍽✨**

src/components/slider/resources.ts
```diff
@@ -0,0 +1,6 @@
+export const CSS = {
+  labelMinValue: "span.handle__label--minValue",
```
driskull, line 2: how come these have `span` inside them? It doesn't seem like we should need to have the element type in the class

src/components/slider/slider.tsx
```diff
@@ -1396,4 +1423,18 @@ export class Slider implements LabelableComponent, FormComponent, InteractiveCom
...
     return intersects(maxLabelBounds, handleBounds);
   }
+
+  /**
+   * Returns a string representing the localized label value based on groupSeparator prop being on or off.
+   *
+   * @param value
+   */
+  private determineGroupSeparator = (value): string => {
+    const lang = this.globalAttributes["lang"] || document.documentElement.lang || "en";
```
driskull, line 1433: I wonder instead of `document.documentElement.lang` we should do `this.el.closes(["lang"]).lang`? What do you think @jcfranco @Elijbet ? Right now, it doesn't honor parent elements lang.

I'm ok with enhancing this in globalAttributes helper if we can do that instead. Maybe we can enhance globalAttributes to do closest()?

driskull (approved):
LGTM! Maybe just fix the span elements in the CSS?

src/components/slider/resources.ts
```diff
@@ -1,6 +1,6 @@
 export const CSS = {
-  labelMinValue: "span.handle__label--minValue",
-  labelValue: "span.handle__label--value",
-  tickMin: "span.tick__label--min",
-  tickMax: "span.tick__label--max"
+  labelMinValue: ".handle__label--minValue",
```
driskull, line 2: @Elijbet these shouldn't have the dot either. `.`

src/components/slider/slider.tsx
jcfranco: This is a compromise right now to avoid querying the DOM between renders. We can doc this limitation, create an issue for adding `lang` lookup and wait for user feedback since users would be expect that to work. cc @geospatialem @benelan

.gitignore
```diff
@@ -33,3 +33,4 @@ UserInterfaceState.xcuserstate
 log.txt
 src/components.d.ts
 src/components/icon/assets
+
```
jcfranco, line 36: This and the `package-lock.json` changes are unrelated, please restore.

src/components/slider/slider.e2e.ts
```diff
@@ -721,64 +730,30 @@ describe("calcite-slider", () => {
...
+          return await getDisplayedValuesArray();
+        });
+        expect(withSeparator).toEqual(formattedValuesArr);
+      };
 
-      withSeparator = await page.$eval("calcite-slider", async (): Promise<string[]> => {
-        return await getDisplayedValuesArray();
-      });
-      expect(withSeparator).toEqual(["2500", "500.000,5", "1000", "1.000.000,5"]);
+      Object.entries(formattedValuesPerLanguageObject);
```
jcfranco, line 753: Straggler line, please remove.

src/components/slider/slider.e2e.ts
```diff
@@ -721,64 +730,30 @@ describe("calcite-slider", () => {
...
+      for (const lang of Object.keys(formattedValuesPerLanguageObject)) {
```
jcfranco, line 754: This could be simplified to:

```ts
for (const lang in formattedValuesPerLanguageObject) { /...
```

src/components/slider/resources.ts
```diff
@@ -0,0 +1,6 @@
+export const CSS = {
+  labelMinValue: "handle__label--minValue",
```
jcfranco, line 2: Nice! Next step would be to use this object within the slider's rendering.

jcfranco (changes requested):

Issue #4813, opened 2022-06-28
Implement an `action` indicator with icon color override with screen reader context (active/inactive)


geospatialem:
### Summary

Screen readers should receive confirmation of the `action`s state (active/inactive).

**_Additional issues stemmed from this issue include #5003, #5005_**

(Some additional issues for consideration: #1220, #3950) 

### Actual Behavior

Context of the `active` property is not provided to screen reader users.

### Expected Behavior

Provide screen reader context based on the `active` prop (e.g., action / button is active) using `aria-selected`, or similar.

### Reproduction Sample

https://codepen.io/geospatialem/pen/QWmWdoQ

### Reproduction Steps

1. Open the [Codepen sample](https://codepen.io/geospatialem/pen/QWmWdoQ)
2. Keyboard in or mouse over the second action (speaker) to depict the difference between an `active` state vs  `:focus`/`:active`/`:hover` states.

### Reproduction Version

1.0.0-beta.83

geospatialem:
Some additional research is needed to proceed.

Additional context provided by @macandcheese from https://github.com/Esri/calcite-components/pull/5139#issuecomment-1212478332:

> I think the issue is that [`action`] can be used in multiple ways.

> If you were to use it alongside other actions, in a group, bar, or pad - an app needs to manage which is active. This would be a menu item or potentially more of a radio group type of interaction, even if the action itself doesn't have knowledge of the other action's states.

> It can also be used as a toggle, individually from other actions (dark mode on / off) - this is where the toggle / checkbox role seems like it could apply

> It sounds like to me, the aria-pressed could be keyed off of the active prop - even if the action component itself doesn't manage that.

> Enter / Space could still be applicable even when there isn't an "active" state associated in the UI - for instance, a "Save" action - this would be interacted by click, enter, space, but wouldn't have a visual representation once clicked.

geospatialem:
This issue came up while putting together the component sample, as context isn't provided to assistive technology (AT) users.

When selection an `action` in an `action-menu`, this would confuse the AT user that isn't provide context when interacting with the `action`s.

driskull:
I think this issue needs more discussion and direction before we can do anything.

We have an `active` property which is probably better representative of `aria-selected`. But since `action` is a role of `button` I think we have to use `aria-pressed` here in order to not misuse aria attributes.

I'm not sure what aria attribute `indicator` would map to exactly.

driskull:
[...]

PR #5139, opened 2022-08-11
fix(action): Assistive technology support and announcement for "active" prop

geospatialem:
**Related Issue:** #4813 

## Summary
Adds an `aria-pressed` attribute that provides context to screen reader users (tested with JAWS, NVDA, and VoiceOver) when the `action` component is pressed via click, or the Enter or spacebar keys.

Additionally, toggles the `active` prop when the button is clicked, or pressed via the Enter or spacebar keys, via `calciteActionClickHandler`.

### DIFF
--- a/src/components/action/action.tsx
+++ b/src/components/action/action.tsx
@@ -220,6 +220,7 @@ export class Action implements InteractiveComponent {
           aria-busy={toAriaBoolean(loading)}
           aria-disabled={toAriaBoolean(disabled)}
           aria-label={ariaLabel}
+          aria-pressed={this.active ? "true" : "false"}
           class={buttonClasses}
           disabled={disabled}
           ref={(buttonEl): HTMLButtonElement => (this.buttonEl = buttonEl)}
@@ -240,6 +241,7 @@ export class Action implements InteractiveComponent {
   calciteActionClickHandler = (): void => {
     if (!this.disabled) {
       this.calciteActionClick.emit();
+      this.active = !this.active;
     }
   };
 }


geospatialem:
@macandcheese Can you verify if this fix fits the UX/UI of the component, specifically for mouse users?

macandcheese:
I think the application should still be managing the state of the action (`active` or not), since there are use cases for an action to be interacted with and not become `active`

I'm not sure if `aria-pressed` will always be applicable - since (from my understanding) - this would be used when the action is representative of a toggle.

geospatialem:
> I think the application should still be managing the state of the action (`active` or not), since there are use cases for an action to be interacted with and not become `active`
> 
> I'm not sure if `aria-pressed` will always be applicable - since (from my understanding) - this would be used when the action is representative of a toggle.

Appreciate the insights! 

Currently screen readers are announcing to activate the component folks must press Enter or spacebar, so its a bit misleading as-is, as neither key alters the state, or provides additional feedback.

Is the intent of `active` to only be visual, similar to `:focus`? If so, we wouldn't need to announce the prop value to screen readers, as they'd hear the component text/label on focus. The current experience doesn't allow assistive technologies to switch the `active` prop, despite what is being read back to them.

Otherwise, since the component is housed in a native button, we might need to consider a differing role to support the component intent a bit more, perhaps a [`role="checkbox"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role) where `aria-checked` could be `true`, `false`, or `mixed`. We could potentially use `mix` to showcase a few different states.

@macandcheese What route (or even "door number three") would you recommend based on the UX/UI?

macandcheese:
I think the issue is that is can be used in multiple ways. 

If you were to use it alongside other actions, in a group, bar, or pad - an app needs to manage which is active. This would be a menu item or potentially more of a radio group type of interaction, even if the action itself doesn't have knowledge of the other action's states. 

It can also be used as a toggle, individually from other actions (dark mode on / off) - this is where the toggle / checkbox role seems like it could apply

It sounds like to me, the `aria-pressed` could be keyed off of the `active` prop - even if the action component itself doesn't manage that.

Enter / Space could still be applicable even when there isn't an "active" state associated in the UI - for instance, a "Save" action - this would be interacted by click, enter, space, but wouldn't have a visual representation once clicked.

geospatialem:
Ah, lots of potential outcomes as a result then. 

I'm going to close this PR, since we'll likely need to do work across a few different components and verify their roles. Going to add some of the great insights you listed here in the original issue for context, and add a `research` label to it to determine some possibilities moving forward.

Issue #5022, opened 2022-07-26
calcite-input: browser-saved logins work but do not display


gpbmike:
### Actual Behavior

If you use calcite-input in a login form and the user saves login information to the browser, on return visits, the inputs appear to be empty. However, the "hidden-form-input" will have the values.

https://user-images.githubusercontent.com/8754/181050115-3c7c77d5-b266-41ca-83d8-6ddd888dd0e7.mov

### Expected Behavior

Changes to the value of "hidden-form-input" should propagate to the visible inputs so the user knows the autofill worked.

### Reproduction Sample

https://mikehorn.dev/calcite-input-saved-login/

### Reproduction Steps

1. Enter whatever email/password you want.
2. Submit.
3. Instruct browser to save email/password.
4. Return to login.
5. Refresh.
6. Values are autofilled, but hidden from the user.

### Reproduction Version

beta.86

jcfranco:
@gpbmike Thanks for reporting this. Just a quick update, the team is looking into this. We have a potential fix that could ship soon. Do you think the Site Scan team could help us do some additional browser testing w `@next` once it's installed?

gpbmike:
> @gpbmike Thanks for reporting this. Just a quick update, the team is looking into this. We have a potential fix that could ship soon. Do you think the Site Scan team could help us do some additional browser testing w `@next` once it's installed?

Yep. We can help test.

gpbmike:
Thanks all

Issue #5026, opened 2022-07-26
calcite-input: input loses focus on every keypress when login info saved (safari)


gpbmike:
### Actual Behavior

Using Safari, If you use calcite-input in login form and choose to save email/password, when you return the inputs will lose focus on every keypress.

https://user-images.githubusercontent.com/8754/181073838-52e50e9a-0b9f-4a98-bae3-8115ba5372ec.mov

### Expected Behavior

User should be able to freely type email and password without having to re-focus on the field.

### Reproduction Sample

https://mikehorn.dev/calcite-input-saved-login/

### Reproduction Steps

1. Use Safari.
2. Enter any email/password.
3. Save login info to browser.
4. Return to form.
5. Try to type email and password.
6. Every keypress causes input to blur.

### Reproduction Version

beta.86

driskull:
This one is tricky. Safari is trying to apply the saved form info to the hidden input elements instead of the inputs in the shadowRoot.

So the positioning of the autofill is all incorrect and causing the component to become unfocused.

We could just disable autoFill for these hidden inputs?

benelan:
I guess we could do that, but it would be sad to completely disable autofill for inputs just because of Safari. It works in Chrome and Firefox. IIRC @eriklharper you looked into this right, do you remember what you found?

I remember I suggested making the hidden inputs [`inert`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert), but it wasn't supported in the latest two Safari versions at the time. Not sure if that worked or not though, I don't have a Mac.

driskull:
I'll try some more options 👍

eriklharper:
Yeah, the `inert` method was the only way I was able to get a "fix" in for this: https://github.com/Esri/calcite-components/pull/5221/files  The other workaround is create a `scoped` input component, but that seems like a lot just to accommodate autofill on Safari.

driskull:
`inert` doesn't do the autofill in safari either. If I set inert on the input it no longer gets autofilled.

The problem is that safari focuses on the hidden input on autofill which means the component's input loses focus and it gets into that kind of a loop.

geospatialem:
Reallocating to February's priorities for the team to explore further.

anveshmekala:
This bug is fixed with PR #5221. Tested in Safari 16.4 Version.

jcfranco:
[...]

PR #5149, opened 2022-08-13
fix(input): display autofilled username/password input values

benelan:
**Related Issue:** #5022

## Summary

Fixes an issue where login values were autofilled in the form's hidden inputs but not visible in `calcite-input`. A couple notes:

1. ~~I didn't remove the change listener from the hidden inputs. Based on the code it looks like the hidden inputs are removed from the DOM when they are no longer in sync with a form component's value ([here](https://github.com/Esri/calcite-components/blob/master/src/utils/form.tsx#L242) and [here](https://github.com/Esri/calcite-components/blob/master/src/utils/form.tsx#L291)). My understanding is that removing elements from the DOM removes their event listeners, but I can also do it manually if that makes everyone feel safer.~~ Nevermind, I went ahead and removed them
2. Not sure if its possible to add an e2e test for this one. Logins don't autofill w/o https, and even if they did I don't think we can save form entries in our tests. But if anyone has an idea I'd be happy to give it a shot.
3. On Firefox the saved login autofills on page load. 
4. On Chrome the autofill won't happen until the user interacts with the page (e.g. click anywhere). This is for security and I'm not aware of any way around it. [Here is the chromium issue marked as "Wont fix"](https://bugs.chromium.org/p/chromium/issues/detail?id=669724#c6)
5. Can someone please test this on Safari? Steps below

## Testing
I created [a repo](https://github.com/benelan/calcite-components-https-testing) with some unsigned certs to test this since it needs HTTPS. Follow these steps:
1. Run this in terminal
```bash
git clone git@github.com:benelan/calcite-components-https-testing.git && cd calcite-components-https-testing && npm i && npm run build && npm start
```
2. Open your browser to [https://localhost:8443/](https://localhost:8443/)
3. Follow the instructions on the page (don't put in your real password, it's logged to the console for testing)


gpbmike:
# Safari

Autofilled fields show as having a value now

https://user-images.githubusercontent.com/8754/184658332-4eb2e811-e968-4759-b2d0-ebeea8c01880.mov

But using the input is very frustrating. Focus is moved to the hidden field. In the video, I keep typing after losing focus. When I blur the hidden field, the typed value moves to the visible input. Related issue: https://github.com/Esri/calcite-components/issues/5026

https://user-images.githubusercontent.com/8754/184658347-7d39c1db-0b26-474e-92e9-62975648d063.mov

benelan:
Thanks for testing @gpbmike. Unfortunately I don't have a Mac so someone else will probably need to work on #5026. I wasn't able to repro on Firefox or Chrome. I'll see who has bandwidth.

src/utils/form.tsx
```diff
@@ -225,6 +225,15 @@ export function afterConnectDefaultValueSet<T>(component: FormComponent<T>, valu
   component.defaultValue = value;
 }
 
+const hiddenInputChangeHandler = (event: Event) => {
+  event.target.dispatchEvent(
+    new CustomEvent("calciteInternalHiddenInputChange", {
+      bubbles: true,
+      composed: true
```
jcfranco, line 232: It _think_ we don't need the event to be `composed` since it's emitted and listened to on the light DOM.

src/utils/form.tsx
```diff
@@ -225,6 +225,15 @@ export function afterConnectDefaultValueSet<T>(component: FormComponent<T>, valu
...
+    })
```
jcfranco, line 233: Can you stop propagation on the internal event?

src/utils/form.tsx
```diff
@@ -282,13 +294,19 @@ function syncHiddenFormInput(component: FormComponent): void {
...
+
     defaultSyncHiddenFormInput(component, input, value);
   });
 
   if (docFrag) {
     el.append(docFrag);
   }
-  extra.forEach((input) => input.remove());
+  extra.forEach((input) => {
+    input.removeEventListener("change", hiddenInputChangeHandler);
```
jcfranco, line 307: Nitpick: event removal code can be extracted into a separate method.

src/components/input/input.tsx
```diff
@@ -720,6 +722,23 @@ export class Input implements LabelableComponent, FormComponent, InteractiveComp
     }
   }
 
+  hiddenInputChangeHandler = (event: Event): void => {
+    if ((event.target as HTMLInputElement).name === this.name) {
```
jcfranco, line 726: Code for this should be moved into a util module to make it easier for components to apply. This can be done separately.

jcfranco (approved):
👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍
👍🎉👍👍👍🎉👍👍🎉🎉👍👍🎉🎉🎉👍👍🎉🎉🎉👍🎉🎉🎉🎉👍🎉👍
👍🎉🎉👍👍🎉👍🎉👍👍🎉👍👍🎉👍👍🎉👍👍👍👍🎉👍👍👍👍🎉👍
👍🎉👍🎉👍🎉👍🎉👍👍🎉👍👍🎉👍👍🎉👍👍👍👍🎉🎉🎉👍👍🎉👍
👍🎉👍👍🎉🎉👍🎉👍👍🎉👍👍🎉👍👍🎉👍👍👍👍🎉👍👍👍👍👍👍
👍🎉👍👍👍🎉👍👍🎉🎉👍👍🎉🎉🎉👍👍🎉🎉🎉👍🎉🎉🎉🎉👍🎉👍
👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍

src/components/input/input.tsx
benelan: Created a new issue #5171

eriklharper (approved):
This looks good to me.

One question/concern I had about Franco's idea to move `hiddenInputChangeHandler` functionality to a separate util is that in order for that to work with each `FormComponent`, the setValue pattern would need to be universal and currently there are some differences with how the value is set across them, for instance input-number uses `setNumberValue`.  I'll provide more thoughts on this idea in the related issue #5171

benelan:
I was thinking about creating an InputComponent class that could be extended and then each input component could have their own `setValue` functionality. We would just need to name them all the same thing.

eriklharper:
> I was thinking about creating an `InputComponent` class that could be extended and then each input component could have their own `setValue` functionality. We would just need to name them all the same thing.

This will be cool to explore.

Issue #5127, opened 2022-08-10
Regression with `calciteStepperItemChange` event


nwhittaker:
### Actual Behavior

Clicking anywhere within a step item's content causes the stepper to emit its [`calciteStepperItemChange`][1] event.

[1]: https://developers.arcgis.com/calcite-design-system/components/stepper/#component-api-events-calciteStepperItemChange

### Expected Behavior

Clicking anywhere within a step item's content area does not emit the `calciteStepperItemChange` event.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/OJvwaaK

### Reproduction Steps

1. Visit the code sample and open the **Console**
2. Click a few times on the `"Step 3 Content here dolor set amet et geographica"` text
3. Notice `"calciteStepperItemChange"` is logged to the console for each click

### Reproduction Version

next.557

### Relevant Info

Possibly introduced by https://github.com/Esri/calcite-components/pull/4367.

### Regression?

beta.86

geospatialem:
Verified on`next.561`

PR #5150, opened 2022-08-13
fix(stepper-item): avoid emitting change event when clicking content in horizontal layout

jcfranco:
**Related Issue:** #5127 

## Summary

When the stepper is in horizontal layout, the content is displaced from the item's selection hit area, so it needs to be ignored when clicked.


driskull (approved):
🆒 😎

PR #5153, opened 2022-08-15
fix(radio-button-group): fix incorrect event payload type

jcfranco:
**Related Issue:** N/A

## Summary

This fixes a type regression introduced by https://github.com/Esri/calcite-components/pull/5001/files#diff-1db34e36763c50cd41364b15a24af2d25c23a0d61dc2a53e31f570b061a9c17eR128 where `calciteRadioButtonGroupChange`'s payload was typed as `void` instead of `any`.

cc @AdelheidF @paulcpederson

### DIFF
--- a/src/components/radio-button-group/radio-button-group.tsx
+++ b/src/components/radio-button-group/radio-button-group.tsx
@@ -125,7 +125,7 @@ export class RadioButtonGroup {
   /**
    * Fires when the component has changed.
    */
-  @Event({ cancelable: false }) calciteRadioButtonGroupChange: EventEmitter<void>;
+  @Event({ cancelable: false }) calciteRadioButtonGroupChange: EventEmitter<any>;
 
   //--------------------------------------------------------------------------
   //


driskull (approved):
👍

src/components/radio-button-group/radio-button-group.tsx
```diff
@@ -125,7 +125,7 @@ export class RadioButtonGroup {
   /**
    * Fires when the component has changed.
    */
-  @Event({ cancelable: false }) calciteRadioButtonGroupChange: EventEmitter<void>;
+  @Event({ cancelable: false }) calciteRadioButtonGroupChange: EventEmitter<any>;
```
paulcpederson, line 128: Pretty sure this is a string, right?

jcfranco: Most commonly a string, but `value` allows for anything. This is useful for environments that set props to rich data to skip mapping a string to an object or something else.

paulcpederson: oh interesting! I assumed this corresponded to the radio input value property

Issue #5102, opened 2022-08-05
Dropdown placement/positioning wrong


daniellaumer:
### Actual Behavior

When placing a dropdown on the bottom of a page, sometimes the items are places below the button and therefore not visible/accessible. This only happens for some window sizes and can be best observed when changing the window height: 

![BugCalciteDropdown2](https://user-images.githubusercontent.com/93542103/183086331-c9d8c3e8-1da2-4537-9b21-0c112752a8b5.gif)

### Expected Behavior

The dropdown should be displayed on top of the button.

### Reproduction Sample

https://codepen.io/dlaumer/pen/oNqdLZm

### Reproduction Steps

1. Open the codepen example
2. Resize the height of the window (see screen gif)
3. Observe how the item list of the dropdown disappears below the button

### Reproduction Version

beta.86

### Relevant Info

Win10, Chrome

Maybe related to this issue: https://github.com/Esri/calcite-components/issues/4272

benelan:
This works as expected if you set `placement` to "top".

```html
 <calcite-dropdown id="exampleDropdown" open placement="top">
```

benelan:
Reopening, @daniellaumer reached out and provided another repro sample:
https://codepen.io/benesri/pen/VwXVXor?editors=1000

It is inconsistent, but open the dropdown and it is above the trigger, refresh and try again and it will eventually be below the trigger. I wasn't able to repro in `beta.86` so it may be due to the switch from popper to floating-ui

benelan:
verified on beta.94 
https://codepen.io/benesri/pen/VwXVXor?editors=0010

PR #5169, opened 2022-08-16
fix(dropdown): Correct placement when opening

driskull:
**Related Issue:** #5102

## Summary

fix(dropdown): Correct placement when opening. (#5102)

The dropdown group needs to remain visible so floating UI can calculate its height. Instead, we set the width/height on its container to be 0 when not opened. This prevents the dropdown group from taking up space when not opened.

cc @benelan

### DIFF
--- a/src/assets/styles/_floating-ui.scss
+++ b/src/assets/styles/_floating-ui.scss
@@ -86,11 +86,17 @@ $floating-ui-default-z-index: 900;
 @mixin floatingUIWrapper {
   visibility: hidden;
   pointer-events: none;
+  inline-size: 0;
+  block-size: 0;
+  overflow: hidden;
 }
 
 @mixin floatingUIWrapperActive {
   pointer-events: initial;
   visibility: visible;
+  inline-size: unset;
+  block-size: unset;
+  overflow: unset;
 }
 
 @mixin floatingUIHost($zIndex: $floating-ui-default-z-index) {
--- a/src/components/dropdown/dropdown.tsx
+++ b/src/components/dropdown/dropdown.tsx
@@ -236,9 +236,7 @@ export class Dropdown implements InteractiveComponent, OpenCloseComponent, Float
             ref={this.setScrollerAndTransitionEl}
             role="menu"
           >
-            <div hidden={!open}>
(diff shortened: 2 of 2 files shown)


driskull:
@benelan can you test this with that sample? I tested it and it seemed to fix the problem.

benelan:
It's definetly a lot better. I had to resize and refresh for a good minute before I was able to reproduce the issue
image

And then after refreshing again (and not resizing) the issue goes away. Super weird

driskull:
@benelan In that case, can you call `reposition()` when it is messed up?

benelan:
Yeah that fixes it 
image

driskull:
I think we should proceed.

benelan:
> I think we should proceed.

Yeah I agree, it definitely helped a lot. It repros way less often now. It's not ideal, but positioning it w/ CSS fixes the issue since it only repros when using the JSAPI's View UI for placement.

```css
calcite-dropdown {
  position: fixed;
  z-index: 1;
  bottom: 2rem;
  right: 1rem;
} 
```

As you suggested, it seems like this fixes it too.
```js
document.addEventListener("calciteDropdownBeforeOpen", ({ target }) =>
  target.reposition()
);
```

jcfranco (approved):
💥

jcfranco:
@driskull WDYT about introducing a prop for floating-UI-owning components to opt into the [autoUpdate](https://floating-ui.com/docs/autoUpdate) behavior? It could help eliminate boilerplate code across apps for these types of scenarios.

driskull:
@jcfranco we do use that for every floatingUI component already. Here: https://github.com/Esri/calcite-components/blob/master/src/utils/floating-ui.ts#L381

driskull:
Maybe if I get rid of the `component.open` requirement it might work better?

jcfranco:
Ah, my bad. I thought we weren't using since the use case above still needed to use `reposition`.

> Maybe if I get rid of the component.open requirement it might work better?

Worth exploring, IMO.

driskull:
@benelan can you test this PR out? https://github.com/Esri/calcite-components/pull/5192

Issue #5137, opened 2022-08-11
No longer able to transition scrim from `hidden` state


nwhittaker:
### Actual Behavior

Given `<calcite-scrim>` with transition styles, toggling its `hidden` attribute shows/hides the scrim abruptly.

### Expected Behavior

Given `<calcite-scrim>` with transition styles, toggling its `hidden` attribute shows/hides the scrim in accordance to its transition styles.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/zYWJNJz

### Reproduction Steps

1. Visit the code sample.
2. Click the **Toggle scrim** button a few times to show/hide the scrim.

The scrim should take `3s` to show/hide itself, instead of immediately.

### Reproduction Version

beta.91

### Relevant Info

Appears to have been introduced by https://github.com/Esri/calcite-components/pull/5059 which changed `:host([hidden])` to force its display style:

https://github.com/Esri/calcite-components/blob/eed9168bea13b2198d847e4682d0fc59cb9d4eb3/src/assets/styles/includes.scss#L18-L20

I'm open to suggestions on alternative ways to transition the scrim, but, IMO, making this style `!important` breaks consumer's assumption about how Calcite hides elements and is not aligned with how native HTML elements use it. From [the doc][1]:

>Note: Changing the value of the CSS [`display`][2] property on an element with the `hidden` attribute overrides the behavior. For instance, elements styled `display: flex` will be displayed despite the `hidden` attribute's presence.

[1]: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden
[2]: https://developer.mozilla.org/en-US/docs/Web/CSS/display

### Regression?

beta.90

jcfranco:
You bring up a good point. When we made the change, we were focused on ensuring that `hidden` worked consistently across components, but ended up breaking the native HTML element expectation you mentioned. We'll revisit. Sorry for the hassle! 

@Elijbet Let's revert this and work on preventing components from overriding the global `hidden` style instead. Maybe https://stylelint.io/user-guide/rules/list/selector-disallowed-list would be a good way to accomplish this.

jcfranco:
@benelan triaged for this release.

geospatialem:
Verified in last night's release, beta.92.

![verify scrim](https://user-images.githubusercontent.com/5023024/185416310-f4b1ee4e-909a-4887-a63a-3b464e7fa70c.gif)

PR #5174, opened 2022-08-17
fix(scrim): reverts change to hidden attribute that caused a transition regression on scrim

Elijbet:
**Related Issue:** #5137

## Summary

Reverts change to `hidden` attribute that caused a transition regression on `scrim`.


jcfranco (approved):
💥

driskull (approved):
💥

Issue #5136, opened 2022-08-11
date-picker: the weekday and date are not correlated in some locales (including en-gb, en-au, nb, es, de ...)


fendiaoxiaoshuazi:
### Actual Behavior

The weekday and date are not actually correlated.

### Expected Behavior

The weekday and date should be correlated. As you can see in the screenshot below, August 11 is a Thursday, but in some locales, August 11 appears below Saturday.

**en-gb**
![image](https://user-images.githubusercontent.com/26078106/184159834-0776ac29-8bb7-4c3c-9a5c-936421fa093c.png)

**en-au**
![image](https://user-images.githubusercontent.com/26078106/184159850-db8a1910-a96d-489b-a04c-48e7f6811bc3.png)

**nb**
![image](https://user-images.githubusercontent.com/26078106/184159949-5d48ab75-6a39-4c11-99b6-03ea723636af.png)

**es**
![image](https://user-images.githubusercontent.com/26078106/184159967-36ad38ee-8304-4377-83c9-79cad06f3233.png)

**de**
![image](https://user-images.githubusercontent.com/26078106/184159989-10ed7293-823b-4b26-83f4-53914d9ed66f.png)

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/date-picker/

### Reproduction Steps

1. Open the website
2. Change the locale to the locale listed above
3. Observe the weekday and date

### Reproduction Version

beta.86

### Regression?

beta.81

geospatialem:
Thanks for bringing this forth, @fendiaoxiaoshuazi! 

Adding additional context that this seems to occur in August 2022 and May 2023, but no other month in 2022/2023.

An additional Codepen for context: https://codepen.io/geospatialem/pen/dymqNVd.

fendiaoxiaoshuazi:
Hi @geospatialem, will this fix be included in the next release? If possible could you please let me know the general release date of the next build? We will schedule a hotfix to upgrade to the new build.

geospatialem:
> Hi @geospatialem, will this fix be included in the next release? If possible could you please let me know the general release date of the next build? We will schedule a hotfix to upgrade to the new build.

@fendiaoxiaoshuazi We're working through the fix this afternoon, and will keep you apprised as soon as its merged in. Hoping to have a fix via `next` very soon. 👍🏻

geospatialem:
[...]

PR #5175, opened 2022-08-17
fix(date-picker): no longer misalign dates with day in `en-gb`,`en-au`,`nb`,`es`,`de` locales

anveshmekala:
**Related Issue:** #5136 

Fixes misalignemnt of days with dates in `en-gb`,`en-au`,`nb`,`es`,`de` locales

Side bar:

`Math.abs` is not required to make sure if the indexed value of startofWeek is  greater than lastDay of the previous month we don't need to display any dates from previous in current month

Example : startOfWeek = 1 (Monday) and lastDay of previous month is 0 ( sunday)

### DIFF
--- a/src/components/date-picker-month/date-picker-month.tsx
+++ b/src/components/date-picker-month/date-picker-month.tsx
@@ -272,7 +272,8 @@ export class DatePickerMonth {
     if (day - 6 === startOfWeek) {
       return days;
     }
-    for (let i = Math.abs(lastDate.getDay() - startOfWeek); i >= 0; i--) {
+    console.log(lastDate, lastDate.getDay(), startOfWeek);
+    for (let i = lastDate.getDay() - startOfWeek; i >= 0; i--) {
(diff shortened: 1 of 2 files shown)


src/components/date-picker-month/date-picker-month.tsx
```diff
@@ -272,7 +272,7 @@ export class DatePickerMonth {
     if (day - 6 === startOfWeek) {
       return days;
     }
-    for (let i = Math.abs(lastDate.getDay() - startOfWeek); i >= 0; i--) {
```
driskull, line 275: Can you add a comment in the PR description about why this Math.abs is not needed?

driskull (approved):
👍 Looks good! 💯

src/components/date-picker-month/date-picker-month.tsx
anveshmekala: Sure.

PR #5180, opened 2022-08-18
fix: wait for body to load before initializing theme change event

benelan:
**Related Issue:** #5104

## Summary
@jcfranco noticed an error in the `demo-app-advanced-2-shell-header.html` and `demo-app-advanced.html` caused by trying to access `body`'s `classList` before it was available.

image

### DIFF
--- a/src/utils/globalScript.ts
+++ b/src/utils/globalScript.ts
@@ -7,7 +7,7 @@ import { initThemeChangeEvent } from "./theme";
  */
 export default function (): void {
   if (isBrowser()) {
-    initThemeChangeEvent();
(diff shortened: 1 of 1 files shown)


Elijbet (approved):

src/utils/globalScript.ts
```diff
@@ -7,7 +7,7 @@ import { initThemeChangeEvent } from "./theme";
  */
 export default function (): void {
   if (isBrowser()) {
-    initThemeChangeEvent();
+    document.addEventListener("DOMContentLoaded", () => initThemeChangeEvent());
```
jcfranco, line 10: Do you need to check if the content has already been loaded and therefore the event already fired?

PR #5184, opened 2022-08-18
fix: initialize calciteThemeChange when DOM content is loaded before globalScript executes

benelan:
**Related Issue:** #5180

## Summary
The fix in the PR linked above could cause `calciteThemeChange` to not fire if the DOM content is loaded before globalScript invokes.

Edit:
Switched to `readystate==="interactive"` because `readystate==="complete"` comes after `DOMContentLoaded` so there would still be a small window of time where the theme event initialization won't happen. See:
https://codepen.io/benesri/pen/xxWmMBM?editors=0012

### DIFF
--- a/src/utils/globalScript.ts
+++ b/src/utils/globalScript.ts
@@ -7,7 +7,7 @@ import { initThemeChangeEvent } from "./theme";
  */
 export default function (): void {
   if (isBrowser()) {
-    initThemeChangeEvent();
+    document.addEventListener("DOMContentLoaded", () => initThemeChangeEvent());
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/utils/globalScript.ts
```diff
@@ -7,7 +7,11 @@ import { initThemeChangeEvent } from "./theme";
  */
 export default function (): void {
   if (isBrowser()) {
-    document.addEventListener("DOMContentLoaded", () => initThemeChangeEvent());
+    if (document.readyState === "interactive") {
+      initThemeChangeEvent();
+    } else {
+      document.addEventListener("DOMContentLoaded", () => initThemeChangeEvent());
```
jcfranco, line 13: Can you use the `once` argument so the listener is removed after the event is handled?

jcfranco (approved):
✨🔧✨

Issue #5182, opened 2022-08-18
calcite-popover not re-registering itself with PopoverManager during connectedCallback method


rweber-esri:
### Actual Behavior

`calcite-popover` does not work as expected after re-appending it to the DOM

### Expected Behavior

`calcite-popover` works as expected after removing from DOM, then re-appending to DOM.

### Reproduction Sample

https://codepen.io/rweber-esri/pen/QWmzgxZ

### Reproduction Steps

1. Add a `calcite-action` and obtain a reference to it
2. Add a `calcite-popover` and pass reference to calcite-action as `referenceElement` prop
3. Append the `calcite-action` and `calcite-popover` to a `div`
4. Append the `div` to `body`
5. Observe popover works as expected
6. After some delay, remove the `div` from the DOM (also removes `calcite-action` and `calcite-popover` from DOM)
7. Re-append the `div` to the DOM 
8. Observe popover does not work as expected

### Reproduction Version

beta.92

### Relevant Info

Per the [Stencil connectedCallback docs](https://stenciljs.com/docs/component-lifecycle#connectedcallback):

> Called every time the component is connected to the DOM. When the component is first connected, this method is called before componentWillLoad.

The component's [disconnectedCallback method](https://github.com/Esri/calcite-components/blob/b184ce3cf6d71e5612e9426544365ef2ff21aaab/src/components/popover/popover.tsx#L260) unregisters the popover instance from the [PopoverManager](https://github.com/Esri/calcite-components/blob/b184ce3cf6d71e5612e9426544365ef2ff21aaab/src/components/popover/PopoverManager.ts) when the component instance is removed from the DOM. However, the component does not re-register itself with [PopoverManager](https://github.com/Esri/calcite-components/blob/b184ce3cf6d71e5612e9426544365ef2ff21aaab/src/components/popover/PopoverManager.ts) upon being reinserted into the DOM, which prevents the popover from functioning properly.

I believe the fix to entail conditionally calling [setUpReferenceElement](https://github.com/Esri/calcite-components/blob/b184ce3cf6d71e5612e9426544365ef2ff21aaab/src/components/popover/popover.tsx#L366) when `this.element` is truthy from [connectedCallback](https://github.com/Esri/calcite-components/blob/b184ce3cf6d71e5612e9426544365ef2ff21aaab/src/components/popover/popover.tsx#L239).

driskull:
NIce catch.

geospatialem:
Verified on next
[...]

PR #5185, opened 2022-08-18
fix(tooltip, popover): reconnecting the component to the DOM should work

driskull:
## Summary

fix(tooltip, popover): reconnecting the component to the DOM should work. (#5182)


jcfranco (approved):
Code LGTM!

Can you check that https://github.com/Esri/calcite-components/pull/2725 is still covered? 

Sidebar: we should add a test for ☝️ to avoid manual checks in the future.

jcfranco (approved):
💥

PR #5187, opened 2022-08-18
fix(modal): calciteModalOpen and calciteModalBeforeOpen now emit on page load

Elijbet:
**Related Issue:** #4689

## Summary

`calciteModalOpen` and `calciteModalBeforeOpen` would not emit on page load like the rest of the components and only when you close and reopen the modal. 

This change ensures that the open styles for the underlay get applied initially and the modal transition styles are applied once the component is fully hydrated.


jcfranco:
@Elijbet I tweaked the approach since I remembered we can't rely on `[calcite-hydrated]` for the `dist-custom-elements` bundle since the attribute isn't applied there. I also added a test for this particular use case, but testing should be moved to a common test helper for all open/close components. I'll create an issue for this.

driskull (approved):
👍

Issue #4880, opened 2022-07-08
[Dropdown] Unwanted scrolling when max-items >= number of dropdown items


macandcheese:
### Actual Behavior

When `max-items` is used on `calcite-dropdown`, and the value equals or exceeds the number of `calcite-dropdown-items`, a small bit of unwanted scrolling appears in the dropdown container.

### Expected Behavior

There is no extra scrolling when the number of dropdown items is equal to or less than the `max-items` value

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/WNzwzwa?editors=1000

### Reproduction Steps

1 - Go to codepen
2 - See the extra scrolling in the two examples with max-items set

### Reproduction Version

83

geospatialem:
The scrollbar is fixed with smaller `max-item` numbers, but when `max-items` is set to 8 or greater the scroller seems to still be added in. 

It looks like this is a regression, last working in `beta.74`. @driskull, can you take a look? 

- [`next`](https://codepen.io/geospatialem/pen/zYjrgeY) - scrollbar present when it shouldn't be
- [`beta.74`](https://codepen.io/geospatialem/pen/PoeZMxO) - expected behavior

driskull:
@geospatialem the scrollbar is still dependent on the window size as well. That might be the reason for it. It has a max height of `45vh`.

geospatialem:
Thanks @driskull for the background! Was able to get the list without the scrollbar when opening in a larger window. Marking this as verified. ✔️

PR #5188, opened 2022-08-18
fix(dropdown): Remove scrolling when max-items >= number of dropdown items

driskull:
**Related Issue:** #4880

## Summary

fix(dropdown): Remove scrolling when max-items >= number of dropdown items. (#4880)


jcfranco (approved):
💥

Issue #5078, opened 2022-08-03
Make reproduction bot smarter


benmccann:
### Summary

I filed an https://github.com/Esri/calcite-components/issues/5077 and had the "incomplete issue report" label applied eventhough I provided a reproduction

### Desired Outcome

Github reproductions should be supported. codepen, codesandbox, and jsbin cannot be used to reproduce Node.js issues. stackblitz could also be added as a supporter reproduction platform

benelan:
Thanks for the feedback! To provide some background, [here is the bot](https://github.com/marketplace/actions/need-more-info) I created to check for samples. It works by checking for text specified in [a config](https://github.com/Esri/calcite-components/blob/master/.github/need-info.yml). I'll find a way to match GitHub URLs as samples without getting false positives when linking to other issues or PRs. Could probably update the bot to use regex.

geospatialem:
Additional effort is needed to improve the bot per https://github.com/Esri/calcite-components/pull/5190#issue-1344728781. Research should be conducted to execute next steps for improvement.

benelan:
I think #5190 and the change to a form issue template with the sample field marked as required is sufficient here.

PR #5190, opened 2022-08-19
ci: allow github repos as repro samples

benelan:
**Related Issue:** #5078

## Summary
This will make the bot allow GitHub repos as reproduction samples. We were getting a lot of false positives asking people to provide a sample when they already did. 

Note that this is only a temporary solution. If someone links to a related GitHub issue and doesn't provide a sample then the bot will not ask for more info. I will eventually fix the bot, but for now I think adding the label manually is better than alienating users who provide repro repos.


driskull (approved):
👍 nice

PR #5192, opened 2022-08-19
fix: Reposition components regardless of open state

driskull:
**Related Issue:** N/A

## Summary

fix: Reposition components regardless of open state


driskull:
The only downside I could see to this is performance. It will reposition any non opened components during a page scroll.

Issue #3772, opened 2021-12-22
[Tree Item] Add support for disabled tree-item


ptmdmusique:
### Question

Is there any way to disable a single `tree-item` on a list, e.g.: make it not focusable/clickable 

I tried with `pointer-events: none` and it works, but it seems like if the `tree-item` is in the first layer of the tree, it's still focusable
Also, notice if a disabled `tree-item` is clicked first, the tree automatically select the first item in the list

https://codepen.io/ptmdmusique/pen/poWrxvJ

I can put `pointer-events: none` to the whole tree, but that wouldn't be ideal in a lot of cases

macandcheese:
Currently this isn't possible - a valid enhancement request though. Updating title.

geospatialem:
Verified on `next.597`

![tree item disabled verification](https://user-images.githubusercontent.com/5023024/196204144-43f8cb9c-6cc9-4027-bdf5-4a6d7234bf52.png)

PR #5194, opened 2022-08-20
feat(tree-item): allow disabling tree-item

jcfranco:
**Related Issue:** #3772 

## Summary

✨🌳🚫👈✨


jcfranco:
@paulcpederson @macandcheese Could you give this a test drive and see if it's merge-worthy? 🧽

macandcheese:
Noticed some weird focus behavior - seems possible to click to focus the first item only before the tree item sibling that is active is expanded. Once that is expanded, clicking the first disabled item no longer focuses it. It remains possible to keyboard focus the first item. The nested disabled item doesn't seem to be focusable at all.

Not sure which of the interactions makes the most sense from an a11y perspective but it seems a little inconsistent:

https://user-images.githubusercontent.com/4733155/186312416-14c3e095-30c8-494b-bd35-7dedbbab0ef3.mov

jcfranco:
@macandcheese Fixed the issue you called out. Can you give it another try?

Focus styling was being applied regardless of component being disabled. We'll need to make sure any interactive component considers the `disabled` attribute for host styles using `:focus` since elements with `tabindex=-1` can still be focused via mouse click or `focus()`. cc @benelan @geospatialem

src/components/tree/tree.e2e.ts
```diff
@@ -145,24 +143,58 @@ describe("calcite-tree", () => {
     expect(grandchildOne).not.toHaveAttribute("selected");
     expect(grandchildTwo).not.toHaveAttribute("selected");
 
-    await two.click();
+    // Puppeteer's element click will happen in the center of a component,
+    // so we call the method to ensure it happens on the component of interest
+
+    await two.callMethod("click");
```
eriklharper, line 149: So cool, had no idea you could do this!

src/components/tree/tree.tsx
```diff
@@ -246,94 +252,143 @@ export class Tree {
...
+      }
+
+      // When focus is on a root node that is also either an end node or a closed node, does nothing.
+      return;
+    }
+
+    if (event.key === "ArrowRight" && !target.disabled) {
+      if (target.hasChildren) {
+        if (target.expanded && getRootNode(this.el).activeElement === target) {
+          {
```
benelan, line 305: Just out of curiosity, what's the purpose of all these extra curly brackets

benelan (approved):
There may be extra curley brackets in the ` keyDownHandler` `event.key` conditionals, but if they have a purpose then the code LGTM

src/components/tree/tree.tsx
jcfranco: To break the interwebs! 💻🔨

J/K, good catch. I'll remove these unnecessary extra blocks.

Issue #5191, opened 2022-08-19
[Combobox] Combobox now requires two tabs to navigate through the component.


macandcheese:
### Actual Behavior

Combobox now requires two tabs to navigate through the component:
The first tab focuses the component but not the input. The second tab correctly focuses the input.

Version 91: https://codepen.io/mac_and_cheese/pen/ExErbYd?editors=100

### Expected Behavior

A single tab will focus the component and the internal input is immediately actionable:

Version 86: https://codepen.io/mac_and_cheese/pen/yLKZPBp?editors=100

### Reproduction Sample

See above

### Reproduction Steps

View two above Codepens - see the correct behavior on version 86, and incorrect on version 91.

### Reproduction Version

91

### Regression?

86

driskull:
Looks like a regression from https://github.com/Esri/calcite-components/commit/1798571cc7e3185c35f68aaf7f63d9377b3bad31 cc @geospatialem

anveshmekala:
we may need to remove the `tabIndex` to get this fixed.

geospatialem:
> Looks like a regression from [1798571](https://github.com/Esri/calcite-components/commit/1798571cc7e3185c35f68aaf7f63d9377b3bad31)

That it is. Thanks for diagnosing, @driskull!

> we may need to remove the `tabIndex` to get this fixed.

@anveshmekala Removing the `tabIndex` will work for this fix. Thanks for tackling so fast!

jcfranco:
Removed `needs triage` label and assigned to this milestone based on the assignment, status label and linked PR.

geospatialem:
Verified on beta.93

PR #5198, opened 2022-08-22
fix(combobox): focuses input when tab through the component

anveshmekala:
**Related Issue:** #5191

### DIFF
--- a/src/components/combobox/combobox.e2e.ts
+++ b/src/components/combobox/combobox.e2e.ts
@@ -1,3 +1,4 @@
+/* eslint-disable jest/no-focused-tests */
 import { E2EPage, E2EElement, newE2EPage } from "@stencil/core/testing";
 import {
   renders,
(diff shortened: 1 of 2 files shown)


jcfranco (approved):
LGTM! Can you add an E2E test that covers the keyboard usage? A simple use case will suffice for this.

jcfranco:
> Can you add an E2E test that covers the keyboard usage? A simple use case will suffice for this.

Can you submit a follow-up PR for this? It'd be good to include this for the upcoming patch release.

jcfranco:
Test failures are unrelated. Merging!

anveshmekala:
> LGTM! Can you add an E2E test that covers the keyboard usage? A simple use case will suffice for this.

I tried adding an e2e test for this use case to assert `activeElement` in the document which should be the `input`.  DOM recognizes `combobox` as the activeElement rather than `input` inside shadow root.

PR #5209, opened 2022-08-23
ci(release): fix Windows errors

benelan:
**Related Issue:** NA

## Summary
Windows doesn't like `echo`


support/prepReleaseCommit.ts
```diff
@@ -84,14 +79,14 @@ import yargs from "yargs";
     const target = next ? "next" : "beta";
     const targetVersionPattern = new RegExp(`-${target}\\.\\d+$`);
 
-    await exec(`echo ${semverTags}`);
+    console.log(semverTags);
```
jcfranco, line 82: These look like debug logs. Let's remove them if this is the case.

jcfranco (approved):
👍

Issue #5131, opened 2022-08-10
calcite-checkbox needs to hide the SVG element from screen readers


rpanichakit9541:
### Summary

From an accessibility audit, a recommendation was made to add `aria-hidden="true"` `focusable="false"` and `role="none"` to the `<svg class="check-svg">` element inside `calcite-checkbox`

### Actual Behavior

`<svg class="check-svg">` is not hidden from screen readers

### Expected Behavior

`<svg class="check-svg">` should be hidden from screen readers

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/checkbox/

### Reproduction Steps

1. Use `<calcite-checkbox>` and inspect the SVG element

### Reproduction Version

beta.86

jcfranco:
@rpanichakit9541 Thanks for reporting this. Can you share more info about the audit? Is it affecting a particular screen reader + browser combination? cc @geospatialem

rpanichakit9541:
@jcfranco - I'm not sure about the screen reader + browser combination.  The issue is filed here https://devtopia.esri.com/webgis/arcgis-portal-app/issues/39881 .

I just saw the "Note" recommendation and the screenshot below, so I filed the issue accordingly

## Note
```
[Recommendation]
Provide alternative text for images. Non-meaningful images must be marked as decorative, which causes screen readers to ignore them. To mark an <svg> element as decorative, add aria-hidden="true", focusable="false" and role="none" to the element.

[Compliant Code Example]
<svg aria-hidden="true" focusable="false" role="none" class="check-svg" viewbox="0 0 16 16"><path d=""></path></svg>
```
![calcite-checkbox-svg](https://user-images.githubusercontent.com/44479773/186280533-cb8a330c-a122-4111-bff2-c98529cd56ae.jpg)

jcfranco:
@rpanichakit9541 Awesome, thanks!

geospatialem:
Verified in `beta.95`

![Screen Shot 2022-10-04 at 12 07 19 PM](https://user-images.githubusercontent.com/5023024/193882700-7cfb7160-a4e2-4014-8aca-02366ffc4b4a.png)

PR #5211, opened 2022-08-23
fix(checkbox): hide check icon from screen readers

jcfranco:
**Related Issue:** #5131 

## Summary

Updates check SVG to hide the check icon from screen readers. cc @geospatialem


src/components/checkbox/checkbox.tsx
```diff
@@ -226,7 +226,7 @@ export class Checkbox implements LabelableComponent, CheckableFormComponent, Int
           role="checkbox"
           tabIndex={this.disabled ? undefined : 0}
         >
-          <svg class="check-svg" viewBox="0 0 16 16">
+          <svg aria-hidden="true" class="check-svg" role="none" viewBox="0 0 16 16">
```
geospatialem, line 229: We won't need the `role` attribute for Chrome/VoiceOver skip over the svg. So if we're looking to limit attributes, we can get this one down to one new attribute. ✂️ 

`<svg aria-hidden="true" class="check-svg" viewBox="0 0 16 16">`

geospatialem (commented):
Klara was able to identify the culprit combination, which is VoiceOver and Chrome. All other screen readers are already avoiding the SVG. To support the case, we will only need the `aria-hidden` attribute. 👍🏻

jcfranco:
@geospatialem Belated thanks for the review and recommendations. I've made the suggested changes and will proceed to install. Let me know if we need to tweak anything after. cc @benelan

jcfranco:
Failing test is unrelated. Proceeding with merge.

Issue #5204, opened 2022-08-23
[Card] Remove hover / press effect


macandcheese:
### Description

The card has a hover / press effect right now. That doesn’t really make sense, as a card can contain one or many internally rendered hit areas.

We should remove the hover and press effects from the card.

Additionally, like we did for border radius, we should have a component level css variable to override the single (and “empty”) default shadow: https://github.com/Esri/calcite-components/issues/4855

### Acceptance Criteria

The hover and press state of card is removed.

### Relevant Info

Cc @ashetland

### Which Component

Card

ashetland:
I'd like to second this. Can we implement it for 1.0?

jcfranco:
Triaged to this sprint.

jcfranco:
PR is up! 

@macandcheese @ashetland Can you confirm if we need any of these states when the card is selectable? I'm assuming we don't since pressing on the component does nothing for selection.

macandcheese:
Correct

geospatialem:
Verified on next
![verify-card-hover](https://user-images.githubusercontent.com/5023024/188229619-15bd3962-d588-4902-857a-0f3408bc8306.gif)

PR #5212, opened 2022-08-23
fix(card): drop hover/press styling

jcfranco:
**Related Issue:** #5204 

## Summary

Drops hover/active styling from `calcite-card` (including when card is selectable).


driskull (approved):
👍

PR #5214, opened 2022-08-24
test(list): use placeholder image util

benelan:
**Related Issue:** #5213

## Summary

The list accessible e2e test is failing because it is using an image generating website which is down:
https://github.com/Esri/calcite-components/runs/7984809614?check_suite_focus=true#step:6:2030

I changed it to use the placeholder image util.


jcfranco (approved):
🎉

Issue #4213, opened 2022-03-08
Deprecate `locale` prop in favor of global `lang` attr/prop


jcfranco:
### Description

The following components have a custom prop (`locale`) for users to specify the language. These should be deprecated in favor of the global `lang` attr/prop:

### _public_

* `input`
* `input-time-picker`
* `date-picker`

### _internal_

* `time-picker`
* `date-picker-month-header`

### Acceptance Criteria

For public usage:

1. deprecate `locale` 
2. use the global attribute watcher util to react to changes to the element's `lang` (see https://github.com/Esri/calcite-components/pull/3596 for more info on this).

For internal usage:
1. remove `locale`
2. same as 2 from public usages

Once this lands, https://github.com/Esri/calcite-components/issues/3781 should be updated accordingly.

### Relevant Info

Related to https://github.com/Esri/calcite-components/issues/3939

### Which Component

Listed in description.

geospatialem:
Verified deprecation of `locale` and full support of `lang` prop in `beta.95`.

PR #5216, opened 2022-08-24
feat: use global lang attribute to determine locale, deprecates `locale` prop

jcfranco:
**Related Issue:** #4213 

## Summary

This deprecates the custom `locale` prop that some components added in favor of `lang`, which is readily available on `HTMLElement`. 

This aligns with https://www.rfc-editor.org/rfc/rfc6067#section-2

> 2.  BCP 47 Required Information
> 
>    Language tags, as defined by [BCP47], are useful for identifying the
>    language of content.  They are also used as locale identifiers (or
>    can be mapped to locales) in many operating environments and APIs.

**Note**: testing will be covered by https://github.com/Esri/calcite-components/pull/5065


src/utils/locale.ts
```diff
@@ -135,3 +136,28 @@ export function localizeNumberString(
...
+}
+
+/**
+ * This util helps resolve a component's locale.
+ * It will also fall back on the deprecated `locale` if a component implemented this previously.
+ *
+ * @param component
+ */
+export function getLocale(component: LangComponent): string {
+  return component.globalAttributes.lang || component.locale || document.documentElement.lang || "en";
```
driskull, line 162: @jcfranco I wonder if we should do...

```ts
return component.globalAttributes.lang || component.locale || component.el.closest("[lang]")?.lang || "en";
```

Would we want to honor a lang attribute set on a parent element?

driskull (approved):
Looks good @jcfranco 🥂 

Only question is if we want to honor lang set on parent elements or document that lang must be set on the element or the HTML document.

src/utils/locale.ts
anveshmekala: Does supporting the lang attribute from parent elements would require actively updating the children component if the `lang` attribute on the parent changes? Considering there could be a chance that the parent element might not be a calcite component.

driskull: I don't think we could support changes to parent elements like that.

eriklharper: > I don't think we could support changes to parent elements like that.

Agreed.  I feel like if the lang is changing in a parent or perhaps at the root `<html>` tag level, the app should re-render every component anyway with the new lang, at least that is probably the 99% most common use case.

src/utils/locale.ts
```diff
@@ -1,4 +1,5 @@
 import { sanitizeDecimalString, sanitizeExponentialNumberString, isValidNumber, BigDecimal } from "./number";
```
eriklharper, line 1: Wondering if we should rename this file to `lang.ts`?

driskull: Yeah good call

src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -133,9 +140,10 @@ export class InputTimePicker
    * BCP 47 language tag for desired language and country format
    *
    * @internal
+   * @deprecated set the global `lang` attribute on the element instead.
+   * @mdn [lang](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang)
    */
-  @Prop({ attribute: "lang", mutable: true }) locale: string =
-    document.documentElement.lang || navigator.language || "en";
+  @Prop({ mutable: true }) locale: string;
```
eriklharper, line 146: Removing the `attribute: "lang"` portion here is causing a regression where the input's value isn't relocalizing when the `lang` attribute is being set.  I updated the demo file to use `setAttribute("lang", newLang)` (which we should probably do as part of this PR) and it fails to relocalize the input's value because the `localeWatcher` isn't firing unless you tell Stencil to have it watch the lang attribute.  There might be another way to solve this, but that's just something I noticed from testing this locally.

eriklharper: Basically I did this and now clicking the locales on the Input Time Picker's demo page no longer update the visible value in the inputs:

image

Since `locale` is deprecated, we should definitely update this demo page to use setAttribute, but we'll want to make sure that works with the changes in this PR too.

eriklharper: Since our global attributes util only updates a component's `globalAttributes` property, one thought perhaps is maybe we could add a feature to `updateGlobalAttributes` where it can take in an additional prop as an argument for each watched global attr (in `lang`'s case, this would be `locale`) that it can sync the new value in addition to `component.globalAttributes`.  That would negate the need for this additional `attribute: "lang"` argument to the prop decorator and still allow for the `localeWatcher` to fire.

eriklharper: Forgot, we're also going to need to fix this in `time-picker` as well, since that component uses this same pattern.

src/utils/locale.ts
anveshmekala: Sounds good. I think we need to document that changes to `lang` attr at the component level will trigger re-render but not at the parent level components.

jcfranco: Good points. I have something in the works for https://github.com/Esri/calcite-components/pull/5065/files that would allow us to look up lang and have components react appropriately. It still needs to go through some testing to spot any perf hits. I think it'll be OK since most use cases will not be changing lang often (builders are the only case that come to mind, atm).

src/utils/locale.ts
jcfranco: I thought about that, and I believe we can keep using `locale` in our doc and internally for consistency across product terminology. The way I'm thinking about this is we're deriving locale from the language, per the spec referenced in the description.

src/components/input-time-picker/input-time-picker.tsx
jcfranco: Good catch. I think this should be fixed with the work that's going on with the built-in localization (mentioned above).

src/utils/locale.ts
eriklharper: I like that.  `locale` is a more specific term too, because it includes both language and the region where the language is spoken.

src/components/input-time-picker/input-time-picker.tsx
jcfranco: @eriklharper Pushed a fix. Can you take another look?

src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -133,13 +144,16 @@ export class InputTimePicker
...
 
+  @Watch("globalAttributes")
```
eriklharper, line 152: Great idea!  Can't believe I didn't think of that :)

src/components/input-time-picker/input-time-picker.tsx
eriklharper: Looks excellent, works with both setAttribute and setting the property directly :)

eriklharper (approved):

Issue #5210, opened 2022-08-23
[calcite-modal] close button padding is broken


joeyHarig:
### Actual Behavior

Notice the misaligned spacing on the top example:

<img src="https://user-images.githubusercontent.com/4733155/186274286-8306499a-dc54-4af9-af56-e7567f83bd55.jpg">

### Expected Behavior

Spacing from the bottom example.

### Reproduction Sample

https://codepen.io/joeyHarig/pen/OJvGbOq?editors=1010

### Reproduction Steps

1. inspect the modal close button
2. notice that the padding is not being assigned correctly

### Reproduction Version

1.0.0-beta.92

### Relevant Info

Likely culprit: https://github.com/Esri/calcite-components/commit/87e6aa8feaa681a8e144ce58f96ed68fa2e4fb98

It looks like the deprecated props should have been removed once the internal, fallback ones were introduced.

### Regression?

1.0.0-beta.87

jcfranco:
@joeyHarig Maybe I'm missing something, but I don't see a typo. It's assigning the padding var value if defined and using the internal variable as a fallback:

https://codepen.io/jcfranco/pen/eYMoBwR?editors=1000

joeyHarig:
@jcfranco Oh I see! I wasn't aware of that syntax for css variable fallbacks - my apologies!

macandcheese:
There was definitely a style regression at some point - notice the misaligned spacing on the top example (latest), compared to before the change (beta 86).
![Group 75](https://user-images.githubusercontent.com/4733155/186274286-8306499a-dc54-4af9-af56-e7567f83bd55.jpg)

jcfranco:
@macandcheese Thanks for catching that. We'll repurpose the issue to fix the spacing discrepancy. Thanks for starting the chain reaction, @joeyHarig! 🎉

anveshmekala:
This is an interesting regression that was not reported by the screener and `padding` is corrected to `padding-block`  during the build which adjusted the spacing.

geospatialem:
Verified on next.579

![verify modal close button padding](https://user-images.githubusercontent.com/5023024/188228915-0fd9ba81-ddb6-43f8-ba1b-3fa792683a82.png)

PR #5217, opened 2022-08-24
fix(modal): close button padding is aligned

anveshmekala:
**Related Issue:** #5210 

This will fix the `padding` misalignment of the `modal` close button.


driskull (approved):
👍 Code changes look good!

Issue #2381, opened 2021-06-23
Question: Should input clearable button be keyboard navigable?


macandcheese:
### Question
Currently, the clearable button for an input (when explicitly requested via prop, or with a default clear button like type search), is keyboard navigable. 

Native "clear" functionality does not allow a user to tab to the affordance. It can be clicked but a keyboard user can use `esc` while the input is in focus: 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search

Should we remove the ability to navigate to this button to better match native functionality? All having the button keyboard-navigable does is add more key strokes needed to tab to, and then invoke that clear action, which can already be accomplished while the input is focused.

macandcheese:
Suggested change reduces amount of keystrokes needed for keyboard users to move through input controls. This also would match our "number up / number down" controls - they cannot be focused since the action can be performed in-field.

benelan:
@geospatialem, thoughts on this?

geospatialem:
### Current behavior
Currently the `clearable` property displays a clear button, but is read back to screen readers as "button" without context. We should include some context, such as "clear entry" to provide meaning behind the button, per [WCAG Success Criterion 3.3.2 Labels or Instructions](https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html).

### Discussion on keyboard navigation with clear button on `input`
It is discouraged to not provide all users with access to content, *unless* there is an alternative available to them. However, in this case users could use the `delete` key, `backspace` key, or even select the entire field length and remove the text content.

We probably shouldn't encourage the use of the `esc` key in this case. While `esc` could be used in many cases, it doesn't have full capabilities across OS' and browsers (Android/Firefox and iOS/Safari are known to not get a delete control to use `esc`). For other combinations where `esc` is supported, the control isn't announced. Since its not a common function, usually only power users, trained users, or those that happen upon it through trial and error will use the `esc` key.

### tldr;
We could set the `clearable` button to have a `tabindex="-1"` as long as we add in an `aria-label="clear entry"` to the button. That way if assistive technologies were to discover it, it would be exposed with the proper role and respond via click or keypress events.

macandcheese:
Thanks for the clarity! Just to confirm, this proposal would match the functionality of the "x" in this native example, correct?
[...]

PR #5219, opened 2022-08-25
feat(input, input-number, input-text): improve keyboard navigation by preventing the clear button from being focusable

Apahadi73:
**Related Issue:** #2381

## Summary

This PR removes ability to perform keyboard navigation on input clear button as discussed in the issue: [Question: Should input clearable button be keyboard navigable? #2381](https://github.com/Esri/calcite-components/issues/2381)


jcfranco (changes requested):
@Apahadi73 Thanks for helping out with this one! I have a few comments:

* `calcite-input-text` needs to be updated as well
* The E2E tests suites will need to be updated to cover the changes made
* The scope of this commit affects more than `input`, so it needs to reflect this (I can tackle this)

Once the above are done, I will proceed to 'adopt' the PR so it goes through all checks and can then get merged.

Let us know if you have any questions.

Apahadi73:
> @Apahadi73 Thanks for helping out with this one! I have a few comments:
> 
> * `calcite-input-text` needs to be updated as well
> * The E2E tests suites will need to be updated to cover the changes made
> * The scope of this commit affects more than `input`, so it needs to reflect this (I can tackle this)
> 
> Once the above are done, I will proceed to 'adopt' the PR so it goes through all checks and can then get merged.
> 
> Let us know if you have any questions.



Thanks, @jcfranco for all of the suggestions. I have made all of the suggested changes. This PR is ready for final review now.

src/components/pick-list/shared-list-tests.ts
```diff
@@ -186,7 +186,7 @@ export function keyboardNavigation(listType: ListType): void {
       await page.keyboard.press("Tab");
       await page.keyboard.press("Tab");
 
-      expect(await getFocusedItemValue(page)).toEqual("one");
```
jcfranco, line 189: This test is unrelated. Can you revert it and update the corresponding input E2E tests:

* https://github.com/Esri/calcite-components/blob/master/src/components/input-number/input-number.e2e.ts
* https://github.com/Esri/calcite-components/blob/master/src/components/input-text/input-text.e2e.ts
* https://github.com/Esri/calcite-components/blob/master/src/components/input/input.e2e.ts

?

jcfranco: Apologies, this is related, but needs to be updated differently. The double tab key presses were meant to move focus from the input, to the clear button and then to the next item. Since this makes the clear button not focusable, that second tab press should no longer be needed (should now tab from input to item). The assertion needs to stay the same.

Apahadi73: Hello @jcfranco, I have applied your suggestion. ☺️

jcfranco: Awesome! Thanks for making the changes. 

I think we'll be able to merge this w/o adopting. 🎉  I scheduled another E2E test run since it looks like it failed on an unrelated test.

jcfranco (approved):
💥

jcfranco:
@Apahadi73 Thanks again for working on this! 🏆

Issue #5027, opened 2022-07-26
calcite-input-date-picker's range mode doesn't provide the correct time for the end date


sonofflynn89:
### Actual Behavior

Using the `range` mode on the input picker, select a valid range and inspect the `calciteDatePickerRangeChange` event. 

Notice that `detail.start` date has a timestamp of 00:00 (Local Time), indicating the start of the day. 
Notice that the `detail.end` date has a timestamp of 00:00 (Local Time), indicating the **_start_** (not the end) of the day

### Expected Behavior

Using the `range` mode on the input picker, select a valid range and inspect the `calciteDatePickerRangeChange` event. 

Notice that `detail.start` date has a timestamp of 00:00 (Local Time), indicating the start of the day. 
Notice that the `detail.end` date has a timestamp of 23:59 (Local Time), indicating the **_end_** (not the start) of the day

When I select a range, (i.e. July 24 - July 26), I expect the date picker to return values that match the sentiment, _I'm interested in items from the start of July 24 through the end of July 26_. 

The existing behavior effectively chops off the last day. If I want items through the 26, I have to select the 27.

### Reproduction Sample

https://codepen.io/sonofflynn89/pen/QWmqNXZ?editors=1111

### Reproduction Steps

1. Make the console tab within codepen visible
2. Select a valid date range
3. See the console output
4. Verify that the end date is the correct day, but the wrong time (shows 00:00 instead of 23:59)

### Reproduction Version

1.0.0-beta.86

### Relevant Info

I believe that you have access to knowing which date picker has been changed before you emit the event. If it's the end date, just set the time before you emit (see https://bobbyhadz.com/blog/javascript-get-start-and-end-of-day for an example).

geospatialem:
Verified in `beta.95`.

![verified end of day in input date picker component](https://user-images.githubusercontent.com/5023024/193875913-3a3b0c15-7038-4aa8-965d-c97a112455b5.png)

PR #5220, opened 2022-08-25
fix(input-date-picker): range end date is set to end of the day

anveshmekala:
**Related Issue:** #5027 

This will fix the `endDate` time in the `date-picker` and `input-date-picker` components which is now set to `23.59.59.999` local time .


src/components/date-picker/date-picker.tsx
```diff
@@ -193,7 +200,7 @@ export class DatePicker {
   // --------------------------------------------------------------------------
   connectedCallback(): void {
     if (Array.isArray(this.value)) {
-      this.valueAsDate = this.value.map((v) => dateFromISO(v));
+      this.valueAsDate = mapValueToValueAsDate(this.value);
```
driskull, line 203: nitpick: I'm not sold on the name of this utility function.

I actually think it was more readable before :D

src/utils/date.ts
```diff
@@ -212,3 +216,14 @@ export function getDaysDiff(date1: Date, date2: Date): number {
...
   return (ts1 - ts2) / (1000 * 3600 * 24);
 }
+
+/**
+ * Set time of the day to the end.
+ *
+ * @param {Date} date Date.
+ * @returns {Date} Date with time set to end of day .
+ */
+export function setEndOfDay(date: Date): Date {
```
driskull, line 226: THE END OF DAYS

driskull (approved):
🔥

src/components/date-picker/date-picker.tsx
anveshmekala: How about `getValueAsDate` ?

driskull: Maybe `getValueAsDateRange`

Issue #4212, opened 2022-03-08
Audit components for `HTMLElement#click` behavior


jcfranco:
### Background

Stems from https://github.com/Esri/calcite-components/issues/4211

Components similar to switch (e.g., checkbox) with single-click behavior should be updated to work with `HTMLElement.click`.

This issue will focus on identifying these components and which ones are not handling the expected `myElement.click()` behavior.

### Desired Outcome

A list of components that need to handle `click()`. We will create a follow-up issue/epic to tackle each.

- [x] checkbox
- [x] switch
- [x] radio-button

benelan:
verified on beta.94
https://codepen.io/benesri/pen/MWGJPrV?editors=0010

PR #5222, opened 2022-08-26
test(radio-button): update to help test click behavior via HTMLElement.click()

Elijbet:
**Related Issue:** #4212

## Summary

Update tests that cover behavior on element click to use the .callMethod("click") in addition to .click(). In the case of the `radio-button`, click triggers a boolean toggle on the checked prop.

### DIFF
--- a/src/components/radio-button/radio-button.e2e.ts
+++ b/src/components/radio-button/radio-button.e2e.ts
@@ -195,7 +195,9 @@ describe("calcite-radio-button", () => {
     expect(await first.getProperty("checked")).toBe(true);
     expect(await second.getProperty("checked")).toBe(false);
 
-    //in addition to .click() check with .callMethod("click") as well
+    /* call a 'click' @method on a component with stencil utility function .callMethod("click") as an additional check to
+    the HTMLElement.click() method that simulates a mouse click on an element */
+
     await second.callMethod("click");
     await page.waitForChanges();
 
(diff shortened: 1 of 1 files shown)


src/components/radio-button/radio-button.e2e.ts
```diff
@@ -272,6 +285,17 @@ describe("calcite-radio-button", () => {
 
     expect(focusEvent).toHaveReceivedEventTimes(1);
     expect(blurEvent).toHaveReceivedEventTimes(1);
+
+    //in addition to .click() check with .callMethod("click") as well
```
jcfranco, line 289: I don't think we need to assert on focus/blur events. The click toggling `checked` should suffice.

src/components/radio-button/radio-button.e2e.ts
```diff
@@ -194,6 +194,19 @@ describe("calcite-radio-button", () => {
 
     expect(await first.getProperty("checked")).toBe(true);
     expect(await second.getProperty("checked")).toBe(false);
+
+    //in addition to .click() check with .callMethod("click") as well
```
jcfranco, line 198: Let's rephrase this to give context behind the different clicks. The code already lets us know we're using `click` and `.callMethod("click")`, but doesn't explain why.

jcfranco (changes requested):

src/components/radio-button/radio-button.e2e.ts
Elijbet: See if this solves it.

jcfranco (approved):
💥

PR #5225, opened 2022-08-26
test(switch): update to help test click behavior via HTMLElement.click()

Elijbet:
**Related Issue:** #4212

## Summary

Update tests that cover behavior on element click to use the `.callMethod("click")` in addition to `.click()`. In the case of the `switch`, click triggers a boolean toggle on the `checked` prop.


jcfranco (approved):
🏆

PR #5229, opened 2022-08-26
ci: close past due milestone when its last issue is moved

benelan:
**Related Issue:** #

## Summary
We have a GitHub Action that closes a milestone when all of the issues are completed. Currently it is only triggered when an issue is closed. Sometimes an issue does not pass the QA verification and the issue needs to be moved to the current sprint or backburner. With this PR the milestone will properly close.


jcfranco (approved):
Noice!

Issue #5218, opened 2022-08-24
combobox: tabbing through the component should focus input field. 


anveshmekala:
### Test type

new test to verify if tabbing through the component focuses the `input` field in the `combobox`

### Which Component(s)

combobox

### PR skipped, if applicable

#

### Additional Info

Related to #5191

geospatialem:
Verified test on master branch.

PR #5234, opened 2022-08-29
test(combobox): add test to verify user is able to tab through the component

anveshmekala:
**Related Issue:** #5218 

## Summary

This will add an E2E test to verify if the user is able to tab through the `Combobox` by the `tab` key once and able to type.


driskull (approved):
👍

Issue #4957, opened 2022-07-18
Input-time-picker: left-hand field initially displays number instead of am/pm when am/pm precedes time for locales with 12 hr time period 


annierm18:
### Actual Behavior

After completing step 4 in the repro steps, the user will see a number. 
![Screenshot (94)](https://user-images.githubusercontent.com/25759835/179626677-a1416fe4-67a2-4610-b2af-c9d933756fc2.png)

However, after step 5, the left field changes to am/pm (translated) and the number shifts to the right.
![Screenshot (95)](https://user-images.githubusercontent.com/25759835/179626684-0eff5bea-0cb1-438c-97df-252cc3441cdd.png)

### Expected Behavior

Every time (including the first time) I click the up/down arrows for the left-hand field of the input-time-picker component, I expect that the correct data should appear i.e. 
- a number appears for locales where the formatting is time followed by am/pm OR 
- am/pm (translated) for locales where am/pm precedes the time.

### Reproduction Sample

https://jsbin.com/wapobulexo/edit?html,output

### Reproduction Steps

Note: So far, my testing has not been impacted by locale set by operating system 
1. Change the browser locale to a locale that has 12 hour time period and its am/pm text precede the time, e.g. Korean or Arabic (Ar)
2. Navigate to https://developers.arcgis.com/calcite-design-system/components/input-time-picker/
3. Click the input-time-picker
4. Use the up/down arrows on the left-most field
5. Use the up/down arrows in the middle or right field

### Reproduction Version

v1.0.0-beta.83

### Relevant Info

Reproduced on Chrome and Firefox
Windows10 OS

jcfranco:
Related issue: https://github.com/Esri/calcite-components/issues/4893

geospatialem:
Verified on `next`.

![verify-time-picker-locale](https://user-images.githubusercontent.com/5023024/189237757-8a30305e-64d1-477a-8a9a-38588897939a.gif)

PR #5246, opened 2022-08-30
fix(time-picker, input-time-picker): initialize meridiem at the start of the time for rtl locales

benelan:
**Related Issue:** #4957 

## Summary
Makes sure the meridiem location is determined on load.


driskull (approved):
👍

Issue #4896, opened 2022-07-11
Active tab underline doesn't fill the width if an icon is dynamically added


mpayson:
### Actual Behavior

If an icon is dynamically added to an active tab title, the underline for the title does not fill the width
<img width="218" alt="Screen Shot 2022-07-11 at 2 53 44 PM" src="https://user-images.githubusercontent.com/5892789/178337502-9166c40b-5a13-4265-b9d4-42e8f6a826ce.png">

### Expected Behavior

The underline width matches the width of the title

### Reproduction Sample

https://codepen.io/mpayson/pen/jOzrXaE

### Reproduction Steps

1. Open the code pen
2. Click the button
3. Notice an indicator is added and the underline is short

### Reproduction Version

beta.83

geospatialem:
Verified in `beta.95`.

![verified active tab style for tab with indicator](https://user-images.githubusercontent.com/5023024/193875035-aa88adc2-8d4c-4c48-94e8-ae45cde1606f.png)

PR #5250, opened 2022-08-31
fix(tab): active indicator width grows with tab's width

anveshmekala:
**Related Issue:** #4896 

This PR will fix the active indicator width when user add's `icon` at the start or end of `tab-title`  dynamically.


src/components/tab-title/tab-title.tsx
```diff
@@ -216,6 +218,7 @@ export class TabTitle implements InteractiveComponent {
     updateHostInteraction(this, () => {
       return this.selected;
     });
+    this.resizeObserver?.observe(this.containerEl);
```
driskull, line 221: Can you move this to where you set the containerEl? `ref={(el) => (this.containerEl = el as HTMLDivElement)}`

driskull (changes requested):
Looks good just one change

driskull (approved):
👍 Looks good

PR #5256, opened 2022-09-01
feat(flow-item): Add back button tooltip.

driskull:
**Related Issue:** N/A

## Summary

feat(flow-item): Add back button tooltip.


benelan (approved):
LGTM 👍 

Side note, in the Flow stories, the [Flow Items are called Panels](https://github.com/Esri/calcite-components/blob/830f7675b08fc97036e3234978ac8089d7658b6d/src/components/flow/flow.stories.ts#L158-L159). Should we change it to "Flow Item 1" and Flow Item 2" or is there a reason it is like that? Flow Item doesn't have a doc page, maybe that's why. Should it have a doc page?

driskull:
> Should we change it to "Flow Item 1" and Flow Item 2"

Yes we should!

jcfranco:
Apologies for not chiming in sooner, but this looks great code-wise, but we now have an inconsistent pattern for setting tooltips. [Slot](https://github.com/Esri/calcite-components/pull/5293) vs prop (this). @driskull Can you create a follow-up issue to define and adopt a consistent tooltip pattern for all components?

driskull:
https://github.com/Esri/calcite-components/issues/5319

Issue #5168, opened 2022-08-16
Overflowing subtree content is clipped


nwhittaker:
### Actual Behavior

Overflowing content within a nested `<calcite-tree>` is clipped by the `"children"` slot container.

<img width="286" alt="Screen Shot 2022-08-16 at 5 08 50 PM 2" src="https://user-images.githubusercontent.com/421496/184986620-2bd41faa-f7fb-4a1b-a06a-4063ef0eeaa2.png">

### Expected Behavior

Overflowing content is allowed to flow out of the `"children"` slot container.

<img width="286" alt="Screen Shot 2022-08-16 at 5 11 10 PM 2" src="https://user-images.githubusercontent.com/421496/184986666-37e6ddad-b06a-4a58-b8d1-2d8454b3ac5a.png">

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/ZExmMbQ

### Reproduction Steps

1. Visit the sample
2. Click the overflow menu for _Layer 1_ and note that the menu flows out of the `<calcite-tree-item>` container.
3. Click the overflow menu for _Layer 2.1_ and note that the menu causes the layer name to scroll up and become partially obscured.

### Reproduction Version

beta.91

### Relevant Info

`<calcite-tree>` is uniquely useful for displaying a hierarchy of layers and nested groups of layers. It'd also be helpful to expose within each `<calcite-tree-item>` a menu of actions that can be performed on that layer.

FWIW, removing `overflow: hidden` from the `.children-container` element in the `<calcite-tree-item>` shadow DOM seemed to get things most of the way there (save some `z-index` futzing). The expand/collapse animation appeared to be unaffected and not depending on the overflow style.

jcfranco:
@paulcpederson @macandcheese I have a couple of questions:

* Do you recall why tree items use [`overflow: hidden`](https://github.com/Esri/calcite-components/blob/master/src/components/tree-item/tree-item.scss#L111) on the children container?
* Would it be possible to transition on the [children container's height vs transform](https://github.com/Esri/calcite-components/blob/master/src/components/tree-item/tree-item.scss#L116-L117)? The latter will mess up `floating-ui` positioning.

paulcpederson:
>  Do you recall why tree items use [overflow: hidden](https://github.com/Esri/calcite-components/blob/master/src/components/tree-item/tree-item.scss#L111) on the children container?

[...]

PR #5261, opened 2022-09-02
fix(tree-item): overflow slotted elements are no longer hidden

anveshmekala:
**Related Issue:** #5168 

## Summary

This fix will no longer hide the overflow slotted elements in `tree-item` and the `floating-ui` element no longer overlaps with the nested tree elements.


src/components/tree-item/tree-item.tsx
```diff
@@ -165,6 +166,13 @@ export class TreeItem implements ConditionalSlotComponent {
     this.updateAncestorTree();
   }
 
+  componentDidrender(): void {
+    if (this.hasChildren) {
+      readTask(() => {
+        this.getChildrenContainerHeight();
```
driskull, line 172: Should this be renamed to `setChildrenContainerHeight`?

driskull: Nice use of `readTask`. @jcfranco should we use read/write task more often? In other cases I've just modified the element at various times but we might be better off using this for perf.

driskull (approved):
This looks good to me. Can we get some more eyes on it first?

src/components/tree-item/tree-item.tsx
```diff
@@ -165,6 +166,13 @@ export class TreeItem implements ConditionalSlotComponent {
     this.updateAncestorTree();
   }
 
+  componentDidrender(): void {
```
jcfranco, line 169: Can you verify this get called? There's a typo in the lifecycle method.

src/components/tree/tree.stories.ts
```diff
@@ -80,3 +80,40 @@ export const darkThemeRTL_TestOnly = (): string => html`
   </calcite-tree>
 `;
 darkThemeRTL_TestOnly.parameters = { themes: themesDarkDefault };
+
+export const OverflowingSubtree_TestOnly = stepStory(
```
jcfranco, line 84: Can you avoid using steps if you update the dropdown to be open?

src/components/tree-item/tree-item.tsx
jcfranco: Yeah, we could use `readTask`/`writeTask` more to minimize layout thrashing (for the cases we truly need to make changes that would cause this), etc.

src/components/tree-item/tree-item.tsx
anveshmekala: Good catch . This is not being invoked and animation is broken.

src/utils/openCloseComponent.ts
```diff
@@ -66,8 +71,9 @@ function transitionEnd(event: TransitionEvent): void {
  * Used to emit (before)open/close events both for when the opacity transition is present and when there is none (transition-duration is set to 0).
  *
  * @param component
+ * @param nonOpenCloseComponent
  */
-export function onToggleOpenCloseComponent(component: OpenCloseComponent): void {
+export function onToggleOpenCloseComponent(component: OpenCloseComponent, nonOpenCloseComponent = false): void {
   readTask((): void => {
```
Elijbet, line 77: Is there any other `transitionProp` we expect to handle besides `open` and `expanded`? If not, maybe we can name it `expandedCollapsedComponent` instead of `nonOpenCloseComponent` in here? And check for `expanded` in the ternary below... Or we could at some point refactor/rename the util and have one `transitionProp` which we set from the getgo.

src/utils/openCloseComponent.ts
```diff
@@ -100,13 +114,15 @@ export function onToggleOpenCloseComponent(component: OpenCloseComponent): void
...
+ * For component which do not have open prop, use `onToggleOpenCloseComponent` implementation.
+ *
  * @param component
  */
 export function connectOpenCloseComponent(component: OpenCloseComponent): void {
   disconnectOpenCloseComponent(component);
   if (component.transitionEl) {
     const boundOnTransitionStart: (event: TransitionEvent) => void = transitionStart.bind(component);
-    const boundOnTransitionEnd: (event: TransitionEvent) => void = transitionEnd.bind(component);
+    const boundOnTransitionEnd: (event: PageTransitionEventInit) => void = transitionEnd.bind(component);
```
Elijbet, line 125: We're not going to be needing this anymore, since we're removing the connect logic altogether, along with the comment above.

Elijbet (approved):
Looks good so far :) Let's figure out how to handle `selection=multiple` like you said.

src/utils/openCloseComponent.ts
anveshmekala: i think this was a typo.

src/utils/openCloseComponent.ts
anveshmekala: i don't have a strict preference here. the reason behind is to convey the idea of deviating from the original naming pattern `openClose` . This might help if we have another set of `transitionProp` like expanded in some other component or we can refactor the interface to something like `transitionComponent`.

src/utils/openCloseComponent.ts
Elijbet: Sorry, I know it was a draft.

src/components/tree/tree.stories.ts
```diff
@@ -80,3 +80,40 @@ export const darkThemeRTL_TestOnly = (): string => html`
...
+            appearance="transparent"
+            color="neutral"
+            icon-start="ellipsis"
+            slot="dropdown-trigger"
+          ></calcite-button>
+        </calcite-dropdown>
+      </calcite-tree-item>
+    </calcite-tree>`,
+  createSteps("calcite-tree").click("calcite-button").snapshot("OverFlowingSubtree")
+);
```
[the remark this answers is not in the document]
benelan: Yeah no more StepStories, I believe I removed the helper but I'll double check. I'll make an announcement when we are ready for storybook interaction tests but there is still some more work to get there. So no interactive visual snapshots for now

Issue #5259, opened 2022-09-02
Add groupSeparator prop to Pagination


Elijbet:
### Description

Currently, the `pagination` component automatically adds the language's corresponding number separator to the page value, but not all situations require it displayed as separated. Devs would need the option to toggle it to match the context of where they are used, eg. AGOL does not show separators for the content.

### Acceptance Criteria

`groupSeparator` prop is added to `pagination` which displays numbers as separated and localized when true and does not when false.

### Which Component

pagination

geospatialem:
Verified on `next`.

PR #5265, opened 2022-09-03
feat(pagination)!: add groupSeparator prop to enable toggling display of separator in pages

Elijbet:
**Related Issue:** #5259

## Summary

`groupSeparator` prop  is added to `calcite-pagination` to display numbers as separated and localized when true (sometimes, You just gotta keep 'em separated!).

Not all situations require pages displayed as group-separated. Devs would need the option to toggle it to match the context of where they are used.

Tests for number locale support: check the display of group separators for multiple locales only when `group-separator` is present and true.


src/components/pagination/pagination.e2e.ts
```diff
@@ -179,4 +180,81 @@ describe("calcite-pagination", () => {
...
+
+    beforeEach(async () => {
+      page = await newE2EPage();
+      await page.setContent(
+        html`<calcite-pagination lang="en" group-separator total="150000000" num="10"></calcite-pagination>`
+      );
+      element = await page.find("calcite-pagination");
+
+      // assign temporary class to the last displayed page to query outside of evaluate
+      await page.evaluate(() => {
```
jcfranco, line 208: Wouldn't it be simpler to do something like this?

```ts
const buttons = await page.findAll(`calcite-pagination >>> .${CSS.page}`);
const last = buttons[buttons.length - 1];
await last.click(); 
```

src/components/pagination/pagination.e2e.ts
```diff
@@ -179,4 +180,81 @@ describe("calcite-pagination", () => {
...
+        const el = document.querySelector("calcite-pagination");
+        const lastButton = Array.from(el.shadowRoot.querySelectorAll("button")).find((el) => {
+          return el.innerText === "15,000,000";
+        });
+        lastButton.classList.add("lastPage");
+      });
+      const lastButton = await page.find(`calcite-pagination >>> .lastPage`);
+      await lastButton.click();
+
+      const buttonListAbridged = await (await page.findAll(`calcite-pagination >>> .${CSS.page}`)).slice(-4);
```
jcfranco, line 218: I'd suggest creating a utility function inside this `describe` block that other tests can reference within the node testing context. For example:

```ts
async function getDisplayedValues(): Promise<string[]> {
  const buttons = await page.findAll("calcite-pagination >>> .page");
  const buttonsWithGroupSeparator = buttons.slice(-4);
  return buttonsWithGroupSeparator.map(button => button.innerText);
}

// within it("something", async () => {

expect(noSeparator).toEqual(
  // tests can use this method directly without having to run it within the browser context
  await getDisplayedValues()
); 
```

jcfranco (changes requested):
Looking good!

src/components/pagination/pagination.e2e.ts
```diff
@@ -179,4 +180,70 @@ describe("calcite-pagination", () => {
       expect(toggleSpy).toHaveReceivedEventTimes(0);
     });
   });
+
+  describe("number locale support", () => {
+    let page: E2EPage;
+    let element: E2EElement;
+    let noSeparator: string[];
+    let withSeparator: string[];
+    let getDisplayedValuesArray;
```
jcfranco, line 189: This is no longer needed.

src/components/pagination/pagination.e2e.ts
```diff
@@ -179,4 +180,70 @@ describe("calcite-pagination", () => {
...
+      const buttonsTestedForSeparator = buttons.slice(-4);
+      return buttonsTestedForSeparator.map((button) => button.innerText);
+    }
+
+    beforeEach(async () => {
+      page = await newE2EPage();
+      await page.setContent(
+        html`<calcite-pagination lang="en" group-separator total="150000000" num="10"></calcite-pagination>`
+      );
+      await page.exposeFunction("getDisplayedValuesArray", getDisplayedValuesArray);
```
jcfranco, line 211: This is no longer needed.

src/components/pagination/pagination.e2e.ts
```diff
@@ -179,4 +180,70 @@ describe("calcite-pagination", () => {
       expect(toggleSpy).toHaveReceivedEventTimes(0);
     });
   });
+
+  describe("number locale support", () => {
+    let page: E2EPage;
+    let element: E2EElement;
+    let noSeparator: string[];
```
jcfranco, line 187: This and the variable below can be moved inside their respective functions. Reasoning: they are not initialized in `beforeEach` and are used in different contexts (not shared across tests).

jcfranco (approved):
Aside from a few comments, this...

✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅
✅👍✅✅✅👍👍👍👍✅👍👍👍👍👍✅👍✅✅✅👍✅👍✅
✅👍✅✅✅👍✅✅✅✅✅✅👍✅✅✅👍👍✅👍👍✅👍✅
✅👍✅✅✅👍✅👍👍✅✅✅👍✅✅✅👍✅👍✅👍✅👍✅
✅👍✅✅✅👍✅✅👍✅✅✅👍✅✅✅👍✅✅✅👍✅✅✅
✅👍👍👍✅👍👍👍👍✅✅✅👍✅✅✅👍✅✅✅👍✅👍✅
✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅

Issue #5069, opened 2022-08-02
Changes to `autofocus` property not reflected in `calcite-input` attributes


nwhittaker:
### Actual Behavior

Given a `<calcite-input>` element, toggling its `autofocus` property does not add/remove its `autofocus` attribute.

### Expected Behavior

Given a `<calcite-input>` element, toggling its `autofocus` property does add/remove its `autofocus` attribute.

### Reproduction Sample

https://ember-twiddle.com/9f794f79e6c1d25b00a276a2ee8524b9

### Reproduction Steps

1. Visit the sample and wait for the `<calcite-input>` to load.
2. Use dev tools to inspect the `<calcite-input>` and the `<input>` within its shadow-dom.
3. Click the **Toggle autofocus** button repeatedly and notice the `autofocus` attribute is added/removed on the `<input>` but not the `<calcite-input>`.

### Reproduction Version

beta.86

### Relevant Info

The above sample is contrived, but it's indicative of how Ember renders templates. Given something like:

```hbs
<calcite-input autofocus={{...}} />
```

Ember ends up setting the `autofocus` property instead of the attribute, and the logic to compute the `autofocus` value could occur after the `<calcite-input>` is hydrated.

The problem with not adding/removing the `autofocus` attribute from the `<calcite-input>` is that it becomes difficult to query for the auto-focused element (i.e. `document.querySelector(['autofocus'])`). It's useful to find this element so we can call `setFocus()` on it for cases where we need to shift the focus relative to an Ember render hook (e.g. after opening a modal, visiting a new step in a stepper, etc...).

This is essentially the same issue as https://github.com/Esri/calcite-components/issues/4765, but with less confusion and a more Ember-focused test case.

geospatialem:
@nwhittaker Would you be able to verify if the above fix works with Ember and `beta.95`? 

Wasn't able to get the [reproduction sample above](https://ember-twiddle.com/9f794f79e6c1d25b00a276a2ee8524b9) to load as a result of a API limit on the gist.

Issue #4765, opened 2022-06-16
Selector `calcite-input[autofocus]` not matched


nwhittaker:
### Actual Behavior

Given a `<calcite-input>` element with the `autofocus` attribute enabled, the `calcite-input[autofocus]` selector never matches the element.

### Expected Behavior

Given a `<calcite-input>` element with the `autofocus` attribute enabled, the `calcite-input[autofocus]` selector matches the element.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/dydreNL?editors=1011

### Reproduction Steps

1. Visit the sample
2. Open the **Console** and notice a value of `null` is logged instead of the `calcite-input` element

### Reproduction Version

beta.83

### Relevant Info

It's possible other attributes on this, or other elements, exhibit the same behavior.

jcfranco:
@nwhittaker Can you clarify whether you are using the attribute or the property? Styling should work via the attribute: https://codepen.io/jcfranco/pen/NWYgvdd?editors=1111

Also, could you provide more info on your use case?

nwhittaker:
@jcfranco, I think you're right, this should be working as expected. I think my confusion was compounded by a couple things:
1. A workaround in app code that cleared the `autofocus` attribute. This was working around a now-fixed Safari bug where it ignored the `FocusOptions.preventScroll` option.
2. Autofocusing not working correctly in Codepen (at least for me) due to a security restriction: `Blocked autofocusing on a <calcite-input> element in a cross-origin subframe.`

Thanks for taking a look and verifying the correct behavior.

>Also, could you provide more info on your use case?

When dynamically rendering a layout in Ember that contains an `autofocus` element, we need to manually focus it. Finding the element via an `[autofocus]` selector was handiest.

PR #5269, opened 2022-09-04
feat: reflect props if they do not represent content nor rich-data

jcfranco:
**Related Issue:** #5069, #4765

## Summary

This updates props to reflect if they conform to the following:

- are static or will not be updated frequently during the component lifespan (e.g., a number that represents a range min or max would be reflected, but a number that represents a value that will constantly be updated by the user would not)
- value represents non-rich data or booleans/numbers/strings that are not used as content (e.g., a string that represents a mode would be reflected, but a string that represents a placeholder, title or summary would not)
- are public and belong to a public component
- required for internal styling or would make internal styling easier

cc @nwhittaker

### Notes

* the conventions doc has been updated with this guidance
* `locale` was omitted since https://github.com/Esri/calcite-components/pull/5216 will remove its default assigned value


src/components/avatar/avatar.tsx
```diff
@@ -28,16 +28,16 @@ export class Avatar {
   @Prop({ reflect: true }) scale: Scale = "m";
 
   /** Specifies the `src` to an image (remember to add a token if the user is private). */
-  @Prop() thumbnail: string;
+  @Prop({ reflect: true }) thumbnail: string;
```
driskull, line 31: @jcfranco would some of these be content? Like `thumbnail`, `fullName`? Maybe the "content" part is a little vague and we can make the rule more concrete by defining it. Would "content" be a sentence or sentences? Anything with spaces in it?

src/components/popover-manager/popover-manager.tsx
```diff
@@ -32,7 +32,7 @@ export class PopoverManager {
    *
    * @default `[data-calcite-popover-reference]`
    */
-  @Prop() selector = "[data-calcite-popover-reference]";
```
driskull, line 35: This component is deprecated so maybe we don't need to update it?

src/components/tooltip-manager/tooltip-manager.tsx
```diff
@@ -20,7 +20,7 @@ export class TooltipManager {
    *
    * @default `[data-calcite-tooltip-reference]`
    */
-  @Prop() selector = "[data-calcite-tooltip-reference]";
+  @Prop({ reflect: true }) selector = "[data-calcite-tooltip-reference]";
```
driskull, line 23: This component is deprecated so maybe we don't need to update it?

driskull (approved):
👍 LGTM!

src/components/avatar/avatar.tsx
jcfranco: I actually had doubts about these avatar props because of what you mentioned. I proceeded to reflect since I saw these more along the lines of `icon` where it does affect the UI, but not 1-1 (i.e., the icon name isn't rendered).

src/components/popover-manager/popover-manager.tsx
jcfranco: I had a note to exclude deprecated props, but it was simple to set up that I just went for it for the greater good. 

greater-good

driskull: Sounds good to me 👍

PR #5277, opened 2022-09-06
fix(list-item): Use margin instead of padding when slotting icons.

driskull:
**Related Issue:** N/A

## Summary

fix(list-item): Use margin instead of padding when slotting icons.


jcfranco (approved):
💥

PR #5279, opened 2022-09-07
docs(storybook): add storybook guidelines to doc conventions

jcfranco:
**Related Issue:** N/A

## Summary

This provides storybook guidelines that came from https://github.com/Esri/calcite-components/pull/5257/

TL;DR: we introduced 2 separate storybook builds (playground and screenshot testing) to keep our screenshot quota down and as a general story cleanup pass.


conventions/Documentation.md
```diff
@@ -34,34 +34,67 @@ You can provide code snippets demonstrating a specific behavior or pattern for a
...
+  export default {
+    title: "Components/My Component",
+    parameters: {
+      notes: readme
+    },
+    ...storyFilters()
+  };
+  ```
+- Update the [main custom theme](https://github.com/Esri/calcite-components/blob/master/src/05-custom-theme.stories.mdx) story instead of adding a specific story showing how to use a custom CSS prop
+- Should only have HTML for the component or use case itself (e.g., no need to wrap in calcite-label)
```
eriklharper, line 78: ☝️ I'm wondering if this matters for the Storybook a11y tab to function properly or not?  Maybe that's not a concern on this storybook anymore?

conventions/Documentation.md
```diff
@@ -34,34 +34,67 @@ You can provide code snippets demonstrating a specific behavior or pattern for a
...
+- Skip adding or updating a story if it is covered by an existing one
+
+#### Interactive showcase stories
+
+- Should have knobs configured to allow users to toggle different options available to the component
+- The simplest and base configuration should be named `simple` for consistency
+- Should have one story per unique, supported use case
+- Stories that can be covered by adding a new prop/knob should update an existing playground story
+
+#### Screenshot test stories
```
eriklharper, line 89: ☝️ I would revise this title to say `Screenshot Test-Only Stories` since the `simple` story is also technically a "test" story since it is also tested by the visual regression tool, in addition to providing knobs, etc.

conventions/Documentation.md
```diff
@@ -34,34 +34,67 @@ You can provide code snippets demonstrating a specific behavior or pattern for a
...
+    parameters: {
+      notes: readme
+    },
+    ...storyFilters()
+  };
+  ```
+- Update the [main custom theme](https://github.com/Esri/calcite-components/blob/master/src/05-custom-theme.stories.mdx) story instead of adding a specific story showing how to use a custom CSS prop
+- Should only have HTML for the component or use case itself (e.g., no need to wrap in calcite-label)
+- Update the `simple` story with a corresponding knob instead of adding a story specific to a new prop
+- Skip adding or updating a story if it is covered by an existing one
```
eriklharper, line 80: Might be cool to add one more bullet point here along the lines of: `- A full "kitchen sink" example of the component preferred for the primary `simple` story to showcase all the possible props available and to allow for the most comprehensive state for visual regression testing`

conventions/Documentation.md
driskull: I don't think we really use or used the a11y tab so I'm not against removing it if we want to.

conventions/Documentation.md
```diff
@@ -34,66 +34,76 @@ You can provide code snippets demonstrating a specific behavior or pattern for a
 
 ## Storybook
 
-Calcite Components uses [Storybook](https://storybook.js.org/) to provide an interactive showcase of components with accompanying documentation and also to provide previews for screenshot testing.
+Calcite Components uses [Storybook](https://storybook.js.org/) to provide an interactive showcase of components with accompanying documentation and also to provide previews for screenshot testing. Storybook is deployed during every `next` release, whereas as the [SDK site](https://developers.arcgis.com/calcite-design-system/components/) is only updated after `beta` releases. Storybook is a great resource for testing the newest changes and viewing the newest API reference.
```
jcfranco, line 37: `whereas as the` => `whereas the`

benelan (approved):
Pushed some cleanups. LGTM, feel free to merge after you look through my changes

conventions/Documentation.md
```diff
@@ -34,66 +34,76 @@ You can provide code snippets demonstrating a specific behavior or pattern for a
...
 #### Screenshot test-only stories
 
-- Should have the minimal HTML to reproduce the test scenario
-- Should not have knobs as screenshot tests will not interact with them
-- Stories that are only meant for testing should use the `_TestOnly` suffix, which ensures it is only used for the screenshot test build
-- Stories that are not meant for testing (e.g., require interaction not supported by the screenshot test environment) should use the `_NoTest` suffix, which ensures it is only used for the internal playground build
-- The `simple` story and any other significant configuration should have a matching `darkThemeRtl_TestOnly` story
+- Should have the minimal HTML needed to reproduce the test scenario
+- Should not have knobs since screenshot tests cannot interact with them
+- Stories that are only meant for testing should use the `_TestOnly` suffix, which ensures it is only used for the screenshot test build. For example, if a snapshot with specific properties is needed for testing, it wouldn't need to be the playground where knobs can be used to display the state in the `simple` story.
```
jcfranco, line 104: > it wouldn't need to be the playground

Seems like this is missing something between `to be` and `the playground`.

WDYT about adding `displayed in` or `included in`?

conventions/Documentation.md
```diff
@@ -34,66 +34,76 @@ You can provide code snippets demonstrating a specific behavior or pattern for a
...
 
 export default {
-  title: "Components/My component",
-  parameters: { notes }
+  title: "Components/My Component",
+  parameters: { notes },
+  ...storyFilters()
 };
 
-export function simple = (): string => html`
```
jcfranco: crying

jcfranco (commented):
Thanks for the edits!

conventions/Documentation.md
benelan: May be worth mentioning this somewhere in the a11y conventions doc so we can point to it if people say "Storybook says you're components aren't accessible" cc @geospatialem

benelan: I meant we should mention that the stories may not be accessible since we aren't adding labels. So I think this isn't true anymore:

> As new components and enhancements are added, ensure stories are updated to test accessibility. This includes as properties are added to components, to ensure we're upholding high standards to fit our audience's needs.

But correct me if I'm wrong @driskull @jcfranco

Issue #5186, opened 2022-08-18
Option to hide selection-mode from calcite trees


ade10176:
### Description

The bullet indicator on a selected/hovered state of a calcite tree conflicts with the pending changes indicator that we've been using in Field Maps. We're looking for a way to relax the appearance of these states to the degree that the bullet is not present.

![image](https://user-images.githubusercontent.com/49533268/185497143-07afd773-1f4f-4690-9162-9015845a86be.png)

cc: @nwhittaker 

### Acceptance Criteria

Provide a way (possibly an `appearance` or `selection-mode` option) that has, or includes, the effect of suppressing the bullet for selected/hovered states of tree items.

### Which Component

calcite-tree

jcfranco:
Would the [`none` selection mode](https://github.com/Esri/calcite-components/issues/3121) help here? Otherwise, we'll design to review this. cc @macandcheese

nwhittaker:
Thanks, I do think this new selection mode does get us most of the way there. One difference is that the disclosure arrows are no longer outdented.

| A "blank" selection mode | `none` selection mode |
| - | - |
| <img width="160" alt="Screen Shot 2022-09-08 at 10 00 22 AM" src="https://user-images.githubusercontent.com/421496/189144046-91102c23-1b64-4300-a898-bfa034cfbe4a.png"> | <img width="162" alt="Screen Shot 2022-09-08 at 9 59 39 AM" src="https://user-images.githubusercontent.com/421496/189144216-c208c4c8-cda9-464b-a0d1-a85349f659b2.png"> |

Not sure if that's a bug or by design, though, but I imagine we can work with or around it (@ade10176?).

ade10176:
Thanks @jcfranco I think the relationships are a bit confusing in the `none` selection mode. I'd expect all children to be aligned regardless whether they have a grand child or not. Are you able to clarify if that's by design or a bug @jcfranco?

jcfranco:
@macandcheese @ashetland WDYT about ☝️? If we did want to preserve alignment, we can use the blank icon for none-selection. Luckily, we have a [old commit with this](https://github.com/Esri/calcite-components/pull/5128/commits/4b17d9ffe184cde7ba19388022251d636cc8c6dc) that we can reapply.

macandcheese:
Agree w @ade10176 on this one :)

ashetland:
Agree w/ @ade10176 as well, alignment should be preserved. 😁

jcfranco:
[...]

PR #5292, opened 2022-09-08
fix(tree-item): preserve the alignment of tree-items with selection-mode: none regardless of whether they have children

Elijbet:
**Related Issue:** #5186

## Summary

Preserve the alignment of the `tree-items` regardless of whether they have children: outdent the disclosure arrows.

### DIFF
--- a/src/components/tree-item/tree-item.scss
+++ b/src/components/tree-item/tree-item.scss
@@ -7,7 +7,7 @@
 
 :host([selection-mode="none"]:not([has-children])) {
   .node-container {
-    margin-inline: theme("margin.3");
+    margin-inline: theme("margin.2");
   }
 }
 
--- a/src/components/tree-item/tree-item.scss
+++ b/src/components/tree-item/tree-item.scss
@@ -149,7 +149,6 @@
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/components/tree-item/tree-item.scss
```diff
@@ -155,6 +149,19 @@
   }
 }
 
+// selection-mode none, dropdown without children
```
jcfranco, line 152: IMO, the selector is readable, so we can drop this comment. Also, there's no dropdown here, right?

src/components/tree/tree.stories.ts
```diff
@@ -61,6 +61,22 @@ export const simple = (): string => html`
   </calcite-tree>
 `;
 
+export const selectionModeNoneScales = stepStory(
```
jcfranco, line 64: Per our convo, can you replace this with a single test-only story that doesn't set scale (default)?

jcfranco (approved):

jcfranco:
Adding @macandcheese for another pair of 👀 on this.

macandcheese (approved):
Looks good! - cc @ashetland for any necessary Figma changes.

PR #5299, opened 2022-09-12
fix: remove unsafe-eval to support CSP script-src

benelan:
**Related Issue:** #5104

## Summary
@dasa noticed this while we were looking into an unrelated JSAPI issue. He suggested I use the [`isBrowser` check the JSAPI copied from dojo](https://devtopia.esri.com/WebGIS/arcgis-js-api/blob/4master/esri/core/has.ts#L229-L235) and that I add the `no-eval`, `no-implied-eval`, and `no-new-func` ESLint rules, which the JSAPI uses, so we don't run into this problem in the future. 

[More info from MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#unsafe_eval_expressions)

Edit:
I ran `npm run test:prerender` and there were no errors.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -73,6 +73,9 @@
     "lines-between-class-members": ["error", "always"],
     "no-multiple-empty-lines": ["error", { "max": 1 }],
     "no-unneeded-ternary": "error",
+    "no-eval": "error",
+    "no-implied-eval": "error",
+    "no-new-func": "error",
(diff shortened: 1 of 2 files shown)


jcfranco (approved):
LGTM! Bonus points if you can split up the linting changes as these are preventative and not specific to the fix.

Issue #4893, opened 2022-07-09
add numberingSystem property to components that display numbers


benelan:
### Description

We added a numberingSystem property to `calcite-input` for localization. We should also add it to all other components that display numbers out of the box.

### Acceptance Criteria

New `numberingSystem` property that behaves like it does for `calcite-input`

### Which Component

- [x] calcite-alert (Dibs'd by Ben)
- [x] calcite-slider
- [x] calcite-stepper (Dibs'd by Ben)
- [x] calcite-date-picker (Dibs'd by Ben)
- [x] calcite-time-picker (Dibs'd by Ben)
- [x] calcite-input-time-picker (Dibs'd by Ben)
- [x] calcite-input  (follow up bug Dibs'd by Ben: #5252)
- [x] calcite-pagination

### Example Use Case

Example use case is in [`calcite-input`](https://github.com/Esri/calcite-components/blob/master/src/components/input/input.tsx). There is also relevant code in the [`locale`](https://github.com/Esri/calcite-components/blob/master/src/utils/locale.ts) util.

geospatialem:
Verified for all components above, with the exception of pagination. ✔️ 

Created #5648 for pagination-specific tracking.

### alert
![image](https://user-images.githubusercontent.com/5023024/198128320-76a981ef-f27e-4b73-9481-6cf3c49b47bd.png)

### slider
![image](https://user-images.githubusercontent.com/5023024/198128549-55ac6f6e-5b69-41f0-9ea3-9cfc71bb767c.png)

### stepper
![image](https://user-images.githubusercontent.com/5023024/198129030-0d2185b0-b983-4ea2-8237-73afac2ae9e1.png)

### date-picker
![image](https://user-images.githubusercontent.com/5023024/198129322-fe5ff777-3e7e-4786-9b30-75a69ebfb71a.png)

### time-picker
![image](https://user-images.githubusercontent.com/5023024/198129696-7b958702-71db-4587-bfb3-eebfc835e110.png)

### input-time-picker
![image](https://user-images.githubusercontent.com/5023024/198129764-d4a6aed6-6718-4e2a-8083-d2eed4fc5ed0.png)

### input
![image](https://user-images.githubusercontent.com/5023024/198129865-f8777469-b3d3-4d88-811d-7dd898fb2aa2.png)

PR #5301, opened 2022-09-13
feat(time-picker, input-time-picker): add numberingSystem property

benelan:
**Related Issue:** #4893

## Summary
Adds `numberingSystem` property to Time Picker and Input Time Picker.


src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -234,7 +242,12 @@ export class InputTimePicker
   @Watch("effectiveLocale")
   effectiveLocaleWatcher(): void {
     this.setInputValue(
-      localizeTimeString(this.value, this.effectiveLocale, this.shouldIncludeSeconds())
+      localizeTimeString(
+        this.value,
+        this.effectiveLocale,
+        this.numberingSystem,
+        this.shouldIncludeSeconds()
+      )
```
eriklharper, line 250: Wonder if it would be appropriate to turn the arguments here into key/value pairs as an object for readability?

src/utils/time.ts
```diff
@@ -154,22 +163,31 @@ export function localizeTimePart(value: string, part: TimePart, locale: string):
...
+  const formatter = createLocaleDateTimeFormatter(locale, numberingSystem);
   const parts = formatter.formatToParts(date);
   return getLocalizedTimePart(part, parts);
 }
 
-export function localizeTimeString(value: string, locale = "en", includeSeconds = true): string {
+export function localizeTimeString(
+  value: string,
+  locale = "en",
+  numberingSystem: NumberingSystem,
```
eriklharper, line 174: ☝️ should we initialize this to `defaultNumberingSystem` here as well?

benelan: Nice catch, it should actually be using `getSupportedNumberingSystem()` instead using a default value. I changed the time util before writing that function for the Input PR. I'll fix that.

src/components/input-time-picker/input-time-picker.tsx
benelan: I'll probably leave it as is for now. We will probably end up refactoring and removing the function in the long run to cache the formatter; same as I just did when creating `NumberStringFormat`. So no reason to clean it up twice.

src/utils/time.ts
eriklharper: Yup, totally.  Does that mean the other utils should use `getSupportedNumberingSystem` also?

src/components/input-time-picker/input-time-picker.tsx
eriklharper: Sounds fine to me :)

src/components/input-time-picker/input-time-picker.e2e.ts
```diff
@@ -244,7 +244,7 @@ describe("calcite-input-time-picker", () => {
     date.setSeconds(59);
 
     const expectedValue = "13:59:59";
-    const expectedDisplayValue = localizeTimeString(expectedValue, lang);
+    const expectedDisplayValue = localizeTimeString(expectedValue, lang, "latn");
 
     inputTimePicker.setProperty("value", expectedValue);
```
eriklharper, line 250: Might be good to also add at least 1 arab numberingSystem test here as well?

src/utils/time.ts
benelan: Yeah I'll go through all the time utils and make sure they are using a supported numbering system

src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -498,6 +539,8 @@ export class InputTimePicker
             disabled={this.disabled}
             icon="clock"
             label={getLabelText(this)}
+            locale={this.locale}
```
jcfranco, line 542: `locale` is deprecated. Can you use `lang` instead?

src/utils/time.ts
jcfranco: 👍 on using params object. Once a method goes beyond 2 or 3 params, it's a good time to think about switching the signature.

src/components/input-time-picker/input-time-picker.e2e.ts
jcfranco: Agreed.

jcfranco (approved):
😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎
😎👍😎😎😎👍😎👍😎😎😎👍😎👍😎😎😎👍😎😎😎👍😎😎😎👍😎👍👍👍😎😎👍👍👍😎👍👍👍👍😎👍😎
😎👍👍😎👍👍😎👍😎😎😎👍😎😎👍😎👍😎😎😎😎👍👍😎😎👍😎😎👍😎😎👍😎😎😎😎👍😎😎😎😎👍😎
😎👍😎👍😎👍😎👍😎😎😎👍😎😎😎👍😎😎😎😎😎👍😎👍😎👍😎😎👍😎😎👍😎😎😎😎👍👍👍😎😎👍😎
😎👍😎😎😎👍😎👍😎😎😎👍😎😎😎👍😎😎😎😎😎👍😎😎👍👍😎😎👍😎😎👍😎😎😎😎👍😎😎😎😎😎😎
😎👍😎😎😎👍😎😎👍👍👍😎😎😎😎👍😎😎😎😎😎👍😎😎😎👍😎👍👍👍😎😎👍👍👍😎👍👍👍👍😎👍😎
😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎

src/utils/time.ts
benelan: Cool I went ahead and changed the signatures and used `getSupportedNumberingSystem()` instead of a default value

src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -260,15 +273,25 @@ export class InputTimePicker
...
+    };
+
+    const delocalizedValue = numberStringFormatter.delocalize(calciteInputEl.value);
+
+    const localizedInputValue = localizeTimeString({
+      value: delocalizedValue,
+      includeSeconds: shouldIncludeSeconds,
+      locale,
+      numberingSystem
+    });
```
eriklharper, line 291: This looks SOO good, thank you!

eriklharper (approved):

PR #5302, opened 2022-09-13
ci: replace pify with promisify from node standard library

benelan:
**Related Issue:** NA

## Summary
Ran into errors in `prepReleaseCommit` while trying to release:

image

@jcfranco ran into the same errors with the `checkSquashMergeableBranch` earlier today. Not sure why it suddenly started happening, but the error is coming from `pify`. I replaced `pify` with Node's standard library `promisify` util which fixed the issue. They do the same thing for our use case, here are the [differences according to `pify`](https://github.com/sindresorhus/pify#faq)


support/prepReleaseCommit.ts
```diff
@@ -82,7 +81,7 @@ import yargs from "yargs";
     // we keep track of `beta` and `next` releases since `standard-version` resets the version number when going in between
     // this should not be needed after v1.0.0 since there would no longer be a beta version to keep track of
     const targetDescendingOrderTags = semverTags.filter((tag) => targetVersionPattern.test(tag)).sort(semver.rcompare);
-    const targetReleaseVersion = semver.inc(targetDescendingOrderTags[0], "prerelease", target);
+    const targetReleaseVersion = semver.inc(targetDescendingOrderTags[0], "prerelease", target) || "";
```
jcfranco, line 84: Sidebar: maybe we should throw here instead of falling back to `""` to make it clear that the target version is invalid.

jcfranco (approved):
Awesome! ✨🔧✨

TIL Node has a `promisify` util. Great find!

Issue #3489, opened 2021-11-10
Enhancement: [_animation.scss] add reduced motion support for .calcite-animate global CSS class


caripizza:
### Description
Add [`prefers-reduced-motion`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion) CSS media feature for the [`.calcite-animate`](https://github.com/Esri/calcite-components/blob/master/src/assets/styles/_animation.scss#L48) global CSS class. This will ensure animations only apply when the prefers-reduced-motion media feature matches `no-preference` on a system (ie., the class will only animate if a user hasn't enabled "Reduce motion" on their system).

Proposed change (courtesy [@tim-white-esri](https://github.com/Esri/calcite-components/issues/3081#issuecomment-960101416)):
```scss
// because default state is opacity: 0; need to make sure that is explicitly set to 1
@media (prefers-reduced-motion: reduce) {
    .calcite-animate {
        opacity: 1;
        animation: none;
    }
}

@media (prefers-reduced-motion: no-preference) {
    .calcite-animate {
        opacity: 0;
        animation-fill-mode: both;
        animation-duration: var(--calcite-animation-timing);
    }
}

```

### Acceptance Criteria 
The src/assets/styles/_animation.scss global stylesheet includes the new CSS media features (above snippet) for the .calcite-animate class.

### Relevant Info 
See pr #2607 related file changes for helpful information on using the global class.

#### Which Component
n/a (global class)

#### Example Use Case
n/a

tim-white-esri:
Any traction on this? I just received another complaint from a site visitor about animations not respecting reduced motion. 

This is important for accessibility.

geospatialem:
@tim-white-esri Thanks for re-pinging this issue. We'll be adding this to our upcoming sprint. For a workaround [Codepen](https://codepen.io/jcfranco/pen/JjvjpOV) in the meantime.

geospatialem:
The solution will update the internal duration factor based on user preferences. This will still give developers the flexibility to override the animation, but would not be supported by Calcite.

geospatialem:
Experiencing some unexpected behavior in Windows with the `loader` component in `beta.95` when animations are disabled. The expectation is no animation is present when users set the system preferences to disable animations.

[...]

PR #5314, opened 2022-09-13
feat(animation): support reduced motion preference for animations

anveshmekala:
**Related Issue:** #3489 

This PR will add support for the `prefers-reduced-motion` media feature for animations across Calcite Components and user's leveraging global animation classes can experience reduced motion when enabled.


src/assets/styles/_animation.scss
```diff
@@ -71,3 +71,16 @@
...
+@media (prefers-reduced-motion: reduce) {
+  .calcite-animate {
+    opacity: 1;
+    animation: none;
+  }
+}
+
+@media (prefers-reduced-motion: reduce) {
+  :root {
+    --calcite-internal-duration-factor: 0.01s;
```
anveshmekala, line 84: Once [5206](https://github.com/Esri/calcite-components/issues/5206) is installed, we can refactor this to be `0s`

src/assets/styles/_animation.scss
```diff
@@ -71,3 +71,16 @@
 .calcite-animate__in-scale {
   animation-name: in-scale;
 }
+
+@media (prefers-reduced-motion: reduce) {
+  .calcite-animate {
+    opacity: 1;
+    animation: none;
```
jcfranco, line 78: Is this needed if the duration factor is modified below? If truly needed, can you move this into the block below?

jcfranco (approved):
Nice! 🎉🌮🦖🦃

I just had a question about possible cleanup.

src/assets/styles/_animation.scss
anveshmekala: This got me thinking and I don't think any component is leveraging animation global classes including `calcite-animate` class. We may not even require this.

macandcheese: Consumers do - really we could as well. I'd imagine this comes from `calcite-styles` at some point instead of from CC. It does seem like we should have "from left" and "from right" animations as well...

anveshmekala: Makes sense. I updated the media query as per your suggestion.

macandcheese (approved):
Looks good, I am not sure if reduce motion should apply to alert dismissal timer since there are associated dismissal intervals that may be separate from the animation in / out of the component on the page (which correctly seems to have the animation duration on container div)

Aside from verifying that, these seem like good changes!

anveshmekala:
> Looks good, I am not sure if reduce motion should apply to alert dismissal timer since there are associated dismissal intervals that may be separate from the animation in / out of the component on the page (which correctly seems to have the animation duration on container div)
> 
> Aside from verifying that, these seem like good changes!

Agree that progress bar shouldn't reduce motion. Updated.

Issue #5318, opened 2022-09-14
Tooltip does not work with disabled button


mpayson:
### Actual Behavior

Tooltip does not show up on hover when its target button is disabled

### Expected Behavior

Tooltip shows up like the doc example
<img width="490" alt="Screen Shot 2022-09-14 at 11 52 25 AM" src="https://user-images.githubusercontent.com/5892789/190202900-54b9f3e8-f18e-4777-a4f1-973b104c2c6e.png">

#### Additional context from #6461 
When a disabled button contains information available by tooltip, using only aria-disabled communicates the disable state while permitting keyboard focus to be moved to the button: https://codepen.io/FelFly/pen/gOdYQQq

### Reproduction Sample

- https://codepen.io/mpayson/pen/poVNZPX?editors=1000
- https://codepen.io/FelFly/pen/RwYbqqe

### Reproduction Steps

1. Open codepen
2. Hover over button, notice tooltip does not display

If `disabled` prop is removed, it works

### Reproduction Version

beta.93

driskull:
@mpayson For this, I think you would need to wrap the button in a div so that you can add the tooltip to that div instead.

Disabled elements don't fire hover events so we can't really do anything here. Its the same if you change <calcite-button> to <button>.

mpayson:
Ah that makes sense, thanks for clarifying!

driskull:
@mpayson actually, we may be able to do something here. I'll see if we can but for now wrapping it would be a good solution.

driskull:
@benelan @jcfranco can we use this issue as an area for discussion on if we should remove the `pointer-events:none` on disabled `calcite-button`?

benelan:
Sure, @macandcheese do you have any design thoughts?

jcfranco:
@benelan FYI, there's some initial discussion in the [related PR](https://github.com/Esri/calcite-components/pull/5320#issuecomment-1247337691).

geospatialem:
Added to the `Stalled` milestone to discuss and prioritize after January's release.

yannikmesserli:
Note that we encountered the exact same problem for Calcite Radio Group's items. When disabled, the tooltip explaining why it was disabled was not showing up. Should I open a new issue for it? Or is the discussion you are having (soon?) will include the `pointer-event: none` removal on any disabled Calcite components?

geospatialem:
[...]

PR #5320, opened 2022-09-14
refactor(tooltip): Use pointer events in favor of mouse events

driskull:
**Related Issue:** #5318

## Summary

refactor(tooltip): Use pointer events in favor of mouse events. (#5318)

pointer events will fire on disabled elements.

### DIFF
--- a/src/components/tooltip/TooltipManager.ts
+++ b/src/components/tooltip/TooltipManager.ts
@@ -97,8 +97,8 @@ export default class TooltipManager {
 
   private addListeners(): void {
     document.addEventListener("keydown", this.keyDownHandler);
-    document.addEventListener("mouseover", this.mouseEnterShow, { capture: true });
-    document.addEventListener("mouseout", this.mouseLeaveHide, { capture: true });
+    document.addEventListener("pointerover", this.mouseEnterShow, { capture: true });
+    document.addEventListener("pointerout", this.mouseLeaveHide, { capture: true });
     document.addEventListener("pointerdown", this.clickHandler, { capture: true });
     document.addEventListener("focusin", this.focusShow, { capture: true });
     document.addEventListener("focusout", this.blurHide, { capture: true });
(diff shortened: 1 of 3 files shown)


driskull:
@jcfranco I think these code changes make sense. However, because we disable pointer events on our disabled elements the tooltips will still not show. This does fix the issue for disabled native buttons though.

test

@jcfranco Do you think we should try and support tooltips on our disabled elements?

jcfranco:
> Do you think we should try and support tooltips on our disabled elements?

@darrenDoesntCode @ashetland @SkyeSeitz @geospatialem WDYT?

ashetland:
> > Do you think we should try and support tooltips on our disabled elements?
> 
> @darrenDoesntCode @ashetland @SkyeSeitz @geospatialem WDYT?

Definitely useful, especially if the tooltip tell me why it's disabled.

macandcheese:
Hm, I'm not sure I would expect a disabled component to be interactive at all. I might expect another, not disabled UI element to invoke the tooltip. There are likely screenreader implications here to consider.

geospatialem:
Tested with NVDA and JAWS, and the disabled context isn't provided to them, which is the expectation with assistive technology. 

However, if implemented, this might be a case where we're providing content to some users, but not all. Not ideal to accommodate a larger audience.  

As Adam mentioned there could be another, non-disabled UI element that could invoke the tooltip for these use cases instead.

driskull:
The native button and native tooltip still work when the button is disabled.

Screen Shot 2022-09-14 at 3 35 50 PM

https://codepen.io/driskull/pen/yLjVWVK?editors=1000

geospatialem:
> The native button and native tooltip still work when the button is disabled.

Ah, thanks for the context! If that's the native behavior, then makes sense to proceed. 👍🏻

ashetland:
> The native button and native tooltip still work when the button is disabled.
> 
> Screen Shot 2022-09-14 at 3 35 50 PM
> 
> https://codepen.io/driskull/pen/yLjVWVK?editors=1000

I knew I had seen this somewhere!

macandcheese:
I still am not sure the tooltip should be invoked. Sometimes tooltips are just the name of button / action, but they are also used to contain descriptive or supplementary text, and only making that available via hover (and not focus, or click) seems weird.

I think it’s valid to maintain the native functionality of showing a label on hover, but to me a label and tooltip are not totally the same.

jcfranco:
> I think it’s valid to maintain the native functionality of showing a label on hover, but to me a label and tooltip are not totally the same.

@macandcheese Agreed, but that sounds like that will be up to the developers since we wouldn't use title + `calcite-tooltip` (**a**), but rather just `calcite-tooltip`s for both use cases (**b**), right? If we're all in agreement w/ **b**, this PR is good to go!

BTW, I'm assuming you used label to mean native tooltip, but please correct me if you meant something else.

driskull:
I think this PR is good to go regardless as it just changes from mouse to pointer events.

If we want to remove the `pointer-events:none` on our disabled `calcite-button` that would be the part that allows our tooltips on our disabled buttons.

jcfranco:
Good point. Moving forward!

WRT allowing tooltips on disabled components, let me know if y'all prefer an issue or discussion.

jcfranco (approved):
🐭👈

Issue #4246, opened 2022-03-16
Bug: Missing types when building from a new clone


benelan:
### Actual Behavior

Some of the components have missing types when building from a new clone. For example, here are the correct files for `calcite-combobox`:
https://unpkg.com/browse/@esri/calcite-components@1.0.0-beta.78/dist/types/components/combobox/

And here is what it looks like when files are missing:
https://unpkg.com/browse/@esri/calcite-components@1.0.0-beta.79/dist/types/components/combobox/

### Expected Behavior

No missing files when building

### Reproduction Sample

N/A

### Reproduction Steps

1. `git clone https://github.com/Esri/calcite-components.git`
2. `cd calcite-components`
3. `npm ci && npm run build`
4. Check for missing files in `dist/types/components/combobox`

### Reproduction Version

beta.79

### Relevant Info

There are no errors in the console. The work around is to build again and make sure all the files are there.

I repro'd on WSL (Ubuntu 20.04)
@jcfranco repro'd on Mac

Zay:
I can confirm that 4 type definitions are missing. 
I used https://github.com/Esri/calcite-components-examples/angular (currently on beta-78)

and ran commands:
- npm install
- ng build
- ng serve

This works fine.

Then upgraded to version "1.0.0-beta.79" using 
npm install @esri/calcite-components@latest

Getting the following errors on ng serve now:
ERROR in node_modules/@esri/calcite-components/dist/types/components.d.ts:19:61 - error TS2307: Cannot find module './components/combobox/interfaces'.

19 import { ComboboxChildElement, ComboboxSelectionMode } from "./components/combobox/interfaces";
                                                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
node_modules/@esri/calcite-components/dist/types/components.d.ts:28:34 - error TS2307: Cannot find module './components/pagination/pagination'.

28 import { PaginationDetail } from "./components/pagination/pagination";
                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
node_modules/@esri/calcite-components/dist/types/components.d.ts:29:29 - error TS2307: Cannot find module './components/pick-list/shared-list-logic'.

29 import { ListFocusId } from "./components/pick-list/shared-list-logic";
                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #5323, opened 2022-09-14
chore: build prior to stencil test to fix missing types error

benelan:
**Related Issue:** #4246

## Summary

We figured out that the issue linked above is due to a missing `components.d.ts` file, which just happens to be true with fresh clones. This is a stencil bug which is linked to our issue. Building twice or commiting the `components.d.ts` file are the two options for fixing the problem.

This PR will prevent the missing types issue for releases by "double building" without adding additional time, since I am creating a production build and then using that for the tests. After the additional build at the end of the `release:prepare` script, all of the types will be correct.

https://github.com/Esri/calcite-components/blob/e6c3ec661d4132549c4c566d323471d491e75627/package.json#L40

Edit:
The `build` script runs `util:copy-icons` which is why I was able to remove it from `util:run-tests`


jcfranco (approved):
Clever! 💪🧠

To prevent regressions, is there a way we can express the intent of these changes in this script or elsewhere?

Issue #4802, opened 2022-06-24
calcite-card selection checkbox can obscure title slot content


mjuniper:
### Actual Behavior

When I use `calcite-card` with `selectable` _and_ have content in the title slot that sits at the right side of the card, that content is obscured by the selection checkbox.

### Expected Behavior

The selection checkbox will not obscure content.

### Reproduction Sample

https://codepen.io/mjuniper/pen/KKQOeBq?editors=1100

### Reproduction Steps

1. Visit https://codepen.io/mjuniper/pen/KKQOeBq?editors=1100
2. Observe that the selection checkbox partially obscures the "Card Title" text.

### Reproduction Version

beta.83

nick-esri:
Field Maps web is also affected by this bug:
<img src="https://user-images.githubusercontent.com/3908500/185446752-58ae509a-b180-4565-a0a7-cdace68591fa.png" width="300"> 

Ideally the title should wrap before it collides with the checkbox. Possible to prioritize this fix @benelan ? Thanks

geospatialem:
Verified on beta.95
![verified card select padding](https://user-images.githubusercontent.com/5023024/193696560-01bf72d2-48dd-4401-b834-86d457ba8475.png)

PR #5325, opened 2022-09-14
fix(card): prevent header from overlapping with checkbox

jcfranco:
**Related Issue:** #4802 

## Summary

This adds some padding to the header when the card is selectable to prevent overlap with the checkbox.


src/components/card/card.scss
```diff
@@ -69,6 +69,13 @@
     font-medium;
 }
 
+:host([selectable]) {
+  .header {
+    // prevents overlap with checkbox
+    padding-inline-end: calc(theme("spacing.3") + theme("spacing.5"));
```
anveshmekala, line 75: Nitpick : can we replace `calc` with `theme("spacing[8.5]")`

anveshmekala (approved):
🚀

src/components/card/card.scss
jcfranco: Yeah, I think I'll replace that with a comment to explain that it's the heading + no-overlap spacing.

src/components/card/card.stories.ts
```diff
@@ -235,6 +235,21 @@ export const thumbnailRounded = (): string => html`
   </div>
 `;
 
+export const headerDoesNotOverlapWithCheckbox_TestOnly = html`
```
benelan, line 238: Take 3:

```js
export const headerDoesNotOverlapWithCheckbox_TestOnl = (): string => html`
```

jcfranco: Thanks for catching this.

Issue #5206, opened 2022-08-23
OpenCloseComponents do not emit when setting `--calcite-duration-factor`


jcfranco:
### Actual Behavior

Setting `--calcite-duration-factor` to 0 allows users to skip animations, but prevents (before)open/close events from being emitted.

### Expected Behavior

All (before)open/close events to be emitted regardless of transition duration.

### Reproduction Sample

https://codepen.io/jcfranco/pen/MWVReVQ?editors=1111

### Reproduction Steps

Click on toggle and notice how events are not emitted in the console.

### Reproduction Version

@esri/calcite-components@1.0.0-beta.92

geospatialem:
Verified on `next` 💯 

![image](https://user-images.githubusercontent.com/5023024/206047015-f4657acf-f82e-4e90-85f5-a0d57ab4f610.png)

jcfranco:
https://github.com/Esri/calcite-components/issues/6582 demonstrated that this only addressed the issue with modal. I created a [separate issue](https://github.com/Esri/calcite-components/issues/6604) to address all open/close components.

Issue #4398, opened 2022-04-12
Ensure events are emitted on user-interaction and not programmatically


jcfranco:
### Description

This is a continuation of https://github.com/Esri/calcite-components/issues/2033. Non-change/input events are listed here.

---

### Actual Behavior

Some components emit events regardless of internal or external (user) modifications.

### Expected Behavior

Public events should emit from user-interaction and not when props are modified programmatically.

##### Component/Event candidates

- [x] `action-bar.tsx` https://github.com/Esri/calcite-components/pull/4275 (@anveshmekala )
  - [x] `66 this.calciteActionBarToggle.emit();`
- [x] `action-pad.tsx` https://github.com/Esri/calcite-components/pull/4275 (@anveshmekala)
  - [x] `52 this.calciteActionPadToggle.emit();`
- ~[ ] `combobox.tsx`~ (implements `openClose`)
  - ~[ ] `462 this.active ? this.calciteComboboxOpen.emit() : this.calciteComboboxClose.emit();`~
- [x] `dropdown.tsx`
  - [x] `314 this.calciteDropdownSelect.emit();` - – already emits on user interaction only @driskull
  - ~[ ] `386 this.active ? this.calciteDropdownOpen.emit() : this.calciteDropdownClose.emit();`~ (implements `openClose`)
- [x] `inline-editable.tsx`
  - [x] `280 this.calciteInlineEditableEditCancel.emit();` – already emits on user interaction only @jcfranco
- ~[ ] `input-date-picker.tsx`~ (implements `openClose`)
  - ~[ ] `530 ? this.calciteInputDatePickerOpen.emit()`~
  - ~[ ] `531 : this.calciteInputDatePickerClose.emit();`~
- ~[ ] `modal.tsx`~ (implements `openClose`)
  - ~[ ] `333 this.active ? this.calciteModalOpen.emit() : this.calciteModalClose.emit();`~
- ~[ ] `popover.tsx`~ @driskull @jcfranco (implements `openClose`)
  - [x] `388 this.open ? this.calcitePopoverOpen.emit() : this.calcitePopoverClose.emit();`

##### Things to look out for

- Events emitted in a prop watcher may need to be moved elsewhere to avoid firing when props are programmatically modified
- Internal events can be refactored in a later milestone, the priority is for public ones
- Deprecated events (current or new ones) will not be modified

#### Related issues

https://github.com/Esri/calcite-components/issues/3453
https://github.com/Esri/calcite-components/issues/3191
[...]

PR #5326, opened 2022-09-14
fix(modal): OpenCloseComponent emits when setting `--calcite-duration-factor` to 0

Elijbet:
**Related Issue:** #5206

All (before)open/close events are now emitted on modal regardless of transition duration.

Add the `onToggleUnanimatedComponent` function to the `openCloseComponent Interface` to query props on the `transitionEl`. Use that to get the `transition-duration` on the `openTransitionProp`, which in the case of modal it’s the `opacity`. This will allow us to emit the open/close and beforeOpen/beforeClose events in cases where `opacity` transition duration is set to  0.

Because calling computed style forces a layout, we use `readTask()` instead which schedules a DOM-read task. The provided callback will be executed at the best moment to perform DOM reads without causing layout thrashing as per Stencil docs. 

Updated to rework `openCloseComponent Interface` to call one function instead of 3 separate calls in every component by removing connect/disconnect and event listener mapping logic and passing argument `once` to `.addEventListener` on `transitionEl` instead.


src/components/modal/modal.e2e.ts
```diff
@@ -151,6 +151,31 @@ describe("opening and closing behavior", () => {
     expect(beforeOpenSpy).toHaveReceivedEventTimes(1);
     expect(openSpy).toHaveReceivedEventTimes(1);
   });
+
+  it("emits when setting --calcite-duration-factor to 0", async () => {
+    const page = await newE2EPage();
+    await page.setContent(`<calcite-modal active></calcite-modal>`);
```
jcfranco, line 157: This and future tests should use the new `open` property.

src/utils/openCloseComponent.ts
```diff
@@ -60,6 +61,26 @@ function transitionEnd(event: TransitionEvent): void {
     this.open ? this.onOpen() : this.onClose();
   }
 }
+
+/**
+ * Helper to determine globally set transition duration setting.
+ *
+ * @param component
+ */
+export function onToggleUnanimatedComponent(component: OpenCloseComponent): void {
```
jcfranco, line 70: Is there another name we could use here? I'm not sure what 'unanimated component' means in the context of open/close. 😅

src/utils/openCloseComponent.ts
```diff
@@ -60,6 +61,26 @@ function transitionEnd(event: TransitionEvent): void {
...
+  readTask((): void => {
+    if (component.transitionEl) {
+      const tokens: string[] = getComputedStyle(component.transitionEl).transition.split(" ");
```
jcfranco, line 73: Let's rename this to better express what it's storing. I also think you can skip the type since TS will infer from the array's return type. This last comment applies to other variable declarations.

src/components/modal/modal.tsx
```diff
@@ -419,6 +420,7 @@ export class Modal implements ConditionalSlotComponent, OpenCloseComponent {
   @Watch("open")
   async toggleModal(value: boolean): Promise<void> {
     this.active = value;
+    onToggleUnanimatedComponent(this);
```
jcfranco, line 423: This will have to be called on all open/close components.

On a related note, it'd be great if we had a single way for handling both the duration > 0 and duration === 0 cases. We can leave this as a future refactor to not hold this up, but in that case, let's make it clear in the util's doc that it is required and where it should be called.

src/utils/openCloseComponent.ts
```diff
@@ -60,6 +61,26 @@ function transitionEnd(event: TransitionEvent): void {
     this.open ? this.onOpen() : this.onClose();
   }
 }
+
+/**
+ * Helper to determine globally set transition duration setting.
```
jcfranco, line 66: This needs more info on the why and how it should be used.

src/utils/openCloseComponent.ts
```diff
@@ -60,6 +61,26 @@ function transitionEnd(event: TransitionEvent): void {
...
+ *
+ * @param component
+ */
+export function onToggleUnanimatedComponent(component: OpenCloseComponent): void {
+  readTask((): void => {
+    if (component.transitionEl) {
```
jcfranco, line 72: What's the scenario where this method would be called and `transitionEl` is not available? And how does the method get called again once `transitionEl` is available?

src/components/modal/modal.e2e.ts
```diff
@@ -64,9 +64,9 @@ describe("calcite-modal properties", () => {
 });
 
 describe("opening and closing behavior", () => {
-  it.skip("opens and closes", async () => {
+  it("opens and closes", async () => {
     const page = await newE2EPage();
-    await page.setContent(`<calcite-modal></calcite-modal>`);
+    await page.setContent(`<calcite-modal style="transition: opacity 0.3ms"></calcite-modal>`);
```
driskull, line 69: Cool! can you use the css variable we have for setting the transition instead of setting the style directly?

src/components/modal/modal.tsx
```diff
@@ -134,6 +130,7 @@ export class Modal
     setUpLoadableComponent(this);
     // when modal initially renders, if active was set we need to open as watcher doesn't fire
     if (this.open) {
+      onToggleOpenCloseComponent(this);
```
driskull, line 133: Calling this here will emit these events on load of the component. Is there some history on why we are emitting on load here? Usually for components, we don't emit events initially, we start emitting after load on user interaction.

If this is intentional, the test will need to be updated to expect the event to be called initially.

src/components/modal/modal.tsx
```diff
@@ -134,6 +130,7 @@ export class Modal
     setUpLoadableComponent(this);
     // when modal initially renders, if active was set we need to open as watcher doesn't fire
     if (this.open) {
```
alisonailea, line 132: there is the prop "open" and the state "isOpen" how are these different and can we choose just one?

alisonailea (changes requested):
There are some issues and questions to resolve here.

src/components/modal/modal.tsx
Elijbet: The discussion on this is moved to #4398.

src/components/modal/modal.tsx
Elijbet: `isOpen` was used in conjunction with the `active `prop, which has been deprecated in favor of `open`. It’s still used in other components as well. It would be out of scope for this issue, but I can look at it separately in another one to determine whether we need to keep it or refactor out.

driskull (approved):
Looks good @Elijbet !! Nice work!!! 🎉

alisonailea (approved):

Issue #5329, opened 2022-09-15
Panel: Escape key closes panel even when closable is true


driskull:
### Actual Behavior

Panel: Escape key closes panel even when closable is true

### Expected Behavior

Panel: Escape key should only close panel if the panel is closable

### Reproduction Sample

https://codepen.io/driskull/pen/MWGJrga?editors=1010

### Reproduction Steps

1. Open codepen
2. Focus on panel
3. Hit esc key

### Reproduction Version

Latest

driskull:
@benelan @jcfranco can we triage this one?

jcfranco:
Triaged! 🚀

geospatialem:
Verified in beta.95

PR #5330, opened 2022-09-15
fix(panel): Do not close non closable panel on ESC key

driskull:
**Related Issue:** #5329

## Summary

fix(panel): Do not close non closable panel on ESC key. (#5329)


jcfranco (approved):
💥

Issue #3682, opened 2021-12-10
Bordered Calcite Tabs does not allow for long tab titles


tyanthony:
### Actual Behavior

Calcite `Tabs` - `bordered`, long tab titles overflow. This happens only with bordered Tabs.

### Expected Behavior

Long tab titles should not overflow and the tabs should allow for titles of any length (same as non-bordered tabs).

### Reproduction Steps and Sample

1. Implement Calcite bordered `Tabs` with at least 2 tabs.
2. The second tab has a long title (at least longer than the first)
3. Notice that the tab title overflows

_Sample_:
https://codepen.io/solowt/pen/poWEWJq?editors=1000

### Relevant Info

_Reproducible version_: 
`@esri/calcite-components@1.0.0-beta.69` (our app version)
`@esri/calcite-components@1.0.0-beta.71` (codepen sample version)
Seeing the bug in multiple versions of Tabs

- [ ] CDN
- [x] NPM package

macandcheese:
![](https://github.trello.services/images/mini-trello-icon.png) [Tab Border Issue](https://trello.com/c/UqOzhCAJ/38-tab-border-issue)

paulcpederson:
@macandcheese I ran into a similar issue. Not sure if it's the same. Basically when using layout centered, the overflow will chop off the first part of the tab. https://codepen.io/paulcp/pen/rNJqbZO Should I open another issue for this centered case, or is this covered here?

macandcheese:
Thanks for adding the issue, I imagine they are related but when this is picked up, let’s confirm it solves as well.

Cc @benelan maybe this could be prioritized? Likely a css fix

macandcheese:
This was resolved by https://github.com/Esri/calcite-components/pull/4714


Example: https://codepen.io/mac_and_cheese/pen/qBoPNba?editors=1000

paulcpederson:
@macandcheese there was a regression in beta 91 and the centered tabs are broken again. https://codepen.io/paulcp/pen/rNJqbZO

If I use beta.86 this codepen works great, but if I use anything 91 and newer it's broken. 

Not sure what caused this regression. We can open a new issue too, if that's easier.

Elijbet:
There are several problems with tabs raised in this issue with regard to regression.

Part 1: Calcite Tabs confirmed to allow for long tab titles in the current release:
https://codepen.io/elijbet/pen/zYjNpwz?editors=1010
Part 2: Is addressed in the PR.
[...]

PR #5332, opened 2022-09-15
fix(tab-nav, tab-title): centered layout allows to scroll to the extent

Elijbet:
**Related Issue:** #3682

## Summary

The centering currently is achieved by applying justify-content: center to the flex container.
However, there is a problem with this method when the flex item is bigger than the flex container.
For horizontal overflow, the left section becomes inaccessible (or the right section, in RTL languages).

https://codepen.io/elijbet/pen/dyeNJZj

Solution: With auto margins, an overflowing flex item can be horizontally centered without losing access to any part of it.

The result, however, looks a bit different. So I’d like to feedback on whether I can keep the solution. 

Current Centering:
Screen Shot 2022-09-15 at 3 36 17 PM

Suggested Centering:
Screen Shot 2022-09-15 at 3 36 58 PM


ashetland:
Good to go with this change. It actually corresponds with how it's currently built in Figma. 🥳

macandcheese (approved):
Looks good!

PR #5062, opened 2022-08-01
ci: check generated types before publishing

jcfranco:
**Related Issue:** N/A

## Summary

Adds a util script for testing types. This can be wired into our CI to prevent deploying if we know the types have issues.

PR #5335, opened 2022-09-16
fix(action-bar, combobox, dropdown, filter, input-date-picker, popover, tooltip): make methods using debounce/throttle private to prevent lodash type errors

benelan:
**Related Issue:** NA

## Summary
@dasa reported a `lodash` type error from `calcite-tooltip` when bumping the JSAPI to `beta.94`. You can see lodash being imported in the types on line 59 of [`tooltip.d.ts`](https://unpkg.com/browse/@esri/calcite-components@1.0.0-beta.94/dist/types/components/tooltip/tooltip.d.ts).

Making the method private prevents the generated interface member from having a type defined. 

Edit:
Looks like there were additional components besides `tooltip`:
image

I added a bash command to Franco's type testing PR #5062 so this won't happen again 
```bash
! grep -rnw 'dist/types' -e '<reference types='
```

### DIFF
--- a/src/components/combobox/combobox.tsx
+++ b/src/components/combobox/combobox.tsx
@@ -452,7 +452,7 @@ export class Combobox
   //
   // --------------------------------------------------------------------------
 
-  debouncedReposition = debounce(() => this.reposition(), repositionDebounceTimeout);
+  private debouncedReposition = debounce(() => this.reposition(), repositionDebounceTimeout);
 
   setFilteredPlacements = (): void => {
     const { el, flipPlacements } = this;
--- a/src/components/combobox/combobox.tsx
+++ b/src/components/combobox/combobox.tsx
@@ -736,7 +736,7 @@ export class Combobox
     return [...this.groupItems, ...this.items];
   }
 
-  filterItems = (() => {
+  private filterItems = (() => {
     const find = (item: ComboboxChildElement, filteredData: ItemData[]) =>
       item &&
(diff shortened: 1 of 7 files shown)


driskull (approved):

driskull:
@benelan do we need to do this for other components? There are others that do the same. `action-bar`, `dropdown`, `filter`, etc.

benelan:
Ah yes we do, I'll add those to this PR. We should try to make an ESLint rule somehow

src/components/combobox/combobox.tsx
```diff
@@ -776,14 +776,14 @@ export class Combobox
...
-  emitCalciteLookupChange = debounce(this.internalCalciteLookupChangeEvent, 0);
+  private emitCalciteLookupChange = debounce(this.internalCalciteLookupChangeEvent, 0);
 
   internalComboboxChangeEvent = (): void => {
     const { selectedItems } = this;
     this.calciteComboboxChange.emit({ selectedItems });
   };
 
-  emitComboboxChange = debounce(this.internalComboboxChangeEvent, 0);
+  private emitComboboxChange = debounce(this.internalComboboxChangeEvent, 0);
```
benelan, line 786: @driskull before I merge, making the event emitters private won't break types for our users right?

jcfranco: This should be safe. The function invoking the emitter is being made private, not the emitter itself.

Issue #4840, opened 2022-06-30
[Progress] Displayed `text` color not set / does not swap with theme class


macandcheese:
### Actual Behavior

It seems we aren't setting a defined text color on Progress `text` string display - which means it doesn't respond to light / dark mode changes and inherits color from a parent.

### Expected Behavior

Explicit text style is set and responds correctly to light / dark mode change.

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/KKowxGr?editors=1000

### Reproduction Steps

1 - Go to Codepen
2 - Notice provided `text` string is not swapping with theme
3 - Notice setting color style on parent component incorrectly overrides `text` string

### Reproduction Version

83

### Relevant Info

Longer term we could look at perhaps removing the `text` string altogether if we don't think it has value / gets use, similar to some of @asangma's proposed changes to Loader : https://github.com/Esri/calcite-components/issues/4470.

geospatialem:
Verified in `beta.95`. ✔️

PR #5336, opened 2022-09-17
fix(progress): text now responds to light/dark theme mode change

Elijbet:
**Related Issue:** #4840

## Summary

Sets the text style through a theme color variable declared in the global file, to enable theming to be applied and inherited throughout the component tree.


src/components/progress/progress.stories.ts
```diff
@@ -31,3 +31,14 @@ export const darkThemeRTL_TestOnly = (): string => html`
 `;
 
 darkThemeRTL_TestOnly.parameters = { themes: themesDarkDefault };
+
+export const darkThemeText = (): string => html`
```
jcfranco, line 35: Can you update the `darkThemeRTL_TestOnly` story above to cover your changes instead of adding a dedicated story? Since it's related to the dark theme, I think it'd make sense to set there. This would keep our snapshot quota down while covering the use case.

jcfranco (approved):
💥

PR #5339, opened 2022-09-19
feat: use pointer events instead of mouse events to improve device compatibility

jcfranco:
**Related Issue:** N/A

## Summary

Stems from https://github.com/Esri/calcite-components/pull/5320. This moves components to use pointer events for better device compatibility.


driskull (approved):
Nice! this is a good practice going forward!

Issue #5258, opened 2022-09-01
Date Picker: missing year and incorrect date order for "zh-CN" locale


benelan:
### Actual Behavior

An Esri Community user reported that the year is missing in the header of the Date Picker when using the "zh-CN" locale. I confirmed with @annierm18 that the year is missing, and she said the date's order is also incorrect. Additionally, the first day of the week is incorrect, but thats not on us ([CLDR issue](https://unicode-org.atlassian.net/browse/CLDR-11510)).

### Expected Behavior

The year should appear in the date picker header when using the "zh-CN" locale, and the date's order should be fixed (check in with @annierm18 for more info on this part if necessary).

### Reproduction Sample

https://codepen.io/benesri/pen/NWMWqzw?editors=1000

### Reproduction Steps

1. Learn how to read Chinese
2. Go to the repro sample and read the date

### Reproduction Version

beta.93

### Relevant Info

Here is the [Community post](https://community.esri.com/t5/calcite-design-system-questions/date-picker-issue-for-local-language-setting-quot/m-p/1208697#M212).

In version `beta.90` it looks like this:
![image](https://user-images.githubusercontent.com/10986395/188026997-bad100c6-2a2b-47a8-b4ae-b56de76bba7f.png)

#5047 and #5028 both made changes to Date Picker between `beta.90` and `beta.91` but niether PR jump out as the culprit.

### Regression?

beta.90

fendiaoxiaoshuazi:
It also happens on `ko` and `ja`.

geospatialem:
Verified in `next.601` for `zh-CN`, `ja`, and `ko`. 

Keeping the issue open for Figma change confirmation from @ashetland and @SkyeSeitz. 

### zh-CN
![image](https://user-images.githubusercontent.com/5023024/196741243-0e120246-9a49-47e1-9127-8b26ca0ec5d4.png)

SkyeSeitz:
Thanks, @geospatialem. This does not effect the Figma build of Date Picker.

geospatialem:
Thanks @SkyeSeitz, closing this issue. 👍🏻

PR #5344, opened 2022-09-19
fix(date-picker): no longer hides year for zh-CN locale

anveshmekala:
**Related Issue:** #5258 

This fix will unhide the year in date-picker component introduced as regression from `beta.84` version

A047343F-4A2D-4783-BEFA-52BE59C105D7_4_5005_c


src/components/date-picker-month-header/date-picker-month-header.tsx
```diff
@@ -152,14 +151,7 @@ export class DatePickerMonthHeader {
               type="text"
               value={localizedYear}
             />
-            {suffix && (
-              <span class="suffix">
-                <span aria-hidden="true" class="suffix__invisible">
```
anveshmekala, line 157: Wasn't sure why we have the `span` element with `visibility: hidden`

src/components/date-picker-month-header/date-picker-month-header.scss
```diff
@@ -117,22 +117,16 @@
...
+  @apply top-0
     left-0
-    w-16
     whitespace-nowrap
-    text-left;
-}
-
-.suffix__invisible {
-  @apply invisible;
+    text-right;
```
driskull, line 131: Should this be `text-right` or something that flips in RTL?

src/components/date-picker-month-header/date-picker-month-header.scss
```diff
@@ -117,22 +117,16 @@
 }
 
 .year--suffix {
-  @apply w-16 text-left;
+  @apply text-left;
```
driskull, line 120: Should this flip in RTL?

src/components/date-picker-month-header/date-picker-month-header.tsx
driskull: Looks like this span was here to hide it from screen readers. Seems like that functionality is removed here. Are we sure we can no longer hide this text from screen readers?

driskull (commented):
Looking good @anveshmekala . Just had some questions/comments!

src/components/date-picker-month-header/date-picker-month-header.scss
anveshmekala: nope, it doesn't need to flip.

src/components/date-picker-month-header/date-picker-month-header.scss
anveshmekala: same here,  it doesn't need to flip. The order has to respect the direction.

src/components/date-picker-month-header/date-picker-month-header.tsx
anveshmekala: The `span` element here is masking the `input` element which displays the year. If we want to hide the input element then this needs to be restored. Might be missing some context here.  @geospatialem  WDYT?

geospatialem: I'm hearing some invalid entries in the demos with NVDA. Wondering if the `aria-hidden` prop is used to hide those entries? Without this change, its registering as blank in the Arabic demos.

geospatialem: The comment above is when the screen reader is set to English. Once NVDA's language is set to the appropriate locale, the year number, or "year" text is read back to the screen reader. Great work, @anveshmekala! 👍🏻

anveshmekala:
Adding @macandcheese to review the changes as we drop the `span` element which masks the `input` field. This change moves the suffix of the year by `6px` towards the right. 

OG version before the bug :
188026997-bad100c6-2a2b-47a8-b4ae-b56de76bba7f

After the fix :
A047343F-4A2D-4783-BEFA-52BE59C105D7_4_5005_c

cc: @ashetland

jcfranco (approved):
Code changes LGTM!

PR #5346, opened 2022-09-19
ci(next): stop checking if changes pushed before publishing to npm

benelan:
**Related Issue:**

## Summary
Next hasn't actually deployed since 582. Looking at the error, it is failing a check to that makes sure the changes were pushed:
https://github.com/Esri/calcite-components/actions/runs/3050425472/jobs/4917508918#step:5:7366

However the following commit is the changes being pushed. According to the commit history we are on `next.590`. It seems like there is some lag happening between pushing and confirming that the push happened which is causing the deployment to fail right before publishing to NPM. So I'm removing that check. 

Hopefully it is an unnecessary check since we are using an atomic push which should provide a non-zero exit code of it doesn't work for any reason.

### DIFF
--- a/support/deployNextFromCI.ts
+++ b/support/deployNextFromCI.ts
@@ -27,11 +27,6 @@
     console.log(" - pushing tags...");
     await exec(`git push --atomic --follow-tags origin master`);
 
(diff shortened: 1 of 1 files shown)


driskull (approved):

jcfranco (approved):
💥

PR #5362, opened 2022-09-20
docs: update color-picker JSDoc link

jcfranco:
**Related Issue:** N/A

## Summary

Tidy up for color-picker's `value` doc.

This also works as a side-effect fix for a build error in Stencil v2.18. 🎉 cc @driskull

### DIFF
--- a/src/components/color-picker/color-picker.tsx
+++ b/src/components/color-picker/color-picker.tsx
@@ -264,12 +264,13 @@ export class ColorPicker implements InteractiveComponent {
   /**
    * The color value.
    *
-   * This value can be either a {@link https://developer.mozilla.org/en-US/docs/Web/CSS/color|CSS string}
+   * This value can be either a CSS color string,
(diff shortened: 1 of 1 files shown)


driskull (approved):
👍

src/components/color-picker/color-picker.tsx
```diff
@@ -264,12 +264,13 @@ export class ColorPicker implements InteractiveComponent {
   /**
    * The color value.
```
geospatialem, line 265: What would ya think about the following below, in line with #4442 for the `value` prop? 

> The ~color~ component's value.

geospatialem (commented):
Very nice! 🎨 Small consideration with the initial doc line, otherwise we can get this one in later, too.

src/components/color-picker/color-picker.tsx
driskull: I changed it.

PR #5364, opened 2022-09-20
chore: specify esri product team when logging issues

benelan:
**Related Issue:** #

## Summary
Adds a required dropdown field for specifying an esri product team to the bug/a11y/enhancement issue templates. A github action will parse the field and add a label for filtering.


jcfranco (approved):
yes

PR #5367, opened 2022-09-21
ci(next releases): check types before pushing tags

benelan:
**Related Issue:** #5062

## Summary
Reorganize the `next` release NPM scripts so that they are in line with the normal release ones. The push should happen in the `util:publish-next` script so that we don't run into an issue where a version pushes but doesn't get published to NPM.


jcfranco (approved):
💥

benelan:
The failed test is unrelated and one of the unstable ones, I'm going to merge. cc @jcfranco 

>   ● calcite-action-bar › overflow actions › should slot 'menu-actions' on resize of component

Issue #3668, opened 2021-12-09
Some components are missing RTL equivalent CSS


driskull:
### Actual Behavior

Some components are missing RTL equivalents for animation, box-shadow, gradients, etc.

- [x] alert
- [x] date-picker-day
- [x] date-picker-month-header
- [x] input-date-picker
- [x] loader
- [x] panel

### Expected Behavior

All components offer RTL support when these styles are used. 

Use logical properties for text-align, floats/positions/resize, and box model properties width/height and margin/padding/border to be set based on the language used. Provide RTL support wherever these are not available. 

### Relevant Info

Eg. some of the date box shadows use x-offset.
```
:host([start-of-range]) :not(.calcite--rtl),
:host([end-of-range]) .calcite--rtl {
  .day-wrapper {
    border-start-start-radius: 40%;
    border-end-start-radius: 40%;
    box-shadow: inset 4px 0 var(--calcite-ui-foreground-1);
  }
  .day {
    opacity: 1;
  }
}
```

_Reproducible version_:  v1.0.0-beta.94

- [x] CDN
- [x] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-<version>`

driskull:
@jcfranco can we maybe shelve this issue to the backburner? I think we might just wait to revisit these.

jcfranco:
I think we should at least look for any animation props that would look odd in RTL. The rest may be minor and worth tackling later. I'll defer to the UXperts™: @macandcheese @asangma

macandcheese:
@jcfranco @driskull Are there

> I think we should at least look for any animation props that would look odd in RTL. The rest may be minor and worth tackling later. I'll defer to the UXperts™: @macandcheese @asangma

Maybe we can open individual issues for each offense as we find them, like the date picker one provided?

jcfranco:
Sounds good. Let’s repurpose this one for the date picker and I’ll create separate issues for anything that came up in https://github.com/Esri/calcite-components/pull/3636#issuecomment-989262308.

Elijbet:
Actually, I decided to rework this issue to cover all the components that need physical positioning swapped with logical, and have multiple PRs instead of separate issues. I noticed that there are quite a bit of components just with box-shadow physical positioning not accounting for RTL.

geospatialem:
Trying to verify this across the components, but not seeing updates/changes across components between `beta.86` and `next` with `dir="rtl"`.

[...]

PR #5368, opened 2022-09-21
fix(alert, date-picker-month-header, input-date-picker, loader, panel): adds RTL equivalent CSS

Elijbet:
**Related Issue:** #3668

## Summary

Refactors to add support for RTL.


src/components/loader/loader.stories.ts
```diff
@@ -31,7 +31,7 @@ export const inline_NoTest = (): string => html`
 </div>
 `;
 
-export const customTheme_NoTest = (): string => html`
+export const customThemeAndRTL_NoTest = (): string => html`
```
jcfranco, line 34: This story seems unrelated to `dir` (custom styling + RTL). 

I notice that the simple test is skipped for tests. @eriklharper @benelan can you confirm if this is intentional? If so, is it because the story's animation causes instability? 

If this is not being tested, I think we could skip testing it in RTL for the same reasons. At least until we switch to Chromatic.

src/components/date-picker-month-header/date-picker-month-header.scss
```diff
@@ -117,7 +117,8 @@
 }
 
 .year--suffix {
-  @apply w-16 text-left;
+  @apply w-16
+  text-start;
```
jcfranco, line 121: According to https://github.com/Esri/calcite-components/pull/5344, this needs to stay the same. cc @anveshmekala

Elijbet: Thank you! I missed that.

src/components/loader/loader.stories.ts
benelan: We shouldn't be taking any snapshots of Loader because every time Screener runs, the Loader position changes due to the animation. In Screener's UI we ignored the component, so every time there were visual regressions it wouldn't show up as a change. But it still took the snapshot every time, which was a waste.

Chromatic may have a better way of handling animated components, I'll look into that once we migrate.

Elijbet: That's a great recap, thanks for the info. I'll remove the RTL test and check back with you when we bring in Chromatic to address it then.

jcfranco (approved):
🎉

PR #5370, opened 2022-09-21
feat(shell-center-row): add collapsed property

benelan:
**Related Issue:** https://devtopia.esri.com/WebGIS/calcite-components/issues/149

## Summary
Our first Tech Support ticket, and it's from NASA 🚀🚀 (literally)

Adds a `collapsed` property to `calcite-shell-center-row` which sets `display: none;` on the host el. This compliments the `collapsed` property for `calcite-shell-panel`.

#### Edit
I know aliasing a global attribute is kind of weird. FWIW it is temporary though, @macandcheese plans to deprecate `calcite-shell-center-row` in in favor of  adding `top` / `bottom` slots to go along with the `start` / `end` ones for Panels within Shell.


src/components/shell-center-row/shell-center-row.scss
```diff
@@ -27,6 +27,10 @@
   border-0;
 }
 
+:host([collapsed]) {
+  @apply hidden;
```
driskull, line 31: Should we just hide the internals of the component instead of the component itself? Because essentially this is just emulating the `hidden` attribute and they may not play nice together.

driskull (commented):
Just a question❓

src/components/shell-center-row/shell-center-row.scss
driskull: edit: Saw the edit that this is temporary. If its temporary then I'm ok with it but ideally we don't want to mess with the host in these cases

benelan: It is temporary, but I don't know how long the rework will take so I might as well do it right so nothing breaks until then.

src/components/shell-center-row/shell-center-row.e2e.ts
```diff
@@ -36,6 +40,18 @@ describe("calcite-shell-center-row", () => {
     expect(actionBarContainer).toBeNull();
   });
 
+  it("should not render when collapsed", async () => {
+    const page = await newE2EPage();
+    await page.setContent("<calcite-shell-center-row collapsed></calcite-shell-center-row>");
+
+    const element = await page.find("calcite-shell-center-row");
+
+    await page.waitForChanges();
```
jcfranco, line 49: This shouldn't be necessary as `setContent` calls `waitForChanges` internally.

src/components/shell-center-row/shell-center-row.scss
benelan: Sorry by the rework I meant @macandcheese thinks we should completely deprecate the `calcite-shell-center-row` component in favor of new top/bottom slots for `calcite-shell`. But idk how long that will take, I assume way past v1. I'm just going to close this PR, it only took a sec to put together and the users should just use `hidden`

Issue #4721, opened 2022-06-13
Slider handle misaligned if browser default font is adjusted


gsoosalu:
### Actual Behavior

Calcite slider's handle appears below or above the slider line: 
![image](https://user-images.githubusercontent.com/85623288/173317041-6b5ddd14-3429-4a8c-af7a-3211c425aa80.png)
![image](https://user-images.githubusercontent.com/85623288/173317270-2f6682ac-8bfe-4b26-9287-977d8a82e314.png)

### Expected Behavior

Slider's handle should be in the middle of the line
![image](https://user-images.githubusercontent.com/85623288/173317382-480081da-2304-4f7b-b52f-8d2c6446296f.png)

### Reproduction Sample

https://developers.arcgis.com/javascript/latest/sample-code/widgets-shadow-cast/live/

### Reproduction Steps

1. Change in browser setting the font size to either small or large. Screenshot from Chrome:
![image](https://user-images.githubusercontent.com/85623288/173316634-74f73bae-4b30-4bbd-9a87-5f082a8817d2.png)

1. Open JS API sample: https://developers.arcgis.com/javascript/latest/sample-code/widgets-shadow-cast/live/
2. Open color-picker
3. The opacity slider's handle is misaligned

This also appears when enabling  in the Calcite storybook: 
https://esri.github.io/calcite-components/?path=/story/components-controls-slider--single-value

### Reproduction Version

beta.82

### Relevant Info

MacOS and Windows
Chrome 102.0.5005.115
Firefox 101.0.1

driskull:
Seems like the current positioning of the handle is based on a fixed offset size. That would need to change to be relative instead of fixed.

geospatialem:
Hi @gsoosalu, thanks for reaching out. We plan to address to our February release.

geospatialem:
Verified on [Storybook](https://esri.github.io/calcite-components/?path=/story/components-controls-slider--simple) and in `1.0.8-next.1`.

![image](https://user-images.githubusercontent.com/5023024/221719945-1e621f51-e162-475a-bbe5-6dfbcab29643.png)

PR #5372, opened 2022-09-21
fix(slider): slider handle aligns with track when font size changes

anveshmekala:
**Related Issue:** #4721 

### Summary:
This PR will align slider thumb handles with track with respect to font-size of the browser.


src/components/slider/slider.stories.ts
```diff
@@ -279,3 +279,17 @@ export const wordBreakDoesNotAffectLabels_TestOnly = (): string =>
...
+  <head>
+    <style>
+      html {
+        font-size: 20px;
+      }
+    </style>
+  </head>
+  <body>
+    <calcite-slider label-handles min-value="50" max-value="70" range></calcite-slider>
+    <calcite-slider label-handles value="70"></calcite-slider>
```
anveshmekala, line 293: should we add all types of slider's  available to users in this test?

jcfranco (commented):
Is there a way we could achieve this primarily with HTML + CSS?

anveshmekala:
> Is there a way we could achieve this primarily with HTML + CSS?

I am able to prototype the solution using CSS . Earlier i assumed the handle element doesn't scale with increase in font-size (track, ticks & font of labels are scaling wrt font-size) which is why we need to determine the translate co-ordinates for `thumb` dynamically.  The main reason for blocking auto scaling with browser is due the use of `px` sizing rather than `rem`  me thinks.  Replaced the existing `px` value with `rem`  for CSS variables in slider which can make the thumb handles bigger when the fontsize in browser settings increase .

src/components/slider/slider.stories.ts
anveshmekala: I ended up adding more slider variants in a single test. This can be hard to maintain in terms of updating the test but adds value in identifying regression for larger font-size.

jcfranco (approved):
Awesome!

Issue #5382, opened 2022-09-26
At min level 0, slider handle appears above scale, and label ticks no longer display 0 value


capegreg:
### Actual Behavior

Issue 1. 
As of 1.0.0-beta.94, the slider handle is positioned above the scale when value is 0.

Issue 2. 
Handle jumps to scale line as slider value increases.

Issue 3.
The minimum tick level 0 no longer displays.

![image](https://user-images.githubusercontent.com/1051118/192314048-3eb7ba7e-f83d-4cf4-b901-29a7054a3706.png)

### Expected Behavior

In previous version 1.0.0-beta.93, handle positioned on same line as scale, and min tick level 0 displayed.

![image](https://user-images.githubusercontent.com/1051118/192314246-89f5ce33-4b63-43cc-bda1-eea9d4572389.png)

### Reproduction Sample

https://codepen.io/capegreg/pen/XWqzvxy?editors=1010

### Reproduction Steps

1. Increase slider value. Handle jumps to scale line.

### Reproduction Version

1.0.0-beta.94

### Regression?

1.0.0-beta.93

### Esri team

N/A

geospatialem:
Looks like a regression from https://github.com/Esri/calcite-components/pull/5138 cc @Elijbet

anveshmekala:
I stumbled up on this one while working on #4721 .

geospatialem:
Verified on `next` ✔️

PR #5397, opened 2022-09-28
fix(slider): thumb no longer appears above track at min

anveshmekala:
**Related Issue:** #5382 

## Summary

This will fix the thumb position to be on the slider track at min value.

After the fix :
EDA939A0-5D0C-4976-A034-DDD9442BF380_4_5005_c

### DIFF
--- a/src/components/slider/slider.tsx
+++ b/src/components/slider/slider.tsx
@@ -1432,10 +1432,15 @@ export class Slider
    *
    * @param value
    */
-  private determineGroupSeparator = (value: number): number | string => {
+  private determineGroupSeparator = (value: number): string => {
     const lang = this.globalAttributes["lang"] || document.documentElement.lang || "en";
-    return this.groupSeparator && value
-      ? localizeNumberString(value.toString(), lang, this.groupSeparator, this.numberingSystem)
-      : value;
+    if (typeof value === "number") {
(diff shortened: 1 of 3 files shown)


src/components/slider/slider.tsx
```diff
@@ -1428,16 +1428,14 @@ export class Slider
...
    */
-  private determineGroupSeparator = (value): string => {
+  private determineGroupSeparator = (value: number): number | string => {
     const lang = this.globalAttributes["lang"] || document.documentElement.lang || "en";
-    if (value) {
-      return this.groupSeparator
-        ? localizeNumberString(value.toString(), lang, this.groupSeparator, this.numberingSystem)
-        : value;
-    }
+    return this.groupSeparator && value
```
driskull, line 1437: What if `value` is `0`? Should it localize it still?

anveshmekala: It should. Updated the code.

anveshmekala: And also, earlier it is required to parse `group-separator` boolean to localize the numbers in `slider`. With this fix `group-separator `is not required. 

cc: @benelan , @geospatialem

driskull (approved):
👍 🎉

PR #5401, opened 2022-09-30
feat(list): Add filtering support.

driskull:
**Related Issue:** #

## Summary

feat(list): Add filtering support.


driskull:
@jcfranco Should we add in filter support for List at this time?

I think filtering makes sense but I think we need to make a plan for sorting/dragging. If we have the `sortable-list` then how does that play with this? Should the `sortable-list` be used internally here?

PR #5402, opened 2022-10-01
build(deps): bump stencil and typescript

benelan:
## Summary
Bumps stencil to `v2.18`, which [adds support for typescript@4.7.4](https://github.com/ionic-team/stencil/blob/main/CHANGELOG.md#-2180-2022-09-12). I also bumped the `jest` packages to match those [used in `@stencil/core@2.18`](https://github.com/ionic-team/stencil/blob/v2.18.0/package.json#L74-L111) in a separate commit. If we don't want to bump the `jest` packages I can revert it, but it looks like the tests are passing. 

- The build patches still look good
- `npm run util:test-types` passes


jcfranco (approved):
💥

PR #5404, opened 2022-10-03
feat(date-picker,input-date-picker): add support for built-in translation

anveshmekala:
**Related Issue:** #4961

## Summary

This PR will add support for translations in `calcite-date-picker` and `calcite-input-date-picker` components.


src/components/color-picker/color-picker.e2e.ts
```diff
@@ -319,7 +319,7 @@ describe("calcite-color-picker", () => {
   };
 
   function assertUnsupportedValueMessage(value: string | object | null, format: string): void {
-    expect(consoleSpy).toBeCalledTimes(1);
+    expect(consoleSpy).toHaveBeenCalledTimes(1);
```
jcfranco, line 322: This is unrelated, please revert.

src/components/date-picker/date-picker.e2e.ts
```diff
@@ -240,4 +227,6 @@ describe("calcite-date-picker", () => {
 
     expect(await date.getProperty("intlYear")).toEqual("Year");
   });
+
+  it("supports translations", () => t9n("<calcite-date-picker> </calcite-date-picker>"));
```
jcfranco, line 231: Our commonTest helpers take the tag name for easier setup:

```tsx
  it("supports translations", () => t9n("calcite-date-picker"));

```

src/components/date-picker/date-picker.tsx
```diff
@@ -120,22 +128,25 @@ export class DatePicker implements LocalizedComponent {
    * Localized string for "previous month" (used for aria label)
    *
    * @default "Previous month"
```
jcfranco, line 130: I think these no longer need defaults since it depends on the locale. I've been removing these entries in my PRs. @geospatialem @benelan WDYT?

src/components/date-picker/date-picker.tsx
```diff
@@ -120,22 +128,25 @@ export class DatePicker implements LocalizedComponent {
...
+   * @deprecated - translations are now built-in, if you need to override a string, please use `messageOverrides`
    */
-  @Prop() intlPrevMonth?: string = TEXT.prevMonth;
+  @Prop() intlPrevMonth?: string;
```
jcfranco, line 133: IMO, we don't need to mark props as optional as Stencil will do this for us for the generated custom element type.

src/components/date-picker/date-picker.tsx
```diff
@@ -202,13 +234,16 @@ export class DatePicker implements LocalizedComponent {
 
   @State() globalAttributes = {};
 
+  @State() defaultMessages: Messages;
```
jcfranco, line 237: This and other state props should be moved to the private state/prop section.

src/components/date-picker/date-picker.tsx
```diff
@@ -202,13 +234,16 @@ export class DatePicker implements LocalizedComponent {
 
   @State() globalAttributes = {};
```
jcfranco, line 235: Sidebar: this doesn't seem to be used at all.

src/components/date-picker/date-picker.tsx
```diff
@@ -462,9 +504,10 @@ export class DatePicker implements LocalizedComponent {
         <calcite-date-picker-month-header
           activeDate={activeDate}
           headingLevel={this.headingLevel || HEADING_LEVEL}
-          intlNextMonth={this.intlNextMonth}
-          intlPrevMonth={this.intlPrevMonth}
-          intlYear={this.intlYear}
+          intlNextMonth={this.messages.nextMonth}
+          intlPrevMonth={this.messages.prevMonth}
+          intlYear={this.messages.year}
+          lang={this.effectiveLocale}
```
jcfranco, line 510: We don't need to set lang here nor pass the locale, right?

src/components/date-picker/date-picker.tsx
```diff
@@ -462,9 +504,10 @@ export class DatePicker implements LocalizedComponent {
         <calcite-date-picker-month-header
           activeDate={activeDate}
           headingLevel={this.headingLevel || HEADING_LEVEL}
-          intlNextMonth={this.intlNextMonth}
-          intlPrevMonth={this.intlPrevMonth}
-          intlYear={this.intlYear}
+          intlNextMonth={this.messages.nextMonth}
```
jcfranco, line 507: I think this is fine to keep (month-header not being a localized/t9n component) since it's a supporting component that won't be used outside of this context. Maybe we can add a note here?

src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -522,6 +529,7 @@ export class InputDatePicker
                   intlYear={this.intlYear}
                   max={this.max}
                   maxAsDate={this.maxAsDate}
+                  messageOverrides={this.messageOverrides}
```
jcfranco, line 532: Can you update the conventions doc and/or util doc about this scenario (non-localized/t9n component forwarding props)?

jcfranco (approved):
Aside from a few comments, this LGTM! 💥

src/components/date-picker/date-picker.tsx
anveshmekala: Nope, we don't. It slipped through while merging locally. Good catch.

src/components/date-picker/date-picker.tsx
anveshmekala: I think that's given considering it is not documented as a public component.

src/components/color-picker/color-picker.e2e.ts
anveshmekala: Its' the pre-commit linting job doing the change. Once #5431 lands-in, it will disappear.

src/components/date-picker/date-picker.tsx
eriklharper: This is for accessing the global `lang` attribute.

src/components/date-picker/date-picker.tsx
anveshmekala: marking it as optional helps us leverage documenting the prop as optional for the user.

src/components/date-picker/date-picker.tsx
anveshmekala: that's wired-up in the new localization interface.

jcfranco: `LocalizedComponent` does not rely on this.

src/components/color-picker/color-picker.e2e.ts
jcfranco: Oh, I see. Ignore my comment then.

src/components/date-picker/date-picker.tsx
jcfranco: We could make it clearer by adding a note under conventions/t9n module about internal components only needing to provide a prop to forward messages (`messages`?).

PR #5412, opened 2022-10-04
build: drop ES5 build

jcfranco:
# Request for comments 🗣️

**Related Issue:** N/A

## Summary

We kept the ES5 build option enabled when we moved to Stencil v2.0 (see https://github.com/Esri/calcite-components/issues/935#issuecomment-686168804) since this was the default behavior of Stencil v1 builds and we still had teams supporting legacy browsers. As legacy browsers are no longer supported, this rolls back that build option. Based on the browser support story, this isn't considered a breaking change at the moment.

cc @cosbyr @MikeTschudi


benelan (approved):
south-park-and-its-gone

jcfranco:
No concerns have been brought up, so moving forward with the install.

PR #5415, opened 2022-10-05
build: fix typescript errors

benelan:
## Summary
Fixes some typescript errors that have been floating around for a while. You can reproduce by running `tsc`:

image

- The `tile-select.e2e.ts` and `dropdown.e2e.ts` errors are self explanatory. 
- The storybook errors coming from `node_modules` were because we were using a storybook util in the e2e tests, e.g. [here](https://github.com/Esri/calcite-components/blob/73e80af97a76ec6ca028a3d4f979e469b3a0962d/src/components/card/card.e2e.ts#L3). I fixed that by ~~excluding e2e files in `tsconfig.json`~~ moving the `placeholderImage` util into `src/tests/utils.ts`.
- The `stencil.config.ts` error was because our tailwind config is incompatible with the  `Config` type from `"tailwindcss/types/config"`; probably because of some custom theme stuff.  The tailwind plugin also accepts a path to the tailwind config as a string, but then it was importing the config as a commonjs module causing import errors. I just cast the type as `any`.
    
    image
 
- The `tailwind.config.ts` errors were because [`variants` configuration no longer exists in tailwind v3](https://tailwindcss.com/docs/upgrade-guide#remove-variant-configuration). Apparently variants are now automatically available for every utility. The `variants("borderColor")` wasn't doing anything because we no longer have a variants section in our config (and because the helper function doesn't exist anymore lol). See [tailwind's v2 `variants` doc](https://v2.tailwindcss.com/docs/plugins#variants) for more info, specifically:
    
    image


tsconfig-demos.json
```diff
@@ -1,4 +1,5 @@
 {
   "extends": "./tsconfig-base",
-  "include": ["src/demos", "src/components.d.ts"]
+  "include": ["src/demos", "src/components.d.ts"],
+  "exclude": ["node_modules"]
```
jcfranco, line 4: Nitpick: `node_modules` is already excluded by default: https://www.typescriptlang.org/tsconfig#exclude

jcfranco (approved):
💥🧹💥

PR #5421, opened 2022-10-06
docs(flow-item, panel): use @example jsdoc tag to fix formatting

benelan:
**Related Issue:** #

## Summary
Use the `example` jsdoc tag instead of fenced code, so the formatting isn't so goofy in the afd doc

image


jcfranco (approved):
📚✅

PR #5437, opened 2022-10-11
feat(rating): add built-in translations

jcfranco:
**Related Issue:** #4961

## Summary

Wires up translations to the rating component.


src/tests/commonTests.ts
```diff
@@ -974,25 +974,35 @@ export async function t9n(componentSetup: TagOrHTML | TagAndPage): Promise<void>
...
+      fakeBundleIdentifier
+    );
 
     component.setAttribute("lang", "es");
     await page.waitForChanges();
     await page.waitForTimeout(3000);
     const esMessages = await getCurrentMessages();
 
-    expect(esMessages).toEqual({});
+    expect(esMessages).toHaveProperty(fakeBundleIdentifier);
```
anveshmekala, line 1005: we can assert `esMessages` toEqual `enMessage` without the need of adding `fakeBundleIdentifier`.

anveshmekala (approved):
LGTM! 💯

src/tests/commonTests.ts
jcfranco: If we did so, this would no longer catch problems with locale bundle switching as the `en` one wouldn't change.

anveshmekala: make sense.

PR #5447, opened 2022-10-11
feat(button): add built-in translations

anveshmekala:
**Related Issue:** #4961 

## Summary
This PR will add built-in translation support for `calcite-button`


src/components/button/button.tsx
```diff
@@ -129,16 +154,29 @@ export class Button implements LabelableComponent, InteractiveComponent, FormOwn
...
+
   //--------------------------------------------------------------------------
   //
   //  Lifecycle
   //
   //--------------------------------------------------------------------------
 
-  connectedCallback(): void {
+  async connectedCallback(): Promise<void> {
+    if (Build.isBrowser) {
```
jcfranco, line 171: Why do we need to check for `isBrowser` here?

src/components/button/button.tsx
```diff
@@ -129,16 +154,29 @@ export class Button implements LabelableComponent, InteractiveComponent, FormOwn
...
+      connectLocalized(this);
+      connectMessages(this);
+    }
+
+    connectLabel(this);
```
jcfranco, line 176: If moving this line was not necessary, please make sure label behavior works. Otherwise, you can restore this.

jcfranco (approved):
Other than a question regarding the use of `Build.isBrowser` and modified `connectLabel` call, this LGTM! 🎉

src/components/button/button.tsx
anveshmekala: Invoke `mutationObserver` only in the browser context and avoid hydrate errors for prerender builds. This is an interesting case where this occurs only for a few components.

anveshmekala: it is resolved now. we longer need to check for build context at the component level.

PR #5457, opened 2022-10-11
feat(inline-editable): add built-in translations

anveshmekala:
**Related Issue:** #4961 

This PR will add built-in translations for `calcite-inline-editable`.


src/components/inline-editable/assets/inline-editable/t9n/messages.json
```diff
@@ -1,5 +1,5 @@
 {
-  "enablingEditing": "Click to edit",
```
jcfranco, line 2: Great catch!

jcfranco (approved):

PR #5458, opened 2022-10-11
feat(input-number, input, input-text): add built-in translations

anveshmekala:
**Related Issue:** #4961 

This will add built-in localization support for `calcite-input`, `calcite-input-number`, `calcite-input-text` components.


jcfranco (approved):
💥

Issue #3723, opened 2021-12-15
calcite-input-date-picker calendar should not change year in certain case


AdelheidF:
Reported by Beyondsoft

Instruction:
- place cursor in input box and hit backspace 1-3 times
- the calendar popover updates to weird years
- the probability that users actually want to switch from year 2021 to year 202, 20 or 2 is probably nil
- therefore the calendar popover should not change its year until a valid date is entered in the input box or the user clicked on a date inside the calendar

From Beyondsoft:
10. Click the calendar icon of date textbox.
11. Select a date and delete the date by press backspace keyboard.  –key step
12. Observe.
13. Continue to delete date in date textbox by press backspace keyboard again. –key step
14. Observe.
15. Continue to delete date in date textbox by press backspace keyboard again. –key step
16. Observe.

Actual result:
Step 12: It got stuck and then skipped to invalid date if deleting date in date textbox by press backspace keyboard.
Step 14: It skips to another invalid date if deleting date in date textbox by press backspace keyboard again.
Step 16: It displays default date value in date textbox if deleting date in date textbox by press backspace keyboard again.

Expected:
The deleted value in date should be removed but the original date should not be changed in date textbox if deleting date in date textbox by press backspace keyboard.

### Actual Behavior

year inside the calendar changes to 202, 20, 2

### Expected Behavior

calendar should stick to year 2021

### Reproduction Steps and Sample

https://codepen.io/afreitag/pen/KKXaYQa?editors=1010

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-beta.71`

- [ ] CDN
- [ ] NPM package

macandcheese:
I agree our current implementation there is a bit lacking. A native (or calcite-input) of type="date" also provides helpers for focusing day, month, year independently and incrementing / decrementing with arrows. I wonder if having this component leverage that would improve the overall UX.

eriklharper:
@AdelheidF Is the ask here that the calendar shouldn't update unless at least a 4-digit year is typed?  Technically `202`, `20` and `2` are all valid years, so it sounds like the ask is to only update the calendar when a full 4 digits are typed, otherwise you'll prevent someone from typing in valid single, double and triple digit years (in number format as opposed to padded zero string representations).

AdelheidF:
Yes, I'd prefer that we ask the user to always enter a 4 digit year, otherwise there are too many options of what it could mean.

AdelheidF:
verified with beta.98. Nice!

PR #5459, opened 2022-10-11
fix(input-date-picker): commit value and refresh calendar on blur or Enter keypress in the input

eriklharper:
**Related Issue:** #3723 

## Summary

This fix changes the input to no longer commit the changed value and refresh the calendar while the user is typing, only doing so when they commit the value either by pressing the Enter key or unfocusing (blurring) the input.  Clicking dates in the calendar popup still continue to work as before.

### TODO:

- [x] Fix issue where the input's value isn't formatted correctly in DD/MM/YYYY locales when a date is clicked in the Calendar
- [x] Refresh calendar while typing only when a valid date is typed with at least a 4-digit year in range mode


src/utils/date.ts
```diff
@@ -70,6 +73,33 @@ export function dateFromISO(iso8601: string | Date, isEndDate = false): Date | n
...
+/**
+ * Parse a localized date string into a valid Date.
+ * return false if date is invalid, or out of range
+ *
+ * @param value
+ * @param localeData
+ */
+export function dateFromLocalizedString(value: string, localeData: DateLocaleData): Date | false {
+  if (!localeData) {
+    return false;
```
jcfranco, line 85: Returning a boolean from a function that indicates a date is returned seems odd to me. Let's return `null` or `undefined` instead.

src/utils/date.ts
```diff
@@ -164,25 +194,41 @@ export function parseNumber(str: string, localeData: DateLocaleData): number {
...
  *
  * @param str
  * @param localeData
+ * @param returnType
  */
-export function parseDateString(str: string, localeData: DateLocaleData): { day: number; month: number; year: number } {
+export function parseDateString(
+  str: string,
+  localeData: DateLocaleData,
+  returnType: "number" | "string" = "number"
```
jcfranco, line 208: Can we leave it up to the component to stringify the numbers? I'm assuming this is to avoid parsing from string to format again, but I think this is negligible. We can revisit if we notice perf issues.

src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -347,7 +366,7 @@ export class InputDatePicker
   /**
    * Fires when the component's value changes.
    */
-  @Event({ cancelable: false }) calciteInputDatePickerChange: EventEmitter<void>;
+  @Event({ cancelable: false }) calciteInputDatePickerChange: EventEmitter<DateRangeChange | void>;
```
jcfranco, line 369: Let's roll this back. Per our conventions, event payloads should not include data that is accessible via a property.

src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -895,35 +861,110 @@ export class InputDatePicker
...
+    const { value: oldValue } = this;
+    const oldValueIsArray = Array.isArray(oldValue);
+    const valueIsArray = Array.isArray(value);
+
+    const newStartDate = valueIsArray ? value[0] : "";
+    const newStartDateISO = valueIsArray ? dateToISO(newStartDate) : "";
+    const newEndDate = valueIsArray ? value[1] : "";
+    const newEndDateISO = valueIsArray ? dateToISO(newEndDate) : "";
+
+    // TODO: ensure start and end dates are within min and max
```
jcfranco, line 903: Are these still pending?

jcfranco (changes requested):
Looking good, @eriklharper! 😎

src/utils/date.ts
```diff
@@ -164,25 +194,41 @@ export function parseNumber(str: string, localeData: DateLocaleData): number {
...
   const { separator, unitOrder } = localeData;
   const order = getOrder(unitOrder);
   const values = replaceArabicNumerals(str).split(separator);
+  const day = values[order.indexOf("d")];
+  const month = values[order.indexOf("m")];
+  const year = values[order.indexOf("y")];
+  if (returnType === "number") {
+    return {
+      day: parseInt(day),
+      month: parseInt(month) - 1, // this subtracts by 1 because the month in the Date contructor is zero-based https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getMonth
```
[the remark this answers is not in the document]
eriklharper: Yeah, thanks to Paul I understood that only months for some reason are zero-based!  Huge gotcha!

src/components/input-date-picker/input-date-picker.tsx
eriklharper: Good catch.  I got confused because I thought at first that we needed to include that event payload to keep from removing a deprecated feature, but since this is the new event that replaces the deprecated event, it doesn't make sense to keep emitting the payload.

eriklharper: Cool, rolled that back.  I also added value checks in the test of what would be the EventTarget for the change event to make sure it looks right since the event detail is no longer being emitted.

src/components/date-picker/date-picker.tsx
```diff
@@ -46,6 +46,9 @@ export class DatePicker implements LocalizedComponent {
   //
   //--------------------------------------------------------------------------
 
+  /** Active date */
```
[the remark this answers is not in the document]
eriklharper: Yes, this is needed because it shows what the value would be in the calendar pre-commit, so it highlights it in gray before the value is committed, at which point it turns to blue to denote it is selected.

src/components/input-date-picker/input-date-picker.tsx
eriklharper: These are not, good catch, just removed them.

src/utils/date.ts
eriklharper: The `string` returnType is not just for convenience.  It allows for checking if a user entered years with leading zeros like `0001`, `0010`, `0100`, which is the centerpiece of the original issue, which is to prevent the calendar from updating on years less than 1000.

I could create a separate util function that returns `day`, `month` and `year` as strings, or a simpler function like `isFourDigitYear` that literally just checks if a date value contains at least a 4 digit year, but both those seem unnecessary and redundant because I still would have to handle `unitOrder` and `separator` from the locale data.

src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -679,9 +696,58 @@ export class InputDatePicker
...
+    const date = dateFromLocalizedString(focusedInputValue, this.localeData) as Date;
+    const dateAsISO = dateToISO(date);
+    const valueIsArray = Array.isArray(value);
+    if (this.range) {
+      const focusedInputValueIndex = focusedInput === "start" ? 0 : 1;
+      if (valueIsArray) {
+        if (dateAsISO === value[focusedInputValueIndex]) {
+          return;
+        }
+        if (date) {
```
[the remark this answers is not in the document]
eriklharper: I'll revisit this in a follow-up

src/utils/date.ts
jcfranco: Per our conversation, I like your suggestions to add a method that parses a string into the specific parts. We could keep `parseDateString`'s signature the same and internally it could use the date splitting method and it would just need to convert them to numbers. My main reason was to keep the return type simple and consistent and not have to think about switching things around. It's only used in the date picker for now, so I guess it's not a big deal to have a flag for the return type.

eriklharper: Ok, I created a new function `datePartsFromLocalizedString` that moves the logic for capturing the day, month and year values as strings from an incoming localized string so that we can properly determine if a 4-digit year is typed and not have to overload the return type on the parseDateString function.  Let me know if that looks good :)

src/utils/date.spec.ts
```diff
@@ -11,6 +11,7 @@ describe("inRange", () => {
   it("returns true if no min/max", () => {
     expect(inRange(new Date())).toEqual(true);
   });
+
```
jcfranco, line 14: Since there were no relevant code changes here, can you submit this separately as a `style` PR?

jcfranco (approved):
👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈
👈📆👈👈👈📆👈👈📆📆👈👈📆📆📆👈👈📆📆📆👈📆📆📆📆👈📆👈
👈📆📆👈👈📆👈📆👈👈📆👈👈📆👈👈📆👈👈👈👈📆👈👈👈👈📆👈
👈📆👈📆👈📆👈📆👈👈📆👈👈📆👈👈📆👈👈👈👈📆📆📆👈👈📆👈
👈📆👈👈📆📆👈📆👈👈📆👈👈📆👈👈📆👈👈👈👈📆👈👈👈👈👈👈
👈📆👈👈👈📆👈👈📆📆👈👈📆📆📆👈👈📆📆📆👈📆📆📆📆👈📆👈
👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈👈

eriklharper:
Ran screener locally as a sort of smoke test and things appeared fine with date picker stories.

PR #5465, opened 2022-10-12
feat(chip): add built-in translations

anveshmekala:
**Related Issue:** #4961 

This PR will add built-in translations for `calcite-chip`.


src/tests/commonTests.ts
```diff
@@ -910,13 +910,20 @@ export async function floatingUIOwner(
  * Helper to test t9n component setup
  *
  * @param {TagOrHTML|TagAndPage} componentSetup - A component tag, html, or an object with e2e page and tag for setting up a test
+ * @param intlProps
+ * @param nonIntlProps
  */
-export async function t9n(componentSetup: TagOrHTML | TagAndPage): Promise<void> {
+export async function t9n(
+  componentSetup: TagOrHTML | TagAndPage,
+  intlProps = true,
```
jcfranco, line 918: This signature might cause issues for components that have both intl props in addition to extra ones. I suggest dropping `intlProps` (always testing for those) and changing `nonIntlProps` to `extraMessageProps` and testing for those if a value was indeed passed along. We can revisit this if we come up with a consistent way to group the non-intl ones (e.g., a11y labels).

jcfranco (approved):
Other than a suggestion to the `t9n` test helper, this LGTM!

src/tests/commonTests.ts
anveshmekala: make sense. proposing `additionalMessageProps` inplace of `extraMessageProps`.  WDYT ?

PR #5466, opened 2022-10-12
refactor: use getters and setters in NumberStringFormat

benelan:
**Related Issue:** #5301

## Summary
Uses getter and setter methods in `NumberStringFormat` for safesies. Also caches the number formatter in `calcite-alert` by serializing and comparing the options object, and using the whole object for the `Intl.NumberFormat` options when there are changes.

PR #5467, opened 2022-10-12
feat(stepper, stepper-item): add numberingSystem property

benelan:
**Related Issue:** #4893

## Summary
Adds `numberingSystem` property to `calcite-stepper`. A couple lines in `calcite-stepper-item` will need to be changed due to a refactor in #5466. The small change can go in which ever PR is merged last.

### DIFF
--- a/src/components/stepper-item/stepper-item.tsx
+++ b/src/components/stepper-item/stepper-item.tsx
@@ -8,6 +8,7 @@ import {
   Listen,
   Method,
   Prop,
+  State,
   VNode,
   Watch
 } from "@stencil/core";
@@ -19,6 +20,12 @@ import {
   StepperItemEventDetail,
   StepperItemKeyEventDetail
 } from "../stepper/interfaces";
+import {
+  numberStringFormatter,
+  LocalizedComponent,
+  disconnectLocalized,
+  connectLocalized
+} from "../../utils/locale";
 
 /**
  * @slot - A slot for adding custom content.
@@ -28,7 +35,7 @@ import {
   styleUrl: "stepper-item.scss",
   shadow: true
 })
-export class StepperItem implements InteractiveComponent {
+export class StepperItem implements InteractiveComponent, LocalizedComponent {
   //--------------------------------------------------------------------------
   //
   //  Element
@@ -127,6 +134,8 @@ export class StepperItem implements InteractiveComponent {
   //
   //--------------------------------------------------------------------------
 
+  @State() effectiveLocale = "";
+
   headerEl: HTMLDivElement;
 
   //--------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


src/components/stepper-item/stepper-item.tsx
```diff
@@ -193,7 +203,17 @@ export class StepperItem implements InteractiveComponent {
     updateHostInteraction(this, true);
   }
 
+  disconnectedCallback(): void {
+    disconnectLocalized(this);
+  }
+
   render(): VNode {
+    numberStringFormatter.setOptions({
```
jcfranco, line 211: Is there another hook we can use to update the options? You could watch for effectiveLocale, but that wouldn't handle changes to the parent's numbering system. Would that be a common scenario, you think?

jcfranco (approved):
Aside from a considering another way to update the formatter options outside of `render`, this LGTM! 

🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺
🎺🏇🎺🎺🎺🏇🎺🏇🎺🎺🎺🏇🎺🏇🏇🏇🎺🎺🏇🏇🏇🏇🎺🏇🏇🏇🏇🎺🏇🎺
🎺🏇🏇🎺🏇🏇🎺🏇🎺🎺🎺🏇🎺🏇🎺🎺🏇🎺🏇🎺🎺🎺🎺🏇🎺🎺🎺🎺🏇🎺
🎺🏇🎺🏇🎺🏇🎺🏇🎺🎺🎺🏇🎺🏇🏇🏇🎺🎺🏇🎺🏇🏇🎺🏇🏇🏇🎺🎺🏇🎺
🎺🏇🎺🎺🎺🏇🎺🏇🎺🎺🎺🏇🎺🏇🎺🏇🎺🎺🏇🎺🎺🏇🎺🏇🎺🎺🎺🎺🎺🎺
🎺🏇🎺🎺🎺🏇🎺🎺🏇🏇🏇🎺🎺🏇🎺🎺🏇🎺🏇🏇🏇🏇🎺🏇🏇🏇🏇🎺🏇🎺
🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺🎺

src/components/stepper-item/stepper-item.tsx
benelan: Yeah I'm not really sure there is a use case for changing locale and numberingSystem besides testing/debugging. I've just been updating the options to be safe lol. 

I guess we could create an internal event or a mutation observer for the numberingSystem prop on the parent and a watcher for effectiveLocale, if you think that's better

jcfranco: I'm leaning towards only updating the formatter on `effectiveLocale` changes for now.

Issue #5369, opened 2022-09-21
setFocus() requires a timeout in componentDidLoad()


AdelheidF:
### Description

There is a change with setFocus in beta.94. The call itself works fine, but in the past I could just do this:
```
componentDidLoad(): void {
    this.panelNode.setFocus("back-button");
}
```
This no longer works. I have to add the setFocus() call into a timeout now. Not sure though what a reliable wait time is.

This is the case both for panel or flow-item, or if I want to set the focus on a close button I add myself. No difference there.

From @driskull 

I think we can improve this in the future by setting up a promise internally and waiting for the component to load before resolving the promise and then setFocus would wait for that promise first.
But I think we can wait for the component to load before focusing internally.

### Acceptance Criteria

setFocus should not require a timeout in componentDidLoad()

### Which Component

panel, flow-item, ...

### Esri team

N/A

jcfranco:
Could be related to https://github.com/ionic-team/stencil/issues/3580.

driskull:
I did a little bit of testing, this seems to work fine if there are no async lifecycle methods in the child component.

@AdelheidF is this still an issue? If so, can you point me to where its occurring in code? I'd like to get some more info.

AdelheidF:
I can still repro with next.650. I have a few flow-items with footer buttons and in some cases it still requires a timeout of 200 for the footer button to get the focus on forward or backward flow movement. Same issue for header back button of flow-items. No async calls when opening the panels.

![image](https://user-images.githubusercontent.com/44481292/205157395-05bdb53d-b3b3-4df1-804d-c7a78c5b7ed5.png)![image](https://user-images.githubusercontent.com/44481292/205157539-7eab5a03-9547-41e1-a2e0-c5b05a97d34f.png)

geospatialem:
@AdelheidF Would it be possible to put a Codepen together for us to reproduce for a fix, and/or some context on where it might be occurring in the code? Want to make sure we're understanding the complexity fully to diagnose.

AdelheidF:
@driskull has repro cases using our components. I have not managed to get it to happen in a simple codepen.

geospatialem:
Will be re-evaluating post 1.0 release to determine if this is in the scope of the components, and if so, prioritization for a future 2023 release.

AdelheidF:
Using `requestAnimationFrame()` fixes it in many places. I'm wondering if this could be added to all calcite `setFocus()` calls.

[...]

PR #5471, opened 2022-10-13
feat: add built-in translations

anveshmekala:
**Related Issue:** #4961 

## Summary

This PR will add built-in translations to the components across calcite .

Notable features  :

- Users no longer need to parse translated strings, translations are now built-in. List of the supported locale's can be found [here](https://developers.arcgis.com/calcite-design-system/localization/#locale-support).
- One can overwrite default translations for a given locale using `messageOverrides`  property.
- Specifying `lang` at document level , component level and ancestor level is supported.  
   - component level always takes precedence and then ancestor's followed by doc level.
      ```
       <!--Though arabic is specified as lang at doc level , button component will follow Spanish for string localization . -->
       <html lang="ar"> 
       <body>
       <calcite-button lang="es"> </calcite-button>
       </body>
       </html>
       ```

### Migration 

To migrate calcite components that use `intl` props &  properties for accessible labels for overriding default values, refer to the following example to start using `messageOverrides` object property.

Before:
```
<!-- overriding via intl props -->
<calcite-color-picker intl-saturation="{custom value}"  intl-save="{custom value}"> </calcite-color-picker>
```
Current :
```
<!-- leveraging building translations. 
 display translations for Spanish lang -->
<calcite-color-picker lang="es"> </calcite-color-picker>

<!-- if the user wants to customize the default translations provided, one can override as below -->
<calcite-color-picker lang="es"> </calcite-color-picker>
<script>
const colorpickEl = document.querySelector('calcite-color-picker')
colorpickerEl.messageOverrides = { blue = "custom value" , red = "custom value" }
</script>

### CODE REVIEW

File: src/components/action-bar/action-bar.tsx [hunk 1]
```diff
@@ -158,10 +204,16 @@ export class ActionBar implements ConditionalSlotComponent {
     connectConditionalSlotComponent(this);
   }
 
+  async componentWillLoad(): Promise<void> {
+    await setUpMessages(this);
```
driskull, line 208: @jcfranco now that we're delaying `componentDidLoad` more with fetching async messages, I wonder if we need to prioritize updating our `setFocus()` methods to wait for the component to be rendered the first time before proceeding with focusing?

#5369 is a priority for MapViewer and I'd like to try and fix it and make them not have to increase any timeouts they have currently for focusing.

src/components/action-bar/action-bar.tsx
```diff
@@ -75,11 +88,15 @@ export class ActionBar implements ConditionalSlotComponent, LoadableComponent {
 
   /**
    * Specifies the label of the expand icon when the component is collapsed.
+   *
+   * @deprecated - translations are now built-in, if you need to override a string, please use `messageOverrides`
```
alisonailea, line 92: are we removing @deprectaed before 1.0?

src/components/date-picker/date-picker.tsx
```diff
@@ -486,10 +524,11 @@ export class DatePicker implements LocalizedComponent {
       this.localeData && [
         <calcite-date-picker-month-header
           activeDate={activeDate}
-          headingLevel={this.headingLevel}
-          intlNextMonth={this.intlNextMonth}
-          intlPrevMonth={this.intlPrevMonth}
-          intlYear={this.intlYear}
+          headingLevel={this.headingLevel || HEADING_LEVEL}
+          //t9n props are used here to forward the messages only.
+          intlNextMonth={this.messages.nextMonth}
```
driskull, line 529: Whats the plan for when these are deprecated? Will there be a way to pass these along?

src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -597,11 +604,13 @@ export class InputDatePicker
                   activeRange={this.focusedInput}
                   endAsDate={this.endAsDate}
                   headingLevel={this.headingLevel}
+                  //t9n props are used here to forward the messages only.
                   intlNextMonth={this.intlNextMonth}
```
driskull, line 608: Shouldn't these be forwarded with messagesOverrides? Applies to other places too.

src/components/input/input.tsx
```diff
@@ -945,16 +987,17 @@ export class Input
   // --------------------------------------------------------------------------
 
   render(): VNode {
+    console.log("messages", this.messages);
```
driskull, line 990: remove console.log. Search for other occurances.

src/components/rating/rating.tsx
```diff
@@ -176,7 +216,7 @@ export class Rating
                 <calcite-icon icon="star-f" scale={this.scale} />
               </div>
             )}
-            <span class="visually-hidden">{this.intlStars.replace("${num}", `${i}`)}</span>
+            <span class="visually-hidden">{this.messages.stars.replace("${num}", `${i}`)}</span>
```
driskull, line 219: for variables in messages, do they prefer `{}` or `${}`? Should we make one or the other a preferred best practice?

driskull (commented):
Comments/questions

src/components/value-list/utils.ts
```diff
@@ -58,6 +42,6 @@ export function getHandleAndItemElement(event: KeyboardEvent | FocusEvent): {
 
 export function replacePlaceholders(text: string, label: string, position: number, total: number): string {
   const replacePosition = text.replace("${position}", position.toString());
-  const replaceLabel = replacePosition.replace("${item.label}", label);
+  const replaceLabel = replacePosition.replace("${itemLabel}", label);
```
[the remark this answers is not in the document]
anveshmekala: Yup, this is changed intentionally with respect to strings in the bundle.

src/components/date-picker/date-picker.tsx
anveshmekala: Good point . There could be two ways we can do this, one is to keep the `intl` props on `date-picker-month-header` component because it's `internal` or introduce `messageOverrides` and utilize the object to keep it consistent.

src/components/input-date-picker/input-date-picker.tsx
anveshmekala: These need to be removed as we are passing `messageOverrides` here anyways.

src/components/rating/rating.tsx
anveshmekala: It's more of a preference. Using `${ }` makes it easier to understand that it is a custom value. I am open to use `{ }` as well.

driskull: cool is that in the conventions file?

src/components/date-picker/date-picker.tsx
driskull: I think we should introduce messagesOverrides so we get rid of all the intl props

src/components/alert/alert.tsx
```diff
@@ -122,6 +131,24 @@ export class Alert implements OpenCloseComponent, LocalizedComponent, LoadableCo
   /** Specifies the size of the component. */
   @Prop({ reflect: true }) scale: Scale = "m";
 
+  /**
+   * Made into a prop for testing purposes only
+   *
+   * @internal
+   */
+  @Prop({ mutable: true }) messages: Messages;
```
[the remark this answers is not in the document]
anveshmekala: this is turned into Prop for testing purposes only. if we use state, I don't think it is accessible in test environment.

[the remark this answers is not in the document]
anveshmekala: reagrding `@readonly` tag, it is to help the developers or you were referring to documentation .

src/utils/locale.ts
```diff
@@ -3,6 +3,50 @@ import { createObserver } from "./observers";
 import { closestElementCrossShadowBoundary, containsCrossShadowBoundary } from "./dom";
 
 export const defaultLocale = "en";
+
+export const t9nLocales = [
```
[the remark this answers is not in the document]
anveshmekala: Added.

src/components/alert/alert.tsx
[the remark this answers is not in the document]
driskull: I was thinking developers since its not documented.

src/components/input-date-picker/input-date-picker.tsx
anveshmekala: `input-date-picker` is not a `t9nComponent` which means when the user sets `intl-*` props or `messageOverrides` they won't be in sync and it is the `date picker's job to merge those and keep them in sync which is why we need to prase them.

src/utils/locale.ts
```diff
@@ -124,8 +124,14 @@ export const numberingSystems = [
   "tibt"
 ] as const;
 
+export const supportedLocales = [...new Set([...t9nLocales, ...locales])] as const;
+
 export type NumberingSystem = typeof numberingSystems[number];
 
+export type SupportedLocales = typeof supportedLocales[number];
+
+console.log(supportedLocales);
```
driskull, line 133: console.log :)

src/components/input-date-picker/input-date-picker.tsx
driskull: Gotcha. Yeah maybe we can just remove documentation for this component.

src/components/alert/alert.tsx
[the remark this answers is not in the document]
anveshmekala: i don't think this is can be `@readonly` because it's mutable in `t9nComponent` instance. Thoughts?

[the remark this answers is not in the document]
driskull: We can ignore for now 👍

src/components/rating/rating.tsx
anveshmekala: nope, this is not in the conventions yet.

src/components/date-picker/date-picker.tsx
anveshmekala: this needs a bit of correction. we may introduce `messages` prop in `date-picker-month-header` rather than `messagesOverrides` because the component is internal and user's can only override via the date-picker which already has `messageOverrides` property . `messages` property is the single source of truth and hence it will be parsed down to the child components which are non-localized / non-t9n.

driskull: Yeah that makes the most sense. These individual small components aren't documented for use by themselves so they can just take the large messages prop from their parent.

src/components/block/block.tsx
```diff
@@ -257,14 +316,14 @@ export class Block implements ConditionalSlotComponent, InteractiveComponent {
           headerContent
         )}
         {loading ? (
-          <calcite-loader inline is-active label={intlLoading} />
+          <calcite-loader inline is-active label={messages.loading} />
```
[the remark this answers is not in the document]
anveshmekala: Good catch. Not really sure the purpose of this, this is currently in master and was added a while ago.  I don't see a prop in the loader, assuming its safe to remove.

src/components/action-bar/action-bar.tsx
```diff
@@ -290,28 +340,16 @@ export class ActionBar implements ConditionalSlotComponent, LoadableComponent {
...
-    const expandLabel = intlExpand || TEXT.expand;
-    const collapseLabel = intlCollapse || TEXT.collapse;
 
     const expandToggleNode = !expandDisabled ? (
       <ExpandToggle
         el={el}
         expanded={expanded}
-        intlCollapse={collapseLabel}
-        intlExpand={expandLabel}
+        intlCollapse={messages.collapse}
```
driskull, line 351: Can we add a note for 1.0 to rename these function component params so they aren't `intl-` anymore?

src/utils/locale.ts
```diff
@@ -95,31 +143,54 @@ export const defaultNumberingSystem =
...
-  if (locales.indexOf(locale) > -1) {
-    return locale;
-  }
+/**
+ * Gets the locale that best matches the context.
+ *
+ * @param locale – the BCP 47 locale code
+ * @param context - specifies whether the locale code should match in the context of CLDR or T9N (translation)
+ */
+export function getSupportedLocale(locale: string, context: "cldr" | "t9n" = "cldr"): SupportedLocales {
```
driskull, line 152: Do you think we should use the type `SupportedLocales` here or anywhere else?

src/utils/locale.ts
```diff
@@ -95,31 +143,54 @@ export const defaultNumberingSystem =
...
   }
 
-  return locales.includes(locale) ? locale : defaultLocale;
+  // we can `zh-CN` as base translation for chinese locales which has no corresponding bundle.
+  if (locale === "zh") {
+    return "zh-CN";
+  }
+
+  if (!contextualLocales.includes(locale)) {
+    console.warn(`Translations for "${locale}" not found or invalid, falling back to english`);
```
driskull, line 189: nitpick: `Translations for "${locale}" are not available, falling back to English (en).`

@geospatialem maybe you could suggest some better warning text?

driskull (approved):
👍

src/utils/locale.ts
geospatialem: @driskull WDYT about:
> Translations for the "${locale}" locale are not available and will fall back to the default, English (en).

Would like to point developers to a list of currently available locales on our docs site, but would likely be after our 1.0 release.

src/components/action-bar/action-bar.tsx
anveshmekala: good idea. Will add that notes.

src/utils/locale.ts
driskull: sounds good

src/utils/locale.ts
anveshmekala: we can add it for the `lang` property on each component but it's a global attribute. I don't think we can add it anywhere else. Am I missing anything?

driskull: > we can add it for the lang property on each component but it's a global attribute. 

No, lets not do that.

> Am I missing anything?

Nope. Just wondering if anywhere else is applicable.

src/components/action-group/action-group.tsx
```diff
@@ -50,6 +60,8 @@ export class ActionGroup implements ConditionalSlotComponent {
 
   /**
    * Specifies a text string for the `calcite-action-menu`.
+   *
+   * @deprecated - translations are now built-in, if you need to override a string, please use `messageOverrides`
```
geospatialem, line 64: nitpick: For the `@deprecated` translation message, can we add a capital on the first letter and a period at the end for API style consistency? Added a recommended refactored line below for use across components, if its possible to integrate easily:

> @deprecated - Translations are built-in. To override the built-in message string, use `messageOverrides`.

geospatialem (commented):
A few comments regarding doc conventions, a lot of nitpicks if time allows. Looking forward to having this in, nice work @anveshmekala! 💪🏻

src/components/action-group/action-group.tsx
anveshmekala: These will be removed in #5946. Can we skip this for now?

geospatialem: Yes, definitely OK to skip. Thanks for clarifying on the removal PR 👍🏻

benelan:
As discussed on Teams, we can see in the [Chromatic canvas that the stories are working correctly](https://www.chromatic.com/component?appId=6266d45509d7eb004aa254fb&csfId=components-action-pad&specName=Romanian%20Moldova%20Locale%20Test%20Only&buildNumber=448), even though the snapshots are blank. I have accepted the snapshots for now so they aren't blocking this PR. Anvesh messaged Chromatic support and we will fix the blank snapshots when they provide additional info.

Issue #5333, opened 2022-09-15
Bug: Error when Tree is the root node in a custom element


benelan:
### Actual Behavior

When `calcite-tree` is the root node, it doesn't render and there is an error in the console.
![image](https://user-images.githubusercontent.com/10986395/190527476-df1545a6-aa2d-446f-85d6-ce8c9f2097ff.png)

### Expected Behavior

`calcite-tree` still works correctly when it is the root node, e.g. the behavior when you wrap the tree with a `div` in the repro sample below.

### Reproduction Sample

https://codepen.io/benesri/pen/oNdBPRQ

### Reproduction Steps

1. Open dev tools and notice the errors in the console
2. Wrap the `calcite-tree` in a `div` (uncomment lines 6 and 40)
3. The errors go away and the tree renders

### Reproduction Version

beta.94

### Relevant Info

I don't think we can assume that `calcite-tree` will have a parent. #adoptatree

https://github.com/Esri/calcite-components/blob/ed9306124c6b0570fe3606a47b33840aeab129bb/src/components/tree/tree.tsx#L74

This issue stemmed from #4814, but I saw a different error using the github repro sample. The two errors are _probably_ caused by the same thing.

jcfranco:
Triaged for this release. cc @benelan @geospatialem

geospatialem:
Verified on `next.601`.

PR #5472, opened 2022-10-13
fix(tree, tree-item): works when tree is the topmost element in a shadow root where it has no parent

### LINKED ISSUE #5464: `CalciteTree.componentWillRender()` throws error when `<calcite-tree>` is detached
### Actual Behavior

A `TypeError` is sometimes thrown by the `CalciteTree.componentWillRender()` hook.

### Expected Behavior

A `TypeError` is never thrown by the `CalciteTree.componentWillRender()` hook.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/KKRbYOQ

### Reproduction Steps

1. Visit sample and open the browser's console.
2. Click the **Mutate and detach** button
3. Observe error output in console.

### Reproduction Version

beta.95

### Relevant Info

Due to stencil's asynchronous nature when updating props or calling methods, a race condition exists when mutating a property immediately before removing the element from the DOM. A scenario more likely encountered when writing unit tests.

Specifically, the property mutation triggers a re-render at the same time the element is detached. By the time the `componentWillRender()` is called, the element may already be detached. When this is the case, the following error and stacktrace (abbreviated) is thrown:

```
TypeError: Cannot read properties of null (reading 'closest')
    componentWillRender (p-e1c801a6.entry.js:formatted:252)
    ...
    requestAnimationFrame (async)
    ...
    set (p-975aba66.js:6)
    (anonymous) (tests.js:19450)
...
```

The failing line is:
https://github.com/Esri/calcite-components/blob/f5716f345f71a14d99267e1bf72d22fa28a9870f/src/components/tree-item/tree-item.tsx#L160

If the element is detached at this point, `parentTree.parentElement` is `null` and `parentTree.isConnected` is `false`.

### Regression?

_No response_

### Esri team

ArcGIS Field Apps

Elijbet:
**Related Issue:** #5333 

Resolves #5464 as well, stems from #4814.

## Summary

Adds optional chaining to get rid of the `typeError` in cases when there is no `parentEl`.

Project #adoptatree is good to go.


src/components/tree/tree.e2e.ts
```diff
@@ -904,5 +904,50 @@ describe("calcite-tree", () => {
...
+      await page.evaluate(async (): Promise<string> => {
+        customElements.define(
+          "test-tree-element",
+          class extends HTMLElement {
+            constructor() {
+              super();
+            }
+
+            connectedCallback() {
+              this.attachShadow({ mode: "open" }).appendChild(
```
jcfranco, line 940: Nitpick: use `append`.

src/components/tree/tree.e2e.ts
```diff
@@ -904,5 +904,50 @@ describe("calcite-tree", () => {
 
       expect(await getActiveElementId(page)).toEqual("child-1");
     });
+
+    it("renders when tree is in the root node (there is no parent)", async () => {
```
jcfranco, line 908: Test name is misleading, this is testing when the tree is the topmost element in a shadow root where it has no parent.

src/components/tree/tree.e2e.ts
```diff
@@ -904,5 +904,50 @@ describe("calcite-tree", () => {
...
+                (document.getElementById("test-tree-element-template") as HTMLTemplateElement).content.cloneNode(true)
+              );
+            }
+          }
+        );
+        return null;
+      });
+      await page.waitForChanges();
+
+      renders("calcite-tree", { display: "block" });
```
jcfranco, line 950: This test should assert that no error is thrown instead of using `renders` since the bug is prone to happen after initial render (e.g., in `updateAncestorTree`).

src/components/tree/tree.e2e.ts
```diff
@@ -904,5 +904,50 @@ describe("calcite-tree", () => {
...
+                <calcite-tree-item>Great-grandchild 2</calcite-tree-item>
+              </calcite-tree>
+            </calcite-tree>
+          </calcite-tree-item>
+          <calcite-tree-item id="three">3</calcite-tree-item>
+        </calcite-tree>
+      </template>`;
+
+      const page = await newE2EPage();
+      await page.setContent(html`${templateHTML} <test-tree-element></test-tree-element>`);
```
jcfranco, line 928: Nitpick: for readability, can you move the template and test element to separate lines?

```ts
html`
  ${templateHTML}
  <test-tree-element></test-tree-element>
`
```
?

src/components/tree/tree.e2e.ts
```diff
@@ -904,5 +904,50 @@ describe("calcite-tree", () => {
...
+      await page.waitForChanges();
+
+      await page.evaluate(async (): Promise<string> => {
+        customElements.define(
+          "test-tree-element",
+          class extends HTMLElement {
+            constructor() {
```
jcfranco, line 935: The constructor isn't doing anything, so let's drop it.

src/components/tree/tree.e2e.ts
```diff
@@ -904,5 +904,50 @@ describe("calcite-tree", () => {
...
+            }
+
+            connectedCallback() {
+              this.attachShadow({ mode: "open" }).appendChild(
+                (document.getElementById("test-tree-element-template") as HTMLTemplateElement).content.cloneNode(true)
+              );
+            }
+          }
+        );
+        return null;
```
jcfranco, line 946: This doesn't have to return anything.

src/components/tree/tree.e2e.ts
```diff
@@ -904,5 +904,50 @@ describe("calcite-tree", () => {
 
       expect(await getActiveElementId(page)).toEqual("child-1");
     });
+
+    it("renders when tree is in the root node (there is no parent)", async () => {
+      const templateHTML = html` <template id="test-tree-element-template">
+        <calcite-tree>
```
jcfranco, line 910: Is this the minimal repro case? I think the error would happen with a single tree and tree item, no?

jcfranco (approved):
This is great, @Elijbet. Aside from some test comments, this LGTM!

jcfranco (commented):
👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐
👐👐🌳👐🌳👐👐🌳🌳🌳🌳👐🌳🌳🌳👐👐👐🌳🌳👐👐🌳🌳🌳👐👐🌳🌳🌳🌳🌳👐🌳🌳🌳🌳👐🌳🌳🌳🌳🌳👐🌳🌳🌳👐👐🌳🌳🌳🌳👐🌳🌳🌳🌳👐
👐🌳🌳🌳🌳🌳👐🌳👐👐🌳👐🌳👐👐🌳👐🌳👐👐🌳👐🌳👐👐🌳👐👐👐🌳👐👐👐🌳👐👐🌳👐👐👐🌳👐👐👐🌳👐👐🌳👐🌳👐👐👐👐🌳👐👐👐👐
👐👐🌳👐🌳👐👐🌳🌳🌳🌳👐🌳👐👐🌳👐🌳👐👐🌳👐🌳🌳🌳👐👐👐👐🌳👐👐👐🌳🌳🌳🌳👐👐👐🌳👐👐👐🌳🌳🌳👐👐🌳🌳🌳👐👐🌳🌳🌳👐👐
👐🌳🌳🌳🌳🌳👐🌳👐👐🌳👐🌳👐👐🌳👐🌳👐👐🌳👐🌳👐👐👐👐👐👐🌳👐👐👐🌳👐👐🌳👐👐👐🌳👐👐👐🌳👐🌳👐👐🌳👐👐👐👐🌳👐👐👐👐
👐👐🌳👐🌳👐👐🌳👐👐🌳👐🌳🌳🌳👐👐👐🌳🌳👐👐🌳👐👐👐👐👐👐🌳👐👐👐🌳👐👐🌳👐👐👐🌳👐👐👐🌳👐👐🌳👐🌳🌳🌳🌳👐🌳🌳🌳🌳👐
👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐👐

src/components/tree/tree.e2e.ts
```diff
@@ -904,5 +904,39 @@ describe("calcite-tree", () => {
...
+                (document.getElementById("test-tree-element-template") as HTMLTemplateElement).content.cloneNode(true)
+              );
+            }
+          }
+        );
+      });
+      await page.waitForChanges();
+
+      renders("calcite-tree", { display: "block" });
+      await expect(() => defineCustomElement).not.toThrow(TypeError);
```
jcfranco, line 939: I _think_ this needs to wait for page changes in the function that goes over the error scenario.

Can you try the following updates:

```ts
const defineTestElementAndWaitForChanges = async () => {
  await page.evaluate(async (): Promise<void> => {
        customElements.define(
          // ...
  });
  await page.waitForChanges();
};

// no `renders` call as this should be captured in `defineTestElementAndWaitForChanges`

await expect(() => defineTestElementAndWaitForChanges).not.toThrow(Error);
```
?

If this fails when you roll back your fix and passes when you restore them. This should be good to go.

jcfranco (changes requested):
One more pass on the test and I think this will be good to merge!

src/components/tree/tree.e2e.ts
Elijbet: Yeah, it fails when the wait is moved inside the function. Good point about the render call.

macandcheese:
A accepted Screener - it was an unrelated new screenshot from Block changes.

src/components/tree/tree.tsx
```diff
@@ -283,7 +283,7 @@ export class Tree {
       }
 
       // When focus is on a child node that is also either an end node or a closed node, moves focus to its parent node.
-      const parentItem = target.parentElement.closest("calcite-tree-item");
+      const parentItem = target.parentElement?.closest("calcite-tree-item");
```
jcfranco, line 286: Per our convo, this and the updates below (`updateAncestorTree`) wouldn't need the optional chaining operator as the event is processed after it bubbles, so we can assume there's an ancestor chain. This can be refactored later.

jcfranco (approved):
✨🌳✨

Issue #5399, opened 2022-09-29
Popover reference element position issues


rslibed:
### Actual Behavior

When opening a popover with a trigger element, the popover will be incorrectly positioned for a brief moment, then correct itself. 

### Expected Behavior

When opening a popover with a trigger element, the popover will correctly open next to the reference element

### Reproduction Sample

https://codepen.io/rslibed/pen/bGMvBPL

### Reproduction Steps

1. Open CodePen
2. Click on `i` icons to open popovers. 
3. Notice that the popovers are incorrectly positioned and eventually corrects itself. 

Note: Issue is intermittent, so it may take a few refreshes to repro. 

### Reproduction Version

beta.95

### Relevant Info

Chrome, Mac OS

### Regression?

beta.92

### Esri team

ArcGIS Instant Apps

ethanbdev:
+1 watching this one

jmanke:
Here is another [sample](https://codepen.io/jmanke/pen/LYmMbJb) that is reproducible 100% of the time on my machine (using Chrome).

+1 Dashboards is watching this one as well

PR #5484, opened 2022-10-14
fix: fix jarring positioning when a closed component is first opened

jcfranco:
**Related Issue:** #5399 

## Summary

This ensures proper positioning of `FloatingUIComponent`s by addressing a few items:

* sets floating element's `position` to match the `overlay-positioning` value for initial positioning
* avoids positioning components if an arrow should be included in positioning calculation phase

Additionally, this refactors the `floating-ui.ts` module to provide a convenience function to make it easier for components to implement `reposition` in a consistent, and performant fashion (e.g., debounce batch calls, avoid positioning on closed components).

### Notes

* When positioned for the first time, the transitioned element may transition in from a different direction (IMO, barely noticeable). Subsequent transitions should work as intended.
* This issue affected any component with `overlay-positioning="fixed"` as the default is `absolute` and would trip up the initial positioning.
* An additional utility for when a floating-UI element/component is closed was introduced to reset the styles for absolutely-positioned elements. This is necessary to avoid having hidden floating elements from affecting layout. Worth noting that this will share the same issue as https://github.com/Esri/calcite-components/issues/5206, but shares the same workaround (to set a very small duration). This should be refactored once #5206 lands for a more consistent approach. 
* Spec tests have also been added to cover the latest changes.

### Questions

* `reposition`'s signature was updated to include a `delayed` boolean to toggle between 1. immediate repositioning calls (e.g., when resizing the page) vs. 2. debounced calls (e.g., when multiple props may be updated that trigger repositioning).
  a. Does `delayed` make sense as a name? 
  b. Should we doc this? One reason to do so is for wrapping components to be able to use debounced repositioning (2)

~**Note:** I tried to set up a spec test for this, but ran into some issues:~

* ~`ResizeObserver` not being defined when using `floating-ui` utils~ – enabled `autoUpdate` (`floating-ui`) for browser contexts only
* ~`jest.spyOn()` not working on the `floating-ui` helper module (spies not being called)~ – used jest's `moduleNameMapper` to load the CJS version of lodash in spec tests instead of the ESM one, which will cause an error)

~I'll revisit this in a future PR.~

### Pending

- updating spec tests to cover moving-offscreen changes (I'll tackle this in a follow-up PR)

cc @driskull


src/components/popover/popover.stories.ts
```diff
@@ -101,3 +101,22 @@ export const flipPlacements_TestOnly = (): string => html`
...
+  </calcite-panel>
+  <script>
+    (async function () {
+      await customElements.whenDefined("calcite-popover");
+      const popover = document.querySelector("calcite-popover");
+      await popover.componentOnReady();
+
+      // need to set open after initial render to reproduce the issue
+      popover.open = true;
+    })();
```
benelan, line 120: I don't think this will work, it probably needs to be an interaction test eventually.

src/utils/floating-ui.ts
```diff
@@ -380,6 +380,9 @@ export function connectFloatingUI(
 
   disconnectFloatingUI(component, referenceEl, floatingEl);
 
+  // ensure position matches for initial positioning
+  // floatingEl.style.position = component.overlayPositioning;
+
```
benelan, line 385: How is this a fix, there is no change to the component code other than comments?

src/components/popover/popover.stories.ts
jcfranco: Switching over to steps.

src/utils/floating-ui.ts
jcfranco: I forgot to uncomment after some local testing. My bad!

benelan (approved):
LGTM!

jcfranco:
Moving this back to draft. Ran an example where this is still noticeable. 

I think we'll need to roll back the debouncing. Might not be impactful since `floating-ui` reverted the performance-affecting code.

jcfranco:
I think I figured out the cause for the initial position flash. The repositioning logic is debounced, but the prop that toggles the floating-ui animation class is set as `toggle` is opened. Hence why the component is visible right before it shifts to the right spot. I see two options:

1. roll back debouncing
2. provide a hook for components to pass any classes to apply right when a component will be repositioned.

As hinted previously, going with 1 for now.

src/utils/floating-ui.ts
```diff
@@ -380,6 +380,9 @@ export function connectFloatingUI(
 
   disconnectFloatingUI(component, referenceEl, floatingEl);
 
+  // ensure position matches for initial positioning
+  floatingEl.style.position = component.overlayPositioning;
```
driskull, line 384: 👍

jcfranco:
Tried option 1., but noticed there's a noticeable perf impact when dealing with multiple floating-ui-owning components (similar to https://github.com/Esri/calcite-components/issues/5286). Going to give option 2 a try.

jcfranco:
I think I've got a viable solution for this. Coming up!

src/utils/floating-ui.ts
```diff
@@ -380,13 +433,21 @@ export function connectFloatingUI(
...
+
+  const runAutoUpdate = Build.isBrowser
+    ? autoUpdate
+    : (_refEl: HTMLElement, _floatingEl: HTMLElement, updateCallback: Function): (() => void) => {
```
driskull, line 441: 🆒

src/utils/floating-ui.ts
```diff
@@ -301,12 +347,14 @@ export async function positionFloatingUI({
...
   flipPlacements?: EffectivePlacement[];
+
   offsetDistance?: number;
   offsetSkidding?: number;
   arrowEl?: HTMLElement;
+  includeArrow?: boolean;
   type: UIType;
 }): Promise<void> {
-  if (!referenceEl || !floatingEl) {
+  if (!referenceEl || !floatingEl || (includeArrow && !arrowEl)) {
```
driskull, line 357: nice

src/utils/floating-ui.ts
```diff
@@ -380,13 +433,21 @@ export function connectFloatingUI(
...
+        updateCallback();
+        return () => {
+          /* noop */
+        };
+      };
+
   cleanupMap.set(
     component,
-    autoUpdate(referenceEl, floatingEl, () => {
-      if (component.open) {
```
driskull, line 386: No more open check right? I don't think we need the open check anymore.

driskull (approved):
>  Does delayed make sense as a name?

Yeah, I think so. Or maybe `scheduleReposition`? Like scheduleRender?

> Should we doc this? One reason to do so is for wrapping components to be able to use debounced repositioning (2)

I think so.

src/utils/floating-ui.ts
jcfranco: The `reposition` util takes care of it now, so no need for it here.

jcfranco:
@driskull I updated this to handle the case where hidden popovers would affect layout. We can iterate on this if needed, but  it seems to work without further affecting transitions. FWIW, I tried the suggested removal of `visibility` in our styles, but this still affected layout.

driskull:
> FWIW, I tried the suggested removal of visibility in our styles, but this still affected layout.

The suggestion to remove `visibility` is for hiding the FloatingUI when the reference is no longer in view. Currently, the floating-ui JS util and styles are competing and its not working.

Issue #5428, opened 2022-10-10
Calcite-Panel: content is no longer wrapped in a <section> element when panel.hasFab is false


zaramatheson:
### Actual Behavior

The panel content is no longer wrapped in a `<section>` element when panel.hasFab is false, which refers to a "floating action button". Not totally sure if this was an intended change. (Note: This is also the case when replacing the panel with flow-item.)

It is causing issues with button display in the panel's footer in Map Viewer. 

cc @jcfranco 

![dev screenshot](https://user-images.githubusercontent.com/5023024/194960772-95d5f4bf-a83b-47f4-a88a-679c2bf41b5f.png)

### Expected Behavior

panel content wrapped in a `<section>` element.

More specifically, the panel/flow-item is missing the `display: flex;` from the `.content-container`, which is resulting in an `overflow: auto;` from `.content-wrapper` to display the scrollbar.

The `Editor` widget typically relies on a child of the `<calcite-panel>` to display a scrollbar around its own content (FeatureForm, Attachments) so the main container element inside the panel/flow-item container may need `display: flex;` at the minimum to override the `overflow: auto;`.

![prod screenshot](https://user-images.githubusercontent.com/5023024/194960793-88a046ef-7047-4cef-8497-813018443f92.png)

### Reproduction Sample

https://jsbin.com/jisujuyofu/edit?html,output

Other samples:
- https://codepen.io/banuelosj/pen/poVBZoV
- Also, a [more advanced use case with JSAPI](https://codepen.io/hgonzago/pen/ExLQazO)

### Reproduction Steps

1. Visit the [sample](https://jsbin.com/jisujuyofu/edit?html,output)
2. Observe the scrollbar in panel

### Reproduction Version

beta.95

### Relevant Info

- Issue was first found in version 91
    - Most recent version where this issue was NOT reproducible was version 86.
- If remove the `<calcite-tab>` and `calcite-tabs` components in the example and keep the `calcite-flow` and `calcite-panel`, then it works as-expected.

### Regression?

Potentially related to the https://github.com/Esri/calcite-components/pull/5129 update [[Commit]( https://github.com/Esri/calcite-components/commit/6f6316c8532e1876aadc63f71c3e4095db2af3b6#diff-e58f98beb295fc3692d5fe7bbc2c3d1dc8a0b4379ebacd2de304680f9d7ad7f3L521)]. 

### Esri team

ArcGIS Map Viewer

geospatialem:
[...]

PR #5486, opened 2022-10-15
fix(flow-item, panel): fix layout issue that would cause double scrollbars

jcfranco:
**Related Issue:** #5428

## Summary

https://github.com/Esri/calcite-components/pull/5129 refactored panel's content rendering function and it missed toggling some CSS classes on the parent node based on whether there was a fab or not (before: https://github.com/Esri/calcite-components/blob/6f6316c8532e1876aadc63f71c3e4095db2af3b6~/src/components/panel/panel.tsx#L533 + https://github.com/Esri/calcite-components/blob/6f6316c8532e1876aadc63f71c3e4095db2af3b6~/src/components/panel/panel.tsx#L523; after: https://github.com/Esri/calcite-components/blob/6f6316c8532e1876aadc63f71c3e4095db2af3b6/src/components/panel/panel.tsx#L629 ).

**Notable changes:**

- adds missing flow-item stories (to include in screenshot testing)
- removes unnecessary keys for nodes that are no longer conditionally rendered

cc @driskull

### DIFF
--- a/src/components/flow-item/flow-item.stories.ts
+++ b/src/components/flow-item/flow-item.stories.ts
@@ -13,7 +13,7 @@ import { html } from "../../../support/formatting";
 import { storyFilters } from "../../../.storybook/helpers";
 
 export default {
-  title: "Components/Panel",
+  title: "Components/Flow Item",
   parameters: {
     notes: readme
   },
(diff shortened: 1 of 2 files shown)


src/components/flow-item/flow-item.stories.ts
```diff
@@ -0,0 +1,219 @@
...
+        commit(): Attribute {
+          this.value = boolean("loading", false);
+          delete this.build;
+          return this;
+        }
+      },
+      {
+        name: "intl-close",
+        commit(): Attribute {
+          this.value = text("intlClose", TEXT.close);
```
benelan, line 71: there is no `TEXT.close` in `flow-item.resources.ts`. Either add it to the resources file or hardcode a value this line.

benelan (approved):
Look at all those properly exported stories 🥳🍾🥂🎆 

One small thing but otherwise LGTM

src/components/flow-item/flow-item.stories.ts
jcfranco: Yikes, good catch! I went with reusing panel's resources file as `flow-item` delegates to it.

PR #5490, opened 2022-10-17
fix(input, input-number): decimals no longer contain groupSeparators and remove leading zeros

benelan:
**Related Issue:** N/A

Repro sample: https://codepen.io/benesri/pen/zYaZaGX

## Summary

This fixes a bug caused by adding the `numberingSystem` property to components. We have to format the integers and decimals separately since we support yuuge numbers using the `BigDecimal` class. This caused long decimal numbers to have group separators, which I fixed by formatting each digit individually. It also caused leading zeros in decimals to be removed, e.g. `0.00123` became `0.123`.

I also did some small cleanup in BigDecimal like caching regex and removing duplicate code


src/utils/number.spec.ts
```diff
@@ -111,6 +111,15 @@ describe("BigDecimal", () => {
     expect(negativeZero).toBe("-0");
   });
 
+  it("correctly formats long decimal numbers", () => {
+    numberStringFormatter.numberFormatOptions = {
+      locale: "en",
+      numberingSystem: "latn",
+      useGrouping: true
+    };
+    expect(new BigDecimal("123.123456789").format(numberStringFormatter)).toBe("123.123456789");
```
alisonailea, line 120: can we add one or two other "expects" here to confirm this change for other locales and non latin numbering systems?

benelan: I added coverage for all of the supported numbering systems in the locale spec tests. `numberStringFormatter.localize` uses `BigDecimal.format`.

src/utils/locale.spec.ts
```diff
@@ -35,7 +36,7 @@ describe("NumberStringFormat", () => {
   });
 
   locales.forEach((locale) => {
-    it(`integers localize and delocalize in "${locale}"`, () => {
+    it(`locale: integers localize and delocalize in "${locale}"`, () => {
```
jcfranco, line 39: You could place these in a `describe("locale", () => { /*...*/` block for grouping instead of updating the individual test names.

jcfranco (approved):
✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨
✨🔢✨✨✨🔢✨🔢🔢🔢✨✨🔢🔢🔢✨🔢🔢🔢🔢✨🔢✨
✨🔢🔢✨✨🔢✨✨🔢✨✨🔢✨✨✨✨🔢✨✨✨✨🔢✨
✨🔢✨🔢✨🔢✨✨🔢✨✨🔢✨✨✨✨🔢🔢🔢✨✨🔢✨
✨🔢✨✨🔢🔢✨✨🔢✨✨🔢✨✨✨✨🔢✨✨✨✨✨✨
✨🔢✨✨✨🔢✨🔢🔢🔢✨✨🔢🔢🔢✨🔢🔢🔢🔢✨🔢✨
✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨

PR #5512, opened 2022-10-18
ci: cleanup 3rd party action dependencies

benelan:
**Related Issue:** #5478

## Summary

Cleans up a couple 3rd party dependencies that may cause us issues. Some of these ones were easily replaced with `actions/github-scripts` which is maintained by GitHub. I also added the [`workflow_dispatch`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch) to more of the actions we may want to run manually, like creating new milestones.

### NPM Audit

I had to completely delete the audit action because `npm audit` runs `npm install` internally and there is a bug that prevents passing `--legacy-peer-deps`.
- [Error message here](https://github.com/Esri/calcite-components/actions/runs/3269835363/jobs/5377797979#step:4:43)

### Browserslist

The browserlist action has also been failing due to not installing with `--legacy-peer-deps`. But it was easy to do it manually, which allows us to install with `--legacy-peer-deps` before using `npx update-browserslist-db@latest` (which does `npm install caniuse-lite` internally)
- [Error message here](https://github.com/Esri/calcite-components/actions/runs/3257644462/jobs/5349040957#step:4:96)

This was tested and confirmed working on my test repo, where I replicated our deps problem.

### PR Bot

This PR removes the job that assigns the author to the PR. It's annoying getting emails everytime you log a PR. @jcfranco what was the use case for wanting the author automatically assigned? If it was for filtering the CC PRs when trying to find one you wrote in the past, you can do that by `author`. If you want to see your current PRs, you can check [here](https://github.com/pulls). If there is another use case I'm missing, I'm happy to add it back. It will only take a couple lines in a `actions/github-scripts` job.

I also replaced 3rd party labeler action with a simple `actions/github-scripts` one. The changes work on my test repo

### Release Teams Notification

Replaced the unmaintained Action with a simple  `actions/github-scripts` one. The changes work on my test repo, I used a new private channel to make sure.


jcfranco:
> what was the use case for wanting the author automatically assigned? If it was for filtering the CC PRs when trying to find one you wrote in the past, you can do that by author. If you want to see your current PRs, you can check [here](https://github.com/pulls). If there is another use case I'm missing, I'm happy to add it back. It will only take a couple lines in a actions/github-scripts job.

I _think_ it was to help filter consistently for both issues and PRs since PR author doesn't necessarily mean assignee (although for 99% of the time this is the case). There have also been some PRs w/ multiple assignees, so going by author will likely not work the same. Can we restore this and revisit later?

jcfranco (approved):
Aside from the comment about PR assignment, this LGTM! 🎉

Issue #5445, opened 2022-10-11
Regression with `<calcite-tab>` and children with percentage heights


nwhittaker:
### Actual Behavior

Given a `<calcite-tabs>` element with a defined height, placing a block element with a percentage height inside a `<calcite-tab>` causes the element (red) to hug its content rather than fill up a percentage of the tab area (pink):

<img width="324" alt="Screen Shot 2022-10-11 at 12 04 59 PM" src="https://user-images.githubusercontent.com/421496/195143190-8c3f5c84-00d7-4d6e-a678-5717ab1d8bbc.png">

### Expected Behavior

The element (red) does not hug its content but instead fills up a percentage (in this case, `100%`) of the tab area (pink):

<img width="324" alt="Screen Shot 2022-10-11 at 12 05 20 PM" src="https://user-images.githubusercontent.com/421496/195143164-0c6ab5e2-bfe1-4ab6-8d9f-69aae836b7f0.png">

### Reproduction Sample

1. https://codepen.io/nwhittaker-esri/pen/bGMQjob (layout)
2. https://codepen.io/nwhittaker-esri/pen/JjvVLPg (scrolling)

### Reproduction Steps

Layout issue:
1. Visit the first code sample
2. See layout matches actual behavior instead of expected behavior

Scrolling issue:
1. Visit the second code sample
2. Scroll the red area and notice the **Scroll top** output does not change, but it should be updating as the content as scrolled
3. To see the expected behavior, pull in the `next.552` deps instead

### Reproduction Version

beta.95

### Relevant Info

Appears to have been introduced by https://github.com/Esri/calcite-components/pull/5061.

One apparent workaround is to give the `<calcite-tab>` flex display, but that introduces an issue where the child element no longer respects a percentage width.

Field Maps has a feature where the tab area should fill the height of the window and dragging and dropping onto that area adds to the content of the tab. With this regression, this behavior is no longer possible and an alternative implementation is a large effort due to how the various layout templates are sliced up.

### Regression?

next.552

### Esri team

ArcGIS Field Apps

nwhittaker:
Adding that this regression is also impacting some scroll behaviors Field Maps has when dragging an element towards the top or bottom of the tab.

nick-esri:
@benelan - it would be great to get this one addressed. It's causing a pretty obvious visual defect in Field Maps web. Thanks!

nwhittaker:
[...]

PR #5516, opened 2022-10-18
fix(tab): applies section styles onto the enclosing parent

Elijbet:
**Related Issue:** #5445

## Summary
Reverts regression introduced by #5061 by carrying section styles onto the new parent.


src/components/tab/tab.scss
```diff
@@ -1,6 +1,6 @@
 :host([selected]) {
   section,
-  :has(section) {
+  [role="tabpanel"] {
```
jcfranco, line 3: Let's use a, new, CSS class instead. I'll suggest `.container` for consistency.

jcfranco (approved):
Aside from the CSS class suggestion, this LGTM!

Issue #5513, opened 2022-10-18
calcite-popover should only display content on left-click


richiecarmichael:
### Actual Behavior

[calcite-popover](https://developers.arcgis.com/calcite-design-system/components/popover/) is displaying content on right click. 

### Expected Behavior

[calcite-popover](https://developers.arcgis.com/calcite-design-system/components/popover/) should ignore right-clicks.

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/popover/

### Reproduction Steps

1. Open [sample page](https://developers.arcgis.com/calcite-design-system/components/popover/)
2. Right click on "Activate Popover" button.
  - "Popover content here" is displayed. It should not.

### Reproduction Version

unknown

### Esri team

ArcGIS API for JavaScript

geospatialem:
Verified on `next.608`.

![verify-left-click-only-popover](https://user-images.githubusercontent.com/5023024/197832663-2c2be3f0-ec6c-4463-b378-be7d1d13f50c.gif)

PR #5519, opened 2022-10-18
fix: components should only react to primary button pointer events.

driskull:
**Related Issue:** #5513

## Summary

fix: components should only react to primary button pointer events.

This PR checks that the button is `0` for `pointerup` and `pointerdown` events throughout the components.

https://www.w3.org/TR/pointerevents/#the-button-property

image


driskull:
@jcfranco should we also have a check for `pointerup` events?

src/utils/dom.ts
```diff
@@ -289,3 +289,15 @@ export function intersects(rect1: DOMRect, rect2: DOMRect): boolean {
...
+
+/**
+ * This helper returns true if the pointer event fired from the primary button of the device.
+ *
+ * See https://www.w3.org/TR/pointerevents/#the-button-property.
+ *
+ * @param event
+ * @returns {boolean}
+ */
+export function isPrimaryPointerButton(event: PointerEvent): boolean {
```
jcfranco, line 301: Can you add a spec test for this?

src/components/slider/slider.tsx
```diff
@@ -989,12 +993,16 @@ export class Slider
     return ticks;
   }
 
-  private dragStart(prop: ActiveSliderProperty): void {
```
jcfranco, line 992: Nitpick: for consistency, can you bring this back and use it in `pointerDownDragStart` (similar to `pointerDownDragEnd`)?

jcfranco (approved):
Thanks for fixing this! 🐁👈

Side bar: we should revisit how we're using pointer events as clicks/taps. `pointerdown` & `click` fire at different times and with different context (e.g., click has a slight delay before firing, and `mousedown` & `mouseup` needs to fire from the same element), so this could throw some users off. Maybe we could bring back listening to `click` and for the rest, we can rely on pointer events.

src/utils/dom.ts
driskull: Yesssir \( ﾟヮﾟ)/

src/components/slider/slider.tsx
driskull: done! (*・‿・)ノ⌒*:･ﾟ✧

jcfranco (approved):
💥

Issue #5245, opened 2022-08-30
[Button / FAB] Improve scale of internally rendered icons


macandcheese:
### Description

Noted as feedback on Button / FAB - but possibly relevant in other places where icons are rendered internally.

Button / FAB of scale S, M, L all render a `<calcite-icon>` of scale S internally. We should consider bumping the scale of that icon to M when the parent Button / FAB is scale L.

Unfortunately the icons don't scale with the component sizes, but in the case of L Button / FAB - the rendered icon is too small.

### Acceptance Criteria

Button / FAB (and other components if needed) render an M size icon from provided Calcite UI Icon strings instead of S

### Relevant Info
cc @ashetland 

### Which Component

Button / FAB

### Example Use Case

Using a FAB or Button of scale L - the rendered S icon is comically small, and takes away from the intent of using a large button / fab to begin with.

ashetland:
Is this something we want to look at for v1/ui-kit-v2?

macandcheese:
If possible - from a code standpoint its a small change so we'd just need to look at design implications.

ashetland:
On the Design and Figma side, we would be good to include this update to Button and FAB. Padding adjustments would be made to maintain height.

Should we look at applying this approach across Calcite? For example, Chip uses a S icon at the L scale.

jcfranco:
Priority bumped and triaged to this sprint. cc @brittneyzwolfer @benelan @geospatialem

adepottey:
@jcfranco Checking in on this - are we still on track to get a fix for this by mid-November?  I see the milestones have shifted a few times.

Elijbet:
Here is a list of relevant Items with a similar issue.  What do we think about these? @ashetland @macandcheese @jcfranco 
Accordion-Item
Action
Chip
Combobox and Combobox-item
Dropdown
Input and input-date-picker
Link
Notice
Pick-list-item
Radio-group
Tab-title

jcfranco:
@Elijbet Thanks for putting that list together. I'll defer to @ashetland @macandcheese on that.

@adepottey I meant to reply sooner, but yes, this should be available by mid-November.

ashetland:
We discovered an issue with this related merged PR (https://github.com/Esri/calcite-components/pull/5521) that needs a design adjustment. It looks like the height of the component changes when the new 24px icon is present. Our recommendation would be to maintain the 44px height with or without icon.

[...]

PR #5521, opened 2022-10-19
fix(button, fab): bumping the scale of icon to M when the parent Button / FAB is scale L

Elijbet:
**Related Issue:** #5245

## Summary

Bumping the scale of `icon` to M when the parent `button `/ `fab` is scale L for a visual distinction between larger and smaller components.

Added stories and cleaned up the demo for `fab`, which was missing `scale="l"` visuals.

### DIFF
--- a/src/demos/fab.html
+++ b/src/demos/fab.html
@@ -135,7 +135,7 @@
 
           <div>
             <calcite-fab
-              scale="m"
+              scale="l"
               icon="arrow-left"
               label="FAB"
               text="FAB"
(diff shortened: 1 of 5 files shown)


src/demos/fab.html
```diff
@@ -427,7 +427,7 @@
 
           <div>
             <calcite-fab
-              scale="m"
```
jcfranco, line 430: Nice, great catch!

jcfranco (approved):
As we discussed, let's drop the scale-specific story since we're not covering these in our screenshot tests to keep our quota under control (permutations can get out of hand).

Code-wise, this LGTM! 🎉

I'll let @macandcheese and @ashetland chime in on the design changes.

ashetland:
Design is good to go.

Issue #5520, opened 2022-10-19
Refactor to drop the optional chaining operator ? from tree within the event handler.


Elijbet:
### Description

In the context of events from child items, there is no need for the optional chaining operator as the event is processed after it bubbles, so we can assume there's an ancestor chain.

### Proposed Advantages

Remove redundancy.

### Which Component

`calcite-tree`

### Relevant Info

Stems from changes in #5472

geospatialem:
Verified on `next.608`

PR #5523, opened 2022-10-19
refactor(tree): drop the optional chaining operator ? from tree within the event handler

Elijbet:
**Related Issue:** #5520

## Summary

Some of the optional chaining `?` added to get rid of the typeError in cases when there is no `parentEl` (#5472) was redundant. In the context of events from child items, there is no need for the `?` operator as the event is processed after it bubbles, so we can assume there's an ancestor chain.


jcfranco (approved):
💥

PR #5526, opened 2022-10-19
fix(slider): Only focus handle if handle element is defined.

driskull:
**Related Issue:** N/A

## Summary

fix(slider): Only focus handle if handle element is defined.


jcfranco:
FYI, I've got a fix coming up for [this](https://github.com/Esri/calcite-components/issues/5321) with accompanying tests. The issue is not whether the thumbs are rendered, but when the method is called in the `pointerdown` handler.

driskull:
cool i'll close this one then.

Issue #5321, opened 2022-09-14
calcite-slider throws error


AdelheidF:
### Actual Behavior

When clicking left of handle on slider an error occurs and no calciteSliderInput event gets called. After that nothing works anymore. This started with beta.91.

![image](https://user-images.githubusercontent.com/44481292/190230955-f1cd3a76-73d0-4f83-a7b6-20eee7de5740.png)

### Expected Behavior

no errors

### Reproduction Sample

https://codepen.io/afreitag/pen/ExLNOeN

### Reproduction Steps

open the codepen and the browser console
click twice on the right side of the handle, all good
click left of the handle and you'll see the error

### Reproduction Version

beta.93 and latest next

### Regression?

beta.90

RalucaNicola:
I ran into the same issue with the calcite react components. I have a repro here: https://codesandbox.io/s/brave-benz-cfi8vg?file=/src/index.js

AdelheidF:
looks good with next.603

PR #5533, opened 2022-10-20
fix(slider): fix slider (single-value) error when clicking range

jcfranco:
**Related Issue:** #5321 

## Summary

This fixes a regression from https://github.com/Esri/calcite-components/commit/24a6b9548e8eaa2ab8b8c9faa05869c0c29d6136 where the thumb was being focused before the drag prop was set, causing an error to be thrown when clicking on the track, on opposite sides of the thumb.

### Notes

* `activeProp` is no longer a state prop as it does not affect rendering (takes care of Stencil dev-time rendering warning)
* makes slider UX more consistent by focusing the single-thumb immediately on `pointerdown` (currently, this only happens with range slider)

### DIFF
--- a/src/components/slider/slider.e2e.ts
+++ b/src/components/slider/slider.e2e.ts
@@ -270,6 +270,60 @@ describe("calcite-slider", () => {
     expect(changeEvent).toHaveReceivedEventTimes(1);
   });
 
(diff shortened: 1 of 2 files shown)


driskull (approved):
👍

Issue #3821, opened 2021-12-30
[calcite-loader, calcite-input-message] Drop `active` in favor of `hidden`


jcfranco:
### Description 

Stems from https://github.com/Esri/calcite-components/pull/3778/files#r774706383

It seems that `calcite-loader/calcite-input-message[active]` is used to toggle visibility, so we should deprecate/remove it in favor of the global `hidden` attribute.

https://codepen.io/jcfranco/pen/KKXQErV?editors=1000

To avoid breaking changes:

* `active` should be deprecated 
* styles for `hidden` should be added (these have higher precedence than `active`)

### Proposed Advantages 

One less custom prop to manage, and it aligns with existing HTML conventions.

### Relevant Info 

N/A

#### Which Component

* `calcite-loader`
* `calcite-input-message`

driskull:
Agreed.

We do have some components that are not visible by default...

- modal
- popover
- tooltip
- ?

However, these use `visibility`  to hide them instead of `display`. Also, we would expect those components to not be visible by default. In this case, I think I would expect loader to be visible by default.

macandcheese:
Do we need to allow this to be configured at all? 

Should users just be adding and removing the element, or does having the ability to toggle visibility via prop aid in a transition or fade-in or something? cc @asangma

driskull:
> Do we need to allow this to be configured at all?

I don't think so. We just choose what we want to do.

> Should users just be adding and removing the element, or does having the ability to toggle visibility via prop aid in a transition or fade-in or something? 

That's a good question. If a user had to construct a `calcite-modal` at the time that the user wanted to display it would that be problematic? It might be an inconvenience. Currently, a user can just write the HTML for a modal and then enable it at a later time.

As for `hidden` vs `active`... hidden will set `display:none` which might be problematic for animations of modals/alerts/tooltips/popovers. I think that's why we are using our own property, to handle animation. We might be able to do animation using `hidden` but it might be a little tricky.

driskull:
In the case of `calcite-loader`, we probably don't need an active property and can just use the hidden attribute.

I don't see any animation going on when setting active for this component. It's also not a component that contains any content, so there's no reason not to just remove it or set hidden property on it.

jcfranco:
I updated the issue to include `calcite-input-message`. I think the other components brought up earlier can stay as is since, `active`/`open` is linked to their behavior . Loader and input message are more like atoms that can be displayed/hidden by an owning component.

[...]

PR #5537, opened 2022-10-21
fix(calcite-loader, calcite-input-message): drop active in favor of hidden

Elijbet:
**Related Issue:** #3821

## Summary

Loader and Input-Message are small building blocks that an owning component should be able to display or hide.

This deprecates/removes the role of `active` prop in toggling visibility in favor of the global attribute `hidden`. Styles for hidden take precedence over active. 

The benefit is one less custom prop to manage, and it aligns with existing HTML conventions.

### DIFF
--- a/src/components/input-message/input-message.tsx
+++ b/src/components/input-message/input-message.tsx
@@ -33,13 +33,6 @@ export class InputMessage {
    */
   @Prop({ reflect: true }) active = false;
 
-  /**
-   * When `true`, the component will not be visible.
-   *
-   * @mdn [hidden](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden)
-   */
-  @Prop({ reflect: true }) hidden = false;
-
   /** Specifies an icon to display. */
   @Prop({ reflect: true }) icon: boolean | string;
 
--- a/src/components/input-message/input-message.tsx
+++ b/src/components/input-message/input-message.tsx
@@ -30,6 +30,7 @@ export class InputMessage {
    * When `true`, the component is active.
    *
    * @deprecated use global `hidden` attribute instead.
+   * @mdn [hidden](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden)
    */
   @Prop({ reflect: true }) active = false;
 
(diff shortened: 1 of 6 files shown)


src/components/input-message/input-message.tsx
```diff
@@ -26,9 +26,20 @@ export class InputMessage {
...
+   * When `true`, the component is active.
+   *
+   * @deprecated use global `hidden` attribute instead.
+   */
   @Prop({ reflect: true }) active = false;
 
+  /**
+   * When `true`, the component will not be visible.
+   *
+   * @mdn [hidden](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden)
```
driskull, line 39: @Elijbet we shouldn't need to document or add this property since its already a global property on every dom element.

src/components/loader/loader.tsx
```diff
@@ -20,9 +20,21 @@ export class Loader {
...
+   * When `true`, the component is active.
+   *
+   * @deprecated use global `hidden` attribute instead.
+   */
   @Prop({ reflect: true }) active = false;
 
+  /**
+   * When `true`, the component will not be visible.
+   *
+   * @mdn [hidden](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden)
```
driskull, line 34: we shouldn't need to document or add this property since its already a global property on every dom element.

driskull (changes requested):
Looks good after removal of hidden property and doc

src/components/input-message/input-message.tsx
```diff
@@ -26,7 +26,11 @@ export class InputMessage {
   //
   //--------------------------------------------------------------------------
 
-  /** When `true`, the component is active. */
+  /**
+   * When `true`, the component is active.
+   *
+   * @deprecated use global `hidden` attribute instead.
```
driskull, line 32: nitpick: Should we add the link to the `mdn` hidden attribute here? Not necessary but might be nice.

driskull (approved):
\( ﾟヮﾟ)/

Looks good!~ ✨

src/components/input-message/input-message.tsx
Elijbet: I think that'd be great. I'll add it now.

Issue #5429, opened 2022-10-10
Popover and Tooltip do not auto-position itself correctly when placed inside other components on non-chromium browsers


sonofflynn89:
#### Popover
I was able to reproduce this bug with a `calcite-popover`  /  `calcite-card` combo, though I suspect that other component combinations might also exhibit this behavior. Essentially, you have to position the popover trigger in such a way that the popover would be hidden if it flowed to the right (default behavior).

In a Chrome window, the popover correctly positions itself to the left of the trigger to avoid overflow:

![image](https://user-images.githubusercontent.com/30124777/194966621-249379f9-2728-4406-b289-dc0a96439569.png)

On either Firefox or Safari windows, the popover positions itself to the right of the trigger, is hidden due to overflow, and is not completely vertically centered:

<img width="419" alt="image" src="https://user-images.githubusercontent.com/30124777/194966663-6d2e2733-7e4e-421b-9866-8ced24995f2a.png">

#### Tooltip

**Firefox/Safari**:
![image](https://user-images.githubusercontent.com/5023024/196545264-9a695b31-1841-4cab-8903-9b5e6eb74d93.png)

**Chromium**:
Tooltips and popovers with `overlay-positioning="fixed"` are not initially positioned properly. On first target hover the tooltip or popover becomes visible and then moves to correct position. All subsequent openings are fine.

### Expected Behavior

On Safari / Firefox, the popover correctly positions itself to the left of the trigger to avoid overflow, just like on Chrome. The popover is also vertically centered.

#### Popover
https://codepen.io/sonofflynn89/pen/YzLROEP

#### Tooltip
- Firefox/Safari browsers: https://codesandbox.io/s/gifted-tess-bujev9
- Chromium browsers: https://codepen.io/COV-GIS/pen/vYjwZLN

#### Popover
- Open the [codepen](https://codepen.io/sonofflynn89/pen/YzLROEP) in Chrome
- Click the button
- Notice that the popover auto-positions itself to the left, avoiding overflow

- Open the [codepen](https://codepen.io/sonofflynn89/pen/YzLROEP) in Safari / Firefox
- Click the button
- Notice that the popover auto-positions itself to the right, is hidden by overflow, and is not vertically centered

#### Tooltip
- Firefox/Safari browsers: Hover over the items in the shell and see that "absolute" positioning has an offset the height of the header in the calcite shell
- Chromium browsers: Hover over target, optionally uncomment js for workaround

### Reproduction Version

[...]

PR #5542, opened 2022-10-21
fix: add custom logic for floating-ui positioning across shadow DOM in non-Chromium browsers

jcfranco:
**Related Issue:** #5429

## Summary

This restores positioning logic that was reverted from [`@floating-ui/dom@1.0.2`](https://github.com/floating-ui/floating-ui/releases/tag/%40floating-ui%2Fdom%401.0.2) due to the performance impact. The fix will only be applied to non-Chromium browsers (targeting Firefox and Safari specifically) by default and can be disabled by providing the following config flag:

```html
<script>
  var calciteComponentsConfig = {
    floatingUINonChromiumPositioningFix: false
  }
</script>
```

### Notes

* introduces a global config for components to allow users to opt-out of the fix in favor of performance
* bumps `@floating-ui/dom` to [1.0.3](https://github.com/floating-ui/floating-ui/releases/tag/%40floating-ui%2Fdom%401.0.3)
* dynamically loads custom code for non-Chromium browsers for proper positioning
* the nature of this fix requires using some perf-degrading code for non-Chromium browsers, but with #5399/#5484  (less frequent positioning updates), this should be less severe though
* the config should not be documented until we solidify a few items first (naming, scope, feedback on defaults), but should be fine to share as a workaround
* the module for the fix depends on copying over some `floating-ui` code (doc has move info) and we'll have to keep this in sync until we submit a PR to only require passing  our custom `getOffsetParent` implementation.

### DIFF
--- a/src/utils/floating-ui.ts
+++ b/src/utils/floating-ui.ts
@@ -15,6 +15,7 @@ import {
 } from "@floating-ui/dom";
 import { getElementDir } from "./dom";
 import { Build } from "@stencil/core";
+import { config } from "./config";
 
 const floatingUIBrowserCheck = patchFloatingUiForNonChromiumBrowsers();
 
@@ -37,6 +38,7 @@ async function patchFloatingUiForNonChromiumBrowsers(): Promise<void> {
 
   if (
     Build.isBrowser &&
+    config.floatingUINonChromiumPositioningFix &&
     // ⚠️ browser-sniffing is not a best practice and should be avoided ⚠️
     /firefox|safari/i.test(getUAString())
   ) {
--- a/src/utils/floating-ui.ts
+++ b/src/utils/floating-ui.ts
@@ -36,7 +36,7 @@ async function patchFloatingUiForNonChromiumBrowsers(): Promise<void> {
   }
 
   if (
-    !Build.isBrowser ||
+    Build.isBrowser &&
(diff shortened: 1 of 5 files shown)


src/utils/floating-ui.ts
```diff
@@ -1,18 +1,56 @@
...
+      return uaData.brands.map((item) => `${item.brand}/${item.version}`).join(" ");
+    }
+
+    return navigator.userAgent;
+  }
+
+  if (
+    Build.isBrowser &&
+    config.floatingUINonChromiumPositioningFix &&
+    // ⚠️ browser-sniffing is not a best practice and should be avoided ⚠️
```
driskull, line 42: 👍

driskull (approved):
Since the default `floatingUINonChromiumPositioningFix` is true, I wonder if we should change the config value to `disableFloatingUINonChromiumPositioningFix`?

Do you think there would ever be a need for a single component to have this fix and not the rest? If so then the config idea might not work.

> the config should not be documented until we solidify a few items first (naming, scope, feedback on defaults), but should be fine to share as a workaround

Agreed.

> the module for the fix depends on copying over some floating-ui code (doc has move info) and we'll have to keep this in sync until we submit a PR to only require passing our custom getOffsetParent implementation.

👍

Issue #5541, opened 2022-10-21
Drag handle icon is not accessible in Value List / List


ashetland:
### Summary

The color of the drag handle (ui-border-1) on ui-foreground-1 has a contrast ratio of 1.63:1. 

Changing the color to ui-border-input would increase contrast to 3:1 while still allowing the handle to be less prominent.

### Actual Behavior

<img width="914" alt="CleanShot 2022-10-21 at 13 43 39@2x" src="https://user-images.githubusercontent.com/108549080/197285405-15acd221-48b1-4162-b77e-f111a95002fc.png">

### Expected Behavior

<img width="916" alt="CleanShot 2022-10-21 at 13 45 12@2x" src="https://user-images.githubusercontent.com/108549080/197285650-5b792962-cadb-404b-b43b-fd9dfec76446.png">

### Reproduction Sample

https://codepen.io/ashetland/pen/YzLmLyq

### Reproduction Steps

n/a

### Reproduction Version

beta.95

### Esri team

Calcite (design)

geospatialem:
Verified on `next.608` with light and dark themes. 

cc: @ashetland + @SkyeSeitz for Figma changes. Once the Figma file is updated, the issue can be closed out.

SkyeSeitz:
Thanks, @geospatialem. This has now been updated in Figma 🚀

PR #5543, opened 2022-10-21
fix(value-list-item): Change drag handle color.

driskull:
**Related Issue:** #5541

## Summary

fix(value-list-item): Change drag handle color.


benelan (approved):
👍

benelan:
Kind of unfortunate screener didn't catch this change

Issue #4628, opened 2022-05-26
Add scale to popover.


Elijbet:
### Actual Behavior

Follow up on PR #4627.

Add scale on `popover`, so it's consistent with other components that have scale. Also to be able to accommodate nested items that also have scale such as an `action` slot with an `icon`. 

### Expected Behavior

`popover` should have a scale prop.

### Reproduction Sample

N/A

### Reproduction Steps

N/A

### Reproduction Version

1.0.0-next.478

driskull:
Currently, only scale relationships have existed for components that have a child component. For example, components like `dropdown` & `dropdown-item`.

If we want to start having scale relationships with components that are not related in a parent/child sense, then we should probably move scale to a CSS custom property. I don't think we should have unrelated components rely on each other for scale.

I think this might direct us back to this issue: https://github.com/Esri/calcite-components/issues/2868

Elijbet:
> Currently, only scale relationships have existed for components that have a child component. For example, components like `dropdown` & `dropdown-item`.
> 
> If we want to start having scale relationships with components that are not related in a parent/child sense, then we should probably move scale to a CSS custom property. I don't think we should have unrelated components rely on each other for scale.
> 
> I think this might direct us back to this issue: #2868

Thanks for pointing to the cascading scale discussion, that's very helpful. I was thinking that if all visual components have a scale (this one is missing it by omission rather than by design), whatever is nested within would take the default or user-defined scale of the parent. I’ll read up more on the existing trail of related issues to catch up on the background.

jcfranco:
@Elijbet please correct me if I'm wrong, but this issue stemmed from the popover not having a `scale` prop to adjust its dismissible button icon accordingly (like we do for other icon-owning components).

> I think this might direct us back to this issue: https://github.com/Esri/calcite-components/issues/2868

Agreed. We'll pick up that discussion soon.

jcfranco:
> > I think this might direct us back to this issue: https://github.com/Esri/calcite-components/issues/2868
>
> Agreed. We'll pick up that discussion soon.

We've decided to not support inheriting scale, but we still have the question whether popover needs a `scale` prop to enable users to set the scale of the dismissible button icon as stated above. Adding @macandcheese @ashetland @SkyeSeitz for their UXpertise™.

macandcheese:
I'd expect the `heading` string to also be affected by a scale change, in addition to the close button.

driskull:
Agreed ^

[...]

PR #5560, opened 2022-10-25
fix(popover): adds a new `scale` prop

Elijbet:
**Related Issue:** #4628

## Summary

Adds scale to `popover` 

- so it's consistent with all other visual components that have a `scale` prop

- to coordinate with the scales of the `heading`, the optional dismissible nested `action` element, and the slotted x `icon`

- to ensure visual height consistency when the icon is collapsed


driskull (approved):
LGTM!

＼(＾O＾)／

src/components/popover/popover.scss
```diff
@@ -1,6 +1,31 @@
 @include floatingUIHost();
 @include floatingUIArrow();
 
+// scale variants for child
```
jcfranco, line 4: Let's drop this comment. The selectors are expressive enough.

src/components/popover/popover.stories.ts
```diff
@@ -101,3 +101,29 @@ export const flipPlacements_TestOnly = (): string => html`
     document.querySelector(".my-popover").flipPlacements = ["right"];
   </script>
 `;
+
+export const scaleConsistencyPopoverHeadingActionSlottedIcon = stepStory(
```
jcfranco, line 105: Per https://github.com/Esri/calcite-components/blob/master/conventions/Documentation.md#screenshot-test-only-stories, please add the `_TestOnly` suffix.

src/components/popover/popover.stories.ts
```diff
@@ -101,3 +101,29 @@ export const flipPlacements_TestOnly = (): string => html`
...
+  (): string => html`
+    <div style="width: 400px;">
+      ${referenceElementHTML}
+      <calcite-popover
+        heading="Dreams didn't make us kings. Dragons did. 🐉"
+        reference-element="reference-element"
+        placement="${select("placement", placements, defaultPopoverPlacement)}"
```
jcfranco, line 112: Please remove the knobs as these are not needed for test-specific stories.

src/components/popover/popover.stories.ts
```diff
@@ -101,3 +101,29 @@ export const flipPlacements_TestOnly = (): string => html`
...
+        ${boolean("open", true)}
+        ${boolean("dismissible", true)}
+        scale="${select("scale", ["s", "m", "l"], "m")}"
+      >
+        ${contentHTML}
+      </calcite-popover>
+    </div>
+  `,
+  createSteps("calcite-popover")
```
jcfranco, line 121: Let's drop the stepped stories to keep our screenshot quota down.

jcfranco (approved):

src/components/popover/popover.stories.ts
Elijbet: Ok, I added them for designers to be able to give me feedback on visuals. I'll prep a demo instead. I think having an interactive demo would be better for a design review anyway.

Issue #2709, opened 2021-08-03
Enhancement: Create calcite-input-time component


AdelheidF:
### Description

Create a new `calcite-input-time` component to be used by `calcite-input-time-picker` to solve issues https://github.com/Esri/calcite-components/issues/2331 and https://github.com/Esri/calcite-components/issues/2329

-----------
From @eriklharper

I discussed this with @jcfranco and we came up with the idea to create a new `calcite-input-time` component that will replace the current input box inside `calcite-input-time-picker` which will be modeled after the native browser implementations like this one from Chrome:

![image](https://user-images.githubusercontent.com/821864/127046380-2442a56d-fd06-44a0-9858-6f1ffac4dd81.png)

The behavior will essentially change from being a free-form text field, to established input fields that act as a mask for entering time values.

jcfranco:
Linking to discussion: https://github.com/Esri/calcite-components/discussions/4360

BlakeStearman:
The home app team is also interested in this enhancement. We're currently using `input-time-picker` in our UI for scheduling tasks where a start or end time is needed when rebuilding hosted scene layers on a schedule, and had also run into the behaviors that https://github.com/Esri/calcite-components/issues/2331 and https://github.com/Esri/calcite-components/issues/2329 cover. Notebook task scheduling is still using a very old Dojo time input, so we are looking to move to `input-time-picker` presently as well.

eriklharper:
> The home app team is also interested in this enhancement. We're currently using `input-time-picker` in our UI for scheduling tasks where a start or end time is needed when rebuilding hosted scene layers on a schedule, and had also run into the behaviors that #2331 and #2329 cover. Notebook task scheduling is still using a very old Dojo time input, so we are looking to move to `input-time-picker` presently as well.

Great to hear, I am currently working on this component!

geospatialem:
This new component will continue work in February and is anticipated to land in the February 2023 release.

geospatialem:
Reallocating to the April milestone to tackle.

geospatialem:
Localization research has been performed. Implementation has been moved to the upcoming May release.

geospatialem:
Will be adding millisecond support to `input-time-picker`, this new component will leverage when it lands. As a result this issue is blocked by millisecond effort in #6591. 

Targeting the effort for later this year in December.

eriklharper:
Related issue that can be addressed by a masked input: https://github.com/Esri/calcite-design-system/issues/2709

[...]

PR #5639, opened 2022-10-25
feat(input-time): calcite-input-time component

eriklharper:
**Related Issue:** #2709 

## Summary

Introduces the new `calcite-input-time` component for simplified time entry.


src/components/input-time/input-time.tsx
```diff
@@ -0,0 +1,872 @@
...
+    }
+    if (composedEventPath.includes(this.minuteEl)) {
+      return;
+    }
+    if (composedEventPath.includes(this.secondEl)) {
+      return;
+    }
+    if (composedEventPath.includes(this.meridiemEl)) {
+      return;
+    }
```
benelan, line 281: Can you use one if statement for all of these, it seems overly verbose. Maybe just a bunch of `||`'s, or throw the props in any array and use `some()`.

```js
if ([this.hourEl, this.minuteEl, this.secondEl, this.meridiemEl].some((el) =>
    composedEventPath.includes(el))) return;
```

`||` is better performance wise though

src/components/input-time/input-time.tsx
```diff
@@ -0,0 +1,872 @@
...
+  //--------------------------------------------------------------------------
+
+  /**
+   * Sets focus on the component.
+   *
+   * @param target
+   */
+  @Method()
+  async setFocus(target?: TimePart): Promise<void> {
+    this[`${target || "hour"}El`]?.focus();
```
benelan, line 357: I forget the name, but you need to await a util from `loadable.ts` before calling `focus()`. See any other components for an example.

src/components/input-time/input-time.tsx
```diff
@@ -0,0 +1,872 @@
...
+      }
+    }
+    this.localizedMeridiem = this.value
+      ? localizeTimeStringToParts({ value: this.value, locale })?.localizedMeridiem || null
+      : localizeTimePart({ value: this.meridiem, part: "meridiem", locale });
+  };
+
+  private timePartFocusHandler = (event: FocusEvent): void => {
+    this.activeEl = event.currentTarget as HTMLSpanElement;
+  };
```
benelan, line 723: Oh man, there's lot of if/else and case/switch statements in all the methods and handlers above lol 😵
It would be nice to refactor this at some point to make it more readable, maybe create some functional utils or something.

src/components/input-time/input-time.tsx
```diff
@@ -0,0 +1,872 @@
...
+          )}
+          {this.localizedSecondSuffix && (
+            <span class={CSS.delimiter}>{this.localizedSecondSuffix}</span>
+          )}
+          {showMeridiem && (
+            <span
+              aria-label={this.intlMeridiem}
+              aria-valuemax="2"
+              aria-valuemin="1"
+              aria-valuenow={(this.meridiem === "PM" && "2") || "1"}
```
benelan, line 850: We probably don't want to hardcode "PM" right, since we do translations?

benelan (changes requested):
Here are some comments from a quick first pass. I'll do some demo testing and a more in depth review Friday morning after you have time to work through the feedback.

src/components/input-time/input-time.tsx
```diff
@@ -0,0 +1,872 @@
...
+            [CSS.readonly]: this.readonly,
+            [CSS[`scale-${this.scale}`]]: true,
+            [CSS.showMeridiem]: showMeridiem,
+            [CSS.showSecond]: this.showSecond
+          }}
+          dir="ltr"
+        >
+          <calcite-icon class={CSS.clockIcon} flipRtl={this.iconFlipRtl} icon="clock" scale="s" />
+          <span
+            aria-label={this.intlHour}
```
anveshmekala, line 782: This needs to use `t9nComponent` interface . `intl-*` pattern is deprecated.

anveshmekala: applies to other instances as well.

src/components/input-time/input-time.tsx
```diff
@@ -0,0 +1,872 @@
...
+   * @default "AM/PM"
+   */
+  @Prop() intlMeridiem = TEXT.meridiem;
+
+  /**
+   * aria-label for the minute input
+   *
+   * @default "Minute"
+   */
+  @Prop() intlMinute = TEXT.minute;
```
anveshmekala, line 104: any label's that needs to be translated should use `messages` & `messageOverrides` properties. Please ref component with `t9nComponent` interface.

package-lock.json
```diff
@@ -21,7 +21,7 @@
       "devDependencies": {
```
alisonailea, line 21: I think this file should be reverted and the lock file changes done in a separate PR or a publish.

src/components/input-time/input-time.e2e.ts
```diff
@@ -0,0 +1,360 @@
...
+  focusable,
+  formAssociated,
+  labelable,
+  reflects,
+  renders
+} from "../../tests/commonTests";
+import { CSS } from "./resources";
+
+describe("calcite-input-time", () => {
+  it("renders", async () => renders("calcite-input-time", { display: "inline-block" }));
```
alisonailea, line 16: suggestion: group all tests run by commonTests resources under a `describe("common tests", () => {...`

src/components/input-time/input-time.e2e.ts
```diff
@@ -0,0 +1,360 @@
...
+  it("is labelable", async () => labelable("calcite-input-time"));
+
+  it("should focus the hour input when setFocus is called", async () =>
+    focusable(`calcite-input-time`, {
+      shadowFocusTargetSelector: "span.hour"
+    }));
+
+  it("can be disabled", () => disabled("calcite-input-time"));
+
+  it("when set to readOnly, element still focusable but won't display the controls or allow for changing the value", async () => {
```
alisonailea, line 49: can you rephrase this using the `it("should..."` convention?

src/components/input-time/input-time.e2e.ts
```diff
@@ -0,0 +1,360 @@
...
+    const page = await newE2EPage();
+    await page.setContent(`<calcite-input-time id="canReadOnly" readonly></calcite-input-time>`);
+
+    const component = await page.find("calcite-input-time");
+
+    await component.callMethod("setFocus");
+    await page.waitForChanges();
+
+    expect(await page.evaluate(() => document.activeElement.id)).toBe("canReadOnly");
```
alisonailea, line 58: this `expect` is unnecessary and it's validating anything from the component so it should be removed

src/components/input-time/input-time.e2e.ts
```diff
@@ -0,0 +1,360 @@
...
+    accessible(`
+    <calcite-label>
+      Input Time
+      <calcite-input-time name="test"></calcite-input-time>
+    </calcite-label>
+  `));
+
+  it("has defaults", async () =>
+    defaults("calcite-input-time", [
+      { propertyName: "scale", defaultValue: "m" },
```
alisonailea, line 28: does it make sense to standardize around storing defaults in `./resources.ts` so they can be referenced in the component and tests?

src/components/input-time/input-time.e2e.ts
alisonailea, line 58: can you move this expect to the bottom of the test. I'd like to group the expects as much as possible for easier readability.

src/components/input-time/input-time.e2e.ts
```diff
@@ -0,0 +1,360 @@
...
+    await page.waitForChanges();
+
+    await component.type("12");
+    await page.keyboard.press("Tab");
+    await component.type("30");
+    await page.waitForChanges();
+
+    expect(await component.getProperty("value")).toBeNull();
+  });
+
```
alisonailea, line 70: group these lang specific tests under a describe. Something like... `describe("lang specific UI value changes", () => { ...`

src/components/input-time/input-time.e2e.ts
```diff
@@ -0,0 +1,360 @@
...
+    expect(inputEvent).toHaveReceivedEventTimes(2);
+    expect(changeEvent).toHaveReceivedEventTimes(1);
+
+    await page.keyboard.down("Shift");
+    await page.keyboard.press("Tab");
+    await page.keyboard.up("Shift");
+    await page.keyboard.press("Tab");
+    await page.waitForChanges();
+
```
alisonailea, line 288: you can remove all the `expect`s before this one to improve the speed of the tests without losing test integrity. If you want to better describe what is happening, add some inline comments.

src/components/input-time/input-time.e2e.ts
```diff
@@ -0,0 +1,360 @@
...
+    await page.keyboard.press("Enter");
+    await page.waitForChanges();
+
+    expect(inputEvent).toHaveReceivedEventTimes(1);
+    expect(changeEvent).toHaveReceivedEventTimes(1);
+
+    await page.keyboard.press("Enter");
+    await page.waitForChanges();
+
+    expect(inputEvent).toHaveReceivedEventTimes(1);
```
alisonailea, line 323: same as test above. Let's improve test performance by removing unnecessary `expect`. Add some inline comments to describe when the input event is called and why you can press the enter key multiple times and the change event only calls once.

src/components/input-time/input-time.e2e.ts
```diff
@@ -0,0 +1,360 @@
...
+  it("sets initial value to null when it is not a valid time value", async () => {
+    const page = await newE2EPage({
+      html: `<calcite-input-time value="invalid"></calcite-input-time>`
+    });
+    const inputTime = await page.find("calcite-input-time");
+
+    expect(await inputTime.getProperty("value")).toBeNull();
+  });
+
+  it("is form-associated", () => formAssociated("calcite-input-time", { testValue: "03:23", submitsOnEnter: true }));
```
alisonailea, line 359: move this test up with the other common tests

src/components/input-time/input-time.e2e.ts
```diff
@@ -0,0 +1,360 @@
...
+    const inputTime = await page.find("calcite-input-time");
+
+    await page.evaluate(() => {
+      const inputTime = document.querySelector("calcite-input-time");
+      inputTime.addEventListener("calciteInputTimeInput", (event) => {
+        event.preventDefault();
+      });
+    });
+
+    expect(await inputTime.getProperty("value")).toBe("14:59");
```
alisonailea, line 340: remove this unnecessary `expect`

src/utils/time.ts
```diff
@@ -66,12 +66,42 @@ export function formatTimeString(value: string): string {
...
+  const minuteIndex = parts.indexOf(parts.find(({ type }): boolean => type === "minute"));
+  const secondIndex = parts.indexOf(parts.find(({ type }): boolean => type === "second"));
+  const minuteSuffix = parts[minuteIndex + 1];
+  return minuteSuffix && minuteSuffix.type === "literal" && secondIndex - minuteIndex === 2
+    ? minuteSuffix.value?.trim() || null
+    : null;
+}
+
+export function getLocaleSecondSuffix(locale: string): string {
+  const formatter = createLocaleDateTimeFormatter(locale);
```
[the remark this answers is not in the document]
alisonailea: this should probably be done once in connectedCallback or componentWillLoad. if the user is changing the locale, the whole page is going to be reloading anyway.

src/components/input-time/input-time.tsx
```diff
@@ -0,0 +1,872 @@
...
+
+  // --------------------------------------------------------------------------
+  //
+  //  Private Properties
+  //
+  // --------------------------------------------------------------------------
+
+  private activeEl: HTMLSpanElement;
+
+  defaultValue: InputTime["value"];
```
alisonailea, line 174: if these are private, please add `private` in front of them

src/components/input-time/input-time.tsx
```diff
@@ -0,0 +1,872 @@
...
+  private secondEl: HTMLSpanElement;
+
+  private userChangedValue = false;
+
+  // --------------------------------------------------------------------------
+  //
+  //  State
+  //
+  // --------------------------------------------------------------------------
+
```
alisonailea, line 201: Is there a way to simplify these State props? It seems like we only need the localized state or regular state, not both as updating one could drive the other.

src/components/input-time/input-time.tsx
```diff
@@ -0,0 +1,872 @@
...
+   */
+  @Event() calciteInternalInputTimeFocus: EventEmitter<void>;
+
+  //--------------------------------------------------------------------------
+  //
+  //  Event Listeners
+  //
+  //--------------------------------------------------------------------------
+
+  @Listen("blur")
```
alisonailea, line 260: we need to create some documentation alignment on how we are handling event listeners. I prefer this syntax but other devs are saying these should be setup as on event listeners in the component render itself. @benelan thoughts?

src/components/input-time/input-time.tsx
alisonailea: the performance difference is very slight however so in this case I would defer to which seems more readable.

src/components/input-time/input-time.tsx
```diff
@@ -0,0 +1,872 @@
...
+          break;
+        case " ":
+        case "Spacebar":
+          event.preventDefault();
+          break;
+      }
+    }
+  };
+
+  onLabelClick(): void {
```
alisonailea, line 566: private?

src/components/input-time/input-time.tsx
alisonailea: agreed. I'd like to start pulling as many private functions out into their own single testable files in a ./[component]/functions/[fileName]` pattern. It will let us better test our components and maybe recognize more functional patterns we already have.

alisonailea (changes requested):
Left a few comments. Major priorities are 
1. organizing e2e tests
2. think about how much of the input-time component can be written as functional utils/component(s).

src/utils/time.ts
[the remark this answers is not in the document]
benelan: The whole page won't necessarily be reloaded if the user changes the `lang` property on the specific component.

src/components/input-time/input-time.tsx
benelan: Yeah I don't feel strongly in either direction but I agree on the need for consistency. I'll run it by Franco when he returns.

src/components/input-time/resources.ts
```diff
@@ -0,0 +1,25 @@
...
+  second: "second",
+  showMeridiem: "show-meridiem",
+  showSecond: "show-second"
+};
+
+export const TEXT = {
+  hour: "Hour",
+  meridiem: "AM/PM",
+  minute: "Minute",
+  second: "Second"
```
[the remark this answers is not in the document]
anveshmekala: yup we have these translated for `calcite-time-picker`. currently, we don't have shared assets set up yet, so we have to copy the message bundles from time-picker component and generate the type using `util:generate-t9n-types` command. Please make sure that the component path is added in `manifest` file.

src/components/input-time/input-time.e2e.ts
benelan: Yeah I like that idea

eriklharper:
Closing in favor of https://github.com/Esri/calcite-design-system/issues/8624

PR #5640, opened 2022-10-25
fix(stepper-item): make sure numberingSystem is rendered on load

benelan:
**Related Issue:** #5467

## Summary

Providing `numberingSystem` to the formatter isn't working correctly in the `effectiveLocale` watcher. ~~I'm changing it back to the render block so it works in this upcoming release. I'll get it working in a hook or lifecycle method after the release, when we aren't so crunched.~~

Looks like I just needed to add it to `componentWillLoad` as well, for the initial rendering on load. Thanks Franco!

### DIFF
--- a/src/components/stepper-item/stepper-item.tsx
+++ b/src/components/stepper-item/stepper-item.tsx
@@ -136,6 +136,15 @@ export class StepperItem implements InteractiveComponent, LocalizedComponent {
 
   @State() effectiveLocale = "";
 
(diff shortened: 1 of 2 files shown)


jcfranco (approved):
Code LGTM! 💥

Is this already covered in an existing test? If not, let's create a follow-up issue to see if we can provide a test helper for all components that support `numberingSystem`.

Issue #5538, opened 2022-10-21
calcite-tooltip with close-on-click will show up if clicked quickly


gpbmike:
### Actual Behavior

Using `<calcite-tooltip close-on-click>`, the tooltip will open if the reference element is clicked before the tooltip opens.

It's clear when calcite-tooltip and calcite-popover share the same reference element. The tooltip will overlap the popover.

### Expected Behavior

When using `close-on-click` the tooltip should not show up when the reference-element is clicked.

### Reproduction Sample

https://codepen.io/gpbmike/pen/mdLNBMz

### Reproduction Steps

1. Click on reference element before tooltip opens.

### Reproduction Version

beta.95

### Relevant Info

![image](https://user-images.githubusercontent.com/8754/197244352-ea327902-0f9a-4e10-99a0-8f4ca7730d5a.png)

### Esri team

ArcGIS Site Scan

gpbmike:
@geospatialem what are you using to record your clicks in your gif? i've been looking for something like that

geospatialem:
> what are you using to record your clicks in your gif? i've been looking for something like that

@gpbmike [ScreenToGif](https://www.screentogif.com) - it's been a handy tool! 👍🏻

PR #5643, opened 2022-10-25
fix(tooltip): Prevent opening when closeOnClick is true and referenceElement is clicked quickly

driskull:
**Related Issue:** #5538

## Summary

fix(tooltip): Prevent opening when closeOnClick is true and referenceElement is clicked quickly. (#5538)

### DIFF
--- a/src/components/tooltip/tooltip.e2e.ts
+++ b/src/components/tooltip/tooltip.e2e.ts
@@ -561,6 +561,35 @@ describe("calcite-tooltip", () => {
     expect(await tooltip.getProperty("open")).toBe(false);
   });
 
+  it("should close tooltip when closeOnClick is true and referenceElement is clicked quickly", async () => {
+    const page = await newE2EPage();
+
+    await page.setContent(
+      html`
+        <calcite-tooltip reference-element="ref" close-on-click>Content</calcite-tooltip>
+        <button id="ref">Button</button>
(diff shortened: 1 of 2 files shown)


src/components/tooltip/tooltip.e2e.ts
```diff
@@ -561,6 +561,35 @@ describe("calcite-tooltip", () => {
...
+    const page = await newE2EPage();
+
+    await page.setContent(
+      html`
+        <calcite-tooltip reference-element="ref" close-on-click>Content</calcite-tooltip>
+        <button id="ref">Button</button>
+      `
+    );
+
+    await page.waitForChanges();
```
jcfranco, line 574: Nitpick: Can you confirm if this is needed here? I was under the impression that `setContent` called this internally, but there's [no mention of this in the doc](https://stenciljs.com/docs/end-to-end-testing#:~:text=setContent(html%3A%20string)). Worth noting that this behavior is documented in the [spec page doc](https://stenciljs.com/docs/unit-testing#:~:text=return%20null.-,setContent(html),-Convenience%20function%20to).

jcfranco (approved):
LGTM! 🎉

Can you update the PR title to reflect the fix from a user's perspective? Not sure if it'll be clear from reading the changelog entry as is.

Issue #863, opened 2020-08-13
New Component: calcite-text-area


asangma:
### Description

* Separate out textarea from CalciteInput.
* Retain a slot for action(s)
* Layout actions at the bottom in a sort of footer
* Add support for resize #1528

### Design and spec

Aaron's Updated Figma Design: https://www.figma.com/file/OQF8Y5t1VFk8uaZ0Hyjot9/Textarea-%5Benhancement%5D-%5Bnew-component%5D-%5Bissue-%23863%5D?node-id=635%3A6742

#### Related:
- #2866
- #3312

macandcheese:
I think this is a good time to split `textarea` into its own component vs `<calcite-input type=textarea>` which has always been a little weird. We will want to leave the existing `input-action` slot for inputs as is.

I could see textarea component having a `footer-actions` slot that is placed and spaced inside the textarea, and we could use for save / cancel functionality internally and folks could use to place items like your example, if needed.

bstifle:
Agreed! would be cool for Bold Italic/etc actions too

asangma:
Cool.
Though I suggest just using the same slot name and putting the actions at the bottom.
I think if a full-on rich text editor happens, it should be its own component.

As a note, I think the `input-slot` name could be simplified to just `action` or `actions`.
It's inside `input` already so the context should handle it.

macandcheese:
As long as we document or limit it to button components, at least for input... Action sounds like it's for the action component which is not really what we want folks to use adjacent to inputs. Even in textarea I'd suggest buttons be used.

asangma:
Hmmm...isn't the idea behind the slot to allow the user to decide what goes in there?

But if you don't wanna rename the slot, I think that's OK. The `input-` part just seemed unneeded to me.
If that's the convention, we'd name the slot in TextArea `textarea-action`, yeah?

macandcheese:
For the most part, yeah. But there are certainly places where we want to prevent bad UX / incentivize or recommend good practices. For the `input-action` in particular, it's placed in a trailing position as a helper for constraining button heights in an [input][button] lockup. Users can always flex-position a div of their own choosing.

`textarea-action` seems reasonable, if it's a full-width div perhaps a `footer-leading`/ `footer-trailing` paradigm for placing on either side like in `calcite-card` makes sense. Keep in mind we may want to place some items in there from the component itself, like save / cancel or a character count we don't want folks to re-create on their own if we offer it.

macandcheese:
[...]

PR #5644, opened 2022-10-25
feat(text-area): add calcite-text-area component

anveshmekala:
**Related Issue:** #863 

## Summary

This PR will add new component `calcite-text-area` which will replace `<calcite-input type="textarea" />`

#### Demo : http://calcite-dev.esri.com:3009/demos/text-area.html


## calcite-text-area


<!-- Auto Generated Below -->


## Usage

### Basic

```html
<calcite-text-area placeholder="Add notes"></calcite-text-area>
```

## Properties

| Property           | Attribute          | Description                                                                                                       | Type                                                                                                                                                                                                                                    | Default     |
| ------------------ | ------------------ | ----------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------- |
| `autofocus`        | `autofocus`        | When `true`, the component is focused on page load.                                                               | `boolean`                                                                                                                                                                                                                               | `false`     |
| `columns`          | `columns`          | Specifies the number or columns allowed.                                                                          | `number`                                                                                                                                                                                                                                | `undefined` |
| `disabled`         | `disabled`         | When `true`, interaction is prevented and the component is displayed with lower opacity.                          | `boolean`                                                                                                                                                                                                                               | `false`     |
| `groupSeparator`   | `group-separator`  | When `true`, number values are displayed with a group separator corresponding to the language and country format. | `boolean`                                                                                                                                                                                                                               | `false`     |
| `label`            | `label`            | Accessible name for the component.                                                                                | `string`                                                                                                                                                                                                                                | `undefined` |
| `maxLength`        | `max-length`       | Specifies the maximum number of characters allowed.                                                               | `number`                                                                                                                                                                                                                                | `undefined` |
| `messageOverrides` | --                 | Use this property to override individual strings used by the component.                                           | `{ invalid?: string; overLimit?: string; }`                                                                                                                                                                                             | `undefined` |
| `name`             | `name`             | Specifies the name of the component                                                                               | `string`                                                                                                                                                                                                                                | `undefined` |
| `numberingSystem`  | `numbering-system` | Specifies the Unicode numeral system used by the component for localization.                                      | `"arab" \| "arabext" \| "bali" \| "beng" \| "deva" \| "fullwide" \| "gujr" \| "guru" \| "hanidec" \| "khmr" \| "knda" \| "laoo" \| "latn" \| "limb" \| "mlym" \| "mong" \| "mymr" \| "orya" \| "tamldec" \| "telu" \| "thai" \| "tibt"` | `undefined` |
| `placeholder`      | `placeholder`      | Specifies the placeholder text for the component.                                                                 | `string`                                                                                                                                                                                                                                | `undefined` |
| `readonly`         | `readonly`         | When `true`, the component's value can be read, but cannot be modified.                                           | `boolean`                                                                                                                                                                                                                               | `false`     |
| `required`         | `required`         | When `true`, the component must have a value in order for the form to submit.                                     | `boolean`                                                                                                                                                                                                                               | `false`     |
| `resize`           | `resize`           | When `true`, disables horizontally and vertically resizing the component.                                         | `"both" \| "horizontal" \| "none" \| "vertical"`                                                                                                                                                                                        | `"both"`    |
| `rows`             | `rows`             | Specifies the number or rows allowed.                                                                             | `number`                                                                                                                                                                                                                                | `undefined` |
| `scale`            | `scale`            | Specifies the size of the component.                                                                              | `"l" \| "m" \| "s"`                                                                                                                                                                                                                     | `"m"`       |
| `value`            | `value`            | The component's value.                                                                                            | `string`                                                                                                                                                                                                                                | `undefined` |
| `wrap`             | `wrap`             | Specifies the wrapping mechanism for the text.                                                                    | `"hard" \| "soft"`                                                                                                                                                                                                                      | `"soft"`    |

## Events

| Event                   | Description                                           | Type                |
| ----------------------- | ----------------------------------------------------- | ------------------- |
| `calciteTextAreaChange` | Fires each time a new `value` is typed and committed. | `CustomEvent<void>` |
| `calciteTextAreaInput`  | Fires each time a new `value` is typed.               | `CustomEvent<void>` |

## Methods

### `selectText() => Promise<void>`

Selects all text of the component's `value`.

#### Returns

Type: `Promise<void>`

### `setFocus() => Promise<void>`

Sets focus on the component.

#### Returns

Type: `Promise<void>`

## Slots

| Slot             | Description                          |
| ---------------- | ------------------------------------ |
|                  | A slot for adding text.              |
| `"footer-end"`   | A slot for adding a trailing footer. |
| `"footer-start"` | A slot for adding a leading footer.  |

---



#### Known limitations: 

when the `calcite-text-area` with fixed height and width is wrapped inside a flex container with height and width, resizing behavior doesn't restrict the component from increasing both height and width.


src/components/textarea/textarea.scss
```diff
@@ -0,0 +1,149 @@
+:host {
+  @apply flex flex-col;
+}
+
+:host textarea {
```
driskull, line 5: You shouldn't need the `:host` here

src/components/textarea/textarea.scss
```diff
@@ -0,0 +1,149 @@
...
+    &:focus {
+      @apply border-color-danger border-2 border-solid;
+    }
+  }
+  &.footer-slotted {
+    min-inline-size: theme("spacing.72");
+  }
+}
+
+:host .footer {
```
driskull, line 29: You shouldn't need the `:host` here

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,321 @@
...
+  NumberingSystem,
+  numberStringFormatter
+} from "../../utils/locale";
+import { createObserver } from "../../utils/observers";
+@Component({
+  tag: "calcite-textarea",
+  styleUrl: "textarea.scss",
+  shadow: true
+})
+export class Textarea implements FormComponent, LabelableComponent, LocalizedComponent {
```
driskull, line 31: Can you implement the `LoadableComponent` as well?

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,321 @@
...
+  /** Specifies name of the component  */
+  @Prop({ reflect: true }) name: string;
+
+  /** Specifies the size of `textarea` component. */
+  @Prop({ reflect: true }) scale: "l" | "m" | "s" = "m";
+
+  /** Specifies wrapping mechanism for the text.  */
+  @Prop({ reflect: true }) wrap: "soft" | "hard" = "soft";
+
+  // @Prop() autocomplete: boolean;
```
driskull, line 75: Whats this for?

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,321 @@
...
+  /** The id of the form `textarea` is associated with. */
+  @Prop({ reflect: true }) form: string;
+
+  /** The label of the component */
+  @Prop({ reflect: true }) label: string;
+
+  /**
+   * When `true`, the component will not be visible.
+   */
+  @Prop({ reflect: true }) hidden = false;
```
driskull, line 104: You don't need this as a prop as its a global attribute. I don't think you would have to pass this down to the textarea either.

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,321 @@
...
+  @Prop({ reflect: true }) horizantalResizeDisabled: boolean;
+
+  /** When `true`, disables resizing textarea vertically. */
+  @Prop({ reflect: true }) verticalResizeDisabled: boolean;
+
+  /** When `true`, marks this component as required in form. */
+  @Prop({ reflect: true }) required: boolean;
+
+  /** The id of the form `textarea` is associated with. */
+  @Prop({ reflect: true }) form: string;
```
driskull, line 96: I don't think other Form components have a form id like this. Do we need this?

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,321 @@
...
+  // @Prop() autocomplete: boolean;
+
+  /** The component's value. */
+  @Prop({ mutable: true }) value: string;
+
+  /** When `true` , footer will be added to the component. */
+  @Prop({ reflect: true }) footer = false;
+
+  /** When `true`, disables the resizing handle. */
+  @Prop({ reflect: true }) resizeDisabled: boolean;
```
driskull, line 84: @benelan @geospatialem @alisonailea @eriklharper It seems like our components have `disableFeature` and `featureDisabled` as props throughout our components.

For example, combobox has `filterDisabled` whereas dropdown has `disableCloseOnSelect`. I think we need to choose a format and update all of our components. 

Was this brought up in consistency issues before?

driskull: I think I prefer the way @anveshmekala has it as it shows as `resize-disabled` which is consistent with `aria-disabled="true"`

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,321 @@
...
+  EventEmitter,
+  VNode,
+  Watch,
+  Method,
+  Host,
+  State
+} from "@stencil/core";
+import { connectForm, disconnectForm, FormComponent, HiddenFormInputSlot } from "../../utils/form";
+import { connectLabel, disconnectLabel, getLabelText, LabelableComponent } from "../../utils/label";
+import { getSlotted } from "../../utils/dom";
```
driskull, line 16: Can you remove use of `getSlotted()` as it is deprecated now. You can see the deprecation notes on how to migrate.

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,321 @@
...
+          readonly={this.readonly}
+          ref={this.setTextareaEl}
+          required={this.required}
+          rows={this.rows}
+          value={this.value}
+          wrap={this.wrap}
+        />
+        <slot />
+        {this.footer && (
+          <footer
```
driskull, line 189: I think we might want to add a `key` on this VNode since its conditional

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,321 @@
...
+  //--------------------------------------------------------------------------
+  //
+  //  Public Methods
+  //
+  //--------------------------------------------------------------------------
+
+  /** Set's focus on the `textarea`. */
+  @Method()
+  async setFocus(): Promise<void> {
+    this.textareaEl.focus();
```
driskull, line 212: For both of these methods, await the `componentLoaded()` util function.

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,321 @@
...
+      numberingSystem: this.numberingSystem
+    };
+    return numberStringFormatter.localize(this.value?.length.toString());
+  }
+
+  disableSlottedElements(disabled: boolean): void {
+    const slottedEl = getSlotted(this.el, [SLOTS.footerLeading, SLOTS.footerTrailing], {
+      all: true
+    });
+    slottedEl.forEach((el) => (el["disabled"] = disabled));
```
driskull, line 299: Do we know that `disabled` is a valid property for all slotted elements?

driskull (commented):
Looking good! Heres my initial code review. I'll run a build and test it out as well. 👍

src/components/textarea/textarea.tsx
eriklharper: FWIW, I'd say that the pattern `<property>Disabled` makes sense if the type is `boolean` and the pattern `disable<property-or-feature>` makes sense if this is a function that performs the disabling.  Since the vast majority of these types of props are likely `boolean` my guess is we'll want the former rather than the latter.

driskull: I added it to this issue: https://github.com/Esri/calcite-components/issues/4688

src/components/textarea/textarea.scss
anveshmekala: Will get that removed.

driskull (commented):
@anveshmekala I think you'll need to add this component to the stencil config as well.

src/components/textarea/textarea.tsx
anveshmekala: Good catch .this needs to be removed. I was trying to declare global attributes earlier and forgot to clean it.

src/components/textarea/textarea.tsx
anveshmekala: To add more context, the combobox `filterDisabled` prop needed to be renamed and was added to [API consistency list](https://github.com/Esri/calcite-components/wiki/Consistency-Audit) (item 12)  and I agree with using <property>Disabled to indicate booleans.

src/components/textarea/textarea.tsx
anveshmekala: Just checked with other form components. I don't think we need this, FormComponent interface has this built-in.

src/components/textarea/textarea.tsx
anveshmekala: i dont think so, we only have one `footer` element here.

src/components/textarea/textarea.e2e.ts
```diff
@@ -0,0 +1,254 @@
...
+    await page.keyboard.press("Tab");
+    await page.waitForChanges();
+
+    await page.keyboard.type("rocky mountains");
+    await page.waitForChanges();
+
+    expect(await element.getProperty("value")).toBe("rocky mountains");
+  });
+
+  // it("should focus textarea on pageload when autofocus is true", async () => {
```
driskull, line 175: remove commented test

src/components/textarea/textarea.stories.ts
```diff
@@ -0,0 +1,46 @@
...
+      placeholder="${text("placeholder", "Add Notes")}"
+      ${boolean("footer", false)}
+      ${boolean("disabled", false)}
+      cols="${number("cols", 20)}"
+      rows="${number("rows", 2)}"
+      ${boolean("resize-disabled", true)}
+      ${boolean("horizantal-resize-disabled", false)}
+      ${boolean("vertical-resize-disabled", false)}
+      ${boolean("required", false)}
+      ${boolean("hidden", false)}
```
driskull, line 27: do we need the hidden attribute?

src/components/textarea/textarea.stories.ts
```diff
@@ -0,0 +1,46 @@
...
+      ${boolean("invalid", false)}
+      label="${text("label", "")}"
+      form="${text("form", "")}"
```
driskull, line 30: I dont think we need the form attribute

tailwind.config.ts
```diff
@@ -197,7 +197,10 @@ export default {
         2.5: "0.625rem",
         3.5: "0.875rem",
         9: "2.25rem",
-        11: "2.75rem"
+        11: "2.75rem",
+        29: "7.25rem",
+        31: "7.75rem",
+        33: "8.25rem"
```
alisonailea, line 203: how were these new spacing variables determined?

anveshmekala: These were defined in earlier iterations in sync with Figma design specifications where the width and height is 320px * 160px. Will get them removed.

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,417 @@
...
+    if (this.value?.length > this.maxlength) {
+      input.setCustomValidity(this.messages.overLimit);
+    }
+    if (this.invalid) {
+      input.setCustomValidity(this.messages.invalid);
+    }
+  }
+
+  setHeightAndWidthToAuto(): void {
+    this.timeOutId = window.setTimeout(() => {
```
driskull, line 407: Do you need to use `window.setTimeout` or will just `setTimeout` work?

Maybe rename `timeOutId` to `resizeTimeout`?

driskull: I think the timeout should be cleared right before being created.

anveshmekala: > Do you need to use `window.setTimeout` or will just `setTimeout` work?


both setTimeout and window.setTimeout works here, I see other components are using `window.setTimeout` and followed it.

anveshmekala: > Maybe rename `timeOutId` to `resizeTimeout`?

makes sense, you mean resizeTimeoutId ?

anveshmekala: > I think the timeout should be cleared right before being created.

the reason behind not adding clearTimeout before creating it to set the height and with to auto while user is dragging the resize handle. If we clear it and user play around with the drag handle not fixing on a position , then the height & width wont be set to `auto` because the resize invokations would be quicker than the 100ms or 150ms timeout causing overlap with neighbor elements. For this use case, it is safe to clear it on `disconnectedCallback` because once the timeout is cleared `setHeightAndWidthToAuto` method wont be invoked further.

driskull: Have you tried using `throttle` instead of a timeout here? It sounds like `throttle` would work with how you are describing.

anveshmekala: tried `throttle` with `leading:false` to avoid first invokation to execute and it is working as expected.

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,422 @@
...
+  @Element() el: HTMLCalciteTextareaElement;
+
+  //--------------------------------------------------------------------------
+  //
+  //  Properties
+  //
+  //--------------------------------------------------------------------------
+
+  /** When `true`, focuses the `textarea` element on page render. */
+  @Prop({ reflect: true }) autofocus: boolean;
```
driskull, line 71: Should these undefined boolean props just be set to false by default?

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,422 @@
...
+  @State() leadingSlotElements: Element[];
+
+  @Watch("effectiveLocale")
+  effectiveLocaleChange(): void {
+    updateMessages(this, this.effectiveLocale);
+  }
+
+  @State() defaultMessages: TextareaMessages;
+
+  timeOutId: number;
```
driskull, line 303: no longer needed :)

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,422 @@
...
+    setComponentLoaded(this);
+  }
+
+  disconnectedCallback(): void {
+    disconnectLabel(this);
+    disconnectForm(this);
+    disconnectLocalized(this);
+    disconnectMessages(this);
+    this.resizeObserver?.disconnect();
+    window.clearTimeout(this.timeOutId);
```
driskull, line 195: no longer needed :)

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,422 @@
...
+  footerLeadingSlotChangeHandler = (event: Event): void => {
+    this.leadingSlotElements = slotChangeGetAssignedElements(event);
+    if (this.disabled) {
+      this.disablePointerEvents(this.disabled, this.leadingSlotElements);
+    }
+  };
+
+  contentSlotChangeHandler = (): void => {
+    if (!this.value) {
+      const nodes = this.el.childNodes;
```
driskull, line 344: Can you just use `this.el.textContent.trim()`?

If not, can you use the slot event's assigned elements instead of looping through the children? I think it should be more accurate.

```
contentSlotChangeHandler = (event: Event): void => {
    this.value = slotChangeGetAssignedElements(event).map((el) => el.textContent).toString() ;
}
```

driskull (commented):
Looking good @anveshmekala!!

@benelan & @eriklharper can you review this PR once @anveshmekala is ready for another review? Then we can get it merged!

src/components/textarea/textarea.tsx
anveshmekala: yup, this needs to be `false` by default.

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,422 @@
...
+  @Prop({ reflect: true }) scale: "l" | "m" | "s" = "m";
+
+  /** Specifies wrapping mechanism for the text.  */
+  @Prop({ reflect: true }) wrap: "soft" | "hard" = "soft";
+
+  /** The component's value. */
+  @Prop({ mutable: true }) value: string;
+
+  /** When `true` , footer will be added to the component. */
+  @Prop({ reflect: true }) footer = false;
```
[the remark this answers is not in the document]
anveshmekala: i think it makes sense not to have a prop. even if the user enables footer, they have to parse either of `maxlenght` or slot elements to render the footer.

src/components/textarea/textarea.tsx
anveshmekala: `slotChangeGetAssignedElements()` wouldn't work in this scenario as the method look for event.target as `HTMLSlotElement` which in this case is a plain text rather than an Element.

anveshmekala: and the Event emitted from slot change doesn't capture textContent without an element.

src/components/textarea/textarea.tsx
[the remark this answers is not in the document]
anveshmekala: > Is there a better name for this? `footer` sounds like content to me.

`footer` is more descriptive than `content` IMO. @macandcheese , @ashetland Thoughts?

Even though this Prop will be removed , `footer` is used as prefix for slots.

eriklharper:
Based on the Figma design, shouldn't this box highlight color turn to the danger color? 
image

anveshmekala:
> Based on the Figma design, shouldn't this box highlight color turn to the danger color? image

The reason for decoupling invalid focus with character limit is to differentiate invalid state user has set via `prop`. It does look a bit odd but if we are going with `maxlength` invalidity turning textarea red, then we may need to include the footer area which might have slotted elements.
cc: @ashetland

ashetland:
The intent is to leave that under the user's control. When the limit is exceeded, additional UI may be desirable, such as an input-message or there may be a form-level event that needs accounting for. Most of the time, it will likely make sense to mark the field as `invalid` when maxlength has been exceeded, but since it's up to the user, we should document it as a best practice, e.g.:

CleanShot 2023-01-03 at 14 18 46@2x

cc: @geospatialem

ashetland:
After a lengthy conversation we've decided to auto-set `invalid` when `maxlength` is exceeded. Since `invalid` is a public property, it can be un-set if needed. This decision will help enforce/suggest a desired ui state when the limit is exceeded, but still allow users to customize the ui state if desired.

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,411 @@
...
+  @Prop() maxlength: number;
+
+  /** Specifies name of the component  */
+  @Prop({ reflect: true }) name: string;
+
+  /** Specifies the size of `textarea` component. */
+  @Prop({ reflect: true }) scale: "l" | "m" | "s" = "m";
+
+  /** Specifies wrapping mechanism for the text.  */
+  @Prop({ reflect: true }) wrap: "soft" | "hard" = "soft";
```
benelan, line 98: > Specifies the wrapping mechanism for the text

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,411 @@
...
+  /** Specifies number or rows allowed. */
+  @Prop({ reflect: true }) rows: number;
+
+  /** Specifies number or columns allowed. */
+  @Prop({ reflect: true }) cols: number;
+
+  /** Specifies maximum number of characters allowed. */
+  @Prop() maxlength: number;
+
+  /** Specifies name of the component  */
```
benelan, line 91: > Specifies name of the component.

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,411 @@
...
+  /** Whne `true`, the component's value can be read, but cannot be modified.  */
+  @Prop({ reflect: true }) readonly = false;
+
+  /** Specifies number or rows allowed. */
+  @Prop({ reflect: true }) rows: number;
+
+  /** Specifies number or columns allowed. */
+  @Prop({ reflect: true }) cols: number;
+
+  /** Specifies maximum number of characters allowed. */
```
benelan, line 88: > Specifies the maximum number of characters allowed.

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,411 @@
...
+  /** Specifies the placeholder text for the input. */
+  @Prop() placeholder: string;
+
+  /** Whne `true`, the component's value can be read, but cannot be modified.  */
+  @Prop({ reflect: true }) readonly = false;
+
+  /** Specifies number or rows allowed. */
+  @Prop({ reflect: true }) rows: number;
+
+  /** Specifies number or columns allowed. */
```
benelan, line 85: > Specifies the number of columns allowed.

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,411 @@
...
+  /** When `true`, disables the component. */
+  @Prop({ reflect: true }) disabled = false;
+
+  /** Specifies the placeholder text for the input. */
+  @Prop() placeholder: string;
+
+  /** Whne `true`, the component's value can be read, but cannot be modified.  */
+  @Prop({ reflect: true }) readonly = false;
+
+  /** Specifies number or rows allowed. */
```
benelan, line 82: > Specifies the number of rows allowed.

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,411 @@
...
+  //
+  //--------------------------------------------------------------------------
+
+  /** When `true`, focuses the `textarea` element on page render. */
+  @Prop({ reflect: true }) autofocus = false;
+
+  /** When `true`, disables the component. */
+  @Prop({ reflect: true }) disabled = false;
+
+  /** Specifies the placeholder text for the input. */
```
benelan, line 76: > Specifies the placeholder text for the component.

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,411 @@
...
+  //
+  //  Properties
+  //
+  //--------------------------------------------------------------------------
+
+  /** When `true`, focuses the `textarea` element on page render. */
+  @Prop({ reflect: true }) autofocus = false;
+
+  /** When `true`, disables the component. */
+  @Prop({ reflect: true }) disabled = false;
```
benelan, line 74: > When \`true\`, interaction is prevented and the component is displayed with lower opacity.

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,411 @@
...
+  @Element() el: HTMLCalciteTextareaElement;
+
+  //--------------------------------------------------------------------------
+  //
+  //  Properties
+  //
+  //--------------------------------------------------------------------------
+
+  /** When `true`, focuses the `textarea` element on page render. */
+  @Prop({ reflect: true }) autofocus = false;
```
benelan, line 71: Can you change the comment to match `calcite-input`:
```js
  /**
   * When `true`, the component is focused on page load.
   *
   * @mdn [autofocus](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus)
   */
```

Can you also add the `@mdn` jsdoc tag where applicable? See `calcite-input` for examples. We plan to eventually add it as a chip/link  in the documentation site

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,411 @@
...
+  @Prop({ reflect: true }) resizeDisabled = false;
+
+  /** When `true`, disables resizing textarea horizantally. */
+  @Prop({ reflect: true }) horizantalResizeDisabled = false;
+
+  /** When `true`, disables resizing textarea vertically. */
+  @Prop({ reflect: true }) verticalResizeDisabled = false;
+
+  /** When `true`, marks this component as required in form. */
+  @Prop({ reflect: true }) required = false;
```
benelan, line 113: > When \`true\`, the component must have a value in order for the form to submit.

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,411 @@
...
+  /** Specifies the size of `textarea` component. */
+  @Prop({ reflect: true }) scale: "l" | "m" | "s" = "m";
+
+  /** Specifies wrapping mechanism for the text.  */
+  @Prop({ reflect: true }) wrap: "soft" | "hard" = "soft";
+
+  /** The component's value. */
+  @Prop({ mutable: true }) value: string;
+
+  /** When `true`, disables the resizing handle. */
```
benelan, line 103: > When \`true\`, disables horizantally and vertically resizing the component.

benelan (commented):
Looking good! I added comments for some cleanup items

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,411 @@
...
+  /** When `true`, disables resizing textarea horizantally. */
+  @Prop({ reflect: true }) horizantalResizeDisabled = false;
+
+  /** When `true`, disables resizing textarea vertically. */
+  @Prop({ reflect: true }) verticalResizeDisabled = false;
+
+  /** When `true`, marks this component as required in form. */
+  @Prop({ reflect: true }) required = false;
+
+  /** The label of the component */
```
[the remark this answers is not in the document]
geospatialem: We went with:
> Accessible name for the component.

Will take a peek at the remainder of the docs shortly. 📜

geospatialem (commented):
Looking great, @anveshmekala! Some minor doc recommendations. Can't wait for this new component to be incorporated soon!

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,411 @@
...
+
+  //--------------------------------------------------------------------------
+  //
+  //  Event Emitters
+  //
+  //--------------------------------------------------------------------------
+  /**
+   * Fires each time a new `value` is typed.
+   */
+  @Event({ cancelable: true }) calciteTextareaInput: EventEmitter<void>;
```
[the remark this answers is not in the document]
anveshmekala: Initially i did considered naming `calcite-text-area` but to keep it in sync with the native element chose `calcite-textarea`. I dont have a strict preference but `calcite-text-area` is a bit odd :thinking:

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,411 @@
...
+  T9nComponent,
+  updateMessages
+} from "../../utils/t9n";
+import { TextareaMessages } from "./assets/textarea/t9n";
+import { throttle } from "lodash-es";
+
+/**
+ * @slot - A slot for adding text.
+ * @slot footer-leading - A slot for adding a leading footer.
+ * @slot footer-trailing - A slot for adding a trailing footer.
```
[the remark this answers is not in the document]
anveshmekala: I think `footer` is more descriptive than `content` and it is consistent with the existing pattern in components like `card`, `modal` & `flow` etc.

src/components/textarea/textarea.e2e.ts
```diff
@@ -0,0 +1,230 @@
...
+      const page = await newE2EPage();
+      await page.setContent(`<calcite-textarea resize-disabled></calcite-textarea>`);
+
+      const element = await page.find("calcite-textarea >>> textarea");
+      await page.waitForChanges();
+
+      expect(element).toHaveClass(CSS.resizeDisabled);
+    });
+
+    it("should have resize-disabled--x class when horizantal-resize-disabled property is parsed", async () => {
```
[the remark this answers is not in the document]
anveshmekala: unfortunately, there isn't a way to trigger resize using javascript or via keyboard for native `textarea`  element. The alternate solution is to make sure we add a CSS class which adds `resize` property.

src/components/textarea/textarea.e2e.ts
```diff
@@ -0,0 +1,230 @@
...
+
+describe("calcite-textarea", () => {
+  it("renders", async () => {
+    const page = await newE2EPage();
+    await page.setContent("<calcite-textarea></calcite-textarea>");
+
+    const element = await page.find("calcite-textarea");
+    expect(element).toHaveAttribute(HYDRATED_ATTR);
+  });
+
```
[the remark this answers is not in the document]
anveshmekala: Good catch. Added.

src/components/textarea/textarea.tsx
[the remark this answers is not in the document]
ashetland: Agree with @anveshmekala here for naming the footer slots, but should they be `footer-start` and `footer-end` for more consistency?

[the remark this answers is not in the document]
anveshmekala: i don't think we are using `footer-start`, `footer-end` across the repo in any component. I see `footer-leading`, `footer-trailing` (chip , link)  & `footer-actions` as of now.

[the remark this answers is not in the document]
ashetland: Ah okay, probably makes sense to follow that trend for the moment then. I thought we had included those when we updated naming to `-start` and `-end`, but those were apparently not included. 😂

src/components/textarea/textarea.tsx
anveshmekala: Reverting back to looping through childNodes and filter nodes with name `text` as `this.el.textContent.trim()` will fetch the text from slotted elements too.

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,411 @@
...
+        {numberStringFormatter.localize(this.maxlength.toString())}
+      </span>
+    ) : null;
+  };
+
+  getLocalizedCharacterLength(): string {
+    numberStringFormatter.numberFormatOptions = {
+      locale: this.effectiveLocale,
+      numberingSystem: this.numberingSystem
+    };
```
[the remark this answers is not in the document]
benelan: Yup my bad, nice catch

src/components/textarea/textarea.tsx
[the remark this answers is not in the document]
benelan: I'm okay with going with footer, but I think we should rename the chip/link ones to `footer-start` and `footer-end` for consistency and use those here as well

src/components/textarea/textarea.tsx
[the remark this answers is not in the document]
benelan: @macandcheese @eriklharper any thoughts here? We have deviated from the native naming pattern of smushing two words together in the past, although it was related to property not element names. Here is an example: https://github.com/Esri/calcite-components/pull/3382#issuecomment-962253064

Native can't hyphenate element names, but as custom elements we can, so in my opinion we might as well.

> [Custom elements] contain a hyphen, used for namespacing and to ensure forward compatibility __(since no elements will be added to HTML, SVG, or MathML with hyphen-containing local names in the future)__.
   -- [spec](https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name)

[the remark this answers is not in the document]
macandcheese: Native tag is <textarea>, this is a replication of that element so matching that makes sense.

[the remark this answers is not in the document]
eriklharper: I lean towards `textarea` too for the same reasons Adam mentioned.  I actually also thought as a pattern that we wanted all our "input" components to prefix the word "input" in front of them, but maybe that only applies to the native `<input type="type">` components we're breaking away from `input` and not `textarea`.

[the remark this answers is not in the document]
eriklharper: `calcite-input-textarea` is longer, but does have more semantic meaning.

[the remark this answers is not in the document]
benelan: Yeah we shouldn't put `input` in the name since the whole point is breaking it out. It was between `calcite-text-area` and `calcite-textarea`. Sounds like both Erik and Adam vote for `calcite-textarea` so we can stick with that

src/components/textarea/textarea.tsx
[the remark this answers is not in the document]
benelan: Here is the pr for changing the slot names in `calcite-card` #6249

src/demos/textarea.html
```diff
@@ -0,0 +1,279 @@
...
+      }
+
+      calcite-textarea {
+        width: 320px;
+        height: 160px;
+      }
+    </style>
+    <script src="./_assets/head.js"></script>
+  </head>
+  <body>
```
benelan, line 49: Can you add the `demo-dom-swapper` custom element like the rest of the demos have

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,439 @@
...
+  };
+
+  footerTrailingSlotChangeHandler = (event: Event): void => {
+    this.trailingSlotElements = slotChangeGetAssignedElements(event);
+    if (this.disabled) {
+      this.disablePointerEvents(this.disabled, this.trailingSlotElements);
+    }
+  };
+
+  footerLeadingSlotChangeHandler = (event: Event): void => {
```
benelan, line 353: Nitpick: should we change these handlers to Start/End instead of Leading/Trailing?

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,439 @@
...
+   *
+   * @mdn [disabled](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled)
+   */
+  @Prop({ reflect: true }) disabled = false;
+
+  /** Specifies the placeholder text for the component. */
+  @Prop() placeholder: string;
+
+  /** Whne `true`, the component's value can be read, but cannot be modified.  */
+  @Prop({ reflect: true }) readonly = false;
```
benelan, line 95: Can you continue adding the `@mdn` jsdoc tag? Any property that we pass directly to the native textarea element should have the tag. 

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly

src/components/textarea/textarea.stories.ts
```diff
@@ -0,0 +1,89 @@
...
+</div>`;
+
+export const readonly = (): string => html`<div style="width:300px;max-width:100%;">
+<calcite-textarea
+readonly
+>
+</caclite-textarea>
+</div>`;
+
+export const resizeDisabled = (): string => html`<div style="width:300px;max-width:100%;">
```
benelan, line 57: The `resizeDisabled`, `readonly`, and `disabled` stories can all be TestOnly. You can think of it like anything that a user can achieve by using knobs in the `simple` or other stories doesn't need to be repeated in the public stories.

So stuff like adding/removing properties can be TestOnly, but adding stuff to slots or other things that can't be done using knobs warrent another public story

src/components/textarea/textarea.stories.ts
```diff
@@ -0,0 +1,89 @@
...
+  },
+  ...storyFilters()
+};
+
+export const simple = (): string => html`
+  <div style="width:300px;text-align:center;">
+    <calcite-textarea
+      scale="${select("scale", ["s", "m", "l"], "m")}"
+      placeholder="${text("placeholder", "Add Notes")}"
+      ${boolean("disabled", false)}
```
benelan, line 19: can you add a `readonly` knob

src/components/textarea/textarea.stories.ts
```diff
@@ -0,0 +1,89 @@
...
+export const withLabel = (): string => html`<div style="width:300px;max-width:100%;">
+<calcite-label> Add Notes
+<calcite-textarea
+>
+</caclite-textarea>
+</calcite-label>
+</div>`;
+
+export const arabicLang_TestOnly = (): string => html`<div style="width:300px;text-align:center;">
+  <calcite-textarea value="${text("placeholder", "Rocky Mountains National Park")}" lang="ar"> </calcite-textarea>
```
benelan, line 73: TestOnly stories shouldn't use Knobs, see https://github.com/Esri/calcite-components/blob/master/conventions/Documentation.md#screenshot-test-only-stories

Can you change this one to

`<calcite-textarea lang="fr" max-length="123456" group-separator></calcite-textarea>`

so we can make sure the group separator is changing correctly

src/components/textarea/textarea.stories.ts
```diff
@@ -0,0 +1,89 @@
...
+</div>`;
+
+export const exceedingMaxlength_TestOnly = (): string => html`<div style="width:300px;max-width:100%;">
+<calcite-textarea
+value="${text("placeholder", "Rocky Mountains National Park")}"
+maxlength="${number("maxlength", 10)}"
```
benelan, line 79: These shouldn't be knobs either

src/components/textarea/textarea.stories.ts
```diff
@@ -0,0 +1,89 @@
...
+>
+</caclite-textarea>
+</div>`;
+
+export const chineseLangNumberingSystem_TestOnly = (): string => html`div style="width:300px;text-align:center;">
+<calcite-textarea value="${text(
+  "placeholder",
+  "Rocky Mountains National Park"
+)}" lang="zh-cn" numbering-system="hans"> </calcite-textarea>
```
benelan, line 88: Can you change this one to use lang/numberingSystems we support and to have a maxLength, so that we can make sure the numbering system is changing correctly

`<calcite-textarea lang="zh-CN" numbering-system="hanidec" group-separator max-length="654321"></calcite-textarea>`

src/components/textarea/textarea.scss
```diff
@@ -0,0 +1,164 @@
...
+    min-block-size: theme("spacing.24");
+    block-size: calc(100% - 2.75rem);
+  }
+  .footer {
+    @apply py-3 px-4 h-11;
+  }
+  @include slotted("footer-leading", "*") {
+    @apply pr-4;
+  }
+  @include slotted("footer-trailing", "*") {
```
benelan, line 122: This should be `footer-end`

src/components/textarea/textarea.scss
```diff
@@ -0,0 +1,164 @@
...
+  @apply text-0;
+  textarea {
+    @apply py-3 px-4;
+    min-block-size: theme("spacing.24");
+    block-size: calc(100% - 2.75rem);
+  }
+  .footer {
+    @apply py-3 px-4 h-11;
+  }
+  @include slotted("footer-leading", "*") {
```
benelan, line 119: this should be `footer-start`

src/components/textarea/textarea.scss
```diff
@@ -0,0 +1,164 @@
...
+    min-block-size: theme("spacing.20");
+    block-size: calc(100% - 2.25rem);
+  }
+  .footer {
+    @apply py-2 px-3 h-9;
+  }
+  @include slotted("footer-leading", "*") {
+    @apply pr-3;
+  }
+  @include slotted("footer-trailing", "*") {
```
benelan, line 96: same with these two

src/components/textarea/textarea.scss
```diff
@@ -0,0 +1,164 @@
...
+    min-block-size: theme("spacing.16");
+    block-size: calc(100% - 1.75rem);
+  }
+  .footer {
+    @apply py-1 px-2 h-7;
+  }
+  @include slotted("footer-leading", "*") {
+    @apply pr-2;
+  }
+  @include slotted("footer-trailing", "*") {
```
benelan, line 71: and these

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,439 @@
...
+
+  /** Specifies the number or rows allowed. */
+  @Prop({ reflect: true }) rows: number;
+
+  /** Specifies the number or columns allowed. */
+  @Prop({ reflect: true }) cols: number;
+
+  /** Specifies the maximum number of characters allowed. */
+  @Prop() maxlength: number;
```
benelan, line 104: Can you change this to `maxLength` to match `calcite-input`

src/components/textarea/textarea.stories.ts
```diff
@@ -0,0 +1,89 @@
...
+  </div>
+`;
+
+export const withSlottedElements = (): string => html`<div style="width:300px;max-width:100%;">
+<calcite-textarea
+scale="${select("scale", ["s", "m", "l"], "m")}"
+placeholder="${text("placeholder", "Add Notes")}"
+maxlength="${number("maxlength", 50)}"
+>
+<calcite-button  slot="${text("slot", "footer-leading")}">RESET</calcite-button>
```
benelan, line 39: This needs to be `footer-start`. Can you also add something into `footer-end`, maybe Action/Chip/Link, if any of those make sense UI-wise

src/components/textarea/textarea.stories.ts
```diff
@@ -0,0 +1,89 @@
...
+>
+</caclite-textarea>
+</div>`;
+
+export const chineseLangNumberingSystem_TestOnly = (): string => html`div style="width:300px;text-align:center;">
+<calcite-textarea value="${text(
+  "placeholder",
+  "Rocky Mountains National Park"
+)}" lang="zh-cn" numbering-system="hans"> </calcite-textarea>
+</div>`;
```
benelan, line 89: Can you add an `darkModeRTL_TestOnly` story please, see https://github.com/Esri/calcite-components/blob/master/conventions/Documentation.md#screenshot-test-only-stories

benelan (changes requested):
Some more comments. It looks like something happened to the `0` when the textarea is empty too 
image

anveshmekala:
> Some more comments. It looks like something happened to the `0` when the textarea is empty too image

I couldn't able to reproduce this in locale and hosted build. Is there a specific scenario that cause this issue?

src/components/textarea/textarea.tsx
anveshmekala: `maxlength` should be single word without camelcase to match the spec from native attribute https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength.

benelan:
Using this as the demo repros for me

```html
<!DOCTYPE html>
<html>

<head>
  <meta charset="UTF-8" />
  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
  <meta http-equiv="X-UA-Compatible" content="ie=edge" />
  <title>TextArea</title>
  <style>
    calcite-textarea {
      width: 320px;
      height: 160px;
    }
  </style>
  <script src="./_assets/head.js"></script>
</head>

<body>
  <calcite-textarea maxlength="100"></calcite-textarea>
</body>

</html>
```

anveshmekala:
> Using this as the demo repros for me
> 
> ```
> <!DOCTYPE html>
> <html>
> 
> <head>
>   <meta charset="UTF-8" />
>   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
>   <title>TextArea</title>
>   <style>
>     calcite-textarea {
>       width: 320px;
>       height: 160px;
>     }
>   </style>
>   <script src="./_assets/head.js"></script>
> </head>
> 
> <body>
>   <calcite-textarea maxlength="100"></calcite-textarea>
> </body>
> 
> </html>
> ```

Great catch. When there is a space between the open tag and closing tag `0` is displayed.  

Using `<calcite-textarea maxlength="100"> </calcite-textarea>` display `0` but when the space is removed  between the tag it is disappeared.

src/components/textarea/textarea.tsx
anveshmekala: Changed this to `maxLength` to be consistent with `calcite-input`.

src/demos/textarea-issue.html
```diff
@@ -0,0 +1,131 @@
+<!DOCTYPE html>
```
anveshmekala, line 1: Added this file for debugging only. this will be removed before install.

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,498 @@
...
+   * @mdn [rows](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea#attr-rows)
+   */
+  @Prop({ reflect: true }) rows: number;
+
+  /**
+   * Specifies the number or columns allowed.
+   *
+   * @mdn [cols](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea#attr-cols)
+   */
+  @Prop({ reflect: true }) cols: number;
```
altrae, line 117: I've got a basic question/concern about the overall organization of this component. I do like that areas are split up using sectional comment headers, but would it be possible to alpha-sort where possible such as the Props and move lifecycles to the bottom of the component at the very least? Also, adding private to non-public properties might help differentiate what is public and what is not.

I know these are just my personal preferences so I'm curious as to what others think. I've seen similar organization as what I'm suggesting in other Calcite components (button, card and others seem to have started out this way but looks like they weren't kept up with the addition of messages props amongst other things), but I haven't found any Calcite guidelines for this specifically. I do know there is a [component boilerplate](https://github.com/Esri/calcite-components/blob/master/BOILERPLATE_COMPONENT.md), but it doesn't seem to be organized in any meaningful way itself and the [styleguide](https://github.com/ionic-team/stencil/blob/main/STYLE_GUIDE.md) also doesn't make mention of these things yet.

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,498 @@
...
+
+  footerEl: HTMLElement;
+
+  @State() effectiveLocale = "";
+
+  @State() endSlotElements: Element[];
+
+  @State() startSlotElements: Element[];
+
+  @Watch("effectiveLocale")
```
altrae, line 352: Should this be grouped with the other Watchers?

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,498 @@
...
+  };
+
+  footerEndSlotChangeHandler = (event: Event): void => {
+    this.endSlotElements = slotChangeGetAssignedElements(event);
+    if (this.disabled) {
+      this.disablePointerEvents(this.disabled, this.endSlotElements);
+    }
+  };
+
+  footerStartSlotChangeHandler = (event: Event): void => {
```
altrae, line 389: Could possibly combine both footerEndSlotChangeHandler and footerStartSlotChangeHandler into one function (haven't tested this but saw a pattern here):

```
private footerSlotChangeHandler = (event: Event, position: 'end' | 'start' = 'start') => {
  const slotElements = `${position}SlotElements`;

  this[slotElements] = slotChangeGetAssignedElements(event);

  if (this.disabled) this.disablePointerEvents(this.disabled, this.[slotElements]);
};
```

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,498 @@
...
+    LoadableComponent,
+    T9nComponent,
+    InteractiveComponent
+{
+  //--------------------------------------------------------------------------
+  //
+  //  Element
+  //
+  //--------------------------------------------------------------------------
+  @Element() el: HTMLCalciteTextareaElement;
```
altrae, line 69: I always wonder why this variable isn't just named `host`. It always references the host so it makes sense to name it as host and it also reduces confusion later in the component because we know we are referring to the host and not some other random element.

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,498 @@
...
+  syncHiddenFormInput(input: HTMLInputElement): void {
+    input.setCustomValidity("");
+    if (this.value?.length > this.maxLength) {
+      input.setCustomValidity(this.messages.overLimit);
+    }
+  }
+
+  setHeightAndWidthToAuto = throttle(
+    (): void => {
+      this.verticalResizeDisabled || (this.el.style.height = "auto");
```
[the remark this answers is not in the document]
jcfranco: This doesn't seem to be that performance intensive. Was this being throttled to batch style updates? If so, I think it'd be better to use `writeTask` for this.

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,498 @@
...
+    { leading: false, trailing: true }
+  );
+
+  setTextareaEl = (el: HTMLTextAreaElement): void => {
+    this.textareaEl = el;
+    this.resizeObserver.observe(el);
+  };
+
+  setTextareaHeight(): void {
+    const { textareaHeight, elHeight, footerHeight } = this.getHeightandWidthOfElements();
```
jcfranco, line 470: Might be worth testing w/ `readTask` for this. This can be done in a follow-up PR.

jcfranco (changes requested):
This is looking great, @anveshmekala! 😎

src/index.html
```diff
@@ -386,6 +386,19 @@ <h1 id="calcite-demo">Calcite demo</h1>
...
         </div>
 
+        <div>
+          <a href="/demos/textarea.html">
+            <calcite-action scale="s" text="Textarea" alignment="start" text-enabled appearance="solid">
+            </calcite-action>
+          </a>
+        </div>
+        <div>
+          <a href="/demos/textarea-issue.html">
```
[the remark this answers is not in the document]
anveshmekala: Yup. this will be removed.

src/components/textarea/textarea.stories.ts
```diff
@@ -0,0 +1,118 @@
...
+readonly
+>
+</caclite-textarea>
+</div>`;
+
+export const resizeDisabled_TestOnly = (): string => html`<div style="width:300px;max-width:100%;">
+<calcite-textarea
+resize-disabled
+>
+</caclite-textarea>
```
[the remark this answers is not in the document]
anveshmekala: Great catch. The screenshots were working at some point. Might have made the change after that.

src/components/textarea/textarea.stories.ts
```diff
@@ -0,0 +1,118 @@
...
+${boolean("readonly", false)}
+label="${text("label", "")}"
+name="${text("name", "")}"
+>
+<calcite-button  slot="${text("slot", "footer-start")}">RESET</calcite-button>
+<calcite-action  icon="code" slot="${text("slot", "footer-end")}"></calcite-action>
+</caclite-textarea>
+</div>`;
+
+export const withSlottedElementsDarkModeRTL_TestOnly = (): string => html`<div style="width:300px;max-width:100%;">
```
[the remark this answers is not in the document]
anveshmekala: The reason was to limit the height and width of the textarea rather than setting it to fullwidth.

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,498 @@
...
+      !!this.startSlotElements?.length || !!this.endSlotElements?.length || !!this.maxLength;
+
+    return (
+      <Host>
+        <textarea
+          aria-invalid={toAriaBoolean(this.value?.length > this.maxLength)}
+          aria-label={getLabelText(this)}
+          autofocus={this.autofocus}
+          class={{
+            textarea: true,
```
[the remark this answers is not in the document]
anveshmekala: Good catch, this isn't required. removed it.

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,498 @@
...
+          onInput={this.handleInput}
+          placeholder={this.placeholder}
+          readonly={this.readonly}
+          ref={this.setTextareaEl}
+          required={this.required}
+          rows={this.rows}
+          value={this.value}
+          wrap={this.wrap}
+        />
+        <span class="content">
```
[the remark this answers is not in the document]
anveshmekala: These are non conditional classes, do we need to add them in CSS object?

src/components/textarea/textarea.tsx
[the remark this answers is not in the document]
anveshmekala: This is being throttle to avoid flickering of drag handle when the user drags quick where and the height and width not set to `auto` yet. This is inconsistent and with throttling we are delaying it by 100ms which provides a better experience. And also subsequent tasks will be batched and executed.

src/components/textarea/textarea.tsx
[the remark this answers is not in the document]
altrae: I've had similar conversations with my teams regarding attributes such as `autocomplete` where we decided to keep it as to remain less confusing when external developers attempt to apply the attributes in the way they are used to doing so. In regards to component naming though, I would like to add to the argument for separation due to the fact they are already different via the `calcite-` prefix (i.e. `calcite-textarea` vs `textarea` so taking this one step further with `calcite-text-area` shouldn't really be any more confusing :)).

src/components/textarea/textarea.tsx
altrae: That makes sense! Based on your comment I now see it is watching a State variable. I honestly haven't seen Watchers used for State before, only Props, so I presumed this was for a Prop when I made my suggestion. :)

src/components/textarea/textarea.stories.ts
```diff
@@ -0,0 +1,118 @@
...
+</div>`;
+
+export const resizeDisabled_TestOnly = (): string => html`<div style="width:300px;max-width:100%;">
+<calcite-textarea
+resize-disabled
+>
+</caclite-textarea>
+</div>`;
+
+export const withLabel_TestOnly = (): string => html`<div style="width:300px;max-width:100%;">
```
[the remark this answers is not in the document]
anveshmekala: i think this is needed to understand how the label is showing up in with text-area component.

src/components/textarea/textarea.tsx
anveshmekala: yeah we can combine them but in the new iteration these functions are no longer needed.

src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,498 @@
...
+  @Prop({ mutable: true }) messageOverrides: Partial<TextareaMessages>;
+
+  @Watch("messageOverrides")
+  onMessagesChange(): void {
+    /* wired up by t9n util */
+  }
+
+  @Watch("disabled")
+  disabledHandler(value: boolean): void {
+    this.disablePointerEvents(value, this.startSlotElements);
```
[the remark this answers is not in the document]
anveshmekala: yeah this is not needed. I was unaware of it , updated the code accordingly.

src/demos/textarea.html
```diff
@@ -0,0 +1,281 @@
...
+        </div>
+        <div class="child">
+          <calcite-textarea placeholder="add notes" scale="m" vertical-resize-disabled> </calcite-textarea>
+        </div>
+        <div class="child">
+          <calcite-textarea placeholder="add notes" scale="l" horizantal-resize-disabled> </calcite-textarea>
+        </div>
+      </div>
+
+      <!-- With max-length -->
```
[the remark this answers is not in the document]
anveshmekala: removed for this component.

src/components/textarea/textarea.scss
```diff
@@ -0,0 +1,147 @@
...
+:host textarea.block-size--full {
+  @apply h-full;
+}
+
+.content,
+.hide {
+  @apply hidden;
+}
+
+@include disabled();
```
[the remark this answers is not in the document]
anveshmekala: i am moving them to the last of the page for now.

src/components/textarea/textarea.e2e.ts
```diff
@@ -0,0 +1,223 @@
...
+    await page.keyboard.press("Tab");
+    await page.waitForChanges();
+
+    await page.keyboard.type("rocky mountains");
+    await page.waitForChanges();
+
+    expect(await element.getProperty("value")).toBe("rocky mountains");
+  });
+
+  it("should disable slotted elements when disabled", async () => {
```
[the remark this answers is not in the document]
anveshmekala: i don't think the disabled test helper is checking if the slotted elements are disabled.

src/components/textarea/textarea.tsx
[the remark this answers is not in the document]
anveshmekala: as discussed added them in the CSS object.

src/components/textarea/textarea.stories.ts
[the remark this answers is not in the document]
anveshmekala: removed the wrapper and added a test specific to wrapper.

src/components/text-area/resources.ts
```diff
@@ -8,11 +8,12 @@ export const CSS = {
   resizeDisabledY: "resize--disabled-y",
   characterOverlimit: "character--over-limit",
   readonly: "readonly",
-  textareaInvalid: "textarea--invalid",
+  textAreaInvalid: "text-area--invalid",
   footerSlotted: "footer--slotted",
   borderColor: "border--color",
   hide: "hide",
-  blocksizeFull: "block-size--full"
+  blocksizeFull: "block-size--full",
```
jcfranco, line 15: `blocksizeFull` ➡️ `blockSizeFull`

src/components/text-area/text-area.e2e.ts
```diff
@@ -70,29 +70,13 @@ describe("calcite-text-area", () => {
       submitsOnEnter: false
     }));
 
-  it("should emit calciteTextareaInput event when user starts typing", async () => {
+  it("should emit calciteTextareaInput event when user type in the textarea and emit calciteTextareaChange when users tabs out", async () => {
```
jcfranco, line 73: `calciteTextareaInput` ➡️ `calciteTextAreaInput`
`calciteTextareaChange` ➡️ `calciteTextAreaChange`

src/components/text-area/text-area.tsx
```diff
@@ -200,12 +194,12 @@ export class TextArea
   /**
    * Fires each time a new `value` is typed.
    */
-  @Event({ cancelable: true }) calciteTextareaInput: EventEmitter<void>;
+  @Event({ cancelable: true }) calciteTextAreaInput: EventEmitter<void>;
```
jcfranco, line 197: I don't see any logic to deal with cancelation. If these are not cancelable, can you update the event decorator options?

src/components/text-area/text-area.tsx
```diff
@@ -317,13 +312,13 @@ export class TextArea
   @Method()
   async setFocus(): Promise<void> {
     await componentLoaded(this);
-    this.textareaEl.focus();
+    this.textAreaEl.focus();
   }
 
   /** Selects all text of the component's `value`. */
```
jcfranco, line 318: +@geospatialem for editorial review.

src/components/text-area/text-area.tsx
```diff
@@ -337,13 +332,13 @@ export class TextArea
 
   labelEl: HTMLCalciteLabelElement;
 
-  textareaEl: HTMLTextAreaElement;
+  textAreaEl: HTMLTextAreaElement;
 
   footerEl: HTMLElement;
 
-  endSlotHasElements: boolean;
+  @State() endSlotHasElements: boolean;
```
jcfranco, line 339: Can you make sure component props/variables in alphabetical order? I think we can keep variables w/o decorators separate from the ones that use them.

src/components/label/label.stories.ts
```diff
@@ -77,6 +77,10 @@ export const simple = (): string => html`
         <calcite-option>c</calcite-option>
       </calcite-select>
     </calcite-label>
+    <calcite-label>
+      Add Notes
+      <calcite-text-area> </calcite-text-area>
```
jcfranco, line 82: Can you double check all `calcite-text-area` usages don't have a space?

This applies to other files as well (e.g., stories).

src/components/text-area/assets/text-area/t9n/messages_en.json
```diff
@@ -0,0 +1,4 @@
+{
+  "invalid": "Invalid",
+  "overLimit": "Character limit exceeded"
```
jcfranco, line 3: Suggestion: as a shorter alternative, the native textarea uses `tooLong` as the name for this error. I'll leave this up to you.

src/components/text-area/text-area.stories.ts
```diff
@@ -0,0 +1,85 @@
...
+export const readonly_TestOnly = (): string => html` <calcite-text-area readonly> </calcite-text-area> `;
+
+export const resizeDisabled_TestOnly = (): string => html` <calcite-text-area resize="none"> </calcite-text-area> `;
+
+export const groupSeparator_TestOnly = (): string => html`
+  <calcite-text-area value="Rocky Mountains National Park" lang="fr" max-length="123456" group-separator>
+  </calcite-text-area>
+`;
+
+export const exceedingMaxlength_TestOnly = (): string => html`
```
jcfranco, line 69: `exceedingMaxlength_TestOnly` ➡️ `exceedingMaxLength_TestOnly`

jcfranco (approved):
Awesome work, @anveshmekala! 

One the last comments are addressed, this is good to merge! 🎉

src/components/text-area/text-area.tsx
anveshmekala: matched with `calcite-input` but i feel this can be modified to `Selects the text of the component ` may be.

src/components/text-area/text-area.stories.ts
anveshmekala: i swear i made this change earlier. :open_mouth:

src/components/text-area/assets/text-area/t9n/messages_en.json
anveshmekala: I prefer using more descriptive error. This suggestion was made during design discussions. WDYT @ashetland ?

src/components/text-area/text-area.e2e.ts
```diff
@@ -172,39 +141,27 @@ describe("calcite-text-area", () => {
     expect(element).toHaveClass(CSS.footerSlotted);
   });
 
-  describe("resize disabled", () => {
-    it("should have resize-disabled class", async () => {
+  describe("resize", () => {
+    it("should set CSS resize property to horizontal", async () => {
```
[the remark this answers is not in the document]
anveshmekala: Yup, added https://github.com/Esri/calcite-components/issues/6624 with the discussed approach.

src/components/text-area/assets/text-area/t9n/messages_en.json
ashetland: Are we talking about the name or the message?

anveshmekala: checked with Franco, it is regarding the name. Sorry for the confusion. I think prop name can be changed to `tooLong`.

src/components/text-area/text-area.tsx
```diff
@@ -0,0 +1,476 @@
...
+  setUpMessages,
+  T9nComponent,
+  updateMessages
+} from "../../utils/t9n";
+import { TextAreaMessages } from "./assets/text-area/t9n";
+import { throttle } from "lodash-es";
+import { InteractiveComponent, updateHostInteraction } from "../../utils/interactive";
+
+/**
+ * @slot - A slot for adding text.
```
geospatialem, line 44: For default slots and consistency we tried to mention content, not sure if that's applicable here, but WDYT about:

> A slot for adding content.

src/components/text-area/text-area.tsx
```diff
@@ -0,0 +1,476 @@
...
+ * @slot footer-start - A slot for adding a leading footer.
```
geospatialem, line 45: nitpick:

> A slot for adding content to the start of the component's footer.

src/components/text-area/text-area.tsx
```diff
@@ -0,0 +1,476 @@
...
+  @Element() el: HTMLCalciteTextAreaElement;
+
+  //--------------------------------------------------------------------------
+  //
+  //  Properties
+  //
+  //--------------------------------------------------------------------------
+
+  /**
+   * When `true`, the component is focused on page load.
```
geospatialem, line 78: Could this be the case if the component is added to the DOM after load? If so, perhaps mention initialization? 

> When `true`, the component is focused on initialization.

src/components/text-area/text-area.tsx
```diff
@@ -0,0 +1,476 @@
...
+   *
+   * @mdn [autofocus](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus)
+   */
+  @Prop({ reflect: true }) autofocus = false;
+
+  /**
+   * Specifies the number or columns allowed.
```
geospatialem, line 85: > Specifies the component's number of columns.

src/components/text-area/text-area.tsx
geospatialem: This and `input` does read a little difficult (I'll tackle that one in a different PR).

Concur with the proposal you have above:
> Selects the text of the component.

geospatialem (commented):
A few editorial comments below.

src/components/text-area/text-area.tsx
anveshmekala: content sounds better.

anveshmekala: i think it is mentioned as `text` for some components in the cc repo ( Ex: chip, button, link, input-text). i will leave it up to you @geospatialem

src/components/text-area/text-area.tsx
anveshmekala: I don't think if initialization captures the meaning. When the page is loaded with components  rendered  then the autofocus will kick in and applied to the component. Initialization could mean it is first attached to the DOM not essentially rendered. Thoughts?

src/components/text-area/text-area.tsx
geospatialem: Ah, yes, in those cases we're trying to be focused on text, which might be the case here as well. Let's leave as-is for consistency with the other components. Thanks @anveshmekala!

src/components/text-area/text-area.tsx
geospatialem: Ah okay, how about adding another line for context instead?:

```
When `true`, the component is focused on page load. Only one element can contain `autofocus`. If multiple elements have `autofocus`, the first element will receive focus.
```

anveshmekala: we are adding a reference to `mdn` website which provides the additional information. do we want to add it in here as well?

anveshmekala: Interesting.  Will add the additional context. FWIW, this needs to be updated for input components as well.

src/components/text-area/text-area.tsx
```diff
@@ -0,0 +1,476 @@
...
+  @Method()
+  async setFocus(): Promise<void> {
+    await componentLoaded(this);
+    this.textAreaEl.focus();
+  }
+
+  /** Selects the text of the component's `value`. */
+  @Method()
+  async selectText(): Promise<void> {
+    this.textAreaEl.select();
```
benelan, line 321: Should `selectText`  await `componentLoaded(this)` like `setFocus` does?

src/components/text-area/text-area.tsx
benelan: I plan on supporting `@mdn` as a custom jsdoc tag so we can link to it in the API Ref when our prop is passed to a native attribute

benelan (approved):
A couple comments but LGTM!

src/components/text-area/text-area.tsx
anveshmekala: i think it should.

PR #5645, opened 2022-10-26
fix(date-picker): cached formatter does not conflict with our numberingSystem components

benelan:
**Related Issue:** #5488

## Summary
There is an inconsistent bug when there are multiple date pickers with different numberingSystems. Sometimes one will impact the numberingSystem of the others due to the timing of when the formatter was cached. We couldn't think of a use case for multiple numberingSystems, but it still seemed like something that needed to be fixed.

I was setting the formatter options in date-picker, it also needs to be set it in `date-picker-day` and `date-picker-month-header`. Setting the formatter options in `componentWillLoad` works for the first render. However, these ones do need to go in the render block, because currently if you switch months in one `date-picker` and then switch months in another, sometimes the `numberingSystem` would also switch. This PR resolves the bug.


eriklharper (approved):
Looks awesome!  I tested this locally and it is legit!  Nice work :)

PR #5649, opened 2022-10-26
fix(types): fix type issue caused by unintentionally moving @floating-ui/dom as a dev dependency

jcfranco:
**Related Issue:** N/A

## Summary

This fixes a regression caused by https://github.com/Esri/calcite-components/pull/5542 (bumped as a devdep vs dep 😅 #derp). cc @AdelheidF


driskull (approved):
💯 
☉ ‿ ⚆

PR #5654, opened 2022-10-27
ci: stop generating readmes on every build

benelan:
**Related Issue:** #

## Summary

All those readmes that are generated on every build have been slowly driving me to madness. I've had it - NO MORE, I SAY!

The only time we actually need the readmes built is in the GH Action, which doesn't even use `npm run build`. 
It runs `npx stencil build --docs`, which will still generate the docs after this change. 
We also have a util script specifically for building docs, which will still do so: `npm run util:build-docs`.

We can free ourselves from the opressive reign of these markdown clad menaces! They may be many, but we are strong. Who will ~~ride with me~~ approve this PR?

FCm6URe

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,9 +17,9 @@
     "hydrate/"
   ],
   "scripts": {
-    "build": "npm run util:prep-build-reqs && stencil build && npm run util:patch",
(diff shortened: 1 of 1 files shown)


jcfranco (approved):
Twas merely a doc-generation wound. ⚔🛡️🩸

PR #5656, opened 2022-10-27
fix(input-message,loader): reverts breaking change

geospatialem:
**Related Issue:** #3821 

## Summary
Reverts breaking change for `loader` and `input-message`'s `active` props from #5537.

To be incorporated in the upcoming CC release, `beta.97`.


driskull (commented):
Looks good @geospatialem Thanks for getting this together!

Can we update the changelog here too to remove the entry for the feature?

geospatialem:
> Can we update the changelog here too to remove the entry for the feature?

@driskull Good call! Removed from the entry.

driskull (approved):
👍

PR #5659, opened 2022-10-27
fix(button, fab): adjust padding on 'l' scale button to accommodate 'm' scale icon without change in height

Elijbet:
**Related Issue:** #5245

## Summary

Design adjustment on [#5521](https://github.com/Esri/calcite-components/pull/5521). The height of the `large` component changed when a new `medium` (24px) icon was introduced. This PR will meet the design recommendation to maintain the 44px height with or without an icon, accounting for both the default `solid` appearance with a `border` and `transparent` without. Every other appearance option also has a border.


src/components/button/button.scss
```diff
@@ -659,12 +659,26 @@
 :host([scale="l"]:not([appearance="transparent"])) a.content--slotted {
   --calcite-button-padding-y: 11px;
 }
+//for 'l' solid buttons shrink the padding if icon is present to preserve the height
+:host([scale="l"][icon-start]:not([appearance="transparent"])) button.content--slotted,
```
driskull, line 663: This is a pretty advanced selector. Should we move this logic into the component and add a class for it? Could always be a refactor issue if we don't want to do it now.

driskull (commented):
Will screener pick up these changes somewhere? We need a visual test here so it doesn't break :)

src/components/button/button.scss
Elijbet: I'm all up for doing it now.

src/components/button/button.scss
```diff
@@ -229,15 +229,16 @@
 }
 
 // button styles
-// solid
-:host([appearance="solid"]) {
+:host([appearance="solid"]),
+:host([appearance=""]) {
```
driskull, line 233: Does this need the `appearance=""`? 

Would `[appearance]` be sufficient?

src/components/button/button.scss
```diff
@@ -655,15 +656,26 @@
...
-:host([scale="l"]:not([appearance="transparent"])) button.content--slotted,
-:host([scale="l"]:not([appearance="transparent"])) a.content--slotted {
-  --calcite-button-padding-y: 11px;
+
+:host([scale="l"]:not([appearance="transparent"])) {
+  .button-padding {
+    --calcite-button-padding-y: 11px;
+  }
+  //shrink the padding if icon is present to preserve the height
+  .button-padding-shrunk {
```
driskull, line 665: nitpick: I think the selector for this ideally would be `button-padding--shrunk`. using BEM naming methodology.  https://getbem.com/

src/components/button/button.scss
```diff
@@ -655,15 +656,26 @@
...
+    --calcite-button-padding-y: 9px;
+  }
 }
 // accommodate for transparent buttons not having borders
-:host([scale="l"][appearance="transparent"]) button.content--slotted,
-:host([scale="l"][appearance="transparent"]) a.content--slotted {
-  --calcite-button-padding-x: theme("padding.4");
-  --calcite-button-padding-y: theme("padding.3");
+:host([scale="l"][appearance="transparent"]) {
+  .button-no-border-padding {
```
driskull, line 671: nitpick: I think the selector for this ideally would be `button--no-border-padding` using BEM naming methodology.  https://getbem.com/

src/components/button/button.tsx
```diff
@@ -204,6 +204,10 @@ export class Button implements LabelableComponent, InteractiveComponent, FormOwn
       <Tag
         aria-label={getLabelText(this)}
         class={{
+          [CSS.buttonNoBorderPadding]: !this.iconStart && !this.iconEnd,
```
driskull, line 207: Maybe add a const for `!this.iconStart && !this.iconEnd`?

Like 

```ts
const noStartEndIcons = !this.iconStart && !this.iconEnd;
```

```ts
[CSS.buttonNoBorderPadding]: noStartEndIcons,
[CSS.buttonNoBorderPaddingShrunk]: !noStartEndIcons,
[CSS.buttonPadding]: noStartEndIcons,
[CSS.buttonPaddingShrunk]: !noStartEndIcons,
```

Looking at the above, it seems that two of those classes are the same right? Could it be reduced to...


```ts
[CSS.noStartEndIcons]: noStartEndIcons,
[CSS.hasStartEndIcons]: !noStartEndIcons,
```

src/components/button/button.stories.ts
```diff
@@ -117,6 +118,24 @@ export const withIconEndEmpty_TestOnly = (): string => html` <calcite-button ico
 
 withIconEndEmpty_TestOnly.storyName = "With icon-end set to empty";
 
+export const largeWithMediumIcons_TestOnly = stepStory(
```
driskull, line 121: 👍

src/components/button/button.scss
driskull: If this is looking for when there is no appearance set, I think `:not([appearance])` would work. Or a css class could be used here.

driskull (changes requested):
Review feedbackz! ʘ‿ʘ

src/components/button/button.scss
Elijbet: Looks like I’ve been testing for redundant cases, such as having the appearance set to an empty string. That’s not really an option and I’ve learned that we’ve decided not to address invalid cases. So I simplified the code quite a bit as a result.

src/components/button/button.scss
```diff
@@ -229,15 +229,16 @@
 }
 
 // button styles
-// solid
-:host([appearance="solid"]) {
+:host([appearance="solid"]),
+:host([appearance]) {
```
driskull, line 233: I'm not sure why we need the `:host([appearance="solid"]),` if `:host([appearance]) {` covers all cases right?

src/components/button/button.scss
```diff
@@ -629,41 +630,49 @@
...
 }
+
 // accommodate for transparent buttons not having borders
 :host([scale="s"][appearance="transparent"]) button.content--slotted,
 :host([scale="s"][appearance="transparent"]) a.content--slotted {
   --calcite-button-padding-x: theme("padding.2");
-  --calcite-button-padding-y: theme("padding.1");
 }
+
+:host([scale="s"]) button,
```
driskull, line 640: nitpick: Ideally, both `button` and `a` have a class so we can just use that instead of the tag.

driskull (approved):
👍 Looks good @Elijbet

src/components/button/button.scss
Elijbet: Would this require quite a bit of refactoring, since button and a have been used throughout the file?

driskull: yeah, feel free to ignore it.

PR #5666, opened 2022-10-28
fix: Floating-ui should handle visibility and pointer-events on the floating element.

driskull:
**Related Issue:** #

## Summary

fix: Floating-ui should handle visibility and pointer-events on the floating element.

- `floating-ui` sets `visibility` on a component depending on if the reference element is out of view.  https://floating-ui.com/docs/hide#strategy
- `floating-ui` sets pointer-events to none depending on the visibility.

Because of the above, we should not be setting `visibility` or `pointer` via CSS and let `floating-ui` handle it so that there are no conflicts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,11 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [1.0.0-beta.96](https://github.com/Esri/calcite-components/compare/v1.0.0-beta.95...v1.0.0-beta.96) (2022-10-27)
 
+### ⚠ BREAKING CHANGES
+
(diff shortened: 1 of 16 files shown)


jcfranco (approved):
Awesome! 🎉

Can you update the spec test before merging? I think we only need to assert on the initial styles [here](https://github.com/Esri/calcite-components/blob/master/src/utils/floating-ui.spec.ts#L106) and [here](https://github.com/Esri/calcite-components/blob/master/src/utils/floating-ui.spec.ts#L117).

driskull:
Seems to have broke the `dropdown`. Need to investigate more 🔍

driskull:
Working good now. 👍

Issue #5419, opened 2022-10-05
[Combobox] Long Item text-labels should wrap


omarkawach:
### Actual Behavior

With a combobox and combobox items set to a fixed width of 250px, an item's text-label should get wrapped if the label is too long. 

When a label is long and has spaces, I noticed that the wrapping works well. However, if the label does not have spaces, there is no wrapping.

Not all users name their layers with spaces. Some use underscores, camel cases, etc.

![Screen Shot 2022-10-05 at 11 30 29](https://user-images.githubusercontent.com/49499098/194138555-2e07181d-b6b5-4b90-958b-049309bc23ce.png)

### Expected Behavior

Long labels with spaces wrap without issue, but I expect all long labels to wrap so they stay contained within a fixed width and with no need for a scrollbar.

### Reproduction Sample

https://codepen.io/omarkawach/pen/wvjxXOK

### Reproduction Steps

1. Open the codepen
2. Go to the combobox on the right
3. Click to see the combobox items
4. Observe how one long label is wrapped but not the other
5. Uncomment the other blocks of HTML code to see additional cases

### Reproduction Version

beta.95

### Relevant Info

MacOS, Version 106

### Esri team

ArcGIS Online

geospatialem:
Verified on `next.631`.

![image](https://user-images.githubusercontent.com/5023024/202221957-0033f7a0-4ca2-429d-a5ad-a937f3edf780.png)

PR #5672, opened 2022-10-31
fix(combobox): Wrap and break text on long items

driskull:
**Related Issue:** #5419

## Summary

fix(combobox): Wrap and break text on long items. (#5419)


jcfranco (approved):
🎁

Issue #5270, opened 2022-09-04
Auto focus modal content by default


janett-baresel:
### Summary

When using a modal dialog the content is not auto focused.

### Actual Behavior

Opening a modal from a button will not auto focus the content. The focus stays in the document and continues "behind".

### Desired Behavior

When opening a modal dialog the focus is automatically moved to the content or close button. Basically, call setFocus by default when the modal is opened. I cannot think of any user who wouldn't want this behavior.

### Reproduction Sample

https://codepen.io/Programmine/pen/RwyNpJe

### Reproduction Steps

1. create calcite-button that has an onclick event
2. onclick creates modal dialog and opens it
-> focus remains on button and behind modal

### Reproduction Version

latest

### Relevant Info

reproduced on chrome

<sup>SV-CH</sup>

janett-baresel:
@jcfranco I would like to change this from a bug report to an enhancement suggestion. Would I need to create a new ticket and close this one?

benelan:
Hi Jannet, we can use this issue, I just changed the labels. Thanks!

geospatialem:
Verified the `modal` auto focus  in `beta.98` across Chrome, Firefox, and Edge browsers.

Issue #2133, opened 2021-05-13
Bug: calcite-popover - trap focus inside


AdelheidF:
All 3 sub-issues here a related, so I added them into the same issue.

<details>
<summary>(1) (fixed with the use of `calcite-list`)</summary>
<br>
Add an option to value list to not allow selection of the text, only `action-end`.

![image](https://user-images.githubusercontent.com/44481292/118058980-4a433a80-b344-11eb-8140-8c3f70458446.png)

This is a lists of values in the filter component, clicking on a value name does not call an action. Only `action-end` matters here.
![image](https://user-images.githubusercontent.com/44481292/118060144-f1c16c80-b346-11eb-8f8a-77a6baa72a5b.png)

This is a list of detailed information inside a popover. Nothing should happen when you click on the text, only `action-end` matters here as well.
![image](https://user-images.githubusercontent.com/44481292/118060204-0c93e100-b347-11eb-9661-d393a12d27d3.png)
</details>

<details>
<summary>(2) (fixed with the use of `calcite-list`)</summary>
<br>
Navigating in this value list is a bit weird. Initially only the first row, first text then action-end, can be selected. If you then tab again it leaves the list. You need to use the down arrow to navigate inside the list, but the down arrow only works on the text, not on action-end. Same with `calcite-pick-list`.
</details>

(3) 
Focus gets lost if there if no other focusable item after the value-list when using tab key after `action-end` is selected. This is a problem for keyboard only users in the popover case above. How can I catch the focus (or know when the focus leaves value-list) and redirect it back to the close icon of the popover? I tried `onBlur` but that didn't work.
Scenario:
popover opens with close button on focus. Then I tab my way through the list. After the last focusable item in the list is active and I hit tab the focus gets lost.

____________
New list component https://github.com/Esri/calcite-components/issues/2117

### Reproduction Steps or Sample

1.

### Relevant Info

_Version_: `@esri/calcite-components@<version>`

- [ ] CDN
- [ ] NPM package

TheBlueDog:
Follow

TheBlueDog:
Follow

AdelheidF:
(1) and (2) are fixed with the use of the new `calcite-list`. (3) is still valid. Changed the title.

macandcheese:
@AdelheidF can you put together a codepen that illustrates this bug? Thanks!

AdelheidF:
https://codepen.io/afreitag/pen/LYjmaWv?editors=1000

[...]

Issue #1169, opened 2020-10-22
Research: Research using different focus trap npm helper


driskull:
### Background 

Research: Research using different focus trap npm helper

https://www.npmjs.com/package/focus-trap

### Desired Outcome 

Using best focus trap helper

driskull:
@jcfranco I tried using the other focus trap package without any luck. The problem is that I couldn't find a good way to start the focus trap after the modal becomes active but after DOM nodes were already created or updated.

Do you want to take a stab at it?

driskull:
I think the issue is with the tubbable dependency not supporting shadow dom.

driskull:
Going to close this until the focus helper supports shadow dom.

geospatialem:
Works great! Verified in `beta.98` in Chrome, Firefox, and Edge browsers.

Issue #5147, opened 2022-08-13
Consistent focus/blur eventing across components


jcfranco:
**monday.com sync:** #18048591137

### Overview

Currently, developers have to rely on `focusin`/`focusout` (composable + bubbling) from component internals or using, internal, blur/focus events for a few components.† 
The goal for this effort is to provide a general solution for focus/blur events given that these events are commonly used and expected in many web development scenarios.

Also for consideration, `setFocus` has introduced a pattern that can lead to inconsistent focus behavior. Existing usage should be revisited to see if we can consolidate in favor of consistency (one focus pattern to rule them all).   

† In general, this works but leads to boilerplate code as `focusin`/`focusout` fires when you move between elements inside a component and would need to check where focus lies when handling the event.

### Tasks

- [ ] determine if `delegatesFocus` is a viable solution
  - [x] consult design regarding potential double focus outline 
      - **Design suggests hidding the focus ring on the host and only displaying it on the element it delegates focus to. a11y agrees with the recommendation**
      - ** Note I haven't seen any double focus rings yet, it may have been a quirk that is worked out now**

  - [x] check browsers focusing is consistent, if not weigh options
      - **As usual, browsers behave differently when it comes to focus  ([example](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#clicking_and_focus)).  However, since `delegatesFocus` is a web standard, I think we will have better consistency than we would if we try to jerry-rig up our own solution. Once we get it set up in the code we can do more browser testing, but all the browsers have adopted the pattern and are reletaively consistent, based on my research.***

  - [ ] check that focus/blur events will behave as expected for both light/shadow DOM content (if so, we use these and drop our internal focus/blur events)
  - [ ] confirm dev expectations light/shadow DOM content focus/blur (e.g., blurring slotted content not emitting blur on custom element) - https://codepen.io/jcfranco/pen/ZExMgwx?editors=1000
  - [x] check if https://www.npmjs.com/package/delegates-focus-polyfill will fit the bill since Safari 14 doesn't support it
      - **Safari 16 released so we don't need the polyfill**

- [ ] consult design, a11y concerns from having multiple focus targets from `setFocus`
  - if keeping single focusable target, we should consider promoting `calciteComponent.focus/blur()` for shadow-targeted focus/blur or `slottedContent.focus/blur()` for light-targeted focus/blur

### Additional resources

- https://github.com/whatwg/html/issues/7207
[...]

PR #5676, opened 2022-11-01
refactor(modal): Update modal to use focus-trap module.

driskull:
**Related Issue:** #1169 #5270

## Summary

refactor(modal): Update modal to use focus-trap module.

- Replaces outdated `@a11y/focus-trap` module with more popular and supported `focus-trap` module. 
- https://www.npmjs.com/package/focus-trap
- https://www.npmjs.com/package/@a11y/focus-trap

Precursor to #2133


driskull:
@macandcheese @geospatialem @jcfranco could you play around with this and let me know if its working correctly?

src/components/modal/modal.tsx
```diff
@@ -388,6 +375,11 @@ export class Modal implements ConditionalSlotComponent, OpenCloseComponent {
 
   private setTransitionEl = (el): void => {
     this.transitionEl = el;
+    this.trap = focusTrap.createFocusTrap(this.transitionEl, {
```
driskull, line 378: its-a-trap-trap

macandcheese:
Seems to work well! The "tab fencing" demo doesn't seem to work as described, but that may just be out of date sample content. Is it still possible to specify an element to focus first?

macandcheese:
> Yes, that's possible with the module. Do we need to support that for modal? I didn't see anything like that.
> 
> https://github.com/focus-trap/focus-trap#createoptions `initialFocus`.

Just going off the existing content in the "tab fencing" example modal. That may just be out of date demo - [we already have `setFocus` available](https://developers.arcgis.com/calcite-design-system/components/modal/#component-api-methods-setFocus)

driskull:
From what I understand, it's ideal to focus on the first focusable element when a component's `focus/setFocus` method is called.

In most cases, this will be the close button but if that is disabled it would go to the next focusable element which could be some modal content or if no focusable modal content then any footer focusable content.

If we're following the rules, we shouldn't prioritize modal "content" over whatever the first focusable element is.

jcfranco:
@driskull We should also confirm if this impacts the a11y/design of things. IMO, for closable dialogs with interactive content, as a keyboard user I would expect to be able to interact with the content without an extra keypress to move focus. Note that I'm not vouching for `setFocus` options, but we'll need to find another approach if focusing the close button in all cases isn't ideal and we don't want to provide focus IDs and keep the same design. cc @ashetland @SkyeSeitz @geospatialem 

FWIW, @benelan was looking into this topic for https://github.com/Esri/calcite-components/issues/5147.

> If we're following the rules, we shouldn't prioritize modal "content" over whatever the first focusable element is.

Which rules are you referring to here?

ashetland:
> @driskull We should also confirm if this impacts the a11y/design of things. IMO, for closable dialogs with interactive content, as a keyboard user I would expect to be able to interact with the content without an extra keypress to move focus. Note that I'm not vouching for `setFocus` options, but we'll need to find another approach if focusing the close button in all cases isn't ideal and we don't want to provide focus IDs and keep the same design. cc @ashetland @SkyeSeitz @geospatialem
> 
> FWIW, @benelan was looking into this topic for #5147.
> 
> > If we're following the rules, we shouldn't prioritize modal "content" over whatever the first focusable element is.
> 
> Which rules are you referring to here?

I agree with @jcfranco here regarding interactive content. I like to think of this in terms of primary action. When the dialog appears, what's my primary action? If there is interactive content, my primary action would be to interact with that content before I close the dialog.

driskull:
> Which rules are you referring to here?

W3C guidelines for modal, but they are all pretty similar regarding focus.

https://www.w3.org/WAI/ARIA/apg/example-index/dialog-modal/dialog

> Note that I'm not vouching for setFocus options, but we'll need to find another approach if focusing the close button in all cases isn't ideal and we don't want to provide focus IDs and keep the same design

I agree its not ideal in all cases, but in those cases the user is slotting the content and the footer buttons so they can always focus on one of those once the modal is opened. I don't think we need `setFocus` options because of that. I don't think there are any modal shadowRoot elements that need to specifically be focused.

> I agree with @jcfranco here regarding interactive content. I like to think of this in terms of primary action. When the dialog appears, what's my primary action? If there is interactive content, my primary action would be to interact with that content before I close the dialog.

For these cases, we wouldn't be able to use any setFocus options to determine a primary action anyway. The author of the modal would have to focus on the primary action once the modal is opened. Which they can do because they are slotting the primary actions and those actions are in light DOM.

driskull:
Just to be clear, the only thing changing is we focus the `close-button` by default now because that's what the first focusable element in the component is.

driskull:
If we want to provide an option in the future for which element to focus by default, we could do that with `setFocus` by passing in the element or some selector.

https://github.com/focus-trap/focus-trap#what-it-does

> When a focus trap is activated, this is what should happen:

> Some element within the focus trap receives focus. By default, this will be the first element in the focus trap's tab order (as determined by [tabbable](https://github.com/focus-trap/tabbable)). Alternately, you can specify an element that should receive this initial focus.

driskull:
`delegatesFocus` and `focus-trap` are two different topics. 

- delegatesFocus: when a non-focusable part of the shadow DOM is clicked, the first focusable part is given focus, and the shadow host is given any available :focus styling.
- focusTrap: Trap focus within a DOM node.

benelan:
Ah, thanks for clarifying. I thought focus-trap was being suggested as an alternative.

driskull:
> Ah, thanks for clarifying. I thought focus-trap was being suggested as an alternative.

Nope! but ideally we should be consistent with how `delegatesFocus` works (by focusing first focusable element). Which `focus-trap` does.

geospatialem:
This looks great, @driskull! 🌟 Concur with your thoughts on the `:focus` placement for the close button, in particular around a11y.

If folks think the ordering seems odd with the close button being the first focusable element we could consider moving the close button to the footer by default. It would be a new pattern for the component so we'd need some design buy in for it, but perhaps an option if folks would rather see focusable elements from the content in focus first.

benelan:
> If folks think the ordering seems odd with the close button being the first focusable element we could consider moving the close button to the footer by default.

I'm not design, so take this with a grain of salt, but I don't think we should do this. If we do, we would probably need to change the "X" to "Close". I don't think I've ever seen a close "X" button in the footer, they are also in the top right. Imo changing that would cause more confusion/annoyance for users than whatever the first focusable element is.

macandcheese:
Many Modal's won't have footers either, so the close button as currently situated makes the most sense IMO. I think the focus on the close button should be expected as we've done that for awhile.

driskull:
>  Many Modals won't have footers either, so the close button as currently situated makes the most sense IMO. I think the focus on the close button should be expected as we've done that for awhile.

Correct. We've gotten feedback from a11y team that focusing close button on panel first is correct. This should align with that.

jcfranco:
> I think the focus on the close button should be expected as we've done that for awhile.
> Correct. We've gotten feedback from a11y team that focusing close button on panel first is correct. This should align with that.

Fair enough. As long as this is design/a11y-approved and we are consistent.

Just curious, but is there an a11y example out in the wild that has a close button similar to where we place ours?

> If we want to provide an option in the future for which element to focus by default, we could do that with setFocus by passing in the element or some selector.
https://github.com/focus-trap/focus-trap#what-it-does

I like this. We can revisit supporting this later based on user feedback.

src/components/modal/modal.tsx
```diff
@@ -355,11 +346,13 @@ export class Modal implements ConditionalSlotComponent, OpenCloseComponent {
    */
   @Method()
   async setFocus(focusId?: "close-button"): Promise<void> {
-    const closeButton = this.closeButtonEl;
+    const { closeButtonEl } = this;
 
-    return focusElement(
-      focusId === "close-button" ? closeButton : getFocusableElements(this.el)[0] || closeButton
-    );
+    if (closeButtonEl && focusId === "close-button") {
```
jcfranco, line 351: Since `close-button` is the only supported ID, I think you could just check for `closeButtonEl` being present or not. Passing any other ID would technically be invalid, right?

src/tests/commonTests.ts
```diff
@@ -196,6 +196,11 @@ interface FocusableOptions {
    * selector used to assert the focused shadow DOM element
    */
   shadowFocusTargetSelector?: string;
+
+  /**
+   * Used to delay setFocus call in ms.
+   */
+  delay?: number;
```
jcfranco, line 203: If this is for the open/close transition, I'd suggest using the `skipAnimation` helper instead of introducing a delay argument. If it's something specific to `focus-trap` behavior, can you add more info about this to the documentation?

src/utils/focusTrapComponent.ts
```diff
@@ -0,0 +1,72 @@
...
+  if (!focusTrapEl) {
+    return;
+  }
+
+  if (focusTrapEl.tabIndex == null) {
+    focusTrapEl.tabIndex = -1;
+  }
+
+  const focusTrapOptions: FocusTrapOptions = {
+    allowOutsideClick: true,
```
jcfranco, line 38: Do we need to set the defaults on `delayInitialFocus`, `clickOutsideDeactivates`, `returnFocusOnDeactivate`?

src/utils/focusTrapComponent.ts
```diff
@@ -0,0 +1,72 @@
...
+    setReturnFocus: (el) => {
+      focusElement(el as FocusableElement);
+      return false;
+    },
+    tabbableOptions: {
+      getShadowRoot: true
+    }
+  };
+
+  component.focusTrap = createFocusTrap(focusTrapEl, focusTrapOptions);
```
jcfranco, line 53: Could you add a spec test for these helpers?

jcfranco (approved):
Awesome stuff!

src/components/modal/modal.tsx
driskull: Correct. I think its nicer to check and fallback incase the user entered an invalid ID>

src/utils/focusTrapComponent.ts
driskull: Nope. i'll remove :)

src/tests/commonTests.ts
driskull: @jcfranco the problem is the `delayInitialFocus` is set to true.

I tried `delayInitialFocus: !Build.isTesting,` but the problem is that its not returning true.

```
{"testing":false,"browser":true,"dev":false,"server":false} 
    Location: http://localhost:3333/build/p-371d7086.entry.js:15:2306
```

src/utils/focusTrapComponent.ts
driskull: done

Issue #5663, opened 2022-10-28
Dropdown - max-items is not accurate in FF and Safari


jmanke:
### Actual Behavior

The `max-items` prop on the `<calcite-dropdown>` is not being reflected accurately in FF and Safari.

| Chrome  | FF | Safari |
| ------------- | ------------- | ------------- | 
| ![Screen Shot 2022-10-28 at 11 42 05 AM](https://user-images.githubusercontent.com/19629809/198709621-abcc9914-9a84-44a9-ace4-d29ca36eaacb.png) | <img width="340" alt="Screen Shot 2022-10-28 at 11 43 40 AM" src="https://user-images.githubusercontent.com/19629809/198709665-b7a19b88-a25f-4865-858c-5f6e767c8f39.png"> | <img width="340" alt="Screen Shot 2022-10-28 at 11 43 51 AM" src="https://user-images.githubusercontent.com/19629809/198709679-134bbc63-4724-4ae3-a990-db9b9a6b29cd.png"> |

### Expected Behavior

All browsers should behave the same when the `max-items` prop is set on the `calcite-dropdown`. In this case, they should match what chrome looks like.

### Reproduction Sample

https://codepen.io/jmanke/pen/NWzqLKK?editors=1000

### Reproduction Steps

1. Open [codepen](https://codepen.io/jmanke/pen/NWzqLKK?editors=1000)
2. Open dropdown in FF or Safari
3. Notice `max-items` is not behaving correctly

### Reproduction Version

beta.97

### Regression?

beta.96

### Esri team

ArcGIS Dashboard

jcfranco:
Going to nab this for this milestone as it's a regression from https://github.com/Esri/calcite-components/issues/5399. cc @geospatialem @benelan @brittneyzwolfer

geospatialem:
Verified on `next.631`

Firefox:
![image](https://user-images.githubusercontent.com/5023024/202225360-dc782564-9d2f-4f2d-968e-9df208e974c1.png)

Safari:
![image 2](https://user-images.githubusercontent.com/5023024/202226669-3b3414a0-2afa-4fa6-80a4-c89cbf9bd5af.png)

PR #5677, opened 2022-11-01
fix(dropdown): fix calculation of max-items for non-Chromium browsers.

jcfranco:
**Related Issue:** #5663 

## Summary

This fixes a regression caused by #5399 where the floating-ui element is placed offscreen and its styles change the offsetParent in non-Chromium browsers.

### DIFF
--- a/src/components/dropdown-group/dropdown-group.scss
+++ b/src/components/dropdown-group/dropdown-group.scss
@@ -1,5 +1,6 @@
 :host {
-  @apply block;
+  // we make the host relative, so items can consistently compute their offsetTop based on the group
+  @apply block relative;
 }
 
 .container {


driskull (approved):
Nice! A one liner!

Guess we need to start getting screenshots on other browsers :)

jcfranco:
> Guess we need to start getting screenshots on other browsers :)

Yeah, we can reconsider expanding our coverage in the near future. Maybe for starters we can add screenshots for non-Chrome fixes? cc @benelan

eriklharper:
> Maybe for starters we can add screenshots for non-Chrome fixes? cc @benelan

+1

benelan:
Yeah Chromatic supports Firefox, I'll look to see if there is a way to specify which browser to run on per snapshot

eriklharper:
> Yeah Chromatic supports Firefox, I'll look to see if there is a way to specify which browser to run on per snapshot

That would be very cool if that's possible.

benelan:
Unfortunately it's not possible. We would have to run all snapshots on both chrome and firefox in order to add coverage for the non-Chrome fixes. Could be a candiate for our first official enhancement request

eriklharper:
> Unfortunately it's not possible. We would have to run all snapshots on both chrome and firefox in order to add coverage for the non-Chrome fixes. Could be a candiate for our first official enhancement request

Had a feeling that was the case.  Certainly would make for a great experience though if we could add browser-specific snapshots since we do get reports on specific UI issues that would be excellent to capture with just a single or small set of very specific cases that are browser-specific.  We should definitely request this feature.

Issue #5489, opened 2022-10-17
calcite-inline-editable: end with ellipses if string is too long


AdelheidF:
### Actual Behavior

if the string is too long to fit it should end with ellipses

### Expected Behavior

string ends with ellipses

### Reproduction Sample

https://codepen.io/afreitag/pen/ExLJpRa

### Reproduction Steps

open codepen and see string is just cut off.

### Reproduction Version

beta.95

### Esri team

ArcGIS Map Viewer

geospatialem:
@AdelheidF The request to add an ellipses is out of scope for the component at this time. As such, will be closing the request.

At this time it would be on the developer to implement an ellipse if interested in doing so.

ashetland:
Already built. 🥳
<img width="276" alt="CleanShot 2022-11-16 at 09 31 30@2x" src="https://user-images.githubusercontent.com/108549080/202251887-757f9c45-a9e6-4a44-896a-3a98409a36b1.png">

PR #5679, opened 2022-11-02
fix(inline-editable): Add text-ellipsis when not editing

driskull:
**Related Issue:** #5489

## Summary

fix(inline-editable): Add text-ellipsis when not editing. (#5489)

## Example

image


driskull:
@geospatialem do you know why the issue above was closed?

geospatialem:
> @geospatialem do you know why the issue above was closed?

@driskull The sprint discussion mentioned this as a design that was out of scope for the component. Maybe good to get some insights from some UI and/or design folks? cc @macandcheese

driskull:
Yeah. It seems to me its a simple fix that makes it look a lot better.

macandcheese:
This fix makes sense to me, - the un-truncated string is available for users in the editing state.

eriklharper (approved):
LGTM 👍

Issue #4333, opened 2022-03-30
calcite-pick-list needs an event specific to its filter to get filtered items


AdelheidF:
### Description

When using `calcite-pick-list` with `filterEnabled:true` the only way to get to the filtered items is by listening to `calciteFilterChange` and using `event.path[0].filteredItems`.

Before November 2021 we could get the filtered items via `calciteFilterChange` and `event.detail`.
See: https://github.com/Esri/calcite-components/pull/3449

cc @dhrumil83

### Acceptance Criteria

There should be an event on `calcite-pick-list` specific to the filter to get the filtered items.

### Which Component

calcite-pick-list

### Example Use Case

```
<calcite-pick-list filter-enabled>
  <calcite-pick-list-item label="Chocolate" value="chocolate">
    <calcite-action slot="actions-end" icon="ellipsis-circle"></calcite-action>
  </calcite-pick-list-item>
  <calcite-pick-list-item label="Vanilla" description="Oldie but goodie" value="vanilla">
    <calcite-action slot="actions-end" icon="ellipsis-circle"></calcite-action>
  </calcite-pick-list-item>
  <calcite-pick-list-item label="Strawberry" description="no metadata on this one" value="strawberry">
    <calcite-action slot="actions-end" icon="ellipsis-circle"></calcite-action>
  </calcite-pick-list-item>
</calcite-pick-list>
```

jcfranco:
[Implementor's note] Both `calcite-value-list` and `calcite-pick-list` need to be updated. 

#### Proposal

```ts
interface List {
  calciteListFilter: EventEmitter<void>;
  filterText: string;
}
```

#### Question

* Should filtering affect the selected values (not the items)? Both components have methods to get the selected items, but we may need an additional convenience prop for filtered ones if filtering is independent of selection.

@driskull @eriklharper @Elijbet WDYT about ☝️? Any name suggestions?

driskull:
It depends on the purpose of the filter. If the filter is just to help the end user narrow results to select, it might not need to affect the value.

Maybe we shouldn't filter out selected values if we always want to show them?

driskull:
The way the combobox gets away with this is that it is always showing the selected items as chips.

AdelheidF:
Use case in table component.

User can filter a certain subset of fields then either hit select all or deselect all on that subset of fields. So it needs to know the subset of fields to make the adjustments in the table.

[...]

PR #5681, opened 2022-11-02
feat(pick-list, value-list): Add calciteListFilter event, filteredItems prop, filterText prop and filteredData prop.

driskull:
**Related Issue:** #4333

## Summary

feat(pick-list, value-list): Add calciteListFilter event, filteredItems prop, filterText prop and filteredData prop. (#4333)

### DIFF
--- a/src/components/filter/filter.tsx
+++ b/src/components/filter/filter.tsx
@@ -120,7 +120,7 @@ export class Filter implements InteractiveComponent {
   //--------------------------------------------------------------------------
 
   componentWillLoad(): void {
-    this.filter(this.value);
+    this.updateFiltered(filter(this.items, this.value));
   }
 
   // --------------------------------------------------------------------------
--- a/src/components/filter/filter.tsx
+++ b/src/components/filter/filter.tsx
@@ -11,7 +11,7 @@ import {
   Watch
 } from "@stencil/core";
 import { debounce } from "lodash-es";
-import { CSS, DEBOUNCE_TIMEOUT, ICONS, TEXT } from "./resources";
+import { CalciteFilterChangeDetail, CSS, DEBOUNCE_TIMEOUT, ICONS, TEXT } from "./resources";
 import { Scale } from "../interfaces";
 import { focusElement } from "../../utils/dom";
 import { InteractiveComponent, updateHostInteraction } from "../../utils/interactive";
@@ -111,7 +111,7 @@ export class Filter implements InteractiveComponent {
   /**
    * This event fires when the filter text changes.
    */
-  @Event({ cancelable: false }) calciteFilterChange: EventEmitter<void>;
+  @Event({ cancelable: false }) calciteFilterChange: EventEmitter<CalciteFilterChangeDetail>;
 
   //--------------------------------------------------------------------------
   //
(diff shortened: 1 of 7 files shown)


src/components/pick-list/shared-list-logic.ts
```diff
@@ -11,6 +11,11 @@ type List<T> = T extends PickList ? PickList : ValueList;
 
 export type ListFocusId = "filter";
 
+export interface CalciteListFilterDetail {
+  calciteListFilter: ItemData;
```
jcfranco, line 15: Can we introduce a `filterText` (or similar) prop on the lists to skip the event payload?

jcfranco (changes requested):
Looking good, I just had a suggestion regarding the event payload.

src/components/pick-list/shared-list-logic.ts
driskull: @jcfranco what about the `ItemData`? Right now we just have `@State() dataForFilter: ItemData = [];`?

I like the `filterText` idea because that allows the users to set the filter as well.

driskull: Would we want `filterData` prop as well? readonly.

src/components/filter/filter.tsx
```diff
@@ -120,7 +120,7 @@ export class Filter implements InteractiveComponent {
   //--------------------------------------------------------------------------
 
   componentWillLoad(): void {
-    this.filter(this.value);
+    this.updateFiltered(filter(this.items, this.value));
```
driskull, line 123: This is so that creating the filter initially is not debounced. Allows the `filteredItems` to be defined initially.

src/components/pick-list/shared-list-logic.ts
driskull: Updated! Let me know how it looks now @jcfranco

driskull:
Addressed review feedback. Can I get another review?

src/components/pick-list/pick-list.tsx
```diff
@@ -68,6 +69,13 @@ export class PickList<
    */
   @Prop({ mutable: true }) filteredItems: HTMLCalcitePickListItemElement[] = [];
 
+  /**
+   * **read-only** The currently filtered items
+   *
+   * @readonly
```
jcfranco, line 75: @macandcheese Do you think `@readonly` could be picked up by the doc process? It'd prevent us from explicitly adding it to the JSDoc.

jcfranco (approved):
🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦
🍦🧊🍦🍦🍦🧊🍦🧊🧊🧊🍦🍦🧊🧊🧊🍦🧊🧊🧊🧊🍦🧊🍦
🍦🧊🧊🍦🍦🧊🍦🍦🧊🍦🍦🧊🍦🍦🍦🍦🧊🍦🍦🍦🍦🧊🍦
🍦🧊🍦🧊🍦🧊🍦🍦🧊🍦🍦🧊🍦🍦🍦🍦🧊🧊🧊🍦🍦🧊🍦
🍦🧊🍦🍦🧊🧊🍦🍦🧊🍦🍦🧊🍦🍦🍦🍦🧊🍦🍦🍦🍦🍦🍦
🍦🧊🍦🍦🍦🧊🍦🧊🧊🧊🍦🍦🧊🧊🧊🍦🧊🧊🧊🧊🍦🧊🍦
🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦🍦

Issue #5680, opened 2022-11-02
Bug: Tooltip arrow issue when the container has text-align: right set


rpanichakit9541:
### Actual Behavior

When the container element has the `text-align: right` style set, the tooltip arrow looks broken.

### Expected Behavior

The tooltip arrow should render correctly regardless of the parent container text-align style

### Reproduction Sample

https://codepen.io/rpanichakit/pen/JjZXOLo?editors=100

### Reproduction Steps

Hover over the link to see the arrow issue.

### Reproduction Version

beta.95

### Relevant Info

Windows 10 Chrome Version 107.0.5304.87 (Official Build) (64-bit)

### Impact

Not currently blocking, was able to override the container to have `text-align: start` to work around it.

### Esri team

ArcGIS Online

geospatialem:
Verified on `next.631` in Firefox and Chrome.

![image](https://user-images.githubusercontent.com/5023024/202229597-eb4c96ed-04f2-440d-8b41-f4c2f4625703.png)

PR #5682, opened 2022-11-02
fix(tooltip, popover): Fix pointer when text-align is right

driskull:
**Related Issue:** #5680

## Summary

fix(tooltip, popover): Fix pointer when text-align is right. #5680


jcfranco (approved):
👍

Issue #5690, opened 2022-11-04
value-list: screen reader only instructions appears with keyboard interaction when dragEnabled


anveshmekala:
### Actual Behavior

When the user interacts with the `drag-enabled` value-list component, screen reader helper text is displayed.

### Expected Behavior

Screen reader only instructions should be hidden.

### Reproduction Sample

https://codepen.io/ANVESHMEKALA/pen/zYaBeKZ

### Reproduction Steps

1. In the codepen, try to tab into the draghande of any value-list item.
2.  Press space key.
3. observe a text being displayed at the top of the list with instructions.

### Reproduction Version

0.96

### Regression?

beta.86

### Esri team

N/A

jcfranco:
Express-triaged for this milestone. cc @benelan @geospatialem @brittneyzwolfer

jcfranco:
Verified with the latest release: https://codepen.io/jcfranco/pen/ExRZqbG?editors=100

PR #5691, opened 2022-11-04
fix(value-list): no longer display screen reader only text when drag-enabled

anveshmekala:
**Related Issue:** #5690 

## Summary

This PR will hide the screen reader only text when drag-enabled in `value-list` component.


driskull (approved):
👍 Nice

Issue #5693, opened 2022-11-04
calcite-value-list + calcite-dropdown position issue


jcfranco:
### Actual Behavior

Dragging the last element from a value list will cause incorrect positioning from slotted dropdown.

cc @rslibed

### Expected Behavior

Dropdown menu positioning takes into consideration latest reference element placement.

### Reproduction Sample

https://codepen.io/rslibed/pen/ZEROGjX

### Reproduction Steps

1. Drag bottom 'calcite-value-list-item' to top
2. Click on '...' action on 'Bottom (Drag to top)' value list item
3. Observe

### Reproduction Version

beta.97

### Regression?

beta.87

### Impact

Affects layout and can prevent interaction with the dropdown menu.

### Esri team

ArcGIS Instant Apps

driskull:
Somewhat related to #5694 as well.

geospatialem:
Verified on `next.631`.

![image](https://user-images.githubusercontent.com/5023024/202232433-289b4806-f35c-430d-a75a-b1f029e4b5dd.png)

PR #5695, opened 2022-11-07
fix: avoid size changes before and after positioning floating-ui elements

jcfranco:
**Related Issue:** #5693 

## Summary

This modifies styles to avoid size changes before and after positioning:

> Initial layout
To ensure positioning works smoothly, the dimensions of the floating element should not change before and after being positioned.

☝️ from https://floating-ui.com/docs/computePosition#initial-layout).


driskull (approved):
👍 Nice.

driskull:
Yeah this seems like it would fix a lot of the issues. I don't think these sizes were there before the other changes.

driskull:
I think the failing test can be updated 👍

Issue #5694, opened 2022-11-04
floating-ui offscreen placement causes drag-and-drop issues in Chrome and Safari


jcfranco:
### Actual Behavior

Chrome's ghost image while dragging appears in an unexpected place and in Safari items cannot be dragged at all.

cc @rslibed

### Expected Behavior

Drag and drop is consistent across browsers.

### Reproduction Sample

https://codepen.io/rslibed/pen/ZEROGjX

### Chrome

1. Drag any 'calcite-value-list-item''s handle and move
2. Notice the ghost image is offset and becomes obscured

### Safari

1. Drag any 'calcite-value-list-item''s handle and try to move
2. Notice that the item cannot be dragged at all

### Reproduction Version

beta.97

### Regression?

beta.97

### Esri team

ArcGIS Instant Apps

geospatialem:
Verified in `next.631`.

![verify-drag-handle](https://user-images.githubusercontent.com/5023024/202233679-f88911eb-e8ab-4e43-9b56-6b48b8205746.gif)

PR #5696, opened 2022-11-07
fix: tweak position reset logic to avoid layout interference

jcfranco:
**Related Issue:** #5694 

## Summary

This resets absolutely-positioned floating-ui elements according to the doc:

> Initial layout
...
These properties prevent the floating element from resizing when it overflows a container, removing layout interference that can cause incorrect measurements.

☝️ from https://floating-ui.com/docs/computePosition#initial-layout)


driskull (approved):
awesome sauce. ! 👍

PR #5703, opened 2022-11-08
ci(actions): remove deprecated set-output workflow commands

benelan:
**Related Issue:** #

## Summary

Gets rid of `set-output` deprecation notices

image

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


driskull (approved):
👍

Issue #5701, opened 2022-11-08
Activating drag button with Space on Value List makes the page scroll down unexpectedly


cont-kolomeets:
### Actual Behavior

Activating drag button with the keyboard Space key on Value List makes the page scroll down unexpectedly.

![image](https://user-images.githubusercontent.com/5318527/200470940-0a14c87e-c1bb-4bcd-a6c2-2fa4749fcd0a.png)

### Expected Behavior

The default scroll behavior on Space should be prevented.

### Reproduction Sample

https://codepen.io/cont-kolomeets/pen/qBKqVaK?editors=100

### Reproduction Steps

(1) Create a long value list to have a scroll bar on the right.
(2) Using Shift+Tab navigate to the drag button.
(3) Hit Space. The drag is activated, but the default Space key behavior is not prevented and the page scrolls way down, moving the focused element up and out of sight.

### Reproduction Version

beta.95

### Esri team

ArcGIS Business/Community Analyst

geospatialem:
Verified in `next.631`

![verify-keyboard-value-list](https://user-images.githubusercontent.com/5023024/202235802-a1b56523-9919-4444-b97e-c3cb221712ff.gif)

PR #5709, opened 2022-11-08
fix(value-list-item): Prevent scrolling when space is pressed on drag button

driskull:
**Related Issue:** #5701

## Summary

fix(value-list-item): Prevent scrolling when space is pressed on drag button. #5701


jcfranco (approved):
Nice!

PR #5711, opened 2022-11-08
ci(screener): fix yaml syntax

benelan:
**Related Issue:** #

## Summary
Fix indent and use GITHUB_OUTPUT instead of GITHUB_ENV


driskull (approved):
👍

PR #5712, opened 2022-11-08
ci(e2e): use step output in github action

benelan:
**Related Issue:** #5711

## Summary
Forgot to change the e2e action as well


driskull (approved):
👍

PR #5714, opened 2022-11-08
ci(next): fix error due to updating action runner

benelan:
**Related Issue:** #

## Summary

I just updated the GH Action runners from node12 to node16 and I've been slowly chipping away at some problems as they pop up. For this round, we got an error during a next deployment:
https://github.com/Esri/calcite-components/actions/runs/3422971878/jobs/5701302669#step:5:8861

The husky prepush that caused the error is checking to see if the shell is interactive. I ran into this error with GH Actions a while ago in #3782

Not sure why the error didn't show up here until now. I changed the hook's tty check to run in a subshell so it won't throw errors in the main script. We only care about the exit code for the conditional. I Also cleaned up the hook to use quotes so we don't run into any parameter expansion issues down the road.

Additionally, I moved the `npm publish` to go before the `push`. The deploy action gets cancelled if another commit comes in, so that deployable commits are batched instead of having a bunch of releases back to back. So having the push before the publish could potentially cancel deployment.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,14 +1,14 @@
 #!/bin/bash
 # from https://riptutorial.com/git/example/16164/pre-push
 
-protected_branch='master'
+protected_branch="master"
 current_branch=$(git rev-parse --abbrev-ref HEAD)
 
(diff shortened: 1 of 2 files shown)


driskull (approved):
💯

PR #5724, opened 2022-11-09
ci(next): add back default npm registry url

benelan:
**Related Issue:** #5719

## Summary

I just remembered that gitignored files aren't considered untracked, which means `git clean` won't remove them. So my last PR (linked above) probably won't fix the issue. I'm starting to run out of ideas, but maybe it is due to removing the default NPM registry url? Not sure why we would be required to specify the default registry, but serves me right for trying to tidy up, I guess.

If this isn't it, I'll revert back to the node12 runner until I have more time to troubleshoot.


driskull (approved):
Just keep trying :)

PR #5725, opened 2022-11-09
feat(popover): Add focus-trap to popover and disableFocusTrap property.

driskull:
**Related Issue:** #2133

## Summary

feat(popover): Add focus-trap to popover and disableFocusTrap property.. #2133


jcfranco (approved):
Awesome! 

Before merging, can you add an E2E test for popover + focus trap behavior?

Issue #5661, opened 2022-10-28
[a11y] Tabs component: keyboarding events for Home and End keys


stefanobettinelli:
### Summary

Using tab button to cycle through tab (first to last) is not working as expected:
`Move focus to next tab. If the currently focused tab is the last, the focus will leave the component`

**Instead it skips the tabs and jumps to the next element**

But it works correctly with the arrow keys, with a small difference from the tab behaviour:
`Move focus to previous tab. if the current focus is on last tab, the focus will cycle to the first tabs`

Can also be reproduced from the sample in the official docs
https://developers.arcgis.com/calcite-design-system/components/tabs/

### Actual Behavior

Using tab button to cycle through tab (first to last) is not working as expected:
`Move focus to next tab. If the currently focused tab is the last, the focus will leave the component`

**Instead it skips the tabs and jumps to the next element**

### Expected Behavior

It would be expected that pressing tab would cycle through the tabs until the last one and from the last one to the next element in the DOM (that can ben focused)

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/tabs/

### Reproduction Steps

Use the sample from the official docs
https://developers.arcgis.com/calcite-design-system/components/tabs/

Focus the first tab and press the tab key from the keyboard

### Reproduction Version

v1.0.0-beta.95

### Relevant Info

Reproduced on a Mac on Chrome, Safari and Firefox

### Esri team

ArcGIS Scene Viewer

macandcheese:
Potentially related prior work / discussion: https://github.com/Esri/calcite-components/issues/2124. @geospatialem / @benelan we may want to look at putting the a11y descriptions in js-doc output if we can to have one source of truth for those descriptions.

geospatialem:
Good catch and agreed on future a11y-related doc updates, @macandcheese.

We recently did update the keyboard support functionality for `tabs`, however it does look like we're missing the use of the `Home` and `End` keys.

This issue could cover a few updates:
- [ ] 1. Updating the `tabs` component to support the `Home` and `End` keys, where the:
    - `Home` key moves focus to the first `tab-title`
    - `End` key moves focus to the last `tab-title`
- [ ] 2.  Update the a11y navigation on the [docs page](https://developers.arcgis.com/calcite-design-system/components/tabs/) for:
    - `Tab` and `Tab`/`Shift` to indicate focus moves to the `tab` then to the next element in the page.
[...]

PR #5727, opened 2022-11-09
feat(tabs): Add keyboarding events for Home and End keys

driskull:
**Related Issue:** #5661

## Summary

feat(tabs): Add keyboarding events for Home and End keys #5661

### DIFF
--- a/src/components/tab-nav/tab-nav.tsx
+++ b/src/components/tab-nav/tab-nav.tsx
@@ -223,6 +223,26 @@ export class TabNav {
     event.preventDefault();
   }
 
+  @Listen("calciteInternalTabsFocusFirst")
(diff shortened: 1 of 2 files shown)


src/components/tab-nav/tab-nav.tsx
```diff
@@ -223,6 +223,26 @@ export class TabNav {
     event.preventDefault();
   }
 
+  @Listen("calciteInternalTabsFocusFirst")
+  focusFirstTabHandler(event: CustomEvent): void {
```
jcfranco, line 227: Can you extract the common logic between the event handlers into a separate method?

jcfranco (approved):
Code LGTM, but this needs a test before merging.

Also, shouldn't the PR title be more about adding support for Home/End vs the events themselves?

Issue #5099, opened 2022-08-04
calcite-panel: Set the default heading to be a `div` to accommodate more audiences


kevindoshier:
### Summary

The mapViewer has been undergoing accessibility review, and one of the issues that came up was the headers within calcite-panels which we use all over the app.  The comment from those reviewers is

`[Issue]
There are dialogs present that have headings that do not match their visual level. The first heading starts with <h3> instead of <h2>.`

Could we consider setting the default for `headingLevel` to be 2 to match this review, otherwise we would need to scour everywhere in the app where we use calcite-panels and manually make this change

### Actual Behavior

The `headingLevel` appears to default to 3

### Expected Behavior

Default the `headingLevel` to 2 to satisfy the accessibility review requirements

### Reproduction Sample

https://codepen.io/kevindoshier/pen/dymmBwy

### Reproduction Steps

1. Click repro link above

### Reproduction Version

beta.86

macandcheese:
It does seem a bit weird that we have a default set at all. I almost wonder if it should be rendering a plain div / span by default, and only render h1-h6 when a user sets this property. 

Ultimately, even if in Map Viewer defaulting to h2 makes sense, it may not be the case for another app, and they would need to adjust that similarly.

macandcheese:
Correct, I think we said the default internally rendered element would be a div, or span (defer to @geospatialem), and it would only render h1-h6 when requested via prop. No visual changes for any element.

geospatialem:
> @geospatialem @macandcheese I noticed we didn't update this issue after discussion. 😅 I believe we agreed to not set a default as we can't guarantee it'll be correct. Can you confirm?

Exactly what @macandcheese said (he beat me to it!) 🐇 

@jcfranco Good memory, and thanks for catching this one! 

Yes, we'd like to have the default be set to a `span`, and where a heading is defined change it to the appropriate header class (e.g., `heading-level="2"` would correspond to a `h2`).

driskull:
A `div` would be better since its a headings are block elements

geospatialem:
Verified on `next.634` for:
`block`, `date-picker`, `list-item-group`, `panel`, `pick-list-group`, `popover`, `tip`, and `tip-manager`. (Also `flow-item` and `input-date-picker`. 👍🏻 

![image](https://user-images.githubusercontent.com/5023024/202580271-b6100ea5-0f7f-4cf0-9672-275a14a26b77.png)

PR #5728, opened 2022-11-09
fix(block, date-picker, list-item-group, panel, pick-list-group, popover, tip, tip-manager)!: Set default internal heading to a div.

driskull:
**Related Issue:** #5099

## Summary

fix(block, date-picker, list-item-group, panel, pick-list-group, popover, tip, tip-manager)!: Set default internal heading to a div.

- Set default heading to be a div to accommodate more audiences
- Applies to block, date-picker, list-item-group, panel, pick-list-group, popover, tip, and tip-manager


t9nmanifest.txt
```diff
@@ -29,4 +29,4 @@ src\components\shell-panel\assets\shell-panel\t9n
 src\components\time-picker\assets\time-picker\t9n
 src\components\tip-manager\assets\tip-manager\t9n
 src\components\tip\assets\tip\t9n
-src\components\value-list\assets\value-list\t9n
```
jcfranco, line 32: Can you revert this unrelated change?

jcfranco:
Devs will need to update their code to preserve heading levels, so this would be a breaking change, right?

driskull:
> Devs will need to update their code to preserve heading levels, so this would be a breaking change, right?

Yes, I guess that would be a breaking change if they were not setting any value.

driskull:
Updated to breaking change.

driskull:
Created issue: https://github.com/Esri/calcite-components/issues/5768

PR #5730, opened 2022-11-10
ci(output targets): move patches to postbuild to include them in CCR

benelan:
**Related Issue:** https://github.com/ArcGIS/calcite-components-output-targets/pull/10

## Summary

When releasing CCR, `stencil build` must be the last command  in CC's `build` script because [CCR appends the `--config` flag](https://github.com/ArcGIS/calcite-components-output-targets/blob/ae776a58efc3212802cd7a79192dcf09964f92b8/packages/calcite-components-output-target-generator/package.json#L15) to use a different config that generates the react output target.

I had been removing the patch because I thought something it changed in the CC build was breaking the CCR build. Moving the patch to `postbuild` will allow the patches to be applied to CCR as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,8 @@
     "hydrate/"
   ],
   "scripts": {
-    "build": "npm run util:prep-build-reqs && stencil build --no-docs && npm run util:patch",
+    "build": "npm run util:prep-build-reqs && stencil build --no-docs",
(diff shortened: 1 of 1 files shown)


jcfranco:
> When releasing CCR

🧠💭👉https://www.youtube.com/watch?v=Aae_RHRptRg

PR #5737, opened 2022-11-10
ci(pr-bot): skip assignment for dependabot PRs

benelan:
**Related Issue:** #5704

## Summary
I replaced an unmaintained 3rd party action, which still uses node12, with a [`github-script`](https://github.com/actions/github-script) workflow, but forgot to skip assignment for dependabot PRs


driskull (approved):
👍

PR #5748, opened 2022-11-14
build(docs): generate docs-json for afd usage

benelan:
**Related Issue:** #5654

## Summary
Turns out adding `--no-docs` to the build prevents the `docs-json` output target from being created. The reason I added it was the annoying readmes being generated for each component. The `git restore` I added to `postbuild` resolves that, while still generating `docs-json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,8 +17,8 @@
     "hydrate/"
   ],
   "scripts": {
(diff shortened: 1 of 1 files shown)


jcfranco (approved):
🎉

Issue #5207, opened 2022-08-23
Input date picker with range ignores valueAsDate parameter


arjanvanzutphen:
### Actual Behavior

InputDatePicker with the range parameter ignores the valueAsDate parameter value and the input element stays empty.
The single inputDatePicker honours the valueAsDate parameter with a pre filled date value in the input element
![image](https://user-images.githubusercontent.com/25850072/186250375-5f3f09b2-e9a7-40e0-8410-c2b20fc5f6dc.png)

### Expected Behavior

The inputDatePicker element with the range parameters should contain the pre filled values when passing the [new Date(), new Date()] array in the valueAsDate parameter

### Reproduction Sample

https://codepen.io/arjan_esrinl/pen/JjLVKwV?editors=1001

### Reproduction Steps

1. Open the codepen: https://codepen.io/arjan_esrinl/pen/JjLVKwV?editors=1001
2. Notice the missing values in the second inputDatePicker

### Reproduction Version

1.0.0-beta.91

### Relevant Info

Windows 11
Chrome 104.0.5112.81 (Official Build) (64-bit)
Firefox 95.0 (64-bit)

driskull:
@eriklharper can we prioritize this one for the next release?

eriklharper:
> @eriklharper can we prioritize this one for the next release?

I have a fix ready to go, just writing some tests for it :)  PR should be ready EOW.

geospatialem:
Verified on `next.642`  ✔️ 

![image](https://user-images.githubusercontent.com/5023024/204327809-50581a35-0a39-4bba-b89a-e5c908909284.png)

PR #5758, opened 2022-11-15
fix(input-date-picker): display updated valueAsDate in the two range inputs

eriklharper:
**Related Issue:** #5207 

## Summary

This PR fixes an issue where the input boxes are completely empty when the `valueAsDate` property is directly changed with javascript when using the `range` prop.


driskull (approved):
Looks good @eriklharper 🎉 🌮

PR #5759, opened 2022-11-15
ci(pr-bot): don't label dependabot PRs

benelan:
**Related Issue:** #5756

## Summary
Dependabot labels itself and behaves differently than normal PRs, so we can just skip labelling it.

### DIFF
--- a/src/utils/locale.spec.ts
+++ b/src/utils/locale.spec.ts
@@ -1,5 +1,6 @@
 import {
   locales,
+  numberingSystems,
   numberStringFormatter,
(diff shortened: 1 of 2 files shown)


src/utils/locale.spec.ts
```diff
@@ -1,5 +1,6 @@
 import {
   locales,
+  numberingSystems,
```
driskull, line 3: Seems like this file shouldn't be changed right?

driskull (changes requested):

PR #5761, opened 2022-11-15
fix(calcite-loader, calcite-input-message)!: drop active in favor of hidden

Elijbet:
**Related Issue:** #3821

## Summary

`Loader` and `Input-Messag`e are small building blocks that an owning component should be able to display or hide.

This deprecates/removes the role of `active` prop in toggling visibility in favor of the global attribute `hidden`. Styles for hidden take precedence over `active`. 

The benefit is one less custom prop to manage, and it aligns with existing HTML conventions.

Removes `active` knob from stories and subs the ones set to `false` with static `hidden` for consistency, as we don't provide knobs for native global attributes.


src/components/input-number/input-number.stories.ts
```diff
@@ -83,9 +83,7 @@ export const withSlottedAction = (): string => html`
     >
       <calcite-button slot="action">${text("action button text", "Go")}</calcite-button>
     </calcite-input-number>
-    <calcite-input-message
-      ${boolean("input-message-active", false)}
-      status="${select("input message status", ["idle", "valid", "invalid"], "idle")}"
+    <calcite-input-message hidden status="${select("input message status", ["idle", "valid", "invalid"], "idle")}"
```
driskull, line 86: Same here, we need a way to toggle hidden in the story right?

src/components/input-text/input-text.stories.ts
```diff
@@ -71,9 +71,7 @@ export const withSlottedAction = (): string => html`
     >
       <calcite-button slot="action">${text("action button text", "Go")}</calcite-button>
     </calcite-input-text>
-    <calcite-input-message
-      ${boolean("input-message-active", false)}
-      status="${select("input message status", ["idle", "valid", "invalid"], "idle")}"
+    <calcite-input-message hidden status="${select("input message status", ["idle", "valid", "invalid"], "idle")}"
```
driskull, line 74: Same here with regards to toggling hidden.

driskull: Applies to all stories.

src/components/input/input.stories.ts
```diff
@@ -118,7 +116,7 @@ export const textarea = (): string => html`
     >
     </calcite-input>
     <calcite-input-message
-      ${boolean("input-message-active", false)}
+      ${boolean("hidden", true)}
```
driskull, line 119: nice :)

src/components/loader/readme.md
```diff
@@ -8,40 +8,39 @@ The `calcite-loader` component can act as a determinate or indeterminate loading
 
 ### Basic
 
-You can use the `active` attribute to toggle visibility:
+You can use the global attribute hidden to toggle visibility:
```
driskull, line 11: Do we need this doc? This could apply to any component. `hidden` is a global attribute that would hide any component. Just seems odd to have that doc here.

driskull (approved):
Looks good.

I think just stories need to be updated with regards to toggling hidden in some cases.

We also should decide if we need to document the global `hidden` attribute.

src/components/inline-editable/inline-editable.stories.ts
```diff
@@ -62,7 +62,7 @@ export const darkThemeRTL_TestOnly = (): string => html`
         </calcite-input>
       </calcite-inline-editable>
       <calcite-input-message
-        ${boolean("active", false, "InputMessage") && "active"}
+        hidden
```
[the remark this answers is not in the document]
Elijbet: I talked to Franco about how to address this. I was wondering for cases when we have a story (not `_TestOnly`) with a toggle `active` set to false, which isn't available anymore... do we just remove that and set it to static `hidden`, or does hidden need to be a toggle boolean? And we decided on static hidden for consistency, as we don't provide knobs for native global attributes. In the one case where I have input textarea story with a boolean hidden, I don't see the knob on the storybook. It seems to be ignored. The question is does the storybook even accept an HTML native attribute for a knob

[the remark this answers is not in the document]
Elijbet: @jcfranco  for feedback on the hidden toggle knob conundrum.

src/components/input-number/input-number.stories.ts
jcfranco: A few thoughts:

1. I'm not sure if we need the input message for the slotted action story/use case. This sort of applies to other types too.
2. I think it'd be more helpful to have a usage entry or simple example showing how input message can be toggled dynamically. Maybe this is a future validation story candidate.
3. Shouldn't any `withInputMessage` story always show it?

src/components/loader/readme.md
jcfranco: Agreed.

src/components/loader/loader.e2e.ts
```diff
@@ -3,15 +3,15 @@ import { renders, hidden } from "../../tests/commonTests";
 
 describe("calcite-loader", () => {
   it("renders", async () => {
-    await renders("calcite-loader", { display: "none", visible: false });
-    await renders(`<calcite-loader active></calcite-loader>`, { display: "flex", visible: true });
+    await renders(`<calcite-loader hidden></calcite-loader>`, { display: "none", visible: false });
```
jcfranco, line 6: This is no longer needed as it's covered by the `hidden` test below.

src/components/loader/loader.tsx
```diff
@@ -21,10 +21,12 @@ export class Loader {
   //
   //--------------------------------------------------------------------------
 
-  /** When `true`, the component is active. */
-  @Prop({ reflect: true }) active = false;
-
-  /** When `true`, displays smaller and appears to the left of the text. */
+  /**
+   * When `true`, the component is active.
```
jcfranco, line 25: Can you drop this doc for `active` and restore the doc for `inline`?

driskull (approved):
🎉 🌮

PR #5769, opened 2022-11-16
Revert "fix(block, date-picker, list-item-group, panel, pick-list-gro…

driskull:
…up, popover, tip, tip-manager)!: Set default internal heading to a div. (#5728)"

This reverts commit 38ca639010b8bd1d1fe32c9cf9b54dfc38cf9877.

**Related Issue:** #


driskull:
This PR is ready if we need to revert the breaking change.

Issue #5734, opened 2022-11-10
Add events to Tooltip to have more control over its content and behavior


cont-kolomeets:
### Description

In our code we need to show tooltip text depending on the current context. The trick we would like to use is to place text on the fly before the tooltip is shown. Unlike popover, tooltip doesn't emit any events so there is no way to know when the tooltip is open.

Of course, we could use popover for this case, but popover has different style and doesn't have the built-in hover-to-show behavior.

### Acceptance Criteria

Need `calcite-tooltip` to emit `calciteTooltipBeforeOpen`, `calciteTooltipOpen`, `calciteTooltipBeforeClose`, `calciteTooltipClose` events.

### Which Component

https://developers.arcgis.com/calcite-design-system/components/tooltip/

### Example Use Case

One possible example. Hover over the color picker to get more information about the selected color.
A workaround for now is to update the tooltip every time the color picker changes, but this is not very efficient to do this everywhere. It's better to build a tooltip on demand when needed.

![image](https://user-images.githubusercontent.com/5318527/201005368-a95eac0b-a363-445a-b21b-f56d3207caa0.png)

### Esri team

ArcGIS Business/Community Analyst

driskull:
> Need calcite-tooltip to emit calciteTooltipBeforeOpen, calciteTooltipOpen, calciteTooltipBeforeClose, calciteTooltipClose events.

Agreed. I think we can add these events.

> Of course, we could use popover for this case, but popover has different style and doesn't have the built-in hover-to-show behavior.

For a11y, its best to use tooltip in this instance :)

> A workaround for now is to update the tooltip every time the color picker changes, but this is not very efficient to do this everywhere.

I think this is actually better than modifying the content after the tooltip is open. The tooltip positions itself based on the content in many examples so this wouldn't be good if you have a large tooltip that would open in different positions depending on size of the content.

PR #5772, opened 2022-11-17
feat(tooltip): Add tooltip open, close, beforeOpen, and beforeClose events

driskull:
**Related Issue:** #5734

## Summary

feat(tooltip): Add tooltip open, close, beforeOpen, and beforeClose events. (#5734)


Elijbet (approved):
Looks good. We're making some adjustments to this pattern to account for cases when `--calcite-duration-factor` is 0. Still in the works though. We'll update this and all other openClose components after it's installed.

Issue #5540, opened 2022-10-21
[Combobox] Programmatically selecting a combobox item focuses the component


macandcheese:
### Actual Behavior

When programmatically selecting a combobox item, focus is moved to the Combobox. This differs from how other form components work, where the acted upon component does not gain focus.

### Expected Behavior

I wouldn't expect the Combobox to gain focus when programmatically setting the selected value of contained combobox items.

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/rNvXGgz?editors=1000

### Reproduction Steps

1. Open codepen
2. Click the button to set selected on the combobox item.
3. Notice that the combobox has received focus
4. Repeat for input and slider - those components to not receive focus

### Reproduction Version

95

### Esri team

Calcite (design)

driskull:
I think this is the issue; https://github.com/Esri/calcite-components/blob/1b8624e4d2785523bc5670c968065c91b625a3df/src/components/combobox/combobox.tsx#L1027

driskull:
hey @alisonailea don't close the issue, just label it as installed so that it can get verified.

geospatialem:
@alisonailea Can you speak to the skipped tests in `filter`, `input-number`, and `input`? 

The fix works well with `next.642`, but want to make sure the skipped tests are expected.

driskull:
@alisonailea any update on restoring these tests? Do you know why they were skipped?

alisonailea:
> @alisonailea any update on restoring these tests? Do you know why they were skipped?

yep! sorry for the delay. I skipped these tests because they were blocking me from getting my work in. These tests were failing before I did any work. I  will have a PR up soon with some fixes but for the few I still need to skip, I will be added them to the "skipped tests" issue

geospatialem:
> > @alisonailea any update on restoring these tests? Do you know why they were skipped?
> 
> yep! sorry for the delay. I skipped these tests because they were blocking me from getting my work in. These tests were failing before I did any work. I will have a PR up soon with some fixes but for the few I still need to skip, I will be added them to the "skipped tests" issue

@alisonailea Have the skipped tests been added back in? Can you post the PR, or when its merged in, add the PR to this issue?

geospatialem:
A good majority of the skip tests have been addressed via https://github.com/Esri/calcite-components/pull/5885, but there still seems to be a skipped test in `filter`.

[...]

PR #5774, opened 2022-11-17
fix(combobox): 5540 - handle focus

alisonailea:
**Related Issue:** #5540 

## Summary

added a prop to combobox-item that will allow it to inform it's parent if it was clicked or not. 
The combobox parent no longer sets focus on it's text input whenever a combobox-item is updated. Instead it only sets text input focus when there is a keyboard interaction or the updated combobox-item is passed with a clicked prop equal to _true_. This allows combobox-items to be selected programmatically without changing the document.activeElement

- [x] feature or fix has a corresponding test
- [] changes have been tested with demo page in Edge (I still need to get a VM setup)


driskull (changes requested):
Looking good, just have some feedback on how we can approach this without adding a property

src/components/combobox-item/combobox-item.tsx
```diff
@@ -158,6 +165,7 @@ export class ComboboxItem implements ConditionalSlotComponent, InteractiveCompon
       return;
     }
     this.selected = !this.selected;
+    this.calciteComboboxItemFocus.emit(this.el);
```
driskull, line 168: @alisonailea let me see if I'm understanding the issue correctly. From what I understand, the problem is that the `combobox` gets focus on its text input anytime a `combobox-item` is selected. right?

I think this is because the watcher on the combobox item emits the selection event no matter if its from user input or programatic.

```
 @Watch("selected")
  selectedWatchHandler(): void {
    this.calciteComboboxItemChange.emit(this.el);
  }
```

What I was suggesting, is that we change the watcher to emit an internal event named `calciteInternalComboboxItemChange`. This event would not be documented (marked as `@internal`).

This internal event would emit on watch change and its event handler can update selection on the `combobox` but NOT focus the internal input. I think this is ideal because this event really should only fire on user interaction anyway and not programatic changes.

Then, we only emit the public `calciteComboboxItemChange` when a user input occurs. For example in onClick (and keyboard events?) we would call this event. The public event would do the same stuff as the internal event listener plus handle focusing. If we only want to focus on click and not keyboard events then we probably need that internal focus event that you created.

Does that make sense?

src/components/combobox-item/combobox-item.tsx
```diff
@@ -126,6 +126,13 @@ export class ComboboxItem implements ConditionalSlotComponent, InteractiveCompon
    */
   @Event({ cancelable: false }) calciteComboboxItemChange: EventEmitter<DeprecatedEventPayload>;
 
+  /**
+   * Emits whenever the component is focused.
+   *
+   * **Note:**: The event's payload is deprecated, please use the event's `target`/`currentTarget` instead
+   */
+  @Event({ cancelable: false }) calciteComboboxItemFocus: EventEmitter<DeprecatedEventPayload>;
```
driskull, line 134: @alisonailea Do you think we could call this event `calciteInternalComboboxItemChange` instead? I don't think we want to make a public focus event for this fix.

Curious, If this is a new event, why is the payload deprecated?

src/components/combobox-item/combobox-item.tsx
driskull: Just curious, do we want the text box to get focus only on click? Or via keyboard enter/space events as well?

driskull (changes requested):
Thanks for cleaning up the other files @alisonailea!

Looking good. Just had suggestions on the event name and making it private. 

The tests look 💯 🎉

src/components/combobox-item/combobox-item.tsx
alisonailea: after reviewing your comment and reassessing the code I realized that only the combobox has to be aware of focus at all. Since we are already handling all keypress events on combobox we can handle the click events as well. I've updated the code and removed `calciteComboboxItemFocus` completely then added a click event handler to the connectedCallback(and disconnnectedCallback) of combobox

src/components/combobox/combobox.tsx
```diff
@@ -386,6 +388,7 @@ export class Combobox
   disconnectedCallback(): void {
     this.mutationObserver?.disconnect();
     this.resizeObserver?.disconnect();
+    this.el.removeEventListener("click", this.comboboxFocusHandler);
```
driskull, line 391: Is this necessary? I see `onClick={this.clickHandler}` on the div that contains everything inside of the `<host>` and it seems to set focus already...

https://github.com/Esri/calcite-components/blob/c86f7910b9f007419063a863642a595c19ea3a71/src/components/combobox/combobox.tsx#L661-L668

If it is necessary, can you use an onClick on the host element to call the `comboboxFocusHandler`? This way the JSX handles creating and cleaning up the event. ;)

driskull (changes requested):
💥

src/components/combobox/combobox.tsx
alisonailea: there are actually two ways in combobox that focus is set:

setFocus - which puts focus on the textInput AND sets activeChipIndex and activeItemIndex to -1

comboboxFocusHandler - which only sets focus on the textInput

the click event on the div requires setFocus specifically whereas a click event on the combobox-item can not change activeChipIndex or activeItemIndex

driskull (approved):
🎉 Nice!!

src/components/filter/filter.e2e.ts
```diff
@@ -173,7 +173,7 @@ describe("calcite-filter", () => {
       });
     });
 
-    it("updates filtered items after filtering", async () => {
+    it.skip("updates filtered items after filtering", async () => {
```
driskull, line 176: @alisonailea why were these filter tests skipped

src/components/input-number/input-number.e2e.ts
```diff
@@ -522,7 +522,7 @@ describe("calcite-input-number", () => {
       expect(await input.getProperty("value")).toBe(`${finalNudgedValue}`);
     });
 
-    it("when both 'ArrowUp' and 'ArrowDown' are pressed at the same time most recently pressed key takes over", async () => {
+    it.skip("when both 'ArrowUp' and 'ArrowDown' are pressed at the same time most recently pressed key takes over", async () => {
```
driskull, line 525: @alisonailea why were these input tests skipped

src/components/input/input.e2e.ts
```diff
@@ -312,7 +312,7 @@ describe("calcite-input", () => {
       expect(await element.getProperty("value")).toBe("25");
     });
 
-    it("correctly increments and decrements on long hold on mousedown and step is set to a decimal", async () => {
+    it.skip("correctly increments and decrements on long hold on mousedown and step is set to a decimal", async () => {
```
driskull, line 315: @alisonailea why were these input tests skipped

Issue #5776, opened 2022-11-17
Block loading icon spinning too fast


jcfranco:
### Actual Behavior

Loading indicator spins way too fast.

### Expected Behavior

Loading icon to spin like it in version 94.

### Reproduction Sample

https://codepen.io/jcfranco/pen/RwJLZQJ

### Reproduction Steps

Observe the spinner! 🏃🔁

### Reproduction Version

beta.98

### Regression?

beta.94

### Esri team

Calcite (dev)

geospatialem:
Verified in `next.639`.

PR #5778, opened 2022-11-17
fix(block): slow down loading icon spin

jcfranco:
**Related Issue:** #5776

## Summary

This fixes a regression from https://github.com/Esri/calcite-components/pull/5314/ where the block loading icon would spin way too fast.

It uses the same speed as [`calcite-loader`](https://github.com/Esri/calcite-components/blob/master/src/components/loader/loader.scss#L23) for consistency.

Tested manually with the block demo page as animations are not supported by automated testing.

### DIFF
--- a/src/components/block/block.scss
+++ b/src/components/block/block.scss
@@ -98,7 +98,7 @@ calcite-handle {
 }
 
 .status-icon.loading {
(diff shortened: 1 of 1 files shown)


driskull (approved):
89157841-88e3-474a-a1c7-4e0160cd32f4_text

PR #5780, opened 2022-11-17
docs: update changelog's breaking changes.

driskull:
**Related Issue:** #

## Summary

docs: update changelog's breaking changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - cleanup: restore visibility on some cases of input-message
 
 - clean loader readme file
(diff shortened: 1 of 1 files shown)


jcfranco (approved):
👍

jcfranco:
@benelan @geospatialem These will get overridden by the changelog-generating script in the next release, right? What would you suggest to make sure we apply these updates then?

benelan:
I think so, but I'm not 100% sure. I made a note for myself to check during the next release. If it is overwritten I'll fix it and look into fixing the script

driskull:
Yeah ideally the script wouldn't overwrite any modified entries.

driskull:
Looks like the script did overwrite it. We'll need to figure out a way to keep modified changes somehow.

PR #5791, opened 2022-11-22
ci: changelog modifications aren't overwritten after next releases

benelan:
**Related Issue:** #

## Summary
It turns out the default behavior of conventional-changelog is to generate the changelog items since the most recent tag. Whoever wrote the script was deleting all of the next tags and readding them so that the whole changelog would be regenerated. I ended up writing some helpers to combine the existing unreleased changelog items and the new ones.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,45 +10,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### ⚠ BREAKING CHANGES
 
 - **list, list-item, list-item-group:** To know when `calcite-list-item` content is selected, listen to the event `calciteListItemSelect` instead of `click`.
-
-* `List`
-  - Adds `label` property to specify an accessible name for the component.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -10,45 +10,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### ⚠ BREAKING CHANGES
 
 - **list, list-item, list-item-group:** To know when `calcite-list-item` content is selected, listen to the event `calciteListItemSelect` instead of `click`.
-
```
driskull, line 13: @benelan why are you removing this?

driskull: This part should stay.

driskull (approved):
👍

CHANGELOG.md
benelan: Oh okay I thought it was accidently added like Eliza's was. I'll put it back

Issue #5735, opened 2022-11-10
Action inside Tree item can't be clicked with keyboard Space button


cont-kolomeets:
### Actual Behavior

Need to add an action inside a tree item. After focusing and hitting Space, no click event is emitted by the action.
Clicking works when using the mouse, though.

### Expected Behavior

Clicking with Space must work for this case.
I know it might conflict with the selection of tree items, but I'm using the tree with selection-mode='none'.

Also when putting a checkbox in the tree item (see the sample below), it works fine, probably because it emits a different event (`calciteCheckboxChange`).

### Reproduction Sample

https://codepen.io/cont-kolomeets/pen/JjZWYwv?editors=100

### Reproduction Steps

See the sample above.
(1) Add an action inside a tree child. 
(2) Navigate with the keyboard.
(3) Hit Space.
(4) Nothing happens.

### Reproduction Version

beta.95

### Esri team

ArcGIS Business/Community Analyst

driskull:
Valid bug. Seems like we need to not preventDefault on the keyboard events when the tree is in `selection-mode="none"`

geospatialem:
Verified in `next.652`.

![verify-actions-tree-item-keyboard](https://user-images.githubusercontent.com/5023024/205144490-062efcfd-51e9-45cf-be96-5a48faab7f1b.gif)

PR #5800, opened 2022-11-22
fix(tree-item): Allow space and enter key events when selectionMode is "none"

driskull:
**Related Issue:** #5735

## Summary

fix(tree-item): Allow space and enter key events when selectionMode is "none". #5735


eriklharper (approved):
LGTM! 👍

PR #5801, opened 2022-11-22
docs(tree): Update interface for selectionMode.

driskull:
**Related Issue:** 

## Summary

docs(tree): Update interface for selectionMode.

cc @macandcheese


macandcheese (approved):
Awesome, great cleanup for docs site. Will get pulled in next time we bump CC.

PR #5806, opened 2022-11-23
ci(chromatic): use diffThreshold value saved in secrets

benelan:
**Related Issue:** #

## Summary

I forgot to pass the diff threshold GH secret to the chromatic workflow's environment variables 😅
That explains why bumping up the value wasn't doing anything


driskull (approved):
👍

Issue #4079, opened 2022-02-08
calcite-input: pass through native input attributes


noahmulfinger:
### Description

The `accept` attribute restricts the type of files that can be selected when an input has the type `file`. It would be useful to have this attribute available on calcite-input.

### Acceptance Criteria

The `accept` property is passed down from the `calcite-input` into the internal `input`.

### Relevant Info

MDN docs for this attribute: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept

### Which Component

calcite-input

### Example Use Case

Profile image upload:
```html
<calcite-label>
  Upload profile image (must be png or jpg)
  <calcite-input type="file" accept="image/png, image/jpeg"></calcite-input>
</calcite-label>


```

benelan:
@macandcheese @jcfranco can we triage this to the next sprint?

geospatialem:
Verified `accept` and `multiple` props. ✔️ 

@benelan Seeking your expertise 🧠 for verifying the `autocomplete` and `pattern` props are coming in as-expected with the [following Codepen](https://codepen.io/geospatialem/pen/zYaJyzN), where:

1. [`autocomplete`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete):
Unable to reproduce similar behavior to the native input where information is populated, such as email address.

2. [`pattern`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern):
Unable to discern how to use the new prop similar to the native input without using JavaScript to interpret.

benelan:
My understanding is we have a post v1 epic to improve form validation, and for now users will still need JS like you mentioned. Here's an example:
https://codepen.io/benesri/pen/xxzaNKW?editors=1010

cc @driskull in case I'm wrong about that. I can another stab at it if so.

For autocomplete you have to have it turned on in your browser and have saved info. This works for me:
https://codepen.io/benesri/pen/LYrJozq?editors=1010

![image](https://user-images.githubusercontent.com/10986395/204962769-c002ea5b-cece-4ad5-b33c-08d695f53c9e.png)

yelenakreyndel:
@macandcheese  why do we have this issue listed on confluence in the Next section here: https://confluencewikidev.esri.com/display/Calcite/New+Web+Components

macandcheese:
@yelenakreyndel opened a new issue for tracking, will update Confluence: https://github.com/Esri/calcite-components/issues/6545

PR #5807, opened 2022-11-23
feat(input, input-text, input-number): add attributes autocomplete, accept, multiple, pattern

benelan:
**Related Issue:** #4079 

## Summary
Adds native input attributes:

- [multiple](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/multiple) to `calcite-input` for the types specified on MDN
- [accept](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept)  to `calcite-input` for the types specified on MDN
- [pattern](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern) to `calcite-input-text` and  `calcite-input` for the types specified on MDN
- [autocomplete](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) to `calcite-input-number`, `calcite-input-text`, and  `calcite-input` for the types specified on MDN

## Note
 just saw on MDN after committing that Safari may not support `autocomplete`:
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#browser_compatibility

Should I remove it from the PR?


src/components/input-number/input-number.tsx
```diff
@@ -258,6 +258,14 @@ export class InputNumber
    */
   @Prop({ reflect: true }) step?: number | "any";
 
+  /**
+   * Specifies the type of content to autocomplete, for use in forms.
+   * Read the native attribute's documentation on MDN for more info.
+   *
+   * @mdn [step](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete)
+   */
+  @Prop({ reflect: true }) autocomplete?: "string";
```
driskull, line 267: @benelan shouldn't it be `autocomplete: string;`?

We don't need the optional question mark and the type should be string not "string" for these right?

driskull: Do these need to be reflected?

benelan: Woops my bad, fixing. I copy and pasted from the `step` property which has "any" in quotes for some reason. I'll fix that in a separate PR

driskull (approved):
Looks good!

Do you think it would beneficial to make sure these are passed to their internal inputs via tests? if not i'm ok with that.

benelan:
Good idea, I'll add some coverage. I did test them in the demos

PR #5808, opened 2022-11-23
refactor(accordion-item)!: Remove deprecated properties

driskull:
BREAKING CHANGE: Removed the properties `active`, `itemTitle`, `itemSubtitle`, and `icon`.

- Removed the property `active`, use `expanded` instead.
- Removed the property `itemTitle`, use `heading` instead.
- Removed the property `itemSubtitle`, use `description` instead.
- Removed the property `icon`, use `iconStart` or `iconEnd` instead.


benelan (approved):
👍

PR #5809, opened 2022-11-24
refactor(action-menu)!: Remove deprecated calciteActionMenuOpenChange event payload

driskull:
BREAKING CHANGE: Removed the `event.detail` value from the `calciteActionMenuOpenChange` event on the `action-menu` component.

- When listening to `calciteActionMenuOpenChange`, use the `open` property on the `event.target` instead of `event.detail`.


benelan (approved):
👍

PR #5812, opened 2022-11-28
refactor(popover-manager)!: Remove deprecated component

driskull:
BREAKING CHANGE: Removed the `calcite-popover-manager` component. This component is no longer necessary for `calcite-popover`s.


benelan (approved):
🎉

PR #5815, opened 2022-11-28
refactor(block)!: Remove deprecated properties.

driskull:
BREAKING CHANGE: Removed the `summary` and `disablePadding` properties.

- Removed the property `summary`, use `description` instead.
- Removed the property `disablePadding`, use the CSS variable `--calcite-block-padding` instead.


src/components/block/block.stories.ts
```diff
@@ -174,12 +174,6 @@ export const disabled_TestOnly = (): string => html`<calcite-block
   </calcite-block-section>
 </calcite-block>`;
 
-export const paddingDisabled_TestOnly = (): string => html` <calcite-panel heading="Properties">
-  <calcite-block heading="Example block heading" summary="example summary heading" collapsible open disable-padding>
-    <div>calcite components ninja</div>
-  </calcite-block>
-</calcite-panel>`;
```
benelan, line 181: Should we replace this with a snapshot setting `--calcite-block-padding` to zero, instead of completely losing coverage?

benelan (approved):
LGTM. One comment about the padding story, but if you don't think the CSS Variable needs coverage like the property did, I'm cool with that

src/components/block/block.stories.ts
driskull: yeah ill put it back. 👍

PR #5818, opened 2022-11-28
refactor(list, list-item, list-item-group)!: Remove deprecated properties.

driskull:
BREAKING CHANGE: Removed the `headingLevel` and `nonInteractive` properties.

- `headingLevel ` property on the `list` and `list-item-group` is no longer necessary.
- `nonInteractive` property on the `list-item` is no longer necessary.


benelan (approved):
LGTM

PR #5820, opened 2022-11-28
refactor(shell-panel)!: Remove deprecated `calciteShellPanelToggle` event.

driskull:
BREAKING CHANGE: Removed the `calciteShellPanelToggle` event.

- Use a `ResizeObserver` on the component to listen for changes to its size. (https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver)


benelan (approved):
:+1:

PR #5821, opened 2022-11-28
docs: consistent api reference throughout components

geospatialem:
**Related Issue:** #4442

## Summary
1. Includes consistency across API references, this PR finalizes the style guide across all components.

2. Also includes the removal of the optional "`?`" JSDoc tags for the components listed below. 
    - A separate PR will be created for the remaining components.

### Components included:
- ` date-picker`
- `dropdown`
- `dropdown-group`
- `dropdown-item`
- `fab`
- `filter`
- `flow`
-  `flow-item`
- `icon`
- `inline-editable`


src/components/dropdown-group/dropdown-group.tsx
```diff
@@ -36,17 +36,19 @@ export class DropdownGroup {
...
-  /** optionally set a group title for display */
-  @Prop({ reflect: true }) groupTitle?: string;
+  /** Specifies and displays a group title. */
+  @Prop({ reflect: true }) groupTitle: string;
 
   /**
-    specify the selection mode - multiple (allow any number of (or no) active items), single (allow and require one active item),
-   none (no active items), defaults to single
+   * Specifies the component's selection mode, where
+   * `"multi"` allows any number of (or no) selected `calcite-dropdown-item`s,
```
driskull, line 44: Do we need these in quotes?  `"multi"` vs `multi`?

src/components/inline-editable/inline-editable.tsx
```diff
@@ -199,12 +203,12 @@ export class InlineEditable implements InteractiveComponent, LabelableComponent,
...
 
   /**
-   * Emitted when the cancel button gets clicked.
+   * Emits when the cancel editing button is pressed.
    */
   @Event({ cancelable: false }) calciteInlineEditableEditCancel: EventEmitter<void>;
 
   /**
-   * Emitted when the check button gets clicked.
+   * Emits when the confirm edits button is pressed.
```
driskull, line 211: maybe `"confirm edits"`

src/components/inline-editable/inline-editable.tsx
```diff
@@ -199,12 +203,12 @@ export class InlineEditable implements InteractiveComponent, LabelableComponent,
   //--------------------------------------------------------------------------
 
   /**
-   * Emitted when the cancel button gets clicked.
+   * Emits when the cancel editing button is pressed.
```
driskull, line 206: maybe "cancel editing"

driskull (approved):
❤️‍🔥

src/components/date-picker-day/date-picker-day.tsx
```diff
@@ -40,19 +40,19 @@ export class DatePickerDay implements InteractiveComponent {
   /** Day of the month to be shown. */
   @Prop() day!: number;
```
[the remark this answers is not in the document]
geospatialem: We can remove, as long as the documentation makes sense to the devs and is understandable without the comments when interacting with the components. 

Tried to only match the common or same named props for consistency for the devs in those cases, and we could definitely omit altogether if that's the recommend approach. WDYT @driskull?

src/components/dropdown-group/dropdown-group.tsx
geospatialem: For some reason we didn't document it in the original issue, but its to discern string values in the docs similar to this [Popover `overlayPositioning` prop](https://developers.arcgis.com/calcite-design-system/components/popover/#component-api-properties-overlayPositioning).

src/components/inline-editable/inline-editable.tsx
geospatialem: We were reserving the double quotes for property values:
> Use double quotes (") for the values of properties/attributes and event details.

WDYT about:

> Emits when the [component's] confirm edits button is pressed.

and

> Emits when the [component's] cancel editing button is pressed.

src/components/date-picker-day/date-picker-day.tsx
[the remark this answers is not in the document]
driskull: Lets revisit it later before 1.0.

This component isn't designed for a user to ever read its doc or place it in the DOM.

src/components/dropdown-group/dropdown-group.tsx
driskull: Ok sounds good. I just thought the strings were maybe unnecessary.

src/components/inline-editable/inline-editable.tsx
driskull: hmm It seems to weird to reserve quotations for values. the tick in markdown is reserved for code values.

It seems like the button value needs to be surrounded by something. Could we use single quotes?

geospatialem: This is good feedback, updated to double quotes.

PR #5822, opened 2022-11-29
refactor(popover)!: Remove deprecated properties.

driskull:
BREAKING CHANGE: Removed the `closeButton` and `dismissible` properties.

- Removed the property `closeButton`, use `closable` instead.


src/components/popover/popover.stories.ts
```diff
@@ -1,7 +1,7 @@
 import { select, number, text } from "@storybook/addon-knobs";
 import { html } from "../../../support/formatting";
-import { boolean, createSteps, stepStory, storyFilters } from "../../../.storybook/helpers";
```
benelan, line 3: :1st_place_medal:

benelan (approved):
LGTM!

PR #5823, opened 2022-11-29
refactor(panel)!: Remove deprecated properties and events.

driskull:
BREAKING CHANGE: Removed deprecated events and properties.

- Removed the property `dismissed`, use `closed` instead.
- Removed the property `summary`, use `description` instead.
- Removed the property `intlBack`, use the `calcite-flow-item` component instead.
- Removed the property `showBackButton`, use the `calcite-flow-item` component instead.
- Removed the property `beforeBack`, use the `calcite-flow-item` component instead.
- Removed the event `calcitePanelDismiss`, use `calcitePanelClose` instead.
- Removed the event `calcitePanelDismissedChange`, use `calcitePanelClose` instead.
- Removed the event `calcitePanelBackClick`, use the `calcite-flow-item` component instead.


src/components/flow-item/flow-item.tsx
```diff
@@ -302,7 +301,6 @@ export class FlowItem implements InteractiveComponent, LoadableComponent {
           heading={heading}
           headingLevel={headingLevel}
           heightScale={heightScale}
-          intlBack={intlBack}
```
benelan, line 305: Looks like `intlBack` was removed from `calcite-flow-item` in this PR as well. Copy/paste and Search/replace gets the best of us :sweat_smile:

src/components/panel/panel.e2e.ts
```diff
@@ -69,39 +49,19 @@ describe("calcite-panel", () => {
 
     await page.waitForChanges();
 
-    expect(await element.getProperty("dismissed")).toBe(true);
```
benelan, line 72: I think this line should be changed to `closed` instead of completely removed

benelan (approved):
Couple comments, but LGTM.

Small nitpick about the breaking change description:

When listing what was removed and what replaced it, it would be nice if they were all in the same order (e.g 
> removed option ... replacement option

or vice versa)

 For example:

> Use the property `description` instead of `summary`.
> Removed the property `intlBack`. Use `calcite-flow-item` instead.

It would be easier for our users to quickly go through all the changes if the order was the same, so:


> Removed the property `description`, use `description`  instead.
> Removed the property `intlBack`. Use `calcite-flow-item` instead.

Not a huge deal and I don't think we need to go back and change the existing PRs, but lets try to do that moving forward

src/components/flow-item/flow-item.tsx
driskull: The flow-item uses the panel internally. Since this prop has been removed in this PR I need to no longer set it on the panel.

benelan: Ahh I missed `intlBack` being used in `getBackLabel`, I thought it wasn't being used at all anymore. Thanks for clarifying

PR #5826, opened 2022-11-29
refactor(tip-manager)!: Remove deprecated `calciteTipManagerToggle` event.

driskull:
BREAKING CHANGE: Removed the `calciteTipManagerToggle` event, use `calciteTipManagerClose` instead.


benelan (approved):
LGTM after the e2e test is fixed. It's using the removed `toggleEventSpy` in expect:

```js
 expect(toggleEventSpy).toHaveReceivedEvent();
```

Issue #5648, opened 2022-10-26
Pagination: `numberingSystem` and `lang` are only respected when `groupSeparator` prop is present


geospatialem:
### Actual Behavior

The component will only render a specified `numberingSystem` and `lang` when the `groupSeparator` prop is used on the component.

![image](https://user-images.githubusercontent.com/5023024/198126422-08bdbc5e-ca26-46f7-a4dc-44c495cf3e9b.png)

### Expected Behavior

The component should render the specified `numberingSystem` and `lang` when defined, regardless of the `groupSeparator` prop.

### Reproduction Sample

https://codepen.io/geospatialem/pen/ZEREoZV

### Reproduction Steps

1. Open the [sample](https://codepen.io/geospatialem/pen/ZEREoZV)
2. Observe the first example works as-expected when the `groupSeparator` prop is present
3. Observe the second example renders the numbers in `"en"` when the `groupSeparator` prop is not present

### Reproduction Version

next.611

### Relevant Info

Related to work performed on `numberingSystem` added via #4893 and #5259.

### Esri team

Calcite (dev)

geospatialem:
Verified on `next.650`.

![image](https://user-images.githubusercontent.com/5023024/205102805-c277725e-4db8-4391-bc83-4e042f390628.png)

PR #5828, opened 2022-11-29
fix(pagination): numberingSystem and lang properties work without groupSeparator

benelan:
**Related Issue:** #5648

## Summary
`calcite-pagination` was only formatting (localizing) the page numbers when the groupSeparator property was set to `true`. Now lang/numberingSystem work correctly with or without groupSeparator


driskull (approved):
👍

PR #5838, opened 2022-11-29
refactor(chip)!: remove deprecated properties and values.

anveshmekala:
BREAKING CHANGE: Removed the `dismissible` property and the `clear` value for the `appearance` property.

- Use the value `transparent` instead of `clear` for `appearance` property.


anveshmekala:
> @anveshmekala can you check for any readme, html, story, and test files that need to be updated?
> 
> Can you change the description/title to:
> 
> ```
> BREAKING CHANGE: Removed the `dismissible` property.
> 
> - Removed the property `dismissible`, use `closable` instead.
> ```

For readme, we can rely on update-doc Github action.

driskull:
> For readme, we can rely on update-doc Github action.

@anveshmekala can you update the usage file markdown?

src/components/chip/usage/Basic.md
```diff
@@ -1,3 +1,3 @@
 ```html
-<calcite-chip value="Global" dismissible icon="globe" appearance="clear" color="green">Global</calcite-chip>
+<calcite-chip value="Global" closable icon="globe" appearance="clear" color="green">Global</calcite-chip>
```
driskull, line 2: @anveshmekala one more thing...

I think the `appearance` value of `clear` is deprecated. Can we change them all to `transparent`?

driskull: We could always do this later if you would like that better.

anveshmekala: Good catch. will change it to `transparent`. And get the `conditionalSlotComponent` removed as well.

driskull: We can do the `connectConditionalSlotComponent` in a separate pr/epic. Lets wait on that one.

PR #5846, opened 2022-11-30
ci: more tweaks to prevent chromatic diffs for text aliasing

benelan:
**Related Issue:** #

## Summary

Some more changes to stop text aliasing from triggering Chromatic diffs

### DIFF
--- a/src/components/slider/slider.stories.ts
+++ b/src/components/slider/slider.stories.ts
@@ -7,7 +7,11 @@ import { html } from "../../../support/formatting";
 export default {
   title: "Components/Controls/Slider",
   parameters: {
-    notes: readme
+    notes: readme,
+    chromatic: {
+      // https://www.chromatic.com/docs/threshold
+      diffThreshold: Number(process.env.CHROMATIC_DIFF_THRESHOLD) || 0.3
+    }
   },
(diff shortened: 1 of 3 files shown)


driskull (approved):
👍

src/components/slider/slider.stories.ts
```diff
@@ -7,7 +7,11 @@ import { html } from "../../../support/formatting";
 export default {
   title: "Components/Controls/Slider",
   parameters: {
-    notes: readme
+    notes: readme,
+    chromatic: {
+      // https://www.chromatic.com/docs/threshold
+      diffThreshold: Number(process.env.CHROMATIC_DIFF_THRESHOLD) || 0.3
```
driskull, line 13: Do you think setting this globally would be too relaxed?

benelan: I think 0.15 will probably be okay globally, its more than double the default value. I think we've been using the default this whole time because I didn't put the property in `chromatic: {}` 😅

I bumped the fallbacks in case the environment variable doesn't work, but I'm still attempting to tweak it from the GH Secret. And then I'll commit the value once we get something that works

PR #5853, opened 2022-11-30
test(custom-theme): use placeholderImage

benelan:
**Related Issue:** #5812

## Summary
The image service we were using for the card's thumbnail in the custom theme story is/was down, which caused CI failures. For example:
https://www.chromatic.com/test?appId=6266d45509d7eb004aa254fb&id=6386b5d2e340d812b9c9a134

I switched it out to use our placeholderImage util.


driskull (approved):
Sweeet 🤘🎸

PR #5482, opened 2022-10-14
refactor(storybook): fix type errors

benelan:
**Related Issue:** #

## Summary
Fixes type errors found by running `tsc --p ./tsconfig-storybook.json`. Also excludes autogenerated files from the eslint tsconfig.

PR #5854, opened 2022-11-30
ci(chromatic): hardcode diffThreshold to 0.7

benelan:
# Edit

Alright, this is ready for review now. I set the diffThreshold to 0.7 which is higher than I had hoped, but it seems to be working. We need to keep an eye out for a bit to make sure actual changes are caught, but it should be fine. We are still getting false positives from the floating-ui components, but [specifying the viewport](https://www.chromatic.com/docs/viewports) may help. I'll work on that in a separate PR.

---

**Related Issue:** #

## Summary
I bumped the GH Secret up to `1` and we still got diffs. I'm hoping its just something with the Secret -> Env Var pipeline that's causing issues.

To make sure I'm barking up the right tree, I bumped up the diff threshold for slider and date-picker to the most relaxed. I also added the threshold per story for slider, because that is how they show it documented:
https://www.chromatic.com/docs/threshold

I'm hoping that date-picker passes as well so I don't have to go through and add it to every story lol. 

I've also been trying to save snapshots by checking real PRs for the chromatic results, but I'm just going to rerun it in this PR until I get it working. We have plenty of snapshots left and its the end of the month.

### DIFF
--- a/src/components/slider/slider.stories.ts
+++ b/src/components/slider/slider.stories.ts
@@ -1,4 +1,5 @@
 import { text, number, array, boolean as booleanFn, select } from "@storybook/addon-knobs";
+
 import { boolean, storyFilters } from "../../../.storybook/helpers";
 import { themesDarkDefault } from "../../../.storybook/utils";
 import readme from "./readme.md";
@@ -10,7 +11,7 @@ export default {
     notes: readme,
     chromatic: {
       // https://www.chromatic.com/docs/threshold
-      diffThreshold: 0.65
+      diffThreshold: 0.7
     }
   },
   ...storyFilters()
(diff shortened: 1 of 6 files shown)


src/components/slider/slider.stories.ts
```diff
@@ -158,8 +155,8 @@ export const rangeLabeledTicksEdgePositioningAtMin_TestOnly = (): string => html
   ></calcite-slider>
 `;
 
-export const Histogram = (): HTMLCalciteSliderElement => {
-  const slider = document.createElement("calcite-slider") as HTMLCalciteSliderElement;
+export const Histogram = (): any => {
```
driskull, line 158: Why the change to type as `any` in these cases?

src/components/slider/slider.stories.ts
```diff
@@ -186,37 +183,8 @@ export const Histogram = (): HTMLCalciteSliderElement => {
   return slider;
 };
 
-export const HistogramWithColors = (): HTMLCalciteSliderElement => {
```
driskull, line 189: Did we need to remove this story?

driskull (approved):
👍

src/components/slider/slider.stories.ts
benelan: We don't have access to the component types from the stories so it throws linting errors
image

I'm pretty sure I tried importing `components.d.ts` at one point, which fixed the linting errors but caused build errors. I can try to dig that up again when I have some time

benelan: Yeah it was in #5482

Here was the error that popped up when importing `components.d.ts`:
https://github.com/Esri/calcite-components/actions/runs/3275079841/jobs/5389458993#step:6:50

alisonailea (approved):

PR #5858, opened 2022-11-30
refactor(tree)!: Remove deprecated `inputEnabled` property

driskull:
BREAKING CHANGE: Removed the `inputEnabled` property.

- Removed the property `inputEnabled`, use `selectionMode="ancestors"` instead.

### DIFF
--- a/src/components/tree-item/tree-item.tsx
+++ b/src/components/tree-item/tree-item.tsx
@@ -91,14 +91,6 @@ export class TreeItem implements ConditionalSlotComponent, InteractiveComponent 
   @Prop({ reflect: true, mutable: true }) lines: boolean;
 
   /**
-   * Displays checkboxes (set on parent).
-   *
-   * @internal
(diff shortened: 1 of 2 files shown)


src/components/tree-item/tree-item.tsx
```diff
@@ -90,14 +90,6 @@ export class TreeItem implements ConditionalSlotComponent, InteractiveComponent
    */
   @Prop({ reflect: true, mutable: true }) lines: boolean;
 
-  /**
-   * Displays checkboxes (set on parent).
-   *
-   * @internal
-   * @deprecated Use `selectionMode="ancestors"` for checkbox input.
-   */
-  @Prop() inputEnabled: boolean;
```
driskull, line 99: Looks like this was already a breaking change at some point since it isn't used anywhere.

alisonailea (approved):

PR #5861, opened 2022-11-30
refactor(label)!: Remove deprecated properties

driskull:

- Removed the property `status`, set the `status` property on the component the label is bound to instead.
- Removed the property `disabled`, set the `disabled` property on the component the label is bound to instead.
- Removed the property `disableSpacing`, use the CSS variable `--calcite-label-margin-bottom` instead.


alisonailea (approved):

src/components/label/label.scss
```diff
@@ -101,9 +101,3 @@
     @apply bg-opacity-0;
   }
 }
-
-:host([disable-spacing]) {
-  .container {
-    @apply m-0 gap-0;
```
driskull, line 107: @macandcheese @asangma do we need another CSS var for setting gap to zero?  `--calcite-label-margin-bottom` only sets the margin so a gap still exists. Could we just have one or another somehow?

driskull:
@eriklharper @macandcheese the CSS var `--calcite-label-margin-bottom` can't replace `disablePadding` because the `disablePadding` attribute set the CSS gap to zero and the CSS var only handled margin.

Is there a way we can just use gap or margin? If not, do I need to add another CSS var to remove the gap?

src/components/label/label.scss
driskull: ping @eriklharper @macandcheese do we need both gap and margin? If so, should I introduce another variable for devs to remove the gap?

Issue #5866, opened 2022-12-01
[Modal] Restore deprecated css var for scrim bg


macandcheese:
### Actual Behavior

During the consistency / cleanup, it seems we accidentally deprecated the ‘--calcite-scrim-background’ css var: https://developers.arcgis.com/calcite-design-system/components/modal/#component-api-styles---calcite-scrim-background

The rest of the deprecations for Modal css vars seem correct.

### Expected Behavior

The variable shouldn’t be deprecated.

### Reproduction Sample

No

### Reproduction Steps

No

### Reproduction Version

98

### Esri team

Calcite (design)

geospatialem:
While the CSS var is back is back in the docs, its not functional in `next.668`.

- Working example when using a css var for `scrim` when coupled with `modal`: https://codepen.io/geospatialem/pen/rNKbmWE
- Non-working example with `modal`: https://codepen.io/geospatialem/pen/jOKRmeJ 
  - Going back many versions, unable to see this functional 

Wondering the best approach forward, should we do one or many of the following?:
1. Document the use of `scrim` and the `--calcite-scrim-background` css var
2. Try to add functionality to the `modal --calcite-scrim-background` css var
3. Another alternative?

cc: @macandcheese

macandcheese:
Maybe the `calcite-scrim` component should have a `--calcite-scrim-background-color` variable, and `calcite-modal` should have its own `--calcite-modal-scrim-background-color` variable? 

I've added a note in this issue for the former: https://github.com/Esri/calcite-components/issues/5958

geospatialem:
Going to close out this issue since its documented in the above issue (also updated the description). 

@macandcheese The above issue isn't slotted for the 1.0 release at this time, but we could prioritize variables as we see fit.

PR #5868, opened 2022-12-01
fix(modal): restore deprecated scrim background css property

anveshmekala:
**Related Issue:** #5866 

This PR will restore the deprecated `--calcite-scrim-background` CSS property.


driskull (approved):
👍

PR #5869, opened 2022-12-01
refactor(stepper-item)!: Remove deprecated properties

driskull:

- Removed the property `active`, use `selected` instead.
- Removed the property `itemTitle`, use `heading` instead.
- Removed the property `itemSubtitle`, use `description` instead.


benelan (approved):
👍

PR #5871, opened 2022-12-01
docs(changelog): fix breaking change indent levels

benelan:
**Related Issue:** #


driskull (approved):
👍 nice! 

I had noticed this as well :)

benelan:
Yeah I wonder if maybe indenting the bullet points in the PR description will help. It doesn't show up as a bullet point in the PR body, but it should show up as a subpoint in the changelog. For example

```
    - Removed x, use y instead
```

instead of 

```
- Removed x, use y instead
```

Issue #3338, opened 2021-10-22
Enhancement: Alert - Usability improvements to auto-dismiss.


macandcheese:
### Description
Two related proposals to solve potential annoyances for users:

- If an alert is auto-dismissible, currently we remove the close button. We should keep the close button there.

- If an alert is auto-dismissible, it may dismiss while a user tries to interact with it. We should pause the dismiss timer while a user is hovering over, or focusing on an element within an alert. I've seen this in the wild and its a very nice interaction (only reference I could find immediately was this https://bootstrap-vue.org/docs/components/toast#comp-ref-b-toast-props)

#### Which Component
Alert

#### Example Use Case
 
For the first - a user may want to dismiss the alert before the dismiss duration completes. 

For the second, although disincentivized, apps may put actions or links inside of an auto-dismissible alert. Currently, the alert dismissal timer continues, even if they are hovering over the alert, or using keyboard to navigate and focus within.

jul11557:
@macandcheese +100 on this! We have a custom alert in ArcGIS Mission that does exactly what you're suggesting:

- It allows users to manually dismiss the alert with an `x` button _and_ we set an automatic dismiss timer on the alert. That way, we can allow un-intrusive messages to automatically disappear, and impatient users could easily tap the X to close.
- It remains on-screen when users click and hold the alert (allowing them more time to read and digest the message, if needed). I like the suggestion above about keeping it on hover.

We just swapped to the calcite alert, but are currently evaluating whether we should keep the calcite component or switch back to the custom alert we previously had. The lack of the close action might be the deal-breaker. The calcite alert looks more visually appealing, but without the enhancements you've listed in the issue above, the compenent can be a bit frustrating to interact with.

One interesting use case we have that may be worth considering: 
Our app has enabled third-party sensor buttons. Someone can tap a smart button and it triggers an alert in the web app. These alerts are displayed with an `alert` component, and can include audio and sound. (Think of an emergency scenario: someone is responding to a search and rescue mission and has a sudden emergency. The person taps a button they are wearing on their vest, which triggers an emergency alert. An analyst, who is monitoring the mission from their desktop, sees the alert and hears sirens playing through the web app. The alert remains on-screen, un-blocked from the user's view, until the analyst has time to address the situation.)

jul11557:
No prob! 

[...]

PR #5872, opened 2022-12-01
fix(alert): auto-dismissible retains close button and dismisses timer while a user is hovering over

Elijbet:
**Related Issue:** #3338

## Summary

If an `alert` is `auto-dismissible`, 

- keep the `close` button there, a user may want to dismiss the alert before the dismiss duration completes.

- pause the dismiss timer while a user is hovering over, or focusing on an element within an alert. Although disincentivized, apps may put actions or links inside of an `auto-dismissible alert`.


src/components/alert/alert.stories.ts
```diff
@@ -203,3 +203,11 @@ export const actionsEndQueued_TestOnly = (): string => html`
     }, "1000");
   </script>
 `;
+
+export const autoDismissableRetainsCloseButton_TestOnly = (): string => html`
+  <calcite-alert auto-dismiss="" auto-dismiss-duration="medium" open="" scale="m" color="blue">
```
driskull, line 208: nitpick: ` auto-dismiss=""` should just be ` auto-dismiss`

driskull: Same with `open`.

src/components/alert/alert.tsx
```diff
@@ -226,6 +226,10 @@ export class Alert implements OpenCloseComponent, LocalizedComponent, LoadableCo
             queued,
             [placement]: true
           }}
+          onMouseLeave={
```
driskull, line 229: Can you use the pointer event equivalent instead of mouse event? `onPointerLeave/onPointerOver`?

I think `pointerout` might be better than pointerleave.

src/components/alert/alert.tsx
```diff
@@ -423,4 +429,17 @@ export class Alert implements OpenCloseComponent, LocalizedComponent, LoadableCo
   private actionsEndSlotChangeHandler = (event: Event): void => {
     this.hasEndActions = slotChangeHasAssignedElement(event);
   };
+
+  private handleMouseOver = (): void => {
+    window.clearTimeout(this.autoDismissTimeoutId);
```
driskull, line 434: Do any of these timeouts need to be cleared on destroy?

driskull (changes requested):
Looking good!

driskull (approved):
👍

macandcheese:
Extremely awesome work @Elijbet !! The only thing I notice is that while it pauses while hovering, it's not pausing when focus is inside the alert and the cursor is not. For instance, if there is a slotted `calcite-link` that has focus, the timer should pause as well. Maybe we can just add this as a follow up for later?

Elijbet:
> Extremely awesome work @Elijbet !! The only thing I notice is that while it pauses while hovering, it's not pausing when focus is inside the alert and the cursor is not. For instance, if there is a slotted `calcite-link` that has focus, the timer should pause as well. Maybe we can just add this as a follow up for later?

Yeah, that's a good point, I don't think I thought of that. Lets log it as an issue and I'll address that as well.

Issue #5388, opened 2022-09-27
Tooltip appears above modal overlay


nwhittaker:
### Actual Behavior

If a tooltip is visible when a modal appears, the tooltip is layered above the modal's background overlay.

<img width="426" alt="Screen Shot 2022-09-27 at 10 54 09 AM" src="https://user-images.githubusercontent.com/421496/192561049-a467015d-c181-48d4-8dd7-9f8efba703d1.png">

### Expected Behavior

The tooltip is layered below the overlay.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/eYryBbY

### Reproduction Steps

1. Visit the sample
2. Hover the button until a tooltip appears
3. Click the button to open the modal

### Reproduction Version

beta.94

### Relevant Info

Field Maps has many icon-only buttons with tooltips (e.g. delete map, copy map, etc...) that launch modals.

### Esri team

ArcGIS Field Apps

jcfranco:
This is caused by our predefined z-index layer values where tooltip uses a popup z-index value of 999 and modal and its overlay use 700 and 800, respectively.

As a workaround, you could apply a [z-index of 1000 to `calcite-modal`](https://codepen.io/jcfranco/pen/GRdzyYg?editors=1010).

@macandcheese @ashetland @SkyeSeitz @driskull Do we to tweak the z-index used by tooltips (and other floating-ui components) to use the same as modal?

driskull:
> Do we to tweak the z-index used by tooltips (and other floating-ui components) to use the same as modal?

That might be the best thing to do. If a tooltip is inside a modal it should still appear above the modal right?

ashetland:
That was going to be my question as well, can a tooltip be invoked inside a modal and how would the z-index affect that? One thing I noticed in the example was that I expected the Tooltip to dismiss as soon as the Modal opened. That might be another way to solve this, but I'm not sure about feasibility.

macandcheese:
> > Do we to tweak the z-index used by tooltips (and other floating-ui components) to use the same as modal?
> 
> That might be the best thing to do. If a tooltip is inside a modal it should still appear above the modal right?

I'd agree yes

driskull:
Making the tooltip/popover the same zIndex as the overlay doesn't seem to work because when you put a tooltip inside the modal it wont' display. This is because the modal is positioned fixed.

I think the best solution is to allow a user to set the zIndex of a component using a CSS variable for cases like this.

[...]

PR #5873, opened 2022-12-01
fix(tooltip): Prevent tooltip from appearing above modal overlay

driskull:
**Related Issue:** #5388

## Summary

fix(tooltip): Prevent tooltip from appearing above modal overlay. #5388

- Updates floating-ui to use tailwind configuration
- Sets default z-index to be that of a dropdown
- Adds ability for users to customize zindex of tooltips and popovers
  - CSS vars `--calcite-tooltip-z-index` and `--calcite-popover-z-index`

```html
 <calcite-tooltip style="--calcite-tooltip-z-index: 500;" label="Open modal" reference-element="button">
Open modal
</calcite-tooltip>
```

### DIFF
--- a/src/assets/styles/_floating-ui.scss
+++ b/src/assets/styles/_floating-ui.scss
@@ -5,7 +5,7 @@ $floating-ui-transform-right: translateX(-5px);
 
 :root {
   --calcite-floating-ui-transition: var(--calcite-animation-timing);
-  --calcite-floating-ui-z-index: theme("zIndex.popover");
+  --calcite-floating-ui-z-index: theme("zIndex.dropdown");
 }
 
 @mixin floatingUIAnim {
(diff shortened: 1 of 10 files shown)


src/assets/styles/_floating-ui.scss
```diff
@@ -2,10 +2,10 @@ $floating-ui-transform-bottom: translateY(-5px);
 $floating-ui-transform-top: translateY(5px);
 $floating-ui-transform-left: translateX(5px);
 $floating-ui-transform-right: translateX(-5px);
-$floating-ui-default-z-index: 900;
 
 :root {
   --calcite-floating-ui-transition: var(--calcite-animation-timing);
+  --calcite-floating-ui-z-index: theme("zIndex.dropdown");
```
driskull, line 8: Makes the default zIndex that of a dropdown.

src/components/combobox/combobox.scss
```diff
@@ -143,6 +143,7 @@
 }
 
 .floating-ui-container {
+  --calcite-floating-ui-z-index: theme("zIndex.dropdown");
```
benelan, line 146: Nitpick: Should we rename `zIndex.dropdown` to `zIndex.default` since it is now being used as the default, and no longer specific to dropdown?

benelan (approved):
👍

src/components/combobox/combobox.scss
driskull: hmm our tailwind already has a zindex.default of 1. I think this is fine as is.

PR #5875, opened 2022-12-02
fix(action): Add screen reader support for `active` and `indicator` props

driskull:
**Related Issue:** #4813

## Summary

fix(action): Add screen reader support for `active` and `indicator` properties. #4813

- Uses `aria-pressed` to describe toggle state for `active` property
- Uses `aria-live` region to describe `Unsaved changes` for `indicator` property


src/components/action/action.tsx
```diff
@@ -60,6 +60,18 @@ export class Action implements InteractiveComponent, LoadableComponent {
...
+   * Specifies the text label to display `indicator` is `true`.
+   */
+  @Prop() indicatorMessage = "";
+
+  /**
+   * Specifies the text label to display `indicator` is `true`.
+   *
+   * @default "Unread changes"
+   */
+  @Prop() intlIndicator?: string = TEXT.indicator;
```
geospatialem, line 73: I think we were going to be removing the optional props with JSDoc references of "`?`" from the docs? 

And in this case if the prop isn't present there is no message read back to AT if line 196 remains the same.

src/components/action/action.tsx
```diff
@@ -165,6 +183,21 @@ export class Action implements InteractiveComponent, LoadableComponent {
...
+    const { indicator, intlIndicator, indicatorMessage, indicatorId, buttonId } = this;
+    return (
+      <div
+        aria-labelledby={buttonId}
+        aria-live="polite"
+        class={CSS.indicatorText}
+        id={indicatorId}
+        role="region"
+      >
+        {indicator ? indicatorMessage ?? intlIndicator : null}
```
geospatialem, line 196: When the `indicatorMessage` prop is provided with a string, this works really well. However, no prop present provides no context. Would it be possible to have a default value for the indicator such as "indicator present for ${text}"?

geospatialem (approved):
Fantastic improvement, @driskull! 🙌🏻 

We're giving context when the indicator is added to the component on-the-fly. It's missing context if an indicator is already present on the component. 

It provides instant feedback, which is a great improvement and good starting point for the component.

A few comments below regarding the `indicatorMessage` prop, which would be awesome to have a default value assigned to it.

src/components/action/action.tsx
driskull: @anveshmekala will have to add a string for this and update it in his PR.

driskull:
> It's missing context if an indicator is already present on the component.

Not sure how we can do anything in this case. Do you know?

geospatialem:
> Not sure how we can do anything in this case. Do you know?

If we have a default text value present in `indicatorMessage` and add context to the `ariaLabel`, we can provide context on load. 

Very rough example where if present the value is read back with the `aria-label`, otherwise the empty string will ignore an indicator status:

```ts
const {
      active,
      compact,
      disabled,
      loading,
      textEnabled,
      label,
      text,
      indicator,
      indicatorId,
      indicatorMessage,
      intlIndicator,
      buttonId
    } = this;

    const indicatorPresent = indicator ? indicatorMessage ?? intlIndicator : "";

    const ariaLabel = label + indicatorPresent || text + indicatorPresent;
```

driskull:
@geospatialem I could use a concatenated `aria-label` instead of the live region if thats the way we want to go. Let me know @geospatialem. I was concerned that concatenating wouldn't work well in all languages.

geospatialem:
> @geospatialem I could use a concatenated `aria-label` instead of the live region if thats the way we want to go. Let me know @geospatialem. I was concerned that concatenating wouldn't work well in all languages.

We'd still need the live region for providing context when the indicator is changed, so both would need to be present for context to AT users.

Good call on the concatenation for language support, what if we used a template literal, such as:

```ts
const indicatorStatus = indicator ? indicatorMessage ?? intlIndicator : "";
const ariaLabel = `${label} ${indicatorStatus}` || `${text} ${indicatorStatus}`;
```

driskull:
> Good call on the concatenation for language support, @driskull what if we used a template literal, such as:

I can do that if we don't think it will be an issue to combine the two strings.

driskull:
@geospatialem can you test again?

geospatialem:
> @geospatialem can you test again?

@driskull Looking great! 😎 

JAWS transcript below, with a similar experience in NVDA, with `indicatorMessage` set as  `"My indicator message"`:

If the indicator is present on load:
```
hello world (My indicator message) Toggle button Pressed 
To toggle the state press spacebar.
```

after the indicator is removed:
```
hello world
```

and after the indicator is added:
```
hello world (My indicator message)
```

driskull:
Ready for review.

src/components/action/action.tsx
```diff
@@ -196,9 +224,21 @@ export class Action implements InteractiveComponent, LoadableComponent {
...
+      textEnabled,
+      label,
+      text,
+      indicator,
+      indicatorId,
+      buttonId,
+      intlIndicator
+    } = this;
+
+    const ariaLabel = `${label || text}${indicator ? ` (${intlIndicator})` : ""}`;
```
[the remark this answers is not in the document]
anveshmekala: nvm. it will be covered [ here](https://github.com/Esri/calcite-components/pull/5471/files#diff-6d2b1195f68eccf57de81ad892c63c9b647f05afca4bb4fd55b42daa9005dcb0R45).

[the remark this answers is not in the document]
driskull: Yep. Can you approve?

anveshmekala (approved):
🚀

Issue #3032, opened 2021-09-14
Enhancement: Add a slot in List Item for specialized content or a slot for label and a slot for description


ffaubry:
### Description

Currently, the only way to have content is by providing label and/or description.
It would be useful to also support a slot 'content' for specialized content.
In my particular case, we would like a second line for the description:

Label
Description 1
Description 2

#### Which Component

calcire-list-item

#### Example Use Case

Support more specialized content for list item

geospatialem:
Verified on `next.664` with the [following Codepen](https://codepen.io/geospatialem/pen/zYaXovV), where the `content` slot will override the `label` and `description` props.

PR #5876, opened 2022-12-02
feat(list-item): Add content slot for specialized content

driskull:
**Related Issue:** #3032

## Summary

feat(list-item): Add slot for specialized content #3032

- Adds a new `content` slot to the `list-item` for customized content.
- `content` slot takes precedence over `label` and `description` properties.

### DIFF
--- a/src/components/list-item/list-item.tsx
+++ b/src/components/list-item/list-item.tsx
@@ -175,6 +175,8 @@ export class ListItem implements InteractiveComponent, LoadableComponent {
 
   @State() hasActionsEnd = false;
 
+  @State() hasContent = false;
+
   @State() hasContentStart = false;
 
   @State() hasContentEnd = false;
@@ -321,6 +323,15 @@ export class ListItem implements InteractiveComponent, LoadableComponent {
     );
   }
 
+  renderContent(): VNode {
+    const { hasContent } = this;
+    return (
+      <div class={CSS.content} hidden={!hasContent}>
+        <slot name={SLOTS.content} onSlotchange={this.handleContentSlotChange} />
+      </div>
+    );
+  }
+
   renderContentEnd(): VNode {
     const { hasContentEnd } = this;
     return (
@@ -330,10 +341,10 @@ export class ListItem implements InteractiveComponent, LoadableComponent {
     );
   }
 
-  renderContent(): VNode {
-    const { label, description } = this;
+  renderContentProperties(): VNode {
+    const { label, description, hasContent } = this;
 
-    return !!label || !!description ? (
+    return !hasContent && (!!label || !!description) ? (
       <div class={CSS.content} key="content">
         {label ? (
           <div class={CSS.label} key="label">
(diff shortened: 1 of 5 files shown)


src/components/list-item/list-item.tsx
```diff
@@ -33,6 +33,7 @@ import {
  * @slot - A slot for adding `calcite-list-item` and `calcite-list-item-group` elements.
  * @slot actions-start - A slot for adding actionable `calcite-action` elements before the content of the component.
  * @slot content-start - A slot for adding non-actionable elements before the label and description of the component.
+ * @slot content - A slot for adding non-actionable centered content in place of the label and description of the component.
```
benelan, line 36: Can you change the description to

> A slot for adding non-actionable, centered content in place of the \`label\` and \`description\` of the component.

benelan (changes requested):
Something is funky with the focus when using the `content` slot. You can repro in the story after removing the `disabled` attributes, or put this at the end of the list demo:

```html
    <!-- custom content -->
    <div class="parent">
      <style>
        .list-chip {
          margin-inline-end: 0.5rem;
        }
      </style>

      <div class="child right-aligned-text">custom content</div>
      <div class="child">
        <calcite-list>
          <calcite-list-item>
            <div slot="content">
              <strong>Cras iaculis ultricies nulla.</strong>
              <div>Lorem ipsum dolor sit amet, consectetuer adipiscing elit.</div>
            </div>
          </calcite-list-item>
          <calcite-list-item>
            <div slot="content">
              <strong>Cras iaculis ultricies nulla.</strong>
              <div>Lorem ipsum dolor sit amet, consectetuer adipiscing elit.</div>
            </div>
          </calcite-list-item>
          <calcite-list-item>
            <div slot="content">
              <strong>Cras iaculis ultricies nulla.</strong>
              <div>Lorem ipsum dolor sit amet, consectetuer adipiscing elit.</div>
            </div>
          </calcite-list-item>
        </calcite-list>
      </div>
    </div>
```

For the most part, I can't get the focus ring to show up when clicking or tabbing. However if you click/tab for long enough the focus ring will eventually show up on one of the items. I also noticed some inconsistent focus ring behavior for the "start/end content slots" demo, so it might not be completely related to this PR. But the content slot focus behavior seems a lot worse

driskull:
> Something is funky with the focus when using the content slot. You can repro in the story after removing the disabled attributes, or put this at the end of the list demo:

@benelan can you create an issue for this one? I do see the problem.

benelan (approved):
Yeah I confirmed it's reproducible on our `next` storybook deployment, so it isn't a result of the changes in this PR. 
https://esri.github.io/calcite-components/?path=/story/components-list--start-and-end-content-slots

If you rather address it in a different PR I'm okay with that; I'll create an issue. Other than that and the doc nitpick, LGTM

driskull:
> If you rather address it in a different PR I'm okay with that; I'll create an issue. Other than that and the doc nitpick, LGTM

Yeah, it would be easier to track and report that way 👍

PR #5879, opened 2022-12-02
build(tsconfig): bump target to es2020 and lib to es2021

benelan:
**Related Issue:** #

## Summary

Creating this PR based on the Teams discussion about increasing our `target`. I confirmed with @dasa that these changes are safe for our stakeholders. But any `target` higher than 2020 has the potential to cause them problems.

lib 2021 has some useful stuff we can benefit from, like `Promise.any`, `string.prototype.replaceAll`, and logical assignment operators.


driskull (approved):
👍

alisonailea (approved):
Thanks Ben!

benelan:
Visual regressions are unrelated to this PR. Installing!

- Chromatic changes are due to that text shifting problem we've seen with list, slider, date-picker, and pagination. It seems to be worse in the snapshots using Arabic numerals. Still investigating this; I may try to add a short delay before the snapshot is taken.
- Screener changes are due to a focus ring either showing up or not showing up around popover's close button. I've seen these change back and forth a few times; it's likely timing related.

Issue #5855, opened 2022-11-30
The popup of the input-date-picker and  input-time-picker didn't change language after setting the "lang" property.


nanzhang01:
### Actual Behavior

After changing the "lang" of a loaded input-date-picker or input-time-picker, the language of the popup didn't update.
Sample: https://codepen.io/nan9266/pen/PoadpNx
Screenshot:
![beta97gif](https://user-images.githubusercontent.com/29697780/204747062-3c53e0a6-bfdc-4392-bf25-947e282e3e37.gif)

### Expected Behavior

After changing the "lang" of a loaded input-date-picker or input-time-picker, the language of the popup should update, like the 1.0.0-beta.95 version did.
Sample: https://codepen.io/nan9266/pen/QWxVdEB

Screenshot: 
![beta95gif](https://user-images.githubusercontent.com/29697780/204747405-605fec56-e789-4bdf-b85f-e832774f34c8.gif)

### Reproduction Sample

https://codepen.io/nan9266/pen/PoadpNx

### Reproduction Steps

1. Load a input-date-picker and a input-time-picker in html, set the `lang` as 'en'.
2. After page loaded, change the `lang` of the elements, like 'ko', 'zh-CN', click the picker to check the language in the popup.
3. Observe, will find the language in the pupup didn't change.

### Reproduction Version

1.0.0-beta.97

### Regression?

1.0.0-beta.95

### Impact

Regression in survey123 web app: switching language does not change the date/time controller.

### Esri team

ArcGIS Survey123

ESRIi18n:
@benelan and @geospatialem I added the installed label by accident and then removed it, but it looks like some changes were automatically made to the ticket. Sorry!

annierm18:
@benelan and @geospatialem I added the installed label by accident and then removed it, but it looks like some changes were automatically made to the ticket. Sorry!

benelan:
This is fixed on `next`, although I uncovered a similar bug where `input-date-picker`'s internal input doesn't update when changing locales #5885

I'll work on that one next.

geospatialem:
Verified in `next.668`

![image](https://user-images.githubusercontent.com/5023024/206813929-93f753a9-b924-4200-bfc6-6d6f39ccdd3e.png)

PR #5887, opened 2022-12-03
fix(input-time-picker, input-date-picker): picker UI and state updates when changing locales

benelan:
**Related Issue:** #5855 

## Summary
The effectiveLocale is kept up to date using a MutationObserver, so it doesn't need to be passed down via attribute.

The `input-date-picker` e2e test I added for coverage is kinda funky, but I couldn't think of another way to make sure the internals are being localized. If you have any other ideas I'm happy to add to or change it.


driskull (approved):
Looks good! 👍

PR #5889, opened 2022-12-05
refactor(combobox,combobox-item)!: remove deprecated properties ,events & event payload. 

anveshmekala:
BREAKING CHANGE: Removed deprecated properties, events, and event payload.

calcite-combobox :
- Removed the property `active`, use `open` instead.
- Removed the event`calciteLookupChange`,use `calciteComboboxChange`  event instead.
- Removed the payload information from the event`calciteComboboxChipDismiss`, use the `value` property on the component to determine the removed value instead.

calcite-combobox-item:
- Removed the property `constant`, use `filterDisable` instead.


driskull (approved):
Thanks @anveshmekala! Just make sure you search for all occurances.

PR #5895, opened 2022-12-05
feat(action): add built-in translation support for indicator text

anveshmekala:
**Related Issue:** #4813 

## Summary

This PR will enable built-in translations for `indicator` text in `calcite-action`.


driskull:
@macandcheese thats why they will be able to override it with their own text via a prop.

The doc does say this:

```
/**
   * When `true`, indicates unread changes.
   */
  @Prop({ reflect: true }) indicator = false;
```

driskull (approved):

macandcheese:
I know it can be overridden, I just don't think that's an appropriate default string. The doc could be changed as well. 

"Unread changes" is pretty opinionated. That said, I don't really have a better suggestion besides "Indicator active" or "Indicator displayed" or something more generic.

driskull:
I'm fine with `Indicator present`. Just `indicator` is pretty vague.

anveshmekala:
So `indicator present` is the one we are going with right ?

cc: @geospatialem , @macandcheese , @driskull

driskull:
Fine with me.

geospatialem:
> Small thing, can we update the property description as well, here or in a separate PR? Maybe, "When true, display a visual indicator" or something like that? cc @geospatialem for description verbiage :)

Good catch @macandcheese - we had the original `intlIndicator` string listed as well. Added a new PR: https://github.com/Esri/calcite-components/pull/5938 for inclusion.

Issue #5896, opened 2022-12-05
flow inside flow not stacking inner flow-items


AdelheidF:
### Actual Behavior

Please see the codepen below. I added 2 shell panels that have a flow inside a flow so we can compare. First look at the end panel and you see that my inner flow doesn't stack the flow-items properly. In the start panel I 'fix' this by commenting out the flow-item of the outer flow.

I tried adding the inner flow programmatically but this didn't change anything.

cc @driskull 

### Expected Behavior

inner flow stacks flow-items 

### Reproduction Sample

https://codepen.io/afreitag/pen/rNKQOrR

### Reproduction Steps

Open codepen and look at right hand panel

### Reproduction Version

next.654

### Impact

high

Existing workflow for Map Viewer sketch, which seems to work currently, but I'm worried that this is an unplanned side effect and will go away.

### Esri team

ArcGIS Map Viewer

AdelheidF:
Looks good with next.659. Thanks for fixing so quickly!

PR #5897, opened 2022-12-05
fix(flow): Allow nested flow-items.

driskull:
**Related Issue:** #5896

## Summary

fix(flow): Allow nested flow-items.

Allows nesting `flow`s inside other `flows` and the flow should only query items that are not nested inside another flow.


driskull:
@benelan Can you review this one?

benelan (approved):
Can you add an e2e test of story, unless I'm missing existing coverage

PR #5904, opened 2022-12-06
refactor(panel): Remove messages related to deprecated properties.

anveshmekala:
**Related Issue:** #

## Summary

This PR will removed `back` from message bundles which is no longer required as `showBackButton` prop is removed from `calcite-panel` #5823.


driskull (approved):
👍

PR #1410, opened 2020-12-16
feat(color, color-hex-input, color-swatch): add support for no-color

jcfranco:
**Related Issue:** #917

## Summary

This enhances `calcite-color` to allow it to be unset by using the `allowEmpty` property. 

When `allowEmpty` is enabled, users can clear the color by either setting the `value` attribute_ as `""` OR by setting the `value` _property_ as `null`. Also, when working with hex/RGB/HSV inputs, the color can be cleared by clearing an input and committing (Enter or blurring the input).

Additional notes:

### `calcite-color`

* supports `null` as a value (toggled via `allowEmpty`)
* backdoor `color` prop will also be `null` when empty
* value nudging is not allowed when empty
* saving/removing when empty is now allowed
* hex, RGB/HSV inputs will display empty when there is no color set
* thumbs from color field and hue slider will not be rendered until a new value is selected

### `calcite-color-hex-input`

* supports `null` as a value (toggled via `allowEmpty`)

## Preview

color-empty

**Note**: color swatch no longer renders when there is no color.

## Questions

* ~Is `allowEmpty` a good name for this behavior? When I write code or doc to describe this, I tend to use 'no color'. Originally, the prop was named `allowNoColor`, but I thought that may read as do not allow any colors.~ Proceeding w/ `allowEmpty`.

cc @bstifle @pemberdom @subgan82

PR #5908, opened 2022-12-06
refactor(input)!: remove deprecated props and event payloads

benelan:
BREAKING CHANGE: remove deprecated properties and event payload    

- Removed `maxlength` property, use `maxLength` instead.
- Removed `locale` property, use `lang` instead.
- Removed `calciteInputInput`'s  `el`/`value` event payload properties, use the event's `target`/`currentTarget` instead.

### DIFF
--- a/src/components/color-picker/color-picker.tsx
+++ b/src/components/color-picker/color-picker.tsx
@@ -972,16 +972,16 @@ export class ColorPicker implements InteractiveComponent, LoadableComponent {
 
   private renderChannelsTabTitle = (channelMode: this["channelMode"]): VNode => {
     const { channelMode: activeChannelMode, intlRgb, intlHsv } = this;
-    const active = channelMode === activeChannelMode;
+    const selected = channelMode === activeChannelMode;
     const label = channelMode === "rgb" ? intlRgb : intlHsv;
 
     return (
       <calcite-tab-title
-        active={active}
         class={CSS.colorMode}
         data-color-mode={channelMode}
         key={channelMode}
         onCalciteTabsActivate={this.handleTabActivate}
+        selected={selected}
       >
         {label}
       </calcite-tab-title>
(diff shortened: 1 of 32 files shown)


src/components/color-picker/color-picker.tsx
```diff
@@ -482,7 +481,6 @@ export class ColorPicker implements InteractiveComponent, LoadableComponent {
     }
 
     input.value = inputValue;
-    internalInput.value = inputValue;
```
benelan, line 485: @eriklharper do you remember why you were setting the value for both the internal input and the `calcite-input`? None of the e2e tests are failing :sweat_smile:

benelan: Nevermind sorry Erik, I was git blaming the wrong line. It's from #1410

@driskull was I not supposed to remove the whole event payload? Or I guess I could go into the shadowRoot...

driskull: @eriklharper do you know if the whole event payload is deprecated or just the value?

I would say whole thing if a user can still get all the information.

driskull: I'm not sure why we were exposing the native input event.

eriklharper: > @eriklharper do you know if the whole event payload is deprecated or just the value?

I think the whole event payload as long as the values sent to the listener can all be retrieved by the event target object?

benelan: It's bad practices, but in order to remove the whole event payload I would need to go into  `calcite-input`'s shadowRoot to set the internal input's value. When Franco gets back I can ask him why we needed to do that and if it can be removed. 

The alternative is not setting the value on the internal input (what this PR is currently doing), and hoping it doesn't break anything lol.

Any preferences @driskull / @eriklharper ?

driskull: @eriklharper do you know why its setting it like this? 

@benelan lets wait for franco on this one. For now, lets leave it alone and add a note to revisit via a comment with `@deprecated`

driskull (approved):
👍

Issue #5898, opened 2022-12-05
`calcite-block` has no padding on the content slot


ethanbdev:
### Actual Behavior

The calcite block lost padding - it used to have `0.5rem` by default and now has none, so content goes right up against the edge. Setting `--calcite-block-padding` has no effect.

### Expected Behavior

There should be some way to add padding to this slot without adding it to each inidividual thing put into the slot. Beta.98 codepen: https://codepen.io/eborgen/pen/WNyLzNx?editors=100

### Reproduction Sample

https://codepen.io/eborgen/pen/ZERVxEJ?editors=1000

### Reproduction Steps

Open the codepen and observe no padding. Try adding the css variable to the block element, and see it has no effect.

### Reproduction Version

next.654

### Regression?

beta.98

### Impact

High!

### Esri team

ArcGIS Map Viewer

nel11211:
Was just wondering about this myself, tried setting `--calcite-block-padding` but no luck so far.

driskull:
@geospatialem can we get this one triaged and assigned? It seems like a simple fix.

geospatialem:
> @geospatialem can we get this one triaged and assigned? It seems like a simple fix.

@driskull Triage'd 👍🏻

driskull:
@geospatialem unassigned myself. Hoping another dev can grab it.

geospatialem:
Could be a regression as a result of `.content--spaced`, which is the css class that uses `--calcite-block-padding` from https://github.com/Esri/calcite-components/pull/5815/files#diff-ef7c4f2da3f37de2979bfaf2d13e5c0abc7cb9f1928e3f1dd8729872ebc381c9L140-L144

driskull:
Installed.

geospatialem:
Verified on `next.660`.

![image](https://user-images.githubusercontent.com/5023024/206044538-c1fb14dd-59e0-4ba0-abeb-76a7f7ca829c.png)

driskull:
@AdelheidF looking at that codepen in `beta98` and `next` and they look the same. Is that correct?

AdelheidF:
Yeah, the codepen is not right. Sorry, I can't find a test case.

padding-inline doesn't seem to fallback to 0.625rem.
![image](https://user-images.githubusercontent.com/44481292/206304043-96e0b294-831c-4aa3-b151-7541caeac106.png)

removing the variable makes it work.
[...]

PR #5918, opened 2022-12-06
fix(block): Fix content spacing. #5898

driskull:
fix(block): Fix content spacing. #5898

### DIFF
--- a/src/components/block/block.scss
+++ b/src/components/block/block.scss
@@ -142,6 +142,11 @@ calcite-handle {
   grid-area: control;
 }
 
+.content--spaced {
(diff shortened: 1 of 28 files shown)


src/components/block/block.scss
```diff
@@ -142,6 +142,11 @@ calcite-handle {
   grid-area: control;
 }
 
+.content--spaced {
```
anveshmekala, line 145: I don't think we need a separate class here. we can use `.content`  itself.

anveshmekala (approved):
LGTM!

src/components/block/block.scss
anveshmekala: nitpick ^

PR #5921, opened 2022-12-06
refactor(switch)!: Remove deprecated property and event payload.

driskull:
BREAKING CHANGE: Removed deprecated `switched` property and `calciteSwitchChange` event payload.

- Removed the property `switched`, use `checked` instead.
- Removed the `event.detail` from `calciteSwitchChange`, use `event.target.checked` instead.


benelan (approved):
:+1:

PR #5922, opened 2022-12-06
refactor(tabs, tab-nav, tab-title, tab)!: Remove deprecated properties and values.

driskull:
BREAKING CHANGE: Removed deprecated properties and values.

- Removed the property `active` from `calcite-tab-title`, use `selected` instead.
- Removed the property `active` from `calcite-tab`, use `selected` instead.
- Removed the `above` value from the `position` property on `calcite-tabs`, use `top` instead.
- Removed the `below` value from the `position` property on `calcite-tabs`, use `bottom` instead.


benelan (approved):
:tada:

PR #5935, opened 2022-12-07
refactor(handle)!: Remove deprecated event payload property on `calciteHandleNudge`.

driskull:
BREAKING CHANGE: Removed deprecated event payload property on `calciteHandleNudge`.

- Removed the `event.detail.handle` property on the event `calciteHandleNudge`, use `event.target` instead.


benelan (approved):
:rocket: I merged master to fix the e2e issue and the previous chromatic build looked good

PR #5936, opened 2022-12-07
refactor(button)!: Remove deprecated properties and values.

driskull:
BREAKING CHANGE: Removed deprecated properties and values.

- Removed the property `form`, this property is no longer needed if the component is placed inside a form.
- Removed the value `clear` from the property `appearance`, use the value `minimal` instead.


driskull:
@anveshmekala can you comment on why `"clear"` is replaced by `"minimal"`? I would expect it to be replaced by `"transparent"`. https://github.com/Esri/calcite-components/commit/29e2811feffa7ed3fcfddaf23ec2c129bd7a712d#diff-80c708f46f5ec49cbd3566c30ea3ab15523be503141c6c42713918a32febe1edR405

Issue #5312, opened 2022-09-13
[Rating] Unexpected component behavior after setting value as user


macandcheese:
### Actual Behavior

As an end-user, after interacting with the rating, there is unexpected behavior that occurs after setting (or clearing) a user-set value.

https://user-images.githubusercontent.com/4733155/189963194-da8ee607-f4a8-46bf-8e21-9b36d2aeef64.mov

- If I've set, or set and cleared, the component will reset on a subsequent keyboard-invoked focus.
- I cannot immediately use space or enter to set the value again, I need to navigate off and back to the desired rating.

### Expected Behavior

- I would not expect subsequent component focuses to reset the value, once I as a user have set it.
- I would expect to be able to toggle between set / clear value while remaining focused on the same rating. 

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/ZEoBWGj?editors=100

### Reproduction Steps

For clear issue:
- Open codepen
- Set a value on a rating
- Use keyboard to leave / return to component focus
- Rating will be cleared

For not being able to "re select" issue:
- Open codepen
- Set a value on a rating
- Use keyboard to clear value
- Attempt to use same key to set value again
- Value will not set
- Navigate to an adjacent star element, and return
- Value will be set-able with the same key.

### Reproduction Version

91

driskull:
@macandcheese the codepen seems to be incorrect for the issue.

macandcheese:
> @macandcheese the codepen seems to be incorrect for the issue.

Wrong link - updated. It's just a single instance of a rating component, but it illustrates the issue.

driskull:
I'm not sure we should allow toggling between a set value and no value when hitting enter or clicking. Radio buttons don't usually work like that.

If we want, maybe we could support the `escape` or `delete` key to reset to no value?

driskull:
I'm also not quite sure while hitting enter or space needs to emit an event. It seems like a user already clicked or navigated to the value with the arrow keys and the emit has occurred. Any idea on why we need to emit again on enter or space?

macandcheese:
Agree on not needing the extra events while navigating. I do think enter / space / click again could be expected to clear the rating, but will defer to @geospatialem there for a11y.

driskull:
> I do think enter / space / click again could be expected to clear the rating

Can you elaborate on why? These are radio buttons and I don't think there is a way to clear radio buttons

macandcheese:
[...]

PR #5948, opened 2022-12-08
fix(rating): 5312 improve user interface

alisonailea:
**Related Issue:** #5312

## Summary

Resolves bugs keyboard interaction and focus-in/focus-out
Cleans up component events, state, methods, and props
Refactors the component to make the value prop the source of truth and driver of the component state
Refactor tests to remove duplicates and improve scan-ability.

### Before
https://user-images.githubusercontent.com/4733155/189963194-da8ee607-f4a8-46bf-8e21-9b36d2aeef64.mov

### After
https://user-images.githubusercontent.com/3362490/208055841-20e6630a-c2ab-43b6-815d-18f14b0cd911.mp4


src/components/rating/rating.tsx
```diff
@@ -53,6 +52,20 @@ export class Rating
   /** The component's value. */
   @Prop({ reflect: true, mutable: true }) value = 0;
 
+  @Watch("value")
+  setValue(newValue: number, oldValue: number): void {
+    if (newValue == oldValue && !this.required) {
```
driskull, line 57: I don't think `newValue == oldValue` will ever be true since the `@Watch` only fires on changes to the newValue. So this if statement probably will never be entered.

src/components/rating/rating.tsx
```diff
@@ -92,6 +105,13 @@ export class Rating
    */
   @Prop({ reflect: true }) required = false;
 
+  /**
+   * Set the maximum number of stars
+   *
+   * @default 5
+   */
+  @Prop() max = 5;
```
driskull, line 113: Do we want to introduce a property here? I think we always want 5 stars for consistency within the design system. Maybe this could be a private variable?

src/components/rating/rating.tsx
```diff
@@ -101,6 +121,7 @@ export class Rating
   connectedCallback(): void {
     connectLabel(this);
     connectForm(this);
+    this.renderArray = Array.from({ length: this.max }, (_, i) => i + 1);
```
driskull, line 124: nitpick: maybe this array could be called `renderStars`?

driskull (commented):
Looking good @alisonailea! Just had some feedback/comments. Let me know when you want another round of reviews :D

src/components/rating/rating.tsx
alisonailea: right. that was from when I had this code in updateValue. thanks for the catch. fixed

src/components/rating/rating.tsx
alisonailea: I've moved this.max to private vars so this isn't an issue. I can update the name to be more descriptive.

src/components/rating/rating.tsx
```diff
@@ -203,28 +207,27 @@ export class Rating
...
-          onTouchEnd={() => (this.hoverValue = null)}
-        >
-          <legend class="visually-hidden">{intlRating}</legend>
+      <Host
+        onBlur={this.handleRatingFocusLeave}
+        onFocus={this.handleRatingFocusIn}
+        onKeyDown={this.handleKeyDown}
+        onPointerOut={this.handleRatingPointerOut}
+        onPointerOver={this.handleRatingPointerOver}
+        tabindex="0"
```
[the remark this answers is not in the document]
alisonailea: does setting the events listeners on the host interfere with users setting event listeners on the component too?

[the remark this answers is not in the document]
driskull: Nope that wont affect anything. These event listeners are added via addEventListener via JSX from my understanding.

src/components/rating/rating.tsx
```diff
@@ -238,27 +241,88 @@ export class Rating
...
+    //  when switching from keyboard to mouse this prop change is required to remove the focus styles from the element which are needed when navigating with a keyboard.
+    this.focusValue = null;
+  };
+
+  private handleKeyDown = (ev: KeyboardEvent) => {
+    if (ev.target["nodeName"] === "CALCITE-RATING") {
+      this.isKeyboardInteraction = true;
     } else {
-      this.focusValue = selectedRatingValue;
+      const inputVal = Number(ev.target["value"]);
```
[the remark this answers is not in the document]
alisonailea: I did try replacing ev.target with inputFocusRef. this creates a circular dependency error because inputFocusRef relies on this.value to understand it's current context.

src/components/rating/rating.tsx
```diff
@@ -238,27 +241,88 @@ export class Rating
...
+  private handleInputChange = (ev: Event) => {
+    if (this.isKeyboardInteraction === true) {
+      const inputVal = Number(ev.target["value"]);
+      this.focusValue = inputVal;
+      this.hoverValue = inputVal;
+    }
+  };
+
+  private handleLabelPointerDown = (ev: PointerEvent) => {
+    const inputVal = Number(ev.target["parentElement"]["nextElementSibling"]["value"]);
```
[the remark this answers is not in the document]
alisonailea: I did try replacing ev.target with inputFocusRef. this creates a circular dependency error because inputFocusRef relies on this.value to understand it's current context.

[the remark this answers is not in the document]
driskull: Lets create another private var for the input element that doesn't rely on the value. Then the dom structure isn't so dependent for this to work.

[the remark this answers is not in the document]
alisonailea: How often is the internal shadow DOM structure going to change?  I could do a complex function where I get the id of the input from the htmlFor and then query the DOM but that's a much more expensive query than simply navigating the DOM object that we already have available. I did simplify the DOM so the object traversal is simplified.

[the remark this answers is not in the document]
driskull: > How often is the internal shadow DOM structure going to change

Probably not often but it does happen.

I think its best if we just refactor this to not depend on the value when setting the `inputFocusRef`

```
ref={(el) =>
              (i === 1 || i === this.value) && (this.inputFocusRef = el as HTMLInputElement)
            }
```

Wherever the current code depends on the `inputFocusRef`, we should just add the if statement there.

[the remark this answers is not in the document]
alisonailea: > Wherever the current code depends on the `inputFocusRef`, we should just add the if statement there.

I don't know what you mean here. The inputFocusRef is useful for React users and in the case of focusIn where the inputFocusRef has already been correctly assigned after a repaint. Because inputFocusRef is assigned on repaint and not before, it can not  be used as the basis for new value assignment. While I could refactor this to assign the inputFocusRef as a part of the updateValue method, this would also require a DOM traversal and would be less performant. For these event listeners set on an element other than the input itself which update the component value , a DOM traversal is the most performant solution to getting the correct new value.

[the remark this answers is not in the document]
driskull: Sorry, maybe I'm not understanding correctly.

In general, we have been preferably been storing DOM elements after they are created using the `ref` attribute. This is preferred because when the element has been created by the host component, we can just store a reference to it rather than have to query the internal shadowRoot for it (which is less performant) or have to navigate a specific DOM structure that may change or be altered by future feature updates. So it would be preferable to store a reference to the element when its created.

From my understanding of the stencil lifecycle, the elements are created before rendering in the `componentWillRender` lifecycle method and the same element is used on repaint and modified as necessary.

I think the issue here is that the `inputFocusRef` changes depending on the value. We should probably just store each input element so we have a reference to them as needed. 

Maybe `inputFocusRef` could be changed to an array of elements? `inputFocusRef: HTMLInputElement[]`?

src/components/rating/rating.tsx
```diff
@@ -238,27 +254,115 @@ export class Rating
...
-      this.focusValue = selectedRatingValue;
+  };
+
+  private handleRatingFocusLeave = (): void => {
+    this.focusValue = null;
+    this.hoverValue = null;
+    this.hasFocus = false;
+  };
+
+  private handleKeyDown = (event: KeyboardEvent) => {
```
driskull, line 301: Can we separate this function into two functions. One for `handleComponentKeyDown` on the Host and one for `handleInputKeyDown` on the input? The host element can only set `isKeyBoardNavigation`?  Then we don't need to check the currentTarget anymore. I'm not sure why the same function is being called for both elements because they do different things.

src/components/rating/rating.scss
```diff
@@ -32,60 +32,68 @@
...
+  display: flex;
+  border-width: 0;
+  padding: 0;
+  margin-inline-end: var(--calcite-rating-spacing-unit);
+  align-items: center;
 }
 
 .wrapper {
-  @apply inline-block;
+  display: inline-block;
```
driskull, line 45: Can we continue to use `@apply inline-block`? For consistency with other components. We try to use tailwind if possible for now. We may be able to get rid of tailwind at a later time.

Applies to other CSS changes in this file.

src/components/rating/rating.tsx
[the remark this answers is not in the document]
alisonailea: Thanks for hoping on a call! As we discussed I have refactored the DOM to make DOM traversal simpler. I had added inputFocusRef on the focusIn event handler with a fallback to traverse the shadowDOM if inputFocusRef is unassigned for some edge case reason. For other event handlers simple DOM traversal (without shadowDom query) is the most performant option.

src/components/rating/rating.tsx
```diff
@@ -238,27 +254,115 @@ export class Rating
...
+    this.hoverValue = null;
+    this.focusValue = null;
+    this.hasFocus = false;
   };
 
-  private onFocusChange = (selectedRatingValue: number): void => {
+  private handleRatingFocusIn = (): void => {
+    let selectedInput = this.inputFocusRef;
+
+    if (!selectedInput) {
```
driskull, line 271: Instead of traversing the shadowRoot, my suggestion would be to have two private arrays where we could store references to the input and label elements. Then we don't need to have this conditional `(!selectedInput)` logic that has to traverse the DOM.

```
starInputs: HTMLCalciteInputElement = []
starLabels: HTMLCalciteLabelInputElement = []
```

On render, use the `ref` attribute to store them in the array. 

```
ref={(el) =>
              starInputs[i] = el as HTMLInputElement)
            }
```

Then, all references to inputs and labels are available at all times and the DOM doesn't need to be traversed.

driskull: @benelan @eriklharper what do you think? If you're ok with querying the shadowRoot here then we can go with this.

I was just thinking that we are rendering the elements so we should just store references to them.

driskull (commented):
Looking good @alisonailea.

@eriklharper @benelan can you review this one as well?

src/components/rating/rating.tsx
alisonailea: > Instead of traversing the shadowRoot, my suggestion would be to have two private arrays 

I've added starInputs array to the rendering but I don't see a use-case for starLabels

src/components/rating/rating.scss
alisonailea: no. I've already spoken with Franco about tailwind. This @apply is an anti-pattern in the tailwind docs itself and we will need to start moving away from it in the future. I will not be adding tech dept to this work.

src/components/rating/rating.tsx
```diff
@@ -3,27 +3,32 @@ import {
...
 import {
   setUpLoadableComponent,
   setComponentLoaded,
   LoadableComponent,
   componentLoaded
 } from "../../utils/loadable";
+import { Star, StarIconProps } from "./interfaces";
+
+const StarIcon: FunctionalComponent<StarIconProps> = ({ icon, scale, partial }) => (
+  <calcite-icon aria-hidden="true" class={partial ? undefined : "icon"} icon={icon} scale={scale} />
```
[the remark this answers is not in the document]
alisonailea: we should look into making all icons functional components in the future with a dynamic component wrapper.

alisonailea:
Was trying to figure out what PR refactor broke e2e tests so I ended up creating a new branch off master and cherry picking commits back onto it, fixing tests and commits as I went. This created a new commit history so I force pushed the new commit history to the branch.

benelan (commented):
The code LGTM. One thing I noticed is that shift tabbing back to a rating w/o a value focuses the first star, whereas previously it would focus the last star. Was that change intentional? I didn't see it mentioned in the linked issue.

driskull:
> The code LGTM. One thing I noticed is that shift tabbing back to a rating w/o a value focuses the first star, whereas previously it would focus the last star. Was that change intentional? I didn't see it mentioned in the linked issue.

I think ideally it should focus the star that was last value selected. If none were selected, it should be the first star.

It should act like a roving tabindex where only 1 star has the tabindex at any time. https://www.freecodecamp.org/news/html-roving-tabindex-attribute-explained-with-examples/

src/components/rating/rating.tsx
```diff
@@ -3,27 +3,38 @@ import {
...
   setComponentLoaded,
   LoadableComponent,
   componentLoaded
 } from "../../utils/loadable";
+import { Star, StarIconProps } from "./interfaces";
+
+const StarIcon: FunctionalComponent<StarIconProps> = ({ icon, scale, partial }) => (
+  <calcite-icon
+    {...{
+      "aria-hidden": "true",
```
driskull, line 31: "aria-hidden" isn't needed here as `calcite-icon` already sets it to true.

src/components/rating/rating.tsx
```diff
@@ -3,27 +3,38 @@ import {
...
-import { isActivationKey } from "../../utils/key";
 import {
   setUpLoadableComponent,
   setComponentLoaded,
   LoadableComponent,
   componentLoaded
 } from "../../utils/loadable";
+import { Star, StarIconProps } from "./interfaces";
+
+const StarIcon: FunctionalComponent<StarIconProps> = ({ icon, scale, partial }) => (
```
driskull, line 28: Actually, a functional component isn't really necessary here as the only thing that is really changing is the class name. The class name could just be applied on the `calcite-icon` component. I think the functional component would be more useful if it was setting the star icon based on a `full: boolean` variable?

src/components/rating/rating.tsx
```diff
@@ -131,100 +179,93 @@ export class Rating
...
+                      disabled={this.disabled || this.readOnly}
+                      id={id}
+                      name={this.guid}
+                      onChange={this.handleInputChange}
+                      onKeyDown={this.handleInputKeyDown}
+                      ref={(el) => {
+                        this.inputRefs[idx] = el;
+                        return (
+                          (value === 1 || value === this.value) &&
+                          (this.inputFocusRef = el as HTMLInputElement)
```
driskull, line 239: Do we still need `inputFocusRef` since we have a reference to all inputs now? Can we use `inputRefs[this.value]` wherever we were previously using `inputFocusRef`?

benelan:
> I think ideally it should focus the star that was last value selected. If none were selected, it should be the first star.

To clarify, if star 3 was selected and then cleared, do you mean it should focus star 3 or 1 when shift tabbing back? It currently focuses a selected star or star 1 if none are selected.

alisonailea:
> The code LGTM. One thing I noticed is that shift tabbing back to a rating w/o a value focuses the first star, whereas previously it would focus the last star. Was that change intentional? I didn't see it mentioned in the linked issue.

If there is no value but an average is set it defaults to the average. This was not part of the linked issue. I will remove it because it is out of scope but I would like @SkyeSeitz or @ashetland to weigh in on what is the best user experience as it seems jarring from a UX perspective to go from four yellow average stars to one selected blue star.

src/components/rating/rating.tsx
```diff
@@ -131,100 +179,93 @@ export class Rating
...
+                        );
+                      }}
+                      type="radio"
+                      value={value}
+                    />
+                    <StarIcon icon={selected || average ? "star-f" : "star"} scale={this.scale} />
```
driskull, line 245: If we do use a functional component here, would we move the logic for applying the `icon` into the functional component?

alisonailea:
> To clarify, if star 3 was selected and then cleared, do you mean it should focus star 3 or 1 when shift tabbing back?

In this case it should focus on Star 1, which is what it does.

alisonailea:
> I think ideally it should focus the star that was last value selected. If none were selected, it should be the first star.
> 
> It should act like a roving tabindex where only 1 star has the tabindex at any time. https://www.freecodecamp.org/news/html-roving-tabindex-attribute-explained-with-examples/

That's what it does.

driskull (commented):
Had some comments/questions. 👍

benelan:
Yup the keyboard nav matches the clarified, expected behavior.

I do agree that focusing the average (when present) makes sense when there is no selection. But it should probably go through design approvals first like you mentioned.

LGTM, I'll defer to Matt and Erik for final approvals.

src/components/rating/rating.tsx
alisonailea: inputFocusRef is still potentially important for react consumers?
I'm hesitant to remove it as it's legacy and I don't fully grok all the "in-the-wild" use cases yet

driskull: @alisonailea what do you mean react consumers? This is an internal property.

src/components/rating/rating.tsx
alisonailea: sorry I made you mention this twice. thanks for the catch

src/components/rating/rating.tsx
eriklharper: > I don't think newValue == oldValue will ever be true since the @Watch only fires on changes to the newValue. 

You would think, but I have seen it happen before.

driskull: If it does, its a stencil bug :D

src/components/rating/rating.tsx
eriklharper: React consumers using `ref` would obtain an instance to the `CalciteRating` element, not the internal `<input>` element inside the shadowRoot.  They should be able to do anything they need to do with that instance since the public API for the component would be exposed through that ref.

ashetland:
Okay folks, just walked through this one. When there is an average, but no value, tabbing into the rating should hide the average and focus on the first star. Hover behavior should follow the same pattern.

alisonailea:
> Okay folks, just walked through this one. When there is an average, but no value, tabbing into the rating should hide the average and focus on the first star. Hover behavior should follow the same pattern.

updated the PR per @ashetland's comment

rating-deselect-average

driskull:
Works great!

driskull (approved):
Looks good @alisonailea!

Only final suggestion would be to remove `inputFocusRef` but its only a nitpick.

alisonailea:
> remove inputFocusRef

I don't want to because I think it's needed for React support. Do we know that it's not?

driskull:
> I don't want to because I think it's needed for React support. Do we know that it's not?

Its definitely not needed for react support. React has no reference to our stencil code. Once built, there is no `ref`.

alisonailea:
> Its definitely not needed for react support. React has no reference to our stencil code. Once built, there is no `ref`.

I know. That's my concern. Due to the differences between the React virtual DOM and the web component DOM there are edge cases where changes to the internals of the component will not be captured by a wrapping React app. Providing the `inputFocusRef` allows the consuming app to use it to help track changes. I'm assuming that's why it was introduced in the first place.

benelan:
How would the consuming app use `inputFocusRef`?

driskull:
>  Due to the differences between the React virtual DOM and the web component DOM

When react is using our components, our components are compiled into JS files using vanilla web components so there is no JSX/virtualDOM at that time.

driskull:
> Providing the inputFocusRef allows the consuming app to use it to help track changes. I'm assuming that's why it was introduced in the first place.

It wasn't introduced for any consumer at all, Its just there to get a reference to the element that we are creating in VDOM using Stencil.js

alisonailea:
> How would the consuming app use inputFocusRef?

I don't know for sure. why was it introduced in the first place?

driskull:
https://stenciljs.com/docs/templating-jsx#getting-a-reference-to-a-dom-element

alisonailea:
> https://stenciljs.com/docs/templating-jsx#getting-a-reference-to-a-dom-element

Yes I completely understand that. I'm not confused about how stencil works. That's not my question. Never mind, I'll just remove it.

driskull:
Hey @alisonailea I'd like to understand the concerns so we're not just removing something. Maybe you can elaborate on the following...

> there are edge cases where changes to the internals of the component will not be captured by a wrapping React app

I'm not following how react would be able to see/monitor the internals of the compiled component here. I'm not sure react could or should react to our internals.

benelan:
> Yes I completely understand that. I'm not confused about how stencil works. That's not my question. Never mind, I'll just remove it.

I think he was saying the reason the ref was added is because it is recommended in the documentation. It didn't have anything to do with React

alisonailea:
I will not be removing inputFocusRef. It is used by setFocus and removing it breaks most of the common tests. There may be a work around but because it is still in use in other places in the component and I'm still not sure why it was added to begin with I have concerns. Removing this can be addressed in a future refactor of which this component already needs.

eriklharper:
> Due to the differences between the React virtual DOM and the web component DOM there are edge cases where changes to the internals of the component will not be captured by a wrapping React app.

No changes inside a Stencil component's shadow dom would be picked up by React (afaik) since React only responds to virtual DOM it manages via its own render methods, which is always regular (non-shadow) DOM.  React knows nothing about the DOM internals of shadowed components (as it should) so there is no need to think about how vdom consumers like React will respond to internal shadow dom changes.  Any internal change in the component that React needs to know about should be communicated via events.  Unless I'm misunderstanding the concern...

I understand better your concern with removing `inputFocusRef`.  It's still needed to determine which invisible radio input (star) should gain focus when the `setFocus` method is used by a React or other consumer of the component.  I agree we need to keep it in there so we don't break our focus method.

driskull:
> It's still needed to determine which invisible radio input (star) should gain focus when the setFocus method is used by a React or other consumer of the component

I was just saying it's not really needed because we have `this.inputRefs` now. `setFocus` can just use `this.inputRefs` by focusing the correct one to focus.

eriklharper:
> I was just saying it's not really needed because we have `this.inputRefs` now. `setFocus` can just use `this.inputRefs` by focusing the correct one to focus.

Gotcha, makes total sense.

eriklharper:
> When react is using our components, our components are compiled into JS files using vanilla web components so there is no JSX/virtualDOM at that time.

JSX is removed for sure, but the virtual DOM is in the production code.

Issue #5882, opened 2022-12-02
Infinite loop when calling `setFocus` on new `calcite-input`


nwhittaker:
### Actual Behavior

Programmatically inserting a `<calcite-input>` into a `<calcite-modal>` and calling `setFocus()` on the input results in an infinite loop that locks up the window.

### Expected Behavior

Programmatically inserting a `<calcite-input>` into a `<calcite-modal>` and calling `setFocus()` on it shifts focus to the input.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/VwdEody

### Reproduction Steps

1. Visit the sample.
2. Wait for the modal to open and click the **Add input** button.
3. Notice the preview area locks up after a few seconds, and the browser presents some sort of **Page Unresponsive** dialog after awhile.

### Reproduction Version

next.653

### Relevant Info

The regression appears to have been introduced over a couple releases:
1. `next.632`: #5676
   * Browser does not lock up, but new input elements do not gain immediate focus. 
3. `next.636`: #5749
   * Browser locks up when setting focus on the new input element.

The likely root of the problem is the new focus trap library isn't aware of the new input element yet, so it thinks an element outside of the modal is being focused. The loop that emerges is:
1. `setFocus()` called on the new `<calcite-input>`.
2. `focus()` called on the shadow `<input>`.
3. The focus traps's `focusin` handler reverts focus to the previously focused element.
4. The `<calcite-input>` element's `focus` handler runs and calls `setFocus()` on itself which leads back to step 1.

### Regression?

next.631

### Impact

The Field Maps web app uses a modal for editing coded value domains. When a user wants to add a new coded value, a new row of inputs is inserted and focus should shift to the row's first input. This is blocking us from upgrading Calcite beyond `next.631`.

### Esri team

ArcGIS Field Apps

benelan:
Yeah, now this one isn't a doc cleanup issue 😬 

Triaging to the current sprint, cc @driskull @geospatialem

driskull:
I think this might be because when the modal is opened it is trapping focus. We might need to provide an option to disable focus trap for modal here.

driskull:
@eriklharper can you take this one? It seems like an issue with `calcite-input`. It works fine if a native input is used.

I narrowed it down to a infinite loop in setFocus/input

When `setFocus()` is called it focuses on the child element.

[...]

PR #5961, opened 2022-12-09
fix(input, input-number, input-text): Fix infinite loop crashing browser. #5882

driskull:
**Related Issue:** #5882

## Summary

- Adds click handler to focus on internal inputs
- Focus handler only emits event


eriklharper:
@driskull these changes look good, about the same as what I have (so far) come up with for resolving the infinite loop bug, but I still don't know how to get the `setFocus` call to work right after the programmatic insertion of the input element.  Let's sync up on this maybe tomorrow.  Focusing with the keyboard works fine, but for some reason clicks into the input don't focus the element and the `setFocus` calls don't work either.  Very strange.

driskull:
> I still don't know how to get the setFocus call to work right after the programmatic insertion of the input element.

Yeah, I think something with focus-trap isn't allowing this. Maybe we can do this via a follow-up.

I will add a way to disable the focus trap for Modal.

driskull:
@eriklharper can we get this one installed today?

eriklharper (approved):
> @eriklharper can we get this one installed today?

Yes, let's do it.  Approved :)

geospatialem:
> I will add a way to disable the focus trap for Modal.

@driskull Do you have an open PR for the `modal` focus trap disable? Also noticed the behavior when trying to verify the related issue.

driskull:
Yeah here's the PR: https://github.com/Esri/calcite-components/pull/5965

PR #5962, opened 2022-12-09
refactor(time-picker)!: Remove deprecated locale property

eriklharper:
BREAKING CHANGE: Removed the `locale` property, use `lang` instead.


driskull (approved):
👍

PR #5965, opened 2022-12-09
fix(modal, popover): Add `disableFocusTrap` property to toggle focus trapping.

driskull:
fix(modal, popover): Add `disableFocusTrap` property to toggle focus trapping.


benelan (approved):
LGTM. Would it be possible to make an e2e test for this that wouldn't crash everything if it fails? It can be in another PR if getting this in for MV is urgent so they can test on `next`

driskull:
> LGTM. Would it be possible to make an e2e test for this that wouldn't crash everything if it fails? It can be in another PR if getting this in for MV is urgent so they can test on next

That was a different PR, but I think `input` would need a test for that.

eriklharper (commented):
LGTM!  Not sure how it might affect accessibility to disable focus trapping, but as far as implementation goes for enabling a feature like this, it is spot on!

benelan:
Ah, I thought this PR was needed to disable the focus trap in Input to fix the crashing issue. Could modal/popover have a test that opens the component, tabs a full circle with focus trap on, turns focus trap off, and then tabs until it gets to an element outside of the modal?

PR #5966, opened 2022-12-09
test(floating-ui): fix type errors

benelan:
**Related Issue:** #5939

## Summary
I kept getting type errors coming from `floating-ui.spec.ts` when trying to run the demos. After some quick maths I determined fixing them would be faster than restarting the dev server every 15 minutes

image

One of them was my fault from the PR linked above, I made the `placements` array a const so it could be used as a type but didn't check the spec tests. Not sure where the `referenceEl` error came from but renaming the param fixed it.


driskull (approved):
Thanks!

Issue #5917, opened 2022-12-06
[Inputs] Add `inputmode` and `enterKeyHint` property to Input Components


macandcheese:
### Description

We currently don't explicitly accept an `inputmode` property, although it is spread to the internally rendered component. As a result, it is not documented. We should add this as a public property.

### Acceptance Criteria

Add `inputmode` and `enterKeyHint` property to appropriate input-* components.

### Which Component

Input-*

### Esri team

Calcite (design)

macandcheese:
Might be worth adding `enterKeyHint` as well while we are at is, since that is similarly inferred from `this.el` currently...

geospatialem:
Adding this to the January 2023 release as a low priority issue to tackle. If we're unable to tackle it with other priorities, we will reevaluate the priority after release for the February or March release.

geospatialem:
Related to a [Community Idea](https://community.esri.com/t5/calcite-design-system-ideas/have-calcite-input-type-be-reflected-on-the/idi-p/1232286/jump-to/first-unread-message).

geospatialem:
Verified in [the following Codepen](https://codepen.io/geospatialem/pen/bGKPqLV) on `next.672` using the `input-mode` and `enter-key-hint` attributes.

Mobile verification on iOS `16.1.2`.

![image](https://user-images.githubusercontent.com/5023024/207401415-76762fa4-1bd2-4ac1-965f-870e1bbcbecd.png)
![image](https://user-images.githubusercontent.com/5023024/207401525-de45d4c2-2274-4a0e-9fca-fb3b416fe4ba.png)

PR #5976, opened 2022-12-09
feat(input, input-number, input-text): add inputMode and enterKeyHint properties

benelan:
**Related Issue:** #5917

## Summary

Adds [`inputMode`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode) and [`enterKeyHint`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/enterkeyhint) properties to `input`, `input-text`, and `input-number`


src/components/input-number/input-number.tsx
```diff
@@ -906,8 +925,8 @@ export class InputNumber
         autofocus={this.autofocus ? true : null}
         defaultValue={this.defaultValue}
         disabled={this.disabled ? true : null}
-        enterKeyHint={this.el.enterKeyHint}
-        inputMode={this.el.inputMode}
```
driskull, line 910: nice. We shouldn't have these this.el props anymore :D

src/components/input/input.tsx
```diff
@@ -280,6 +283,22 @@ export class Input
    */
   @Prop() multiple = false;
 
+  /**
+   * Specifies the type of content to help devices display an appropriate virtual keyboard.
+   * Read the native attribute's documentation on MDN for more info.
+   *
+   * @mdn [step](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode)
```
macandcheese, line 290: Should we be passing this @mdn tag through to the documentation site UI at all to indicate these are native html attributes?

PR #5977, opened 2022-12-09
refactor(input-date-picker,date-picker)!: Removing deprecated locale properties

eriklharper:
BREAKING CHANGE: Removed the `locale` property, use `lang` instead.


driskull (approved):
LGTM! 👍 💯

PR #5981, opened 2022-12-10
fix(loader): no longer animates when reduced motion is enabled

anveshmekala:
**Related Issue:** #3489 

## Summary

This PR will stop the loader from animating when reduced motion is enabled.


driskull (approved):
👍

alisonailea (approved):

PR #5983, opened 2022-12-10
feat(shell): Add slots for Modal and Alert

macandcheese:
**Related Issue:** 

Proposed as part of Shell / Layout enhancements: https://confluencewikidev.esri.com/pages/viewpage.action?pageId=222267264 (VPN needed)

## Summary
This is a draft PR to add two new slots to `calcite-shell`: `modal` and `alerts`. 

This pattern was proposed as part of our upcoming Shell / layout improvements, and would be leveraged by the upcoming proposed `calcite-sheet` component.

When expected components are slotted into named slots, conditional styles are applied to constrain their extent to that of the `calcite-shell` (including covering the existing `header` and `footer` slots). This is useful in many cases, specifically when the `calcite-shell` is embedded amongst other content, like tutorials, documentation pages, or any place the user experience is not a completely full-screen `calcite-shell`. 

When these components are _not_ slotted into the new `calcite-shell` slots - their behavior remains unchanged.

A demo is available locally at `/demos/shell/basic-slotted-elements.html`

cc @asangma @mitc7862 

Screen recording of proposed behavior: 

https://user-images.githubusercontent.com/4733155/207209472-000f38d5-d44c-4ead-a277-dd3fc5ef905c.mov


src/components/alert/alert.tsx
```diff
@@ -153,6 +153,8 @@ export class Alert implements OpenCloseComponent, LocalizedComponent, LoadableCo
       this.calciteInternalAlertRegister.emit();
     }
     connectOpenCloseComponent(this);
+    this.isSlottedInShell =
```
driskull, line 156: Can't the `shell` just add some styling in its CSS like...

```
::slotted(alert) {
  @apply absolute;
}
```

macandcheese: Alert is a simpler example, but the Modal required some internal configuration and I tried to keep the pattern consistent between the two. I also tried to keep the styles off of the Alert host and applied to an internal container, but I can re-evaluate this.

driskull: Yeah. In theory your logic should be fine if this is the only way. Its just we don't currently do this anywhere else.

driskull: Maybe we could introduce a property if necessary? Could be an internal property that only shell knows about.

Its just better if the parent controls its children rather than the children having to figure out their parent and do things.

macandcheese: @driskull I updated the pattern to use internal properties on `Alert` and `Modal`, which is set by the parent `Shell` - let me know if this looks better.

macandcheese: @driskull I updated the pattern to use internal properties on Alert and Modal, set by the parent Shell - let me know if this looks better.

driskull (approved):
Looks good @macandcheese! 🧀

src/components/shell/shell.tsx
```diff
@@ -14,7 +14,10 @@ import {
  * @slot panel-start - A slot for adding the starting `calcite-shell-panel`.
  * @slot panel-end - A slot for adding the ending `calcite-shell-panel`.
  * @slot center-row - A slot for adding content to the center row.
+ * @slot modal - A slot for adding a `calcite-modal`. When placed in this slot, the modal position will be constrained to the extent of the shell.
```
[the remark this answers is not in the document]
macandcheese: Yeah, they should be consistent one way or the other. I had used `alerts` since we kind of incentivize multiple with the `queued` display and only one `modal` is shown to a user at once, but I'll make them both plural.

src/components/shell/shell.tsx
```diff
@@ -54,16 +65,74 @@ export class Shell implements ConditionalSlotComponent {
...
+
+  handleFooterSlotChange = (event: Event): void => {
+    this.hasFooter = !!slotChangeHasAssignedElement(event);
+  };
+
+  handleAlertsSlotChange = (event: Event): void => {
+    this.hasAlerts = !!slotChangeHasAssignedElement(event);
+    slotChangeGetAssignedElements(event)?.map((el) => {
+      if (el.nodeName === "CALCITE-ALERT") {
+        (el as HTMLCalciteAlertElement).slottedInShell = true;
```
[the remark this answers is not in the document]
macandcheese: Will update

src/components/shell/shell.tsx
```diff
@@ -54,16 +65,74 @@ export class Shell implements ConditionalSlotComponent {
...
+      }
+    });
+  };
+
+  handleModalSlotChange = (event: Event): void => {
+    this.hasModal = !!slotChangeHasAssignedElement(event);
+    slotChangeGetAssignedElements(event)?.map((el) => {
+      if (el.nodeName === "CALCITE-MODAL") {
+        (el as HTMLCalciteModalElement).slottedInShell = true;
```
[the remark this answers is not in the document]
macandcheese: Will update

src/components/shell/shell.tsx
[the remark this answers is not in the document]
macandcheese: @driskull - updated to `modals` - going to merge.

PR #5990, opened 2022-12-12
test(chromatic): add delay and disable animation on some stories

benelan:
**Related Issue:** https://github.com/Esri/calcite-components/issues/4825#issuecomment-1335921930

## Summary
Adds a chromatic snapshot delay to the components giving false positives. I also disabled animation for some of the floating-ui components, but only the NoTest ones so it doesn't effect the hosted stories. If the delay doesn't help I'll set the diffThreshold to `1` for the specific stories that fail, until I can do some additional digging.


driskull (approved):
👍 Nice!

Issue #5993, opened 2022-12-12
popover/modal focus trap causes input elements not to get the focus set via mouse click


AdelheidF:
### Actual Behavior

popover/modal focus trap causes input elements, inside or outside of popovers, not to get the focus when you click on them via mouse click.

E.g. a popover is open (focus trap set by default), now I can no longer edit an input box in a shell panel.

cc @driskull 

### Expected Behavior

A mouse click anywhere in the app should set the focus on that element. Focus trap set for an open popover should not affect this.

### Reproduction Sample

https://codepen.io/afreitag/pen/XWzOjwK

### Reproduction Steps

Try to edit value 0 in the panel

### Reproduction Version

next.665

### Impact

very high

### Esri team

ArcGIS Map Viewer

AdelheidF:
looks good in my various codepens, will test in map viewer later

geospatialem:
Verified in `next.673`.

PR #5994, opened 2022-12-12
fix(popover, modal): deactivate focus trap on outside click

driskull:
**Related Issue:** #5993

## Summary

fix(popover, modal): deactivate focus trap on outside click.

https://github.com/focus-trap/focus-trap/blob/master/README.md


benelan (approved):
👍

Issue #5975, opened 2022-12-09
Alert: demo page for `alert` component appears to be broken and `buttons` no longer trigger the `alert` dialogues to open


Elijbet:
### Actual Behavior

`buttons` on the demo pages won't trigger the `alert` dialogues to open any longer

### Expected Behavior

Demo page for `alert` component works as expected, where `buttons` trigger the `alert` dialogues to open. 

### Reproduction Sample

N/A

### Reproduction Steps

N/A

### Reproduction Version

v1.0.0-beta.98

### Esri team

Calcite (dev)

geospatialem:
Verified on `master`. 💯 🔥

PR #6002, opened 2022-12-13
chore(alert): Fix local demo page attribute

macandcheese:
**Related Issue:** #5975 

## Summary
The local demo page was using the outdated `active` attribute instead of `open`.


driskull (approved):
👍

PR #6003, opened 2022-12-13
refactor(loader)!: remove deprecated properties

anveshmekala:

- Removed the property `active`, use global attribute `hidden` instead.
- Removed the property `noPadding`, use `--calcite-loader-padding` CSS property instead.


alisonailea (approved):

driskull (approved):
👍 Nice!

Issue #5798, opened 2022-11-22
Epic: Audit components and remove deprecations


geospatialem:
### Purpose
Conduct an audit across components, remove deprecations, and update/remove extraneous/failing tests and stories prior to the launch of 1.0.

cc @driskull

### Plan of action for each component
1. Remove deprecations cc #3781 
2. Audit of the following:
    - Props (or any deprecated values)
    - Methods (or any deprecated arguments)
    - Events, including removal of unnecessary (`event.detail`)
    - CSS variables
    - Named slots
    - Unused utilities/resources
3. Remove them. 
4. Once deprecations are removed:
    - Remove failing/extraneous Tests
    - Update Storybook Stories
    - Update usage readme files
    - Update readme file description of component if necessary.
    - Check for any components using removed deprecations and update them.
5. Add a note with the string "deprecated" to any shared files where deprecations need to be removed after this issue has been completed.

### Sample Pull Request

When submitting a PR for a component, please refer to https://github.com/Esri/calcite-components/pull/5799 for guidance.

### Sample commit message

```
refactor(action)!: Remove deprecated properties and values. #5798
```

### Sample PR Title

Note the component and what was removed. This can be properties, events, methods, arguments, values, CSS variables, slots.

```
refactor(action)!: Remove deprecated properties and values.
```

### Sample PR Body

Note the general changes and migration options. This can span multiple lines.

```
BREAKING CHANGE: Removed the `summary` and `disablePadding` properties.

- Removed the property `summary`, use `description` instead.
- Removed the property `disablePadding`, use the CSS variable `--calcite-block-padding` instead.
```

```
BREAKING CHANGE: Removed deprecated events and properties.

- Removed the property `dismissed`, use `closed` instead.
- Removed the property `dismissible`, use `closable` instead.
- Removed the property `summary`, use `description` instead.
- Removed the property `intlBack`, use the `calcite-flow-item` component instead.
- Removed the property `showBackButton`, use `calcite-flow-item` instead.
- Removed the property `beforeBack`, use the `calcite-flow-item` component instead.
- Removed the event `calcitePanelDismiss`, use `calcitePanelClose` instead.
- Removed the event `calcitePanelDismissedChange`, use `calcitePanelClose` instead.
- Removed the event `calcitePanelBackClick`, use the `calcite-flow-item` component instead.
```

```
BREAKING CHANGE: Removed the `primary-panel` and `contextual-panel` slots.

[...]

PR #6004, opened 2022-12-13
refactor(notice)!: remove deprecated properties

anveshmekala:
**Related Issue:** #5798

Removed the property `active`, use `open` instead.
Removed the property `dimissible`, use `closable` property instead.


driskull (approved):
👍

Issue #3127, opened 2021-09-28
Enhancement: [calcite-tree] Support actions/menu in tree items


ellenupp:
### Description
We would like to include a menu of actions in each tree item, ex: 
![image](https://user-images.githubusercontent.com/14304581/135178348-7963846b-a815-4f82-9fa9-341b39caa7f1.png)

Including a calcite-dropdown within a calcite-tree-item does not render in a usable way as the dropdown is constrained to the current tree level. It would be great to have a slot in the tree item to accommodate such a menu.
https://codepen.io/ellupp/pen/PojWvmR?editors=1100
![image](https://user-images.githubusercontent.com/14304581/135178285-633c6843-d2a1-47db-851f-1b45dee4eeb2.png)

### Acceptance Criteria 
Tree items can include a menu of actions/dropdown. When opened, the menu should render on top of all tree items such that it is entirely visible.

### Relevant Info 
Supporting a menu is the key enhancement needed but we also need to support multiple actions in some tree items. Slotting a calcite-button into the tree item works for this at the moment. Also, in our use case we only want the actions to be visible on hover or when focus is on/within the tree item. 
![image](https://user-images.githubusercontent.com/14304581/135178820-98de4acb-4a7a-449b-80bb-56bbb6f5e033.png)

#### Which Component
calcite-tree/calcite-tree-item

#### Example Use Case
Please contact me for link to a custom tree implementation in dashboards that has the desired functionality. 

cc @vijayendranj @julio8a

macandcheese:
Seems like this could adopt the action / content start / end slot paradigm to allow for these customizations, similar to a list item.

Elijbet:
@ellenupp 
Just noticed that all `tree-items` from the example I had from you earlier today have an `icon-start`. Is it needed for your use case, I could plan to add that as well if it is.

ellenupp:
Yeah that would be great, thanks.

geospatialem:
Verified in `1.0.3` with the [following Codepen](https://codepen.io/geospatialem/pen/rNrrLew) with the `iconStart` prop and `actions-end` slot.

<img width="403" alt="Screenshot 2023-01-28 at 9 59 41 PM" src="https://user-images.githubusercontent.com/5023024/215304161-710d2152-956c-403d-8560-33f88d93fe09.png">

ellenupp:
Getting around to trying this out- it looks great, thanks!

PR #6005, opened 2022-12-13
feat(tree-item): support actions-end and icon-start

Elijbet:
**Related Issue:** #3127

## Summary

This PR aims to enhance the `calcite-tree-item` with an `actions-end` feature to accommodate a dropdown menu.

- `calcite-tree-item` now supports `actions-end` slot for adding actions to the end of the component, with a recommendation to use two or fewer actions.
- Including a `calcite-dropdown` within a `calcite-tree-item` renders as expected.
- Includes `icon-start` per the use case described in the issue. 
- Adds a story to demo the use case.
- Screenshots to cover various scenarios.

Thanks @macandcheese for help with styling.


Elijbet:
@SkyeSeitz @ashetland We will need to update the Figma files to include this enhancement and the use case presented in the stories.

driskull (approved):
Looks good!! 🌟

Elijbet:
> Looks good! 

I have more work on it - just  opened to check screenshots. But it's close.

driskull:
Ah, you can open it as a draft PR next time so we know not to review it yet :)

Elijbet:
@SkyeSeitz @ashetland We will need to update Figma files to include this enhancement and the use case presented in the stories.

src/components/tree-item/tree-item.tsx
```diff
@@ -230,6 +241,22 @@ export class TreeItem implements ConditionalSlotComponent, InteractiveComponent
...
+    const slotNode = (
+      <slot
+        key="actionsEndSlot"
+        name={SLOTS.actionsEnd}
+        onSlotchange={this.actionsEndSlotChangeHandler}
+      />
+    );
+    const iconStartEl = (
+      <calcite-icon
+        class="calcite-tab-title--icon icon-start"
```
driskull, line 254: can you add these to the CSS const in the resources file and use them from that?

```jsx
class={{[CSS.tabTitle]: true, [CSS.iconStart]: true}}
```

driskull (approved):
Looks good, just a change for the CSS const.

src/components/tree-item/tree-item.tsx
Elijbet: Perfect!

SkyeSeitz:
Update on the design decisions thus far: 

**1. `icon-start` will have padding on each according to scale**

> s=4px
> m=8px 
> l=12px

image

**2. Slotted Actions in `actions-end` changing the height of the Tree Item**

Since Action comes with internal padding, they are taller than our current Tree Items. Initially, our proposal was to document using the small `Action` in the medium scale Tree Item in order to maintain its 32px height. This would also be reinforced in the UI Kit through prescribing that size on the medium as a default. 

image

However, though this works great for the medium scale, we run into issues with the small scale needing to use the small `Action` as well and therefore increasing its size to basically match the size of medium. 
image

Large is slightly less of an issue since it would only grow from 44px to 48px. But overall, we are still trying to avoid the Tree component from becoming more spread out between items. 

We will be addressing early next week to decide on the best direction.

Elijbet:
> Update on the design decisions thus far:
> 
> **1. `icon-start` will have padding on each according to scale**
> 
> > s=4px
> > m=8px
> > l=12px
> 
> image
> 
> **2. Slotted Actions in `actions-end` changing the height of the Tree Item**
> 
> Since Action comes with internal padding, they are taller than our current Tree Items. Initially, our proposal was to document using the small `Action` in the medium scale Tree Item in order to maintain its 32px height. This would also be reinforced in the UI Kit through prescribing that size on the medium as a default.
> 
> image
> 
> However, though this works great for the medium scale, we run into issues with the small scale needing to use the small `Action` as well and therefore increasing its size to basically match the size of medium. image
> 
> Large is slightly less of an issue since it would only grow from 44px to 48px. But overall, we are still trying to avoid the Tree component from becoming more spread out between items.
> 
> We will be addressing early next week to decide on the best direction.

To recap, we discussed this during the engineering meeting and the general practice is not to set any props on the slotted content. The reasoning is that slots are a way to provide means of inserting content, and we have no idea what will be slotted (it might not be our components that are slotted). Adam suggested demonstrating in Figma how the `dropdown` needs to be slotted in `action-end` to work with the parent scales, eg setting the scale="s" for the small `tree-item`. Instead of matching the scales, I will look into having the slots fill their container, per Skye's suggestion, following the set up in `Alert`.

src/components/tree-item/tree-item.tsx
```diff
@@ -266,11 +275,27 @@ export class TreeItem
...
       />
     ) : null;
 
     const hidden = !(this.parentExpanded || this.depth === 1);
+    const { hasEndActions } = this;
+    const slotNode = (
+      <slot
+        key="actionsEndSlot"
+        name={SLOTS.actionsEnd}
+        onSlotchange={this.actionsEndSlotChangeHandler}
```
benelan, line 288: should this be `onSlotChange` with a capital C?

src/components/tree-item/tree-item.tsx
```diff
@@ -218,12 +222,16 @@ export class TreeItem
 
   private isSelectionMultiLike: boolean;
 
+  /** Specifies the scale of the tree-item controlled by parent, defaults to m */
+  scale: Scale = "m";
```
benelan, line 226: It looks like the `@Prop` decorator is missing, maybe that's causing issues somehow?

src/components/tree-item/tree-item.scss
```diff
@@ -5,6 +5,10 @@
     cursor-pointer;
 }
 
+[hidden] {
+  @apply hidden;
+}
```
benelan, line 10: Do we need this? `hidden` is a global attribute so I think it should just work (famous last words)

benelan (commented):
Added a couple comments, not sure if any of them are causing the font size issue though. Are you seeing any build or runtime errors in the demos?

src/components/tree-item/tree-item.tsx
Elijbet: I left it out because we didn't want scale on children to be public. Is there a better way to do this?

src/components/tree-item/tree-item.tsx
Elijbet: I think maybe because the event that is fired on an [HTMLSlotElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement) instance when the node(s) contained in that slot change is called `slotchange`, as in one word, that's why we use `onSlotchange`. Which means I should be changing the handler name instead, to match.

src/components/tree-item/tree-item.scss
Elijbet: I remember having this discussion before with others and as it turns out even though you'd expect the global attribute "hidden" to have that behavior by default, it doesn't and needs to be set.

src/components/tree-item/tree-item.tsx
Elijbet: You were right about this causing an issue in that not reflecting it was causing the styling to not apply. I brought the prop back for the time being to address it later in a more relevant issue.

src/components/tree-item/tree-item.scss
jcfranco: > even though you'd expect the global attribute "hidden" to have that behavior by default, it doesn't and needs to be set.

Natively, `hidden` does hide the element, but in this case, we need an attribute selector so it can have a higher specificity and override the `display` prop set in the [`.actions-end` style here](https://github.com/Esri/calcite-components/pull/6005/files#diff-c7e860104581dc1e9bebfb6a5733bb56066510cf1e9abfb39102bbecec90ec3dR110) ([example](https://codepen.io/jcfranco/pen/abjYOoL)).

Sidebar: maybe we need to add a selector for `hidden` within our component's shadow DOM to [`includes.scss`](https://github.com/Esri/calcite-components/blob/master/src/assets/styles/includes.scss#L18-L20).

src/components/tree-item/resources.ts
```diff
@@ -1,14 +1,18 @@
...
+  actionsEnd: "actions-end",
   checkboxLabel: "checkbox-label",
   checkbox: "checkbox",
   chevron: "chevron",
   nodeContainer: "node-container",
   childrenContainer: "children-container",
   bulletPointIcon: "bullet-point",
-  checkmarkIcon: "checkmark"
+  checkmarkIcon: "checkmark",
+  tabTitle: "calcite-tab-title--icon",
```
jcfranco, line 10: `tabTitle`/`calcite-tab-title--icon` doesn't seem to be referenced anywhere.

src/components/tree-item/tree-item.tsx
```diff
@@ -266,11 +277,27 @@ export class TreeItem
...
     const hidden = !(this.parentExpanded || this.depth === 1);
+    const { hasEndActions } = this;
+    const slotNode = (
+      <slot
+        key="actionsEndSlot"
+        name={SLOTS.actionsEnd}
+        onSlotchange={this.actionsEndSlotChangeHandler}
+      />
+    );
+    const iconStartEl = (
```
jcfranco, line 293: Nitpick: For readability, can you add a newline before the const declaration?

src/components/tree/tree.stories.ts
```diff
@@ -64,11 +115,17 @@ export const simple = (): string => html`
...
-export const selectionModeNone = (): string => html`<calcite-tree
-  ${boolean("lines", false)}
-  selection-mode="${select("selection-mode", selectionModes, "none")}"
+export const selectionModeNone = (): string => html`
+  <calcite-tree ${boolean("lines", false)} selection-mode="${select("selection-mode", selectionModes, "none")}"
+    >${treeItems}</calcite-tree
+  >
+`;
+
+export const actionsEndDropdownsAndIconStart = (): string => html`<calcite-tree
```
jcfranco, line 124: This looks like a test-specific story. If so, can you add the `_TestOnly` suffix?

src/components/tree-item/tree-item.tsx
```diff
@@ -266,11 +277,27 @@ export class TreeItem
...
+      <calcite-icon
+        class={{ [CSS.tabTitle]: true, [CSS.iconStart]: true }}
+        flipRtl={this.iconFlipRtl === "start" || this.iconFlipRtl === "both"}
+        icon={this.iconStart}
+        scale={this.scale === "l" ? "m" : "s"}
```
jcfranco, line 298: Sidebar: I think we can extract some utils to have consistent icon scale applied within components.

jcfranco (approved):
🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴
🌴🌳🌳🌳🌳🌳🌴🌳🌳🌳🌴🌴🌳🌳🌳🌳🌴🌳🌳🌳🌳🌴🌳🌳🌳🌴🌴🌳🌳🌳🌴🌴🌳🌳🌳🌴🌳🌳🌳🌳🌴🌳🌳🌳🌴🌴🌳🌳🌳🌴🌳🌴
🌴🌴🌴🌳🌴🌴🌴🌳🌴🌴🌳🌴🌳🌴🌴🌴🌴🌳🌴🌴🌴🌴🌳🌴🌴🌳🌴🌳🌴🌴🌳🌴🌴🌳🌴🌴🌳🌴🌴🌴🌴🌴🌳🌴🌴🌳🌴🌴🌴🌴🌳🌴
🌴🌴🌴🌳🌴🌴🌴🌳🌳🌳🌴🌴🌳🌳🌳🌴🌴🌳🌳🌳🌴🌴🌳🌳🌳🌴🌴🌳🌳🌳🌴🌴🌴🌳🌴🌴🌳🌳🌳🌴🌴🌴🌳🌴🌴🌳🌴🌴🌴🌴🌳🌴
🌴🌴🌴🌳🌴🌴🌴🌳🌴🌳🌴🌴🌳🌴🌴🌴🌴🌳🌴🌴🌴🌴🌳🌴🌳🌴🌴🌳🌴🌳🌴🌴🌴🌳🌴🌴🌳🌴🌴🌴🌴🌴🌳🌴🌴🌳🌴🌴🌴🌴🌴🌴
🌴🌴🌴🌳🌴🌴🌴🌳🌴🌴🌳🌴🌳🌳🌳🌳🌴🌳🌳🌳🌳🌴🌳🌴🌴🌳🌴🌳🌴🌴🌳🌴🌳🌳🌳🌴🌳🌴🌴🌴🌴🌳🌳🌳🌴🌴🌳🌳🌳🌴🌳🌴
🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴🌴

src/components/tree-item/tree-item.tsx
Elijbet: I feel like extracting utils is becoming my favorite thing to do. I'll open an issue.

PR #6012, opened 2022-12-13
refactor(date-picker)!: remove deprecated start and end properties

eriklharper:
BREAKING CHANGE: Removed the `start` and `end` properties, set `value` as an array with the start as the first value and the end as the second value instead.


driskull (approved):
👍

benelan (commented):
The properties are used in the demos, can you remove them there as well please

benelan:
And the usage snippets

driskull:
Don't forget to sniff out all the places these things are used. demos, usage, stories, etc.

I will find them later if not now 😈

PR #6014, opened 2022-12-13
refactor(label)!: Set default display to be `flex`

driskull:
BREAKING CHANGE: The default display for label is now `flex` instead of `inline`.

- Use `--calcite-label-margin-bottom` CSS variable to disable space when in `layout` is `inline`.


alisonailea (approved):

src/components/label/label.e2e.ts
```diff
@@ -2,7 +2,7 @@ import { newE2EPage } from "@stencil/core/testing";
 import { renders, hidden } from "../../tests/commonTests";
 
 describe("calcite-label", () => {
-  it("renders", () => renders("calcite-label", { display: "inline" }));
+  it("renders", () => renders("calcite-label", { display: "flex" }));
```
alisonailea, line 5: why are we testing the existence of CSS props here? Shouldn't that be covered by the screenshot tests?

alisonailea (approved):

src/components/label/label.e2e.ts
driskull: We're testing the default display of components because that may not necessary change a screenshot and we don't want to break changes for users.

macandcheese:
Looks good, I commented in Chromatic, but this breaking change will cause users to have to use the `--calcite-label-margin-bottom` css variable, which is expected 👍 .

PR #6026, opened 2022-12-13
refactor(tip)!: Rename property

driskull:
BREAKING CHANGE: Renamed property.

- Renamed the property `nonDismissible`, use `closeDisabled` instead.


benelan (approved):
:+1:

Issue #5130, opened 2022-08-10
Inactive `input-message` adds white-space below a `label`


nwhittaker:
### Actual Behavior

Given a `<calcite-label>` that contains a `<calcite-input>` and an inactive `<calcite-input-message>`, its bottom white-space is greater than a `<calcite-label>` that only contains a `<calcite-input>`.

### Expected Behavior

Given a `<calcite-label>` that contains a `<calcite-input>` and an inactive `<calcite-input-message>`, its bottom white-space is the same as a `<calcite-label>` that only contains a `<calcite-input>`.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/QWmBoEe

### Reproduction Steps

1. Visit the code sample
2. Note how the boxes around each `<calcite-label>` are different heights

The height of the 2nd box should match the height of the 1st box.

### Reproduction Version

beta.86

### Relevant Info

This inconsistency adds difficulty to maintaining a vertical rhythm within a form that includes only some inputs with an assoc. input message.

A quick fix could be to give `<calcite-input-message>` a `display: none` style when its `active` attribute is absent.

benelan:
I'm still able to reproduce on `next`, none of the changes effected the repro sample:
https://codepen.io/benesri/pen/oNyrqLY

alisonailea:
given the implementation of the global hidden attribute, it seems like this doesn't require a code change. Rather the user should apply the "hidden" attribute when the input message should be hidden. A solution which enables an empty input message to be applied in this context without adding any additional space will require a refactor of both label and input message. Is that what we want to do?

driskull:
@eriklharper @macandcheese @benelan wasn't it the plan for this component to go away and be rolled into `input`?

If so, should we deprecate it?

macandcheese:
I know it has been mentioned before, although I can't find an issue for it. I think it might have to be rolled into `label` if possible, since it is applicable to a lot of form components, not just `input`. 

It might be good to re-evaluate this as part of the larger form / validation story. I'd imagine that will be a larger epic for another major release, so not sure if we are quite ready to deprecate it?

benelan:
I honestly thought we were going into the opposite direction by splitting up `input`. If we roll `input-message` into `input` do we just delete `input-text` and `input-number`?

macandcheese:
[...]

PR #6037, opened 2022-12-14
fix(input-message): auto-hide message when there is no content

alisonailea:
**Related Issue:** #5130

## Summary
Auto hides the input message when there is no content.

Before:
Screen Shot 2022-12-14 at 1 03 51 PM

After:
Screen Shot 2022-12-14 at 1 02 37 PM


src/components/input-message/input-message.tsx
```diff
@@ -35,6 +35,9 @@ export class InputMessage {
   /** Specifies the status of the input field, which determines message and icons. */
   @Prop({ reflect: true, mutable: true }) status: Status = "idle";
 
+  /** Specifies if the message should be visible to the user */
+  @Prop({ reflect: true, mutable: true }) hidden = false;
```
driskull, line 39: Hey @alisonailea, instead of adding a `hidden` property, can you create a container `<div>` inside the component  that surrounds all the internal elements and hide it based on the textContent logic? The reasoning is that there's already a global `hidden` attribute on HTMLElement's and we don't want to override any `hidden` attribute or `display:none` that a user sets on the host element. 

For example, if a user added `hidden` to this component it would be removed if it had textContent.

It's been a best practice that we don't change the display of an element during its lifecycle. I think I can add this to the conventions if its not already there.

Proposal:

```jsx
<Host calcite-hydrated-hidden={hidden}>
    <div class={{["has-content"]: hasContent}}>
        {this.renderIcon(this.requestedIcon)}
        <slot />
    </div>
</Host>
```

src/components/input-message/input-message.tsx
```diff
@@ -48,8 +51,13 @@ export class InputMessage {
   //--------------------------------------------------------------------------
 
   connectedCallback(): void {
-    this.scale = getElementProp(this.el, "scale", this.scale);
-    this.requestedIcon = setRequestedIcon(StatusIconDefaults, this.icon, this.status);
+    const content = this.el.textContent;
+    if (content === "" && !this.icon) {
```
driskull, line 55: I think we will need to setup a mutationObserver to watch for `textConttent` changes. Otherwise, if a user adds or removes `textContent` after the component `connectedCallback` has been called then this if statement won't change. 

We need conventions on this as well. eek :(

I'll make a note for these conventions to be added.

driskull (changes requested):
Nice work 👍

alisonailea:
this is not the right fix here. As mentioned by @driskull this shouldn't control the hidden attribute on host. but no other solution will work without a refactor of calcite-label as well. A workaround solution is simply for users to add "hidden" attribute to empty input-message.

PR #6042, opened 2022-12-14
refactor(alert, notice)!: Renamed properties and updated values

macandcheese:
BREAKING CHANGE: Renamed `color` properties and updated values.

- Renamed the property `color`, use `kind` instead.
- Updated the accepted values of `kind` to `brand`, `danger`, `info`, `success`, and `warning`.


src/components/interfaces.ts
```diff
@@ -16,3 +17,11 @@ export type Width = "auto" | "half" | "full";
 // used to help track of event payloads to remove at 1.0.0 – see https://github.com/Esri/calcite-components/issues/3781
 /* Note : should be removed before `1.0 */
 export type DeprecatedEventPayload = any;
+
+export enum KindIcons {
```
macandcheese, line 21: Is there a better place for this? Previously `notice` was referencing this from the `alert` `interfaces.ts`.

src/components/alert/alert.tsx
```diff
@@ -46,6 +46,7 @@ import {
   LoadableComponent,
   componentLoaded
 } from "../../utils/loadable";
+import { MenuPlacement } from "../../utils/floating-ui";
```
macandcheese, line 49: Note: I updated to use this shared resource instead of explicitly defining in `interfaces.ts`, since the accepted values are the same. That said, `alert` doesn't use `floating-ui`, so I can revert if needed.

macandcheese:
@driskull @benelan is there a better format for `Updated the accepted values...` part of the changelog entry? I couldn't find any previous examples that renamed values in this way.

src/components/interfaces.ts
driskull: I thought the enum types were causing issues on the doc site?

This is the correct place for types. constants would go in a resources file in the folder.

driskull (approved):
👍

src/components/interfaces.ts
macandcheese: These aren't publicly exposed, so not an issue for doc site. But they are used by more than one component - it seemed weird that previously Notice was referencing this from the Alert `resources` file - maybe a `resources` file in the component root is better than this location (although this would be the only thing)?

driskull: Yeah maybe resources is better.

macandcheese: 👍 added a shared `resources.ts` file to the `/component` folder root. Will merge after checks run.

macandcheese:
Merging this - it's approved in Chromatic, not sure why that isn't being reflected for the "UI Tests" check.

benelan:
> Merging this - it's approved in Chromatic, not sure why that isn't being reflected for the "UI Tests" check.

The Chromatic UI Review functionality isn't a required check, we haven't setup any workflows for using it yet. In order to pass the CI you need to accept/deny the individual snapshots by clicking on the UI Tests check's ["details" link](https://www.chromatic.com/build?appId=6266d45509d7eb004aa254fb&number=460).

Additionally, when the you push more changes to the branch you need to toggle the `pr ready for visual snapshots` label to get the snapshots to run again. This saves us a lot of snapshots, which allows us to add more coverage. Previously screener was running every time changes were pushed to the PR's branch.

I created a bash function to automatically toggle the label when you are ready to merge, or when you need to see the results of the snapshots. It requires the [GitHub CLI](https://cli.github.com/).

```sh
function ccvsnaps() {
    if [[ "$(gh repo view --json name -q ".name")" = "calcite-components" ]]; then
        local current_branch
        current_branch="$(git symbolic-ref --short HEAD)"
        gh pr edit "$current_branch" --remove-label "pr ready for visual snapshots"
        gh pr edit "$current_branch" --add-label "pr ready for visual snapshots"
    fi
}
```

We also have some [known false positives](https://github.com/Esri/calcite-components/issues/4825#issuecomment-1335921930) that you can approve if they come up. I'll fix the false positives and add all of this info to CONTRIBUTING when I have a sec.

PR #6045, opened 2022-12-15
refactor(rating)!: remove `event.detail` payload from events

driskull:

- Removed the `event.detail` property on the event `calciteRatingChange`, use `event.target` instead.


anveshmekala (approved):
LGTM!

PR #6046, opened 2022-12-15
refactor(tab-nav, tab-title)!: remove `event.detail` payload from events

driskull:

- TabNav
  - Removed the `event.detail` property on the event `calciteTabChange`, use `event.target` instead.
- TabTitle
  - Removed the `event.detail` property on the event `calciteTabsActivate`, use `event.target` instead.


src/components/tab-title/tab-title.tsx
```diff
@@ -275,7 +275,7 @@ export class TabTitle implements InteractiveComponent {
    *
    * @see [TabChangeEventDetail](https://github.com/Esri/calcite-components/blob/master/src/components/tab/interfaces.ts#L1)
```
anveshmekala, line 276: this detail needs to be removed

anveshmekala: applies to `tab-nav` occurence.

anveshmekala (approved):
👍

PR #6047, opened 2022-12-15
refactor(pagination)!: remove `event.detail` payload from events

driskull:

- Removed the `event.detail` property on the event `calcitePaginationChange`, use `event.target` instead.

### DIFF
--- a/src/components/pagination/pagination.tsx
+++ b/src/components/pagination/pagination.tsx
@@ -157,8 +157,6 @@ export class Pagination implements LocalizedComponent, LocalizedComponent, T9nCo
 
   /**
    * Emits when the selected page changes.
-   *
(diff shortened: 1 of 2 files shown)


src/components/pagination/pagination.tsx
```diff
@@ -160,7 +160,7 @@ export class Pagination implements LocalizedComponent, LocalizedComponent, T9nCo
    *
    * @see [PaginationDetail](https://github.com/Esri/calcite-components/blob/master/src/components/pagination/pagination.tsx#L23)
```
anveshmekala, line 161: nitpick: payload documentation link should be removed here.

anveshmekala (approved):
🚀

PR #6051, opened 2022-12-15
refactor(tree)!: remove `event.detail` payload from events

driskull:
BREAKING CHANGE: Removed `event.detail` payload from events and added `selectedItems` property.

- Added property `selectedItems`.
- Removed the `event.detail` property on the event `calciteTreeSelect`, use `event.target` instead.


anveshmekala (approved):
💯

Issue #6038, opened 2022-12-14
Refactor `getElementProp` across child components as an outdated pattern.


Elijbet:
### Description

Refactor `getElementProp` across child components as an outdated pattern.

Parents should be setting props on the child if it’s a tightly coupled component pair. With this current pattern if the parent prop changes the child doesn't know about the changes to the parent.

For all relevant components props are passed down and children no longer look up for the prop. To achieve this the parent gets a mutation observer to modify children where necessary.

### Proposed Advantages

- Bring consistency to child styling, where there is a tightly coupled component pair. 
- Children are in sync with parent props.
- Independent components retain autonomy, eg. a large-scale modal with small-scale child components.

### Which Component

All coupled components: where the child shares their parent’s name, like `list/list-item, flow/flow-item, combobox/combobox-item`.

- [x] accordion-item
    - [x] `iconType`
    - [x] `iconPosition`
    - [x] `scale`
    - [x] `selectionMode`
- [x] combobox-item
    - [x] `scale`
    - [x] `selectionMode`
- [x] combobox-item-group
   - [x] `scale`
- [x] dropdown-group
    - [x] `scale`
- [x] dropdown-item
    - [x] `scale`
    - [x] `selectionMode`
- [x] segmented-control-item
    - [x] `scale`
    - [x] `appearance`
    - [x] `layout`
- [x] stepper-item
    - [x] `icon`
    - [x] `numbered`
    - [x] `layout`
    - [x] `scale`
- [ ] tab-title
    - [x] `position`
    - [x] `scale`

eriklharper:
Prior art: https://github.com/Esri/calcite-components/blob/master/src/components/radio-button-group/radio-button-group.tsx#L106-L117

geospatialem:
Verified in the `main` branch. 🎉

Issue #5698, opened 2022-11-07
Improve scale of internally rendered icons across relevant components


Elijbet:
### Description

Follow-up to [#5245](https://github.com/Esri/calcite-components/issues/5245)

This issue will be rounding up all the components that need an `icon scale` upgrade, where the large-scale component gets a medium-scale icon (24px) instead of a small one, this includes `chevron` icons. Needs to maintain the height with or without an icon, accounting for all the prop variations that can affect height, eg. the default solid appearance with a border and transparent without.

🏗 ⚠️
WIP: Design in Figma is in progress for most of the listed components. Will update by linking Figma for each component as they become available. Please contact @SkyeSeitz if you need permission to access the provided link. 

### Proposed Advantages

This will provide a better visual distinction between scales. 

### Which Component

- [x] Accordion-Item ([Finished in Figma](https://www.figma.com/file/lfeQxz39iCKeKqkwvmFztV/24px-icons-%5Benhancement%5D-%5Bissue-%235698%5D?node-id=0%3A1))
- [x] Action (icon is contained in a button)
- [x] Button: https://github.com/Esri/calcite-components/pull/5659
- [x] FAB: https://github.com/Esri/calcite-components/pull/5659
- [x] Combobox-item
- [x] Dropdown-item
- [x] Input (all types with icons)
- [x] Input-number
- [x] Input-text
- [x] Input-date-picker
- [x] Input-time-picker
- [x] Inline Editable (icon is contained in a button)
- [x] Tab-title
- [x] Chip - https://github.com/Esri/calcite-components/pull/6058
- [x] Select
- [x] Pagination 

### Relevant Info
Follow-up issues:

- List, List-group, and List-item currently do not have any scale at any level and will be considered separately at a later time. 
- Tree-item icon satisfies the requirements but `action-end` needs to be reworked to not have an impact on height. 

_No response_

SkyeSeitz:
**Accordion Item** design updates are completed and available at this [Figma link](url). I will continue to add completed designs that same enhancements file for the list above as they are ready. cc @brittneyzwolfer for visibility 

<img width="894" alt="image" src="https://user-images.githubusercontent.com/29716057/200431039-ff3a44d6-8b4a-4434-9a63-f6fcbcf0d0fc.png">

<img width="1131" alt="image" src="https://user-images.githubusercontent.com/29716057/200431620-5df7499f-5a6d-473e-af7e-adaa23a06a16.png">

Elijbet:
[...]

PR #6055, opened 2022-12-15
refactor(accordion, accordion-item): `getElementProp` on child is factored out and parent gets a mutation observer to update children

Elijbet:
**Related Issue:** #6038, #5698

## Summary
`getElementProp` on child is factored out and parent gets a mutation observer to update children.


Elijbet:
I implemented this new pattern to just the `scale` property, this is necessary to move ahead with #5698.
I can merge this as is with just `scale` and continue working on `accordion-item` for #5698. I could make a follow-up issue to address all other properties that need to be inherited.
Or I can put all other properties that need to be inherited in this PR as well since most of the work is done. I would just need to know which additional ones need to be inherited (`appearance`, `iconPosition`, `iconType`).

src/components/accordion/accordion.tsx
```diff
@@ -200,4 +227,13 @@ export class Accordion {
 
   private sortItems = (items: any[]): any[] =>
     items.sort((a, b) => a.position - b.position).map((a) => a.item);
+
+  private passPropsToAccordionItems = (): void => {
+    const accordionItems = this.el.querySelectorAll("calcite-accordion-item");
```
driskull, line 232: nitpick:

```ts
Array.from(this.el.querySelectorAll("calcite-accordion-item")).forEach((accordionItem) => (accordionItem.scale = this.scale))
```

src/components/accordion-item/accordion-item.tsx
```diff
@@ -43,23 +43,26 @@ export class AccordionItem implements ConditionalSlotComponent {
...
   /** Specifies an icon to display at the start of the component. */
   @Prop({ reflect: true }) iconStart: string;
 
   /** Specifies an icon to display at the end of the component. */
   @Prop({ reflect: true }) iconEnd: string;
 
-  /** When `true`, the icon will be flipped when the element direction is right-to-left (`"rtl"`). */
-  @Prop({ reflect: true }) iconFlipRtl: FlipContext;
+  /** Specifies the size of the component inherited from the `accordion`. */
+  @Prop({ reflect: true }) scale: Scale = "m";
```
macandcheese, line 65: If this is not meant to be public, can we add an `* @internal` tag so it is excluded from documentation?

src/components/accordion/accordion.tsx
```diff
@@ -140,4 +167,12 @@ export class Accordion {
 
   private sortItems = (items: any[]): any[] =>
     items.sort((a, b) => a.position - b.position).map((a) => a.item);
+
+  private passPropsToAccordionItems = (): void => {
+    (
+      Array.from(
```
macandcheese, line 173: Can we use the existing `this.items` instead of querying for the child items again?

src/components/accordion/accordion.e2e.ts
```diff
@@ -47,23 +47,58 @@ describe("calcite-accordion", () => {
...
+      const item1: E2EElement = await page.find(`calcite-accordion-item[id='1']`);
+      expect(await item1.getProperty("scale")).toEqual(scale.s);
+
+      const icon1: E2EElement = await page.find(`calcite-accordion-item[id='1'] >>> .${CSS.iconStart}`);
+      expect(await icon1.getProperty("scale")).toEqual(scale.m);
+    });
+
+    it("renders expected scale icon on child when scale is set on child level (no parent override)", async () => {
+      const accordion: E2EElement = await page.find("calcite-accordion");
+      await accordion.getProperty("scale", scale.l);
```
macandcheese, line 93: should this be a `setProperty`?

src/components/accordion/accordion.e2e.ts
```diff
@@ -47,23 +47,58 @@ describe("calcite-accordion", () => {
...
-    </calcite-accordion-item>
-    </calcite-accordion>`);
-    const icon1 = await page.find(`calcite-accordion-item[id='1'] >>> .${CSS.iconStart}`);
-    const icon2 = await page.find(`calcite-accordion-item[id='2'] >>> .${CSS.iconStart}`);
-    const icon3 = await page.find(`calcite-accordion-item[id='3'] >>> .${CSS.iconStart}`);
-    expect(icon1).not.toBe(null);
-    expect(icon2).toBe(null);
-    expect(icon3).not.toBe(null);
+  describe("icon behavior", () => {
+    let page: E2EPage;
```
macandcheese, line 51: Does this need to be imported?

src/components/accordion/accordion.e2e.ts
```diff
@@ -47,23 +47,58 @@ describe("calcite-accordion", () => {
...
+    it("renders expected scale icon on child when scale is set on child level (no parent override)", async () => {
+      const accordion: E2EElement = await page.find("calcite-accordion");
+      await accordion.getProperty("scale", scale.l);
+
+      const item1: E2EElement = await page.find(`calcite-accordion-item[id='1']`);
+      await item1.setProperty("scale", scale.m);
+      await page.waitForChanges();
+
+      expect(await accordion.getProperty("scale")).toEqual(scale.l);
+      expect(await item1.getProperty("scale")).toEqual(scale.m);
```
macandcheese, line 100: Don't we want the scale of `calcite-accordion` and `calcite-accordion-item` to match in this case?

src/components/accordion/accordion.e2e.ts
```diff
@@ -47,23 +47,58 @@ describe("calcite-accordion", () => {
...
+    it("renders corresponding scale on accordion-item when parent scale changes, icon scale not affected", async () => {
+      const accordion: E2EElement = await page.find("calcite-accordion");
+      await accordion.setProperty("scale", scale.s);
+      await page.waitForChanges();
+
+      const item1: E2EElement = await page.find(`calcite-accordion-item[id='1']`);
+      expect(await item1.getProperty("scale")).toEqual(scale.s);
+
+      const icon1: E2EElement = await page.find(`calcite-accordion-item[id='1'] >>> .${CSS.iconStart}`);
+      expect(await icon1.getProperty("scale")).toEqual(scale.m);
```
macandcheese, line 88: Should this icon be a scale `s`?

src/components/accordion-item/accordion-item.tsx
```diff
@@ -133,6 +136,8 @@ export class AccordionItem implements ConditionalSlotComponent {
     ) : null;
   }
 
+  iconScaleAdjustment: Scale = this.scale === "l" ? "m" : "s";
```
macandcheese, line 139: I don't think this will always respond to changes, you'll likely need to watch the internal `scale` prop and update based on that:

```
  /**
   * Specifies the size of the component inherited from the `accordion`.
   * 
   * @internal
   */

  @Watch("scale")
  onScaleChange(): void {
    this.internalIconScale = this.scale === "l" ? "m" : "s";
  }

  private internalIconScale: Scale;
```

src/components/accordion/accordion.tsx
eriklharper: This makes sense as long as `this.items` gets updated when `accordion-item`s get added or removed from the light dom.  Is this currently handled in any way?

macandcheese: Yeah good point the internal registration events are kind of an outdated pattern and wouldn't respond - an observer that responds to addition / removal of children and updates `this.items` is probably better. There's a handful of our ... more senior 👴👵 ... components that could probably benefit from a change like that.

src/components/accordion/accordion.e2e.ts
eriklharper: > Don't we want the scale of `calcite-accordion` and `calcite-accordion-item` to match in this case - shouldn't let folks set scale "m" on the child if the parent is "l". (assuming scale on `calcite-accordion-item` is internal property)

I agree it seems odd to allow this, but if you think about it, this pattern breaks component autonomy.  When a component defines its own `scale` prop, I as a developer expect it to honor whatever value I'm explicitly setting it to, excepting the default fallback value.  If a parent value comes in and overrides that value I've explicitly set on the child, the property pattern on that child component no longer behaves according to standard convention.

src/components/accordion/accordion.e2e.ts
eriklharper: > Erik also brought up that we should not override an item with the parent if a user sets it individually to a different scale, although I'm not sure what the use case for that would be.

Not sure what the specific use case would be either, but I still think that if the `scale` property is explicitly applied either as an attribute or by setting the property directly on the child component, that this value should be honored and not overridden by the parent component.  We can actively discourage doing this in the documentation, and encourage users to set the value at the parent level to ensure visual consistency, but I don't think we should break from convention with regards to how property values are honored as the individual component level.

src/components/accordion/accordion.e2e.ts
macandcheese: Yeah, although if the property is internal and is set by / managed by the parent maybe that wouldn't be expected. Though to follow through on that the scale probably shouldn't have the default to "m" until it receives the scale from `calcite-accordion`.

eriklharper: This same principle applies to object inheritance in programming.  If I have Object A that defines a `foo` property, an Object B that extends Object A, I don't expect setting the `foo` property on Object A to override what I've explicitly set it to on Object B.  I expect the lowest level of specificity on that property to win above all others.

macandcheese: Right - but its not a public property (or, is proposed to not be an undocumented, internal property) - so the test case of it being set in that way should not be valid?

eriklharper: > Yeah, although if the property is internal and is set by / managed by the parent maybe I wouldn't expect to be able to set this at all, as its undocumented. Though to follow through on that the scale probably shouldn't have the default to "m" until it receives the scale from `calcite-accordion`.

For sure.  I think with a component like `accordion`, its children items are so tightly interwoven with the parent that this pattern does make sense where you would explicitly control children properties like `scale` from the parent level.  It gets trickier when you have child components that can live standalone by themselves like `radio-button` and `tile-select` where they need to work on their own regardless if their corresponding parent wrapper component is being used.

> Right - but its not a public property (or, is proposed to not be an undocumented, internal property) - so the test case of it being set in that way should not be valid?

Agreed.  I think if we mark these properties as internal, and call out in our documentation that modifying internal properties is not supported, then this pattern can be adopted for certain parent/child component systems.

Elijbet:
Closing this, as it was resolved in #7191.

PR #6056, opened 2022-12-15
refactor!: enabled delegatesFocus

benelan:
BREAKING CHANGE: Removed the `focusId` parameter from `action-bar` and `action-pad`'s `setFocus` method

- **action-bar, action-pad:** Removed `focusId` paramter `setFocus` method, focus is delegated to the first focusable element.

---
I'll remove this part before merging.

**Related Issue:** #5147

## Summary
A lot of the components require feedback from Franco. More info in this issue comment: https://github.com/Esri/calcite-components/issues/5147#issuecomment-1349616672

These are the uncontroversial components that we can install before Franco gets back.


src/components/filter/filter.tsx
```diff
@@ -184,7 +185,7 @@ export class Filter
   async setFocus(): Promise<void> {
     await componentLoaded(this);
 
-    focusElement(this.textInput);
+    this.el?.focus();
```
driskull, line 188: Wouldn't this no longer focus the textInput?

driskull (approved):
Nice. Just had a question. If we call `focus()` on a component with delegatesFocus enabled does it automatically focus the first element?

src/components/filter/filter.tsx
benelan: Yeah the Input is the first focusable element. I double checked to be sure, putting this in the demos for this branch correctly focuses the Input:

```html
  <calcite-filter></calcite-filter>
  <script>
    customElements.whenDefined("calcite-filter").then(() => 
        document.querySelector("calcite-filter").setFocus());
  </script>
```

> If we call `focus()` on a component with delegatesFocus enabled does it automatically focus the first element?



It depends on whether everything is loaded. For example this works, but not without the timeout:

```js
    customElements
      .whenDefined("calcite-filter")
      .then(() => setTimeout(() => document.querySelector("calcite-filter").focus(), 1000));
```

We still need the async `setFocus()` methods.

PR #6058, opened 2022-12-15
refactor(chip)!: Renamed property and updated values

macandcheese:
BREAKING CHANGE: Renamed `color` property and updated values, and updated `appearance` values.

- Renamed the property `color`, use `kind` instead.
- Updated the accepted values of `kind` to `brand`, `inverse`, and `neutral` (default).
- Updated the accepted values of `appearance` to , `outline`, `outline-fill` and `solid` (default).

### DIFF
--- a/src/components/chip/chip.scss
+++ b/src/components/chip/chip.scss
@@ -6,16 +6,16 @@
   --calcite-chip-spacing-unit-s: theme("spacing.1");
   .image-container {
     @apply h-5 w-5;
-    padding-inline-start: 1px;
+    padding-inline-start: theme("spacing[0.5]");
   }
 }
 :host([scale="m"]) {
   @apply text-n1 h-8;
   --calcite-chip-spacing-unit-l: theme("spacing.3");
-  --calcite-chip-spacing-unit-s: 6px;
+  --calcite-chip-spacing-unit-s: calc(theme("spacing.3") / 2);
   .image-container {
     @apply h-6 w-6;
-    padding-inline-start: theme("padding.1");
+    padding-inline-start: theme("spacing.1");
   }
 }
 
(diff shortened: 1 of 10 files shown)


macandcheese:
Along with the refactored properties and values, these changes incorporate some other small updates to avoid further breaking change in a few months when we add `chip-group`:

https://github.com/Esri/calcite-components/issues/4837,
https://github.com/Esri/calcite-components/issues/3838,
https://github.com/Esri/calcite-components/issues/2506,
https://github.com/Esri/calcite-components/issues/5698

These are being added to Figma - cc @ashetland @SkyeSeitz.

https://user-images.githubusercontent.com/4733155/207926416-b0728c1a-a5bd-4fda-8703-1a7ba547ce68.mov

src/components/chip/chip.scss
```diff
@@ -45,47 +46,95 @@
   @apply inline-flex h-full max-w-full items-center;
 }
 
-.title {
+:host .title {
```
driskull, line 49: I don't think `:host` is needed here.

src/components/chip/chip.scss
```diff
@@ -6,15 +6,16 @@
   --calcite-chip-spacing-unit-s: theme("spacing.1");
   .image-container {
     @apply h-5 w-5;
+    padding-inline-start: theme("spacing[0.5]");
```
driskull, line 9: new styles?

src/components/chip/chip.scss
```diff
@@ -45,47 +46,95 @@
...
 
-.title {
+:host .title {
   @apply truncate;
+  padding-block: 0;
 }
 
-:host span {
-  padding-block: 0;
+:host .content--slotted .title {
```
driskull, line 54: I don't think `:host` is needed here.

src/components/chip/chip.tsx
```diff
@@ -178,10 +184,34 @@ export class Chip
     this.closed = true;
   };
 
+  private updateHasContent() {
+    const slottedContent = this.el.textContent.trim().length > 0 || this.el.childNodes.length > 0;
+    this.hasContent =
+      (getSlotted(this.el, SLOTS.image) || this.el.childNodes.length === 1) &&
```
driskull, line 190: `getSlotted()` internal util is deprecated. If we put this in, can we create an issue to replace it with `onSlotChange`? the `onSlotChange` can set `hasContent` depending on how many assigned elements it has.

driskull (approved):
I'm ok with doing some bug fixes in the refactoring but I think this one needs to be tested pretty well since its doing more than just refactoring.

src/components/chip/chip.tsx
macandcheese: I'll update to use `onSlotChange` and open a separate issue to track / remove places where `getSlotted` is currently used.

macandcheese: https://github.com/Esri/calcite-components/issues/6059

src/components/chip/chip.scss
macandcheese: Yeah, there were some new rules needed, mostly to achieve the proper spacing for icon/image-slot-only fixes from https://github.com/Esri/calcite-components/issues/4837, the `spacing[0.5]` itself already exists.

src/components/chip/chip.scss
macandcheese: Correct - removed 🧹

src/components/chip/chip.scss
macandcheese: Correct - removed 🧹

macandcheese:
@ashetland @SkyeSeitz I've added you as reviewers here- as @driskull mentioned this has a few extra changes to take a look at - I know these are noted for inclusion for Figma but want to triple check :)

SkyeSeitz:
> @ashetland @SkyeSeitz I've added you as reviewers here- as @driskull mentioned this has a few extra changes to take a look at - I know these are noted for inclusion for Figma but want to triple check :)

Yes! These changes are slotted for Figma 🚀

SkyeSeitz (approved):

Issue #6007, opened 2022-12-13
Refactor event payload for consistency


geospatialem:
### Description

The following list of events could have their payloads deprecated`*`, or cleaned up for consistency. cc @driskull 

Note (`*`) : For those where the event payload could be removed, some may require introducing a public property.

- [x] `calciteComboboxFilterChange` - @driskull 
- [x] `calciteComboboxChange` - @driskull 
- [x]  `calciteDatePickerChange`  - @driskull 
- [x] `calciteDatePickerRangeChange` - @driskull 
- [x] `calciteDropdownSelect` - @driskull 
- [x] `calciteRatingChange` - @driskull 
- [x] `calciteTabChange` - @driskull 
- [x] `calciteTreeSelect` - @driskull 
- [x] `calcitePaginationChange` - @driskull 
- [x] `calciteDatePickerActiveDateChange` - @driskull
- [x] `calciteDatePickerSelect` - @driskull
- [x] `calciteRadioGroupChange` - @driskull 
- [x] `calciteTabsActivate` - @driskull 
- [ ] `calciteListChange`. (PickList/ValueList) - TBD
- [ ] `calciteListOrderChange` (PickList/ValueList)  - TBD

### Proposed Advantages

For consistency across components prior to the January release.

### Which Component

- combobox
- date-picker
- dropdown
- list
- pagination
- radio-group
- rating
- tab
- tabs
- tree

### Relevant Info

An audit was conducted to determine the events sought for event payload deprecation or clean up. Visit https://github.com/Esri/calcite-components/issues/2925#issuecomment-1349503060 for details.

driskull:
@eriklharper @benelan @geospatialem in regards to removing `event.detail` from some events...

These events are on the DatePickerMonth/DatePickerMonthHeader. Do we need to document these or can we just mark them internal? These components aren't designed to be used on their own.

- `calciteDatePickerSelect` - DatePickerMonth/DatePickerMonthHeader
- `calciteDatePickerActiveDateChange` - DatePickerMonthHeader

These are on the ValueList/PickList. Should we modify them or just let them be since we will likely deprecate them?

- `calciteListChange`
- `calciteListOrderChange`

My vote would be to deprecate the list events and mark the datepicker ones as internal. Thoughts?

anveshmekala:
I agree with marking datepicker ones as `@internal` and deprecating the valueList/pickList related events considering the components will be deprecated.

benelan:
Yup agree on both points

[...]

PR #6061, opened 2022-12-15
docs(pick-list, value-list): Deprecate events

driskull:
docs(pick-list, value-list): Deprecate events `calciteListOrderChange` and `calciteListChange`. #6007


benelan (approved):
Is this un-deprecating the events, or are they going to be removed later?

driskull:
I can leave them alone, but the issue with them is the `event.detail`. I didn't think it was worth fixing since these components may go away at some point next year.

PR #6063, opened 2022-12-15
refactor(color-picker-hex-input)!: removed `intl*` properties

anveshmekala:
BREAKING CHANGE: removed `intl` properties. 

- Removed, `intlHex` property, aria-label of color-picker-hex-input is set to `hex` by default.
- Removed ,`intlNoColor` property.

_note: color-picker-hex-input is `internal` component._


src/components/color-picker-hex-input/color-picker-hex-input.tsx
```diff
@@ -223,14 +208,14 @@ export class ColorPickerHexInput implements LoadableComponent {
...
-    const { intlHex, value } = this;
+    const { value } = this;
     const hexInputValue = this.formatForInternalInput(value);
 
     return (
       <div class={CSS.container}>
         <calcite-input
           class={CSS.input}
-          label={intlHex}
+          label={"hex"}
```
driskull, line 218: Does this ever need to be translated?

driskull (approved):
👍

src/components/color-picker-hex-input/color-picker-hex-input.tsx
anveshmekala: this needs to be translated. one way is to parse it from the `color-picker` component as `messages` similar to what we did for `date-picker-month-header`.

src/components/color-picker-hex-input/color-picker-hex-input.tsx
```diff
@@ -42,6 +41,13 @@ export class ColorPickerHexInput implements LoadableComponent {
 
   @Element() el: HTMLCalciteColorPickerHexInputElement;
 
+  /**
+   * Specifies accessible label for `calcite-input`.
+   *
+   * @readonly
```
driskull, line 47: I don't think this is readOnly since its set by a parent component.

src/components/color-picker-hex-input/color-picker-hex-input.tsx
```diff
@@ -42,6 +41,13 @@ export class ColorPickerHexInput implements LoadableComponent {
...
+   */
+  @Prop() hexLabel = "Hex";
```
driskull, line 49: Maybe we can just leave it as empty string?

driskull (approved):
👍

src/components/color-picker-hex-input/color-picker-hex-input.tsx
anveshmekala: leaving this empty would break `a11y` test in `color-picker-hex-input`  with `label` being empty.

driskull: Ok. this is fine then.

PR #6068, opened 2022-12-16
chore: Update interface references

macandcheese:
Removes individually specified `Appearance` interfaces and instead uses `Extract<>` to source values from a single defined reference.

### DIFF
--- a/src/components/color-picker/interfaces.ts
+++ b/src/components/color-picker/interfaces.ts
@@ -1,7 +1,5 @@
 import type Color from "color";
 
-export type ColorAppearance = "default" | "minimal" | "solid";
-
 export type ColorMode = "rgb" | "hsv";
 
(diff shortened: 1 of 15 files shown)


src/components/color-picker/interfaces.ts
```diff
@@ -1,7 +1,5 @@
 import type Color from "color";
 
-export type ColorAppearance = "default" | "minimal" | "solid";
```
macandcheese, line 3: I couldn't find any references to this "default" value.

anveshmekala: `default` is deprecated [here](https://github.com/Esri/calcite-components/pull/4744). I think its safe to remove it.

driskull (approved):
💯 Awesome. Super nice that these will all be using the same interface for consistency.

Issue #5616, opened 2022-10-25
Tree Item - Use aria-hidden for SVGs - (2036570418)


dqateam:
### Module: 
 38 Tree and Tree Item 

### Violation: 
Hide SVGs from assistive technologies using the `aria-hidden` prop.

![image](https://user-images.githubusercontent.com/5023024/202574349-d447e3b7-b45f-4ea8-b16b-fe67ef56cef0.png)

### WCAG Reference: 
- [1.1.1 Non-text Content](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html) 

### Instance ID: 
 2036570418 

### Severity: 
 10 

### Description: 
 [Issue]
There are SVG images that serve a decorative purpose but are exposed to screen readers.

[User Impact]
Screen readers may announce <img> elements with no alt attribute by their filename and SVGs with no accessible name as "group." This gives screen reader users irrelevant and confusing information.

[Code Reference] 
 ```
<svg class="check-svg" viewbox="0 0 16 16"><path d=""></path></svg>
 ``` 

### Note: 
 [Suggestion]
Provide alternative text for images. Non-meaningful images must be marked as decorative, which causes screen readers to ignore them. To mark an ``` <svg> ``` element as decorative, add aria-hidden="true" to the element.

[Compliant Code Example] 
 ```
<svg aria-hidden="true" class="check-svg" viewbox="0 0 16 16"><path d=""></path></svg>
 ```
 

### Media Type: 
 Images 

### Additional Resources: 
- Report Source: Product accessibility evaluation conducted on the Calcite Design System in August, 2022.

geospatialem:
Verified in `next.698`.

PR #6069, opened 2022-12-16
fix(icon, graphic, loader): Set aria-hidden on internal svg elements

driskull:
**Related Issue:** #5616

## Summary

fix(icon, graphic, loader): Set aria-hidden on internal svg elements

### DIFF
--- a/src/components/graph/graph.tsx
+++ b/src/components/graph/graph.tsx
@@ -71,6 +71,7 @@ export class Graph {
     if (!data || data.length === 0) {
       return (
         <svg
+          aria-hidden="true"
           class="svg"
           height={height}
           preserveAspectRatio="none"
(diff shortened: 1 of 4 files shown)


src/components/graph/graph.tsx
```diff
@@ -100,6 +101,7 @@ export class Graph {
     const fill = colorStops ? `url(#linear-gradient-${id})` : undefined;
     return (
       <svg
+        aria-hidden="true"
```
alisonailea, line 104: given how all these SVG's share many attributes and classes this seems like a good opportunity for a utility function or functional component to normalize calcite svgs.

alisonailea (approved):

geospatialem (approved):
Great a11y convention additions included in this fix, @driskull!    

🚀🌟

Issue #5109, opened 2022-08-05
combobox throws error when typing a custom value and then trying to open the menu


jcfranco:
### Actual Behavior

It seems the combobox expects custom values to be committed when opening the item menu, doing so throws the following:

```
p-40825758.entry.js:6 Uncaught TypeError: Cannot read properties of undefined (reading 'offsetHeight')
    at U.calculateSingleItemHeight (p-40825758.entry.js:6:11612)
    at U.scrollToActiveItem (p-40825758.entry.js:6:8224)
    at U.shiftActiveItemIndex (p-40825758.entry.js:6:14819)
    at U.keydownHandler (p-40825758.entry.js:6:3629)
```

### Expected Behavior

At a minimum, no error should be thrown. What may need discussion is whether to allow the regular menu open experience and preserve the in-progress text if a selection is made.

### Reproduction Sample

https://codepen.io/jcfranco/pen/abYGwWj?editors=1010

### Reproduction Steps

1. click on combobox input
2. type extraneous value ("asdf"), w/o committing
3. press down key

### Reproduction Version

beta.90

geospatialem:
Verified in `beta.99`.

PR #6071, opened 2022-12-16
fix(combobox): Fix error when typing a custom value

driskull:
**Related Issue:** #5109

## Summary

fix(combobox): Fix error when typing a custom value and then trying to open the menu. #5109


macandcheese (approved):
🧹

PR #6073, opened 2022-12-16
chore: Clean outdated properties

macandcheese:
Cleans up some stragglers from the recent `kind` property work. Also removes the `Color` section from the `Styling` conventions.

There are still some instances of `color` in generated Readme files but they should be fixed next time those are auto-generated.


driskull (approved):
Nice catches

PR #6076, opened 2022-12-16
refactor(input-date-picker)!: Removed deprecated start, startAsDate, end, endAsDate properties

eriklharper:
BREAKING CHANGE: Removed the `start`, `end`, `startAsDate`, and `endAsDate` properties.

- Removed the property `start`, use `value` instead.
- Removed the property `end`, use `value` instead.
- Removed the property `startAsDate`, use `valueAsDate` instead.
- Removed the property `endAsDate`, use `valueAsDate` instead.


driskull (approved):
👍 Code looks good!

PR #6077, opened 2022-12-16
refactor(modal)!: Remove `backgroundColor` property and added a CSS variable

driskull:
BREAKING CHANGE: Removed `backgroundColor` property.

- Removed the property `backgroundColor`, use the CSS variable `--calcite-modal-content-background` instead.


src/components/modal/modal.scss
```diff
@@ -5,6 +5,7 @@
  *
  * @prop --calcite-modal-padding:Specifies the padding of the modal.
  * @prop --calcite-scrim-background: The component's semi-transparent background color.
+ * @prop --calcite-modal-content-background: Specifies the background color of the component's content.
```
macandcheese, line 8: nitpick - "of content placed in the `content` slot" or something along those lines?

macandcheese (approved):
Looks good! Small doc comment.

PR #6080, opened 2022-12-16
refactor(input-time-picker)!: removed deprecated `locale` property

anveshmekala:
BREAKING CHANGE:  Removed deprecated `locale` property.

- Removed the property `locale`, use `lang` instead.


driskull (approved):
sweeet

PR #6082, opened 2022-12-16
refactor(shell-panel)!: removed deprecated `intlResize` property

anveshmekala:
BREAKING CHANGE: Removed deprecated `intlResize` property.

- Removed the property `intlResize`, use `messagesOverrides.resize` instead.


driskull (approved):
👍

PR #6084, opened 2022-12-16
refactor(date-picker)!: removed deprecated properties

benelan:
BREAKING CHANGE: Removed `endAsDate` and `startAsDate` properties, use `valueAsDate` instead

- Removed `endAsDate` and `startAsDate` properties, use `valueAsDate` instead


alisonailea (approved):

driskull (approved):
👍

PR #6085, opened 2022-12-16
refactor(date-picker): Remove deprecated startAsDate and endAsDate properties

eriklharper:
## Summary

- Removed the property `startAsDate`, use `valueAsDate` instead.
- Removed the property `endAsDate`, use `valueAsDate` instead.


driskull (approved):
Looks good to me 👍

Issue #6081, opened 2022-12-16
[Accordion] Remove extraneous code with home/end/arrow keys


macandcheese:
### Actual Behavior

I am working on making some of our focus helpers into a general utility : https://github.com/Esri/calcite-components/pull/6075#discussion_r1051100417

While doing so, I discovered our accordion keyboard navigation (besides `tab` / `shift + tab`) stopped working at some point: https://codepen.io/mac_and_cheese/pen/LYBPyPr?editors=1000

### Expected Behavior

After discussion, this is the intentional behavior, the stale code was never removed - this issue has been repurposed to remove that code.

macandcheese:
After discussing with @geospatialem - this may be intentional / expected. But we still have a bunch of code in the components that supposedly is handling this, so we can remove if this is indeed desired behavior.

geospatialem:
We're going to proceed with leaving the `Home` / `End` / arrow keys as-is, instead repurposing this issue to remove the extraneous code to support the pattern.

geospatialem:
Verified on `master` and in the `accordion` component in `beta.99`.

The `dropdown`, `stepper`, and `tabs` components added enhancement utility is also working great in `beta.99`.

PR #6086, opened 2022-12-16
refactor(accordion, dropdown, stepper, tabs): Updates keyboard handling

macandcheese:
**Related Issue:** #6081 [#6803](https://github.com/Esri/calcite-components/issues/6083)

## Summary
As suggested by @alisonailea [in this comment](https://github.com/Esri/calcite-components/pull/6075#discussion_r1051100417), extracts some common keyboard navigation focus helpers into a shared utility.

Along the way discovered a few things that are fixed alongside the refactors:

Accordion:
- Cleanup: The keyboard expected navigation changed at some point, which I confirmed with @geospatialem, but there was a lot of related stale code that is removed.

Dropdown:
- Fix: Previously, tab / shift tab would skip elements resulting in navigating to every other item.
- Fix: Previously, there was inconsistent close behavior when using tab / shift tab at the first / last item.
- Fix: Previously, when `accordion-item` had a populated `href` property, there was a "double focus" issue.
- Cleanup: Uses the new utility to handle keyboard navigation / focus

Stepper / Tabs:
- Cleanup: Uses the new utility to handle keyboard navigation / focus

All existing tests are passing, but we may want some extra passes of manual testing, especially in the dropdown case, since there were so many weird bugs previously.


src/utils/dom.ts
```diff
@@ -343,3 +343,37 @@ export function slotChangeGetAssignedElements(event: Event): Element[] {
...
+
+/**
+Focus helpers for keyboard navigation
+ *
+ * @param elements : An array of elements
+ * @param elCurrent : The current element
+ * @param destination : "next" | "prev" | "first" | "last"
+ */
+
+export const focusElementInGroup = (elements: Element[], elCurrent: Element, destination: string): void => {
```
driskull, line 355: nice!

src/utils/dom.ts
```diff
@@ -343,3 +343,37 @@ export function slotChangeGetAssignedElements(event: Event): Element[] {
...
+  destination =
+    destination === "prev" && isFirstItem ? "last" : destination === "next" && isLastItem ? "first" : destination;
+
+  let focusTarget;
+
+  switch (destination) {
+    case "next":
+      focusTarget = elements[currentIndex + 1] || elements[0];
+      break;
+    case "prev":
```
driskull, line 368: I'd prefer just being specific with `"previous"` but just a nitpick :)

src/utils/dom.ts
driskull, line 355: An enhancement would be to return the focusTarget instead of just void. That would let any component know if something were focused.

driskull (approved):
👍 Nice cleanup. Removed a bit of code!

src/utils/dom.ts
macandcheese: Like this?

```
export const focusElementInGroup = (elements: Element[], elCurrent: Element, destination: string): Element => {
 
  ...

  focusElement(focusTarget);
  return focusTarget;
};
```

driskull: Yep!

macandcheese: Cool, updated.

src/utils/dom.ts
macandcheese: Updated

PR #6088, opened 2022-12-17
refactor(time-picker)!: remove target parameter from setFocus

benelan:
BREAKING CHANGE: remove `target` parameter from `setFocus()`

- Removed  `target` parameter from `setFocus()`, focus will be delegated to the first focusable element instead.


anveshmekala (approved):
🚀

driskull (approved):
👍

PR #6090, opened 2022-12-17
chore(block): Expose css var

macandcheese:
- Documents the `--calcite-block-padding` css variable


driskull (approved):
👍 

Should we have a story to test this is working?

macandcheese:
Yep, added some tests @driskull

PR #6091, opened 2022-12-17
chore(scrim): Expose css var

macandcheese:
- Documents the `--calcite-scrim-background` css variable

Note: This also changes the defaults, defined in `global.scss`, to have an `-internal` suffix. I'm not sure the history of why they are in that file and not locally defined in `scrim.scss` - maybe we prefer to move them to `scrim.scss`?


driskull (approved):
👍 

Do we need a story to verify this is working?

macandcheese:
This one already had a test, the variable just wasn't documented:
 https://github.com/Esri/calcite-components/blob/234e21aa3bfe58be0e1b8951aaeea2509019a690/src/components/scrim/scrim.e2e.ts#L145

PR #6093, opened 2022-12-17
chore(modal): Expose css var 

macandcheese:
Documents the `--calcite-modal-scrim-background` css variable.

### DIFF
--- a/src/components/modal/modal.e2e.ts
+++ b/src/components/modal/modal.e2e.ts
@@ -507,16 +507,16 @@ describe("calcite-modal accessibility checks", () => {
     });
     const scrimStyles = await page.evaluate(() => {
       const scrim = document.querySelector("calcite-modal").shadowRoot.querySelector(".scrim");
-      return window.getComputedStyle(scrim).getPropertyValue("--calcite-scrim-background").trim();
+      return window.getComputedStyle(scrim).getPropertyValue("--calcite-scrim-background");
     });
-    expect(scrimStyles).toEqual("rgba(0, 0, 0, 0.85)");
+    expect(scrimStyles).toEqual(" rgba(0, 0, 0, 0.85)");
   });
 
   it("when modal css override set, scrim should adhere to requested color", async () => {
     const overrideStyle = "rgba(160, 20, 10, 0.5)";
     const page = await newE2EPage({
       html: `
-      <calcite-modal aria-labelledby="modal-title" open style="--calcite-modal-scrim-background: ${overrideStyle}">
+      <calcite-modal aria-labelledby="modal-title" open style="--calcite-modal-scrim-background:${overrideStyle}">
         <h3 slot="header" id="modal-title">Title of the modal</h3>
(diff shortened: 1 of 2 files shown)


src/components/modal/modal.e2e.ts
```diff
@@ -509,7 +509,29 @@ describe("calcite-modal accessibility checks", () => {
       const scrim = document.querySelector("calcite-modal").shadowRoot.querySelector(".scrim");
       return window.getComputedStyle(scrim).getPropertyValue("--calcite-scrim-background");
     });
-    expect(scrimStyles).toEqual("rgba(0, 0, 0, 0.75)");
+    expect(scrimStyles).toEqual(" rgba(0, 0, 0, 0.85)");
```
macandcheese, line 512: For some reason there is an extra space being returned with the new fallback css var setup, but the value is correct.

Screen Shot 2022-12-16 at 8 24 33 PM

src/components/modal/modal.scss
```diff
@@ -3,16 +3,18 @@
...
+ * @prop --calcite-modal-scrim-background: Specifies the background color of the modal scrim.
  */
 
 :host {
   @apply absolute flex inset-0 opacity-0 z-overlay;
   visibility: hidden !important;
   transition: visibility 0ms linear var(--calcite-internal-animation-timing-slow),
     opacity var(--calcite-internal-animation-timing-slow) $easing-function;
+  // the modal should always use a dark scrim, regardless of light / dark mode - matches value in global.scss
+  --calcite-modal-scrim-background-internal: #{rgba($blk-240, 0.85)};
```
macandcheese, line 17: We were already hardcoding this default (it's dark in both light and dark mode), but variable-ized it ✨

macandcheese, line 17: @SkyeSeitz @ashetland - I'm not sure why this was a different value than the regular scrim, I couldn't find an example of this in the Figma kit, but I confirmed it matches scrim. Let me know if this needs to be adjusted.

src/components/modal/modal.e2e.ts
benelan: Are we able to fix the fallback css var setup instead of doing this?

macandcheese: I tried but I think it gets added during css processing, unless I’m missing something obvious. We could ‘trim()’ the expected value in the test, although it’s not ideal either way.

driskull: We could maybe just do `expect(scrimStyles.trim()).toEqual`

src/components/modal/modal.scss
driskull: Nice!

driskull (approved):
👍 🎉

src/components/modal/modal.e2e.ts
macandcheese: Yeah just going with `trim()`, the value is correctly set - I think since the first value in the fallback sequence is not provided it's processed as an empty space.

PR #6096, opened 2022-12-17
refactor(radio-group)!: Update `appearance` value

macandcheese:
BREAKING CHANGE:

- Removed `minimal` appearance value, use `outline` instead.

### DIFF
--- a/src/components/radio-group-item/radio-group-item.scss
+++ b/src/components/radio-group-item/radio-group-item.scss
@@ -61,7 +61,7 @@
 }
 
 :host([checked]) .label--outline {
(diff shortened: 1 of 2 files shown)


macandcheese:
This one never had an issue opened but was using the outdated value. I also updated the display to use transparent to truly reflect the m̶e̶a̶n̶i̶n̶g̶ ̶o̶f̶ ̶l̶i̶f̶e̶ meaning of the value. 

cc @ashetland @SkyeSeitz in case this needs to be added to Figma, but it's always presented on the same color background so shouldn't affect any design.

driskull (approved):
🤘

PR #6099, opened 2022-12-17
refactor(input)!: remove deprecated event payload

benelan:
BREAKING CHANGE: remove deprecated event payload

- Removed the `nativeEvent` payload property which was being used internally.

### DIFF
--- a/src/components/color-picker/color-picker.tsx
+++ b/src/components/color-picker/color-picker.tsx
@@ -366,7 +366,7 @@ export class ColorPicker
 
   private handleChannelInput = (event: CustomEvent): void => {
     const input = event.currentTarget as HTMLCalciteInputElement;
-    const internalInput = event.detail.nativeEvent.target as HTMLInputElement;
+    const internalInput = input?.shadowRoot.querySelector("input") as HTMLInputElement;
     const channelIndex = Number(input.getAttribute("data-channel-index"));
 
     const limit =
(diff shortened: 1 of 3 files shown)


src/components/color-picker/color-picker.tsx
```diff
@@ -366,7 +366,7 @@ export class ColorPicker
 
   private handleChannelInput = (event: CustomEvent): void => {
     const input = event.currentTarget as HTMLCalciteInputElement;
-    const internalInput = event.detail.nativeEvent.target as HTMLInputElement;
+    const internalInput = input?.shadowRoot.querySelector("input") as HTMLInputElement;
```
driskull, line 369: I think this is ok for now but lets create an issue and add a comment here with the issue to refactor this.

driskull (approved):
👍 🥂

src/components/color-picker/color-picker.tsx
driskull: Maybe the input needs an undocumented method to set the internal input value? That might be a safer way to go

driskull:
@benelan don't forget to remove the extra stuff from the body of the commit when merging.

> If we want to get rid of the deprecated nativeEvent event payload, going into the shadowRoot in calcite-color-picker should be the equivelent. It is still getting the calcite-input from the event target. This would be the last deprecated item that we wanted to remove before v1 🚀

src/components/color-picker/color-picker.tsx
benelan: Good idea. I ended up going with an internal method that just syncs the internal input's `value` with `this.value`. Seems like the most stable option for now. Can I get a re-review if you're still on @driskull

driskull (approved):
Looks good 😊

PR #6104, opened 2022-12-18
refactor(translations): prepend component name to Messages type

benelan:
**Related Issue:** #

## Summary

Prepends the component's name to the name of the `Messages` type


driskull (approved):
Looks good. Seems like this shouldn't be necessary but it's also helpful to just have them named uniquely

benelan:
Yeah I thought it might be a stencil compiler issue, but I double checked and you're right. I think I forgot to do the double build thing which caused type errors for me locally. 

Apparently the build cancellation that we do to batch `next` releases happened in the millisecond between releasing to npm and pushing to github lol.

We are at `next.700` on NPM but only `next.699 `on GitHub, so it's throwing errors trying to publish 700 again. I'm fixing it. Should I still install this PR? It is a bit nicer anyway.

driskull:
Yeah go for it

PR #6107, opened 2022-12-18
chore(types): Adds missing value to SelectionMode

macandcheese:
Adds a missing value to SelectionMode type - no errors are thrown when you try to extract a value that doesn't exist in the type. Not sure if there's a way to validate the requested Extract values?

For example:   
`@Prop({ reflect: true }) selectionMode: Extract<"single" | "single-persist" | "multiple", SelectionMode> = "multiple";`

doesn't throw an error since the default `multiple` exists in the defined type, however:

`@Prop({ reflect: true }) selectionMode: Extract<"single" | "single-persist" | "multiple", SelectionMode> = "single-persist";`

would have thrown an error.

### DIFF
--- a/src/components/interfaces.ts
+++ b/src/components/interfaces.ts
@@ -7,7 +7,14 @@ export type Kind = "brand" | "danger" | "info" | "inverse" | "neutral" | "warnin
 export type Layout = "horizontal" | "vertical" | "grid";
 export type LogicalFlowPosition = "inline-start" | "inline-end" | "block-start" | "block-end";
 export type Position = "start" | "end";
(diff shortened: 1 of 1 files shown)


benelan (approved):
I'm pretty sure you're example is how TypeScript should behave. When using Extract on two Unions, the resulting Union type ends up being all of the values that are on both sides of the comma. 

So in the first example you're setting the default to a value in the type, and in the second example you're setting the default to a value that's not in the type, so its throwing errors

benelan:
This should be a `fix` instead of `chore` right? Because I think trying to use `single-persist` currently wouldn't work

benelan:
Nevermind it does still work https://codepen.io/benesri/pen/JjBjYbq

PR #6109, opened 2022-12-19
docs(changelog): modify indent levels and styling

geospatialem:
**Related Issue:** N/A

## Summary
Adds readability improvements to the changelog for the upcoming release.

PR #6110, opened 2022-12-19
docs(changelog): resort and modify indent levels and styling

geospatialem:
**Related Issue:** N/A

## Summary
Either this PR or #6109 will be incorporated in. The difference is the sorting/organization of the components (_see first bullet point below for organization details_).

**This is the recommended PR for merging (as long as developers agree) so those interpreting our change log can discern changes by component instead of having to sift through the entire 700+ line breaking changes between `beta.98` and `beta.99`.** 🚀 

- Adds major refactoring to the changelog so components show breaking changes more effectively.
  - Components now have all breaking changes grouped under their name once
  - Exceptions are grouped breaking changes where multiple components were listed (such as `list`, `list-item`, `list-group`) 
- Adds readability improvements to the changelog for the upcoming release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -169,8 +169,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
     `success`, and `warning`.
 
 - **tab-nav**: Removed `event.detail` payload from events.
+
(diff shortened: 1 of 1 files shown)


driskull (approved):
👍

Issue #6113, opened 2022-12-19
Tooltip does not always display on calcite-action button


zaramatheson:
### Actual Behavior

When hovering over the button, the tooltip does not consistently display. It seems ok the first time you hover, and then it seems to depend on placement. In some cases it will not display at all. 

![hover](https://user-images.githubusercontent.com/50747646/208496106-b21a8548-5579-4778-9dc9-5920293abf9d.gif)

cc @kevindoshier @kitt

### Expected Behavior

Tooltip should consistently display when hovering over the button. 

### Reproduction Sample

https://codepen.io/kevindoshier/pen/WNyBLVO

### Reproduction Steps

1. Open the codepen above
2. Hover mouse over the back button (Try a few times)
![image](https://user-images.githubusercontent.com/50747646/208495674-c8e59852-2f33-4db9-be70-841a049b99ab.png)
3. Sometimes tooltip will display, sometimes it will not

### Reproduction Version

next

### Esri team

ArcGIS Map Viewer

geospatialem:
@kevindoshier Would it be possible to get a more refined/directed Codepen for us to reproduce and diagnose further? cc @zaramatheson 

There might be a workaround if a heading is added the `flow-item` component.

kevindoshier:
@geospatialem - im not sure what you are asking for, that codepen is basically how the app uses the flow-item's, would just the basic flow-item sample work that you can access through the calcite-component flow-item docs?

geospatialem:
> @geospatialem - im not sure what you are asking for, that codepen is basically how the app uses the flow-item's, would just the basic flow-item sample work that you can access through the calcite-component flow-item docs?

@kevindoshier - we're going to continue to troubleshoot and debug on our end to help determine a fix. Stay tuned for updates.

geospatialem:
Verified on `next.705`.

PR #6119, opened 2022-12-19
fix(tooltip): Fix hover logic for elements within shadowRoot

driskull:
**Related Issue:** #6113

## Summary

fix(tooltip): Fix hover logic for elements within shadowRoot. #6113

It looks like the events `pointerover` and `pointerout` won't be triggered on elements within a shadowRoot. 

The alternative this PR uses it to switch to using just `pointermove` so that It can query for a tooltip anytime the pointer moves.  Because this can cause a lot of events to occur, this PR also moves `setTimeout` logic on `pointermove` so there is no performance decrease.


driskull:
@benelan @eriklharper can you review plz?

benelan (approved):
Changes LGTM. At some point in a separate PR, could you refactor the timeout functionality to use lodash's throttle util instead? It's already bundled in our dist since color-picker uses it

PR #6137, opened 2022-12-20
refactor(radio-button-group)!: remove `event.detail` payload from event and add property

driskull:
BREAKING CHANGE: Removed `event.detail` payload from event and added `selectedItem` property.

- Added property `selectedItem`.
- Removed the `event.detail` property on the event `calciteRadioButtonGroupChange`, use `event.target` and the property `selectedItem` instead.

### DIFF
--- a/src/components/radio-button-group/radio-button-group.e2e.ts
+++ b/src/components/radio-button-group/radio-button-group.e2e.ts
@@ -419,7 +419,7 @@ describe("calcite-radio-button-group", () => {
     `);
 
     const getSelectedItemValue = async (): Promise<string> => {
-      return await page.evaluate((): Promise<string> => {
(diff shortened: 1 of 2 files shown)


src/components/radio-button-group/radio-button-group.e2e.ts
```diff
@@ -418,6 +418,12 @@ describe("calcite-radio-button-group", () => {
       </calcite-radio-button-group>
     `);
 
+    const getSelectedItemValue = async (): Promise<string> => {
```
macandcheese, line 421: Nice! This was added in the Stepper PR as well, maybe a good candidate for helper util?

macandcheese (approved):
Nice ✨

PR #6138, opened 2022-12-20
refactor(stepper)!: remove event.detail payload from event and add property

driskull:
BREAKING CHANGE: Removed `event.detail` payload from event and added `selectedItem` property.

- Added property `selectedItem`.
- Removed the `event.detail` property on the event `calciteStepperItemChange`, use `event.target` and the property `selectedItem` instead.


macandcheese (approved):
Nice 🧹 !

Issue #6140, opened 2022-12-21
Support updating focus trap with new elements


driskull:
### Description

Support updating focus trap with new elements

### Acceptance Criteria

Elements can be added to a focus trap after the focus trap has been created.

### Relevant Info

https://github.com/Esri/calcite-components/issues/5882#issuecomment-1360051698

### Which Component

At this time, popover and modal.

### Example Use Case

```html
<body>

  <calcite-modal open>
    <p slot="content"></p>
    <button slot="primary" onclick="addInput();">Add input</button>
  </calcite-modal>
  <script>const content = document.querySelector('p');

    async function addInput() {
      const modal = document.querySelector("calcite-modal");
      const input = document.createElement('calcite-input');
      content.appendChild(input);
      await input.componentOnReady(); // Needed so that the input has been set up with all its internal elements
      await modal.updateFocusTrapElements();
      requestAnimationFrame(() => { // needed so that the input has been rendered
        input.setFocus();
      });
    }
  </script>
</body>
```

### Esri team

N/A

driskull:
@nwhittaker installed this feature.

geospatialem:
Verified in `next.728` using the sample above (also mentioned below for context):

```html
<body>

  <calcite-modal open>
    <p slot="content"></p>
    <button slot="primary" onclick="addInput();">Add input</button>
  </calcite-modal>
  <script>const content = document.querySelector('p');

    async function addInput() {
      const modal = document.querySelector("calcite-modal");
      const input = document.createElement('calcite-input');
      content.appendChild(input);
      await input.componentOnReady(); // Needed so that the input has been set up with all its internal elements
      await modal.updateFocusTrapElements();
      requestAnimationFrame(() => { // needed so that the input has been rendered
        input.setFocus();
      });
    }
  </script>
</body>
```

PR #6141, opened 2022-12-21
feat(popover, modal): Add the ability to update focus trap elements after initialization.

driskull:
**Related Issue:** #6140

## Summary

feat(popover, modal): Add the ability to update focus trap elements after initialization.

- Adds `updateFocusTrapElements` method to each `FocusTrap` component.
  - This method allows a user to update the focus trap elements after the focus trap has been initialized

## Example use case

```js
const modal = document.querySelector("calcite-modal");
const input = document.createElement("calcite-input");
content.appendChild(input);
await input.componentOnReady(); // To make sure the input has been rendered
await modal.updateFocusTrapElements(); // The new method to update focusable elements
requestAnimationFrame(() => input.setFocus()); // Wait for next frame to set focus
 ```

### DIFF
--- a/src/utils/focusTrapComponent.ts
+++ b/src/utils/focusTrapComponent.ts
@@ -105,6 +105,13 @@ export function focusFirstTabbable(component: FocusTrapComponent): void {
  * Helper to update the element(s) that are used within the FocusTrap component.
  *
  * @param {FocusTrapComponent} component The FocusTrap component.
+ * @example
+ * const modal = document.querySelector("calcite-modal");
+ * const input = document.createElement("calcite-input");
+ * content.appendChild(input);
+ * await input.componentOnReady();
+ * await modal.updateFocusTrapElements();
+ * requestAnimationFrame(() => input.setFocus());
  */
 export function updateFocusTrapElements(component: FocusTrapComponent): void {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by driskull** [comment]
@benelan @eriklharper can you review this one?

File: src/utils/focusTrapComponent.ts [hunk 1]
```diff
@@ -95,3 +100,19 @@ export function deactivateFocusTrap(component: FocusTrapComponent): void {
...
+ *
+ * @param {FocusTrapComponent} component The FocusTrap component.
+ * @example
+ * const modal = document.querySelector("calcite-modal");
+ * const input = document.createElement("calcite-input");
+ * content.appendChild(input);
+ * await input.componentOnReady();
+ * await modal.updateFocusTrapElements();
+ * requestAnimationFrame(() => input.setFocus());
+ */
```
benelan, line 115: @macandcheese do we recieve the `example` tag via `docs-json`? It would be nice to add examples to the API ref table. Maybe something like the "show code snippet" action you have in the component preview. We get a lot of feedback about a lack of javascript examples, so this could be a good solution.

Thanks for adding it @driskull!

benelan (approved):
Nice!

PR #6147, opened 2022-12-21
refactor(modal, panel, popover)!: remove `focusId` param from `setFocus` method.

driskull:
BREAKING CHANGE: Removed `focusId` parameter from `setFocus` methods. When the `setFocus` method is called the first focusable element will be focused.


src/components/popover/popover.tsx
```diff
@@ -385,23 +384,12 @@ export class Popover
...
-
-    if (focusId === "close-button" && closeButtonEl) {
-      forceUpdate(closeButtonEl);
-      closeButtonEl.setFocus();
-
-      return;
-    }
-
-    focusFirstTabbable(this);
+    forceUpdate(this);
```
benelan, line 391: this should be `forceUpdate(this.el)` right?

src/components/popover/popover.tsx
```diff
@@ -552,7 +540,7 @@ export class Popover
   renderCloseButton(): VNode {
     const { messages, closable } = this;
     return closable ? (
-      <div class={CSS.closeButtonContainer}>
+      <div class={CSS.closeButtonContainer} key="close-button-container">
```
benelan, line 543: Can you use `CSS.closeButtonContainer` for the `key` attribute instead of hardcoding it

src/components/popover/popover.tsx
```diff
@@ -575,7 +563,7 @@ export class Popover
     ) : null;
 
     return headingNode ? (
-      <div class={CSS.header}>
+      <div class={CSS.header} key="header">
```
benelan, line 566: Reuse the string from `resources` here as well

benelan (approved):
A couple comments, but LGTM. I think that just leaves pick/value list as the only two left with `focusId` right?

Issue #6148, opened 2022-12-21
tab-nav - Add `selectedTab` property to support `calciteTabChange` event


geospatialem:
### Description

With our event payload changes from `event.target` (previously `event.detail`) users are unable to get at the selected `tab`.

### Acceptance Criteria

When using the `calciteTabChange` users can use the `event.target.selectedTab` to receive the selected tab.

### Which Component

`calcite-tab-nav`

### Esri team

Calcite (dev)

geospatialem:
Verified on `next.710` with the `tab-nav`'s property `selectedTitle`.

```js
const myTabNav = document.querySelector("calcite-tab-nav");
myTabNav.addEventListener("calciteTabChange", (evt) => {
  console.log(evt.target.selectedTitle);
});
```

PR #6149, opened 2022-12-21
feat(tab-nav): Add `selectedTitle` property

driskull:
feat(tab-nav): Add `selectedTitle` property. #6148

### DIFF
--- a/src/components/tab-nav/tab-nav.tsx
+++ b/src/components/tab-nav/tab-nav.tsx
@@ -61,7 +61,7 @@ export class TabNav {
    *
    * @readonly
    */
-  @Prop({ mutable: true }) selectedTabTitle: HTMLCalciteTabTitleElement = null;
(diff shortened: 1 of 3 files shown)


src/components/tab-nav/tab-nav.tsx
```diff
@@ -56,6 +56,13 @@ export class TabNav {
    */
   @Prop({ reflect: true }) syncId: string;
 
+  /**
+   * Specifies the component's selected tab-title.
+   *
+   * @readonly
+   */
+  @Prop({ mutable: true }) selectedTabTitle: HTMLCalciteTabTitleElement = null;
```
benelan, line 64: For consistency, should it be `selectedTitle`? I don't think we we add the component name in our props. 

For example `selectedItems` rather than `selectedTreeItems`, and `filteredItems` instead of `filteredComboboxItems`

driskull: Changed!

benelan (approved):
Thanks, LGTM!

PR #6150, opened 2022-12-21
build(vscode): add completion and hover support

benelan:
## Summary
Adds VSCode completion and hover functionality for our components. Our users can enable it as well by adding this to their project:

```jsonc
  // .vscode/settings.json
  "html.customData": [
    "./node_modules/@esri/calcite-components/dist/extras/vscode-data.json"
  ]
```
I don't even use VSCode, but happy holidays! 🎅 (Don't expect any more gifts)

image

### Resources
- https://stenciljs.com/docs/docs-vscode
- https://github.com/microsoft/vscode-custom-data

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,5 +6,9 @@
   "editor.formatOnSave": true,
   "editor.quickSuggestions": {
     "strings": true
-  }
+  },
+  "html.customData": [
+    "./dist/extras/vscode-data.json",
+    "./node_modules/@esri/calcite-components/dist/extras/vscode-data.json"
+  ]
 }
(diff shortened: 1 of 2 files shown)


driskull (approved):
🎅 🎄 🎁

benelan:
@driskull this is kind of on the edge between `build` and `feat`, what do you think? It might be nice to get the word out by getting it on our changelog. I also asked Kitty to write a blog post about it (after v1). Maybe part of a bigger "Calcite tooling" blog post.

driskull:
I'm not really sure it qualifies as a feat since it doesn't enhance our components in any way. It does enhance our developer experience so I do see why you would lean toward feat.

Maybe we can keep it as `build` and still blog about it? Could we add some doc somewhere in a different PR that might get notification?

PR #6154, opened 2022-12-21
fix(date-picker): modify weekStart value for ar locale

annierm18:
**Related Issue:** n/a

## Summary
Previously, date-picker had the weekStart value as "7" for the locale "ar", when it's actual weekStart as returned by intl object is "6". JSAPI team asked for the weekStart to be changed to "6" so that the week start day for the "ar" locale can be consistent across Esri web products.


driskull (approved):
👍

eriklharper (approved):
Thanks Aine!

PR #6161, opened 2022-12-22
refactor(popover)!: remove `toggle` method.

driskull:
BREAKING CHANGE: Removed the `toggle` method, use the `open` property instead.


eriklharper (approved):
LGTM! 👍 .  This will make the React crowd very happy!  Should we add any tests, visual or otherwise for this?

PR #6162, opened 2022-12-22
refactor(combobox-item)!: remove `toggleSelected` method.

driskull:
BREAKING CHANGE: Removed the `toggleSelected` method, use the `selected` property instead.


macandcheese (approved):
Nice 👍

PR #6166, opened 2022-12-22
feat(modal)!: Updates accepted `width` values, adds css variables for width and height

macandcheese:
BREAKING CHANGE: Updates accepted `width` values, adds css variables for width and height.

- The `width` property no longer accepts a custom width. Accepted values are `s`, `m`, `l`.
- Adds `--calcite-modal-width` and `--calcite-modal-height` css variables.


src/components/modal/modal.scss
```diff
@@ -6,6 +6,9 @@
  * @prop --calcite-modal-content-background: Specifies the background color of content placed in the `content` slot.
  * @prop --calcite-modal-content-padding: Specifies the padding of the modal `content` slot.
  * @prop --calcite-modal-scrim-background: Specifies the background color of the modal scrim.
+ * @prop --calcite-modal-width: Specifies a width of the modal. Will never exceed the width of the viewport. Does not apply when `fullscreen` is true.
```
macandcheese, line 9: I'm reconsidering not having these apply to fullscreen and instead always having them be honored. 

It seems more consistent that the components would adhere to any requested overrides.

src/components/modal/modal.tsx
```diff
@@ -119,10 +119,10 @@ export class Modal
   /** Specifies the size of the component. */
   @Prop({ reflect: true }) scale: Scale = "m";
 
-  /** Specifies the width of the component. Can use scale sizes or pass a number (displays in pixels). */
-  @Prop({ reflect: true }) width: Scale | number = "m";
+  /** Specifies the width of the component. */
+  @Prop({ reflect: true }) width: Scale = "m";
```
driskull, line 123: Would this be considered a breaking change?

driskull (approved):
👍 Looks good. If its a breaking change lets change the PR title and body to reflect that.

src/components/modal/modal.tsx
macandcheese: Yeah we should probably consider it as one. Need to clean up a few things on this, will update with breaking change when ready.

macandcheese:
@benelan @Esri/calcite-components-reviewers could you take a look at this one again? It will be a breaking change, as setting width via "width" prop is no longer available.

src/components/modal/modal.tsx
macandcheese: Yep, will update.

benelan: Does `scale` not change the width of the component? Having css vars for width/height, and then also having props for width/scale seems confusing. If the `scale` prop doesn't currently change the width in this PR, maybe we should update it to do so, since it accepts the same type as the `width` prop anyway? Then we can just remove the width property since we have the CSS var. @SkyeSeitz @ashetland thoughts?

macandcheese: Scale only changes the padding / text size. Width independently sets the modal to one of three default widths. Previously you could use the `| number` to pass a px value, or set directly on the modal, but the styles were moved inside the shadow root so the css vars replicate that.

We have a handful of components that have default widths that are able to be overridden - shell-panel, alert, etc., and I opened an issue about some of their inconsistent naming: https://github.com/Esri/calcite-components/issues/6172

benelan: Based on the issue you logged shoud this prop be renamed widthScale to be consistent with the other components?

macandcheese: Perhaps as a result of a more comprehensive audit, but short term I don't think we want to make a change. I still think it could just be a single default value that is only changed by css var. The linked issue is valid but post 1.0 since it would affect many other components.

src/components/modal/modal.tsx
```diff
@@ -302,9 +323,15 @@ export class Modal
 
   modalContent: HTMLDivElement;
 
-  private mutationObserver: MutationObserver = createObserver("mutation", () =>
-    this.updateFooterVisibility()
-  );
+  private mutationObserver: MutationObserver = createObserver("mutation", (mutations) => {
+    mutations.map((mutation) => {
```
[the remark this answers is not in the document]
macandcheese: It was changed to allow updates to the css vars to affect the Modal size after `connectedCallback`. I'll use the above suggestion and use a separate MutationObserver that isn't observing the subtree like the existing one for footer slot content is.

src/components/modal/modal.tsx
ashetland: I agree that the `width = s | m | l` prop doesn't seem to make much sense with this change. A default width that is customizable via the css var would seem fairly straightforward.

macandcheese: Let's try and tackle that change along with the other two issues - Modal has upcoming change after 1.0 with the proposed `fullscreen` changes and possibly removing `docked` when we add a `sheet`

src/components/modal/modal.tsx
```diff
@@ -268,30 +276,41 @@ export class Modal
...
-            flex: 1 1 auto !important;
-            max-height: unset !important;
-          }
-        }
-      `}
-      </style>
-    ) : null;
+    if (!this.fullscreen && (this.cssWidth || this.cssHeight)) {
+      return (
+        <style>
```
alisonailea, line 281: why put this here instead of in the .scss file? better yet, set this with CSS vars rather than static rules.

src/components/modal/modal.tsx
```diff
@@ -320,6 +343,12 @@ export class Modal
 
   contentId: string;
 
+  @State() modalSlottedWidth: number;
+
+  @State() cssWidth: string | number;
```
alisonailea, line 348: these should be css vars

src/components/modal/modal.tsx
macandcheese: This is using the existing pattern that was setting the `width` (as number) value. The css variable when in unpredictable format (calc, vw, etc) is not compatible with the css file media query, was I believe the original reason for it.

src/components/modal/modal.tsx
macandcheese: First one was stale code - removed it. The other two are set based on the css vars for use in `renderStyles()`.

benelan (approved):
Approving this because we need to get the breaking change in, and because the `width` property and MutationObserver existed before this PR.  We can address those concerns in one of the issues you linked in the comments

PR #6169, opened 2022-12-23
chore(modal): Remove invalid kind values

macandcheese:
Extracts only valid kind values from the interface.


anveshmekala (approved):
LGTM!

PR #6173, opened 2022-12-26
refactor(panel, shell-panel)!: Update available css variables

macandcheese:
BREAKING CHANGE: Update available css variables

- Removes width, height, and multiplier css variables for Panel - the component will now fill available space.
- Documents public css variables for Shell Panel

### DIFF
--- a/src/components/flow-item/flow-item.tsx
+++ b/src/components/flow-item/flow-item.tsx
@@ -13,7 +13,6 @@ import {
 } from "@stencil/core";
 import { getElementDir } from "../../utils/dom";
 import { HeadingLevel } from "../functional/Heading";
-import { Scale } from "../interfaces";
 import { CSS, ICONS, SLOTS } from "./resources";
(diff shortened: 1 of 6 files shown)


macandcheese:
Similar to #5983, this PR continues to implement proposed updates to our Layout component set.

Currently, the story to adjust the width of shell-panel, and their child panel, in an application layout is confusing and in many cases hard to accomplish. 

This removes all css variables used to adjust width / height from Panel, so it behaves more like a content container and should fill the space provided by any parent component. 

When used in a shell-panel, users should set css variables on the shell-panel to achieve a desired width. 

cc @asangma / @mitc7862 would like some eyes on this to ensure it won't break anything, from what I can tell with manual testing it is much easier to simply set a min / max / width on shell-panel than previously.

driskull (approved):
👍 Looks good!

asangma (approved):
Looks good based on my local demo that kind of replicates an app. Hopefully that will translate to the real apps.

Issue #6178, opened 2022-12-27
[Flow Item] Make `show-back-button` a private, non-documented property.


macandcheese:
### Description

The `show-back-button` is used by the parent Flow component and should not be publicly exposed.

### Proposed Advantages

The current property is meant to be set by Flow

### Which Component

Flow Item, Flow

### Relevant Info

cc @driskull

geospatialem:
Verified on `master`.

PR #6179, opened 2022-12-27
docs(flow-item): make showBackButton internal

benelan:
**Related Issue:** #6178

## Summary

Makes the showBackButton property used by the parent flow internal

### DIFF
--- a/src/components/flow-item/flow-item.tsx
+++ b/src/components/flow-item/flow-item.tsx
@@ -124,7 +124,7 @@ export class FlowItem
    *
    * @internal
    */
-  @Prop({ reflect: true }) showBackButton = false;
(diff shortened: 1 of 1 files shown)


src/components/flow-item/flow-item.tsx
```diff
@@ -120,8 +120,9 @@ export class FlowItem
   @Prop({ mutable: true }) messages: FlowItemMessages;
 
   /**
-   * When true, displays a back button in the header.
    * When `true`, displays a back button in the component's header.
+   *
+   * @internal
    */
   @Prop({ reflect: true }) showBackButton = false;
```
driskull, line 127: Can we update it to not reflect?

driskull (approved):
👍

PR #6181, opened 2022-12-27
feat: add part designation to block-section heading so that it can be custom styled

MikeTschudi:
**Related Issue:** _none_

## Summary

Added a `part` attribute to the span holding the block-section's heading so that the heading can be styled by the consuming app.

For example, our PE has requested that "Terms of Use" match the styling of the other headings in
image

With this modification, we can add two examples to this repository's demos/block.html:
image


driskull:
Hey @MikeTschudi, this PR looks 💯 

We just need to strategize on how and when we are going to start adding and documenting CSS part selectors for the design system. I think we need to wait on this PR so we can discuss how to best do this and support it in the long run.

I'll have @jcfranco take a look when he's back

MikeTschudi:
@driskull, makes sense to me. I puzzled how to add it to the interactive doc.

driskull:
Yeah, I think we may have to add a custom JSDoc tag for CSS Parts.

MikeTschudi:
@driskull, @jcfranco, anything that I can do to help the doc task that Matt identified?

jcfranco:
@MikeTschudi Thanks for the PR. 🎉 As Matt mentioned, we need to figure out what our strategy will be before we can proceed with this. We'll look into CSS parts after 1.0 is out of beta. 

cc @geospatialem @macandcheese @SkyeSeitz

jcfranco:
@MikeTschudi Discussed this with the team and we decided to wait until the design system matures a bit (e.g. providing guidelines and patterns to promote consistency) before we can move towards supporting parts. We'd like to thank you again for your time and effort on this.

Going back to the original use case, can we set something up to discuss an alternative solution or workaround? cc @geospatialem @ashetland @SkyeSeitz

MikeTschudi:
@jcfranco, we have an acceptable alternative solution via calcite-block:
image

Issue #3097, opened 2021-09-23
Docs: generalize default slots that are too specific


benelan:
### Description 
Some of the default slots are too specific which has caused some confusion for users who want to add their own custom content. As an example, `calcite-action`'s default slot is "A slot for adding a `calcite-icon`." However, users can also add their own custom content, such as images to this default slot. 

@macandcheese suggested that all of the default slots that aren't specifically used for child components should be changed to something generic, such as "A slot for adding custom content". 

Alternatively, for components like `calcite-action`, we can say something like "A slot for adding custom content, primarily `calcite-icon`." Any thoughts/suggestions on this?

cc @caripizza

macandcheese:
I prefer the "generic" copy for non-child-component slots - although maybe I'm just opposed to that specific recommendation of `calcite-icon` inside `calcite-action`, since the built in icon prop should be used _almost_ all the time there.

benelan:
Maybe `calcite-action` isn't a good example since it has a prop for the icon. Another example would be `calcite-scrim`. It could be generic, but it may be better to provide some additional insight so people use it correctly. For example: "

> A slot for adding custom content, primarily loading states and messages.

geospatialem:
Verified on `master` across component descriptions for default slots.

PR #6190, opened 2022-12-28
docs: consistent slot messaging across components

geospatialem:
**Related Issue:** #3097  

## Summary
Adds consistency to slot documentation across components. Particular focus on default slots to describe as "A slot for adding custom content". 

Also includes an in-depth review and update across all component slots.


benelan (approved):
👍

PR #6192, opened 2022-12-28
ci: fix timing issue in esri team labeler action

benelan:
**Related Issue:** #5851

## Summary
The Esri product team labeler was running into the same timing issue as another action linked above.


driskull (approved):
👍

Issue #5969, opened 2022-12-09
Input-date-picker: Incorrect date separator appearing for "it-CH", incorrect date appearing for "zh-CN" and "zh-TW", and cannot select date in beta.98


annierm18:
### Actual Behavior

**Issue 1:**
With input-date-picker [Calcite Input Date Picker](http://calcite-dev.esri.com:3008/demos/input-date-picker.html), the separator in between the dates did not change between "it-CH" and "it". I saw it was changing as expected for other locales like for fr-CH --> fr for "it-CH" but saw "25/3/2023".  In beta.98,  when I put "it-CH", I see "." (see issue 3 below for more info on beta.98).

**Issue 2:**
First, the date looks like this right when I change the lang (see standard small component)
![Screenshot (98)](https://user-images.githubusercontent.com/25759835/206760931-167efe1d-c4d5-4469-a14b-02c5153b33d2.png)

Then, it looks like this once I select another date (see standard small component)
![Screenshot (99)](https://user-images.githubusercontent.com/25759835/206760961-d405ba98-8924-4cac-9a94-093974a3a5f6.png)

Lastly, changes to this after I select a date once more (see standard small component)
![Screenshot (100)](https://user-images.githubusercontent.com/25759835/206760975-9a0951a6-40f4-4664-ac31-9ae54e8914b5.png)

**Issue 3:** 
Reproduced in Chrome & Edge: beta.98  with or without the lang attribute included, I can click on the input and get the calendar to appear, but cannot select anything. When I try to select, the calendar dropdown goes away. Note: when I put "it-CH", I see "." in the input field as expected, but I cannot confirm the behavior once a date is clicked. 

### Expected Behavior

**Issue 1:**
I would expect the separator to be ".". Correct separator and date formatting for "it-CH" can be found [here](https://icu4c-demos.unicode.org/icu-bin/locexp?d_=en&_=it_CH). It works correctly in beta.97 in Codepen. Interestingly, the "." appears correctly for "fr-CH" in the same environment where it does not work for "it-CH". 

**Issue 2:**
I would expect the date to remain in the correct...e.g. 2022 would stay as 2022 and not get converted somehow to 190X.

**Issue 3:** 
On beta.98, I would expect to be able to select dates from the dropdown.

### Reproduction Sample

**Issue 3:** https://codepen.io/ainermfc/pen/qBKvQRx?editors=100

### Reproduction Steps

**Issue 1:**
1. Open input date picker 
2. Add lang="it-CH"
[...]

Issue #5886, opened 2022-12-03
Input Date Picker: internal input out of sync with component's locale


benelan:
### Actual Behavior

The internal input's displayed value does not update when changing the component's locale after the initial rendering.

### Expected Behavior

The internal input's displayed value updates to the correct format when changing the component's locale

### Reproduction Sample

https://codepen.io/benesri/pen/mdKQejR?editors=1010

### Reproduction Steps

1. Click the "Change locale" button, and notice the input's displayed value stays the same
2. Switch the `calcite-components` version from `next.622` to `next.621`
3. Do step 1, notice the input's displayed value changes to match the new locale

### Reproduction Version

next.622

### Relevant Info

Looks like the regression came from #5459 based on the [changes between next.622 and next.621](https://github.com/Esri/calcite-components/compare/v1.0.0-next.621...v1.0.0-next.622).

I noticed this while working on a similar issue #5855, but unfortunately the solution I have for that one doesn't fix this one as well.

 Interestingly, in the repro sample you can't select a date from the internal `date-picker` in both `next.622` and `next.621`. That's fixed now though so I didn't spend any time investigating, it seems unrelated to this issue.

I can tackle this issue afterwards unless you have any ideas @eriklharper.

### Regression?

next.621

### Esri team

Calcite (dev)

PR #6197, opened 2022-12-29
fix(input-date-picker): update input value when changing locale

benelan:
**Related Issue:** #5886 #5969
## Summary
Makes sure the internal input's value is updated/rerendered when the `lang` property changes


driskull (approved):
👍

eriklharper (approved):
Awesome man!  Thanks for fixing this :)

src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -912,17 +914,21 @@ export class InputDatePicker
...
+
   private formatNumerals = (value: string): string =>
     value
       ? value
           .split("")
           .map((char: string) =>
-            numberKeys.includes(char)
-              ? numberStringFormatter.numberFormatter.format(Number(char))
+            this.commonDateSeparators?.includes(char)
+              ? this.localeData?.separator
```
benelan, line 929: As a note before I merge, I added this commonDateSeparator code as a convienence in #5488. It used to automatically convert incorrect date separators as you typed, but it shouldn't have affected anything if the separator was typed correctly. 

Here is a demo from `beta.97`, If you type "2020-1-1" it will automatically convert to "2020.1.1". https://codepen.io/benesri/pen/qByZmXd

I removed it in #5887, because at that point the auto conversion no longer worked and it was causing JS Handle errors in the e2e test I added. For some reason removing it caused "Issue 1" in #5969, so this PR adds it back.

But at some point we should probably figure out why the auto conversion is no longer working and why the date separator doesn't work without this code, which wasn't meant for that.

cc @eriklharper

eriklharper: I bet the reason the e2e environment had a problem with it could be due to the encoding on the characters, but yeah this is a good item to research further.  Just an idea, but one way to set `commonDateSeparators` are to get the separator chars from the locales themselves using the Intl.DateTimeFormat.formatToParts api that way we can ensure the correct character is being used.

benelan: The separator being used is coming from `localeData`, so that will already always be correct. But I'll put in a PR in a sec to make sure all date separators get converted since the code relies on it now lol

PR #6205, opened 2022-12-29
fix(modal): close button does not change header height

benelan:
**Related Issue:** #1707

## Summary
`padding-block` was changing the header height for me when toggling the back button. Removing it didn't seem to change the close button positioning in any way. `padding-inline` was added in #5210 for reference

### DIFF
--- a/src/components/modal/modal.scss
+++ b/src/components/modal/modal.scss
@@ -148,7 +148,6 @@
     border-none
     bg-transparent;
   border-start-end-radius: theme("borderRadius.DEFAULT");
-  padding-block: var(--calcite-modal-padding-internal);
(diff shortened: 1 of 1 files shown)


driskull (approved):
👍

benelan:
The change did cause some minor chromatic diffs, but they don't seem noticable to the human eye. I think it's worth fixing the bug, but @ashetland @SkyeSeitz can I get a design approval/denial?

SkyeSeitz:
> The change did cause some minor chromatic diffs, but they don't seem noticable to the human eye. I think it's worth fixing the bug, but @ashetland @SkyeSeitz can I get a design approval/denial?

Visual changes approved.

Issue #6201, opened 2022-12-29
calcite-input non-responsive increment/decrement 


itsminj:
### Actual Behavior

When I set a minimum value on the calcite-input (e.g. 3), and input a value such as 1 and try to click increment or decrement, nothing happens. The same occurs when I set a max value (e.g. 40) and input a number such as 44.

### Expected Behavior

A user should be able to still click increment/decrement a value even if it is lower or higher than the minimum value or maximum value, respectively.

### Reproduction Sample

https://codepen.io/mjkim12/pen/JjBXWzV

### Reproduction Steps

1. Go to the calcite-input
2. Enter a number lower than the minimum or higher than the maximum
3. Notice how that clicking on the increment or decrement doesn't do anything

### Reproduction Version

beta.99

### Impact

Using this component for ArcGIS Online

### Esri team

ArcGIS Online

driskull:
@eriklharper it seems like the min or max number should be set if there is no value or the value is out of range.

@geospatialem do you think this should be pre 1.0?

geospatialem:
@driskull We could add it to the January milestone and if we're unable to get it in on time have it reallocated to the February milestone since its a bug. WDYT?

driskull:
Yep. sounds good

eriklharper:
> @eriklharper it seems like the min or max number should be set if there is no value or the value is out of range.

Doesn't sound like the native number input sets a min or max if no value is provided or if the provided min/max value is not a valid number, so I think we should follow suit in our component to match that:

![image](https://user-images.githubusercontent.com/821864/210016062-33024f0a-0bd9-4b7d-8a58-748a0423749a.png)

Also, we should probably remove `type="number"` support in `calcite-input` before 1.0 in favor of `calcite-input-number` and resolve this behavior in that component, if it exists.  Thoughts?

driskull:
It does set the min value if the down button is pressed

benelan:
So if the value 44 and the max is 46, when the deincrement button is clicked would the value should change to 44

driskull:
yeah. that's how the native input works.

benelan:
Cool, I can take this issue

PR #6207, opened 2022-12-29
fix(input, input-number): increment/decrement to the min/max when value is below/above

benelan:
**Related Issue:** #6201

## Summary
Follows the native input's behavior by incrementing/decrementing to the `min`/`max` property when `value` is below/above


driskull (approved):
Awesome 👍 👏

Issue #6198, opened 2022-12-29
Input Date Picker: internal input doesn't display numerals in the numberingSystem


benelan:
### Actual Behavior

The internal input no longer displays numerals in the specified numbering system

### Expected Behavior

The internal input should display numerals in the specified numbering system

### Reproduction Sample

https://codepen.io/benesri/pen/xxJZvmB

### Reproduction Steps

1. Type numbers into the input and notice they are `latn` numerals, even though the `numberingSystem` property is `arab`
2. Switch the version to `next.621`
3. Repeat step 1 and notice the numbers are `arab` numerals

### Reproduction Version

next.703

### Relevant Info

This is a regression from #5459

Related issue: #5886

@eriklharper do you have bandwidth to help with this one? The component is completely different than the last time I looked at it and most of the i18n code I added is gone. You have a better idea of what's going on in the component right now and it would probably take me a while to get back up to speed. If you're swamped I can take a stab at it though, let me know.

### Regression?

next.621

### Esri team

Calcite (dev)

eriklharper:
> @eriklharper do you have bandwidth to help with this one? The component is completely different than the last time I looked at it and most of the i18n code I added is gone. You have a better idea of what's going on in the component right now and it would probably take me a while to get back up to speed. If you're swamped I can take a stab at it though, let me know.

Yeah, I can assign myself to this, I'm going to be adding numberingSystem support to Input Time anyway, so all the info will be fresh in my head anyway.

benelan:
Awesome, thanks! Let me know if you have any questions about the `numberingSystem` utils I wrote as you work through the issues.

eriklharper:
After some fresh thinking on this, I'm wondering if the computer's Input Source should be the "source of truth" when it comes to what characters display as the result of typing into inputs, regardless of any component-specific settings.  Meaning, if I have a U.S. keyboard layout, I, as a user probably expect to see latin numerals anywhere I type, even on Arabic date picking/text/number inputs.  By Input Source, I'm referring to the setting of the same name in macOS:

![image](https://user-images.githubusercontent.com/821864/214132468-289bcd3b-6b0c-48e0-94e8-e9c3aa0c907d.png)

The logical outcome to a setting like this is that the component would still render read-only and committed content in the specified language/character system, but respect the user's Input Source setting for data entry only.  This could mean that when the user focuses any editable portion of UI, they would see characters native to the Input Source, but after committing the value and blurring the input the component would display those characters in the specified language on the component.

[...]

PR #6208, opened 2022-12-29
refactor(input-date-picker): use Intl to get unique date separators

benelan:
**Related Issue:** https://github.com/Esri/calcite-components/pull/6197#discussion_r1059175604

## Summary

Makes sure all date separators get converted into the one used by the component's `lang`. This shouldn't be necessary, but better safe than sorry. I _think_ fixing #6198 will make incorrect date separators automatically convert to the correct one when typing again. They seem to both be caused by lifecycle changes that stopped localizing/updating/rerendering the internal input's value per character typed, which is necessary for `numberingSystem` and the date separator conversion.

### DIFF
--- a/src/utils/date.ts
+++ b/src/utils/date.ts
@@ -1,11 +1,26 @@
 import { DateLocaleData } from "../components/date-picker/utils";
-import { numberStringFormatter } from "./locale";
+import { numberStringFormatter, locales } from "./locale";
 
 export interface HoverRange {
   focused: "end" | "start";
   start: Date;
   end: Date;
 }
+
+const getUniqueDateSeparators = () => {
+  const uniqueDateSeparators = new Set();
+  const dummyDate = new Date(2020, 4, 20);
+  locales.forEach((locale) => {
+    const dateTimeFormat = new Intl.DateTimeFormat(locale, { year: "numeric", month: "numeric", day: "numeric" });
+
+    const parts = dateTimeFormat.formatToParts(dummyDate);
+    const separator = parts.filter((part) => part.type === "literal")[0]?.value;
+    uniqueDateSeparators.add(separator);
+  });
+  return [...uniqueDateSeparators];
+};
+
+export const dateSeparators = getUniqueDateSeparators();
 
 /**
  * Check if date is within a min and max
@@ -199,7 +214,7 @@ export function parseDateString(
   const { day, month, year } = datePartsFromLocalizedString(string, localeData);
   return {
     day: parseInt(day),
-    month: parseInt(month) - 1, // this subtracts by 1 because the month in the Date contructor is zero-based https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getMonth
+    month: parseInt(month) - 1, // this subtracts by 1 because the month in the Date constructor is zero-based https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getMonth
     year: parseInt(year)
   };
 }
--- a/src/components/input-date-picker/input-date-picker.tsx
+++ b/src/components/input-date-picker/input-date-picker.tsx
@@ -20,7 +20,8 @@ import {
   dateFromISO,
   dateToISO,
   dateFromLocalizedString,
-  datePartsFromLocalizedString
+  datePartsFromLocalizedString,
+  dateSeparators
 } from "../../utils/date";
 import { HeadingLevel } from "../functional/Heading";
 import { CSS } from "./resources";
@@ -918,14 +919,12 @@ export class InputDatePicker
     );
   }
 
-  private commonDateSeparators = [".", "-", "/"];
-
   private formatNumerals = (value: string): string =>
     value
       ? value
           .split("")
           .map((char: string) =>
-            this.commonDateSeparators?.includes(char)
+            dateSeparators?.includes(char)
               ? this.localeData?.separator
               : numberKeys?.includes(char)
               ? numberStringFormatter?.numberFormatter?.format(Number(char))


src/utils/date.ts
```diff
@@ -1,12 +1,27 @@
...
 
 export interface HoverRange {
   focused: "end" | "start";
   start: Date;
   end: Date;
 }
 
+const getUniqueDateSeparators = () => {
+  const uniqueDateSeparators = new Set();
+  const dummyDate = new Date(2020, 4, 20);
```
driskull, line 12: giphy


Could have done pi day or star wars day

 ¯\_(⊙︿⊙)_/¯

src/utils/date.ts
```diff
@@ -1,12 +1,27 @@
...
+    const dateTimeFormat = new Intl.DateTimeFormat(locale, { year: "numeric", month: "numeric", day: "numeric" });
+
+    const parts = dateTimeFormat.formatToParts(dummyDate);
+    const separator = parts.filter((part) => part.type === "literal")[0]?.value;
+    uniqueDateSeparators.add(separator);
+  });
+  return [...uniqueDateSeparators];
+};
+
+export const dateSeparators = getUniqueDateSeparators();
```
driskull, line 23: Do we expect this to grow with new locales added? I guess this is safer if locales change.

driskull (approved):
👍 💯

src/utils/date.ts
benelan: If we add support for additional locales they should be added to the array I'm iterating through, so any additional separators would be represented here. Or am I misunderstanding?

driskull: Yeah that's correct. I was just wondering if we expect the list to grow or if we could just generate the const once and not have it dynamic like this.

driskull: Maybe we should add a spec test for this?

benelan: Yeah generating the const once was what I originally did, which came out to `["/", ".", "-"]` after trimming. I'm fine with either. Realistically this code shouldn't make our break our localization lol. Something in the component's lifecycle needs to be fixed, this is just a temporary bandaid.

eriklharper (approved):
This looks good to me, although I would like to sync up on our approach to locales in general to see if maybe the best solution longer term since we're moving toward full translation implementations for all the components (eventually) is to just have these characters stored as part of the hard-coded locale data that we get from the Intl team.  Franco told me not to worry about locales Esri doesn't officially support, so it would be a narrower list to work from anyway and we wouldn't have to rely on this somewhat-hacky way of retrieving a locale's separator character.  Another good reason to opt for storing this eventually in the locale data is that our Intl team can provide those exact char codes to us rather than rely on what the browser spits out, which would improve predictability and also take advantage of our locale subject matter experts.

benelan:
Yeah storing it in an array is what I originally did but you asked me to generate it dynamically. If you're cool with the way it currently is then we can stick with that

driskull:
> Yeah storing it in an array is what I originally did but you asked me to generate it dynamically. If you're cool with the way it currently is then we can stick with that

Yep 👍

benelan:
>  I would like to sync up on our approach to locales in general to see if maybe the best solution longer term 

I would still like to sync up with the locale team as well once Franco gets back and things calm down a bit. There are other places where we can potentially use hard coded values instead of `Intl` data. As an example, we store locale's numerals in the nls data, although for the most part it's just latin numerals and doesn't match up with `Intl'`s numberingSystem data. But we could potentially refactor the numberingSystem stuff to use those at some point, if that's an option. 

https://github.com/Esri/calcite-components/blob/0038162ce593a522f48b985ea290788aa55e257e/src/components/date-picker/assets/date-picker/nls/ar.json#L13

Not sure how big of a benefit it would be, but it may avoid weird stuff like #6209 (assuming that's a bug on `Intl`'s end).

benelan:
> we wouldn't have to rely on this somewhat-hacky way of retrieving a locale's separator character

We shouldn't/didn't need this for retrieving a locale's separator. This code was originally just to auto convert the separator to the one specified in `this.localeData` if someone typed the wrong separator, as a convienence. 

But the separator logic from `localeData` and the auto conversion both broke at some point in one of the refactors since `beta.97`. Adding this back fixes the stuff that should be covered by `localeData` but doesn't fix its intended auto-conversion purpose lol. The component definitely needs a refactor when we have time.

eriklharper:
> Not sure how big of a benefit it would be, but it may avoid weird stuff like #6209 (assuming that's a bug on `Intl`'s end).

Cool, yeah this is a great example of a case I had in mind where relying on our own Intl team to provide the proper characters could circumvent platform issues beyond our control.  Thanks for sharing!

eriklharper:
> > we wouldn't have to rely on this somewhat-hacky way of retrieving a locale's separator character
> 
> We shouldn't/didn't need this for retrieving a locale's separator. This code was originally just to auto convert the separator to the one specified in `this.localeData` if someone typed the wrong separator, as a convienence.

I guess I was thinking of the pattern I adopted to glean the separator character for time formats using the Intl.DateTimeFormat api.  I assumed the same thing was happening for Date formats too.  I think it was a cool experiment and could be utilized indefinitely, but I think in the longer term it makes more sense to just store these in our own locale data formats that we get from the Intl team.  It is cool (in theory) to just make components like Date and Time inputs utilize all the available browser-supported locales, but because they don't have 2-way translation for all the non-latn locales (yet, if ever) our approach to translation will be necessary anyway to fully support typing in non-latn keyboard layouts, etc (Arabic mostly but there are quite a few other writing systems as well).

benelan:
>  It is cool (in theory) make components like Date and Time inputs utilize all the available browser-supported locales, but because they don't have 2-way translation for all the non-latn locales (yet, if ever) our approach to translation will be necessary anyway to fully support typing in non-latn keyboard layouts

Why would translations be required for users to type using non-latn keyboard layouts? Translations don't have anything to do with user inputted values, they are for the hardcoded strings that we use in our components. I also don't think we every plan on supporting two way translation, so that won't be an issue either.

Issue #6139, opened 2022-12-20
Remove `focusId` params from all `setFocus` methods


driskull:
### Description

Remove all `focusId` params from `setFocus` methods.

### Proposed Advantages

Ideally, the focus should be on the first focusable element as defined by a11y guidelines and consistent with how `delegatesFocus` works.

If we have `focusId` then there could be UX inconsistencies where developers are determining where to focus first. Ideally, this is handled by the component either by focusing the first element or using a focus trap to do so.

### Which Component

20 results - 5 files

src/components/modal/modal.tsx:
  383     *
  384:    * @param focusId
  385     */
  386    @Method()
  387:   async setFocus(focusId?: "close-button"): Promise<void> {
  388      await componentLoaded(this);

  391  
  392:     if (closeButtonEl && focusId === "close-button") {
  393        return focusElement(closeButtonEl);

src/components/panel/panel.tsx:
  323     *
  324:    * @param focusId
  325     */
  326    @Method()
  327:   async setFocus(focusId?: "back-button" | "dismiss-button"): Promise<void> {
  328      await componentLoaded(this);

  331  
  332:     if (focusId === "back-button") {
  333        backButtonEl?.setFocus();

  336  
  337:     if (focusId === "dismiss-button") {
  338        closeButtonEl?.setFocus();

src/components/pick-list/pick-list.tsx:
   14  import {
   15:   ListFocusId,
   16    calciteListItemChangeHandler,

  272     *
  273:    * @param focusId
  274     */
  275    @Method()
  276:   async setFocus(focusId?: ListFocusId): Promise<void> {
  277      await componentLoaded(this);
  278  
  279:     return setFocus.call(this, focusId);
  280    }

src/components/popover/popover.tsx:
  387     *
  388:    * @param focusId
  389     */
  390    @Method()
  391:   async setFocus(focusId?: "close-button"): Promise<void> {
  392      await componentLoaded(this);

  395  
  396:     if (focusId === "close-button" && closeButtonEl) {
  397        forceUpdate(closeButtonEl);

src/components/value-list/value-list.tsx:
   16  import {
   17:   ListFocusId,
   18    calciteListFocusOutHandler,

  416     *
  417:    * @param focusId
  418     */
  419    @Method()
  420:   async setFocus(focusId?: ListFocusId): Promise<void> {
  421      await componentLoaded(this);
  422  
  423:     return setFocus.call(this, focusId);
  424    }

driskull:
[...]

PR #6212, opened 2022-12-30
docs: update setFocus method across components

geospatialem:
**Related Issue:** #6139

## Summary
Adds context ~to the `panel` and `popover` docs~ **across our components** for the `setFocus` refactor. ~~Where when called will set focus on the component's first focusable element.~~ 

Where when called will either set focus on the component or its first focusable element. Some components first focusable element will be a close button.

Unchanged components with `setFocus` where focus is set on the component, such as `calcite-action`, and the context of "Set focus on the component" is still valid.

### DIFF
--- a/src/components/action-bar/action-bar.tsx
+++ b/src/components/action-bar/action-bar.tsx
@@ -228,7 +228,7 @@ export class ActionBar
   }
 
   /**
-   * Sets focus on the component.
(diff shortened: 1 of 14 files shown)


driskull (approved):
👍 Looks good!

I think all setFocus methods should be consistent with this text.

Issue #6123, opened 2022-12-20
`calcite-list-item` missing pointer cursor style


mpayson:
### Actual Behavior

`calcite-list-item` does not use a pointer style for the cursor. Looks like this was fixed in https://github.com/Esri/calcite-components/issues/5880 but the pointer may have changed again?

cc @ethanbdev @AdelheidF @geospatialem 

### Expected Behavior

`calcite-list-item` uses the pointer style so the items appear clickable, see gif under "Expected behavior" here: https://github.com/Esri/calcite-components/issues/5880

### Reproduction Sample

https://codepen.io/mpayson/pen/mdjdNBq

### Reproduction Steps

1. Open code pen
2. Hover over an item

### Reproduction Version

beta.99

### Regression?

next.660

### Esri team

ArcGIS Online

geospatialem:
This seems to be a result of the `selection-mode` attribute, which is set to `"none"` by default. If adding `selection-mode="multiple"` or `"single"` the cursor is present. 

```html
<calcite-list selection-mode="single">
...
</calcite-list>
```

This seems to be the expected behavior, but we could add documentation to the component for additional context.

mpayson:
Ah interesting, is there a way to do this without the visual selected state? Our use case is a user selects a processing tool from a list of options and that tool gets added to a processing model, I'm not sure it makes sense to keep the tool highlighted after the user chooses an option

geospatialem:
> Ah interesting, is there a way to do this without the visual selected state?

Not at this time. Prior to the refactor we had a `nonInteractive` property, but [opted to remove it as we couldn't justify a use case compared to a `selectionMode` of `"none"`](https://github.com/Esri/calcite-components/discussions/4356#discussioncomment-2481815). 

The visual feedback is useful for more diverse audiences, so we decided to remove the `nonInteractive` prop. We'd likely need to consider the enhancement after next month's release after some team discussion.

> Our use case is a user selects a processing tool from a list of options and that tool gets added to a processing model, I don't think it makes sense to keep the tool highlighted after the user chooses an option. They can select the same option twice and that tool should get added to the model twice, for example, and they aren't choosing one exclusive option

Good to know, don't believe we knew of a use case so good to get additional context. 

For us to re-evaluate after next month's release, could you or the team add additional visuals for the expected behavior so when we are able to convene we can determine some viable options for consideration in a future release?

ethanbdev:
[...]

PR #6213, opened 2022-12-30
fix(list-item): use pointer cursor when selection mode is none

benelan:
**Related Issue:** #6123

## Summary
Add pointer cursor on hover

### DIFF
--- a/src/components/list-item/list-item.scss
+++ b/src/components/list-item/list-item.scss
@@ -19,7 +19,7 @@
   );
 }
 .container:hover {
-  @apply bg-foreground-2;
+  @apply bg-foreground-2 cursor-pointer;
 }
 
 .container--border-selected {
--- a/src/components/list-item/list-item.scss
+++ b/src/components/list-item/list-item.scss
@@ -48,10 +48,6 @@
     p-0;
 }
 
-.content-container--selectable {
-  @apply cursor-pointer;
-}
-
 tr,
 td {
   @apply focus-base;


src/components/list-item/list-item.scss
```diff
@@ -19,7 +19,7 @@
   );
 }
 .container:hover {
-  @apply bg-foreground-2;
+  @apply bg-foreground-2 cursor-pointer;
```
driskull, line 22: 👍 LOOKS GOOD.

I think we could probably remove this though...

https://github.com/Esri/calcite-components/blob/1fc4e760b8924ef8903f1283b7fa45e6d40e7231/src/components/list-item/list-item.scss#L51-L53

driskull (approved):
I'm ok with this, my only reservation is that if we allow these list items to basically become buttons when selection is disabled, then I think we lose a11y support because a keyboard user won't know these are actionable.

cc @geospatialem

geospatialem:
@benelan This looks great, nice work! 🎆 

> I'm ok with this, my only reservation is that if we allow these list items to basically become buttons when selection is disabled, then I think we lose a11y support because a keyboard user won't know these are actionable.

Agree with Matt's comment regarding disabled states and a11y if we try to implement in the future.

geospatialem:
@benelan Can this be merged to verify 6123?

benelan:
Woops sorry I thought I already merged it lol

Issue #5544, opened 2022-10-22
Bug: the end date in date picker range is a square


benelan:
### Actual Behavior

The end date in date-picker range is a square 

![image](https://user-images.githubusercontent.com/10986395/197309240-d2b2fee3-8f03-4497-9c46-9d9bf655439b.png)

### Expected Behavior

The end date to match the nice, smooth start date #dontbeasquare

### Reproduction Sample

https://codepen.io/benesri/pen/NWMQzyQ

### Reproduction Steps

1. Open the sample
2. Look at the shapes

### Reproduction Version

beta.94 through next

### Regression?

beta.93

### Esri team

Calcite (dev)

geospatialem:
Observed on `hover`, or `focus` the end date returns to a filled circle.

![date-picker-range-hover](https://user-images.githubusercontent.com/5023024/197826033-4bd14d0f-5535-4126-919e-138ddaadd125.gif)

geospatialem:
Verified in `next.729` ✨ 

![image](https://user-images.githubusercontent.com/5023024/213800590-61e3b387-90f1-45e2-abec-80d845db7cf3.png)

PR #6216, opened 2023-01-03
fix(date-picker): end-range is now rounded and has the correct box-shadow

Elijbet:
**Related Issue:** #5544

## Summary

Improvements to `date-picker-day` styling include:

- `end-range` is now rounded 
- has the correct `box-shadow`
- styling is consistent for both `dir="ltr"` and `dir="rtl"`


src/components/date-picker-day/date-picker-day.scss
```diff
@@ -221,7 +221,7 @@
 }
 
 :host([highlighted]:not([active]:focus)) .day {
-  @apply text-color-1 rounded-none;
+  @apply text-color-1 rounded-full;
```
anveshmekala, line 224: I don't think `rounded-full` is required here.

anveshmekala (approved):
LGTM!

src/components/date-picker-day/date-picker-day.scss
Elijbet: This change is what fixes the bug.

eriklharper:
End selected highlight circle looks good!  One thing I'm noticing that we should also do is make sure the selected end date's hover styles match that of the start date.  On your branch looks like there is still a visible gray square that appears behind the blue circle on just the end date of the range.  We should match that to look just like the start date, where there is no gray square behind it:

image

src/components/date-picker-day/date-picker-day.scss
anveshmekala: we don't need to add `rounded-full` here as it is the default style applied on `.day`

Elijbet:
> End selected highlight circle looks good! One thing I'm noticing that we should also do is make sure the selected end date's hover styles match that of the start date. On your branch looks like there is still a visible gray square that appears behind the blue circle on just the end date of the range. We should match that to look just like the start date, where there is no gray square behind it:
> 
> image

Yeah, I noticed that as well after the fact and working on it now. Good catch :)

Elijbet:
I have a couple of design questions on `date-picker` (mostly hover behavior). @SkyeSeitz @ashetland @macandcheese 

- Should there be a hover style for unselected dates outside of the range, such as an empty bordered circle?
- Should this also match the hover style of unselected dates inside the range, aside from the blue-to-white linear transition?
- What should it do when selected range ends are hovered on? 

This could be separate issues for after 1.0, it just seems that the `date-picker` design is a bit incomplete.

alisonailea:
I would defer to the designers here but there are use cases where a user would want to know what date they're hovering over after they have selected a date range. This goes double for someone interacting with the date picker via keyboard.

Same for highlighting a date within a selected range.

Elijbet:
> I would defer to the designers here but there are use cases where a user would want to know what date they're hovering over after they have selected a date range. This goes double for someone interacting with the date picker via keyboard.
> 
> Same for highlighting a date within a selected range.

Yeah, right now the only highlighting I see is a slightly darker text color, which I only noticed after I specifically looked for a distinction.

ashetland:
Hey folks, as of beta.97 it looks like all the hover states had been sorted out. Am I missing / misundestanding something?

https://user-images.githubusercontent.com/108549080/211356583-94f302ab-8fd7-4265-b356-12c19d6e5dc0.mp4

SkyeSeitz:
> Hey folks, as of beta.97 it looks like all the hover states had been sorted out. Am I missing / misundestanding something?
> 

Tis upon clicking outside the `date-picker`

https://user-images.githubusercontent.com/29716057/211389269-3d0c2986-7189-423b-8b2f-b202d4b0473a.mov

ashetland:
> > Hey folks, as of beta.97 it looks like all the hover states had been sorted out. Am I missing / misundestanding something?
> 
> Tis upon clicking outside the `date-picker`
> 
>  Screen.Recording.2023-01-09.at.11.13.57.AM.mov

I thought @Elijbet was asking about hover states not the square end-date.

SkyeSeitz:
> > > Hey folks, as of beta.97 it looks like all the hover states had been sorted out. Am I missing / misundestanding something?
> > 
> > 
> > Tis upon clicking outside the `date-picker`
> > Screen.Recording.2023-01-09.at.11.13.57.AM.mov
> 
> I thought @Elijbet was asking about hover states not the square end-date.

You are correct- sorry! It would help if I read the latest comment, not just the issue title 😅

Elijbet:
> Hey folks, as of beta.97 it looks like all the hover states had been sorted out. Am I missing / misundestanding something?
> CleanShot.2023-01-09.at.08.20.03.mp4

Wow, ok. So turns out that because I had my contrast just one notch up, the grey highlighting completely faded out. I wasn't even aware it was there. Even with it down to normal I can barely see it. 

Screenshot 2023-01-09 at 12 22 37 PM

jcfranco:
@Elijbet Was there anything pending for this? It looks like it's good to go. Let me know if I missed something.

Also, can you mark this as ready for review? I think we can skip draft reviews by default unless there's an explicit need for early feedback.

jcfranco:
One more thing, since `date-picker-day` is internal, can you use the public component in the PR title's scope: 

`fix(date-picker): end-range ...`

?

Elijbet:
> @Elijbet Was there anything pending for this? It looks like it's good to go. Let me know if I missed something.
> 
> Also, can you mark this as ready for review? I think we can skip draft reviews by default unless there's an explicit need for early feedback.

There is more working code on this branch that isn't pushed up. Let me finalize it now and open if for review.

src/components/date-picker/date-picker.stories.ts
```diff
@@ -121,17 +121,24 @@ export const range = (): string =>
     )}
   </div>`;
 
+const selectRangeString = html`
+  <script>
+    let weekDays = document.querySelector(".week-days");
+    weekDays.querySelector("calcite-date-picker-day:first-child").selected = true;
+    weekDays.querySelector("calcite-date-picker-day:last-child").selected = true;
+  </script>
```
benelan, line 129: Yeah unfortunately this won't work. Any JS needs to go into a [`play` function](https://storybook.js.org/docs/react/writing-stories/play-function) which don't work OOTB with custom elements. I will have some time to look into workarounds after v1. Can you add a `TODO` comment about adding a selectedRange story In lieu of this?

jcfranco: > Can you add a TODO comment about adding a selectedRange story In lieu of this?

Can we create a follow-up issue instead? It'll give it more visibility and we can track it.

src/components/date-picker-day/date-picker-day.scss
jcfranco: @Elijbet Based on ☝️, is the util class still needed?

src/components/date-picker/date-picker.stories.ts
Elijbet: So I should remove this test and create a follow up to to add to add the test later when the play function is implemented?

jcfranco (approved):

Issue #6211, opened 2022-12-30
[Tooltip] Native "Reference element" label appearing upon hover in Addition to Tooltip


SkyeSeitz:
### Actual Behavior

The title field on the reference element, `calcite-button`, is populated with `"Reference element"` 
<img width="174" alt="image" src="https://user-images.githubusercontent.com/29716057/210104504-ef7f5595-55cb-4a21-b67c-cf9789a01400.png">

causing the native label to appear on reference element upon hover, as seen below.

Windows:
<img width="547" alt="image" src="https://user-images.githubusercontent.com/29716057/210104213-b8965bd4-96fc-4b64-99e7-6c04434e65e3.png">

Mac:
<img width="340" alt="image" src="https://user-images.githubusercontent.com/29716057/210104206-61cee621-0946-4177-aec1-6fdd265e0ae3.png">

### Expected Behavior

Do we need this that "title" field populated? 

cc @geospatialem can you confirm if it is required for a11y?

### Reproduction Sample

https://esri.github.io/calcite-components/?path=/story/components-tooltip--simple

### Reproduction Steps

1. Open the Storybook Tooltip sample
2. Hover on the `calcite-button` reference element
3. Wait for `calcite-tooltip` and native label to appear

### Reproduction Version

v1.0.0-beta.99

### Esri team

Calcite (design)

driskull:
It seems like the storybook just needs to be updated to remove the title attribute on the button.

geospatialem:
> It seems like the storybook just needs to be updated to remove the title attribute on the button.

Agree, there also seems to be a missing `span` for AT support. Will have a PR ready for a doc fix shortly.

geospatialem:
Verified on [Storybook](https://esri.github.io/calcite-components/?path=/story/components-tooltip--simple).

PR #6221, opened 2023-01-03
docs(tooltip): assistive technology story support

geospatialem:
**Related Issue:** #6211 

## Summary
- Adds assistive technology (AT) support with the tooltip content reference in a `span` element.
- Removes the button's `title` attribute.

### DIFF
--- a/src/components/tooltip/tooltip.stories.ts
+++ b/src/components/tooltip/tooltip.stories.ts
@@ -7,7 +7,7 @@ import { themesDarkDefault } from "../../../.storybook/utils";
 
 const contentHTML = `Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua`;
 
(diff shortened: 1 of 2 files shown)


macandcheese (approved):
Looks good, this is definitely something a user probably won't expect to have to add for the component to satisfy a11y needs, maybe an issue on the docs site repo to add a note for this?

Is there anyway to automatically set this without a `span` being required as the first child?

geospatialem:
> Looks good, this is definitely something a user probably won't expect to have to add for the component to satisfy a11y needs, maybe an issue on the docs site repo to add a note for this?

Good idea, will put in a PR for incorporation this week.


> Is there anyway to automatically set this without a `span` being required as the first child?

Unfortunately, there isn't - there are still ways assistive technology can receive the context, but its not as straightforward (reading it back on demand). Saw this first hand with a blind user that was unable to access the contents until hitting the down arrow, which isn't expected behavior for screen reader users.

Issue #6152, opened 2022-12-21
Rename `radio-group` and `radio-group-item` components


geospatialem:
### Description

Rename the components to eliminate confusion. As an example [Ant Design](https://ant.design/components/segmented) uses the name `segmented`.

### Proposed Advantages

To eliminate the confusing name hopscotch.

### Which Component

`radio-group`, `radio-group-item`

### Relevant Info

Rename should be considered by designers and developers. 

cc @ashetland @SkyeSeitz @driskull @macandcheese

geospatialem:
Stems from #2512

macandcheese:
I like ‘segmented-something’, I think both ‘segmented-toggle’ and ‘segmented-control’ are pretty familiar names for this element.

ashetland:
I like segmented-control. This is a separate issue, but could we allow icon-only as well?

SkyeSeitz:
I agree with `segmented-control` 👍🏽

geospatialem:
@driskull Sounds like `segmented-control` is what we're working towards. Would you or another dev be able to pick up the change this week?

driskull:
@geospatialem I don't have the bandwidth to take this one. 

Do we need the `-control` part? or would `calcite-segmented` and `calcite-segmented-item` be enough?

@benelan @macandcheese @eriklharper @anveshmekala @Elijbet does one of you have the bandwidth to rename these two components as a breaking change?

macandcheese:
I can pick this up. I think it should probably be `calcite-segmented-control` though, "segmented" is just a descriptor. Or we could go with "chunky buttons" like giffgaff system : https://component.gallery/components/segmented-control/

driskull:
I just saw that `ant` doesn't use the `-control` part.

ashetland:
Omg "chunky buttons"! But seriously, unless there's a good reason to omit `-control`, I'm in the camp of `calcite-segmented-control` and `calcite-segmented-control-item`.

ethanbdev:
Should the event be renamed? https://github.com/Esri/calcite-components/blob/master/src/components/segmented-control/segmented-control.tsx#L253

macandcheese:
@ethanbdev - thanks for the catch, pr incoming...

geospatialem:
Verified on `next.714` with `calcite-segmented-control` and `calcite-segmented-control-item`:

```html
<calcite-segmented-control>
    <calcite-segmented-control-item value="layers" checked>Layers
[...]

PR #6224, opened 2023-01-03
refactor(radio-group, radio-group-item)!: Renames radio group components

macandcheese:
BREAKING CHANGE: Renames radio group components
- `calcite-radio-group` has been renamed to `calcite-segmented-control`
- `calcite-radio-group-item` has been renamed to `calcite-segmented-control-item`


macandcheese:
Addresses #6152 - seemed like this was the agreed upon name, IMO just `calcite-segmented` sounds like it is missing a part of the name, but now would be the time to adjust if needed.

alisonailea (approved):

Issue #6222, opened 2023-01-03
Hovering over the calcite alert when saving a map will not automatically dismiss


rebollin:
### Description

The final calcite alert for saving in Map Viewer will not automatically dismiss after a certain allotment of time. This seems to happen after the user hovers over the previous notifications. Talked with @geospatialem and it is likely a result of this PR https://github.com/Esri/calcite-components/pull/5872. Here is a codepen for this issue: https://codepen.io/kevindoshier/pen/ZEjQPjB

![savefreezing](https://user-images.githubusercontent.com/121894974/210432869-d3ade707-76c0-47f1-a528-ddea1eb55ca6.gif)

### Acceptance Criteria

The alert should automatically dismiss after the allotted time rather than having the user manually close it.

### Which Component

alert component 

### Esri team

ArcGIS Map Viewer

geospatialem:
Could see a use case for potentially having the alert not dismiss automatically when pausing due to hovering (and eventually focusing via #5960). 

Should we consider introducing a boolean property to handle both use cases?

Will evaluate prioritization after this month's 1.0 release for a future milestone and/or assignment.

macandcheese:
I think we should prioritize the bug fix for 1.0, if we can.

> Could see a use case for potentially having the alert not dismiss automatically when pausing due to hovering (and eventually focusing via https://github.com/Esri/calcite-components/issues/5960).

I think this would just be a use case to not have the auto-dismiss property enabled at all by an app. I would expect the dismiss timer to resume once hover / focus leaves, I feel like having an option to "leave the timer halfway done" is kind of strange?

There is precedent for a `no-hover-pause` property: (https://bootstrap-vue.org/docs/components/toast#comp-ref-b-toast-props), but slightly unrelated I suppose.

geospatialem:
Adding to the January milestone, if time permits from the team to tackle.

macandcheese:
Somewhat separately from the component but, looking again at the gif provided, I think invoking the “saving” alert, then removing it without user interaction to replace it with another “save complete” alert is a bit of a weird experience, and might feel broken to a user.

Alternatively, a loading state / loading spinner could be used on the folder action in the action bar instead of the initial “saving” alert - then just provide a single success alert.

macandcheese:
@rebollin @kevindoshier this should be available to test on next release if you could give a look and test it out :)

Would also recommend the UX suggestion above be implemented, check with @asangma for interaction details there.

[...]

PR #6228, opened 2023-01-04
fix(alert): Correctly dismiss after hovering

macandcheese:
**Related Issue:** #6222 

## Summary
This should fix the problem described in the issue - it now handles multiple "hover on and off" events and, at least in local testing, seems to handle large queues that may be present pretty well. The events used were changed to fire less often, and there's an accrued timer of hover time that now is used to determine remaining duration before dismissal.

The watcher for changing the duration now resets it to the default, so unless a user interacts with the alert, it may display for a bit longer. Although, I can't think of a use case where this would be changed while an alert is open, we do use it in demos, and we could refactor this code to better support that if needed.

https://user-images.githubusercontent.com/4733155/210485461-c77e8948-af76-4c61-b965-306fc0a9d973.mov

### DIFF
--- a/src/components/alert/alert.tsx
+++ b/src/components/alert/alert.tsx
@@ -168,7 +168,7 @@ export class Alert implements OpenCloseComponent, LoadableComponent, T9nComponen
       window.clearTimeout(this.autoCloseTimeoutId);
       this.autoCloseTimeoutId = window.setTimeout(
         () => this.closeAlert(),
-        DURATIONS[this.autoCloseDuration] - (Date.now() - this.initialOpenTime)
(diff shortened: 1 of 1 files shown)


macandcheese:
Should also resolve https://github.com/Esri/calcite-components/issues/6132

alisonailea (approved):

macandcheese:
Surprisingly the related change in Chromatic for Alert seems to be unrelated - exists on other PR https://www.chromatic.com/build?appId=6266d45509d7eb004aa254fb&number=663 - if someone could approve :)

PR #6249, opened 2023-01-06
refactor(card)!: rename footer-leading and footer-trailing slots to footer-start and footer-end

benelan:
BREAKING CHANGE: rename footer-leading and footer-trailing slots to footer-start and footer-end

- Removed `footer-leading` and `footer-trailing` slots, use `footer-start` and `footer-end` instead


anveshmekala (approved):
LGTM!

Issue #6241, opened 2023-01-06
calcite-button slotted in a calcite-input can't be disabled by default


jgrayson-apl:
### Actual Behavior

calcite-button slotted in a calcite-input can't be disabled by default

### Expected Behavior

calcite-button slotted in a calcite-input should be allowed to be disabled by default

### Reproduction Sample

https://codepen.io/john-grayson/pen/mdjOwbP

### Reproduction Steps

1. open above codepen
2. see calcite-button slotted inside of calcite-input is not disabled

### Reproduction Version

beta.99

### Esri team

N/A

benelan:
In the following example, would you expect the slotted action to still be `disabled` after the javascript executes.

```html
<calcite-input><calcite-button slot="action" disabled>Action</calcite-button></calcite-input>
<script>
    document.querySelector("calcite-input").disabled = true
    document.querySelector("calcite-input").disabled = false
</script>
```

jgrayson-apl:
> I would expect the button to still be `disabled`, which means we would need to track whether the property was set internally or not.

When the calcite-input disable state is un-disabled I would expect the calcite-button disable state to return to what it was previously.  I would handle this in the calcite-button so that if it's slotted in any of our calcite-input-* components, it can react to the parent disabled state as needed.

benelan:
> I would handle this in the calcite-button so that if it's slotted in any of our calcite-input-* components, it can react to the parent disabled state as needed.

I ended up implementing this in the Input components instead because we can't assume what users will put in the slot. They may slot a `calcite-action` or their own custom button component, both of which should also handle `disable` correctly.

geospatialem:
Verified on `next.728` with an `action` and `button` slotted.

### `button` slotted
![image](https://user-images.githubusercontent.com/5023024/213281943-c80b7f02-11eb-48fe-b124-9b7a52085354.png)

### `action` slotted
![image](https://user-images.githubusercontent.com/5023024/213292516-ca046a9a-8599-406f-a100-e58a2a298e15.png)

PR #6250, opened 2023-01-07
fix(input, input-number, input-text): allow slotted action to be independently disabled

benelan:
**Related Issue:** #6241

## Summary
This PR allows using the `disabled` attribute on the slotted action independently of the Input. Previous the slotted action would mirror the Input's disabled state, regardless of the value of its `disabled` property. 

The slotted action will still be automatically `disabled` if the input is `disabled`. The slotted action will revert to its previous `disabled` state when the Input is un-`disabled`.


src/components/input-number/input-number.e2e.ts
```diff
@@ -1407,6 +1407,59 @@ describe("calcite-input-number", () => {
...
+
+    const input = await page.find("calcite-input-number");
+    const button = await page.find("calcite-button");
+
+    await input.callMethod("setFocus");
+    await typeNumberValue(page, "1");
+    await page.waitForChanges();
+    expect(await input.getProperty("value")).toBe("1");
+    expect(await button.getProperty("disabled")).toBeTruthy();
+    expect(await input.getProperty("disabled")).toBeFalsy();
```
eriklharper, line 1424: nitpick: any reason why `disabled` values won't always be either `true` or `false` in order to use a more explicit assertion like `toBe(true|false)`?

src/components/input-number/input-number.tsx
```diff
@@ -756,9 +758,15 @@ export class InputNumber
       return;
     }
 
-    this.disabled
-      ? slottedActionEl.setAttribute("disabled", "")
-      : slottedActionEl.removeAttribute("disabled");
+    if (this.disabled) {
+      if (slottedActionEl.getAttribute("disabled") == null) {
```
eriklharper, line 762: nitpick: Would `!slottedActionEl.hasAttribute("disabled")` read better here?

src/components/input-number/input-number.e2e.ts
benelan: If whatever the user puts in the slot doesn't reflect the `disabled` attribute it could be `""` or `null`

src/components/input-number/input-number.tsx
benelan: `""` means true for attributes but it's falsy in js

src/components/input-number/input-number.e2e.ts
benelan: Actually I think this one can be changed since I'm explicitly getting the property. I'll double check

eriklharper (approved):

PR #6252, opened 2023-01-08
fix(accordion-item): bumping the scale of icon to M when parent accordion is scale L

Elijbet:
**Related Issue:** #5698

## Summary

Bumping the scale of the icon to M when the parent `accordion-item` is `scale="l"` for a visual distinction between larger and smaller components without affecting the height of the `accordion-item` when icon(s) are added or removed. Added a _testOnly snapshot.


benelan (approved):
👍

PR #6253, opened 2023-01-09
fix(combobox-item): bumping the scale of icon to M when parent combobox is scale L

Elijbet:
**Related Issue:** #5698

Bumping the scale of the icon to M when the parent `combobox-item` is scale="l" for a visual distinction between larger and smaller components without affecting the height of the `combobox-item` when icon(s) are added or removed. Added a _testOnly snapshot.

### DIFF
--- a/src/components/combobox-item/combobox-item.tsx
+++ b/src/components/combobox-item/combobox-item.tsx
@@ -86,6 +86,7 @@ export class ComboboxItem implements ConditionalSlotComponent, InteractiveCompon
 
   isNested: boolean;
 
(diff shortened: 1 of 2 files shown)


benelan (approved):
👍 

In the future these 3 changes can be in one PR since they do the same thing, just for different components. It would save us chromatic snapshots too. The PR title could be something like:

`fix(combobox-item, accordion-item, dropdown-item): increase child icon scale to M when component is L`

alisonailea (approved):

Issue #749, opened 2020-07-22
[color-picker] add support for alpha


jcfranco:
The color picker needs to support alpha channel updates. https://github.com/Esri/calcite-components/issues/64#issuecomment-583678167 has some designs with opacity.

Not sure if the design needs any tweaks, but one suggestion would be to introduce an alpha channel input as well. The color picker from Chrome dev tool's has something similar: 

![cdt-cp-colors](https://user-images.githubusercontent.com/197440/88196177-c02fb700-cbf5-11ea-9807-058e8fc6a327.gif)

bstifle:
👀

jcfranco:
Another thing to consider in the design is the selected color + transparency preview. One idea would be to update `calcite-color-swatch` to support transparency (rough mockup 😄):

<img width="294" alt="Screen Shot 2020-07-29 at 1 31 12 PM" src="https://user-images.githubusercontent.com/197440/88851391-8f292680-d1a1-11ea-869d-fd76f27daa9b.png">

bstifle:
yeah i dig it

macandcheese:
Yeah looks good... Think that's a pretty universal treatment for opacity so should be familiar.

bstifle:
![image](https://user-images.githubusercontent.com/14083530/88860959-c6073880-d1b1-11ea-925a-8de58e6b135b.png)
here's a comp with the transparency checker fill. 

light theme using 4x4 #000000 squares
dark theme using 4x4 #FFFFFF squares

bstifle:
![image](https://user-images.githubusercontent.com/14083530/88863040-9dce0880-d1b6-11ea-9ae0-6175d300d29d.png)
ok this is better. 

#CACACA for light theme
#6A6A6A for dark

smambrose:
Is there any progress on this one? Testers for charts keep running into it and noticing they can't control this.

jcfranco:
@smambrose Not for the time being. I'll move this to backlog until it's ready for development. I estimate this will land after 1.0.0 ships (after April).

tyanthony:
Having the ability to change the Opacity would be a great enhancement for our team as well. An example of a use case for us: 

![image](https://user-images.githubusercontent.com/17069279/124000773-ec28f000-d9a1-11eb-8b10-45c6ae931ec9.png)

julio8a:
@bstifle, is this one ready for implementation?

bstifle:
Here's the design with an opacity slider:

[...]

PR #6265, opened 2023-01-10
test(color-picker): alpha support

alisonailea:
**Related Issue:** #749

## Summary

Cleans up tests for alpha feature on color picker. There is something wrong with Puppeteer causing an unknown error when puppeteer attempts to make programmatic changes to the alpha input box but these changes have been tested manually and are good to go for the 1.0 release. A ticket has been made to address the unstable tests.


src/components/color-picker/color-picker.tsx
```diff
@@ -974,10 +974,7 @@ export class ColorPicker
   }
 
   private renderOpacitySection(): VNode {
-    const { color, previousColor } = this;
-    const sliderOpacity = alphaToOpacity((color ? color : previousColor).alpha()); // slider keeps previous alpha when null
-    const inputOpacity = color ? alphaToOpacity(color.alpha()).toString() : undefined;
```
benelan: Why was this changed to match the sliderOpacity value? I assume there was a reason they were different in the first place, although I have no idea what it might be lol

src/components/color-picker/color-picker.e2e.ts
```diff
@@ -1060,61 +1142,124 @@ describe("calcite-color-picker", () => {
...
+            await rInput.callMethod("setFocus");
+            await selectText(rInput);
+            await page.keyboard.press("Delete");
+            await page.keyboard.press("Enter");
+            await page.waitForChanges();
 
-              await page.keyboard.press("ArrowUp");
-              await page.waitForChanges();
-              expect(await calciteInput.getProperty("value")).toBe(consistentRgbHsvChannelValue);
+            expect(await rInput.getProperty("value")).toBe("");
```
benelan, line 1163: Why did this change from "0" to "", was that a result of changes to the test or the component code?

package.json
```diff
@@ -138,6 +139,7 @@
     "prettier": "2.8.1",
     "puppeteer": "10.0.0",
     "quicktype-core": "6.1.0",
+    "react-textarea-autosize": "8.4.0",
```
benelan, line 142: Adding these dependencies was to fix the storybook type errors right? Can you move these to a separate PR? The way our tsconfig/eslint is set up storybook type errors shouldn't effect builds or lintstaged. I'm pretty sure I asked Franco about adding these awhile ago so I want his eyes on the separate PR

benelan (commented):
Looking good pretty good, thanks for taking this on! Did you clean up some tests that weren't related to alpha in this commit?

https://github.com/Esri/calcite-components/pull/6265/commits/8a610f7dc614b745d186c7979fc756e87ade1469

If so could you move them to a separate PR please

package.json
alisonailea: I don't know what this is. It was already a part of the code base.

src/components/color-picker/color-picker.e2e.ts
alisonailea: changes to the component code. The test was always resulting in ""

src/components/color-picker/color-picker.tsx
alisonailea: the only change here was to make the code more readable. using "this." rather than destructuring at the top of the page and there is no need for "inputOpacity" as this.color is a required prop with a fallback.

alisonailea:
> Looking good pretty good, thanks for taking this on! Did you clean up some tests that weren't related to alpha in this commit?

I kept running into weird timeout errors in the tests which I suspect were caused by the layered nested functions in the multi-layered "describes" these were resolved when I made each test stand alone.

PR #6268, opened 2023-01-10
refactor(accordion)!: remove minimal and default values from appearance property

benelan:
BREAKING CHANGE: removed "minimal" and "default" values for `calcite-accordion`'s `appearance` property.

- Removed `"default"` value for the `appearance` property, use `"solid"` instead.
- Removed `"minimal"` value for the `appearance` property.


anveshmekala (approved):
LGTM!  👍 

Can you also remove CSS class [here](https://github.com/Esri/calcite-components/blob/master/src/components/accordion/accordion.scss#L37)

Issue #5331, opened 2022-09-15
Button neutral and outline uses wrong color for border


asangma:
### Actual Behavior

When Button is `appearance="outline"` and `color="neutral"`, the border color is set to `foreground-3`.

### Expected Behavior

Border color should be `border-1`.

### Reproduction Sample

https://codepen.io/mtb-alan/pen/vYjgrJE

### Reproduction Steps

Inspect the button in the example.

### Reproduction Version

latest

asangma:
cc @ashetland

asangma:
cc @macandcheese

ashetland:
@asangma Sadly, the codepen is blank, but I agree with this one.

asangma:
LOLz. I pasted the wrong url.

macandcheese:
👍 agreed

geospatialem:
Verified in `next.727` with the following html:

```html
  <calcite-button appearance="outline" kind="neutral">Test</calcite-button>
```

![image](https://user-images.githubusercontent.com/5023024/213256574-be5ca2af-471d-4503-aacf-074937cd19dd.png)

PR #6269, opened 2023-01-11
fix(button): neutral and outline button now has correct border color

Elijbet:
**Related Issue:** #5331

## Summary
When `Button` is `appearance="outline"` and `color="neutral"`, the border color is set to `borderColor.color.1`.

### DIFF
--- a/src/components/button/button.stories.ts
+++ b/src/components/button/button.stories.ts
@@ -156,3 +156,6 @@ export const darkModeRTL_TestOnly = (): string => html`
 `;
 
 darkModeRTL_TestOnly.parameters = { modes: modesDarkDefault };
+
+export const outlineNeutralBorderColor_TestOnly = (): string =>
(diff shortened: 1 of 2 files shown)


anveshmekala (approved):
LGTM!

src/components/button/button.stories.ts
```diff
@@ -156,3 +156,6 @@ export const darkModeRTL_TestOnly = (): string => html`
 `;
 
 darkModeRTL_TestOnly.parameters = { modes: modesDarkDefault };
+
+export const outlineNeutralBorderColor_TestOnly = (): string =>
+  html`<calcite-button appearance="outline" kind="neutral">Test</calcite-button>`;
```
anveshmekala, line 161: Nitpick: do we need to add kind here? it will be set to default value if not parsed.

Elijbet: The issue specifies that specific combination of props that gives a wrong border.

PR #6276, opened 2023-01-12
refactor(color-picker)!: remove appearance property

benelan:
BREAKING CHANGE: removed `appearance` property

- Removed `appearance` property, use `--calcite-ui-border-1: transparent` to remove the border instead of the `"minimal"` value


macandcheese (approved):
🖼️

Issue #6285, opened 2023-01-13
Rename `pagination` props to be descriptive


geospatialem:
### Description

The `pagination` components props should be renamed to be more descriptive 
cc @anveshmekala 

### Proposed naming changes
- `num` → `pageSize`
- `total` → `totalItems`
- `start` → `startItem`

### Proposed Advantages

API Consistency

### Which Component

pagination

geospatialem:
This breaking change will be discussed for either the January, or for inclusion in a future release in the next few weeks and the milestone may change.

jcfranco:
Since this component mainly represents pages and not item/value pairs like other components, I'm leaning towards:

* `num` → `pageSize`/`itemsPerPage` (change default to 1; breaking for REST API use cases not setting this)  
* `total` → `pages`/`totalPages` (change default to 1; breaking for REST API use cases not setting this)  
* `start` → `page`/`selectedPage` (no change to default)

WDYT?

Note: I included default changes as well, but these could be handled separately if convenient.

anveshmekala:
I can get behind `num`→ `pageSize` .

For `total` → `pages`/`totalPages`, naming with `pages` as suffix or prefix might be contradictory here because the value is related to total number of items which will be divided with `pageSize`. Either of the suggested names sound more like total no of pages irrespective of the number of items. 

start → `page`/`selectedPage` (no change to default) This seems like a good idea and i checked with other design system which does allow user to specify which page to show up on initial rendering . This might require some change in component code as well. If we are considering only name change then user might enter a value and expect the valued page number is shown where as actual behavior shows up page in which the item index is located.

benelan:
Based on Franco's comment it sounds like the suggestion includes changing the prop's value from total items to total pages too?

 I think logically that makes sense for a component about pages, but it may make it harder to use for everyone, not just AGOL. Currently you can use an array's length for the `total` property, which makes dynamically changing the items easier. [Here is an example](https://codepen.io/benelan/pen/bGvdEgR).

Either way, I don't think we should consider changing the `total` prop from items to pages for v1. It's a lot of surface area that would need refactoring for all of the teams using pagination with the REST API. I think it may be worth messaging stakeholders to see how big of an impact it will be for them if that's what we are thinking about.

If we aren't discussing changing from total items to total pages then I agree with Anvesh's comment about the `totalPages` prop name. Maybe something like `totalItems` instead?

jcfranco:
[...]

PR #6286, opened 2023-01-13
refactor(pagination)!: rename properties.

anveshmekala:
**Related Issue:** #6285 

BREAKING CHANGE: Rename properties.

- Removed the property `num`, use `pageSize` instead.
- Removed the property `total`, use `totalItems` instead.
- Removed the property `start`, use `startItem` instead.

### DIFF
--- a/src/05-custom-theme.stories.mdx
+++ b/src/05-custom-theme.stories.mdx
@@ -162,7 +162,7 @@ export const Template = () =>
           <calcite-chip appearance="transparent" kind="inverse">Inverse</calcite-chip>
           <calcite-chip appearance="outline-fill" kind="brand">Brand</calcite-chip>
         </div>
-        <calcite-pagination total="1200" num="100" start="1"></calcite-pagination>
+        <calcite-pagination total-items="1200" page-size="100" start-item-index="1"></calcite-pagination>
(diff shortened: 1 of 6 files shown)


ashetland:
@SkyeSeitz we'll need to update prop names in Figma.

SkyeSeitz:
> @SkyeSeitz we'll need to update prop names in Figma.

Thanks for the heads up on this, Aaron. I confirmed if any changes are needed in Figma. It looks like none of these would apply to the kit as Figma does not set `start` or `total` properties and every number is the same internal component with a shared text property that needs to stay generically called `number` 👍🏽

ashetland:
> > @SkyeSeitz we'll need to update prop names in Figma.
> 
> Thanks for the heads up on this, Aaron. I confirmed if any changes are needed in Figma. It looks like none of these would apply to the kit as Figma does not set `start` or `total` properties and every number is the same internal component with a shared text property that needs to stay generically called `number` 👍🏽

Excellent news. Thanks @SkyeSeitz!

jcfranco (approved):
LGTM!

Can you make sure to use the prop names, instead the attributes, in the breaking change notes right before merging?

Issue #5816, opened 2022-11-28
Date picker range `value` undefined


geospatialem:
### Actual Behavior

When using the `range` prop on the date-picker component, there are a few unexpected behaviors, one programmatic and one visual:

1. Programmatic: 
If the `range` prop is present and a predefined range is not selected, even upon selecting a range, the `value` prop is not available on the range. Instead, the values are added to the deprecated `start` and `end` props.

3. Visual: 
The component doesn't seem to hold the change either, perhaps a regression from #5459?

![image](https://user-images.githubusercontent.com/5023024/204373818-ef107613-f1a7-4b3b-b8c3-8de5133627d9.png)

![image](https://user-images.githubusercontent.com/5023024/204373844-6d5f2942-e9cb-4c07-8299-5538d1ef6622.png)

![image](https://user-images.githubusercontent.com/5023024/204373876-54fac9c7-681f-4d85-ab19-434e73548bae.png)

### Expected Behavior

Once values have been selected for the component's value, they should be accessible via the `value` prop since the `start` and `end` props are deprecated.

### Reproduction Sample

https://codepen.io/geospatialem/pen/mdKjwxx

### Reproduction Steps

1. Open the [sample](https://codepen.io/geospatialem/pen/mdKjwxx)
2. Open the console
3. Observe the first `date-picker` with a predefined range and its `value` depicted in the console
4. Select a range of dates in the second `date-picker`
5. Observe the second `date-picker`'s `value` is depicted as "`undefined`" whereas the `start` and `end` properties contain values.

### Reproduction Version

beta.98

### Esri team

Calcite (dev)

macandcheese:
Noted as high priority https://devtopia.esri.com/WebGIS/arcgis-portal-app/issues/40558#issuecomment-3883901 cc @geospatialem

eriklharper:
I verified that this issue is not a regression as a result of #5459, this appears to have been a prior issue in that `this.value` isn't properly set in `range` mode, most likely because `input-date-picker` never used `value` directly, instead relying on `valueAsDate`, `startAsDate` and `endAsDate`.  We'll correct this behavior, but in the meantime the best workaround is to utilize `valueAsDate` instead.

PR #6289, opened 2023-01-14
fix(date-picker): range value property updates correctly

eriklharper:
**Related Issue:** #5816 

## Summary

`calcite-date-picker` correctly updates its `value` property when start and end dates are selected.


alisonailea (approved):

benelan (approved):
Code LGTM!

Nitpick: 
Can you add spec test(s) for `dateToISO`? Looks like we have some for `dateFromISO` already so we can copy that format

Issue #6270, opened 2023-01-11
Trying to type a negative number into `calcite-input` after programmatically clearing it causes the old number to come back


ethanbdev:
### Actual Behavior

If I have a number in `calcite-input`, then clear the input by setting the `value` to `""`, trying to input a negative number will bring back the old, cleared number on the `-` keypress.
![87AF2C15-6FEF-4665-8336-F1A3C0007DD4](https://user-images.githubusercontent.com/52869490/211689205-5963c641-633f-494e-bc4d-d4f955686716.GIF)

### Expected Behavior

I should be able to type a negative number as usual

### Reproduction Sample

https://codepen.io/eborgen/pen/wvxJZMP?editors=1010

### Reproduction Steps

1. Type any number into the input
2. Use the button to clear the input
3. Now try type a negative number
4. Observe the number comes back

### Reproduction Version

beta.99

### Impact

Prohibiting a certain workflow by changing a user input.

The use case is the user types a number, then commits it. We then clear the input after the commit happens. Now, trying to do a negative number next is a frustrating experience

### Esri team

ArcGIS Map Viewer

ethanbdev:
cc: @benelan

benelan:
Thanks for reporting, I'll try to get this fixed before v1. `calcite-input-number` works correctly as a temporary workaround if this is causing issues with your testing. Should be relatively easy to fix otherwise though.

benelan:
It's the same as `calcite-input type="number"` at the moment, just without the craziness of handling multiple types lol. We planned to split off some of the input components for v1, including new `calcite-textarea` (#5644) and `calcite-input-time` (#5639), but we aren't quite done yet. 

We will message out the new Input setup when it's all done, but you won't have to change anything until May at the earliest. So for now I'd continue using `calcite-input` if this bug isn't killing your CI. I'll have a PR in a sec

ethanbdev:
> Yeah it ended up not being what I expected and I got pulled away into a high priority Tree issue for the Arcade team. I should still have this done for v1 though.

Np, thanks for the update

geospatialem:
Verified on `next.736`

![verify-input](https://user-images.githubusercontent.com/5023024/214474717-3c5d20e9-2686-4001-b7ad-1e23fa30a912.gif)

ethanbdev:
Thanks for getting this in! 🍻

PR #6306, opened 2023-01-18
fix(input, input-number): correctly handle '-' and '.' values when sanitizing number

benelan:
**Related Issue:** #6270

## Summary
Fixes issues in input and input-number involving how to handle `-` and `.` values. We don't want to set it as the `value` property because they aren't numbers, but we also don't want to sanitize them because people should be allowed to type negative and decimal numbers. Added a couple code comments for clarity


src/components/input-number/input-number.tsx
```diff
@@ -806,6 +806,7 @@ export class InputNumber
     };
 
     const sanitizedValue = sanitizeNumberString(
+        // no need to delocalize a string that ia already in latn numerals
```
eriklharper, line 809: typo: "is" instead of "ia"

eriklharper (approved):
Looks good, tested locally.  Just noticed a typo in the comment, other than that, ready to merge :)

src/components/input-number/input-number.tsx
benelan: Nice catch on the typo. Its breaking some tests which I'm looking into before merging

src/components/input-number/input-number.e2e.ts
```diff
@@ -914,13 +914,16 @@ describe("calcite-input-number", () => {
...
 
-  it("allows clearing value with an empty string", async () => {
+  it("allows negative numbers after clearing value with an empty string", async () => {
     const page = await newE2EPage();
     await page.setContent(html`<calcite-input-number value="1"></calcite-input-number>`);
     const input = await page.find("calcite-input-number");
     input.setProperty("value", "");
     await page.waitForChanges();
     expect(await input.getProperty("value")).toBe("");
+    typeNumberValue(page, "-123");
```
jcfranco, line 924: This needs to be awaited on, right?

src/components/input/input.e2e.ts
```diff
@@ -1070,13 +1070,16 @@ describe("calcite-input", () => {
...
 
-    it("allows clearing value with an empty string", async () => {
+    it("allows negative numbers after clearing value with an empty string", async () => {
       const page = await newE2EPage();
       await page.setContent(html`<calcite-input type="number" value="1"></calcite-input>`);
       const input = await page.find("calcite-input");
       input.setProperty("value", "");
       await page.waitForChanges();
       expect(await input.getProperty("value")).toBe("");
+      typeNumberValue(page, "-123");
```
jcfranco, line 1080: Missing await here too.

PR #6308, opened 2023-01-19
test(date): add test for dateToISO utility function

eriklharper:
**Related Issue:** https://github.com/Esri/calcite-components/pull/6289#pullrequestreview-1252734747

## Summary

Adds a spec test for `dateToISO`.  Also updates the function to return a proper empty string value when any non-`Date` value is passed in.

### DIFF
--- a/src/utils/date.spec.ts
+++ b/src/utils/date.spec.ts
@@ -1,5 +1,15 @@
 import { DateLocaleData } from "../components/date-picker/utils";
-import { dateFromISO, dateFromRange, getOrder, inRange, nextMonth, parseDateString, prevMonth, sameDate } from "./date";
+import {
+  dateFromISO,
+  dateFromRange,
+  dateToISO,
+  getOrder,
+  inRange,
+  nextMonth,
+  parseDateString,
+  prevMonth,
+  sameDate
+} from "./date";
 
 import arabic from "../components/date-picker/assets/date-picker/nls/ar.json";
 import english from "../components/date-picker/assets/date-picker/nls/en.json";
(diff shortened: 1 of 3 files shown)


src/utils/date.ts
```diff
@@ -126,10 +126,7 @@ export function datePartsFromLocalizedString(
  *
  * @param date
  */
-export function dateToISO(date?: Date | string): string {
-  if (typeof date === "string") {
-    return date;
-  }
+export function dateToISO(date?: Date): string {
```
eriklharper, line 129: This _shouldn't_ break anything since all usages are sending in `Date` objects or empty strings, and nothing should be relying on the improper value coming back.

src/utils/date.spec.ts
```diff
@@ -63,6 +73,23 @@ describe("dateFromISO", () => {
...
+describe("dateToISO", () => {
+  it("returns empty string from bad input", () => {
+    expect(dateToISO("" as any)).toEqual("");
+    expect(dateToISO("asdflkjasdhoui" as any)).toEqual("");
+  });
+  it("correctly returns string in simplified ISO format (YYYY-MM-DD)", () => {
+    const date = new Date(2011, 10, 29);
+    const expectedValue = "2011-11-29";
+    expect(dateToISO(date)).toEqual(expectedValue);
+  });
```
benelan, line 85: I think this test is redundant since the one below tests the same base case plus the zero padding. We could probably remove it, what do you think?

src/utils/date.ts
benelan: Yeah typescript would complain if it was recieving a string since you changed the signature. So we should be good

benelan (approved):
LGTM, thanks for adding it! I added a comment about a potentially redundant test, but I'll leave it up to you

Issue #6284, opened 2023-01-13
Expanded Tree Item doesn't render properly under TSX 


ffaubry:
### Actual Behavior

![image](https://user-images.githubusercontent.com/3506166/212389816-b3d6061b-7bb3-4cfe-b6fb-7896b7a81465.png)

### Expected Behavior

![image](https://user-images.githubusercontent.com/3506166/212389881-04884939-e95f-4dfa-b8e5-d137c98483e2.png)

### Reproduction Sample

Use Arcade Editor on qawebgis

### Reproduction Steps

1. https://qawebgis.esri.com/arcade-editor/main/
2. Click Run in upper left corner
3. Result tab should open
4. A tree is supposed to be expanded and rendered 
5. Inspect tree in debugger and notice that everything is in place
6. search for children-container in the calcite-tree-item shadow dom
7. change the style by adding `transform: none`

### Reproduction Version

next.723

### Regression?

Was working before

### Impact

Confuses users when using Arcade Editor.

### Esri team

ArcGIS Map Viewer

CodeBarker:
This ones a critical issue for the Arcade Editor team and it will impact multiple apps such as Field Maps and potentially Dashboard

benelan:
Removing the `incomplete issue report` label since I was able to reproduce. It appears to only occur in the output dist builds and not the CDN, e.g. [this repros](https://codesandbox.io/s/cc-expanded-not-rendering-7o2fk8), but [this doesn't](https://codepen.io/benesri/pen/MWBOMzY). I'm working on a fix now, but here is a temporary workaround in case this is impacting anyone's CI/testing:

```css
:root {
  --calcite-duration-factor: 0;
}
```

geospatialem:
Verified in `next.736` with Ben's [Code sandbox](https://codesandbox.io/s/cc-expanded-not-rendering-7o2fk8).

geospatialem:
@omarkawach Good catch - can also repo on my end. We do plan to have a patch for 1.0.3 in the next week or so. @jcfranco this could be a good candidate to tackle, WDYT?

omarkawach:
Thanks @geospatialem 

I was also able to reproduce this on the Calcite Design System site

![image](https://user-images.githubusercontent.com/49499098/214966324-c2fd1a78-bbc1-4103-aaa9-cd8840f14058.png)

geospatialem:
Opened a new issue to track the vertical line on `tree` introduced with the above fix.

PR #6320, opened 2023-01-20
fix(tree): expanded item renders correctly on initial load

benelan:
**Related Issue:** #6284

## Summary

This issue only occurs on the `custom-elements` and `dist` output builds, not the CDN. It is a timing issue where the item is expanded on initial load before the transition event listeners that handle styling are added. The fix uses a flag to make sure the transition start/end event listeners are added on initial load before the tree item expands.

### DIFF
--- a/src/components/tree-item/tree-item.tsx
+++ b/src/components/tree-item/tree-item.tsx
@@ -71,7 +71,6 @@ export class TreeItem
   expandedHandler(newValue: boolean): void {
     this.updateParentIsExpanded(this.el, newValue);
     onToggleOpenCloseComponent(this, true);
-    this.treeItemNeverExpanded = false;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/components/tree-item/tree-item.tsx
```diff
@@ -425,6 +430,14 @@ export class TreeItem
   //
   //--------------------------------------------------------------------------
 
+  /**
+   * Used to make sure initially expanded tree-item can properly
+   * transition and emit events from closed state when rendered.
+   *
+   * @private
+   */
+  @State() transitionListenersAddedOnInitialLoad = false;
```
jcfranco, line 439: WDYT about:

1. renaming this to `updateAfterInitialRender`
1. removing the line in `componentWillLoad` that sets it
1. adding a `ref` to host or the expanding container that sets the flag if `expanded === true`?

I think we should at least do 1 even if it's with another name. Reason being that there's no direct connection between the transition listeners being added and how/when the flag is set.

jcfranco (approved):
I had some comments regarding naming and a possible alternative to simplify. Other than that, this LGTM!

Issue #5284, opened 2022-09-08
calcite-input: required and type attributes should behave like HTML input


gpbmike:
### Actual Behavior

- The `required` attribute does not prevent form submission when the value is empty.
- The `type` attribute does not prevent form submission when value is invalid (ex: email)

### Expected Behavior

- When `required` attribute is present and value is empty (""), form submission should be blocked.
- When `type` attribute is set to "email" and value is not a valid email address, form submission should be blocked.

### Reproduction Sample

https://codesandbox.io/s/calcite-input-required-type-email-nxrvh7?file=/src/App.js

### Reproduction Steps

1. Submit form without entering any values.
2. Form submission is blocked by the HTML input.
3. Enter valid email address into HTML input.
4. Submit form.
5. Form submission is NOT blocked by invalid calcite-input value.

### Reproduction Version

beta.93

### Relevant Info

I'm in the process of upgrading from an old version of calcite-components (43), where this behavior previously worked.

### Regression?

beta.43

benelan:
FWIW its not just Input, the `required` property doesn't work as expected for other form components either, e.g. https://codesandbox.io/s/cc-form-validation-jqqrge

I'm pretty sure it's due to #2777. We are planning to do a big form validation epic after v1 where this will be addressed.

As a workaround, users can do their own form validation, here is a small example: https://codepen.io/benesri/pen/xxzaNKW

jcfranco:
Just checked and there is a bug preventing some form components from providing type info to the form, but the rest should be working as expected. One thing I noticed in the repro cases was the lack of `name`. The `name` attribute is required for our components to participate in form submission. I don't think we doc this at the moment, which can lead to confusion. cc @geospatialem

geospatialem:
Verified with the [Codesandbox sample](https://codesandbox.io/s/calcite-input-required-type-email-nxrvh7?file=/src/App.js) with React `1.0.1` and adding an accompanying `name` attr.

PR #6322, opened 2023-01-21
fix(checkbox, input, radio-button, switch): improve form validation by adding type to internal inputs

jcfranco:
**Related Issue:** #5284 

## Summary

This updates the form component pattern to sync hidden inputs with their corresponding value type.


src/components/input/input.tsx
```diff
@@ -833,22 +833,21 @@ export class Input
...
   syncHiddenFormInput(input: HTMLInputElement): void {
-    if (this.type === "number") {
-      input.type = "number";
+    const { type } = this;
+
+    input.type = type;
+
+    if (type === "number") {
       input.min = this.min?.toString(10) ?? "";
       input.max = this.max?.toString(10) ?? "";
```
benelan, line 842: Nitpick: I know you didn't touch this, but `10` is the default radix and I think putting it there is more confusing than omitting it.

src/tests/commonTests.ts
```diff
@@ -531,14 +536,38 @@ export async function formAssociated(componentTagOrHtml: TagOrHTML, options: For
...
+        );
+
+        return hiddenFormInput.type;
+      },
+      name,
+      hiddenFormInputSlotName
+    );
+
+    if (checkable) {
+      expect(hiddenFormInputType).toMatch(/radio|checkbox/);
```
benelan, line 565: Would `expect(["radio", "checkbox"]).toContain(hiddenFormInputType)` work here so we don't need to use regex?

src/utils/form.tsx
```diff
@@ -324,15 +324,12 @@ function defaultSyncHiddenFormInput(
...
+
     // keep in sync to prevent losing reset value
     input.defaultChecked = component.defaultChecked;
     // heuristic to support default/on mode from https://html.spec.whatwg.org/multipage/input.html#dom-input-value-default-on
     input.value = component.checked ? value || "on" : "";
-
-    // we disable the component when not checked to avoid having its value submitted
-    if (!disabled && !component.checked) {
-      input.disabled = true;
-    }
```
benelan, line 335: For my own understanding, why don't we need this anymore?

benelan (approved):
Couple questions/comments but LGTM!

src/components/input/input.tsx
jcfranco: We can remove it [since it's the default](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toString#radix). Maybe it was conflated with [`parseInt` radix recommendation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#radix): 

> If 0 or not provided, the radix will be inferred based on string's value. Be careful — this does NOT always default to 10!

src/utils/form.tsx
jcfranco: Good question. We don't need it anymore because the internal input has the matching type that will prevent submitting the value when unchecked: https://codepen.io/jcfranco/pen/XWBwmoy?editors=1001

src/tests/commonTests.ts
jcfranco: I would definitely change it if the regex was slightly more complex. For now, both seem pretty straightforward. We can revisit this later.

PR #6333, opened 2023-01-24
ci: add npm script to remove js files compiled by tsc

benelan:
## Summary

Adds an NPM script that removes all of the javascript files compiled by tsc when doing the type check before releases.


jcfranco (approved):
💥🤖⚡📁💥

PR #6335, opened 2023-01-25
fix(select): bumping scale of chevron icon to M when host is scale l

Elijbet:
**Related Issue:** #5698

## Summary
Bumping the scale of the chevron to M when host is scale="l" for a visual distinction between larger and smaller components without affecting the height of the component when icon(s) are added or removed. Added a _testOnly snapshot.


jcfranco (approved):
💥

PR #6336, opened 2023-01-25
ci: prepare release scripts for v1

benelan:
## Summary

Modifies the release scripts to actually change major/minor/patch version instead of using prerelease tags for both `latest` and `next`. Previously we were telling `standard-version` which to version use, now we are getting the release version from them.


jcfranco (approved):
✨:shipit:✨

Issue #5803, opened 2022-11-23
Padding on new Popover scales is incorrect


ashetland:
### Actual Behavior

Padding as built:
<img width="716" alt="CleanShot 2022-11-22 at 16 14 53@2x" src="https://user-images.githubusercontent.com/108549080/203446505-65c4cf43-0e3f-4543-a456-6e932d06989b.png">
<img width="791" alt="CleanShot 2022-11-22 at 16 14 58@2x" src="https://user-images.githubusercontent.com/108549080/203446509-36b249ed-c88f-4c5c-aae4-c876ee93e386.png">
<img width="853" alt="CleanShot 2022-11-22 at 16 15 03@2x" src="https://user-images.githubusercontent.com/108549080/203446514-f5fb9448-b116-42c5-b7c1-45417b5ecf8a.png">

### Expected Behavior

Correct padding to match other Calcite norms:
<img width="693" alt="CleanShot 2022-11-22 at 16 18 00@2x" src="https://user-images.githubusercontent.com/108549080/203446715-3d3bfeeb-b006-468c-bfcb-9951c45a8e67.png">
<img width="710" alt="CleanShot 2022-11-22 at 16 18 06@2x" src="https://user-images.githubusercontent.com/108549080/203446720-110b8f14-7e04-451c-b857-f20061fd10ae.png">
<img width="686" alt="CleanShot 2022-11-22 at 16 18 10@2x" src="https://user-images.githubusercontent.com/108549080/203446725-7753a4af-0493-4b90-87b7-a6edb6ccad90.png">

### Reproduction Sample
- https://codepen.io/geospatialem/pen/qBKoPVR

- See Popover demo on DevVM

### Reproduction Steps

-

### Reproduction Version

-

### Esri team

Calcite (design)

ashetland:
@Elijbet One of the "correct behavior" screenshots above is wrong. 🤦‍♂️ Please refer to this instead.
<img width="1252" alt="CleanShot 2022-11-30 at 16 02 39@2x" src="https://user-images.githubusercontent.com/108549080/204934181-f4331ee2-386f-44e4-ba7c-7157b9e291e8.png">

geospatialem:
Verified on `1.0.4-next.2`

<img width="718" alt="Screenshot 2023-01-28 at 10 14 23 PM" src="https://user-images.githubusercontent.com/5023024/215304603-dbecd98d-0906-47a6-9a5c-fbe91c29b2bb.png">

PR #6341, opened 2023-01-25
fix(popover): fix heading padding for m and l scales

jcfranco:
**Related Issue:** #5803 

## Summary

Fixes the x and y padding applied to the popover's heading (they were inverted).

**Note**: I included simple test stories for s, m and l scales. I can tweak these as needed based on our Chromatic setup.


ashetland (approved):

Elijbet (approved):
✨💫

Issue #6310, opened 2023-01-19
[TabNav] TypeError when used inside Panel


benelan:
### Actual Behavior

There is a type error when using Tabs inside of Panel

> p-4c5a1247.entry.js:6 Uncaught TypeError: Cannot read properties of undefined (reading 'style')
    at ResizeObserver.<anonymous>

### Expected Behavior

There should not be a TypeError when using Tabs inside of Panel

### Reproduction Sample

https://codepen.io/fcbasson/pen/KKByZzV

### Reproduction Steps

1. Open the repro sample and check the console

### Reproduction Version

next.728

### Relevant Info

This was reported on Esri Community:
https://community.esri.com/t5/calcite-design-system-questions/calcite-tabs-creating-resizeobserver-type-error/m-p/1249383#M290

Here is the line causing issues:
https://github.com/esri/calcite-components/blob/f146a0991799c1d44261bb808fc05c54515f78d2/src/components/tab-nav/tab-nav.tsx#L307

### Esri team

Calcite (dev)

AdelheidF:
also reproducible with next.716

geospatialem:
Verified in `1.0.4-next.4`

PR #6342, opened 2023-01-25
fix(tabs): fix error when tabs is resized before initial render

jcfranco:
**Related Issue:** #6310

## Summary

This prevents resize logic from running if the indicator element has not been rendered and stored internally.


eriklharper (approved):

SaschaBrunnerCH:
Could it be that the bot did something wrong with the adding to "2023 February Priorities - No breaking changes"?
The issue itself is in "2023 January Priorities Patch - Bugs on"
https://github.com/Esri/calcite-components/milestone/115?closed=1
@jcfranco

Issue #6292, opened 2023-01-16
Tooltip ESC key press propagation


Anastasiia-Boleiko:
### Summary

Calcite tooltip is dismissable that complies with Success Criterion [1.4.13 Content on Hover or Focus](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html ) (Level AA) 

But when pressing the 'Esc' key in an opened calcite-tooltip, that is located in a calcite-modal, the whole modal is closed.
The expected behavior is to only close the tooltip and close the Dialog on the next 'ESC' key press.

Generally speaking, it would be nice to close a context element at the moment (with the Esc key).

### Actual Behavior

After pressing 'Esc' key to close a tooltip, the whole dialog gets closed.

### Expected Behavior

After pressing 'Esc' key - close the tooltip, and close the Dialog on the next 'ESC' key press.

### Reproduction Sample

https://codepen.io/anastasiia-zrh/pen/bGjrZYX

### Reproduction Steps

1. Click 'Open Modal' button
2. Press 'Esc'

One more example, where it would be nice to close a contextual element (with the Esc key), following visual order:
https://codepen.io/anastasiia-zrh/pen/MWBvZdB 

### Reproduction Version

1.0.0-beta.97

### Relevant Info

A similar issue was found in our product by LevelAccess (so we tried calcite components first and found a similar issue):
### **Ensure that content that appears on hover or focus may be dismissed by the user**
**Violation:**
Ensure that content that appears on hover or focus may be dismissed by the user.
**Description:** 
Examples include:
Tooltips
In this case, pressing the Esc key does close the tooltip, but it also closes the entire dialog.
**User Impact**
Users may press the Esc key to close the tooltip and may unwittingly and unexpectedly close the entire dialog.

### Esri team

ArcGIS Urban

geospatialem:
Verified with https://codepen.io/anastasiia-zrh/pen/bGjrZYX in `1.0.5`.

![verify-tooltip-esc-dismiss](https://user-images.githubusercontent.com/5023024/217853411-f11dec3d-89a8-4edf-96ac-8513b33ec0f0.gif)

Does not include the https://codepen.io/anastasiia-zrh/pen/MWBvZdB sample. @jcfranco, should we open a new issue to close a contextual element (with the Esc key), following visual order?

Anastasiia-Boleiko:
Thank you for the update!

Recently, I've checked the link:

[...]

PR #6343, opened 2023-01-25
fix(tooltip): prevent closing of Esc-key-closing parent components when dismissing a tooltip with Esc

jcfranco:
**Related Issue:** #6292

## Summary

This updates tooltip manager to cancel the Esc key press if handled to prevent closing parent components that also close with Esc.

Additionally, extracts a utility to get the tooltip's effective reference element.


benelan (approved):
👍

PR #6351, opened 2023-01-26
ci: fix release script sanity check error

benelan:
**Related Issue:** #6336

## Summary
The git sanity check in the prepReleaseCommit for making sure the releaser doesn't have a dirty git state is failing:
https://github.com/Esri/calcite-components/actions/runs/4003255063/jobs/6871188483#step:5:9033

This is because the `prepReleaseCommit` script execution was moved to the "precommit" `standard-version` lifecycle hook, which triggers after the version bumps in `package.json`, causing the dirty git state.

I moved that sanity check to the beginning of the release process since it doesn't need to be so close to the commit like it does when checking to make sure we are up date with origin.

The tag also didn't end up getting pushed, and It looks like the tag being created isn't annotated, which is required for `--follow-tags`. So I explicitly push the most recent tag instead.

Edit:

I had to comment out that sanity check last night, which is how I missed it. I had to change the version in the `package.json` when releasing last night because `standard-version` was trying to bump to `1.0.0` instead of `1.0.3`


jcfranco (approved):
Nice! Bonus points if we can extract the git working tree checks (for readability).

PR #6355, opened 2023-01-26
ci(linting): improve results readability and add markdownlint

benelan:
## Summary
- Redirects stdout to /dev/null in the `npm run lint:*` scripts. When linting it would previously print a line for every file being linted, which buried errors/warnings when they popped up in stderr. Now we can actually see the warnings/errors.
- Adds `markdownlint` for our documentation. We were having some weirdness with the changelog formatting which this will resolve, among other things.
- Skipped the following [markdownlint rules](https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md), which we don't seem to care about based on all of the errors in our code lol
    - MD013 - Line length (everywhere)
    - MD024 - Multiple headers with the same content (e.g in our changelog)
    - MD033 - Inline HTML
    - MD041 - First line in file should be a top level header (e.g in our main readme)


jcfranco (approved):
Sweet improvement! ✨💻🧹✨

Issue #5143, opened 2022-08-12
Accordion's selection mode can become stale


nwhittaker:
### Actual Behavior

Given a `<calcite-accordion selection-mode="single">`, then dynamically changing its `selection-mode` to `multi`, the user is unable to expand multiple sections.

### Expected Behavior

Given a `<calcite-accordion selection-mode="single">`, then dynamically changing its `selection-mode` to `multi`, the user is able to expand multiple sections.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/JjLaVNB

### Reproduction Steps

1. Visit the sample and note the accordion behaves in accordance with the `"single"` selection mode.
2. Click the **Multi selection** button and use dev tools to inspect the accordion and note it now has `"multi"` selection mode.
3. Attempt to expand more than one accordion section and note that the previously expanded section auto-collapses.

### Reproduction Version

beta.91

### Relevant Info

Toggling the selection-mode is convenient for re-purposing the accordion for different states. Specifically, Field Maps could use an accordion to show different groups of layers one-at-a-time. But, when the user enters some text to filter by, we want to expand all the sections and show the matching layers.

geospatialem:
Verified in `1.0.4-next.3` with the following [Codepen sample](https://codepen.io/geospatialem/pen/jOpvMrw).

![image](https://user-images.githubusercontent.com/5023024/215549174-30ec505a-2114-48f1-9dbf-ee989361d634.png)

PR #6356, opened 2023-01-26
fix(accordion): supports selection mode updates

anveshmekala:
**Related Issue:** #5143

### DIFF
--- a/src/components/accordion-item/accordion-item.tsx
+++ b/src/components/accordion-item/accordion-item.tsx
@@ -90,7 +90,6 @@ export class AccordionItem implements ConditionalSlotComponent {
 
   connectedCallback(): void {
     this.parent = this.el.parentElement as HTMLCalciteAccordionElement;
-    this.selectionMode = getElementProp(this.el, "selection-mode", "multiple");
     this.iconType = getElementProp(this.el, "icon-type", "chevron");
     this.iconPosition = getElementProp(this.el, "icon-position", this.iconPosition);
     this.scale = getElementProp(this.el, "scale", this.scale);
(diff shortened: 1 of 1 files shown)


src/components/accordion-item/accordion-item.tsx
```diff
@@ -272,6 +271,7 @@ export class AccordionItem implements ConditionalSlotComponent {
   //--------------------------------------------------------------------------
 
   private determineActiveItem(): void {
+    this.selectionMode = getElementProp(this.el, "selection-mode", "multiple");
```
macandcheese, line 274: Instead of checking here, could this be set by the parent Accordion when selectionMode is updated `@Watch("selectionMode") ...` ?

anveshmekala: i don't think that can be used here considering we are determining the selection of items inside the accordion-item.

src/components/accordion/accordion.e2e.ts
```diff
@@ -195,4 +195,33 @@ describe("calcite-accordion", () => {
...
+    const page = await newE2EPage();
+    await page.setContent(`
+    <calcite-accordion selection-mode="single">
+    ${accordionContent}
+    </calcite-accordion>`);
+    const element = await page.find("calcite-accordion");
+    expect(element).toEqualAttribute("selection-mode", "single");
+    element.setAttribute("selection-mode", "multiple");
+    await page.waitForChanges();
+    const item1 = await element.find("calcite-accordion-item[id='1']");
```
jcfranco, line 209: Suggestion: `const [item1, item2, item3] = await element.findAll("calcite-accordion-item");`. Totally up to you. 😄

jcfranco (approved):
Nice! 🎉

jcfranco:
@Elijbet This one's another candidate for https://github.com/Esri/calcite-components/issues/6038.

src/components/accordion/accordion.e2e.ts
anveshmekala: Clean 🧹 . applied this to other tests too.

PR #6360, opened 2023-01-26
fix(input-date-picker): input renders numbers in the specified numbering system

eriklharper:
**Related Issue:** #6198 

## Summary

This PR restores the normal behavior of displaying the proper characters according to the specified numbering system in the input.

### DIFF
--- a/src/components/input-date-picker/input-date-picker.e2e.ts
+++ b/src/components/input-date-picker/input-date-picker.e2e.ts
@@ -228,54 +228,92 @@ describe("calcite-input-date-picker", () => {
     expect(await calendar.isVisible()).toBe(true);
   });
 
-  it("syncs lang changes to internal date-picker and input", async () => {
-    const lang = "it-CH";
-    const newLang = "nl";
-    const year = "2020";
-    const month = "4";
-    const day = "19";
+  describe("localization", () => {
+    it("renders arabic numerals while typing in the input when numbering-system is set to arab", async () => {
+      const page = await newE2EPage();
+      await page.setContent(
+        `<calcite-input-date-picker lang="ar" numbering-system="arab"></calcite-input-date-picker>`
+      );
+      const getInputValue = async () =>
+        await page.evaluate(
+          () =>
+            document
+              .querySelector("calcite-input-date-picker")
+              .shadowRoot.querySelector("calcite-input")
+              .shadowRoot.querySelector("input").value
+        );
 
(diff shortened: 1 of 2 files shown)


src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -939,4 +944,14 @@ export class InputDatePicker
           )
           .join("")
       : "";
+
+  private parseNumerals = (value: string): string =>
```
jcfranco, line 948: Sidebar: @benelan Is there a util from input or input-number that could be used here? If so, we should refactor in a separate PR.

src/components/input-date-picker/input-date-picker.e2e.ts
```diff
@@ -228,54 +228,95 @@ describe("calcite-input-date-picker", () => {
...
-      await page.evaluate(
-        async (MONTH_HEADER_CSS) =>
-          document
-            .querySelector("calcite-input-date-picker")
-            .shadowRoot.querySelector("calcite-date-picker")
-            .shadowRoot.querySelector("calcite-date-picker-month-header")
-            .shadowRoot.querySelector(`.${MONTH_HEADER_CSS.month}`).textContent,
-        MONTH_HEADER_CSS
-      );
+      // NOTE: This asserted value was copied from the received value in a test failure caused by
```
jcfranco, line 255: ✨📚✨

jcfranco (approved):
Nice!

src/components/input-date-picker/input-date-picker.tsx
benelan: This isn't used in input or input number, the reason this is here is because trying to delocalize dates was causing issues since they aren't numbers due to the separators. So I ended up iterating over the characters and just localizing the numbers. It could be useful in input-time for the same reason so it may be worth pulling out into a util or even adding it as a method to the NumberStringFormat class

benelan: To clarify, I added this method along with the formatNumerals one when I set up numbering system in the component but it got removed at some point

benelan (approved):
Awesome, thanks for taking this on!

src/components/input-date-picker/input-date-picker.tsx
jcfranco: Thanks for the background, sir! 🎩

Issue #6348, opened 2023-01-25
[Alert] Incorrect border radius when icon is present


macandcheese:
### Actual Behavior

When an icon is present (default `icon` or supplied as an override - `icon="layers"`) - there is no bottom-left border-radius:
<img width="687" alt="Screen Shot 2023-01-25 at 1 07 09 PM" src="https://user-images.githubusercontent.com/4733155/214691303-a8243e17-796b-4a52-ac54-0dc1d6fa09c9.png">
<img width="687" alt="Screen Shot 2023-01-25 at 1 06 08 PM" src="https://user-images.githubusercontent.com/4733155/214691311-32a2ba03-1d8d-404c-94b0-755e254e1f8c.png">

### Expected Behavior

The alerts with icon have correct border radius.

with icon: https://codepen.io/mac_and_cheese/pen/BaPxpqR?editors=1000
without icon: https://codepen.io/mac_and_cheese/pen/yLqjgjj

### Reproduction Sample

Above

### Reproduction Steps

1. Visit codepen links
2. View difference in styling

### Reproduction Version

1.0.3

### Esri team

Calcite (design)

geospatialem:
cc @ashetland @SkyeSeitz - Would this fix require a Figma UI kit change? We're hoping for this to land in the upcoming patch.

ashetland:
Nope, the Figma component has a 4px radius on all corners in all variants. 🚀

geospatialem:
Verified on `1.0.4-next.2`

<img width="1227" alt="Screenshot 2023-01-28 at 8 57 46 PM" src="https://user-images.githubusercontent.com/5023024/215302084-9b6f2eeb-dc82-4173-a9e7-165879f5ff95.png">

PR #6368, opened 2023-01-27
fix(alert): ensure `border-radius` is consistent for prescribed `slots`

Elijbet:
**Related Issue:** #6348

## Summary
The bottom left corner radius was missing when the icon was present. Upon inspection, it was clear that slots added around content don't have the proper radius set and end up covering the radius on the host. For each occasion the slot is placed against either end of the host, I adjusted the relevant radius to match the host container. 

Because the close button is now always present, the cases addressed are the right side of the close button and the left side of the content (no icon) along with the left side of the icon (when placed in front of the content).

Added a screenshot for the radius with options where `icon` is present and absent.

### DIFF
--- a/src/components/alert/alert.tsx
+++ b/src/components/alert/alert.tsx
@@ -398,7 +398,7 @@ export class Alert implements OpenCloseComponent, LoadableComponent, T9nComponen
   @State() queued = false;
 
   /** the close button element */
-  private closeButton?: HTMLButtonElement;
+  private closeButton: HTMLButtonElement;
 
   private autoCloseTimeoutId: number = null;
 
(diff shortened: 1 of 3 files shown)


src/components/alert/alert.tsx
```diff
@@ -398,7 +398,7 @@ export class Alert implements OpenCloseComponent, LoadableComponent, T9nComponen
   @State() queued = false;
 
   /** the close button element */
-  private closeButton?: HTMLButtonElement;
```
jcfranco, line 401: This is fine for now, but for the future, let's keep these cleanup changes separate unless the code specific PR changes require a refactor or if it's in the modified line vicinity. Reasoning: there are no additional changes here that relevant to the PR.

jcfranco (approved):
Other than a small comment, this LGTM! 🚀

jcfranco:
@Elijbet I forgot to add, can you revisit the PR title to summarize the changes from a consumer perspective (i.e., what does this fix for users)?

Issue #6194, opened 2022-12-28
Input Date Picker Range: start date's hover style starts at the beginning of the month


benelan:
### Actual Behavior

Hovering over start date adds hover style starting at the first day of the month, even when the component has no value.

### Expected Behavior

- For the start date, only the day under the cursor has hover style when the component does not have a value. See `next.701` functionality for reference

### Reproduction Sample

https://esri.github.io/calcite-components/?path=/story/components-controls-inputdatepicker--range

### Reproduction Steps

1. Open the start date picker and hover over the days

### Reproduction Version

next.703

### Relevant Info

I'm pretty sure it's a regression due to #6085, @eriklharper can you take a look?

### Regression?

next.701

### Esri team

Calcite (dev)

geospatialem:
Verified on Storybook and in `1.1.0-next.2`.

PR #6369, opened 2023-01-28
fix(date-picker): days previous to the currently hovered day when no range value exists display correctly with no hover styles

eriklharper:
**Related Issue:** #6194 

## Summary

This PR fixes an issue where the days previous to the currently hovered day are highlighted incorrectly when no start or end value exists.  There should be no hover styles applied to any days other than the currently hovered day unless a start or an end date exists.

### DIFF
--- a/src/components/date-picker-month/date-picker-month.tsx
+++ b/src/components/date-picker-month/date-picker-month.tsx
@@ -480,8 +480,9 @@ export class DatePickerMonth {
       !insideRange &&
       ((!isStart && date >= this.endDate && (date < end || sameDate(date, end))) ||
         (isStart &&
-          (date < this.startDate || (this.endDate && sameDate(date, this.startDate))) &&
-          (date > start || sameDate(date, start))));
+          ((this.startDate && date < this.startDate) ||
+            (this.endDate && sameDate(date, this.startDate))) &&
+          ((start && date > start) || sameDate(date, start))));
     return cond1 || cond2;
   }
 }
--- a/src/components/date-picker/date-picker.tsx
+++ b/src/components/date-picker/date-picker.tsx
@@ -386,15 +386,17 @@ export class DatePicker implements LocalizedComponent, T9nComponent {
           this.hoverRange.focused = "end";
         }
       } else {
-        if (date < start) {
-          this.hoverRange = {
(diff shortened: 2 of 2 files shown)


eriklharper:
My attempt at a visual snapshot test for hovering is a no-go.  Apparently hover events can't be simulated by Storybook's play function according to this in their docs:

image
https://www.chromatic.com/docs/hoverfocus

jcfranco (commented):
~In case any date used in this logic is 0  (e.g., `Wed Dec 31 1969 16:00:00 GMT-0800 (Pacific Standard Time)`), should these date checks explicitly be for null/undefined vs anything falsy?~

Ignore my comment. I thought the conditional would use the Date's value and not the instance. This isn't the case.

jcfranco:
@Elijbet @geospatialem Can you help with manually test this one?

geospatialem:
> @Elijbet @geospatialem Can you help with manually test this one?

@eriklharper @jcfranco Maybe I'm missing something, but the behavior looks similar to what is experienced in the original issue story - maybe because today is Feb 1? Would that be the expectation?

input-date-picker-manual-test

eriklharper:
> > @Elijbet @geospatialem Can you help with manually test this one?
> 
> @eriklharper @jcfranco Maybe I'm missing something, but the behavior looks similar to what is experienced in the original issue story - maybe because today is Feb 1? Would that be the expectation?
> 
> input-date-picker-manual-test [ input-date-picker-manual-test ](https://user-images.githubusercontent.com/5023024/216241355-680e6dac-92bf-485b-b670-f5fb20866de2.gif) [ ](https://user-images.githubusercontent.com/5023024/216241355-680e6dac-92bf-485b-b670-f5fb20866de2.gif)

Yes, because today is the 1st, you won't really be able to repro this because the bug has to do with days previous to the current day being incorrectly highlighted.  Try this again tomorrow and you should see the effect!  :)

jcfranco:
> Try this again tomorrow and you should see the effect!

Or you could tweak your system settings too. :)

delorean-time-travel

geospatialem:
This looks good on my end - nice work @eriklharper! 🌟 📅 🌟 


input-date-picker-manual-test2

Issue #6367, opened 2023-01-27
Tree item has visual vertical line present


geospatialem:
### Actual Behavior

A vertical line is present across the `tree` and its child `tree-item`s throughout, whether or not the `tree-item`s are expanded.

Was introduced with the fix from https://github.com/Esri/calcite-components/issues/6284. 

cc: @omarkawach

![image](https://user-images.githubusercontent.com/5023024/215210344-dcd036a9-0545-416c-9d47-8b72bcc148a1.png)

### Expected Behavior

A verifical line present on the expanded `tree` and its `tree-item`s.

![image](https://user-images.githubusercontent.com/5023024/215210490-3bb3ae3b-0669-4269-90d8-a1de33310848.png)

### Reproduction Sample

https://codepen.io/geospatialem/pen/abjKEaO

### Reproduction Steps

1. Open the [Codepen sample](https://codepen.io/geospatialem/pen/abjKEaO)
2. Observe the vertical line present when the `lines` prop is present

### Reproduction Version

1.0.3

### Regression?

beta.99

### Impact

Moderate. A visual regression, which we should try to mitigate, but does not impact the functionality.

### Esri team

Calcite (dev)

ashetland:
@jcfranco, @geospatialem beat me to it!

geospatialem:
Verified on `1.0.4-next.4`

![image](https://user-images.githubusercontent.com/5023024/215670833-cf70ec23-f126-4284-8b9f-624a412ea7fe.png)

PR #6372, opened 2023-01-28
fix(tree): prevent lines from expanded item from bleeding out of container

jcfranco:
**Related Issue:** #6367

## Summary

Restores styles needed to properly contain expanded item line that removed in https://github.com/Esri/calcite-components/pull/6005.


ashetland (approved):

Elijbet (approved):
🫧✨ You're right. Looks like this bit did not need to get removed in the previous fix. 🫧✨

eriklharper (approved):
Verified locally, LGTM! 👍

Issue #6331, opened 2023-01-24
stepper : numbering-system displaying inconsistent results with two stepper components in a page


anveshmekala:
### Actual Behavior

if two numbered stepper components are rendered in a page with different numbering-system, when user alternate the clicks between stepper components the numbers are showing inconsistent results.

### Expected Behavior

stepper components should follow numbering-system defined on them or should fall back to `latn` numbers.

### Reproduction Sample

https://codepen.io/ANVESHMEKALA/pen/YzjaveB?editors=100

### Reproduction Steps

1. Click on the stepper item with arabic numbers
2. Then, click on the stepper item above with latin numbers
3.  Observe that latin numbers changes to arabic.

### Reproduction Version

next.729

### Regression?

beta.97

### Esri team

N/A

geospatialem:
Verified in `1.0.4-next.5`

PR #6380, opened 2023-01-30
fix(stepper-item): no longer refer numberingSystem from neighbor stepper component

anveshmekala:
**Related Issue:** #6331 

## Summary

This PR will fix the issue of `stepper` component's numbers being localized to the `numberingSystem` specified by neighboring `stepper` component rendered in same page.

### DIFF
--- a/src/components/stepper-item/stepper-item.tsx
+++ b/src/components/stepper-item/stepper-item.tsx
@@ -182,12 +182,6 @@ export class StepperItem implements InteractiveComponent, LocalizedComponent, Lo
     if (this.selected) {
       this.emitRequestedItem();
     }
-
-    numberStringFormatter.numberFormatOptions = {
-      locale: this.effectiveLocale,
-      numberingSystem: this.parentStepperEl?.numberingSystem,
-      useGrouping: false
-    };
   }
 
   componentDidLoad(): void {
@@ -219,11 +213,7 @@ export class StepperItem implements InteractiveComponent, LocalizedComponent, Lo
             }
           >
             {this.icon ? this.renderIcon() : null}
-            {this.numbered ? (
-              <div class="stepper-item-number">
-                {numberStringFormatter.numberFormatter.format(this.itemPosition + 1)}.
-              </div>
-            ) : null}
+            {this.numbered ? <div class="stepper-item-number">{this.renderNumbers()}.</div> : null}
             <div class="stepper-item-header-text">
               <span class="stepper-item-heading">{this.heading}</span>
               <span class="stepper-item-description">{this.description}</span>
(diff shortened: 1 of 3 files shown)


src/components/stepper/stepper.tsx
```diff
@@ -143,7 +143,8 @@ export class Stepper {
     event.stopPropagation();
   }
 
-  @Listen("calciteInternalStepperItemRegister") registerItem(event: CustomEvent): void {
+  @Listen("calciteInternalStepperItemRegister")
+  registerItem(event: CustomEvent): void {
```
anveshmekala, line 147: made this change to avoid JSDOC warning . unrelated to the issue.

src/components/stepper/stepper.tsx
```diff
@@ -153,7 +154,8 @@ export class Stepper {
     event.stopPropagation();
   }
 
-  @Listen("calciteInternalStepperItemSelect") updateItem(event: CustomEvent): void {
+  @Listen("calciteInternalStepperItemSelect")
+  updateItem(event: CustomEvent): void {
```
anveshmekala, line 158: made this change to avoid JSDOC warning . unrelated to the issue.

src/components/stepper/stepper.e2e.ts
```diff
@@ -577,4 +577,30 @@ describe("calcite-stepper", () => {
...
+    </calcite-stepper>`);
+    const [stepper1, stepper2] = await page.findAll("calcite-stepper");
+    expect(stepper2.getAttribute("numbering-system")).toEqual("arab");
+
+    await stepper1.click();
+    await page.waitForChanges();
+    await stepper2.click();
+    await page.waitForChanges();
+    await stepper1.click();
+    await page.waitForChanges();
```
anveshmekala, line 601: Main reason behind adding clicks is to make sure we reproduce the error state. Sometimes, on initial render the error may appear.

src/components/stepper/stepper.tsx
jcfranco: You mean linting warning?

jcfranco (approved):
Nice! 

I believe the issue is with the formatter cache itself, but this is good for the issue at hand. Adding @benelan to take a look. 

If this looks good to merge, @anveshmekala, can you create an issue to fix the cache?

benelan (approved):
LGTM, thanks!

src/components/stepper-item/stepper-item.tsx
```diff
@@ -219,11 +213,7 @@ export class StepperItem implements InteractiveComponent, LocalizedComponent, Lo
             }
           >
             {this.icon ? this.renderIcon() : null}
-            {this.numbered ? (
-              <div class="stepper-item-number">
-                {numberStringFormatter.numberFormatter.format(this.itemPosition + 1)}.
-              </div>
-            ) : null}
+            {this.numbered ? <div class="stepper-item-number">{this.renderNumbers()}.</div> : null}
```
benelan, line 216: We might want to move the numberStringFormatter check to the render block for all of its components to prevent similar bugs. It won't recreate a new NumberFormatter if the lang/numberingSystem is the same so putting the check in the render block won't be too expensive performance-wise. 

I can do that in a follow up PR

src/components/stepper/stepper.tsx
anveshmekala: yup. linting warning.

anveshmekala:
> If this looks good to merge, @anveshmekala, can you create an issue to fix the cache?

Added an issue https://github.com/Esri/calcite-components/issues/6399.

Issue #5227, opened 2022-08-26
[Stepper-item] vertical layout bottom padding only applied when there is a slotted element


ashetland:
### Actual Behavior

On vertical layout:
- No bottom padding when content is text
- 20px bottom padding when content is slotted element
<img width="339" alt="CleanShot 2022-08-26 at 11 15 47@2x" src="https://user-images.githubusercontent.com/108549080/186967318-e8dc7631-2f44-46d6-a52b-08657956bcb8.png">

### Expected Behavior

Bottom padding should be consistent regardless of content type.

### Reproduction Sample

https://codepen.io/ashetland/pen/WNzBqpv

### Reproduction Steps

1. Insert calcite-stepper
2. Add text to content slot in stepper-item. No bottom padding present.
3. Replace text with calcite-notice. 20px bottom padding is added.

### Reproduction Version

beta.93

### Relevant Info

Mac 12.5.1
Chrome 104

geospatialem:
The Figma UI kit contains the above fix, as such the fix should be included in the January release.

benelan:
Should the bottom padding be 20px or 0?

ashetland:
There should always be 20px bottom padding.

ashetland:
16px on small scale, and 24px on large

benelan:
Seems like the only way to target text nodes via CSS is experimental: [`::target-text`](https://developer.mozilla.org/en-US/docs/Web/CSS/::target-text)

You could wrap the next node in a `span`, is that sufficient or should I try to figure something else out?
https://codepen.io/benesri/pen/poZyGJa

benelan:
The alternative is setting the margin on the container instead of the slotted content, but then the line on the left won't extend the full height of the focus ring

![image](https://user-images.githubusercontent.com/10986395/210118732-f79c4d44-ec3d-4ea5-8b84-1ce3a6246a76.png)

benelan:
Nevermind lol, PR coming up

anveshmekala:
Added a prototype solution in #6385 which leverages `onSlotChange`  event and adds margin if text node with a value is rendered.  @benelan , @jcfranco   WDYT?

geospatialem:
Verified in `1.1.0-next.2`

<img width="754" alt="Screenshot 2023-02-06 at 9 11 27 PM" src="https://user-images.githubusercontent.com/5023024/217139094-402e58b3-d299-484d-9fc6-3c3f444fe720.png">

PR #6385, opened 2023-01-31
fix(stepper): add margin bottom for text in vertical layout

anveshmekala:
**Related Issue:** #5227 

## Summary
This PR will add bottom margin for text node in vertical layout.

### DIFF
--- a/src/components/stepper/stepper.stories.ts
+++ b/src/components/stepper/stepper.stories.ts
@@ -191,3 +191,9 @@ export const arabicNumberingSystem_TestOnly = (): string => html` <calcite-stepp
     </calcite-notice>
   </calcite-stepper-item>
 </calcite-stepper>`;
+
+export const verticalLayout_TestOnly = (): string => html`<calcite-stepper layout="vertical">
+  <calcite-stepper-item heading="Choose method" description="Add members without sending invitations"
+    >Step 1 Content Goes Here</calcite-stepper-item
+  >
+</calcite-stepper>`;
--- a/src/components/stepper/stepper.stories.ts
+++ b/src/components/stepper/stepper.stories.ts
@@ -192,8 +192,20 @@ export const arabicNumberingSystem_TestOnly = (): string => html` <calcite-stepp
   </calcite-stepper-item>
 </calcite-stepper>`;
 
-export const verticalLayout_TestOnly = (): string => html`<calcite-stepper layout="vertical">
-  <calcite-stepper-item heading="Choose method" description="Add members without sending invitations"
-    >Step 1 Content Goes Here</calcite-stepper-item
-  >
-</calcite-stepper>`;
+export const verticalLayout_TestOnly = (): string => html`<calcite-stepper layout="vertical" scale="s">
+    <calcite-stepper-item heading="Scale s" description="Add members without sending invitations"
+      >Step 1 Content Goes Here</calcite-stepper-item
+    >
+  </calcite-stepper>
+
+  <calcite-stepper layout="vertical">
+    <calcite-stepper-item heading="Scale m" description="Add members without sending invitations"
+      >Step 1 Content Goes Here</calcite-stepper-item
+    >
+  </calcite-stepper>
+
+  <calcite-stepper layout="vertical" scale="l">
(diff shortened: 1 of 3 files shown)


anveshmekala:
Updated the solution to use `:empty` CSS selector as suggested by @jcfranco .

anveshmekala:
> Could you add small and large scales to the Chromatic review? The bottoms space should be 16px for small and 24px for large.

Good idea. Added.

jcfranco (approved):
Very nice and clean! 💯

src/components/stepper/stepper.stories.ts
```diff
@@ -191,3 +191,21 @@ export const arabicNumberingSystem_TestOnly = (): string => html` <calcite-stepp
     </calcite-notice>
   </calcite-stepper-item>
 </calcite-stepper>`;
+
+export const verticalLayout_TestOnly = (): string => html`<calcite-stepper layout="vertical" scale="s">
```
jcfranco, line 195: @benelan WDYT about adding a screenshot test convention for covering scale-specific issues in a single story? @Elijbet You've added similar stories, right?

Elijbet: A convention would be good. Here are some examples. Sometimes for the layout testing, I include other prop variations to see if they affect/interact with the prop I'm testing at different scales(eg props: heading/description/long title/slots, etc).

```mediumIconForLargeAccordionItem_TestOnly
withActions
actionsEndNoQueue_TestOnly
actionsEndQueued_TestOnly
mediumIconForLargeComoboboxItem_TestOnly
mediumIconForLargeDropdownItem_TestOnly
mediumIconForLargeScaleStyling_TestOnly
largeScaleLayout_TestOnly

Combining non default scale test with other tests:

hebrewNumberingSystemAndMediumIconForLargeInputStyling_TestOnly
disabled_TestOnly
laoNumberingSystemAndMediumIconForLargeInput_TestOnly
darkModeFrenchLocaleAndLargeScaleGetsMediumChevron_TestOnly
disabledAndLargeScaleGetsMediumChevron_TestOnly
disabledTabsAndMediumIconsForLargeTabsTitle_TestOnly```

anveshmekala: if the UI isnt affected by other props apart from `scale`, i think its better to include all the scales in a single test so that we can saveup on screenshots. If we had to include other props and see how it behaves , then we may have tests them separately as mentioned above. Thoughts @Elijbet , @jcfranco , @benelan ?

Elijbet: Not sure, here for example I put all variations in the same screenshot https://github.com/Esri/calcite-components/pull/6403
```export const iconStartAndActionsEnd = (): string => html`
  <div style="width: 650px">
    <calcite-tree style="margin: 80px" scale="l"> ${iconStartLargeActionAndActionsEnd} </calcite-tree>
    <calcite-tree style="margin: 80px" scale="m"> ${slottedSmallAction} </calcite-tree>
    <calcite-tree style="margin: 80px" scale="s"> ${slottedSmallActionsEndAndAction} </calcite-tree>
  </div>
`;```

anveshmekala: makes sense. The same applies for this PR too, me thinks.

benelan: Yeah I like that idea of putting them all in the same story. I had been recommending setting different scales for existing stories, as long as it doesn't conflict with the existing coverage. For example changing the scale to `s` on a story testing dark mode or a different lang/numberingSystem. We would just need to alter the story name to mention both coverage cases.

If there isn't a way to alter an existing story, I'd say putting all three in the same story is our second best option. 

In terms of adding conventions, I'd like to wait another month or two when we'll know what our Chromatic snapshot usage is under normal conditions. The first month we went over the limit due to testing during initial setup. The second month we were way under the limit due to holidays. So I'd like to see if we have any wiggle room before deciding how we should handle the situation in the long term.

And to clarify, we should definitely be adding coverage to scale issues as we fix them right now, but if we have some wiggle room it could be nice to add scale coverage to all of our components to catch issues before they arise. Just not sure the best way to do that until we know our usage.

Issue #6386, opened 2023-01-31
Extra space in front of the `input` in `input-number` and `input-text`


Elijbet:
### Actual Behavior

The extra space in front of the input in input-number and input-text is coming from specifying `icon`, but not having a recommended icon for these 2 types. The solution is to not support the boolean type on the icon on them. 

### Expected Behavior

Get rid of the space. 

### Reproduction Sample

https://codepen.io/elijbet/pen/WNKggyE

### Reproduction Steps

See demo.

### Reproduction Version

1.0.3

### Esri team

Calcite (dev)

geospatialem:
Verified on `master`.

cc: @macandcheese - The change includes the JSDoc tag of `@futureBreaking`, we might need to consider some styling options for the doc site for these use cases.

PR #6394, opened 2023-01-31
docs(input-number, input-text): remove unsupported boolean type from type prop doc

Elijbet:
**Related Issue:** #6386

## Summary

In the future, we will be removing the support of the `boolean` type for the `icon` prop on `input-number` and `input-text` as these don't have a `default` recommended icon for an unspecified icon attribute and result in unwanted spacing preceding the `input`. 

For now though, since it is potentially a minor breaking change that could possibly be in use somewhere, we removed the `boolean` scenario from the documentation and marked it with a newly introduced custom `@futureBreaking` tag.

### DIFF
--- a/src/components/input-number/input-number.scss
+++ b/src/components/input-number/input-number.scss
@@ -137,15 +137,15 @@
 
 // position placeholder/value text in relation to icons
 
-:host([icon][scale="s"]) input {
+:host([icon]:not([icon=""])[scale="s"]) input {
   padding-inline-start: theme("padding.8");
 }
 
-:host([icon][scale="m"]) input {
+:host([icon]:not([icon=""])[scale="m"]) input {
   padding-inline-start: theme("padding.10");
 }
 
-:host([icon][scale="l"]) input {
+:host([icon]:not([icon=""])[scale="l"]) input {
   padding-inline-start: theme("padding.14");
 }
 
--- a/src/components/input-number/input-number.scss
+++ b/src/components/input-number/input-number.scss
@@ -137,15 +137,15 @@
 
 // position placeholder/value text in relation to icons
 
-:host([icon]:not([icon=""])[scale="s"]) input {
+:host([icon][scale="s"]) input {
   padding-inline-start: theme("padding.8");
 }
 
-:host([icon]:not([icon=""])[scale="m"]) input {
+:host([icon][scale="m"]) input {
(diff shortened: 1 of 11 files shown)


src/components/input-number/input-number.tsx
```diff
@@ -808,7 +808,7 @@ export class InputNumber
     };
 
     const sanitizedValue = sanitizeNumberString(
-        // no need to delocalize a string that ia already in latn numerals
+      // no need to delocalize a string that ia already in latn numerals
```
jcfranco, line 811: Was formatted by the linter?

Typo (unrelated): `ia` -> `is`

src/components/input-number/input-number.scss
```diff
@@ -137,15 +137,15 @@
 
 // position placeholder/value text in relation to icons
 
-:host([icon][scale="s"]) input {
+:host([icon]:not([icon=""])[scale="s"]) input {
```
jcfranco, line 140: If the doc and type no longer allows boolean, I think we can drop `:not([icon=""])` and assume `icon` has a value.

src/components/input-number/input-number.tsx
Elijbet: Yes it is.

src/components/input-number/input-number.tsx
```diff
@@ -131,9 +131,9 @@ export class InputNumber
   @Prop({ reflect: true }) hidden = false;
 
   /**
-   * When `true`, shows a default recommended icon. Alternatively, pass a Calcite UI Icon name to display a specific icon.
+   * Pass a Calcite UI Icon name to display a specific icon.
```
geospatialem, line 134: WDYT about the following (in line with other components such as `action` and `chip`:

> Specifies an icon to display.

src/components/input-text/input-text.tsx
```diff
@@ -109,9 +109,9 @@ export class InputText
   @Prop({ reflect: true }) hidden = false;
 
   /**
-   * When `true`, shows a default recommended icon. Alternatively, pass a Calcite UI Icon name to display a specific icon.
+   * Pass a Calcite UI Icon name to display a specific icon.
```
geospatialem, line 112: Similar to above:

> Specifies an icon to display.

jcfranco (approved):
💥

Issue #6401, opened 2023-01-31
value-list: screen reader instructions are missing when drag handle is activated


anveshmekala:
### Actual Behavior

when the user tabs into the dragHandle and activate with space key, no instruction is displayed in screen reader. 

### Expected Behavior

screen reader instructions should be displayed when the dragHandle is activated/ deactivated. 

### Reproduction Sample

https://codepen.io/ANVESHMEKALA/pen/BaPqVgV?editors=1010

### Reproduction Steps

1. Start the screen reader mode in the codepen . 
2. Try navigation into the first value-list-item. Make sure the screen reader reads out  the instruction 
3. Hit space to activate the dragHandle.
4. Observe no instruction is announced by the screen reader.

### Reproduction Version

bet.99

### Relevant Info

cancelling out the event [here](https://github.com/Esri/calcite-components/pull/5709/files#diff-9cb5506aef9cc2ff9af5ee350a918a33151cda203737921476258b8014387de7R179) is not allowing the screen reader text to update.

### Esri team

N/A

geospatialem:
This resolution will likely only apply to Mac and VoiceOver, as there was no previous support for Windows machines with NVDA and JAWS. cc https://github.com/Esri/calcite-components/issues/3982#issuecomment-1206487248. 

In the long term we should explore some alternatives with the `list` component when we add dragging functionality to the component.

geospatialem:
Verified on `1.0.8-next.3` with Safari and VoiceOver. Both NVDA and JAWS are unable to access when activated, related to #3982.

Issue #5739, opened 2022-11-10
value-list: Unstable test


driskull:
### Test type

e2e

### Which Component(s)

value-list

### Unstable Tests

it.skip("is drag and drop list accessible", async () => {

### Test error, if applicable

```
calcite-value-list › drag and drop › is drag and drop list accessible
    expect(received).toBe(expected) // Object.is equality
    Expected: "Reordering One, current position 1 of 3."
    Received: ""
      271 |       await page.waitForChanges();
      272 |
    > 273 |       expect(assistiveTextElement.textContent).toBe(
          |                                                ^
      274 |         `Reordering ${itemLabel}, current position ${startIndex + 1} of ${items.length}.`
      275 |       );
      276 |
      at Object.<anonymous> (src/components/value-list/value-list.e2e.ts:273:48)
          at runMicrotasks (<anonymous>)
```

### PR skipped, if applicable

#5738

anveshmekala:
the test is stable and the failure reported earlier is a true error case. closing this as verified.

geospatialem:
Closing this as it has been verified.

PR #6402, opened 2023-01-31
fix(value-list): add back instructions for screen reader when drag handle is activated

anveshmekala:
**Related Issue:** #6401 

## Summary

This PR will add back instructions for screen reader when drag handle is activated via `Space` key. 

Sidebar: This will also cover #5739

### DIFF
--- a/src/components/value-list-item/value-list-item.tsx
+++ b/src/components/value-list-item/value-list-item.tsx
@@ -187,7 +187,7 @@ export class ValueListItem
   /**
    * @internal
    */
-  @Event({ cancelable: true })
+  @Event({ cancelable: false })
   calciteValueListItemDragHandleBlur: EventEmitter<ListItemAndHandle>;
 
   @Listen("calciteListItemChange")
--- a/src/components/value-list-item/value-list-item.tsx
+++ b/src/components/value-list-item/value-list-item.tsx
@@ -199,7 +199,6 @@ export class ValueListItem
 
   handleKeyDown = (event: KeyboardEvent): void => {
     if (event.key === " ") {
-      event.preventDefault();
       this.handleActivated = !this.handleActivated;
     }
   };
--- a/src/components/value-list-item/value-list-item.tsx
+++ b/src/components/value-list-item/value-list-item.tsx
@@ -185,11 +185,10 @@ export class ValueListItem
   @Event({ cancelable: true }) calciteListItemRemove: EventEmitter<void>; // wrapped pick-list-item emits this
 
   /**
-   * Fires when the drag handle is focused
    * @internal
    */
   @Event({ cancelable: true })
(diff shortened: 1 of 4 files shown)


src/components/value-list/value-list.tsx
```diff
@@ -457,6 +458,14 @@ export class ValueList<
     }
   };
 
+  @Listen("calciteValueListItemDragHandleBlur")
+  handleValueListItemBlur(event: CustomEvent<ListItemAndHandle>): void {
+    const { item, handle } = event.detail;
```
jcfranco, line 463: [Per our conventions](https://github.com/Esri/calcite-components/blob/master/conventions/README.md#privateinternal-events), the propagation of custom events should be stopped after being handled.

src/components/value-list-item/value-list-item.tsx
```diff
@@ -181,6 +184,12 @@ export class ValueListItem
    */
   @Event({ cancelable: true }) calciteListItemRemove: EventEmitter<void>; // wrapped pick-list-item emits this
 
+  /**
+   * @internal
+   */
+  @Event({ cancelable: true })
+  calciteValueListItemDragHandleBlur: EventEmitter<ListItemAndHandle>;
```
jcfranco, line 191: Is it possible to use the `getHandleAndItemElement` util to avoid passing the item and handle here?

src/components/value-list/value-list.tsx
```diff
@@ -457,6 +458,14 @@ export class ValueList<
     }
   };
 
+  @Listen("calciteValueListItemDragHandleBlur")
```
jcfranco, line 461: I _think_ this can be achieved by using `focusout`, but let's proceed with this for now and please open a follow-up issue to revisit this approach.

src/components/value-list-item/value-list-item.tsx
```diff
@@ -181,6 +184,12 @@ export class ValueListItem
    */
   @Event({ cancelable: true }) calciteListItemRemove: EventEmitter<void>; // wrapped pick-list-item emits this
 
+  /**
+   * @internal
+   */
+  @Event({ cancelable: true })
```
jcfranco, line 190: This event is marked as cancelable, but there doesn't appear to be any code paths that handle cancelation. Please set to `false` if this is the case.

jcfranco (approved):

src/components/value-list-item/value-list-item.tsx
anveshmekala: i tried doing so but the `focusEvent`   doesn't show up the handle element `composedPath` or `event.target`

src/components/value-list/value-list.tsx
anveshmekala: Added a follow-up issue https://github.com/Esri/calcite-components/issues/6526

Issue #6361, opened 2023-01-26
fix(tree-item): consistent height with or w/t `action-end`


Elijbet:
### Actual Behavior
Expected consistent height with our without action-end as per spec:

Small: 24
Medium: 32
Large: 44

Padding on the action-end slot currently affects the height of the host. For the slot in tree-item to not affect the tree layout, we need to take it outside of the layout, sitting it next to node-container. Similar to notice. This way the padding on the node-container will not affect it. Assuming the user sets a proper action scale it should match the height.

### Expected Behavior
Follows the design spec.

### Reproduction Sample
https://codepen.io/elijbet/pen/YzjvKdR?editors=1010

### Reproduction Steps
Use inspect to explore how bringing in action-end bumps up the height of the tree-item.

### Reproduction Version
1.0.3

### Relevant Info
No response

### Regression?
No response

### Impact
No response

### Esri team
Calcite (dev)

geospatialem:
Migrating to the upcoming February milestone for the fix to land later this month.

geospatialem:
Verified in `1.0.6`. There was a tradeoff with the focus pertaining to `action-end`, and a new issue #6481 was created to track progress.

PR #6403, opened 2023-01-31
fix(tree-item): preserves consistent height with or w/t actions-end 

Elijbet:
**Related Issue:** #6361, #3127

## Summary
Moved the `actions-end` slot outside of the `node-container` layout and placed it next to it instead, similar to `notice`. This way the padding on the `node-container` does not add to the action-ends.

Expected consistent height on the `tree-item` when bringing in `actions-end` can be preserved like this (here to show designers how this can be demo-ed to end-users).

s (24) tree-item:                               s slotted component + s action 
m (32) tree-item:                              m slotted component + m action 
l (48) tree-item:                                l slotted component + l action 

https://user-images.githubusercontent.com/19231036/216716464-8413540a-85f5-4183-a2b1-b67652ae1384.mov

We need to follow up on clipping the action:

- We’re not normally clipping actions to accommodate component heights, it’s inconsistent with how we normally handle slots. So here I’m clipping the actions for the time being and marking this internally as a known issue (Figma and implementation side)
- Large clips outline: clips top and bottom, will be added to the note above.


src/components/tree-item/tree-item.tsx
```diff
@@ -311,22 +311,25 @@ export class TreeItem
...
-              [CSS_UTILITY.rtl]: rtl
-            }}
-            data-selection-mode={this.selectionMode}
-            ref={(el) => (this.defaultSlotWrapper = el as HTMLElement)}
-          >
-            {chevron}
-            {itemIndicator}
-            {this.iconStart ? iconStartEl : null}
-            {checkbox ? checkbox : defaultSlotNode}
+          <div class={CSS.nodeAndActionsContainer}>
```
jcfranco, line 314: Would it be possible to merge this with the above container node?

src/components/tree-item/tree-item.scss
```diff
@@ -239,7 +251,7 @@ slot[name="actions-end"]::slotted(*),
     transform: rotate(180deg);
   }
 
-  .item--expanded > .node-container > & {
+  .item--expanded .node-container > & {
```
jcfranco, line 254: @benelan can you double check this one? I noticed it was left intact [here](https://github.com/Esri/calcite-components/pull/6320/files).

src/components/tree-item/tree-item.tsx
jcfranco: Per our conversation, disregard this one. I didn't notice `nodeAndActionsContainer` didn't contain all of the expanded node content.

src/components/tree-item/tree-item.scss
benelan: I added a div with the `item--expanded` class between the `:host` and the `node-container` elements. So the `node-container` is now the direct child of `item--expanded` and the grandchild of `:host`. That's why I had to change all of the `:host > .node-container` selectors. 

It's fine with the change if it makes it easier to read/discern, the way it was before is technically more performant but not enough to be noticible. I'm just a weirdo who considers CSS selector optimization lool

Elijbet: Cool. I removed the direct descendant selector because now there is an added div in-between `.nodeAndActionsContainer`.

jcfranco (approved):
Code LGTM!

+@macandcheese Can you review the styles to see if there's any noticeable layout changes? We may want to ping teams for extra testing if this is the case.

jcfranco:
+@ashetland for updated screenshot review.

macandcheese:
Reviewed the Chromatic - The heights themselves look good! The focus outline will always be cut off at the top and bottom with the way the action element is hidden outside the list-item extent.

ashetland:
For some reason the small Tree Items with small Actions (cropped to 24px) show the full focus outline, but the other scales don't. Not sure what special about that one.

Also, I just noticed the Tree Items themselves have a white focus outline instead of brand. Is that something we need to look at?

Elijbet:
> For some reason the small Tree Items with small Actions (cropped to 24px) show the full focus outline, but the other scales don't. Not sure what special about that one.
> 
> Also, I just noticed the Tree Items themselves have a white focus outline instead of brand. Is that something we need to look at?

The outline is being worked on right now, it happens because the action is cropped, as requested by design. Adam mentioned the white focus and I cannot see it on my end for some reason.

geospatialem:
> The focus outline will always be cut off at the top and bottom with the way the action element is hidden outside the list-item extent.

> it happens because the action is cropped, as requested by design

@ashetland @macandcheese With the above comments, should we open a new issue to mitigate the action focus? Seems to be an unexpected behavior.

verify-tree

PR #6406, opened 2023-02-01
ci(release): lint readme and changelog before committing

benelan:
**Related Issue:** #6355

## Summary
I added markdown lint in the PR above but it isn't running during releases because we have `standard-version` set to `noVerify`. So this is manually linting the readme/changelog before committing.


jcfranco (approved):
✨🧹✨

PR #6415, opened 2023-02-03
test(heading): remove unnecessary dummy component for spec test

jcfranco:
**Related Issue:** N/A

## Summary

Passing a component is not necessary for this spec page test, so this cleans that up. This came up after some digging around where local test runs (w/ `--watchAll`) would fail after making changes to code:

> [ ERROR ]  Rollup: Parse Error
           Unexpected token (Note that you need plugins to import files that are not JavaScript) 94: export {
           CalciteValueList, defineCustomElement as defineCustomElementCalciteValueList } from
           'CalciteValueList'; 95: export { CalciteValueListItem, defineCustomElement as
           defineCustomElementCalciteValueListItem } from 'CalciteValueListItem'; 96: export { {DummyComponent
           } from 'DummyComponent'; ^

### DIFF
--- a/src/components/functional/Heading.spec.tsx
+++ b/src/components/functional/Heading.spec.tsx
@@ -1,4 +1,4 @@
-import { Component, h } from "@stencil/core";
+import { h } from "@stencil/core";
 import { newSpecPage } from "@stencil/core/testing";
 import { constrainHeadingLevel, Heading } from "./Heading";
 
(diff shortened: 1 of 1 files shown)


benelan (approved):
Yeah we only want smart components up in hurrr

> Hey `calcite-accordion`, what's the weather like today?

PR #6416, opened 2023-02-03
ci(eslint): ignore private/internal code for jsdoc rules

benelan:
**Related Issue:** N/A

## Summary

Prevents  [`eslint-jsdoc-plugin`](https://www.npmjs.com/package/eslint-plugin-jsdoc) from complaining when an interface/function/class/etc is marked as `@internal` or `@private` in its jsdoc block. We will need to go through the code and appropriately mark a lot of our utils as `@internal`.


benelan:
Nice find, not sure how I missed that. Her

- current number of warnings: 629
- warnings when using ignorePrivate and ignoreInternal: 589
- warnings when disabling the 3 rules: 121

So it's a bit better but still a lot of noise. We could go through and mark all of the util functions as internal I guess

jcfranco:
> We could go through and mark all of the util functions as internal I guess

I think we would want to start doing this eventually. TBH, I'm torn on this one. On the one hand, I'm hesitant to disable valid rules, but on the other, it's hard to spot valid public ones from the rest. Maybe we can coordinate updating all of the internal ones to lessen the noise?

benelan:
> Maybe we can coordinate updating all of the internal ones to lessen the noise?

Yeah I'm cool with this option. Marking stuff as internal has an effect on the types though, so we will need to be careful. I switched up this PR to ignore private/internal code using the settings you found

jcfranco (approved):

PR #6418, opened 2023-02-03
docs: fix deprecated tag to be in line with Calcite JSDoc guidance

geospatialem:
**Related Issue:** n/a

## Summary
Updates the `@deprecated` tag to the `@futureBreaking` tag for the `pick-list-*` and `value-list-*` components.

Noticed the `@deprecated` tag was creating an issue with loading in the slots on our doc site, so hoping this PR:
1. Loads the slots appropiately, and
2. Provides clarity with our direction using the `@featureBreaking` JSDoc tag.

cc @jcfranco for awareness

### DIFF
--- a/src/components/pick-list-group/pick-list-group.tsx
+++ b/src/components/pick-list-group/pick-list-group.tsx
@@ -12,7 +12,7 @@ import { CSS, SLOTS } from "./resources";
  * @slot - A slot for adding `calcite-pick-list-item` elements.
  */
 
-/** @futureBreaking Use the `list` component instead. */
+/** @futureBreaking Deprecated in 1.0.3. Use the `list` component instead. */
 @Component({
   tag: "calcite-pick-list-group",
   styleUrl: "pick-list-group.scss",
--- a/src/components/pick-list-group/pick-list-group.tsx
+++ b/src/components/pick-list-group/pick-list-group.tsx
@@ -12,7 +12,7 @@ import { CSS, SLOTS } from "./resources";
  * @slot - A slot for adding `calcite-pick-list-item` elements.
  */
 
-/** @deprecated Use the `list` component instead. */
+/** @futureBreaking Use the `list` component instead. */
 @Component({
   tag: "calcite-pick-list-group",
   styleUrl: "pick-list-group.scss",
--- a/src/components/pick-list-item/pick-list-item.tsx
+++ b/src/components/pick-list-item/pick-list-item.tsx
@@ -41,7 +41,7 @@ import { CSS, ICONS, SLOTS } from "./resources";
  * @slot actions-start - A slot for adding `calcite-action`s or content to the start side of the component.
  */
 
-/** @futureBreaking Use the `list` component instead. */
+/** @futureBreaking Deprecated in 1.0.3. Use the `list` component instead. */
 @Component({
   tag: "calcite-pick-list-item",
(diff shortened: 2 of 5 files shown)


macandcheese (commented):
I think we'd want to maintain the `deprecated` tag - I haven't set up any logic to handle and display `futureBreaking` yet, so it won't work in a local doc site preview referencing the latest `docs-json`.

One issue we'll see is that Gatsby / graphql implodes when an expected piece of data is missing from a query (ie, we are looking for a deprecated tag, and none exists), so we might need to fudge that somehow in the future.

jcfranco:
> I think we'd want to maintain the deprecated tag

Correct. This is on me for not clarifying sooner. 😅 We should continue using `@deprecated` as needed for public API changes. `@futureBreaking` is more intended to help keep track of internal tweaks that could be breaking, but won't necessarily make sense to mark as deprecated in our public doc. At least that's how I was thinking about it. 

For more context, this came up in https://github.com/Esri/calcite-components/pull/6394 where the input-number/text's `icon` prop is a boolean, which is not applicable (number and text have no default icon). This was easily fixed by updating the doc, but removing the boolean type would break anyone setting it this way (very unlikely to be the case, but it would break this scenario if we made the change and released). I guess another way of thinking about it would be "low priority, breaking change candidate". Having said that, I'm up for suggestions if this would end up causing more confusion. cc @Elijbet

macandcheese:
The issue of slots not making in to the `docs-json` also happens with `futureBreaking` - it is because they are two separate commented sections. 

Neither

```
/**
 * @slot - A slot for adding `calcite-value-list-item` elements. List items are displayed as a vertical list.
 * @slot menu-actions - A slot for adding a button and menu combination for performing actions, such as sorting.
 */

/** @futureBreaking Use the `list` component instead. */

```
or 

```
/**
 * @slot - A slot for adding `calcite-value-list-item` elements. List items are displayed as a vertical list.
 * @slot menu-actions - A slot for adding a button and menu combination for performing actions, such as sorting.
 */

/** @deprecated Use the `list` component instead. */
```
will add both the second tag and slots to the `docs-json`, but this works:

```
/**
 * @deprecated Use the `list` component instead.
 * @slot - A slot for adding `calcite-value-list-item` elements. List items are displayed as a vertical list.
 * @slot menu-actions - A slot for adding a button and menu combination for performing actions, such as sorting.
 */
 ```

macandcheese:
For reference, this is the output of the above: 

`docs-json` adds the slots to their own section in addition to appearing in `docsTags`, as well as a separate `deprecation` metadata item. Since `futureBreaking` is a custom tag, it only shows in `docsTags`, which is fine / usable.
Screen Shot 2023-02-03 at 10 31 26 AM

Elijbet:
> For reference, this is the output of the above:
> 
> `docs-json` adds the slots to their own section in addition to appearing in `docsTags`, as well as a separate `deprecation` metadata item. Since `futureBreaking` is a custom tag, it only shows in `docsTags`, which is fine / usable. 

Thanks for that explanation.

macandcheese:
To preview the output, running the build command will generate the updated file located in the ‘/dist/extras’ folder.

geospatialem:
Thanks for investigating, @macandcheese. The above solution makes sense, and we can use it for future use cases. Closing out this PR.

Issue #6272, opened 2023-01-12
[Icon] Some recently added, special case icons not available in dist


macandcheese:
### Actual Behavior

Some recently added icons with special cases (from what I gather, names that have `-f` or start with a number: https://github.com/Esri/calcite-components/blob/efd08f11effbd109a334e36ce817658966422a69/src/components/icon/utils.ts#L64) are not available in when used with CC / CC React. In a local development environment, some do seem to work, but still not all are available:
<img width="1725" alt="Screen Shot 2023-01-11 at 4 00 07 PM" src="https://user-images.githubusercontent.com/4733155/211943911-86cab4dd-9575-4e12-949e-4c136ed4acad.png">
<img width="1722" alt="Screen Shot 2023-01-11 at 3 57 22 PM" src="https://user-images.githubusercontent.com/4733155/211943913-6fe64688-60e0-42c3-9f78-4f9ec2ab23ac.png">
<img width="750" alt="Screen Shot 2023-01-11 at 4 05 21 PM" src="https://user-images.githubusercontent.com/4733155/211944457-f5378e66-ff29-4f1a-9a01-b1547fccdf6e.png">

### Expected Behavior

All icons are available for use, within `calcite-icon` component and when passed as property value to other components, like `combobox-item`

Codepen: http://localhost:3363/demos/icon.html
CodeSandbox: https://codesandbox.io/s/funny-jackson-tb8ufc?file=/src/App.js

### Reproduction Sample

Above

### Reproduction Steps

Visit either reproduction link, see missing icons.

### Reproduction Version

Latest

### Esri team

Calcite (design)

jcfranco:
I think this partly a `calcite-ui-icon` issue. This is what I'm noticing:

1. `bookmark-f` – the convention for filled icons has been [`<iconName>-<size>(-f)`](https://github.com/Esri/calcite-ui-icons/commit/87423ee157e9107990c34e839caeceb86197d534), but for this one [it's `<iconName>(-f)-<size>`](https://github.com/Esri/calcite-ui-icons/commit/1aef262233f162c0b294d34d21effcc23a5719a4). This is causing the utility to not build the correct file name.
2. `1-2x`, `8x`, `1-4x`, and similar ones – these have no matching data path files ([SVGs start with 1-2x](https://github.com/Esri/calcite-ui-icons/tree/master/icons), [JSON starts with aZ16](https://github.com/Esri/calcite-ui-icons/tree/master/js)).
[...]

PR #6422, opened 2023-02-04
fix(icon): fix icon normalization to handle x-times-named icons

jcfranco:
**Related Issue:** #6272

## Summary

This fixes an issue where our icon names weren't mapped correctly to some newly added icons (e.g., `1x`, `1-8x`).

### DIFF
--- a/src/components/icon/utils.ts
+++ b/src/components/icon/utils.ts
@@ -58,24 +58,32 @@ export async function fetchIcon({ icon, scale }: FetchIconProps): Promise<Calcit
  * Normalize the icon name to match the path data module exports.
  * Exported for testing purposes.
  *
- * @param name
+ * @param name – an icon name that can be either kebab or camel-cased
  * @private
  */
 export function normalizeIconName(name: string): string {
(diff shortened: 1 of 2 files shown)


src/components/icon/utils.ts
```diff
@@ -58,24 +58,32 @@ export async function fetchIcon({ icon, scale }: FetchIconProps): Promise<Calcit
...
+  const kebabCased = parts.length > 0;
 
-  if (parts.length === 1) {
-    return numberLeadingName ? `i${name}` : name;
-  }
+  if (kebabCased) {
+    const firstNonDigitInPartPattern = /[a-z]/i;
+
+    name = name
+      .split("-")
```
benelan, line 73: We also already split `name` into `parts` on line 66, so we can use that here instead of splitting again?

benelan (approved):

src/components/icon/utils.ts
jcfranco: Doh! Excellent catch! Fixing!

Issue #6376, opened 2023-01-30
[Tree-item] No focus present when navigating via keyboard


rpanichakit9541:
### Summary

When using keyboard navigation, the `tree` and `tree-item` components do not have a focus present.

### Actual Behavior

No focus present on the components.

### Expected Behavior

A focus should be present on the `tree` and `tree-item` components.

### Reproduction Sample

https://codepen.io/rpanichakit/pen/wvxEJvG

### Reproduction Steps

Use your keyboard to navigate through the calcite-tree-item, and notice that there's no focus present.

### Reproduction Version

1.0.3

### Regression?

beta.99

### Esri team

ArcGIS Online

geospatialem:
@rpanichakit9541 Thanks for reaching out! Accessibility for tree structures follows the [tree role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tree_role), which would expand/collapse `tree-item`s with the `Arrow right`/`Arrow left` keys respectively. We've documented a similar keyboard navigation in the components [keyboard navigation](https://developers.arcgis.com/calcite-design-system/components/tree/#keyboard-navigation).

However, there does seem to be a different with the component's focus, which occurred somewhere between `beta.99` and `1.0.3`. @jcfranco This might be a good one to tackle either in the upcoming patch of February release - WDYT?


Example behavior with focus below:

### beta.99
![beta 99-tree](https://user-images.githubusercontent.com/5023024/215594340-738300a3-be3a-4b4a-87a7-8e467f7f016f.gif)

### 1.0.3
![1 0 3-tree](https://user-images.githubusercontent.com/5023024/215594689-e3f4eafa-64ec-4761-b2ce-bc1ecf367ec1.gif)

rpanichakit9541:
@geospatialem - Thank you for the quick reply.  I apologize I missed that keyboard navigation section.  Feel free to reuse this for the focus issue or close it.

geospatialem:
Updated the issue labels, as the focus functionality works as-expected in Firefox, but does not display in Chrome or Microsoft Edge. Identified when @Elijbet was working through a fix.

geospatialem:
Verified in Storybook with Chrome, Edge, and Firefox.

PR #6424, opened 2023-02-06
fix(tree-item): reverses regression to bring back focus when navigating with keyboard

Elijbet:
**Related Issue:** #6376

## Summary
Reverses regression to bring back focus when navigating with the keyboard and a small cleanup on the existing code.

### DIFF
--- a/src/components/tree-item/tree-item.scss
+++ b/src/components/tree-item/tree-item.scss
@@ -102,10 +102,10 @@
 // focus styles
 :host {
   @apply focus-base;
-  &:focus,
-  &:active {
-    @apply focus-inset outline-none;
-  }
+}
+
+:host(:focus:not([disabled])) {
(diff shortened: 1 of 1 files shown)


src/components/tree-item/tree-item.scss
```diff
@@ -102,10 +102,10 @@
 // focus styles
 :host {
   @apply focus-base;
-  &:focus,
-  &:active {
-    @apply focus-inset outline-none;
-  }
+}
+
+:host(:focus:not([disabled])) {
```
jcfranco, line 107: Are focus styles applied when the item is disabled? IIRC, for that case we block pointer events, prevent tabbing and will blur the item if it becomes disabled and already had focus.

jcfranco (approved):
Other than a question that could lead to simplification, this LGTM!

Issue #6427, opened 2023-02-06
[Segmented Control] Update `calciteSegmentedControlChange` event listener with updated event 


geospatialem:
### Description

Update event context of `calciteSegmentedControlChange` with the new pattern of `event.target`.

Came up on a Community question, with the following context:

> the calciteSegmentedControlChange has no `detail` attribute. The documentation states that it should be the value of the selected item (this was the behaviour of the previous radio group component)

### Which Component

segemented-control

### Resources

https://community.esri.com/t5/calcite-design-system-questions/calcitesegmentedcontrolchange-has-no-detail/m-p/1255474#M319

geospatialem:
Verified with the `1.0.4` release. Documentation page content will update with next Wednesday's documentation release for `1.0.5`.

PR #6428, opened 2023-02-06
docs(segmented-control): update event description for calciteSegmentedControlChange

geospatialem:
**Related Issue:** #6427 

## Summary
Adds context to the event description and removes the `event.detail` reference, where the payload is now set by `event.target`.


benelan (approved):
🧹

Issue #4232, opened 2022-03-14
No `change` event if input cleared after setting value programmatically with focus


nwhittaker:
### Actual Behavior

If an empty `calcite-input` element has focus and its value is set programmatically, it will not dispatch a `calciteInputChange` event if it's manually cleared and blurred.

### Expected Behavior

A `calciteInputChange` event is dispatched because the value was changed due to a user interaction.

### Reproduction Sample

https://ember-twiddle.com/84ec2591fdac5ae61ae10e6dc804d632/3e2cba5bcd55a0fb71c82258be68d252a4fea18e

### Reproduction Steps

1. Visit the sample and open the browser's dev tools console
2. Click the "Seed value with focus shift" button
3. After focus shifts to the input, clear the contents of the input, and then blur the input in some way
4. Observe that "didChangeSeededInput" does not appear in the console indicating that the input's `calciteInputChange` handler was not run

### Reproduction Version

1.0.0-next.415

### Relevant Info

A follow-on to https://github.com/Esri/calcite-components/issues/4229.

jcfranco:
Non-Ember repro case: https://codepen.io/jcfranco/pen/yLqrPMp

geospatialem:
Migrating to the upcoming February milestone for the fix to land later this month.

geospatialem:
Verified in `1.0.5`

![image](https://user-images.githubusercontent.com/5023024/217850846-2e928f95-d50a-439b-8616-16cf2902dc8b.png)

PR #6431, opened 2023-02-07
fix(input, input-number, input-text): emit change value when clearing programmatically-set value

jcfranco:
**Related Issue:** #4232

## Summary

Updates value-setting code to treat programmatically-set values as the last emitted one, which is used to determine whether to emit or not.

### DIFF
--- a/src/components/input/input.tsx
+++ b/src/components/input/input.tsx
@@ -681,11 +681,7 @@ export class Input
       return;
     }
     if (event.key === "Enter") {
-      this.setValue({
-        committing: true,
-        origin: "user",
-        value: this.value
-      });
+      this.emitChangeIfUserModified();
     }
   };
 
@@ -749,11 +745,7 @@ export class Input
     const isShiftTabEvent = event.shiftKey && event.key === "Tab";
     if (supportedKeys.includes(event.key) && (!event.shiftKey || isShiftTabEvent)) {
       if (event.key === "Enter") {
-        this.setValue({
-          committing: true,
-          origin: "user",
-          value: this.value
-        });
+        this.emitChangeIfUserModified();
       }
       return;
     }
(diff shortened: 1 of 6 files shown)


src/components/input/input.tsx
```diff
@@ -898,22 +899,16 @@ export class Input
     this[`child${this.type === "number" ? "Number" : ""}El`].value = newInputValue;
   };
 
-  private setPreviousEmittedValue = (newPreviousEmittedValue: string): void => {
-    this.previousEmittedValue =
-      this.type === "number"
-        ? isValidNumber(newPreviousEmittedValue)
-          ? newPreviousEmittedValue
-          : ""
-        : newPreviousEmittedValue;
```
eriklharper, line 907: Wow, didn't realize there was still number logic in here!

src/components/input/input.tsx
```diff
@@ -929,6 +924,9 @@ export class Input
     previousValue?: string;
     value: string;
   }): void => {
+    this.setPreviousValue(previousValue ?? this.value);
```
eriklharper, line 927: Great call using the `??` operator here, that makes so much more sense.

src/components/input/input.tsx
```diff
@@ -955,19 +953,18 @@ export class Input
...
-      this.setPreviousValue(previousValue || this.value);
-      this.previousValueOrigin = origin;
       this.userChangedValue = origin === "user" && this.value !== value;
       this.value = value;
-      origin === "direct" && this.setInputValue(value);
+    }
+
+    if (origin === "direct") {
+      this.setInputValue(value);
+      this.previousEmittedValue = value;
```
eriklharper, line 967: Setting previousEmittedValue here is also crucial to do in the direct case.  Glad to see the `origin` pattern being put to good use!

eriklharper (approved):
Tested all of this thoroughly, and its working flawlessly!  Thanks for tackling this :)

Issue #6281, opened 2023-01-13
Modal loses focus trap after clicking inside


nwhittaker:
### Actual Behavior

Clicking inside a modal causes the modal's focus trap to become deactivated.

### Expected Behavior

Clicking inside a modal does not deactivate the modal's focus trap.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/JjBJYee

### Reproduction Steps

1. Visit the sample and tab around to see the focus trap is in place.
2. Click on the modal's header, its content, or on any of its buttons (slotted or otherwise).
3. Tab around and see focus now moves outside of the modal.

### Reproduction Version

next.721

### Impact

A very high adverse impact to accessibility.

### Esri team

ArcGIS Field Apps

jcfranco:
Preliminary findings on this one:

* `focus-trap` is [configured to disable the trap when clicking outside](https://github.com/Esri/calcite-components/blob/master/src/utils/focusTrapComponent.ts#L50).
* By default, `clickOutsideDeactivates: true` is deactivating the trap when clicking inside the modal
* Using a custom `clickOutsideDeactivates` handler (deactivating if the click target is outside of the modal trap) helps fix this issue, but is now causing another issue where clicking on the focusable items within the modal (buttons in the example case) aren't consistently focused. **This is the main blocker at the moment.**
* [`document` is set to the trap's owner document](https://github.com/Esri/calcite-components/blob/master/src/utils/focusTrapComponent.ts#L51) instead of the component's shadow host (as [recommended in the doc](https://github.com/focus-trap/focus-trap#user-content-shadow-dom)). 
* The tab index for the fallback focus element is set to -1 [via prop](https://github.com/Esri/calcite-components/blob/master/src/utils/focusTrapComponent.ts#L45-L47) instead of attribute (as [recommended in the doc](https://github.com/focus-trap/focus-trap#your-trap-should-include-a-tabbable-element-or-a-focusable-container)). This doesn't reflect and could potentially mess up anything querying by the attribute.

At this point, it's unclear to me whether this is a limitation of `focus-trap` + shadow DOM, a bug or something we're missing in our `focus-trap` options. We'll have to dig a bit more.

geospatialem:
Migrating to the upcoming February milestone for the fix to land later this month.

geospatialem:
Verified in `1.0.5`

[...]

PR #6434, opened 2023-02-07
fix(modal): no longer loses focus trap after clicking inside the component.

anveshmekala:
**Related Issue:** #6281 

## Summary
This PR will fix `focusTrap` issue when user clicks inside the `modal`. When user clicks inside the modal, `focusTrap` behavior wont be affected.

### DIFF
--- a/src/components/popover/popover.e2e.ts
+++ b/src/components/popover/popover.e2e.ts
@@ -735,8 +735,8 @@ describe("calcite-popover", () => {
       }));
   });
 
-  // refer to https://github.com/Esri/calcite-components/issues/5993 for context
   it("should focus input element in the page with popover when user click", async () => {
+    // refer to https://github.com/Esri/calcite-components/issues/5993 for context
     const page = await newE2EPage();
     await page.setContent(html` <calcite-shell content-behind>
         <calcite-shell-panel slot="panel-end">
--- a/src/components/popover/popover.e2e.ts
+++ b/src/components/popover/popover.e2e.ts
@@ -734,4 +734,48 @@ describe("calcite-popover", () => {
         shadowFocusTargetSelector: `.${CSS.closeButton}`
       }));
   });
+
+  // refer to https://github.com/Esri/calcite-components/issues/5993 for context
+  it("should focus input element in the page with popover when user click", async () => {
+    const page = await newE2EPage();
+    await page.setContent(html` <calcite-shell content-behind>
+        <calcite-shell-panel slot="panel-end">
+          <calcite-label>
+            value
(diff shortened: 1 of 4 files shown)


src/components/popover/popover.e2e.ts
```diff
@@ -734,4 +734,48 @@ describe("calcite-popover", () => {
         shadowFocusTargetSelector: `.${CSS.closeButton}`
       }));
   });
+
+  // refer to https://github.com/Esri/calcite-components/issues/5993 for context
```
anveshmekala, line 738: added this to provide context.

src/utils/focusTrapComponent.ts
```diff
@@ -42,12 +42,15 @@ export function connectFocusTrap(component: FocusTrapComponent): void {
     return;
   }
 
-  if (focusTrapEl.tabIndex == null) {
-    focusTrapEl.tabIndex = -1;
+  if (focusTrapEl.tabIndex === null) {
```
jcfranco, line 45: This `if` block won't be executed as `tabIndex` is a number and won't ever be null.

src/utils/focusTrapComponent.ts
```diff
@@ -42,12 +42,15 @@ export function connectFocusTrap(component: FocusTrapComponent): void {
...
+    focusTrapEl.setAttribute("tabindex", "-1");
   }
 
   const focusTrapOptions: FocusTrapOptions = {
-    clickOutsideDeactivates: true,
+    clickOutsideDeactivates: (event: MouseEvent | TouchEvent) => {
+      return !event.composedPath().find((el) => (el as HTMLElement) === focusTrapEl);
+    },
+    allowOutsideClick: true,
```
jcfranco, line 53: Nitpick: can you sort the props alphabetically? Makes it easier to know where they go and to look up.

src/components/popover/popover.e2e.ts
jcfranco: Nice! WDYT about moving the comment inside the test, so if for whatever reason it's modified or moved, it goes with it?

jcfranco (approved):
Aside from updating the if statement to set `tabindex="-1"` when needed, this LGTM!

src/components/popover/popover.e2e.ts
anveshmekala: make sense.

src/utils/focusTrapComponent.ts
jcfranco: When no `tabIndex` is defined, the value will be `-1` if it's a non-focusable element by default. Otherwise, it will be `0`.

PR #6437, opened 2023-02-08
test(tabs): delay story due to potential timing chromatic diff

benelan:
**Related Issue:** #5284

## Summary
Adds a delay to the story because it registered a chromatic diff in what seems like a timing issue
https://www.chromatic.com/test?appId=6266d45509d7eb004aa254fb&id=63e30231eac9ea4af6de885d


src/components/tabs/tabs.stories.ts
```diff
@@ -174,3 +174,6 @@ export const TabChilrenWithPercentageHeights = (): string => html`
     </calcite-tab>
   </calcite-tabs>
 `;
+TabChilrenWithPercentageHeights.parameters = {
+  chromatic: { delay: 1000 }
```
jcfranco, line 178: Sidebar: I think we could extract this delay in a helper module for reuse.

jcfranco (approved):

src/components/tabs/tabs.stories.ts
benelan: Yeah, I have it sprinkled around on other stories too. The only thing is I'd rather not delay all stories, only the ones we need to

src/components/tabs/tabs.stories.ts
```diff
@@ -174,3 +174,6 @@ export const TabChilrenWithPercentageHeights = (): string => html`
     </calcite-tab>
   </calcite-tabs>
 `;
+TabChilrenWithPercentageHeights.parameters = {
```
eriklharper, line 177: Should this be spelled `Children`? ☝️

eriklharper:
Issue https://github.com/Esri/calcite-components/issues/5284 doesn't seem related to this PR.  Is that intentional?

src/components/tabs/tabs.stories.ts
jcfranco: `Chilrun` 😂 Good catch.

src/components/tabs/tabs.stories.ts
jcfranco: I meant extracting the delay value, so we would import the same reference.

benelan: Gotcha. I've been putting different delay values at this point to see what works without eating too much time. But once I figure that out I'll create a const

PR #6476, opened 2023-02-15
ci(action): fix edge case where '`' causes breaks in webhook

alisonailea:


jcfranco (approved):

.github/workflows/sync_to_airtable.yml
```diff
@@ -31,11 +31,12 @@ jobs:
         with:
           script: |
             const { title, body, assignees, closed_at, created_at, id, labels, milestone, number, state, updated_at } = JSON.parse(process.env.ISSUE);
+            const bodyData = JSON.stringify(body);
```
jcfranco, line 34: Suggestion: I think you could specify the `replacer` here.

Issue #6454, opened 2023-02-10
Inputs not always focusable in Modal


ellenupp:
### Actual Behavior
Inputs used inside calcite-modal are not always focusable, making them non-functional. Might be related to when they are not initially visible, like when used in a tab.

### Expected Behavior
Inputs should always be focusable when not disabled.

### Reproduction Sample
https://codepen.io/ellupp/pen/BaPgVQV?editors=1000

### Reproduction Steps
click button to open the modal
click on 'tab 2'
try to click on the input. you cannot focus on the input.
Reproduction Version
1.0.5

### Relevant Info
No response

### Regression?
1.0.31.0.4

### Impact
This is a blocking issue as it prohibits many editing workflows

### Esri team
ArcGIS Dashboards

mpayson:
We're seeing this too with the item properties modal in our app and the inputs are all initially visible

https://user-images.githubusercontent.com/5892789/218186663-77457b0a-067c-4df0-8841-d1512bb61bf0.mov

ellenupp:
cc @vijayendranj

ellenupp:
Hi @alisonailea, why did you edit this issue to remove all the details?

paulcpederson:
Impacts many teams. As a workaround we've found we can call `updateFocusTrapElements` on the modal after the new content is available, but hard to know when to time this. 

Very much seems to be related to https://github.com/Esri/calcite-components/issues/6281

geospatialem:
> Hi @alisonailea, why did you edit this issue to remove all the details?

@ellenupp Thanks for pinging us, there were some errors in automation. Added in the Dashboards label, so should be set now. cc @alisonailea

jcfranco:
@mpayson @paulcpederson Can you confirm if the original repro case represents your use cases? If not, could you share a repro case? 

@geospatialem We should look into this for an upcoming patch because of its impact.

jcfranco:
@paulcpederson That'd be very helpful. Thanks!

jcfranco:
Quick update: I've got a fix in the works. PR coming up!

jcfranco:
@geospatialem @benelan and myself are stepping in to help verify this one.

jcfranco:
Verified locally by testing original repro case in:

* OSX: Chrome (110.0.5481.100), Firefox (109, 110), Edge (110.0.1587.41), Safari (Version 16.2)
[...]

PR #6479, opened 2023-02-15
fix(focus-trap): prevent host from receiving initial focus

jcfranco:
**Related Issue:** #6454 

## Summary

This fixes an issue caused by setting the `tabindex` attribute on the host element instead of the property ([previous behavior](https://github.com/Esri/calcite-components/pull/6434)). 

### Notes

* When `tabindex` was set, `focus-trap` seemed to move focus to the host as focus fallback even if focus was already set within the focus trap. We should look into only setting the host as focus fallback if there is no focusable content within the trap.
* The `tabindex` setting logic was removed instead of being rolled back because the [previous version](https://github.com/Esri/calcite-components/blob/059d6eee6a8dc0a0d74db6e113d30cafebac25bb/src/utils/focusTrapComponent.ts#L33-L35) was misleading and would never be invoked (`HTMLElement.tabIndex` [is always an integer](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/tabIndex#value)).
* Some focus-related tests were fixed as they were not properly set up and produced false positives. 
* A test util was added to help match elements against the focused element.

### DIFF
--- a/src/components/modal/modal.e2e.ts
+++ b/src/components/modal/modal.e2e.ts
@@ -1,8 +1,8 @@
-import { newE2EPage } from "@stencil/core/testing";
+import { E2EPage, newE2EPage } from "@stencil/core/testing";
 import { focusable, renders, slots, hidden, t9n } from "../../tests/commonTests";
 import { html } from "../../../support/formatting";
 import { CSS, SLOTS, DURATIONS } from "./resources";
-import { newProgrammaticE2EPage, skipAnimations } from "../../tests/utils";
+import { isElementFocused, newProgrammaticE2EPage, skipAnimations, waitForAnimationFrame } from "../../tests/utils";
 
 describe("calcite-modal properties", () => {
   it("renders", () => renders("calcite-modal", { display: "flex", visible: false }));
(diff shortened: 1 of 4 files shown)


src/components/modal/modal.e2e.ts
```diff
@@ -303,83 +303,90 @@ describe("opening and closing behavior", () => {
...
     await page.waitForChanges();
     await modal.setProperty("open", false);
     await page.waitForChanges();
-    expect(document.activeElement).toEqual($button);
+    expect(await isElementFocused(page, initiallyFocusedIdSelector)).toBe(true);
   });
 
   it("traps focus within the modal when open and disabled close button", async () => {
+    const button1Id = "button1";
+    const button2Id = "button2";
```
benelan, line 363: Nitpick: we can probably move these consts outside of the `it` scope at the top of the file so we don't have to declare them more than once

benelan (approved):
:+1:

src/components/modal/modal.e2e.ts
jcfranco: I'll do this in a follow-up PR.

PR #6483, opened 2023-02-15
revert(focus-trap): drop focus-trap options causing inconsistent focus behavior

jcfranco:
**Related Issue:** #6281 #6454

## Summary

Reverts 764609de7b8c96aa331e364ca790eefdb44dd1ab and df144dc30e66d7e11fda326f289c6b8c931c34f8 to give us more time to look into other scenarios where focus is being blocked after updating focus-trap configuration.


anveshmekala (approved):
LGTM!

PR #6488, opened 2023-02-15
chore(linting): restore pattern to lint TS and TSX files

jcfranco:
**Related Issue:** N/A

## Summary

Restores TS(X) file linting on commit.


benelan (approved):
Just out of curiousity, why do we want to lint the assets?

jcfranco:
> Just out of curiousity, why do we want to lint the assets?

I mainly made the change to restore linting on TS files, but I'm guessing we want to lint any _staged_ files regardless of where they reside.

PR #6489, opened 2023-02-15
chore(linting): drop filter for JS file linting

jcfranco:
**Related Issue:** N/A

## Summary

Drops an obsolete filter since we no longer have JS files in the codebase.


benelan (approved):
Can you also remove the js,jsx file extensions in `lint:ts`

https://github.com/Esri/calcite-components/blob/cad286dd84163d61aae15c55dac3c01808aea221/package.json#L34

Issue #4800, opened 2022-06-23
[Modal] Provide sticky header / footer slots adjacent to content slot


macandcheese:
### Description

A common request we get is to provide styling / css for a sticky portion of content above the content slot in a modal. Similarly this could be applicable to a bottom content section.

Because there are internally rendered / shadow DOM divs, it's tricky for a consumer to create this styling on their own. A proposal can be made to add sticky header / footer slots to a modal that provide this with built in styles.

### Acceptance Criteria

Sticky top / bottom slot/s are added to the modal to provide this styling.

### Which Component

Modal

### Example Use Case

Users often want to provide a bit of contextual information (besides the header) above scrolling content within a modal. This is often the case when tabs / a data grid is used in the content slot.

shawnmgoulet:
Here's a practical example:

![image](https://user-images.githubusercontent.com/3179974/177357812-3eead10d-3278-45e1-8453-5b1285f1dd30.png)

geospatialem:
The design for the enhancement is in the works, and we plan to address in the February release.

geospatialem:
@ashetland @SkyeSeitz - Dev is looking to address this in the February release, are there any design changes or recommendations needed before proceeding?

ashetland:
Final spec available in Figma. Sending link to @Elijbet.

geospatialem:
Verified on [Storybook](https://esri.github.io/calcite-components/?path=/story/components-modal--slots).

![image](https://user-images.githubusercontent.com/5023024/225368648-a74db035-1b7c-4428-9b9b-8577d172267d.png)

PR #6490, opened 2023-02-16
feat(modal): provides `content-top` and `content-bottom` slots

Elijbet:
**Related Issue:** #4800

## Summary
Provides `content-top` and `content-bottom` slots. 

For simplicity, we matched the `padding` on the `content-top` and `content-bottom` to the `content`, with one public prop for the latter. This can further be developed if requested.

We decided to move forward with `content-top` and `content-bottom` naming, as this should provide enough context and also aligns with existing `<x>-start/<x>-end` naming.


src/components/modal/modal.scss
```diff
@@ -5,6 +5,8 @@
  *
  * @prop --calcite-modal-content-background: Specifies the background color of content placed in the `content` slot.
  * @prop --calcite-modal-content-padding: Specifies the padding of the modal `content` slot.
+ * @prop --calcite-modal-content-header-padding: Specifies the padding of the modal `content-header` slot.
```
macandcheese, line 8: cc @ashetland / @SkyeSeitz - slot names could be `content-top` and `content-bottom` to avoid confusion with existing elements.

ashetland: Would `content-start` and `content-end` be more inline with existing conventions or do we reserve start/end for slots on the x-axis?

Elijbet: What do you mean by this?

ashetland: We use start/end in multiple places, for example icon-start/end and actions-end (I think we have actions-start somewhere?). So instead of naming these slots `content-top` and `content-bottom` would `content-start` and `content-end` make sense? Or would top/bottom be preferred here?

SkyeSeitz: I don't think it would hurt to reserve start & end with the x-axis in case we ever needed both top/bottom & start/end, so `content-top` & `content-bottom` would be my vote unless there were additional concerns.

src/components/modal/modal.scss
```diff
@@ -184,13 +189,44 @@
...
+  @apply bg-foreground-1
+  border-color-3
+  flex
+  min-w-0
+  max-w-full
+  rounded-t
+  border-0
+  border-b
+  border-solid
+  z-header;
```
jcfranco, line 207: Is `z-header` needed for the top and bottom areas?

src/components/modal/modal.scss
```diff
@@ -184,13 +189,44 @@
...
+  padding: var(--calcite-modal-content-padding, var(--calcite-modal-padding-internal));
+}
+
+.content-top {
+  @apply bg-foreground-1
+  border-color-3
+  flex
+  min-w-0
+  max-w-full
+  rounded-t
```
jcfranco, line 203: Is rounding needed for when there's no header nor footer?

src/components/modal/modal.scss
```diff
@@ -184,13 +189,44 @@
...
+  border-0
+  border-b
+  border-solid
+  z-header;
+  flex: 0 0 auto;
+  padding: var(--calcite-modal-padding-internal);
+}
+
+.content-bottom {
```
jcfranco, line 212: Can you DRY up the common utils from both `.content-top` and `.content-bottom`?

src/components/modal/modal.scss
```diff
@@ -184,13 +189,44 @@
...
+    border-0
+    border-t
+    border-solid
+    z-header;
+  flex: 0 0 auto;
+  padding: var(--calcite-modal-padding-internal);
 }
 
-.content--no-footer {
+.content--no-modal-footer {
```
jcfranco, line 229: Since the slots were renamed, can this go back to the previous name?

This applies to other modal-footer renames in this PR.

src/components/modal/modal.tsx
```diff
@@ -255,6 +264,22 @@ export class Modal
     ) : null;
   }
 
+  renderContentTop(): VNode {
+    return (
+      <div class={CSS.contentTop} hidden={!this.hasContentTop} key="content-top">
```
jcfranco, line 269: It looks like the content top and bottom nodes are always rendered, so these don't need `key` (see [Stencil doc on conditional rendering](https://stenciljs.com/docs/templating-jsx#:~:text=If%20this%20array%20of%20children%20is%20dynamic%2C%20i.e.%2C%20if%20any%20nodes%20may%20be%20added%2C%20removed%2C%20or%20reordered%2C%20then%20it%27s%20a%20good%20idea%20to%20set%20a%20unique%20key%20attribute%20on%20each%20element)).

jcfranco (approved):

jcfranco:
@Elijbet Let's wait until the patch release is done before merging this.

src/components/modal/modal.scss
Elijbet: It's on the same level as the header, so I kept it.

src/components/modal/modal.scss
Elijbet: I made a correction here to have the top rounded on the `content-top` if there is no `header`, and the rounded bottom on the `content-bottom` if there is no `footer`.

PR #6491, opened 2023-02-16
ci: add Dockerfile for running demos or e2e tests

benelan:
**Related Issue:** https://github.com/ionic-team/stencil/issues/3853

## Summary

I went down a bit of a rabbit hole on this one. I'm honestly not sure why our tests aren't working on Ubuntu 22.04 (more info in the comments of the issue linked above). I can't repro with stencil's starter code, and after an hourish of adding stuff from CC to the starter I still couldn't repro.

 I want to be able to run tests locally so I ended up creating a Dockerfile for it, which seems to be working. There are inconsistent timeouts, which I'll look into when I have more time. But I've seen them outside of a container in the past, so Docker may be aggrevating an existing issue. 

I added some npm scripts to make the Docker image easier to use. The most relevant two - `npm run docker:run-start` and `npm run docker:run-test-watch` - correctly watch for code changes in the local repo and update the container accordingly.

I'm not sure if this is even something we want committed to the repo or not, but I wanted to at least create a PR so I could point to it if anyone else runs into this issue. @jcfranco thoughts?

Alternatively I could add instructions and inline the Dockerfile into one of our convention docs or a wiki.


jcfranco:
For posterity, we decided to move this to a wiki as a resource while we flesh out Docker usage in our dev environment.

benelan:
Note: I reopened and closed the issue because I fixed some errors and the changes didn't show up in the files tab. I linked the PR in the Stencil issue so others could find the work around, and I wanted to make sure the correct version is easy to find.

For reference, I removed some user permissions stuff from the Dockerfile when I was cleaning up and it continued to work due to [Docker's caching](https://docs.docker.com/build/cache/). Once I cleared the cache the errors popped up again.

Issue #6248, opened 2023-01-06
calcite-block: handle label not translated


AdelheidF:
### Actual Behavior
![image](https://user-images.githubusercontent.com/3362490/219098686-86cc7477-43b7-4d5e-b72f-db9812087829.png)

### Expected Behavior
handle should be translated

### Reproduction Sample
https://codepen.io/afreitag/pen/RwBbgxK

### Reproduction Steps
Hover over the drag handle and you'll see the English text "handle". Hover over the expand/collapse icon an you'll see French text.

### Reproduction Version
beta.99

### Relevant Info
No response

### Regression?
don't think so

### Impact
medium/high

### Esri team
ArcGIS Map Viewer

geospatialem:
Verified on `master`.

### `en`:
![image](https://user-images.githubusercontent.com/5023024/225348157-f169f1cf-a3ef-4dd4-948e-2a453c83f9c9.png)

### `ar`:
![image](https://user-images.githubusercontent.com/5023024/225348401-c5ef9252-f298-4031-942e-70d3b035f46a.png)

PR #6493, opened 2023-02-16
chore(handle): add message bundles for translation.

anveshmekala:
**Related Issue:** #6248 

## Summary

This PR will add message bundles required for translation build.

### DIFF
--- a/src/components/handle/assets/handle/t9n/messages.json
+++ b/src/components/handle/assets/handle/t9n/messages.json
@@ -1,3 +1,3 @@
 {
-  "title": "handle"
+  "dragHandle": "Drag handle"
 }
--- a/src/components/handle/assets/handle/t9n/messages.json
+++ b/src/components/handle/assets/handle/t9n/messages.json
@@ -1,0 +1,3 @@
+{
+  "title": "handle"
+}
--- a/src/components/handle/assets/handle/t9n/messages_en.json
+++ b/src/components/handle/assets/handle/t9n/messages_en.json
@@ -1,3 +1,3 @@
 {
-  "title": "handle"
+  "dragHandle": "Drag handle"
 }
--- a/src/components/handle/assets/handle/t9n/messages_en.json
+++ b/src/components/handle/assets/handle/t9n/messages_en.json
@@ -1,0 +1,3 @@
+{
+  "title": "handle"
+}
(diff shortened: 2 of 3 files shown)


src/components/handle/assets/handle/t9n/messages.json
```diff
@@ -0,0 +1,3 @@
+{
+  "title": "handle"
```
anveshmekala, line 2: wondering if `dragHandle` makes more sense here. Thoughts @jcfranco , @geospatialem ?

jcfranco: Agreed and now that you bring it up, using it as the value may also be more helpful. I'll defer to @geospatialem on that one.

jcfranco (approved):
🎉

src/components/handle/assets/handle/t9n/messages.json
geospatialem: Is the context for when users are moving items? And if so, would it override what we have in place for [`value-list`](https://github.com/Esri/calcite-components/blob/master/src/components/value-list/assets/value-list/t9n/messages.json#L1-L6)? 

If its separate from `value-list`, and doesn't override those strings, dig the `dragHandle` suggestion, perhaps could also update the value? 

WDYT about:

> "dragHandle": "Drag handle"

anveshmekala: This is separate from `value-list` in terms of implementation  which doesn't override the instruction we have in place for `value-list`.  Since you bought it up, i wonder if `sortable-list` and block should have the same a11y support for dragEnabled as value-list 🤔 ?

anveshmekala: > WDYT about:
> 
> > "dragHandle": "Drag handle"

I prefer using `dragHandle`  both for value and key and since it is internal component wont be a breaking change and dragHandle is more self explanatory  than handle .

geospatialem: > I wonder if `sortable-list` and `block` should have the same a11y support for `dragEnabled` as `value-list` considering the functionality is similar 🤔

Oh, good call. Yeah, would imagine we should add for inclusion as well.

Issue #5627, opened 2022-10-25
Label for Select and Slider (Form) - Provide a valid label for form fields - (2036570477)


dqateam:
### Module: 
 40 Form 

### Violation: 
 Provide a valid label for form fields

![image](https://user-images.githubusercontent.com/5023024/202567060-8abe9d2a-546e-444e-ab4e-7cc3b79054fc.png) 

### WCAG Reference: 
- [1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html)
- [3.3.2 Labels or Instructions](https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html)
- [4.1.2 Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html) 

### Instance ID: 
 2036570477 

### Severity: 
 10 

### Description: 
 [Issue]
There are form controls with visual labels that are not programmatically associated with the control. Examples include:

- Where did you hear about Calcite components?
- How many components will you be using?
- What is your favorite component(s)?

[User Impact]
Screen reader users will have difficulty determining the purpose of these controls. Speech input users will have difficulty navigating to them.

[Code Reference] 
 ```
<select class="select">
  <option label="Please selectâ¦" value="">Please selectâ¦</option>
  <option label="Esri DevSummit" value="">Esri DevSummit</option>
  <option label="Search engine" value="">Search engine</option>
  (...)
</select>
 ``` 

### Note: 
 [Suggestion]
Provide a valid label for form fields. The best way to do this is to use a ``` <label> ``` element with a for attribute set to the ID of the field it labels. For custom controls, use aria-labelledby on the control set to the ID of the labelling text.

[Compliant Code Example] 
 ```
<label for="dropdown-label">Where did you hear about Calcite components?</label>
<select id="dropdown-label" class="select">
  <option label="Please selectâ¦" value="">Please selectâ¦</option>
  <option label="Esri DevSummit" value="">Esri DevSummit</option>
  <option label="Search engine" value="">Search engine</option>
  (...)
</select>
 ```
 

### Media Type: 
 Forms 

### Additional Resources: 
- Report Source: Product accessibility evaluation conducted on the Calcite Design System in August, 2022.

geospatialem:
`label` is not added to [the form](https://geospatialem.github.io/calcite-samples/0-form.html) for the following components:
- `select`
- `slider`
- `combobox`

[...]

PR #6500, opened 2023-02-22
fix(select, slider, combobox): display label in screen reader instructions. 

anveshmekala:
**Related Issue:** #5627 

## Summary

This PR will add  valid label to the component with `aria-label` .  User's can wrap the component with `calcite-label` or use `label` prop .


jcfranco (approved):
🎉

jcfranco:
Can you add a bit more context to the PR title about this fix? Technically, some components were already using `aria-label`.

Also, we should look into providing a testing util or pattern for these use cases as it's currently a bit cumbersome to setup. Can you create a follow-up issue for this?

PR #6504, opened 2023-02-23
fix(modal): no longer loses focus trap after clicking inside the component

jcfranco:
**Related Issue:** #6281

## Summary

* restores changes reverted by https://github.com/Esri/calcite-components/pull/6483
* bumps focus-trap dep


jcfranco:
Closing. Will revisit approach in a separate PR.

PR #6505, opened 2023-02-23
chore(linting): wire up px-to-rem linting rule

jcfranco:
**Related Issue:** N/A

## Summary

Sets up linting rule to automatically convert px-to-rem on commit. This should help with sizing when the browser font size is modified.


anveshmekala (approved):
🥇 🥇 🥇

anveshmekala:
Nitpick: can you add PR description.

alisonailea:
To align with this change we should review design tokens as well. Not all platforms use rems but we could convert tokens at SCSS generation to REMs if that is preferable.

jcfranco:
Closing until we confirm this is something we want to pursue.

Issue #6494, opened 2023-02-16
`calcite-modal`'s scale overrides `calcite-input`'s scale


omarkawach:
### Actual Behavior

When calcite-input gets loaded by the lazy loader (the first time), it disregards its own scale and searches the ancestry for a scale.

In the case of a modal set to scale="l", the calcite-input will find the modal and use it for its scale even though the calcite-input was set to scale "m".

Our subsequent calcite-input instances are working correctly in the Arcade editor

### Expected Behavior

The scale of calcite-input should not be overwritten by another component when a scale is already set

```js
<calcite-input
      value={""}
      icon="magnifying-glass"
      clearable
      scale="s"
/>
```

### Reproduction Sample

https://github.com/Esri/calcite-components/blob/cad286dd84163d61aae15c55dac3c01808aea221/src/components/input/input.tsx#L452

### Reproduction Steps

Sharing the problematic line of code. Talk to @jcfranco

Difficult to reproduce in a codepen https://codepen.io/omarkawach/pen/VwGYMmo

### Reproduction Version

1.0.7

### Impact

We had to do a workaround for now

```js
<div scale="m"><calcite-input ... /></div>
```

This way calcite input finds an ancestor with a scale other than the modal's scale.

### Esri team

ArcGIS Online

jcfranco:
This is legacy behavior that snuck into 1.0.0, and can have unintended consequences if component ancestors have `scale` or `status` set. 

Related issues: 

* https://github.com/Esri/calcite-components/issues/3755
* https://github.com/Esri/calcite-components/pull/3984
* https://github.com/Esri/calcite-components/issues/3781

geospatialem:
@omarkawach Wasn't able to repro the original issue in the Codepen to verify if the fix is working now. Could you verify if the Arcade editor is working as-expected with `1.3.0-next.4`?

omarkawach:
Verified the fix in the editor 🙂

I will reopen this issue if the problem comes up again.

geospatialem:
Awesome, thank you @omarkawach! Marking this as verified and closing it out for the upcoming April release.

PR #6506, opened 2023-02-23
fix(inline-editable, input-message, input-number, input-text, input): prevent components from unintentionally picking up a different scale/status value from an ancestor

jcfranco:
**Related Issue:** #6494

## Summary

This PR updates remaining non-children components to no longer look up their ancestry to set props.


joeyHarig:
@jcfranco Wanted to check in and see if there is any status update on this fix. ArcGIS Monitor cannot upgrade to v1.2.0 until issue #6494 is resolved.

joeyHarig:
Thanks for the update @jcfranco!

alisonailea (approved):
👍🏻

PR #6513, opened 2023-02-23
ci: add user-defined priority

geospatialem:
**Related Issue:** n/a

## Summary
1. Adds a dropdown to the bug issue template so users can select a priority impact from a dropdown menu, which includes:
        - `p3 - not time sensitive`
        - `p2 - want for current milestone`
        - `p1 - need for current milestone`
        - `p0 - emergency`

2. Adds an action that automatically adds a label (and will create a label if its not yet created) similar to the `add-esri-product-label.yml`

cc @brittneytewks

### DIFF
--- a/.github/workflows/add-bug-priority-label.yml
+++ b/.github/workflows/add-bug-priority-label.yml
@@ -46,7 +46,7 @@ jobs:
                   repo,
                   name: bugPriority,
                 });
-              } catch (e) {
+              } catch (error) {
                 await github.rest.issues.createLabel({
                   owner,
                   repo,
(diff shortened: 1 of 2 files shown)


.github/workflows/add-bug-priority-label.yml
```diff
@@ -0,0 +1,66 @@
...
+
+            if (bugPriority && bugPriority !== "N/A") {
+              /** Creates a label if it does not exist */
+              try {
+                await github.rest.issues.getLabel({
+                  owner,
+                  repo,
+                  name: bugPriority,
+                });
+              } catch (e) {
```
jcfranco, line 49: Nitpick: the linter won't catch this, but for consistency, the error argument should be `error`.

jcfranco (approved):
💥

Issue #6512, opened 2023-02-23
Tree item content no longer wraps long words


nwhittaker:
### Actual Behavior

Long words are clipped to one line:

<img width="317" alt="Screenshot 2023-02-23 at 1 04 45 PM" src="https://user-images.githubusercontent.com/421496/220992175-22175a36-3c6f-4b69-885b-831dafb454cd.png">

### Expected Behavior

Long words wrap:

<img width="317" alt="Screenshot 2023-02-23 at 1 05 15 PM" src="https://user-images.githubusercontent.com/421496/220992193-ababc126-c918-4f2c-8da7-12da87d82dee.png">

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/LYJZQaO

### Reproduction Steps

1. Visit sample

### Reproduction Version

1.0.7

### Relevant Info

Possibly fixable by adding a `min-width: 0` style to the shadow dom's `.node-container` element.

### Regression?

1.0.5

### Impact

The Field Maps web app uses the tree element to organize layers and nested layers. The tree-item content is the layer's name -- as such it is not impossible to imagine users who have constructed long, non-breaking layer names.

### Esri team

ArcGIS Field Apps

geospatialem:
Verified in `1.0.8-next.3`

![image](https://user-images.githubusercontent.com/5023024/222013972-80123621-1c2a-4264-b18c-ba0a40a7e755.png)

PR #6518, opened 2023-02-24
fix(tree): restore wrapping in tree-item text content

jcfranco:
**Related Issue:** #6512 

## Summary

Updates styles to restore text wrapping. Thanks @nwhittaker for the fix suggestion!


benelan (approved):
:+1:

Issue #6514, opened 2023-02-23
Tree item content now clipped if wrapping beyond two lines


nwhittaker:
### Actual Behavior

Tree item clips content:

<img width="265" alt="Screenshot 2023-02-23 at 1 17 45 PM" src="https://user-images.githubusercontent.com/421496/220995689-c023c516-3a49-49f0-a497-77ddd38d9f66.png">

### Expected Behavior

Tree item hugs content:

<img width="265" alt="Screenshot 2023-02-23 at 1 17 28 PM" src="https://user-images.githubusercontent.com/421496/220996007-14b7b0d8-03c5-4b71-8c34-495eb2209c9c.png">

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/rNZeXVq

### Reproduction Steps

1. See sample

### Reproduction Version

1.0.7

### Relevant Info

Somewhat related to https://github.com/Esri/calcite-components/issues/6512

### Regression?

1.0.5

### Impact

The Field Maps web app uses the tree element to organize layers and nested layers. The tree-item content is the layer's name -- as such it is not impossible to imagine users who have constructed long layer names.

The tree also exists within a resizable panel -- so even short layer names could wrap.

### Esri team

ArcGIS Field Apps

jcfranco:
Installed.

geospatialem:
Verified in `1.0.8-next.3`

![image](https://user-images.githubusercontent.com/5023024/222013805-875c7e5e-b337-4843-8871-c69de46360cf.png)

PR #6519, opened 2023-02-24
fix(tree): prevent tree-item content from being clipped

jcfranco:
**Related Issue:** #6514 

## Summary

This drops any clipping and also ensures the min-height of tree-items adheres to the design spec while still allowing items to grow if needed to accommodate their content.

cc @SkyeSeitz @ashetland


anveshmekala (approved):
LGTM!  👍

Issue #6300, opened 2023-01-17
Floating UI issues with Tree and Dropdown


geospatialem:
### Actual Behavior
The first tree-item has the expected floating UI behavior, but subsequent tree-items placement looks off that surfaced while verifying #5168 cc @anveshmekala

![1__212780237-ea5ab162-0128-45cd-bf3b-34c433228b7f](https://user-images.githubusercontent.com/3362490/219099506-370650ec-d5e7-40f7-9cd1-5d601290edf9.gif)

### Expected Behavior
The Dropdown item should be present in line with its parent Dropdown, similar to the first item.

### Reproduction Sample
https://codepen.io/geospatialem/pen/QWBqWeN

### Reproduction Steps
Open the [Codepen sample](https://codepen.io/geospatialem/pen/QWBqWeN)
Select the first dropdown and observe the dropdown-item's placement (as-expected)
Select either the second or third dropdown and observe the behavior of the dropdown-item's placement
![2__213015711-9d144bcd-b15f-4227-83dc-7e7a6a642c3f](https://user-images.githubusercontent.com/3362490/219099546-d1387162-087a-4b03-98e0-01ce58f5cf9f.png)

### Reproduction Version
next.726

### Relevant Info
Applicable to Chrome and Edge browsers in the latest version (Version 109.0.5414.74 (Official Build) (64-bit)).

Firefox is not affected.

### Regression?
beta.94

### Impact
High

### Esri team
Calcite (dev)

ethanbdev:
Oh nice - I noticed this issue and was just about to open one.

If its helpful, I found this issue did _not_ reproduce on chrome v108 🤷 

Here was my codepen anyway: https://codepen.io/eborgen/pen/vYaWgoz?editors=1010

I could message some screenshots of it being bad in the map viewer, but its the same with a dropdown inside an open block

jcfranco:
@ethanbdev Interesting. Thanks for sharing the browser version info. I'm wondering if we should hold off on this to see if it gets fixed in a future Chrome version.

geospatialem:
Moving to the February 2023 milestone per Franco's note above regarding checking for a potential fix in a future Chrome/Edge browser update.

atomiks:
It's not Chrome's bug per se, they're now [adhering to the spec](https://github.com/shoelace-style/shoelace/issues/1135#issuecomment-1387325386). 

[...]

PR #6520, opened 2023-02-24
fix: apply offsetParent polyfill for Chrome 109+

jcfranco:
**Related Issue:** #6300 

## Summary

This addresses positioning issues in Chrome 109+. Version 109 updated offsetParent behavior to follow the latest spec changes from https://github.com/w3c/csswg-drafts/issues/159.


src/utils/floating-ui.ts
```diff
@@ -27,29 +27,44 @@ async function patchFloatingUiForNonChromiumBrowsers(): Promise<void> {
     platform: string;
   }
 
+  function getUAData(): NavigatorUAData | undefined {
+    return (navigator as any).userAgentData;
```
benelan, line 31: I assume you had to assert `navigator` as `any` because `userAgentData` is experimental and not included in its type yet?

https://developer.mozilla.org/en-US/docs/Web/API/Navigator/userAgentData#browser_compatibility

benelan (approved):
LGTM

src/utils/floating-ui.ts
jcfranco: Correct and this particular code is based on `@floating-ui/dom` https://github.com/floating-ui/floating-ui/blob/f63411e85ae12bfe0c983e7c8104c57eb7849d0a/packages/dom/src/utils/userAgent.ts.

Issue #6203, opened 2022-12-29
Initials in avatar do not pass color contrast in all combinations


samanthahunter:
### Summary

Various combinations of the background color and initial text do not pass color contrast. ArcGIS Hub uses the avatar component heavily in our discussions UI (this feature is gated to premium users, so can contact Sam Hunter for credentials if needed) which is where we noticed the issue due to problems with custom themes being applied. However, after checking the Design System and Figma examples in a contrast checker, various combinations are not passing even when a custom theme is absent so the issue isn't just in Hub.

### Actual Behavior

Calcite avatar in pink/light mode, and Calcite avatar in green/dark mode - do not pass contrast. We did not check all combinations.

<img width="825" alt="Screenshot 2022-12-29 at 2 40 51 PM" src="https://user-images.githubusercontent.com/33425289/210004283-773a8c0c-5c10-4154-9135-cc920e281df0.png">

Also seeing issues when testing options from Figma combos which are different than the Design System site:
<img width="748" alt="Screenshot 2022-12-29 at 2 40 43 PM" src="https://user-images.githubusercontent.com/33425289/210004336-fac27981-d400-48a4-8650-181891f0f765.png">

### Expected Behavior

All text and color combos pass color contrast when no custom theme is applied (uses Calcite defaults)

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/avatar/

### Reproduction Steps

Looking at combinations from the [Design System site](https://developers.arcgis.com/calcite-design-system/components/avatar/) using dev tools I inspect for the hex colors to put in in [web Aim's contrast checker](https://webaim.org/resources/contrastchecker/)

### Reproduction Version

v1.0.0-beta.98

### Working W3C Example/Tutorial

https://www.w3.org/TR/WCAG21/#contrast-minimum

### Relevant Info

I was testing against AA - where "Large text is defined as 14 point (typically 18.66px) and bold or larger, or 18 point (typically 24px) or larger."

<img width="332" alt="Screenshot 2022-12-29 at 2 49 11 PM" src="https://user-images.githubusercontent.com/33425289/210004984-aaeadf2a-8880-4b01-862d-70a6f2f88b80.png">

Given initials text is not larger than 18px, it does not qualify as 'Large' so it needs to pass on 'Normal' text criteria on [WebAim](https://webaim.org/resources/contrastchecker/) if I am understanding the requirements correctly.

From 2.1 guidelines:
[...]

PR #6528, opened 2023-02-26
fix(avatar): pass color contrast in all combinations

Elijbet:
**Related Issue:** #6203

Modifies the function `stringToContrastCompliantHex` that takes in the `avatar id/name` to generate a `hex` value based on the provided string with an added WCAG contrast validator based on the theme. Also adds a special case to handle initial colors being `#fff or #000.`

The `adjustmentFactor = 0.1` can be used to adjust the `lightness` value in the `hsl`.


src/components/avatar/utils.ts
```diff
@@ -63,3 +77,51 @@ export function rgbToHue(rgb: RGB): number {
...
+
+/**
+ * For RGB color, find luminance
+ *
+ * @param colorObject { r: number; g: number; b: number }
+ * @param colorObject.r
+ * @param colorObject.g
+ * @param colorObject.b
+ */
+export function rgbToLuminance(colorObject: { r: number; g: number; b: number }): number {
```
jcfranco, line 89: I believe the `color` module (used by `calcite-color-picker`) already has utilities to get a color's luminance and contrast. Can you look at using those instead?

src/components/avatar/utils.ts
```diff
@@ -5,20 +5,34 @@ import { hexToRGB } from "../color-picker/utils";
...
  */
-export function stringToHex(str: string): string {
-  let hash = 0;
-  for (let i = 0; i < str.length; i++) {
-    hash = str.charCodeAt(i) + ((hash << 5) - hash);
-  }
+export function stringToHex(name: string, theme?: string, scale?: string): string {
+  const { constName, constTheme, constScale } = { constName: name, constTheme: theme, constScale: scale };
+
+  const generateNewHexValue = () => {
```
jcfranco, line 15: Does this need to be a separate function? It seems to only be used once.

src/components/avatar/utils.ts
Elijbet: Sure this is my first pass figuring out the basic logic. Still working it out. Will take that note into account.

src/components/avatar/utils.ts
```diff
@@ -63,3 +78,39 @@ export function rgbToHue(rgb: RGB): number {
...
+  const getLuminance = (hex) => {
+    const rgbObj = hexToRGB(hex);
+    return rgbToLuminosity(rgbObj);
+  };
+  const themeInitials = theme === "light" ? "#6A6A6A" : "#9f9f9f";
+
+  //dark mode uses ui-text-3-dark #9f9f9f (lighter grey initials, darker background)  luminosity 0.35
+  //light mode uses ui-text-3 #6a6a6a (darker grey initials, lighter background) luminosity 0.144
+  const k = 0.05; //small offset to prevent division by zero
+  const [l1, l2] =
```
jcfranco, line 109: `color` already provides this util: https://github.com/Qix-/color/blob/master/index.js#L276

src/components/avatar/utils.ts
```diff
@@ -63,3 +78,39 @@ export function rgbToHue(rgb: RGB): number {
...
+
+/**
+ * Verify the generated color is WCAG contrast compliant
+ *
+ * @param hex {string} - form of "#------"
+ * @param theme {string}
+ */
+export function isContrastCompliant(hex: string, theme: string): string | null {
+  const getLuminance = (hex) => {
+    const rgbObj = hexToRGB(hex);
```
jcfranco, line 101: This isn't needed as `color` can deal with different color models (i.e., you should be able to get luminosity from a hex-created color instance).

src/components/avatar/utils.ts
```diff
@@ -1,24 +1,39 @@
...
+
+export function stringToHexCompliant(args: stringArgs): string {
+  const { id, name, theme } = args;
+  let hex;
+
+  if (id) {
+    hex = id;
+  } else {
+    //returns a 6-digit color code based on a string input
+    const hash = Array.from(name).reduce((acc, char) => char.charCodeAt(0) + ((acc << 5) - acc), 0);
```
jcfranco, line 22: Can you explain these changes (compared to the original implementation) and why they are needed?

Elijbet:
The color of the name initials for light and dark modes are not far apart enough to create interesting background colors that would pass contrast. 

To make light letters lighter and dark letters dark, I  swapped the `ui-text-3` with `ui-text-2`. 

Original (master)
Screenshot 2023-03-09 at 5 22 06 PM
After passing contrast while applying more contrast to themed text.
Screenshot 2023-03-09 at 5 16 01 PM
You can see that the original is visibly muddier.

paulcpederson:
I agree the higher contrast text is much better!

I am a bit concerned, though, as to how these more vivid backgrounds are going to feel on a page full of member thumbs. The higher saturation is going to make something like a table of users pretty intense. Bryan's original intention here was to desaturate these background colors so that they are more muted.

ashetland:
If we try text-1 we should be able to reduce the saturation of the bg colors.

ashetland:
I think our solution to the contrast problem with some of the color combos might be simpler than we originally thought. In this mockup, I sampled all the bg colors from the Original screenshot and made versions using text-2 for both light and dark. All these pass AA with just changing color for the text and we keep the less saturated bg colors.

Frame 12@2x

paulcpederson:
Yeah, here is a codepen of all possible background colors in the old implementation, but with text-2: https://codepen.io/paulcp/pen/PodEpRW  

The saturation and lightness are always the same and only the hue is dynamically generated. With text-2 they all pass so I think you would have a 100% compliant set.

Elijbet:
> I guess I don't really understand. If you look at something like `IC` in the provided example, the more vivid green is going to have way _less_ contrast with the text. If I grab those values and check their contrast, the paler green with ui-text-2 has a 7.57:1 contrast ratio while the more vivid green with the same text color only has a 5.3:1 ratio.
> 
> So it seems like making the backgrounds more vivid actually reduces contrast which is sort of the opposite of what we want, right? Totally possible that I don't fully understand the nuances here, but on it's surface in my mind making the backgrounds vivid is counterproductive to the stated goals of this pr.

Makes sense, another question is: does the pale avatar background need to also pass the contrast with the overall background?

paulcpederson:
> That said, I agree that pale color solution looks more consistent (less intense), I guess another question is: does the pale avatar background need to also pass the contrast ration with the overall background, because it's hard to see against white?

After looking over the contrast rules for WCAG 2.0 I don't think we need to.

Elijbet:
To recap:

- initials use `ui-text-3` for its color
- fill-generating logic falls back to `text-2` if no proper hex is generated, 
- fill-generating logic does not test for the contrast between itself and the text
- fill-generating logic always generates unsaturated color (per Paul's PR comment)
- The latest proposed solution is to use `ui-text-2` to improve contrast
- initials might not be displayed always (e.g., if a thumbnail is provided)

`text-color-2` is using `--calcite-ui-text-2` via the tailwind config and the simplified approach does not take into consideration the foreground color being overridden by the user. So we may end up with the same issue again.

Elijbet:
Maybe a good compromise would be that colors may get funky if users play with the base text/foreground colors, which means we decided to do both :

1. use a different CSS prop for text color (text-2) and 
2. adjust the background color if there is not enough contrast. 

* Anyone using CC out of the box w/o changing any of the CSS props will get a pretty similar appearance for initials & background color since this change will prevent the custom color tweaking from kicking in because contrast should be sufficient (per the PR discussion).
* Anyone changing CSS props might cause the generated color to fail required contrast checks and then activate custom color-tweaking logic. 

I will look into that green avatar being super bright. That is actually the only color that seems to get “vivid” with contrast adjustment, the other colors are just a bit darker. 

Splitting this pr into 2. I will submit a new one quickly on text contrast separately. And repurpose this one to "adjusting-if-user-modified" case?

Elijbet:
Creating a refactor issue to make sure hex-generating never fails to generate a valid hex.

paulcpederson:
> I will look into that green avatar being super bright. That is actually the only color that seems to get “vivid” with contrast adjustment, the other colors are just a bit darker.

@Elijbet I think you are comparing the straight hex with the text color in your compliance check. But that is not actually what will be shown to the user. The generated hex is only used to get a hue. You need to do your contrast check with the hue at `hsl(${hue}, 60%, 90%)` for light and `hsl(${hue}, 60%, 20%)` for dark. Right?

Elijbet:
The draft doesn’t have the reworked version of the code.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

jcfranco:
@Elijbet Per our chat, closing this until there's a demand to adjust user-defined colors (via CSS props). https://github.com/Esri/calcite-design-system/pull/6592 addressed the associated issue.

PR #6550, opened 2023-03-01
build: leanify CSS

jcfranco:
**Related Issue:** N/A

## Summary

🧑‍🔬 Wired up [`cssnano`](https://github.com/cssnano/cssnano) to see if we can optimize our CSS even further.


stencil.config.ts
```diff
@@ -111,6 +112,9 @@ export const create: () => Config = () => ({
         stylelint({
           configFile: ".stylelintrc-postcss.json",
           fix: true
+        }),
+        cssnano({
+          preset: "default"
```
alisonailea, line 117: this default comes with SVGO which I think is already built into icons. This might cause an issue? Not sure, just want to point it out.

jcfranco: Thanks for bring this up. The icon assets (path data) [shouldn't get processed by `postcss`](https://github.com/stencil-community/stencil-postcss/tree/master#valid-file-extensions), so I _think_ we're good here.

benelan (approved):
:+1:

jcfranco:
This ended up increasing the build size (12.53 MB ➡️ 12.56 MB). 🤡  We can revisit later, closing for now.

womp-womp

Issue #6523, opened 2023-02-24
[List] `filterText` and `filterEnabled` not initializing properly on load


geospatialem:
### Actual Behavior

The `list` component's `filterEnabled` and `filterText` are not performing as-expected on initialization, where:

1. When a `filterText` value is added on initialization, all child `list-item`s are hidden.

2. Additionally, when `filterEnabled` is `true` on initialization a console message is displayed:
> "No data was passed to the filter function. The data argument should be an array of objects"

### Expected Behavior

On initialization:
1. `filter-text` should depict the correct child `list-item` elements
2. `filter-enabled` should not load a console message

### Reproduction Sample

https://codepen.io/geospatialem/pen/BaOLpbN

### Reproduction Steps

1. Open the [sample](https://codepen.io/geospatialem/pen/BaOLpbN)
2. Observe the console error of `"No data was passed to the filter function. The data argument should be an array of objects"`
3. Observe no `list-item`s are present when a `filter-text` value is entered during initialization
4. Remove the `filter-text` value and observe all `list-item`s present
5. Enter in a new `filter-text` value in the input and observe `list-item`s will filter as the text is entered

### Reproduction Version

1.0.7

### Relevant Info

We had a similar fix with https://github.com/Esri/calcite-components/pull/4207, which could be a similar solution to mitigate. cc @jcfranco 

### Priority impact

p1 - need for current milestone

### Esri team

Calcite (dev)

geospatialem:
Verified in `1.0.8-next.4`

![image](https://user-images.githubusercontent.com/5023024/222465643-f706d332-c3a1-4492-8c2f-1924bbf45c4a.png)

PR #6551, opened 2023-03-01
fix(filter, list): filter properly on initialization

jcfranco:
**Related Issue:** #6523

## Summary

This fixes initial filtering on consumers of `calcite-filter`. 

#### `list`

* `updateListItems` (private) is no longer debounced to allow updated values to be passed to the internal filter. This was also called whenever the items are modified (`MutationObserver`), so this could be revisited later if we see noticeable impact on performance
* Corrected assertion where filteredItems and filteredData were not matching (one reflects the other)
* Added new test

#### `filter`

* Removed a duplicate `filter` call that would interfere with the initial one – (introduced [here](https://github.com/Esri/calcite-components/pull/5471), possibly an oversight)
* Restored skipped test
* Updated test to cover filtering with default filter text

### DIFF
--- a/src/components/filter/filter.e2e.ts
+++ b/src/components/filter/filter.e2e.ts
@@ -173,24 +173,36 @@ describe("calcite-filter", () => {
       });
     });
 
-    it.skip("updates filtered items after filtering", async () => {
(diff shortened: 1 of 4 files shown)


anveshmekala (approved):
🚀

PR #6552, opened 2023-03-01
ci(sync_from_airtable): update gh issues from airtable records

alisonailea:
## Summary

In Airtable there are a set of automations listening for changes to certain cells and passing those changes along to Github via a webhook. This webhook is sent along with a Github PAC which helps secure the passed data.

Once triggered this workflow expects the Github issue number, the issue title, a data object, and an action name. This action name matches the action types expected by Github on issues. These action types are assigned/unassigned, labeled/unlabeled, milestoned/demilestoned, open/closed, and edited (changes to the title or description).
The workflow first uses github.rest.issues.get to query the API for an existing issue based on the passed issue number or passed issue title. If no issue is found the workflow assumes this is a new issue and jumps to the new_issue step of the workflow. Otherwise, the found issue is updated based on it's action type and the passed data object.

Heavy testing has already been conducted using the same code in calcite-design-tokens so no major issues are anticipated. However, all changes to the assignees, milestone, labels, title and description are tracked by Github and can be overwritten seen and, if issues to the update are found, overwritten. Once merged, I will be closely following reviewing all triggers to this workflow to ensure no unexpected changes occur.


.github/workflows/sync_from_airtable.yml
```diff
@@ -0,0 +1,325 @@
...
+                        issue_number: NUMBER,
+                        milestone: null,
+                    });
+                    core.info(`Issue ${ NUMBER }'s Milestone updated to 'null'`)
+                }
+            } catch (error) {
+                core.info(error)
+            }
+  open_closed:
+    name: Opened, ReOpened, or Closed
```
jcfranco, line 255: Nitpick: `ReOpened` -> `Reopened`

.github/workflows/sync_from_airtable.yml
```diff
@@ -0,0 +1,325 @@
...
+                            await github.rest.issues.removeLabel({
+                                owner: OWNER,
+                                repo: REPO,
+                                issue_number: NUMBER,
+                                name
+                            });
+                            core.info(`Remove Label: ${name}`)
+                        }));
+                    } catch (err) {
+                        if (!error.includes('Label does not exist')) { core.error(`Label not removed: ${err}`) }
```
jcfranco, line 207: For consistency, can you move the if statement contents to its own line?

.github/workflows/sync_from_airtable.yml
```diff
@@ -0,0 +1,325 @@
+name: Sync from Airtable
+
+on:
+  # publish blog using webhook
```
jcfranco, line 4: Is `blog` applicable?

jcfranco (commented):
Code's looking good. 

Can you share a bit more info on how this works and the plan for monitoring changes and possibly how to roll back/restore changes if needed?

alisonailea:
> Can you share a bit more info on how this works and the plan for monitoring changes and possibly how to roll back/restore changes if needed?

In Airtable there are a set of automations listening for changes to certain cells and passing those changes along to Github via a webhook. This webhook is sent along with a Github PAC which helps secure the passed data.

Once triggered this workflow expects the Github issue number, the issue title, a data object, and an action name. This action name matches the action types expected by Github on issues. These action types are assigned/unassigned, labeled/unlabeled, milestoned/demilestoned, open/closed, and edited (changes to the title or description).
The workflow first uses `github.rest.issues.get` to query the API for an existing issue based on the passed issue number or passed issue title. If no issue is found the workflow assumes this is a new issue and jumps to the `new_issue` step of the workflow. Otherwise, the found issue is updated based on it's action type and the passed data object. 

Heavy testing has already been conducted using the same code in calcite-design-tokens so no major issues are anticipated. However, all changes to the assignees, milestone, labels, title and description are tracked by Github and can be overwritten seen and, if issues to the update are found, overwritten. Once merged, I will be closely following reviewing all triggers to this workflow to ensure no unexpected changes occur.

.github/workflows/sync_from_airtable.yml
alisonailea: fixed

.github/workflows/sync_from_airtable.yml
alisonailea: fixed

.github/workflows/sync_from_airtable.yml
alisonailea: removed

.github/workflows/sync_from_airtable.yml
```diff
@@ -0,0 +1,330 @@
...
+            const repoName = `${OWNER}/${REPO}`;
+            try {
+                if (milestone) {
+                    const { repo, milestone_number } = milestone;
+                    if (repoName === repo) {
+                        await github.rest.issues.update({
+                            owner: OWNER,
+                            repo: REPO,
+                            issue_number: NUMBER,
+                            milestone: milestone_number || null,
```
benelan, line 238: Is it possible for a milestone to exist in airtable but not github? Do we know what happens when trying to set a milestone that doesn't exist? (This is assuming `env.DATA` is airtable info that you converted to github's schema)

.github/workflows/sync_from_airtable.yml
```diff
@@ -0,0 +1,330 @@
...
+          DATA: ${{ fromJSON(env.DATA) }}
+          NUMBER: ${{env.NUMBER}}
+          OWNER: ${{env.OWNER}}
+          REPO: ${{env.REPO}}
+        with:
+          script: |
+            const { DATA, NUMBER, OWNER, REPO } = process.env;
+            const data = JSON.parse(DATA);
+
+            await github.rest.issues.update({
```
benelan, line 298: I think `update` run into the same timing issue as #5851, which was tricky to troubleshoot. Something to keep an eye on.

benelan (commented):
Looks pretty good! Let me know on Teams when you finish working through these comments and I'll take another look. Sometimes my GitHub emails get buried :sweat_smile:

benelan:
> > Can you share a bit more info on how this works and the plan for monitoring changes and possibly how to roll back/restore changes if needed?
> 
> In Airtable there are a set of automations listening for changes to certain cells and passing those changes along to Github via a webhook. This webhook is sent along with a Github PAC which helps secure the passed data.
> 
> Once triggered this workflow expects the Github issue number, the issue title, a data object, and an action name. This action name matches the action types expected by Github on issues. These action types are assigned/unassigned, labeled/unlabeled, milestoned/demilestoned, open/closed, and edited (changes to the title or description). The workflow first uses `github.rest.issues.get` to query the API for an existing issue based on the passed issue number or passed issue title. If no issue is found the workflow assumes this is a new issue and jumps to the `new_issue` step of the workflow. Otherwise, the found issue is updated based on it's action type and the passed data object.
> 
> Heavy testing has already been conducted using the same code in calcite-design-tokens so no major issues are anticipated. However, all changes to the assignees, milestone, labels, title and description are tracked by Github and can be overwritten seen and, if issues to the update are found, overwritten. Once merged, I will be closely following reviewing all triggers to this workflow to ensure no unexpected changes occur.

Can you put this description for the PR body so it gets used for squashed commit message body?

.github/workflows/sync_from_airtable.yml
benelan: In this case, running `update` could overwrite edits to the issue body, title, labels, etc., which would be frustrating for a user. It might be worth looking into a different way to do this that specifically uses the add/remove methods instead of changing the whole issue to match the data provided in the request.

.github/workflows/sync_from_airtable.yml
```diff
@@ -0,0 +1,330 @@
...
+
+                return issue.status === 200 ? JSON.stringify(issue) : JSON.stringify(null);
+            } catch (error) {
+                
+                const issue = await github.rest.search.issuesAndPullRequests({
+                    q: `"${TITLE}" in:title is:open is:issue`,
+                    sort: 'created',
+                    order: 'asc',
+                    per_page: 1
+                }).then((result) =>  result[0]);
```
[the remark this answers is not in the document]
jcfranco: `result` can't be null. This method either resolves to an array or throws due to an error or invalid query. This may need a try/catch to deal with the latter.

.github/workflows/sync_from_airtable.yml
```diff
@@ -0,0 +1,330 @@
...
+
+                return issue ? JSON.stringify(issue) : JSON.stringify(null);
+            }
+
+            return JSON.stringify(null);
+
+  assigned_unassigned:
+    name: Assign or Unassign
+    needs: existing_issue
+    if: (fromJSON(needs.existing_issue.outputs.github_issue).status == 200 ) && (github.event.client_payload.action == 'assigned/unassigned')
```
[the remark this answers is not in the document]
alisonailea: these are custom actions being sent by Airtable so I have complete control over the payload. On the Airtable side I'm listening for the default Github action expressions but when being sent back I'm combining them because the behavior is the same.

.github/workflows/sync_from_airtable.yml
```diff
@@ -0,0 +1,330 @@
...
+                }
+                for (let i in assignees) {
+                    if (typeof ghLookup[assignees[i]] != 'undefined') {
+                    keepAssignees.push(assignees[i])
+                    delete ghLookup[assignees[i]];
+                    } else {
+                    addAssignees.push(assignees[i]);
+                    }
+                }
+                const removeAssignees = Object.keys(ghLookup);
```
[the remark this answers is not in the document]
alisonailea: I'm not a big fan of looping over the data twice. I can replace it with a single reduce, which might be more readable.

.github/workflows/sync_from_airtable.yml
```diff
@@ -0,0 +1,330 @@
...
+                }
+                for (let i in labels) {
+                    if (typeof ghLookup[labels[i]] != 'undefined') {
+                        keepLabels.push(labels[i])
+                        delete ghLookup[labels[i]];
+                    } else {
+                        addLabels.push(labels[i]);
+                    }
+                }
+                const removeLabels = Object.keys(ghLookup);
```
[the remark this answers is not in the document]
alisonailea: I'm not a big fan of looping over the data twice. I can replace it with a single reduce, which might be more readable.

.github/workflows/sync_from_airtable.yml
```diff
@@ -0,0 +1,330 @@
...
+                if (addLabels && addLabels.length > 0) {
+                    try {
+                        await github.rest.issues.addLabels({
```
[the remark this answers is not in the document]
alisonailea: addLabels will throw an error if the label doesn't already exist. Hence the trycatch

.github/workflows/sync_from_airtable.yml
```diff
@@ -0,0 +1,330 @@
...
+                                    await github.rest.issues.addLabels({
+                                        owner: OWNER,
+                                        repo: REPO,
+                                        issue_number: NUMBER,
+                                        labels: label
+                                    });
+                                    core.info(`Label added: ${label})`)
+                                } catch (error) {
+                                    try {
+                                        await github.rest.issues.createLabel({
```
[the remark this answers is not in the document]
alisonailea: This is for the edge case where someone might need to create a new label while working in Airtable and have that translate to a new label in Github. There is no way to add a description unless we want to make a whole new table in Airtable for all our Labels.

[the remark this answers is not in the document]
alisonailea: Unless we want to set a default description like "created in Airtable"

.github/workflows/sync_from_airtable.yml
alisonailea: Edited only runs when the Title or Description have changed. The history of both are logged in the GH issue.

benelan: If the title/description is changed in Airtable (by us) and GH (most likely the by user, but possibly us) at the same time then the changes to GH will be overwritten. It will be in the history but I don't check the history when working on an issue so I might miss some info. I'm fine with leaving it as is for now, but it's something to keep an eye on.

.github/workflows/sync_from_airtable.yml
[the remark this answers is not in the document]
benelan: Issues usually only have 0 or 1 assignees so I wasn't too concerned about iterating twice in this case, which is why I went with the more readable/concise option. I'd personally still go with the two filters in this scenario but I'm fine with leaving it as is

.github/workflows/sync_from_airtable.yml
```diff
@@ -0,0 +1,344 @@
...
+                    issue_number: NUMBER,
+                    owner: OWNER,
+                    repo: REPO,
+                });
+
+                return issue.status === 200 ? JSON.stringify(issue) : null;
+            } catch (error) {
+                
+                return null;
+
```
benelan, line 44: Nitpick: can you remove the extra white space above/below the return statement

.github/workflows/sync_from_airtable.yml
```diff
@@ -0,0 +1,344 @@
...
+                    }
+
+                    if (idx === labels.length - 1) {
+                        // We are in our final loop. Remove empty indices from keepLabels and set those priorLabels for removal.
+                        acc.keepLabels = acc.keepLabels.filter((label, keepLabelIdx) => {
+                            if (label === undefined) {
+                                acc.removeLabels.push(acc.priorLabels[keepLabelIdx]);
+                                return false;
+                            } else {
+                                return true;
```
benelan, line 148: Nitpick: `return true` doesn't need to be in an else statement because the if block returns

.github/workflows/sync_from_airtable.yml
```diff
@@ -0,0 +1,344 @@
...
+                    }
+
+                    if (idx === labels.length - 1) {
+                        // We are in our final loop. Remove empty indices from keepAssignees and set those priorAssignees for removal.
+                        acc.keepAssignees = acc.keepAssignees.filter((label, keepLabelIdx) => {
+                            if (label === undefined) {
+                                acc.removeAssignees.push(acc.priorAssignees[keepLabelIdx]);
+                                return false;
+                            } else {
+                                return true;
```
benelan, line 84: Nitpick: this return doesn't need to be in an else block either

benelan (approved):
A couple nitpicks but otherwise LGTM! Thanks for working through my comments/questions

.github/workflows/sync_from_airtable.yml
```diff
@@ -73,38 +62,48 @@ jobs:
...
-                const addAssignees = [];
-                const ghLookup = {};
-                for (let j in ghAssignees) {
-                    ghLookup[ghAssignees[j]] = ghAssignees[j];
-                }
-                for (let i in assignees) {
-                    if (typeof ghLookup[assignees[i]] != 'undefined') {
-                    keepAssignees.push(assignees[i])
-                    delete ghLookup[assignees[i]];
+                const { removeAssignees, addAssignees } = assignees.reduce((acc, labelName, idx) => {
```
jcfranco, line 65: Nitpick: is there a more expressive name we could use instead of `acc`? Applies to similar callbacks.

.github/workflows/sync_from_airtable.yml
```diff
@@ -73,38 +62,48 @@ jobs:
...
-                }
-                if (removeAssignees && removeAssignees.length > 0) {
-                  await github.rest.issues.removeAssignees({
-                      owner: OWNER,
-                      repo: REPO,
-                      issue_number: NUMBER,
-                      assignees: removeAssignees
-                  });
-                  core.info('Assignees Removed: ', removeAssignees.join(','))
+                if (removeAssignees.length > 0) {
```
jcfranco, line 99: Nitpick: for readability, a newline would help visually separate adding vs removing assignee blocks.

.github/workflows/sync_from_airtable.yml
jcfranco, line 65: For consistency, `index` vs `idx`. Applies to similar callbacks.

.github/workflows/sync_from_airtable.yml
```diff
@@ -127,23 +126,67 @@ jobs:
...
+                    }
+
+                    return acc;
+                    
+                }, { keepLabels: new Array(issue.data.labels.length), removeLabels: [], addLabels: [], priorLabels: issue.data.labels.map((label) => label.name)});
+
+                if (addLabels.length > 0) {
+                    await Promise.all(addLabels.map(async (label) => {
+                        try {
+                            await github.rest.issues.getLabel({
```
jcfranco, line 157: Is `getLabel` used as another way to check for label presence? If so, does it handle something that `addLabels` wouldn't cover?

.github/workflows/sync_from_airtable.yml
```diff
@@ -127,23 +126,67 @@ jobs:
...
-                        addLabels.push(labels[i]);
+                        // Keep the label indices from the priorLabel list
+                        acc.keepLabels[currentIdx] = labelName;
                     }
-                }
-                const removeLabels = Object.keys(ghLookup);
-                if (addLabels && addLabels.length > 0) {
+
+                    if (idx === labels.length - 1) {
+                        // We are in our final loop. Remove empty indices from keepLabels and set those priorLabels for removal.
```
jcfranco, line 140: ` // We are in our final loop.`

Can you use a 'self-documenting variable' to get rid of the comment? 

```js
const inFinalLoop = idx === labels.length - 1;

if (inFinalLoop) {
/* ... */
```

This sort of forces us to update the variable if it's context changes later on vs the code and the comment.

jcfranco (approved):
Awesome, stuff @alisonailea! 

Once the non-nitpicks are addressed, this is good to merge! Thanks for addressing our comments. 💪

🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖
🤖🤖🔁🔁🔁🤖🔁🤖🤖🤖🔁🤖🔁🤖🤖🤖🔁🤖🤖🔁🔁🔁🤖
🤖🔁🤖🤖🤖🤖🤖🔁🤖🔁🤖🤖🔁🔁🤖🤖🔁🤖🔁🤖🤖🤖🤖
🤖🤖🔁🔁🤖🤖🤖🤖🔁🤖🤖🤖🔁🤖🔁🤖🔁🤖🔁🤖🤖🤖🤖
🤖🤖🤖🤖🔁🤖🤖🤖🔁🤖🤖🤖🔁🤖🤖🔁🔁🤖🔁🤖🤖🤖🤖
🤖🔁🔁🔁🤖🤖🤖🤖🔁🤖🤖🤖🔁🤖🤖🤖🔁🤖🤖🔁🔁🔁🤖
🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖🤖

alisonailea:
We are going to abandon Airtable

Issue #5979, opened 2022-12-10
Stepper: changing the "numbering-systems" attribute via dev tools does not cause page to re-render with updated numbering system


annierm18:
### Actual Behavior

In Stepper, changing the "numbering-systems" attribute via dev tools (not via the code in codepen), does not cause the page to re-render with the updated numbering systems. 

![Screenshot (108)](https://user-images.githubusercontent.com/25759835/206820929-a731eb6e-f217-484e-86d9-6d1bf6011a20.png)

### Expected Behavior

When a user changes the numbering system, the component should update and show the new numbering system.

### Reproduction Sample

https://codepen.io/ANVESHMEKALA/pen/NWzmpdd?editors=1010

### Reproduction Steps

1. Open either Stepper in codepen 
2. Open the dev tools. 
3. Change the numbering-system to a non-latn numbering system

### Reproduction Version

http://calcite-dev.esri.com:3008/demos/input-time-picker.html

### Esri team

N/A

benelan:
FWIW, changing `numberingSystem` in the code doesn't update it either; it's not specific to dev tools:
https://codepen.io/benesri/pen/OJEGgzY?editors=1010

However, it will rerender if you also change `lang` afterwards (uncomment line 3 in the JS section of the codepen above). 

What is the use case for dynamically changing the numbering system and not `lang`? When I spoke to Franco about this we couldn't think of one other than testing. See: https://github.com/Esri/calcite-components/pull/5467#discussion_r995020457

At minimum our doc needs to be updated with this information though.

annierm18:
@benelan Good point, this issue seems mostly applicable to testing. cc: @anveshmekala

anveshmekala:
I think what @annierm18  meant is when she changes the prop in codepen HTML page it re-renders the component and hence updated numbering system is displayed.

The re-render happens only when the `lang` prop is changed not the numbering system.

Regarding use case , if the app has to show an option of switching between two different numbering systems for a given lang (ex: Arabic locale). I agree we have to decide if it's the most common use-case with Esri applications at least to make the change.

benelan:
Ah gotcha. Yeah that's Codepen's auto-reload doing that but if you turn it off it won't re-render.

> Regarding use case , if the app has to show an option of switching between two different numbering systems for a given lang (ex: Arabic locale). I agree we have to decide if it's the most common use-case with Esri applications at least to make the change.

[...]

PR #6563, opened 2023-03-04
fix(stepper): rerender stepper items when parent numbering system changes

jcfranco:
**Related Issue:** #5979 

## Summary

Updates stepper to re-render items when its `numberingSystem` value is modified.


src/components/stepper/stepper.tsx
```diff
@@ -104,7 +112,9 @@ export class Stepper {
           const items = (event.currentTarget as HTMLSlotElement)
             .assignedElements()
             .filter((el) => el?.tagName === "CALCITE-STEPPER-ITEM");
-
+          items.forEach((item: HTMLCalciteStepperItemElement) => {
```
jcfranco, line 115: Can you extract this into a method and reuse it here and in the watcher above?

jcfranco (commented):
👍 Approved! I can't approve my own PR 😅

Thanks for making these updates, @anveshmekala!

Issue #5067, opened 2022-08-02
Opening modal immediately after hydration yields inconsistent transitions


nwhittaker:
### Actual Behavior

Once a `calcite-modal` element becomes hydrated, setting the `open` attribute causes the modal's backdrop to appear abruptly while the dialog area transitions in smoothly.

![ezgif com-gif-maker](https://user-images.githubusercontent.com/421496/182421026-5834f1a7-d7b5-45a2-80b2-0b4019322143.gif)

### Expected Behavior

The modal's backdrop also transitions in smoothly.

![ezgif com-gif-maker (1)](https://user-images.githubusercontent.com/421496/182421111-a6fd2f31-48ec-44d8-af98-574c16bbbaad.gif)

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/qBoxdgg

### Reproduction Steps

1. Visit code sample and notice how the modal's backdrop appears immediately.
2. Close the modal.
3. Click the **Open modal** button to re-open; notice how modal's backdrop transitions smoothly now.

It may be helpful to use [Chrome's Animations][1] tab to slow down the transitions.

[1]: https://developer.chrome.com/docs/devtools/css/animations/

### Reproduction Version

beta.86

### Relevant Info

A workaround is to wait about .5s after hydration before setting `open` to produces the expected result.

In Ember it's typical to render the `<calcite-modal>` element at the time it needs to open (ideally w/out a delay). `open` is set after hydration because setting it initially currently has a couple drawbacks:
1. Both backdrop and dialog appear abruptly (possibly expected?).
2. https://github.com/Esri/calcite-components/issues/4764

jcfranco:
Looks like the initial transition is skipped entirely because the host is initially rendered as end state (host initially rendered w/ `open`), so there's no initial set of props to animate/transition from.

nwhittaker:
There may be some confusion on my part around when an element is considered rendered relative to when it is considered hydrated.

I understand if there's no transition when `open` is enabled before hydration. But enabling it any time after hydration feels like it should be safe to assume the styles for the unopened state were applied.

geospatialem:
Reallocating this issue to the February priorities milestone.

jcfranco:
> There may be some confusion on my part around when an element is considered rendered relative to when it is considered hydrated.
> I understand if there's no transition when open is enabled before hydration. But enabling it any time after hydration feels like it should be safe to assume the styles for the unopened state were applied.

[...]

PR #6564, opened 2023-03-04
fix(modal): ensure modal transitions are in sync

jcfranco:
**Related Issue:** #5067 

## Summary

Moves an internal open state class to the topmost container so all transitions are in sync.


anveshmekala:
Tested in local and LGTM!

Issue #5920, opened 2022-12-06
i18n Numbering Systems Testing: Issue with # of digits using up and down arrows


annierm18:
### Actual Behavior

In the Calcite input component:

- With a greater than a certain number of digits, like >15 digits after the decimal and 2 digits before the decimal, using the up and down arrows rounds the number in an unexpected way.
- Without including a decimal separator: Is it expected that "e" appears? With "e", the decimal separator gets pushed to right after the first digit. When I backspace to make the number smaller, and manually remove the "e", the decimal separator remains in the same place. So my number goes from very large, like "8173847023850385108501835981", to very small, like "13.17384702"

Info from @benelan : "Using the increment buttons is converting it to a number"

### Expected Behavior

With decimal - the number should not round.
Without decimal - I am not sure what the expected behavior should be, but I imagine if I remove the symbol that means "number to the power of x", then decimal separator should move back to the original spot...e.g. my number "8173847023850385108501835981" would not become "13.17384702". 

### Reproduction Sample
https://codepen.io/ainermfc/pen/XWYMqNE

### Reproduction Steps
(With group-separator...the repro steps may involve a different # of digits without group-separator
1. Enter a number with > 17 digits and include a decimal separator somewhere in the number
6. Click the up or down arrows 
7. Enter a number with > 17 digits and no decimal
8. Use arrows and add digits until you see "e" appear in the number
9. Remove digits manually with the back space until e is removed

### Reproduction Version

1.0.0-beta.97

### Regression?

v1.0.0-beta.97

### Esri team

N/A

geospatialem:
Created a new issue to track the first portion of the original issue cc @annierm18 @benelan @jcfranco

geospatialem:
Verified on `master`

![verify-input](https://user-images.githubusercontent.com/5023024/225423325-422cbdc8-2082-4ac7-86f1-c752626691eb.gif)

PR #6580, opened 2023-03-09
fix(input, input-number): increment/decrement unsafe numbers without loss of precision

benelan:
**Related Issue:** #5920

## Summary

Numbers that cannot be represented in JavaScript's IEEE-754 double precision were being rounded when incrementing/decrementing because they were cast as Number (we store the value as a string). We use BigDecimal class to support numbers that do not fit into JavaScript's encoding method. I had overlooked the increment/decrement functionality when implementing BigDecimal.

BigDecimal uses BigInt internally, and unfortunately BigInt doesn't support exponential notation. For example, `BigInt("1e2") ` throws the error:

> Uncaught SyntaxError: invalid BigInt syntax

To work around the constraint, I manually expand any exponential notation numbers into normal decimal numbers. All of this is done via string manipulation, so sorry in advance for the burned retinas you are about to receive from this PR review.

For more info:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger#description


src/utils/number.ts
```diff
@@ -52,6 +55,10 @@
     return `${this.isNegative ? "-" : ""}${integers}${decimals.length ? "." + decimals : ""}`;
   }
 
+  toFixed(decimalPlaces: number): string {
+    return this.toString().replace(new RegExp(`-?\d+\.?\d{0, ${decimalPlaces}}`), "");
```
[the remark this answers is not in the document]
benelan: Huh, TIL. 

After looking into this, I don't think we even need to use toFixed() anymore. It was originally added in #3473 to deal with Javascript's precision issues, but using BigDecimal is a much more robust approach. I'll see if nuking the scary regex breaks everything or not

src/components/input-number/input-number.tsx
```diff
@@ -751,7 +761,7 @@ export class InputNumber
     event.stopPropagation();
   };
 
-  private setChildNumberElRef = (el) => {
+  private setChildNumberElRef = (el: HTMLInputElement) => {
```
jcfranco, line 764: ✨🏆✨

src/utils/math.ts
```diff
@@ -1,7 +1,9 @@
 export const clamp = (value: number, min: number, max: number): number => Math.max(min, Math.min(value, max));
 
-export const decimalPlaces = (value: number): number => {
-  const match = ("" + value).match(/(?:\.(\d+))?(?:[eE]([+-]?\d+))?$/);
+const decimalNumberRegex = new RegExp(/(?:\.(\d+))?(?:[eE]([+-]?\d+))?$/);
+
+export const decimalPlaces = (value: string): number => {
+  const match = value.match(decimalNumberRegex);
```
jcfranco, line 6: Suggestion: would `test` help simplify this?

src/components/tooltip/tooltip.e2e.ts
```diff
@@ -12,6 +12,8 @@ describe("calcite-tooltip", () => {
   /**
    * Helps assert the canceled Esc key press when closing tooltips
    * Must be called before the tooltip is closed via keyboard.
+   *
+   * @param page
```
jcfranco, line 16: Can you roll this file back? I think it was auto-fixed by the linter, but the file is unrelated to this PR.

src/components/slider/slider.tsx
```diff
@@ -1200,7 +1200,7 @@ export class Slider
     const percent = (x - left) / width;
     const mirror = this.shouldMirror();
     const clampedValue = this.clamp(this.min + range * (mirror ? 1 - percent : percent));
-    let value = Number(clampedValue.toFixed(decimalPlaces(this.step)));
+    let value = Number(clampedValue.toFixed(decimalPlaces(`${this.step}`)));
```
jcfranco, line 1203: If this fixes precision on the slider, can you update the PR title to reflect this? If it's to fix types, can you submit this in a separate PR? If it's the latter, we can merge the separate PR first and we can skip rolling it back here.

src/utils/number.ts
```diff
@@ -157,6 +166,52 @@ export function sanitizeExponentialNumberString(numberString: string, func: (s:
...
+ *
+ * @param {string} numberString - pre-validated exponential or decimal number
+ * @returns {string} numberString in decimal notation
+ */
+export function expandExponentialNumberString(numberString: string): string {
+  const exponentialParts = numberString.split(/[eE]/);
+  if (exponentialParts.length === 1) {
+    return numberString;
+  }
+  if (Number.isSafeInteger(+numberString)) {
```
jcfranco, line 181: Nitpick: for readability, can you add a newline between if statements?

jcfranco, line 181: Can you store `+numberString` in a local variable and reuse it?

src/components/input/input.tsx
```diff
@@ -610,26 +615,32 @@ export class Input
...
+      nudgedValue.subtract(`${inputMin}`).isNegative;
+
+    const nudgedValueAboveInputMax =
+      typeof inputMax === "number" &&
+      !isNaN(inputMax) &&
+      !nudgedValue.subtract(`${inputMax}`).isNegative;
 
-    const inputValPlaces = decimalPlaces(inputVal);
-    const inputStepPlaces = decimalPlaces(inputStep);
+    const finalValue = nudgedValueBelowInputMin
```
jcfranco, line 633: Suggestion: can this be refactored to compute and evaluate a nudged value check at a time? It's currently computing two values where one might possibly be thrown away.

src/components/slider/slider.tsx
benelan: The parameter needs to be a string so the decimalPlaces function can also be used with BigDecimal. Previously `step` was being converted to a string in the decimalPlaces function itself, but we would lose precision if converting BigDecimal to `number` before passing it to the function. So it doesn't fix anything in slider, it's just a small, required refactor.

src/utils/math.ts
benelan: We need to use the values of the match a couple lines down

src/components/slider/slider.tsx
benelan: Nevermind, Input isn't using decimalPlaces anymore, removing!

Issue #6531, opened 2023-02-27
Navigation Component


yelenakreyndel:
### Description

Implement Navigation component per design prototype: http://calcite-dev.esri.com:3001/demos/nav.html  and spec: https://www.figma.com/file/6GrtrMONiH13kk9DeJtIMd/Design-Specs?node-id=157%3A2032&t=nLLbnXFvznVkZnqR-1 

### User Stories

High priority for March

### Acceptance Criteria

Design specifications and interaction behavior demonstrated in a prototype are followed

### Relevant Info

For questions about Navigation component interaction behavior, contact @macandcheese. For Figma spec questions, please contact @ashetland. 

### Esri team

Calcite (dev)

yelenakreyndel:
New issue is created and assigned to @anveshmekala. Cc'ing @jcfranco, @geospatialem, @macandcheese

yelenakreyndel:
@anveshmekala I invited you to the file. Try accessing from email link if this one doesn't work: https://www.figma.com/file/6GrtrMONiH13kk9DeJtIMd/Design-Specs?node-id=157%3A2032&t=nLLbnXFvznVkZnqR-1 (I updated the url in the description as well).

geospatialem:
Reallocated to the April milestone.

geospatialem:
Migrating to the May milestone for additional testing.

geospatialem:
Verified on `master` with `nav-logo`, `nav-user`, `menu` and `menu-item`.

cc @ashetland @SkyeSeitz for Figma changes.

PR #6583, opened 2023-03-09
feat(nav): navigation component

anveshmekala:
**Related Issue:** #6531 

## Summary
WIP


src/components/nav-menu-item/nav-menu-item.tsx
```diff
@@ -13,7 +13,12 @@ import {
   VNode
 } from "@stencil/core";
 import { FlipContext } from "../interfaces";
-import { getElementDir, getSlotted } from "../../utils/dom";
+import {
+  getElementDir,
```
driskull, line 17: `getElementDir` shouldn't be used if at all possible.

src/components/nav-menu-item/nav-menu-item.tsx
```diff
@@ -431,48 +510,52 @@ export class CalciteNavMenuItem {
...
+          class={{
+            container: true,
+            "nav-item-vertical-parent": this.topLevelLayout === "vertical"
+          }}
           role="none"
         >
           <div class="item-content">
             <a
-              aria-current="todo"
+              aria-current="page"
```
geospatialem, line 532: "Current" is currently being read back on all nav menu items, which is great to have for the selected menu item, but not properly identified for nav menu items which are not selected.

Can a conditional be added where the  `aria-current="page"` attribute is read by the currently active page, and other non-selected menu items are set to `aria-current="false"`?

(Similar to this [MDN example](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current#example).)

src/components/nav-menu-item/nav-menu-item.tsx
```diff
@@ -281,6 +330,37 @@ export class CalciteNavMenuItem {
...
+    }
+  };
+
+  private hasSlottedItems(): boolean {
+    return this.el.querySelectorAll("[slot=menu-item-dropdown]").length > 0;
+  }
+
+  private focusFirst(): void {
+    setTimeout(() => this.subMenuItems[0].setFocus(), 1000);
+  }
```
geospatialem, line 345: Think I recall additional R&D is needed on the focus - wanted to point it out again, just to make sure. The delay is a bit disruptive to AT, where if the user tries to press any key, focus is completely lost.

geospatialem (commented):
Looking good! A few small behaviors noted regarding a11y, will need to do another manual test when we get further along.

src/components/nav-menu-item/nav-menu-item.tsx
anveshmekala: added an conditional where the `aria-current ="page"` when the nav-menu-item is focused.

Issue #6582, opened 2023-03-09
Empty transition-duration results in no popover events available


cont-kolomeets:
### Actual Behavior

In some browsers, especially Edge, popover doesn't emit events. This is observed on some Windows machines in our team. Looks like it appears mostly on slower machines. Please try the sample provided below on different Windows machines.

Looks like the root of the problem is that internally Calcite listens to "transitionstart/end" events on the "calcite-floating-ui-anim" element. But the `transition-duration` is not specified. Looks like some browsers play hide/show animation instantly and don't even through "transitionstart/end" native events. We hacked this problem by setting `transition-duration` to 100ms.

![image](https://user-images.githubusercontent.com/5318527/224013476-6c0315b2-666c-4963-af2a-e5d0a552f619.png)

![image](https://user-images.githubusercontent.com/5318527/224013517-63363963-4929-44c8-8f3a-66a3927cdb24.png)

### Expected Behavior

Popover must emit events and `transition-duration` must be specified explicitly.

### Reproduction Sample

https://codepen.io/cont-kolomeets/pen/QWVOvOR?editors=100

### Reproduction Steps

See the sample above.

### Reproduction Version

1.0.7

#### Machine 1:
Windows 10
Chrome 110.0.5481.178 - not reproducible
Edge 110.0.1587.63 - issue is observed

#### Machine 2
Windows 10
Chrome 110.0.5481.77 - issue is observed
Edge 109.0.1518.78 - issue is observed
FF 109.0.1 - issue is observed

### Regression?

Looks like a regression from 0.98

### Priority impact

p0 - emergency

### Esri team

ArcGIS Business/Community Analyst

jcfranco:
Doing some preliminary digging into this one. ⛏️👷‍♂️

jcfranco:
### TL;DR

I've submitted #6605 with a fix while #6604 is worked on.

### Findings

* the issue is present whenever `prefers-reduced-motion` is enabled
  * [the following](https://codepen.io/jcfranco/pen/gOdeVpq?editors=1000) could be used as a workaround
* it looks like the utils for `openCloseComponent` were never updated to handle the case where duration is set to 0s. https://github.com/Esri/calcite-components/issues/5206 addressed a special case for modal, but all open/close components will share the same problem. 
* the regression was likely introduced [here](https://github.com/Esri/calcite-components/pull/5981/files) where a workaround for `openCloseComponent` working w/ `prefers-reduced-motion` was removed

[...]

PR #6605, opened 2023-03-15
fix(alert, combobox, dropdown, input-date-picker, popover, tooltip): prefers-reduced-motion no longer prevents open/close components from emitting before + open/close events

jcfranco:
**Related Issue:** #6582 

## Summary

This restores a workaround to get open/close components to emit before+open/close events when `prefers-reduced-motion` is enabled. Also, a Sass utility was added help scale durations for animations that require fine-tuning. This technically restores some animation, but it is very reduced:

### `calcite-block`

block-reduced-motion

### `calcite-loader`
loader-reduced-motion

### `calcite-progress`

progress-reduced-motion

I _think_ this still aligns with the expected behavior of `prefers-reduced-motion`, but we should check to just in case. cc @geospatialem

### DIFF
--- a/src/assets/styles/_animation.scss
+++ b/src/assets/styles/_animation.scss
@@ -74,6 +74,16 @@
 
 @media (prefers-reduced-motion: reduce) {
   :root {
-    --calcite-internal-duration-factor: 0;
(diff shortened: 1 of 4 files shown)


driskull (approved):
👍 Nice!

Issue #5683, opened 2022-11-03
Accordion Item: Longer titles extend outside of the header area


kellyhutchins:
### Actual Behavior

If the accordion item header title is long (simulated this with underscores) the title will extend outside the container instead of wrapping.  
<img width="835" alt="Screen Shot 2022-11-03 at 9 39 25 AM" src="https://user-images.githubusercontent.com/2124187/199781419-2760537e-f712-47cb-8d9c-288c2a0c58cb.png">

![image](https://user-images.githubusercontent.com/5023024/201743676-f52f652c-746f-4d2a-9733-d2c796e46ea8.png)

### Expected Behavior

I expect the title to wrap not to overflow out of the container. 

### Reproduction Sample

https://codepen.io/geospatialem/pen/QWxgzJE

### Reproduction Steps

1. Go to the [code sample](https://codepen.io/geospatialem/pen/QWxgzJE)
2. Observe `accordion-item`s containing long `heading`s with underscores don't wrap, and overflow outside the header area. 

### Reproduction Version

beta.98

### Esri team

ArcGIS Instant Apps

ashetland:
I think this one might just be contained to Accordion Item as it looks like a bug. The presence of underscores shouldn't affect  text wrapping. Underscores should allow for line breaks just like using dashes or spaces currently do.

geospatialem:
@jcfranco Going to assign to the upcoming March milestone per Aaron's note above.

> Underscores should allow for line breaks just like using dashes or spaces currently do.

geospatialem:
Verified with the [following Codepen](https://codepen.io/geospatialem/pen/QWxgzJE) with `1.1.1-next.0`:

![image](https://user-images.githubusercontent.com/5023024/225966404-f77d2593-ecb4-4f2d-8c1b-5b48e0056786.png)

PR #6608, opened 2023-03-15
fix(accordion, accordion-item): now wraps long words in header (title & description)

Elijbet:
**Related Issue:** #5683

## Summary
The `accordion-item header` longer `title`/`description` no longer extends outside the container instead of wrapping.

Note: Tailwind's `break-words` implements the `overflow-wrap: break-word` CSS property, but it seems to have no impact for me in Chrome. Used  our `word-break() mixin` instead to achieve the desired effect.

Provided a `_TestOnly` screenshot that simulates long words with underscores.

### DIFF
--- a/src/components/accordion-item/accordion-item.scss
+++ b/src/components/accordion-item/accordion-item.scss
@@ -102,7 +102,7 @@
     items-center
     duration-150
     ease-in-out;
-  word-break: break-word;
+  @include word-break();
 }
 
 :host .content {
--- a/src/components/accordion-item/accordion-item.scss
+++ b/src/components/accordion-item/accordion-item.scss
@@ -102,6 +102,7 @@
     items-center
     duration-150
     ease-in-out;
+  word-break: break-word;
 }
 
 :host .content {
--- a/src/components/accordion/accordion.stories.ts
+++ b/src/components/accordion/accordion.stories.ts
@@ -264,28 +264,42 @@ export const withIconStartAndEnd_TestOnly = (): string =>
     `
   );
 
-export const mediumIconForLargeAccordionItem_TestOnly = (): string => html`
-  <calcite-accordion scale="l" style="width: 400px">
-    <calcite-accordion-item heading="Accordion Item" scale="l"></calcite-accordion-item>
-    <calcite-accordion-item
-      heading="Accordion Item IconStart/End"
-      icon-start="plane"
-      icon-end="plane"
-      scale="l"
-    ></calcite-accordion-item>
-    <calcite-accordion-item heading="Accordion Item" scale="l" description="A great subtitle"></calcite-accordion-item>
-    <calcite-accordion-item
-      heading="Accordion Item IconStart/End"
-      icon-start="plane"
-      icon-end="plane"
-      scale="l"
-      description="A great subtitle"
-    ></calcite-accordion-item>
-    <calcite-accordion-item
-      heading="Accordion Item IconStart/End with a potentially two line title"
-      icon-start="banana"
-      icon-end="banana"
-      scale="l"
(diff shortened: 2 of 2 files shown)


src/components/accordion-item/accordion-item.scss
```diff
@@ -102,6 +102,7 @@
     items-center
     duration-150
     ease-in-out;
+  word-break: break-word;
```
driskull, line 105: We have this mixin that we could use here: https://github.com/Esri/calcite-components/blob/14fff178e50a62cb58fc66aa43b7b2332d4d3eff/src/assets/styles/includes.scss#L22-L25

src/components/accordion/accordion.stories.ts
```diff
@@ -264,28 +264,42 @@ export const withIconStartAndEnd_TestOnly = (): string =>
...
-      description="A great subtitle"
-    ></calcite-accordion-item>
-    <calcite-accordion-item
-      heading="Accordion Item IconStart/End with a potentially two line title"
-      icon-start="banana"
-      icon-end="banana"
-      scale="l"
-    ></calcite-accordion-item>
-  </calcite-accordion>
+const useCases = [
```
driskull, line 267: Incase someone wants to add more of these in the future maybe you could type `useCases`? Like...

`const useCases: {icon:string; heading:string; description: string;}`

driskull (approved):
LGTM! ❤️ 👍

src/components/accordion/accordion.stories.ts
Elijbet: Kewl. I got two opinions in favor of plain HTML instead of this, what would you say in regard to that?

driskull: IMO what you have is fine. Its why ES6 templates exist. Why duplicate the same HTML?

src/components/accordion/accordion.stories.ts
```diff
@@ -264,28 +264,42 @@ export const withIconStartAndEnd_TestOnly = (): string =>
...
+  },
+  {
+    icon: "plane",
+    heading: "Extra long description and icons m /scale l",
+    description:
+      "Planes, helicopters, and jets. Planes, helicopters, and jets. Planes, helicopters, and jets. Planes, helicopters, and jets. Planes, helicopters, and jets. Planes, helicopters, and jets. Planes, helicopters, and jets. Planes, helicopters, and jets. Planes, helicopters, and jets. Planes, helicopters, and jets. Planes, helicopters, and jets. Planes, helicopters, and jets. Planes, helicopters, and jets. Planes, helicopters, and jets. Planes, helicopters, and jets."
+  }
+];
+
+const accordionItems = useCases
```
jcfranco, line 290: Might be good to rename `accordionItems` and `useCases` to be closer to the related test. They are generic at the moment.

src/components/accordion/accordion.stories.ts
```diff
@@ -264,28 +264,42 @@ export const withIconStartAndEnd_TestOnly = (): string =>
...
+        icon-start="${useCase.icon}"
+        icon-end="${useCase.icon}"
+        scale="l"
+        heading="${useCase.heading}"
+        description="${useCase.description}"
+      ></calcite-accordion-item>`
+  )
+  .join("");
+
+export const longHeading_MediumIconForLargeAccordionItem_testOnly = (): string => html`
```
jcfranco, line 303: Per [our guidelines](https://github.com/Esri/calcite-components/blob/3bd1b540aad7254bd34f726da926cc31e0897fc5/conventions/Documentation.md#screenshot-test-only-stories), suffix should be `_TestOnly`.

jcfranco (approved):
🎉

PR #6611, opened 2023-03-17
build: fix peer dependency resolution error

jcfranco:
**Related Issue:** N/A

## Summary

Updates dependencies that required `--legacy-peer-deps` for initial installs.


benelan (approved):
Nice find! As discussed on Teams, I removed a workaround that is no longer necessary and pushed it to this branch. Otherwise it LGTM!

PR #6623, opened 2023-03-21
fix: restore form control validation in Safari

jcfranco:
**Related Issue:** #6626

## Summary

This restores consistent usage of `requestSubmit` when submitting forms. `submit` and `requestSubmit` [behave differently](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/requestSubmit#usage_notes) and it was currently not going through form validation in Safari. This moves the polyfill to the form util to ensure it's available for components using the form util.


driskull (approved):
This looks good @jcfranco. The changes make sense, I don't think we considered inputs submitting previously.

Issue #6617, opened 2023-03-20
[Modal] Slot documentation should be kebab-cased, not camelCased.


macandcheese:
### Description

The slot doc for Modal's new slots should use `content-top` and `content-bottom` instead of `contentTop` and `contentBottom`.

This ends up being rendered in documentation as `contentTop`, when the slot name needs to be written as `content-top`.

Example in Notice: 
https://github.com/Esri/calcite-components/blob/cbf3f5a011b5f2a30611e3938c6ff787b1016bdd/src/components/notice/notice.tsx#L49

### Which Component

Modal

PR #6625, opened 2023-03-21
docs(modal): kebab-case slot names and sticky context

geospatialem:
**Related Issue:** #6617

## Summary
1. Updates `content-top` and `content-bottom` to kabob case
2. Adds additional context to the `content-top` and `content-bottom` slot descriptions


Elijbet (approved):
✨💎✨

macandcheese (approved):
Looks good! Not sure if we need this last ", where content remains..." bit for either of these.

Issue #6561, opened 2023-03-04
[accordion-item] Doc actions-start/end slots


jcfranco:
### Description

Stems from https://github.com/Esri/calcite-components/issues/2884#issuecomment-1454349330.

Both `actions-start` and `actions-end` slots are missing from [`accordion-item`'s doc](https://github.com/Esri/calcite-components/pull/4958).

### Which Component

`accordion-item`

### Resources

https://github.com/Esri/calcite-components/pull/4958

macandcheese:
We could add an additional sample to show recommended pattern / expected use case for these slots as well cc @geospatialem

PR #6631, opened 2023-03-22
docs(accordion-item): add actions-start and actions-end slots

geospatialem:
**Related Issue:** #6561

## Summary

Add `accordion`s `actions-start` and `actions-end` slots. 🎰


macandcheese (commented):
This should be for `accordion-item` right?

alisonailea (approved):
Looks good thanks @geospatialem and good catch @macandcheese 🎉

PR #6634, opened 2023-03-22
docs(input, input-number, input-text): api consistency across components

geospatialem:
**Related Issue:** n/a

## Summary
Document API consistency across components, in particular while working through the new component, `textarea`. cc: https://github.com/Esri/calcite-components/pull/5644

Update includes consistency with `input`, `input-number`, and `input-text` (in alignment with `textarea`) for the following:

1. `autofocus` prop:
> When `true`, the component is focused on page load. Only one element can contain `autofocus`. If multiple elements have `autofocus`, the first element will receive focus.

2. `selectText()` method:
> Selects the text of the component's `value`.

### DIFF
--- a/src/components/input-number/input-number.tsx
+++ b/src/components/input-number/input-number.tsx
@@ -99,7 +99,7 @@ export class InputNumber
   @Prop({ reflect: true }) alignment: Position = "start";
 
   /**
-   * When `true`, the component is focused on page load.
+   * When `true`, the component is focused on page load. Only one element can contain `autofocus`. If multiple elements have `autofocus`, the first element will receive focus.
    *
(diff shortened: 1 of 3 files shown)


src/components/input-number/input-number.tsx
```diff
@@ -99,7 +99,7 @@ export class InputNumber
   @Prop({ reflect: true }) alignment: Position = "start";
 
   /**
-   * When `true`, the component is focused on page load.
+   * When `true`, the component is focused on page load. Only one element can contain `autofocus`. If multiple elements have `autofocus`, the first element will receive focus.
```
jcfranco, line 102: Nice!

jcfranco (approved):
💥

PR #6635, opened 2023-03-23
refactor(locale): document and move utils for ease of reuse

benelan:
## Summary

Adds some comments and clarifying name changes to a few locale/number utils. It also moves the utils for parsing/formatting numerals one-by-one from input-date-picker to the NumberStringFormat class so they can be reused in time-picker.

---

@eriklharper - in time-picker, these are the steps you should take after user input:

1. Use `delocalizeNumerals` to parse the numbers into Latin numerals
2. Validate the time now that it is computer readable. If it is a valid time, set it to state.
3. Use Intl.DateTimeFormat to make sure the meridian, time separator, etc. are localized correctly
4. Use `localizeNumerals` to switch the numbers back into the user's numberingSystem, and display that localized value in the input

Don't hesitate to reach out if you have any questions, I'm always happy to help!


src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -941,30 +941,4 @@ export class InputDatePicker
       `The specified value "${value}" does not conform to the required format, "YYYY-MM-DD".`
     );
   }
-
-  private commonDateSeparators = [".", "-", "/"];
```
benelan, line 945: This was a convience thing I added which stopped working at some point during our breaking changes push to v1. It would convert common date separators into the separator for the user's locale as you typed. It made it easier to test in different locales and it seemed kinda nice so I left it in when committing. Since it's not working anyway, I left it out of this PR so the util could be reused for time-picker.

benelan: Based on the failing tests it looks like input-date-picker actually relies on this to format the date separator now, which wasn't the case before. Unfortunately I don't have the cycles to dig into the new date picker logic and figure out what's going on so I may close this PR.

@eriklharper these are the [utils you'll need for the `numberingSystem` part of time-picker](https://github.com/Esri/calcite-components/blob/84aaf3561080c42be9f078fef1faa152f3c60a50/src/utils/locale.ts#L441-L464). You can get the separator and other stuff from [`Intl.DateTimeFormat.formatToParts()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/formatToParts). That's all you were looking at date-fns to solve right?

eriklharper: RE: "I don't have the cycles to dig into the new date picker logic and figure out what's going on so I may close this PR."  This is most likely because of the behavior change I instituted that doesn't attempt to parse what the user types in the date picker input while they're typing any longer, only when they hit enter or blur the input is when it parses, validates and then attempts to update the value if it is different than the current value.

eriklharper: > @eriklharper these are the [utils you'll need for the numberingSystem part of time-picker](https://github.com/Esri/calcite-components/blob/84aaf3561080c42be9f078fef1faa152f3c60a50/src/utils/locale.ts#L441-L464). You can get the separator and other stuff from [Intl.DateTimeFormat.formatToParts()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/formatToParts). That's all you were looking at date-fns to solve right?

The utils will be good for translating the numbers of the time value from other numbering systems back into latn for sure, and we'll be able to get separator characters with formatToParts as you pointed out.  One thing we can't rely on cross-browser is `hourCycle` because of Firefox's lack of support for the `Intl.Locale.hourCycles` property, so that will be something we'll have to hard-code in a locale definition file if we roll our own solution, or rely on a package like `date-fns` that would provide it.  

@jcfranco and I discussed trying out datetime parsing libraries based on learnings he had from talking with Dasa Paddock when he was trying to do the same thing for the JS API.  They decided to use Luxon, but also evaluated `date-fns` too.  `date-fns` provides a comprehensive datetime parsing setup that does have some locale gaps to address, but long term could potentially save us on maintenance since its a well-used open source project, although based on how their activity has shown as of late (lots of open PRs and issues currently in their repo), I do have some concerns about how timely changes might be able to happen on their end with PRs and such.

I can see the pros and cons of both approaches for sure.  Franco and I also talked today about possibly maintaining a fork of `date-fns` if submitting PRs turns out to not be a suitable pathway due to lack of engagement on their end.  I'm going to play around more with your codepen and see what we might be able to come up with though, I do like the idea of it relying as much on the browser APIs as possible.

benelan: > behavior change I instituted that doesn't attempt to parse what the user types in the date picker input while they're typing any longer 

I don't know if `commonSeparators` will work for all of the locales we support, in terms of being the only thing displaying the proper date separator while the user types. It would be nice if we could use Intl or our NLS data for that like we did in the previous implementation.

> The utils will be good for translating the numbers of the time value from other numbering systems back into latn for sure, and we'll be able to get separator characters with formatToParts as you pointed out. One thing we can't rely on cross-browser is hourCycle 

Now that we know we can get the hourCycle, meridian, and separator from Intl, is there anything else preventing us from using a home-built solution to parse/format the times? As you mentioned, we would need to use our own custom logic to parse the time as the user types for the numberingSystem and separators anyway (assuming date-fns won't parse/format invalid dates). I guess we could put it through our own utils first and then try using `date-fns` once it's a valid date, but at that point the hard part is already done.

eriklharper: There's nothing technically blocking us from doing this ourselves, its just a question of whether we want to do the work of maintaining that work versus using a library.  Also, we don't have to parse the time value as the user types.  At least with these current iterations of date and time inputs that allow the user to type freely, we should let them type whatever they want and only attempt to parse when they submit or blur the input.  I guess by "parse as the user types" you probably are just referring to translating to their preferred numbering system, which is different than actually trying to convert the whole value in the input to a Date object.  We should still continue to translate the characters as they type, but can hold off on converting that typed value to a real Date when the value is committed.

Issue #5590, opened 2022-10-25
Pagination - Ensure custom controls provide proper textual name, role, and state information - (2036570530)


dqateam:
### Module: 
 22b Pagination 

### Violation: 
 Ensure custom controls provide proper textual name, role, and state information 

![image](https://user-images.githubusercontent.com/5023024/202749227-8faf69f7-2ddc-4775-894c-39dea53e09ca.png)

### WCAG Reference: 
- [1.1.1 Non-text Content](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html)
- [4.1.2 Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html) 

### Instance ID: 
 2036570530 

### Severity: 
 7 

### Description: 
 [Issue]
The current state information of the "pagination component" is not provided.

[User Impact]
Screen reader users will have difficulty determining the purpose and state of these controls.

[Code Reference] 
 ```
<button class="page is-selected">58</button>
 ``` 

### Note: 
 [Suggestion]
Ensure custom controls provide proper textual name, role, and state information. Properly structured native controls provide this information automatically. For custom controls, developers must explicitly add this information by using ARIA attributes.

Use the aria-current="page" class to highlight the currently active page.

[Compliant Code Example] 
 ```
<button class="page is-selected" aria-current="page">58</button>
 ```
 

### Media Type: 
 Custom Controls 

### Additional Resources: 
- Report Source: Product accessibility evaluation conducted on the Calcite Design System in August, 2022.

brittneytewks:
Removed "needs triage" since a priority is set and estimate is indicated

geospatialem:
Verified on `master` with NVDA and JAWS.

PR #6637, opened 2023-03-23
fix(pagination): add current page information for screen readers

jcfranco:
**Related Issue:** #5590

## Summary

Add `aria-current` to provide state information.


geospatialem:
> @geospatialem Do you have any suggestions on how we could automate testing for this?

Can't think of a way for us to add in an automated test with axe specific to the attribute, other than the general `it("it accessible"...` asynchronous overall test we perform since its not prop-related.

We could potentially add in more attributes to the current test... (_sample below_). @jcfranco WDYT?

```html
<calcite-pagination></calcite-pagination>
```

to

```html
<calcite-pagination page-size="10" start-item="50" total-items="100"></calcite-pagination>
```

Issue #5585, opened 2022-10-25
Card `selectable` state - Ensure custom controls provide proper textual name, role, and state information - (2036570012) 


dqateam:
### Module: 
 18b Card 

### Violation: 
 Ensure custom controls provide proper textual name, role, and state information 

![image](https://user-images.githubusercontent.com/5023024/202752959-f25fc7d0-fef4-4c60-84ff-e81dafc5667e.png)

### WCAG Reference: 
- [1.1.1 Non-text Content](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html)
- [4.1.2 Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html) 

### Instance ID: 
 2036570012 

### Severity: 
 7 

### Description: 
 [Issue]
There are checkboxes present under the following sections without appropriate name information. Currently the checkboxes have aria-label as "select/deselect" which in inappropriate. Examples include:

- Test 3. Selected Card
- Test 2. Selectable Card

[User Impact]
Screen reader users will be unable to determine that these controls can be checked or unchecked. They will also be unable to review which options they have chosen.

[Code Reference] 
 ```
<div aria-checked="false" aria-label="Select" class="toggle" role="checkbox" tabindex="0"><svg class="check-svg" viewBox="0 0 16 16"><path d=""></path></svg><slot></slot></div>
 ``` 

### Note: 
 [Suggestion]
Ensure custom controls provide proper textual name, role, and state information.

The best way to do this is to use native controls, as they come with this information built in. In this case, ``` <input type="checkbox"> ``` elements would be most appropriate.

Developers can often achieve the same visual appearance using progressively enhanced CSS. Alternatively, they can visually hide the checkboxes with opacity: 0 and use SVGs to fake their appearance.

If conversion to native checkboxes is not possible, the following information is expected:
- The element functioning as the checkbox must have role="checkbox".
- This checkbox element must have a label provided by internal text, aria-label, or aria-labelledby.
- When checked, the checkbox element must have aria-checked="true", and when not checked, it must have aria-checked="false". If partially checked, it must have aria-checked="mixed".
- Checkboxes in a logical group must be children of an element with role="group". That element must have aria-labelledby set to the ID of text labelling the group.

[Compliant Code Example] 
 ```
[...]

PR #6657, opened 2023-03-24
fix(card): provide more meaningful screen reader label for selectable cards (deprecates `deselect` message override)

jcfranco:
**Related Issue:** #5585

## Summary

Changes the text used for the card's checkbox to provide correct state information. Before, screen readers would read "Select, unchecked, checkbox"/ "Deselect, checked, checkbox". This text should now be more meaningful: "Selectable card, checked, checkbox"/ "Selectable card, unchecked, checkbox". 

## Notes

* The suggested fix from the audit included an identifier for the card and using the card's title/subtitle was considered, but not used as we can't reliably use it since neither are technically required.
* This reuses the `select` message and deprecates `deselect`. This is not considered a breaking change as any existing text overrides from `select` will still be used and provide applicable context and any provided `deselect` will end up being ignored.


jcfranco:
@geospatialem Any ideas for automated testing here? Not sure if asserting on the rendered string is of value or not as this change is more about the interpretation of the text. LMK what you think.

driskull (approved):
👍 

Looks good.

We might want to check other places we use `aria-checked` and make sure they represent the item description and not the action.

geospatialem:
> @geospatialem Any ideas for automated testing here? Not sure if asserting on the rendered string is of value or not as this change is more about the interpretation of the text. LMK what you think.

@jcfranco I don't think we could run automated testing in this case since it would have technically passed prior to this fix through automation.

jcfranco:
> We might want to check other places we use aria-checked and make sure they represent the item description and not the action.

@geospatialem Do you think our a11y audit caught all of these?

geospatialem:
> @geospatialem Do you think our a11y audit caught all of these?

It looks like there was only one other instance across components, which was in `stepper`/`stepper-item` with https://github.com/Esri/calcite-components/issues/5602. 

@jcfranco We have it slotted for the July release, but perhaps we could add it in the current one if its fix is similar?

jcfranco:
> perhaps we could add it in the current one if its fix is similar?

Let's keep it where it is and we can move it to an earlier milestone if our backlog allows.

Issue #5660, opened 2022-10-28
Buttons with long text should cut them off


yannikmesserli:
### Actual Behavior

When an unbreakable text is set to a button which is longer that what the button's width is capable of handling, the text overflow and break the button's layout.

### Expected Behavior

The text should be cut-off.

### Reproduction Sample

https://codepen.io/yannikm/pen/xxzGPRE?editors=100

### Reproduction Steps

Check the codepen's sample above.

### Reproduction Version

beta.97

### Esri team

ArcGIS Scene Viewer

alisonailea:
We have had some conflicting requests for the right resolution to text overflows in components. We may need to provide an optional prop or find another solution per designer's recommendations.

SkyeSeitz:
Thanks, @alisonailea. Let's connect on all the text overflow requests as a collection. 

@yannikmesserli do either of these examples below demonstrate the expected behavior you are describing?

<img width="1183" alt="image" src="https://user-images.githubusercontent.com/29716057/199351835-30a786fe-f749-4f0c-8c31-88163c471cb7.png">

ashetland:
The truncated version feels like the best failure state in the case. @alisonailea, we should sync up on other components where this occurs. Hopefully a similar strategy will work kit-wide. 🤞

jcfranco:
Hopefully, we can arrive at a single way of handling this to avoid props if possible. Some additional things to consider:

1. slotted content – depending on the slotted HTML, it won't be possible for us to handle truncation via CSS
2. translations/overrides (near future) – different languages may have strings that impact the original component layout, these can also be overridden by users, which seems closer to this issue
3. a11y – we should be fine here as long as truncation is handled at a visual level and not by manipulating text content

yannikmesserli:
> The truncated version feels like the best failure state in the case. @alisonailea, we should sync up on other components where this occurs. Hopefully a similar strategy will work kit-wide. 🤞

Agreed 🪁 

> We may need to provide an optional prop or find another solution per designer's recommendations. 

It probably makes sense to have an optional prop (`truncated`?). There are probably good uses cases that embed html element that needs to be on the edge, e.g. thumbs, so truncated inline element will break their design.

geospatialem:
cc fix from https://github.com/Esri/calcite-components/issues/5683#issuecomment-1474113754

jcfranco:
[...]

PR #6664, opened 2023-03-26
fix(button): truncate long button text

Elijbet:
**Related Issue:** #5660

## Summary

Truncate long text on fixed-width components that should not grow in height. 

To avoid unnecessary noise, detect the `buttons` with text that has been truncated and apply `title attribute` conditionally only to these, to reveal the full text for sighted users. This frees up the `calcite-tooltip` to provide additional context.

Added an e2e test to ensure the long text is truncated and the `title attribute` is added when it is.


jcfranco (commented):
Truncation looks good, but there isn't any context for sighted-users regarding what got truncated. cc @SkyeSeitz @geospatialem

SkyeSeitz:
> Truncation looks good, but there isn't any context for sighted-users regarding what got truncated. cc @SkyeSeitz @geospatialem

Summarizing decision made together: To provide context for sighted-users regarding what got truncated, we will use `calcite-tooltip` on text only when truncated.

alisonailea (approved):
👍🏻

benelan (approved):

src/components/button/button.tsx
```diff
@@ -201,6 +201,18 @@ export class Button
 
   componentDidLoad(): void {
     setComponentLoaded(this);
+
+    const slotted = this.el.shadowRoot
```
driskull, line 205: can you use `onSlotChange` instead of querying the slot?

src/components/button/button.tsx
```diff
@@ -201,6 +201,18 @@ export class Button
...
+      .querySelector("span.content")
+      .querySelector("slot") as HTMLSlotElement;
+
+    const textNodes = slotted.assignedNodes().filter((node) => node.nodeName === "#text");
```
driskull, line 209: Do you need the textNodes or can you just see if the parent node will work?

src/components/button/button.tsx
```diff
@@ -201,6 +201,18 @@ export class Button
...
+    const initialText = textNodes.map((node) => node.textContent).join("");
+
+    const spanContent = this.el.shadowRoot.querySelector("span.content");
+    const spanContentOffsetWidth = (spanContent as HTMLElement).offsetWidth;
+
+    spanContentOffsetWidth < initialText.length ? this.el.setAttribute("title", initialText) : null;
```
driskull, line 215: We shouldn't set the title on the host element which is in light dom.

src/components/button/button.tsx
Elijbet: Could you expand on this one?

src/components/button/button.tsx
```diff
@@ -358,4 +369,9 @@ export class Button
       resetForm(this);
     }
   };
+
+  private handleSlotchange = (event): void => {
+    const textNodes = event.target.assignedNodes().filter((node) => node.nodeName === "#text");
+    this.initialSlottedText = textNodes.map((node) => node.textContent).join("");
```
driskull, line 375: Is there a reason you need to access the text nodes? Seems like a container around the slot should suffice for getting the offsetWidth/scrollWidth.

src/components/button/button.tsx
```diff
@@ -200,7 +200,16 @@ export class Button
   }
 
   componentDidLoad(): void {
+    const { initialSlottedText, el } = this;
     setComponentLoaded(this);
+
+    const spanContent = el.shadowRoot.querySelector("span.content");
+    const spanContentOffsetWidth = (spanContent as HTMLElement).offsetWidth;
+    const button = el.shadowRoot.querySelector("button");
```
driskull, line 208: Instead of querying for these nodes, can you safe these nodes using the `ref` property on them?

src/components/button/button.tsx
Elijbet: Yeah, you're right. Not sure why I couldn't get it to work before. I think I needed to clean up the demo page noise to see the output more clearly.

src/components/button/button.tsx
```diff
@@ -200,7 +200,12 @@ export class Button
   }
 
   componentDidLoad(): void {
+    const { contentEl, childEl, fullLengthSlottedText } = this;
     setComponentLoaded(this);
+
+    contentEl.offsetWidth < contentEl.scrollWidth
```
driskull, line 206: @Elijbet can you create an internal state for whether to set the title attribute? Maybe you can just reuse the `fullLengthSlottedText` something like so...

```ts
private handleSlotchange = (event): void => {
const {contentEl} = this;
    const textNodes = event.target.assignedNodes().filter((node) => node.nodeName === "#text");
    this.fullLengthSlottedText = contentEl.offsetWidth < contentEl.scrollWidth ? textNodes.map((node) => node.textContent).join("") : null;
  };
```

Then in the JSX you can just do this...

```jsx
<childEl
title={this.fullLengthSlottedText}
```

I think you will also still need to setup a resize observer on the childEl because if the text size changes after componentDidLoad you will need to run the logic whether to apply the title again. The size of the button could change due to other props as well so a resizeObserver is probably necessary.

src/components/button/button.scss
```diff
@@ -69,6 +70,16 @@
   }
 }
 
+:host[title]:after {
```
driskull, line 73: This doesnt apply to the childEl so its not needed.

src/components/button/button.scss
```diff
@@ -69,6 +70,16 @@
...
+  @apply absolute inset-x-1/2 bottom-full w-max opacity-0;
+  content: attr(title);
+  transition: opacity var(--calcite-internal-animation-timing-fast) ease-in-out;
+}
+
+:host[title]:hover:after {
```
driskull, line 79: This doesnt apply to the childEl so its not needed.

src/components/button/button.tsx
```diff
@@ -241,8 +246,8 @@ export class Button
     );
 
     const contentEl = (
-      <span class={CSS.content}>
-        <slot />
+      <span class={CSS.content} ref={(el) => (this.contentEl = el)}>
+        <slot onSlotchange={this.handleSlotchange} />
```
driskull, line 250: I'm not sure you need onSlotChange actually, a resize observer should be enough. The resize observer can just get the text from the childEl. Probably from the textContent property on the element.

Elijbet: I thought we decided that there is no resize happening here for resizeObserver to pick up.

src/components/button/button.tsx
driskull: Can you try this?

```ts
this.fullLengthSlottedText = contentEl.offsetWidth < contentEl.scrollWidth ?
  contentEl.innerText : 
   null;
```

Elijbet: Ah, ok. Now I understand.

src/components/button/button.tsx
driskull: I'm suggesting having a function to determine whether or not to show the title attribute which can be called initially in componentDidLoad and anytime a resize of the component occurs.

Someting like:

```ts
setFullLengthText = () :void => {
const {contentEl} = this;
this.fullLengthSlottedText = contentEl.offsetWidth < contentEl.scrollWidth ?
  contentEl.innerText : 
   null;
}
```

```
resizeObserver = createObserver("resize", () => this.setFullLengthText());
```

Elijbet: Right, I understand.

src/components/button/button.tsx
```diff
@@ -332,6 +334,14 @@ export class Button
     this.mutationObserver?.observe(this.el, { childList: true, subtree: true });
   }
 
+  /** inital full length slotted text before the truncation */
+  private fullLengthSlottedText: string;
```
driskull, line 338: I think this should be `@State()` so that it triggers a re-render when this value changes.

driskull: Maybe we can think of a better name like just `fullText` or `tooltipText`?

driskull (changes requested):
Nice!

driskull (approved):
LGTM! 🎉 🍾

Issue #6185, opened 2022-12-28
[tile-select] Can't make selection by clicking tile with dist-custom-elements


mpayson:
### Actual Behavior

It looks like when using `dist-custom-elements`, users can't click a tile to make a selection for tile select components, they have to click the radio button circle. This differs from the `dist` behavior, where the user can click the entire tile to make a selection.

### Expected Behavior

The user can click the tile to make a selection like with the `dist` output.

### Reproduction Sample

https://github.com/mpayson/calcite-components-examples/tree/tile-select-issue

### Reproduction Steps

1. Clone the repository and checkout the `tile-select-issue` branch
2. `cd react && npm install && npm start`
3. Click a tile, see it does not make a selection
5. Go to this codepen: https://codepen.io/mpayson/pen/yLqezWB
6. Click a tile, see it does make a selection

### Reproduction Version

beta.99

### Esri team

ArcGIS Online

geospatialem:
Research will be conducted in March to determine next steps.

geospatialem:
Verified in `1.2.0-next.0`

![verify-select](https://user-images.githubusercontent.com/5023024/228124509-b5c017f0-7d69-4745-a382-0d31370be2ba.gif)

PR #6665, opened 2023-03-27
fix(tile-select): fix click not firing in custom-elements build

jcfranco:
**Related Issue:** #6185 

## Summary

Renames event handlers to prevent click listener not firing when the tile-select area was clicked in the `dist-custom-elements` output target. This seems like a Stencil bug caused when an event handler is identical to its listened event name (needs further investigation to confirm cc @benelan).


eriklharper:
Wow, super interesting to see this.  Had no idea that the name of the event handlers had any effect on click behavior.  Definitely a bug they should fix for sure.

eriklharper (approved):

Issue #6188, opened 2022-12-28
[modal] TypeError when calling `setFocus` with dist-custom-elements


mpayson:
### Actual Behavior

When calling `setFocus` on calcite-modal with dist-custom-elements, the TypeError below gets thrown. The error stems from `focusTrapEl` being undefined when focusFirstTabbable is called [here](https://github.com/Esri/calcite-components/blob/master/src/components/modal/modal.tsx#L387), so maybe the fix from #5749 does not work with the custom-elements output type? Also, it does look like `setFocus` gets called internally [here](https://github.com/Esri/calcite-components/blob/master/src/components/modal/modal.tsx#L463) and focuses the close button as expected, so consuming apps may not need to call it?

```
focusTrapComponent.js:76 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'tagName')
    at getCandidatesIteratively (focusTrapComponent.js:76:1)
    at tabbable (focusTrapComponent.js:418:1)
    at focusFirstTabbable (focusTrapComponent.js:1250:1)
    at Modal.open.setFocus (calcite-modal.js:226:1)
```

### Expected Behavior

`setFocus` method waits until the component has loaded and the TypeError is not thrown

### Reproduction Sample

https://github.com/mpayson/calcite-components-examples/tree/set-focus-issue

### Reproduction Steps

1. Clone the repository and checkout the set-focus-issue branch
2. `cd react && npm install && npm start`
3. Open developer console
4. Click button to open the modal, see console error
5. Switch the [loading strategy here](https://github.com/mpayson/calcite-components-examples/blob/set-focus-issue/react/src/App.js#L7) by commenting lines 5-6 and un-commenting lines 7-8
6. Refresh the page
7. Click button to open the modal, don't see console error

### Reproduction Version

beta.99

### Esri team

ArcGIS Online

benelan:
Note: this isn't specific to React, it repro's in the vite sample as well:

<details open>

<summary>index.html</summary>

```html
<!DOCTYPE html>
<html lang="en">
	<head>
		<meta charset="UTF-8" />
		<link rel="icon" type="image/svg+xml" href="favicon.svg" />
		<meta name="viewport" content="width=device-width, initial-scale=1.0" />
		<title>Calcite Vite</title>
	</head>
	<body>
      <calcite-button>Click me!</calcite-button>
        <calcite-modal open>
          <div slot="header">Demo</div>
          <div slot="content">hi!</div>
[...]

PR #6666, opened 2023-03-27
fix(modal): prevent error when calling setFocus on a recently rendered and opened modal (`dist-custom-elements`)

jcfranco:
**Related Issue:** #6188

## Summary

This is a workaround for modal throwing an error in the `dist-custom-elements` build when setting focus on the first focusable item after the component initially renders. Fortunately for modal, focus is set properly after the open animation.

The problem seems to be the focus-trap's element `ref` not being called on `componentDidLoad`, not even after using `requestAnimationFrame` as used in [this Ionic util](https://github.com/ionic-team/ionic-framework/blob/main/core/src/utils/helpers.ts#L72-L79), nor `setTimeout` as suggested [here](https://github.com/ionic-team/stencil/issues/3246#issuecomment-1048802446). We'll investigate and open an issue if this is indeed a lifecycle bug.

### DIFF
--- a/src/utils/dom.ts
+++ b/src/utils/dom.ts
@@ -175,6 +175,10 @@ export async function focusElement(el: FocusableElement): Promise<void> {
  * @param {HTMLElement} element The html element containing tabbable elements.
  */
 export function focusFirstTabbable(element: HTMLElement): void {
+  if (!element) {
+    return;
+  }
+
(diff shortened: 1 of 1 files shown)


driskull (approved):
👍

benelan (approved):

Issue #5164, opened 2022-08-16
Regression with `calcite-button#form` where form is not submitted/reset 


nwhittaker:
### Actual Behavior

Give a submit or reset button associated with a non-ancestor form, clicking the button does not submit/reset the form.

### Expected Behavior

Give a submit or reset button associated with a non-ancestor form, clicking the button does submit/reset the form.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/wvmQjMO

### Reproduction Steps

1. Visit the sample and enter some text into the input.
2. Click the reset button and not the input does not clear.
3. Open the console and click the submit button and note `"onsubmit"` is not printed.

### Reproduction Version

beta.91

### Relevant Info

Possibly introduced by https://github.com/Esri/calcite-components/pull/3287.

I'd also ask to revisit the decision to deprecate the `form` attribute. Marking it as deprecated suggests it'll be removed from the API at some point in the future. I understand it's no longer needed in the case where the buttons are descendants of the `<form>`, but if they're not, the attribute is still needed and perfectly valid.

### Regression?

next.290

jcfranco:
> I understand it's no longer needed in the case where the buttons are descendants of the `<form>`, but if they're not, the attribute is still needed and perfectly valid.

Agreed. We can revisit our form-associated pattern to accommodate this. In the meantime, can you use the form-descendant approach as a workaround? 

For prioritization, can you share more info about your use case and the impact of this?

nwhittaker:
>For prioritization, can you share more info about your use case and the impact of this?

One use-case where I came up against this issue was when I tried to treat the contents of a modal like a form, and the modal's buttons like form buttons:

```html
<calcite-modal>
  <form id="form-id" onsubmit="…" slot="content">
    <calcite-input />
  </form>
  <calcite-button form="form-id" slot="primary" type="submit" />
  <calcite-button form="form-id" slot="secondary" type="reset" />
</calcite-modal>
```

This lets me consolidate the submission logic and retain native submission behavior:
* Click submit button
* Hit `Enter` key when input is focused

Currently achieving this behavior would require separate handlers and added logic:

```html
<calcite-modal>
  <form slot="content">
    <calcite-input onkeyup="…" />
  </form>
  <calcite-button onclick="…" slot="primary" type="submit" />
  <calcite-button onclick="…" slot="secondary" type="reset" />
</calcite-modal>
```

[...]

PR #6682, opened 2023-03-28
feat: support setting form ID on form components

jcfranco:
**Related Issue:** #5164 

## Summary

This allows developers to associate form components by providing the form ID. 

### Notes

* the form must be available in the DOM before the component is attached to the DOM
* components will not look up ancestors for form if the ID is invalid or if the form is not found

### Follow-up

I'll DRY up some duplication in the `commonTests#formAssociated` test helper in a separate PR. I went this route to expedite testing.

### DIFF
--- a/src/components/button/button.tsx
+++ b/src/components/button/button.tsx
@@ -1,6 +1,5 @@
 import { Build, Component, Element, h, Method, Prop, State, VNode, Watch } from "@stencil/core";
-import { closestElementCrossShadowBoundary } from "../../utils/dom";
-import { FormOwner, resetForm, submitForm } from "../../utils/form";
+import { findAssociatedForm, FormOwner, resetForm, submitForm } from "../../utils/form";
 import { InteractiveComponent, updateHostInteraction } from "../../utils/interactive";
 import { connectLabel, disconnectLabel, getLabelText, LabelableComponent } from "../../utils/label";
 import {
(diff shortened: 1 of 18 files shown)


src/components/button/button.tsx
```diff
@@ -75,6 +74,14 @@ export class Button
   /**  When `true`, interaction is prevented and the component is displayed with lower opacity. */
   @Prop({ reflect: true }) disabled = false;
 
+  /**
+   * The ID of the form owner to associate this component with.
```
benelan, line 78: Maybe "The ID of the form owner to associate with the component"? IIRC ending a sentence with "with" is frowned upon by  the dictionary dwellers. Applies to the other components as well.

src/components/button/button.tsx
```diff
@@ -75,6 +74,14 @@ export class Button
...
+   *
+   * When not set, this component will be associated with its ancestor `<form>` element, if any.
```
benelan, line 80: Nitpick: we have been saying "the component" rather than "this component" in all of the API ref.  Applies to the other components as well.

src/utils/form.spec.ts
```diff
@@ -19,4 +21,44 @@ describe("form", () => {
...
+      formEl.append(el);
+      const formOwner: FormOwner = { formEl, form, el };
+
+      expect(findAssociatedForm(formOwner)).toBe(formEl);
+    });
+
+    it("finds form via ID", async () => {
+      const formEl = document.createElement("form");
+      formEl.id = "test-form";
+      const form = "test-form";
```
benelan, line 56: Nitpick: this can be

```js
const form = "test-form";
formEl.id = form;
```

benelan (approved):
Couple comments, otherwise LGTM with the commonTest stuff being hung up to DRY in a follow up PR

src/components/button/button.tsx
jcfranco: Great suggestion with!

PR #6707, opened 2023-03-30
refactor(tailwind): extract tailwind preset

jrowlingson:
**Related Issue:** https://github.com/Esri/calcite-components/issues/6602

## Summary

This also bumps tailwind to v3.3.1.


mpriour:
I really like this idea. It does give end users an easy way to incorporate our styling without having to manually mesh our tailwind config with theirs.

jcfranco:
Awesome! Thanks for extracting the config into its own preset. We are planning to distribute the config in its own package once the design tokens are ready and applied to the config. From that perspective, I'd like to hold off on including the preset as part of the `calcite-component`'s package, but we can move forward installing the preset extraction. cc @alisonailea

jrowlingson:
Thanks for taking a :eyes:, I've removed the preset from inclusion in the package.

jcfranco:
@jrowlingson Awesome. I'll help adopt this PR and get it installed this week. Thanks again! 🎉

PR #6708, opened 2023-03-31
test(date-picker): stabilize screenshot test

jcfranco:
**Related Issue:** N/A

## Summary

Sets fixed date on screenshot test that would show a diff every month.


alisonailea (approved):
👍🏻

Issue #6404, opened 2023-01-31
Double clicking the reference element of an open popover causes unexpected placement


ethanbdev:
### Actual Behavior

![image](https://user-images.githubusercontent.com/52869490/215910566-f0b92ab9-51fa-4b5c-9b59-0556375af7b5.png)

Double clicking the reference element of an open popover makes it position weird, going to the top left of the screen

### Expected Behavior

I would expect it to not change position at all, possibly close? Not clear

### Reproduction Sample

https://codepen.io/eborgen/pen/OJwBrPK?editors=1000

### Reproduction Steps

1. Click the button to open the popover
2. Double click the button

### Reproduction Version

1.0.3

### Relevant Info

```
Additional information:
1. This behavior also occurs on other tools.
2. This behavior also occurs on FF/Safari/Edge chromium browsers.

Test case:
https://github.com/ArcGIS/I18N-BYS-TestCases/issues/10286

Configuration:
1. Build: AGOL 2023.1 devext site
2. OS: ENU Win 10
3. Browser: Chrome (109.0.5414.120)
4. Found by date: 1/31/2023

Build Number : 1/31/2023
```

### Impact

Its pretty consistent and easy for a user to accidentally do this

### Esri team

ArcGIS Map Viewer

ethanbdev:
Also repro'd in the calcite components public doc site: https://developers.arcgis.com/calcite-design-system/components/popover/

macandcheese:
Adding on, this issue (or an issue somewhat similar to it), is also evident in the component sample app UI itself, but with Tooltip. 

Quickly double clicking the Action Bar Actions can occasionally cause the Tooltip to reposition and in that case, appear cutoff: 
<img width="375" alt="Screen Shot 2023-01-31 at 8 59 39 PM" src="https://user-images.githubusercontent.com/4733155/215927208-db9aa640-3408-453a-85ca-393408f64d19.png">

Not 100% the same but another weird Floating UI thing, might help diagnose.

amring1:
Can we get a timeline on a fix for this? It's frustrating for many of our analysis components.

geospatialem:
Adding priority labels, which weren't available at the time of issue creation.

Priority is mentioned since it was an issue reported by support, and aiming to get a resolution as soon as possible.

jcfranco:
This is scheduled for the April milestone. For now, setting `overlay-positioning="fixed"` can be used as a workaround.

geospatialem:
[...]

PR #6709, opened 2023-03-31
fix(combobox, dropdown, input-date-picker, popover, tooltip): fix misplaced floating-ui elements when associated-components are closed

jcfranco:
**Related Issue:** #6404

## Summary

This removes a utility added in https://github.com/Esri/calcite-components/pull/5484 that would reset position of floating-UI elements using the absolute strategy to avoid scroll bars when hidden. Unfortunately, the util would lead to undesired position resets in some scenarios (e.g., quick toggling between transitions). Moving forward, it is recommended to use  `overlay-positioning='fixed'`to escaping clipping containers and prevent hidden floating-ui elements from affecting layout/creating scrollbars.


driskull (approved):
Nice. Yeah this revert sounds good 👍

Issue #6547, opened 2023-03-01
Use `composed-offset-position` package instead of embedded ponyfill 🐴


jcfranco:
### Description

https://github.com/Esri/calcite-components/pull/6520 copied a util while the composed offset ponyfill package was in progress. Once stable, we can use https://www.npmjs.com/package/composed-offset-position instead of embedding it [here](https://github.com/Esri/calcite-components/blob/ba8c0688dd92a99e147f0e6434a969fef10ab3e3/src/utils/floating-ui/utils.ts).

### Proposed Advantages

Picking upstream fixes/enhancements and less code on our side.

### Which Component

https://github.com/Esri/calcite-components/blob/ba8c0688dd92a99e147f0e6434a969fef10ab3e3/src/utils/floating-ui.ts#L64

jcfranco:
@geospatialem This got installed in the April milestone, but I forgot to update the issue. Marking as self-verified.

PR #6710, opened 2023-03-31
refactor: use composed-offset-position instead of custom utils

jcfranco:
**Related Issue:** #6547

## Summary

The composed offset position code is now available as a package we can use instead. See https://github.com/floating-ui/floating-ui/pull/2160#issuecomment-1445067026 for more context.


benelan (approved):
:+1:

Issue #6449, opened 2023-02-09
[Shell / Shell Panel / Action Bar / Panel] Tracking Epic for upcoming enhancements


macandcheese:
### Description

This issue serves as a catch-all epic for the proposed and upcoming enhancements to this set of layout components:

In order of need to prevent blockers or duplicate work:

- [x] #5983
- [x] **Needed for others below** #6389
- [x] **Couple with 6447** #6390 
- [x] **Couple with 6390** #6447
- [x] _Independent of above_ #6448
- [x] **Development should start after 6389 is complete* #6388 
- [x] (Needs issue, last to tackle) When slotted in top / bottom, resizable is vertical, not horizontal

Adding here for reference as its part of the design story but ultimately a separate component / implementation:

- [ ] #6391 
- [ ] Once all above complete, deprecate the Shell Center Row component when functionality exists 

### Acceptance Criteria

The above issues are completed and align with proposed design and functionality upon implementation.

Consider some of the comments from https://github.com/Esri/calcite-components/issues/2966#issuecomment-916364665, which include:

> `content-behind` is meant to work with a mapview's [view-padding](https://developers.arcgis.com/javascript/latest/sample-code/view-padding/) so that the view ui (container that holds the on-map widgets) can update without affecting the position of the map.
>
> Not moving the map as panels open/close has historically been one of the requirements of the Map Viewer.
>
> The expectation is that the app developer will update the view-padding in concert with the panels opening and closing.
>
> As it is now, Shell defaults to `content-behind="false"` and the content in the default slot behaves as a center column and responds as such.
>
> As far as `detached`, indeed it does need to work with `content-behind`. This is would be useful in Map Viewer but definitely confusing in general.

### Relevant Info

cc @brittneytewks 

### Which Component

Shell / Shell Panel / Action Bar / Sheet

### Esri team

Calcite (design)

geospatialem:
The epic's issues have been assigned to Matt in the upcoming April release. I will own the epic to track progress as the April milestone approaches.

geospatialem:
Migrating the epic to July, as sheet will be tackled after the May release in July.

macandcheese:
@brittneytewks @geospatialem - I think this can be closed. All the issues have been addressed - the Sheet already has a separate issue and we can address deprecation of Shell Center Row separately.

geospatialem:
Completed for the May milestone.

Issue #6389, opened 2023-01-31
[Shell] Add "panel-top" slot, rename "center-row" to "panel-bottom"


macandcheese:
Description

Following up with more issues from proposals...

Acceptance Criteria

Rename "center-row" slot to "panel-bottom" to be more consistent with existing slot names.
Keep the existing "center-row" for a period of deprecation before removal.
Add a reciprocal "panel-top" slot.
Handle collision between these in the same way "panel-start/end" are.
Relevant Info

In future updates, Shell Panel will be the recommended component for use in these two slots (replacing Shell Center Row).

Which Component

Shell

Example Use Case

Slotted content is desired at the top of a shell component.

Esri team

Calcite (design)

geospatialem:
Verified in `1.3.0-next.4`.

PR #6730, opened 2023-04-04
feat(shell): Add panel-top slot

macandcheese:
**Related Issue:** #6389 

## Summary
Adds a slot for positioning content at the top of the center area of a Shell. As part of the larger set of layout updates (#6449), this will eventually be a use case for Shell Panel.

For now, this will continue to use the Shell Center Row, just like the `center-row` slot. That slot has been "deprecated" (as much as you can deprecate a slot), and a new `panel-bottom` slot has been added.

This has some of the same "swim lane" configuration corners as the current Shell Panel when combining `detached` without `content-behind`, but those exist already and will be alleviated with documentation and further component enhancements.

Screen Shot 2023-04-04 at 12 52 16 PM
Screen Shot 2023-04-04 at 12 51 40 PM
Screen Shot 2023-04-04 at 12 51 30 PM
Screen Shot 2023-04-04 at 12 52 29 PM
Screen Shot 2023-04-04 at 1 21 06 PM
Screen Shot 2023-04-04 at 1 20 43 PM

cc @driskull @asangma


src/components/shell/shell.tsx
```diff
@@ -13,7 +13,9 @@ import { CSS, SLOTS } from "./resources";
  * @slot footer - A slot for adding footer content. This content will be positioned at the bottom of the component.
  * @slot panel-start - A slot for adding the starting `calcite-shell-panel`.
  * @slot panel-end - A slot for adding the ending `calcite-shell-panel`.
- * @slot center-row - A slot for adding content to the center row.
+ * @slot panel-top - A slot for adding the top `calcite-shell-panel`.
+ * @slot panel-bottom - A slot for adding the bottom `calcite-shell-panel`.
+ * @slot center-row - A slot for adding the bottom `calcite-shell-panel`. (Deprecated - use `panel-bottom` instead)
```
driskull, line 18: I think this should start with deprecated for consistency.

```
 * @slot center-row - [Deprecated] use "panel-bottom" instead. A slot for adding content to the center row.
```

driskull (commented):
Looks good! Needs a screenshot and e2e test though.

macandcheese:
@ashetland @SkyeSeitz - I updated the helper that generates story content for Shell Panel, Shell Center Row, so there may be some additional changes in Chromatic in addition to the ones added.

src/components/shell/shell.tsx
macandcheese: Updated - @geospatialem this is the correct way to format "deprecated" slots right? We don't have any other current cases. Would this be a breaking change even though it's a "doc only" deprecation?

geospatialem: We can verify in the next doc update with `1.3.0`, but it seems to be on par with what we had prior to 1.0. 👍🏻 

It looks like any text with `[deprecated]` mentioned receives the deprecation chip on the docs site (lower/upper case character combinations shouldn't matter). 

So `[Deprecated]` seems like a good pattern, and is what we had as of late November 2022 in [shell.txt](https://github.com/Esri/calcite-components/pull/5819/files#diff-b5b8c066968e2720736cedb2d71b8f55fa99bae5fbb8d8b28729cb47212de564) with `[DEPRECATED]` - so pick your poison with lower/upper case combinations.

src/components/shell/shell.tsx
```diff
@@ -13,9 +13,9 @@ import { CSS, SLOTS } from "./resources";
  * @slot footer - A slot for adding footer content. This content will be positioned at the bottom of the component.
  * @slot panel-start - A slot for adding the starting `calcite-shell-panel`.
  * @slot panel-end - A slot for adding the ending `calcite-shell-panel`.
- * @slot panel-top - A slot for adding the top `calcite-shell-panel`.
- * @slot panel-bottom - A slot for adding the bottom `calcite-shell-panel`.
- * @slot center-row - A slot for adding the bottom `calcite-shell-panel`. (Deprecated - use `panel-bottom` instead)
+ * @slot panel-top - A slot for adding the top `calcite-shell-center-row`.
+ * @slot panel-bottom - A slot for adding the bottom `calcite-shell-center-row`.
+ * @slot center-row - [Deprecated] use "panel-bottom" instead. A slot for adding the bottom `calcite-shell-center-row`.
```
geospatialem, line 18: nitpick: Add backticks for API doc consistency to the referenced slot name.

```
[Deprecated] use `"panel-bottom"` instead.
```

macandcheese:
Merging following Chromatic approval: https://www.chromatic.com/pullrequest?appId=6266d45509d7eb004aa254fb&number=6730

Issue #6039, opened 2022-12-14
Calcite-Input-Time-Picker: UI doesn't refresh when updating step property


JonUihlein:
### Actual Behavior

`CalciteInputTimePicker` UI doesn't update after changing the `step` property between different values.
Seconds remain visible (or invisible) in the `input` depending on the previous `step` value until a subsequent interaction. 
Additionally, the time popover never updates to show the correct configuration (separate bug `CalciteTimePicker). 

This bug may also affect other configurations as well (e.g. hours/minutes only) but I didn't test those configurations. 

Using specific values, changing `step` to `1` (from default `60`) should cause the `CalciteInputTimePicker` to show "seconds" in the input. This does not happen until a subsequent interaction (clicking the input). Additionally, the time popover never updates to show seconds, even on subsequent interactions (open/close).

Two different bugs to address:
The `<calcite-input>` used inside `CalciteInputTimePicker` does not update after the `step` value changes, until an interaction. I think this property may need a watcher to trigger a UI update.

The `CalciteTimePicker` contained in the `<calcite-popover>` inside `CalciteInputTimePicker` does not update after the `step` value on the parent component changes. This is also reproducible using a stand-alone `CalciteTimePicker`, but it may need additional logic to react to `step` changes on the parent component. 

### Expected Behavior

Seconds/minutes should be visible/invisible depending on `step`. 

Time Picker Popover should correctly reflect current `step` value. 

### Reproduction Sample
`CalciteInputTimePicker` (button):
https://jsbin.com/fisopopavi/edit?html,output

`CalciteTimePicker` (button):
https://jsbin.com/piqecanefi/edit?html,output

(old) `CalciteInputTimePicker` (setTimeout):
https://jsbin.com/pupiviseza/edit?html,output

### Reproduction Steps

1) Create `CalciteInputTimePicker`
2) Update `CalciteInputTimePicker.step` from 60 (default) to 1
3) Notice the UI does not update
4) Click the input to trigger the popover
5) Notice the popover does not show seconds
6) Click away
7) Input should now display time with seconds
8) Click the input to trigger the popover
9) Notice the popover still does not show seconds

### Reproduction Version

1.0.0-next.665

### Impact

Noticed this while integrating `CalciteInputDatePicker` and `CalciteInputTimePicker` into our FeatureForm widget. 

### Esri team

ArcGIS API for JavaScript

geospatialem:
[...]

PR #6731, opened 2023-04-05
fix(input-time-picker, time-picker): render when input-time-picker or time-picker's step property changes

jcfranco:
**Related Issue:** #6039 

## Summary

The `step` property toggles the display of the seconds in the UI. This adds some watchers to trigger rendering when the prop is updated.


eriklharper (approved):
Thanks for jumping in on this one! :)

PR #6732, opened 2023-04-05
feat: allow disabled elements to emit pointer events without triggering activation

jcfranco:
**Related Issue:** #5318 

## Summary

This tweaks the InteractiveComponent pattern to block all pointer/mouse on children (light/shadow) within host components, and to ignore clicks on the host to allow pointer events to be emitted while not activating on click.

**Note**: in order to prevent focus moving to the host when clicked, the `pointerdown` event has to be canceled in the capture phase.


driskull:
@jcfranco what about components like the `calcite-action` where a user sets a click event on their own? They will have to check for the disabled property right?

jcfranco:
> what about components like the calcite-action where a user sets a click event on their own? They will have to check for the disabled property right?

Yeah, great catch. [The spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#enabling-and-disabling-form-controls%3A-the-disabled-attribute) does mention preventing clicks on disabled form elements:

> A form control that is [disabled](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#concept-fe-disabled) must prevent any [click](https://w3c.github.io/uievents/#event-type-click) events that are [queued](https://html.spec.whatwg.org/multipage/webappapis.html#queue-a-task) on the [user interaction task source](https://html.spec.whatwg.org/multipage/webappapis.html#user-interaction-task-source) from being dispatched on the element.

Based on ☝️, I think we could stop propagation of clicks, but let me dig a bit.

jcfranco:
Updated the PR to be closer to https://github.com/web-platform-tests/wpt/blob/master/html/semantics/disabled-elements/event-propagate-disabled.tentative.html with the exception of allowing mouse/pointer events to reach children of disabled components. We can revisit this later if needed as it will require more time and effort to achieve. It'll also allow us time to gauge the need for this.

src/utils/interactive.ts
```diff
@@ -22,6 +22,13 @@ function onPointerDown(event: PointerEvent): void {
   event.preventDefault();
 }
 
+const nonBubblingWhenDisabledMouseEvents = ["mousedown", "mouseup", "click"];
```
driskull, line 25: Do we need to worry about pointer events as well?

driskull (commented):
Looks good. I think the disabled tests need to be updated to account for no more click events.

src/utils/interactive.ts
jcfranco: Not according to the [intended spec changes](https://github.com/whatwg/html/issues/5886).

driskull (approved):
Approved! 👍 💯

Issue #6448, opened 2023-02-09
[Panel] Add Action Bar slot


macandcheese:
### Description

As proposed, the Panel gains an "action bar slot" beneath the rendered header to provide familiar and consistent placement of actions in a workflow setting. This will enable the migration path forward from using Shell Center Row to using Shell Panel in the "panel-bottom" and "panel-top" (upcoming) shell slots.

### Acceptance Criteria

The Panel has a positioned action bar that is located beneath the rendered header / slotted header content. Action bar within this slot is set to horizontal.

### Which Component

Panel, Action Bar

### Example Use Case

cc @asangma, @mitc7862

MV - Analysis, etc. 

### Esri team

Calcite (design)

brittneytewks:
@driskull can you please add an estimate to this issue? thank you!

ethanbdev:
Watching 👀

geospatialem:
Verified on `1.3.0-next.4`.

![image](https://user-images.githubusercontent.com/5023024/231895474-ad7b2ffb-b377-4a76-9674-eca8580c5af1.png)

driskull:
@geospatialem @macandcheese can you create a new issue for adding an action bar slot to the flow item?

driskull:
nevermind. created https://github.com/Esri/calcite-components/issues/6886

PR #6738, opened 2023-04-05
feat(panel): Add slot for an action-bar component.

driskull:
**Related Issue:** #6448

## Summary

Panel gains an "action bar slot" beneath the rendered header to provide familiar and consistent placement of actions in a workflow setting.


asangma:
@driskull One note. Does content in the action-bar slot stay sticky along with the header node? Seems like we would want that behavior.

driskull:
@asangma no, that was not defined in the requirements. It would need to be in the header then and not below.

macandcheese:
I think it's working as expected currently: 


https://user-images.githubusercontent.com/4733155/230679518-03a73fda-fd2c-4c0e-9eab-5f5c9e00da16.mov

The only thing that sticks out is part of Action Bar - the 1px border-right set on the `bottom-actions` slot of the Action Bar. Maybe we can clean that up separately.

Re: design - this really doesn't have a visual of its own, the Action Bar is doing all of the UI work here aside from the bottom border on the slot, which seems correct to me.

macandcheese (approved):
Looks good, seems to behave as expected. Only visual thing of note is the border-right on the `bottom-actions` slot in Action Bar. Maybe conditionally don't set that based on the Action Bar layout?

driskull:
> Only visual thing of note is the border-right on the bottom-actions slot in Action Bar.

@macandcheese that might not be related to this PR. I'm not setting any borders on the action-bar. Can you create a new issue for some CSS changes?

driskull:
Since the action-bar goes horizontal now, should we rename "botttom-actions" to "end-actions"? Might want to reevaluate the names

macandcheese:
> Since the action-bar goes horizontal now, should we rename "botttom-actions" to "end-actions"? Might want to reevaluate the names

Ha, yep see above issue. Agreed. I'll open issues for the border and slot name change on Action Bar.

src/components/panel/panel.stories.ts
```diff
@@ -162,3 +162,15 @@ export const darkModeRTL_TestOnly = (): string =>
   );
 
 darkModeRTL_TestOnly.parameters = { modes: modesDarkDefault };
+
+export const withActionBar_TestOnly = (): string => `<calcite-panel height-scale="s">
```
jcfranco, line 166: Can you use the `html` formatting helper so prettier does its thang?

src/components/panel/panel.tsx
```diff
@@ -358,6 +371,14 @@ export class Panel
     ) : null;
   }
 
+  renderActionBar(): VNode {
+    return (
+      <div class={CSS.actionBarContainer} hidden={!this.hasActionBar} key="action-bar">
```
jcfranco, line 376: If this is always rendered, `key` isn't needed.

jcfranco (approved):
🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥
🎥💥💥💥💥🎥🎥💥💥💥🎥💥💥💥💥💥🎥💥💥💥🎥🎥💥💥🎥🎥💥🎥🎥🎥💥🎥
🎥💥🎥🎥💥🎥💥🎥🎥🎥🎥🎥🎥💥🎥🎥🎥🎥💥🎥🎥💥🎥🎥💥🎥💥💥🎥🎥💥🎥
🎥💥💥💥💥🎥💥🎥🎥🎥🎥🎥🎥💥🎥🎥🎥🎥💥🎥🎥💥🎥🎥💥🎥💥🎥💥🎥💥🎥
🎥💥🎥🎥💥🎥💥🎥🎥🎥🎥🎥🎥💥🎥🎥🎥🎥💥🎥🎥💥🎥🎥💥🎥💥🎥🎥💥💥🎥
🎥💥🎥🎥💥🎥🎥💥💥💥🎥🎥🎥💥🎥🎥🎥💥💥💥🎥🎥💥💥🎥🎥💥🎥🎥🎥💥🎥
🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥

Issue #6390, opened 2023-01-31
[Action Group] Infer Action Group `layout` from parent Action Bar or Action Pad


macandcheese:
### Description

Currently, the Action Group has a `layout` property, as does its containing Action Bar or Action Pad. This can lead to situations where the parent is in one mode, and Action Group in another, creating undesirable layout. 

### Acceptance Criteria

If possible, infer the `layout` of the Action Group from the parent and respond to changes made to the parent. This could be set by the parent. Remove the public property from Action Group as it shouldn't be used on its own (verify this ... maybe it needs to remain but still be set by parent in most cases).

### Which Component

Action Group (Action Pad / Action Bar)

### Esri team

Calcite (design)

driskull:
> This could be set by the parent.

Yes, the parent should manage it and set it. We don't want to look back up the DOM.

brittneytewks:
@driskull as we're working through the milestone, can you add a priority and estimate to this issue?

geospatialem:
Dev work will be initiated once https://github.com/Esri/calcite-components/issues/6389 is complete, and should be coordinated with #6447.

geospatialem:
Verified in `1.3.0-next.4`.

geospatialem:
future breaking change for the `action-group`s `layout` prop

PR #6739, opened 2023-04-05
feat(action-bar, action-pad): Set layout property on child action-group elements.

driskull:
**Related Issue:** #6390

## Summary

Deprecates the layout property on the `action-group` element. The parent component of the child `action-group` elements will set the `layout` property.

### DIFF
--- a/src/components/action-group/action-group.tsx
+++ b/src/components/action-group/action-group.tsx
@@ -60,7 +60,7 @@ export class ActionGroup
   /**
    * Indicates the layout of the component.
    *
-   * @deprecated Use the `layout` property on the comoponent's parent component instead.
+   * @deprecated Use the `layout` property on the component's parent component instead.
    */
   @Prop({ reflect: true }) layout: Layout = "vertical";
 
--- a/src/components/action-group/action-group.tsx
+++ b/src/components/action-group/action-group.tsx
@@ -60,7 +60,7 @@ export class ActionGroup
   /**
    * Indicates the layout of the component.
    *
-   * @deprecated Use the `layout` property on the component's parent component instead.
+   * @deprecated Use the `layout` property on the component's parent instead.
    */
   @Prop({ reflect: true }) layout: Layout = "vertical";
 
--- a/src/components/action-group/action-group.tsx
+++ b/src/components/action-group/action-group.tsx
@@ -59,6 +59,8 @@ export class ActionGroup
 
   /**
    * Indicates the layout of the component.
+   *
+   * @deprecated Use the `layout` property on the comoponent's parent component instead.
    */
   @Prop({ reflect: true }) layout: Layout = "vertical";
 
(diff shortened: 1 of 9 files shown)


src/components/action-group/action-group.tsx
```diff
@@ -59,6 +59,8 @@ export class ActionGroup
 
   /**
    * Indicates the layout of the component.
+   *
+   * @deprecated Use the `layout` property on the comoponent's parent component instead.
```
macandcheese, line 63: Typo: "comoponent's"

macandcheese (approved):
This will be a great QoL improvement, thanks!

src/components/action-group/action-group.tsx
```diff
@@ -59,6 +59,8 @@ export class ActionGroup
 
   /**
    * Indicates the layout of the component.
+   *
+   * @deprecated Use the `layout` property on the component's parent component instead.
```
jcfranco, line 63: Suggestion: ``Use the `layout` property on the component's parent instead.``

cc @geospatialem

src/components/action-pad/action-pad.tsx
```diff
@@ -125,6 +131,12 @@ export class ActionPad
 
   @Element() el: HTMLCalciteActionPadElement;
 
+  mutationObserver = createObserver("mutation", () => {
+    const actionGroups = Array.from(this.el.querySelectorAll("calcite-action-group"));
+
+    this.setGroupLayout(actionGroups);
```
jcfranco, line 137: Nitpick: this could be inlined. Applies to similar functions.

src/components/action-pad/action-pad.tsx
```diff
@@ -207,6 +219,26 @@ export class ActionPad
     this.expandToggleEl = el;
   };
 
+  updateGroups = (): void => {
```
jcfranco, line 222: I think this can be a class method and not need to store a bound function. Applies to similar functions.

jcfranco (approved):
🎉

jcfranco:
Forgot to mention, can you update the PR title to include the deprecation? We want these type of changes to show up in the changelog as well.

src/components/action-group/action-group.tsx
```diff
@@ -59,6 +59,8 @@ export class ActionGroup
 
   /**
    * Indicates the layout of the component.
+   *
+   * @deprecated Use the `layout` property on the component's parent instead.
```
macandcheese, line 63: There is the "grid" mode on the action-group, which currently isn't on the parent for some reason. I'm not really familiar with how those modes interact, but should "grid" be added to the parent `action-bar` layout as a value?

driskull: I think the grid layout is just for the action-pad so I think as it is makes sense.

Issue #2620, opened 2021-07-23
Enhancement: Add `closable` functionality to calcite-tabs-title


am-maneaters:
### Description
Add action button to the calcite-tab-title component to allow for closing tabs

*Obviously very rough implementation below
<img width="552" alt="image" src="https://user-images.githubusercontent.com/50883428/126845046-2e9caf15-c592-4e11-850b-d80995b035d3.png">

### Acceptance Criteria 
Include action button in tabs with custom icons and custom onClick event

#### Which Component

calcite-tab-title

#### Example Use Case

Allow users to close opened tabs using the action button in the tab titles

AdamWMoqrane:
Hello! XI Design would definitely have a need for this issue with the Knowledge Studio project we are working on. Prioritizing this would help many out!

RSantosGIS:
Just wanted to bump this requirement from the ArcGIS Knowledge Studio team again following our conversation with Adam T. yesterday.

macandcheese:
This was mentioned in a Community post as well: https://community.esri.com/t5/calcite-design-system-questions/calcite-tab-dismissible-closable/m-p/1238589#M278

lkoumis:
Similar to the attribute table in web appbuilder, it would be a great feature to be able to add a tab containing the  feature table of a loaded layer and then be able to remove the tab when the layer is removed from the map.

macandcheese:
@geospatialem @jcfranco - this is high-impact for Knowledge Studio and other teams.

geospatialem:
Should consider with the efforts of https://github.com/Esri/calcite-components/issues/2721, anticipated for the May 2023 milestone.

ashetland:
Quick question about behavior on this as we work out details. I would expect [closable] to be globally set, e.i. all tabs are closable or not-closable. This way we would avoid a confusing possible state of mixed closable and non-closable tabs. Does this thinking jive with expectations?

macandcheese:
I’d imagine it would be more flexible to set per tab. 

Perhaps the tabs represent tables or sheets, and the app has a minimum number to display, or for the same workflow, some tabs are locked due to permission (and denoted as such with icon-start/end).

Those are just two use cases that come to mind but I’d expect some other similar needs. 

Some related questions:
Should we allow the last tab to be closed?

Should we consider allowing tabs to be re-ordered (log as separate enhancement). If tabs can be closed, we can imagine they may be used in workflows where users could reorder the representative tabs.

[...]

PR #6740, opened 2023-04-06
feat(tab-nav): adds optional closable functionality to individual `tab-titles`

Elijbet:
**Related Issue:** #2620

## Summary
`handleTabTitleClose` sequence is as follows:
> if `closed` item is `selected`, fall back on the next
> unless it's the last item, which falls back on previous
> closing non-selected items doesn't affect the `selected` tab
> last remaining item becomes `disabled`, if `closable`

- Added `closed` property
- Added `closable` property
- Added `calciteInternalTabsClose` event
- Added `calciteTabsClose` event
- Added t9n messages for `Close` text.
- Updated styling per new Figma specs
- Added tests

`handleTabTitleClose` keeps track of `visibleTabTitlesIndices` instead of having Dom queries filter out `hidden`. This helps to sync `tabs` with `tab-titles` using their corresponding `id`s set on the original render. `visibleTabIndeces` is an array of `id`s of the non-hidden `tab-titles`. When `tab-title` is `closed` this `id` is used to navigate to the next visible item. This makes figuring out what the next item should be when closing out of order is now more straightforward.

Tests coverage for 

- simple closing behavior 
- closing sequence behavior:
> - when closing `tab-titles` in sequence 1 (first selected) through 4, `tab-title` and corresponding `tab` become hidden, and selection fallback is the next tab
> - reverse fallback: when closing `tab-titles` in sequence 4 (last selected) through 1, `tab-title` and corresponding `tab` become hidden, and selection fallback is the previous tab
> - closing an unselected `tab-title` does not deselect the current selection
> - 2 additional cases: works with a randomized closing sequence with mixed selected and not


src/components/tab-nav/tab-nav.tsx
```diff
@@ -3,6 +3,7 @@ import {
   Element,
   Event,
   EventEmitter,
+  forceUpdate,
```
driskull, line 6: Instead of calling forceUpdate, is there an internal prop that could be added on the element or something in order to more naturally trigger a rerender?

src/components/tab-nav/tab-nav.tsx
```diff
@@ -234,9 +242,17 @@ export class TabNav {
     event.preventDefault();
   }
 
-  @Listen("calciteTabsActivate") activateTabHandler(event: CustomEvent<void>): void {
+  @Listen("calciteTabsActivate")
+  activateTabHandler(event: CustomEvent<void>): void {
     this.calciteTabChange.emit();
+    event.stopPropagation();
+    event.preventDefault();
```
driskull, line 249: Is there a reason preventDefault is called here? I don't think there's a default action for tabsactivate right?

src/components/tab-title/tab-title.scss
```diff
@@ -154,25 +194,25 @@
...
 :host([bordered][scale="l"]) {
-  .container {
-    @apply px-4;
-  }
   .icon-present {
     padding-block: 11px;
   }
 }
 
+// hover
```
driskull, line 207: I think we can remove the hover comment.

src/components/tab-title/tab-title.tsx
```diff
@@ -130,6 +175,10 @@ export class TabTitle implements InteractiveComponent {
     }
   }
 
+  componentWillUpdate(): void {
+    this.isFirstLoad = false;
```
driskull, line 179: There might be a better pattern to use here. Would a mutation observer help here?  I'm just wondering why a first load is special in this case?

src/components/tab-title/tab-title.tsx
```diff
@@ -215,13 +289,14 @@ export class TabTitle implements InteractiveComponent {
...
       return;
     }
-
-    if (this.tab) {
-      this.selected = this.tab === event.detail.tab;
-    } else {
-      this.getTabIndex().then((index) => {
-        this.selected = index === event.detail.tab;
-      });
+    if (!this.isFirstLoad) {
```
driskull, line 292: This is kind of an unusual pattern. Would a mutation observer help here?

src/components/tab-title/tab-title.tsx
```diff
@@ -355,12 +449,25 @@ export class TabTitle implements InteractiveComponent {
...
   }
 
+  //--------------------------------------------------------------------------
+  //
+  //  Private Methods
+  //
+  //--------------------------------------------------------------------------
+
+  closeClickHandler = (event: MouseEvent): void => {
+    event.stopPropagation();
```
driskull, line 459: Do we need to stop propagation here? Seems like we should allow clicks to bubble here.

driskull (commented):
Looking good @Elijbet 👍 😎 

(◠﹏◠)

src/components/tab-nav/tab-nav.tsx
Elijbet: It fails to update `tabs` in a few places when `tabTitle` changes due to the selection moving after `tab-title` closes. I'll be following this up with a refactor issue on all tab-related components, as they are a bit more complicated than might need to be. And to address inconsistent variable naming. I'll add `forceUpdate` as an item to address.

src/components/tab-nav/tab-nav.tsx
```diff
@@ -370,9 +385,49 @@ export class TabNav {
...
   get enabledTabTitles(): HTMLCalciteTabTitleElement[] {
-    return filterDirectChildren<HTMLCalciteTabTitleElement>(
+    const directEnabled = filterDirectChildren<HTMLCalciteTabTitleElement>(
       this.el,
       "calcite-tab-title:not([disabled])"
     );
+    return directEnabled.filter((tabTitle) => !tabTitle.closed);
+  }
+
+  handleTabTitleClose(closedTabTitleEl: HTMLCalciteTabTitleElement): void {
```
macandcheese, line 395: Currently, if you close a tab more than one tab "away" from the currently selected tab, the selected tab is shifted. I would not expect the selected tab to change unless I close it or navigate to another, regardless of distance from selected item.

In this example I'd expect "Bicycles" in the bottom example to remain selected when I close the first tab.

https://github.com/Esri/calcite-components/assets/4733155/896df332-96a6-4a06-928e-16d05cbd839a

macandcheese: Similarly - in this example, I close "Bicycles" and the selected tab is shifted to "Aircrafts" - when I think we'd want it to stay on "Watercraft"

https://github.com/Esri/calcite-components/assets/4733155/e57a0f8d-e9b6-4453-b591-7c9a598e0a84

macandcheese:
Found it possible to enter a state with nothing selected - maybe due to the proximity of each closed tab to the selected one?

https://github.com/Esri/calcite-components/assets/4733155/8294523c-a4c9-4f99-aa4d-a657f9a21307

src/components/tab-nav/tab-nav.tsx
Elijbet: Cool thanks. I know why that is happening. Pushing a fix and a corresponding test.

Elijbet: Pushed. Let me know if it works now.

src/components/tab-title/tab-title.scss
```diff
@@ -17,12 +17,20 @@
...
-  @apply focus-base;
+:host([closed]) {
+  @apply hidden;
 }
 
-:host(:focus) .container {
-  @apply focus-inset;
+// focus styles
+.tab-title-close,
+.container {
```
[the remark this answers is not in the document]
Elijbet: Pushed, if you could check this as well por favor.

src/components/tab-title/tab-title.scss
```diff
@@ -91,18 +102,40 @@
...
+      h-full
+      appearance-none
+      overflow-hidden
+      border-none
+      justify-center
+      items-center;
+}
+
+.close-button {
+  @include tab-title-element-base;
```
jcfranco, line 133: This is the only place where `tab-title-element-base` is used, so let's inline its contents here.

src/components/tab-title/tab-title.tsx
```diff
@@ -59,6 +74,12 @@ export class TabTitle implements InteractiveComponent {
     }
   }
 
+  /** When `true`, a close button is added to the component. */
+  @Prop({ reflect: true }) closable = false;
+
+  /** When `true`, does not display or position the component. */
```
jcfranco, line 80: +@geospatialem for editorial review.

src/components/tab-title/tab-title.tsx
```diff
@@ -232,7 +300,6 @@ export class TabTitle implements InteractiveComponent {
     if (this.disabled) {
       return;
     }
-
```
jcfranco, line 235: For readability, let's restore this newline (the return and emitting have more visual separation with the newline).

Applies to similar changes.

src/components/tab-title/tab-title.tsx
```diff
@@ -369,6 +468,15 @@ export class TabTitle implements InteractiveComponent {
   /** determine if there is slotted text for styling purposes */
   @State() hasText = false;
 
+  @State() effectiveLocale: "";
```
jcfranco, line 471: Nitpick: can these new props be placed in alphabetical order?

src/components/tab-title/tab-title.e2e.ts
```diff
@@ -1,14 +1,48 @@
...
+      </calcite-tab>
+      <calcite-tab id="bikingTab">
+        <calcite-notice icon="biking" open>
+          <div slot="message">Healthy and gets you from point A to B</div>
+        </calcite-notice>
+      </calcite-tab>
+    </calcite-tabs>
+  `;
   const iconStartHtml = `calcite-tab-title >>> .${CSS.titleIcon}.${CSS.iconStart}`;
   const iconEndHtml = `calcite-tab-title >>> .${CSS.titleIcon}.${CSS.iconEnd}`;
```
jcfranco, line 40: Nitpick: These should be renamed to indicate they are selectors and not HTML.

src/components/tab-title/tab-title.e2e.ts
```diff
@@ -50,6 +84,241 @@ describe("calcite-tab-title", () => {
...
+      expect(containerElOne).toHaveAttribute(HYDRATED_ATTR);
+
+      await closeOne.click();
+      await page.waitForChanges();
+
+      containerElOne = await page.find(`calcite-tab-title[id='one']>>> .${CSS.container}`);
+      expect(await containerElOne.getProperty("hidden")).toBe(true);
+
+      const closeTwo = await page.find(`calcite-tab-title[id='two'] >>> .${CSS.closeButton}`);
+      expect(await closeTwo.getProperty("closable")).not.toBe(true);
```
jcfranco, line 127: Nitpick: for readability, use `toBe(false)` vs `.not.toBe(true)`.

src/components/tab-title/tab-title.e2e.ts
```diff
@@ -50,6 +84,241 @@ describe("calcite-tab-title", () => {
...
+    it(`when closing tab-titles in sequence 1 (first selected) through 4, 
+        tab-title and corresponding tab become hidden, 
+        and selection fallback is the next tab`, async () => {
+      await closeTabsInSequenceOfGivenArrayOfIds(["embark", "car", "plane", "biking"]);
+    });
+
+    it(`when closing tab-titles in sequence 4 (last selected) through 1, 
+        tab-title and corresponding tab become hidden, 
+        and selection fallback is the previous tab`, async () => {
+      const arrayOfReversedIds = ["embark", "car", "plane", "biking"].reverse();
```
jcfranco, line 186: Is there a more expressive name for this? What do these IDs represent? Applies to other ID arrays.

src/components/tab-title/tab-title.e2e.ts
```diff
@@ -50,6 +84,241 @@ describe("calcite-tab-title", () => {
...
+    });
+  });
+
+  describe("closing sequence", () => {
+    let page: E2EPage;
+
+    let matchingTabEl: E2EElement;
+    let close: E2EElement;
+
+    const closeTabsInSequenceOfGivenArrayOfIds = async (arrayOfIds: string[]): Promise<void> => {
```
jcfranco, line 137: Let's make this a function that takes a page as an argument as well as an array of IDs.

Also, can we rename `arrayOfIds` to be more expressive? Once this is done, I think we could drop `OfGivenArrayOfIds` from the function name as the function signature provides context.

src/components/tab-title/tab-title.e2e.ts
```diff
@@ -50,6 +84,241 @@ describe("calcite-tab-title", () => {
...
+      const closeTwo = await page.find(`calcite-tab-title[id='two'] >>> .${CSS.closeButton}`);
+      expect(await closeTwo.getProperty("closable")).not.toBe(true);
+    });
+  });
+
+  describe("closing sequence", () => {
+    let page: E2EPage;
+
+    let matchingTabEl: E2EElement;
+    let close: E2EElement;
```
jcfranco, line 135: Is there a more expressive name for this?

src/components/tab-title/tab-title.e2e.ts
```diff
@@ -108,49 +377,72 @@ describe("calcite-tab-title", () => {
...
+      it("should inherit small scale from tab-nav", async () => {
+        await page.setContent(
+          html`
+            <calcite-tab-nav scale="s">
+              <calcite-tab-title selected>Tab Title</calcite-tab-title>
+              <calcite-tab-title>Tab 2 Title</calcite-tab-title>
+            </calcite-tab-nav>
+          `
+        );
+        await page.waitForChanges();
```
jcfranco, line 415: I think you don't need to call `await page.waitForChanges();` after setting content.

Applies to similar calls.

src/components/tab-title/tab-title.e2e.ts
```diff
@@ -108,49 +377,72 @@ describe("calcite-tab-title", () => {
...
+
+        expect(await page.find("calcite-tab-nav")).toEqualAttribute("scale", "s");
+        expect(await page.find("calcite-tab-title")).toEqualAttribute("scale", "s");
+
+        const content = await page.find(`calcite-tab-title >>> .${CSS.content}`);
```
jcfranco, line 420: Can you reorder the tests in this describe block to have the setup first and then the assertions (`expect` calls) at the end (similar to the grouping from the previous version)?

src/components/tab-title/tab-title.e2e.ts
```diff
@@ -1,14 +1,48 @@
...
+import { newE2EPage, E2EPage, E2EElement } from "@stencil/core/testing";
 import { disabled, HYDRATED_ATTR, renders, hidden } from "../../tests/commonTests";
+import { html } from "../../../support/formatting";
 import { CSS } from "./resources";
 
 describe("calcite-tab-title", () => {
   const tabTitleHtml = "<calcite-tab-title></calcite-tab-title>";
+  const tabTitleClosableHtml = "<calcite-tab-title closable></calcite-tab-title>";
+  const multiTabTitleClosableHtml = `
+    <calcite-tabs bordered position="top">
```
jcfranco, line 10: I think it'd be beneficial to update the markup to use some sort of numbering in the IDs and/or content to make it easy to track the order in the tests below.

src/components/tab-nav/tab-nav.tsx
```diff
@@ -370,9 +375,42 @@ export class TabNav {
...
+    const { tabTitles } = this;
+
+    const visibleTabTitlesIndices = tabTitles.reduce((acc, curr, i) => {
```
jcfranco, line 388: I think this method can be simplified a bit further, but let's sync up on this tomorrow.

jcfranco (approved):
Awesome!

src/components/tab-title/tab-title.tsx
```diff
@@ -176,20 +222,42 @@ export class TabTitle implements InteractiveComponent {
...
+    const { closable, messages } = this;
+
+    return closable ? (
+      <button
+        aria-label={messages.close}
+        class={CSS.closeButton}
+        disabled={false}
+        key={CSS.closeButton}
+        onClick={this.closeClickHandler}
+        onKeyDown={this.closeKeyHandler}
```
driskull, line 252: is the keydown needed? I thought native button elements don't need keydown for enter/space support.

src/components/tab-title/tab-title.e2e.ts
jcfranco: Cool. This is better, but can be improved by also renaming the element reference variables below. 

For comparison, look at the following: https://github.com/Esri/calcite-components/blob/ee5444c6fd986c7ea87c0daf56054f64c13d525d/src/components/combobox/combobox.e2e.ts#L945-L968 You can see it's easier to visualize the order in the assertions without having to look at the HTML or selectors used.

src/components/tab-title/tab-title.tsx
geospatialem: Is the behavior that the component won't be displayed? If so, perhaps similar to [List item](https://github.com/Esri/calcite-components/blob/master/src/components/list-item/list-item.tsx#L81-L82)?:

> When `true`, hides the component.

PR #6741, opened 2023-04-06
chore: switch unpkg CDN to jsdelivr

jcfranco:
**Related Issue:** N/A

## Summary

Based on jsdelivr's [about page](https://www.jsdelivr.com/about), [unpkg comparison page](https://www.jsdelivr.com/unpkg) and also recent experiences, jsdelivr seems to be a more reliable, fast and secure option than unpkg.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
   "jsnext:main": "dist/esm/index.js",
   "types": "dist/types/components.d.ts",
   "type": "module",
+  "jsdelivr": "dist/calcite/calcite.js",
   "unpkg": "dist/calcite/calcite.js",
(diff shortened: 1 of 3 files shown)


driskull (approved):
👍 

Should we still put a disclaimer where we use jsdlvr saying to use our CDN? or is jsdlvr production quality?

jcfranco:
> Should we still put a disclaimer where we use jsdlvr saying to use our CDN? or is jsdlvr production quality?

From what I can gather, it's production quality. cc @benelan @dasa

benelan (approved):
:+1:

Issue #6742, opened 2023-04-06
Panel - close button should always be in the corner and not before the menu actions.


driskull:
### Actual Behavior

A closable panel has the close button left of the actions menu.

https://developers.arcgis.com/calcite-design-system/components/panel/

### Expected Behavior

The close button is all the way to the corner.

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/panel/

### Reproduction Steps

1. Go to https://developers.arcgis.com/calcite-design-system/components/panel/
2. Set the panel to closable
3. view

<img width="456" alt="image" src="https://user-images.githubusercontent.com/1231455/230456453-dee77240-993a-42e1-892b-ef2e83d9d94c.png">

### Reproduction Version

1.2.x

### Priority impact

p2 - want for current milestone

### Esri team

N/A

driskull:
@geospatialem can we get this triaged? I can take it.

asangma:
➕

PR #6746, opened 2023-04-06
fix(panel): ensure close button is placed in the corner of the panel header.

driskull:
**Related Issue:** #6742

## Summary

Ensures the close button is always placed in the far corner of the panel header.


src/components/panel/panel.tsx
```diff
@@ -387,12 +387,13 @@ export class Panel
...
 
   renderHeaderActionsEnd(): VNode {
-    const { close, hasEndActions, messages, closable } = this;
+    const { close, hasEndActions, messages, closable, hasMenuItems } = this;
     const text = messages.close;
 
     const closableNode = closable ? (
       <calcite-action
         aria-label={text}
+        data-test="close"
```
jcfranco, line 396: Can you create a follow-up issue to look into having these data attributes removed from production builds? I think we may be able to do that by using Stencil's `Build.isTesting` flag.

jcfranco (approved):
↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️
↗️❌↗️↗️↗️❌↗️❌❌❌↗️↗️❌❌❌↗️❌❌❌❌↗️❌↗️
↗️❌❌↗️↗️❌↗️↗️❌↗️↗️❌↗️↗️↗️↗️❌↗️↗️↗️↗️❌↗️
↗️❌↗️❌↗️❌↗️↗️❌↗️↗️❌↗️↗️↗️↗️❌❌❌↗️↗️❌↗️
↗️❌↗️↗️❌❌↗️↗️❌↗️↗️❌↗️↗️↗️↗️❌↗️↗️↗️↗️↗️↗️
↗️❌↗️↗️↗️❌↗️❌❌❌↗️↗️❌❌❌↗️❌❌❌❌↗️❌↗️
↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️↗️

src/components/panel/panel.tsx
driskull: Can I just strip it out instead?

driskull: Might be a stencil bug, but I tried `data-test={Build.isTesting ? "close" : undefined}` and the test failed.

driskull: https://github.com/Esri/calcite-components/issues/6760

Issue #5903, opened 2022-12-05
dropdown button with width="full" wider than containing block


AdelheidF:
### Actual Behavior

![image](https://user-images.githubusercontent.com/44481292/205768062-495157a1-7af7-4719-8d64-68656da1b389.png)

The dropdown button is wider than its containing block if the button text has no dashes or spaces.
Button uses `width="full"`.

### Expected Behavior

text is wrapped 

### Reproduction Sample

https://codepen.io/afreitag/pen/GRGPdwQ

### Reproduction Steps

open codepen and observe. Red box displays block dimensions.

### Reproduction Version

next.657

### Regression?

Don't think so

### Esri team

ArcGIS Map Viewer

geospatialem:
Applicable in the case mentioned above where text has no breaks, such as "`BirdObservationCommentBirdObservationComment`" and where "`_`" characters are present, such as "`Bird_Observation_Comment_Bird_Observation_Comment`".

![image](https://user-images.githubusercontent.com/5023024/205976751-55dd0600-0a27-4a1a-b234-d1ea68c35eba.png)

AdelheidF:
will this be part of 1.0?

benelan:
How smart do you want this to be; are you expecting it to understand word breaks such as, capitalization, underscores, periods, etc.

>BirdObservationCommentBirdObservation
>Comment

Or just wrap on the overflow
>BirdObservationCommentBirdObservationCom
>ment

AdelheidF:
> Or just wrap on the overflow

I think this is fine. Don't you do this already in other places like this?

macandcheese:
FWIW, it breaks in a different way when the button is not the dropdown-trigger:
<img width="421" alt="Screen Shot 2022-12-08 at 1 32 07 PM" src="https://user-images.githubusercontent.com/4733155/206572164-831881ef-b3a8-4ccf-822b-3d1446acc134.png">

geospatialem:
@AdelheidF We're putting this in a list for Franco when he returns, but due to staffing, this one might land in time for the February release. Will post an update a few weeks into 2023.

geospatialem:
Seeking design expertise for the `dropdown` button, if wrapping and/or truncating the text should be considered.

Elijbet:
> Seeking design expertise for the `dropdown` button, if wrapping and/or truncating the text should be considered.

[...]

PR #6747, opened 2023-04-06
fix(dropdown): trigger should break words when overflowing container.

driskull:
**Related Issue:** #5903

## Summary

Applies css word break to the container of the trigger slotted element. If it overflows its bounds, words will be broken.


Elijbet (commented):
Great! Per update here lets merge it as a temporary solution until the other more advanced one lands: https://github.com/Esri/calcite-components/issues/5903#issuecomment-1500570630 Thanks @driskull :)

Elijbet (approved):

Issue #6384, opened 2023-01-30
Combobox Item Groups no longer change indentation levels when nested


ashetland:
### Actual Behavior
When nesting Combobox Item Groups, there is no visual indication of nesting.

### Expected Behavior
Nested groups should increase in indentation levels, matching the behavior of Combobox Items.

### Reproduction Sample
https://codepen.io/ashetland/pen/JjBaOga?editors=100

### Reproduction Steps
Open the codepen.
Note that groups are nested in code, but there is no visual representation of the nesting.
See this [beta.97 codepen](https://codepen.io/ashetland/pen/dyjqZLw?editors=1000) for comparison.

### Reproduction Version
1.0.3

### Relevant Info
No response

### Regression?
beta.97

### Impact
No response

### Esri team
Calcite (design)

ashetland:
This change may be an unintended byproduct of a recent update to Combobox Item Groups. It was discovered that groups were unnecessarily increasing the initial indentation level of their children. This was fixed, but the now nesting groups shows no visual change.

Anastasiia-Boleiko:
@ashetland Thank you for filing the issue, will be very helpful for ArcGIS Urban.

ashetland:
cc @geospatialem for triage

geospatialem:
Reallocated to the upcoming April milestone.

geospatialem:
Verified in `1.3.0-next.3`

![image](https://user-images.githubusercontent.com/5023024/231849749-44c1e457-8f57-42c6-b006-4a86d978c9cf.png)

PR #6749, opened 2023-04-07
fix(combobox): Visually nest group items properly

driskull:
**Related Issue:** #6384

## Summary

Fix css for nesting group combobox items.

### DIFF
--- a/src/components/combobox-item-group/combobox-item-group.scss
+++ b/src/components/combobox-item-group/combobox-item-group.scss
@@ -2,24 +2,21 @@
   @apply text-n2h;
   --calcite-combobox-item-spacing-unit-l: theme("spacing.2");
   --calcite-combobox-item-spacing-unit-s: theme("spacing.1");
-  --calcite-combobox-item-spacing-indent-1: theme("spacing.2");
-  --calcite-combobox-item-spacing-indent-2: theme("spacing.4");
+  --calcite-combobox-item-spacing-indent: theme("spacing.2");
 }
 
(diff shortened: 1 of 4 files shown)


driskull:
I can haz review? :)

jcfranco (approved):
Code changes LGTM! 

@ashetland Can you review the visual diffs? 

Good to merge once ☝️ approveth.

ashetland:
It looks like when `items` get slotted into an `item-group`, the var `--calcite-combobox-item-spacing-indent-multiplier` gets bumped by +1: what is 0, 1, 2... when not in a group becomes 1, 2, 3... when in a group. This bumping of the multiplier isn't ideal and complicates the hierarchical relationships when item-groups are used. Is it possible to change this behavior of `item-group`?

driskull:
@ashetland isn't that what the issue is asking for? item-groups should provide nesting indentation?

driskull:
This PR does make it look how it did previously.

ashetland:
Thanks @driskull, you're totally right. This PR fixes the behavior of nesting groups. Should the other behavior fix be logged as its own issue? Sometimes we've been making little fixes like this as we catch them, but maybe that's not the best practice. Happy to log a separate issue if that the preferred approach.

driskull:
> Should the other behavior fix be logged as its own issue

Yes, definitely. Otherwise we get scope creep

ashetland:
Awesome. Approving now. 🚀

Issue #6581, opened 2023-03-09
[Modal / Popover] Search Widget not receiving click event if Modal / Popover open after initial load


macandcheese:
### Actual Behavior

TBD

### Expected Behavior

TBD

### Reproduction Sample

TBD

### Reproduction Steps

TBD

### Reproduction Version

TBD

### Priority impact

p1 - need for current milestone

### Impact

Affecting multiple components, end users.

### Esri team

Calcite (design)

geospatialem:
Research will be conducted to determine next steps in May, as it relates to focus trap.

jcfranco:
Sharing workaround suggested by @driskull: using `focus-trap-disabled` mitigates this issue, but also allows users to tab into surrounding elements.

geospatialem:
Verified with `1.3.0-next.3`.

PR #6753, opened 2023-04-07
fix(modal, popover): Stop preventing focus on elements outside of the component.

driskull:
**Related Issue:** #6581

## Summary

@jcfranco this needs to be tested but I think it might fix the issue with clicks not being allowed in some instances.

I think what the doc is saying is that we should set the `document` property to our host element. However, the typings won't allow for that. I opened up an issue with `focus-trap` for clarity.

https://github.com/focus-trap/focus-trap/issues/939


jcfranco:
> this needs to be tested but I think it might fix the issue with clicks not being allowed in some instances.

👍 on testing. I think I tried this for https://github.com/Esri/calcite-components/issues/6281, but maybe additional configuration options used interfered with the desired behavior.

driskull:
I tested with the codepen from https://github.com/Esri/calcite-components/issues/6581 and this does fix that instance.

PR #6759, opened 2023-04-10
fix: normalize form component value when undefined or null

jcfranco:
**Related Issue:** N/A

## Summary

Prevents internal input used for form submission from being set to `"undefined"`/`"null"` vs `""`.

## Pending

* - [ ] test helper update


jcfranco:
Closing as this hasn't come up in a real use case.

PR #6769, opened 2023-04-11
fix(modal, popover): fix focus-trap from preventing first click

driskull:
**Related Issue:** #6581

## Summary

- Uses the component's host element for the focus trap. This seems to fix the issue in the related PR. I guess we should always use the host element in order to not run into any issues. Previously, it was using an element inside the shadowRoot.
- Sets up a mutation observer to update focusable elements when new elements are slotted
- Remove setting tabIndex on an element as a fallback


src/components/modal/modal.tsx
```diff
@@ -190,6 +190,7 @@ export class Modal
     connectConditionalSlotComponent(this);
     connectLocalized(this);
     connectMessages(this);
+    connectFocusTrap(this);
```
jcfranco, line 193: Does `connectFocusTrap` need to be updated to clean up existing focus traps?

src/components/popover/popover.tsx
```diff
@@ -255,6 +256,10 @@ export class Popover
   //
   // --------------------------------------------------------------------------
 
+  mutationObserver: MutationObserver = createObserver("mutation", () =>
```
jcfranco, line 259: As more components may need focus traps, we might need to provide a util to reuse a single mutation observer for all focus-traps.

src/components/modal/modal.tsx
driskull: I don't think so. There's no destroy method in the focus-trap. Its already getting deactivated on disconnectedCallback.

src/components/popover/popover.tsx
driskull: Agreed. Should we do this part of this PR or later?

jcfranco: We can tackle this separately. https://github.com/Esri/calcite-components/issues/6668 may lead to similar components implementing `focusTrapComponent` too, so I could look into it for that one.

jcfranco (approved):
⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️
⌨️🪤⌨️⌨️⌨️🪤⌨️⌨️🪤🪤⌨️⌨️🪤🪤🪤⌨️⌨️🪤🪤🪤⌨️🪤🪤🪤🪤⌨️
⌨️🪤🪤⌨️⌨️🪤⌨️🪤⌨️⌨️🪤⌨️⌨️🪤⌨️⌨️🪤⌨️⌨️⌨️⌨️🪤⌨️⌨️⌨️⌨️
⌨️🪤⌨️🪤⌨️🪤⌨️🪤⌨️⌨️🪤⌨️⌨️🪤⌨️⌨️🪤⌨️⌨️⌨️⌨️🪤🪤🪤⌨️⌨️
⌨️🪤⌨️⌨️🪤🪤⌨️🪤⌨️⌨️🪤⌨️⌨️🪤⌨️⌨️🪤⌨️⌨️⌨️⌨️🪤⌨️⌨️⌨️⌨️
⌨️🪤⌨️⌨️⌨️🪤⌨️⌨️🪤🪤⌨️⌨️🪤🪤🪤⌨️⌨️🪤🪤🪤⌨️🪤🪤🪤🪤⌨️
⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️

jcfranco:
This can't be easily tested with our setup, right? If so, can you add some info regarding testing?

jcfranco:
> Remove setting tabIndex on an element as a fallback

Forgot to bring this up earlier, but can we emphasize in the `focusTrapComponent` doc about focusable content being a requirement for components implementing focus trapping?

driskull:
> This can't be easily tested with our setup, right? If so, can you add some info regarding testing?

Correct, the best way to test is to use the codepen example.

https://codepen.io/mac_and_cheese/pen/rNZYxZE?editors=1000

1. Close the modal that is open initially
2. Open the modal using the button
3. Type in the search until a suggestion appears
4. Click the suggestion and make sure it navigates to that suggestion.

PR #6772, opened 2023-04-12
ci: fix next releases by increasing maxBuffer

benelan:
**Related Issue:** #

## Summary
Next releases are failing due to exceding the maxBuffer size of stdout/stderr. Here is the error message:
https://github.com/Esri/calcite-components/actions/runs/4671369609/jobs/8272269667#step:5:247

This increases the maxBuffer which should resolve the issue.

### DIFF
--- a/support/deployNextFromCI.ts
+++ b/support/deployNextFromCI.ts
@@ -7,6 +7,7 @@
   const childProcess = await import("child_process");
   const { promisify } = await import("util");
   const exec = promisify(childProcess.exec);
+  const maxBuffer = 1024 * 2048;
 
   async function deployNextFromCI(): Promise<void> {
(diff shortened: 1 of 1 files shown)


support/deployNextFromCI.ts
```diff
@@ -7,6 +7,7 @@
   const childProcess = await import("child_process");
   const { promisify } = await import("util");
   const exec = promisify(childProcess.exec);
+  const maxBuffer = 1024 * 2048;
```
jcfranco, line 10: Nitpick: WDYT about `1024 * 1024 * <factor>`? I saw that format used in a few places to make it more clear what the adjusted value is. Up to you.

jcfranco (approved):
Thanks for taking a look at this!

PR #6774, opened 2023-04-12
feat(navigation): remove prefix nav for menu and menu-item

anveshmekala:
**Related Issue:** #

## Summary

-  This PR will rename  `nav-menu` to `menu` and `nav-menu-item` to `menu-item`.
-  No longer autoFocus first & last items on ArrowUp & ArrowDown respectively in horizontal layout. 
-  Renamed `subMenuOpen` to `open`  in menu-item.
- Refactor keyboard support . All the interactions are handled in parent calcite-menu.


src/components/menu/menu.e2e.ts
```diff
@@ -0,0 +1,7 @@
+import { hidden, renders } from "../../tests/commonTests";
+
+describe("calcite-menu", () => {
+  it("renders", async () => renders("calcite-menu", { display: "flex" }));
+
+  it("honors hidden attribute", async () => hidden("calcite-menu"));
```
macandcheese, line 6: Do we want to try adding a11y tests at this point or wait for that?

src/components/menu/menu.tsx
```diff
@@ -36,7 +36,7 @@ export class CalciteNavMenu {
   @Prop() label: string;
 
   // todo evaluate slotted content and determine if it is a nav menu item, then limit # rendered when auto-collapsing based on width of parent
-  @Prop({ mutable: true }) minCollapsedItems?;
+  @Prop() minCollapsedItems: boolean;
```
macandcheese, line 39: We can remove this, for the time being, until we add back the auto-collapse functionality?

src/components/nav-menu-item/menu-item.tsx
```diff
@@ -77,15 +77,15 @@ export class CalciteNavMenuItem implements LoadableComponent {
...
 
   /** Displays the `text` */
-  @Prop({ mutable: true }) textEnabled: boolean;
+  @Prop() textEnabled: boolean;
 
   /**
    * @internal
    */
-  @Prop({ mutable: true }) subMenuOpen = false;
+  @Prop({ mutable: true }) open = false;
```
macandcheese, line 88: I know this is internal - 2 questions - is `open` descriptive enough? And, should this be publicly documented (is there a reason a dev user might want to be able to toggle this? I could see it maybe being useful in `layout="vertical"` menu use cases.

macandcheese (approved):
Awesome! I think this is good to merge into the wip navigation branch. Some questions, can address once merged.

src/components/nav-menu-item/menu-item.tsx
anveshmekala: i gave a thought about changing it to `open`, the reason being consistency. Though it is internal, keeping it consistent will be useful if we choose to make it public in future as you mentioned for `layout="vertical"`

src/components/menu/menu.tsx
anveshmekala: sure, removed.

src/components/menu/menu.e2e.ts
anveshmekala: added a11y for `menu-item` and `menu` components.

Issue #6555, opened 2023-03-02
[List and List Item] Support List Item closed functionality


geospatialem:
### Description

Add full parity to the `list` component, which is provided in the `value-list` and `pick-list` components.

Children should not be destroyed, or removed as part of the effort.

### Acceptance Criteria

- [x] Closed functionality of `list-item`s
   - [x] Mouse and keyboard support
   - [x] Keyboard support when screen readers are active for both Windows (NVDA, JAWS) and Mac (VoiceOver) users

### Relevant Info

Part of the https://github.com/Esri/calcite-components/issues/6430 epic

### Which Component

- List
- List Item

### Example Use Case

Depicted with `value-list` and `value-list-item`: 
https://codepen.io/geospatialem/pen/abjxWwL

### Esri team

Calcite (dev)

driskull:
@jcfranco It seems like a user would be best to specify the action for removal and handle removing the element. Otherwise, we may need to get into creating a `beforeRemoval` callback incase some async thing needs to happen. Even an undo UI.

With the list, the user manages the DOM elements. It might not be best for us to be removing DOM elements.

We might want to discuss a pattern for when removing DOM elements that are not in a shadowRoot is okay.

geospatialem:
Will be consulting the team on a pattern prior to any work to determine next steps.

geospatialem:
@jcfranco We're exploring closed functionality based on group discussions on 4/11, rather than removable/destroyed children in the DOM. WDYT about the proposed functionality and approach?

cc @driskull

geospatialem:
Verified in `1.3.0-next.7` with the following example:

```html
<calcite-list selection-mode="single" label="test" filter-enabled>
    <calcite-list-item closable label="test1" value="test1" description="hello world 1">
      <calcite-icon icon="banana" slot="content-start" style="color: var(--calcite-ui-success)"></calcite-icon>
      <calcite-action appearance="transparent" icon="ellipsis" text="menu" label="menu" slot="actions-end"></calcite-action>
    </calcite-list-item>
    <calcite-list-item closable label="test2" value="test2" description="hello world 2">
      <calcite-icon icon="compass" slot="content-start" style="color: var(--calcite-ui-success)"></calcite-icon>
    </calcite-list-item>
    <calcite-list-item closable label="test3" value="test3" description="hello world 3">
      <calcite-icon icon="compass" slot="content-start" style="color: var(--calcite-ui-success)"></calcite-icon>
[...]

PR #6775, opened 2023-04-12
feat(list, list-item): Adds the ability to close a list-item

driskull:
**Related Issue:** #6555

## Summary

- Added `closed` property
- Added `closable` property
- Added `calciteListItemClose` event
- Added t9n messages for "Close" text.
- Updated story
- Added tests


src/components/list-item/list-item.e2e.ts
```diff
@@ -115,4 +115,37 @@ describe("calcite-list-item", () => {
...
+  it("should fire close event when closed", async () => {
+    const page = await newE2EPage({ html: "<calcite-list-item closable>test</calcite-list-item>" });
+
+    const calciteListItemClose = await page.spyOnEvent("calciteListItemClose", "window");
+
+    const closeButton = await page.find(`calcite-list-item >>> .${CSS.actionsEnd} calcite-action`);
+
+    await closeButton.click();
+
+    expect(calciteListItemClose).toHaveReceivedEvent();
```
jcfranco, line 149: Can you use `toHaveReceivedEventTimes` to increase coverage?

src/components/list/list.stories.ts
```diff
@@ -337,3 +338,32 @@ export const customContent_TestOnly = (): string => html`<calcite-list disabled>
     </div></calcite-list-item
   >
 </calcite-list>`;
+
+export const closableListItems_TestOnly = (): string => html`<calcite-list
```
jcfranco, line 342: Can this be revisited so that the minimal HTML required for the test is used?

src/components/list-item/list-item.tsx
```diff
@@ -431,6 +501,7 @@ export class ListItem implements InteractiveComponent, LoadableComponent {
             [CSS.containerBorderSelected]: borderSelected,
             [CSS.containerBorderUnselected]: borderUnselected
           }}
+          hidden={closed}
```
jcfranco, line 504: We should consolidate how we handle closed/removed items. I've noticed we have the following discrepancies:

1. naming: `closable` vs `removable`
2. approach: hiding vs removing the node

I'll create a separate issue for this.

jcfranco (approved):
❎a📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝
📝❎📝📝📝❎📝❎📝📝📝❎📝📝❎❎❎📝📝❎❎❎📝📝❎❎❎📝❎📝
📝📝❎📝❎📝📝❎📝📝📝❎📝❎📝📝📝📝❎📝📝📝📝❎📝📝📝📝❎📝
📝📝📝❎📝📝📝❎📝📝📝❎📝📝❎❎📝📝📝❎❎📝📝📝❎❎📝📝❎📝
📝📝📝❎📝📝📝❎📝📝📝❎📝📝📝📝❎📝📝📝📝❎📝📝📝📝❎📝📝📝
📝📝📝❎📝📝📝📝❎❎❎📝📝❎❎❎📝📝❎❎❎📝📝❎❎❎📝📝❎📝
📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝

src/components/list-item/list-item.tsx
driskull: I don't think we use removable anywhere else besides value-list, right?

I think sometimes we can't remove the node because we need the slot to still be there so we hide it.

Issue #3392, opened 2021-10-29
Enhancement: Focus outline inset


joeyHarig:
### Description
Allow specification of focus outline on components to be inset. Components like `calcite-action` and `calcite-input` utilize inset outlines when focused on while `calcite-button` and `calcite-dropdown` utilize an outset outline. When the surrounding elements prevent overflow, outset outlines can be covered hidden.

### Acceptance Criteria 
- add prop to components with outset focus outlines to allow them to be inset

#### Which Component
- `calcite-button`
- `calcite-stepper`
- `calcite-tabs`

_possibly_
- `calcite-switch`

#### Example Use Case
<img width="750" alt="Screen Shot 2021-10-29 at 3 52 02 PM" src="https://user-images.githubusercontent.com/55154968/139505987-1a74efa5-8763-4551-a794-7325ac0f61bf.png">

macandcheese:
We will need to examine how to allow this while ensuring focus is visible, for example a focus outline that is inset on a filled blue button is not perceivable.

geospatialem:
Reallocating this issue to tackle in February's priorities.

geospatialem:
Re-adjusting the timeline to accommodate upcoming Esri events and will tackle as part of the March priorities in coordination with #4633.

ashetland:
### Proposal to make this a system-wide customization


Add two CSS variables for customization at the root or component level (actual text TBD):
- `--calcite-ui-focus-offset` (default to what is currently most common (-2px?))
- `--calcite-ui-focus-color` (defaults to ui-brand)

cc: @jcfranco @macandcheese @geospatialem

ashetland:
Also related: https://github.com/Esri/calcite-components/issues/6588

geospatialem:
Further research is needed to carry out the [above proposal of two exposed CSS vars](https://github.com/Esri/calcite-components/issues/3392#issuecomment-1481920535).

In April we'll research further to assess a time estimate and priority prior to assigning to a future milestone for implementation.

geospatialem:
Verified on `master`:

```css
--calcite-ui-focus-color: var(--calcite-ui-danger);
```

![button focus var verify](https://user-images.githubusercontent.com/5023024/234133922-2eb51652-2b09-474f-b289-2ec58cbb4882.png)
![switch focus var verify](https://user-images.githubusercontent.com/5023024/234134201-ce74f23c-8716-47f0-ad68-04fffd7903a8.png)



```css
[...]

PR #6782, opened 2023-04-14
feat: add global CSS props for focus offset and color

jcfranco:
**Related Issue:** #3392 

## Summary

Adds the following global CSS props:

* `--calcite-ui-focus-color`
* `--calcite-ui-focus-offset-invert` – inverts the applied offset when set to `1` or keeps as is when `0`


macandcheese (approved):
🧘‍♂️ excellent

src/assets/styles/global.scss
```diff
@@ -44,9 +44,11 @@
 
   --calcite-border-radius: 4px;
   --calcite-border-radius-base: 0;
-  --calcite-ui-opacity-disabled: 0.5;
   --calcite-panel-width-multiplier: 1;
+  --calcite-ui-focus-color: var(--calcite-ui-brand);
```
macandcheese, line 48: Cc @geospatialem for inclusion when we add / improve css var and token docs

PR #6786, opened 2023-04-14
chore: Adds ability to remove demo page header via a close button.

driskull:
**Related Issue:** N/A

## Summary

image

Also increased header z-index to make sure its on top.


jcfranco:
Could you describe the use case for this?

driskull:
Sure, the use case is that it gets in the way of full page demos like the shell component demos and this provides a quick way to close it.

jcfranco (approved):

Issue #5927, opened 2022-12-07
Shell action bar expand tooltip displays behind shell center row


mpayson:
### Actual Behavior

The `expand-tooltip` slot for the primary action bar in calcite shell displays the tooltip behind the shell center row when the center row is visible

### Expected Behavior

The tooltip is displayed in front of the center row

### Reproduction Sample

https://codepen.io/mpayson/pen/eYKxWrz

### Reproduction Steps

1. Hover over the 'Expand' button in the bottom left
2. See that the tooltip displays underneath the maroon center row

### Reproduction Version

beta.98

### Relevant Info

Happens in both Firefox and Chrome

### Esri team

ArcGIS Online

driskull:
@macandcheese @jcfranco should the side panels be a higher zindex than the center ones?

driskull:
@macandcheese @ashetland I need some advice on which shell panels should be higher z-index. Should we have the side panels always appear over the inner panels? Or do we need to have some sort of CSS var or property to allow users to set z-indexing?

macandcheese:
I'd agree that side panels should appear over inner panels by default. That said, if there's a way to allow it to be overridden with css variable or other tactic, that's a good option to provide.

geospatialem:
Verified on `v1.3.0-next.12` in Chrome

<img width="181" alt="expand tooltip verified" src="https://user-images.githubusercontent.com/5023024/233805685-8d9196f2-1380-408d-a596-d8101d18152d.png">

PR #6787, opened 2023-04-14
fix(shell-panel): Side panels should appear over center panels

driskull:
**Related Issue:** #5927

## Summary

- Makes `shell-panels` that have `layout="vertical"` have a higher z-index than ones that are `layout="horizontal"` by default.
- Adds `--calcite-shell-panel-z-index` CSS property to modify z-index per shell-panel if necessary.


benelan (approved):
👍

Issue #6737, opened 2023-04-05
Localization support for `de-AT`


geospatialem:
### Description

Provide support to the language, `German - Austria (de-AT)`, in particular with the `date-picker` component.

### Acceptance Criteria

- [ ] The `date-picker` component's month should be  set to `Jänner` (currently set to German value of `Januar`) for  `German - Austria (de-AT)` languages. - @annierm18
- [ ] Update [supported languages](https://developers.arcgis.com/calcite-design-system/resources/localization/#calcite-supported-languages) to include `German - Austria (de-AT)*` - @geospatialem 

### Relevant Info

Enhancement request

### Which Component

[`date-picker`](https://github.com/Esri/calcite-components/tree/master/src/components/date-picker/assets/date-picker/nls)

### Priority impact

p3 - want for upcoming milestone

### Esri team

N/A

annierm18:
@geospatialem for this change, we can make a copy of this [de.json](https://github.com/Esri/calcite-components/blob/master/src/components/date-picker/assets/date-picker/nls/de.json) file, rename the copy to "de-AT", and then update the following two values:

1. In months.abbreviated: change "Jan" to "Jän"
2. In months.wide: change "Januar" to "Jänner"

geospatialem:
Verified in `1.3.0-next.5` with the [following Codepen](https://codepen.io/geospatialem/pen/PoyNzgB).

annierm18:
@geospatialem the distributor that requested this change tested it on the Codepen that you shared; he said the change looks good.

PR #6788, opened 2023-04-14
feat: add json for de-AT locale with locale-specific data

annierm18:
**Related Issue:** #6737 

## Summary
The change in this PR is to add support for the German - Austria (de-AT) locale in the date-picker component. It was requested by an international distributor. The change entails adding a new json file for the de-AT locale with locale-specific data.

### DIFF
--- a/src/components/date-picker/assets/date-picker/nls/de-AT.json
+++ b/src/components/date-picker/assets/date-picker/nls/de-AT.json
@@ -1,0 +1,32 @@
+{
+  "default-calendar": "gregorian",
+  "separator": ".",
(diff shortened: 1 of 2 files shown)


driskull (approved):
👍

eriklharper (approved):

jcfranco:
Awesome! 🎉 Thanks for the PR, @annierm18! 

I believe the new locale also needs to be [added here](https://github.com/Esri/calcite-components/blob/master/src/utils/locale.ts#L57) to treat it as supported. I can help take care of this later today if you don't have the bandwidth.

Issue #5867, opened 2022-12-01
calcite-tree doesn't initialize properly when options are pre-selected with selection-mode="ancestors"


sonofflynn89:
### Actual Behavior

In a simple tree with one parent node and one child node, the parent node will initialize with an indeterminate state even if both parent and child are marked as selected:

![MicrosoftTeams-image (13)](https://user-images.githubusercontent.com/30124777/205094629-8a0e2705-4526-48f8-ad9f-6b82a49fc7cc.png)

The same thing happens when only the child is marked as selected (and the option remains unbolded).

### Expected Behavior

If either:
- the parent node and the child node are marked as selected
- the child node is marked as selected

then the parent node's checkbox icon and label should reflect that reality:

![MicrosoftTeams-image (14)](https://user-images.githubusercontent.com/30124777/205094957-aba99801-5e6f-454f-ab11-5d1a5b07da00.png)

### Reproduction Sample

https://codepen.io/sonofflynn89/pen/ExRdNwR

### Reproduction Steps

1. Open the code pen. Note that the top example has both parent and child selected, while the bottom example has only the child selected
2. Note that both parent nodes have checkboxes with the `indeterminate` icon, even though they should be checked
3. Uncheck and recheck the parent nodes
4. Note that the parent nodes have checkboxes with the `checked` icon

### Reproduction Version

1.0.0-beta.98

### Impact

minor visual bug

### Esri team

ArcGIS Hub

geospatialem:
Looks to be a regression from `beta.66`, possibly https://github.com/Esri/calcite-components/pull/3111?

## `beta.65`:
![image](https://user-images.githubusercontent.com/5023024/205104696-f45e16d8-bff3-4f29-b1e9-a39fa1e005c0.png)

## `beta.66`:
![image](https://user-images.githubusercontent.com/5023024/205104648-a35bfbc4-46f5-4a2f-bc7c-6a0489f82894.png)

sonofflynn89:
@geospatialem, same thing happens when just the child node is marked as selected. Should I update the description to account for both scenarios?

sonofflynn89:
@geospatialem done! Codepen has been updated with both scenarios as well

sonofflynn89:
@geospatialem any updates on this? I'm wrestling with a lot of other issues trying to programmatically select / unselect tree items and this particular bug is foiling all my workarounds.

geospatialem:
[...]

PR #6795, opened 2023-04-15
fix(tree-item): ensure tree-item properly reflect ancestor-mode selection on initialization

jcfranco:
**Related Issue:** #5867 

## Summary

This updates the initial tree-item initialization logic to consider each tier of selected items in the tree hierarchy.

### DIFF
--- a/src/components/tree-item/tree-item.tsx
+++ b/src/components/tree-item/tree-item.tsx
@@ -496,18 +496,33 @@ export class TreeItem
     items.forEach((item) => (item.parentExpanded = expanded));
   };
 
-  private updateAncestorTree = (): void => {
-    if (this.selected && this.selectionMode === "ancestors") {
-      const ancestors: HTMLCalciteTreeItemElement[] = [];
-      let parent = this.parentTreeItem;
-      while (parent) {
-        ancestors.push(parent);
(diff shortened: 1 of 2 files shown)


src/components/tree-item/tree-item.tsx
```diff
@@ -496,18 +496,37 @@ export class TreeItem
...
-        parent = parent.parentElement?.closest("calcite-tree-item");
-      }
-      ancestors.forEach((item) => (item.indeterminate = true));
+  /**
+   * This is meant to be called in `componentDidLoad` in order to take advantage of the hierarchical component lifecycle
+   * and help check for item selection as items are initialized
+   *
+   * @private
+   */
+  private updateAncestorTree(): void {
```
driskull, line 505: Do we have an issue, or do we want to open one, where the `tree` is in in charge of handling selection instead of individual `tree-items`? It would be a refactor but might be beneficial for the future.

driskull (approved):
👍 Nice

src/components/tree-item/tree-item.tsx
jcfranco: Yeah, I was thinking the same. I'll create a refactor issue for tracking purposes. Tree is a beast for sure. 👹

Issue #5167, opened 2022-08-16
ValueListItem has no emitter for onCalciteListItemChange


jcfranco:
### Description

`calcite-value-list-item` reuses `onCalciteListItemChange` from its wrapped `calcite-pick-list-item`, but doesn't have an emitter for it, so it is missing from the types.

Stems from https://github.com/Esri/calcite-components/issues/5075#issuecomment-1217151710

### Acceptance Criteria

Add an event emitter for `onCalciteListItemChange` so the types are in sync (similar to https://github.com/Esri/calcite-components/blob/09aebeb2d2ba295ad47fa22f3f9e5256358a98f5/src/components/value-list-item/value-list-item.tsx#L157-L160).

### Which Component

`calcite-value-list-item`

jcfranco:
@geospatialem Can we triage this for this milestone? I had an existing branch for this one and I just submitted the PR.

geospatialem:
Verified in `1.3.0-next.8` with the calcite-components-examples [preact-typescript](https://github.com/Esri/calcite-components-examples/tree/master/preact-typescript) example with the following:

```html
<calcite-value-list>
  <calcite-value-list-item oncalciteListItemChange={(e) => console.log(e.target.value, e.target.selected)} id="water" label="Waterfalls" description="Vertical drops from a river." value="waterfalls" />
  <calcite-value-list-item oncalciteListItemChange={(e) => console.log(e.target.value, e.target.selected)} label="Rivers" description="Large naturally flowing watercourses." value="rivers" />
  <calcite-value-list-item oncalciteListItemChange={(e) => console.log(e.target.value, e.target.selected)} label="Estuaries" description="Where the river meets the sea." value="estuaries" />
</calcite-value-list>
```

PR #6797, opened 2023-04-16
fix(value-list-item): add missing event emitter

jcfranco:
**Related Issue:** #5167

## Summary

Adds event emitter, from supporting pick-list-item component, to include into typings.


driskull (approved):
👍

PR #6802, opened 2023-04-17
docs(list-item): add context to selected prop with parent list's selectionMode prop

geospatialem:
**Related Issue:** n/a

## Summary
Received anonymous feedback this morning where `list-item`s `selected` attr is not always reflected, when the parent `list` has a `selection-mode` of `"none"`.

Updates the `selected` attribute to reflect changes when the parent `list` has a `selection-mode`  

Also, audited `selected` and `checked` attributes across components, which have already been handled or do not apply to this case  
    - cc: @macandcheese for inclusion in `chip`/`chip-group` in the near future.

### DIFF
--- a/src/components/list-item/list-item.tsx
+++ b/src/components/list-item/list-item.tsx
@@ -121,7 +121,7 @@ export class ListItem
   @Prop() setPosition: number = null;
 
   /**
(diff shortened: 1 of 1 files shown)


driskull (approved):
👍

Issue #6798, opened 2023-04-17
Improve tooltip arrow to allow semi-transparent background color


cont-kolomeets:
### Actual Behavior

We have customizable tooltips and if a user changes the background transparency, the arrow artifacts become visible.

![image](https://user-images.githubusercontent.com/5318527/232388014-0200a7ae-3cc8-4db6-b327-c9dc783d7b75.png)

### Expected Behavior

Please fix the arrow, so it doesn't go inside the container.

### Reproduction Sample

https://codepen.io/cont-kolomeets/pen/JjmKGGb?editors=100

### Reproduction Steps

See the sample.

### Reproduction Version

1.0.7

### Priority impact

p0 - emergency

### Esri team

ArcGIS Business/Community Analyst

geospatialem:
@cont-kolomeets We're interested in learning more of the use case behind adding transparency to the `tooltip`s background as it could impact accessibility and contrast.

driskull:
If we want to fix this, we can use css `clip-path` instead of just rotating the div 45deg.

driskull:
If we want to support transparent backgrounds on all of our components we may need to set up some screenshot tests for each one.

cont-kolomeets:
@geospatialem 

> We're interested in learning more of the use case behind adding transparency to the tooltips background as it could impact accessibility and contrast.

Agree that usage of transparency should be limited, but it's just something we allow for tables and charts in our infographic builder. In some cases alpha of 0.8 can be used. Also you never now in which context users might want to use the tooltips.

![image](https://user-images.githubusercontent.com/5318527/232674662-f5628c69-fecf-4211-98ee-cc90bb12658f.png)

driskull:
@jcfranco @macandcheese if we want to support semi-transparent backgrounds in all of our components, we need to do some work to make sure we're only setting the background on one element within the component and any components used within the component have `appearance="transparent"`. Otherwise, the backgrounds get overlapped which compounds the transparency and color.

This would probably be an audit of every component to setup screenshot tests and make sure that backgrounds with some transparency look good. Do we want to create an epic for this?

driskull:
Going forward, we may want to introduce an internal component (possibly functional component) to handle rendering a triangle SVG and styling it with a stroke/fill.

See: https://github.com/floating-ui/floating-ui/pull/2195

geospatialem:
Verified in `1.4.0-next.6`

[...]

PR #6803, opened 2023-04-17
fix(tooltip, popover): Support transparent backgrounds

driskull:
**Related Issue:** #6798

## Summary

- Uses css `clip-path` instead of `transform: rotate` to create the caret arrow tail.
- Removes CSS where we are setting the BG on two separate elements.
- Screener tests

### DIFF
--- a/src/assets/styles/_floating-ui.scss
+++ b/src/assets/styles/_floating-ui.scss
@@ -112,7 +112,6 @@ $pointer_offset: -$pointer_size * 0.5;
 
   .arrow::before {
     content: "";
-    @apply shadow-1;
     background: var(--calcite-ui-foreground-1);
   }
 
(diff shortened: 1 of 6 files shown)


jcfranco (approved):
🎉

driskull:
@jcfranco @ashetland Let me know if you'd like to proceed with this or not.

ashetland:
Losing the border on the arrow is a deal breaker for me on this one. Without the border, the arrow practically disappears and thus reads like a bug. I think supporting transparency, even if minimal, is valid though. What would it mean to support it? Are we talking breaking changes / major rework?

driskull:
@jcfranco you'll have to make the call on what is more important here.

If we have an arrow behind an arrow that will defeat the purpose of transparency as well due to compounding background colors.

driskull:
Going to close this as it's not possible to use just clip-path and have a border with it. I even tried putting a rotated box div inside and trying to clip it but it still didn't render correctly.

Going forward, we may want to introduce an internal component (possibly functional component) to handle rendering a triangle SVG and styling it with a stroke/fill.

See: https://github.com/floating-ui/floating-ui/pull/2195

CSS just isn't good at doing this even in 2023.

Issue #6773, opened 2023-04-12
Closed date picker popup requires tabbing out and back in to be reopened


jcfranco:
### Actual Behavior

Closing a date picker via Escape force a user to tab outside and back into the component to display again.

### Expected Behavior

An affordance in the picker to toggle the popup (akin to combobox) and adding support to reopen on ⬇️ key press.

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/input-date-picker/

### Reproduction Steps

1. Click sandbox above input
2. Press Tab
3. Picker popup opens
4. Press Escape
5. Picker closes

Once closed, the picker can no longer be opened unless users tab outside and back in

### Reproduction Version

1.2.0

### Priority impact

p4 - not time sensitive

### Esri team

Calcite (dev)

geospatialem:
Design implementation will follow `combobox`. Routing to development. cc @ashetland @jcfranco

ashetland:
cc @SkyeSeitz for Figma updates

geospatialem:
Verified on `master` via mouse, keyboard and AT - Windows JAWS and NVDA. 

cc @SkyeSeitz @ashetland - once we're good on Figma changes, can close out the issue.

ashetland:
Did the PR add anything to Input Time Picker or just Input Date Picker? This issue implies both, but I think the work I reviewed only included Input Date Picker.

geospatialem:
> Did the PR add anything to Input Time Picker or just Input Date Picker? This issue title implies both, but I think the work I reviewed only included Input Date Picker.

@ashetland - Good catch. Refactored this issue to date-picker, and opened #6830 to tackle the time picker effort. The design should be the same for both components.

geospatialem:
Figma changes are noted - closing out this issue.

PR #6805, opened 2023-04-18
feat(input-date-picker): allow toggling date picker by clicking the input or entering the down/escape key

jcfranco:
**Related Issue:** #6773 

## Summary

This enhances the input-date-picker to allow it to be toggled via mouse and keyboard vs initial focus.

This also updates input to allow overriding role to accommodate other a11y structures.


jcfranco:
+@geospatialem in case additional screen reader testing is needed.

src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -503,22 +508,27 @@ export class InputDatePicker
...
             <div
               aria-hidden={toAriaBoolean(!this.open)}
               aria-label={messages.chooseDate}
+              aria-live="polite"
               aria-modal="true"
               class={{
                 [CSS.menu]: true,
                 [CSS.menuActive]: this.open
               }}
+              id="date-picker-dialog"
```
geospatialem, line 531: I *think* if multiple dialogs are present on the page it's causing some confusion with screen readers. 

Can this `id` be more unique generating a guid from the [guid utils](https://github.com/Esri/calcite-components/blob/master/src/utils/guid.ts), associated with the name, similar to [`combobox`](https://github.com/Esri/calcite-components/blob/master/src/components/combobox/combobox.tsx#L1134)? 

NVDA picks up collapse, but JAWS isn't providing context. 😢

geospatialem (commented):
The `id` might need to be unique when more than one `date-picker` is present in an app?

There's also a bunch of random numbers read back in both NVDA and JAWS, not sure where they are being read from?

JAWS transcript when pressing the ⬇️ key to show the `date-picker` component, and changing the date from **3/7/2023** to **3/8/2023** :
```
Calcite Input Date Picker
Input Date Picker Label edit combo 
3/7/2023
To set the value use the Arrow keys or type the value.
1
26
25
27
19
5
10
8
9
16
17
14
12
4
3
22
24
13
15
2
1
28
26
29
Choose date
March 2023
Previous month Button 
To activate press Enter.
Year Edit 
2023
Type in text.
Calcite Input Date Picker
Next month
Button 
To activate press Enter.
table
Column 3, Row 3
Grid 
Tuesday, March 7, 2023 Button 
Tu
column 2 row 2
To activate press Enter.
Wednesday, March 8, 2023 Button 
We
To activate press Enter.
Calcite Input Date Picker
```

src/components/input-date-picker/input-date-picker.tsx
driskull: Yeah these should probably be unique since we don't know how screen readers will deal with shadowRoot.

jcfranco: And here I was, thinking shadow DOM would make this easier. 🥲 

Serious talk, thanks for pointing this out. I'll fix this right away. Can we add this to our a11y guidelines?

src/components/input/input.tsx
```diff
@@ -1164,6 +1169,7 @@ export class Input
               value={this.value}
               // eslint-disable-next-line react/jsx-sort-props
               ref={this.setChildElRef}
+              {...this.globalAttributes}
```
driskull, line 1172: @jcfranco can you elaborate on this one?

If a user sets a role on the host element, the same role will be set internally on this shadowRoot element?

Won't the nested roles conflict with each other? I'm not 100% on how screen readers read shadowRoot but it seems like a parent and a shadow child having the same role might be problematic?

driskull (commented):
Code looks good to me 👍 

Just have questions on the role setting.

src/components/input/input.tsx
driskull: Also, if a user sets lang on the host it will be set on this element. That's not a problem but just seems redundant/not necessary.

jcfranco: I think overriding the internal input's role is necessary to be able to implement the [combobox  patterns](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/#examples) from the ARIA Authoring Practices Guide:

> | Role	| Attribute	| Element |	Usage |
> |--------|--------|--------|--------|
> | combobox |		| input | 	Identifies the input element as a combobox.

> | Role	| Attribute	| Element |	Usage |
> |--------|--------|--------|--------|
> | combobox |		| input[type="text"] | 	Identifies the input element as a combobox.

Otherwise, the host will have one role and the actual input will have another one which might interfere with the screen reader interpretation. In my testing, VO reads and offers different options before and after making this change (compared to the example from https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-datepicker). I'll defer to @geospatialem on this one.

> Also, if a user sets lang on the host it will be set on this element. That's not a problem but just seems redundant/not necessary.

The util is configured to only watch `role` in this component (see https://github.com/Esri/calcite-components/pull/6805/files#diff-fe3d603d6390f87fdcdd3f40e34778ccdb07ff9edc52e440ad13c9571cfceadaR475 above).

geospatialem: Agreed on the combobox role here. We did some testing with the light and shadow DOM to be sure, and both should be interpreted separately.

driskull (approved):
👍

Issue #6463, opened 2023-02-13
Date picker is fixed on screen when scrolling


jlias-devtester:
### Actual Behavior

The calendar of start date is fixed on screen while scrolling up or down on 'Create report' dialog with resolution 1440 x 900 or below.

### Expected Behavior

Calendar should be moved along with the scroll bar after clicked calendar icon of start date and then scroll on 'Create report' dialog when resolution is set to 1440 x 900 or below.

### Reproduction Sample

https://codepen.io/rpanichakit/pen/WNgePqY?editors=100

### Reproduction Steps

Steps:
1. Prepare a clean EN browser on machine with resolution is set to 1440 x 900.
2. Login https://qaext.arcgis.com with a valid account.
3. Go to Organization->Status->Reports tab.
4. Click 'Create report' dropdown and select 'Single report'.
5. Click 'Report type' dropdown and select "Item view report", make sure scroll bar is appeared. (key step)
6. Click calendar icon of start date Under 'Timeframe'. (key step)
7. Scroll the mouse or drag the scroll bar.
8. Observe that the calendar of start date is fixed on screen while scrolling up or down on 'Create report' dialog with resolution 1440 x 900 or below.

### Reproduction Version

1.0.3

### Esri team

ArcGIS Online

geospatialem:
Research to be conducted in June for a priority and estimate for an upcoming milestone.

geospatialem:
Verified on `master`.

PR #6815, opened 2023-04-19
fix(input-date-picker): Correctly position open component when scrolling

driskull:
**Related Issue:** #6463

## Summary

Sets up floating UI when the component is opened in order to get all scrollable parents.


src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -644,6 +644,7 @@ export class InputDatePicker
 
   transitionEl: HTMLDivElement;
 
+  @Watch("open")
```
driskull, line 647: @jcfranco the problem seems to be when the input-date-picker is rendered the `floatingEl` has no parent yet. Not sure if its because it's not rendered or if some other stuff isn't ready yet.

So if we watch for `open` and then setup the floating ui again it can get all the scrolling parents correctly.

Seems like this is the only component affected.

src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -654,7 +654,7 @@ export class InputDatePicker
         ? endWrapper || startWrapper
         : startWrapper || endWrapper;
 
-    connectFloatingUI(this, this.referenceEl, this.floatingEl);
+    requestAnimationFrame(() => connectFloatingUI(this, this.referenceEl, this.floatingEl));
```
driskull, line 657: Need to wait for next frame to connect floating UI because some rendering is still going on. The rendering needs to happen first before trying to get all scrollable parents.

src/components/input-date-picker/input-date-picker.e2e.ts
```diff
@@ -580,4 +580,39 @@ describe("calcite-input-date-picker", () => {
     expect(changeEvent).toHaveReceivedEventTimes(1);
     expect(await datepickerEl.getProperty("value")).toEqual(["2022-08-15", "2022-08-20"]);
   });
+
+  it("should position on scroll when overlayPositioning is fixed", async () => {
```
jcfranco, line 584: Nice! Could you try using `commonTests#floatingUIOwner`? It should cover scrolling updates.

src/components/input-date-picker/input-date-picker.tsx
jcfranco: Interesting. On a related note, I was thinking `input-date-picker` should be refactored to use `popover` for consistency and reusability. Maybe that would help eliminate since this doesn't appear to be an issue with the `input-time-picker`.

jcfranco (approved):
👍

src/components/input-date-picker/input-date-picker.e2e.ts
driskull: The test already has a floatingUIOwner test. Maybe it should be updated to test with `overlay-positioning="fixed"` as well?

driskull: Tried this and it passed so its not catching the issue.

```tsx
 it("owns a floating-ui 2", () =>
    floatingUIOwner(
      `<calcite-input-date-picker overlay-positioning="fixed" value="2022-11-27" min="2022-11-15" max="2024-11-15"></calcite-input-date-picker>`,
      "open",
      { shadowSelector: ".menu-container" }
    ));
```

jcfranco: Hmm... I think the helper might need some tweaking. Can you create a follow-up issue for me? Let's proceed with the existing test here.

driskull: created: https://github.com/Esri/calcite-components/issues/6838

Issue #6668, opened 2023-03-27
[Input Date Picker] Add keyboard trap to Date Picker component


geospatialem:
### Description

Add in keyboard trapping so assistive technology (AT) users can cycle through the different options, ensuring they know the full extent of the dialog and what's available to them.

Spurred from https://github.com/Esri/calcite-components/issues/5582

### Acceptance Criteria

Add keyboard trapping to `date-picker`, where users can close the dialog using the `Esc` key.

### Relevant Info

https://github.com/Esri/calcite-components/issues/5582

### Which Component

`input-date-picker` and `date-picker`

### Priority impact

p2 - want for current milestone

### Esri team

Calcite (dev)

jcfranco:
Waiting on #6773 to address some concerns mentioned ☝️.

geospatialem:
Verified in `1.3.0-next.13` on Mac and Windows.

PR #6816, opened 2023-04-20
feat(input-date-picker): add focus trap support

jcfranco:
**Related Issue:** #6668 

## Summary

Adds focus-trap to date-picker dialog to improve a11y.

**Note:** This also adds a util to help retrieve props from the currently focused element.

## Notable `focusTrapComponent` changes

* `connectFocusTrap` now supports passing a different focus-trap target element (for setting the focus trap on a component's subtree) and overrides to certain options from `focus-trap`'s `focusTrap()`.
* `activateFocusTrap`/`deactivateFocusTrap` now allow passing options to `focus-trap`'s `activate`/`deactivate` methods.
* `updateFocusTrapElements` is now optional.


jcfranco:
FYI, this is based on https://github.com/Esri/calcite-components/pull/6805. [These](https://github.com/Esri/calcite-components/pull/6816/files/1642a84c5ec7ba2aaf881124ab0c4f6645f91011..d9b5becb56e67e744eba14f102e13ea1087bd66b) are the relevant changes for review.

jcfranco:
+@geospatialem for a11y testing.

driskull (approved):
FocusTrap related changes look good!

geospatialem (commented):
Likely out of scope for the focus trap support, and also *nitpick*:

Would we want to support keyboard focus where the "previous month" button is focused on first? 

Could see this being confusing for mouse users, too, but it seems like an odd behavior for AT to have to tab in after the dialog is displayed. Also makes it harder for AT users to close the dialog, as they would need to tab in first, then press `esc`.

jcfranco:
@geospatialem I updated it so that when the date-picker is opened via the ⬇ key, it gets focused. Clicking won't move focus to allow users to type the value if desired.

PR #6820, opened 2023-04-21
test(stepper): adds test to verify stepper-item re-render when numbering-system changes on parent

anveshmekala:
**Related Issue:** #5979 

## Summary

Add a test to verify stepper-item re-renders when numbering system is changed on parent stepper.


driskull (approved):
👍

PR #6826, opened 2023-04-21
fix(stepper): rerender items when numbering system changes

benelan:
**Related Issue:** #5979

## Summary

Changing `calcite-stepper`'s `numberingSystem` property will force a rerender so the item step numbers are updated. Thanks @jcfranco for the implementation.


driskull (approved):
👍

src/components/stepper/stepper.tsx
```diff
@@ -52,6 +54,14 @@ export class Stepper {
    */
   @Prop({ reflect: true }) numberingSystem?: NumberingSystem;
 
+  @Watch("numberingSystem")
+  numberingSystemChange(): void {
+    for (const item of this.itemMap.keys()) {
+      // trigger item rendering to pick up parent numbering system
+      forceUpdate(item);
```
anveshmekala, line 61: can we avoid forceUpdating the whole component and parse down the `numberingSystem` prop onto `stepper-item` and mark it as `internal`?

anveshmekala: Similar to [this](https://github.com/Esri/calcite-components/pull/6563/files#diff-7d3850db7c4d6df6d974db7f03ca7c67b4cfadc998eae076b508c7b78b266167R58) . Thoughts?

src/components/stepper/stepper.e2e.ts
```diff
@@ -602,5 +602,10 @@ describe("calcite-stepper", () => {
 
     const stepper1Number = await page.find("calcite-stepper-item[id='step-one'] >>> .stepper-item-number");
     expect(stepper1Number.textContent).toBe("1.");
+
+    stepper2.setProperty("numberingSystem", "thai");
+    await page.waitForChanges();
+    const stepper2Number = await page.find("calcite-stepper-item[id='step-two'] >>> .stepper-item-number");
+    expect(stepper2Number.textContent).toBe("๑.");
```
anveshmekala, line 609: Is it possible to avoid hardcoding the value that are non `latn` numbers so that we don't  need to change this if the localization API updates the string.

src/components/stepper/stepper.tsx
benelan: Do you mean adding an internal `numberingSystem` prop to `stepper-item` and setting it to `stepper`'s `numberingSystem`? I think I tried doing that for `date-picker` when I first implemented `numberingSystem` but I don't remember the issue I faced. @jcfranco do you have any insights on this? For reference, I just yoinked this code his PR #6563

src/components/stepper/stepper.e2e.ts
benelan: Yeah I was going to use the `numerals` in `date-picker`'s Thai nls data but they are Latin:
https://github.com/Esri/calcite-components/blob/master/src/components/date-picker/assets/date-picker/nls/th.json#L13

Is there somewhere else that we put that data now after adding translations?

src/components/stepper/stepper.tsx
benelan: Oh I see you updated his PR, I thought I was going crazy for a second there lol. Does the test I wrote in this PR work in the branch with your updates?

anveshmekala: haha yeah i updated it. yup the test included in here works in the PR https://github.com/Esri/calcite-components/pull/6563

anveshmekala: i added a test in there which kind of runs through all the supported numberingSystem. Looking back at it now, i think it is a overkill.

benelan: Yeah also we don't want to use `numberStringFormatter.numberFormatOptions` in the test because we want to make sure it's working for our users and that's an internal API. I'd replace the test in #6563 with the one in this PR and then Franco can decide which one he wants to install. Thanks for digging into this!

anveshmekala: Sounds good. will update the test in #6563.

jcfranco:
Closing in favor of https://github.com/Esri/calcite-components/pull/6563. Thank you both!

Issue #6830, opened 2023-04-22
Closed time picker popup requires tabbing out and back in to be reopened


geospatialem:
### Actual Behavior

Closing a time picker via Escape force a user to tab outside and back into the component to display again.

Spurred from #6773 cc @Elijbet @jcfranco @ashetland 

### Expected Behavior

An affordance in the picker to toggle the popups (akin to combobox) and adding support to reopen on ⬇️ key press.

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/input-time-picker/

### Reproduction Steps

1. Click sandbox above input
2. Press Tab
3. Picker popup opens
4. Press Escape
5. Picker closes

Once closed, the picker can no longer be opened unless users tab outside and back in

### Reproduction Version

1.2.0

### Priority impact

p4 - not time sensitive

### Esri team

Calcite (dev)

geospatialem:
Verified on `master`. cc @ashetland @SkyeSeitz for UI kit changes to `input-time-picker`.

PR #6832, opened 2023-04-23
feat(input-time-picker): allow toggling time picker by clicking the input or entering the down/escape key

Elijbet:
**Related Issue**: #6830

### Summary
This enhances the `input-time-picker` to allow it to be toggled via mouse and keyboard vs initial focus.

This also updates input to allow overriding `role` to accommodate other a11y structures.

### DIFF
--- a/src/components/input-time-picker/input-time-picker.e2e.ts
+++ b/src/components/input-time-picker/input-time-picker.e2e.ts
@@ -1,4 +1,4 @@
-import { newE2EPage } from "@stencil/core/testing";
+import { E2EPage, newE2EPage, E2EElement } from "@stencil/core/testing";
 import { localizeTimeString } from "../../utils/time";
 import {
   accessible,
@@ -11,6 +11,8 @@ import {
   reflects,
   renders
 } from "../../tests/commonTests";
+import { skipAnimations } from "../../tests/utils";
+import { html } from "../../../support/formatting";
 
 describe("calcite-input-time-picker", () => {
   it("renders", async () => renders("calcite-input-time-picker", { display: "inline-block" }));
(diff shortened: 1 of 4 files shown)


src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -113,9 +115,6 @@ export class InputTimePicker
     }
   }
 
-  /**
```
jcfranco, line 116: Can you restore this doc?

src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -190,6 +189,8 @@ export class InputTimePicker
 
   popoverEl: HTMLCalcitePopoverElement;
 
+  private dialogId = `time-picker-dialog--${guid()}`;
```
jcfranco, line 192: Nitpick: can you move this above so that the private properties are in alphabetical order?

jcfranco (approved):
✨🕐🤏✨

src/components/input-time-picker/input-time-picker.e2e.ts
```diff
@@ -395,4 +372,64 @@ describe("calcite-input-time-picker", () => {
...
+    });
+
+    it("does not open the time picker on input keyboard focus", async () => {
+      const popover = await page.find("calcite-input-time-picker >>> calcite-popover");
+
+      await page.keyboard.press("Tab");
+      await page.waitForChanges();
+
+      expect(await popover.getProperty("open")).not.toBe(true);
+    });
```
eriklharper, line 395: This is cool and more inline with how Chrome's native time input works.  This will also setup users to expect the focus to be more on the input than in the popup, which will transition usage nicely over to the popover-less `input-time` component.

Elijbet: Awesome! Thanks for the note.

PR #6834, opened 2023-04-24
feat(input-time-picker): add focus trap support

Elijbet:
**Description**
Related Issue: https://github.com/Esri/calcite-components/issues/6833

### <ins>Summary<ins>
Adds `focus-trap` to `time-picker` dialog to improve a11y.


src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -449,6 +500,11 @@ export class InputTimePicker
     }
   };
 
+  private restoreInputFocus(): void {
```
jcfranco, line 503: I think you can drop this method and call `calciteInputEl.setFocus()` where needed instead.

src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -84,12 +92,39 @@ export class InputTimePicker
...
+      activateFocusTrap(this, {
+        onActivate: () => {
+          if (this.focusOnOpen) {
+            this.calciteTimePickerEl.setFocus();
+            this.focusOnOpen = false;
+          }
+        }
+      });
+    } else {
+      deactivateFocusTrap(this);
```
jcfranco, line 105: Similar to `activateFocusTrap` above, can you pass an options object with `onDeactivate` where you set the following lines (restore focus and set `focusOnOpen`)? Just noticed we should do this for `input-date-picker` as well (separately).

jcfranco (approved):
➡👁️ 🕸⬅

jcfranco:
@eriklharper Can you give this a look as well and LMK if you spot anything? I can address any comments in a follow-up issue/PR.

eriklharper:
> @eriklharper Can you give this a look as well and LMK if you spot anything? I can address any comments in a follow-up issue/PR.

This looks really good, nothing jumps out at me as weird.  Nice work! :)

Issue #6835, opened 2023-04-24
[Input-Time-Picker] Support keyboard a11y improvements


Elijbet:
### Description

Related Issue: https://github.com/Esri/calcite-components/issues/6680, #6830 

Summary
Updates `input-time-picker` to support keyboard a11y improvements for its time-picker container along with descriptive text.

### Acceptance Criteria

N/A

### Which Component

input-time-picker

### Priority impact

p4 - not time sensitive

### Esri team

Calcite (dev)

geospatialem:
Verified on `master`.

PR #6837, opened 2023-04-24
fix(input-time-picker): implement dialog behavior to improve a11y

Elijbet:
**Related Issue:** #6835

## Summary
Updates `input-time-picker` to use the dialog role for its time-picker popover along with descriptive text.


src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -504,34 +530,38 @@ export class InputTimePicker
...
-            readOnly={this.readOnly}
+            readOnly={readOnly}
             scale={this.scale}
             step={this.step}
             // eslint-disable-next-line react/jsx-sort-props
             ref={this.setCalciteInputEl}
           />
         </div>
         <calcite-popover
+          aria-hidden={toAriaBoolean(!this.open)}
```
jcfranco, line 547: Since `popover` handles a lot of these roles for us, I think you only need to set `label` here. The rest are [already applied](https://github.com/Esri/calcite-components/blob/master/src/components/popover/popover.tsx#L579-L584).

jcfranco (approved):
LGTM! Can you add a `t9n` test to `input-time-picker.e2e.ts`? You may have to skip it if it fails consistently as [stated here](https://github.com/Esri/calcite-components/pull/6669#issuecomment-1514993359).

src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -543,6 +566,7 @@ export class InputTimePicker
           </div>
         </div>
         <calcite-popover
+          aria-label={messages.chooseTime}
           aria-live="polite"
```
jcfranco, line 570: Can you remove `aria-live` here? It's also handled by the popover.

jcfranco: Also `aria-label` and set `label` instead.

Elijbet: Sorry, I accidentally merged it back when resolving conflicts.

PR #6844, opened 2023-04-24
test(input-date-picker): add accessible test for input-date-picker

jcfranco:
**Related Issue:** N/A

## Summary

Wires up `accessible` test helper for input-date-picker suite. 

**Note**: this rolls back overriding `role` on `calcite-input`'s internal input (see https://github.com/Esri/calcite-components/pull/6805/). This ended up causing invalid markup as some roles require matching aria roles, which refs cannot be resolved between shadow/light DOM:

```
 ● calcite-input-date-picker › is accessible

    expect(received).toHaveNoViolations(expected)

    Expected the HTML found at $('calcite-input-date-picker,calcite-input,input[aria-label=""]') to have no violations:

    <input aria-label="" class="" inputmode="text" placeholder="MM/DD/YYYY" type="text" role="combobox">

    Received:

    "Required ARIA attributes must be provided (aria-required-attr)"

    Fix any of the following:
      Required ARIA attributes not present: aria-expanded, aria-controls
```

@geospatialem helped test and verified that both JAWS & NVDA provide relevant options regardless of this change. VoiceOver is the only one that will lack context when interpreting the component:

#### passing `role="combobox"` through to the internal input

Screenshot 2023-04-24 at 11 59 24 AM

#### not passing `role="combobox"` through to the internal input
Screenshot 2023-04-24 at 11 57 43 AM

cc @Elijbet

### DIFF
--- a/src/components/input-date-picker/input-date-picker.e2e.ts
+++ b/src/components/input-date-picker/input-date-picker.e2e.ts
@@ -7,7 +7,8 @@ import {
   floatingUIOwner,
   renders,
   hidden,
-  t9n
+  t9n,
+  accessible
 } from "../../tests/commonTests";
 import { html } from "../../../support/formatting";
 import { CSS } from "./resources";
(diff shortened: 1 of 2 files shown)


driskull (approved):
👍 

Wish that a11y with shadowDom had more documentation and support!

PR #6846, opened 2023-04-25
test: split up form-associated test helper to run a test per use case

jcfranco:
**Related Issue:** N/A

## Summary

This should help mitigate timeouts caused by additional test coverage added to the `formAssociated` test helper.

This also changes the way we have to use the helpers in E2E tests. I've updated the helper doc to reflect this as well.


Elijbet (approved):
Awesome!

PR #6847, opened 2023-04-25
fix(tooltip, popover): Support transparent backgrounds #6803

driskull:
**Related Issue:** #6798

## Summary

- Creates a `FloatingArrow` functional component to generate an SVG for the arrow.
- Uses `FloatingArrow` instead of `transform: rotate` to create the caret arrow tail.
- Introduces optional state property `floatingLayout` on `FloatingUIComponent` class for sizing the arrow properly.
- Removes CSS where we are setting the BG on two separate elements.
- Sets `Action` used in the popover to be transparent.
- Removes unnecessary styles
- Screener tests
- See: floating-ui/floating-ui#2195


src/utils/floating-ui.ts
```diff
@@ -197,8 +199,19 @@ export interface FloatingUIComponent {
    * @param delayed – (internal) when true, it will reposition the component after a delay. the default is false. This is useful for components that have multiple watched properties that schedule repositioning.
    */
   reposition(delayed?: boolean): Promise<void>;
+
+  /**
+   * Specifies the component's layout.
+   *
+   * Possible values: "vertical" or "horizontal".
```
jcfranco, line 206: Can you add some info regarding this only being needed for floating-ui owners that support arrows? Additionally, can you add a note on this using the `@State` decorator (similar to https://github.com/Esri/calcite-components/blob/be3a8b2414794562f5efa8a8223fdacd93192d1f/src/utils/locale.ts#L210 or https://github.com/Esri/calcite-components/blob/c683f6d4d52ecfffca9ffcb2ab7dc4dc758d0cad/src/utils/t9n.ts#L122)?

src/utils/floating-ui.ts
```diff
@@ -343,34 +363,55 @@ const debouncedReposition = debounce(positionFloatingUI, repositionDebounceTimeo
...
-  flipDisabled?: boolean;
-  flipPlacements?: EffectivePlacement[];
-  offsetDistance?: number;
-  offsetSkidding?: number;
-  arrowEl?: HTMLElement;
-  includeArrow?: boolean;
-  type: UIType;
-}): Promise<void> {
-  if (!referenceEl || !floatingEl || (includeArrow && !arrowEl)) {
+export async function positionFloatingUI(
```
jcfranco, line 388: ✨:star_struck:✨ Nice!

src/utils/floating-ui.ts
```diff
@@ -399,12 +440,22 @@ export async function positionFloatingUI({
     })
   });
 
-  if (middlewareData?.arrow) {
-    const { x: arrowX, y: arrowY } = middlewareData.arrow;
+  if (arrowEl && middlewareData?.arrow) {
```
jcfranco, line 443: No need for optional chaining on `middlewareData` as [it is guaranteed](https://github.com/floating-ui/floating-ui/blob/master/packages/core/src/types.ts#L87).

src/components/functional/FloatingArrow.tsx
```diff
@@ -0,0 +1,60 @@
...
+  arrowStroke: "calcite-floating-ui-arrow__stroke"
+};
+
+const DEFAULTS = {
+  width: 12,
+  height: 6,
+  strokeWidth: 1
+};
+
+export const FloatingArrow: FunctionalComponent<FloatingArrowProps> = ({
```
jcfranco, line 20: Can you add some doc on the usage of this component? This should reference `floating-ui.ts` and viceversa.

src/components/functional/ExpandToggle.tsx
```diff
@@ -3,16 +3,16 @@ import { getElementDir } from "../../utils/dom";
 import { queryActions } from "../action-bar/utils";
 import { SLOTS as ACTION_GROUP_SLOTS } from "../action-group/resources";
 import { Position, Scale } from "../interfaces";
+import { JSXAttributes } from "@stencil/core/internal";
 
-interface ExpandToggleProps {
```
jcfranco, line 7: Let's update this separately as it's unrelated to the PR's scope.

jcfranco (approved):
Awesome stuff, @driskull! 

🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯
🗯🗯▶▶▶🗯▶🗯🗯🗯🗯🗯▶🗯▶▶▶▶🗯▶▶▶▶🗯▶▶▶▶▶🗯▶🗯
🗯▶🗯🗯🗯🗯▶🗯🗯🗯🗯🗯▶🗯▶🗯🗯🗯🗯▶🗯🗯🗯🗯🗯🗯▶🗯🗯🗯▶🗯
🗯🗯▶▶🗯🗯▶🗯🗯▶🗯🗯▶🗯▶▶▶🗯🗯▶▶▶🗯🗯🗯🗯▶🗯🗯🗯▶🗯
🗯🗯🗯🗯▶🗯▶🗯▶🗯▶🗯▶🗯▶🗯🗯🗯🗯▶🗯🗯🗯🗯🗯🗯▶🗯🗯🗯🗯🗯
🗯▶▶▶🗯🗯🗯▶🗯🗯🗯▶🗯🗯▶▶▶▶🗯▶▶▶▶🗯🗯🗯▶🗯🗯🗯▶🗯
🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯

driskull:
@ashetland @macandcheese can one of you review the visual changes? Thanks!

ashetland:
This is looking great! Should the triangle be inheriting the opacity?

driskull:
>  This is looking great! Should the triangle be inheriting the opacity?

Yes, it is.

geospatialem:
> > This is looking great! Should the triangle be inheriting the opacity?
> 
> Yes, it is.

@driskull Wondering if this is truly inheriting the opacity? The more opacity that is set shows up a bit darker, it looks like set to `--calcite-ui-foreground-3`, maybe in the [floating ui scss](https://github.com/Esri/calcite-components/pull/6847/files#diff-4003a5655b0d0a9ba327a93211e690d642e5cdea753d8aea6bcf0ae78e81bb16R110)?

https://codepen.io/geospatialem/pen/GRwNvde

ashetland:
It almost looks like the triangle is underneath the shadow instead of on top of it.

driskull:
Ah, I think for the SVG we need to use `fill` and `fill-opacity` to work correctly. Do we want to use a new issue?

driskull:
nevermind, that didnt do anything :(

ashetland:
> nevermind, that didnt do anything :(

😭

Is there a border getting applied? And maybe something else happening as a result?
https://codepen.io/ashetland/pen/GRwWQyb?editors=1000
CleanShot 2023-06-28 at 10 19 22@2x

geospatialem:
> Do we want to use a new issue?

@driskull I can open the OG issue  so we have a comment trail of effort/updates.

driskull:
The stroke is defined with `"var(--calcite-ui-border-3)"`. Not sure that has anything to do.

I think the SVG fill color is just rendered differently. Its using the correct var. `var(--calcite-ui-foreground-1)`

ashetland:
Could it be an error in the SVG code?

driskull:
I don't think so. Maybe someone else could take a look.

It seems like SVG just isn't handling opacity rendering the same as a HTMLElement. Maybe its not blending in the background of what its on top of.

driskull:
Yeah its the blending of the drop shadow and the fact that the caret is over the tooltip so the stroke shows through.

Issue #6856, opened 2023-04-25
`calcite-chip` does not render when the parent is a custom element / shadow root


ethanbdev:
### Actual Behavior

When the chip is a direct child in the shadow root, it will not render.

### Expected Behavior

It should render, since 1.0.3 is not supposed to have breaking changes / it was not documented

### Reproduction Sample

https://codepen.io/eborgen/pen/OJBmJQe?editors=1010

### Reproduction Steps

Open the codepen, observe it is not rendering

### Reproduction Version

1.3.0

### Regression?

1.2.0

### Priority impact

p0 - emergency

### Impact

Need a workaround or patch as soon as possible

### Esri team

ArcGIS Web Analysis

ethanbdev:
The issue is there here:
https://github.com/Esri/calcite-components/pull/6075/files#diff-b237b0dcdb25a0941df7fe4dc75a403c120893bee940115495a9b6216358d92cR201

Will return `null` if the parent element is across the shadow boundary. I.e.
```
<custom-parent>
  <shadow-root>
    <calcite-chip />
  </shadow-root>
</custom-parent>
```

Then, when it gets to the render function,
https://github.com/Esri/calcite-components/pull/6075/files#diff-b237b0dcdb25a0941df7fe4dc75a403c120893bee940115495a9b6216358d92cR201

Gives a reference error
```
TypeError: Cannot read properties of null (reading 'label')
    at y.render (p-41dbef13.entry.js:6:3490)
    at V (p-21300df2.js:6:7216)
    at q (p-21300df2.js:6:7062)
    at p-21300df2.js:6:6879 <calcite-chip class=​"analysis-chip-button" calcite-hydrated>​…​</calcite-chip>​flex
fe @ p-21300df2.js:6
```

Maybe it should use optional chaining as I don't think a parent element should be required

ethanbdev:
Workaround is to insert a div wrapping the chip(s), so they have a light-dom parent 👍

PR #6858, opened 2023-04-25
fix(chip): Ensure Chip displays without Chip Group label in Custom Element

macandcheese:
**Related Issue:** #6856 

## Summary
Shifts responsibility of setting `aria-labelledby` to a parent Chip Group, to prevent the component from failing from lack of parent label.

I wonder if our `renders()` test helper could add a check for rendering component within a Custom Element.

### DIFF
--- a/src/components/chip-group/chip-group.e2e.ts
+++ b/src/components/chip-group/chip-group.e2e.ts
@@ -461,7 +461,7 @@ describe("calcite-chip-group", () => {
 
   it("correctly assigns aria-labelledby to child Chip elements", async () => {
     const page = await newE2EPage();
-    const label = "test-label";
+    const label = "test-group-label";
     await page.setContent(
       html`<calcite-chip-group label=${label} selection-mode="multiple">
         <calcite-chip label="test-label"></calcite-chip>
--- a/src/components/chip-group/chip-group.e2e.ts
+++ b/src/components/chip-group/chip-group.e2e.ts
@@ -458,6 +458,26 @@ describe("calcite-chip-group", () => {
     expect(await element.getProperty("selectedItems")).toHaveLength(2);
     await assertSelectedItems(page, { expectedItemIds: [chip4.id, chip5.id] });
   });
+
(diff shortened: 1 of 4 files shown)


src/components/chip-group/chip-group.tsx
```diff
@@ -193,6 +193,7 @@ export class ChipGroup implements InteractiveComponent {
       el.interactive = true;
       el.scale = this.scale;
       el.selectionMode = this.selectionMode;
+      el.setAttribute("aria-labelledby", this.label);
```
driskull, line 196: Shouldn't this just be `aria-label`? I thought labelledby takes an `id`?

src/components/chip-group/chip-group.e2e.ts
```diff
@@ -458,6 +458,26 @@ describe("calcite-chip-group", () => {
...
+        <calcite-chip id="chip-4" label="test-label"></calcite-chip>
+        <calcite-chip id="chip-5" selected label="test-label"></calcite-chip>
+      </calcite-chip-group>`
+    );
+    const chip4 = await page.find("#chip-4");
+    const chip5 = await page.find("#chip-5");
+    await page.waitForChanges();
+
+    expect(await chip4.getAttribute("aria-labelledby")).toEqual(label);
+    expect(await chip5.getAttribute("aria-labelledby")).toEqual(label);
```
driskull, line 479: Maybe also run some HTML through the a11y test for aXe verification.

src/components/chip-group/chip-group.tsx
macandcheese: @geospatialem can you advise here? Is this not needed at all?

macandcheese:
@driskull @jcfranco - Spoke with @geospatialem and we are removing the `aria-labelledby` as it is not needed for AT.

macandcheese:
Preview of fix with code sample provided in original issue: https://codepen.io/mac_and_cheese/pen/oNaWjZE?editors=1010

driskull (approved):
💥 👍

geospatialem (approved):
👍🏻

Issue #6857, opened 2023-04-25
 `--calcite-ui-brand` not inherited to  `--calcite-ui-focus-color`


geospatialem:
### Actual Behavior

When setting the `--calcite-ui-brand` to a customized color, the `--calcite-ui-focus-color` var is not overridden, and must be set by the user related to the feature in #3392

cc @macandcheese 

### Expected Behavior

When setting the `--calcite-ui-brand` to a customized color,  `--calcite-ui-focus-color` should inherit the value.

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/JjmNPOX

### Reproduction Steps

1. Open the [Codepen](https://codepen.io/mac_and_cheese/pen/JjmNPOX)
2. Observe the first button's focus color has the default `--calcite-ui-brand` color
3. Observe the second button, when using the `.themed-focus` to `--calcite-ui-focus-color` does change the focus color

### Reproduction Version

1.3.0

### Priority impact

p0 - emergency

### Impact

This seems to be a breaking change, as the focus color would change prior to the `1.3.0` release.

### Esri team

Calcite (dev)

geospatialem:
Verified in `1.3.1-next.0`

![verify-focus-color-inheritance](https://user-images.githubusercontent.com/5023024/234374547-8ecde57d-ec58-4d79-867d-1ed800410a3f.gif)

PR #6859, opened 2023-04-25
fix: fix inheritance of default focus-color value

jcfranco:
**Related Issue:** #6857 

## Summary

This fixes an issue where `--calcite-ui-focus-color`'s default was resolved at `:root` and would not allow it to inherit `--calcite-ui-brand` further down the DOM tree.


macandcheese (approved):
👍 Excellent

PR #6862, opened 2023-04-25
fix(chip): Uses correct aria role in a selection-mode:none Chip Group

macandcheese:
## Summary
Handles the case where a parent Chip Group of `selection-mode="none"` caused the Chip to incorrectly have role of "radio" set. 

cc @geospatialem


geospatialem (approved):
Nice catch - sounds great in JAWS! 🦈

jcfranco (approved):
💥

driskull (approved):
👍

PR #6829, opened 2023-04-22
feat(nav, menu): Add Nav, Nav Logo, Nav User, Menu, Menu Item components

macandcheese:
**Related Issue:** #6531 

## Summary
Adds initial implementation of these features. Future enhancements and items pushed from this MVP release can be found here: https://github.com/Esri/calcite-components/issues/6606

Issue #5564, opened 2022-10-25
Avatar and Chip - Provide alternative text for images - (2036569783)


dqateam:
### Module: 
 07b Avatar and Chip 

### Violation: 
 Provide alternative text for images 

### WCAG Reference: 
- [1.1.1 Non-text Content](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html) 

### Instance ID: 
 2036569783 

### Severity: 
 10 

### Description: 
 [Issue]
There are meaningful images without a textual equivalent. Examples include:
- Test 2. Medium Avatar Thumbnail
- Test 4. Small Avatar Thumbnail

[User Impact]
Screen reader users will be unable to determine what these images represent.

[Code Reference] 
 ```
N/A
 ``` 

### Note: 
 [Suggestion]
Provide alternative text for images. Meaningful images must have a concise but descriptive textual equivalent. To add a textual equivalent to an ``` <img> ``` element, set its alt attribute to a descriptive value. To add a textual equivalent to a ``` <svg> ``` element, add an aria-label attribute set to the desired equivalent.

[Compliant Code Example] 
 ```
N/A
 ```
 

### Media Type: 
 Images 

### Additional Resources: 
- Report Source: Product accessibility evaluation conducted on the Calcite Design System in August, 2022.

geospatialem:
For reference, the [testing page](https://geospatialem.github.io/calcite-samples/avatar.html) for tests 2 and 4.

geospatialem:
For a solution, we should pursue requiring an existing prop, such as `username` (or similar) to assign as an `aria-label` to provide context to assistive technologies.

Alternatively we could have a default `aria-label` assigned, and when the user provides a value to the `username` (or similar prop), assign the value to the `aria-label`.

geospatialem:
Verified in `1.4.0-next.9` with Windows JAWS and NVD. Also includes the fix for #5563 where the role is set to figure and provided an `aria-label`.

Issue #5563, opened 2022-10-25
Avatar and Chip - Ensure content is visible to assistive technologies - (2036569782)


dqateam:
### Module: 
 07b Avatar and Chip 

### Violation: 
 Ensure content is visible to assistive technologies 

![image](https://user-images.githubusercontent.com/5023024/202801089-657d9090-0047-4bbe-8294-b61fc7c73c17.png)

### WCAG Reference: 
- [1.1.1 Non-text Content](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html)
- [4.1.2 Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html) 

### Instance ID: 
 2036569782 

### Severity: 
 10 

### Description: 
 [Issue]
There are elements that are hidden for assistive technology users using aria-hidden arrtibute. Example included:
- mj below Test1. Medium Avatar
- mj below Test 3. Small Avatar

[User Impact]
Assistive technology users might miss the information which is provided.

[Code Reference] 
 ```
<span aria-hidden="true" class="initials">MJ</span>

 ``` 

### Note: 
 [Suggestion]
Developers should ensure that content is available for all the assistive technology users. Remove aria-hidden attribute from the elements.
Provide hidden text as "Michael Jordan" to make it descriptive.

[Complaint Code Example]
``` <span class="initials"> ```MJ
``` <span class="sr-only"> ```Michael Jordan``` </span> ```
``` </span> ```
 

### Media Type: 
 Page Structure 

### Additional Resources: 
- Report Source: Product accessibility evaluation conducted on the Calcite Design System in August, 2022.

geospatialem:
Maybe we can expose text similar to the suggestion from a prop, possibly making it required?

geospatialem:
Could be coupled with #5564

geospatialem:
Completed as part of #5564.

PR #6873, opened 2023-04-27
feat(navigation, navigation-logo, navigation-user): Add navigation, navigation-logo & navigation-user components.

anveshmekala:
**Related Issue:** #6531 

## Summary
This PR adds the following components:

- calcite-navigation
- calcite-navigation-user
- calcite-navigation-logo

Co-authored by @macandcheese 
Extracted from #6829 

#### Basic

```html
<calcite-shell>
  <calcite-navigation slot="header">
    <calcite-chip-group slot="content-center">
      <calcite-chip>nav item 1</calcite-chip>
      <calcite-chip>nav item 2</calcite-chip>
      <calcite-chip>nav item 3</calcite-chip>
    </calcite-chip-group>
  </calcite-navigation>
</calcite-shell>
```

### Properties

| Property             | Attribute           | Description                                                                                              | Type      | Default     |
| -------------------- | ------------------- | -------------------------------------------------------------------------------------------------------- | --------- | ----------- |
| `label` _(required)_ | `label`             | When `navigationAction` is `true`, specifies the label of the `calcite-action`.                          | `string`  | `undefined` |
| `navigationAction`   | `navigation-action` | When `true`, displays a `calcite-action` and emits a `calciteNavActionSelect` event on selection change. | `boolean` | `false`     |

### Events

| Event                           | Description                                                                         | Type                |
| ------------------------------- | ----------------------------------------------------------------------------------- | ------------------- |
| `calciteNavigationActionSelect` | When `navigationAction` is true, emits when the displayed action selection changes. | `CustomEvent<void>` |

#### `setFocus() => Promise<void>`

When `navigation-action` is `true`, sets focus on the component's action element.

##### Returns

Type: `Promise<void>`

### CSS Custom Properties

| Name                                | Description                                          |
| ----------------------------------- | ---------------------------------------------------- |
| `--calcite-navigation-background`   | Specifies the background color of the component.     |
| `--calcite-navigation-border-color` | Specifies the border color of the component.         |
| `--calcite-navigation-width`        | Specifies the width of the component's content area. |

#### Depends on

- [calcite-action](../action)

----------------------------------------------

#### Basic

```html
<calcite-navigation-logo active thumbnail="./_assets/images/esri-logo.svg"></calcite-navigation-user>
```

### Properties

| Property      | Attribute     | Description                                                                                                                               | Type      | Default     |
| ------------- | ------------- | ----------------------------------------------------------------------------------------------------------------------------------------- | --------- | ----------- |
| `active`      | `active`      | When true, the component is highlighted.                                                                                                  | `boolean` | `undefined` |
| `description` | `description` | A description for the component, which displays below the `heading`.                                                                      | `string`  | `undefined` |
| `heading`     | `heading`     | Specifies heading text for the component, such as a product or organization name.                                                         | `string`  | `undefined` |
| `href`        | `href`        | Specifies the URL destination of the component, which can be set as an absolute or relative path.                                         | `string`  | `undefined` |
| `label`       | `label`       | Describes the appearance or function of the `thumbnail`. If no label is provided, context will not be provided to assistive technologies. | `string`  | `undefined` |
| `rel`         | `rel`         | Defines the relationship between the `href` value and the current document.                                                               | `string`  | `undefined` |
| `target`      | `target`      | Specifies where to open the linked document defined in the `href` property.                                                               | `string`  | `undefined` |
| `thumbnail`   | `thumbnail`   | Specifies the `src` to an image.                                                                                                          | `string`  | `undefined` |

#### `setFocus() => Promise<void>`

Sets focus on the component.

##### Returns

Type: `Promise<void>`

----------------------------------------------

#### Basic

```html
<calcite-navigation-user full-name="Jhon Doe" user-id="Jhon123" active></calcite-navigation-user>
```

### Properties

| Property       | Attribute       | Description                                                                                                              | Type      | Default     |
| -------------- | --------------- | ------------------------------------------------------------------------------------------------------------------------ | --------- | ----------- |
| `active`       | `active`        | When true, the component is highlighted.                                                                                 | `boolean` | `undefined` |
| `fullName`     | `full-name`     | Specifies the full name of the user.                                                                                     | `string`  | `undefined` |
| `label`        | `label`         | Describes the appearance of the avatar. If no label is provided, context will not be provided to assistive technologies. | `string`  | `undefined` |
| `textDisabled` | `text-disabled` | When `true`, hides the `fullName` and `username` contents.                                                               | `boolean` | `false`     |
| `thumbnail`    | `thumbnail`     | Specifies the`src` to an image (remember to add a token if the user is private).                                         | `string`  | `undefined` |
| `userId`       | `user-id`       | Specifies the unique id of the user.                                                                                     | `string`  | `undefined` |
| `username`     | `username`      | Specifies the username of the user.                                                                                      | `string`  | `undefined` |

#### `setFocus() => Promise<void>`

Sets focus on the component.

##### Returns

Type: `Promise<void>`

#### Depends on

- [calcite-avatar](../avatar)


src/components/nav/nav.tsx
```diff
@@ -0,0 +1,116 @@
...
+  //--------------------------------------------------------------------------
+  //
+  //  Properties
+  //
+  //--------------------------------------------------------------------------
+
+  /**
+   * When true, display a hamburger  icon and emits `calciteNavMenuActionSelect` event on user interaction.
+   */
+  @Prop({ reflect: true }) displayMenuAction = false;
```
anveshmekala, line 39: open for suggestions regarding the prop name

src/components/nav-logo/nav-logo.tsx
```diff
@@ -0,0 +1,96 @@
+import { Component, Element, EventEmitter, h, Host, Prop, Event } from "@stencil/core";
+import { CSS } from "./resources";
+
+@Component({
+  tag: "calcite-nav-logo",
+  styleUrl: "nav-logo.scss",
+  shadow: true
```
driskull, line 7: Should these new components have delegatesFocus? Applies to all new components

src/components/nav-logo/nav-logo.tsx
```diff
@@ -0,0 +1,96 @@
...
+  //--------------------------------------------------------------------------
+  @Element() el: HTMLCalciteNavLogoElement;
+
+  //--------------------------------------------------------------------------
+  //
+  //  Public Properties
+  //
+  //--------------------------------------------------------------------------
+  /** When `true`, visually highlight the component */
+  @Prop({ reflect: true }) active: boolean;
```
driskull, line 23: Naming: should this be selected or is active correct terminoligy?

src/components/nav-logo/nav-logo.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+  /** Specifies the href destination of the component */
+  @Prop({ reflect: true }) href: string;
+
+  /** Specifies accesible label for the component */
+  @Prop() label: string;
+
+  /** Specifies the subtext to display, for example an organization or application description */
+  @Prop({ reflect: true }) subText: string;
```
driskull, line 32: Do we need to reflect this prop?

src/components/nav-logo/nav-logo.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+  /** Specifies the text to display, for example a product name */
+  @Prop({ reflect: true }) text: string;
```
driskull, line 35: Do we need to reflect this prop?

src/components/nav-logo/nav-logo.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+  /** When `true`, makes `text` and `subText` visible */
+  @Prop({ reflect: true }) textEnabled: boolean;
+
+  /** Specifies the `src` to an image  */
+  @Prop({ reflect: true }) thumbnail: string;
```
driskull, line 41: Do we need to reflect this prop?

src/components/nav-logo/nav-logo.tsx
```diff
@@ -0,0 +1,96 @@
...
+  //
+  // --------------------------------------------------------------------------
+
+  private clickHandler = (): void => {
+    this.calciteNavLogoSelect.emit();
+  };
+
+  private keyDownHandler = (event: KeyboardEvent): void => {
+    if (event.key === "Enter" || event.key === " ") {
+      this.calciteNavLogoSelect.emit();
```
driskull, line 64: should this event be `preventDefault()`?

src/components/nav-logo/nav-logo.tsx
```diff
@@ -0,0 +1,96 @@
...
+    return (
+      <Host>
+        <a
+          aria-label={this.label}
+          href={this.href}
+          onClick={this.clickHandler}
+          onKeyDown={this.keyDownHandler}
+          tabIndex={0}
+        >
+          {this.thumbnail && <img src={this.thumbnail} />}
```
driskull, line 83: should the `img` tag have a key if its conditional?

src/components/nav-logo/nav-logo.tsx
```diff
@@ -0,0 +1,96 @@
...
+          {(this.text || this.subText) && this.textEnabled && (
+            <div class={CSS.textContainer}>
```
driskull, line 85: should this div have a key?

driskull: Applies to all other conditional nodes.

src/components/nav-user/nav-user.tsx
```diff
@@ -0,0 +1,104 @@
...
+  @Prop() fullName: string;
+
+  /** Specifies accesible label for the component */
+  @Prop() label: string;
+
+  /** Specifies the unique id of the user. */
+  @Prop({ reflect: true }) userId: string;
+
+  /** Specifies the subtext to display, for example a user organization or role */
+  @Prop({ reflect: true }) username: string;
```
driskull, line 36: Should this be reflected?

src/components/nav-user/nav-user.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  /** When `true`, hides the `fullName` and `username`. */
+  @Prop({ reflect: true }) hideText = false;
+
+  /** Specifies the `src` to an image to display in the Avatar (remember to add a token if the user is private). */
+  @Prop({ reflect: true }) thumbnail: string;
```
driskull, line 42: Should this be reflected?

src/components/nav-user/nav-user.tsx
```diff
@@ -0,0 +1,104 @@
...
+  //
+  // --------------------------------------------------------------------------
+
+  private clickHandler = (): void => {
+    this.calciteNavUserSelect.emit();
+  };
+
+  private keyDownHandler = (event: KeyboardEvent): void => {
+    if (event.key === "Enter" || event.key === " ") {
+      this.calciteNavUserSelect.emit();
```
driskull, line 65: should default be prevented here?

src/components/nav-user/resources.ts
```diff
@@ -0,0 +1,5 @@
+export const CSS = {
+  textContainer: "text-container",
+  userFullName: "user-full-name",
```
driskull, line 3: seems weird with `userFullName` and `userUsername`. Can we make these consistent? Maybe just `username` and `fullname`?

src/components/nav/nav.tsx
```diff
@@ -0,0 +1,116 @@
...
+  styleUrl: "nav.scss",
+  shadow: true
+})
+
+/**
+ * @slot logo - A slot for adding a `calcite-logo` component to the primary nav level.
+ * @slot user - A slot for adding a `calcite-user` component to the primary nav level.
+ * @slot progress - A slot for adding a `calcite-progress` component to the primary nav level.
+ * @slot primary-content-start - A slot for adding a `calcite-menu`, `calcite-action`, or other interactive elements in the start position of the primary nav level.
+ * @slot primary-content-center - A slot for adding a `calcite-menu`, `calcite-action`, or other interactive elements in the center position of the primary nav level.
```
driskull, line 15: Do we need to say `primary-content-center` or could we just do `primary-content`?

src/components/nav/nav.tsx
driskull: This actually renders a calcite-action, not an icon. Should the doc be updated?

Instead of having a prop for this, why not just let the user slot a calcite-action here? If we do need a name, how about just `menu` or `menuEnabled`?

src/components/nav/nav.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  /**
+   * When true, display a hamburger  icon and emits `calciteNavMenuActionSelect` event on user interaction.
+   */
+  @Prop({ reflect: true }) displayMenuAction = false;
+
+  /**
+   * Specifies the label of the hamburger icon.
+   */
+  @Prop({ reflect: true }) label!: string;
```
driskull, line 44: does this need to reflect?

src/components/nav/nav.tsx
```diff
@@ -0,0 +1,116 @@
...
+  };
+
+  //--------------------------------------------------------------------------
+  //
+  //  Render Methods
+  //
+  //--------------------------------------------------------------------------
+
+  renderMenuAction(): VNode {
+    return <calcite-action icon="hamburger" onClick={this.clickHandler} text={this.label} />;
```
driskull, line 75: I would prefer allowing a user to slot this so they can modify the action properties like indicator, active, etc.

src/components/nav/nav.tsx
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  renderNavLevel(level: "primary" | "secondary" | "tertiary"): VNode {
+    const progress = this.el.querySelector('[slot="progress"]');
```
driskull, line 79: I think we should find a way to not have to query the dom for a slot. If the slot changes, this won't change unless another render happens. I think this is a pattern we want to try and avoid if possible.

src/components/nav/nav.tsx
```diff
@@ -0,0 +1,116 @@
...
+    const logo = this.el.querySelector('[slot="logo"]');
+    const user = this.el.querySelector('[slot="user"]');
+    const start = this.el.querySelector(`[slot="${level}-content-start"]`);
+    const center = this.el.querySelector(`[slot="${level}-content-center"]`);
+    const end = this.el.querySelector(`[slot="${level}-content-end"]`);
+    const showMenu =
+      (level === "primary" && (logo || user || this.displayMenuAction)) || center || start || end;
+    return showMenu ? (
+      <div
```
driskull, line 88: conditional div should probably have a key

driskull: applies to other nodes

stencil.config.ts
```diff
@@ -12,51 +12,51 @@ export const create: () => Config = () => ({
   namespace: "calcite",
   bundles: [
     { components: ["calcite-accordion", "calcite-accordion-item"] },
-    { components: ["calcite-action"] },
     { components: ["calcite-action-bar"] },
-    { components: ["calcite-action-menu"] },
```
driskull, line 17: Is the menu still here?

driskull (changes requested):
Initial review! LOoking good! 👀

src/components/nav/nav.tsx
macandcheese: Have gone back and forth on slot vs. prop and supplied UI here - the usual pro/cons apply…

Providing this would lead to more consistent UX across platform, but of course necessitates accepting a label string, etc.

stencil.config.ts
macandcheese: Might have been a merge issue as this PR shouldn’t have touched this component.

src/demos/_assets/images/dashboard-logo.svg
macandcheese, line 1: Is there a better way to use assets in demo pages? Maybe we can use url if these are publicly available to not include in project.

src/components/nav-logo/nav-logo.scss
```diff
@@ -0,0 +1,54 @@
...
+  & a {
+    @apply flex m-0 items-center justify-center cursor-pointer transition-default focus-base no-underline;
+    border-block-end: 2px solid transparent;
+  }
+  & img {
+    @apply flex h-7 m-0;
+  }
+}
+
+a:hover,
```
driskull, line 12: nitpicky, but maybe we should use a class for these instead of an element tag. What if this component adds another anchor tag in its shadowRoot.

driskull: Same with `img`

driskull: Applies to other CSS files as well.

src/components/nav/nav.tsx
macandcheese: Yeah supporting both a built-in “hamburger” and allowing folks to slot something else if they really need to makes sense to me.

Hesitant to call the prop only “menu” since we would only be rendering the “action that emits an event devs can use to open a Sheet / Panel” - “menu-toggle”?

driskull: `menu-action` or `menu-toggle` makes sense to me

src/components/nav-logo/nav-logo.tsx
macandcheese: I think `active` aligns with List, `selected` would align with Dropdown, Tab. No preference on this one.

src/components/nav/nav.tsx
macandcheese: Discussed in other comment - seems like providing this + a slot of the same name that overrides the property seems to work.

src/components/nav/nav.tsx
driskull: I don't think theres any need to add `enable` onto the name. When you add it as a HTML attribute or set it to `true` that is already enabling it.

src/components/nav-logo/nav-logo.tsx
anveshmekala: yup, these components need `delegatesFocus`.

src/components/nav-logo/nav-logo.tsx
anveshmekala: the reason behind `active` is the UX we are displaying to the user. for `selected` user expects the focus kind of border for the component where as for `active` it is just border-bottom.

src/components/nav-logo/nav-logo.tsx
anveshmekala: this is not required.

src/components/nav-logo/nav-logo.tsx
anveshmekala: nope. removed .

src/components/nav-logo/nav-logo.tsx
anveshmekala: i don't think its required in this case, using key would recreate the node which isn't required in this case.

src/components/nav-user/nav-user.tsx
anveshmekala: yup.

src/components/nav-user/nav-user.tsx
anveshmekala: removed.

src/components/nav-user/nav-user.tsx
anveshmekala: removed.

src/components/nav/nav.tsx
anveshmekala: since we have both `primary-content-start` & `primary-content-end` , i think its better to be descriptive by adding `content` at the end.

stencil.config.ts
anveshmekala: menu will be added in a separate PR.

src/components/nav/nav.tsx
anveshmekala: nope, removed.

stencil.config.ts
driskull: @anveshmekala calcite-action-menu already exists.

src/components/nav/nav.tsx
driskull: It's inconsistent with list-item though.

https://github.com/Esri/calcite-components/blob/3b2f4a0377ff63d67f42b43e28b1bbe8b2b7d251/src/components/list-item/list-item.tsx#L43-L47

src/components/nav/nav.tsx
macandcheese: `menu-action` sounds good - this will become more valuable once we have a Sheet component - have some ideas about leveraging that and a slot so folks won't have to set that up themselves but... this will be useful either way in the interim.

macandcheese: Only other thought here is to update with `nav` terminology - `nav-action` - since it may or may not invoke a Menu.

src/components/nav/nav.stories.ts
```diff
@@ -0,0 +1,109 @@
...
+      </calcite-nav>
+    </calcite-shell>
+  `;
+
+export const primaryAndSecondarySlots_TestOnly = (): string =>
+  html`
+    <calcite-shell>
+      <calcite-nav slot="header">
+        <calcite-chip slot="logo"> logo </calcite-chip>
+        <calcite-action slot="nav-action" text="menu" icon="hamburger"> </calcite-action>
```
macandcheese, line 49: For these can we just use the `nav-action` prop and not the slot so we can test what the component renders here? 

Or if we want to test the slot on some, just use the Chip as the slot content like the others?

anveshmekala: there is a test which adds both `nav-action` prop and slot to test out what component renders. I can add a separate one just for the prop. 

thought using `calcite-action` in `nav-action` slot is the recommended practice.

macandcheese: It is for that slot in practice, but this is just showing the slot positioning is correct, like we are not placing a `calcite-nav-logo/user` in the `logo/user` slot. For now I think we want to incentivize the property over the slot, especially in a case like this where it's the same as what the component renders. For the purposes of this test it could just be a chip indicating what's there is correct.

src/components/nav/nav.tsx
macandcheese: Because `content-center` isn't the most common slotting location, having the extra `-center` makes sense in this case, IMO. In List, (Modal, etc.), the `content` slot is "the main area".

driskull: @jcfranco what do you think?

stencil.config.ts
anveshmekala: my bad, thought it was `calcite-menu`. added it back.

src/components/nav-logo/nav-logo.scss
anveshmekala: Since this component is very basic, may be we can add a class when there is a requirement to add more anchor element or img.

src/components/nav-logo/nav-logo.stories.ts
```diff
@@ -0,0 +1,70 @@
...
+  </calcite-nav>
+</calcite-shell>`;
+
+export const textAndThumbnail_TestOnly = (): string => html` <calcite-shell>
+  <calcite-nav slot="header">
+    <calcite-nav-logo slot="logo" text="ArcGIS Online" thumbnail="${placeholderImage({ width: 50, height: 50 })}" />
+  </calcite-nav>
+</calcite-shell>`;
+
+export const subTextAndThumbnail_TestOnly = (): string => html` <calcite-shell>
```
anveshmekala, line 51: I don't think using subText only is a common use-case and a separate test isn't required IMO. Can we combine `subText` tests with `text` ones?

src/components/nav/nav.tsx
```diff
@@ -0,0 +1,201 @@
...
+  }
+})
+
+/**
+ * @slot logo - A slot for adding a `calcite-logo` component to the primary nav level.
+ * @slot user - A slot for adding a `calcite-user` component to the primary nav level.
+ * @slot progress - A slot for adding a `calcite-progress` component to the primary nav level.
+ * @slot menu-action - A slot for adding a `calcite-action` component to the primary nav level.
+ * @slot primary-content-start - A slot for adding a `calcite-menu`, `calcite-action`, or other interactive elements in the start position of the primary nav level.
+ * @slot primary-content-center - A slot for adding a `calcite-menu`, `calcite-action`, or other interactive elements in the center position of the primary nav level.
```
geospatialem, line 31: Would "primary nav level", "secondary nav level", "tertiary..." be the same as "component's primary level", "component's secondary level", "component's tertiary..."?

src/components/nav/nav.tsx
```diff
@@ -0,0 +1,201 @@
...
+  @State() teritiarySlotHasElements: boolean;
+
+  // --------------------------------------------------------------------------
+  //
+  //  Events
+  //
+  // --------------------------------------------------------------------------
+
+  /**
+   * Emits whenever the component is selected or unselected.
```
geospatialem, line 85: > When `navAction` is `true`, emits when the component's selection changes.

src/components/nav/nav.tsx
```diff
@@ -0,0 +1,201 @@
...
+  @Element() el: HTMLCalciteNavElement;
+
+  //--------------------------------------------------------------------------
+  //
+  //  Properties
+  //
+  //--------------------------------------------------------------------------
+
+  /**
+   * When true, display a hamburger `calcite-action` and emits `calciteNavMenuActionSelect` event on user interaction.
```
geospatialem, line 53:
```
When `true`, displays a hamburger icon and emits a `calciteNavMenuActionSelect` event on selection change.
```

src/components/nav-user/nav-user.tsx
```diff
@@ -0,0 +1,111 @@
...
+          aria-label={this.label}
+          onClick={this.clickHandler}
+          onKeyDown={this.keyDownHandler}
+          tabIndex={0}
+        >
+          <calcite-avatar
+            full-name={this.fullName}
+            thumbnail={this.thumbnail}
+            user-id={this.userId}
+          />
```
geospatialem, line 91: Per our Teams discussion, we'll need some solutions from #5563 and #5564 to address a11y/context for AT.

src/components/nav-logo/nav-logo.tsx
```diff
@@ -0,0 +1,105 @@
...
+  //
+  //--------------------------------------------------------------------------
+  @Element() el: HTMLCalciteNavLogoElement;
+
+  //--------------------------------------------------------------------------
+  //
+  //  Public Properties
+  //
+  //--------------------------------------------------------------------------
+  /** When `true`, visually highlight the component */
```
geospatialem, line 24: doc consistency recommendation:

> When `true`, the component is highlighted.

src/components/nav-user/nav-user.tsx
```diff
@@ -0,0 +1,111 @@
...
+  //
+  //--------------------------------------------------------------------------
+  /** When `true`, visually highlight the component */
+  @Prop({ reflect: true }) active: boolean;
+
+  /** Specifies the text to display, for example a user name or full name */
+  @Prop() fullName: string;
+
+  /** Specifies accesible label for the component */
+  @Prop() label: string;
```
geospatialem, line 32: doc consistency recommendation - if needed, depends on https://github.com/Esri/calcite-components/issues/5563 and https://github.com/Esri/calcite-components/issues/5564, perhaps?:

> Specifies the label of the component.

src/components/nav-user/nav-user.tsx
```diff
@@ -0,0 +1,111 @@
...
+  /** Specifies the unique id of the user. */
+  @Prop() userId: string;
+
+  /** Specifies the subtext to display, for example a user organization or role */
+  @Prop() username: string;
+
+  /** When `true`, hides the `fullName` and `username`. */
+  @Prop({ reflect: true }) hideText = false;
+
+  /** Specifies the `src` to an image to display in the Avatar (remember to add a token if the user is private). */
```
geospatialem, line 43: doc consistency recommendation:

```
Specifies the `src` to an image to display to the `calcite-avatar` (remember to add a token if the user is private).

**Review by geospatialem — COMMENTED** [review]
Looking great! 👍🏻 

Have a few doc-related recommendations for consistency, and some feedback of changing the `label` to an `alt` on the `nav-logo` for a11y support.

File: src/components/nav-logo/nav-logo.tsx [hunk 30]
```diff
@@ -0,0 +1,105 @@
...
+  /** Specifies the `src` to an image  */
+  @Prop() thumbnail: string;
+
+  //--------------------------------------------------------------------------
+  //
+  //  Events
+  //
+  //--------------------------------------------------------------------------
+
+  /** Emits when user select the component. */
```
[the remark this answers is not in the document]
anveshmekala: i see that most of the components uses either `select` or `selects` . Are we planning to update them to `selecting`?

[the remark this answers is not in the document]
geospatialem: Good catch on consistency - we're aiming for present tense throughout. How about:

> Emits when the component's selection changes.

src/components/nav/nav.tsx
anveshmekala: based on the feedback [here](https://github.com/Esri/calcite-components/pull/6873#discussion_r1179656339) changed the hamburger icon to hamburger `calcite-action` . 

cc: @driskull

macandcheese: Can we not use `hamburger`? I know it is the icon name but maybe just "displays a `calcite-action`" ?

geospatialem: I'm on board with removing the hamburger reference, if Anvesh and Matt agree. It seems a bit confusing just as `calcite-action` though, what would folks think about (a similar change would need to occur on the `label` prop for consistency as well):

```
When `true`, displays a `calcite-action` menu and emits a `calciteNavMenuActionSelect` event on selection change.
```

src/components/nav/nav.tsx
geospatialem: Good point, how about a reference to the `calcite-action` for the second reference? Perhaps something like:

> When `navAction` is `true`, emits when the `calcite-action` selection changes.

src/components/nav-logo/nav-logo.stories.ts
anveshmekala: discussed in teams and keeping the tests for now . If the screenshot count exceeds then we can merge them into a single test. The benifit of keeping them separate is to check indentations.

src/components/nav/nav.tsx
anveshmekala: > ```
> When `true`, displays a `calcite-action` menu and emits a `calciteNavMenuActionSelect` event on selection change.
> ```

This sounds good to me. Wanted to confirm its `calciteNavActionSelect` not `calciteNavMenuActionSelect` .

src/components/nav/nav.tsx
anveshmekala: i think its better to mention `calcite-action` **menu** since user can slot calcite-action in any of the slots available and assume selection of those will emit the event.

macandcheese: I still don't think we want to use `menu` with this property - it could be used to launch something without a menu. This would be the only action we are rendering from inside the `nav` component.

What about:

`When `navAction` is true, emits when the user interacts with the displayed action.`

anveshmekala: i think menu makes sense since we are displaying menu kind of icon. This can help eliminate the confusion between other slotted elements which can be `calcite-action`'s . And the event is emitted only on user interaction with hamburger icon.

src/components/nav/nav.tsx
```diff
@@ -0,0 +1,198 @@
...
+  //--------------------------------------------------------------------------
+  //
+  //  Render Methods
+  //
+  //--------------------------------------------------------------------------
+
+  renderMenuAction(): VNode {
+    return (
+      <slot name={SLOTS.navAction} onSlotchange={this.handleMenuActionSlotChange}>
+        {this.navAction && (
```
driskull, line 146: I don't think you need the check for `this.navAction` because this is fallback content by default. If a user adds this slot then the content inside it won't render.

src/utils/dom.ts
```diff
@@ -367,6 +367,7 @@ export function isPrimaryPointerButton(event: PointerEvent): boolean {
  * @param elements An array of elements
  * @param currentElement The current element
  * @param destination The target destination element to focus
+ * @param cycle Should navigation cycle through elements or stop at extent - defaults to true
```
driskull, line 370: Should we add a spec test for this?

driskull (approved):
👍 💯 🥇

src/components/nav/nav.tsx
anveshmekala: didn't knew that. Will update this.

src/components/nav-user/nav-user.tsx
```diff
@@ -0,0 +1,106 @@
...
+  @Prop() label: string;
+
+  /** Specifies the unique id of the user. */
+  @Prop() userId: string;
+
+  /** Specifies the username of the user. */
+  @Prop() username: string;
+
+  /** When `true`, hides the `fullName` and `username` contents. */
+  @Prop({ reflect: true }) hideText = false;
```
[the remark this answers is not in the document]
anveshmekala: I think `hideName` makes more sense here.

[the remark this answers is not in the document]
driskull: I think for consistency, `thing-hidden` or `thing-disabled` makes more sense. Thats how we've been naming properties that hide or turn off some functionality.

The only other component that uses `somethingHidden` is ColorPicker. cc @jcfranco 

@jcfranco for consistency, we should probably decide which makes more sense. We had already settled on `-disabled` for turning off functionality, do we need to distinguish for hiding certain pieces of UI? If so, it probably makes more sense for feature`-hidden` right?

hidden also aligns with native HTML better.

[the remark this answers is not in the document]
anveshmekala: i see the point here. 

Thinking back at the `*-disabled` props, may be we should change these also into `hidden` . disabled can mean the functionality is visible but user cant interact with it.  Thoughts?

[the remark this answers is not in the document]
driskull: I think it depends on what they do. disabling functionality vs just disabling UI.

src/utils/dom.ts
anveshmekala: Added.

src/components/nav/nav.tsx
anveshmekala: updated to 
```
When `navAction` is `true`, emits when the displayed action selection changes.
```

src/components/nav/nav.tsx
macandcheese: Agreed sounds good

src/components/nav-user/nav-user.tsx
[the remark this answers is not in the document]
jcfranco: This has come up before and we decided to adopt `<foo>Disabled`/`<foo>Enabled` pattern (see https://github.com/Esri/calcite-components/issues/2115). Worth noting that the color-picker has recently deprecated its `hide-<foo>` props in favor of the aforementioned naming.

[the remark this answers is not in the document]
driskull: Sounds good to me. So enabled/disabled can refer to functionality and UI. 👍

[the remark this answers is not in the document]
anveshmekala: `textDisabled ` will be restored in place of `hideText`.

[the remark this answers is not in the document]
anveshmekala: `nameDisabled` sounds better to me . Regarding two props one for each ,It depends on how often users choose to hide name but show username and vice-versa. 
Even with design requirements ,  If a user has space to display username and don't want to show the fullName they can simply ignore it while rendering the UI. I don't think we need separate variables for this.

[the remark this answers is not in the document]
macandcheese: I think ‘textDisabled’ is sufficient with meaningful doc and it’s a reciprocal of ‘textDisabled’ nomenclature we have elsewhere Action, etc.. ‘name-‘ could be confused for only hiding one of the texts.

src/components/nav/nav.tsx
jcfranco: After some back and forth, we'll be moving forward with `x-content-center`. Reasoning:

* unlike list-item, any slot can be used as the main content slot
* center slot is purely placing and not replacing content (compared to list-item)
* `content` appears to be the most common/appropriate word for that area in navigation bars
* `center` matches CSS terminology (along with `start`/`end`)

src/components/nav-user/nav-user.tsx
anveshmekala: #5563 and #5564 are resolved.

src/components/nav-user/nav-user.tsx
```diff
@@ -0,0 +1,107 @@
...
+  /** Specifies the unique id of the user. */
+  @Prop() userId: string;
+
+  /** Specifies the username of the user. */
+  @Prop() username: string;
+
+  /** When `true`, hides the `fullName` and `username` contents. */
+  @Prop({ reflect: true }) textDisabled = false;
+
+  /** Specifies the `src` to an image to display in the Avatar (remember to add a token if the user is private). */
```
jcfranco, line 43: > remember to add a token if the user is private

Without additional context, I think this seems out of place.

src/components/nav-user/nav-user.stories.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export default {
+  title: "Components/Nav/Nav User",
+  parameters: {
+    notes: readme
+  },
+  ...storyFilters()
+};
+
+export const simple = (): string =>
```
jcfranco, line 15: In order to keep these stories fully-focused on navigation user, would it be possible to:

* only use the nav user component here
* either update navigation stories to include an all-navigation-components story or a separate stories file to cover that?

Applies to other stories.

src/components/nav-logo/usage/Basic.md
```diff
@@ -0,0 +1,3 @@
+```html
+<calcite-nav-logo  active  thumbnail="./_assets/images/esri-logo.svg"></calcite-nav-user>
```
jcfranco, line 2: Nitpick: there's double space between attributes here.

src/components/nav-user/nav-user.tsx
```diff
@@ -0,0 +1,107 @@
...
+  @Prop() fullName: string;
+
+  /** Describes the appearance of the avatar. If no label is provided, context will not be provided to assistive technologies. */
+  @Prop() label: string;
+
+  /** Specifies the unique id of the user. */
+  @Prop() userId: string;
+
+  /** Specifies the username of the user. */
+  @Prop() username: string;
```
jcfranco, line 38: Sidebar: we should probably revisit this way in the future, but the casing of `username` and `fullName`/`full-name` does stand out when both are used. I get the reasoning, but something I noticed.

src/components/nav-logo/nav-logo.tsx
```diff
@@ -0,0 +1,104 @@
...
+  @Prop() thumbnail: string;
+
+  //--------------------------------------------------------------------------
+  //
+  //  Events
+  //
+  //--------------------------------------------------------------------------
+
+  /** Emits when the component's selection changes. */
+  @Event() calciteNavLogoSelect: EventEmitter<void>;
```
jcfranco, line 52: Since there is no selection state, how is this (and other select events for navigation and menu components) different from a "click" event? If there's no distinction, I'd suggest removing them and have developers rely on the event target or provide specific event handlers to these components. 

For background: we did something similar some time ago with `action` and removed the redundant custom click event.

jcfranco (approved):
@anveshmekala @macandcheese Once again, great stuff!

🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭
🧭⛵⛵⛵⛵🧭⛵🧭🧭🧭🧭🧭⛵🧭⛵⛵⛵⛵🧭🧭⛵⛵⛵🧭🧭⛵⛵🧭🧭⛵🧭🧭🧭⛵🧭⛵⛵⛵⛵🧭
🧭⛵🧭🧭⛵🧭⛵🧭🧭🧭🧭🧭⛵🧭⛵🧭🧭🧭🧭⛵🧭🧭🧭🧭⛵🧭🧭⛵🧭⛵⛵🧭⛵⛵🧭⛵🧭🧭🧭🧭
🧭⛵⛵⛵⛵🧭⛵🧭🧭⛵🧭🧭⛵🧭⛵⛵⛵🧭🧭🧭⛵⛵🧭🧭⛵🧭🧭⛵🧭⛵🧭⛵🧭⛵🧭⛵⛵⛵🧭🧭
🧭⛵🧭🧭⛵🧭⛵🧭⛵🧭⛵🧭⛵🧭⛵🧭🧭🧭🧭🧭🧭🧭⛵🧭⛵🧭🧭⛵🧭⛵🧭🧭🧭⛵🧭⛵🧭🧭🧭🧭
🧭⛵🧭🧭⛵🧭🧭⛵🧭🧭🧭⛵🧭🧭⛵⛵⛵⛵🧭⛵⛵⛵🧭🧭🧭⛵⛵🧭🧭⛵🧭🧭🧭⛵🧭⛵⛵⛵⛵🧭
🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭🧭

Several comments from https://github.com/Esri/calcite-components/pull/6901/ are also applicable here, but not showstoppers by any means.

The only things I'd like to see addressed before merging are:

* `nav` -> `navigation` (impacts all naming)
* dropping of select events if there is no substantial difference between a click (impacts menu-item)

src/components/nav-user/usage/Basic.md
```diff
@@ -0,0 +1,3 @@
+```html
+<calcite-nav-user full-name="Jhon Doe" user-id="jhon123" active></calcite-nav-user>
```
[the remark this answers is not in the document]
macandcheese: Tell that to this guy https://www.transfermarkt.us/jhon-jhon/profil/spieler/964582

src/components/nav-user/nav-user.tsx
macandcheese: Taken from Avatar so should address there as well if needed.

src/components/nav/nav.tsx
```diff
@@ -0,0 +1,198 @@
...
+  //--------------------------------------------------------------------------
+  //
+  //  Properties
+  //
+  //--------------------------------------------------------------------------
+
+  /**
+   * When `true`, displays a `calcite-action` and emits a `calciteNavActionSelect` event on selection change.
+   */
+  @Prop({ reflect: true }) navAction = false;
```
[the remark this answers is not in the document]
macandcheese: This one has been discussed before, and found naming tough. If we want to just leave it as a slot for now we can do that and add this functionality in when we have the Sheet to better support it (this was always kind of a stopgap until we could build the Sheet and an associated slot and handle hiding / showing that with the built-in action).

src/components/nav/nav.tsx
```diff
@@ -0,0 +1,198 @@
...
+          {level === LEVEL.primary && (
+            <slot name={SLOTS.logo} onSlotchange={this.handleLogoSlotChange} />
+          )}
+          <slot
+            name={`${level}-content-start`}
+            onSlotchange={(event) => this.handleContentSlotChange(event, level)}
+          />
+          <slot
+            name={`${level}-content-center`}
+            onSlotchange={(event) => this.handleContentSlotChange(event, level)}
```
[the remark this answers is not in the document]
anveshmekala: Awesome!

src/components/nav/nav.scss
```diff
@@ -0,0 +1,109 @@
...
+slot[name="tertiary-content-start"] {
+  @apply mr-auto;
+}
+
+slot[name="primary-content-end"],
+slot[name="secondary-content-end"],
+slot[name="tertiary-content-end"],
+slot[name="logo"] ~ slot[name="user"],
+slot[name="user"]:only-child {
+  @apply ml-auto;
```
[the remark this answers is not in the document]
anveshmekala: Yup, updated to use tailwind logical props util classes.

src/components/nav-user/nav-user.tsx
```diff
@@ -0,0 +1,107 @@
...
+  // --------------------------------------------------------------------------
+  //
+  //  Render Methods
+  //
+  // --------------------------------------------------------------------------
+
+  render() {
+    return (
+      <Host>
+        <a onClick={this.clickHandler} onKeyDown={this.keyDownHandler} tabIndex={0}>
```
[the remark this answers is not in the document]
anveshmekala: `<a>` should have a href attribute to help the user agent add `tabIndex=0`  default value.

src/components/nav-user/nav-user.tsx
anveshmekala: Thoughts on adding additional context @geospatialem  ?

src/components/nav-user/nav-user.stories.ts
anveshmekala: Updated navigation component stories to utilize navigation-logo and navigation-user.

src/components/navigation-logo/navgation-logo.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  // --------------------------------------------------------------------------
+  //
+  //  Render Methods
+  //
+  // --------------------------------------------------------------------------
+  render(): VNode {
+    return (
+      <Host>
+        <a href={this.href} tabIndex={0}>
```
driskull, line 53: Is href an optional property? Was it considered that the user could just wrap the component in their own <a> so that the user can set other anchor attributes like rel, title, target, etc?

src/components/navigation/navigation.tsx
```diff
@@ -0,0 +1,214 @@
...
+  //
+  //  Render Methods
+  //
+  //--------------------------------------------------------------------------
+
+  renderMenuAction(): VNode {
+    return (
+      <slot name={SLOTS.navAction} onSlotchange={this.handleMenuActionSlotChange}>
+        {this.navAction && (
+          <calcite-action icon="hamburger" onClick={this.clickHandler} text={this.label} />
```
driskull, line 149: nitpick: would be nice if the icons were in the resources file.

src/components/navigation-logo/navgation-logo.tsx
macandcheese: In other places we accept href like Button, we also accept rel and target props, which we could add here. 

Asking to wrap a component in anchor seems like it would introduce a lot of potential style, focus, behavior fragmentation.

src/components/navigation/navigation.tsx
```diff
@@ -0,0 +1,214 @@
...
+          <slot
+            data-level={level}
+            name={
+              isPrimaryLevel
+                ? SLOTS.primaryContentStart
+                : level === LEVEL.secondary
+                ? SLOTS.secondaryContentStart
+                : SLOTS.tertiaryContentStart
+            }
+            onSlotchange={this.handleContentSlotChange}
```
driskull, line 179: Question: rather than having specific slots for levels, would it be more future proof and HTML like if the content could just be nested? 

This is more inline with other components we have like combobox-item where you can nest one inside the other in order to have levels.

Currently, it is hard restricted to 3 levels and requires a user to really define that level. It could also lead to users adding a tertiary level without a primary level.

More future proof would be to just allow nesting of components.

driskull: cc @macandcheese

src/components/navigation-logo/navgation-logo.tsx
driskull: yeah makes sense. 👍

src/components/navigation/navigation.tsx
macandcheese: Do you mean nesting additional nav components inside the parent nav? The primary reason for having the levels is to enforce nav logo and nav user at the top, as well as the primary nav has different height. If we add more support for nested nav, we’d probably need to introduce a scale to achieve the various levels, and this could break platform consistency quickly. 

What would the markup look like for nested nav? It would simplify the theming and css variable story which would be great, but we might lose consistency of implementation. We don’t want logo / user to ever reside under the primary navigation.

Not opposed if that makes the markup simpler and doesn’t affect intended design.

driskull: I think the way it is is fine as long as we are committed to 3 levels.

src/components/navigation-user/navigation-user.tsx
```diff
@@ -0,0 +1,81 @@
...
+  // --------------------------------------------------------------------------
+  //
+  //  Render Methods
+  //
+  // --------------------------------------------------------------------------
+
+  render(): VNode {
+    return (
+      <Host>
+        <a tabindex={0}>
```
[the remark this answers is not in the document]
anveshmekala: updated.

src/components/navigation/navigation.tsx
macandcheese: Took a quick shot at something like you describe - it simplifies the component a bit and seems to work well IMO - https://github.com/Esri/calcite-components/tree/macandcheese/slotted-navs

This ultimately may be more flexible - we can still can enforce certain slotted components in the "primary" nav, it will just require a bit of doc.

src/components/menu-item/menu-item.scss
```diff
@@ -16,6 +15,10 @@
   @apply w-full;
 }
 
+:host(:not([layout="vertical"])) {
+  @apply h-full;
+}
```
anveshmekala, line 20: This change is related to visual regression found while debugging navigation component.

src/components/navigation-user/navigation-user.e2e.ts
```diff
@@ -0,0 +1,36 @@
...
+        propertyName: "active",
+        value: "true"
+      },
+      {
+        propertyName: "textDisabled",
+        value: ""
+      }
+    ]));
+
+  it("defaults", async () =>
```
jcfranco, line 29: Nitpick: `async` not needed since the promise is returned.

src/components/navigation-user/usage/Basic.md
```diff
@@ -0,0 +1,3 @@
+```html
+<calcite-navigation-user full-name="Jhon Doe" user-id="Jhon123" active></calcite-navigation-user>
```
jcfranco, line 2: Typo: `Jhon`, `Jhon123`

src/components/navigation/navigation.e2e.ts
```diff
@@ -0,0 +1,58 @@
...
+    const hamburgerMenu = await page.find(`calcite-navigation >>> calcite-action`);
+
+    await page.keyboard.press("Tab");
+    expect(eventSpy).toHaveReceivedEventTimes(0);
+
+    await page.keyboard.press("Enter");
+    expect(eventSpy).toHaveReceivedEventTimes(1);
+
+    await page.keyboard.press("Space");
+    await page.waitForChanges();
```
jcfranco, line 49: I don't think `waitForChanges` is needed here as it wasn't needed for `Enter`.

src/components/navigation/navigation.tsx
```diff
@@ -0,0 +1,221 @@
...
+  /** When navAction is true, emits when the displayed action selection changes.*/
+  @Event({ cancelable: false }) calciteNavActionSelect: EventEmitter<void>;
+
+  //--------------------------------------------------------------------------
+  //
+  //  Private Methods
+  //
+  //--------------------------------------------------------------------------
+
+  private clickHandler = () => {
```
jcfranco, line 98: Should this be `actionClickHandler`?

src/components/navigation/navigation.tsx
```diff
@@ -0,0 +1,221 @@
...
+    }
+  };
+
+  private handleProgressSlotChange = (event: Event): void => {
+    if (this.isPrimaryLevel()) {
+      this.progressSlotHasElement = slotChangeHasAssignedElement(event);
+    }
+  };
+
+  private isPrimaryLevel = (): boolean => {
```
jcfranco, line 155: This can be a class function instead of an assigned bound function. `this` is already set where it's called.

src/utils/dom.ts
```diff
@@ -489,6 +489,7 @@ export const focusElementInGroup = (
   const isFirstItem = currentIndex === 0;
   const isLastItem = currentIndex === elements.length - 1;
 
+
```
jcfranco, line 492: There's an extra newline here. 🤓

jcfranco (approved):
Awesome stuff, @anveshmekala! 

Once the `setFocus` and navigation action prop/event renames are addressed, this should be good to merge!

src/components/navigation-logo/navgation-logo.tsx
```diff
@@ -0,0 +1,74 @@
...
+  @Prop({ reflect: true }) active: boolean;
+
+  /** Specifies the URL destination of the component, which can be set as an absolute or relative path.*/
+  @Prop({ reflect: true }) href: string;
+
+  /** Describes the appearance or function of the `thumbnail`. If no label is provided, context will not be provided to assistive technologies. */
+  @Prop() label: string;
+
+  /** Specifies the subtext to display, such as an organization or application description. */
+  @Prop() subtext: string;
```
driskull, line 34: Should this maybe be named `description` to be consistent with other props?

src/components/navigation-user/navigation-user.tsx
```diff
@@ -0,0 +1,81 @@
...
+  //--------------------------------------------------------------------------
+  //
+  //  Public Properties
+  //
+  //--------------------------------------------------------------------------
+  /** When true, the component is highlighted.*/
+  @Prop({ reflect: true }) active: boolean;
+
+  /** Specifies the text to display, such as a username or full name.*/
+  @Prop() fullName: string;
```
driskull, line 29: doc: This doc says this could be a username but theres already a username prop.

src/components/navigation-user/navigation-user.tsx
```diff
@@ -0,0 +1,81 @@
...
+  //
+  //  Render Methods
+  //
+  // --------------------------------------------------------------------------
+
+  render(): VNode {
+    return (
+      <Host>
+        <button aria-label={this.label}>
+          <calcite-avatar
```
driskull, line 56: note: I wonder if it might make more sense to just have them slot the `<calcite-avatar>` instead of this component having all the same props?

src/components/navigation-logo/navgation-logo.tsx
```diff
@@ -0,0 +1,74 @@
...
+  // --------------------------------------------------------------------------
+  //
+  //  Render Methods
+  //
+  // --------------------------------------------------------------------------
+
+  render(): VNode {
+    return (
+      <Host>
+        <a href={this.href} tabindex={0}>
```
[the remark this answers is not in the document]
anveshmekala: the tabIndex still needs to be set here because the default 0 is applied only when href has a valid value or an empty string. I tried adding empty string which results in reload of page. Later changed to "#" which will a11y instructions.

driskull (approved):
Nice! 👍 💯

src/components/navigation-logo/navgation-logo.tsx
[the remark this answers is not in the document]
driskull: Maybe you could do `tabindex={!this.href ? 0 : null}`. href should be a required prop though right?

src/components/navigation/navigation.tsx
anveshmekala: the reason for having this is to avoid having different scope for `this` when used inside `onSlotchange` event handlers.

src/components/navigation-user/navigation-user.tsx
anveshmekala: good catch, this needs to be updated to just full name.

src/components/navigation-logo/navgation-logo.tsx
anveshmekala: we can use description here considering we change the `text` attribute to something like `name` or `label`.

cc @jcfranco

src/components/navigation-logo/navgation-logo.tsx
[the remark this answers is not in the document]
anveshmekala: `tabindex={!this.href ? 0 : null}` this will over-ride the default tabindex value `0` with `null` when href is defined and `href` is not a required prop.

src/components/navigation/navigation.e2e.ts
anveshmekala: tests are failing without `waitForChanges` on `Space` key . I see that `waitForChanges` is added in many other instances on `Space` keydown.

src/components/navigation-logo/navgation-logo.tsx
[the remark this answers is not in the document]
anveshmekala: Should we add `rel` & `target` prop here?

src/components/navigation-user/navigation-user.tsx
anveshmekala: The reason for having a separate component is to restrict the users from setting the scale which can cause layout issues in navigation context me thinks.  We can considering re-using `avatar` component by rendering `username` & `fullName`  in the DOM and adjusting scaling based on CSS selectors to identify if its slotted inside navigation component.

WDYT @macandcheese @jcfranco

src/components/navigation-logo/navgation-logo.tsx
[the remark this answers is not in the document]
driskull: maybe `href` should just be required?

src/components/navigation-user/navigation-user.tsx
driskull: We could always modify the scale of the avatar onslotchange.

src/components/navigation-logo/navgation-logo.tsx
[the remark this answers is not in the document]
macandcheese: I think there are times where clicking the logo might just emit an event to reset a route or something it might not always be a href to navigate.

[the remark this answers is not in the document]
driskull: If that's the case, it should probably follow `calcite-button` logic and render either a button or a anchor depending on the existence of href.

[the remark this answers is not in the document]
anveshmekala: i don't think `href` should be a required attribute for logo. Is there an added advantage of rendering a button when href is `undefined` besides click event being emitted on `Space` keydown ?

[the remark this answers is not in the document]
driskull: I think it would just be a11y concerns. An anchor should take you somewhere (either on the page or off) whereas as button should perform some action.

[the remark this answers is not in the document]
anveshmekala: when href is `undefined` & text-enabled is `false` , VoiceOver is reads out the logo as ` {label} name, image`.
when href is `undefined` & text-enabled is `true` ,  VoiceOver focuses the image reading the instruction as ` {label} name, image`  and then user can move on to the text which is read out.
 
 The only time VoiceOver recognizes logo as `link` is when user adds href.  When both `href` and `label` are `undefined` , the instructions recognizes logo as `clickable` element.
 
 Tested with VoicOver and safari 16.4

src/components/navigation-logo/navgation-logo.tsx
anveshmekala: changed to heading and description. Wondering if we should drop `textEnabled` prop and render `heading` & `description` if provided by the user.

src/components/nav-user/nav-user.tsx
geospatialem: I still think the token is important to note here. Maybe some context clarity can help?

Can we update a bit to fit the same description used in Avatar, where it specifies as an image.

Perhaps something like:

```
Specifies the `src` to an image (remember to add a token if the user is private).
```

Otherwise, if referencing Avatar makes sense here in the doc, how about the following:

```
Specifies the `src` of the `calcite-avatar` image (remember to add a token if the user is private).
```

src/components/navigation-logo/navgation-logo.tsx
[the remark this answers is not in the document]
anveshmekala: synced-up with kitty today. 

Updates regarding navigation-logo a11y:

- tabindex default value is removed  because logo shouldn't be in the focus order when href is null value
- AT user should be able to hear out image  (if label is defined), heading & description context 
- Screen readers will recognize logo as link if href is defined.

src/components/navigation/navigation.tsx
jcfranco: All of the `onSlotchange` event handlers already have `this` bound to the instance. Hence why we don't need it here.

Issue #6886, opened 2023-04-28
FlowItem. Inherit action-bar slot


asangma:
### Description

Let FlowItem use the action-bar slot that Panel has.

### Acceptance Criteria

Work the same way it works with Panel

### Relevant Info

Sometimes we need an action bar when we're drilled in.

### Which Component

FlowItem

### Example Use Case

When an ActionBar is needed in a "drilled-in" experience.

### Priority impact

p4 - not time sensitive

### Esri team

N/A

geospatialem:
Verified in `1.4.0-next.7`

![verify-action-bar-slot-flow](https://user-images.githubusercontent.com/5023024/236321692-9e61d8a7-e239-4aac-a5c2-f0cd791b2a34.gif)

PR #6887, opened 2023-04-28
feat(flow-item): Add action bar slot

driskull:
**Related Issue:** #6886

## Summary

- Adds action-bar slot to flow item so it can be passed to the panel internally.


macandcheese (approved):
🎈 Nice!

asangma (approved):
Skadoosh!

Issue #6758, opened 2023-04-10
[Action Bar] Improve border display when slotted in Panel's `action-bar` slot


macandcheese:
### Description

Brought up in https://github.com/Esri/calcite-components/pull/6738, when `bottom-actions` is populated and the Action Bar is in a Panel's `action-bar` slot, an undesired border is visible at the right extent of the Action Bar. Might be being set by the Shell Panel?

https://user-images.githubusercontent.com/4733155/230954220-b41f6e5d-4290-495f-b8a8-e9c4ef32fa78.mov

### Acceptance Criteria

There is no visible border at either edge of the Action Bar when slotted.

### Which Component

Shell Panel / Action Bar / Panel

### Priority impact

p4 - not time sensitive

### Esri team

Calcite (design)

geospatialem:
Verified in `1.4.0-next.7`

PR #6888, opened 2023-04-28
feat(action-bar): Improve border display in horizontal layout

macandcheese:
**Related Issue:** #6758 

## Summary
Adds a css case to prevent the border at the edge of horizontal action bar bottom-actions.

Before:

Screen Shot 2023-04-28 at 3 05 45 PM

After:

Screen Shot 2023-04-28 at 3 05 38 PM

### DIFF
--- a/src/components/action-bar/action-bar.scss
+++ b/src/components/action-bar/action-bar.scss
@@ -50,6 +50,10 @@
   }
 }
 
+:host([layout="horizontal"][expand-disabled]) ::slotted(calcite-action-group:last-of-type) {
+  @apply border-r-0;
+}
+
 ::slotted(calcite-action-group:last-child) {
   @apply border-b-0 border-r-0;
 }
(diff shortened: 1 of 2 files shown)


src/components/action-bar/action-bar.scss
```diff
@@ -50,6 +50,10 @@
   }
 }
 
+:host([layout="horizontal"][expand-disabled]) ::slotted(calcite-action-group:last-of-type) {
+  @apply border-r-0;
```
driskull, line 54: hmm instead of border-right should we be using border-block-end or border-inline-end? Does this look ok in RTL?

driskull (approved):
👍

src/components/action-bar/action-bar.scss
macandcheese: Ha, yeah it looks good in `rtl` because we were only applying the border with `border-r`. I'll update the file to use logical css properties.

macandcheese: Updated @driskull

driskull (approved):
👍 ✨

Issue #6890, opened 2023-05-01
Center row action bar does not display correctly


mpayson:
### Actual Behavior

Center row action bar displays horizontally by default, this is a breaking change and the documentation says vertical should be the default

Also, there seems to be a bug where setting the layout to vertical does not work immediately, at least in our app? And setting `layout=vertical` in the reproduction case does not have an effect

https://user-images.githubusercontent.com/5892789/235479646-6743491a-e704-42fc-9ae4-897887e21e4a.mov

### Expected Behavior

Center row action bar displays vertically

### Reproduction Sample

https://codepen.io/mpayson/pen/jOeGWvq

### Reproduction Steps

1. Open code pen
2. See center row is horizontal

### Reproduction Version

1.3.0

### Regression?

1.2.0

### Priority impact

p0 - emergency

### Impact

Unable to update to the latest calcite version, set the priority to emergency since it's a breaking change, similar to: https://github.com/Esri/calcite-components/issues/6856

### Esri team

ArcGIS Online

macandcheese:
Yeah, only Shell Panel should have had this, looks like a quick revert will fix - https://github.com/Esri/calcite-components/pull/6745/files#diff-157b42657037cdcaaea0a7d90c9c9cda7769d7573329df56da34f7738147e3c4R81

PR #6891, opened 2023-05-01
fix(shell-center-row): Correctly do not set Action Bar layout

macandcheese:
**Related Issue:** Fix for #6890 

## Summary
I missed this during review, it should only be applying this to Shell Panel. Adds test only Story to capture this state.


driskull (approved):
👍

Issue #6875, opened 2023-04-27
composed-offset-position is a devDepencency but is used in a dynamic import in utils/floating-ui.ts


rweber-esri:
### Actual Behavior

When Hub bumps from `@esri/calcite-components@1.07` to `@esri/calcite-components@1.3.1`, we observe JS errors when we trigger any popover to open. The problem seems to relate to [this dynamic import](https://github.com/Esri/calcite-components/blob/3ba1b50e6a76a64ef75fe7df35a7aca769fee1bb/src/utils/floating-ui.ts#L64).  I noticed that the `composed-offset-position` package is a [devDependency](https://github.com/Esri/calcite-components/blob/3ba1b50e6a76a64ef75fe7df35a7aca769fee1bb/package.json#L119) of `@esri/calcite-components@1.3.1`, but I'm curious if this should be in `dependencies` vs `devDependencies`? We're able to resolve the JS errors on our end by adding `composed-offset-position@0.0.4` as a `devDependency` to our project's package.json, but I feel like we probably shouldn't need to do this? Interestingly, I do not observe the same JS error from the [codepen template](https://codepen.io/pen?template=RwgrjEx) linked from the `Reproduction Sample` section below when I add a `calcite-popover`.

### Expected Behavior

Hub (or any client) doesn't need to explicitly add `composed-offset-position@0.0.4` to our project dependencies.

### Reproduction Sample

NA

### Reproduction Steps

NA

### Reproduction Version

1.3.1

### Priority impact

p4 - not time sensitive

### Esri team

N/A

jcfranco:
@rweber-esri Agreed, `composed-offset-position` should have been a core dependency. We'll correct this soon. Thanks!

PR #6895, opened 2023-05-02
fix(deps): move `composed-offset-position` to dependencies

jcfranco:
**Related Issue:** #6875

## Summary

`composed-offset-position` was unintentionally installed as a dev dependency and would require developers to install it themselves.


driskull (approved):
👍

Issue #6743, opened 2023-04-06
Create generic item layout component


driskull:
### Description

Create generic item layout component

This component would have the following slots to allow users to layout content better...

- actions-start
- content-start
- DEFAULT content slot
- content-end
- actions-end

### User Stories

This will help users be able to layout content that is slotted within a panel footer for example. It will also allow a use to disable specific parts within the component if necessary since some parent components don't allow disabling partial UI.

### Acceptance Criteria

Name of component can be decided later.

### Relevant Info

- actions-start
- content-start
-(default slot) for cotent
- content-end
- actions-end

related to #5664

### Helpful Details

This would be able to emulate the layout of a calcite-panel header or a calcite-list-item 

```tsx
<calcite-layout-item>
<calcite-action text="select" icon="checkmark" slot="actions-start" />
<calcite-avatar slot="content-start" / >
My Content
<calcite-action text="delete" icon="trash" slot="actions-end" />
</calcite-layout-item>
```

### Priority impact

p2 - want for current milestone

### Esri team

N/A

driskull:
cc @geospatialem @jcfranco can we get this triaged?

macandcheese:
Does it complicate this unnecessarily to have a `layout-group` or parent by another name to wrap these `layout-item` component? That could control spacing / margin around the children, any potential alignment or orientation options, etc.

driskull:
I'm not sure it needs a parent since this layout component since wherever its slotted should handle that.

geospatialem:
> cc @geospatialem @jcfranco can we get this triaged?

@driskull You thinking for the May release?

driskull:
yeah that sounds good

geospatialem:
Verified on `master` with the demo pages, including use of the two CSS props: 
- `--calcite-stack-padding-inline` and
- `--calcite-stack-padding-block`

Example use:
```html
<calcite-panel>
  <div slot="header-content">Panel</div>
  <p>Content</p>
  <calcite-stack slot="footer">
    <calcite-action text="select" icon="check" slot="actions-start"></calcite-action>
    <calcite-input disabled></calcite-input>
    <calcite-avatar slot="content-start" thumbnail="http://placekitten.com/105/105" scale="s"> </calcite-avatar>
    <calcite-chip slot="content-end" value="chip" scale="s" appearance="outline">A chip</calcite-chip>
    <calcite-action text="delete" icon="trash" slot="actions-end"></calcite-action>
  </calcite-stack>
[...]

Issue #5664, opened 2022-10-28
Allow list-item action slots to have an independent disabled state form the list-item


ethanbdev:
### Description

The use case is that a list item (GP tool) is disabled, but we want to show an information icon that the user can interact with either on hover / click to give help to the user on why this item (GP tool) is disabled. This would occur when the licensing is not sufficient to use the tool and we need to direct them to information on getting the license / seeing more information about the licensing level.

Right now, the actions inherit the disabled state from the list-item, so there is no way to provide user interaction on those actions.

### Acceptance Criteria

The scope of the issue would be the list-item respecting the `disabled` attribute on itself and any slotted actions independently.

### Relevant Info
![image](https://user-images.githubusercontent.com/52869490/198727917-3deddbac-3ec5-451b-8549-45db2e4404fd.png)
![image](https://user-images.githubusercontent.com/52869490/198727952-6272e011-ccb0-41e8-abe0-455af941410f.png)

### Which Component

calcite-list-item

### Example Use Case

https://codepen.io/eborgen/pen/WNyvagJ

### Esri team

ArcGIS Map Viewer

driskull:
@jcfranco @benelan @eriklharper  What are your thoughts about how disabling could work here?

- Would we introduce another property to disable only the content of a list item?
  - `content-disabled`?
  - Would we introduce a mode? `disableMode="content";  // probably not its ugly.
- Would we update `disabled` to not affect slotted actions? If so, a user would have to disable them separately.
  - If we do this, we would need to update the interaction helper to not set pointer-events for the whole component.

Other ideas?

benelan:
@geospatialem any a11y concerns with a disabled component having non-disabled child(ren)?

geospatialem:
> @geospatialem any a11y concerns with a disabled component having non-disabled child(ren)?

@benelan This sounds like a very solid use case. From the a11y perspective, its not a known or expected pattern, so would need to consult with a few other a11y professionals to determine the best path forward on the enhancement. Can report back here in a few weeks once a few of us can convene together.

driskull:
Yeah if we disable the component it will get `aria-disabled` on the host. I'm not sure if that will affect slotted children or not.

nattarnoff:
Disabled is not a valid attribute on the `<li>`. What needs to happen here is the disabled attribute needs to be applied to the button. It is also preferred to use this attribute over the aria-disabled attribute.
`<li>`
[...]

PR #6903, opened 2023-05-02
feat(stack): Adds new `Stack` component to arrange content and actions

driskull:
**Related Issue:** #6743 #5664

## Summary

- Adds new component `Stack` to render a row of actions and content.
- Can be used to slot into places where no layout options are present.
- Can be used to disable specific parts of slotted content.

## Background

This will help users be able to layout content that is slotted within a panel footer for example.

- The panel footer slot doesn't offer a way to slot things at the start or end and this would allow them the ability to do that without having to dig into CSS or apply padding.

It will also allow a use to disable specific parts within the component if necessary since some parent components don't allow disabling partial UI.

- Instead of disabling the component, the user would use this component and slot it into the component. The user can disable only a portion of slotted content inside this component.

## Questions

- Name of this component


src/components/layout-item/layout-item.scss
```diff
@@ -0,0 +1,71 @@
+:host {
```
macandcheese, line 1: Should we consider a margin variable or other spacing modifier like we have on input or rely on the content placed inside?

src/components/layout-item/layout-item.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * @slot - A slot for adding content.
+ * @slot actions-start - A slot for adding actionable `calcite-action` elements before the content of the component.
+ * @slot content-start - A slot for adding non-actionable elements before content of the component.
+ * @slot content-end - A slot for adding non-actionable elements after content of the component.
+ * @slot actions-end - A slot for adding actionable `calcite-action` elements after the content of the component.
+ */
+@Component({
+  tag: "calcite-layout-item",
```
macandcheese, line 14: Is this going to be public or internal component? The “-item” portion of the name makes me think there would be a parent of some kind - layout / layout-group.

src/components/layout-item/layout-item.scss
```diff
@@ -0,0 +1,71 @@
...
+  @apply flex flex-col flex-1;
+}
+
+.container {
```
macandcheese, line 5: Should we consider a bordered option or rely on users to add that if needed?

src/components/layout-item/usage/Basic.md
```diff
@@ -0,0 +1,9 @@
+```html
```
macandcheese, line 1: For best practices, what is the distinction between using this and using List with selection mode none?

src/components/layout-item/layout-item.scss
driskull: I think we can rely on users or where this is being slotted to handle that.

driskull:
@macandcheese @jcfranco can we deprecate `detachedHeightScale` in favor of just `heightScale` and then in the doc we can say it will only work for `displayMode="float"`? That would give us freedom to add other displayModes where this scale may apply in the future

src/components/layout-item/layout-item.tsx
macandcheese: ‘calcite-content-row’ / ‘layout-row’ (if we don’t expect it to have a “column” layout) ?

src/components/layout-item/layout-item.scss
driskull: We can include padding by default and allow a user to remove it with a variable if that sounds good?

src/components/layout-item/layout-item.tsx
driskull: Good question. We can always just start with row layout and not have column until the need arises.

driskull:
@macandcheese @jcfranco I looked at other web component libraries to find a name for this component and the closest seems to be `stack`.

The other closest I saw was `slats`. https://ux.mailchimp.com/patterns/slats

We could also go with something more generic like:

- box
- section
- group

Let me know what you think so I can get the name finalized.

driskull:
Can I get a review plz :) ✨

src/components/stack/stack.tsx
```diff
@@ -0,0 +1,126 @@
...
+  shadow: true
+})
+export class Stack {
+  //--------------------------------------------------------------------------
+  //
+  //  Properties
+  //
+  //--------------------------------------------------------------------------
+
+  /**  When `true`, content interaction is prevented and the displayed with lower opacity. */
```
jcfranco, line 24: Typo: `prevented and the displayed`

src/components/stack/stack.stories.ts
```diff
@@ -0,0 +1,45 @@
+import { modesDarkDefault } from "../../../.storybook/utils";
+import readme from "./readme.md";
+import { html } from "../../../support/formatting";
+import { storyFilters } from "../../../.storybook/helpers";
+
+export default {
+  title: "Components/Stack",
```
jcfranco, line 7: WDYT about adding this under `Components/Support/` to better convey it's purpose for now?

jcfranco (approved):
💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲
💲💲💵💵💵💲💵💵💵💵💵💲💵💵💵💵💲💲💵💵💵💲💵💲💲💵💲💲💵💵💵💲
💲💵💲💲💲💲💲💲💵💲💲💲💵💲💲💵💲💵💲💲💲💲💵💲💵💲💲💵💲💲💲💲
💲💲💵💵💲💲💲💲💵💲💲💲💵💵💵💵💲💵💲💲💲💲💵💵💲💲💲💲💵💵💲💲
💲💲💲💲💵💲💲💲💵💲💲💲💵💲💲💵💲💵💲💲💲💲💵💲💵💲💲💲💲💲💵💲
💲💵💵💵💲💲💲💲💵💲💲💲💵💲💲💵💲💲💵💵💵💲💵💲💲💵💲💵💵💵💲💲
💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲💲

Awesome! This component is going to be extremely useful.

PR #6911, opened 2023-05-03
test(list): skip unstable test

driskull:
Skips unstable list test.

The filter component needs to be refactored to be able to access its data after changing the filterText since its debounced.


jcfranco (approved):
Should the title be "skip unstable test" instead?

Issue #6893, opened 2023-05-01
Tooltips within Shadow DOM components sometimes fail to open when focused


sonofflynn89:
### Actual Behavior

According to our accessibility standards, tooltips should appear when their trigger is hovered over OR when their trigger is focused via the keyboard. However, tooltips in a Shadow DOM environment do not always open when their trigger is focused via the keyboard. I spent a few days trying to figure out what conditions cause this and here is what I've found:
- 2+ tooltips must be placed within a Shadow Root
- Clicking on the page and tabbing to the first trigger will show the first tooltip. However, tabbing to the second trigger _will not show the second tooltip_

I made a codepen and recorded the behavior. I also showed that creating the same scenario with the Light DOM does not produce the same error:

https://user-images.githubusercontent.com/30124777/235545686-2482f6a8-e99a-4d38-a32a-c5ae30c1f49a.mov

### Expected Behavior

Focusing on the 2nd tooltip trigger within a Shadow Root should display the 2nd tooltip, just like it does in the Light DOM

### Reproduction Sample

https://codepen.io/sonofflynn89-the-reactor/pen/poxWRLY

### Reproduction Steps

- Go to the codepen in a Chrome window (must be v111 or later for declarative Shadow DOM to work)
- Hover over both buttons in the `No Shadow DOM` section, note that both tooltips appear
- Hover over both buttons in the `With Shadow DOM` section, note that both tooltips appear
- Click above the `No Shadow DOM` section
- Hit the tab button and focus both buttons in the `No Shadow DOM` section, note that both tooltips appear
- Hit the tab button and focus the _first_ button in the `With Shadow DOM` section, note that the tooltip appears
- Hit the tab button and focus the _second_ button in the `With Shadow DOM` section. Note that the tooltip **does not** appear

### Reproduction Version

v1.3.1

### Priority impact

p3 - want for upcoming milestone

### Impact

We are working on an incredibly complex, space-limited gallery component and were hoping to leverage icon buttons to help users toggle between different views. Since space is so limited, we opted to move the button descriptions into a popover.

However, the `calcite-popover` forces keyboard navigation users to click the `esc` button after every time an icon button comes into focus, otherwise they are locked in and cannot navigate anywhere else. Not a great user experience.

We hoped to leverage the auto close-on-tab capability of `calcite-tooltip` to solve this problem, only to find out that it won't work for Shadow DOM. We cannot change our complex component to use `scoped` at this time.

[...]

PR #6915, opened 2023-05-03
fix(tooltip): fix focusing tooltip when a referenceElement is within a shadowDOM

driskull:
**Related Issue:** #6893

## Summary

- When a tooltip's reference element is inside a shadowRoot, the `focusin` and `focusout` handlers are not activating
  - This is because the event handlers are on the document and wont' work inside a shadowRoot
- This PR sets up these focus handlers on each shadowRoot necessary
- Only sets up the focus listeners once per shadowRoot by keeping a count of referenceElements with the same shadowRoot.
- Adds `getShadowRootNode` dom utility to get an elements rootNode if the rootNode is a shadowRoot element.

### DIFF
--- a/src/components/tooltip/TooltipManager.ts
+++ b/src/components/tooltip/TooltipManager.ts
@@ -12,7 +12,7 @@ export default class TooltipManager {
 
   private registeredElements = new WeakMap<ReferenceElement, HTMLCalciteTooltipElement>();
 
-  private registeredShadowRoots = new WeakMap<ReferenceElement, ShadowRoot>();
+  private registeredShadowRootCounts = new WeakMap<ShadowRoot, number>();
 
   private hoverTimeout: number = null;
 
@@ -30,14 +30,18 @@ export default class TooltipManager {
 
   registerElement(referenceEl: ReferenceElement, tooltip: HTMLCalciteTooltipElement): void {
     this.registeredElementCount++;
-
     this.registeredElements.set(referenceEl, tooltip);
-
-    const shadowRoot = referenceEl instanceof Element ? getShadowRootNode(referenceEl) : null;
+    const shadowRoot = this.getReferenceElShadowRootNode(referenceEl);
 
     if (shadowRoot) {
-      this.registeredShadowRoots.set(referenceEl, shadowRoot);
-      this.addShadowListeners(shadowRoot);
+      const { registeredShadowRootCounts } = this;
+      const count = registeredShadowRootCounts.get(shadowRoot) || 0;
+
+      if (count === 0) {
+        this.addShadowListeners(shadowRoot);
+      }
+
+      registeredShadowRootCounts.set(shadowRoot, count + 1);
     }
 
     if (this.registeredElementCount === 1) {
@@ -46,9 +50,10 @@ export default class TooltipManager {
   }
 
   unregisterElement(referenceEl: ReferenceElement): void {
-    const shadowRoot = this.registeredShadowRoots.get(referenceEl);
+    const shadowRoot = this.getReferenceElShadowRootNode(referenceEl);
+    const count = this.registeredShadowRootCounts.get(shadowRoot);
 
-    if (shadowRoot) {
+    if (!count) {
       this.removeShadowListeners(shadowRoot);
     }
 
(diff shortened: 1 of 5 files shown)


src/components/tooltip/TooltipManager.ts
```diff
@@ -28,15 +30,25 @@ export default class TooltipManager {
 
   registerElement(referenceEl: ReferenceElement, tooltip: HTMLCalciteTooltipElement): void {
     this.registeredElementCount++;
-
     this.registeredElements.set(referenceEl, tooltip);
+    const shadowRoot = this.getReferenceElShadowRootNode(referenceEl);
```
jcfranco, line 34: This works assuming there's only one layer of shadow DOM, right? I believe the following would still have issues:

```html
<composite-component-a>
  <!-- shadow DOM -->
    <composite-component-b>
        <!-- shadow DOM -->
           <calcite-link id="tooltip-button">Data disclaimer</calcite-link>
           <calcite-tooltip label="Data disclaimer" reference-element="tooltip-button">
             <span>This data was collected over a 24 hour period</span>
           </calcite-tooltip>
        <!-- shadow DOM -->    
    </composite-component-b>    
  <!-- shadow DOM -->
</composite-component-a>
```

driskull: That's a good question, I haven't tested that but I would have assumed it should work since this just has to deal with focusing and once you focus into the 2nd layer of shadowDom the event would be triggered right?

driskull: For the example above, the `getReferenceElShadowRootNode` should return the 2nd layer shadow DOM right? Since the rootNode for that component would be the 2nd layer shadow DOM. right?

driskull: This does work:

```html
<template shadowrootmode="open">
    <div>
      <template shadowrootmode="open">
        <calcite-button id="shadow-trigger-1">
          <calcite-icon icon="grid" />
        </calcite-button>
        <calcite-tooltip reference-element="shadow-trigger-1" placement="bottom">
          Display Grid
        </calcite-tooltip>
        <calcite-button id="shadow-trigger-2">
          <calcite-icon icon="list" />
        </calcite-button>
        <calcite-tooltip reference-element="shadow-trigger-2" placement="bottom">
          Display List
        </calcite-tooltip>
      </template>
    </div>
  </template>
```

driskull: Tested and seems to work as expected. Getting the `getReferenceElShadowRootNode` will get the closest shadowRoot and adding focusIn/focusout events on that shadowRoot works since anything inside that shadowRoot getting focus will trigger the events.

driskull: Added test

alisonailea (approved):

src/components/tooltip/tooltip.e2e.ts
```diff
@@ -853,4 +853,68 @@ describe("calcite-tooltip", () => {
...
+          );
+        }
+
+        document.body.innerHTML = "<shadow-component-b></shadow-component-b>";
+      });
+      await page.waitForChanges();
+    }
+
+    async function isTooltipOpen(page: E2EPage): Promise<boolean> {
+      return await page.evaluate((): boolean => {
```
jcfranco, line 894: No need to `await` here since you're returning the promise.

src/components/tooltip/TooltipManager.ts
jcfranco: Awesome, thanks for testing!

jcfranco (approved):
Pure awesomeness, @driskull! 

🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍
🔍🔍👁👁👁🔍🔍👁👁🔍🔍🔍👁👁🔍🔍👁🔍🔍🔍👁🔍
🔍👁🔍🔍🔍🔍👁🔍🔍👁🔍👁🔍🔍👁🔍👁🔍🔍🔍👁🔍
🔍👁🔍🔍🔍🔍👁🔍🔍👁🔍👁🔍🔍👁🔍👁🔍🔍🔍👁🔍
🔍👁🔍🔍🔍🔍👁🔍🔍👁🔍👁🔍🔍👁🔍👁🔍🔍🔍🔍🔍
🔍🔍👁👁👁🔍🔍👁👁🔍🔍🔍👁👁🔍🔍👁👁👁🔍👁🔍
🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍🔍

Issue #6865, opened 2023-04-26
[Panel] `action-bar` slot should have the same background color as header


macandcheese:
### Description

Since Action Bar are not full width - they can show the Panel background behind, leading to undesired appearance:

![Screen Shot 2023-04-26 at 12 11 47 PM](https://user-images.githubusercontent.com/4733155/234678911-98cf7a0f-5069-456d-b811-d58b5b347046.png)

### Acceptance Criteria

I'd expect the slot to have a container with background that matched that of header, to better match the default color of action bg: 

![Screen Shot 2023-04-26 at 12 11 14 PM](https://user-images.githubusercontent.com/4733155/234678886-8851beb6-c7b8-4b75-9d51-ff4c14d55869.png)

### Which Component

Panel

### Priority impact

p4 - not time sensitive

### Esri team

Calcite (design)

driskull:
Seems like the action bar should always be full width when slotted here right?

macandcheese:
Yeah, that would also solve 👍

driskull:
@macandcheese do you have a codepen for this one? I think it could be fixed?

geospatialem:
@driskull I observed it when `expand-disabled` is `true` and few `action`s are added: https://codepen.io/geospatialem/pen/LYgOOPK. 

Possibly associated with the container?

geospatialem:
Verified in `1.4.0-next.8` with `panel` and `flow`.

![image](https://user-images.githubusercontent.com/5023024/236329948-a58a8e5a-ae6c-4805-87ed-2869d4fc7ba6.png)

![image](https://user-images.githubusercontent.com/5023024/236330306-6b275be1-8a2f-41a9-a1e2-7af633a156bf.png)

PR #6917, opened 2023-05-03
fix(action-bar): Set background color on action-bar

driskull:
**Related Issue:** #6865

## Summary

- Removes `transparent` bg color on action-bar.
- Adds screenshot test.


macandcheese (approved):
🪄

Issue #6388, opened 2023-01-31
[Shell Panel] Add Overlay display mode


macandcheese:
### Description

As proposed, add an "overlay" mode to Shell Panel that allows it to position over any content adjacent to it in the shell.

- Does not extend over header / footer slots (use case for upcoming proposed Sheet)
- Does not have a blocking scrim (use case for upcoming proposed Sheet)
- Action bar remains visible when the child panel is open or closed (same as existing functionality)
- Mutually exclusive from "detached"
- Animation of width, box shadow applied.
- Animation direction, box shadow direction determined by shell panel slotted location within shell.

### Acceptance Criteria

As proposed, this mode will allow users to use a Shell Panel to overlay other content in the main slotted content area of Shell.

### Relevant Info

More information in Confluence proposal

### Which Component

Shell Panel

### Example Use Case

At a smaller viewport size, it may be desirable to use Shell Panel as an "overlay". This will allow consuming apps to change the property and gracefully hide "shown on larger viewport" shell panels.

### Esri team

Calcite (design)

driskull:
@macandcheese how is this different from the `contentBehind` property on the `shell`?

macandcheese:
Much of this functionality does focus on the "without `contentBehind` set on parent shell" use case as that was not intended to be changed:

- default (current behavior): Slotted Shell Panel's action bar and panel push over content slotted in default Shell slot.
- overlay display (proposed additional behavior) Slotted Shell Panel's action bar pushes over content slotted in default Shell slot. Panel is overlaid absolutely on top with animation / transition. start/end slotted shell panel would display on top of top/bottom if multiple were open (but documented against)
- detached mode used with "Default behavior" in this case creates the "swimlanes" where the shell panel appears floating above a plain bg adjacent to the map. 


So, yep, in some ways these proposed changes do create an alternative to using `contentBehind` while leaving that functionality alone to keep those use cases when used with jsapi view padding. I can whip up a demo if needed to demonstrate better, I think the one on Confluence is outdated if still working.

driskull:
> overlay display (proposed additional behavior) Slotted Shell Panel's action bar pushes over content slotted in default Shell slot. Panel is overlaid absolutely on top with animation / transition. start/end slotted shell panel would display on top of top/bottom if multiple were open (but documented against)

Are you proposing that the action bar is on the side of the center content but the panel is on top of it? These are inside a single component (shell-panel) so I'm not sure how we would achieve that.

[...]

PR #6919, opened 2023-05-04
feat(shell-panel): Add `displayMode` and `heightScale` properties and deprecate `detached` and `detachedHeightScale`

driskull:
**Related Issue:** #6388

## Summary

- Adds the property `displayMode` on the `calcite-shell-panel`.
  - DisplayMode = "dock" | "float" | "overlay"; 
- Deprecates the property `detachedHeightScale` on the `calcite-shell-panel`, use `heightScale` instead.
- Deprecates the property `detached` on the `calcite-shell-panel`, use `displayMode` instead.
- Deprecates the CSS property `--calcite-shell-panel-detached-max-height`, use `--calcite-shell-panel-max-height` instead.
- Moves the resize handle to inside of the content node.
- Test app
- tests

## Future

- It would be nice if `content-behind` on the shell and the `displayMode` on the shell-panel were tied together more in order to prevent unwanted display combinations


src/components/shell-panel/interfaces.d.ts
```diff
@@ -0,0 +1 @@
+export type DisplayMode = "docked" | "detached" | "overlaid";
```
jcfranco, line 1: WDYT about these not being in the past (`dock | detach | overlay`)? Not sure about `detach` though. Maybe `isolate` or `separate`?

src/components/shell-panel/shell-panel.tsx
```diff
@@ -47,9 +49,31 @@ export class ShellPanel implements ConditionalSlotComponent, LocalizedComponent,
...
+  handleDetached(value: boolean): void {
+    if (value) {
+      this.displayMode = "detached";
+    } else if (this.displayMode === "detached") {
+      this.displayMode = "docked";
+    }
+  }
+
+  /**
+   * Specifies the display mode - `"docked"` (full height, displays adjacent to center content), `"detached"` (not full height, content separated detached from action bar, displays on top of center content),
```
jcfranco, line 67: +@geospatialem for editorial consideration

jcfranco (approved):
✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨
✨✨🐚🐚🐚✨✨🐚🐚✨✨✨🐚🐚✨✨🐚✨✨✨🐚✨
✨🐚✨✨✨✨🐚✨✨🐚✨🐚✨✨🐚✨🐚✨✨✨🐚✨
✨🐚✨✨✨✨🐚✨✨🐚✨🐚✨✨🐚✨🐚✨✨✨🐚✨
✨🐚✨✨✨✨🐚✨✨🐚✨🐚✨✨🐚✨🐚✨✨✨✨✨
✨✨🐚🐚🐚✨✨🐚🐚✨✨✨🐚🐚✨✨🐚🐚🐚✨🐚✨
✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨

Should we update the usage files as well? I think we could do an independent pass at all components. cc @geospatialem

src/components/shell-panel/interfaces.d.ts
macandcheese: It seems to match other tenses (`escapeDisabled`, `textEnabled`) - and keeps the existing `detached` nomenclature. I think `isolate` or `separate` is... maybe less clear?

jcfranco: > It seems to match other tenses (escapeDisabled, textEnabled)

Those are prop names and not the values. I don't think we have anything similar for existing modes. I could be wrong.

> and keeps the existing detached nomenclature. I think isolate or separate is... maybe less clear?

Nothing good ol' doc won't fix, although I do hear you on `isolate` and `separate`. Do you think `detach` would still be less clear?

src/components/shell-panel/shell-panel.tsx
geospatialem: nitpick:

> .....content is separated from `calcite-action-bar`, displays....

src/components/shell-panel/interfaces.d.ts
driskull: I agree since they are now values and not properties we should probably have them to not be past tense.

@macandcheese can you come up with the values? I think the only one in question is detach right? What about "float"?

geospatialem (commented):
With the deprecation, would this be a good candidate for a future breaking change? cc @jcfranco

src/components/shell-panel/interfaces.d.ts
jcfranco: 👍🏼  for `float`! 🎪🤡🎈

src/components/shell-panel/shell-panel.scss
```diff
@@ -20,8 +20,23 @@
     flex
     flex-initial
     items-stretch;
-  --calcite-shell-panel-detached-max-height: unset;
   z-index: var(--calcite-shell-panel-z-index, theme("zIndex.default"));
+  --calcite-shell-panel-detached-max-height: unset;
```
driskull, line 24: @macandcheese should we rename these CSS props to not use detached and float instead?

src/components/shell-panel/shell-panel.scss
```diff
@@ -52,15 +67,15 @@
   --calcite-shell-panel-min-width-internal: var(--calcite-shell-panel-min-width, 340px);
 }
 
-:host([detached-height-scale="s"]) .content--detached {
+:host([detached-height-scale="s"]) .content--float {
```
driskull, line 70: Should we update this detached prop to say float as well? deprecate and add new?

driskull: Or could we just deprecate this prop and use a CSS var instead?

src/components/shell-panel/shell-panel.scss
macandcheese: Yeah we should update to the final value name. At some point we can probably look to combine this with the other css vars, maybe when we look at the content-behind / Shell cases.

src/components/shell-panel/shell-panel.scss
macandcheese: I think we could look at changing how this behaves in the future, there’s a need for a “full” and “auto” height. “Float” sounds a bit weird in this context but if that’s the value name we should update to match.

driskull: @jcfranco should we deprecate `detatchedHeightScale` in favor of a new prop?

jcfranco: Good point. I believe we should once we have the new prop in place.

macandcheese: I wonder if we need this at all with the available css variables, especially if it only affects the one mode. 

Detached shell panel in vertical layout could be height: auto to fit content, and could set a max height  of 100% (minus margin). Not sure how common having a defined height that is less than 100% but more than the height of the content is.

cc @asangma for general Shell FYI / use cases

driskull: I added it but if we want to remove we can.

driskull:
Ready for review again.

src/components/shell-panel/shell-panel.scss
driskull: @macandcheese do you want me to remove the `heightScale` prop and just make it so a user can control the height via a CSS variable?

alisonailea (approved):

macandcheese:
@ashetland @SkyeSeitz could you take a look at this Chromatic build? A few expected height changes and new story. Thanks!

Issue #6785, opened 2023-04-14
Need better control over tooltip duration and close behavior


cont-kolomeets:
### Description

The current behavior of the tooltip is not very user friendly.

* The tooltips stays until I stop moving the mouse cursor.
* The tooltip stays visible too long after the mouse cursor moves away.

![Tooltip](https://user-images.githubusercontent.com/5318527/232012235-c50a9b93-83e5-42fc-bcef-7f35bf562b4a.gif)

### Acceptance Criteria

(1) Need to be able to hide the tooltip right after the mouse cursor moves away.
(2) Need to reduce the animation duration.

### Which Component

https://developers.arcgis.com/calcite-design-system/components/tooltip/

### Priority impact

p1 - need for current milestone

### Esri team

ArcGIS Business/Community Analyst

macandcheese:
To confirm, do you want to adjust the animation duration (fade in transition), or the time it takes for the tooltip to invoke (and that animation duration to begin)?

jcfranco:
The issues brought up in the description are duplicates of the following:

1. https://github.com/Esri/calcite-components/issues/6278
2. https://github.com/Esri/calcite-components/issues/6396

ddred11:
Just to add to this, In the Business Analyst web app we have already experienced internal user dissatisfaction with the lag time given to the visibility of the tooltip after the user has moved the cursor away from the hotspot. Would be nice if the component had timing control. Or at the very least reduce the lag time so the tooltip is removed immediately after the cursor leaves the hotspot.

cont-kolomeets:
@macandcheese 

> To confirm, do you want to adjust the animation duration (fade in transition), or the time it takes for the tooltip to invoke (and that animation duration to begin)?

Good question. Would be nice to have control over all of these.

* The time it takes for the tooltip to invoke, so it doesn't appear right away after hover.
* The animation transition duration. We would like to reduce that so that the tooltip appears faster.
* The time it takes to go away. Our goal is to be able to remove the tooltip right away after the cursor leaves.

Please let me know if there is a way to control those in 1.0.7 somehow.

geospatialem:
Re-assigning to address the cursor movement outside of the reference element, where the `tooltip` stays open when moving outside of the reference element and still moving the mouse.

geospatialem:
Verified in `1.4.0-next.9`

[...]

PR #6922, opened 2023-05-04
fix(tooltip): close tooltip when pointer is moving

driskull:
**Related Issue:** #6785

## Summary

- Closes tooltip when pointer is not moving on the tooltip or its reference element.
- Adds test


jcfranco (approved):
💥

Issue #6912, opened 2023-05-03
Calcite tree - allow selection of parent category w/out selecting children


capeoples:
### Description

We would like to have a property on the calcite tree that just allows for selecting the parent, without selecting children. We also want to make sure that this enhancement is accessible. 

_____

Currently in the calcite tree, when a parent is expanded / selected, it automatically selects all of the children. 
<img width="609" alt="Screen Shot 2023-05-03 at 11 40 02 AM" src="https://user-images.githubusercontent.com/30473086/236012618-b50e3da3-d635-440e-be06-6039cf778325.png">

When applied to filters in ArcGIS Online, this creates a scenario that is confusing / overwhelming for users, as there could be many nested filters under a parent. There is also a need to select just the parent in many filter use cases. Current example:
<img width="316" alt="Screen Shot 2023-05-03 at 11 45 52 AM" src="https://user-images.githubusercontent.com/30473086/236013962-12bf786d-ea38-404a-b2c8-aa253dc8f093.png">

### Acceptance Criteria

Allow for a selecting just the parent in a tree without selecting children.

### Relevant Info

[This example from w3.org of a File Directory for a Treeview](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-1a/) illustrates the behaviour we are looking for. Here's a link directly to the [codepen example](https://codepen.io/pen). 

### Which Component

Calcite tree

### Priority impact

p1 - need for current milestone

### Esri team

ArcGIS Online

geospatialem:
@SkyeSeitz @ashetland Could you take a peek at this one? The solution will likely be a new prop on `tree`, which would only select the parent item and not the child(ren). This pattern would differ from what is currently offered when a parent item is expanded and selected.

geospatialem:
Possible related issue: https://github.com/Esri/calcite-components/issues/6509, which is a bug request on `tree-item` related to child `action` and selection - also targeted for the May release.

SkyeSeitz:
Some thinky thoughts and notes on interaction behavior so we can align and discuss any open ended questions :)
Documented at [this file](https://www.figma.com/file/R0B6ljDBIyKPtEa5GXjSjx/Tree-%5Bparent%2Fchildren-selection%5D-%5B6912%5D?node-id=1%3A77844&t=AtfixlXqewQTJzV0-1).

paulcpederson:
[...]

Issue #6509, opened 2023-02-23
Clicking tree item action also selects it


nwhittaker:
### Actual Behavior

Clicking a tree item's slotted `actions-end` action(s) also selects the tree-item.

### Expected Behavior

Clicking a tree item's slotted `actions-end` action(s) does not selects the tree-item.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/poObpdK

### Reproduction Steps

1. Visit sample
2. Click the tree item's trash icon (all the way on the right side of the viewport)
3. The tree item becomes selected

### Reproduction Version

1.0.7

### Impact

On a technical level, in the Field Maps web app, selecting a tree item can be costly since it loads a layer and form. We want to be able to perform actions on other layers (e.g. remove them) without deselecting the current layer.

On a usability level, it can be surprising that the tree-item becomes selected here since the user can only anticipate with certainty that the clicked action will be performed.

One workaround is to manually stop event propagation of `click` and `keydown` events on tree-item actions.

### Esri team

ArcGIS Field Apps

geospatialem:
Verified in `1.4.0-next.10` via mouse and keyboard selection of the slotted `action`, where the `tree-item` selection does not occur.

![verify-tree-slotted-action-unrelated-to-select](https://user-images.githubusercontent.com/5023024/236598412-d6249eab-60da-42e6-8351-2fa68be5e76f.gif)

Issue #6444, opened 2023-02-08
Tree component parents will not open or close when user clicks on parent name


ddred11:
### Actual Behavior
Tree component only opens and close when user clicks on carat to the left of a parent's name in selection mode: single

Additionally, there are similar hit area issues across the other modes.

This is a usability issue in that the click target is too small and will cause frustration.

### Expected Behavior
User will expect to be able to click on parents' names in addition to carat to open and close content. This is especially the case since the component currently shows a cursor change (suggesting clickability) during hover over the parent.

### Reproduction Sample
https://developers.arcgis.com/calcite-design-system/components/tree/

### Reproduction Steps
This can be reproduced directly from the Developers site
https://developers.arcgis.com/calcite-design-system/components/tree/

Click on Mid West or Westcoast and nothing happens

Note that cursor changes when hovering over these parents leading user to believe they are clickable

### Reproduction Version
1.04

### Relevant Info
All

### Regression?
No response

### Impact
Currently we have had to add wrappers to the tree instances in BA WEB app after adoption to remedy this behavior.

In the BA widget for Experience Builder, there is not time for a solution and usability for this release will be affected.

### Esri team
ArcGIS Business/Community Analyst

macandcheese:
Our visual treatments are pretty inconsistent across modes… @ashetland @SkyeSeitz we should look to improve hit area distinction between expand / collapse and selection state where applicable across these.

ftiru:
Currently, the calcite tree does not allow clicking on the body (in red) to expand/collapse the tree. You have to specifically click on the chevron icon (in green). This is not a good experience and we wanted to know when this can be fixed/implemented.

![image](https://user-images.githubusercontent.com/16827336/225700125-5981ba3b-434c-4802-ad24-02ad47bc98ef.png)

subgan82:
We would benefit on this improvement as well for analysis-spatial-reference component , where we do this on our side today to allow expand/collapse on the tree item name

http://arcgiswebanalysis.esri.com:7080/analysis-components/storybook/?path=/story/components-analysis-spatial-reference-input--basic

geospatialem:
Reallocating to the May release for design considerations.

geospatialem:
Verified in `1.4.0-next.10`

![verify-tree-parent-click-open-close](https://user-images.githubusercontent.com/5023024/236598235-5492e124-52f2-48f0-a01e-ac9c82067d00.gif)

PR #6926, opened 2023-05-05
fix(tree): allow selection of parent category w/out selecting children

paulcpederson:
**Related Issue:** #6912, #6444, #6509

## Summary

Updates the interaction model of trees to adhere to the [new design](https://www.figma.com/file/R0B6ljDBIyKPtEa5GXjSjx/Tree-%5Bparent%2Fchildren-selection%5D-%5B6912%5D?node-id=1-76826&t=GOElQQiBF6PkKrgN-0). The key changes to the interaction model are:

1. Tree items will expand when they are clicked (#6444)
2. Child selection modes will not automatically display all children of a selected item as selected (#6912)
3. Clicking a slotted action inside a tree item will not select the tree item (#6509)

Note: this pr does not address the hover states in that design. We are coming down to the wire and I want to give time for code review. I can help address the styling/focus/hover/accessibility aspects of the tree next release.

### DIFF
--- a/src/components/tree-item/tree-item.tsx
+++ b/src/components/tree-item/tree-item.tsx
@@ -361,7 +361,8 @@ export class TreeItem
 
   @Listen("click")
   onClick(event: Event): void {
-    if (this.disabled) {
+    const isActionClick = (event.target as HTMLCalciteActionElement).slot === "actions-end";
+    if (this.disabled || isActionClick) {
       return;
     }
 
--- a/src/components/tree-item/tree-item.tsx
+++ b/src/components/tree-item/tree-item.tsx
@@ -365,9 +365,7 @@ export class TreeItem
 
   @Listen("click")
   onClick(event: Event): void {
-    const composedPath = event.composedPath();
-    const isActionClick = composedPath.includes(this.actionSlotWrapper);
-    if (this.disabled || isActionClick) {
+    if (this.disabled || this.isActionEndEvent(event)) {
       return;
     }
 
@@ -394,6 +392,10 @@ export class TreeItem
   @Listen("keydown")
   keyDownHandler(event: KeyboardEvent): void {
     let root;
+
+    if (this.isActionEndEvent(event)) {
+      return;
+    }
 
     switch (event.key) {
       case " ":
@@ -499,6 +501,11 @@ export class TreeItem
   //  Private Methods
   //
   //--------------------------------------------------------------------------
+
+  private isActionEndEvent(event: Event): boolean {
+    const composedPath = event.composedPath();
+    return composedPath.includes(this.actionSlotWrapper);
+  }
 
   private updateParentIsExpanded = (el: HTMLCalciteTreeItemElement, expanded: boolean): void => {
     const items = getSlotted<HTMLCalciteTreeItemElement>(el, SLOTS.children, {
--- a/src/components/tree-item/tree-item.tsx
+++ b/src/components/tree-item/tree-item.tsx
@@ -326,7 +326,11 @@ export class TreeItem
               {this.iconStart ? iconStartEl : null}
               {checkbox ? checkbox : defaultSlotNode}
             </div>
-            <div class={CSS.actionsEnd} hidden={!hasEndActions}>
+            <div
+              class={CSS.actionsEnd}
+              hidden={!hasEndActions}
+              ref={(el) => (this.actionSlotWrapper = el as HTMLElement)}
+            >
               {slotNode}
             </div>
           </div>
@@ -361,7 +365,8 @@ export class TreeItem
 
   @Listen("click")
   onClick(event: Event): void {
-    const isActionClick = (event.target as HTMLCalciteActionElement).slot === "actions-end";
+    const composedPath = event.composedPath();
+    const isActionClick = composedPath.includes(this.actionSlotWrapper);
     if (this.disabled || isActionClick) {
       return;
     }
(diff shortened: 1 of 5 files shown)


src/components/tree-item/tree-item.tsx
```diff
@@ -361,7 +361,8 @@ export class TreeItem
 
   @Listen("click")
   onClick(event: Event): void {
-    if (this.disabled) {
+    const isActionClick = (event.target as HTMLCalciteActionElement).slot === "actions-end";
```
driskull, line 364: I think this is a fine solution but it would probably be better if the `onClick` was just removed from the `host` and set only onto the internal nodes that should allow the `onClick`.

Otherwise, as we add new slots, we would have to start filtering them out here. 

What do you think?

driskull: If we do go this route, it might be better to use composedPath and check if the slot is in it.

driskull (commented):
Looks good to me! Just had some feedback on filtering out specific elements from the onclick handler.

src/components/tree-item/tree-item.tsx
paulcpederson: Yeah I think `@Listen("click"` is weird and kind of an anti-pattern. I can try to move this to be more targeted.

paulcpederson: I just looked at this a bit and it would be pretty substantial to rework this. It may also have accessibility impact. Let me try to use composedPath though, that way this will work in more cases.

paulcpederson: Updated. This will catch all clicks anywhere in the slotted dom and prevent selections.

src/components/tree-item/tree-item.tsx
```diff
@@ -361,7 +365,9 @@ export class TreeItem
 
   @Listen("click")
   onClick(event: Event): void {
-    if (this.disabled) {
+    const composedPath = event.composedPath();
+    const isActionClick = composedPath.includes(this.actionSlotWrapper);
+    if (this.disabled || isActionClick) {
```
geospatialem, line 370: Would we need to add in support for keyboard users here for consistency? cc @driskull 

### keyboard
keyboard-tree

### mouse
mouse-tree

paulcpederson: oh sure I can handle that case

driskull: A11y wise, if there are slotted actions the role may need to change from tree to treegrid. Not sure though.

paulcpederson: Yeah there are some other accessibility updates that would be good to make on the tree component as a whole. I could probably try to dig into that next release as it's part of our roadmap for Online

paulcpederson:
Running the e2e again as they somehow started failing? Is this normal?

driskull (approved):
👍 ✨

paulcpederson:
@driskull or @jcfranco  do you have any ideas on the e2e tests? Seeing it locally too. I went back to the commit that passed and it also failed, so wondering what could have changed?

```
Error: Protocol error (Runtime.callFunctionOn): Session closed. Most likely the page has been closed.
```

src/components/tree/tree.tsx
```diff
@@ -165,23 +165,26 @@ export class Tree {
         (target.hasChildren &&
           (this.selectionMode === "children" || this.selectionMode === "multichildren")));
 
+    // deselecting a parent in multichildren should deselect all the children
```
jcfranco, line 168: I think this comment can be dropped as the line below is expressive enough.

jcfranco (approved):
↖↖↖↖↖↖↖↖↖↖↖↖↖↖↖↖↖↖↖↖↖↖↖
↖🌳↖↖↖🌳↖🌳🌳🌳↖↖🌳🌳🌳↖🌳🌳🌳🌳↖🌳↖
↖🌳🌳↖↖🌳↖↖🌳↖↖🌳↖↖↖↖🌳↖↖↖↖🌳↖
↖🌳↖🌳↖🌳↖↖🌳↖↖🌳↖↖↖↖🌳🌳🌳↖↖🌳↖
↖🌳↖↖🌳🌳↖↖🌳↖↖🌳↖↖↖↖🌳↖↖↖↖↖↖
↖🌳↖↖↖🌳↖🌳🌳🌳↖↖🌳🌳🌳↖🌳🌳🌳🌳↖🌳↖
↖↖↖↖↖↖↖↖↖↖↖↖↖↖↖↖↖↖↖↖↖↖↖


I'll take a look at the unstable test. 👀

jcfranco:
> I'll take a look at the unstable test. 👀

Talked to @driskull and I think found a way to stabilize the tooltip tests. Feel free to skip them (https://github.com/Esri/calcite-components/blob/master/src/components/tooltip/tooltip.e2e.ts#L730 and https://github.com/Esri/calcite-components/blob/master/src/components/tooltip/tooltip.e2e.ts#L791) to get this installed and I'll submit a separate PR for restoring/stabilizing the skipped tests.

paulcpederson:
I skipped the tests and addressed the pr comment, will see how the e2e tests go

jcfranco:
Thanks again, @paulcpederson! Can you open a follow-up issue for the pending styling/focus/hover/accessibility updates?

geospatialem:
Added a follow-up issue targeted for August regarding the hover states in #6930. 

@paulcpederson were there other considerations with styling, focus or a11y for another follow-up issue? Didn't see other context in Figma or the linked issues.

paulcpederson:
Nope, the styling was left as is for now. I think your hover states issue should cover updating that. Accessibility has it's own issue IIRC. Nothing more to open for this, thanks for opening that issue. And thanks @jcfranco for the merge

PR #6927, opened 2023-05-05
chore: find/replace master to main

benelan:
**Related Issue:** #

## Summary
Search/replaced master to main in preparation of setting `main` as the default branch


jcfranco (approved):
Awesome! 

Let's sync up before installing this.

PR #6935, opened 2023-05-08
ci: add label to skip visual snapshots

benelan:
**Related Issue:** #

## Summary
Adds a `skip visual snapshots` label which can be used for PRs that don't need to run chromatic.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -4,7 +4,7 @@ on:
     branches: [master]
   pull_request:
     branches: [master]
-    types: [labeled]
+    types: [labeled, synchronize]
 jobs:
   run:
     if: |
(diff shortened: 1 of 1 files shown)


macandcheese:
What is preferred - to opt in with `pr ready for visual snapshots` or opt out with this new label?

benelan:
the new label will skip the need to run visual snapshots. When a PR has that label, the PR check will automatically show up as successful every time you push. 

The `pr ready for visual snapshots` label is used to run the snapshots, and it only does so during the act of adding it. So if a PR has the ready label and you push and need to rerun the snapshots, you need to toggle the label to get it to run again. This is to prevent unnecessary chromatic runs for cleanup commits and other stuff because we are charged per snapshot.

Without using either label, the PR check will be in the yellow "Expected" state, which blocks merging since it is a required check.

alisonailea (approved):

PR #6938, opened 2023-05-08
chore(menu-item): adds base message bundles.

anveshmekala:
**Related Issue:** #6531 

## Summary
This will add base message bundles to enable translations build process.

### DIFF
--- a/src/components/menu-item/assets/menu-item/t9n/messages.json
+++ b/src/components/menu-item/assets/menu-item/t9n/messages.json
@@ -1,0 +1,4 @@
+{
+  "submenu": "submenu",
+  "open": "open"
+}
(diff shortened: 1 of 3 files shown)


src/components/menu-item/assets/menu-item/t9n/messages.json
```diff
@@ -0,0 +1,4 @@
+{
+  "submenu": "submenu",
```
jcfranco, line 2: Should these be capitalized (like other message bundles)?

jcfranco (approved):

src/components/menu-item/assets/menu-item/t9n/messages.json
anveshmekala: Good Catch, the value should be capitalized.

PR #6957, opened 2023-05-11
chore: drop nomodule script usage

jcfranco:
**Related Issue:** N/A

## Summary

`nomodule` fallback scripts are no longer needed as we don't support ES5 anymore.


driskull (approved):
Awesome 👏

PR #6958, opened 2023-05-11
fix(dropdown): button `width=full` is enclosed within containing block 

Elijbet:
**Related Issue:** #5903

## Summary
The `dropdown trigger button` is enclosed in its containing block if the button text has no dashes or spaces when the button is set to `width="full"`.

**Note:** This should be scheduled for a future planned breaking change release. 

## Relevant Info
https://github.com/Esri/calcite-components/issues/5903#issuecomment-1555399923


src/components/dropdown/dropdown.e2e.ts
```diff
@@ -22,7 +22,7 @@ describe("calcite-dropdown", () => {
   });
 
   describe("renders", () => {
-    renders(simpleDropdownHTML, { display: "inline-flex" });
+    renders(simpleDropdownHTML, { display: "block" });
```
driskull, line 25: @jcfranco would a change to the default display be considered a breaking change? If not now, then in the future?

driskull (commented):
The code changes look good but wouldn't changing from `inline-flex` to `block` be a breaking change for someone who expects the dropdown to not span the full width by default?

src/components/dropdown/dropdown.e2e.ts
driskull: I'm kind of surprised the screenshot tests don't show any changes when going from `inline-flex` to `block`

Elijbet: Well, the screenshot below that has 2 dropdowns next to each other is adjusted for the containing block to apply flex, because when the dropdown is a `block` is will otherwise stack.

driskull (approved):
I'm ok with this change if @jcfranco is. 👍  We may want to note it in the release docs.

Ideally, any changes to host styles should be noted for users. Whether we consider them breaking changes or not is debatable 🧑‍⚖️

src/components/dropdown/dropdown.e2e.ts
```diff
@@ -811,22 +811,26 @@ describe("calcite-dropdown", () => {
...
-      <calcite-dropdown id="dropdown-2">
-        <calcite-button id="trigger" slot="trigger">Open dropdown</calcite-button>
-        <calcite-dropdown-group id="group-1" selection-mode="single">
-          <calcite-dropdown-item id="item-1"> Dropdown Item Content </calcite-dropdown-item>
-          <calcite-dropdown-item id="item-2" selected> Dropdown Item Content </calcite-dropdown-item>
-          <calcite-dropdown-item id="item-3"> Dropdown Item Content </calcite-dropdown-item>
-        </calcite-dropdown-group>
-      </calcite-dropdown>`);
+    await page.setContent(html`
+      <div style="display: flex">
```
jcfranco, line 815: I'll roll back this change as it is not critical for this test.

macandcheese:
Would adopting Floating UI / `reference-element` pattern for Dropdown solve for this? That way we won't need to handle any cases like this where the slotted `trigger` element isn't behaving, it'll be completely outside the Dropdown component.

jcfranco:
@macandcheese Great idea. That'd be way cleaner and we could include it in `rc`. I'll take a stab and see how it affects other components.

jcfranco:
> Would adopting Floating UI / reference-element pattern for Dropdown solve for this?

@macandcheese I started looking into this and the complexity of this approach is way higher than I expected:

* it requires logic from tooltip/popover managers, so we would need to refactor to share the code between reference-element components
* I looked into other design system components, but I couldn't find one that uses a trigger defined outside of the dropdown
  * very IMO, but unlike tooltip/popover, something about having the trigger outside for a dropdown strikes me odd. Maybe it's because my mental model of a dropdown includes both trigger and content.
* this would also need split-button to be updated to use the same pattern

jcfranco:
Screenshot testing shows that this will affect layout in the opposite direction and would also require changing the dropdown's `display` to have it work as before. Compare https://codepen.io/jcfranco/pen/JjxvbOJ with https://codepen.io/jcfranco/pen/KKJRNYV. 

I've submitted https://github.com/Esri/calcite-design-system/pull/8253 as an alternative solution that will make it easier to enable truncation without affecting layout.

Closing based on ☝️.

Issue #6464, opened 2023-02-14
Inconsistent padding around calcite-tip elements in different browsers


timmorey:
### Actual Behavior

When a `<calcite-tip>` is used inside a `<calite-tip-manager>`, the padding/margin around the tip is rendered differently in safari and firefox than it is in chrome.

This is reproducible in the "whats new" tips used in the Field Maps Designer web app.  It is also observable in the calcite-components documentation

Safari example from documentation: 
<img width="443" alt="Screenshot 2023-02-14 at 10 21 14 AM" src="https://user-images.githubusercontent.com/2340894/218780950-c3a9325c-0f0a-4b55-93b5-26b4baecb73f.png">

Chrome example from documentation:
<img width="395" alt="Screenshot 2023-02-14 at 10 21 48 AM" src="https://user-images.githubusercontent.com/2340894/218781017-a2dad6a5-5227-4b5d-952a-ebed218a8613.png">

This appears to be related to the use of the `:host-context` pseudo-class, which safari/firefox don't support.

### Expected Behavior

Padding around `calcite-tip` elements should be consistent across browsers.

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/tip-manager/

### Reproduction Steps

1. Scroll down to the sample in the doc
2. Compare padding around tip in various browsers

### Reproduction Version

1.0.5

### Relevant Info

Tested using
- safari Version 16.3 (18614.4.6.1.5)
- firefox 109.0.1 (64-bit)
- chrome Version 110.0.5481.77 (Official Build) (arm64)

### Impact

Not blocking progress - its just an inconsistency we'd like to fix.

### Esri team

ArcGIS Field Apps

geospatialem:
Verified in Chrome, FF, and Edge on Windows. 

@ashetland did you log the Figma changes? Want to make sure we're good to go prior to closing the work out.

PR #6959, opened 2023-05-11
fix(tip-manager): Set padding for tips and tip-groups consistently

driskull:
**Related Issue:** #6464

## Summary

- Removes use of `:host-context` CSS selector. https://developer.mozilla.org/en-US/docs/Web/CSS/:host-context
  -  It's not supported in FF or Safari. @jcfranco can be ban its usage?
- Sets padding inside the tip manager container instead of on slotted elements for consistency.
- Sets tip max-width on tips within the `calcite-tip-manager` and within `calcite-tip-group`'s
- Adds screenshot tests
- Fixes link to tooltip manager html file in index of demos


src/components/tip-group/tip-group.scss
```diff
@@ -8,12 +8,5 @@
 
 ::slotted(calcite-tip) {
   @apply m-0 border-none;
-}
-
-:host-context(calcite-tip-manager) {
-  @apply my-3;
-}
-
-:host-context(calcite-tip-manager) ::slotted(calcite-tip) {
```
driskull, line 17: Can we setup a lint error/warning for using host-context so its not checked into any files?

src/components/tip-manager/tip-manager.scss
```diff
@@ -61,7 +61,8 @@
     flex
     items-start
     justify-center
-    overflow-auto;
+    overflow-auto
+    p-4;
```
driskull, line 65: Used `p-4` to line up with header

src/components/tip-group/tip-group.scss
jcfranco: Good idea. I'll see what we can set up for this.

jcfranco (approved):
💥

driskull:
@ashetland @macandcheese can you review the screenshot changes for this PR?

PR #6964, opened 2023-05-12
fix(panel): Make sure the header doesn't appear over other components

driskull:
**Related Issue:** None

## Summary

- Makes sure z-index on header doesn't leak and cover items outside of the panel.

Example issue:

image


driskull:
@jcfranco it may be nice if we had a class for this we could apply to all applicable containers within components shadowRoot.

jcfranco:
Unfortunately, this goes against the z-index layering we implemented (see https://github.com/Esri/calcite-components/issues/5514#issuecomment-1282917677 and https://github.com/Esri/calcite-components/issues/5514#issuecomment-1282967568). For these cases, developers should apply the same z-index layers or tweak ours (coming soon via https://github.com/Esri/calcite-components/issues/6558).

driskull:
Ah I forgot about that.

Is that because sometimes the users want the scrolling content to be higher than the header and so we give them the ability to do that?

It just seems odd for us to let the zindex of an internal panel header leak outside of the component and mess with other UI.

jcfranco:
Were you able to fix this by tweaking z-index on the popup side?

driskull:
> Were you able to fix this by tweaking z-index on the popup side?

In the JSAPI, I basically added this:

```css
calcite-flow {
  position:relative;
  z-index:1;
}
```

Issue #6659, opened 2023-03-24
Support multiple focus-traps on an app/page


driskull:
### Description

Support multiple focus-traps on an app/page

### Acceptance Criteria

Apps can share the focus trap stack so that they don't clash with each other in different apps.

### Relevant Info

https://devtopia.esri.com/WebGIS/arcgis-js-api/issues/50190#issuecomment-4017341

### Which Component

All focus-trap components

### Example Use Case

This sample loads a webmap that will prompt for authentication.
https://codepen.io/ellupp/pen/QWVbPPb?editors=1000

Open the sample, and remove "focus-trap-disabled" from the calcite-modal. The UI becomes unresponsive and you see the exceptions in the console. warning- this entirely froze codepen for me even after killing the tab and opening a new one, I needed to restart chrome.

### Priority impact

p4 - not time sensitive

### Esri team

N/A

geospatialem:
Workaround is to disable the focus trap in the component.

driskull:
I think we should elevate this issue. The workaround is to basically disable an accessibility feature.

geospatialem:
> I think we should elevate this issue. The workaround is to basically disable an accessibility feature.

@jcfranco WDYT about adding this in to the current (May) milestone?

jcfranco:
@driskull June is a patch-only release, though. If July is too late, let's sync up tomorrow to hash out a plan.

driskull:
July is probably fine then. Lets just get it on the board.

driskull:
https://stackoverflow.com/a/70184233

geospatialem:
Verified in `master` where the `FocusTrap` is imported into `config.ts` and into the `focusTrapComponent` utility.

PR #6965, opened 2023-05-16
feat: support focus traps outside of components

driskull:
**Related Issue:** #6659

## Summary

- Adds `focusTrapStack` property to all components supporting focusTraps.
- This makes it possible to share the same focus trap stack outside of calcite-components.
- A user would create their own `trapStack` array and pass it to all the components they need to use that have the `focusTrapStack` property.
- In the future, maybe we can expose another way to set it just one time and not per component.

### DIFF
--- a/src/components/input-date-picker/input-date-picker.tsx
+++ b/src/components/input-date-picker/input-date-picker.tsx
@@ -130,6 +130,13 @@ export class InputDatePicker
   }
 
   /**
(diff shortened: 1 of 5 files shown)


jcfranco:
Thanks for putting in the time and effort on this, but I'd rather we focus on the global solution vs passing the trap stack per component.

The associated issue doesn't have a milestone associated, so let's sync up on this once the May release is wrapped up.

Issue #4738, opened 2022-06-15
Combobox blank after selecting the same option twice


paulcpederson:
### Actual Behavior

When a combobox option is selected in single select mode, selecting it again does not show the option as selected. It stays blank with the text input visible. 

### Expected Behavior

Selecting an option should show that option as selected. This will keep the combobox consistent with calcite select and consistent internally with itself as after selecting a new option.

### Reproduction Sample

https://codepen.io/paulcp/pen/abqXgqE

### Reproduction Steps

1. Open the combobox
2. Select Altitude (already selected)

### Reproduction Version

1.0.0-beta.83

### Relevant Info

Looking at different versions, it seems like this has behaved very differently in different versions. 

- beta.83 the combobox will close, the input remains focused and active, and shows as blank
- beta.81 will close the options, hide the input, but show as blank
- beta.69 and earlier worked as expected

### Regression?

1.0.0-beta.69

macandcheese:
@paulcpederson would one of the suggestions for an additional "mode" or type of Combobox selection work here? Both UX could be desirable I think... https://github.com/Esri/calcite-components/issues/3074

paulcpederson:
This issue is exclusively tied to single select to me. In multi select I think the behavior is fine, but if I'm using single select and I select an item, I expect it to behave like a select menu. Selecting one item vs another should not have different behavior, IMO.

macandcheese:
We could look at the behavior of MUI as an example - they by default have a clear affordance that is opt-out with `disableClear` - this could be combined with fixing the regression to solve all the requests in - https://github.com/Esri/calcite-components/issues/3074

geospatialem:
For the dev taking this on - the bug is still present in `1.2.0`, and when a value is already selected, should remain in the `combobox`, even if its selected again. Showing current behavior, and expected (from [MUI](https://mui.com/material-ui/react-autocomplete/#combo-box)).

It also looks like the `value` prop isn't working, perhaps related to a fix?

Updated Codepen: https://codepen.io/geospatialem/pen/QWZLPyQ

## Current behavior (still in `1.2.0`)
![repro](https://user-images.githubusercontent.com/5023024/230116979-86fe7748-4d44-4017-b324-15dbce291de5.gif)

## Expected behavior (from [MUI](https://mui.com/material-ui/react-autocomplete/#combo-box))
[...]

PR #6972, opened 2023-05-17
feat(combobox): make combobox clearable

jcfranco:
**Related Issue:** #4738 

## Summary

This updates the combobox to display a clear-value button by default and adds a `clearDisabled` property to disable the clear-button. Additionally, this prevents single-selection from deselecting a value when clicked again making it similar to `calcite-select`.

## Noteworthy changes

* Includes a functional component + style mixin to create an internal close or x button used in other places. This can be used as the base for an internal x-button component if needed.
* Drops the `coerce` param from `combobox-item#toggleSelected` as it's internal and isn't used.


src/components/combobox/combobox.tsx
```diff
@@ -688,11 +707,20 @@ export class Combobox
   };
 
   clickHandler = (event: MouseEvent): void => {
-    if (event.composedPath().some((node: HTMLElement) => node.tagName === "CALCITE-CHIP")) {
+    const composedPath = event.composedPath();
+
+    if (composedPath.some((node: HTMLElement) => node.tagName === "CALCITE-CHIP")) {
       this.open = false;
-      event.stopPropagation();
+      event.preventDefault();
```
jcfranco, line 714: @macandcheese Tests seem to pass, but do you recall why the original event needed to stop propagating?

src/components/combobox/combobox.tsx
```diff
@@ -506,6 +512,19 @@ export class Combobox
   //
   // --------------------------------------------------------------------------
 
+  private clearValue(): void {
+    this.ignoreSelectedEventsFlag = true;
```
jcfranco, line 516: @driskull @macandcheese Do you have a suggestions on how we could generalize this? There's a few places where bits and pieces are reused, but I haven't found a pattern. 😵

driskull: Maybe `combobox-item` needs an (undocumented?) method that could be called where `calciteComboboxItemChange` isn't fired.

I agree that these flags are super hacky looking and prone to errors. We should try to come up with a pattern where we're more deliberate with what we're trying to accomplish.

src/components/functional/XButton.tsx
```diff
@@ -0,0 +1,39 @@
+import { FunctionalComponent, h } from "@stencil/core";
+import { JSXAttributes, JSXBase } from "@stencil/core/internal";
+import { Scale } from "../interfaces";
+
+export interface XButtonOptions extends JSXAttributes {
```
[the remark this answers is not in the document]
driskull: maybe it should be `ClearButton`/`ResetButton` since the `x` icon is used for close as well?

image

src/components/combobox/combobox.tsx
macandcheese: It allowed focus to be set on the clicked Chip's close button, when user clicked the Chip, which was a change to support how Chips behaved post some Chip Group related changes.  

Although looking at it in action (and compared to past behavior), that probably wasn't needed, so this should be fine to remove.

src/components/functional/XButton.tsx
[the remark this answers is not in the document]
jcfranco: Bounced names w/ @ashetland and we landed on x-button as it could be used for both clearing and closing. Is there another name that could represent clear/reset/close? Luckily, it's internal for now, so we can rename if we land on a better name.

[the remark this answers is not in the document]
driskull: sounds good!

src/components/combobox/combobox.tsx
jcfranco: I don't think they're hacky, tbh. Sometimes we need to update props internally. We should definitely spend some time to develop a pattern for components to adopt. It's a bit sporadic at the moment.

src/assets/styles/includes.scss
```diff
@@ -74,3 +74,61 @@
...
+    self-center
+    focus-base
+    transition-default
+    text-color-3
+    m-0
+    border-2
+    cursor-pointer
+    items-center
+    bg-transparent;
+    -webkit-appearance: none;
```
driskull, line 111: do we need `appearance: none` as well?

src/assets/styles/includes.scss
```diff
@@ -74,3 +74,61 @@
...
+    display: flex;
```
driskull, line 112: flex is already applied using `apply`.

driskull: I think some other css statements are duplicated as well.

driskull (approved):
Fantastic ✨

src/assets/styles/includes.scss
jcfranco: Yikes! Can't believe I missed that. 😅 

Thanks for the great catches. Fixing right up!

src/components/combobox-item/combobox-item.tsx
```diff
@@ -127,11 +127,14 @@ export class ComboboxItem implements ConditionalSlotComponent, InteractiveCompon
   //
   // --------------------------------------------------------------------------
 
-  toggleSelected(coerce?: boolean): Promise<void> {
-    if (this.disabled) {
+  toggleSelected(): Promise<void> {
+    const isSingleSelect = getElementProp(this.el, "selection-mode", "multiple") === "single";
```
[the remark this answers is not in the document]
jcfranco: Yeah, I remember. I'll submit a PR to rename it today, so it's not misleading when reading code.

src/components/functional/XButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+  scale
+}: XButtonOptions): FunctionalComponent {
+  return (
+    <button
+      aria-label={label}
+      class={CSS.button}
+      disabled={disabled}
+      key={key}
+      onClick={onClick}
+      tabIndex={-1}
```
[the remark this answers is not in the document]
jcfranco: Great point. I think I had it this way because of how it works in `input`, but `Escape` doesn't clear the value here. @ashetland WDYT?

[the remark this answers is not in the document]
jcfranco: Synced up w/ @ashetland. Following `input`'s example, the clear button won't be focusable and `Escape` will clear the value if the combobox is closed and focused.

src/assets/styles/includes.scss
jcfranco: Fixed!

src/components/combobox/combobox.tsx
```diff
@@ -628,6 +628,10 @@ export class Combobox
         }
         break;
       case "Escape":
+        if (!this.open) {
```
driskull, line 631: should `this.clearDisabled` be checked here?

src/components/combobox-item/combobox-item.tsx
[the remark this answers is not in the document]
driskull: related to https://github.com/Esri/calcite-components/issues/6038

src/components/combobox/combobox.tsx
jcfranco: Yes, it should. Sorry about that. 😞

jcfranco: FIxed!

PR #6973, opened 2023-05-17
fix(combobox): ensure most recent selected item is active when combobox is opened

jcfranco:
**Related Issue:** N/A

## Summary

This should help users recall the most recent selection when opening the combobox. Previously, this was not ideal if there is a large list of items since the active item index would start from the top when opened. This is especially noticeable in single-selection mode where the input showing the selected value is cleared on open.

Stems from https://github.com/Esri/calcite-components/pull/6947.


driskull (approved):
👏 👍

PR #6975, opened 2023-05-17
refactor(navigation): allow nesting secondary and tertiary navigation slots.

macandcheese:
## Summary
cc @anveshmekala

### DIFF
--- a/src/components/navigation/navigation.scss
+++ b/src/components/navigation/navigation.scss
@@ -4,43 +4,32 @@
  * These properties can be overridden using the component's tag as selector.
  *
  * @prop --calcite-navigation-width: Specifies the width of the component's content area.
- * @prop --calcite-navigation-primary-background: Specifies the background color of the primary navigation.
- * @prop --calcite-navigation-primary-border-color: Specifies the background color of the primary navigation.
- * @prop --calcite-navigation-secondary-background: Specifies the background color of the secondary navigation.
- * @prop --calcite-navigation-secondary-border-color: Specifies the background color of the primary navigation.
- * @prop --calcite-navigation-tertiary-background: Specifies the width of the component.
- * @prop --calcite-navigation-tertiary-border-color: Specifies the background color of the primary navigation.
+ * @prop --calcite-navigation-background: Specifies the background color of the component.
+ * @prop --calcite-navigation-border-color: Specifies the border color of the component.
  */
 
 .container {
   @apply flex 
   flex-col 
   mx-auto 
-  w-full;
+  w-full
+  h-16;
   margin-block: 0;
   margin-inline: auto;
   border-block-end: 1px solid;
-  background-color: var(--calcite-navigation-primary-background, var(--calcite-ui-foreground-1));
-  border-block-end-color: var(--calcite-navigation-primary-border-color, var(--calcite-ui-border-3));
+  background-color: var(--calcite-navigation-background, var(--calcite-ui-foreground-1));
+  border-block-end-color: var(--calcite-navigation-border-color, var(--calcite-ui-border-3));
   &.hide {
     display: none;
   }
 }
 
-.primary {
-  @apply h-16;
-}
-
 .secondary {
   @apply h-12;
-  background-color: var(--calcite-navigation-secondary-background, var(--calcite-ui-foreground-1));
-  border-block-end-color: var(--calcite-navigation-secondary-border-color, var(--calcite-ui-border-3));
 }
 
 .tertiary {
   @apply h-12;
-  background-color: var(--calcite-navigation-tertiary-background, var(--calcite-ui-foreground-1));
-  border-block-end-color: var(--calcite-navigation-tertiary-border-color, var(--calcite-ui-border-3));
 }
 
 .container-content {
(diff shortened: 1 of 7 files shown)


src/components/navigation/navigation.tsx
```diff
@@ -152,62 +146,37 @@ export class CalciteNavigation {
...
-        {this.renderNavLevel("secondary")}
-        {this.renderNavLevel("tertiary")}
+        <div
+          class={{
+            [CSS.container]: true,
+            [SLOTS.navSecondary]: this.el.slot === SLOTS.navSecondary,
+            [SLOTS.navTertiary]: this.el.slot === SLOTS.navTertiary
+          }}
+        >
+          {isPrimaryLevel && <slot name={SLOTS.progress} />}
```
driskull, line 161: With all this `isPrimaryLevel` logic that doens't apply to secondary/tertiary, I'm thinking maybe "primary" is a separate component that uses the "secondary/tertiary" component inside it? 

Only the "primary" component should have the slots for "secondary" and "tertiary".

It would also clean up all this logic depending on the slot name.

Something like:

```html
<calcite-navigation>
  <calcite-subnavigation slot="secondary"></calcite-subnavigation>
  <calcite-subnavigation slot="tertiary"></calcite-subnavigation>
</calcite-navigation>
```

Internally to "calcite-navigation" it could use the "calcite-subnavigation" component to reduce rendering logic as well.

macandcheese: I'd prefer to avoid a second component, as I think that complicates the story quite a bit - this is already pretty simplified compared to the original attempt with a single component and all the slots. Open to better ways to manage the slotted position though, maybe an internal prop / state vs. checking in render.

driskull: Yeah I talked to @anveshmekala about a `@State` which makes sense. 

Something like...

```tsx
@State() activeLevel: "secondary" | "primary" | "tertiary";

// onslotchange do this
this.activeLevel = "secondary"// or whatever it is based on this.el.slot

// Then in render you can just use
const isPrimaryLevel = this.activeLevel === "primary";
```

driskull: The only weird part is that some slots would be conditional.

Instead of conditionally rendering the slots, we should hide the slots. Otherwise, the `onslotchange` event won't ever be fired if the slot isn't first rendered.

src/components/navigation/navigation.tsx
```diff
@@ -23,17 +22,15 @@ type Level = "primary" | "secondary" | "tertiary";
 })
 
 /**
- * @slot logo - A slot for adding a `calcite-logo` component to the primary nav level.
- * @slot user - A slot for adding a `calcite-user` component to the primary nav level.
- * @slot progress - A slot for adding a `calcite-progress` component to the primary nav level.
- * @slot nav-action - A slot for adding a `calcite-action` component to the primary nav level.
- * @slot primary-content-start - A slot for adding a `calcite-menu`, `calcite-action`, or other interactive elements in the start position of the primary nav level.
```
alisonailea, line 30: Changing slot names seems like a breaking change to me. Can we continue to support prior slots here until the next major version?

macandcheese: This is just a PR into the other open PR, so hasn’t made it into the codebase yet :)

src/components/navigation/navigation.scss
```diff
@@ -24,6 +24,19 @@
   }
 }
 
+.user,
+.logo {
+  @apply flex;
+}
+
+.hide {
```
driskull, line 32: you could also have done the following instead...

```css
[hidden] {
@apply hidden;
}
```

src/components/navigation/navigation.tsx
```diff
@@ -124,12 +138,24 @@ export class CalciteNavigation {
...
+  };
+
+  private handleProgressSlotChange = (event: Event): void => {
+    if (this.isPrimaryLevel(this.el)) {
+      this.progressSlotHasElement = slotChangeHasAssignedElement(event);
     }
   };
 
+  private isPrimaryLevel(el: HTMLCalciteNavigationElement): boolean {
+    return el.slot !== SLOTS.navSecondary && el.slot !== SLOTS.navSecondary;
```
driskull, line 156: Instead of passing in `this.el` into this function as an argument can you just do...

```tsx
const {el} = this;
return el.slot !== SLOTS.navSecondary && el.slot !== SLOTS.navSecondary;
```

src/components/navigation/navigation.tsx
```diff
@@ -147,34 +173,45 @@ export class CalciteNavigation {
...
+      this.primaryContentStartSlotHasElements ||
+      this.navAction;
     return (
       <Host>
         <div
           class={{
             [CSS.container]: true,
             [SLOTS.navSecondary]: this.el.slot === SLOTS.navSecondary,
-            [SLOTS.navTertiary]: this.el.slot === SLOTS.navTertiary
+            [SLOTS.navTertiary]: this.el.slot === SLOTS.navTertiary,
```
driskull, line 190: nitpick: Can you create a const for `this.el.slot` since its repeated.

src/components/navigation/navigation.tsx
```diff
@@ -147,34 +173,45 @@ export class CalciteNavigation {
...
         >
-          {isPrimaryLevel && <slot name={SLOTS.progress} />}
+          <div class={{ hide: !this.progressSlotHasElement, progress: true }}>
+            <slot name={SLOTS.progress} onSlotchange={this.handleProgressSlotChange} />
+          </div>
           <div class={CSS.containerContent}>
-            {isPrimaryLevel && this.renderMenuAction()}
-            {isPrimaryLevel && <slot name={SLOTS.logo} onSlotchange={this.handleLogoSlotChange} />}
+            {this.renderMenuAction()}
+            <div class={{ hide: !this.userSlotHasElements, logo: true }}>
```
driskull, line 199: nitpick: Would be nice if all the classes were in a CSS object.

driskull (commented):
👍

src/components/navigation/navigation.tsx
anveshmekala: the problem would be scope either we have to convert this function to an arrow function of pass down the element scope from onSlotChange handler method.

driskull: Oh ok. I was just thinking that passing the same element every time doesn't make much sense as an argument. IMO this would make more sense....

```tsx
private isPrimaryLevel = (el: HTMLCalciteNavigationElement): boolean => {
```

driskull (approved):
👍

Issue #6455, opened 2023-02-10
[safari] Popover returns focus to the wrong element on close


mpayson:
### Actual Behavior

In Firefox & Chrome, when the user closes a popover, the popover returns focus to the reference element. In Safari, it returns the focus to a different element.

### Expected Behavior

The popover returns focus to the reference element in all browsers

### Reproduction Sample

https://codepen.io/mpayson/pen/mdjNdbG

### Reproduction Steps

1. Open code pen
2. Click panel "i" action in the top right to open the popover
3. Click "x" to close the popover
4. See focus is returned to the "i" action in FF / Chrome but not Safari

### Reproduction Version

1.0.4

### Relevant Info

Safari 16.2

### Esri team

ArcGIS Online

geospatialem:
Reproduced in Safari on Mac only. 

Windows browsers - FF, Chrome, and Edge are not affected.

geospatialem:
Research will be conducted in June to determine next steps, and if the focus trap could be responsible. A time estimate could be the resolution of the research.

driskull:
I'm not sure what we can do for this one. Seems to be pretty specific to safari.

driskull:
This one is still broken after focus-trap `7.4.2`. 

After investigating some more, it seems its because in the PopoverManager, we set focus to the opened popup during the `pointerdown` event. If I change it to `pointerup` it allows the target to get focused before we set focus to the popover which then lets the focus be returned properly.

@jcfranco can we change `pointerdown` to `poinerup` without worry here? 

https://github.com/Esri/calcite-components/blob/bcf97cfc89b2756aa980b41d9fb37ac062607f0e/src/components/popover/PopoverManager.ts#L93-L100

I don't think we need to worry about tooltip since it doesn't get focus.

driskull:
I found this: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#clicking_and_focus

Seems like safari just doesn't focus on a button when clicking it. Which is why it doesn't go back to the focus of that button.

@geospatialem I think we can mark this as wontfix.

geospatialem:
Good find, @driskull. Unfortunate it's something known to Safari. 😞 

@mpayson, this is a specific issue to Safari, as Matt mentions above related to Safari [bug 22261](https://bugs.webkit.org/show_bug.cgi?id=22261), so this is something Calcite cannot fix as its browser-specific.

PR #6977, opened 2023-05-17
fix(popover): change event listener to 'pointerup'

driskull:
**Related Issue:** #6455

## Summary

- Changes `PopoverManager` event listener to be 'pointerup' instead of `pointerdown` to allow elements to become focused first. (seems like safari was the only browser with an issue)
  - `focus-trap` listens for `pointerdown` to get the element to return focus to.
- If these elements become focused first then their focus is returned correctly.

### DIFF
--- a/src/components/popover/PopoverManager.ts
+++ b/src/components/popover/PopoverManager.ts
@@ -91,12 +91,12 @@ export default class PopoverManager {
   };
 
   private addListeners(): void {
-    document.addEventListener("pointerdown", this.clickHandler, { capture: true });
+    document.addEventListener("pointerup", this.clickHandler, { capture: true });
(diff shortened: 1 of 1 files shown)


jcfranco (approved):
Awesome!

Is it possible to add a test for this?

driskull:
> Is it possible to add a test for this?

I'm not sure how that would be done. The issue seems to only exist in Safari osx. :\

geospatialem:
@driskull Tested this in Safari, but still getting the same result in `1.4.0-next.23` with Safari using https://codepen.io/geospatialem/pen/RweEYNp, where on click the focus is shifted to the `flow-item` back button.

Do you get a different result?

Issue #6851, opened 2023-04-25
[Time Picker] Group controls for hour, minute and second for keyboard support


geospatialem:
### Description

Create group controls for: hour, minute and second to ensure keyboard users can access content as-expected.

![image](https://user-images.githubusercontent.com/5023024/234159193-2786eeda-c29d-438e-92d3-f05d78198ece.png)

### Acceptance Criteria

- [ ] When tabbing, move between the groups (hour - minute - second), and each group with focus can be nudged via the ⬆️ and ⬇️  keys.
- [ ] Support focus when the component is on focus after opening from the `input-time-picker` component
- [ ] If feasible, support both tabbing and `Left arrow` and `Right arrow` interactions.

### Relevant Info

Came up while working through `input-time-picker` and  `time-picker` a11y updates as part of the April release in #6830 

### Which Component

`time-picker`

### Example Use Case

Keyboard support/access

### Priority impact

p2 - want for current milestone

### Esri team

Calcite (dev)

ashetland:
If focus is moved to a column group, would it still be obvious to the user that they can input numbers as well as arrow up/down?
<img width="202" alt="CleanShot 2023-04-27 at 12 24 04@2x" src="https://user-images.githubusercontent.com/108549080/234970458-03ceff52-33d5-4150-80cb-3071e56f322c.png">

geospatialem:
Verified in `1.4.0-next.23`

![verify-6851](https://github.com/Esri/calcite-components/assets/5023024/3c30a8b3-93c6-4cfe-8423-c623c266b75b)

PR #6982, opened 2023-05-19
fix(time-picker): prevent time-picker part steppers from being focusable

jcfranco:
**Related Issue:** #6851 

## Summary

This ensures the hour, minute and second inputs are the only focusable elements within the component. Previously, focus would land on the hour-increment stepper (due to `tabIndex=1` on steppers and the component using `delegatesFocus: true`), which doesn't listen to up/down keys.


driskull (approved):
👍

eriklharper:
Many thanks for fixing this!  I started noticing that as a result of the focus trap changes :)

Issue #6024, opened 2022-12-13
after sorting blocks a list within a block no longer sorts


AdelheidF:
### Actual Behavior

I have a sortable list of blocks and some of them contain a sortable list. All is well until you re-order the blocks, then the inner list no longer sorts.

### Expected Behavior

sorting keeps working

### Reproduction Sample

https://codepen.io/afreitag/pen/eYKweaz

### Reproduction Steps

1. sort inner list, Food, Rend... - works
2. move block B below block C
3. sorting inner list no longer works

### Reproduction Version

beta.98

### Regression?

don't think so

### Impact

moderate-high

### Esri team

ArcGIS Map Viewer

lau11123:
cc @driskull for when he's back.

@geospatialem is there any status on this? Can a milestone be set please?

geospatialem:
@lau11123 Apologies for the lack of communication, we'll be prioritizing issues after this month's release for inclusion in upcoming releases. Will post an update once we're able to assign a milestone in a few weeks.

lau11123:
@geospatialem thanks!

geospatialem:
Additional research and follow-up needed pertaining to the `list` component sorting in 2023.

driskull:
The problem lies with the `sortable-list` component.  It's `handleSelector` is set to `calcite-handle` which also targets the `value-list`'s calcite-handle. The `handleSelector` prop isn't very useful because you can't target within a specific shadowRoot of a component.

If the `calcite-block` allowed you to slot a handle instead of just adding it for you inside the shadowRoot then this would be doable as you could update the sortable list to only target specific handle elements.

geospatialem:
Verified on `master`.

![verify-6024](https://github.com/Esri/calcite-components/assets/5023024/4e462184-5d08-4c96-8a48-a0041d6a557b)

PR #6983, opened 2023-05-19
fix(value-list, sortable-list): fix nested sorting components

driskull:
**Related Issue:** #6024

## Summary

The main problem is:

- Sortable components set up a `Sortable` instance on `connectedCallback` and destroy `Sortable` on `disconnectedCallback`.
  - When dragging a `Sortable` component with another `Sortable` component inside of it, it triggers the child component's  `connectedCallback`/`disconnectedCallback` methods which initialize/destroy a `Sortable`. This causes JS errors and UI problems.
  - When a Sortable instance is being dragged (active), we should prevent `connectedCallback`/`disconnectedCallback` from doing anything with the inactive `Sortable` instances. (not destroy them & recreate them).

Solution:

- Create an interface for `SortableComponent` components to follow.
  - Make sure the interface does not destroy other nested sortable instances while dragging.


src/demos/sortable-list-nested.html
```diff
@@ -0,0 +1,75 @@
...
+      hr {
+        margin: 35px 0;
+        border-top: 1px solid var(--calcite-ui-border-2);
+      }
+    </style>
+
+    <script src="./_assets/head.js"></script>
+  </head>
+
+  <body>
```
jcfranco, line 38: I think this can be moved into `src/demos/sortable-list.html` instead of adding a new demo page.

src/utils/sortableComponent.ts
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Helper to keep track of a SortableComponent. This should be called in the `connectedCallback` lifecycle method as well as any other method necessary to rebuild the sortable instance.
+ *
+ * @param {SortableComponent} component - The sortable component.
+ * @param {HTMLElement} element - Any variety of HTMLElement.
+ * @param {SortableComponent} [options] - Sortable options object.
+ */
+export function connectSortableComponent(
+  component: SortableComponent,
+  element: HTMLElement,
```
jcfranco, line 24: Do we have any use cases where `element` is different than the component's host? If not, using `component.el` could help simplify the util a bit until we need to support a different root sortable element.

src/utils/sortableComponent.ts
```diff
@@ -0,0 +1,77 @@
...
+  if (inactiveSortableComponentSet.has(component)) {
+    return;
+  }
+
+  component.sortable?.destroy();
+  component.sortable = null;
+}
+
+/**
+ * Helper to deactivate other SortableComponent listeners on `Sortable.onStart`.
```
jcfranco, line 54: I think this doc needs revisiting as the function doesn't appear to deactivate any sortables. 

Applies to `onSortingEnd` too.

jcfranco (commented):
This is looking good! 😎 I think we could iterate on it to make the utils more seamless (i.e., unify observing and sorting steps, if possible).

src/components/sortable-list/sortable-list.tsx
```diff
@@ -165,49 +161,52 @@ export class SortableList implements InteractiveComponent {
 
     this.items = Array.from(this.el.children);
 
-    handle.activated = true;
-    handle.setFocus();
     this.beginObserving();
+    requestAnimationFrame(() => focusElement(handle));
+
+    if ("activated" in handle) {
+      handle.activated = true;
```
driskull, line 168: since handleSelector can be anything we need to do this check. Maybe we shouldn't support a handleSelector if we want to maintain drag consistency thoughout components and require use of a calcite-handle.

src/components/sortable-list/sortable-list.tsx
```diff
@@ -128,7 +121,10 @@ export class SortableList implements InteractiveComponent {
 
   handleNudgeEvent(event: CustomEvent<HandleNudge>): void {
     const { direction } = event.detail;
-    const handle = event.target as HTMLCalciteHandleElement;
+
+    const handle = event
+      .composedPath()
+      .find((el: HTMLElement) => el.matches(this.handleSelector)) as HTMLElement;
```
driskull, line 127: Need to find the handle in the composedPath. This was broken.

src/components/value-list/value-list.tsx
```diff
@@ -313,31 +327,34 @@ export class ValueList<
     this.filteredItems = filteredItems;
   };
 
-  setUpDragAndDrop(): void {
-    this.cleanUpDragAndDrop();
+  setUpSorting(): void {
```
jcfranco, line 330: Would this be something we need defined in the interface as well?

driskull: It could be but it doesn't need to be. I left it out for now since it wouldn't do anything by being added to the interface.

Issue #6574, opened 2023-03-07
Child list-items do not appear in filter results


nwhittaker:
### Actual Behavior

Matching children are absent from a `<calcite-list>` element's filtered results.

### Expected Behavior

Matching children are present in a `<calcite-list>` element's filtered results.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/bGxeMVq

### Reproduction Steps

1. Visit the sample
2. Expand the **Rivers** list-item to see its child list-item named **Estuaries**
3. Type "Estuaries" into the list's filter input

### Reproduction Version

1.0.8

### Priority impact

p2 - want for current milestone

### Impact

A blocker for migrating the Field Map web app to use a `<calcite-list>` element for managing lists of nested layers.

### Esri team

ArcGIS Field Apps

geospatialem:
Verified in `1.4.1-next.2`

![image](https://github.com/Esri/calcite-components/assets/5023024/0c8cb69c-5161-4288-aaee-8a5fd301d474)

PR #6989, opened 2023-05-23
fix(list): filter nested children correctly

driskull:
**Related Issue:** #6574

## Summary

- Refactors and fixes filtering logic in the list component.
- Starts with the filtered list items that are not children of any other list items
- Then works its way up the dom using closest to get any parent groups or list items and sets their `hidden` property if necessary.


src/components/list/list.tsx
```diff
@@ -338,56 +339,56 @@ export class List implements InteractiveComponent, LoadableComponent {
...
+    visible
+  }: {
+    el: HTMLCalciteListItemElement | HTMLCalciteListItemGroupElement;
+    filteredItems: HTMLCalciteListItemElement[];
+    visibleParents: WeakSet<HTMLElement>;
+    visible?: boolean;
+  }): void {
+    const isVisible = filteredItems.includes(el as HTMLCalciteListItemElement) || visible;
+
+    el.hidden = visibleParents.has(el) ? false : !isVisible;
```
jcfranco, line 355: I think this could be simplified as `el.hidden = !visibleParents.has(el) && !isVisible;`.

src/components/list/list.tsx
```diff
@@ -338,56 +339,56 @@ export class List implements InteractiveComponent, LoadableComponent {
...
+    filteredItems,
+    visibleParents,
+    visible
+  }: {
+    el: HTMLCalciteListItemElement | HTMLCalciteListItemGroupElement;
+    filteredItems: HTMLCalciteListItemElement[];
+    visibleParents: WeakSet<HTMLElement>;
+    visible?: boolean;
+  }): void {
+    const isVisible = filteredItems.includes(el as HTMLCalciteListItemElement) || visible;
```
jcfranco, line 353: IMO, the `visible` override might make this slightly tricky to debug.

src/components/list/list.tsx
```diff
@@ -338,56 +339,56 @@ export class List implements InteractiveComponent, LoadableComponent {
...
-        }
-
-        const matches = filterText ? values.includes(item.value) : true;
-
-        item.hidden = !matches;
-
-        if (!hasSelectedMatch) {
-          hasSelectedMatch = matches && item.selected;
-        }
+      listItems.filter((item) => (filterText ? values.includes(item.value) : true)) || [];
```
jcfranco, line 385: This can be simplified to `!filterText || values.includes(item.value)`.

src/components/list/list.tsx
```diff
@@ -338,56 +339,56 @@ export class List implements InteractiveComponent, LoadableComponent {
...
+
   private updateFilteredItems = (emit = false): void => {
     const { listItems, filteredData, filterText } = this;
 
     const values = filteredData.map((item) => item.value);
 
-    const groups = new Set<HTMLCalciteListItemGroupElement>();
-    let hasSelectedMatch = false;
+    const lastDescendantItems = listItems?.filter((listItem) =>
+      listItems.every((li) => (li !== listItem ? !listItem.contains(li) : true))
```
jcfranco, line 381: This can be simplified to `li === listItem || !listItem.contains(li)`.

jcfranco:
Can you update the PR title so that it conveys what was fixed vs what the original issue is?

src/components/list/list.tsx
driskull: Refactored!

PR #6990, opened 2023-05-23
fix(input-time-picker): ensure dynamic date/time locale config imports are compatible with bundlers that require static import paths

jcfranco:
**Related Issue:** #7019

## Summary

This corrects `calcite-input-time-picker` treating `dayjs` locale modules as assets and uses explicit module imports to avoid build errors in bundlers (e.g., Webpack).


driskull (approved):

eriklharper (approved):
Nice, cleans up those scripts back to the way they were :)

PR #6993, opened 2023-05-23
fix(menu): ensure role type is compatible with Stencil's HTMLAttributes interface

jcfranco:
**Related Issue:** #7000 

## Summary

This fixes a TS error caused by `role` prop type not matching Stencil's `HTMLAttributes`. 

Went with the `globalAttributes` approach because making `role` optional would fix the error in a strict-compilation context, but would cause the prerender build to fail with the same error. 

Worth noting that this also adds support for specifying default `globalAttributes` for initial rendering.


anveshmekala (approved):
👍

Issue #7001, opened 2023-05-23
prerender build fails fetching asset bundles


jcfranco:
### Actual Behavior

Running the prerender build shows the following error:

```
Error: Error: could not fetch component message bundle
    at throwMessageFetchError (/Users/juan6600/dev/projects/calcite-components/hydrate/index.js:7762:9)
    at /Users/juan6600/dev/projects/calcite-components/hydrate/index.js:7758:18
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
unhandledRejection Error: could not fetch component message bundle
    at throwMessageFetchError (/Users/juan6600/dev/projects/calcite-components/hydrate/index.js:7762:9)
    at /Users/juan6600/dev/projects/calcite-components/hydrate/index.js:7758:18
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
...
```

### Expected Behavior

No error is shown.

### Reproduction Sample

https://github.com/Esri/calcite-components/

### Reproduction Steps

1. clone repo
2. run `npx stencil build --prerender`
3. see errors race by 🏎️

### Reproduction Version

1.4.0

### Regression?

1.0.0-beta.99

### Priority impact

p2 - want for current milestone

### Impact

Unsure as the prerender build finishes despite the errors.

### Esri team

Calcite (dev)

PR #7002, opened 2023-05-23
fix(hydrate): fix errors caused by message bundle fetch attempts during prerendering

jcfranco:
**Related Issue:** #7001

## Summary

This skips message bundle loading when prerendering.


driskull (approved):
👍

Issue #7010, opened 2023-05-24
[Navigation User / Navigation Logo] Incorrectly inheriting line-height when JSAPI css present


macandcheese:
### Actual Behavior

As reported by @mpayson, these components can be subject to inheriting style from JSAPI css when present on page. This issue was present in `4.26` or before, but `4.27` and later exhibit the issue to a worse extent due to line-height change in the additional css. Not sure what changed on their end, but we can limit any inheritance by adding a couple of line height declarations in our components.

No JSAPI css:  https://codepen.io/mac_and_cheese/pen/oNamozV?editors=1000
JSAPI 4.27 css: https://codepen.io/mac_and_cheese/pen/NWOoXVo
JSAPI 4.26 css: https://codepen.io/mac_and_cheese/pen/ZEqwvdY

### Expected Behavior

The spacing of the components should not look as it does with no JSAPI css on page

### Reproduction Sample

Above

### Reproduction Steps

View the various Codepen links, notice that the one without the JSAPI CSS is the correct spacing.

### Reproduction Version

1.4.0

### Priority impact

p1 - need for current milestone

### Esri team

Calcite (design)

geospatialem:
Verified in `1.4.1-next.2`

PR #7011, opened 2023-05-24
fix(navigation-logo, navigation-user): Prevent inheriting line-height

macandcheese:
**Related Issue:** #7010 

## Summary
This more explicitly sets the line height within our components, adds class name where we were targeting elements previously, and does a bit of clean up of related css.

Before (with css from JSAPI `4.27` - this wasn't an issue with `4.26` or prior):

Screen Shot 2023-05-23 at 7 06 56 PM

After:
Screen Shot 2023-05-23 at 7 06 08 PM

### DIFF
--- a/src/components/navigation-logo/navigation-logo.scss
+++ b/src/components/navigation-logo/navigation-logo.scss
@@ -1,47 +1,45 @@
 :host {
   @apply inline-flex outline-none;
-  & a {
-    @apply flex 
+}
+
+.anchor {
+  @apply flex 
     m-0 
     items-center 
     justify-center 
     cursor-pointer 
     transition-default 
     focus-base 
-    no-underline;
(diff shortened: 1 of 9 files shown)


jcfranco (approved):
Awesome! 📏🚫

Is this already covered by screenshot tests? If not, could you add one before merging?

macandcheese:
> Awesome! 📏🚫
> 
> Is this already covered by screenshot tests? If not, could you add one before merging?

We do have the various states in a few tests, but it wouldn't have caught this from the external css.  I'll add another with the `description` and `username` strings in the main Navigation one though for better coverage.

Separately, maybe we can update the `theme` story to have a few more components and include a variant that loads the latest JSAPI CSS, so we could have a chance at catching components that allow styles to leak in.

driskull (approved):

PR #7016, opened 2023-05-24
fix(list-item): Improve item toggle behavior

driskull:
**Related Issue:** #7017

## Summary

- Honors `event.defaultPrevented` for click/key events so users can prevent them from doing their thing.
- Will no longer toggle selection when selection-mode is set to "none".
- Changes key from toggling selection from space to enter.


src/components/list-item/list-item.tsx
```diff
@@ -604,28 +605,35 @@ export class ListItem
...
     const composedPath = event.composedPath();
-    const { containerEl, contentEl, actionsStartEl, actionsEndEl, open, openable } = this;
+    const { containerEl, contentEl, actionsStartEl, actionsEndEl, open, openable, selectionMode } =
+      this;
 
     const cells = [actionsStartEl, contentEl, actionsEndEl].filter(Boolean);
     const currentIndex = cells.findIndex((cell) => composedPath.includes(cell));
 
-    if (key === " ") {
+    if (isActivationKey(key) && selectionMode && selectionMode !== "none") {
```
driskull, line 644: `treegrid` role says enter should perform default row action and not space. cc @geospatialem what do you think? Should we just honor `Enter` key like the spec defines? 

https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/examples/treegrid-1/

src/components/list-item/list-item.tsx
```diff
@@ -604,28 +605,35 @@ export class ListItem
...
+      this.selected = !selected;
+    } else if (selectionMode === "single-persist") {
       this.selected = true;
     }
 
     this.calciteListItemSelect.emit();
   };
 
   handleItemKeyDown = (event: KeyboardEvent): void => {
+    if (event.defaultPrevented) {
```
driskull, line 632: We can honor users preventing default for keydown.

src/components/list-item/list-item.tsx
geospatialem: Yes, concur with changing to `Enter` from `space` for alignment, and also seems more intuitive. Once updated, can update the docs site to reflect the change (looks like `space` was not previously doc'd).

driskull: @geospatialem should the "default action" be to toggle selection or toggle open state of children?

geospatialem: Was the children open state previously handled by `enter`? Thought it was through the arrow keys? Would assume the default would be toggle selection.

jcfranco (approved):
💥

Could you add a test before merging?

Also, can you update the PR title to reflect what the improvement was?

jcfranco:
Would "improve item toggle behavior" or similar work?

Issue #7025, opened 2023-05-24
[Tab-Title]: tabs no longer center with `layout="center"` 


Elijbet:
### Actual Behavior

<img width="894" alt="Screenshot 2023-05-24 at 2 37 23 PM" src="https://github.com/Esri/calcite-components/assets/19231036/f0fcc075-9101-4898-98cb-33589e7f8244">

### Reproduction Sample

https://codepen.io/nsulzberger_esri/pen/MWPLRgB

### Reproduction Version

1.4.0

### Relevant Info

Possible fix: `:host([centered]) .content { @apply mi-auto; }` on or around https://github.com/Esri/calcite-components/blame/master/src/components/tab-title/tab-title.scss#L95

Possible workaround: https://codepen.io/mpriour/pen/bGmzJZm

### Regression?

Regression caused by https://github.com/Esri/calcite-components/issues/2620

### Priority impact

p1 - need for current milestone

### Esri team

Calcite (dev)

geospatialem:
Verified in `master`.

nsulzberger:
thanks a lot for fixing this, it looks fine now in Scene Viewer with 1.4.1 :)

PR #7026, opened 2023-05-24
fix(tab-title): tabs center when set to `layout='center'`

Elijbet:
**Related Issue:** #7025

## Summary
Enhancement to `tab-title` to make it optionally `closable` introduced structural changes that caused a regression in the tab alignment when centered.


src/components/tab-title/tab-title.scss
```diff
@@ -7,7 +7,9 @@
 :host([layout="center"]) {
   @apply my-0 mx-5 text-center;
   flex-basis: theme("spacing.48");
-  margin: auto;
+  .content {
+    @apply m-auto;
```
alisonailea, line 11: given that there's no difference between `margin: auto` and `@apply m-auto` I'd recommend using `margin: auto` it's more readable

alisonailea (approved):

macandcheese:
Can we add screenshot tests for some specific pre-configured cases like centered + closable, closable + bordered + centered, to hopefully catch some of these?

driskull (approved):
👍

src/components/tab-title/tab-title.scss
Elijbet: Since we're using tailwind for now, it makes sense to stick to tailwind classes if they are available, for consistency. This is what we've been doing so far. Also, overlapping uses of tailwind and regular css props might create oversight doubling use of both by accident.

src/components/tabs/tabs.stories.ts
```diff
@@ -184,6 +184,53 @@ export const disabledTabsAndMediumIconsForLargeTabsTitle_TestOnly = (): string =
...
+  </calcite-tabs>
+`;
+
+export const centeredClosable_TestOnly = (): string => html`
+  <calcite-tabs layout="center">
+    <calcite-tab-nav slot="title-group">
+      <calcite-tab-title closable>Tab 1 Title</calcite-tab-title>
+      <calcite-tab-title icon-start="arrow-left" closable>Tab 2 Title</calcite-tab-title>
+      <calcite-tab-title icon-end="arrow-right" closable>Tab 3 Title</calcite-tab-title>
+      <calcite-tab-title icon-start="arrow-left" icon-end="arrow-right" selected>Tab 4 Title</calcite-tab-title>
```
macandcheese, line 208: This one needs `closable` as well.

Elijbet:
> Can we add some additional screenshot tests for some more specific pre-configured cases like centered + closable, closable + bordered + centered, to hopefully catch some of these?

Adding these tests revealed some mismatch with the centering with regards to Figma. There are various ways to center text/icon and the x button within a tab, as discussed with @macandcheese. 

We were wondering what to prioritize: matching the Figma design to make it absolutely centered (but with variable padding to offset for added close button), or making sure the text/icon is centered between the edge of the tab and the close button. I've applied the latter for your review @SkyeSeitz @ashetland

src/components/tab-title/tab-title.scss
driskull: Agreed, lets keep tailwind for now for consistency. Hopefully we can remove tailwind in the future.

src/components/tab-title/tab-title.scss
```diff
@@ -190,11 +192,21 @@
   }
 }
 
+:host([closable]) {
+  .content {
+    padding-inline-start: 1.5rem;
```
driskull, line 197: @alisonailea @Elijbet are there any tokens we can use as values for these paddings?f

driskull (approved):
🦭 Seal of approval! ☑️ 👍 💯 👏

src/components/tab-title/tab-title.scss
driskull: If this padding should be referenced from `theme()`/tailwind we could do it as a followup.

Elijbet: Let me see what design says about the centering approach in general and yeah that would be the next question.

SkyeSeitz:
Let's prioritize absolutely centering the Tab Titles with variable padding to offset for added close button width.

```
small x-button offset: 24px; core.sizing.9
medium x-button offset: 24px; core.sizing.9
large x-button offset: 32px; core.sizing.11
```

Secondly, I would propose truncating the text when it reaches 8px away from the x-button at all scales. Not sure if this is in scope for this particular issue, though, but this is a better behavior than the current wrapping. 
image

Elijbet:
> Let's prioritize absolutely centering the Tab Titles with variable padding to offset for added close button width.
> 
> ```
> small x-button offset: 24px; core.sizing.9
> medium x-button offset: 24px; core.sizing.9
> large x-button offset: 32px; core.sizing.11
> ```
> 
> Secondly, I would propose truncating the text when it reaches 8px away from the x-button at all scales. Not sure if this is in scope for this particular issue, though, but this is a better behavior than the current wrapping. image

I'll open a separate issue for truncation, as it would need to have the tooltip applied for sighted users, just like the button does.

src/components/tab-title/tab-title.scss
```diff
@@ -7,7 +7,40 @@
...
+    padding-inline-start: 32px; //28px button width + 0.25rem padding
+  }
+}
+
+:host([layout="center"][bordered][closable][scale="s"]) {
+  .content {
+    padding-inline-start: 36px; //28px button width + 0.5rem padding
+  }
+}
+
```
Elijbet, line 27: I worked the offset px/rem amount for the text to stay static in the center when buttons are added or removed, but I'm not sure how it needs to be expressed in terms of tokens. 

@alisonailea suggested I use format var(--core-sizing-9) to reference button width, but the offset here is cumulative of the width and the existing padding. 

Could I get some help with this?  @driskull

src/components/tab-title/tab-title.scss
```diff
@@ -7,7 +7,40 @@
...
+  .content {
+    @apply m-auto;
+  }
+}
+
+// center the text visually and not affected by the x button so as to avoid moving when on or off
+:host([layout="center"][closable]) {
+  .content {
+    padding-inline-start: 32px; //28px button width + 0.25rem padding
+  }
```
Elijbet, line 19: Can I get guidance on how to implement tokens here? 
eg. `padding-inline-start: var(--core-sizing-11);`
This doesn't seem to take any effect.

driskull: @Elijbet this is good for the release. Can you open an issue to address this by using our design tokens/tailwind for this padding? Ideally both the button width and padding should be tied to variables.

Issue #4072, opened 2022-02-07
calcite-color-picker: Add ability to paste hex (with # symbol) into hex input


jmanke:
### Description

Currently, pasting a color with the full hex string (ex. #FF00FF) into the hex input will not work. The last character is truncated and the value is not accepted. This is something users routinely do in Dashboards. 

### Acceptance Criteria

Pasting a full hex string (with the # symbol included) should update the color-picker's color.

### Which Component

calcite-color-picker

### Example Use Case

A user is configuring their theme using the calcite-color-picker and would like to use hex colors from an external source. The user will attempt to paste a hex color into the hex input but it will fail to update the color. A user would expect the hex input to accept hex strings that include the # symbol.

jcfranco:
@benelan I triaged this one to this milestone since I already had something in the works for it.

benelan:
Unfortunately, in `beta.94` it doesn't let you paste valid hex at all now. Also, should it be using `internalSetValue()` instead of only setting the Input's value? 

https://github.com/Esri/calcite-components/blob/ed9306124c6b0570fe3606a47b33840aeab129bb/src/components/color-picker-hex-input/color-picker-hex-input.tsx#L187

The way it is now won't update Color Picker's value property, as well as some other states like `previousNonNullValue` and `internalColor`. It also won't emit, which is taken care of in `internalSetValue()`.

https://codepen.io/benesri/pen/vYjgQBX?editors=100

jcfranco:
Taking another look! 👀

geospatialem:
This issue will be prioritized after the 1.0 release next month.

jcfranco:
This was addressed via https://github.com/Esri/calcite-components/issues/749 and will be available in the next release ([see updated demo](https://codepen.io/jcfranco/pen/vYVpXYx)). cc @geospatialem

geospatialem:
Verified in `1.4.0-next.9` 🎨

AdelheidF:
I tried this with 1.4.0 and I don't think it works correctly.

When you paste a new hex value the color does not change. It also does not change if you click anywhere else outside of the hex box.

If you type in a new hex value the color only changes if you click directly on the RGB/HSV label or the +/- buttons.

https://codepen.io/afreitag/pen/bGmZGPR

jcfranco:
Thanks for catching this, @AdelheidF. Need to improve test coverage for this. 

I'll take a look to see if this can be fixed for the upcoming patch.

[...]

PR #7030, opened 2023-05-25
fix(color-picker): fix pasting hex with leading # character

jcfranco:
**Related Issue:** #4072 

## Summary

This ensures pasted values with `#` are committed on blur since the event is canceled to intercept and modify the text value.  

**Note**: there is no corresponding E2E test due to the puppeteer limitations outlined [here](https://github.com/Esri/calcite-components/pull/5189#issue-1344620750).


benelan (approved):
:+1:

Issue #7022, opened 2023-05-24
[Button] Buttons are briefly not clickable after being enabled


LanceStasinski:
### Actual Behavior

When a `<calcite-button>` is clicked immediately after the `disabled` attribute is set to `false`, the click will not trigger an event. This mostly impacts tests where the disabled state change and click are milliseconds apart, but I cannot definitively say that this does not affect real (non-test) use cases. 

### Expected Behavior

A `<calcite-button>` that has it's `disabled` attribute set to `false` should become immediately clickable.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/xxyBKyP

### Reproduction Steps

1. Click on the button.
2. Notice that 'click' is not logged in the console.

### Reproduction Version

1.4.0

### Relevant Info

MacOS: Ventura 13.3.1, Chrome 113.0.5672.126

### Regression?

1.2.0

### Priority impact

p2 - want for current milestone

### Impact

This regression breaks most of our tests that involve toggling the `disabled` state of a `<calcite-button>` immediately followed by triggering an event on the same `<calcite-button>`.

### Esri team

ArcGIS Field Apps

geospatialem:
Verified in `master`

PR #7031, opened 2023-05-25
fix: ensure mouse events are blocked only when components are disabled

jcfranco:
**Related Issue:** #7022

## Summary

This updates `interactive.ts` to prevent blocking of mouse/pointer events unnecessarily. This is needed if mouse/pointer events happen right after toggling `disabled` on an interactive component.


driskull (approved):
👍

PR #7033, opened 2023-05-25
docs(changelog): split beta entries into separate file

benelan:
## Summary

Separates `beta` changelog entries into a separate file to slim down the file size.


driskull:
Looks good but don't we need to update our scripts to handle building these?

driskull (approved):

benelan:
We should be fine since we aren't adding to the beta changelog and the current changelog filename didn't change

PR #7037, opened 2023-05-25
fix(input-time-picker): resolve dayjs type error

benelan:
**Related Issue:** #6990

## Summary

Fixes a type error due to dynamically importing the dayjs module in an exported Map. Caught by `npm run util:test-types`.

Maybe we should run the type test in a PR Check? cc @jcfranco


driskull (approved):
👍

Issue #7036, opened 2023-05-25
input-time-picker doesn't support sending in a lowercase locale code as lang


eriklharper:
### Actual Behavior

Input Time Picker throws a console error and fails to render when sending in certain lang codes in lowercase that are typically spelled with uppercase geographical regions, like `en-us` (which is typically typed as `en-US`.  

![image](https://github.com/Esri/calcite-components/assets/821864/d691a846-d687-467a-9f09-c83cd43bbf06)

### Expected Behavior

Input Time Picker should support sending in any valid locale code with all lowercase letters.

### Reproduction Sample

https://codepen.io/eriklharper/pen/ZEqPMYR?editors=1111

### Reproduction Steps

1.  Render  `<calcite-input-time-picker lang="en-us"></calcite-input-time-picker>`
2. Result: nothing renders

### Reproduction Version

1.4.0

### Priority impact

p1 - need for current milestone

### Esri team

ArcGIS Map Viewer

geospatialem:
Verified in `master`.

PR #7038, opened 2023-05-25
fix(input-time-picker): allow en-us and other supported lowercase locale codes

eriklharper:
**Related Issue:** #7036 

## Summary

This fixes an issue where passing in `en-us` instead of `en-US` (or inheriting the lang from the HTML tag) was blocking rendering and throwing a console error.

### DIFF
--- a/src/components/input-time-picker/input-time-picker.tsx
+++ b/src/components/input-time-picker/input-time-picker.tsx
@@ -501,27 +501,18 @@ export class InputTimePicker
   };
 
   private async loadDateTimeLocaleData(): Promise<void> {
-    const { effectiveLocale } = this;
-    let dayjsLocale = effectiveLocale.toLowerCase();
+    const normalizedLocale = this.getNormalizedLocale();
 
-    if (dayjsLocale === "en" || dayjsLocale === "en-us") {
+    if (normalizedLocale === "en" || normalizedLocale === "en-us") {
       return;
     }
 
-    if (dayjsLocale === "pt-PT") {
-      dayjsLocale = "pt";
-    }
-
-    if (dayjsLocale === "no") {
-      dayjsLocale = "nb";
-    }
-
     const { default: localeConfig } = await supportedDayJsLocaleToLocaleConfigImport.get(
-      dayjsLocale
+      normalizedLocale
     )();
 
     dayjs.locale(localeConfig, null, true);
-    dayjs.updateLocale(dayjsLocale, this.getExtendedLocaleConfig(dayjsLocale));
+    dayjs.updateLocale(normalizedLocale, this.getExtendedLocaleConfig(normalizedLocale));
   }
 
   private getExtendedLocaleConfig(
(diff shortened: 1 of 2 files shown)


src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -502,18 +502,17 @@ export class InputTimePicker
 
   private async loadDateTimeLocaleData(): Promise<void> {
     const { effectiveLocale } = this;
+    let dayjsLocale = effectiveLocale.toLowerCase();
```
driskull, line 505: could `effectiveLocale` be null? If so should there be a check?

driskull, line 505: Should we call this normalizedLocale to be consistent with the other code? Maybe a function to normalize the locale would promote consistency since `toLowerCase` is used twice for the same thing.

This logic is repeated, I think a function would be beneficial. Something that converts the effectiveLocale to the format we expect.

```ts
const normalizedLocale = effectiveLocale.toLowerCase();

    if (normalizedLocale === "en-us") {
      effectiveLocale = "en";
    }
    if (effectiveLocale === "pt-PT") {
    if (normalizedLocale === "pt-pt") {
      effectiveLocale = "pt";
    }
    if (effectiveLocale === "no") {
    if (normalizedLocale === "no") {
      effectiveLocale = "nb";
    }
```

driskull (approved):
👍 

Looks good but I think we can make the code more efficient with some kind of function that will format the effectiveLocale to a format we need to work with. The code added seems a little redundant.

src/components/input-time-picker/input-time-picker.tsx
eriklharper: Good call, I'll follow this up with just such a function.  I agree it would be nice to handle this logic all in one place.

eriklharper: It should always default to `en` if this is sent in as a null value, but couldn't hurt to do a null check here.

jcfranco: Under normal circumstances, `effectiveLocale` will never be null, so need for the null check.

jcfranco: I think we could add another context to [`locale.getSupportedLocale`](https://github.com/Esri/calcite-components/blob/master/src/utils/locale.ts#L153) for this since it already tackles normalization. Let's sync up to do this separately.

On related note, I was thinking on adding a module to abstract away `dayjs` to not have to do this per component.

src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -706,13 +705,15 @@ export class InputTimePicker
     connectLocalized(this);
 
     let { effectiveLocale } = this;
-    if (effectiveLocale === "en-US") {
+    const normalizedLocale = effectiveLocale.toLowerCase();
```
jcfranco: I think you could keep the previous code if you use `locale.getSupportedLocale`, but that's a moot point if we add a new `dayjs` (or other named) context to it.

jcfranco (approved):
Thanks for fixing this.

Please add a new test before merging.

Issue #6407, opened 2023-02-01
Modal `beforeClose` behaves differently when removing `open` attribute


nwhittaker:
### Actual Behavior

Closing a modal via its built-in UI (close button, escape key, or outside click) can be interrupted by having its `beforeClose()` callback return a rejected promise.

However closing a modal by removing its `open` attribute can not be interrupted in the same way. The modal also enters into an inconsistent state where its `open` and `isOpen` properties are out-of-sync and an `overflow-hidden` class remains on the document element.

### Expected Behavior

Closing the modal by removing the `open` attribute and having its `beforeClose()` callback return a rejected promise does no leave the modal in an inconsistent state.

Ideally the `open` attribute is restored, but if not, maybe generate a warning for this scenario and then provide an API method that can be used to programmatically close a modal in a way that is interruptible by the `beforeClose()` callback.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/JjBeJVr

### Reproduction Steps

1. Visit the pen and click the **Open** button to open the modal.
2. Click the modal's **X** button and see it does not close, as expected because `beforeClose` returns a rejected promise.
3. Click the modal's **Close** button and see the modal disappears, unexpectedly.
4. Use devtools to see the pen iframe's `<html>` element retains the modal's `overflow-hidden` CSS class and the `.modal` element within the modal's shadow-dom retains the `modal--open` class.

### Reproduction Version

1.0.3

### Impact

Having `beforeClose()` reject is useful for flows where the app needs to confirm the closure of a modal.

One workaround is for the consumer to run `beforeClose()` themselves before actually removing the `open` attribute. However this leads to some more complicated code to prevent `beforeClose()` from running more than once per close attempt.

### Esri team

ArcGIS Field Apps

geospatialem:
Research needed to determine next steps, if related to #6279, or if additional mitigation is needed.

brittneytewks:
Hey @driskull, for research consideration in this milestone for time estimate and priority to be added

driskull:
Maybe we need to introduce a public close method that would handle this and return a promise. cc @jcfranco 

> Use devtools to see the pen iframe's <html> element retains the modal's overflow-hidden CSS class and the .modal element within the modal's shadow-dom retains the modal--open class.

I think its weird that the watcher is calling beforeClose. The watcher should just honor what was done and remove the necessary classes.

driskull:
Changed the estimate since this requires a new pattern discussion.

[...]

Issue #6379, opened 2023-01-30
Closing a modal using built-in UI calls `beforeClose` twice


nwhittaker:
### Actual Behavior

Closing a modal via its built-in close button, the `esc` key, or clicking outside the modal invokes the modal's `beforeClose` callback twice.

### Expected Behavior

Closing a modal via its built-in close button, the `esc` key, or clicking outside the modal invokes the modal's `beforeClose` callback once.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/QWBVvYP

### Reproduction Steps

1. Visit the pen and open the console.
2. Close the modal via either its close button, pressing the `esc` key, or clicking outside the modal.
3. Notice `beforeClose` is logged twice.

### Reproduction Version

1.0.3

### Impact

The Field Maps web app has some modal's that need to confirm their closure in cases where data loss could occur. We'd like to use the `beforeClose` callback to do this, but having it called twice in some cases could introduce unintended side-effects.

### Esri team

ArcGIS Field Apps

driskull:
I think this is because the watcher for `open` is calling close as well as the click handler.

geospatialem:
Verified with `1.7.0-next.9`.

PR #7042, opened 2023-05-26
fix(modal): changing the open property no longer calls beforeClose

driskull:
**Related Issue:** #6407 #6379

## Summary

- Changing the property `open` will no longer call `beforeClose`.
  -  Its up to the developer to call `beforeClose` when modifying the property.
  - This ensures that the classes added by modal will always be cleaned up.
- User interaction such as clicking the close button will call `beforeClose`
- Cleanup


driskull:
@jcfranco I heard back from Stencil regarding validation in the watcher. Here is there response:

> Using a watcher to validate and potentially revert property changes based on certain conditions is a valid approach in Ionic apps. The code snippet you provided demonstrates the usage of a watcher to observe changes to the myProperty and perform validation logic.

> Based on the information you have provided the approach you described using a watcher to validate and potentially revert property changes is acceptable and aligns with the purpose of watchers in Ionic apps.
Thank you for using ionic,

So we can revert changes in the watcher. However, since our `beforeClose` property is async, there is a delay before the property is reverted which leads to unwanted animation changes.

Jun-02-2023 11-51-13

If the `beforeClose` takes any significant time it will basically close the modal and reopen it. The shifting is even noticeable with a function that just returns a resolved promise.

I think I lean towards not having the watcher flip the property for this reason. What do you think? If we don't flip the property, users will just have to call `beforeClose` before toggling the property.

benelan (commented):
Code LGTM, but some people may see this as a breaking change since they may be relying on this behavior. I'll wait for Franco for final review due to that.

Nitpick: can you change "should" to "does" in the PR title. Or: "... property no longer calls beforeClose"

driskull:
Something else that we could do instead of this would be to introduce an internal state that handles the open prop as well as the beforeClose. So if a user sets the open prop to false it won't actually close until both the open prop and beforeClose have resolved. That would eliminate any temporary style changes. We would likely need to update our CSS to not use the open attribute and use a class instead.

@jcfranco is this what you want to pursue?

driskull:
Ping @jcfranco

PR #7047, opened 2023-05-26
fix(tab-nav): justify centered `tab-titles` evenly to extend across the width of `tab-nav`

Elijbet:
**Related Issue:** #7025 

## Summary

Follow-up to fix to #7026.

Centered `tabs` should spread the `tab-titles` evenly to extend across the entire width of `tab-nav`, as opposed to `inline tab-titles` that only justify against the start of `tab-nav` width.

Added coverage for both `centeredTabsAreEvenlyJustifiedAcrossNavWidth` and `inlineTabsJustifyAgainstTheStartOfTheNavWidth`.

### DIFF
--- a/src/components/tabs/tabs.stories.ts
+++ b/src/components/tabs/tabs.stories.ts
@@ -220,7 +220,6 @@ export const centeredBorderedClosable_TestOnly = (): string => html`
   <calcite-tabs layout="center" bordered>
     <calcite-tab-nav slot="title-group">
       <calcite-tab-title closable>Tab 1 Title</calcite-tab-title>
-      <calcite-tab-title icon-start="arrow-left" closable>Tab 2 Title</calcite-tab-title>
       <calcite-tab-title icon-end="arrow-right" closable>Tab 3 Title</calcite-tab-title>
       <calcite-tab-title icon-start="arrow-left" icon-end="arrow-right" closable selected
         >Tab 4 Title</calcite-tab-title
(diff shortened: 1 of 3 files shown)


driskull (approved):
👍 

Will this trigger a screenshot failure? If not can we add a test for it?

Elijbet:
> 👍
> 
> Will this trigger a screenshot failure? If not can we add a test for it?

Yeah, I think because I had too many tabs in the screenshot setup to cover all prop variations, it wasn't obvious tabs fail to spread to the whole length. That's how it got missed. I'll modify one now.

src/components/tabs/tabs.stories.ts
```diff
@@ -220,7 +220,6 @@ export const centeredBorderedClosable_TestOnly = (): string => html`
   <calcite-tabs layout="center" bordered>
     <calcite-tab-nav slot="title-group">
       <calcite-tab-title closable>Tab 1 Title</calcite-tab-title>
-      <calcite-tab-title icon-start="arrow-left" closable>Tab 2 Title</calcite-tab-title>
```
driskull, line 223: I think this one should be reverted.

src/components/tabs/tabs.stories.ts
```diff
@@ -233,6 +232,21 @@ export const centeredBorderedClosable_TestOnly = (): string => html`
   </calcite-tabs>
 `;
 
+export const centeredTabsAreEvenlyJustifiedAcrossNavWidth_TestOnly = (): string => html`
```
driskull, line 235: 👍

Issue #7040, opened 2023-05-26
Input-time-picker does not accept normal locale (nl-nl) code anymore 


arjanvanzutphen:
### Actual Behavior

The input time picker does not get rendered with the same lang property value as in 1.3.0.
The behavior looks this bug: https://github.com/Esri/calcite-components/issues/7036

Changing the locale value from 'nl-nl' to 'nl' works.

### Expected Behavior

Expecting to see a input time picker with the lang property set to 'nl-nl'

### Reproduction Sample

Codepen of the input-time-picker sample from https://developers.arcgis.com/calcite-design-system/components/input-time-picker/

### Reproduction Steps

1. Go to the input time picker page: https://developers.arcgis.com/calcite-design-system/components/input-time-picker/
2. Go to it's codepen sample
3. Add lang='nl-nl' property to the input-time-picker web component
4. The input-time-picker does not get rendered
5. Console errors a missing file. nl-nl.js is missing in dayjs, because it's now nl.js
6. Changing the lang property to 'nl' makes the input-time-picker appear again

### Reproduction Version

1.4.0

### Relevant Info

Uncaught (in promise) TypeError: Failed to fetch dynamically imported module: https://js.arcgis.com/calcite-components/1.4.0/assets/input-time-picker/nls/dayjs/locale/nl-nl.js

### Regression?

1.3.0

### Priority impact

p3 - want for upcoming milestone

### Esri team

Professional Services - Services Delivery

geospatialem:
Verified in `master`.

PR #7049, opened 2023-05-26
fix(input-time-picker): only load valid dayjs locale files and fallback to base lang when region code is unsupported

eriklharper:
**Related Issue:** #7040

## Summary

This fixes an issue where setting `lang` to a locale code with an unsupported region (e.g. `nl-nl`) failed to fall back to the base language and didn't render the component.

### DIFF
--- a/src/components/input-time-picker/input-time-picker.e2e.ts
+++ b/src/components/input-time-picker/input-time-picker.e2e.ts
@@ -37,6 +37,14 @@ describe("calcite-input-time-picker", () => {
     renders(`<calcite-input-time-picker lang="nl-nl"></calcite-input-time-picker>`, { display: "inline-block" });
   });
 
+  describe("renders with pt-PT locale", () => {
+    renders(`<calcite-input-time-picker lang="pt-PT"></calcite-input-time-picker>`, { display: "inline-block" });
+  });
+
+  describe("renders with no locale", () => {
+    renders(`<calcite-input-time-picker lang="no"></calcite-input-time-picker>`, { display: "inline-block" });
+  });
+
   describe("honors hidden attribute", () => {
     hidden("calcite-input-time-picker");
   });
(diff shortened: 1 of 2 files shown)


src/components/input-time-picker/input-time-picker.e2e.ts
```diff
@@ -643,7 +647,7 @@ describe("calcite-input-time-picker", () => {
     it("localizes initial display value in arab numbering system", async () => {
       const page = await newE2EPage();
       await page.setContent(
-        `<calcite-input-time-picker step="1" lang="ar" numbering-system="arab" value="14:02:30"></calcite-input-time-picker>`
+        `<calcite-input-time-picker step="1" lang="ar" numbering-system  ="arab" value="14:02:30"></calcite-input-time-picker>`
```
jcfranco: Nitpick: unnecessary extra space here.

src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -555,18 +556,16 @@ export class InputTimePicker
...
+    const supportedLocale = getSupportedLocale(this.effectiveLocale).toLowerCase();
 
-    if (normalizedLocale === "en" || normalizedLocale === "en-us") {
-      return;
-    }
-
-    const { default: localeConfig } = await supportedDayJsLocaleToLocaleConfigImport.get(
-      normalizedLocale
-    )();
+    if (supportedDayJsLocaleToLocaleConfigImport.has(supportedLocale)) {
```
jcfranco, line 561: You still need the previous mappings you had for `pt-PT` and `no`. Once this is in place, you no longer need to check if the locale is in the import map since all supported locales are mapped to a corresponding locale config.

jcfranco (commented):
Looking good! 😎

jcfranco (approved):
Awesome! 🎉

> Once this is in place, you no longer need to check if the locale is in the import map since all supported locales are mapped to a corresponding locale config.

This is still pending, but we can tackle that separately.

src/components/input-time-picker/input-time-picker.tsx
eriklharper: good catch, updated!

Issue #6357, opened 2023-01-26
Unable to select label text in radio group


nwhittaker:
### Actual Behavior
Hovering the text label of a <calcite-radio-button> inside a <calcite-radio-button-group> shows a text style cursor. Holding down the mouse button and dragging the label text does not select the text and does check the assoc. radio control when the mouse button is released.

### Expected Behavior
Hovering the text label of a <calcite-radio-button> inside a <calcite-radio-button-group> shows an arrow cursor. Dragging the cursor across the text label does select the text and does not select the assoc. radio control.

### Reproduction Sample
https://developers.arcgis.com/calcite-design-system/components/radio-button-group/

### Reproduction Steps
Visit the radio-button-group doc page and try to perform selections on the radio button labels.
### Reproduction Version
1.0.3

### Relevant Info
See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio for an example of the expected behavior.

### Regression?
No response

### Impact
Impact is to users who expect radio button labels to behave in the same way native radio button labels behave. A common use case might be to want to copy label text for some other use.

### Esri team
ArcGIS Field Apps

geospatialem:
Research will be conducted to determine feasibility/next steps in June.

anveshmekala:
Need to determine the focus behavior when the text is selected to solve this issue since focus and selection behavior is different in chrome/edge compared to safari/FF.

PR #7050, opened 2023-05-26
fix(radio-button-group): no longer focus first radio button on label click and adds `setFocus` method.

### LINKED ISSUE #6698: Clicking a radio button label flashes focus on the group's first radio button
### Actual Behavior

Given a radio button group with no descendants in focus, mouse down on any of the group's radio button labels except the first. While the mouse button is down, focus is given to the group's first radio button.

Screenshot 2023-03-29 at 12 12 15 PM

### Expected Behavior

Given a radio button group with no descendants in focus, mouse down on any of the group's radio button labels except the first. While the mouse button is down, focus is given to the radio button associated with the label that the mouse is on top of.

Screenshot 2023-03-29 at 12 22 05 PM

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/radio-button-group/

### Reproduction Steps

1. Visit the doc site and scroll to the **Sample** section.
2. Mouse down on the _Bushes_ label and see focus is given to the _Trees_ radio button instead of the _Bushes_ radio button.

### Reproduction Version

1.2.0-next.0

### Relevant Info

macOS 13, Chrome 111.

### Regression?

_No response_

### Priority impact

p4 - not time sensitive

### Impact

Could lead to brief user confusion about what radio button they're actually clicking.

### Esri team

ArcGIS Field Apps

anveshmekala:
**Related Issue:**  #6698

## Summary

This PR will remove the focus flash from the first `calcite-radio-button` in group on label click 

Additional enhancements:
- Adds `setFocus( )` method in `calcite-radio-button-group`
- Users will be able to select label text in chrome. #6357 

Note:
With this PR, `delegateFocus` is removed from `calcite-radio-button-group` which is causing #6698. Because the tabIndex is set to 0 for the first focusable element on page load to enable keyboard navigation into the group it makes  `delegateFocus` to apply focus styles on the first focusable element when the user clicks on the label (invokes .focus( ))  resulting in the behavior mentioned in the issue 6698.  To resolve this, `delegateFocus` is removed and added `setFocus( )` method.


src/components/radio-button-group/radio-button-group.tsx
```diff
@@ -126,6 +143,20 @@ export class RadioButtonGroup {
     }
   };
 
+  private getFocusableRadioButton(): HTMLCalciteRadioButtonElement | null {
```
driskull, line 146: Could you use the `focusFirstTabbable` helper in the dom utils?

This is what the modal does...

```tsx
 @Method()
    await componentLoaded(this);
    focusFirstTabbable(this.el);
  }
```

Would we want only 1 radio button within a group tabbable at a time? Seems like that might make for a good enhancement in the future.

src/components/radio-button-group/radio-button-group.tsx
```diff
@@ -137,6 +168,26 @@ export class RadioButtonGroup {
...
+  /** Sets focus on the fist focusable `calcite-radio-button` element in the component. */
+  @Method()
+  async setFocus(): Promise<void> {
+    await componentLoaded(this);
+    if (this.selectedItem && !this.selectedItem.disabled) {
+      this.selectedItem.setFocus();
+      return;
+    }
+    if (this.radioButtons.length > 0) {
+      const focusableRadioButton = this.getFocusableRadioButton();
```
driskull, line 186: Would this work?

```
this.getFocusableRadioButton()?.setFocus();
```

driskull (commented):
Looks good 👍

src/components/radio-button-group/radio-button-group.tsx
anveshmekala: tried implementing `focusFirstTabbable` helper here,  it works well if none if the radio-button's in the group are disabled. The reason for this is the way we assign tabIndex at the `radio-button` level. The `tabIndex=0` only if the radio-button is `checked` or if its the first element in the group. I think this is a bug. Wonder if the assignment of `tabIndex` should stay at the group level not the `radio-button` level. 


> Would we want only 1 radio button within a group tabbable at a time? Seems like that might make for a good enhancement in the future.

The above is following w3 radio group pattern keyboard specs i believe. One button is tabbable and the user can use Arrow Keys to navigate between the items.

src/components/radio-button-group/radio-button-group.tsx
anveshmekala: this works as expected. Updated.

packages/calcite-components/src/components/radio-button-group/radio-button-group.tsx
```diff
@@ -126,6 +143,20 @@ export class RadioButtonGroup {
     }
   };
 
+  private getFocusableRadioButton(): HTMLCalciteRadioButtonElement | null {
```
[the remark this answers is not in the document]
driskull: The other option would to be to use the `tabbable` dependency to just get the first tabbable element and focus it.

packages/calcite-components/src/components/radio-button-group/radio-button-group.tsx
```diff
@@ -6,24 +6,30 @@ import {
...
+} from "../../utils/loadable";
 
 /**
  * @slot - A slot for adding `calcite-radio-button`s.
  */
 @Component({
   tag: "calcite-radio-button-group",
   styleUrl: "radio-button-group.scss",
-  shadow: {
-    delegatesFocus: true
```
driskull, line 23: Do we need to remove delegatesFocus for this component? If so can you summarize in the PR body? Just to document it since it seems like delegatesFocus is actually something we want to do more of.

driskull (commented):
comments! :)

packages/calcite-components/src/components/radio-button-group/radio-button-group.tsx
```diff
@@ -126,6 +143,20 @@ export class RadioButtonGroup {
     }
   };
 
+  private getFocusableRadioButton(): HTMLCalciteRadioButtonElement | null {
+    let index = 0;
+    let focusableEle = this.radioButtons[index];
```
driskull, line 148: For the variable name, can it be `focusableEl` instead of `focusableEle` to be consistent with other variable names?

packages/calcite-components/src/components/radio-button-group/radio-button-group.tsx
anveshmekala: Added.

packages/calcite-components/src/components/radio-button-group/radio-button-group.tsx
anveshmekala: yup it should be `focusableEl`.

packages/calcite-components/src/components/radio-button-group/radio-button-group.tsx
```diff
@@ -126,6 +143,11 @@ export class RadioButtonGroup {
     }
   };
 
+  private getFocusableRadioButton(): HTMLCalciteRadioButtonElement | null {
+    const index = this.radioButtons.findIndex((radiobutton) => !radiobutton.disabled);
+    return index >= 0 ? this.radioButtons[index] : null;
```
driskull, line 148: @anveshmekala if you're returning the button, why not just use `.find()` instead of `.findIndex`?

```ts
return this.radioButtons.find((radioButton) => !radiobutton.disabled) ?? null;
```

packages/calcite-components/src/components/radio-button/radio-button.tsx
```diff
@@ -275,7 +278,7 @@ export class RadioButton
     if (this.disabled) {
       return undefined;
     }
-    return this.checked || this.isDefaultSelectable() ? 0 : -1;
+    return this.checked || this.isFocusable() ? 0 : -1;
```
driskull, line 281: If a radio-button is added as the first radio-button after all the other radio buttons have already been added to the DOM my guess is that there will be two focusable radio buttons.

We may want to consider creating a registry of radio buttons that are keyed by the name property so that we can ensure only one is ever focusable. @anveshmekala what do you think?

driskull (approved):
👍

packages/calcite-components/src/components/radio-button-group/radio-button-group.tsx
anveshmekala: my bad, i checked in the wrong commit. `.find( )` is the optimal choice here.

packages/calcite-components/src/components/radio-button/radio-button.tsx
driskull: Since radio button's don't always belong to a group we can't rely on a mutation observer of a parent group so we likely need some kind of map/registry to maintain the active focusable radio button for a specific `name` property. This logic would make sure only one radio button with the same name value can be focusable. It should be set/updated anytime a radio-button connectedCallback/disconnectedCallback occurs.

driskull: Probably something like...

```ts
type RootNode = Document | ShadowRoot;
type ActiveRadioButtonLookup = WeakMap<RootNode, Map<string, HTMLCalciteRadioButtonElement>>;
```

anveshmekala: There is a method `queryButtons` in radio-button component which queries all the radio-buttons with a similar name for a given radio-button. This will ensure any mutations to the tabIndex attribute of radio-button will be in the same group. 

I tried addressing edge cases  with the existing setup but it is breaking other features.  Will probably split this PR tomorrow in to individual PR's since the issue we have is related to#7113 and prototype with `WeakMap`.

driskull (approved):
👍 Looks good!

PR #7058, opened 2023-05-30
fix(list-item): always emit calciteListItemSelect on activation

driskull:
**Related Issue:** https://github.com/Esri/calcite-components/issues/7059

## Summary

- Always emit calciteListItemSelect on activation


jcfranco (approved):
💥

Issue #7013, opened 2023-05-24
[Navigation] `label` property is required even if `navigationAction` is disabled


mpayson:
### Actual Behavior

The `label` property is typed as required, but it is only used to label the navigation action, so it shouldn't be required when the navigation action is disabled. cc @macandcheese 

### Expected Behavior

The `label` property is optional

### Reproduction Sample

https://github.com/Esri/calcite-components/blob/master/src/components/navigation/navigation.tsx#L56

### Reproduction Steps

See typings and doc description

### Reproduction Version

1.4.0

### Priority impact

p4 - not time sensitive

### Esri team

ArcGIS Online

PR #7084, opened 2023-06-01
fix(navigation): label is no longer a required property

anveshmekala:
**Related Issue:** #7013 

## Summary

This PR will remove `label` as required property from `calcite-navigation` component.


driskull (approved):
👍

PR #7089, opened 2023-06-02
test: set up `disabled` helper to run a test per use case

Elijbet:
**Related Issue:** #N/A

## Summary
Refactor `disabled` test helper to help mitigate timeouts.

Splitting up our common test helpers into individual tests helps mitigate timeouts that occurred when a helper's coverage ended up taking longer than the default test timeout.

The updated doc reflects how to use the helpers in E2E tests.


jcfranco (commented):
Looking good! 😎

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -965,125 +959,178 @@ export async function disabled(
...
-    const focusTarget = element.shadowRoot.activeElement || element;
-    const rect = focusTarget.getBoundingClientRect();
+    if (options.focusTarget === "none") {
+      await page.click(tag);
+      await expectToBeFocused(page, "body");
 
-    return [rect.x + rect.width / 2, rect.y + rect.height / 2];
-  });
+      /* eslint-disable-next-line jest/no-conditional-expect --
+       * Conditional logic here is confined to a test helper and its purpose is to handle specific scenarios/variations in the test setup.
```
jcfranco, line 1015: Could the explainer be simplified a bit? We might need to reuse it in other helpers and it'd be awesome if we could keep it short and sweet. 

Sidebar: we could disable `jest/no-conditional-expect` in the whole file since a lot of test helpers need to assert conditionally.

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -916,45 +916,39 @@ async function getTagAndPage(componentTestSetup: ComponentTestSetup): Promise<Ta
...
-  const eventsExpectedToNotBubble = ["mousedown", "mouseup", "click"];
-  const allExpectedEvents = [...eventsExpectedToBubble, ...eventsExpectedToNotBubble];
-
-  const eventSpies: EventSpy[] = [];
-
-  for (const event of allExpectedEvents) {
-    eventSpies.push(await component.spyOnEvent(event));
-  }
+): void {
+  const addClickListenersWithRedirectPrevention = async (page: E2EPage, tag: string): Promise<void> => {
```
jcfranco, line 933: I think we can drop the click listeners part since it's an implementation detail. The redirect prevention is the key part. `addClickListenersWithRedirectPrevention` ➡️ `addRedirectPrevention`/`preventRedirects`?

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -965,125 +959,178 @@ export async function disabled(
...
-    await page.click(tag);
-    await expectToBeFocused("body");
+  const createEventSpiesForExpectedEvents = async (component: E2EElement): Promise<EventSpy[]> => {
+    const eventSpies: EventSpy[] = [];
 
-    assertOnMouseAndPointerEvents(eventSpies, (spy) => expect(spy).toHaveReceivedEventTimes(1));
+    for (const event of allExpectedEvents) {
+      eventSpies.push(await component.spyOnEvent(event));
+    }
+    return eventSpies;
```
jcfranco, line 973: Nitpick: add newline above the return to preserve spacing between function parts.

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -965,125 +959,178 @@ export async function disabled(
...
+    const eventSpies = await createEventSpiesForExpectedEvents(component);
 
-  expect(tabFocusTarget).not.toBe("body");
-  await expectToBeFocused(tabFocusTarget);
+    expect(component.getAttribute("aria-disabled")).toBeNull();
 
-  const [shadowFocusableCenterX, shadowFocusableCenterY] = await page.$eval(tabFocusTarget, (element: HTMLElement) => {
-    const focusTarget = element.shadowRoot.activeElement || element;
-    const rect = focusTarget.getBoundingClientRect();
+    if (options.focusTarget === "none") {
```
jcfranco, line 1010: This test still uses resources to set up the test when the focus target is not `none`. I think this one needs to get merged into `it("prevents focusing via keyboard and mouse")` where you have different focus assertions depending on `options.focusTarget`.

jcfranco (approved):
Awesome! Thanks for helping out with this one. 🎉

packages/calcite-components/src/tests/commonTests.ts
Elijbet: Take a look at the merge of those 2 tests now, does the combined test need a new title?

jcfranco: Looks great, thanks!

Issue #6712, opened 2023-03-31
`calciteRadioButtonGroupChange` and `calciteRadioButtonChange` events emit when selecting the checked radio button


nwhittaker:
### Actual Behavior

Given a `<calcite-radio-button-group>` of `<calcite-radio-button>` elements, clicking, or focusing it and pressing the space key, the checked radio button causes:
1. It to emit the `calciteRadioButtonChange` event.
2. The group to emit the `calciteRadioButtonGroupChange` event.

### Expected Behavior

These events are only emitted when selecting an unchecked radio button.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/yLxdyda

### Reproduction Steps

1. Visit code sample and open the console.
2. Click the **Trees** label and see both change events are logged.
3. Click the **Trees** radio button and see both change events are logged.
4. With the **Trees** radio button in focus, press the Space key and see both change events are logged.

### Reproduction Version

1.2.0

### Priority impact

p4 - not time sensitive

### Impact

Native radio inputs only emit the `change` event when the selected radio changes. As a result, emitting these events when the selected radio button hasn't actually changed is unexpected and can be wasteful if the events' handlers are expensive.

Additionally, if the handlers are not idempotent, there is risk of data corruption or the need for a workaround to add extra validation logic.

### Esri team

ArcGIS Field Apps

geospatialem:
Research needed to compare between native and Calcite functionality to determine next steps.

geospatialem:
Verified in `1.5.0-next.4`.

PR #7102, opened 2023-06-02
fix(radio-button, radio-button-group): prevent emitting events when selecting a checked radio button

driskull:
**Related Issue:** #6712

## Summary

- Prevents emitting `calciteRadioButtonChange` and `calciteRadioButtonGroupChange` events when the `checked` property is already true.
- Update tests


benelan (approved):
:+1:

Issue #7043, opened 2023-05-26
Disabled buttons are clickable in Firefox


mpayson:
### Actual Behavior

Clicking a disabled button fires the `onClick` event

### Expected Behavior

Clicking a disabled button does not fire the `onClick` event

### Reproduction Sample

https://codepen.io/mpayson/pen/VwENeyv

### Reproduction Steps

1. Open sample & developer console
2. Click button
3. See logged message

### Reproduction Version

1.4.0

### Priority impact

p2 - want for current milestone

### Impact

Maybe this could be lower priority, we can work around it by ignoring the click even when the button is disabled

### Esri team

N/A

geospatialem:
Was able to repro in FF, even with last night's `1.4.1` release, which included https://github.com/Esri/calcite-components/pull/7031.

tristan-morrison:
If it matters for prioritization purposes, this impacts the Editor widget in the JS SDK.

geospatialem:
There are two workarounds available, however it will prevent providing a `tooltip` to a disabled `button`:

```html
<calcite-button style="pointer-events:none" disabled id="example-button">Open modal</calcite-button>
```

or checking if the button is disabled in the event listener:

```js
button.addEventListener("click", (evt) => {
  if (evt.target.disabled) {
    return;
  } else {
   modal.open = true;
  }
});
```

jcfranco:
@geospatialem [Here's a codepen](https://codepen.io/jcfranco/pen/zYMqBQR?editors=1000) I used to look for regressions related to #5318. Sharing in case it helps.

geospatialem:
Verified with [Franco's codepen above](https://github.com/Esri/calcite-components/issues/7043#issuecomment-1595276148) and https://codepen.io/geospatialem/pen/rNQMPXg in `1.5.0-next.4`.

PR #7107, opened 2023-06-02
fix: ensure mouse events are blocked for disabled components in Firefox

jcfranco:
**Related Issue:** #7043

## Summary

This applies a workaround for Firefox not firing capture events before non-capture ones to block mouse events when components are disabled.

### Notable changes

* added `connectInteractive`/`disconnectInteractive` utils, solely to support Firefox
* moved user agent utils to `browser.ts`

### Additional info

* [Firefox recently fixed this](https://bugzilla.mozilla.org/show_bug.cgi?id=1731504) to follow the spec and it should ship in version 116.
* https://wpt.live/dom/events/EventTarget-dispatchEvent.html is the relevant web platform test.
* https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#useCapture 👇

  > Note: For event listeners attached to the event target, the event is in the target phase, rather than the capturing and bubbling phases. Event listeners in the capturing phase are called before event listeners in any non-capturing phases.


packages/calcite-components/src/utils/interactive.ts
```diff
@@ -104,3 +121,76 @@ export function updateHostInteraction(
 
   component.el.removeAttribute("aria-disabled");
 }
+
+function blockInteraction(component: InteractiveComponent): void {
+  component.el.click = interceptedClick;
```
driskull, line 126: Is there any chance a user could already have a click property on an el?

packages/calcite-components/src/utils/interactive.ts
```diff
@@ -24,6 +26,15 @@ type HostIsTabbablePredicate = () => boolean;
  */
 export type InteractiveHTMLElement = HTMLElement & Pick<InteractiveComponent, "disabled">;
 
+// ⚠️ browser-sniffing is not a best practice and should be avoided ⚠️
```
driskull, line 29: 👍

packages/calcite-components/src/utils/interactive.ts
```diff
@@ -104,3 +121,76 @@ export function updateHostInteraction(
...
+ *
+ * @param component
+ */
+export function connectInteractive(component: InteractiveComponent): void {
+  if (!component.disabled || !isFirefox) {
+    return;
+  }
+
+  const parent =
+    component.el.parentElement || component.el; /* assume element is host if it has no parent when connected */
```
driskull, line 176: would `getShadowRootNode()` help here?

driskull (approved):
LGTM! 👍

packages/calcite-components/src/utils/interactive.ts
jcfranco: Only if they're also hijacking `HTMLElement.click` for some magical purposes. We can revisit this if users report odd behavior.

packages/calcite-components/src/utils/interactive.ts
jcfranco: It would help for sure, but I wanted to keep it simple since I only need to check for the parent or the host. Worth noting that most of these changes will be dropped once Firefox ESR is bumped to a version matching 116+. 🎉

Issue #6554, opened 2023-03-02
[List and List Item] Provide list item sorting


geospatialem:
### Description

Add full parity to the `list` component, which is provided in the `value-list` and `pick-list` components.

May also include a refactor of the `sortable` utility, which some teams use for their own use and should be consulted if a breaking change is necessary.

### Acceptance Criteria

- [ ] Drag and drop functionality of the `list-item`s
   - [ ] Mouse and keyboard support
   - [ ] Keyboard support when screen readers are active for both Windows (NVDA, JAWS) and Mac (VoiceOver) users

### Relevant Info

Part of the https://github.com/Esri/calcite-components/issues/6430 epic

### Which Component

- List
- List item

### Example Use Case

Depicted with `value-list` and `value-list-item`: 
https://codepen.io/geospatialem/pen/abjxWwL

### Esri team

Calcite (dev)

driskull:
Before we can tackle this, we should discuss whether we should be using the SortableList component internally here or getting rid of that component.

AdelheidF:
This is what we're using currently.

![image](https://user-images.githubusercontent.com/44481292/225123512-c311740b-1bca-42a9-a546-d8cfe2d75c8c.png)![image](https://user-images.githubusercontent.com/44481292/225123569-13d9e9c9-d477-4a1e-9a0b-e657dc83e7d6.png)

driskull:
We need a whole sorting strategy before we can tackle this. The strategy should cover...

- Sorting between two separate lists or instances of components
- Fixing sorting when a child has sorting of its own (currently these are conflicting with list/block/etc)

geospatialem:
Adding a tag to this issue to Hub components and cc @vivzhang in regards to the issue description above:

> May also include a refactor of the sortable utility, which some teams use for their own use and should be consulted if a breaking change is necessary.

macandcheese:
Adding two pieces of feedback we got about Value List for consideration here w/r/t keyboard usage from this closed issue: https://github.com/Esri/calcite-components/issues/4231

> 
> Following up from Teams chat - some improvements to the value list while in drag-enabled mode to make using with a keyboard a bit easier:
> 
> Add support for esc to exit that mode (currently only hitting spacebar again exits the mode)
> Add support for home / end to quickly move an item to either extent (and preventDefault so the page doesn't scroll up / down)

Anastasiia-Boleiko:
[...]

PR #7109, opened 2023-06-03
feat(list): Add support for dragging items.

driskull:
**Related Issue:** #6554

## Summary

- SortableComponent
  - Moves all configuration options into the interface so a component can consistently setup SortableJS.
  - Changes usage of `onUpdate` to `onSort` for moving between lists and getting `calciteListOrderChange` event.
  - Adds support for `canPut`/`canPull` so users can configure whether an item can be dragged to another list and vice versa.
- List
  - Sets up sorting
    - keyboard sorting only works within a list. Cannot keyboard sort across lists at this time (including nested lists).
  - Adds dragHandle rendering to `list-item`
  - ListItemGroup and List emit an internal event when its default slot changes in order to update whether an expand caret shows or not.
- Handle
  - Updates handle to support displaying ariaLabel (logic taken from value-list)
  - Handle will emit an internal event for parent components to update an aria-live region. 
- No breaking changes necessary. We can advise users to nest another `calcite-list` to work with sorting on children.
- `calcite-list-item-group` will not be draggable/sortable at this time.


ibrahimelement:
Thanks for working on this, eager to use it once it has been merged!

driskull:
@geospatialem do you think you could test this out for a11y?

@jcfranco this is ready for review, I'm just having some issues with a test. Not sure what's going on.

geospatialem (commented):
There are a few issues with JAWS and NVDA, but this is also true in [`value-list`](https://github.com/Esri/calcite-design-system/issues/3982#issuecomment-1206487248). 

As-is JAWS can hear the context of the sorting button, but cannot move items, which would be confusing for AT users. NVDA hears the context and can move an item once, but context is lost. 

Could be a combination of keys used and the role, and what is causing the conflict when enabling AT on the component.

We should follow-up to provide keyboard accessibility via some additional efforts outside of Sortable.js, with the intent to prioritize before year's end, if possible. cc @jcfranco

packages/calcite-components/src/demos/list.html
```diff
@@ -576,6 +576,81 @@ <h1 style="margin: 0 auto; text-align: center">List</h1>
           </calcite-list>
         </div>
       </div>
+
+      <!-- simple list drag enabled -->
```
jcfranco, line 580: I think we can drop the comments for each demo sections. This info is already displayed in each section's HTML.

packages/calcite-components/src/utils/sortableComponent.ts
```diff
@@ -12,27 +18,100 @@ export interface SortableComponent {
...
 
   if (inactiveSortableComponentSet.has(component)) {
     return;
   }
 
-  component.sortable = Sortable.create(component.el, options);
+  const dataIdAttr = "id";
+  const { group, handleSelector: handle, dragSelector: draggable } = component;
+
+  component.sortable = Sortable.create(component.el, {
```
jcfranco, line 88: Can we type the callback functions in this init object?

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -245,6 +327,14 @@ export class List implements InteractiveComponent, LoadableComponent {
 
   @Element() el: HTMLCalciteListElement;
 
+  @State() assistiveText: string;
+
+  sortable: Sortable;
+
+  handleSelector = "calcite-handle";
```
jcfranco, line 334: Nitpick: can these be sorted in alphabetical order?

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -574,4 +728,70 @@ export class List implements InteractiveComponent, LoadableComponent {
...
+    }
+
+    const { enabledListItems } = this;
+
+    const sameParentItems = enabledListItems.filter((item) => item.parentElement === parentEl);
+
+    const lastIndex = sameParentItems.length - 1;
+    const startingIndex = sameParentItems.indexOf(sortItem);
+    let appendInstead = false;
+    let buddyIndex: number;
```
jcfranco, line 758: kramer-turkey-hey-buddy

packages/calcite-components/src/utils/sortableComponent.ts
```diff
@@ -12,27 +18,100 @@ export interface SortableComponent {
...
+   */
+  handleSelector: string;
+
   /**
    * The Sortable instance.
    */
   sortable: Sortable;
+
+  /**
+   * Whether the element can  move from the list.
```
jcfranco, line 47: Nitpick: extra space between `can` and `move`. 🤓

packages/calcite-components/src/utils/sortableComponent.ts
```diff
@@ -12,27 +18,100 @@ export interface SortableComponent {
...
+   */
+  dragCanPull: (event: DragEvent) => boolean;
```
jcfranco, line 49: In order to simplify the API a bit, would `canPull` and `canPut` be concise and expressive enough for these props?

jcfranco (approved):
Amazing stuff here, @driskull! 🎉

Great job on the consolidated sortable component util.

packages/calcite-components/src/utils/sortableComponent.ts
driskull: Yeah I can make that change

packages/calcite-components/src/utils/sortableComponent.ts
driskull: These are typed by Sortable already. 🎉

packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -545,7 +577,10 @@ export class ListItem
             [CSS.nestedContainerHidden]: openable && !open,
           }}
         >
-          <slot onSlotchange={this.handleDefaultSlotChange} />
+          <slot
+            onSlotchange={this.handleDefaultSlotChange}
+            ref={(el: HTMLSlotElement) => (this.defaultSlotEl = el)}
```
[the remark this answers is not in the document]
driskull: I think its only called once when the ref is established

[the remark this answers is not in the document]
eriklharper: I've just seen patterns where we declare a private method on the class for assigning the ref so that the function doesn't have to get recalled on every render, so that stood out to me.

Issue #6633, opened 2023-03-22
calcite-list - changing filter-text property/attribute does not filter the list items


jgrayson-apl:
### Actual Behavior

The list of calcite-list-items is not updated/changed/filtered when changing the filter-text property/attribute of the calcite-list via code.

### Expected Behavior

I would expect the list of calcite-list-items to be updated/changed/filtered when changing the filter-text property/attribute of the calcite-list via code.

### Reproduction Sample

https://codepen.io/john-grayson/pen/ZEMMqdX

### Reproduction Steps

1. click on the 'FILTER BANANAS' button
2. see how the list is not filtered

### Reproduction Version

1.07

### Priority impact

p4 - not time sensitive

### Esri team

N/A

jgrayson-apl:
Q: Any current workarounds to this issue?  

I just need to change the filter via code... we should be able to do this somehow, no?

jcfranco:
@jgrayson-apl Not ideal by any means, but could you give this a try? https://codepen.io/jcfranco/pen/PoyzwKx?editors=0010

jgrayson-apl:
@jcfranco yes kinda hacky but ok for a workaround, I was able to adapt the concepts to a couple of apps.  Thank you very much.

driskull:
I think this needs to be fixed in the filter component.

The filter component doesn't emit an event when the filterText changes programmatically. The issue is that when filterText is changed programmatically the filteredItems property changes after a debounce so there's no way to know when to grab that new data. The filter would either need to emit or have another event for when the filteredItems changes.

driskull:
@jcfranco what are your thoughts regarding how filter can let other components know when the filterText changes?

driskull:
@jcfranco I'd like to get this going. What do you think about...

- Making a public filter method that would resolve once debouncing is complete.
- Changing the `calciteFilterChange` event to emit on both programatic and user changes.
- Removing debouncing from the filter and making other components handle debouncing

jgrayson-apl:
Since you guys are thinking about this issue...  I would like to have the ability to set/override a 'filter' function/callback (like we can do with simple arrays) where I could perform my own more complex checks to determine if an item passes the test.  There are many times when entering text in the input is not what we need.  Sometimes we have multiple criteria to consider or even a simple discrete set of options that we want to filter on.  Probably beyond what is going on here, but thought I'd mention anyways.

jcfranco:
[...]

PR #7127, opened 2023-06-05
feat(filter): Add filter method

driskull:
**Related Issue:** #6633

## Summary

- Adds public filter method to the calcite-filter component.
- Fixes setting `filteredItems` on filter.
- Adds test


packages/calcite-components/src/components/filter/filter.e2e.ts
```diff
@@ -273,4 +273,50 @@ describe("calcite-filter", () => {
...
+            name: "Harry",
+            description: "developer",
+            value: "harry",
+            metadata: { haircolor: "red", favoriteBand: "MetallicA" }
+          },
+          {
+            name: "Matt",
+            description: "developer",
+            value: "matt",
+            metadata: { haircolor: "black", favoriteBand: "Radiohead" }
```
benelan, line 296: Can we please trade Matt for whoever Harry is? Metallica-loving gingers ftw

packages/calcite-components/src/components/filter/filter.tsx
```diff
@@ -219,16 +235,16 @@ export class Filter
...
-    this.filter("", true);
+    this.filterDebounced("", true);
     this.setFocus();
   };
 
-  updateFiltered(filtered: any[], emit = false): void {
-    this.filteredItems.length = 0;
-    this.filteredItems = this.filteredItems.concat(filtered);
+  updateFiltered(filtered: object[], emit = false, callback?: () => void): void {
+    this.filteredItems = [...filtered];
```
benelan, line 243: is a shallow copy good enough here, considering `filteredItems` are objects?

benelan (approved):
Do we need to do a deep copy of the items in `updateFiltered`? I don't think we are allowed to use [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) yet but there is always the JSON stringify/parse hack lol

If a shallow copy is fine then LGTM!

packages/calcite-components/src/components/filter/filter.tsx
driskull: Thats a good question. It was a shallow copy previously right?

driskull: I think a shallow copy is fine here since we weren't cloning before.

benelan: Yeah I looked at the code and we definitely don't need a deep copy and probably don't even need a shallow one. `filteredItems` is readonly and our [filter util](https://github.com/esri/calcite-components/blob/dris0000/filter-method/packages/calcite-components/src/utils/filter.ts) uses [`Array.prototype.filter()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter)  which already creates a shallow copy. So we can do:

```ts
updateFiltered(filtered: object[], emit = false, callback?: () => void): void {
    this.filteredItems = filtered;
```

packages/calcite-components/src/components/filter/filter.tsx
```diff
@@ -181,6 +182,20 @@ export class Filter
   //
   // --------------------------------------------------------------------------
 
+  /**
+   * Sets focus on the component.
```
jcfranco, line 186: Doc description does not match function.

On a related note, I think we should add a little snippet regarding the need for this method since filtering is delayed and asynchronous.

packages/calcite-components/src/components/filter/filter.tsx
```diff
@@ -169,6 +169,7 @@ export class Filter
   disconnectedCallback(): void {
     disconnectLocalized(this);
     disconnectMessages(this);
+    this.filterDebounced.cancel();
```
jcfranco, line 172: Nice! 

Would you need to filter again on `connectedCallback`? Might be an edge case, but I'm thinking of the case where filtering is running as the component is moved between parents.

packages/calcite-components/src/components/filter/filter.tsx
```diff
@@ -219,16 +235,16 @@ export class Filter
...
     this.value = "";
-    this.filter("", true);
+    this.filterDebounced("", true);
     this.setFocus();
   };
 
-  updateFiltered(filtered: any[], emit = false): void {
-    this.filteredItems.length = 0;
-    this.filteredItems = this.filteredItems.concat(filtered);
+  updateFiltered(filtered: object[], emit = false, callback?: () => void): void {
```
jcfranco, line 242: WDYT about `onFilter` or something along those lines for the callback name? Up to you.

jcfranco (approved):
Awesome! 🎉

This should be good to merge once the June (patch-only) release is done.

packages/calcite-components/src/components/filter/filter.tsx
driskull: Maybe we could address that if it ever comes up? Since we don't know if a component is moved vs removed we would have to maintain state like this that isn't set via a prop. It would be something to discuss across all components that have some async method or state to maintain.

jcfranco: For sure, we can keep an eye out for it. I think this is a very edge use case anyways since the filter is a supporting component and it'd be odd for a component to move it around during filtering. 🤔

PR #7134, opened 2023-06-06
chore(t9n): move manifest file to root level.

anveshmekala:
**Related Issue:** #

## Summary

This PR will move the `t9nmanifest.txt` file from `packages/calcite-components/support` to the root directory as requested by the t9n team and updates the paths in `t9nmanifest.txt`


jcfranco (approved):
Shouldn't the `t9nmanifest.txt` paths be updated (via the script) as well?

Once this is addressed, this should be good to merge!

Issue #6600, opened 2023-03-14
calcite-list missing slot for menu-action like calcite-pick-list


AdelheidF:
### Actual Behavior

In order to move from pick-list to list we need a way to add a menu-action like shown below.

![image](https://user-images.githubusercontent.com/44481292/225129562-cd7c2ea2-1801-4a08-859c-5422a4d41299.png)![image](https://user-images.githubusercontent.com/44481292/225129638-9651b83b-2951-4b4d-a0e0-74a6c10ac737.png)

### Expected Behavior

menu-action slot like pick-list

### Reproduction Sample

https://codepen.io/afreitag/pen/WNgzWee

### Reproduction Steps

Open codepen and compare the 2 lists.

### Reproduction Version

1.0.8

### Priority impact

p3 - want for upcoming milstone

### Impact

Needs to happen before pick-list gets removed.

### Esri team

ArcGIS Map Viewer

geospatialem:
Will consider a `filter-menu-options` slot for the `list` component when a filter is present on the component.

driskull:
@jcfranco should this slot be built into the filter component? The current slot name "menu-actions" doesn't really work in the context of the list but could work in the context of the filter. We would need a better slot name for the list.

geospatialem:
Verified in `1.5.0-next.9` with the following [Codepen](https://codepen.io/geospatialem/pen/oNQWEWE).

![image](https://github.com/Esri/calcite-components/assets/5023024/0d840ee1-21bd-4ff5-a568-99f6f935b1ab)

PR #7148, opened 2023-06-06
feat(list): Add slots for filter actions

driskull:
**Related Issue:** #6600

## Summary

- adds slot `filter-actions-start`.
- adds slot `filter-actions-end`.
- Uses `calcite-stack` internally for layout.
- Adds story
- Updates demo file


benelan (approved):
:+1:

jcfranco:
We may need to revert and install this later as it was installed into a patch-only milestone.

driskull:
ah sorry. I'll revert.

Issue #7113, opened 2023-06-03
radio-button : When tab into radio-group no radio-button is focused if the first radio-button is disabled.


anveshmekala:
### Actual Behavior

When the first `calcite-radio-button` element in the group is` disabled`  and user tries to `Tab` into the group then the next focusable `calcite-radio-button` element is not focused.

### Expected Behavior

When the user `Tab` into a radio-group, first focusable radio-button should be focused.

### Reproduction Sample

https://codepen.io/ANVESHMEKALA/pen/NWOQzxp?editors=1000

### Reproduction Steps

1. Open the codepen
2.  Try `Tab` into the `calcite-radio-group`.
3. Observe none of the `calcite-radio-button` is focused.

### Reproduction Version

1.4.1

### Priority impact

p3 - want for upcoming milestone

### Esri team

Calcite (dev)

PR #7152, opened 2023-06-07
fix(radio-button):  focuses first focusable radio-button element in group.

anveshmekala:
**Related Issue:** #7113 

## Summary

This PR will focus the first focusable `calcite-radio-button` when the user `Tab` in to the group.


packages/calcite-components/src/components/radio-button/radio-button.tsx
```diff
@@ -271,11 +274,26 @@ export class RadioButton
...
+    });
+  }
+
   private getTabIndex(): number | undefined {
     if (this.disabled) {
       return undefined;
     }
-    return this.checked || this.isDefaultSelectable() ? 0 : -1;
+    if (this.checked || this.isFocusable()) {
+      this.updateTabIndexOfOtherRadioButtonsInGroup();
```
eriklharper, line 292: I'm wondering if `updateTabIndexOfOtherRadioButtonsInGroup` should be called in `connectedCallback` instead of on every render?  Seems better from a performance standpoint to only run this when we know the actual DOM has been modified.

eriklharper (changes requested):
Could you try updating other radio-buttons in a group in `connectedCallback` instead?  If you can get that to work, let's also add a test for this that ensures the focus behavior is correct when adding or removing buttons.

packages/calcite-components/src/components/radio-button/radio-button.tsx
anveshmekala: i don't think moving this to `connectedCallback` is required. This method executed only once per radio-group on initial render because only one radio-button will have the `tabindex` set to `0`. Any new addition or deletion of radio-buttons in the group causing change in `tabIndex` will invoke this method which is likely to happen once per deletion or addition .

eriklharper: > This method executed only once per radio-group on initial render

I'm seeing this method getting called on initial load and anytime a new value is selected, so its not just happening on initial render:

image

eriklharper: Seems to work just fine if I simply move the call to `this.updateTabIndexOfOtherRadioButtonsInGroup()` into `connectedCallback`.  I now only see those forceUpdate calls happening when I add or remove radios.

anveshmekala: This works as expected and enhances the  performance.

packages/calcite-components/src/components/radio-button/radio-button.tsx
```diff
@@ -286,11 +289,25 @@ export class RadioButton
...
 
+  private updateTabIndexOfOtherRadioButtonsInGroup(): void {
+    const radioButtons = this.queryButtons();
+    const otherFocusableRadioButtons = radioButtons.filter(
+      (radioButton) => radioButton.guid !== this.guid && !radioButton.disabled
+    );
+    otherFocusableRadioButtons.forEach((radiobutton) => {
+      forceUpdate(radiobutton);
+    });
+  }
```
eriklharper, line 300: I tested completely removing `updateTabIndexOfOtherRadioButtonsInGroup` and this works great with just the changes you made to `isFocusable`.  I tested with multiple radio button groups and tabbing between groups works correctly.  Try completely removing the call to this function, preserving the rest of the changes and see if that works for you too.

eriklharper: Scratch that, we need this for adding/removing radios.

packages/calcite-components/src/components/radio-button/radio-button.tsx
```diff
@@ -291,11 +304,26 @@ export class RadioButton
     });
   }
 
+  private updateTabIndexOfOtherRadioButtonsInGroup(): void {
+    console.log("update tab index", this.el.value);
```
eriklharper, line 308: Console log here to remove.

packages/calcite-components/src/components/radio-button/radio-button.tsx
```diff
@@ -478,6 +508,7 @@ export class RadioButton
 
   render(): VNode {
     const tabIndex = this.getTabIndex();
+    console.log("re render", this.el.value);
```
eriklharper, line 511: We should remove this one as well.

eriklharper (approved):
This looks excellent, thanks for adding those extra tests!  Just remove those console logs and we should be good to go :)

packages/calcite-components/src/components/radio-button/radio-button.tsx
```diff
@@ -94,6 +100,11 @@ export class RadioButton
   /** When `true`, the component is not displayed and is not focusable or checkable. */
   @Prop({ reflect: true }) hidden = false;
```
jcfranco, line 101: Sidebar: we should revisit custom `hidden` props since it's a reflected prop on `HTMLElement`.

packages/calcite-components/src/components/radio-button/radio-button.tsx
```diff
@@ -184,9 +195,11 @@ export class RadioButton
     ) as HTMLCalciteRadioButtonElement[];
   };
 
-  isDefaultSelectable = (): boolean => {
+  isFocusable = (): boolean => {
     const radioButtons = this.queryButtons();
-    return !radioButtons.some((radioButton) => radioButton.checked) && radioButtons[0] === this.el;
+    const firstFocusable = radioButtons.find((radiobutton) => !radiobutton.disabled);
```
jcfranco, line 200: `radiobutton` ➡️ `radioButton`

packages/calcite-components/src/components/radio-button/radio-button.tsx
```diff
@@ -291,11 +304,25 @@ export class RadioButton
...
+      forceUpdate(radiobutton);
+    });
+  }
+
   private getTabIndex(): number | undefined {
     if (this.disabled) {
       return undefined;
     }
-    return this.checked || this.isDefaultSelectable() ? 0 : -1;
+    if (this.checked || this.isFocusable()) {
```
jcfranco, line 321: No love for the previous ternary? 😭

jcfranco (approved):
⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪
⚪🔘⚪⚪⚪🔘⚪🔘🔘🔘⚪⚪🔘🔘🔘⚪🔘🔘🔘🔘⚪🔘⚪
⚪🔘🔘⚪⚪🔘⚪⚪🔘⚪⚪🔘⚪⚪⚪⚪🔘⚪⚪⚪⚪🔘⚪
⚪🔘⚪🔘⚪🔘⚪⚪🔘⚪⚪🔘⚪⚪⚪⚪🔘🔘🔘⚪⚪🔘⚪
⚪🔘⚪⚪🔘🔘⚪⚪🔘⚪⚪🔘⚪⚪⚪⚪🔘⚪⚪⚪⚪⚪⚪
⚪🔘⚪⚪⚪🔘⚪🔘🔘🔘⚪⚪🔘🔘🔘⚪🔘🔘🔘🔘⚪🔘⚪
⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪⚪

packages/calcite-components/src/components/radio-button/radio-button.tsx
anveshmekala: Oops, there was some extra code in `if` block earlier. Will revert to ternary.

PR #6988, opened 2023-05-23
ci: transition to monorepo

benelan:
**Related Issue:** #6532

## Summary

Transitions to a monorepo using Turbo Repo. The existing files are moved to `packages/calcite-components` and `calcite-components-react` is added to the repo. For more information see: 
https://github.com/Esri/calcite-components/wiki/Monorepo

PR #7153, opened 2023-06-07
ci(release-please): change PR title and ensure git tag consistency

benelan:
**Related Issue:** #6988

## Summary

This PR changes the `release-please` PR title to `chore: release latest`. The default is `chore: release master`. 

It also makes sure a `v` isn't added before the version in git tags. Their schema doc says it doesn't by default, 
but I'm not convinced (see links below) so I want to hardcode the config value. The `v` being in the tag wouldn't break anything in our CI, but Lerna doesn't add a `v` and we should make sure our tag format is consistent for our consumer's CI.

https://github.com/googleapis/release-please/blob/bdd9b0158ce79d958da01ebf54cb6b07925bc125/schemas/config.json#L77-L79

https://github.com/googleapis/release-please/blob/bdd9b0158ce79d958da01ebf54cb6b07925bc125/src/strategies/base.ts#L139


driskull (approved):
👍

Issue #6378, opened 2023-01-30
[calcite-input-date-picker]: should display placeholder text when the value is reset to null


vannizhang:
### Actual Behavior
I have two simple simple components, a `calcite-input-date-picker` and a `calcite-button`:
![cal1](https://user-images.githubusercontent.com/3362490/218831044-a97d2fcd-060f-41eb-811f-c31ce3cad851.png)

I can use the date picker to select a date without any issue:
![cal2](https://user-images.githubusercontent.com/3362490/218831078-0652cb5c-1c12-4118-9719-47f1d638d0d6.png)

The selected date should be reset when I click on the reset button, and I can see selected date become unselected in the calendar component after clicking the button:
![cal3](https://user-images.githubusercontent.com/3362490/218831104-7053b779-ce58-4be9-a7af-dc2e96bd493f.png)

But the input still shows previously selected date even after we reset the selected date:
![cal4](https://user-images.githubusercontent.com/3362490/218831617-c8e6f433-a2b0-47a9-a16a-7490f9356301.png)

### Expected Behavior
the input should show placeholder text (e.g. YYYY/MM/DD) once the selected date is reset to undefined, so that the user knows that they have successfully reset the selected date.

### Reproduction Sample
https://codepen.io/vannizhang/pen/ZEjMLaL?editors=1010

### Reproduction Steps
open the codepen sample,
select a date using the date-picker component and you should see the selected date in input-date-picker
click on reset selected date button
you still see the previously selected date in the input-date-picker, which is incorrect, should see the placeholder text
#### Reproduction Version
1.0.3

### Relevant Info
No response

### Regression?
No response

### Impact
The Living Atlas Curation App (a internal tool to curate all Living Atlas contents) uses the calcite-input-date-picker in a Modal to let the curator to select a auto removal date for any Living Atlas content, we also gives them a button to reset the previously selected auto removal date, when that button is clicked, we want them to see that change in the calcite-input-date-picker:

![cal5](https://user-images.githubusercontent.com/3362490/218831215-5498e826-46b8-4964-871e-5a0b5b616fd2.png)

### Esri team
Calcite (dev)

geospatialem:
Also surfaced in a [Community post](https://community.esri.com/t5/calcite-design-system-questions/reset-input-date-picker-component/m-p/1297746#M423) on June 9.

PR #7156, opened 2023-06-08
fix(input-date-picker): Fix showing the placeholder when resetting the value

driskull:
**Related Issue:** #6378

## Summary

- Updates internal input's value when the value on the input-date-picker is changed.
- Adds a test

### DIFF
--- a/packages/calcite-components/src/components/input-date-picker/input-date-picker.e2e.ts
+++ b/packages/calcite-components/src/components/input-date-picker/input-date-picker.e2e.ts
@@ -805,4 +805,24 @@ describe("calcite-input-date-picker", () => {
       expect(await getFocusedElementProp(page, "id")).toBe("next-sibling");
     });
   });
+
+  it("should reset input value", async () => {
+    const page = await newE2EPage();
+    const expectedValue = "2022-10-01";
+    const expectedInputValue = "10/1/2022";
+
+    await page.setContent(html` <calcite-input-date-picker value="${expectedValue}"></calcite-input-date-picker>`);
+
+    const inputDatePickerEl = await page.find("calcite-input-date-picker");
+    const input = await page.find("calcite-input-date-picker >>> calcite-input");
+
+    expect(await inputDatePickerEl.getProperty("value")).toEqual(expectedValue);
+    expect(await input.getProperty("value")).toEqual(expectedInputValue);
+
+    inputDatePickerEl.setProperty("value", "");
+    await page.waitForChanges();
+
+    expect(await inputDatePickerEl.getProperty("value")).toEqual("");
+    expect(await input.getProperty("value")).toEqual("");
+  });
 });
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/input-date-picker/input-date-picker.e2e.ts
```diff
@@ -805,4 +805,24 @@ describe("calcite-input-date-picker", () => {
...
+    const expectedValue = "2022-10-01";
+    const expectedInputValue = "10/1/2022";
+
+    await page.setContent(html` <calcite-input-date-picker value="${expectedValue}"></calcite-input-date-picker>`);
+
+    const inputDatePickerEl = await page.find("calcite-input-date-picker");
+    const input = await page.find("calcite-input-date-picker >>> calcite-input");
+
+    expect(await inputDatePickerEl.getProperty("value")).toEqual(expectedValue);
+    expect(await input.getProperty("value")).toEqual(expectedInputValue);
```
eriklharper, line 820: nitpick: I feel like what we want to assert here is the actual displayed value in the input instead of `calcite-input`'s `value` property, so I've used code like this to do just that:

```
const getInputValue = async () =>
        await page.evaluate(
          () =>
            document
              .querySelector("calcite-input-date-picker")
              .shadowRoot.querySelector("calcite-input")
              .shadowRoot.querySelector("input").value
        );

      expect(await getInputValue()).toBe(expectedInputValue);
```
We could probably create this as a shared util at some point.

packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -158,21 +158,31 @@ export class InputDatePicker
...
         newValueAsDate = undefined;
       }
 
+      const { focusedInput } = this;
+
+      this.setInputValue(
+        valueIsArray ? newValue[focusedInput === "end" ? 1 : 0] : newValue,
+        focusedInput
+      );
+
```
eriklharper, line 178: This looks good, the only thing I'm wondering is if we should move this logic over to the `localizeInputValues()` method?  I'm realizing that method only handles localizing truthy input values, but it could just as well handle empty values as well.  Up to you, no hard opinion either way.

eriklharper (approved):

packages/calcite-components/src/components/input-date-picker/input-date-picker.e2e.ts
driskull: That would be testing the internals of the `calcite-input` and not the internals of just the `calcite-input-date-picker`. I think its out of scope to test the internals of the input here.

eriklharper: Assuming the `this.value` of `calcite-input` 100% matches the input value, which I believe in the case of `calcite-input type="text"` would apply, then testing `calcite-input.value` should suffice.  I agree with you that its testing the internals of `calcite-input` too, and probably worth writing a few tests that capture different localized string values within `calcite-input.e2e.ts` as well.  Definitely out of scope for this PR, but can be a separate nice-to-have test enhancement.

eriklharper (approved):
Looks amazing!  Such a smaller diff too :)

Issue #6363, opened 2023-01-27
alert: auto-close-duration required for custom-elements but not dist


mpayson:
### Actual Behavior
When using the dist build, an auto-closing calcite-alert will work when just the auto-close attribute is set. 

When using the custom elements build, `calcite-alert` needs both the` auto-close` and `auto-close-duration` attributes to be set, otherwise the `alert` will immediately dismiss and not display to the user.

### Expected Behavior
`auto-close-duration` is not required (or is required) for both distributions and this is documented

### Reproduction Sample
~https://codepen.io/mpayson/pen/zYLaELw~
https://codesandbox.io/s/cc-bug-calcitealert-autoclose-autocloseduration-9ht5xd?file=/src/App.js

### Reproduction Steps
The codepen uses the dist build, and it looks like the `auto-close-duration` property is automatically set on the element, despite it not being set in the markup

1. [Open the sample](https://codesandbox.io/s/cc-bug-calcitealert-autoclose-autocloseduration-9ht5xd?file=/src/App.js)
2. Observe `alert` doesn't open on load
3. Add `autoCloseDuration={"medium"}` to `alert`
4. Observe `alert` now loads with `autoClose`

<img width="1680" alt="1__215131605-4b33d2c2-9081-4b08-b89f-3d37f5b64162" src="https://user-images.githubusercontent.com/3362490/218834855-179def1e-81d3-4b8f-9df2-b7fd8b9237cb.png">
-27 at 10 58 23 AM

In a project that uses the custom-elements build, I used the same markup with different results

 ```javascript
return (
    <calcite-alert
      icon="rangefinder"
      kind="brand"
      open
      label="A report alert"
      auto-close
    >
      <div slot="title">Trail Camera Report </div>
      <div slot="message">We thought you might want to take a look</div>
      <calcite-link slot="link">Take action</calcite-link>
    </calcite-alert>
  );
```
<img width="519" alt="2__215131567-292f55c9-7741-43f1-8796-23258cbeb09c" src="https://user-images.githubusercontent.com/3362490/218834898-ea7f70e3-f3ff-415f-afd8-67d1c4427251.png">

### Reproduction Version
1.0.3 (repro is in 1.2.0)

### Relevant Info
custom-elements build

### Regression?
No response

### Impact
Seems low with the workaround, but should be mitigated as its not a great bug.

### Esri team
ArcGIS Online

macandcheese:
[...]

PR #7157, opened 2023-06-08
fix(alert): Sets autoCloseDuration to "medium" by default

driskull:
**Related Issue:** #6363

## Summary

fix(alert): Sets autoCloseDuration to "medium" by default. #6363

@jcfranco ~~can we have linting to prevent having any conditional defaults? If not, we should maybe update conventions to say not to do this~~ Lets discuss how we can properly document and test this pattern.

### DIFF
--- a/packages/calcite-components/conventions/README.md
+++ b/packages/calcite-components/conventions/README.md
@@ -153,6 +153,22 @@ It is recommended to reflect properties that fit the following criteria:
 
 Doing so will give developers more flexibility when querying the DOM. This is important in framework environments where we can't safely assume components will have their attributes set vs properties.
 
+### Property defaults
+
+Properties should have static default values. There should be no conditional logic when defining the default value of a property.
+
+#### Incorrect example
+
+```ts
+@Prop({ reflect: true }) autoCloseDuration: AlertDuration = this.autoClose ? "medium" : null;
+```
+
+#### Correct example
+
+```ts
+@Prop({ reflect: true }) autoCloseDuration: AlertDuration = "medium";
+```
+
 ### `ref` usage
 
 Due to a [bug in Stencil](https://github.com/ionic-team/stencil/issues/4074), `ref` should be set as the last property in JSX to ensure the node's attributes/properties are up to date.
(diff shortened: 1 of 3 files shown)


benelan (approved):
I didn't find a rule for that after a quick search, were you thinking we should add one to our custom calcite eslint plugin?

driskull:
Yeah that's what I was thinking if its possible.

packages/calcite-components/conventions/README.md
```diff
@@ -153,6 +153,22 @@ It is recommended to reflect properties that fit the following criteria:
 
 Doing so will give developers more flexibility when querying the DOM. This is important in framework environments where we can't safely assume components will have their attributes set vs properties.
 
+### Property defaults
```
jcfranco, line 156: In order to not block the fix, can we discuss this separately before making it a convention? There might be some other cases to consider (e.g., [the color-picker ensures the format matches and set default value to hex or hexa depending on alpha being enabled or not](https://github.com/Esri/calcite-components/blob/master/packages/calcite-components/src/components/color-picker/color-picker.tsx#L219-L221)).

jcfranco (approved):
Aside from discussing the proposed convention separately, this LGTM!

packages/calcite-components/conventions/README.md
driskull: Yeah lets discuss it. 

As a user, its nice to know what the expected default value for a prop will be. Having these conditional default values does complicate things so we should discuss whether it matters or not.

We may be able to make rules for how these conditionals should be documented since there are other components like this. (such as datepicker `value` where it depends on if the component is a range or not)

We could at least update the doc for colorPicker to document the `default` since it only lists the hex right now. We may also want to enhance our `defaults` test to be able to handle these conditionals.

driskull: @jcfranco with the colorPicker, if `alphaChannel` is set to `true` after initialization, the value doesn't get updated right? Would a user expect the value to be updated? We should discuss that part as well.

driskull: Maybe we can make exceptions for props that are mutable? In most cases `value` props are mutable and can be modified by the user and the component.

Issue #7039, opened 2023-05-26
Calcite numeric input deleting decimal point


nid13240:
### Actual Behavior

When I am editing a decimal value on the calcite input component, the decimal point itself gets deleted when I delete the number in the tenths position. For example, if I have 1.5 and want to change it to 1.8, on trying to delete 5, it would also remove the "." and I would end up typing 18. 

### Expected Behavior

I expect the decimal position to stay in the input box like it does in the native input element.

### Reproduction Sample

https://codepen.io/geospatialem/pen/poxYOdO

### Reproduction Steps

Using the codepen above, try to use backspace on the three different input boxes and observe the difference in behavior between calcite-input and native input.

### Reproduction Version

1.4.0

### Priority impact

p3 - want for upcoming milestone

### Esri team

ArcGIS Map Viewer

geospatialem:
Verified on `master`

PR #7159, opened 2023-06-08
fix(input, input-number): no longer removes trailing decimal separator

anveshmekala:
**Related Issue:** #7039 

## Summary

This PR will allow trailing decimal separator while user is editing the input value in `calcite-input` & `calcite-input-number`

Before the Change:

- If the user removes the `1` value from `0.00001`  , the input will be sanitized and the new value will be `0`
- If the user removes the `1` value from `0.1`  , the input will be sanitized and the new value will be `0`

After the Change:

- If the user removes the `1` value from `0.01`  , the input value will be `0.0`
- If the user removes the `1` value from `0.1`  , the input value will be `0.`


packages/calcite-components/src/components/input-number/input-number.e2e.ts
```diff
@@ -715,6 +715,21 @@ describe("calcite-input-number", () => {
...
+      const calciteInput = await page.find("calcite-input");
+      const inputEventSpy = await calciteInput.spyOnEvent("calciteInputInput");
+
+      await calciteInput.click();
+      await page.keyboard.press("Backspace");
+      await page.waitForChanges();
+      expect(inputEventSpy).toHaveReceivedEvent();
+      const input = await page.find("calcite-input >>> input");
+
+      expect(await input.getProperty("value")).toBe("1.");
```
benelan, line 731: the component's value should still be `1`. You will need to go into the shadowRoot and check the value of the internal input to verify it is `1.`

benelan: Nevermind that is what you were doing it was other input tests failing. I need read closer before opening my big mouth lol

benelan (commented):
Try adding default decimal/group/minus values for `en` to NumberStringFormat. 

https://github.com/esri/calcite-components/blob/dd1dfbdd2dc89ad5b2ea8be2f7adf70086217220/packages/calcite-components/src/utils/locale.ts#L322-L347

```jsx
export class NumberStringFormat {
  /**
   * The actual group separator for the specified locale.
   * Some white space group separators don't render correctly in the browser,
   * so we replace them with a normal <SPACE>.
   */
  private _actualGroup = ",";

  /** the corrected group separator */
  private _group = ",";

  get group(): string {
    return this._group;
  }

  private _decimal = ".";

  get decimal(): string {
    return this._decimal;
  }

  private _minusSign = "-";

  get minusSign(): string {
    return this._minusSign;
  }
```

packages/calcite-components/src/components/input-number/input-number.e2e.ts
anveshmekala: no problem. The issues with other tests is the immutation of `localizedValue` when the user tries to delete the decimal part.  The logic applies even if the user is typing out a value like `0.05`. By the time user types `0.0` the validation is happening hence changing the localizedValue to `0` again. Current setup avoids this by not changing the `localizedValue` variable until the user inputs a valid value which allows users to type the value of their without any auto correction.  

Trying to figure out a reliable way to identify if the user is adding a value in the input field or deleting the value so that we can apply the logic only for deletion case.

benelan: Gotcha. 

> Trying to figure out a reliable way to identify if the user is adding a value in the input field or deleting the value so that we can apply the logic only for deletion case.

Some potential options for investigation:

1. only sanitzing the trailing decimal if the length of the previous value is more than the current one
2. only sanitzing the trailing decimal on backspace and delete key events. And the increment down click event too probably

anveshmekala: yeah i was using the second option which is not feasible with `inputEvent` . i had to create a combination of above. Realized there was an another bug with the input when the user has leading `0`'s in the decimal. If the user deletes the `2` in `0.00002` then the value becomes `0` which is similar to #7039 . I tried adding those trailing zero's back in the decimal but the localize method is removing them too. We may have to follow similar technique of adding back those zeros like we did for decimal separator.

anveshmekala: or we can refactor [this](https://github.com/Esri/calcite-components/blob/master/packages/calcite-components/src/utils/number.ts#L47) method.

anveshmekala:
Open for suggestions to avoid adding trailing decimal zero values during sanitization and also localization of the value.

packages/calcite-components/src/components/input-number/input-number.e2e.ts
benelan: Yeah the input lifecycle methods will probably need a refactor. Once you start doing crazy stuff like this to fix bugs you know you've gone too far lol.

packages/calcite-components/src/utils/number.ts
```diff
@@ -217,3 +230,33 @@ export function expandExponentialNumberString(numberString: string): string {
...
+  const decimals = value.split(".")[1];
+
+  if (decimals && localizedDecimals !== decimals) {
+    localizedValue = localizedValue + decimalSeparator;
+    [...decimals].forEach((decimal) => {
+      localizedValue += formatter.localize(decimal);
+    });
+  }
+  return localizedValue;
+}
```
benelan, line 262: This function is a bit confusing to me. Is it only for adding trailing 0s? If so, then why is it iterating over all of the initial value's decimal places and concat'ing them to the localizedValue? For example

value is `123.3210`
localizedValue is `123,321`

would this function return `123,321,3210`?

Also, can you use template literals instead of concatenating and add some unit tests for this function.

benelan (commented):
A couple comments, but looking good nice work! I'll do some testing on my end too because changing the input's lifecycle methods can have some weird side effects.

 I think eventually we will need a larger refactor because doing these case by case patches isn't sustainable.

packages/calcite-components/src/utils/number.ts
```diff
@@ -137,14 +137,27 @@ export const sanitizeNumberString = (numberString: string): string =>
...
 
+export function getBigDecimalAsString(sanitizedValue: string): string {
+  const decimals = sanitizedValue.split(".")[1];
+  const newdecimals = new BigDecimal(sanitizedValue).toString().split(".")[1];
+
+  // adds back trailing decimal zeros
+  if (decimals && !newdecimals && BigInt(decimals) === BigInt(0) && decimals !== newdecimals) {
+    const value = new BigDecimal(sanitizedValue).toString() + ".";
+    const newvalue = value.padEnd(value.length + decimals.length, "0");
+    return newvalue;
```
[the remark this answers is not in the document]
benelan: I've been more conscious about overworking our garbage collectors after seeing the results of their strike in France.

packages/calcite-components/src/utils/number.ts
benelan: Example of changing the concat'ing to template literals is this:

```ts
 if (decimals && localizedDecimals !== decimals) {
    localizedValue = localizedValue + decimalSeparator;
    [...decimals].forEach((decimal) => {
      localizedValue += formatter.localize(decimal);
    });
  }
  return localizedValue;
```
to something like:
```ts
return decimals && localizedDecimals !== decimals
    ? `${localizedValue}${decimalSeparator}${Array.from(decimals)
        .map((d) => formatter.localize(d))
        .join("")}`
    : localizedValue;
``````

Comment (anveshmekala) [reply packages/calcite-components/src/utils/number.ts:262]: yes, it is supposed to add only trailing zeros since those are the ones that are trimmed off during localization and sanitization.  Great catch and i agree that iteration can be avoided and didn't factored in numbering systems like french :frowning_face: . Will refactor this one and add a test as well.

Comment (benelan) [reply packages/calcite-components/src/utils/number.ts:274]: To clarify, the template literals vs string concatenation thing is about weirdness with RTL langauges. Other than that the two snippets above are pretty much the same. I just use `map()` instead of `forEach` in that specific example.

File: packages/calcite-components/src/components/input-number/input-number.tsx [hunk 4]
```diff
@@ -840,12 +841,19 @@ export class InputNumber
...
 
-    const sanitizedValue = sanitizeNumberString(
-      // no need to delocalize a string that ia already in latn numerals
-      (this.numberingSystem && this.numberingSystem !== "latn") || defaultNumberingSystem !== "latn"
-        ? numberStringFormatter.delocalize(value)
-        : value
-    );
+    const isValueDeleted =
+      this.previousValue?.length > value.length || this.value?.length > value.length;
+    const hasTrailingDecimalSeparator = value.charAt(value.length - 1) === ".";
```
benelan, line 846: I think the `.` should be `numberFormatter.decimal` here, since it is being done before delocalization

packages/calcite-components/src/components/input-number/input-number.tsx
```diff
@@ -840,12 +841,19 @@ export class InputNumber
...
+    const sanitizedValue =
+      hasTrailingDecimalSeparator && isValueDeleted
+        ? value
```
benelan, line 849: Is there a reason why the value doesn't need to be delocalized when there is a trailing 0 and it is a delete action?

packages/calcite-components/src/utils/number.ts
```diff
@@ -217,3 +226,30 @@ export function expandExponentialNumberString(numberString: string): string {
...
+  formatter: NumberStringFormat
+): string {
+  const decimalSeparator = formatter.decimal;
+  const localizedIntegers = localizedValue.includes(decimalSeparator)
+    ? localizedValue.split(decimalSeparator)[0]
+    : localizedValue;
+  const decimals = value.split(".")[1];
+
+  return decimals
+    ? `${localizedIntegers}${decimalSeparator}${Array.from(decimals)
```
benelan, line 251: It looks like the decimal separator is being added here as well as in input - 

```js
this.localizedValue =
        hasTrailingDecimalSeparator && isValueDeleted
          ? `${newLocalizedValue}${numberStringFormatter.decimal}`
```

Wouldn't that result in too many decimals?

packages/calcite-components/src/utils/number.ts
```diff
@@ -217,3 +226,30 @@ export function expandExponentialNumberString(numberString: string): string {
...
+        .map((d) => formatter.localize(d))
```
benelan, line 252: I'm still not sure this is what we want, could it potentially add non 0 decimal places? Could you add some spec tests including numbers like `123.456000`. If we are iterating over all of the decimal places it would end up appending the `456` as well right? When we only want to add back the `000`.

benelan (commented):
Couple more comments, but looking good!

packages/calcite-components/src/components/input-number/input-number.tsx
anveshmekala: i don't think so, if you see that `setNumberValue` parses delocalized value in `inputNumberInputHandler`. The only instance when we try to check if there is a trailingDecimalSeparator is when the value is deleted which invokes ` inputNumberInputHandler` method.

packages/calcite-components/src/utils/number.ts
anveshmekala: these are two diff scenarios, in the util method we add `decimalSeparator` only when we are appending decimal values. If there are decimal values then the condition `hasTrailingDecimalSeparator` return `false`.

packages/calcite-components/src/utils/number.ts
anveshmekala: Yeah, i thought about it. When i was trying to add just the missing zeros the function is getting a bit more complex but i see your point where this function localizes characters which are localized by then resulting in redundancy. It can be an issue for super long decimals. 

how about this 

```onst decimalSeparator = formatter.decimal;
  const decimals = value.split(".")[1];
  const localizedDecimals = localizedValue.split(decimalSeparator)[1];

  const diff = decimals && localizedDecimals ? decimals.length - localizedDecimals.length : 0;
  return localizedValue.padEnd(localizedValue.length + diff, formatter.localize("0"));
}

File: packages/calcite-components/src/components/input-number/input-number.tsx (hunk 5 above)
Comment (anveshmekala) [reply packages/calcite-components/src/components/input-number/input-number.tsx:849]: i assume you are referring to `trailingDecimalSeparator` , the value is already delocalized by then.

File: packages/calcite-components/src/utils/number.ts (hunk 7 above)
Comment (anveshmekala) [reply packages/calcite-components/src/utils/number.ts:252]: never mind, the above will fail

File: packages/calcite-components/src/components/input-number/input-number.tsx (hunk 5 above)
Comment (benelan) [reply packages/calcite-components/src/components/input-number/input-number.tsx:849]: If the value is already delocalized by then, we could remove the delocalization a couple lines down right?

File: packages/calcite-components/src/utils/number.ts (hunk 7 above)
Comment (benelan) [reply packages/calcite-components/src/utils/number.ts]: What you probably want is a regex expression like `/\..*0$/`. to check for trailing zeros after a decimal, but excluding non-zero numbers after the decimal

Comment (benelan) [reply packages/calcite-components/src/utils/number.ts]: Updating/simplified the regex. We don't care about the actual decimal because it is already being split, so we just need to grab the trailing zeros and append them to the localized value

File: packages/calcite-components/src/components/input-number/input-number.tsx (hunk 5 above)
Comment (anveshmekala) [reply packages/calcite-components/src/components/input-number/input-number.tsx:849]: i think so, it was added [here](https://github.com/Esri/calcite-components/pull/5427/files#diff-fe3d603d6390f87fdcdd3f40e34778ccdb07ff9edc52e440ad13c9571cfceadaR841) . When removed , all the tests are passing which is a good sign.

Comment (benelan) [reply packages/calcite-components/src/components/input-number/input-number.tsx]: Are we sure it is already delocalized in all cases? (e.g. typing, pasting, and setting the value via JS?)?

**Comment by anveshmekala** [comment]
Open question: Do we want to keep the trailing decimal zeros on initial render  ?  

Example: 
```
<calcite-input-number value="123.0" ></calcite-input-number>
```
Would the above should display `123` or `123.0` in input field after this change

File: packages/calcite-components/src/utils/number.spec.ts [hunk 8]
```diff
@@ -169,3 +173,34 @@ describe("expandExponentialNumberString", () => {
...
+      const localizedValue = numberStringFormatter.localize("123456.000");
+      const localizedZeroValue = numberStringFormatter.localize("0");
+      const result = `${localizedValue}${numberStringFormatter.decimal}`.padEnd(
+        localizedValue.length + 4,
+        localizedZeroValue
+      );
+      expect(addLocalizedTrailingDecimalZeros(localizedValue, "123456.000", numberStringFormatter)).toBe(result);
+    });
+
+    it(`returns same value if no trailing decimal zero value is removed- ${locale}`, () => {
```
jcfranco, line 195: Nitpick: `...removed- ${locale}` ➡️ `...removed - ${locale}` (space between `-`)

jcfranco (approved):
Code LGTM!

@benelan for another look-see. 1️⃣.0️⃣0️⃣0️⃣🔍👁️👄👁️

jcfranco:
> Open question: Do we want to keep the trailing decimal zeros on initial render ?

I think we'd want to keep them in order to align with developer expectations and with the behavior of adding it later if the value was initially `123`.

anveshmekala:
> > Open question: Do we want to keep the trailing decimal zeros on initial render ?
> 
> I think we'd want to keep them in order to align with developer expectations and with the behavior of adding it later if the value was initially `123`.

makes sense, the current expectation is to keep it to `123`.

packages/calcite-components/src/components/input-number/input-number.tsx
anveshmekala: yup, while typing or pasting InputNumberInput handler will delocalize the strings. For setting the value via JS, i don't think users can parse non latin number values which sounds strange tbh. I guess it is intentional to avoid localization twice.

packages/calcite-components/src/utils/number.spec.ts
```diff
@@ -169,3 +173,34 @@ describe("expandExponentialNumberString", () => {
...
+        locale,
+        // the group separator is different in arabic depending on the numberingSystem
+        numberingSystem: locale === "ar" ? "arab" : "latn",
+        useGrouping: true
+      };
+      const localizedValue = numberStringFormatter.localize("123456.000");
+      const localizedZeroValue = numberStringFormatter.localize("0");
+      const result = `${localizedValue}${numberStringFormatter.decimal}`.padEnd(
+        localizedValue.length + 4,
+        localizedZeroValue
```
benelan, line 190: can you add a few more spec tests for the following numbers:

- `1230000000000000000000000000000.00000000000000000000045000000000000000000000000`
- `-102.02100e1000000`

The regex may need to be adjusted to not consider trailing zeros after an `e`.

packages/calcite-components/src/components/input/input.e2e.ts
```diff
@@ -1291,6 +1291,80 @@ describe("calcite-input", () => {
...
+          await page.setContent(html`<calcite-input value="0.0000" lang="${locale}" type="number"></calcite-input>`);
+
+          numberStringFormatter.numberFormatOptions = {
+            locale,
+            numberingSystem: "latn",
+            useGrouping: false
+          };
+          const input = await page.find("calcite-input >>> input");
+          expect(await input.getProperty("value")).toBe("0");
+        });
```
benelan, line 1367: Can you add an e2e test (doesn't have to be for every locale) for this number and then type three backspaces: `2.100e10`.

packages/calcite-components/src/utils/number.ts
```diff
@@ -217,3 +227,29 @@ export function expandExponentialNumberString(numberString: string): string {
...
+ * @param {NumberStringFormat} numberStringFormatter - numberStringFormatter instance to localize the number value
+ * @returns {string} localized number string value
+ */
+export function addLocalizedTrailingDecimalZeros(
+  localizedValue: string,
+  value: string,
+  formatter: NumberStringFormat
+): string {
+  const trailingDecimalZeros = value.match(hasTrailingDecimalZeros)[0];
+  const decimalSeparator = formatter.decimal;
```
benelan, line 246: Nitpick: this line can be moved into the if statement

benelan (approved):
Added a couple comments - there may be one more edge case concerning exponential numbers. After that LGTM, great stuff! 

Any feedback on code clarity and/or places that could use some doc regarding the number sanitazation/localization logic would also be appreciated. I think you're the first person to touch that stuff since I did the big rework so I want to make sure its legible lol.

> Do we want to keep the trailing decimal zeros on initial render ?

Great question. I don't have a strong preference either way, @jcfranco @eriklharper any thoughts?

FWIW the native input doesn't remove decimal or trailing zero so following their lead would make sense

packages/calcite-components/src/utils/number.spec.ts
benelan: It would probably be easier to make sure `indexOf("e") === -1` rather than adding a negative lookaheads or something to the regex

packages/calcite-components/src/components/input/input.e2e.ts
benelan: For context, `2.100e` is `NaN` so I want to make sure the logic still works after deleting the `e`.

anveshmekala: `2.100e10` is  auto-corrected to `2.1e10` by default.

packages/calcite-components/src/utils/number.spec.ts
anveshmekala: Good catch, the 0 after the `e` should be ignored.

Issue #7130, opened 2023-06-05
Cannot type a decimal in a number input when the locale is arabic - Firefox only


ethanbdev:
### Actual Behavior

If I try to type a number such as "5.5: into a `calcite-input` with type `number`, it does not enter the decimal and just types "55"

### Expected Behavior

It should type "5.5" like it does in Chrome

### Reproduction Sample

https://codepen.io/eborgen/pen/PoxYxdR?editors=1010

### Reproduction Steps

1. Open the codepen
2. Try to type "5.5"
3. Observe "55" is displayed

You can also,
1. Type "5.5" somewhere else and copy/paste, it will change it to `5,5`
2. This is not correct per https://i18n.esri.com/website/i18ntools (Use Arabic - Saudi Arabia for testing)

### Reproduction Version

1.4.1

### Relevant Info

I don't think this has worked for a while

### Priority impact

p3 - want for upcoming milestone

### Esri team

ArcGIS Web Analysis

geospatialem:
Perhaps could be tackled with https://github.com/Esri/calcite-components/issues/7039?

anveshmekala:
From the initial debugging , it is identified that `Intl.NumberFormat` API is returning `latn` numbers with `.` as decimal separator  in chrome where as in FireFox the decimal separator is `,` with `arab` numbers though the numbering-system is set to `latn` in both instances.

geospatialem:
Verified in `1.4.3-next.5` with Firefox v`102.12.0esr`.

PR #7173, opened 2023-06-13
fix(input,input-number): allow typing decimal separator in firefox for arabic locale

anveshmekala:
**Related Issue:** #7130 

## Summary

This PR will allow users to type `latn` decimal separator for `arabic` lang in  Firefox & safari for `calcite-input` and `calcite-input-number`.

### DIFF
--- a/packages/calcite-components/src/utils/locale.ts
+++ b/packages/calcite-components/src/utils/locale.ts
@@ -402,7 +402,11 @@ export class NumberStringFormat {
     ].reverse();
 
     const index = new Map(this._digits.map((d, i) => [d, i]));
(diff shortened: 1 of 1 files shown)


driskull (approved):
👍

driskull:
@anveshmekala can you add an e2e test for this? would that be doable?

anveshmekala:
> can you add an e2e test for this? would that be doable?

I am not certain if we can simulate Firefox browser in e2e test with the current setup. I was hoping to do that in chromatic but looks like we can only enable Firefox browser for the whole repo rather than a single story or a file.

Issue #7158, opened 2023-06-08
`calcite-popover` does not open in a specific case with `calcite-tooltip`


jmanke:
### Actual Behavior

When a `calcite-action` (element type may not be important) is the target of both a `calcite-popover` and `calcite-tooltip`, the `calcite-popover` will not open if the `calcite-tooltip` is in the process of appearing. For example, the user hovers over the `calcite-action` and the tooltip starts to appear - at this exact time, clicking on the action will not open the `calcite-popover`. The user will need to click the `calcite-action` two more times for the `calcite-popover` to appear.

### Expected Behavior

The `calcite-popover` should appear when the user clicks its target.

### Reproduction Sample

https://codepen.io/jmanke/pen/ExOabVd?editors=1011

### Reproduction Steps

The following repro steps are for this [codepen](https://codepen.io/jmanke/pen/ExOabVd?editors=1011):

  1. Ensure page is refreshed
  2. Hover over action with mouse. Right as the tooltip appears, click on the action. This step is difficult to time just right.
  3. Observe that the popover does not open. It requires clicking on the action two more time to open it.

### Reproduction Version

v1.4.2

### Relevant Info

Tested on Chrome v114 on MacOS Ventura 13.4

### Priority impact

p3 - want for upcoming milestone

### Impact

Medium - it feels like the popover is broken for some users in this particular workflow.

### Esri team

ArcGIS Dashboards

driskull:
It seems to have to do with debouncing the reposition() method internally. If you resize the page when it gets into one of those bad states it seems to bring the component up to date because it calls reposition again.

@jmanke does this only happen when there is both a tooltip and popover or can you reproduce with just a popover?

jmanke:
@driskull I can't seem to repro the issue when I remove the tooltip. It may be a timing issue on my end, but I've tried ~50 times and couldn't repro the bug. The only way I've been able to repro is with the steps above.

geospatialem:
Verified in `1.5.0-next.4`.

PR #7178, opened 2023-06-14
fix(combobox, dropdown, input-date-picker, input-time-picker, popover, tooltip): Prevent repositioning from affecting other floating components

driskull:
**Related Issue:** #7158

## Summary

This prevent repositioning from affecting other floating components caused by using a shared, debounced reposition function.

Additionally, this will cancel any in-progress debounced calls if the component is disconnected.


jcfranco:
@driskull Updated the PR per our conversation.

packages/calcite-components/src/utils/floating-ui.ts
```diff
@@ -53,6 +53,138 @@ async function patchFloatingUiForNonChromiumBrowsers(): Promise<void> {
...
+    });
+
+    if (arrowEl && middlewareData.arrow) {
+      const { x, y } = middlewareData.arrow;
+      const side = effectivePlacement.split("-")[0] as Side;
+      const alignment = x != null ? "left" : "top";
+      const transform = ARROW_CSS_TRANSFORM[side];
+      const reset = { left: "", top: "", bottom: "", right: "" };
+
+      if ("floatingLayout" in component) {
```
benelan, line 158: should we use `Object.hasOwn(component, "floatingLayout)` here instead of `in` so we don't need spend extra time checking component's prototype chain? Or do we care about inherited properties in this scenario?

benelan (approved):
One comment but otherwise LGTM!

packages/calcite-components/src/utils/floating-ui.ts
jcfranco: Great observation. I think we could look at our `in` usage across the board. Maybe we could use ESLint to help enforce this (haven't found an existing rule at the moment 🤔). 

Currently, I'm not too worried about the prototype lookup, but it'd be good to make this change for the future.

WRT `hasOwn`, I think we could to wait for Safari 17 before we can start using it. Unless we use `hasOwnProperty` and then switch to `hasOwn` along with https://eslint.org/docs/latest/rules/prefer-object-has-own.

eriklharper (approved):
LGTM :)

PR #7179, opened 2023-06-15
revert: Add slots for filter actions"

driskull:
Reverts Esri/calcite-components#7148


jcfranco:
Thanks for reverting, @driskull!

For context, this is being reverted to keep the June milestone patch-only. This will be reinstalled once the release is wrapped up.

Issue #7147, opened 2023-06-06
calcite-list - initial empty list with loading=true is not visually tall enough to show loading animation


jgrayson-apl:
### Actual Behavior

When an empty calcite-list with loading=true is initially displayed it's not visually tall enough to show loading animation.

### Expected Behavior

I would expect the calcite-list to be tall enough to display the loading animation even if the list is empty

### Reproduction Sample

https://codepen.io/john-grayson/pen/gOQOmZE

### Reproduction Steps

1. Load the above codepen
2. Watch how the list is initially not tall enough to display the loading animation

### Reproduction Version

1.4.2

### Priority impact

p4 - not time sensitive

### Esri team

N/A

geospatialem:
Also appears a bit cut off when only one `list-item` is present: https://codepen.io/geospatialem/pen/MWzWmzQ.

jgrayson-apl:
FYI: I updated my codepen with the workaround I'm using in my apps.  I basically unset/reset the 'position' style property of the internal container.

ashetland:
Could List accommodate the Loader with some minimal padding? In this example, the empty List would be 80px tall while loading.

<img width="438" alt="CleanShot 2023-06-13 at 16 15 50@2x" src="https://github.com/Esri/calcite-components/assets/108549080/43623099-a9a6-4bd2-88b4-5e94778c826f">

driskull:
Yes, that is super doable.

driskull:
I think it would be better if the `calcite-scrim` was just more responsive to adjust the loading spinner size relative to available height. That way, the list would't get taller just for a loading spinner if only 1 item exists.

driskull:
When the loader component inside the scrim is is set to `scale="s"` it looks good.
 
![image](https://github.com/Esri/calcite-components/assets/1231455/93a7d85e-b0ec-47f8-a0e7-5c03ccec95d2)

ashetland:
And assuming List gets small and large scales added to it, we'd probably need to ensure enough height at small scale for the small scale Loader inside Scrim.

geospatialem:
Verified with https://codepen.io/geospatialem/pen/MWzWmzQ in `1.5.0-next.4`.

PR #7182, opened 2023-06-15
fix(scrim): Responsively set the scale of the loading spinner

driskull:
**Related Issue:** #7147

## Summary

- Updates the scrim to set the scale of the internal loading spinner based on the scrim size.
- Scrim is now responsive with a resize observer.
- Breakpoints added based on design. These could maybe go into some kind of component token in the future.
- Breakpoint is based on the minimum value of width or height.
- Adds test

### DIFF
--- a/packages/calcite-components/src/components/scrim/resources.ts
+++ b/packages/calcite-components/src/components/scrim/resources.ts
@@ -4,6 +4,7 @@ export const CSS = {
 };
 
 export const BREAKPOINTS = {
-  s: 72,
-  l: 480
+  s: 72, // Less than 72px.
+  // medium is assumed default.
+  l: 480 // Greater than or equal to 480px.
 };
--- a/packages/calcite-components/src/components/scrim/resources.ts
+++ b/packages/calcite-components/src/components/scrim/resources.ts
@@ -4,6 +4,6 @@ export const CSS = {
 };
 
 export const BREAKPOINTS = {
-  s: 150,
-  l: 350
+  s: 72,
+  l: 480
 };
--- a/packages/calcite-components/src/components/scrim/resources.ts
+++ b/packages/calcite-components/src/components/scrim/resources.ts
@@ -2,3 +2,8 @@ export const CSS = {
   scrim: "scrim",
   content: "content"
 };
+
+export const BREAKPOINTS = {
+  s: 150,
+  l: 350
+};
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/scrim/scrim.tsx
```diff
@@ -58,11 +60,11 @@ export class Scrim implements LocalizedComponent, T9nComponent {
 
   @Element() el: HTMLCalciteScrimElement;
 
-  // --------------------------------------------------------------------------
-  //
```
jcfranco, line 62: ✨🧹✨

packages/calcite-components/src/components/scrim/scrim.tsx
```diff
@@ -116,4 +120,35 @@ export class Scrim implements LocalizedComponent, T9nComponent {
...
+    if (size < BREAKPOINTS.s) {
+      return "s";
+    } else if (size >= BREAKPOINTS.l) {
+      return "l";
+    } else {
+      return "m";
+    }
+  }
+
+  private handleResize = (): void => {
```
jcfranco, line 145: This can be a class method (instead of an assigned bound function) because its call sites already bind `this` to the instance.

packages/calcite-components/src/components/scrim/resources.ts
```diff
@@ -2,3 +2,8 @@ export const CSS = {
   scrim: "scrim",
   content: "content"
 };
+
+export const BREAKPOINTS = {
+  s: 72,
+  l: 480
```
jcfranco, line 8: Can you add a comment explaining why `m` is left out?

jcfranco (approved):
📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏
📏🔄📏📏📏🔄📏🔄🔄🔄📏📏🔄🔄🔄📏🔄🔄🔄🔄📏🔄📏
📏🔄🔄📏📏🔄📏📏🔄📏📏🔄📏📏📏📏🔄📏📏📏📏🔄📏
📏🔄📏🔄📏🔄📏📏🔄📏📏🔄📏📏📏📏🔄🔄🔄📏📏🔄📏
📏🔄📏📏🔄🔄📏📏🔄📏📏🔄📏📏📏📏🔄📏📏📏📏📏📏
📏🔄📏📏📏🔄📏🔄🔄🔄📏📏🔄🔄🔄📏🔄🔄🔄🔄📏🔄📏
📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏

Would this be possible to simplify by using container queries? We'd need to set some polyfills, but it looks like it could help here.

driskull:
> Would this be possible to simplify by using container queries? We'd need to set some polyfills, but it looks like it could help here.

I think a resize observer makes most sense in this case since we're in JS land. If it were a pure CSS change then container queries.

Issue #6601, opened 2023-03-14
Calcite Components React should auto import web component


am-maneaters:
### Summary

Having to import components twice in the React wrapper adds extra complexity to using the package. 
i.e.

```ts
import '@esri/calcite-components/dist/components/calcite-notice';
import { CalciteNotice } from "calcite-components-react";
```

Intellisense in VSCode knows to auto import the `CalciteNotice` when I use it in JSX, but then I have to manually include web-component import as well. I've resorted to having a single file where I import every calcite component import that I just include in my index file.

### Desired Outcome

Since the react wrappers are already generated using a set of scripts, we could extend those scripts or write a secondary script that auto attaches `calcite-component` imports to the respective `calcite-components-react` component. 

```ts
import { CalciteNotice } from "calcite-components-react";
```

I have a local version of this working using dynamic imports in the `createComponent.js` file that is included in the stencil-react-output-target.

### Resources

the way i implemented this was at this [line](https://github.com/ionic-team/stencil-ds-output-targets/blob/967739d7ac3bc46a52c0d48bdafa8fa909e787d6/packages/react-output-target/react-component-lib/createComponent.tsx#L31), i added

```
import(`../../../calcite-components/dist/components/${tagName}`)
```

oh, or probably could just use this? https://stenciljs.com/docs/react#includeimportcustomelements

jcfranco:
> oh, or probably could just use this? https://stenciljs.com/docs/react#includeimportcustomelements

Great suggestion. Enabling this will make the wrappers behave closer to the custom-elements output target. cc @benelan

geospatialem:
Verified in the [example repo's App.js](https://github.com/Esri/calcite-components-examples/blob/master/react/src/App.js) with `1.5.0-next.11` with the following imports:

```js
import React, { useState } from 'react';
import { CalciteButton, CalciteIcon, CalciteSlider } from '@esri/calcite-components-react';
import '@esri/calcite-components/dist/calcite/calcite.css';
```

PR #7185, opened 2023-06-15
feat: automatically import and define Calcite Components when importing their React wrapper

benelan:
**Related Issue:** #6601

## Summary

Automatically imports and defines Calcite Components when importing the React wrappers. For example, here is a snippet from our React sample:

```jsx
import '@esri/calcite-components/dist/components/calcite-button';
import '@esri/calcite-components/dist/components/calcite-icon';
import '@esri/calcite-components/dist/components/calcite-slider';
import { CalciteButton, CalciteIcon, CalciteSlider } from '@esri/calcite-components-react';
```

The above snippet can now be simplified to the following:

```jsx
import { CalciteButton, CalciteIcon, CalciteSlider } from '@esri/calcite-components-react';
```

### Note

I had to patch the build because Stencil was generating an incorrect import path to the JSX type. I tested in our old output targets repo and it reproduced there as well, so this isn't related to the monorepo change.


jcfranco (approved):
Awesome!

Will this feature bump `calcite-components` version or will it ship silently under the wrapper's synced version, or both? If this does bump the version, let's merge after the June (patch-only) release is done.

benelan:
> Awesome!
> 
> Will this feature bump `calcite-components` version or will it ship silently under the wrapper's synced version, or both? If this does bump the version, let's merge after the June (patch-only) release is done.

Yeah it would bump CC's version too because we are only doing patches for CC at the moment. So we can hold off for now.

Issue #6616, opened 2023-03-20
calcite-alert: queue corruption after removal of visible alert


timmorey:
### Actual Behavior

If there is a queue of `<calcite-alert>`s, and the first alert is forcibly removed from the DOM without being properly dismissed, the alert queue will become corrupt and future alerts will not be rendered.

### Expected Behavior

The alert queue should handle removal of alerts from the DOM.

With dynamic rendering in modern apps, it is quite possible that an alert may become severed from the document before it has been properly closed by the user.  While we may recommend that alerts be properly closed and allowed to animate out before they are removed from the DOM, we should be tolerant of cases where they are not, and we should attempt to keep internal state consistent to allow future alerts to render.

### Reproduction Sample

https://codepen.io/timmorey/pen/yLxqdoE

### Reproduction Steps

1. Click "Add alert" twice
2. Observe that "alert 0" is visible, and that there is one more alert queued.
3. Click "Remove alert" once
4. Observe that no alerts are visible.  

The expected behavior is that "alert 1" should be visible at step 4.  If you inspect the dom at this time, you'll see that the second alert remains queued.

### Reproduction Version

1.1.0

### Priority impact

p3 - want for upcoming milestone

### Impact

We are building components to be shared with other teams, and sometimes these components want to show alerts to the user to indicate an error condition or report the success of an operation.  We would like to declaratively render these alerts in the components that trigger them, rather than insist that the parent application manage the alerts.  However, we cannot guarantee that all alerts will have been gracefully dismissed by the time our components are destroyed, and we don't want to break the alert queue for the parent application.

### Esri team

ArcGIS Field Apps

geospatialem:
Verified with `1.5.0-next.4`.

PR #7189, opened 2023-06-16
fix(alert): update alert queue when an alert is removed from the DOM

driskull:
**Related Issue:** #6616

## Summary

- Adds new internal event `calciteInternalAlertUnregister` to dispatch when an alert is removed from the DOM.
- Event is dispatched on the window since that is where the listeners are and dispatching on the element won't reach the window since the element would no longer be in the DOM.
- When the event is received, the queue is updated to remove the removed alert element.
- queue is kept n'sync.

giphy


packages/calcite-components/src/components/alert/alert.tsx
```diff
@@ -203,6 +203,12 @@ export class Alert implements OpenCloseComponent, LoadableComponent, T9nComponen
   }
 
   disconnectedCallback(): void {
+    window.dispatchEvent(
```
driskull, line 206: Instead of having to do this it might be nice if we had a register of alert elements on the alert instance. Should we update this pattern? Would we need an epic to do it in other components as well?

driskull: https://github.com/Esri/calcite-components/issues/7200

packages/calcite-components/src/components/alert/alert.tsx
```diff
@@ -353,6 +359,20 @@ export class Alert implements OpenCloseComponent, LoadableComponent, T9nComponen
...
 
+  // when an alert is unregistered, trigger a queue sync
+  @Listen("calciteInternalAlertUnregister", { target: "window" })
+  alertUnregister(event: CustomEvent<Unregister>): void {
+    const queue = this.queue.filter((el) => el !== event.detail.alert);
+    this.queue = queue;
+
+    window.dispatchEvent(
+      new CustomEvent<Sync>("calciteInternalAlertSync", {
+        cancelable: false,
```
jcfranco, line 370: Custom events are not cancelable by default, so this can be dropped.

packages/calcite-components/src/components/alert/alert.tsx
```diff
@@ -353,6 +359,20 @@ export class Alert implements OpenCloseComponent, LoadableComponent, T9nComponen
     this.determineActiveAlert();
   }
 
+  // when an alert is unregistered, trigger a queue sync
```
jcfranco, line 362: I think we can drop this comment since the code is pretty expressive. Maybe it'd be better to add a note on why we need to dispatch on the window instead.

jcfranco (approved):
🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺
🕺🎤🎤🎤🎤🕺🎤🎤🎤🎤🕺🎤🎤🎤🎤🕺🎤🕺🕺🕺🎤🕺
🕺🎤🕺🕺🎤🕺🎤🕺🕺🎤🕺🎤🕺🕺🎤🕺🕺🎤🕺🎤🕺🕺
🕺🎤🎤🎤🕺🕺🎤🎤🎤🎤🕺🎤🎤🎤🕺🕺🕺🕺🎤🕺🕺🕺
🕺🎤🕺🕺🎤🕺🎤🕺🕺🎤🕺🎤🕺🕺🎤🕺🕺🕺🎤🕺🕺🕺
🕺🎤🎤🎤🎤🕺🎤🕺🕺🎤🕺🎤🎤🎤🎤🕺🕺🕺🎤🕺🕺🕺
🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺


🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺
🕺🎤🎤🎤🎤🕺🎤🕺🕺🕺🎤🕺🎤🎤🎤🎤🕺
🕺🎤🕺🕺🎤🕺🕺🎤🕺🎤🕺🕺🎤🕺🕺🕺🕺
🕺🎤🎤🎤🕺🕺🕺🕺🎤🕺🕺🕺🎤🎤🎤🕺🕺
🕺🎤🕺🕺🎤🕺🕺🕺🎤🕺🕺🕺🎤🕺🕺🕺🕺
🕺🎤🎤🎤🎤🕺🕺🕺🎤🕺🕺🕺🎤🎤🎤🎤🕺
🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺

🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺
🕺🎤🎤🎤🎤🕺🎤🕺🕺🕺🎤🕺🎤🎤🎤🎤🕺
🕺🎤🕺🕺🎤🕺🕺🎤🕺🎤🕺🕺🎤🕺🕺🕺🕺
🕺🎤🎤🎤🕺🕺🕺🕺🎤🕺🕺🕺🎤🎤🎤🕺🕺
🕺🎤🕺🕺🎤🕺🕺🕺🎤🕺🕺🕺🎤🕺🕺🕺🕺
🕺🎤🎤🎤🎤🕺🕺🕺🎤🕺🕺🕺🎤🎤🎤🎤🕺
🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺

🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺
🕺🎤🎤🎤🎤🕺🎤🕺🕺🕺🎤🕺🎤🎤🎤🎤🕺
🕺🎤🕺🕺🎤🕺🕺🎤🕺🎤🕺🕺🎤🕺🕺🕺🕺
🕺🎤🎤🎤🕺🕺🕺🕺🎤🕺🕺🕺🎤🎤🎤🕺🕺
🕺🎤🕺🕺🎤🕺🕺🕺🎤🕺🕺🕺🎤🕺🕺🕺🕺
🕺🎤🎤🎤🎤🕺🕺🕺🎤🕺🕺🕺🎤🎤🎤🎤🕺
🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺🕺

PR #7190, opened 2023-06-16
chore: issue template context for issue check and package

geospatialem:
**Related Issue:** n/a

## Summary
Adds the following to the a11y, bug, and enhancement templates:

1. Adds a required checkbox verifying the submitter has checked for existing issues
image

2. Adds checkbox options to specify the package of the request
image
 

cc @benelan for the ci/action on the checkbox options to add the `framework-react` label to the CCR checkbox when checked.


benelan:
Made a couple tweaks to make my life easier:

- The action will add a label named after the NPM package (minus the `@esri` scope), e.g. `calcite-components` or `calcite-components-react`
- As of now, this won't work if the package isn't part of the `@esri` NPM scope. We can refactor later if we end up adding a package in a different scope, but I don't think any `@arcgis` packages will end up in here.

.github/workflows/add-package-label.yml
```diff
@@ -0,0 +1,56 @@
...
+            if (!body) {
+              console.log("could not determine the issue body");
+              return;
+            }
+
+            // NOTE: assumes all packages will be in the @esri NPM scope
+            const packageRegex = /(?<=\[X\]\s@esri\/)[\w-]*$/gm;
+            const packages = body.match(packageRegex);
+
+            if (packages?.length) {
```
jcfranco, line 29: We could eliminate this check, and its associated nesting, by providing an empty array as a fallback when assigning `packages` above:

```ts
const packages = body.match(packageRegex) || [];
```

jcfranco (approved):
Very nice! ⚡🤖⚡

Issue #6200, opened 2022-12-29
Documenting conventions for the coupled parent-child relationship.


Elijbet:
### Description

Parents should be setting props on the child if it’s a tightly coupled component pair.  We don’t want people setting them individually. We’d have inconsistencies with different scales for example, which might not be intended. The logic to achieve this is outlined in #6038.

Not applicable for independent components, where it makes sense for the child to have autonomy, eg. a large modal with smaller child components. These children of course need to retain their scale.

Define what a coupled parent-child is and what isn’t:
Tightly coupled components will be required to have the child share their parent’s name, like `list/list-item`, `flow/flow-item`, `combobox/combobox-item`.

Questions:
Does this apply to all props on all coupled children? Do we just refactor all at once or go case by case?

To do:
Children’s props should be refactored to not be public (remove `@Prop` decorator). 
Undocument these new internal props, so it’s clear the user shouldn’t be setting these. 
Keep defaults for children’s internal props that are controlled by parents.
Don’t reflect, as this would expose them. 
This means that CSS classes need to be updated to use classes instead of attributes.

Once we decide the rules and document the convention in a markdown then we can spin up an epic issue.

### Which Component

All

geospatialem:
The milestone should be associated with https://github.com/Esri/calcite-components/issues/6038 as the two are related and should be addressed together.

Elijbet:
Closing, as it is now combined with #6038.

Elijbet:
Closing, as it is now combined with https://github.com/Esri/calcite-design-system/issues/6038.

PR #7191, opened 2023-06-17
fix(accordion, accordion-item): `icon-position`, `icon-type`, `selection-mode` and `scale` can now be set as props or attributes

### LINKED ISSUE #6199: Accordion icon-position can only be set as an attribute
### Actual Behavior

If you set `accordionEl.iconPosition = "start"` the icons of the children accordion items will still appear at the end of the accordion. 

However, if you use `accordionEl.setAttribute("icon-position", "start")` or just set it with kebab case in the DOM, it works great. 

### Expected Behavior

All props should function using either attributes or... props. We generally prefer using the camel case versions as you get much better type hinting for the values. 

### Reproduction Sample

https://codepen.io/paulcp/pen/VwBaPEd?editors=0010

### Reproduction Steps

1. look at the icon position (will be at the end)
2. now change the code to use the attribute version
3. notice the icon moves to the start

### Reproduction Version

All

### Relevant Info

_No response_

### Regression?

_No response_

### Impact

_No response_

### Esri team

ArcGIS Online

Elijbet:
**Related Issue:** #6199, #6038, #6200

## Summary

This should resolve #6199, where you're able to set `accordionEl.setAttribute("icon-position", "start")`, but not  `accordionEl.iconPosition = "start"`, happening because of a logical issue within `getElementProp` function.

`getElementProp` is being refactored across child components as an outdated pattern in #6038, so this will also kick off that issue as well.

Instead of the `accordion-item` looking up the parent for `iconPosition, iconType, selectionMode, or scale`, these get set by the `accordion` parent. 

The logic for setting these props thus moves to the parent, getting rid of the `getElementProp` altogether. The parent component gets a `mutation observer` to do this as well as `watchers` for when it needs to modify the children.

With the `mutationObserver` added, there is no longer a need for the `calciteInternalAccordionItemRegister` event logic.


packages/calcite-components/src/components/accordion/accordion.tsx
```diff
@@ -116,6 +145,9 @@ export class Accordion {
   //  Private State/Props
   //
   //--------------------------------------------------------------------------
+  mutationObserver = createObserver("mutation", () => this.passPropsToAccordionItems());
+
+  @State() accordionItems: HTMLCalciteAccordionItemElement[] = [];
```
driskull, line 150: I don't think this needs to be state since its not rendering.

packages/calcite-components/src/components/accordion-item/accordion-item.tsx
```diff
@@ -69,6 +68,37 @@ export class AccordionItem implements ConditionalSlotComponent {
...
+   *
+   * @internal
+   */
+  @Prop({ reflect: false, mutable: false }) iconPosition: Position = "end";
+
+  /** Specifies the type of the icon in the header inherited from the `calcite-accordion`.
+   *
+   * @internal
+   */
+  @Prop({ reflect: false, mutable: false }) iconType: "chevron" | "caret" | "plus-minus" =
```
driskull, line 82: `mutable: false` is default so this can be removed.

packages/calcite-components/src/components/accordion-item/accordion-item.tsx
```diff
@@ -98,10 +128,6 @@ export class AccordionItem implements ConditionalSlotComponent {
 
   connectedCallback(): void {
     this.parent = this.el.parentElement as HTMLCalciteAccordionElement;
```
driskull, line 130: The parent could probably be set as well.

packages/calcite-components/src/components/accordion/accordion.tsx
```diff
@@ -45,6 +57,14 @@ export class Accordion {
     SelectionMode
   > = "multiple";
 
+  @Watch("iconPosition")
+  @Watch("iconType")
+  @Watch("scale")
+  @Watch("selectionMode")
+  onInternalPropChange(): void {
```
driskull, line 64: maybe `handlePropsChange`? Since these aren't internal at this point.

packages/calcite-components/src/components/accordion/accordion.tsx
```diff
@@ -62,13 +82,22 @@ export class Accordion {
   //
   //--------------------------------------------------------------------------
 
+  connectedCallback(): void {
+    this.passPropsToAccordionItems();
```
driskull, line 86: maybe call this `updateAccordionItems`?

packages/calcite-components/src/components/accordion/accordion.tsx
driskull: there's already a var that can be used for the items `this.items` instead of `accordionItems`

driskull (commented):
Looking good! Not sure what the issue is. Can you debug and see where the failure is occuring?

packages/calcite-components/src/components/accordion-item/accordion-item.tsx
Elijbet: Hmm... not sure how to achieve this. What do you have in mind?

packages/calcite-components/src/components/accordion-item/accordion-item.tsx
```diff
@@ -69,6 +68,36 @@ export class AccordionItem implements ConditionalSlotComponent {
   /** Displays the `iconStart` and/or `iconEnd` as flipped when the element direction is right-to-left (`"rtl"`). */
   @Prop({ reflect: true }) iconFlipRtl: FlipContext;
 
+  /**
+   * Specifies the placement of the icon in the header inherited from the `calcite-accordion`.
+   *
+   * @internal
+   */
+  @Prop({ reflect: false }) iconPosition: Position = "end";
```
driskull, line 76: @Elijbet `reflect: false` is the default so I don't think this needs to be specifically defined in the code. Applies to other props.

driskull: For these new internal props, do they need to have a default set or could they be undefined by default?

packages/calcite-components/src/components/accordion-item/accordion-item.tsx
driskull: What I was thinking is that the owning accordion could set itself on the items as the parent. Just like how its setting the icontype, iconPosition, scale.

packages/calcite-components/src/components/accordion/accordion.tsx
```diff
@@ -116,9 +144,17 @@ export class Accordion {
   //  Private State/Props
   //
   //--------------------------------------------------------------------------
+  mutationObserver = createObserver("mutation", () => this.updateAccordionItems());
 
   /** created list of Accordion items */
-  private items = [];
+  accordionItems: HTMLCalciteAccordionItemElement[] = [];
```
driskull, line 150: Instead of adding `accordionItems` can we just use the existing `this.items`? It just seems like adding a redundant private property.

driskull (dismissed):
Looking good @Elijbet ✨ 👍

packages/calcite-components/src/components/accordion/accordion.tsx
```diff
@@ -62,13 +81,22 @@ export class Accordion {
   //
   //--------------------------------------------------------------------------
 
+  connectedCallback(): void {
+    this.mutationObserver?.observe(this.el, { childList: true, subtree: true });
```
jcfranco, line 85: I _think_ `childList` should suffice since we would only update items when the direct children change, and not the children of each item (i.e., nesting is not supported).

packages/calcite-components/src/components/accordion/accordion.tsx
```diff
@@ -116,9 +144,17 @@ export class Accordion {
   //  Private State/Props
   //
   //--------------------------------------------------------------------------
+  mutationObserver = createObserver("mutation", () => this.updateAccordionItems());
```
jcfranco, line 147: Please add a newline above this to preserve comment divider spacing.

packages/calcite-components/src/components/accordion/accordion.tsx
```diff
@@ -82,17 +85,21 @@ export class Accordion {
...
 
   componentDidLoad(): void {
     if (!this.sorted) {
-      this.items = this.sortItems(this.items);
+      this.accordionItems = this.sortItems(this.accordionItems);
       this.sorted = true;
     }
   }
 
+  // componentDidRender(): void {
```
driskull, line 99: Can you remove this empty comment block?

packages/calcite-components/src/components/accordion/accordion.e2e.ts
```diff
@@ -33,11 +41,26 @@ describe("calcite-accordion", () => {
...
-    expect(element).toEqualAttribute("icon-type", "chevron");
+
+    expect(await element.getProperty("appearance")).toBe("solid");
+    expect(await element.getProperty("iconPosition")).toBe("end");
+    expect(await element.getProperty("scale")).toBe("m");
+    expect(await element.getProperty("selectionMode")).toBe("multiple");
+    expect(await element.getProperty("iconType")).toBe("chevron");
+  });
+
+  it("renders inheritable props: `iconPosition`, `iconType`, `selectionMode`, and `scale` to match those set on parent (non-default)", async () => {
```
jcfranco, line 52: Could you also add a test to cover #6199?

packages/calcite-components/src/components/accordion/accordion.tsx
```diff
@@ -117,8 +135,26 @@ export class Accordion {
...
+    for (const mutation of mutationList) {
+      if (mutation.type === "childList") {
+        const addedNodes = Array.from(mutation.addedNodes);
+        addedAccordionItem = addedNodes.find(
+          (node) => node instanceof HTMLCalciteAccordionItemElement
+        ) as HTMLCalciteAccordionItemElement;
+      }
+    }
+    const childNodes = Array.from(this.el.childNodes);
+    const position = childNodes.indexOf(addedAccordionItem);
```
jcfranco, line 149: Do you still need `position` and sort by it after setting up the mutation observer? I think this could be simplified by updating accordion items by their order in the DOM.

jcfranco:
> This should resolve https://github.com/Esri/calcite-components/issues/6199

This means this PR is a fix and no longer a refactor. 🥳

packages/calcite-components/src/components/accordion/accordion.tsx
```diff
@@ -132,6 +153,24 @@ export class Accordion {
   //
   //--------------------------------------------------------------------------
 
-  private sortItems = (items: any[]): any[] =>
-    items.sort((a, b) => a.position - b.position).map((a) => a.item);
+  private updateAccordionItems = (addedAccordionItem?: HTMLCalciteAccordionItemElement): void => {
+    if (addedAccordionItem) {
```
benelan, line 157: What happens if an accordion item is removed from the DOM? Does it get removed from `accordionItems` somehow or will that cause issues?

packages/calcite-components/src/components/accordion/accordion.tsx
```diff
@@ -132,6 +153,24 @@ export class Accordion {
...
+      this.accordionItems.push(addedAccordionItem);
+    } else {
+      const accordionItemsQueryArray = Array.from(
+        this.el.querySelectorAll("calcite-accordion-item")
+      );
+      this.accordionItems = accordionItemsQueryArray;
```
benelan, line 163: What is the purpose of pushing `addedAccordionItems` into `accordionItems` if `accordionItems` is being re-referenced to `accordionItemsQueryArray` directly afterwards?

benelan (commented):
I have a couple quick questions questions. Also can you change the PR title to be about #6199 rather than the refactor, since the fix is what will show up in the changelog

packages/calcite-components/src/components/accordion/accordion.tsx
benelan: Nevermind it looks like it will get removed on line 163, but my question below still stands

benelan: Although it looks like `updateAccordionItems()` exits early if there were no items added. So do we need to handle items being removed separately? I haven't spent a lot of time working in accordion so it's an open question, could be fine as is.

packages/calcite-components/src/components/accordion/accordion.tsx
Elijbet: Looks like it isn't needed anymore. Removing position removed the need for accordionItems being an array of objects, which makes that bit redundant.

packages/calcite-components/src/components/accordion/accordion.tsx
```diff
@@ -128,18 +128,7 @@ export class Accordion {
...
-    let addedAccordionItem: HTMLCalciteAccordionItemElement;
-    for (const mutation of mutationList) {
-      if (mutation.type === "childList") {
-        const addedNodes = Array.from(mutation.addedNodes);
-        addedAccordionItem = addedNodes.find(
-          (node) => node instanceof HTMLCalciteAccordionItemElement
-        ) as HTMLCalciteAccordionItemElement;
-      }
-    }
-    this.updateAccordionItems(addedAccordionItem);
```
benelan: Hmm idk how I feel about this. Removing this part will have a performance impact because now the DOM is being queried on every mutation, whereas before it was only being queried when nodes were added. 

`querySelectorAll` is a pretty intensive task, especially for something like accordion which can have a yuuge amount of child nodes.

benelan: Do we really need to query the DOM at all? We should be able to get the added/removed/target nodes from the mutation observer right

Elijbet: updateAccordionItems() runs also on watcher for a number of props that are set by parents on children. We need to keep track of items for updateAccordionItems to map set props/prop changes onto them. Looks like at this point that's all updateAccordionItems is doing, as we no longer keep track of position, or items being added because of the query.

Mutation observer track additions, removals, and attribute changes, right? If that's the case why do we even need a watcher? We could just run the update function to sync props on component load in addition to the observer, but we'd need to query still.

benelan: Yeah now that the constraint of only querying when the mutation included addedNodes is removed, I'm pretty sure the watchers are redundant. 

However I'm still concerned about using querySelectorAll on every mutation. There has to be a better way to do this that doesn't have the potential to explode someone's computer. Found this explaination after some googling: https://stackoverflow.com/a/39332340

benelan: Nevermind, it looks like the MutationObserver isn't set up to observe attributes so I think we would still need to watchers

benelan: Also it's not watching the whole subtree so the performance hit won't be too bad, I guess. Does the current logic cover nested accordions or accordion items?

packages/calcite-components/src/components/accordion/accordion.e2e.ts
```diff
@@ -33,11 +41,54 @@ describe("calcite-accordion", () => {
...
+    </calcite-accordion>`);
+    const element = await page.find("calcite-accordion");
+
+    expect(await element.getProperty("iconPosition")).toBe("start");
+    expect(await element.getProperty("iconType")).toBe("plus-minus");
+    expect(await element.getProperty("selectionMode")).toBe("single-persist");
+    expect(await element.getProperty("scale")).toBe("l");
+  });
+
+  it("accordion properties can be set as an attribute as well as a prop", async () => {
```
jcfranco, line 66: This test would be covered by using the `reflects` test helper.

jcfranco (approved):
Once the review comments have been addressed, this should be good to merge! 🚀

PR #7201, opened 2023-06-20
chore: add eslint-plugin-calcite-components package

jcfranco:
**Related Issue:** N/A

## Summary

Adds https://github.com/Esri/eslint-plugin-calcite-components as a package.


packages/eslint-plugin-calcite-components/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "dependencies": {
+    "stencil-eslint-core": "0.3.1"
+  },
+  "peerDependencies": {
+    "eslint": "^8"
+  },
+  "license": "SEE LICENSE.md",
+  "engines": {
+    "node": ">=14.0.0"
+  }
```
benelan, line 22: Can you copy the volta section from CC into this package.json too

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,31 @@
...
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: npm install, build, and test
+        run: |
+          npm install
+          npm run build --if-present
+          npm test
+        env:
+          CI: true
```
benelan, line 31: Is the matrix testing required for the eslint plugin? We already have the `pr-e2e.yml` action that runs tests on PRs, and then we have `deploy-next.yml` that runs the tests on pushes.

If this is only supposed to run on the eslint plugin it would need to use the [`--workspace` flag](https://docs.npmjs.com/cli/v8/using-npm/workspaces?v=true) and [filter via path](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore). 

As is, both actions will run tests for all packages that have a `test` npm script. Turbo's remote caching is used to make sure tests only run on the packages that were changed, but we haven't set that up yet since afaik it's a paid service.

benelan: We could also rename the eslint plugin's `test` npm script so it doesn't get executed by turbo's `test` pipeline.

.github/workflows/pr-tests_eslint-plugin-calcite-components.yml
```diff
@@ -0,0 +1,35 @@
...
+    steps:
+      - uses: actions/checkout@v1
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: npm install, build, and test
+        run: |
+          npm install
+          npm run build --if-present
```
benelan, line 32: You can add the `--workspace=packages/eslint-plugin-calcite-components` flag here too, unless building CC and CCR is required for the eslint plugin's tests. That should fix the error in the action's error. 

Also we don't need the `--if-present` flag when running `build` from the root, because turbo will only run a package's `build` script if it is present.

benelan (approved):
One more small change that should fix the GH Action's error, then LGTM! My guess is the error is due to using stencil in Node 14x, but not sure without looking into it further.

Also, reminder to create the git tag and github release after installing the PR following the [steps in the monorepo doc](https://github.com/Esri/calcite-components/wiki/Monorepo#adding-new-packages).

Issue #7202, opened 2023-06-20
calcite-list selectedItem incorrect when unselecting all via code


jgrayson-apl:
### Actual Behavior

the calciteList.selectedItem property is incorrect when we unselect all list items via code

### Expected Behavior

I would expect the CalciteList.selectedItem property to be an empty list if there are no selected list items 

### Reproduction Sample

https://codepen.io/john-grayson/pen/poQbMpe

### Reproduction Steps

1. select items in list to verify in the console that selecting/unselecting works correctly if interacting directly with list via list items
2. click the top 'Clear Selection' button
3. notice that the selectedItem property still contains the previous selected items
4. click button again to verify that the selectedItem property is still incorrect

### Reproduction Version

1.4.2

### Priority impact

p4 - not time sensitive

### Esri team

N/A

driskull:
Nice find @jgrayson-apl 👍

geospatialem:
Verified in `1.5.0-next.4` without the workaround of `list.selectedItems = [];`

PR #7204, opened 2023-06-20
fix(list): update selectedItems property on all item selection changes

driskull:
**Related Issue:** #7202

## Summary

- Updates `selectedItems` when an item's selection is toggled.
- Fixes current logic which depends on value being `true`.
- Adds test coverage

### DIFF
--- a/packages/calcite-components/src/components/list/list.tsx
+++ b/packages/calcite-components/src/components/list/list.tsx
@@ -185,11 +185,13 @@ export class List implements InteractiveComponent, LoadableComponent {
     const target = event.target as HTMLCalciteListItemElement;
     const { listItems, selectionMode } = this;
 
-    listItems.forEach((listItem) => {
-      if (selectionMode === "single" || selectionMode === "single-persist") {
-        listItem.selected = listItem === target;
-      }
-    });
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/list/list.tsx
```diff
@@ -185,11 +185,13 @@ export class List implements InteractiveComponent, LoadableComponent {
     const target = event.target as HTMLCalciteListItemElement;
     const { listItems, selectionMode } = this;
 
-    listItems.forEach((listItem) => {
-      if (selectionMode === "single" || selectionMode === "single-persist") {
-        listItem.selected = listItem === target;
-      }
-    });
+    if (target.selected) {
```
jcfranco, line 188: Should this also check if the selection mode is not `single` nor `single-persist`? When so, the `forEach` loop goes through all items without applying any changes.

jcfranco (approved):
💥

packages/calcite-components/src/components/list/list.tsx
driskull: yeah nice catch 🎣

PR #7206, opened 2023-06-21
ci: set design complete label conditionals

geospatialem:
**Related Issue:** n/a

## Summary
Updates the design complete workflow for conditions, in the event labels are not present.

In this case, will continue to run if the `design` and/or `1 - assigned` label(s) are not included in the issue. 🤖

### DIFF
--- a/.github/workflows/design-complete.yml
+++ b/.github/workflows/design-complete.yml
@@ -40,17 +40,25 @@ jobs:
 
               /* Modify labels */
 
-              // Remove "Design" label
-              await github.rest.issues.removeLabel({
-                ...issueProps,
-                name: "design"
-              });
+              // Try to remove "Design" label
+              try {
+                await github.rest.issues.removeLabel({
+                  ...issueProps,
+                  name: "design"
+                });
+              } catch(err) {
+                console.log(err);
+              }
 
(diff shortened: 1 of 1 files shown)


.github/workflows/design-complete.yml
```diff
@@ -40,17 +40,25 @@ jobs:
...
-                name: "design"
-              });
+              // Try to remove "Design" label
+              try {
+                await github.rest.issues.removeLabel({
+                  ...issueProps,
+                  name: "design"
+                });
+              } catch(err) {
+                console.log(err);
```
jcfranco, line 50: Can you add an additional message for context in addition to the caught error?

This also applies to the assigned label removal below.

.github/workflows/design-complete.yml
```diff
@@ -40,17 +40,25 @@ jobs:
 
               /* Modify labels */
 
-              // Remove "Design" label
-              await github.rest.issues.removeLabel({
-                ...issueProps,
-                name: "design"
-              });
+              // Try to remove "Design" label
```
jcfranco, line 43: Does this (and the comment below) need some info on why it may fail?

jcfranco (approved):
⚡🤖⚡

Issue #6854, opened 2023-04-25
Input type number does not work with French (F) AZERTY keyboard layout


ChristopheSuter:
### Actual Behavior

Problem when using calcite input type number with French (France) keyboard layout AZERTY on keyboards without numeric pad.

The user needs to hold the [SHIFT] key to enter numbers and this does not work (no input token) on calcite input but does work on native input type number

### Expected Behavior

Input should work as native input

### Reproduction Sample

https://codepen.io/geospatialem/pen/QWZpepg

### Reproduction Steps

Change your keyboard layout to French (France) and try to input number :
https://codepen.io/geospatialem/pen/QWZpepg

### Reproduction Version

1.2.0

### Relevant Info

MacOS/Chrome

### Priority impact

p1 - need for current milestone

### Impact

User can not use input

### Esri team

N/A

geospatialem:
This was also reported as BUG-000157826.

geospatialem:
possibly related #7039

eriklharper:
@ChristopheSuter what platform and browser are you running that allows you to type the non-number pad numeric keys in the native html number input?  I'm on macOS Ventura 13.4 and this the behavior I see when I type 1234567890 on the non-number pad keys in the following browsers:

Safari:
![image](https://github.com/Esri/calcite-components/assets/821864/cce4c235-8493-4ef9-b7fd-580a39299167)

Firefox
![image](https://github.com/Esri/calcite-components/assets/821864/bc6fdac3-6154-482a-8a88-573d3a4cbe5d)

Chrome
![image](https://github.com/Esri/calcite-components/assets/821864/b7e28842-a5d5-4011-a364-4a5ba9f6f973)

eriklharper:
This is the keyboard I've selected in the system settings input sources option 
![image](https://github.com/Esri/calcite-components/assets/821864/7452fea5-7350-4f32-965d-16a9da209742)

ChristopheSuter:
Ventura 13.3.1 and Chrome
Make sure it is the only layout in your settings ..
<img width="733" alt="image" src="https://github.com/Esri/calcite-components/assets/78811924/14c0a00d-93a8-4e69-82d2-9cbb534c741c">

<img width="790" alt="image" src="https://github.com/Esri/calcite-components/assets/78811924/9cdaaf21-2513-4e6b-83d4-35814db10d0b">

eriklharper:
[...]

PR #7213, opened 2023-06-22
fix(input, input-number): allows numeric characters.

anveshmekala:
**Related Issue:** #6854 

## Summary

This PR will support French AZERTY keyboard layout in `calcite-input` & `calcite-input-number`

### DIFF
--- a/packages/calcite-components/src/components/input-number/input-number.e2e.ts
+++ b/packages/calcite-components/src/components/input-number/input-number.e2e.ts
@@ -899,7 +899,7 @@ describe("calcite-input-number", () => {
     expect(Number(await element.getProperty("value"))).toBe(195);
   });
 
-  it("disallows typing characters with shift modifier key down", async () => {
+  it("disallows typing non-numeric characters with shift modifier key down", async () => {
     const page = await newE2EPage();
     await page.setContent(html`<calcite-input-number></calcite-input-number>`);
     const calciteInput = await page.find("calcite-input-number");
@@ -915,8 +915,7 @@ describe("calcite-input-number", () => {
     }
   });
 
-  // refer to issue here https://github.com/Esri/calcite-components/issues/6854
-  it("allows typing numbers with shift modifier key down", async () => {
+  it("allows typing numeric characters with shift modifier key down (#6854)", async () => {
     const page = await newE2EPage();
     await page.setContent(html`<calcite-input type="number"></calcite-input>`);
     const calciteInput = await page.find("calcite-input");
--- a/packages/calcite-components/src/components/input-number/input-number.e2e.ts
+++ b/packages/calcite-components/src/components/input-number/input-number.e2e.ts
@@ -13,7 +13,7 @@ import {
   t9n
 } from "../../tests/commonTests";
 import { getElementXY, selectText } from "../../tests/utils";
-import { letterKeys } from "../../utils/key";
+import { letterKeys, numberKeys } from "../../utils/key";
 import { locales, numberStringFormatter } from "../../utils/locale";
 
 describe("calcite-input-number", () => {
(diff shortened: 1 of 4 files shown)


packages/calcite-components/src/components/input-number/input-number.e2e.ts
```diff
@@ -923,6 +915,26 @@ describe("calcite-input-number", () => {
     }
   });
 
+  // refer to issue here https://github.com/Esri/calcite-components/issues/6854
```
jcfranco, line 918: Can you add the issue number in the test name instead? Similar to https://github.com/Esri/calcite-components/blob/14036c1354419222bc5b839af8b68675b17ebf62/packages/calcite-components/src/components/tree/tree.e2e.ts#L962.

packages/calcite-components/src/components/input-number/input-number.tsx
```diff
@@ -675,6 +675,11 @@ export class InputNumber
       return;
     }
 
+    // identifies French AZERTY keyboard input
```
jcfranco, line 678: Can you generalize this comment? I think this would apply to other keyboard layouts. Bonus points if you use a self-documenting const for this. ⭐

Actually, do we need to care about the Shift key being pressed? Wouldn't we want to process any number from `KeyboardEvent.key`? I tracked that part of logic down to https://github.com/Esri/calcite-components/issues/2127. @eriklharper Do you recall why we bail when the Shift key is pressed? This could possibly extend to the Alt, Ctrl and Meta keys.

packages/calcite-components/src/components/input/input.e2e.ts
```diff
@@ -1055,20 +1055,12 @@ describe("calcite-input", () => {
       expect(Number(await element.getProperty("value"))).toBe(195);
     });
 
-    it("disallows typing any letter or number with shift modifier key down", async () => {
+    it("disallows typing any characters with shift modifier key down", async () => {
```
jcfranco, line 1058: Would `non-numeric characters` be more precise for this?

Applies to other test names.

packages/calcite-components/src/components/input-number/input-number.tsx
anveshmekala: The reason for verifying Shift key here is to make sure the numbers are being returned. In AZERTY keyboard , users need to keep shift key activated for numbers to be printed.  

Ideally we would like to process any number but as you mentioned there is a custom logic that prevents numberKeys, ArrowKeys and Meta keys with Tab+Shift which results in the bug #6854.  I wonder if there will be a case where the `event.key` will return  a numberKey when Shift+Tab or Shift key is activated. 

Either we need to remove `numberKeys` from supportedKeys array or remove the check for `shiftKey`. With input and input-number , i am very hesitant to remove any existing code since it is very impractical to iterate tests for every single locale we support.

jcfranco: > The reason for verifying Shift key here is to make sure the numbers are being returned. In AZERTY keyboard , users need to keep shift key activated for numbers to be printed.

>  I wonder if there will be a case where the event.key will return a numberKey when Shift+Tab or Shift key is activated.

This is what I'm saying. In AZERTY, when the key combination for a number is pressed, you get a number in `KeyboardEvent.key`. 

> or remove the check for shiftKey. 

Unless I'm missing something, I think this is the way to go (`Shift` + `Tab` being the exception).

jcfranco (approved):
⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️
⌨️🇫🇷🇫🇷🇫🇷🇫🇷⌨️⌨️🇫🇷🇫🇷⌨️⌨️⌨️🇫🇷🇫🇷⌨️⌨️🇫🇷⌨️⌨️⌨️🇫🇷⌨️🇫🇷⌨️
⌨️🇫🇷⌨️⌨️🇫🇷⌨️🇫🇷⌨️⌨️🇫🇷⌨️🇫🇷⌨️⌨️🇫🇷⌨️🇫🇷🇫🇷⌨️🇫🇷🇫🇷⌨️🇫🇷⌨️
⌨️🇫🇷🇫🇷🇫🇷⌨️⌨️🇫🇷⌨️⌨️🇫🇷⌨️🇫🇷⌨️⌨️🇫🇷⌨️🇫🇷⌨️🇫🇷⌨️🇫🇷⌨️🇫🇷⌨️
⌨️🇫🇷⌨️⌨️🇫🇷⌨️🇫🇷⌨️⌨️🇫🇷⌨️🇫🇷⌨️⌨️🇫🇷⌨️🇫🇷⌨️⌨️⌨️🇫🇷⌨️⌨️⌨️
⌨️🇫🇷🇫🇷🇫🇷🇫🇷⌨️⌨️🇫🇷🇫🇷⌨️⌨️⌨️🇫🇷🇫🇷⌨️⌨️🇫🇷⌨️⌨️⌨️🇫🇷⌨️🇫🇷⌨️
⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️⌨️

jcfranco:
@anveshmekala I think the PR title can be generalized since this now applies to other keyboard layouts besides AZERTY.

Issue #6143, opened 2022-12-21
[Tooling] Generate translation reference JSON during build


macandcheese:
### Summary

The documentation site uses the `docs-json.json` file in the `extras` folder of our distributable to format and display docs to users. We would benefit from having a similar output for the updated translation files. 

`docs-json.json` includes each file's `messageOverride` string names, but not the default values, which would be valuable to show in documentation. It would be great if these were included as well.

Alternatively, and maybe easier than fighting jsdoc, we could generate our own `translations-json.json` at build time, since the files seem to be in a predictable location and format: (`/dist/assets/<component>/t9n`).

### Desired Outcome

Something like this, but any format works as long as the default are available somehow.

```
{
    "components": {
        "calcite-action": {
            "loading": {
                "en": "Close",
                "es": "Cerrar"
                ...
            },
            "disabled": {
                "en": "Close",
                "es": "Cerrar"
                ...
            }
        },
        "calcite-alert": {
            "loading": {
                "en": "Close",
                "es": "Cerrar"
                ...
            },
            "disabled": {
                "en": "Close",
                "es": "Cerrar"
                ...
            }
        }
    }
}
```

jcfranco:
@macandcheese I think we could generate something for this quite easily. What's the desired timeline for this?

macandcheese:
As soon as reasonable without a rush - not high priority. 

Whenever it is available I can add it into documentation site. Currently we only show the English string.

benelan:
Yeah I can whip something up after the patch period and other v1 stuff quiets down

jcfranco:
@benelan Were you able to whip something up for this? We modified the milestone, but can move it sooner if it's feasible.

benelan:
I haven't gotten to this yet, there are some higher priority items on my plate

geospatialem:
Verified with `1.4.3` in the calcite-components-examples repo.

PR #7214, opened 2023-06-23
build(t9n): generate json file containing t9n values

benelan:
**Related Issue:** #6143

## Summary

Generates `dist/extras/translations-json.json` which contains the t9n values per component in the following format:

```jsonc
{
    "components": {
        "calcite-action": {
            "loading": {
                "en": "Close",
                "es": "Cerrar"
                // ...
            },
            "disabled": {
                "en": "Close",
                "es": "Cerrar"
                // ...
            }
        },
        "calcite-alert": {
            "loading": {
                "en": "Close",
                "es": "Cerrar"
               // ...
            },
            "disabled": {
                "en": "Close",
                "es": "Cerrar"
                // ...
            }
        }
    }
}
```


packages/calcite-components/support/generateT9nDocsJSON.ts
```diff
@@ -0,0 +1,52 @@
...
+    const components = await readdir(assetsPaths);
+
+    const data = {};
+    const messagesFilenameRegex = /messages_(.*)\.json/;
+
+    for (const component of components) {
+      const t9nPath = resolve(assetsPaths, component, "t9n");
+      if (existsSync(t9nPath)) {
+        data[component] = {};
+        // the english file is the only one without the lang in the filename
```
jcfranco, line 23: English does have a corresponding bundle (see https://github.com/Esri/calcite-components/blob/master/packages/calcite-components/src/components/action/assets/action/t9n/messages_en.json). The one without a language code is the main one.

Could you update the regex pattern to match bundles with or without a language code?

packages/calcite-components/package.json
```diff
@@ -19,7 +19,7 @@
   ],
   "scripts": {
     "build": "npm run util:prep-build-reqs && stencil build",
-    "postbuild": "npm run util:patch && git restore src/components/*/readme.md",
+    "postbuild": "npm run util:patch && npm run util:generate-t9n-docs-json && git restore src/components/*/readme.md || true",
```
jcfranco, line 22: Is `util:generate-t9n-docs-json` throwing unnecessarily or is there another reason for the fallback `true`?

jcfranco (changes requested):
Looking good! 😎

packages/calcite-components/package.json
benelan: Woops that was left over from my local testing. The docker container I use isn't a git repo so `git restore` fails. This is my workaround but I forgot to change it back, fixing!

packages/calcite-components/support/generateT9nDocsJSON.ts
benelan: Oh gotcha. What should I be doing with the main one then, does it need to be included?  And if so what should I call the key, "main"? cc @macandcheese

benelan: It seems like `messages.json` and `messages_en.json` are always the same, but I created a separate field called `main` for `messages.json` anyway.

jcfranco: Yeah, in most cases they will be the same. WDYT about omitting `main` for now until we have a strong reason to include it? Sorry, this would mean reverting the regex update I suggested earlier. 😅

jcfranco (approved):
✨🔧🧰✨ Awesome stuff!

Issue #6575, opened 2023-03-07
[Tree] expanded tree item not showing children


rpanichakit9541:
### Actual Behavior

If the container of the calcite-tree starts out as hidden (display: none), when the container is toggled to be visible the children under an expanded parent are invisible.

### Expected Behavior

The children calcite-tree-item under expanded parent should not be invisible.

### Reproduction Sample

https://codepen.io/rpanichakit/pen/eYLGJGw

### Reproduction Steps

1. Go to [the codepen](https://codepen.io/rpanichakit/pen/eYLGJGw)
2. Click on "Show tree"
3. Observe that the children under "Parent 1" are invisible even though it's expanded
4. If you click the icon to toggle close/open "Parent 1" you'll see the children then show up correctly

### Reproduction Version

1.0.8

### Relevant Info

Windows 10, Chrome Version 110.0.5481.178

### Regression?

v1.0.0-beta.99

### Priority impact

p2 - want for current milestone

### Impact

Currently, we're using the tree to display the menu navigation in the new metadata editor and this bug make the tree nav look broken to the user.

### Esri team

ArcGIS Online

geospatialem:
Research will be conducted as part of the June milestone for next steps.

nsulzberger:
I have a related issue. If a parent node is expanded, I remove the children and add them again later, the children are not displayed, only an empty space like in the example above.

geospatialem:
Verified on `master`.

PR #7216, opened 2023-06-23
fix(tree-item): ensure expanded tree-item is displayed when expanded and made visible

jcfranco:
**Related Issue:** #6575 

## Summary

This updates the styling on tree items to ensure they are rendered regardless of its expand/collapse transition.

**Note:** 7209f9bc9fa5aaad361532409bc129400643f459 removes `OpenCloseComponent` implementation since the hooks no longer did anything. We can reapply it once we add corresponding events to tree items.

Release-As: 1.4.3


benelan (approved):
:+1:

Issue #6604, opened 2023-03-15
OpenCloseComponents do not emit when setting `--calcite-duration-factor`


jcfranco:
### Actual Behavior

Setting `--calcite-duration-factor` to 0 allows users to skip animations, but prevents (before)open/close events from being emitted.

cc @Elijbet 

### Expected Behavior

All (before)open/close events to be emitted regardless of transition duration.

### Reproduction Sample

https://codepen.io/jcfranco/pen/WNgJemJ?editors=1010

### Reproduction Steps

Click on toggle and notice how events are not emitted in the console.

### Reproduction Version

1.0.8

### Priority impact

p3 - not time sensitive

### Impact

This will impact any users explicitly setting `--calcite-duration-factor` to 0 that rely on the before+open/close events.

### Esri team

Calcite (dev)

geospatialem:
Workaround: https://codepen.io/jcfranco/pen/gOdeVpq?editors=1000 (Context: https://github.com/Esri/calcite-components/issues/6582#issuecomment-1469253860)

jcfranco:
This should be fixed by https://github.com/Esri/calcite-design-system/pull/7258.

geospatialem:
Verified in `1.5.0-next.38`

Issue #6495, opened 2023-02-17
Input date picker - calendar doesn't adjust to programmatically specified date


jlias-devtester:
### Actual Behavior

In ArcGIS Online, it is possible to get the date picker in a state where you cannot select the start and end dates you need

### Expected Behavior

Regardless of how you select a start and end date, the calendar picker should update to show the correct month that you're trying to select.

### Reproduction Sample

https://codepen.io/rpanichakit/pen/RwYPYvq

### Reproduction Steps

1. Prepare a clean ENU browser.
2. ArcGIS online
3. Switch to Organization page and click Status tab.
4. Navigation to Reports and click Create Single report.
5. Expand Report type and select Activity.
Scenario 1:
6. Navigation to Month and expand end date calendar.
7. Select the previous month of current month. (e.g. Select January in 2023 since current month is February) -key step
8. Navigation to Quarter and expand end date calendar.
9. Click > Next icon on calendar and observe the result.
10. Expand start date calendar.
11. Click < Previous icon on calendar and observe the result.
Scenario 2:
12. Navigate to Week and expand start date calendar after step 5.
13. Select certain date which is in current quarter. (e.g. Select 1/1/2023 date since current quarter is from January to March)-key step
14. Navigation to Quarter and expand end date calendar.
15. Click > Next icon on calendar and observe the result.
16. Expand start date calendar.
17. Click < Previous icon on calendar and observe the result.

Actual Result:
1. It is skipped from January 2023 to December 2022 after clicking > Next icon on calendar in step 9 and step 15.
2. It is skipped from December 2022 to September 2022 after clicking < Previous icon on calendar in step 11 and step 17.

### Reproduction Version

not sure

### Regression?

I do not believe so as this is reproducible in production, so it's been a bug since at least November 2022.

### Esri team

ArcGIS Online

geospatialem:
Related https://github.com/Esri/calcite-components/issues/6469

geospatialem:
Verified on `1.5.0-next.8`

![verify-list-filter](https://github.com/Esri/calcite-components/assets/5023024/07fe3334-3ed4-418d-98a1-6febc6561b42)

PR #7219, opened 2023-06-24
fix(input-date-picker): reset active date picker date after closing

jcfranco:
**Related Issue:** #6495 

## Summary

This ensures the active date is reset properly after a `calcite-input-date-picker`'s date picker is closed (either by blurring or selecting a date).

## Notes

* This adds an internal `reset` method that can be made public in the future if needed. 
* There was an existing `reset` method prior to these changes that seems to clear the active date when the date picker is blurred or when `Escape` is pressed. We should revisit this behavior since it was [added when the input and date picker were a single component](https://github.com/Esri/calcite-components/blob/v1.0.0-beta.22/src/components/calcite-date/calcite-date.tsx#L263-L270) and it does seem odd to have the date-picker reset in these scenarios. cc @macandcheese @SkyeSeitz @ashetland 
* Utility test methods to interact with internal components were added and tests were updated to leverage them
* Slightly increases the duration factor in the tests to work around #6604

### DIFF
--- a/packages/calcite-components/src/components/input-date-picker/input-date-picker.e2e.ts
+++ b/packages/calcite-components/src/components/input-date-picker/input-date-picker.e2e.ts
@@ -57,6 +57,64 @@ describe("calcite-input-date-picker", () => {
 
   it.skip("supports t9n", () => t9n("calcite-input-date-picker"));
 
+  async function navigateMonth(page: E2EPage, direction: "previous" | "next"): Promise<void> {
+    const linkIndex = direction === "previous" ? 0 : 1;
+
+    await page.evaluate(
+      async (MONTH_HEADER_CSS, linkIndex: number): Promise<void> =>
+        document
+          .querySelector("calcite-input-date-picker")
+          .shadowRoot.querySelector("calcite-date-picker")
+          .shadowRoot.querySelector("calcite-date-picker-month-header")
+          .shadowRoot.querySelectorAll<HTMLAnchorElement>(`.${MONTH_HEADER_CSS.chevron}`)
+          [linkIndex].click(),
+      MONTH_HEADER_CSS,
+      linkIndex
+    );
+    await page.waitForChanges();
+  }
+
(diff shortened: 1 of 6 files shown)


packages/calcite-components/src/components/input-date-picker/input-date-picker.e2e.ts
```diff
@@ -294,38 +345,30 @@ describe("calcite-input-date-picker", () => {
...
-              .shadowRoot.querySelector("calcite-input")
-              .shadowRoot.querySelector("input").value
-        );
 
       await page.keyboard.press("Tab");
       await page.keyboard.type("1/");
       await page.waitForChanges();
 
-      expect(await getInputValue()).toBe("١‏/");
+      expect(await getDateInputValue(page)).toBe("١‏/");
```
benelan, line 353: thanks for hanging that up to DRY!

packages/calcite-components/src/tests/utils.ts
```diff
@@ -268,7 +268,7 @@ export async function newProgrammaticE2EPage(): Promise<E2EPage> {
  */
 export async function skipAnimations(page: E2EPage): Promise<void> {
   await page.addStyleTag({
-    content: `:root { --calcite-duration-factor: 0; }`
+    content: `:root { --calcite-duration-factor: 0.01; }`
```
benelan, line 271: Can you explain why changing this to 0.1 is required, is it a chromatic thing?

benelan (approved):
LGTM

packages/calcite-components/src/tests/utils.ts
jcfranco: Sure. It's to work around [this issue](https://github.com/Esri/calcite-components/issues/6604).

Good catch though. I'll add a comment to this for context.

jcfranco:
Extracted the restoring of missing focus assertions to a separate PR: https://github.com/Esri/calcite-components/pull/7235

PR #7222, opened 2023-06-26
ci(release-please): pin action version and allow manually running action

benelan:
**Related Issue:** #

## Summary
The release-please action randomly started failing 3 days ago with this [unhelpful error](https://github.com/Esri/calcite-components/actions/runs/5360304923/jobs/9725004712#step:2:234). It looks like they [released 3 days ago](https://github.com/google-github-actions/release-please-action/releases/tag/v3.7.10), so I'm pinning the version to the one before that and hoping it will fix it. 

I also added the [`workflow_dispatch`](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/) trigger so we can run the action manually.

PR #7225, opened 2023-06-26
ci: hardcode release version due to reverting feat

benelan:
**Related Issue:** #7222

## Summary
Pinning the `release-please` Github Action version in the PR above fixed the error, so their release must have broken something for us. `release-please` is trying to release `v1.5.0` instead of `v1.4.3` because we installed a feat and subsequently reverted it, but our current version is still `1.5.0-next.5`. 

[The logs](https://github.com/Esri/calcite-components/actions/runs/5383995298/jobs/9771404018#step:2:102) say it's skipping the conventional commit versioning strategy and forcing `1.5.0`. I'm hoping I can force `1.4.3` instead, as described in their [manifest doc](https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md#configfile). If this doesn't work I may need to change the package.json versions back to `1.4.3-next.7`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -43876,7 +43876,7 @@
     },
     "packages/calcite-components": {
       "name": "@esri/calcite-components",
-      "version": "1.5.0-next.5",
+      "version": "1.4.3-next.7",
       "license": "SEE LICENSE.md",
       "dependencies": {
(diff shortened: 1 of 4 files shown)


benelan:
pushing this thru cc @jcfranco

PR #7227, opened 2023-06-27
ci: double build cc when publishing to workaround stencil types bug

benelan:
**Related Issue:** #

## Summary

Add a `prepublishOnly` npm script hook to calcite-components, which builds and makes sure the types aren't broken. In addition to the Stencil bug, Calcite Components also needs to be built after versioning so that the preamble is correct. It has been [a version behind in `next` releases](https://unpkg.com/browse/@esri/calcite-components@1.5.0-next.5/dist/calcite/calcite.esm.js). 

ref: https://github.com/lerna/lerna/tree/main/libs/commands/publish#lifecycle-scripts


jcfranco (approved):
👍

Could we wrap this in an external script to add some context?

Issue #6485, opened 2023-02-15
Block: too many loading indicators


ashetland:
### Actual Behavior

Block's header has two loading indicators. 
![image](https://user-images.githubusercontent.com/108549080/219140725-96af0fc1-9614-44be-a8cd-bc5c466e0e60.png)

### Expected Behavior

One loading indicator, either on the left or the right. Prefer the inline calcite-loader for consistency.

### Reproduction Sample

https://codepen.io/ashetland/pen/GRBbOXK

### Reproduction Steps

1. Open codepen.
2. Collapse and expand Blocks to see loading behavior.

### Reproduction Version

1.0.5

### Relevant Info

cc @alan

### Esri team

Calcite (design)

macandcheese:
Cc @asangma / @mitc7862 for insight / historical context / need for MV

ashetland:
After a quick chat with @asangma, propose we use the calcite-loader for consistency, but use the icon slot like before. This would reduce adding more items to the right side.

![image](https://user-images.githubusercontent.com/108549080/219192591-632c063a-bc24-4471-924c-b8585f4117ae.png)

geospatialem:
Moving back to development per https://github.com/Esri/calcite-design-system/pull/7239#issuecomment-1618721579 and https://github.com/Esri/calcite-design-system/pull/7239#issuecomment-1619163107

PR #7239, opened 2023-06-28
fix(block): removes extra loading indicator

Elijbet:
**Related Issue:** #6485

## Summary

`calcite-block header` had two loading indicators. This removes the extra `calcite-loader` from the right end, and places it as an option in `icon-slot`,`statusIcons: valid or invalid` as well as `controls` now appear regardless of `loading` state.


driskull (approved):
👍

packages/calcite-components/src/components/block/block.tsx
```diff
@@ -321,9 +322,7 @@ export class Block
         ) : (
           headerContent
         )}
-        {loading ? (
-          <calcite-loader inline label={messages.loading} />
-        ) : hasControl ? (
+        {hasControl && !loading ? (
```
jcfranco, line 325: I _think_ `loading` doesn't affect this anymore since we want it always rendered (if a control is provided).

packages/calcite-components/src/components/block/block.tsx
```diff
@@ -237,16 +237,17 @@ export class Block
...
 
     const hasIcon = getSlotted(el, SLOTS.icon) || statusIcon;
 
     const iconEl = !statusIcon ? (
       <slot key="icon-slot" name={SLOTS.icon} />
+    ) : loading ? (
+      <calcite-loader inline label={messages.loading} />
     ) : (
       <calcite-icon
         class={{
```
jcfranco, line 253: I think `[CSS.loading]: showingLoadingStatus`, and its corresponding `CSS` prop and styles, need to go as well since it's no longer applicable (we have the single loader now to indicate loading).

jcfranco (changes requested):

packages/calcite-components/src/components/block/block.tsx
Elijbet: Ok, it seemed like in the original code the icons rendered in the absence of loading: looks like they were built as else-if conditional ternary chains. Both for status (valid/invalid) and controls.

packages/calcite-components/src/components/block/resources.ts
```diff
@@ -35,6 +34,5 @@ export const ICONS = {
   opened: "chevron-up",
   closed: "chevron-down",
   valid: "check-circle",
-  invalid: "exclamation-mark-triangle",
-  refresh: "refresh"
```
jcfranco, line 39: ✨🧹✨

jcfranco (approved):
Awesome!

geospatialem:
It looks like the block's `icon` slot isn't getting overridden in this case per https://github.com/Esri/calcite-components/issues/6485#issuecomment-1432121944 in the updated [sample](https://codepen.io/geospatialem/pen/poQPgNG). 

@Elijbet Is this the expected structure and behavior code-wise? cc @ashetland

jcfranco:
@geospatialem I noticed the sample above uses `1.5.0-next.8`, could you try `next` or `1.5.0-next.9` instead?

geospatialem:
Ah, great catch, Franco!

geospatialem:
It does look like this mitigates the multiple loaders, but there still seems to be a missing loader when the component is collapsed (not `open`) with `1.5.0-next.9`: https://codepen.io/geospatialem/pen/poQPgNG.

Observe the first example does not have a loading indicator, but the `open` one does.

Would the `icon` slot need to account for both `loading` and when the component is not `open`?

Elijbet:
Cool, I'll follow up on this issue with the last comments. @ashetland @geospatialem

PR #7240, opened 2023-06-28
refactor(floating-ui): apply composed getOffsetParent consistently

jcfranco:
**Related Issue:** N/A

## Summary

We no longer need to check for Chromium 109+ before applying `getOffsetParent` ponyfill since our supported Chrome versions are now 113 & 114.

**Note**: this also removes internal config flag to bypass the custom logic to handle positioning across shadow DOM (no longer an issue as the ponyfill does not impact performance).


driskull (approved):
Nice! 💥

PR #7243, opened 2023-06-29
test(input): stabilize increment tests

benelan:
## Summary

Stabilizes a few of the unstable input tests brought up on Teams.


packages/calcite-components/src/components/input/input.e2e.ts
```diff
@@ -252,7 +252,9 @@ describe("calcite-input", () => {
       expect(await element.getProperty("value")).toBe("100000000000000000000000000000000000000000000000010");
       element.setProperty("step", 0.1);
       await page.waitForChanges();
-      Array.from({ length: 10 }, async () => await numberHorizontalItemDown.click());
+      for (let i = 0; i < 10; i++) {
+        numberHorizontalItemDown.click();
```
jcfranco, line 256: Would awaiting on all the clicks in parallel achieve the same and keep everything chained? 

```ts
await Promise.all(
  Array.from({ length: 10 }, () => numberHorizontalItemDown.click())
);
```

jcfranco: Actually, it looks like the test doesn't need these to run in parallel and we can click in sequence, so we could await on each one.

jcfranco (approved):

jcfranco:
Approving and merging in the name of stability. 🧪

Issue #6329, opened 2023-01-24
Tooltip label acts like a description and obscures tooltip content


nwhittaker:
### Summary
A tooltip's label property becomes the textual representation of the tooltip. This means that string is what is applied back to the reference element via its aria-describedby attribute. However, a label and description are two different things:

…a label should be concise, while a description is intended to provide more verbose information; a label describes the essence of an object, while a description provides more information that the user might need.[1](https://github.com/Esri/calcite-components/issues/6329#user-content-fn-1-7bfafa444cda74d8385b1634b8e6d646)

As a result, when a tooltip is given a concise label, that label becomes the description of the reference element and the browser no longer has an accessible way to expose/apply the actual text content of the tooltip.

### Actual Behavior
From the API, it's unclear that the tooltip's label property is actually behaving as the description of the reference element. If the fix is to duplicate the tooltip's text content as its label, it's unclear why the label property is required since aria-describedby falls back to an element's text content anyway.

### Expected Behavior
Possibly rename the label property to a less overloaded term (e.g. "description"?) and not making it required. Otherwise, improve doc to better explain what makes for a good label.

### Reproduction Sample
https://developers.arcgis.com/calcite-design-system/components/tooltip/

### Reproduction Steps
Visit the tooltip doc page.
Use devtools to inspect the accessibility tree of the code sample's reference element.
Notice the reference element's description is the same as the tooltip's label and does not include any of the tooltip's text content.
<img width="824" alt="Calcite QA" src="https://user-images.githubusercontent.com/3362490/218839096-9a509b3c-0e7c-4188-b376-a5f89ec8249d.png">

### Reproduction Version
next.731

### Working W3C Example/Tutorial
https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/

### Relevant Info
It looks like a working example is still in development, but I didn't see any examples in the current work where the tooltip element had an aria-label attribute, FWIW.

### Regression?
No response

### Esri team
ArcGIS Field Apps

Footnotes
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby#description [↩](https://github.com/Esri/calcite-components/issues/6329#user-content-fnref-1-7bfafa444cda74d8385b1634b8e6d646)

brittneytewks:
Removed "needs triage" since a priority is set and estimate is indicated

geospatialem:
Additional research needed on prop requirements and a11y supported.

[...]

PR #7247, opened 2023-06-29
fix(tooltip): deprecate the label property due to the description coming from the component's content

driskull:
**Related Issue:** #6329

## Summary

- Deprecates the `label` property.
  - The already provided `aria-describedby` attribute is sufficient for accessibility.
  - Setting aria attributes on the component is not recommended.
- Can still be set by user so no breaking changes.
- Set to optional by default so its no longer required.
- Update tests
- Update Storybook
- Update HTML

### DIFF
--- a/packages/calcite-components/src/components/tooltip/tooltip.tsx
+++ b/packages/calcite-components/src/components/tooltip/tooltip.tsx
@@ -59,7 +59,7 @@ export class Tooltip implements FloatingUIComponent, OpenCloseComponent {
   /**
    * Accessible name for the component.
    *
-   * @deprecated No longer necessary.
+   * @deprecated No longer necessary due to the label coming from the component's content. Setting aria attributes on the component is not recommended.
    */
   @Prop() label: string;
 
--- a/packages/calcite-components/src/components/tooltip/tooltip.tsx
+++ b/packages/calcite-components/src/components/tooltip/tooltip.tsx
@@ -59,7 +59,7 @@ export class Tooltip implements FloatingUIComponent, OpenCloseComponent {
   /**
    * Accessible name for the component.
    *
-   * @deprecated No longer necessary due to the label coming from the component's content. Setting aria attributes on the component is not recommended.
+   * @deprecated This is no longer necessary because the component's content provides the description.
    */
   @Prop() label: string;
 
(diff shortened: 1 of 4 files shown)


packages/calcite-components/src/components/tooltip/tooltip.tsx
```diff
@@ -56,8 +56,12 @@ export class Tooltip implements FloatingUIComponent, OpenCloseComponent {
   /** Closes the component when the `referenceElement` is clicked. */
   @Prop({ reflect: true }) closeOnClick = false;
 
-  /** Accessible name for the component. */
-  @Prop() label!: string;
+  /**
+   * Accessible name for the component.
+   *
+   * @deprecated No longer necessary.
```
jcfranco, line 62: Can we add some info on this no longer being needed because the label will come from the content/setting ARIA props on it not being recommended?

jcfranco:
Could you update the description to include a summary of why the property is now optional & deprecated? Had to track it down in the linked issue.

jcfranco (approved):
🚀

packages/calcite-components/src/components/tooltip/tooltip.tsx
jcfranco: From the changelog perspective, could you also revisit the PR title to include something about this?

driskull:
@jcfranco @geospatialem can you review again?

packages/calcite-components/src/components/tooltip/tooltip.tsx
jcfranco: WDYT about:

```
@deprecated The component's content provides the description, so this should no longer necessary.
```

?

Deferring to @geospatialem for some docspertise™.

jcfranco:
Just had a comment about some extra info in the `@deprecated` tag. Other than that, this LGTM! 🚀

packages/calcite-components/src/components/tooltip/tooltip.tsx
driskull: How about..

> @deprecated This is no longer necessary because the component's content provides the description.

I like "is" better than "should" and this ordering. Should sounds less confident :D

geospatialem: WDYT about some context as to what the `label` would provide if used? Something like:

> @deprecated No longer necessary. Overrides the context of the component's description, which could confuse assistive technology users.

driskull: sounds good to me

PR #7248, opened 2023-06-29
test(input): replaces input-number with input component

anveshmekala:
**Related Issue:** # N/A

## Summary

This will replace  any instance of `calcite-input-number` used in `calcite-input` E2E tests.


jcfranco (approved):
OMG, great catch! Can't believe we missed this. 😵

anveshmekala:
> OMG, great catch! Can't believe we missed this. 😵

One has to do similar mistake to catch those lol.

Issue #6400, opened 2023-01-31
flow-item back button focus issues and tooltip at wrong location


AdelheidF:
### Actual Behavior

![image](https://user-images.githubusercontent.com/44481292/215876929-ed5b2c4b-0ca9-4aeb-a6eb-0be267e93ace.png)

After creating the second flow-item I want to have the focus set on the back button.

Try 1:
`flowItem2.setFocus();`
-> doesn't set the focus on back button

Try 2: 
`requestAnimationFrame(() => flowItem2.setFocus());`
-> doesn't set the focus on back button
-> this method worked for me with beta.99, but I can't repro it via the codepen
  
Try 3:
`setTimeout(() => flowItem2.setFocus(), 100);`
-> sets the focus on back button, but tooltip is somewhere else

Try 4:
`setTimeout(() => flowItem2.setFocus(), 1000);`
-> sets the focus on back button, with tooltip on it, but it takes too long to wait for the correct focus

### Expected Behavior

1. requestAnimation should set the focus; no random timeout required
2. tooltip is always next to the back button

### Reproduction Sample

https://codepen.io/afreitag/pen/WNKaJww

### Reproduction Steps

Open codepen and click on blue button.

### Reproduction Version

1.0.3

### Regression?

beta.98 for tooltip placement
beta.99 for using requestAnimation 

### Impact

high

### Esri team

ArcGIS Map Viewer

macandcheese:
Not 100% the same but similar enough to add - I have encountered this when a Tooltip referencing an Action is clicked and the containing Shell Panel has a Panel open as a result of that interaction. The previous Action tooltip will linger at the extent of the Shell Panel container: 

<img width="485" alt="Screen Shot 2023-01-31 at 8 50 02 PM" src="https://user-images.githubusercontent.com/4733155/215926126-3f72ce50-b9de-4276-8b7e-133f7d2e1c58.png">

I use `setFocus()` (without timeout) on the Panel, and when the Panel closes, to return focus to the Action. It's a bit unpredictable, as it works most of the time but you can occasionally see it while navigating via the Action Bar items in the component sample on the documentation site.

jcfranco:
FWIW, I narrowed down the regression to [this commit](https://github.com/Esri/calcite-components/commit/f490e5ee0a4ae75f0e3b727f4ce0f7925bc8e53c). I'll keep digging after I wrap some issues we've got queued for our patch release.

jcfranco:
[...]

PR #7252, opened 2023-06-30
feat(flow): Adds setFocus method

driskull:
**Related Issue:** #6400 #5369

## Summary

- Depends on #7255
- Adds `setFocus` method.
- Implements `LoadableComponent`.
- Add test

## Gif

issue-example


jcfranco:
`Calls setFocus when the back button is clicked.` is a separate feature from adding `setFocus`. Is there a PR title that could work to cover both from a higher level? Otherwise, we should split this up into the following to adhere to conventional commits:

* adding `setFocus` to `flow`
* focus `flow-item` content when backing into a `flow`

I'd be more than happy to help with the PR splitting.

jcfranco (commented):
Looking 🔥!

jcfranco (approved):
Awesome! 🔥

PR #7255, opened 2023-06-30
fix(action): correctly focus the button after rendering updates.

driskull:
**Related Issue:** #6400 #5369

## Summary

- `forceUpdate()` is called in `setFocus()` in order to make sure the internal button has been rendered.
- The internal button is then focused after `requestAnimationFrame()` to ensure its been rendered.


packages/calcite-components/src/components/action/action.tsx
```diff
@@ -206,8 +206,15 @@ export class Action
...
+    forceUpdate(this);
 
-    this.buttonEl?.focus();
+    const { buttonEl } = this;
+
+    if (!buttonEl) {
+      return;
+    }
+
+    requestAnimationFrame(() => buttonEl.focus());
```
driskull, line 217: I'm wondering if we should update all components that call `focus()` on a child element within the component to also forceUpdate and requestAnimationFrame?

Components that call `setFocus()` on other components wouldn't need to do this.

thoughts @jcfranco ?

eriklharper: I'm curious why the `forceUpdate` is needed instead of just checking if the `buttonEl` is defined?

driskull: I believe its necessary to render the component before trying to focus the button element in cases where the action was just created and setFocus was called on it.

eriklharper: If the action was just created, wouldn't that mean it had already rendered it though?

eriklharper: When you say "created" that just means an instance of `CalciteActionButton` was made, not that it was appended to the DOM, right?

driskull: Nope. Not from what I understand.

```js
const action = document.createElement("calcite-action");
document.body.appendChild(action);
action.setFocus(); // not guaranteed anything within action is rendered at this time.
```

eriklharper: Interesting.  What way could you guarantee that `action` is fully rendered before calling setFocus, I wonder?

driskull: Yeah I think forcing an update and then calling requestAnimationFrame seems to do the trick. There may be a better way? @jcfranco 

We could explore having a flag like `componentRendered = true;` set in `componentDidRender()`

driskull: Ok i tried that but it didn't work.

```patch
diff --git forkSrcPrefix/packages/calcite-components/src/utils/renderable.ts forkDstPrefix/packages/calcite-components/src/utils/renderable.ts
new file mode 100644
index 0000000000000000000000000000000000000000..85bbcea923f563daabdbcd92d0cb008605078ce7
--- /dev/null
+++ forkDstPrefix/packages/calcite-components/src/utils/renderable.ts
@@ -0,0 +1,20 @@
+export interface RenderableComponent {
+  componentWillLoad: () => Promise<void> | void;
+  componentDidRender: () => Promise<void> | void;
+}
+
+const resolveMap = new WeakMap<RenderableComponent, (value: void | PromiseLike<void>) => void>();
+
+const promiseMap = new WeakMap<RenderableComponent, Promise<void>>();
+
+export function setUpRenderableComponent(component: RenderableComponent): void {
+  promiseMap.set(component, new Promise((resolve) => resolveMap.set(component, resolve)));
+}
+
+export function setComponentRendered(component: RenderableComponent): void {
+  resolveMap.get(component)();
+}
+
+export function componentRendered(component: RenderableComponent): Promise<void> {
+  return promiseMap.get(component);
+}
diff --git forkSrcPrefix/packages/calcite-components/src/components/action/action.tsx forkDstPrefix/packages/calcite-components/src/components/action/action.tsx
index f11cf45717c5dbee5234c9d77b5cc2df015fb46b..8c06a56e0b6419f64b52bdcc147850c62473d706 100644
--- forkSrcPrefix/packages/calcite-components/src/components/action/action.tsx
+++ forkDstPrefix/packages/calcite-components/src/components/action/action.tsx
@@ -25,6 +25,12 @@ import {
   setComponentLoaded,
   setUpLoadableComponent
+import {
+  componentRendered,
+  RenderableComponent,
+  setComponentRendered,
+  setUpRenderableComponent
+} from "../../utils/renderable";
 import { connectLocalized, disconnectLocalized, LocalizedComponent } from "../../utils/locale";
 import { createObserver } from "../../utils/observers";
 import {
@@ -48,7 +54,12 @@ import { CSS, SLOTS } from "./resources";
   assetsDirs: ["assets"]
 })
 export class Action
-  implements InteractiveComponent, LocalizedComponent, T9nComponent, LoadableComponent
+  implements
+    InteractiveComponent,
+    LocalizedComponent,
+    T9nComponent,
+    LoadableComponent,
+    RenderableComponent
 {
   //
@@ -176,6 +187,7 @@ export class Action
 
     setUpLoadableComponent(this);
+    setUpRenderableComponent(this);
     if (Build.isBrowser) {
       await setUpMessages(this);
     }
@@ -185,6 +197,11 @@ export class Action
     setComponentLoaded(this);
   }
 
+  componentDidRender(): void {
+    setComponentRendered(this);
+    updateHostInteraction(this);
+  }
+
     disconnectInteractive(this);
     disconnectLocalized(this);
@@ -192,10 +209,6 @@ export class Action
   }
 
-  componentDidRender(): void {
-    updateHostInteraction(this);
-  }
-
   //
   //  Methods
@@ -206,7 +219,8 @@ export class Action
   @Method()
     await componentLoaded(this);
-    forceUpdate(this);
+    //forceUpdate(this);
+    await componentRendered(this);
 
     const { buttonEl } = this;
 
@@ -214,7 +228,9 @@ export class Action
       return;
     }
 
-    requestAnimationFrame(() => buttonEl.focus());
+    buttonEl.focus();
+
+    //requestAnimationFrame(() => buttonEl.focus());
   }
 
```

jcfranco: Refs should be set by the time `componentDidLoad` is called, but we may need util if this is a lifecycle bug. A few questions:

* Does this happen for both the `dist` and `components` output targets?
* In the action snippet above, is an error thrown when `setFocus` is called or does it silently fail? 
* Would we only want to ensure focus is called properly if the element is attached to the DOM? This would match native behavior.
* `Ok i tried that but it didn't work.` Is this referring to the `forceUpdate` + `rAF` technique or the proposed pattern above?

FWIW, `forceUpdate` schedules a render phase, but does not guarantee immediate rendering as it's not synchronous and up to the component's rendering (see https://stenciljs.com/docs/api#other).

driskull: > Refs should be set by the time componentDidLoad is called, but we may need util if this is a lifecycle bug. A few questions:

The refs are there, however calling `focus()` on the element has no effect. This is why I'm forcing an update and waiting for the requestAnimationFrame.

> In the action snippet above, is an error thrown when setFocus is called or does it silently fail?

Silently fail. The element is just not focused.

> Would we only want to ensure focus is called properly if the element is attached to the DOM? This would match native behavior.

Potentially. I'm not sure how we would achieve this. Ideally, when setFocus is called, we want to ensure an element within the component gets focused. Currently, this PR Is what I found works.

> Ok i tried that but it didn't work. Is this referring to the forceUpdate + rAF technique or the proposed pattern above?

Its referring to setting up a promise and resolving it on `componentDidRender()`. This didn't work for focusing correctly.

driskull: It looks like it is added to the DOM, the focus just isn't happening without the forceUpdate and requestAnimationFrame.

jcfranco: Yeah, that makes sense. I recall `focus` not working in cases where `focus` was called right after an element is appended, but before the browser renders it. 

Taking all your findings into consideration, do you think the following could work?

* Add a `setFocus(component: LoadableComponent): Promise<void>)` (or similarly named) util that resolves after the following:
* waits for `component` to resolve, then
* calls `forceUpdate` and then
* calls `focusElement` in `requestAnimationFrame`'s callback?

For `setFocus` methods that rely on internal components, these could simply return their `setFocus()` call and not need the util, right?

driskull: Yeah I was thinking the same thing as a util that handles the forceUpdate and returns after requestAnimationFrame. 👍

driskull: @jcfranco something like what I just checked in?

packages/calcite-components/src/utils/loadable.ts
```diff
@@ -111,3 +113,26 @@ export function setComponentLoaded(component: LoadableComponent): void {
 export function componentLoaded(component: LoadableComponent): Promise<void> {
   return promiseMap.get(component);
 }
+
+/**
+ * This helper util can be used to ensure a component is ready to be focused (The "componentDidLoad" stencil lifecycle method has been called and any internal elements are ready to be focused).
```
jcfranco, line 118: Nitpick: `stencil` => `Stencil`

packages/calcite-components/src/utils/loadable.ts
```diff
@@ -111,3 +113,26 @@ export function setComponentLoaded(component: LoadableComponent): void {
...
+ *
+ * Requires "setUpLoadableComponent" and "setComponentLoaded" to be called first.
```
jcfranco, line 120: WDYT about simplifying this to ``requires `LoadableComponent` to be implemented`` (similar to [`T9nComponent`'s doc](https://github.com/Esri/calcite-components/blob/14036c1354419222bc5b839af8b68675b17ebf62/packages/calcite-components/src/utils/t9n.ts#L105)).

packages/calcite-components/src/utils/loadable.ts
jcfranco: Could you rephrase this part to state that this helps focusing by ensuring the component is loaded and rendered by the browser?

jcfranco (approved):
Awesome! We'll need to add a test helper and spec test for this as well. Could you open up a follow-up issue?

jcfranco:
> We'll need to add a test helper and spec test for this as well.

Actually, maybe we can update our `focusable` helper instead of adding a new one.

driskull:
@jcfranco would we want another issue to update all other setFocus methods that `.focus()` an internal element as well?

This shouldn't be necessary if a `setFocus` method calls another `.setFocus()`.

driskull:
Created issue: https://github.com/Esri/calcite-design-system/issues/7276

driskull:
It might just be good to update all `setFocus()` methods to use this new helper instead of `componentLoaded()`. Just to make sure things are working well.

jcfranco:
Agreed! Thanks for creating the follow-up issue. 🎉

PR #7256, opened 2023-06-30
refactor(flow-item): setFocus method should return setFocus calls to child components

driskull:
**Related Issue:** #7270

## Summary

- Updates setFocus method within flow-item to return calls to child components setFocus method.


jcfranco (approved):
Nice! If this is fixing something, can we rename the commit type, and add a test (if feasible) to reflect this?

This is a good idea. We should look at following this pattern consistently. I'll create an issue for this.

driskull:
Its not fixing any issue, just something I noticed.

jcfranco:
Gotcha. I wasn't sure from the branch name. 

Issue created: https://github.com/Esri/calcite-design-system/issues/7270

Issue #6279, opened 2023-01-12
Tooltip `beforeClose` and `close` events sometimes not emitted


nwhittaker:
### Actual Behavior

If a tooltip is open inside a container when a `display: none` style is set on the container, the tooltip never emits `calciteTooltipBeforeClose` and `calciteTooltipClose` events.

### Expected Behavior

If a tooltip is open inside a container when a `display: none` style is set on the container, the tooltip still emits `calciteTooltipBeforeClose` and `calciteTooltipClose` events.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/bGjRNEW

### Reproduction Steps

1. Visit the sample and open the console.
2. Mouse into the red-bordered box and see the "Hover for Tooltip" text appear.
3. Mouse onto the "Hover for Tooltip" text and wait for the tooltip to open.
4. Quickly mouse out of the red-bordered box and wait a few seconds. Notice that none of the tooltip's close events are logged to the console.

### Reproduction Version

next.721

### Relevant Info

IIRC, `transitionstart` and `transitionend` events don't fire if the element doesn't have display which may explain why the tooltip isn't emitting its close events.

### Impact

For performance reasons, the Field Maps web-app has some logic that remove a tooltip after it closes. In some cases, those tooltips appear alongside buttons that gain/lose display depending on whether a containing element is being hovered.

The current solution is to wait a hard-coded amount of time before removing the tooltip, but waiting for the event would be cleaner and easier to maintain.

### Esri team

ArcGIS Field Apps

geospatialem:
Related `modal` issues to `beforeClose` to possibly investigate together: #6407, #6381, #6379

geospatialem:
Verified in the `main` branch.
![verify-tooltip-events](https://github.com/Esri/calcite-design-system/assets/5023024/670e13c1-0adb-47b4-8b88-b2aecc6b3e80)

PR #7258, opened 2023-06-30
fix(tooltip): emits `close` and `beforeClose` events when container is set to `display:none`

Elijbet:
**Related Issue:** #6279

## Summary
The `transitionend` event is fired when a transition has been completed. In the case where a transition is removed before completion, such as if the transition property is removed or `display` is set to `none`, then the event will not be generated.

This PR implements a more recent implementation of the `openCloseComponent interface` with the `onToggleOpenCloseComponent` function import, modified to emit in all cases where there is no transition happening.

An alternate solution is to call getComputedStyle():
`if (transitionDuration === "0s" || window.getComputedStyle(component.el).visibility === "hidden")`

For context, here is what the difference why it works with `visibility: hidden`, but not `display: none`:

`visibility: hidden`: the element is not visible but still takes up space in the layout. The element is rendered with transparent content, and it can still undergo opacity transitions. 
See it working here: https://codepen.io/elijbet/pen/wvQdyxw?editors=1111

`display: none`: the element is not rendered at all, and it does not take up any space in the layout. This means that the element cannot undergo any transitions, including opacity transitions. Since the element is not present in the layout, transition events related to `opacity`, such as `transitionend`, will not be fired.

### DIFF
--- a/packages/calcite-components/src/utils/openCloseComponent.ts
+++ b/packages/calcite-components/src/utils/openCloseComponent.ts
@@ -77,18 +77,16 @@ export function onToggleOpenCloseComponent(component: OpenCloseComponent, nonOpe
   readTask((): void => {
     if (component.transitionEl) {
       const allTransitionPropsArray = getComputedStyle(component.transitionEl).transition.split(" ");
+      const visibility = getComputedStyle(component.transitionEl).visibility;
+      const display = getComputedStyle(component.transitionEl).display;
+      console.log("visibility", visibility);
+      console.log("display", display);
+
       const openTransitionPropIndex = allTransitionPropsArray.findIndex(
         (item) => item === component.openTransitionProp
       );
-      const transitionDuration = allTransitionPropsArray[openTransitionPropIndex + 1];
-      if (transitionDuration === "0s") {
-        (nonOpenCloseComponent ? component[component.transitionProp] : component.open)
-          ? component.onBeforeOpen()
-          : component.onBeforeClose();
-        (nonOpenCloseComponent ? component[component.transitionProp] : component.open)
-          ? component.onOpen()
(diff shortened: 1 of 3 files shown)


driskull:
Since `connectOpenCloseComponent` and `disconnectOpenCloseComponent` are outdated we should update the doc for them and mark them as `@deprecated` as well. Just so nobody else tries to implement them in another component

jcfranco:
It might be more scalable and perf-conscious to listen to `transitioncancel` than looking at the computed `display` value to determine the open/close 'finale' 🎩🪄. Could you give that a try?

Elijbet:
> It might be more scalable and perf-conscious to listen to `transitioncancel` than looking at the computed `display` value to determine the open/close 'finale' 🎩🪄. Could you give that a try?

Hm.. I think getting computed display was just a console.log for demo purposes.

packages/calcite-components/src/utils/openCloseComponent.ts
```diff
@@ -70,25 +70,30 @@ function transitionEnd(event: TransitionEvent): void {
  * Helper to determine globally set transition duration on the given openTransitionProp, which is imported and set in the @Watch("open").
  * Used to emit (before)open/close events both for when the opacity transition is present and when there is none (transition-duration is set to 0).
  *
+ * @example
+ * async componentWillLoad() {
+ * // When componenet initially renders, if `open` was set we need to trigger on load as watcher doesn't fire.
```
jcfranco, line 75: Typo: `componenet` ➡️ `component`

Elijbet:
This seems to be working after the #7422 fix is implemented.

jcfranco (approved):
🙈🗣️

Issue #7268, opened 2023-07-03
Unstable input tests


driskull:
### Test type

e2e test

### Which Component(s)

input-number, input-time-picker

### Test error, if applicable

```
@esri/calcite-components:test: Summary of all failing tests
@esri/calcite-components:test: FAIL src/components/input-time-picker/input-time-picker.e2e.ts (177.907 s)
@esri/calcite-components:test:   ● calcite-input-time-picker › correctly relocalizes the display value when the lang and numbering systems change
@esri/calcite-components:test: 
@esri/calcite-components:test:     expect(received).toBe(expected) // Object.is equality
@esri/calcite-components:test: 
@esri/calcite-components:test:     Expected: "02:30:25 م"
@esri/calcite-components:test:     Received: "14.30.25"
@esri/calcite-components:test: 
@esri/calcite-components:test:       642 |     await page.waitForChanges();
@esri/calcite-components:test:       643 |
@esri/calcite-components:test:     > 644 |     expect(await getInputValue(page)).toBe("02:30:25 م");
@esri/calcite-components:test:           |                                       ^
@esri/calcite-components:test:       645 |
@esri/calcite-components:test:       646 |     inputTimePicker.setProperty("numberingSystem", "arab");
@esri/calcite-components:test:       647 |     await page.waitForChanges();
@esri/calcite-components:test: 
@esri/calcite-components:test:       at Object.<anonymous> (src/components/input-time-picker/input-time-picker.e2e.ts:644:39)
@esri/calcite-components:test:           at runMicrotasks (<anonymous>)
@esri/calcite-components:test: 
@esri/calcite-components:test: FAIL src/components/input-number/input-number.e2e.ts (978.91 s)
@esri/calcite-components:test:   ● calcite-input-number › increment/decrement functionality › correctly increments/decrements numbers greater than MAX_SAFE_INTEGER
@esri/calcite-components:test: 
@esri/calcite-components:test:     expect(received).toBe(expected) // Object.is equality
@esri/calcite-components:test: 
@esri/calcite-components:test:     Expected: "100000000000000000000000000000000000000000000000009"
@esri/calcite-components:test:     Received: "100000000000000000000000000000000000000000000000009.4"
@esri/calcite-components:test: 
@esri/calcite-components:test:       231 |       Array.from({ length: 10 }, async () => await numberHorizontalItemDown.click());
@esri/calcite-components:test:       232 |       await page.waitForChanges();
[...]

PR #7269, opened 2023-07-03
test(input-number, input-time-picker): Skip unstable tests

driskull:
**Related Issue:** #7268

## Summary

test(input-number, input-time-picker): Skip unstable tests


eriklharper:
Do we have any insight as to the root cause of these unstable tests, whether these are component bugs that are fixable via `requestAnimationFrame` (or something related) or if these are test environment bugs?   Should we reach out to Stencil support on these kinds of topics?

Issue #7254, opened 2023-06-30
Keyboard navigation issue in calcite-list


nid13240:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-components/issues) to avoid duplicates

### Actual Behavior

When using custom logic to enable/disable items in the calcite-list, items which were previously disabled and reenabled get skipped during keyboard navigation.

### Expected Behavior

I would expect to be able to tab through all the items, even if they were previously disabled and reenabled.

### Reproduction Sample

https://codepen.io/nid13240/pen/VwVmzpx

### Reproduction Steps

1. open the codepen above - the logic to enable/disable items is there to mimc some logic we have in the Map Viewer
2. select the first item in the list with your keyboard (the next 2 items should become disabled)
3. deselect the first item in the list (those items should become reenabled)
4. try to tab through the rest of the list
5. notice the tab skips those 2 items which were previously disabled

### Reproduction Version

1.4.3

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Map Viewer

geospatialem:
It looks like this one is mitigated when using the arrow keys. 👍🏻 

However, keyboard a11y is not mitigated with the `tab` or `tab` + `shift` keys for navigating, where `list-item` slotted actions are still skipped with https://codepen.io/geospatialem/pen/XWyzpXX. @driskull Is the expectation to also account for `tab` and `shift` + `tab`? If not, I can update the doc.


![verify-list-keyboard-a11y](https://github.com/Esri/calcite-design-system/assets/5023024/32cbad00-5bcf-4760-aafa-42d6c760af13)

driskull:
@geospatialem this might actually be unrelated. Its something that the `interactive` helper is doing. Can we get a separate issue for @jcfranco?

The issue is that the tabindex is being set to -1 but not being removed. @jcfranco I don't think the tabindex should be set at all since this component is set to "managed".

geospatialem:
New issue created via #7335, closing the above out.

PR #7275, opened 2023-07-03
fix(list): fix keyboard navigation after a listItem's disabled or closed property changes

driskull:
**Related Issue:** #7254

## Summary

- Emits internal event when `closed` or `disabled` changes on a list item.
- This internal event triggers the parent list to update its navigable items.
- Adds test for keyboard navigation

### DIFF
--- a/packages/calcite-components/src/components/list-item/list-item.tsx
+++ b/packages/calcite-components/src/components/list-item/list-item.tsx
@@ -88,7 +88,7 @@ export class ListItem
 
   @Watch("closed")
   handleClosedChange(): void {
-    this.emitcalciteInternalListItemChange();
+    this.emitCalciteInternalListItemChange();
   }
 
   /**
@@ -103,7 +103,7 @@ export class ListItem
 
   @Watch("disabled")
   handleDisabledChange(): void {
-    this.emitcalciteInternalListItemChange();
+    this.emitCalciteInternalListItemChange();
   }
 
   /**
@@ -557,7 +557,7 @@ export class ListItem
   //
   // --------------------------------------------------------------------------
 
-  private emitcalciteInternalListItemChange(): void {
+  private emitCalciteInternalListItemChange(): void {
     this.calciteInternalListItemChange.emit();
   }
 
--- a/packages/calcite-components/src/components/list-item/list-item.tsx
+++ b/packages/calcite-components/src/components/list-item/list-item.tsx
@@ -86,6 +86,11 @@ export class ListItem
   /** When `true`, hides the component. */
   @Prop({ reflect: true, mutable: true }) closed = false;
 
+  @Watch("closed")
+  handleClosedChange(): void {
+    this.emitcalciteInternalListItemChange();
+  }
+
   /**
    * A description for the component. Displays below the label text.
    */
@@ -95,6 +100,11 @@ export class ListItem
    * When `true`, interaction is prevented and the component is displayed with lower opacity.
    */
   @Prop({ reflect: true }) disabled = false;
+
+  @Watch("disabled")
+  handleDisabledChange(): void {
+    this.emitcalciteInternalListItemChange();
+  }
 
   /**
    * The label text of the component. Displays above the description text.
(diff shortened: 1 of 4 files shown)


packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -541,6 +557,10 @@ export class ListItem
   //
   // --------------------------------------------------------------------------
 
+  private emitcalciteInternalListItemChange(): void {
```
jcfranco, line 560: `emitcalcite...` ➡️ `emitCalcite...`

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -200,8 +200,8 @@ export class List implements InteractiveComponent, LoadableComponent {
     this.updateSelectedItems();
   }
 
-  @Listen("calciteListItemClose")
-  handleCalciteListItemClose(): void {
+  @Listen("calciteInternalListItemChange")
+  handlecalciteInternalListItemChange(): void {
```
jcfranco, line 204: `handlecalcite...` ➡️ `handleCalcite...`

jcfranco: Per our conventions, the internal event needs to stop propagating when handled.

packages/calcite-components/src/components/list/list.e2e.ts
```diff
@@ -339,4 +339,87 @@ describe("calcite-list", () => {
     expect(await listItemOne.getProperty("selected")).toBe(false);
     expect(await list.getProperty("selectedItems")).toHaveLength(0);
   });
+
+  describe.only("keyboard navigation", () => {
```
jcfranco, line 343: Can you remove `only` here? The linter should've caught this. 🤔

packages/calcite-components/src/components/list/list.e2e.ts
```diff
@@ -339,4 +339,87 @@ describe("calcite-list", () => {
...
+    async function evaluateActiveElement(page: E2EPage, activeElementSelector: string): Promise<void> {
```
jcfranco, line 344: I think you could use the [`isElementFocused` test util](https://github.com/Esri/calcite-components/blob/91b2a1b92d49cdd573650952ee09971c59bd1649/packages/calcite-components/src/tests/utils.ts#L291-L290) instead.

packages/calcite-components/src/components/list/list.e2e.ts
```diff
@@ -339,4 +339,87 @@ describe("calcite-list", () => {
...
+      expect(await page.evaluate((selector) => document.activeElement?.matches(selector), activeElementSelector)).toBe(
+        true
+      );
+    }
+
+    it("should navigate via ArrowUp, ArrowDown, Home, and End", async () => {
+      const page = await newE2EPage({
```
jcfranco, line 351: For consistency, can you create the page and then populate it with `setContent`?

packages/calcite-components/src/components/list/list.e2e.ts
```diff
@@ -339,4 +339,87 @@ describe("calcite-list", () => {
...
+              description="hello world"
+            ></calcite-list-item>
+          </calcite-list>
+        `
+      });
+      await page.waitForChanges();
+      const list = await page.find("calcite-list");
+      await list.callMethod("setFocus");
+      await page.waitForChanges();
+      await page.waitForTimeout(0);
```
jcfranco, line 378: Will the extra timeout still be needed after the focus-ensuring util is applied?

jcfranco (approved):
Nice!

packages/calcite-components/src/components/list/list.e2e.ts
driskull: eek. Hmm my linter isn't catching this. No bueno!

Maybe because its on the describe???

packages/calcite-components/src/components/list/list.e2e.ts
driskull: Nope removed.

packages/calcite-components/src/components/list/list.e2e.ts
driskull: Hmmm nope. No linting going on with `.only` for me. Does it work for you?

jcfranco: Weird. Looks like the linter does catch it, but the commit hook isn't being blocked. Something going on with our `lint-staged` configuration? cc @benelan

PR #7277, opened 2023-07-03
feat: improve focus behavior in components

driskull:
**Related Issue:** None

## Summary

- Updates all `setFocus()` methods to use `componentFocusable()` instead of `componentLoaded()`.
- `TimePicker` component also updates in the `focusPart` private method to use `componentFocusable()`.
- `TextArea` uses `componentLoaded()` outside of the `setFocus()` method. This was in the `selectText()` and `resizeObserver`. I left those them alone since they are not doing any focusing specifically.
- Added `await page.waitForChanges();` to tests that required it.
  - action-menu
  - split-button
  - commonTests helper (removed setTimeout of 0 as well)


jcfranco (approved):
This is awesome. Thanks for updating all components! 🚀

If these changes improve focus behavior, should this be marked as `feat: improve focus behavior` (or similar) instead?

PR #7285, opened 2023-07-05
fix(flow): Call setFocus() on back button click

driskull:
**Related Issue:** None

## Summary

- Depends on #7252
- When the back button is clicked, `setFocus()` is called and the newly active panel is focused.
- Adds test


packages/calcite-components/src/components/flow/flow.e2e.ts
```diff
@@ -55,6 +56,24 @@ describe("calcite-flow", () => {
...
+    it("should call setFocus() on back button click", async () => {
+      const page = await newE2EPage();
+
+      await page.setContent(
+        html`<calcite-flow
+          ><calcite-flow-item id="one"></calcite-flow-item><calcite-flow-item id="two"></calcite-flow-item
+        ></calcite-flow>`
+      );
+
+      await page.$eval("#two", (elm: HTMLCalciteFlowItemElement) =>
```
jcfranco, line 68: In order to remove implementation details, can we click the back button vs firing the custom event?

packages/calcite-components/src/components/flow/flow.e2e.ts
```diff
@@ -55,6 +56,24 @@ describe("calcite-flow", () => {
...
+        elm.dispatchEvent(new CustomEvent("calciteFlowItemBack"))
+      );
+      await page.waitForTimeout(0);
```
jcfranco, line 71: Curious, do we still need the timeout delay after the `setFocus` changes? If so, maybe we could add a new util (separately) to ensure we wait consistently for `setFocus`-related tests.

jcfranco (approved):
Sweet, thanks for splitting this up! 🪵🪓

packages/calcite-components/src/components/flow/flow.e2e.ts
driskull: yup!

packages/calcite-components/src/components/flow/flow.e2e.ts
driskull: Seems to not be necessary here. Ill remove.

Issue #7270, opened 2023-07-03
Ensure all `setFocus` calls return `setFocus` from internal components


jcfranco:
### Description

For consistency, components implementing `setFocus` that end up calling `setFocus` on internal, supporting components should return the promise from this method call.

cc @driskull 

### Proposed Advantages

All `setFocus` calls are chained properly.

### Which Component

Any component implementing `setFocus` whose method calls `setFocus` or the `focusElement` helper.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

geospatialem:
Verified on `master`.

PR #7287, opened 2023-07-06
refactor: Ensure all setFocus methods return calls to setFocus

driskull:
**Related Issue:** #7270

## Summary

For consistency, components implementing `setFocus` that end up calling `setFocus` on internal, supporting components should return the promise from this method call.

- Any component implementing `setFocus` whose method calls `setFocus` or the `focusElement` helper.
- depends on #7277


jcfranco (approved):
Awesome! 👓

I just noticed that we could also await vs return if that's simpler/more consistent, but I could go either way.

PR #7289, opened 2023-07-06
fix(card): ensure teardown logic is called when disconnected

jcfranco:
**Related Issue:** N/A

## Summary

This fixes a lifecycle hook typo that prevented teardown logic from running.


driskull (approved):
oh wow. I wonder if we can catch these somehow?

driskull:
Can we create a eslint plugin that ensures lifecycle method names are correct?

Something like this:
Something like this: https://github.com/angular-eslint/angular-eslint/blob/main/packages/eslint-plugin/docs/rules/use-lifecycle-interface.md

jcfranco:
A custom rule for this would be awesome, but it might be tricky to implement. Maybe we could start simple with a spell checker plugin? https://www.npmjs.com/package/@cspell/eslint-plugin looks promising and it would have definitely caught this. Might need some tweaking to avoid unnecessary noise.

jcfranco:
Depends on https://github.com/Esri/calcite-design-system/pull/7295 landing first. Cherry picked into this branch to get a head start on the E2E test run.

Issue #7122, opened 2023-06-03
Create additional packages/workspaces within the monorepo


benelan:
### Summary

The initial monorepo PR is installed! Once we ensure everything is running smoothly, we can start splitting out some of the code in `packages/calcite-components` into separate workspaces. [Turbo has an example project](https://github.com/vercel/turbo/tree/main/examples/design-system) where they have separate workspaces for eslint, tsconfig, and storybook. We could potentially split the demo pages into its own workspace as well.

In addition to creating workspaces from existing code in the monorepo, we can start moving in packages from other repos. Here is a list copied from #6532:

- [x] `@esri/calcite-components-react`
- [ ] `@esri/eslint-plugin-calcite-components`
- [ ] `@esri/calcite-design-tokens`
- [ ] `@esri/calcite-ui-icons` (potentially)
- [ ]  [framework examples](https://github.com/Esri/calcite-components-examples)
- [ ] shared eslint configuration (see turbo's example linked above)
- [ ] tsconfigs (see turbo's example linked above)
- [ ] storybook (see turbo's example linked above)
- [ ] component utils (future)
- [ ] test utils (future)
- [ ] demos (future)

### Desired Outcome

Adding more Calcite packages to the repo and splitting existing functionality into encapsulated workspaces will improve maintainability and discoverability.

driskull:
> @esri/calcite-ui-icons (potentially)

Seems like this would be good to be in the monorepo right? Anything calcite should be in this repo ideally.

benelan:
Yeah I think Franco put the "potentially" there before he talked with @SkyeSeitz and the other maintainers. I just copied it from his list in the issue I linked. But last I heard we are on track to move the icons to the monorepo.

SkyeSeitz:
Thanks, @benelan. Can you remind me of the impact so so I can confirm with the Graphic Illustration team on any updates to their workflow - if any? A url change? Any impacts on maintaining[ this site](https://esri.github.io/calcite-ui-icons/)?

benelan:
Yeah the code would be moved to this repo so they would need to get set up here with permissions etc. The github pages site would need to be moved as well

alisonailea:
I'd like to propose we create a /configs folder at root for all shared configs (typescript, jest, eslint, etc.) and a /tools folder for shared tooling (tokens-transformer)

benelan:
It would just be the shared configs for the design system. Here is an example of a shared eslint config from turbo: 
[...]

PR #7294, opened 2023-07-07
build: make sure eslint-plugin-calcite-components is set up correctly

benelan:
**Related Issue:** #

## Summary

I forgot to mention that the packages within the monorepo are "local dependencies", which means `npm install` won't install them, they have to be built locally. A common workflow is `cd`ing directly into calcite-components to do work. This caused issues with `eslint-plugin-calcite-components` because it wasn't being built so it wasn't working during development in CC. 

I added a `start` turbo pipeline which will ensure all dependencies are built first. So running `npm start` from the root directory will no longer cause this issue.

I also cleaned up the Contributing doc and tweaked some eslint settings that were causing issues in VSCode.


driskull (approved):
Awesome 👏

benelan:
@jcfranco I had to remove the on-exit cleanup of js demo files from `npm start` because it was breaking the turbo pipeline. I'm guessing turbo's concurrency was getting tangled up `concurrently` and the exit code watcher. 

Instead, I clean out all of the js files created by `tsc` before running `npm start` and make sure the files that were previously being cleaned are gitignored. 

Is that solution okay with you or should I spend more time troubleshooting the concurrency issue?

CONTRIBUTING.md
```diff
@@ -112,11 +112,11 @@ If your IDE supports the [Language Server Protocol (LSP) specification](https://
 
 ## Starting the demos
 
-First, clone the repo and install the NPM dependencies from within the `calcite-components` directory:
+First, clone the repo, install the NPM dependencies:
```
jcfranco, line 115: Nitpick: I feel like we need an `"and then"` between cloning and installing deps. I'll leave this up to you.

packages/eslint-plugin-calcite-components/.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "ignorePatterns": ["*"]
```
jcfranco, line 2: Any package with this configuration will skip linting entirely, right? If so, won't this make the contributing instructions on linting misleading?

jcfranco: For now, maybe the contributing doc can be updated to indicate that ESLint will only run on `calcite-components` and we can tweak our linting config to lint properly outside `calcite-components` after this lands?

jcfranco (approved):
Aside from a comment regarding linting on other packages, this LGTM! 🚀

packages/eslint-plugin-calcite-components/.eslintrc.json
benelan: Yeah I added this to CCR and the eslint plugin so eslint won't complain about there not being a config which covers their files.  

I'm working on pulling shared eslint/tsconfigs into separate workspaces as part of the lint-staged fix. After that happens, the base eslint config will be used in these packages. 

So I'm going to leave the doc as is for now so I don't have to change it and then change it back.

benelan:
Note: I couldn't get the tsconfig/eslint separating to work at the moment so I updated the doc. See https://github.com/Esri/calcite-design-system/issues/7122#issuecomment-1626229399

PR #7297, opened 2023-07-07
ci: allow lint-staged to find typescript parser config file

benelan:
**Related Issue:** N/A

## Summary

`lint-staged` correctly fails now when there are eslint errors in `packages/calcite-components`.


packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -918,7 +918,7 @@ async function getTagAndPage(componentTestSetup: ComponentTestSetup): Promise<Ta
  * });
  *
  * @param {ComponentTestSetup} componentTestSetup - A component tag, html, or the tag and e2e page for setting up a test.
- * @param {DisabledOptions} [options={ focusTarget: "host" }] - Disabled options.
```
jcfranco, line 921: This looks like an autofix, but the file itself looks unrelated to these changes. Let's exclude it from this PR and then do a full lint sweep afterwards.

packages/calcite-components/.eslintrc.cjs
```diff
@@ -1,23 +1,25 @@
-{
-  "env": {
+module.exports = {
```
jcfranco, line 1: I was going to originally suggest keeping this as `json` if it didn't break anything, but this will align perfectly with the changes we'll need to make for [future of configuring ESLint](https://eslint.org/docs/latest/use/configure/configuration-files-new). 🎉

jcfranco (approved):
Thanks for fixing this! 🛠️

packages/calcite-components/.eslintrc.cjs
benelan: it needs to be a cjs file so I can use `__dirname`

Issue #6630, opened 2023-03-22
[TextArea] Provide context to assistive technologies when character count is exceeded


geospatialem:
### Description

Provide context to assistive technology (AT) users when the `textarea` component's character count is exceeded, something like:

> Maximum character length is `${maxLength}`, current length is `${currentLength}`. 

The above will require support for a wide audience, including:
1. AT support, and
2. Translation support
    - The `maxLength` and `currentLength` variables will need to be replaced before reading back to users to ensure wording is translated in a grammatically acceptable manner across locales.

### Acceptance Criteria

1. Provide context to assistive technologies (e.g., JAWS, NVDA, VoiceOver) when `textarea`'s character count is exceeded
2. Include translation strings, which provide proper grammatical context across our supported locales

### Relevant Info

Pertains to the new component, `textarea` from PR https://github.com/Esri/calcite-components/pull/5644

### Which Component

`textarea`

### Example Use Case

https://codepen.io/geospatialem/pen/GRXwJKy

The solution could consider a `.sr-only`, which is not displayed visually on the page, but provides context to assistive technologies (AT).

### Priority impact

p1 - need for current milestone

### Esri team

Calcite (dev)

geospatialem:
Reallocating to the May release, work will be conducted after the nav component work.

geospatialem:
Current context is provided on form submission 👍🏻. 

Additional efforts will be sought for context as the user is interacting with the component directly, moving back to development to finalize the effort.

geospatialem:
Verified in `1.5.0-next.33` with JAWS in Chrome and NVDA in Firefox.

PR #7299, opened 2023-07-07
feat(text-area): provide additional context for AT users when character limit exceeds

anveshmekala:
**Related Issue:** #6630 

## Summary

This PR will provide additional context for Assistive Technology users when character limit exceeds.

After this change, an example error message looks like below:

`The current character length is 14, which exceeds the maximum character length of 10.`


packages/calcite-components/src/components/text-area/text-area.tsx
```diff
@@ -403,25 +403,37 @@ export class TextArea
...
-        <span class={{ [CSS.characterOverLimit]: this.value?.length > this.maxLength }}>
-          {this.getLocalizedCharacterLength()}
+    if (this.maxLength) {
+      const { currentLength, maxLength } = this.getLocalizedCharactersLength();
+      return (
+        <span class={CSS.characterLimit}>
+          <span class={{ [CSS.characterOverLimit]: this.value?.length > this.maxLength }}>
+            {currentLength}
+          </span>
+          {"/"}
```
jcfranco, line 413: Unrelated nitpick: We don't need to interpolate the string here, so can we use `/` directly or do you think it'd be confusing?

packages/calcite-components/src/components/text-area/text-area.tsx
```diff
@@ -403,25 +403,37 @@ export class TextArea
   };
 
   renderCharacterLimit = (): VNode => {
```
jcfranco, line 405: Can you return null when there's no max length and update the return type accordingly?

packages/calcite-components/src/components/text-area/text-area.tsx
```diff
@@ -403,25 +403,37 @@ export class TextArea
...
+        locale: this.effectiveLocale,
+        numberingSystem: this.numberingSystem,
+        signDisplay: "never",
+        useGrouping: this.groupSeparator
+      };
+      return {
+        currentLength: numberStringFormatter.localize(currentLength),
+        maxLength: numberStringFormatter.localize(maxLength)
+      };
+    } else {
```
jcfranco, line 434: Some readability suggestions:

* Invert the if conditions so that the shorter block comes up first.
* Since you're already returning above, you can drop the `else`.

```ts
if (this.numberingSystem === "latn") {
  return { currentLength, maxLength };
}

numberStringFormatter.numberFormatOptions = {
  locale: this.effectiveLocale,  
  numberingSystem: this.numberingSystem,
  signDisplay: "never",
  useGrouping: this.groupSeparator
};

return {
  currentLength: numberStringFormatter.localize(currentLength),
  maxLength: numberStringFormatter.localize(maxLength)
};
```

jcfranco (approved):
Aside from the comment to avoid computing localized characters twice, this LGTM!

packages/calcite-components/src/components/text-area/text-area.tsx
anveshmekala: i think its better to interpolate indicating it is part of the UI.

packages/calcite-components/src/components/text-area/text-area.tsx
anveshmekala: i think this needs to be update across the repo. Should we open an issue for this one?

Issue #6412, opened 2023-02-01
Combobox Item Group - add space after grouped items


ashetland:
### Description

It is currently very difficult to know where a `combobox-item-group` ends. Adding space at the bottom of grouped items will provide visual separation between the group and any following content. 

### Acceptance Criteria

- Bottom padding added to `combobox-item-group` at all scales.
- Added padding would not be present if the group has no items, allowing for the use of only the group heading (with no extra padding) when nesting groups.

### Relevant Info

Thank you to @Anastasiia-Boleiko for bringing this to our attention!

A design, with a possible final spec, is available in Figma.

### Which Component

Combobox Item Group

### Esri team

Calcite (dev)

Anastasiia-Boleiko:
Thanks a lot for filing this enhancement! It will be really helpful👍

ashetland:
Design spec available in this [Figma file](https://www.figma.com/file/IMJzSur0pnbTcw7RW719yf/Combobox-Group-Spacing-%5BIssues-%236412%2C-%236384%5D-%5Benhancement%5D?type=design&node-id=107-35777&t=fSD1StP4Bv6l8Pi5-11).

geospatialem:
Verified in the `main` branch:

![image](https://github.com/Esri/calcite-design-system/assets/5023024/82483f46-fb8c-4ff7-9bac-998d0a0c9582)

PR #7302, opened 2023-07-09
fix(combobox): add space after grouped items

Elijbet:
**Related Issue:** #6412

## Summary
Adding space at the bottom of grouped items to provide visual separation between the group and any following content.

- Bottom padding added to combobox-item-group at all scales.
- Added padding would not be present if the group has no items, allowing for the use of only the group heading (with no extra padding) when nesting groups.

CleanShot 2023-07-10 at 09 19 44@2x


packages/calcite-components/src/components/combobox-item-group/combobox-item-group.tsx
```diff
@@ -50,6 +50,24 @@ export class ComboboxItemGroup {
 
   scale: Scale = "m";
 
+  // --------------------------------------------------------------------------
+  //
+  //  Private Methods
+  //
+  // --------------------------------------------------------------------------
+
+  private isPaddingPresent(): boolean {
```
driskull, line 59: nitpick: can we come up with a better name for this private method? Something that describes what the function is testing for.

driskull (approved):
This looks good. Should `ancestors` be marked as `@internal` in doc? We don't want or need users to set ancestors right? cc @geospatialem @jcfranco

anveshmekala (approved):
🙌  

Other than naming everything looks great!

packages/calcite-components/src/components/combobox-item-group/combobox-item-group.scss
```diff
@@ -49,3 +49,8 @@
   padding-inline: var(--calcite-combobox-item-spacing-unit-s);
   margin-inline-start: var(--calcite-combobox-item-indent-value);
 }
+
+.trailingPadding {
```
anveshmekala, line 53: can we change this to `padding--trailing` or `list-padding--trailing` as per BEM™ ?

jcfranco:
> Should ancestors be marked as @internal in doc? We don't want or need users to set ancestors right?

Yes, it should be internal. Nice catch. Let's do this separately to confirm that using `@internal` only removes it from the doc (i.e., prop and associated type are unaffected). Otherwise, this would fall under breaking change.

packages/calcite-components/src/components/combobox-item-group/combobox-item-group.tsx
```diff
@@ -50,6 +50,24 @@ export class ComboboxItemGroup {
...
+    const groupEl = this.ancestors[this.ancestors.length - 1];
+    if (groupEl) {
```
jcfranco, line 61: If there's no group, this method returns `undefined` instead of a boolean.

packages/calcite-components/src/components/combobox-item-group/combobox-item-group.scss
Elijbet: Directly quoting a Reddit comment I received a couple of days ago: "You won't get in trouble for not including the ™ btw" :)

packages/calcite-components/src/components/combobox-item-group/combobox-item-group.tsx
```diff
@@ -63,7 +63,10 @@ export class ComboboxItemGroup {
     return (
       <ul
         aria-labelledby={this.guid}
-        class={{ [CSS.list]: true, [`scale--${scale}`]: true }}
+        class={{
```
jcfranco, line 66: This file isn't related to the fix, so can you remove it from the PR?

jcfranco (approved):
Awesome!

👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾
👾👩‍🚀👩‍🚀👩‍🚀👩‍🚀👾👩‍🚀👾👾👾👾👾👩‍🚀👾👩‍🚀👩‍🚀👩‍🚀👩‍🚀👾👾👩‍🚀👩‍🚀👩‍🚀👾👾👩‍🚀👩‍🚀👾👾👩‍🚀👾👾👾👩‍🚀👾👩‍🚀👩‍🚀👩‍🚀👩‍🚀👾
👾👩‍🚀👾👾👩‍🚀👾👩‍🚀👾👾👾👾👾👩‍🚀👾👩‍🚀👾👾👾👾👩‍🚀👾👾👾👾👩‍🚀👾👾👩‍🚀👾👩‍🚀👩‍🚀👾👩‍🚀👩‍🚀👾👩‍🚀👾👾👾👾
👾👩‍🚀👩‍🚀👩‍🚀👩‍🚀👾👩‍🚀👾👾👩‍🚀👾👾👩‍🚀👾👩‍🚀👩‍🚀👩‍🚀👾👾👾👩‍🚀👩‍🚀👾👾👩‍🚀👾👾👩‍🚀👾👩‍🚀👾👩‍🚀👾👩‍🚀👾👩‍🚀👩‍🚀👩‍🚀👾👾
👾👩‍🚀👾👾👩‍🚀👾👩‍🚀👾👩‍🚀👾👩‍🚀👾👩‍🚀👾👩‍🚀👾👾👾👾👾👾👾👩‍🚀👾👩‍🚀👾👾👩‍🚀👾👩‍🚀👾👾👾👩‍🚀👾👩‍🚀👾👾👾👾
👾👩‍🚀👾👾👩‍🚀👾👾👩‍🚀👾👾👾👩‍🚀👾👾👩‍🚀👩‍🚀👩‍🚀👩‍🚀👾👩‍🚀👩‍🚀👩‍🚀👾👾👾👩‍🚀👩‍🚀👾👾👩‍🚀👾👾👾👩‍🚀👾👩‍🚀👩‍🚀👩‍🚀👩‍🚀👾
👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾

driskull (approved):
LGTM! 👏

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -967,6 +967,15 @@ export class Combobox
     if (!this.allowCustomValues) {
       this.setMaxScrollerHeight();
     }
+
+    this.groupItems.forEach((groupItem, index, array) => {
+      if (groupItem.children.length === 0) {
+        if (index + 1 < array.length) {
+          const nextGroupItem = array[index + 1];
+          nextGroupItem.classList.add(CSS.groupThatFollowsChildlessGroup);
```
jcfranco, line 975: Can you add a reflected, internal property (`afterEmptyGroup`?) to `combobox-group`s that `combobox` sets here instead of applying the class here?

jcfranco (approved):
Awesome! 🎉

packages/calcite-components/src/components/combobox/combobox.stories.ts
```diff
@@ -449,17 +449,31 @@ export const withSelectorIndicatorAndIcons_TestOnly = (): string => html`
...
 
-export const nestedGroups_TestOnly =
-  (): string => html`<calcite-combobox open selection-mode="single" style="width:400px" placeholder="Type to filter">
-<calcite-combobox-item-group label="Level 1">
-  <calcite-combobox-item-group label="Level 2">
-    <calcite-combobox-item-group label="Level 3">
-      <calcite-combobox-item value="Item 1" text-label="Item 1">
-      </calcite-combobox-item>
-      <calcite-combobox-item value="Item 2" text-label="Item 2">
+export const nestedGroups = (): string => html`
```
jcfranco, line 452: Should this still use the `_TestOnly` suffix? This is not critical for this PR and can be tackled separately.

Elijbet: yep, let me fix that before this merge

PR #7303, opened 2023-07-10
fix(block): loader now appears for all loading cases

Elijbet:
**Related Issue:** #6485

## Summary
Continues on #7239 to cover additional cases where loader needs to appear when there is no `status icon` or if there is a `slottedIcon` so that the loading icon shows both when a collapsible block is `open` and `closed`. Any slotted icons get overridden when loading.


packages/calcite-components/src/components/block/block.stories.ts
```diff
@@ -215,3 +215,31 @@ export const contentSpacing_TestOnly = (): string =>
       <div>Some text that has padding built in</div>
     </calcite-block>
   `;
+
+export const loadingCaseLoadingWithSlottedIcon_TestOnly = (): string =>
```
jcfranco, line 219: I think we can drop `loadingCase` since the name indicates it's loading-related and `_TestOnly` reflects a test case.

packages/calcite-components/src/components/block/block.tsx
```diff
@@ -237,17 +237,19 @@ export class Block
...
+      </div>
+    );
 
-    const iconEl = !statusIcon ? (
-      <slot key="icon-slot" name={SLOTS.icon} />
-    ) : loading ? (
-      <calcite-loader inline label={messages.loading} />
-    ) : (
+    return loading ? (
+      loaderIcon
```
jcfranco, line 251: Can you inline `loaderIcon`'s JSX here? Otherwise, under the hood, it's creating a `VNode` that may not get used in most cases.

packages/calcite-components/src/components/block/block.tsx
```diff
@@ -237,17 +237,19 @@ export class Block
...
+    ) : statusIcon && statusIcon !== "" ? (
```
jcfranco, line 252: This can be simplified to `!!statusIcon`.

jcfranco (approved):
Aside from a few minor suggestions, code changes LGTM!

packages/calcite-components/src/components/block/block.tsx
Elijbet: It's the linter's preference to have it wrapped.

packages/calcite-components/src/components/block/block.tsx
```diff
@@ -237,29 +237,31 @@ export class Block
...
-        scale="m"
-      />
-    );
+    const { loading, messages, status } = this;
+
+    const hasSlottedIcon = !!getSlotted(this.el, SLOTS.icon);
 
-    return hasIcon ? <div class={CSS.icon}>{iconEl}</div> : null;
+    return loading ? (
+      <div class={CSS.icon}>
```
jcfranco, line 245: Looks like the conditionally rendered container for the icon is the same and could be reused across renders. Could you set `key` on `<div class={CSS.icon}>` for this?

Issue #6462, opened 2023-02-13
[Action Bar] "Stray" clicks on Action Bar focus bottom action / toggle / first action


macandcheese:
### Actual Behavior

I noticed this in the documentation site app, but it seems to happen in our plain component setup as well.

If there are no bottom-actions, the toggle button gains focus.

If there are no bottom-actions and `expandDisabled` (so no expand toggle), the first item in the action bar is focused.

This video shows the issue both in the app itself and the action bar sample:

https://user-images.githubusercontent.com/4733155/218555803-0f26b253-b860-4b02-9d78-af41fca1bdb5.mov

### Expected Behavior

I would not expect a "stray" click on the Action Bar to focus the first bottom-actions action, the expand toggle, or the first action.

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/dyqbrOz?editors=100

### Reproduction Steps

1. Go to Codepen
2. Click the "empty space" between the top and bottom actions.

### Reproduction Version

1.0.5

### Esri team

Calcite (design)

geospatialem:
Research to be conducted to determine a time estimate for an upcoming milestone.

macandcheese:
Potential css-only fix: Adjust flex / margin on the bottom action group elements to not fill available space.

geospatialem:
Verified in `1.5.0-next.25`

PR #7310, opened 2023-07-11
fix(action-bar): no longer delegates focus when clicked on non-focusable region

anveshmekala:
**Related Issue:** #6462 

## Summary

This PR will fix focusing `calcite-action` in `calcite-action-bar` when user clicks on non-focusable area.  Neither the collapse button nor slotted `calcite-action` will be focused when clicked on non-focusable area.


packages/calcite-components/src/components/action-bar/action-bar.scss
```diff
@@ -53,5 +56,5 @@
 }
 
 .action-group--bottom {
-  @apply flex-grow justify-end;
```
driskull, line 56: Why was this removed?

anveshmekala: With flex-grow, the `calcite-action-group` is filling up the space in full width mode. Any click on the space filled is `delegatingFocu`s in `calcite-action-group`. To avoid this, the empty space will be filled by `calcite-action-bar`  and `delegateFocus` will be removed from `calcite-action-bar`.

driskull (approved):
Looks great 👍

PR #7313, opened 2023-07-12
ci: create more specific codeowners to limit unnecessary notifications

benelan:
## Summary
Adds more specific codeowners to stuff that most devs don't need email notifications about. If a PR contains anything not specified in the list of globs it will still go to the calcite-components-reviewers team.

ref: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners


jcfranco (approved):
💥

Issue #6921, opened 2023-05-04
Calcite-drop-down-item label property not working as expected


Vani8021:
### Summary

The `label` property as per the documentation should be used as an accessible name. But looks like it is not working as expected. The `aria-label` property seems to be set only when we use the `href` attribute too. 

### Actual Behavior

`label` property - Not working as an accessible name.

### Expected Behavior

`label` property - should be used to set the `aria-label` so that the screen readers can read them.

### Reproduction Sample

https://codepen.io/geospatialem/pen/poxpyEb

### Reproduction Steps

1. Copy paste the code pen sample on Chrome or any other browser
2. Right click to inspect the drop-down-item and look at the `Accessibility` Tab on the right side to look at the aria-attributes
3. Notice the aria-label is not being set.

### Reproduction Version

1.3.1

### Priority impact

p1 - need for current milestone

### Esri team

ArcGIS Dashboards

geospatialem:
There is a workaround where instead of using the expected `label` attribute, the `aria-label` can be substituted in its place.

Workaround is depicted in the [Codepen](https://codepen.io/geospatialem/pen/poxpyEb) where the first element contains the non-working `label`, and the second element contains an `aria-label`.

geospatialem:
Verified with JAWS with `1.5.0-next.25` with both the `dropdown-item`'s `label` and `aria-label`.

PR #7316, opened 2023-07-13
fix(dropdown-item): provides accessible label when href is not parsed

anveshmekala:
**Related Issue:** #6921 

## Summary

This PR provides accessible label for `calcite-dropdown-item` irrespective of `href` property value.


driskull (approved):
👍

PR #3771, opened 2021-12-22
ci(action): cleanup teams release notification

benelan:
**Related Issue:** NA

## Summary
- Removed unnecessary 30 min delay
- Changed the name of the GH secret. Since creating this, I added another teams notification for builds with the secret `TEAMS_WEBHOOK_URI_BUILD`. So I want to specify that this secret is for release for consistency
- Added `#changelog` to the link so you don't have to scroll to find the changes
- Hardcoded the action version just incase the owner pushes breaking changes to master
    - https://github.com/marketplace/actions/ms-teams-action?version=v1.0.1

PR #7317, opened 2023-07-13
ci: notify teams about calcite-components releases only

benelan:
**Original PRs:** #3771 #5512

## Summary

Adding back the Teams release notification action. I had to refactor it after the monorepo transition so it wouldn't spam the channel with release notifications for all of the packages. 

We can make additional steps for different packages that go to different channels at some point. For example,  a `design-tokens` release notification to "Get Help - Design and FIGMA"


jcfranco (approved):
✨🗞️✨

PR #7319, opened 2023-07-14
style: lint and format code

benelan:
**Related Issue:** #7318

## Summary

Lint and format the code. Prettier will now add trailing commas when possible, which is the default in 3x. 

I know already brought up disabling some jsdoc rules in the past, but I think we can at least disable the ` jsdoc/require-param-type` one. We are using TypeScript so defining the types twice seems redundant and likely to get out of sync when things change.


packages/calcite-components/src/components/date-picker-day/date-picker-day.scss
```diff
@@ -31,7 +31,6 @@
 
 .text {
   @apply mt-px mb-0;
-  margin-inline-start: theme("spacing.px");
```
benelan, line 34: @eriklharper can you confirm that eslint or prettier removed the correct one of these duplicate css properties?

driskull (approved):
👍

packages/calcite-components/src/components/date-picker-day/date-picker-day.scss
benelan: I blamed it and it's from this commit: https://github.com/Esri/calcite-design-system/commit/4169f30984f92f5146d1b3709a99fbe9062d6b68 

The one not being used was removed so I assume that's correct, but just wanted to double check before its gone forever.

PR #7326, opened 2023-07-17
chore(eslint): disable jsdoc/require-param-type rule in favor of typescript

benelan:
**Related Issue:** #7319

## Summary

Disables the [`jsdoc/require-param-type`](https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/require-param-type.md) eslint rule since we use TypeScript for that.


jcfranco:
Sharing this from our convo for additional context: this will reduce linting noise we would get by not specifying the type in JSDoc, especially since the JSDoc type info isn't available in the build output.

jcfranco (approved):

Issue #5978, opened 2022-12-10
Color Picker: Number greater than 255 changes back to latn from other numbering systems


annierm18:
### Actual Behavior

When a number greater than 255 is entered in the RGB fields, the numbering systems changes from that specified in the numbering system attribute to latn.

![Screenshot (107)](https://user-images.githubusercontent.com/25759835/206819798-597d25d0-b96c-4112-a519-d24a3b25ef7e.png)

### Expected Behavior

I expect the number to remain in the numbering system that I specified using the numbering-system attribute and not change to latn.

### Reproduction Sample

https://codepen.io/ainermfc/pen/JjZVJdN?editors=100

### Reproduction Steps

1. Open a codepen for color picker
2. Set numbering-system to a numbering system other than latn, like "arab"
3. Click in either the R, G, or B input
4. Type in two numbers

### Reproduction Version

beta.97

### Esri team

N/A

jcfranco:
The culprit lies in the channel input's clamping logic: https://github.com/Esri/calcite-components/blob/master/src/components/color-picker/color-picker.tsx#L366-L391. Need to take a closer look.

geospatialem:
Re-allocating to the upcoming March milestone.

geospatialem:
Verified in `1.5.0-next.25`.

![image](https://github.com/Esri/calcite-design-system/assets/5023024/7250212f-4300-43fe-a601-f10f1c658fbc)

PR #7327, opened 2023-07-18
fix(color-picker): maintains correct numbering system when entering invalid RGB value

benelan:
**Related Issue:** #5978

## Summary

Entering an RGB channel value greater than 255 or less than 0 clamps the value so it doesn't become invalid. This caused the numbering system to revert to the default `latn`. 

It is a timing issue with `calcite-input`'s lifecycle methods and it reproduces outside of `calcite-color-picker`. Here is a demo showing that using a timeout or requesting an animation frame (kind of) fixes the issue:

https://codepen.io/benesri/pen/vYQjYJK

The issue isn't reproducible in `calcite-input-number`.


driskull (approved):
🎉 🌮

PR #7333, opened 2023-07-18
test(block): disable loader animation in chromatic snapshots

benelan:
**Related Issue:** noticed the failing snapshots in #7327

For example: https://www.chromatic.com/test?appId=6266d45509d7eb004aa254fb&id=64b6d984dc6b56c885b8872f

## Summary

Sets the duration factor to 0 in chromatic snapshots to prevent failures due to loader animation

### DIFF
--- a/packages/calcite-components/src/components/block/block.stories.ts
+++ b/packages/calcite-components/src/components/block/block.stories.ts
@@ -218,6 +218,11 @@ export const contentSpacing_TestOnly = (): string =>
 
 export const loadingWithSlottedIcon_TestOnly = (): string =>
   html`
+    <style>
+      :root {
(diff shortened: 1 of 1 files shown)


driskull (approved):
nice!

Maybe we should just set this var for every story?

benelan:
> nice!
> 
> Maybe we should just set this var for every story?

Yeah good call, I'll refactor this PR to do that

benelan:
After looking around for a bit, I haven't found a good way to globally apply this to only the TestOnly stories. I assume we still want animations in the hosted storybook. I'll document this in our conventions for now and I added it to my list to keep digging if/when have some extra time.

Issue #6642, opened 2023-03-23
[Rating] Focus outline not present with mouse selection


geospatialem:
### Actual Behavior

When selecting the `rating` (in focus), shows the focus outline. However, the focus outline does not show when interacting and selecting the `rating` with a mouse. 

Discovered while conducting an audit on https://github.com/Esri/calcite-components/issues/4633 cc @ashetland 

https://user-images.githubusercontent.com/5023024/227353740-908572cc-08fd-4bad-acf8-527fd405199a.mp4

### Expected Behavior

Should perform similar to the `tabs` or `pagination` components when selection occurs with a mouse.

![expected-result](https://user-images.githubusercontent.com/5023024/227366823-517af13d-aa29-4faa-9af1-a8177a53e23b.gif)

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/rating/

### Reproduction Steps

1. Open the [sample](https://developers.arcgis.com/calcite-design-system/components/rating/)
2. Navigate through the component using the keyboard and observe the proper focus
3. Navigate the component with a mouse and observe the focus is not present

### Reproduction Version

1.1.0

### Priority impact

p3 - want for upcoming milestone

### Impact

Moderate - a nice to have feature, but not a critical a11y prioritization.

### Esri team

Calcite (design)

geospatialem:
Verified in `1.7.0-next.9`.

Issue #4633, opened 2022-05-27
Hitbox focus on components


geospatialem:
### Summary

Complete an audit of all the components to reduce block elements in focus. For instance, the color picker selector.

This came up as an action as part of https://github.com/Esri/calcite-components/issues/1359.

### Actual Behavior

Some components elements are depicted as block elements when focused.

### Expected Behavior

Focus highlights the extent of the element.

#### Components
- [x] `color-picker`: #7378 (verified in `1.5.0-next.30`)
- [x] `link`: #7390 (verified in `1.5.0-next.30`)
- [x] `pagination`: no change
- [x]  `rating`: Focus outline should appear when clicked #7341 
- [x]  `stepper`: no change
- [x] `tabs`: no change
- [x] `switch`: #7390 (verified in `1.5.0-next.30`)

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/color-picker/#sample

### Reproduction Steps

1. Open the [sample page](https://developers.arcgis.com/calcite-design-system/components/color-picker/#sample).
2. Navigate into the color picker using keyboard keys (`tab`) to see the focus highlighting a box for the color picker selector.

### Reproduction Version

beta.82

### Working W3C Example/Tutorial

N/A

### Relevant Info

N/A

geospatialem:
## Audit results:

### 1. `color-picker`
Includes boxy focus on the following elements of the component (displayed in one image below for clarity):
- color handle
- swatch handle and
- RGB and HSV groupings (depicted in a 1px black rectangle)
![color picker screenshot with multiple focus combinations](https://user-images.githubusercontent.com/5023024/209837622-2d635769-b1db-4b63-b209-3ec81ca0236b.png)

### 2. `link`
The hitbox needs a bit more padding, it's a bit hard to discern on the left/right, includes spacing where links could extend multiple lines.
![link focus screenshot simple](https://user-images.githubusercontent.com/5023024/209840220-aa1a90f6-01d9-4471-ac7a-5e353ae56378.png)
![link focus screenshot multiple lines](https://user-images.githubusercontent.com/5023024/209839813-9183da78-2efe-44cd-b458-2e6b072cf2fc.png)

### 3. `pagination` cc https://github.com/Esri/calcite-components/issues/5113
The focus and selection conflict, maybe a differing focus, such as a circle would help differentiate?
[...]

PR #7341, opened 2023-07-19
fix(rating): adds focus outline on click

anveshmekala:
**Related Issue:** #4633 , https://github.com/Esri/calcite-design-system/issues/6642

## Summary

This PR will add focus outline for `calcite-rating` component on click.


packages/calcite-components/src/components/rating/rating.tsx
```diff
@@ -407,35 +353,77 @@ export class Rating
...
-    const newPointerValue = Number(target.firstChild["value"] || 0);
-    this.hoverValue = newPointerValue;
-    this.focusValue = null;
+    this.hoverValue = this.getValueFromLabelEvent(event);
   };
 
   private handleLabelPointerDown = (event: PointerEvent) => {
     const target = event.currentTarget as HTMLLabelElement;
-    const inputVal = Number(target.firstChild["value"] || 0);
+    const inputValue = Number(target.firstChild["value"]);
```
driskull, line 374: maybe instead of using `target.firstChild` you can use the `data-value` attribute?

packages/calcite-components/src/components/rating/rating.tsx
```diff
@@ -407,35 +353,77 @@ export class Rating
...
+    }
+  }
+
+  private setLabelEl = (el: HTMLLabelElement): void => {
+    this.labelElements.push(el);
+  };
+
+  private getValueFromLabelEvent(event: FocusEvent | PointerEvent | KeyboardEvent): number {
+    const target = event.currentTarget as HTMLLabelElement;
+    return Number(target.firstChild["value"]) || 0;
```
driskull, line 416: maybe instead of using `target.firstChild` you can use the `data-value` attribute?

packages/calcite-components/src/components/rating/rating.tsx
```diff
@@ -259,31 +239,26 @@ export class Rating
...
                       star: true,
-                      focused,
                       selected,
                       hovered,
                       average,
                       partial,
                     }}
+                    data-value={value}
                     htmlFor={id}
+                    onClick={this.handleLabelClick}
```
driskull, line 254: could this be done in `pointerDown`?

packages/calcite-components/src/components/rating/rating.tsx
```diff
@@ -259,31 +239,26 @@ export class Rating
...
+                    onFocus={this.handleLabelFocus}
```
driskull, line 255: just to note:
We could use `onfocusin` and `onfocusout` on the host instead of `onBlur` and `onFocus` on different elements since `onfocusin` and `onfocusout` bubble whereas blur/focus do not bubble.

driskull (changes requested):
Looking good! 🌮

packages/calcite-components/src/components/rating/rating.tsx
```diff
@@ -446,8 +434,8 @@ export class Rating
   @Method()
   async setFocus(): Promise<void> {
     await componentFocusable(this);
-
-    this.inputFocusRef?.focus();
+    const focusableEl = this.labelElements.find((rating) => rating.tabIndex === 0);
```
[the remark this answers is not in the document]
anveshmekala: dang, i forgot this name and searched for `firstFocusable`

packages/calcite-components/src/components/rating/rating.tsx
anveshmekala: preventing default in pointerDown handler wouldn't stop `click` event from being invoked.

packages/calcite-components/src/components/rating/rating.tsx
anveshmekala: This made me think and we don't need a `blur` event handler here. And tried using `focusin` at `Host` level which showed weird behavior when user is tabbing out of the rating and then tabbing back in using `Shift+Tab` which is not recognized as focus by `focusin` event.

driskull (approved):
👍 LGTM. Lets get one other review before merging.

packages/calcite-components/src/components/rating/rating.tsx
```diff
@@ -407,35 +353,73 @@ export class Rating
...
+  private handleLabelClick = (event: Event) => {
+    event.preventDefault();
+  };
+
+  private handleLabelFocus = (event: FocusEvent) => {
+    const inputValue = this.getValueFromLabelEvent(event);
+    this.hoverValue = inputValue;
+  };
+
+  private updatefocus(): void {
```
jcfranco, line 390: `updatefocus` ➡️ `updateFocus`

packages/calcite-components/src/components/rating/rating.tsx
jcfranco: Interesting. I thought this was a browser bug, but it seems to be working as intended (see _Pointer Event: Suppress compatibility mouse events on click_ [test](https://github.com/web-platform-tests/wpt/blob/master/pointerevents/pointerevent_suppress_compat_events_on_click.html) and [results](https://w3c.github.io/test-results/pointerevents/all.html#test-file-48). Can you add a comment about this for posterity?

packages/calcite-components/src/components/rating/rating.e2e.ts
```diff
@@ -519,8 +521,15 @@ describe("calcite-rating", () => {
...
       await page.waitForChanges();
 
+      const hoveredEl = await page.findAll("calcite-rating >>> .star.hovered");
+      const selectedEl = await page.findAll("calcite-rating >>> .star.selected");
+      const focusedElId = labels[2].getAttribute("for");
+
       expect(await element.getProperty("value")).toBe(0);
       expect(changeEvent).toHaveReceivedEventTimes(1);
+      expect(hoveredEl.length).toBe(3);
+      expect(await isElementFocused(page, `[for=${focusedElId}]`, { shadowed: true })).toBe(true);
```
jcfranco, line 531: Not critical, but is it possible to query using the `star` class via `label` or `label` attributes? This would make the test less brittle in case the implementation changes.

packages/calcite-components/src/components/rating/rating.tsx
```diff
@@ -259,31 +239,26 @@ export class Rating
...
+                    data-value={value}
                     htmlFor={id}
+                    onClick={this.handleLabelClick}
+                    onFocus={this.handleLabelFocus}
+                    onKeyDown={this.handleLabelKeyDown}
                     onPointerDown={this.handleLabelPointerDown}
                     onPointerOver={this.handleLabelPointerOver}
+                    tabIndex={tabIndex}
+                    // eslint-disable-next-line react/jsx-sort-props
+                    ref={this.setLabelEl}
```
jcfranco, line 261: ✨🧹✨

packages/calcite-components/src/components/rating/rating.tsx
```diff
@@ -259,31 +239,26 @@ export class Rating
...
                       partial,
                     }}
+                    data-value={value}
                     htmlFor={id}
+                    onClick={this.handleLabelClick}
+                    onFocus={this.handleLabelFocus}
+                    onKeyDown={this.handleLabelKeyDown}
                     onPointerDown={this.handleLabelPointerDown}
                     onPointerOver={this.handleLabelPointerOver}
+                    tabIndex={tabIndex}
```
jcfranco, line 259: Does making `label` have any impact on how screen readers interpret the value?

jcfranco (approved):
Awesome! 

Had a few comments, but this LGTM! 🚀

Risk-wise, I think we should wait after the upcoming maintenance release is done before landing this one.

jcfranco:
> fix(rating): adds focus outline on click

One more thing, should the title reflect that focus is added for both pointer and keyboard usage? It reads as only covering the pointer case.

anveshmekala:
> > fix(rating): adds focus outline on click
> 
> One more thing, should the title reflect that focus is added for both pointer and keyboard usage? It reads as only covering the pointer case.

Since we marked this as `fix` i think its safe to mention click only and we attached the issue too.

packages/calcite-components/src/components/rating/rating.tsx
anveshmekala: the screen reader experience has no impact in VoiceOver.  cc @geospatialem 

I do see a lot of intermediate states with VoiceOver where visually hidden area are being focused when user is navigating using voiceover commands + rightArrow.  we need to add `aria-hidden` attribute to remove the visually hidden elements from accessibility tree. This can be done in a different PR though, wanted to confirm similar behavior with NVDA/JAWS before diving in.

anveshmekala: One addition could be announcing to AT user when user reset the selection ( not  available in native behavior).

geospatialem: Not following the logic above, is this to remove the `label`?

geospatialem (commented):
Changes look good for mouse users. Also verified with JAWS, just for testing sake, and we're good to go. 👍🏻

Issue #6588, opened 2023-03-13
Link: add ability to use outset focus outline positioning 


paulcpederson:
### Description

We are receiving issues where users think the focus style of the link element is a bug. The default position of the outline ring on a link is -2px inset. This means the outline overlaps the text. There is no way to customize this. 

We are asking for either:

1. The focus outline to be outset by default _or_
2. Some customization mechanism like a CSS var or a prop to allow us to give the outline space around the content.

### Acceptance Criteria

We should be able to have the outline at 2px offset

### Relevant Info

Link actually used to use the outset focus style.  https://github.com/Esri/calcite-components/commit/d2f95a67ae1f5d8c3935e12930542bebbdce6e9e#diff-9f60eb4702a4bb9e8e5a80c284b3695e8dba637ae48738791a6961b8de9c12c1R46

I'm not sure why it was changed, presumably some need from using the link component internally? Outset seems like a more usable default for links in the run of text. 

### Which Component

calcite-link

### Example Use Case

Really any link use case. 

### Esri team

ArcGIS Online

macandcheese:
Agree - @ashetland / @SkyeSeitz are there any places reverting to outset as the default would cause an issue? Maybe somewhere link is expected to be slotted?

ashetland:
Related issue https://github.com/Esri/calcite-components/issues/4633

macandcheese:
For this one in particular, we should just revert link to use outset.

PR #7342, opened 2023-07-19
fix(link): adds outline-offset to avoid overlapping with text.

anveshmekala:
**Related Issue:** #6588 

## Summary
This PR will add back the `2px` outline-offset for `calcite-link`.


jcfranco (approved):
🔗🔍👁️

Issue #7344, opened 2023-07-19
Tooltip not honoring --calcite-app-z-index-tooltip token


jcfranco:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

All components should be using tokens from `calcite-design-tokens`, but tooltip still has a hard-coded value.

As a workaround, developers can use the existing `--calcite-tooltip-z-index` on the tooltip.

### Expected Behavior

To be able to set `--calcite-app-z-index-tooltip` and have tooltip honor it.

### Reproduction Sample

https://codepen.io/jcfranco/pen/wvQXKvG

### Reproduction Steps

After opening the codepen, notice how:

1. The first tooltip is not hidden by modal's overlay (using `--calcite-app-z-index-tooltip`)
2. The second tooltip is hidden by modal's overlay (using `--calcite-tooltip-z-index`)

### Reproduction Version

1.4.3

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

Calcite (dev)

geospatialem:
Verified in `1.5.0-next.24` with `--calcite-app-z-index-tooltip`.

PR #7345, opened 2023-07-19
fix(tooltip): ensure --calcite-app-z-index-tooltip is applied

jcfranco:
**Related Issue:** #7344 

## Summary

This updates the tooltip z-index layer to use the corresponding design token CSS var.


driskull (approved):
👍

Issue #7325, opened 2023-07-17
Global CSS file has grown 5000 lines.


driskull:
### Description

1. Can we reduce the file size of the calcite.css style sheet? The CSS file grew from ~300 lines to ~5300 lines. So basically 5000 new lines. 
    
    1.3.0: https://cdn.jsdelivr.net/npm/@esri/calcite-components@1.3.0/dist/calcite/calcite.css
    1.4.0: https://cdn.jsdelivr.net/npm/@esri/calcite-components@1.4.0/dist/calcite/calcite.css

2. Can we move the animation classes into a new file for utility classes that can be used by teams if necessary. We can document this new file and its utility classes. The animation keyframes should also be renamed to be BEM so that they do not have any conflicts.

```css
@keyframes in {
  0% {
    opacity: 0;
  }
  100% {
    opacity: 1;
  }
}
```

3. Remove the `overflow-hidden` class only being used by modal. This could change to an inline style where Modal can just set the overflow to hidden on the document instead of using a class.

```css
.overflow-hidden {
  overflow: hidden;
}
```

4. Remove the following, it should be done in a components CSS not in the global CSS. 
```css
host([hidden]) {
  display: none;
}
```

### Proposed Advantages

Reduced file size, no conflicting code, utility classes are separated and documented properly.

### Which Component

All

- [x] Refactor utility classes (Matt, Adam)
- [ ] Refactor tokens (Ali, Franco)

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

driskull:
cc @jcfranco @alisonailea

jcfranco:
We could treat the order to also reflect priority/impact too. 

1. It looks like the [wiring up of tokens](https://github.com/Esri/calcite-design-system/pull/6884) caused the increase. @alisonailea Are there any strategies we could apply to strike a balance between token availability/flexibility and global CSS bloat (possibly moving component-specific tokens to the component's styles)? This will increase even more as we add more tokens and components.
2. For the time being, I'd like to focus on generalizing as much animations as possible and moving them to `_animations.scss`. 
3. Your suggestion sounds perfect.
4. Is your concern here the style impacting non-Calcite components? I think this one might still be good to keep in the global styles.

driskull:
> For the time being, I'd like to focus on generalizing as much animations as possible and moving them to _animations.scss.

[...]

PR #7346, opened 2023-07-19
fix: remove style modifying all host components with hidden attribute

driskull:
**Related Issue:** #7325

## Summary

- Removes style modifying all host components with hidden attribute. This is modifying components outside of calcite and applying host styles outside of the host. Host styles should have no effect outside of a shadow DOM. https://developer.mozilla.org/en-US/docs/Web/CSS/:host
- Adds includes mixin for common component internal styles such as hidden attribute styling
- Includes mixin in every component file.

### DIFF
--- a/packages/calcite-components/src/assets/styles/includes.scss
+++ b/packages/calcite-components/src/assets/styles/includes.scss
@@ -15,10 +15,6 @@
   }
 }
 
-:host([hidden]) {
-  display: none;
-}
-
 @mixin word-break() {
   word-wrap: break-word;
   word-break: break-word;
--- a/packages/calcite-components/src/assets/styles/includes.scss
+++ b/packages/calcite-components/src/assets/styles/includes.scss
@@ -71,6 +71,16 @@
   }
 }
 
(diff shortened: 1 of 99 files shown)


packages/calcite-components/src/assets/styles/includes.scss
```diff
@@ -75,6 +71,16 @@
   }
 }
 
+@mixin calciteComponentCommon() {
```
jcfranco, line 74: We don't prefix other mixins w/ `calciteComponent`. Maybe `base` or `common`?

jcfranco (approved):
👁️🚫👻

packages/calcite-components/src/assets/styles/includes.scss
driskull: I also noticed some mixins use camel case vs dash separated? do we want to pick one?

- calciteHydratedHidden
- xButton
- hidden-form-input
- word-break

driskull: Renamed to `baseComponent`

PR #7347, opened 2023-07-19
refactor(modal): use inline style to set overflow instead of a class

driskull:
**Related Issue:** #7325

## Summary

- refactor(modal): use inline style to set overflow instead of a class
- We don't want to expose css utility classes this way.

### DIFF
--- a/packages/calcite-components/src/components/modal/modal.e2e.ts
+++ b/packages/calcite-components/src/components/modal/modal.e2e.ts
@@ -520,10 +520,10 @@ describe("calcite-modal accessibility checks", () => {
     const modal = await page.find("calcite-modal");
     await modal.setProperty("open", true);
     await page.waitForChanges();
-    const documentClass = await page.evaluate(() => {
-      return document.documentElement.classList.contains("overflow-hidden");
+    const isOverflowHIdden = await page.evaluate(() => {
+      return document.documentElement.style.overflow === "hidden";
     });
-    expect(documentClass).toEqual(true);
+    expect(isOverflowHIdden).toEqual(true);
   });
 
   it("correctly does not add overflow class on document when open and slotted in shell modals slot", async () => {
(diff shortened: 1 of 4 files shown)


driskull:
@jcfranco can I get a review?

packages/calcite-components/src/components/modal/modal.e2e.ts
```diff
@@ -520,10 +520,10 @@ describe("calcite-modal accessibility checks", () => {
     const modal = await page.find("calcite-modal");
     await modal.setProperty("open", true);
     await page.waitForChanges();
-    const documentClass = await page.evaluate(() => {
-      return document.documentElement.classList.contains("overflow-hidden");
+    const isOverflowHIdden = await page.evaluate(() => {
```
jcfranco, line 523: Typo: `isOverflowHIdden` (uppercase `I` in `Hidden`)

Applies to similar util used below.

packages/calcite-components/src/components/modal/modal.tsx
```diff
@@ -532,7 +532,7 @@ export class Modal
     this.contentId = ensureId(contentEl);
 
     if (!this.slottedInShell) {
```
jcfranco, line 534: Could you add some explainers to the inline style logic?

packages/calcite-components/src/components/modal/modal.tsx
```diff
@@ -532,7 +532,7 @@ export class Modal
...
-      document.documentElement.classList.add(CSS.overflowHidden);
+      document.documentElement.style.setProperty("overflow", "hidden");
```
jcfranco, line 535: In order to play nice with existing inline styles, could you store the value of `overflow` (if set) and restore when our override is removed?

jcfranco (approved):
Awesome!

PR #7337, opened 2023-07-18
docs: add more chromatic/storybook info

benelan:
**Related Issue:** #7333

## Summary

- Add contributing information about running chromatic

PR #7348, opened 2023-07-20
ci(chromatic): disable animations globally to prevent false positive visual diffs

benelan:
**Related Issue:** https://github.com/Esri/calcite-design-system/pull/7337#discussion_r1267378922

## Summary

Uses `--calcite-duration-factor` to disable animations for all snapshots when running in Chromatic.


jcfranco (approved):
✨🎥✨

Issue #6018, opened 2022-12-13
Refactor all `OpenCloseComponent`s with the new pattern where `onToggleOpenCloseComponent` replaces `connect` logic


Elijbet:
### Description

- Implement `OpenCloseComponent` interface on all relevant components.
- Use the new pattern where `onToggleOpenCloseComponent` replaces both `connectOpenCloseComponent` and `disconnectOpenCloseComponent`.
- ~Once implemented on all components, clean up the util to remove the connect pattern helpers in favor of `onToggleOpenCloseComponent` helper.~

### Proposed Advantages

Consistency across relevant components and less functions to import into each component from the util. This new pattern simplifies the overall logic. 

### Which Component

All components and the util itself.

- [x] alert
- [x] dropdown
- [x] tooltip
- [x] popover
- [x] input-date-picker
- [x] combobox

geospatialem:
The top two bullet points have been verified on the `main` branch. ✨ 

The last bulletpoint was edited as it is not part of the effort. Instead future refactoring will be taking place via https://github.com/Esri/calcite-design-system/issues/6017, https://github.com/Esri/calcite-design-system/issues/4398, https://github.com/Esri/calcite-design-system/issues/5056, and https://github.com/Esri/calcite-design-system/issues/5074, with the following priorities:
1. Apply utility across all components
2. Refactor the utility
3. Incorporate a test utility

PR #7356, opened 2023-07-21
refactor(alert, dropdown): `onToggleOpenCloseComponent` replaces connect logic

Elijbet:
**Related Issue:** #6018

## Summary
Use the new `onToggleOpenCloseComponent` pattern to replace both `connectOpenCloseComponent` and `disconnectOpenCloseComponent` on `alert` and `dropdown`. This new pattern is intended to simplify the overall eventing logic.

### DIFF
--- a/packages/calcite-components/src/components/alert/alert.tsx
+++ b/packages/calcite-components/src/components/alert/alert.tsx
@@ -31,11 +31,7 @@ import {
   NumberingSystem,
   numberStringFormatter,
 } from "../../utils/locale";
-import {
-  connectOpenCloseComponent,
-  disconnectOpenCloseComponent,
-  OpenCloseComponent,
-} from "../../utils/openCloseComponent";
+import { onToggleOpenCloseComponent, OpenCloseComponent } from "../../utils/openCloseComponent";
 import {
   connectMessages,
   disconnectMessages,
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/alert/alert.tsx
```diff
@@ -89,6 +85,7 @@ export class Alert implements OpenCloseComponent, LoadableComponent, T9nComponen
   openHandler(): void {
     if (this.open && !this.queued) {
       this.calciteInternalAlertRegister.emit();
+      onToggleOpenCloseComponent(this);
```
jcfranco, line 88: This needs to be called when the component is both open and closed, right?

jcfranco (changes requested):
Can you look at adding tests for these changes? This behavior hasn't been tested so far and I'd like to avoid any possible regressions from the refactor.

We could use them as the base for the test helper. 🎉

Elijbet:
> Can you look at adding tests for these changes? This behavior hasn't been tested so far and I'd like to avoid any possible regressions from the refactor.
> 
> We could use them as the base for the test helper. 🎉

The behavior is whether the events are emitted right? 
eg.
  ```it("should emit component status for transition-chained events: 'calciteInputDatePickerBeforeOpen', 'calciteInputDatePickerOpen', 'calciteInputDatePickerBeforeClose', 'calciteInputDatePickerClose'", async () => {```

jcfranco:
> The behavior to test is whether the events are emitted right?

Correct.

jcfranco (approved):
Awesome! Thanks for pointing out that the existing tests already covered this changes. 

Per our conversation, let's extract those tests into a test helper and update it to cover disabled animations.

PR #7368, opened 2023-07-24
fix(panel): Remove double border styling when content isn't provided

driskull:
**Related Issue:** #7423

## Summary

- Fixes display issue when no content is slotted within a panel but a footer or action-bar is displayed then a double border occurs.
- Cleans up code to use `slotChangeHasAssignedElement` dom utility for slot changes
- Fixes conditional element surrounding default slot.
- Adds screenshot tests


jcfranco:
Can you double check the screenshot diffs? I think there's some spacing inconsistencies between `panel` and `flow-item`.

driskull:
> Can you double check the screenshot diffs? I think there's some spacing inconsistencies between panel and flow-item.

@jcfranco updated.

packages/calcite-components/src/components/panel/panel.tsx
```diff
@@ -525,27 +511,16 @@ export class Panel
   };
 
   renderContent(): VNode {
-    const { hasFab } = this;
-
-    const defaultSlotNode: VNode = <slot key="default-slot" />;
-    const containerNode = hasFab ? (
-      <section class={CSS.contentContainer}>{defaultSlotNode}</section>
```
driskull, line 532: @asangma do you think you could review these changes? I changed this because it was causing some problems with `onSlotChange` since this default slot was getting moved into a section element depending on if there was a FAB or not. I don't think its necessary if we always have the section element present and the FAB content classes are no longer conditional.

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -29,9 +29,8 @@
 }
 
 .header {
-  border-block-end: 1px solid;
+  border-block-end: 1px solid var(--calcite-ui-border-3);
```
driskull, line 32: moved these together since `border-block-end` isnt exactly the same as `border-bottom`. Applies to other CSS

packages/calcite-components/src/components/panel/panel.tsx
```diff
@@ -525,27 +511,16 @@ export class Panel
...
-          [CSS.contentContainer]: !hasFab,
-          [CSS.contentHeight]: hasFab,
-        }}
+        class={CSS.contentWrapper}
         onScroll={this.panelScrollHandler}
         // eslint-disable-next-line react/jsx-sort-props
         ref={this.setPanelScrollEl}
       >
-        {containerNode}
+        <section class={CSS.contentContainer}>
```
driskull, line 521: @asangma do you remember why this `section` was necessary for the FAB? Is there any harm in always having it there for consistency and so the default slot doesn't get rearranged?

driskull:
@jcfranco can you review? I'd like to get this one in if possible.

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -40,9 +38,12 @@
 }
 
 .action-bar-container {
-  border-block-end: 1px solid;
   @apply w-full
-  z-header border-b-color-3;
+  z-header;
+}
+
+.border-block-end {
```
jcfranco, line 45: Nitpick: is there a more expressive name for this class?

jcfranco (approved):
Nice! 🎉

Can you also submit an issue to link to this? Otherwise, we have no test case nor issue to verify. cc @geospatialem

jcfranco:
@asangma Friendly ping! 😄

packages/calcite-components/src/components/panel/panel.tsx
asangma: I think the section was there to keep the FAB sticky. But if it can be sticky without the section, that's cool too.

driskull: Ok. i have it so the section is always there regardless of the FAB or not. Ill move forward with that since its safer.

PR #7377, opened 2023-07-25
refactor: clean up global stylesheets

driskull:
**Related Issue:** #7325

## Summary

- Reduces root selector amount.
  - Only a single root selector now
  - animation and floating-ui CSS were spitting out their own root selectors just for CSS vars
- Moves animation helper classes into a mixin
- Moves reduced motion CSS into a mixin


jcfranco (approved):
Awesome! ✨🧹✨

PR #7378, opened 2023-07-25
feat(color-picker): replaces thumb focus outline to rounded

anveshmekala:
**Related Issue:** #4633 

## Summary
This PR will replace the existing focus outline for thumb in `calcite-color-picker` to rounded. 

After the change:
494C307B-315E-4842-93C5-911C6F3810A2
ED840D87-BFA5-4602-8D1C-51FF78CD04E2


packages/calcite-components/src/components/color-picker/color-picker.tsx
```diff
@@ -793,8 +807,12 @@ export class ColorPicker
                 aria-valuenow={color?.round().hue() || DEFAULT_COLOR.round().hue()}
                 class={{ [CSS.scope]: true, [CSS.hueScope]: true }}
                 onKeyDown={this.handleHueScopeKeyDown}
+                onPointerDown={this.handleHueSliderPointerDown}
                 role="slider"
-                style={{ top: `${hueTop}px`, left: `${hueLeft}px` }}
+                style={{
+                  top: `${hueTop - DIMENSIONS.scopeNode / 2}px`,
```
driskull, line 813: since a few places are dividing a value by the scopeNode and converting to px maybe convert that into a function for reuse.

packages/calcite-components/src/components/color-picker/color-picker.tsx
```diff
@@ -597,6 +606,7 @@ export class ColorPicker
     const { activeCanvasInfo, el } = this;
 
     if (!el.isConnected || !activeCanvasInfo) {
+      console.log("move handler");
```
driskull, line 609: remove console logs.

driskull (approved):
👍 🎉 🌮

packages/calcite-components/src/components/color-picker/utils.ts
```diff
@@ -254,3 +255,26 @@ export function toNonAlphaMode(mode: SupportedMode): SupportedMode {
...
+
+/**
+ * Returns adjusted left and top values of thumb node by substracting half of width/height of the thumb from computed left and top values.
+ * The current thumb node has 1px in width and height.
+ *
+ * @param left
+ * @param top
+ * @returns {[number, number]}
+ */
+export function getAdjustedLeftAndTop(left: number, top: number): [number, number] {
```
jcfranco, line 267: Can you move these functions to the color component? The current utils in this module are color-related.

packages/calcite-components/src/components/color-picker/utils.ts
```diff
@@ -254,3 +255,26 @@ export function toNonAlphaMode(mode: SupportedMode): SupportedMode {
...
+}
+
+/**
+ * Returns center of the thumb node. The current thumb node has 1px in width and height.
+ *
+ * @param X
+ * @param Y
+ * @returns {[number, number]}
+ */
+export function getCenterXYOfThumbNode(X: number, Y: number): [number, number] {
```
jcfranco, line 278: Nitpick: doesn't center imply x,y?

packages/calcite-components/src/components/color-picker/utils.ts
jcfranco, line 267: Can you revisit the name to be more expressive? It's not clear to me by name alone what the context of left & top is.

packages/calcite-components/src/components/color-picker/color-picker.tsx
```diff
@@ -557,6 +563,9 @@ export class ColorPicker
       context: this.hueSliderRenderingContext,
       bounds: this.hueSliderRenderingContext.canvas.getBoundingClientRect(),
     };
+    if (event.target === this.hueScopeNode) {
+      return;
+    }
     this.captureHueSliderColor(offsetX);
```
jcfranco, line 569: Color should still be captured when the hue scope is the event target. Each pixel along the hue slider can update the underlying color.

This also applies to the color field scope.

On a related note, since the scopes are mostly for keyboard interaction, should we disable pointer events on them to simplify the code a bit? Pointer interactions are already handled and we didn't have to worry about pointer events on the scopes since they were 0x0.

packages/calcite-components/src/components/color-picker/color-picker.e2e.ts
```diff
@@ -2157,7 +2161,7 @@ describe("calcite-color-picker", () => {
       const scope = await page.find(`calcite-color-picker >>> .${CSS.colorFieldScope}`);
 
       const initialStyle = await scope.getComputedStyle();
-      expect(initialStyle.left).toBe("0px");
+      expect(initialStyle.left).toBe("-0.5px");
```
jcfranco, line 2164: Per our conversation, would it be possible to avoid fractional pixel values? Not a deal breaker if not possible.

packages/calcite-components/src/components/color-picker/resources.ts
```diff
@@ -93,4 +93,5 @@ export const DIMENSIONS = {
       radius: 10,
     },
   },
+  thumbNode: 1,
```
jcfranco, line 96: Can this be moved to the component itself or a separate constant? Currently, I find it a bit confusing:

1. the dimensions object stores the size of different parts across all scales that are drawn onto canvas elements
2. I know it's a bit verbose, but repeated values across scales have a corresponding prop
3. none of the existing part names use `node`, and there's already one for thumb

jcfranco (changes requested):
Looking good, @anveshmekala! ✨🖍️😎✨

packages/calcite-components/src/components/color-picker/color-picker.scss
```diff
@@ -78,11 +78,14 @@
...
-    z-default;
-
+    z-default
+    rounded-full
+    bg-transparent
+    w-px 
+    h-px;
   &:focus {
     @apply focus-outset;
-    outline-offset: 16px;
```
jcfranco, line 85: One more thing I noticed. The design spec seems to have a 2px space between the thumb and the focus outline. The spacing in the PR screenshot looks larger than that. Maybe adjusting this to 14px can help better match that.

anveshmekala: if we increase the spacing from 12px to 14px it will increase the gap between the thumb and outline. 

Thoughts @ashetland

anveshmekala: Adding @macandcheese.

macandcheese: If I’m following, I agree it should be a 2px gap to align with other “outset focus” styles.

packages/calcite-components/src/components/color-picker/color-picker.e2e.ts
```diff
@@ -2236,9 +2243,64 @@ describe("calcite-color-picker", () => {
...
       expect(await hueSliderScope.getComputedStyle()).toMatchObject({
-        top: "10px",
-        left: "0px",
+        top: "9.5px",
+        left: "-0.5px",
       });
     });
   });
+
+  describe("mouse interaction with scope", () => {
```
jcfranco, line 2252: Awesome coverage here. Can you move this closer to the keyboard scope interaction block? Bonus points if these are grouped within a `scope interaction` describe block.

```tsx
describe("scope interaction", () => {
  describe("mouse", async () => { /* ... */ });
  describe("keyboard", async () => { /* ... */ });
})
```

This is not critical and can be done in a follow-up PR.

packages/calcite-components/src/components/color-picker/color-picker.e2e.ts
```diff
@@ -2236,9 +2243,64 @@ describe("calcite-color-picker", () => {
...
+
+    it("should update value when hue scope is moved", async () => {
+      const page = await newE2EPage();
+      await page.setContent(`<calcite-color-picker ></calcite-color-picker>`);
+      const colorPicker = await page.find("calcite-color-picker");
+
+      const [hueScopeX, hueScopeY] = await getElementXY(page, "calcite-color-picker", `.${CSS.hueScope}`);
+      const value = await colorPicker.getProperty("value");
+
+      await page.mouse.move(hueScopeX + 0.8, hueScopeY);
```
jcfranco, line 2280: What does the 0.8 represent? If it's an arbitrary move in pixels, can you simplify this by using whole numbers? Also, can you assign it to a self-documenting const?

This applies to the 10 and 2 values below too.

packages/calcite-components/src/components/color-picker/color-picker.e2e.ts
```diff
@@ -2236,9 +2243,64 @@ describe("calcite-color-picker", () => {
...
+    it("should update value when opacity scope is moved", async () => {
+      const page = await newE2EPage();
+      await page.setContent(`<calcite-color-picker alpha-channel></calcite-color-picker> <p>BROKEN</p>`);
+      let [opacityScopeX, opacityScopeY] = await getElementXY(page, "calcite-color-picker", `.${CSS.opacityScope}`);
+      const opacityInputEl = await page.find(`calcite-color-picker >>> [data-channel-index='3']`);
+      await page.mouse.move(opacityScopeX - 10, opacityScopeY);
+      await page.mouse.down();
+      await page.mouse.up();
+      await page.waitForChanges();
+      const value = await opacityInputEl.getProperty("value");
```
jcfranco, line 2296: Can you avoid going through the internal input? The component's value (including alpha) should update when  the scopes are interacted with.

packages/calcite-components/src/components/color-picker/resources.ts
```diff
@@ -94,3 +94,5 @@ export const DIMENSIONS = {
     },
   },
 };
+
+export const SCOPE_DIMENSION = 1;
```
jcfranco, line 98: Suggestion: `SCOPE_SIZE`.

jcfranco (approved):
Aside from some minor comments, this LGTM! 🚀

packages/calcite-components/src/components/color-picker/color-picker.e2e.ts
anveshmekala: the `0.8` represent the pixel inside the scope element and the reason for using value above 0.5  is to avoid getting rounded off and assert a change in value.

packages/calcite-components/src/components/color-picker/color-picker.e2e.ts
anveshmekala: currently `mouse interaction with scope` is placed near to keyboard scope interaction block.

anveshmekala: will group them.

Issue #7273, opened 2023-07-03
Skip unstable modal test


driskull:
### Test type

e2e

### Which Component(s)

modal

### Unstable Tests

"emits when duration is set to 0"

### Test error, if applicable

```
@esri/calcite-components:test: Summary of all failing tests
@esri/calcite-components:test: FAIL src/components/modal/modal.e2e.ts (66.694 s)
@esri/calcite-components:test:   ● opening and closing behavior › emits when duration is set to 0
@esri/calcite-components:test: 
@esri/calcite-components:test:     expected event "calciteModalBeforeOpen" to have been called 1 times, but was called 0 time
@esri/calcite-components:test: 
@esri/calcite-components:test:       291 |     await openSpy;
@esri/calcite-components:test:       292 |
@esri/calcite-components:test:     > 293 |     expect(beforeOpenSpy).toHaveReceivedEventTimes(1);
@esri/calcite-components:test:           |                           ^
@esri/calcite-components:test:       294 |     expect(openSpy).toHaveReceivedEventTimes(1);
@esri/calcite-components:test:       295 |
@esri/calcite-components:test:       296 |     await page.evaluate(() => {
@esri/calcite-components:test: 
@esri/calcite-components:test:       at Object.<anonymous> (src/components/modal/modal.e2e.ts:293:27)
@esri/calcite-components:test: 
@esri/calcite-components:test: 
@esri/calcite-components:test: Test Suites: 1 failed, 1 skipped, 117 passed, 118 of 119 total
```

### PR skipped, if applicable

#7274

Elijbet:
This was refactored as part of #7258.

geospatialem:
Verified on the `main` branch with `modal`'s e2e tests.

PR #7379, opened 2023-07-25
test(commonTests, alert, combobox, dropdown, input-date-picker, modal, popover, tooltip): extracts a general test utility `OpenClose` for all components implementing `openCloseComponent` util

Elijbet:
**Related Issue:** #4968

## Summary
Extracts a common test utility `openClose` for all components implementing `openCloseComponent` util. 

Implemented on all relevant components. 

- [x] alert
- [x] combobox
- [x] dropdown
- [x] input-date-picker
- [x] modal
- [x] popover
- [x] tooltip


packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -1323,3 +1323,67 @@ export async function t9n(componentTestSetup: ComponentTestSetup): Promise<void>
...
+ *
+ * @example
+ * import { emitsOpenCloseTransitionChainedEvents } from "../../tests/commonTests";
+ *
+ * describe("emitsOpenCloseTransitionChainedEvents", () => {
+ *   emitsOpenCloseTransitionChainedEvents("calcite-dropdown", "open", ".container");
+ * });
+ *
+ */
+export function emitsOpenCloseTransitionChainedEvents(
```
jcfranco, line 1340: Can you simplify the name so it matches the module it supports (similar to other common test helpers)?

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -1323,3 +1323,67 @@ export async function t9n(componentTestSetup: ComponentTestSetup): Promise<void>
...
+  componentTestSetup: ComponentTestSetup,
+  toggleProp: string,
+  containerSelector: string
+): void {
+  it(`should emit component status for transition-chained events: 'calciteComponentBeforeOpen', 'calciteComponentOpen', 'calciteComponentBeforeClose', 'calciteComponentClose'`, async () => {
```
jcfranco, line 1345: A few questions WRT the test name:

* what does `component status` mean?
* what are `transition-chained events`? These aren't mentioned here nor in the module, I believe.
* for readability, can we generalize the event name part to `before and when the component is toggled` (or something along these lines)?

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -1323,3 +1323,67 @@ export async function t9n(componentTestSetup: ComponentTestSetup): Promise<void>
...
+    const { page, tag } = await getTagAndPage(componentTestSetup);
+    await page.setContent(`${componentTestSetup}`);
+    const element = await page.find(tag);
+
+    const camelCaseTag = tag.replace(/-([a-z])/g, (g) => g[1].toUpperCase());
```
jcfranco, line 1350: I'm not sure what `g` represents, so could you rename `g` to something more expressive? It's the RegExp's match, right?

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -1323,3 +1323,67 @@ export async function t9n(componentTestSetup: ComponentTestSetup): Promise<void>
...
+
+    const beforeOpenEvent = page.waitForEvent(`${camelCaseTag}BeforeOpen`);
+    const openEvent = page.waitForEvent(`${camelCaseTag}Open`);
+
+    const beforeOpenSpy = await element.spyOnEvent(`${camelCaseTag}BeforeOpen`);
```
jcfranco, line 1355: Nitpick: can you extract the constructed event names before their use? They're referenced only twice, so I'll leave this up to you.

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -1323,3 +1323,67 @@ export async function t9n(componentTestSetup: ComponentTestSetup): Promise<void>
...
+ */
+export function emitsOpenCloseTransitionChainedEvents(
+  componentTestSetup: ComponentTestSetup,
+  toggleProp: string,
+  containerSelector: string
+): void {
+  it(`should emit component status for transition-chained events: 'calciteComponentBeforeOpen', 'calciteComponentOpen', 'calciteComponentBeforeClose', 'calciteComponentClose'`, async () => {
+    const { page, tag } = await getTagAndPage(componentTestSetup);
+    await page.setContent(`${componentTestSetup}`);
+    const element = await page.find(tag);
```
jcfranco, line 1348: Disabling animations might improve test robustness.

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -1323,3 +1323,67 @@ export async function t9n(componentTestSetup: ComponentTestSetup): Promise<void>
...
+    const beforeCloseEvent = page.waitForEvent(`${camelCaseTag}BeforeClose`);
+    const closeEvent = page.waitForEvent(`${camelCaseTag}Close`);
+
+    const beforeCloseSpy = await element.spyOnEvent(`${camelCaseTag}BeforeClose`);
+    const closeSpy = await element.spyOnEvent(`${camelCaseTag}Close`);
+
+    element.setProperty(toggleProp, true);
+    await page.waitForChanges();
+
+    const container = await page.find(`calcite-dropdown-group >>> ${containerSelector}`);
```
jcfranco, line 1367: Fixme: Avoid hardcoded selectors.

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -1323,3 +1323,67 @@ export async function t9n(componentTestSetup: ComponentTestSetup): Promise<void>
...
+    await page.waitForChanges();
+
+    expect(await container.isVisible()).toBe(true);
+
+    await beforeOpenEvent;
+    await openEvent;
+
+    expect(beforeOpenSpy).toHaveReceivedEventTimes(1);
+    expect(openSpy).toHaveReceivedEventTimes(1);
+
```
jcfranco, line 1377: Can you also assert on the close events not being called?

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -1323,3 +1323,67 @@ export async function t9n(componentTestSetup: ComponentTestSetup): Promise<void>
...
+    element.setProperty(toggleProp, false);
+    await page.waitForChanges();
+
+    expect(await container.isVisible()).toBe(false);
+
+    await beforeCloseEvent;
+    await closeEvent;
+
+    expect(beforeCloseSpy).toHaveReceivedEventTimes(1);
+    expect(closeSpy).toHaveReceivedEventTimes(1);
```
jcfranco, line 1387: Can you assert on the open events not being called one more time?

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -1323,3 +1323,67 @@ export async function t9n(componentTestSetup: ComponentTestSetup): Promise<void>
     await page.waitForChanges();
   }
 }
+
+/*
+ * @param {ComponentTestSetup} ComponentTestSetup - A component tag, html, or the tag and e2e page for setting up a test.
+ * @param {string} toggleProp - Toggle property to test. Currently, either "open" or "expanded".
+ * @param {string} containerSelector - Container element selector to test for visibility.
```
jcfranco, line 1330: Testing for visibility seems out of scope for this test helper. `onToggleOpenCloseComponent` helps components easily emit respective open/close events, but doesn't handle visibility, right?

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -1323,3 +1323,67 @@ export async function t9n(componentTestSetup: ComponentTestSetup): Promise<void>
...
+    const beforeOpenSpy = await element.spyOnEvent(`${camelCaseTag}BeforeOpen`);
+    const openSpy = await element.spyOnEvent(`${camelCaseTag}Open`);
+
+    const beforeCloseEvent = page.waitForEvent(`${camelCaseTag}BeforeClose`);
+    const closeEvent = page.waitForEvent(`${camelCaseTag}Close`);
+
+    const beforeCloseSpy = await element.spyOnEvent(`${camelCaseTag}BeforeClose`);
+    const closeSpy = await element.spyOnEvent(`${camelCaseTag}Close`);
+
+    element.setProperty(toggleProp, true);
```
jcfranco, line 1364: In order to simplify this helper's usage, can you get the initial value of `toggleProp` dynamically and toggle accordingly? Otherwise, this requires the HTML to always be closed. I realize this may be most common setup, but it'd be more flexible to allow both options.

jcfranco (changes requested):
This is looking pretty sweet, @Elijbet! 🍬

Can you wire up this helper in other open/close component test suites?

packages/calcite-components/src/tests/commonTests.ts
jcfranco: I forgot to add if you could look into also adding test coverage for when animations are disabled completely? Our animation disabling util technically still allows for transitions to happen (it uses a tiny duration), so you'll need to explicitly set the animation duration to 0.

Elijbet:
> This is looking pretty sweet, @Elijbet! 🍬
> 
> Can you wire up this helper in other open/close component test suites?

Is it ok if I wire up this helper in other components separately from this PR?

I think it's going to bloat up the PR and make it hard to merge. I was hoping to address each case or several cases in a bunch within continuing PRs, so I can progressively adjust the util.

packages/calcite-components/src/tests/commonTests.ts
Elijbet: Does this work or did I miss what you meant?

packages/calcite-components/src/tests/commonTests.ts
Elijbet: Found a better way!

jcfranco:
> Is it ok if I wire up this helper in other components separately from this PR?

I'd prefer to have the test helper cover all existing use cases from the get go. There's 6 other components that would need to be included, so I don't think it'll add much bloat since we're replacing any existing tests.

packages/calcite-components/src/tests/commonTests.ts
jcfranco: Latest test name is perfect, thanks!

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -1323,3 +1323,63 @@ export async function t9n(componentTestSetup: ComponentTestSetup): Promise<void>
...
+ *
+ * @example
+ * import { emitsOpenCloseTransitionChainedEvents } from "../../tests/commonTests";
+ *
+ * describe("emitsOpenCloseTransitionChainedEvents", () => {
+ *   emitsOpenCloseTransitionChainedEvents("calcite-dropdown", "open", ".container");
+ * });
+ *
+ */
+export function emitsOpenCloseEvents(
```
jcfranco, line 1340: It might be better to avoid implementation specifics in the helper name (namely, `emits` and `events`). This would allow the test helper to cover additional scenarios if needed without having to rename. WDYT about `openClose`? This is going by our current helper naming, 

* `formAssociated`
* `labelable`
* `focusable`
* `reflects`
* `defaults`
* `hidden`
* `accessible`
* `slots`
* `floatingUIOwner`
* `t9n`

packages/calcite-components/src/tests/commonTests.ts
Elijbet: Right. We're passing the selector in to `expect(await container.isVisible()).toBe(true); ` before we await for the events. I rewrote it based on some previous comments to pass this instead.     
`emitsOpenCloseEvents(simpleDropdownHTML, "open", `${CSS.calciteDropdownGroup} >>> .${CSS.container}`);`
Does that look ok? I can rewrite that bit to say `Container element selector.`

packages/calcite-components/src/tests/commonTests.ts
Elijbet: Sure, what additional scenarios are you thinking about?

jcfranco: I don't have specific scenarios in mind, but if another use case comes up, we don't have to worry about renaming the util.

packages/calcite-components/src/components/dropdown/resources.ts
```diff
@@ -4,4 +4,6 @@ export const SLOTS = {
 
 export const CSS = {
   calciteDropdownContent: "calcite-dropdown-content",
+  calciteDropdownGroup: "calcite-dropdown-group",
```
jcfranco, line 7: These classes don't appear to belong here as they are not used in the component's rendering.

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -1360,7 +1360,7 @@ export function emitsOpenCloseEvents(
     element.setProperty(toggleProp, true);
     await page.waitForChanges();
 
-    const container = await page.find(`calcite-dropdown-group >>> ${containerSelector}`);
+    const container = await page.find(containerSelector);
```
jcfranco, line 1363: I think we can focus solely on the events and assume the element will be visible (or not visible) to avoid checking visibility. This should make the helper simpler to use.

packages/calcite-components/src/components/modal/modal.e2e.ts
```diff
@@ -13,6 +13,14 @@ describe("calcite-modal properties", () => {
     hidden("calcite-modal");
   });
 
+  describe("openClose", () => {
+    openClose("<calcite-modal></calcite-modal>", "open", false);
```
jcfranco, line 17: If you update `openClose` to use `simplePageSetup` (similar to other test helpers), it will allow devs to set up a test by passing the tag directly in this case. 

```ts
openClose("calcite-modal", "open", false);
```

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -1341,3 +1345,124 @@ export async function t9n(componentTestSetup: ComponentTestSetup): Promise<void>
...
+ *  openTooltip: (page: E2EPage) => Promise<void>;
+ *  closeTooltip: (page: E2EPage) => Promise<void>;
+ * }} userInputDevice - Optional argument with functions to simulate user input (mouse or keyboard), to open or close the component.
+ */
+
+export function openClose(
+  componentTestSetup: ComponentTestSetup,
+  toggleProp: string,
+  toggleValue: boolean,
+  userInputDevice?: {
```
jcfranco, line 1381: Is there another name we could use for this? Not sure I would understand what `userInputDevice` would be used for without going into the code.

jcfranco, line 1381: Can you look at introducing an `options` object? If you look at other test utils, additional props not required for the test usually go in an `options` argument.

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -1341,3 +1345,124 @@ export async function t9n(componentTestSetup: ComponentTestSetup): Promise<void>
...
+ * @param {ComponentTestSetup} ComponentTestSetup - A component tag, html, or the tag and e2e page for setting up a test.
+ * @param {string} toggleProp - Toggle property to test. Currently, either "open" or "expanded".
+ * @param {boolean} toggleValue - Indicates the initial value of the toggle property.
+ * @param {{
+ *  openTooltip: (page: E2EPage) => Promise<void>;
+ *  closeTooltip: (page: E2EPage) => Promise<void>;
+ * }} userInputDevice - Optional argument with functions to simulate user input (mouse or keyboard), to open or close the component.
+ */
+
+export function openClose(
```
jcfranco, line 1377: Can you make sure the the body of this function is typed as much as possible?

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -30,9 +30,13 @@ function isHTML(tagOrHTML: string): boolean {
 
 function getTag(tagOrHTML: string): ComponentTag {
   if (isHTML(tagOrHTML)) {
-    const regex = /[>\s]/;
+    const regex = /<calcite-[a-z0-9-]+/i;
```
jcfranco, line 33: * `regex` was there already, but it might be good to rename it to express what it'll match.
* `match` could also be renamed to better convey its purpose.
* The error message could have more info on why it didn't match (e.g., no calcite-component tag used). Also, WDYT about throwing if it finds more than 1? Might not happen much, but if it does, it'll be useful to have this early warning when using the helper.

jcfranco (commented):
Looking awesome, @Elijbet! 😎

Elijbet:
Cool, so far mostly syntax comments. I'll work on that and modal failing tests next.

Elijbet:
This should resolve the failing modal test issue #7273.

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -1376,3 +1380,165 @@ export async function t9n(componentTestSetup: ComponentTestSetup): Promise<void>
...
+ */
+
+type userInputDevice = {
+  open: (page: E2EPage) => Promise<void>;
+  close: (page: E2EPage) => Promise<void>;
+};
+
+export async function openClose(
+  componentTagOrHTML: TagOrHTML,
+  toggleProp = "open",
```
jcfranco, line 1416: `open` is the default, but it is still being specified in all tests. 

Most of our test helpers provide arguments for required values, and extra options are provided in an options parameters. I think we should do this to simplify this helper's usage:

```ts
interface OpenCloseOptions {
  /**
   * ...
   */
  initialToggleValue?: boolean;
  
  /**
   * ...
   */
  toggleProp?: string;

  /**
   * ...
   */
  userInputDevice?: UserInputDevice;
}

export async function openClose(
  componentTagOrHTML: TagOrHTML,
  options?: OpenCloseOptions
)
```

The main benefit we get from this is that we avoid passing defaults when we need to use `userInputDevice`:

```ts
openClose("calcite-magic", {
  userInputDevice: { /* ... */ }
});
```

vs

```ts
openClose("calcite-magic", "open", false, {
  userInputDevice: { /* ... */ }
});
```

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -1376,3 +1380,165 @@ export async function t9n(componentTestSetup: ComponentTestSetup): Promise<void>
...
+ *   },
+ * });
+ *
+ * @param {string} componentTagOrHTML - the component tag or HTML markup to test against
+ * @param {string} toggleProp - Toggle property to test. Currently, either "open" or "expanded".
+ * @param {boolean} initialToggleValue - Indicates the initial value of the toggle property to determine whether to configure a `simplePageSetup` or `newProgrammaticE2EPage`.
+ * @param {userInputDevice} userInputDevice - Optional argument with functions to simulate user input (mouse or keyboard), to open or close the component.
+ */
+
+type userInputDevice = {
```
jcfranco, line 1409: Please add JSDoc to the interface similar to other test util options. Applies to all test-helper-related types/interfaces.

packages/calcite-components/src/tests/commonTests.ts
jcfranco: On a related note, I think some parameter names could be renamed to better convey their intent. This is just me, but this is how I'm currently reading them:

* `toggleProp` – most likely from `toggle`, but this can read as a boolean to enable toggling
* `initialToggleValue` – the initial value before toggling takes place or to toggle with
* `userInputDevice` – an object to specify how the user will interact with the component

Let's sync up on this to hash out some names.

packages/calcite-components/src/components/tooltip/tooltip.e2e.ts
```diff
@@ -680,7 +762,7 @@ describe("calcite-tooltip", () => {
   describe("beforeOpen, open, beforeClose, close event emitting", () => {
```
jcfranco, line 762: Can these be replaced by using the helper?

packages/calcite-components/src/components/action-bar/action-bar.tsx
```diff
@@ -407,8 +407,8 @@ export class ActionBar
     return (
       <calcite-action-group
         class={CSS.actionGroupEnd}
-        label={actionsEndGroupLabel}
         hidden={this.expandDisabled && !(this.hasActionsEnd || this.hasBottomActions)}
+        label={actionsEndGroupLabel}
```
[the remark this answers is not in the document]
Elijbet: Not sure how this happened TBH, and now I think have to add the `eslint-disable` for it to move back.

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -1376,3 +1380,204 @@ export async function t9n(componentTestSetup: ComponentTestSetup): Promise<void>
...
+interface OpenCloseOptions {
+  /**
+   * Toggle property to test. Currently, either "open" or "expanded".
+   */
+  openPropName?: string;
+
+  /**
+   * Indicates the initial value of the toggle property to determine whether to configure a `simplePageSetup` or `newProgrammaticE2EPage`.
+   */
+  initialToggleValue?: boolean;
```
jcfranco, line 1405: Synced up w/ @Elijbet and we decided to see if we could eliminate the need for this prop by looking up the open prop name dynamically from the test HTML or component (separate PR).

jcfranco (approved):
🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧
🔧🔧🧪🧪🧪🔧🔧🧪🧪🔧🔧🔧🧪🧪🔧🔧🧪🔧🔧🔧🧪🔧
🔧🧪🔧🔧🔧🔧🧪🔧🔧🧪🔧🧪🔧🔧🧪🔧🧪🔧🔧🔧🧪🔧
🔧🧪🔧🔧🔧🔧🧪🔧🔧🧪🔧🧪🔧🔧🧪🔧🧪🔧🔧🔧🧪🔧
🔧🧪🔧🔧🔧🔧🧪🔧🔧🧪🔧🧪🔧🔧🧪🔧🧪🔧🔧🔧🔧🔧
🔧🔧🧪🧪🧪🔧🔧🧪🧪🔧🔧🔧🧪🧪🔧🔧🧪🧪🧪🔧🧪🔧
🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧🔧

Issue #7386, opened 2023-07-27
[Menu/Menu Item] Doubling up borders when last nested Menu Item is a dropdown


SkyeSeitz:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When the last `menu-item` in a nested `menu-item` is itself a dropdown, two borders double up with the last `menu-item` open, as seen below:

<img width="269" alt="image" src="https://github.com/Esri/calcite-design-system/assets/29716057/6521830c-d88a-424e-83f2-2d9764d0ff01">

### Expected Behavior

I would expect one border to be present- the longest border associated with the top parent `menu-item`

### Reproduction Sample

https://codepen.io/skyeseitz/pen/eYQPoOG

### Reproduction Steps

1. Open **Vegetables** Menu Item
2. Open **Peppers** Menu Item
3. Double border is below **Habanero**

### Reproduction Version

v1.4.3

### Priority impact

p4 - not time sensitive

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

Calcite (design)

geospatialem:
Verified in `1.5.0-next.28`.

PR #7387, opened 2023-07-28
fix(menu-item): Prevent duplicate border in nested vertical menu-items

macandcheese:
**Related Issue:** #7386

## Summary
Adds css rule to prevent duplicate border in certain nested cases when the last item itself is a parent of other Menu Item components.

Before:
Screenshot 2023-07-27 at 1 16 11 PM

After:
Screenshot 2023-07-27 at 1 16 19 PM


benelan (approved):
:+1:

PR #7390, opened 2023-07-28
feat(switch): updates focus outline to be rounded

anveshmekala:
**Related Issue:** #4633 

## Summary

This PR will replace the block focus outline of `calcite-switch` with a rounded one.

Before the change:

25C47646-4B10-45A7-94E3-3CD884D35370_4_5005_c

After the change:

92B6D3BC-5B7E-4041-B424-BAB2F89E3C32_4_5005_c

### DIFF
--- a/packages/calcite-components/src/components/switch/switch.stories.ts
+++ b/packages/calcite-components/src/components/switch/switch.stories.ts
@@ -35,3 +35,14 @@ export const darkModeRTL_TestOnly = (): string => html`
 darkModeRTL_TestOnly.parameters = { modes: modesDarkDefault };
 
 export const disabled_TestOnly = (): string => html`<calcite-switch disabled checked></calcite-switch>`;
+
+export const Focus_TestOnly = (): string =>
+  html` <div>
+    <calcite-switch> </calcite-switch>
+    <script>
+      (async () => {
+        await customElements.whenDefined("calcite-date-picker");
+        await document.querySelector("calcite-switch").focus();
+      })();
+    </script>
+  </div>`;
--- a/packages/calcite-components/src/components/switch/switch.stories.ts
+++ b/packages/calcite-components/src/components/switch/switch.stories.ts
@@ -37,12 +37,12 @@ darkModeRTL_TestOnly.parameters = { modes: modesDarkDefault };
 export const disabled_TestOnly = (): string => html`<calcite-switch disabled checked></calcite-switch>`;
 
 export const Focus_TestOnly = (): string =>
-  html` <div>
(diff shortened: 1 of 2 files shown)


driskull (approved):
👍 

Can we add a screenshot test? It would require some JS but might be worth it if we can.

packages/calcite-components/src/components/switch/switch.stories.ts
```diff
@@ -35,3 +35,14 @@ export const darkModeRTL_TestOnly = (): string => html`
 darkModeRTL_TestOnly.parameters = { modes: modesDarkDefault };
 
 export const disabled_TestOnly = (): string => html`<calcite-switch disabled checked></calcite-switch>`;
+
+export const Focus_TestOnly = (): string =>
```
jcfranco, line 39: I think you may need to add a delay to the story to ensure the screenshot is taken when the test setup is done. 

@benelan do you have Any additional tips?

packages/calcite-components/src/components/switch/switch.stories.ts
```diff
@@ -35,3 +35,14 @@ export const darkModeRTL_TestOnly = (): string => html`
...
+  html` <div>
+    <calcite-switch> </calcite-switch>
+    <script>
+      (async () => {
```
jcfranco, line 43: Should this check for the switch being defined and not the date picker?

packages/calcite-components/src/components/switch/switch.stories.ts
```diff
@@ -35,3 +35,14 @@ export const darkModeRTL_TestOnly = (): string => html`
 darkModeRTL_TestOnly.parameters = { modes: modesDarkDefault };
 
 export const disabled_TestOnly = (): string => html`<calcite-switch disabled checked></calcite-switch>`;
+
+export const Focus_TestOnly = (): string =>
+  html` <div>
```
jcfranco, line 40: Nitpick: is the parent div needed here?

packages/calcite-components/src/components/switch/switch.stories.ts
jcfranco: It’s currently not showing the focus outline. 

1FE0E813-3A0C-4D40-BFDC-2F690F88BF8D

jcfranco (approved):

packages/calcite-components/src/components/switch/switch.stories.ts
anveshmekala: side effects.

PR #7392, opened 2023-07-29
test(input-time-zone): ensure time zone options are captured in the screenshot

jcfranco:
**Related Issue:** N/A

## Summary

Updates story parameters so the time zone options are captured.


driskull (approved):
👍

Issue #7005, opened 2023-05-23
[Color Picker] Hue thumb handle partially hidden under swatch


geospatialem:
### Actual Behavior

The thumb handle on the `color-picker` is partially hidden under the swatch.

![image](https://github.com/Esri/calcite-components/assets/5023024/b17d6c45-1b1b-4a29-a4d3-12e7ac517de6)

### Expected Behavior

The thumb should fall within the slider bounds, but still map to the edges.

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/color-picker/#sample

### Reproduction Steps

1. Open the [sample](https://developers.arcgis.com/calcite-design-system/components/color-picker/#sample)
2. Observe when the color is to the far left the handle is partially covered by the swatch

### Reproduction Version

1.4.0

### Regression?

1.3.1

### Priority impact

p2 - want for current milestone

### Esri team

ArcGIS Map Viewer

ashetland:
Discovered some related issues that are present at each scale, but most pronounced on `scale="s"`. See the attached screen recording.

- Using a mouse, the farthest right extent of the sliders cannot be reached
- The opacity slider doesn't respond to arrow keys
- Hue slider nitpick: I think a more expected UX is for the hue slider to range from 0 to 360, not 0 to 359.

https://github.com/Esri/calcite-components/assets/108549080/5f010f73-6042-4bf7-80cd-06deb070a2c0

geospatialem:
Verified in `1.5.0-next.36` 

![image](https://github.com/Esri/calcite-design-system/assets/5023024/8b77fcbe-f1cf-4044-9b48-d72d0656eb46)

PR #7398, opened 2023-07-31
fix(color-picker): draw slider thumbs within bounds

jcfranco:
**Related Issue:** #7005

## Summary

This addresses slider thumbs appearing cut off.

Screenshot 2023-07-31 at 2 44 23 PM
Screenshot 2023-07-31 at 2 44 07 PM

This also ensures the max hue slider value doesn't overlap with 0 and adds a util to map a value between ranges.

### DIFF
--- a/packages/calcite-components/src/components/color-picker/color-picker.tsx
+++ b/packages/calcite-components/src/components/color-picker/color-picker.tsx
@@ -23,6 +23,7 @@ import {
   DEFAULT_STORAGE_KEY_PREFIX,
   DIMENSIONS,
   HSV_LIMITS,
+  HUE_LIMIT_CONSTRAINED,
   OPACITY_LIMITS,
   RGB_LIMITS,
   SCOPE_SIZE,
@@ -1095,7 +1096,7 @@ export class ColorPicker
         slider: { width },
       },
     } = this;
-    const hue = ((HSV_LIMITS.h - 1) / width) * x;
+    const hue = (HUE_LIMIT_CONSTRAINED / width) * x;
 
     this.internalColorSet(this.baseColorFieldColor.hue(hue), false);
   }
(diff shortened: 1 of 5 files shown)


packages/calcite-components/src/components/color-picker/color-picker.tsx
```diff
@@ -1095,7 +1095,7 @@ export class ColorPicker
         slider: { width },
       },
     } = this;
-    const hue = (360 / width) * x;
+    const hue = ((HSV_LIMITS.h - 1) / width) * x;
```
anveshmekala, line 1098: Nitpick: can we assign `1` to a self explanatory constant since its is used in other places or have a property which stores `HSV_LIMITS.h -1`

packages/calcite-components/src/components/color-picker/color-picker.tsx
```diff
@@ -1565,12 +1573,13 @@ export class ColorPicker
 
     const x = alphaToOpacity(hsvColor.alpha()) / (OPACITY_LIMITS.max / width);
     const y = radius;
+    const sliderBoundX = remap(x, 0, width, radius, width - radius);
```
anveshmekala, line 1576: This will cause the thumb to not center on `pointerup` event. `sliderBoundX` is remapped on to a new range which is adjusting the thumb to move away from the center. Applies to opacity slider also.

anveshmekala (changes requested):
Need to tweak a little to make sure we center the thumb on `pointerup`

packages/calcite-components/src/components/color-picker/color-picker.tsx
jcfranco: I'm not following. Both thumbs and scopes are centered. 🤔

One thing to note is that the scopes will look misaligned in non-HD screens (this is currently how it's rendering on `@next`). Is this what you're observing?

### HD

Screenshot 2023-08-02 at 1 28 27 PM
Screenshot 2023-08-02 at 1 28 31 PM
Screenshot 2023-08-02 at 1 28 34 PM

### Standard

Screenshot 2023-08-02 at 1 29 10 PM
Screenshot 2023-08-02 at 1 29 14 PM
Screenshot 2023-08-02 at 1 29 17 PM

jcfranco: Confirmed w/ @anveshmekala that there's an issue when sampling colors individually vs sliding. Thanks for clarifying!

I'll look into only mapping if the x lies within the slider edges.

anveshmekala (approved):
🥇

Issue #2249, opened 2021-06-02
New Component: Meter


macandcheese:
### Description
A consistently formatted and styled meter component that leverages the native meter element.

#### User Stories
Consistently display data limits, quotas, disc space, etc. 

Can also be useful for item completeness, profile "filled out" progression messaging, etc.

### Helpful Details
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meter

Adobe Spectrum's implementation: https://spectrum.adobe.com/page/meter/#Anatomy

They don't seem to leverage the native "levels" that are associated with color, but they DO have consistently placed and formatted descriptive and value text areas.

- Ensure it works both with / without the "step" attributes (always one color OR stepped color based on those "steps")

Online "item completeness": 
![Screen Shot 2021-06-07 at 9 49 17 AM](https://user-images.githubusercontent.com/4733155/121059584-7badfb00-c776-11eb-8b53-3d89c3566095.png)

Enterprise Manager :
![Bitmap](https://user-images.githubusercontent.com/4733155/121067383-d6982000-c77f-11eb-8209-ef860a5202b6.png)

bstifle:
gonna jump on this design this week

paulcpederson:
Here's a couple other versions of this UI from Online:  

<img width="368" alt="Screen Shot 2021-06-07 at 10 48 25 AM" src="https://user-images.githubusercontent.com/1031758/121065743-f595b280-c77d-11eb-807b-272774bf2457.png">

<img width="502" alt="Screen Shot 2021-06-07 at 10 48 09 AM" src="https://user-images.githubusercontent.com/1031758/121065772-faf2fd00-c77d-11eb-9407-18a6c68947ba.png">

bstifle:
![image](https://user-images.githubusercontent.com/14083530/121101447-788c2c80-c7c1-11eb-8038-616e6256122e.png)
![image](https://user-images.githubusercontent.com/14083530/121101458-7f1aa400-c7c1-11eb-8d51-41d073836219.png)
![image](https://user-images.githubusercontent.com/14083530/121101487-8b066600-c7c1-11eb-84a7-4ab3a80886e4.png)

Initial designs. any thoughts/additions/concerns lemme know!

bstifle:
Leaving discussion open for this until next Wednesday 6/16

bstifle:
Alrighty! Moving this one over to developer hands

driskull:
[...]

PR #7401, opened 2023-08-01
feat(meter): Add Meter component

macandcheese:
**Related Issue:** #2249 

## Summary
Adds new Meter component with a variety of display capabilities.

https://github.com/Esri/calcite-design-system/assets/4733155/dfca6fa0-ebb9-45d4-b97d-f749270559b4


packages/calcite-components/src/components/meter/meter.scss
```diff
@@ -0,0 +1,126 @@
...
+  color: var(--calcite-ui-text-1);
+}
+
+.meter-label-range {
+  inset-block-start: calc(100% + 0.5em);
+  color: var(--calcite-ui-text-2);
+}
+
+.meter-unit-label {
+  font-weight: 500;
```
driskull, line 75: Is there a tailwind value for this?

packages/calcite-components/src/components/meter/meter.scss
```diff
@@ -0,0 +1,126 @@
...
+}
+
+.meter-label {
+  position: absolute;
+  font-size: var(--calcite-meter-font-size-internal);
+}
+
+.meter-label-value {
+  inset-block-end: calc(100% + 0.5em);
+  font-weight: 600;
```
driskull, line 65: Is there a tailwind value for this?

packages/calcite-components/src/components/meter/meter.scss
```diff
@@ -0,0 +1,126 @@
...
+  display: block;
+  position: absolute;
+  z-index: 200;
+  background-color: var(--calcite-ui-brand);
+  inset-inline-start: var(--calcite-meter-space-internal);
+  inset-block: var(--calcite-meter-space-internal);
+  border-radius: var(--calcite-meter-height-internal);
+  max-inline-size: calc(100% - (var(--calcite-meter-space-internal) * 2));
+  min-inline-size: calc(var(--calcite-meter-height-internal) - (var(--calcite-meter-space-internal) * 2));
+  transition: inline-size 0.25s ease-in-out, background-color 0.25s ease-in-out, box-shadow 0.25s ease-in-out;
```
driskull, line 93: Do we have a reduced motion equivalent?

packages/calcite-components/src/components/meter/meter.tsx
```diff
@@ -0,0 +1,291 @@
...
+  @Prop() numberingSystem: NumberingSystem;
+
+  /** When `true`, number values are displayed with a group separator corresponding to the language and country format. */
+  @Prop({ reflect: true }) groupSeparator = false;
+
+  /** The type of appearance of the meter */
+  @Prop() appearanceType: Extract<"outline" | "outline-fill" | "solid", Appearance> =
+    "outline-fill";
+
+  /** Determine if the meter should be displayed with a percent based fill, or based on discreet numbers. This will affect displayed labels */
```
geospatialem, line 71: WDYT about the following?:

```Specifies the component's unit type. When `displayValue` or `displayRangeLabels` is `true`, also specifies the component's labels.```

File: packages/calcite-components/src/components/meter/meter.stories.ts [hunk 5]
```diff
@@ -0,0 +1,76 @@
...
+
+export default {
+  title: "Components/Meter",
+  parameters: {
+    notes: readme,
+  },
+  ...storyFilters(),
+};
+
+export const simple = (): string => html`<calcite-meter min="0" max="100" low="25" high="75" />`;
```
driskull, line 13: Do we need closing tags here?

driskull: Applies to other usages without closing tags and not JSX

packages/calcite-components/src/components/meter/readme.md
```diff
@@ -0,0 +1 @@
+# calcite-meter
```
driskull, line 1: Should we add some `usage` files with examples?

packages/calcite-components/src/components/meter/meter.tsx
```diff
@@ -0,0 +1,291 @@
...
+  @Prop() labelType: "percent" | "units" = "percent";
+
+  /** A string optionally display beside the value and range values when not in `percent` `fillType` */
+  @Prop() unitLabel: "";
+
+  /** When true, displays the values of `high`, `low`, `min`, and `max` */
+  @Prop() displayRangeLabels: boolean;
+
+  /** When true, displays the current value */
+  @Prop() displayValue: boolean;
```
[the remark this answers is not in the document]
driskull: Maybe `valueLabel`?

driskull (commented):
Review round 1! 🥊

packages/calcite-components/src/components/meter/meter.tsx
```diff
@@ -0,0 +1,291 @@
...
+  NumberingSystem,
+  numberStringFormatter,
+} from "../../utils/locale";
+
+@Component({
+  tag: "calcite-meter",
+  styleUrl: "meter.scss",
+  shadow: true,
+})
+export class Meter implements LocalizedComponent {
```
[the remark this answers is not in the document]
driskull: - Since it has `disabled` I would think `InteractiveComponent` at minimum. 
- Since it has a `value` I would think maybe its a `FormComponent`?
- If it needs a `setFocus` method then `LoadableComponent`?
- If it can be wrapped in a `calcite-label` then `LabelableComponent`?

packages/calcite-components/src/components/meter/meter.tsx
```diff
@@ -0,0 +1,291 @@
...
+  /** Specifies the maximum possible value of the component. */
+  @Prop() max = 100;
+
+  /** Optionally provide a low value - above this, when `fillType` is 'range', the component display the appropriate color  */
+  @Prop() low: number;
+
+  /** Optionally provide a high value - above this, when `fillType` is 'range', the component display the appropriate color  */
+  @Prop() high: number;
+
+  /** Does the fill appear as a single "brand" color or does it display indications of low / high / "success", "danger", or "warning" based on low, high if provided, or value compared to max */
```
geospatialem, line 58: WDYT about something like:

```Specifies the component's display, where `"single"` displays a single "brand" color and `"range"` displays a range of colors which can specify `low`, `high`, `min` or `max` values with "success", "warning", and "danger".```

File: packages/calcite-components/src/components/meter/meter.tsx [hunk 10]
```diff
@@ -0,0 +1,291 @@
...
+    const displayLow = !!low && low > value && low > min && (!high || low < high);
+    const displayHigh = !!high && high > value && high < max && (!low || high > low);
+
+    return (
+      <Host>
+        <div
+          aria-label={label}
+          aria-valuemax={isPercent ? maxPercent : max}
+          aria-valuemin={isPercent ? minPercent : min}
+          aria-valuenow={isPercent ? currentPercent : value}
```
geospatialem, line 263: For some reason `aria-valuenow`/`aria-valuetext` isn't getting read back to AT, will need to dive into why this could be occurring tomorrow.

geospatialem (commented):
💥💥Great work!💥💥

A few doc-related considerations. 

Also, unable to hear JAWS reading out the value (NVDA works in FF), so thinking it might be related to the `aria-label` attribute. Will dive into the a11y tomorrow.

packages/calcite-components/src/components/meter/meter.tsx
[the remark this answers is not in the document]
macandcheese: Yeah, wasn’t sure on these. 

I don’t think it needs ‘disabled’ at all. I’ll remove that property and style.

It also isn’t interactive for a user, so while it can be placed inside a Label, a user can’t input or provide any value, so I’m not sure if it needs the Form / Labelable 🤔 WDYT?

packages/calcite-components/src/components/meter/meter.tsx
```diff
@@ -0,0 +1,291 @@
...
+    "outline-fill";
+
+  /** Determine if the meter should be displayed with a percent based fill, or based on discreet numbers. This will affect displayed labels */
+  @Prop() labelType: "percent" | "units" = "percent";
+
+  /** A string optionally display beside the value and range values when not in `percent` `fillType` */
+  @Prop() unitLabel: "";
+
+  /** When true, displays the values of `high`, `low`, `min`, and `max` */
+  @Prop() displayRangeLabels: boolean;
```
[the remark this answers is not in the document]
macandcheese: ‘rangeLabels’ works for me.

packages/calcite-components/src/components/meter/meter.tsx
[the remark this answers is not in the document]
macandcheese: ‘valueLabel’ works for me and matches the proposed ‘rangeLabels’ name change.

packages/calcite-components/src/components/meter/meter.tsx
```diff
@@ -0,0 +1,291 @@
...
+  /** Does the fill appear as a single "brand" color or does it display indications of low / high / "success", "danger", or "warning" based on low, high if provided, or value compared to max */
+  @Prop() fillType: "single" | "range" = "range";
+
+  /** Specifies the Unicode numeral system used by the component for localization. */
+  @Prop() numberingSystem: NumberingSystem;
+
+  /** When `true`, number values are displayed with a group separator corresponding to the language and country format. */
+  @Prop({ reflect: true }) groupSeparator = false;
+
+  /** The type of appearance of the meter */
```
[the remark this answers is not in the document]
macandcheese: Nope - updated.

packages/calcite-components/src/components/meter/meter.scss
macandcheese: Yep, updated.

packages/calcite-components/src/components/meter/meter.scss
macandcheese: Yep, updated.

packages/calcite-components/src/components/meter/meter.tsx
```diff
@@ -0,0 +1,291 @@
...
+  /** Specifies the current value of the component. */
+  @Prop() value: number;
+
+  /** Specifies the miniumum possible value of the component. */
+  @Prop() min = 0;
+
+  /** Specifies the maximum possible value of the component. */
+  @Prop() max = 100;
+
+  /** Optionally provide a low value - above this, when `fillType` is 'range', the component display the appropriate color  */
```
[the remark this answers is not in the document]
macandcheese: For `low` / `high` - they are displayed when `rangeLabels` is `true`, not only when `fillType="range"` - so would this work for both?:

`Specifies a <low/high> value.  When `fillType` is `"range"`, displays different color when <above/below> the specified threshold.`

packages/calcite-components/src/components/meter/meter.tsx
macandcheese: For brevity, what about this - I don't think listing brand / status colors are needed - those would be shown as tokens. 

`Specifies the component's display, where `"single"` displays a single color and `"range"` displays one of a range of colors when `low` or `high` are specified.

packages/calcite-components/src/components/meter/meter.tsx
macandcheese: Maybe `Specifies the format of displayed labels, when either "valueLabel" and/or "rangeLabels" are true.`

packages/calcite-components/src/components/meter/meter.tsx
```diff
@@ -0,0 +1,291 @@
...
+  @Prop({ reflect: true }) groupSeparator = false;
+
+  /** The type of appearance of the meter */
+  @Prop() appearanceType: Extract<"outline" | "outline-fill" | "solid", Appearance> =
+    "outline-fill";
+
+  /** Determine if the meter should be displayed with a percent based fill, or based on discreet numbers. This will affect displayed labels */
+  @Prop() labelType: "percent" | "units" = "percent";
+
+  /** A string optionally display beside the value and range values when not in `percent` `fillType` */
```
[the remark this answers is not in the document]
macandcheese: Will update - it is expected only the first "range label" has the unit added.

packages/calcite-components/src/components/meter/meter.stories.ts
macandcheese: Updated

packages/calcite-components/src/components/meter/meter.tsx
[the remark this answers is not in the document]
driskull: hmm not sure. The labelable component also helps to focus the component when the label is clicked with and without the "for" attribute.

packages/calcite-components/src/components/meter/meter.tsx
```diff
@@ -0,0 +1,284 @@
...
+    const { low, high, min, max, value } = this;
+    const lowest = low ? low : min;
+    const highest = high ? high : max;
+    const aboveLowest = value >= lowest;
+    const belowLowest = value < lowest;
+    const aboveHighest = value >= highest;
+    const belowHighest = value < highest;
+
+    if (!value || (!low && belowHighest) || belowLowest) {
+      return "success";
```
driskull, line 133: These are used as CSS classes right? if so should they be in the CSS object?

packages/calcite-components/src/components/meter/meter.tsx
geospatialem: Awesome! 🥇 

nitpick: on the display of props. Could backticks be added, in addition to the double quotes to all the values (e.g. `"single"`, `"range"`, etc.)?

packages/calcite-components/src/components/meter/meter.tsx
geospatialem: Could we flip your recommendation, for alignment with other docs where we specify the props conditions first? Something like:

```When either `valueLabel` and/or `rangeLabels` are `true`, specifies the format of displayed labels.```

File: packages/calcite-components/src/components/meter/meter.tsx (hunk 15 above)
Comment (macandcheese) [reply packages/calcite-components/src/components/meter/meter.tsx:133]: Yeah, I was using `Kind` type as return but will change to string type add to resources 👍

File: packages/calcite-components/src/components/meter/meter.scss (hunk 3 above)
Comment (driskull) [reply packages/calcite-components/src/components/meter/meter.scss:91]: Would there be tailwind values for the timing too?

Comment (macandcheese) [reply packages/calcite-components/src/components/meter/meter.scss:91]: Yeah need to clean this up, not sure the best way to only specify certain transitions with Tailwind like that. Will look into it.

File: packages/calcite-components/src/components/meter/readme.md (hunk 6 above)
Comment (macandcheese) [reply packages/calcite-components/src/components/meter/readme.md:1]: Yep, added.

File: packages/calcite-components/src/components/meter/meter.tsx (hunk 10 above)
Comment (geospatialem) [reply packages/calcite-components/src/components/meter/meter.tsx:256]: Was able to diagnose the issue, the `aria-valuetext` is interfering with the `aria-valuenow` context for AT. If needed for context, could we add the `aria-valuetext`  value as part of `aria-label` so the `aria-valuenow` is provided to AT as the value changes?

Comment (macandcheese) [reply packages/calcite-components/src/components/meter/meter.tsx:256]: I think we can add the value to the `valuenow` to have it read as "50 percent", or "2,400 credits" (from slider example: https://www.w3.org/WAI/ARIA/apg/patterns/slider/examples/slider-temperature/)

Comment (geospatialem) [reply packages/calcite-components/src/components/meter/meter.tsx:259]: For what we're aiming for this looks good to go a11y-wise. 🚀 

NVDA does lose a "beep" indicator that content has changed in the page (when only `aria-valuenow` is provided over `aria-valuetext`). But that seems minor so the unit context is provided to AT. Nice work! 💪🏻

Comment (macandcheese) [reply packages/calcite-components/src/components/meter/meter.tsx:259]: Does the above case occur when `labelType="unit"` and there is no `unitLabel` provided? 

If so, we could consider having a translated default of "units" if we always want `aria-valuetext` to have some value?

File: packages/calcite-components/src/components/meter/meter.tsx [hunk 16]
```diff
@@ -0,0 +1,291 @@
...
+    const displayHigh = !!high && high > value && high < max && (!low || high > low);
+
+    return (
+      <Host>
+        <div
+          aria-label={label}
+          aria-valuemax={isPercent ? maxPercent : max}
+          aria-valuemin={isPercent ? minPercent : min}
+          aria-valuenow={isPercent ? currentPercent : value}
+          // todo translate "percent"
```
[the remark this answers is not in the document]
macandcheese: Removed todo - added T9n support and populated the "en" message for use in future translations.

macandcheese:
@geospatialem Can you give this another look for a11y?

packages/calcite-components/src/components/meter/meter.tsx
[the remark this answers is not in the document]
macandcheese: Component has been updated, I still am not sure Form / Interactive / Labelable apply as the component itself won't accept input or be focusable.

geospatialem (commented):
> Can you give this another look for a11y?

a11y is good to go! ✨

packages/calcite-components/src/components/meter/meter.stories.ts
```diff
@@ -0,0 +1,240 @@
...
+    appearance="${select("appearance", ["solid", "outline", "outline-fill"], "single")}"
+    range-label-type="${select("range-label-type", ["percent", "units"], "units")}"
+    value-label-type="${select("value-label-type", ["percent", "units"], "percent")}"
+    unit-label="${text("unit-label", "credits")}"
+    ${boolean("group-separator", true)}
+    ${boolean("range-labels", true)}
+    ${boolean("value-label", true)}
+  ></calcite-meter>`;
+
+export const meterSwapLabelPlacementWhenCloseToMax_TestOnly = (): string => html`<calcite-meter
```
jcfranco, line 50: Nitpick: I think `meter` can be dropped from the name since all of these stories are meter-related.

packages/calcite-components/src/components/meter/meter.stories.ts
```diff
@@ -0,0 +1,240 @@
...
+    value-label
+    range-labels
+    min="0"
+    max="12400"
+    low="4600"
+    high="7600"
+    value="2200"
+  ></calcite-meter>`;
+
+export const meterComplexUnitRTL_TestOnly = (): string =>
```
jcfranco, line 228: Awesome screenshot test coverage! ✨🏆✨

packages/calcite-components/src/components/meter/resources.ts
```diff
@@ -0,0 +1,14 @@
+export const CSS = {
+  meter: "meter",
```
jcfranco, line 2: I think we can also simplify naming/classes here by dropping meter since it's the main context for this object.

packages/calcite-components/src/components/meter/usage/Advanced.md
```diff
@@ -0,0 +1,16 @@
+A complex meter component.
```
jcfranco, line 1: Not critical for this PR and more of a doc suggestion, but it might be good to describe what makes it complex or the use case it's representing. cc @geospatialem

packages/calcite-components/src/components/modal/assets/modal/t9n/messages_ar.json
```diff
@@ -1,3 +0,0 @@
-{
-  "close": "إغلاق"
```
jcfranco, line 2: This seems unintentional, can you restore `ar` and `et` modal translations?

packages/calcite-components/src/components/meter/meter.scss
```diff
@@ -0,0 +1,117 @@
...
+  color: var(--calcite-ui-text-3);
+}
+
+.meter-label-value .meter-unit-label {
+  color: var(--calcite-ui-text-2);
+  font-weight: var(--calcite-font-weight-bold);
+}
+
+.meter-fill {
+  @apply block absolute duration-150 ease-in-out bg-brand z-sticky;
```
jcfranco, line 78: Should this use the default z-index util instead? The sticky level seems too high, IMO.

packages/calcite-components/src/components/meter/meter.tsx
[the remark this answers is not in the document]
macandcheese: Cool. Can add `FormComponent` - WDYT the others?

packages/calcite-components/src/components/meter/meter.tsx
```diff
@@ -0,0 +1,522 @@
...
+  //--------------------------------------------------------------------------
+  //
+  //  Lifecycle
+  //
+  //--------------------------------------------------------------------------
+
+  async componentWillLoad(): Promise<void> {
+    await setUpMessages(this);
+    setUpLoadableComponent(this);
+    this.calculateValues();
```
jcfranco, line 146: Can `value`, `low`, `high` be computed if not specified at this point? This would align w/ `<meter>`'s behavior (see https://codepen.io/jcfranco/pen/dywyPQb?editors=1010).

This might also need to happen when certain props are modified at runtime.

This could be tackled in a follow-up PR if needed.

packages/calcite-components/src/components/meter/meter.scss
```diff
@@ -0,0 +1,117 @@
...
+
+.meter {
+  @apply flex relative items-center w-full m-0;
+  block-size: var(--calcite-meter-height-internal);
+  background-color: var(--calcite-ui-foreground-2);
+  border: 1px solid var(--calcite-ui-border-3);
+  border-radius: var(--calcite-meter-height-internal);
+}
+
+.meter.solid {
```
jcfranco, line 28: Would it be possible to use BEM-like classes along with the cascade to reduce specificity?

packages/calcite-components/src/components/meter/meter.e2e.ts
```diff
@@ -0,0 +1,19 @@
+import { accessible, renders, hidden, t9n } from "../../tests/commonTests";
+
+describe("calcite-meter", () => {
+  describe("renders", () => {
```
jcfranco, line 4: There might be other test helpers that could be used here (e.g., `defaults`, `reflects`).

packages/calcite-components/src/components/meter/meter.tsx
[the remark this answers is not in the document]
jcfranco: Agreed on not needing to implement the others.

jcfranco (approved):
Awesome stuff, @macandcheese! 

Any refactor/formatting comments unrelated to types, z-index and test helpers are not critical for this PR to land. If tackling separately, please create a follow-up issue/PR to keep track of remaining items.

📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏
📏🌡️📏📏📏🌡️📏📏🌡️🌡️📏📏🌡️🌡️🌡️📏📏🌡️🌡️🌡️📏🌡️🌡️🌡️🌡️📏🌡️📏
📏🌡️🌡️📏📏🌡️📏🌡️📏📏🌡️📏📏🌡️📏📏🌡️📏📏📏📏🌡️📏📏📏📏🌡️📏
📏🌡️📏🌡️📏🌡️📏🌡️📏📏🌡️📏📏🌡️📏📏🌡️📏📏📏📏🌡️🌡️🌡️📏📏🌡️📏
📏🌡️📏📏🌡️🌡️📏🌡️📏📏🌡️📏📏🌡️📏📏🌡️📏📏📏📏🌡️📏📏📏📏📏📏
📏🌡️📏📏📏🌡️📏📏🌡️🌡️📏📏🌡️🌡️🌡️📏📏🌡️🌡️🌡️📏🌡️🌡️🌡️🌡️📏🌡️📏
📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏

packages/calcite-components/src/components/meter/meter.tsx
macandcheese: To clarify - set `value` to "0" if not provided by a user?

Do we want to set an internal `low` and `high` even if not provided? 

Currently we don't have `optimum` as part of the API so the `low` and `high`, if unset by a user, are not used in any calculations, as they might be in native Meter.

packages/calcite-components/src/demos/meter.html
```diff
@@ -0,0 +1,885 @@
...
+
+        const controlType = document.getElementById("control-type");
+        const controlScale = document.getElementById("control-scale");
+        const controlValueLabelType = document.getElementById("control-value-label-type");
+        const controlValueRangeType = document.getElementById("control-range-label-type");
+
+        const controlAppearance = document.getElementById("control-appearance");
+
+        inputMin.addEventListener("calciteInputNumberInput", function () {
+          meter.min = event.target.value;
```
jcfranco, line 828: One more thing. Can you add an `event` param to each handler?

packages/calcite-components/src/components/meter/meter.tsx
jcfranco: > To clarify - set value to "0" if not provided by a user?

It should match `min`, if not provided.

> Do we want to set an internal low and high even if not provided?

I was thinking setting the props directly. Do you see an issue in doing so?

> Currently we don't have optimum as part of the API so the low and high, if unset by a user, are not used in any calculations, as they might be in native Meter (we could add optimum in the future and address this then as well).

Sorry about the confusion, I added `optimum` in my codepen for exhaustiveness.

packages/calcite-components/src/components/meter/meter.tsx
```diff
@@ -0,0 +1,522 @@
...
+  private formatLabel = (value: number, labelType: MeterLabelType): string => {
+    if (typeof value === "number" && labelType !== "percent") {
+      numberStringFormatter.numberFormatOptions = {
+        locale: this.effectiveLocale,
+        numberingSystem: this.numberingSystem,
+        useGrouping: this.groupSeparator,
+      };
+      return numberStringFormatter.localize(value.toString());
+    } else if (typeof value === "number" && labelType === "percent") {
+      return Intl.NumberFormat(this.effectiveLocale, {
```
jcfranco, line 260: Yet one more thing™ 😅, this may be called multiple times, so it'd be good to create and cache the formatter and reuse.

packages/calcite-components/src/components/meter/meter.tsx
```diff
@@ -0,0 +1,522 @@
...
+    "outline-fill";
+
+  /** When either `rangeLabels` is `true`, specifies the format of displayed labels. */
+  @Prop({ reflect: true }) rangeLabelType: MeterLabelType = "percent";
+
+  /** When either `valueLabel` is `true`, specifies the format of displayed label. */
+  @Prop({ reflect: true }) valueLabelType: MeterLabelType = "percent";
+
+  /** When `labelType` is `"units"` and either `valueLabel` or `rangeLabels` are `true`, displays beside the `value` and/or  `min` values. */
+  @Prop() unitLabel: "";
```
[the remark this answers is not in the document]
macandcheese: Nice catch!

packages/calcite-components/src/components/meter/meter.tsx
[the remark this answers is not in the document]
macandcheese: Implemented `FormComponent` and related props.

packages/calcite-components/src/components/meter/meter.tsx
macandcheese: Yeah, that makes sense. To clarify:

If no `low` provided, set `low` to `min`.
If no `high` provided, set `high` to `max`.
If no `value` provided, set `value` to `min`.

This can be set in `calculateValues()`, which is called in other component updates as well as `componentWillLoad()` - are there other cases we should check for?

packages/calcite-components/src/components/meter/meter.e2e.ts
macandcheese: Added!

packages/calcite-components/src/components/meter/meter.scss
macandcheese: Don't think we need this at all - removing.

packages/calcite-components/src/components/meter/meter.tsx
```diff
@@ -0,0 +1,522 @@
...
+    const minLowOverlap = this.comparePosition(minLabelEl, lowLabelEl);
+    const minHighOverlap = this.comparePosition(minLabelEl, highLabelEl);
+    const minMaxOverlap = this.comparePosition(minLabelEl, maxLabelEl);
+    const lowMaxOverlap = this.comparePosition(lowLabelEl, maxLabelEl);
+    const lowHighOverlap = this.comparePosition(lowLabelEl, highLabelEl);
+    const highMaxOverlap = this.comparePosition(highLabelEl, maxLabelEl);
+
+    if (lowLabelEl) {
+      const hideLowLabel = minLowOverlap || lowMaxOverlap || lowHighOverlap;
+      lowLabelEl.style.opacity = hideLowLabel ? "0" : "100";
```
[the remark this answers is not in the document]
macandcheese: Well it WAS because I was encountering issues using Tailwind's `@apply hidden` in a conditionally applied class. But taking another look, I see that I've been unintentionally setting `display: none` and not `visibility: hidden` - which Tailwind of course calls `invisible` instead 😄 

With it toggling `display: none` instead - it was messing with all the pixel width calculations we do for overlap. Of course - using the correct TW name makes it work fine, lol.

packages/calcite-components/src/components/meter/meter.tsx
```diff
@@ -0,0 +1,522 @@
...
+    return (
+      <div
+        class={{ [CSS.meterLabel]: true, [CSS.meterLabelValue]: true }}
+        key="low-label-line"
+        // eslint-disable-next-line react/jsx-sort-props
+        ref={(el) => (this.valueLabelEl = el as HTMLDivElement)}
+      >
+        {label}
+        {unitLabel && valueLabelType !== "percent" && (
+          <span class={CSS.meterUnitLabel}>&nbsp;{unitLabel}</span>
```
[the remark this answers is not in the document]
macandcheese: I believe so - see https://www.chromatic.com/test?appId=6266d45509d7eb004aa254fb&id=64dea7e4cacb0207a04160c8

Let me know if that is not a correct treatment in RTL. I think (?) the unit display would swap whereas the percent display would not?

packages/calcite-components/src/components/meter/meter.tsx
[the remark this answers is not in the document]
macandcheese: The other locations where styles are set inline are where we calculate position and set a width or position value, so I don't know if those would work in classes.

packages/calcite-components/src/components/meter/meter.tsx
macandcheese: Updated to the above - added some checks for "out of bounds" values, and updated with some tests to check that various "improper" configurations are correctly adjusted and set.

packages/calcite-components/src/components/meter/meter.tsx
```diff
@@ -0,0 +1,522 @@
...
+    const aboveHighest = value >= highest;
+    const belowHighest = value < highest;
+
+    if (!value || (!low && belowHighest) || belowLowest) {
+      return CSS.success;
+    } else if (aboveLowest && belowHighest) {
+      return CSS.warning;
+    } else if (aboveHighest) {
+      return CSS.danger;
+    }
```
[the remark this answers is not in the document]
macandcheese: Not intentionally at least, but I'll add an else case to return success.

packages/calcite-components/src/components/meter/meter.tsx
```diff
@@ -0,0 +1,522 @@
...
+    const { low, high, min, max, value } = this;
+    const lowest = low ? low : min;
+    const highest = high ? high : max;
+    const aboveLowest = value >= lowest;
+    const belowLowest = value < lowest;
+    const aboveHighest = value >= highest;
+    const belowHighest = value < highest;
+
+    if (!value || (!low && belowHighest) || belowLowest) {
+      return CSS.success;
```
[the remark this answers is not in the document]
macandcheese: I've updated this naming to reflect that this is only returning the css class. It is not set to any property, it's solely used to style the bar. Maybe we'll get request for exposing this as a readOnly property? Hesitant to add as it only applies to fillType:"range" and not "single".

packages/calcite-components/src/components/meter/usage/Advanced.md
macandcheese: Updated

geospatialem (approved):
Final a11y check looks good. 🟢

packages/calcite-components/src/components/meter/meter.tsx
macandcheese: Updated

ashetland:
Looking good on my end 🚀

packages/calcite-components/src/components/meter/meter.tsx
```diff
@@ -0,0 +1,543 @@
...
+    this.lowPercent = lowPercent;
+    this.highPercent = highPercent;
+    this.currentPercent = value ? currentPercent : 0;
+    this.lowActive = !!low && low > min && (!value || low > value) && (!high || low < high);
+    this.highActive =
+      !!high && min <= high && high < max && (!value || high > value) && (!low || high > low);
+  }
+
+  /**
+   *
```
jcfranco, line 254: This can probably be dropped since the function signature is pretty expressive and typed.

packages/calcite-components/src/components/meter/meter.tsx
```diff
@@ -0,0 +1,543 @@
...
+      return CSS.warning;
+    } else if (aboveHighest) {
+      return CSS.danger;
+    } else {
+      return CSS.success;
+    }
+  }
+
+  private intersects(el1: HTMLDivElement, el2: HTMLDivElement): boolean {
+    return el1 && el2
```
jcfranco, line 301: Suggestion: 

```ts
return el1 && el2 && intersects(el1.getBoundingClientRect(), el2.getBoundingClientRect());
```

You already guard against elements being defined, so no need for optional chaining. Sidebar: I think we could also improve the util to handle null to simplify this type of logic.  🤓

jcfranco (approved):
🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭
🍭🎚️🎚️🎚️🎚️🍭🎚️🍭🍭🍭🍭🍭🎚️🍭🎚️🎚️🎚️🎚️🍭🍭🎚️🎚️🎚️🍭🍭🎚️🎚️🍭🍭🎚️🍭🍭🍭🎚️🍭🎚️🎚️🎚️🎚️🍭🎚️🍭
🍭🎚️🍭🍭🎚️🍭🎚️🍭🍭🍭🍭🍭🎚️🍭🎚️🍭🍭🍭🍭🎚️🍭🍭🍭🍭🎚️🍭🍭🎚️🍭🎚️🎚️🍭🎚️🎚️🍭🎚️🍭🍭🍭🍭🎚️🍭
🍭🎚️🎚️🎚️🎚️🍭🎚️🍭🍭🎚️🍭🍭🎚️🍭🎚️🎚️🎚️🍭🍭🍭🎚️🎚️🍭🍭🎚️🍭🍭🎚️🍭🎚️🍭🎚️🍭🎚️🍭🎚️🎚️🎚️🍭🍭🎚️🍭
🍭🎚️🍭🍭🎚️🍭🎚️🍭🎚️🍭🎚️🍭🎚️🍭🎚️🍭🍭🍭🍭🍭🍭🍭🎚️🍭🎚️🍭🍭🎚️🍭🎚️🍭🍭🍭🎚️🍭🎚️🍭🍭🍭🍭🍭🍭
🍭🎚️🍭🍭🎚️🍭🍭🎚️🍭🍭🍭🎚️🍭🍭🎚️🎚️🎚️🎚️🍭🎚️🎚️🎚️🍭🍭🍭🎚️🎚️🍭🍭🎚️🍭🍭🍭🎚️🍭🎚️🎚️🎚️🎚️🍭🎚️🍭
🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭🍭

PR #7412, opened 2023-08-01
feat(text-area): provide additional context for AT users when character limit exceeds

anveshmekala:
**Related Issue:** #6630

### DIFF
--- a/packages/calcite-components/src/components/text-area/text-area.tsx
+++ b/packages/calcite-components/src/components/text-area/text-area.tsx
@@ -260,12 +260,12 @@ export class TextArea
     return (
       <Host>
         <textarea
-          aria-invalid={toAriaBoolean(this.value?.length > this.maxLength)}
+          aria-invalid={toAriaBoolean(this.isCharacterLimitExceeded())}
           aria-label={getLabelText(this)}
           autofocus={this.autofocus}
           class={{
             [CSS.readOnly]: this.readOnly,
-            [CSS.textAreaInvalid]: this.value?.length > this.maxLength,
+            [CSS.textAreaInvalid]: this.isCharacterLimitExceeded(),
             [CSS.footerSlotted]: this.endSlotHasElements && this.startSlotHasElements,
             [CSS.blockSizeFull]: !hasFooter,
             [CSS.borderColor]: !hasFooter,
@@ -317,6 +317,11 @@ export class TextArea
           {this.renderCharacterLimit()}
         </footer>
         <HiddenFormInputSlot component={this} />
+        {this.isCharacterLimitExceeded() && (
+          <span aria-hidden={true} aria-live="polite" class={CSS.assistiveText}>
+            {this.replacePlaceHoldersInMessages()}
+          </span>
+        )}
       </Host>
     );
   }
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/text-area/text-area.tsx
```diff
@@ -260,12 +260,12 @@ export class TextArea
     return (
       <Host>
         <textarea
-          aria-invalid={toAriaBoolean(this.value?.length > this.maxLength)}
+          aria-invalid={toAriaBoolean(this.isCharacterLimitExceeded())}
```
geospatialem, line 263: To mitigate with NVDA and FF, add in an `aria-describedby` to provide additional context.

packages/calcite-components/src/components/text-area/text-area.tsx
```diff
@@ -410,7 +415,7 @@ export class TextArea
       this.localizedCharacterLengthObj = this.getLocalizedCharacterLength();
       return (
         <span class={CSS.characterLimit}>
-          <span class={{ [CSS.characterOverLimit]: this.value?.length > this.maxLength }}>
+          <span class={{ [CSS.characterOverLimit]: this.isCharacterLimitExceeded() }}>
```
geospatialem, line 418: To support context with NVDA and FF, add a unique `id`, which can be referenced in the `aria-describedby` in the `textarea`, similar to [`combobox-item`](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/combobox-item/combobox-item.tsx#L228), where a guid is created.

geospatialem (changes requested):
This will be a fantastic addition, nice work, @anveshmekala! 🙌🏻 

Overall testing shows great support for JAWS in Chrome and Firefox and NVDA in Chrome.

There was some unexpected behavior with NVDA and Firefox, but with the effort working in Chrome. It does appear adding an `aria-describedby` to the `textarea` mitigates the issue and doesn't affect the fix to the other working combinations (JAWS and NVDA). 

Some screenshots of NVDA during review:

1. NVDA and FF:
nvda-ff

2. NVDA and Chrome:
nvda-chrome

Proposed fix:
- Add `aria-describedby` to the `textarea` referencing the unique `id` of the `span`
- Add a unique `id` to the `span`

After the fix in NVDA and FF:
image

packages/calcite-components/src/components/text-area/text-area.tsx
```diff
@@ -260,12 +261,13 @@ export class TextArea
...
         <textarea
-          aria-invalid={toAriaBoolean(this.value?.length > this.maxLength)}
+          aria-describedby={this.guid}
+          aria-invalid={toAriaBoolean(this.isCharacterLimitExceeded())}
           aria-label={getLabelText(this)}
           autofocus={this.autofocus}
           class={{
             [CSS.readOnly]: this.readOnly,
-            [CSS.textAreaInvalid]: this.value?.length > this.maxLength,
+            [CSS.textAreaInvalid]: this.isCharacterLimitExceeded(),
```
jcfranco, line 270: Nitpick: this is called a few times during rendering, so maybe this could be a state prop computed when `value` or `maxLength` changes.

packages/calcite-components/src/components/text-area/text-area.tsx
```diff
@@ -514,4 +502,29 @@ export class TextArea
       footerWidth,
     };
   }
+
+  private replacePlaceHoldersInMessages(): string {
```
jcfranco, line 506: Nitpick: was the move of this function intentional?

jcfranco (approved):
Aside from a11y tweaks, code changes LGTM!

geospatialem (approved):
Thanks for putting in these changes so quickly! ✨ 🚀 ✨

packages/calcite-components/src/components/text-area/text-area.tsx
anveshmekala: yup, this is intentional to move all the private method to the end.

Issue #7402, opened 2023-08-01
[Switch] When used within Label, focus outline incorrect


macandcheese:
### Check existing issues

- [ ] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When used in a `calcite-label`, the focus outline of `calcite-switch` extends past the element. It may be affected by other parent elements and their styles, but noticed it within a `calcite-label`.

```
<calcite-label>
  <calcite-switch />
</calcite-label>
```

<img width="1728" alt="Screenshot 2023-07-31 at 8 15 02 PM" src="https://github.com/Esri/calcite-design-system/assets/4733155/4461b6f7-0bb9-40a1-9d11-393ca7f77e1e">

### Expected Behavior

I would expect the Outline to hug the `calcite-switch`

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/VwVqXwP?editors=1100

### Reproduction Steps

1. Open Codepen
2. Click in Label to show focus outline

### Reproduction Version

1.5.0

### Regression?

Yes

### Priority impact

p2 - want for current milestone

### Impact

Will be pretty noticeable in most label + switch layouts.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

Calcite (design)

geospatialem:
There are a few workarounds for others that may have encountered the above issue:

1. Adding a `layout="inline"`:
```html
  <calcite-label layout="inline">
    <calcite-switch />
  </calcite-label>
```

2. Adding a `for` attribute:
 ```html
  <calcite-label for="switchEl">A label</calcite-label>
    <calcite-switch id="switchEl" />
  
```

geospatialem:
Verified in `1.5.0-next.31`.

PR #7414, opened 2023-08-01
fix(switch): Fix for focus outline style in certain cases

macandcheese:
**Related Issue:** #7402 

## Summary
Fix for focus issue where container was getting the focus outline. With this change, the component is still taking up 100% width of the Label, but the focus outline is limited to the visual area. We could look at adjusting the click area in a separate update if needed.

Before:
Screenshot 2023-08-01 at 11 54 31 AM

After:
Screenshot 2023-08-01 at 11 54 11 AM

### DIFF
--- a/packages/calcite-components/src/components/switch/switch.scss
+++ b/packages/calcite-components/src/components/switch/switch.scss
@@ -47,13 +47,8 @@
 
 @include disabled();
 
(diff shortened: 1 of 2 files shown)


driskull (approved):
👍

Issue #7417, opened 2023-08-01
Chip not cleaning up its MutationObserver when disconnected


jcfranco:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Memory leak when adding/removing chips.

via @ycabon

### Expected Behavior

Chips properly clean up after being removed from the DOM.

### Reproduction Sample

https://codepen.io/jcfranco/pen/JjexPVj

### Reproduction Steps

1. open devtools memory tab
2. click "add chip" then "remove chip" to warm up all the statics and what not
3. take a memory snapshot
4. click "add chip" then "remove chip"
5. take another snapshot
6. compare snapshot 2 with 1

### Reproduction Version

1.4.3

### Priority impact

p1 - need for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Maps SDK for JavaScript

jcfranco:
Even though the observer is properly cleaned up on removal, I think there's something else going on that could be causing a memory leak (possibly w/ Stencil). Even adding and removing a very simple component yields a bump in memory.

geospatialem:
Verified the mutation observer cleanup in `main`. ✔️ 

@jcfranco Do we need anew  research issue to determine what could be happening with the memory leak?

PR #7418, opened 2023-08-02
fix(chip): disconnect mutation observer when component is disconnected from the DOM

jcfranco:
**Related Issue:** #7417 

## Summary

✨🧹✨


driskull (approved):
👍

Issue #7396, opened 2023-07-31
Tooltip emitting too many open/close events


jcfranco:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Multiple events are emitted before open/close, and on open/close.

### Expected Behavior

A single event is emitted before open/close, and on open/close.

### Reproduction Sample

https://codepen.io/jcfranco/pen/PoxXjWe?editors=1000

### Reproduction Steps

Hover over the link to display the tooltip, then hover out to have the tooltip hide.

Note there's more than 1 open/close event emitted when this happens.

### Reproduction Version

next (1.5.0)

### Regression?

1.4.3

### Priority impact

p1 - need for current milestone

### Impact

This will cause tooltip's open/close event handlers to be invoked multiple times vs the expected 1.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

Calcite (dev)

geospatialem:
Verified on `next` with 4 events. ✔️

PR #7422, opened 2023-08-02
fix(tooltip): avoid extra before open/close event emitting

jcfranco:
**Related Issue:** #7396

## Summary

Updates `TooltipManager` logic to avoid toggling hovered tooltips if they are the active ones.

### DIFF
--- a/packages/calcite-components/src/components/tooltip/TooltipManager.ts
+++ b/packages/calcite-components/src/components/tooltip/TooltipManager.ts
@@ -208,6 +208,7 @@ export default class TooltipManager {
   }
 
   private openHoveredTooltip = (tooltip: HTMLCalciteTooltipElement): void => {
+    this.clearHoverOpenTimeout();
     this.hoverOpenTimeout = window.setTimeout(
       () => {
         if (this.hoverOpenTimeout === null) {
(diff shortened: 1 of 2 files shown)


Elijbet:
Tested with tooltip demos: hover over a trigger that is visible or set to `display: none` (with the new `onToggleOpenClose` refactors and so far it seems to be emitting only once for each before open/close)

packages/calcite-components/src/components/tooltip/TooltipManager.ts
```diff
@@ -215,6 +215,11 @@ export default class TooltipManager {
         }
 
         this.clearHoverCloseTimeout();
+
+        if (this.activeTooltip === this.hoveredTooltip) {
+          return;
```
driskull, line 220: ✨ clean ✨

driskull (approved):
Awesome!

Can you update the PR body with the changes? Still has the 🚧 sign posted. 👷

packages/calcite-components/src/components/tooltip/TooltipManager.ts
Elijbet: 🧹 🫧

Issue #6517, opened 2023-02-24
[calcite-action-bar] bottom-actions slot name not consistent with similar slots


jcfranco:
### Actual Behavior

All action slot names are named `actions-<placement>`, and `bottom-actions` does not follow this pattern.

### Expected Behavior

`bottom-actions` to be renamed `actions-bottom`.

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/action-bar/#api-reference-slots-bottom-actions

### Reproduction Steps

👀 slot name

### Reproduction Version

1.0.7

### Relevant Info

This name was introduced when the component got moved from the calcite-app-components repo and the name probably didn't get revisited since.

Not urgent, but we can deprecate the existing slot and introduce the new one. 

### Priority impact

p3 - not time sensitive

### Esri team

Calcite (dev)

brittneytewks:
Removed "needs triage" since a priority is set and estimate is indicated

geospatialem:
Verified in `1.6.0-next.3` with: 

```html
<calcite-action-group slot="actions-end">
```

PR #7435, opened 2023-08-03
feat(action-bar): add "actions-end" slot (deprecates "bottom-actions")

driskull:
**Related Issue:** #6517

## Summary

- Add "actions-end" slot (deprecates "bottom-actions")
- Update tests
- Update stories
- Update demos
- Update usage
- Update CSS


jcfranco (approved):
🎉

jcfranco:
Since this is using `feat`, I think the message should be rephrased as `add "actions-end" slot (deprecates "bottom-actions")` (or similar) to emphasize the feature vs the deprecation.

Issue #6375, opened 2023-01-30
[Slider] Unexpected behavior when programmatically updating props


macandcheese:
### Actual Behavior

When programmatically updating slider, the changes don't accurately reflect in the component:

https://user-images.githubusercontent.com/4733155/215389203-c45ffe12-c1c2-46d4-8921-a9e5cf88e1dc.mov

### Expected Behavior

I'd expect the Slider to accurately reflect the updated property values

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/gOjjjZO?editors=1000

### Reproduction Steps

1. Open Codepen
2. Use the buttons to adjust the Slider component
3. See component not reflect changes correctly.

### Reproduction Version

1.0.3

### Regression?

Could not find a working previous version

### Esri team

Calcite (design)

geospatialem:
Verified on the `main` branch.

PR #7439, opened 2023-08-03
fix(slider): rerender ticks when prop is modified

jcfranco:
**Related Issue:** #6375

## Summary

This adds a watcher for `ticks` to update slider ticks.


driskull (approved):
👍

Issue #7437, opened 2023-08-03
Panel header and action-bar z-indexing display issue


driskull:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The panel header and action-bar slotted inside a panel have the same z-index stacking number.

This makes displaying tooltips or opening menus not work correctly as seen below

![image](https://github.com/Esri/calcite-design-system/assets/1231455/286b56e9-7017-47df-a7e9-f56bd9f95b24)

### Expected Behavior

There should be a way to have tooltips or menus work on both the action-bar and panel header so that one is not covering the other.

### Reproduction Sample

https://codepen.io/driskull/pen/RwqdwQg

### Reproduction Steps

1. Open the codepen
2. Open the overflow menu in the top right header of the panel
3. Notice the menu is displayed behind the action-bar
4. There is no way to adjust the z-index of the action-bar or header because they share the same CSS var for z-index. (`--calcite-app-z-index-header`)

### Reproduction Version

1.4.3

### Relevant Info

N/A

### Regression?

No

### Priority impact

p3 - want for upcoming milestone

### Impact

This is affecting adding tooltips to the header actions for the Popup in the Maps SDK.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Maps SDK for JavaScript

geospatialem:
Verified in `1.5.1-next.1`.

PR #7440, opened 2023-08-03
fix(panel): Correct header and action-bar z-indexing display issues

driskull:
**Related Issue:** #7437

## Summary

- Places action-bar within the header so the z-index isn't an issue anymore
- Adjust styles as necessary
- Adjust rendering as necessary
- Adds screenshot test


jcfranco (approved):
Noice! 
⬆️
↕️
⬇️

PR #7441, opened 2023-08-03
ci: remove duplicate action for notifying teams about releases

benelan:
**Related Issue:** #7317

## Summary
I thought I had removed this original teams notification action but it must have been added back while I was fixing merge conflicts. This one caused two notifications, one for each package. The one added in the PR linked above will only notify about CC releases.


jcfranco (approved):
🤖💬💬❌

Issue #7436, opened 2023-08-03
flow-item back button tooltip at wrong location initially


AdelheidF:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

![image](https://user-images.githubusercontent.com/44481292/215876929-ed5b2c4b-0ca9-4aeb-a6eb-0be267e93ace.png)

After creating a second flow-item the focus moves to the back button and the tooltip initially displays at the top left corner.

leftover from https://github.com/Esri/calcite-design-system/issues/6400

### Expected Behavior

tooltip is always next to the back button

### Reproduction Sample

https://codepen.io/afreitag/pen/rNQRNzE

### Reproduction Steps

open Codepen
Click on blue button and watch top left corner.

### Reproduction Version

1.5.0-next.36

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Map Viewer

geospatialem:
Verified in `1.5.1-next.1` with the updated native tooltip.

![image](https://github.com/Esri/calcite-design-system/assets/5023024/391d6d73-f8cb-466f-a27c-36e6c0605f19)

Issue #7433, opened 2023-08-03
[calcite-flow-item] set tooltip to open bottom by default


lboyd93:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

The back button on the flow item component has a tooltip that opens above by default. We would like to request the tooltip to open below by default so that we can avoid this issue in the Map Viewer and the JS Maps SDK.

![flow-item-back-tooltip](https://github.com/Esri/calcite-design-system/assets/22916889/bfb4f400-fc35-48b3-9956-b55b75c84ec4)

cc @driskull 

### Acceptance Criteria

Setting the back button's tooltip to show below by default.

### Which Component

[Calcite Flow Item](https://next.sites.afd.arcgis.com/calcite-design-system/components/flow-item/)

### Priority impact

p2 - want for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Maps SDK for JavaScript

geospatialem:
Verified in `1.5.1-next.1`, where the placement will be below the mouse cursor. Context is provided to AT users, but will not be present via keyboard keys.

![verify-title-bottom-default](https://github.com/Esri/calcite-design-system/assets/5023024/55b892a2-ea9e-4f97-8942-0d65919fe0b0)

PR #7442, opened 2023-08-03
fix(flow-item): Use a native tooltip for the back button

driskull:
**Related Issue:** #7436 #7433

## Summary

- Use a native tooltip for the back button. #7436
- This is the only component that uses a tooltip by default so it seems like we shouldn't do this
- A native tooltip probably makes more sense as the back icon is pretty recognizable


driskull:
@macandcheese can you review this one?

macandcheese (approved):
Nice!
Screenshot 2023-08-04 at 9 43 19 AM

.. Also, that's just the text in the demo page, I didn't write it for you specifically 😄  @driskull

Issue #7451, opened 2023-08-04
[Action Bar / Action Pad] Use native tooltip if no slotted Tooltip provided


macandcheese:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Currently, in order to get the expand / collapse string (default or overridden) to display, a user needs to slot a tooltip. We should still allow that, but by default adding a native tooltip / title would be an improvement.

### Acceptance Criteria

Native tooltip exists on expand / collapse button by default - users shouldn't have to slot a tooltip (although they can - in which case - it overrides the native tooltip).

### Relevant Info

Aligns with https://github.com/Esri/calcite-design-system/pull/7442

### Which Component

Action Bar / Pad

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

Calcite (design)

geospatialem:
Verified in `1.5.1-next.2` with `action-bar` and `action-pad`.

action bar example:
![image](https://github.com/Esri/calcite-design-system/assets/5023024/099b16a8-ffd0-490e-93a0-162467c5bfdd)

PR #7452, opened 2023-08-04
fix(action-bar, action-pad): Add native tooltip to expand action

driskull:
**Related Issue:** #7451

## Summary

- Add native tooltip to expand action when no tooltip is slotted.


macandcheese (approved):
⏩ Very fast. Nice!

Issue #7453, opened 2023-08-04
Calcite-panel missing display:flex on content-wrapper


JonUihlein:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

For JSAPI.

FeatureTable isn't appearing as expected when embedded in a simple calcite-panel:
https://codepen.io/jonuihlein/pen/wvQOqeO?editors=1000

Background:
> Originally reported from Map Viewer team, but reproducible with simple sample above.
> 
> It looks like in there was an internal "content" node inside calcite-panel that had the content-container class removed (there in 1.4.2) which was responsible for setting the content height to 100%.
> 
> I feel like we went through something similar when Editor was using calcite-panel over calcite-flow but were able to ignore the issue as a result of switching over to flow. However, the above case has worked as expected in the map viewer up until this point so it is likely our users may run into this as well.

Spoke with @driskull and he identified that "content-wrapper" is missing `display: flex`. 

### Expected Behavior

content node should have display: flex;

### Reproduction Sample

https://codepen.io/jonuihlein/pen/wvQOqeO?editors=1000

### Reproduction Steps

Open sample, table does not appear as expected. 
Change version to 1.4.2, table appears.

### Reproduction Version

1.5.0

### Regression?

1.4.3

### Priority impact

p1 - need for current milestone

### Impact

FeatureTable does not work in AGOL Map Viewer

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Maps SDK for JavaScript

subgan82:
Please add ArcGIS charts project as well we are impacted by it , is there a CSS workaround ?

<img width="715" alt="2023-08-04_15-03-41" src="https://github.com/Esri/calcite-design-system/assets/1017712/3ec0068b-46b0-4c06-b971-2e46f006247b">

paulcpederson:
Adding Online as we are also running into this

driskull:
This would be a good one to test on the next build once its deployed.

geospatialem:
Verified in `1.5.1-next.2`

![image](https://github.com/Esri/calcite-design-system/assets/5023024/664d9534-49ba-4b28-a88e-5816db0723e4)

PR #7454, opened 2023-08-04
fix(panel): Allow panel content to take full height.

driskull:
**Related Issue:** #7453

## Summary

- Combine wrapper and container nodes/classes to simplify implementation.
- Fixes full height content for children of panel
- It seems that the `<section>` isn't necessary.
- Add stories


jcfranco (approved):
🎉

Awesome job on the improved screenshot coverage. 📸

Issue #6059, opened 2022-12-15
Audit components to remove `getSlotted` util instances


macandcheese:
### Description

Many components use `getSlotted` and will need to be refactored in favor of a new pattern.

### Proposed Advantages

The `getSlotted` util is no longer a recommended pattern - `onSlotChange` is preferred. 

### Which Component

**Note: prioritize `card` effort in the initial stages first, if at all possible -- related issue: https://github.com/Esri/calcite-design-system/issues/10152**

- [x] `accordion-item` #10447
- [x] `action-bar`
- [x] `action-group`
- [x] `action-pad`
- [x] `alert` #10448
- [x] `block` #10449
- [x] `card` ❗  - prioritize in October 2024
- [x] `combobox-item` #10450
- [x] `inline-editable` #10456
- [x] `input` #10458
- [x] `input-number` #10458
- [x] `input-text` #10458
- [x] ~`modal`~ (deprecated) #10469
- [x] `notice` #10453
- [x] ~`pick-list-group`~ (deprecated)
- [x] ~`pick-list-item`~ (deprecated)
- [x] `shell`
- [x] `shell-center-row` #10451
- [x] `shell-panel`
- [x] `table`
- [x] `tile`
- [x] `tile-group`
- [x] ~`tip`~ (deprecated) #10470
- [x] `tree-item` #10462
- [x] ~`value-list-item`~ (deprecated)

### Relevant Info

cc @driskull

driskull:
This one would be nice to get onto a milestone sometime in the next year.

rweber-esri:
See my [comment](https://github.com/Esri/calcite-design-system/issues/10152#issuecomment-2368479741) on [10152](https://github.com/Esri/calcite-design-system/issues/10152). Hub is observing slotted content not working in `calcite-card` even when we _immediately_ slot the content into the component during initial component render on `2.12.1`. This is observed when rendering `calcite-card` from other Stencil components (JSX) Hub internally develops/maintains that are lazy loaded (may play a factor) in our Ember application. Not sure if that influences priority, but we seem to be unable to reliably slot certain content in `calcite-card` at this time.

driskull:
@geospatialem @DitwanP besides `card` which are next highest priority? Should we prioritize this list for future implementation?

jcfranco:
@driskull's been crushing it with this audit! Some of the associated PRs are fixes, so we might need to adjust this issue to accommodate for testing and verification. cc @geospatialem @DitwanP

[...]

PR #7464, opened 2023-08-04
refactor(action-bar, action-pad, action-group): Remove usage of getSlotted utility

driskull:
**Related Issue:** #6059

## Summary

- Depends on #7435


packages/calcite-components/src/components/action-bar/action-bar.e2e.ts
```diff
@@ -165,10 +165,12 @@ describe("calcite-action-bar", () => {
       const page = await newE2EPage();
 
       await page.setContent(html`<calcite-action-bar expand-disabled></calcite-action-bar>`);
+      await page.waitForChanges();
 
       const buttonGroup = await page.find(`calcite-action-bar >>> .${CSS.actionGroupEnd}`);
 
-      expect(buttonGroup).toBeNull();
+      expect(buttonGroup.getAttribute("hidden")).toBe("");
```
jcfranco, line 172: For consistency, can you check the prop instead?

packages/calcite-components/src/components/action-bar/action-bar.tsx
```diff
@@ -345,6 +351,18 @@ export class ActionBar
     this.setGroupLayout(groups);
   };
 
+  handleActionsEndSlotChange = (event: Event): void => {
+    this.hasActionsEnd = slotChangeHasAssignedElement(event);
+  };
+
+  handleTooltipSlotChange = (event: Event): void => {
+    const tooltips = slotChangeGetAssignedElements(event).filter((el) =>
+      el?.matches("calcite-tooltip")
```
jcfranco, line 360: Just curious, when would this be `null`?

jcfranco (approved):
This LGTM and also looks low-risk.

packages/calcite-components/src/components/action-bar/action-bar.tsx
driskull: are you talking about `el`? If there are no assigned elements.

packages/calcite-components/src/components/action-bar/action-bar.e2e.ts
driskull: @jcfranco I can't do this I don't think?

image

driskull: Can't use getProperty as its not a component.

image

packages/calcite-components/src/components/action-bar/action-bar.tsx
jcfranco: > If there are no assigned elements.

If it returns an empty array for this case, the `filter` callback won't be invoked.

packages/calcite-components/src/components/action-bar/action-bar.e2e.ts
jcfranco: Maybe it's a doc thing, but we already do this in [`tab-title.e2e.ts`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden).

driskull: yeah must be a doc thing.

Issue #7448, opened 2023-08-04
`calcite-accordion-item` that is not direct child of `calcite-accordion` does not expand when clicked


ethanbdev:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When using the `calcite-accordion-item` standalone or with a shadow boundary between it and the parent `calcite-accordion`, it does not expand on click.

> use case is we have a component that uses shadow dom wrapping each accordion item, so they need to look up the accordion parent with a closestAcrossShadowDOM Looks like they are just assuming the parent element will be an accordion and that is not the case.

### Expected Behavior

It should expand when clicked

### Reproduction Sample

https://codepen.io/eborgen/pen/eYQXReo?editors=1000

### Reproduction Steps

1. Try to expand the accordion item and see that it does not

### Reproduction Version

1.5.0

### Regression?

Yes - 1.4.3

### Priority impact

p0 - emergency

### Impact

This is preventing the item browser from working correctly pending a refactor of the dom structure. Its okay to accomodate in a breaking change release, but there is no immediate way to support this + older calcite versions.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Map Viewer

ethanbdev:
cc: @paulcpederson

jcfranco:
@ethanbdev Can you confirm that the repro case up to date? Earlier versions show the same behavior. Went as far back as 1.0.3.

ethanbdev:
> @ethanbdev Can you confirm that the repro case up to date? Earlier versions show the same behavior. Went as far back as 1.0.3.

Yes, I think the problem is actually the shadow boundary between them, let me work on that

ethanbdev:
@jcfranco repro case updated

jcfranco:
Thanks. It looks like the regression was introduced by https://github.com/Esri/calcite-design-system/commit/2b09abacf4f265866f44f210a8c83b580e865fdf. cc @Elijbet

geospatialem:
Verified in `1.5.1-next.3`.

PR #7466, opened 2023-08-05
fix(accordion): restore accordion-items working with accordion across shadow DOM

jcfranco:
**Related Issue:** #7448 

## Summary

This updates accordion to sync its children's properties across shadow boundaries via events.

### DIFF
--- a/packages/calcite-components/src/components/accordion-item/accordion-item.tsx
+++ b/packages/calcite-components/src/components/accordion-item/accordion-item.tsx
@@ -119,15 +119,10 @@ export class AccordionItem implements ConditionalSlotComponent {
 
   connectedCallback(): void {
     connectConditionalSlotComponent(this);
-    document.addEventListener("calciteInternalAccordionItemsSync", this.updateAccordionChildSync);
   }
 
   disconnectedCallback(): void {
     disconnectConditionalSlotComponent(this);
-    document.removeEventListener(
-      "calciteInternalAccordionItemsSync",
-      this.updateAccordionChildSync
-    );
   }
 
   // --------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/accordion-item/accordion-item.tsx
```diff
@@ -243,12 +241,35 @@ export class AccordionItem implements ConditionalSlotComponent {
 
   @Listen("calciteInternalAccordionChange", { target: "body" })
```
driskull, line 242: maybe this one should be targeted to document for consistency? 👍

driskull: or vice versa?

jcfranco: I'll create an issue to audit all listeners not targeting the element itself and possibly define some guidelines. I think `document` makes sense for a direct custom event bus where we don't need capturing nor bubbling.

driskull (approved):
👍 🥼 🔬 👀 Looks good

packages/calcite-components/src/components/accordion-item/accordion-item.tsx
```diff
@@ -243,12 +241,42 @@ export class AccordionItem implements ConditionalSlotComponent {
...
+    this.determineActiveItem(parent.selectionMode, event.detail.requestedAccordionItem);
+    event.stopPropagation();
+  }
+
+  @Listen("calciteInternalAccordionItemsSync", { target: "document" })
+  updateAccordionChildSync(event: CustomEvent): void {
+    const [accordion] = event.composedPath();
+    const accordionItem = this.el;
+
+    const willBeSyncedByDirectParent = accordionItem.parentElement === accordion;
```
driskull, line 263: I think the name for this const is great but maybe a link to an issue or more info might be useful in a comment. What do you think?

Ideally, we can just do things one way and this will be cleaned up in the future.

driskull (approved):
👍 Looks good!

packages/calcite-components/src/components/accordion-item/accordion-item.tsx
jcfranco: Good idea. I'll submit a follow-up PR for this to expedite this fix landing.

PR #7470, opened 2023-08-07
fix(modal): handle removal of open attribute and prevent multiple beforeClose calls

driskull:
**Related Issue:** #6407 #6379

## Summary

- Add internal property `opened` to maintain open state when beforeBack is rejected and to handle initial animation.
- Refactor `beforeClose` default value.
- Rename internal close method name to be consistent with internal open method name
- Prevent `beforeClose` from being called twice by adding an internal flag to ignore watch changes when necessary.
- Tests


packages/calcite-components/src/components/modal/modal.e2e.ts
```diff
@@ -123,14 +123,40 @@ describe("calcite-modal properties", () => {
         (elm.beforeClose = (window as typeof window & Pick<typeof elm, "beforeClose">).beforeClose)
     );
     await page.waitForChanges();
-    await modal.setProperty("open", true);
```
alisonailea, line 126: is there any way to fix this without changing the property name? A prop name change is technically a breaking change.

driskull: This isn't a breaking change, no prop is being removed.

alisonailea (approved):

driskull:
@jcfranco can you review this one?

packages/calcite-components/src/components/modal/modal.tsx
```diff
@@ -343,6 +342,15 @@ export class Modal
   //
   //--------------------------------------------------------------------------
 
+  /**
+   * We use an internal property to handle styles for when a modal is actually opened, not just when the open attribute is applied. This is a property because we need to apply styles to the host element and to keep the styles present while beforeClose is .
```
jcfranco, line 346: Nitpicks: 

* `...while beforeClose is .` ➡️  `...while beforeClose is.`
* `@internal.` ➡️ `@internal`

packages/calcite-components/src/components/modal/modal.e2e.ts
```diff
@@ -123,14 +123,40 @@ describe("calcite-modal properties", () => {
...
+  const page = await newE2EPage();
+  const mockCallBack = jest.fn();
+  await page.exposeFunction("beforeClose", mockCallBack);
+  await page.setContent(`
+    <calcite-modal open></calcite-modal>
+  `);
+  const modal = await page.find("calcite-modal");
+  await page.$eval(
+    "calcite-modal",
+    (elm: HTMLCalciteModalElement) =>
```
jcfranco, line 147: Nitpick: `elm` ➡️ `el` for consistency.

Sidebar, I think we can have ESLint catch/fix this one.

packages/calcite-components/src/components/modal/modal.e2e.ts
```diff
@@ -123,14 +123,40 @@ describe("calcite-modal properties", () => {
...
+      (elm.beforeClose = (window as typeof window & Pick<typeof elm, "beforeClose">).beforeClose)
+  );
+  await page.waitForChanges();
+  modal.setProperty("open", true);
+  await page.waitForChanges();
+  expect(await modal.getProperty("opened")).toBe(true);
```
jcfranco, line 153: `opened` is internal, so maybe we can avoid asserting on it. I'll leave this up to you.

packages/calcite-components/src/components/modal/modal.e2e.ts
```diff
@@ -123,14 +123,40 @@ describe("calcite-modal properties", () => {
...
+  const mockCallBack = jest.fn();
+  await page.exposeFunction("beforeClose", mockCallBack);
+  await page.setContent(`
+    <calcite-modal open></calcite-modal>
+  `);
+  const modal = await page.find("calcite-modal");
+  await page.$eval(
+    "calcite-modal",
+    (elm: HTMLCalciteModalElement) =>
+      (elm.beforeClose = (window as typeof window & Pick<typeof elm, "beforeClose">).beforeClose)
```
jcfranco, line 148: Suggestion: I _think_ you could use our global test type for this `GlobalTestProps<{ beforeClose: HTMLCalciteModalElement["beforeClose"] }>`.

packages/calcite-components/src/components/modal/modal.tsx
```diff
@@ -343,6 +342,15 @@ export class Modal
...
+   *
+   * @internal.
+   */
+  @Prop({ mutable: true, reflect: true }) opened = false;
```
jcfranco, line 350: For consistency, can we move this to properties despite it being internal?

packages/calcite-components/src/components/modal/modal.e2e.ts
```diff
@@ -109,7 +109,7 @@ describe("calcite-modal properties", () => {
     expect(styleH).toEqual("800px");
   });
 
-  it("calls the beforeClose method prior to closing", async () => {
+  it("calls the beforeClose method prior to closing via click", async () => {
```
jcfranco, line 112: Can you add a similar test covering closing via `Escape` key?

packages/calcite-components/src/components/modal/modal.tsx
```diff
@@ -87,8 +87,7 @@ export class Modal
   @Prop({ mutable: true, reflect: true }) open = false;
 
   /** Passes a function to run before the component closes. */
-  @Prop()
-  beforeClose: (el: HTMLElement) => Promise<void> = () => Promise.resolve();
+  @Prop() beforeClose: (el: HTMLCalciteModalElement) => Promise<void>;
```
jcfranco, line 90: For posterity, this is removing the default value, but we should be OK as it's [documented as `undefined`](https://developers.arcgis.com/calcite-design-system/components/modal/#api-reference-properties-beforeClose).

jcfranco (approved):
Awesome! Nice solution and clean to boot! ✨🥾✨

packages/calcite-components/src/components/modal/modal.e2e.ts
driskull: Replaced all var names. I think someone else named these but yeah eslint would be good.

packages/calcite-components/src/components/modal/modal.e2e.ts
driskull: I'll leave it as it ensures the prop is being added.

Issue #7151, opened 2023-06-07
[calcite-slider] If the group separator is space, the number will be cut off at group separators


fendiaoxiaoshuazi:
### Actual Behavior

Language like ru uses space as the group separator, and when the space is used as the group separator, the number will be cut off in between.
![image](https://github.com/Esri/calcite-components/assets/26078106/a9af51b4-bbb0-4bf3-82cd-b1de0fdb9381)
![image](https://github.com/Esri/calcite-components/assets/26078106/47c007f4-6997-4794-9e36-417e7bfef9ef)

### Expected Behavior

Even the space is used as the group separator, the number should not be broken into multiple lines.

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/slider/

### Reproduction Steps

1. Enable group separator
2. Change the locale to ru

### Reproduction Version

v1.4.1

### Priority impact

p4 - not time sensitive

### Esri team

ArcGIS Survey123

geospatialem:
Verified in `1.5.1-next.4`.

![image](https://github.com/Esri/calcite-design-system/assets/5023024/aa17f68e-8b24-46e3-9a3d-23a687574565)

PR #7472, opened 2023-08-07
fix(slider): numbers remain on one line for locales with space group separators

benelan:
**Related Issue:** #7151

## Summary

Numbers with a space group separator are breaking onto multiple lines. This replaces the space with a non-breaking space (nbsp) HTML entity.

ref: https://en.wikipedia.org/wiki/Non-breaking_space


packages/calcite-components/src/utils/locale.ts
```diff
@@ -409,8 +409,8 @@ export class NumberStringFormat {
     } as Intl.NumberFormatOptions).formatToParts(-12345678.9);
 
     this._actualGroup = parts.find((d) => d.type === "group").value;
-    // change whitespace group characters that don't render correctly
-    this._group = this._actualGroup.trim().length === 0 ? " " : this._actualGroup;
+    // change whitespace group characters that don't render correctly to a unicode no-break space
+    this._group = this._actualGroup.trim().length === 0 || this._actualGroup == "\u0020" ? "\u00A0" : this._actualGroup;
```
driskull, line 413: It might be prettier to use html entities here instead of unicode. What do you think?

```
this._actualGroup === " " ? "&nbsp;" : this._actualGroup;
```

benelan: Yeah I agree that's nicer for the nbsp, but I like being very explicit about what is in `" "`, since there are a ton of white space characters that could have been copy/pasted in there. And then I didn't want to mix and match unicode and HTML entities. But I don't feel strongly about it if you'd like me to switch. Also story incoming

driskull (approved):
✨ Beautiful ✨ 👍

Issue #7475, opened 2023-08-07
Input clearable not rendering in CC React


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The `input` isn't loading when using `clearable` and ` value` of undefined.
![image](https://github.com/Esri/calcite-design-system/assets/5023024/a2cf5164-8de4-4b5f-b451-5c71fc669326)

### Expected Behavior

![image](https://github.com/Esri/calcite-design-system/assets/5023024/91a8ef00-a34a-4810-a4ee-2765136b0cdb)

### Reproduction Sample

https://codesandbox.io/s/input-clearable-error-1-5-forked-ykqyq8?file=/src/App.js

### Reproduction Steps

1. Open the [sample](https://codesandbox.io/s/input-clearable-error-1-5-forked-ykqyq8?file=/src/App.js)
2. Observe the error message
3. Open the [1.4.0 sample](https://codesandbox.io/s/input-clearable-error-1-5-forked-mqmfn5?file=/src/App.js)
4. Observe no error message

### Reproduction Version

1.5.0

### Relevant Info

[1.5.0](https://codesandbox.io/s/input-clearable-error-1-5-forked-ykqyq8?file=/src/App.js) and [1.4.3](https://codesandbox.io/s/input-clearable-error-1-5-forked-mqmfn5?file=/src/App.js) examples

### Regression?

1.4.3

### Priority impact

p1 - need for current milestone

### Impact

Error messages. In the doc site, prevents the component from loading altogether. cc @macandcheese 

### Calcite package

- [ ] @esri/calcite-components
- [X] @esri/calcite-components-react

### Esri team

Calcite (dev)

macandcheese:
Workaround is setting `""` instead of `undefined`...

driskull:
Installed a fix. There could be a refactor that handles more use cases so we may want to keep this issue open for another milestone.

geospatialem:
[Verified by Ben in the PR](https://github.com/Esri/calcite-design-system/pull/7476#issuecomment-1668653884)

geospatialem:
Reopening if we want to include more use cases mentioned by [Matt above](https://github.com/Esri/calcite-design-system/issues/7475#issuecomment-1668673075).

jcfranco:
[...]

PR #7476, opened 2023-08-07
fix(input): fix clearable from throwing an error when value is undefined

driskull:
**Related Issue:** #7475

## Summary

- Makes sure `value` is defined before trying to get its length.

### DIFF
--- a/packages/calcite-components/src/components/input/input.tsx
+++ b/packages/calcite-components/src/components/input/input.tsx
@@ -419,7 +419,7 @@ export class Input
   private childNumberEl?: HTMLInputElement;
 
   get isClearable(): boolean {
-    return !this.isTextarea && (this.clearable || this.type === "search") && this.value.length > 0;
+    return !this.isTextarea && (this.clearable || this.type === "search") && this.value?.length > 0;
   }
 
   get isTextarea(): boolean {


driskull:
@benelan or @eriklharper can one of you test this with react to verify it fixes it?

jcfranco (approved):

jcfranco:
@driskull I should've noticed this sooner, but according to https://github.com/Esri/calcite-design-system/pull/3627/files, the `value` watcher shouldn't allow `null` nor `undefined`, so I think the value isn't corrected at load time. If time permits, can you make this adjustment instead?

driskull:
> @driskull I should've noticed this sooner, but according to https://github.com/Esri/calcite-design-system/pull/3627/files, the value watcher shouldn't allow null nor undefined, so I think the value isn't corrected at load time. If time permits, can you make this adjustment instead?

I think its risky to be doing more refactoring to value on initialization. We can always patch again with further changes or hold off on this one.

Its a bit more complex because maybe on initialize it should just call the `valueHandler()`. Maybe @eriklharper would be best to make further changes. I still think this change is useful regardless.

benelan:
confirmed working in react

Issue #7032, opened 2023-05-25
Content added to empty scrim is not rendered


nwhittaker:
### Actual Behavior

Given an empty scrim that is already rendered, appending text content or child nodes results in no visual change to the scrim.

<img width="1496" alt="Screenshot 2023-05-25 at 3 43 55 PM" src="https://github.com/Esri/calcite-components/assets/421496/64159ac7-5251-406b-aaf3-7e5c6b357aa2">

### Expected Behavior

The appended content becomes visible within the scrim.

<img width="1496" alt="Screenshot 2023-05-25 at 3 48 10 PM" src="https://github.com/Esri/calcite-components/assets/421496/9bb48011-c00a-4b52-9229-99e33b112ec8">

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/yLRwjZw

### Reproduction Steps

1. Visit the sample and use dev tools to inspect the scrim
2. Within a second of rendering the scrim, see the DOM is updated with the scrim's new text content, however the text is not rendered in the preview area.

### Reproduction Version

1.4.0

### Relevant Info

Interestingly, if the scrim is not initially empty, updates to its content are rendered. A workaround is to seed the empty scrim with a comment.

```hbs
<calcite-scrim><!-- --></calcite-scrim>
```

### Priority impact

p4 - not time sensitive

### Esri team

ArcGIS Field Apps

geospatialem:
This looks as-expected in `1.6.0-next.3`, but differs from the screenshots above. cc @nwhittaker In case additional verification is needed with the use case above, as it differs from the [repro sample](https://codepen.io/nwhittaker-esri/pen/yLRwjZw).

![image](https://github.com/Esri/calcite-design-system/assets/5023024/d6c61fe8-4ecf-4400-9292-9253a0ff1d7d)

geospatialem:
Thanks for following up, @nwhittaker!

It does look like this seems to be a regression between `1.2.0` and `1.3.0`. The only change I can see between the two releases is the introduction of design tokens and utilities in the [comparison](https://github.com/Esri/calcite-design-system/compare/v1.2.0...v1.3.0).

Will follow up with research to determine next steps.

geospatialem:
The fix will be added as part of the September release.

There is a workaround when adding a child `div` element to the `scrim`: https://codepen.io/geospatialem/pen/wvQLwLB.

geospatialem:
Verified in `1.6.0-next.5`

PR #7477, opened 2023-08-07
fix(scrim): handle slotted children correctly

driskull:
**Related Issue:** #7032

## Summary

- Support adding slotted content after initial render
- Uses `onSlotChange` to detect slotted elements
- Hides content container via the hidden attribute
- Updates test


jcfranco (approved):
🎉

Issue #7397, opened 2023-07-31
`docs-vscode` output target not included in dist/


am-maneaters:
### Summary

Trying to use the generated VSCode docs to get better html intellisense but the `vscode-data.json` is not being included in `dist/extras/vscode-data.json`. when i change the `stencil.config.ts` to output the vscode docs to `./vscode-data.json` it works but it doesnt work with the dist path

### Desired Outcome

Fix the vscode docs output target

jcfranco:
@benelan Can you take a look at this?

benelan:
Hi @am-maneaters, it looks like vscode data is included in the `next` release, can you confirm on your end?
https://cdn.jsdelivr.net/npm/@esri/calcite-components@1.5.1-next.3/dist/extras/vscode-data.json

There was an issue with our release CI for v1.5.0 but it is resolved now. Sorry for the inconvenience!

am-maneaters:
@benelan no worries, thank you!

benelan:
Reopening because the vscode-data.json file wasn't included in the 1.5.1 release for some reason.

benelan:
Verification note: the fix can be confirmed locally by running the following script from the root of the repo

```sh
#!/usr/bin/env sh

fix() { rm -rf packages/calcite-components/dist/extras/vscode-data.json; }
expect() { echo "build $1 expected: vscode data $2"; }
assert() {
    npm run build --workspace=@esri/calcite-components >/dev/null
    [ -f packages/calcite-components/dist/extras/vscode-data.json ] &&
        echo "build $1 actual: vscode data exists" ||
        echo "build $1 actual: vscode data missing"
}

echo "this may take a while..."
fix
assert "pre-fix 1"
expect "pre-fix 1" "exists"
assert "pre-fix 2"
expect "pre-fix 2" "missing"
fix
assert "post-fix 1"
expect "post-fix 1" "exists"
fix
assert "post-fix 2"
expect "post-fix 2" "exists"

```

Unfortunately there is no way to verify for realsies until the next `latest` release happens.

geospatialem:
Verified with `1.6.0`

PR #7480, opened 2023-08-08
ci: ensure vscode-data is generated during every build

benelan:
**Related Issue:** #7397 

## Summary

The vscode-data Stencil output is not being generated if one already exists. It works on `next` because the components get built 3 times, once for tests, once for storybook deployment, and then a third time for the stencil types workaround. I'll try to find a way to remove one of those builds for `next` deployments.


driskull (approved):
👏

jcfranco (approved):
💥

PR #7487, opened 2023-08-08
fix(chip-group): Add existence checks

macandcheese:
## Summary
Running into some error using Chip Group / Chip in React environment with programmatically populated Chips...

Screenshot 2023-08-08 at 10 45 44 AM
Screenshot 2023-08-08 at 10 33 17 AM

Adds check for existence of unfound element.


driskull (approved):
👍

Issue #7489, opened 2023-08-08
[Panel] No longer has border under header


macandcheese:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Panel no longer has a border under a header:
<img width="1049" alt="Screenshot 2023-08-08 at 1 25 32 PM" src="https://github.com/Esri/calcite-design-system/assets/4733155/ae530966-1699-49c1-8e8f-3473917ecebf">

### Expected Behavior

I'd expect / want Panel to still have a border under header:
<img width="1107" alt="Screenshot 2023-08-08 at 1 25 27 PM" src="https://github.com/Esri/calcite-design-system/assets/4733155/e7d3deca-d734-4685-bb9c-f15a17ff9a7a">

### Reproduction Sample

Below

### Reproduction Steps

Compare the two Codepen:
1.4.3 - border: https://codepen.io/mac_and_cheese/pen/qBQweyr?editors=1000
1.5.1 - no border: https://codepen.io/mac_and_cheese/pen/GRwLVGd?editors=1000

### Reproduction Version

1.5.1

### Regression?

1.4.3

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

Calcite (design)

geospatialem:
Verified in `1.6.0-next.3`

![image](https://github.com/Esri/calcite-design-system/assets/5023024/7b0a10b7-7c58-4e48-bd38-fa3166930eac)

PR #7491, opened 2023-08-08
fix(panel): Fix heading border when only text content is slotted

driskull:
**Related Issue:** #7489

## Summary

- Restores border-block-end by default.
- Adds internal CSS variable for temporary usage
- Issue to add panel collapsing support:  https://github.com/Esri/calcite-design-system/issues/7498
- Adds story tests


packages/calcite-components/src/utils/dom.ts
```diff
@@ -426,6 +426,68 @@ export function toAriaBoolean(value: boolean): string {
...
+ * @param {Event} event The event.
+ * @returns {string} The slots text.
+ */
+export function slotChangeGetTextContent(event: Event): string {
+  return slotChangeGetAssignedNodes(event)
+    .filter((node) => node.nodeType === Node.TEXT_NODE)
+    .map((node) => node.textContent)
+    .join("")
+    .trim();
+}
```
driskull, line 445: @anveshmekala I think this utility could be used for...

https://github.com/Esri/calcite-design-system/blob/bb07831eb518d1d617b765e1f14279480c6051fc/packages/calcite-components/src/components/text-area/text-area.tsx#L406-L415

What do you think?

driskull: I'm a little confused about the logic above though, it seems like the last textNode will be set as `this.value` and any previous nodes would be overwritten.

jcfranco (commented):
Looking good, but can we separate the utils that aren't related to the fix for now?

packages/calcite-components/src/utils/dom.ts
```diff
@@ -426,6 +426,68 @@ export function toAriaBoolean(value: boolean): string {
...
+ *
+ * ```
+ * <slot onSlotchange={(event) => this.mySlotHasNode = slotChangeHasAssignedNode(event)} />}
+ * ```
+ *
+ * @param {Event} event The event.
+ * @returns {boolean} Whether the slot has any assigned nodes.
+ */
+export function slotChangeHasAssignedNode(event: Event): boolean {
+  return !!slotChangeGetAssignedNodes(event).length;
```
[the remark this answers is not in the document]
driskull: Yeah but its a nice convenience rather than having to do `!!slotChangeGetAssignedNodes(event).length` a lot.

packages/calcite-components/src/components/panel/panel.tsx
```diff
@@ -267,7 +268,7 @@ export class Panel
   };
 
   handleDefaultSlotChange = (event: Event): void => {
-    this.hasDefaultContent = slotChangeHasAssignedElement(event);
+    this.hasDefaultContent = slotChangeHasAssignedElement(event) || slotChangeHasTextContent(event);
```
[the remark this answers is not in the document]
driskull: Yes but you can get an assigned text node just from a return carriage :\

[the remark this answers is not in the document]
driskull: So, no not really :)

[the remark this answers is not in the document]
driskull:
```html
<calcite-panel height-scale="s">
      <div slot="header-content">Header!</div>
      test
    </calcite-panel>
```

```js
console.log(slotChangeGetAssignedNodes(event).map((el) => el.nodeValue));
```

Equals

```console
(2) ['\n      ', '\n      test\n    ']
```

macandcheese:
This is a great update but I think a lot of the doc site use cases would still have the missing border - can we only hide the one of the duplicate borders border when footer content is populated or footer / header containers are intersecting?

driskull:
>  I think a lot of the doc site use cases would still have the missing border

Can you explain why?

macandcheese:
> > I think a lot of the doc site use cases would still have the missing border
> 
> Can you explain why?

Because they don't have text or any content in them. In an app use case, maybe a Panel is unpopulated before an element is added:

Previous behavior with totally empty Panel:
Screenshot 2023-08-08 at 4 43 41 PM

Current: 
Screenshot 2023-08-08 at 4 44 21 PM


I think the fix here is still valid but maybe there is another way to determine when to remove the border.

driskull:
> I think the fix here is still valid but maybe there is another way to determine when to remove the border, if footer / header overlap remove it?

We'll have to brainstorm because it's not just the footer that is the problem. 

When a popup is collapsed, we don't want the border line showing here as it would divide nothing and put a line through the pointer tail.

image

driskull:
Maybe we need to build collapsing into the panel/flow-item to handle this correctly? Then it would only be removed when `collapsed`?

driskull:
Another option would be the ability to turn off the border when necessary through a CSS variable.

packages/calcite-components/src/components/panel/panel.tsx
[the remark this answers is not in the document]
jcfranco: Ah, those pesky whitespace text nodes. This LGTM then. 

Would we ever want to treat element vs text presence differently when it comes to slotted content? Wondering if we could simplify a bit by introducing a `slotChangeHasVisibleContent` or similar in the future.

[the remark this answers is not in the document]
driskull: Yeah thats a good point. Another util function :)

I'm happy to make another function if you see it being useful.

packages/calcite-components/src/utils/dom.ts
driskull: @anveshmekala I think this breaks the logic:

```html
<calcite-text-area>
Hello
world
hi</calcite-text-area>
```

driskull: https://codepen.io/driskull/pen/oNQREpy?editors=1111

It seems to work but could you double check?

packages/calcite-components/src/components/panel/panel.stories.ts
```diff
@@ -313,3 +313,11 @@ export const noOverflowContentWithFab_TestOnly = (): string => html` <calcite-pa
   ><div>My Content</div>
   <calcite-fab slot="fab"></calcite-fab
 ></calcite-panel>`;
+
+export const withTextContentOnly = (): string =>
+  html`<calcite-panel height-scale="s" heading="My Panel">Slotted content!</calcite-panel>`;
+
+export const withNoHeaderBorderBlockEnd = (): string =>
```
jcfranco, line 320: Should this use the `_TestOnly` suffix since it's internal?

jcfranco (approved):
Awesome! 🚀

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -12,6 +12,7 @@
 
   --calcite-min-header-height: calc(var(--calcite-icon-size) * 3);
   --calcite-panel-footer-padding: theme("spacing.2");
+  --calcite-internal-panel-header-border-block-end: 1px solid var(--calcite-ui-border-3);
```
macandcheese, line 15: Can the `-internal` be at the end? I think that's the pattern elsewhere - maybe we can add that to conventions? https://github.com/Esri/calcite-design-system/blob/bb07831eb518d1d617b765e1f14279480c6051fc/packages/calcite-components/src/components/chip/chip.scss#L3

jcfranco: ☝️ this is actually inconsistent with some other internal CSS props. 😅

I think `internal` should be more prominent in the name. This might be easier to spot if it's part of the prefix.

jcfranco: > maybe we can add that to conventions?

Excellent idea. I don't think we have this documented at the moment.

packages/calcite-components/src/utils/dom.ts
anveshmekala: it works with multi line text including special characters.

packages/calcite-components/src/components/panel/panel.scss
macandcheese: > 

LOL - yeah it's about 50/50. I feel like it's easier to spot at the end with nothing following it but... as long as its consistent either way.

driskull: I feel like we should follow events as they are prefixed with `calciteInternal`.

macandcheese: 🚢  it

driskull:
@macandcheese @jcfranco good to merge?

jcfranco:
Good to murge! 🚀

Issue #6484, opened 2023-02-15
Block: status icon size


ashetland:
### Description

Status icons default to `m` at 24px. To be consistent with existing icon size patterns in Calcite, they should be 16px, `scale=s`.
![image](https://user-images.githubusercontent.com/108549080/219136176-f91f40de-5825-4e99-94fc-abcdb06ba228.png)

### Acceptance Criteria

Status icons default to `scale=s`.

### Relevant Info

Reproducible on v1.0.5.
[[https://codepen.io/ashetland/pen/wvEBdMW?editors=1000]](https://codepen.io/ashetland/pen/wvEBdMW?editors=1000)

### Which Component

Block

### Esri team

Calcite (design)

geospatialem:
Verified in `1.6.0-next.5`

PR #7503, opened 2023-08-10
fix(block): defaults the status icon to `scale=s`

anveshmekala:
**Related Issue:** #6484 

## Summary

Defaults the scale of status icon to `s`


driskull (approved):
👍

PR #7504, opened 2023-08-10
docs(modal): update usage file with correct prop

driskull:
**Related Issue:** None

## Summary

- update usage file with correct prop


geospatialem (approved):
Great catch! 🦅 👀

Issue #6387, opened 2023-01-31
[Combobox] Using `space` key while focusing on a Combobox Item advances selection


macandcheese:
### Actual Behavior

Using the `space` key while focus is on a Combobox Item ("combobox open") advances focus to the next item.

### Expected Behavior

I might expect `space` to behave like `enter` - cc @geospatialem for a11y opinion.

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/combobox/

### Reproduction Steps

1. View sample link
2. Open the combobox
3. Use the `space` key
4. See that the focused item "advances" to the next one without selection.

### Reproduction Version

1.0.3

### Regression?

Functioned this way as far back as beta.86, at least.

### Impact

Unsure, but struck me as unexpected.

### Esri team

Calcite (design)

geospatialem:
> I might expect `space` to behave like `enter`

Would concur, per the [W3C combobox](https://w3c.github.io/aria-practices/examples/combobox/combobox-select-only.html) where `spacebar` and `Enter` keys perform the same action on its items.

Also, of note also applies to the `tab` key, but that might be out of scope for this issue.

geospatialem:
Verified in `1.6.0-next.5`

PR #7505, opened 2023-08-10
fix(combobox): prevents navigation list with Space key

anveshmekala:
**Related Issue:** #6387 

## Summary

This prevents `Space` key navigating combobox list when opened.

### DIFF
--- a/packages/calcite-components/src/components/combobox/combobox.tsx
+++ b/packages/calcite-components/src/components/combobox/combobox.tsx
@@ -605,8 +605,10 @@ export class Combobox
       case " ":
         if (!this.textInput.value) {
           event.preventDefault();
-          this.open = true;
-          this.shiftActiveItemIndex(1);
+          if (!this.open) {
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -901,6 +901,7 @@ export class Combobox
       this.emitComboboxChange();
       if (this.textInput) {
         this.textInput.value = item.textLabel;
+        console.log("toggle selection", item.textLabel);
```
driskull, line 904: remove console log :)

driskull (approved):
👍

packages/calcite-components/src/components/combobox/combobox.tsx
anveshmekala: missed it here 65bd4651f0f459155a8c92bfe84d49971059b9d0

Issue #6569, opened 2023-03-06
`calcite-progress` has incomplete animation in RTL


ethanbdev:
### Actual Behavior

The animation does not use the entire progress bar

### Expected Behavior

Both RTL and LTR should have a similar animation, using the entire progress bar

### Reproduction Sample

https://codepen.io/eborgen/pen/LYJjmEp?editors=1000

### Reproduction Steps

Observed in the codepen, change the direction to `ltr` for comparison

Also can be seen on the doc site after changing direction:
https://developers.arcgis.com/calcite-design-system/components/progress/

### Reproduction Version

1.0.7

### Relevant Info

Windows 10, Chrome Version 110.0.5481.178 (Official Build) (64-bit)

### Priority impact

p2 - want for current milestone

### Impact

Its already in production, would be hoping for a fix by the R2 timeline.

Affects both RTL supported languages in the map viewer

### Esri team

ArcGIS Web Analysis

ethanbdev:
BYS issue: https://github.com/ArcGIS/I18N-BYS-Bugs/issues/14165
CR-BY-Beyondsoft_I18N,QV-A-Bug,QV-S-New,QV-C-I18N,QV-C-AR;HE,QV-B-Low

Web analysis issue: https://devtopia.esri.com/WebGIS/arcgis-web-analysis/issues/2611

PR #7511, opened 2023-08-11
fix(progressbar): completes animation for `dir='rtl'`

anveshmekala:
**Related Issue:** #6569


driskull (approved):
👍 

No way to screenshot test this right?

anveshmekala:
> 👍
> 
> No way to screenshot test this right?

yup, animations can be very complex to screenshot with consistent keyframes.

Issue #7104, opened 2023-06-02
Clicking the root option of a calcite tree doesn't select all child options


sonofflynn89:
### Actual Behavior

On the ArcGIS Hub Team, we leverage the tree component in  `selection-mode=ancestors` to generate search facets based on the `categories` that users define for their organization. As such we have to dynamically generate a different tree for each user's site.

While performing testing, we increased our nested category levels from 3 to 4 (i.e. `/Root/Child/Grandchild` to `/Root/Child/Grandchild/Great-Grandchild`) and we discovered that the tree component selection breaks when you select the root option:

https://github.com/Esri/calcite-components/assets/30124777/b7f154e3-3be8-4ece-b91e-e415b235a5bb

This does not happen if you only have 3 nested levels:

https://github.com/Esri/calcite-components/assets/30124777/ff74c25c-f029-4a98-bd8e-ea8bfe3224c9

This issue doesn't just affect the visual presentation of the component. The underlying `calciteTreeSelect` also indicates that the 2nd (i.e.,`Child`) level item is _not selected_, even when it should be. This causes all sorts of problems on our application side as the post-condition that we rely on (i.e., that _all_ child elements are selected) is no longer true.

### Expected Behavior

Clicking the root option of a tree with any number of nested levels should select _all_ child options.

### Reproduction Sample

https://codepen.io/sonofflynn89-the-reactor/pen/OJBKZMG

### Reproduction Steps

- Go to the codepen
- Click the root of the first tree
- Expand the tree all the way
- Note that the 2nd level option is not selected, but the 3rd and 4th level options are
- Click the root of the first tree again
- Note that 2nd level option IS selected, but no other option is.
- Click the root of the second tree
- Expand the tree all the way
- Note that all options are selected as expected

### Reproduction Version

1.4.2

### Regression?

1.0.0

### Priority impact

p3 - want for upcoming milestone

### Impact

While we don't have a way of knowing how many customers will actually need the 4 level nesting, any client that does could block themselves into an corner and will have some search ui rendered useless unless they know to double click the specific offending option.

### Esri team

ArcGIS Hub

ashetland:
This can also be seen in the documentation sample https://developers.arcgis.com/calcite-design-system/components/tree/

geospatialem:
Research needed to determine an estimate for inclusion in a future release.

PR #7518, opened 2023-08-14
fix(tree): selects all child items when selection-mode is set to ancestors

anveshmekala:
**Related Issue:** #7104 

## Summary

Selects all the child `calcite-tree-item` elements when the root element is selected with `selection-mode="ancestors"`


driskull (approved):
👍 ⭐

Issue #7486, opened 2023-08-08
Calcite React tests fail in vitest


mpayson:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Calcite React tests fail in Vitest with the following error in Vite 0.33
```
SyntaxError: Named export 'Build' not found. The requested module '@stencil/core/internal/client/index.js' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@stencil/core/internal/client/index.js';
const { Build } = pkg;
```

In the latest Vitest (0.34.1), this can be worked around by in-lining Stencil so that the tests can run, but running Stencil in this environment leads to additional errors

```
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Unhandled Rejection ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
TypeError: Invalid URL
 ❯ new NodeError node:internal/errors:405:5
 ❯ new URL node:internal/url:611:13
 ❯ Module.getAssetPath node_modules/@stencil/core/internal/client/index.js:39:22
     37| const createTime = (fnName, tagName = '') => {
     38|     if (BUILD.profile && performance.mark) {
     39|         const key = `st:${fnName}:${tagName}:${i++}`;
[...]

PR #7521, opened 2023-08-15
fix: prevent vitetest/ssr errors due to defining components on the server

benelan:
**Related Issue:** #7486

## Summary

Enabling Stencil's [`includeImportCustomElements`](https://stenciljs.com/docs/react#includeimportcustomelements) option prevented SSR and vitetest users from being able to explicitly define the custom elements on the client. This patches the Stencil build to ensure the custom elements aren't defined while on the server.

### DIFF
--- a/packages/calcite-components-react/src/auto-define.ts
+++ b/packages/calcite-components-react/src/auto-define.ts
@@ -4,7 +4,7 @@ const isBrowser = (): boolean =>
   window.location === location &&
   window.document === document;
 
-export function autoDefine(component: string) {
+export function autoDefine(component: string): () => Promise<void> | undefined {
   if (isBrowser()) {
     return async () => (await import(`@esri/calcite-components/dist/components/${component}.js`)).defineCustomElement();
   }
--- a/packages/calcite-components-react/src/auto-define.ts
+++ b/packages/calcite-components-react/src/auto-define.ts
@@ -1,0 +1,12 @@
+const isBrowser = (): boolean =>
+  ![typeof window, typeof document, typeof location].includes("undefined") &&
+  [typeof process, typeof global].includes("undefined") &&
+  window.location === location &&
+  window.document === document;
+
+export function autoDefine(component: string) {
+  if (isBrowser()) {
+    return async () => (await import(`@esri/calcite-components/dist/components/${component}.js`)).defineCustomElement();
+  }
+  return undefined;
+}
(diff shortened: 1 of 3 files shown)


packages/calcite-components-react/src/auto-define.ts
```diff
@@ -0,0 +1,12 @@
+const isBrowser = (): boolean =>
+  ![typeof window, typeof document, typeof location].includes("undefined") &&
+  [typeof process, typeof global].includes("undefined") &&
+  window.location === location &&
+  window.document === document;
+
+export function autoDefine(component: string) {
```
jcfranco, line 7: Can we type the return arguments?

packages/calcite-components-react/src/auto-define.ts
```diff
@@ -0,0 +1,12 @@
+const isBrowser = (): boolean =>
```
jcfranco, line 1: Suggestion (minor): The result of this can be computed earlier or cached. Multiple `autoDefine` calls run the same logic.

packages/calcite-components-react/support/patchSSR.ts
```diff
@@ -0,0 +1,32 @@
...
+  promises: { readFile, writeFile },
+} = require("fs");
+const { dirname, resolve } = require("path");
+
+(async () => {
+  try {
+    const filePath = resolve(`${__dirname}/../src/components.ts`);
+    const contents = await readFile(filePath, { encoding: "utf8" });
+
+    const defineCustomElementImports = /import { defineCustomElement as defineCalcite.*(\r\n|\r|\n)/gm;
```
jcfranco, line 16: Can you add a short explainer on what these lines are doing (e.g., prevent static imports of `dist` output target by making CE definition lazy to accommodate SSR environments)?

jcfranco (approved):
Minor comments, but this LGTM! 🎉

packages/calcite-components-react/src/auto-define.ts
benelan: Per our Teams conversation, we decided to leave this as is for now just to be very safe.

It is possible (but unlikely) that if we change isBrowser to a constant its could be cached on the server as `false` by an SSR framework trying to be too smart. If this were to happen and the constant didn't get reevaluated on the client, the components would never be defined.

This needs a bit more investigation.

benelan:
[Chromatic snapshots looked good on a previous build](https://www.chromatic.com/build?appId=6266d45509d7eb004aa254fb&number=2151) so I skipped them because they don't need to be ran.

Issue #6610, opened 2023-03-16
calcite-dropdown: focus lost after tabbing through


rpanichakit9541:
### Summary

When I tab to a `calcite-dropdown` then hit "Spacebar" to open it and tab through all the items. Once I pass the last item the dropdown closes and the focus is lost.  I would think that the focus would either focus back onto the button that opened the dropdown or the next focusable element after the dropdown, but neither is true.  Interestingly if I then hit tab again it then focuses on the next focusable element.

### Actual Behavior

Focus disappears after tabbing through all the option of the dropdown

### Expected Behavior

It should focus on the next focusable element after the dropdown.

### Reproduction Sample

https://codepen.io/rpanichakit/pen/RwYJKeL?editors=100

### Reproduction Steps

1. Tab to the "Dropdown button" and hit "Spacebar" to open it
2. Tab through all the options
3. Note that the dropdown auto closes, but the focus is not lost
4. If you tab an extra time, then the focus goes to the next focusable element

### Reproduction Version

1.0.8

### Esri team

ArcGIS Online

geospatialem:
Discussing option to no longer allow tabbing through the component - only arrow keys will be used. Closing only occurs with `esc`, and the focus will be retained back to the trigger element.

anveshmekala:
The focus is restored on to trigger element with calcite latest version `1.5.0` . This issues will be repurposed to avoid tabbing through the dropdown items when opened as per [W3C spec](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/#:~:text=When%20focus%20is%20on%20a%20menuitem%20in%20a%20menu%20or%20menubar%2C%20move%20focus%20out%20of%20the%20menu%20or%20menubar%2C%20and%20close%20all%20menus%20and%20submenus.).

geospatialem:
Verified in `1.7.0-next.9`. 

Where once the component is focused the arrow keys, `Home` (first item) and `End` (last item), `Esc` (closed when opened) are achievable. The `tab` key will proceed to the next focusable element in the page (`shift` + `tab` to the previous focusable element in the page).

PR #7527, opened 2023-08-15
fix(dropdown): prevents navigating dropdown items with Tab key

anveshmekala:
**Related Issue:** #6610 

## Summary

- Prevents navigating the dropdown list using `Tab` key
- When `Tab` key is down with dropdown open the focus shifts on to next focusable element
- When `Shift+Tab` key is down with dropdown open the focus shifts back on to trigger element.


packages/calcite-components/src/components/dropdown/dropdown.tsx
```diff
@@ -387,10 +387,9 @@ export class Dropdown
 
     switch (keyboardEvent.key) {
       case "Tab":
-        if (this.items.indexOf(target) === this.items.length - 1 && !keyboardEvent.shiftKey) {
-          this.closeCalciteDropdown();
-        } else if (this.items.indexOf(target) === 0 && keyboardEvent.shiftKey) {
-          this.closeCalciteDropdown();
+        if (this.items.indexOf(target) > -1) {
```
jcfranco, line 390: The handled event would come a child item, so wouldn't this always be > -1?

packages/calcite-components/src/components/dropdown/dropdown.e2e.ts
```diff
@@ -724,98 +724,169 @@ describe("calcite-dropdown", () => {
     await page.waitForChanges();
     expect(await dropdownWrapper.isVisible()).toBe(false);
   });
+  describe("keyboard support", () => {
```
jcfranco, line 727: I think tests and describe blocks may need rearranging. This top one is for keyboard support and there's nested ones covering pointer interaction.

packages/calcite-components/src/components/dropdown/dropdown.tsx
```diff
@@ -387,10 +387,9 @@ export class Dropdown
...
+          this.open = false;
+          this.updateTabIndexOfItems(event);
```
jcfranco, line 392: Should we call this when the active item is updated vs only on Tab?

jcfranco (commented):

packages/calcite-components/src/components/dropdown/dropdown.e2e.ts
anveshmekala: removed the keyboard support describe block since most of the tests are using `click` to open the dropdown.

packages/calcite-components/src/components/dropdown/dropdown.tsx
```diff
@@ -675,4 +672,10 @@ export class Dropdown
       this.el.addEventListener("calciteDropdownOpen", this.toggleOpenEnd);
     }
   };
+
+  private updateTabIndexOfItems(event: CustomEvent<ItemKeyboardEvent>): void {
```
jcfranco, line 676: To make it more general and decouple the event from this, can you update the signature to pass the item directly instead of the event?

jcfranco (approved):
Noice! 👇⌨️👇

packages/calcite-components/src/components/dropdown/dropdown.tsx
jcfranco: Gotcha. I was wondering it the items could be left in an odd state.

packages/calcite-components/src/components/dropdown/dropdown.tsx
anveshmekala: makes sense. updated.

geospatialem (approved):
Awesome! 💪🏻 Great a11y update - works well with keyboard navigation and with JAWS active.

Issue #7534, opened 2023-08-16
[Action Bar] Deprecated "bottom-actions" slot doesn't work


macandcheese:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

We recently added "actions-end" slot to Action Bar, and deprecated "bottom-actions". But when using "bottom-actions" - the component does not render slot contents.

### Expected Behavior

I'd expect the deprecated slot to still render.

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/VwqZjbB?editors=1000

### Reproduction Steps

1. Go to Codepen
2. View "bottom-actions" slot in HTML
3. Change to "actions-end"
4. See slotted contents appear

### Reproduction Version

1.6.0

### Priority impact

p2 - want for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

Calcite (dev)

geospatialem:
Verified in `1.6.1-next.2` with the `"bottom-actions"` and `"actions-end"` slots.

PR #7535, opened 2023-08-16
fix(action-bar): restore "bottom-actions" slot functionality.

driskull:
**Related Issue:** #7534

## Summary

- Fix incorrect resources slot value
- Separate state from different slots so they aren't' competing with each other.
- Add tests


packages/calcite-components/src/components/action-bar/action-bar.stories.ts
```diff
@@ -82,6 +82,7 @@ export const horizontal = (): string => html`
         <calcite-action text="Layers" icon="layers"> </calcite-action>
       </calcite-action-group>
       <calcite-action slot="actions-end" text="hello world" icon="layers"> </calcite-action>
+      <calcite-action slot="bottom-actions" text="hello world 2" icon="information"> </calcite-action>
```
jcfranco, line 85: Can you add a comment somewhere around this to indicate that the slot is deprecated? Applies to the `horizontalSmall` story too.

jcfranco (approved):
✨🔧✨

PR #7541, opened 2023-08-16
chore: remove designated Element comment block and move  to Private Props

Elijbet:
**Related Issue:** #N/A

## Summary
Remove the designated `Element` comment block and move `@Element()` to the `Private Props` block.

We don't have an automated way of enforcing a designated slot for just the `Element` at the moment. `@Element()` is not really that different from any other private prop in a component, no different from `@State()` for example, and it's one less thing to maintain or keep track of.


packages/calcite-components/src/components/option/option.tsx
```diff
@@ -59,8 +59,7 @@ export class Option {
   //
   //--------------------------------------------------------------------------
 
-  @Element()
-  private el: HTMLCalciteOptionElement;
+  @Element() private el: HTMLCalciteOptionElement;
```
jcfranco, line 62: This formatting-only change is unrelated to a move/dropping of Element section.

Could you remove those from this PR? We can do a separate PR with linting fixes.

Applies to other files with only lint fixes.

packages/calcite-components/src/components/rating/rating.tsx
```diff
@@ -337,6 +329,8 @@ export class Rating
   //
   //--------------------------------------------------------------------------
 
+  @Element() el: HTMLCalciteRatingElement;
```
jcfranco, line 332: This is in the wrong section.

packages/calcite-components/src/components/accordion-item/accordion-item.tsx
```diff
@@ -289,6 +281,8 @@ export class AccordionItem implements ConditionalSlotComponent {
   //
   //--------------------------------------------------------------------------
 
+  @Element() el: HTMLCalciteAccordionItemElement;
+
   /** handle clicks on item header */
   private itemHeaderClickHandler = (): void => this.emitRequestedItem();
```
jcfranco, line 287: Can you move this to the section below?

packages/calcite-components/src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -310,6 +302,7 @@ export class InputTimePicker
   //  Private Properties
   //
   //--------------------------------------------------------------------------
+  @Element() el: HTMLCalciteInputTimePickerElement;
```
jcfranco, line 305: Can you restore the newline between the section and `el`?

packages/calcite-components/src/components/split-button/split-button.tsx
```diff
@@ -37,6 +37,12 @@ import { CSS } from "./resources";
...
 export class SplitButton implements InteractiveComponent, LoadableComponent {
+  //--------------------------------------------------------------------------
+  //
+  //  Private Properties
+  //
+  //--------------------------------------------------------------------------
+
   @Element() el: HTMLCalciteSplitButtonElement;
 
   /** Specifies the appearance style of the component. */
```
jcfranco, line 48: Can you add a `Properties` section here? 🙇

jcfranco (approved):
Awesome! 🚀🪐

Had some minor comments, but this should be good to merge once they are addressed.

jcfranco:
Forgot to confirm that this is low-risk. ✅

Issue #7180, opened 2023-06-15
Design token phase II - Integration across components


geospatialem:
### Description

Addresses the final, or second phase, of our design token integration from #6558.

Majority of the effort will be performed in June and July, targeted for early August for additional documentation and pages for the September main release.

cc @alisonailea @jcfranco 

### Acceptance Criteria

Work though component specific tokens and css/static references to:

- Identify existing variables
- Discuss with designers (@SkyeSeitz, @ashetland, @jacqueskeet) on existing variables to either:
    - Replace with a token **or**
    - Add to component tokens
- Changes will be incorporated into one branch until finalized 

### Relevant Info

For specific component issues, we can use the `design-tokens` label.

### Which Component

- [ ] `accordion` 
- [ ] `accordion-item` 
- [ ] `action`
- [ ] `action-bar`
- [ ] `action-group`
- [ ] `action-menu` 
- [ ] `action-pad`
- [ ] `alert`
- [ ] `avatar`
- [ ] `block`
- [ ] `block-section`
- [ ] `button`
- [ ] `card`
- [ ] `checkbox`
- [ ] `chip`
- [ ] `chip-group`
- [ ] `color-picker`
- [ ] `color-picker-hex-input`
- [ ] `color-picker-swatch`
- [ ] `combobox`
- [ ] `combobox-item`
- [ ] `combobox-item-group`
- [ ] `date-picker`
- [ ] `date-picker-day`
- [ ] `date-picker-month-header`
- [ ] `date-picker-month`
- [ ] `dropdown`
- [ ] `dropdown-group`
- [ ] `dropdown-item`
- [ ] `fab`
- [ ] `filter`
- [ ] `flow`
- [ ] `flow-item`
- [ ] functional
    - [ ] ExpandToggle
    - [ ] FloatingArrow
    - [ ] Heading
    - [ ] XButton
- [ ] `graph` 
- [ ] `handle`
- [ ] `icon`
- [ ] `inline-editable`
- [ ] `input`
- [ ] `input-date-picker`
- [ ] `input-message`
- [ ] `input-number`
- [ ] `input-text`
- [ ] `input-time-picker`
- [ ] `label`
- [ ] `link`
- [ ] `list`
- [ ] `list-item`
- [ ] `list-item-group`
- [ ] `loader`
- [ ] `menu`
- [ ] `menu-item`
- [ ] `modal`
- [ ] `navigation`
- [ ] `navigation-logo`
- [ ] `navigation-user`
[...]

PR #7551, opened 2023-08-18
fix: fix open/close event emitting in Safari

jcfranco:
**Related Issue:** #7180

## Summary

This updates how `onToggleOpenCloseComponent` determines the duration for the `openTransitionProp` to be more robust to work consistently across browsers.

The previous approach relied on the computed `transition` property having all associated transition values laid out per prop, which didn't work in Safari since it won't include all values if shared.

This change should be covered by existing tests.

cc @Elijbet


benelan (approved):

anveshmekala (approved):
🥇

PR #7553, opened 2023-08-18
fix: ensure label only focuses the first labelable child

jcfranco:
**Related Issue:** #5070 

## Summary

Ensures label clicks focus on the first labelable child and ignores clicks on labelable children (as they will already receive focus).

#### `label.ts`

* no longer binds label click handler per labelable

#### `commonTests.ts`/`labelable` test helper

* now asserts that a clicked labelable child will own focus
* grouped tests for both wrapped and sibling use cases
* removed unnecessary `for` usage in wrapped test coverage

### `dom.ts`

* add `isBefore` util to determine DOM order of elements (useful for sorting)


driskull (approved):
Looks good. nice util functions too!

Issue #7540, opened 2023-08-16
Icons disappear while dragging in a nested calcite-list


nid13240:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

I am using a calcite list with a nested drag enabled list because I only want certain items to be draggable. While dragging an item in this nested list, the drag icons disappear but reappear when I release the item.

### Expected Behavior

I expect the icons to remain visible while dragging.

### Reproduction Sample

https://codepen.io/nid13240/pen/NWEZYdN

### Reproduction Steps

1. open the codepen above
2. try to drag items in the first list
3. observe that the icons disappear while dragging

I've also added a regular non-nested drag enabled list for comparison

### Reproduction Version

1.6.0

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Map Viewer

geospatialem:
Verified in `1.7.0-next.9`.

PR #7555, opened 2023-08-21
fix(list): fixes dragging nested list items

driskull:
**Related Issue:** #7540

## Summary

- Fixes parent and nested lists from competing with each other and setting dragEnabled on children
- Only closest parent will now set the dragEnabled property on its direct children
- The furthest parent will control setting `selectionMode` and `selectionAppearance` on all child items.


jcfranco (approved):
Noice! Would it be possible to test this with our drag-and-drop test util?

driskull:
> Noice! Would it be possible to test this with our drag-and-drop test util?

Can we have a followup issue to handle this?

jcfranco:
> Can we have a followup issue to handle this?

👍

Issue #5565, opened 2022-10-25
Block, Block Section - Ensure custom controls provide proper textual name, role, and state information - (2036569799)


dqateam:
### Module: 
 08 Block &amp; Block Section 

### Violation: 
 Ensure custom controls provide proper textual name, role, and state information 

![image](https://user-images.githubusercontent.com/5023024/202800663-da4d0812-5631-4c10-84d4-2a4f95a294ff.png)

### WCAG Reference: 
- [1.1.1 Non-text Content](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html)
- [4.1.2 Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html) 

### Instance ID: 
 2036569799 

### Severity: 
 7 

### Description: 
 [Issue]
There are accordions without appropriate state information. Examples include:

- River

[User Impact]
Screen reader users will be unable to determine that these controls expand and collapse content. They will also have difficulty determining whether the controls are currently expanded or collapsed.

[Code Reference] 
 ```
<button aria-controls="8baa2714-5762-3308-30b8-ee324ecd6671-region" class="section-header toggle" id="8baa2714-5762-3308-30b8-ee324ecd6671-button" name="Expand" aria-label="Expand">
<calcite-icon aria-hidden="true" icon="chevron-right" scale="s" calcite-hydrated=""></calcite-icon><span class="section-header__text">River</span>
</button>
 ``` 

### Note: 
 [Suggestion]
Ensure custom controls provide proper textual name, role, and state information.

For accordions, the following information is expected:
- Each accordion header must be enclosed in a ``` <button> ``` element.
- Each of these buttons must be enclosed in a heading element of an appropriate level. The ``` <button> ``` element must be the only child of this heading element.
- When the panel is visible, the header button must have aria-expanded="true". When it is hidden, the header button must have aria-expanded="false".
- The button must have aria-controls set to the ID of the panel container.
- If a panel is visible and locked in the expanded state, its header button must have aria-disabled="true".

 

### Media Type: 
 Custom Controls 

### Additional Resources: 
- Report Source: Product accessibility evaluation conducted on the Calcite Design System in August, 2022.

geospatialem:
Could be related to a similar solution needed for #5553.

geospatialem:
Also came up in Map Viewer's accessibility audit.

geospatialem:
After initial R&D a few small modifications are needed, which include:
[...]

PR #7557, opened 2023-08-22
fix(block, block-section): improve a11y

jcfranco:
**Related Issue:** #5565 

## Summary

Updated HTML to improve a11y.

**Note**: this removes references to outdated `TEXT` constants, which were used before translations were built-in.

### DIFF
--- a/packages/calcite-components/src/components/block-section/block-section.tsx
+++ b/packages/calcite-components/src/components/block-section/block-section.tsx
@@ -25,14 +25,13 @@ import {
 import { Status } from "../interfaces";
 import { BlockSectionMessages } from "./assets/block-section/t9n";
 import { BlockSectionToggleDisplay } from "./interfaces";
-import { CSS, ICONS } from "./resources";
+import { CSS, ICONS, IDS } from "./resources";
 import {
   componentFocusable,
   LoadableComponent,
   setComponentLoaded,
   setUpLoadableComponent,
 } from "../../utils/loadable";
-import { guid } from "../../utils/guid";
 
 /**
  * @slot - A slot for adding custom content.
@@ -116,8 +115,6 @@ export class BlockSection implements LocalizedComponent, T9nComponent, LoadableC
   // --------------------------------------------------------------------------
 
   @Element() el: HTMLCalciteBlockSectionElement;
-
-  private guid = guid();
 
   @State() effectiveLocale: string;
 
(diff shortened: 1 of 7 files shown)


driskull (approved):
Nice improvements! 🦾

geospatialem (commented):
Know this isn't part of the PR, but the messages are repeating the collapse/expand twice. Would it be possible to remove the `toggleLabel`? The `aria-expanded` should provide context, but it looks like context is provided in the `title`, which might be the duplication presented in AT. 😢

jcfranco:
@geospatialem I made the following changes according to https://www.w3.org/WAI/ARIA/apg/patterns/accordion/examples/accordion/:

* removed `aria-label` since we have the same content as visible text
* removed `aria-expanded` on the content area since it's only needed in the toggle button (changed container back to `section`)
* ensured toggle button has `button` role and `aria-expanded`
* associated the content with the toggle button w/ `aria-labelledby`
* added `aria-hidden` to the switch since it is used as presentation (toggle button has all the relevant state)

Could you test again? 🙇‍♂️

packages/calcite-components/src/components/block-section/block-section.tsx
```diff
@@ -259,13 +259,7 @@ export class BlockSection implements LocalizedComponent, T9nComponent, LoadableC
...
       <Host>
         {headerNode}
-        <section
-          aria-expanded={toAriaBoolean(open)}
-          aria-labelledby={buttonId}
-          class={CSS.content}
-          hidden={!open}
-          id={regionId}
-        >
+        <section aria-labelledby={buttonId} class={CSS.content} hidden={!open} id={regionId}>
```
geospatialem, line 262: Not to the full context of the PR, but adding the following to the `section` provide additional context with JAWS when the section's content contains text and is expanded:
```ts
aria-live="polite"
role="region"
```

packages/calcite-components/src/components/block/block.tsx
```diff
@@ -303,7 +303,6 @@ export class Block
           <button
             aria-controls={regionId}
             aria-expanded={collapsible ? toAriaBoolean(open) : null}
-            aria-label={toggleLabel}
             class={CSS.toggle}
             id={buttonId}
             onClick={this.onHeaderClick}
```
geospatialem, line 308: It seems the `title` attribute is conflicting with AT, but if we add an `aria-describedby` and an unique `id` using `guid` to the `header`, similar to the `buttonId` and `regionId`, we could provide both contexts.

Something like:

```ts
const headerId = `${guid}-header`;

...

<header class={CSS.header} id={headerId}>

...

<button
...
aria-describedby={headerId}
...
>


```

geospatialem (commented):
A few small tweaks for a11y, and this should be good to go!

packages/calcite-components/src/components/block-section/block-section.tsx
geospatialem: Per our Teams convo, and Franco's sleuthing, [`section` inherits a region `role`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section#technical_summary), so the `role` is not needed in addition to `aria-live`.

jcfranco:
@geospatialem Belated update, I rolled back a few changes to where the AT interpretation was better and also restructured the `block-section` internals to avoid nested interactive elements. 

I think you beat me to the 💥🥊 with https://github.com/Esri/calcite-design-system/pull/7652, but we can make further adjustments if needed.

Issue #5553, opened 2022-10-25
Accordion and Accordion Item - Ensure custom controls provide proper textual name, role, and state information - (2036569668)


dqateam:
### Module: 
 03 Accordion and Accordion Item 

### Violation: 
 Ensure custom controls provide proper textual name, role, and state information 

![image](https://user-images.githubusercontent.com/5023024/202766183-52688513-9ed0-47d6-9af1-d7c4045e4224.png)

### WCAG Reference: 
- [1.1.1 Non-text Content](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html)
- [4.1.2 Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html) 

### Instance ID: 
 2036569668 

### Severity: 
 7 

### Description: 
 [Issue]
Accordions lack full implementation, missing some roles and/or states information. 

[User Impact]
Screen reader users will be unable to determine that these controls expand and collapse content. They will also have difficulty determining whether the controls are currently expanded or collapsed.

[Code Reference] 
 ```
<calcite-accordion appearance="default" icon-position="end" icon-type="chevron" scale="m" selection-mode="single" calcite-hydrated=""> ... </calcite-accordion>
 ``` 

### Note: 
 [Suggestion]
Ensure custom controls provide proper textual name, role, and state information.

For accordions, the following information is expected:
- Each accordion header must be enclosed in a ``` <button> ``` element.
- Each of these buttons must be enclosed in a heading element of an appropriate level. The ``` <button> ``` element must be the only child of this heading element.
- When the panel is visible, the header button must have aria-expanded="true". When it is hidden, the header button must have aria-expanded="false".
- The button must have aria-controls set to the ID of the panel container.
- If a panel is visible and locked in the expanded state, its header button must have aria-disabled="true".

In addition, it is recommended that container elements have a region role as it greatly increases accessibility for non-visual users.  This is not required though.  More information in the link below.

[Recommended Reading]
For more information about creating accessible accordions, please see the W3C Web Accessibility Initiative&#039;s ARIA Authoring Practices Guide: https://www.w3.org/WAI/ARIA/apg/patterns/accordion/ 

### Media Type: 
 Custom Controls 

### Additional Resources: 
- Report Source: Product accessibility evaluation conducted on the Calcite Design System in August, 2022.

brittneytewks:
Removed "needs triage" since a priority is set and estimate is indicated

[...]

PR #7560, opened 2023-08-22
fix(accordion, accordion-item): improve a11y

jcfranco:
**Related Issue:** #5553 

## Summary

Updated HTML to improve a11y.

### DIFF
--- a/packages/calcite-components/src/components/accordion-item/accordion-item.tsx
+++ b/packages/calcite-components/src/components/accordion-item/accordion-item.tsx
@@ -206,7 +206,7 @@ export class AccordionItem implements ConditionalSlotComponent {
             </div>
             {this.renderActionsEnd()}
           </div>
-          <section aria-labelledby={IDS.section} class={CSS.content} id={IDS.section}>
+          <section aria-labelledby={IDS.sectionToggle} class={CSS.content} id={IDS.section}>
             <slot />
           </section>
         </div>
--- a/packages/calcite-components/src/components/accordion-item/accordion-item.tsx
+++ b/packages/calcite-components/src/components/accordion-item/accordion-item.tsx
@@ -206,7 +206,7 @@ export class AccordionItem implements ConditionalSlotComponent {
             </div>
             {this.renderActionsEnd()}
           </div>
-          <section aria-labelledby={IDS.section} class={CSS.content}>
+          <section aria-labelledby={IDS.section} class={CSS.content} id={IDS.section}>
             <slot />
           </section>
         </div>
--- a/packages/calcite-components/src/components/accordion-item/accordion-item.tsx
+++ b/packages/calcite-components/src/components/accordion-item/accordion-item.tsx
@@ -174,9 +174,10 @@ export class AccordionItem implements ConditionalSlotComponent {
           <div class={{ [CSS.header]: true, [CSS_UTILITY.rtl]: dir === "rtl" }}>
             {this.renderActionsStart()}
             <div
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/accordion-item/resources.ts
```diff
@@ -18,3 +18,7 @@ export const CSS = {
   iconEnd: "icon--end",
   headerContainer: "header-container",
 };
+
+export const IDS = {
```
jcfranco, line 22: Sidebar: we can update our conventions on using ID if needed within shadow DOM since IDs are encapsulated.

packages/calcite-components/src/components/accordion-item/accordion-item.tsx
```diff
@@ -204,7 +205,7 @@ export class AccordionItem implements ConditionalSlotComponent {
             </div>
             {this.renderActionsEnd()}
           </div>
-          <div class={CSS.content}>
+          <div aria-live="polite" class={CSS.content} id={IDS.slotContainer} role="region">
```
driskull, line 208: I thought we were still trying to use unique ids for each component instance? Even though its in shadowDOM.

driskull (commented):
question ❔

driskull:
I think a basic structure test would be fine here. 👍

packages/calcite-components/src/components/accordion-item/accordion-item.tsx
jcfranco: Yeah, I think we were following this pattern since the early days because we didn't have shadow DOM enabled consistently. If AT picks this up smoothly, we can keep things simple for the only-need-unique-ID-in-the-component's-shadow-DOM case.

packages/calcite-components/src/components/accordion-item/resources.ts
jcfranco: Whoops, forgot to comment earlier. 😅

driskull:
> If AT picks this up smoothly, we can keep things simple for the only-need-unique-ID-in-the-component's-shadow-DOM case.

Sounds good. We'll just have to test AT.

packages/calcite-components/src/components/accordion-item/accordion-item.tsx
geospatialem: Hm, yeah, I'm not sure if this is an unique id issue and/or shadow DOM issue but JAWS isn't providing context with the current solution. Maybe we can try out an unique id?

jcfranco: @geospatialem I made some tweaks. Could you test again?

geospatialem (approved):
Good to go in both JAWS and NVDA. 🙌🏻

geospatialem:
> Good to go in both JAWS and NVDA. 🙌🏻

@jcfranco After looking at the [failing e2e](https://github.com/Esri/calcite-design-system/actions/runs/6035205853/job/16387124926?pr=7560#step:6:480), wondering if this could be fixed with the guid utility? It looks like the current value of  `IDS.section` is `"section"`. While it does work with AT, it would fail the test since its not an unique id?

geospatialem (approved):
🙌🏻 LGTM

jcfranco:
@geospatialem Thanks for testing!

@driskull It looks like we'll be able to use IDs unique to each shadow root after all. 🎉

driskull:
Nice! 🎉

PR #7562, opened 2023-08-22
refactor(combobox, combobox-item, combobox-group): `getElementProp` is refactored out across child components as an outdated pattern in favor of inheritable props set directly on parent

Elijbet:
**Related Issue:** #6038

## Summary

Instead of the `combobox-item` and `combobox-group` looking up the parent for scale, these get set by the combobox parent.

The logic for setting these props thus moves to the parent, getting rid of the `getElementProp altogether`. The parent component gets a mutation observer to do this as well as watchers for when it needs to modify the children.

Inherited props addressed:
- [x] `scale`
- [x] `selectionMode`


driskull (approved):
Awesome. 👏 LGTM!

benelan (approved):
LGTM!

Issue #7205, opened 2023-06-20
Color picker value does not update when alphaChannel is changed from false to true


driskull:
### Actual Behavior

If you start with a basic color picker and then set the alphaValue to true after initialization, the value does not update with an alpha value.

Even if you type a new hex color in the textbox or change the alpha text box number from 100 to 50, the value of the component does not update.

### Expected Behavior

The alphaChannel will update the value of the component.

I would expect any changes to alphaChannel as well as typing into any of the textboxes to update the value of the component.

### Reproduction Sample

https://codepen.io/driskull/pen/gOQwYPM

### Reproduction Steps

1. Open codepen
2. See default value
3. Type in a new alpha value
4. Verify that value has not changed. (it should);

### Reproduction Version

Latest

### Relevant Info

N/A

### Regression?

No

### Priority impact

p4 - not time sensitive

### Esri team

N/A

geospatialem:
Verified in `1.7.0-next.9`

PR #7563, opened 2023-08-22
fix(color-picker): update value when alphaChannel is toggled

jcfranco:
**Related Issue:** #7205 

## Summary

This updates the color picker to update its state properly when `alphaChannel` is toggled.


driskull (approved):
👍 🙇

Issue #6383, opened 2023-01-30
[Loader] Animation buggy on iOS


macandcheese:
### Actual Behavior

Animation of Loader is buggy on iOS (Safari and Chrome): 

https://user-images.githubusercontent.com/4733155/215604731-64df3324-1c5b-47c9-b7a2-782e70d00047.MP4

### Expected Behavior

I'd expect the loader to animate smoothly.

### Reproduction Sample
https://developers.arcgis.com/calcite-design-system/components/loader/

### Reproduction Steps

A simple loader in isolation in code pen or in the developer site sample displays the issue.

### Reproduction Version

1.0.3

### Relevant Info

I noticed this on my iOS devices. I don't have an Android device to test but would be good to confirm. Can use XCode Simulator to recreate locally.

### Esri team

Calcite (design)

jcfranco:
Worth noting that this is also reproducible on OS X (Ventura 13.5, Safari 16.6).

geospatialem:
Verified with iOS 16.6 in Chrome.

PR #7564, opened 2023-08-22
fix(loader): tweak loading animations to work in Safari

jcfranco:
**Related Issue:** #6383 

## Summary

This revisits how the loader animates its different segments to play nicely across all browsers. It seems like the percentages calculated [here](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/loader/loader.scss#L179-L181) are causing Safari to interrupt the animation before resetting (there are some [existing](https://bugs.webkit.org/show_bug.cgi?id=258897) [issues](https://bugs.webkit.org/show_bug.cgi?id=249307) on this behavior). It comes close to the original, but it's not identical, so I'm open to suggestions on how to improve this. cc @driskull @macandcheese @paulcpederson 

Also, the `loader-clockwise` (rotating) animation is now applied to the SVG container to avoid it being overwritten by the `loader-offset-#` animations.

This also fixes the determinate loader logic that mimics a loader updating dynamically (demo page).

### DIFF
--- a/packages/calcite-components/src/components/loader/loader.scss
+++ b/packages/calcite-components/src/components/loader/loader.scss
@@ -84,7 +84,10 @@ $loader-circumference: ($loader-scale - (2 * $stroke-width)) * 3.14159;
   block-size: var(--calcite-loader-size);
   inset-inline-start: 50%;
   margin-inline-start: calc(var(--calcite-loader-size) / 2 * -1);
-  transform: scale(1, 1);
+  animation-iteration-count: infinite;
+  animation-timing-function: linear;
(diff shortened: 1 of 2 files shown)


jcfranco:
You can browse the updated stories here: https://www.chromatic.com/build?appId=6266d45509d7eb004aa254fb&number=2207

macandcheese (approved):
Looks good in Simulator with a Codepen referencing Chromatic build

https://github.com/Esri/calcite-design-system/assets/4733155/e65947be-e993-42f0-a13a-224bce6c5990

jcfranco:
Merging in favor of having loader work consistently across browsers. We can make further adjustments if needed.

Issue #5736, opened 2022-11-10
Segmented Control becomes unfocusable after adding new children


cont-kolomeets:
### Actual Behavior

We create `calcite-segmented-control` and then add items to it programmatically. As a result, the segmented control becomes unfocusable with Tab. Looks like `tabindex` property is not assigned to the children internally.

Compare below.
When creating using static html, the `tabindex` is present.

![image](https://user-images.githubusercontent.com/5318527/201098812-500ae801-57f2-4aa0-9b11-20dba95136d1.png)

When adding items programmatically, no `tabindex` is present.

![image](https://user-images.githubusercontent.com/5318527/201099111-df0b43a7-1e5e-41e5-9e73-42c4725dd928.png)

### Expected Behavior

The segmented control must be focusable after re-populating.

### Reproduction Sample

https://codepen.io/cont-kolomeets/pen/JjZWEdN?editors=1000

### Reproduction Steps

See the sample above.

### Reproduction Version

beta.95

### Esri team

ArcGIS Business/Community Analyst

driskull:
Seems like `radio-group` needs to setup a mutation observer 👍

jcfranco:
Adding to the suggestion ☝️, most likely [this](https://github.com/Esri/calcite-components/blob/cfb14ce96528c0d68d8c1ef26d92af49c4532825/src/components/segmented-control/segmented-control.tsx#L135-L144) needs to be run by the observer.

geospatialem:
Verified in `1.7.0-next.9` with `calcite-segmented-control` and `calcite-segmented-control-item`.

PR #7567, opened 2023-08-22
fix(segmented-control): refresh items when added dynamically

jcfranco:
**Related Issue:** #5736 

## Summary

Sets up a mutation observer to update the item state when children are added/removed after initialization.

**Note**: left/right + up/down arrow navigation test were consolidated.


packages/calcite-components/src/components/segmented-control/segmented-control.tsx
```diff
@@ -332,4 +329,15 @@ export class SegmentedControl
       match.focus();
     }
   }
+
+  private setUpItems(): void {
+    const items = this.getItems();
+    const lastChecked = items.filter((item) => item.checked).pop();
```
driskull, line 335: could use array.find as well right?

driskull (approved):
👍

packages/calcite-components/src/components/segmented-control/segmented-control.tsx
jcfranco: Good idea, but maybe once Safari 17 is out. I need to find the last checked item and Safari added support for `findLast` in v15.4. 🧭🥲

PR #7573, opened 2023-08-22
fix(modal): Add close tooltip and aria-label on scrim when it can be clicked to close

driskull:
**Related Issue:** #

## Summary

- Add close tooltip and aria-label on scrim when it can be clicked to close

### DIFF
--- a/packages/calcite-components/src/components/modal/modal.tsx
+++ b/packages/calcite-components/src/components/modal/modal.tsx
@@ -198,6 +198,7 @@ export class Modal
   }
 
   render(): VNode {
+    const { outsideCloseDisabled, messages } = this;
     return (
       <Host
         aria-describedby={this.contentId}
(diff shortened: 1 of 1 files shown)


packages/calcite-components/src/components/modal/modal.tsx
```diff
@@ -212,7 +213,12 @@ export class Modal
             [CSS.slottedInShell]: this.slottedInShell,
           }}
         >
-          <calcite-scrim class={CSS.scrim} onClick={this.handleOutsideClose} />
+          <calcite-scrim
```
driskull, line 216: cc @geospatialem 

Should we have an aria-label and title attribute on the scrim if clicking it will close the modal? Should we also have a role of button or something? Its not keyboard accessible so maybe the aria-label isn't needed?

I was just thinking the tooltip is nice to have though.

geospatialem: Did this surface from an existing issue? Overall don't think a `label` is needed for AT context, since as you mentioned the `scrim` wouldn't be accessible to keyboard or AT users, so wouldn't provide context. With `modal` having a dialog role, should cover a11y. 

The `tooltip` seems odd visually, but would defer to design on the change. From an a11y perspective, wouldn't add/remove value to users, if that helps for context.

image

driskull: Ok ill close

Issue #7382, opened 2023-07-26
[Date Picker] Calendar should start the week with Monday in `zh-CN`


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

In the `zh-CN` locale, the `date picker` starts the week on Sunday (`周日`).

### Expected Behavior

The first day of the week should be set to Monday (`周一`), where the [`weekStart`](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/date-picker/assets/date-picker/nls/zh-CN.json#L5) is set to `1` per @annierm18.

### Reproduction Sample

https://codepen.io/geospatialem/pen/qBQJPOd

### Reproduction Steps

1. Open the [sample](https://codepen.io/geospatialem/pen/qBQJPOd)
2. Observe the day of the week is set to Sunday (`周日`) and not Monday (`周一`)

### Reproduction Version

1.4.3

### Priority impact

p1 - need for current milestone

### Impact

Moderate priority for the MapViewer team, found on 6/21 in `arcgis-webviewer-app` issue 7602.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Map Viewer

geospatialem:
Verified in `1.7.0-next.9`

PR #7578, opened 2023-08-22
fix(date-picker): set start of week to monday in zh-CN

geospatialem:
**Related Issue:** #7382

## Summary
Updates the `date-picker` first day of the week in the `zh-CN` locale to Monday (`周一`), previously set as Sunday (`周日`).


driskull (approved):
👍

PR #7561, opened 2023-08-22
feat(sheet): Add Sheet component

driskull:
**Related Issue:** #6391

## Summary

- Adds new `calcite-sheet` component.
- Adds e2e tests
- Adds screenshot tests
- Adds html page with examples
- Modeled after the Modal component.
- Added to index page
- Added to stencil config

## After merged

- [Shell] Add "Sheet" Slot
  - Will try to do for this release if it gets merged in time 
  - Issue: #7154 
  - PR: #7579

## Future

- Resizable (need issue) @macandcheese 

### Example

```html
<calcite-sheet open label="libero nunc" position="inline-start">
  <calcite-panel closable heading="Ultrices neque"
    ><p>
      Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore
      magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
      consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.
      Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
    </p>
    <calcite-button slot="footer" width="half" appearance="outline">tincidunt lobortis</calcite-button>
    <calcite-button slot="footer" width="half" appearance="outline">amet porttitor</calcite-button>
  </calcite-panel>
  <script>
    document.addEventListener("calcitePanelClose", () => {
      document.querySelector("calcite-sheet").open = false;
    });
  </script>
</calcite-sheet>
```

## Properties

| Property               | Attribute                | Description                                                                                                                 | Type                                                             | Default          |
| ---------------------- | ------------------------ | --------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------- | ---------------- |
| `beforeClose`          | --                       | Passes a function to run before the component closes.                                                                       | `(el: HTMLCalciteSheetElement) => Promise<void>`                 | `undefined`      |
| `displayMode`          | `display-mode`           | Specifies the display mode - `"float"` (content is separated detached), or `"overlay"` (displays on top of center content). | `"float" \| "overlay"`                                           | `"overlay"`      |
| `escapeDisabled`       | `escape-disabled`        | When `true`, disables the default close on escape behavior.                                                                 | `boolean`                                                        | `false`          |
[...]

Issue #7154, opened 2023-06-08
[Shell] Add "Sheet" Slot


macandcheese:
### Description

Similar to Alert, and Modal, we should offer a slot within Shell to accept a https://github.com/Esri/calcite-components/issues/6391.

### Acceptance Criteria

Depends on https://github.com/Esri/calcite-components/issues/6391

### Which Component

Sheet, Shell

### Priority impact

p4 - not time sensitive

### Esri team

Calcite (design)

geospatialem:
Verified in `1.7.0-next.13`.

PR #7579, opened 2023-08-22
feat(shell): Add "Sheets" Slot

driskull:
**Related Issue:** #7154

## Summary

- depends on #7561
- Adds `sheets` slot to the shell where users can place `calcite-sheet` components
- Adds screenshot test
- Adds example HTML

cc @macandcheese


packages/calcite-components/src/components/shell/shell.tsx
```diff
@@ -129,6 +141,14 @@ export class Shell implements ConditionalSlotComponent {
     );
   }
 
+  renderSheets(): VNode {
+    return (
+      <div hidden={!this.hasSheets}>
```
anveshmekala, line 146: Q: is hidden required here considering slotted elements are rendered only if user parses them.

packages/calcite-components/src/components/shell/shell.tsx
```diff
@@ -53,6 +54,8 @@ export class Shell implements ConditionalSlotComponent {
 
   @State() hasModals = false;
 
+  @State() hasSheets = false;
```
anveshmekala, line 57: nitpick: can this be `hasSheet` or `hasSlottedSheet` instead of `hasSheets`.

anveshmekala (approved):
LGTM!  👍

packages/calcite-components/src/components/shell/shell.tsx
driskull: This is because we don't want to display the div if no content is slotted here.

packages/calcite-components/src/components/shell/shell.tsx
driskull: I'd like to maintain consistency with the above `hasModals`. It can be multiple sheets.

driskull:
@jcfranco let me know if I need to make any changes post merge.

packages/calcite-components/src/components/shell/shell.tsx
anveshmekala: makes sense. wasn't aware that users will slot multiple sheets.

PR #7582, opened 2023-08-23
refactor(dropdown, dropdown-item-group, dropdown-item): `getElementProp` is refactored out in favor of inheritable props set directly on parent

Elijbet:
**Related Issue:** #6038

## Summary

The logic for setting these props thus moves to the parent, getting rid of the `getElementProp` altogether. The parent component gets a `mutationObserver` to do this as well as `watchers` for when it needs to modify the children.

Inherited props addressed:

- [x]  `scale` is inherited from `dropdown`,
- [x]  `selectionMode` is inherited from `dropdown-item-group`, so there can be a mix of types within one dropdown.

### DIFF
--- a/packages/calcite-components/src/components/dropdown-group/dropdown-group.tsx
+++ b/packages/calcite-components/src/components/dropdown-group/dropdown-group.tsx
@@ -13,6 +13,7 @@ import {
 import { Scale, SelectionMode } from "../interfaces";
 import { RequestedItem } from "./interfaces";
 import { createObserver } from "../../utils/observers";
+import { CSS } from "../dropdown-item/resources";
 
 /**
  * @slot - A slot for adding `calcite-dropdown-item`s.
@@ -94,8 +95,8 @@ export class DropdownGroup {
       <Host aria-label={this.groupTitle} role="group">
         <div
           class={{
-            ["container"]: true,
-            [`container--${this.scale}`]: true,
+            [CSS.container]: true,
+            [`${CSS.container}--${this.scale}`]: true,
           }}
         >
           {dropdownSeparator}
--- a/packages/calcite-components/src/components/dropdown-group/dropdown-group.tsx
+++ b/packages/calcite-components/src/components/dropdown-group/dropdown-group.tsx
@@ -9,10 +9,9 @@ import {
   Prop,
   VNode,
 } from "@stencil/core";
-import { getElementProp } from "../../utils/dom";
 import { Scale, SelectionMode } from "../interfaces";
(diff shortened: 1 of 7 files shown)


packages/calcite-components/src/components/dropdown/dropdown.tsx
```diff
@@ -486,6 +490,10 @@ export class Dropdown
     this.updateSelectedItems();
 
     this.reposition(true);
+
+    this.items.forEach((item) => {
```
driskull, line 494: nitpick: could be simplified to

```
this.items.forEach((item) => (item.scale = this.scale));
```

packages/calcite-components/src/components/dropdown-item/dropdown-item.tsx
```diff
@@ -183,11 +200,9 @@ export class DropdownItem implements LoadableComponent {
       <Host aria-checked={itemAria} aria-label={!href ? label : ""} role={itemRole} tabindex="0">
         <div
           class={{
-            container: true,
+            ["container"]: true,
```
driskull, line 203: can you add this one to the CSS const?

packages/calcite-components/src/components/dropdown-item/dropdown-item.tsx
```diff
@@ -183,11 +200,9 @@ export class DropdownItem implements LoadableComponent {
...
             [CSS.containerLink]: !!href,
-            [CSS.containerSmall]: scale === "s",
-            [CSS.containerMedium]: scale === "m",
-            [CSS.containerLarge]: scale === "l",
+            [`container--${scale}`]: true,
```
driskull, line 205: can you add this one to the CSS const?

driskull: Just the `container--` part?

packages/calcite-components/src/components/dropdown-group/dropdown-group.tsx
```diff
@@ -83,12 +94,9 @@ export class DropdownGroup {
       <Host aria-label={this.groupTitle} role="group">
         <div
           class={{
-            container: true,
-            [CSS.containerSmall]: scale === "s",
-            [CSS.containerMedium]: scale === "m",
-            [CSS.containerLarge]: scale === "l",
+            ["container"]: true,
+            [`container--${this.scale}`]: true,
```
driskull, line 98: can you add this one to the CSS const? Just the `container--` part

driskull (approved):
💯 🎈 👍

Issue #7585, opened 2023-08-23
[Chip Group] TypeError in certain cases when populating in React programmatically 


macandcheese:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Related https://github.com/Esri/calcite-design-system/issues/7490

### Expected Behavior

No response

### Reproduction Sample

No response

### Reproduction Steps

No response

### Reproduction Version

1.6.1

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [ ] @esri/calcite-components
- [X] @esri/calcite-components-react

### Esri team

Calcite (design)

geospatialem:
Verified in `main`.

PR #7586, opened 2023-08-23
fix(chip-group): Add existence checks

macandcheese:
**Related Issue:** #7585 

## Summary
Adds check for existence with optional chaining.


driskull (approved):
👍 

Could also be solved by seeing where `this.items` is not being set to an array.

PR #7592, opened 2023-08-24
refactor(segmented-control, segmented-control-item): `getElementProp` is refactored out in favor of inheritable props set directly on parent

Elijbet:
**Related Issue:** #6038

## Summary

Instead of the `segmented-control-item` looking up the parent for `appearance`, `layout`, and `scale`, these get set by the `segmented-control` parent.

The logic for setting these props thus moves to the parent, getting rid of the `getElementProp` altogether. The parent component gets a `mutationObserver` to do this and watchers for when it needs to modify the children.

Inherited props addressed:

- [x] appearance
- [x] layout
- [x] scale

### DIFF
--- a/packages/calcite-components/src/components/segmented-control-item/segmented-control-item.tsx
+++ b/packages/calcite-components/src/components/segmented-control-item/segmented-control-item.tsx
@@ -71,6 +71,12 @@ export class SegmentedControlItem {
    */
   @Prop() scale: Scale = "m";
 
+  //--------------------------------------------------------------------------
+  //
+  //  Lifecycle
+  //
+  //--------------------------------------------------------------------------
+
   render(): VNode {
     const { appearance, checked, layout, scale, value } = this;
 
@@ -98,9 +104,12 @@ export class SegmentedControlItem {
       <Host aria-checked={toAriaBoolean(checked)} aria-label={value} role="radio">
         <label
           class={{
-            [`label--scale-${scale}`]: true,
-            [`label--${layout}`]: true,
-            [`label--${appearance}`]: true,
+            "label--scale-s": scale === "s",
+            "label--scale-m": scale === "m",
+            "label--scale-l": scale === "l",
+            "label--horizontal": layout === "horizontal",
+            "label--outline": appearance === "outline",
+            "label--outline-fill": appearance === "outline-fill",
           }}
         >
           {this.iconStart ? iconStartEl : null}
--- a/packages/calcite-components/src/components/segmented-control-item/segmented-control-item.tsx
+++ b/packages/calcite-components/src/components/segmented-control-item/segmented-control-item.tsx
@@ -9,7 +9,7 @@ import {
   VNode,
   Watch,
 } from "@stencil/core";
-import { getElementProp, toAriaBoolean } from "../../utils/dom";
+import { toAriaBoolean } from "../../utils/dom";
 import { Appearance, Layout, Scale } from "../interfaces";
 import { CSS, SLOTS } from "./resources";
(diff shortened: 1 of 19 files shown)


driskull (approved):
👏 Looks good @Elijbet! Can you shorten the PR title a bit?

packages/calcite-components/src/components/segmented-control/segmented-control.tsx
```diff
@@ -76,19 +72,30 @@ export class SegmentedControl
...
+  /** Defines the layout of the component. */
+  @Prop({ reflect: true }) layout: Layout = "horizontal";
+
   /** Specifies the size of the component. */
   @Prop({ reflect: true }) scale: Scale = "m";
 
+  @Watch("appearance")
+  @Watch("layout")
+  @Watch("scale")
+  handlePropsChange(): void {
```
jcfranco, line 95: Nitpick: the current name can apply to any sort of prop being modified, so can you rename it to reflect the context or purpose? Maybe `handleItemPropChange`?

packages/calcite-components/src/components/segmented-control/segmented-control.e2e.ts
```diff
@@ -284,14 +336,18 @@ describe("calcite-segmented-control", () => {
...
-    expect(element).toEqualAttribute("scale", "m");
-    expect(element).toEqualAttribute("layout", "horizontal");
-    expect(element).toEqualAttribute("appearance", "solid");
-    expect(element).toEqualAttribute("width", "auto");
+    await page.setContent(html`
+      <calcite-segmented-control appearance="outline" layout="vertical" scale="l"></calcite-segmented-control>
+    `);
+    const segmentedControlItems = await page.findAll("calcite-segmented-control-item");
+
+    segmentedControlItems.forEach(async (item) => {
```
jcfranco, line 346: You'll need to use a `for...of` loop since `forEach` doesn't play well w/ `await`.

packages/calcite-components/src/components/segmented-control-item/segmented-control-item.tsx
```diff
@@ -84,12 +98,9 @@ export class SegmentedControlItem {
       <Host aria-checked={toAriaBoolean(checked)} aria-label={value} role="radio">
         <label
           class={{
-            "label--scale-s": scale === "s",
-            "label--scale-m": scale === "m",
-            "label--scale-l": scale === "l",
-            "label--horizontal": layout === "horizontal",
-            "label--outline": appearance === "outline",
-            "label--outline-fill": appearance === "outline-fill",
+            [`label--scale-${scale}`]: true,
```
jcfranco, line 101: Might be better to keep the previous class object since dynamically-generated classes won't work w/ our `CSS` lookup pattern (when this component is updated to use it).

packages/calcite-components/src/components/segmented-control-item/segmented-control-item.tsx
```diff
@@ -50,15 +50,29 @@ export class SegmentedControlItem {
...
+   */
+  @Prop() layout: Layout = "horizontal";
+
+  /**
+   * Specifies the size of the component inherited from the `calcite-segmented-control`, defaults to `m`.
+   *
+   * @internal
+   */
+  @Prop() scale: Scale = "m";
+
```
jcfranco, line 73: Could you add a `lifecycle` comment divider here? 🙇‍♂️

jcfranco (approved):
Awesome! 🤘🪨🤘

This should be good to merge once comments are addressed.

PR #7593, opened 2023-08-24
refactor(stepper, stepper-item): `getElementProp` is refactored out in favor of inheritable props set directly on parent

Elijbet:
**Related Issue:** #6038

## Summary

Instead of the `stepper-item` looking up the parent for `icon, layout, numbered, and scale`, these get set by the `stepper` parent.

The logic for setting these props thus moves to the parent, getting rid of the `getElementProp` altogether. The parent component gets a `mutationObserver` to do this as well as watchers for when it needs to modify the children.

Inherited props addressed:

- [x] icon
- [x] layout
- [x] numbered
- [x] scale


packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -40,6 +41,21 @@ export class Stepper {
   /** When `true`, displays the step number in the `calcite-stepper-item` heading. */
   @Prop({ reflect: true }) numbered = false;
 
+  /** Specifies the parent of the component. */
+  @Prop({ reflect: true }) stepperParent: HTMLCalciteStepperElement;
```
driskull, line 45: I dont think this prop is necessary.

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -258,6 +277,20 @@ export class Stepper {
   /** keep track of the currently active item position */
   private currentPosition: number;
 
+  private mutationObserver = createObserver("mutation", () => this.updateItems());
+
+  private updateItems(): void {
+    if (this.items) {
```
driskull, line 283: Shouldn't it be querying the items here instead of relying on the register event? Otherwise, watching this.items is ncessary to updateItems().

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -258,6 +277,20 @@ export class Stepper {
...
+      for (const item of this.items) {
+        item.icon = this.icon;
+        item.numbered = this.numbered;
+        item.layout = this.layout;
+        item.stepperParent = this.el;
```
driskull, line 288: nitpick: `stepperParentEl`

packages/calcite-components/src/components/tree-item/tree-item.tsx
```diff
@@ -399,7 +399,7 @@ export class TreeItem implements ConditionalSlotComponent, InteractiveComponent
           });
         }
         event.preventDefault();
-      }
+    }
```
jcfranco, line 402: Please submit this separately as it's unrelated to the PR's changeset.

jcfranco (approved):
This should be good to merge after addressing all comments.

PR #7605, opened 2023-08-25
refactor(tabs, tab, tab-nav, tab-title): `getElementProp` is refactored out in favor of inheritable props set directly on parent

Elijbet:
**Related Issue:** #6038

## Summary

`getElementProp` is refactored out across child components as an outdated pattern in favor of inheritable props set directly on the parent. Instead of the `tab, tab-title, tab-nav` looking up the parent for `position` and `scale`, these get set by the `tabs` parent.

The logic for setting these props thus moves to the parent, getting rid of the `getElementProp` altogether. The parent component gets a `mutationObserver` to achieve this and watchers for when it needs to modify the children. Child attribute selectors are subbed with new classes added dynamically to reflect the scale and position, as attributes are no longer reflected to be targeted. 

Inherited props addressed:

- [x] position
- [x] scale


packages/calcite-components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -55,19 +56,23 @@ export class TabNav {
...
   @Prop({ reflect: true, mutable: true }) layout: TabLayout = "inline";
 
   /**
-   * @internal
+   * Specifies the position of `calcite-tab-nav` and `calcite-tab-title` components in relation to, and is inherited from the parent `calcite-tabs`, defaults to `top`.
+   *
+   *  @internal
    */
-  @Prop({ reflect: true, mutable: true }) position: TabPosition = "bottom";
+  @Prop() position: TabPosition = "top";
```
driskull, line 75: How come the default was flipped here?

packages/calcite-components/src/components/tab-nav/tab-nav.scss
```diff
@@ -2,15 +2,15 @@
   @apply relative flex;
 }
 
-:host([scale="s"]) {
+:host .scale-s {
```
driskull, line 5: nitpick: `:host` isn't needed in these selectors

packages/calcite-components/src/components/tab-nav/tab-nav.scss
```diff
@@ -59,7 +59,7 @@
   inset-block-end: unset; // display active blue line above instead of below
 }
 
-:host([bordered][position="bottom"]) .tab-nav-active-indicator-container {
+:host([bordered]) .position-bottom .tab-nav-active-indicator-container {
```
driskull, line 62: This seems to be the only selector where host is needed.

packages/calcite-components/src/components/tab-title/tab-title.scss
```diff
@@ -87,20 +87,20 @@
   }
 }
 
-:host([scale="s"]) {
+:host .scale-s {
```
driskull, line 90: host isn't needed in this selector.

packages/calcite-components/src/components/tab/tab.scss
```diff
@@ -18,15 +18,15 @@ section,
   @apply hidden h-full w-full;
 }
 
-:host([scale="s"]) {
+:host .scale-s section {
```
driskull, line 21: host isn't needed in these selectors

driskull (commented):
Looks good @Elijbet ✨

packages/calcite-components/src/components/tab-nav/tab-nav.tsx
Elijbet: The usual. 
🎩 🪄 🐇

packages/calcite-components/src/components/tabs/resources.ts
```diff
@@ -1,3 +1,21 @@
+export const CSS = {
```
jcfranco, line 1: I don't think this is the proper place for these moved constants since tabs does not use any of these. Please move and split up as needed per component.

packages/calcite-components/src/components/tabs/tabs.e2e.ts
```diff
@@ -34,6 +34,14 @@ describe("calcite-tabs", () => {
     ]);
   });
 
+  describe("reflects", () => {
```
jcfranco, line 37: ✨🏆✨

packages/calcite-components/src/components/tabs/tabs.e2e.ts
```diff
@@ -16,7 +16,7 @@ describe("calcite-tabs", () => {
     <calcite-tab>Tab 3 Content</calcite-tab>
     <calcite-tab>Tab 4 Content</calcite-tab>
   `;
-  const tabsSnippet = `<calcite-tabs>${tabsContent}</calcite-tabs>`;
+  const tabsSnippet = html`<calcite-tabs>${tabsContent}</calcite-tabs>`;
```
jcfranco, line 19: ✨🏅✨

packages/calcite-components/src/components/tabs/tabs.e2e.ts
```diff
@@ -258,11 +240,25 @@ describe("calcite-tabs", () => {
         document.body.innerHTML = `<${wrapperName}></${wrapperName}>`;
 
         const wrapper = document.querySelector(wrapperName);
+
+        async function waitForAnimationFrames(count) {
```
jcfranco, line 244: We already have a [helper to wait an animation frame](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/tests/utils.ts#L255-L257). Can you use that here?

packages/calcite-components/src/components/tabs/tabs.e2e.ts
```diff
@@ -346,4 +342,19 @@ describe("calcite-tabs", () => {
 
     expect(selectedTitleOnEmit).toBe("boats");
   });
+
+  it("inheritable props `position` and `scale` get passed to `tab-nav` and `tab-titles`", async () => {
```
jcfranco, line 346: This test seems to be a partial duplicate of `should render itself and child tab elements with corresponding scale`.

packages/calcite-components/src/components/tabs/tabs.e2e.ts
```diff
@@ -346,4 +342,19 @@ describe("calcite-tabs", () => {
...
+    const page = await newE2EPage();
+    await page.setContent(html`<calcite-tabs position="bottom" scale="l">${tabsContent}</calcite-tabs>`);
+    const tabNav = await page.find("calcite-tab-nav");
+    const tabTitles = await page.findAll("calcite-tab-titles");
+
+    expect(await tabNav.getProperty("position")).toBe("bottom");
+    expect(await tabNav.getProperty("scale")).toBe("l");
+
+    tabTitles.forEach(async (item) => {
```
jcfranco, line 355: You'll need to use a `for` or `for...x` loop since `forEach` is not async.

packages/calcite-components/src/components/tabs/tabs.tsx
```diff
@@ -45,6 +52,19 @@ export class Tabs {
   //
   //--------------------------------------------------------------------------
 
+  connectedCallback(): void {
+    this.mutationObserver?.observe(this.el, { childList: true, subtree: true });
```
jcfranco, line 56: I'd like to avoid using `subtree` here since it means the observer will run on every child update within the components (i.e., all tab-related components + their respective content). Some alternatives would be:

1. update `tab-nav` to forward relevant properties to its `tab-title` children
2. update the mutation observer callback to bail if the mutation target is not a tab-related component
3. reuse the tabs observer on its tab-nav

packages/calcite-components/src/components/tabs/tabs.tsx
```diff
@@ -133,6 +153,30 @@ export class Tabs {
    */
   @State() tabs: HTMLCalciteTabElement[] = [];
 
+  mutationObserver = createObserver("mutation", () => this.updateItems());
+
+  private updateItems = (): void => {
+    const { position, scale } = this;
+
+    Array.from(this.el.querySelectorAll("calcite-tab-nav")).forEach(
```
jcfranco, line 161: There should only be a single, direct, `tab-nav` child per tabs, right?

packages/calcite-components/src/components/tabs/tabs.tsx
```diff
@@ -133,6 +153,30 @@ export class Tabs {
...
+      (nav: HTMLCalciteTabNavElement) => {
+        nav.position = position;
+        nav.scale = scale;
+      }
+    );
+
+    Array.from(this.el.querySelectorAll("calcite-tab")).forEach((tab: HTMLCalciteTabElement) => {
```
jcfranco, line 168: I think you'll need to update queries to ignore nested tabs.

jcfranco (changes requested):
Looking good, @Elijbet! 🙂➕🕶🟰😎

packages/calcite-components/src/components/tab-nav/tab-nav.tsx
Elijbet: No it's not. Must have been testing something and left it accidentally. I just commutes a  revert, if I haven't pushed it yet, I will when addressing the new feedback.

packages/calcite-components/src/components/tab-title/tab-title.e2e.ts
```diff
@@ -338,153 +347,56 @@ describe("calcite-tab-title", () => {
...
-
-        expect(tabTitleEl).toEqualAttribute("scale", "l");
-        expect(tabTitleNav).toEqualAttribute("scale", "l");
-        expect(contentStyles.fontSize).toEqual("16px");
-        expect(contentStyles.lineHeight).toEqual("20px");
-      });
-    });
-  });
+    for (const { scale, styles } of scaleStyles) {
+      describe("when immediate parent element is `tabs`", () => {
```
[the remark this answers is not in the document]
Elijbet: Probably not. It was an extra ting test but I modified, I think it's better to remove that.

packages/calcite-components/src/components/tabs/tabs.tsx
Elijbet: Going with: update the mutation observer callback to bail if the mutation target is not a tab-related component.

packages/calcite-components/src/components/tabs/tabs.tsx
```diff
@@ -25,15 +26,21 @@ export class Tabs {
...
 
   /**
-   * Specifies the size of the component.
+   * Specifies the size of the component, defaults to `m`.
    */
   @Prop({ reflect: true }) scale: Scale = "m";
 
+  @Watch("position")
+  @Watch("scale")
+  handlePropsChange(): void {
```
jcfranco, line 40: `handlePropsChange` is a bit vague. Can you rename it to better convey what it does? It could be something simple as `handlePositionScaleChange` or `handleInheritableProps`

packages/calcite-components/src/components/tabs/tabs.tsx
```diff
@@ -133,6 +153,43 @@ export class Tabs {
    */
   @State() tabs: HTMLCalciteTabElement[] = [];
 
+  mutationObserver = createObserver("mutation", (mutationsList: MutationRecord[]) => {
+    console.log("mutationsList", mutationsList);
```
jcfranco, line 157: FIXME: Can you remove this console log?

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -194,7 +194,7 @@ function propToAttr(name: string): string {
 }
 
 /**
- * Helper for asserting that a property's value is its default
+ * Helper for asserting that a property's value is its default.
```
jcfranco, line 197: Thanks for updating the doc here. 

Can this be submitted separately? There aren't any changes related to this PR, so it's out of scope.

packages/calcite-components/src/components/tabs/tabs.tsx
```diff
@@ -133,6 +153,43 @@ export class Tabs {
...
+    for (const mutation of mutationsList) {
+      const target = mutation.target as HTMLElement;
+      if (
+        target.nodeName === "CALCITE-TAB-NAV" ||
+        target.nodeName === "CALCITE-TAB-TITLE" ||
```
jcfranco, line 162: FIXME: the current configuration of the mutation observer would only be invoked when direct children are added/removed. `tab-title`s are children of `tab-nav`s, so the observer would miss these.

Mutations from other children might be helping update `tab-titles`, so we could improve this in a follow-up PR for the current milestone.

packages/calcite-components/src/components/tabs/tabs.tsx
```diff
@@ -133,6 +153,43 @@ export class Tabs {
...
+  private updateItems = (): void => {
+    const { position, scale } = this;
+
+    const nav = this.el.querySelector("calcite-tab-nav");
+    if (nav) {
+      nav.position = position;
+      nav.scale = scale;
+    }
+
+    Array.from(this.el.querySelectorAll("calcite-tab:not(.calcite-tab-nested)")).forEach(
```
jcfranco, line 179: Where is `calcite-tab-nested` coming from?

packages/calcite-components/src/components/tabs/tabs.tsx
```diff
@@ -133,6 +153,43 @@ export class Tabs {
...
+      (tab: HTMLCalciteTabElement) => {
+        tab.scale = scale;
+      }
+    );
+
+    Array.from(this.el.querySelectorAll("calcite-tab-title")).forEach(
```
jcfranco, line 185: Can you update the selector so that it targets direct children of a `tabs`' direct `tab-nav` child?

packages/calcite-components/src/components/tab-title/resources.ts
```diff
@@ -4,8 +4,9 @@ export const CSS = {
   content: "content",
   contentHasText: "content--has-text",
   iconEnd: "icon-end",
-  iconStart: "icon-start",
   iconPresent: "icon-present",
+  iconStart: "icon-start",
+  tab: "tab",
```
jcfranco, line 9: FIXME: This new prop doesn't seem to be used in `tab-title`.

packages/calcite-components/src/components/tab-nav/resources.ts
```diff
@@ -0,0 +1,3 @@
+export const CSS = {
+  tabNav: "tab-nav",
```
jcfranco, line 2: For consistency, can you rename the key to be `container`? We can change the value to match in a follow-up PR.

jcfranco (changes requested):
Awesome stuff, @Elijbet. This is looking very close to landing. 🛬

packages/calcite-components/src/components/tab-title/tab-title.e2e.ts
```diff
@@ -338,153 +347,44 @@ describe("calcite-tab-title", () => {
...
-        const contentStyles = await content.getComputedStyle();
-
-        expect(tabTitleEl).toEqualAttribute("scale", "l");
-        expect(tabTitleNav).toEqualAttribute("scale", "l");
-        expect(contentStyles.fontSize).toEqual("16px");
-        expect(contentStyles.lineHeight).toEqual("20px");
-      });
-    });
-  });
+  describe("scale property", () => {
```
jcfranco, line 364: Sorry for not catching this earlier, but let's drop this test entirely in favor of a screenshot test.

Elijbet:
Awesome, thanks @jcfranco. Will take care of these today.

packages/calcite-components/src/components/tab-nav/usage/Basic.md
```diff
@@ -1,9 +1,11 @@
-When tab-nav is the only parent, tab-title can inherit its `scale` and `position` from tab-nav:
+Tab-nav and tab-title inherit their `scale` and `position` from tabs parent.
```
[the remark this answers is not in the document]
geospatialem: Doc consideration:

```
Tabs `scale` and `position` properties are inherited by it's child components, tab-nav and tab-title.
```

packages/calcite-components/src/components/tabs/tabs.e2e.ts
Elijbet: Opening a follow-up refactor Pr for this one.

packages/calcite-components/src/components/tab/tab.scss
```diff
@@ -18,17 +18,16 @@ section,
   @apply hidden h-full w-full;
 }
 
-:host([scale="s"]) {
+.scale-s section {
```
jcfranco, line 21: Not needed for this PR, but we can make these styles more robust by using a class vs an element.

packages/calcite-components/src/components/tab/tab.e2e.ts
```diff
@@ -18,43 +17,7 @@ describe("calcite-tab", () => {
...
   });
-
-  describe("when nested within calcite-tabs component", () => {
-    it("should render with medium scale", async () => {
-      const page = await newE2EPage({
-        html: `<calcite-tabs>${tabHtml}</calcite-tabs>`,
-      });
-      const element = await page.find("calcite-tab");
-      expect(element).toEqualAttribute("scale", "m");
-      expect(await (await element.getComputedStyle())["font-size"]).toEqual("14px");
```
jcfranco, line 32: Can you confirm these are covered by screenshot tests? If not, let's make sure to add coverage.

jcfranco (approved):
👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾
👾🕹️🕹️🕹️🕹️👾🕹️👾👾👾👾👾🕹️👾🕹️🕹️🕹️🕹️👾👾🕹️🕹️🕹️👾👾🕹️🕹️👾👾🕹️👾👾👾🕹️👾🕹️🕹️🕹️🕹️👾🕹️👾
👾🕹️👾👾🕹️👾🕹️👾👾👾👾👾🕹️👾🕹️👾👾👾👾🕹️👾👾👾👾🕹️👾👾🕹️👾🕹️🕹️👾🕹️🕹️👾🕹️👾👾👾👾🕹️👾
👾🕹️🕹️🕹️🕹️👾🕹️👾👾🕹️👾👾🕹️👾🕹️🕹️🕹️👾👾👾🕹️🕹️👾👾🕹️👾👾🕹️👾🕹️👾🕹️👾🕹️👾🕹️🕹️🕹️👾👾🕹️👾
👾🕹️👾👾🕹️👾🕹️👾🕹️👾🕹️👾🕹️👾🕹️👾👾👾👾👾👾👾🕹️👾🕹️👾👾🕹️👾🕹️👾👾👾🕹️👾🕹️👾👾👾👾👾👾
👾🕹️👾👾🕹️👾👾🕹️👾👾👾🕹️👾👾🕹️🕹️🕹️🕹️👾🕹️🕹️🕹️👾👾👾🕹️🕹️👾👾🕹️👾👾👾🕹️👾🕹️🕹️🕹️🕹️👾🕹️👾
👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾

packages/calcite-components/src/components/tab/tab.e2e.ts
Elijbet: These are some stories we have that cover different scale scenarios. 
bordered
closable
withIcons
setWidth
justTabNav

These 2 screenshots are stories with knobs (including scale) actually that have `_TestOnly` suffix. Technically we can use them to check more cases with scale by removing the suffix when testing. Otherwise, we'd have to rewrite them to run screenshots for each scale, currently, the output is `m`.

simpleDarkModeRTL_TestOnly 
borderedDarkModeRTL_TestOnly

jcfranco: If existing stories provide the same coverage that the E2E tests focused on, no further action is required. Otherwise, let's update existing tests to make sure we don't sacrifice coverage provided by removed E2E tests.

> Technically we can use them to check more cases with scale by removing the suffix when testing.

This would not be covered by automated testing though, right?

Elijbet: Not specifically for dark mode/RTL/bordered combinations

packages/calcite-components/src/components/tab-nav/tab-nav.scss
```diff
@@ -14,6 +14,36 @@
...
+  }
+  .scale-s {
+    ::slotted(calcite-tab-title) {
+      margin-inline-end: theme("margin.4");
+    }
+  }
+}
+
+//override margin-inline-end for the last child for tab-nav to implement the 20px padding on both ends instead
+:host([layout="center"]:not([bordered])) {
```
jcfranco, line 39: Can you merge this with the styles above? They share the same parent selector.

packages/calcite-components/src/components/tab/tab.tsx
```diff
@@ -69,7 +69,7 @@ export class Tab {
           role="tabpanel"
           tabIndex={this.selected ? 0 : -1}
         >
-          <section>
+          <section class={CSS.content}>
```
jcfranco, line 72: Not for this PR, but it'd be good to replace all selectors using `section` with the `content` class.

packages/calcite-components/src/components/tabs/tabs.stories.ts
```diff
@@ -163,16 +163,94 @@ export const setWidth = (): string => html`
   </div>
 `;
 
-export const justTabNav = (): string => html`
-  <calcite-tab-nav
-    position="${select("position", ["top", "bottom"], "top")}"
-    scale="${select("scale", ["s", "m", "l"], "l")}"
-  >
+const TabNavHTMLSimple = html`
```
jcfranco, line 166: Awesome job adding more coverage! ✨🧪🧪🧪✨

packages/calcite-components/src/components/tabs/tabs.tsx
```diff
@@ -166,7 +166,7 @@ export class Tabs {
     }
   });
 
-  private updateItems = (): void => {
+  private updateItems(): void {
```
jcfranco, line 169: ✨🏆✨

packages/calcite-components/src/components/tabs/tabs.tsx
jcfranco: Queries target all descendants, so this will be an issue when you have nested tabs with different props each. Not sure how common `position` and `scale` being different between nested tabs would be, so this could be an edge case.

Elijbet: Interim fix, I'll think of a way to have the selector do this for us instead.

Issue #6237, opened 2023-01-05
[calcite-flow-item] does not have a back button if within a shadow root


ethanbdev:
### Actual Behavior

With `showBackButton` being internal and defaulting to false, by default a flow item within a shadow root is missing the back button, breaking the functionality of a flow item.

### Expected Behavior

It should show a back button when it is the active flow item

### Reproduction Sample

https://codepen.io/eborgen/pen/vYaygZy

### Reproduction Steps

Observe that with a structure that puts a shadow root boundary between the `calcite-flow` and `calcite-flow-item`, the back button does not render.

### Reproduction Version

next.707

### Regression?

between beta.99 and next.707

### Impact

High - it breaks the workflow unless I set an internal prop and ignore the type error

### Esri team

ArcGIS Map Viewer

geospatialem:
Could be related to the [utilities](https://github.com/esri/calcite-components/blob/f4a55a81cc087ab88e11b51e40ca6a37ed379e75/src/utils/dom.ts#L113-L154) used, will need a refactor in the component to mitigate without the use of the utility functions.

jcfranco:
@ethanbdev I could be misinterpreting the issue, but from what I can recall, `flow` never supported `flow-items` within the shadow DOM of extraneous components. Can you update your codepen or provide one that shows this behavior?

ethanbdev:
> @ethanbdev I could be misinterpreting the issue, but from what I can recall, `flow` never supported `flow-items` within the shadow DOM of extraneous components. Can you update your codepen or provide one that shows this behavior?

That makes sense, I think we are not using the flow-item quite correctly. But with the removal of the `showBackButton` attribute it made it not work at all. It could be reasonable to have a tree structure like:
```
<calcite-flow>
  <my-custom-component>
    <calcite-flow-item />
  </my-custom-component>
</calcite-flow>
```
If that is not supported we can adjust the structure or consider a different component, like a panel with a slotted action in the header.

geospatialem:
Allocating to the February release for further discussion.

AdelheidF:
Yes, I use this in a few places

![image](https://user-images.githubusercontent.com/44481292/214936257-4a0d9eb8-b318-4157-adce-be1244e494e2.png)

ethanbdev:
[...]

PR #7608, opened 2023-08-26
feat(flow): add support for custom flow-item elements

jcfranco:
**Related Issue:** #6237 

## Summary

This enables `flow` to work with custom components using shadow DOM wrapping `flow-item`. 

Custom components will have to implement the [`FlowItemLike`](https://github.com/Esri/calcite-design-system/pull/7608/files#diff-82c222ab365cde13a1f1288d936611519dfd9bee1e283164b260ca554c04a191R3-R7) interface and set the new `custom-item-selectors`/`customItemSelectors` attr/prop to target custom components (see [E2E test](https://github.com/Esri/calcite-design-system/pull/7608/files#diff-9b86e64de24dfca441533c63ae0f6834bff10bffbde23fd8bb3989a2259e356cR315-R387) for vanilla JS example).

**Note**: `customItemSelectors` is intentionally marked internal since we don't have any documentation yet on developing custom components, which this is meant to support. We could additionally hide `FlowItemLike` in the doc site to avoid confusion until we have more documentation on this use case. I'm open to suggestions on this. @geospatialem @macandcheese @driskull

### DIFF
--- a/packages/calcite-components/conventions/README.md
+++ b/packages/calcite-components/conventions/README.md
@@ -409,3 +409,33 @@ The [`globalAttributes`](../src/utils/globalAttributes.ts) util was specifically
 ### BigDecimal
 
 `BigDecimal` is a [number util](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/utils/number.ts) that helps with [arbitrary precision arithmetic](https://en.wikipedia.org/wiki/Arbitrary-precision_arithmetic). The util is adopted from a [Stack Overflow answer](https://stackoverflow.com/a/66939244) with some small changes. There are some usage examples in [`number.spec.ts`](../src/utils/number.spec.ts).
+
+### Custom child element support
+
+In order to support certain architectures, parent components might need to handle custom elements that wrap their expected child items within shadow DOM that would prevent discovery when querying the DOM.
+
+For such cases, the following pattern will enable developers to create custom child/item components and have them work seamlessly with parent components.
+
+#### Parent component
+
+- Must provide a `customItemSelectors` property to allow querying for custom elements in addition to their expected children.
+- An interface for `HTML<ChildComponentItem>Element` must be created in the parent's `interfaces.d.ts` file, where the necessary child APIs must be extracted.
+  **`parent/interfaces.d.ts`**
+  ```ts
+  type ChildComponentLike = Pick<HTMLCalciteChildElement, "required" | "props" | "from" | "parent"> & HTMLElement;
+  ```
+  **`parent/parent.tsx`**
+  ```tsx
+    @Prop() selectedItem: HTMLChildComponentElement | ChildComponentLike;
+  ```
+
+#### Custom child component
+
+- Must implement the element interface expected by the parent (e.g., `ChildComponentLike`).
+
+#### Notes
+
+- This pattern should be applied sparingly and on a case-by-case basis.
+- We can refine this pattern as we go on, but additional modifications needed to handle the custom items workflow will be considered out-of-scope and thus not supported.
+- Until we have documentation covering creating custom elements, `customItemSelectors` must be made internal and any `ChildComponentLike` types must be excluded from the doc.
+- Please refer to https://github.com/Esri/calcite-design-system/pull/7608/ as an example on how this pattern is applied.
--- a/packages/calcite-components/src/components/flow/flow.tsx
+++ b/packages/calcite-components/src/components/flow/flow.tsx
@@ -137,11 +137,11 @@ export class Flow implements LoadableComponent {
   };
 
   updateFlowProps = (): void => {
-    const { el, items } = this;
+    const { customItemSelectors, el, items } = this;
 
     const newItems = Array.from<FlowItemLikeElement>(
       el.querySelectorAll(
-        `calcite-flow-item${this.customItemSelectors ? `,${this.customItemSelectors}` : ""}`
+        `calcite-flow-item${customItemSelectors ? `,${customItemSelectors}` : ""}`
       )
     ).filter((flowItem) => flowItem.closest("calcite-flow") === el);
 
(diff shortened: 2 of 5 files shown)


packages/calcite-components/src/components/flow/flow.tsx
```diff
@@ -61,6 +61,19 @@ export class Flow implements LoadableComponent {
...
+  //  Public Properties
+  //
+  // --------------------------------------------------------------------------
+
+  /**
+   * This property enables the component to consider other elements implementing flow-item's interface.
+   *
+   * @internal
+   */
+  @Prop() extraItemSelector: string;
```
driskull, line 75: I think we'll need to decide on this name so it can be standardized ™️ and used for any components that may need to adopt this pattern.

Would we want to expose the fact that this is a selector or just allow other item types?

Maybe this should be an array? Then we can just join them with a comma? Like `extraItems: string[]`

packages/calcite-components/src/components/flow/flow.tsx
```diff
@@ -126,9 +139,11 @@ export class Flow implements LoadableComponent {
...
     const { el, items } = this;
 
-    const newItems: HTMLCalciteFlowItemElement[] = Array.from(
-      el.querySelectorAll("calcite-flow-item")
-    ).filter((flowItem) => flowItem.closest("calcite-flow") === el) as HTMLCalciteFlowItemElement[];
+    const newItems = Array.from<FlowItemLikeElement>(
+      el.querySelectorAll(
+        `calcite-flow-item${this.extraItemSelector ? `,${this.extraItemSelector}` : ""}`
+      )
+    ).filter((flowItem) => flowItem.closest("calcite-flow") === el);
```
driskull, line 146: maybe something like:

```ts
const newItems = Array.from<FlowItemLikeElement>(
      el.querySelectorAll(
        [...extraItems, "calcite-flow-item"].join(",")
      )
    ).filter((flowItem) => flowItem.closest("calcite-flow") === el);
```

driskull (commented):
This is looking awesome @jcfranco. I think we just need to decide on the name & finalize it since this may pop up for other components.

packages/calcite-components/src/components/flow/flow.tsx
jcfranco: Agreed on standardizing. I plan to update the conventions doc with info on this.

I went with the current name because it captures what it'll be used for and the type of value that's expected. 

WRT the type, I think we should leave it as a string to keep it simple on both sides and because it would also align with DOM-querying APIs where a string `selectors` argument supports one or more selectors. 

Based on the above and your suggestions, I kind of like `customItemSelector(s)` since it conveys the use case better than the current name.

Alternatively, we can buy some extra time to think about alternative names if we keep everything internal for this release.

driskull: Sounds good!

packages/calcite-components/conventions/README.md
```diff
@@ -409,3 +409,33 @@ The [`globalAttributes`](../src/utils/globalAttributes.ts) util was specifically
...
+
+### Custom child element support
+
+In order to support certain architectures, parent components might need to handle custom elements that wrap their expected child items within shadow DOM that would prevent discovery when querying the DOM.
+
+For such cases, the following pattern will enable developers to create custom child/item components and have them work seamlessly with parent components.
+
+#### Parent component
+
+- Must provide a `customItemSelectors` property to allow querying for custom elements in addition to their expected children.
```
driskull, line 421: The plurality of `customItemSelectors` threw me off a little because its a single string. Would it make more sense as `customItemSelector`? I guess technically it can be a single or multiple selectors but its a single string. 🤔

packages/calcite-components/src/components/flow/flow.tsx
```diff
@@ -1,6 +1,6 @@
-import { Component, Element, h, Listen, Method, State, VNode } from "@stencil/core";
+import { Component, Element, h, Listen, Method, Prop, State, VNode } from "@stencil/core";
 import { createObserver } from "../../utils/observers";
-import { FlowDirection } from "./interfaces";
+import { FlowDirection, FlowItemLikeElement } from "./interfaces";
```
driskull, line 3: I think `FlowItemLikeElement` is fine but it could also be `CustomFlowItemElement`? I'm fine either way.

packages/calcite-components/conventions/README.md
driskull: I guess its fine since its referred to as `selectors` here: https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector

Just ignore my comment then:)

packages/calcite-components/src/components/flow/flow.tsx
```diff
@@ -126,9 +139,11 @@ export class Flow implements LoadableComponent {
   updateFlowProps = (): void => {
     const { el, items } = this;
 
-    const newItems: HTMLCalciteFlowItemElement[] = Array.from(
-      el.querySelectorAll("calcite-flow-item")
-    ).filter((flowItem) => flowItem.closest("calcite-flow") === el) as HTMLCalciteFlowItemElement[];
+    const newItems = Array.from<FlowItemLikeElement>(
+      el.querySelectorAll(
+        `calcite-flow-item${this.customItemSelectors ? `,${this.customItemSelectors}` : ""}`
```
driskull, line 144: nitpick: since `this.customItemSelectors` is used twice, might be nice to include it here:

```
const { el, items, customItemSelectors } = this;
```

driskull (approved):
Awesome! MV team will appreciate this :) 💯 🏅

packages/calcite-components/src/components/flow/flow.tsx
jcfranco: I like your name suggestion better, but the interface only requires the props expected by the parent and not all from `flow-item`.

packages/calcite-components/conventions/README.md
jcfranco: FWIW, I feel the same way about the name. 😅

Issue #6706, opened 2023-03-30
[number input] property to allow integers only


JonUihlein:
### Description

Add a property to `<calcite-number-input>` to only allow integers (no decimals).

Alternatively, adding a `places` property to control significant figures might be interesting (with `0` also forcing integers only). 

### Acceptance Criteria

In most cases, this would simply prevent users from typing `.` and `,`.
Invalid values supplied via the constructor, programmatically, or via a direct copy+paste could either be rounded or ignored. 

### Relevant Info

Currently, we must enforce this ourselves by overriding "bad" values in the associated `onchange` event.

Inspired by:
https://github.com/Esri/calcite-components/issues/2009#issuecomment-839204684

### Which Component

`<calcite-input-number>`

### Priority impact

p2 - want for current milestone

### Esri team

ArcGIS Maps SDK for JavaScript

hgonzago:
The `maximumfractiondigits` property in `Intl.NumberFormat` takes care of this.

JonUihlein:
Actually, this seems like a regression.

`<calcite-input type="number" />` prevented decimals in a previous release:
https://codepen.io/jonuihlein/pen/JjaqrJe

`<calcite-input type="number" />` and `<calcite-input-number />` allowing decimals by default:
https://codepen.io/jonuihlein/pen/NWLVaRr

This change in default behavior changed might have been outlined in release notes, but how do we go back to the previous behavior of "integers only"?

geospatialem:
> @geospatialem is there someone we can assign this too, or get an update? This use to work before as @JonUihlein mentioned. This behavior is preventing a migration to use this calcite input in Editor for the JS SDK. Thanks!

@banuelosj thanks for re-pinging this enhancement request - we're targeting this for the upcoming August milestone. cc @hgonzago

hgonzago:
Thanks @geospatialem 👍

jcfranco:
@benelan Can you take a look at this one?

benelan:
> The `maximumfractiondigits` property in `Intl.NumberFormat` takes care of this

![](https://i.imgflip.com/7xakt0.jpg)

Every single time...

geospatialem:
Verified in `1.7.0` with:

### Only integers (result = 55)
```html
<calcite-input-number integer value="5.5"></calcite-input-number>
```

### Floats allowed  (result = 5.5)
```html
<calcite-input-number value="5.5"></calcite-input-number>
```

PR #7612, opened 2023-08-28
feat(input, input-number): add property to set the maximum decimal places

benelan:
**Related Issue:** #6706

## Summary

Add a `places` property to `<calcite-input type="number">` and `<calcite-input-number>` for specifying the maximum number of decimal places. User input is prevented for decimals greater than the maximum and the extra decimals are stripped when pasting.

### Why not maximumFractionDigits
Unfortunately we can't use `Intl.NumberFormat`'s [maximumFractionDigits option](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#maximumfractiondigits) because of our floating point precision and unsafe integer implementation. 

We use `BigDecimal.format()` for localization. It uses `Intl.NumberFormat` to format the integers (as `BigInt`), and then slaps the decimals on afterwards, changing their numbering system character-by-character. This works because all of the non-numberingSystem localization changes (minus/plus sign, group separator, decimal separator) happen before the decimal places. 

https://github.com/Esri/calcite-design-system/blob/482c4eab3561f7a7412eeb47ad4841163da4b987/packages/calcite-components/src/utils/number.ts#L69-L81

Trying to treat the decimal places as an integer also wouldn't work because it would strip leading zeros and do other auto-formatting, e.g. `1.001` would become `1.1`. Additionally, trying to format the number without splitting it would result in a loss of precision.

### Exponential notation edge case 

We need *special* logic for exponential numbers, or else `1.2e5` would become `1.2` if `places === 1`.

When the exponent is negative (meaning the number is fractional), it is expanded to a normal decimal number to check if it has more places than the max. If the decimal places are less than the max, the number stays in exponential notation. Otherwise, the extra places are stripped and the input value is changed from exponential notation to decimal notation. Converting it back to exponential notation after stripping the extra decimal places probably wouldn't be possible in all cases.

The UX is clunky when this edge case occurs and any better ideas on how to handle it would be appreciated. As an example, `1.2e-3` (which is `0.0012`) becomes `0.00` when `places === 2`. The change happens immediately after typing the `3` which is a bit disorienting.

This is a good example of why restricting user input isn't great. Maybe it would be better to implement `places` with form validation instead? Or wait until blur to sanitize the decimal `places` of exponential numbers?


eriklharper:
Curious why we wouldn't just want to use `step` to inform the decimal precision for numbers?  A step of `.001` would imply 3 decimal places, `.0001` would be 4, etc.

eriklharper:
> Steps just specifies how much the increment/decrement buttons change the value by

To the contrary, what I notice with native chrome behavior is that when you have a step of `5` for example and you type any value less than 5 except `0` (like `3`), then when you nudge the value up with the keyboard it moves to the value 5 instead of adding 5 to the current value.  See here for a codepen demonstrating this: https://codepen.io/eriklharper/pen/rNoeyJa

This behavior aligns with my understanding of what the spec says about `step`:

> The step attribute indicates the granularity that is expected (and required) of the value or values by _limiting_ the allowed values.
https://html.spec.whatwg.org/multipage/input.html#attr-input-step

Limiting the allowed values is exactly what nudging on the native number input is doing.  Rather than simply adding 5 to the current value, it determines which is the next highest _allowed_ value based on the step when the up arrow key is pressed.

The [MDN docs on the step attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/step) don't do a good job of explaining this even though the link to the spec is on that page.

benelan:
I was being simplistic but I was trying to say that its a difference between constraints for increment/decrement buttons and typing/direct user input. In your example the value is limited to multiples of the step when nudging, but you can still type whatever you want. 

Just to make sure we're on the same page, is this the step to places mapping you had in mind?

```
step=1337, places=0
step=5, places=0
step=0.1, places=1
step=0.012345, places=6
```

Using the same prop for these would be a breaking change because Inputs with an integer step like `5` would no longer accept any decimal places, which is definitely a valid use case. 

Take money, for example. The places would need to be 2, but I doubt anyone wants to increment by cents; they would want to increment by dollars (step=1).

eriklharper:
>  In your example the value is limited to multiples of the step when nudging, but you can still type whatever you want.

Correct, yes, it just puts the input into an invalid state when it is submitted inside a `<form>` in that case, which is expected behavior.

> Just to make sure we're on the same page, is this the step to places mapping you had in mind?

Yes, that mapping is what I had in mind for using `step` to determine this, but now that I've thought about it more, I don't think `step` should actually restrict what the user can type into the input, because that's not what `step` is for.  I feel like since the original ask of this issue involves restricting what the user can type, and we try as much as possible not to keyboard-block the user with our generic/multi-purpose input components, it just _feels_ wrong to do this in `input-number`, either with `step` or a custom prop.  It feels like maybe a more purpose-built component would be better suited to this, one that maybe doesn't just use the built-in `<input>` element, and the input functionality can be completely controlled and customized to fit the requirements.  This is just my opinion, of course and I don't think it would be too terrible and extra maintenance burden to maintain a separate `input-integer` type component for this.

> Using the same prop for these would be a breaking change because Inputs with an integer step like 5 would no longer accept any decimal places, which is definitely a valid use case.

Exactly, totally agree, which is why I don't think we should use `step` to restrict what the user can type.

> Take money, for example. The places would need to be 2, but I doubt anyone wants to increment by cents; they would want to increment by dollars (step=1).

If we have asked-for use cases for currency, that's another good reason to create a new `input-currency` component that can be purpose-built specifically for that kind of value.

Overall, I think generic number inputs need to allow all types of number expressions to be typed, since they're just for entering numbers, which can be positive, negative, decimals, exponential, integer, etc.  When we start adding off-switches into the number input that disallow typing any character that composes a valid number value, that starts opening up the possibility of adding more off switches to restrict other types of characters like `-`, `e`, `,` etc.  At that point it probably makes more sense to just add a single regular expression prop so that the developer can supply their own custom restriction instead of implementing each unique restriction as a one-off property?

eriklharper:
> At that point it probably makes more sense to just add a single regular expression prop so that the developer can supply their own custom restriction instead of implementing each unique restriction as a one-off property?

The more I think about this idea, I'm also unsure if the the regex prop should actually restrict what the user can type or if it should behave like the `pattern` attribute which is just to control constraint validity.  Nevertheless, if we do decide to use a regex approach, we should name that property something different than `pattern` if the intention is to actually restrict typing certain characters unmatched in the expression.

driskull:
> The more I think about this idea, I'm also unsure if the the regex prop should actually restrict what the user can type or if it should behave like the pattern attribute which is just to control constraint validity.

Agreed. We shouldn't mess with what a user is doing because its not a good UX to stop them while they are typing. We should show an error state.

eriklharper:
> We should show an error state.

I kinda wonder if our components already had error state UI implemented, if that would satisfy developers asking for number input restriction?  If the component automatically handled error communication, it would take the burden off the developer to have to do that themselves in order to provide a better UX for the end user, and maybe at that point having to restrict input is unnecessary since the end user would be informed that their value is invalid?

benelan:
Yeah I don't like restricting user input either like I mentioned in the pull request body. However I don't think creating a dozen separate input components for slightly different features is the way to go.

Imo we should either do it with props like this pull request or do it with form validation instead, which I think is the better option. 

cc @jcfranco what are your thoughts on this?

eriklharper:
Another option we could consider is exposing a `calciteInputNumberKeyDown` event that you could listen for and call `event.preventDefault()` whenever the key matches a regex expression like `/[^A-z.-]/g` which would disallow all alpha characters, `.` and `-` and `e`.

jcfranco:
Sorry for the late reply, but here's my take:

* 👍 on not restricting user input as they type as much as possible
* There's value in providing convenience props to avoid unnecessary workflows and reduce boilerplate by relying solely on validation. For example, in a color-picker or similar context, I know my RGB inputs will never use fractions, so
    * why do I need to allow the user make a mistake they would need to correct after showing a validation error?
    * why not add a convenience to reduce boilerplate? I can intercept this myself, but this is a common scenario that I would need to duplicate logic or create a reusable helper that might not make sense to make public
* 👍 on `pattern` behaving the same as it does w/ `<input>`.
* 👍 on not introducing an integer-specific component for this as it still seems within `input-number`'s scope.
* Exposing component-specific key events would be flexible, but also open a can of worms for us

I also synced up w/ @benelan separately and we discussed and agreed to:

1. add a simple/integer-only mode that wouldn't handle exponential numbers nor unsafe integers (we'll need to highlight this in the doc) to address #6706 
2. abandon `places` in favor of `pattern`, which plays better w/ validation workflows.
3. as part of the validation effort, explore adding convenience props to handle common validation patterns/messages for consistency

Issue #7046, opened 2023-05-26
List: Expose Events for dragging between Lists


Csmith246:
### Actual Behavior

For the Calcite List component, when you set `dragEnabled` to `true`, and the `group` value of more than two lists to the same key, and then move items between the lists, none of the current Event's on the List get triggered. This makes it difficult to react to drags between lists.

### Expected Behavior

When a List Item gets dragged from one List to a different List, maybe trigger events such as `calciteListDragOut` and `calciteListDragIn` on the respective lists. This would broadcast what specifically has occurred within the lists, and allow the wider application to perform any behaviors that are tied to a move between lists. 

### Reproduction Sample

~https://codepen.io/csmith55/pen/YzJMzYg?editors=1001~
https://codepen.io/geospatialem/pen/OJaGeVN

### Reproduction Steps

Notice in the sample that:

- Each list has the same `group` value
- All events are being tracked on List 1

If you drag any list items from list 2 or 3 into list 1, no events are fired. If you drag any out of List 1, still no events are fired.

### Reproduction Version

~https://www.npmjs.com/package/@esri/calcite-components/v/1.4.1-next.4~
1.5.1

### Priority impact

p2 - want for current milestone

### Impact

Technically, I don't believe this behavior has ever worked, but we got around that in our application by finding a reference to the internal SortableJS object and adding `onAdd` and `onRemove` methods. This fix on the Calcite side (https://github.com/Esri/calcite-components/commit/b4bbdf3d709b8b1adb04fe090957451f731f48bb) caused our work around not to work anymore, so an official solution is now desirable. Thanks!

### Esri team

ArcGIS Instant Apps

Csmith246:
@driskull Would you have any update on this?

geospatialem:
With the `list` sorting support in `1.5.1`, it looks like the events are exposed with the exception if a new `list-item` is added to the `list`. The request should include adding a listener if a new `list-item` is added, perhaps with the `calciteListOrderChange` event?

Updated Codepen to `list`/`list-item` in `1.5.1`: https://codepen.io/geospatialem/pen/OJaGeVN

Recording showing events are registered, unless a `list-item` is added to the first `list`:
![list-event-listeners](https://github.com/Esri/calcite-design-system/assets/5023024/f12335b3-49d6-48c1-bdfb-e0a93d07b376)

Csmith246:
[...]

PR #7614, opened 2023-08-29
fix(list, sortable-list, value-list): Emit calciteListOrderChange when dragging between lists

driskull:
**Related Issue:** #7046

## Summary

TLDR: Basically, we need to pause draggable component connected/disconnected lifecycle events while a component is being dragged. Having these lifecycle methods kick off during a drag causes SortableJS errors.

- SortableComponent
  - Refactors logic to prevent any SortableJS component from doing its lifecycle logic when a component is being dragged
  - Previously, it was preventing all SortableJS components Sortable from being destroyed or created but that was causing issues by not emitting events when an item was moved from one list to another.
  - The nested component check that was previously being used isn't ideal because two different lists don't have to be nested to drag items between each other.
  - We need all lists to still continue emitting events when necessary, we just don't want their lifecycle methods to kick off when an item is being dragged. Otherwise, JS errors are thrown.
- Components
  - Updates SortableComponent components to not do any lifecycle callbacks when an item is being dragged to prevent any JS errors that SortableJS was throwing.
  - This was because in connectedCallback, sortable components were setting up the sortable instance, connecting the observer, modifying items, etc. We don't want the component to do this while an item is being dragged.  
  - The same thing as above was happening on disconnectedCallback.
  -  This fix stops all those errors that occurred while dragging an item from one list to another.

## Assumptions

It is reasonable to not do any lifecycle events for any draggable component while a component is being dragged


packages/calcite-components/src/utils/sortableComponent.ts
```diff
@@ -59,14 +57,14 @@ export interface SortableComponent {
   onDragSort: (event: Sortable.SortableEvent) => void;
 
   /**
-   * Element dragging started.
+   * Called when any sortable component drag starts.
    */
-  onDragStart?: (event: Sortable.SortableEvent) => void;
+  onDragStart: () => void;
```
driskull, line 62: Should we rename these to something like `disableObservers`? Since that is what they are basically for.

packages/calcite-components/src/utils/sortableComponent.ts
```diff
@@ -59,14 +57,14 @@ export interface SortableComponent {
...
 
   /**
-   * Element dragging ended.
+   * Called when any sortable component drag ends.
    */
-  onDragEnd?: (event: Sortable.SortableEvent) => void;
+  onDragEnd: () => void;
```
driskull, line 67: Should we rename these to something like `enableListeners`? Since that is what they are basically for.

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -477,6 +486,7 @@ export class List implements InteractiveComponent, LoadableComponent, SortableCo
   }
 
   onDragSort(event: SortableEvent): void {
+    this.parentListEl = this.el.parentElement.closest("calcite-list");
```
driskull, line 489: Since connectedCallback won't be fired, we should update the parentList here once dragging/sorting is done.

driskull (commented):
questions

packages/calcite-components/src/components/list/list.e2e.ts
```diff
@@ -471,6 +471,7 @@ describe("calcite-list", () => {
 
     it("works using a mouse", async () => {
       const page = await createSimpleList();
+      //const listOrderChangeSpy = await page.spyOnEvent("calciteListOrderChange");
```
driskull, line 474: @jcfranco I need your help with this one. If I add this spy, the test fails with the following...

```
ReferenceError: You are trying to `import` a file after the Jest environment has been torn down. From src/components/list/list.e2e.ts.

      at Object.get (../../node_modules/@stencil/core/testing/index.js:7:14)
      at Object.<anonymous> (../../node_modules/@stencil/core/testing/index.js:3772:79)

 RUNS  src/components/list/list.e2e.ts
/Users/dris0000/Sites/git/calcite-design-system/node_modules/@stencil/core/testing/index.js:3925
    throw new Error("The page was already closed");
          ^

Error: Caught error after test environment was torn down
```

jcfranco: Took a look and it looks like this error is caused by the drag event payload being serialized and there being circular JSON structures from the payload elements. I don't know yet what the proper fix for this would be, but you can work around it by manually adding an event listener to the page and tracking/storing the times it was emitted changes on `window` and then using that value to assert the final call count.

driskull:
Test refactor issue: https://github.com/Esri/calcite-design-system/issues/7643

packages/calcite-components/src/components/list/list.e2e.ts
driskull: Cool yeah that works for now. Thanks 👍

jcfranco (approved):
📃🤏➡️

Issue #7616, opened 2023-08-29
[Design] Consistent width-scale , height-scale values across components


macandcheese:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Components like Shell Panel, Sheet, and Modal have `width-scale` properties - we should consolidate these into a shared set of widths if possible, and reference those. Same for `height-scale`.

Dropdown also has `width` - but that property might not be needed anymore - just have it set by the `scale`, or longest text string.

### Acceptance Criteria

A shared set of default width values is available for components that use `width-scale` and `height-scale`

### Relevant Info

cc @ashetland @SkyeSeitz 

### Which Component

Shell Panel, Sheet, Modal, Dropdown (?)

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

Calcite (design)

geospatialem:
For design expertise - to confirm components prior to assigning to a future dev milestone.

SkyeSeitz:
For confirmation - was the aim of this issue to establish shared s/m/l values for the width property across components? And was that resolved in https://github.com/Esri/calcite-design-system/pull/10786?

PR #7617, opened 2023-08-29
feat(sheet): Update default widths

macandcheese:
Updates the default Sheet css variables. Users can of course override any of these as needed but these work better for most expected Sheet use cases.

## Summary
Opened this to create a set of shared widths for other components: https://github.com/Esri/calcite-design-system/issues/7616


driskull (approved):
Looks good 👍

Issue #6861, opened 2023-04-25
Unexpected keyboard navigation experience with nested trees


SaraLawrence:
### Summary

Questioning the use of nested trees and the effect on keyboard only users. Also, wondering why slots are implemented at the tree-item level rather than at the tree level? 

### Actual Behavior

Keyboard only users expect to be able to navigate between nodes with the up and down arrow keys and also move into a group of children using the down arrow. It is confusing to have to first press right or left arrow in order to move up or down.

Also, screen reader support for trees is not consistent and implementing them as a series of trees when visually they appear to be one component, can be confusing for blind screen reader users. 

Example: JAWS only recognizes two of the four patterns in the [sample codepen](https://codepen.io/paulcp/pen/GRYjaoY):

![Tree views in JAWS list view](https://user-images.githubusercontent.com/112504381/234399156-527ada15-7ecc-4beb-b80c-7a2f7b1484e3.JPG)

### Expected Behavior

- Expected behavior for `tab` is to move into the tree and out of the tree. With Calcite, `tab` moves to what visually appears to be the next "depth 1" item. This is confusing for screen reader users because they can't see the relationship that appears visually between "Grandfather" and "Great uncle".
- Ideally, if it's one cohesive unit visually and there are expected keyboard commands associated with the role, it should be presented to screen readers as one cohesive unit, i.e, one `role="tree"`, and the commands should follow the pattern, with down-arrow and up-arrow always available.	

All of the W3C and Deque examples use `role="group"` to contain child items, rather than nesting trees and the keyboard behavior is more straightforward. 

- [W3C tree file list pattern example (using computed properties) ](https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-1/treeview-1a.html)opens in a new window
- [W3C tree file list pattern example (using declared properties) ](https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-1/treeview-1b.html)opens in a new window
- [W3C tree navigation pattern example (with computed properties) ](https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-2/treeview-2a.html)opens in a new window
- [W3C tree navigation pattern example (with declared properties) ](https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-2/treeview-2b.html)opens in a new window

### Reproduction Sample

https://codepen.io/paulcp/pen/GRYjaoY

### Reproduction Steps

[...]

PR #7618, opened 2023-08-29
fix(tree): improve keyboard navigation

jcfranco:
**Related Issue:** #6861 

## Summary

Improves tree's keyboard navigation to move across all items and users can now tab to the next focusable element without having to traverse all items.

### Notes

* items are now programmatically-focusable only (e.g., `tabindex="-1"`) since the tree moves focus
* keyboard navigation based on https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-1a/#kbd_label
* consolidated tree tests and improved coverage
* moved all navigation logic to the tree-level
* refactored item traversal logic to cover all cases
* moved tree `keyDownHandler` to use VDOM event.

### DIFF
--- a/packages/calcite-components/src/components/tree/tree.tsx
+++ b/packages/calcite-components/src/components/tree/tree.tsx
@@ -253,16 +253,9 @@ export class Tree {
     const root = this.el;
     const target = event.target as HTMLCalciteTreeItemElement;
 
-    if (
-      !(isTreeItem(target) && this.el.contains(target)) ||
-      (event.key !== "ArrowRight" &&
-        event.key !== "ArrowDown" &&
-        event.key !== "ArrowLeft" &&
-        event.key !== "ArrowUp" &&
-        event.key !== "Home" &&
-        event.key !== "End" &&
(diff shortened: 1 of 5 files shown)


packages/calcite-components/src/components/tree/tree.tsx
```diff
@@ -244,72 +245,103 @@ export class Tree {
...
+      return;
+    }
+
+    const root = this.el;
     const target = event.target as HTMLCalciteTreeItemElement;
 
-    if (!(root === this.el && target.tagName === "CALCITE-TREE-ITEM" && this.el.contains(target))) {
+    if (
+      !(isTreeItem(target) && this.el.contains(target)) ||
+      (event.key !== "ArrowRight" &&
```
driskull, line 258: nitpick: Might be nicer as `const allowedKeys = ["ArrowDown", "ArrowLeft",...]`

Then `if(!allowedKeys.includes(event.key)) { return; }`

driskull (approved):
👏 👍

Issue #7588, opened 2023-08-23
[Calcite Date Picker] Year value adds leading zeros when two numbers are added to the component


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Filed as Online bug: `BUG-000159719`.

When a user types in a date with two digits for the year the year is provided a value with two leading zeros. 

For instance: `8/1/23` > `0023-08-01`

Came up in the efforts of https://github.com/Esri/calcite-design-system/issues/3526, but is a separate bug. Sectioning out for clarity and prioritization. 

cc: @banuelosj @amring1

### Expected Behavior

Not sure if we should automatically add in the first two numerical values, but at a minimum the value shouldn't register since its invalid. Perhaps in this case the value remains as `""`? cc: @jcfranco 

### Reproduction Sample

https://codepen.io/banuelosj/pen/qBQyNrQ

### Reproduction Steps

1. Open https://codepen.io/banuelosj/pen/qBQyNrQ
2. Type a date with the year ending in two digits, for instance 8/1/23
3. Select outside the date to close the component
4. Observe the date is displayed as `0023-08-01`

### Reproduction Version

1.4.3

### Priority impact

p1 - need for current milestone

### Impact

Should be mitigated ASAP for inclusion in the upcoming R3 releases.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

Calcite (dev)

anveshmekala:
Regarding expected behavior, native input of type="date" does allow users to type dates before the specified `min` and return them on `change` event.  If we consider the example [here ]( min="2023-08-01") , the change event target value is formatted similar to `calcite-date-picker` one. The difference in behavior is related to the format we display in the input field rather than the target value ( 8/1/23 vs 08/01/0023). Since validation effort is handled in https://github.com/Esri/calcite-design-system/issues/4276, we can utilize this to display right format `08/01/0023` in the input field.


Thoughts? @geospatialem @jcfranco

jcfranco:
@anveshmekala I like the idea of normalizing the value when the input is blurred or when the date is modified via the calendar, but I think this is slightly different. 

[...]

PR #7622, opened 2023-08-29
feat(input-date-picker): normalize year to current century

anveshmekala:
**Related Issue:** #7588

## Summary

Normalizes year to current century. 

Ex: If the user types `01/01/20` the `value` will be corrected to `01/01/2020`

### DIFF
--- a/packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
+++ b/packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
@@ -17,6 +17,7 @@ import {
   dateFromISO,
   dateFromLocalizedString,
   dateFromRange,
+  datePartsFromISO,
   datePartsFromLocalizedString,
   dateToISO,
   inRange,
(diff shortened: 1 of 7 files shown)


packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -1110,4 +1126,29 @@ export class InputDatePicker
           )
           .join("")
       : "";
+
+  private normalizeToCurrentCentury(twoDigitYear: number) {
```
driskull, line 1130: Would this be beneficial as a util outside of the component that other components may use? Or is this the only place a user would type in a year?

driskull (approved):
👍 Nice!

packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
anveshmekala: One component i couldn't think of is date-picker but there is a strange behavior in date-picker which is normalizing year to 19th century when user types any two digit year.

packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -155,8 +157,16 @@ export class InputDatePicker
       let newValueAsDate: Date | Date[];
 
       if (Array.isArray(newValue)) {
+        if (isTwoDigitYear(newValue[0]) || isTwoDigitYear(newValue[1])) {
```
anveshmekala, line 160: Reason for having separate method `isTwoDigitYear` is to avoid multiple method invokations when we are mutating `value` prop.

driskull (approved):
LGTM!

jcfranco (commented):
Awesome some, @anveshmekala! 

I did have some notes:

Nice job on https://github.com/Esri/calcite-design-system/pull/7622/files. I did spot a few things:

 

1. Can we avoid `normalizeYear` and have the behavior enabled by default?
2. We should only be normalizing when users type in a 'shorthand' year value and not when the value is directly set as `YY-MM-DD`. Our doc explicitly calls out ISO format for `value`.

PR #7624, opened 2023-08-30
fix(list): Stop emitting calciteListChange when a list-item is disabled or closed.

driskull:
**Related Issue:** #7627

## Summary

- calciteListChange shouldn't be emitting when an item is disabled or closed.


jcfranco (approved):
💥

Issue #7575, opened 2023-08-22
auto-define.js issue prevents Calcite Components React from loading


arjanvanzutphen:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Many, if not all, components don't load due to an auto-define.js error log. 
I noticed that also components are mentioned in the error logs that I don't use in my application
![image](https://github.com/Esri/calcite-design-system/assets/25850072/91bde87b-d876-472e-9a7d-6daca285fc10)

### Expected Behavior

The application loads as expected

### Reproduction Sample
https://github.com/arjanvanzutphen/ccr-auto-define-issue

### Reproduction Steps

1. Clone the repo
2. Run `npm install`
3. Run `npm run dev`
4. Open the page. Most likely: http://localhost:5173
5. See errors in Devtools' console

### Reproduction Version

1.6.0. in sample

### Relevant Info

1.6.1 shows the same behavior and is not showing the components as they should

### Regression?

1.5.1

### Priority impact

p3 - want for upcoming milestone

### Impact

The app doesn't get loaded. Version 1.6.0 is in this way unusable

### Calcite package

- [ ] @esri/calcite-components
- [X] @esri/calcite-components-react

### Esri team

Professional Services - Services Delivery

geospatialem:
CodeSandbox can sometimes takes a few days to catch up to the `latest`, but if you open [the template](https://cdn.jsdelivr.net/npm/@esri/calcite-components@latest/dist/calcite/calcite.esm.js) and select the "Upgrade to latest" button in the "Dependencies" it will update to `1.6.1`. 
![image](https://github.com/Esri/calcite-design-system/assets/5023024/e8c699c8-25ca-40a6-8830-cfdc4bec95a4)

Or if its easier to setup a use case, here's [the sample with `1.6.1` setup](https://codesandbox.io/s/vigilant-james-x86ff5). 

@arjanvanzutphen WRT the above, could you add a sample with either the template upgraded to latest, or with the above link?

jcfranco:
Looks like the `isBrowser` check we set up is failing in this case. `window`, `document`, `global` and `process` are available on the browser preview. cc @benelan

jcfranco:
Maybe we can add some additional checks on the `process` object since we can't rely on these not being available because of bundlers. We can also look at using https://www.npmjs.com/package/browser-or-node or beefing up our checks based on that.

[...]

PR #7632, opened 2023-08-31
fix: make sure components are defined in environments like in codesandbox

benelan:
**Related Issue:** #7575

## Summary

CodeSandbox exposes `process`, which makes it look like NodeJS. The only way to determine it should be treated as the browser is the non-standard value they use for `process.platform`.

ref: https://nodejs.org/api/process.html#processplatform


jcfranco (approved):
💥

PR #7634, opened 2023-08-31
fix(combobox): prevent deselecting items via keyboard in single-persist mode

jcfranco:
**Related Issue:** #4738 

## Summary

This also adds test coverage for keyboard toggling for all selection-modes.


driskull (approved):
👍 Nice work 🍾

PR #7638, opened 2023-08-31
feat(input-date-picker): normalize year to current century for user typed values only

anveshmekala:
**Related Issue:** #7588 

## Summary

This will normalize two digit year typed by user to current century. 

Ex: If the user types `01/01/20` the value will be corrected to `01/01/2020` when user focus out of the input field or press a `Enter` key.


jcfranco (approved):
Awesome! 🎉

PR #7639, opened 2023-08-31
fix: dropdown context for accessibility

alisonailea:
**Related Issue:** #4866

## Summary
Provides context for the Dropdown component using aria-label


packages/calcite-components/src/components/dropdown/dropdown.tsx
```diff
@@ -606,7 +613,43 @@ export class Dropdown
   };
 
   private updateSelectedItems(): void {
+    const selectedLabels: string[] | string[][] = [[], []];
     this.selectedItems = this.items.filter((item) => item.selected);
+    this.label = this.selectedItems
+      .reduce((acc, item, idx) => {
+        // @ts-expect-error TODO: write special type for selectedItems which includes the parentElement as a DropDownGroup
+        const group = item.parentElement.groupTitle;
```
driskull, line 621: maybe this should use `item.closest("calcite-dropdown-group")` to get the group

packages/calcite-components/src/components/dropdown/dropdown.tsx
```diff
@@ -237,7 +237,11 @@ export class Dropdown
   render(): VNode {
     const { open, guid } = this;
     return (
-      <Host>
+      <Host
+        {...{
+          "aria-label": this.label,
```
driskull, line 242: can this be moved onto the `calcite-trigger-container` element so that its within the shadowDOM?

driskull (commented):
Just a few comments. Looks good 👍

jcfranco:
Can you update the PR title to follow our contributing doc? Just noticed the [Stencil example link](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md#commit-message-format) needs to be updated. I'll submit a PR for this shortly.

alisonailea:
> Can you update the PR title to follow our contributing doc? Just noticed the [Stencil example link](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md#commit-message-format) needs to be updated. I'll submit a PR for this shortly.

I removed the issue number from the title. Was that the only thing wrong?

packages/calcite-components/src/components/dropdown/dropdown.tsx
alisonailea: aria- properties don't work when used on shadowDOM elements

driskull:
Hey @alisonailea I think the title still needs a little tweaking. It should read something like: `fix(dropdown): Add context for accessibility`

Just needs to highlight that the dropdown is the component being fixed.

I use a cheat sheet for the titles and commits: https://cheatography.com/albelop/cheat-sheets/conventional-commits/

packages/calcite-components/src/components/dropdown/dropdown.tsx
driskull: Hey @alisonailea, Is it just not reading correctly? We have many other aria attributes inside shadowDOM of components so I think this should be doable.

The reason I think it would be better in the shadowDOM is so that an end user doesn't have the ability to override it or remove it accidentally. We try not to expose these kind of internals outside of the component.

alisonailea:
closing in favor of larger bugfix which will fix the larger missing aria context

PR #7648, opened 2023-08-31
fix(input, input-number): correctly sanitize numbers when pasting string with 'e'

benelan:
**Related Issue:** #

## Summary

Prevent error when pasting a string like `123test` into `calcite-input-number` and `calcite-input type="number`. The error was due to our sanitization logic, which splits on `e` for exponential notation and sanitizes each part. In the `123test` example, the rest of the letters after the `e` are sanitized causing a `reduce()` on nothing. On the other hand, `123fine` works fine because the `e` is at the end of the string so the falsy check kicks in before the `reduce()`. The fix uses `join()`, which doesn't throw errors on empty strings.

## Reproduction
Sample: https://developers.arcgis.com/calcite-design-system/components/input-number/#sample

Steps:
1. Open the sample and paste `123test` into the input
2. Notice `test` isn't sanitized
3. Open the console and you'll see the following error

```
Uncaught TypeError: reduce of empty array with no initial value
    o https://js.arcgis.com/calcite-components/1.6.1/p-06a878f3.js:6
    f https://js.arcgis.com/calcite-components/1.6.1/p-06a878f3.js:6
    f https://js.arcgis.com/calcite-components/1.6.1/p-06a878f3.js:6
    o https://js.arcgis.com/calcite-components/1.6.1/p-06a878f3.js:6
    inputNumberInputHandler https://js.arcgis.com/calcite-components/1.6.1/p-1c7861e8.entry.js:6
```

### DIFF
--- a/packages/calcite-components/src/utils/number.spec.ts
+++ b/packages/calcite-components/src/utils/number.spec.ts
@@ -41,7 +41,7 @@ describe("parseNumberString", () => {
 
     expect(parseNumberString(lettersAndSymbols)).toBe("");
     expect(parseNumberString(lettersAndSymbolsWithLeadingNegativeSign)).toBe("");
-    expect(parseNumberString("123test")).toBe("123");
+    expect(parseNumberString("123test")).toBe("123e");
   });
 
   it("returns valid number string for string values that compute to a valid number", () => {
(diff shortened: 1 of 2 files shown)


jcfranco (approved):
🎉

Issue #7533, opened 2023-08-16
Time Picker: Clicking nudge value buttons should focus the corresponding input


eriklharper:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Clicking `time-picker`'s up/down nudge buttons don't gain their own focus, but should probably focus the input the buttons are affecting the value of.

### Acceptance Criteria

Clicking the up or down buttons in `time-picker` for any of the editable values should focus the corresponding input.

### Which Component

`time-picker`

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

Calcite (dev)

geospatialem:
Verified with `1.8.0-next.0`

![verify-7533](https://github.com/Esri/calcite-design-system/assets/5023024/cb9bc82d-51a0-42f2-a6ae-4eee4265ca84)

PR #7650, opened 2023-09-01
fix(time-picker): focus corresponding input when nudge buttons are clicked

eriklharper:
**Related Issue:** #7533

## Summary

This fixes an issue where clicking the up/down buttons doesn't focus the input that the buttons are changing the value of.


packages/calcite-components/src/components/time-picker/time-picker.tsx
```diff
@@ -309,6 +312,11 @@ export class TimePicker
     }
   }
 
+  @Listen("mousedown")
```
jcfranco, line 315: For compatibility, can you use `pointerdown` instead? Applies to all `mouse`-related code.

packages/calcite-components/src/components/time-picker/time-picker.tsx
```diff
@@ -559,6 +614,8 @@ export class TimePicker
     }
   };
 
+  private mouseActivated = false;
```
jcfranco, line 617: Can you move this to the `Private Properties` section?

packages/calcite-components/src/components/time-picker/time-picker.tsx
```diff
@@ -309,6 +312,11 @@ export class TimePicker
     }
   }
 
+  @Listen("mousedown")
+  hostMouseDownHandler(): void {
```
jcfranco, line 316: For consistency, can you drop `host` from the name? Bonus points if this is moved to the `<Host>` element itself.

jcfranco (changes requested):
Looking great! 😎 

Could you add a test for this? Also, can you confirm whether this is low-risk or not? It seems so from my perspective.

packages/calcite-components/src/components/time-picker/time-picker.tsx
eriklharper: It needs to use the `@Listen` decorator for it to work correctly, for some reason.  When I inline the pointerdown handler on the Host element, the `focusHandler` method is getting an out-of-date value for `pointerActivated`, resulting in the flash of focus styling.  I can go into more detail if you'd like to see it in action.

packages/calcite-components/src/components/time-picker/time-picker.tsx
eriklharper: It needs to use the `@Listen` decorator for it to work correctly, for some reason.  When I inline the pointerdown handler on the Host element, the `focusHandler` method is getting an out-of-date value for `pointerActivated`, resulting in the flash of focus styling.  I can go into more detail if you'd like to see it in action.

jcfranco: Interesting. Both should behave the same. 🤔

In order to get this moving for the upcoming release, let's focus on:

* dropping `host` for consistency
* using `pointerdown` vs `mousedown`
* confirming if this is low-risk
* if time permits, adding an E2E to cover this – this one can be tackled immediately after the release if we need more time as long as we create an issue or PR for tracking purposes

Afterwards, let's look at the change in behavior from moving this listener to `Host`.

eriklharper: - [ ] dropping `host` for consistency
- [ ] using `pointerdown` vs `mousedown`
- [ ] confirming if this is low-risk
- [ ] if time permits, adding an E2E to cover this

All done!  The e2e test is comprehensive, testing all up and down nudge buttons, including milliseconds.

packages/calcite-components/src/components/time-picker/time-picker.e2e.ts
```diff
@@ -59,6 +59,118 @@ describe("calcite-time-picker", () => {
...
+      await page.setContent(`<calcite-time-picker step=".001"></calcite-time-picker>`);
+
+      const minuteEl = await page.find(`calcite-time-picker >>> .${CSS.minute}`);
+      const minuteDownEl = await page.find(`calcite-time-picker >>> .${CSS.buttonMinuteDown}`);
+
+      await minuteDownEl.click();
+      await page.waitForChanges();
+
+      expect(
+        await page.evaluate(() => document.activeElement.shadowRoot.activeElement.getAttribute("aria-label"))
```
jcfranco, line 129: You might be able to use `getFocusedElementProp` for this:

```tsx
getFocusedElementProp(page, "ariaLabel", { shadow: true });
```

packages/calcite-components/src/components/time-picker/time-picker.e2e.ts
```diff
@@ -1122,7 +1234,7 @@ describe("calcite-time-picker", () => {
       expect(fractionalSecondEl.innerHTML).toEqual("00");
     });
 
-    it("upward nudge of empty fractional second sets to 000 for step=0.001", async () => {
+    it("upward nudge of empty fractional second sets to 000 for step=0.001.", async () => {
```
jcfranco, line 1237: Is this trailing period intentional? We typically don't use periods in test names.

packages/calcite-components/src/components/time-picker/time-picker.tsx
```diff
@@ -904,7 +973,9 @@ export class TimePicker
             class={{
               [CSS.input]: true,
               [CSS.hour]: true,
+              [CSS.inputFocus]: this.activeEl && this.activeEl === this.hourEl,
```
jcfranco, line 976: I think this could be simplified to 

```tsx
this.activeEl === this.hourEl 
```

Since `hourEl` would be truthy when focus applicable.

packages/calcite-components/src/components/time-picker/time-picker.tsx
```diff
@@ -221,6 +223,8 @@ export class TimePicker
 
   @Listen("blur")
   hostBlurHandler(): void {
```
jcfranco, line 225: Would you be able to remove `host` here too since you're already updating the function?

jcfranco (approved):
Awesome!

packages/calcite-components/src/components/time-picker/time-picker.e2e.ts
eriklharper: woopsie, forgot to remove that.  I added that just to trigger a diff for the test runner, but didn't intend to commit it.

packages/calcite-components/src/components/time-picker/time-picker.tsx
eriklharper: The truthy check on `this.activeEl` is there in order to prevent a visual indicator of focus on initial render when both `this.activeEl` and `this.hourEl` are both `undefined`.

eriklharper: Removing the truthy check ends up making it look like this on initial render, which we don't want:

image

image

eriklharper: Since we set `this.activeEl` explicitly when the picker gains focus either from keyboard tabbing or pointer clicking, we'll know when we need to highlight the focused element once that property gets set to something other than `undefined`.

jcfranco: Ah, I see. Thanks for the great breakdown! Ignore my comment then. 🙊

packages/calcite-components/src/components/time-picker/time-picker.e2e.ts
eriklharper: done!

Issue #7531, opened 2023-08-15
CI: Add automation to low risk milestones as a fail safe


geospatialem:
### Summary

Add automation to our CI 🤖⚙️ supporting low risk milestones, where we should have a failsafe and only allow PRs to be merged in after review if any of the following labels are added:
- `low risk`
- `p - high`
- `p - critical`
- `regression`

### Desired Outcome

Add a failsafe 🔏, so only critical and/or low risk PRs can be merged in during our low risk milestones.

### Resources

Examples of low risk milestone naming conventions:
~- [1.6.0 - Low risk issues only](https://github.com/Esri/calcite-design-system/milestone/135)~
~- [2023 October Priorities - Low risk only](https://github.com/Esri/calcite-design-system/milestone/126)~

With our new naming of `Maintenance`, let's modify the original ask to just `Maintenance`, similar to:
- [1.6.1 Maintenance Release](https://github.com/Esri/calcite-design-system/milestone/136)

geospatialem:
Performed a test with https://github.com/Esri/calcite-design-system/pull/7611 via [this action](https://github.com/Esri/calcite-design-system/actions/runs/5988880275). 

It seems like if the Maintenance milestone is up next, it continues to the next milestone where it succeeds? To make sure the milestone dates weren't off I modified [this test Maintenance milestone](https://github.com/Esri/calcite-design-system/milestone/137) to 8/28 and our upcoming [August milestone](https://github.com/Esri/calcite-design-system/milestone/123) to 8/29. I reset both dates with the upcoming release, so they will look different when viewing this comment later.

@benelan I might be missing a step in the workflow, can you take a peek to confirm?

geospatialem:
@benelan Reassigning to September's milestone for additional R&D. Wasn't able to verify [earlier this week](https://github.com/Esri/calcite-design-system/issues/7531#issuecomment-1694568933).

geospatialem:
Verified with [action 1](https://github.com/Esri/calcite-design-system/actions/runs/6053273992/job/16428416938?pr=7649) _with no `low risk` label_ and [action 2](https://github.com/Esri/calcite-design-system/actions/runs/6053438909/job/16428932311?pr=7657) _with a `low risk` label_.

PR #7656, opened 2023-09-01
ci: make sure to exit on maitenance milestone failure

benelan:
**Related Issue:** #7531

## Summary

The action correctly determined that it was a maintenance milestone and [set the error message](https://github.com/Esri/calcite-design-system/actions/runs/5988880275/job/16244580547#step:3:39). However, it looks like [`core.setFailed()`](https://github.com/actions/toolkit/tree/main/packages/core#exit-codes) sets the exit code but doesn't actually exit. And then the else block with `process.exit(0)` overrode the exit code in the next iteration. Explicitly exiting will fix the issue.


jcfranco (approved):
🤖✋🚫

Issue #6033, opened 2022-12-14
CalciteButton: aria-expanded not being read by screenreader


phenderson00:
### Summary

Screenreader does not indicate to user that button is expandable despite aria-expanded being properly set on the `calcite-button` component.

### Actual Behavior

We are using a `calcite-popover` with a `calcite-button` as its reference element. When the button gets focus, the button label and "button" are read by the screenreader but "expandable" is not read aloud. I can see in the Dev Tools that the aria-expanded state is being correctly updated on the button.

### Expected Behavior

The expectation is that after reading the button label and "button", the screenreader will read "expandable" (or something to that effect) to indicate to the user that clicking the button will reveal more content, a `calcite-popover` in this case.

### Reproduction Sample

https://codesandbox.io/s/calcite-button-aria-expanded-not-read-by-screenreader-oh4y6s?file=/src/App.js

### Reproduction Steps

1. Turn on screenreader
2. Tab until focus arrives at button
3. Hear "Any age groups, button" be read

### Reproduction Version

0.35.0

### Relevant Info

Windows 10, Chrome 108.0.5359.99, Narrator, JAWS

### Esri team

Professional Services - Midwest Delivery Center

brittneytewks:
Removed "needs triage" since a priority is set and estimate is indicated

geospatialem:
@phenderson00 Thanks for re pining this! This request seems to be more in line of an enhancement, but in line with the native `button` functionality.

We'll take a look at an upcoming milestone to determine next steps.

Some findings below during triage/next steps:
- The native `button` functionality will announce "expand" or "collapse" to AT when `aria-expanded` is toggled and coupled with an `aria-controls` attribute associated with the content's `id`.
- A [Codepen sample](https://codepen.io/geospatialem/pen/gOBOBRK) for context

geospatialem:
Verified in `1.9.0-next.16` with JAWS and NVDA in https://codepen.io/geospatialem/pen/xxmagmG

PR #7658, opened 2023-09-01
fix(button): provides context for AT users when used as reference element for collapsible content

anveshmekala:
**Related Issue:** #6033 

## Summary

Provides context of expanded for AT users when `calcite-button` is used as reference element for collapsible content.


driskull (approved):
👍 

I wonder if this deserves its own `expanded` prop? Otherwise users would never know about setting aria-expanded and having it work as expected.

anveshmekala:
> I wonder if this deserves its own `expanded` prop? Otherwise users would never know about setting aria-expanded and having it work as expected.

Using prop seems like a good option. we can explore having an prop which needs to be set by the user when they use button as trigger for any floating/expandable UI. Defining this helps a lot of components which can be used as trigger elements. (ex: fab , action) . It will be additional work for user but they will have the benifit of consistent screen reader experience. 

Thoughts @jcfranco

jcfranco:
For the time being, I think we should skip the prop for a few reasons:

* API concerns
  * we already have `expanded` in some cases where this might fit nicely, but it doesn't translate as well w/ `open`, which is the preferred prop name.
  * we would have to add matching props for other, if not all, attributes.
* This follows existing a11y practices.
* At this point, this is mostly needed for triggers (reference elements). We could revisit this later as components are used in more a11y patterns.

packages/calcite-components/src/components/button/button.e2e.ts
```diff
@@ -644,4 +644,27 @@ describe("calcite-button", () => {
     expect(button1.textContent.length).toBeLessThan(longText.length);
     expect(button1.getAttribute("title")).toEqual(longText);
   });
+
+  it("should set aria-expanded attribute on shadowDOM element when used as trigger", async () => {
```
jcfranco, line 648: Could you test if the `accessible` helper picks this up (also checking that it fails when the ARIA attribute isn't set on the internal button)? If so, I'd suggest using that to simplify this test.

packages/calcite-components/src/utils/globalAttributes.ts
```diff
@@ -1,8 +1,8 @@
 import { createObserver } from "./observers";
 
 type AttributeObject = { [k: string]: any };
-type AllowedGlobalAttribute = "lang" | "role";
-const allowedGlobalAttributes = ["lang", "role"];
+type AllowedGlobalAttribute = "lang" | "role" | "aria-expanded";
```
jcfranco, line 4: Popover also sets `aria-controls`, so does that also need to be set here? Not sure if we need `aria-describedby` to support tooltip. cc @geospatialem

jcfranco (approved):
🎸🎸🎸🎸🎸🎸🎸🎸🎸🎸🎸🎸🎸🎸🎸🎸🎸🎸🎸🎸
🎸🤘🎸🎸🎸🤘🎸🤘🎸🎸🎸🤘🎸🎸🤘🤘🤘🎸🤘🎸
🎸🎸🤘🎸🤘🎸🎸🤘🎸🎸🎸🤘🎸🤘🎸🎸🎸🎸🤘🎸
🎸🎸🎸🤘🎸🎸🎸🤘🎸🎸🎸🤘🎸🎸🤘🤘🎸🎸🤘🎸
🎸🎸🎸🤘🎸🎸🎸🤘🎸🎸🎸🤘🎸🎸🎸🎸🤘🎸🎸🎸
🎸🎸🎸🤘🎸🎸🎸🎸🤘🤘🤘🎸🎸🤘🤘🤘🎸🎸🤘🎸
🎸🎸🎸🎸🎸🎸🎸🎸🎸🎸🎸🎸🎸🎸🎸🎸🎸🎸🎸🎸

packages/calcite-components/src/components/button/button.e2e.ts
anveshmekala: `accessible` helper is unable to pick this one up.

packages/calcite-components/src/utils/globalAttributes.ts
anveshmekala: FWIW, `aria-controls` need not be parsed on to the native button for the following reasons:
-It has minimal support across screen readers in terms of providing context as per https://a11ysupport.io/tech/aria/aria-controls_attribute. 
- For scenarios where screen reader should jump directly onto the opened element, the popover is currently doing that with or without aria-controls.

geospatialem (commented):
This will be a fantastic fix! 🌟 

Observed the `button` doesn't appear to be selectable/clickable via the `Enter` key when testing with NVDA and Firefox, but not sure what could be causing the issue. 

Here's the code tested with (taken from the original issue Codepen):
```html
<calcite-label layout="inline"
  ><span>Age groups:</span>
  <calcite-button id="test-category-filter-button" icon-end="chevron-down" width="half" label="Any age groups">
    Any age groups
  </calcite-button>
  <calcite-popover
    label="Select types"
    reference-element="test-category-filter-button"
    auto-close
    placement="bottom"
    id="category-filter-popover"
  >
    <p style="padding: 12px; margin: 0">
      <calcite-label layout="inline">
        <calcite-checkbox value="infant"></calcite-checkbox>
        infant
      </calcite-label>
      <calcite-label layout="inline">
        <calcite-checkbox value="toddler"></calcite-checkbox>
        toddler
      </calcite-label>
      <calcite-label layout="inline">
        <calcite-checkbox value="preschool"></calcite-checkbox>
        preschool
      </calcite-label>
      <calcite-label layout="inline">
        <calcite-checkbox value="school age"></calcite-checkbox>
        school age
      </calcite-label>
    </p>
  </calcite-popover>
</calcite-label>
```

packages/calcite-components/src/utils/globalAttributes.ts
anveshmekala: In VoiceOver, context is provided when user check/uncheck the checkbox before closing the popup.

anveshmekala:
> This will be a fantastic fix! 🌟
> 
> Observed the `button` doesn't appear to be selectable/clickable via the `Enter` key when testing with NVDA and Firefox, but not sure what could be causing the issue.
> 
 Is it possible that NVDA is in browse mode? Browse mode can intercept certain keys from working as expected.

geospatialem:
> > Observed the `button` doesn't appear to be selectable/clickable via the `Enter` key when testing with NVDA and Firefox, but not sure what could be causing the issue.
> 
> Is it possible that NVDA is in browse mode? Browse mode can intercept certain keys from working as expected.


That was the trick! ✨ Only occurs in Firefox, and was discovered as a similar report (which was closed) in https://github.com/nvaccess/nvda/issues/4903. More info: https://dequeuniversity.com/screenreaders/nvda-keyboard-shortcuts#nvda-browse_focus_modes

packages/calcite-components/src/utils/globalAttributes.ts
geospatialem: Windows OS + JAWS updates this week mitigated the issue from last week. 💪🏻

geospatialem (approved):
LGTM 🚀

PR #7662, opened 2023-09-01
ci: stop linting auto generated component doc

benelan:
## Summary
The doc update script is failing due to linting issues with the auto-generated component readmes. Removing the lint step.

### DIFF
--- a/.github/workflows/update-doc.yml
+++ b/.github/workflows/update-doc.yml
@@ -15,7 +15,6 @@ jobs:
         run: |
           npm install
           npx --workspace=@esri/calcite-components stencil build --docs
-          npm run --workspace=@esri/calcite-components lint:md
       - name: commit readme files/create pull request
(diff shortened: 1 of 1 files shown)


jcfranco:
Do you know what the linting error is?

benelan:
> Do you know what the linting error is?

`MD049/emphasis-style Emphasis style should be consistent [Expected: underscore; Actual: asterisk]`

I could also disable that rule but other errors may pop up. We can't guarantee that Stencil's markdown formatting will match the markdownlint rules.

jcfranco:
Could we enforce `asterisk` or `underscore` (per the [docs](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md049---emphasis-style-should-be-consistent))? I agree that auto-generated content might use mixed and would still cause failures with the default (`consistent`).

PR #7664, opened 2023-09-02
fix(sheet): Update shadow style in display-mode="float"

macandcheese:
## Summary
Fixes shadow set on incorrect container in Sheet while in `display-mode="float"`, and updates to use the same shadow variable as Modal. Was only noticed when the scrim color was overridden to a lighter color. cc @driskull cc @SkyeSeitz 

Before:
Screenshot 2023-09-01 at 6 01 13 PM

After:
Screenshot 2023-09-01 at 6 01 03 PM


jcfranco (approved):
😱👻

PR #7665, opened 2023-09-02
fix(table): Allow wrapping text in table-header and table-cell

macandcheese:
## Summary
Adds styles to allow long text in header or table cell to wrap. Adds story.


jcfranco (approved):
🏓

PR #7666, opened 2023-09-02
feat(flow): split up custom flow item interfaces

jcfranco:
**Related Issue:** #7608 

## Summary

Splits up interfaces for custom flow item class and elements. 

This also updates the conventions doc with info on interface usage.


driskull (approved):
Nice, maybe we should patch for this if they need it for implementation.

Issue #7659, opened 2023-09-01
Custom ESLint rule to disable sorting for `ref` attribute, as it needs to be the last property in JSX to ensure it is up-to-date.


Elijbet:
### Description

- This GitHub issue relates to an existing bug in the Stencil library (https://github.com/ionic-team/stencil/issues/4074). The current behavior is that the order of `ref` usage can result in outdated props/attributes when the callback is invoked.

- The expected behavior is for `ref` to be called immediately after the element is fully initialized with all props/attributes, aligning with similar mechanisms in other libraries like React and Preact.  Properties are applied in the order they are specified. If you place the `ref` at the end of the component's properties, it ensures that the component's render and updates are processed before the ref callback is triggered. This means the ref callback will have access to the fully rendered and updated component or DOM element.

- The suggested solution involves setting `ref` as the last property in JSX to ensure that node attributes/properties are up-to-date. However, this requires disabling the `jsx-sort-props` rule specifically for this case using an ESLint comment (`eslint-disable-next-line`) (#6530, #7584).

- The request is to create a custom ESLint rule that catches this pattern instead of relying on manual comments throughout the repository. 

### Proposed Advantages

Provides an automated solution to manual comments to disable sorting rules throughout the repository. 

### Which Component

All. 

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [X] @esri/eslint-plugin-calcite-components

jcfranco:
Updated labels and verified locally. 🎉

The next step for this rule will be tackled in #8080.

PR #6530, opened 2023-02-27
refactor: move ref prop last to ensure ref node is in sync

jcfranco:
**Related Issue:** N/A

## Summary

This updates `ref` usage to work around a [Stencil bug](https://github.com/ionic-team/stencil/issues/4074) where the ref node can be out of sync.

**Note**: this also updates the conventions doc with this info.

PR #7669, opened 2023-09-03
chore: custom ESLint rule to disable sorting for `ref` attribute to keep it as last property in JSX to ensure it is up-to-date

Elijbet:
**Related Issue:** #7659

## Summary
Custom ESLint rule to provide a more granular prop ordering mechanism by forcing the `ref` attribute to be kept as the last property so we can ensure it is up-to-date.

Added a test file for the new custom rules: the file references different use cases, which are 1 good and 1 bad usage files.


Elijbet:
> It doesn't look like jsx-eslint/eslint-plugin-react has kept their docs up to date. There is a prop called reservedList available for the eslint react/jsx-sort-props that should allow "ref" to be at the bottom of a multiline list of attributes.
> 
> ```json
> "react/jsx-sort-props": [
>       "error",
>       {
>         reservedList: ["ref"],
>       },
>  ],
> ```

I would rather not use an undocumented prop, it doesn't feel like a good practice.
It's possible they just forgot to document it, but it is also possible that it wasn’t documented intentionally and only exists in the source code because they don’t want it being used for whatever reason. Have we not ever done the latter before?

But I'll defer to @driskull and @jcfranco for perspective.

alisonailea:
I ran quick test on my local branch before I made the comment. However, after opening a clean branch off of Main and running some additional tests, I don’t think it’s working the way I thought.

That said, I think it would be better to open an issue and PR to eslint-jsx/eslint-react with a "reservedLast" prop (or something) rather than making (and then having to maintain) a new eslint plugin.

driskull:
> I would rather not use an undocumented prop, it doesn't feel like a good practice.

Agreed.

> I ran quick test on my local branch before I made the comment. However, after opening a clean branch off of Main and running some additional tests, I don’t think it’s working the way I thought.

`reservedList` sounds like its something that isnt allowed to be used?

> That said, I think it would be better to open an issue and PR to eslint-jsx/eslint-react with a "reservedLast" prop (or something) rather than making (and then having to maintain) a new eslint plugin.

Agreed. We should open an issue/PR but in the meantime we could do our own thing with a custom plugin. Who knows if they would add or accept the feature we are requesting.

packages/eslint-plugin-calcite-components/src/rules/enforce-ref-last-prop.ts
```diff
@@ -0,0 +1,52 @@
+import { Rule } from "eslint";
+import type { JSXAttribute, JSXSpreadAttribute, JSXOpeningElement } from "@babel/types";
+
+const rule: Rule.RuleModule = {
+  meta: {
+    docs: {
+      description: "Ensure that the ref attribute is placed as the last property in JSX elements to keep it up-to-date",
```
driskull, line 7: Maybe:

` "Ensures the ref attribute is ordered last in a JSXElement to keep it up-to-date",`?

packages/eslint-plugin-calcite-components/src/rules/enforce-ref-last-prop.ts
```diff
@@ -0,0 +1,52 @@
...
+
+          openingElement.attributes.forEach((attr: JSXAttribute | JSXSpreadAttribute) => {
+            if (attr.type === "JSXAttribute" && attr.name?.type === "JSXIdentifier") {
+              attributes.push(attr.name.name);
+            }
+          });
+
+          const refAttribute = attributes.find((attr: string) => attr === "ref");
+
+          console.log("refAttribute", refAttribute);
```
driskull, line 38: remove console log.

packages/eslint-plugin-calcite-components/src/rules/enforce-ref-last-prop.ts
```diff
@@ -0,0 +1,52 @@
...
+      description: "Ensure that the ref attribute is placed as the last property in JSX elements to keep it up-to-date",
+      category: "Best Practices",
+      recommended: true,
+    },
+    fixable: "code",
+    schema: [],
+    type: "problem",
+  },
+
+  // https://unpkg.com/browse/@babel/types@7.6.0/lib/index.d.ts
```
driskull, line 16: remove comments

packages/eslint-plugin-calcite-components/src/rules/enforce-ref-last-prop.ts
```diff
@@ -0,0 +1,52 @@
...
+          });
+
+          const refAttribute = attributes.find((attr: string) => attr === "ref");
+
+          console.log("refAttribute", refAttribute);
+
+          if (refAttribute && attributes.indexOf(refAttribute) !== attributes.length - 1) {
+            context.report({
+              node,
+              message: `The "ref" attribute should be placed as the last property in JSX elements.`,
```
driskull, line 43: `The "ref" attribute should be the last attribute in a JSXElement.`

driskull (commented):
Looks good! Maybe the wording could be cleaned up a little.

alisonailea (approved):

packages/calcite-components/.eslintrc.cjs
```diff
@@ -12,7 +12,7 @@ module.exports = {
     "plugin:jsdoc/recommended",
     "prettier",
   ],
-  ignorePatterns: ["dist", "docs", "hydrate", "www"],
+  ignorePatterns: ["dist", "docs", "hydrate", "www", ".eslintrc.cjs"],
```
jcfranco, line 15: Why is `.eslintrc.cjs` ignored here?

packages/eslint-plugin-calcite-components/README.md
```diff
@@ -35,7 +35,7 @@ Add a new `lint` script to `package.json`:
 
 Then you can run the linter:
 
-```
+```javascript
```
jcfranco, line 38: `shell` might be more appropriate since this is not JS.

packages/eslint-plugin-calcite-components/README.md
```diff
@@ -49,6 +49,10 @@ This rule helps prevent usage of specific events and allows suggesting alternati
 
 This rule catches props/attributes that should be in the encapsulated HTML structure and not on the host element.
 
+- [`@esri/calcite-components/enforce-ref-last-prop`](./docs/enforce-ref-last-prop.md)
+
+This rule ensures the `ref` attribute is ordered last in a JSXElement to keep it up-to-date.
```
jcfranco, line 54: Can you add info from https://github.com/Esri/calcite-design-system/pull/6530 to provide more context? The rule description and its doc file could also benefit from this extra info.

packages/eslint-plugin-calcite-components/src/rules/enforce-ref-last-prop.ts
```diff
@@ -0,0 +1,43 @@
...
+              attributes.push(attr.name.name);
+            }
+          });
+
+          const refAttribute = attributes.find((attr: string) => attr === "ref");
+
+          if (refAttribute && attributes.indexOf(refAttribute) !== attributes.length - 1) {
+            context.report({
+              node,
+              message: `The "ref" attribute should be the last attribute in a JSXElement.`,
```
jcfranco, line 34: Could you add more info from https://github.com/Esri/calcite-design-system/pull/7584 to the error message?

packages/eslint-plugin-calcite-components/tests/lib/rules/enforce-ref-last-prop/enforce-ref-last-prop.good.tsx
```diff
@@ -0,0 +1,22 @@
+// @ts-nocheck
+@Component({ tag: "sample-tag" })
+export class SampleTag {
+  render() {
+    return (
+      <Host>
+        <div
+          class="some-class"
```
jcfranco, line 8: Suggestion: could you also add `tabIndex` to tie in the [scenario](https://github.com/jcfranco/stencil-ref-called-in-prop-order-instead-of-after-node-being-initialized/blob/main/src/components/my-component/my-component.tsx#L12-L15) that this rule helps avoid?

jcfranco (approved):
🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️
🕶️😎😎😎😎🕶️😎🕶️🕶️🕶️🕶️🕶️😎🕶️😎😎😎😎🕶️🕶️😎😎😎🕶️🕶️😎😎🕶️🕶️😎🕶️🕶️🕶️😎🕶️😎😎😎😎🕶️😎🕶️
🕶️😎🕶️🕶️😎🕶️😎🕶️🕶️🕶️🕶️🕶️😎🕶️😎🕶️🕶️🕶️🕶️😎🕶️🕶️🕶️🕶️😎🕶️🕶️😎🕶️😎😎🕶️😎😎🕶️😎🕶️🕶️🕶️🕶️😎🕶️
🕶️😎😎😎😎🕶️😎🕶️🕶️😎🕶️🕶️😎🕶️😎😎😎🕶️🕶️🕶️😎😎🕶️🕶️😎🕶️🕶️😎🕶️😎🕶️😎🕶️😎🕶️😎😎😎🕶️🕶️😎🕶️
🕶️😎🕶️🕶️😎🕶️😎🕶️😎🕶️😎🕶️😎🕶️😎🕶️🕶️🕶️🕶️🕶️🕶️🕶️😎🕶️😎🕶️🕶️😎🕶️😎🕶️🕶️🕶️😎🕶️😎🕶️🕶️🕶️🕶️🕶️🕶️
🕶️😎🕶️🕶️😎🕶️🕶️😎🕶️🕶️🕶️😎🕶️🕶️😎😎😎😎🕶️😎😎😎🕶️🕶️🕶️😎😎🕶️🕶️😎🕶️🕶️🕶️😎🕶️😎😎😎😎🕶️😎🕶️
🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️🕶️

packages/eslint-plugin-calcite-components/tests/lib/rules/enforce-ref-last-prop/enforce-ref-last-prop.good.tsx
Elijbet: That's a good point! ☉

packages/eslint-plugin-calcite-components/tests/lib/rules/enforce-ref-last-prop/enforce-ref-last-prop.wrong.tsx
```diff
@@ -0,0 +1,22 @@
...
+      <Host>
+        <div
+          ref={(el: HTMLDivElement): void => {
+            /* refEl */
+          }}
+          class="some-class"
+          id={`${guid}-element`}
+          onClick={() => {
+            /* click! */
+          }}
```
jcfranco, line 15: `tabIndex` should go here to match the good case.

packages/eslint-plugin-calcite-components/README.md
```diff
@@ -49,6 +49,10 @@ This rule helps prevent usage of specific events and allows suggesting alternati
 
 This rule catches props/attributes that should be in the encapsulated HTML structure and not on the host element.
 
+- [`@esri/calcite-components/enforce-ref-last-prop`](./docs/enforce-ref-last-prop.md)
+
+This updates `ref` usage to work around a Stencil bug where ref is called in the specified order and not after initializing element with all its attributes/properties. This can cause attributes/properties to be outdated by the time the callback is invoked. This rule ensures the `ref` attribute is ordered last in a JSXElement to keep it up-to-date.
```
jcfranco, line 54: The `this updates ref usage...` part doesn't really apply here because the rule isn't changing the code. I think it'll read better if we flip it a bit where we describe what the rule does `This rule ensures...` and then follow it with
 
`This updates ref usage to workaround is a workaround for a Stencil bug (<add link to bug*>) where ref is...`

jcfranco (approved):
This should be good to merge once the last comments are addressed. 🚀

PR #7671, opened 2023-09-05
fix: resolve vite errors due to using expressions in dynamic imports

benelan:
**Related Issue:** #7575

## Summary

Vite uses [`@rollup/plugin-dynamic-import-vars`](https://github.com/rollup/plugins/tree/master/packages/dynamic-import-vars) for resolving dynamic imports that use an expression for the path, rather than a fixed string. The rollup plugin has a [limitation](https://github.com/rollup/plugins/tree/master/packages/dynamic-import-vars#imports-must-start-with--or-) that requires dynamic imports with expressions to use relative paths. 

> All imports must start relative to the importing file. The import should not start with a variable, an absolute path or a bare import

Changing the module path to a fixed string in the patch script resolves the errors. It also ensures that only the components being used in the app are bundled.


jcfranco (approved):
Awesome! 🚀👨‍🚀🛰️🪐

~Can you update the issue description with the latest findings before merging?~ Never mind, I misread the plugin part.

Issue #7599, opened 2023-08-24
[CalciteList] Dropdowns at action-end are cut off in Safari if overflow-y: auto is set on the list


nikki-lea:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When a dropdown is placed in a CalciteList with overflow-y: auto, the DropdownList will be truncated with no way to access the dropdown.

<img width="350" alt="Screenshot 2023-08-24 at 9 39 39 AM" src="https://github.com/Esri/calcite-design-system/assets/61667766/8faf265c-471d-442b-9932-80f9c191d52b">

### Expected Behavior

Expect to see the dropdown display and not be truncated

### Reproduction Sample

https://codesandbox.io/s/peaceful-worker-6clsy5?file=/src/App.js

### Reproduction Steps

1. Open with safari
2. Click the ellipses on the item2 list item
3. Dropdown does not display fully

### Reproduction Version

1.4.2

### Relevant Info

Safari only

### Priority impact

p4 - not time sensitive

### Impact

Release blocking, workaround provided

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Enterprise

geospatialem:
Spike to confirm the use of z-index mentioned in [Franco's comment above](https://github.com/Esri/calcite-design-system/issues/7599#issuecomment-1692175316).

driskull:
@geospatialem @jcfranco it was added in the big list enhancements PR for filtering and treegrid. https://github.com/Esri/calcite-design-system/commit/f489c57095ec21df1f427176d2d635675eea95d3

I think we can remove it. I don't see any reason it is needed.

geospatialem:
@driskull Awesome news! If this seems like a low risk fix, we could add into inclusion for the upcoming maintenance release.

driskull:
Its tough to determine the risk here. Ill defer to @jcfranco.

jcfranco:
If we can confirm there's no visual regression in our test suite, we can consider this low risk since it's removing an unnecessary stacking context that might stir up trouble in similar configurations.

driskull:
Yeah no regressions found in screenshot tests

geospatialem:
Verified in `1.8.0` with https://codepen.io/geospatialem/pen/rNojOxe.

PR #7678, opened 2023-09-05
fix(list): remove unnecessary z-index

driskull:
**Related Issue:** #7599

## Summary

- removes z-index on a container within the list component

### DIFF
--- a/packages/calcite-components/src/components/list/list.scss
+++ b/packages/calcite-components/src/components/list/list.scss
@@ -13,9 +13,7 @@
   flex
   w-full
   flex-col
-  bg-transparent
-  relative
-  z-default;
+  bg-transparent;
   * {
     @apply box-border;
   }
--- a/packages/calcite-components/src/components/list-item/list-item.scss
+++ b/packages/calcite-components/src/components/list-item/list-item.scss
@@ -55,7 +55,7 @@ td {
 
 tr:focus,
 td:focus {
-  @apply focus-inset z-sticky;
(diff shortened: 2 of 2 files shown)


packages/calcite-components/src/components/list/list.scss
```diff
@@ -13,9 +13,7 @@
   flex
   w-full
   flex-col
-  bg-transparent
-  relative
-  z-default;
+  bg-transparent;
```
macandcheese, line 16: We might still need to bump the z-index of the `filter-enabled` UI elements - with this change it is behind the focused list element:

https://github.com/Esri/calcite-design-system/assets/4733155/b57ad989-531a-410b-868d-a20c2b5b1109


https://github.com/Esri/calcite-design-system/assets/4733155/07f6b932-7e9e-4cc8-ba90-e7001f094b53

driskull: Nice catch @macandcheese. I removed the z-index on the active element. Not sure why we had it there. @jcfranco any idea?

jcfranco: I don't recall. Nice cleanup though! 🧼

jcfranco (approved):
Low-risk approved! 🥇

driskull:
@macandcheese can you test this one again? Por favor 🙇

macandcheese (approved):
Looks good with this change!

Issue #7529, opened 2023-08-15
[ArcGIS Online] [META] I18N-FCN-AR [FF/Safari]: Date is cleared after selecting a date and clicking on a blank area on Metadata page.


Olga8614:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Date is cleared after selecting a date and clicking on a blank area on Metadata page.
Please refer the attachment "data_cant_be_selected.gif" for details.

### Expected Behavior

Date should not be cleared after selecting a date and clicking on a blank area on Metadata page.

### Reproduction Sample

- https://codepen.io/rpanichakit/pen/OJrLMRX?editors=100

- https://github.com/ArcGIS/I18N-BYS-TestCases/issues/11650

### Reproduction Steps

Codepen Steps:
1. Open the [Codepen](https://codepen.io/rpanichakit/pen/OJrLMRX?editors=100) in Firefox
2. Click calendar icon and select a date.
3. Click other blank area.
4. Observe.

DevExt Steps:
1. Prepare a clean AR FF browser.
3. Launch URL https://devext.arcgis.com and sign in with valid account.
4. Go to My content page and open item details page of a feature layer.
5. Click Metadata button.
6. Click "All metadata" tab, expand Metadata Info->Metadata Details.
7. Scroll to Metadata Data section.
8. Click calendar icon and select a date.
9. Click other blank area.
10. Observe.

### Reproduction Version

1.5.1

### Relevant Info

Additional Information:
1. This behavior also occurs with all other date fields.
2. This behavior also occurs with Safari (16.6) browser.
3. This behavior does not occur with Edge Chromium (115.0.1901.200)/Chrome (115.0.5790.171) browsers.
4. This behavior does not occur on AGOL 2023.r02 production site.

Configuration:
1. Build: AGOL 2023.r03 devext site
2. OS: AR Win10 64bit
3. Browser: Firefox (116.0.2)
4. Found by date: 8/11/2023

Build Number : 8/11/2023

### Regression?

N/A

### Priority impact

p3 - want for upcoming milestone

### Impact

N/A

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Online

Olga8614:
https://github.com/ArcGIS/I18N-BYS-Bugs/issues/16589

Olga8614:
[...]

PR #7682, opened 2023-09-06
fix(input-date-picker): apply default numbering system to avoid browser inferring from locale

jcfranco:
**Related Issue:** #7529 

## Summary

This applies our default numbering system to `input-date-picker`'s date formatting.


benelan (approved):
:+1:

Issue #7676, opened 2023-09-05
List: <Enter> keypress behavior within list-items


jrowlingson:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Summary

I'm unsure if this is the intended behavior for [`<calcite-list-item>`](https://developers.arcgis.com/calcite-design-system/components/list/), but a cursory look through the issues did not yield anything similar.

I see that an `<Enter>` keypress will open the `href=...` of both a `<calcite-link>` and `<calcite-button>` (ex. 1,2), however, when these are slotted into `actions-end` of `<calcite-list-item>` this no longer works (ex. 3,4). Given the naming of the slot, I am assuming a `<calcite-action>` is expected here, however, this too seems to produce a similar behavior (ex. 6). Note, `onKeydown` does work for a slotted `<calcite-action>`.

### Actual Behavior

1. Focus a `actions-end` slotted `<calcite-link href=...>`, `<calcite-button href=...>` (ex. 3,4), or `<calcite-action onClick=...>` (ex. 6). 
2. Press `<Enter>`
3. The browser navigates to the href if its an anchor, or calls the onClick handler if its an action.

### Expected Behavior

1. Focus a `actions-end` slotted `<calcite-link href=...>`, `<calcite-button href=...>` (ex. 3,4), or `<calcite-action onClick=...>` (ex. 6). 
2. Press `<Enter>`
3. The browser does not navigate to the href if its an anchor, or call the onClick handler if its an action.

### Reproduction Sample

https://codepen.io/jrowls/pen/OJrbRqj?editors=1000

### Reproduction Steps

See codepen sample

### Reproduction Version

1.7.0

### Relevant Info

Firefox 117.0

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Field Apps

geospatialem:
Verified in `1.8.0`.

PR #7684, opened 2023-09-06
fix(list-item): Do not call preventDefault on enter key within slotted actions

driskull:
**Related Issue:** #7676

## Summary

- Enter key on any of the actions containers will not do anything
- So enter key will no longer be prevented
- Adds test

### DIFF
--- a/packages/calcite-components/src/components/list-item/list-item.tsx
+++ b/packages/calcite-components/src/components/list-item/list-item.tsx
@@ -703,7 +703,11 @@ export class ListItem
     const cells = [actionsStartEl, contentEl, actionsEndEl].filter(Boolean);
     const currentIndex = cells.findIndex((cell) => composedPath.includes(cell));
 
-    if (key === "Enter") {
+    if (
+      key === "Enter" &&
+      !composedPath.includes(actionsStartEl) &&
+      !composedPath.includes(actionsEndEl)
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -703,7 +703,11 @@ export class ListItem
     const cells = [actionsStartEl, contentEl, actionsEndEl].filter(Boolean);
     const currentIndex = cells.findIndex((cell) => composedPath.includes(cell));
 
-    if (key === "Enter") {
+    if (
+      key === "Enter" &&
+      !composedPath.includes(actionsStartEl) &&
```
jcfranco, line 708: Nice!

Sidebar: it would be useful to do an audit to see if other components might need this too. cc @geospatialem

jcfranco (approved):
👇💥🚫

packages/calcite-components/src/components/list-item/list-item.tsx
geospatialem: Added https://github.com/Esri/calcite-design-system/issues/7686 for a spike in next month's milestone.

jcfranco:
[Belated] low-risk confirmed!

Issue #7687, opened 2023-09-06
Action-group no longer honoring `flex-grow`


nel11211:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

`calcite-action-group` flex-grow is not being honored anymore.

Before:
![image](https://github.com/Esri/calcite-design-system/assets/66378014/a17ebae2-5eef-45db-bddd-3fe72400ab2c)

After:
![image](https://github.com/Esri/calcite-design-system/assets/66378014/7af46c96-24a3-48ac-a876-18d9cbad9b5d)

### Expected Behavior

The "Before" image above is the expected behavior.

### Reproduction Sample

https://codepen.io/nel11211/pen/wvRoZQd

### Reproduction Steps

See codepen for one example of a setup that yields the issue.

### Reproduction Version

1.7.0

### Regression?

1.6.0

### Priority impact

p4 - not time sensitive

### Impact

Low impact since we were able to resolve our issue by switching to using a horizontal `calcite-action-bar` instead of the `calcite-action-group`. But it is a bug with the action-group nonetheless.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Charts

driskull:
@geospatialem can we add this to the current release?

jcfranco:
Added to the upcoming maintenance release.

geospatialem:
Verified in `1.8.0`

![image](https://github.com/Esri/calcite-design-system/assets/5023024/7d189501-17a6-4260-8511-672df444b85e)

PR #7690, opened 2023-09-06
fix(action-group): Honor flex-grow on slotted actions

driskull:
**Related Issue:** #7687

## Summary

- Add flex grow to container
- Add screenshot test

### DIFF
--- a/packages/calcite-components/src/components/action-group/action-group.scss
+++ b/packages/calcite-components/src/components/action-group/action-group.scss
@@ -17,7 +17,7 @@
 }
 
 .container {
-  @apply flex flex-col;
+  @apply flex flex-col flex-grow;
 }
 
 :host([columns="1"]) {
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/action-group/action-group.scss
```diff
@@ -17,7 +17,7 @@
 }
 
 .container {
-  @apply flex flex-col;
+  @apply flex flex-col flex-grow;
```
jcfranco, line 20: Sidebar: WDYT about doing an audit and possibly defining guidelines on expected styles on the root internal container? cc @macandcheese @geospatialem

jcfranco (approved):
✨🌱➡️🌴✨

jcfranco:
Belated, low-risk confirmed! 😅

Issue #5602, opened 2022-10-25
Stepper and Stepper Item - Ensure custom controls provide proper textual name, role, and state information - (2036570226)


dqateam:
### Module: 
 31 Stepper and Stepper Item 

### Violation: 
 Ensure custom controls provide proper textual name, role, and state information 

![image](https://user-images.githubusercontent.com/5023024/202740101-5bef66c5-e407-467f-9a4f-66b1c741d3f1.png)

### WCAG Reference: 
- [1.1.1 Non-text Content](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html)
- [4.1.2 Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html) 

### Instance ID: 
 2036570226 

### Severity: 
 7 

### Description: 
 [Issue]
The current step information is not communicated to JAWS users

[User Impact]
Screen reader users will have difficulty determining the purpose and state of these controls.

[Code Reference] 
 ```
<calcite-stepper-item item-title="Error property" error="" aria-expanded="true" layout="horizontal" scale="m" tabindex="0" calcite-hydrated="" active="">Step 2 contains an "error" property.</calcite-stepper-item>
 ``` 

### Note: 
 [Suggestion]
Ensure custom controls provide proper textual name, role, and state information. 

A control&#039;s state indicates several possible pieces of information. Common states include expanded, current, selected, checked, and disabled. Most are indicated with similarly named ARIA attributes. Certain states can only be used with certain roles. For a list of roles and their allowed states, see the ARIA specification: https://www.w3.org/TR/wai-aria/#role_definitions

A proper ordered list can also be used with a list item that includes every step and it&#039;s respective text. Visually hidden text can be used to indicate the current and completed steps. 

[Compliant Code Example] 
 ```
<calcite-stepper-item error="" aria-expanded="true" layout="horizontal" scale="m" tabindex="0" calcite-hydrated="" active="">
<span class="visuallyhidden">Current: </span>
Step 2 contains an "error" property.
</calcite-stepper-item>


 ```
[Recommended Reading]
More information on how to indicate progress, reference approach 4 in "Multi-page forms" best practices:
https://www.w3.org/WAI/tutorials/forms/multi-page/

More information about the correct use of ARIA states, roles and properties can be found on the Level Access website:

https://www.levelaccess.com/how-not-to-misuse-aria-states-properties-and-roles/ 

### Media Type: 
 Custom Controls 

[...]

PR #7691, opened 2023-09-06
fix(stepper): improves AT Users experience with screen readers

anveshmekala:
**Related Issue:** #5602

### DIFF
--- a/packages/calcite-components/src/components/stepper-item/stepper-item.tsx
+++ b/packages/calcite-components/src/components/stepper-item/stepper-item.tsx
@@ -12,7 +12,6 @@ import {
   VNode,
   Watch,
 } from "@stencil/core";
-import { toAriaBoolean } from "../../utils/dom";
 import { Layout, Scale } from "../interfaces";
 import {
   connectInteractive,
@@ -210,11 +209,12 @@ export class StepperItem implements InteractiveComponent, LocalizedComponent, Lo
   render(): VNode {
     return (
       <Host
-        aria-expanded={toAriaBoolean(this.selected)}
+        aria-current={this.selected ? "step" : "false"}
         onClick={this.handleItemClick}
         onKeyDown={this.keyDownHandler}
       >
         <div class="container">
+          {this.complete && <span class="visually-hidden">{" Completed Step"}</span>}
           <div
             class="stepper-item-header"
             tabIndex={
--- a/packages/calcite-components/src/components/stepper/stepper.tsx
+++ b/packages/calcite-components/src/components/stepper/stepper.tsx
@@ -4,6 +4,7 @@ import {
   Event,
   EventEmitter,
   h,
+  Host,
   Listen,
   Method,
   Prop,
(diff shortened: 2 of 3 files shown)


anveshmekala:
In terms of testing this fix will provide context regarding `current` step user one and the status of step ( complete). Tested with NVDA in Chrome and VoiceOver in safari.

geospatialem (commented):
The current and completed step sounds great in both NVDA and JAWS announcing if the step is current (not on selection which is out of scope for AT support) and when the step is complete.

packages/calcite-components/src/components/stepper-item/stepper-item.tsx
```diff
@@ -210,11 +209,16 @@ export class StepperItem implements InteractiveComponent, LocalizedComponent, Lo
...
       <Host
-        aria-expanded={toAriaBoolean(this.selected)}
+        aria-current={this.selected ? "step" : "false"}
         onClick={this.handleItemClick}
         onKeyDown={this.keyDownHandler}
       >
         <div class="container">
+          {this.complete && (
+            <span aria-live="polite" class="visually-hidden">
+              {"Completed step"}
```
driskull, line 219: should reference message here.

packages/calcite-components/src/components/stepper-item/stepper-item.tsx
```diff
@@ -210,11 +209,16 @@ export class StepperItem implements InteractiveComponent, LocalizedComponent, Lo
...
     return (
       <Host
-        aria-expanded={toAriaBoolean(this.selected)}
+        aria-current={this.selected ? "step" : "false"}
         onClick={this.handleItemClick}
         onKeyDown={this.keyDownHandler}
       >
         <div class="container">
+          {this.complete && (
+            <span aria-live="polite" class="visually-hidden">
```
driskull, line 218: can you put the class in a CSS const?

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -112,17 +113,19 @@ export class Stepper {
...
-          const items = (event.currentTarget as HTMLSlotElement)
-            .assignedElements()
-            .filter((el) => el?.tagName === "CALCITE-STEPPER-ITEM");
-          const spacing = Array(items.length).fill("1fr").join(" ");
-          this.el.style.gridTemplateAreas = spacing;
-          this.el.style.gridTemplateColumns = spacing;
-          this.setStepperItemNumberingSystem();
-        }}
-      />
+      <Host aria-label={"Progress steps"} role="region">
```
driskull, line 116: aria-label should reference message

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -112,17 +113,19 @@ export class Stepper {
...
+        <slot
+          onSlotchange={(event: Event) => {
+            const items = (event.currentTarget as HTMLSlotElement)
```
driskull, line 119: We have a dom util helper for getting the assigned elements.

driskull (changes requested):
Looking good!

packages/calcite-components/src/components/stepper-item/stepper-item.tsx
anveshmekala: Without `T9nComponent` interface implemented, we cant test the message bundles for all supported locales. Once the translations are back this will be replaced.

packages/calcite-components/src/components/stepper-item/stepper-item.tsx
anveshmekala: updated.

packages/calcite-components/src/components/stepper/stepper.tsx
anveshmekala: Good catch, updated this.

packages/calcite-components/src/components/stepper-item/stepper-item.tsx
anveshmekala: added #7759.

driskull: This should reference the messages now right?

packages/calcite-components/src/components/stepper/stepper.tsx
driskull: This should reference the messages now right?

driskull (approved):
👍

anveshmekala:
Nope, the strings wont reference the messages because the translations bundles are available yet. Once the translation are back , we can reference the strings from messages. The reason for dependency on translated strings is related to testing util. Translation part will be addressed in #7759.

geospatialem (commented):
The current and completed step sounds great in both NVDA and JAWS announcing if the step is current (not on selection which is out of scope for AT support) and when the step is complete. 👍🏻

Issue #6814, opened 2023-04-19
[Combobox] filtering does not work when Item text labels are not unique


mjuniper:
### Actual Behavior

When calcite-combobox contains multiple calcite-combobox-items with the same text-label, filtering breaks.

### Expected Behavior

I expect filtering to work.

### Reproduction Sample

https://codepen.io/mjuniper/pen/JjmRQpV?editors=1000

### Reproduction Steps

1. Click in the combobox
2. type "conf"

Expected: the combobox groups/items would be filtered to just the ones in the "arcgis-configuration-editor" group.

Observed: The items are not filtered.

### Reproduction Version

1.2.0

### Relevant Info

[This issue](https://github.com/Esri/calcite-components/issues/4348) seems similar but it is closed.

### Priority impact

p4 - not time sensitive

### Esri team

ArcGIS Hub

geospatialem:
Verified in `1.8.1-next.0`
![verify-filter](https://github.com/Esri/calcite-design-system/assets/5023024/1d09b644-6433-4189-938d-ec0cb2a7b874)

PR #7704, opened 2023-09-08
fix(combobox): fix filtering to avoid showing unrelated items

jcfranco:
**Related Issue:** #6814 

## Summary

This fixes an issue with the logic matching items to the internal filtered data. Filtered data corresponds to items, so both `label` and `value` need to match vs one or the other.

Also, hidden groups are visually hidden now. They were being displayed because the host was missing the base `hidden` styles.


packages/calcite-components/src/components/combobox-item-group/combobox-item-group.scss
```diff
@@ -53,3 +53,5 @@
 ::slotted(calcite-combobox-item-group:not([after-empty-group])) {
   padding-block-start: var(--calcite-combobox-item-spacing-unit-l);
 }
+
+@include base-component();
```
driskull, line 57: nice catch.

driskull (approved):
👍

Issue #431, opened 2020-04-02
Dropdown contents should not be dependent on the parent element


dhrumil83:
Enhancement to this fix: https://github.com/Esri/calcite-components/issues/396
Codepen: https://codepen.io/dhrumil83/pen/poJBbEW

Can end up with 2 scrollbars when parent element to the dropdown has a height specified.

![image](https://user-images.githubusercontent.com/14285436/78198450-7dc34e00-743d-11ea-99a3-eb077f2058fe.png)

Behavior should be like the regular select, where the scrolling is independent of the parent element.
![image](https://user-images.githubusercontent.com/14285436/78198599-ead6e380-743d-11ea-9176-d78a1bbc93f0.png)

Potential fix can be to open the dropdown contents in a popover/popper?
Short term fix can be to let the client specify the max allowed height.

jcfranco:
I submitted a PR to allow users to specify the max items to show before we show the scroller.

PR #7705, opened 2023-09-08
feat(input-time-zone): add max-items support

jcfranco:
**Related Issue:** #7738

## Summary

Adds `maxItems` prop to allow users to define the max items to show before displaying a scrollbar. 

**Note**: this delegates max-items behavior to the internal combobox.


packages/calcite-components/src/components/input-time-zone/input-time-zone.tsx
```diff
@@ -301,6 +304,7 @@ export class InputTimeZone
           disabled={this.disabled}
           label={this.messages.chooseTimeZone}
           lang={this.effectiveLocale}
+          maxItems={this.maxItems}
```
alisonailea, line 307: how will this prop be used? What happens when more items exist than the max allowed items? Will items simply not render or will the customer receive some kind of warning/alert that they have exceeded the max items? Will this be included in a separate PR?

alisonailea (approved):

packages/calcite-components/src/components/input-time-zone/input-time-zone.tsx
jcfranco: Good questions. To summarize, this prop basically defines the amount of items displayed before a scrollbar is displayed.

packages/calcite-components/src/components/input-time-zone/input-time-zone.tsx
```diff
@@ -80,6 +80,9 @@ export class InputTimeZone
    */
   @Prop({ reflect: true }) form: string;
 
+  /** Specifies the maximum number of input time zone options to display before displaying a scrollbar. */
```
geospatialem, line 83: WDYT about the following for doc consistency?:

> Specifies the component's maximum number of options to display before displaying a scrollbar.

driskull (approved):
This looks good 👍 

Do we know what the purpose of maxItems is? Why are users limiting the items displayed? If this is about being more responsive maybe we can solve the problem without them having to limit items.

jcfranco:
> Do we know what the purpose of maxItems is? Why are users limiting the items displayed? If this is about being more responsive maybe we can solve the problem without them having to limit items.

It was introduced to address https://github.com/Esri/calcite-design-system/issues/431.

PR #7709, opened 2023-09-08
pagination container query proto

driskull:
**Related Issue:** #

## Summary

Sep-08-2023 11-36-14


alisonailea:
Assuming there are more than 4 pages, I'd like to always see 5 character slots displayed. something like....

<  1   2   3   ...   12  >
<  1   ...  4   ...   12  >
<  1   ...  10  11  12  >

PR #7720, opened 2023-09-09
fix(dropdown, dropdown-group, dropdown-item): enhance AT support

alisonailea:
**Related Issue:** #4866

## Summary
Update aria- attributes on the light DOM to provide context to dropdown item selections when closing the menu for screen reader users


packages/calcite-components/src/components/dropdown-group/dropdown-group.tsx
```diff
@@ -92,7 +95,14 @@ export class DropdownGroup {
       this.groupPosition > 0 ? <div class="dropdown-separator" role="separator" /> : null;
 
     return (
-      <Host aria-label={this.groupTitle} role="group">
+      <Host
+        {...{
+          "aria-label": this.groupTitle,
+          role: "menu",
+          id: "demo-menu-list",
```
driskull, line 102: This id should already be set on connectedCallback correct? I don't think it needs to be set here. Its also not unique in this instance.

packages/calcite-components/src/components/dropdown-group/dropdown-group.tsx
```diff
@@ -74,6 +76,7 @@ export class DropdownGroup {
   //--------------------------------------------------------------------------
 
   connectedCallback(): void {
+    this.el.setAttribute("id", this.el.getAttribute("id") || `dropdown-group-${guid()}`);
```
driskull, line 79: Can `dropdown-group-${guid()}` be created on the instance instead of in connectedCallback? otherwise, this guid will change if the component is moved in the DOM and connectedCallback is called again.

driskull: This applies to other components in this PR as well.

packages/calcite-components/src/components/dropdown/dropdown.tsx
```diff
@@ -380,32 +372,65 @@ export class Dropdown
...
       case "End":
-        focusElementInGroup(this.items, target, "last");
+        focusTarget = focusElementInGroup(this.items, target, "last");
         break;
     }
 
+    if (focusTarget) {
+      this.triggers.forEach((trigger) => {
+        trigger.setAttribute("aria-activedescendant", focusTarget.id);
+      });
```
driskull, line 399: nitpick cleanup:

```ts
this.triggers.forEach((trigger) => (trigger.setAttribute("aria-activedescendant", focusTarget.id)));
```

packages/calcite-components/src/components/dropdown/dropdown.tsx
```diff
@@ -380,32 +372,65 @@ export class Dropdown
...
+    const { selectionMode } = event.detail.requestedDropdownItem;
+    const selectionArray = [
+      event.detail.requestedDropdownGroup.id,
+      event.detail.requestedDropdownItem.id,
+    ];
+    let newSelection = [];
+
     this.updateSelectedItems();
+
+    this.triggers.forEach((trigger) => {
```
driskull, line 416: nitpick: maybe this logic can be moved into a function so its apparent what its goal is?

Something like `private _setSelectionAria()`? I think thats what its doing right?

packages/calcite-components/src/components/dropdown/dropdown.tsx
```diff
@@ -606,7 +649,43 @@ export class Dropdown
   };
 
   private updateSelectedItems(): void {
+    const selectedLabels: string[] | string[][] = [[], []];
     this.selectedItems = this.items.filter((item) => item.selected);
+    this.label = this.selectedItems
+      .reduce((acc, item, idx) => {
+        // @ts-expect-error TODO: write special type for selectedItems which includes the parentElement as a DropDownGroup
+        const group = item.parentElement.groupTitle;
```
driskull, line 657: maybe this should use `item.closes("calcite-dropdown-group")?.groupTitle`?

driskull (commented):
Code looks good @alisonailea. 👏 💯 

I think we should do a little research and get some clarity on if roles and aria-attributes can span across shadow DOMs. I know we do this in other places and it hasn't come up as an issue. For example, in `list-item` the role is a `grid-cell` inside the shadowDOM and `list` has a shadowed role of `treegrid`.

If its possible, i'd prefer to have the roles and tabindex itnernal to the component so an end user can't modify it and mess things up. That goes for the labels and id's as well. We wouldn't need to make the id's unique either.

I'll defer to Franco for his approval 👍

geospatialem (commented):
Awesome @alisonailea! 🎉 

Concur with Matt's thoughts on `aria`-attributes across the shadow DOM. We've had success across AT while leveraging shadow DOM in some cases, so wondering if we could do the same with `dropdown`-named components as well.

As far as Windows AT support, JAWS 🦈 sounds good, but NVDA 🟣 doesn't provide additional context, so we might need to do some additional testing to support both.

With the PR name, could you add context to the `dropdown`-named components to be reflected in the [changelog](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/CHANGELOG.md) so the change provides more context to folks? For additional details, take a peek at the [contributing.md](https://github.com/Esri/calcite-design-system/blob/main/CONTRIBUTING.md#pull-requests) and [recent commits](https://github.com/Esri/calcite-design-system/commits/main). Perhaps something like:

`fix(dropdown, dropdown-group, dropdown-item): enhance AT support`

alisonailea:
Thanks Matt!

> I think we should do a little research and get some clarity on if roles and aria-attributes can span across shadow DOMs. I know we do this in other places and it hasn't come up as an issue. For example, in `list-item` the role is a `grid-cell` inside the shadowDOM and `list` has a shadowed role of `treegrid`.

The problem is primarily between elements that must be semantically connected via ID (like aria-activedescendant). These elements must both be in the light-dom because otherwise shadow-dom encapsulation blocks access. This is a good article on the topic https://alice.pages.igalia.com/blog/how-shadow-dom-and-accessibility-are-in-conflict/

> If its possible, i'd prefer to have the roles and tabindex itnernal to the component so an end user can't modify it and mess things up. That goes for the labels and id's as well. We wouldn't need to make the id's unique either.

Breaking w3c standards for fear of customer error is not sustainable. If we are seeing a regular pattern of customers changing roles and tabindexes then complaining that their app is not working correctly we should update the documentation or offer a workshop on building accessible applications to our customers.

packages/calcite-components/src/components/dropdown-group/dropdown-group.tsx
alisonailea: Thanks for the catch!

packages/calcite-components/src/components/dropdown/dropdown.e2e.ts
```diff
@@ -1148,35 +1147,53 @@ describe("calcite-dropdown", () => {
       ).toBe(false);
     });
 
-    it("dropdown wrapper should have height when filter results empty and combined with a PickList in Panel  #3048", async () => {
+    it("Dropdown wrapper should not extend the height of the Picklist filter input when filter results are empty in Panel  #3048", async () => {
```
alisonailea, line 1150: This test was failing but after a conversation with @anveshmekala I updated the test to check for the original use case. Dropdown wrapper now only has a height when the Dropdown is open.

packages/calcite-components/src/components/dropdown/dropdown.e2e.ts
```diff
@@ -1060,7 +1060,7 @@ describe("calcite-dropdown", () => {
       expect(item7).not.toHaveAttribute("aria-checked");
     });
 
-    it("item selection should work when placed inside shadow DOM (#992)", async () => {
+    it.skip("item selection should work when placed inside shadow DOM (#992)", async () => {
```
alisonailea, line 1063: Currently debugging this test with @jcfranco. However, I was able to "un-skip" two other tests with the code changes I made.

alisonailea:
The e2e error is coming from axe-core which is out of date with current ARIA recommended patterns and only when tests are not run locally.

packages/calcite-components/src/components/dropdown/dropdown.tsx
```diff
@@ -380,40 +372,57 @@ export class Dropdown
   calciteInternalDropdownItemKeyEvent(event: CustomEvent<ItemKeyboardEvent>): void {
     const { keyboardEvent } = event.detail;
     const target = keyboardEvent.target as HTMLCalciteDropdownItemElement;
+    let focusTarget;
```
jcfranco, line 375: Can you type this?

packages/calcite-components/src/components/dropdown/dropdown.e2e.ts
```diff
@@ -1179,35 +1178,53 @@ describe("calcite-dropdown", () => {
...
+                    <calcite-dropdown-item>Type</calcite-dropdown-item>
+                  </calcite-dropdown-group>
+                </calcite-dropdown>
+                <calcite-pick-list-item label="calcite" description="calcite!"> </calcite-pick-list-item>
+                <calcite-pick-list-item label="calcite" description="calcite"> </calcite-pick-list-item>
+              </calcite-pick-list>
+            </calcite-panel>
+          </div>`,
       });
+      const picklistInput = await page.find("calcite-pick-list >>> calcite-filter >>> calcite-input");
```
jcfranco, line 1215: Nitpick: `pickListInput`

packages/calcite-components/src/components/dropdown/dropdown.tsx
```diff
@@ -498,9 +553,13 @@ export class Dropdown
...
       .assignedElements({ flatten: true })
       .filter((el) => el?.matches("calcite-dropdown-group")) as HTMLCalciteDropdownGroupElement[];
-
+    const groupIds = groups.map((group) => group.id);
     this.groups = groups;
 
+    this.triggers.forEach((trigger) => {
+      trigger.setAttribute("aria-controls", groupIds.join(" "));
+    });
+
```
driskull, line 562: nitpick:

```suggestion
this.triggers.forEach((trigger) => (trigger.setAttribute("aria-controls", groupIds.join(" "))));
```

packages/calcite-components/src/components/dropdown/dropdown.e2e.ts
jcfranco: For posterity, the issue was the dropdown not being fully initialized by the time the item was clicked, so it was causing some event handlers to fire that assumed the dropdown being ready.

driskull (approved):
Looks good to me if @geospatialem approves 👍 🎉

jcfranco (commented):
Looking good. I had some comments regarding the current approach and whether we could simplify a few things by making some adjustments.

jcfranco (changes requested):
I don't think the skipped tests were failing before, so these should be fixed vs skipping before proceeding.

packages/calcite-components/src/components/dropdown-group/dropdown-group.tsx
```diff
@@ -92,7 +94,7 @@ export class DropdownGroup {
       this.groupPosition > 0 ? <div class="dropdown-separator" role="separator" /> : null;
 
     return (
-      <Host aria-label={this.groupTitle} role="group">
+      <Host aria-label={this.groupTitle} id={this.el.id || this.guid} role="menu" tabindex="-1">
```
geospatialem, line 97: Could you speak to the use case for changing the role from `"group"` to `"menu"`? Overall `"group"` provides more support for AT, in particular NVDA. This could be why NVDA isn't receiving context with the changes.

Role and AT support conveying the role and name:
- [Group role support](https://a11ysupport.io/tech/aria/group_role)
  - Supports role and name for JAWS, Narrator, NVDA and VoiceOver on desktop
- [Menu role support](https://a11ysupport.io/tech/aria/menu_role)
  - Supports role and name for JAWS and Narrator
  - Supports role only for NVDA and VoiceOver

packages/calcite-components/src/components/dropdown/dropdown.tsx
```diff
@@ -452,11 +461,44 @@ export class Dropdown
...
+  private _setSemanticAriaAssociation(
+    selectionMode,
+    selectedDropdownItemID,
+    selectedDropdownGroupID
+  ) {
+    let newSelection = [];
+
+    return (trigger) => {
+      if (selectionMode === "multiple") {
+        const currentSelection = trigger.getAttribute("aria-labeledby")?.split(" ") || [];
```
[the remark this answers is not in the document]
geospatialem: Even in updating the above attributes in the code, there is no support for NVDA, which might be related to the `"menu"` role. We might need to consider keeping the role as-is (`"group"`) or refactoring the components to support NVDA users.

geospatialem (commented):
There is context provided for JAWS, which is excellent! Did observe NVDA is still not picking up context, and think it could be related to the `"menu"` role change. Wondering if we could explore alternatives with the previously defined role of `"group"`, which provides more support to the role and name across AT.

Issue #2162, opened 2021-05-18
[Combobox] clear user entered text on blur


avezina:
## The description below is outdated and the issue title has been updated accordingly. Please read all the comments below before implementing, specifically https://github.com/Esri/calcite-components/issues/2162#issuecomment-1087759131.

### Description
My team is working on a widget where we need to be able to do multiple selections from a drop-down. For example in the mockup below with the multiple check-boxes.
![image](https://user-images.githubusercontent.com/50338557/118719736-b1963a00-b7dd-11eb-9d06-2783ba51a7e9.png)
![image](https://user-images.githubusercontent.com/50338557/118719933-e2766f00-b7dd-11eb-9a22-2185ad377843.png)

### Acceptance Criteria 
Ideally, we need a way to select multiple choices, and we need the selected choices to be listed against the header item so the end-user can see which items were selected without expanding the dropdown.

#### Which Component
It seems the Combobox provides the option to expose the selected options as tabs, but we would like to disable the input text parameter and only rely on selecting options from the dropdown similar to the Button>Dropdown>Simple component.
![image](https://user-images.githubusercontent.com/50338557/118720360-5c0e5d00-b7de-11eb-8d2d-053f95582ec9.png)
![image](https://user-images.githubusercontent.com/50338557/118720728-d4751e00-b7de-11eb-80d7-2f4bfe42ab3b.png)

We may just use the Combobox for now, and remove the input text later if that parameter is made optional.

bstifle:
Hi @avezina ! 

The dropdown seems like unnecessary redundancy. In this use case, can we just remove the dropdown and rely on the checkboxes?

macandcheese:
@avezina - checking in on this. Is this still valid?

avezina:
@macandcheese thanks for looping back on this. Yes this is still a valid request. For example, we are using the Combobox in the [UtilityNetworkTrace widget (see sample)](https://developers.arcgis.com/javascript/latest/sample-code/widgets-untrace/) where one or multiple values can be selected. While the text could be used to search a value, I feel like it should not persist?
![image](https://user-images.githubusercontent.com/50338557/161552460-2d34c3ff-fde7-4bc4-9dac-b676012c8ff5.png)

macandcheese:
I can’t seem to preview that UX in the sample (it asks for authentication to view layers - after which the panel is disabled), however - the combobox shouldn’t persist text after making a selection - after it is entered and an item is selected, the text clears.

[...]

PR #7721, opened 2023-09-11
fix(combobox): clears input value on blur

anveshmekala:
**Related Issue:** #2162 

## Summary

Clears the user entered input on `blur` when `allowCustomValues` is set to `false`


driskull (approved):
LGTM!! Nice 👍

PR #7743, opened 2023-09-14
chore: add util to generate breakpoint stories

jcfranco:
**Related Issue:** N/A

## Summary

Adds helper to ease adding responsive layout screenshot tests.

### DIFF
--- a/packages/calcite-components/.storybook/utils.tsx
+++ b/packages/calcite-components/.storybook/utils.tsx
@@ -20,6 +20,9 @@ import { colors } from "../../../node_modules/@esri/calcite-colors/dist/colors";
 import { Description, DocsPage } from "@storybook/addon-docs";
 import { Theme as Mode } from "storybook-addon-themes/dist/models/Theme";
 import React from "react";
+import { Scale } from "../src/components/interfaces";
+import { html } from "../support/formatting";
+import { Breakpoints } from "../src/utils/responsive";
 
 const autoValue = {
   name: "Auto",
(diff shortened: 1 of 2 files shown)


packages/calcite-components/.storybook/utils.tsx
```diff
@@ -139,3 +142,69 @@ export const filterComponentAttributes = (
...
+ *   This should only be used if multiple stories cannot be displayed side-by-side.
+ */
+export function createBreakpointStories(
+  singleStoryHtml: string,
+  focused?: { breakpoint: keyof Breakpoints["width"]; scale: Scale }
+): string {
+  // we hard-code breakpoint values because we can't read them directly from the page when setting up a story
+  // based on https://github.com/Esri/calcite-design-tokens/blob/2e8fc1b8f410b5443fa53ca1c12ceef71e651b9a/tokens/core.json#L1533-L1553
+  const widthBreakpoints: { name: keyof Breakpoints["width"]; maxWidth: number }[] = [
+    { name: "xsmall", maxWidth: 476 },
```
driskull, line 160: Once we can get these values from the tokens we should reference them here.

packages/calcite-components/.storybook/utils.tsx
```diff
@@ -139,3 +142,69 @@ export const filterComponentAttributes = (
...
+            ${singleStoryHtml.replace(placeholderPattern, (_match, placeholder: string) =>
+              placeholder === "scale" ? scale : placeholder
+            )}
+          </div>`;
+        });
+    });
+
+  return html`<div class="${css.storiesContainer}">
+    <style>
+      .${css.storiesContainer} {
```
driskull, line 193: I wonder if we should just add this CSS file instead of in HTML?

driskull (approved):
👍

packages/calcite-components/.storybook/utils.tsx
jcfranco: Do you mean as a stylesheet import in [`preview-html`](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/.storybook/preview-head.html)? FWIW, I went this route to keep the breakpoint-story styling closer to its usage, but I agree it'd be nice to have a single storybook CSS file to reference. I'll create a refactor issue to extract all styling to a global storybook stylesheet.

Issue #6694, opened 2023-03-28
[Stepper] Responsive layout


brittneytewks:
### Description

Responsiveness across Calcite Design System, focused on web (traditionally desktop sized web browser), tablet, phone screen sizes, and large monitors.

### Acceptance Criteria

1. Consider breaking changes for folks. Consultation should occur on both the design and development teams.
2. Components need to be designed for the following breakpoints:

- [x]  Small (s), 476px
- [x]  Medium (m), 768px
- [x]  Large (l), 1152px
- [x]  Extra larger (xl), 1440px

### Relevant Info

Part of epic https://github.com/Esri/calcite-components/issues/6670

### Which Component

`stepper`

### Priority impact

p4 - not time sensitive

### Esri team

Calcite (design)

SkyeSeitz:
Update from team discussion today:
I presented iterations for a Stepper on smaller screen sizes and mobile devices. Below is the solution we landed on, which I will refine before our final design proposal with additional variants and vertical orientation.
<img width="890" alt="image" src="https://user-images.githubusercontent.com/29716057/229244718-675e8645-bcdb-46d4-ac73-0d8c3dfa8ece.png">

SkyeSeitz:
Added [all variants](https://www.figma.com/file/zBGnAOEOxqKGocUjqiYbed/Responsive-Layout?node-id=3624%3A105603&t=mjOP1jFnYACx7TSE-1) for us to review in Tuesday's meeting with a few follow up questions. 
<img width="1670" alt="image" src="https://user-images.githubusercontent.com/29716057/231034037-0439bbf9-8038-4cf6-9629-7ecf8a5deb35.png">

geospatialem:
Verified on `2.0.0-next.1` with default, `numbered`, `icon`, and `numbered` w/ `icon`

![image](https://github.com/Esri/calcite-design-system/assets/5023024/a9bdbd0d-058a-4ac9-a896-4f86b5b2ff49)

PR #7744, opened 2023-09-14
feat(stepper): enable responsive layout

anveshmekala:
**Related Issue:** #6694 

## Summary

Enables responsive layout for stepper component.

#### Additional Enhancements:

- stepper element `host` display is changed to `flex` from `grid`
- grid column width will use `minmax( )` instead of fixed fractional units.
- When none of the stepper item's are selected and the first one is disabled, the first available enabled stepper will be selected.


packages/calcite-components/src/components/stepper/stepper.stories.ts
```diff
@@ -212,3 +212,63 @@ export const verticalLayout_TestOnly = (): string => html`<calcite-stepper layou
...
+      </calcite-notice>
+    </calcite-stepper-item>
+  </calcite-stepper>`;
+
+export const responsiveLayoutXsmallScaleS_TestOnly = (): string => getBreakpointsStoryTemplate("s");
+export const responsiveLayoutXsmallScaleM_TestOnly = (): string => getBreakpointsStoryTemplate("m");
+export const responsiveLayoutXsmallScaleL_TestOnly = (): string => getBreakpointsStoryTemplate("l");
+
+responsiveLayoutXsmallScaleL_TestOnly.parameters = {
+  chromatic: { disableSnapshot: true },
```
anveshmekala, line 265: these will be enabled before install

packages/calcite-components/src/components/action/action.scss
```diff
@@ -76,7 +76,9 @@
 
 :host([scale="s"]) {
   .button {
-    @apply text-n2h px-2 py-1 font-normal;
+    @apply text-n2h font-normal;
```
driskull, line 79: Maybe these changes to action should be in a separate PR so this change can be tested and documented?

packages/calcite-components/src/components/stepper-item/stepper-item.tsx
```diff
@@ -123,6 +124,11 @@ export class StepperItem implements InteractiveComponent, LocalizedComponent, Lo
    */
   @Prop({ reflect: true }) scale: Scale = "m";
 
+  /**
+   * @internal
+   */
+  @Prop({ reflect: true }) responsiveMode = false;
```
driskull, line 130: It seems odd to have an internal prop for being responsive. Shouldn't it always be in "responsive mode"? Maybe the prop name is just odd?

packages/calcite-components/src/components/stepper-item/stepper-item.tsx
```diff
@@ -335,6 +358,34 @@ export class StepperItem implements InteractiveComponent, LocalizedComponent, Lo
...
+          (itemPosition === 0 && isPositionStart) ||
+          (itemPosition === totalItems - 1 && !isPositionStart)
+        }
+        icon={path}
+        iconFlipRtl={iconFlipRtl}
+        // eslint-disable-next-line react/jsx-no-bind
+        onClick={(event) => this.handleActionClick(event, position)}
+        onKeyDown={(event) => this.handleActionKeyDown(event, position)}
+        scale={this.scale}
+        text={isPositionStart ? "Previous Step" : "Next Step"}
```
driskull, line 384: This should be in the messages property right?

packages/calcite-components/src/components/stepper/step-bar.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export const CSS = {
+  stepBar: "step-bar",
+  stepBarActive: "step-bar--active",
+  stepBarInActive: "step-bar--inactive",
+  stepBarComplete: "step-bar--complete",
+  stepBarError: "step-bar--error",
+};
+
+export const StepBar: FunctionalComponent<StepBarProps> = ({ isActive, isComplete, isError }) => (
```
driskull, line 12: nice!

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -100,6 +103,12 @@ export class Stepper {
   connectedCallback(): void {
     this.mutationObserver?.observe(this.el, { childList: true });
     this.updateItems();
+    this.resizeObserver?.observe(document.body);
```
driskull, line 106: Question: Why are we observing the document body? Shouldn't we just be observing a component's size?

driskull (changes requested):
Have some initial comments and questions 👍

packages/calcite-components/src/components/stepper/stepper.tsx
anveshmekala: The component do not collapse entirely when viewport size decreases as per spec to avoid shrinking stepper-items. To identify if the viewport size has decreased and showcase UX for small screen when required we are observing `document.body`.  

Update:
After discussing with Skye regarding this, the suggestion is to have a min-width on each stepper item which helps us to calculate the breakpoint width more precisely. Changes coming soon 🚧

packages/calcite-components/src/components/stepper-item/stepper-item.tsx
anveshmekala: Yup, will create a separate PR for this.

packages/calcite-components/src/components/stepper-item/stepper-item.tsx
anveshmekala: I agree it is always in the responsive mode, the idea here was to change UI since the width has gone below the breakpoint but the name is confusing. Will come-up with a better name.

packages/calcite-components/src/components/stepper-item/stepper-item.tsx
anveshmekala: Installed PR #7918

packages/calcite-components/src/components/action/action.scss
anveshmekala: Redesigned to avoid changes at `calcite-action` level.

macandcheese:
Few comments... most related to UI / UX.

- bug? Seems like I can get the "responsive artifacts" to display over the original display mode. I can't easily replicate but grabbed this screenshot:
Screenshot 2023-11-09 at 10 18 34 AM

- previously - a user could tab and focus the title of the current step. In this new mode, there is no way to tab to the title / active item. The first focused item is the "previous" button. Is this a strange experience? Is this ok / expected from an a11y perspective?
- previously - a user could interact with and explicitly navigate to a specific step. Are we ok with removing this behavior with no way to disable?
- you can navigate to a disabled item:
Screenshot 2023-11-09 at 10 19 25 AM
- the icon is not aligned with step number / title
Screenshot 2023-11-09 at 10 19 05 AM
- additionally, as shown above - the icon (IMO) seems to indicate the step directly above it is active.
- It's just not clear what step I am on

anveshmekala:
> Few comments... most related to UI / UX.
> 
> * bug? Seems like I can get the "responsive artifacts" to display over the original display mode. I can't easily replicate but grabbed this screenshot:
>  Intresting, Will try to reproduce this one. Those artifacts shouldnt be visible.

> * previously - a user could tab and focus the title of the current step. In this new mode, there is no way to tab to the title / active item. The first focused item is the "previous" button. Is this a strange experience? Is this ok / expected from an a11y perspective?

The click on the title is restricted since we display the slotted content by default and user can view only one step at a time. In the current version for bigger screen clicking the title region is used for both navigation and displaying the slotted content. In terms of a11y though focusing is prevented the screen reader will still provide context to the user since `aria-hidden` is not added.

> * you can navigate to a disabled item:

the thought behind this is to help the user understand that disabled items are rendered. In larger screen mode, disabled items are visible to user by default where as in singleView mode for smaller screen hiding them from users will make them kind of non existent unless we provide an alternative way to convey the message. This is a tricky one  and open to suggestions.

> * the icon is not aligned with step number / title
Good catch, will fix the layout.

> * additionally, as shown above - the icon (IMO) seems to indicate the step directly above it is active.
> * It's just not clear what step I am on

i agree, we can make the other steps more opaque to counter this. Thoughts? 

> * Is it valuable to display disabled arrow for first / last items?

No preference for this one.

> * previously - a user could interact with and explicitly navigate to a specific step. Are we ok with removing this behavior with no way to disable?

Had discussion about this during the initial phase. There was a thought of making the step-bar's interactive so that user can switch across items more easily. The concern is regarding the hit area dimensions being tiny (especially smaller screens) and the length of step-bar can diminish if there are more steps. We can add this as an enhancement also. 


cc : @SkyeSeitz

packages/calcite-components-angular/projects/component-library/src/public-api.ts
```diff
@@ -1,3 +1,3 @@
-export * from './lib/component-library.module';
-export { DIRECTIVES } from './lib/stencil-generated';
-export * from './lib/stencil-generated/components';
+export * from "./lib/component-library.module";
```
jcfranco, line 1: This is unrelated to the PR, so could you roll it back?

packages/calcite-components/src/components/icon/icon.tsx
```diff
@@ -142,7 +142,13 @@ export class Icon {
       return;
     }
 
-    this.pathData = await fetchIcon({ icon, scale });
+    const pathData = await fetchIcon({ icon, scale });
```
jcfranco, line 145: Could you add a brief explainer on why this is needed?

packages/calcite-components/src/components/stepper-item/stepper-item.tsx
```diff
@@ -142,6 +142,11 @@ export class StepperItem
    */
   @Prop({ reflect: true }) scale: Scale = "m";
 
+  /**
+   * @internal
+   */
```
jcfranco, line 147: Could you add an explainer on what this is for and how it's used?

packages/calcite-components/src/components/stepper-item/stepper-item.tsx
```diff
@@ -358,13 +363,14 @@ export class StepperItem
...
   private renderIcon(): VNode {
-    const path = this.selected
-      ? "circleF"
-      : this.error
-      ? "exclamationMarkCircleF"
-      : this.complete
-      ? "checkCircleF"
-      : "circle";
+    const path =
+      this.selected && (this.singleViewMode ? !this.error && !this.complete : true)
```
jcfranco, line 367: This is getting unwieldy. It might be time to use an if/else for this.

Also, I think you can replace

```ts
this.singleViewMode ? !this.error && !this.complete : true
```

with 

```ts
!this.singleViewMode || (!this.error && !this.complete)
```

packages/calcite-components/src/components/stepper/stepper.scss
```diff
@@ -20,6 +38,62 @@
...
+
+:host([layout="horizontal"][scale="s"]) {
+  gap: theme("spacing.1") var(--calcite-internal-stepper-item-spacing-unit-s);
+}
+
+:host([layout="horizontal"][scale="l"]) {
+  gap: theme("spacing.3") var(--calcite-internal-stepper-item-spacing-unit-s);
+}
+
+.step-bar {
```
jcfranco, line 52: Could you use a mixin to isolate all of the step bar styling?

packages/calcite-components/src/components/stepper/stepper.stories.ts
```diff
@@ -210,3 +210,87 @@ export const verticalLayout_TestOnly = (): string => html`<calcite-stepper layou
...
+      heading="Confirm and complete"
+      description="Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua"
+    >
+      <calcite-notice open width="full">
+        <div slot="message">Step 4 Content Goes Here</div>
+      </calcite-notice>
+    </calcite-stepper-item>
+  </calcite-stepper>`;
+
+export const responsiveLayoutScaleS_TestOnly = (): string => getBreakpointsStoryTemplate({ scale: "s", width: 400 });
```
jcfranco, line 259: Could you use the [`createBreakpointStories` util](https://github.com/Esri/calcite-components/blob/0ee6c16e2e9e3f45185f06eb87705b98bc87aba8/packages/calcite-components/.storybook/utils.tsx#L153-L152) for this? 

BTW, what's the reason behind testing widths between the defined max-width breakpoints?

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -310,25 +330,50 @@ export class Stepper implements LocalizedComponent, T9nComponent {
...
+  @State() elWidth: number;
+
+  @Watch("elWidth")
+  handleElWidthChange(): void {
+    requestAnimationFrame(() => {
+      this.determineActiveStepper();
+    });
+  }
 
   /** list of enabled Stepper items */
```
jcfranco, line 352: [Unrelated] I think we can drop this comment as the variable name is clear.

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -380,17 +459,83 @@ export class Stepper implements LocalizedComponent, T9nComponent {
...
+        // eslint-disable-next-line react/jsx-no-bind
+        onClick={(event) => this.handleActionClick(event, position)}
+        scale={this.scale}
+        text={isPositionStart ? "Previous Step" : "Next Step"}
+      />
+    ) : null;
+  }
+
+  private handleActionClick(event: MouseEvent, position: Position): void {
+    event.stopPropagation();
```
jcfranco, line 494: Is it possible to remove this? We should not be stopping clicks from propagating.

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -310,25 +330,50 @@ export class Stepper implements LocalizedComponent, T9nComponent {
...
-  /** keep track of the currently active item position */
-  private currentPosition: number;
+  private itemMap = new Map<HTMLCalciteStepperItemElement, { position: number; content: Node[] }>();
+
+  /** list of sorted Stepper items */
+  private items: HTMLCalciteStepperItemElement[] = [];
 
   private mutationObserver = createObserver("mutation", () => this.updateItems());
 
+  private singleViewMode = false;
```
jcfranco, line 362: Similar to the item, could you add some doc for this prop?

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -310,25 +330,50 @@ export class Stepper implements LocalizedComponent, T9nComponent {
     updateMessages(this, this.effectiveLocale);
   }
 
-  private itemMap = new Map<HTMLCalciteStepperItemElement, { position: number; content: Node[] }>();
+  /** keep track of the currently active item position */
+  @State() currentPosition: number;
```
jcfranco, line 334: Can this be renamed to `currentActiveItemPosition` to avoid the comment?

packages/calcite-components/src/components/stepper/stepper.e2e.ts
```diff
@@ -683,7 +706,136 @@ describe("calcite-stepper", () => {
...
+          <div>Step 3 content</div>
+        </calcite-stepper-item>
+      </calcite-stepper>`);
+
+      const stepper = await page.find("calcite-stepper");
+      const [actionStart, actionEnd] = await page.findAll("calcite-stepper >>> calcite-action");
+      const [stepperItem1, stepperItem2, stepperItem3] = await page.findAll("calcite-stepper-item");
+      const eventSpy = await stepper.spyOnEvent("calciteStepperItemChange");
+
+      expect(await actionStart.isVisible()).toBe(true);
```
jcfranco, line 813: Do we need to assert on visibility if this test is focusing on event emitting? If so, the test name might need to be renamed. May apply to other tests.

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -338,6 +383,48 @@ export class Stepper implements LocalizedComponent, T9nComponent {
...
+        } else {
+          item.style.display = "contents";
+          item.singleViewMode = true;
+        }
+      });
+    } else if (this.elWidth > totalMinWidthOfItems + totalRowGap) {
+      if (!this.singleViewMode) {
+        return;
+      }
+      this.el.style.display = "grid";
```
jcfranco, line 417: Can you look at pushing static styles to a CSS class and only use inline styles for computed values?

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -310,25 +330,50 @@ export class Stepper implements LocalizedComponent, T9nComponent {
...
 
-  private itemMap = new Map<HTMLCalciteStepperItemElement, { position: number; content: Node[] }>();
+  /** keep track of the currently active item position */
+  @State() currentPosition: number;
 
-  /** list of sorted Stepper items */
-  private items: HTMLCalciteStepperItemElement[] = [];
+  @Watch("currentPosition")
+  handlePositionChange(): void {
+    requestAnimationFrame(() => {
```
jcfranco, line 338: Since `determineActiveStepper` is calling `getComputedStyle`, could you try using [`readTask`](https://stenciljs.com/docs/api#other) vs `requestAnimationFrame` here?

jcfranco (changes requested):

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -152,14 +156,31 @@ export class Stepper implements LocalizedComponent, T9nComponent {
...
+    this.resizeObserver?.disconnect();
     disconnectMessages(this);
     disconnectLocalized(this);
     this.mutationObserver?.disconnect();
   }
 
   render(): VNode {
     return (
-      <Host aria-label={this.messages.label} role="region">
+      <Host aria-label={"Progress steps"} role="region">
```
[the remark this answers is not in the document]
anveshmekala: This handled in main already. need to merge master.

packages/calcite-components/src/components/stepper/stepper.tsx
anveshmekala: I am trying hard to recall, why this was added in first place. Will remove it and see if any test fails.

packages/calcite-components/src/components/stepper/stepper.tsx
anveshmekala: Dropped.

packages/calcite-components/src/components/stepper/resources.ts
```diff
@@ -0,0 +1,14 @@
+//todo : replace with design tokens
```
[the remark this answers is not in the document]
anveshmekala: added https://github.com/Esri/calcite-design-system/issues/8174

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -310,25 +330,50 @@ export class Stepper implements LocalizedComponent, T9nComponent {
...
+  }
 
   /** list of enabled Stepper items */
   private enabledItems: HTMLCalciteStepperItemElement[] = [];
 
-  /** keep track of the currently active item position */
-  private currentPosition: number;
+  private itemMap = new Map<HTMLCalciteStepperItemElement, { position: number; content: Node[] }>();
+
+  /** list of sorted Stepper items */
```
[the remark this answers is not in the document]
anveshmekala: i am thinking to remove the comment and keep the name `items`.  The reason why i added this earlier is to differentiate between `itemMap` and `items`. Looks like we can remove `itemMap` since it is not used anywhere else. Thoughts?

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -380,17 +459,83 @@ export class Stepper implements LocalizedComponent, T9nComponent {
...
+    if (
+      this.currentPosition &&
+      currentActivePosition !== this.currentPosition &&
+      !this.items[this.currentPosition].disabled
+    ) {
+      this.calciteStepperItemChange.emit();
+    }
+  }
 
   private getFirstEnabledStepperPosition(): number {
```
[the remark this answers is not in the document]
anveshmekala: Makes sense.Extracted `currentPosition =0` outside of the method.

packages/calcite-components/src/components/stepper/stepper.stories.ts
anveshmekala: Stepper responsive layout is designed a bit different compared to others. The component doesn't depend upon a fixed breakpoint width and uses min-width to calculate how many items can fit within the page dimensions. The [createBreakpointStories util](https://github.com/Esri/calcite-components/blob/0ee6c16e2e9e3f45185f06eb87705b98bc87aba8/packages/calcite-components/.storybook/utils.tsx#L153-L152) is using hard-corded value to test which has the lowest breakpoint at `476px`. This works currently since we are using 4 `stepper-item's` in each story and the lowest possible width is `480px`, but if there is a change to the value for `xsmall` width the tests will break. 

I don't think changing the value of `xsmall` should break these stories. Perhaps we can enhance the util to add a custom width scale. Thoughts?

packages/calcite-components/src/components/stepper/step-bar.tsx
```diff
@@ -0,0 +1,45 @@
+import { FunctionalComponent, h } from "@stencil/core";
+interface StepBarProps {
```
jcfranco, line 2: Could you add a newline here to separate the imports from the interface?

packages/calcite-components/src/components/stepper/stepper.scss
```diff
@@ -9,17 +27,80 @@
...
   display: grid;
   // grid-templates-columns are dynamically generated
   // grid-templates-rows are dynamically generated
   grid-template-areas:
     "items"
     "content";
+  gap: theme("spacing.2") var(--calcite-internal-stepper-item-spacing-unit-s);
+}
+
+:host([layout="horizontal"][scale="s"]) .conatiner {
```
jcfranco, line 44: Typo: `.conatiner` ➡️ `.container`

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -310,25 +343,50 @@ export class Stepper implements LocalizedComponent, T9nComponent {
...
+  private items: HTMLCalciteStepperItemElement[] = [];
 
   private mutationObserver = createObserver("mutation", () => this.updateItems());
 
+  /** Specifies if the user is viewing one `stepper-item` at a time when the page width is less than sum of min-width of each item. */
+  private singleViewMode = false;
+
+  private guid = `calcite-stepper-action-${guid()}`;
+
+  containerEl: HTMLDivElement;
```
jcfranco, line 377: Could you make this private as well?

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -380,17 +471,86 @@ export class Stepper implements LocalizedComponent, T9nComponent {
...
+      (el) => el?.tagName === "CALCITE-STEPPER-ITEM"
+    );
+    this.items = items as HTMLCalciteStepperItemElement[];
+    this.setGridTemplateColumns(items);
+    this.setStepperItemNumberingSystem();
+  };
+
+  setGridTemplateColumns = (items: Element[]): void => {
+    const minWidth = this.getMinWidthOfStepperItem();
+    const spacing = Array(items.length).fill(`minmax(${minWidth}px,1fr)`).join(" ");
```
jcfranco, line 548: For readability, can you add a space between the `minmax` arguments?

packages/calcite-components/src/components/stepper-item/stepper-item.scss
```diff
@@ -348,6 +335,44 @@
...
+:host([layout="horizontal"][single-view-mode][complete][selected]) .container {
+  & .stepper-item-icon {
+    @apply opacity-disabled;
+  }
+  & .stepper-item-number {
+    @apply text-color-3;
+  }
+}
+
+:host([layout="horizontal"][data-display="hide"]) {
```
jcfranco, line 372: Is it possible to use `hidden` vs `data-display="hide"` instead? `data-display="visible"` doesn't seem to be used.

packages/calcite-components/src/components/stepper-item/assets/stepper-item/t9n/messages.json
```diff
@@ -1,5 +1,3 @@
 {
-  "complete": "Completed step",
-  "previousStep": "Previous step",
```
jcfranco, line 3: Message removal treads into a territory of technically being breaking, but probably not affecting many, so we have a couple of options:

1. Restore these strings and remove them in a separate one (similar to https://github.com/Esri/calcite-design-system/pull/8099) so that the responsive feature isn't seen as itself being breaking in the changelog
2. Mark this PR as breaking and be clear about it only affecting anyone actually using the strings object or its type

I'm leaning more towards 1.

packages/calcite-components/src/components/stepper/stepper.e2e.ts
```diff
@@ -460,8 +475,10 @@ describe("calcite-stepper", () => {
           const item2 = wrapper.shadowRoot.querySelector<HTMLElement>("#item-2");
           await new Promise<void>((resolve) => requestAnimationFrame(() => resolve()));
```
jcfranco, line 476: I totally missed this earlier, but there's a test util ([`waitForAnimationFrame`](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/tests/utils.ts#L279)) you can use to simplify. Not critical for this PR.

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -380,17 +473,86 @@ export class Stepper implements LocalizedComponent, T9nComponent {
...
+  handleDefaultSlotChange = (event: Event): void => {
+    const items = slotChangeGetAssignedElements(event).filter(
+      (el) => el?.tagName === "CALCITE-STEPPER-ITEM"
+    );
+    this.items = items as HTMLCalciteStepperItemElement[];
+    this.setGridTemplateColumns(items);
+    this.setStepperItemNumberingSystem();
+  };
+
+  setGridTemplateColumns = (items: Element[]): void => {
```
jcfranco, line 548: This and `getMinWidthOfStepperItem` can be class functions instead of assigned bound functions since these are not event handlers and `this` is bound to the component instance at their call site. https://www.joshmorony.com/understanding-jsx-for-stencil-js-applications/#event-binding has some nice and concise info on this.

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -338,38 +396,73 @@ export class Stepper implements LocalizedComponent, T9nComponent {
...
+      }
+
+      this.singleViewMode = true;
+
+      this.items.forEach((item: HTMLCalciteStepperItemElement, index) => {
+        if (index !== activePosition) {
+          item.setAttribute("hidden", "true");
+          item.setAttribute("aria-hidden", "true");
+        } else {
+          item.removeAttribute("hidden");
```
jcfranco, line 422: All HTML elements have a [`hidden`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden) prop, so you can it directly instead of setting the attribute.

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -338,38 +396,73 @@ export class Stepper implements LocalizedComponent, T9nComponent {
...
+      if (this.singleViewMode && !currentActivePositionChanged) {
+        return;
+      }
+
+      this.singleViewMode = true;
+
+      this.items.forEach((item: HTMLCalciteStepperItemElement, index) => {
+        if (index !== activePosition) {
+          item.setAttribute("hidden", "true");
+          item.setAttribute("aria-hidden", "true");
```
jcfranco, line 420: I _think_ `aria-hidden` may be redundant since we're using `hidden`. cc @geospatialem

jcfranco (approved):
Awesome stuff, @anveshmekala!

🌊🌊🌊🌊🌊🌊🌊🌊🌊🌊🌊🌊🌊🌊🌊🌊🌊🌊
🌊🏄🏄🏄🌊🌊🏄🏄🏄🏄🌊🏄🏄🏄🌊🌊🏄🌊
🌊🏄🌊🌊🏄🌊🏄🌊🌊🏄🌊🏄🌊🌊🏄🌊🏄🌊
🌊🏄🏄🏄🌊🌊🏄🏄🏄🏄🌊🏄🌊🌊🏄🌊🏄🌊
🌊🏄🌊🏄🌊🌊🏄🌊🌊🏄🌊🏄🌊🌊🏄🌊🌊🌊
🌊🏄🌊🌊🏄🌊🏄🌊🌊🏄🌊🏄🏄🏄🌊🌊🏄🌊
🌊🌊🌊🌊🌊🌊🌊🌊🌊🌊🌊🌊🌊🌊🌊🌊🌊🌊

packages/calcite-components/src/components/stepper/stepper.e2e.ts
anveshmekala: Tried this but for some reason it is not working in browser context. Am i missing anything ?

jcfranco: Oh, my bad. I didn't realize this was within an `evaluate` callback. Ignore my comment then. 

There is a way to split up the `evaluate` to allow us use the helper in the test context, but we can tackle that later.

packages/calcite-components/src/components/stepper-item/assets/stepper-item/t9n/messages.json
anveshmekala: Added https://github.com/Esri/calcite-design-system/pull/8222

packages/calcite-components/src/components/stepper/step-bar.tsx
```diff
@@ -0,0 +1,46 @@
...
+export const CSS = {
+  stepBar: "step-bar",
+  stepBarActive: "step-bar--active",
+  stepBarComplete: "step-bar--complete",
+  stepBarDisabled: "step-bar--disabled",
+  stepBarError: "step-bar--error",
+  stepBarInActive: "step-bar--inactive",
+};
+
+export const StepBar: FunctionalComponent<StepBarProps> = ({
```
driskull, line 20: Should this file be in a `functional` folder or something so we know its a functional component? Our other functional components are in the functional folder.

packages/calcite-components/src/components/stepper/resources.ts
```diff
@@ -0,0 +1,14 @@
+export const ITEM_MIN_WIDTH = {
+  s: 120,
```
driskull, line 2: Should these sizes be defined by any tokens?

packages/calcite-components/src/components/stepper/stepper.e2e.ts
```diff
@@ -68,14 +68,29 @@ describe("calcite-stepper", () => {
           <div>Step 4 content</div>
         </calcite-stepper-item>
       </calcite-stepper>`,
-      { display: "grid" }
+      { display: "flex" }
```
driskull, line 71: would this be a breaking change?

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -160,7 +172,29 @@ export class Stepper implements LocalizedComponent, T9nComponent {
...
+        <div
+          class={{ container: true, [CSS.singleView]: this.singleViewMode }}
+          ref={this.setContainerEl}
+        >
+          {this.singleViewMode && (
+            <div class={{ [CSS.stepBarContainer]: true }}>
+              {this.items.map((item, index) => (
+                <StepBar
+                  disabled={item.disabled && index !== this.currentActivePosition}
+                  isActive={index === this.currentActivePosition}
```
driskull, line 184: do you need the `isActive` or could it just be `active`?

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -160,7 +172,29 @@ export class Stepper implements LocalizedComponent, T9nComponent {
...
+                  isComplete={item.complete && index !== this.currentActivePosition && !item.error}
```
driskull, line 185: do you need the `is` or could it just be `complete`?

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -160,7 +172,29 @@ export class Stepper implements LocalizedComponent, T9nComponent {
...
+                  isError={item.error && index !== this.currentActivePosition}
```
driskull, line 186: should this be `hasError` or `showError`  instead of `isError`?

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -310,25 +343,50 @@ export class Stepper implements LocalizedComponent, T9nComponent {
...
-  /** keep track of the currently active item position */
-  private currentPosition: number;
+  private itemMap = new Map<HTMLCalciteStepperItemElement, { position: number; content: Node[] }>();
+
+  private items: HTMLCalciteStepperItemElement[] = [];
 
   private mutationObserver = createObserver("mutation", () => this.updateItems());
 
+  /** Specifies if the user is viewing one `stepper-item` at a time when the page width is less than sum of min-width of each item. */
+  private singleViewMode = false;
```
driskull, line 373: Should this be true by default? Assume the smallest size first? I think thats how we decided to handle our breakpoints.

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -380,17 +470,86 @@ export class Stepper implements LocalizedComponent, T9nComponent {
...
+          [CSS.actionIcon]: true,
+        }}
+        compact={true}
+        data-position={position}
+        disabled={
+          (currentActivePosition === 0 && isPositionStart) ||
+          (currentActivePosition === totalItems - 1 && !isPositionStart)
+        }
+        icon={path}
+        iconFlipRtl={dir === "rtl"}
```
driskull, line 495: this should be set to true and not depend on the dir. It handles the dir internally.

driskull (changes requested):
Looking good! 🎉

packages/calcite-components/src/components/stepper/resources.ts
anveshmekala: yeah they will be defined by tokens. Added an issue https://github.com/Esri/calcite-design-system/issues/8174

packages/calcite-components/src/components/stepper/stepper.tsx
anveshmekala: it can be `active`, renamed `isError` and `isComplete` too.

packages/calcite-components/src/components/stepper/stepper.tsx
anveshmekala: I think we can just use `error` since it is a UI state rather than a `boolean`.

packages/calcite-components/src/components/stepper/step-bar.tsx
anveshmekala: Agreed, moved to the functional folder.

packages/calcite-components/src/components/stepper/stepper.e2e.ts
driskull: Visually, it shouldn't but we should at least make a note of this change incase anyone was relying on this element being a grid display.

packages/calcite-components/src/components/stepper/stepper.tsx
anveshmekala: make sense, changed this to `multipleViewMode = false`

driskull (approved):
🎉 🌮 !

Issue #6374, opened 2023-01-29
Audit and update component usage snippets


geospatialem:
### Description

Audit and update all component usage snippets in the `<component>/usage` folder 📁. 

The audit should determine:
1. If all usage snippets are up-to-date with props, slots, etc.
  - If not, usage snippets should be cleaned up and/or removed
2. If additional use snippets should be added to showcase the component effectively.

### Which Component

All components

<details>
  <summary><h2>Progress checklist</h2></summary>

- [ ] `accordion` 
- [ ] `accordion-item` 
- [ ] `action` 
- [ ] `action-bar` 
- [ ] `action-group` 
- [ ] `action-menu` 
- [ ] `action-pad`
- [ ] `alert`
- [ ] `avatar`
- [ ] `block`
- [ ] `block-section`
- [ ] `button` 
- [ ] `card` 
- [ ] `checkbox`
- [ ] `chip`
- [ ] `color-picker`
- [ ] `combobox`
- [ ] `combobox-item` 
- [ ] `combobox-item-group` 
- [ ] `date-picker` 
- [ ] `dropdown` 
- [ ] `dropdown-group` 
- [ ] `dropdown-item` 
- [ ] `fab` 
- [ ] `filter` 
- [ ] `flow` 
- [ ] `flow-item` 
- [ ] `icon` 
- [ ] `inline-editable` 
- [ ] `input` 
- [ ] `input-date-picker` 
- [ ] `input-message` 
- [ ] `input-time-picker` 
- [ ] `label` 
- [ ] `link` 
- [ ] `list` 
- [ ] `list-item` 
- [ ] `list-item-group` 
- [ ] `loader` 
- [ ] `modal` 
- [ ] `notice` 
- [ ] `option` 
- [ ] `option-group`
- [ ] `pagination` 
- [ ] `panel` 
- [ ] `pick-list` 
- [ ] `pick-list-group`
- [ ] `pick-list-item` 
- [ ] `popover` 
- [ ] `progress` 
- [ ] `radio-button` 
- [ ] `radio-button-group` 
- [ ] `rating`
- [ ] `scrim`
- [ ] `segmented-control` 
- [ ] `segmented-control-item` 
- [ ] `select`
- [ ] `shell`
- [ ] `shell-center-row` 
- [ ] `shell-panel`
- [ ] `slider` 
- [ ] `split-button` 
- [ ] `stepper` 
- [ ] `stepper-item` 
- [ ] `switch` 
- [ ] `tab`
[...]

PR #7750, opened 2023-09-15
docs(action, action-pad): update usage and action's `tooltip` slot 

geospatialem:
**Related Issue:** #6374

## Summary
Includes a doc cleanup 🧹  for our a-named components, which include:
- `action`: discovered the use for a `tooltip` slot, which wasn't documented, should it be? ❓ 
- `action-pad`: updates the tooltip use sample to the `"expand-tooltip"` slot


driskull (approved):
👍

Issue #6689, opened 2023-03-28
[Tabs] Responsive Layout


brittneytewks:
### Description

Responsiveness across Calcite Design System, focused on web (traditionally desktop sized web browser), tablet, phone screen sizes, and large monitors.

### Acceptance Criteria

1. Consider breaking changes for folks. Consultation should occur on both the design and development teams.
2. Components need to be designed for the following breakpoints:

- [ ]  Small (s), 476px
- [ ]  Medium (m), 768px
- [ ]  Large (l), 1152px
- [ ]  Extra larger (xl), 1440px

### Relevant Info

Part of epic https://github.com/Esri/calcite-components/issues/6670

### Which Component

`tabs`

### Priority impact

p4 - not time sensitive

### Esri team

Calcite (design)

SkyeSeitz:
Update from team discussion today:
@jacqueskeet presented iterations for Tabs on smaller screen sizes and mobile devices. Below is the solution we landed on, which Jacques will refine with matching scale Actions and UX interactions notes. 

https://user-images.githubusercontent.com/29716057/231278532-02058fe7-c5d9-4d72-bea4-ed7458cfdb6a.mov

DitwanP:
🍬 Verified

PR #7751, opened 2023-09-15
feat(tabs, tab-nav, tab-title, tab): make component responsive

Elijbet:
**Related Issue:** #6689

## Summary

Implementation Details:

`tab-title`
- can vary in content, including text length and the presence of `icon-end, icon-start, and close button`
- hug the content and allow for variable width
- for `inline layout`, `tab-titles` overflow if they don't fit within the bounds of `tab-nav` and allow for arrow scroll navigation
- responsiveness occurs through `tab-title text` truncation for `center layout`

Detect overflow:
- The `getOverflowIcons()` function detects overflow over the right and left edges of the tab nav.
- If the last tab-title is outside of the tab-nav boundary, but the first isn't, it adds only the right arrow icon and vice versa.
- If both are outside the tab-nav boundary, it adds both chevrons.

`action button` Callback:
- The function handles scrolling to the `next | previous` set of `tab-titles` in a tabbed interface when the `right | left chevron` is clicked.
- It finds the last visible tab-title index within the mobile-sized tab navigation and calculates the scroll amount needed to bring the next set of tab titles into view.
- The scroll amount is applied to the tab navigation element using `requestAnimationFrame` for smooth scrolling.


packages/calcite-components/src/components/tab-nav/tab-nav.scss
```diff
@@ -41,6 +42,20 @@
     ease-out;
 }
 
+calcite-action {
+  @apply absolute;
+  inset-block-start: 50%;
+  transform: translateY(-50%);
+}
+
+.chevron-right {
```
driskull, line 51: question: for these `arrowRight` and `chevron-right` should we be using terms `chevron-end` and `arrowEnd` instead?

packages/calcite-components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -306,7 +349,18 @@ export class TabNav {
...
+  @State() defaultMessages: TabNavMessages;
+
+  @State() effectiveLocale = "";
+
+  @Watch("effectiveLocale")
+  effectiveLocaleChange(): void {
+    updateMessages(this, this.effectiveLocale);
+  }
+
+  @State() elWidth: number;
```
driskull, line 361: What is this state used for? Do we need it?

packages/calcite-components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -323,6 +381,58 @@ export class TabNav {
   //
   //--------------------------------------------------------------------------
 
+  private findVisibleTabTitleIndex = (tabTitles: NodeListOf<Element>, isNext: boolean): number => {
+    const mobilePageWidth = this.el.getBoundingClientRect().width;
+    let visibleTabTitleIndex = -1;
+
+    for (let i = 0; i < tabTitles.length; i++) {
```
driskull, line 388: Can you use `Array.from(tabTitles).forEach((tabTitle, index){...})`?

packages/calcite-components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -323,6 +381,58 @@ export class TabNav {
   //
   //--------------------------------------------------------------------------
 
+  private findVisibleTabTitleIndex = (tabTitles: NodeListOf<Element>, isNext: boolean): number => {
+    const mobilePageWidth = this.el.getBoundingClientRect().width;
```
driskull, line 385: This const is a bit confusing because this is the component's width and not a page width.

packages/calcite-components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -323,6 +381,58 @@ export class TabNav {
...
+    let visibleTabTitleIndex = -1;
+
+    for (let i = 0; i < tabTitles.length; i++) {
+      const tabTitle = tabTitles[i];
+      const tabTitleRect = tabTitle.getBoundingClientRect();
```
driskull, line 390: note: If we can store `getBoundingClinetRect` anywhere and not call it unless necessary that would be best as it is an expensive dom query.

packages/calcite-components/src/components/tab-nav/tab-nav.tsx
driskull: Can you use `this.el.clientWidth` or do you need to call `getBoundingClientRect()`?

packages/calcite-components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -406,4 +516,55 @@ export class TabNav {
       tabTitles[this.selectedTabId].focus();
     });
   }
+
+  private getOverflowIcons(): VNode | VNode[] {
+    const { messages } = this;
+    const tabNavWidth = this.el.offsetWidth;
+
+    const tabTitles = Array.from(this.el.querySelectorAll("calcite-tab-title"));
```
driskull, line 524: Looks like tab titles are queried here and inside `findVisibleTabTitleIndex`. Is there a way the query can be stored and used in both?

packages/calcite-components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -406,4 +516,55 @@ export class TabNav {
...
+
+    const firstTitle = tabTitles[0].getBoundingClientRect();
+    const lastTitle = tabTitles[tabTitles.length - 1].getBoundingClientRect();
+
+    const isOverflowingRight = lastTitle.right > tabNavWidth;
+    const isOverflowingLeft = firstTitle.left < 0;
+
+    const getActionChevronDirection = (overflowDirection: string): VNode => {
```
driskull, line 532: nitpick: should `overflowDirection` be typed to `"left" | "right"`? Maybe it should be `start` and `end`?

packages/calcite-components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -406,4 +516,55 @@ export class TabNav {
...
+        overflowDirection === "right"
+          ? this.scrollToNextTabTitles()
+          : this.scrollToPreviousTabTitles();
+
+      return (
+        <calcite-action
+          class={dirActionClass}
+          icon={dirChevronIcon}
+          onClick={() => dirScroll(overflowDirection)}
+          onKeyDown={() => dirScroll(overflowDirection)}
```
driskull, line 550: you shouldn't need onkeydown event here if you just want this to occur on enter/space. Otherwise, this needs to check which key should call the dirScroll.

driskull (commented):
Looking good @Elijbet!

My only concerns are with the getClientBoundingRect. If we can cache them as much as possible that would be great as we don't want to call them unless necessary.

packages/calcite-components/src/components/tab-nav/tab-nav.scss
Elijbet: Hm... I don't think so. I think start/end vocabulary is meant to indicate to a user where they are slotting their actions or icons. In this case, action is built in and conditionally rendered, and that naming is probably irrelevant.

driskull: Do these flip in RTL direction? If so then "right" isn't right anymore.

The start/end vocabulary is used to describe placement independent of direction (right/left) for page directions that are flipped.

Elijbet: Hm.. you're right. It would only flip if we have tab order flipping. Let me check.

Elijbet: It looks like it does swap the direction, so that was a good one to bring up. This makes sense, because tabs might be numbered and ordered by the user.

SkyeSeitz: This is a good point that we should pass along to the icon team. Are there hesitations with asking them to rename all icons with reference to right/left in favor of start/end?

driskull: I don't think the icon names are an issue, just the CSS class names and object properties.

SkyeSeitz: Oh missed that - thanks! 😁

packages/calcite-components/src/components/tab-nav/tab-nav.tsx
Elijbet: Nope. Relic of the past.

packages/calcite-components/src/components/tab-nav/tab-nav.e2e.ts
```diff
@@ -192,4 +174,96 @@ describe("calcite-tab-nav", () => {
...
+        }, 0);
+
+        return { tabTitlesTotalWidth, tabNavWidth };
+      });
+
+      expect(tabTitlesTotalWidth).toBeGreaterThan(tabNavWidth);
+    });
+
+    overflowScenarios.forEach(async (overflowScenario) => {
+      if (overflowScenario === "end") {
```
jcfranco, line 223: Would it be possible to add the specific scenarios the test needs to cover vs conditionally running different tests per scenario? Something similar to [this (note that it has some pending cleanup 😅)](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/tree/tree.e2e.ts#L1148). This would avoid disabling the `jest/no-conditional-expect` rule.

packages/calcite-components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -406,4 +511,46 @@ export class TabNav {
...
+      dir !== "rtl" ? lastTitle.right > tabNavWidth : lastTitle.right < tabNavWidth;
+    const isOverflowingStart = firstTitle.left < 0;
+
+    const getActionChevronDirection = (overflowDirection: string): VNode => {
+      const isEnd = overflowDirection === "end";
+      const dirActionClass: string = isEnd ? CSS.arrowEnd : CSS.arrowStart;
+      const dirChevronIcon: string = isEnd && dir !== "rtl" ? ICON.chevronRight : ICON.chevronLeft;
+      const dirText: string = isEnd ? messages.previousTabTitles : messages.nextTabTitles;
+
+      const dirScroll = () =>
```
jcfranco, line 533: Can this be an instance method instead? This is being created on each render.

jcfranco, line 533: Could you rename this to better express its intent?

packages/calcite-components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -406,4 +511,46 @@ export class TabNav {
...
+          scale={this.scale}
+          text={dirText}
+        />
+      );
+    };
+
+    const showEndAction = getActionChevronDirection("end");
+    const showStartAction = getActionChevronDirection("start");
+
+    const action = (isOverflowingEnd ? [showEndAction] : []).concat(
```
jcfranco, line 550: Can this be simplified? It's a bit hard to read.

packages/calcite-components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -406,4 +511,46 @@ export class TabNav {
...
+          class={dirActionClass}
+          icon={dirChevronIcon}
+          onClick={() => dirScroll()}
+          scale={this.scale}
+          text={dirText}
+        />
+      );
+    };
+
+    const showEndAction = getActionChevronDirection("end");
```
jcfranco, line 547: Both the end and start action vars are being created, but only one used. Can you update them to be created when needed?

packages/calcite-components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -187,6 +229,7 @@ export class TabNav {
               ref={(el) => (this.activeIndicatorEl = el as HTMLElement)}
             />
           </div>
+          {this.layout === "inline" && this.getOverflowIcons()}
```
jcfranco, line 232: `getOverflowIcons` has multiple `getBoundingClientRect` calls and is called during render. Can you decouple the dimension calculation from rendering to improve performance?

jcfranco (commented):
Looking good, @Elijbet. 😎✌️

packages/calcite-components/src/components/tab-nav/tab-nav.e2e.ts
```diff
@@ -1,20 +1,14 @@
...
+import { CSS } from "./resources";
 
 describe("calcite-tab-nav", () => {
   const tabNavHtml = "<calcite-tab-nav></calcite-tab-nav>";
 
-  describe("renders", () => {
-    renders(tabNavHtml, { display: "flex" });
-  });
-
-  describe("honors hidden attribute", () => {
```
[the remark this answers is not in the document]
Elijbet: We no longer have `tab-nav` not nested in the `tabs` now that it inherits from the tabs. And the parent already is covered with those tests. I made a note to add to docs that `tab-nav` cannot be used standalone anymore.

packages/calcite-components/src/components/tab-nav/tab-nav.e2e.ts
```diff
@@ -192,4 +174,96 @@ describe("calcite-tab-nav", () => {
...
+          expect(await page.find(`#testSubjectNav >>> .${CSS.arrowEnd}`)).not.toBe(null);
+          expect(await page.find(`#testSubjectNav >>> .${CSS.arrowStart}`)).toBe(null);
+        });
+      } else if (overflowScenario === "left") {
+        it("should show action buttons with correct chevrons for overflow to the left", async () => {
+          const isOverflowingLeft = await page.evaluate(async () => {
+            const tabNav = document.getElementById("testSubjectNav") as HTMLCalciteTabNavElement;
+            const tabTitles = Array.from(document.querySelectorAll("calcite-tab-title"));
+
+            const mobilePageWidth = tabNav.getBoundingClientRect().width;
```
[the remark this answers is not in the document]
Elijbet: Oh, this one isn't named right, it isn't really page width. I'll rename the variable.

packages/calcite-components/src/components/tab-nav/tab-nav.e2e.ts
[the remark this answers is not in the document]
jcfranco: I think we still need them despite not being able to be used standalone. From a consistency-perspective, our components should be able to render and be hidden, so we should cover that in tests too.

packages/calcite-components/src/components/tab-nav/tab-nav.tsx
Elijbet: Not sure if it was used before, but it is now.

Elijbet:
> Looking good @Elijbet!
> 
> My only concerns are with the getClientBoundingRect. If we can cache them as much as possible that would be great as we don't want to call them unless necessary.

I refactored the logic and no longer need `getBoundingClientRect`. I now use the index to determine what is visible in the viewport. Just need to clean it from the tests as well.

Elijbet:
I think we need to step back a bit and review the logic of clipping vs responsive scrolling. For some reason, this is pretty complex actually and we need to make sure we're all on the same page. Some thoughts. 

@SkyeSeitz could you attach a Figma mockup based on that little sketch I shared for a discussion here?
 
I think maybe the difference between the two is that the center scrolls both ways when it overflows, and for the inline, we limit the number of tabs we can add for both clipped and non-clipped options. 
 
I think the central reasoning behind adding responsive action buttons is an imaginary boundary at the ends of the width. The end tabs should not be able to detach and pass the start/end width boundaries and go past them inside the visible area (if we did allow this, we'd have to abandon the concept of inline vs center). 
 
But if we give the user the ability to set the clipping, then they shouldn't be able to retract the setting because, in the inline clipped text version, the number of tabs will easily overflow if not clipped. So we should let them set the clipped/unclipped prop upfront and limit tab numbers based on this prop. The Inline should not overflow on the right end. 

@jcfranco @macandcheese @ashetland

Elijbet:
> I think we need to step back a bit and review the logic of clipping vs responsive scrolling. For some reason, this is pretty complex actually and we need to make sure we're all on the same page. Some thoughts.
> 
> @SkyeSeitz could you attach a Figma mockup based on that little sketch I shared for a discussion here?
> 
> I think maybe the difference between the two is that the `center` scrolls both ways when it overflows, and for the `inline`, we limit the number of tabs we can add for both clipped and non-clipped options.
> 
> **I think the central reasoning behind adding responsive action buttons is an imaginary boundary at the ends of the width. The end tabs should not be able to detach and pass the start/end width boundaries and go past them inside the visible area** (if we did allow this, we'd have to abandon the concept of `inline` vs `center`).
> 
> But if we give the user the ability to set the clipping, then they shouldn't be able to retract the setting because, in the `inline` clipped text version, the number of tabs will easily overflow if not clipped. So we should let them set the clipped/unclipped prop upfront and limit tab numbers based on this prop. The `inline` should not overflow on the right end.
> 
> @jcfranco @macandcheese @ashetland

SkyeSeitz:
Thanks @Elijbet! Messaged ya to sync up today and share notes. Here is the link to the [Figma spec](https://www.figma.com/file/JAINbGHSykI8JQCVzOHiSz/Tabs---responsive-design-%5B6689%5D?type=design&node-id=511%3A11482&mode=dev) defining the behavior for inline vs center. 

Quick summary though: buttons + scrollable function would only come into play on `layout="inline"`. Center layout should continue to fill the container with Tab Titles simply getting more and more spread out or more and more tight as the component grows and shrinks. This means `layout="inline" would be the better layout option for larger amounts of Tabs.

Elijbet:
> Thanks @Elijbet! Messaged ya to sync up today and share notes. Here is the link to the [Figma spec](https://www.figma.com/file/JAINbGHSykI8JQCVzOHiSz/Tabs---responsive-design-%5B6689%5D?type=design&node-id=511%3A11482&mode=dev) defining the behavior for inline vs center.
> 
> Quick summary though: buttons + scrollable function would only come into play on `layout="inline"`. Center layout should continue to fill the container with Tab Titles simply getting more and more spread out or more and more tight as the component grows and shrinks. This means `layout="inline" would be the better layout option for larger amounts of Tabs.

I see what you mean.

Regarding the `center` layout:
I think what I was thinking is that if there are too many tabs, or tabs have very long titles, the ones that don't fit in the viewport will just be overflowing on both sides of the `center` layout. Figma snippet currently doesn't solve for truncation, so tab min-width would just allow for very long titles, contributing to the issue. So I was solving for the `center` by imagining an overflow.

I gather now that's not what the intention is, but I think maybe this needs to be addressed in Figma somehow. The current approach implies that we limit the number of tabs that can be added, truncated, or not.

Some additional questions: 
1. How does `inline/center` behave with really long tab titles in narrow tabs?
2. How will a user scroll back to tabs after clicking on the right arrow? 

> I imagine when overflow is to the right, the arrow chevron is pointing to the right. And once we click or scroll all the way to the right, and the overflow is to the left end, then the arrow chevron is flipped to the left so we can click back to the 1st tab.

4. There's a configuration where inline/center might blend, would users be confused?
5. Would lack of tab title truncation be an issue or do we need it?

jcfranco:
[Belated] I'll be picking this up and will close in favor of https://github.com/Esri/calcite-design-system/pull/8616. Thanks @Elijbet!

Issue #5689, opened 2022-11-04
Unable to use `dispatchEvent()` on the `dropdown-trigger` to open/close a dropdown


nwhittaker:
### Actual Behavior

The dropdown does not respond to keyboard events invoked with [`dispatchEvent()`][1] on the `dropdown-trigger` slot.

[1]: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent

### Expected Behavior

The dropdown does respond to these keyboard events.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/oNyLQwr

### Reproduction Steps

In codepen, dispatched events don't seem to trigger the listeners on custom elements, for some reason. So the code sample may need to be recreated and run locally.

1. Visit the code sample
2. Click the **"Enter" keydown** button and notice the dropdown does not open

### Reproduction Version

beta.97

### Relevant Info

The `keyDownHandler` is bailing early here because `target` is the `dropdown-trigger` element, and `this.referenceEl` is the `.calcite-dropdown-trigger-container` element:

https://github.com/Esri/calcite-components/blob/90e9831bdb8a29e2ce8925ef3464549c1295882b/src/components/dropdown/dropdown.tsx#L554

### Impact

This issue makes it difficult to write tests asserting keyboard events propagate correctly to a dropdown.

### Esri team

ArcGIS Field Apps

driskull:
@nwhittaker why do you need to dispatch the event? Can you just set the dropdown to `open`?

nwhittaker:
I'm dispatching the event in a test because I'm interested in asserting the event made it to the dropdown. Specifically, I have a dropdown inside a tree-item. In this scenario, if the dropdown is focused, keyboard events appear to have no effect or are consumed by the tree-item, unless their propagation is stopped.

If it's a high level of effort to fix, I could add my own handlers, but it seemed reasonable enough to expect the `open` property to have been toggled when dispatching the event on the trigger.

driskull:
Yeah this is tricky because the event listener is in the shadowRoot. To fix this, I think we would need to setup a composed event or refactor the dropdown to use a referenceElement approach.

geospatialem:
Verified in `1.9.0-next.2` with https://codepen.io/geospatialem/pen/YzdExqq

```html
<p>Simulate event on the trigger:
  <button type="button" onclick="keydown('Enter');">"Enter" keydown</button>
</p>

<calcite-dropdown>
[...]

PR #7752, opened 2023-09-15
fix(dropdown): support dispatching enter key event on dropdown trigger

driskull:
**Related Issue:** #5689

## Summary

- Fix keydown handler event to support dispatching events from any elements within the trigger slot
- Add test 
- Note: the codepen needs the slot changed from `dropdown-trigger` to `trigger`
- Note: the codepen needs the dispatched event to bubble.

### DIFF
--- a/packages/calcite-components/src/components/dropdown/dropdown.e2e.ts
+++ b/packages/calcite-components/src/components/dropdown/dropdown.e2e.ts
@@ -868,6 +868,7 @@ describe("calcite-dropdown", () => {
         expect(await dropdownWrapper.isVisible()).toBe(false);
 
         await page.$eval("calcite-button[slot='trigger']", (triggerEl: HTMLCalciteButtonElement) => {
+          // intentionally not pressing to avoid emitting `click`
           triggerEl.dispatchEvent(new KeyboardEvent("keydown", { key: "Enter", bubbles: true }));
         });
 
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/dropdown/dropdown.e2e.ts
```diff
@@ -847,6 +847,36 @@ describe("calcite-dropdown", () => {
...
+        const element = await page.find("calcite-dropdown");
+        const dropdownWrapper = await page.find(`calcite-dropdown >>> .calcite-dropdown-wrapper`);
+        const calciteDropdownOpen = await element.spyOnEvent("calciteDropdownOpen");
+        const calciteDropdownClose = await element.spyOnEvent("calciteDropdownClose");
+        const waitForCalciteDropdownOpen = page.waitForEvent("calciteDropdownOpen");
+
+        expect(await dropdownWrapper.isVisible()).toBe(false);
+
+        await page.$eval("calcite-button[slot='trigger']", (triggerEl: HTMLCalciteButtonElement) => {
+          triggerEl.dispatchEvent(new KeyboardEvent("keydown", { key: "Enter", bubbles: true }));
```
jcfranco, line 871: Could you add a comment here about only emitting `keydown` to avoid emitting `click` on the button?

jcfranco: If you wanted to stick to puppeteer APIs, I think you could also do the following here:

```tsx
const trigger = await page.find("calcite-button[slot='trigger']");
await trigger.callMethod("setFocus");
await page.keyboard.down("Enter"); // intentionally not pressing to avoid emitting `click`
```

Totally up to you.

jcfranco (approved):
👇⌨️🗯

packages/calcite-components/src/components/dropdown/dropdown.e2e.ts
driskull: I think I like the dispatch event better since it targets the issue more accurately and doesn't require any focusing.

PR #7753, opened 2023-09-16
chore: create a shared configs folder for monorepo

alisonailea:
**Related Issue:** None

## Summary

Following the example from turbo -> kitchensink. Let's create a separate location for shared configuration files


configs/tsconfig/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "tsconfig",
+  "version": "0.0.0",
+  "private": true,
+  "license": "MIT",
+  "publishConfig": {
+    "access": "public"
+  }
```
benelan, line 8: We can remove the `publishConfig` field

configs/eslint-config-custom/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "eslint-config-custom",
+  "license": "MIT",
```
benelan, line 3: Either remove the license field or point to the LICENSE.md file in the root of the repo

packages/calcite-components/package.json
```diff
@@ -75,14 +75,16 @@
     "sortablejs": "1.15.0"
   },
   "devDependencies": {
-    "@esri/calcite-design-tokens": "1.0.0",
+    "@esri/calcite-design-tokens": "*",
```
benelan, line 78: The packages in the monorepo that are published to NPM need to have the actual version here for `lerna`/`release-please`, rather than `*`. Packages that aren't versioned/published (like the configs) can use `*`.

benelan (commented):
IIRC when I briefly tried this I ran into some issues with the typescript parser and lintstaged. Did you hit and resolve those issues?

alisonailea:
> IIRC when I briefly tried this I ran into some issues with the typescript parser and lintstaged. Did you hit and resolve those issues?

I was able to fix the issues with typescript and eslint however I'm seeing a lot of SCSS errors when I run stylelint

benelan:
> I was able to fix the issues with typescript and eslint

Nice!!

> however I'm seeing a lot of SCSS errors when I run stylelint

Looks like the tests for `packages/eslint-plugin-calcite-components` are failing too :sob:

alisonailea:
> > I was able to fix the issues with typescript and eslint
> 
> Nice!!
> 
> > however I'm seeing a lot of SCSS errors when I run stylelint
> 
> Looks like the tests for `packages/eslint-plugin-calcite-components` are failing too 😭

oops missed a tsconfig in packages/eslint-plugin-calcite-components/tests/tsconfig. 
Fixed!

benelan (commented):
Is the typescript parser is supposed to cause eslint errors when there are typescript problems? When I was testing, lint-staged still allows commits when there are typescript errors that will break the build.

Issue #6673, opened 2023-03-27
[Alert] Responsive layout


geospatialem:
### Description

Mobile responsiveness across Calcite Design System, focused on web (traditionally desktop sized web browser), tablet, phone screen sizes, and large monitors.

### Acceptance Criteria

- [ ] 1. Consider breaking changes for folks. Consultation should occur on both the design and development teams.
- [ ] 2. Components need to be designed for the following breakpoints:
    - [ ] Small (s), `476px`
    - [ ] Medium (m), `768px`
    - [ ] Large (l), `1152px`
    - [ ] Extra larger (xl), `1440px`

### Relevant Info

Part of epic https://github.com/Esri/calcite-components/issues/6670

### Which Component

`alert`

### Priority impact

p4 - not time sensitive

### Esri team

Calcite (design)

SkyeSeitz:
Update from team discussion today:
@rmstinson presented Alert footer iterations for smaller screen sizes and mobile devices, which we will put on hold to first finalize the [Alert stacking enhancement](https://github.com/Esri/calcite-components/issues/2835) currently in progress.

<img width="1061" alt="image" src="https://user-images.githubusercontent.com/29716057/232863479-3c9981be-2fa2-4ad7-adea-3a4c1f8d8cf0.png">

rmstinson:
It looks like I can't assign myself to this issue @SkyeSeitz but at least I can comment now!

geospatialem:
Just added 'ya @rmstinson 👍🏻

rmstinson:
Compact Alerts have been built - they're intended for use at a width of 767px and lower. 

Building them also presented an opportunity to streamline the regular Alert component - currently we have a variant for autodismiss, and a variant for default. However, it's possible to build the component with a show/hide boolean for the Progress component, so that we can have a configuration option for autodismiss, and cut the number of variants in the component in half.

Take a look here: https://www.figma.com/file/eradSav1w0c4viNljdROKk/Alert---responsive-design-%5B6673%5D?type=design&node-id=517%3A12655&mode=design&t=eM0Y3CNz80p4B0wH-1

![Screenshot 2023-06-28 at 4 01 30 PM](https://github.com/Esri/calcite-components/assets/12650705/51fbeb2c-f16e-4de3-9aaa-91e9670423ab)

If we're happy with this approach I can change the rest of the Alerts as well. This might have ramifications for the Notice component as well @ashetland

SkyeSeitz:
[...]

PR #7755, opened 2023-09-16
feat(alert): make component responsive

jcfranco:
**Related Issue:** #6673

## Summary

Updates alert to be responsive per the design spec.

## Additional changes

* simplified class names (`alert` prefix is no longer necessary)
* simplified styles by fixing issues reported by [`no-descending-specificity`](https://github.com/stylelint/stylelint/blob/main/lib/rules/no-descending-specificity/README.md) stylelint rule
* `requestedIcon` is now computed at render time to simplify component 
* refactored class names (e.g., used BEM-like convention to state/modifier classes) and applied them according to conventions
* clean up test story


packages/calcite-components/src/components/alert/alert.scss
```diff
@@ -6,11 +6,162 @@
...
+  pointer-events-none
+  shadow-2
+  text-start
+  w-full
+  z-toast;
+
+  container: responsive-container / inline-size;
+  border-radius: var(--calcite-border-radius);
+  border-block-start: 0 solid transparent;
+  border-inline: 1px solid var(--calcite-ui-border-3);
```
driskull, line 34: I wonder if these border px values should come from a token?

packages/calcite-components/src/components/alert/alert.scss
```diff
@@ -6,11 +6,162 @@
...
+  border-block-end: 1px solid var(--calcite-ui-border-3);
+  inline-size: var(--calcite-alert-width);
+  max-inline-size: calc(100% - (var(--calcite-alert-edge-distance) * 2 + 2px));
```
driskull, line 37: What is the `2px` here from? can that come from a token?

packages/calcite-components/src/components/alert/alert.scss
```diff
@@ -6,11 +6,162 @@
...
+  }
+}
+
+.dismiss-progress {
+  @apply absolute
+  block
+  w-full
+  overflow-hidden;
+  inset-inline: 0;
+  inset-block-start: -2px;
```
driskull, line 125: For these 2px values maybe a sass variable or token would be nice

packages/calcite-components/src/components/alert/alert.tsx
```diff
@@ -257,41 +229,107 @@ export class Alert implements OpenCloseComponent, LoadableComponent, T9nComponen
...
+    );
+  }
+
+  private renderIcon(icon: string): VNode {
+    return (
+      <div class={CSS.icon}>
+        <calcite-icon
+          flipRtl={this.iconFlipRtl}
+          icon={icon}
+          scale={this.scale === "l" ? "m" : "s"}
```
driskull, line 327: Maybe a function for setting this scale value? This logic appears twice.

driskull (approved):
👍 💯 🥇

packages/calcite-components/src/components/alert/alert.scss
jcfranco: Agreed on using internal CSS props or Sass vars for these. @macandcheese could you help us out with this?

packages/calcite-components/src/components/alert/alert.tsx
jcfranco: I like the idea of having a util for internal icon scale adjustment. Especially, since it's sprinkled across many components. These seem to be the only exceptions I found:

* https://github.com/Esri/calcite-design-system/blob/8a575e451acb6b33d65124988b2e2f7ea8ec8b81/packages/calcite-components/src/components/popover/popover.tsx#L545
* https://github.com/Esri/calcite-design-system/blob/796bf90cacd68595a108f39e2a66bcc5d71de0e9/packages/calcite-components/src/components/modal/modal.tsx#L300

@ashetland @SkyeSeitz All internal icons scales should be following this (`m` icon for `scale="l"` and `s` icon for `scale="s"` + `scale="m"`) , right?

ashetland: Yep, that's correct.

packages/calcite-components/src/components/alert/alert.scss
macandcheese: Hm, this is an older one, maybe to compensate for the 1px border on either side? May not be necessary anymore.

jcfranco:
@macandcheese @ashetland @SkyeSeitz Could you review the [visual changes](https://www.chromatic.com/build?appId=6266d45509d7eb004aa254fb&number=2492)? 🙇

packages/calcite-components/src/components/alert/alert.scss
jcfranco: Doesn't seem to be necessary. Removed!

packages/calcite-components/src/components/alert/alert.scss
jcfranco: I stored them in a sass variable for now to DRY things up w/o adding a CSS prop.

packages/calcite-components/src/components/alert/alert.scss
jcfranco: Maybe we can do this in a follow-up PR. I'm not entirely sure if all 2px values represent the same thing. 🤔

driskull: Yep sounds good 👍

PR #7747, opened 2023-09-15
chore: disable eslint rule "jsdoc/require-returns-type"

driskull:
**Related Issue:** None

## Summary

Disables ESlint rule ["jsdoc/require-returns-type"](https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/require-returns-type.md#readme) since we rely on TypeScript.

This builds upon https://github.com/Esri/calcite-design-system/pull/7326.

PR #7756, opened 2023-09-17
ci(eslint): disable jsdoc/require-property-type

benelan:
**Related Issue:** #7747, #7326

## Summary

Disable [`jsdoc/require-property-type`](https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/require-property-type.md) in favor of using Typescript.


driskull (approved):
👍

Issue #6821, opened 2023-04-21
Tab selection underline is not aligned when icon is added to tab


mpayson:
### Actual Behavior

When an icon is added to another tab, the selection underline for the current tab title is not aligned to the title content

<img width="215" alt="Screenshot 2023-04-21 at 7 21 03 AM" src="https://user-images.githubusercontent.com/5892789/233623749-4fd2e480-13b9-4058-b457-558e3b480782.png">

### Expected Behavior

The selection underline adjusts to align with the title content

### Reproduction Sample

https://codepen.io/mpayson/pen/YzJpbpe

### Reproduction Steps

1. Open sample
2. Click "Add indicator"
3. See underline is not aligned

### Reproduction Version

1.2.0

### Relevant Info

Similar to https://github.com/Esri/calcite-components/issues/4896, but this issue occurs when the indicator is added to a tab other than the selected tab

### Priority impact

p3 - want for upcoming milestone

### Esri team

ArcGIS Online

macandcheese:
cc @ashetland @SkyeSeitz regarding that conversation about this indicator element / animation.

geospatialem:
Verified in `1.9.0-next.5`

![image](https://github.com/Esri/calcite-design-system/assets/5023024/c7a94389-ea10-427e-ab08-728a01648483)

PR #7768, opened 2023-09-19
fix(tab-nav): update indicator position when tab icon changes

driskull:
**Related Issue:** #6821

## Summary

- Updates offset position when tab icon is added
- Add screenshot test


alisonailea (approved):
👍🏻

Issue #4032, opened 2022-02-01
calcite-select: Allow having no initial value


jgibson02:
### Description

It is sometimes useful to have no initial value for a `select` to coerce the user to make an explicit decision. Currently this seems to not be supported by `calcite-select` since it seems to always be initialized with the first `calcite-option` child's value if `value` is undefined and no `calcite-option` is `selected`. In plain HTML, `<option hidden disabled />` might be used, however doing `<calcite-option hidden disabled />` still causes it to be visible in the menu.

### Acceptance Criteria

- [ ] When no `value` is set, and no `calcite-option` is selected, the `calcite-select` renders its placeholder and does not initialize itself with a default value

### Which Component

calcite-select

### Example

https://codepen.io/arjan_esrinl/pen/ZEmgaVv

macandcheese:
This one's tough as we match native behavior and there is no native placeholder attribute.

The single-select combobox could accomplish this:

```
<calcite-combobox placeholder="Select a value" selection-mode="single">
    <calcite-combobox-item value="Engineering" text-label="Engineering"></calcite-combobox-item>
    <calcite-combobox-item value="Architecture" text-label="Architecture"></calcite-combobox-item>
    <calcite-combobox-item value="Construction" text-label="Construction"></calcite-combobox-item>
</calcite-combobox>
```

Although the above [does not allow you to clear selection](https://github.com/Esri/calcite-components/issues/3074) - not sure if that is desirable for your workflow.

jgibson02:
@macandcheese what're your thoughts on supporting `hidden` on a `calcite-option` to hide it from the list and act like a placeholder akin to the native behavior?

macandcheese:
That would work - it seems like that works only when `hidden` is the first option. Maybe we could just render that internally based on the presence of an (added) placeholder attribute? Would be open to either option there.

jgibson02:
If it only works for the first option, then your suggestion certainly sounds cleaner to me rather than adding a `hidden` prop and depending on documenting that behavior.

ReneU:
+1 We need this to allow having no initial value for some components in ArcGIS Velocity. We have some required fields that don't allow for good defaults such as this one:
![image](https://user-images.githubusercontent.com/5748802/176642918-c15ebd63-8777-4f10-8138-bb9b5a17d5b9.png)

ReneU:
[...]

PR #7769, opened 2023-09-19
fix(select): allow setting an option value to an empty string

jcfranco:
**Related Issue:** #4032

## Summary

This fixes an issue where an empty string value was treated as value not set and would therefore fall back to the label as the value.


alisonailea (approved):
👍🏻

Issue #6692, opened 2023-03-28
[Checkbox] Responsive Layout


brittneytewks:
### Description

Responsiveness across Calcite Design System, focused on web (traditionally desktop sized web browser), tablet, phone screen sizes, and large monitors.

### Acceptance Criteria

1. Consider breaking changes for folks. Consultation should occur on both the design and development teams.
2. Components need to be designed for the following breakpoints:

- [ ]  Small (s), 476px
- [ ]  Medium (m), 768px
- [ ]  Large (l), 1152px
- [ ]  Extra larger (xl), 1440px

### Which Component

`checkbox`

### Priority impact

p4 - not time sensitive

### Esri team

Calcite (design)

SkyeSeitz:
Update from team discussion today:
Checkbox needs a size adjustment on the large scale for use on small screen and mobile devices. The following proposal is to change the large 16px Checkbox to 18px with a 44px hit target.
<img width="297" alt="image" src="https://user-images.githubusercontent.com/29716057/236292391-f454d671-f426-4ca1-8725-19b4df1e3c32.png">
[Link to specs](https://www.figma.com/file/zBGnAOEOxqKGocUjqiYbed/Responsive-Layout?node-id=1616%3A113187&t=G1UGmY26oFahPopH-1)

SkyeSeitz:
Updated [Figma file](https://www.figma.com/file/RB5r1PlfjOxRBgx4aA00cq/Checkbox---responsive-%5B6692%5D?type=design&node-id=1%3A2&mode=design&t=ynQz9Cciv1wv04SD-1) with AA hit target at 24px 🚀

geospatialem:
Verified in `1.9.0-next.7` with `checkbox` and `card` when `selectable`, where the differing sizing of `--calcite-checkbox-size` will adjust to the `24px` hit target at a minimum.

PR #7773, opened 2023-09-19
feat(checkbox): add WCAG AA recommended minimum 24px square hotspot optimized for touch users

eriklharper:
**Related Issue:** #6692 

## Summary

This adds an invisible 24px hotspot for pointer and touch clicks around the checkbox for all scales.  If the `--calcite-checkbox-size` custom property is modified to be larger than 24px, this hotspot won't prevent the clickable area from expanding to the new size; it will just serve to enlarge the clickable area when the size is less than the WCAG AA recommended minimum size of 24px.


packages/calcite-components/src/components/checkbox/checkbox.scss
```diff
@@ -58,13 +58,23 @@
...
   &:focus,
   &:focus-visible {
     @apply focus-outset;
   }
+
+  &::after {
+    content: "";
+    inset-block-start: 50%;
+    inset-inline-start: 50%;
+    min-block-size: 24px;
```
driskull, line 73: I think you can use `min-block-size: theme("width.6")` so that it uses tokens.

driskull (changes requested):
Looks good, can we just use the design tokens?

packages/calcite-components/src/components/checkbox/checkbox.scss
eriklharper: I'm wondering since `width.6` evalutes to a `rem` value if the 24px AA minimum could go under if the page's default font-size is reduced far enough.  Maybe this is something we document as a best practice not to do?

driskull: Yeah thats a good point. Ill defer to others judgement. We could maybe store a const somewhere else for 24px minimum. Ideally, this 24px is stored somewhere in either a global CSS custom prop or sass var.

jcfranco (approved):
Awesome! 

Since this doesn't affect the hit area visually, can you add a single, new E2E test that asserts clicks on the updated hit area across all scales and the custom sizing case?

driskull (approved):
👍 Looks good!

PR #7855, opened 2023-09-20
feat(flow, flow-item): Allow calciteFlowItemBack event to be cancelled

driskull:
**Related Issue:**#7869

## Summary

- Supports the cancelling of the `calciteFlowItemBack` event via `preventDefault()`
- Changes event to be cancellable
- Adds test

### DIFF
--- a/packages/calcite-components/src/components/flow-item/flow-item.tsx
+++ b/packages/calcite-components/src/components/flow-item/flow-item.tsx
@@ -168,7 +168,7 @@ export class FlowItem
   /**
    * Fires when the back button is clicked.
    */
-  @Event({ cancelable: false }) calciteFlowItemBack: EventEmitter<void>;
+  @Event({ cancelable: true }) calciteFlowItemBack: EventEmitter<void>;
 
   /**
    * Fires when the content is scrolled.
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/flow-item/flow-item.tsx
```diff
@@ -168,7 +168,7 @@ export class FlowItem
   /**
    * Fires when the back button is clicked.
    */
-  @Event({ cancelable: false }) calciteFlowItemBack: EventEmitter<void>;
+  @Event({ cancelable: true }) calciteFlowItemBack: EventEmitter<void>;
```
driskull, line 171: @jcfranco I wonder if we should have a common test helper to test all the custom events on a component to make sure that an event doesn't switch from `cancelable: true` to `cancelable: false` as that would be a breaking change.

I'm thinking kind of like a `defaults` test to make sure events are setup as they should be.

jcfranco: This would be a great helper, but would be hard to generalize since we would need to still pass a function that triggers and asserts the canceled behavior. The actual event being canceled is half of what this helper would need to assert.

jcfranco (approved):
Awesome!

⬅️✋🚫

jcfranco:
Sorry for not catching this sooner, but can you create an issue for this feature so that it goes through verification? cc @geospatialem

packages/calcite-components/src/components/flow-item/flow-item.tsx
driskull: Yeah, might be overkill. Just a thought.

driskull:
created #7869

Issue #7100, opened 2023-06-02
[List], [List Item], [List Item Group]: add scales, update padding and spacing, update font sizes


ashetland:
### Description

Add scales and related updates to List, List Item, and List Item Group. See this [Figma file](https://www.figma.com/design/91sfHrYNt0Il6B4pEZZ749/List-scales%2C-sorting%2C-groups%2C-and-interaction?node-id=982-37071&m=dev&focus-id=1066-43440) for full specs. Note that file also contains other related updates to List.

### Acceptance Criteria

- [x] Add scales
- [x] Update padding and spacing
- [x] Update font sizes
- [x] Align bottom border to content area
- [x] Revised hit targets to meet AA minimum
- [x] Add default padding to content with CSS vars for overrides - 🚧 `breaking change` ❗ 
- [x] Use inner-shadow for `selection-appearance="border"` to maintain alignment

### Relevant Info

These are updated designs since this issue was originally created. They prioritize reducing unnecessary whitespace around elements to conserve horizontal real estate, 24px AA minimum hit targets, and improving visual hierarchy when nesting.

Pair with #9173, #9174, and #9175

Full specs available in [Figma file](https://www.figma.com/design/91sfHrYNt0Il6B4pEZZ749/List-scales%2C-sorting%2C-groups%2C-and-interaction?node-id=982-37071&m=dev&focus-id=1066-43440), before/after reference below:
![Slice 1](https://github.com/user-attachments/assets/b7c92582-840c-4135-8c11-4b34aea299fc)

### Which Component

List
List Item
List Item Group

### Priority impact

p3 - want for upcoming milestone

### Esri team

Calcite (design)

driskull:
We may wan to reconsider how we indent child items as well. Currently, we only indent within the item but not the whole item.

<img width="310" alt="image" src="https://github.com/Esri/calcite-design-system/assets/1231455/d619ba44-11d1-481b-8cd5-a6188177c145">

The problem with this is that the selection is still far to the side and when dragging a child there isnt a good way to distinguish where its going because they are all indented the same. No matter how deep the child is, the element spans the whole width. 

I think we did this originally so that the hover/active/focus spans the whole width of the list but I think we should reconsider.

<img width="493" alt="image" src="https://github.com/Esri/calcite-design-system/assets/1231455/a0ded77d-e94d-4512-bd48-fec3500df173">

driskull:
[...]

Issue #6700, opened 2023-03-29
Have a clearer unselected state in Calcite List 


CodeBarker:
### Description

We've had users share that the unselected state of the calcite list doesnt immediately communicate it is selectable.  When one or more items is selected by default it is not as critical, However for lists where there is no default selection it just looks like empty space and users are confused

#### Current 
<img width="431" alt="image" src="https://user-images.githubusercontent.com/3944714/228623592-0f83a54c-c26f-4a19-ba37-d153450e2c01.png">

#### Desired
<img width="412" alt="image" src="https://user-images.githubusercontent.com/3944714/228623989-0c4e5b2c-3891-46a2-9259-543a834f5618.png">

### Which Component

calcite list / calcite list item 

### Priority impact

p3 - want for upcoming milestone

### Esri team

ArcGIS Map Viewer

macandcheese:
@asangma Agreed, either using the circle-empty + check-circle-f or square and square-check-f here would clear this up.

asangma:
@macandcheese I could also see something like an `appearance="minimal"` or something that would mimic PickList's selection indicators but with an added hover state that shows the check or dot. I can add mockups if that's helpful.

asangma:
💯

asangma:
And yeah...the spacing needs some love. :)

asangma:
Suggested spacing between the selection icon and the text:
Use the same spacing that Button has between its content and icon...and maaaaybeee that spacing could be a little tighter (0.5rem for medium)? 😬

hcampos-esri:
Another feedback we got, specifically with regards to the `selection-appearance="border"` was that the focus highlight and the border can cause confusion. When the focus outline is visible, it seems as though the list item is selected.
![1a7938f1-da23-4b3f-b404-5fb74ee86225](https://github.com/Esri/calcite-design-system/assets/136828178/b60a0ce0-3d75-4fad-88e7-a2efbc1da8ae)
![dbd3d086-456e-4656-bf55-9f8da73e333b](https://github.com/Esri/calcite-design-system/assets/136828178/34c09e98-5802-44cd-b6cd-0c9d83a84363)

Anastasiia-Boleiko:
Hi there👋
Also, some feedback regarding the List component😉. 
[...]

PR #7862, opened 2023-09-21
fix(list-item): Make width of caret icon consistent with selection icon

driskull:
**Related Issue:** None

## Summary

- Currently, the hit area for the expand/collapse of a list item is a bit small. (18px)
- This adds padding consistent with the selection icon

## Current

image

## New

image

### DIFF
--- a/packages/calcite-components/src/components/list-item/list-item.scss
+++ b/packages/calcite-components/src/components/list-item/list-item.scss
@@ -104,8 +104,12 @@ td:focus {
 }
 
 .selection-container {
-  @apply flex px-3;
+  @apply px-3;
   color: var(--calcite-list-item-icon-color);
+}
+
+.open-container {
+  @apply px-3;
 }
 
 .actions-start,


jcfranco (approved):
Code changes LGTM!

@SkyeSeitz @ashetland Could you give this a looksie? 🔍👁️‍🗨️👄👁️‍🗨️

ashetland:
This seems to conflict a bit with what I put together for issue #7100 slated for December ([Figma file](https://www.figma.com/file/LV1M8Hjoc2aD5zBr61XCa2/Nested-Indent-Spacing-%5Bepic-6632%5D?type=design&node-id=308-128306&mode=dev)).

driskull:
This can wait for December if necessary.

Looking at the design, it would be nice if the design included the hit area for the caret, selection, etc. It looks like its just putting padding in-between the icons but not really showing each icon's hit area. Would the icon include 6px padding on each side?

driskull:
I guess for this release I was hoping to improve the hit area of the caret without messing with the other paddings. Can we do that?

ashetland:
> It looks like its just putting padding in-between the icons but not really showing each icon's hit area.

Great catch! Yep, those designs will be revised to include hit areas. I actually started looking into that for #6700, but stopped because it wasn't relevant to the issue 🤦‍♂️. _That_ [Figma file](https://www.figma.com/file/d7UBycO6x1Vz8zskFj5dTI/Epic---update-icons-for-system-consistency-%5B7181%5D?type=design&node-id=1004-10047&mode=dev) also calls for changing the caret to a chevron for system consistency. 

Due to all the related issues, it _might_ be worth holding off for now? Merging this would up the space between the chevron and the next element to 2rems which is a lot and doesn't improve the visual hierarchy when nesting.

ashetland:
> Would the icon include 6px padding on each side?

That sounds right, but I recall some of the paddings getting tweaked depending on whether or not there was nesting.

driskull:
Yeah lets just close this and wait for december.

PR #7867, opened 2023-09-22
feat(action-menu): Add appearance property to configure trigger appearance

driskull:
**Related Issue:** None

## Summary

- Adds `appearance` property to `calcite-action-menu`.
- Allows menu to configure the default action's appearance
- Adds test
- Adds screenshot tests

### DIFF
--- a/packages/calcite-components/src/components/action-menu/action-menu.stories.ts
+++ b/packages/calcite-components/src/components/action-menu/action-menu.stories.ts
@@ -20,6 +20,24 @@ export const simple = (): string =>
     </calcite-action-menu>
   `;
 
+export const simpleSolid_TestOnly = (): string =>
+  html`<div style="background-color:#99ccff">
+    <calcite-action-menu>
+      <calcite-action text="Plus" icon="plus" text-enabled></calcite-action>
+      <calcite-action text="Minus" icon="minus" text-enabled></calcite-action>
+      <calcite-action text="Table" icon="table" text-enabled></calcite-action>
+    </calcite-action-menu>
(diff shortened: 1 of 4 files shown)


packages/calcite-components/src/components/action-menu/action-menu.stories.ts
```diff
@@ -20,6 +20,24 @@ export const simple = (): string =>
...
   `;
 
+export const simpleSolid_TestOnly = (): string =>
+  html`<div style="background-color:#99ccff">
+    <calcite-action-menu>
+      <calcite-action text="Plus" icon="plus" text-enabled></calcite-action>
+      <calcite-action text="Minus" icon="minus" text-enabled></calcite-action>
+      <calcite-action text="Table" icon="table" text-enabled></calcite-action>
+    </calcite-action-menu>
+  </div> `;
```
benelan, line 30: Does the background color need to be `#99ccff` for this test? If not the `simple` one above should cover the solid use case so we should be able to remove this one.

benelan (approved):
:+1:

packages/calcite-components/src/components/action-menu/action-menu.stories.ts
benelan: Woah, TIL GitHub renders a little color icon next to color codes

driskull: Yeah the background needs to be set to something to test that the transparent appearance is working correctly. I can switch it to just `red` to be a little more obvious.

PR #7873, opened 2023-09-25
fix(action-bar): overflow actions when overflowActionsDisabled is set to false

driskull:
**Related Issue:** #7891

## Summary

- When overflowActionsDisabled is set to `false` from `true`, actions should be overflowed again.
- Adds test


packages/calcite-components/src/components/action-bar/action-bar.tsx
```diff
@@ -114,9 +114,13 @@ export class ActionBar
 
   @Watch("overflowActionsDisabled")
   overflowDisabledHandler(overflowActionsDisabled: boolean): void {
-    overflowActionsDisabled
-      ? this.resizeObserver.disconnect()
-      : this.resizeObserver.observe(this.el);
+    if (overflowActionsDisabled) {
+      this.resizeObserver.disconnect();
```
jcfranco, line 118: For consistency, can you add an optional chaining operator to observers?

packages/calcite-components/src/components/action-bar/action-bar.e2e.ts
```diff
@@ -368,6 +368,43 @@ describe("calcite-action-bar", () => {
       expect(await page.findAll(slottedActionsSelector)).toHaveLength(7);
     });
 
+    it("should slot 'menu-actions' on sublist changes after being enabled", async () => {
+      const page = await newE2EPage({
```
jcfranco, line 372: In an upcoming PR, all tests will be creating the E2E page and then setting the content. Would you mind updating this test to follow that pattern?

jcfranco (approved):
🚀🪐✨

Awesome!

Before merging, can you create an issue for this so that it goes through verification? cc @geospatialem

PR #7874, opened 2023-09-25
feat(list): add newIndex and oldIndex event details to calciteListOrderChange event

driskull:
**Related Issue:** #7875

## Summary

- Add more useful event details to `calciteListOrderChange` event
- Adds `oldIndex` and `newIndex` properties to the event detail
- Adds test

### DIFF
--- a/packages/calcite-components/src/components/list/list.e2e.ts
+++ b/packages/calcite-components/src/components/list/list.e2e.ts
@@ -6,6 +6,7 @@ import { debounceTimeout } from "./resources";
 import { CSS } from "../list-item/resources";
 import { DEBOUNCE_TIMEOUT as FILTER_DEBOUNCE_TIMEOUT } from "../filter/resources";
 import { GlobalTestProps, dragAndDrop, isElementFocused } from "../../tests/utils";
+import { DragDetail } from "../../utils/sortableComponent";
 
 const placeholder = placeholderImage({
   width: 140,
@@ -474,6 +475,8 @@ describe("calcite-list", () => {
 
     type TestWindow = GlobalTestProps<{
       calledTimes: number;
+      newIndex: number;
+      oldIndex: number;
     }>;
 
     it("works using a mouse", async () => {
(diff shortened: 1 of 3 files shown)


benelan (commented):
What do you think about `previousIndex` and `currentIndex`? It would line up with the native event's `currentTarget`. Unless we have some precedent for new/old versus current/previous.

driskull:
I prefer old/new vs previous/current because `current` can be a bit ambiguous.

driskull:
Event has `originalTarget` but that property is non-standard. `previous` would go better with `next`.

packages/calcite-components/src/components/list/list.e2e.ts
```diff
@@ -616,24 +631,30 @@ describe("calcite-list", () => {
 
       let totalMoves = 0;
 
-      const eventSpy = await page.spyOnEvent("calciteListOrderChange");
+      // Workaround for page.spyOnEvent() failing due to drag event payload being serialized and there being circular JSON structures from the payload elements. See: https://github.com/Esri/calcite-design-system/issues/7643
```
jcfranco, line 634: Nice!

packages/calcite-components/src/components/list/list.e2e.ts
```diff
@@ -482,8 +485,12 @@ describe("calcite-list", () => {
       // Workaround for page.spyOnEvent() failing due to drag event payload being serialized and there being circular JSON structures from the payload elements. See: https://github.com/Esri/calcite-design-system/issues/7643
       await page.$eval("calcite-list", (list: HTMLCalciteListElement) => {
         (window as TestWindow).calledTimes = 0;
```
jcfranco, line 487: Suggestion: can you store the typed window in a local `win` variable or similar to reduce typecasting lines?

packages/calcite-components/src/utils/sortableComponent.ts
```diff
@@ -5,6 +5,8 @@ export interface DragDetail {
   toEl: HTMLElement;
```
jcfranco, line 5: Sidebar: we should probably add a conventions entry about meaningful event details and when they should be allowed. This is a perfect example of how this info helps add context behind an event that would be hard to represent in another way.

jcfranco (approved):
Awesome! ✨👶📇👴📇 ✨

packages/calcite-components/src/utils/sortableComponent.ts
driskull: Yes a convention document on this would be super useful!

PR #7876, opened 2023-09-25
ci(chromatic): fix node v18 SSL issue and remove dependabot/markdown skip conditions

benelan:
**Related Issue:** #7871

## Summary

We bumped Node to v18 because v16 has been deprecated. Chromatic/storybook is failing because they are still using Node v16 which uses a legacy SSL provider. We can use the [`--openssl-legacy-provider`](https://nodejs.org/dist/latest-v18.x/docs/api/cli.html#--openssl-legacy-provider) flag to resolve the issue.

Additionally, we don't need the dependabot conditions and the check for markdown-only PRs now that we use renovatebot and have a label for skipping visual snapshots.


packages/calcite-components/src/components/accordion/accordion.stories.ts
```diff
@@ -187,7 +187,7 @@ export const darkModeRTL_TestOnly = (): string =>
     `
   );
 
-darkModeRTL_TestOnly.parameters = { modes: modesDarkDefault };
+darkModeRTL_TestOnly.parameters = { modes: modesDarkDefault, chromatic: { disableSnapshot: false }, };
```
anveshmekala, line 190: nitpick: is this added to test the option ? if yes, can we revert this change.

anveshmekala (approved):
LGTM! 👍

PR #7880, opened 2023-09-26
build(node): bump npm version and pin node types to v18

benelan:
**Related Issue:** #7695

## Summary

- Pin the Node types to v18.x.x so we don't get incorrect info
- Prevents Renovate from trying to update the Node types
- Bump NPM to the version that ships with Node v18


jcfranco (approved):
✨🧷✨

PR #7904, opened 2023-09-28
fix(tree): avoid modifying selected items for "none" selection-mode

jcfranco:
**Related Issue:** #7909

## Summary

`selectedItems` should no longer be updated when selection mode is `none`.


packages/calcite-components/src/components/tree/tree.tsx
```diff
@@ -235,7 +235,7 @@ export class Tree {
     }
 
     this.selectedItems = isNoneSelectionMode
-      ? [target]
```
driskull, line 238: 💥

driskull (approved):
💯 🦅

Issue #7903, opened 2023-09-28
[Table] Flip border selection on RTL


SkyeSeitz:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The RTL border selection appearance is not flipped and sits on the inside edge of the Table row.
<img width="809" alt="image" src="https://github.com/Esri/calcite-design-system/assets/29716057/6548d1b2-50d1-42bd-a0f4-b16a81d5887d">

### Expected Behavior

When in RTL, the border selection appearance should be flipped to the outside edge of the Table row.
<img width="812" alt="image" src="https://github.com/Esri/calcite-design-system/assets/29716057/827644fb-d3c6-49c8-92ee-e23f0ff2671f">

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/table/

### Reproduction Steps

1. Open Table sample
2. Switch to RTL
3. Change selection-mode prop to single or multi
4. Select row in sample

### Reproduction Version

1.8.0

### Priority impact

p4 - not time sensitive

### Impact

This chance would match the intended design for both LTR & RTL users.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

geospatialem:
Verified in `1.9.0-next.16` when the `html`, `body`, or component are set to `dir="rtl"`:

![image](https://github.com/Esri/calcite-design-system/assets/5023024/02f30334-d375-41e2-9d67-8fab39695e70)

PR #7907, opened 2023-09-29
fix(table): Fix selection display in RTL

macandcheese:
**Related Issue:** #7903 

## Summary
Fixes the RTL display of selection indicator in Table Cell. Note, since this is set in render method, like other instances, this won't "update" until the component is interacted with..'

Screenshot 2023-09-29 at 10 31 23 AM

Screenshot 2023-09-29 at 10 31 30 AM


driskull (approved):
Looks good with the caveat you mentioned. Should there be a screenshot test for this?

Issue #7910, opened 2023-09-29
[Action Bar/Action Menu] active actions in action bar are not active in collapsed menu


gopal-y:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When active property of calcite-action is being used and when the container is resized, the calcite actions are moved to menu-actions and are not shown as active. This can be observed when more than one calcite-action is active.

### Expected Behavior

If more than one calcite-action item is active, it should be reflected in the collapsed menu-items as well.

### Reproduction Sample

https://codepen.io/im-the-developer/full/qBLooXP

### Reproduction Steps

1. Open the code pen
2. select more than one calcite action in the first action-group
3. now drag the side shell-panel to reduce the width and trigger the overflow
4. now open the menu to see the text-enabled actions, only one or none of the calcite-actions reflect the active behavior

### Reproduction Version

1.4.2

### Priority impact

p1 - need for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

gopal-y:
https://github.com/Esri/calcite-design-system/assets/30383517/dac1aa03-1e4f-4539-86d1-6237bb18726e

driskull:
Valid. This is because the action menu uses the `active` property to highlight the currently active action. It should probably use another internal property to do this visual change because `active` can be defined by the user.

geospatialem:
Verified with `1.9.0-next.16`

![verify-7910](https://github.com/Esri/calcite-design-system/assets/5023024/97a2cf09-5653-4b43-b8cc-d86d725ec450)

PR #7911, opened 2023-09-29
fix(action-menu): update active selection to not use the action's active property

driskull:
**Related Issue:** #7910 

## Summary

- Use a data attribute instead of the action property for styling actively "focused" action within the action-menu.
- Updates test


packages/calcite-components/src/components/action-menu/action-menu.e2e.ts
```diff
@@ -237,17 +237,17 @@ describe("calcite-action-menu", () => {
       await page.waitForChanges();
 
       expect(await actionMenu.getProperty("open")).toBe(true);
-      expect(await actions[0].getProperty("active")).toBe(true);
-      expect(await actions[1].getProperty("active")).toBe(false);
-      expect(await actions[2].getProperty("active")).toBe(false);
+      expect(actions[0].getAttribute(activeAttr)).toBe("");
```
jcfranco, line 240: For later, but this might be better off as a screenshot test since this is now an internal concern.

packages/calcite-components/src/components/action-menu/action-menu.tsx
```diff
@@ -95,7 +95,7 @@ export class ActionMenu implements LoadableComponent {
   openHandler(open: boolean): void {
     this.activeMenuItemIndex = this.open ? 0 : -1;
     if (this.menuButtonEl) {
-      this.menuButtonEl.active = open;
+      this.menuButtonEl.toggleAttribute(activeAttr, open);
```
jcfranco, line 98: We typically rely on internal props for this instead of setting data attributes on the host. Could we make this consistent here or in a follow-up PR if pressed for time with the upcomin release?

jcfranco (approved):
🎉🦖🌮

packages/calcite-components/src/components/action-menu/action-menu.tsx
driskull: Yeah I thought about the internal prop but the internal prop wouldn't actually do anything on the action.

Thought about a class but didn't want to pollute action with a class incase users remove all classes

We should have a good best practice doc for this though

packages/calcite-components/src/components/action-menu/action-menu.e2e.ts
driskull: Screenshot tests will already capture the styles for this but this is a nice to test the keyboard navigation.

Issue #4326, opened 2022-03-29
[Combobox] Add an option to limit the amount of displayed chips


macandcheese:
### Description

The Combobox can often become crowded when many chips are selected. 

Propose to add a `limitChips`(name tbd) property that appends all additional chips after that numeric value to a non-dismissible "+n" chip.

Questions to answer:
- What chips are visible - is it decided by order entered, alphabetically, etc.
- Should there be an affordance for "no individual chips" - as in, "4 selected", 6 selected", "1 selected", etc.
   - If so, is that exclusive from this request? If not, how do they interact?
- What happens when the combobox is interacted with - do the "+3" chips appear in-combobox? in-dropdown? nowhere?

### Acceptance Criteria

As a user, I can choose a number of chips to display, after which, all selected items would be appended to a non-dismissible catch-all.

Also note of [Aaron's comment](https://github.com/Esri/calcite-design-system/issues/5226#issuecomment-1613918426)

For example, limit-chips="2" would result in: 

This would result in a display of <chip (1) ><chip 2 (x) ><+n>

### Relevant Info

MUI's "limit tags" https://mui.com/components/autocomplete/#limit-tags

_No response_

### Which Component

Combobox

### Example Use Case

My product offers a state-selection Combobox to a user. Often times, a user can select 5+ states, which clutters the combobox and causes my UI to break. I want to limit the amount of chips displayed.

LeviRemi:
Customer example where they dislike how other containers in the same row grow in height to accommodate the ever-expanding combobox. This layout also does not scale well for large quantities of selection. 
<img width="869" alt="Screen Shot 2022-03-29 at 2 20 27 PM" src="https://user-images.githubusercontent.com/21701076/160709136-b9d01749-887e-4d82-81ac-368b4785ccb5.png">

mMa2:
The displayed selected chips keep growing and occupy too much real estate, hide other components on the UI, it is too crowded for the user to see other items. Please consider options as 1)shrink the display text 2) have a textbox labelled down to show selected items, no displaying chips

ashetland:
Proposing a new `display-mode` prop that would address the need to limit the number of chips displayed while also making Combobox a more responsive component. 

Full designs and context with other issues from epic #6720 are available in [**this Figma file**](https://www.figma.com/file/ujZYHdimnVC3yA2PyYOH85/Epic%3A-Combobox-Redesign?type=design&node-id=370-86756&mode=dev).

[...]

PR #7912, opened 2023-09-29
feat(combobox): limit display of selected items with new selection-display prop

eriklharper:
**Related Issue:** #4326 

## Summary

This PR introduces a new `selection-display` property to Combobox to control the chip display with the available values `all`(default), `fit` and `single`.  This new property only takes effect when `selection-mode` is `multiple` or `ancestors`.


packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -761,6 +793,72 @@ export class Combobox
...
+      return;
+    }
+    if (this.displayMode === "fit-to-line") {
+      const chipEls = Array.from(this.el.shadowRoot.querySelectorAll("calcite-chip")).filter(
+        (chipEl) => chipEl !== this.selectedIndicatorChipEl
+      );
+      const { fontSize, fontFamily } = getComputedStyle(this.textInput);
+      const inputContainerElWidth = getElementWidth(this.inputContainerEl);
+      const inputContainerElGap = parseInt(
+        getComputedStyle(this.inputContainerEl).gap.replace("px", "")
```
eriklharper, line 820: Might be able to reference the `gap` value from a design token here, need to look into that.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -782,6 +816,107 @@ export class Combobox
...
+      chipEls.forEach((chipEl: HTMLCalciteChipElement) => {
+        if (chipEl.selected) {
+          const chipElWidth = getElementWidth(chipEl);
+          if (chipElWidth && chipElWidth < availableHorizontalChipElSpace) {
+            availableHorizontalChipElSpace -= chipElWidth + chipContainerElGap;
+            showChip(chipEl);
+          } else {
+            hideChip(chipEl);
+          }
+        } else {
```
jcfranco, line 898: For readability, can you flip the order of these? This way the shorter block will be easier to spot and scan and the bigger one can follow. Bonus points for returning early.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -782,6 +816,107 @@ export class Combobox
...
+  }
+
+  private maxCompactBreakpoint: number;
+
+  @State() compactDisplayMode = false;
+
+  private refreshDisplayMode = async () => {
+    await componentLoaded(this);
+
+    if (isSingleLike(this.selectionMode)) {
```
jcfranco, line 834: Let's create a perf follow-up issue for this. I see some opportunities to optimize this since it is an expensive call that gets called frequently:

1. Observe/Unobserve only when `displayMode` is set to `fit-to-line`. 
2. Split DOM reads and writes using `readTask`/`writeTask` (see https://stenciljs.com/docs/api#other)
3. Avoid awaiting on the component to be ready if we know it's single-like or the required input element is not available.
4. Avoid recomputing `inputTextWidth` unless `placeholder` changes.

packages/calcite-components/src/utils/dom.ts
```diff
@@ -81,6 +81,19 @@ export function getElementProp(el: Element, attribute: string, fallbackValue: an
...
 
+/**
+ * This helper returns the computed width in pixels of a rendered HTMLElement.
+ *
+ * @param {HTMLElement} el An element.
+ * @returns {number} The element's width.
+ */
+export function getElementWidth(el: HTMLElement): number {
+  if (!el) {
+    return;
```
jcfranco, line 92: Should this return 0?

packages/calcite-components/src/utils/dom.ts
```diff
@@ -102,6 +115,23 @@ export function getShadowRootNode(el: Element): ShadowRoot | null {
...
+/**
+ * This helper returns the computed width in pixels a given text string takes up on screen.
+ *
+ * See https://stackoverflow.com/questions/118241/calculate-text-width-with-javascript
+ *
+ * @param {string} text The string of text to measure.
+ * @param {string} font The CSS font attribute's value, which should include size and face, e.g. "12px Arial".
+ */
+export function getTextWidth(text: string, font: string): number {
+  if (!text) {
```
jcfranco, line 127: Should this return 0?

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1183,6 +1336,153 @@ export class Combobox
...
+      selectedVisibleChipsCount,
+      setSelectedIndicatorChipEl,
+    } = this;
+    let chipInvisible, label;
+    if (compactDisplayMode) {
+      chipInvisible = true;
+    } else {
+      if (displayMode === "single") {
+        const selectedItemsCount = getSelectedItems().length;
+        if (isAllSelected()) {
```
jcfranco, line 1407: Either call this with `this` (e.g., `this.isAllSelected()`) or store the value and reuse. The way it's called is requires you to unnecessarily bind the function where it's declared. Applies to other assigned, bound functions.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1135,6 +1283,10 @@ export class Combobox
     }
   }
 
+  private isAllSelected = (): boolean => {
```
jcfranco, line 1286: This, and similar variables, can be class functions instead of assigned bound functions since these are not event handlers and `this` is bound to the component instance at their call site. https://www.joshmorony.com/understanding-jsx-for-stencil-js-applications/#event-binding has some nice and concise info on this.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1340,9 +1642,24 @@ export class Combobox
...
+            class={{
+              "grid-input": true,
+              [CSS.displayModeFitToLine]: fitToLineDisplayMode,
+              [CSS.displayModeSingle]: singleDisplayMode,
+            }}
+            ref={this.setChipContainerEl}
+          >
             {this.renderIconStart()}
-            {!single && this.renderChips()}
+            {!singleSelectionMode && !singleDisplayMode && this.renderChips()}
```
jcfranco, line 1654: Can you look at simplifying this? IMO, it's kind of hard to follow because some variables are similarly named and there's some duplication going on.  

One option would be to render multiple parts that share the same condition:

```tsx
{!singleSelectionMode &&
  !showAllDisplayMode && [
    this.renderSelectedIndicatorChip(),
    this.renderSelectedIndicatorChipCompact(),
    this.renderAllSelectedIndicatorChip(),
    this.renderAllSelectedIndicatorChipCompact(),
  ]}
```

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1183,6 +1336,153 @@ export class Combobox
...
 
+  renderAllSelectedIndicatorChip(): VNode {
+    const {
+      compactDisplayMode,
+      isAllSelected,
+      scale,
+      selectedVisibleChipsCount,
+      setAllSelectedIndicatorChipEl,
+    } = this;
+    const label = this.messages.allSelected || "All selected";
```
jcfranco, line 1347: Let's remove the fallback since we've already got messages set up.

packages/calcite-components/src/components/combobox/assets/combobox/t9n/messages.json
```diff
@@ -1,4 +1,6 @@
 {
+  "all": "All",
```
jcfranco, line 2: You need to also update `messages_en.json` (see https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/conventions/Internationalization.md#t9ncomponent-pattern).

We have an NPM script to help with this (`util:sync-t9n-en-bundles`), but it appears it needs to be broken. Probably needs to be updated to account for the monorepo structure).

Sidebar: @anveshmekala we need to add a note about updating strings to the doc.

jcfranco (changes requested):
Looking good, @eriklharper! 

🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊
🧊🧊😎😎😎🧊🧊😎😎🧊🧊🧊😎😎🧊🧊😎🧊🧊🧊😎🧊
🧊😎🧊🧊🧊🧊😎🧊🧊😎🧊😎🧊🧊😎🧊😎🧊🧊🧊😎🧊
🧊😎🧊🧊🧊🧊😎🧊🧊😎🧊😎🧊🧊😎🧊😎🧊🧊🧊😎🧊
🧊😎🧊🧊🧊🧊😎🧊🧊😎🧊😎🧊🧊😎🧊😎🧊🧊🧊🧊🧊
🧊🧊😎😎😎🧊🧊😎😎🧊🧊🧊😎😎🧊🧊😎😎😎🧊😎🧊
🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -112,6 +120,15 @@ export class Combobox
...
   @Prop({ reflect: true }) clearDisabled = false;
 
+  /**
+   * Controls the display behavior of multiple selected items.
+   * This property does not apply when selection-mode is set to "single" or "single-persist".
+   * - "show-all" displays every selected chip, wrapping to the next line if necessary
+   * - "single" displays a single chip indicating the total amount of selected items
+   * - "fit-to-line" only displays selected chips that fit horizontally, with a non-closable chip indicating additional selected items as needed
+   */
+  @Prop() displayMode: DisplayMode = "show-all";
```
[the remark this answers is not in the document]
eriklharper: Good catch!  Will do :)

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -533,6 +563,10 @@ export class Combobox
 
   transitionEl: HTMLDivElement;
 
+  private allSelectedIndicatorChipEl: HTMLCalciteChipElement;
```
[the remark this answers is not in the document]
eriklharper: Yeah, I like having everything alphabetized too.  The names are all over the place, but I'll do the best I can with moving the new ones and if you want in a follow-up PR I can re-alphabetize everything per-section like in other components.  Let me know if you want me to do that :)

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1183,6 +1336,153 @@ export class Combobox
...
+      displayMode,
+      getSelectedItems,
+      isAllSelected,
+      scale,
+      selectedHiddenChipsCount,
+    } = this;
+    let chipInvisible, label;
+    if (compactDisplayMode) {
+      const selectedItemsCount = getSelectedItems().length;
+      chipInvisible = isAllSelected() ? true : false;
```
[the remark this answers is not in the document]
eriklharper: Nice catch, that line is completely unnecessary now, must have been a holdout from an older point in time that I just missed.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1183,6 +1336,153 @@ export class Combobox
...
+          (isAllSelected() && selectedVisibleChipsCount === 0) ||
+          selectedHiddenChipsCount === 0
+        ) {
+          chipInvisible = true;
+        } else {
+          chipInvisible = false;
+        }
+        label =
+          selectedVisibleChipsCount > 0
+            ? `+${selectedHiddenChipsCount}`
```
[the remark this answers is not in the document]
eriklharper: Excellent point, thank you for that catch also!  Maybe I should make a follow-up issue to address that so as not to delay this install?  I need to think about that and probably reach out to i18n on that.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -782,6 +816,107 @@ export class Combobox
...
+      selectedIndicatorChipEl,
+      showChip,
+      textInput,
+    } = this;
+
+    const chipContainerElGap = parseInt(getComputedStyle(chipContainerEl).gap.replace("px", ""));
+    const chipContainerElWidth = getElementWidth(chipContainerEl);
+    const { fontSize, fontFamily } = getComputedStyle(textInput);
+    const inputTextWidth = getTextWidth(placeholder, `${fontSize} ${fontFamily}`);
+    const inputWidth = (inputTextWidth || 48) + chipContainerElGap;
```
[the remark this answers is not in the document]
eriklharper: Even better Aaron suggested using a core design token here.  I'll create a follow-up issue perhaps and loop in @alisonailea to find out how to reference that core token here in JS.

packages/calcite-components/src/components/combobox/combobox.tsx
eriklharper: I just now realized I should probably have a message for `${num} selected` too.

packages/calcite-components/src/components/combobox/combobox.tsx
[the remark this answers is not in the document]
eriklharper: In RTL the `+` seems to display correctly. 
image

packages/calcite-components/src/components/combobox/combobox.tsx
eriklharper: Done.  Added one for `selected` as well.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -782,6 +816,107 @@ export class Combobox
     this.updateActiveItemIndex(targetIndex);
   }
 
+  private hideChip(chipEl: HTMLCalciteChipElement): void {
+    chipEl.classList.add(CSS.chipInvisible);
+  }
+
+  private showChip(chipEl: HTMLCalciteChipElement): void {
+    chipEl.classList.remove(CSS.chipInvisible);
```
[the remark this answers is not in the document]
eriklharper: I had issues using toggle, so I kept this pattern in place.  It also reads clearer IMO.

packages/calcite-components/src/utils/dom.ts
```diff
@@ -81,6 +81,19 @@ export function getElementProp(el: Element, attribute: string, fallbackValue: an
...
+ * This helper returns the computed width in pixels of a rendered HTMLElement.
+ *
+ * @param {HTMLElement} el An element.
+ * @returns {number} The element's width.
+ */
+export function getElementWidth(el: HTMLElement): number {
+  if (!el) {
+    return;
+  }
+  return Math.round(parseFloat(getComputedStyle(el).inlineSize.replace("px", "")));
```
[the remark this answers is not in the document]
eriklharper: It will return decimal values, so to keep the calculation simpler, I thought it best to round the result to make the math cleaner.  Probably is a negligible difference, could go either way.

image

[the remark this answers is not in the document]
eriklharper: It will make the calculation that much more accurate I suppose if we don't round the value.  I'll update that.

packages/calcite-components/src/components/combobox/combobox.tsx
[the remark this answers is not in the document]
eriklharper: Done!  Uses `CoreSizing15`.

jcfranco (approved):
I have some comments for follow-ups, but we can proceed to land this. Thanks for addressing the review barrage! 💪

🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️
🎛️🛬🎛️🎛️🎛️🛬🛬🛬🛬🎛️🛬🛬🛬🛬🛬🎛️🛬🎛️🎛️🎛️🛬🎛️🛬🎛️
🎛️🛬🎛️🎛️🎛️🛬🎛️🎛️🎛️🎛️🎛️🎛️🛬🎛️🎛️🎛️🛬🛬🎛️🛬🛬🎛️🛬🎛️
🎛️🛬🎛️🎛️🎛️🛬🎛️🛬🛬🎛️🎛️🎛️🛬🎛️🎛️🎛️🛬🎛️🛬🎛️🛬🎛️🛬🎛️
🎛️🛬🎛️🎛️🎛️🛬🎛️🎛️🛬🎛️🎛️🎛️🛬🎛️🎛️🎛️🛬🎛️🎛️🎛️🛬🎛️🎛️🎛️
🎛️🛬🛬🛬🎛️🛬🛬🛬🛬🎛️🎛️🎛️🛬🎛️🎛️🎛️🛬🎛️🎛️🎛️🛬🎛️🛬🎛️
🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️🎛️

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -112,6 +119,15 @@ export class Combobox
...
   @Prop({ reflect: true }) clearDisabled = false;
 
+  /**
+   * Controls the display behavior of multiple selected items.
+   * This property does not apply when selection-mode is set to "single" or "single-persist".
+   * - "show-all" displays every selected chip, wrapping to the next line if necessary
+   * - "single" displays a single chip indicating the total amount of selected items
+   * - "fit-to-line" only displays selected chips that fit horizontally, with a non-closable chip indicating additional selected items as needed
+   */
+  @Prop({ reflect: true }) displayMode: DisplayMode = "show-all";
```
macandcheese, line 129: Could we use "all", "single", "auto" here? I think we could explain the "fit-to-line" in the value documentation string. I think the "show-" part of "show-all" is implied and "fit-to-line" is a little confusing.

eriklharper: @ashetland @SkyeSeitz ☝️ thoughts?

ashetland: I love it. Nice simplification.

SkyeSeitz: I agree the "fit-to-line" value is tricky! Only confusion with "auto" is how we have used it in the system thus far. Feels like it might introduce a somewhat conflicting meaning here as keep to a single line when elsewhere I believe we consistently use it as a hug contents value speaking to width/height.

Would simply `display-mode="all | fit | single"` work?

eriklharper: Based on a discussion, the team voted to use `selection-display: "all" | "fit" | "single"`, `all` being the default behavior to wrap all chips, `fit` to be the mode with the overflow chips on a single line, and `single` for just a single indicator chip.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -112,6 +120,15 @@ export class Combobox
    */
   @Prop({ reflect: true }) clearDisabled = false;
 
+  /**
+   * Controls the display behavior of multiple selected items.
+   * This property does not apply when selection-mode is set to "single" or "single-persist".
```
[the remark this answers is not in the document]
geospatialem: @eriklharper WDYT about the following for doc consistency, similar to [sheet](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/sheet/sheet.tsx#L55-L58):

```
When `selectionMode` is `"ancestors"` or `"multiple"`, specifies the display of multiple `calcite-combobox-item` selections - `"all"` (displays all selections with individual `calcite-chip`s), `"single"` (display one `calcite-chip` with the total number of selections), or `"auto"` (displays individual `calcite-chip`s that scale to the component's size, including a non-closable `calcite-chip`, which provides the number of additional `calcite-combobox-item` selections not visually displayed).
```

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -112,6 +119,14 @@ export class Combobox
    */
   @Prop({ reflect: true }) clearDisabled = false;
 
+  /**
+   * When `selectionMode` is `"ancestors"` or `"multiple"`, specifies the display of multiple `calcite-combobox-item` selections
+   * - `"all"` (displays all selections with individual `calcite-chip`s)
+   * - `"fit"` (displays individual `calcite-chip`s that scale to the component's size, including a non-closable `calcite-chip`, which provides the number of additional `calcite-combobox-item` selections not visually displayed).
+   * - `"single"` (display one `calcite-chip` with the total number of selections)
+   */
```
geospatialem, line 127: Two small copyedit suggestions for consistency:

- Only show one dash `-` before the `"all"` reference - adding commas after the other items for consistency 
    - i.e., `....with individual calcite-chips),`
- Add a period at the end of the single parenthesis
    - i.e., `.... with the total number of selections).`

packages/calcite-components/src/components/combobox/combobox.tsx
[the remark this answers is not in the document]
eriklharper: image

Issue #7430, opened 2023-08-03
input-time-zone enhancements


jcfranco:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

The following would be great additions to the time-zone picking experience:

* [ ] add a way to display a preview of local time across time zone options (⚠️ postponed until there's a greater demand)
  ![time-zone-labels](https://user-images.githubusercontent.com/197440/238069360-18abd52d-a355-4813-bf50-9363d983f3fa.png)
* [ ] display more info per time zone and also allow users to search for a time zone regardless of being listed or not
  ![time-zone-labels](https://user-images.githubusercontent.com/197440/238069470-db99ec09-be6f-492b-8008-9c9d94fa3fb6.png)
  Users should also have a way to set the time zone option by specifying the IANA time zone (must match a value from `Intl.supportedValuesOf("timeZone")`).

cc @ashetland

### Acceptance Criteria

- [ ] A property to enable each of the options from the description.
- [ ] Explore one of the following from [Richie's comment](https://github.com/Esri/calcite-design-system/issues/7430#issuecomment-1681069819) for accommodating timestamps:
  - Have two components `calcite-input-time-zone-offset` and `calcite-input-iana-time-zone`
  - Have one component, `calcite-input-time-zone`, with two modes (e.g. `"offset"`, `"iana-time-zone"` or `"name"`)
  - Keep `calcite-input-time-zone` as-is, where Ian uses a `calcite-combobox` with items populated from `Intl.supportedValuesOf('timeZone')` and translated using [this](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/input-time-zone/assets/input-time-zone/t9n/messages.json).
- [ ] Provide Daylight Savings Support or using a reference date from `calcite-date-picker` per [Richie's comment below](https://github.com/Esri/calcite-design-system/issues/7430#issuecomment-1681099248)
  -   _See the suggestion from Richie in the [comment link](https://github.com/Esri/calcite-design-system/issues/7430#issuecomment-1681099248)_
- [ ] Support internationalized string: _See the suggestion from Richie in the [following comment](https://github.com/Esri/calcite-design-system/issues/7430#issuecomment-1681116301)_

### Relevant Info

Stems from https://github.com/Esri/calcite-design-system/issues/6590#issuecomment-1654552443

### Which Component

`calcite-input-time-zone`

### Priority impact

[...]

PR #7913, opened 2023-10-01
feat(input-time-zone): add time zone offset and name mode

jcfranco:
**Related Issue:** #7430 

## Summary

This adds a `mode` property to the component to list:

* Time zone offsets with (searchable) time zone groups
* IANA time zone names

### API changes

```ts
interface InputTimeZone {
  /**
   * Sets the time zone display and value used by the component.
   * 
   * when set to "offset" (default), displays list of time zone offsets
   * when set to "name", displays list of IANA time zone names
   */
  mode: "offset" | "name"; 

  /**
   * The reference date used when processing offsets (affects daylight savings time).
   * When not provided, the reference date will be today's date.
   *  
   * Note: this only affects to the `offset` mode
   */  
  referenceDate: Date | string; 
}
```

#### `mode="offset"`

* wires up [`timezone-groups`](https://www.npmjs.com/package/timezone-groups) package to list cities along with each respective offset entry
* city names are translated
* allows searching by time zone regardless of being displayed in the label or not
* E2E tests extend `Intl.DateTimeFormat` to patch missing behavior (limited subset of time zones) from Chromium v92 (bundled w/ Stencil's Puppeteer v10).

input-time-zone-enhancements

### Pending

CI testing needs looking into. Except for `labelable` (most likely needs updating to handle `Intl.DateTimeFormat` patching), all tests pass locally. 

Screenshot 2023-10-02 at 11 43 17 AM

Will create a separate issue if these can't be stabilized for this milestone.

**Update:** Figured out the issue. Each E2E test needs to emulate the same default time zone.


packages/calcite-components/src/demos/input-time-zone.html
```diff
@@ -34,23 +34,470 @@
...
+              "Pacific/Saipan",
+              "Pacific/Tahiti",
+              "Pacific/Tarawa",
+              "Pacific/Tongatapu",
+              "Pacific/Truk",
+              "Pacific/Wake",
+              "Pacific/Wallis",
+            ];
+
+            console.log("EYO");
```
driskull, line 477: eyo?

packages/calcite-components/src/demos/input-time-zone.html
```diff
@@ -34,23 +34,470 @@
...
+    <script>
+      (function () {
+        const originalSupportedValuesOf = Intl.supportedValuesOf;
+
+        Intl.supportedValuesOf = function (key) {
+          if (key === "timeZone") {
+            /**
+             * Polyfilled due to lack of support for `Intl.DateTimeFormatOptions#timeZoneName` in Chromium v92 (bundled in Puppeteer v10).
+             */
+            const fallbackTimeZoneNames = [
```
driskull, line 46: nitpick: Could this be saved in some other JS file? could maybe be reusable.

packages/calcite-components/src/components/input-time-zone/input-time-zone.tsx
```diff
@@ -317,17 +381,17 @@ export class InputTimeZone
...
-            const selected = this.selectedTimeZoneGroup === group;
-            const label = group.offsetLabel;
-            const value = group.offsetValue;
+          {this.timeZoneItems.map((group) => {
+            const selected = this.selectedTimeZoneItem === group;
+            const { label, value } = group;
 
             return (
               <calcite-combobox-item
+                data-value={value}
```
driskull, line 390: We should have a best practices document regarding use of data attributes 👍

driskull (approved):
✨ 👀 Looks good!

packages/calcite-components/src/demos/input-time-zone.html
jcfranco: Whoopsie! This was meant to the time zone polyfill used in E2E tests. Removing! 



> could maybe be reusable.

Possibly, but it may also not be needed once we bump Stencil in November. I think E2E is the only place where this is needed.

packages/calcite-components/src/demos/input-time-zone.html
jcfranco: Boom, boom, boom! Let me hear you say **EYO**! 

Sorry, stowaway test log. 😅

EYO

jcfranco:
Figured out the issue. Each E2E test needs to emulate the same default time zone. I'll skip the common ones and create a follow-up issue since those might take a bit more work to do this consistently.

packages/calcite-components/src/components/input-time-zone/utils.ts
```diff
@@ -1,43 +1,132 @@
...
+  referenceDate: Date
+): Promise<TimeZoneItem[]> {
+  const referenceDateInMs: number = referenceDate.getTime();
+  const timeZoneNames = Intl.supportedValuesOf("timeZone");
+
+  if (mode === "offset") {
+    if (!timeZoneGroups) {
+      timeZoneGroups = Promise.all([
+        import("timezone-groups/dist/index.js"),
+        import("timezone-groups/dist/strategy/native/index.js"),
```
benelan, line 61: Is the `timezone-groups` directory being generated in `dist` or somewhere in `src`? I'm kinda curious what these dynamic imports are being compiled to, can you search for one of the paths in `dist` to see if these strings remain intact? If they are, can we be certain the paths will resolve correctly in all of our consumer's different use cases?

jcfranco: `timezone-groups` is a new dependency. Lean chunks for each import are created. 🥩

jcfranco:
It looks like screenshots are taken early in the component initialization. Going to the actual story preview shows the right thing. I'll play with larger delays in a separate PR and accept the current ones to avoid blocking this one. @ashetland I'll ping you once that's ready for review.

packages/calcite-components/src/components/input-time-zone/utils.ts
benelan: Nevermind, I thought we were creating a directory named `timezone-groups` but it's actually an NPM package so we're good.

PR #7922, opened 2023-10-02
ci: the current milestone ends after release day

benelan:
## Summary

Ensure the PR check that blocks installs if the current milestone is for maintenance doesn't think the milestone changed until after the release.


geospatialem (approved):
🥇

PR #7927, opened 2023-10-03
fix(panel): fix collapse action title and reverse icon direction

driskull:
**Related Issue:** #7930

## Summary

- invert expand/collapse title attribute value.
- reverse icon direction.


jcfranco (approved):
💥

Issue #7930, opened 2023-10-03
Collapsible Panel affordance corrections


ashetland:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Panel's collapse button is reversed when compared to other components like Accordion and Block.

The `title` attribute and `aria-label` also don't match.

### Expected Behavior

The default orientation matches Accordion and Block. Proposing adding a `collapse-inverted` prop to replicate the current behavior.

<img width="808" alt="CleanShot 2023-10-03 at 15 49 59@2x" src="https://github.com/Esri/calcite-design-system/assets/108549080/e7ea0582-9db0-4d31-8157-50c45b458019">

### Reproduction Sample

https://codepen.io/ashetland/pen/jOXeOXK

### Reproduction Steps

Compare with Block and Accordion in the provided codepen.

### Reproduction Version

v1.9.0

### Relevant Info

cc @driskull @SkyeSeitz 

### Priority impact

p2 - want for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

macandcheese:
Do we need to allow this to be customized? Can we just have the default match the  other places this chevron is used?

ashetland:
> Do we need to allow this to be customized? Can we just have the default match the other places this chevron is used?

Good question for @asangma and @driskull. I believe there thinking was there particular use case required it be inverted, but this may be adding unnecessary complication.

driskull:
I don’t think it’s special to popover. It really applies to any component that is anchored at the bottom of something to where when expanded it goes up instead of down.

SkyeSeitz:
Would we consider this similar to Action Bar's position prop to determine expand direction based on if it is anchored to the start or end?

geospatialem:
Verified in `1.9.1-next.0`. 

@driskull do we need to flip the behavior for the `collapseDirection` prop values? It looks like the chevron is opposite the value.

### `collapseDirection="up"`:
![image](https://github.com/Esri/calcite-design-system/assets/5023024/61336c17-2b7d-46d8-bda6-3aaf4445893a)

### `collapseDirection="down"`:
![image](https://github.com/Esri/calcite-design-system/assets/5023024/f48741a9-ceac-4b25-a47d-a09a9741ad56)

[...]

PR #7931, opened 2023-10-03
test(flow-item, panel): add tests for internal property.

driskull:
**Related Issue:** #7930

## Summary

- Adds stories for internal collapseDirection property.

### DIFF
--- a/packages/calcite-components/src/components/flow-item/flow-item.stories.ts
+++ b/packages/calcite-components/src/components/flow-item/flow-item.stories.ts
@@ -66,9 +66,9 @@ const createAttributes: (options?: { exceptions: string[] }) => Attributes = ({ 
         },
       },
       {
-        name: "collapse-inverted",
+        name: "collapse-direction",
         commit(): Attribute {
-          this.value = boolean("collapseInverted", false);
+          this.value = select("collapseDirection", ["down", "up"], "down");
           delete this.build;
           return this;
         },
@@ -167,8 +167,8 @@ export const collapsed_TestOnly = (): string => html`
   <calcite-flow-item collapsed collapsible closable> Hello World! </calcite-flow-item>
 `;
 
-export const collapseInverted_TestOnly = (): string => html`
-  <calcite-flow-item collapsed collapsible collapse-inverted closable> Hello World! </calcite-flow-item>
+export const collapseDirection_TestOnly = (): string => html`
+  <calcite-flow-item collapsed collapsible collapse-direction="up" closable> Hello World! </calcite-flow-item>
 `;
 
 export const disabledWithStyledSlot_TestOnly = (): string => html`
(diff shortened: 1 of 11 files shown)


packages/calcite-components/src/components/flow-item/flow-item.stories.ts
```diff
@@ -159,6 +167,10 @@ export const collapsed_TestOnly = (): string => html`
   <calcite-flow-item collapsed collapsible closable> Hello World! </calcite-flow-item>
 `;
 
+export const collapsedInverted_TestOnly = (): string => html`
```
jcfranco, line 170: `collapsedInverted_TestOnly` ➡️ `collapseInverted_TestOnly`

jcfranco (commented):
If this prop only affects the icon and not the way the panel expands, for consistency, could we rename it to follow something similar to `flipRtl`?

driskull:
> If this prop only affects the icon and not the way the panel expands, for consistency, could we rename it to follow something similar to flipRtl?

@jcfranco I had considered `flipRTL` but it actually doesn't follow our naming standard. Ideally, it would be `rtl-flipped` to be more inline with how HTML names attributes right? Do we want to consider updating that property to be more inline with our other props?

alisonailea (approved):

macandcheese:
Since ‘collapseInverted’ doesn’t really explain what the default direction is, maybe ‘collapseDirection’ top / bottom with a meaningful default is easier to understand? 


It does seem like it would maybe be confusing or lead to misuse as an option in Block, Accordion, etc, so would this only be available in Panel?

driskull:
> Since ‘collapseInverted’ doesn’t really explain what the default direction is, maybe ‘collapseDirection’ top / bottom with a meaningful default is easier to understand?

I agree. Maybe you me and @jcfranco can figure out a name? I can also just make this internal for now.

> It does seem like it would maybe be confusing or lead to misuse as an option in Block, Accordion, etc, so would this only be available in Panel?

I think it could be useful in anything that uses expand arrows to indicate expand direction. Currently, we assume that accordion and block open below but if the component is anchored at the bottom of something then that isn't the case anymore.

jcfranco:
> Ideally, it would be rtl-flipped to be more inline with how HTML names attributes right? Do we want to consider updating that property to be more inline with our other props?

Good point. I don't think we focus on this explicitly in our conventions. From what I recall, we've suggested concise adjectives for boolean names. We could update the conventions and do another audit on boolean names.

FWIW, I was more suggesting to use `flip` in the name. 😅

> Since ‘collapseInverted’ doesn’t really explain what the default direction is, maybe ‘collapseDirection’ top / bottom with a meaningful default is easier to understand?

Agreed. Would we ever have other directions? 

I'm fine with keeping it internal while we hash this out to land the increased test coverage.

jcfranco (approved):
Aside from the naming discussion, LGTM!

Good to merge as is if the prop is made internal.

macandcheese:
> I think it could be useful in anything that uses expand arrows to indicate expand direction. Currently, we assume that accordion and block open below but if the component is anchored at the bottom of something then that isn't the case anymore.

Yeah, although I think flipping the direction of the icon is unneeded because the area that expands is still opening in the same space / direction - the containing element just gets taller. A "real flip" IMO would be the content area opening on the "other side" of the element.

driskull:
Reverted to previous name and `@internal` state.

driskull:
@ashetland can you review screenshot tests?

Issue #7944, opened 2023-10-05
[Input Time Zone] Doesn't display in Firefox browsers and console errors are present


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The Maps SDK for JS Editor form relies on the Time Zone component, which doesn't display in `1.9.0`. There are also errors on the page:

> Uncaught (in promise) TypeError: format: expected string, got undefined
> createTimeZoneItems https://js.arcgis.com/calcite-components/1.9.0/p-fcc3a585.entry.js:6
> createTimeZoneItems https://js.arcgis.com/calcite-components/1.9.0/p-fcc3a585.entry.js:6

### Expected Behavior

Similar to `1.8.0`, the component displays and no console errors are present.

### Reproduction Sample

https://codepen.io/geospatialem/pen/zYyMrpY

### Reproduction Steps

1. Open https://codepen.io/geospatialem/pen/zYyMrpY with Firefox
2. _Contact Kitty for username/password_
3. Click on the Edit Features "Select"
4. Select a point to edit
5. Select one of the objects in the list
6. Open the browser console
7. Observe the behavior and errors of:

> Uncaught (in promise) TypeError: format: expected string, got undefined
> createTimeZoneItems https://js.arcgis.com/calcite-components/1.9.0/p-fcc3a585.entry.js:6
> createTimeZoneItems https://js.arcgis.com/calcite-components/1.9.0/p-fcc3a585.entry.js:6

### Reproduction Version

1.9.0

### Relevant Info

Seems specific to Firefox browsers

### Regression?

1.8.0

### Priority impact

p1 - need for current milestone

### Impact

Very high - impacts MapViewer and Maps SDK for JS teams in the upcoming release Monday.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Map Viewer

geospatialem:
cc @rebollin @jonuihlein @hgonzago

geospatialem:
Verified in `1.9.1-next.1` with Firefox.

jcfranco:
For context, we reopened due to bug in certain time zone grouping scenario.

geospatialem:
Verified in `1.9.2-next.3` with Firefox.

PR #7947, opened 2023-10-05
fix(input-time-zone): fix handling of unknown and cityless time zones from offset display mode

jcfranco:
**Related Issue:** #7944

## Summary

This updates the component to remove cityless time zone entries in the `offset` mode and for unknown entries (either from browser differences or updated values) their city part will be displayed as is.

**Note**: due to time constraints, testing will be tackled in #7951.

### DIFF
--- a/packages/calcite-components/src/components/input-time-zone/utils.ts
+++ b/packages/calcite-components/src/components/input-time-zone/utils.ts
@@ -4,6 +4,48 @@ import { InputTimeZoneMessages } from "./assets/input-time-zone/t9n";
 
 const hourToMinutes = 60;
 
+const timeZoneNameBlockList = [
+  "CET",
+  "CST6CDT",
+  "EET",
+  "EST",
+  "EST5EDT",
+  "Etc/GMT+1",
+  "Etc/GMT+10",
+  "Etc/GMT+11",
+  "Etc/GMT+12",
+  "Etc/GMT+2",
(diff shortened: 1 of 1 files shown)


packages/calcite-components/src/components/input-time-zone/utils.ts
```diff
@@ -4,6 +4,48 @@ import { InputTimeZoneMessages } from "./assets/input-time-zone/t9n";
 
 const hourToMinutes = 60;
 
+const timeZoneNameBlockList = [
```
alisonailea, line 7: Is there a way to get this list from a trusted source rather than having to maintain it ourselves?

jcfranco: Good idea. I could reliably exclude `Etc/X` ones, but would have to investigate separately if entries with no delimiter are reliable to block as well.

driskull (approved):
👏 👍 💯

packages/calcite-components/src/components/input-time-zone/utils.ts
jcfranco: I focused too much on having this dynamically generated that I didn't address the "trusted source" part of your question. Unfortunately, `Intl` doesn't provide metadata for time zone names, so we have to handle this on our side. That would definitely make things simpler here.

Issue #7934, opened 2023-10-04
Combobox - filtering combobox items and selecting by tap fails to update the selected item


ian13262:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Here is a very specific workflow in **1.9 only** that does not follow expected combobox behavior:

1.) Click to open a combobox with multiple selected items
2.) Filter the items by typing some input
3.) Choose a filtered item, but carefully! This item **must** be:
* **Not** in its original position in the list
* Selected by **tapping** a trackpad (and occasionally by clicking with a mouse, but this is variable)

4.) The new item will not be selected appropriately.

### Expected Behavior

The workflow above should correctly select that tapped item, but it doesn't.

### Reproduction Sample

https://codepen.io/iyouth66/pen/LYMgEOW

### Reproduction Steps

See above.

Demo: 
https://github.com/Esri/calcite-design-system/assets/131201379/4367da0a-936d-44e5-b685-8349a3df30fc

### Reproduction Version

1.9

### Relevant Info

Windows 10 (must use trackpad **tap** to reliably reproduce this issue); Chrome Version 117.0.5938.150 (Official Build) (64-bit)

_Note_: Also seems to apply to the Mac OS trackpad.

cc @geospatialem - hoping to have this for 1.9.1, if time permits.

### Regression?

1.8

### Priority impact

p1 - need for current milestone

### Impact

All comboboxes are affected, particularly `single-persist` ones (time-zone component, field selection, etc.).

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Online

geospatialem:
Possibly related to https://github.com/Esri/calcite-design-system/pull/7721. The above sample works as-expected in `1.8.0-next.0`, but shows starting at `1.8.0-next.1`.

geospatialem:
Verified in `1.9.1-next.1` with Windows trackpad in Chrome `Version 117.0.5938.149 (Official Build) (64-bit)`

PR #7954, opened 2023-10-05
fix(combobox): fix issue causing value to be cleared when selecting an item (Windows + trackpad)

jcfranco:
**Related Issue:** #7934 

## Summary

Updates the component to consistently bail on blur handling logic if the input is blurred, but focus is still within the component.

### Notes

* the regression was introduced by https://github.com/Esri/calcite-design-system/pull/7721/
* due to time constraints, testing will be tackled in a follow-up issue as it's not easy to reproduce in an E2E test.

### DIFF
--- a/packages/calcite-components/src/components/combobox/combobox.tsx
+++ b/packages/calcite-components/src/components/combobox/combobox.tsx
@@ -761,17 +761,17 @@ export class Combobox
     this.updateActiveItemIndex(targetIndex);
   }
 
-  setInactiveIfNotContained = (event: Event): void => {
+  private setInactiveIfNotContained = (event: Event): void => {
     const composedPath = event.composedPath();
+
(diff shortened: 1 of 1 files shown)


driskull (approved):
Awesome! 🌈 ☁️ 🦄

jcfranco:
@anveshmekala Could you give this a look and comment if you notice anything I may have missed?

anveshmekala:
> @anveshmekala Could you give this a look and comment if you notice anything I may have missed?

This will break the issue #[2162](https://github.com/Esri/calcite-design-system/issues/2162) when the user chooses to enter some custom value and tab  out of the component. The tests should have cover this scenario but they are validating against the value property which will be empty since there are no `selectedItems`.

jcfranco:
Yikes, thanks for catching that! I knew this fix felt too easy. 😂

PR #7956, opened 2023-10-06
fix(combobox): clear custom input value on blur

anveshmekala:
**Related Issue:** #2162 

## Summary

Clears input value on `blur` if the value has no matching `combobox-item` when `allowCustomValue` property is set to `false`


packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -298,15 +298,19 @@ export class Combobox
       return;
     }
 
-    this.setInactiveIfNotContained(event);
+    const composedPath = event.composedPath();
+    if (!this.open || composedPath.includes(this.el) || composedPath.includes(this.referenceEl)) {
```
alisonailea, line 302: Can you add a test that explicitly checks that the custom input value doesn't clear when the compose path includes this.el?

alisonailea (changes requested):

packages/calcite-components/src/components/combobox/combobox.tsx
anveshmekala: we do have a test for this one currently. Any item selection test covers this usecase.

packages/calcite-components/src/components/combobox-item/combobox-item.tsx
```diff
@@ -160,7 +160,7 @@ export class ComboboxItem implements ConditionalSlotComponent, InteractiveCompon
     this.selected = !this.selected;
   }
 
-  itemClickHandler = (event: MouseEvent): void => {
+  itemPointerHandler = (event: PointerEvent): void => {
     event.preventDefault();
```
jcfranco, line 164: Can you confirm the event still needs to be canceled here? I strongly suspect it is no longer needed because:

1. It was added in a very early stage of the component
2. In this early iteration, it was set on an `<a>` 
3. Not sure why it was canceled in 2 since `href` was set to `#`, which won't redirect – maybe it was an extra step to prevent redirects (canceling `click` stops `<a>`'s default behavior)
4. This PR uses `pointerdown` (vs `click`) and canceling it will prevent `mousedown` and `mouseup` from firing (this was not the case when `click` was canceled)
5. The pointer event is set on `<li>`, which does not have any default click behavior (unlike `<a>`)

If not needed, it would help us avoid adding `pointerEventPreventDefaultOnClick` to the `disabled` test helper options.

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -975,6 +975,14 @@ interface DisabledOptions {
...
    */
   shadowAriaAttributeTargetSelector?: string;
+
+  /**
+   * Use this to specify if any pointerEvent is disabled on click.
+   *
+   *  Note: mousedown and mouseup events are skipped when pointerEvent has preventDefault
+   *  https://github.com/web-platform-tests/wpt/blob/master/pointerevents/compat/pointerevent_mouse-pointer-preventdefault.html#L54
+   */
+  pointerEventPreventDefaultOnClick?: boolean;
```
jcfranco, line 985: This argument is for a very particular use case and I'm not sure about its wider applicability.

packages/calcite-components/src/components/combobox-item/combobox-item.tsx
```diff
@@ -160,7 +160,7 @@ export class ComboboxItem implements ConditionalSlotComponent, InteractiveCompon
     this.selected = !this.selected;
   }
 
-  itemClickHandler = (event: MouseEvent): void => {
+  itemPointerHandler = (event: PointerEvent): void => {
```
jcfranco, line 163: Can you rename this to better match the particular event it's handling?

packages/calcite-components/src/components/combobox/combobox.e2e.ts
```diff
@@ -1097,7 +1097,7 @@ describe("calcite-combobox", () => {
       await page.waitForChanges();
       await page.keyboard.press("Tab");
       await page.waitForChanges();
-      expect(await combobox.getProperty("value")).toBe("");
+      expect(await inputEl.getProperty("value")).toBe("");
```
jcfranco, line 1100: Great catch, but can we submit these renames separately? There are no other changes in this file related to the PR, so it's out of scope.

jcfranco (changes requested):
I have some concerns regarding the changes to the `disabled` test helper. Other than that, this is looking good. 😎👍🖥

packages/calcite-components/src/components/combobox-item/combobox-item.tsx
anveshmekala: The reason for using `event.preventDefault` in this scenario is to avoid emitting compatibility `mouse` 

1.  Without preventDefault `mouse` compatibility events will be fired thus causing `blur` on `input` which results in loosing focus.
2.  Preventing mouse  events wouldn't cause any issues since we don't have any listeners or handlers for mouse events in combobox-item and combobox. 
3.  Though `<li>` element does not have any default click behavior , the preventDefault is used to prevent mouse and focus related events.

packages/calcite-components/src/components/combobox/combobox.e2e.ts
anveshmekala: These changes are related i think. This helps in asserting the input clear on blur functionality via keyboard & Mouse. Though these tests were added earlier as part of https://github.com/Esri/calcite-design-system/pull/7721 , they couldn't identify regression when https://github.com/Esri/calcite-design-system/pull/7954 is installed . Made changes to this test so that they can detect any regressions from now on.

packages/calcite-components/src/tests/commonTests.ts
anveshmekala: Yup, it is a specific use case . Added this to the options so that we can extend this if required to any component in future.  `pointerDownEventDefaultPrevented` might be a better name?

anveshmekala:
Closing in favor https://github.com/Esri/calcite-design-system/pull/8070

Issue #7959, opened 2023-10-06
flow-item: collapse property is not toggled when collapse button is clicked


driskull:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

collapse property is not toggled when collapse button is clicked

### Expected Behavior

collapse property should be toggled when collapse button is clicked

### Reproduction Sample

https://codepen.io/driskull/pen/BavGbXW

### Reproduction Steps

1. Open codepen
2. Click on collapse button
3. Notice the collapse property on the flow-item is not set to true

### Reproduction Version

1.9.1

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

geospatialem:
Verified in `1.9.2-next.0`.

PR #7960, opened 2023-10-06
fix(flow-item): update collapsed property when collapse button is clicked

driskull:
**Related Issue:** #7959

## Summary

- Updates collapsed property when `calcitePanelToggle` event is fired.
- Add test


jcfranco (approved):
LGTM!

PR #7971, opened 2023-10-10
fix(input-time-zone): fix error caused by time zone group filtering

jcfranco:
**Related Issue:** #7944

## Summary

Addresses an issue caused by time zone label index adjustment (from time zone group filtering) ended up as -1 and cause an error when using it as a lookup index.

This also adds missing tests from https://github.com/Esri/calcite-design-system/pull/7947/.


driskull (approved):
✨ 👍

PR #7980, opened 2023-10-11
fix: global css file reduce 5000 lines

alisonailea:
**Related Issue:** #7325

## Summary

calcite-design-tokens currently combines all modes and components into three files (headless, light, dark) which are then all included in the global.scss making the final global calcite.css file bloated.
There are some minor size improvements that can be made by utilizing design token references and removing duplications but the major issue is all of the component specific design tokens which do not need to be included in the global css file.

To remove this bloat the solution is...

1. update the token-transformer to output better SCSS files which utilize @use and @forward to link related styles
2. remove component tokens from global output
3. stop relying on class based mode setting and get/set the mode via a local storage key which may be accessed and set as an attribute per component
4. pull component tokens directly into the individual component styles


packages/calcite-design-tokens/src/$themes.json
```diff
@@ -3,297 +3,28 @@
     "id": "59a65f6d4a3657738c9d46d61315ae0b6fc6e9fd",
```
alisonailea, line 3: This whole file is outdated and set to be deleted.

alisonailea:
Some notes on the changes to tokens. You will notice a lot of the file changes are simply converting token references from the "$[token.reference.path]" to "{[token.reference.path]}". The use of the "$" to delineate token references was set in Token Studio by designers because it made their workflow easier within Figma. However, designers are no longer using Token Studio and the default file parser looks for "{}" token references. So, by converting to "{}" we remove the need for a custom parser, improving maintenance and performance.

packages/calcite-components/src/components/accordion-item/accordion-item.tsx
```diff
@@ -108,6 +111,11 @@ export class AccordionItem implements ConditionalSlotComponent {
   //  Lifecycle
   //
   //--------------------------------------------------------------------------
+  componentWillLoad(): void {
+    // TODO: check if this attribute is manually set by the user before getting the theme set in local storage.
+    // TODO: make this a part of the generic component setup.
+    this.calciteMode = getMode();
```
driskull, line 117: This would likely need to happen on `connectedCallback()` so a component updates its mode if its moved position in the DOM.

driskull:
> stop relying on class based mode setting and get/set the mode via a local storage key which may be accessed and set as an attribute per component

@alisonailea can you elaborate on why we would need to do this one?

I thought we would just be moving the component specific variables to the component files which would get placed into the CSS of the component instead of the global.css file therefore reducing the global.css file for users who are not using every component.

I have concerns with switching from using a CSS class approach to using an attribute. We had originally had a theme attribute but moved away from that because using CSS is a better practice than using an attribute for purely stylistic options. Using an attribute requires implementing our own cascading logic which is something that CSS just already does.

alisonailea:
>
> @alisonailea can you elaborate on why we would need to do this one?
> 

We will not be fully taking the approach listed in this draft as we don't currently want to rely on localstore. @jcfranco and I are working on a mutationObserver pattern solution similar to localization.

> 
> I thought we would just be moving the component specific variables to the component files
> 

We still need a way to scope applying the mode specific tokens to the component in the CSS.

':host([calcite-mode="light"]): { ...design tokens as CSS custom props go here } '

> 
> I have concerns with switching from using a CSS class approach to using an attribute.
> 

Without Container Style Queries (currently only available under a feature flag in chrome), using an attribute to specify the CSS is the only solution unless you want everything in global CSS.

driskull:
> We still need a way to scope applying the mode specific tokens to the component in the CSS.

Can you share some example mode specific component tokens? I'd like to take a look at them. Maybe we could just keep these in global if there aren't too many.

jcfranco:
This was addressed by https://github.com/Esri/calcite-design-system/pull/8215. Closing!

PR #7983, opened 2023-10-12
ci: don't link design-token version to CC version

benelan:
**Related Issue:** #

## Summary

We don't want the design tokens version to be in sync with the CC version, that is just for the component wrapper packages.


benelan:
Pushing this one thru to fix releases, cc @jcfranco

PR #7994, opened 2023-10-12
feat: reorganize breakpoints 

alisonailea:
**Related Issue:** n/a

## Summary

Different platforms need to target breakpoints in different ways. It might be easier to be explicit about min vs max breakpoint sizes. This is not a breaking change as it retains the existing “breakpoint-width-“ tokens for the time being but does mark them for deprecation.

### DIFF
--- a/packages/calcite-design-tokens/src/core.json
+++ b/packages/calcite-design-tokens/src/core.json
@@ -1528,6 +1528,11 @@
     },
     "breakpoint": {
       "width": {
+        "xxs": {
+          "value": "320px",
+          "type": "sizing",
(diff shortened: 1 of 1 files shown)


jcfranco (approved):
Awesome, I really dig this updated structure. Clearer and they can be used for both width or height.

jcfranco:
Very, very, very nitpicky: `re-organize` -> `reorganize` 🤓

alisonailea:
> they can be used for both width or height

While any token can technically be used anywhere I've updated the tokens to include the explicit breakpoint heights detailed in Figma so they are available for use as well.

PR #7998, opened 2023-10-12
chore: align expected node engine with volta

alisonailea:
**Related Issue:** n/a

## Summary
Trying to build the Calcite Design System packages with node 16 fails. It requires at least node 18.

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,9 +110,6 @@
     "workbox-build": "7.0.0"
   },
   "license": "SEE LICENSE.md",
-  "engines": {
-    "node": ">=18.0.0"
-  },
   "packageManager": "npm@10.1.0",
   "volta": {
     "node": "18.18.0",
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
   },
   "license": "SEE LICENSE.md",
   "engines": {
-    "node": ">=16.0.0"
+    "node": ">=18.0.0"
   },
   "packageManager": "npm@10.1.0",
   "volta": {
--- a/packages/calcite-components/package.json
+++ b/packages/calcite-components/package.json
@@ -86,5 +86,8 @@
     "@stencil/sass": "3.0.5",
     "@stencil/state-tunnel": "1.0.1"
   },
-  "license": "SEE LICENSE.md"
+  "license": "SEE LICENSE.md",
+  "volta": {
+    "extends": "../../package.json"
+  }
 }
--- a/packages/calcite-components/package.json
+++ b/packages/calcite-components/package.json
@@ -86,8 +86,5 @@
     "@stencil/sass": "3.0.5",
     "@stencil/state-tunnel": "1.0.1"
   },
-  "license": "SEE LICENSE.md",
-  "engines": {
-    "node": ">=16.0.0"
-  }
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -111,7 +111,7 @@
   },
   "license": "SEE LICENSE.md",
   "engines": {
-    "node": ">=16.0.0"
+    "node": ">=18.0.0"
```
jcfranco, line 114: Should this be `>=18.18.0`?

benelan (changes requested):
Specifying `engines` displays warnings for our users if they install CC with a node version that is not in the specified range. I'm pretty sure CC will still work on node v16 so we shouldn't throw warnings when people try to use it. Volta is what we use to specify our development version of Node.

ref: https://docs.npmjs.com/cli/v6/configuring-npm/package-json#engines

package.json
alisonailea: anything over 18 will work. I wanted to keep it flexible but I can update

packages/calcite-components/package.json
```diff
@@ -87,7 +87,7 @@
     "@stencil/state-tunnel": "1.0.1"
   },
   "license": "SEE LICENSE.md",
-  "volta": {
```
alisonailea, line 90: I'm removing volta at this level because a calcite-components consumer will not have the context of the monorepo to reference the volta node version.

benelan: The volta version is not for the calcite consumer. It is for automatically using the correct version of Node during development. Can you please add it back?

benelan:
That only applies to the engine listed in the calcite-components package, which doesn't currently have an engine listed at all. In the CONTRIBUTING docs it does not say you must use Volta but rather directs you to use the major version of node listed in the package.json (implying the engine key).

Following up on our Teams conversation, it looks like there was never an `engines` field before switching to the monorepo. I must have added it when copy/pasting from somewhere :sweat_smile: 

We can remove `engines` from the root and the newly added one in CC's `package.json`. Setting `engines` specifies the supported versions of node for using the package, and there aren't any packages being published from the root of the monorepo. I hesitate to add one to CC without doing the proper testing to determine the oldest supported version.

The supported versions for usage and the supported versions for development can be completely different, so I don't want to use the `engines` field for something it wasn't designed to do.

We can update the CONTRIBUTING doc to specify that the version listed under the `volta` field should be used for development.

packages/calcite-components/package.json
benelan: If Volta adds another place to specify the development versions of node/npm then we can switch to that. But for now `package.json` is the only place to store pinned versions. We need to keep the `volta` field in CC's `package.json` so everyone who uses Volta will use the correct version of node. 

We are following the recommendation for monorepo setups in Volta's doc:
https://docs.volta.sh/advanced/workspaces

The `volta` field in our `package.json` has no impact on our consumers, even if they use Volta. They would need to pin a version on their end.

Can you add volta back here and then I'll approve the PR? Thanks!

alisonailea: after another discussion with Ben over teams I'm adding volta back to calcite-components

benelan (approved):
Awesome, thanks for the catch!

Issue #7953, opened 2023-10-05
stepper: No stepper-item is selected when first stepper-item is disabled


anveshmekala:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

If the first stepper-item is `disabled` then  no stepper-item is `selected` by default.

### Expected Behavior

If first `stepper-item` is selected, the next enabled `stepper-item` should be `selected`.

### Reproduction Sample

https://codepen.io/ANVESHMEKALA/pen/bGOQgJr?editors=1100

### Reproduction Steps

1. Open the CodePen.
2. Observe that second `stepper-item` is not selected.

### Reproduction Version

1.9.0

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

geospatialem:
Verified in `1.10.0-next.7`.

PR #8004, opened 2023-10-13
fix(stepper): selects next enabled stepper-item when first one is disabled

anveshmekala:
**Related Issue:** #7953 

## Summary

Selects next enabled `calcite-stepper-item` when the first one is disabled.

### DIFF
--- a/packages/calcite-components/src/components/stepper/stepper.e2e.ts
+++ b/packages/calcite-components/src/components/stepper/stepper.e2e.ts
@@ -677,7 +677,7 @@ describe("calcite-stepper", () => {
 
     const [stepperItem1, stepperItem2] = await page.findAll("calcite-stepper-item");
 
-    expect(stepperItem1).not.toHaveAttribute("selected");
-    expect(stepperItem2).toHaveAttribute("selected");
+    expect(await stepperItem1.getProperty("selected")).toBe(false);
+    expect(await stepperItem2.getProperty("selected")).toBe(true);
   });
 });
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/stepper/stepper.stories.ts
```diff
@@ -212,3 +212,9 @@ export const verticalLayout_TestOnly = (): string => html`<calcite-stepper layou
       >Step 1 Content Goes Here</calcite-stepper-item
     >
   </calcite-stepper>`;
+
+export const firstItemDisabled_TestOnly = (): string => html`<calcite-stepper>
```
jcfranco, line 216: Do we need this test if the E2E test already covers it?

jcfranco (approved):
Awesome!

Low-risk confirmed. ✅

packages/calcite-components/src/components/stepper/stepper.e2e.ts
```diff
@@ -663,4 +663,21 @@ describe("calcite-stepper", () => {
...
+        <div>Step 1 content</div>
+      </calcite-stepper-item>
+      <calcite-stepper-item heading="Step 2" id="step-2">
+        <div>Step 2 content</div>
+      </calcite-stepper-item>
+    </calcite-stepper>`);
+
+    const [stepperItem1, stepperItem2] = await page.findAll("calcite-stepper-item");
+
+    expect(stepperItem1).not.toHaveAttribute("selected");
```
jcfranco, line 680: Sorry I missed this, for consistency can you check the prop vs the attribute?

packages/calcite-components/src/components/stepper/stepper.stories.ts
anveshmekala: Since there are visual differences for selected stepper-item it is better to have a screenshot. I think we can utilize existing `disabled` story. Thoughts?

Issue #8006, opened 2023-10-13
Allow for overriding scales for `CalciteIcon`s


meescool:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

To be able to override CalciteIcon scales.

### Acceptance Criteria

To be able to override the CalciteIcon scales. This way we can account for when the base size for rem differs from the default and rescale them if necessary. For example we override it to 65.2%.

### Relevant Info

This is blocking us from upgrading to JSSDK 4.28 since they recently have refactored a couple widgets to use CalciteIcon components. We can't override any of these that are nested in other Calcite components due to the shadow dom. 
Please look at channel discussion [here](https://teams.microsoft.com/l/message/19:fd15b51dacd24e70895ec1218a54ae06@thread.skype/1697060245939?tenantId=aee6e3c9-711e-4c7c-bd27-04f2307db20d&groupId=56fae21a-9407-4943-859f-a9bfcf0bbad3&parentMessageId=1697060245939&teamName=Calcite%20Design%20System&channelName=Get%20Help%20-%20Development&createdTime=1697060245939&allowXTenantAccess=false)

We use Calcite Icons (just the icons) ourselves and we don't have any issues with their sizes and we'd like the consistent with any `CalciteIcon` components being used in the JSSDK.

### Which Component

CalciteIcon

### Example Use Case

There are cases where websites override the base size for rem to 65.2%, which is what we currently do. 
Here's a code pen of a JSSDK example.
https://codepen.io/meescool/pen/gOZQEja

![image](https://github.com/Esri/calcite-design-system/assets/43302795/5af3a556-9d5b-49bc-bcfd-6561464a66ed)

### Priority impact

p1 - need for current milestone

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS StoryMaps

jcfranco:
We'll be updating our icon component to scale based on icon asset scales (16px, 24px and 32px) and not use `rems`. Scale overrides will not be pursued at this time.

geospatialem:
Verified in `1.9.3-next.2`

![image](https://github.com/Esri/calcite-design-system/assets/5023024/4d437aa1-e5e6-491a-b1a7-08358c2f499b)

oevans:
[...]

PR #8009, opened 2023-10-13
fix(icon): use pixel sizes for icons

driskull:
**Related Issue:** #8006

## Summary

- Use fixed pixel sizes for icons
- Add screenshot test

### DIFF
--- a/packages/calcite-components/src/components/icon/icon.scss
+++ b/packages/calcite-components/src/components/icon/icon.scss
@@ -10,22 +10,29 @@
   @apply text-color-icon inline-flex;
 }
 
+$icon-size-s: 16px;
+$icon-size-m: 24px;
+$icon-size-l: 32px;
+
 :host([scale="s"]) {
-  @apply h-4 w-4;
-  min-inline-size: theme("spacing.4");
-  min-block-size: theme("spacing.4");
+  inline-size: $icon-size-s;
+  min-inline-size: $icon-size-s;
+  block-size: $icon-size-s;
+  min-block-size: $icon-size-s;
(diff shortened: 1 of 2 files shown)


jcfranco (approved):
💥

packages/calcite-components/src/components/icon/icon.scss
```diff
@@ -10,22 +10,29 @@
...
 
+$icon-size-s: 16px;
+$icon-size-m: 24px;
+$icon-size-l: 32px;
+
 :host([scale="s"]) {
-  @apply h-4 w-4;
-  min-inline-size: theme("spacing.4");
-  min-block-size: theme("spacing.4");
+  inline-size: $icon-size-s;
```
driskull, line 18: Not sure If I need the min sizes as well but the previous code had height and min size.

alisonailea: Min sizes are superfluous when width and height are pixels.

macandcheese: I think this prevents icon from squishing in cases where it doesn’t have room but may need to test that….

alisonailea: Most icon patterns I've seen set the width and height of the icon to 100% of its container and then size the container. In that case min- sizing would be important to prevent what you're describing but if width and height are explicit then min- sizing doesn't change anything

driskull: I'll leave it in so it doesn't break anything.

Issue #2093, opened 2021-05-04
Bug: Pagination - setting `pageSize` to 0 logs console error


macandcheese:
### Actual Behavior

Setting "pageSize" to 0 will crash the Storybook and cause the page to be killed (at least in Chrome)

### Relevant Info

https://codepen.io/mac_and_cheese/pen/eYLLbbN?editors=1000

Chrome Version 90.0.4430.93 (Official Build) (x86_64)

jcfranco:
I originally thought this was a storybook configuration issue (similar to https://github.com/Esri/calcite-components/issues/1127), but it looks like the problem is at the root of the component. I think we should update `pagination` to prevent `num` (number of items per page) being less than 1. @anveshmekala @macandcheese @driskull WDYT?

jcfranco:
Repro case: https://codepen.io/jcfranco/pen/dyJRrgB?editors=1000

anveshmekala:
Yup agreed. It is better to make sure `num` prop value is always greater than 1. Does this mean we mask the text to be one even if the user type in zero ?

macandcheese:
> I think we should update `pagination` to prevent `num` (number of items per page) being less than 1.

Agreed. I think it’s named to correlate to a specific ArcGIS data structure but ‘num’ as a prop name is a little confusing here - we don’t use shorthand names like that anywhere else

macandcheese:
In the component's latest state, the page doesn't crash, but the component does throw a console error : https://codepen.io/mac_and_cheese/pen/eYLLbbN?editors=1000

geospatialem:
Verified the component's `pageSize` is set to `1` if a user or change tries to modify the property to `0` (or less). ✨

PR #8017, opened 2023-10-16
fix(pagination): prevents console error when page-size is set to zero

anveshmekala:
**Related Issue:** #2093 

## Summary

Prevents console error messages when `page-size` is set to `0` via attribute or programmatically.

### DIFF
--- a/packages/calcite-components/src/components/pagination/pagination.tsx
+++ b/packages/calcite-components/src/components/pagination/pagination.tsx
@@ -113,13 +113,12 @@ export class Pagination
   @Prop({ reflect: true }) totalItems = 0;
 
   @Watch("totalItems")
-  handleTotalItems(): void {
-    this.totalPages = this.totalItems / this.pageSize;
-  }
-
   @Watch("pageSize")
-  handlePageSize(): void {
-    this.handleTotalPages();
+  handleTotalPages(): void {
+    if (this.pageSize < 1) {
+      this.pageSize = 1;
+    }
+    this.totalPages = this.totalItems / this.pageSize;
   }
 
   // --------------------------------------------------------------------------
@@ -301,13 +300,6 @@ export class Pagination
     this.emitUpdate();
   };
 
-  private handleTotalPages(): void {
-    if (this.pageSize < 1) {
-      this.pageSize = 1;
-    }
-    this.totalPages = this.totalItems / this.pageSize;
-  }
-
   //--------------------------------------------------------------------------
   //
   //  Render Methods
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/pagination/pagination.tsx
```diff
@@ -382,6 +393,12 @@ export class Pagination
 
   renderPage(start: number): VNode {
     const { pageSize } = this;
+
+    // if(pageSize === 0 ) {
```
driskull, line 397: remove comments :)

packages/calcite-components/src/components/pagination/pagination.tsx
```diff
@@ -297,6 +301,13 @@ export class Pagination
     this.emitUpdate();
   };
 
+  private handleTotalPages(): void {
+    if (this.pageSize < 1) {
+      this.pageSize = 1;
+    }
+    this.totalPages = this.totalItems / this.pageSize;
```
driskull, line 308: Would this be better? Do we need to set `this.pageSize` or can we just make sure totalItems is correct?

```
this.totalPages = this.totalItems / Math.max(this.pageSize, 1);
```

packages/calcite-components/src/components/pagination/pagination.tsx
```diff
@@ -112,12 +112,16 @@ export class Pagination
   /** Specifies the total number of items. */
   @Prop({ reflect: true }) totalItems = 0;
 
-  @Watch("pageSize")
   @Watch("totalItems")
-  handleTotalPages(): void {
+  handleTotalItems(): void {
     this.totalPages = this.totalItems / this.pageSize;
```
driskull, line 117: It seems like `handleTotalPages` and `handleTotalItems` does the same logic, anyway to combine them?

driskull (changes requested):
Had some refactoring feedback. 👍 ❔

packages/calcite-components/src/components/pagination/pagination.tsx
anveshmekala: how did i miss that 😵

packages/calcite-components/src/components/pagination/pagination.tsx
anveshmekala: Avoid setting `this.pageSize` value to `1` means leaving it to zero which can cause issues in other places where `pageSize` is used across the component. Otherwise using `Math.max` looks clean and helps us combine the `handlePageSize` & `handleTotalItems` into one method like it is in current version. Thoughts?

driskull: Ok, lets keep the page size validation fix and combine those two methods so that they can share the same function.

driskull (approved):
🎉 🌮

Issue #6981, opened 2023-05-18
[Segmented Control] `layout` inherits `"grid"` from shared resource


geospatialem:
### Actual Behavior

The segmented control's `layout` inherits `"grid"` from a [shared resource](https://github.com/Esri/calcite-components/blob/master/src/components/segmented-control/segmented-control.tsx#L32), but does not perform any layout action.

### Expected Behavior

We should exclude the `"grid"` layout option in the component from the shared resource, similar to [Stepper item](https://github.com/Esri/calcite-components/blob/98a880342149619c607e3c5719bee0b00b6d0fa8/src/components/stepper-item/stepper-item.tsx#L90).

### Reproduction Sample

https://codepen.io/geospatialem/pen/xxyyNzm

### Reproduction Steps

1. Open the [sample](https://codepen.io/geospatialem/pen/xxyyNzm)
2. Observe there is a `layout` prop option for `"grid"` in the third example
    - Also mentioned in our [docs](https://developers.arcgis.com/calcite-design-system/components/segmented-control/#api-reference-properties-layout)
3. Observe the value does nothing

### Reproduction Version

1.3.1

### Relevant Info

Surfaced in [Community](https://community.esri.com/t5/calcite-design-system-questions/grid-in-calcite-segmented-control-doesn-t-work/m-p/1290693#M400).

### Priority impact

p4 - not time sensitive

### Impact

Medium-ish priority

### Esri team

Calcite (dev)

macandcheese:
Need to extract the valid values here https://github.com/Esri/calcite-design-system/blob/429ed06e42a353b8f7b05daf08297ff19252a3af/packages/calcite-components/src/components/segmented-control/segmented-control.tsx#L87

geospatialem:
Verified in `main` on `segmented-control`'s `layout` prop with the extraction of just the `"horizontal"` and `"vertical"` values.

PR #8019, opened 2023-10-16
docs(segmented-control): removes grid option from layout

anveshmekala:
**Related Issue:** #6981


driskull (approved):
Just want to confirm, nobody should have been using `grid` right? it would be in an error state or something not good?

macandcheese:
> Just want to confirm, nobody should have been using `grid` right? it would be in an error state or something not good?

It just had no effect / associated styles - the component displayed as if the default value was set.

driskull:
Sounds good!

Issue #7250, opened 2023-06-30
Low color contrast in Button and Segmented Control in dark mode


Anastasiia-Boleiko:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-components/issues) to avoid duplicates

### Summary

The Calcite Button and the selected option of Calcite Segmented Control (appearance: solid) have low contrast Components in the dark mode.

WCAG 2.0 level AA requires a contrast ratio of at least 4.5:1 for normal text and 3:1 for large text. WCAG 2.1 requires a contrast ratio of at least 3:1 for graphics and user interface components (such as form input borders). WCAG Level AAA requires a contrast ratio of at least 7:1 for normal text and 4.5:1 for large text.

Large text is defined as 14 point (typically 18.66px) and bold or larger, or 18 point (typically 24px) or larger.

### Actual Behavior

In both cases, the `font-size: 14px`, `color: #353535` and `background: #009AF2 `
**Color contrast is 4.03.**

<img width="300" alt="Screenshot 2023-06-30 at 13 15 12" src="https://github.com/Esri/calcite-components/assets/119309703/39b0aeaa-765e-4a4e-bcf1-a4afc1646853">

### Expected Behavior

Proposal: Change the color to `color: var(--calcite-ui-text-inverse)` (#151515) for example.
**Color contrast is 6.**
<img width="300" alt="Screenshot 2023-06-30 at 17 20 06" src="https://github.com/Esri/calcite-components/assets/119309703/83f7ab02-5e35-409a-a415-815c1492802e">

### Reproduction Sample

https://codepen.io/anastasiia-zrh/pen/JjeNrRy?editors=100

### Reproduction Steps

1. Add a Segmented Control/Button to a page
2. Make sure that the page is in the dark mode
3. Make sure that the [appearance](url) parameter is set to` default (solid)`
4. Make sure that the `kind` parameter is set to ` default (brand)`

### Reproduction Version

v1.4.3

### Working W3C Example/Tutorial

https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html

### Priority impact

p4 - not time sensitive

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Urban

ashetland:
Preliminary digging showed that Segmented Control is set to ui-background instead of ui-text-inverse. Further digging need to confirm all configs on Button and Segmented Control.

[...]

PR #8036, opened 2023-10-20
fix(segmented-control-item): fix text color contrast

jcfranco:
**Related Issue:** #7250

## Summary

✨👀👍✨

### DIFF
--- a/packages/calcite-components/src/components/segmented-control-item/segmented-control-item.scss
+++ b/packages/calcite-components/src/components/segmented-control-item/segmented-control-item.scss
@@ -56,8 +56,7 @@
 }
 
 :host([checked]) label {
-  @apply bg-brand border-color-brand cursor-default;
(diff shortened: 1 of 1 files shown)


ashetland:
I accepted the unrelated screens, but it looks like something is off with the Dark Mode RTL Test Only page as chromatic didn't catch the change. When I look at storybook, that page is actually in Light Mode and LTR. If I click the canvas or just press a key it switches to Dark Mode, but is still LTR. All that said, the changes look good!

jcfranco:
@ashetland Great catch! 🎣 Looks like the dark + RTL story was using the wrong class. It should be fixed in the next build.

ashetland:
✨✨✨✨🧹✨✨✨✨

Elijbet (approved):
🔆 🕶 Great!

PR #7992, opened 2023-10-12
feat: add xxs breakpoint to tokens

alisonailea:
**Related Issue:** #7856

## Summary

generates a new SCSS variable `$breakpoint-width-xxs: 320px;`

Issue #7856, opened 2023-09-20
[Pagination] Responsive layout for specific XS layouts/dimensions


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

A follow-up issue to https://github.com/Esri/calcite-design-system/issues/6687 to address specific XS screen dimensions, where `pagination` should be sized to accommodate specific XS dimensions, as not all use cases fit the 475px breakpoint.

_Note_ the start/end buttons are not yet in the Calcite UI icons, Skye added a high priority ticket for them to be added. The current button icons are local SVGs (currently noted in Figma).

cc @SkyeSeitz for design expertise, specifics with the breakpoint needed, and follow-up when working through the responsive update.

### Acceptance Criteria

- The last responsive addition should be consulted with Skye
- [Figma file](https://www.figma.com/file/iCQBMp5Q5zlSfyE9NxkdO7/Pagination---responsive?type=design&node-id=608%3A56424&mode=dev)
- Screenshot and GIF:
<img width="638" alt="MicrosoftTeams-image (21)" src="https://github.com/Esri/calcite-design-system/assets/5023024/495e5bb9-a505-4c6d-8fb4-7e33786b2fbf">

https://github.com/Esri/calcite-design-system/assets/5023024/751b0b3f-ec9e-475c-a76f-029227ff38d6

### Relevant Info

Part of epic https://github.com/Esri/calcite-design-system/issues/6670, originated from https://github.com/Esri/calcite-design-system/issues/6687

### Which Component

`pagination`

### Priority impact

p2 - want for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

driskull:
A blocker for this would be adding a new breakpoint for a smaller size.

geospatialem:
> A blocker for this would be adding a new breakpoint for a smaller size.

@driskull Should we migrate this issue to October to work out the new breakpoint?

driskull:
@geospatialem Thats a good question for the design team, might be a good idea unless they can do it sooner.

geospatialem:
> A blocker for this would be adding a new breakpoint for a smaller size.

> Should we migrate this issue to October to work out the new breakpoint?

> Thats a good question for the design team, might be a good idea unless they can do it sooner.

[...]

PR #8041, opened 2023-10-23
chore(responsive): add xxs breakpoint prop

jcfranco:
**Related Issue:** #7856

## Summary

Updates `breakpoints` lookup object to include `xxs` (added in https://github.com/Esri/calcite-design-system/pull/7992).

### DIFF
--- a/packages/calcite-components/src/utils/responsive.spec.ts
+++ b/packages/calcite-components/src/utils/responsive.spec.ts
@@ -8,10 +8,11 @@ describe("getBreakpoints()", () => {
     document.head.innerHTML = `
       <style>
         :root {
-        --calcite-app-breakpoint-width-lg: 1000px;
-        --calcite-app-breakpoint-width-md: 100px;
-        --calcite-app-breakpoint-width-sm: 10px;
-        --calcite-app-breakpoint-width-xs: 1px;
+        --calcite-app-breakpoint-width-lg: 10000px;
+        --calcite-app-breakpoint-width-md: 1000px;
+        --calcite-app-breakpoint-width-sm: 100px;
+        --calcite-app-breakpoint-width-xs: 10px;
+        --calcite-app-breakpoint-width-xxs: 1px;
       }
       </style>
     `;
@@ -22,6 +23,7 @@ describe("getBreakpoints()", () => {
         medium: toBeInteger(),
         small: toBeInteger(),
         xsmall: toBeInteger(),
+        xxsmall: toBeInteger(),
       },
     });
   });
--- a/packages/calcite-components/src/utils/responsive.ts
+++ b/packages/calcite-components/src/utils/responsive.ts
@@ -4,6 +4,7 @@ export interface Breakpoints {
     medium: number;
     small: number;
     xsmall: number;
+    xxsmall: number;
   };
 }
 
@@ -35,6 +36,7 @@ export async function getBreakpoints(): Promise<Breakpoints> {
           medium: breakpointTokenToNumericalValue(rootStyles, "--calcite-app-breakpoint-width-md"),
           small: breakpointTokenToNumericalValue(rootStyles, "--calcite-app-breakpoint-width-sm"),
           xsmall: breakpointTokenToNumericalValue(rootStyles, "--calcite-app-breakpoint-width-xs"),
+          xxsmall: breakpointTokenToNumericalValue(rootStyles, "--calcite-app-breakpoint-width-xxs"),
         },
       });
     });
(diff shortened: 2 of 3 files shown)


jcfranco:
cc @driskull

packages/calcite-components/src/utils/responsive.spec.ts
```diff
@@ -8,10 +8,11 @@ describe("getBreakpoints()", () => {
...
         :root {
-        --calcite-app-breakpoint-width-lg: 1000px;
-        --calcite-app-breakpoint-width-md: 100px;
-        --calcite-app-breakpoint-width-sm: 10px;
-        --calcite-app-breakpoint-width-xs: 1px;
+        --calcite-app-breakpoint-width-lg: 10000px;
+        --calcite-app-breakpoint-width-md: 1000px;
+        --calcite-app-breakpoint-width-sm: 100px;
+        --calcite-app-breakpoint-width-xs: 10px;
+        --calcite-app-breakpoint-width-xxs: 1px;
```
alisonailea, line 15: Breakpoints as CSS Custom Props don't actually work. They can be queried by JS but this is not the ideal way to retrieve these variables. These tokens are also provided by calcite-design-tokens as CSS custom props. However the values do not align.

```css
--calcite-app-breakpoint-width-lg: 1440px; /* min-width size */
  --calcite-app-breakpoint-width-md: 1152px; /* min-width size */
  --calcite-app-breakpoint-width-sm: 768px; /* min-width size */
  --calcite-app-breakpoint-width-xs: 476px; /* min-width size */
  --calcite-app-breakpoint-width-xxs: 320px;
  ```

Or as SCSS variables

```scss
$breakpoint-width-lg: 1440px; // min-width size
$breakpoint-width-md: 1152px; // min-width size
$breakpoint-width-sm: 768px; // min-width size
$breakpoint-width-xs: 476px; // min-width size
$breakpoint-width-xxs: 320px; // Small handheld devices and mini-windows
```


The new PR ([8044](https://github.com/Esri/calcite-design-system/pull/8044)) also provides the same breakpoints now as a JS object 

```js
import tokenJSON from "@esri/calcite-design-tokens/js";

const breakpointXxs = tokenJSON.core.breakpoint.width.xxs.value;
```

or as tree-shakeable named exports

```js
import { CoreBreakpointWidthXxs } from "@esri/calcite-design-tokens/es6";
```

Do we still need this PR?

alisonailea (changes requested):
These breakpoints do not align with designer approved breakpoints

packages/calcite-components/src/utils/responsive.spec.ts
jcfranco: > Breakpoints as CSS Custom Props don't actually work. 

Can you provide more info on this? I'm not sure I follow.

> They can be queried by JS but this is not the ideal way to retrieve these variables. These tokens are also provided by calcite-design-tokens as CSS custom props. However the values do not align.

There are a few factors on why we get the breakpoints by evaluating the CSS at runtime:

1. no JS object at the time
2. unclear if we were allowing users to override breakpoints via CSS props or not
3. wanted to avoid introducing a third way to reference tokens within the code base

Depending on 2, we could revisit this module's implementation (in a follow-up PR) and set some guidelines for 3.

> These breakpoints do not align with designer approved breakpoints

FWIW, this test cover looking up the breakpoints CSS props and isn't meant to match the actual breakpoint values.

alisonailea: > Can you provide more info on this? I'm not sure I follow.

https://bholmes.dev/blog/alternative-to-css-variable-media-queries/

> FWIW, this test cover looking up the breakpoints CSS props and isn't meant to match the actual breakpoint values.

Got it! Thanks for the clarification.

alisonailea (approved):

packages/calcite-components/src/utils/responsive.spec.ts
alisonailea: Container Style Queries can be used to set breakpoint styles with CSS Custom Props but that's currently only available in chrome https://developer.chrome.com/blog/style-queries/

jcfranco: > While a consumer could override any custom prop including --calcite-app-breakpoint-width-lg it won't do them any good because CSS will not use custom props in media breakpoint queries as shown in the link above. This can be accomplished via SCSS but that would rely on the SCSS var $breakpoint-width-lg before compiling.
> Container Style Queries can be used to set breakpoint styles with CSS Custom Props but that's currently only available in chrome https://developer.chrome.com/blog/style-queries/

Ah, that's right. Thanks for catching that and clarifying.

The `responsive` util does pick up breakpoint CSS props overrides at load time, but we would still have a disconnect in a few places (namely stylesheet and conditional rendering) because we can't use style queries/environment variables yet. 

I'll update the util in a follow-up PR to use the JS exports instead. ✨🔧✨

Issue #7335, opened 2023-07-18
[List/List Item] Mitigate action access when toggling the `disabled` state


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Summary

Ensure all List Items and their actions are available to users when toggling the `disabled` states.

![a11y-listitem-disabled-actions](https://github.com/Esri/calcite-design-system/assets/5023024/b3b2f056-a0dd-4f34-beea-6fcd8b0e3232)

### Actual Behavior

When List Items `disabled` state is toggled, its actions remain inaccessible to keyboard users, while available to mouse users.

### Expected Behavior

Either no `tabindex` should be present via the interactive helper, or the `tabindex` should be updated when changing between the `disabled` state. [Additional context provided via Matt](https://github.com/Esri/calcite-design-system/issues/7254#issuecomment-1629396533) addressed in #7254.

### Reproduction Sample

https://codepen.io/geospatialem/pen/XWyzpXX

### Reproduction Steps

1. Open the [sample](https://codepen.io/geospatialem/pen/XWyzpXX)
2. Select the first item in the list by pressing the `Enter` key
3. Press the `Enter` key again to remove any disabled List items
4. Press `Tab` to navigate to the apple's action
5. Press `Tab` again and observe the next item is the mango's action

### Reproduction Version

1.5.0-next.17

### Relevant Info

https://github.com/Esri/calcite-design-system/issues/7254#issuecomment-1629396533

### Regression?

N/A

### Priority impact

p2 - want for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

Calcite (dev)

jcfranco:
[...]

PR #8042, opened 2023-10-23
fix(list-item): restore tabbability when an item's disabled prop is toggled

jcfranco:
**Related Issue:** #7335 

## Summary

Drop "managed" as the `hostIsTabbable ` argument since the parent does not control its tabbability (`tabindex`).

This also adds info to `updateHostInteraction`'s doc.

### DIFF
--- a/packages/calcite-components/src/utils/interactive.ts
+++ b/packages/calcite-components/src/utils/interactive.ts
@@ -88,7 +88,7 @@ const captureOnlyOptions = { capture: true } as const;
  * technically, users can override `tabindex` and restore keyboard navigation, but this will be considered user error
  *
  * @param component
- * @param hostIsTabbable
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/utils/interactive.ts
```diff
@@ -88,7 +88,7 @@ const captureOnlyOptions = { capture: true } as const;
  * technically, users can override `tabindex` and restore keyboard navigation, but this will be considered user error
  *
  * @param component
- * @param hostIsTabbable
+ * @param hostIsTabbable – when set to true or its predicate returns true, the host is tabbable and will be managed by the helper. Set to "managed" for cases where a component's parent determines which item is tabbable (i.e., sets `tabindex` to allow tabbing).
```
benelan, line 91: :writing_hand: :scroll: :trophy:

benelan (approved):
:+1:

Issue #7955, opened 2023-10-05
[combobox] add test coverage for #7934


jcfranco:
### Test type

new test

### Which Component(s)

calcite-combobox

### PR skipped, if applicable

#

### Additional Info

The issue is not consistently reproducible in E2E testing nor in OSX, so this will need some research with a Windows + trackpad machine.

geospatialem:
Verified in the `main` branch with the `does not clear filter if pointer down/up on an item has a delay in between events` test. 🧪

PR #8045, opened 2023-10-23
test(combobox): add test for filter being cleared if there's a delay between pointer down and up

jcfranco:
**Related Issue:** #7955 

## Summary

Adds missing test from #7934.


alisonailea (approved):
🧪

PR #8050, opened 2023-10-24
test: add token snapshot tests

alisonailea:
**Related Issue:** n/a

## Summary

Adds a debugger script for calcite-design-tokens in VSCode (launch.json)
Adds snapshot tests to enforce token output


jcfranco (commented):
Noice. I might be missing something, but I do have some questions:

1. What is the expected code change/PR review/test workflow? It's comparing against all output targets, so does this mean any change to tokens could potentially cause a test failure?
2. If ☝️ is yes, what is the difference between a. reviewing a change to the design tokens package source and b. accepting the diff from the updated snapshot? There might be a bit of redundancy between a and b.
3. What is the main objective of these tests? If it's to make sure the generated files are not impacted in a major way, would the components package testing help catch this type of change? If it's to make sure all changes get double checked, how can we make sure it won't get noisy and reviews will be effective?
4. Is there an alternative that wouldn't require us to keep a copy of all output targets as the base for snapshot tests? This would only grow as we add more tokens and output targets, right?

alisonailea:
> * What is the expected code change/PR review/test workflow? It's comparing against all output targets, so does this mean any change to tokens could potentially cause a test failure?

Yes any change to platform outputs will cause a test failure.

> * If ☝️ is yes, what is the difference between a. reviewing a change to the design tokens package source and b. accepting the diff from the updated snapshot? There might be a bit of redundancy between a and b.

Because changes may be made to the token transformer which would cause a change in the output even if the tokens themselves remain the same. This also helps future proof tokens so changes aren't accidentally made in Figma and pushed up. It's a lot easier to miss mistakes on the design side.

> * What is the main objective of these tests? If it's to make sure the generated files are not impacted in a major way, would the components package testing help catch this type of change? If it's to make sure all changes get double checked, how can we make sure it won't get noisy and reviews will be effective?

The main objective is to catch changes made to the platform output of tokens.

> * Is there an alternative that wouldn't require us to keep a copy of all output targets as the base for snapshot tests? This would only grow as we add more tokens and output targets, right?

Along with being the only way to fully test tokens prior to building calcite-components as well, keeping a static record of how we expect tokens to look is valuable because it allows us to review token output from Github without pulling the branch down and building locally.

alisonailea:
7984 relies on this PR for verification. It would also help for verifying 7986. And 7986, 7984, and this PR are required to accomplish 7325

jcfranco:
Thanks for the additional context. Agreed that this would be super useful for token test coverage. 🎉

I am curious to see what the update workflow ends up looking like since the snapshot test base is volatile.

jcfranco (approved):
👍

PR #8051, opened 2023-10-25
refactor(responsive): use JS module to create breakpoints lookup object

jcfranco:
**Related Issue:** N/A

## Summary

Stems from https://github.com/Esri/calcite-design-system/pull/8041#discussion_r1370773208. 

This changes how breakpoints are built and used (`responsive` module):

* uses design tokens JS output target to reference breakpoint values
* prevents possible disconnect between stylesheet and conditional rendering breakpoint values (only the latter could be overridden via CSS props)
* the breakpoints lookup object no longer needs to be obtained asynchronously

This also updates and unskips the `responsive` spec test.

**Note**: this includes an update to the testing Jest config to apply Stencil's transformer to the design token's headless ESM file as the default one was not transforming properly and leading to `Jest encountered an unexpected token` errors when running the spec test.


jcfranco:
@alisonailea I'm going to proceed with merging, but please LMK if you spot anything and I can address afterwards.

Issue #3280, opened 2021-10-15
Add Chromatic Test for text wrap  for block


anveshmekala:
### Description
Add a screener test to verify text wrap in `calcite-block-section`  
### Acceptance Criteria 
Parse a  long string or URL in the `calcite-block-section` and verfiy if the content is wrapped when the width of the `calcite-`block is set to `400px`
### Relevant Info 
Refer to issue #3069 
#### Which Component
`calcite-block-section`

jcfranco:
@anveshmekala Can you double check this is still needed? If not, we can close this.

geospatialem:
Verified on `main` branch

PR #8066, opened 2023-10-26
test(block): Add screenshot test for long Block and Block Section props

macandcheese:
**Related Issue:** #3280

### DIFF
--- a/packages/calcite-components/src/components/block/block.stories.ts
+++ b/packages/calcite-components/src/components/block/block.stories.ts
@@ -235,6 +235,40 @@ export const loadingWithNoStatusNorSlottedIcon_TestOnly = (): string =>
     </calcite-block>
   `;
 
+export const longWrappingTextInBlockAndBlockSection_TestOnly = (): string =>
+  html`
+    <calcite-panel style="width:250px">
+      <calcite-block
+        collapsible
+        open
+        heading="Planes, trains, and automobiles are some examples of modes of transportation"
+        description="Planes, trains, and automobiles are some examples of modes of transportation"
+      >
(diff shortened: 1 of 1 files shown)


macandcheese:
Chromatic link: https://www.chromatic.com/test?appId=6266d45509d7eb004aa254fb&id=653a97a872bb1a8cc020f92f

packages/calcite-components/src/components/block/block.stories.ts
```diff
@@ -235,6 +235,40 @@ export const loadingWithNoStatusNorSlottedIcon_TestOnly = (): string =>
     </calcite-block>
   `;
 
+export const longWrappingTextInBlockAndBlockSection_TestOnly = (): string =>
+  html`
+    <calcite-panel style="width:250px">
+      <calcite-block
+        collapsible
+        open
+        heading="Planes, trains, and automobiles are some examples of modes of transportation"
```
driskull, line 244: Might be nice in the future if we had a JSON file with sample content like this so all of our storybook screenshot tests could be consistent with good content. Maybe JSON file with `paragraph`, `sentence`, `snippet`, `heading`, etc. Could be something as a "good first issue" to cleanup.

driskull (approved):
👍

packages/calcite-components/src/components/block/block.stories.ts
macandcheese: Agreed `sampleText.long`, `sampleText.short` would be nice helpers to be able to reference. 

Or maybe a helper to generate "esri ipsum" with geography terms for a given character count... `getSampleText(400)`

Issue #5626, opened 2022-10-25
List Filter Control Label - Provide a valid label for form fields - (2036570437)


dqateam:
### Module: 
~39 Value List and Value List Item~
List and List Item

### Violation: 
 Provide a valid label for the component's filter list item

![image](https://user-images.githubusercontent.com/5023024/202567795-5b898332-e277-4076-840c-22b1607ed40e.png)

### WCAG Reference: 
- [1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html)
- [3.3.2 Labels or Instructions](https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html)
- [4.1.2 Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html) 

### Instance ID: 
 2036570437 

### Severity: 
 10 

### Description: 
 [Issue]
There is filter form control in the "Test 2. Value List with Filter" section without a programmatically associated label describing their purpose.

[User Impact]
Screen reader users and users with cognitive disabilities may be unable to determine what these controls are for. Speech input users will have difficulty navigating to them.

[Code Reference] 
 ```
<input aria-label="" class="" enterkeyhint="" inputmode="" placeholder="Filter list items" type="text">
 ``` 

### Note: 
 [Suggestion]
Provide a valid label for form fields. Provide aria-label attributes in the ``` <input> ``` elements.

[Compliant Code Example] 
 ```
<input aria-label="Filter list items" class="" enterkeyhint="" inputmode="" placeholder="Filter list items" type="text">
 ```
 

### Media Type: 
 Forms 

### Additional Resources: 
- Report Source: Product accessibility evaluation conducted on the Calcite Design System in August, 2022.

driskull:
This one would apply to list component as well.

geospatialem:
Updated to apply to the `list` component

driskull:
The fix should be to change from using `aria-label` to `label` within the `calcite-filter` component. 

https://github.com/Esri/calcite-design-system/blob/331247f250ad7bfb892cd77f32c6711c8365bacb/packages/calcite-components/src/components/filter/filter.tsx#L273

In the future, we could add a `filterLabel` property to specifically configure the list filter label.

geospatialem:
[...]

PR #8069, opened 2023-10-26
fix(filter): corrects the accessible label

driskull:
**Related Issue:** #5626

## Summary

- Correctly set the label for the filter component

### DIFF
--- a/packages/calcite-components/src/components/filter/filter.tsx
+++ b/packages/calcite-components/src/components/filter/filter.tsx
@@ -270,10 +270,10 @@ export class Filter
         <div class={CSS.container}>
           <label>
             <calcite-input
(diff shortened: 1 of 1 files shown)


geospatialem (commented):
This sounds a bit odd in JAWS mentioning the tree view item. It seems okay in NVDA, but doesn't provide a lot of context other than "filter", but maybe there's more context that could be added, such as "Filter list items", or perhaps the additional prop would help cover use cases. WDYT?

### JAWS
```
Filter tree view item 
```

### NVDA
```
Filter edit blank
```

driskull:
Agreed a new prop could be added in the future. I think as a separate issue.

jcfranco (approved):

Issue #7860, opened 2023-09-21
Add an Angular wrapper


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Determine feasibility and implement an Angular wrapper, similar to the React wrapper to support users working in the Angular framework.

This came up at DevSummit and UC, where users were asking for integration with their Angular applications.

### Acceptance Criteria

Criteria taken from [Stencil's docs](https://stenciljs.com/docs/angular) after installing the [`@stencil\angular-output-target`](https://stenciljs.com/docs/angular#adding-the-angular-output-target) dependency:
- Angular component wrappers will be detached from change detection, preventing unnecessary repaints of your web component.
- Web component events will be converted to RxJS observables to align with Angular's `@Output()` and will not emit across component boundaries.
- Optionally, form control web components can be used as control value accessors with Angular's reactive forms or `[ngModel]`.
- It is not necessary to include the [Angular `CUSTOM_ELEMENTS_SCHEMA`](https://angular.io/api/core/CUSTOM_ELEMENTS_SCHEMA) in all modules consuming your Stencil components.

### Relevant Info

Additional [framework integration](https://stenciljs.com/docs/overview) support from Stencil's documentation.

Could potentially mitigate the issue identified in https://github.com/Esri/calcite-design-system/issues/7729 cc: @andygup

### Which Component

All components

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

omarkawach:
Watching.

geospatialem:
Additional testing and review will take place later this week, and should be available as part of `next`. For now moving to the November milestone.

andygup:
Will the wrapper be aimed at the latest shipping version of Angular? Angular 17 is scheduled for release next week. I did some limited testing on it with @arcgis/core and didn't see any issues (other than the broken data binding).

We'll also need to do some basic sanity testing with [Angular Control Flow](https://blog.angular.io/meet-angulars-new-control-flow-a02c6eee7843) and @arcgis/core + components.

martinbojnansky:
[...]

PR #8084, opened 2023-10-30
build: add @esri/calcite-components-angular package

benelan:
**Related Issue:** #7860

## Summary

Add a package of angular components that wrap the web component library following [Stencil's documentation](https://stenciljs.com/docs/v2/angular).

There were a lot of type related build errors and I ended up adding two patches:

1. Rebuild the web component library if the CC types aren't generated correctly as a workaround for https://github.com/ionic-team/stencil/issues/3239
2. Fix a couple type import paths for event details that were generated by `@stencil/angular-output-target`. Not sure if this one is an issue on our end or theirs.

All of the build/CI stuff is working correctly. The outputs of `lerna version` and `npm --workspace @esri/calcite-components-angular pack` are both correct. It was a bit wonky at first because we need an angular workspace just to build a single component library.

I still need to do some testing in an example app, but I don't foresee any issues so the PR is ready for review. I'll add setup steps to the package's README after I walk through them while setting up the example.


packages/calcite-components-angular/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "name": "angular-workspace",
+  "private": true,
+  "scripts": {
+    "ng": "ng",
+    "prebuild": "./support/stencilDoubleBuildTypesWorkaround.sh && ts-node ./support/patchEventDetailTypeImports.ts",
+    "build": "ng build",
+    "clean": "rimraf dist node_modules .turbo .angular projects/component-library/dist"
+  },
+  "dependencies": {
+    "@angular/animations": "^16.2.0",
```
jcfranco, line 11: Can we pin these deps and have renovate handle them?

Applies to all `package.json` files in this PR.

jcfranco (approved):
This is looking great, @benelan!

I was curious about the `projects/component-library/` folders. Is this required for 👇?

> All of the build/CI stuff is working correctly. The outputs of lerna version and npm --workspace @esri/calcite-components-angular pack are both correct. It was a bit wonky at first because we need an angular workspace just to build a single component library.

benelan:
> I was curious about the projects/component-library/ folders. 

Yup, that's related to using an Angular workspace, which is how the [Stencil doc says to do it](https://stenciljs.com/docs/v2/angular#creating-an-angular-component-library) and how the [ArcGIS Web Components monorepo does it](https://devtopia.esri.com/WebGIS/arcgis-web-components/blob/main/packages/wrapper-packages/angular-components/projects/map-components-angular).

There may be a way to set this up without using an Angular workspace, but as someone who has never used Angular I wouldn't be the one to do it. I just followed the doc :man_shrugging:

benelan:
Here is Angular's doc about their workspace file structure: https://angular.io/guide/file-structure

It doesn't seem like there are any other options for developing "shareable libraries", which is what we are doing.

> Angular also supports workspaces with [multiple projects](https://angular.io/guide/file-structure#multiple-projects). This type of development environment is suitable for advanced users who are developing [shareable libraries](https://angular.io/guide/glossary#library), and for enterprises that use a "monorepo" development style, with a single repository and global configuration for all Angular projects.

Here is the Angular doc about creating libraries: https://angular.io/guide/creating-libraries

packages/calcite-components-angular/package.json
andygup: Nice work @benelan!  What's your plan for updating Angular deps, e.g. each Calcite release? Mainly wondering how Calcite, Components and @arcgis/core can stay roughly in sync. For example, Angular 17 is being released next week.

andygup: @jcfranco @benelan thoughts on this? Are you going with @latest, for each Calcite release? I don't know how 'renovate' is tied into the process. 

Also, will there be a `next` release of this that I can test on?

benelan: Hi @andygup! Yeah renovate will submit a PR to bump the angular dependencies when new versions come out. 

And yes, a `next` release will be deployed automatically once there are "deployable" changes. I'll let you know when one goes out. Like the react wrapper, the angular wrapper's version will always match CC's to prevent confusion.

andygup:
> There may be a way to set this up without using an Angular workspace

Using an Angular workspace is the recommended, but non-intuitive, way to go.

packages/calcite-components-angular/projects/component-library/README.md
```diff
@@ -0,0 +1,89 @@
...
+<calcite-slider min="1" max="100" [value]="sliderValue" (calciteSliderInput)="onSliderInput($event)"></calcite-slider>
+```
+
+## Contributing
+
+We welcome contributions to this project. See [CONTRIBUTING.md](https://github.com/Esri/calcite-design-system/blob/main/CONTRIBUTING.md) for an overview of the guidelines.
+
+## License
+
+COPYRIGHT © 2021 Esri
```
jcfranco, line 79: Can we update the copyright? 

Sidebar: we should check/update all license entries.

packages/calcite-components-angular/projects/component-library/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "peerDependencies": {
+    "@angular/common": "^16.2.0",
+    "@angular/core": "^16.2.0",
+    "@esri/calcite-components": "1.10.0"
+  },
+  "devDependencies": {
+    "@esri/calcite-components": "1.10.0"
+  },
+  "dependencies": {
+    "tslib": "2.3.0"
```
jcfranco, line 13: Is this needed by Angular or can we update the `tsconfig.json` target to avoid this dependency?

benelan: It is added by the angular cli so I'm pretty sure it's required

PR #8074, opened 2023-10-27
fix(combobox-item): do not show icon for disabled item 

BrunoCaimar:
**Related Issue:** #7719 

## Summary

Do not show the icon if the combobox-item is disabled.

Issue #7719, opened 2023-09-08
Disabled Combobox Items reveal icon onHover


ashetland:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When hovering over a disabled Combobox Item, the `circle-disallowed` icon appears.

### Expected Behavior

I would expect no interaction on `disabled` items and thus no icon should appear.

### Reproduction Sample

https://codepen.io/ashetland/pen/YzdZPEM

### Reproduction Steps

Hover over `disabled` items in the open Combobox.

### Reproduction Version

v1.8.0

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

Calcite (design)

jcfranco:
@BrunoCaimar Is helping us with this one! 🎉

jcfranco:
Thanks again @BrunoCaimar for helping us out with this! 🏆

geospatialem:
Verified in `1.11.0-next.3`. 🎉

PR #8095, opened 2023-10-31
fix(combobox-item): hide disabled item icon

jcfranco:
# Adopted https://github.com/Esri/calcite-design-system/pull/8074
 
cc @BrunoCaimar
 
---
 
**Related Issue:** #7719 

## Summary

Do not show the icon if the combobox-item is disabled.


benelan (approved):
:+1:

Issue #7698, opened 2023-09-07
Calcite-block-section does not account for switch space when rendering text


ian13262:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When rendering text, `calcite-block-sections` do not account for the space taken up by a switch. Text should wrap on an additional line in this situation.
![image](https://github.com/Esri/calcite-design-system/assets/131201379/fca7963a-c535-43f4-ad77-b7fba2a54735)

![image](https://github.com/Esri/calcite-design-system/assets/44481292/4c5b64ee-095e-4bd5-857a-d0f81f43e8a8)

### Expected Behavior

Text should stop _before_ the switch and continue onto the next line if necessary.

### Reproduction Sample

https://codepen.io/iyouth66/pen/gOZgomO?editors=1100

### Reproduction Steps

(see codepen)

### Reproduction Version

1.7.0

### Relevant Info

N/A

### Regression?

`1.6.1`

### Priority impact

p2 - want for current milestone

### Impact

Smart mapping currently suffers from this issue:
![image](https://github.com/Esri/calcite-design-system/assets/131201379/2b6654ae-617d-4170-9bc4-be8c9e141a3b)

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Online

geospatialem:
Investigate if using the `stack` component could mitigate the above.

AdelheidF:
Could this be get fixed before January? It's all over the place ...

![image](https://github.com/Esri/calcite-design-system/assets/44481292/fd843a29-237f-4022-962e-f8db2222146f)

geospatialem:
@AdelheidF @ian13262 It looks like this is a regression [between `1.6.1` and `1.7.0`](https://github.com/Esri/calcite-design-system/compare/%40esri/calcite-components%401.6.1...%40esri/calcite-components%401.7.0) as a result of accessibility improvements to the component - added context to the issue summary above. 

More context:
- Works in `1.7.0-next.21`: https://codepen.io/geospatialem/pen/VwgaboK
- Doesn't work in `1.7.0-next.22`: https://codepen.io/geospatialem/pen/bGzpWPL
- Looks like a result of commit https://github.com/Esri/calcite-design-system/commit/1f44f6b0e05bcaf708015432ef64c8cb660101bd


[...]

PR #8101, opened 2023-11-01
fix(block-section): wraps long text over to a new line when toggle switch is displayed

anveshmekala:
**Related Issue:** #7698 

## Summary

Wraps long text to a new line when `toggle-display='switch'`

### DIFF
--- a/packages/calcite-components/src/components/block-section/block-section.scss
+++ b/packages/calcite-components/src/components/block-section/block-section.scss
@@ -39,9 +39,6 @@
     px-0
     py-2;
 
-  &:focus {
-    @apply focus-outset;
-  }
   &:hover {
     @apply text-color-1;
   }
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/block-section/block-section.scss
```diff
@@ -62,6 +62,7 @@
 
 .toggle--switch-container {
   @apply flex items-center relative bg-transparent w-full;
+  word-break: break-word;
```
driskull, line 65: Is there no tailwind option for this?

anveshmekala: nope, currently only `break-all` , `normal`  &  `keep-all` are supported

driskull (approved):
👍 🎉

PR #8107, opened 2023-11-02
ci(renovate): run on weekdays and don't pin peerDeps

benelan:
**Related Issue:** #8102

## Summary
We don't want to pin peer deps, because our users should be allowed to update the version of packages they use in their app w/o throwing errors due to calcite. Also, increasing the frequency of renovate updates to ensure we don't fall behind.


driskull (approved):
👍

Issue #7861, opened 2023-09-21
Upgrade Stencil to version 4


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Upgrade Stencil from v2 to v4, which include the following breaking changes:
- [Version 3](https://github.com/ionic-team/stencil/blob/main/BREAKING_CHANGES.md#stencil-v300)
- [Version 4](https://stenciljs.com/docs/introduction/upgrading-to-stencil-four)

A more comprehensive list can be found on [GitHub](https://github.com/ionic-team/stencil/blob/v4.0.0/BREAKING_CHANGES.md).

### Acceptance Criteria

Upgrade to the latest version of Stencil to support users and teams also using Stencil in their projects.

### Relevant Info

- Since Calcite has come out of beta, Stencil has already bumped 2 versions (they are on version 4, Calcite is on version 2)
- This means that any projects building on top of the Calcite Components are using version 2 and that puts us at in a little bit of a bind because Stencil is focusing all their efforts on the latest release, so if we want to pick up any of the most recent fixes or features, we need to be on version 4
- To be cautious, Calcite is treating the update as a breaking change because with all the the projects building on top of Calcite Components and Stencil. As a result there might be some things that we're not considering that might pop up once we make the change.

### Which Component

All components

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [X] @esri/calcite-components-react
- [X] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

omarkawach:
Watching

Issue #6696, opened 2023-03-28
Add support for `getAssetPath` to `@esri/calcite-components/dist/components/index`


dasa:
### Description

Can `getAssetPath` be exported at `@esri/calcite-components/dist/components/index`? I'd like to be able to call it to see if it throws an error. Even better would be a better way to see if the assets path has been set or not, but I don't see one. Related issue: https://github.com/ionic-team/stencil/issues/3208

### Acceptance Criteria

Function is available.

### Relevant Info

`setAssetPath` is already exported.

### Which Component

N/A

### Priority impact

p2 - want for current milestone

### Esri team

ArcGIS Maps SDK for JavaScript

PR #5233, opened 2022-08-27
build: patch rollup tree shaking stencil issue

benelan:
**Related Issue:** https://github.com/ionic-team/stencil/issues/3470

## Summary

It looks like Stencil wont fix the rollup tree shaking bug until v3 ([source](https://github.com/ionic-team/stencil/pull/3556)), so I'm adding another post-build patch until then. 

I looked through the [breaking PR](https://github.com/ionic-team/stencil/commit/ff0e8cc54e5e68631cd83302d59b19f2626d43cb#diff-474e24df74c0221a170d7c067c32acfa2501ed400f4694df986bd84ddd43b7ee), and it didn't look like any changes occured that would cause problems due to removing the component exports in `dist/components/index.js`. To be safe I tested with the [CC rollup example](https://github.com/Esri/calcite-components-examples/tree/master/rollup), `npm link`, and my [`build-sizes`](https://github.com/benelan/build-sizes) package. 

Using the latest version of Stencil (which contains the tree-shaking issue) these are the build sizes:

```
-----------------------------
|> Application Build Sizes <|
-----------------------------
Build
 --> file count: 2548
 --> size: 2.19 MB
 --> size on disk: 2.31 MB
-----------------------------
Main JS bundle
 --> name: main.js
 --> size: 1.07 MB
 --> gzip size: 181.39 KB
 --> brotli size: 130.63 KB
-----------------------------
```
And these were the build sizes after applying the patch:

```
-----------------------------
|> Application Build Sizes <|
-----------------------------
Build
 --> file count: 2548
 --> size: 1.25 MB
 --> size on disk: 1.38 MB
-----------------------------
Main JS bundle
 --> name: main.js
 --> size: 138.4 KB
 --> gzip size: 25.45 KB
 --> brotli size: 21.35 KB
-----------------------------
```

There were no build or runtime errors when using the patched build and the components rendered correctly. So the patch seems safe and effective.

cc @dasa

PR #8108, opened 2023-11-02
build(deps)!: bump Stencil to v4

jcfranco:
**Related Issue:** #7861 

## Summary

Bumps Stencil to v4.

### Noteworthy changes

* add `components.d.ts` to source control per [Stencil recommendation](https://stenciljs.com/docs/typed-components)
* add workaround for JSX imports not matching what Stencil React output target expects
* dropped workaround for https://github.com/Esri/calcite-design-system/issues/6696 (fixed in v4)
* drop obsolete tree-shaking patch (#5233) (fixed in v4)
* skips additional locales that are showing formatting differences between test/node and browser environments
* updates `volta` and `engines` fields from `eslint-plugin-calcite-components`' `package.json` to match other packages
* dep changes:
  * bumps `@typescript-eslint`†, `angular`†, `jest`†, `puppeteer`, `typescript` to latest († and related deps)
  * removed unused `@stencil/state-tunnel`
  * replaced `@stencil/postcss` with `@stencil-community/postcss`
  * added missing `globby` and `jsdom`
  * override `@jest/transform` to ensure node types match

BREAKING CHANGE: We are treating this core version bump as a precautionary measure, particularly due to its potential impact on projects using `calcite-components` and Stencil.


benelan (approved):
Awesome! [This patch](https://github.com/Esri/calcite-design-system/blob/jcfranco/7861-bump-stencil-to-v4/packages/calcite-components-angular/support/stencilDoubleBuildTypesWorkaround.sh) can also be removed now that we committed `components.d.ts`.

Otherwise LGTM!

jcfranco:
> Awesome! [This patch](https://github.com/Esri/calcite-design-system/blob/jcfranco/7861-bump-stencil-to-v4/packages/calcite-components-angular/support/stencilDoubleBuildTypesWorkaround.sh) can also be removed now that we committed components.d.ts.

Noice. I'll remove that patch in a follow-up PR.

jcfranco:
Skipping screenshots before installing as last change was specific to E2E tests and previous screenshot tests were ✅.

driskull (commented):
👍 Nice!

ffaubry:
@jcfranco I'm adding question here even if this PR is merged. Let me know there is a better place.

- You are using a fix version for stencil core. Why not using ^4.7.1? 
- You move the angular wrapper to 17. What's your thoughts about angular 16?


BTW, we stopped using ts-node and we are using tsx. Faster IMO.

jcfranco:
> You are using a fix version for stencil core. Why not using ^4.7.1?

IIRC, mainly to ensure a consistent test/debug environment and to reduce merge conflicts in `package-lock.json`.

> You move the angular wrapper to 17. What's your thoughts about angular 16?

We thought we would use the latest since we were updating other dependencies, but we could downgrade to v16. @benelan WDYT?

> BTW, we stopped using ts-node and we are using tsx. Faster IMO.

Noted. We will look into this and possibly make the switch too.

Issue #5625, opened 2022-10-25
List and List Item - Indicate live regions for dynamically changing content when filtering items - (2036570436)


dqateam:
### Module: 
~39 Value List and Value List Item~ 
List and List Item

### Violation: 
 Indicate live regions for dynamically changing content. This applies to the component's filtering, to indicate filtered items have changed.

![image](https://user-images.githubusercontent.com/5023024/202568186-c3ad762a-1349-4fe6-a1a2-255b94e8a97b.png)

### WCAG Reference: 
- [1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html) 

### Instance ID: 
 2036570436 

### Severity: 
 9 

### Description: 
 [Issue]
There are dynamically changing content areas when enter letter(s) in the filter input field in the "Test 2. Value List with Filter" section that do not indicate live regions.

[User Impact]
Screen reader users will be unaware of important updates to this content when they occur.

[Code Reference] 
 ```
<input aria-label="" class="" enterkeyhint="" inputmode="" placeholder="Filter list items" type="text">
 ``` 

### Note: 
 [Suggestion]
Indicate live regions for dynamically changing content. Live regions can be created by adding a role attribute set to "log", "status", "alert", "progressbar", "marquee", or "timer" as appropriate. Alternatively, custom behavior can be created by using the aria-live, aria-atomic, and aria-relevant attributes. Text injected into this live region element will be announced by screen readers.

Importantly, the element with the ARIA live attributes must be available when the page loads. Otherwise, many screen readers will not detect updates to the element. Additionally, the element must be empty on page load unless an immediate screen reader announcement is desired.

 

### Media Type: 
 Live Regions 

### Additional Resources: 
- Report Source: Product accessibility evaluation conducted on the Calcite Design System in August, 2022.

driskull:
This one would apply to list component as well.

geospatialem:
Here's an example of what we could implement when items are filtered: https://codepen.io/geospatialem/pen/KKrjwbY

The solution could use a combination of `aria-live="polite"` and [`aria-relevant="all"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-relevant) (Note the Codepen example above shows `"additions"`, but in our case we would also have removals).

The JAWS transcript for when interacting with the change:
```
add another message to chat above Button 
To activate press Enter.

[...]

PR #8125, opened 2023-11-04
chore: add aria-live filtered items functional component.

driskull:
**Related Issue:** #5625

## Summary

WIP

### DIFF
--- a/packages/calcite-components/src/components/functional/FilteredItems.tsx
+++ b/packages/calcite-components/src/components/functional/FilteredItems.tsx
@@ -1,0 +1,16 @@
+import { FunctionalComponent, h } from "@stencil/core";
+
+interface ExpandToggleProps {
+  filterEnabled: boolean;
+  items: string[];
+}
+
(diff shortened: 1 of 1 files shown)


packages/calcite-components/src/components/functional/FilteredItems.tsx
```diff
@@ -0,0 +1,16 @@
+import { FunctionalComponent, h } from "@stencil/core";
+
+interface ExpandToggleProps {
+  filterEnabled: boolean;
+  items: string[];
+}
+
+export const ExpandToggle: FunctionalComponent<ExpandToggleProps> = ({ filterEnabled, items }) => {
```
driskull, line 8: @anveshmekala @jcfranco is there any mechanism for fectching some common message strings from a functional component?

If not, I can just pass in messages from the list component but that kind of defeats the benefit of the functional component and I would just have this in the List component itself.

driskull: The strings would be pretty generic like `"Filter enabled"`, `"Displaying {num} Items"`, `"Item {x}: My Item"`.

Issue #7692, opened 2023-09-06
Split button `width` appears to size relative to content instead of container


nwhittaker:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Given a split button with a width of `full`, the button does not grow to fill the width of its container:

<img width="853" alt="Screenshot 2023-09-06 at 5 43 14 PM" src="https://github.com/Esri/calcite-design-system/assets/421496/2081801a-636f-4027-825e-cb5014647c1b">

Additionally, a width of `half` appears to size the button to half the width of its content, truncating its text content.

### Expected Behavior

Given a split button with a width of `full`, the button grows to fill the width of its container:

<img width="853" alt="Screenshot 2023-09-06 at 5 43 34 PM" src="https://github.com/Esri/calcite-design-system/assets/421496/05e7e2e5-4274-406e-b4bd-08b234c35500">

A width of `half` should size the button to half the width of its container.

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/split-button/

### Reproduction Steps

1. Visit the split button's doc page and set the `width` attribute to `full` or `half` in the **Sample** area.
2. Visit the regular button's doc page and compare its behavior when setting the `width` attribute.

### Reproduction Version

1.7.0

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Field Apps

geospatialem:
Spike to determine the working instance with `button`, and where `split-button` may need additional effort for consistency. An estimate and next steps will be provided, with implementation in a future milestone.

geospatialem:
Spike efforts might have similar groupings/relationships to https://github.com/Esri/calcite-design-system/issues/5903.

Consider tackling as a research effort in coordination.

jcfranco:
@geospatialem Sidebar, it looks like the brand new spike action isn't clearing all labels prior to assigning `0 - new`.

geospatialem:
> @geospatialem Sidebar, it looks like the brand new spike action isn't clearing all labels prior to assigning `0 - new`.

Ah, yes, this looks like its because it was set to `2 - in development`. @jcfranco Is that a current pattern we should account for, in addition to `1 - assigned`? I can update the script to account for both scenarios.

geospatialem:
[...]

PR #8133, opened 2023-11-06
fix(split-button): fix width layout

jcfranco:
**Related Issue:** #7692

## Summary

This fixes an issue where styles related to `width` were being applied internally and had no impact on the component.


driskull (approved):
👍

Issue #7546, opened 2023-08-17
Borders inconsistent when list comprises both grouped and ungrouped list items


nwhittaker:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Given a list containing a mix of list items and grouped list items:
1. A border _does not_ appear before an item if the previous item is in a group.
2. A border _does_ appear before a group heading if the previous item is not in a group.

<img width="776" alt="Screenshot 2023-08-17 at 4 16 46 PM" src="https://github.com/Esri/calcite-design-system/assets/421496/874eba6f-fe2d-4cac-814d-4bd5f8fed0ec">

### Expected Behavior

I expected borders to behave as they do in non-hybrid lists:
1. A border _always_ appears before an item unless it's the first item.
2. A border _never_ appears before a group heading.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/LYMYpGj

### Reproduction Steps

1. Visit reproduction sample
2. Verify no border appears between the **Waterfalls** and **Rivers** items
3. Verify a border appears before the **Building** group header

### Reproduction Version

1.6.0

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Field Apps

geospatialem:
Verified on the `main` branch:  
![image](https://github.com/Esri/calcite-design-system/assets/5023024/f9169a11-8c99-4b28-a702-6424a9c47939)

nwhittaker:
@geospatialem, I updated the [repro sample][1] to `1.11.0`. I see the border is gone from above the Buildings header, but I don't see a border between the Waterfalls and Rivers list items. Fair enough if this is the expected design, but I wanted to double-check…

<img width="283" alt="Screenshot 2023-11-27 at 10 27 01 AM" src="https://github.com/Esri/calcite-design-system/assets/421496/6b4c9c2d-1e35-45c2-99e6-c0db2f9937fa">

[1]: https://codepen.io/nwhittaker-esri/pen/LYMYpGj

anveshmekala:
[...]

PR #8134, opened 2023-11-06
fix(list-item): adds border between grouped and ungrouped list-items

anveshmekala:
**Related Issue:** #7546 

## Summary

Adds border to the `calcite-list-item` excluding the first `calcite-list-item` in the `calcite-list-item-group` or the `calcite-list`.

### DIFF
--- a/packages/calcite-components/calcite-preset.ts
+++ b/packages/calcite-components/calcite-preset.ts
@@ -165,7 +165,7 @@ export default {
         "2-lg": "0 12px 32px -2px rgba(0, 0, 0, 0.1), 0 4px 20px 0 rgba(0, 0, 0, 0.08)",
         "2-sm": "0 2px 12px -4px rgba(0, 0, 0, 0.2), 0 2px 4px -2px rgba(0, 0, 0, 0.16)",
         "border-bottom": "0 1px 0 var(--calcite-ui-border-3)",
-        "border-top": "0px -1px 0 var(--calcite-ui-border-3)",
+        "border-top": "0 -1px 0 var(--calcite-ui-border-3)",
         "outline-active": "0 0 0 1px var(--calcite-ui-brand)",
         none: "none",
         xs: "0 0 0 1px rgba(0, 0, 0, 0.05)",
--- a/packages/calcite-components/calcite-preset.ts
+++ b/packages/calcite-components/calcite-preset.ts
@@ -165,6 +165,7 @@ export default {
         "2-lg": "0 12px 32px -2px rgba(0, 0, 0, 0.1), 0 4px 20px 0 rgba(0, 0, 0, 0.08)",
         "2-sm": "0 2px 12px -4px rgba(0, 0, 0, 0.2), 0 2px 4px -2px rgba(0, 0, 0, 0.16)",
         "border-bottom": "0 1px 0 var(--calcite-ui-border-3)",
+        "border-top": "0px -1px 0 var(--calcite-ui-border-3)",
         "outline-active": "0 0 0 1px var(--calcite-ui-brand)",
         none: "none",
         xs: "0 0 0 1px rgba(0, 0, 0, 0.05)",
--- a/packages/calcite-components/src/components/list-item-group/list-item-group.scss
+++ b/packages/calcite-components/src/components/list-item-group/list-item-group.scss
@@ -20,11 +20,7 @@
   margin-block-start: 1px;
 }
 
-::slotted(calcite-list-item:first-of-type) {
-  @apply shadow-none;
-}
-
-// removes shadow for the first item in filteredItems of the group.
+// removes shadow for the first item of the group for both filtered and unfiltered items.
 ::slotted(calcite-list-item:nth-child(1 of :not([hidden]))) {
   @apply shadow-none;
 }
--- a/packages/calcite-components/src/components/list-item-group/list-item-group.scss
+++ b/packages/calcite-components/src/components/list-item-group/list-item-group.scss
@@ -23,6 +23,7 @@
 // removes shadow for the first item of the group for both filtered and unfiltered items.
 ::slotted(calcite-list-item:nth-child(1 of :not([hidden]))) {
   @apply shadow-none;
+  margin-block-start: 0px;
 }
 
 @include base-component();
--- a/packages/calcite-components/src/components/list-item-group/list-item-group.scss
+++ b/packages/calcite-components/src/components/list-item-group/list-item-group.scss
@@ -20,7 +20,7 @@
   margin-block-start: 1px;
 }
 
-::slotted(calcite-list-item:first-child) {
+::slotted(calcite-list-item:first-of-type) {
   @apply shadow-none;
 }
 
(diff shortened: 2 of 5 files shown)


driskull (approved):
Code changes look good. 🎉

ashetland:
Based on what was documented in the issue, I think this isn't quite fixing what was asked. I noted on a couple screenshots borders that now appear before the first item, but the issue requested that not happen.

anveshmekala:
> Based on what was documented in the issue, I think this isn't quite fixing what was asked. I noted on a couple screenshots borders that now appear before the first item, but the issue requested that not happen.

Agreed, observed that fix is causing an regression when `filter-enabled`.

ashetland:
Do we need to add a story to mimic the config in the issue? Not sure if that actually gets covered in our current tests.

anveshmekala:
> Do we need to add a story to mimic the config in the issue? Not sure if that actually gets covered in our current tests.

All the stories which has groups and/or nesting would cover the config mentioned the issue. Don't think we need a new one.

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -742,6 +748,51 @@ export class List implements InteractiveComponent, LoadableComponent, SortableCo
...
+    const topLevelAnchor = this.getTopLevelAncestorItemElement(firstFilteredItem);
+    if (topLevelAnchor) {
+      this.topLevelAncestorsMap.set(firstFilteredItem, topLevelAnchor);
+    }
+
+    filteredItems[0].setAttribute("data-filter", "0");
+    this.topLevelAncestorsMap.get(filteredItems[0])?.setAttribute("data-filter", "0");
+  };
+
+  private getTopLevelAncestorItemElement = (
```
jcfranco, line 778: I think you could use the [`walkUpAncestry`](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/utils/dom.ts#L197C23-L197C31) util for this.

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -742,6 +748,51 @@ export class List implements InteractiveComponent, LoadableComponent, SortableCo
...
+    }
+
+    filteredItems.forEach((item) => {
+      if (item.hasAttribute("data-filter")) {
+        item.removeAttribute("data-filter");
+      }
+    });
+
+    const firstFilteredItem = filteredItems[0];
+    const topLevelAnchor = this.getTopLevelAncestorItemElement(firstFilteredItem);
```
jcfranco, line 769: Per our convo, this `Anchor` ➡️ `Ancestor`.

packages/calcite-components/src/components/list-item-group/list-item-group.scss
```diff
@@ -16,10 +16,16 @@
...
+  @apply shadow-border-top;
+  margin-block-start: 1px;
 }
 
-::slotted(calcite-list-item:last-child) {
+::slotted(calcite-list-item:first-of-type) {
+  @apply shadow-none;
+}
+
+// removes shadow for the first item in filteredItems of the group.
```
jcfranco, line 27: Not sure if the comment or the selector needs updating, but this would affect both filtered and unfiltered items, no?

packages/calcite-components/calcite-preset.ts
```diff
@@ -165,6 +165,7 @@ export default {
         "2-lg": "0 12px 32px -2px rgba(0, 0, 0, 0.1), 0 4px 20px 0 rgba(0, 0, 0, 0.08)",
         "2-sm": "0 2px 12px -4px rgba(0, 0, 0, 0.2), 0 2px 4px -2px rgba(0, 0, 0, 0.16)",
         "border-bottom": "0 1px 0 var(--calcite-ui-border-3)",
+        "border-top": "0px -1px 0 var(--calcite-ui-border-3)",
```
jcfranco, line 168: You can drop `px` from 0-length units.

packages/calcite-components/src/components/list-item-group/list-item-group.scss
anveshmekala: Yup it affects both filtered and unfiltered items. will update the comment and remove the selector above `::slotted(calcite-list-item:first-of-type)` which is redundant.

ashetland:
LGTM 🎉🎉🎉🎉🎉🎉🎉🚀🚀🚀🚀🚀🚀🚀🚀🚀🎉🎉🎉🎉🎉🎉🎉🚀🚀🚀🚀🚀🚀🚀🚀🚀🎉🎉🎉🎉🎉🎉🎉

packages/calcite-components/src/components/list-item-group/list-item-group.scss
```diff
@@ -16,11 +16,14 @@
...
-  @apply shadow-border-bottom mb-px;
+  @apply shadow-border-top;
+  margin-block-start: 1px;
 }
 
-::slotted(calcite-list-item:last-child) {
+// removes shadow for the first item of the group for both filtered and unfiltered items.
+::slotted(calcite-list-item:nth-child(1 of :not([hidden]))) {
   @apply shadow-none;
+  margin-block-start: 0px;
```
jcfranco, line 26: Nitpick: `px` not needed here.

packages/calcite-components/src/components/list/list.tsx
jcfranco: Per our convo, we can look at updating some of our `dom` utils to help cover this use case later (i.e., recursive `closest` without crossing shadow DOM).

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -742,6 +748,35 @@ export class List implements InteractiveComponent, LoadableComponent, SortableCo
...
   };
 
+  private findAncestorOfFirstFilteredItem = (filteredItems: HTMLCalciteListItemElement[]): void => {
+    this.ancestorOfFirstFilteredItem?.removeAttribute("data-filter");
+    filteredItems.forEach((item) => {
+      item.removeAttribute("data-filter");
+    });
+
+    this.ancestorOfFirstFilteredItem = this.getTopLevelAncestorItemElement(filteredItems[0]);
+    filteredItems[0].setAttribute("data-filter", "0");
```
jcfranco, line 758: Totally missed this last time, but you can use `toggleAttribute` if we don't care about the value. Not critical for this PR.

jcfranco (approved):
Awesome! 🚀

jcfranco:
This is approved and tests are looking ✅✅✅. I'm going to ahead and merge it.

@anveshmekala Can you submit a follow-up PR for any additional refactoring? Thanks!

packages/calcite-components/src/components/list/list.tsx
anveshmekala: i think we can rename the attribute to be something like `data-filtered-first` so that we can leverage `toggleAttribute`.

anveshmekala:
> This is approved and tests are looking ✅✅✅. I'm going to ahead and merge it.
> 
> @anveshmekala Can you submit a follow-up PR for any additional refactoring? Thanks!

Thank you 👍

Issue #7426, opened 2023-08-02
[List and List Item] Support sorting for AT users


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Summary

With parity of `list`, we should provide support to AT users of list when using sorting to finalize the efforts initiated in #6554.

### Actual Behavior

When active, both JAWS and NVDA are unable to sort `list-items` in the `list`.

### Expected Behavior

Same behavior as keyboard navigation without AT enabled, where `space` activates the move, and the arrow keys place the item.

### Reproduction Sample

https://codepen.io/geospatialem/pen/bGQzavG

### Reproduction Steps

1. Activate JAWS or NVDA
2. Open the [code sample](https://codepen.io/geospatialem/pen/bGQzavG)
    - JAWS: Chrome
    - NVDA: Firefox
3. Using your keyboard navigate to the first List Item's movable button
4. Press the `space` key
5. Press the down arrow key
6. Observe the breakage (_Note: NVDA takes two down arrows to break_)

### Reproduction Version

1.5.0-next.33 (1.5.0)

### Relevant Info

- List parity: https://github.com/Esri/calcite-design-system/pull/7109#pullrequestreview-1553114334
- Screen reader support: https://github.com/Esri/calcite-design-system/issues/3982#issuecomment-1206487248
- R&D efforts: https://github.com/Esri/calcite-design-system/issues/3982#issuecomment-1449143638

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

Calcite (dev)

geospatialem:
Effort should be considered with the effort in https://github.com/Esri/calcite-design-system/issues/7537

geospatialem:
This also applies to the sorting with the `block` component in `2.0.0` with both Windows AT, JAWS and NVDA, as outlined in https://codepen.io/geospatialem/pen/PoVXeVL cc @driskull

driskull:
@geospatialem what exactly is the breakage? Is it not reading something, is it not moving?

geospatialem:
> @geospatialem what exactly is the breakage? Is it not reading something, is it not moving?

It's the same for the issue above and `block`, where the context is provided to AT, but sorting/moving is not possible when either JAWS or NVDA are active.

driskull:
@geospatialem according to the following:

[...]

Issue #6399, opened 2023-01-31
Epic: move localizing numbers using numberingSystem to render block  


anveshmekala:
### Actual Behavior

While localizing numbers for a specific `numberingSystem`, it is cached by the browser and used by other components in the same page. 

### Expected Behavior

localized numbers  using `numberingSystem` should be specific for a component and should not be applied to other components in the same page.

### Reproduction Sample

https://codepen.io/ANVESHMEKALA/pen/YzjaveB?editors=100

### Reproduction Steps

1.Click on the stepper item with arabic numbers
2.Then, click on the stepper item above with latin numbers
3.Observe that latin numbers changes to arabic.

### Reproduction Version

beta.97

### Relevant Info

List of affected components: 

- [ ] input-date-picker
- [x] stepper-item  (#6331)

Please feel free to add if any other components experience similar behavior.

### Esri team

N/A

benelan:
>  While localizing numbers for a specific numberingSystem, it is cached by the browser and used by other components in the same page.
> localized numbers using numberingSystem should be specific for a component and should not be applied to other components in the same page.

Small clarification in case someone grabs this before I do - we want the components to continue using the same NumberFormatter if they have matching lang/numberingSystem (I assume this is the case in 99% of apps).

Adding it to the render block just makes sure the component is always using a NumberFormatter for its lang/numberingSystem if the app has components with different i18n.

benelan:
@anveshmekala `numberFormatOptions` is already set in `calcite-input-date-picker`'s render block:

https://github.com/Esri/calcite-design-system/blob/4b4c43c952b54a6a74d41e4257607ad98bc2efe4/packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx#L505-L511

When you created the checklist above, did you go through and check all the components' code, or was it based on some behavior you saw? Let me know if there is more investigation needed on my end. Thanks!

anveshmekala:
i remember going through each component. I would recommend more investigation  just to make sure no other components had implemented this since the creation of the epic.

benelan:
Looked into this a bit more and `numberStringFormatter.numberFormatOptions` only needs to be called in the `render` block if a number is being formatted in the render block via `numberStringFormatter.localize()`.

Some components like `calcite-input` store the localized number as State instead. This works because even if a different component changes the lang of the formatter cached by NumberStringFormat, `calcite-input` rerenders won't cause the value to be relocalized so there won't be a conflict.

[...]

PR #8148, opened 2023-11-08
fix(list): Add live region for dynamically changing list items

driskull:
**Related Issue:** #5625

## Summary

- Adds aria-live region to show currently enabled and filtered items


packages/calcite-components/src/components/list/list.tsx
```diff
@@ -459,6 +516,42 @@ export class List implements InteractiveComponent, LoadableComponent, SortableCo
...
+      filterText,
+      filteredData,
+    } = this;
+
+    numberStringFormatter.numberFormatOptions = {
+      locale: effectiveLocale,
+      numberingSystem,
+    };
+
+    const count = numberStringFormatter.localize(enabledListItems.length.toString());
```
driskull, line 536: @anveshmekala @jcfranco would this count need to be localized like this?

anveshmekala: i don't think this should be localized considering it is a dynamic property.

driskull: Curious, why does being dynamic not make it good to localize? If we're rendering a number shouldn't it be formatted & localized? Even if the number isn't displayed but available to be read to an end user.

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -459,6 +516,42 @@ export class List implements InteractiveComponent, LoadableComponent, SortableCo
...
+
+    const showFilteredExt = filterEnabled && filterText && filteredData?.length;
+
+    return !parentListEl ? (
+      <div aria-live="polite" class={CSS.assistiveText}>
+        {showFilteredExt ? <div key="aria-filter-enabled">{messages.filterEnabled}</div> : null}
+        <div key="aria-item-count">{messages.total.replace("{count}", count)}</div>
```
driskull, line 543: Maybe we should have or use a substitution utility for substituting message text? Instead of just using string replace? cc @jcfranco.

packages/calcite-components/src/components/list/list.tsx
anveshmekala: Thought you were asking regarding translations, sorry for misunderstanding.  If the question regarding the placement of localization in render block , it is suggested to localize inside the render block unless the count is being used in other places . 
The reason is the caching issue we faced earlier https://github.com/Esri/calcite-design-system/pull/6380. When the numberStringFormatter is used to localize, the formatter is caching the request per lang/numberingSystem. To avoid such cases we are localizing in render block.

More discussion here https://github.com/Esri/calcite-design-system/issues/6399#issuecomment-1411139505

cc @benelan

geospatialem (commented):
WRT a11y testing on Windows: both NVDA and JAWS pick this up well. 🙌🏻 

There's still some issues with AT and the sorting, but that is out of scope for this update and will be part of https://github.com/Esri/calcite-design-system/issues/7426 targeted 🎯 for next month's release.

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -328,12 +356,40 @@ export class List implements InteractiveComponent, LoadableComponent, SortableCo
...
 
+  @State() effectiveLocale = "";
+
+  @Watch("effectiveLocale")
+  effectiveLocaleChange(): void {
+    updateMessages(this, this.effectiveLocale);
+  }
+
+  @Watch("effectiveLocale")
+  effectiveLocaleWatcher(): void {
```
jcfranco, line 384: I think we can drop this watcher since the formatter is updated right before its used in `renderItemAriaLive`.

jcfranco (approved):
✨📝✨

PR #8151, opened 2023-11-09
fix(angular): publish from the dist directory

benelan:
**Related Issue:** #7860

## Summary

Angular has an unusual workflow where you need to `cd` into the library's `dist` directory and run `npm publish` from there. I found related Lerna issues that culminated in the ability to publish a particular package from a different directory. This is the doc I found for the config option:
https://github.com/lerna/lerna/tree/main/libs/commands/publish#publishconfigdirectory

Unfortunately, it didn't end up working, so `v1.10.1-next.3` of the angular wrapper is broken. I'm not aware of a way to test the results of `lerna publish` without actually publishing.

I ended up copying the config from the `arcgis-web-components` repo's angular wrapper once I remembered they use Lerna.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -46311,25 +46311,6 @@
         "node": ">=12"
       }
     },
-    "packages/calcite-components-angular/node_modules/@esri/calcite-components": {
-      "version": "1.10.0",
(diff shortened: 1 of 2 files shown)


jcfranco (approved):
🚀🪐

PR #8163, opened 2023-11-14
refactor(input, input-number, input-text): avoid redundant `:host`

jcfranco:
**Related Issue:** N/A

## Summary

✨🧹✨


driskull (approved):
Nice cleanup! I wonder if we could have an eslint for `:host %selector%` to prevent it?

jcfranco:
> I wonder if we could have an eslint for :host %selector% to prevent it?

Oooh, I like this very much. PR coming soon.

Issue #4970, opened 2022-07-19
Setting tabindex="-1" on calcite-button does not work


mjuniper:
### Actual Behavior

When I set tabindex="-1" on a calcite-button I would expect to not be able to give it focus by tabbing to that element.

### Expected Behavior

calcite-buttons with tab-index="-1" gets focus when tabbing.

### Reproduction Sample

https://codepen.io/mjuniper/pen/MWVpLdj?editors=1000

### Reproduction Steps

1. visit https://codepen.io/mjuniper/pen/MWVpLdj?editors=1000
2. get focus into the preview pane
3. tab sequentially

observe that both calcite-buttons get focus (one has tabindex="-1" and one does not.

### Reproduction Version

beta.83

driskull:
@jcfranco I think this is an issue with the `updateHostInteraction` utility function. Its removing tabindex set by the user.

jcfranco:
@driskull We could look into having the util check for any user-set `tabindex` before taking control, but there could be cases where we still need to override it at runtime (e.g., to prevent tabbing into children).

@mjuniper May I ask what your use case is? For context, all interactive components control the component/host element's `tabindex` to support the `disabled` state.

driskull:
> but there could be cases where we still need to override it at runtime (e.g., to prevent tabbing into children).

Yeah thats the tricky part. It would need to store the user defined value and set it back when necessary.

driskull:
> For context, all interactive components control the component/host element's tabindex to support the disabled state.

I do think the use case is reasonable though. A user would expect their defined tabindex to be honored. I think it falls into the category of not messing with host attributes/classes.

jcfranco:
> @mjuniper May I ask what your use case is? For context, all interactive components control the component/host element's tabindex to support the disabled state.

Friendly ping. 😄

---

Side note: it looks like [support for `inert` is getting there](https://caniuse.com/?search=inert), at least from a `last 2 versions` perspective. Once it's widely supported, we'll definitely refactor our interactive component pattern and we would no longer need to mess w/ the host's tab index.

If this is a showstopper, we could look at introducing an internal component solely to prevent interaction with its content. @driskull Most, if not all, of our components have a root container, right?

driskull:
@jcfranco correct. Most but not all.

geospatialem:
Blocked by Firefox, will investigate once this can be adopted across browsers.

jcfranco:
[...]

PR #8166, opened 2023-11-14
fix: allow users to control tabindex on interactive components

jcfranco:
**Related Issue:** #4970 

## Summary

Updates the interactive component pattern to avoid the need to control `tabindex` on the host to prevent tabbing into a component's subtree.

This works by introducing a new functional component (`InteractiveContainer`) that leverages `inert` (prevent tabbing into its contents) and `display: contents` (allows us to use it as a root container without affecting layout), which should be the root container under the host. 

Updating components to this pattern should be straightforward as the styling of the helper component is bundled within the `disabled` mixin that interactive components already include and `updateHostInteraction` no longer needs its 2nd parameter. For cases that were setting the 2nd parameter (`hostIsTabbable`) all they have to do to migrate depends on the value:

1. `false` (default) – no further changes necessary
2. `true` – set `tabIndex={0}` on the host
3. predicate function – set `tabIndex={myPredicateFunction() ? 0 : -1 }` on the host
4. `"managed"` – no further changes necessary as owner/parent sets `tabIndex` on the component already

Any use of `Fragment` can be replaced by `InteractiveContainer` directly.

2 and 3 might also need to determine the tab index based on `disabled`.


packages/calcite-design-tokens/support/token-transformer/styleDictionary/formatter/utils/formatExtraOutput.ts
```diff
@@ -49,7 +49,8 @@ export function formatExtraOutput(
                 const m = ensureIfArray(outputObject[`${output}.${args.platform}`]);
                 const cssProps = m.map((o) => `${o}`.replaceAll("$", "--"));
                 return Array.isArray(m)
-                  ? `${output === "light" ? `:root {${cssProps.join("")}}` : ""
+                  ? `${
```
driskull, line 52: This just a formatting change?

driskull (approved):
This looks good @jcfranco ! 🎉 

We'll just have to do some testing to make sure it won't mess up any existing styles with the added container.

packages/calcite-design-tokens/support/token-transformer/styleDictionary/formatter/utils/formatExtraOutput.ts
jcfranco: Yes, I think it got formatted when the latest from `main` got merged.

jcfranco:
> We'll just have to do some testing to make sure it won't mess up any existing styles with the added container.

Based on the description of [`display: contents`](https://developer.mozilla.org/en-US/docs/Web/CSS/display#contents), I don't foresee any issues, but worth keeping an eye out anyways. 👁️👄👁️

PR #8167, opened 2023-11-14
build(angular): chage calcite-components to a dep from peerDep/devDep

benelan:
## Summary

Changes calcite-components to a dependency of the angular wrapper, versus a peerDep and devDep. 

Stencil has it as a peerDep because it is required for the user. arcgis-web-components and us added it as a devDep so the monorepo dependency tree would catch it.

Putting it as a dep resolves both of the above, and it prevents the user from needing to install it on their end. It seems like lerna doesn't bump the peerDep when versioning local packages so it falls behind, which causes issues.

Note: we did the same thing for the react wrapper for convenience, even before we moved it to the monorepo

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21941,12 +21941,6 @@
         "node": ">= 6"
       }
     },
-    "node_modules/form-request-submit-polyfill": {
-      "version": "2.0.0",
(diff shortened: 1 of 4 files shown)


jcfranco (approved):
🚀👨‍🚀

PR #8168, opened 2023-11-14
chore: drop unnecessary double build type workaround since `components.d.ts` is now tracked

jcfranco:
**Related Issue:** #8108 

## Summary

Follow up from https://github.com/Esri/calcite-design-system/pull/8108#pullrequestreview-1723826233.


jcfranco:
@benelan @alisonailea I'm getting the following error:

```cli
@esri/calcite-components:test: Summary of all failing tests
@esri/calcite-components:test: FAIL src/utils/responsive.spec.ts
@esri/calcite-components:test:   ● Test suite failed to run
@esri/calcite-components:test: 
@esri/calcite-components:test:     Cannot find module '@esri/calcite-design-tokens/dist/es6/calcite-headless' from 'src/utils/responsive.ts'
@esri/calcite-components:test: 
@esri/calcite-components:test:     Require stack:
@esri/calcite-components:test:       src/utils/responsive.ts
@esri/calcite-components:test:       src/utils/responsive.spec.ts
@esri/calcite-components:test: 
@esri/calcite-components:test:     > 1 | import {
@esri/calcite-components:test:         | ^
@esri/calcite-components:test:       2 |   CoreBreakpointWidthDefaultLg,
@esri/calcite-components:test:       3 |   CoreBreakpointWidthDefaultMd,
@esri/calcite-components:test:       4 |   CoreBreakpointWidthDefaultSm,
@esri/calcite-components:test: 
@esri/calcite-components:test:       at Resolver._throwModNotFoundError (../../node_modules/jest-resolve/build/resolver.js:427:11)
@esri/calcite-components:test:       at Object.<anonymous> (src/utils/responsive.ts:1:1)
@esri/calcite-components:test:       at Object.<anonymous> (src/utils/responsive.spec.ts:1:1)
```

I dug into this a bit and noted the following:

1. `/dist/es6/calcite-headless` is missing from the tokens `package.json`'s `exports` mapping. 
2. The `tsconfig.json` file used by the components is using `node` as its module resolution. It needs `node16` for it to leverage the token `exports`.

We can update the module resolution for 2, but it may require a bit more testing and other configs may need to be updated as well. Adding `/dist/es6/calcite-headless` to the `exports` map seems like a simple fix for the time being. WDYT about updating the `exports` map for now (undocumented) and we can deprecate/remove once we've had some time to test the module resolution switch?

jcfranco:
The above error may be caused by something else. Our TS configs use `node` (`node10`) for module resolution, which should not be affected by `package.json` 's `exports` field (see https://www.typescriptlang.org/docs/handbook/modules/reference.html#packagejson-exports).

jcfranco:
@benelan I noticed the `rc` test runs are ✅, so I'm not sure what's going on here. I see the same error running the spec tests locally (after running a full monorepo build).

benelan:
> @benelan I noticed the `rc` test runs are ✅, so I'm not sure what's going on here. I see the same error running the spec tests locally (after running a full monorepo build).

Can you try testing locally on the `rc-force` branch I just pushed, which rebases to get rid of the copied commits. There is no diff between `rc` and `rc-force` but not sure what else could be causing issues with the design tokens. 

Maybe try deleting node_modules and package-lock.json too to see if that resolves the issue.

jcfranco:
🚫🎲😭

jcfranco:
PR recreated on updated `rc` base: https://github.com/Esri/calcite-design-system/pull/8199.

Issue #6760, opened 2023-04-10
Look into stripping out testing `data-` attributes from production code.


driskull:
**monday.com sync:** #18048707802

### Background

https://github.com/Esri/calcite-components/pull/6746#discussion_r1162056759

### Desired Outcome

Any data attributes or testing specific code is removed from the production code.

_May have some risk involved and it could impact current component functionality._ 🚧

jcfranco:
I think I have something we could use for this: https://github.com/Esri/calcite-design-system/pull/8171.

driskull:
We could convert all the `data-test` attributes to `id`. there aren't too many

<img width="372" alt="image" src="https://github.com/user-attachments/assets/01203712-23c9-48d3-a214-3514f8a2a258">

jcfranco:
Based on https://github.com/vitejs/vite/discussions/10408, we might be able to handle this at the build level. https://github.com/jacobbogers/rollup-plugin-jsx-remove-attributes looks promising, but we’ll need to wait until https://github.com/estools/escodegen/pull/465 lands and the plugin is updated to use it. Currently, it produces the following error:

```bash
x Build failed in 10.16s
error during build:
[vite-plugin-jsx-remove-attributes] this[type] is not a function
file: /.../calcite-design-system/packages/calcite-components/src/utils/number.js
at CodeGenerator.generateExpression (/.../calcite-design-system/node_modules/escodegen/escodegen.js:2517:28)
at /.../calcite-design-system/node_modules/escodegen/escodegen.js:1133:38
at withIndent (/.../calcite-design-system/node_modules/escodegen/escodegen.js:581:9)
at CodeGenerator.ClassBody (/.../calcite-design-system/node_modules/escodegen/escodegen.js:1128:13)
at CodeGenerator.generateStatement (/.../calcite-design-system/node_modules/escodegen/escodegen.js:2530:33)
at CodeGenerator.ClassDeclaration (/.../calcite-design-system/node_modules/escodegen/escodegen.js:1159:30)
at CodeGenerator.generateStatement (/.../calcite-design-system/node_modules/escodegen/escodegen.js:2530:33)
at CodeGenerator.ExportNamedDeclaration (/.../calcite-design-system/node_modules/escodegen/escodegen.js:1237:42)
at CodeGenerator.generateStatement (/.../calcite-design-system/node_modules/escodegen/escodegen.js:2530:33)
[...]

PR #8171, opened 2023-11-14
refactor: use utility to ensure test-specific data attributes are removed in prod builds

jcfranco:
**Related Issue:** #6760

## Summary

This new util should help remove test-specific data attributes in the `dist` build and in tree-shaken builds using the `components` output build.

We could also enable an ESLint rule to flag `data-test-x` attributes and recommend using this pattern instead.

# Questions

1. Should we recommend spreading the test attribute object at the end or in place of where the data attribute would reside?
2. Would it be helpful to rename the util to include data attribute or something along those lines? Open to suggestions here since I'm not keen on the current name.


jcfranco:
Closing in favor of [this proposed solution](https://github.com/Esri/calcite-design-system/issues/6760#issuecomment-2660543976).

Issue #6678, opened 2023-03-27
[Date Picker] Responsive layout


geospatialem:
### Description

Mobile responsiveness across Calcite Design System, focused on web (traditionally desktop sized web browser), tablet, phone screen sizes, and large monitors.

### Acceptance Criteria

- [x] 1. Consider breaking changes for folks. Consultation should occur on both the design and development teams.
- [x] 2. Components need to be designed for the following breakpoints:
    - [x] Small (s), `476px`
    - [x] Medium (m), `768px`
    - [x] Large (l), `1152px`
    - [ ] Extra larger (xl), `1440px`

### Relevant Info

Part of epic https://github.com/Esri/calcite-components/issues/6670

### Which Component

`input-date-picker`

### Priority impact

p4 - not time sensitive

### Esri team

Calcite (design)

manuelalcosta:
Just applied the change on the variants on the 3 scales on the default variant

SkyeSeitz:
Two recordings demonstrating the proposed behavior change:

**BEFORE**
https://user-images.githubusercontent.com/29716057/230445031-26be2f0b-e5f4-4335-b104-8d2534be116c.mov
**AFTER**
https://user-images.githubusercontent.com/29716057/230445076-0c1a36bd-8f1e-412f-ac75-ca6fbeee95d8.mov

SkyeSeitz:
Completed responsive design available [here](https://www.figma.com/file/zBGnAOEOxqKGocUjqiYbed/Responsive-Layout?node-id=8279%3A183723&t=H0cPsE1ZwYk0UsRU-1).

jcfranco:
I have a PR ready, but it is not low risk, so this would be better suited for the November release. cc @geospatialem

geospatialem:
Verified in `1.11.0-next.3`

PR #8172, opened 2023-11-14
fix(input-date-picker): fix date-picker wrapper displaying beyond its bounds

jcfranco:
**Related Issue:** #6678

## Summary

This removes unnecessary sizing applied to the date-picker's wrapper which displayed an artifact when configuring the `input-date-picker` with `scale=l` and `range`.

MicrosoftTeams-image (2)

### DIFF
--- a/packages/calcite-components/src/components/input-date-picker/input-date-picker.scss
+++ b/packages/calcite-components/src/components/input-date-picker/input-date-picker.scss
@@ -24,9 +24,7 @@
 @include disabled();
 
 .calendar-picker-wrapper {
-  @apply static
-    w-full
-    shadow-none;
+  @apply shadow-none static;
   transform: translate3d(0, 0, 0);
 }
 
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/input-date-picker/input-date-picker.scss
```diff
@@ -24,9 +24,7 @@
 @include disabled();
 
 .calendar-picker-wrapper {
-  @apply static
-    w-full
-    shadow-none;
+  @apply shadow-none static;
```
driskull, line 27: is static needed? its the default position.

driskull (commented):
👏 Nice! 👏

packages/calcite-components/src/components/input-date-picker/input-date-picker.scss
jcfranco: Nope. Good catch. 🎣 Will remove!

PR #8175, opened 2023-11-15
ci: fix versioning script async issue that committed too early

benelan:
## Summary

Resolve an async issue that caused the versioning script to commit before it finished writing changes to the angular package's changelog and README. This issue arose in #8167 so it never actually caused any issues in a real release.


jcfranco (approved):
🚀

PR #8123, opened 2023-11-03
feat(list): Specify the element types in the `calciteListOrderChange` event detail.

driskull:
**Related Issue:** #8111

## Summary

- More specifically types the elements within the drag event

PR #8177, opened 2023-11-15
build(angular): remove event detail types patch by exporting them from the entry point

benelan:
**Related Issue:** #8084

## Summary

I ran into build errors related to event detail types when adding the Angular output target to the monorepo. Stencil's auto-generated Angular code expected the types to be exported from the `@esri/calcite-components` entry point, but they were not. Here is one of the error messages:

```
angular-workspace:build: projects/component-library/src/lib/stencil-generated/components.ts:1014:15 - error TS2459: Module '"@esri/calcite-components"' declares 'HandleNudge' locally, but it is not exported.
angular-workspace:build:
angular-workspace:build: 1014 import type { HandleNudge as ICalciteHandleHandleNudge } from '@esri/calcite-components';
angular-workspace:build:                    ~~~~~~~~~~~
angular-workspace:build:
angular-workspace:build:   ../calcite-components/dist/types/components.d.ts:42:24
angular-workspace:build:     42 import { HandleChange, HandleNudge } from "./components/handle/interfaces";
angular-workspace:build:                               ~~~~~~~~~~~
angular-workspace:build:     'HandleNudge' is declared here.
```

To resolve the errors, I added a prebuild patch to the Angular output target that fixed the import paths for the three event detail types. As a result, a Calcite Components fix/feat that added a new event detail would create an unrelated entry in the Angular changelog. This is because the new event detail would need to be added to the patch, which lived in the Angular package's directory. This came up in: https://github.com/Esri/calcite-design-system/pull/8123#issuecomment-1804500831

This pull request exports the event detail types from the entry point, so the patch can be removed. **When adding a new event detail, it must be exported from [`packages/calcite-components/src/index.ts`](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/index.ts) to prevent build errors.**


jcfranco (approved):
🎉

jcfranco:
I think Stencil 4 includes custom event types in `components.d.ts` by default, so this may become even simpler soon. 🎉

PR #8178, opened 2023-11-15
ci(angular): rebuild before publishing to fix outdated version in dist

benelan:
## Summary

Add a script to re-build the Angular output target before publishing to fix an outdated version issue.

Angular generates a `package.json` file in the dist directory and expects users to `cd` into the dist and publish from there. A Lerna config option for publishing from a different directory was added due to this Angular workflow.

Unfortunately, the version in Angular's auto-generated `package.json` is outdated if you build before versioning with Lerna, which we do to ensure tests pass. Building again will ensure the version is correct and the Angular output target gets published.


jcfranco (approved):
🚀

PR #8180, opened 2023-11-15
ci: remove old, unused actions

benelan:
## Summary

Removes some of the actions I added back when I was the only product person. They are now done (much better) by Brittney/Kitty.

- Milestone add/move/close - no longer used (taken over by Brittney/Kitty)
- Release Teams notification - still occurs at the moment, but is redundant (Brittney's is much more detailed)
- Need info (require sample) - we now use GitHub's form issue templates, where we can mark fields as required

Whoever said "the robots are coming for your job" got it all wrong. Kitty/Brittney came for my robot's jobs :sob:


jcfranco (approved):
🤖⁉️

Issue #8173, opened 2023-11-14
List Item - Add a new slot to display content at the bottom of the content in a list item


sagewall:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

We are working on refactoring the JavaScript Maps SDK LayerList widget to use the Calcite Design System.  It turns out there isn't a good slot to put the legend in the list item. When adding the legend to the `content` slot the list item stretches the action buttons to match the height of the content and centers them vertically.

![MicrosoftTeams-image (6)](https://github.com/Esri/calcite-design-system/assets/2636825/d232af21-0521-40be-93b1-098cf13b6eff)

Using the `default` slot would result in the legend not being show unless the list item is open as well as producing alignment issues with nested children.

A new content slot (content-bottom) would be a great addition to list items to display content under the label description and actions.

### Acceptance Criteria

The ability to show a legend widget without alignment issues in a list item, similar to the way legends are currently shown in layer list items. https://developers.arcgis.com/javascript/latest/sample-code/widgets-layerlist-legend/

### Which Component

List Item

### Example Use Case

Refactoring the ArcGIS Maps SDK for JavaScript LayerList widget to use the Calcite Design System

### Priority impact

p2 - want for current milestone

### Calcite package

- [X] @esri/calcite-components
- [X] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

sagewall:
cc @driskull @asangma

driskull:
@geospatialem can we triage this one for November?

macandcheese:
Is it possible to use a Block in any way for this?

driskull:
I don't think so.

macandcheese:
Is there an example of the desired UI? It's tough to tell from the screenshot - are there siblings? 

In the [linked example](https://developers.arcgis.com/javascript/latest/sample-code/widgets-layerlist-legend/), there's just a single item in that content area. For that case, a Block or even a Panel could make sense...

driskull:
Here's another example: https://developers.arcgis.com/javascript/latest/sample-code/layers-imagerytilelayer/

The issue with block is we can't just insert it in the middle of a layerlist. We need a way to display content within a single list item that can be toggled by a user. This content should also be indented with the list item and not depend on the list item being open.

[...]

PR #8183, opened 2023-11-16
feat(list-item): Add content-bottom slot for placing content below the label and description of the component

driskull:
**Related Issue:** #8173

## Summary

- This is a good starting point. We just need input from design. cc @macandcheese 
- Adds a new slot `content-bottom`. Could also be called `content-block-end`?
- `content-bottom` is indented the same as the item.
- screenshot tests

### DIFF
--- a/packages/calcite-components/src/components/list-item/list-item.tsx
+++ b/packages/calcite-components/src/components/list-item/list-item.tsx
@@ -51,7 +51,7 @@ import {
  * @slot content - A slot for adding non-actionable, centered content in place of the `label` and `description` of the component.
  * @slot content-end - A slot for adding non-actionable elements after the label and description of the component.
  * @slot actions-end - A slot for adding actionable `calcite-action` elements after the content of the component.
- * @slot content-bottom - A slot for adding content below the label and description of the component.
+ * @slot content-bottom - A slot for adding content below the component's `label` and `description`.
  */
 @Component({
   tag: "calcite-list-item",
--- a/packages/calcite-components/src/components/list-item/list-item.tsx
+++ b/packages/calcite-components/src/components/list-item/list-item.tsx
@@ -276,6 +276,8 @@ export class ListItem
 
   @State() hasContentEnd = false;
 
+  @State() hasContentBottom = false;
+
   containerEl: HTMLTableRowElement;
 
   contentEl: HTMLTableCellElement;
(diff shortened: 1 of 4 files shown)


macandcheese:
Nice! No issue with the visuals but cc @SkyeSeitz @ashetland for visibility. I think we need to do a larger pass on the "inline/bottom/block/top" stuff, but for now maybe "bottom" is clearer since we already have "content/actions-end" and those don't use "inline".

ashetland:
> Nice! No issue with the visuals but cc @SkyeSeitz @ashetland for visibility. I think we need to do a larger pass on the "inline/bottom/block/top" stuff, but for now maybe "bottom" is clearer since we already have "content/actions-end" and those don't use "inline".

I vote for `content-bottom`. Follows Modal which has `content`, `-top`, `-bottom`.

> I think we need to do a larger pass on the "inline/bottom/block/top" stuff

Agree.

ashetland:
I think I know the answer to this, but it ties into other things I'm working on. Should the bottom slot have some default padding on it or do we leave it up to the user? At first I was thinking about x-padding to help align it with the heading+description, but then of course that leads to asking about y-padding too.

SkyeSeitz:
Agreed on `content-bottom`. Just did a quick pass and I believe Card's thumbnailPosition is now the only prop still using block-end/start language, which we could consider changing to follow the top/bottom paradigm moving forward.

macandcheese:
I do think inline/block-start/end (as in Card) is a more accurate verbiage that aligns with css terminology and is writing mode agnostic. 

But in our current nomenclature here it wouldn’t fit, so bottom IMO is good for now.

Places like Shell, other components that use top/bottom/start/end, maybe can all use inline/block-start/end in the future.

driskull:
@jcfranco @macandcheese can I get a review of this one?

macandcheese (approved):
Awesome! Can you add this slot to the jsdoc definitions at the top of `list-item.tsx` so it comes through in docs? Looks good otherwise, we can work to add "good use cases" to the doc site.

driskull:
Nice catch. Added slot doc. @geospatialem can you review it?

jcfranco:
> I do think inline/block-start/end (as in Card) is a more accurate verbiage that aligns with css terminology and is writing mode agnostic.

I disagree here. The `inline/block`-related CSS logical prop terminology is very much `writing-mode` dependent. All logical props have something similar to this in their description:

> It corresponds to the top, right, bottom, or left property depending on the values defined for writing-mode, direction, and text-orientation.

 ☝️from https://developer.mozilla.org/en-US/docs/Web/CSS/inset-inline-end

jcfranco:
Scratch what I said above. I read that wrong. 😅

nothing-to-see

jcfranco:
> I think we need to do a larger pass on the "inline/bottom/block/top" stuff, but for now maybe "bottom" is clearer since we already have "content/actions-end" and those don't use "inline".

> I do think inline/block-start/end (as in Card) is a more accurate verbiage that aligns with css terminology and is writing mode agnostic.

@macandcheese Do we need an audit for this for the future? Seems like we should for consistency.

packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -51,6 +51,7 @@ import {
  * @slot content - A slot for adding non-actionable, centered content in place of the `label` and `description` of the component.
  * @slot content-end - A slot for adding non-actionable elements after the label and description of the component.
  * @slot actions-end - A slot for adding actionable `calcite-action` elements after the content of the component.
+ * @slot content-bottom - A slot for adding content below the label and description of the component.
```
geospatialem, line 54: Awesome! Small recommended tweak and consistency with backticks in to differentiate the props:

```
A slot for adding content below the component's `label` and `description`.
```

Issue #5860, opened 2022-11-30
Enhance indicator positioning with a wide `calcite-action`


ethanbdev:
### Description

The positioning of the indicator looks a bit weird when the action is wide:
![image](https://user-images.githubusercontent.com/52869490/204918346-dc0dfc79-569b-4603-93bb-fdf273d48924.png)

It would be nice if no matter what the width of the action, the indicator would show in the bottom right of the icon.

### Acceptance Criteria

The indicator is in the bottom right of the icon on a calcite-action without text enabled, regardless of action width.

### Relevant Info

From @macandcheese:

> Right now it’s just absolutely positioned to the action component entire container, when text-enabled is not true maybe it could 
> be absolutely positioned to the icon container

### Which Component

calcite-action

### Example Use Case

https://codepen.io/eborgen/pen/RwJYExY?editors=100

### Esri team

ArcGIS Map Viewer

geospatialem:
Added an issue creation priority label of `p3 - want for upcoming milestone`, which weren't available at the time of issue creation.

macandcheese:
This will be fixed in https://github.com/Esri/calcite-design-system/issues/6879 and is expected to land for the next release.

Issue #6879, opened 2023-04-27
Action: text-only actions render off-centered


paulcpederson:
### Actual Behavior

There is an extra margin to the right of a calcite action which prevents it from being properly centered in cases where the action is displayed inline with an auto width.

### Expected Behavior

Shoudl be able to render an action which is just text where the text is centered in the bounding box of the action. 

### Reproduction Sample

https://codepen.io/paulcp/pen/zYmzEdQ

### Reproduction Steps

1. Notice the extra space to the right of the word "testing"

### Reproduction Version

v1.3.1

### Relevant Info

All OS, all browsers.

Caused by https://github.com/Esri/calcite-components/pull/6562 , broken in 1.1.0

### Regression?

v1.0.8

### Priority impact

p4 - not time sensitive

### Impact

We'd like to display a text-only action. We already have these in some interfaces so it will be broken until the upstream calcite issue is fixed. 

### Esri team

ArcGIS Online

macandcheese:
+1 - this is pretty noticeable in some interfaces. cc @SkyeSeitz @ashetland

ashetland:
I think ideally the fix in pr #6562 would only apply to Actions inside an Action Bar. Another possible way to address this might be to have Action Bar account for the space needed by the indicator instead of the Actions themselves. Are vertical Action Bars the only place the indicator space matters? I'd expect horizontal Action Bars could be a little more fluid, like the behavior in Chip Group.

ashetland:
[Figma file](https://www.figma.com/file/Q16YuSQQTJKCpeT4CN0XCA/Action%3A-text-only-actions-render-off-centered-%5B6879%5D?type=design&node-id=530-170&mode=dev) for this issue. Restores Action to previous behavior and requires Action Bar accommodate the needed space for the indicator.

SkyeSeitz:
> [Figma file](https://www.figma.com/file/Q16YuSQQTJKCpeT4CN0XCA/Action%3A-text-only-actions-render-off-centered-%5B6879%5D?type=design&node-id=530-170&mode=dev) for this issue. Restores Action to previous behavior and requires Action Bar accommodate the needed space for the indicator.

Accounting for the indicator space at the Action Bar level makes a lot of sense 👍🏽

driskull:
Can we just position the indicator absolutely like we do when there is an icon? This should be a simple fix.

I'm not sure how we're expecting the action bar to account for extra space. The width of the action bar is based off of the action content inside.

[...]

PR #8184, opened 2023-11-16
fix(action): Ensure action content is correctly spaced

macandcheese:
**Related Issue:** #6879 

## Summary
Ensures extra space is not added to an action. 

This is solved by:
When an icon is present, the indicator will always be appended to the bottom right. This ensures it remains in the same place whether a parent Action Bar is expanded, or not. The indicator moved a couple of px away - in the future I think we can improve this with a border or clipped ring around the indicator to prevent overlap. 

Also resolves https://github.com/Esri/calcite-design-system/issues/5860, for icon-only uses.

When an icon is not present, the indicator will be positioned adjacent to the text container.

| Before  | After |
| ------------- | ------------- |
| Screenshot 2023-11-15 at 5 00 35 PM  | Screenshot 2023-11-15 at 4 58 47 PM  |

| Before  | After |
| ------------- | ------------- |
| Screenshot 2023-11-15 at 5 00 25 PM  | Screenshot 2023-11-15 at 4 59 26 PM  |

(Note, in this one, the "beaker" image is NOT an `icon` property, it is just a slotted element - so the indicator is not positioned over it. This isn't really a documented / expected use case.)

| Before  | After |
| ------------- | ------------- |
| Screenshot 2023-11-15 at 5 00 47 PM  | Screenshot 2023-11-15 at 5 01 02 PM |

| Before  | Screenshot 2023-11-15 at 6 17 38 PM |
| ------------- | ------------- |
| After  | Screenshot 2023-11-15 at 6 17 57 PM |

| Before  | Screenshot 2023-11-15 at 6 21 22 PM |
| ------------- | ------------- |
| After  | Screenshot 2023-11-15 at 6 18 30 PM |

### DIFF
--- a/packages/calcite-components/src/components/action/action.scss
+++ b/packages/calcite-components/src/components/action/action.scss
@@ -12,6 +12,20 @@
   --calcite-action-indicator-color: theme("colors.brand");
 }
 
+@mixin action-indicator() {
+  position: relative;
+  &::after {
+    content: "";
+    @apply absolute
+      h-2
+      w-2
+      rounded-full;
+    inset-block-end: calc(-0.275rem);
+    inset-inline-end: calc(-0.275rem);
+    background-color: var(--calcite-action-indicator-color);
+  }
+}
+
 @include disabled();
 
 .button {
@@ -70,7 +84,6 @@
 
   .text-container--visible {
     @apply w-auto flex-auto opacity-100;
-    margin-inline-end: theme("spacing.4");
   }
 }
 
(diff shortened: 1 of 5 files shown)


macandcheese:
cc @ashetland @SkyeSeitz as there are visual changes here.

driskull (approved):
Amazing ✨ 🪄

jcfranco:
Awesome! 🤩

> Should this get breaking tag (good timing...), or should we just lump the indicator placement change into the fix?

I don't think this qualifies as a breaking change, but we should definitely call this out for the next release since it is noticeable in certain configurations. cc @brittneytewks @geospatialem

PR #8185, opened 2023-11-16
fix(list-item, stack): stretch action-menu and handle when placed inside a list-item or stack.

driskull:
**Related Issue:** #8187

## Summary

- Adds css to stack and list-item to treat an action-menu the same as an action
- This makes the action take up the full height of the list-item/stack
- Previously, the action-menu would stop at 32x32 pixels.

### DIFF
--- a/packages/calcite-components/src/components/list-item/list-item.scss
+++ b/packages/calcite-components/src/components/list-item/list-item.scss
@@ -136,7 +136,8 @@ td:focus {
 
 .actions-start,
 .actions-end {
-  ::slotted(calcite-action) {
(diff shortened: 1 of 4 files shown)


macandcheese:
Any hesitation with doing this for the drag handle element as well? Right now it ends up being shorter than a List Item quite often: 
Screenshot 2023-11-15 at 5 52 07 PM

driskull:
Yeah that makes sense @macandcheese

jcfranco (approved):
⬅️⬜⬜⬜⬜⬜⬜➡️

Issue #3834, opened 2022-01-01
Action appearance="clear" hover state should follow general pattern


asangma:
### Actual Behavior
When Action has `appearance="clear"`, its hover state is kind of a border.

### Expected Behavior
It should probably be a solid background color and follow the same pattern that Button uses and utilize `--calcite-button-transparent-hover`.

Yes, that's what Button uses for `appearance="transprent"`, but that's a different discussion.

### Reproduction Steps and Sample

_Sample_: https://codepen.io/mtb-alan/pen/abLYBNr

1. Hover over the Action.

### Relevant Info

_Reproducible version_: `@esri/calcite-components@1.0.0-<version>`

- [ ] CDN
- [ ] NPM package

Regression? _Last working version_: `@esri/calcite-components@1.0.0-<version>`

macandcheese:
@asangma are any other changes to existing components needed to make this consistent across the library?

macandcheese:
@asangma - I'm going to move this to an enhancement request. There are other components that use the `clear` prop name so we should audit these consistently if we decide to make a change.

macandcheese:
@ashetland @SkyeSeitz can we take a look at this now that the nomenclature has changed and see what's needed here?

ashetland:
Agree. The `appearance`=`transparent` should follow the same hover pattern as Button. Is this another audit situation of hover states on transparent appearance?

ashetland:
[Figma file](https://www.figma.com/file/9w88XL7TvWob68kHIcFWZx/Action-transparent-hover-%5B3834%5D?type=design&node-id=401-8255&mode=dev) for reference.

geospatialem:
Verified in the `main` branch. 
![verify-transparency-action-hover-state](https://github.com/Esri/calcite-design-system/assets/5023024/2242095c-6a9d-4ca4-89e3-b0680696e828)

Issue #7538, opened 2023-08-16
List item: slotted actions obstruct portion of list outline with active


COV-GIS:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

1.4.3 and higher:

![image](https://github.com/Esri/calcite-design-system/assets/14297903/4432a429-96aa-4d2a-bd3a-c69bc9e274ba)

### Expected Behavior

1.4.2:

![image](https://github.com/Esri/calcite-design-system/assets/14297903/1a50f08a-ee75-4331-a21b-dbdb72bb2005)

### Reproduction Sample

https://codepen.io/geospatialem/pen/VwOXGgr
https://codepen.io/COV-GIS/pen/wvRwmPV?editors=100

### Reproduction Steps

Switch between 1.4.2 and 1.4.3 in sample.

### Reproduction Version

1.6.0

### Regression?

1.4.2

### Priority impact

impact - p3 - not time sensitive

### Impact

Does not affect usability or workflows. Just aesthetics.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

N/A

jcfranco:
As a workaround, action's `appearance` prop can be set to `transparent` (via @driskull 🏆). cc @macandcheese

geospatialem:
Additional information needs to determine a potential fix that doesn't impact the original fix creating the regression.

geospatialem:
A spike will be conducted before the end of 2023 to determine:
1. A solution between the current bug and the previous border issue
    - Issue: https://github.com/Esri/calcite-design-system/issues/6373
    - Commit: https://github.com/Esri/calcite-design-system/commit/a810943fdea2c1f90f5deca35ab0501287e45489
2. Should we consider moving the DOM in `list` with `filter`? (this might be risky change, should we pursue it?)
3. Are there other alternatives we should consider?

cc @macandcheese

geospatialem:
Spike effort to determine next steps - refer to https://github.com/Esri/calcite-design-system/issues/7538#issuecomment-1785350627 for additional context cc @macandcheese @brittneytewks

josercarcamo:
Met with @driskull. There is probably nothing that can be done because the slotted action has a higher z-index than the td. @jcfranco may be we can discuss further.

driskull:
[...]

PR #8194, opened 2023-11-17
fix(action): Update transparent action styles

macandcheese:
**Related Issue:** #3834 

## Summary
cc @asangma @SkyeSeitz @ashetland


SkyeSeitz:
Looking good! A few small notes before approving the rest of the snapshots..
* The active fg-3 background [goes away when loading=“true”](https://github.com/Esri/calcite-design-system/assets/29716057/d7301312-1f8e-4398-9cc1-151a5621b495)
* The loader not animating is a Storybook artifact, correct?
* Should the hover leave 1px top & bottom or can it fill the height of the container? 
hello world

* Same here?
Estuaries

* Super smol but there is corner radius mismatch here on all [dismissible popover](https://6266d45509d7eb004aa254fb-ufvywejadj.chromatic.com/?path=/story/components-popover--large-scale-layout-test-only ) scales (I think this is just for the test tho?)

Pasted Graphic 2

macandcheese:
> Super smol but there is corner radius mismatch here on all [dismissible popover](https://6266d45509d7eb004aa254fb-ufvywejadj.chromatic.com/?path=/story/components-popover--large-scale-layout-test-only) scales

Nice catch! Should be unrelated to this PR though - could you open a new issue?

> Should the hover leave 1px top & bottom or can it fill the height of the container?
This is the current behavior, it might be resolved by https://github.com/Esri/calcite-design-system/issues/7538, could you add a +1 there?

For the "Estuaries" one that is not close to full-height, only Action elements are slotted and automatically stretched - that is slotted into the Dropdown trigger so isn't being stretched. I think this could be improved, but also something a user can adjust. Upcoming changes to Dropdown might resolve this by using a `reference-element` instead of user needing to slot a trigger inside Dropdown.

I'll take a look at the FG change in loading, can you confirm the behavior for this transparent mode and solid? Thanks!

SkyeSeitz:
> I'll take a look at the FG change in loading, can you confirm the behavior for this transparent mode and solid? Thanks!

I would expect the transparent active Action to retain its fg-3 background when loading="true" as it does in solid. 

> This is the current behavior, it might be resolved by https://github.com/Esri/calcite-design-system/issues/7538, could you add a +1 there?

Added a +1 to https://github.com/Esri/calcite-design-system/issues/7538

> Nice catch! Should be unrelated to this PR though - could you open a new issue?

Good point - thought that was new to this build. Issue created ✅ https://github.com/Esri/calcite-design-system/issues/8208

jcfranco (approved):
Code changes LGTM! 🚀

PR #8196, opened 2023-11-17
build: swap ts-node for tsx

jcfranco:
**Related Issue:** N/A

## Summary

🧰 ➡️ 🧰

https://www.npmjs.com/package/ts-node
https://github.com/privatenumber/ts-runtime-comparison

No type-checking shouldn't be an issue per [`tsx`'s README](https://github.com/privatenumber/tsx#does-it-type-check-the-code-it-runs):

> ### Does it type check the code it runs?
> No. tsx is designed to be a simple TypeScript runner.
> 
> If you need type-checking, you can use an IDE like [VS Code](https://code.visualstudio.com/) and it will type-check as you code via [IntelliSense](https://code.visualstudio.com/docs/languages/typescript). Alternatively, you can run the TypeScript Compiler only for type-checking (e.g. tsc --noEmit) as a linting step.


benelan (approved):
@jcfranco told me to move forward with this PR via Teams. #8372 Needs to be installed first, since I had to skip those rules in this PR too due to `lint-staged` errors.

driskull (approved):
🎉 🌮

PR #8197, opened 2023-11-17
chore(design-tokens): delete unnecessary lock file

jcfranco:
**Related Issue:** N/A

## Summary

The design-tokens package no longer has package-level dependencies, so we can drop the lock file.


benelan (approved):
:broom:

PR #8204, opened 2023-11-17
fix(list-item, stack): stretch dropdown when placed inside a list-item or stack

driskull:
**Related Issue:** #8185

## Summary

- stretch dropdown component in list-item and stack


jcfranco (approved):
🚀👨‍🚀🛰️

PR #8207, opened 2023-11-17
fix(list-item): fix rendering of open icon.

driskull:
**Related Issue:** N/A

## Summary

@jcfranco I'm running into an issue where once I drag a list-item that has children, the caret icon disappears.

Nov-17-2023 14-02-24

Screenshot 2023-11-17 at 2 03 41 PM

I can't reproduce it with just calcite. Do you think this might fix it?

### DIFF
--- a/packages/calcite-components/src/components/list-item/list-item.tsx
+++ b/packages/calcite-components/src/components/list-item/list-item.tsx
@@ -392,16 +392,23 @@ export class ListItem
     const { el, open, openable, parentListEl } = this;
     const dir = getElementDir(el);
 
+    const icon = openable
+      ? open
+        ? ICONS.open
+        : dir === "rtl"
+        ? ICONS.closedRTL
+        : ICONS.closedLTR
+      : ICONS.blank;
+
+    const iconNode = <calcite-icon icon={icon} key={icon} scale="s" />;
+
     return openable ? (
-      <td class={CSS.openContainer} key="open-container" onClick={this.toggleOpen}>
-        <calcite-icon
-          icon={open ? ICONS.open : dir === "rtl" ? ICONS.closedRTL : ICONS.closedLTR}
-          scale="s"
-        />
+      <td class={CSS.openContainer} key="open-container-openable" onClick={this.toggleOpen}>
+        {iconNode}
       </td>
     ) : parentListEl?.openable ? (
-      <td class={CSS.openContainer} key="open-container" onClick={this.itemClicked}>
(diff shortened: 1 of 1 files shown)


driskull:
Seems to be an issue where the path isn't being updated or fetched properly.

jcfranco (approved):
🐛💥🔨

Could you create a follow-up issue to dig into why the icon isn't loading properly in this case a bit closer?

packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -392,16 +392,23 @@ export class ListItem
...
+
+    const iconNode = <calcite-icon icon={icon} key={icon} scale="s" />;
+
     return openable ? (
-      <td class={CSS.openContainer} key="open-container" onClick={this.toggleOpen}>
-        <calcite-icon
-          icon={open ? ICONS.open : dir === "rtl" ? ICONS.closedRTL : ICONS.closedLTR}
-          scale="s"
-        />
+      <td class={CSS.openContainer} key="open-container-openable" onClick={this.toggleOpen}>
```
jcfranco, line 406: Do you need unique keys on the parent for the fix to work? I would think the icon's key would suffice as it would ensure it gets recreated.

Could you also add a note about this workaround?

driskull:
Created https://github.com/Esri/calcite-design-system/issues/8228

packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -392,16 +392,23 @@ export class ListItem
...
+        {iconNode}
       </td>
     ) : parentListEl?.openable ? (
-      <td class={CSS.openContainer} key="open-container" onClick={this.itemClicked}>
-        <calcite-icon icon={ICONS.blank} scale="s" />
+      <td class={CSS.openContainer} key="open-container-blank" onClick={this.itemClicked}>
```
jcfranco, line 410: Sidebar: if the open container is the same element, could we dynamically assign the event handler instead of having 2 seemingly different `VNodes`?

packages/calcite-components/src/components/list-item/list-item.tsx
driskull: Nope, removed.

packages/calcite-components/src/components/list-item/list-item.tsx
driskull: Agreed. Made changes.

PR #8222, opened 2023-11-20
feat(stepper-item)!: remove support for previousStep and nextStep in messages

anveshmekala:
**Related Issue:** #8223 

## Summary
BREAKING CHANGE: Removed `previousStep` and `nextStep`  in messages. They are not available for overriding via `messageOverrides`


jcfranco (approved):
🪨💥🔨

Issue #7188, opened 2023-06-16
`calcite-input-number` - input keeps increasing/decreasing if the mouseup happens outside the increment/decrement button


ethanbdev:
### Actual Behavior

If a user clicks the increment/decrement button on the number input, the moves the mouse button off the increment/decrement button, the input value continues to increase/decrease. The user has to re-click to make it stop

Observe in the video the increment button loses focus but the value keeps increasing until I click again

### Expected Behavior

I think it should stop increment/decrement as soon as the mouse out happens and the button loses focus

### Reproduction Sample

https://codepen.io/eborgen/pen/ExOKNLP?editors=1010

### Reproduction Steps

![issue-long-mouse-press](https://github.com/Esri/calcite-design-system/assets/5023024/67a9d7dc-8450-42f4-8ad3-ca4e2ede4566)

See gif above

### Reproduction Version

1.4.2

### Relevant Info

Reproducible in latest chrome/ff - does not seem browser specific

Not reproducible using `calcite-input` type `number`

### Priority impact

p3 - want for upcoming milestone

### Impact

I think its an odd user behavior but  changing the value without user action / no easy way to stop it is confusing

### Esri team

ArcGIS Web Analysis

ethanbdev:
Another test case I just found - if you use the keyboard up/down keys, then tab to lose the focus on the input, the same behavior happens.

anveshmekala:
Hi @ethanbdev , i am unable to repro the issue in the Codepen link  using Chrome 118.0 version and FF 118.0.2 . Additionally tested with latest calcite version 1.9.2 and couldn't repro either. Is this still reproducible?

ethanbdev:
> Hi @ethanbdev , i am unable to repro the issue in the Codepen link using Chrome 118.0 version and FF 118.0.2 . Additionally tested with latest calcite version 1.9.2 and couldn't repro either. Is this still reproducible?

Let me check with the team

ethanbdev:
> Hi @ethanbdev , i am unable to repro the issue in the Codepen link using Chrome 118.0 version and FF 118.0.2 . Additionally tested with latest calcite version 1.9.2 and couldn't repro either. Is this still reproducible?

It still reproduces for me, same chrome version and calcite version. I update the codepen to the latest.

anveshmekala:
Thanks for the confirmation @ethanbdev , I am able to repro this by clicking on the increment/decrement button for longer period of time before moving it away.

[...]

PR #8226, opened 2023-11-21
fix(input-number): prevents mutating value on blur

anveshmekala:
**Related Issue:** #7188 

## Summary

Prevents increasing/decreasing the value in `calcite-input-number` on `blur` via keyboard or mouse


jcfranco (approved):
Noice! Do we also need to update `input` with similar logic?

anveshmekala:
> Noice! Do we also need to update `input` with similar logic?

Yup, we need to make this change to `input` also.

Issue #8214, opened 2023-11-18
Closed floating elements are taking up visual space when being dragged


driskull:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Floating elements are taking up visual space because they are set to `visibility:hidden`.

Is there any way we can make them not introduce scrollbars in certain scenarios?

### Expected Behavior

I would expect that any hidden menus are not taking up any visual space

### Reproduction Sample

https://codepen.io/driskull/pen/Rwvxyaj

### Reproduction Steps

1. Open sample
2. Notice the scrollbar appearing in the panel
3. The scrollbar is because of the dropdown which is not opened.
4. The banana icon opens a dropdown which is causing the issue

### Reproduction Version

1.10

### Priority impact

p2 - want for current milestone

### Impact

This is noticeable when a large menu is placed inside a list item and makes dragging experience pretty bad.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

geospatialem:
Verified in `1.12.0-next.6`

Issue #7979, opened 2023-10-11
calcite-block / calcite-sortable-list: Avatars fly in from left on Mac


AdelheidF:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Using Chrome on Mac when reordering calcite-blocks avatars fly in from the left.
It's fine on Windows.

Video that shows the issue here:
https://github.com/Esri/calcite-design-system/assets/44481292/6c37d88a-fc0a-4188-a636-663a37fb57e2

Matt discovered that it has something to do with the calcite-dropdown on the block headers.

cc @driskull 

### Expected Behavior

Same behavior on Mac and Windows.

### Reproduction Sample

https://codepen.io/afreitag/pen/eYKweaz

### Reproduction Steps

Reorder blocks on right panel.

### Reproduction Version

1.9.1

### Relevant Info

Mac, Chrome

### Priority impact

p2 - want for current milestone

### Impact

medium

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Map Viewer

driskull:
This is because the `calcite-dropdown`'s `overlayPositioning`  is `fixed` and the dropdown menu is originally positioned in the top left corner because of this. When you drag the element, the whole element (including the menu which is in the top left corner) is being dragged. 

![Screenshot 2023-10-12 at 2 29 48 PM](https://github.com/Esri/calcite-design-system/assets/1231455/e7fe9d87-51f5-4fbc-9724-edd70678dfbb)

![Screenshot 2023-10-12 at 2 29 57 PM](https://github.com/Esri/calcite-design-system/assets/1231455/b5da2163-ef15-42d4-b7fd-1deb8481592c)

If you open the dropdown and then drag everything works fine because the component is now positioned correctly.

1. One solution is to remove the `open` check here so that the component is positioned correctly even when not open yet:
https://github.com/Esri/calcite-design-system/blob/8ea00eccee8ee6c40b0a1ae9992beb575247532f/packages/calcite-components/src/utils/floating-ui.ts#L424-L426
[...]

PR #8230, opened 2023-11-21
fix: dragging floating ui components

driskull:
**Related Issue:** #8214 #7979

## Summary

- Update some component `open` handlers to call reposition before open and after open
  - Previously, they were just calling when open was set to true.
- Update `floating-ui` utility
  - Handle device pixel ratio: https://floating-ui.com/docs/misc#subpixel-and-accelerated-positioning 
  - Only set `top`, `left`, and `transform` when the component is open.
    - This makes it so that the positioned container is never going to "fly in" from anywhere. 
- Update sortable styles to hide any overflow on dragged elements.

### DIFF
--- a/packages/calcite-components/src/utils/floating-ui.ts
+++ b/packages/calcite-components/src/utils/floating-ui.ts
@@ -157,8 +157,8 @@ export const positionFloatingUI =
       pointerEvents,
       position,
       transform: open ? `translate(${roundByDPR(x)}px,${roundByDPR(y)}px)` : "",
-      left: open ? `0` : "",
-      top: open ? `0` : "",
+      left: open ? "0" : "",
+      top: open ? "0" : "",
     });
   };
 
--- a/packages/calcite-components/src/utils/floating-ui.ts
+++ b/packages/calcite-components/src/utils/floating-ui.ts
@@ -27,6 +27,11 @@ import { offsetParent } from "composed-offset-position";
   }
 })();
 
+function roundByDPR(value) {
+  const dpr = window.devicePixelRatio || 1;
+  return Math.round(value * dpr) / dpr;
+}
+
 /**
  * Positions the floating element relative to the reference element.
  *
@@ -145,15 +150,15 @@ export const positionFloatingUI =
 
     floatingEl.setAttribute(placementDataAttribute, effectivePlacement);
 
-    const transform = `translate(${Math.round(x)}px,${Math.round(y)}px)`;
+    const { open } = component;
 
     Object.assign(floatingEl.style, {
       visibility,
       pointerEvents,
       position,
-      top: "0",
-      left: "0",
-      transform,
+      transform: open ? `translate(${roundByDPR(x)}px,${roundByDPR(y)}px)` : null,
+      left: open ? `0` : null,
+      top: open ? `0` : null,
     });
   };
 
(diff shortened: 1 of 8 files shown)


driskull:
@jcfranco this seems to fix the drag and drop issues with flying in from the right/left.

However, closed floating ui elements are still taking up space. What do you think?

driskull:
Actually, if I add overflow hidden on elements that are dragged it seems to be good.

Can you review/test? @jcfranco

packages/calcite-components/src/utils/floating-ui.ts
```diff
@@ -145,15 +150,15 @@ export const positionFloatingUI =
...
+    const { open } = component;
 
     Object.assign(floatingEl.style, {
       visibility,
       pointerEvents,
       position,
-      top: "0",
-      left: "0",
-      transform,
+      transform: open ? `translate(${roundByDPR(x)}px,${roundByDPR(y)}px)` : null,
```
jcfranco, line 159: Type-wise, should we pass `""` vs `null`?

jcfranco (approved):
🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈
🎈🆙🆙🆙🆙🎈🆙🎈🎈🎈🎈🎈🆙🎈🆙🆙🆙🆙🎈🎈🆙🆙🆙🎈🎈🆙🆙🎈🎈🆙🎈🎈🎈🆙🎈🆙🆙🆙🆙🎈🆙🎈
🎈🆙🎈🎈🆙🎈🆙🎈🎈🎈🎈🎈🆙🎈🆙🎈🎈🎈🎈🆙🎈🎈🎈🎈🆙🎈🎈🆙🎈🆙🆙🎈🆙🆙🎈🆙🎈🎈🎈🎈🆙🎈
🎈🆙🆙🆙🆙🎈🆙🎈🎈🆙🎈🎈🆙🎈🆙🆙🆙🎈🎈🎈🆙🆙🎈🎈🆙🎈🎈🆙🎈🆙🎈🆙🎈🆙🎈🆙🆙🆙🎈🎈🆙🎈
🎈🆙🎈🎈🆙🎈🆙🎈🆙🎈🆙🎈🆙🎈🆙🎈🎈🎈🎈🎈🎈🎈🆙🎈🆙🎈🎈🆙🎈🆙🎈🎈🎈🆙🎈🆙🎈🎈🎈🎈🎈🎈
🎈🆙🎈🎈🆙🎈🎈🆙🎈🎈🎈🆙🎈🎈🆙🆙🆙🆙🎈🆙🆙🆙🎈🎈🎈🆙🆙🎈🎈🆙🎈🎈🎈🆙🎈🆙🆙🆙🆙🎈🆙🎈
🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈

jcfranco:
> Can you review/test?

FYI, I only reviewed. Haven't had a chance to test.

packages/calcite-components/src/utils/floating-ui.ts
driskull: Updated.

driskull:
@jcfranco good to merge?

Issue #8223, opened 2023-11-20
stepper-item: Remove unused messages 


anveshmekala:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Currently, `previousStep` and `nextStep` message strings are not used in the component  which are added as part of https://github.com/Esri/calcite-design-system/pull/7918/ and they are moved into `stepper` after the spec discussions.

### Acceptance Criteria

Remove `previousStep` and `nextStep` message strings from locale bundles.

### Which Component

calcite-stepper-item

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

geospatialem:
Verified the translations are no longer present on the `main` branch.

PR #8234, opened 2023-11-21
refactor(stepper-item)!: no longer supports previousStep and nextStep messages

anveshmekala:
**Related Issue:** #8223 

## Summary
BREAKING CHANGE: Removed `previousStep` and `nextStep`  in messages. They are not available for overriding via `messageOverrides`


jcfranco (approved):
Changes LGTM!

Conventional-commit-wise, would `refactor`/`chore` be a more appropriate type for this?

anveshmekala:
> Conventional-commit-wise, would `refactor`/`chore` be a more appropriate type for this?

I gave a thought about it. `chore` might not be a better fit considering this can break some user's code. It's a hard choice between `refactor` and `feat ` . From calcite perspective it is a refactor since those messages are not utilized yet, for the end user it could still be breaking some custom feature. Thoughts?

jcfranco:
> rom calcite perspective it is a refactor since those messages are not utilized yet, for the end user it could still be breaking some custom feature

The type should reflect the changes not what it could impact, so I think `refactor` makes sense here.

PR #8240, opened 2023-11-21
fix(list-item): an item with an empty slotted list should be openable.

driskull:
**Related Issue:** #8241

## Summary

- Sets a list item as `openable` when a list without any items is slotted inside it.
- Previously, an empty list was not accounted for.
- Add story


packages/calcite-components/src/components/list-item/utils.ts
```diff
@@ -4,6 +4,12 @@ const listSelector = "calcite-list";
 const listItemGroupSelector = "calcite-list-item-group";
 const listItemSelector = "calcite-list-item";
 
+export function getListItemChildLists(slotEl: HTMLSlotElement): HTMLCalciteListElement[] {
```
jcfranco, line 7: Sidebar: we should generalize these and move into `dom`. These are some pretty nice utils.

jcfranco (approved):
✨📕➡️📖✨

Issue #8243, opened 2023-11-21
input: value is increasing/decreasing if the user tabs out of the input with `ArrowUp/ArrowDown` key down.


anveshmekala:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

If a user `tab` out of the input with `ArrowUp/ArrowDown` key down, the value keeps increasing/decreasing. 

### Expected Behavior

The value should stop increasing/decreasing when the user `tab` out.

### Reproduction Sample

https://codepen.io/ANVESHMEKALA/pen/ExrEGEa

### Reproduction Steps

1. Click or Tab into the input component.
2. Use either of ArrowUp or ArrowDown key to increase/decrease the value.
3. Observe that value changes in input.
4. Hold the Arrow key and tab out of the input.
5. Observe that value in input keeps changing.

### Reproduction Version

1.10.0

### Relevant Info

Extension of https://github.com/Esri/calcite-design-system/issues/7188

### Priority impact

p2 - want for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

PR #8245, opened 2023-11-21
fix(input): prevents mutating value on `blur` when `type="number"`

anveshmekala:
**Related Issue:** #8243 

## Summary
Prevents increasing/decreasing the value in `<calcite-input type="number" />` on `blur` using `Tab`


driskull (approved):
👍

jcfranco (approved):
✨🏆✨

Issue #8156, opened 2023-11-10
Calcite Shell Panel: Resizable with iframe in calcite shell 


rslibed:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Resize handle does not expand the shell panel as expected when calcite shell contains an iframe. 

### Expected Behavior

Resize handle does expand the shell panel when calcite shell contains an iframe. 

### Reproduction Sample

https://codepen.io/rslibed/pen/MWLmmBY

### Reproduction Steps

Use resize handle to increase/decrease shell panel size. Notice that reducing the panel size does not have any issues; however, increasing the panel size does not continue to expand as expected.

### Reproduction Version

1.10.0

### Priority impact

p3 - want for upcoming milestone

### Impact

Received a request to be able to resize panel within ArcGIS Instant Apps configuration experience. 

### Calcite package

- [X] @esri/calcite-components
- [X] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Instant Apps

macandcheese:
Note (can make a separate issue but adding from Teams) - There's a similar issue with iFrame / Tooltip dismissal - if the cursor moves from the tooltip reference element to an iFrame, it's never dismissed.

geospatialem:
In `1.12.0-next.6` this works well when the mouse cursor hovers over the `shell`'s `"header"` slot and with the arrow keys, but when the mouse cursor is anywhere else, it doesn't seem to respond at all - is that intended, @driskull?

![resizing-shell-panel-iframe](https://github.com/Esri/calcite-design-system/assets/5023024/f20d734e-22c8-4b66-a81d-4024a4ce36e3)

driskull:
No, not intended. Ill see whats going on there.

jcfranco:
Querying for all iframes and modifying them could get tricky and might still run into issues for complex content (e.g., iframes within shadow roots 😱). Would it be possible to use an overlay on the shell or document to track mouse movement when resizing is initiated via pointer?

driskull:
Yeah I can try that. It would need to be handled in the shell and not the shell-panel so maybe via some internal event.

DitwanP:
Verified locally on `main` 🍬

PR #8247, opened 2023-11-22
fix(color-picker, popover, shell-panel, slider, tooltip): Register events on the window instead of the document

driskull:
**Related Issue:** #8156

## Summary

- Sets up global event listeners on the window instead of the document.
- This is because the events should continue to work beyond the document in cases where an iframe is being used.
- We want to continue to listen to these events for the entire window.


driskull:
@jcfranco we may want to add some kind of eslint rule(s) for this as a followup issue.

alisonailea (approved):

benelan (approved):
:+1:

PR #8253, opened 2023-11-22
fix(dropdown)!: change display to inline-block to ease truncation setup

jcfranco:
**Related Issue:** #5903

## Summary

BREAKING CHANGE: Dropdown's default `display` was changed from `inline-flex` to `inline-block` to make it easier to prompt truncation in trigger button text with minimal impact to layout (by setting an explicit width **or** setting `width: 100%` or `display: block` on the dropdown of a width-constrained parent).


benelan (approved):
LGTM! Small nitpick in the deprecation notice - maybe change the first "trigger" to "activate", which sounds less clunky

PR #8262, opened 2023-11-27
fix(list-item): adds border between last item in a group and slotted item

anveshmekala:
**Related Issue:** #7546 

## Summary

Adds border between last `calcite-list-item` in a `calcite-list-item-group` and next slotted `calcite-list-item`.


jcfranco (approved):
📝🔲

Issue #8007, opened 2023-10-13
[input-time-zone] updates items when related props are updated


jcfranco:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Changing any of the props that are used to generate the offset options recreate items but these are not rendered. 

### Expected Behavior

Changing `lang`, `messageOverrides`, `mode` or `referenceDate` updates and rerenders items.

### Reproduction Sample

https://codepen.io/jcfranco/pen/QWzYOLe

### Reproduction Steps

1. look at `applyTip` and notice that the component needs to be recreated to pick up the updated options when rendering
2. update the method to reuse the current component and notice that items will not update

### Reproduction Version

1.9.2

### Relevant Info

[This line](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/input-time-zone/input-time-zone.tsx#L119) needs to be updated, so that the recreated items are used in rendering.

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

geospatialem:
Verified with `1.12.0-next.6`

PR #8271, opened 2023-11-28
fix(input-time-zone): update time zone items when item-dependent props change

jcfranco:
**Related Issue:** #8007

## Summary

This addresses an issue where `input-time-zone` was recreating time zone groups when related props were modified, but these were not affecting state, so the items would remain unchanged.


driskull (approved):
💯 Beautiful 🌈

Issue #8202, opened 2023-11-17
calcite-action-menu should allow groups of related actions


sagewall:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

We are refactoring the JavaScript Maps SDK [LayerList](https://developers.arcgis.com/javascript/latest/api-reference/esri-widgets-LayerList.htm) widget to use calcite components. The LayerList [ListItem](https://developers.arcgis.com/javascript/latest/api-reference/esri-widgets-LayerList-ListItem.html) has a property [actionsSections](https://developers.arcgis.com/javascript/latest/api-reference/esri-widgets-LayerList-ListItem.html#actionsSections) that defines a nested 2-dimensional collection of actions that could be triggered on the item. In the previous design of the LayerList these action sections were visually separated by some empty space.
![image](https://github.com/Esri/calcite-design-system/assets/2636825/20e6c588-4cc9-49c8-8ef7-4971a22c2b63)

It would be nice if there was a way to create groups of actions inside a calcite-action-menu to be able to categorize and group the actions. Currently we have no way to visually separate the two defined action sections.
![image](https://github.com/Esri/calcite-design-system/assets/2636825/6bd3b8f1-4486-469a-9b30-44a0b8f73934)

cc @driskull 

### Acceptance Criteria

The ability to visually distinguish between groups of related actions in the calcite-action-menu

### Which Component

calcite-action-menu

### Example Use Case

https://codepen.io/sagewall/pen/ExroQVb

### Priority impact

p2 - want for current milestone

### Calcite package

- [X] @esri/calcite-components
- [X] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

macandcheese:
An option with the current components could be to use a Dropdown there. It supports Groups, and in selection-mode="none" should behave like an Action menu. You can choose to add `group-title`, but when not present, has a visual line between the items.
<img width="436" alt="Screenshot 2023-11-17 at 11 23 09 AM" src="https://github.com/Esri/calcite-design-system/assets/4733155/07226f3e-c023-4a44-b5d5-d123bea58881">
[...]

PR #8273, opened 2023-11-28
feat(action-menu): support action-groups

driskull:
**Related Issue:** #8202

## Summary

- Adds support for `calcite-action-group` within `calcite-action-menu`
- Adds screenshot test
- Adds e2e test


packages/calcite-components/src/components/action-menu/action-menu.tsx
```diff
@@ -401,7 +401,21 @@ export class ActionMenu implements LoadableComponent {
       .assignedElements({
         flatten: true,
       })
-      .filter((el) => el?.matches("calcite-action")) as HTMLCalciteActionElement[];
+      .reduce((previousValue: HTMLCalciteActionElement[], currentValue) => {
+        if (currentValue?.matches("calcite-action")) {
+          return [...previousValue, currentValue];
```
benelan, line 406: Could we we do `return previousValue.push(currentValue)` for better performance? I don't think we need to do a shallow clone here by spreading.

packages/calcite-components/src/components/action-menu/action-menu.tsx
```diff
@@ -401,7 +401,21 @@ export class ActionMenu implements LoadableComponent {
...
+        }
+
+        if (currentValue?.matches("calcite-action-group")) {
+          const groupActions = Array.from(
+            currentValue.querySelectorAll("calcite-action")
+          ) as HTMLCalciteActionElement[];
+
+          return [...previousValue, ...groupActions];
```
benelan, line 414: This can be `return previousValue.push(...groupActions)` for better performance

benelan (approved):

packages/calcite-components/src/components/action-menu/action-menu.tsx
driskull: Yep!

packages/calcite-components/src/components/action-menu/action-menu.tsx
driskull: done!

jcfranco: > for better performance

Just my two cents: unless we're dealing with a critical path, prioritizing readability/maintainability over micro-optimizations might be more beneficial. It's easy to get caught in the perf optimization rabbit hole with little to gain. Also, it's worth noting that modern browsers are pretty good at optimizing code themselves.

Issue #8277, opened 2023-11-28
Break Core Design Tokens into separate files for the different token group types


alisonailea:
### Description

Similar to Semantic tokens, Core tokens should be split into separate files organized in a "core" directory to facilitate ease of use when looking for a specific variable or group of variables.

This will have no effect on the behavior or output of the Calcite Design Tokens

### Proposed Advantages

- Make it easier to find existing tokens
- Make it easier to add tokens to specific token groups

### Which Component

all core tokens

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [X] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

jcfranco:
@alisonailea I went ahead and marked this one as installed for v2. Could you update the labels as needed?

geospatialem:
Verified in [calcite-design-tokens](https://www.npmjs.com/package/@esri/calcite-design-tokens) in `2.0.0`.

PR #8279, opened 2023-11-28
refactor(core tokens): split core tokens into groups

alisonailea:
**Related Issue:** #8277

## Summary

This is a refactor only and will have no effect on existing output. Similar to Semantic tokens, this PR breaks the Core tokens into separate token type groups and puts them in a "core" token directory.


packages/calcite-design-tokens/src/index.ts
```diff
@@ -17,19 +16,4 @@ export const globalTokens = await createCalciteTokenFiles({
   references: coreTokens.source,
 });
 
-export const componentTokens = await getFiles(resolve(__dirname, "component")).then((components) =>
```
jcfranco, line 20: Please correct me if I'm wrong, but this seems unrelated to the group changes. If so, we can either revert here or merge #8280 before merging this one.

jcfranco (commented):
This should use the `refactor` commit type as it is not providing a fix.

packages/calcite-design-tokens/src/index.ts
alisonailea: oop thanks for the catch

jcfranco (approved):
✨👥👥👥➡️👤👤👤✨

Issue #8278, opened 2023-11-28
Remove unused Component Design Tokens


alisonailea:
### Description

There is a directory of "component" tokens in the calcite-design-tokens which are out of date and not currently in use. These should be removed.

### Proposed Advantages

- reduce the size of the calcite-design-token repo
- reduce confusion for consumers who may think the "component" tokens should be used

### Which Component

all component files in calcite-design-tokens

### Relevant Info

If/when component tokens are align on in Design these can be added back on an as-needed basis.

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [X] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

jcfranco:
@alisonailea I went ahead and marked this one as installed for v2. Could you update the labels as needed?

geospatialem:
Verified in [calcite-design-tokens](https://www.npmjs.com/package/@esri/calcite-design-tokens) in `2.0.0`.

PR #8280, opened 2023-11-28
refactor(component tokens): remove old component tokens

alisonailea:
**Related Issue:** #8278

## Summary

These component tokens are out of date and should not be used. To avoid confusion and reduce the size of the repo we should remove these files.


driskull (approved):
🗑️ 👋

jcfranco (approved):
Awesome! ✨🧹✨ 

A couple of things I spotted:

1. This should probably use `refactor` as the type since removing tokens isn't a fix 
2. There's a space between the type and the scope that needs to be removed for this to adhere to conventional commits

Other than that, this LGTM!

PR #8281, opened 2023-11-28
fix: allow users to control tabindex on interactive components

jcfranco:
**Related Issue:** #4970 

## Summary

Updates the interactive component pattern to avoid the need to control `tabindex` on the host to prevent tabbing into a component's subtree.

This works by introducing a new functional component (`InteractiveContainer`) that leverages `inert` (prevent tabbing into its contents) and `display: contents` (allows us to use it as a root container without affecting layout), which should be the root container under the host. 

Updating components to this pattern should be straightforward as the styling of the helper component is bundled within the `disabled` mixin that interactive components already include and `updateHostInteraction` no longer needs its 2nd parameter. For cases that were setting the 2nd parameter (`hostIsTabbable`) all they have to do to migrate depends on the value:

1. `false` (default) – no further changes necessary
2. `true` – set `tabIndex={0}` on the host
3. predicate function – set `tabIndex={myPredicateFunction() ? 0 : -1 }` on the host
4. `"managed"` – no further changes necessary as owner/parent sets `tabIndex` on the component already

Any use of `Fragment` can be replaced by `InteractiveContainer` directly.

2 and 3 might also need to determine the tab index based on `disabled`.


jcfranco:
Recreated https://github.com/Esri/calcite-design-system/pull/8166 on `rc` to avoid merge conflicts from `main`.

Issue #8282, opened 2023-11-28
Handle: add disabled property


driskull:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

A disabled property on the handle would be super useful to be able to prevent dragging on a specific item.

### Acceptance Criteria

Add disabled property to handle to prevent interaction and appear disabled

### Which Component

handle

### Example Use Case

<calcite-handle disabled></calcite-handle>

### Priority impact

p2 - want for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

DitwanP:
Verified locally on `main`

PR #8283, opened 2023-11-28
feat(handle): Add disabled property

driskull:
**Related Issue:** #8282

## Summary

- Adds disabled property to handle component
- Use interactive component on handle
- Adds e2e test
- Adds story screenshot test


benelan (approved):
:+1:

Issue #8287, opened 2023-11-28
fix tokens output for Gatby documentation


alisonailea:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

the json output for semantic.color.border.input breaks when we try to automate documentation in Gastby. Then word "input" is reserved word and must be updated.

js output is only used by gatsby but it needs json

### Expected Behavior

Change to `semantic.color.border.user-input`

Change js format to json

### Reproduction Sample

N/a

### Reproduction Steps

1. take the JSON output of calcite-design-tokens
2. load it into Gatsby
3. Observe a build failure

### Reproduction Version

N/A

### Priority impact

p0 - emergency

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [X] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

alisonailea:
Per conversation on https://github.com/Esri/calcite-design-system/pull/8288 and also over chat, we will leave "input" as is and instead create a new output format for Gatsby which uses an array of design tokens instead of setting each design token to a named key. 

```json
"color":  [
   ...,
  { name: "input", ....},
  ...
]
```

instead of 

```json
"color": {
  ...,
  input: { ... },
  ...
}
```

jcfranco:
@alisonailea I went ahead and marked this one as installed. Could you add a quick summary of changes from the PR?

geospatialem:
Verified with Adam's internal doc site preview branch.

PR #8288, opened 2023-11-28
fix(color tokens): "input" should be "user-input"

alisonailea:
**Related Issue:** #8287

## Summary
Resolves an issue with reserved words in Gasby when trying to generate design tokens documentation

### DIFF
--- a/packages/calcite-design-tokens/src/semantic/color.json
+++ b/packages/calcite-design-tokens/src/semantic/color.json
@@ -147,7 +147,7 @@
           "value": { "light": "{core.color.neutral.blk-030}", "dark": "{core.color.neutral.blk-180}" },
           "type": "color"
         },
-        "input": {
+        "user-input": {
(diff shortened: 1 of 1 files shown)


jcfranco:
I'm hesitant about modifying our data/tokens to address the Gatsby issue. Mostly from a maintenance/scalability perspective. Could we explore an alternative solution to avoid these issues?

alisonailea:
> I'm hesitant about modifying our data/tokens to address the Gatsby issue. Mostly from a maintenance/scalability perspective. Could we explore an alternative solution to avoid these issues?

I don't know what an alternative would be. We need to document tokens and we need to automate that process. This is the only instance where there is a problem because it's the only place we are using the word "input" alone. Adam and I did try to find an alternative but were unable to do so. I have already run this by design and they have agreed this is a good solution.

jcfranco:
> We need to document tokens and we need to automate that process. 

This means there could be other automation scenarios that could also choke on our design token output, right?

IMO, this is a build/tooling issue and should be addressed at that level. cc @macandcheese

PR #8290, opened 2023-11-28
feat: add json to design token output

alisonailea:
**Related Issue:** #8287

## Summary

updates token output to change js to json.
also fixes files array in package.json to allow SCSS imports using "~@esri/calcite-design-tokens" in calcite-components


jcfranco (changes requested):
Since this is dropping the `js` output, this should be marked as a breaking change.

packages/calcite-design-tokens/support/token-transformer/styleDictionary/formatter/utils.ts
```diff
@@ -1,14 +1,14 @@
...
-import { CalciteJs } from "./javascript.js";
+import { CalciteJson } from "./json.js";
 import { CalciteScss } from "./scss.js";
 
 const formatters: Partial<Record<PlatformFormats, string>> = {
   css: CalciteCss,
   scss: CalciteScss,
   sass: CalciteScss,
-  js: CalciteJs,
+  json: CalciteJson,
```
jcfranco, line 11: Do we need a custom formatter or can we use the predefined formatter for JSON?

jcfranco: Did a shallow test and it looks like contents are identical.

jcfranco:
> Since this is dropping the js output, this should be marked as a breaking change.

Alternatively, we could keep the JS output and add the JSON output.

jcfranco:
Sorry for not catching this sooner, but can you also update the spec tests?

alisonailea:
> Sorry for not catching this sooner, but can you also update the spec tests?

opened a different PR to fix the snapshot tests but added the test for the JSON output here

packages/calcite-design-tokens/support/token-transformer/styleDictionary/formatter/utils.ts
alisonailea: updated to use "json"

packages/calcite-design-tokens/package.json
```diff
@@ -9,31 +9,10 @@
     "tokens"
   ],
   "type": "module",
-  "main": "./dist/js/global.js",
-  "types": "./dist/js/global.d.ts",
+  "main": "./dist/es6/global.js",
```
jcfranco, line 12: Does this need to be reverted since we're keeping the JS output?

jcfranco (approved):
LGTM! 🚀

Since this is now adding the JSON output, can you update the PR title similar to bf9e03a7b0629d1a3db18c6b104f665666167fe9 before merging?

packages/calcite-design-tokens/package.json
jcfranco: Makes sense, but this is now unrelated to this PR and I _think_ falls under breaking change territory.

Issue #7974, opened 2023-10-11
[List] Accessibility Issues during keyboard navigation to "action-end" slot of List-Item


pllcoster:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Summary

When navigating a list-item using the keyboard, and navigating to the "action-end" slot of the list item:
- navigating using tab, I can press enter/spacebar to open a calcite dropdown in the "action-end" slot
- navigating using left-right arrow keys, pressing enter/spacebar on the same item has no effect

Navigating using arrow keys makes the action-end item look focused, however the item cannot be used as expected, which may cause accessibility problems.

Related issue https://github.com/Esri/calcite-design-system/issues/7936

### Actual Behavior

I have found that if you have a Calcite List, you can navigate from the List Item to the item in its "actions-end" slot using tab, and then (e.g. if the action-ends slot contains a Calcite Dropdown trigger action) you can use enter or spacebar to trigger the dropdown:
 

List Item Focused:
![image](https://github.com/Esri/calcite-design-system/assets/146282673/ca096a73-7088-4b26-ba84-d8028395763b)

 
Tab to actions-end slot - pressing enter or spacebar will open the dropdown:
![image](https://github.com/Esri/calcite-design-system/assets/146282673/67e70304-c07f-4fc9-b55b-bc515dcc4930)

 
However once you have the actions-end slot focused, you can navigate around using the keyboard arrow keys, if you navigate away from the actions-end icon and then back to it, it looks like it is focused again (although the styling is different):
![image](https://github.com/Esri/calcite-design-system/assets/146282673/a017bffb-36fb-4b83-b0cb-20d2853e306c)

 

And now pressing enter or spacebar has no effect - the dropdown does not open.

### Expected Behavior

Expected that if the action-end item is focused it should open by pressing enter or spacebar.

### Reproduction Sample

https://codepen.io/pcoster/pen/oNJmZmj

### Reproduction Steps

Follow the steps outlined in 'Actual Behavior' above

### Reproduction Version

1.9.1

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Knowledge

geospatialem:
Research spike to determine next steps regarding tree grid role.

geospatialem:
[...]

PR #8291, opened 2023-11-29
fix(list-item): focus on the first focusable element within the component when using arrow keys

driskull:
**Related Issue:** #7974

## Summary

- Focuses on the first tabbable element inside the grid cell when using arrow keys.
- Excludes hidden cells from the arrow key logic
- Adds dom utility `getFirstTabbable` to get first focusable element from within an element.

### DIFF
--- a/packages/calcite-components/src/components/list-item/list-item.tsx
+++ b/packages/calcite-components/src/components/list-item/list-item.tsx
@@ -12,7 +12,12 @@ import {
   VNode,
   Watch,
 } from "@stencil/core";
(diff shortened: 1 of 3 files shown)


jcfranco (approved):
Awesome! 1️⃣🔍👁️

Can you add a test for this before merging?

geospatialem (approved):
Nice work! 🥇 💪🏻 

Testing is working great with JAWS, NVDA, and keyboard without AT active. 🙌🏻

Issue #7590, opened 2023-08-23
[Calcite Button] Loading buttons are clickable in Firefox


mpayson:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Similar to https://github.com/Esri/calcite-design-system/issues/7043, buttons in a `loading` state are clickable in Firefox

### Expected Behavior

Since the button visually looks disabled, I'd expect it to have the same behavior as the `disabled` state and not be clickable

### Reproduction Sample

https://codepen.io/mpayson/pen/poqJZpY

### Reproduction Steps

1. Open code pen and devtools in Firefox
2. Click loading button, see console message
3. Click disabled, don't see console message

### Reproduction Version

1.6.1

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [X] @esri/calcite-components-react

### Esri team

ArcGIS Online

geospatialem:
Update pattern for consistency across components, such as `block`, and `button` to truly be disabled when loading is enabled. For `panel` could use `block`'s scrim paradigm, where the header isn't covered and actions are available for use.

amring1:
Can we up the priority on this one? It causes issues in analysis where users can submit multiple jobs in quick succession which causes failures.

jcfranco:
This also affects Safari. 

We fixed stacked opacity on `calcite-button` when they are both `disabled` and `loading` in v1.10.0, so as a workaround, you can set both props on the component. https://codepen.io/jcfranco/pen/xxMELeX?editors=1111

rebollin:
Hi there, would also like to up the priority of this bug if possible. This was reported as a release issue when saving in Map Viewer (it allows the user to create multiple maps). The workaround only partially resolved the issue, it works in chrome but not in FF

geospatialem:
For consistency across the design system Calcite will keep the `loading` and `disabled` states as separate. 

In cases where `loading` is present and a `disabled` state is desired, folks can add both attributes, or properties, in to achieve the desired use case where the button in its loading state is disabled, as depicted in [Franco's reply above](https://github.com/Esri/calcite-design-system/issues/7590#issuecomment-1791059087), such as:

```html
<calcite-button id="loading" loading disabled>Add layer</calcite-button>
<script>
document.getElementById("loading").onclick = () => console.log("loading clicked")
</script>
```

[...]

PR #8292, opened 2023-11-29
fix(button, list, pick-list, value-list)!: prevent loading prop from affecting interactivity

jcfranco:
**Related Issue:** #7590

## Summary

This updates some components to avoid setting `disabled` on internal, supporting components when `loading` was set to true.

BREAKING CHANGE: Setting `loading` prop to true no longer prevents interaction nor applies disabled styles. If you'd like to block interaction when loading, please set `disabled` along with `loading`.

### DIFF
--- a/packages/calcite-components/src/components/button/button.scss
+++ b/packages/calcite-components/src/components/button/button.scss
@@ -160,15 +160,6 @@
 
 @include disabled();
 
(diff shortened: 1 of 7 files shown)


driskull (approved):
Awesome sauce! 🛸 👍 

w00t w00t w00t

Issue #6121, opened 2022-12-20
calcite-dropdown: open dropdown not moving on scroll after re-ordering blocks


AdelheidF:
### Actual Behavior

Before re-ordering blocks the open dropdown moves along with the panel when scrolling via mouse wheel.
After re-ordering blocks the open dropdown stays in place.

cc @driskull 

### Expected Behavior

open dropdown always moves with the panel scroll

### Reproduction Sample

https://codepen.io/afreitag/pen/RwBbgxK

### Reproduction Steps

1. open "Preview count" dropdown and scroll panel via mouse wheel, watch the open dropdown move correctly
2. re-order block C before block B
3. now again open "Preview count" dropdown and scroll panel, notice the open dropdown sticks in place

### Reproduction Version

beta.99

### Regression?

don't think so

### Impact

medium

### Esri team

ArcGIS Map Viewer

geospatialem:
Additional research is needed, there may be a workaround related to the #6520 PR.

driskull:
@AdelheidF is this issue still present?

lau11123:
@driskull yeah, it's still an issue.

AdelheidF:
Looking good for me too. Thanks.

PR #8296, opened 2023-11-29
fix(combobox, dropdown, input-date-picker, popover, tooltip): fix positioning of component when component is moved

driskull:
**Related Issue:** #6121

## Summary

- When a component is moved (via drag and drop or whatever), we need to reconnect floating UI (connectFloatingUI).
- This is needed because it sets up autoUpdating of the reference element so that the reference element is tracked and positioned properly
- Currently, when an a component is moved, this tracking is lost because we were only calling `connectFloatingUI` when the reference element and floating element were created.
- This PR makes sure that we reestablish the floating UI autoUpdate when the component itself is moved.

### DIFF
--- a/packages/calcite-components/src/components/combobox/combobox.tsx
+++ b/packages/calcite-components/src/components/combobox/combobox.tsx
@@ -439,6 +439,7 @@ export class Combobox
       this.openHandler();
       onToggleOpenCloseComponent(this);
     }
+    connectFloatingUI(this, this.referenceEl, this.floatingEl);
   }
 
   async componentWillLoad(): Promise<void> {
--- a/packages/calcite-components/src/components/dropdown/dropdown.tsx
+++ b/packages/calcite-components/src/components/dropdown/dropdown.tsx
@@ -209,6 +209,7 @@ export class Dropdown
     }
     connectInteractive(this);
     this.updateItems();
+    connectFloatingUI(this, this.referenceEl, this.floatingEl);
   }
 
   componentWillLoad(): void {
--- a/packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
+++ b/packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
@@ -475,6 +475,7 @@ export class InputDatePicker
     if (this.open) {
       onToggleOpenCloseComponent(this);
     }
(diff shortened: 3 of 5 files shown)


jcfranco (approved):
🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯
🗯🎈🗯🗯🗯🎈🗯🗯🎈🎈🗯🗯🎈🎈🎈🗯🗯🎈🎈🎈🗯🎈🎈🎈🎈🗯🎈🗯
🗯🎈🎈🗯🗯🎈🗯🎈🗯🗯🎈🗯🗯🎈🗯🗯🎈🗯🗯🗯🗯🎈🗯🗯🗯🗯🎈🗯
🗯🎈🗯🎈🗯🎈🗯🎈🗯🗯🎈🗯🗯🎈🗯🗯🎈🗯🗯🗯🗯🎈🎈🎈🗯🗯🎈🗯
🗯🎈🗯🗯🎈🎈🗯🎈🗯🗯🎈🗯🗯🎈🗯🗯🎈🗯🗯🗯🗯🎈🗯🗯🗯🗯🗯🗯
🗯🎈🗯🗯🗯🎈🗯🗯🎈🎈🗯🗯🎈🎈🎈🗯🗯🎈🎈🎈🗯🎈🎈🎈🎈🗯🎈🗯
🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯🗯

jcfranco:
Would it be possible to set up an E2E test for this or is it too complex? If it's the latter, can you create a follow-up issue to add a test? I think testing might be simpler if we had  storybook interactions in place.

driskull:
> Would it be possible to set up an E2E test for this or is it too complex? If it's the latter, can you create a follow-up issue to add a test? I think testing might be simpler if we had storybook interactions in place.

Yeah would be pretty difficult to automate a test here. Ill create a followup issue.

driskull:
https://github.com/Esri/calcite-design-system/issues/8297

PR #8299, opened 2023-11-29
feat!: update default main file output for design tokens

alisonailea:
**Related Issue:** n/a

## Summary

Engineers prefer to use es6 to import specific tokens when necessary. The es6 format (which is tree-shakable) should be the one imported by default.

BREAKING CHANGE:

- `import * as tokens from "@esri/calcite-design-tokens";` now imports a tree-shakable list of design tokens in camelCase format rather than a JSON object.


jcfranco (approved):
Noice! 🚀👾🛰️

Issue #7966, opened 2023-10-09
calcite-list: support same 'shift' selection support as calcite-value-lists and don't select filtered out items


AdelheidF:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

value-list supported using the shift key to select multiple rows at once. This does not work with the new list.

https://codepen.io/afreitag/pen/jOXdbwq

If value list has a filter applied and you do this shift key selection then it also selects the hidden items. This should be fixed in list, filtered out items should not get selected.

### Acceptance Criteria

I can select 10 rows at once by clicking on the first row and clicking on the 10th row while holding the shift key pressed. Filtered out items should not be selected in this case.

### Which Component

calcite-list

### Example Use Case

Map Viewer attributes component which used value-list so far. But value-list has an issue when selecting multiple rows at once while a filter is in place. It ignores the filter in this case and selects too many items.

### Priority impact

p2 - want for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Map Viewer

macandcheese:
This would be nice to add to Table as well.

AdelheidF:
Looks goo to me in 2.0 as well. Thanks.

PR #8301, opened 2023-11-29
feat(list): support multiple selection using the shift key

driskull:
**Related Issue:** #7966

## Summary

- support multiple selection using the shift key

### DIFF
--- a/packages/calcite-components/src/components/list-item/list-item.tsx
+++ b/packages/calcite-components/src/components/list-item/list-item.tsx
@@ -225,7 +225,7 @@ export class ListItem
   /**
    * Emits when the item's content is selected.
    */
-  @Event({ cancelable: false }) calciteListItemSelect: EventEmitter<void>;
+  @Event({ cancelable: false }) calciteListItemSelect: EventEmitter<{ multiple: boolean }>;
 
   /**
    * Fires when the close button is clicked.
@@ -724,16 +724,16 @@ export class ListItem
     this.open = !this.open;
   };
 
-  itemClicked = (event: Event): void => {
+  itemClicked = (event: PointerEvent): void => {
     if (event.defaultPrevented) {
       return;
     }
 
-    this.toggleSelected();
+    this.toggleSelected(event.shiftKey);
     this.calciteInternalListItemActive.emit();
   };
 
-  toggleSelected = (): void => {
+  toggleSelected = (shiftKey: boolean): void => {
     const { selectionMode, selected } = this;
 
     if (this.disabled) {
@@ -746,7 +746,7 @@ export class ListItem
       this.selected = true;
     }
 
-    this.calciteListItemSelect.emit();
+    this.calciteListItemSelect.emit({ multiple: shiftKey && selectionMode === "multiple" });
   };
 
   handleItemKeyDown = (event: KeyboardEvent): void => {
@@ -767,7 +767,7 @@ export class ListItem
       !composedPath.includes(actionsEndEl)
     ) {
       event.preventDefault();
-      this.toggleSelected();
+      this.toggleSelected(event.shiftKey);
     } else if (key === "ArrowRight") {
       event.preventDefault();
       const nextIndex = currentIndex + 1;
--- a/packages/calcite-components/src/components/list/list.tsx
+++ b/packages/calcite-components/src/components/list/list.tsx
@@ -269,9 +269,26 @@ export class List
   }
 
   @Listen("calciteListItemSelect")
-  handleCalciteListItemSelect(): void {
+  handleCalciteListItemSelect(event: CustomEvent<{ multiple: boolean }>): void {
     if (!!this.parentListEl) {
       return;
+    }
+
+    const { enabledListItems, lastSelectedItem } = this;
+    const { multiple } = event.detail;
+    const selectedItem = event.target as HTMLCalciteListItemElement;
+
+    if (multiple && lastSelectedItem) {
+      const lastSelectedIndex = enabledListItems.indexOf(lastSelectedItem);
+      const currentIndex = enabledListItems.indexOf(selectedItem);
+      const startIndex = Math.min(lastSelectedIndex, currentIndex);
+      const endIndex = Math.max(lastSelectedIndex, currentIndex);
+
+      enabledListItems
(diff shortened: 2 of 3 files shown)


packages/calcite-components/src/components/list/list.tsx
```diff
@@ -269,11 +269,27 @@ export class List
   }
 
   @Listen("calciteListItemSelect")
-  handleCalciteListItemSelect(): void {
+  handleCalciteListItemSelect({ target, detail }: CustomEvent<{ multiple: boolean }>): void {
```
jcfranco, line 272: Is `multiple` strictly for internal purposes? If so, can we prevent `multiple` from being exposed in the public event type?

driskull: @jcfranco how would we do that? It might still be useful for public but correct its basically just for us.

packages/calcite-components/src/components/list/list.e2e.ts
```diff
@@ -252,6 +252,69 @@ describe("calcite-list", () => {
...
+    expect(await items[2].getProperty("selected")).toBe(false);
+    expect(await items[3].getProperty("selected")).toBe(false);
+
+    await page.$eval("#item-4", clickItemContent, `.${CSS.contentContainer}`);
+    await page.waitForChanges();
+    await page.waitForTimeout(listDebounceTimeout);
+    expect(eventSpy).toHaveReceivedEventTimes(2);
+    expect(await list.getProperty("selectedItems")).toHaveLength(4);
+
+    items.forEach(async (item) => expect(await item.getProperty("selected")).toBe(true));
```
benelan, line 295: `forEach` expects a synchronous function so using async/await doesn't always do what you expect. 

Can you change this to a `for...of` loop? Also applies to the two other instances in this `it` block.

packages/calcite-components/src/components/list/list.tsx
driskull: value-list did expose this as well as `shiftPressed`. https://github.com/Esri/calcite-design-system/blob/2b2506d07978b2fb66a63c1fed57c3f36c0d2996/packages/calcite-components/src/components/value-list-item/value-list-item.tsx#L192-L196

benelan (approved):
:+1:

packages/calcite-components/src/components/list/list.e2e.ts
driskull: Ill just get rid of the loop.

packages/calcite-components/src/components/list/list.tsx
driskull: I'm ok with it being public but i'll defer to @jcfranco. Not sure how to mark an event detail as `@internal`

driskull: I can create and emit an internal event instead but then I'm not sure if its guaranteed the internal one would be handled first before updating the selected items.

benelan:
Exporting the new event detail here will fix the build error:

https://github.com/Esri/calcite-design-system/blob/2b2506d07978b2fb66a63c1fed57c3f36c0d2996/packages/calcite-components/src/index.ts#L7-L13

packages/calcite-components/src/components/list/list.tsx
jcfranco: I think we could hide it from the public types by doing the following:

1. Type the item's select event emitter as `void`.
2. Use an internal type for the detail in the event handler (looks like `ListItemSelectDetail` in this case).
3. Use the internal type in the list's select event handler (as it currently does)
4. Not sure, but exporting the new event might no longer be needed as it wouldn't be used in the event type.

Based on the latest changes in this PR, you would need to do something like the following:

#### `list-item.tsx`

```ts
@Event({ cancelable: false }) calciteListItemSelect: EventEmitter<void>;
// ...
const internalDetail: ListItemSelectDetail = { 
  selectMultiple: shiftKey && selectionMode === "multiple" 
};
this.calciteListItemSelect.emit(internalDetail as any);
```

In addition, we could rename the prop in the interface to `_selectMultiple` to make it clear that it is internal.

WDYT?

driskull: @jcfranco I updated to use a separate internal event. Can you review again?

packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -238,6 +238,15 @@ export class ListItem
    */
   @Event({ cancelable: false }) calciteInternalListItemSelect: EventEmitter<void>;
 
+  /**
+   *
+   * @internal
+   */
+  @Event({ cancelable: false })
+  calciteInternalListItemSelectMultiple: EventEmitter<{
```
jcfranco, line 246: Nitpick: I think the event name or the detail prop can be renamed to avoid some redundancy.

jcfranco (approved):
🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏
🤏📝🤏🤏🤏📝📝📝📝🤏📝📝📝📝📝🤏📝🤏🤏🤏📝🤏📝🤏
🤏📝🤏🤏🤏📝🤏🤏🤏🤏🤏🤏📝🤏🤏🤏📝📝🤏📝📝🤏📝🤏
🤏📝🤏🤏🤏📝🤏📝📝🤏🤏🤏📝🤏🤏🤏📝🤏📝🤏📝🤏📝🤏
🤏📝🤏🤏🤏📝🤏🤏📝🤏🤏🤏📝🤏🤏🤏📝🤏🤏🤏📝🤏🤏🤏
🤏📝📝📝🤏📝📝📝📝🤏🤏🤏📝🤏🤏🤏📝🤏🤏🤏📝🤏📝🤏
🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏🤏

packages/calcite-components/src/components/list-item/list-item.tsx
driskull: Yeah I couldn't think of anything better. Since its internal i'll leave as is.

Issue #8286, opened 2023-11-28
[Docs]: Shell Panel's `heightScale` property not reflected properly


geospatialem:
### Description

Shell Panel's [`heightScale`](https://developers.arcgis.com/calcite-design-system/components/shell-panel/#api-reference-properties-heightScale) property currently lists all `layout` types as supported, however this is not the case. 

Instead, the doc should reflect only when the `layout` is `"horizontal"` - omitting the commas and the content between the commas.

### Which Component

`shell-panel`

### Resources

This came up in a [Community post](https://community.esri.com/t5/calcite-design-system-questions/heightscale-not-working-in-shell-panel/m-p/1347321#M529), where the documentation isn't accurately depicting the attribute/property's intent.

PR #8302, opened 2023-11-29
docs(shell-panel): remove incorrect information from heightScale prop

DitwanP:
**Related Issue:** #8286 

## Summary
Updated the description of the `heightScale` property for `Shell Panel` to correctly reflect the actual behavior.


driskull (approved):
👍 Looks good!

Issue #8057, opened 2023-10-25
Add Input Message to form components for form validation


benelan:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

We are in the process of improving our form validation story. One of the items is to replace the native validation message UI with `calcite-input-message` (#8000). First wee need to add an Input Message to our form components.

### Acceptance Criteria

- Add a `calcite-input-message` to our form components, which will display under the internal  element. 
- Add `validation-message` and `validation-icon` properties so users can change the message's values on their end.
- Investigate adding a way to adjust spacing between the internal form element and the input-message via a CSS variable.

### Which Component

The following input components need an internal input-message, and a `status` property if they don't have one.

- [x] combobox
- [x] input-time-zone
- [x] input-date-picker
- [x] input-number
- [x] input-text
- [x] input-time-picker
- [x] input
- [x] text-area

The following have the `required` property, which means they will need a validation message displayed via #8000.

- [x] select
- [x] checkbox [2]
- [x] radio-button-group [1]
- [ ] slider [1]
- [ ] rating [1] [3]
- [x] segmented-control [1] [3]

> [1]: needs design for `status` prop due to different UI than the "input components" listed above.
> [2]: needs design for internal `input-message` due to potential spacing issues
> [3]: `required` prop is marked as `internal` in #2777 pending design. Can be made public as part of this issue.

The following don't need an internal `input-message` for #8000 because they don't have any [constraint validation](https://developer.mozilla.org/en-US/docs/Web/HTML/Constraint_validation) properties. Should they receive an internal `input-message` anyway? If so, do they need a `status` property as well?

- [ ] switch
- [ ] meter

### Example Use Case

Form validation and other help messages.

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

benelan:
> @ashetland, @SkyeSeitz - I updated the issue body, but TLDR; the following components needs designs for the `status` property:
> 
> * `slider`
> * `radio-button-group`
> * `rating`
> * `segmented-control`
> 
> 
> Not sure where the invalid coloring should be applied for slider. Just the handle? Or the whole slider?
> 
[...]

Issue #3406, opened 2021-11-02
Enhancement: combobox invalid prop and status


sam11927:
### Description
Combobox should have an invalid state (add a status prop)
### Acceptance Criteria 
Combobox shows red border and can accept an input-message as well.
#### Which Component
`calcite-combobox`
#### Example Use Case
want to show that the combobox is a required field, so when someone clicks in and doesn't select something, set the status to invalid and display an error message (in the form of an `input-message`).

macandcheese:
@sam11927 do you still have a need for this invalid state on Combobox? 

The `input-message` can be placed adjacent to the Combobox itself like this: https://codepen.io/mac_and_cheese/pen/bGxQEMv?editors=1000

geospatialem:
Rescoped to fit the combobox's invalid status and include a prop.

geospatialem:
Part of a larger validation epic effort slotted for late 2023.

geospatialem:
Refer to the [Figma file](https://www.figma.com/file/5QDPzXr21qQcPTkXR1Ouga/Form-Validation-%5BEpic-4598%5D?type=design&node-id=956-94284&mode=design&t=tPqHJcVdxmJmZ6L2-0) and image below:

![image](https://github.com/Esri/calcite-design-system/assets/5023024/52eaf790-91bb-47e4-9bb1-9522c348c445)

geospatialem:
Verified in `2.1.0` with https://codepen.io/geospatialem/pen/PoLYZoL with the `status`, `validationIcon`, and `validationMessage` props.

PR #8304, opened 2023-11-30
feat(combobox, checkbox, input-time-zone, select, text-area): add status property

benelan:
**Related Issue:** #8057 #3406

## Summary

Add the `status` property to additional form components in preparation of adding internal input messages for validation. I split these changes out from #8305, which is where the new stories are.


jcfranco (approved):
⌨️👇💻✅❌😮

Issue #8000, opened 2023-10-12
Display a custom UI for validation messages instead of the native popover


benelan:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Our form components currently display the native validation popover when a form is submitted with invalid values. The native validation popover is styled differently depending on the browser. The plan is to replace the native validation popover with a builtin message.

This is required before some other validation issues can be completed. Otherwise there will be two different UIs for validation messages. Some of the validation issues blocked by this one are:
- #5723
- #4566
- #4276
- #3526

### Acceptance Criteria

The form util needs to be modified to prevent the native validation popover from displaying, and to display a custom one instead. ~~Eventually we need to add a way to display validation messages to all form components (maybe via a prop). I think design is TBD on that. In the meantime, we can try to use a sibling `calcite-input-message`.~~

> **Edit:** Design said we need to add the `calcite-input-message` to the form components as part of this issue due to spacing weirdness when displaying the message as a sibling component. See the prototypes below

### Prototypes

Current behavior using the native UI: 
https://codepen.io/benesri/pen/ExGpjaL
 
Sibling `calcite-input-message` created after form submission when there are invalid values (a lot of layout shifting): 
https://codepen.io/benesri/pen/YzBzVKE
 
Sibling `calcite-input-message` created on initial render (extra padding initially, plus a bit of layout shifting after submitting):
https://codepen.io/benesri/pen/PoVojVb
 
`calcite-popover` (will not be used in implementation):
https://codepen.io/benesri/pen/VwgwzPj?editors=1010

### Which Component

Changes to `form.ts` will be required. Adding Input Message to the components will be addressed by #8057

### Example Use Case

Forms

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

geospatialem:
Verified for the following components with the `required` attribute in `2.2.0-next.22` via https://codepen.io/geospatialem/pen/abMJqxP:

- [x] `input-text`
- [x] `input`
- [x] `input-date-picker`
- [x] `input-number`
- [x] `text-area`

![image](https://github.com/Esri/calcite-design-system/assets/5023024/065c6f80-ad4e-4734-a1d9-4dbb09239237)

PR #8305, opened 2023-11-30
feat: add `validation-message` and `validation-icon` attributes to form components for new internal `calcite-input-message`

benelan:
**Related Issue:** #8057

## Summary

Add an internal `calcite-input-message` to the following components 
- combobox
- input-date-picker
- input-number
- input-text
- input-time-picker
- input-time-zone
- input
- select
- text-area

The new properties will be used to replace the browser's popover element that displays validation messages on form submission (#8000).


anveshmekala:
Is it possible to avoid using `message` prefix for props since we have `messages` and `messageOverrides` props as part of translation effort. Thoughts?

benelan:
> Is it possible to avoid using `message` prefix for props since we have `messages` and `messageOverrides` props as part of translation effort. Thoughts?

Any other ideas? Maybe `validationText` (or `validationMessage` if suffix is okay) and `validationIcon`? cc @jcfranco

ashetland:
> > Is it possible to avoid using `message` prefix for props since we have `messages` and `messageOverrides` props as part of translation effort. Thoughts?
> 
> Any other ideas? Maybe `validationText` (or `validationMessage` if suffix is okay) and `validationIcon`? cc @jcfranco

`statusMessage`?

jcfranco:
> Any other ideas? Maybe validationText (or validationMessage if suffix is okay) and validationIcon? cc @jcfranco

If the message and icon are only displayed when the component goes through validation, I think `validationMessage` and `validationIcon` fit best. IMO, their intent is clearer vs using `statusMessage`/`statusIcon`. It would also align nicely w/ ElementInternal's [`validationMessage` prop](https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals/validationMessage).

packages/calcite-components/src/components/select/select.e2e.ts
```diff
@@ -22,7 +22,7 @@ describe("calcite-select", () => {
   `;
 
   describe("renders", () => {
-    renders(simpleTestMarkup, { display: "flex" });
+    renders(simpleTestMarkup, { display: "inline" });
```
jcfranco, line 25: The host display is `flex`, so this test should be failing. Can you double check?

jcfranco (approved):
✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️
✔️✅✔️✔️✔️✅✔️✔️✅✅✔️✔️✅✅✅✔️✔️✅✅✅✔️✅✅✅✅✔️✅✔️
✔️✅✅✔️✔️✅✔️✅✔️✔️✅✔️✔️✅✔️✔️✅✔️✔️✔️✔️✅✔️✔️✔️✔️✅✔️
✔️✅✔️✅✔️✅✔️✅✔️✔️✅✔️✔️✅✔️✔️✅✔️✔️✔️✔️✅✅✅✔️✔️✅✔️
✔️✅✔️✔️✅✅✔️✅✔️✔️✅✔️✔️✅✔️✔️✅✔️✔️✔️✔️✅✔️✔️✔️✔️✔️✔️
✔️✅✔️✔️✔️✅✔️✔️✅✅✔️✔️✅✅✅✔️✔️✅✅✅✔️✅✅✅✅✔️✅✔️
✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️✔️

As a follow-up, can we extract and reuse the common rendering logic and possibly helper CSS class?

PR #8308, opened 2023-11-30
feat: update json shape for docs

alisonailea:
**Related Issue:** #8287

## Summary

This changes the new format "json" to "docs" to better capture where this platform output should be used. This replaces the folder "json" with "docs".

### DIFF
--- a/packages/calcite-design-tokens/src/$config.ts
+++ b/packages/calcite-design-tokens/src/$config.ts
@@ -15,7 +15,7 @@ export const config: CalciteTokenTransformConfig = {
   },
   output: {
     dir: resolve(__dirname, "../dist"),
-    platforms: [Platform.SCSS, Platform.CSS, Platform.JS, Platform.JSON, Platform.ES6],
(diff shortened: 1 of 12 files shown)


benelan (commented):
This would change import paths for consumers right? If so it should be a breaking change

jcfranco:
> This would change import paths for consumers right? If so it should be a breaking change

The docs JSON output hasn't made it to a stable release yet, so I wouldn't consider it breaking from that perspective.

alisonailea:
Screenshot 2023-12-01 at 12 03 36 AM

jcfranco (approved):
Awesome! 🟦➡️🔵

Will `index.spec.ts` be updated to cover the docs JSON?

alisonailea:
> Awesome! 🟦➡️🔵
> 
> Will `index.spec.ts` be updated to cover the docs JSON?

I'm not currently watching the "docs" or "js" output in `index.spec.ts` because there is an absolute file path in each token at the moment which will break tests because the test is not running on my local machine. I'll make a ticket to fix this bug.

Issue #7364, opened 2023-07-24
Change to label's `for` prop not propagated


nwhittaker:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Given a `<calcite-label>` that does not contain an input element, if the label's `for` property is assigned/updated after the label element is initially rendered, then:
1. Clicking the label does not focus the input.
2. The input's `aria-label` is empty.

### Expected Behavior

Given a `<calcite-label>` that does not contain an input element:
1. Clicking the label always focuses the input assoc. via the label's `for` prop.
2. The input's `aria-label` always matches its label's text content.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/PoxdNJd

### Reproduction Steps

1. Visit the code sample
2. Use dev tools to verify the label's `for` prop is set to the ID of the input.
3. Verify that clicking the label does not focus the input and that the input's `<input>` element's `aria-label` is empty.

### Reproduction Version

1.4.3

### Relevant Info

macOS 13.4.1, Chrome 115

Related to, or perhaps a more generalized version of https://github.com/Esri/calcite-design-system/issues/6524. Unfortunately the workarounds in that issue do not work for keeping `aria-label` updated.

### Priority impact

p2 - want for current milestone

### Impact

This issue breaks accessibility for cases where the input's ID is inserted after the label component's initial render cycle. This usage can be typical when using a templating language like Ember to pass a dynamically generated ID.

The Field Apps team is planning on writing our own Stencil components that encapsulate many of Calcite's label and input -related components. A fix for this bug would cut down on the number of workarounds and/or broken accessibility issues for us.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Field Apps

geospatialem:
Research to determine feasibility for implementation, if using the label within the component or as part of its own standalone (`calcite-label`). cc @jcfranco

nwhittaker:
>@nwhittaker Can you confirm if this is critical for R3? If so, we can take a stab at it this week to get an idea on risk and go from there.

Not critical for R3. Post October is fine but we are beginning a larger refactor effort where it'd be nice to have this issue spiked/resolved sooner rather than later.

DitwanP:
Verified locally on Main

PR #8309, opened 2023-12-01
fix(label): associate label to component when `for` prop is set after initialization

jcfranco:
**Related Issue:** #7364 

## Summary

This allows `label` to find its corresponding labelable when `for` is dynamically set after the label and labelable have gone through the label matching process that happens when they are connected to the DOM.


driskull (approved):
Nice! I tried to think of a better name for the function `associateExplicitLabelToUnlabeledComponent` because its quite a mouthful but couldn't think of anything better.

PR #8317, opened 2023-12-01
fix(shell, shell-panel): support resizing shell panel when there is an iframe slotted in shell content

driskull:
**Related Issue:** #8156

## Summary

- When resizing starts & stops on a shell panel, emit an event to the shell.
- The shell will listen for that event and set pointer events none on the default slot container when resizing is active.
- This will prevent the iframe from capturing any pointer events and not bubbling them up which fixes the resize functionality


packages/calcite-components/src/components/shell-panel/shell-panel.tsx
```diff
@@ -216,6 +227,19 @@ export class ShellPanel implements ConditionalSlotComponent, LocalizedComponent,
...
+  //
+  //  Events
+  //
+  // --------------------------------------------------------------------------
+
+  /**
+   * @internal
+   */
+  @Event({ cancelable: false }) calciteInternalShellPanelResize: EventEmitter<{
+    resizing: boolean;
```
jcfranco, line 240: Nitpick: having a `resizing` boolean on a resize event sticks out to me. WDYT of `state: "start" | "end"`, `status: "start" | "end"` or something similar? I'll leave this up to you.

packages/calcite-components/src/components/shell/shell.tsx
```diff
@@ -38,6 +38,18 @@ export class Shell implements ConditionalSlotComponent {
...
 
+  //--------------------------------------------------------------------------
+  //
+  //  Event Listeners
+  //
+  //--------------------------------------------------------------------------
+
+  @Listen("calciteInternalShellPanelResize")
+  handleCalciteInternalShellPanelResize(event: CustomEvent<{ resizing: boolean }>): void {
+    this.panelIsResizing = !!event.detail?.resizing;
```
jcfranco, line 49: I think you can drop the optional chaining operator since `detail` is guaranteed for this custom event, right?

packages/calcite-components/src/components/shell/shell.scss
```diff
@@ -53,6 +57,10 @@
   display: initial;
 }
 
+.content--non-interactive {
+  @apply pointer-events-none;
```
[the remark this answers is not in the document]
driskull: I don't think so because this is getting applied within shadowRoot of shell.

packages/calcite-components/src/components/shell-panel/shell-panel.tsx
driskull: Yeah that's fine. Its internal so it doesn't really matter too much.

driskull: Ill just use two events so its clearer.

PR #8325, opened 2023-12-02
ci: remove workarounds no longer needed with Stencil v4

benelan:
## Summary

Removes some workarounds that are no longer needed. The first types test was because we didn't commit `components.d.ts` and needed to build twice to get the correct types. After updating to Stencil v4, we started getting JSDOM type errors from that test.

```
../../node_modules/puppeteer/lib/types.d.ts:7584:5 - error TS18028: Private identifiers are only available when targeting ECMAScript 2015 and higher.

7584     #private;
         ~~~~~~~~

Found 22 errors in the same file, starting at: ../../node_modules/puppeteer/lib/types.d.ts:33
```

It looks like it is using the wrong tsconfig or something, because we are above ES2015. Alternatively we could use `skipLibCheck`, but that can cause a variety of issues, especially with monorepos.

Also removes the extra build that happens before publishing CC. It was originally for the `component.d.ts` types issue and to make sure the version is correct in the source code preamble. Now that we don't have the types issue anymore, we can move the versioning before the first build in the CI to get rid of the second build.


alisonailea (approved):

jcfranco (approved):
✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨
✨🧹✨✨✨🧹✨✨🧹🧹✨✨🧹🧹🧹✨✨🧹🧹🧹✨🧹🧹🧹🧹✨🧹✨
✨🧹🧹✨✨🧹✨🧹✨✨🧹✨✨🧹✨✨🧹✨✨✨✨🧹✨✨✨✨🧹✨
✨🧹✨🧹✨🧹✨🧹✨✨🧹✨✨🧹✨✨🧹✨✨✨✨🧹🧹🧹✨✨🧹✨
✨🧹✨✨🧹🧹✨🧹✨✨🧹✨✨🧹✨✨🧹✨✨✨✨🧹✨✨✨✨✨✨
✨🧹✨✨✨🧹✨✨🧹🧹✨✨🧹🧹🧹✨✨🧹🧹🧹✨🧹🧹🧹🧹✨🧹✨
✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨

Issue #8343, opened 2023-12-06
Color context (mode) tokens should only be applied with prefers-color-scheme when `.calcite-mode-auto` is applied


alisonailea:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When you change the light or dark appearance on a Mac, the light or dark color context tokens are applied automatically.

### Expected Behavior

Color context tokens should only be applied in css with prefers-color-scheme when a `.calcite-mode-auto` class is applied.

### Reproduction Sample

https://codepen.io/benesri/pen/RwgrjEx

### Reproduction Steps

1. Open a calcite-component that is not wrapped in a `.calcite-mode-auto` class
2. switch to the "light" or "dark" mode on your computer
3. observe the calcite-component does not change it's colors

### Reproduction Version

2.0

### Priority impact

p0 - emergency

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [X] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

omarkawach:
This is greatly impacting us on the ArcGIS Maps SDK for JavaScript components team. Watching.

jmanke:
Dashboards is also seeing this in the authoring experience - dark mode will briefly flash before our theming logic kicks in (when system preference is set to dark).

omarkawach:
Some examples of areas where the design system is dictating the styling for us 

Arcade editor (action)
![image](https://github.com/Esri/calcite-design-system/assets/49499098/e4eb4ccf-ea65-45ae-9092-9a625c033d99)

Our Storybook web component documentation site (tabs, chips, and cards)
![image (1)](https://github.com/Esri/calcite-design-system/assets/49499098/1cf8c220-baa4-4b4a-96d8-eb1e0b93c18c)
![image (2)](https://github.com/Esri/calcite-design-system/assets/49499098/069a8264-ae5a-4479-8ffa-594f077c43cd)
![image (3)](https://github.com/Esri/calcite-design-system/assets/49499098/18dfa91c-ffbc-4d21-be35-cf589f440fdf)

geospatialem:
Verified in `1.12.0-next.7`  when the system settings are set to dark mode in Windows.

PR #8344, opened 2023-12-06
fix(color-context tokens): only apply when `.calcite-mode-auto` is applied

alisonailea:
**Related Issue:** #8343

## Summary

Resolves bug where the @media query, which applied light and dark color-context tokens on prefers-color-scheme in CSS, was applied at `:root` instead of `.calcite-mode-auto`

### DIFF
--- a/packages/calcite-design-tokens/support/token-transformer/styleDictionary/formatter/utils/formatExtraOutput.ts
+++ b/packages/calcite-design-tokens/support/token-transformer/styleDictionary/formatter/utils/formatExtraOutput.ts
@@ -36,14 +36,14 @@ export function formatExtraOutput(
             ? index.mixin.map(([mixinName, output]) => {
                 const m = ensureIfArray(outputObject[`${output}.${args.platform}`]);
                 return Array.isArray(m)
-                  ? `@mixin ${mixinName} {\n\t${m.map((o) => `${o}`.replace("$", "--")).join("\n\t")}\n}`
+                  ? `@mixin ${mixinName} {\n\t${m.map((o) => `${o}`.replaceAll("$", "--")).join("\n\t")}\n}`
                   : "";
               })
             : [];
           const medias = index.media
             ? index.media.map(([mediaSchemed, output]) => {
                 const m = ensureIfArray(outputObject[`${output}.${args.platform}`]);
-                const cssProps = m.map((o) => `${o}`.replace("$", "--"));
+                const cssProps = m.map((o) => `${o}`.replaceAll("$", "--"));
                 return Array.isArray(m)
                   ? `${
                       output === "light" ? `:root {\n\t${cssProps.join("\n\t")}\n}\n` : ""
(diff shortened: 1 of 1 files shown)


packages/calcite-design-tokens/support/token-transformer/styleDictionary/formatter/utils/formatExtraOutput.ts
```diff
@@ -44,7 +44,7 @@ export function formatExtraOutput(
             ? index.media.map(([mediaSchemed, output]) => {
                 const m = ensureIfArray(outputObject[`${output}.${args.platform}`]);
                 return Array.isArray(m)
-                  ? `@media (${mediaSchemed}) {\n\t:root {\n\t\t${m
+                  ? `@media (${mediaSchemed}) {\n\t.calcite-mode-auto {\n\t\t${m
```
benelan, line 47: As a follow up (not in this PR), we shouldn't be using `\n` since it can cause issues on Windows. Is there an issue for tracking the Windows related design token build errors that Kitty ran into? If not, can you please log one?

For reference, you can do this instead:

```js
import { EOL }from "os";

const osAgnosticString = `${EOL}lorem${EOL}ipsum${EOL}`
```

benelan (commented):
Did you do testing to confirm that it no longer uses dark mode if `prefer-color-scheme` is "dark"? If I remember correctly, `calcite-mode-auto` uses the value from `prefer-color-scheme`, so this change may end up with the same outcome.

I think we don't want to add any of the mode classes by default, which defaults to the components to "light".

ref: https://developers.arcgis.com/calcite-design-system/core-concepts/#modes

driskull (approved):
Nice 👍

alisonailea:
> Did you do testing to confirm that it no longer uses dark mode if `prefer-color-scheme` is "dark"? If I remember correctly, `calcite-mode-auto` uses the value from `prefer-color-scheme`, so this change may end up with the same outcome.

The e2e tests caught that light mode was not being applied by default.

benelan (approved):
Couple more nitpicks, but otherwise code LGTM! Thanks for the quick turn around! 

Just to confirm, the `.calcite-mode-auto` class isn't being added by default right?

---

It's fine for this PR, but in the future can we split the mode fix and the EOL refactor into two separate PRs? It makes it easier to determine what actually resolved the bug, as well as easier to revert one w/o losing the other's work if a regression occurs.


---

Stepping back a bit, would it make more sense to run the transformed strings/files through [Prettier's API](https://prettier.io/docs/en/api.html) instead of doing all the tab/newline formatting manually?

alisonailea:
> Just to confirm, the .calcite-mode-auto class isn't being added by default right?

No. It is provided as an option. Light mode tokens are set by default.

jcfranco (approved):
Noice! 🚀

alisonailea:
> It's fine for this PR, but in the future can we split the mode fix and the EOL refactor into two separate PRs? It makes it easier to determine what actually resolved the bug, as well as easier to revert one w/o losing the other's work if a regression occurs.

Good point. Refactoring the branch to remove EOL commits and moving those to a separate PR.

Issue #8350, opened 2023-12-06
'\n' in token output files causing issues in Windows


alisonailea:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

'\n' is written into platform output in tokens

### Expected Behavior

tokens transformer should use npm `os.EOL` or `Prettier.format()`

### Reproduction Sample

N/a

### Reproduction Steps

1. try to run the token transformer on a windows machine
2. observe the token transformer fails to build

### Reproduction Version

2.0

### Relevant Info

Originally brought up by Ben in https://github.com/Esri/calcite-design-system/pull/8344#pullrequestreview-1766629720

### Priority impact

p1 - need for current milestone

### Impact

Blocking usage of token output files on Windows machines

### Calcite package
-  @esri/calcite-design-tokens

### Esri team

Calcite (dev)

geospatialem:
Verified on the `main` branch with Windows and git bash with the demo pages.

PR #8352, opened 2023-12-06
fix: replace "\n" to support Windows for tokens output

alisonailea:
**Related Issue:** #8350

## Summary

Fixes an issue that was causing problems in Windows by handing off control of formatting to the experts and by experts I mean Prettier.


benelan (commented):
Awesome, this looks a lot cleaner! Has Kitty tested on this branch? We should make sure there aren't other Windows issues before we say it supports Windows.

jcfranco (approved):
Noice! ✨💅✨

geospatialem (commented):
Tested locally in a new clone on Windows. Perhaps unrelated to the `\n` use, but still running into issues with Windows WRT to the `--calcite-modal-scrim-background` variable. @alisonailea Do we need a new issue to track the effort?

image

image (1)

alisonailea:
> Tested locally in a new clone on Windows. Perhaps unrelated to the `\n` use, but still running into issues with Windows WRT to the `--calcite-modal-scrim-background` variable. @alisonailea Do we need a new issue to track the effort?
> 

Thanks for catching this. Weird that none of the other tests threw an error here.

geospatialem (approved):
🎉 🏆 🎉 

Windows is up and running again! Nice work, @alisonailea! 🥇

PR #8358, opened 2023-12-06
ci: stop formatting components.d.ts

benelan:
**Related Issue:** #7861 

## Summary

Make sure `component.d.ts` does not get formatted via `lint-staged` when committing. `lint-stage` runs from the root of the monorepo, so the file needs to be ignored there.


driskull (approved):
👍

Elijbet (approved):

PR #8381, opened 2023-12-08
build(node): bump node version to v20 (latest LTS)

benelan:
## Summary

- Bump Volta's pinned Node version to v20 (latest LTS).
- Make sure all of the GitHub Action workflows are using the Node version pinned by Volta.


driskull (approved):
👍

PR #8385, opened 2023-12-08
fix(button): don't needlessly overwrite title

maxpatiiuk:
In my application I set `title` attribute on the `<calcite-button>`, but it wasn't being displayed because the shadow DOM's `<button>` had an empty title attribute (`title=""`) which was overwriting the title set on the calcite-button.

Reproduction:
1. [Codepen example](https://codepen.io/maxpatiiuk/pen/KKJEaER?editors=1000)
2. Hover over the button - no tooltip is shown
3. Remove the "title" attribute from the `<button>`, and tooltip is visible

Video reproduction:

https://github.com/Esri/calcite-design-system/assets/40512816/d338be3e-667d-45ab-a29f-bc79d84df353

### DIFF
--- a/packages/calcite-components/src/components/tile-select/resources.ts
+++ b/packages/calcite-components/src/components/tile-select/resources.ts
@@ -11,6 +11,12 @@ export const CSS = {
   inputAlignmentStart: "input-alignment-start",
   inputEnabled: "input-enabled",
   largeVisual: "large-visual",
+  tile: "tile",
+  tileContentContainer: "tile-content-container",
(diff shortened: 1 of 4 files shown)


jcfranco (changes requested):
Thanks for submitting this PR, @maxpatiiuk!

Before proceeding, would you mind doing the following?

1. Creating an issue so this goes through our verification workflow.
2. Add an E2E test that provides coverage for this fix.

Let us know if you have any questions, thanks!

Issue #7445, opened 2023-08-03
Sortable list item's action menu don't close correctly after item is reordered


sonofflynn89:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

`<calcite-block>` items within a `<calcite-sortable-list>` have a built in actions menu popover. On initial render, the actions menu works fine. Clicking the toggle opens the popover. The popover can be closed by clicking an action, clicking outside, or clicking the toggle again.

https://github.com/Esri/calcite-design-system/assets/30124777/728416da-069c-4a0e-95ad-cd7e73b4acea

_However_, the moment that a `<calcite-block>` is reordered, the actions menu breaks. Clicking the toggle still opens the popover, but now the popover will not close after clicking an action or clicking outside. It will only close if the toggle is clicked

https://github.com/Esri/calcite-design-system/assets/30124777/7ba043d6-0de2-4f9c-9319-a568729d024e

### Expected Behavior

The actions menu popover should close when an action is clicked or the user clicks outside, just like it did on the initial render.

### Reproduction Sample

https://codepen.io/sonofflynn89-the-reactor/pen/qBQvEbj

### Reproduction Steps

The code pen has them enumerated

### Reproduction Version

1.5.0

### Priority impact

p3 - want for upcoming milestone

### Impact

While flows are not prohibited, it is jarring to see that after an action is clicked, the menu remains open. Our users could have up to 5 of these popovers stuck open at a time unless they know they know that they need to click the toggle button again.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Hub

geospatialem:
For next steps, Calcite should consider:
- Sorting on components, and 
- Creating a list of components for support, to ensure support is upheld and configurations are sound
  - Design/usability should be considered as part of the list considerations

cc @jcfranco @SkyeSeitz @ashetland

DitwanP:
Tested locally on `main` 

Results: 
- After sorting, popoverer closes when clicking the toggle.  ✅
- After sorting, popoverer closes when clicking outside.  ✅
- After sorting, popoverer closes when clicking the action.  ❌

https://github.com/Esri/calcite-design-system/assets/60022782/ae2bbbf4-f4da-447e-8b93-9930ec340ca5

driskull:
[...]

PR #8387, opened 2023-12-08
fix(action-menu): keep internal popover open property in sync

driskull:
**Related Issue:** #7445

## Summary

- Keep open property in sync with open property of internal popover
- Update menu button element when connectedCallback occurs.
- Add test


jcfranco (approved):
🚀👨‍🚀

Issue #8315, opened 2023-12-01
Tokens | Updates for docs output


alisonailea:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Currently when tokens are generated, each token has a key called "filePath" set by StyleDictionary which provides an absolute file path to the json file where the token is set. While useful this field will break snapshot tests as the snapshot is generated on a local machine while the tests are run in the Github CI.

This issue may be resolved by transforming the absolute file path to a path relative to the current working directory. Until this issue is resolved we can not have snapshots of the "js" or "docs" platform output.

Additionally the docs platform output should stringify all token values (needed by Gatsby) and need more metadata to make tokens easy to search and organize on the doc-site.

### Acceptance Criteria

- each token's `filePath` key is a string with a file path relative to the current working directory.
- docs platform output will stringify all token values
- Each core and semantic token gets a "extensions>calcite-schema" which defines all token naming schema meta data for use by documentation

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- @esri/calcite-design-tokens

### Esri team

Calcite (dev)

DitwanP:
🍫 Verified locally on `main`

`dist` folder includes `docs`

PR #8395, opened 2023-12-11
feat: update tokens for better documentation

alisonailea:
**Related Issue:** #8315

## Summary

This PR updates tokens to include a "calcite-schema" in each token's attributes. This schema fully defines the token and it's role in the design system. It also changes the Docs output to always stringify the token value and fix the filePath to be relative to the tokens directory.


jcfranco (approved):
LGTM! 🕹️👾

I have a couple of questions regarding the attributes/schema field:

* Are these automatically populated from Figma? If not, what is the process to determine the schema?
* To reduce duplication, would it possible to provide a common schema at the token parent level and have each token extend or override the parent schema?

alisonailea:
> * Are these automatically populated from Figma? If not, what is the process to determine the schema?

Not at this time. Automated Figma variables integration is still TBD. However, the schema organization was developed with the designers and the organization of Figma variables should mirror tokens. However I'm waiting for Skye to get back so we can do a final review of the schema before merging.

I'm also very open to feedback from devs as well. This naming schema will likely have long term effects on the organization of things. If needed I can move this into draft until the next engineering discussions.

> * To reduce duplication, would it possible to provide a common schema at the token parent level and have each token extend or override the parent schema?

I like this. It lends itself to part of a larger typescript refactor I was considering.

Issue #8139, opened 2023-11-07
[Tabs] Vertical scrollbar regression in Chromium browsers when housed inside a `panel`


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When `tabs` are housed inside a `panel` a scrollbar was introduced in `1.10.0`:

![image](https://github.com/Esri/calcite-design-system/assets/5023024/bc024f70-d2e2-4536-8f29-7a4dcd2ba2fc)

### Expected Behavior

No scrollbar present in Chromium browsers.

![image](https://github.com/Esri/calcite-design-system/assets/5023024/3e2eacc9-a123-4441-b394-cff02f9bef8f)

### Reproduction Sample

https://codepen.io/fcbasson/pen/rNPjPYM

### Reproduction Steps

1. Open the [sample](https://codepen.io/fcbasson/pen/rNPjPYM)
    - Also, can be observed in [this sample](https://codepen.io/geospatialem/pen/eYxgaqW)
2. Observe the scrollbar added into `tabs` in `1.10.0` 

### Reproduction Version

`1.10.0`

### Relevant Info

Effects Chromium browsers (Chrome, Edge on Windows). Firefox does not appear to be effected.

Presented in `1.10.0-next.11` with https://github.com/Esri/calcite-design-system/commit/e070d7db3f54c18c4586a558f3c0767fd1def3df when `tabs` are within a `panel`, could be related to the [`tab` section's `CSS.content` import](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/tab/tab.tsx#L72)?

### Regression?

1.10.0

### Priority impact

p4 - not time sensitive

### Impact

Visual regression.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

geospatialem:
This surfaced on [Community](https://community.esri.com/t5/calcite-design-system-questions/calcite-1-10-0-content-class-causing-overflow/m-p/1346551#M525) on 11/7/2023.

timmorey:
I'm seeing something that seems very similar, but my use case does not involve a `panel`, and it is reproducible across browsers (seeing in safari, firefox too).

[...]

PR #8399, opened 2023-12-11
fix(tab): prevent vertical scrollbar on content pane when the height of outer elements are specified

eriklharper:
**Related Issue:** #8139 

## Summary

This PR fixes an issue where applying a fixed CSS height on the `calcite-tabs` element or using tabs inside a shell panel caused child `tab`s to sometimes show a vertical scrollbar.

### DIFF
--- a/packages/calcite-components/src/components/tab/tab.scss
+++ b/packages/calcite-components/src/components/tab/tab.scss
@@ -13,6 +13,10 @@
   @apply block h-full w-full overflow-auto;
 }
 
+.content {
+  @apply box-border;
+}
+
 .scale-s .content {
   @apply text-n2h py-1;
 }
--- a/packages/calcite-components/src/components/tabs/tabs.stories.ts
+++ b/packages/calcite-components/src/components/tabs/tabs.stories.ts
@@ -378,3 +378,34 @@ export const updateIndicatorOffset_TestOnly = (): string => html`<calcite-tabs>
 updateIndicatorOffset_TestOnly.parameters = {
   chromatic: { delay: 1000 },
 };
+
+export const fixedHeightNoVerticalScrollbar_TestOnly = (): string => html`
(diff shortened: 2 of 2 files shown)


packages/calcite-components/src/components/tab/tab.scss
```diff
@@ -13,6 +13,10 @@
   @apply block h-full w-full overflow-auto;
 }
 
+.content {
+  @apply box-border;
```
jcfranco, line 17: Sidebar: @driskull Do you think we could split up https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/assets/styles/_host.scss for supporting/child components if they don't need background or other styles? Or if all components should be using `%component-host`, we can set up a rule to enforce it across all component style files.

jcfranco (approved):
📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑
📑📦📑📑📑📦📑📑📦📦📑📑📦📦📦📑📑📦📦📦📑📦📦📦📦📑📦📑
📑📦📦📑📑📦📑📦📑📑📦📑📑📦📑📑📦📑📑📑📑📦📑📑📑📑📦📑
📑📦📑📦📑📦📑📦📑📑📦📑📑📦📑📑📦📑📑📑📑📦📦📦📑📑📦📑
📑📦📑📑📦📦📑📦📑📑📦📑📑📦📑📑📦📑📑📑📑📦📑📑📑📑📑📑
📑📦📑📑📑📦📑📑📦📦📑📑📦📦📦📑📑📦📦📦📑📦📦📦📦📑📦📑
📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑📑

PR #8401, opened 2023-12-11
chore(dropdown-item): remove non-applicable selectors due to lack of visual `::before` pseudo-element

jcfranco:
**Related Issue:** N/A

## Summary

Existing selectors using `::before` are not targeting pseudo-elements due to missing `content`.

### DIFF
--- a/packages/calcite-components/src/components/dropdown-item/dropdown-item.scss
+++ b/packages/calcite-components/src/components/dropdown-item/dropdown-item.scss
@@ -121,39 +121,12 @@
   @apply bg-foreground-3;
 }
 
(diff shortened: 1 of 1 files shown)


driskull (approved):
This looks good as long as we can verify we don't need any of it. 👍

jcfranco:
> This looks good as long as we can verify we don't need any of it. 👍

Testing looks good so far and [this commit](https://github.com/Esri/calcite-design-system/commit/3457694b0664185c6f1229a316835eaaaa4ecd62) shows that `::before` was used for icons only and the remaining selectors are removed in this PR. 🎉

Issue #6321, opened 2023-01-20
[date-picker]: rounds up all remaining styling bugs


Elijbet:
### Actual Behavior

Rounds up all remaining styling bugs on date-picker:

- [x]  Hover starts selecting from the beginning of the month (https://github.com/Esri/calcite-components/issues/6194)
- [ ]  Unwanted blue gradient highlight when hovering/active/selected on the start-range date
- [ ]  Same behavior when the start-range date is selected and the end-range is being hovered/active/selected
- [ ]  Hovering/active past end-range also creates an unwanted white highlight over the grey gradient
- [ ]  All these issues are true and need to be addressed for both rtl and ltr
<img width="477" alt="1__213818983-9019a308-aaa1-4668-9573-21b1de5411a1" src="https://user-images.githubusercontent.com/3362490/218834211-31418a3c-4401-42ef-8243-d1b0cbfddaa4.png">
<img width="435" alt="2__213818984-ae875dcf-48a2-45fb-a92e-9463783166f8" src="https://user-images.githubusercontent.com/3362490/218834242-8fc80623-4cb3-416e-88f3-d38ac60815e5.png">
<img width="410" alt="3__213818980-5b3eb4ba-efb9-44a8-ad14-fe6c14c14956" src="https://user-images.githubusercontent.com/3362490/218834264-590d5e6f-4beb-4910-96c5-697464bd0ce7.png">

### Expected Behavior
Remove the highlight.

### Reproduction Sample
See demo.

### Reproduction Steps
N/A

### Reproduction Version
v1.0.0-beta.99

### Relevant Info
No response

### Regression?
No response

### Impact
No response

### Esri team
Calcite (dev)

geospatialem:
Need design expertise and input across the date-picker component.

SkyeSeitz:
Remaining listed bugs should be solved by design specs mentioned in this comment on the epic: https://github.com/Esri/calcite-design-system/issues/6721#issuecomment-1593774733

- [ ]  Unwanted blue gradient highlight when hovering/active/selected on the start-range date
- [ ]  Same behavior when the start-range date is selected and the end-range is being hovered/active/selected
- [ ]  Hovering/active past end-range also creates an unwanted white highlight over the grey gradient
- [ ]  All these issues are true and need to be addressed for both rtl and ltr

cc @geospatialem

jcfranco:
[...]

Issue #6410, opened 2023-02-01
Date picker: year is slightly cut off


jlias-devtester:
### Actual Behavior

The year in the date-picker component is slightly cut off on the end:

![image](https://user-images.githubusercontent.com/62298389/216129029-df56fa81-4043-4794-b426-c057ff8f04df.png)

### Expected Behavior

The year shouldn't be cut off

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/KKxEMoK

### Reproduction Steps

1. Visit the [sample](https://codepen.io/mac_and_cheese/pen/KKxEMoK) on Mac OS
2. Observe the year is cut off

### Reproduction Version
1.1.0

### Relevant Info
Seems to be specific to Mac OS across browsers (Chrome, FF, Safari). Windows browsers appear to be unaffected (tested on Chrome, FF, Edge).

### Esri team

ArcGIS Online

benelan:
Hi @jlias-devtester - did you close the issue because you figured out a solution? I'm not able to reproduce this on my end with this sample https://codepen.io/benesri/pen/qByQYyJ

If you are still experiencing the issue can you provide some additional information, such as browser, operating system, etc? A sample that reproduces the issue would also be very helpful. Thanks!

macandcheese:
It's reproducible in scale "L" : https://codepen.io/mac_and_cheese/pen/KKxEMoK?editors=1010

geospatialem:
> It's reproducible in scale "L" : https://codepen.io/mac_and_cheese/pen/KKxEMoK?editors=1010

Am able to reproduce the `scale="l"` bug in https://codepen.io/mac_and_cheese/pen/KKxEMoK on Mac OS across browsers (Chrome, FF, Safari), but Windows browsers are unaffected (tested on Chrome, FF, and Edge).

DitwanP:
🍡 Verified on `2.14.0-next.18` 
https://codepen.io/Ditwan-Price/pen/BaXYEoR

Issue #3455, opened 2021-11-09
Enhancement: add month-year and year appearance to calcite date picker


fendiaoxiaoshuazi:
### Description
Support month-year and year only appearance in the calcite date picker.

Here are the screenshots of the month-year and year only appearance in bootstrap date picker
![image](https://user-images.githubusercontent.com/26078106/140918561-ee99475e-689a-474f-9780-3346542b5645.png)
![image](https://user-images.githubusercontent.com/26078106/140918589-baeb0b95-a078-479b-a2c0-5925a3130f0d.png)

### Relevant Info 
Blocked issues: #9167, #8624, #10291, #10243, #10490

#### Which Component
`calcite-date-picker`

fendiaoxiaoshuazi:
Hi calcite team, I am a PE of the Survey123 team, and we are currently working on replacing our date and time picker with the calcite date and time picker for better A11y and I18N support. The month/year and year only appearance is critical for us, so we cannot fulfill the replacement at this point.

Do you have a timeframe time for this enhancement to be implemented?

fendiaoxiaoshuazi:
Hi Calcite team, we have partially replaced our pervious Wijimo date/time picker with the Calcite date/time picker. Since Calcite does not support month year and year only appearances, we have to keep the Wijimo date picker to support these appearances. 

Therefore, we need to load 2 different libs for the date/time question, and there are many month year and year only appearances I18N issues are impeded by the replacement work. 

We are waiting for Calcite to support month year and year only appearances to proceed with the replacement work.

fendiaoxiaoshuazi:
Hi Calcite team, we need to upgrade node.js to the latest version as required by the Security team, and we probably have to drop the Wijimo lib which we currently use for month-year and year-only appearances. 

Therefore, Calcite's month-year and year-only appearances support become the pre-requisite of our node upgrade. We plan to upgrade node.js and drop Wijimo in our UC Release. If possible could you please prioritize this issue, and address this issue before May?

benelan:
Hi @fendiaoxiaoshuazi, sorry for the delay! I'll make sure to bring it up again during our next triage. It is definitely a valid use case and something we would like to implement. However it is unlikely to be done by May, since it would require design and i18n feedback before development can start.

Which Node version are you currently on? If you're on an Active or Maintenance LTS version, updating to the latest minor/patch version should be sufficient security-wise. That should give us more time until your current date components need to be replaced.

[...]

PR #8402, opened 2023-12-12
feat(input-date-picker, date-picker): improve date picking experience

anveshmekala:
**Related Issue:** #3455, #10113 

## Summary

Update `calcite-date-picker` & `calcite-input-date-picker` UI & UX.

4D1CFC3C-8FF9-4493-9178-4DEDA0417031

### Key changes
- display two calendars for range irrespective of layout.
- No longer switches focus from day to end input when startDate is selected initially.
- Month selection is possible via select menu
- No longer positions the date-picker component relative to endInput when endInput is focused in range.
- Dates from previous months are not visible in range calendar.
- Divider indicator icons are removed in horizontal layout for range in input-date-picker.
- No longer uses chevron icon which indicate the open status of input-date-picker in startInput field.

Other issues resolved :
https://github.com/Esri/calcite-design-system/issues/6321
https://github.com/Esri/calcite-design-system/issues/6410
https://github.com/Esri/calcite-design-system/issues/10301
https://github.com/Esri/calcite-design-system/issues/10291
https://github.com/Esri/calcite-design-system/issues/10113
https://github.com/Esri/calcite-design-system/issues/10243
https://github.com/Esri/calcite-design-system/issues/10490

Blocked issues: #9167 

Wiki https://github.com/Esri/calcite-design-system/wiki/date%E2%80%90picker-enhancement-%238402


macandcheese:
Is the intention that these are separate public components? Or all housed within one public component?

I know the native input can be `type="month"`, but from my understanding, across other design systems, this approach seems (more?) common: https://mui.com/x/react-date-pickers/date-picker/#views

anveshmekala:
> Is the intention that these are separate public components? Or all housed within one public component?
> 
> I know the native input can be `type="month"`, but from my understanding, across other design systems, this approach seems (more?) common: https://mui.com/x/react-date-pickers/date-picker/#views

The idea is to have independent `input-month-picker` and `input-date-picker` components which makes it easier to maintain. Started off with `month-picker` though which can be used in `input-month-picker` and also integrated to current date-picker to allow user change month from a dropdown style.  

Concurrently, I am testing  a different approach to use combobox for input-month-picker and input-year-picker (similar to input-time-zone) which seems promising but wouldn't work if we need to display two scroll bars one for month and one for year.

I also dig the idea mentioned in the issue for input-month-picker which is more like a table with year at the top. It doesn't require two scroll bars.

macandcheese:
> Concurrently, I am testing a different approach to use combobox for input-month-picker and input-year-picker (similar to input-time-zone) which seems promising but wouldn't work if we need to display two scroll bars one for month and one for year.

IMO, this approach is just an implementation of a Combobox, a user can already set something like that up. I'd think ours would be more custom picking experiences like the examples linked in the original issue / Material example (however the components are structured).

anveshmekala:
_sidenote_: Suggestions for PR title are welcome

driskull:
Now that they are side-by-side, are we worried about the increased space being taken up? Are they responsive to the view width?

packages/calcite-components/src/components/date-picker/date-picker.tsx
```diff
@@ -124,6 +151,11 @@ export class DatePicker implements LocalizedComponent, LoadableComponent, T9nCom
     this.maxAsDate = dateFromISO(max);
   }
 
+  /**
+   * Specifies if the month abbreviations are used by the component.
+   */
+  @Prop() monthAbbreviations: boolean = false;
```
jcfranco, line 157: We might be able to use `Intl.DateTimeFormat` for this in the future, so WDYT about changing the prop to:

```tsx
monthStyle = "short" | "long"; // "narrow" could be added later
```

packages/calcite-components/src/components/date-picker/date-picker.tsx
```diff
@@ -361,17 +369,19 @@ export class DatePicker implements LocalizedComponent, LoadableComponent, T9nCom
     this.dateTimeFormat = getDateTimeFormat(this.effectiveLocale, DATE_PICKER_FORMAT_OPTIONS);
   }
 
-  monthHeaderSelectChange = (event: CustomEvent<Date>): void => {
-    const date = new Date(event.detail);
+  monthHeaderSelectChange = (event: CustomEvent<{ date: Date; position: string }>): void => {
```
jcfranco, line 372: Can you make this private? Applies to other updated functions.

packages/calcite-components/src/components/date-picker/date-picker.tsx
```diff
@@ -468,54 +496,48 @@ export class DatePicker implements LocalizedComponent, LoadableComponent, T9nCom
    * @param minDate
    * @param date
    * @param endDate
+   * @param position
```
jcfranco, line 499: There is no `position` argument.

packages/calcite-components/src/components/date-picker/date-picker.tsx
```diff
@@ -532,12 +554,13 @@ export class DatePicker implements LocalizedComponent, LoadableComponent, T9nCom
...
   private setEndDate(date: Date): void {
     const startDate = this.getStartDate();
-    const newEndDate = date ? setEndOfDay(date) : date;
+    this.userChangeRangeValue = true;
     this.value = [dateToISO(startDate), dateToISO(date)];
     this.valueAsDate = [startDate, date];
-    this.mostRecentRangeValue = newEndDate;
-    this.calciteDatePickerRangeChange.emit();
-    this.activeEndDate = date || null;
+    // avoid emitting change event when user select stat date beyond end date
```
jcfranco, line 560: Can we use a self-documenting var for this instead? Applies to similar comments.

jcfranco, line 560: Typo: `stat` ➡️ `start`

packages/calcite-components/src/components/date-picker-month/date-picker-month.tsx
```diff
@@ -62,6 +85,11 @@ export class DatePickerMonth {
   /** Specifies the latest allowed date (`"yyyy-mm-dd"`). */
   @Prop() max: Date;
 
+  /**
+   * When `true`, activates the component's range mode which renders two calendars for selecting ranges of dates.
+   */
+  @Prop({ reflect: true }) range: boolean = false;
```
jcfranco, line 91: Can you sort props alphabetically?

jcfranco:
@driskull brings up a great question. How will the updated design look/behave in narrow layouts? Would we need to opt out of the 2nd month picker for certain cases? [This example](https://codepen.io/jcfranco/pen/OJYjMbM?editors=100) shows some min widths for each range layout. cc @SkyeSeitz @ashetland

jcfranco (commented):
This is looking great, @anveshmekala! ✨📅✨

Props for adding extra test coverage and cleaning some tests up as well. 🏆

🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓
🗓🗓📅📅📅🗓🗓📅📅🗓🗓🗓📅📅🗓🗓📅🗓🗓🗓📅🗓
🗓📅🗓🗓🗓🗓📅🗓🗓📅🗓📅🗓🗓📅🗓📅🗓🗓🗓📅🗓
🗓📅🗓🗓🗓🗓📅🗓🗓📅🗓📅🗓🗓📅🗓📅🗓🗓🗓📅🗓
🗓📅🗓🗓🗓🗓📅🗓🗓📅🗓📅🗓🗓📅🗓📅🗓🗓🗓🗓🗓
🗓🗓📅📅📅🗓🗓📅📅🗓🗓🗓📅📅🗓🗓📅📅📅🗓📅🗓
🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓

jcfranco:
> sidenote: Suggestions for PR title are welcome

`feat(input-date-picker, date-picker): improve date picking experience`

?

SkyeSeitz:
> @driskull brings up a great question. How will the updated design look/behave in narrow layouts? Would we need to opt out of the 2nd month picker for certain cases? [This example](https://codepen.io/jcfranco/pen/OJYjMbM?editors=100) shows some min widths for each range layout. cc @SkyeSeitz @ashetland

I think the horizontal range can either wrap in narrow veiw ports or drop the second date picker. Wrapping might be the simpler option since going to one date picker would require a layout change to the header.

SkyeSeitz:
> > @driskull brings up a great question. How will the updated design look/behave in narrow layouts? Would we need to opt out of the 2nd month picker for certain cases? [This example](https://codepen.io/jcfranco/pen/OJYjMbM?editors=100) shows some min widths for each range layout. cc @SkyeSeitz @ashetland
> 
> I think the horizontal range can either wrap in narrow view ports or drop the second date picker. Wrapping might be the simpler option since going to one date picker would require a layout change to the header.

Thinking about the variety of use-cases, I think there are valid instances where range needs to be a single month and other times where a 2 month span is incredibly helpful. Would adding a month display prop to choose between a 1 or 2 month span be an option here? This would help with narrow viewports, but the 2 month span could still wrap regardless. 
image

alisonailea:
> I think there are valid instances where range needs to be a single month and other times where a 2 month span is incredibly helpful.

In the 2-month span case, would it make sense to convert the layout from horizontal to vertical at less than the smallest viewport?

SkyeSeitz:
> In the 2-month span case, would it make sense to convert the layout from horizontal to vertical at less than the smallest viewport?

Exactly. It would essentially wrap once the min-widths for each date picker are reached.

packages/calcite-components/src/components/input-date-picker/resources.ts
```diff
@@ -2,15 +2,21 @@ export const CSS = {
...
   calendarWrapper: "calendar-wrapper",
   calendarWrapperEnd: "calendar-wrapper--end",
+  container: "container",
+  dividerContainer: "divider-container",
+  divider: "divider",
   horizontalArrowContainer: "horizontal-arrow-container",
   inputBorderTopColorOne: "border-top-color-one",
   inputContainer: "input-container",
   inputNoBottomBorder: "no-bottom-border",
+  inputNoRightBorder: "no-right-border",
```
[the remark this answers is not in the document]
anveshmekala: Agree it would improve readability.

packages/calcite-components/src/components/input-date-picker/input-date-picker.scss
```diff
@@ -48,6 +48,11 @@
     .chevron-icon {
       color: var(--calcite-color-text-1);
     }
+    // chevron icon is available in endInput only for range.
+    // changes color of chevron icon in endInput when startInput is hovered/focused.
```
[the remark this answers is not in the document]
anveshmekala: I think this is straightforward. Don't remember why i added these comments in first place :sweat_smile:

packages/calcite-components/src/components/date-picker/date-picker.e2e.ts
```diff
@@ -551,20 +475,98 @@ describe("calcite-date-picker", () => {
...
     await page.waitForChanges();
     expect(await datePicker.getProperty("value")).toEqual(["2020-09-15", ""]);
 
-    await selectDay("20200930", page, "mouse");
+    await selectDayInMonthById("20200930", page);
     await page.waitForChanges();
     expect(await datePicker.getProperty("value")).toEqual(["2020-09-15", "2020-09-30"]);
   });
+
+  //todo: refactor to storybook tests
```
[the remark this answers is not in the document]
anveshmekala: Added https://github.com/Esri/calcite-design-system/issues/9570

packages/calcite-components/src/components/select/select.scss
```diff
@@ -79,11 +80,26 @@
 }
 
 .select {
-  @apply bg-foreground-1 border-color-input text-color-2 font-inherit focus-base m-0 box-border w-full cursor-pointer appearance-none truncate rounded-none border border-solid;
+  @apply bg-foreground-1
```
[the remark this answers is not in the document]
anveshmekala: Yusss!

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
```diff
@@ -124,66 +136,95 @@ export class DatePickerMonthHeader {
...
+        role="button"
+        scale={this.scale === "l" ? "l" : "m"}
+        text={direction === "right" ? this.messages.nextMonth : this.messages.prevMonth}
+      />
+    );
+  }
+
+  private getYearList(): void {
+    this.yearList = [];
+    for (let i = this.min?.getFullYear() || 1950; i <= (this.max?.getFullYear() || 2050); i++) {
```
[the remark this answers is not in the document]
anveshmekala: Currently, setting the default min value to 1900 and max  value to 2100 unless specified by the user.  We can minimize the interval to a shorter range and let the user specify min and max. Thoughts ? @SkyeSeitz

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
```diff
@@ -124,66 +136,95 @@ export class DatePickerMonthHeader {
...
+              {numberStringFormatter?.localize(yearString)}
+              {this.localeData?.year?.suffix}
+            </calcite-option>
+          );
+        })}
+      </calcite-select>
+    );
+  }
+
+  private renderChevron(direction: "left" | "right"): VNode {
```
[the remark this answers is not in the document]
anveshmekala: The reason behind using `right/left` is to keep that in sync with chevron icon names.

packages/calcite-components/src/components/input-date-picker/input-date-picker.e2e.ts
```diff
@@ -1281,4 +1491,58 @@ describe("calcite-input-date-picker", () => {
...
+    await page.waitForChanges();
+
+    expect(await calendar.isVisible()).toBe(true);
+    await page.keyboard.press("ArrowDown");
+    await page.waitForChanges();
+    expect(await getActiveMonth(page)).toBe("September");
+
+    await page.keyboard.press("ArrowDown");
+    await page.waitForChanges();
+    await page.waitForTimeout(2000);
```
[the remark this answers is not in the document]
anveshmekala: Removed these, added as part of debugging with `headful` mode.

packages/calcite-components/src/components/date-picker/date-picker.tsx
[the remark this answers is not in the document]
anveshmekala: Good catch. added issues https://github.com/Esri/calcite-design-system/issues/9579

anveshmekala:
Appreciate the feedback @macandcheese . Most of the design specs are updated regarding select menu width, spacing and chevron dimensions. 

Some of the items that might need discussion:

Hover state for select menu. Current we do not have a hover state styles for `calcite-select` and date-picker uses it internally. Do we want to add hover styles specific to date-picker case or is that something we would prefer adding for select component irrespective of the use case 

> We might want another visual state for a "day item you are about to select but is already part of a range" - it currently just appears grey - when in reality it is about to be selected - MUI has an additional "bordered" visual state for this.

Regarding above, i agree this can be misleading. In the current build while using keyboard, we preserve the highlighted color for the day and add a border to indicate the current focused date. Does this sound like a good option to opt for mouse users as well? 

Image 6-28-24 at 10 48 AM


>I think the grey "current day" should only apply to the current month + year. Currently it shows for the current "day number" in other months / years in range:

I think this will help the user to understand the active date in a given month. Without the indication, it is hard to guess atleast for the keyboard users which day will be focused when tabbing. Alternate could be focusing first day of the month without the visual grey representation for month-year which are non-current. Thoughts? 

cc @SkyeSeitz

I18nRobo:
@anveshmekala I have the en-CA data. Just a note, the existing data for months-abbreviated and days-abbreviated have been updated.

[ca-gregorian.json](https://github.com/user-attachments/files/16034524/ca-gregorian.json)

packages/calcite-components/src/components/date-picker/date-picker.tsx
anveshmekala: i think using monthStyle might give the idea to the user that number format is possible when `monthStyle="short"` (ex: 08 for Aug/August) as per UNICODE specs https://cldr.unicode.org/translation/date-time/date-time-patterns#h.aa5zjyepm6vh.  Additionally in context of calcite  these values might be confused for size of the month.

We can use "abbreviated | wide " and add "narrow" in future. Thoughts?

packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -1010,6 +1019,8 @@ export class InputDatePicker
     return !!(endValue && !startValue && this.focusedInput === "end" && this.startInput);
   }
 
+  //update these logics to allow focus restoration during editing
```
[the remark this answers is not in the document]
anveshmekala: nope, these needs to stay for keyboard users once selection is complete using "Enter".

packages/calcite-components/src/components/date-picker-day/date-picker-day.tsx
```diff
@@ -196,12 +196,10 @@ export class DatePickerDay implements InteractiveComponent, LoadableComponent {
         tabIndex={this.active && !this.disabled ? 0 : -1}
       >
         <InteractiveContainer disabled={this.disabled}>
-          <div aria-hidden="true" class={{ "day-v-wrapper": true }}>
```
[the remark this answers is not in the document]
geospatialem: In testing with JAWS this sounds great, and didn't observe any additional chatter on the date.  cc @anveshmekala 🏆

geospatialem (commented):
a11y-wise this is good to go - working great with JAWS and via keyboard. Nice work, @anveshmekala!! ✨

SkyeSeitz:
> Hover state for select menu. Current we do not have a hover state styles for calcite-select and date-picker uses it internally. Do we want to add hover styles specific to date-picker case or is that something we would prefer adding for select component irrespective of the use case

Since Select currently matches Combobox, I don't think I would recommend adding a hover state beyond the existing behavior to Select for now. It somewhat matches MUI's month picker with just a hover state on the chevron itself.


> Regarding above, i agree this can be misleading. In the current build while using keyboard, we preserve the highlighted color for the day and add a border to indicate the current focused date. Does this sound like a good option to opt for mouse users as well?

Agreed, this works great for keyboard usage and would improve mouse usage as well. 
image


> I think this will help the user to understand the active date in a given month. Without the indication, it is hard to guess atleast for the keyboard users which day will be focused when tabbing. Alternate could be focusing first day of the month without the visual grey representation for month-year which are non-current. Thoughts?

IMO current day should be specific to “today’s” month and year. Same day in another year or month should not display as current day. Focusing when current day is not in view, focusing first day of the month feels expected.

packages/calcite-components/src/components/select/select.scss
```diff
@@ -79,11 +61,27 @@
...
+    rounded-none
+    border-solid;
+  color: var(--calcite-select-text-color, var(--calcite-color-text-2));
+  border-width: var(--calcite-select-internal-border-width, 1px);
   font-size: var(--calcite-select-font-size);
   padding-inline: var(--calcite-select-spacing-inline);
+  padding-block: var(--calcite-internal-select-spacing-block);
   border-inline-end-width: theme("borderWidth.0");
-
+  font-weight: var(--calcite-select-font-weight, 400);
```
jcfranco, line 83: Can you use the corresponding core token props for the fallback values?

packages/calcite-components/src/components/date-picker-month/date-picker-month.tsx
```diff
@@ -487,30 +574,143 @@ export class DatePickerMonth {
...
+
+        <div class="week-days" role="row">
+          {days.map((day, index) => this.renderDateDay(day, isEndCalendar ? 50 + index : index))}
+        </div>
+      </div>
+    );
+  }
+
+  private monthHeaderSelectChange = (event: CustomEvent<Date>): void => {
+    const date = new Date(event.detail);
```
jcfranco, line 697: Can you stop propagation of this internal event? Applies to other internal event handlers.

packages/calcite-components/src/components/date-picker-month/date-picker-month.tsx
```diff
@@ -267,28 +315,35 @@ export class DatePickerMonth {
...
    * Add n days to the current date
    *
    * @param step
+   * @param targetDate
    */
-  private addDays(step = 0) {
-    const nextDate = new Date(this.activeDate);
-    nextDate.setDate(this.activeDate.getDate() + step);
-    this.calciteInternalDatePickerActiveDateChange.emit(
+  private addDays(step = 0, targetDate: Date) {
```
jcfranco, line 337: Can you add the return type to this? Applies to other updated functions.

packages/calcite-components/src/components/date-picker-month/date-picker-month.tsx
```diff
@@ -176,14 +250,19 @@ export class DatePickerMonth {
...
+    this.calciteInternalDatePickerMonthMouseOut.emit();
   }
 
   //--------------------------------------------------------------------------
   //
   //  Lifecycle
   //
   //--------------------------------------------------------------------------
+
+  connectedCallback(): void {
```
jcfranco, line 262: Is there another lifecycle hook you could use? It might be too early for this. This might be more of an issue with the `components` output build.

packages/calcite-components/src/components/date-picker-month/date-picker-month.tsx
```diff
@@ -119,43 +193,43 @@ export class DatePickerMonth {
     }
 
     const isRTL = this.el.dir === "rtl";
-
+    const dateValue = (event.target as HTMLCalciteDatePickerDayElement).value;
     switch (event.key) {
```
jcfranco, line 197: For readability, can you add a newline before the switch statement?

packages/calcite-components/src/components/date-picker-month/date-picker-month.tsx
```diff
@@ -86,26 +141,45 @@ export class DatePickerMonth {
...
+    position: string;
+  }>;
+
+  //--------------------------------------------------------------------------
+  //
+  //  Private State/Props
+  //
+  //--------------------------------------------------------------------------
+
+  /** current focused date */
```
jcfranco, line 181: Do we need this comment? It seems a bit redundant. Might apply to other comments.

packages/calcite-components/src/components/date-picker-month/date-picker-month.tsx
```diff
@@ -72,8 +92,43 @@ export class DatePickerMonth {
...
-  @Prop() hoverRange: HoverRange;
+  /** Specifies the latest allowed date (`"yyyy-mm-dd"`). */
+  @Prop() max: Date;
+
+  /**
+   * Made into a prop for testing purposes only
+   *
+   * @internal
+   */
+  // eslint-disable-next-line @stencil-community/strict-mutable -- updated by t9n module
```
jcfranco, line 103: Mutable and the disable line shouldn't be necessary since this is not implementing `T9nComponent`.

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
```diff
@@ -192,13 +290,27 @@ export class DatePickerMonthHeader {
...
 
-  @State() nextMonthDate: Date;
+  private yearPickerEl: HTMLCalciteSelectElement;
 
-  @State() prevMonthDate: Date;
+  private monthPickerEl: HTMLCalciteSelectElement;
+
+  private defaultMinYear = 1950;
+
+  private defaultMaxYear = 2050;
```
jcfranco, line 307: Could the min/max years be derived from the min/max ISO year vars?

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
```diff
@@ -124,66 +163,125 @@ export class DatePickerMonthHeader {
...
+        })}
+      </calcite-select>
+    );
+  }
+
+  private renderYearPicker(): VNode {
+    return (
+      <calcite-select
+        class={CSS.yearPicker}
+        label="Year menu"
```
jcfranco, line 225: This should use `messages`. Applies to other labels.

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
```diff
@@ -124,66 +163,125 @@ export class DatePickerMonthHeader {
...
+        onCalciteSelectChange={this.handleYearChange}
+        ref={(el) => (this.yearPickerEl = el)}
+        width="auto"
+      >
+        {this.yearList.map((year: number) => {
```
jcfranco, line 230: We may have discussed this previously, but could we explore using an input to select the year if there is no `min`/`max` provided? cc @SkyeSeitz

packages/calcite-components/src/components/date-picker/date-picker.tsx
jcfranco: > We can use "abbreviated | wide " and add "narrow" in future. Thoughts?

I dig it!

jcfranco (commented):
📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅
📅✨📅📅📅✨📅📅✨✨📅📅✨✨✨📅📅✨✨✨📅✨✨✨✨📅✨📅
📅✨✨📅📅✨📅✨📅📅✨📅📅✨📅📅✨📅📅📅📅✨📅📅📅📅✨📅
📅✨📅✨📅✨📅✨📅📅✨📅📅✨📅📅✨📅📅📅📅✨✨✨📅📅✨📅
📅✨📅📅✨✨📅✨📅📅✨📅📅✨📅📅✨📅📅📅📅✨📅📅📅📅📅📅
📅✨📅📅📅✨📅📅✨✨📅📅✨✨✨📅📅✨✨✨📅✨✨✨✨📅✨📅
📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅


This is looking great, @anveshmekala! ✨🥇✨

packages/calcite-components/src/components/date-picker/assets/date-picker/t9n/messages.json
```diff
@@ -2,6 +2,5 @@
   "nextMonth": "Next month",
   "prevMonth": "Previous month",
   "monthMenu": "Month menu",
-  "yearMenu": "Year menu",
-  "year": "Year"
```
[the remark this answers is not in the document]
anveshmekala: Good Catch!

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.e2e.ts
```diff
@@ -1,13 +1,6 @@
 import { newE2EPage } from "@stencil/core/testing";
-import { html } from "../../../support/formatting";
-import { renders } from "../../tests/commonTests";
 import { DateLocaleData } from "../date-picker/utils";
-
 describe("calcite-date-picker-month-header", () => {
-  describe("renders", () => {
```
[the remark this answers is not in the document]
anveshmekala: Since `setSelectMenuIconOffset` method invoked in `componentDidLoad`, `render` test util would return build error if component is rendered without `activeDate`.

packages/calcite-components/src/components/input-date-picker/input-date-picker.scss
```diff
@@ -59,25 +62,43 @@
...
+  background-color: var(--calcite-color-foreground-1);
+}
+
+:host([layout="horizontal"]) .divider-container {
+  @apply w-px;
+}
+
+.divider {
+  @apply my-1 inline-block w-px;
+  background-color: #d9d9d9;
```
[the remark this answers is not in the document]
anveshmekala: @SkyeSeitz

packages/calcite-components/src/components/input-date-picker/input-date-picker.stories.ts
```diff
@@ -234,6 +241,60 @@ export const widthSetToBreakpoints_TestOnly = (): string =>
...
+    const datePicker = document.querySelector("calcite-input-date-picker");
+    datePicker.valueAsDate = [new Date("2025-09-08"), new Date("2025-12-10")];
+  </script>
+`;
+
+rangeWithValueAsDate.parameters = {
+  chromatic: { delay: 2000 },
+};
+
+export const rangeWithValue = (): string => html`
```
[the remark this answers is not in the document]
anveshmekala: The test is added to make sure there is no visual difference in setting `value` or `valueAsDate`.

packages/calcite-components/src/components/input-date-picker/input-date-picker.scss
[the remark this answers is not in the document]
SkyeSeitz: Absolutely. This raw hex-code slipped through. Corrected it to `--calcite-color-border-2` (#d4d4d4) in the Figma specs. Good catch!

packages/calcite-components/src/components/input-date-picker/input-date-picker.stories.ts
```diff
@@ -234,6 +241,60 @@ export const widthSetToBreakpoints_TestOnly = (): string =>
...
+    const datePicker = document.querySelector("calcite-input-date-picker");
+    datePicker.value = ["2025-09-08", "2026-12-10"];
+  </script>
+`;
+
+rangeWithValue.parameters = {
+  chromatic: { delay: 2000 },
+};
+
+export const rangeWithMinBeforeCurrentDate = (): string => html`
```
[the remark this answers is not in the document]
anveshmekala: Prefer to keep them separate as they display different behavior in each case. 
When min is before currentDate, current day is viewed
When max is before currentDate, last valid day is  viewed since it is the closest day to the current day.

packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -261,6 +260,11 @@ export class InputDatePicker
     this.maxAsDate = dateFromISO(max);
   }
 
+  /**
+   * When `true`, month will be abbreviated.
+   */
+  @Prop() monthAbbreviations: boolean = false;
```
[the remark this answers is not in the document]
anveshmekala: `monthStyle` can be used with `abbreviated | wide` as values?

packages/calcite-components/src/components/action/action.scss
```diff
@@ -95,7 +95,8 @@
 
 :host([scale="s"]) {
   .button {
-    @apply text-n2h px-2 py-1 font-normal;
+    @apply text-n2h px-2  font-normal;
+    padding-block: var(--calcite-internal-action-padding-block, 0.25rem);
```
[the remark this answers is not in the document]
anveshmekala: In the context of the  PR, this prop is @internal and used by date-picker only.

packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -1063,14 +1067,25 @@ export class InputDatePicker
...
     }
 
-    const focusedInput = this.focusedInput === "start" ? this.startInput : this.endInput;
-    focusedInput.setFocus();
+    if (restore) {
+      this.focusInput();
+      return;
+    }
+
+    this.rangeStartValueChangedByUser = !restore && this.focusedInput === "start";
```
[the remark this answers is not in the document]
anveshmekala: Don't think it will be `true` always. When the user submit form or closes the date-picker with `escape` key !restore returns `false`

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.e2e.ts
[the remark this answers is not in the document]
jcfranco: You could use `newProgrammaticE2EPage` to set up the component before rendering (see [example](https://github.com/Esri/calcite-design-system/blob/bcbb79f8c925d505bb4ee5e6a54861c5f6bb88b9/packages/calcite-components/src/components/date-picker-day/date-picker-day.e2e.ts#L7-L23)). `disabled` and other common test helpers support this pattern.

packages/calcite-components/src/components/action/action.scss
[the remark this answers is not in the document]
jcfranco: Gotcha. We should bring this up as a potential new component prop.

packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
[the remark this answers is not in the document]
jcfranco: In the `if` statement above this line, we return when `restore === true`, so `restore` can't be `true` after that block, so these lines simplify to:

```tsx
this.rangeStartValueChangedByUser = this.focusedInput === "start";
this.focusedInput = false;
```

packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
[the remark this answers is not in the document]
jcfranco: I dig it!

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
anveshmekala: Good Catch! This needs to be updated to use messages as translations are back.

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
anveshmekala: pending discussion.

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
anveshmekala: pending Discussion.

packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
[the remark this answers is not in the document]
anveshmekala: I see it now :anguished: .

Requires a minute tweak. `this.focusedInput = false` isn't required.

packages/calcite-components/src/utils/date.ts
```diff
@@ -202,6 +202,31 @@ export function prevMonth(date: Date): Date {
   return nextDate;
 }
 
+/**
```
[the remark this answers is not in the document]
anveshmekala: Added.

packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -1063,14 +1067,25 @@ export class InputDatePicker
     this.restoreInputFocus();
   };
 
-  private restoreInputFocus(): void {
+  private restoreInputFocus(restore = false): void {
```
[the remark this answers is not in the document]
anveshmekala: How about `isDatePickerClosed = false` for `disableRestore`, when set to `true` we return and focus the input.

packages/calcite-components/src/components/date-picker/date-picker.stories.ts
```diff
@@ -83,6 +83,19 @@ export const rangeHighlighted_TestOnly = (): string => html`
   </script>
 `;
 
+export const rangeValuesNotInSameMonthAndYear_TestOnly = (): string => html`
+  <div style="width: 400px">
+    <calcite-date-picker range></calcite-date-picker>
+  </div>
+  <script>
+    (async () => {
+      await customElements.whenDefined("calcite-date-picker");
```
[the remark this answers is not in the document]
anveshmekala: specific to parsing `value/valueAsDate` in range or `<script/>` in general for stories?

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
anveshmekala: Moving forward with `input` for year.

packages/calcite-components/src/components/date-picker-month/date-picker-month.tsx
```diff
@@ -195,54 +273,23 @@ export class DatePickerMonth {
...
-      ...nextMonDays.map((day) => {
-        return {
-          active: false,
-          day,
-          dayInWeek: getDayInWeek(),
-          date: new Date(year, month + 1, day),
-        };
-      }),
-    ];
+    const endCalendarPrevMonDays = this.getPreviousMonthDays(month + 1, year, startOfWeek);
```
jcfranco, line 276: Suggestion: `month + 1` ➡️ `nextMonth` to DRY up a bit

packages/calcite-components/src/components/date-picker-month/date-picker-month.tsx
```diff
@@ -435,34 +490,38 @@ export class DatePickerMonth {
...
-      <div class="day" key={key} role="gridcell">
+      <div class="day-container" key={key} role="gridcell">
         <calcite-date-picker-day
           active={active}
           class={{
-            "hover--inside-range": this.startDate && isHoverInRange,
-            "hover--outside-range": this.startDate && !isHoverInRange,
-            "focused--start": isFocusedOnStart,
-            "focused--end": !isFocusedOnStart,
+            "current-day": currentDay,
```
jcfranco, line 509: Can we move these, and other inlined classes, to a `CSS` object? This can be done in a follow-up PR.

packages/calcite-components/src/components/date-picker-month/date-picker-month.tsx
```diff
@@ -487,30 +573,143 @@ export class DatePickerMonth {
...
+
+    const isDateBeforeStartDateAndAfterStart = date > start && date < this.startDate;
+    const isDateAfterEndDateAndBeforeEnd = date < end && date > this.endDate;
+    const isDateBeforeEndDateAndAfterEnd = date > end && date < this.endDate;
+    const isDateAfterStartDateAndBeforeStart = date < start && date > this.startDate;
+    const isDateAfterStartDateAndBeforeEnd = date < end && date > this.startDate;
+    const isDateBeforeEndDateAndAfterStart = date > start && date < this.endDate;
+
+    if (insideRange) {
+      if (!!this.startDate && !!this.endDate) {
```
jcfranco, line 606: `!!` might be redundant. Applies to similar usages in this block.

packages/calcite-components/src/components/date-picker-month/date-picker-month.tsx
```diff
@@ -487,30 +573,143 @@ export class DatePickerMonth {
...
+        <div class="week-header-container" role="row">
+          {weekDays.map((weekday) => (
+            <span class="week-header" role="columnheader">
+              {weekday}
+            </span>
+          ))}
+        </div>
+
+        <div class="week-days" role="row">
+          {days.map((day, index) => this.renderDateDay(day, isEndCalendar ? 50 + index : index))}
```
jcfranco, line 689: Can you store `50` in a local variable to avoid magic numbers?

packages/calcite-components/src/components/date-picker-month/date-picker-month.tsx
jcfranco: This is unresolved. Ignore if bubbling is intentional (maybe add a comment to make this clear).

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
```diff
@@ -70,9 +84,21 @@ export class DatePickerMonthHeader {
    * @internal
    * @readonly
    */
-  // eslint-disable-next-line @stencil-community/strict-mutable -- updated by t9n module
+  // eslint-disable-next-line @stencil-community/strict-mutable
   @Prop({ mutable: true }) messages: DatePickerMessages;
```
jcfranco, line 88: `mutable` and the `eslint-disable-next-line` comment are not needed since this doesn't implement `T9nComponent`. This might apply to other components.

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
```diff
@@ -119,68 +149,130 @@ export class DatePickerMonthHeader {
...
-        >
-          <calcite-icon flip-rtl icon={ICON.chevronRight} scale={getIconScale(this.scale)} />
-        </a>
       </Fragment>
     );
   }
 
+  private renderMonthYearContainer(reverse: boolean): VNode {
+    return reverse ? (
+      <Fragment>
```
jcfranco, line 182: Can we DRY this up?

```tsx
const content =  reverse ?  [/* ... */] : [/* ... */]; // or similar

return (
  <Fragment>{content}</Fragment>
);
```

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
```diff
@@ -119,68 +149,130 @@ export class DatePickerMonthHeader {
...
+    const isDisabled =
+      hasSameMonthAndYear(
+        isDirectionRight ? this.nextMonthDate : this.prevMonthDate,
+        this.activeDate,
+      ) || !inRange(this.activeDate, this.min, this.max);
+
+    return (
+      <calcite-action
+        alignment="center"
+        aria-disabled={`${isDisabled}`}
```
jcfranco, line 258: Can you use `toAriaBoolean` for ARIA boolean attrs? Applies to other ARIA booleans.

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
```diff
@@ -279,9 +340,33 @@ export class DatePickerMonthHeader {
...
+    if (!inRange(newDate, this.min, this.max)) {
+      newDate = dateFromRange(newDate, this.min, this.max);
+    }
+    this.calciteInternalDatePickerMonthHeaderSelect.emit(newDate);
+    this.setSelectMenuWidth(this.monthPickerEl);
+  };
+
+  private handleKeyDown = (event: KeyboardEvent): void => {
+    if (event.key === "ArrowDown" || event.key === "ArrowUp") {
+      event.stopPropagation();
```
jcfranco, line 368: FIXME: let's avoid stopping the propagation of native events.

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
```diff
@@ -323,16 +408,127 @@ export class DatePickerMonthHeader {
...
     }
 
     if (commit) {
-      yearInput.value = this.formatCalendarYear((inRangeDate || activeDate).getFullYear());
+      yearInputEl.value = this.formatCalendarYear((inRangeDate || activeDate).getFullYear());
+    }
+  }
+
+  private async setSelectMenuIconOffset(select: HTMLCalciteSelectElement): Promise<void> {
+    const iconEl = select.shadowRoot.querySelector("calcite-icon");
```
jcfranco, line 425: FIXME: let's avoid querying shadow roots.

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
```diff
@@ -323,16 +408,127 @@ export class DatePickerMonthHeader {
...
+      const prevMonthDate = dateFromRange(prevMonth(this.activeDate), this.min, this.max);
+      isTargetLastValidMonth = hasSameMonthAndYear(prevMonthDate, this.min);
+    } else if (this.max) {
+      const nextMonthDate = dateFromRange(nextMonth(this.activeDate), this.min, this.max);
+      isTargetLastValidMonth = hasSameMonthAndYear(nextMonthDate, this.max);
+    }
+
+    if (isTargetLastValidMonth) {
+      if (!this.position) {
+        direction === "left"
```
jcfranco, line 526: Can we DRY this up a bit?

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
```diff
@@ -323,16 +408,127 @@ export class DatePickerMonthHeader {
...
 
     if (commit) {
-      yearInput.value = this.formatCalendarYear((inRangeDate || activeDate).getFullYear());
+      yearInputEl.value = this.formatCalendarYear((inRangeDate || activeDate).getFullYear());
+    }
+  }
+
+  private async setSelectMenuIconOffset(select: HTMLCalciteSelectElement): Promise<void> {
+    const iconEl = select.shadowRoot.querySelector("calcite-icon");
+    await componentOnReady(iconEl);
```
jcfranco, line 426: Child components should be ready at this point (see https://stenciljs.com/docs/component-lifecycle#lifecycle-hierarchy). If not, there might be a watcher being explicitly invoked.

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
```diff
@@ -323,16 +408,127 @@ export class DatePickerMonthHeader {
...
+    const iconWidth = iconEl.getBoundingClientRect().width;
+    const iconContainerWidth = iconContainer.getBoundingClientRect().width;
+    this.selectMenuIconOffsetWidth = iconContainerWidth + (iconContainerWidth - iconWidth) / 2;
+    this.setSelectMenuWidth(select);
+  }
+
+  private setSelectMenuWidth(select: HTMLCalciteSelectElement): void {
+    const selectEl = select.shadowRoot.querySelector("select");
+    let selectedOptionWidth: number;
+    const fontStyle = getComputedStyle(selectEl).font;
```
jcfranco, line 437: Can the component's host be used to get the `font` property? The props listed [in the doc](https://github.com/Esri/calcite-design-system/blob/4c2c623d59441be8ddc66866748b7b104810b5ac/packages/calcite-components/src/utils/dom.ts#L125) are inherited.

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
```diff
@@ -323,16 +408,127 @@ export class DatePickerMonthHeader {
...
+      const activeMonthIndex = this.activeDate.getMonth();
+      selectedOptionWidth = getTextWidth(localeMonths[activeMonthIndex], fontStyle);
+    } else {
+      selectedOptionWidth = getTextWidth(
+        numberStringFormatter.localize(
+          `${parseCalendarYear(this.activeDate.getFullYear(), this.localeData)}`,
+        ),
+        fontStyle,
+      );
+      if (this.localeData?.year?.suffix) {
```
jcfranco, line 451: Isn't `localeData` guaranteed here?

jcfranco (changes requested):
Great job, @anveshmekala! There's a few FIXMEs, but this is very close! Excellent work on the E2E test coverage. ✨🏆✨

jcfranco (changes requested):
Great job, @anveshmekala! There's a few FIXMEs, but this is very close! Excellent work on the E2E test coverage. ✨🏆✨

packages/calcite-components/src/components/input-date-picker/input-date-picker.stories.ts
```diff
@@ -50,6 +50,11 @@ export default {
       control: { type: "select" },
     },
   },
+  parameters: {
+    chromatic: {
+      delay: 2000,
```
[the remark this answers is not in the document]
anveshmekala: Was debugging with diff set of delay values to make sure date-picker is opened when screenshot is captured. 
Surprisingly, wrapping a container with `height` and no custom delay solved the issue. 

WIN-WIN! ⭐

packages/calcite-components/src/components/date-picker/assets/date-picker/t9n/messages.json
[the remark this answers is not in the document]
anveshmekala: This is no longer required since we are using `input` for year.  Am i missing anything?

packages/calcite-components/src/components/input-date-picker/input-date-picker.e2e.ts
```diff
@@ -1399,4 +1550,301 @@ describe("calcite-input-date-picker", () => {
...
+async function selectDayInMonthByIndex(page: E2EPage, day: number): Promise<void> {
+  const dayIndex = day - 1;
+  const days = await page.findAll("calcite-input-date-picker >>> calcite-date-picker-day[current-month]");
+  await days[dayIndex].click();
+  await page.waitForChanges();
+}
+
+async function getActiveMonth(page: E2EPage, position: Extract<"start" | "end", Position> = "start"): Promise<string> {
+  const [startMonth, endMonth] = await page.findAll(
+    "calcite-input-date-picker >>> calcite-date-picker-month-header >>> .header >>> calcite-select.month-select",
```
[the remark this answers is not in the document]
anveshmekala: CSS object of input-date-picker do not have `header`  Is it worth adding it for testing purposes only?

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
anveshmekala: `mutable` is required for testing purposes.

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
anveshmekala: Didnt' knew we could use array in JSX. Very Cool & Clean.

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
anveshmekala: After taking a closer look at this, figured currently font in select is set via CSS variable `--calcite-select-font-size`. And the prop is used in select element rather than `:host`. 

We could add `inherit` [here](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/components/select/select.scss#L83)  & reference the font css prop at `:host`. Don't think it was intentional to reference CSS prop in native element [here](https://github.com/Esri/calcite-design-system/pull/1112/files#diff-6292969ff9b2667a39f133113a074309c3d39badaf733e69a7e14b691af447e4R44).

Thoughts?

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.e2e.ts
```diff
@@ -38,62 +34,36 @@ describe("calcite-date-picker-month-header", () => {
...
-
-      document.body.innerHTML = "";
-      document.body.append(dateMonthHeader);
-    }, localeDataFixture);
-    await page.waitForChanges();
-    const date = await page.find(`calcite-date-picker-month-header >>> input`);
-
-    expect(await date.getAttribute("aria-label")).toBe("Year");
+    const yearInput = await page.find(`calcite-date-picker-month-header >>> input`);
+    expect(yearInput.getAttribute("aria-label")).toBe("Year");
```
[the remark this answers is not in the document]
anveshmekala: message bundle is required for each test.  Fetching message bundle is added in `beforeEach` script.

packages/calcite-components/src/components/date-picker-month/date-picker-month.tsx
anveshmekala: Bubbling isn't required. Event propagation is stopped.

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
anveshmekala: This is intentional to avoid closing `date-picker` in `input-date-picker` on ArrowKeys.

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.scss
```diff
@@ -63,72 +80,109 @@
...
+  }
+}
+
+:host([scale="l"]) .month-year-container {
+  .month-select {
+    --calcite-select-spacing-inline: var(--calcite-size-sm);
+    --calcite-internal-select-spacing-block: var(--calcite-size-sm);
+    --calcite-internal-select-icon-container-padding-inline: var(--calcite-size-sm);
+    --calcite-internal-select-block-size: #{$calcite-size-44};
+    --calcite-select-font-size: var(--calcite-font-size-lg);
```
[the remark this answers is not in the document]
anveshmekala: make sense. Public props are documented currently and internal props might need review. Will add a comment in https://github.com/Esri/calcite-design-system/issues/7180.

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
anveshmekala: Yup, the `value` Watcher in `select` is invoked .

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
anveshmekala: This is required to set dynamic width for month select.

packages/calcite-components/src/components/action/action.scss
```diff
@@ -148,7 +148,8 @@
 
 :host([scale="s"]) {
   .button {
-    @apply text-n2h px-2 py-1 font-normal;
+    @apply text-n2h px-2 font-normal;
+    padding-block: var(--calcite-internal-action-padding-block, var(--calcite-size-xxs));
```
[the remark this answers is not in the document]
anveshmekala: `action` tokens are installed i believe and tokens added in this PR are `internal`. Adding @alisonailea for review.

 Will add a comment in the issue for `select` & `date-picker` related.

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
jcfranco: We cancel the event for these cases. `input-date-picker` will [ignore canceled arrow key presses](https://github.com/Esri/calcite-design-system/blob/anveshmekala/3455-add-month-and-year-picker/packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx#L943-L945).

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
jcfranco: Could we get the icon container dimensions from the design spec and tokens? If not, I’m inclined to hardcode the sizes (based on the spec) while we explore alternatives in a follow-up issue.

jcfranco (approved):
Aside from a few comments, I think this is good to go! Awesome job, @anveshmekala! 😎

📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅
📅👈📅📅📅👈📅📅👈👈📅📅👈👈👈📅📅👈👈👈📅👈👈👈👈📅👈📅
📅👈👈📅📅👈📅👈📅📅👈📅📅👈📅📅👈📅📅📅📅👈📅📅📅📅👈📅
📅👈📅👈📅👈📅👈📅📅👈📅📅👈📅📅👈📅📅📅📅👈👈👈📅📅👈📅
📅👈📅📅👈👈📅👈📅📅👈📅📅👈📅📅👈📅📅📅📅👈📅📅📅📅📅📅
📅👈📅📅📅👈📅📅👈👈📅📅👈👈👈📅📅👈👈👈📅👈👈👈👈📅👈📅
📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅📅

packages/calcite-components/src/components/input-date-picker/input-date-picker.stories.ts
```diff
@@ -53,23 +53,47 @@ export default {
 };
 
 export const simple = (args: InputDatePickerStoryArgs): string => html`
-  <div style="width: 400px">
+  <style>
```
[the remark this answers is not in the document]
anveshmekala: Will create a followup PR.

packages/calcite-components/src/components/input-date-picker/input-date-picker.e2e.ts
```diff
@@ -1399,4 +1549,323 @@ describe("calcite-input-date-picker", () => {
...
+    await page.keyboard.press("ArrowRight");
+    await page.waitForChanges();
+    await page.keyboard.press("ArrowRight");
+    await page.waitForChanges();
+    expect(await getDayById(page, "20240703")).not.toHaveAttribute("range-hover");
+
+    await page.keyboard.press("Enter");
+    await page.waitForChanges();
+    expect(await calendar.isVisible()).toBe(false);
+    expect((await inputDatePicker.getProperty("value"))[1]).toEqual("2024-07-02");
```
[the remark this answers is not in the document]
anveshmekala: Is the below syntax suggested?

``` 
    const [, endValue] = await inputDatePicker.getProperty("value");
    expect(endValue).toEqual("2024-07-02");

File: packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx (hunk 44 above)
Comment (anveshmekala) [reply packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx:460]: Initial idea is to use hardcoded values but it can be brittle if the layout of select or sizing of icon changes. Prefer using tokens but they aren't defined for the `inline-size` & `block-size` of icon yet. 

Looking at it now, i realized we can get rid of all the querySelectors since we use only `scale="m"` for `select` in `date-picker-month-header` & `icon` size is fixed. Additionally, `tokens` for spacing can be used to calibrate offset width.

File: packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx (hunk 40 above)
Comment (anveshmekala) [reply packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx:88]: 🧠  fade!
 
 removed the `mutable ` from messages.

**Review by SkyeSeitz — APPROVED** [review]

File: packages/calcite-components/src/components/date-picker/date-picker.scss [hunk 56]
```diff
@@ -7,25 +7,47 @@
     inline-block
     overflow-visible
     rounded-none
-    w-auto;
+    w-full;
+}
+
+.month-header {
```
anveshmekala, line 13: Not required,will be removed.

Issue #8135, opened 2023-11-06
[Menu Item] Allow keyboard navigation to open `href` link


dmbrinkman:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Using a keyboard to open a link to a menu item does not open the link specified in `href`.

To reproduce:  See code pen example and repro steps

### Expected Behavior

Keyboard navigation opens the linked page.

Notes from Adam Tirella:
I think we just need to add case for when href is populated, do not prevent default on the rendered internal anchor for enter / space.

### Reproduction Sample

https://codepen.io/dmbrinkman/pen/qBgRPxp?editors=1000

### Reproduction Steps

1.  Run the codepen example
2. Click on Link using mouse
3. Page opens in a new tab
4. Use the keyboard `Enter` or `Space bar` to select `Link`
5. Page does not open

### Reproduction Version

1.10.0

### Priority impact

p2 - want for current milestone

### Calcite package

- [X] @esri/calcite-components
- [X] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

macandcheese:
@geospatialem could you confirm the expected a11y behavior here? Both `space` and `enter` key should Navigate to the `href` if populated?

geospatialem:
> @geospatialem could you confirm the expected a11y behavior here? Both `space` and `enter` key should Navigate to the `href` if populated?

@macandcheese You're spot on - we'd want to support both `space` and `enter` - it looks like the [`a` element and `role` are setup already](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/menu-item/menu-item.tsx#L473-L488), so we'd just need to support the same functionality with the keys that's provided with a click event.

macandcheese:
FYI for verification - After speaking with @geospatialem - this should match other elements that conditionally render button / anchor (button, link, etc.) and only the `enter` key should open the link, not `space`.

DitwanP:
🍰 Verified locally on `2.1.0-next.10`

PR #8408, opened 2023-12-12
fix(menu-item): Improve keyboard navigability when `href` populated

macandcheese:
**Related Issue:** #8135


driskull (approved):
👍 

Any way to test this?

packages/calcite-components/src/components/menu-item/menu-item.e2e.ts
```diff
@@ -50,6 +50,39 @@ describe("calcite-menu-item", () => {
...
+      await page.setContent(html`${testEl}`);
+      const originalUrl = page.url();
+      await page.waitForChanges();
+
+      const menuItem = await page.find("calcite-menu-item");
+      await page.waitForChanges();
+      await menuItem.click();
+      await page.waitForChanges();
+      const newUrl = page.url();
+      expect(newUrl).toEqual(originalUrl + testHref);
```
macandcheese, line 68: @driskull @jcfranco any other ideas for how to test this? All I could come up with is checking an for a named anchor with `#`, any `href` that changed the url, I wasn't having luck testing against.

packages/calcite-components/src/components/menu-item/menu-item.tsx
```diff
@@ -266,61 +266,54 @@ export class CalciteMenuItem implements LoadableComponent, T9nComponent, Localiz
...
-    if (event.key === " " || event.key === "Enter") {
-      this.selectMenuItem(event);
-      if (
-        this.hasSubmenu &&
-        (!this.href || (this.href && event.target === this.dropdownActionEl))
-      ) {
-        this.open = !this.open;
+    const { hasSubmenu, href, layout, open, submenuItems } = this;
+    const key = event.key;
+    const targetIsDropdown = event.target === this.dropdownActionEl;
```
macandcheese, line 271: @anveshmekala this look ok to you? I think at some point we might want to break the dropdown key / click and the anchor key / click apart so we can simplify some of this logic.

packages/calcite-components/src/components/menu-item/menu-item.tsx
```diff
@@ -266,61 +266,54 @@ export class CalciteMenuItem implements LoadableComponent, T9nComponent, Localiz
...
+    if (key === " " || key === "Enter") {
+      if (hasSubmenu && (!href || (href && targetIsDropdown))) {
```
jcfranco, line 273: I think 

```ts
!href || (href && targetIsDropdown)
```
can be simplified as 

```ts
!href || targetIsDropdown
```

packages/calcite-components/src/components/menu-item/menu-item.tsx
```diff
@@ -266,61 +266,54 @@ export class CalciteMenuItem implements LoadableComponent, T9nComponent, Localiz
...
-      this.selectMenuItem(event);
-      if (
-        this.hasSubmenu &&
-        (!this.href || (this.href && event.target === this.dropdownActionEl))
-      ) {
-        this.open = !this.open;
+    const { hasSubmenu, href, layout, open, submenuItems } = this;
+    const key = event.key;
+    const targetIsDropdown = event.target === this.dropdownActionEl;
+    if (key === " " || key === "Enter") {
```
jcfranco, line 272: We have a util to check for this condition: `isActivationKey`. This is optional since there’s other individual space and Enter key checks going on.

packages/calcite-components/src/components/menu-item/menu-item.tsx
```diff
@@ -266,61 +266,54 @@ export class CalciteMenuItem implements LoadableComponent, T9nComponent, Localiz
...
-    } else if (event.key === "ArrowDown" || event.key === "ArrowUp") {
+    } else if (key === "ArrowDown" || key === "ArrowUp") {
       event.preventDefault();
-      if (
-        (event.target === this.dropdownActionEl || !this.href) &&
-        this.hasSubmenu &&
-        !this.open &&
-        this.layout === "horizontal"
-      ) {
+      if ((targetIsDropdown || !href) && hasSubmenu && !open && layout === "horizontal") {
```
jcfranco, line 291: Can we store `(targetIsDropdown || !href) && hasSubmenu && !open` in a local var and reuse?

packages/calcite-components/src/components/menu-item/menu-item.e2e.ts
jcfranco: I think this works. I don’t think we can use Puppeteer’s `waitForNavigation` method, but this seems like a good alternative. If you were able to make the test fail when expected (e.g. removing/bypassing  the fix), this should be good to go. 🚀

jcfranco (approved):
Noice! 👉🏼🔗

packages/calcite-components/src/components/menu-item/menu-item.tsx
macandcheese: Added

packages/calcite-components/src/components/menu-item/menu-item.tsx
macandcheese: Refactored this section to store all re-used vars / checks.

Issue #8386, opened 2023-12-08
calcite-action-bar: tooltip causing wrong width for action bar


AdelheidF:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Before and after look of right action bar when adding calcite-blocks in flow-item panel.
![image](https://github.com/Esri/calcite-design-system/assets/44481292/7abf1e9a-0627-423c-a5ea-7e9631d94f83)  ![image](https://github.com/Esri/calcite-design-system/assets/44481292/f7c9e090-28ba-4db2-b06d-3c75b4fd994e)

Contact me for a test case.

cc @driskull 

### Expected Behavior

Action bar width doesn't change.

### Reproduction Sample

N/A

### Reproduction Steps

Open MV and notice 2 action bars, right and left, in the viewer. Click on popup config action to open that panel. Then click 'add content' and 'Fields list'. Notice that the action bar on the left disappears and there is blank space at the right of the MV.
If you hover over the expand icon of the action bar it corrects itself.

### Reproduction Version

2.0.0

### Regression?

Yes

### Priority impact

p0 - emergency

### Impact

Very visible in MV

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Map Viewer

driskull:
@AdelheidF can you try creating a reproduction of this to help with verification?

AdelheidF:
Here is an attempt, but adding a block or re-ordering blocks doesn't cause it in this codepen.
https://codepen.io/afreitag/pen/JjxVMXQ

driskull:
@kevindoshier could this also be caused by the MV not using version 2.0 at the time?

AdelheidF:
not fixed with 2.1.0-next.8

driskull:
I'll take another look. 👀

driskull:
I can't reproduce this one. ok to close @AdelheidF?

AdelheidF:
We installed a workaround using `overlayPositioning="fixed"` on `calcite-action-group` and `calcite-tooltip`. With that we don't see the issue anymore.

We can re-open if someone else runs into it.

PR #8001, opened 2023-10-12
fix: floating components will now get an initial position even if they are not opened

driskull:
**Related Issue:** #7979

## Summary

- Removes floating-ui logic that requires a component to be open before any positioning occurs
- Updates tests
  - commonTest
  - floatingUI spec test
- Add jsdoc

PR #8409, opened 2023-12-12
fix(floating-ui): improve floating element performance

driskull:
**Related Issue:** #7979 #8214 #8386 #8419 #5697

## Summary

- reverts #8001
  - Keeps doc updates 
- reverts partially #8230
  - Keeps drag classes that are setting `overflow:hidden` when a drag is occurring.
  - Keeps roundByDPR floating-ui fix. https://floating-ui.com/docs/misc#subpixel-and-accelerated-positioning
- Hides overflow for `calcite-sortable--chosen` class
- Only sets `top` and `left` on floating element once positioned. **This fixes native drag and drop issue**

### DIFF
--- a/packages/calcite-components/src/components/dropdown/dropdown.tsx
+++ b/packages/calcite-components/src/components/dropdown/dropdown.tsx
@@ -78,13 +78,14 @@ export class Dropdown
 
   @Watch("open")
   openHandler(): void {
+    onToggleOpenCloseComponent(this);
+
     if (this.disabled) {
       this.open = false;
       return;
     }
 
     this.reposition(true);
-    onToggleOpenCloseComponent(this);
   }
 
   /**
--- a/packages/calcite-components/src/components/dropdown/dropdown.tsx
+++ b/packages/calcite-components/src/components/dropdown/dropdown.tsx
@@ -78,12 +78,13 @@ export class Dropdown
 
   @Watch("open")
   openHandler(): void {
-    if (!this.disabled) {
-      onToggleOpenCloseComponent(this);
+    if (this.disabled) {
+      this.open = false;
       return;
     }
 
-    this.open = false;
+    this.reposition(true);
+    onToggleOpenCloseComponent(this);
   }
 
   /**
@@ -549,7 +550,6 @@ export class Dropdown
   };
 
   onBeforeOpen(): void {
-    this.reposition(true);
     this.calciteDropdownBeforeOpen.emit();
   }
 
@@ -563,7 +563,6 @@ export class Dropdown
 
   onClose(): void {
     this.calciteDropdownClose.emit();
-    this.reposition(true);
   }
 
   setReferenceEl = (el: HTMLDivElement): void => {
--- a/packages/calcite-components/src/utils/floating-ui.ts
+++ b/packages/calcite-components/src/utils/floating-ui.ts
@@ -157,6 +157,9 @@ export const positionFloatingUI =
       pointerEvents,
       position,
       transform: open ? `translate(${roundByDPR(x)}px,${roundByDPR(y)}px)` : "",
+      width: "max-content",
+      top: "0",
+      left: "0",
     });
   };
 
@@ -491,9 +494,6 @@ export function connectFloatingUI(
 
     // initial positioning based on https://floating-ui.com/docs/computePosition#initial-layout
     position: component.overlayPositioning,
-    width: "max-content",
-    top: "0",
-    left: "0",
   });
 
   const runAutoUpdate = Build.isBrowser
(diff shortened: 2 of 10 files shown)


packages/calcite-components/src/utils/floating-ui.ts
```diff
@@ -489,8 +494,6 @@ export function connectFloatingUI(
 
     // initial positioning based on https://floating-ui.com/docs/computePosition#initial-layout
     position: component.overlayPositioning,
-    top: "0",
```
driskull, line 492: @jcfranco do you know why this was here?

It was added here: https://github.com/Esri/calcite-design-system/commit/e220686fa366b86dda2647da25144eb45a4a193d#diff-3c419c81de1cc684c4d1d859881782bc87d089da410675b5af3f6c468b95ef13R458

I'd like to remove this to fix the drag and drop positioning issue. #7979

driskull: Instead, this is only set once an actual position occurs.

jcfranco (approved):
🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎
🏎🎈🎈🎈🎈🏎🎈🏎🏎🏎🏎🏎🎈🏎🎈🎈🎈🎈🏎🏎🎈🎈🎈🏎🏎🎈🎈🏎🏎🎈🏎🏎🏎🎈🏎🎈🎈🎈🎈🏎
🏎🎈🏎🏎🎈🏎🎈🏎🏎🏎🏎🏎🎈🏎🎈🏎🏎🏎🏎🎈🏎🏎🏎🏎🎈🏎🏎🎈🏎🎈🎈🏎🎈🎈🏎🎈🏎🏎🏎🏎
🏎🎈🎈🎈🎈🏎🎈🏎🏎🎈🏎🏎🎈🏎🎈🎈🎈🏎🏎🏎🎈🎈🏎🏎🎈🏎🏎🎈🏎🎈🏎🎈🏎🎈🏎🎈🎈🎈🏎🏎
🏎🎈🏎🏎🎈🏎🎈🏎🎈🏎🎈🏎🎈🏎🎈🏎🏎🏎🏎🏎🏎🏎🎈🏎🎈🏎🏎🎈🏎🎈🏎🏎🏎🎈🏎🎈🏎🏎🏎🏎
🏎🎈🏎🏎🎈🏎🏎🎈🏎🏎🏎🎈🏎🏎🎈🎈🎈🎈🏎🎈🎈🎈🏎🏎🏎🎈🎈🏎🏎🎈🏎🏎🏎🎈🏎🎈🎈🎈🎈🏎
🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎🏎

Do we have a perf impact issue we could link to for this PR?

driskull:
> Do we have a perf impact issue we could link to for this PR?

No, can we get one? cc @paulcpederson

packages/calcite-components/src/components/dropdown/dropdown.tsx
```diff
@@ -78,12 +78,14 @@ export class Dropdown
...
-      onToggleOpenCloseComponent(this);
+    onToggleOpenCloseComponent(this);
+
+    if (this.disabled) {
+      this.open = false;
       return;
     }
 
-    this.open = false;
+    this.reposition(true);
```
jcfranco, line 88: Great catch on this perf fix, btw!

packages/calcite-components/src/utils/floating-ui.ts
driskull: Looks like that doc says only the `position` needs to be there initially so the top and left can happen later.

jcfranco:
@driskull Could you also check if this introduces a regression for https://github.com/Esri/calcite-design-system/issues/5697? 🙇‍♂️

paulcpederson:
https://github.com/Esri/calcite-design-system/issues/8419

driskull:
>@driskull Could you also check if this introduces a regression for https://github.com/Esri/calcite-design-system/issues/5697? 🙇‍♂️

@jcfranco It seems like this would be a regression here. This makes sense because the component doesn't get its position necessary for a proper animation until after its opened. Once it has been opened once the position is there and the correct animation takes place.

I think we'll have to accept this for now. Thoughts?

jcfranco:
> I think we'll have to accept this for now. Thoughts?

Agreed. Let's reopen the issue and comment on the reasoning. cc @geospatialem

Issue #6501, opened 2023-02-22
[Input Date Picker Range]: save previously selected start and end dates on the calendar


svlesiv:
### Actual Behavior

Whenever the start and end dates have been set, the latest state of the calendar will be shown when clicking any of them. 

https://user-images.githubusercontent.com/35533801/220742202-e8ff1777-6ff1-4800-8d22-2034d7337773.mp4

### Expected Behavior

The value of the start date on the calendar will be shown when clicking on the start date input, and the end date when clicking on the end date input field.

https://user-images.githubusercontent.com/35533801/220742374-9c73290c-d789-43ae-b5c9-ce1a45e25a3b.mp4

### Reproduction Sample

https://codepen.io/svlesiv/pen/WNgwBNL

### Reproduction Steps

* Set the Start Date to Feb 1st, 2023
* Set the End Date to April 22nd, 2023
* Click on the Start Date input field
* The latest calendar state (April) will be shown

### Reproduction Version

v1.0.7

### Esri team

ArcGIS Site Scan

Olga8614:
Also reported by BYS: https://devtopia.esri.com/WebGIS/arcgis-portal-app/issues/46825

jcfranco:
It looks like the issue is being caused by the logic handling the caret/toggles (e.g., blur/focus? 🤔❓). I did a quick test using `pointer-events: none` on the toggles and the component seems to work as expected.

DitwanP:
🍠 Verified locally on `main`
![Screenshot 2023-12-15 at 11 15 11 AM](https://github.com/Esri/calcite-design-system/assets/60022782/8c479bb0-851a-4a0b-89b8-ea96c20ea85d)

PR #8414, opened 2023-12-13
fix(input-date-picker): ensure range input toggling is consistent

jcfranco:
**Related Issue:** #6501 

## Summary

This fixes an issue where the start/end picker would not update after being toggled and displayed on the toggled side.

### DIFF
--- a/packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
+++ b/packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
@@ -517,12 +517,14 @@ export class InputDatePicker
     };
 
     return (
-      <Host onBlur={this.deactivate} onKeyDown={this.keyDownHandler}>
+      <Host onBlur={this.blurHandler} onKeyDown={this.keyDownHandler}>
         {this.localeData && (
-          <div class="input-container">
+          <div class={CSS.inputContainer}>
             <div
-              class="input-wrapper"
+              class={CSS.inputWrapper}
+              data-position="start"
               onClick={this.onInputWrapperClick}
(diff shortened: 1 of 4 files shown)


packages/calcite-components/src/components/input-date-picker/resources.ts
```diff
@@ -1,6 +1,15 @@
 export const CSS = {
   assistiveText: "assistive-text",
+  calendarWrapper: "calendar-wrapper",
+  calendarWrapperEnd: "calendar-picker-wrapper--end",
```
eriklharper, line 4: ☝️ should the value here omit the word `picker` to be more inline with the key pattern?

jcfranco: Great catch! 🎣 Fixing.

packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -532,15 +534,15 @@ export class InputDatePicker
                 aria-describedby={this.placeholderTextId}
                 aria-expanded={toAriaBoolean(this.open)}
                 aria-haspopup="dialog"
-                class={`input ${
-                  this.layout === "vertical" && this.range ? `no-bottom-border` : ``
-                }`}
+                class={{
+                  [CSS.input]: true,
+                  [CSS.inputNoBottomBorder]: this.layout === "vertical" && this.range,
+                }}
```
eriklharper, line 540: 🧹

eriklharper (approved):
LGTM! 👍

Issue #8397, opened 2023-12-11
Calcite Stepper: Issues with disabled state, change events, and styles in v2.0.0


noahmulfinger:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

I found the following issues when testing the stepper on calcite-components v2.0.0. They are reproducible in the linked repro sample.
1. It is not possible to disable the next/previous step action buttons when the stepper is in single step view mode.
2. The `onCalciteStepperItemChange` event does not emit when switching to a step that is disabled, or when switching to the first step (using the step action buttons)
3. The `heading`s for non-disabled stepper items are not visible.
4. The content for disabled stepper items is not visible.

### Expected Behavior

1. It should be possible to disable the next/previous step action buttons, and if possible they should be disabled if the next or previous step is marked as disabled.
2. The `onCalciteStepperItemChange` should always emit when a step changes.
3. The headings should always be visible
4. The content should always be visible.

### Reproduction Sample

https://codesandbox.io/p/sandbox/runtime-morning-tqdjmw?file=%2Fsrc%2FApp.js%3A15%2C40

### Reproduction Steps

1. Open the repro sample
2. Switch between steps using the next/previous action buttons
3. View the console log to see the stepper item change events.

### Reproduction Version

2.0.0

### Relevant Info

Tested in Chrome 120.0.6099.71 and Firefox 122.0a1

### Priority impact

p1 - need for current milestone

### Impact

Needed for updating the underlying theme functionality for developer website.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [X] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Developer Experience

jcfranco:
Created a [vanilla repro case](https://codepen.io/jcfranco/pen/YzBbKzV?editors=1000) for comparison purposes based on @noahmulfinger's sample.

Prior to the responsiveness updates, disabled steps could not be accessed, so we need to handle this path as well. There may be some UI/UX tweaks needed as well. cc @anveshmekala @SkyeSeitz @ashetland

anveshmekala:
[...]

PR #8422, opened 2023-12-13
fix(stepper): emits `calciteStepperItemChange` event when switched to first step

anveshmekala:
**Related Issue:** #8397 

## Summary

Emits `calciteStepperItemChange` event when switched to the first enabled step in single view layout.

### DIFF
--- a/packages/calcite-components/src/components/stepper/stepper.tsx
+++ b/packages/calcite-components/src/components/stepper/stepper.tsx
@@ -118,8 +118,7 @@ export class Stepper implements LocalizedComponent, T9nComponent {
    * Fires when the active `calcite-stepper-item` changes.
    *
    */
-  @Event({ cancelable: false })
-  calciteStepperItemChange: EventEmitter<void>;
+  @Event({ cancelable: false }) calciteStepperItemChange: EventEmitter<void>;
 
   /**
    * Fires when the active `calcite-stepper-item` changes.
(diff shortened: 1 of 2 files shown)


alisonailea (approved):

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -506,7 +504,6 @@ export class Stepper implements LocalizedComponent, T9nComponent {
     }
 
     if (
-      this.currentActivePosition &&
```
driskull, line 509: should this maybe just check the type? Something like...

- `this.currentActivePosition != null`?
- `typeof this.currentActivePostion === "number"`?

driskull (approved):
👍

packages/calcite-components/src/components/stepper/stepper.tsx
anveshmekala: Good catch, this made me realize we need that condition to support when all the item's are disabled. Will get that added.

driskull: cool. yeah im guessing the issue was that the currentActivePostion was `0` which is a valid index number but causing this if statement to not get entered.

anveshmekala: yup.

Issue #8421, opened 2023-12-13
Build Issue with webpack - 2.1.0-next.0


jmanke:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When building calcite-components on version `2.1.0-next.0` and up, webpack will fail to build (happens in Dashboard build). Used https://github.com/Esri/calcite-components-examples/tree/main/webpack to reproduce the issue.

<img width="516" alt="Screenshot 2023-12-13 at 3 16 27 PM" src="https://github.com/Esri/calcite-design-system/assets/19629809/99d1698a-fade-4813-b33d-f7dbaac0cbab">

### Expected Behavior

Building with webpack should work.

### Reproduction Sample

https://github.com/Esri/calcite-components-examples/tree/main/webpack

### Reproduction Steps

1. Use calcite-components-examples/webpack
2. Switch calcite-components to use `2.1.0-next.0` or above
3. Run `npm install && npm run build`

### Reproduction Version

2.1.0-next.8

### Relevant Info

MacOS 14.1.1 (23B81)

### Regression?

2.0.0

### Priority impact

p0 - emergency

### Impact

Very high - Dashboards cannot test or bump its calcite version until this is resolved.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Dashboards

@jcfranco

DitwanP:
🍡 Verified on `2.1.0-next.9`
![Screenshot 2023-12-15 at 10 19 39 AM](https://github.com/Esri/calcite-design-system/assets/60022782/ca278b50-56e5-4e73-a43b-f03e3bb214bd)

PR #8423, opened 2023-12-14
build: ensure patches run after prerender test build

benelan:
**Related Issue:** #8421 

## Summary

Ensure the build used for `next` releases includes patches. This fixes a regression caused by #8325, where we removed the extra build on prepublish. The extra build was fixing some issues caused by `npm run test:prerender`, where a build occurred that didn't run pre/post build scripts, including the ESM resolution patch added in #5145.


jcfranco (approved):
🤖🔨

Issue #7677, opened 2023-09-05
Question - Table overflow (scrollbars)


jmanke:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Currently, I'm having difficulty achieving desired horizontal scrollbar behavior with the `calcite-table`. In my use case, the user should be able to scroll vertically and horizontally, regardless of where in the table they are. In this example, there is a wrapping div with `overflow: auto` to allow the user to navigate the table. However, there is an `overflow-x: auto` within the `.table-container` inside the shadow root. This is causing a scrollbar to appear at the bottom of the rows, so the user must scroll to the bottom before scrolling left/right - not only that, but now there are two scrollbars. Removing the wrapping div's `overflow: x` doesn't work either since the user would need to scroll to the bottom of the table before scrolling horizontally. The provided [codepen](https://codepen.io/jmanke/pen/XWoNNbv?editors=1100) goes into more detail. 

Example: [codepen](https://codepen.io/jmanke/pen/XWoNNbv?editors=1100)

| Current double x-scrollbar behavior  | Desired behavior |
| ------------- | ------------- |
| ![Screenshot 2023-09-05 at 10 04 13 AM](https://github.com/Esri/calcite-design-system/assets/19629809/d8134abf-260a-4b9c-a07d-12e47d801ad5) | ![Screenshot 2023-09-05 at 9 59 38 AM](https://github.com/Esri/calcite-design-system/assets/19629809/61f5b378-4612-449f-b048-61cfadbba3be)  |

### Question
Is the desired behavior compatible with A11Y, or should I go about this another way? Is this something that will be addressed in #6646?

### Acceptance Criteria

Allow user to scroll the `calcite-table` vertically and horizontally at any point while navigating the table. In my use case, I'm using an outer div to control the overflow. Either the`calcite-table` shouldn't implement internal scrollbars (or expose the ability to override it), or the `calcite-table` should allow the user to scroll vertically/horizontally as seen in the "Desired behavior" screenshot in the description.

### Which Component

calcite-table

### Example Use Case

Ability for user to navigate the `calcite-table` with scrollbars regardless of their current position in the table.

![Screenshot 2023-09-05 at 9 59 38 AM](https://github.com/Esri/calcite-design-system/assets/19629809/be896cc0-7851-48b3-ad27-2e964aa4cbdc)

### Priority impact

p4 - not time sensitive

### Calcite package

[...]

PR #8424, opened 2023-12-14
fix(table): Improve Table overflow behavior

macandcheese:
**Related Issue:** #7677 

## Summary
Fixes table overflow issues and adds conditional styles to prevent double borders on last visible table row when pagination is present.

previous (from #7677): https://codepen.io/jmanke/pen/XWoNNbv?editors=1100
with this pr: https://codepen.io/mac_and_cheese/pen/abXgNdb?editors=1100

previous (from [#7677 comment](https://github.com/Esri/calcite-design-system/issues/7677#issuecomment-1856176584)): https://codepen.io/Olga-Knyazeva-the-bashful/pen/yLZweJG
with this pr: https://codepen.io/mac_and_cheese/pen/BaMgLLK


packages/calcite-components/src/components/table/table.scss
```diff
@@ -15,20 +15,20 @@
 }
 
 :host {
-  @apply table;
+  @apply flex;
```
driskull, line 18: This would be a breaking change, right? Are we allowing a breaking change here?

driskull (commented):
❓

packages/calcite-components/src/components/table/table.scss
macandcheese: Yeah, not sure if there's another way to fix this without this change. The resultant Chromatic changes are minimal (https://www.chromatic.com/build?appId=6266d45509d7eb004aa254fb&number=3263) - maybe @jcfranco can make a call?

driskull: Sounds good.

We had discussed changes like this being breaking changes since users apps may be expecting the component to display a certain way. Maybe we can get away with these changes if we note them more prominently.

macandcheese: Marking as draft for now.

jcfranco: I think we can proceed with this w/o it being breaking for the following reasons:
 
* `display: table` ➡️  `display: flex` still has a block flow parent-wise (flex-specific styling might kick in, but not likely due to its current display and any overrides to use flex would still be honored)
* Slotted children of `calcite-table` are all table-related, so the styles of these should be encapsulated already, and user content would be unaffected

jcfranco (approved):
📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜
📜🏓📜📜📜🏓📜📜🏓🏓📜📜🏓🏓🏓📜📜🏓🏓🏓📜🏓🏓🏓🏓📜🏓📜
📜🏓🏓📜📜🏓📜🏓📜📜🏓📜📜🏓📜📜🏓📜📜📜📜🏓📜📜📜📜🏓📜
📜🏓📜🏓📜🏓📜🏓📜📜🏓📜📜🏓📜📜🏓📜📜📜📜🏓🏓🏓📜📜🏓📜
📜🏓📜📜🏓🏓📜🏓📜📜🏓📜📜🏓📜📜🏓📜📜📜📜🏓📜📜📜📜📜📜
📜🏓📜📜📜🏓📜📜🏓🏓📜📜🏓🏓🏓📜📜🏓🏓🏓📜🏓🏓🏓🏓📜🏓📜
📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜📜

Issue #6473, opened 2023-02-14
Audit components `open`/`closed` props for consistency


SkyeSeitz:
### Description
An audit to define open and `open`/`closed` verses `expanded`/`collapsed` properties and use consistently across Calcite components.

**Audit summary**: The audit effort was completed in June 2024, and deprecations and new functionality will be added during the `3.0` breaking change release, with anticipated removal in the `4.0` breaking change release.

cc @geospatialem @ashetland @macandcheese @jcfranco 

### Acceptance Criteria
New components and props should follow this updated pattern:
`open`/`close`: Toggles visibility of entire component, or its menu
`expand`/`collapse`: Toggles visibility of more content within a component 

Certain components may have need for both an open/close state as well as a expanded/collapsed such as:
`[panel]` can be `open` (component is visible) & `collapsed` (content area is hidden) at the same time

Action items for `3.0`:
`[list-item]`  deprecate `open` & add functionality to `expanded`
`[block]`  deprecate `open` & add functionality to `expanded`
`[block-section]`  deprecate `open` & add functionality to `expanded`
Consider an **eslint rule** to ensure a component does not have both `open`/`closed` -or- `expanded`/`collapsed`

Events:
Leverage the `open`/`close` utility for `expanded` and `collapsed`. Likely just needs some small functional tweaks. _Refer to [Eliza's comment below](https://github.com/Esri/calcite-design-system/issues/6473#issuecomment-2148144692) for more context._

🔗 Link to [Figma file](https://www.figma.com/design/MDMxs2PqRTQUaiOrFrmXQE/Open%2FExpanded?node-id=609-40577&t=CbeVDC35k9WvqaAb-11)

<img width="13142" height="2558" alt="Image" src="https://github.com/user-attachments/assets/438ad2fe-d531-4c47-acf9-736ecd04161c" />

<img width="8020" height="2202" alt="Image" src="https://github.com/user-attachments/assets/17731a9b-5038-48c7-89e4-9fd20990edbe" />

### Which Component

List Item, Block

Applies across multiple components

### Esri team

Calcite (design)

**monday.com sync:** #9366737060

SkyeSeitz:
This is a proposal to start the discussion with follow up questions:
<img width="1459" alt="image" src="https://user-images.githubusercontent.com/29716057/218886971-cf17b043-761b-426b-ad46-f8f0c23f7527.png">

1. Depending on how we define `open`, does anything in the `open` category fit better in `active` or `expanded`?
[...]

PR #8433, opened 2023-12-15
feat(list-item): Add calciteListItemToggle event.

driskull:
**Related Issue:** #8434

## Summary

- Adds `calciteListItemToggle ` event for when a list item open is toggled via user interaction. 
- Adds test

### DIFF
--- a/packages/calcite-components/src/components.d.ts
+++ b/packages/calcite-components/src/components.d.ts
@@ -6286,6 +6286,7 @@ declare global {
     interface HTMLCalciteListItemElementEventMap {
         "calciteListItemSelect": void;
         "calciteListItemClose": void;
+        "calciteListItemOpen": void;
         "calciteInternalListItemSelect": void;
         "calciteInternalListItemSelectMultiple": {
     selectMultiple: boolean;
@@ -10079,6 +10080,10 @@ declare namespace LocalJSX {
           * Fires when the close button is clicked.
          */
         "onCalciteListItemClose"?: (event: CalciteListItemCustomEvent<void>) => void;
+        /**
+          * Fires when the open button is clicked.
+         */
+        "onCalciteListItemOpen"?: (event: CalciteListItemCustomEvent<void>) => void;
         /**
           * Emits when the item's content is selected.
          */
--- a/packages/calcite-components/src/components.d.ts
+++ b/packages/calcite-components/src/components.d.ts
@@ -6286,7 +6286,7 @@ declare global {
     interface HTMLCalciteListItemElementEventMap {
         "calciteListItemSelect": void;
         "calciteListItemClose": void;
-        "calciteListItemOpen": void;
+        "calciteListItemToggle": void;
         "calciteInternalListItemSelect": void;
         "calciteInternalListItemSelectMultiple": {
     selectMultiple: boolean;
@@ -10081,13 +10081,13 @@ declare namespace LocalJSX {
          */
         "onCalciteListItemClose"?: (event: CalciteListItemCustomEvent<void>) => void;
         /**
-          * Fires when the open button is clicked.
-         */
-        "onCalciteListItemOpen"?: (event: CalciteListItemCustomEvent<void>) => void;
-        /**
           * Emits when the item's content is selected.
          */
         "onCalciteListItemSelect"?: (event: CalciteListItemCustomEvent<void>) => void;
+        /**
+          * Fires when the open button is clicked.
+         */
+        "onCalciteListItemToggle"?: (event: CalciteListItemCustomEvent<void>) => void;
         /**
           * When `true`, the item is open to show child components.
          */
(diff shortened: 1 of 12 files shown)


packages/calcite-components/src/components.d.ts
```diff
@@ -10079,6 +10080,10 @@ declare namespace LocalJSX {
           * Fires when the close button is clicked.
          */
         "onCalciteListItemClose"?: (event: CalciteListItemCustomEvent<void>) => void;
+        /**
+          * Fires when the open button is clicked.
+         */
+        "onCalciteListItemOpen"?: (event: CalciteListItemCustomEvent<void>) => void;
```
macandcheese, line 10086: Mostly because we already have ‘close’ event for the “close button” interaction, but is it possible to use Toggle here or something besides ‘open’? We use ‘calcitePanelToggle‘ in Panel. It could also fire when collapsed?

driskull: Yeah I can do that. its mapped to the open property though.

macandcheese: Yeah, it’s tough with existing ‘close’ as that, in practice, is more of “dismiss” - is Toggle is too close to ‘select’ event?  OpenToggle?

driskull: Updated. Ill let @jcfranco comment on name as well.

jcfranco: **TL;DR**: I think we should go with `calciteListItemToggle` event firing when opened/closed. 

I'd like to avoid introducing an event name that mixes both open/close and toggling. We could introduce an individual toggle event (similar to `block`'s toggle event before we added open/close events) and check the event target's `open` value for now until we look more into the following:

* revisit close/dismiss/remove type props and their respective events (e.g., `tip` has a [closed](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/tip/tip.tsx#L50) prop and [dismissed](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/tip/tip.tsx#L141) event, `flow-item` has a [closed](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/flow-item/flow-item.tsx#L198) prop and a [close](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/flow-item/flow-item.tsx#L198) event, `pick-list-item` [deprecated] has no `close` prop, but similar affordance and a [removed](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/pick-list-item/pick-list-item.tsx#L244) event).
* use `open`/`close`, `expand`/`collapse` per https://github.com/Esri/calcite-design-system/issues/6473#issuecomment-1730103153.

Unfortunately, `list` is in a bit of a tough spot with some of these props and patterns we're adopting.

driskull: Thanks 👍

packages/calcite-components/src/components/list-item/list-item.e2e.ts
```diff
@@ -275,4 +275,20 @@ describe("calcite-list-item", () => {
 
     expect(calciteListItemClose).toHaveReceivedEventTimes(1);
   });
+
+  it("should fire calciteListItemToggle event when opened", async () => {
```
jcfranco, line 279: Can you update the test to assert on the event emitting for both opening and closing?

packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -740,11 +745,16 @@ export class ListItem
     this.handleOpenableChange(event.target as HTMLSlotElement);
   };
 
-  toggleOpen = (): void => {
-    this.open = !this.open;
+  handleToggleOpenClick = (): void => {
+    this.toggleOpen();
+  };
+
+  toggleOpen = (value = !this.open): void => {
```
jcfranco, line 752: For readability, WDYT about renaming this to `toggle`?

#### before

`toggleOpen()`
`toggleOpen(true)`
`toggleOpen(false)`

#### after

`toggle()`
`toggle(true)`
`toggle(false)`

jcfranco (approved):
Noice! 😎

Once the E2E test is updated, this should be good to merge!

packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -669,7 +674,7 @@ export class ListItem
     this.calciteInternalListItemChange.emit();
   }
 
-  closeClickHandler = (): void => {
+  handleCloseClick = (): void => {
```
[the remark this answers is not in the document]
driskull: Yeah we should probably have an issue to do this across all components.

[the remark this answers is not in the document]
driskull: Should we make render functions (not render() but the ones that return VNodes) private as well?

Issue #8380, opened 2023-12-08
Calcite Table: Cell background and border styles do not work


Anastasiia-Boleiko:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Applying `--calcite-table-cell-background` and `--calcite-table-cell-border-color` does not work.

### Expected Behavior

Properties can be applied to a cell.

### Reproduction Sample

https://codepen.io/anastasiia-zrh/pen/qBgvbYq?editors=1100

### Reproduction Steps

1. Create a table or use [codepan example](https://developers.arcgis.com/calcite-design-system/components/table/) from the docs
2. Try changing cell color and background using style parameters from the [table cell docs](https://developers.arcgis.com/calcite-design-system/components/table-cell/#api-reference-styles---calcite-table-cell-background)

### Reproduction Version

v1.11.0

### Priority impact

p0 - emergency

### Impact

This feature is needed for the redesigned component that will be released in Feb 2024. Code freeze is in the first half of January.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [X] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Urban

macandcheese:
As a workaround for now, you can use the `--calcite-internal-table-cell-background` to set the cell background, and `--calcite-ui-focus-color` for the focused ring, if you'd like to match that: https://codepen.io/mac_and_cheese/pen/abXMpRx?editors=1000

![Screenshot 2023-12-08 at 2 14 34 PM](https://github.com/Esri/calcite-design-system/assets/4733155/23eb870f-2b4a-4e74-b255-7a01240bdd3a)


Currently, the border color is set partially by the parent row as well, we may need to change how that is applied if we want folks to be able to easily configure border, per cell, via css.

Anastasiia-Boleiko:
@macandcheese Thanks a lot for a quick response! 😊
I think the proposed workaround serves our case. Thanks a lot 👍

geospatialem:
Verified in `2.1.0-next.10` with the following:

```html
<style>
.error {
      --calcite-table-cell-background: rgba(200, 50, 20, 0.15);
}
</style>
<calcite-table caption="Simple table">
    <calcite-table-row slot="table-header">
[...]

PR #8439, opened 2023-12-16
fix(table-cell): Fix background css variable

macandcheese:
**Related Issue:** #8380 

## Summary
Fixes a typo that caused this variable not to work, and removes an unused border variable from the initial Table PR that snuck through.


jcfranco (approved):
✨🔧✨

Issue #8436, opened 2023-12-15
Console error appears when selecting just an end date for input date picker


abp6318:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

![Screenshot 2023-12-15 at 4 24 40 PM](https://github.com/Esri/calcite-design-system/assets/60486980/d5f0f346-548d-499f-85a9-1abdbe7d6754)

Console error appears when user selects the end date in the date range input.

### Expected Behavior

No error appears when selecting an end date.

### Reproduction Sample

https://codepen.io/AaronPutterman/pen/PoVrdoY?editors=100

### Reproduction Steps

1. Open the repro sample or [visit the first variant example on the official calcite docs](https://developers.arcgis.com/calcite-design-system/components/input-date-picker/)
2. Open the console
3. Select an end date
4. Error appears in the console

### Reproduction Version

1.11.0

### Relevant Info

MacBook Pro, 16-inch, 2023
Chip: Apple M2 Pro
Memory: 32 GB
macOS: Sonoma 14.1.2
Browser: Chrome

### Priority impact

p2 - want for current milestone

### Impact

Little to no known impact yet, but may have residual effects elsewhere and should be investigated.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Hub

DitwanP:
🍦 Verified locally on `main`

PR #8444, opened 2023-12-18
fix(date-picker): prevent console error when selecting just an end date for input date picker

driskull:
**Related Issue:** #8436

## Summary

- Quick fix to remove console error
- Ideally, I think a user should be able to select the end date first and the whole hover range experience should work. I tried getting that working quickly but wasn't successful. I think we'll need a followup issue.


jcfranco (approved):
> Ideally, I think a user should be able to select the end date first and the whole hover range experience should work. 

Agreed. Per your suggestion, can you create a follow-up issue?

driskull:
Added #8454

PR #8462, opened 2023-12-20
fix(list): fix event detail newIndex when down arrow pressed to sort

driskull:
**Related Issue:** #8464

## Summary

- fixes `newIndex` event detail
- adds test


jcfranco (approved):
🤘🏼🗿🤘🏼

PR #8483, opened 2023-12-22
feat(handle): Add `blurUnselectDisabled` property to disable unselecting handle on blur.

driskull:
**Related Issue:** #8485

## Summary

- Adds `blurUnselectDisabled` property
- This is needed so that the element is not deactivated after blur when moving an item via keyboard.
- Adds test for functionality


driskull:
@benelan can you review this one?

packages/calcite-components/src/components/handle/handle.tsx
```diff
@@ -106,6 +106,13 @@ export class Handle implements LoadableComponent, T9nComponent, InteractiveCompo
    */
   @Prop() label: string;
 
+  /**
+   * When `true`, disables unselecting the component when blurred.
+   *
+   * @internal
+   */
+  @Prop() blurUnselectDisabled = false;
```
macandcheese, line 114: Is there a way to tie this name to `activated`, or would that not be accurate? Just a nitpick as its an internal prop, but its a bit confusing if unfamiliar with the component.

packages/calcite-components/src/components/handle/handle.e2e.ts
```diff
@@ -37,6 +37,51 @@ describe("calcite-handle", () => {
     expect(await handle.getProperty("activated")).toBe(true);
   });
 
+  it("sets activated to false when blurred", async () => {
```
macandcheese, line 40: Does this fix any behavior in list or another consuming component? If so, is it worth adding a check there? Disregard if not the case.

macandcheese (approved):
🏆 Nice! Just a couple nitpick / q's.

packages/calcite-components/src/components/handle/handle.tsx
driskull: I was thinking that but I think the next step is going to be renaming `active` to `selected` and making it public.

Issue #7071, opened 2023-05-31
Update API reference descriptions for consistency across components


geospatialem:
### Description

Comb through all components and update all API ref descriptions to fit our style guide from https://github.com/Esri/calcite-components/issues/4442, which include:

- Use complete sentences with proper grammar, capitalization, and punctuation.  
- No abbreviations, e.g. use "property" instead of "prop".  
- Verbs must be in present tense (e.g., `"emits"` instead of `"emitted"`).  
- When referring to the component that the property/slot/method/css variable belongs to, call it "the component".  
- Use the full tag name when referencing other Calcite Components (prefix with `calcite-`), e.g. `calcite-button` instead of `button`.  
- For plural context, use `calcite-button`s instead of `calcite-button` elements.  
- Use backticks (``` ` ```) for the names of slots, events, properties, CSS variables, and component names.  
    - In line with the JSAPI docs [example](https://developers.arcgis.com/javascript/latest/api-reference/esri-views-View.html#ready)  
    - Docs will incorporate styling with [afd-calcite-documentation issue #490](https://github.com/ArcGIS/afd-calcite-documentation/issues/490)  
- Use double quotes (`"`) for the values of properties/attributes and event details.  
- Only use single quotes (`'`) as apostrophes.  
- No links or URLs allowed in descriptions. If a link is necessary, a [custom JSDoc tag](https://stenciljs.com/docs/docs-json#custom-jsdocs-tags) should be added and parsed in the SDK site.  
- Refrain from using "e.g." or "i.e." references. Leverage "such as" (or similar) where examples are referenced.  
- Use "Accessible" instead of "Aria" language  
- Verify slots don't use the text "optional" in their descriptions  
- Remove optional JSDoc tags for props (`?`)

### Original Issue Description
Our property/slot/method descriptions are all over the place right now. We need to decide on some conventions, and then audit and proofread to make sure they are consistent. Some things to consider:

- Should they be sentences with proper case/punctuation/grammar, or should they be more like bullet points?
- Which verb tense should we use?
[...]

PR #8484, opened 2023-12-22
feat(handle): add selected property and calciteHandleChange event.

driskull:
**Related Issue:** #8522

## Summary

- depends on #8483
- Renames `activated` property to `selected` to align with other components
- Makes `selected` public
- Adds public event `calciteHandleChange` for when `selected` is changed via user
- Renames internal event `calciteInternalHandleChange` to `calciteInternalAssistiveTextChange` to better illustrate its purpose
- adds tests


packages/calcite-components/src/components/handle/handle.tsx
```diff
@@ -180,14 +180,21 @@ export class Handle implements LoadableComponent, T9nComponent, InteractiveCompo
   // --------------------------------------------------------------------------
 
   /**
-   * Emitted when the handle is activated and the up or down arrow key is pressed.
+   * Emits whenever the component is selected or unselected.
```
benelan, line 183: @geospatialem @DitwanP do we have a convention on "Emits" vs "Emitted"? I've seen them both throughout the code base. Maybe we should pick one and do some cleanup in a different PR? I'm partial to "Emits" because we use "Specifies" for properties.

benelan (approved):
LGTM! We already have `selected` in other components so it's safe addition, naming-wise. Otherwise I'd say we should wait for @jcfranco to get back.

packages/calcite-components/src/components/handle/handle.tsx
geospatialem: Good question, we've moved towards "Emits" for the reason you specify above, to keep present tense on the docs. It looks like there are two outliers in `list` and `list-item-group`. Otherwise there is mention of "emitted" throughout the code, but not in the doc. Added this note to our conventions for #7071.

driskull:
@jcfranco when you're back. let me know if we should revert this or keep it internal.

Issue #8493, opened 2023-12-23
list-item: should store last focused cell from focusing on elements within a cell


driskull:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Currently, with the list's tree-grid role and keyboard navigation, only one row and cell is focusable at a time.

When an action or handle is clicked within a cell, the cell and row do not get an tabindex.

### Expected Behavior

When an action or handle is clicked within a cell, the cell and row should get an updated tabindex so that it is the current active cell and row.

### Reproduction Sample

https://codepen.io/driskull/pen/NWJWzPN?editors=1000

### Reproduction Steps

1. Open the codepen
2. Click on drag handle within the third row.
3. Use the up arrow key. 
4. Notice how it goes to the filter element as if the currently focused row is the first one.
5. Click the 3rd row drag handle again.
6. Use the down arrow key.
7. It focuses the second row instead of the 4th row.

### Reproduction Version

2.1.0

### Priority impact

p2 - want for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

geospatialem:
Verified in `2.2.0-next.6`:

![verify-list-drag](https://github.com/Esri/calcite-design-system/assets/5023024/400c6934-b28b-4bf6-a2bf-572063f00671)

PR #8494, opened 2023-12-23
fix(list-item): store last focused cell from focusing on elements within a cell.

driskull:
**Related Issue:** #8493

## Summary

- add `onFocusWithin` listeners for each appropriate grid cell to set which cell is focused
- add `onFocusWithin` listener for when a row is selected, emit active event to update list to the currently active row
- add data-attribute for testing 
- add test


packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -27,7 +27,7 @@ import {
 } from "../../utils/interactive";
 import { SelectionMode } from "../interfaces";
 import { SelectionAppearance } from "../list/resources";
-import { CSS, ICONS, SLOTS } from "./resources";
+import { CSS, dataTestActiveCellAttr, ICONS, SLOTS } from "./resources";
```
benelan, line 30: Very Nitpick: I think we try to avoid abbreviations, so should we do `dataTestActiveCellAttribute`? 
I get why it was abbreviated though, because that's pretty long lol

benelan (approved):

packages/calcite-components/src/components/list-item/list-item.tsx
driskull: lol yeah I agree about abbr. :)

 I'll rename.

Issue #4759, opened 2022-06-16
[Tile / Tile Group] Add visual scales


macandcheese:
### Description

We should bring these components in line with others by adding additional scales to provide S, M, L.

### Acceptance Criteria

Additional scales are added to these components, and to the Figma component library.

### Which Component

Tile / Tile Group

SkyeSeitz:
Link to Figma specs: https://www.figma.com/file/8CiAVHkXIMTsI9hTSbotEz/Tile-Redesign-%5BEpic-6662%5D?type=design&node-id=720%3A2738&mode=dev

geospatialem:
Verified in `2.2.0-next.18` when `tile`'s `scale` is set to `"s"`, `"m"`, or `"l"` (`"m"` is the default).

![image](https://github.com/Esri/calcite-design-system/assets/5023024/5bb275b2-0e62-4fdc-9fc8-b7a1176e6144)

cc @SkyeSeitz + @eriklharper It looks like the original ask was for both `tile` and `tile-group`. The [Figma file](https://www.figma.com/file/8CiAVHkXIMTsI9hTSbotEz/Tile-Redesign-%5BEpic-6662%5D?node-id=720%3A2738&mode=dev) mentions adding a new `tile-group` component, but I'm not seeing the addition in the code. Is that part of this issue, or a different issue? Want to verify before closing the above out. TIA! 🙌🏻 

![image](https://github.com/Esri/calcite-design-system/assets/5023024/10f1038b-7ac2-4772-8548-def6cd6bb3de)

geospatialem:
Closing with `tile` completed per https://github.com/Esri/calcite-design-system/issues/4759#issuecomment-1893985006, and `tile-group` will be addressed in next months release with https://github.com/Esri/calcite-design-system/issues/8615.

Issue #6602, opened 2023-03-14
Publish Tailwind Config as Preset


am-maneaters:
### Summary

The tailwind config for Calcite Components does not get exported with the components so when I am using Tailwind in a project, there are a lot of class name conflicts. I'd also just like to use the same colors/typography as Calcite for a consistent UI. Currently, I just copy and paste the entire config from the calcite-components repo and save it locally.

### Desired Outcome

Publish the Calcite Components tailwind config so users can use it as a "Preset" in their tailwind projects.

### Resources

https://tailwindcss.com/docs/presets

jrowlingson:
Related https://github.com/Esri/calcite-components/issues/2143

jcfranco:
I mentioned this in the linked PR, but we are planning to distribute the preset in its own package once the design tokens are ready and applied to the config. 

Related: https://github.com/Esri/calcite-components/issues/6558

DitwanP:
🍠

PR #8496, opened 2023-12-23
feat(tile): add visual scales

eriklharper:
**Related Issue:** #4759 

## Summary

This PR adds a `scale` property to control visual display for small, medium and large scales.


packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -1,4 +1,31 @@
+:host([scale="s"]) {
+  --calcite-tile-content-gap: var(--calcite-spacing-sm);
```
macandcheese, line 2: Should these be marked `--calcite-internal-x`? 

Anything intended to be public should have a doc / description. I'd think we want to "start small" with what we offer here, maybe just `--calcite-tile-heading-color` / `--calcite-tile-description-color`.

eriklharper: > Should these be marked `--calcite-internal-x`?
> 
> Anything intended to be public should have a doc / description. I'd think we want to "start small" with what we offer here, maybe just `--calcite-tile-heading-color` / `--calcite-tile-description-color`.

Good question for @alisonailea and @SkyeSeitz.  If these are intended to eventually be "component" tokens specific to Tile, should these have the `internal` label included in their name?

eriklharper: Decided to go the route of keeping this as simple as possible and only defining component tokens that are absolutely necessary for theming purposes only, mainly color.

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -47,75 +43,108 @@
...
-    break-words
-    font-medium
-    duration-150
-    ease-in-out;
+    color: var(--calcite-color-text-2);
+    font-weight: 500;
+    overflow-wrap: break-word;
+    pointer-events: none;
+    transition-duration: 150ms;
+    transition-timing-function: cubic-bezier(0.4, 0, 0.2, 1);
```
macandcheese, line 51: Can this be stored + referenced? It's used below as well. Do other components have a need for this value?

eriklharper: Not sure if other components need it and I can't remember why this is even needed in Tile.  I can reference it at the least with a Sass variable.

eriklharper: I removed these styles from `.heading` and `.description` since they don't seem to be needed because the host level rules cascade down to them, which is cleaner and less redundant.  I think it is out of scope of this PR to store and reference these values anyway since this is just about adding `scale` support, but in the forthcoming tile redesign PR I could look at adding a sass mixin for this.

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -1,41 +1,38 @@
 :host {
-  @apply bg-foreground-1
-    text-color-3
-    box-border
-    inline-block
-    select-none
-    duration-150
-    ease-in-out;
+  background-color: var(--calcite-color-foreground-1);
```
jcfranco, line 2: Can we revert the Tailwind ➡️ CSS changes? I'd like to keep TW utils until we wrap up https://github.com/Esri/calcite-design-system/issues/6602 (mainly for visual regression testing purposes).

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -47,75 +44,108 @@
...
+  }
+}
+
+:host([scale="l"]) {
+  .container {
+    gap: var(--calcite-spacing-lg);
+  }
+  .heading {
+    font-size: var(--calcite-font-size-0);
+    line-height: 1.375rem;
```
jcfranco, line 98: Is there a design token or TW util we could use for line height?

jcfranco (approved):
🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦
🟦🟨🟦🟦🟦🟨🟦🟦🟨🟨🟦🟦🟨🟨🟨🟦🟦🟨🟨🟨🟦🟨🟨🟨🟨🟦🟨🟦
🟦🟨🟨🟦🟦🟨🟦🟨🟦🟦🟨🟦🟦🟨🟦🟦🟨🟦🟦🟦🟦🟨🟦🟦🟦🟦🟨🟦
🟦🟨🟦🟨🟦🟨🟦🟨🟦🟦🟨🟦🟦🟨🟦🟦🟨🟦🟦🟦🟦🟨🟨🟨🟦🟦🟨🟦
🟦🟨🟦🟦🟨🟨🟦🟨🟦🟦🟨🟦🟦🟨🟦🟦🟨🟦🟦🟦🟦🟨🟦🟦🟦🟦🟦🟦
🟦🟨🟦🟦🟦🟨🟦🟦🟨🟨🟦🟦🟨🟨🟨🟦🟦🟨🟨🟨🟦🟨🟨🟨🟨🟦🟨🟦
🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦

Once the TW ➡️ CSS changes are rolled back, this is good to merge!

packages/calcite-components/src/components/tile/tile.scss
eriklharper: I'll let @SkyeSeitz and @alisonailea chime in on this.  Ali informed me that there aren't Figma variables for certain token types, including `line-height`. 
image

I am noticing some line-height css variables in calcite-design-tokens, but only 1 of the values on the Tile Scale Figma file matches up with any of these, which is `1.375rem` for `--calcite-font-line-height-relative-snug`.

image

SkyeSeitz (approved):

PR #8517, opened 2023-12-28
feat(checkbox, combobox, input-date-picker, input-time-picker, segmented-control, select): add required property

benelan:
**Related Issue:** #8057

## Summary

Remove the `@internal` tag from the `required` property on form components.


driskull (approved):
👍

jcfranco (approved):
🎉

To provide more context in the changelog, can you revisit the PR title to include the enhanced components in the scope?

`feat(a, b, c...): add required property`

benelan:
@aaronpi and @SkyeSeitz provided design feedback on #8057 and @SkyeSeitz [asked](https://github.com/Esri/calcite-design-system/issues/8057#issuecomment-1875973007):


>  Is the required prop applicable [to Segmented Control] since it should always have a selected item?

@jcfranco what do you think, should we keep `required` internal for Segmented Control? Slider and Rating may also be worth reevaluating, since I think both of them always have a value.

If we keep `required` internal, then they won't have any constraint validation properties and wouldn't need the internal input-message for validation.

jcfranco:
> @aaronpi and @SkyeSeitz provided design feedback on https://github.com/Esri/calcite-design-system/issues/8057 and 
> @SkyeSeitz https://github.com/Esri/calcite-design-system/issues/8057#issuecomment-1875973007:
> 
> Is the required prop applicable [to Segmented Control] since it should always have a selected item?
> 
> @jcfranco what do you think, should we keep required internal for Segmented Control? Slider and Rating may also be worth reevaluating, since I think both of them always have a value.

I [replied](https://github.com/Esri/calcite-design-system/issues/8057#issuecomment-1879360542) in the referenced issue, but the TLDR is: we still need a public `required` prop for segmented control.

> Slider and Rating may also be worth reevaluating, since I think both of them always have a value.

Maybe `slider` does, but `rating` does support no initial value.

jcfranco:
@benelan pointed out that rating has a [default value of 0](https://github.com/Esri/calcite-design-system/blob/81cb5cc58d082c5d62d91a5dcaa5cfa8d993e626/packages/calcite-components/src/components/rating/rating.tsx#L123-L124) when not set, which confirms that both slider and rating always have a value. Thanks for the catch!

Issue #7958, opened 2023-10-06
[input-date-picker] sometimes render the date incorrectly


rpanichakit9541:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

`calcite-input-date-picker` sometimes render the date incorrectly (see screenshot)

![MicrosoftTeams-image (1)](https://github.com/Esri/calcite-design-system/assets/44479773/79cce23b-b611-4c2e-94ae-26e016ff59c9)

### Expected Behavior

`calcite-input-date-picker` should always render the date correctly

### Reproduction Sample

~Have not been able to get it to happen in Codepen~

https://codepen.io/benesri/pen/YzdodgJ

### Reproduction Steps

I don't have a great set of repro steps.  Seem to happen randomly and inconsistently.  I'm seeing it in the metadata editor on devext.arcgis.com in Firefox, but it's not reproducible 100% of the time.

@jcfranco mentioned that @benelan may know the cause and might be able to generate a repro case.

### Reproduction Version

1.9.0

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Online

benelan:
Sorry for the delay, here is the repro sample:
https://codepen.io/benesri/pen/YzdodgJ

@jcfranco and I discussed and agreed on a solution, so when this issue gets triaged to a milestone please assign it to me. Thanks!

cc @geospatialem

mcollins-GPI:
@benelan I thought I was going crazy! Is there any workaround for this that you have found? It looks like it will get fixed when it does.

DitwanP:
🧁 Verified locally on `main`

PR #8518, opened 2023-12-28
fix(input-date-picker): hard to reproduce numbering-system caching issue

benelan:
**Related Issue:** #7958

## Summary

Resolves an extremely hard to reproduce caching issue that occurs due to Alert's timeouts. The fix is to cache Alert's number formatter per component so it doesn't impact other components.


packages/calcite-components/src/components/alert/alert.tsx
```diff
@@ -201,12 +208,6 @@ export class Alert implements OpenCloseComponent, LoadableComponent, T9nComponen
   }
 
   render(): VNode {
-    numberStringFormatter.numberFormatOptions = {
```
driskull, line 204: This looks good but there are other components doing this same thing. Should we update them as well?

driskull (approved):
👍

packages/calcite-components/src/components/alert/alert.tsx
benelan: Yeah I was thinking the same thing. I'll start doing that after I finish some of my other issues for this milestone.

benelan: FWIW this is the only one causing an actual bug that I'm aware of because of the delay. See the repro sample:
https://codepen.io/benesri/pen/YzdodgJ

jcfranco:
@benelan A bit late, but could you revisit the PR title and changelog entry to be brief, yet descriptive (as best as possible)? The start of the description seems like a good starting point:

> Resolves an extremely hard to reproduce caching issue that occurs due to Alert's timeouts.

Issue #8519, opened 2023-12-28
Transparent Action hover state missing background color


ashetland:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When `appearance="transparent"` there is no background color change onHover.
![CleanShot 2023-12-28 at 14 14 45](https://github.com/Esri/calcite-design-system/assets/108549080/eaa2d09a-6ea0-4665-a3db-0da076e1b9be)

### Expected Behavior

The background color should change per issue #3834.

### Reproduction Sample

https://codepen.io/ashetland/pen/xxBGPZM?editors=1000

### Reproduction Steps

Hover over transparent Action in code sample.

### Reproduction Version

v2.1.0

### Regression?

v1.11.0

### Priority impact

p2 - want for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

geospatialem:
Since this is a regression between `1.11.0` and `2.0.0`, wondering if this might be related to design tokens. cc @alisonailea

driskull:
The var being referenced is `--calcite-button-transparent-hover`

alisonailea:
--calcite-button-transparent-hover: and --calcite-button-transparent-press are exclusively used by the action component. So these should be re-added as action-component tokens. There are existing semantic tokens that can be used. 

--calcite-color-transparent-hover is a 0.04 vs 0.05 opacity but --calcite-color-transparent-press is an exact match.

--calcite-action-color-transparent-hover: var(--calcite-color-transparent-hover);
--calcite-action-color-transparent-press: var(--calcite-color-transparent-press);

DitwanP:
🍡 Verified on `2.2.0-next.15`

PR #8532, opened 2024-01-02
fix(action): update component tokens to support transparent

alisonailea:
**Related Issue:** #8519

## Summary

Action token was mislabeled as a button token and was missed in the design token refactor. This adds the token back to the action component


driskull (approved):
👍 Looks good @alisonailea

PR #8536, opened 2024-01-02
docs(b and c components): consistent api description refs

DitwanP:
**Related Issue:** #7071

## Summary
Updates doc consistency across b and c-named components defined in the above issue for props, events, methods, and css vars, including:

- `button`
- `block`
- `block-section`
- `color-picker-hex-input`
- `color-picker`
- `combobox`

cc @geospatialem

PR #8540, opened 2024-01-02
docs(d, f, g, and h-named components): update api description refs

DitwanP:
**Related Issue:** #7071 

## Summary
Updates doc consistency across b and c-named components defined in the above issue for props, events, methods, and css vars, including:

- `date-picker-day`
- `date-picker-month`
- `date-picker-month-header`
- `dropdown-group`
- `dropdown-item`
- `filter`
- `graph`
- `handle`

cc @geospatialem


packages/calcite-components/src/components/dropdown-group/dropdown-group.tsx
```diff
@@ -43,10 +43,10 @@ export class DropdownGroup {
...
 
   /**
-   * Specifies the selection mode for `calcite-dropdown-item` children, defaults to `single`:
-   * - `multiple` allows any number of selected items,
-   * - `single` allows only one selection (default),
-   * - `none` doesn't allow for any selection.
+   * Specifies the selection mode for `calcite-dropdown-item` children:
+   * - `"multiple"` allows any number of selected items,
+   * - `"single"` allows only one selection,
+   * - `"none"` doesn't allow for any selection.
```
geospatialem, line 49: Similar to https://github.com/Esri/calcite-design-system/pull/8536#pullrequestreview-1801192987 - nitpick: _Believe_ in this case both the `:` and `-` / `,` and `-` would be shown together. Maybe remove the `-` throughout so only one character is displayed in the doc?

packages/calcite-components/src/components/date-picker-month/date-picker-month.tsx
```diff
@@ -47,14 +47,14 @@ export class DatePickerMonth {
...
   @Prop() activeDate: Date = new Date();
 
   /** Start date currently active. */
-  @Prop() startDate?: Date;
+  @Prop() startDate: Date;
 
-  /** End date currently active  */
-  @Prop() endDate?: Date;
+  /** End date currently active.  */
+  @Prop() endDate: Date;
```
geospatialem, line 57: These optional props _may_ still be needed for `startDate` and `endDate`, can they be added back in?

geospatialem (approved):
Awesome! Reviewed the doc'd components, `dropdown-group`, `dropdown-item`, and `filter`. 🏆 

It looks like there were some changes to the optional props in `calcite-date-picker-month`, which _couldt_ have ramifications elsewhere. Since its a JSDoc change and we aren't officially documenting the component, wondering if we could add the `?` back in for good measure?

Otherwise, the non-doc'd component doc changes can be merged in as part of the PR, but out of scope for the review. 🔍

DitwanP:
> Awesome! Reviewed the doc'd components, `dropdown-group`, `dropdown-item`, and `filter`. 🏆
> 
> It looks like there were some changes to the optional props in `calcite-date-picker-month`, which _could_ have ramifications elsewhere. Since its a JSDoc change and we aren't officially documenting the component, wondering if we could add the `?` back in for good measure?
> 
> Otherwise, the non-doc'd component doc changes can be merged in as part of the PR, but out of scope for the review. 🔍

For the optional tag, theres a bullet in the list of things to check that says to remove them or is that referring to something else?

geospatialem:
> For the optional tag, theres a bullet in the list of things to check that says to remove them or is that referring to something else?

Yes, that's accurate, but for the documented components. Its likely outdated now with our move to `1.0.0` last January, since that's not something we've documented since the first production release.

With the undocumented components there is the possibility of tests that could fail as a result of removing the optional JSDoc tag.

DitwanP:
> With the undocumented components there is the possibility of tests that could fail as a result of removing the optional JSDoc tag.

Ahh okay, got it! 👍

geospatialem (approved):
Great additions! 🎇

Issue #8539, opened 2024-01-02
`<calcite-list-item>` elements with `hidden` attributes are visible


nwhittaker:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The `<calcite-list-item>` elements with `hidden` attributes are visible.

### Expected Behavior

The `<calcite-list-item>` elements with `hidden` attributes are hidden.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/MWxaLRp

### Reproduction Steps

1. Open the repro sample.
2. See the hiking trails and waterfall list items are visible despite having `hidden` attributes.

### Reproduction Version

2.1.0

### Relevant Info

Spun off from https://github.com/Esri/calcite-design-system/issues/6544#issuecomment-1874647103.

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Field Apps

DitwanP:
🍠 Verified locally on `main`

PR #8541, opened 2024-01-02
fix(list, list-item, list-item-group): honor hidden attribute on list-item and list-item-group

driskull:
**Related Issue:** #8539

## Summary

- Adds internal `filterHidden` property to `list-item` and `list-item-group`
  - This property is used to hide an item when it is filtered out instead of the native `hidden` attribute
  - This allows the `hidden` attribute to be set by an end user
- Renames internal vars
- Adds screenshot tests
- Add e2e tests
- Added a developer discussion topic to consider not modifying global `hidden` attribute on host elements.

### DIFF
--- a/packages/calcite-components/src/components.d.ts
+++ b/packages/calcite-components/src/components.d.ts
@@ -2846,6 +2846,10 @@ export namespace Components {
          */
         "dragSelected": boolean;
         /**
+          * Hides the component when filtered.
+         */
+        "filterHidden": boolean;
+        /**
           * The label text of the component. Displays above the description text.
          */
         "label": string;
@@ -2902,6 +2906,10 @@ export namespace Components {
           * When `true`, interaction is prevented and the component is displayed with lower opacity.
          */
         "disabled": boolean;
+        /**
+          * Hides the component when filtered.
+         */
+        "filterHidden": boolean;
         /**
           * The header text for all nested `calcite-list-item` rows.
          */
@@ -4526,7 +4534,7 @@ export namespace Components {
          */
         "scale": Scale;
         /**
-          * Specifies the component's selected tab-title.
+          * Specifies the component's selected `calcite-tab-title`.
           * @readonly
          */
         "selectedTitle": HTMLCalciteTabTitleElement;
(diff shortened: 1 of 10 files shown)


driskull:
These other components modify the hidden attribute of host elements:

```
src/components/combobox/combobox.tsx:
  1041          const hidden = !find(item, filteredData);
  1042:         item.hidden = hidden;
  1043          const [parent, grandparent] = item.ancestors;

  1045          if (find(parent, filteredData) || find(grandparent, filteredData)) {
  1046:           item.hidden = false;
  1047          }

  1049          if (!hidden) {
  1050:           item.ancestors.forEach((ancestor) => (ancestor.hidden = false));
  1051          }

src/components/flow/flow.tsx:
  164          itemNode.showBackButton = itemNode === activeItem && newItemCount > 1;
  165:         itemNode.hidden = itemNode !== activeItem;
  166        });

src/components/stepper/stepper.tsx:
  412          if (index !== activePosition) {
  413:           item.hidden = true;
  414          } else {
  415:           item.hidden = false;
  416            item.multipleViewMode = false;

  426        this.items.forEach((item: HTMLCalciteStepperItemElement) => {
  427:         item.hidden = false;
  428          item.multipleViewMode = true;

src/components/table/table.tsx:
  355        const inView = rowPos >= this.pageStartRow && rowPos < this.pageStartRow + this.pageSize;
  356:       row.hidden = this.pageSize > 0 && !inView && !this.footRows.includes(row);
  357        row.lastVisibleRow =

src/components/tip-manager/tip-manager.tsx:
  201        tip.selected = isSelected;
  202:       tip.hidden = !isSelected;
  203      });
```

alisonailea (approved):

jcfranco:
Nice! I like this approach very much since it frees up `hidden` for developers. Naming-wise, maybe `filtered` would avoid reading the prop as affecting a filter in the component? 

I think we could extend this to other filtering components once we add a convention, and possibly some shareable utils, for it.

driskull:
> Nice! I like this approach very much since it frees up hidden for developers. Naming-wise, maybe filtered would avoid reading the prop as affecting a filter in the component?

Wouldn't `filtered` mean visible though?

driskull:
Added issue: https://github.com/Esri/calcite-design-system/issues/8623

PR #8542, opened 2024-01-03
ci(update-doc): build design tokens to prevent errors

benelan:
## Summary

Resolves an error in the GitHub Action that updates the component readmes, which was due to not building the design tokens before building the components.

error: https://github.com/Esri/calcite-design-system/actions/runs/7381163545/job/20079386682#step:4:178


driskull (commented):
👍 Looks good.  Thanks @benelan

Issue #7866, opened 2023-09-22
Syntax error setting `Infinity` on number input


nwhittaker:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Setting the value of a `<calcite-input-number>` input to `Infinity` generates a console error:

```
SyntaxError: Cannot convert Infinity0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000 to a BigInt (at p-a48cbcc5.js:6:798)
    at BigInt (<anonymous>)
    at new a (p-a48cbcc5.js:6:798)
    at p-a48cbcc5.js:6:2771
    at p-a48cbcc5.js:6:2840
    at f (p-a48cbcc5.js:6:3094)
    at $ (p-a48cbcc5.js:6:2646)
    at K.setNumberValue (p-ceeae984.entry.js:6:4983)
    at K.valueWatcher (p-ceeae984.entry.js:6:6697)
    at p-5f56ab0e.js:6:8473
    at Array.map (<anonymous>) 
```

### Expected Behavior

No error is generated, with possibly the input's value showing as `"Infinity"`.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/jOXZVGJ

### Reproduction Steps

1. Visit the code sample and open the console
2. Click the **Set Infinity** button and see error logged to console

### Reproduction Version

1.8.0

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Field Apps

geospatialem:
@nwhittaker Thanks for filing the above, could you speak more to the use case of adding infinity in, and expecting an error in the log as a result?

ethanbdev:
We ran into this too - I think since `Infinity` is a valid number in javascript, its reasonable for the prop to be set to this via jsx

nwhittaker:
>…could you speak more to the use case of adding infinity in…

The non-reduced test case is when the input is used in a data-down, action-up flow:
1. The numeric input is rendered with a value from a data model.
2. The user enters a number larger than `Number.MAX_VALUE`.
3. The input's change handler uses `Number(input.value)` to coerce a number which resolves to `Infinity`.
4. This value then gets propagated back to the data model.
[...]

PR #8547, opened 2024-01-03
fix(input, input-number): support setting value property to Infinity

benelan:
**Related Issue:** #7866

## Summary

- Setting the value of `calcite-input` and `calcite-input-number` to `Infinity` or `-Infinity` no longer causes errors
- `Infinity` or `-Infinity` will be displayed in the input.
- Entering additional characters is prevented, and pressing Backspace or Delete will clear the value.


benelan:
We might want to wait for @jcfranco on this one. I'm not sure if he'll want us to translate `Infinity` or not. FWIW we don't translate the `e` in exponential numbers.

packages/calcite-components/src/components/input/input.e2e.ts
```diff
@@ -1922,7 +1943,7 @@ describe("calcite-input", () => {
     expect(await button.getProperty("disabled")).toBe(true);
     expect(await input.getProperty("disabled")).toBe(false);
 
-    await input.setProperty("disabled", true);
+    input.setProperty("disabled", true);
```
jcfranco, line 1946: ✨🏆✨

jcfranco (approved):
♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️
♾️♾️😎😎😎♾️😎♾️♾️♾️♾️♾️😎♾️😎😎😎😎♾️😎😎😎😎♾️😎😎😎😎😎♾️😎♾️
♾️😎♾️♾️♾️♾️😎♾️♾️♾️♾️♾️😎♾️😎♾️♾️♾️♾️😎♾️♾️♾️♾️♾️♾️😎♾️♾️♾️😎♾️
♾️♾️😎😎♾️♾️😎♾️♾️😎♾️♾️😎♾️😎😎😎♾️♾️😎😎😎♾️♾️♾️♾️😎♾️♾️♾️😎♾️
♾️♾️♾️♾️😎♾️😎♾️😎♾️😎♾️😎♾️😎♾️♾️♾️♾️😎♾️♾️♾️♾️♾️♾️😎♾️♾️♾️♾️♾️
♾️😎😎😎♾️♾️♾️😎♾️♾️♾️😎♾️♾️😎😎😎😎♾️😎😎😎😎♾️♾️♾️😎♾️♾️♾️😎♾️
♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️♾️

Since this is displaying a special value, could you add a screenshot test for all infinity cases?

> We might want to wait for @jcfranco on this one. I'm not sure if he'll want us to translate Infinity or not. FWIW we don't translate the e in exponential numbers.

I think we can proceed w/ `Infinity` for the time being as, IMO, it's an edge case. From a design and localization-perspective, would displaying `∞` vs `Infinity` (value setting/event emitting would still use `Infinity`) be a good alternative that doesn't need translation? cc @SkyeSeitz @ashetland @annierm18

benelan:
@jcfranco is this okay to merge or should we wait for i18n feedback from @annierm18 regarding:

> From a design and localization-perspective, would displaying `∞` vs `Infinity` (value setting/event emitting would still use `Infinity`) be a good alternative that doesn't need translation?

ashetland:
Don't have a strong opinion on this one. `Infinity` _might_ be more clear, but defer to @SkyeSeitz and @annierm18.

SkyeSeitz:
I don't see a large issue with either option so I don't have a strong opinion here either, but the symbol does seem to read clear enough and may worth it if translation is an issue, especially if we have other similar symbols in use. 
image

jcfranco:
> @jcfranco is this okay to merge

Yus. Sorry for not being explicit about the `∞` suggestion being a follow-up.

packages/calcite-components/src/components/input/input.tsx
```diff
@@ -128,8 +128,7 @@ export class Input
    *
    * When not set, the component will be associated with its ancestor form element, if any.
    */
-  @Prop({ reflect: true })
-  form: string;
+  @Prop({ reflect: true }) form: string;
```
benelan, line 131: @jcfranco do you know why this keeps ending up on a new line? Lol

jcfranco: It looks like it's the result of my habit of placing decorators above of their target and https://github.com/stencil-community/stencil-eslint/blob/main/docs/decorators-style.md not auto fixing for some reason. Does it auto fix this for you when explicitly running the linter?

PR #8550, opened 2024-01-03
docs(various components): consistent api description refs

DitwanP:
**Related Issue:** #7071

## Summary
Updates doc consistency across i and l-named components defined in the above issue for props, events, methods, and css vars, including:

- `input-date-picker`
- `input-time-zone`
- `list-item-group`
- `list-item`
- `list`

cc @geospatialem


packages/calcite-components/src/components/list/list.tsx
```diff
@@ -189,7 +189,11 @@ export class List
   @Prop({ mutable: true }) selectedItems: HTMLCalciteListItemElement[] = [];
 
   /**
-   * Specifies the selection mode - `"multiple"` (allow any number of selected items), `"single"` (allow one selected item), `"single-persist"` (allow one selected item and prevent de-selection), or `"none"` (no selected items).
+   * Specifies the selection mode of the component:
+   * `"multiple"` allow any number of selected items,
+   * `"single"` allow only one selected item,
+   * `"single-persist"` allow one selected item and prevent de-selection,
+   * `"none"` does not allow any selections.
```
geospatialem, line 196: Saw this looks similar to your recent PR, but it differs from some of the other components, such as `accordion` and `table`. WDYT is the best approach for the `selectionMode` prop across components? Maybe this PR should include an update across all components for consistency based on what we think works the best comprehensively.

geospatialem (commented):
Have a few comments, and maybe we can work towards a standard for the `selectionMode` property across the design system. WDYT?

packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -229,7 +229,7 @@ export class ListItem
   //--------------------------------------------------------------------------
 
   /**
-   * Emits when the item's content is selected.
+   * Fires when the item's content is selected.
```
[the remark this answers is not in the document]
DitwanP: Yeah I think that would be fine.

packages/calcite-components/src/components/list/list.tsx
DitwanP: I was thinking about this as well actually, was going to probably make a separate PR for it but I can include it here since its essentially the same kind of changes.

As for best practice, I looked into it and there's not really anything definitive, I think what we have is probably fine and we just need to better align the wording across the different components since they should pretty much be the same. 

One thing that was bothering me is the way these types of descriptions looked, the way I see it in code wasn't reflected in how it ends up being formatted on the site. I'm talking about each value being on a new line which to me makes it way easier to read. 

So there's two ways to actually get it to make a new line. One way is to add two spaces and the end of each line and the other way is to separate the lines by an empty line. I think the two spaces looks cleaner but its easy to not realize that they are even present in the code. 

The two spaces at the end of each line would look identical to what you see now but here's what the other way looks like:
Screenshot 2024-01-04 at 9 54 04 AM
(⬆️ I think this way is probably the one to go with IMO)


Both of them results in the description looking like this:image

Instead of what they look like now: image

Now I don't know if there will be a change between what we see and how it ends up on the doc site but assuming it's the same formatting, I think the individual line one looks better but LMKWYT.

DitwanP: Also I was thinking they would actually look like this: Screenshot 2024-01-04 at 10 12 40 AM

Where each value has a ":" and has it's description as a standalone sentence.

geospatialem: > I was thinking about this as well actually, was going to probably make a separate PR for it but I can include it here since its essentially the same kind of changes.

Awesome! 
✨ 💪🏻✨  



> One thing that was bothering me is the way these types of descriptions looked, the way I see it in code wasn't reflected in how it ends up being formatted on the site. I'm talking about each value being on a new line which to me makes it way easier to read.

Agreed 💯

> Now I don't know if there will be a change between what we see and how it ends up on the doc site but assuming it's the same formatting, I think the individual line one looks better but LMKWYT.

There may be a few changes, albeit minor, but here are some examples for context from [`accordion`](https://developers.arcgis.com/calcite-design-system/components/accordion/), [`combobox`](https://developers.arcgis.com/calcite-design-system/components/combobox), and [`list`](https://developers.arcgis.com/calcite-design-system/components/list/):

image
image
image

There are a few other properties throughout the design system that have a similar structure as well, so we may want to also change those depending on the change, such as removing the parenthesis `"("` and `")"` characters, such as [note: not a comprehensive list]:
- `block-section`'s `toggleDisplay`
- `combobox`'s `selectionDisplay`
- `list`'s `selectionAppearance`
- `shell-panel`'s `displayMode`

> Also I was thinking they would actually look like this: Screenshot 2024-01-04 at 10 12 40 AM
> 
> Where each value has a ":" and has it's description as a standalone sentence.

While that might be the case in the repo/typings, the doc site would look a bit different, and would defer to keeping the doc site look and feel consistent over the repo to ensure consistency for our users there, which is usually the first place folks navigate to.


> The two spaces at the end of each line would look identical to what you see now but here's what the other way looks like:
> Screenshot 2024-01-04 at 9 54 04 AM
> (⬆️ I think this way is probably the one to go with IMO)

With the above in mind, your proposal above looks fantastic. 🌟 It would be easier for folks to read, and understand within the repo/typings and on the doc site where the parenthesis aren't present. 🏆

You could open a new PR to accommodate the additional properties with parenthesis mentioned above (and additional ones identified), or use this existing PR for the new consistency - whichever you prefer.

With the above changes, the PR has changed a bit, WDYT about changing the title to something more generic like `(docs) consistent api description refs` or similar?

packages/calcite-components/src/components/accordion/accordion.tsx
```diff
@@ -40,8 +40,13 @@ export class Accordion {
...
   /**
-   * Specifies the selection mode - `"multiple"` (allow any number of open items), `"single"` (allow one open item),
-   * or `"single-persist"` (allow and require one open item).
+   * Specifies the selection mode of the component:
+   *
+   * `"multiple"`: Allows any number of selections.
+   *
+   * `"single"`: Allows only one selection.
+   *
+   * `"single-persist"`: Allows one selection and prevents de-selection.
```
geospatialem, line 49: I think based on [your comment here](https://github.com/Esri/calcite-design-system/pull/8550#discussion_r1442091135), this would be modified to exclude the `:` between the property values, and each property would have commas between them?

> The two spaces at the end of each line would look identical to what you see now but here's what the other way looks like:
> Screenshot 2024-01-04 at 9 54 04 AM
> (⬆️ I think this way is probably the one to go with IMO)

packages/calcite-components/src/components/chip-group/chip-group.tsx
```diff
@@ -51,10 +51,14 @@ export class ChipGroup implements InteractiveComponent {
...
-   * `"single-persist"` allow and require one selected item,
-   * `"none"` does not allow any selections.
+   *
+   * `"multiple"`: Allows any number of selections.
+   *
+   * `"single"`: Allows only one selection.
+   *
+   * `"single-persist"`: Allows one selection and prevents de-selection.
+   *
+   * `"none"`: Does not allow any selections.
```
geospatialem, line 61: Same as the comment above.

packages/calcite-components/src/components/combobox-item/combobox-item.tsx
```diff
@@ -84,11 +84,15 @@ export class ComboboxItem implements ConditionalSlotComponent, InteractiveCompon
...
-   * - "ancestors" is like multiple, but shows ancestors of selected items as selected, with only deepest children shown in chips.
+   * Specifies the selection mode of the component:
+   *
+   * `"multiple"`: Allows any number of selections.
+   *
+   * `"single"`: Allows only one selection.
+   *
+   * `"single-persist"`: Allows one selection and prevents de-selection.
+   *
+   * `"ancestors"`: Allows multiple selections, but shows ancestors of selected items as selected, with only deepest children shown in chips.
```
geospatialem, line 95: Refer to the comment above.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -218,11 +218,15 @@ export class Combobox
...
-   * `"ancestors"` is like multiple, but shows ancestors of selected items as selected, with only deepest children shown in chips.
+   * Specifies the selection mode of the component:
+   *
+   * `"multiple"`: Allows any number of selections.
+   *
+   * `"single"`: Allows only one selection.
+   *
+   * `"single-persist"`: Allows one selection and prevents de-selection.
+   *
+   * `"ancestors"`: Allows multiple selections, but shows ancestors of selected items as selected, with only deepest children shown in chips.
```
geospatialem, line 229: Refer to the comment above.

packages/calcite-components/src/components/dropdown-group/dropdown-group.tsx
```diff
@@ -43,10 +43,13 @@ export class DropdownGroup {
...
-   * `"multiple"` allows any number of selected items,
-   * `"single"` allows only one selection,
-   * `"none"` doesn't allow for any selection.
+   * Specifies the selection mode of the component:
+   *
+   * `"multiple"`: Allows any number of selections.
+   *
+   * `"single"`: Allows only one selection.
+   *
+   * `"none"`: Does not allow any selections.
```
geospatialem, line 52: Refer to the comment above.

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -189,7 +189,15 @@ export class List
...
-   * Specifies the selection mode - `"multiple"` (allow any number of selected items), `"single"` (allow one selected item), `"single-persist"` (allow one selected item and prevent de-selection), or `"none"` (no selected items).
+   * Specifies the selection mode of the component:
+   *
+   * `"multiple"`: Allows any number of selections.
+   *
+   * `"single"`: Allows only one selection.
+   *
+   * `"single-persist"`: Allows one selection and prevents de-selection.
+   *
+   * `"none"`: Does not allow any selections.
```
geospatialem, line 200: Refer to the comment above.

packages/calcite-components/src/components/table/table.tsx
```diff
@@ -80,7 +80,15 @@ export class Table implements LocalizedComponent, LoadableComponent, T9nComponen
...
-  /** Specifies the selection mode - `"none"` (no `calcite-table-row` selections), `"single"` (allow one `calcite-table-row` selection), or `"multiple"` (allow any number of `calcite-table-row` selections). */
+  /**
+   * Specifies the selection mode of the component:
+   *
+   * `"multiple"`: Allows any number of selections.
+   *
+   * `"single"`: Allows only one selection.
+   *
+   * `"none"`: Does not allow any selections.
+   */
```
geospatialem, line 91: Refer to the comment above.

packages/calcite-components/src/components/accordion/accordion.tsx
DitwanP: No the first comment I made with screenshot was what it looked like before I formatted it to look the way it looks now. I figured that the format with the commas between them worked better if they weren't going to be on their own lines.

I don't have a strong opinion either way so let me know which way you think looks best or makes the most sense.

geospatialem: > No the first comment I made with screenshot was what it looked like before I formatted it to look the way it looks now. I figured that the format with the commas between them looked better when the values were listed in a regular sentence format.

Gotcha, that's my mistake. I thought it was the new format proposal.

> I don't have a strong opinion either way so let me know which way you think looks best or makes the most sense.

I think the format mentioned in the picture seems the most clear, both from a repo/typings and doc site perspective - as long as you agree. So, something like:

```ts
/*
* Specifies the selection mode of the component, where:
* `"multiple"` allows any number of selected items,
* `"single"` allows only one selected item,
* `"single-persist"` allows one selected item and prevent de-selection, and
* `"none"` does not allow any selections.
*/
```

Snippets from the doc site and repo/typings projected below:

image

image

packages/calcite-components/src/components/list/list.tsx
DitwanP: I think another PR for the additional properties makes sense to keep this one from getting too big.

geospatialem (approved):
Awesome - LGTM!

🌟🚀🌟

Issue #7965, opened 2023-10-09
[input-date-picker] Using range, 2nd input's dropdown icon shows first date-picker 


arjanvanzutphen:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Clicking on the second dropdown icon (marked yellow) shows the date-picker underneath the first input element
![image](https://github.com/Esri/calcite-design-system/assets/25850072/847b18d9-5270-4b5b-a548-ed9a21e889f4)

Note that clicking on the second input-date-picker element itself shows the date-picker underneath the second input element.
Behavior occurs with vertical and horizontal layout.

### Expected Behavior

Clicking on the second dropdown icon (marked yellow) in the second input element shows the first date-picker underneath

### Reproduction Sample

See documentation (https://developers.arcgis.com/calcite-design-system/components/input-date-picker/)

### Reproduction Steps

1. Go the the [input-date-picker](https://developers.arcgis.com/calcite-design-system/components/input-date-picker/) documentation sample
2. Enable the range property. A second input element for an end date appears
3. Make sure the input elements are not active. Click somewhere outside the input-date-picker element(s). Active elements are blue bordered, like below:
![image](https://github.com/Esri/calcite-design-system/assets/25850072/3cdbfc2b-7cac-4c49-a22c-a6d5a78baec2)
4. Click on the dropdown icon in the second input element.
5. The date-picker components will appear underneath the first input element.

### Reproduction Version

1.9.0

### Regression?

`1.3.0`, likely from https://github.com/Esri/calcite-design-system/pull/6715

### Priority impact

p4 - not time sensitive

### Impact

The end user must select the input element itself to have the date-picker component appear.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Professional Services - Services Delivery

abp6318:
Dropping a comment as we found this bug over on Hub Discovery as well and are hoping it can be bumped up in time sensitivity. We are looking to have this fixed for our current milestone. Also leaving this here: https://devtopia.esri.com/dc/hub/issues/8086

@thomas-hervey looks like this issue has been logged by another team already.

geospatialem:
[...]

PR #8554, opened 2024-01-04
fix(input-date-picker): ensure range icon toggles open corresponding date-picker

jcfranco:
**Related Issue:** #7965

## Summary

This fixes an issue that caused the start date picker to be toggled when the end date icon was clicked.


driskull (approved):
👍

Issue #4060, opened 2022-02-04
Expose CSS variable to style accordion item header in active state 


rslibed:
### Description

When an accordion item is selected within an accordion, provide a css variable to configure the color of just the accordion item header to more obviously indicate that it is selected. 

There appears to be `--calcite-accordion-item-background` but that seems to color the entire accordion background (header and contents).

### Acceptance Criteria

![image](https://user-images.githubusercontent.com/29421380/152454224-39300388-ae0c-4aab-a1cc-8e8f5f8862f8.png)

### Which Component

calcite-accordion-item

macandcheese:
Thoughts @asangma ? 

Before adding something like this I’d like to try making the active accordion item more obvious by default than providing this variable, we run into issues here as some accordion appearance types has no x-padding on the header.

geospatialem:
This also came up in a [Community post](https://community.esri.com/t5/calcite-design-system-questions/add-style-to-dynamically-added-calcite-accordion) on 9/21/2022 to modify the header color.

As mentioned in the description above the background can be changed. Additionally, a customized border and icon color (if present) can be accessed, but not the header or description text color.

A Codepen example for context: https://codepen.io/geospatialem/pen/dyezJQN

geospatialem:
While not related to `accordion-item`, a similar question was asked on [Community](https://community.esri.com/t5/calcite-design-system-questions/changing-style-of-collapsible-calcite-blocks/m-p/1260640#M330) on 2/22/2023 related to `block` headings. 

Perhaps a pattern we could consider across components with `heading` attributes?

macandcheese:
Still not sure we need to allow for this level of customization.

CharanGIS:
Hi, would it be better to have <calcite-blocks> within <calcite-accordion-item> similar to his [example](https://developers.arcgis.com/calcite-design-system/components/accordion/)

geospatialem:
Will revisit once tokens have been implemented to determine next feasible steps.

ashetland:
Agree with the above that the requested level of customization may not be advisable — Accordion Items can't be selected, only expanded or collapsed. It may be worth looking at so padding tweaks to make the expanded state more clear. See [Carbon's Accordion](https://carbondesignsystem.com/components/accordion/usage/#live-demo) as an example.

macandcheese:
Agree - tokens would allow this level of customization as well depending on granularity of token exposed per component.

geospatialem:
[...]

Issue #4012, opened 2022-01-26
calcite-accordion-item content padding


COV-GIS:
### Description

`calcite-accordion-item`'s shadow dom `div[class="accordion-item-content"]` has a rendered padding of `0 0.75rem 0.5rem 0.75rem`. The top padding of `0` puts any content right up against the bottom of the accordion header and the header focus can overlap the content by a pixel or so.

### Acceptance Criteria

~~Put some space between `calcite-accordion-item` header and content.~~
See [Calcite Design Solution](https://github.com/Esri/calcite-design-system/issues/4012#:~:text=No%20response-,Calcite%20Design%20Solution,-Apply%20default%20padding) below

### Which Component

`calcite-accordion-item`

## Calcite Design Solution

- [ ] Apply default padding to Accordion Item's content slot to match the header padding at each scale.
~~Add separate CSS vars for the header area and the content area, per [comment](https://github.com/Esri/calcite-design-system/issues/4012#issuecomment-1367487005)~~
- [ ] Add CSS var to adjust the space around the content slot

Medium scale example:
![CleanShot 2024-09-19 at 16 39 19@2x](https://github.com/user-attachments/assets/99ebe317-69a3-48bb-9dcf-1b16cf88d8a3)

paulcpederson:
I also ran into this. I tried to use the `--calcite-accordion-item-padding` CSS var but it still doesn't allow adding padding to the top of the content. Further, it also applies the padding to the item's title/header area. Would be nice to make this padding configurable and also allow setting just the content padding.

ashetland:
Design recommendations add to description above.

geospatialem:
This was not addressed in https://github.com/Esri/calcite-design-system/pull/9861, will need to be a follow-up consideration for design tokens.

alisonailea:
Ah! Got it. Will address ASAP

geospatialem:
The above will have two different solutions:
1. **Phase I: Component tokens** to support user-defined spacing via CSS variables (targeted in September 2024)
2. **Phase II: Default spacing** which will alter the component's default spacing. Added the `visual changes` label as this will have visual changes to the component (targeted in November 2024)

ashetland:
[...]

PR #8559, opened 2024-01-05
feat(accordion): add component tokens for accordion and accordion-item

alisonailea:
**Related Issue:** #4060 #4012 #7180

## Summary

Adds component tokens within the SCSS file for accordion and accordion item.

### DIFF
--- a/packages/calcite-components/src/components/accordion-item/accordion-item.scss
+++ b/packages/calcite-components/src/components/accordion-item/accordion-item.scss
@@ -1,3 +1,32 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden
+ *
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/accordion-item/accordion-item.scss
```diff
@@ -1,3 +1,32 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden
+ *
+ * @prop --calcite-accordion-item-background: Specifies the background color of the accordion item.
+ * @prop --calcite-accordion-item-space: Specifies the base spacing multiplier for padding and margin of the accordion item.
```
macandcheese, line 7: Do we want to expose `-space` and `-size` here? The rest look good IMO to be public.

PR #8561, opened 2024-01-05
feat(radio-button-group, segmented control): add validationMessage, validationIcon, and status properties

benelan:
**Related Issue:** #8057

## Summary

Add `validationMessage`, `validationIcon`, and `status` properties to `calcite-radio-button-group` and `calcite-segmented-control` for form validation.


jcfranco (approved):
📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝
📝✅📝📝📝✅📝✅✅✅✅📝✅📝📝📝✅✅✅📝✅✅✅📝📝✅📝
📝✅📝📝📝✅📝✅📝📝✅📝✅📝📝📝📝✅📝📝✅📝📝✅📝✅📝
📝✅📝📝📝✅📝✅✅✅✅📝✅📝📝📝📝✅📝📝✅📝📝✅📝✅📝
📝📝✅📝✅📝📝✅📝📝✅📝✅📝📝📝📝✅📝📝✅📝📝✅📝📝📝
📝📝📝✅📝📝📝✅📝📝✅📝✅✅✅📝✅✅✅📝✅✅✅📝📝✅📝
📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝

jcfranco:
@benelan Once all components are updated, could we also update E2E tests to cover these props with `defaults` and `reflects`?

benelan:
> @benelan Once all components are updated, could we also update E2E tests to cover these props with `defaults` and `reflects`?

Should `validationMessage` and `validationIcon` be reflected?

jcfranco:
> Should validationMessage and validationIcon be reflected?

According to [our conventions](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/conventions/README.md#reflecting-to-attributes), `validationIcon` could be reflected.

benelan:
@jcfranco - I'm re-requesting review because there were a lot of changes since the first one. A couple notes:

- There is a [diff on the "Simple" radio-button-group story](https://www.chromatic.com/test?appId=6266d45509d7eb004aa254fb&id=659f4b5fe6fddfcc8aef0d02) because you can't set a `string` knob's value to `undefined`. I [set it to `""` instead](https://github.com/Esri/calcite-design-system/blob/benelan/radio-validation-message/packages/calcite-components/src/components/radio-button-group/radio-button-group.stories.ts), but that makes the prop truthy, which ends up matching [these CSS selectors](https://github.com/Esri/calcite-design-system/blob/8be513b40a54b882ae8889e5d6a42ab98a55e378/packages/calcite-components/src/components/radio-button-group/radio-button-group.scss#L31-L37). Is there a better way to do this or is it okay as is?

- I wanted to do something like the following snippet, but [`--calcite-input-message-spacing-value`](https://github.com/Esri/calcite-design-system/blob/8be513b40a54b882ae8889e5d6a42ab98a55e378/packages/calcite-components/src/components/input-message/input-message.scss#L6) doesn't seem to have any effect, [even when setting it to `500px` for testing](https://codepen.io/benesri/pen/OJqRxNm?editors=1100). Unless I'm doing something wrong?
  ```css
  :host([validation-message]) {
    --calcite-input-message-spacing-value: calc(var(--calcite-label-margin-bottom) * -1);
  }
  ```

jcfranco:
> There is a [diff on the "Simple" radio-button-group story](https://www.chromatic.com/test?appId=6266d45509d7eb004aa254fb&id=659f4b5fe6fddfcc8aef0d02) because you can't set a string knob's value to undefined. I [set it to "" instead](https://github.com/Esri/calcite-design-system/blob/benelan/radio-validation-message/packages/calcite-components/src/components/radio-button-group/radio-button-group.stories.ts), but that makes the prop truthy, which ends up matching [these CSS selectors](https://github.com/Esri/calcite-design-system/blob/8be513b40a54b882ae8889e5d6a42ab98a55e378/packages/calcite-components/src/components/radio-button-group/radio-button-group.scss#L31-L37). Is there a better way to do this or is it okay as is?

Unfortunately, there doesn't seem to be a clean way to set `undefined` on knobs. Is this the only story that is showing different styling due to the presence of `validation-message`? If so, I think we could do one of the following:

1. keep as is
2. create a separate story to set validation props
3. remove validation props from simple story

It'd be good to add a note in the story regardless of the option.

> I wanted to do something like the following snippet, but --calcite-input-message-spacing-value doesn't seem to have any effect, even when setting it to 500px for testing. Unless I'm doing something wrong?

It looks like the label CSS prop [doesn't have a default value and label's bottom margin will use a fall back in this case](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/label/label.scss#L34).

jcfranco (approved):
💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊
💊🔘💊💊💊🔘💊💊🔘🔘💊💊🔘🔘🔘💊💊🔘🔘🔘💊🔘🔘🔘🔘💊🔘💊
💊🔘🔘💊💊🔘💊🔘💊💊🔘💊💊🔘💊💊🔘💊💊💊💊🔘💊💊💊💊🔘💊
💊🔘💊🔘💊🔘💊🔘💊💊🔘💊💊🔘💊💊🔘💊💊💊💊🔘🔘🔘💊💊🔘💊
💊🔘💊💊🔘🔘💊🔘💊💊🔘💊💊🔘💊💊🔘💊💊💊💊🔘💊💊💊💊💊💊
💊🔘💊💊💊🔘💊💊🔘🔘💊💊🔘🔘🔘💊💊🔘🔘🔘💊🔘🔘🔘🔘💊🔘💊
💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊💊

benelan:
> It looks like the label CSS prop [doesn't have a default value and label's bottom margin will use a fall back in this case](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/label/label.scss#L34).

`--calcite-input-message-spacing-value` doesn't have any effect even when removing the label in the codepen I linked in my previous comment - https://codepen.io/benesri/pen/OJqRxNm?editors=1100

Any other ideas?

jcfranco:
I think there's a specificity issue in `input-message`: `margin-block-start` is set to the variable's value [here](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/input-message/input-message.scss#L20), but is overridden [here](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/input-message/input-message.scss#L56-L62) (bypassing the prop).

macandcheese:
> I think there's a specificity issue in `input-message`: `margin-block-start` is set to the variable's value [here](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/input-message/input-message.scss#L20), but is overridden [here](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/input-message/input-message.scss#L56-L62) (bypassing the prop). You could [override it directly](https://codepen.io/jcfranco/pen/qBvawML), but might be good to fix at the `input-message` level (in a separate PR).

For consistency across forms - does this space between the input and the input message need to be configurable at all?

jcfranco:
> For consistency across forms - does this space between the input and the input message need to be configurable at all?

image

We can keep an eye out for requests.

benelan:
After additional testing I found that setting `--calcite-label-margin-bottom: 0` caused a spacing regression when the horizontal layout wrapped.

Now that #8592 is installed I changed the code to set a negative margin-top on the input message to inverse the label's margin-bottom. @ashetland can you please review the snapshots when you get a chance? Thanks!

benelan:
> After additional testing I found that setting `--calcite-label-margin-bottom: 0` caused a spacing regression when the horizontal layout wrapped.
> 
> Now that #8592 is installed I changed the code to set a negative margin-top on the input message to inverse the label's margin-bottom. @ashetland can you please review the snapshots when you get a chance? Thanks!

I was using incorrect spacing token levels and after fixing it the diffs disappeared in the most recent Chromatic build. Here are the two relevant stories:

- [radio-button-group](https://www.chromatic.com/component?appId=6266d45509d7eb004aa254fb&csfId=components-controls-radio-radio-button-group--validation-message-test-only&buildNumber=3490&k=65a1dc9570ad7dd8388b8a26-1200px-interactive-true&h=6&b=-2)
- [segmented-control](https://www.chromatic.com/component?appId=6266d45509d7eb004aa254fb&csfId=components-controls-radio-segmented-control--validation-message-test-only&buildNumber=3490&k=65a1dc9570ad7dd8388b8a37-1200px-interactive-true&h=2&b=-2)

ashetland:
I think those two stories are looking good. Are these the only ones affected by this or were there others? 

@jcfranco I think the longterm goal is to remove the Label margin altogether, right? If so, when we build label directly into components we'll likely need to reevaluate/refactor how this spacing is achieved on some components.

benelan:
@ashetland the only other one that changed (other than the usual false positives) is this one:

https://www.chromatic.com/component?appId=6266d45509d7eb004aa254fb&csfId=components-controls-radio-radio-button-group--dark-mode-rtl-test-only&buildNumber=3490&k=65a1dc9570ad7dd8388b8a25-1200px-interactive-true&h=4&b=-2

I made it vertical and added the validation message

Issue #8564, opened 2024-01-05
[List] Add a notice when list items are filtered down to no results.


sagewall:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Currently when a list component is filtered, and there are no results, there is no indication that the filtered list contains no items.
<img width="205" alt="image" src="https://github.com/Esri/calcite-design-system/assets/2636825/1de8c650-9bfd-407c-b079-420333ea02f6">

It would be nice if a calcite notice was presented to users when the list is filtered down to zero list items. Possibly something like this: https://codepen.io/mtb-alan/pen/JjzGvJP
<img width="214" alt="image" src="https://github.com/Esri/calcite-design-system/assets/2636825/f71c6ef8-3d10-4d0b-aaaf-f21dc388adb9">

cc @driskull @asangma @geospatialem

### Acceptance Criteria

Some sort of notification to users that no list items remain.

### Relevant Info

This has been requested as part of the ArcGIS Maps SDK Layer List widget redesign.

Related issue: https://github.com/Esri/calcite-design-system/issues/6133

### Which Component

List

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

macandcheese:
I feel like this should be up to the implementing team. 

Maybe a team wants to display a simple message-only Notice, but maybe it's a notice populated with title, description, icon, or it's another type of content entirely.

It should be pretty simple to have a conditionally displayed Notice based on the length of `filteredItems` property:

https://codepen.io/mac_and_cheese/pen/ZEPWbWL?editors=1000

macandcheese:
If we want the List / Filter component to handle the show / hide of this arbitrary content, maybe a "no results slot" that is conditionally shown?

driskull:
I like the idea of the slot since it would make it a lot easier to implement and still allow it to be customizable.

sagewall:
I like the slot idea as well

driskull:
To clarify, this slot is only for when a filter exists? or would we want it displayed anytime there are no items within the list?

driskull:
[...]

PR #8569, opened 2024-01-08
feat(list): Add "filter-no-results" slot to display content when no filtered items are shown

driskull:
**Related Issue:** #8564

## Summary

- Adds slot `"filter-no-results"`.
- When filter is enabled, a filter text is present and there are no items being shown, this slot will be displayed
  - The slot is for showing any warning information that there are no items displayed
  - Allows a user to customize any messaging for no results shown.
- Adds screenshot test
- Adds e2e test

### DIFF
--- a/packages/calcite-components/src/components/list/list.e2e.ts
+++ b/packages/calcite-components/src/components/list/list.e2e.ts
@@ -106,6 +106,17 @@ describe("calcite-list", () => {
         <calcite-list-item label="test" description="hello world"></calcite-list-item>
       </calcite-list>`,
     );
+    accessible(
+      html`<calcite-list filter-enabled filter-text="Bananas" selection-appearance="border" selection-mode="single">
+        <calcite-list-item label="Apples" value="apples"></calcite-list-item>
+        <calcite-list-item label="Oranges" value="oranges"></calcite-list-item>
+        <calcite-list-item label="Pears" value="pears"></calcite-list-item>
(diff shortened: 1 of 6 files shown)


packages/calcite-components/src/components/list/list.e2e.ts
```diff
@@ -530,6 +543,35 @@ describe("calcite-list", () => {
     expect(await list.getProperty("selectedItems")).toHaveLength(0);
   });
 
+  it("should show noFilterResults", async () => {
```
jcfranco, line 546: Nitpick: `noFilterResults` reads like a prop and does not match the slot name. WDYT about `"should show no-results content when filter does not match"`?

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -553,6 +557,16 @@ export class List
               <slot onSlotchange={this.handleDefaultSlotChange} />
             </tbody>
           </table>
+          <div
+            aria-live="polite"
+            class={CSS.noResultsContainer}
```
jcfranco, line 562: Is this class mainly used for testing purposes? If so, maybe we can use a test data attribute (`data-test-id="no-results-container"` would be compatible with [this proposed test pattern](https://github.com/Esri/calcite-design-system/pull/8171)).

jcfranco (approved):
Awesome! 

Before merging, can you update the description with the updated slot name?

Issue #7351, opened 2023-07-20
Bug(color-picker): Value does not change when using shift+up/down arrow


maxpatiiuk:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The color picker value updates correctly if using `up/down arrow keys` in any of the `r/g/b` or `h/s/v` input boxes, but the value does not update if using `shift+up/down arrow keys` (unless you press `up/down arrow keys` again without shift)

https://github.com/Esri/calcite-design-system/assets/40512816/d478574f-b845-4b1a-85e5-325b9e3b4eb5

### Expected Behavior

The value would update regardless if `shift` is pressed or not

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/color-picker/

### Reproduction Steps

1. Render calcite color picker `<calcite-color-picker />` (or use the official sample - https://developers.arcgis.com/calcite-design-system/components/color-picker/)
2. Click on any of the `r/g/b` or `h/s/v` input fields (like in the video)
3. Click `up/down arrow keys` and see that HEX value and previews update with each click
4. Click `shift + up/down arrow keys` and see that HEX value and previews do not update

---

This might be a separate issue, but while input fields, and hue slider support the `shift+up/down arrow keys` gesture, the `color-field` canvas (the rectangle at the top of the color picker) does not respect the shift key - the value changes by the same amount regardless if shift is pressed or not. The issue is that this is inconsistent - all but one component in color picker support the `shift` key

### Reproduction Version

1.4.x (May 2023)

### Relevant Info

Chrome 114 (latest version at the moment), on macOS 13.4.1 (the latest allowed by Esri)

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Maps SDK for JavaScript

DitwanP:
🍯 Verified locally on `main`

PR #8579, opened 2024-01-09
fix(color-picker): emit color change when nudging color channels by using the shift key

jcfranco:
**Related Issue:** #7351

## Summary

When color channels are nudged, the value is modified and set directly on input. This worked fine when using up/down arrows as change events are emitted by default, but for the modified nudging behavior (shift key), the modified input value logic was suppressing the change event.


driskull (approved):
👍 Nice

Issue #7221, opened 2023-06-26
Closing the active tab does not fire events for tab change


RSantosGIS:
### Actual Behavior

Originally documented in https://github.com/Esri/calcite-components/issues/7155 but broken into a smaller bug report at the request of @mpriour .

When you close the active tab, the active tab automatically changes (as intended) but this does not fire the onCalciteTabsActivate or onCalciteTabChange events for CalciteTabTitle and CalciteTabNav, respectively

### Expected Behavior

When you close the active tab and it moves you to a new tab, all of the events for tab navigation should fire.

### Reproduction Sample

https://codesandbox.io/s/new-feather-qsyw8j?file=/src/App.js

### Reproduction Steps

1. Open the dev console
2. Close the active tab - notice the console.log statements don't fire
3. click to navigate to another tab - notice that the console.log statements do fire

### Reproduction Version

1.4.2

### Priority impact

p3 - want for upcoming milestone

### Impact

This prevents ArcGIS studio from being able to utilize callbacks on tab navigation events reliably.  We have to do workarounds where tab closure triggers special handling to also consider navigation, rather than just have navigation events be handled by the actual navigation event handler on the component.

### Esri team

ArcGIS Knowledge

geospatialem:
Research to be conducted in August to determine the time estimate needed for a fix in a future milestone.

Elijbet:
> Research to be conducted in August to determine the time estimate needed for a fix in a future milestone.

**Research Summary**

The question revolves around whether the automatic movement of selection to the next tab after closing one should be considered a user-triggered activation or a programmatic action. 

Currently, no `TabsActivate` or `TabChange` events, either public or internal, are being fired as part of the `closable` logic. The suggestion is to emit public events `TabsActivate(userTriggered = true / TabChange(userTriggered = true)` related to close events based on how the tab was closed (user-triggered vs. programmatic). 

Here are some precedent examples where related events are emitted in response to user-initiated actions in other components:

- `pick/value-list` emits the remove item event and, after being removed, also emits the list change event.
- `combobox` emits filter and combobox change events when the value is cleared (there may be other similar examples).
- `list` emits a change event along with its item's select event.

It seems to make sense that emitting associated events should only occur when the user initiates an action that would trigger those events because internal events may not always be necessary; they have been used in specific cases as implementation details. 

[...]

PR #8582, opened 2024-01-10
feat(tabs): emit selection-related events when selection is modified after closing the selected tab

jcfranco:
**Related Issue:** #7221

## Summary

This will help users know when the selected `tab-title` and `tab` have changed after closing.


jcfranco:
@Elijbet Since the closing behavior affects `tabs` children, we should probably move, and increase coverage, all tab-closing-related tests from https://github.com/Esri/calcite-design-system/pull/6740/files#diff-c7f6cef044b115506885fe23d350cb1c87b2535c77d3888863efbcf37e7fe500 to `tabs.e2e.ts`.

packages/calcite-components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -396,6 +397,10 @@ export class TabNav {
     if (totalVisibleTabTitles === 1 && tabTitles[visibleTabTitlesIndices[0]].closable) {
       tabTitles[visibleTabTitlesIndices[0]].closable = false;
       this.selectedTabId = visibleTabTitlesIndices[0];
+
+      if (selectionModified) {
+        tabTitles[visibleTabTitlesIndices[0]].click(); // trigger selection-related events since they were modified after user interaction
```
driskull, line 402: nitpick: It seems a little odd to be handling this using a click since a click will bubble up and could potentially trigger other unknown things higher in the DOM.

Maybe in the future we could refactor this to handle the use case without having to trigger a click of an element? Just throwing thoughts out there :) I know we do use `.click()` in other places but its something we should think about possibly.

driskull (commented):
The code looks good 👍  APPROVED

packages/calcite-components/src/components/tab-nav/tab-nav.tsx
jcfranco: 100% agreed. I'll update this PR to add an internal method to only trigger the tab-related events and will create a refactor issue to introduce a tabs controller that `tabs` and its structure could share to make this cleaner. There's a lot of eventing going back and forth and it's getting quite complex.

jcfranco: Internal method added and `click` no longer used. Can you take another look? 🙇‍♂️

driskull (approved):
Much nicer! 🎉 🌮

Issue #8581, opened 2024-01-10
Custom tooltip for ListItem drag handle


annelfitz:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

We would like to be able to customize the drag handle tooltip in a calcite list item.

For example, in the Bookmarks widget, we were previously able to display a custom tooltip based on the bookmark title.
![Screenshot 2024-01-09 at 4 12 44 PM](https://github.com/Esri/calcite-design-system/assets/51344676/1e65082a-5cae-4154-bb68-518a3a2c061e)

Now we are in the process of refactoring the Bookmarks widget to use calcite components, taking advantage of the List component, but we are unable to provide that same experience in the tooltip. This enhancement seems like it would greatly help people using screen readers as they may need more information to know what they are dragging. The custom tooltip allows us to provide that information, instead of just the default `Drag handle` which is what the tooltip currently says.
![Screenshot 2024-01-09 at 4 15 44 PM](https://github.com/Esri/calcite-design-system/assets/51344676/0150a499-03b4-4c12-a453-b20168a0b180)

### Acceptance Criteria

- Allow dragHandle tooltip to be customizable

### Which Component

List Item

### Example Use Case

https://zrh-app-sd-2.esri.com/arcgis-play/a/afitz_jsapi/bookmarks-calcite-refactor-2

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

annelfitz:
cc @driskull @geospatialem

geospatialem:
This looks like a relatively low effort, to add the capability via the component's `messageOverrides` property. 

@annelfitz We just had our translation deadline pass this week, if we aimed for completion shortly after 24.R1 next month, would that work with the Maps SDK timeline for 24.R2?

annelfitz:
sure, sounds good to me. thanks @geospatialem !

driskull:
What if we just update the message to be `Move: {itemLabel}` or `Move or reorder: {itemLabel}`? Would that suffice?

Items can be dragged or moved via keyboard

geospatialem:
[...]

PR #8584, opened 2024-01-10
feat(handle, block, list-item): improve drag handle tooltip to include item label

driskull:
**Related Issue:** #8581

## Summary

- Renames `dragHandle` message to include `label` of the handle.
- Add new message for `dragHandleUntitled`.
- Update block to set label on handle.
- If handle is used without a label it will just say `Reposition`
- Add resources const for `SUBSTITUTION` vars.
- Add test


packages/calcite-components/src/components/handle/assets/handle/t9n/messages.json
```diff
@@ -1,5 +1,5 @@
 {
-  "dragHandle": "Drag handle",
+  "dragHandle": "Reposition or move: {itemLabel}",
```
macandcheese, line 2: Should we choose one of “Reposition” or “move” here?

driskull: We could just go with reposition. I was just thinking of sort vs move out

macandcheese (approved):

Issue #5548, opened 2022-10-24
Refactor to remove any instances of added hidden prop


Elijbet:
### Description

We shouldn't need to add or document this property since it’s already a global attribute common to all HTML elements.

### Proposed Advantages

Remove redundancy. 

### Which Component

Input
Input-number
Input-text

jcfranco:
Stencil v4 allows us to [watch global attributes](https://stenciljs.com/docs/reactive-data#watching-native-html-attributes), so we can leverage this wherever needed.

geospatialem:
Verified in `2.2.1-next.0` with https://codepen.io/geospatialem/pen/OJqmqEm, where the `hidden` attribute is supported as a global attribute.

PR #8591, opened 2024-01-12
refactor: drop `hidden` prop since it is a global attribute

jcfranco:
**Related Issue:** #5548  

## Summary

This removes `hidden` props since it's available on `HTMLElement` (base class for custom elements).


driskull (approved):
👍 Nice cleanup/refactoring!

Issue #8340, opened 2023-12-06
Focus traps don't work as expected with multiple modals


ellenupp:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When configuring a dashboard it's possible to have two modals open where the second stacks on top of the first (Ex: JS API sign-in dialog can open from a configuration dialog when adding a map/layer requiring authentication).

When opening the second calcite-modal on top of the other, the focus does not behave as expected. The second modal doesn't auto-focus on open and is not focus trapped. 

### Expected Behavior

The topmost modal should trap the focus (when no focus-trap-disabled present). Modal should auto-focus on open. 

Example where two modals are rendered and second one opens and closes on top. This works as expected. The issue seems to be when the second modal is dynamically added to the page. https://codepen.io/ellupp/pen/WNPxRwJ?editors=1000

### Reproduction Sample

https://codepen.io/ellupp/pen/LYqZxgx?editors=1000

### Reproduction Steps

1. Click button to open second modal. Note new modal is not auto-focused on open and is not focus trapped. 

### Reproduction Version

2.0.0

### Relevant Info

related issue #6659

### Priority impact

p4 - not time sensitive

### Impact

Minor but is an a11y concern

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Dashboards

geospatialem:
Would prefer to find a way to mitigate this across products where multiple modals are opened by default, such as with the Maps SDK and Dashboards. This may start to pop up even more in the coming weeks, so Calcite should address it ASAP.

dasa:
It might be worth considering implementing Modal using [HTMLDialogElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement) since that could avoid this issue.

geospatialem:
Verified with https://codepen.io/ellupp/pen/LYqZxgx in `2.2.0-next.22`:

![verify-modal-trap](https://github.com/Esri/calcite-design-system/assets/5023024/7f1345c7-4f6b-4633-9fdc-b768cb07656d)

PR #8593, opened 2024-01-12
fix(modal): ensure focus trapping in dynamically created, subsequently opened modals

jcfranco:
**Related Issue:** #8340 

## Summary

This fixes an issue caused by the internal `opened` watcher being invoked and not having the [`transitionEl` ref available](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/modal/modal.tsx#L519-L523), which would bypass the open/close component hooks and not [activate the focus trap](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/modal/modal.tsx#L489).

### DIFF
--- a/packages/calcite-components/src/components/modal/modal.tsx
+++ b/packages/calcite-components/src/components/modal/modal.tsx
@@ -52,7 +52,7 @@ import {
 } from "../../utils/t9n";
 import { ModalMessages } from "./assets/modal/t9n";
 
-import { getIconScale } from "../../utils/component";
+import { componentOnReady, getIconScale } from "../../utils/component";
 
 let totalOpenModals: number = 0;
 let initialDocumentOverflowStyle: string = "";
@@ -177,7 +177,7 @@ export class Modal
     setUpLoadableComponent(this);
     // when modal initially renders, if active was set we need to open as watcher doesn't fire
     if (this.open) {
-      requestAnimationFrame(() => this.openModal());
+      this.openModal();
     }
   }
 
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/modal/modal.tsx
```diff
@@ -532,7 +529,8 @@ export class Modal
     this.open = false;
   };
 
-  private openModal() {
+  private async openModal(): Promise<void> {
+    await componentOnReady(this.el);
```
driskull, line 533: @jcfranco is there any way we could put this in `activateFocusTrap` and/or `connectFocusTrap` so that this would happen for any component using focus trapping?

driskull (approved):
Looks good. Just wondering if we can move this logic into the focus trap component logic so that it won't occur for any component implementing focus trapping.

packages/calcite-components/src/components/modal/modal.tsx
jcfranco: It might not be that helpful since the main issue is related to watchers and refs. `focusTrapComponent` utils seem disconnected from this, right?

driskull: Yeah that's true. I was just thinking that this problem might arise in other components as well.

jcfranco: Agreed. Maybe we could wire it up if we made some additional changes to `focusTrapComponent`. 🤔❓

PR #8596, opened 2024-01-12
feat(inline-editable): add component tokens

jcfranco:
**Related Issue:** #7180 

## Summary

* `--calcite-inline-editable-background-color`
* `--calcite-inline-editable-size`

**Note**: this removes disabled styling for the editing button wrapper as it was not being applied (no border style set).

### DIFF
--- a/packages/calcite-components/src/components/inline-editable/inline-editable.scss
+++ b/packages/calcite-components/src/components/inline-editable/inline-editable.scss
@@ -4,7 +4,6 @@
  * These properties can be overridden using the component's tag as selector.
  *
  * @prop --calcite-inline-editable-background-color: Specifies the background color of the component.
- * @prop --calcite-inline-editable-size: Specifies the size of the component.
  */
 :host {
   --calcite-inline-editable-background-color: var(--calcite-ui-foreground-1);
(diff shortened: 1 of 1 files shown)


packages/calcite-components/src/components/inline-editable/inline-editable.scss
```diff
@@ -1,36 +1,37 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-inline-editable-background-color: Specifies the background color of the component.
+ * @prop --calcite-inline-editable-size: Specifies the size of the component.
```
macandcheese, line 7: Should this be public? I feel like this kind of conflates with `scale`. If it is public, should it be `height` as it is applied to `block-size` ?

jcfranco: Great catch. 🎣 Will fix.

alisonailea (approved):

jcfranco:
Closing in favor of updated PR: https://github.com/Esri/calcite-design-system/pull/8788.

PR #8599, opened 2024-01-13
feat(select): add component tokens

jcfranco:
**Related Issue:** #7180

## Summary

* `--calcite-select-background-color`
* `--calcite-select-background-color-hover`
* `--calcite-select-border-color`
* `--calcite-select-text-color`

and deprecates the following:

* `--calcite-select-font-size`
* `--calcite-select-spacing`

**Note**: `--calcite-select-space-inline` was removed from the doc as it is not defined nor used


packages/calcite-components/src/components/select/select.scss
```diff
@@ -3,104 +3,112 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
+ * @prop --calcite-select-background-color: The background color of the component.
+ * @prop --calcite-select-border-color: The border color of the component.
  * @prop --calcite-select-font-size: The font size of `calcite-option`s in the component.
```
macandcheese, line 8: I think we should keep the font-size and the space-inline internal. Also, adding a corner-radius that defaults to `sharp` and a box-shadow that defaults to `none` would help customization stories. Sorry, realize this is still in draft but I'll be OOO for a bit after today.

packages/calcite-components/src/components/select/select.scss
```diff
@@ -3,104 +3,112 @@
...
- * @prop --calcite-select-spacing: The padding around the selected option text.
+ * @prop --calcite-select-space-inline: The padding around the selected option text.
```
alisonailea, line 9: this can be an internal var

packages/calcite-components/src/components/select/select.scss
```diff
@@ -3,104 +3,112 @@
...
+ * @prop --calcite-select-text-color: The text color of the component.
  */
 
 :host {
+  --calcite-select-background-color: var(--calcite-color-foreground-1);
+  --calcite-select-border-color: var(--calcite-color-border-input);
+  --calcite-select-text-color: var(--calcite-color-text-2);
+
+  --calcite-internal-select-icon-space-inline: var(--calcite-color-foreground-1);
+  --calcite-internal-select-border-color: var(--calcite-select-border-color);
```
alisonailea, line 19: why is there an internal var and component token of the same name here?

alisonailea (changes requested):
Looks pretty good. I think the tokens can be simplified and `font-size` and `space` made internal

packages/calcite-components/src/components/select/select.scss
jcfranco: My bad, I shouldn't have added the internal one. Will update.

packages/calcite-components/src/components/select/select.scss
jcfranco: `--calcite-select-font-size` and `--calcite-select-spacing` were added prior to the component token effort. I'll need to deprecate.

alisonailea (approved):

alisonailea (changes requested):
There seems to be something off with the VRT. Looks like the font size changed?

jcfranco:
@alisonailea E2E tests are ✅ now.

alisonailea (approved):

Issue #8469, opened 2023-12-20
[Combobox] When `selectionDisplay="fit"/"single"`, unexpected `delete` key UX


macandcheese:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When Combobox has "fit" (and has "overflowed") or "single" value for `selectionDisplay` - there is unexpected behavior when using the `delete` key and an empty input. It behaves as if it is in "all" value - deleting unseen Chips and unselecting the items.

https://github.com/Esri/calcite-design-system/assets/4733155/ed55142b-4e05-4df6-8837-576c517f562c

### Expected Behavior

I would only expect that behavior in "all", since the `delete` deletes a visible chip. As Chip have `closable`, maybe this is unexpected and users should be able to "arrow left" into a Chip Group and use the Chip close button interaction / focus to close?

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/KKEPrYP?editors=1000

### Reproduction Steps

1. Visit Codepen
2. Select items
3. Use `delete` in an un-filled input

### Reproduction Version

2.0

### Priority impact

p2 - want for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

geospatialem:
Verified in `2.4.0` - adjusted the milestone to the January `2.2.0` release.

PR #8603, opened 2024-01-13
fix(combobox): only allow deleting visible chips with the keyboard

eriklharper:
**Related Issue:** #8469 

## Summary

This PR fixes a bug where pressing the Delete/Backspace key in combobox resulted in hidden chips being removed, which wasn't intuitive or expected.  The new behavior is deleting items is only allowed in `all` selection-display mode or `fit` when there are no overflowed chips.


packages/calcite-components/src/components/combobox/combobox.e2e.ts
```diff
@@ -1120,14 +1120,118 @@ describe("calcite-combobox", () => {
         chips = await page.findAll("#myCombobox >>> calcite-chip");
         expect(chips.length).toEqual(2);
       });
+    });
+  });
 
-      it("should delete last chip on Delete", async () => {
-        await element.click();
+  describe("deleting items with the keyboard", () => {
```
jcfranco, line 1126: I think this can go in the above describe block since it also falls within `keyboard interaction`.

packages/calcite-components/src/components/combobox/combobox.e2e.ts
```diff
@@ -1120,14 +1120,118 @@ describe("calcite-combobox", () => {
...
+      await item2.click();
+      await item3.click();
+
+      await element.click();
+      await element.press("Backspace");
+      expect((await element.getProperty("selectedItems")).length).toBe(3);
+    });
+
+    it("should delete last item on Delete in fit selection-display mode when there are no overflowed chips", async () => {
+      const page = await newE2EPage();
```
jcfranco, line 1211: It looks like there's a lot of common markup between the tests, so could you look at generalizing page setup logic in a `beforeEach` block to reduce duplication?

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -742,6 +742,12 @@ export class Combobox
         break;
       case "Delete":
       case "Backspace":
+        if (
+          this.selectionDisplay === "single" ||
```
jcfranco, line 746: For readability, can you store this value in a local self-documenting variable and use here?

jcfranco (approved):
Looking great, @eriklharper! 😎

Once the requested changes are made, this is good to merge!

PR #8605, opened 2024-01-13
feat(card): Add component tokens

macandcheese:
**Related Issue:** #7180 

## Summary

`--calcite-card-background-color`: Specifies the background color of the component.
`--calcite-card-border-color`: Specifies the border color of the component.
`--calcite-card-box-shadow`: Specifies the box shadow of the component.
`--calcite-card-corner-radius`: Specifies the corner radius of the component.

Adds a Chromatic test and a demo page example. Also replaces broken image links in local demo.


packages/calcite-components/src/components/card/card.scss
```diff
@@ -1,87 +1,93 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-card-background-color: Specifies the background color of the component.
+ * @prop --calcite-card-border-color: Specifies the border color of the component.
+ * @prop --calcite-card-box-shadow: Specifies the box shadow of the component.
+ * @prop --calcite-card-corner-radius: Specifies the corner radius of the component.
```
eriklharper, line 9: I feel like the names of CSS variables should mirror the web-platform names, so instead of `corner-radius` here, it should be `border-radius`.  That doesn't mean that the source token name can't still be called `corner-radius` in the JSON because when that source token is transformed to an iOS or Android platform variable, it will be named specific to that platform.  Since we're using the CSS terms `box-shadow`, `background-color`, `border-color`, `color`, etc. in the name of a css property, I would expect the name `border-radius` to be used in the name of the variable that assigns to the `border-radius` css property.

macandcheese: Will defer to @alisonailea on this one.

eriklharper: I just think that makes sense since the target audience for these variables are web developers, so we should use terms they are familiar with.

jcfranco: My 2 cents, I think we should follow the token names as it'll make it easier for documentation and maintenance. It also doesn't tie it to the implementation.

packages/calcite-components/src/components/card/card.scss
```diff
@@ -1,87 +1,93 @@
...
+ * @prop --calcite-card-subtitle-color: Specifies the color of the component's `"subtitle"` slot.
```
macandcheese, line 10: Do we want to provide these tokens for slotted content? We could just set defaults and the user could override themselves on their slotted elements.

macandcheese: Places we accept both a prop and a slot, it could make sense to have this. I could also see card having props in the future for title / subtitle, keeping the slot for more advanced cases, so maybe these are fine to leave.

packages/calcite-components/src/components/card/card.scss
alisonailea: @eriklharper while this particular instance of token names is web specific it would be a mistake to make naming decisions exclusively based on supporting a single platform given that we are already in discussions with teams building for multiple other platforms. Using the name `corner-radius`, it is still clear to everyone what we are discussing but does not limit token patterns to a web platform.

packages/calcite-components/src/components/card/card.scss
alisonailea: we should not create tokens for slotted elements.

packages/calcite-components/src/components/card/card.scss
```diff
@@ -1,87 +1,93 @@
...
+ * @prop --calcite-card-title-color: Specifies the color of the component's `"title"` slot.
```
alisonailea, line 11: remove this.

alisonailea (changes requested):
Looks good. Let's just remove the tokens for slotted content and allow those elements to style themselves.

macandcheese:
@alisonailea could you give this one another check after the last updates? Thanks!

alisonailea (approved):

alisonailea:
Looks good!

Issue #7425, opened 2023-08-02
Chip/Chip Group selection icon padding


ashetland:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When in `selection-mode="single" | "single-persist"`, there are a few errors in padding around the selection icon. 

### Expected Behavior

See this [Figma file](https://www.figma.com/design/HB1Ollf1D1Csy2PtXJ3Hai/Chip-selection-affordance-%26-padding?node-id=501-15580&m=dev) for the correct padding.

### Reproduction Sample

https://codepen.io/ashetland/pen/zYMLqyd?editors=1000

### Reproduction Steps

1. Open the codepen
2. Note padding around the selection icon

### Reproduction Version

1.4.3

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

Calcite (design)

ashetland:
Here's the design spec for this one https://www.figma.com/design/HB1Ollf1D1Csy2PtXJ3Hai/Chip-selection-affordance-%26-padding?node-id=501-15580&m=dev

Elijbet:
Closing, as this was address with [feat(chip): enhance multi-select group affordance](https://github.com/Esri/calcite-design-system/pull/9286#top) #9286.

Elijbet:
Closing, as this was address with [feat(chip): enhance multi-select group affordance](https://github.com/Esri/calcite-design-system/pull/9286#top) #9286.

Elijbet:
Closing, as this was address with https://github.com/Esri/calcite-design-system/pull/9286#top https://github.com/Esri/calcite-design-system/pull/9286.

PR #8607, opened 2024-01-13
feat(chip, chip group): Add component tokens

macandcheese:
Related Issue: https://github.com/Esri/calcite-design-system/issues/7180

Summary
`--calcite-chip-background-color`: Specifies the background color of the component.
`--calcite-chip-border-color`: Specifies the border color of the component.
`--calcite-chip-shadow`: Specifies the box shadow of the component.
`--calcite-chip-corner-radius`: Specifies the corner radius of the component.
`--calcite-chip-icon-color`: Specifies the color of the component's icon property.
`--calcite-chip-text-color`: Specifies the color of the component's `"default"` slot.
`--calcite-chip-close-icon-color`: Specifies the color of the component's `closable` icon.
`--calcite-chip-close-background-color`: Specifies the color of the component's `closable` background color.
 `--calcite-chip-close-background-color-hover`: Specifies the hover color of the component's `closable` background color.
`--calcite-chip-close-background-color-active`: Specifies the active color of the component's `closable` background color.
`--calcite-chip-close-background-color-focus`: Specifies the focus color of the component's `closable` background color.
`--calcite-chip-close-focus-outline-color`: Specifies the focus color of the component's `closable` background color.

Adds a Chromatic test and a demo page example. Also replaces broken image links in local demo.


packages/calcite-components/src/components/chip-group/chip-group.scss
```diff
@@ -6,19 +6,19 @@
   display: flex;
   inline-size: 100%;
   flex-wrap: wrap;
-  gap: theme("spacing.2");
+  gap: var(--calcite-spacing-sm);
```
macandcheese, line 9: @ashetland @SkyeSeitz this is a small change but one that prevents focus overlap at small scale. For now, this is not a documented public component.

packages/calcite-components/src/components/chip/chip.scss
```diff
@@ -1,73 +1,80 @@
...
   focus-base
   justify-center
-  border
-  border-solid
-  border-color-1
+  cursor-default
   box-border
   font-medium;
+  border: var(--calcite-border-width-sm) solid var(--calcite-chip-border-color, var(--calcite-color-border-3));
+  border-radius: var(--calcite-chip-corner-radius, 9999px);
```
macandcheese, line 70: This `9999px` needs to made into a token - in TW this is `rounded-full`

packages/calcite-components/src/components/chip/chip.scss
```diff
@@ -1,73 +1,86 @@
...
+ *
+ * @prop --calcite-chip-background-color: Specifies the background color of the component.
+ * @prop --calcite-chip-border-color: Specifies the border color of the component.
+ * @prop --calcite-chip-box-shadow: Specifies the box shadow of the component.
+ * @prop --calcite-chip-corner-radius: Specifies the corner radius of the component.
+ * @prop --calcite-chip-icon-color: Specifies the color of the component's icon property.
+ * @prop --calcite-chip-text-color: Specifies the color of the component's `"default"` slot.
+ * @prop --calcite-chip-close-icon-color: Specifies the color of the component's `closable` icon.
+ * @prop --calcite-chip-close-background-color: Specifies the color of the component's `closable` background color.
+ * @prop --calcite-chip-close-background-color-hover: Specifies the hover color of the component's `closable` background color.
```
alisonailea, line 14: Replace state specific tokens with a token reassignment in the `:host(:[state])`

packages/calcite-components/src/components/chip/chip.scss
```diff
@@ -1,73 +1,86 @@
...
+ * @prop --calcite-chip-close-background-color-active: Specifies the active color of the component's `closable` background color.
+ * @prop --calcite-chip-close-background-color-focus: Specifies the focus color of the component's `closable` background color.
+ * @prop --calcite-chip-close-focus-color: Specifies the focus color of the component's `closable` background color.
+ *
+ */
+
+/**
+  * Local props
+  * These properties are intended for internal component use only. It is not recommended that these properties be overwritten.
```
alisonailea, line 23: 🌟

packages/calcite-components/src/components/chip/chip.scss
```diff
@@ -78,7 +91,7 @@
 }
 
 :host([scale="s"]) .container.image--slotted {
-  padding-inline-start: calc(theme("spacing[0.5]") / 2);
+  padding-inline-start: calc(var(--calcite-internal-chip-spacing-s) / 2);
```
alisonailea, line 94: instead of adding style props to each rule based on :host props, can you reassign a single CSS variable instead?

alisonailea (changes requested):
Lift all tokens into `:host` and only apply each style prop to an HTML element once with a CSS variable (token) then reassign the variable when the host state changes.

packages/calcite-components/src/components/chip/chip.scss
macandcheese: I think that would prevent overrides from being possible.

This allows for "when the close button is hovered, change the close button background color", not "when the host element is hovered, change the close button background color"

macandcheese: Applies to other component with similar internal div as well.

alisonailea: It is still possible it just requires people writing overrides to add the override to each state as well. I went over the pros and cons of this with Franco and he agreed.

alisonailea: Sorry I forgot about internal component state.

alisonailea (approved):

packages/calcite-components/src/components/chip/chip.scss
macandcheese: Updated all the style-specific ones based on appearance / kind. The spacing ones will need to be revisited as they currently are not always based on reflected attributes, but also internal conditional classes. Should be addressed with https://github.com/Esri/calcite-design-system/issues/7425.

Issue #7576, opened 2023-08-22
tabs: focuses tabpanel when first element inside panel is focusable element


anveshmekala:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When the user activates the tab, the tabpanel is focusable irrespective of whether the first element in the content is focusable or not.

### Expected Behavior

If the first element in the tabpanel content is focusable then the tabpanel should not be included in the keyboard Tab sequence as per [W3Cspecs](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-manual/#:~:text=Note%20that%20since,is%20not%20focusable.).

### Reproduction Sample

https://codepen.io/ANVESHMEKALA/pen/rNoVOrO?editors=1000

### Reproduction Steps

1. Open the Codepen
2. Activate the second calcite-tab with title `Component2` by clicking or space key (when focused).
3. Tab into the content 
4. Observe that content container is focused
5. Tab again and observe that `calcite-tab` link is focused.

### Reproduction Version

1.6.1

### Priority impact

p4 - not time sensitive

### Impact

This will impact `calcite-color-picker` component

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

N/A

Elijbet:
After some thought, it seems like this issue is not valid. We don't know what the contents of the `tabpanel` are. These elements might not be tabbable. If they aren't we would need to be able to tab into the content.

DitwanP:
Closing due to the point made above. 👆

PR #8616, opened 2024-01-16
feat(tabs): make component responsive

jcfranco:
**Related Issue:** #6689 

## Summary

This improves how `tabs` works at narrow widths where `tab-title`s can't be displayed entirely. 

### Notable changes

* overflowing `tab-title`s will be clipped
* affordances to scroll clipped `tab-titles` forward/backward will be displayed accordingly
* selecting a `tab-title` that is partially clipped will be scrolled into view
* `tab-title`s can be scrolled via mousewheel

See [design spec](https://www.figma.com/file/JAINbGHSykI8JQCVzOHiSz/Tabs---responsive-design-%5B6689%5D?node-id=862%3A13837&mode=dev) for more details.


packages/calcite-components/.storybook/utils.tsx
```diff
@@ -146,7 +146,7 @@ export const filterComponentAttributes = (
 /**
  * This helper creates a story that captures all breakpoints across all scales for testing.
  *
- * @param singleStoryHtml – HTML story template with placeholders for `scale` attributes (e.g., `{scale}`).
+ * @param singleStoryHtml – HTML story template with placeholders for `scale` attributes (e.g., `{scale}`). You can additionally use `.breakpoint-stories-container` and `.breakpoint-story-container` to style breakpoint story containers.
```
jcfranco, line 149: Extracted this over to https://github.com/Esri/calcite-design-system/pull/8739.

packages/calcite-components/src/components/tab-title/tab-title.scss
```diff
@@ -26,29 +26,29 @@
...
     padding-inline-start: 32px; //28px button width + 0.25rem padding
   }
-}
 
-:host([layout="center"][bordered][closable]) .scale-s {
-  .content {
-    padding-inline-start: 36px; //28px button width + 0.5rem padding
+  .scale-l {
+    .content {
+      padding-inline-start: 40px; //36px button width + .25 padding
```
driskull, line 32: Maybe to get rid of these comments we should have an internal var for button width and padding? Same goes for other uses below.

packages/calcite-components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -301,35 +412,197 @@ export class TabNav {
...
+  private storeContainerRef = (el: HTMLDivElement) => (this.containerEl = el);
+
+  private storeTabTitleWrapperRef = (el: HTMLDivElement) => {
+    this.tabTitleContainerEl = el;
+
+    if (!el) {
+      return;
+    }
+
+    this.intersectionObserver = createObserver("intersection", () => this.updateScrollingState(), {
```
driskull, line 504: Any chance `storeTabTitleWrapperRef` could be called again and this intersectionObserver may need to be disconnected before the new one is created?

packages/calcite-components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -301,35 +412,197 @@ export class TabNav {
...
+
+    this.hasOverflowingStartTabTitle = isOverflowStart;
+    this.hasOverflowingEndTabTitle = isOverflowEnd;
+  }
+
+  private scrollToTabTitles = (direction: "forward" | "backward"): void => {
+    readTask(() => {
+      const tabTitleContainer = this.tabTitleContainerEl;
+      const containerBounds = tabTitleContainer.getBoundingClientRect();
+      const tabTitles = Array.from(this.el.querySelectorAll("calcite-tab-title"));
```
driskull, line 540: maybe these could be stored with the onSlotChange and not queried via DOM?

packages/calcite-components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -414,9 +673,38 @@ export class TabNav {
...
+    const isEnd = overflowDirection === "end";
+
+    return (
+      <div
+        class={{
+          [CSS.scrollButtonContainer]: true,
+          [CSS.scrollBackwardContainerButton]: !isEnd,
+          [CSS.scrollForwardContainerButton]: isEnd,
+        }}
+        hidden={(isEnd && !hasOverflowingEndTabTitle) || (!isEnd && !hasOverflowingStartTabTitle)}
```
driskull, line 693: Instead of hiding, should it just not be rendered? Not a deal breaker just wondering.

driskull (approved):
Looks good! 🏆 ✨

macandcheese:
Nice! Some design / UX feedback mostly (cc @ashetland @SkyeSeitz)

- If I use the buttons via keyboard, then "shift tab back" to the tab items, and the now focused item is out of view, it does not scroll into view, so it is impossible to tell the tab has been successfully focused.

- Using the arrow keys, it seems the trigger to "scroll to focused item" is either a step off or happens to be below the width threshold sometimes. What this means is I've used the arrow key, navigated to a now hidden tab, and it hasn't re-adjusted yet. Navigating to another will scroll this one into view, so maybe there needs to be a width buffer etc.

- If I have navigated with the arrow keys away from the active tab to focus-but-not-active a sibling, and then use Tab, the focus doesn't go to the arrow (or content). Not sure if intentional but it does re-focus the active tab before allowing you to tab again through.

- Using the arrow buttons (or arrow keys) doesn't always increment a full title into view. Sometimes I can hit "next" or "prev" 2/3 times and a long title will still be in view - can we ensure a new tab is visible?

- Can we use a rule to make scrolling respond better? Right now I need to swipe-swipe-swipe and it doesn't seem to have inertia scrolling making it tough to use on a phone or trackpad / mouse.

- Should we support long-press on these arrow keys in any way on mobile?

Style issues:

- This focus area is tabbable - if this is intentional we should use a default focus style:
- Is it a bug that there is no border on top of these arrow buttons?
Screenshot 2024-02-13 at 2 08 09 PM

- Design - do we want to limit these to squares? These are taking up too much space IMO:
- Design - is this the correct spacing here? It seems really cramped and spacious at the same time.
Screenshot 2024-02-13 at 2 07 24 PM

- The focus area outline is misaligned:
Screenshot 2024-02-13 at 2 07 03 PM

- The padding on some of these is a bit weird - compare the first tabs to the subsequent ones.
Screenshot 2024-02-13 at 2 06 29 PM

- I suggest we get rid of the "zoom border animation", it is adding a bit of confusion to a crowded area (IMO). Just render the active state visual with a border or inset box-shadow.

driskull:
It might be nice if instead of having to horizontally scroll through tabs there was an overflow menu. Not a huge fan of having to horizontally "carousel" through tabs.

image

SkyeSeitz:
> It might be nice if instead of having to horizontally scroll through tabs there was an overflow menu. Not a huge fan of having to horizontally "carousel" through tabs.
> 
> image

Might be subjective, but I think an overflow menu is less expected for Tabs specifically. See below for an example with the scroll function working..

https://github.com/Esri/calcite-design-system/assets/29716057/cdb66601-d8f0-4359-b0da-4b98cceea986

SkyeSeitz:
Agree with all Adam's points above and have similar thoughts (tried to not repeat any notes, but apologies if there is overlap)
* > This focus area is tabbable - if this is intentional we should use a default focus style:

Referring to the content area, correct? I had the same question, although unless there is something clickable in the content area, I am not sure it needs to be focusable. Looks like this is how it works & is styled currently, however. 
    
* > do we want to limit these to squares? These are taking up too much space IMO:

Agreed. These should have the following square dimensions
image


Design UX questions: 
* When arrowing through tabs, I am not sure the last tab needs to roll into the 1st tab? Thoughts?
* When keyboard navigating to one of the ends of Tab via the chevron buttons, the focus ring goes away after reaching the end. I think I would expect it to move to the selected tab when the focused button goes away?

jcfranco:
> * Using the arrow buttons (or arrow keys) doesn't always increment a full title into view. 
Sometimes I can hit "next" or "prev" 2/3 times and a long title will still be in view - can we ensure a new tab is visible?
> * If I use the buttons via keyboard, then "shift tab back" to the tab items, and the now focused item is out of view, it does not scroll into view, so it is impossible to tell the tab has been successfully focused.
> * Using the arrow keys, it seems the trigger to "scroll to focused item" is either a step off or happens to be below the width threshold sometimes. What this means is I've used the arrow key, navigated to a now hidden tab, and it hasn't re-adjusted yet. Navigating to another will scroll this one into view, so maybe there needs to be a width buffer etc.
> * Is it a bug that there is no border on top of these arrow buttons?
> * Design - do we want to limit these to squares? These are taking up too much space IMO:

Fixed! ✨🔨✨

> * Can we use a rule to make scrolling respond better? Right now I need to swipe-swipe-swipe and it doesn't seem to have inertia scrolling making it tough to use on a phone or trackpad / mouse.

Updated swipe-scrolling to be ↕️on desktop and ↔️on mobile.

> * Should we support long-press on these arrow keys in any way on mobile?

The updated swipe-scrolling might take care of this now.

> * If I have navigated with the arrow keys away from the active tab to focus-but-not-active a sibling, and then use Tab, the focus doesn't go to the arrow (or content). Not sure if intentional but it does re-focus the active tab before allowing you to tab again through.

We could refine this if needed. The roving tab helper might help here.

> * This focus area is tabbable - if this is intentional we should use a default focus style:
> Referring to the content area, correct? I had the same question, although unless there is something clickable in the content area, I am not sure it needs to be focusable. Looks like this is how it works & is styled currently, however.

This is by intentional for a11y purposes, there is an issue to revisit making it tabbable only when interactive (https://github.com/Esri/calcite-design-system/issues/7576).

> * The focus area outline is misaligned:

Is this specific to the close button? If so, we can fix this separately since this issue is present in production too (https://github.com/Esri/calcite-design-system/issues/8772).

> * The padding on some of these is a bit weird - compare the first tabs to the subsequent ones.

Updated spacing to better match the spec. There is an issue though with center + closable layouts when they're used in narrow widths. You can see this in the UI kit:

Screenshot 2024-02-16 at 4 52 21 PM

cc @SkyeSeitz @ashetland

> * I suggest we get rid of the "zoom border animation", it is adding a bit of confusion to a crowded area (IMO). Just render the active state visual with a border or inset box-shadow.

I'm down to remove/simplify this. Do you think we could tackle it as a separate issue or is it critical for this PR?

> * When arrowing through tabs, I am not sure the last tab needs to roll into the 1st tab? Thoughts?

Do you mean using keyboard left/right arrows when focus is on a tab or clicking on the prev/next tab actions? If it's the latter, this behavior is specified under Q.3 under `"Franco questions 01.17.24"` from the responsive tabs spec.

> * When keyboard navigating to one of the ends of Tab via the chevron buttons, the focus ring goes away after reaching the end. I think I would expect it to move to the selected tab when the focused button goes away? 

This might be confusing since you'll scroll back to the selected tab after 'paging' all the way to the end. E.g., selected tab is 2nd one, you page via keyboard with the chevron button until the end, it scrolls back to the 2nd one. I might be missing something though.

macandcheese:
Works a lot better now!

>This is by intentional for a11y purposes, there is an issue to revisit making it tabbable only when interactive (https://github.com/Esri/calcite-design-system/issues/7576).
Maybe we can use a shared focus style so it isn't browser default?

>I'm down to remove/simplify this. Do you think we could tackle it as a separate issue or is it critical for this PR?
Separate is good, it might resolve some of the alignment items, but it would be a refactor to remove associated code.

>This might be confusing since you'll scroll back to the selected tab after 'paging' all the way to the end. E.g., selected tab is 2nd one, you page via keyboard with the chevron button until the end, it scrolls back to the 2nd one. I might be missing something though.

Maybe the arrow buttons do not need to be in the tab order. They are likely to be used via tap / click. But, to a keyboard-only user, using those doesn't provide any use - it's just moving what is visible, not selecting a new tab. Keyboard-only users can already do this more efficiently with arrow keys. 

So, I think we could exclude them from tab order. This would be similar to some input functionality like number spinner, clear button - that are clickable entities but excluded from tab focus stack because the keyboard-only user has better usability without them.

SkyeSeitz:
1. >So, I think we could exclude them from tab order. This would be similar to some input functionality like number spinner, clear button - that are clickable entities but excluded from tab focus stack because the keyboard-only user has better usability without them.

This is a super good point and would take care of my question. 

2. >Updated swipe-scrolling to be ↕️on desktop and ↔️on mobile.

Is there a reason ↔️ does not scroll on desktop? I would expect both directions to scroll.

jcfranco:
> Maybe we can use a shared focus style so it isn't browser default?

Agreed. Created https://github.com/Esri/calcite-design-system/issues/8799 for this.

> Separate is good, it might resolve some of the alignment items, but it would be a refactor to remove associated code.

I will create an issue for this today.

> Maybe the arrow buttons do not need to be in the tab order. 

Removed from tab order! I think this works way better for keyboard usage (IMO).

> Is there a reason ↔️ does not scroll on desktop? I would expect both directions to scroll.

When testing locally, it seems when you scroll with the mouse wheel and trackpad (at least mine), I get both delta x and delta y values. I wouldn't know which one to use if I handle both horizontal/vertical scrolling.

packages/calcite-components/src/components/tab-nav/tab-nav.scss
```diff
@@ -1,78 +1,199 @@
...
+        transparent,
+        transparent var(--calcite-internal-tab-nav-scroller-button-width),
+        white var(--calcite-internal-tab-nav-scroller-button-width),
+        white $last-mask-color-stop-position,
+        transparent $last-mask-color-stop-position
+      );
+  }
+}
+
+.tab-nav::-webkit-scrollbar {
```
driskull, line 78: Whats this for? we're hiding the scrollbar but still allowing scrolling?

driskull (approved):
👍

packages/calcite-components/src/components/tab-nav/tab-nav.scss
jcfranco: It was added [here](https://github.com/Esri/calcite-design-system/commit/712fb31655e851973502737764588881b73b4e3c). Removing it doesn't seem to have any effect. @Elijbet Do you recall why this is needed? If not, I can drop this. Good catch! ✨🎣✨

jcfranco:
> Is there a reason ↔️ does not scroll on desktop? I would expect both directions to scroll.

Managed to make this work thanks to https://github.com/react-component/tabs. 🎉

jcfranco:
This is now officially good to land, but the extent of changes aren't low-risk. I suggest we install this as soon tomorrow's maintenance release is wrapped up. cc @geospatialem @brittneytewks

PR #8618, opened 2024-01-16
feat(button): add component tokens

DitwanP:
**Related Issue:** #7180 

## Summary
Add the following component tokens to the button component:

`--calcite-button-background-color`: Specifies the background color of the component.
`--calcite-button-background-color-hover`: Specifies the background color of the component when in hover state.
`--calcite-button-background-color-focus`: Specifies the background color of the component when in focus state.
`--calcite-button-background-color-active`: Specifies the background color of the component when in active state.
`--calcite-button-text-color`: Specifies the text color of the component.
`--calcite-button-text-color-hover`: Specifies the text color of the component when in hover state.
`--calcite-button-text-color-focus`: Specifies the text color of the component when in focus state.
`--calcite-button-text-color-active`: Specifies the text color of the component when in active.
`--calcite-button-icon-start-color`: Specifies the color of the component's start position icon.
`--calcite-button-icon-start-color-hover`: Specifies the color of the component's start position icon when in hover state.
`--calcite-button-icon-start-color-focus`: Specifies the color of the component's start position icon when in focus state.
`--calcite-button-icon-start-color-active`: Specifies the color of the component's start position icon when in active state.
`--calcite-button-icon-end-color`: Specifies the color of the component's end position icon.
`--calcite-button-icon-end-color-hover`: Specifies the color of the component's end position icon when in hover state.
`--calcite-button-icon-end-color-focus`: Specifies the color of the component's end position icon when in focus state.
`--calcite-button-icon-end-color-active`: Specifies the color of the component's end position icon when in active state.
`--calcite-button-box-shadow`: Specifies the box shadow of the component.
`--calcite-button-box-shadow-hover`: Specifies the box shadow of the component when in hover state.
`--calcite-button-box-shadow-focus`: Specifies the box shadow of the component when in focus state.
`--calcite-button-box-shadow-active`: Specifies the box shadow of the component when in active state.
`--calcite-button-border-color`: Specifies border colors of the component.
`--calcite-button-border-color-hover`: Specifies border colors of the component when in hover state.
`--calcite-button-border-color-focus`: Specifies border colors of the component when in focus state.
`--calcite-button-border-color-active`: Specifies border colors of the component when in active state`.
 
 Also adds the correct `appearance` for some button demos.

### DIFF
--- a/packages/calcite-components/src/components/button/button.scss
+++ b/packages/calcite-components/src/components/button/button.scss
@@ -4,25 +4,63 @@
  * These properties can be overridden using the component's tag as selector.
  *
  * @prop --calcite-button-background-color: Specifies the background color of the component.
- * @prop --calcite-button-text-color: Specifies the text and icon color of the component.
- * @prop --calcite-button-box-shadow: Specifies the border color of the component.
- * @prop --calcite-button-border-color: Specifies all border colors of the component.
- * @prop --calcite-button-border-top-color: Specifies the top border color of the component.
- * @prop --calcite-button-border-right-color: Specifies the right border color of the component.
- * @prop --calcite-button-border-bottom-color: Specifies the bottom border color of the component.
- * @prop --calcite-button-border-left-color: Specifies the left border color of the component.
- *
+ * @prop --calcite-button-background-color-hover: Specifies the background color of the component when in hover state.
+ * @prop --calcite-button-background-color-focus: Specifies the background color of the component when in focus state.
+ * @prop --calcite-button-background-color-active: Specifies the background color of the component when in active state.
+ * @prop --calcite-button-text-color: Specifies the text color of the component.
+ * @prop --calcite-button-text-color-hover: Specifies the text color of the component when in hover state.
+ * @prop --calcite-button-text-color-focus: Specifies the text color of the component when in focus state.
+ * @prop --calcite-button-text-color-active: Specifies the text color of the component when in active.
+ * @prop --calcite-button-icon-start-color: Specifies the color of the component's start position icon.
+ * @prop --calcite-button-icon-start-color-hover: Specifies the color of the component's start position icon when in hover state.
+ * @prop --calcite-button-icon-start-color-focus: Specifies the color of the component's start position icon when in focus state.
+ * @prop --calcite-button-icon-start-color-active: Specifies the color of the component's start position icon when in active state.
+ * @prop --calcite-button-icon-end-color: Specifies the color of the component's end position icon.
+ * @prop --calcite-button-icon-end-color-hover: Specifies the color of the component's end position icon when in hover state.
+ * @prop --calcite-button-icon-end-color-focus: Specifies the color of the component's end position icon when in focus state.
+ * @prop --calcite-button-icon-end-color-active: Specifies the color of the component's end position icon when in active state.
+ * @prop --calcite-button-box-shadow: Specifies the box shadow of the component.
+ * @prop --calcite-button-box-shadow-hover: Specifies the box shadow of the component when in hover state.
+ * @prop --calcite-button-box-shadow-focus: Specifies the box shadow of the component when in focus state.
+ * @prop --calcite-button-box-shadow-active: Specifies the box shadow of the component when in active state.
+ * @prop --calcite-button-border-color: Specifies border colors of the component.
+ * @prop --calcite-button-border-color-hover: Specifies border colors of the component when in hover state.
+ * @prop --calcite-button-border-color-focus: Specifies border colors of the component when in focus state.
+ * @prop --calcite-button-border-color-active: Specifies border colors of the component when in active state.
  */
 
 :host {
   --calcite-button-background-color: var(--calcite-color-background);
-  --calcite-button-text-color: var(--calcite-color-text-1);
+  --calcite-button-text-color: var(--calcite-color-text-inverse);
   --calcite-button-box-shadow: var(--calcite-shadow-none);
   --calcite-button-border-color: var(--calcite-color-transparent);
-  --calcite-button-border-top-color: var(--calcite-color-transparent);
-  --calcite-button-border-right-color: var(--calcite-color-transparent);
-  --calcite-button-border-bottom-color: var(--calcite-color-transparent);
-  --calcite-button-border-left-color: var(--calcite-color-transparent);
+  --calcite-button-corner-radius: var(--calcite-corner-radius-sharp);
+  --calcite-button-icon-start-color: var(--calcite-color-text-inverse);
+  --calcite-button-icon-end-color: var(--calcite-color-text-inverse);
+
+  // hover
+  --calcite-button-text-color-hover: var(--calcite-color-text-inverse-hover);
+  --calcite-button-background-color-hover: var(--calcite-color-background);
+  --calcite-button-border-color-hover: var(--calcite-color-transparent);
+  --calcite-button-box-shadow-hover: var(--calcite-shadow-none);
+  --calcite-button-icon-start-color-hover: var(--calcite-color-text-inverse-hover);
+  --calcite-button-icon-end-color-hover: var(--calcite-color-text-inverse-hover);
+
+  // focus
+  --calcite-button-text-color-focus: var(--calcite-color-text-inverse-hover);
+  --calcite-button-background-color-focus: var(--calcite-color-background);
+  --calcite-button-border-color-focus: var(--calcite-color-transparent);
+  --calcite-button-box-shadow-focus: var(--calcite-shadow-none);
+  --calcite-button-icon-start-color-focus: var(--calcite-color-text-inverse-focus);
+  --calcite-button-icon-end-color-focus: var(--calcite-color-text-inverse-focus);
+
+  // active
+  --calcite-button-text-color-active: var(--calcite-color-text-inverse-press);
+  --calcite-button-background-color-active: var(--calcite-color-background);
+  --calcite-button-border-color-active: var(--calcite-color-transparent);
+  --calcite-button-box-shadow-active: var(--calcite-shadow-none);
+  --calcite-button-icon-start-color-active: var(--calcite-color-text-inverse-active);
+  --calcite-button-icon-end-color-active: var(--calcite-color-text-inverse-active);
 
   @apply inline-block w-auto align-middle;
 }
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/button/button.scss
```diff
@@ -1,4 +1,29 @@
...
+ * @prop --calcite-button-border-left-color: Specifies the left border color of the component.
+ *
+ */
+
 :host {
+  --calcite-button-background-color: var(--calcite-color-background);
+  --calcite-button-text-color: var(--calcite-color-text-1);
+  --calcite-button-box-shadow: var(--calcite-shadow-none);
+  --calcite-button-border-color: var(--calcite-color-transparent);
+  --calcite-button-border-top-color: var(--calcite-color-transparent);
```
macandcheese, line 22: Do we want to expose these individual border color values? If so, should we use inline / block nomenclature? Unless this is needed for split-button, maybe these specific ones could be internal to begin?

packages/calcite-components/src/components/button/button.scss
```diff
@@ -1,4 +1,29 @@
...
+  --calcite-button-border-right-color: var(--calcite-color-transparent);
+  --calcite-button-border-bottom-color: var(--calcite-color-transparent);
+  --calcite-button-border-left-color: var(--calcite-color-transparent);
+
```
eriklharper, line 26: Since there are so many places where these component variables need to be set based on prop combinations, it would be much more readable and discoverable if instead of assigning and reassigning all of them in the existing selector blocks below, if you could move all the css variable assignments up to the top of the file here, assigning them based on their prop combinations in a new set of `:host` css selectors that mirror their current placement in the rest of the css file.  That way you can see at a glance which prop combinations result in which default component variable settings.  You can then use assign those variables in the appropriate places below.

For example:
```
:host([kind="brand"]) {
    --calcite-button-text-color: var(--calcite-color-text-inverse);
    --calcite-button-background-color: var(--calcite-color-brand);
}
:host([kind="danger"]) {
    --calcite-button-text-color: var(--calcite-color-text-inverse);
    --calcite-button-background-color: var(--calcite-color-status-danger);
}
// additional kinds here and other prop combinations...

// And then you only have to set these values once with a rule like this that sits below all the prop-based assignments:
:host {
    button, a {
        color: var(--calcite-button-text-color);
        background-color: var(--calcite-button-background-color);
    }
}
```

packages/calcite-components/src/components/button/button.scss
```diff
@@ -228,271 +253,288 @@
...
+    box-shadow: var(--calcite-button-box-shadow);
   }
 }
 
 :host([kind="brand"]) {
   button,
   a {
-    @apply text-color-inverse bg-brand;
+    --calcite-button-text-color: var(--calcite-color-text-inverse);
+    --calcite-button-background-color: var(--calcite-color-brand);
```
eriklharper, line 267: Looks like you forgot to assign `color` and `background-color` here.  Applies to many other places below too.  See my earlier comment about revising this to assign all the variants of these css properties at the top of the file.

packages/calcite-components/src/components/button/button.scss
```diff
@@ -228,271 +253,288 @@
 :host([appearance]) {
   button,
   a {
-    @apply border-color-transparent
-      border
-      border-solid;
+    background-color: var(--calcite-button-background-color);
+    color: var(--calcite-button-text-color);
+    border: var(--calcite-border-width-sm) solid var(--calcite-button-border-color);
+    box-shadow: var(--calcite-button-box-shadow);
```
eriklharper, line 259: Based on this usage, `--calcite-button-box-shadow` defines the entire box shadow, not just the color as the description above indicates.  Also, should we be setting both `border` and `box-shadow` together?  Seems like we should stick with just one for simplicity.

packages/calcite-components/src/components/button/button.scss
```diff
@@ -228,271 +253,288 @@
 :host([appearance]) {
   button,
   a {
-    @apply border-color-transparent
-      border
-      border-solid;
+    background-color: var(--calcite-button-background-color);
+    color: var(--calcite-button-text-color);
```
eriklharper, line 257: These styles look like they should just be assigned with a generic `:host { button, a {}}` selector instead of here where it is looking just for the `appearance` attribute.  Adopting the pattern I described earlier will necessitate moving these styles to a generic `:host` selector anyway.

packages/calcite-components/src/components/button/button.scss
```diff
@@ -1,4 +1,29 @@
...
+ * @prop --calcite-button-border-color: Specifies all border colors of the component.
+ * @prop --calcite-button-border-top-color: Specifies the top border color of the component.
+ * @prop --calcite-button-border-right-color: Specifies the right border color of the component.
+ * @prop --calcite-button-border-bottom-color: Specifies the bottom border color of the component.
+ * @prop --calcite-button-border-left-color: Specifies the left border color of the component.
+ *
+ */
+
 :host {
+  --calcite-button-background-color: var(--calcite-color-background);
```
macandcheese, line 18: Can we add `corner-radius` and default it to `sharp`?

eriklharper (changes requested):
See comments.

packages/calcite-components/src/components/button/button.scss
DitwanP: Yeah this one was tricky, I asked Ali about it and they asked Aaron about it, but there was no precedence set for how we wanted to handle that yet so for now they said that there should just be one token called `--calcite-button-box-shadow` and for now just to apply the existing values that were there before to it.

macandcheese: This makes sense IMO as our box-shadow tokens set all portions of a box-shadow rule. Border-color should be separate as it is here.

eriklharper: We should update the @prop doc above then to say "Specifies the box shadow of the component" or something to that effect.  I also again think we should revisit border vs. box-shadow.  I know we use box-shadow to create border effects instead of CSS `border` in other places because of all the extra control you get with it, but IMO we should stick to one vs. the other unless there's a specific use case we're trying to solve for that necessitates both.

Also, not sure if naming the css property to include the words "box shadow" makes sense if one of our goals is not to leak implementation details or not.  I don't have a hard opinion on that since this is would be applying a web-platform variable and we should use those terms in our documentation since that's the target platform for these variables in this case, but thought that would be a good point to bring up too.

@macandcheese @ashetland @SkyeSeitz: is there a use case for supporting both a `border` _and_ a `box-shadow` on button?

eriklharper: I take that back what I said about border vs box-shadow.  I think its valid to include both of them just because of the nature of what box-shadow allows you to do.  I'm noticing Adam's work on Card is similar to what someone might want to do with button.

packages/calcite-components/src/components/button/button.scss
```diff
@@ -1,4 +1,29 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-button-background-color: Specifies the background color of the component.
+ * @prop --calcite-button-text-color: Specifies the text and icon color of the component.
```
macandcheese, line 7: We should separate the icon css props - add `--calcite-button-icon-start-color` and `--calcite-button-icon-end-color`, which can default to thi text-color value.

macandcheese: I suppose you could also have a `--calcite-button-icon-color` var that would set both, and these explicitly for each, but not sure how far down the rabbit hole we want to go there. Ostensibly, you could set `--calcite-icon-color` on a button to achieve the same as `--calcite-button-icon-color`, so only individual ones for start/end makes sense and doesn't prevent a theming story.

DitwanP: Yeah that makes sense 👍

packages/calcite-components/src/components/button/button.scss
DitwanP: I'm not sure if they are 100% necessary but there is one spot in the button file where we are using `border-t-color-input` and `border-b-color-input` so I thought we might have wanted the option of individual border colors. I removed it for now because there's a lot going on but can revisit when there's some conclusion as to how we want to go about it.

packages/calcite-components/src/components/button/button.scss
DitwanP: So I made some changes based on the suggestions you gave, the only one I didn't do was moving some things up to the top, I kept them relatively in the same positions because I didn't want to have to deal with potentially unwanted issues from repositioning some styles above others. I think the changes already went above the main focus of this effort which was just to swap out some values for tokens based on other PRs I saw. 

Let me know what you think about the changes and if they are close to what you had in mind.

macandcheese: My only question remaining is the need for explicit border sides. Can we just do one? I think this aligns with other component efforts unless there is another reason @DitwanP

packages/calcite-components/src/components/button/button.scss
```diff
@@ -1,4 +1,67 @@
...
 :host {
+  --calcite-button-background-color: var(--calcite-color-background);
+  --calcite-button-text-color: var(--calcite-color-text-inverse);
+  --calcite-button-box-shadow: var(--calcite-shadow-none);
+  --calcite-button-border-color: var(--calcite-color-transparent);
+  --calcite-button-corner-radius: var(--calcite-corner-radius-sharp);
+  --calcite-button-icon-start-color: var(--calcite-color-text-inverse);
+  --calcite-button-icon-end-color: var(--calcite-color-text-inverse);
+
+  // hover
```
eriklharper, line 41: These css comments should be inside `/* ... */` syntax.  Applies to others below.

packages/calcite-components/src/components/button/button.scss
```diff
@@ -1,4 +1,67 @@
...
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-button-background-color: Specifies the background color of the component.
+ * @prop --calcite-button-background-color-hover: Specifies the background color of the component when in hover state.
+ * @prop --calcite-button-background-color-focus: Specifies the background color of the component when in focus state.
+ * @prop --calcite-button-background-color-active: Specifies the background color of the component when in active state.
+ * @prop --calcite-button-text-color: Specifies the text color of the component.
+ * @prop --calcite-button-text-color-hover: Specifies the text color of the component when in hover state.
+ * @prop --calcite-button-text-color-focus: Specifies the text color of the component when in focus state.
+ * @prop --calcite-button-text-color-active: Specifies the text color of the component when in active.
```
eriklharper, line 13: Should this read "when in active **state**" to match the others?

eriklharper (approved):
Just spotted a couple things, so this is good to merge after you address those 2 items and we get Chromatic snapshots to run.  Looks amazing, excellent work!

packages/calcite-components/src/components/button/button.scss
DitwanP: I got rid of the border sides so the code referenced in this comment is now outdated there's just one again. @macandcheese

PR #8622, opened 2024-01-17
fix(list-item): always show hover and pointer styling

driskull:
**Related Issue:** #6700

## Summary

It was decided that the majority of cases use the layer list to perform some kind of action so the default should be to have hover and pointer styling by default. In the future, we may add another selectionMode to provide a better UX for "non interactive" lists.

Further information: https://github.com/Esri/calcite-design-system/issues/6123#issuecomment-1361770131

- Always has the hover and pointer styling on a list item regardless of selectionMode
- Add e2e test

### DIFF
--- a/packages/calcite-components/src/components.d.ts
+++ b/packages/calcite-components/src/components.d.ts
@@ -916,8 +916,6 @@ export namespace Components {
          */
         "setFocus": () => Promise<void>;
     }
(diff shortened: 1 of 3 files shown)


jcfranco (approved):
Thanks for fixing! 🏆 Can you add a recap of the decision in the summary for our future selves?

PR #8627, opened 2024-01-19
fix(action-menu): clicking an action menu item should call click event.

driskull:
**Related Issue:** #8577 #8628

## Summary

- This reverts commit ccfbd0c69f3f83149808f16dfaa389dc463c2442.
- Adds test

### DIFF
--- a/packages/calcite-components/src/components/action-menu/action-menu.e2e.ts
+++ b/packages/calcite-components/src/components/action-menu/action-menu.e2e.ts
@@ -436,7 +436,7 @@ describe("calcite-action-menu", () => {
       expect(await actionMenu.getProperty("open")).toBe(false);
     });
 
-    it("should click the active action and close the menu", async () => {
+    it("should click the active action on Enter key and close the menu", async () => {
       const page = await newE2EPage({
         html: html`<calcite-action-menu>
           <calcite-action id="first" text="Add" icon="plus" text-enabled></calcite-action>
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/action-menu/action-menu.e2e.ts
```diff
@@ -466,9 +466,43 @@ describe("calcite-action-menu", () => {
...
+      await page.waitForChanges();
+
+      const clickSpy = await actions[0].spyOnEvent("click");
+
+      expect(await actionMenu.getProperty("open")).toBe(true);
+      expect(actions[0].getAttribute(activeAttr)).toBe("");
+      expect(actions[1].getAttribute(activeAttr)).toBe(null);
+      expect(actions[2].getAttribute(activeAttr)).toBe(null);
+
+      await page.$eval("calcite-action", (el: HTMLCalciteActionElement) => el.click());
```
jcfranco, line 504: Is there a reason why this is calling `click` directly vs clicking via the E2E APIs? If so, can you add an explainer comment, otherwise, I'd suggest `await page.click("calcite-action");` to simplify.

jcfranco (approved):
💥

packages/calcite-components/src/components/action-menu/action-menu.e2e.ts
driskull: Yeah it wasn't triggering the closing of the menu when `await page.click("calcite-action")` was used.

Issue #8413, opened 2023-12-12
[Tab] Add css custom property to override built-in padding


macandcheese:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Currently, Tab provides built in y-padding that is not overridable in a supported way.

### Acceptance Criteria

We should expose a css custom property, such as `--calcite-tab-padding`. This aligns with Block, Modal, and other components where built-in padding is provided.

### Relevant Info

Related to this comment: https://github.com/Esri/calcite-design-system/issues/8139#issuecomment-1830397218

### Which Component

Tab / Tabs

### Example Use Case

As a user, I want to have a full-height content in a Tab, with no built-in padding.

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

RSantosGIS:
Adding from our Teams thread I brought up:

I just updated Studio to Calcite 2.1 from 1.9.  I noticed that padding is now being applied to the CalciteTab "<selection> element, which is causing our layout to get thrown off and creating an undesirable white stripe above the content.  Is there a way to turn off this padding, similar to the x-content-padding prop that exists on other components, or some kind of recommended way to do this?
 
It looks like the difference is occurring because of a change to the way the padding-block style is being applied.  It looks like with the old way, it was a non-inheritable property and so was never actually applied (perhaps your team considered this a bug?), but the new selector is being applied in 2.1 and I can't figure out a way to override it.
 
Please see the screenshots below:
 
1.9 (no padding, desired)
![image (4)](https://github.com/Esri/calcite-design-system/assets/36513418/841b9914-406f-4af8-871b-1c82d9b89aa3)

2.1 (padding, not desired)
![image (5)](https://github.com/Esri/calcite-design-system/assets/36513418/c97dc855-c748-45eb-97a1-f5acc647029b)

RSantosGIS:
Adding a follow up comment on this due to the recent R1 completion announcement.  This is a style regression in ArcGIS Knowledge Studio.  If the proposed property isn't going to make it, we need some kind of guidance on a way to hack/workaround to remove the padding.

geospatialem:
[...]

PR #8629, opened 2024-01-21
feat(tab): add `--calcite-tab-content-block-padding` to override built-in block-padding

Elijbet:
**Related Issue:** #8413

## Summary
Introduces a CSS custom property `--calcite-tab-content-block-padding` to align with other components that have built-in padding. 

Currently, the Tab component has a built-in `block-padding` that cannot be overridden in a supported way, leading to a bug with scrollbars when tabs are housed inside a panel. 

The introduction of the custom property allows for more flexibility in the component, as users can achieve their desired layout and padding by adding it to their slotted content. (https://github.com/Esri/calcite-design-system/issues/8139#issuecomment-1830397218)

### DIFF
--- a/packages/calcite-components/src/components/tab/tab.e2e.ts
+++ b/packages/calcite-components/src/components/tab/tab.e2e.ts
@@ -1,4 +1,20 @@
+import { newE2EPage } from "@stencil/core/testing";
 import { defaults, renders, hidden } from "../../tests/commonTests";
+import { html } from "../../../support/formatting";
+import { CSS } from "../tab-title/resources";
+
+const tabBlockPadding = html`
+  <calcite-tabs>
+    <calcite-tabs>
+      <calcite-tab-nav slot="title-group">
+        <calcite-tab-title selected>Sandwiches</calcite-tab-title>
+      </calcite-tab-nav>
+      <calcite-tab style="--calcite-block-padding: 0;">
+        <div>Chicken with Pesto</div>
+      </calcite-tab>
+    </calcite-tabs>
+  </calcite-tabs>
+`;
 
 describe("calcite-tab", () => {
   const tabHtml = "<calcite-tab>A tab</calcite-tab>";
@@ -19,5 +35,33 @@ describe("calcite-tab", () => {
       { propertyName: "selected", defaultValue: false },
       { propertyName: "scale", defaultValue: "m" },
     ]);
+  });
+
+  it("should allow the CSS custom property to be overridden when applied to :root", async () => {
+    const overrideStyle = "0px";
+    const page = await newE2EPage();
+    await page.setContent(html`
+      <style>
+        :root {
+          --calcite-block-padding: ${overrideStyle};
+        }
+      </style>
+      ${tabBlockPadding}
+    `);
+    const content = await page.find(`calcite-tab >>> .${CSS.content}`);
+    const contentStyles = await content.getComputedStyle();
+    const contentPadding = contentStyles.getPropertyValue("padding");
+    expect(contentPadding).toEqual(overrideStyle);
+  });
+
+  it("should allow the CSS custom property to be overridden when applied to element", async () => {
+    const overrideStyle = "0px";
+    const page = await newE2EPage();
+    await page.setContent(tabBlockPadding);
+
+    const content = await page.find(`calcite-tab >>> .${CSS.content}`);
+    const contentStyles = await content.getComputedStyle();
+    const contentPadding = await contentStyles.getPropertyValue("padding");
+    expect(contentPadding).toEqual(overrideStyle);
   });
 });
--- a/packages/calcite-components/src/components/tab/tab.e2e.ts
+++ b/packages/calcite-components/src/components/tab/tab.e2e.ts
@@ -1,20 +1,4 @@
-import { newE2EPage } from "@stencil/core/testing";
 import { defaults, renders, hidden } from "../../tests/commonTests";
-import { html } from "../../../support/formatting";
-import { CSS } from "../tab-title/resources";
-
-const tabBlockPadding = html`
-  <calcite-tabs>
-    <calcite-tabs>
-      <calcite-tab-nav slot="title-group">
-        <calcite-tab-title selected>Sandwiches</calcite-tab-title>
-      </calcite-tab-nav>
-      <calcite-tab style="--calcite-block-padding: 0;">
-        <div>Chicken with Pesto</div>
-      </calcite-tab>
-    </calcite-tabs>
-  </calcite-tabs>
-`;
 
 describe("calcite-tab", () => {
   const tabHtml = "<calcite-tab>A tab</calcite-tab>";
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/tab/tab.e2e.ts
```diff
@@ -20,4 +36,32 @@ describe("calcite-tab", () => {
       { propertyName: "scale", defaultValue: "m" },
     ]);
   });
+
+  it("should allow the CSS custom property to be overridden when applied to :root", async () => {
```
jcfranco, line 40: I think these tests are better suited as screenshot tests as they are more visual.

packages/calcite-components/src/components/tab/tab.e2e.ts
```diff
@@ -20,4 +36,32 @@ describe("calcite-tab", () => {
...
+    const overrideStyle = "0px";
+    const page = await newE2EPage();
+    await page.setContent(html`
+      <style>
+        :root {
+          --calcite-block-padding: ${overrideStyle};
+        }
+      </style>
+      ${tabBlockPadding}
```
jcfranco, line 49: This test is misleading because the CSS prop is being set both at the root and component-levels (because of the reused test HTML). Sidebar: [block tests](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/block/block.e2e.ts#L289-L292) share the same issue. cc @macandcheese

packages/calcite-components/src/components/tab/tab.scss
```diff
@@ -1,3 +1,11 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-block-padding: Specifies the padding of the block `default` slot.
```
jcfranco, line 6: CSS prop name and doc does not match the component they belong to.

jcfranco (changes requested):

packages/calcite-components/src/components/tab/tab.scss
Elijbet: I meant the 'block' here as y-padding. Is there a better way to name it?

ashetland: Should the be `--calcite-tab-content-padding`? I think that matches other similar var names.

jcfranco: All component-specific tokens share the following naming pattern:

`--calcite-<component>-<prop>`.

Interestingly, I couldn't find this in our conventions, so we should add an entry for this.

macandcheese: IMO with the name “content-padding” it should apply a full padding rule if overridden, not just block padding.

packages/calcite-components/src/components/tab/tab.e2e.ts
macandcheese: Could you open an issue for that - we can update Block to match the tests here.

packages/calcite-components/src/components/tab/tab.scss
jcfranco: With the ongoing component-token work, should this be `--calcite-tab-space`, `--calcite-tab-content-space`? cc @alisonailea

packages/calcite-components/src/components/tab/tab.e2e.ts
Elijbet: I'll open it right now.

Elijbet: https://github.com/Esri/calcite-design-system/issues/8630

packages/calcite-components/src/components/tab/tab.scss
Elijbet: Does this not have the same issue as above, where tab-space or tab-content-space would be expected to override full padding around, not just y-padding?

alisonailea: @Elijbet if it is only applied to the content element inside tab then the name should be `--calcite-tab-content-space`

macandcheese: This seems like a different approach than other components? IMO it isn’t descriptive for a component level property IMO. Is the “space” referencing space between the tab nav and the tab etc? If not -  Should we change other components that use padding to match this?

macandcheese: > It's the spacing around the tab content, right?
Yeah definitely; just meant it’s not a self explanatory name like “padding”.

eriklharper: > All component-specific tokens share the following naming pattern:
> 
> `--calcite-<component>-<prop>`.
> 
> Interestingly, I couldn't find this in our conventions, so we should add an entry for this.

I was thinking about this the other day and using the Token Naming Schema as a guide, you could expand the template to something like:

`--calcite-<component>-<stencil-prop?>-<css-property?>-<modifier?>`

Where the `?` denote optional, so you could end up with a var like `--calcite-tile-background-color` which leaves out the stencil prop since that isn't applicable, but you could also have a var called `--calcite-tile-heading-text-color` and `--calcite-tile-heading-text-color-hover` which use more of those elements.

alisonailea: Your right I think I need to be more explicit. It is preferable that we create a single `-space` token from which margin and padding are calculated and applied as `-internal-` variables. This should be set on the `:host`. If you are creating a variable that only applied to padding properties then it should end with `-padding`.

macandcheese: For the tile example, it’s also possible the hover color could be set by the user on a pseudo class right? Only internal or inaccessible containers would need to be exposed? Close buttons, etc.

jcfranco: > It is preferable that we create a single -space token from which margin and padding are calculated and applied as -internal- variables. This should be set on the :host. If you are creating a variable that only applied to padding properties then it should end with -padding.

We should decide on one or the other or set some guidelines to deal with the following scenario:

1. component adds `--calcite-component-content-space` (used internally for both margin/padding)
2. [🇫🇷2 hours later🇫🇷] component needs to support different props for margin and padding, so it adds `--calcite-component-content-space-padding` and `--calcite-component-content-space-margin`

Does `--calcite-component-content-space` need to be deprecated? If so, it means we need to support the 3 CSS props until the next breaking change.

LMK if I misread your suggestion.

jcfranco: Regardless, as @Elijbet stated, let's move forward w/ `--calcite-tab-content-padding` for consistency and we can revisit ☝️ soon.

packages/calcite-components/src/components/tab/tab.scss
```diff
@@ -1,3 +1,11 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-tab-content-space: Specifies the padding of the block `default` slot.
```
geospatialem, line 6: Should the css prop description fit the title a bit more? 

Maybe something like:

```suggestion
 * @prop --calcite-tab-content-space: Specifies the spacing of the component's content in the  `"default"` slot.
```

macandcheese: Shouldn’t this remain “padding”? Rest of the suggestion looks good.

geospatialem: Good catch 🎣 - updated the suggestion!

packages/calcite-components/src/components/tab/tab.scss
jcfranco: Actually, this might have to be `--calcite-tab-content-block-padding` (semi full circle, eh? 😅) since the content padding isn't equal and I don't think we can use a single CSS prop for block and inline padding if their values aren't equal.

jcfranco: Got @alisonailea's blessing on ☝️ (via Teams):

> yes. In this case `--calcite-tab-content-block-padding` is probably best

jcfranco (approved):

PR #8637, opened 2024-01-24
feat(accordion, accordion-item): refactor for custom props

alisonailea:
**Related Issue:** #7180 #4060 #4012

## Summary

Refactors accordion and accordion item for custom props and component tokens.

### DIFF
--- a/packages/calcite-components/src/components/accordion-item/accordion-item.scss
+++ b/packages/calcite-components/src/components/accordion-item/accordion-item.scss
@@ -1,20 +1,20 @@
 /**
  * CSS Custom Properties
  *
- * @prop --calcite-accordion-item-background: The background color of the accordion item.
- * @prop --calcite-accordion-item-border-color: The border color of the accordion item.
- * @prop --calcite-accordion-item-outline-color-focus: The outline color of an accordion item when it is in focus;
- * @prop --calcite-accordion-item-outline-color: The outline color of an accordion item;
+ * @prop --calcite-accordion-item-background-color: defines the background color of the accordion item. May be set by the accordion parent.
+ * @prop --calcite-accordion-item-border-color: defines the border color of the accordion item. May be set by the accordion parent.
+ * @prop --calcite-accordion-item-heading-text-color: defines the color of heading text.
+ * @prop --calcite-accordion-item-heading-background-color: defines the background color of the heading. Defaults to --calcite-accordion-item-background-color.
  * @prop --calcite-accordion-item-padding: Deprecated. Please use --calcite-accordion-spacing instead.
- * @prop --calcite-accordion-item-secondary-text-color-selected: The text color of the secondary text in the accordion item when selected.
- * @prop --calcite-accordion-item-text-color-selected: The text color of the accordion item when it is selected.
- * @prop --calcite-accordion-item-text-color: The text color of the accordion item.
+ * @prop --calcite-accordion-item-icon-color: defines the color of the icons in an accordion item.
+ * @prop --calcite-accordion-item-description-text-color: defines the color of the accordion item's description text.
+ * @prop --calcite-accordion-item-content-text-color: defines the color of the accordion item's content text.
  *
  */
 
 /*
  * Internal
- *
+ * --calcite-internal-accordion-item-outline-color: defines the outline color of an accordion item.
  * --calcite-internal-accordion-item-icon-rotation: defines the rotation of the icon element. This is altered by setting classes within the ShadowDom.
  * --calcite-internal-accordion-item-flex-direction: defines the flex layout direction of the elements within the accordion item.
  * --calcite-internal-accordion-item-padding: defines the padding around the header and content elements within the accordion item ShadowDom.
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/accordion-item/accordion-item.scss
```diff
@@ -1,195 +1,207 @@
...
-}
+/**
+ * CSS Custom Properties
+ *
+ * @prop --calcite-accordion-item-background: The background color of the accordion item.
+ * @prop --calcite-accordion-item-border-color: The border color of the accordion item.
+ * @prop --calcite-accordion-item-outline-color-focus: The outline color of an accordion item when it is in focus;
+ * @prop --calcite-accordion-item-outline-color: The outline color of an accordion item;
+ * @prop --calcite-accordion-item-padding: Deprecated. Please use --calcite-accordion-spacing instead.
+ * @prop --calcite-accordion-item-secondary-text-color-selected: The text color of the secondary text in the accordion item when selected.
```
macandcheese, line 9: Should this be `description` to match the prop name that renders the string?

packages/calcite-components/src/components/accordion-item/accordion-item.scss
```diff
@@ -1,195 +1,207 @@
...
+ * @prop --calcite-accordion-item-text-color-selected: The text color of the accordion item when it is selected.
+ * @prop --calcite-accordion-item-text-color: The text color of the accordion item.
```
macandcheese, line 11: Should this be `heading`? Users can already control the color of the content they slot - but this doesn't seem to convey this would change the `heading` rendered via prop.

packages/calcite-components/src/components/accordion-item/accordion-item.scss
```diff
@@ -1,195 +1,207 @@
...
-  --calcite-accordion-item-active-icon-rotation-rtl: calc(theme("rotate.180") * -1);
-}
+/**
+ * CSS Custom Properties
+ *
+ * @prop --calcite-accordion-item-background: The background color of the accordion item.
+ * @prop --calcite-accordion-item-border-color: The border color of the accordion item.
+ * @prop --calcite-accordion-item-outline-color-focus: The outline color of an accordion item when it is in focus;
+ * @prop --calcite-accordion-item-outline-color: The outline color of an accordion item;
+ * @prop --calcite-accordion-item-padding: Deprecated. Please use --calcite-accordion-spacing instead.
```
macandcheese, line 8: Is this still accurate following conversation yesterday about tabs. `--calcite-accordion-spacing` doesn't explain to me as an unfamiliar user what it will adjust. Is this for the Accordion Item content area?

packages/calcite-components/src/components/accordion/accordion.scss
```diff
@@ -1,41 +1,77 @@
+/**
+ * CSS Custom Properties
+ *
+ * @prop --calcite-accordion-space: defines space around and between accordion item elements. Other spacing is calculated at a +4.
+ * @prop --calcite-accordion-border-color: The border around an accordion.
+ * @prop --calcite-accordion-background: The background color of the accordion and accordion items.
+ * @prop --calcite-accordion-header-text-color: The header text color for accordion items.
```
macandcheese, line 7: These `heading` and `description` props are set on the Accordion Items themselves, do we need these at this level? If so, these don't seem to match the Accordion Item variable names.

packages/calcite-components/src/components/accordion/accordion.scss
```diff
@@ -1,41 +1,77 @@
+/**
+ * CSS Custom Properties
+ *
+ * @prop --calcite-accordion-space: defines space around and between accordion item elements. Other spacing is calculated at a +4.
```
macandcheese, line 4: Do we need to expose this? If it's truly between Accordion Items (not something we currently support), should it be `gap`? The `is calculated at a +4` is a little confusing.

macandcheese (commented):
Some questions re: naming / strucutre

packages/calcite-components/src/components/accordion/accordion.scss
```diff
@@ -2,13 +2,8 @@
  * CSS Custom Properties
  *
  * @prop --calcite-accordion-space: defines space around and between accordion item elements. Other spacing is calculated at a +4.
- * @prop --calcite-accordion-border-color: The border around an accordion.
+ * @prop --calcite-accordion-border-color: The border around an accordion and accordion items.
  * @prop --calcite-accordion-background: The background color of the accordion and accordion items.
```
macandcheese, line 6: For consistency, should we deprecate this `background` and add a `background-color` instead? Similar to https://github.com/Esri/calcite-design-system/pull/8662/files

Or, maybe just deprecate without replacement if this background color never "gets seen" because only the Accordion Item has visible background?

alisonailea: missed this one. Thanks!

packages/calcite-components/src/components/accordion/accordion.scss
alisonailea: `gap` is a different style prop. Since it applies to margin and padding I think the right naming convention is `-space`. However I think it does make sense to be explicit that the calculation is based off a semantic token.

macandcheese: Name makes sense, can it be `-internal-` though? Doesn't seem like we'd want this to change (at least for now?)

alisonailea: done.

PR #8645, opened 2024-01-25
feat(notice): Add component tokens

macandcheese:
**Related Issue:** #7180

## Summary
`--calcite-notice-background-color`: Specifies the background color of the component.
`--calcite-notice-border-color`: Specifies the border color of the component.
`--calcite-notice-shadow`: Specifies the box shadow of the component.
`--calcite-notice-corner-radius`: Specifies the corner radius of the component.
`--calcite-notice-icon-color`: Specifies the color of the component's icon property.
`--calcite-notice-close-icon-color`: Specifies the color of the component's `closable` icon color.
`--calcite-notice-close-background-color`: Specifies the color of the component's closable background color.
`--calcite-notice-close-background-color-hover`: Specifies the hover color of the component's closable background color.
`--calcite-notice-close-background-color-active`: Specifies the active color of the component's closable background color.
`--calcite-notice-close-background-color-focus`: Specifies the focus color of the component's closable background color.
`--calcite-notice-close-focus-color`: Specifies the focus color of the component's closable background color.

Adds a Chromatic test and a demo page example.


packages/calcite-components/src/components/notice/notice.scss
```diff
@@ -3,13 +3,24 @@
...
  *
- * @prop --calcite-notice-width: The width of the component.
+ * @prop --calcite-notice-background-color: Specifies the background color of the component.
+ * @prop --calcite-notice-border-color: Specifies the border color of the component.
+ * @prop --calcite-notice-box-shadow: Specifies the box shadow of the component.
+ * @prop --calcite-notice-corner-radius: Specifies the corner radius of the component.
+ * @prop --calcite-notice-icon-color: Specifies the color of the component's icon property.
+ * @prop --calcite-notice-close-icon-color: Specifies the color of the component's `closable` icon color.
+ * @prop --calcite-notice-close-background-color: Specifies the color of the component's `closable` background color.
+ * @prop --calcite-notice-close-background-color-hover: Specifies the hover color of the component's `closable` background color.
```
alisonailea, line 13: Remove state specific tokens. We can just update the single token based on the state of :host

packages/calcite-components/src/components/notice/notice.scss
```diff
@@ -71,18 +82,18 @@
 }
 
 .container {
-  @apply bg-foreground-1
-    pointer-events-none
+  @apply pointer-events-none
     my-0
     box-border
     hidden
     w-full
```
alisonailea, line 89: replace with `inline-size: var(--calcite-container-size-content-fluid);`

packages/calcite-components/src/components/notice/notice.scss
```diff
@@ -98,13 +109,13 @@
 }
 
 :host([open]) .container {
-  @apply shadow-1
-    pointer-events-auto
+  @apply pointer-events-auto
     flex
     max-h-full
```
alisonailea, line 114: `max-block-size: var(--calcite-container-size-content-fluid);`

packages/calcite-components/src/components/notice/notice.scss
```diff
@@ -98,13 +109,13 @@
...
     items-center
     border-2
```
alisonailea, line 116: `border-width: var(--calcite-border-width-md);`

packages/calcite-components/src/components/notice/notice.scss
```diff
@@ -98,13 +109,13 @@
...
     opacity-100;
```
alisonailea, line 117: `opacity: var(--calcite-opacity-full);`

alisonailea (changes requested):

packages/calcite-components/src/components/notice/notice.scss
macandcheese: For these - it's for the internally rendered close button. Users won't be able to do `calcite-notice:hover { --calcite-notice-close-background-color: red }`. We need these because the close button also isn't transparently overlaying backgrounds (which, have their own issues for theming but less immediately need overrides).

macandcheese: Same feedback as https://github.com/Esri/calcite-design-system/pull/8607#discussion_r1488278003

cc @jcfranco for confirmation.

jcfranco: @alisonailea We discussed host-level states and agreed to not add tokens for these, but we missed this scenario (styling states of internal components/elements). We'd need to allow these to properly support theming, right?

alisonailea: Ah yes that's true. For internal component state management we'll need to have state tokens. Thanks!

alisonailea (approved):

packages/calcite-components/src/components/notice/notice.scss
driskull: > Ah yes that's true. For internal component state management we'll need to have state tokens. Thanks!

Wouldn't it be better to be consistent across all components rather than have some be able to define hover states via tokens and others where you need to use an existing component along with some host selector?

Issue #8495, opened 2023-12-23
[Tile Select] Tile Select React components do not correctly display radio buttons / allow selection


pllcoster:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When using Calcite React components after upgrading to 2.1.0,  a Tile Select component with either no `type` property or `type="radio"`, no radio button appears and the tiles are not selectable.

![image](https://github.com/Esri/calcite-design-system/assets/146282673/3a61472a-8a3b-4b0d-9ddb-0ba5944f92f8)

### Expected Behavior

Expect Radio buttons to appear as default or if type is set to "radio"

Correct functionality with calcite 1.9.1 can be seen here:
https://codesandbox.io/p/sandbox/calcite-1-9-1-tile-select-zdt9t7?file=%2Fsrc%2FApp.js%3A22%2C49

### Reproduction Sample

https://codesandbox.io/p/sandbox/caclite-2-1-0-tile-select-wx4tzr?file=%2Fsrc%2FApp.js%3A22%2C49

### Reproduction Steps

Create CalciteTileSelect components as described above, with type set to radio or left as default. Note that in the code sandbox checkboxes also appear to be broken.

### Reproduction Version

2.1.0

### Relevant Info

calcite-components-react

### Regression?

Worked in 1.9

### Priority impact

p2 - want for current milestone

### Impact

We wish to use the tile component as a radio button in our app. With this broken we may have to swap to a different component to have a similar selection experience.

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [X] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Knowledge

pllcoster:
From the discussion:

Your comment pointed me in the right direction - it seems like previously we did not have to separately import `@esri/calcite-components/dist/components/calcite-radio-button` to have the radio button working in the tile select - but if I add this import after updating to 2.1.0, the radio buttons come back on the tile select element

geospatialem:
Spike to investigate if the component could automatically import components, possibly similar to `modal`.

jcfranco:
It looks like dynamic tags passed to `document.createElement` prevent the Stencil build from auto-defining them. Fortunately, the fix for this is simple and is feasible for the current release.

[...]

PR #8648, opened 2024-01-25
fix(tile-select): ensure supporting components are auto-defined

jcfranco:
**Related Issue:** #8495 

## Summary

This fixes an issue that prevented the Stencil build from including supporting components from being auto-defined in the `components` output target.

### DIFF
--- a/packages/calcite-components/src/components/tile-select/tile-select.tsx
+++ b/packages/calcite-components/src/components/tile-select/tile-select.tsx
@@ -275,9 +275,11 @@ export class TileSelect implements InteractiveComponent, LoadableComponent {
   // --------------------------------------------------------------------------
 
   private renderInput(): void {
-    this.input = document.createElement(
-      this.type === "radio" ? "calcite-radio-button" : "calcite-checkbox",
-    );
+    this.input =
+      this.type === "radio"
+        ? /* we need to call createElement(x) separately to ensure supporting components are properly bundled */
+          document.createElement("calcite-radio-button")
+        : document.createElement("calcite-checkbox");
     this.input.checked = this.checked;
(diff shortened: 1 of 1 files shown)


packages/calcite-components/src/components/tile-select/tile-select.tsx
```diff
@@ -275,9 +275,11 @@ export class TileSelect implements InteractiveComponent, LoadableComponent {
   // --------------------------------------------------------------------------
 
   private renderInput(): void {
-    this.input = document.createElement(
-      this.type === "radio" ? "calcite-radio-button" : "calcite-checkbox",
-    );
+    this.input =
+      this.type === "radio"
+        ? /* we need to call createElement(x) separately to ensure supporting components are properly bundled */
+          document.createElement("calcite-radio-button")
```
driskull, line 281: Looks good! 👍 

We might want to add this to a gotcha developer doc or add some kind of eslint for not having vars in createElement?

driskull (approved):
👍

packages/calcite-components/src/components/tile-select/tile-select.tsx
jcfranco: Great idea. Will see if we can automate this. `import`s also share this issue.

jcfranco: `import` is [already covered](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/.eslintrc.cjs#L77). Created https://github.com/Esri/calcite-design-system/issues/8651 for adding a custom rule to catch dynamic `createElement` tag usage.

PR #8649, opened 2024-01-26
fix: only show validation message when status='invalid'

benelan:
## Summary

The validation message for form components should only be displayed when the `status` property is "invalid".


driskull (approved):
This looks good @benelan 👏 

Should we have some screenshot & maybe e2e tests for this as well?

Issue #8647, opened 2024-01-25
validation-related props not preventing form submission


jcfranco:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

`pattern`, `minLength` and other related props should prevent form submission.

### Expected Behavior

Form submit is prevented when the component is in an invalid state.

### Reproduction Sample

https://codepen.io/jcfranco/pen/wvOPvEa

### Reproduction Steps

1. Type in "http://a"
2. Press the submit button

Note that the alert message is displayed due to the submit event not being blocked.

### Reproduction Version

2.3.0

### Relevant Info

The internal hidden input isn't getting all of the validation-related props set on it on initialization and some are missing (see https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/input/input.tsx#L886-L903).

### Priority impact

p1 - need for current milestone

### Impact

This impacts all form-related use cases that involve validation.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

jcfranco:
**Note**: `minlength` and `maxlength` won't trigger constraint validation unless a user interacts with the input (see [`minlength` spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#setting-minimum-input-length-requirements:-the-minlength-attribute:~:text=Constraint%20validation%3A%20If%20an%20element%20has%20a%20minimum,element%20is%20suffering%20from%20being%20too%20short.) and [`maxlength` spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#setting-minimum-input-length-requirements:-the-minlength-attribute:~:text=Constraint%20validation%3A%20If%20an%20element%20has%20a%20maximum,element%20is%20suffering%20from%20being%20too%20long.)). This will have to be a known limitation until #8126 is completed.

Issue #8126, opened 2023-11-04
Epic: Update form components to leverage ElementInternals API


jcfranco:
**monday.com sync:** #18048583564

### Description

Lit is looking to [add support for `ElementInternals`](https://github.com/lit/rfcs/pull/45), so we should explore leveraging this API instead of our custom solution. 

cc @eriklharper @benelan 

Based on recent discussion, we'll use this issue to track just replacing the existing `HiddenFormInputSlot` implementation with `ElementInternals` without changing any existing public APIs.  This will keep the refactor label consistent too since no functionality would change as a result.

### Proposed Advantages

Improved form integration and leaner components.

### Which Component

Checkbox, Combobox, Input Date Picker, Input Number, Input Text, Input Time Picker, Input Time Zone, Input, Meter, Radio Button, Rating, Segmented Control, Select, Slider, Switch, Text Area

### Relevant Info

Safari added support for `ElementInternals` at 16.4, so we may have to consider the following if we update form components to use this API:

- [x] waiting until Safari 18 to ensure last 2 **base** versions support this API
- [x] ~consider [a polyfill](https://www.npmjs.com/package/element-internals-polyfill) to handle Safari 16.0 – 16.3 (requires a lot testing)~
- [x]  ~moving forward and document form integration won't work for 16.0 – 16.3 (nonstarter option as it introduces a regression for this version)~

All components implementing `FormComponent`:

* [x] `autocomplete` @jcfranco #14182 
* [x] `checkbox` @eriklharper #14052 
* [x] `combobox` @jcfranco #14056
* [x] `input-date-picker` @jcfranco https://github.com/Esri/calcite-design-system/pull/14072
* [x] `input-number` @jcfranco https://github.com/Esri/calcite-design-system/pull/14072
* [x] `input-text` @eriklharper #13985 
* [x] `input-time-picker` @jcfranco https://github.com/Esri/calcite-design-system/pull/14072
* [x] `input-time-zone` @jcfranco https://github.com/Esri/calcite-design-system/pull/14135
* [x] `input` @jcfranco https://github.com/Esri/calcite-design-system/pull/14073
* [x] ~`meter` @jcfranco~ [no longer applicable](https://github.com/Esri/calcite-design-system/pull/14071)
* [x] `radio-button` @eriklharper https://github.com/Esri/calcite-design-system/pull/14142
* [x] `rating`  @jcfranco https://github.com/Esri/calcite-design-system/pull/14069
[...]

PR #8655, opened 2024-01-26
fix(input, input-number, input-text, text-area): ensure all applicable props are considered in form validation

jcfranco:
**Related Issue:** #8647 

## Summary

This ensures `pattern`, `minlength`, `maxlength`, `min`, `max`, `step` are set on the internal form input based on the matching type.

**Note**: `minlength` and `maxlength` won't trigger constraint validation unless a user interacts with the input (see [`minlength` spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#setting-minimum-input-length-requirements:-the-minlength-attribute:~:text=Constraint%20validation%3A%20If%20an%20element%20has%20a%20minimum,element%20is%20suffering%20from%20being%20too%20short.) and [`maxlength` spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#setting-minimum-input-length-requirements:-the-minlength-attribute:~:text=Constraint%20validation%3A%20If%20an%20element%20has%20a%20maximum,element%20is%20suffering%20from%20being%20too%20long.)). This will have to be a known limitation until #8126 is completed.

### DIFF
--- a/packages/calcite-components/src/components/input/common/input.spec.ts
+++ b/packages/calcite-components/src/components/input/common/input.spec.ts
@@ -14,7 +14,10 @@ describe("common input utils", () => {
 
     allTypes.forEach((type) => {
       syncHiddenFormInput(type, allValueFakeInputComponent, hiddenFormInput);
-      expect(hiddenFormInput.type).toBe(type);
+
+      const expectedType = type === "textarea" ? "text" : type;
+
+      expect(hiddenFormInput.type).toBe(expectedType);
 
       if (minMaxStepTypes.includes(type)) {
         expect(hiddenFormInput.min).toBe("0");
--- a/packages/calcite-components/src/components/input/common/input.spec.ts
+++ b/packages/calcite-components/src/components/input/common/input.spec.ts
@@ -1,0 +1,35 @@
+/* eslint-disable jest/no-conditional-expect -- Using conditional logic in a confined test helper to handle specific scenarios, reducing duplication, balancing test readability and maintainability. **/
+import { minMaxLengthTypes, minMaxStepTypes, patternTypes, syncHiddenFormInput } from "./input";
+
+describe("common input utils", () => {
+  it("syncHiddenFormInput", async () => {
+    const minMaxLengthTestValues = { minLength: 0, maxLength: 10 };
+    const patternTestValue = { pattern: "test" };
+    const minMaxStepTestValues = { min: 0, max: 10, step: 1 };
+
+    const allTypes = Array.from(new Set([...minMaxLengthTypes, ...patternTypes, ...minMaxStepTypes]));
+    const allValueFakeInputComponent = { ...minMaxLengthTestValues, ...minMaxStepTestValues, ...patternTestValue };
+
+    const hiddenFormInput = document.createElement("input");
+
+    allTypes.forEach((type) => {
+      syncHiddenFormInput(type, allValueFakeInputComponent, hiddenFormInput);
+      expect(hiddenFormInput.type).toBe(type);
+
+      if (minMaxStepTypes.includes(type)) {
+        expect(hiddenFormInput.min).toBe("0");
+        expect(hiddenFormInput.max).toBe("10");
+        expect(hiddenFormInput.step).toBe("1");
+      }
+
+      if (minMaxLengthTypes.includes(type)) {
+        expect(hiddenFormInput.minLength).toBe(0);
+        expect(hiddenFormInput.maxLength).toBe(10);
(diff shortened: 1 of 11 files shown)


packages/calcite-components/src/components/input/common/input.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    const step = numericInputComponent.step;
+    hiddenFormInput.step = step === "any" ? step : toString(step);
+  }
+
+  if (minMaxLengthTypes.includes(type)) {
+    const textualInputComponent = inputComponent as TextualInputComponent;
+
+    if (hiddenFormInput.minLength != null) {
+      hiddenFormInput.minLength = textualInputComponent.minLength;
```
benelan, line 50: This should be `minlength` (and `maxlength` below) on the hidden input right?

benelan (commented):
Adding a couple comments; I'll take another look once this becomes ready for review. Thanks for putting this together so quickly!

packages/calcite-components/src/components/input/common/input.ts
```diff
@@ -0,0 +1,62 @@
...
+  type: HTMLInputElement["type"] | "textarea",
+  inputComponent: InputComponent,
+  hiddenFormInput: HTMLInputElement,
+): void {
+  hiddenFormInput.type = type === "textarea" ? "text" : type;
+
+  if (minMaxStepTypes.includes(type)) {
+    const numericInputComponent = inputComponent as NumericInputComponent;
+    hiddenFormInput.min = toString(numericInputComponent.min);
+    hiddenFormInput.max = toString(numericInputComponent.max);
```
benelan, line 40: I mentioned this on Teams, but adding it here for posterity - I think we will lose the extra precision from BigDecimal when validating the `min`/`max` constraints. I'm guessing the string is cast to a number internally when making the comparison. For example, I'm guessing the form will treat this as valid when it shouldn't:

```html
<calcite-input-number value="1.123456789123456789" max="1.123456789123456788">
```

```js
console.log(1.123456789123456789 > 1.123456789123456788) // => false
```

I don't know if we should care about that until someone logs an issue, but we should at least add a code comment for future reference.

benelan: The potential solution I had in mind was adding custom logic in our `reportValidity()` method, which uses BigDecimal to make the comparison. We will likely need to do the same for other components like input-date-picker and input-time-picker anyway. 

We would need to call `reportValidity()` on all the form components with custom validation logic when submitting the form. If our logic finds invalid values, we'd call `setValidity()` on the hidden inputs to prevent form submission.

jcfranco: I dig the beefed-up validation logic. 💪

I added a note regarding expected loss of precision in some cases based on your comment.

packages/calcite-components/src/components/input/common/input.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+      syncHiddenFormInput(type, allValueFakeInputComponent, hiddenFormInput);
+
+      const expectedType = type === "textarea" ? "text" : type;
+
+      expect(hiddenFormInput.type).toBe(expectedType);
+
+      if (minMaxStepTypes.includes(type)) {
+        expect(hiddenFormInput.min).toBe("0");
+        expect(hiddenFormInput.max).toBe("10");
+        expect(hiddenFormInput.step).toBe("1");
```
benelan, line 25: Nitpick: We could do the following so it's easier to change the values while testing locally

```js
expect(hiddenFormInput.min).toBe(`${minMaxStepTestValues.min}`);
expect(hiddenFormInput.max).toBe(`${minMaxStepTestValues.max}`);
expect(hiddenFormInput.step).toBe(`${minMaxStepTestValues.step}`);
```

Same idea applies to the assertions below.

benelan (approved):
LGTM! 

Once this is installed I can work on using these utils in input-date-picker and input-time-picker so we can add validation for the min/max constraints. 

I'll try comparing the date/time to min/max during the sync process, and then call setValidity() on the hidden input if invalid. That would be a pretty clean approach if it works.

packages/calcite-components/src/components/input/common/input.spec.ts
jcfranco: I’ll clean these up in a separate PR.

jcfranco: ☝️ https://github.com/Esri/calcite-design-system/pull/8733

Issue #8651, opened 2024-01-26
Add rule to prevent dynamic tag usage with `createElement`


jcfranco:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Stems from https://github.com/Esri/calcite-design-system/pull/8648#discussion_r1467083471. 

Dynamic tags passed to `document.createElement` prevent the Stencil build from auto-defining them. This rule should help ensure static element tag usage.

### Acceptance Criteria

Code like the following is flagged

```ts
const tag = someCondition ? "my-component-1" : "my-component-2";
const el = document.createElement(tag);
```

and the following is fine

```ts
const el = someCondition ? 
  document.createElement("my-component-1") : 
  document.createElement("my-component-2");
```

### Which Component

Any component that is dynamically creating supporting components.

### Priority impact

p4 - not time sensitive

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [X] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

DitwanP:
🍠 Verified locally on `main`

The rule is doing what rules do. 🛑
![Screenshot 2024-01-31 at 1 37 47 PM](https://github.com/Esri/calcite-design-system/assets/60022782/0a8ec695-316f-4f2f-abbe-79b6c3cbb28b)

PR #8656, opened 2024-01-26
feat: add `no-dynamic-createelement` rule

jcfranco:
**Related Issue:** #8651

## Summary

This adds a custom ESLint rule to ensure Stencil bundles and auto-defines child components created via `document.createElement`.


driskull (approved):
🚀 👍

PR #8657, opened 2024-01-26
fix(combobox): ensure supporting components are auto-defined

jcfranco:
**Related Issue:** #8495 

## Summary

This fixes an issue that prevented the Stencil build from including supporting components from being auto-defined in the components output target. 

This probably wasn't noticeable since in `combobox` and `combobox-item`s are expected to be used together.


driskull (approved):
👍 Nice

PR #8660, opened 2024-01-26
feat(link): add component tokens

DitwanP:
**Related Issue:** #7180
## Summary

Add the following component tokens to the `link` component:
 * `--calcite-link-text-color`: Specifies the text color of the component.
 * `--calcite-link-icon-start-color`: Specifies the color of the component's start position icon.
 * `--calcite-link-icon-end-color`: Specifies the color of the component's end position icon.
 
 Also fixed the end icon demo for `link`. (the icon was set to the start position)


packages/calcite-components/src/components/fab/fab.scss
```diff
@@ -1,17 +1,33 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-fab-shadow: Specifies the shadow of the component.
+ * @prop --calcite-fab-shadow-hover: Specifies the shadow of the component when in hover state.
+ * @prop --calcite-fab-shadow-active: Specifies the shadow of the component when in active state.
+ */
```
macandcheese, line 9: Should we expose background-color, border-color, corner-radius, text-color here? cc @SkyeSeitz @ashetland @alisonailea

packages/calcite-components/src/components/fab/fab.scss
```diff
@@ -1,17 +1,33 @@
...
+  --calcite-fab-shadow-hover: var(--calcite-shadow-md);
+  --calcite-fab-offset-shadow: 0 0 #0000;
+  background-color: var(--calcite-color-transparent);
 }
 
-@include disabled();
-
 calcite-button {
-  @apply shadow-2;
+  box-shadow: var(--calcite-fab-offset-shadow), var(--calcite-fab-offset-shadow), var(--calcite-fab-shadow);
```
eriklharper, line 21: ☝️ There are 3 different css custom properties used in the `shadow-2` tailwind util that don't seem to be completely replicated here.

image

packages/calcite-components/src/components/fab/fab.scss
```diff
@@ -1,17 +1,33 @@
...
+ * @prop --calcite-fab-shadow-active: Specifies the shadow of the component when in active state.
+ */
+
 :host {
-  @apply flex bg-transparent;
+  @apply flex;
+  --calcite-fab-shadow: 0 6px 20px -4px rgba(0, 0, 0, 0.1), 0 4px 12px -2px rgba(0, 0, 0, 0.08);
+  --calcite-fab-shadow-active: 0 2px 12px -4px rgba(0, 0, 0, 0.2), 0 2px 4px -2px rgba(0, 0, 0, 0.16);
+  --calcite-fab-shadow-hover: var(--calcite-shadow-md);
+  --calcite-fab-offset-shadow: 0 0 #0000;
```
eriklharper, line 16: ☝️ `--calcite-fab-offset-shadow` is missing in the documentation above.

packages/calcite-components/src/components/fab/fab.scss
DitwanP: So I was confused about the shadow-2, so I asked Ali for suggestions and here's two suggestions they gave

```
:host {
  @apply flex;
  
  --calcite-fab-shadow: 0 6px 20px -4px rgba(0, 0, 0, 0.1), 0 4px 12px -2px rgba(0, 0, 0, 0.08);
  --calcite-fab-shadow-active: 0 2px 12px -4px rgba(0, 0, 0, 0.2), 0 2px 4px -2px rgba(0, 0, 0, 0.16);
  --calcite-fab-shadow-hover: var(--calcite-shadow-md);

  --calcite-fab-offset-shadow:  0 0 #0000;

  --calcite-button-box-shadow: var(--calcite-fab-offset-shadow), var(--calcite-fab-offset-shadow), var(--calcite-fab-shadow);

  background-color: var(--calcite-color-transparent);
  &:hover {
    --calcite-button-box-shadow: var(--calcite-fab-offset-shadow), var(--calcite-fab-offset-shadow), var(--calcite-fab-shadow-hover);
  }
  &:active {
    --calcite-button-box-shadow: var(--calcite-fab-offset-shadow), var(--calcite-fab-offset-shadow), var(--calcite-fab-shadow-active);
  }
}

@include disabled();

@include base-component();
```

and 


```
:host {
  @apply flex;
  
  --calcite-fab-shadow: 0 6px 20px -4px rgba(0, 0, 0, 0.1), 0 4px 12px -2px rgba(0, 0, 0, 0.08);
  --calcite-button-box-shadow: var(--calcite-internal-fab-offset-shadow), var(--calcite-internal-fab-offset-shadow), var(--calcite-fab-shadow);
  
  --calcite-internal-fab-offset-shadow:  0 0 #0000;

  background-color: var(--calcite-color-transparent);
  
  &:hover {
      --calcite-fab-shadow: var(--calcite-shadow-md);
  }
  
  &:active {
    --calcite-fab-shadow: 0 2px 12px -4px rgba(0, 0, 0, 0.2), 0 2px 4px -2px rgba(0, 0, 0, 0.16);
  }
}

@include disabled();

@include base-component();
```
The problem I had with those suggestions is that they were not actually applying the shadows because the shadows exists on the `calcite-button` within the fab. So I made some tweaks to the suggestion and visually inspected the component to see if it looked the same as before adding the tokens. It did look similar but I also wanted to see if there would be any changes caught by chromatic.

macandcheese: Is it possible to use tokens to set shadow on the internally rendered Button vs. applying them on the FAB itself? Would require Button to land first.

packages/calcite-components/src/components/fab/fab.scss
macandcheese: I would think this can be an `-internal` prop.

packages/calcite-components/src/components/fab/fab.scss
macandcheese: cc @alisonailea for above

packages/calcite-components/src/components/fab/fab.scss
ashetland: Are we doing that for Button? Would make sense to match, if so.

packages/calcite-components/src/components/fab/fab.scss
DitwanP: Two opposing options, what do you guys think we should go with? Would a user want to be able to change the offset shadow? @macandcheese @eriklharper

macandcheese: I'm not sure what this would be used for. Is this instead of a focus override? See this comment about perhaps setting css property overrides on the internal button directly https://github.com/Esri/calcite-design-system/pull/8660#discussion_r1468210782, which I think we should look into.

macandcheese: Right now FAB applies its own shadow around the internally rendered Button component. But we didn't have the Button css properties in the past to directly set them on Button. So, I think we can do that now, like what you have above.

But, I don't understand the offset-shadow rule. What is that used for? I'd suggest just getting rid of that if possible. Users should be able to override a focus color to achieve that?

alisonailea: @DitwanP @eriklharper offset-shadow should not be a token.

packages/calcite-components/src/components/fab/fab.scss
alisonailea: Token must be refactored to only set it's tokens(css vars) on :host. I'll have a PR up today with a button refactor that fixes the problem with passing tokens down. In the case of `fab` and other components intended to wrap specific calcite-components. I think the best pattern would be to simply document the component tokens for the wrapped component in the docs for this component. Because of the way CSS variables pierce the shadowdom any CSS tokens set in the `:host` of one component will be lifted up into it's parent `:host` when wrapped within the shadow dom of another component.

packages/calcite-components/src/components/fab/fab.scss
```diff
@@ -1,17 +1,34 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-fab-shadow: Specifies the shadow of the component.
+ * @prop --calcite-fab-shadow-hover: Specifies the shadow of the component when in hover state.
+ * @prop --calcite-fab-shadow-active: Specifies the shadow of the component when in active state.
+ * @prop --calcite-fab-offset-shadow: Specifies the offset shadow of the component.
```
alisonailea, line 9: Remove these. After looking it over again, I don't think fab has any of it's own tokens. We should direct customers to the button component for a list of themeable variables.

packages/calcite-components/src/components/fab/fab.scss
alisonailea: Sorry @DitwanP. I know this is different than the code samples I sent you initially.

packages/calcite-components/src/components/fab/fab.scss
```diff
@@ -1,17 +1,34 @@
...
 
-@include disabled();
-
 calcite-button {
-  @apply shadow-2;
+  box-shadow: var(--calcite-fab-offset-shadow), var(--calcite-fab-offset-shadow), var(--calcite-fab-shadow);
+
   &:hover {
-    @apply shadow-2-lg;
+    box-shadow: var(--calcite-fab-offset-shadow), var(--calcite-fab-offset-shadow), var(--calcite-fab-shadow-hover);
```
alisonailea, line 25: Is fab actually supposed to be adding a second drop shadow around the button? Or just taking control of the existing dropshadow of calcite-button?

packages/calcite-components/src/components/link/link.scss
```diff
@@ -1,5 +1,18 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-link-text-color: Specifies the text color of the component.
+ * @prop --calcite-link-icon-start-color: Specifies the color of the component's start position icon.
```
alisonailea, line 7: Do we have requests for each icon to be a different color? If not I think we could reduce this to `--calcite-link-icon-color`

alisonailea (changes requested):

packages/calcite-components/src/components/fab/fab.scss
DitwanP: I think it should be taking control of the existing one.

packages/calcite-components/src/components/link/link.scss
DitwanP: Adam suggested this pattern for the button, and I think it makes sense to allow it. I can't think of an example at the moment but it's not that much extra work and best case is that if there does happen to be cases for it that pop up then it's already there. But LMK what you think and maybe we change it for button as well?

packages/calcite-components/src/components/fab/fab.scss
DitwanP: Yeah I was just starting to realize this and was going to ask about it but then found these comments.

macandcheese: It will be really hard for us to document the component-tokens if they aren't available on the component itself. 

Even if --calcite-fab-shadow just sets --calcite-button-shadow internally, I think that having a set of `--calcite-fab-x` would be more expected DX / would be able to show on the documentation site.

alisonailea (approved):

PR #8662, opened 2024-01-29
feat(scrim): Add component tokens

macandcheese:
**Related Issue:** #7180 

## Summary
`--calcite-scrim-background-color`: Specifies the background color of the component.

Adds a deprecated tag and fallback for the previous `--calcite-scrim-background` css property.


eriklharper (approved):
Change looks good.  This question applies to any of these new component token changes, but I'm wondering if we should add a storybook story that overrides all of the publicly documented tokens for visual regression tracking?

alisonailea (approved):

PR #8663, opened 2024-01-29
fix(chip): Prevent rendering internal icon if not necessary.

driskull:
**Related Issue:**#8674

## Summary

- Prevents rendering the internal calcite-icon if no selection icon is specified
- Causes screenshot tests in the Maps SDK to fail because its expecting/waiting for an icon to load and it will never load since no icon is specified
- Maybe we should make the icon property required on `calcite-icon`?

### DIFF
--- a/packages/calcite-components/src/components/chip/chip.tsx
+++ b/packages/calcite-components/src/components/chip/chip.tsx
@@ -340,7 +340,7 @@ export class Chip
           [CSS.selectIconActive]: this.selectionMode === "multiple" || this.selected,
         }}
       >
-        <calcite-icon icon={icon} scale={getIconScale(this.scale)} />
+        {icon ? <calcite-icon icon={icon} scale={getIconScale(this.scale)} /> : null}
       </div>
     );
   }
--- a/packages/calcite-components/src/components/chip/chip.e2e.ts
+++ b/packages/calcite-components/src/components/chip/chip.e2e.ts
@@ -27,6 +27,18 @@ describe("calcite-chip", () => {
     disabled("<calcite-chip interactive>doritos</calcite-chip>");
   });
 
+  it("should not render a calcite-icon when selectionMode is single and not selected", async () => {
+    const page = await newE2EPage();
+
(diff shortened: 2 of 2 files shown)


packages/calcite-components/src/components/chip/chip.tsx
```diff
@@ -340,7 +340,7 @@ export class Chip
           [CSS.selectIconActive]: this.selectionMode === "multiple" || this.selected,
         }}
       >
-        <calcite-icon icon={icon} scale={getIconScale(this.scale)} />
+        {icon ? <calcite-icon icon={icon} scale={getIconScale(this.scale)} /> : null}
```
macandcheese, line 343: This is only the selection mode icon - the `icon` property icon is already conditionally rendered. How are they experiencing the error? Maybe the `this.selectionMode !== "none"` isn't catching all cases?

I don't think the `icon` property should be required, Chips only have icons specified occasionally.

macandcheese: Ah, yeah makes sense, its the "inactive" selection icon. This looks to be a safe change, I can't see any issues with animation or anything in local testing with the proposed change.

The `icon` property is separate.

driskull: Yeah, not the property. I was thinking maybe we make the `icon` property on `calcite-icon` required to avoid this though?

The issue is that the JSAPI screenshot tests looks for calcite-icon and waits for it to have a path before proceeding. In this case it never gets a path.

cc @jcfranco

macandcheese: Oh, sorry, missed the "required on Calcite Icon" part. Carry on :)

Maybe if no icon is provided to Calcite Icon, we render "blank" icon instead of making required? In this case it would have reserved space but maybe it's a better fallback.

driskull: yeah that could work as well.

macandcheese (approved):
Nice fix, thanks!

Issue #8667, opened 2024-01-30
Nested List Items no longer correctly align


ashetland:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When any items are nested in a List, space for the expand affordance is not accounted for in all items, breaking nested alignment.
![CleanShot 2024-01-29 at 16 03 58@2x](https://github.com/Esri/calcite-design-system/assets/108549080/d2706c0c-5d14-491a-8629-848279e3cb85)

### Expected Behavior

When nested items are present in a List, the space for the expand affordance is accounted for in all items to maintain proper alignment.
![CleanShot 2024-01-29 at 16 03 36@2x](https://github.com/Esri/calcite-design-system/assets/108549080/1848cdf1-624b-4978-8666-6b7b27f77082)

### Reproduction Sample

https://codepen.io/ashetland/pen/VwRyOQa?editors=1000

### Reproduction Steps

Expand the second item and note the alignment of the children as well as the item at the top that has no children.

### Reproduction Version

v2.3.0

### Regression?

v2.1.0

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

ashetland:
Correcting this bug will result in the visual change depicted below
![Frame 304](https://github.com/Esri/calcite-design-system/assets/108549080/f890910a-aec8-498c-9e7c-9f8b03c2ef1d)

ashetland:
This [Figma file](https://www.figma.com/file/91sfHrYNt0Il6B4pEZZ749/List-spacing%2C-sorting%2C-group-heading-%5B7100%5D-%5B8039%5D-%5B8242%5D-%5B8667%5D?type=design&node-id=982%3A37071&mode=dev) covers designs for this issue and multiple related design revisions for List. cc @driskull

driskull:
@ashetland @SkyeSeitz  I have some concerns about the design summary...

The summary says that the "classic" mode should maintain existing padding and spacing. That would mean that each mode would have its own styles. Id rather ovoid that if necessary because its harder to maintain. I thought the purpose of these modes would just be how the chevron is rendered and takes up space?

[...]

PR #8668, opened 2024-01-30
fix(list-item): Always reserve spacing for open icon

driskull:
**Related Issue:** #8667

## Summary

- Always reserve space for open toggle icon
- Screenshot tests show changes


ashetland:
I just noticed that the "Nested List" page isn't expanded in the baseline or new so chromatic isn't showing the changes. It might be worth expanding all the items in that page so we have it as a baseline.

ashetland:
Also the baseline is showing the old icons?
CleanShot 2024-01-30 at 11 02 14@2x

driskull:
I think those icons are slotted in, not in the component

alisonailea (approved):

driskull:
@ashetland is this one good to merge?

ashetland:
Sadly no. We got feedback that folks aren't super happy about the added space in a flat list scenario. I'm drawing up specs for revised spacing in both flat and nested lists as there was a strong desire for this to be configurable. We'll need to push implementation to March. cc @jcfranco

driskull:
> there was a strong desire for this to be configurable

I think that would be a breaking change if we required people to configure allowing nesting and doesn't seem very user friendly to have to do.

If the reserved space for a nesting icon is an issue maybe it should be moved to the opposite side like the block/accordion components?

PR #8670, opened 2024-01-30
feat(panel): Add component tokens

driskull:
**Related Issue:** #7180

## Summary

- Adds screenshot theming test

### Adds the following public component css properties:
 
 * `--calcite-panel-background-color`: Specifies the background color of the component.
 * `--calcite-panel-border-color`: Specifies the border color of the component.
 * `--calcite-panel-description-text-color`: Specifies the color of the component's description.
 * `--calcite-panel-footer-background-color`: Specifies the background color of the component's footer.
 * `--calcite-panel-footer-space`: Specifies the spacing of the component's footer.
 * `--calcite-panel-header-background-color`: Specifies the component header's background color.
 * `--calcite-panel-heading-text-color`: Specifies the component header's  color.
 * `--calcite-panel-header-z-index`: Specifies the component header's z-index.
 * `--calcite-panel-fab-z-index`: Specifies the component fab's z-index.

## Deprecates the following public component css properties:

 - `--calcite-panel-footer-padding`: [Deprecated] Use `--calcite-panel-footer-space` instead. Specifies the padding of the component's footer.
- `--calcite-panel-header-border-block-end`: [Deprecated] No longer necessary. Specifies the component header's block end border.
-  `--calcite-panel-header-color`: [Deprecated] Use `--calcite-panel-heading-text-color` instead. Specifies the component header's  color.

### DIFF
--- a/packages/calcite-components/src/components/panel/panel.scss
+++ b/packages/calcite-components/src/components/panel/panel.scss
@@ -19,8 +19,8 @@
   @extend %component-host;
   position: relative;
   display: flex;
-  inline-size: 100%;
-  block-size: 100%;
+  inline-size: var(--calcite-container-size-content-fluid);
+  block-size: var(--calcite-container-size-content-fluid);
   flex: 1 1 auto;
   --calcite-internal-panel-transition: max-block-size var(--calcite-animation-timing),
     inline-size var(--calcite-animation-timing);
@@ -43,9 +43,14 @@
 .header {
   color: var(--calcite-panel-header-color);
   margin: 0;
-  display: flex;
   align-content: space-between;
   align-items: center;
+  display: flex;
+  flex: 1;
+  flex-direction: column;
+  z-index: var(--calcite-panel-header-z-index);
+  background-color: var(--calcite-panel-header-background-color);
+  border-block-end: var(--calcite-panel-header-border-block-end);
 }
 
 .heading {
@@ -63,7 +68,7 @@
   background-color: var(--calcite-panel-background-color);
   margin: 0;
   display: flex;
-  inline-size: 100%;
+  inline-size: var(--calcite-container-size-content-fluid);
   flex: 1 1 auto;
   flex-direction: column;
   align-items: stretch;
@@ -75,18 +80,10 @@
   display: none;
 }
 
-.header {
-  display: flex;
-  flex-direction: column;
-  z-index: var(--calcite-panel-header-z-index);
-  background-color: var(--calcite-panel-header-background-color);
-  border-block-end: var(--calcite-panel-header-border-block-end);
-}
-
 .header-container {
   display: flex;
   flex-direction: row;
-  inline-size: 100%;
+  inline-size: var(--calcite-container-size-content-fluid);
   align-items: stretch;
   justify-content: flex-start;
   flex: 0 0 auto;
@@ -97,11 +94,11 @@
 }
 
 .action-bar-container {
-  inline-size: 100%;
+  inline-size: var(--calcite-container-size-content-fluid);
 }
 
 .action-bar-container ::slotted(calcite-action-bar) {
-  inline-size: 100%;
+  inline-size: var(--calcite-container-size-content-fluid);
 }
 
 .header-content {
@@ -122,7 +119,7 @@
     margin-block: 0px var(--calcite-spacing-xxs);
     font-weight: var(--calcite-font-weight-medium);
     font-size: var(--calcite-font-size-0);
-    line-height: 1.25rem;
+    line-height: var(--calcite-font-line-height-relative-tight);
     &:only-child {
       margin-block-end: 0;
     }
@@ -130,7 +127,7 @@
   .description {
     color: var(--calcite-panel-description-color);
     font-size: var(--calcite-font-size--1);
-    line-height: 1rem;
+    line-height: var(--calcite-font-line-height-relative);
   }
 }
 
@@ -160,12 +157,12 @@
   align-items: stretch;
   background-color: var(--calcite-panel-background-color);
   overflow: auto;
-  block-size: 100%;
+  block-size: var(--calcite-container-size-content-fluid);
 }
 
 .footer {
   background-color: var(--calcite-panel-footer-background-color);
-  inline-size: 100%;
+  inline-size: var(--calcite-container-size-content-fluid);
   display: flex;
   justify-content: space-evenly;
   flex: 0 0 auto;
--- a/packages/calcite-components/src/components/panel/panel.scss
+++ b/packages/calcite-components/src/components/panel/panel.scss
@@ -3,130 +3,184 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
+ * @prop --calcite-panel-background-color: Specifies the background color of the component.
+ * @prop --calcite-panel-border-color: Specifies the border color of the component.
+ * @prop --calcite-panel-description-color: Specifies the color of the component's description.
+ * @prop --calcite-panel-footer-background-color: Specifies the background color of the component's footer.
  * @prop --calcite-panel-footer-padding: Specifies the padding of the component's footer.
+ * @prop --calcite-panel-header-background-color: Specifies the component header's background color.
  * @prop --calcite-panel-header-border-block-end: Specifies the component header's block end border.
+ * @prop --calcite-panel-header-color: Specifies the component header's  color.
+ * @prop --calcite-panel-header-z-index: Specifies the component header's z-index.
+ * @prop --calcite-panel-fab-z-index: Specifies the component fab's z-index.
  */
 
 :host {
   @extend %component-host;
-  @apply relative flex w-full h-full flex-auto;
-
-  --calcite-min-header-height: calc(var(--calcite-icon-size) * 3);
+  position: relative;
+  display: flex;
+  inline-size: 100%;
+  block-size: 100%;
+  flex: 1 1 auto;
 }
 
 @include disabled();
 
-@import "../../assets/styles/header";
+.header {
+  color: var(--calcite-panel-header-color, var(--calcite-color-text-2));
+  fill: var(
+    --calcite-panel-header-color,
+    var(--calcite-color-text-2)
+  ); // question: not sure if this is needed. Not sure why it was there in the first place.  https://github.com/Esri/calcite-design-system/blame/2e8abc5cea258e188445a223b15c2e469d6baef5/src/assets/styles/_header.scss
+  margin: 0;
+  display: flex;
+  align-content: space-between;
+  align-items: center;
+}
+
+.heading {
+  margin: 0;
+  padding: 0;
+  font-weight: var(--calcite-font-weight-medium); // question: are we doing font-weight tokens?
+}
+
+.header .heading {
+  flex: 1 1 auto;
+  padding: var(--calcite-spacing-xxs);
+}
 
 .container {
-  @apply bg-background m-0 flex w-full flex-auto flex-col items-stretch p-0;
+  background-color: var(--calcite-panel-background-color, var(--calcite-color-background));
+  margin: 0;
+  display: flex;
+  inline-size: 100%;
+  flex: 1 1 auto;
+  flex-direction: column;
+  align-items: stretch;
+  padding: 0;
 
-  transition:
+  transition: var(
+    --calcite-internal-panel-transition,
     max-block-size var(--calcite-animation-timing),
-    inline-size var(--calcite-animation-timing);
+    inline-size var(--calcite-animation-timing)
+  );
 }
 
 .container[hidden] {
-  @apply hidden;
+  display: none;
 }
 
 .header {
-  @apply bg-foreground-1 flex flex-col
-    z-header;
-  border-block-end: var(--calcite-panel-header-border-block-end, 1px solid var(--calcite-color-border-3));
+  display: flex;
+  flex-direction: column;
+  z-index: var(--calcite-panel-header-z-index, var(--calcite-z-index-header));
+  background-color: var(--calcite-panel-header-background-color, var(--calcite-color-foreground-1));
+  border-block-end: var(
+    --calcite-panel-header-border-block-end,
+    var(--calcite-border-width-sm) solid var(--calcite-panel-border-color, var(--calcite-color-border-3))
+  );
 }
 
 .header-container {
-  @apply flex flex-row w-full
-  items-stretch
-  justify-start;
+  display: flex;
+  flex-direction: row;
+  inline-size: 100%;
+  align-items: stretch;
+  justify-content: flex-start;
   flex: 0 0 auto;
 }
 
 .header-container--border-end {
-  border-block-end: 1px solid var(--calcite-color-border-3);
+  border-block-end: var(--calcite-border-width-sm) solid
+    var(--calcite-panel-border-color, var(--calcite-color-border-3));
 }
 
 .action-bar-container {
-  @apply w-full;
+  inline-size: 100%;
 }
 
 .action-bar-container ::slotted(calcite-action-bar) {
-  @apply w-full;
+  inline-size: 100%;
 }
 
 .header-content {
-  @apply flex
-    flex-col
-    overflow-hidden
-    px-3
-    py-3.5;
+  flex-direction: column;
+  display: flex;
+  overflow: hidden;
+  padding-inline: var(--calcite-spacing-sm);
+  padding-block: var(--calcite-spacing-md);
   margin-inline-end: auto;
   .heading,
   .description {
-    @apply block
-      break-words
-      p-0;
+    display: block;
+    overflow-wrap: break-word;
+    padding: 0;
   }
   .heading {
-    @apply text-0h mx-0 mt-0 mb-1 font-medium;
+    margin-inline: 0px;
+    margin-block: 0px var(--calcite-spacing-xxs);
+    font-weight: var(--calcite-font-weight-medium);
+    font-size: var(--calcite-font-size-0); // question: we doing font sizes?
+    line-height: 1.25rem;
     &:only-child {
-      @apply mb-0;
+      margin-block-end: 0;
     }
   }
   .description {
-    @apply text-color-2 text-n1h;
+    color: var(--calcite-panel-description-color, var(--calcite-color-text-2));
+    font-size: var(--calcite-font-size--1); // question: we doing font sizes?
+    line-height: 1rem;
   }
 }
 
 .back-button {
-  @apply border-color-3
-  border-0
-  border-solid;
-  border-inline-end-width: theme("borderWidth.DEFAULT");
+  border-color: var(--calcite-panel-border-color, var(--calcite-color-border-3));
+  border-style: solid;
+  border-width: 0px;
+  border-inline-end-width: var(--calcite-border-width-sm);
 }
 
 .header-actions {
-  @apply flex
-  flex-row
-  flex-nowrap
-  items-stretch;
+  display: flex;
+  flex-direction: row;
+  flex-wrap: nowrap;
+  align-items: stretch;
 }
 
 .header-actions--end {
-  margin-inline-start: theme("margin.auto");
+  margin-inline-start: "auto";
 }
 
 .content-wrapper {
-  @apply flex
-  flex-auto
-  flex-col
-  flex-nowrap
-  items-stretch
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -3,130 +3,184 @@
...
+ * @prop --calcite-panel-background-color: Specifies the background color of the component.
+ * @prop --calcite-panel-border-color: Specifies the border color of the component.
+ * @prop --calcite-panel-description-color: Specifies the color of the component's description.
+ * @prop --calcite-panel-footer-background-color: Specifies the background color of the component's footer.
  * @prop --calcite-panel-footer-padding: Specifies the padding of the component's footer.
+ * @prop --calcite-panel-header-background-color: Specifies the component header's background color.
  * @prop --calcite-panel-header-border-block-end: Specifies the component header's block end border.
+ * @prop --calcite-panel-header-color: Specifies the component header's  color.
+ * @prop --calcite-panel-header-z-index: Specifies the component header's z-index.
+ * @prop --calcite-panel-fab-z-index: Specifies the component fab's z-index.
```
alisonailea, line 15: nitpick: alphabetize component token list.

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -3,130 +3,184 @@
...
+  block-size: 100%;
+  flex: 1 1 auto;
 }
 
 @include disabled();
 
-@import "../../assets/styles/header";
+.header {
+  color: var(--calcite-panel-header-color, var(--calcite-color-text-2));
+  fill: var(
```
alisonailea, line 31: Yeah agreed this shouldn't do anything. Remove? @jcfranco

packages/calcite-components/src/components/panel/panel.scss
alisonailea, line 15: nitpick: alphabetize component token list.

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -3,130 +3,184 @@
...
+  inline-size: 100%;
+  block-size: 100%;
+  flex: 1 1 auto;
 }
 
 @include disabled();
 
-@import "../../assets/styles/header";
+.header {
+  color: var(--calcite-panel-header-color, var(--calcite-color-text-2));
```
alisonailea, line 30: nitpick: Setting initial values of component tokens in `:host` makes the code easier review.

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -3,130 +3,184 @@
...
+  align-content: space-between;
+  align-items: center;
+}
 
-.container {
-  @apply bg-background m-0 flex w-full flex-auto flex-col items-stretch p-0;
+.heading {
+  margin: 0;
+  padding: 0;
+  font-weight: var(--calcite-font-weight-medium); // question: are we doing font-weight tokens?
```
alisonailea, line 44: Yes please use semantic tokens in place of static values if available.

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -3,130 +3,184 @@
...
 
 .header {
-  @apply bg-foreground-1 flex flex-col
-    z-header;
-  border-block-end: var(--calcite-panel-header-border-block-end, 1px solid var(--calcite-color-border-3));
+  display: flex;
+  flex-direction: column;
+  z-index: var(--calcite-panel-header-z-index, var(--calcite-z-index-header));
+  background-color: var(--calcite-panel-header-background-color, var(--calcite-color-foreground-1));
+  border-block-end: var(
```
alisonailea, line 78: I don't think there are any requests from customers to have control of the border width. I'd prefer we split this rule up into `border-block-end-style`, `border-block-end-width`, and `border-block-end-color` then only add a component token for `--calcite-panel-border-color`. Then use the same pattern for other border rules within the component.

alisonailea (changes requested):

alisonailea:
Looks good! Just a few little things.

packages/calcite-components/src/components/panel/panel.scss
driskull: @alisonailea we're not doing component specific font sizing right?

packages/calcite-components/src/components/panel/panel.scss
driskull: This var was already there so cant change it.

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -3,130 +3,184 @@
...
  *
+ * @prop --calcite-panel-background-color: Specifies the background color of the component.
+ * @prop --calcite-panel-border-color: Specifies the border color of the component.
+ * @prop --calcite-panel-description-color: Specifies the color of the component's description.
+ * @prop --calcite-panel-fab-z-index: Specifies the component fab's z-index.
+ * @prop --calcite-panel-footer-background-color: Specifies the background color of the component's footer.
  * @prop --calcite-panel-footer-padding: Specifies the padding of the component's footer.
+ * @prop --calcite-panel-header-background-color: Specifies the component header's background color.
  * @prop --calcite-panel-header-border-block-end: Specifies the component header's block end border.
+ * @prop --calcite-panel-header-color: Specifies the component header's  color.
```
macandcheese, line 14: Should this be `heading` to correspond to the property name and match `description-`?

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -3,130 +3,184 @@
...
+ * @prop --calcite-panel-header-z-index: Specifies the component header's z-index.
  */
```
macandcheese, line 16: Can we add a `corner-radius` that defaults to "sharp" and a `box-shadow` that defaults to "none" to facilitate theming here? It would facilitate theming these types of layout that use Panel:

Screenshot 2024-01-30 at 10 58 53 AM

packages/calcite-components/src/components/panel/panel.scss
driskull: I'm torn on this one. It does set the color for the header but I could see your point. What if there is more text in the header in the future? Would it affect slotted actions?

macandcheese: I suppose both could be valid as well - the `header` here could apply to any slotted `header-content`, and a `heading` one could apply just to the rendered `heading` prop?

driskull: Can we just add the heading one later if we need to? @macandcheese

packages/calcite-components/src/components/panel/panel.scss
driskull: @macandcheese wouldn't this be out of scope? We're not currently setting any border radius within panel.

macandcheese: Yeah, I can open an issue to enable this as a follow up as this should be theme able IMO. We did add these theme able defaults elsewhere in PR for this effort - button, card, meter, some others I think.

packages/calcite-components/src/components/panel/panel.scss
macandcheese: Yeah that’s fine. I guess between the two I’d expect ‘heading’ since we render that text for a user based on prop and slotted ‘header’ content is already user-controllable.

packages/calcite-components/src/components/panel/panel.scss
driskull: Yeah i think if we're adding new CSS we should probably have an issue.

It should be a separate feat IMO. cc @jcfranco

packages/calcite-components/src/components/panel/panel.scss
macandcheese: I don’t think we’d want it to affect slotted actions though, since those are individually theme able already. Would it just be the rendered overflow “ellipsis” action affected by this? Besides slotted content?

driskull: > Would it just be the rendered overflow “ellipsis” action affected by this? Besides slotted content?

I think so.

packages/calcite-components/src/components/panel/panel.scss
macandcheese: https://github.com/Esri/calcite-design-system/issues/8675

packages/calcite-components/src/components/panel/panel.scss
macandcheese: I guess my feeling is `heading` is more useful here for the "I want to change the text color of heading and description" theming need. But, ultimately we can add more later if needed.

re: the border ones:
Can / should the other one be simplified to `--calcite-panel-header-border-color` ? The block-end seems a little in the weeds for the user. This also makes me think there would be a "footer border" prop too.

Maybe we don't even need to allow that to be overridden - trying to think if we'd ever want that to be themed separately from the panel border itself (maybe set to transparent for some reason?)

macandcheese: We could deprecate it and leave it until the next breaking change release if we want to I'd think. But, yeah seems a little strangely named.

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -3,130 +3,184 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
+ * @prop --calcite-panel-background-color: Specifies the background color of the component.
+ * @prop --calcite-panel-border-color: Specifies the border color of the component.
+ * @prop --calcite-panel-description-color: Specifies the color of the component's description.
+ * @prop --calcite-panel-fab-z-index: Specifies the component fab's z-index.
+ * @prop --calcite-panel-footer-background-color: Specifies the background color of the component's footer.
  * @prop --calcite-panel-footer-padding: Specifies the padding of the component's footer.
```
alisonailea, line 11: are there requests/user stories for custom control of padding? I'm wondering if this token is necessary.

packages/calcite-components/src/components/panel/panel.scss
alisonailea: I agree it's strangely named. I'd vote to deprecate if we can.

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -3,130 +3,184 @@
...
+ * @prop --calcite-panel-header-background-color: Specifies the component header's background color.
  * @prop --calcite-panel-header-border-block-end: Specifies the component header's block end border.
+ * @prop --calcite-panel-header-color: Specifies the component header's  color.
+ * @prop --calcite-panel-header-z-index: Specifies the component header's z-index.
```
alisonailea, line 15: Could we make a `--calcite-panel-z-index` instead? I'm curious what the use case would be for a user to need to change the layer position of only the panel header.

packages/calcite-components/src/components/panel/panel.scss
alisonailea: If panel doesn't already have a box-shadow or border-radius and there haven't been any requests from users to add them I would leave them out for now.

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -3,130 +3,184 @@
...
+.header .heading {
+  flex: 1 1 auto;
+  padding: var(--calcite-spacing-xxs);
+}
+
+.container {
+  background-color: var(--calcite-panel-background-color);
+  margin: 0;
+  display: flex;
+  inline-size: 100%;
```
alisonailea, line 66: inline-size: var(--calcite-container-size-content-fluid);

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -3,130 +3,184 @@
...
+  border-block-end: var(--calcite-panel-header-border-block-end);
 }
 
 .header-container {
-  @apply flex flex-row w-full
-  items-stretch
-  justify-start;
+  display: flex;
+  flex-direction: row;
+  inline-size: 100%;
```
alisonailea, line 89: inline-size: var(--calcite-container-size-content-fluid);

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -3,130 +3,184 @@
...
 }
 
 .header-container--border-end {
-  border-block-end: 1px solid var(--calcite-color-border-3);
+  border-block-end: var(--calcite-border-width-sm) solid var(--calcite-panel-border-color);
 }
 
 .action-bar-container {
-  @apply w-full;
+  inline-size: 100%;
```
alisonailea, line 100: inline-size: var(--calcite-container-size-content-fluid);

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -3,130 +3,184 @@
...
+    overflow-wrap: break-word;
+    padding: 0;
   }
   .heading {
-    @apply text-0h mx-0 mt-0 mb-1 font-medium;
+    margin-inline: 0px;
+    margin-block: 0px var(--calcite-spacing-xxs);
+    font-weight: var(--calcite-font-weight-medium);
+    font-size: var(--calcite-font-size-0);
+    line-height: 1.25rem;
```
alisonailea, line 125: var(--calcite-font-line-height-relative-tight)

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -3,130 +3,184 @@
...
     &:only-child {
-      @apply mb-0;
+      margin-block-end: 0;
     }
   }
   .description {
-    @apply text-color-2 text-n1h;
+    color: var(--calcite-panel-description-color);
+    font-size: var(--calcite-font-size--1);
+    line-height: 1rem;
```
alisonailea, line 133: --calcite-font-line-height-relative

alisonailea (changes requested):

packages/calcite-components/src/components/panel/panel.scss
macandcheese: I have heard requests / designs that call for this, I can share in Teams if needed as they are from unreleased product design. Happy to just do it in the issue I added above as well.

macandcheese: I have heard requests / designs that call for this, I can share in Teams if needed as they are from unreleased product design. Happy to just do it in the issue I added above as well which more or less shows a generic version of the request.

packages/calcite-components/src/components/panel/panel.scss
driskull: This var is already there so if we want we can deprecate it.

packages/calcite-components/src/components/panel/panel.scss
driskull: I prefer the specificity here IMO. If we say `--calcite-panel-z-index` users may think its the whole panel's z-index.

They also have different values as is. One uses the header zindex and the other uses the sticky zindex (FAB).

packages/calcite-components/src/components/panel/panel.scss
alisonailea: - [ ] deprecate old calcite-panel-footer-padding variable and replace with a "space" token

packages/calcite-components/src/components/panel/panel.scss
alisonailea: - [ ] deprecate --calcite-panel-header-color

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -3,130 +3,181 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
+ * @prop --calcite-panel-background-color: Specifies the background color of the component.
+ * @prop --calcite-panel-border-color: Specifies the border color of the component.
+ * @prop --calcite-panel-description-color: Specifies the color of the component's description.
```
alisonailea, line 8: - [ ] Change to `--calcite-panel-description-text-color`

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -3,130 +3,181 @@
...
+ * @prop --calcite-panel-fab-z-index: Specifies the component fab's z-index.
+ * @prop --calcite-panel-footer-background-color: Specifies the background color of the component's footer.
  * @prop --calcite-panel-footer-padding: Specifies the padding of the component's footer.
+ * @prop --calcite-panel-header-background-color: Specifies the component header's background color.
  * @prop --calcite-panel-header-border-block-end: Specifies the component header's block end border.
```
alisonailea, line 13: - [ ] deprecate and replace with `--calcite-panel-header-border-width`.

We don't need to specify it only applies to block-end as that's obvious from looking at the component.

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -3,130 +3,181 @@
...
-  @apply relative flex w-full h-full flex-auto;
+  position: relative;
+  display: flex;
+  inline-size: var(--calcite-container-size-content-fluid);
+  block-size: var(--calcite-container-size-content-fluid);
+  flex: 1 1 auto;
+  --calcite-internal-panel-transition: max-block-size var(--calcite-animation-timing),
+    inline-size var(--calcite-animation-timing);
 
-  --calcite-min-header-height: calc(var(--calcite-icon-size) * 3);
```
alisonailea, line 14: does this need to be deprecated?

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -3,130 +3,181 @@
...
 
 .back-button {
-  @apply border-color-3
-  border-0
-  border-solid;
-  border-inline-end-width: theme("borderWidth.DEFAULT");
+  border-color: var(--calcite-panel-border-color);
+  border-style: solid;
+  border-width: 0px;
+  border-inline-end-width: var(--calcite-border-width-sm);
```
alisonailea, line 138: this could also use the `var(--calcite-panel-border-width)`

alisonailea (changes requested):

packages/calcite-components/src/components/panel/panel.scss
driskull: Doesn't seem like it was being used.

macandcheese (approved):
Looks good!

I can do a quick follow up for https://github.com/Esri/calcite-design-system/issues/8675 after this lands - we showed a lot of demo with those customizations this morning 😄

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -3,130 +3,187 @@
...
+ * @prop --calcite-panel-background-color: Specifies the background color of the component.
+ * @prop --calcite-panel-border-color: Specifies the border color of the component.
+ * @prop --calcite-panel-description-text-color: Specifies the text color of the component's description.
+ * @prop --calcite-panel-fab-z-index: Specifies the component fab's z-index.
+ * @prop --calcite-panel-footer-background-color: Specifies the background color of the component's footer.
+ * @prop --calcite-panel-footer-padding: [Deprecated] Use `--calcite-panel-footer-space` instead. Specifies the padding of the component's footer.
+ * @prop --calcite-panel-footer-space: Specifies the spacing of the component's footer.
+ * @prop --calcite-panel-header-background-color: Specifies the component header's background color.
+ * @prop --calcite-panel-header-border-block-end: [Deprecated] Use `--calcite-panel-header-border-width` instead. Specifies the component header's block end border.
+ * @prop --calcite-panel-header-border-width: Specifies the component header's border width.
```
macandcheese, line 15: Should this one be deprecated?

driskull: It is new and replaces the old one> https://github.com/Esri/calcite-design-system/pull/8670#discussion_r1483734423

macandcheese: Oh, gotcha. Is there a use case for making this thicker border or removing? Otherwise not sure we need this.

driskull: The use case was for removing the border

macandcheese: Cool, I guess my vote would be to not add this? Unless you know the exact use case for it, seems strange we'd allow this to be removed. Maybe from a case where the header-content slot is filled?

driskull: Agreed, we can just not add it.

Issue #8396, opened 2023-12-11
Combobox - Make icon/label spacing consistent between dropdown items and selected item


rebollin:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

The spacing between the calcite icon used in the combobox component is different than the spacing seen in the dropdown options.
![image](https://github.com/Esri/calcite-design-system/assets/121894974/fa1415d3-7010-44d9-a557-25830fd7f5e1)

### Acceptance Criteria

The spacing between the label and icon should be consistent between the selected option and dropdown options

### Which Component

Combobox input

### Example Use Case

We just switched to using the combobox in the Update Data modal. We noticed that the spacing is off and not adjustable.

### Priority impact

p2 - want for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Online

ashetland:
@jcfranco @eriklharper I think this snuck in with the updates from #4326

macandcheese:
![Screenshot 2023-12-11 at 10 26 16 AM](https://github.com/Esri/calcite-design-system/assets/4733155/214b7dd1-7068-4c21-b409-6174c352cd39)

There also seems to be some uneven spacing in Input, may be worth looking at all of these to make sure this is cleared up across the board.

jcfranco:
@macandcheese Great catch. Let's create a separate issue for this since `combobox` doesn't use `calcite-input` internally.

jcfranco:
@macandcheese It looks like spacing might look different depending on [the icon used](https://codepen.io/jcfranco/pen/gOEoJWg).

![Screenshot 2024-01-29 at 4 09 20 PM](https://github.com/Esri/calcite-design-system/assets/197440/5c295695-b390-4b57-af5b-f30412e56858)

eriklharper:
> @macandcheese Great catch. Let's create a separate issue for this since `combobox` doesn't use `calcite-input` internally.

Perhaps it should?

jcfranco:
☝️
![](https://media1.tenor.com/m/2M8-iXvYBJoAAAAC/maybe-robert-downey-jr.gif)

geospatialem:
[...]

PR #8672, opened 2024-01-30
fix(combobox): avoid inline-start padding on combobox label when icon is displayed

jcfranco:
**Related Issue:** #8396 

## Summary

This fixes a spacing regression introduced in https://github.com/Esri/calcite-design-system/pull/7912/.

### DIFF
--- a/packages/calcite-components/src/components/combobox/combobox.scss
+++ b/packages/calcite-components/src/components/combobox/combobox.scss
@@ -89,7 +89,7 @@
     truncate
     p-0;
 
-  gap: var(--calcite-combobox-item-spacing-unit-l);
+  gap: var(--calcite-combobox-item-spacing-unit-s);
 
   &.selection-display-fit,
   &.selection-display-single {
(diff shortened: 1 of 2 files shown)


eriklharper:
> This fixes a spacing regression introduced in #7912.

This wasn't introduced by #7912, that line was there from before based on this diff: (notice line 174 doesn't have any changes):

image

Could we remove that from the description?

jcfranco:
@eriklharper The regression was introduced in `v2.0.0-next.1` and the referenced issue is the only one modifying `combobox` (see [example](https://codepen.io/jcfranco/pen/XWGZrdy?editors=1000) and [changes between `v2.0.0-next.0` and `v2.0.0-next.1`](https://github.com/Esri/calcite-design-system/compare/@esri/calcite-components@2.0.0-next.0...@esri/calcite-components@2.0.0-next.1)). I think it's caused by the added [`gap` in `.grid-input`](https://github.com/Esri/calcite-design-system/commit/58317ae6803a3a1716cabeaeac2061fa533a6bad#diff-6801b02decf9e399e6f2886aedfb17f0090ade2ece29c37b15911ffca21df1d0R84).

driskull (approved):
👍

PR #8676, opened 2024-01-30
refactor(button): optimize for component tokens

alisonailea:
**Related Issue:** #7180 

## Summary

Reduce file size from 842 lines to 648
Saves around 40ms on the test page (more once loader is refactored to reduce CLS)

- Lift component tokens up to `:host`
- Utilize `:is`, `:has`, and `:not` to limit necessary selector rules
- Only apply a style prop rule once then rely on resetting the value of CSS variables base on classes and state.

### DIFF
--- a/packages/calcite-components/src/components/button/button.scss
+++ b/packages/calcite-components/src/components/button/button.scss
@@ -4,29 +4,11 @@
  * These properties can be overridden using the component's tag as selector.
  *
  * @prop --calcite-button-background-color: Specifies the background color of the component.
- * @prop --calcite-button-background-color-hover: Specifies the background color of the component when in hover state.
- * @prop --calcite-button-background-color-focus: Specifies the background color of the component when in focus state.
- * @prop --calcite-button-background-color-active: Specifies the background color of the component when in active state.
  * @prop --calcite-button-text-color: Specifies the text color of the component.
- * @prop --calcite-button-text-color-hover: Specifies the text color of the component when in hover state.
- * @prop --calcite-button-text-color-focus: Specifies the text color of the component when in focus state.
- * @prop --calcite-button-text-color-active: Specifies the text color of the component when in active state.
  * @prop --calcite-button-icon-start-color: Specifies the color of the component's start position icon.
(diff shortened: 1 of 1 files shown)


packages/calcite-components/src/components/button/button.scss
```diff
@@ -4,122 +4,51 @@
...
- * @prop --calcite-button-border-color-active: Specifies border colors of the component when in active state.
  */
 
 :host {
   --calcite-button-background-color: var(--calcite-color-background);
   --calcite-button-text-color: var(--calcite-color-text-inverse);
   --calcite-button-box-shadow: var(--calcite-shadow-none);
   --calcite-button-border-color: var(--calcite-color-transparent);
-  --calcite-button-corner-radius: var(--calcite-corner-radius-sharp);
+  --calcite-button-corner-radius: var(--calcite-corner-radius);
```
macandcheese, line 19: Shouldn't we leave this as the `sharp` token? If an app set their `sharp` to be... 4px, it should affect Buttons, Cards, etc.,

packages/calcite-components/src/components/button/button.scss
```diff
@@ -4,122 +4,51 @@
...
-  --calcite-button-icon-end-color-focus: var(--calcite-color-text-inverse-focus);
-
-  /* active */
-  --calcite-button-text-color-active: var(--calcite-color-text-inverse-press);
-  --calcite-button-background-color-active: var(--calcite-color-background);
-  --calcite-button-border-color-active: var(--calcite-color-transparent);
-  --calcite-button-box-shadow-active: var(--calcite-shadow-none);
-  --calcite-button-icon-start-color-active: var(--calcite-color-text-inverse-active);
-  --calcite-button-icon-end-color-active: var(--calcite-color-text-inverse-active);
+  --calcite-button-box-shadow: inset 0 0 0 1px var(--calcite-button-box-shadow-color);
```
macandcheese, line 22: Shouldn't we allow this to be a full box shadow rule, defaulting to our `none` - so a user can override this with, say box-shadow-sm, or box-shadow-md (or whatever they need). Do we need to change our usage of box-shadow within this component to use border? How would I set the border color and the box-shadow in this case?

Issue #6690, opened 2023-03-28
[Tile] Responsive Layout


brittneytewks:
### Description

Responsiveness across Calcite Design System, focused on web (traditionally desktop sized web browser), tablet, phone screen sizes, and large monitors.

### Acceptance Criteria

1. Consider breaking changes for folks. Consultation should occur on both the design and development teams.
2. Components need to be designed for the following breakpoints:

- [ ] Small (s), 476px
- [ ]  Medium (m), 768px
- [ ]  Large (l), 1152px
- [ ]  Extra larger (xl), 1440px

### Relevant Info

Part of epic https://github.com/Esri/calcite-components/issues/6670

### Which Component

`tile`

### Priority impact

p4 - not time sensitive

### Esri team

Calcite (design)

SkyeSeitz:
I will be rolling responsive layout considerations into the [Tile redesign epic ](https://github.com/Esri/calcite-components/issues/6662) in June.

SkyeSeitz:
Link to Figma Specs: https://www.figma.com/file/8CiAVHkXIMTsI9hTSbotEz/Tile-Redesign-%5BEpic-6662%5D?type=design&node-id=720%3A2738&mode=dev

geospatialem:
Closing as effort will instead focus on responsive layout with the effort of the `tile` redesign and new component `tile-group` in https://github.com/Esri/calcite-design-system/issues/6662 and https://github.com/Esri/calcite-design-system/issues/8615 respectively.

geospatialem:
Verified with https://codepen.io/geospatialem/pen/zYbQZYx in `2.5.1` for the `alignment` and min and max widths for each scale. Updated context is #6662.

Issue #6662, opened 2023-03-25
Epic: Tile redesign


geospatialem:
## Background
Enhance and consolidate functionality of `tile-select` into `tile`.

Once `tile` efforts are completed we'll:
- release the new `tile-group` component and
- deprecate the `tile-select` and `tile-select-group` components

### Phase I
- [x] 1. Keep the following props:
    - [x] `iconFlipRtl` - boolean (`false`)
    - [x] `description` - string (`undefined`)
    - [x] `disabled` - boolean (`false`)
    - [x] `heading` - string (`undefined`)
    - [x] `hidden` - boolean (`false`)
    - [x] `href` - string (`undefined`)
    - [x] `icon` - `calcite-ui-icon` (`undefined`)
- [x] 2. Add new props: 
    - [x] `alignment` - boolean (`"start"` - default, `"center"`)
    - [x] `interactive` - boolean (`false`)
    - [x] `label` - string (`undefined`)
    - [x] `layout` - type (`"horizontal"` - default, `"vertical"`)
    - [x] `selection-appearance` - type  (`"icon"` - default, `"border"`)
    - [x] `selection-mode` - type (`"none"` - default, `"single"`, `"single-persist"`, `"multiple"`)
    - [x] `scale` - type (`"s"`, `"m"` - default, `"l"`)
    - [x] `selected`  - boolean (`false`) 
- [x] 4. Deprecate props:
    - [x] `embed` - boolean https://github.com/Esri/calcite-design-system/pull/8833
    - [x] `active` - boolean (no longer needed once selection modes are introduced)
- [x] 5. Add `min` and `max widths` to each `scale`
- [x] 6. Update `href` styling (Refer to the [Figma file](https://www.figma.com/file/8CiAVHkXIMTsI9hTSbotEz/Tile-Redesign-%5BEpic-6662%5D?type=design&node-id=720-2738&mode=design&t=ZAQtpcQRe2IVJM16-0) for a recording.
    - [x] No underline
    - [x] Blue border
    - [x] Blue icon
    - [x] Header on hover
   
![image](https://github.com/Esri/calcite-design-system/assets/5023024/b436a018-4ad3-470f-ac23-5ca64aec3a61)

- [x] 7. Replace `calcite-checkbox` with an icon https://github.com/Esri/calcite-design-system/issues/8834
    - [x] Consult with team members if the component could be a form element
    - [x] Icons: `"circle"` and `"circle-f"`
    
[...]

PR #8681, opened 2024-02-01
feat(tile): add responsive layout features

eriklharper:
**Related Issues:** #6662 #6690

## Summary

This PR adds responsive layout features and some design improvements to the Tile component, including:

- Tile: `min` and `max` inline-size (width) for all 3 scales
- Tile: `alignment` property, supporting `start` and `center`

Other improvements include refactoring Tile to remove the deprecated `ConditionalSlotComponent` interface and improving style encapsulation to allow more defined control of theming with css custom properties.


packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -1,39 +1,59 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-tile-background-color: Specifies the background color of the component.
+ * @prop --calcite-tile-border-color: Specifies the border color of the component.
+ * @prop --calcite-tile-border-color-active: Specifies the border color of the component when in an active state.
```
macandcheese, line 8: I suppose this will be resolved with upcoming direction re: token construction, but this could potentially be set like `calcite-tile:hover {  --calcite-tile-border-color: pink }`. I would think component-specific props for pseudo classes might be limited to internal components the external developer doesn't have ability to set? Applies to other props.

packages/calcite-components/src/components/tile/tile.tsx
```diff
@@ -36,6 +31,11 @@ export class Tile implements ConditionalSlotComponent, InteractiveComponent {
    */
   @Prop({ reflect: true }) active = false;
 
+  /**
+   * Specifies the alignment of the Tile's content.
+   */
+  @Prop({ reflect: true }) alignment: Exclude<Alignment, "end"> = "start";
```
macandcheese, line 37: Elsewhere we use portions of interface definitions, we use `Extract<>` vs. `Exclude<>` - no preference here except that we are consistent - does one have benefits over the other?

packages/calcite-components/src/components/tile-group/tile-group.scss
```diff
@@ -0,0 +1,36 @@
...
+
+  .tile-group {
+    display: grid;
+    grid-auto-rows: minmax(auto, 1fr);
+    grid-template-columns: repeat(auto-fit, minmax(140px, 1fr));
+  }
+
+  ::slotted(calcite-tile) {
+    margin-block-end: 1px;
+    margin-inline-end: 1px;
```
alisonailea, line 13: margin-block-end: var(--calcite-spacing-px);
margin-inline-end: var(--calcite-spacing-px);

alisonailea, line 13: margin-block-end: var(--calcite-spacing-px);
margin-inline-end: var(--calcite-spacing-px);

packages/calcite-components/src/components/tile-group/tile-group.scss
```diff
@@ -0,0 +1,36 @@
+:host {
+  box-sizing: border-box;
+  display: inline-block;
+
+  .tile-group {
+    display: grid;
+    grid-auto-rows: minmax(auto, 1fr);
+    grid-template-columns: repeat(auto-fit, minmax(140px, 1fr));
```
alisonailea, line 8: nitpick: move this rule into `:host([scale="m"]) { .tile-group {` for clarity

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -1,39 +1,59 @@
...
-  }
+  --calcite-tile-background-color: var(--calcite-color-foreground-1);
+  --calcite-tile-border-color: var(--calcite-color-border-2);
+  --calcite-tile-border-color-active: var(--calcite-color-brand);
+  --calcite-tile-border-color-hover: var(--calcite-color-brand);
+  --calcite-tile-description-text-color: var(--calcite-color-text-3);
+  --calcite-tile-description-text-color-hover: var(--calcite-color-text-2);
+  --calcite-tile-heading-text-color: var(--calcite-color-text-2);
+  --calcite-tile-heading-text-color-hover: var(--calcite-color-text-1);
+  --calcite-tile-icon-color: var(--calcite-color-text-3);
```
alisonailea, line 26: Remove state specific tokens.

Instead do 
```css
:host(:hover) {
  --calcite-tile-border-color: var(--calcite-color-brand);
  --calcite-tile-description-text-color: var(--calcite-color-text-2);
  --calcite-tile-heading-text-color: var(--calcite-color-text-1);
}
```

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -1,39 +1,59 @@
...
+
+  box-sizing: border-box;
+  display: inline-block;
+  max-inline-size: 460px;
```
alisonailea, line 30: where are these pixel sizes coming from?

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -1,39 +1,59 @@
...
 
-  .content {
-    @apply flex
-      flex-auto
-      flex-col
-      justify-center;
-    inline-size: 10%;
+  .tile {
+    background-color: var(--calcite-tile-background-color);
+    block-size: 100%;
```
alisonailea, line 35: block-size:  var(--calcite-container-size-content-fluid);

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -1,39 +1,59 @@
...
+    box-sizing: border-box;
+    display: flex;
+    flex-direction: column;
+    gap: var(--calcite-spacing-md);
+    inline-size: 100%;
```
alisonailea, line 40: inline-size: var(--calcite-container-size-content-fluid);

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -45,39 +65,46 @@
   }
 
   .heading {
-    @apply pointer-events-none
-      break-words;
-    color: var(--calcite-color-text-2);
+    color: var(--calcite-tile-heading-text-color);
     font-size: var(--calcite-font-size--1);
     font-weight: 500;
```
alisonailea, line 70: font-weight: var(--calcite-font-weight-medium);

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -45,39 +65,46 @@
...
     line-height: 1.20313rem;
```
alisonailea, line 71: where is this line height coming from?

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -45,39 +65,46 @@
...
+      justify-content: center;
     }
   }
   .description {
-    @apply break-words
-      pointer-events-none;
-    color: var(--calcite-color-text-3);
+    color: var(--calcite-tile-description-text-color);
     font-size: var(--calcite-font-size--2);
     font-weight: 400;
```
alisonailea, line 91: font-weight: var(--calcite-font-weight-regular);

alisonailea (changes requested):

packages/calcite-components/src/components/tile/tile.tsx
eriklharper: Good question, not sure.

packages/calcite-components/src/components/tile-group/tile-group.scss
eriklharper: @jcfranco  what do you think about Ali's idea here?  I did this mainly because in the past if there was a default case in some javascript logic, setting things up like this minimized code branches, but maybe we want to do things differently with CSS?

packages/calcite-components/src/components/tile/tile.scss
eriklharper: From the [Figma specs](https://www.figma.com/file/8CiAVHkXIMTsI9hTSbotEz/Tile-Redesign-%5BEpic-6662%5D?type=design&node-id=1107-2810&mode=dev):

image

packages/calcite-components/src/components/tile/tile.scss
eriklharper: Also from the Figma file.  These are copied verbatim from the styles panel of the corresponding text elements:

image

packages/calcite-components/src/components/tile-group/tile-group.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  disconnectedCallback(): void {
+    disconnectInteractive(this);
+  }
+
+  render(): VNode {
+    return (
+      <InteractiveContainer disabled={this.disabled}>
+        <div class={CSS.tileGroup}>
+          <slot />
```
driskull, line 89: there should be an onslotchange handler here to update tiles that are added after connectedCallback.

eriklharper: Thanks for catching that  👍

eriklharper:
I'm going to split this PR in two and make a separate one just for adding Tile Group on its own.

packages/calcite-components/src/components/tile/tile.scss
alisonailea: @SkyeSeitz we haven't set Semantic Tokens in Tile Component yet right? I think this needs to be replaced with a semantic line-height token. Pick the one closest to this value.

packages/calcite-components/src/components/tile/tile.scss
alisonailea: @SkyeSeitz should these be container tokens or based off container tokens?

packages/calcite-components/src/components/tile-group/tile-group.scss
jcfranco: I agree and think we should consider updating our approach to lean more on being explicit vs having implicit defaults styles.

jcfranco (approved):
I have some minor comments, but once @alisonailea's token comments have been addressed, this should be good to merge!

jcfranco:
@eriklharper Also, can you update your PR title to be a sentence vs noun phrase? This will better align with the Ionic [commit message format guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md#subject) we follow (per [our contributing guidelines](https://github.com/Esri/calcite-design-system/blob/main/CONTRIBUTING.md#commit-message-format)).

eriklharper:
Changed the branch name, closing this in favor of: https://github.com/Esri/calcite-design-system/pull/8691

packages/calcite-components/src/components/tile/tile.scss
eriklharper: Ok, cool, I've addressed this in the new #8691 pr.

packages/calcite-components/src/components/tile/tile.scss
eriklharper: This has been addressed in #8691.  Thanks for the feedback!

packages/calcite-components/src/components/tile/tile.scss
SkyeSeitz: @alisonailea great question. The ones highlighted in blue below are the only values with something close in our container tokens. Do you think we should add the rest as container tokens? 

image

Issue #8659, opened 2024-01-26
[Table] Add a non-keyboard-navigable mode to Table


macandcheese:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Currently, the Table component supports navigation and focus of Table Cell with arrow keys. This is useful, but not all the time - and will be more useful when we add built-in editable table cells and other datagrid functionality.

### Acceptance Criteria

Adds a new `interactionMode` property (or other name) that defaults to the current behavior, but supports a "none" mode that allows the Table to render without focusable Table Cell or keyboard navigation other than default browser behavior between areas like selection, pagination, etc.

### Relevant Info

Focusable content within Table Cell should not be impacted - these will still be keyboard navigable as expected. cc @geospatialem 

### Which Component

Table, Table child components

### Example Use Case

This will allow for more "static table" use cases where the datagrid-navigation  keyboard method could be unexpected.

### Priority impact

p2 - want for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

geospatialem:
The component now bypasses keyboard navigation, but during the PR review when JAWS and NVDA were active both tables and content were still accessible via keyboard. 

During verification observed both JAWS and NVDA can no longer access the content via [this codepen](https://codepen.io/geospatialem/pen/oNVQQOa) in `2.5.0-next.2`. cc @macandcheese

PR #8686, opened 2024-02-01
feat(table): Add `interactionMode` property to control focus behavior

macandcheese:
**Related Issue:** #8659 

## Summary
- Adds an `interactionMode` property with `static` and `interactive` (default) values to Table to allow the table to be used without cell + header focus.
- When set, prevents keyboard navigation with arrow / home / page keys.
- Still allows focus and tab / shift tab for `interactionMode` selection affordances in cell + header.
- Still allows tab to / shift tab to reach focusable content
- Prevent focus of "unused" `interactionMode` footer cell in `static` mode.
- Adds test to check that only interactionMode cells + header are focused in `static` mode.
- Does not change the default behavior.


packages/calcite-components/src/components/table-cell/table-cell.tsx
```diff
@@ -61,6 +61,9 @@ export class TableCell
   /** @internal */
   @Prop() lastCell: boolean;
 
+  /** @internal */
+  @Prop() nonInteractive = false;
```
driskull, line 65: How about something like `interaction-disabled`? This would align better with how we disable features throughout components. Our props should be in the tense as `disabled`.

macandcheese: Yeah, that's a better boolean name.

Any preference between `interactionDisabled` and `interactionMode="non-interactive" | "interactive"` (or another property / value option)?

driskull: we have `selectionMode` on a few other components but i'm not sure its the same thing.

macandcheese: Yeah, different functionality, Table has `selectionMode` as well. Could see the similarity as good or bad. Either way, `interactionDisabled` is better than `nonInteractive`. cc @jcfranco for any thoughts.

packages/calcite-components/src/components/table/table.tsx
```diff
@@ -68,6 +68,9 @@ export class Table implements LocalizedComponent, LoadableComponent, T9nComponen
   /** Specifies the layout of the component. */
   @Prop({ reflect: true }) layout: TableLayout = "auto";
 
+  /** When `true`, prevents focus and keyboard navigation of Table Header and Table Cell components. Selection affordances and slotted content within Table Cell components remains focusable. */
```
geospatialem, line 71: The focus and keyboard navigation will still be present with AT active. WDYT about adding in some context for folks? So they get additional context that could help them decide if they'd like to disable it when AT is not active, without taking away support for AT users?

```suggestion
  /** When `true`, prevents focus and keyboard navigation of the `table-header` and `table-cell`s when assistive technologies are not active. Selection affordances and slotted content within `table-cell`s remain focusable. */
```

geospatialem (commented):
Awesome! ✨ 

This works well with Windows without AT active - both with and without selection, and when JAWS is active - also with and without selection. 🥇

packages/calcite-components/src/components/table-cell/table-cell.tsx
alisonailea: what about `interactionMode="static" | "interactive"`

macandcheese: Yeah, that makes sense to me as well.

packages/calcite-components/src/components/table/table.tsx
macandcheese: Added

packages/calcite-components/src/components/table-cell/table-cell.tsx
macandcheese: I've updated the PR to use interactionMode="static" | "interactive"

packages/calcite-components/src/components/table/table.tsx
macandcheese: @geospatialem I updated this to reflect the new property name and two values. Could you give a look over?

packages/calcite-components/src/components/table/table.tsx
```diff
@@ -65,6 +65,9 @@ export class Table implements LocalizedComponent, LoadableComponent, T9nComponen
   /** When `true`, number values are displayed with a group separator corresponding to the language and country format. */
   @Prop({ reflect: true }) groupSeparator = false;
 
+  /** When `interactive`, allows focus and keyboard navigation of `table-header`s and `table-cell`s.  When `static`, prevents focus and keyboard navigation of `table-header`s and `table-cell`s when assistive technologies are not active. Selection affordances and slotted content within `table-cell`s remain focusable. */
```
geospatialem, line 68: Awesome ✨ , two small nitpicks with the values:

```suggestion
  /** When `"interactive"`, allows focus and keyboard navigation of `table-header`s and `table-cell`s.  When `"static"`, prevents focus and keyboard navigation of `table-header`s and `table-cell`s when assistive technologies are not active. Selection affordances and slotted content within `table-cell`s remain focusable. */
```

packages/calcite-components/src/components/table-cell/table-cell.tsx
jcfranco: I like @alisonailea's suggestion. 👍

macandcheese: Cool, I think this one should be good to go then. I'll re-submit for approval.

packages/calcite-components/src/components/table-cell/resources.ts
```diff
@@ -5,4 +5,5 @@ export const CSS = {
   selectedCell: "selected-cell",
   assistiveText: "assistive-text",
   lastCell: "last-cell",
+  nonInteractive: "non-interactive",
```
jcfranco, line 8: Nitpick: should this and related styles reflect the updated name?

packages/calcite-components/src/components/table-row/table-row.tsx
```diff
@@ -385,7 +390,11 @@ export class TableRow implements InteractiveComponent, LocalizedComponent {
             aria-rowindex={this.positionAll + 1}
             aria-selected={this.selected}
             class={{ [CSS.lastVisibleRow]: this.lastVisibleRow }}
-            onKeyDown={(event) => this.keyDownHandler(event)}
+            onKeyDown={(event) => {
+              if (this.interactionMode === "interactive") {
```
jcfranco, line 394: Could we push this into the event handler to avoid creating a new anonymous function on each render?

packages/calcite-components/src/components/table/table.e2e.ts
```diff
@@ -2464,6 +2503,95 @@ describe("keyboard navigation", () => {
...
+
+    await page.keyboard.press("Tab");
+    await page.waitForChanges();
+
+    expect(
+      await page.$eval(
+        `#${rowHead.id}`,
+        (el) => el.shadowRoot?.activeElement.shadowRoot?.querySelector("th").classList,
+      ),
+    ).toEqual({ "0": CSS.selectionCell, "1": CSS.multipleSelectionCell });
```
jcfranco, line 2547: If these are asserting on styles, we could look into converting these to screenshot tests later.

jcfranco (approved):
✨🏓✨

packages/calcite-components/src/components/table/table.e2e.ts
macandcheese: They are used to test if focus is correctly applied to internally-rendered elements. Is there a helper or better pattern for this I can use?

jcfranco: Not at the moment, I'm afraid. I think we can revisit this once we have play functions set up in Storybook.

packages/calcite-components/src/components/table-cell/resources.ts
macandcheese: Updated to match other css cell modifiers 👍

packages/calcite-components/src/components/table-row/table-row.tsx
macandcheese: Updated and will merge in post-Chromatic run. Thanks for review!

Issue #8615, opened 2024-01-16
[Tile Group] Create new component


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Create a new component, `tile-group`, which will deprecate the `tile-select-group` component and consumes the new consolidated `tile` component, with responsive design in mind via https://github.com/Esri/calcite-design-system/issues/6662.

### Acceptance Criteria

Refer to https://www.figma.com/file/8CiAVHkXIMTsI9hTSbotEz/Tile-Redesign-%5BEpic-6662%5D?node-id=720%3A2738&mode=dev for the specs on `tile-group`. cc @SkyeSeitz

- [x] Create a new `tile-group` component
- [ ] Deprecate the `tile-select-group` component
- [x] Consume the newly consolidated `tile` component
- [x] Adds the `scale` property

![image](https://github.com/Esri/calcite-design-system/assets/5023024/0b48e3d4-73d6-458c-b365-b0e10863879e)

### Relevant Info

Breaks up the addition of the `scale` attr/prop from https://github.com/Esri/calcite-design-system/issues/4759.

### Which Component

- `tile-group` (new)
- related to `tile-select-group` and `tile`

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

Issue #6691, opened 2023-03-28
[Tile Group] Responsive layout


brittneytewks:
### Description

Responsiveness across Calcite Design System, focused on web (traditionally desktop sized web browser), tablet, phone screen sizes, and large monitors.

#### Task breakdown:
- [ ] https://github.com/Esri/calcite-design-system/issues/8615

1. Consider breaking changes for folks. Consultation should occur on both the design and development teams.
2. Components need to be designed for the following breakpoints:

- [ ] Small (s), 476px
- [ ] Medium (m), 768px
- [ ] Large (l), 1152px
- [ ] Extra larger (xl), 1440px

### Which Component

`tile-group`

### Priority impact

p4 - not time sensitive

### Esri team

Calcite (design)

SkyeSeitz:
I will be rolling responsive layout considerations into the [Tile redesign epic ](https://github.com/Esri/calcite-components/issues/6662) in June.

SkyeSeitz:
Link to Figma specs: https://www.figma.com/file/8CiAVHkXIMTsI9hTSbotEz/Tile-Redesign-%5BEpic-6662%5D?type=design&node-id=720%3A2738&mode=dev

eriklharper:
> I will be rolling responsive layout considerations into the [Tile redesign epic ](https://github.com/Esri/calcite-components/issues/6662) in June.

Since this is part of the Tile redesign epic, does that mean we won't apply responsiveness changes to the planned-to-be-deprecated Tile Select Group/Tile Select components?

SkyeSeitz:
Discussed and decided to roll responsiveness into the Tile & Tile Group redesign to avoid applying new functionality to Tile Select and Tile Select Group which will be deprecated shortly after in favor of Tile and Tile Group.

geospatialem:
Closing as effort will instead focus on responsive layout with the effort of the `tile` redesign and new component `tile-group` in https://github.com/Esri/calcite-design-system/issues/6662 and https://github.com/Esri/calcite-design-system/issues/8615 respectively.

PR #8687, opened 2024-02-02
feat(tile-group): add responsive Tile Group component

eriklharper:
**Related Issues:** #8615 #6691 #6662

## Summary

This PR adds the new `calcite-tile-group` component.  It includes these responsiveness features that address #6691:

- Wrapped Tiles match width of Tiles above
- Wrapped Tiles match height of tallest Tile in the group

The changes in this PR were extracted from and depend on the changes in #8681.


macandcheese (changes requested):
Nice! Some things I saw in local testing:

This hover state needs some adjustment:

Screenshot 2024-02-01 at 4 10 52 PM

I'm also unable to see a focus outline when focused - the Tile is focused as I can navigate with keyboard "Enter" when `href` is present but there isn't a visual indication.

Should there be keyboard support for arrow / home / end within these groups? If so, would it change based on `layout`?

eriklharper:
> Should there be keyboard support for arrow / home / end within these groups? If so, would it change based on `layout`?

This PR is just about adding responsive layout features.  I intend on addressing this in follow-up PRs tied to the #6662 epic.

eriklharper:
> This hover state needs some adjustment:
> 
> Screenshot 2024-02-01 at 4 10 52 PM
> I'm also unable to see a focus outline when focused - the Tile is focused as I can navigate with keyboard "Enter" when `href` is present but there isn't a visual indication.

This is also out of scope for this PR.  These changes are part of Tile, which will come in subsequent PRs after this initial PR to add `tile-group` is merged in.

packages/calcite-components/src/components/tile-group/tile-group.scss
```diff
@@ -0,0 +1,36 @@
...
+
+  .tile-group {
+    display: grid;
+    grid-auto-rows: minmax(auto, 1fr);
+    grid-template-columns: repeat(auto-fit, minmax(140px, 1fr));
+  }
+
+  ::slotted(calcite-tile) {
+    margin-block-end: 1px;
+    margin-inline-end: 1px;
```
alisonailea, line 13: var(--calcite-spacing-px);

packages/calcite-components/src/components/tile-group/tile-group.tsx
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  disconnectedCallback(): void {
+    disconnectInteractive(this);
+  }
+
+  render(): VNode {
+    return (
+      <InteractiveContainer disabled={this.disabled}>
+        <div class={CSS.tileGroup}>
```
alisonailea, line 88: nitpick/future improvement: 
It seems silly to me to add a whole extra `div` to the ShadowDOM just for a classname when it would make more sense to allow the `InteractiveContainer` to take a `class` prop and pass it through.

alisonailea (changes requested):

packages/calcite-components/src/components/tile-group/tile-group.tsx
eriklharper: `InteractiveContainer` uses `display: contents;` so its like its not even there and any styles applied wouldn't take effect anyway.

eriklharper:
> Nice! Some things I saw in local testing:
> 
> This hover state needs some adjustment:
> 
> Screenshot 2024-02-01 at 4 10 52 PM
> I'm also unable to see a focus outline when focused - the Tile is focused as I can navigate with keyboard "Enter" when `href` is present but there isn't a visual indication.
> 
> Should there be keyboard support for arrow / home / end within these groups? If so, would it change based on `layout`?

This has been fixed on this branch.  Thanks for the suggestion!

macandcheese (approved):
Looks good!

(design) - should there be a hover state that is distinct from focus state? Right now they are the same cc @SkyeSeitz @ashetland

alisonailea (approved):

PR #8689, opened 2024-02-02
build(storybook): lint component readmes after building docs

benelan:
## Summary

Lint the component readmes after they are generated by the storybook build to reduce the diffs. The readmes are linted in the action that updates them:

https://github.com/Esri/calcite-design-system/blob/2d111761f756689942775174fdc8b1217aca16c1/.github/workflows/update-doc.yml#L19-L20


driskull (approved):
Nice!

PR #8690, opened 2024-02-03
fix: don't override existing validationMessage when displaying after form submission

benelan:
**Related Issue:** #8000 

## Summary

- Only set `validationMessage` property during form submission if it doesn't have an existing value. This allows users to set a custom validation message, which will be displayed instead of the default message provided by the browser.

- Move the form validation e2e tests to the formAssociated common test, which was [requested as a follow up item](https://github.com/Esri/calcite-design-system/pull/8574#discussion_r1453899164).


packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -777,14 +788,52 @@ export function formAssociated(
...
+        ${componentHtml}
+        <calcite-button type="submit">Submit</calcite-button>
+      </form>
+    `;
+
+    await page.setContent(content);
+    await page.waitForChanges();
+    const component = await page.find(tag);
+
+    const submitButton = await page.find("calcite-button");
```
driskull, line 813: nitpick: should this use an id to find the button? just incase another calcite-button exists? What if the component HTML includes a calcite-button inside it not in the shadowDOM?

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -777,14 +788,52 @@ export function formAssociated(
...
   function ensureName(html: string, componentTag: string): string {
     return html.includes("name=") ? html : html.replace(componentTag, `${componentTag} name="testName" `);
   }
 
+  function ensureRequired(html: string, componentTag: string): string {
+    return html.includes("required") ? html : html.replace(componentTag, `${componentTag} required `);
+  }
+
+  function ensureUnchecked(html: string): string {
+    return html.replace(/(checked|selected)/, "");
```
driskull, line 830: hopefully checked and selected are all we need and never more in the future :)

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -983,6 +1032,80 @@ export function formAssociated(
...
+    page: E2EPage,
+    component: E2EElement,
+    options: FormAssociatedOptions,
+    event: EventSpy,
+    tag: string,
+  ) {
+    await component.callMethod("setFocus");
+    await page.waitForChanges();
+
+    if (tag === "calcite-combobox") {
```
driskull, line 1058: Is there a way to move this component specific logic outside of the common test? Maybe this kind of data could be stored in the component resource file or something and passed in as an option on the test?

driskull: Just the combobox, select and input-time picker having specific logic here seems odd since its a common test.

driskull (approved):
Looking good @benelan Just had some feedback.

packages/calcite-components/src/tests/commonTests.ts
benelan: Yeah, luckily the test would fail if the form submits due to any other attribute existing that satisfies the `required` restraint. I'm going to write some validation doc once I'm done and I'll note that this function may need to be updated.

Issue #7761, opened 2023-09-18
Color Picker's alpha-channel slider scope doesn't update


ashetland:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When using the alpha-channel slider, the scope doesn't update to reflect the current opacity.
![CleanShot 2023-09-18 at 13 10 09](https://github.com/Esri/calcite-design-system/assets/108549080/bb54ec3c-5593-4e87-9e94-f6a8e2561b38)

### Expected Behavior

I would expect the color in the scope to change.
![image](https://github.com/Esri/calcite-design-system/assets/108549080/61a4bbba-7093-4273-822a-20f219a1d8c6)

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/color-picker/

### Reproduction Steps

1. Enable alpha-channel
2. Use alpha-channel slider

### Reproduction Version

v1.8.0

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

Calcite (design)

ashetland:
Hey team, I think I expected the color in the alpha channel scope to fade out to white like pictured above instead of being completely transparent.

geospatialem:
> Hey team, I think I expected the color in the alpha channel scope to fade out to white like pictured above instead of being completely transparent.

@Elijbet - can you confirm on @ashetland's comment above? It looks like the scope (where the pointer is the below screenshot) differs from the expected behavior.

### Expected
![image](https://github.com/Esri/calcite-design-system/assets/5023024/b4db12a9-9dfa-41c7-abf2-126a7fdc6b7f)

### With `2.5.0-next.2`
![image](https://github.com/Esri/calcite-design-system/assets/5023024/f08d5587-090b-458b-a563-15620f7c6411)

jcfranco:
@ashetland @geospatialem IMO, the current behavior is expected since the checkerboard pattern is being used to represent full transparency.

This is common in other color picker components:

https://uiwjs.github.io/react-color/
https://jscolor.com/
https://opensource.adobe.com/spectrum-web-components/components/color-area/ 👇
[...]

PR #8700, opened 2024-02-07
fix(color-picker): alpha-channel slider scope updates to reflect current opacity

Elijbet:
**Related Issue:** #7761

## Summary

When using the alpha-channel slider, the scope now updates to reflect the current opacity.


packages/calcite-components/src/components/color-picker/color-picker.tsx
```diff
@@ -1400,14 +1411,28 @@ export class ColorPicker
...
     context.stroke();
 
+    const pattern = context.createPattern(this.getCheckeredBackgroundPattern(), "repeat");
+
+    context.beginPath();
+    context.arc(x, y, radius - 3, startAngle, endAngle);
+    context.fillStyle = pattern;
+    context.fill();
+
+    context.globalCompositeOperation = "source-atop";
```
jcfranco, line 1426: Noice! I was not aware of this prop. Will need to add it to the toolbox. ✨🧰✨

packages/calcite-components/src/components/color-picker/color-picker.tsx
```diff
@@ -1391,6 +1401,7 @@ export class ColorPicker
     x: number,
     y: number,
     color: Color,
+    sliderType: SliderType,
```
jcfranco, line 1404: `sliderType` only seems to be used to control whether the color's alpha is applied or not. Signature-wise, it might be simpler to:

1. replace the argument with a boolean to apply alpha
2. drop the argument and rely entirely on `color` for this by making sure we update the color's alpha value when passing it to this function

I think 1. seems simpler and avoids extra creating color objects.

jcfranco:
@Elijbet Can you update your PR title to be a sentence vs noun phrase? I think the first line in the description is a good starting point. This will better align with the Ionic [commit message format guidelines](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md#subject) we follow (per [our contributing guidelines](https://github.com/Esri/calcite-design-system/blob/main/CONTRIBUTING.md#commit-message-format)). 

Also, can you adjust the description to summarize the key changes and goals? For complex parts, we can refer to specifics in the code.

jcfranco (changes requested):
@Elijbet Looking good! 🌊🏄

packages/calcite-components/src/components/color-picker/color-picker.tsx
Elijbet: Yeah, it's layering operations, photoshop style. Here is a playground: https://www.w3schools.com/jsref/playcanvas.php?filename=playcanvas_globalcompop&preval=source-atop

packages/calcite-components/src/components/color-picker/color-picker.tsx
```diff
@@ -1400,14 +1401,30 @@ export class ColorPicker
...
     context.fillStyle = "#fff";
     context.fill();
+
     context.strokeStyle = "rgba(0,0,0,0.3)";
     context.lineWidth = outlineWidth;
     context.stroke();
 
+    const pattern = context.createPattern(this.getCheckeredBackgroundPattern(), "repeat");
+
+    if (color.alpha() < 1) {
```
jcfranco, line 1411: I think this also needs to take `applyAlpha` into consideration.

packages/calcite-components/src/components/color-picker/color-picker.tsx
```diff
@@ -1400,14 +1401,30 @@ export class ColorPicker
...
+      context.fillStyle = pattern;
+      context.fill();
+    }
+
+    context.globalCompositeOperation = "source-atop";
+
     context.beginPath();
     context.arc(x, y, radius - 3, startAngle, endAngle);
-    context.fillStyle = color.rgb().alpha(1).string();
+    applyAlpha
```
jcfranco, line 1422: Can you simplify this? The `fillStyle` assignment is mostly duplicated (the alpha value is the only variant here).

packages/calcite-components/src/components/color-picker/color-picker.tsx
```diff
@@ -1400,14 +1401,29 @@ export class ColorPicker
     context.arc(x, y, radius, startAngle, endAngle);
     context.fillStyle = "#fff";
     context.fill();
+
     context.strokeStyle = "rgba(0,0,0,0.3)";
     context.lineWidth = outlineWidth;
     context.stroke();
 
+    const pattern = context.createPattern(this.getCheckeredBackgroundPattern(), "repeat");
```
jcfranco, line 1409: Can you move this into the `if` statement block below? `pattern` is created but only used if alpha will be applied.

packages/calcite-components/src/components/color-picker/color-picker.tsx
```diff
@@ -1400,14 +1401,29 @@ export class ColorPicker
...
+
+    if (color.alpha() < 1 && applyAlpha) {
```
jcfranco, line 1411: Nitpick: this isn't applicable in this instance since `alpha()` isn't expensive, but in some cases it's beneficial to test flags before functions to avoid the method call.

jcfranco (approved):
👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻
👻🎨👻👻👻🎨👻👻🎨🎨👻👻🎨🎨🎨👻👻🎨🎨🎨👻🎨🎨🎨🎨👻🎨👻
👻🎨🎨👻👻🎨👻🎨👻👻🎨👻👻🎨👻👻🎨👻👻👻👻🎨👻👻👻👻🎨👻
👻🎨👻🎨👻🎨👻🎨👻👻🎨👻👻🎨👻👻🎨👻👻👻👻🎨🎨🎨👻👻🎨👻
👻🎨👻👻🎨🎨👻🎨👻👻🎨👻👻🎨👻👻🎨👻👻👻👻🎨👻👻👻👻👻👻
👻🎨👻👻👻🎨👻👻🎨🎨👻👻🎨🎨🎨👻👻🎨🎨🎨👻🎨🎨🎨🎨👻🎨👻
👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻

PR #8710, opened 2024-02-09
feat(checkbox): add component tokens

alisonailea:
**Related Issue:** #7180 

## Summary

Adds Checkbox Component tokens


packages/calcite-components/src/components/checkbox/checkbox.scss
```diff
@@ -3,9 +3,18 @@
...
- * @prop --calcite-checkbox-size: Specifies the component's height and width.
+ * @prop --calcite-checkbox-size: defines the checkbox's size vertically and horizontally.
+ * @prop --calcite-checkbox-background-color: defines the background color of the checkbox.
+ * @prop --calcite-checkbox-box-shadow: defines the box shadow of the checkbox.
+ * @prop --calcite-checkbox-icon-color: defines the icon color of the checkbox.
  */
 
+:host {
+  --calcite-checkbox-background-color: var(--calcite-color-foreground-1);
+  --calcite-checkbox-box-shadow: inset 0 0 0 1px var(--calcite-color-border-input);
```
macandcheese, line 14: Maybe a question for later refactor as the component is already built like this - I think as a user I'd expect this to be a border-color variable with a separate one for box-shadow (defaulting to none)

alisonailea: I have a note to talk to design about looking into adding to the semantic box-shadows to cover use cases like this.

macandcheese (approved):
Looks good - should we update `box-shadow` to `shadow` ?

PR #8713, opened 2024-02-09
feat(code): Add syntax highlighter code component. 

driskull:
**Related Issue:** WIP

## Summary

WIP


driskull:
@macandcheese do you know if we have an issue for this type of component?

macandcheese:
> @macandcheese do you know if we have an issue for this type of component?

No, just the key code one which seems different.

Are we planning on passing through configuration options like lang, line numbering, etc, 

We’ll probably want design pass on this for color decisions, spacing, etc. The developer site has a pretty built out one with a ton of functionality, using highlight I think, should we sync with them?

driskull:
> Are we planning on passing through configuration options like lang, line numbering, etc,

Yeah I would think so, I think that depends on the design requirements as well.

> We’ll probably want design pass on this for color decisions, spacing, etc. The developer site has a pretty built out one with a ton of functionality, using highlight I think, should we sync with them?

Yes, can we get an issue created for the design?  Yes, lets sync with them.

Thanks @macandcheese

Issue #8709, opened 2024-02-08
calcite-loader performance issue in Chrome


adepottey:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When the Loader component (indeterminate) is on a page, the "Google Chrome Helper (Renderer)" process CPU utilization begins climbing and we begin to see rendering responsiveness issues elsewhere on the page. The behavior can be observed on the the doc page for the loader: https://developers.arcgis.com/calcite-design-system/components/loader/
We do not see the same impact on responsiveness or CPU utilization on Firefox.

### Expected Behavior

CPU utilization remains stable and does not negatively impact other elements on the page

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/loader/

### Reproduction Steps

1. Open Activity Monitor or Task Manager on your machine
2. Open the page containing calcite-loader indeterminate component
3. See the CPU utilization begin climbing and page responsiveness for rendering-heavy components begin to be impacted

### Reproduction Version

2.3

### Relevant Info

MacOS 14.1.2 (23B92)
Version 121.0.6167.139 (Official Build) (x86_64)

### Priority impact

p1 - need for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Workflow Manager

aart7367:
@geospatialem @jcfranco @DitwanP This is a pretty significant issue for Workflow Manager and is a must-have fix for R1 release. 
Are you planning to release this fix before R1? 
Would this change be picked up by the JS API team?

jcfranco:
I can't speak to the JS API picking it up, but the fix is available on `@next`. We could also coordinate a maintenance release if it will get picked up. cc @bsvensson @dasa 

As a workaround, if you own the loaders, you could apply the following style in the meantime (same as the fix):

```css
calcite-loader {
  will-change: contents;
}
```

**Note:** this won't affect loaders that are used internally by other components (i.e., loaders showing up when toggling a component's `loading` property).

dasa:
> Would this change be picked up by the JS API team?

The `@arcgis/core@rc` release currently has this dependency: @esri/calcite-components: ^2.4.0

[...]

PR #8714, opened 2024-02-09
fix(loader): optimize animation performance

jcfranco:
**Related Issue:** #8709 

## Summary

Improves performance by using `will-change` to hint at browsers to render animations in a separate layer, reducing repaints.

### Current

Screenshot 2024-02-08 at 11 01 37 PM
Screenshot 2024-02-08 at 11 00 19 PM

### Optimized (`will-change: contents`)

Screenshot 2024-02-08 at 11 00 41 PM
Screenshot 2024-02-08 at 11 00 59 PM


eriklharper (approved):
Would be cool to get a primer on `will-change` and also consider the real "why" behind the animation slowness, but this LGTM as a quick fix.

benelan:
> Would be cool to get a primer on `will-change` and also consider the real "why" behind the animation slowness, but this LGTM as a quick fix.

https://developer.mozilla.org/en-US/docs/Web/CSS/will-change

driskull (approved):
1️⃣ Line 🪄 ✨ 

💯

Issue #8728, opened 2024-02-12
[list] - unable to drag and drop list items successfully on mobile devices with drag-enabled


sagewall:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Unable to drag and drop a list item in a list with drag-enabled on mobile devices or when browser developer tools are set to mimic a mobile device. This is impacting the ArcGIS Maps SDK for JavaScript LayerList widget and all the applications that use it.

cc @driskull @geospatialem 

### Expected Behavior

It's expected to be able to move list items on mobile devices using touch gestures.

### Reproduction Sample

https://codepen.io/sagewall/pen/eYXQpmz

### Reproduction Steps

1. Open this codepen: https://codepen.io/sagewall/pen/eYXQpmz using a mobile device
2. Switch to a full screen view
3. Try and move a list item in the list

As an alternative, the same behavior exists when opening the codepen and using the [toggle device bar](https://developer.chrome.com/docs/devtools/device-mode) to switch to a mobile device in the chrome developers tools.

https://github.com/Esri/calcite-design-system/assets/2636825/5829e906-4fc4-40cd-ba9c-ca3950edd021

### Reproduction Version

2.4.0

### Priority impact

p0 - emergency

### Impact

Preventing Map Viewer, Experience Builder, Instant Apps and custom applications using the ArcGIS Maps SDK for JavaScript LayerList widget from having drag-enabled layer lists on mobile devices.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

driskull:
@jcfranco would FrancoCorp™️ have any time to help debug this one?

What i've uncovered is that it seems like when in `responsive mode` the Sortable.js appears to go away from native browser DnD and use the fallback SortableJS logic. However, that logic isn't working. I'm not sure why, it might just not work with web components or something along those lines.

[...]

PR #8730, opened 2024-02-12
#8728 prototype

driskull:
**Related Issue:** #


packages/calcite-components/src/assets/styles/_sortable.scss
```diff
@@ -1,7 +1,9 @@
 @mixin sortable-helper-classes() {
   .calcite-sortable--chosen,
   .calcite-sortable--ghost,
-  .calcite-sortable--drag {
+  .calcite-sortable--drag,
+  .calcite-sortable--fallback {
+    position: relative;
```
driskull, line 6: this part fixes the styling error for the dragging element.

packages/calcite-components/src/components/list-item/list-item.scss
```diff
@@ -2,6 +2,8 @@
   @apply flex flex-col;
   --calcite-list-item-icon-color: theme("colors.brand");
   --calcite-list-item-spacing-indent: theme("spacing.4");
+  // uncomment this and it works with `forceFallback: true,`
+  // padding-block: 10px;
```
driskull, line 6: Introducing a padding between items seems to fix the dragging when fallback is enabled. Need to figure out why dragging on a web component isnt triggering elements to swap unless there is space between them. cc @jcfranco

packages/calcite-components/src/utils/sortableComponent.ts
```diff
@@ -13,6 +13,7 @@ export const CSS = {
   ghostClass: "calcite-sortable--ghost",
   chosenClass: "calcite-sortable--chosen",
   dragClass: "calcite-sortable--drag",
+  fallbackClass: "calcite-sortable--fallback",
```
driskull, line 16: Added fallback class css for usage if necessary.

driskull:
@jcfranco I can get sortable working if I change the following:

https://github.com/SortableJS/Sortable/blob/07708fa1e7d9cf9d4563a7b1a09f7a47771522ed/src/Sortable.js#L764-L770


From:

```js
			if (parent) {
				do {
					if (parent[expando]) {
						let inserted;

						inserted = parent[expando]._onDragOver({
							clientX: touchEvt.clientX,
```

To:

```js
      if (parent) {
        do {
          // todo
          const test = parent[expando] || dragEl.parentNode[expando];
          if (test) {
            var inserted = void 0;
            inserted = test._onDragOver({
              clientX: touchEvt.clientX,
```

jcfranco:
Thanks for working on this! ✨💪✨

Could we apply the fallback and extra padding only for mobile platforms (via [userAgent](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/utils/browser.ts#L13))? We'd have to note this as a known issue for the time being. cc @ashetland @geospatialem 

I was not able to find any similar issue in SortableJS, so this might be an upstream issue that they would need to address. Could you submit an issue for this? Sharing your [findings above](https://github.com/Esri/calcite-design-system/pull/8730#issuecomment-1942934093) would be helpful.

PR #8734, opened 2024-02-13
docs(input-number): deprecate `minLength`/`maxLength` props as they have no effect

jcfranco:
**Related Issue:** N/A

## Summary

Deprecates `minLength`/`maxLength` since they have no effect on numeric inputs. We can remove them at a future breaking change release.

This stems from https://github.com/Esri/calcite-design-system/pull/8655#discussion_r1468938091.


driskull (approved):
👍

geospatialem (approved):
Good catch! 🎣 

We don't have something similar in place now, but maybe the additional context provides additional support of the "why" for folks.

WDYT about `@deprecated Has no effect on the component`?

PR #8736, opened 2024-02-13
feat(combobox, combobox-item): add component tokens

alisonailea:
**Related Issue:** #7180 

## Summary

adds component tokens to combobox and combobox item.


packages/calcite-components/src/components/combobox-item/combobox-item.scss
```diff
@@ -1,37 +1,96 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * --calcite-combobox-item-text-color: defines the text color of a combobox item.
+ * --calcite-combobox-item-selection-icon-color: defines the icon color of the dot or custom icon shown at the start of a combobox item.
+ * --calcite-combobox-item-icon-color: defines the color of a custom combobox-item icon
+ * --calcite-combobox-item-box-shadow: defines the box shadow of a combobox item.
```
macandcheese, line 9: Should box shadow be applied to Combobox Items? or just border?

packages/calcite-components/src/components/combobox/combobox.scss
```diff
@@ -3,10 +3,46 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
- * @prop --calcite-combobox-input-height: Specifies the height of the component's input.
+ * @prop --calcite-combobox-border: defines the border around the combobox input and dropdown. References the --calcite-combobox-border-color by default.
```
macandcheese, line 6: Do we want to have a general border-color prop here? I think it's unlikely an override to the combobox input would also be desired on the dropdown. Maybe just separate combobox-border and combobox-dropdown-border?

As this is a form element I think we'd want to support border-color, shadow, and corner-radius overrides that default to current border color, shadow-none, and sharp.

packages/calcite-components/src/components/combobox-item/combobox-item.scss
alisonailea: removed.

PR #8745, opened 2024-02-14
feat(icon): add component tokens

alisonailea:
**Related Issue:** #7180 

## Summary

Add new component tokens

```
--calcite-icon-color: defines the color of the component. Defaults to `currentColor`.
--calcite-icon-size: defines the size of the component.
```

### DIFF
--- a/packages/calcite-components/src/components/icon/icon.scss
+++ b/packages/calcite-components/src/components/icon/icon.scss
@@ -5,7 +5,12 @@
  *
  * @prop --calcite-ui-icon-color: *deprecated* in favor of `--calcite-icon-color`. The component's color. Defaults to `currentColor`.
  * @props --calcite-icon-color: defines the color of the component. Defaults to `currentColor`.
- * @props --calcite-icon-size: defines the size of the component.
+ */
+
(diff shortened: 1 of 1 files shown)


packages/calcite-components/src/components/icon/icon.scss
```diff
@@ -3,36 +3,32 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
- * @prop --calcite-ui-icon-color: The component's color. Defaults to `currentColor`.
+ * @prop --calcite-ui-icon-color: *deprecated* in favor of `--calcite-icon-color`. The component's color. Defaults to `currentColor`.
+ * @props --calcite-icon-color: defines the color of the component. Defaults to `currentColor`.
+ * @props --calcite-icon-size: defines the size of the component.
```
macandcheese, line 8: I don't think we should expose size - these are really only supposed to be rendered at the three `scale` values.

macandcheese: Just `internal-` is probably fine for this one.

Issue #8729, opened 2024-02-12
Disabled buttons where the 'disabled' attribute is set after the button is created are clickable in Firefox (raw Calcite)


nCastle1:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Calcite buttons (and other controls, like Checkbox) that have been disabled with the `disabled` property still fire `onclick` events.

### Expected Behavior

Calcite buttons and other components should suppress the onclick event when the control is disabled.

### Reproduction Sample

[Gist](https://gist.github.com/nCastle1/ae228f15f5c8aaa7f95514590bed07f0) (I don't have a codepen account)

It's pretty simple though:

```html
<calcite-button id="abc" onclick="alert()">Test button</calcite-button>

<calcite-button onclick="document.getElementById('abc').setAttribute('disabled', 'true')">Disable other button</calcite-button>
```

### Reproduction Steps

1. Click the 'disable other button' button and notice that the other button becomes disabled
2. Click the disabled button, and notice that the onclick event fires

### Reproduction Version

2.4.0

### Relevant Info

Observed in Firefox on Mac.

I believe an earlier issue with firefox buttons not honoring the 'disabled' property was partially addressed in this PR: https://github.com/Esri/calcite-design-system/pull/7107/files.

I played around with a local build and have a [PR](https://github.com/nCastle1/calcite-design-system/pull/1) that appears to fix the problem, but this was made to satisfy my curiosity and is more of a proof of concept than a merge-worthy PR.

There is another issue filed specifically for React that I suspect shares a root cause with this issue, but I don't know. https://github.com/Esri/calcite-design-system/issues/8203

### Priority impact

p3 - want for upcoming milestone

### Impact

The ArcGIS Maps SDK uses Calcite Components in quite a few widgets. The workaround (check disabled state within the event handler) is straightforward, but will have to be done in several places.

Because the code for implementing the current Firefox disablement fix is shared between several components, several components are likely impacted. I only checked Checkbox and Button.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

jcfranco:
[...]

PR #8746, opened 2024-02-14
fix: prevent interaction when component is disabled after initialization (Firefox)

jcfranco:
**Related Issue:** #8729 

## Summary

This updates the interactive util to track disabled component parents (needed for Firefox workaround) as the component is disabled and enabled.

**Note**: test environment is Chromium-based, so no tests were updated.


alisonailea (approved):
👍🏻

PR #8750, opened 2024-02-14
fix: fix styling when dragging items on a mobile device

driskull:
**Related Issue:** #8728

## Summary

- Fixes fallback ghost element styling to not overflow its container.
- adds fallback class for sortable items.


jcfranco (approved):
Awesome! ✨👻✨

Before merging, could you revisit the title and phrase it more from a consumer's perspective? We don't mention Sortable nor its fallback mode in the doc site.

PR #8754, opened 2024-02-14
feat(table): Improve accessibility of interaction-mode: static Table

macandcheese:
**Related Issue:** #8659 

## Summary
Follow up items cc @geospatialem


geospatialem (commented):
JAWS and NVDA are coming in as-expected 🙌🏻

Issue #8461, opened 2023-12-19
Stepper component 2.0 update breaks BA Experience builder widget


ddred11:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Current updated behavior of the stepper widget breaks our UI for the Business Analyst widget in Experience Builder. At the default width of our widget, user can no longer see additional steps or easily click on them to forward the experience. Making this widget wider is a challenge given the canvas space in Experience Builder.
![Screenshot 2023-12-19 at 2 54 07 PM](https://github.com/Esri/calcite-design-system/assets/5332496/7faf6692-27eb-4b73-9f2b-aa04ba70ac4c). 

### Acceptance Criteria

We would like to have the component provided properties that would allow developers to set the width where the responsive design of the widget trigger so that we can return our widget to this experience.
![Screenshot 2023-12-19 at 2 14 27 PM](https://github.com/Esri/calcite-design-system/assets/5332496/677f1289-0682-4d41-8786-df67e73741a4)

### Relevant Info

Breaks existing functionality. Will affect current users and the apps they have created with this widget.

### Which Component

Stepper component

### Priority impact

p1 - need for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Business/Community Analyst

macandcheese:
This will also affect anyone trying to use the Stepper in Modal, as the width is often "below" the current breakpoint to swap to this mode. As a designer / dev user, I want to be able to ensure the Stepper doesn't enter this mode.
![Screenshot 2024-01-04 at 12 20 44 PM](https://github.com/Esri/calcite-design-system/assets/4733155/b05a0423-3088-413b-b697-ad088f5a37ec)

SkyeSeitz:
Issue #8174 is to add the component tokens to do this customization. In the meantime, can we perhaps add the css vars for Stepper Item's `minWidth` value at each scale? @geospatialem Is it possible to prioritize for this milestone? 

cc @alisonailea

geospatialem:
> In the meantime, can we perhaps add the css vars for Stepper Item's `minWidth` value at each scale? @geospatialem Is it possible to prioritize for this milestone?

[...]

PR #8758, opened 2024-02-15
fix(stepper): no longer adds default `min-width` for items when `layout='horizontal'`

anveshmekala:
**Related Issue:**  #8461

## Summary

When `layout="horizontal"`, allows `stepper-item's` to shrink with container width.

### DIFF
--- a/packages/calcite-components/src/components/stepper/stepper.tsx
+++ b/packages/calcite-components/src/components/stepper/stepper.tsx
@@ -423,7 +423,7 @@ export class Stepper implements LocalizedComponent, T9nComponent {
 
     let newIndex = startIndex;
 
-    while (items[newIndex]?.disabled && this.multipleViewMode) {
+    while (items[newIndex]?.disabled && this.layout !== "horizontal-single") {
       newIndex = newIndex + (direction === "previous" ? -1 : 1);
     }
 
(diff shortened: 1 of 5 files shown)


packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -386,7 +388,7 @@ export class Stepper implements LocalizedComponent, T9nComponent {
 
   private resizeObserver = createObserver(
     "resize",
-    (entries) => (this.elWidth = entries[0].contentRect.width),
+    (entries) => (this.elWidth = entries[0].contentRect.width)
```
driskull, line 391: @anveshmekala can you check your stylelint/prettier config? I think the previous should be correct.

driskull (approved):
@anveshmekala can you check the formatting settings as I think trailing commas are expected. Otherwise looks good 👍

jcfranco (approved):
Awesome! 📏

Good to merge once linter/formatting comments have been addressed.

geospatialem:
Added the `low risk` label for the maintenance release, as the above will revert `stepper` to the 2.0 version of the component.

packages/calcite-components/src/components/stepper/stepper.tsx
anveshmekala: Will be handled in followup.

PR #8761, opened 2024-02-15
fix: ensure ui-icons are copied from correct path

benelan:
**Related Issue:** #

## Summary

Re-generate package-lock.json to try to fix the release CI, which had the following error:
https://github.com/Esri/calcite-design-system/actions/runs/7922511667/job/21630265819#step:4:87

This was the suggested solution in a comment on a related issue:
https://github.com/nrwl/nx/issues/15452#issuecomment-1485839378

If that doesn't work, we can try one of these potential solutions:
https://github.com/npm/cli/issues/4828#issuecomment-1913543338
https://github.com/vercel/turbo/issues/3328#issuecomment-1413206817

Unfortunately Calcite Components v2.5.0 does not contain the ui-icons due to releasing locally after re-generating the package-lock. I'm not sure if the corrupted package-lock was installing `@esri/calcite-ui-icons` in the wrong node_modules. Or maybe updating to Node v20 resolved some weirdness with where packages are installed when using workspaces. Either way, the icons are now installed into the node_modules in the calcite-components workspace, rather than the root.


jcfranco (approved):
🚀

PR #8751, opened 2024-02-14
fix: fix dragging items on a mobile device

driskull:
**Related Issue:** #8728

## Summary

- Fixes dragging when on a mobile device.
- Sortablejs uses fallback code instead of native drag and drop on mobile devices.
- Patches `sortablejs` dependency until the fix for the issue is installed and released. https://github.com/SortableJS/Sortable/issues/2346
- Adds and uses the dependency `patch-package` to handle patching the dependency.
- Applies the changes listed here: https://github.com/SortableJS/Sortable/pull/2347

PR #8766, opened 2024-02-16
fix: prevent package patching on install

jcfranco:
**Related Issue:** N/A

## Summary

Cleans up package-patching scripts (introduced in https://github.com/Esri/calcite-design-system/pull/8751) to prevent them from running on `calcite-components` installs.


jcfranco:
Merging to proceed with today’s patch release. LMK if there’s anything that needs following up.

PR #8767, opened 2024-02-16
feat(input): add component tokens

alisonailea:
**Related Issue:** #7180

## Summary

Adds the following component tokens to the calcite-input component styles

```
--calcite-input-corner-radius: defines the border radius of the component.
--calcite-input-text-color: defines the text color of the component.
--calcite-input-border-color: defines the border color of the component.
--calcite-input-background-color: defines the background color of the component.
--calcite-input-button-background-color: defines the background color of a button element in the component.
--calcite-input-button-background-color-hover: defines the background color of a :hover-ed button element in the component.
--calcite-input-button-background-color-active: defines the background color of an :active button element in the component.
--calcite-input-icon-color: defines the color of an icon element in the component.
--calcite-input-button-icon-color-hover: defines the color of an icon element when it's parent is hovered in the component.
--calcite-input-prefix-text-color: defines the prefix text color in the component.
--calcite-input-prefix-background-color: defines the prefix background color in the component.
--calcite-input-suffix-text-color: defines the suffix text color in the component.
--calcite-input-suffix-background-color: defines the suffix background color in the component.
--calcite-input-placeholder-text-color: defines the color of placeholder text in the component.
--calcite-input-shadow: defines the box-shadow of the component.
```

### DIFF
--- a/packages/calcite-components/src/components/input/input.scss
+++ b/packages/calcite-components/src/components/input/input.scss
@@ -9,6 +9,10 @@
   --calcite-input-button-background-color-hover: var(--calcite-color-foreground-2);
   --calcite-input-button-background-color-active: var(--calcite-color-foreground-3);
 
+  // Icon
+  --calcite-input-icon-color: var(--calcite-color-text-3);
+  --calcite-input-icon-color-hover: var(--calcite-color-text-1);
+
   // For props that should follow the initial border-color but not change on statechange.
   --calcite-internal-input-border-color-base: var(--calcite-input-border-color);
 
@@ -150,7 +154,11 @@ textarea[readonly] {
   @apply font-medium;
 }
 calcite-icon {
-  @apply text-color-3;
+  color: var(--calcite-input-icon-color);
+}
+button:hover,
+button:active {
+  --calcite-input-icon-color: var(--calcite-input-icon-color-hover);
 }
 
 //focus
@@ -265,14 +273,11 @@ input[type="time"]::-webkit-clear-button {
     --calcite-input-button-background-color: var(--calcite-input-button-background-color-hover);
 
     calcite-icon {
-      @apply text-color-1 transition-default;
+      @apply transition-default;
     }
   }
   &:active {
     --calcite-input-button-background-color: var(--calcite-input-button-background-color-active);
-    calcite-icon {
-      @apply text-color-1;
-    }
   }
   &:focus {
     @apply focus-inset;
@@ -401,32 +406,10 @@ input[type="number"] {
     border-solid;
   border-color: var(--calcite-internal-input-border-color-base);
   border-inline-end-width: theme("borderWidth.0");
-  &:hover {
-    calcite-icon {
-      @apply text-color-1;
-    }
-  }
 }
 
 .number-button-item.number-button-item--horizontal[data-adjustment="up"] {
   @apply order-5;
-  &:hover {
-    calcite-icon {
-      @apply text-color-1;
-    }
-  }
-}
-
-:host([number-button-type="vertical"]) .number-button-item[data-adjustment="down"]:hover {
-  calcite-icon {
-    @apply text-color-1;
-  }
-}
-
-:host([number-button-type="vertical"]) .number-button-item[data-adjustment="up"]:hover {
-  calcite-icon {
-    @apply text-color-1;
-  }
 }
 
 :host([number-button-type="vertical"]) .number-button-item[data-adjustment="down"] {
@@ -457,9 +440,6 @@ input[type="number"] {
   &:focus,
   &:hover {
     --calcite-input-button-background-color: var(--calcite-input-button-background-color-hover);
-    calcite-icon {
-      @apply text-color-1;
-    }
   }
   &:active {
     --calcite-input-button-background-color: var(--calcite-input-button-background-color-active);
@@ -500,8 +480,7 @@ textarea::-webkit-resizer {
   inset-inline-end: 2px;
   background-color: var(--calcite-input-button-background-color);
 
-  @apply text-color-3
-    pointer-events-none
+  @apply pointer-events-none
     absolute
     h-3
     w-3;
--- a/packages/calcite-components/src/components/input/input.scss
+++ b/packages/calcite-components/src/components/input/input.scss
@@ -1,6 +1,10 @@
 :host {
   --calcite-input-corner-radius: var(--calcite-border-radius-sharp);
   --calcite-input-text-color: var(--calcite-color-text-1);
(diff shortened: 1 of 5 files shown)


packages/calcite-components/src/components/input/input.scss
```diff
@@ -1,4 +1,48 @@
...
+ * @prop --calcite-input-corner-radius: defines the border radius of the component.
+ * @prop --calcite-input-text-color: defines the text color of the component.
+ * @prop --calcite-input-border-color: defines the border color of the component.
+ * @prop --calcite-input-background-color: defines the background color of the component.
+ * @prop --calcite-input-button-background-color: defines the background color of a button element in the component.
+ * @prop --calcite-input-button-background-color-hover: defines the background color of a :hover-ed button element in the component.
+ * @prop --calcite-input-button-background-color-active: defines the background color of an :active button element in the component.
+ * @prop --calcite-input-icon-color: defines the color of an icon element in the component.
+ * @prop --calcite-input-icon-color-hover: defines the color of an icon element when it's parent is hovered in the component.
+ * @prop --calcite-input-prefixsuffix-text-color: defines the text color of a prefix/suffix element in the component.
```
macandcheese, line 15: I could see users wanting to customize these individually pretty easily, wdyt cc  @SkyeSeitz @ashetland

packages/calcite-components/src/components/input/input.scss
```diff
@@ -1,4 +1,48 @@
...
+ *
+ * @prop --calcite-input-corner-radius: defines the border radius of the component.
+ * @prop --calcite-input-text-color: defines the text color of the component.
+ * @prop --calcite-input-border-color: defines the border color of the component.
+ * @prop --calcite-input-background-color: defines the background color of the component.
+ * @prop --calcite-input-button-background-color: defines the background color of a button element in the component.
+ * @prop --calcite-input-button-background-color-hover: defines the background color of a :hover-ed button element in the component.
+ * @prop --calcite-input-button-background-color-active: defines the background color of an :active button element in the component.
+ * @prop --calcite-input-icon-color: defines the color of an icon element in the component.
+ * @prop --calcite-input-icon-color-hover: defines the color of an icon element when it's parent is hovered in the component.
```
macandcheese, line 14: Can the user set this with calcite-input: hover or is it for a different icon besides the one from ‘icon’ prop?

packages/calcite-components/src/components/input/input.scss
```diff
@@ -1,4 +1,48 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-input-corner-radius: defines the border radius of the component.
+ * @prop --calcite-input-text-color: defines the text color of the component.
```
macandcheese, line 7: Can we add an input-shadow that defaults to none?

packages/calcite-components/src/components/input/input.scss
```diff
@@ -1,4 +1,48 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-input-corner-radius: defines the border radius of the component.
```
macandcheese, line 6: Conditionally handle the closable button with this value: 
Screenshot 2024-02-16 at 9 15 36 AM

packages/calcite-components/src/components/input/input.scss
alisonailea: it's different.

packages/calcite-components/src/components/input/input.scss
alisonailea: I already have the styles there to prevent an example like you've shown above. The input will not add a rounded border when a ::slotted action is added. However, I can not style the button as shown above as the button is not currently styled in such a way as to allow that.

macandcheese: Agreed, we shouldn't apply it to slotted content. But that clear button is rendered internally, IMO we should treat it like the number buttons which you already handle:

```
border-start-start-radius: var(--calcite-input-corner-radius);
border-end-start-radius: var(--calcite-input-corner-radius);
```  

It's only an issue when the close button is adjacent to the edge, you're already handling it in these cases, I think it just needs a conditional for the above state:
Screenshot 2024-02-16 at 5 23 32 PM
Screenshot 2024-02-16 at 5 22 30 PM

macandcheese (changes requested):
Looking really good! Two main things:

- Handle the `clearable` case when its the last rendered element.
- I think there are some extra border colors being set on focus / invalid.

Previously:
Screenshot 2024-02-16 at 5 40 20 PM
Screenshot 2024-02-16 at 5 40 34 PM
Screenshot 2024-02-16 at 5 40 41 PM

With this PR:
Screenshot 2024-02-16 at 5 42 12 PM
Screenshot 2024-02-16 at 5 42 14 PM
Screenshot 2024-02-16 at 5 42 30 PM

I don't mind the invalid styling around the suffix / prefix, but the focus color around number spinners shouldn't be applied I don't think.

packages/calcite-components/src/components/input/input.scss
macandcheese: Maybe `input-button-icon-color` and `input-button-icon-color-hover` just to be clearer? From the name / description it just sounds like the hover color for the icon set above (which can be affected with a parent-level pseudo state override)

macandcheese:
Colors look good! 

Border radius looks good on close button in some states :
Screenshot 2024-02-20 at 6 00 35 PM
Screenshot 2024-02-20 at 5 57 32 PM

But I think it might need another pass for certain states : 

Screenshot 2024-02-20 at 5 58 02 PM
Screenshot 2024-02-20 at 5 59 17 PM
Screenshot 2024-02-20 at 5 59 21 PM

alisonailea:
@macandcheese Thanks for the catch! I did find one error but I also updated the storybook story to remove empty "prefix-text" and "suffix-text" which were causing some of the unexpected layout bugs you saw.

Latest screenshots

Screenshot 2024-02-20 at 10 13 26 PM
Screenshot 2024-02-20 at 10 13 35 PM
Screenshot 2024-02-20 at 11 01 51 PM
Screenshot 2024-02-20 at 11 02 09 PM
Screenshot 2024-02-20 at 11 02 30 PM
Screenshot 2024-02-20 at 11 02 46 PM
Screenshot 2024-02-20 at 11 03 37 PM
Screenshot 2024-02-20 at 11 04 12 PM
Screenshot 2024-02-20 at 11 04 23 PM
Screenshot 2024-02-20 at 11 04 29 PM
Screenshot 2024-02-20 at 11 04 38 PM

PR #8771, opened 2024-02-16
refactor: Use component tokens within includes

driskull:
**Related Issue:** #7180

## Summary

- Uses component tokens within include files.
- Removes unused spacing file.
- Moves xButton mixin to its own file
- Converts header to a mixin
- removes animation import from some components (it only has mixins and functions)
- Further issues should be created to fix any components using heading or xButton
- Removes unused global vars


alisonailea:
If we're going to keep using these we should convert them to `@mixins` with passable parameters so each component can use it's own variables.

driskull:
@jcfranco @macandcheese whats the history of `x-button`? Shouldn't we just use a `calcite-action` or `calcite-button` instead? Not sure we need the xButton functional component.

macandcheese:
I think it was built to eventually support all close affordances - accepting arguments for round, square, stretch to fill, etc. as needed, but seems like it hasn't been adopted much.

alisonailea:
Input could also be a mixin, given there are so many very similar components

packages/calcite-components/src/assets/styles/_x-button.scss
```diff
@@ -0,0 +1,58 @@
...
+    align-self: center;
+    color: var(--calcite-color-text-3);
+    transition:
+      all var(--calcite-animation-timing) ease-in-out 0s,
+      outline 0s,
+      outline-offset 0s;
+    border-width: var(--calcite-border-width-md);
+    border-radius: 50%;
+    border-color: transparent;
+    background-color: var(--calcite-color-foreground-2);
```
alisonailea, line 20:
```css
inline-size: var(--calcite-internal-x-button-inline-size);
block-size: var(--calcite-internal-x-button-block-size);
```

packages/calcite-components/src/assets/styles/_x-button.scss
```diff
@@ -0,0 +1,58 @@
...
+      block-size: var(--calcite-spacing-lg);
+    }
+  }
+
+  :host([scale="l"]) {
+    .x-button {
+      inline-size: var(--calcite-spacing-xxl);
+      block-size: var(--calcite-spacing-xxl);
+    }
+  }
```
alisonailea, line 57: replace lines :38 - :57 with 

```css
:host([scale="s"]) {
  --calcite-internal-x-button-size: var(--calcite-spacing-sm);
}

:host([scale="m"]) {
  --calcite-internal-x-button-size: var(--calcite-spacing-lg);
}

:host([scale="l"]) {
    --calcite-internal-x-button-size: var(--calcite-spacing-xxl);
}
```

packages/calcite-components/src/assets/styles/includes.scss
```diff
@@ -67,7 +68,7 @@
     ::slotted([calcite-hydrated][disabled]),
     [calcite-hydrated][disabled] {
       /* prevent opacity stacking */
-      @apply opacity-100;
+      opacity: 1;
```
alisonailea, line 71: var(--calcite-opacity-full)

packages/calcite-components/src/assets/styles/_x-button.scss
```diff
@@ -0,0 +1,59 @@
...
+    align-content: center;
+    cursor: pointer;
+    display: flex;
+    outline-color: transparent;
+    align-items: center;
+    justify-content: center;
+    margin: 0px;
+    align-self: center;
+    color: var(#{$textColor}, var(--calcite-color-text-3));
+    transition:
```
jcfranco, line 20: Could you explicitly list the props that will be transitioned by state? Looks like it's `color` and `background-color`.

packages/calcite-components/src/assets/styles/_x-button.scss
```diff
@@ -0,0 +1,59 @@
+@mixin xButton(
+  $textColor: "--calcite-color-text-3",
+  $backgroundColor: "--calcite-color-foreground-2",
+  $hoverTextColor: "--calcite-color-text-3",
+  $hoverBackgroundColor: "--calcite-color-foreground-3",
+  $activeBorderColor: "--calcite-color-brand",
+  $borderWidth: "--calcite-border-width-md"
+) {
+  .x-button {
+    appearance: none;
```
jcfranco, line 10: I thought we agreed to only inline Tailwind utils related to component tokens. FWIW, we can automate this step later to make it easier for review.

jcfranco (approved):
🕹️👾

packages/calcite-components/src/assets/styles/_sortable.scss
```diff
@@ -9,20 +9,17 @@
...
-      top-0
-      left-0
-      bottom-0
-      right-0
-      z-default;
+    box-sizing: border-box;
+    border-width: var(--calcite-border-width-sm);
+    border-style: dashed;
+    border-color: var(--calcite-color-brand);
+    background-color: var(--calcite-color-foreground-2);
```
alisonailea, line 16: Should this allow for a passed component token similar to `_spacing`?

driskull: Unfortunately, we can't use component tokens for sortable.js. The library uses global classes to style elements so none of these sortable styles are scoped to any one single component.

However, users can restyle these classes since they are global classes. We could have global root vars for these if we want to allow them to style that way. what do you think @jcfranco?

For now, i'd probably like to hold off on that though since these styles are going to change with some upcoming list design changes.

alisonailea (approved):

PR #8780, opened 2024-02-19
feat(handle): add component tokens

Elijbet:
**Related Issue:** #7180

## Summary

```
--calcite-handle-text-color
--calcite-handle-text-color-hover
--calcite-handle-text-color-focus
--calcite-handle-text-color-selected

--calcite-handle-background-color
--calcite-handle-background-color-hover
--calcite-handle-background-color-focus
--calcite-handle-background-color-selected
```


packages/calcite-components/src/components/handle/handle.scss
```diff
@@ -10,7 +22,7 @@
     self-stretch
     border-none
     bg-transparent;
-  color: theme("borderColor.color.input");
+  color: var(--calcite-color-border-input);
```
alisonailea, line 25: I think this should be handled-text-color

packages/calcite-components/src/components/handle/handle.scss
```diff
@@ -23,13 +35,17 @@
 :host(:not([disabled])) .handle {
   @apply cursor-move;
   &:hover {
-    @apply bg-foreground-2 text-color-1;
+    background-color: var(--calcite-handle-background-color);
```
alisonailea, line 38: Rather than re-apply the prop, try reassigning the variable on the host

packages/calcite-components/src/components/handle/handle.scss
```diff
@@ -23,13 +35,17 @@
...
+    color: var(--calcite-handle-text-color);
   }
   &:focus {
-    @apply text-color-1 focus-inset;
+    @apply focus-inset;
+
+    color: var(--calcite-handle-text-color);
```
alisonailea, line 44: Same here. Reassign the custom-prop rather than adding new style props

packages/calcite-components/src/components/handle/handle.scss
```diff
@@ -23,13 +35,17 @@
...
   }
   &--selected {
-    @apply bg-foreground-3 text-color-1;
+    background-color: var(--calcite-color-foreground-3);
+    color: var(--calcite-handle-text-color);
```
alisonailea, line 48: Here too

alisonailea (changes requested):
Move custom props to host and style based on the host state if you can.

packages/calcite-components/src/components/handle/handle.scss
driskull: I think there needs to bet a var for text-color (border-input) and a var for hover/focus (text-color-1).

packages/calcite-components/src/components/handle/handle.scss
driskull: background-color needs two separate vars for default, hover and selected. I think text needs the same 3.

driskull:
Would be nice to have a story that applies these vars for test coverage.

packages/calcite-components/src/components/handle/handle.scss
```diff
@@ -10,27 +22,32 @@
...
-    @apply bg-foreground-2 text-color-1;
-  }
-  &:focus {
-    @apply text-color-1 focus-inset;
-  }
-  &--selected {
-    @apply bg-foreground-3 text-color-1;
-  }
+:host(:not([disabled]):hover) {
+  --calcite-handle-text-color: var(--calcite-color-text-1);
```
driskull, line 39: @Elijbet how can a user set the handle color when hovered? I think we need separate vars for these.

Elijbet: @driskull @alisonailea Ali I changed the previous setup to incorporate separate states for hover, focus, and selected states.

jcfranco: We decided to only add state tokens for component "parts" (e.g., close button on alert), but not for the host itself (see https://github.com/Esri/calcite-design-system/pull/8645#discussion_r1488300762).

driskull: That means if a user wants to customize the hover text color on this component they would need to look into the source code because they wouldn't realize they need this selector: `:host(:not([disabled]):hover) {`. In a lot of cases its not just as simple as `host(:hover){`. IMO it would be much simpler for the end user with a separate component token and they wouldn't have to be aware of these implementation specifics.

packages/calcite-components/src/components/handle/handle.scss
```diff
@@ -1,5 +1,34 @@
...
+
+* @prop --calcite-handle-background-color: Specifies the background color of the component.
+* @prop --calcite-handle-background-color-hover: Specifies the background color of the component when in hover state.
+* @prop --calcite-handle-background-color-selected: Specifies the background color of the component when selected.
+*/
+
 :host {
   @apply flex;
+
+  --calcite-handle-text-color: var(--calcite-color-border-input);
```
jcfranco, line 19: Handle doesn't have text, so this should probably use another name.

packages/calcite-components/src/components/handle/handle.scss
```diff
@@ -1,5 +1,34 @@
...
+  --calcite-handle-background-color: transparent;
+
+  /* hover */
+  --calcite-handle-text-color-hover: var(--calcite-color-text-1);
+  --calcite-handle-background-color-hover: var(--calcite-color-foreground-2);
+
+  /* focus */
+  --calcite-handle-text-color-focus: var(--calcite-color-text-1);
```
jcfranco, line 27: This is missing a background focus prop.

packages/calcite-components/src/components/handle/handle.scss
```diff
@@ -1,5 +1,34 @@
...
+* @prop --calcite-handle-background-color-selected: Specifies the background color of the component when selected.
+*/
+
 :host {
   @apply flex;
+
+  --calcite-handle-text-color: var(--calcite-color-border-input);
+  --calcite-handle-background-color: transparent;
+
+  /* hover */
```
jcfranco, line 22: Nitpick: I don't think we need these comments if we group props accordingly. We should add a note to the workflow doc about this. cc @alisonailea

jcfranco (changes requested):

packages/calcite-components/src/components/handle/handle.stories.ts
```diff
@@ -14,3 +14,20 @@ export const simple = (): string => html` <calcite-handle></calcite-handle> `;
 export const activated_TestOnly = (): string => html` <calcite-handle activated></calcite-handle> `;
 
 export const disabled_TestOnly = (): string => html` <calcite-handle disabled></calcite-handle> `;
+
+export const focused_TestOnly = (): string => html` <calcite-handle focused></calcite-handle> `;
```
jcfranco, line 18: This doesn't seem valid as `handle` doesn't have a `focused` prop.

packages/calcite-components/src/components/handle/handle.stories.ts
```diff
@@ -14,3 +14,20 @@ export const simple = (): string => html` <calcite-handle></calcite-handle> `;
...
+
+export const themedStates_TestOnly = (): string => html`
```
jcfranco, line 20: ✨🏆✨

jcfranco (changes requested):
Awesome stuff, @Elijbet! One more round of changes and I think this will be good to land. 🛬

jcfranco (approved):
🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘
🤘🎸🎸🎸🎸🤘🎸🤘🤘🤘🤘🤘🎸🤘🎸🎸🎸🎸🤘🤘🎸🎸🎸🤘🤘🎸🎸🤘🤘🎸🤘🤘🤘🎸🤘🎸🎸🎸🎸🤘🎸🤘
🤘🎸🤘🤘🎸🤘🎸🤘🤘🤘🤘🤘🎸🤘🎸🤘🤘🤘🤘🎸🤘🤘🤘🤘🎸🤘🤘🎸🤘🎸🎸🤘🎸🎸🤘🎸🤘🤘🤘🤘🎸🤘
🤘🎸🎸🎸🎸🤘🎸🤘🤘🎸🤘🤘🎸🤘🎸🎸🎸🤘🤘🤘🎸🎸🤘🤘🎸🤘🤘🎸🤘🎸🤘🎸🤘🎸🤘🎸🎸🎸🤘🤘🎸🤘
🤘🎸🤘🤘🎸🤘🎸🤘🎸🤘🎸🤘🎸🤘🎸🤘🤘🤘🤘🤘🤘🤘🎸🤘🎸🤘🤘🎸🤘🎸🤘🤘🤘🎸🤘🎸🤘🤘🤘🤘🤘🤘
🤘🎸🤘🤘🎸🤘🤘🎸🤘🤘🤘🎸🤘🤘🎸🎸🎸🎸🤘🎸🎸🎸🤘🤘🤘🎸🎸🤘🤘🎸🤘🤘🤘🎸🤘🎸🎸🎸🎸🤘🎸🤘
🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘🤘

jcfranco:
@alisonailea Would you mind giving this a quick glance? 👀

packages/calcite-components/src/components/handle/handle.scss
```diff
@@ -1,3 +1,19 @@
+/**
+* CSS Custom Properties  
+*
+* These properties can be overridden using the component's tag as selector.
+*
+* @prop --calcite-handle-icon-color: Specifies the icon color of the component.
+* @prop --calcite-handle-icon-color-hover: Specifies the icon color of the component when in hover state.
```
driskull, line 7: question: do we prefer the term `hover state` or `hovered`? cc @geospatialem 

Should this be worded like `defines the color of an icon sub-component when hovered  inside the component.`

Thats how i've seen these worded in other PRs. I guess we'll probably need to do some cleanup/consistency doc checks before we merge the "real" PR?

driskull (approved):
🎉

packages/calcite-components/src/components/handle/handle.scss
driskull: Maybe we can check the epic PR every so often and fix any inconsistencies in wording/description?

geospatialem: > question: do we prefer the term hover state or hovered? cc @geospatialem
> 
> Should this be worded like `defines the color of an icon sub-component when hovered  inside the component.`

Great question. The sentence reads a bit weird when coupled with `"when hovered"`. WDYT about another option altogether, and something like `"on hover"`? For instance:

> Specifies the component's icon color on hover.

alisonailea (approved):

packages/calcite-components/src/components/handle/handle.scss
Elijbet: I like the simplicity of that phrasing.

PR #8782, opened 2024-02-20
feat(tip-manager): add component tokens

Elijbet:
**Related Issue:** #7180

## Summary

```
--calcite-tip-manager-content-height
--calcite-tip-manager-background-color
--calcite-tip-manager-text-color
--calcite-tip-manager-border-color
```


packages/calcite-components/src/components/tip-manager/tip-manager.scss
```diff
@@ -1,16 +1,19 @@
+@import "../../assets/styles/header";
+
 /**
  * CSS Custom Properties
  *
  * These properties can be overridden using the component's tag as selector.
  *
- * @prop --calcite-tip-manager-height: The maximum height of the component.
  * @prop --calcite-tip-max-width: The maximum width of a slotted `calcite-tip` within the component.
```
jcfranco, line 8: Sidebar: we should look into:

1. adding a matching prop on `tip` and deprecate this one or
2. deprecate this prop and let users set max-width/max-inline-size on `tip` directly.

jcfranco (changes requested):
Looking good, @Elijbet! 😎

packages/calcite-components/src/components/tip-manager/tip-manager.scss
```diff
@@ -1,24 +1,35 @@
...
+* These properties can be overridden using the component's tag as selector.
+*
+* @prop --calcite-tip-max-width: [Deprecated] No longer necessary. 
+* @prop --calcite-tip-manager-height: [Deprecated] No longer necessary. 
+* @prop --calcite-tip-manager-background-color: The background color of the component.
+* @prop --calcite-tip-manager-heading-text-color: The heading text color of the component.
+* @prop --calcite-tip-manager-text-color: The text color of the component.
+* @prop --calcite-tip-manager-border-color: The border color of the component.
+* 
+* @prop --calcite-tip-manager-tip-border-color: The border color of the sub-component.
```
jcfranco, line 15: I don't think these tip-specific props are applicable. Tips are slotted into the manager and are not internal.

packages/calcite-components/src/components/tip-manager/tip-manager.scss
```diff
@@ -1,24 +1,35 @@
...
- * These properties can be overridden using the component's tag as selector.
- *
- * @prop --calcite-tip-manager-height: The maximum height of the component.
- * @prop --calcite-tip-max-width: The maximum width of a slotted `calcite-tip` within the component.
- */
+* CSS Custom Properties
+*
+* These properties can be overridden using the component's tag as selector.
+*
+* @prop --calcite-tip-max-width: [Deprecated] No longer necessary.
```
jcfranco, line 8: Shouldn't the deprecated ones point to the matching props to use? Or why are these no longer necessary?

packages/calcite-components/src/components/tip-manager/tip-manager.scss
```diff
@@ -65,15 +71,14 @@
     p-4;
   animation-name: none;
   animation-duration: var(--calcite-animation-timing);
-  block-size: var(--calcite-tip-manager-height);
+  block-size: var(--calcite-internal-tip-manager-block-size, 19vh);
```
jcfranco, line 74: `--calcite-internal-tip-manager-block-size` doesn't seem to be defined internally, so it can probably be removed.

packages/calcite-components/src/components/tip-manager/tip-manager.scss
```diff
@@ -1,16 +1,19 @@
+@import "../../assets/styles/header";
+
 /**
  * CSS Custom Properties
  *
  * These properties can be overridden using the component's tag as selector.
  *
- * @prop --calcite-tip-manager-height: The maximum height of the component.
```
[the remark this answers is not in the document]
alisonailea: They're documented so we should point to the new token.

packages/calcite-components/src/components/tip-manager/tip-manager.scss
alisonailea: These are both documented. My guess is they were requested by end users. So this is probably an edge case where we should keep them. Maybe rename to `--calcite-tip-max-size-x` and `--calcite-tip-max-size-y`

alisonailea (commented):
Looking good. Just a few things.

packages/calcite-components/src/components/tip-manager/tip-manager.scss
jcfranco: In this particular case, `--calcite-tip-max-width` is used to set the max-width directly on slotted `tip`s. Wouldn't we want to deprecate it in favor of rules targeting `tip`s to set the max-width? It'd be one less prop to maintain.

macandcheese: Yeah, in these cases a user could also just set these css rules directly on the Tip element correct, as it is slotted into the Tip Manager? 

Anecdotally, Tip / Tip Manager is one of our oldest and most seldom used components, and is slated for deprecation once Carousel lands. I think it's relatively safe to deprecate these - they'll continue to work and we don't have to add or support new css properties.

packages/calcite-components/src/components/tip-manager/tip-manager.scss
```diff
@@ -1,31 +1,39 @@
...
+* @prop --calcite-tip-manager-text-color: The text color of the component.
+* @prop --calcite-tip-manager-border-color: The border color of the component.
+*
+* @prop --calcite-tip-manager-action-background-color: defines the background color of an action sub-component inside the component.
+* @prop --calcite-tip-manager-action-text-color: defines the text color of an action sub-component inside the component.
+* @prop --calcite-tip-manager-action-background-color-hover: defines the background color of an action sub-component when hovered or focused inside the component.
+* @prop --calcite-tip-manager-action-text-color-hover: defines the text color of an action sub-component when hovered or focused inside the component.
+* @prop --calcite-tip-manager-action-background-color-active: defines the background color of an action sub-component when active inside the component.
+* @prop --calcite-tip-manager-action-text-color-active: defines the text color of an action sub-component when active inside the component.
+* @prop --calcite-tip-manager-height: The maximum height of the component.
```
jcfranco, line 21: This is duplicated.

alisonailea, line 21: I think this should be `--calcite-tip-manager-size-y` for consistency.

packages/calcite-components/src/components/tip-manager/tip-manager.scss
```diff
@@ -1,31 +1,39 @@
...
+* @prop --calcite-tip-max-width: The maximum width of a slotted `calcite-tip` within the component.
```
alisonailea, line 22: I think this should be `--calcite-tip-max-size-x`

packages/calcite-components/src/components/tip-manager/tip-manager.scss
```diff
@@ -1,31 +1,39 @@
...
- * @prop --calcite-tip-manager-action-text-color-active: defines the text color of an action sub-component when active inside the component.
- * @prop --calcite-tip-manager-height: The maximum height of the component.
- * @prop --calcite-tip-max-width: The maximum width of a slotted `calcite-tip` within the component.
- *
- */
+* CSS Custom Properties
+*
+* These properties can be overridden using the component's tag as selector.
+*
+* @prop --calcite-tip-max-width: [Deprecated] No longer necessary, as we don't use mapped tokens for slotted components. Specifies the maximum width of a slotted`calcite-tip` within the component.
```
alisonailea, line 8: Since this was documented in the doc-site, we need to point users to the updated token name.

packages/calcite-components/src/components/tip-manager/tip-manager.scss
```diff
@@ -72,15 +81,14 @@
     p-4;
   animation-name: none;
   animation-duration: var(--calcite-animation-timing);
-  block-size: var(--calcite-tip-manager-height);
+  block-size: var(--calcite-internal-tip-manager-block-size, 19vh);
```
alisonailea, line 84: I think this should be `--calcite-tip-manager-size-y`

packages/calcite-components/src/components/tip-manager/tip-manager.scss
```diff
@@ -72,15 +81,14 @@
...
   &:focus {
     @apply focus-outset;
   }
 }
 
 ::slotted(calcite-tip) {
   @apply m-0 border-none;
-  max-inline-size: var(--calcite-tip-max-width);
```
alisonailea, line 83: I think this should be `--calcite-tip-manager-max-size-x`

alisonailea (changes requested):
Just a few more things.

packages/calcite-components/src/components/tip-manager/tip-manager.scss
Elijbet: Can we talk about this naming to make sure we're all on the same page with renaming width/height as size-x/-y?
@macandcheese @jcfranco

macandcheese: I’m not sure this is needed at all. A user can set a desired width directly on the Tip that is slotted. 

That’s how Carousel will function as well which is the replacement component / migration path for these Tip components.

macandcheese: This one could just be deprecated I think.

alisonailea: I'd agree with deprecation.

alisonailea (approved):

alisonailea (approved):

PR #8783, opened 2024-02-20
fix(tabs): add component tokens

Elijbet:
**Related Issue:** #7180

## Summary

```
--calcite-tabs-background-color
--calcite-tabs-border-color 
 ```

### CODE REVIEW

File: packages/calcite-components/src/components/tabs/tabs.scss [hunk 1]
```diff
@@ -1,10 +1,22 @@
...
+ * @prop --calcite-tabs-background-color: The background color of the component.
+ * @prop --calcite-tabs-border-color: The border color of the component.
+ */
+
 :host {
   @apply flex flex-col;
 }
 
 :host([bordered]) {
+  --calcite-tabs-background-color: var(--calcite-color-foreground-1);
```
driskull, line 15: I think these should be on the host and not `:host([bordered]` right?

driskull (approved):
Looks good once vars are moved to host.

packages/calcite-components/src/components/tabs/tabs.scss
Elijbet: I think background color is only set on bordered.

Elijbet: That's right, made a blunder.

jcfranco (approved):

PR #8786, opened 2024-02-20
feat(switch): add component tokens

jcfranco:
**Related Issue:** #7180 

## Summary

* `--calcite-switch-handle-background-color`
* `--calcite-switch-handle-border-color`
* `--calcite-switch-track-background-color`
* `--calcite-switch-track-border-color`
* `--calcite-switch-corner-radius`
* `--calcite-switch-shadow`

### DIFF
--- a/packages/calcite-components/src/components/switch/switch.scss
+++ b/packages/calcite-components/src/components/switch/switch.scss
@@ -1,125 +1,131 @@
-// sizes
-:host([scale="s"]) {
-  .container {
-    @apply h-3;
-  }
-  .track {
-    @apply h-3 w-6;
-  }
-  .handle {
-    @apply h-2 w-2;
-  }
-}
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-switch-handle-background-color: Specifies the background color of the component handle.
+ * @prop --calcite-switch-handle-border-color: Specifies the border color of the component handle.
+ * @prop --calcite-switch-track-background-color: Specifies the background color of the component track.
+ * @prop --calcite-switch-track-border-color: Specifies the border color of the component track.
+ * @prop --calcite-switch-corner-radius: Specifies the border radius of the component.
+ * @prop --calcite-switch-shadow: Specifies the box shadow of the component.
+ */
 
-:host([scale="m"]) {
-  .container {
-    @apply h-4;
-  }
-  .track {
-    @apply h-4 w-8;
-  }
-  .handle {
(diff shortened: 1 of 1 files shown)


packages/calcite-components/src/components/switch/switch.scss
```diff
@@ -1,125 +1,131 @@
...
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-switch-handle-background-color: Specifies the background color of the component handle.
+ * @prop --calcite-switch-handle-border-color: Specifies the border color of the component handle.
+ * @prop --calcite-switch-track-background-color: Specifies the background color of the component track.
+ * @prop --calcite-switch-track-border-color: Specifies the border color of the component track.
+ * @prop --calcite-switch-corner-radius: Specifies the border radius of the component.
```
macandcheese, line 10: Specifies the "corner radius"

packages/calcite-components/src/components/switch/switch.scss
```diff
@@ -1,125 +1,131 @@
...
+ * @prop --calcite-switch-shadow: Specifies the box shadow of the component.
```
macandcheese, line 11: Specifies the "shadow" (vs. box shadow), I think.

alisonailea (approved):

macandcheese:
There are some "invisible" height changes in Chromatic, lets make sure this won't affect layout somehow: https://www.chromatic.com/test?appId=6266d45509d7eb004aa254fb&id=65d535a602d615a0c2d7deba

jcfranco:
Forgot to bring this up sooner, but name-wise, do we prefer `handle` vs `thumb`? This might be the first instance of handle/thumb in a public property name (related to that UI part). I'm used to "thumb" and is slightly shorter, but wanted to confirm for consistency. cc @geospatialem @DitwanP

geospatialem:
> Forgot to bring this up sooner, but name-wise, do we prefer `handle` vs `thumb`? This might be the first instance of handle/thumb in a public property name (related to that UI part). I'm used to "thumb" and is slightly shorter, but wanted to confirm for consistency.

We use `handle` for the `list-item` event, so lean towards `handle`, unless we want to differentiate between the list event and switch prop.
https://github.com/Esri/calcite-design-system/blob/802ccf2272985d31bd5afa2383c24e669913ee5c/packages/calcite-components/src/components/list-item/list-item.tsx#L249-L251

driskull:
I prefer handle since thumb might be confused with thumbnail.

Issue #4932, opened 2022-07-14
[Block] Add icon property, update slot nomenclature


macandcheese:
### Description

The current Block slot documentation / naming is overly limiting, in practice an action or other element may be applicable :

https://developers.arcgis.com/calcite-design-system/components/block/#component-api-slots-control

### Which Component

Block

Proposing that we rename the current slots to ones more consistent with the system:

Control -> Actions End
Icon -> Content Start

Additionally: 

Icons are extremely common to see in Blocks. The vast majority of the time, they are simply using Calcite UI Icons that could be provided via prop. Sometimes they need to be custom image or other graphic, and for those cases, the "content-start" slot should be sufficient.

Add: Add an icon property that can be used in lieu of icon slot. Of course, folks can still slot a custom icon here, but providing an icon slot will ensure more consistently implemented designs in more cases while still preserving the location for custom icon or image. This will display alongside "content-slot", and if populated, that slot should not replace this icon provided via property.

macandcheese:
Probably a good candidate for the internal Stack component here - even "icon" could probably just be "content-start".

ashetland:
Related #4935

jcfranco:
Similar to https://github.com/Esri/calcite-design-system/issues/6918, existing slots should be deprecated to allow developers time to migrate to the new ones before we remove them in a future breaking change release.

ashetland:
[Figma file](https://www.figma.com/file/7uLiqKs0ZGUeMa8OmNQQoP/Block%2FBlock-Section-icon-prop-and-slot-naming-%5B4932%2C-8110%5D?type=design&node-id=2754-34186&mode=dev) documenting the changes for the above and deprecating the `status` prop to align with #8110. Recommend implementing them together.

DitwanP:
🍡 Verified locally on `dev`

PR #8790, opened 2024-02-20
feat(block): add component tokens

jcfranco:
**Related Issue:** #7180

## Summary

* `--calcite-block-background-color` 
* `--calcite-block-border-color` 
* `--calcite-block-content-space` 
* `--calcite-block-description-text-color`
* `--calcite-block-header-background-color-hover`
* `--calcite-block-header-background-color` 
* `--calcite-block-heading-text-color-open` 
* `--calcite-block-heading-text-color` 
* `--calcite-block-text-color`
* `--calcite-block-toggle-icon-color`

Deprecates the following:

* `--calcite-block-padding`

**Note**: added internal icon color token for [upcoming `icon` prop](https://github.com/Esri/calcite-design-system/issues/4932).


packages/calcite-components/src/components/block/block.scss
```diff
@@ -3,21 +3,42 @@
...
  * These properties can be overridden using the component's tag as selector.
  *
- * @prop --calcite-block-padding: Specifies the padding of the block `default` slot.
+ * @prop --calcite-block-background-color: Specifies the background color of the component.
+ * @prop --calcite-block-border-color: Specifies the border color of the component.
+ * @prop --calcite-block-description-text-color: Specifies the text color of the component's description.
+ * @prop --calcite-block-header-background-color: Specifies the background color of the component's header.
+ * @prop --calcite-block-heading-text-color: Specifies the text color of the component's heading.
+ * @prop --calcite-block-padding: **deprecated** use `--calcite-block-space` instead - Specifies the padding of the component's `default` slot.
+ * @prop --calcite-block-space: Specifies the space of the component's `default` slot.
```
macandcheese, line 12: I think `padding` here is more descriptive - space is a bit ambiguous. Maybe `calcite-block-content-padding` ?

driskull: I think spacing makes more sense. That way we could internally swap between padding/margin/whatever if need be. Do we want to be specific with padding or margin?

driskull: We should decide on space vs spacing too :)

packages/calcite-components/src/components/block/block.scss
```diff
@@ -3,21 +3,42 @@
...
+ * @prop --calcite-block-toggle-icon-color: Specifies the color of the component's toggle icon.
  */
 
+@import "../../assets/styles/animation";
```
driskull, line 16: animation can be removed, its not needed because its already imported in the includes file.

packages/calcite-components/src/components/block/block.scss
macandcheese: Yeah, even `content-space` could make sense I guess. But `space` could be anything - space between text, slots, etc. In Modal we do use `--calcite-modal-content-padding`. In the future "content areas" could be ::parts with defaults.

packages/calcite-components/src/components/block/block.scss
```diff
@@ -3,21 +3,42 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
- * @prop --calcite-block-padding: Specifies the padding of the block `default` slot.
+ * @prop --calcite-block-background-color: Specifies the background color of the component.
+ * @prop --calcite-block-border-color: Specifies the border color of the component.
+ * @prop --calcite-block-description-text-color: Specifies the text color of the component's description.
+ * @prop --calcite-block-header-background-color: Specifies the background color of the component's header.
+ * @prop --calcite-block-heading-text-color: Specifies the text color of the component's heading.
+ * @prop --calcite-block-padding: **deprecated** use `--calcite-block-space` instead - Specifies the padding of the component's `default` slot.
```
macandcheese, line 11: Can you use ` [Deprecated] ` instead here?

packages/calcite-components/src/components/block/block.scss
```diff
@@ -3,21 +3,42 @@
...
+  --calcite-block-toggle-icon-color: var(--calcite-color-text-3);
+
+  --calcite-internal-block-padding-inline: var(
+    --calcite-block-padding,
+    var(--calcite-block-space, theme("spacing[2.5]"))
+  );
+  --calcite-internal-block-padding-block: var(--calcite-block-padding, var(--calcite-block-space, theme("spacing.2")));
+
+  @include base-host();
   @extend %component-spacing;
```
driskull, line 34: This doesn't need to be extended. None of these vars are used and this file is going to go away.

packages/calcite-components/src/components/block/block.scss
```diff
@@ -3,21 +3,42 @@
...
+ * @prop --calcite-block-padding: **deprecated** use `--calcite-block-space` instead - Specifies the padding of the component's `default` slot.
+ * @prop --calcite-block-space: Specifies the space of the component's `default` slot.
+ * @prop --calcite-block-toggle-icon-color: Specifies the color of the component's toggle icon.
  */
 
+@import "../../assets/styles/animation";
+@import "../../assets/styles/header";
+
 :host {
-  @extend %component-host;
```
driskull, line 10: You could use the mixin `@mixin base-host()` instead of this if you want to keep the box-sizing.

packages/calcite-components/src/components/block/block.scss
driskull: We already have spacing vars like `--calcite-spacing-xxs` so it only makes sense to have component spacing vars like `--calcite-block-spacing-` IMO.

driskull: For Panel, we added `calcite-panel-footer-space`

macandcheese: `calcite-block-content-space` works for me. We can update elsewhere we currently use `padding` to follow this. As long as it has a context and isn't just `component-space`.

packages/calcite-components/src/components/block/block.scss
```diff
@@ -3,21 +3,42 @@
...
+ * @prop --calcite-block-description-text-color: Specifies the text color of the component's description.
+ * @prop --calcite-block-header-background-color: Specifies the background color of the component's header.
+ * @prop --calcite-block-heading-text-color: Specifies the text color of the component's heading.
+ * @prop --calcite-block-padding: **deprecated** use `--calcite-block-space` instead - Specifies the padding of the component's `default` slot.
+ * @prop --calcite-block-space: Specifies the space of the component's `default` slot.
+ * @prop --calcite-block-toggle-icon-color: Specifies the color of the component's toggle icon.
  */
 
+@import "../../assets/styles/animation";
+@import "../../assets/styles/header";
```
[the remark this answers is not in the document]
jcfranco: Awesome! I'll pick it up after it's merged.

packages/calcite-components/src/components/block/block.scss
jcfranco: I knew this seemed familiar. We had a similar discussion in https://github.com/Esri/calcite-design-system/pull/8629#discussion_r1463970044 and landed on `--calcite-tab-content-block-padding`. 

> `calcite-block-content-space` works for me. We can update elsewhere we currently use padding to follow this. As long as it has a context and isn't just component-space.

SGTM, but I'd like for @alisonailea to chime on this too. If we needed separate props later on for block/inline values, would we introduce `--calcite-component-space-<block|inline>`? If so, we could add an example of this in our component token guidelines.

> We should decide on space vs spacing too :)

We decided on `space` to align with `size` (vs. `sizing`).

alisonailea: Given that our design system is not necessarily web specific, we need to use platform agnostic naming conventions.   `space` is the preferred terminology here.

jcfranco: For posterity, if we need custom space props that only apply horizontally or vertically, we can use something like `--calcite-component-<part>-space-<x|y>` (see https://github.com/Esri/calcite-design-system/pull/8852#discussion_r1508358308).

PR #8794, opened 2024-02-21
feat(shell): add component tokens

driskull:
**Related Issue:** #7180

## Summary

 * @prop --calcite-shell-background-color: The background color of the component.
 * @prop --calcite-shell-border-color: The border color of the component.
 * @prop --calcite-shell-text-color: The text color of the component.
 * @prop --calcite-shell-tip-spacing: The left and right spacing of the `calcite-tip-manager` when slotted in the component.

### DIFF
--- a/packages/calcite-components/src/components/shell/shell.scss
+++ b/packages/calcite-components/src/components/shell/shell.scss
@@ -3,70 +3,85 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
+ * @prop --calcite-shell-background-color: The background color of the component.
+ * @prop --calcite-shell-border-color: The border color of the component.
+ * @prop --calcite-shell-text-color: The text color of the component.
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/shell/shell.scss
```diff
@@ -3,70 +3,85 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
+ * @prop --calcite-shell-background-color: The background color of the component.
+ * @prop --calcite-shell-border-color: The border color of the component.
+ * @prop --calcite-shell-text-color: The text color of the component.
  * @prop --calcite-shell-tip-spacing: The left and right spacing of the `calcite-tip-manager` when slotted in the component.
```
macandcheese, line 9: I think we should deprecate this one, and move it to Tip Group directly if needed, then deprecate it when Tip Group is deprecated following Carousel work

alisonailea (approved):

Issue #8110, opened 2023-11-02
[Block Section] Remove `status` property, add `icon` property


macandcheese:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Currently, Block Section has a `status` property. There are a few issues here:

1 - When used with `toggle-display="switch"` - there is a visual overlap.

2 - I don't think the component itself should convey status. I think using the component in such a way is a fine use case, but I think associating validity with a "Block Section" component is weird.

![Screenshot 2023-11-02 at 2 43 22 PM](https://github.com/Esri/calcite-design-system/assets/4733155/934b7a18-d7b0-46e0-b960-3b4771eb7410)

Propose we remove the `status` property, and instead add an `icon` property. This would:

1 - Allow the existing use case to continue (via setting an `icon` and using a css variable to style it).
2 - Allow for other use cases of iconography in Block Section besides the couple tied to a `status`

### Acceptance Criteria

Block Section's `status` property is deprecated. Block Section has `icon` property added.

### Relevant Info

 Perhaps a content-start/end slot could be added as we have elsewhere, but I'd do that later and only if requested.

cc @SkyeSeitz @ashetland 

### Which Component

Block Section

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

ashetland:
This brings into question the placement of the `icon`. Is there a reason the current status icons are at the end instead of the start? Either position could be valid, but in general, when there is only one `icon` it is placed at the beginning (like `icon-start`).

ashetland:
Should this also apply to Block? It has a status prop that overrides the icon slot (note this is a slot instead of a prop on this one).

macandcheese:
Yeah, probably, as that use case could be solved with more general slots / props. This one is related https://github.com/Esri/calcite-design-system/issues/4932

asangma:
To my knowledge, this shouldn't cause any major issues.
The intention was to offer something parallel to validation indicators in Stepper.
<img width="229" alt="image" src="https://github.com/Esri/calcite-design-system/assets/12503298/f505915d-4edf-4f62-88fa-f90ce8d02c20">

There are cases where a closed block-section might hide an invalid input.
[...]

PR #8795, opened 2024-02-21
feat(block-section): add component tokens

jcfranco:
**Related Issue:** #7180

## Summary

* `--calcite-block-section-background-color`
* `--calcite-block-section-border-color`
* `--calcite-block-section-heading-text-color`
* `--calcite-block-section-heading-text-color-hover`

**Note**: added internal icon color token for [upcoming `icon` prop](https://github.com/Esri/calcite-design-system/issues/8110).


macandcheese:
> Note: component tokens for status icon color are intentionally skipped since status and icon will be https://github.com/Esri/calcite-design-system/issues/8110.

We might want to just add the icon css property now as that issue actually adds an `icon` property.

jcfranco:
@macandcheese Oh, great catch. I totally misread that issue. 😅 I'll also update `block` since I referenced the same issue, and https://github.com/Esri/calcite-design-system/issues/4932 suggests the same changes.

alisonailea (approved):

PR #8796, opened 2024-02-21
feat(label): add component tokens

jcfranco:
**Related Issue:** #7180 

## Summary

 * `--calcite-label-space-y-end`
 * `--calcite-label-text-color`

Deprecates the following:

* `--calcite-label-margin-bottom`


macandcheese (approved):
Looks good pending alignment with resolution of https://github.com/Esri/calcite-design-system/pull/8790 nomenclature discussion for `margin-bottom` / `space-block-end`

packages/calcite-components/src/components/alert/alert.scss
```diff
@@ -4,19 +4,50 @@
  * These properties can be overridden using the component's tag as selector.
  *
  * @prop --calcite-alert-width: Specifies the width of the component.
+ * @prop --calcite-alert-background-color: Specifies the background color of the component.
```
alisonailea, line 7: I think this file should be removed from this PR.

alisonailea (changes requested):

packages/calcite-components/src/components/alert/alert.scss
jcfranco: Thanks for the catch. 🎣 Fixing!

alisonailea (approved):

jcfranco:
**Note:** if a different base font size was previously used, there will be a difference in spacing between elements in labels. This is because the gap is now using px and no longer rem-based.

Issue #8571, opened 2024-01-08
Optimize `transition-default` Tailwind util to not target all properties


jcfranco:
### Description

Our current [`transition-default`](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/calcite-preset.ts#L280) Tailwind utility applies transitions to all properties, leading to potential performance impact. We should revisit the util to target specific properties.

As a first step, we could look into targeting a reduced set of essential/common transition properties.

### Proposed Advantages

Improved performance by having browser transition relevant, or common, CSS properties. 

### Which Component

[All components using `transition-default`](https://github.com/search?q=repo%3AEsri%2Fcalcite-design-system+%22transition-default%22+language%3ASCSS&type=code&l=SCSS):

* `alert`
* `checkbox`
* `chip`
* `date-picker-day`
* `date-picker-month-header`
* `inline-editable`
* `input-message`
* `input-number`
* `input-text`
* `input`
* `link`
* `modal`
* `navigation-logo`
* `navigation-user`
* `notice`
* `pagination`
* `radio-button`
* `rating`
* `shell-panel`
* `split-button`
* `tab-title`
* `tile-select`
* `tree-item`
* `x-button` (functional component)

### Relevant Info

<details>
<summary>List of all animatable CSS properties</summary>

<br>See https://vallek.github.io/animatable-css/#anim

1. **Colors**
   - `color`
   - `background-color`
   - `border-color`
   - `outline-color`
   - `fill`
   - `stroke`

2. **Dimensions**
   - `width`
   - `height`
   - `max-width`
   - `max-height`
   - `min-width`
   - `min-height`

3. **Margins, Padding, and Borders**
   - `margin`
   - `margin-top`
   - `margin-right`
   - `margin-bottom`
   - `margin-left`
   - `padding`
   - `padding-top`
   - `padding-right`
   - `padding-bottom`
   - `padding-left`
   - `border-width`
   - `border-top-width`
   - `border-right-width`
   - `border-bottom-width`
   - `border-left-width`
   - `border-spacing`

4. **Positioning**
   - `top`
   - `right`
   - `bottom`
   - `left`
   - `z-index`

5. **Backgrounds**
   - `background-size`
   - `background-position`

6. **Fonts and Text**
   - `font-size`
   - `font-weight`
   - `line-height`
   - `letter-spacing`
   - `text-indent`
   - `text-shadow`
   - `word-spacing`

7. **Transforms**
[...]

PR #8797, opened 2024-02-21
perf: update `transition-default` Tailwind util to only target common, animatable properties

jcfranco:
**Related Issue:** #8571 

## Summary

This updates the `transition-default` to explicitly list common, animatable properties to the following:

* `background-color`
* `block-size`
* `border-color`
* `box-shadow`
* `color`
* `inset-block-end`
* `inset-block-start`
* `inset-inline-end`
* `inset-inline-start`
* `inset-size`
* `opacity`
* `outline-color`
* `transform`

This should improve rendering performance by omitting other props that are unintentionally being animated (see https://vallek.github.io/animatable-css/#anim).


alisonailea:
For added improved performance would it make sense to make this a mixin and allow the animated props to be optionally passed in via an SCSS prop? That way, if they are known, a dev can use the shared code but only set animation for "background-color" if that's all that's needed?

jcfranco:
@alisonailea I like that. I'd prefer being explicit and only set up transitions if they are intended in the design. That could be the next phase.

driskull (approved):
👍

PR #8806, opened 2024-02-22
feat(tile-group): add Tile Group component

eriklharper:
**Related Issues:** #8615 #6691 #6662

## Summary

This PR adds the new `calcite-tile-group` component.  It includes these responsiveness features that address #6691:

- Wrapped Tiles match width of Tiles above
- Wrapped Tiles match height of tallest Tile in the group

The changes in this PR were extracted from and depend on the changes in #8681.


packages/calcite-components/src/components/tile-group/resources.ts
```diff
@@ -0,0 +1,3 @@
+export const CSS = {
+  tileGroup: "tile-group",
```
jcfranco, line 2: For consistency, use `container` for the root element class (key and value).

packages/calcite-components/src/components/tile-group/tile-group.scss
```diff
@@ -0,0 +1,40 @@
...
+
+  ::slotted(calcite-tile) {
+    margin-block-end: var(--calcite-spacing-px);
+    margin-inline-end: var(--calcite-spacing-px);
+  }
+}
+
+:host([scale="s"]) {
+  .tile-group {
+    grid-template-columns: repeat(auto-fit, minmax(100px, 1fr));
```
jcfranco, line 18: This could be simplified by using an internal CSS prop for the `min` parameter.

packages/calcite-components/src/components/tile-group/tile-group.stories.ts
```diff
@@ -0,0 +1,964 @@
...
+export default {
+  title: "Components/Tiles/Tile Group",
+  parameters: {
+    notes: readme,
+    chromatic: { delay: 10000 },
+  },
+  ...storyFilters(),
+};
+
+export const playground = (): string => html`
```
jcfranco, line 15: For consistency, let's name this `simple`.

packages/calcite-components/src/components/tile-group/tile-group.stories.ts
```diff
@@ -0,0 +1,964 @@
...
+    ></calcite-tile>
+    <calcite-tile
+      heading="Tile heading lorem ipsum"
+      description="Iterative approaches to corporate strategy foster collab."
+      icon="layers"
+    ></calcite-tile>
+  </calcite-tile-group>
+`;
+
+export const allVariants_NoTest = (): string => html`
```
jcfranco, line 45: Can you create a follow-up issue to include this in screenshot tests once we've addressed the `timeout` issue?

packages/calcite-components/src/components/tile-group/tile-group.scss
```diff
@@ -0,0 +1,40 @@
+:host {
+  box-sizing: border-box;
+  display: inline-block;
+
+  .tile-group {
```
driskull, line 5: I don't think these need to be nested under host selector

packages/calcite-components/src/components/tile-group/tile-group.tsx
```diff
@@ -0,0 +1,94 @@
...
+  @Element() el: HTMLCalciteTileGroupElement;
+
+  //--------------------------------------------------------------------------
+  //
+  //  Private Methods
+  //
+  //--------------------------------------------------------------------------
+
+  private updateTiles = (): void => {
+    this.el.querySelectorAll("calcite-tile").forEach((item) => (item.scale = this.scale));
```
driskull, line 63: Maybe for future but can we get the slotted assigned elements instead of querying the dom?

driskull (approved):
Should an entry be added to the stencil configuration as well?

Looks great 👍

jcfranco:
> Should an entry be added to the stencil configuration as well?

Agreed. Good catch!

packages/calcite-components/src/components/tile-group/tile-group.stories.ts
eriklharper: https://github.com/Esri/calcite-design-system/issues/8829

PR #8813, opened 2024-02-22
fix(link): update component tokens to pass E2E tests

alisonailea:
**Related Issue:** #7180

## Summary

Make some fixes to the link component tokens to pass E2E tests. Also removes an extraneous unused class from the shadow DOM of the link component.


driskull (approved):
👍

PR #8816, opened 2024-02-22
test(option-group): fix

alisonailea:
## Summary

The previous selector included the component’s style tag.


jcfranco (commented):
I'm confused, this test doesn't seem to be failing on `main` and the before and after achieves pretty much the same. IMO, the latter is more brittle since it relies on an element selector.

PR #8826, opened 2024-02-23
Refactor components for fallback + state token approach

alisonailea:
--Draft--

This is a proof of concept intended to document the work needed in various components but is not intended to be merged.

### DIFF
--- a/packages/calcite-components/src/components/card/card.scss
+++ b/packages/calcite-components/src/components/card/card.scss
@@ -7,6 +7,20 @@
  * @prop --calcite-card-border-color: Specifies the border color of the component.
  * @prop --calcite-card-box-shadow: Specifies the box shadow of the component.
  * @prop --calcite-card-corner-radius: Specifies the corner radius of the component.
+ * @prop --calcite-card-label-text-color: defines the label text colorin the component.
+ * @prop --calcite-card-checkbox-background-color-checked: defines the checkbox background color when checked in the component.
+ * @prop --calcite-card-checkbox-background-color: defines the checkbox background color in the component.
+ * @prop --calcite-card-checkbox-icon-color: defines the checkbox icon color in the component.
+ * @prop --calcite-card-checkbox-shadow-checked: defines the checkbox shadow when checked in the component.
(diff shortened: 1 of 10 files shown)


packages/calcite-components/src/components/card/card.scss
```diff
@@ -7,6 +7,20 @@
  * @prop --calcite-card-border-color: Specifies the border color of the component.
  * @prop --calcite-card-box-shadow: Specifies the box shadow of the component.
  * @prop --calcite-card-corner-radius: Specifies the corner radius of the component.
+ * @prop --calcite-card-label-text-color: defines the label text colorin the component.
+ * @prop --calcite-card-checkbox-background-color-checked: defines the checkbox background color when checked in the component.
+ * @prop --calcite-card-checkbox-background-color: defines the checkbox background color in the component.
+ * @prop --calcite-card-checkbox-icon-color: defines the checkbox icon color in the component.
+ * @prop --calcite-card-checkbox-shadow-checked: defines the checkbox shadow when checked in the component.
+ * @prop --calcite-card-checkbox-shadow-hover: defines the checkbox shadow when hovered in the component.
+ * @prop --calcite-card-checkbox-shadow-invalid: defines the checkbox shadow when invalid in the component.
```
macandcheese, line 16: These checkbox ones should be introduced as deprecated as possible as they are no longer used once Card Group lands. That PR has tokens for the replacement. 

The checkbox also cannot be set to invalid or have a label currently as it is rendered inside the component and a user cannot set invalid on either the checkbox or the card.

macandcheese (changes requested):

packages/calcite-components/src/components/card/card.scss
```diff
@@ -7,6 +7,20 @@
...
+ * @prop --calcite-card-checkbox-shadow: defines the checkbox shadow in the component.
+ * @prop --calcite-card-checkbox-size: defines the checkbox size in the component.
+ * @prop --calcite-card-loader-text-color: defines the loader text color in the component.
```
macandcheese, line 19: This cannot be set on Card so it shouldn’t be passed through.

PR #8828, opened 2024-02-23
refactor(tile-group): add review suggestions

eriklharper:
**Related Issue:** #8806

## Summary

This is a follow-up for #8806, addressing comments in the review.  No changes to functionality.


packages/calcite-components/stencil.config.ts
```diff
@@ -77,7 +77,7 @@ export const create: () => Config = () => ({
     { components: ["calcite-switch"] },
     { components: ["calcite-tab", "calcite-tab-title", "calcite-tab-nav", "calcite-tabs"] },
     { components: ["calcite-text-area"] },
-    { components: ["calcite-tile"] },
+    { components: ["calcite-tile-group", "calcite-tile"] },
```
jcfranco, line 80: I think these need to be separate since tile can still be used standalone, right?

jcfranco (approved):
Awesome! This should be good to merge once the bundle entry is updated. 🚀

packages/calcite-components/stencil.config.ts
eriklharper: k just updated, they are separate now.

Issue #8482, opened 2023-12-21
Navigation Logo - using wrong icon color with href


jgrayson-apl:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

~Related Issue #1: The icon color of the calcite-navigation-logo component changes to reflect link-like colors when a href attribute is present.~ Out of scope - refer to https://github.com/Esri/calcite-design-system/issues/8482#issuecomment-1887940278
Related Issue #2: the calcite-navigation-logo component still uses --calcite-ui-icon-color

We don't want the icon color to change just because we add a href attribute... it completely goes against what what we expect.  Additionally, we can't use the new --calcite-color-icon-color to override as it still uses the older CSS variable.

### Expected Behavior

I would expect the calcite-navigation-logo to not change the icon color just because we add a href attribute, and if that is the 'correct' undesired behavior, then at least I'd like to use the newer CSS variables to fix do this.

### Reproduction Sample

https://codepen.io/john-grayson/pen/XWGWXLZ

### Reproduction Steps

1) remove href attribute from calcite-navigation-logo to see how icon should look
2) uncomment css override to see how it still uses --calcite-ui-icon-color

### Reproduction Version

2.0.0

### Priority impact

p4 - not time sensitive

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

geospatialem:
Updated the request above, as the `--calcite-ui-icon-color` css var/token will remain. More details below:

> The icon color of the calcite-navigation-logo component changes to reflect link-like colors when a href attribute is present.

Will work towards addressing in a future milestone. The estimate is based solely on this update.

> the calcite-navigation-logo component still uses --calcite-ui-icon-color

Working towards leaving as-is, as the above in alignment with design tokens and is out of scope for the issue.

geospatialem:
Verified in `2.6.0-next.1`.

PR #8830, opened 2024-02-23
fix(navigation-logo): no longer changes icon color when `href` is parsed

anveshmekala:
**Related Issue:** #8482 

## Summary

Fixes change of icon color in `navigation-logo` when `href` attribute is parsed.


driskull (approved):
👍

Issue #8658, opened 2024-01-26
[Radio Button] remove deprecated `--calcite-radio-size` token


eriklharper:
**monday.com sync:** #18048397861

Remove the deprecated `--calcite-radio-size` token.

### Relevant Info
Deprecated as part of #7180 via https://github.com/Esri/calcite-design-system/pull/11891.

Old context:
> This change can be traced back to this commit all the way back in 2021: https://github.com/Esri/calcite-design-system/pull/1696/files
>
> https://codepen.io/eriklharper/pen/XWGzVgB

jcfranco:
@eriklharper Due to the undocumented nature of the CSS prop, can you update this issue's title and labels to focus more on refactoring vs bug?

jcfranco:
Due to the variable name and uncertainty about its current use, we're considering this as a potential breaking change for 4.0.

driskull:
@eriklharper @jcfranco @DitwanP @geospatialem @alisonailea what should the size token(s) for radio buttons be?

Should `--calcite-radio-size` become:

- `--calcite-radio-size-x` / `--calcite-radio-size-y`?
- `--calcite-radio-width` / `--calcite-radio-height`?
- Or should these just be removed?

eriklharper:
> [@eriklharper](https://github.com/eriklharper) [@jcfranco](https://github.com/jcfranco) [@DitwanP](https://github.com/DitwanP) [@geospatialem](https://github.com/geospatialem) [@alisonailea](https://github.com/alisonailea) what should the size token(s) for radio buttons be?
> 
> Should `--calcite-radio-size` become:
> 
> * `--calcite-radio-size-x` / `--calcite-radio-size-y`?
> * `--calcite-radio-width` / `--calcite-radio-height`?
> * Or should these just be removed?

They should just be removed because the size is controlled internally by the `scale` property, and not meant to be further customized outside of setting the scale.  @ashetland correct?

driskull:
Should we just create `--calcite-radio-button-size` and deprecate `--calcite-radio-size`?

eriklharper:
> I believe that's correct, but to add a question into the mix, the Checkbox component has a similar css prop, but in that case it **is documented**. Is it possible that these should be deprecated and removed from both components? cc [@macandcheese](https://github.com/macandcheese)

[...]

PR #8831, opened 2024-02-26
feat(radio-button, radio-button-group): add component tokens

driskull:
**Related Issue:** #7180

## Summary

WIP

### DIFF
--- a/packages/calcite-components/src/components/radio-button-group/radio-button-group.scss
+++ b/packages/calcite-components/src/components/radio-button-group/radio-button-group.scss
@@ -1,42 +1,58 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-radio-button-group-input-message-spacing-value: Specifies the spacing of the input message.
+ */
+
 :host {
-  @apply flex flex-col;
+  display: flex;
+  flex-direction: column;
 }
 
 :host > .item-wrapper {
-  @apply flex;
+  display: flex;
   max-inline-size: 100vw;
 }
 
 :host([layout="horizontal"]) > .item-wrapper {
-  @apply flex-row flex-wrap;
+  flex-direction: row;
+  flex-wrap: wrap;
 }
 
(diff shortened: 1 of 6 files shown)


packages/calcite-components/src/components/radio-button-group/radio-button-group.scss
```diff
@@ -1,42 +1,58 @@
...
 }
 
 :host([layout="vertical"]) > .item-wrapper {
-  @apply flex-col;
+  flex-direction: column;
 }
 
 :host([scale="s"]) calcite-input-message {
-  --calcite-input-message-spacing-value: calc(var(--calcite-spacing-xxs) * -1);
+  --calcite-input-message-spacing-value: calc(
```
alisonailea, line 41: move this to an `-internal-` var

alisonailea (changes requested):

packages/calcite-components/src/components/radio-button-group/radio-button-group.scss
driskull: @alisonailea `--calcite-radio-button-group-input-message-space` should be internal? Then a user wouldn't be able to set the spacing for the input-message component.

macandcheese: I don't know why we'd want folks to adjust the distance between the input message and the control. Are there examples where we see this?

driskull: I only added this because its a var on the input-message.

macandcheese: Ah. Yeah maybe we can deprecate it there if it isn't. A standalone instance of the component can just be spaced with css on the element.

alisonailea (approved):

alisonailea (approved):

eriklharper:
Once the main tokens branch is merged to `main` we should be able to mark this issue resolved: https://github.com/Esri/calcite-design-system/issues/8658

PR #8839, opened 2024-02-27
fix(card): Do not apply text color to slotted footer items

macandcheese:
**Related Issue:** #

## Summary
This removes the styles set on footer elements, as it could unintentionally style components slotted directly that themselves do not ensure a text color is set. The workaround is to add a wrapping div that itself is slotted in `footer-end` or `footer-start`.

Screenshot 2024-02-27 at 12 01 36 PM

The `heading` and `description` (and deprecated `title` and `subtitle`) should keep their default text styles as these are most often text slotted directly. In the future it might be good to make these string properties to incentivize consistency, keeping the option to slot in rich content like links, icons, etc., when needed.

### DIFF
--- a/packages/calcite-components/src/components/card/card.scss
+++ b/packages/calcite-components/src/components/card/card.scss
@@ -153,12 +153,10 @@
 @include slotted("footer-start", "*") {
   @apply text-n2-wrap self-center;
   margin-inline-end: auto;
-  color: var(--calcite-color-text-3);
 }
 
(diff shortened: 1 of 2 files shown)


jcfranco (approved):
LGTM!

Before merging, can you create an associated issue and link it for the verification step?

macandcheese:
cc @jcfranco  https://github.com/Esri/calcite-design-system/issues/8855

PR #8844, opened 2024-02-27
refactor(flow, flow-item, panel, popover, tooltip): Use fallback method for css vars

driskull:
**Related Issue:** #7180

## Summary

Update installed tokens for the components above to use fallback method.

### DIFF
--- a/packages/calcite-components/src/assets/styles/_floating-ui.scss
+++ b/packages/calcite-components/src/assets/styles/_floating-ui.scss
@@ -10,12 +10,12 @@ $floating-ui-transform-right: translateX(-5px);
     transition-property: transform, visibility, opacity;
     opacity: 0;
     border-style: solid;
-    border-width: var(#{$borderWidth});
-    box-shadow: var(#{$boxShadow});
-    border-radius: var(#{$borderRadius});
+    border-width: #{$borderWidth};
+    box-shadow: #{$boxShadow};
+    border-radius: #{$borderRadius};
     z-index: var(--calcite-z-index);
-    background-color: var(#{$backgroundColor});
-    border-color: var(#{$borderColor});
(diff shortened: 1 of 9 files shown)


driskull:
@alisonailea this uses the fallback var method for PRs I had already installed. (flow, flow-item, panel, tooltip, popover).

packages/calcite-components/src/assets/styles/_floating-ui.scss
```diff
@@ -26,11 +26,11 @@ $floating-ui-transform-right: translateX(-5px);
...
-  $boxShadow: "--calcite-shadow-md",
-  $borderRadius: "--calcite-corner-radius-0",
-  $borderWidth: "--calcite-border-width-none",
-  $borderColor: "--calcite-color-border-3",
-  $backgroundColor: "--calcite-color-foreground-1"
+  $boxShadow: "var(--calcite-shadow-md)",
+  $borderRadius: "var(--calcite-corner-radius-0)",
+  $borderWidth: "var(--calcite-border-width-none)",
+  $borderColor: "var(--calcite-color-border-3)",
+  $backgroundColor: "var(--calcite-color-foreground-1)"
```
alisonailea, line 33: should this be refactored for tokens? Something like `--calcite-floating-el-[shadow, corner-radius, border-color, background-color]`?

alisonailea (approved):

packages/calcite-components/src/assets/styles/_floating-ui.scss
driskull: These are just the default values, components using this mixin would pass in their tokens here.

`--calcite-floating-el` tokens don't make sense because its not actually a component.

PR #8846, opened 2024-02-28
feat(switch): add stateful component tokens

jcfranco:
**Related Issue:** #7180 

## Summary

Updates component tokens to the following (CSS props):

* `--calcite-switch-corner-radius`
* `--calcite-switch-handle-background-color`
* `--calcite-switch-handle-border-color`
* `--calcite-switch-handle-shadow`
* `--calcite-switch-shadow-hover`
* `--calcite-switch-track-background-color`
* `--calcite-switch-track-border-color`

### DIFF
--- a/packages/calcite-components/src/components/switch/switch.scss
+++ b/packages/calcite-components/src/components/switch/switch.scss
@@ -19,23 +19,6 @@
  */
 
 :host {
-  --calcite-switch-handle-background-color: var(--calcite-color-foreground-1);
-  --calcite-switch-handle-border-color: var(--calcite-color-border-input);
-  --calcite-switch-track-background-color: var(--calcite-color-foreground-2);
-  --calcite-switch-track-border-color: var(--calcite-color-border-2);
-  --calcite-switch-corner-radius: var(--calcite-internal-switch-border-radius, --calcite-corner-radius-pill);
-  --calcite-switch-handle-shadow: none;
-
-  --calcite-switch-handle-border-color-hover: var(--calcite-color-brand);
-  --calcite-switch-handle-shadow-hover: inset 0 0 0 1px var(--calcite-color-brand);
-
-  --calcite-switch-handle-border-color-selected: var(--calcite-color-brand);
-  --calcite-switch-track-background-color-selected: var(--calcite-color-brand);
-  --calcite-switch-track-border-color-selected: var(--calcite-color-brand);
-
-  --calcite-switch-handle-border-color-selected-hover: var(--calcite-color-brand-hover);
-  --calcite-switch-handle-shadow-selected-hover: inset 0 0 0 1px var(--calcite-color-brand-hover);
-
   --calcite-internal-switch-border-radius: 9999px; // TODO: drop once --calcite-corner-radius is updated
 
   @apply relative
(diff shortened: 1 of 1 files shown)


packages/calcite-components/src/components/switch/switch.scss
```diff
@@ -3,23 +3,38 @@
...
+ * @prop --calcite-switch-corner-radius: Specifies the corner radius of the component.
  * @prop --calcite-switch-handle-background-color: Specifies the background color of the component handle.
  * @prop --calcite-switch-handle-border-color: Specifies the border color of the component handle.
  * @prop --calcite-switch-track-background-color: Specifies the background color of the component track.
  * @prop --calcite-switch-track-border-color: Specifies the border color of the component track.
- * @prop --calcite-switch-corner-radius: Specifies the corner radius of the component.
- * @prop --calcite-switch-shadow: Specifies the shadow of the component.
+ * @prop --calcite-switch-handle-shadow: Specifies the shadow of the component handle.
+ * @prop --calcite-switch-handle-border-color-hover: Specifies the border color of the component handle when hovered.
+ * @prop --calcite-switch-shadow-hover: Specifies the shadow of the component handle when hovered.
```
macandcheese, line 13: Should these be set by the user, for instance `calcite-switch:hover { --calcite-switch-handle-border-color: purple }`? I think the hover/active/stateful props are only for inaccessible components within the shadowRoot. Unless I missed a convention change (which.. very possible).

jcfranco: I meant to reply sooner, but this is related to a recent guideline update:

> component tokens should be added for the appropriate states as well. I.E. if background-color changes on hover then you should also add a `background-color-hover` token.

macandcheese: Gotcha. In that case, the css prop makes sense - but maybe the description is confusing? Really it's changing the handle-border-color when the entire switch is hovered, not only when the handle is hovered - maybe "... of the component's handle when the component is hovered" ? Kind of verbose, but more accurate.

packages/calcite-components/src/components/switch/switch.scss
```diff
@@ -3,24 +3,22 @@
...
+ * @prop --calcite-switch-handle-border-color-selected: Specifies the border color of the component handle when selected.
+ * @prop --calcite-switch-track-background-color-selected: Specifies the background color of the component track when selected.
```
macandcheese, line 15: General q - should we use selected here or `checked` to match the prop?

packages/calcite-components/src/components/switch/switch.scss
jcfranco: Oh, good catch. I'll double check prop descriptions. There might be some copy pasta afoot. ©️🍝🦶

packages/calcite-components/src/components/switch/switch.scss
jcfranco: Good question. I've been going back and forth on `selected` (consistency) vs `checked` (correctness). I'll update it to `checked` per your suggestion.

packages/calcite-components/src/components/switch/switch.scss
jcfranco: Updated prop descriptions. We'll need to audit the doc once we're all components are updated for clarity and consistency too. cc @geospatialem @DitwanP

alisonailea (approved):

alisonailea (approved):
Nice

PR #8849, opened 2024-02-28
refactor(card-group): Remove unused `scale` property

macandcheese:
**Related Issue:** #

## Summary
Removes this unused property that made it through review. 

Initially there was a plan to add scale to Card, but that change was removed and this property was not cleaned up. Should be safe to remove as it had no effect.

### DIFF
--- a/packages/calcite-components/src/components.d.ts
+++ b/packages/calcite-components/src/components.d.ts
@@ -806,10 +806,6 @@ export namespace Components {
          */
         "label": string;
         /**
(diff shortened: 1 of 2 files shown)


driskull (approved):
👍 

This is a new component so no deprecation needed right?

macandcheese:
> 👍
> 
> This is a new component so no deprecation needed right?

Yep, just went out in 2.6.0 - this just wasn't caught / removed from that PR.

PR #8850, opened 2024-02-29
test(select): fix form-associated E2E test

jcfranco:
**Related Issue:** N/A

## Summary

Uses character-matching behavior for select to ensure the component's value changes. The context-menu is not accessible through Puppeteer.


driskull (approved):
👍

Issue #8800, opened 2024-02-21
Replace animated selected tab-title indicator with simpler selection styling


jcfranco:
### Description

Simplification of tab-title was [suggested in #8616](https://github.com/Esri/calcite-design-system/pull/8616#issuecomment-1942738868):

> I suggest we get rid of the "zoom border animation", it is adding a bit of confusion to a crowded area (IMO). Just render the active state visual with a border or inset box-shadow.

### Proposed Advantages

* reduce component complexity
* avoid alignment issues
* streamline component design

### Which Component

`calcite-tab-nav`

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

DitwanP:
🍡 Verified on `2.11.0-next.23`

PR #8854, opened 2024-02-29
feat(tab-title): add component tokens

jcfranco:
**Related Issue:** #7180

## Summary

* `--calcite-tab-title-background-color`
* `--calcite-tab-title-text-color`
* `--calcite-tab-title-close-button-background-color`
* `--calcite-tab-title-close-button-background-color-active`
* `--calcite-tab-title-close-button-background-color-focus`
* `--calcite-tab-title-close-button-background-color-hover`
* `--calcite-tab-title-close-button-icon-color`
* `--calcite-tab-title-close-button-icon-color-active`
* `--calcite-tab-title-close-button-icon-color-focus`
* `--calcite-tab-title-close-button-icon-color-hover`

**Note:** Removes styles targeting `a` as these are no longer applicable.


alisonailea (approved):

packages/calcite-components/src/components/tab-title/tab-title.scss
```diff
@@ -1,3 +1,25 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-tab-title-background-color: Specifies the background color of the component.
+ * @prop --calcite-tab-title-background-color-hover: Specifies the background color of the component when hovered.
+ * @prop --calcite-tab-title-border-color-selected: Specifies the border color of the component. Only applicable when bordered.
```
macandcheese, line 8: I think this one is a little confusing. I'd interpret this to mean the "accent" or "highlight" color. But it seems like it would just be the two "side borders"? Maybe this can be set by the tab nav if that is the case.

I'm working on #8800 which would make this useful - currently the selection element is not part of Tab Title, but it will when the above fix lands.

jcfranco: > I think this one is a little confusing. I'd interpret this to mean the "accent" or "highlight" color. But it seems like it would just be the two "side borders"? Maybe this can be set by the tab nav if that is the case.

Agreed. Let me see if I can move this up to tab nav and simplify the props here.

> I'm working on #8800 which would make this useful - currently the selection element is not part of Tab Title, but it will when the above fix lands.

Do you estimate this landing in this release? I could proceed with the current props/styles and revisit once the selection indicator is installed.

macandcheese: I think you could still add the highlight css prop and in my PR I can change the element it affects from the "indicator div" to just the border, and keep the name the same.

alisonailea (changes requested):
do we need a box-shadow token for tab-title?

packages/calcite-components/src/components/tab-title/tab-title.scss
```diff
@@ -209,16 +198,18 @@
 }
 
 :host([bordered][selected]) {
-  box-shadow: inset 0 -2px var(--calcite-tab-title-border-color-selected, var(--calcite-color-foreground-1));
+  // we pick this up from tabs, where border is defined and applied
+  box-shadow: inset -1px -1px var(--calcite-internal-tabs-border-color);
```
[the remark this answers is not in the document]
jcfranco: I don't think a `box-shadow` prop is needed. `box-shadow` is used to mask a border when a tab is selected (see [example](https://codepen.io/jcfranco/pen/PogGRzo?editors=100)).

alisonailea (approved):

PR #8876, opened 2024-03-04
refactor(chip): Update token pattern

macandcheese:
**Related Issue:** #7180 

## Summary
Updates token implementation to match latest recommended pattern.

### DIFF
--- a/packages/calcite-components/src/components/chip/chip.scss
+++ b/packages/calcite-components/src/components/chip/chip.scss
@@ -79,7 +79,9 @@
   &.is-circle {
     inline-size: var(--calcite-internal-chip-container-size);
   }
-
+  &:not(.is-circle) {
+    padding-inline: var(--calcite-internal-chip-spacing-s);
+  }
   &.selectable {
     @apply cursor-pointer;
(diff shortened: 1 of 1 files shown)


packages/calcite-components/src/components/chip/chip.scss
```diff
@@ -71,97 +60,111 @@
   --calcite-internal-chip-container-size: 44px;
 }
 
-:host([appearance="outline-fill"]) {
-  --calcite-chip-background-color: var(--calcite-color-foreground-1);
+.container {
+  @apply inline-flex
+  h-full
+  max-w-full
```
alisonailea, line 66: h-full and max-w-full can be changed to 
`block-size: var(--calcite-container-size-content-fluid)`
`max-inline-size: var(--calcite-container-size-content-fluid)`

alisonailea (approved):

packages/calcite-components/src/components/chip/chip.scss
macandcheese: Updated

PR #8878, opened 2024-03-04
feat(accordion, accordion-item): add component tokens

alisonailea:
**Related Issue:** #7180

### Accordion
--calcite-accordion-background-color: defines the background color for the component and it's slotted accordion-items.
--calcite-accordion-border-color: defines the border color for the component and it's slotted accordion-items.
 
### Accordion Item
--calcite-accordion-item-background-color: defines the background color for the component.
--calcite-accordion-item-border-color: defines the border color for the component.
--calcite-accordion-item-description-text-color: defines the text color of the description in the component. Falls back to `--calcite-accordion-item-text-color`.
--calcite-accordion-item-heading-text-color: defines the text color of the heading in the component. Falls back to `--calcite-accordion-item-text-color`.
--calcite-accordion-item-icon-color: defines the icon color for the component.
--calcite-accordion-item-expand-icon-color: defines the icon color for the component.
--calcite-accordion-item-text-color: defines the text color for the component.


packages/calcite-components/src/components/accordion-item/accordion-item.scss
```diff
@@ -63,38 +77,42 @@
...
 }
 
 .header-container {
-  inline-size: 100%;
+  inline-size: var(--calcite-container-size-content-fluid);
 }
 
 .content {
-  padding: var(--calcite-accordion-item-padding);
+  padding: var(--calcite-internal-accordion-item-space-inset);
```
jcfranco, line 109: Nitpick since this is internal, but shouldn't this be `--calcite-internal-accordion-item-content-space`? 

Applies to other inset space CSS props.

packages/calcite-components/src/components/accordion-item/accordion-item.scss
```diff
@@ -106,29 +124,27 @@
...
   text-align: initial;
 }
 
 // accordion item icon
 :host .expand-icon {
-  @apply text-color-3;
-  margin-inline-start: var(--calcite-accordion-item-icon-spacing-start);
-  margin-inline-end: var(--calcite-accordion-item-icon-spacing-end);
-  transform: rotate(var(--calcite-accordion-item-icon-rotation));
+  margin-inline-start: var(--calcite-internal-accordion-item-icon-spacing-start);
```
jcfranco, line 133: Nitpick since this is internal, but for consistency, these should use `space` vs `spacing`.

Also, since these are used for inline spacing, would `--calcite-internal-accordion-item-icon-space-x-start/end` be more appropriate?

packages/calcite-components/src/components/accordion-item/accordion-item.scss
```diff
@@ -140,51 +156,64 @@
...
 
 :host(:focus),
-:host(:hover) {
+:host(:hover),
+:host(:active),
+:host([expanded]) {
   & .heading {
-    @apply text-color-1;
+    color: var(
+      --calcite-accordion-item-heading-text-color,
```
jcfranco, line 184: Per our latest guidelines, should these be stateful CSS props?

> component tokens should be added for the appropriate states as well. I.E. if background-color changes on hover then you should also add a `background-color-hover` token. 

Applies to other CSS props affected by state.

packages/calcite-components/src/components/accordion-item/accordion-item.scss
```diff
@@ -1,40 +1,54 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-accordion-item-background-color: defines the background color for the component.
+ * @prop --calcite-accordion-item-border-color: defines the border color for the component.
+ * @prop --calcite-accordion-item-description-text-color: defines the text color of the description in the component. Falls back to `--calcite-accordion-item-text-color`.
```
jcfranco, line 8: For the time being, can we omit specifying which CSS prop it will fall back to? I think we could do so since it's an implementation detail.

packages/calcite-components/src/components/accordion/accordion.scss
```diff
@@ -1,41 +1,64 @@
...
+  }
+}
+
+.accordion {
+  border-width: var(--calcite-border-width-sm);
+  border-color: var(--calcite-accordion-border-color, var(--calcite-color-border-2));
+  border-block-end-width: var(--calcite-border-width-none);
+  border-style: solid;
+
+  ::slotted(accordion-item) {
```
jcfranco, line 29: If accordion items have these CSS props defined, shouldn't users set overrides on the items themselves since these are slotted and not owned by the accordion?

packages/calcite-components/src/components/accordion-item/accordion-item.scss
```diff
@@ -1,40 +1,54 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-accordion-item-background-color: defines the background color for the component.
+ * @prop --calcite-accordion-item-border-color: defines the border color for the component.
```
jcfranco, line 7: `--calcite-accordion-item-border-color` doesn't seem to be used in `accordion-item`'s styling.

jcfranco (changes requested):
Looking good! 😎 I noticed a few things and also had a question regarding stateful CSS props.

packages/calcite-components/src/components/accordion/accordion.scss
```diff
@@ -1,41 +1,64 @@
...
+    --calcite-accordion-item-background-color: var(
+      --calcite-accordion-background-color,
+      var(--calcite-color-foreground-1)
+    );
+  }
+}
+
+.accordion--transparent {
+  ::slotted(accordion-item) {
+    --calcite-accordion-item-border-color: var(--calcite-accordion-border-color, transparent);
```
jcfranco, line 40: We should add a note to our guidelines about composite component overrides to slotted components.

packages/calcite-components/src/components/accordion-item/accordion-item.scss
alisonailea: because all parts of the component respond to the state of the host, we don't need state tokens here.

packages/calcite-components/src/components/accordion/accordion.scss
alisonailea: nope! Thanks for the catch.

packages/calcite-components/src/components/accordion-item/accordion-item.scss
alisonailea: thanks for the catch!

PR #8879, opened 2024-03-04
refactor(progress): update token pattern

Elijbet:
**Related Issue:** #7180

## Summary
Update component tokens with fallbacks and add stories.

### DIFF
--- a/packages/calcite-components/src/components/input/input.scss
+++ b/packages/calcite-components/src/components/input/input.scss
@@ -18,10 +18,18 @@
  * @prop --calcite-input-placeholder-text-color: defines the color of placeholder text in the component.
  * @prop --calcite-input-prefix-background-color: defines the background color of the prefix sub-component.
  * @prop --calcite-input-prefix-text-color:  defines the text color of the prefix sub-component.
+ * @prop --calcite-progress-background-color: Defines the background color of the progress sub-component.
+ * @prop --calcite-progress-fill-color: Defines the background color of the progress sub-component bar.
+ * @prop --calcite-progress-text-color: Defines the text color of the progress sub-component. 
  * @prop --calcite-input-suffix-background-color:  defines the background color of the suffix sub-component.
  * @prop --calcite-input-suffix-text-color:  defines the text color of the suffix sub-component.
  * @prop --calcite-input-text-color: defines the text color of the component.
  *
  */
 
+calcite-progress {
+  --calcite-progress-background-color: var(--calcite-input-progress-background-color);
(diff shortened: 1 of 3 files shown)


alisonailea (approved):
🚀 Looks good!

packages/calcite-components/src/components/input/input.scss
```diff
@@ -18,10 +18,18 @@
  * @prop --calcite-input-placeholder-text-color: defines the color of placeholder text in the component.
  * @prop --calcite-input-prefix-background-color: defines the background color of the prefix sub-component.
  * @prop --calcite-input-prefix-text-color:  defines the text color of the prefix sub-component.
+ * @prop --calcite-input-progress-background-color: Defines the background color of the progress sub-component.
+ * @prop --calcite-input-progress-fill-color: Defines the background color of the progress sub-component bar.
```
driskull, line 22: @alisonailea @jcfranco are we using "fill-color" or should this be more like `--calcite-input-progress-bar-background-color`? I don't think we need to distinguish "fill" from "background-color" as its more of an implementation detail. IMO this is really just the background color of the "bar".

driskull: I do see that we use `-fill-color` in `meter` and `rating` components as well. Just wanted to bring this up.

driskull (approved):
👍

packages/calcite-components/src/components/input/input.scss
Elijbet: On the other hand, since bar is the active color and background is the grey stretch, it's sort of intuitive to call it fill, visually, from user perspective. I think if the point of name change is to hide the implementation details, fill does it better.

jcfranco: I think `fill` works and is already being used. It looks like it's common too: https://spectrum.adobe.com/page/progress-bar/#Anatomy.

I think it'll be worth it to do an audit once the epic PR is close to ensure all props are consistently named. cc @alisonailea

driskull: sounds good

alisonailea: I would agree fill is a more accurate name. Although I wonder if `--calcite-progress-bar-color` and `--calcite-progress-color` wouldn't be even more accurate?

driskull: It just seems like `fill-color` and `background-color` are pretty vague. Like what is being filled? How does it differ from a `background-color` of a specific thing inside the component.

PR #8888, opened 2024-03-05
docs(inline-editable): add missing background-color CSS prop

jcfranco:
**Related Issue:** #7180

## Summary

Adds doc for `--calcite-inline-editable-background-color`.


jcfranco:
Merging as this was already approved in #8788 and unintentionally removed by https://github.com/Esri/calcite-design-system/pull/8922.

Issue #8892, opened 2024-03-06
[List] Add new `filter-no-results` slot to the doc site


geospatialem:
### Description

The `list` component has a new slot, `filter-no-results`, but its not currently documented on our list page.

### Which Component

`list`

### Resources

Refer to https://developers.arcgis.com/calcite-design-system/components/list/#slots. However, it is mentioned in our [summary notes](https://developers.arcgis.com/calcite-design-system/releases/summaries/2024-02/#list-slot-for-no-filtered-results).

![image](https://github.com/Esri/calcite-design-system/assets/5023024/9a08a79f-c5e5-418b-bc69-42a22539b3ae)

DitwanP:
🍠 Verified!

PR #8893, opened 2024-03-06
docs(list): add filter-no-results slot

geospatialem:
**Related Issue:** #8892

## Summary
Adds the `filter-no-results` slot to the `list` component's doc.


driskull (approved):
👍

PR #8894, opened 2024-03-06
feat(navigation-logo): adds component tokens

anveshmekala:
**Related Issue:** #7180 

## Summary

Adds the following tokens in `calcite-navigation-logo` component

``` 
--calcite-navigation-logo-background-color: Specifies the background color of the component.
--calcite-navigation-logo-border-color: Specifies the border color of the component.
--calcite-navigation-logo-description-text-color: Specifies text color of description of component.
--calcite-navigation-logo-heading-text-color: Specifies the text color of heading of component.
--calcite-navigation-logo-text-color: Specifies the text color of the component.

### CODE REVIEW

File: packages/calcite-components/src/components/navigation-logo/navigation-logo.scss [hunk 1]
```diff
@@ -1,3 +1,20 @@
...
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-navigation-logo-active-border-color: Specifies the border color of the component when active property is true.
+ * @prop --calcite-navigation-logo-active-text-color: Specifies the text color of component when active property is true.
+ * @prop --calcite-navigation-logo-background-color-active: Specifies the background color of the component when active.
+ * @prop --calcite-navigation-logo-background-color-hover: Specifies the background color of the component when hovered.
+ * @prop --calcite-navigation-logo-border-color-active: Specifies the border color of the component when active.
+ * @prop --calcite-navigation-logo-description-text-color: Specifies text color of description of component.
+ * @prop --calcite-navigation-logo-heading-text-color: Specifies the text color of heading of component.
+ * @prop --calcite-navigation-logo-text-color-active: Specifies the text color of component when active.
```
macandcheese, line 13: Seems like we’d want hover mapped for text as well with the new stateful pattern.

But; it does beg the question - do we need stateful tokens for the active state? It’s reasonable to imagine a theming use case that would want the ability to adjust this state on hover. Of course a user can target this state in their css but it breaks the new pattern we were told to use.

anveshmekala: Good catch, don't think `--calcite-navigation-logo-text-color-active` is used in active state of the component. Removed this and updated the components to follow the pattern of reassigning tokens for each state. For this component we can leverage that since all the state styles are applied at root element in shadowDOM rather than subcomponents or children elements. Thoughts?

anveshmekala: Thought the pattern of using stateful vs stateless tokens is dependent on the component structure. navigation-logo is not being used as subcomponent in shadowDOM of any other component  and user can theme the state in css. Do we want to avoid that

macandcheese: Yeah, I'm not sure honestly. I thought the stateless tokens were only for components inaccessible to a dev-user, and others were set by a user with custom-element:hover, etc. cc @alisonailea for clarification.

packages/calcite-components/src/components/navigation-logo/navigation-logo.scss
```diff
@@ -43,39 +63,39 @@
...
-  @apply text-color-1;
+  color: var(--calcite-navigation-logo-text-color-active, var(--calcite-color-text-1));
 }
 
 :host([active]) .anchor {
-  @apply text-color-1;
-  border-color: var(--calcite-color-brand);
+  color: var(--calcite-navigation-logo-active-text-color, var(--calcite-color-text-1));
+  border-color: var(--calcite-navigation-logo-active-border-color, var(--calcite-color-brand));
   --calcite-ui-icon-color: var(--calcite-color-brand);
```
alisonailea, line 72: can you swap this for `--calcite-icon-color: var(--calcite-color-brand);`

packages/calcite-components/src/components/navigation-logo/navigation-logo.scss
```diff
@@ -10,27 +27,30 @@
...
-    no-underline
-    text-0h;
-  color: inherit;
-  border-block-end: 2px solid transparent;
+    no-underline;
+  color: var(--calcite-navigation-logo-text-color, inherit);
+  font-size: var(--calcite-font-size-md);
+  line-height: var(--calcite-font-line-height-fixed-lg);
+  border-block-end: var(--calcite-spacing-base) solid
+    var(--calcite-navigation-logo-background-color, var(--calcite-color-foreground-1));
```
alisonailea, line 35: should this be `--calcite-navigation-logo-border-color`?

packages/calcite-components/src/components/navigation-logo/navigation-logo.scss
alisonailea: I don't think we need stateful tokens here.

alisonailea (changes requested):

packages/calcite-components/src/components/navigation-logo/navigation-logo.scss
anveshmekala: Yup , good catch.

packages/calcite-components/src/components/navigation-logo/navigation-logo.scss
```diff
@@ -10,27 +27,30 @@
...
 }
 
 .anchor:hover,
 .anchor:focus {
-  @apply bg-foreground-2;
+  background-color: var(--calcite-navigation-logo-background-color-hover, var(--calcite-color-foreground-2));
```
alisonailea, line 40: in a case like this you can do one of two things. 
1. set the background color on the base `.anchor` as `background-color: var(--calcite-navigation-logo-background-color, var(--calcite-internal-navigation-logo-background-color), transparent);` and then reassign the internal var `--calcite-internal-navigation-logo-background-color` when `.anchor:hover, .anchor:focus {--calcite-internal-navigation-logo-background-color: var(--calcite-color-foreground-2); }`

2. set style props based on the :host state. `:host(:hover), :host(:focus) { .anchor { background-color: var(--calcite-navigation-logo-background-color-hover, var(--calcite-color-foreground-2)); }}`

3. use state props 
```
.anchor:hover,
.anchor:focus {
  background-color: var(--calcite-navigation-logo-background-color-hover, var(--calcite-color-foreground-2));
```

packages/calcite-components/src/components/navigation-logo/navigation-logo.scss
```diff
@@ -10,27 +27,30 @@
...
 }
 .anchor:focus {
   @apply focus-inset;
 }
 
 .anchor:active {
-  @apply bg-foreground-3;
+  background-color: var(--calcite-navigation-logo-background-color-active, var(--calcite-color-foreground-3));
```
alisonailea, line 47: same as comment above

packages/calcite-components/src/components/navigation-logo/navigation-logo.scss
```diff
@@ -43,39 +63,39 @@
...
 
 :host(:active) .anchor {
-  @apply text-color-1;
+  color: var(--calcite-navigation-logo-text-color-active, var(--calcite-color-text-1));
 }
 
 :host([active]) .anchor {
-  @apply text-color-1;
-  border-color: var(--calcite-color-brand);
+  color: var(--calcite-navigation-logo-active-text-color, var(--calcite-color-text-1));
```
alisonailea, line 70: if you are going to use stateful tokens and not #1 or #2 listed above, this needs to be `--calcite-navigation-logo-text-color-active`

alisonailea (changes requested):
Per Adam's comment. I don't think we need stateful tokens here. I've shown three ways to get to the tokens we need here. I'd vote #1 or #2 but it's up to you.

anveshmekala:
>I don't think we need stateful tokens here.

This part i am bit confused. Initially had them and later removed them and added back. The refactor doc says tokens should be added for appropriate states. I can see the reasoning for adding them since hover and active are appropriate states for navigation-logo component. On the other hand, user can simply set the state on component selector and change the token variable if required. 

Are we adding state tokens for components which are inaccessible to the user (inside shadowDOM) or components which can used as sub-components for other components?

alisonailea (approved):
🎉 Looks great! Thanks

PR #8906, opened 2024-03-08
feat(stepper, stepper-item): add component tokens

anveshmekala:
**Related Issue:** #7180 

## Summary

Adds tokens for `stepper` and `stepper-item` components.

`calcite-stepper-item` tokens:

```  
--calcite-stepper-item-description-text-color: Specifies text color of description of component.
--calcite-stepper-item-heading-text-color: Specifies text color of heading of component.
--calcite-stepper-item-icon-color: Specifies color of icon of component.
--calcite-stepper-item-number-text-color: Specifies text color of number of component.
--calcite-stepper-item-text-color: Specifies text color of component.
```

`calcite-stepper` tokens:

``` 
--calcite-stepper-action-background-color: defines the background color of an action sub-component inside the component.
--calcite-stepper-action-background-color-hover: defines the background color of an action sub-component when hovered or focused.
--calcite-stepper-action-background-color-active: defines the background color of an action sub-component when active.
--calcite-stepper-step-bar-fill-color : Specifies fill color for step-bar.
--calcite-stepper-step-bar-fill-color-hover : Specifies fill color for step-bar when hovered.
--calcite-stepper-step-bar-selected-fill-color : Specifies fill color for selected step-bar.
--calcite-stepper-step-bar-complete-fill-color : Specifies fill color for completed step-bar.
--calcite-stepper-step-bar-completed-fill-color-hover : Specifies fill color for completed step-bar when hovered.
--calcite-stepper-step-bar-error-fill-color : Specifies fill color for error step-bar.
--calcite-stepper-step-bar-error-fill-color-hover : Specifies fill color for error step-bar when hovered.

### DIFF
--- a/packages/calcite-components/src/components/stepper-item/stepper-item.scss
+++ b/packages/calcite-components/src/components/stepper-item/stepper-item.scss
@@ -3,7 +3,6 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
- * @prop --calcite-stepper-item-border-color: Specifies border color of component.
  * @prop --calcite-stepper-item-description-text-color: Specifies text color of description of component.
  * @prop --calcite-stepper-item-heading-text-color: Specifies text color of heading of component.
  * @prop --calcite-stepper-item-icon-color: Specifies color of icon of component.
@@ -85,7 +84,7 @@
   );
   border-color: var(
     --calcite-stepper-item-border-color,
-    var(--calcite-internal-stepper-item-border-color, var(--calcite-color-border-3))
+    var(--calcite-stepper-step-bar-fill-color, var(--calcite-color-border-3))
   );
 }
 
@@ -242,7 +241,7 @@
   border-t-0
   border-solid
   py-0;
-  border-color: var(--calcite-internal-stepper-item-border-color, var(--calcite-color-border-3));
+  border-color: var(--calcite-stepper-step-bar-fill-color, var(--calcite-color-border-3));
   border-inline-start-width: theme("borderWidth.2");
   padding-inline-start: var(--calcite-stepper-item-spacing-unit-l);
 
@@ -260,17 +259,17 @@
 }
 
 :host([layout="vertical"][complete]) .container {
-  border-color: var(--calcite-internal-stepper-item-complete-border-color, #007ac280);
+  border-color: var(--calcite-stepper-step-bar-complete-fill-color, #007ac280);
 }
 :host([layout="vertical"][complete]:hover:not([disabled]):not([selected])) .container,
 :host([layout="vertical"][complete]:focus:not([disabled]):not([selected])) .container {
-  border-color: var(--calcite-internal-stepper-item-complete-border-color-hover, var(--calcite-color-brand));
+  border-color: var(--calcite-stepper-step-bar-complete-fill-color-hover, var(--calcite-color-brand));
 }
 :host([layout="vertical"][error]) .container {
-  border-color: var(--calcite-internal-stepper-item-error-border-color, var(--calcite-color-status-danger));
+  border-color: var(--calcite-stepper-step-bar-error-fill-color, var(--calcite-color-status-danger));
 }
 :host([layout="vertical"][selected]) .container {
-  border-color: var(--calcite-internal-stepper-item-selected-border-color, var(--calcite-color-brand));
+  border-color: var(--calcite-stepper-step-bar-selected-fill-color, var(--calcite-color-brand));
 
   & .stepper-item-content:not(:empty) {
     margin-block-end: var(--calcite-stepper-item-spacing-unit-l);
@@ -278,11 +277,11 @@
 }
 :host([layout="vertical"]:hover:not([disabled]):not([selected])) .container,
 :host([layout="vertical"]:focus:not([disabled]):not([selected])) .container {
-  border-color: var(--calcite-internal-stepper-item-border-color-hover, #007ac280);
+  border-color: var(--calcite-stepper-step-bar-fill-color-hover, #007ac280);
 }
 :host([layout="vertical"][error]:hover:not([disabled]):not([selected])) .container,
 :host([layout="vertical"][error]:focus:not([disabled]):not([selected])) .container {
-  border-color: var(--calcite-internal-stepper-item-error-border-color-hover, var(--calcite-color-status-danger-hover));
+  border-color: var(--calcite-stepper-step-bar-error-fill-color-hover, var(--calcite-color-status-danger-hover));
 }
 
 :host([layout="horizontal"]),
@@ -336,14 +335,14 @@
 :host([layout="horizontal"]),
 :host([layout="horizontal-single"]) {
   .stepper-item-header {
-    border-color: var(--calcite-internal-stepper-item-border-color, var(--calcite-color-border-3));
+    border-color: var(--calcite-stepper-step-bar-fill-color, var(--calcite-color-border-3));
   }
 }
 
 :host([layout="horizontal"][complete]),
 :host([layout="horizontal-single"][complete]) {
   .stepper-item-header {
-    border-color: var(--calcite-internal-stepper-item-complete-border-color, #007ac280);
+    border-color: var(--calcite-stepper-step-bar-complete-fill-color, #007ac280);
   }
 }
 
@@ -352,21 +351,21 @@
 :host([layout="horizontal-single"][complete]:hover:not([disabled]):not([selected])),
 :host([layout="horizontal-single"][complete]:focus:not([disabled]):not([selected])) {
   .stepper-item-header {
-    border-color: var(--calcite-internal-stepper-item-complete-border-color-hover, var(--calcite-color-brand));
+    border-color: var(--calcite-stepper-step-bar-complete-fill-color-hover, var(--calcite-color-brand));
   }
 }
 
 :host([layout="horizontal"][error]),
 :host([layout="horizontal-single"][error]) {
   .stepper-item-header {
-    border-color: var(--calcite-internal-stepper-item-error-border-color, var(--calcite-color-status-danger));
+    border-color: var(--calcite-stepper-step-bar-error-fill-color, var(--calcite-color-status-danger));
   }
 }
 
 :host([layout="horizontal"][selected]),
 :host([layout="horizontal-single"][selected]) {
   .stepper-item-header {
-    border-color: var(--calcite-internal-stepper-item-selected-border-color, var(--calcite-color-brand));
+    border-color: var(--calcite-stepper-step-bar-selected-fill-color, var(--calcite-color-brand));
   }
 }
 
@@ -375,7 +374,7 @@
 :host([layout="horizontal-single"]:hover:not([disabled]):not([selected])),
 :host([layout="horizontal-single"]:focus:not([disabled]):not([selected])) {
   .stepper-item-header {
-    border-color: var(--calcite-internal-stepper-item-border-color-hover, #007ac280);
+    border-color: var(--calcite-stepper-step-bar-fill-color-hover, #007ac280);
   }
 }
 
@@ -384,10 +383,7 @@
 :host([layout="horizontal-single"][error]:hover:not([disabled]):not([selected])),
 :host([layout="horizontal-single"][error]:focus:not([disabled]):not([selected])) {
   .stepper-item-header {
-    border-color: var(
-      --calcite-internal-stepper-item-error-border-color-hover,
-      var(--calcite-color-status-danger-hover)
-    );
+    border-color: var(--calcite-stepper-step-bar-error-fill-color-hover, var(--calcite-color-status-danger-hover));
   }
 }
 
--- a/packages/calcite-components/src/components/stepper-item/stepper-item.scss
+++ b/packages/calcite-components/src/components/stepper-item/stepper-item.scss
@@ -380,7 +380,7 @@
 :host([layout="horizontal-single"]) .stepper-item-header {
   @apply border-none box-border
   me-0;
-  inline-size: 100%;
+  inline-size: var(--calcite-container-size-content-fluid);
   padding-inline: calc(var(--calcite-internal-stepper-action-inline-size) + 0.5rem);
 }
 
--- a/packages/calcite-components/src/components/stepper-item/stepper-item.scss
+++ b/packages/calcite-components/src/components/stepper-item/stepper-item.scss
@@ -7,53 +7,40 @@
  * @prop --calcite-stepper-item-heading-text-color: Specifies text color of heading of component.
  * @prop --calcite-stepper-item-icon-color: Specifies color of icon of component.
  * @prop --calcite-stepper-item-number-text-color: Specifies text color of number of component.
- * @prop --calcite-stepper-item-text-color: Specifies text color of component.
  *
  */
 
 :host([scale="s"]) {
-  --calcite-stepper-item-spacing-unit-s: var(--calcite-spacing-xxs);
-  --calcite-stepper-item-spacing-unit-m: var(--calcite-spacing-md);
-  --calcite-stepper-item-spacing-unit-l: var(--calcite-spacing-xl);
-  --calcite-internal-stepper-action-inline-size: var(--calcite-spacing-xxxl);
-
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: packages/calcite-components/src/components/stepper/stepper.scss [hunk 1]
```diff
@@ -6,27 +8,31 @@
  * @prop --calcite-stepper-action-background-color: defines the background color of an action sub-component inside the component.
  * @prop --calcite-stepper-action-background-color-hover: defines the background color of an action sub-component when hovered or focused.
  * @prop --calcite-stepper-action-background-color-active: defines the background color of an action sub-component when active.
+ * @prop --calcite-stepper-step-bar-fill-color : Specifies fill color for step-bar when `layout="horizontal-single"`.
```
anveshmekala, line 11: Q: Should this tokens be specific ? The reason being that `calcite-stepper-item-border-color` token allows users to specify the border-color at `stepper-item` level.  Adding the above tokens helps users customize them when `layout ="horizontal-single"`. Wonder if its ideal to specify them twice or should this be replaced with `calcite-stepper-item-border-color`.

cc @macandcheese @ashetland @SkyeSeitz

anveshmekala: One way is to get rid of border-top for stepper-item and let user configure step-bar irrespective of layout.

anveshmekala: currently it matches, but if the user wants to customize the colors using tokens they need to setup twice once for horizontal & vertical and again for horizontal-single. Both horizontal and vertical layout depend on stepper-item color where as horizontal-single depends on step-bar component to show border like.

anveshmekala: Sorry for the confusion, i was under the impression that responsive mode is a default setting when the container size decreases but we made a change recently where user needs to specify the layout `horizontal-single` for single stepper view. Make sense to have these separate set of tokens . Ideally one set of tokens would be preferable but the way component is designed can be tricky .

macandcheese: I still don’t know if we need two - can we use the same nomenclature for both? The user still places the steps in the DOM in horizontal-single right? I might be missing something too, lol.

anveshmekala: With the current way the component is design we end up with having one set of tokens to customize border for vertical & horizontal and other set for horizonatal-layout. In horizontal-single layout case the border isn't the border of stepper-item but a functional step-bar which is not attached to any stepper-item.  Using step-bar for all the layout would be ideal but requires a refactor. 


User can customize the border color as following:

``` 
// covers the use-case of layout="horizontal-single"
 calcite-stepper {
      --calcite-stepper-step-bar-fill-color: green;
      --calcite-stepper-step-bar-active-fill-color: #e9f6ff;
      --calcite-stepper-step-bar-complete-fill-color: #280274;
      --calcite-stepper-step-bar-error-fill-color: #fe7a36;
    }
    
    
    
//covers use-case of layout="horizontal" or layout="vertical"

  calcite-stepper-item {
      --calcite-stepper-item-border-color: green;
    }    
    
 calcite-stepper-item[error] {
      --calcite-stepper-item-border-color: #fe7a36;
    }
    
    calcite-stepper-item[complete] {
      --calcite-stepper-item-border-color: #280274;
    }
    
    calcite-stepper-item[selected] {
      --calcite-stepper-item-border-color: #e9f6ff;
    }

Comment (ashetland) [reply packages/calcite-components/src/components/stepper/stepper.scss:11]: I think I refactor might be worth it, in the long-term at least. Not sure if this is the time to do that or not. I think this situation may have come about do to limitations of how we build stuff in Figma. That version needs a separate step-bar for `horizontal-single` out of necessity, but I don't see why we'd need to mimic that in the real version.

Comment (anveshmekala) [reply packages/calcite-components/src/components/stepper/stepper.scss:11]: Agree on the refactor, ideally these colors needs to be customized once irrespective of the layout. Configuring the colors twice is redundant and can be very confusing. 

The reason for adding `step-bar` in the real version is to display the borders which are attached to stepper-item. In `horizontal-single` layout mode where only one item is viewed,  borders of each stepper-item is unavailable. To resolve this situation we came up with the step-bar component. In the initial designs we are displaying only one border in single-view related to currently viewing stepper-item. In further discussion, felt it is better to show the step-bar equivalent to no of stepper-items renders so that user has an idea of how many steps are present.  

i think its better to use step-bar for all the layout and drop the border maybe since `horizontal-single` layout cannot rely on border? (considering stepper-item's are not used outside of stepper) . Open to suggestions.

File: packages/calcite-components/src/components/stepper/stepper.scss [hunk 2]
```diff
@@ -6,27 +8,31 @@
...
  */
 
 :host([scale="s"]) {
-  --calcite-internal-stepper-item-spacing-unit-s: theme("spacing.1");
+  --calcite-internal-stepper-item-spacing-unit-s: var(--calcite-spacing-xxs);
   --calcite-internal-stepper-action-block-size: theme("spacing.11");
-  --calcite-internal-stepper-action-inline-size: theme("spacing.8");
-  --calcite-internal-step-bar-gap: theme("spacing.1");
+  --calcite-internal-stepper-action-inline-size: var(--calcite-spacing-xxxl);
+  --calcite-internal-step-bar-gap: var(--calcite-spacing-xxs);
```
alisonailea, line 22: Leave all rem based styles as REM for now.

packages/calcite-components/src/components/stepper/stepper.scss
```diff
@@ -57,21 +63,21 @@
...
-    gap: theme("spacing.1") var(--calcite-internal-stepper-item-spacing-unit-s);
+    gap: var(--calcite-spacing-xxs) var(--calcite-internal-stepper-item-spacing-unit-s);
   }
 }
 
 :host([layout="horizontal"][scale="l"]),
 :host([layout="horizontal-single"][scale="l"]) {
   .container {
-    gap: theme("spacing.3") var(--calcite-internal-stepper-item-spacing-unit-s);
+    gap: var(--calcite-spacing-md) var(--calcite-internal-stepper-item-spacing-unit-s);
```
alisonailea, line 80: same applies to line 73 and line 80

packages/calcite-components/src/components/stepper-item/stepper-item.scss
```diff
@@ -120,74 +153,79 @@
...
+    var(--calcite-internal-stepper-item-heading-text-color, var(--calcite-color-text-2))
+  );
+  font-weight: var(--calcite-font-weight-medium);
 }
+
 :host .stepper-item-description {
-  @apply text-color-3;
+  color: var(
+    --calcite-stepper-item-description-text-color,
+    var(--calcite-internal-stepper-item-description-text-color, var(--calcite-color-text-3))
```
alisonailea, line 170: Given that --calcite-stepper-item-description-text-color, --calcite-stepper-item-icon-color, and --calcite-stepper-item-number-text-color all share the same default value. Let's set them to all fallback to --calcite-stepper-item-text-color

`color: var(
    --calcite-stepper-item-description-text-color,
    var(--calcite-stepper-item-text-color, var(--calcite-internal-stepper-item-description-text-color, var(--calcite-color-text-3))));`

packages/calcite-components/src/components/stepper-item/stepper-item.scss
```diff
@@ -120,74 +153,79 @@
...
 }
 
 :host .stepper-item-number {
-  @apply text-color-3
-    font-medium
-    duration-150
+  @apply duration-150
     ease-in-out;
   margin-inline-end: var(--calcite-stepper-item-spacing-unit-m);
+  color: var(
```
alisonailea, line 178: same here
`color: var(--calcite-stepper-item-number-text-color, var(--calcite-stepper-item-text-color, var(--calcite-internal-stepper-item-number-text-color, var(--calcite-color-text-3))));`

packages/calcite-components/src/components/stepper-item/stepper-item.scss
```diff
@@ -256,13 +297,12 @@
...
-    @apply border-color-3
-    border-0
+    @apply border-0
     border-solid
-    border-t-2
     focus-base;
     grid-row: items;
-
+    border-block-start-width: var(--calcite-spacing-base);
+    border-color: var(--calcite-color-border-3);
```
alisonailea, line 305: I can't find a storybook example where this border is applied? Can you double check if this even does anything? if not let's remove it. If so, let's give it a component token.

packages/calcite-components/src/components/stepper-item/stepper-item.scss
```diff
@@ -296,46 +336,61 @@
   }
 }
 
+:host([layout="horizontal"]),
+:host([layout="horizontal-single"]) {
+  .stepper-item-header {
+    border-color: var(
```
alisonailea, line 342: same with this one. It doesn't appear that the stepper-item-header border style rules is actually effecting styling. Can you double check?

packages/calcite-components/src/components/stepper/stepper.scss
alisonailea: How much would a refactor take? This seems like a lot of extra tokens.

alisonailea (commented):
Stepper and Stepper Item can not exist without one another so let's try to share tokens/styles where possible

The following should be set only in Stepper and can be simplified/inherited by stepper-item

--calcite-internal-stepper-item-spacing-unit-s = --calcite-stepper-item-spacing-unit-s
--calcite-internal-stepper-action-inline-size = --calcite-internal-stepper-action-inline-size
--calcite-internal-step-bar-gap = stepper-item:host { margin-inline-end }

As to the refactor, there are a few other components that had a bit of a refactor to reduce component tokens. How much of a refactor are we talking about here? I think it might be worth it.

alisonailea (changes requested):

packages/calcite-components/src/components/stepper/stepper.scss
```diff
@@ -118,7 +121,7 @@
   }
 
   .step-bar--active {
-    fill: var(--calcite-stepper-step-bar-active-fill-color, var(--calcite-color-brand));
+    fill: var(--calcite-stepper-step-bar-selected-fill-color, var(--calcite-color-brand));
```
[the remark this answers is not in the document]
anveshmekala: This should be `selected` since we are matching with stepper-item status.

packages/calcite-components/src/components/stepper/stepper.scss
```diff
@@ -87,45 +93,45 @@
 
 .action-container {
   @apply absolute flex justify-between;
-  padding-block: 0.25rem;
-  inline-size: 100%;
+  padding-block: var(--calcite-spacing-xxs);
+  inline-size: var(--calcite-container-size-content-fluid);
 }
 
 @mixin stepBar {
```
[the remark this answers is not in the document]
anveshmekala: The reason for mixin is to isolate the step-bar related styles and improve readability as [suggested](https://github.com/Esri/calcite-design-system/pull/7744#discussion_r1393058738).

packages/calcite-components/src/components/stepper/stepper.scss
anveshmekala: Curious if there is a downside of using semantic tokens for REM values.

packages/calcite-components/src/components/stepper-item/stepper-item.scss
```diff
@@ -89,24 +118,28 @@
...
     h-3
     flex-shrink-0
     self-start
     duration-150
     ease-in-out;
+  color: var(
+    --calcite-stepper-item-icon-color,
+    var(--calcite-internal-stepper-item-icon-color, var(--calcite-color-text-3))
+  );
+  opacity: var(--calcite-opacity-disabled);
```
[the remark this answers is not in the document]
anveshmekala: opacity tokens aren't either public or scale based, adding internal tokens with fallback and reassigning them isn't required right?

packages/calcite-components/src/components/stepper-item/stepper-item.scss
anveshmekala: This redundant with https://github.com/Esri/calcite-design-system/pull/8906/files#diff-c77e15f7b2a230aaeb9146248b360a9ac7d9e7b867732486b0982ad26d59f43dR33. removed.

packages/calcite-components/src/components/stepper-item/stepper-item.scss
anveshmekala: Curious if there is an advantage of setting to `--calcite-stepper-item-text-color` as fallback color  and then adding `--calcite-internal-stepper-item-description-text-color` as fallback for `--calcite-stepper-item-text-color . 


This can be really confusing in terms of readability me thinks and having internal tokens for overriding tokens based on states would be much clearer. 

Looking at  `--calcite-stepper-item-text-color`  much closer and realized it isn't being used any where. We can remove it. Thoughts?

packages/calcite-components/src/components/stepper-item/stepper-item.scss
```diff
@@ -1,36 +1,59 @@
...
-  --calcite-internal-stepper-action-inline-size: theme("spacing.8");
-  @apply text-n1h;
+  --calcite-stepper-item-spacing-unit-s: var(--calcite-spacing-xxs);
+  --calcite-stepper-item-spacing-unit-m: var(--calcite-spacing-md);
+  --calcite-stepper-item-spacing-unit-l: var(--calcite-spacing-xl);
+  --calcite-internal-stepper-action-inline-size: var(--calcite-spacing-xxxl);
+
+  font-size: var(--calcite-font-size);
+  line-height: var(--calcite-font-line-height-fixed-base);
   margin-inline-end: theme("margin.1");
```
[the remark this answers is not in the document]
anveshmekala: Not sure why we required separate ones for margin and spacing in first place.

packages/calcite-components/src/components/stepper-item/stepper-item.scss
```diff
@@ -120,74 +153,79 @@
...
+    --calcite-stepper-item-number-text-color,
+    var(--calcite-internal-stepper-item-number-text-color, var(--calcite-color-text-3))
+  );
+  font-weight: var(--calcite-font-weight-medium);
 }
 
 @include disabled();
 
 :host([complete]) .container {
   // todo dark mode
```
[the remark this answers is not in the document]
anveshmekala: can we remove this @macandcheese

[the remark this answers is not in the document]
macandcheese: I believe so, this comment was for the opacity style that was previously here. If it works in dark mode, feel free to remove it.

packages/calcite-components/src/components/stepper/stepper.scss
```diff
@@ -6,27 +8,32 @@
  * @prop --calcite-stepper-action-background-color: defines the background color of an action sub-component inside the component.
  * @prop --calcite-stepper-action-background-color-hover: defines the background color of an action sub-component when hovered or focused.
  * @prop --calcite-stepper-action-background-color-active: defines the background color of an action sub-component when active.
+ * @prop --calcite-stepper-step-bar-fill-color : Specifies fill color for step-bar.
+ * @prop --calcite-stepper-step-bar-fill-color-hover : Specifies fill color for step-bar when hovered.
+ * @prop --calcite-stepper-step-bar-selected-fill-color : Specifies fill color for selected step-bar.
+ * @prop --calcite-stepper-step-bar-complete-fill-color : Specifies fill color for completed step-bar.
+ * @prop --calcite-stepper-step-bar-completed-fill-color-hover : Specifies fill color for completed step-bar when hovered.
+ * @prop --calcite-stepper-step-bar-error-fill-color : Specifies fill color for error step-bar.
+ * @prop --calcite-stepper-step-bar-error-fill-color-hover : Specifies fill color for error step-bar when hovered.
```
anveshmekala, line 17: Should this be `-fill-color` or `-color` would be a better fit?

alisonailea: I think in other components we're using "fill"

alisonailea (approved):

PR #8918, opened 2024-03-08
fix(navigation-user): allows theming with tokens

anveshmekala:
**Related Issue:** #7180 

## Summary
Allows theming with tokens and removes `--calcite-navigation-user-text-color` token from `navigation-user`.

### DIFF
--- a/packages/calcite-components/src/components/navigation-user/navigation-user.scss
+++ b/packages/calcite-components/src/components/navigation-user/navigation-user.scss
@@ -9,7 +9,6 @@
  * @prop --calcite-navigation-user-border-color: Specifies the border color of the component.
  * @prop --calcite-navigation-user-full-name-text-color: Specifies the text color of full name of the component.
  * @prop --calcite-navigation-user-name-text-color: Specifies the text color of user name of the component.
- * @prop --calcite-navigation-user-text-color: Specifies the text color of the component.
  *
  */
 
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/navigation-user/navigation-user.scss
```diff
@@ -24,28 +23,29 @@
...
+  --calcite-internal-navigation-user-background-color: var(--calcite-color-foreground-2);
 }
 
 :host(:focus) .button {
   @apply focus-inset;
 }
 
 :host(:active) .button {
-  --calcite-navigation-user-background-color: var(--calcite-color-foreground-3);
-  --calcite-navigation-user-text-color: var(--calcite-color-text-1);
```
alisonailea, line 48: I might be wrong but I think you want to add this here?
`--calcite-internal-navigation-user-text-color: var(--calcite-color-text-1);`

alisonailea (commented):
quick question about the text-color. Is the component is expecting for a text color to be set or can it inherit the value from it's parent?

anveshmekala:
> quick question about the text-color. Is the component is expecting for a text color to be set or can it inherit the value from it's parent?

I don't think text-color needs to be set for this component since there are no elements that needs the `color` property. The component has avatar ,username & full name  which can be themed using the specific tokens and slotting is not expected.

alisonailea (approved):

PR #8921, opened 2024-03-09
feat(alert): add component tokens

jcfranco:
**Related Issue:** #7180 

## Summary

* `--calcite-alert-accent-color`
* `--calcite-alert-background-color`
* `--calcite-alert-border-color`
* `--calcite-alert-chip-background-color`
* `--calcite-alert-chip-border-color`
* `--calcite-alert-chip-corner-radius`
* `--calcite-alert-chip-text-color`
* `--calcite-alert-close-background-color-active`
* `--calcite-alert-close-background-color-focus`
* `--calcite-alert-close-background-color-hover`
* `--calcite-alert-close-background-color`
* `--calcite-alert-close-icon-color-active`
* `--calcite-alert-close-icon-color-focus`
* `--calcite-alert-close-icon-color-hover`
* `--calcite-alert-close-icon-color`
* `--calcite-alert-corner-radius`
* `--calcite-alert-shadow`

**Note**: `--calcite-alert-min-height` was removed as it was not used.

### DIFF
--- a/packages/calcite-components/src/components/alert/alert.scss
+++ b/packages/calcite-components/src/components/alert/alert.scss
@@ -48,8 +48,8 @@
   w-full
   z-toast;
 
-  --calcite-internal-border-color: var(--calcite-alert-border-color, var(--calcite-color-border-3));
-  --calcite-internal-corner-radius: var(--calcite-alert-corner-radius, var(--calcite-border-radius));
+  --calcite-internal-alert-border-color: var(--calcite-alert-border-color, var(--calcite-color-border-3));
+  --calcite-internal-alert-corner-radius: var(--calcite-alert-corner-radius, var(--calcite-border-radius));
 
   background-color: var(--calcite-alert-background-color, var(--calcite-color-foreground-1));
   border-radius: var(--calcite-internal-alert-corner-radius, var(--calcite-border-radius));
--- a/packages/calcite-components/src/components/alert/alert.scss
+++ b/packages/calcite-components/src/components/alert/alert.scss
@@ -5,6 +5,9 @@
  *
  * @prop --calcite-alert-background-color: Specifies the background color of the component.
  * @prop --calcite-alert-border-color: Specifies the border color of the component.
+ * @prop --calcite-alert-chip-background-color: Specifies the background color of the queue count chip in the component.
+ * @prop --calcite-alert-chip-border-color: Specifies the border color of the queue count chip in the component.
+ * @prop --calcite-alert-chip-corner-radius: Specifies the corner radius of the queue count chip in the component.
  * @prop --calcite-alert-chip-text-color: Specifies the color of the queue count chip text in the component.
  * @prop --calcite-alert-close-background-color-active: Specifies the background color of the close button in the component when the close button is active.
  * @prop --calcite-alert-close-background-color-focus: Specifies the background color of the close button in the component when the close button is focused.
@@ -135,7 +138,6 @@
   transition-default;
   border-inline: 0 solid transparent;
   border-start-end-radius: 0;
-  color: var(--calcite-alert-chip-text-color, var(--calcite-color-text-2));
 
   &--active {
     @apply visible opacity-100;
@@ -329,6 +331,13 @@ $alertDurations:
  */
 .container--slotted-in-shell {
   @apply absolute;
+}
+
+calcite-chip {
+  --calcite-chip-background-color: var(--calcite-alert-chip-background-color);
+  --calcite-chip-border-color: var(--calcite-alert-chip-border-color);
+  --calcite-chip-corner-radius: var(--calcite-alert-chip-corner-radius);
+  --calcite-chip-text-color: var(--calcite-alert-chip-text-color, var(--calcite-color-text-2));
 }
 
 @include base-component();
--- a/packages/calcite-components/src/components/alert/alert.scss
+++ b/packages/calcite-components/src/components/alert/alert.scss
@@ -3,20 +3,35 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
+ * @prop --calcite-alert-background-color: Specifies the background color of the component.
+ * @prop --calcite-alert-border-color: Specifies the border color of the component.
+ * @prop --calcite-alert-chip-text-color: Specifies the color of the queue count chip text in the component.
+ * @prop --calcite-alert-close-background-color-active: Specifies the background color of the close button in the component when the close button is active.
+ * @prop --calcite-alert-close-background-color-focus: Specifies the background color of the close button in the component when the close button is focused.
+ * @prop --calcite-alert-close-background-color-hover: Specifies the background color of the close button in the component when the close button is hovered.
+ * @prop --calcite-alert-close-background-color: Specifies the background color of the close button in the component.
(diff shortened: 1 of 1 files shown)


packages/calcite-components/src/components/alert/alert.scss
```diff
@@ -3,20 +3,38 @@
...
+ * @prop --calcite-alert-close-background-color-focus: Specifies the background color of the close button in the component when the close button is focused.
+ * @prop --calcite-alert-close-background-color-hover: Specifies the background color of the close button in the component when the close button is hovered.
+ * @prop --calcite-alert-close-background-color: Specifies the background color of the close button in the component.
+ * @prop --calcite-alert-close-icon-color-active: Specifies the color of the close button icon in the component when the close button is active.
+ * @prop --calcite-alert-close-icon-color-focus: Specifies the color of the close button icon in the component when the close button is focused.
+ * @prop --calcite-alert-close-icon-color-hover: Specifies the color of the close button icon in the component when the close button is hovered.
+ * @prop --calcite-alert-close-icon-color: Specifies the color of the close button icon in the component.
+ * @prop --calcite-alert-corner-radius: Specifies the corner radius of the component.
+ * @prop --calcite-alert-link-text-color: Specifies the color of the link text in the component.
```
macandcheese, line 21: Because these are all slotted (link, message, title) - can we just set defaults and have users override on their slotted content (no css prop?)

packages/calcite-components/src/components/alert/alert.scss
```diff
@@ -3,20 +3,38 @@
...
+ * @prop --calcite-alert-message-text-color: Specifies the color of the message text in the component.
+ * @prop --calcite-alert-shadow: Specifies the shadow of the component.
+ * @prop --calcite-alert-status-color: Specifies the status color of the component.
```
macandcheese, line 24: Maybe calcite-alert-accent/highlight-color? This is tied to the `kind` property and not status.

Notice uses `border` - but I think we should change that to accent/highlight as well, and set `notice-border` to none so these can be set independently. Thoughts?

jcfranco: I like the idea of using `accent` to generalize names. Would this properly describe `kind`, `status` and similar props?

macandcheese: Yeah, I think so - in this case the accent is different than the border, I think that wouldn't always be the case (for instance, border-color can represent a selected tab state, that's fine). There are components that have two borders - one "normal border" and one that represents either a selected state, or in this / Notice case, the kind-determined accent / highlight.

macandcheese: If we make the change here I can make one in Notice as well to match, basically adding a second css prop for accent.

packages/calcite-components/src/components/alert/alert.scss
jcfranco: That's **way** cleaner. ✨🧹✨


We could turn your above suggestion into a guideline.

For cases where a component needs to apply specific overrides to slotted content:

* set variables/overrides on the slotted elements
* ☝️ would not necessitate component-level CSS props
* users can override the slotted elements themselves if needed

cc @alisonailea

packages/calcite-components/src/components/alert/alert.scss
alisonailea: I like accent color but can we run it by @ashetland and @SkyeSeitz?

packages/calcite-components/src/components/alert/alert.scss
macandcheese: I *do* think we should have some more consistency across these implementations. I'd like to see props for heading / description as well as slots - often times even simple text is slotted in and contains extra padding / margins / etc., that aren't desirable. Users can still of course have the option of slotted content when needed for rich content like links, images, etc.

But in places like Alert, Notice, Card, etc., it could be nice to have the properties as a first line of defense for consistency, with slots there as option for more advanced use cases. Of course that's out of scope for this PR / effort.

macandcheese: (and, now that I type it out - continuing to re-name all these text lockups to heading / description where applicable - right now we have a mix of that, and title / subtitle / label / etc.)

packages/calcite-components/src/components/alert/alert.scss
ashetland: I like `accent` as well. Moving Notice away from `border` to `accent` would leave border open in the future if Notice ever got a visual update that included a traditional border of some kind. (That's a lot of borders right there, apologies for that.)

packages/calcite-components/src/components/alert/alert.scss
```diff
@@ -26,17 +41,22 @@ $border-style: 1px solid var(--calcite-color-border-3);
...
   text-start
   w-full
   z-toast;
 
-  border-radius: var(--calcite-border-radius);
+  --calcite-internal-alert-border-color: var(--calcite-alert-border-color, var(--calcite-color-border-3));
+  --calcite-internal-alert-corner-radius: var(--calcite-alert-corner-radius, var(--calcite-border-radius));
+
+  background-color: var(--calcite-alert-background-color, var(--calcite-color-foreground-1));
+  border-radius: var(--calcite-internal-alert-corner-radius, var(--calcite-border-radius));
```
alisonailea, line 52: I can't see anywhere the corner radius is redefined. I think this should be `--calcite-alert-corner-radius`?

packages/calcite-components/src/components/alert/alert.scss
```diff
@@ -123,14 +149,15 @@ $border-style: 1px solid var(--calcite-color-border-3);
   inset-inline: 0;
   inset-block-start: -2px;
   block-size: 2px;
-  border-radius: var(--calcite-border-radius) var(--calcite-border-radius) 0 0;
+  border-radius: var(--calcite-internal-alert-corner-radius) var(--calcite-internal-alert-corner-radius) 0 0;
```
alisonailea, line 152: I think this should this be
`border-radius: var(--calcite-alert-corner-radius, var(--calcite-internal-alert-corner-radius) 0 0);`

packages/calcite-components/src/components/alert/alert.scss
```diff
@@ -123,14 +149,15 @@ $border-style: 1px solid var(--calcite-color-border-3);
...
+
   &:after {
     @apply absolute
     top-0
     block;
     block-size: 2px;
     content: "";
-    background-color: var(--calcite-alert-dismiss-progress-background);
+    background-color: var(--calcite-internal-alert-dismiss-progress-background-color);
```
alisonailea, line 160: I think this is only set once so we don't need to do an internal token here.

alisonailea (changes requested):
Just a few more things

packages/calcite-components/src/components/alert/alert.scss
jcfranco: The dismiss progress only needs to match the top corners of the alert.

packages/calcite-components/src/components/alert/alert.scss
jcfranco: Not redefined, but reused.

alisonailea (approved):
Wooo!

PR #8924, opened 2024-03-11
fix(loader): update names

alisonailea:
**Related Issue:** #7180

### Loader
--calcite-loader-color-end: defines the color of a loader. Applies to --calcite-loader-color between 66% - 99% of the animation.
--calcite-loader-color-middle: defines the color of a loader. Applies to --calcite-loader-color between 33% - 66% of the animation.
--calcite-loader-color-start: defines the starting color of a loader. Applies to --calcite-loader-color between 0% - 33% of the animation.


jcfranco (approved):
✨🚦✨

PR #8934, opened 2024-03-13
feat(date-picker, date-picker-month-header): add component tokens

anveshmekala:
**Related Issue:** #7180 

## Summary

Adds following corresponding tokens in `calcite-date-picker` component :

```
--calcite-date-picker-header-background-color: Specifies the background color of month header.
--calcite-date-picker-header-navigation-button-background-color-active: Specifies the background color of navigation button of month header when active.
--calcite-date-picker-header-navigation-button-background-color-hover: Specifies the background color of navigation button of month header when hovered.
--calcite-date-picker-header-navigation-button-background-color: Specifies the background color of navigation button of month header.
--calcite-date-picker-header-navigation-button-text-color-hover: Specifies the text color of navigation button of month header when hovered.
--calcite-date-picker-header-navigation-button-text-color: Specifies the text color of navigation button of month header.
--calcite-date-picker-header-text-color: Specifies the text color of month header.

### DIFF
--- a/packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.scss
+++ b/packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.scss
@@ -1,3 +1,18 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-date-picker-month-header-background-color: Specifies the background color of component.
+ * @prop --calcite-date-picker-month-header-navigation-button-background-color-active: Specifies the background color of navigation button of component when active.
+ * @prop --calcite-date-picker-month-header-navigation-button-background-color-hover: Specifies the background color of navigation button of component when hovered.
+ * @prop --calcite-date-picker-month-header-navigation-button-background-color: Specifies the background color of navigation button of component.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.scss [hunk 1]
```diff
@@ -49,6 +68,11 @@
     px-1
     outline-none
     transition-default;
+  background-color: var(
```
alisonailea, line 71: can we replace this with an assignment to `--calcite-button-background-color`?

The same goes for the other calcite-button tokens.

alisonailea (approved):
Just a quick thing, can we apply date-picker tokens to the calcite-button tokens instead?

alisonailea (approved):
I still wonder how much of this could be simplified/shared between other date picker elements given that you can't have one without the other.

anveshmekala:
> I still wonder how much of this could be simplified/shared between other date picker elements given that you can't have one without the other.

Agree it would be easier to share if the specs are the same. If we consider day and week both have same initial color where as month and year has a diff color as default state. And it would be difficult if want to update/enhance one component if we use same default for all of them.  And the date-picker component shouldn't be worried about the customization of sub components, it has to pass the tokens down the channel.

Issue #3918, opened 2022-01-12
calcite-date-picker-day: API for theming and setting properties from calcite-date-picker or calcite-input-date-picker


jwasilgeo:
### Description

Is there an API currently available to set properties and, separately, change styles/themes of individual `calcite-date-picker-day`s from the level of the parent `calcite-date-picker` or `calcite-input-date-picker`?

The goal is to let individual `calcite-date-picker-day` properties or styles/themes be data-driven and easily applied via the parent `calcite-date-picker` or `calcite-input-date-picker` APIs.

If not, I request a formal API enhancement, or clarified documentation. Thanks!

1. For a property example, I would want to set any number of arbitrary, individual days as `disabled`, as documented as a capability here:
https://github.com/Esri/calcite-components/tree/v1.0.0-beta.74/src/components/calcite-date-picker-day#properties

    As a rough and hacky example of what I could figure out on my own, today at `v1.0.0-beta.74` I'd need to somehow hook into each month change event by the user (<left || right > user button click) when the new month gets rendered, and then try to access an individual `calcite-date-picker-day` in the shadow DOM to change its `disabled` property:
    ```js
    referenceToCalciteInputDatePickerParentComponent
      .shadowRoot.querySelector("calcite-date-picker")
      .shadowRoot.querySelector("calcite-date-picker-month")
      .shadowRoot.querySelectorAll("calcite-date-picker-day")
      // set them all as disabled for the sake of argument,
      // but we're interested in this being data-driven for just some of the individual dates
      .forEach((cdpd) => (cdpd.disabled = true));
    ```

2. As a style/theme example, I would want to change the individual background-color of a individual day to indicate to a user that there is some other attribute at play and they should pay special attention to that day. For example, in the `calcite-*-date-picker` UI, Friday the 13th would appear to have a background of `hotpink` or `deepskyblue` or some other custom styling, because _reasons_.

### Acceptance Criteria

As a developer using `calcite-date-picker` or `calcite-input-date-picker`, I can affect properties and/or styles/themes of individual `calcite-date-picker-day`s rendered in each month of the parent UI.

Or, restated by borrowing a line from the section above:

> The goal is to let individual `calcite-date-picker-day` properties or styles/themes be data-driven and easily applied via the parent `calcite-date-picker` or `calcite-input-date-picker` APIs.

### Which Component

Focused on: `calcite-date-picker-day`

Related to: `calcite-date-picker` and `calcite-input-date-picker`

### Example Use Case

[...]

PR #8936, opened 2024-03-14
refactor(date-picker): removes font token and update naming.

anveshmekala:
**Related Issue:** #7180 

## Summary

Changes in `calcite-date-picker` tokens:

- Removes `--calcite-date-picker-day-font-size` & `--calcite-date-picker-week-headers-border-color ` tokens.
- Renames  `--calcite-date-picker-border-radius` to `--calcite-date-picker-corner-radius`.

### DIFF
--- a/packages/calcite-components/src/components/date-picker-month/date-picker-month.scss
+++ b/packages/calcite-components/src/components/date-picker-month/date-picker-month.scss
@@ -7,13 +7,11 @@
  * @prop --calcite-date-picker-month-day-background-color-selected: Specifies the background color of day range when selected.
  * @prop --calcite-date-picker-month-day-background-color: Specifies the background color of day.
  * @prop --calcite-date-picker-month-day-corner-radius: Specifies the border radius of day.
- * @prop --calcite-date-picker-month-day-font-size: Specifies the font size of day.
  * @prop --calcite-date-picker-month-day-range-background-color-hovered: Specifies the background color of day range when hovered.
  * @prop --calcite-date-picker-month-day-range-background-color-selected: Specifies the background color of day range when selected.
  * @prop --calcite-date-picker-month-day-text-color-hover: Specifies the text color of day when hovered.
  * @prop --calcite-date-picker-month-day-text-color: Specifies the text color of day.
  * @prop --calcite-date-picker-month-day-text-selected: Specifies the text color of day when selected.
- * @prop --calcite-date-picker-month-week-headers-border-color: Specifies the border color of week headers.
  * @prop --calcite-date-picker-month-week-headers-text-color: Specifies the text color of week headers.
 */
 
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/date-picker-month/date-picker-month.scss
```diff
@@ -7,13 +7,11 @@
...
  * @prop --calcite-date-picker-month-day-background-color: Specifies the background color of day.
  * @prop --calcite-date-picker-month-day-corner-radius: Specifies the border radius of day.
- * @prop --calcite-date-picker-month-day-font-size: Specifies the font size of day.
  * @prop --calcite-date-picker-month-day-range-background-color-hovered: Specifies the background color of day range when hovered.
  * @prop --calcite-date-picker-month-day-range-background-color-selected: Specifies the background color of day range when selected.
  * @prop --calcite-date-picker-month-day-text-color-hover: Specifies the text color of day when hovered.
  * @prop --calcite-date-picker-month-day-text-color: Specifies the text color of day.
  * @prop --calcite-date-picker-month-day-text-selected: Specifies the text color of day when selected.
- * @prop --calcite-date-picker-month-week-headers-border-color: Specifies the border color of week headers.
  * @prop --calcite-date-picker-month-week-headers-text-color: Specifies the text color of week headers.
```
alisonailea, line 15: is there a reason we need different variables for month, day, week? Shouldn't all date picker tokens fallback to the same `--calcite-date-picker-` tokens. Especially since some of these components are internal only?

alisonailea (commented):
is there a reason we need different variables for month, day, week? Shouldn't all date picker theme-able styles fallback to the same --calcite-date-picker- tokens. Especially since some of these components are internal only?

anveshmekala:
> is there a reason we need different variables for month, day, week? Shouldn't all date picker theme-able styles fallback to the same --calcite-date-picker- tokens. Especially since some of these components are internal only?

It is possible for user to customize the way each day is rendered and how week headers look. Same applies for month header text which include month and year. There is a request for this already https://github.com/Esri/calcite-design-system/issues/3918 .The naming include these properties to differentiate and fallback values stay inside the component which are internal. Setting the fallback value to same token depends on the design me thinks . Current version has custom default specs for day, month and week text.

alisonailea:
So there's a use case for custom "day" tokens based on https://github.com/Esri/calcite-design-system/issues/3918 but do we need "month" and "week"?

anveshmekala:
> So there's a use case for custom "day" tokens based on #3918 but do we need "month" and "week"?

Yup, i would recommend keeping the week and month tokens too. I don't think there is an issue from user yet but having them would be a good addition me thinks.

packages/calcite-components/src/components/date-picker/date-picker.scss
```diff
@@ -50,12 +48,11 @@
 }
 
 calcite-date-picker-month {
-  --calcite-date-picker-month-week-headers-border-color: var(--calcite-date-picker-week-headers-border-color);
+  // --calcite-date-picker-month-week-headers-border-color: var(--calcite-date-picker-week-headers-border-color);
```
alisonailea, line 51: Remove this?

alisonailea (changes requested):
Looks good! Just one extra comment

PR #8937, opened 2024-03-14
fix(notice): Fix close button focus outline color

macandcheese:
## Summary
Fix for incorrect fallback.

| Before  | After |
| ------------- | ------------- |
| Screenshot 2024-03-14 at 9 38 00 AM  |   Screenshot 2024-03-14 at 9 37 21 AM |


alisonailea (commented):
Outside the scope of this PR but we should make a single "close button" function that all components can share so we don't have to do this in every component with a close button.

alisonailea (approved):

Issue #8928, opened 2024-03-12
calcite-list: filter box disappears in list when scrolling in Safari


AdelheidF:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

![493e2581-f51b-45f0-bcda-af3db60ea50f](https://github.com/Esri/calcite-design-system/assets/44481292/93bf3e4e-0365-4444-8f76-993746da6221)

Chrome and Edge are fine.

### Expected Behavior

filter should stay visible when scrolling in Safari

### Reproduction Sample

https://codepen.io/afreitag/pen/yLrJPpK

### Reproduction Steps

See test cases for `calcite-pick-list` and `calcite-list`. Only `calcite-list` has this issue.

### Reproduction Version

2.6.0

### Relevant Info

OS: macOS Monterey 12.6.8
Browser: Safari (17.3.1)

### Priority impact

p2 - want for current milestone

### Impact

moderate

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Map Viewer

driskull:
Such a quirky safari bug. 

I think we can workaround this by applying the sticky position to the table header instead of the table row within the list.

macandcheese:
Seems similar to https://github.com/Esri/calcite-design-system/issues/7545 where there was a workaround suggested, if it helps in the meantime.

Alijavedofficial:
If the filter box is part of a scrollable container, setting the position of the container to fixed can help prevent the filter box from disappearing during scrolling. This approach is suggested as a solution for a similar issue where elements disappear during scrolling in Safari.

DitwanP:
🍯 Verified `2.7.0-next.14`

PR #8938, opened 2024-03-14
fix(list): fix filter box when scrolling in Safari

driskull:
**Related Issue:** #8928

## Summary

- Moves sticky class to the table header from the table row


alisonailea (approved):

jcfranco (approved):
💥

PR #8940, opened 2024-03-14
feat(table): Add component tokens

macandcheese:
**Related Issue:** #7180 

## Summary
Screenshot 2024-03-14 at 3 51 21 PM

### DIFF
--- a/packages/calcite-components/src/components/table-cell/table-cell.scss
+++ b/packages/calcite-components/src/components/table-cell/table-cell.scss
@@ -55,7 +55,8 @@ td.last-cell {
 }
 
 .footer-cell {
-  @apply font-medium text-color-1;
+  @apply font-medium;
+  color: var(--calcite-color-text-1);
   background-color: var(
     --calcite-table-cell-background-color,
     var(--calcite-table-cell-background, var(--calcite-color-background))
--- a/packages/calcite-components/src/components/table-cell/table-cell.scss
+++ b/packages/calcite-components/src/components/table-cell/table-cell.scss
@@ -85,6 +85,7 @@ td.last-cell {
 }
 
 .selection-cell {
+  color: var(--calcite-table-selection-cell-icon-color, var(--calcite-color-text-3));
   &:not(.footer-cell) {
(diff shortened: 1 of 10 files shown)


packages/calcite-components/src/components/table-row/table-row.scss
```diff
@@ -21,9 +24,23 @@
   }
 }
 
+:host calcite-table-cell {
```
alisonailea, line 27: the :host here isn't adding anything and can be removed.

packages/calcite-components/src/components/table-row/table-row.scss
```diff
@@ -21,9 +24,23 @@
...
+  --calcite-internal-table-cell-background-color: var(
+    --calcite-table-row-background-color,
+    var(--calcite-color-foreground-1)
+  );
+}
+
+:host([selected]) calcite-table-cell {
+  --calcite-table-cell-background-color: var(
```
alisonailea, line 35: I think this should be `--calcite-internal-table-cell-background-color` ?

alisonailea (changes requested):
Looks good! Just a few little things.

packages/calcite-components/src/components/table-cell/table-cell.scss
```diff
@@ -42,44 +54,67 @@ td.last-cell {
...
 }
 
-.number-cell {
-  @apply bg-foreground-2;
-}
-
 .footer-cell {
-  @apply bg-background font-medium text-color-1;
-  border-block-start: 1px solid var(--calcite-color-border-3);
+  @apply font-medium text-color-1;
```
alisonailea, line 58: should this "text-color-1" have a CSS var/token?

alisonailea (approved):
Looks great! One question but approved 🎉

packages/calcite-components/src/components/table-cell/table-cell.scss
macandcheese: Yep, will update

Issue #7732, opened 2023-09-13
Click on a disabled button still causes a popover opening


cont-feofanov:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

A popover whose reference-element is a disabled calcite-button is opened on a button click.

### Expected Behavior

When a reference-element a disabled then a popover should not be displayed  when a disabled button was clicked.

### Reproduction Sample

https://codepen.io/cont-feofanov/pen/zYyzZwN?editors=100

### Reproduction Steps

See the sample

### Reproduction Version

1.8.0

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Business/Community Analyst

geospatialem:
@cont-feofanov Would this [workaround](https://codepen.io/geospatialem/pen/LYMLyBo) support the use case, where the `trigger-disabled` attribute would compliment the `popover` when the `button` is `disabled`?

cont-kolomeets:
@geospatialem Thank you for the workaround, but will it be fixed in the next release? It's a regression from 1.2.0.

jcfranco:
@geospatialem Bumped the priority and updated milestone since we have a few spots where `pointerdown` is used instead of `click`, which may lead to inconsistent behavior.

PR #8943, opened 2024-03-14
fix(action-menu, combobox, dropdown, popover, tooltip): use click instead of pointerdown for click contexts

jcfranco:
**Related Issue:** #7732

## Summary

This updates event handling to use `click` vs `pointerdown` + checking the pointer type wherever clicks are expected. Both events have different semantics and trigger order. Simplified code is also an added bonus.

### DIFF
--- a/packages/calcite-components/src/components/tooltip/TooltipManager.ts
+++ b/packages/calcite-components/src/components/tooltip/TooltipManager.ts
@@ -1,4 +1,4 @@
-import { getShadowRootNode, isPrimaryPointerButton } from "../../utils/dom";
+import { getShadowRootNode } from "../../utils/dom";
 import { ReferenceElement } from "../../utils/floating-ui";
 import { TOOLTIP_OPEN_DELAY_MS, TOOLTIP_CLOSE_DELAY_MS } from "./resources";
 import { getEffectiveReferenceElement } from "./utils";
@@ -119,11 +119,7 @@ export default class TooltipManager {
     }
   };
 
-  private pointerDownHandler = (event: PointerEvent): void => {
-    if (!isPrimaryPointerButton(event)) {
-      return;
-    }
-
+  private clickHandler = (event: PointerEvent): void => {
     const clickedTooltip = this.queryTooltip(event.composedPath());
 
     this.clickedTooltip = clickedTooltip;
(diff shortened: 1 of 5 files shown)


driskull (approved):
click click 💥  🖱️ 👍

packages/calcite-components/src/components/tooltip/TooltipManager.ts
```diff
@@ -119,11 +119,7 @@ export default class TooltipManager {
     }
   };
 
-  private pointerDownHandler = (event: PointerEvent): void => {
-    if (!isPrimaryPointerButton(event)) {
-      return;
-    }
-
+  private clickHandler = (event: PointerEvent): void => {
```
driskull, line 122: should we change to `MouseEvent` for consistency?

driskull: May apply to other places. Doesn't really matter much since PointerEvent extends MouseEvent.

I also notice some other places using onPointerUp/onPointerDown or using PointerEvent typing for click events.

driskull: Should we search and eliminate the util function `isPrimaryPointerButton`?

jcfranco: > should we change to MouseEvent for consistency?

I think for our purposes, we can use `PointerEvent` as it matches the latest spec changes: https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event#event_type. Worth noting that Firefox and Safari haven't updated yet (see [web platform tests](https://wpt.fyi/results/pointerevents?label=master&label=experimental&aligned&q=pointerevents%2F%20is_a_pointerevent)).

> I also notice some other places using onPointerUp/onPointerDown or using PointerEvent typing for click events.

I may have missed some, but I updated the places where `pointerdown` events were being treated as a `click`. I think `isPrimaryPointerButton` is still useful. E.g., color-picker's color field and sliders depend on the primary pointer's up and down events.

jcfranco: If you spot the ones I missed, I can update them in a follow-up PR.

Issue #8931, opened 2024-03-13
[Stepper] When programmatically switching to / from `layout="horizontal-single"`, component breaks 


macandcheese:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When programmatically adjusting the `layout` property to / from `horizontal-single`, the component does not correctly display the steps. It only displays a single step. In some cases, it displays all the step text on top of one another. Switching between only `horizontal` and `vertical` works fine. The issue is worse after adjusting the width of the component - I think the resizeObserver in the component is causing it. And probably unneeded since this is just changed via prop value.

https://github.com/Esri/calcite-design-system/assets/4733155/1b604b1b-b002-4314-83ed-99adc92030fe

### Expected Behavior

I'd expect the component to toggle between states without issue.

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/jORrZvr

### Reproduction Steps

1. Open Codepen
2. Switch to / from the `horizontal-single` `layout` mode. 
3. Notice only a single step will appear in `horizontal-single` with all step text stacked.
4. Adjust the preview window width.
5. Notice the component now behaves differently - showing only a single step, even while switching between the previously-unaffected `horizontal` and `vertical`

### Reproduction Version

2.6.0

### Priority impact

p1 - need for current milestone

### Impact

This happens when programmatically setting the property. The Codepen shows a button-click example, but a more realistic implementation is in the Codepen, a simplified "responsive adjustment" to the property value.

Quick look at the code suggests something in `determineActiveStepper()` or the width-determining `resizeObserver()` could be the issue. Since the property is not set based on width change of the Stepper component, do we need to monitor width? The bug is worse (only shows a single step) after adjusting the page width. The bug is still present before this, just not as bad.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

geospatialem:
Verified in `2.7.0-next.5` with https://codepen.io/geospatialem/pen/qBwbaxp:

![responsive-stepper-validate](https://github.com/Esri/calcite-design-system/assets/5023024/35e8b27f-4795-4fd2-8bd9-ee66800daf99)

Issue #8623, opened 2024-01-18
Respect hidden attribute added by end users on components


driskull:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Some components toggle the hidden attribute which will override any user set hidden attribute on a component.

### Expected Behavior

User added hidden attributes are respected on a component

### Reproduction Sample

N/A Multiple components affected

### Reproduction Steps

1. See following components usage below

### Reproduction Version

2.2.0

### Relevant Info

Blocked issues: https://github.com/Esri/calcite-design-system/issues/8623

```
src/components/combobox/combobox.tsx:
  1041          const hidden = !find(item, filteredData);
  1042:         item.hidden = hidden;
  1043          const [parent, grandparent] = item.ancestors;

  1045          if (find(parent, filteredData) || find(grandparent, filteredData)) {
  1046:           item.hidden = false;
  1047          }

  1049          if (!hidden) {
  1050:           item.ancestors.forEach((ancestor) => (ancestor.hidden = false));
  1051          }

src/components/flow/flow.tsx:
  164          itemNode.showBackButton = itemNode === activeItem && newItemCount > 1;
  165:         itemNode.hidden = itemNode !== activeItem;
  166        });

src/components/stepper/stepper.tsx:
  412          if (index !== activePosition) {
  413:           item.hidden = true;
  414          } else {
  415:           item.hidden = false;
  416            item.multipleViewMode = false;

  426        this.items.forEach((item: HTMLCalciteStepperItemElement) => {
  427:         item.hidden = false;
  428          item.multipleViewMode = true;

src/components/table/table.tsx:
  355        const inView = rowPos >= this.pageStartRow && rowPos < this.pageStartRow + this.pageSize;
  356:       row.hidden = this.pageSize > 0 && !inView && !this.footRows.includes(row);
  357        row.lastVisibleRow =

src/components/tip-manager/tip-manager.tsx:
  201        tip.selected = isSelected;
  202:       tip.hidden = !isSelected;
  203      });
```

### Priority impact

p2 - want for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

josercarcamo:
[...]

PR #8946, opened 2024-03-15
fix(stepper): fix layout when switching modes dynamically to `horizontal-single`

jcfranco:
**Related Issue:** #8931 

## Summary

Fixes an issue that caused step labels to overlap when dynamically switching to `horizontal-single` layout mode.

### DIFF
--- a/packages/calcite-components/src/components/stepper/stepper.tsx
+++ b/packages/calcite-components/src/components/stepper/stepper.tsx
@@ -74,6 +74,7 @@ export class Stepper implements LocalizedComponent, T9nComponent {
   @Watch("scale")
   handleItemPropChange(): void {
     this.updateItems();
+    this.determineActiveStepper();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -401,18 +386,18 @@ export class Stepper implements LocalizedComponent, T9nComponent {
...
-      this.multipleViewMode = false;
-      this.items.forEach((item: HTMLCalciteStepperItemElement, index) => {
-        item.hidden = index !== activePosition;
-      });
-    }
+    const { currentActivePosition, layout } = this;
+
+    this.multipleViewMode = layout !== "horizontal-single";
+    items.forEach((item, index) => {
+      item.hidden = layout === "horizontal-single" && index !== (currentActivePosition || 0);
```
driskull, line 399: Can we use an internal property instead of `hidden`? This would overwrite any user defined `hidden` attributes/properties. Feel free to defer to another issue :)

driskull (approved):
👍

packages/calcite-components/src/components/stepper/stepper.tsx
jcfranco: I totally thought of https://github.com/Esri/calcite-design-system/issues/8623 when I updated this. 🧠 It covers `stepper` already, so will defer to that issue.

Issue #7686, opened 2023-09-06
Audit and mitigate native event cancellation for child elements


geospatialem:
### Background

In https://github.com/Esri/calcite-design-system/pull/7684#discussion_r1317598399 a fix was added to `list-item` where `preventDefault` is not called with the `Enter` key for child elements.

Let's perform an audit of the following components to determine if a similar fix with event cancellation is needed across the design system.

Other utilities and components to confirm:
- [x] `utils/key.ts`
- [x] `color-picker` (and `color-picker-hex-input`)
- [x] `filter`
- [x] `input`
- [x] `input-date-picker`
- [x] `input-number`
- [x] `input-text`
- [x] `input-time-picker`
- [x] `menu-item`

### Desired Outcome

Consistency across the design system.

Elijbet:
It appears that the scope of this issue is a bit more broad and not just focused on the `Enter` key. Based on where we'd like the audit to go, we should rename it to be more generic (something about auditing native event cancelation) and update the description accordingly.

Elijbet:
The following guidelines/patterns can be used to ensure consistent and predictable behavior in nested components with regard to default action prevention in component hierarchies.

For events we handle (like keyDown event with Enter key), we set event.preventDefault() because we have custom logic for the event and we don't want the default logic to happen. This is how events are canceled.

Components that need to be updated:

- input-date-picker
- input-time-picker
- menu-item

Event handlers in components should check `event.defaultPrevented` before handling the event. This is to ensure that the event is not handled multiple times in different components. All component `keyDownHandler`s should apply general ```
`defaultPrevented` check:
		if (defaultPrevented) {
		    return;
		}
```
Components that need to be updated:

- filter
- input
- input-number
- input-text
- menu-item

 
Add docs to recommend users to cancel the event for slotted components if these have a click handler. It is better to check if the event was canceled or not vs checking if it is a slotted component, example scenario:  https://github.com/Esri/calcite-design-system/pull/7684/

isaacbraun:
Verified on `2.8.1` via https://codepen.io/isaacbraun/pen/qEOyBPa. Follow-up changes will be needed for `calcite-filter` in a separate issue (coming soon).

Elijbet:
> Verified on `2.8.1` via https://codepen.io/isaacbraun/pen/qEOyBPa. Follow-up changes to correctly allow input event prevention will be needed for `calcite-filter` in a separate issue (coming soon).

[...]

PR #8952, opened 2024-03-18
fix(input, input-number, input-text, input-date-picker, input-time-picker, filter, menu-item): ignore canceled events and enforce cancellations where needed

Elijbet:
**Related Issue:** #7686

## Summary
Check `event.defaultPrevented` before handling `keyDown` events and `event.preventDefault` for keys we are handling.


packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -893,6 +893,7 @@ export class InputDatePicker
     }
 
     if (key === "Enter") {
+      event.preventDefault();
```
driskull, line 896: Something I'm noticing is that we're having to call `event.preventDefault()` for every key. This is fine, but maybe it would be good to have a dom utility function to handle this better for us.

Something like:

```ts
export function preventDefaultForKeys(event: KeyboardEvent, keys: string[]): void {
  if (!event.defaultPrevented && keys.includes(event.key)) {
    event.preventDefault();
  }
}
```

usage:

```ts
preventDefaultForKeys(["ArrowDown", "ArrowUp", "ArrowLeft", "ArrowRight"], [" ", "Enter"]);
```

what do you think @jcfranco?

packages/calcite-components/src/components/input-text/input-text.tsx
```diff
@@ -415,14 +415,15 @@ export class InputText
...
       return;
     }
+
     if (this.isClearable && event.key === "Escape") {
       this.clearInputTextValue(event);
       event.preventDefault();
     }
-    if (event.key === "Enter" && !event.defaultPrevented) {
+    if (event.key === "Enter") {
       if (submitForm(this)) {
```
driskull, line 427: For the input text, we are only calling preventDefault if the input is inside a form. I wonder if we should call `event.preventDefault()` for the enter key all the time? cc @jcfranco. Is it inconsistent to only call this when inside a form?

packages/calcite-components/src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -665,8 +665,9 @@ export class InputTimePicker
     }
 
     if (key === "Enter") {
+      event.preventDefault();
```
driskull, line 668: In this example, we always call event.preventDefault for enter regardless of the existence of a form. So we should be consistent with the other comment.

packages/calcite-components/src/components/menu-item/menu-item.tsx
```diff
@@ -269,6 +269,11 @@ export class CalciteMenuItem implements LoadableComponent, T9nComponent, Localiz
...
     const key = event.key;
     const targetIsDropdown = event.target === this.dropdownActionEl;
+
+    if (event.defaultPrevented) {
+      return;
+    }
+
     if (key === " " || key === "Enter") {
       if (hasSubmenu && (!href || (href && targetIsDropdown))) {
         this.open = !open;
```
driskull, line 279: In this example, shouldn't we be preventing default for enter and space key regardless?

Currently, it is only preventing default if the key is space, component has href and target is a dropdown.

```ts
if (key === " " || key === "Enter") {
      if (hasSubmenu && (!href || (href && targetIsDropdown))) {
        this.open = !open;
      }
      if (!(href && targetIsDropdown) && key !== "Enter") {
        this.selectMenuItem(event);
      }
      if (key === " " || (href && targetIsDropdown)) {
        event.preventDefault();
      }
    } 
```

driskull (approved):
Looking good to me @Elijbet. Lets just get some feedback from @jcfranco 👍

packages/calcite-components/src/components/input-text/input-text.tsx
jcfranco: > Is it inconsistent to only call this when inside a form?

Not according to [our conventions](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/conventions/README.md#native-event-cancellation). We should only be canceling native events if the component reacts to an event for a specific interaction.

packages/calcite-components/src/components/menu-item/menu-item.tsx
jcfranco: There are if/else statement blocks that won't be executed (e.g., `Enter` + `!hasSubmenu`), so I wouldn't agree we should cancel regardless of key pressed.

packages/calcite-components/src/components/input-time-picker/input-time-picker.tsx
jcfranco: Agreed, we should only cancel when the component reacts to the handled event.

jcfranco:
Looking good, @Elijbet! Is there another way to describe what this PR does? I'm not sure `mitigate native event cancellation` is accurate.

Issue #8954, opened 2024-03-18
[List Item] - calcite-select becomes unresponsive in a calcite-list-item if drag-disabled is true


sagewall:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

A calcite select component becomes unresponsive when placed in a calcite-list-item with drag-disabled set to true.

### Expected Behavior

The calcite select component should behave and respond normally.

### Reproduction Sample

https://codepen.io/sagewall/pen/eYogBjN

### Reproduction Steps

1. Open this [codepen](https://codepen.io/sagewall/pen/eYogBjN)
2. Observe how the calcite select in the test2 list item doesn't respond normally.

### Reproduction Version

2.6.0

### Relevant Info

Doesn't seem to be browser or OS dependent

### Priority impact

p2 - want for current milestone

### Impact

This is impacting custom list item content the LayerList widget in the JavaScript Maps SDK.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

sagewall:
cc @driskull

geospatialem:
Verified in `2.7.0-next.11`

![image](https://github.com/Esri/calcite-design-system/assets/5023024/d82ca218-1f67-44e7-94ee-5ce7f0dc95ea)

PR #8957, opened 2024-03-18
fix(list, list-item): calcite-select becomes unresponsive in a list-item if drag-disabled is true

driskull:
**Related Issue:** #8954

## Summary

- Sortablejs by default calls `event.preventDefault()` because the option `preventOnFilter` is true by default.
- This affects any elements slotted within the `drag-disabled` list item.
- In order to fix this, we move the `filter` selector to just the `calcite-handle` instead of the whole `list-item`.

```js
filter: ".ignore-elements",  // Selectors that do not lead to dragging (String or Function)
preventOnFilter: true, // Call `event.preventDefault()` when triggered `filter`
```

### DIFF
--- a/packages/calcite-components/src/utils/sortableComponent.ts
+++ b/packages/calcite-components/src/utils/sortableComponent.ts
@@ -129,7 +129,7 @@ export function connectSortableComponent(component: SortableComponent): void {
       },
     }),
     handle,
-    filter: "[drag-disabled]",
+    filter: `${handle}[disabled]`,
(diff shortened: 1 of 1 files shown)


jcfranco (approved):
Can the [`dragAndDrop`](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/tests/utils.ts#L48) test util be used for an E2E test for this?

driskull:
That utility wouldn't be able to test this. :\

Issue #8958, opened 2024-03-18
[calcite-table-cell] Allow for styling vertical alignment within a cell


doreenbrinkman:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Allow for styling the vertical alignment within a cell.  Currently, this is not possible because the style that dictates this is in the shadowRoot.

### Expected Behavior

Add a property to set this on the calcite-table-row

### Reproduction Sample

https://codepen.io/doreenbrinkman/pen/KKYNGxP

### Reproduction Steps

1.  Run the code pen sample and try to vertically align the table data to the top.

### Reproduction Version

2.5.1

### Priority impact

p3 - want for upcoming milestone

### Impact

Minor styling issue using `calcite-table`

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Data Pipelines

geospatialem:
It looks like this might also be a visual regression, per https://github.com/Esri/calcite-design-system/pull/8961#issuecomment-2005607867. 

@macandcheese + @jcfranco should we add the `visual changes` to this issue? cc @SkyeSeitz

macandcheese:
Up to you, it’s an intended and minor change.

macandcheese:
Up to you, it’s an intended and minor change, not a regression.

geospatialem:
Verified in `2.7.0-next.11` with https://codepen.io/geospatialem/pen/dyLvaeX and the following HTML:

```html
<calcite-table>
  <calcite-table-row slot="table-header">
    <calcite-table-header heading="Heading - default (start alignment)"></calcite-table-header>
    <calcite-table-header heading="Heading - default (start alignment)"></calcite-table-header>
    <calcite-table-header heading="Heading - end alignment" alignment="end" description="Alignment attribute example"></calcite-table-header>
    <calcite-table-header heading="Heading - center alignment" alignment="center" description="Alignment attribute example"></calcite-table-header>
    <calcite-table-header heading="Heading - default (start alignment)"></calcite-table-header>
  </calcite-table-row>
  <calcite-table-row alignment="start">
    <calcite-table-header heading="Heading - default"></calcite-table-header>
[...]

PR #8961, opened 2024-03-19
feat(table-row): Add alignment property

macandcheese:
**Related Issue:** #8958 

## Summary
Adds an `alignment` property to the Table Row component. When set - aligns slotted Table Header and Table Cell components to this value. Number and Selection cells (internally rendered) will always be vertically centered / middle aligned.

Adjusts padding of Table Header when not slotted in the `table-header` slot to better align with sibling slotted Table Cell components.


macandcheese:
cc @SkyeSeitz @ashetland this does result in slight visual diff - when the Header are in a row with table cell, we no longer apply the 1.5x padding the "table header th" have. FYI and to confirm.

packages/calcite-components/src/components/table-cell/table-cell.scss
```diff
@@ -11,11 +11,11 @@
   @apply contents;
 }
 
-:host([alignment="center"]) td {
+:host([alignment="center"]) td:not(.selection-cell):not(.number-cell) {
```
driskull, line 14: nitpick: I don't love these selectors.

Every time we add another td this selector needs to grow. I know we have a ton of these in the codebase but IMO we should discuss moving this logic into JS and use a css class to represent what this should be styling.

driskull (approved):
This looks good to me. I just think sometimes our CSS selectors can get quite complex. 😨  

👍

packages/calcite-components/src/components/table-cell/table-cell.scss
macandcheese: Agreed, added a class for this.

PR #8965, opened 2024-03-19
fix(tabs): use border prop for bordered styling

jcfranco:
**Related Issue:** #7180 

## Summary

🔳✨🔨

**Note**: this also fixes a regression from https://github.com/Esri/calcite-design-system/pull/8783 where non-bordered tabs displayed a border.

### DIFF
--- a/packages/calcite-components/src/components/tabs/tabs.scss
+++ b/packages/calcite-components/src/components/tabs/tabs.scss
@@ -18,16 +18,18 @@
   box-shadow: inset 0 1px 0 var(--calcite-internal-tabs-border-color);
   background-color: var(--calcite-internal-tabs-background-color);
 
+  section {
+    border-color: var(--calcite-internal-tabs-border-color);
+    border-style: solid;
+  }
+
   ::slotted(calcite-tab-nav) {
     margin-block-end: -1px;
   }
 }
 
 section {
-  @apply border-color-1 border border-solid flex
-  border-t
-  flex-grow
-  overflow-hidden;
+  @apply border flex flex-grow overflow-hidden;
 
   border-block-start-style: solid;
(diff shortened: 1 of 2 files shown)


driskull (approved):

packages/calcite-components/src/components/tabs/tabs.scss
```diff
@@ -15,7 +15,7 @@
 }
 
 :host([bordered]) {
-  box-shadow: inset 0 1px 0 var(--calcite-color-border-1);
+  box-shadow: inset 0 1px 0 var(--calcite-internal-tabs-border-color);
```
driskull, line 18: Should we be using box-shadow as a border??

alisonailea (approved):

packages/calcite-components/src/components/tabs/tabs.scss
jcfranco: I didn't want to affect the scope of this PR, but worth discussing. I know box-shadow + inset is used to achieve a border w/o affecting dimensions.

driskull: Yeah we should discuss these type of styles where we are doing somewhat odd things to accomplish a border. The list component was doing the same thing (using margin and a box-shadow) to accomplish a border. I went ahead and changed it in that PR.

macandcheese: I don't think there's anything inherently _wrong_ with doing this - in certain places it can be helpful to maintain total height, provide a way to avoid having "compensatory padding" (when you have an active border and siblings don't) etc. Or, in cases where border is already used, to provide additional visuals. But, it probably shouldn't be the first option... I can see why it was used in Tabs in this way - I can see if a refactor is possible in #8800

Issue #8685, opened 2024-02-01
[calcite-dropdown] odd behavior when inside a scrollable container


rpanichakit9541:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

calcite-dropdown is not shown correctly when it's started as opened below the fold inside a scrollable container. If I click the button at this point nothing will happen, the dropdown will open if I click it a second time.  A side note, if I scroll the container after the dropdown shows up it will stay fixed in that position even if I scrollwheel the trigger button out of view.

### Expected Behavior

- The dropdown should display as opened correctly even if it's started out of view inside a scrollable container
- The dropdown should stay together with the trigger button when I use the mousewheel to scroll the container

### Reproduction Sample

https://codepen.io/rpanichakit/pen/vYPRVQN?editors=1000

### Reproduction Steps

1. open the Codepen link  https://codepen.io/rpanichakit/pen/vYPRVQN?editors=1000
2. scroll the div with the orange background down to the bottom
3. observe that the dropdown is not shown even though it has `open` prop set
4. click the trigger button once and see that nothing happens
5. click the trigger button again and see the dropdown shows up
6. use the mousewheel to scroll the div up and down to see the dropdown stay put while the trigger scrolls away.

### Reproduction Version

2.4.0

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Online

driskull:
If we call ` connectFloatingUI` again in `componentDidLoad` this seems to be fixed.

We would probably need to do that for all FloatingUI components so that these scrolling listeners are setup correctly. We need to maintain the call in `connectedCallback` as well for when an element is just moved in the DOM.

@jcfranco if this sounds good I can proceed.

since `connectFloatingUI` is also called when the reference and floating elements are created, maybe we should debounce the call to `connectFloatingUI` as well?

jcfranco:
Assigning to this milestone as @driskull has a PR ready. cc @geospatialem

DitwanP:
Verified on `2.7.0-next.14`

[...]

PR #8973, opened 2024-03-20
fix(dropdown): correct positioning behavior when inside a scrollable container

driskull:
**Related Issue:** #8685

## Summary

- Connect floatingUI elements once a component has loaded to correctly setup auto update events
- Automatically call `reposition` inside `connectFloatingUI` for consistency

### DIFF
--- a/packages/calcite-components/src/utils/floating-ui.ts
+++ b/packages/calcite-components/src/utils/floating-ui.ts
@@ -508,8 +508,6 @@ export function connectFloatingUI(
     component,
     runAutoUpdate(referenceEl, floatingEl, () => component.reposition()),
   );
-
-  component.reposition(true);
 }
 
 /**
--- a/packages/calcite-components/src/utils/floating-ui.ts
+++ b/packages/calcite-components/src/utils/floating-ui.ts
@@ -508,6 +508,8 @@ export function connectFloatingUI(
     component,
     runAutoUpdate(referenceEl, floatingEl, () => component.reposition()),
   );
+
+  component.reposition(true);
 }
 
 /**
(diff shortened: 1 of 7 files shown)


packages/calcite-components/src/utils/floating-ui.ts
```diff
@@ -508,6 +508,8 @@ export function connectFloatingUI(
     component,
     runAutoUpdate(referenceEl, floatingEl, () => component.reposition()),
   );
+
+  component.reposition(true);
```
jcfranco, line 512: Won't this cause the component to reposition twice (once via `runAutoUpdate`/`autoUpdate` and again here) when connected?

driskull: I don't think so because when connected, the reference and floatingEl elements are not yet created.

The reason we need this for both connectedCallback and componentDidLoad is because the elements are not yet created in connectedCallback but if a component is moved within the DOM, we do need to reconnect it. So both should be necessary.

driskull: If runAutoUpdate does reposition, initially, then we might not need this but we had it setup like this already so I'm hesitant to remove it.

driskull: @jcfranco do you know if autoUpdate is called initially?

jcfranco: [It appears so](https://github.com/floating-ui/floating-ui/blob/e642116c9e3fea62b8e275a8dcb75c9109b19350/packages/dom/src/autoUpdate.ts#L211).

jcfranco (approved):
Sweet! 🍨

Before merging, would you mind updating the PR title to:

* briefly describe what the odd behavior is
* [per conventions (referenced in our contributing)](https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md#subject), have it be imperative

?

Issue #7886, opened 2023-09-26
[Dialog] Add new component, deprecate modal component


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

The ModelBuider Web team is seeking a movable, non-attached component in early 2024 to support a workflow not requiring an additional click and is not restrictive for the user experience, where the use case cannot use `modal` as a solution.

We should identify if the new dialog component should support:
- Resizing
- Moving

Or, if not, what the recommended workflow would be for support with docking or an alternative approach.

### Acceptance Criteria

Design: **[Figma file](https://www.figma.com/design/wVqftT1aZDjwYVkDZ2V5fC/Dialog?node-id=874-25664&m=dev)** 🎨 

1. Dependency issues
    - **Panel**
        - [x] https://github.com/Esri/calcite-design-system/issues/6980
        - [x] https://github.com/Esri/calcite-design-system/issues/8979
        - [x] https://github.com/Esri/calcite-design-system/issues/8980
        - [x] https://github.com/Esri/calcite-design-system/issues/8981
    - **Modal**
        - [x] https://github.com/Esri/calcite-design-system/issues/7668
2. Add a new `dialog` component
    - "Render a `panel` inside" paradigm - refer to `popover`, `sheet`, and `shell-panel`
    - Properties
        - modal/non-modal
        - `kind`
        - `placement`
        - `heading`
        - `description`
        - `scale`
        - `scale-width`
        - `fullscreen` (under consideration 🚧 )
            - cc https://github.com/Esri/calcite-design-system/issues/6191 
        - `resizable` (under consideration 🚧 )
        - `draggable` (under consideration 🚧 )
    - Slots
        - `content`
        - `header`
3. Deprecate the `modal` component 

### Relevant Info

Originally reported in https://devtopia.esri.com/WebGIS/calcite-design-system/issues/163

### Which Component

New component

### Example Use Case

See the internal issue for additional context.

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

driskull:
> The use case cannot use modal as a solution.

Can you elaborate on this a bit? I don't see why a non blocking modal (scrim turned off) couldn't support this better.

It seems like modal would fit this use case better. 

[...]

PR #8974, opened 2024-03-20
feat(sheet): add component tokens

alisonailea:
**Related Issue:** #7180

### Sheet

--calcite-sheet-scrim-background: [Deprecated] use `--calcite-sheet-scrim-background-color` instead. Specifies the background color of the sheet scrim.
--calcite-sheet-scrim-background-color: Specifies the background color of the sheet scrim.
--calcite-sheet-background-color: Specifies the background color of the sheet.
--calcite-sheet-shadow: Specifies the shadow of the sheet.
--calcite-sheet-text-color: Specifies the text color of the sheet.


packages/calcite-components/src/components/sheet/sheet.scss
```diff
@@ -3,7 +3,11 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
- * @prop --calcite-sheet-scrim-background: Specifies the background color of the sheet scrim.
+ * @prop --calcite-sheet-scrim-background: [Deprecated] use `--calcite-sheet-scrim-background-color` instead. Specifies the background color of the sheet scrim.
+ * @prop --calcite-sheet-scrim-background-color: Specifies the background color of the sheet scrim.
+ * @prop --calcite-sheet-background-color: Specifies the background color of the sheet.
```
macandcheese, line 8: I wonder if this needs to have a BG set at all. I'd probably vote to just remove this and have the container display with transparent BG. A user should be slotting a Panel, etc., so this doesn't seem needed.

packages/calcite-components/src/components/sheet/sheet.scss
```diff
@@ -3,7 +3,11 @@
...
+ * @prop --calcite-sheet-shadow: Specifies the shadow of the sheet.
+ * @prop --calcite-sheet-text-color: Specifies the text color of the sheet.
```
macandcheese, line 10: I don't think we need text-color, everything is slotted by a user.

packages/calcite-components/src/components/sheet/sheet.scss
jcfranco: This would apply to modal too.

macandcheese: Well, Modal does currently have bg color (and separate for Content area). It’s a bit different than Sheet as you don’t slot a Panel into a Modal. That is the current behavior anyway, we have discussed changing that to align more with Sheet, Shell panel, etc.

jcfranco: Gotcha. Do we have an issue for that? For future reference purposes.

alisonailea: would that apply to shell panel too?

macandcheese: #7886 would address that I think.

jcfranco (approved):
📄📄📄📄📄📄📄📄📄📄📄📄📄📄📄📄📄📄📄📄📄📄📄📄
📄✅📄📄📄✅✅✅✅📄✅✅✅✅✅📄✅📄📄📄✅📄✅📄
📄✅📄📄📄✅📄📄📄📄📄📄✅📄📄📄✅✅📄✅✅📄✅📄
📄✅📄📄📄✅📄✅✅📄📄📄✅📄📄📄✅📄✅📄✅📄✅📄
📄✅📄📄📄✅📄📄✅📄📄📄✅📄📄📄✅📄📄📄✅📄📄📄
📄✅✅✅📄✅✅✅✅📄📄📄✅📄📄📄✅📄📄📄✅📄✅📄
📄📄📄📄📄📄📄📄📄📄📄📄📄📄📄📄📄📄📄📄📄📄📄📄

PR #8983, opened 2024-03-22
fix(popover): prevent disabled reference elements from toggling popover

jcfranco:
**Related Issue:** #7732 

## Summary

This changes `PopoverManager` to no longer rely on the capture phase for toggling popovers, so clicks from disabled elements are ignored properly.

**Note**: this also introduces a new DOM util to help determine if a click event was triggered via keyboard or not.


driskull (approved):
💯 🤩

PR #8988, opened 2024-03-22
feat(tip-group): add component tokens

jcfranco:
**Related Issue:** #7180

## Summary

* `--calcite-tip-group-background-color`
* `--calcite-tip-group-text-color`


driskull (approved):
👍

PR #8991, opened 2024-03-23
fix(table): fix style regression

jcfranco:
**Related Issue:** #7180 

## Summary

✨🏓🔨✨

PR #8993, opened 2024-03-25
fix(accordion, accordion-item, action, action-bar, action-group, action-menu, icon, list-item, popover): update components and e2e tests

alisonailea:
**Related Issue:** #7180

## Summary

To be merged after https://github.com/Esri/calcite-design-system/pull/9027

Updates components to fix unit tests, remove extraneous tokens, and clean up styles. 

This is alternative approach to the fixes proposed in https://github.com/Esri/calcite-design-system/pull/8991


packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -1839,3 +1839,158 @@ export function openClose(componentTagOrHTML: TagOrHTML, options?: OpenCloseOpti
     });
   }
 }
+
+export async function themed(
+  componentTagOrHTML: TagOrHTML,
+  tokens: string[],
+  componentName?: string,
+): Promise<{ page: E2EPage; component: E2EElement; themedTokens: Record<string, string> }> {
+  let colorList: string[] = [
```
driskull, line 1848: Do we really need all these colors and the randomness or could we just set everything to `hotpink` since I don't think we'll ever be using a color like that.

packages/calcite-components/src/tests/commonTests.ts
```diff
@@ -1839,3 +1839,158 @@ export function openClose(componentTagOrHTML: TagOrHTML, options?: OpenCloseOpti
...
+
+  colorList = shuffle(colorList);
+
+  const page = await simplePageSetup(componentTagOrHTML);
+  const tag = getTag(isHTML(componentTagOrHTML) ? componentName : componentTagOrHTML);
+  const component = await page.find(tag);
+  const themedTokens = {};
+
+  tokens.forEach((token, i) => {
+    themedTokens[token] = token.includes("color") ? colorList[i] : `${i * 10}${token.includes("z-index") ? "" : "px"}`;
```
driskull, line 1984: nice. 👍 

This might not cover everything though, like shadows, and stuff. I wonder if we should have some kind of lookup function to map all these things? I think we can search for things like space, color, z-index, shadow, anything else?

alisonailea: updated to cover those use cases

packages/calcite-components/src/tests/commonTests.ts
alisonailea: nope. updated.

Issue #6536, opened 2023-02-28
[New Component] Carousel / Carousel Item (from Tip Manager / Tip)


yelenakreyndel:
### Description

Create a new component for Carousel / pane slider. Design work needs to be complete by September for implementation to be complete by January 2024.

### User Stories

@macandcheese 
@mitc7862 
please gather requirements from product teams and update user story based on that information. 

### Acceptance Criteria

please gather requirements from product teams and update acceptance criteria based on that information. 

### Helpful Details

This should re-purpose Tip and Tip Manager - existing components that function similarly to a Carousel. Making those components more generic in name would open up to more use cases while still supporting the few instances where Tip / Tip Manager are used in the wild. The existing use case can be solved for by using the new Carousel / Carousel Item components and slotting content inside each slide - Notice, Blocks, etc., 

### Esri team

Calcite (design)

geospatialem:
cc @benelan for the `stale` label and closing automation. Is there a way we can add the `stale` label, or similar, but not close out issues?

macandcheese:
I think the auto-close is good, but maybe if the tag has been removed since it was last set, do not attempt to re-assign

DitwanP:
🍡Verified locally on `main` 🎠

PR #8995, opened 2024-03-26
feat(carousel): Add Carousel and Carousel Item components

macandcheese:
**Related Issue:** #6536 

## Summary
Opening as draft. Will likely need follow up options for planned features like `controlDisplay`, `arrowDisplay`, `scale`, etc.


packages/calcite-components/src/components/carousel/carousel.tsx
```diff
@@ -0,0 +1,353 @@
...
+  };
+
+  private previousItem = (): void => {
+    this.direction = "retreating";
+    const previousIndex = this.activeIndex === 0 ? this.items?.length - 1 : this.activeIndex - 1;
+    this.setSelectedItem(true, previousIndex);
+  };
+
+  private keyDownHandler = (event: KeyboardEvent): void => {
+    if (event.target !== this.container) {
```
driskull, line 221: I think we should also check for defaultPrevented here.

driskull, line 221:
```suggestion
    if (event.target !== this.container || event.defaultPrevented) {
```

packages/calcite-components/src/components/carousel/carousel.tsx
```diff
@@ -0,0 +1,353 @@
...
+              class={`pagination-item${index === activeIndex ? " active-icon" : ""}`}
+              icon={index === activeIndex ? ICONS.active : ICONS.inactive}
+              id={`${itemGuid}-${index}`}
+              label={item.label}
+              onClick={() => this.setSelectedItem(true, index)}
+              scale="s"
+              text={item.label}
+            />
+            {!this.tooltipsDisabled && (
+              <calcite-tooltip placement="bottom" reference-element={`${itemGuid}-${index}`}>
```
driskull, line 279: nitpick: use the dom element as the reference element if possible. Less lookup in the DOM this way and potential shadow dom issues.

packages/calcite-components/src/components/carousel/carousel.tsx
```diff
@@ -0,0 +1,353 @@
...
+    );
+  }
+
+  render(): VNode {
+    const { direction, activeIndex } = this;
+    return (
+      <Host>
+        <InteractiveContainer disabled={this.disabled}>
+          <div
+            aria-label={this.label}
```
driskull, line 324: I'm not seeing any role for the carousel. Should this follow the following docs? https://www.w3.org/WAI/ARIA/apg/patterns/carousel/examples/carousel-2-tablist/

Missinge "section" elements, aria-roledescription="carousel", etc. cc @geospatialem

packages/calcite-components/src/components/carousel/carousel.tsx
```diff
@@ -0,0 +1,353 @@
...
+            class={{
+              [CSS.container]: true,
+              [CSS.isOverlay]: this.controlOverlay,
+              [CSS.isEdges]: this.arrowType === "edges",
+            }}
+            onKeyDown={this.keyDownHandler}
+            tabIndex={0}
```
driskull, line 331: I'm guessing this is for keyboard support right?

packages/calcite-components/src/components/carousel/carousel.tsx
```diff
@@ -0,0 +1,353 @@
...
+            // eslint-disable-next-line react/jsx-sort-props
+            ref={this.storeContainerRef}
+          >
+            <div
+              class={{
+                [CSS.itemContainer]: true,
+                [CSS.itemContainerAdvancing]: direction === "advancing",
```
driskull, line 338: flow item code detected :) 🤖

packages/calcite-components/src/components/carousel-item/carousel-item.tsx
```diff
@@ -0,0 +1,71 @@
...
+  // --------------------------------------------------------------------------
+  //
+  //  Properties
+  //
+  // --------------------------------------------------------------------------
+
+  /**
+   * When `true`, the component is active if it has a parent `calcite-carousel`.
+   */
+  @Prop({ reflect: true }) active = false;
```
driskull, line 27: should this be `selected` instead?

packages/calcite-components/src/components/carousel/carousel.scss
```diff
@@ -0,0 +1,119 @@
+:host {
+  @apply block w-full;
```
driskull, line 2: should this be block or flex?

packages/calcite-components/src/components/carousel/carousel.scss
```diff
@@ -0,0 +1,119 @@
...
+.item-container {
+  @apply focus-base
+    flex
+    items-start
+    justify-center
+    overflow-auto
+    flex-auto;
+  animation-name: none;
+  animation-duration: var(--calcite-animation-timing);
+  gap: 1rem;
```
driskull, line 35: should this use a space var?

packages/calcite-components/src/components/carousel-item/carousel-item.scss
```diff
@@ -0,0 +1,13 @@
+:host {
+  display: none;
```
driskull, line 2: Instead of display: none on the host could we do it on an internal container so we're not susceptible to any user styles overriding this? Then the default display could be flex or block.

packages/calcite-components/src/components/carousel/carousel.tsx
driskull: This would be the first component that uses a tooltip internally. Its been our pattern before to use a native `title` attribute for tooltips and let a user supply tooltips with slots if they need to. Should we discuss this?

packages/calcite-components/src/components/carousel/carousel.tsx
driskull: Shouldn't keyboard nav work on any element within the component so you don't have to be on the container? Seems like we should support bubbling events here to allow for navigation. Then if a component doesn't want the functionality they preventdefault or stop bubbling.

driskull (changes requested):
This is awesome @macandcheese nice work!!! ⭐

packages/calcite-components/src/components/carousel/carousel.tsx
driskull: From the docs, rotating is optional.

image

packages/calcite-components/src/components/carousel/carousel.tsx
SkyeSeitz: 😮😮 that is good to know. that sounds like the more consistent direction to go?

ashetland: Nice catch. Agree with changing this to `title` attribute. Might also make sense to remove `tooltipsDisabled` too, like @macandcheese suggested, in this case.

packages/calcite-components/src/components/carousel/carousel.tsx
driskull: good question for @jcfranco. We should establish a pattern here. 

I know for things like modal, you can be anywhere in the component and esc will close it. I would expect this to work the same way.

packages/calcite-components/src/components/carousel/carousel.tsx
driskull: Its still missing the `aria-roledescription="carousel"` and `section` elements.

geospatialem: Ah, _believe_ the reason AT isn't getting on-demand live updates when the `action` is selected is there should also be a unique `aria-controls` value linking to the `carousel-item`. 

For instance:

```html
<calcite-action aria-controls="unique-carousel-id-123xyz" />
<calcite-carousel-item id="unique-carousel-id-123xyz" />
```

packages/calcite-components/src/components/carousel-item/carousel-item.tsx
```diff
@@ -60,8 +63,9 @@ export class CarouselItem implements LoadableComponent {
   // --------------------------------------------------------------------------
 
   render(): VNode {
+    const id = this.el.id || this.guid;
```
driskull, line 66: I know we do this in other components but I have a question. Do we watch the `id` for changes? not sure why an id would change, but if it does then this would break right? Because we don't re-render based on the updated id changing. This may not be a realistic scenario but something we should consider. I'm sure other components would have the same issue.

packages/calcite-components/src/components/carousel-item/carousel-item.tsx
```diff
@@ -0,0 +1,79 @@
...
+  //
+  //  Render Methods
+  //
+  // --------------------------------------------------------------------------
+
+  render(): VNode {
+    const id = this.el.id || this.guid;
+    return (
+      <Host id={id}>
+        <div
```
driskull, line 69: Should this use a `<section>` element? https://www.w3.org/WAI/ARIA/apg/patterns/carousel/examples/carousel-1-prev-next/

I think it would be more semantic if it did.

driskull (commented):
Couple comments 👍

packages/calcite-components/src/components/carousel-item/carousel-item.tsx
driskull: Yeah, its never come up but it could at some point.

packages/calcite-components/src/components/carousel-item/carousel-item.tsx
driskull: Yeah, using a section element as the correct container would be nice.

driskull: Yeah I would follow the examples and just have a section at the carousel level.

packages/calcite-components/src/components/carousel-item/carousel-item.tsx
driskull: Looking again, this should be fine since the id is only set on host. If a user changes it on the host nothing needs to happen.

geospatialem (commented):
This is working great with AT and sounds fantastic with JAWS! 🦈✨ 

Observed there isn't context provided when the carousel contains `rotation` and `rotating`, but that doesn't seem like a showstopper - lmk if we should create a follow-up issue to address.

The animations also look good, but curious if we want to support reduced or disabled animations with the component if `rotation` and `rotating` are present? Think it's ok as-is, but if a shorter `duration` is specified, maybe we should add in a doc note for devs regarding a11y? WDYT?

packages/calcite-components/src/components/carousel/interfaces.ts
```diff
@@ -0,0 +1 @@
+export type ArrowType = "inline" | "edges" | "none";
```
jcfranco, line 1: For consistency, can we use `edge`? This will make all types singular.

jcfranco (approved):
🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠
🎠🎪🎪🎪🎪🎠🎪🎠🎠🎠🎠🎠🎪🎠🎪🎪🎪🎪🎠🎠🎪🎪🎪🎠🎠🎪🎪🎠🎠🎪🎠🎠🎠🎪🎠🎪🎪🎪🎪🎠🎪🎠
🎠🎪🎠🎠🎪🎠🎪🎠🎠🎠🎠🎠🎪🎠🎪🎠🎠🎠🎠🎪🎠🎠🎠🎠🎪🎠🎠🎪🎠🎪🎪🎠🎪🎪🎠🎪🎠🎠🎠🎠🎪🎠
🎠🎪🎪🎪🎪🎠🎪🎠🎠🎪🎠🎠🎪🎠🎪🎪🎪🎠🎠🎠🎪🎪🎠🎠🎪🎠🎠🎪🎠🎪🎠🎪🎠🎪🎠🎪🎪🎪🎠🎠🎪🎠
🎠🎪🎠🎠🎪🎠🎪🎠🎪🎠🎪🎠🎪🎠🎪🎠🎠🎠🎠🎠🎠🎠🎪🎠🎪🎠🎠🎪🎠🎪🎠🎠🎠🎪🎠🎪🎠🎠🎠🎠🎠🎠
🎠🎪🎠🎠🎪🎠🎠🎪🎠🎠🎠🎪🎠🎠🎪🎪🎪🎪🎠🎪🎪🎪🎠🎠🎠🎪🎪🎠🎠🎪🎠🎠🎠🎪🎠🎪🎪🎪🎪🎠🎪🎠
🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠

The most pressing items are:

1. event name simplification
2. clear interval bug on `disconnectedCallback`
3. omitting component CSS props until https://github.com/Esri/calcite-design-system/pull/8594 lands.

The rest can be tackled in a follow-up PR.

jcfranco (approved):
🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠
🎠🎪🎪🎪🎪🎠🎪🎠🎠🎠🎠🎠🎪🎠🎪🎪🎪🎪🎠🎠🎪🎪🎪🎠🎠🎪🎪🎠🎠🎪🎠🎠🎠🎪🎠🎪🎪🎪🎪🎠🎪🎠
🎠🎪🎠🎠🎪🎠🎪🎠🎠🎠🎠🎠🎪🎠🎪🎠🎠🎠🎠🎪🎠🎠🎠🎠🎪🎠🎠🎪🎠🎪🎪🎠🎪🎪🎠🎪🎠🎠🎠🎠🎪🎠
🎠🎪🎪🎪🎪🎠🎪🎠🎠🎪🎠🎠🎪🎠🎪🎪🎪🎠🎠🎠🎪🎪🎠🎠🎪🎠🎠🎪🎠🎪🎠🎪🎠🎪🎠🎪🎪🎪🎠🎠🎪🎠
🎠🎪🎠🎠🎪🎠🎪🎠🎪🎠🎪🎠🎪🎠🎪🎠🎠🎠🎠🎠🎠🎠🎪🎠🎪🎠🎠🎪🎠🎪🎠🎠🎠🎪🎠🎪🎠🎠🎠🎠🎠🎠
🎠🎪🎠🎠🎪🎠🎠🎪🎠🎠🎠🎪🎠🎠🎪🎪🎪🎪🎠🎪🎪🎪🎠🎠🎠🎪🎪🎠🎠🎪🎠🎠🎠🎪🎠🎪🎪🎪🎪🎠🎪🎠
🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠🎠

The most pressing items are:

1. event name simplification
2. clear interval bug on `disconnectedCallback`
3. omitting component CSS props until https://github.com/Esri/calcite-design-system/pull/8594 lands.

The rest can be tackled in a follow-up PR.

packages/calcite-components/src/components/carousel-item/carousel-item.scss
```diff
@@ -0,0 +1,14 @@
+:host {
+  @apply flex;
+}
+
+:host([selected]) .container {
```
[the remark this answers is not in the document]
jcfranco: Let's move forward with this since there's already precedent. We can discuss hidden-by-default rules and make any adjustments in 3.x or 4.x. Impact might be low since item/child components aren't expected to be used standalone.

packages/calcite-components/src/components/carousel/carousel.tsx
```diff
@@ -0,0 +1,638 @@
...
+  /** Fires when the carousel rotation state resumes. */
+  @Event({ cancelable: false }) calciteCarouselRotatingResume: EventEmitter<void>;
+
+  // --------------------------------------------------------------------------
+  //
+  //  Private Methods
+  //
+  // --------------------------------------------------------------------------
+
+  private updateItems = (event: Event): void => {
```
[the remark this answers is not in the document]
jcfranco: Sorry, I should've added more context. This is used as an event handler on the default `slot`, so this should be named `onSlotChange` or `handleSlotChange`.

packages/calcite-components/src/components/carousel/carousel.scss
```diff
@@ -0,0 +1,240 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-carousel-slide-item-background-color-active: Specifies the background color of the component's slide item while active.
```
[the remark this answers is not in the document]
jcfranco: Let's rename to be internal to avoid potential breakage from publicly-named props.

packages/calcite-components/src/components/carousel/carousel.e2e.ts
```diff
@@ -0,0 +1,606 @@
...
+      await page.setContent(
+        `<calcite-carousel label="Carousel example"><calcite-carousel-item label="Slide 1"><p>slide content</p></calcite-carousel-item><calcite-carousel-item label="Slide 2"><p>slide content</p></calcite-carousel-item></calcite-carousel>`,
+      );
+      const carousel = await page.find("calcite-carousel");
+      expect(carousel).not.toBeNull();
+      const isVisible = await carousel.isVisible();
+      expect(isVisible).toBe(true);
+    });
+
+    it("should pre-select the correct item if the selected attribute is set", async () => {
```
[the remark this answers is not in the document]
jcfranco: I think we can drop this test since it is somewhat testing Stencil/native browser behavior.

This may apply to other tests.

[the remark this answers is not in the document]
jcfranco: I think we can drop this test since it is sort of testing Stencil/native browser behavior.

This may apply to other tests.

packages/calcite-components/src/components/carousel/carousel.e2e.ts
```diff
@@ -0,0 +1,804 @@
...
+      expect(selectedItem.id).toEqual("two");
+
+      await slide2Button.click();
+      await page.waitForChanges();
+      expect(changeSpy).toHaveReceivedEventTimes(3);
+      selectedItem = await carousel.find(`calcite-carousel-item[selected]`);
+      expect(selectedItem.id).toEqual("two");
+    });
+
+    it("emit autoplay change event but not carousel change on autoplay button click", async () => {
```
jcfranco, line 232: Nitpick: might be helpful to have all autoplay related tests under an autoplay describe block.

packages/calcite-components/src/demos/meter.html
```diff
@@ -817,7 +817,7 @@ <h5>valueLabel</h5>
         const switchvalueLabel = document.getElementById("switch-label");
         const switchStepLabels = document.getElementById("switch-step-labels");
 
-        const controlType = document.getElementById("control-type");
```
jcfranco, line 820: Nitpick: this seems unrelated to carousel PR.

packages/calcite-components/src/components/carousel/carousel.tsx
```diff
@@ -0,0 +1,656 @@
...
+
+  /**
+   * Accessible name for the component.
+   */
+  @Prop() label!: string;
+
+  /**
+   *  When `autoplay` is `true`, specifies in milliseconds the length of time to display each Carousel Item.
+   */
+  @Prop({ reflect: true }) autoplayDuration = DURATION;
```
jcfranco, line 100: This prop is not following alphabetical order. 

This can be done in a follow-up PR.

packages/calcite-components/src/components/carousel/carousel.tsx
```diff
@@ -0,0 +1,656 @@
...
+  onMessagesChange(): void {
+    /* wired up by t9n util */
+  }
+
+  /**
+   * Made into a prop for testing purposes only
+   *
+   * @internal
+   */
+  // eslint-disable-next-line @stencil-community/strict-mutable -- updated by t9n module
```
jcfranco, line 126: Is this line needed? The explainer is related to `t9n`.

packages/calcite-components/src/components/carousel/carousel.tsx
```diff
@@ -0,0 +1,656 @@
...
+
+  disconnectedCallback(): void {
+    disconnectInteractive(this);
+    disconnectLocalized(this);
+    disconnectMessages(this);
+    this.clearIntervals();
+  }
+
+  async componentWillLoad(): Promise<void> {
+    if ((this.autoplay === "" || this.autoplay) && this.autoplay !== "paused") {
```
jcfranco, line 164: Can we add an explainer similar to [this](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/link/link.tsx#L122-L123)?

packages/calcite-components/src/components/carousel/carousel.tsx
jcfranco: > good question for @jcfranco. We should establish a pattern here.

Sorry, I missed this. What was the question? 😅

jcfranco (approved):
✨🎠✨

Issue #6482, opened 2023-02-15
[Stepper] Focus outline appears slowly


macandcheese:
### Actual Behavior

The Stepper Item focus outline appears slowly and makes the component appear sluggish, or broken: 

https://user-images.githubusercontent.com/4733155/219099845-05e26769-0e5a-4b38-b54c-bdcb9b388e24.mov

### Expected Behavior

I'd expect the focus outline to appear without transition, as in other components

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/MWqYjOb?editors=1000

### Reproduction Steps

1. Go to Codepen
2. Use mouse or keyboard to navigate through the outline
3. See the slow transition of focus outline

### Reproduction Version

1.0.5

### Esri team

Calcite (design)

DitwanP:
🍬 Verified on `2.7.2-next.5`

PR #8996, opened 2024-03-26
fix(stepper-item): remove delay in highlighting item

josercarcamo:
**Related Issue:** #6482

## Summary
Removed delay in highlighting a stepper item.


driskull (approved):
This looks good. I wonder if the animation is for something? We could try adjusting the animation to only apply to whatever the animation is for instead of "all" which it likely is doing now.

ashetland:
I took a looky look and I'm okay with losing the transition effect. The fade-in was pretty subtle and removing it makes this feel more snappy.

ashetland:
Just noticed that in `layout="vertical"` the transition effect looks like it's still there.

SkyeSeitz:
> This does remove the transition effect for the step bar highlights on hover. I'm not opposed to this change but let's run it by @ashetland and @SkyeSeitz. We can also maintain that and do what @driskull mentioned and scope the removal to just what affects the focus-ring outline.

I can't say I see a particular reason to remove the highlight bar animation as long as the focus ring effect gets adjusted. It is subtle enough, however, that we are not losing much if it goes away.

jcfranco:
Does anyone recall what `duration-150` is supposed to target? That util is applied, but there's no explicit transition prop, so all props participate in the transition (see [example](https://codepen.io/jcfranco/pen/bGJoMRQ)).

SkyeSeitz:
I think we are safe to move forward with removing the highlight bar animation. Thanks, y'all 👍🏽

jcfranco:
@macandcheese @SkyeSeitz Thanks for confirming. I think this means we can drop all transitions (duration) from `stepper-item` then. We can do this in a separate PR.

Issue #7014, opened 2023-05-24
Calcite Input Number not showing value


previnWong:
### Actual Behavior

I have a variable assigned to the value parameter (10).  On load, it shows a blank.  but if you press the arrow to increase/decrease, it either goes 9 or 11 so it respects the default value.
![image](https://github.com/Esri/calcite-components/assets/11183838/f05852fd-0d92-4cba-899d-92a9cc90fd24)
![image](https://github.com/Esri/calcite-components/assets/11183838/b315f0e3-8d01-4e08-b380-b5b8764f266f)
![image](https://github.com/Esri/calcite-components/assets/11183838/8d608ecd-8d34-4bdc-bae3-05aebdd2a4d9)

### Expected Behavior

Once "result area" is toggled, the Input Number should show 10 in the text area.

### Reproduction Sample

~https://codepen.io/pwongEsri/pen/mdzvBeK?editors=1000~
https://codepen.io/maxpatiiuk/pen/bGOxqaZ?editors=1000

### Reproduction Steps

~Repo app~
~https://codepen.io/pwongEsri/pen/mdzvBeK?editors=1000~
~please ask for login~
 
~1) select "downstream" trace, click any asset on the map, and run~
~2) Enabled the "result area" toggle, notice input number is blank.~

Note, the issue only happens when calcite-input-number is rendered by Maquette via a JS API widget, rather than directly in the browser.

### Reproduction Version

JS SDK 4.27

### Priority impact

p3 - want for upcoming milestone

### Esri team

ArcGIS Solutions

jcfranco:
@previnWong I'm having a hard time reproducing that issue in isolation. Is there any additional info you could share about the issue (e.g., the state of the component before its value is updated)?

previnWong:
When first rendered, the variable storing the value is already set.
![image](https://github.com/Esri/calcite-components/assets/11183838/2f0cf2fd-9a40-4e06-98b1-0baff2d8afe5)

Input Number is in shadow root
![image](https://github.com/Esri/calcite-components/assets/11183838/b243aee7-d178-4e5d-b6de-62fa3e4e4160)

maxxxxxdlp:
I have the same problem. If you are still having issues reproducing this, I can provide a codepen sample

JonUihlein:
[...]

PR #8997, opened 2024-03-26
fix(input, input-number, input-text): ensure values are initialized properly for dist and components output targets

jcfranco:
**Related Issue:** #7014 

## Summary

Moves value initialization logic to `componentWillLoad` for consistent behavior for both `dist` and `components` output targets.

**Note**: no additional tests were added since our testing setup uses the `dist` output target.


benelan (approved):
Nice!

PR #9003, opened 2024-03-26
fix(modal): fix scrim background color prop resolution

jcfranco:
**Related Issue:** #7180

## Summary

✨🔨✨


jcfranco:
Merging to fix syntax error. LMK if anything needs follow up.

PR #9011, opened 2024-03-27
ci: prevent outdated types from blocking release

benelan:
**Related Issue:** #9008

## Summary

Commit changes to `components.d.ts` after building for release to
prevent lerna/npm publishing errors due to an unclean working tree. The
commit will be discarded once the workflow run's container is destroyed.

### Background

The `components.d.ts` file is tracked as a temporary workaround for
https://github.com/ionic-team/stencil/issues/3239

To keep the types updated, `components.d.ts` is added to the index in a
pre-commit hook.

https://github.com/Esri/calcite-design-system/blob/2bb1ee24bb317c95e2bef8165c75583fe98ad714/.husky/pre-commit#L7-L13

However, the file can still be outdated if people make changes and don't
rebuild before creating another commit. This is mostly an issue with
small doc updates, like the PR linked above. If that happens, the
release will fail due to an unclean working tree.

### DIFF
--- a/.github/workflows/deploy-latest.yml
+++ b/.github/workflows/deploy-latest.yml
@@ -36,6 +36,13 @@ jobs:
         run: |
           npm install
           npm run build
+
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+          git config --global user.name "github-actions[bot]"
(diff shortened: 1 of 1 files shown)


driskull (approved):
👍 Nice!

Issue #7923, opened 2023-10-02
Combobox, truncate long single-select values


ashetland:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

When in `selection-mode="single"` Combobox should truncate long values. This would also help Calcite Input Time Zone where the item labels can be quite long.

Truncated long values should use `title attribute` to display the full text. This would leave `calcite-tooltip` available for additional context and configurations.

### Acceptance Criteria

See [Figma file](https://www.figma.com/file/Mn7qNHTsh2ey9mgXj8C2PT/Combobox%2C-truncate-long-single-select-values?type=design&node-id=501-21371&mode=dev). Values populated by long item labels are truncated based on available width inside the input field.

### Relevant Info

cc @jcfranco 

### Which Component

Combobox, Calcite Input Time Zone

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

eriklharper:
> Truncated long values should use html label to display the full text. This would leave calcite-tooltip available for additional context and configurations.

Curious about the "why" behind this?  `<label>`s are normally used to give names to form fields, not be used for arbitrary pieces of text, in this case being the name of the selected item inside a combobox.

eriklharper:
Also, I left a note in the Figma file, but the linked Codepen is using a very old version of Calcite (1.8.0).  As of the current version (2.7 at the time of writing this) truncation is present since this was added as part of the new `selection-display` property introduced in 1.11: https://github.com/Esri/calcite-design-system/pull/7912

ashetland:
> > Truncated long values should use html label to display the full text. This would leave calcite-tooltip available for additional context and configurations.
> 
> Curious about the "why" behind this? `<label>`s are normally used to give names to form fields, not be used for arbitrary pieces of text, in this case being the name of the selected item inside a combobox.

Womp womp, updated this to `title attribute`

ashetland:
> Also, I left a note in the Figma file, but the linked Codepen is using a very old version of Calcite (1.8.0). As of the current version (2.7 at the time of writing this) truncation is present since this was added as part of the new `selection-display` property introduced in 1.11: #7912

Yep, the only thing left is the `title` attribute

ashetland:
[...]

PR #9014, opened 2024-03-28
feat(combobox): display overflowing selected content in a tooltip

Elijbet:
**Related Issue:** #7923

## Summary
For text that has been truncated in `selection-mode="single"`, apply `title attribute` conditionally to reveal full text for sighted users. This frees up the `calcite-tooltip` to provide additional context.

Added an e2e test to ensure the long text is truncated and the title attribute is added.


packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1569,8 +1597,6 @@ export class Combobox
           onInput={this.inputHandler}
           placeholder={placeholder}
           type="text"
-          // eslint-disable-next-line react/jsx-sort-props -- ref should be last so node attrs/props are in sync (see https://github.com/Esri/calcite-design-system/pull/6530)
-          ref={(el) => (this.textInput = el as HTMLInputElement)}
```
driskull, line 1573: I think this is what is breaking tests. 

This needs to remain. See where `textInput` is being used to get the value of the input element. example: `clearInputValue()`

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -463,6 +463,7 @@ export class Combobox
     afterConnectDefaultValueSet(this, this.getValue());
     connectFloatingUI(this, this.referenceEl, this.floatingEl);
     setComponentLoaded(this);
+    this.setTooltipText();
```
driskull, line 466: I think this could probably go where we setup the resizeObserver for this el.

driskull (commented):
Initial comments.

driskull:
Here's the changes I was working with:

```patch
diff --git forkSrcPrefix/packages/calcite-components/src/components/combobox/combobox.tsx forkDstPrefix/packages/calcite-components/src/components/combobox/combobox.tsx
index e50420dfe42d5070d16a9c74f3499a97871bcc9c..cef51f93972f577d82d8e99e67a831983f81c9ae 100644
--- forkSrcPrefix/packages/calcite-components/src/components/combobox/combobox.tsx
+++ forkDstPrefix/packages/calcite-components/src/components/combobox/combobox.tsx
@@ -463,7 +463,6 @@ export class Combobox
     afterConnectDefaultValueSet(this, this.getValue());
     connectFloatingUI(this, this.referenceEl, this.floatingEl);
     setComponentLoaded(this);
-    this.setTooltipText();
   }
 
   componentDidRender(): void {
@@ -774,10 +773,13 @@ export class Combobox
 
   private setTooltipText = (): void => {
     const { textLabelEl } = this;
-    if (textLabelEl) {
-      this.tooltipText =
-        textLabelEl.offsetWidth < textLabelEl.scrollWidth ? this.textLabelEl.innerText : null;
+
+    if (!textLabelEl) {
+      return;
     }
+
+    this.tooltipText =
+      textLabelEl.offsetWidth < textLabelEl.scrollWidth ? textLabelEl.innerText : null;
   };
 
   private setTextInputWrapEl = (el: HTMLInputElement): void => {
@@ -785,6 +787,7 @@ export class Combobox
 
     if (el) {
       this.resizeObserver?.observe(el);
+      this.setTooltipText();
     }
   };
 
@@ -1551,7 +1554,7 @@ export class Combobox
   }
 
   private renderInput(): VNode {
-    const { guid, disabled, placeholder, selectionMode, selectedItems, open } = this;
+    const { guid, disabled, placeholder, selectionMode, selectedItems, open, tooltipText } = this;
     const single = isSingleLike(selectionMode);
     const selectedItem = selectedItems[0];
     const showLabel = !open && single && !!selectedItem;
@@ -1559,10 +1562,10 @@ export class Combobox
     return (
       <span
         class={{
-          "input-wrap": true,
-          "input-wrap--single": single,
+          [CSS.inputWrap]: true,
+          [CSS.inputWrapSingle]: single,
         }}
-        title={this.tooltipText}
+        title={tooltipText}
         // eslint-disable-next-line react/jsx-sort-props -- ref should be last so node attrs/props are in sync (see https://github.com/Esri/calcite-design-system/pull/6530)
         ref={this.setTextInputWrapEl}
       >
@@ -1597,6 +1600,8 @@ export class Combobox
           onInput={this.inputHandler}
           placeholder={placeholder}
           type="text"
+          // eslint-disable-next-line react/jsx-sort-props -- ref should be last so node attrs/props are in sync (see https://github.com/Esri/calcite-design-system/pull/6530)
+          ref={(el) => (this.textInput = el as HTMLInputElement)}
         />
       </span>
     );
diff --git forkSrcPrefix/packages/calcite-components/src/components/combobox/resources.ts forkDstPrefix/packages/calcite-components/src/components/combobox/resources.ts
index d4efe67ab8c2ae47dea2dca7145be9d2a6b683c4..e5e972a45f665a8984857f7b3863a0a80c368776 100644
--- forkSrcPrefix/packages/calcite-components/src/components/combobox/resources.ts
+++ forkDstPrefix/packages/calcite-components/src/components/combobox/resources.ts
@@ -5,6 +5,7 @@ export const ComboboxChildSelector = `${ComboboxItem}, ${ComboboxItemGroup}`;
 export const CSS = {
   chipInvisible: "chip--invisible",
   inputWrap: "input-wrap",
+  inputWrapSingle: "input-wrap--single",
   selectionDisplayFit: "selection-display-fit",
   selectionDisplaySingle: "selection-display-single",
   listContainer: "list-container",
diff --git forkSrcPrefix/packages/calcite-components/src/components/combobox/combobox.e2e.ts forkDstPrefix/packages/calcite-components/src/components/combobox/combobox.e2e.ts
index 4048a7551c04aba976951ec5dfe6eeda2fa46c85..b761c643bd6cd00f07d1fec4f09db94c52e8d3a8 100644
--- forkSrcPrefix/packages/calcite-components/src/components/combobox/combobox.e2e.ts
+++ forkDstPrefix/packages/calcite-components/src/components/combobox/combobox.e2e.ts
@@ -2010,7 +2010,7 @@ describe("calcite-combobox", () => {
     await page.setContent(html`
       <calcite-combobox open placeholder="Select a field" selection-mode="single">
-        <calcite-combobox-item value="NaturalResources" text-label=${longValue}></calcite-combobox-item>
+        <calcite-combobox-item selected value="NaturalResources" text-label=${longValue}></calcite-combobox-item>
         <calcite-combobox-item value="Agriculture" text-label="Agriculture"></calcite-combobox-item>
         <calcite-combobox-item value="Forestry" text-label="Forestry"></calcite-combobox-item>
         <calcite-combobox-item value="Mining" text-label="Mining"></calcite-combobox-item>
@@ -2019,16 +2019,12 @@ describe("calcite-combobox", () => {
     `);
 
     await page.waitForChanges();
-    const combobox = await page.find("calcite-combobox");
-    const firstItem = await combobox.find("calcite-combobox-item[value=NaturalResources]");
 
-    console.log("firstItem", firstItem);
-    await firstItem.click();
+    const title = await page.$eval(`calcite-combobox >>> .${CSS.inputWrap}`, (el) => el.getAttribute("title"));
 
-    const inputWrap = await page.find(`calcite-combobox >>> .${CSS.inputWrap}`);
-    expect(inputWrap).toHaveAttribute("title");
+    const textContent = await page.$eval(`calcite-combobox >>> .${CSS.inputWrap}`, (el) => el.textContent);
 
-    expect(inputWrap.textContent.length).toBeLessThan(longValue.length);
-    expect(inputWrap.getAttribute("title")).toEqual(longValue);
+    expect(textContent.length).toBeLessThan(longValue.length);
+    expect(title).toEqual(longValue);
   });
 });
```

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -785,6 +785,14 @@ export class Combobox
       textLabelEl.offsetWidth < textLabelEl.scrollWidth ? textLabelEl.innerText : null;
   };
 
+  private setTextLabelEl = (el: HTMLSpanElement): void => {
+    this.textLabelEl = el;
+
+    if (el) {
```
driskull, line 791: I think el should always be true here.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -557,6 +561,13 @@ export class Combobox
     this.refreshSelectionDisplay();
   });
 
+  private textLabelElResizeObserver = createObserver("resize", () => {
+    this.setTooltipText();
```
driskull, line 565:
```suggestion
private textLabelElResizeObserver = createObserver("resize",  this.setTooltipText);
```

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -764,6 +775,33 @@ export class Combobox
...
+
+    if (el) {
+      this.setTooltipText();
+    }
+  };
+
+  private setTextInputWrapEl = (el: HTMLInputElement): void => {
+    this.textInput = el;
+
+    if (el) {
```
driskull, line 799: I think el should always be true here.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1097,6 +1135,7 @@ export class Combobox
       this.updateActiveItemIndex(-1);
       this.resetText();
       this.filterItems("");
+      this.setTooltipText();
```
driskull, line 1138: Do you know why it needs to be set before opening?

driskull (commented):
comments :)

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -764,6 +771,31 @@ export class Combobox
...
 
+  private setTooltipText = (): void => {
+    const { textLabelEl } = this;
+    if (!textLabelEl) {
+      return;
+    }
+
+    requestAnimationFrame(() => {
+      this.tooltipText =
+        textLabelEl.offsetWidth < textLabelEl.scrollWidth ? textLabelEl.innerText : null;
```
eriklharper, line 782: I would compare the actual text width using [this util](https://github.com/Esri/calcite-design-system/blob/34896cbd6eb2600823446f5f0e7b17bb2d42ba25/packages/calcite-components/src/utils/dom.ts#L126-L134) with the computed width of the text's container.  Also, rather than making a separate `@State` property for the tooltip text, just use the selected item's display value and in the render method, just conditionally render the `title` attribute.

driskull: I don't think that util makes sense in this scenario. We just want to know if the offsetWidth is less than the scrollWidth which is simpler than creating a canvas and measuring.

I do agree that the `@State()`isn't ideal. It could probably just be a boolean for whether to render the title or not and use the selected items display value. I think it still needs a state because the render method wouldn't update when the resize observer fires unless it tells some state to change.

driskull: Note: Button seems to do the same logic so a util that handles this might be useful. Something like `hasScrollbar(el);`

https://github.com/Esri/calcite-design-system/blob/210e2da30946664a803bdb2cb67d2e83b6fc7532/packages/calcite-components/src/components/button/button.tsx#L414-L420

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -531,6 +532,9 @@ export class Combobox
 
   @State() text = "";
 
+  /** keeps track of the tooltipText */
+  @State() tooltipText: string;
```
driskull, line 536: maybe this should just change to `@State() showTooltipText = false;`

No need to store the string value since its just the selected item's display value.

eriklharper: Agreed.  Storing the value is duplicative.  You should be able to get the selected item's value with the element reference in the ResizeObserver callback function.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -764,6 +772,22 @@ export class Combobox
...
+      this.tooltipText =
+        textLabelEl.offsetWidth < textLabelEl.scrollWidth ? this.textLabelEl.innerText : null;
+    }
+  };
+
+  private setTextInputWrapEl = (el: HTMLInputElement): void => {
+    this.textInput = el;
+
+    if (el) {
+      this.resizeObserver?.observe(el);
```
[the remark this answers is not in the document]
eriklharper: Not sure if we should add a 2nd ResizeObserver specifically for that child element when we could just tag-along with the existing one.  I've heard that its not advisable where possible to attach more than 1 unnecessarily: https://github.com/WICG/resize-observer/issues/59#issuecomment-408098151

packages/calcite-components/src/components/combobox/combobox.tsx
driskull: Yeah I thought that too. IMO that would be the simplest option. It is a tiny bit odd to conditionally render a title attribute.

packages/calcite-components/src/components/combobox/combobox.tsx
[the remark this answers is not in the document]
driskull: Yeah that makes sense. If we only have 1 resize observer, we can look at the entry in the callback to see which logic to perform. cc @Elijbet

packages/calcite-components/src/components/combobox/combobox.tsx
eriklharper: > I also thought, why not just always render the full `title` attribute? Is the intention to not unnecessarily crowd the UI when hovering the selected item with the mouse if the full text is visible? Either way, rendering the title all the time would just be the simplest approach, no need to calculate overflow whatsoever... Just some food for thought.

@ashetland what do you think of just always rendering the `title` attribute?

ashetland: Honestly not sure why this hasn't come up before 😆. I don't see any problems with that. I think there are a few other places we're using the `title` attribute like this, but I'm not sure if any of them are conditional. Seems like we might be able to always render `title` everywhere we do this.

Elijbet: This has come up before in the button truncation. We had a full discussion with @Aaron and @SkyeSeitz about it and there was a small visual to illustrate the crowding. The gist of it was:

To avoid unnecessary noise, detect the buttons with text that has been truncated and apply title attribute conditionally only to these, to reveal the full text for sighted users. This frees up the calcite-tooltip to provide additional context.

https://github.com/Esri/calcite-design-system/pull/6664#issuecomment-1485643872

driskull: Button was the only other place we seem to be doing this conditional logic. It would be much simpler and easier on performance if we just always set the title attribute when we are potentially truncating.

Elijbet: I think this might have come up with tab-title truncation as well.

ashetland: Thanks @Elijbet for the refresher! If we play this out to possible truncation behavior in places like List or Tree, I agree that keeping this conditional is a cleaner UX for the end user. If the text is short enough to be displayed then there isn't much reason to also render the `title` attribute. Add a possible `calcite-tooltip` on top of that and there's suddenly a lot going on. Reducing redundancy feels like a win.

ashetland: We've also been talking about mapping out a system-wide configurable `text-overflow`. With that in mind and since, in this case, the entire text can be read by expanding the Combobox, I want to put it out there that we could abandon this for the time being.

Elijbet: Ok, that makes sense to me.

Elijbet: Closing the issue then.

Issue #9010, opened 2024-03-27
Calcite tabs broken with React + Vite


mpayson:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Calcite tabs don't load and there is an error in the console `Uncaught DOMException: Operation is not supported`

### Expected Behavior

Calcite tabs load

### Reproduction Sample

https://github.com/mpayson/calcite-debug

### Reproduction Steps

1. Clone repo
2. `npm install`
3. `npm run dev`
4. Open page, see tabs don't load

### Reproduction Version

2.7

### Regression?

Yes

### Priority impact

p1 - need for current milestone

### Impact

Can't upgrade

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [X] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Online

DitwanP:
🍡 Verified with provided sample repo on `2.7.1`

PR #9021, opened 2024-03-29
fix(tab-nav): fix custom element creation error in `components` build output

jcfranco:
**Related Issue:** #9010 

## Summary

Class members are defined on the component host in the `components` output build, so we need to avoid using names of  `HTMLElement` members (`dir` in this case).

**Note**: no additional tests were added since our testing setup uses the `dist` output target.


jcfranco:
Sidebar: we could look into improving [`reserved-member-names`](https://github.com/stencil-community/stencil-eslint/blob/main/docs/reserved-member-names.md) to catch this.

driskull:
> Sidebar: we could look into improving [reserved-member-names](https://github.com/stencil-community/stencil-eslint/blob/main/docs/reserved-member-names.md) to catch this.

Yeah thats a good call 👍

driskull (approved):
👍

PR #9024, opened 2024-03-30
chore(actions): extract label values and utils to support module

jcfranco:
**Related Issue:** N/A

## Summary

✨🧹✨


geospatialem (approved):
Awesome refactor! 🧹

LGTM! 💪🏻

.github/scripts/support/labels.js
```diff
@@ -0,0 +1,41 @@
+const labels = {
```
benelan, line 1: Nitpick: should we call this file `resources.js` so we can add additional objects in here like we do with the components?

benelan (approved):
Nice cleanup!

.github/scripts/support/labels.js
jcfranco: Good idea. Will rename!

PR #9038, opened 2024-04-02
fix(list, list-item): support keyboard sorting in screen readers

driskull:
**Related Issue:** #7426

## Summary

- support keyboard sorting in screen readers by setting internal role to radio
- add e2e test.
- https://stackoverflow.com/questions/52261977/why-javascript-if-keycode-enter-key-or-spacebar-key-does-not-work-with-nvda

### DIFF
--- a/packages/calcite-components/src/components/handle/handle.e2e.ts
+++ b/packages/calcite-components/src/components/handle/handle.e2e.ts
@@ -126,13 +126,21 @@ describe("calcite-handle", () => {
     t9n("calcite-handle");
   });
 
-  it("sets application role on handle", async () => {
+  it("sets radio role properly", async () => {
     const page = await newE2EPage();
     const label = "Hello World";
     await page.setContent(`<calcite-handle lang="en" label="${label}"></calcite-handle>`);
     await page.waitForChanges();
 
     const handle = await page.find("calcite-handle");
-    expect(handle.getAttribute("role")).toBe("application");
+
+    const internalHandle = await page.find(`calcite-handle >>> .${CSS.handle}`);
+    expect(internalHandle.getAttribute("role")).toBe("radio");
+    expect(internalHandle.getAttribute("aria-checked")).toBe("false");
+
+    handle.setProperty("selected", true);
+
+    await page.waitForChanges();
+    expect(internalHandle.getAttribute("aria-checked")).toBe("true");
   });
 });
--- a/packages/calcite-components/src/components/handle/handle.e2e.ts
+++ b/packages/calcite-components/src/components/handle/handle.e2e.ts
@@ -125,4 +125,14 @@ describe("calcite-handle", () => {
   describe("translation support", () => {
     t9n("calcite-handle");
   });
+
+  it("sets application role on handle", async () => {
+    const page = await newE2EPage();
+    const label = "Hello World";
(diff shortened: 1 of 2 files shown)


jcfranco:
@driskull Can you clarify which comment or link from the StackOverflow post is referenced? Is it [this](https://github.com/nvaccess/nvda/issues/7898)?

driskull:
Sorry, I was referring to this one:

https://stackoverflow.com/a/52264128

> It probably won't fire because NVDA intercepts the keys if you are in browse mode, which is normal. If you were to manually switch to focus mode (insert + space -- you'll hear a "typewriter" sound), the code should get passed through.

and this one:

https://stackoverflow.com/a/55793672

> Add role="application" in your widget html and try it, it is working for me.

Role of application does seem to fix it. I'm not sure if there are any downsides to doing that.

packages/calcite-components/src/components/handle/handle.e2e.ts
```diff
@@ -125,4 +125,14 @@ describe("calcite-handle", () => {
   describe("translation support", () => {
     t9n("calcite-handle");
   });
+
+  it("sets application role on handle", async () => {
```
jcfranco, line 129: Similar to adding context in the host, could you add a bit more detail to the test name? I think we should add more info to these types of a11y tests (role/ARIA role).

jcfranco:
Awesome, thanks!

packages/calcite-components/src/components/handle/handle.tsx
```diff
@@ -298,24 +299,26 @@ export class Handle implements LoadableComponent, T9nComponent, InteractiveCompo
...
-      </span>
+      <Host role="application">
+        <span
+          // Needs to be a span because of https://github.com/SortableJS/Sortable/issues/1486
+          aria-disabled={this.disabled ? toAriaBoolean(this.disabled) : null}
+          aria-label={this.disabled ? null : this.getAriaText("label")}
+          aria-pressed={this.disabled ? null : toAriaBoolean(this.selected)}
+          class={{ [CSS.handle]: true, [CSS.handleSelected]: !this.disabled && this.selected }}
+          onBlur={this.handleBlur}
+          onKeyDown={this.handleKeyDown}
```
jcfranco, line 310: Based on https://github.com/nvaccess/nvda/issues/7898#issuecomment-529384975, I'm curious if screen readers are triggering a click vs keydown event here. Could you team up w/ @geospatialem to test this out? If it works, it could help avoid the host role.

jcfranco (approved):
Once explainers are added, this looks good to merge!

packages/calcite-components/src/components/handle/handle.tsx
driskull: @geospatialem could you review this one?

We don't want to add a click event to this element but we could change its role if we need to. Switching from "button" to "radio" or whatever as suggested by that article.

jcfranco: Oh, that's right. There's no click handling logic here. Ignore my suggestion then. Changing the inner element's `role` makes sense.

geospatialem (approved):
Works great with both JAWS and NVDA! Nice work, @driskull! ✨

driskull:

jcfranco (approved):
Very noice! 🏆

Issue #8460, opened 2023-12-19
VoiceOver cannot enter into date range input with native controls


FelFly:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Summary

While I understand Tab is the documented keyboard navigation for this component, Tab isn't as commonly used in VoiceOver as it is in say, NVDA. When trying to use the VO keys to move forward Ctrl+Option+right-arrow, I cannot find a method using native keyboard shortcuts to manually type a value nor trigger the calendar widget. (Lacking the correct keyboard to navigate via natural reading order - I cannot say whether this is applicable to JAWS.)

### Actual Behavior

While I can get focus to an input, using the VoiceOver suggestion for Ctrl+Option+Space does not result in any observable interactions and when I try Ctrl+Shift+Down-Arrow to see if I can get inside the group, it does tell me I'm in an editable text field, but still I cannot manually type in numbers in the suggested format.

![Screenshot 2023-12-19 at 5 16 22 PM](https://github.com/Esri/calcite-design-system/assets/7389593/22744e69-8506-4f8d-9f75-35abc1c2fb95)

![Screenshot 2023-12-19 at 5 24 14 PM](https://github.com/Esri/calcite-design-system/assets/7389593/03da87ac-d440-4848-aba5-8e61628a4b32)

### Expected Behavior

I would expect that I could use the VoiceOver keys to navigate into the input and then manually type in numbers if I so wished.

### Reproduction Sample

https://codepen.io/FelFly/pen/poYzNMq

### Reproduction Steps

STR
1. Using VoiceOver navigate to form input via Ctrl+Option+Right-Arrow
2. Attempt to enter into input without use of Tab or to type manual date entry in MM/DD/YYYY format with keyboard numbers

### Reproduction Version

1.11.0

### Working W3C Example/Tutorial

https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/datepicker-dialog/

### Relevant Info

Safari: Version 17.1 (18616.2.9.11.9, 18616)
MacOS: Ventura 13.6.1 

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Hub

[...]

PR #9039, opened 2024-04-02
fix(combobox): fix aria-role for proper voiceover support

driskull:
**Related Issue:** #8460

## Summary

- Move combobox role from container wrapper to the input element
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role#examples


driskull:
@geospatialem could you give this a test?

geospatialem (commented):
This is looking good with Safari + VO and the VO commands. 🙌🏻  

We might need additional testing to ensure all is good, I'm using a Windows keyboard to check using "Alt" instead of "Option", but otherwise LGTM.

Also checked the tab keys with NVDA, JAWS and VO and ensured NVDA and JAWS still work as-intended - everything seems to be working as-intended. ✨

jcfranco (approved):
🎉🦖🌮

Issue #9044, opened 2024-04-02
Update description of headingLevel property to clarify what it's used for


DitwanP:
### Description

Need to update the description of the `headingLevel` property in order to make it more clear how and why it's used. 

### Which Component

- `block`
- `date-picker`
- `flow-item`
- `input-date-picker`
- `panel`
- `pick-list-group`
- `pick-list`
- `popover`
- `tip-manager`
- `tip`

### Resources

An explanation of what the prop actually does as well as the need for the update from Adam:

"Really headingLevel controls only the internally-rendered text container, so folks can as they need determine the heading level used. There is no visual change, so it's just for semantic page structure. Most (all?) cases default to undefined, where the internally-rendered text container is just a div, span, etc.
 
We do get questions about this (one recent in Community), expecting it to change visually, so we can update the doc to explain this better"

[Community Post](https://community.esri.com/t5/calcite-design-system-ideas/calcitepopover-add-a-header-slot/idi-p/1401959/jump-to/first-unread-message)

geospatialem:
> We do get questions about this (one recent in Community), expecting it to change visually, so we can update the doc to explain this better"

@DitwanP Is there a Community post we can link to in the issue above, so we can let folks know once this is live with our upcoming release of the change?

DitwanP:
Verified on `main`

geospatialem:
Thanks for updating these, @DitwanP and verifying! Updating the label for `4 - verified` for tracking purposes. 🚀

PR #9045, opened 2024-04-02
docs: fix the description of headingLevel property

DitwanP:
**Related Issue:** #9044 

## Summary
Updating the description of `headingLevel` prop to eliminate confusion on what it's purpose is.

### DIFF
--- a/packages/calcite-components/src/components/block/block.tsx
+++ b/packages/calcite-components/src/components/block/block.tsx
@@ -93,7 +93,7 @@ export class Block
   @Prop() heading!: string;
 
   /**
-   * Specifies the semantic heading level of the internally-rendered text container for proper document structure, without affecting visual styling.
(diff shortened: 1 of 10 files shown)


packages/calcite-components/src/components/block/block.tsx
```diff
@@ -93,7 +93,7 @@ export class Block
   @Prop() heading!: string;
 
   /**
-   * Specifies the number at which section headings should start.
+   * Specifies the semantic heading level of the internally-rendered text container for proper document structure, without affecting visual styling.
```
geospatialem, line 96: This is a bit hard to read and offers a lot of context for users to sift through. Is there a different approach we can take as an in-between for the previous language and what is proposed?

Potential suggestion below, but flexible on a different direction.

```suggestion
   * Specifies the heading level of the component's `heading` for proper document structure, without affecting visual styling.
```

geospatialem (approved):
LGTM! 👍🏻

Issue #8900, opened 2024-03-07
<calcite-action> has a different size depending on whether text is enabled (in small scale)


hcampos-professional:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

A `<calcite-action>` without `text` and with `scale="s"` is smaller than when `text` is specified but `text-enabled` is `false` (the default):

```html
<-- Text and icon, 32px tall -->
<calcite-action scale="s" icon="home" text="hello" text-enabled></calcite-action>

<-- Only icon, 32px tall -->
<calcite-action scale="s" icon="home" text="hello"></calcite-action>

<-- Only icon, 24px tall -->
<calcite-action scale="s" icon="home"></calcite-action>
```

![CleanShot 2024-03-07 at 08 59 22](https://github.com/Esri/calcite-design-system/assets/136828178/a36d280d-a111-43eb-a2ea-b54fa69b867f)

### Expected Behavior

I would expect the component to have the same height in the two examples where only the icon is shown. If the text isn't visible, the fact that the attribute is set shouldn't cause the component to get a different height.

### Reproduction Sample

https://codepen.io/hccampos/pen/PogPXrY

### Reproduction Steps

1. Open https://codepen.io/hccampos/pen/PogPXrY
2. Notice that the last row is 24px whereas the others are 32px.

### Reproduction Version

2.6.0

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

PR #9051, opened 2024-04-03
fix(action): maintain equal height when text is not enabled in a small scale

driskull:
**Related Issue:** #8900

## Summary

- Maintain equal height when text is not enabled in a small scale
- Add screenshot test

### DIFF
--- a/packages/calcite-components/src/components/action/action.stories.ts
+++ b/packages/calcite-components/src/components/action/action.stories.ts
@@ -210,6 +210,20 @@ export const indicatorNoTextEnabledNoIcon_TestOnly = (): string => html`
   <calcite-action indicator active text="click-me"></calcite-action>
 `;
 
+export const noTextHeight_TestOnly = (): string =>
+  html`<h2>All actions should be the same height</h2>
+    <div style="width: min-content">
+      <div style="border: solid 1px">
+        <calcite-action text="hello" text-enabled icon="home" scale="s"></calcite-action>
+      </div>
+      <div style="border: solid 1px">
+        <calcite-action text="hello" icon="home" scale="s"></calcite-action>
+      </div>
+      <div style="border: solid 1px">
+        <calcite-action icon="home" scale="s"></calcite-action>
+      </div>
+    </div>`;
+
 export const arabicLocale_TestOnly = (): string => html`
   <calcite-action
     dir="rtl"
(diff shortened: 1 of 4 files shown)


packages/calcite-components/src/components/action/action.tsx
```diff
@@ -299,7 +299,10 @@ export class Action
       buttonId,
       messages,
     } = this;
-    const ariaLabel = `${label || text}${indicator ? ` (${messages.indicator})` : ""}`;
+    const labelFallback = label || text;
+    const ariaLabel = labelFallback
+      ? `${labelFallback}${indicator ? ` (${messages.indicator})` : ""}`
```
jcfranco, line 304: This should use a localized template. Since this is already in production, can you create an issue? We can link to action’s doc site page for the repro case since this is a screen reader label.

packages/calcite-components/src/components/action/action.stories.ts
```diff
@@ -210,6 +210,20 @@ export const indicatorNoTextEnabledNoIcon_TestOnly = (): string => html`
   <calcite-action indicator active text="click-me"></calcite-action>
 `;
 
+export const noTextHeight_TestOnly = (): string =>
+  html`<h2>All actions should be the same height</h2>
```
jcfranco, line 214: 🏆

jcfranco (approved):
Noice!

packages/calcite-components/src/components/action/action.tsx
driskull: created https://github.com/Esri/calcite-design-system/issues/9071

ashetland:
Is this another "bug" where the fix might be seen as breaking? Seems like something we should correct, but maybe it should wait? cc @jcfranco

driskull:
Are the height differences also in Figma? I would consider this a fix if not.

SkyeSeitz:
The height differences are not present in Figma. The height of the Action remains the same regardless if text is enabled.

ashetland:
If we're okay with moving forward with this I can review the screenshots and add the `visual changes` label to the issue.

driskull:
Lets see what @jcfranco thinks. I think we can do it if we consider it a bug.

jcfranco:
Sorry for the late reply. We should treat this one as a bug/fix since the design spec has no height differences.

driskull:
@ashetland can you review and accept the screenshot tests?

Issue #9015, opened 2024-03-28
Remove component usage examples from repo


DitwanP:
### Description

Had a conversation with Adam and Franco about another issue that involved documentation changes [6312](https://github.com/Esri/calcite-design-system/issues/6312). It was determined that we should remove the usage files from the repo. The plan is to replicate the examples on the doc site if the example is not already covered by what's currently on the site. 

### Which Component

All of them

DitwanP:
Pushed this one to later milestone because it seems like there's still some discussion that's happening around it.

DitwanP:
Verified locally
<img width="277" alt="Screenshot 2024-05-21 at 11 06 30 AM" src="https://github.com/user-attachments/assets/f537b38a-f38f-4bef-a8c4-948fd2f757c5">

PR #9052, opened 2024-04-03
chore: remove component usage files

DitwanP:
**Related Issue:** #9015 

## Summary
Removing usage sample files from repo since we now have links to each components page on the doc site for more in depth information and samples.


driskull:
I'm not sure I agree with removing these at this time.

@jcfranco what about where we are lacking coverage for some of these on the doc site? For example, there is no virtual element info on popover in the doc site.

If we remove some of these usage files I'm concerned we will be losing info on how to do some things. 

IMO it would be better to leave these until we can assure we have the same coverage of information on the doc site.

Since Stencil automatically uses these files an generates documentation samples, it would be nice if we could keep these and use them to generate some of the doc site samples. That way, the samples are within the code base and can easily be referenced. A developer who is developing a component or functionality, has the ability to define usage easier if they can do so within the code base.

DitwanP:
> I'm not sure I agree with removing these at this time.
> 
> @jcfranco what about where we are lacking coverage for some of these on the doc site? For example, there is no virtual element info on popover in the doc site.
> 
> If we remove some of these usage files I'm concerned we will be losing info on how to do some things.
> 
> IMO it would be better to leave these until we can assure we have the same coverage of information on the doc site.

So the plan was to go through the components on the doc site and improve coverage, fix descriptions, and make sure things are still accurate.
 
Maybe we just leave this for now until the doc site work is done, which is what I was thinking to do initially but wasn't sure if it mattered that much. 

>Since Stencil automatically uses these files an generates documentation samples, it would be nice if we could keep these and use them to generate some of the doc site samples. That way, the samples are within the code base and can easily be referenced. A developer who is developing a component or functionality, has the ability to define usage easier if they can do so within the code base.

I'm uncertain about the advantages here. It seems like the documentation site offers more detailed examples that often encompass most usage scenarios. While the codebase contains numerous usage files with minor differences, such as adding a property or two to a component, the documentation site could likely achieve the same effect with a toggle.

jcfranco:
> what about where we are lacking coverage for some of these on the doc site? For example, there is no virtual element info on popover in the doc site.

Good point on removing usage files that don't have a counter part in the doc site. Can we come up with a list of _useful_ usage files that are missing from the doc site? I have a hunch it will be a small list. In any case, we can prioritize those items for future doc improvements.

> Maybe we just leave this for now until the doc site work is done, which is what I was thinking to do initially but wasn't sure if it mattered that much.

This seems like the best path forward. Once we know we've got proper coverage on the doc site, we should be good to remove usage files and possibly provide guidance on how to add/update doc examples that will get referenced in the doc site.

> I'm uncertain about the advantages here. It seems like the documentation site offers more detailed examples that often encompass most usage scenarios. While the codebase contains numerous usage files with minor differences, such as adding a property or two to a component, the documentation site could likely achieve the same effect with a toggle.

Agreed. For context, we are removing usage files for a couple of reasons:

1. we aim for the doc site to serve as definitive source of usage examples
2. currently, the doc site does not use/reference usage files
3. generated readmes offer incomplete documentation compared to what the doc site provides

@DitwanP @macandcheese LMK if I missed anything.

DitwanP:
The list below Is what I've found for usage samples that might need to get recreated on the doc site, not sure if they all need to so @macandcheese feel free to LMK what we should exclude if any.

- **Button**
  - Internals sample?
- **Input**
  - Native Events?
- **Label**
  - Browser Caveat?
- **Modal**
  - Before close
- **Popover**
  - Virtual
- **Sheet**
  - Before close
- **Shell Panel**
  - With custom element?
- **Tooltip**
  - Virtual

driskull:
We should add the virtual element doc for popover and tooltip.

jcfranco:
Thanks y'all!

I agree with @macandcheese and @driskull's recommendations.

DitwanP:
The usage snippets that were not previously represented on the doc site has now been added, everything else was already covered by the existing samples.

benelan:
This is going the opposite direction of what I had hoped we would do :sob:. 

It would be great if we could have the usage examples source of truth in the monorepo right next to the source code and API reference. 

When I discussed this with @macandcheese before it sounded like there were just some formatting issues/inconsistencies that prevented him from being able to parse the usage example strings and use them in the doc site. 

I think we should be standardizing the format of the usage examples, and from there I could write a simple parser to grab the examples, their descriptions, etc. for use in the doc site.

Speaking for myself, I'm way more likely to learn from and contribute to usage examples if they are in the monorepo, rather than hidden behind a VPN.

Additionally, people may be using the usage fields of docsjson for something (although we don't have it documented so removing it wouldn't be a breaking change)

https://cdn.jsdelivr.net/npm/@esri/calcite-components@2.8.3/dist/extras/docs-json.json

alisonailea (approved):

jcfranco (approved):

Issue #7075, opened 2023-05-31
[Navigation Logo] Add `heading-level` Property


macandcheese:
### Description

The `heading` text in Navigation Logo should be able to optionally render as a heading level, via a `headingLevel` property. This should not affect the visual display of the component, only the internally rendered component.

### Acceptance Criteria

This should match other instances of this `headingLevel` property, such as in Panel: https://developers.arcgis.com/calcite-design-system/components/panel/#api-reference-properties-headingLevel

### Which Component

Navigation Logo

### Example Use Case

A user may want the rendered heading to render as a semantic heading level.

### Priority impact

p4 - not time sensitive

### Esri team

Calcite (design)

DitwanP:
🍡 Verified on `main`

PR #9054, opened 2024-04-03
feat(navigation-logo): add heading level

josercarcamo:
**Related Issue:** #7075 

## Summary
Added the "headingLevel" property to navigation-logo, similar to Panel.


packages/calcite-components/src/components/navigation-logo/navigation-logo.tsx
```diff
@@ -107,34 +113,45 @@ export class CalciteNavigationLogo implements LoadableComponent {
     return <calcite-icon class={CSS.icon} flipRtl={this.iconFlipRtl} icon={this.icon} scale="l" />;
   }
 
+  renderHeaderContent(): VNode {
+    const { heading, headingLevel, description } = this;
+    const headingNode = heading ? (
+      <Heading class={CSS.heading} level={headingLevel}>
+        <span
+          aria-label={this.heading}
```
driskull, line 121: I don't think we need an aria-label here. The heading element should already provide enough information.

packages/calcite-components/src/components/navigation-logo/navigation-logo.tsx
```diff
@@ -107,34 +113,45 @@ export class CalciteNavigationLogo implements LoadableComponent {
...
+          }}
+          key={CSS.heading}
+        >
+          {heading}
+        </span>
+      </Heading>
+    ) : null;
+
+    const descriptionNode = description ? (
+      <span aria-label={this.description} class={CSS.description} key={CSS.description}>
```
driskull, line 134: Whats the need for the aria-label here as well?

driskull (changes requested):
Looking good @josercarcamo 👍

packages/calcite-components/src/components/navigation-logo/navigation-logo.tsx
josercarcamo: I was just trying to keep the same sematics as what was in the original code, but I've removed it.

packages/calcite-components/src/components/navigation-logo/navigation-logo.tsx
```diff
@@ -59,6 +60,11 @@ export class CalciteNavigationLogo implements LoadableComponent {
   /** Specifies the `src` to an image. */
   @Prop() thumbnail: string;
 
+  /**
+   * Specifies the number at which section headings should start.
```
geospatialem, line 64: @josercarcamo Very timely! @DitwanP made similar changes across the components earlier today. CC'ing him for editorial review on the `headingLevel` doc update.

anveshmekala:
Nitpick: PR title should be `feat` since this is an enhancement.

packages/calcite-components/src/components/navigation-logo/navigation-logo.tsx
```diff
@@ -107,34 +113,43 @@ export class CalciteNavigationLogo implements LoadableComponent {
...
+    ) : null;
+
+    const descriptionNode = description ? (
+      <span class={CSS.description} key={CSS.description}>
+        {description}
+      </span>
+    ) : null;
+
+    return headingNode || descriptionNode ? (
+      <div class={CSS.container}>
```
driskull, line 138: nitpick: should we give this one a key for consistency? `<div class={CSS.container} key={CSS.container}>`. up to you :)

driskull (changes requested):
Can you add an test within the e2e test that tests the headingLevel? This would be using the `reflects` common test helper. You can see an example in the `block.e2e.ts`. 

```
 describe("reflects", () => {
    reflects("calcite-block", [
      {
        propertyName: "collapsible",
        value: true,
      },
      {
        propertyName: "headingLevel",
        value: 2,
      },
```

Also a "defaults" test would be good too.

```
describe("defaults", () => {
    defaults("calcite-block", [
      {
        propertyName: "collapsible",
        defaultValue: false,
      },
      {
        propertyName: "headingLevel",
        defaultValue: undefined,
      },
```

Can you also add a storybook story test in the `navigation-logo.stories.ts` that tests with a headingLevel applied? This way we can make sure different heading levels don't render differently.

Something like:

```
export const headingLevel_TestOnly = (): string => html`
 <calcite-navigation-logo
      heading="ArcGIS Online"
      heading-level="2"
      description="City of AcmeCo"
      thumbnail="${placeholderImage({ width: 50, height: 50 })}"
    />
`;
```

driskull (approved):
Looks good to me! 👍

Issue #8242, opened 2023-11-21
List: difficulty dragging into a child list


driskull:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Its difficult to drag a list item into an empty list because the list takes up no space.

It would be nice if an empty list had some reserved space area to drop items.

### Acceptance Criteria

Easier to drag into an empty list

### Relevant Info

https://codepen.io/driskull/pen/WNPzLdK?editors=1000

![Nov-21-2023 14-34-42](https://github.com/Esri/calcite-design-system/assets/1231455/ca4afe99-9155-4d09-a270-32e13b0ff30d)

### Which Component

list

### Example Use Case

See relevant info

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

driskull:
This needs to go to design first.

sagewall:
@geospatialem would it be possible to prioritize this? It's causing come confusion and unpredictable behavior trying to drag items into group layers in the new JavaScript SDK layer list widget.

ashetland:
@driskull Didn't this get merged into v.2.0?

driskull:
I did a quick fix but this needs a proper design.

driskull:
I think this could be solved by #7100 and indenting the whole item.

sagewall:
Thought I'd mention this not only affects the JS SDK, but the applications that use our SDK like Map Viewer, Experience Builder and Instant Apps.

ashetland:
Hey @sagewall and @zaramatheson is there something that @driskull's quick fix isn't addressing? Can you show us demo of the problem? I'm working on some related List designs, but I want to make sure I/we fully understand what's causing this issue. We might be able to expedite a better fix.

sagewall:
Sure, Currently it's impossible to move an item down the list under an open group layer without adding it to the group.  If you move a layer up from further down the list, there isn't a way to add it to the group.

### Reproduction Steps

1. Open this [codepen](https://codepen.io/sagewall/pen/abMWmEB)
2. Open "Group Layer 3"
[...]

PR #9068, opened 2024-04-05
feat(list): add mode to handle flat and nested lists and improve dragging

driskull:
**Related Issue:** #8667 #8242


driskull:
@ashetland this handles the modes and indentation. Can you take a look?

PR #9069, opened 2024-04-05
chore(linting): enable `import/order` rule to collapse newlines between imports

jcfranco:
**Related Issue:** N/A

## Summary

Enables [`import/order`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md) to help take care of any newlines left between imports after updating modules.


driskull (approved):
👍

Issue #7901, opened 2023-09-28
Tree: clicking near the edges around the chevron selects the item


ashetland:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When attempting to click the chevron to expand/collpase a parent, clicking near the edges will select the parent instead of expanding/collapsing it.

![CleanShot 2023-09-28 at 14 44 43](https://github.com/Esri/calcite-design-system/assets/108549080/f816029f-88ce-4881-a3c4-e3c6c0e832f9)

### Expected Behavior

That clicking to the left, top, or bottom of the chevron would expand/collapse the parent. This would apply to all section modes that allow for selecting the parent.

### Reproduction Sample

https://codepen.io/ashetland/pen/jOXKYdV

### Reproduction Steps

Click near the edges around any of the chevrons.

### Reproduction Version

v1.8.0

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

PR #9115, opened 2024-04-11
fix(tree-item): do not select when chevron clicked

josercarcamo:
**Related Issue:** #7901 

## Summary
Clicking the chevron next to a menu item with children no longer selects the menu item.


driskull (approved):
Looks good @josercarcamo ⭐ 

I'm guessing since the padding was changed some screenshot test changes will need to be accepted is that correct? If so, that's good. 👍

packages/calcite-components/src/components/tree-item/tree-item.scss
```diff
@@ -57,14 +61,16 @@
...
-      .chevron,
       .checkmark,
       .bullet-point {
         margin-inline: theme("margin.3");
       }
       .icon-start {
         margin-inline: theme("margin.3");
       }
+      .chevron {
+        padding: 0.7rem;
```
jcfranco, line 72: Should this be using a `theme` or a spacing prop/token? cc @ashetland @driskull

driskull: Good question. cc @ashetland 👍

Issue #8490, opened 2023-12-22
bug(calcite-button): shadow dom width does not match host width


maxpatiiuk:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

If `<calcite-button scale="s">` is told to take 32px x 32px, the host element takes the desired size, but the shadow DOM's `<button>` is 24px x 32px

### Expected Behavior

If `<calcite-button>` has attribute `width="full"`, `<button>` has `width: 100%;` style applied, matching the width of the container

### Reproduction Sample

https://codepen.io/maxpatiiuk/pen/NWJWXBP?editors=1100

### Reproduction Steps

1. Open [the codepen](https://codepen.io/maxpatiiuk/pen/NWJWXBP?editors=1100)
2. See pink - should not be visible

### Reproduction Version

2.1.0

### Relevant Info

Chrome 120, macOS 14.2

### Priority impact

p4 - not time sensitive

### Impact

I need to hardcode 32px x 32px size for the button because button may either display iconStart, or have a legacy Esri font icon slotted in - in both cases, I want it to maintain the same size (it does not right now as styles for when button has slotted content are quite different)

Current workaround:
1. Still have 32px x 32px size set in CSS
2. If button has iconStart, set `scale="m"`
3. If button has slotted in legacy font icon, set `scale="s"`

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

driskull:
It seems like its the `inline-size` being set with `w-6`. here: https://github.com/Esri/calcite-design-system/blob/f199b3cafc40438c32a12cad2aef4d26be4526f6/packages/calcite-components/src/components/button/button.scss#L670C4-L670C22

DitwanP:
🍧 Verified on `2.9.0-next.4`

PR #9145, opened 2024-04-17
fix(calcite-button): Make shadow dom button 100% width

josercarcamo:
**Related Issue:** #8490

## Summary
When the host element button is set to "full" the shadow dom button is set to 100% width to span the entire space.


packages/calcite-components/src/components/button/button.scss
```diff
@@ -685,6 +685,16 @@
...
 }
+
+:host([scale="s"][width="full"]) button:not(.content--slotted),
+:host([scale="s"][width="full"]) a:not(.content--slotted),
+:host([scale="m"][width="full"]) button:not(.content--slotted),
+:host([scale="m"][width="full"]) a:not(.content--slotted),
+:host([scale="l"][width="full"]) button:not(.content--slotted),
+:host([scale="l"][width="full"]) a:not(.content--slotted) {
+  @apply w-full;
+}
```
alisonailea, line 696: alternative approach

```css
:host(:is([scale="s"], [scale="m"], [scale="l"])[width="full"]) {
  a, button {
    &:not(.content--slotted) {
    }
  }
}
```

alisonailea (approved):

packages/calcite-components/src/components/button/button.scss
driskull: Another alternative approach would be to apply a class here using JS since the CSS logic is getting a little complex here.

driskull (approved):
Code looks good! @josercarcamo is this fix captured in a screenshot test? If not can we add one?

PR #9151, opened 2024-04-18
fix(date-picker): update names and E2E tests

alisonailea:
**Related Issue:** #7180 

## Summary

Update token names and add tests

### DIFF
--- a/packages/calcite-components/src/components/date-picker/date-picker.e2e.ts
+++ b/packages/calcite-components/src/components/date-picker/date-picker.e2e.ts
@@ -112,22 +112,7 @@ describe("calcite-date-picker", () => {
         shadowSelector: "calcite-date-picker-month",
       },
     } as const;
-    themed(
-      `<calcite-date-picker-month-header active-date="${new Date()}" locale-data"${{
-        "default-calendar": "gregorian",
-        separator: "/",
-        unitOrder: "DD/MM/YYYY",
-        weekStart: 7,
-        placeholder: "DD/MM/YYYY",
-        numerals: "0123456789",
-        months: {
-          abbreviated: ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"],
-          narrow: ["J", "F", "M", "A", "M", "J", "J", "A", "S", "O", "N", "D"],
-          wide: ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"],
-        },
-      }}">`,
-      tokens,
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/date-picker/date-picker.e2e.ts
```diff
@@ -28,6 +28,108 @@ describe("calcite-date-picker", () => {
...
+        targetProp: "--calcite-date-picker-month-header-text-color",
+        shadowSelector: "calcite-date-picker-month-header",
+      },
+      "--calcite-date-picker-week-header-text-color": {
+        targetProp: "--calcite-date-picker-month-week-header-text-color",
+        shadowSelector: "calcite-date-picker-month",
+      },
+    } as const;
+    themed(
+      `<calcite-date-picker-month-header active-date="${new Date()}" locale-data"${{
```
jcfranco, line 116: Can you use the [`html` tagged template function](https://github.com/Esri/calcite-components/blob/14036c1354419222bc5b839af8b68675b17ebf62/packages/calcite-components/support/formatting.ts#L35-L37) to help Prettier format embedded HTML ([example](https://github.com/Esri/calcite-design-system/blob/f1ff5e1acd0e802f6c5b635a77dedf8138d7e753/packages/calcite-components/src/components/card/card.e2e.ts#L27-L29))?

packages/calcite-components/src/components/date-picker/date-picker.e2e.ts
```diff
@@ -28,6 +28,108 @@ describe("calcite-date-picker", () => {
...
+      },
+      "--calcite-date-picker-corner-radius": [
+        {
+          targetProp: "borderRadius",
+        },
+        {
+          shadowSelector: "calcite-date-picker-month-header",
+          targetProp: "--calcite-date-picker-month-header-corner-radius",
+        },
+      ] as TestSelectToken[],
```
jcfranco, line 44: Do you still need to type cast after recent type updates to `themed`?

jcfranco (approved):

packages/calcite-components/src/components/date-picker/date-picker.e2e.ts
alisonailea: yes

jcfranco: Oh, I see why. Typecasting bypasses type checks at assignment time.

 If we export the type of `themed`'s `tokens` arg, we can type `tokens` here as 

```ts
const tokens: ThemedTokensArg /* example name 😅 */ = {
  // ...
};
```

Alternatively, we can pass the object directly to `themed` so that TS checks the type there.

Issue #7714, opened 2023-09-08
Select: update chevron colors and hover/pressed background color


ashetland:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Select's chevrons are `ui-text-2` in all states. The background also changes to `foreground-2` when hovered and pressed.

![image](https://github.com/Esri/calcite-design-system/assets/5023024/2873122e-1f11-4914-93db-8aa2faeeb8af)

### Expected Behavior

For system consistency, chevrons should be `ui-text-3` when idle, `ui-text-1` when hovered or pressed, and the background color should not change from `foreground-1`.

![image](https://github.com/Esri/calcite-design-system/assets/5023024/1834820f-c475-4736-ad0c-7e330b8a7ba8)

### Reproduction Sample

https://codepen.io/ashetland/pen/MWZpgRd

### Reproduction Steps

Interact with Select.

### Reproduction Version

v1.8.0

### Relevant Info

Figma designs for Select [here](https://www.figma.com/file/d7UBycO6x1Vz8zskFj5dTI/Epic---icon-consistency-%5B7181%5D?type=design&node-id=914-348742&mode=dev).

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

Calcite (design)

brittneytewks:
Removed Figma changes label as this will be tracked in monday.com

DitwanP:
🍠 Verified on `2.8.0-next.10`

PR #9160, opened 2024-04-19
fix(select): update the focus and hover chevron icon color and select focus and hover background color

aPreciado88:
**Related Issue:** [7714](https://github.com/Esri/calcite-design-system/issues/7714)

### Summary 
This updates `select` chevron colors to be `--calcite-color-text-3` when idle and `--calcite-color-text-1` when hovered or pressed.
This also updates `select` background color to `bg-foreground-1` when hovered or pressed.


aPreciado88:
@jcfranco Will do, thank you for taking a look!

ashetland:
Looks great!

Issue #9020, opened 2024-03-29
[input time zone] add a time zone picker null state


emmilaakso212:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

As a map author who wants to define feature template defaults, I want to be able to choose a default timezone for data entry, or leave it blank.

### Acceptance Criteria

- Create the option to include a time zone picker null state
- Use null value "Not set" 
- When picker is active, show current selection “Not set” at the top of the list
- Replace null value if a time zone is chosen
- User can scroll back to the top or search for “not set” to go back to the null state
[
See designs in Figma](https://www.figma.com/file/JXlrtCx8WmsXZRk8Lzd9Oq/Timezone%2Ftime-based-filters?type=design&node-id=244%3A170583&mode=dev&t=9zsUwmYpjRYD6307-1)

### Relevant Info

Requested by Taylor McNeil and Jasmine Reyes for feature editing enhancements related to default time zones for 24-R2

### Which Component

Input time zone

### Priority impact

p1 - need for current milestone

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Map Viewer

emmilaakso212:
cc: @jcfranco

tayMAPNeil:
Just giving this a ping. This enhancement will be really important for the add/delete field components for this release (for authoring TSO fields with no default value). 

@emmilaakso212 everything still on track?

emmilaakso212:
@jcfranco @geospatialem @tayMAPNeil echoing Taylor's question above. Let's make sure we prioritize the time zone picker tickets appropriately based on the remaining time, and our discussion around time zone work.

jcfranco:
@tayMAPNeil @emmilaakso212 Thanks for the ping. We have not triaged this yet on our end, but we are aware of the priority. I'll update this and related `input-time-zone` issues to reflect this and we'll take that into consideration when we do planning next week. cc @geospatialem

geospatialem:
@tayMAPNeil @emmilaakso212 Calcite is targeting this month's release (April) to introduce the null state to land as part of R2.

DitwanP:
[...]

PR #9168, opened 2024-04-22
feat(input-time-zone): allow clearing value

jcfranco:
**Related Issue:** #9020

## Summary

Adds `clearable` property to allow `input-time-zone` to be cleared via the UI or programmatically via `””` or `null`.

### DIFF
--- a/packages/calcite-components/src/components/input-time-zone/input-time-zone.tsx
+++ b/packages/calcite-components/src/components/input-time-zone/input-time-zone.tsx
@@ -78,6 +78,13 @@ export class InputTimeZone
   //--------------------------------------------------------------------------
 
   /**
+   * When `true`, an empty value (`null`) will be allowed as a `value`.
+   *
+   * When `false`, an offset or name value is enforced, and clearing the input or blurring will restore the last valid `value`.
+   */
+  @Prop({ reflect: true }) clearable = false;
+
+  /**
    * When `true`, interaction is prevented and the component is displayed with lower opacity.
    */
   @Prop({ reflect: true }) disabled = false;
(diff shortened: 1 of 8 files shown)


packages/calcite-components/src/components/input-time-zone/input-time-zone.tsx
```diff
@@ -77,6 +77,13 @@ export class InputTimeZone
   //
   //--------------------------------------------------------------------------
 
+  /**
```
jcfranco, line 80: @geospatialem @DitwanP Would you mind reviewing this doc? It's based on `color-picker`'s [`clearable` prop doc](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/color-picker/color-picker.tsx#L130-L134), but I think both need some clarity tweaks.

geospatialem: LGTM - think both use cases for `clearable` below and for `color-picker` although a bit wordy provide good context, but am flexible if @DitwanP has a recommended approach, too.

alisonailea:
What is the expected user experience to clear the value of the InputTimeZone? I took a look at this locally and I couldn't clear values as I expected. I.E. similar to clearing values in Combox.

driskull (approved):
LGTM!

jcfranco:
> What is the expected user experience to clear the value of the InputTimeZone? I took a look at this locally and I couldn't clear values as I expected. I.E. similar to clearing values in Combox.

@alisonailea [This story](https://6266d45509d7eb004aa254fb-savsemwumb.chromatic.com/?path=/story/components-controls-inputtimezone--clearable) should allow you to clear the value via the clear affordance, deselecting the active item.

Issue #1631, opened 2021-02-26
Enhancement: Configurable slider highlighting


bpatterson88:
### Description

Allow slider to highlight different sections of the rail, ex. left of the handle _or_ right of the handle. 

### Acceptance Criteria 

- Single value: rail can highlight right, left, or none

### Relevant Info 

Dashboards currently uses sliders to filter visualizations, and depending on the operator we highlight different parts of the rail. This request will give Calcite parity with the current Dashboards implementation of slider.

#### Which Component

Slider

#### Example Use Case

- Select all values greater than _N_ (ex. to the right of the handle)
- Select _N_ (ex. no highlighting)

paulcpederson:
@bpatterson88  would this need to apply to range as well or just single value?

bpatterson88:
@paulcpederson Nope, just the single value.

Also, updated the request to include a 'none' option for highlighting, where the slider is used to select a single value.

bpatterson88:
@benelan Dashboards uses sliders for less than, greater than, and equals queries. Different parts of the slider rail are highlighted depending on the operator. The screenshots below show our current implementation. 

![Screen Shot 2021-11-29 at 4 30 29 PM](https://user-images.githubusercontent.com/15875886/143963885-9c73b512-229b-4e6e-b47e-568d50357f1e.png)
![Screen Shot 2021-11-29 at 4 30 38 PM](https://user-images.githubusercontent.com/15875886/143963889-92113fd5-3197-46a3-919e-290f27d3d722.png)
![Screen Shot 2021-11-29 at 4 30 47 PM](https://user-images.githubusercontent.com/15875886/143963896-69dd8beb-b076-4acc-938b-dbb13169fe28.png)

ellenupp:
Hi @benelan, just wanted to put this issue on your radar again and see where it's at. We would love to start using calcite-slider in dashboards but this is a blocker for us.

benelan:
@bpatterson88 thanks for providing the use case, and thanks for bumping the issue @ellenupp. We are planning to get this in before v1 and will have a better timeline after our next triage meeting, to determine which sprint has cycles.

richiecarmichael:
This is needed for the new JSAPI **Range Slider** (see [#46032](https://devtopia.esri.com/WebGIS/arcgis-js-api/issues/46032)).

ubatsukh:
[...]

PR #9170, opened 2024-04-23
feat(slider): allow configuring fill behavior

jcfranco:
**Related Issue:** #1631

## Summary

Adds `fillPlacement` property to allow configuration of fill along the track.

### Notes

* Range mode will always display fill between thumbs
* This is purely a visual option, so no E2E test was added

### DIFF
--- a/packages/calcite-components/src/components/slider/slider.tsx
+++ b/packages/calcite-components/src/components/slider/slider.tsx
@@ -93,11 +93,11 @@ export class Slider
   @Prop({ reflect: true, mutable: true }) hasHistogram = false;
 
   /**
-   * Used to configure where the highlight is placed along the slider track.
+   * Used to configure where the fill is placed along the slider track in relation to the value handle.
    *
-   * **Note**: range mode will always display range between min and max handles.
+   * Range mode will always display the fill between the min and max handles.
    */
-  @Prop({ reflect: true }) highlightMode: "default" | "none" | "mirrored" = "default";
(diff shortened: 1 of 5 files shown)


packages/calcite-components/src/components/slider/slider.tsx
```diff
@@ -93,6 +92,13 @@ export class Slider
   /** When `true`, indicates a histogram is present. */
   @Prop({ reflect: true, mutable: true }) hasHistogram = false;
 
+  /**
+   * Used to configure where the highlight is placed along the slider track.
+   *
+   * **Note**: range mode will always display range between min and max handles.
+   */
+  @Prop({ reflect: true }) highlightMode: "default" | "none" | "mirrored" = "default";
```
driskull, line 100: comment: Is there a better name for "default"? Otherwise, default doesn't really mean anything other than its the default whereas the others have some meaning to them.

driskull (approved):
Code looks good, just had a comment on the default value. 👍

packages/calcite-components/src/components/slider/slider.tsx
jcfranco: > "start" (default) / "end" / "none" ? There is also the separate mirrored property on Slider which might be confusing.

jcfranco: Went with `fillPlacement: "start" | "end" | "none"` since [`meter`](https://developers.arcgis.com/calcite-design-system/components/meter/) introduces fill-related props, which would apply to slider too.

Issue #9116, opened 2024-04-11
Add more details about commit message format to the contributing doc


jcfranco:
### Description

We currently link to [Ionic's conventions](https://github.com/Esri/calcite-design-system/blob/main/CONTRIBUTING.md#commit-message-format), but we should explicitly document this for our purposes.

Additionally, this section should include examples.

### Which Component

N/A

### Resources

https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md#commit-message-format

geospatialem:
Verified on [contributing.md](https://github.com/Esri/calcite-design-system/blob/main/CONTRIBUTING.md#commit-message-format).

PR #9185, opened 2024-04-24
docs: add convention for property values

driskull:
**Related Issue:** N/A

## Summary

add convention for property values

### DIFF
--- a/packages/calcite-components/conventions/README.md
+++ b/packages/calcite-components/conventions/README.md
@@ -146,7 +146,7 @@ Private/internal properties should be annotated accordingly to avoid exposing th
 
 Property values should be meaningful. Avoid using values such as "default" for default values. Instead, use a more descriptive value. 
 
-For instance: `placement: "start" | "end" = "start"` is preferred to `placement: "default" | "end" = "default"`
+For instance: `placement: "start" | "end" = "start";` is preferred to `placement: "default" | "end" = "default";`
 
 ### Reflecting to attributes
 
--- a/packages/calcite-components/conventions/README.md
+++ b/packages/calcite-components/conventions/README.md
@@ -146,6 +146,8 @@ Private/internal properties should be annotated accordingly to avoid exposing th
 
 Property values should be meaningful. Avoid using values such as "default" for default values. Instead, use a more descriptive value. 
 
+For instance: `placement: "start" | "end" = "start"` is preferred to `placement: "default" | "end" = "default"`
+
 ### Reflecting to attributes
 
 It is recommended to reflect properties that fit the following criteria:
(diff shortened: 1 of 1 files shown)


packages/calcite-components/conventions/README.md
```diff
@@ -142,6 +142,10 @@ There are a few ways to add event listeners within our components:
 
 Private/internal properties should be annotated accordingly to avoid exposing them in the doc and/or API. You can do this by using the `@private`/`@internal` [JSDoc](https://jsdoc.app/) tags.
 
+### Property values
+
+Property values should be meaningful. Avoid using values such as "default" for default values. Instead, use a more descriptive value.
```
geospatialem, line 147: Great addition! 📝 

nitpick:
Is it possible to provide an example of what the default could look like vs. what could be implemented instead?
> For instance: `highlightPlacement: "start"` over `highlightMode: "default"`

geospatialem (approved):
Awesome! 💪🏻 

nitpick WRT conventional commit:
Would this be a `doc` rather than `chore`?

packages/calcite-components/conventions/README.md
driskull: Updated!

driskull:
> Would this be a doc rather than chore?

 It's not really documentation of any API and its not useful for an end user to see. 🤔 Not sure if there's a better conventional commit type.

driskull:
Looks like we don't show `docs` conventional commits in the changelog so I could use that commit type. Its documentation for us but not the end user so I'm not sure if `docs` or `chore` makes more sense. @jcfranco?

jcfranco:
Regarding the type, I could see us using either since we we don't include entries for `docs` in the changelog. We could

1. use `docs` to refer to both internal/external doc or 
2. keep `docs` for public and `chore(docs)` (or something similar) for internal doc

No strong preference, but 1 might make it simpler. Whatever we decide, we should include it in https://github.com/Esri/calcite-design-system/issues/9116.

jcfranco (approved):
✨🏆✨

geospatialem:
> Regarding the type, I could see us using either since we we don't include entries for `docs` in the changelog. We could
> 
> 1. use `docs` to refer to both internal/external doc or
> 2. keep `docs` for public and `chore(docs)` (or something similar) for internal doc
> 
> No strong preference, but 1 might make it simpler. Whatever we decide, we should include it in #9116.
> 
> [Here](https://github.com/Esri/calcite-design-system/commits/main/packages/calcite-components/conventions/README.md) are some examples of what we've used for these types of updates.

Thanks for laying out some options for us, and for referencing the related issue, @jcfranco! 🙌🏻 

Personally I lean towards `docs` since it seems more intuitive that we're updating docs, whether it be for internal or external uses. But am flexible if other folks think we should differentiate between the two and opt for `docs` for external, and `chore(docs)` for internal uses. What do others think?

driskull:
I'll change to docs.

Issue #9191, opened 2024-04-25
Tooling: Add a priority option to the test template


geospatialem:
### Summary

Update the test issue template to include a dropdown and automation that adds a priority label, based on the dropdown content.

### Desired Outcome

Add in a priority label of `p - high`, `p - medium`, or `p - low` depending on the dropdown selection.

### Resources

- test.yml
- addPriorityLabel.js

DitwanP:
🍬 Verified in test issue template.

PR #9192, opened 2024-04-25
ci: adds support to test issue template priority status

geospatialem:
**Related Issue:** #9191 

## Summary

Enables a priority option in the test issue template, which will add a priority label upon creation.


geospatialem:
WRT this comment:
> Also updates the addPriorityLabel.js to exclude a new label that includes a priority.

Not tied to the above, it is a bit extraneous with regards to the change, so can omit it from the PR - if it is better without the change, both from a code readability standpoint and outcome.

benelan (approved):
:+1:

Issue #7082, opened 2023-06-01
Epic: Audit scale across components


brittneytewks:
### Description

Audit scale across components.

Time estimate
- Design: `estimate-8`
- Development: `estimate-8`

**monday.com sync:** #8745622293

isaacbraun:
Verified that the sub-issue components all have functional `scale` properties added via individual verifications and a spot-checking CodePen: https://codepen.io/isaacbraun/pen/ogZgOae.

Additionally, confirmed with the design team that this epic is good to close, as additional changes and components have been added throughout its life. 🎉

PR #9194, opened 2024-04-25
feat(block-section): add `icon` property (deprecates `status`)

Elijbet:
**Related Issue:** #8110

## Summary
Remove the `status` property and instead add an `icon` property.

Provide a demo and screenshot coverage.


packages/calcite-components/src/components/block-section/block-section.scss
```diff
@@ -1,3 +1,14 @@
+/**
```
jcfranco, line 1: We might want to hold off on documenting these. The review from https://github.com/Esri/calcite-design-system/pull/8594 might affect some props. If we want to be extra careful, we might need to rename these as internal.

Elijbet: Sure, that makes sense.

packages/calcite-components/src/components/block-section/block-section.tsx
```diff
@@ -199,21 +211,35 @@ export class BlockSection implements LocalizedComponent, T9nComponent, LoadableC
...
-    const arrowIcon = open
-      ? ICONS.menuOpen
-      : dir === "rtl"
-        ? ICONS.menuClosedLeft
-        : ICONS.menuClosedRight;
+    const { messages, open, text, toggleDisplay, iconFlipRtl } = this;
+    const arrowIcon = open ? ICONS.menuOpen : ICONS.menuClosed;
+
+    const iconStartEl = this.iconStart ? (
+      <calcite-icon
```
jcfranco, line 218: To simplify a bit, can you look at adding a `renderIcon` method that returns an icon with the corresponding start/end props?

packages/calcite-components/src/components/block-section/block-section.tsx
```diff
@@ -199,21 +211,35 @@ export class BlockSection implements LocalizedComponent, T9nComponent, LoadableC
...
+        class={CSS.iconStart}
+        flipRtl={iconFlipRtl === "both" || iconFlipRtl === "start"}
+        icon={this.iconStart}
+        key="icon-start"
+        scale="s"
```
jcfranco, line 223: Does this need to use the [`getIconScale`](https://github.com/Esri/calcite-design-system/blob/f1ff5e1acd0e802f6c5b635a77dedf8138d7e753/packages/calcite-components/src/utils/component.ts#L4-L6) util?

packages/calcite-components/src/components/block-section/block-section.tsx
```diff
@@ -55,9 +56,20 @@ export class BlockSection implements LocalizedComponent, T9nComponent, LoadableC
 
   /**
    * Displays a status-related indicator icon.
+   *
+   * @deprecated Use `icon-start` instead.
    */
   @Prop({ reflect: true }) status: Status;
 
+  /** Specifies an icon to display at the start of the component. */
```
jcfranco, line 64: Can you make sure props are in alphabetical order?

jcfranco (approved):
Looks great! A few things:

* Let's target this for the May release. The [token work](#8594) might affect it and it'll also give more time for users to prep for the visual changes this will introduce. 
* Can you apply the `visual changes` to the issue?
* Can you rephrase the PR title to be more along the lines of: `feat(component): add X prop (deprecates Y)`? We want to highlight the added prop first to align with the commit type.

jcfranco (approved):
Looks great! A few things:

* Let's target this for the May release. The [token work](#8594) might affect it and it'll also give more time for users to prep for the visual changes this will introduce. 
* Can you apply the `visual changes` to the issue?
* Can you rephrase the PR title to be more along the lines of: `feat(component): add X prop (deprecates Y)`? We want to highlight the added prop first to align with the commit type.

packages/calcite-components/src/components/block-section/block-section.tsx
Elijbet: `block/block-section` don't have `scale` currently. It'll be reviewed in https://github.com/Esri/calcite-design-system/issues/7082. As is, we think of the components that don't have scale as `medium` and use `small` icons for them.

packages/calcite-components/src/components/block-section/block-section.stories.ts
```diff
@@ -0,0 +1,31 @@
+import { html } from "../../../support/formatting";
+
+export default {
+  title: "Components/Block Section",
+};
+
+export const icons_TestOnly = (): string => html`
+  <calcite-block heading="Heading" description="summary" collapsible open>
```
jcfranco, line 8: Can you move this into the block stories? We usually include child stories to the parent one.

packages/calcite-components/src/components/block-section/block-section.tsx
```diff
@@ -230,31 +257,42 @@ export class BlockSection implements LocalizedComponent, T9nComponent, LoadableC
...
           >
+            {this.renderIcon(this.iconStart)}
             <div class={CSS.toggleSwitchContent}>
               <span class={CSS.toggleSwitchText}>{text}</span>
             </div>
+
+            {this.renderIcon(this.iconEnd)}
             {this.renderStatusIcon()}
+            {/* we use calcite-label to use a simple component that will allow us to prevent keyboard focus by setting tabindex="-1" on the host */}
+            <calcite-label class="label" layout="inline" tabIndex={-1}>
```
jcfranco, line 268: Let's add a `label` prop to the CSS lookup object.

packages/calcite-components/src/components/block-section/block-section.scss
```diff
@@ -51,35 +47,58 @@
...
-    inset-inline-end: 0;
-    margin-inline-start: theme("margin.1");
+  .toggle--switch__content {
+    @apply flex flex-auto items-center;
+  }
+
+  .icon--end {
+    @apply flex items-center;
+
+    color: var(--calcite-internal-block-section-icon-end-color, var(--calcite-color-text-3));
```
jcfranco, line 75: Let's keep this if the internal CSS props help with #8594 integration. Otherwise, it might be cleaner to remove these here and add when the PR is synced w/ `main`.

jcfranco (approved):
Good to merge once last round of changes are addressed. 🎉

PR #9195, opened 2024-04-25
fix(slider): rename `highlightPlacement` to `fillPlacement`

jcfranco:
**Related Issue:** #1631

## Summary

Renames the property according to thread in #9170.

### DIFF
--- a/packages/calcite-components/src/components/slider/slider.tsx
+++ b/packages/calcite-components/src/components/slider/slider.tsx
@@ -97,7 +97,7 @@ export class Slider
    *
    * Range mode will always display the fill between the min and max handles.
    */
-  @Prop({ reflect: true }) highlightPlacement: "start" | "none" | "end" = "start";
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/slider/slider.tsx
```diff
@@ -97,7 +97,7 @@ export class Slider
    *
    * Range mode will always display the fill between the min and max handles.
    */
-  @Prop({ reflect: true }) highlightPlacement: "start" | "none" | "end" = "start";
+  @Prop({ reflect: true }) fillPlacement: "start" | "none" | "end" = "start";
```
driskull, line 100: 👍

driskull (approved):
Looks good. Should there be a defaults test that looks for fillPlacement and makes sure its set to "start" by default?

jcfranco:
Good catch. Test added!

PR #9204, opened 2024-04-26
fix(slider): fix fill placement and tick styling when ranged

jcfranco:
**Related Issue:** #1631

## Summary

This ensures ranged sliders place the fill and style ticks within the range.


driskull (approved):
👍

Issue #9177, opened 2024-04-23
Memory leak with adding/removing block components


nwhittaker:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Detached `<calcite-block>` elements are unexpectedly retained in memory.

### Expected Behavior

They are not unexpectedly retained in memory.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/YzMBXVY

### Reproduction Steps

1. Visit the repro and open dev tools to the **Memory** tab.
2. Click the _Replace button_ button several times to remove the current button element and replace it with a new one each time.
3. If no messages appear in the console, click the _Collect garbage_ button in the dev tool's **Memory** tab.
4. At this point, several messages should appear showing the removed button elements have been garbage collected.
5. Click the _Replace block_ button several times to remove the current block element and replace it with a new one each time.
6. Click the _Collect garbage_ button and see no messages are printed to the console. If the blocks were garbage collected, their messages should appear.

### Reproduction Version

2.7.1

### Relevant Info

Looking at the memory dump, it looks like these blocks might be being over-retained by a mutation observer. Since the observer is shared code, other components may be exhibiting a similar memory leak.

### Priority impact

p2 - want for current milestone

### Impact

The memory leak is an issue that may become more noticeable as the team writes larger and more complex custom components. Performance issues may become noticeable to users on slower hardware.

To provide a granular editing experience, we stack panels on top of each other. The top panel organizes different custom components into calcite-blocks. We're seeing each new instance of our custom components being retained. So, if the user opens/closes the panel 6 times, 6 instances of each of our custom components are retained and never garbage collected.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Field Apps

jcfranco:
Thanks for reporting this. Taking a look. 🔍👁️👄👁️

DitwanP:
🍰 Verified in `2.8.0-next.16` 🗑️🧹

PR #9208, opened 2024-04-26
fix: fix memory leak affecting components using conditionally-rendered slots

jcfranco:
**Related Issue:** #9177

## Summary

Fixes an issue caused by `conditionalSlot` implementing `unobserve` on top of `ExtendedMutationObserver`, which already implements it.


driskull (approved):
👍

benelan (approved):
Nice!

Issue #8666, opened 2024-01-29
Add "readonly" prop/style to Combobox (match style of other "readonly" components).


JonUihlein:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Requesting a `readOnly` equivalent for Combobox, to match the styles of other `readOnly` inputs. We avoid using `disabled` to allow selection of text within the input. The styling of `readOnly` is different compared to `disabled` and makes the form look strange.

### Acceptance Criteria

A property to allow matching the styling of other `readOnly` inputs. 

### Which Component

Combobox

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

DitwanP:
🍯 Verified on `main`

Issue #7853, opened 2023-09-20
Add readonly prop to input-time-zone (like input-date-picker, input-time-picker)


JonUihlein:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Both `<calcite-input-date-picker>` and  `<calcite-input-time-picker>` support the `readOnly` property, which has different styling than the `disabled` property. This is a request for adding `readOnly` property to `<calcite-input-time-zone>` so we can keep styling consistent. We prefer `readOnly` for accessibility reasons since `disabled` prevents interaction with the component (and prevents things like copying the value by hand/keyboard). 

Mixed "readOnly" and "disabled":
![editor-timezone-disabled](https://github.com/Esri/calcite-design-system/assets/2378615/6bec2f5c-8c74-470b-ab49-6329a7e89733)

### Acceptance Criteria

Support for `readOnly` on `<calcite-input-time-zone>`

### Which Component

`<calcite-input-time-zone>`

### Example Use Case

Styles used by `<calcite-input-date-picker>` and `<calcite-input-time-picker>`

### Priority impact

p1 - need for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

ashetland:
@josercarcamo Just saw this had a design label on it. Do you need anything from us on this one or just a quick review of the visuals?

josercarcamo:
Hi @ashetland. I'm not sure. I didn't notice the "design" tag. I was just told by @JonUihlein to make it like the other readonly input elements, so I tried to match the UI/UX of the input date and time pickers.

DitwanP:
@josercarcamo I'm having trouble verifying this issue. on this [codepen](https://codepen.io/Ditwan-Price/pen/YzMJVNB?editors=100) you can see that `read-only` is working as expected on the `date-picker` but not on the `input-time-zone`. 

I also ran the demos locally and it seems like it's working in regards to not allowing me to actually open the dropdown, but the styling there doesn't match. It looks like the components that get `read-only` don't show the dropdown arrow on the right side of the input but the `input-time-zone` is still showing it in the demos.

[...]

PR #9222, opened 2024-04-30
fix(combobox, input-time-zone): improve readOnly behavior and styling

jcfranco:
**Related Issue:** #7853 #8666

## Summary

Improves read-only behavior in combobox and applies proper styling.

### DIFF
--- a/packages/calcite-components/src/components/combobox/combobox.scss
+++ b/packages/calcite-components/src/components/combobox/combobox.scss
@@ -74,6 +74,12 @@
   @apply focus-inset;
 }
 
+:host([read-only]) {
+  .wrapper {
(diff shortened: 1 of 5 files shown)


packages/calcite-components/src/components/combobox/combobox.scss
```diff
@@ -74,6 +74,12 @@
   @apply focus-inset;
 }
 
+:host([read-only]) {
+  .wrapper {
+    @apply bg-background font-medium;
```
anveshmekala, line 79: Q: can we use tokens here instead?

driskull (approved):
Looks good! 🥂

Issue #6456, opened 2023-02-10
[Dialog] Close contextual elements with the `Esc` key in visual order


geospatialem:
### Description
Provide similar behavior to keyboard functionality to mouse users where if multiple modal dialogs are opened, the Esc key provides similar functionality to when selecting the components close button.

cc @Anastasiia-Boleiko

### Acceptance Criteria
If multiple modal dialog windows are open and the Esc key is pressed, the most recently opened (highest order, or visually prominent) should be closed first. The order should follow until the last dialog is closed.

### Relevant Info
Came up in #6292 as a nice to have, but was out of scope for the bug fix.

### Which Component
modal

### Example Use Case
https://codepen.io/anastasiia-zrh/pen/MWBvZdB

### Esri team
ArcGIS Urban

cont-kolomeets:
Another case related to this one is to close an open popover first on the 1st Esc press, then close the modal on the 2nd Esc press.
Right now hitting Esc closes both at once.

### Example Use Case

https://codepen.io/cont-kolomeets/pen/RwqzgqP?editors=100

![image](https://github.com/Esri/calcite-design-system/assets/5318527/58347e06-8224-4080-9456-4281239b68c0)

geospatialem:
Spike to determine if the new `dialog` needs to be updated to handle `modal` with the visual order specified above has the current functionality, or if needed, should be supported with the deprecation of `modal` in the upcoming `2.11.0` release (7/30).

geospatialem:
Adding some notes @Elijbet compiled for testing out the functionality in the [attached PR](https://github.com/Esri/calcite-design-system/pull/9231#issuecomment-2397808476) ✨💪🏻✨

DitwanP:
🍡 Verified on `2.14.0-next.18` 

https://codepen.io/Ditwan-Price/pen/MWNQdvw?editors=1011

PR #9231, opened 2024-04-30
feat(dialog, modal, popover, input-date-picker,  input-time-picker, sheet): support stacked component sequential closing with escape

Elijbet:
**Related Issue:** #6456

## Summary
Support component sequential closing with escape by having focus-trap handle the stacking order.


packages/calcite-components/src/components/sheet/sheet.e2e.ts
```diff
@@ -541,4 +541,144 @@ describe("calcite-sheet properties", () => {
...
+    const secondModal = await page.find("#another-modal");
+    await openAndCheckVisibility(page, secondModal);
+
+    async function testInputPicker(page: E2EPage, pickerSelector: string, modal: E2EElement) {
+      const inputPicker = await page.find(pickerSelector);
+      inputPicker.click();
+      await page.waitForChanges();
+      expect(await inputPicker.getProperty("open")).toBe(true);
+
+      await page.keyboard.press("Escape");
```
driskull, line 646: I think each escape should check to make sure only the last opened component is closed. Basically have a stack of all the components that are opened and verify only the expected one is closed at a time. Each call should close the next one. Currently, its only checking one component at a time.

packages/calcite-components/src/demos/sheet.html
```diff
@@ -37,7 +37,7 @@
 
   <body>
     <demo-dom-swapper>
-      <h1 style="margin: 0 auto; text-align: center">Sheet</h1>
+      <!-- <h1 style="margin: 0 auto; text-align: center">Sheet</h1>
```
driskull, line 40: uncomment this.

packages/calcite-components/src/utils/focusTrapComponent.ts
```diff
@@ -11,6 +11,12 @@ export interface FocusTrapComponent {
    */
   el: HTMLElement;
 
+  /** When `true`, disables the default close on escape behavior. */
```
driskull, line 14: nice!

packages/calcite-components/src/utils/focusTrapComponent.ts
```diff
@@ -27,6 +33,8 @@ export interface FocusTrapComponent {
    * This should be implemented for components that allow user content and/or have conditionally-rendered focusable elements within the trap.
    */
   updateFocusTrapElements?: () => Promise<void>;
+
+  onFocusTrapDeactivate?(): void;
```
driskull, line 37: this one needs doc.

packages/calcite-components/src/utils/focusTrapComponent.ts
```diff
@@ -58,9 +66,13 @@ export function connectFocusTrap(component: FocusTrapComponent, options?: Connec
...
 
   const focusTrapOptions: FocusTrapOptions = {
-    clickOutsideDeactivates: true,
-    escapeDeactivates: false,
+    clickOutsideDeactivates: !component.outsideCloseDisabled ?? true,
+    escapeDeactivates: !component.escapeDisabled ?? true,
     fallbackFocus: focusTrapNode,
+    onDeactivate: () => {
+      component.onFocusTrapDeactivate();
+      console.log("focus trap deactivated");
```
driskull, line 74: remove this comment.

Can be simplified to be:

```ts
onDeactivate: () => component.onFocusTrapDeactivate(),
```

driskull (commented):
Initial feedbackz

packages/calcite-components/src/utils/focusTrapComponent.ts
driskull: Here's the doc from focus-trap:

> onDeactivate {() => void}: A function that will be called before returning focus to the node that had focus prior to activation (or configured with the setReturnFocus option) upon deactivation.

packages/calcite-components/src/components/sheet/sheet.e2e.ts
Elijbet: `calcite-input-date-picker` and `time-picker` don't get stacked, so these 2 will have their own stacks to be checked.

driskull: We can probably remove input-date-picker and time-picker from the tests. We don't need to test components that don't focus trap.

packages/calcite-components/src/components/sheet/sheet.e2e.ts
```diff
@@ -541,4 +541,144 @@ describe("calcite-sheet properties", () => {
...
+    expect(await popover.getProperty("open")).toBe(true);
+
+    await page.keyboard.press("Escape");
+    await page.waitForChanges();
+    expect(await popover.getProperty("open")).toBe(false);
+
+    async function pressEscapeAndCheckVisibility(page: E2EPage, element: E2EElement, expectedVisibility: boolean) {
+      page.keyboard.press("Escape");
+      await page.waitForChanges();
+      expect(await element.isVisible()).toBe(expectedVisibility);
```
driskull, line 677: This should check all the other elements each time esc is pressed to make sure only the expected element is closed.

packages/calcite-components/src/components/sheet/sheet.e2e.ts
Elijbet: Maybe there is some benefit in making sure the top component doesn't close focus-trap stack?

packages/calcite-components/src/demos/sheet.html
jcfranco: This is still commented out.

jcfranco: This is still commented out.

packages/calcite-components/src/utils/focusTrapComponent.ts
```diff
@@ -119,3 +133,17 @@ export function deactivateFocusTrap(
...
+ *
+ * @param {FocusTrapComponent} component The FocusTrap component.
+ * @param [options] The FocusTrap activate options.
+ * @param event
+ */
+export function clickOutsideDeactivates(event: MouseEvent): boolean {
+  const path = event.composedPath();
+  const isClickInside = path.some((el: EventTarget) => this.contains(el as Node));
+  // If the click is inside the component, do not deactivate the focus trap
+  return !isClickInside;
```
driskull, line 148:
```suggestion
  return !event.composedPath().includes(component.el);
```

packages/calcite-components/src/utils/focusTrapComponent.ts
```diff
@@ -119,3 +133,17 @@ export function deactivateFocusTrap(
...
 }
+
+/**
+ * Helper to manage focusTrap by determining if a click event occurred outside of the component. When `true`, the focus trap will deactivate.
+ *
+ * @param {FocusTrapComponent} component The FocusTrap component.
+ * @param [options] The FocusTrap activate options.
+ * @param event
+ */
+export function clickOutsideDeactivates(event: MouseEvent): boolean {
```
driskull, line 144:
```suggestion
export function clickOutsideDeactivates(event: MouseEvent, component: FocusTrapComponent): boolean {
```

packages/calcite-components/src/utils/focusTrapComponent.ts
```diff
@@ -119,3 +133,17 @@ export function deactivateFocusTrap(
 export function updateFocusTrapElements(component: FocusTrapComponent): void {
   component.focusTrap?.updateContainerElements(component.el);
 }
+
+/**
+ * Helper to manage focusTrap by determining if a click event occurred outside of the component. When `true`, the focus trap will deactivate.
```
driskull, line 138: Cool. This isn't being used anywhere is it? I'm not seeing anything.

packages/calcite-components/src/utils/focusTrapComponent.ts
```diff
@@ -58,9 +71,10 @@ export function connectFocusTrap(component: FocusTrapComponent, options?: Connec
   }
 
   const focusTrapOptions: FocusTrapOptions = {
-    clickOutsideDeactivates: true,
-    escapeDeactivates: false,
+    clickOutsideDeactivates: component.clickOutsideDeactivates ?? true,
```
driskull, line 74: We could maybe possibly make this something like: 

```suggestion
    clickOutsideDeactivates: component.clickOutsideDeactivates ?? clickOutsideDeactivates.call(null, this),
```

packages/calcite-components/src/utils/focusTrapComponent.ts
```diff
@@ -58,9 +71,10 @@ export function connectFocusTrap(component: FocusTrapComponent, options?: Connec
...
+    escapeDeactivates: !component.escapeDisabled ?? true,
```
jcfranco, line 75: This can be simplified to `!component.escapeDisabled`.

The `!` operator will cast `component.escapeDisabled` to a boolean, so the nullish coalescing operator won't ever get triggered because it only falls back to the right-hand side value only if the left-hand side is `null` or `undefined`.

jcfranco, line 75: FIXME: I noticed one issue with us relying on `focus-trap`'s `escapeDeactivates` feature. Since it uses the [`capture` phase for `keydown` events](https://github.com/focus-trap/focus-trap/blob/master/index.js#L837), it means our components can no longer ignore `Escape` presses that were already handled by a child component (e.g., user presses `Escape` to close open combobox inside a modal).

packages/calcite-components/src/demos/input-date-picker.html
```diff
@@ -53,10 +53,8 @@
       <div class="parent-flex">
         <div class="child-flex font right-aligned-text">Standard</div>
         <div class="child-flex">
-          <calcite-label>
-            Input Date Picker Label
-            <calcite-input-date-picker scale="s" value="2023-03-07"></calcite-input-date-picker>
-          </calcite-label>
+          <calcite-input-date-picker></calcite-input-date-picker>
```
jcfranco, line 56: This change seems out of place. Was it intentional?

packages/calcite-components/src/components/modal/modal.tsx
```diff
@@ -398,20 +397,6 @@ export class Modal
...
   @State() defaultMessages: ModalMessages;
 
-  //--------------------------------------------------------------------------
-  //
-  //  Event Listeners
-  //
-  //--------------------------------------------------------------------------
-
-  @Listen("keydown", { target: "window" })
-  handleEscape(event: KeyboardEvent): void {
```
jcfranco, line 408: FIXME: We need to keep this, or something similar, to preserve closing with `Escape` when developers set `focusTrapDisabled = true`. When `true`, focus traps are no longer active and therefore won't handle `Escape` presses. This would also apply to anything using `clickOutsideDeactivates`.

packages/calcite-components/src/components/sheet/sheet.e2e.ts
```diff
@@ -191,11 +191,16 @@ describe("calcite-sheet properties", () => {
...
     );
+    await skipAnimations(page);
     await page.waitForChanges();
+    await page.waitForEvent("calciteSheetOpen");
     expect(await sheet.getProperty("opened")).toBe(true);
+
     await page.keyboard.press("Escape");
     await page.waitForChanges();
     await page.waitForChanges();
+    await page.waitForTimeout(500);
```
jcfranco, line 202: Can we await on an open/close event instead?

packages/calcite-components/src/components/popover/popover.tsx
```diff
@@ -292,7 +292,23 @@ export class Popover
...
+    connectFocusTrap(this, {
+      focusTrapEl: this.el,
+      focusTrapOptions: {
+        allowOutsideClick: true,
+        clickOutsideDeactivates: (event: MouseEvent) => {
+          const isReferenceElementInPath =
+            this.referenceElement instanceof EventTarget &&
+            event.composedPath().includes(this.referenceElement);
+
+          const outsideClick = !event.composedPath().includes(this.el);
```
jcfranco, line 304: Can you DRY this up by storing `event.composedPath`, and `autoClose && outsideClick` in a local variable?

packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -427,6 +427,10 @@ export class Dialog
     deactivateFocusTrap(this);
   }
 
+  onFocusTrapDeactivate(): void {
```
jcfranco, line 430: I noticed components close on blur or when clicking outside or on another element (e.g., `scrim`). Can you confirm all component-closing `onFocusTrapDeactivate` calls are needed?

jcfranco (changes requested):

jcfranco (changes requested):
Solid progress. Just a few items we need to tackle first.

packages/calcite-components/src/components/modal/modal.e2e.ts
```diff
@@ -154,7 +154,7 @@ describe("calcite-modal", () => {
     const mockCallBack = jest.fn();
     await page.exposeFunction("beforeClose", mockCallBack);
     await page.setContent(`
-      <calcite-modal open></calcite-modal>
```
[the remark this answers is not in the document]
driskull: It shouldn't be necessary.

packages/calcite-components/src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -1034,7 +1031,7 @@ export class InputTimePicker
             {!this.readOnly && this.renderToggleIcon(this.open)}
           </div>
           <calcite-popover
-            focusTrapDisabled={true}
+            focusTrapDisabled={false}
```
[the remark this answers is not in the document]
driskull: Are you sure? If input-time-picker is handling the focus trapping wouldn't the popover always not have a focus trap?

packages/calcite-components/src/components/input-date-picker/input-date-picker.e2e.ts
```diff
@@ -353,23 +353,40 @@ describe("calcite-input-date-picker", () => {
...
+          (window as InputDatePickerEventOrderWindow).events = receivedEvents;
+
+          ["calciteInputDatePickerOpen", "calciteInputDatePickerClose"].forEach((eventType) => {
+            sheet.addEventListener(eventType, (event) => receivedEvents.push(event.type));
+          });
+        });
       });
 
       it("toggles the date picker when clicked", async () => {
+        const openSpy = await inputDatePicker.spyOnEvent("calciteInputDatePickerOpen");
```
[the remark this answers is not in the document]
Elijbet: From what I remember we introduce those in some places to avoid multiple `await page.waitForChanges();`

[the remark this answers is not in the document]
Elijbet: I'll open an issue to add user interaction test to the helper.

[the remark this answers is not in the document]
jcfranco: > From what I remember we introduced those in some places to avoid multiple await page.waitForChanges();

I believe we used `waitForEvent` for that. This is using event spies, which is different.

packages/calcite-components/src/components/input-time-picker/input-time-picker.tsx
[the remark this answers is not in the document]
jcfranco: `input-time-picker` is now delegating to `popover` for focus-trapping, but you do bring up a good point. @Elijbet Utilities for implementing `FocusTrapComponent` are no longer necessary and should be removed.

packages/calcite-components/src/components/dialog/dialog.e2e.ts
```diff
@@ -391,7 +391,9 @@ describe("calcite-dialog", () => {
 
       dialog.setProperty("open", true);
       await page.waitForChanges();
+      await page.waitForChanges();
       expect(await page.find(`calcite-dialog >>> .${CSS.containerOpen}`)).toBeDefined();
+      await page.waitForChanges();
```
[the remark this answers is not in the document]
Elijbet: I noticed that multiple `page.waitForChanges` are no longer needed. Maybe some of your other work on openClose components helped? Not sure.

packages/calcite-components/src/components/input-time-picker/input-time-picker.tsx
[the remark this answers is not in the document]
driskull: Yeah, it shouldn't be implementing FocusTrapComponent any longer then.

packages/calcite-components/src/utils/focusTrapComponent.ts
driskull: hmm that's not good. 

It seems to have been that way from the beginning (first release)

We could ask and see why its that way and if there is any chance of having a focusTrap option to not use capture events.

driskull: related issue: https://github.com/focus-trap/focus-trap/issues/1165

driskull: I guess instead of using escapeDeactivates, we would have to handle closing the appropriate focus trap element in the trapStack

Would have to look at our trapStack and call trap.deactivate()
 

https://github.com/focus-trap/focus-trap?tab=readme-ov-file#trapdeactivate

This means we would set `escapeDeactivates` to false and each component would handle calling some global method for closing the appropriate focusTrap on the trapStack.

packages/calcite-components/src/components/dialog/dialog.tsx
Elijbet: I think in dialog/modal we also have to consider the value of `outsideCloseDisabled`.

packages/calcite-components/src/utils/focusTrapComponent.ts
driskull: We would need to setup an event listener on the document to listen for the escape key and close the top most focusTrap in the focusTrapStack. We could probably have a method in the `focusTrapComponent` to do this.

To setup the event listener, we'll need to track how many active focus traps there are by having some trap count in `connectFocusTrap`. We'll need to add a method `disconnectFocusTrap` to be able to remove the event listener if there are no more focus traps in the stack.

driskull: Here's an example of what I'm thinking:

```patch
diff --git forkSrcPrefix/packages/calcite-components/src/utils/focusTrapComponent.ts forkDstPrefix/packages/calcite-components/src/utils/focusTrapComponent.ts
index 877c3dc04f3329dd1355a30e8af3d888b3e96cf4..e88af0d7cdf6c530382888b3f8a02c5537bc6dd7 100644
--- forkSrcPrefix/packages/calcite-components/src/utils/focusTrapComponent.ts
+++ forkDstPrefix/packages/calcite-components/src/utils/focusTrapComponent.ts
@@ -2,6 +2,9 @@ import { createFocusTrap, FocusTrap as _FocusTrap, Options as FocusTrapOptions }
 import { FocusableElement, focusElement, tabbableOptions } from "./dom";
 import { focusTrapStack } from "./config";
 
+// i'm not sure if this needs to be a map of all documents. Would the keydown event bubble up out of shadow dom elements? If so, then the map isn't necessary.
+const focusTrapDocumentMap = new Map<Document, number>();
+
 /**
  * Defines interface for components with a focus trap. Focusable content is required for components implementing focus trapping with this interface.
  */
@@ -51,6 +54,7 @@ interface ConnectFocusTrapOptions {
  */
 export function connectFocusTrap(component: FocusTrapComponent, options?: ConnectFocusTrapOptions): void {
   const { el } = component;
+  const { ownerDocument } = el;
   const focusTrapNode = options?.focusTrapEl || el;
 
   if (!focusTrapNode) {
@@ -68,12 +72,28 @@ export function connectFocusTrap(component: FocusTrapComponent, options?: Connec
     ...options?.focusTrapOptions,
 
     // the following options are not overridable
-    document: el.ownerDocument,
+    document: ownerDocument,
     tabbableOptions,
     trapStack: focusTrapStack,
   };
 
   component.focusTrap = createFocusTrap(focusTrapNode, focusTrapOptions);
+  const newCount = (focusTrapDocumentMap.get(ownerDocument) || 0) + 1;
+  if (newCount === 1) {
+    addEscapeListener(ownerDocument);
+  }
+  focusTrapDocumentMap.set(ownerDocument, newCount);
+}
+
+// would need to call this on disconnect of all focus trap components.
+export function disconnectFocusTrap(component: FocusTrapComponent): void {
+  const { el } = component;
+  const { ownerDocument } = el;
+  const newCount = (focusTrapDocumentMap.get(ownerDocument) || 1) - 1;
+  if (newCount === 0) {
+    removeEscapeListener(ownerDocument);
+  }
+  focusTrapDocumentMap.set(ownerDocument, newCount);
 }
 
 /**
@@ -119,3 +139,17 @@ export function deactivateFocusTrap(
 export function updateFocusTrapElements(component: FocusTrapComponent): void {
   component.focusTrap?.updateContainerElements(component.el);
 }
+
+function escapeHandler(event: KeyboardEvent): void {
+  if (event.key === "Escape") {
+    focusTrapStack[focusTrapStack.length - 1]?.deactivate(); // deactivate the last active focus trap. This might be the 0 index. I'm not sure.
+  }
+}
+
+function addEscapeListener(document: Document): void {
+  document.addEventListener("keydown", escapeHandler);
+}
+
+function removeEscapeListener(document: Document): void {
+  document.removeEventListener("keydown", escapeHandler);
+}
```

This would let us handle the escape key and close the last active trap.

driskull: Example: https://github.com/Esri/calcite-design-system/pull/10197

Elijbet:
To recap, this is where we are:

- Cleanup modal and dialog by double checking for a single way to close both focus-trap enabled and disabled cases.
That is, scrim has a click handler that does the closing, we want to make sure we don’t handle closing also with focus trap. If scim is meant to close everything maybe we also use it to close the focus trap. If not, lets keep it as is.

- We confirmed that input-date-picker still closes on outside click, so comment regarding that is no longer relevant.

- There is an outstanding PR with focu-strap repo to enhance so it will allow child elements to handle Escape key presses without causing the trap to close. We might have to wait for this to land to pick up the remaining comments.

packages/calcite-components/src/components/popover/popover.tsx
```diff
@@ -293,7 +293,25 @@ export class Popover
...
     connectMessages(this);
-    connectFocusTrap(this);
+    connectFocusTrap(this, {
+      focusTrapEl: this.el,
+      focusTrapOptions: {
+        allowOutsideClick: true,
+        clickOutsideDeactivates: (event: MouseEvent) => {
+          const path = event.composedPath();
+          const isReferenceElementInPath =
+            this.referenceElement instanceof EventTarget && path.includes(this.referenceElement);
```
driskull, line 303: I think this should use `this.effectiveReferenceElement` instead of `this.referenceElement`

jcfranco:
> There is an https://github.com/focus-trap/focus-trap/pull/1253 to enhance so it will allow child elements to handle Escape key presses without causing the trap to close. We might have to wait for this to land to pick up the remaining comments.

We [recently updated `focus-trap`](https://github.com/Esri/calcite-design-system/pull/10281), which now allows child elements to handle Escape key presses without closing the trap, but I noticed that `dialog` doesn't have logic to ignore canceled Escape key presses within `focus-trap`. Can you check that this is handled? Any code handling Escape like [this](https://github.com/Esri/calcite-design-system/blob/elijbet/6456-component-closing-sequence/packages/calcite-components/src/components/dialog/dialog.tsx#L396-L402) should be changed to go through `focus-trap`, similar to this [example](https://github.com/focus-trap/focus-trap/blob/master/docs/js/escape-key-cancelation.js).

jcfranco:
Could you also coordinate `input-date-picker` changes w/ @anveshmekala? Additional coverage was added all date-picker-related test suites (see https://github.com/Esri/calcite-design-system/pull/8402/files).

packages/calcite-components/src/components/dialog/dialog.e2e.ts
```diff
@@ -716,17 +716,15 @@ describe("calcite-dialog", () => {
     const dialog = await page.find("calcite-dialog");
     await page.waitForChanges();
     expect(dialog).toHaveAttribute("open");
-    expect(dialog).toHaveAttribute("open");
+    await page.waitForChanges();
```
jcfranco, line 719: Is this necessary? There's a previous call to `waitForChanges` and no prop is modified before this call.

Elijbet:
> Could you also coordinate `input-date-picker` changes w/ @anveshmekala? Additional coverage was added all date-picker-related test suites (see [#8402 (files)](https://github.com/Esri/calcite-design-system/pull/8402/files)).

Checked with him and there are no conflicts so far.

packages/calcite-components/src/components/dialog/dialog.e2e.ts
Elijbet: I had to add a timeout there and some other places, otherwise tests fail.

packages/calcite-components/src/components/dialog/dialog.e2e.ts
```diff
@@ -697,6 +700,7 @@ describe("calcite-dialog", () => {
 
     dialog.setProperty("open", true);
     await page.waitForChanges();
+    await page.waitForTimeout(delayInMilliseconds);
```
driskull, line 703: maybe `waitForAnimationFrame()` would work instead?

packages/calcite-components/src/components/dialog/dialog.e2e.ts
```diff
@@ -711,22 +715,20 @@ describe("calcite-dialog", () => {
...
   it("closes when Escape key is pressed and dialog is open on page load", async () => {
     const page = await newE2EPage();
-    await page.setContent(`<calcite-dialog  open></calcite-dialog>`);
+    await page.setContent(`<calcite-dialog open></calcite-dialog>`);
 
     const dialog = await page.find("calcite-dialog");
     await page.waitForChanges();
     expect(dialog).toHaveAttribute("open");
-    expect(dialog).toHaveAttribute("open");
+    await page.waitForTimeout(delayInMilliseconds);
```
driskull, line 723: maybe `waitForAnimationFrame()` would work instead?

packages/calcite-components/src/components/dialog/dialog.e2e.ts
```diff
@@ -735,9 +737,11 @@ describe("calcite-dialog", () => {
     await skipAnimations(page);
     const dialog = await page.find("calcite-dialog");
     const container = await page.find(`calcite-dialog >>> .${CSS.container}`);
+    await page.waitForChanges();
 
     dialog.setProperty("open", true);
     await page.waitForChanges();
+    await page.waitForTimeout(delayInMilliseconds);
```
driskull, line 744: maybe `waitForAnimationFrame()` would work instead?

packages/calcite-components/src/components/modal/modal.e2e.ts
```diff
@@ -389,38 +391,47 @@ describe("calcite-modal", () => {
...
+
     await page.keyboard.press("Escape");
     await page.waitForChanges();
     expect(await modal.isVisible()).toBe(false);
     expect(await modal.getProperty("open")).toBe(false);
-    await modal.setProperty("open", true);
+
+    modal.setProperty("open", true);
     await page.waitForChanges();
+    await page.waitForTimeout(delayInMilliseconds);
```
driskull, line 413: maybe `await waitForAnimationFrame()` would work instead?

packages/calcite-components/src/components/modal/modal.e2e.ts
```diff
@@ -389,38 +391,47 @@ describe("calcite-modal", () => {
...
 
   it("closes when Escape key is pressed and modal is open on page load", async () => {
     const page = await newE2EPage();
-    await page.setContent(`<calcite-modal  open></calcite-modal>`);
+    await page.setContent(`<calcite-modal open></calcite-modal>`);
+    await skipAnimations(page);
+
     const modal = await page.find("calcite-modal");
     await page.waitForChanges();
+    await page.waitForTimeout(delayInMilliseconds);
```
driskull, line 424: maybe `await waitForAnimationFrame()` would work instead?

packages/calcite-components/src/components/modal/modal.e2e.ts
```diff
@@ -389,38 +391,47 @@ describe("calcite-modal", () => {
...
     expect(modal).toHaveAttribute("open");
-    expect(modal).toHaveAttribute("open");
+
     await page.keyboard.press("Escape");
     await page.waitForChanges();
+    await page.waitForTimeout(delayInMilliseconds);
```
driskull, line 429: maybe `await waitForAnimationFrame()` would work instead?

packages/calcite-components/src/demos/input-date-picker.html
```diff
@@ -57,6 +57,7 @@
             Input Date Picker Label
             <calcite-input-date-picker scale="s" value="2023-03-07"></calcite-input-date-picker>
           </calcite-label>
+          <div id="next-sibling" tabindex="0">next sibling</div>
```
driskull, line 60: Is this relevant?

driskull (approved):
This looks good to me 👍

packages/calcite-components/src/components/modal/modal.e2e.ts
Elijbet: This is the only place where this substitution works.

driskull: Interesting. I think we could live with the delay but I'm curious on what its waiting for exactly.

Elijbet: I had to add them after this commit: [`6bf5a5f` (#9231)](https://github.com/Esri/calcite-design-system/pull/9231/commits/6bf5a5fc531db43d9bb4f94d9cc98157226b08bb)

packages/calcite-components/src/tests/globalStyles.e2e.ts
```diff
@@ -92,4 +94,130 @@ describe("global styles", () => {
...
+
+        async function testEscapeAndCheckOpenState(elements: E2EElement[]) {
+          for (let i = 0; i < elements.length; i++) {
+            await page.keyboard.press("Escape");
+            await page.waitForChanges();
+            expect(await elements[i].getProperty("open")).toBe(false);
+
+            for (let j = 0; j < elements.length; j++) {
+              const expectedOpenState = j > i;
+              expect(await elements[j].getProperty("open")).toBe(expectedOpenState);
```
jcfranco, line 211: Can we also assert on the focused component?

packages/calcite-components/src/utils/focusTrapComponent.ts
jcfranco, line 74: This setup seems a bit confusing to me. We’re passing options when setting up the focus trap, but some are also defined on the component itself. This could could cause confusion or issues (e.g., `onDeactivate` being set in both `component.onFocusTrapDeactivate` and `options.onFocusTrapDeactivate`). Can we roll it back to just rely on the component passing options for now? I'd like us to revisit this approach later.

packages/calcite-components/src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -432,6 +412,11 @@ export class InputTimePicker
   /** Fires when the component is open and animation is complete. */
   @Event({ cancelable: false }) calciteInputTimePickerOpen: EventEmitter<void>;
 
+  @Listen("calcitePopoverClose")
```
jcfranco, line 415: Can you use the JSX event property instead of using `@Listen`?

packages/calcite-components/src/components/input-date-picker/input-date-picker.e2e.ts
```diff
@@ -453,11 +454,13 @@ describe("calcite-input-date-picker", () => {
         await resetFocus(page);
         await startInput.click();
         await page.waitForChanges();
+        expect(openSpy).toHaveReceivedEventTimes(1);
```
jcfranco, line 457: I should've spotted this before, but why are we asserting on the open/close events? Doesn't the `openClose` test helper cover this? Applies to `input-time-picker` as well.

packages/calcite-components/src/components/popover/popover.tsx
```diff
@@ -293,7 +293,26 @@ export class Popover
...
+        clickOutsideDeactivates: (event: MouseEvent) => {
+          const path = event.composedPath();
+          const isReferenceElementInPath =
+            this.effectiveReferenceElement instanceof EventTarget &&
+            path.includes(this.effectiveReferenceElement);
+
+          const outsideClick = !path.includes(this.el);
+          const shouldCloseOnOutsideClick = this.autoClose && outsideClick;
+
+          return (
```
jcfranco, line 309: This can be simplified to:

```tsx
return shouldCloseOnOutsideClick && 
    (this.triggerDisabled || isReferenceElementInPath);
```

packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -231,7 +230,18 @@ export class Dialog
     this.mutationObserver?.observe(this.el, { childList: true, subtree: true });
     connectLocalized(this);
     connectMessages(this);
-    connectFocusTrap(this);
+    connectFocusTrap(this, {
+      focusTrapOptions: {
+        clickOutsideDeactivates: false,
```
jcfranco, line 235: Can you add a comment on why we set this to `false` here? Applies to modal. It might be helpful for other options too if it's not clear.

jcfranco (changes requested):
Great job, @Elijbet! Just one more round of changes and I think we'll be good to land. 🛬

packages/calcite-components/src/components/popover/popover.tsx
```diff
@@ -293,7 +293,26 @@ export class Popover
     this.setFilteredPlacements();
     connectLocalized(this);
     connectMessages(this);
-    connectFocusTrap(this);
+    connectFocusTrap(this, {
+      focusTrapEl: this.el,
+      focusTrapOptions: {
+        allowOutsideClick: true,
+        clickOutsideDeactivates: (event: MouseEvent) => {
```
driskull, line 300: nitpick: what about making these anonymous functions private methods on the component? Applies to all the `clickOutsideDeactivates` options that are functions.

May also be beneficial for `onDeactivate` too.

driskull (approved):
🎉 🌮 Nice work!

packages/calcite-components/src/components/input-date-picker/input-date-picker.e2e.ts
```diff
@@ -373,19 +373,18 @@ describe("calcite-input-date-picker", () => {
 
       it("toggles the date picker when clicked", async () => {
         let calendar = await page.find(`calcite-input-date-picker >>> .${CSS.calendarWrapper}`);
```
jcfranco, line 375: Can you roll these changes back? This file now only has stylistic changes, so they are no longer needed. Applies to `globalStyles.e2e.ts` too.

packages/calcite-components/src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -571,16 +541,13 @@ export class InputTimePicker
...
-    deactivateFocusTrap(this, {
-      onDeactivate: () => {
-        this.calciteInputEl.setFocus();
-        this.focusOnOpen = false;
-      },
-    });
     this.open = false;
   };
 
+  onFocusTrapDeactivate(): void {
```
jcfranco, line 547: This method is not referenced.

packages/calcite-components/src/components/popover/popover.tsx
```diff
@@ -522,6 +529,22 @@ export class Popover
     this.reposition(true);
   };
 
+  private clickOutsideDeactivates = (event: MouseEvent): boolean => {
```
jcfranco, line 532: Can you confirm callbacks passed to `connectFocusTrap` options need to be `this`-bound?

packages/calcite-components/src/tests/stackedFocusTrap.e2e.ts
```diff
@@ -0,0 +1,127 @@
...
+      </div>
+    </calcite-modal>
+  `;
+
+  it("closes a stack of open components sequentially in visual order", async () => {
+    const page = await newE2EPage();
+    await page.setContent(componentStack);
+    await skipAnimations(page);
+
+    async function testStackEscapeSequence(page: E2EPage, pickerType: string) {
```
jcfranco, line 60: Can you add return types?

packages/calcite-components/src/tests/globalStyles.e2e.ts
```diff
@@ -92,4 +94,130 @@ describe("global styles", () => {
...
+
+        const popover = await page.find("calcite-popover");
+        await openAndCheckVisibility(popover);
+
+        const inputPicker = await page.find(pickerType);
+        inputPicker.click();
+        await page.waitForChanges();
+        expect(await inputPicker.getProperty("open")).toBe(true);
+
+        async function testEscapeAndCheckOpenState(elements: E2EElement[]) {
```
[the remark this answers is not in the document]
jcfranco: This is unresolved in the new test file.

packages/calcite-components/src/utils/focusTrapComponent.ts
```diff
@@ -58,8 +58,6 @@ export function connectFocusTrap(component: FocusTrapComponent, options?: Connec
   }
 
   const focusTrapOptions: FocusTrapOptions = {
-    clickOutsideDeactivates: true,
```
jcfranco, line 61: Can we confirm falling back on the defaults is desired (e.g., `clickOutsideDeactivates: false`, `escapeDeactivates: true`)?

packages/calcite-components/src/tests/stackedFocusTrap.e2e.ts
```diff
@@ -0,0 +1,127 @@
...
+      async function openAndCheckVisibility(element: E2EElement) {
+        element.setProperty("open", true);
+        await page.waitForChanges();
+        expect(await element.isVisible()).toBe(true);
```
jcfranco, line 64: I think you can omit this since you are testing for openness of all components in `testEscapeAndAssertOpenState`.

packages/calcite-components/src/tests/stackedFocusTrap.e2e.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  it("closes a stack of open components sequentially in visual order", async () => {
+    const page = await newE2EPage();
+    await page.setContent(componentStack);
+    await skipAnimations(page);
+
+    async function testStackEscapeSequence(page: E2EPage, pickerType: string) {
+      async function openAndCheckVisibility(element: E2EElement) {
+        element.setProperty("open", true);
+        await page.waitForChanges();
```
jcfranco, line 63: For stability, can we also await on the open event here?

packages/calcite-components/src/components/input-date-picker/input-date-picker.e2e.ts
```diff
@@ -385,7 +385,6 @@ describe("calcite-input-date-picker", () => {
         await inputDatePicker.click();
         await page.waitForChanges();
         calendar = await page.find(`calcite-input-date-picker >>> .${CSS.calendarWrapper}`);
-
```
jcfranco, line 388: @Elijbet This file still has a stylistic change.

Elijbet: Sorry I'm still working on this.

packages/calcite-components/src/components/popover/popover.tsx
Elijbet: Confirmed. Fails tests without.

packages/calcite-components/src/utils/focusTrapComponent.ts
Elijbet: I think we do want escape to deactivate by default on all components if they are focus trap components, and `clickOutsideDeactivates` is false on all except `popover`: under certain conditions. Added a comment for `sheet` as well to make it clear. 

focusTrapOptions: {
        // Scrim has it's own close handler, allow it to take over.
        clickOutsideDeactivates: false,

packages/calcite-components/src/tests/stackedFocusTrap.e2e.ts
```diff
@@ -0,0 +1,130 @@
...
+              await page.keyboard.press("Tab");
+            }
+          }
+          await page.waitForChanges();
+
+          const activeElementIdAfterTab = await page.evaluate(() => document.activeElement?.id);
+
+          const expectedElementId =
+            focusTrapOrderElements[i].id === "sheet"
+              ? "sheet-button"
```
jcfranco, line 92: This can be done in a follow-up refactor PR, but it'd be helpful to add a comment explaining why we check for `sheet-button` when `sheet` is the next element in the list.

jcfranco (approved):
✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅
✅✔✅✅✅✔✅✔✅✅✅✔✅✅✔✔✔✅✅✔✔✔✅✅✔✔✔✅✔✅
✅✅✔✅✔✅✅✔✅✅✅✔✅✔✅✅✅✅✔✅✅✅✅✔✅✅✅✅✔✅
✅✅✅✔✅✅✅✔✅✅✅✔✅✅✔✔✅✅✅✔✔✅✅✅✔✔✅✅✔✅
✅✅✅✔✅✅✅✔✅✅✅✔✅✅✅✅✔✅✅✅✅✔✅✅✅✅✔✅✅✅
✅✅✅✔✅✅✅✅✔✔✔✅✅✔✔✔✅✅✔✔✔✅✅✔✔✔✅✅✔✅
✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅

excited

packages/calcite-components/src/tests/stackedFocusTrap.e2e.ts
Elijbet: I'll add it now.

Elijbet:
This is the gist of the test @DitwanP, @geospatialem.

Whenever a focus-trapping component opens, it traps focus and gets added to the stack. Pressing escape should deactivate each returning focus to the next component in the stack. 
 
I recommend setting up a stack of all focus-trap components as open, mimicking the setup in the e2e, but have 2 separate stacks with popover containing either input-time-picker or input-date-picker, as you can't have both open at the same time. 
 
Click outside also needs to be tested. It's usually set to false. For input-date-picker and input-time-picker closing with click outside is delegated to popover: https://github.com/Esri/calcite-design-system/pull/9231/files#diff-42c23e5a429472ae2ecb636c2ba383162681d40c5ff70c3f89be55d376052e32R1002
So whatever the expected behavior for popover is in this case.

Issue #9235, opened 2024-05-01
Some properties (e.g. `inputMode` not available anymore)


hcampos-professional:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

While testing the latest v2.8.0 of Calcite Components together with the JSAPI, the Typescript compiler reported that `inputMode` is no longer available as a property of `<calcite-input-number>`. It seems like the change got introduced by https://github.com/Esri/calcite-design-system/commit/13ba95375a9cb9091edf469d45827bd8374098b9 and it could affect apps using that property or any of the other properties that were removed.			

### Expected Behavior

Typings should still contain the previously-existing properties. Alternatively, a breaking change and consequent major version should be released.

### Reproduction Sample

https://codepen.io/hccampos/pen/ExJzRRq

### Reproduction Steps

1. Open https://codepen.io/hccampos/pen/ExJzRRq
2. Inspect the input element.
3. Type `$0.inputMode` in the console and notice that the `inputMode` property isn't set.

Alternatively, build an application which makes use of the typings and notice that the property isn't available.

### Reproduction Version

2.8.0

### Regression?

2.7.x

### Priority impact

p0 - emergency

### Impact

While it doesn't directly impact my team (we haven't decided if we'll upgrade to this version of Calcite for the upcoming release), it can potentially affect any application which was using one of the removed properties. Their builds can fail and the components may not work as expected.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

geospatialem:
Closing out the above as the global attributes are provided and the issue for Maps SDK for JS appears to be [related to the maquette VDOM library typings for globals](https://github.com/Esri/calcite-design-system/pull/9236#issuecomment-2089253034).

jcfranco:
One thing that https://github.com/Esri/calcite-design-system/pull/9123 overlooked was how our custom global component prop's were using the kebabified™ case vs the native ones. I think we can add props (`@internal`) for `enterKeyHint` and `inputMode` to allow the VDOM typing tool referenced above to do proper attribute mapping. We can drop support for the kebab-case attributes in 3.x.

DitwanP:
[...]

PR #9123, opened 2024-04-12
chore(linting): restore some ESLint rules

jcfranco:
**Related Issue:** N/A

## Summary

Restores the following rules:

* `@jest/expect-expect`
* `@stencil-community/decorators-context`
* `@stencil-community/strict-boolean-conditions`
* `@stencil-community/reserved-member-names`
* `@stencil-community/ban-exported-const-enums`
* `@typescript-eslint/no-duplicate-enum-values`

Disables the following: 

* `@esri/calcite-components/strict-boolean-attributes` (replaced by `@stencil-community/ban-default-true`)

PR #9236, opened 2024-05-01
revert: restore some ESLint rules

driskull:
**Related Issue:** #9235

## Summary

- partially reverts #9123. Basically, just the TSX changes where props were removed. These are all global attributes so they should be settable on any element. However, since we removed them as properties, they won't have any typings or doc associated with them which teams may think is a breaking change. So should we put these properties back and remove them at a major (breaking change) release.
- Added `"@stencil-community/reserved-member-names": "warn",` to eslint config because eslint was preventing commits due to these props being reserved global attributes.
- All the other eslint cleanup was kept and not reverted

### DIFF
--- a/packages/calcite-components/.eslintrc.cjs
+++ b/packages/calcite-components/.eslintrc.cjs
@@ -48,6 +48,7 @@ module.exports = {
     "@stencil-community/own-methods-must-be-private": "off",
     "@stencil-community/own-props-must-be-private": "off",
     "@stencil-community/prefer-vdom-listener": "warn",
+    "@stencil-community/reserved-member-names": "off",
     "@stencil-community/required-jsdoc": "off",
     "@stencil-community/strict-boolean-conditions": "off",
     "@typescript-eslint/ban-types": "warn",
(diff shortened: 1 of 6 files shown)


driskull:
I was wrong initially, these are all global attributes.

driskull:
Closing, I think the maquette VDOM library just needs to update its typings to support inputMode, etc.

benelan (approved):
:+1: 

The `revert` type should only be used when fully reverting a previous commit. Maybe we could use `docs` or `chore` instead, since that was the problem this PR resolves? For example:

> chore: restore ESLint rule that removed ref for global attributes

benelan:
Imo we should still have these documented if our components don't support **all** global attributes.

driskull:
> Imo we should still have these documented if our components don't support all global attributes.

Could you clarify this @benelan? What do you mean. Our components are web components which are HTML Elements. They do support all global attributes because of that nature.

benelan:
Okay cool, I wasn't sure if all global attributes were handled correctly or not since this PR was restoring custom logic for two of them. But if all global attributes are handled correctly then we don't need to document them.

Issue #4701, opened 2022-06-09
[Tile] Fix spacing of content-start/end slot when used without title, description, icon


macandcheese:
### Description

The tile component has extra leading / trailing padding when only the single slot is present, and a "double" padding between the slots when both are present. 

### Acceptance Criteria

We should ensure that when:
1) a single slot is populated and no title, description, icon are present: padding is consistent around the slot, and:
2) when both are populated - the spacing between the two slots is the same as from the edge of the slot to the component edge.

### Relevant Info

<img width="892" alt="Screen Shot 2022-06-09 at 9 28 34 AM" src="https://user-images.githubusercontent.com/4733155/172898623-7f19e914-2b80-4726-a638-6c51c69076fa.png">

### Which Component

Tile

### Example Use Case

https://codepen.io/mac_and_cheese/pen/gOvBqxy?editors=100

SkyeSeitz:
Link to Figma specs: https://www.figma.com/file/8CiAVHkXIMTsI9hTSbotEz/Tile-Redesign-%5BEpic-6662%5D?type=design&node-id=720%3A2738&mode=dev

jcfranco:
@eriklharper [E2E tests are failing](https://github.com/Esri/calcite-design-system/actions/runs/8916683873/job/24488529692?pr=9241), so it looks like tests might also need updating.

geospatialem:
It looks like slotted content, such as images don't appear to be adjusted, which _might_ be expected. But it doesn't look like devs could adjust the `tile` or `tile-group` CSS to accommodate width.

@eriklharper Wondering if we need to explore more options, or open a new issue? For instance, with `2.9.0-next.9` and the following sample: https://codepen.io/geospatialem/pen/GRaJdwE:

![image](https://github.com/Esri/calcite-design-system/assets/5023024/63b008bf-7aa3-47da-b739-500512f76729)

PR #9241, opened 2024-05-02
fix(tile): fix spacing between content-top and content-bottom slots

eriklharper:
**Related Issue:** #4701 

## Summary

This PR corrects the spacing between the `content-top` and `content-bottom` slotted content in Tile, ensuring there is a scale-specific gap between the two when both of them are present and none of the other content elements are supplied.

image


packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -117,7 +117,7 @@
   .container {
     padding: var(--calcite-spacing-sm);
   }
-  .column,
+  .column.has-content,
```
jcfranco, line 120: To reduce specificity and for consistency, can we change the state modifiers to be more BEM-like? For example:

```scss
.column--has-content,
.row { /* ... */ }
```

Applies to other state modifier classes.

packages/calcite-components/src/components/tile/tile.tsx
```diff
@@ -214,6 +224,12 @@ export class Tile implements InteractiveComponent, SelectableComponent {
     this.calciteTileSelect.emit();
   };
 
+  private handleSlotChange = (event: Event): void => {
+    const slotNameCamelCased = camelCase((event.target as HTMLSlotElement).name);
```
jcfranco, line 228: Could you use a `data-attribute` on `<slot>` with the corresponding state prop name instead of building it?

packages/calcite-components/src/components/tile/tile.tsx
```diff
@@ -312,6 +340,8 @@ export class Tile implements InteractiveComponent, SelectableComponent {
           : interactive
             ? "button"
             : undefined;
+    const hasContent = Boolean(description || hasContentEnd || hasContentStart || heading || icon);
```
jcfranco, line 343: Nitpick: for consistency, can you use `!!` instead? Both are valid, but maybe we should look into having ESLint enforce one vs the other. cc @driskull

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -129,6 +129,9 @@
     font-size: var(--calcite-font-size--3);
     line-height: 0.85938rem;
   }
+  .has-only-content-top-and-bottom slot[name="content-top"]::slotted(*) {
```
jcfranco, line 132: Using an internal CSS prop to assign the scale-specific might reduce selector duplication.

```scss
:host([scale="m"]) {
  --calcite-internal-content-top-spacing: var(--calcite-spacing-md);
  /* ... */
}

/* ... */

.content--has-top-and-bottom-only slot[name="content-top"]::slotted(*) {
  margin-block-end: var(--calcite-internal-content-top-spacing);
}
```

jcfranco (approved):
Once comments have been addressed, this is good to murge!

jcfranco:
@eriklharper [E2E tests are failing](https://github.com/Esri/calcite-design-system/actions/runs/8916683873/job/24488529692?pr=9241), so it looks like tests might also need updating.

packages/calcite-components/src/components/tile/tile.tsx
driskull: I agree. `!!` is preferred since its much shorter and easier to read.

eriklharper:
> @eriklharper [E2E tests are failing](https://github.com/Esri/calcite-design-system/actions/runs/8916683873/job/24488529692?pr=9241), so it looks like tests might also need updating.

Fixed it.  Had to do with removing a conditional for rendering the heading and description divs, which is not necessary for this PR, was a holdover from a previous experiment that I forgot to revert.

PR #9245, opened 2024-05-02
fix(input, input-number, input-text): restore `autofocus`, `enter-key-mode` and `input-mode` attributes

jcfranco:
**Related Issue:** #9235

## Summary

This fixes the following issues introduced by #9123 when component props that shadowed global attributes were removed:

* the kebab-cased attribute names of those props would no longer work (namely, `enter-key-mode`, `input-mode`)
* `autofocus` was missing from component types as Stencil does not include it in the base `HTMLAttribute` type (see https://github.com/ionic-team/stencil/issues/5726).

### DIFF
--- a/packages/calcite-components/src/components/input-number/input-number.tsx
+++ b/packages/calcite-components/src/components/input-number/input-number.tsx
@@ -120,6 +120,14 @@ export class InputNumber
   @Prop({ reflect: true }) alignment: Extract<"start" | "end", Alignment> = "start";
 
   /**
+   * Adds global prop, missing from Stencil's `HTMLElement` type, see https://github.com/ionic-team/stencil/issues/5726
+   *
+   * @internal
+   */
+  // eslint-disable-next-line @stencil-community/reserved-member-names
+  @Prop() autofocus: boolean;
+
+  /**
    * When `true`, a clear button is displayed when the component has a value.
    */
   @Prop({ reflect: true }) clearable = false;
@@ -137,9 +145,9 @@ export class InputNumber
   }
 
   /**
-   * Custom global prop added to support kebab-cased attribute from custom global prop removed in https://github.com/Esri/calcite-design-system/pull/9123
+   * Adds support for kebab-cased attribute, removed in https://github.com/Esri/calcite-design-system/pull/9123
    *
-   * @futureBreaking kebab-case of this attribute will not be supported in a future release
+   * @futureBreaking kebab-cased attribute will not be supported in a future release
    * @internal
    */
   // eslint-disable-next-line @stencil-community/reserved-member-names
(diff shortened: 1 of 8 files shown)


packages/calcite-components/src/components/input-number/input-number.tsx
```diff
@@ -1059,8 +1085,8 @@ export class InputNumber
         autofocus={this.el.autofocus ? true : null}
         defaultValue={this.defaultValue}
         disabled={this.disabled ? true : null}
-        enterKeyHint={this.el.enterKeyHint}
-        inputMode={this.el.inputMode}
+        enterKeyHint={this.el.enterKeyHint || this.el.getAttribute("enterkeyhint")}
```
driskull, line 1088: This looks a little odd to me because you could just do

```tsx
enterKeyHint={this.enterKeyHint}
```

But I get why its like the way you have. So its easier to remove the property in the future breaking change release 👍

packages/calcite-components/src/components/input/input.tsx
```diff
@@ -1176,15 +1202,23 @@ export class Input
...
                 [CSS.inlineChild]: !!this.inlineEditableEl,
               }}
               defaultValue={this.defaultValue}
               disabled={this.disabled ? true : null}
-              enterKeyHint={this.el.enterKeyHint}
-              inputMode={this.el.inputMode}
+              enterKeyHint={
+                this.el.enterKeyHint ||
+                this.el.getAttribute("enterkeyhint") ||
+                this.el.getAttribute("enter-key-hint")
```
driskull, line 1215: this one is here because it could be nested in another component right?

driskull (approved):
Nice! 👏

packages/calcite-components/src/components/input/input.tsx
jcfranco: How do you mean?

This does catch an inconsistency with the fix. I was sure I updated all inputs to be the same. Fixing!

driskull: Yeah i was just wondering why this is the only one that has the `enter-key-hint` kebab casing.

packages/calcite-components/src/components/input/input.tsx
```diff
@@ -1176,15 +1205,15 @@ export class Input
...
               class={{
                 [CSS.editingEnabled]: this.editingEnabled,
                 [CSS.inlineChild]: !!this.inlineEditableEl,
               }}
               defaultValue={this.defaultValue}
               disabled={this.disabled ? true : null}
-              enterKeyHint={this.el.enterKeyHint}
-              inputMode={this.el.inputMode}
+              enterKeyHint={enterKeyHint}
+              inputMode={inputMode}
```
driskull, line 1216: Yeah this makes more sense. We'll just have to put the previous code back once we remove the internal props.

packages/calcite-components/src/components/input-number/input-number.tsx
jcfranco: This is actually needed since the component prop won't be populated if I set the global attribute. The custom props override the native behavior, so the prop will not be set when using the native global attributes.

Issue #6565, opened 2023-03-04
[Accordion Item] Stretch slotted actions to fill height of component


macandcheese:
### Description

Noticed this while testing out https://github.com/Esri/calcite-components/issues/6561 / https://github.com/Esri/calcite-components/issues/2884#issuecomment-1454349330.

Slotted actions don't stretch to fill the height:

<img width="747" alt="Screen Shot 2023-03-04 at 10 49 10 AM" src="https://user-images.githubusercontent.com/4733155/222923774-74958e7b-01de-48ae-931b-2c734554a487.png">

### Acceptance Criteria

We should follow how we handle this in other components - stretching the actions to fill when we expect the component as a slotted child, for example in Alert:

<img width="871" alt="Screen Shot 2023-03-04 at 10 52 27 AM" src="https://user-images.githubusercontent.com/4733155/222923802-f46f47e3-1556-48eb-a43a-1b42bedcf777.png">

### Which Component

Accordion Item

### Esri team

Calcite (design)

geospatialem:
Adding context to the above - once the above lands ✈️  we have a follow-up internal documentation update (issue 703).

DitwanP:
🍠 Verified on `2.14.0-next.6`
![image](https://github.com/user-attachments/assets/819dbc90-eb79-46b0-b79c-1b1ba2aad5db)

PR #9250, opened 2024-05-02
feat(accordion-item): stretch slotted actions to fill its height

josercarcamo:
**Related Issue:** #6565 

## Summary
Made the height of the action item 100% so that it fills its available height.


packages/calcite-components/src/components/accordion-item/accordion-item.scss
```diff
@@ -192,6 +192,10 @@
   @apply flex items-center;
 }
 
+slot[name="actions-start"] {
+  block-size: 100%;
+}
```
driskull, line 197: Is this working @josercarcamo? Do we need some new screenshot tests to account for these changes?

I would have expected either `align-items:stretch` or `align-self:stretch` needs to be used to stretch the items by either using the `::slotted` selector or some element within the shadow DOM of the component.

Something like: https://github.com/Esri/calcite-design-system/blob/b069284abb46dfd9b8384337f3825f5425f9d2b9/packages/calcite-components/src/components/list-item/list-item.scss#L78

or

https://github.com/Esri/calcite-design-system/blob/b069284abb46dfd9b8384337f3825f5425f9d2b9/packages/calcite-components/src/components/list-item/list-item.scss#L190-L193

driskull (commented):
comments!

packages/calcite-components/src/components/accordion-item/accordion-item.scss
josercarcamo: Yes, it is working, at least with manual testing. I tried the "align-items: stretch" but it didn't work in this case. I actually tried to use the solution in alert.css as shown below, but it didn't work.

`.actions-end {
  @apply flex self-stretch;
}`

driskull:
How's this one going @josercarcamo?

packages/calcite-components/src/components/accordion-item/accordion-item.scss
driskull: I'm just wondering why this is doing anything. A slot element is just a placeholder and really isn't an element that needs to be styled.

josercarcamo:
@driskull as we discussed on teams, I'll pick this up when the September release nears or I'll try to squeeze it in during down time in some other release.

alisonailea:
> ry to squeeze it in during down time in some other release.

alisonailea:
> I'll pick this up when the September release nears or I'll try to squeeze it in during down time in some other release.

Sorry for accidental churn. I think I missed this discussion. Can we add the September milestone tag to this? @geospatialem

geospatialem:
> Sorry for accidental churn. I think I missed this discussion. Can we add the September milestone tag to this? @geospatialem

@alisonailea The issue was originally in September, but the above PR was initialized before the milestone started in May. We should be covered for the future work already. 💪🏻

packages/calcite-components/src/components/accordion-item/accordion-item.scss
```diff
@@ -147,7 +147,10 @@
 
 .actions-start,
 .actions-end {
-  @apply flex items-center;
+  @apply flex items-stretch;
+  * {
```
jcfranco, line 151: I think we can skip the `*` rule because:

1. These slots are specifically for actions, which don't have children.
2. Overriding user styles for slotted content should be avoided.

jcfranco (changes requested):
Looking good, @josercarcamo! 😎

packages/calcite-components/src/components/accordion-item/accordion-item.scss
josercarcamo: The thing is that https://github.com/Esri/calcite-design-system/blob/7ddbffb7deba19f84e815a1471186cee9164a9ff/packages/calcite-components/src/components/accordion-item/accordion-item.scss#L92 causes all items to have "items-center" so it has to be overridden which is what this line change does. How should we proceed?

jcfranco: Apologies! I thought this was targeting the slotted content. 😅

Sidebar: I still think we can tidy this up to avoid having to override the universal selector elsewhere. We should consider swapping out the universal selector for explicit ones in a follow-up PR.

jcfranco (approved):
Code changes LGTM!

jcfranco:
I forgot to mention, before installing, could you update the story to ensure it catches any regression if this style is changed or removed? I didn’t notice any difference when I tested by removing it locally and comparing.

Issue #6371, opened 2023-01-28
[Action] - Remove stale tooltip slot


macandcheese:
### Description

The Action component still contains code for the now-outdated use case of slotting a tooltip in a “tooltip” slot. There is also stale example code showing this pattern.

### Proposed Advantages
Removing these outdated patterns will clean the component code, and no longer documents an unsupported pattern.

### Which Component
Action

### Relevant Info
From Teams channel feedback cc @jcfranco

geospatialem:
The slot is no longer recommended, the updated code is the following - for future documentation/release summary:

```html
<calcite-action id="action-with-tooltip" text="Layers" icon="layers" text-enabled></calcite-action>
<calcite-tooltip reference-element="action-with-tooltip">
    <span>More information about layers</span>
</calcite-tooltip>
```

DitwanP:
🍡 Verified

PR #9270, opened 2024-05-06
docs(action): deprecate the slot tooltip

josercarcamo:
**Related Issue:** #6371

## Summary
Deprecated slot tooltip


packages/calcite-components/src/components/action/action.tsx
```diff
@@ -42,7 +42,7 @@ import { CSS, SLOTS } from "./resources";
 
 /**
  * @slot - A slot for adding a `calcite-icon`.
- * @slot tooltip - A slot for adding a `calcite-tooltip`.
+ * @slot tooltip - [Deprecated] use `id` instead. A reference to a `calcite-tooltip`.
```
jcfranco, line 45: The deprecation message should suggest using a tooltip instead. +@geospatialem @DitwanP for doc guidance.

jcfranco (changes requested):
Since this is merely updating the doc, this should use the `docs` commit type. See https://github.com/ionic-team/stencil/blob/main/CONTRIBUTING.md#type for reference (referenced in our contributing doc).

packages/calcite-components/src/components/action/action.tsx
geospatialem: WDYT about the following for folks? cc: @DitwanP

 ```suggestion
 * @slot tooltip - [Deprecated] Use the `calcite-tooltip` component instead.
```

jcfranco (approved):
Doc changes look good! Once the PR title is updated to use the [`docs` type (vs `fix`)](https://github.com/Esri/calcite-design-system/pull/9270#pullrequestreview-2041861379), this should be good to merge!

Issue #4461, opened 2022-04-27
calcite-select: set initial value based on `value` attr without `selected` options


jgibson02:
### Description

Currently, if a `calcite-select` has an initial `value` set, and that matches a value for one of its `calcite-option`s, the `calcite-option` must also have the `selected` attribute set for the `calcite-select` to initially reflect that value.

~https://codesandbox.io/s/calcite-select-initial-value-w-o-selected-z4dt3f~
https://codepen.io/driskull/pen/poqXEEv

### Acceptance Criteria

[ ] When a `calcite-select` has an initial `value` that matches a `calcite-option` child's `value`, the `calcite-select` should reflect that value

### Which Component

calcite-select

### Example Use Case

~https://codesandbox.io/s/calcite-select-initial-value-w-o-selected-z4dt3f~
https://codepen.io/driskull/pen/poqXEEv

PR #9272, opened 2024-05-07
fix(select): fix option selection via initial `value`

jcfranco:
**Related Issue:** #4461

## Summary

This fixes an issue where the initial `value` would not match any `option`.

### DIFF
--- a/packages/calcite-components/src/components/select/select.tsx
+++ b/packages/calcite-components/src/components/select/select.tsx
@@ -27,7 +27,7 @@ import {
   InteractiveContainer,
   updateHostInteraction,
 } from "../../utils/interactive";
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/select/select.tsx
```diff
@@ -279,6 +282,12 @@ export class Select
     this.setFocus();
   }
 
+  private updateItemsFromValue(value: string): void {
+    this.el
+      .querySelectorAll("calcite-option")
+      .forEach((item) => (item.selected = item.value === value));
```
eriklharper, line 288: Should we use `slotEl.assignedElements` here?

driskull (approved):
LGTM! 🎉

Issue #6497, opened 2023-02-20
Avatars of different users (different usernames) have the same background


Anastasiia-Boleiko:
### Actual Behavior

Avatars of different users (with different `username`s) have the same background colors in some cases.
<img width="1059" alt="Screenshot 2023-02-20 at 17 54 44" src="https://user-images.githubusercontent.com/119309703/220164379-32a1b9bf-8487-4617-a337-88ee0cdb6c44.png">

In the docs it was written:

> A randomly generated AA accessible background will fill behind the initials. That generated fill will stay assigned to the user until their account is deleted or an image is used.

### Expected Behavior

Avatars of different users (with different `username`s) have different background colors.

### Reproduction Sample

https://codepen.io/anastasiia-zrh/pen/wvEMKxq

### Reproduction Steps

1. Add several calcite-avatar elements to a page
2. Specify different `username`s (for example: `john_doe`, `john_doe1`, `john_doe2`)

### Reproduction Version

v1.0.5

### Impact

Initials and avatar background simplify finding a needed user (it's like a visual mapping where color+initials=user, which decreases cognitive load) for those, who do not use AT. 
It would be great to have different avatar backgrounds for different users.

### Esri team

ArcGIS Urban

Anastasiia-Boleiko:
If the described case is an expected behavior from your perspective, maybe it would be possible to implement it as an enhancement of the calcite-avatar component. Thanks a lot in advance!

PR #9277, opened 2024-05-07
feat(avatar): ensure different backgrounds are generated for similar usernames, full names and user IDs

jcfranco:
**Related Issue:** #6497

## Summary

This updates `avatar` to create unique background colors for similar name strings. This is done by mixing the input string in a deterministic way before building the hash and mapping the color.


packages/calcite-components/src/components/avatar/utils.ts
```diff
@@ -21,6 +24,23 @@ export function stringToHex(str: string): string {
...
+ * The function splits the string into two halves, reverses each half, and then concatenates them.
+ *
+ * @param {string} string - The input string to be mixed.
+ * @returns {string} - The mixed string.
+ */
+function mixStringDeterministically(string: string): string {
+  const midPoint = Math.floor(string.length / 2);
+  let mixedString = "";
+
+  for (let i = 0; i < string.length; i++) {
```
paulcpederson, line 37: if you're going to reverse both halves, couldn't you just reverse the whole thing and then return the two slices?

```
  const midPoint = Math.floor(string.length / 2);
  const reversed = string.split("").reverse().join("");
  return reversed.substring(midPoint) + reversed.slice(0, midPoint);
```

paulcpederson (approved):

paulcpederson:
This should be fine as it's still deterministic. Little code golf comment, but looks great 👍🏻

packages/calcite-components/src/components/avatar/utils.ts
jcfranco: Great catch! Will fine-tune. ✨🔧✨

Issue #9278, opened 2024-05-07
Tile directly reassigns CSS variables within the component


alisonailea:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Tile component tokens as CSS can not be reassigned by users because the tokens are directly assigned within the component.

### Expected Behavior

As a user I should be able to give each component token new values and have those take precedence over the internal token assignments, including for classes and states

### Reproduction Sample

https://codepen.io/alisonailea/pen/yLWBXbq/c1ce40abeafb2210def6108a2b745c5f

### Reproduction Steps

1. open the link
2. observe the tile that has the class ".selected" 
3. notice the border is still blue and not red

### Reproduction Version

2.8.1-next.3

### Priority impact

p1 - need for current milestone

### Calcite package

- [X] @esri/calcite-components

### Esri team

Calcite (design)

geospatialem:
@alisonailea Would the above also be closed in favor of https://github.com/Esri/calcite-design-system/issues/7180 per https://github.com/Esri/calcite-design-system/pull/9279#issuecomment-2105110408?

geospatialem:
Closing in favor of #7180 as discussed in the team mid-milestone discussion and mentioned in the [comment above](https://github.com/Esri/calcite-design-system/issues/9278#issuecomment-2107959032) on 5/13. cc @alisonailea

PR #9279, opened 2024-05-07
refactor(tile): use internal variables for token value assignment

alisonailea:
**Related Issue:** #9278

## Summary

Resolves an issue where component tokens were being set internally and therefor were not reassignable by end-users.
The bug is only noticeable if someone in Online tries to adjust the component tokens for tile which is not yet documented. Since component tokens aren't technically released yet it could be argued this is a refactor
it does seem low risk. The only changes I made were some edits to the CSS organization and then adding internal component tokens to allow for internal value reassignment. I only noticed because I tried to get the epic #7180 in alignment with main.

### DIFF
--- a/packages/calcite-components/src/components/tile/tile.scss
+++ b/packages/calcite-components/src/components/tile/tile.scss
@@ -205,55 +205,55 @@ calcite-icon {
   }
 }
 
-:host([selection-appearance="border"]) {
-  &:host([layout="horizontal"]),
-  &:host([layout="vertical"]) {
-    .container.selected:focus::before {
-      --calcite-internal-tile-shadow: inset 0px 0px 0px 1px var(--calcite-color-brand);
+:host([selection-appearance="border"][layout="horizontal"]),
+:host([selection-appearance="border"][layout="vertical"]) {
+  .container.selected:focus::before {
+    --calcite-internal-tile-shadow: inset 0px 0px 0px 1px var(--calcite-color-brand);
 
-      block-size: 100%;
-      content: "";
-      display: block;
-      inline-size: 100%;
-      inset-block-start: 0;
-      inset-inline-start: 0;
-      position: absolute;
-    }
+    block-size: 100%;
+    box-shadow: inset 0px 0px 0px 1px var(--calcite-color-brand);
+    content: "";
+    display: block;
+    inline-size: 100%;
+    inset-block-start: 0;
+    inset-inline-start: 0;
+    position: absolute;
   }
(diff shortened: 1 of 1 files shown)


jcfranco:
Should this be targeting the `epic/7180-component-tokens` branch?

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -7,42 +7,56 @@
...
+
+  background-color: var(--calcite-tile-background-color, var(--calcite-color-foreground-1));
   block-size: var(--calcite-container-size-content-fluid);
   box-sizing: border-box;
   inline-size: var(--calcite-container-size-content-fluid);
-  outline: var(--calcite-border-width-sm, 1px) solid var(--calcite-tile-border-color);
+  outline: var(--calcite-border-width-sm, 1px) solid var(--calcite-internal-tile-outline-color);
   user-select: none;
-  &.interactive {
+  box-shadow: var(--calcite-tile-shadow, var(--calcite-internal-tile-shadow, none)) &.interactive {
```
eriklharper, line 42: Syntax error here.  Also, `--calcite-tile-shadow` is a vague name and also not documented above.  I'm wondering if we even need to expose a css variable for this or not since I imagine a user will want to customize the `brand` color and have it apply globally to all components?  And, if they need to override the `brand` color specific to Tiles, they should be able to do `calcite-tile { --calcite-color-brand: #custom; }`, right?

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -94,11 +108,22 @@
...
 }
 
+calcite-icon {
+  --calcite-icon-color: var(
+    --calcite-tile-icon-color,
+    var(--calcite-internal-tile-icon-color, var(--calcite-color-text-3))
+  );
+
+  &.selection-icon {
+    --calcite-icon-color: var(--calcite-internal-tile-select-icon-color);
```
eriklharper, line 123: You'll need to set the `color` here for the variable to take effect, e.g. `color: var(--calcite-icon-color);`.  I'm also wondering if `--calcite-icon-color` should be renamed to map to the internal counterpart, e.g. `--calcite-tile-icon-color`?

packages/calcite-components/src/components/tile/tile.scss
eriklharper: Just tested this css and it works great:
```
.screenshot-test calcite-tile {
  --calcite-color-brand: green;
}
```

image

Issue #8142, opened 2023-11-07
Split Button - consistent divider across appearances 


ashetland:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

The vertical divider in Split Button should visually match across appearance values.

### Acceptance Criteria

See [Figma file](https://www.figma.com/file/hDaUTLscr9X7xeNygsAQIu/Split-Button---consistent-divider-across-appearances?type=design&node-id=501-3613&mode=dev). `outline` and `outline-fill` are updated to match `solid` and `transparent`.
<img width="906" alt="CleanShot 2023-11-07 at 14 02 58@2x" src="https://github.com/Esri/calcite-design-system/assets/108549080/bd1109fc-f311-467b-a409-d68b9bd4f9ba">

### Relevant Info

cc @jcfranco 

### Which Component

Split Button

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

jcfranco:
Added `visual changes` label as this will tweak the `outline-fill` divider.

geospatialem:
In verifying in `2.9.0-next.18` it looks like if a background color is added, the same color is placed in the empty space for the revised divider: https://codepen.io/geospatialem/pen/VwOjjOX

![image](https://github.com/Esri/calcite-design-system/assets/5023024/ec3d231b-9754-4a89-85bb-52c29fbfb733)

@josercarcamo reassigning back to ya and set to in development for further investigation. 🕵🏻

geospatialem:
Verified in `2.9.0-next.21` with https://codepen.io/geospatialem/pen/VwOjjOX 🎉 

![image](https://github.com/Esri/calcite-design-system/assets/5023024/9ec9d997-67f6-44a5-8f21-45f15371bb88)

PR #9298, opened 2024-05-09
feat(split-button): Make dividers consistent

josercarcamo:
**Related Issue:** #8142 

## Summary
Made the dividers look consistent for all split button appearances.


packages/calcite-components/src/components/split-button/split-button.scss
```diff
@@ -98,6 +98,7 @@
 :host([appearance="outline"]) {
   .divider-container {
     border-block: 1px solid var(--calcite-split-button-divider);
+    background-color: initial;
```
jcfranco, line 101: The divider container's [background-color is set via custom CSS prop](https://github.com/Esri/calcite-design-system/blob/josercarcamo/8142-split-button-consistent-dividers/packages/calcite-components/src/components/split-button/split-button.scss#L89), so could you try updating [`--calcite-split-button-background`](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/split-button/split-button.scss#L55-L61) instead of overriding here? It looks like it should be `transparent` for both `outline` and `outline-fill` appearance types.

jcfranco (changes requested):

packages/calcite-components/src/components/split-button/split-button.scss
```diff
@@ -64,16 +64,19 @@
 :host([appearance="outline-fill"]) {
   &:host([kind="brand"]) {
     --calcite-split-button-divider: theme("colors.brand");
+    --calcite-split-button-background: transparent;
```
jcfranco, line 67: Since the same value is applied regardless of `kind`, can you look at moving this declaration one level up?

```scss
:host([appearance="outline"]),
:host([appearance="outline-fill"]) {
  --calcite-split-button-background: transparent;

  &:host[kind="brand"]) {
    --calcite-split-button-divider: theme("colors.brand");
  }
  /* ... */
```

Also, this block overrides [these styles](https://github.com/Esri/calcite-design-system/pull/9298/files/b1f6fd6dd9ecf815c55fae7dc5d58c7dbbba1cbd..bbfc07ca4b2c06250166fa07ca18c2203b056409#diff-65b8f4931c47584fef5447263565a807eaca0e737c51acd4e7137a6b90079c6bR55-R61), so we can probably remove the overridden ones.

packages/calcite-components/src/components/split-button/split-button.scss
```diff
@@ -126,13 +128,15 @@
...
 
 @include disabled() {
-  .divider-container {
-    @apply opacity-disabled;
-  }
-
   calcite-dropdown > calcite-button {
     @apply pointer-events-none;
   }
+  &:host([appearance="outline-fill"]),
```
jcfranco, line 134: I think this selector is not needed anymore since your changes above already set the background color to `transparent`.

packages/calcite-components/src/components/split-button/split-button.scss
josercarcamo: Because of specificity, I can't factor out this line.

packages/calcite-components/src/components/split-button/split-button.scss
josercarcamo: This is needed because specificity for disabled buttons or else a the divider appears like a light line.

packages/calcite-components/src/components/split-button/split-button.scss
jcfranco: Oh, I see. Disregard my comment then. Same with the disabled styling one below.

Sidebar: I think we could refactor this to simplify styles.

jcfranco (approved):
🚀

Issue #9253, opened 2024-05-03
Deprecate [Tile Select& Tile Select Group]


SkyeSeitz:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Functionality from Tile Select and Tile Select Group has now been implemented into Tile & Tile Group as of 2.8.0 allowing us to deprecate and direct users to to the components & functionality. 

### Relevant Info

A part of the Tile Design epic #6662 
cc @eriklharper @ashetland @macandcheese 

### Which Component

Tile Select & Tile Select Group

### Priority impact

p4 - not time sensitive

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

DitwanP:
Verified on `main`
![image](https://github.com/Esri/calcite-design-system/assets/60022782/48fc0fdf-ba6e-4a46-92a7-7b7cb2a4900c)

PR #9285, opened 2024-05-08
docs(tip, tip-group, tip-manager): deprecate components

geospatialem:
**Related Issue:** #9228

## Summary
Deprecates the `tip`, `tip-group` and `tip-manager` components with the following:
- Tip and Tip Manager: `carousel` and `carousel-item`
- Tip: `card`, `notice`, `panel`, or `tile`

PR #9302, opened 2024-05-09
docs(tile, tile-group): deprecate components

DitwanP:
**Related Issue:** #9253 

## Summary
Adds a JSDoc `@deprecated` tag to the following components:
- `tile-select`
- `tile-select-group`


geospatialem (commented):
Awesome! We'll need to tag team on the naming convention, as https://github.com/Esri/calcite-design-system/pull/9285 uses `calcite-` and this is PR currently scoped to the component name.

WDYT is the best approach?

DitwanP:
> Awesome! We'll need to tag team on the naming convention, as #9285 uses `calcite-` and this is PR currently scoped to the component name.
> 
> WDYT is the best approach?

I don't think it makes a huge difference but if we are to make a decision going forward I prefer the "calcite-..." option.

packages/calcite-components/src/components/tile-select-group/tile-select-group.tsx
```diff
@@ -9,6 +9,7 @@ import {
 import { TileSelectGroupLayout } from "./interfaces";
 
 /**
+ * @deprecated  Use the `calcite-tile-group` component instead.
```
geospatialem, line 12: nitpick:

It looks like there's two spaces between the `@deprecated` and `"Use"`.

packages/calcite-components/src/components/tile-select/tile-select.tsx
```diff
@@ -30,6 +30,7 @@ import { TileSelectType } from "./interfaces";
 import { CSS } from "./resources";
 
 /**
+ * @deprecated  Use the `calcite-tile` component instead.
```
geospatialem, line 33: nitpick:

It looks like there's two spaces between the `@deprecated` and `"Use"`.

geospatialem (approved):
**_Awesome!_** 🌟📚🌟

packages/calcite-components/src/components/tile-select/tile-select.tsx
DitwanP: 🦅👁️ Good catch!

geospatialem (approved):

PR #9306, opened 2024-05-10
fix(input, input-number, input-text): restore `autofocus`, `enter-key-mode` and `input-mode` attributes (#9245)

jcfranco:
**Related Issue:** #9235

## Summary

This fixes the following issues introduced by #9123 when component props
that shadowed global attributes were removed:

* the kebab-cased attribute names of those props would no longer work
(namely, `enter-key-mode`, `input-mode`)
* `autofocus` was missing from component types as Stencil does not
include it in the base `HTMLAttribute` type (see
https://github.com/ionic-team/stencil/issues/5726).


jcfranco:
Merging into the `hotfix` branch as this was already approved in https://github.com/Esri/calcite-design-system/pull/9245.

PR #9316, opened 2024-05-10
refactor: use `slotchange` event instead of `MutationObserver` when possible

jcfranco:
Replaces `MutationObserver` + `childList: true` usage with `slotchange` event for simplification and performance reasons ([slots already leverage mutation observer](https://dom.spec.whatwg.org/#mutation-observers)). Another benefit is that using `slotchange` eliminates conditions where callbacks may be invoked before the element has rendered.


jcfranco:
Closing. Will revisit if needed after https://github.com/Esri/calcite-design-system/issues/10310.

Issue #9307, opened 2024-05-10
`type-fest` build error after upgrading to Calcite 2.8.0


nwhittaker:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Encountering a build error after upgrading to Calcite 2.8.0 in a project and then referencing Calcite's `components.d.ts` file in the project's `tsconfig.json`:

```
  "files": [
    "node_modules/@esri/calcite-components/dist/types/components.d.ts",
  ]
```

### Expected Behavior

No build error.

### Reproduction Sample

https://github.com/nwhittaker/stencil-component

### Reproduction Steps

1. Check out the GitHub repo at the `calcite-2.8.0-type-fest-build-error` tag
2. Run `npm i`
3. Run `npm run build` and should see error here

### Reproduction Version

2.8.0

### Relevant Info

```
[ ERROR ]  TypeScript: node_modules/@esri/calcite-components/dist/types/utils/form.d.ts:2:10
           Module '"type-fest"' has no exported member 'Writable'.

      L1:  import { FunctionalComponent } from "../stencil-public-runtime";
      L2:  import { Writable } from "type-fest";
      L3:  import { Status } from "../components";

[21:11.6]  build failed in 1.58 s
```

### Regression?

2.7.1

### Priority impact

p1 - need for current milestone

### Impact

Workarounds appear to be to either enable `skipLibCheck` or add the `type-fest` dependency explicitly.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Field Apps

DitwanP:
🍭 Verified locally

PR #9317, opened 2024-05-11
build(deps): add type-fest to prevent build errors

benelan:
**Related Issue:** #9307

## Summary

Add `type-fest` to the dependencies of `@esri/calcite-components` to prevent build errors. `type-fest` can also be used to simplify/remove the [preact type generator](https://github.com/Esri/calcite-design-system/blob/27a82dcaf8b091c609b3bf765d8c8617c447d8e7/packages/calcite-components/support/preact.ts#L8) in the future.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -101,7 +101,6 @@
         "ts-jest": "29.1.2",
         "tsx": "4.7.2",
         "turbo": "1.13.2",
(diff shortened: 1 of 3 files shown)


jcfranco (approved):

jcfranco:
Should this use the `fix` commit type since it’s addressing a reported bug?

Issue #9232, opened 2024-05-01
[Carousel] Animation on Carousel Item change doesn't always occur


macandcheese:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The animation as Carousel Items cycle is not consistently displayed. Only when "jumping forward / backwards" are the correct animations being set via css class.

https://github.com/Esri/calcite-design-system/assets/4733155/8c683b01-687f-4043-bb0c-9a698bc6a581

### Expected Behavior

The slides should animate in from the correct direction on each Carousel Item change.

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/rNbgGQE?editors=1000

### Reproduction Steps

1. Open demo
2. Interact with component
3. Observe inconsistent animation behavior

### Reproduction Version

2.8.0

### Relevant Info

I think we need to clear the css class or animation at all times, and handle a case where the internal "direction state" isn't being set correctly, in `nextItem()`, `previousItem()`, `setSelectedItem()` internal functions.

### Priority impact

p2 - want for current milestone

### Impact

Should be added to an upcoming patch release if possible.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

geospatialem:
Verified in `2.8.1-hotfix.0`:

![verify-animations-carousel](https://github.com/Esri/calcite-design-system/assets/5023024/c6e80c60-c1bc-4215-b023-e6ecdea6b72d)

PR #9325, opened 2024-05-14
fix(carousel): animate items with the same direction

jcfranco:
**Related Issue:** #9232

## Summary

Updates carousel to wait for item animations before each slide.

**Note**: I'll submit a follow-up PR to refactor [`openCloseComponent.onToggleOpenCloseComponent`](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/utils/openCloseComponent.ts#L91) to use the same DOM util as carousel.


benelan (approved):
LGTM!

jcfranco:
E2E test added!

Issue #7095, opened 2023-06-02
`dropdown`, `dropdown-item`, `dropdown-group` update icon spacing


ashetland:
### Description

Update icon spacing to make more consistent with the rest of Calcite. This was outlined as part of the work to audit and update nested indentations to improve visual hierarchy across components (Epic #6632).

### Acceptance Criteria

![image](https://github.com/Esri/calcite-components/assets/108549080/25cab87e-cf3c-4ac5-ba7e-bbe1e294f442)

### Relevant Info

[Figma file also available](https://www.figma.com/file/ua4NiNaOLrs5nGo5bECBfZ/Dropdown-Icon-Spacing-%5Bissue-7090%5D?type=design&node-id=287-113361) for additional details and a comparison to the current build.

### Which Component

Dropdown Item

### Priority impact

p4 - not time sensitive

### Esri team

Calcite (design)

geospatialem:
Verified spacing across scales in `2.9.0-next.18` with https://codepen.io/geospatialem/pen/bGyeZwM, where:
- `s` is `8px`
- `m` is `12px`
- `l` is `16px`

![image](https://github.com/Esri/calcite-design-system/assets/5023024/22abc734-016d-4181-a47b-0062a94d0622)

PR #9330, opened 2024-05-14
feat(dropdown-item): update spacing of icons

anveshmekala:
**Related Issue:** #7095 

## Summary

Updates spacing of icons in `calcite-dropdown-item`.


packages/calcite-components/src/components/dropdown-item/resources.ts
```diff
@@ -1,9 +1,9 @@
 export const CSS = {
   container: "container",
+  containerNone: "container--none-selection",
```
anveshmekala, line 3: SideNote: Sorting is handled in a followup refactor PR.

anveshmekala: #9349

ashetland:
Lookin' gewd!

packages/calcite-components/src/components/dropdown-item/dropdown-item.tsx
```diff
@@ -125,7 +125,7 @@ export class DropdownItem implements InteractiveComponent, LoadableComponent {
    *
    * @internal
    */
-  @Prop() scale: Scale = "m";
+  @Prop({ reflect: true }) scale: Scale = "m";
```
jcfranco, line 128: Per our conversation, let's discuss reflection of internal properties as a pattern, as it simplifies styling without significant trade-offs.

jcfranco (approved):
✨🏆✨

packages/calcite-components/src/components/dropdown-item/dropdown-item.scss
```diff
@@ -1,38 +1,35 @@
-.container--s {
-  @apply text-n2h py-1;
-  padding-inline-end: theme("padding.2");
-  padding-inline-start: theme("padding.6");
+:host([scale="s"]) {
```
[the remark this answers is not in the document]
jcfranco: Resolving as this is outdated.

packages/calcite-components/src/components/dropdown-item/dropdown-item.scss
```diff
@@ -164,31 +159,28 @@
 // icon start & end
 .container--s {
   .dropdown-item-icon-start {
-    margin-inline-end: theme("margin.2");
-    margin-inline-start: theme("margin.1");
+    padding-inline-end: 8px;
   }
   .dropdown-item-icon-end {
-    margin-inline-start: theme("margin.2");
+    padding-inline-start: 8px;
```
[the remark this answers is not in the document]
jcfranco: Resolving as this is outdated.

packages/calcite-components/src/components/dropdown-item/dropdown-item.tsx
anveshmekala: Reflecting @internal props for styling purposes will help us  improves the maintainability and readability of styles. Though it's internal since they aren't document for public use it is safe to reflect and take advantage of the attribute selector in CSS instead of creating dynamic classes.

Considering this PR as example, reflecting prop helped us avoid cumbersome styling code and there are other instance in the repo where internal props are reflected. We can use this as a pattern if the reflection helps improving the code without affecting UI/UX.

driskull: Yeah its a trade off discussion...

Is showing an undocumented attribute terrible? No.
Is not showing an undocumented attribute better? Probably.

So its really about if its worth the work to use an internal shadow DOM class vs relying on an attribute.

jcfranco: > So its really about if its worth the work to use an internal shadow DOM class vs relying on an attribute.

Agreed. From what I've seen, the extra work of adding root container classes to mimic host and reflected internal attributes (e.g., `scale`-related styling) can feel like overkill. Since we usually avoid reflecting props on the host element, leveraging the reflection of internal properties seems simpler and cleaner in this case.

PR #9343, opened 2024-05-14
fix(input, input-number, input-text): ensure autofocus is available on HTMLElement

jcfranco:
**Related Issue:** #9235

## Summary

Exposes `autofocus` property to ensure the prop/attr is available on the `HTMLElement` type.

This should no longer be necessary once https://github.com/ionic-team/stencil/issues/5726 is fixed and available in Stencil.


benelan (approved):

jcfranco:
**Note**: using `@ignore` to keep the prop/attr in types and to allow filtering from the doc site.

Issue #7723, opened 2023-09-11
Calcite Switch has hover effect in disabled state


tyanthony:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When the `Switch` component is disabled, both the checked and unchecked options have hover states:

disabled, unchecked:
<img width="112" alt="image" src="https://github.com/Esri/calcite-design-system/assets/17069279/a4f3ff29-cd24-4cf5-b7f7-a83c1e3c14d0">

disabled, checked, left has hover, right does not:
<img width="125" alt="image" src="https://github.com/Esri/calcite-design-system/assets/17069279/3f452ebf-7604-4738-b16c-7b3e73d1792b">

### Expected Behavior

There should not be any hover effects when the `switch` is disabled.

### Reproduction Sample

https://codepen.io/tyleranthony/pen/QWzvENZ?editors=100

### Reproduction Steps

1. Add a Calcite Switch, disabled
2. Hover over the thumb slider
3. Notice the hover effect

### Reproduction Version

1.8.0

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Mission

brittneytewks:
Removed Figma changes label as this will be tracked in monday.com

DitwanP:
🍬 Verified on `2.9.0-next.18`

PR #9365, opened 2024-05-18
fix(switch): prevent hover state change when disabled

jcfranco:
**Related Issue:** #7723

## Summary

Prevents interaction states on switch when disabled.


driskull (approved):
Nice work 👍

Issue #8374, opened 2023-12-07
text-area `disabled` state low-contrast


nwhittaker:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The disabled state for a `text-area` element is very low contrast and inconsistent with the disabled state of the other input components.

![image](https://github.com/Esri/calcite-design-system/assets/108549080/be05952a-8425-4839-8844-ff325862d829)

### Expected Behavior

I was expecting the colors used in the text-area's disabled state to match the colors used in the inputs' disabled state.

![image](https://github.com/Esri/calcite-design-system/assets/108549080/6a3c4546-101b-42b1-81e1-c478d8fc8278)

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/ZEwwwMy

### Reproduction Steps

Visit the repro. and visually compare the disabled components.

### Reproduction Version

2.0.0

### Relevant Info

Inspecting the `calcite-text-area` component, it looks like a `0.5` opacity is getting doubly applied. Once to the `:host` and once to the `:host textarea` element.

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Field Apps

geospatialem:
Adding design folks to this issue to determine next steps and evaluation across the components in alignment with textarea.

ashetland:
@geospatialem Next steps are to remove the double opacity, matching the above expected behavior. If there are no other questions, we should be able to mark this `ready for dev`.

geospatialem:
> @geospatialem Next steps are to remove the double opacity, matching the above expected behavior. If there are no other questions, we should be able to mark this `ready for dev`.

@ashetland Awesome, think we'd just need a Figma url, or screenshot of the expected outcome then this is `ready for dev` 🚀 .

DitwanP:
🍬 Verified on `2.9.0-next.18`

PR #9366, opened 2024-05-18
fix(text-area): remove double transparency applied when disabled

jcfranco:
**Related Issue:** #8374

## Summary

Removes bespoke disabled styling applied in addition to the common disabled style mixin.


driskull (approved):
Is there some test that should verify this fix? Screenshot possibly?

Issue #8841, opened 2024-02-27
calcite-checkbox: causing horizontal scrollbars inside panel for RTL


AdelheidF:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

A calcite-checkbox close to the left border of a panel in RTL is causing horizontal scrollbars inside the panel. This is caused by the `::after`.

![image](https://github.com/Esri/calcite-design-system/assets/44481292/31fc3ef9-f374-4811-9f0e-b6f7b11a878c)

### Expected Behavior

no horizontal scrollbars, just like LTR

### Reproduction Sample

https://codepen.io/afreitag/pen/WNWeEKz

### Reproduction Steps

Open codepen and hover over checkbox. You'll see scrollbars appearing at the bottom of the panel.

### Reproduction Version

2.4.0

### Regression?

no

### Priority impact

p2 - want for current milestone

### Impact

we have a workaround setting overflow-x to hidden

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Map Viewer

DitwanP:
🍬 Verified on `2.9.0-next.18`

PR #9367, opened 2024-05-18
fix(checkbox): fix checkbox hit area placement in RTL

jcfranco:
**Related Issue:** #8841

## Summary

Updates styles to position hit area horizontally according to page direction.


driskull (approved):
Looks good 💯

Issue #9240, opened 2024-05-01
`TypeError` in `filter` when disconnecting during update


nwhittaker:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

If a `<calcite-filter>` element is in the middle of its update lifecycle when it's disconnected, it throws a `TypeError`.

### Expected Behavior

If a `<calcite-filter>` element is in the middle of its update lifecycle when it's disconnected, it does not throw a `TypeError`.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/LYvMQMm

### Reproduction Steps

1. Visit the code sample and open the dev tool's console
2. See the `TypeError` logged in the console

### Reproduction Version

2.7.1

### Relevant Info

Looking at the stack trace, it appears as though the filter's update render is setting `messageOverrides` on its `<calcite-input>` element which in turn attempts to call the input's `onMessagesChange` watch method which is `undefined` at that point due to the input having been disconnected.

```
TypeError: f[t] is not a function
    at index.js:2464:38
    at Array.map (<anonymous>)
    at yt (index.js:2461:34)
    at f.set [as messageOverrides] (index.js:2557:25)
    at _ (index.js:926:29)
    at K (index.js:999:9)
    at nt (index.js:1577:17)
    at Z (index.js:1356:13)
    at nt (index.js:1583:13)
    at Z (index.js:1356:13)
```

This issue may not be limited to the `filter` component. It may impact any component that, in their `render()` methods, set `messageOverrides` on another component.

### Priority impact

p3 - want for upcoming milestone

### Impact

Adds verbosity to console logs and can be a red herring when debugging other issues. Can also impact how layouts are designed to mitigate the presence of this error.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Field Apps

geospatialem:
Verified no console messages are populated in `2.9.0-next.22` with https://codepen.io/nwhittaker-esri/pen/LYvMQMm.

PR #9368, opened 2024-05-18
fix(t9n): fix error caused by disconnecting components during the setting of initial message overrides

jcfranco:
**Related Issue:** #9240

## Summary

This updates the `t9n` module to use a noop function (instead of undefined) for the injected watcher callback when disconnected. This change addresses an issue where a component invokes watchers during disconnection.


anveshmekala (approved):
🥇

driskull (approved):
👍

Issue #9321, opened 2024-05-13
[combobox] Uncaught type errors are thrown


mpayson:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Calcite combobox throws type errors in the console logs

### Expected Behavior

No uncaught type errors

### Reproduction Sample

https://github.com/mpayson/calcite-debug

### Reproduction Steps

1. Clone the reproduction sample above and checkout the branch `combobox-repo`
2. Run the app with developer console open
3. Click the "Toggle" button
4. See uncaught error in console

In the reproduction case, the type error is:
> TypeError: this.textInput is null
    documentClickHandler combobox.tsx:357
    hostListenerProxy index.js:2622
    ael index.js:2878
    addHostEventListeners index.js:2608
    addHostEventListeners index.js:2604
    registerHost index.js:2808
    __registerHost index.js:2358
    Combobox2 @esri_calcite-components_dist_components_calcite-combobox.js:3389
    React 13
    onClick main.jsx:18
    React 23
    <anonymous> main.jsx:29

In our app, this is the type error, it happens for comboboxes across our app
> TypeError: Cannot read properties of null (reading 'value')
    at Combobox2.documentClickHandler (combobox.tsx:357:51)
    at HTMLDocument.<anonymous> (index.js:2622:39) undefined

Additional notes:
* Rendering just the combobox as the app root does not reproduce the issue, it seems like it is related to comboboxes being rendered at some point within the app

### Reproduction Version

2.8.0

### Priority impact

p1 - need for current milestone

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [X] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Online

jcfranco:
Dug a bit and this is caused by [this line](https://github.com/Esri/calcite-design-system/blob/45e73bd19f10dbcc1fc18b7db4f891a3df485e37/packages/calcite-components/src/components/combobox/combobox.tsx#L357). It seems the window `click` listener is set up and triggered before the component is rendered.

In terms of workaround, you could toggle the component's `hidden` property vs recreating it:

```tsx
{(
        <CalciteCombobox hidden={!show}>
          <CalciteComboboxItem value="1" textLabel="One" />
[...]

PR #9373, opened 2024-05-20
fix(combobox): fix error that occurs when a click is emitted when the component is appended to the DOM

jcfranco:
**Related Issue:** #9321

## Summary

This fixes an issue where references to internal elements were undefined in the window-level click handler. This would happen because the window click handler was added when the component was connected, but before its internals were rendered.

### DIFF
--- a/packages/calcite-components/src/components/combobox/combobox.tsx
+++ b/packages/calcite-components/src/components/combobox/combobox.tsx
@@ -344,7 +344,7 @@ export class Combobox
 
   @Listen("click", { target: "document" })
   async documentClickHandler(event: PointerEvent): Promise<void> {
-    if (this.disabled || !this.open || event.composedPath().includes(this.el)) {
+    if (this.disabled || event.composedPath().includes(this.el)) {
       return;
     }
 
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -343,16 +343,12 @@ export class Combobox
   //--------------------------------------------------------------------------
 
   @Listen("click", { target: "document" })
```
driskull, line 345: @jcfranco do you think it would be worthwhile to refactor this event to only be added once for all combobox components? Currently, each combobox sets up an event listener on the document. Could be enhanced to create a single event listener for all combobox components.

driskull (approved):
👍

packages/calcite-components/src/components/combobox/combobox.tsx
jcfranco: Not sure. We'd still need to add a mechanism to 1. ensure a single window/doc listener is set up and 2. manage components to be processed by this single callback, no?

driskull: Correct. It may not be beneficial if there aren't many comboboxes used in a document.

Issue #8981, opened 2024-03-21
[Panel & Flow Item]: Refactor `"footer"` slot to account for `modal`


geospatialem:
### Description
Currently the `panel` and `flow-item` have a `"footer"` slot, however to accommodate the Dialog + Modal component effort in https://github.com/Esri/calcite-design-system/issues/7886, we may need to account of additional footer slots, such as `"footer-start"` and `"footer-end"` to match the "back / secondary / start" currently offered in the `modal` component.

### Acceptance Criteria

Refactor `panel` and `flow-item` footer to include `footer-end` & `footer-start` slots with no built in spacing between or within slots

### Which Component
- `panel`
- `flow-item`

### Relevant Info
A dependency needed for the Dialog + Modal component effort in https://github.com/Esri/calcite-design-system/issues/7886

See [Figma file](https://www.figma.com/file/wVqftT1aZDjwYVkDZ2V5fC/Dialog?type=design&node-id=621-5496&mode=design&t=9PfAlfhXFoN5YOXS-11) for specs
<img width="1620" alt="image" src="https://github.com/Esri/calcite-design-system/assets/29716057/61eca8fc-4e0a-463e-a032-e4094402d4f8">

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

macandcheese:
Not sure if this applies to Card cc @ashetland @SkyeSeitz - I think it may have just been mentioned as a component that has footer-start/end slots.

SkyeSeitz:
Agreed. I believe the idea was to confirm if we wanted more consistency between footers across components or if Card should be unique.

Elijbet:
Can we determine the best approach for the footer slots? In most use cases do we want the footer as is without the need for start/end slots or is it the other way around? I'm wondering if the footer needs to be deprecated.

@ashetland @SkyeSeitz @macandcheese

Elijbet:
From what I gather so far after speaking with @macandcheese and @geospatialem , `footer-start/end` will live alongside `footer`:
 
- `footer` - for non modal/dialog scenarios
- `footer-start` - to support modal/dialog components (future)
- `footer-end` - to support modal/dialog components (future)

The question there is what happens when the footer and one or both of the new footer-start/end are populated? Do the new slots take precedence? 

[...]

PR #9374, opened 2024-05-20
feat(panel, flow-item): add `footer-start` and `footer-end` slots

Elijbet:
**Related Issue:** #8981

## Summary

- Add `footer-start` and `footer-end` slots.
- Provide horizontal gap setting within `footer-start` & `footer-end` slots according to scale.
- Provide vertical gap setting within existing `footer` slot according to scale.


packages/calcite-components/src/components/flow-item/flow-item.tsx
```diff
@@ -50,7 +50,8 @@ import { CSS, ICONS, SLOTS } from "./resources";
  * @slot header-menu-actions - A slot for adding an overflow menu with `calcite-action`s inside a `calcite-dropdown`.
  * @slot fab - A slot for adding a `calcite-fab` (floating action button) to perform an action.
  * @slot footer-actions - [Deprecated] Use the `"footer"` slot instead. A slot for adding `calcite-button`s to the component's footer.
- * @slot footer - A slot for adding custom content to the component's footer.
```
driskull, line 53: Why was the footer slot removed? This would be a breaking change.

driskull (changes requested):

packages/calcite-components/src/components/flow-item/flow-item.tsx
Elijbet: Sorry, I'll fix that.

jcfranco: Can you also restore the doc for the `footer` slot?

packages/calcite-components/src/components/flow-item/flow-item.stories.ts
```diff
@@ -279,13 +279,13 @@ export const withActionBarAndContentTop_TestOnly = (): string =>
     </calcite-flow-item>
   </div>`;
 
-export const footerPaddingAndContentBottom_TestOnly = (): string =>
+export const footerStartEndAndContentBottom_TestOnly = (): string =>
```
jcfranco, line 282: Can you merge the content from both `footerStartEndAndContentBottom_TestOnly` and `footerPaddingAndContentBottom_TestOnly` stories? The latest adds coverage for the start/end footer slots, but loses the footer one. 

I think we also need one that demonstrates the footer slot priority (`footer` > `footer-start`/`footer-end`).

☝️ applies to panel stories too.

packages/calcite-components/src/components/flow-item/flow-item.tsx
```diff
@@ -389,8 +391,10 @@ export class FlowItem
             <slot name={SLOTS.fab} slot={PANEL_SLOTS.fab} />
             <slot name={SLOTS.contentTop} slot={PANEL_SLOTS.contentTop} />
             <slot name={SLOTS.contentBottom} slot={PANEL_SLOTS.contentBottom} />
-            <slot name={SLOTS.footerActions} slot={PANEL_SLOTS.footerActions} />
             <slot name={SLOTS.footer} slot={PANEL_SLOTS.footer} />
+            <slot name={SLOTS.footerActions} slot={PANEL_SLOTS.footerActions} />
+            <slot name={SLOTS.footerEnd} slot={PANEL_SLOTS.footerEnd} />
```
jcfranco, line 396: I originally thought all footer slots would render regardless of slotted content, but this will slot into `panel`, which has the mutually exclusive `footer` and `footer-start`/`footer-end` slot rendering, right? I like this!

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -19,17 +19,17 @@
 @import "../../assets/styles/header";
 
 :host([scale="s"]) .content-top,
```
jcfranco, line 21: Let's use nesting to reduce duplication.

packages/calcite-components/src/components/panel/panel.tsx
```diff
@@ -560,13 +574,25 @@ export class Panel
...
+    const { hasFooterEndContent, hasFooterStartContent, hasFooterContent, hasFooterActions } = this;
 
-    const showFooter = hasFooterContent || hasFooterActions;
+    const showFooter =
+      hasFooterStartContent || hasFooterEndContent || hasFooterContent || hasFooterActions;
 
     return (
       <footer class={CSS.footer} hidden={!showFooter}>
-        <slot key="footer-slot" name={SLOTS.footer} onSlotchange={this.handleFooterSlotChange} />
+        <slot key="footer-slot" name={SLOTS.footer} onSlotchange={this.handleFooterSlotChange}>
```
jcfranco, line 584: There is no conditional rendering within the `footer` element, so `key` is not needed on any child (see https://stenciljs.com/docs/templating-jsx#conditionals).

jcfranco (approved):
Once minor comments are addressed, this is good to merge!

👣👣👣👣👣👣👣👣👣👣👣👣👣👣👣👣👣👣👣👣👣👣👣👣
👣🦶👣👣👣🦶🦶🦶🦶👣🦶🦶🦶🦶🦶👣🦶👣👣👣🦶👣🦶👣
👣🦶👣👣👣🦶👣👣👣👣👣👣🦶👣👣👣🦶🦶👣🦶🦶👣🦶👣
👣🦶👣👣👣🦶👣🦶🦶👣👣👣🦶👣👣👣🦶👣🦶👣🦶👣🦶👣
👣🦶👣👣👣🦶👣👣🦶👣👣👣🦶👣👣👣🦶👣👣👣🦶👣👣👣
👣🦶🦶🦶👣🦶🦶🦶🦶👣👣👣🦶👣👣👣🦶👣👣👣🦶👣🦶👣
👣👣👣👣👣👣👣👣👣👣👣👣👣👣👣👣👣👣👣👣👣👣👣👣

packages/calcite-components/src/components/flow-item/flow-item.tsx
Elijbet: Brilliant! ⭒°｡𖥔 ݁ ˖𖦹⭒°｡⋆⭒

Elijbet: The `flow-item` was failing to render `footer-start` and `footer-end`, I figured it was because slotting an empty slot in this structure gives the empty footer precedence. I had to do a little refactoring to make it work. 
Screenshot 2024-06-24 at 2 09 16 PM

https://github.com/Esri/calcite-design-system/pull/9374/files#diff-36d423bf7b7d6ed67ae5d1bd668b38396b501071d0cda69791e514f68bfd5b94R395-R398

packages/calcite-components/src/components/flow-item/flow-item.stories.ts
```diff
@@ -65,15 +65,17 @@ const contentHTML = html`
 `;
 
 const footerHTML = html`
```
jcfranco, line 67: Not critical for this PR, but let's try to rename this to better match its usage. This is named as footer, but it has HTML with start/end footer slot content. Applies to panel stories.

jcfranco (commented):
🚀

jcfranco (approved):
🚀

driskull (approved):
👍

PR #9392, opened 2024-05-21
chore: use local placeholder image in html files

driskull:
**Related Issue:** None

## Summary

- Use local placeholder image for images
- Remove duplicate links for `https://webapps-cdn.esri.com/CDN/fonts/v1.4.1/fonts.css`
- Update Maps SDK url to use the `next` version so it is always up to date.
- Update links to external URLs to use `http://www.esri.com`
- Update input value's that are links to be `http://www.esri.com`


jcfranco (approved):
LGTM! 

We should look at having the linter catch this. `eslint-html`'s [`no-restricted-attr-values`](https://html-eslint.org/docs/rules/no-restricted-attr-values) looks promising.

driskull:
Yeah that would be great. Should we open an issue for it?

jcfranco:
@driskull An issue would be great, thanks!

driskull:
created issue https://github.com/Esri/calcite-design-system/issues/9396

Issue #9361, opened 2024-05-17
Update tests using URL placeholder images to use the placeholderImage utility function


driskull:
### Priority impact

p - low

### Test type

Update test

### Which Component(s)

Carousel, custom-theme-stories.

### Unstable Tests

Anything with placebear or palcem.at

### Test error, if applicable

```
<style>
      .bg-image-example {
        color: red;
        background-image: url("https://placebear.com/3000/2000");
        background-size: cover;
        padding: 1rem;
        height: 300px;
        font-size: 32px;
        font-weight: 600;
        line-height: 32px;
      }
    </style>
```

### PR skipped, if applicable

#

geospatialem:
Verified on the `main` branch

PR #9395, opened 2024-05-22
chore: Update placeholder images in stories and demos

macandcheese:
**Related Issue:** #9361 , #9392 

## Summary
Removes served images from stories and local demo. We keep them in `.md` file, as these need to be hosted somewhere in order to work if the samples are used in an external location. cc @driskull 

Ideally, we could use map tiles or something (https://codepen.io/mac_and_cheese/pen/XWwXXqx?editors=1000) for these, but not sure how reliable those are.

### DIFF
--- a/packages/calcite-components/src/demos/carousel.html
+++ b/packages/calcite-components/src/demos/carousel.html
@@ -54,13 +54,13 @@
       }
 
       .bg-image-example {
-        background-image: url("https://placebear.com/3000/2000");
+        background-image: url("./_assets/images/placeholder.svg");
         background-size: cover;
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/demos/carousel.html
```diff
@@ -54,13 +54,13 @@
       }
 
       .bg-image-example {
-        background-image: url("https://placebear.com/3000/2000");
```
driskull, line 57: nice catch. Ill see if there are any more uses like this in CSS.

driskull (approved):
👍

PR #9402, opened 2024-05-22
feat(split-button): Make dividers consistent

josercarcamo:
**Related Issue:** #8142

## Summary
This is to address a case missed in https://github.com/Esri/calcite-design-system/pull/9298: Fixed background color anomoly when the background of :root is set.


alisonailea (approved):
To be consistent with the incoming tokens work, can we rename this variable from `--calcite-split-button-background` to `--calcite-internal-split-button-background`?
Same with `--calcite-split-button-divider` --> `--calcite-internal-split-button-divider`
It will help lessen merge conflics with the tokens epic.

Issue #7899, opened 2023-09-28
Tree single-persist selection mode bugs


ashetland:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

`selection-mode="single-persist"` has multiple bugs:

- Allows for multiple selections
- Missing single-select indicator
- Likely related to 👆: nested indentations are incorrect

![CleanShot 2023-09-28 at 12 54 50](https://github.com/Esri/calcite-design-system/assets/108549080/fbfc4129-cd08-4ad1-9690-c2f7f5965b5e)

### Expected Behavior

- [x] Allows for single selection only
- [x] Has single select indicator
- [x] Nested indents match other modes

### Reproduction Sample

https://codepen.io/ashetland/pen/OJrEOzE

### Reproduction Steps

1. Open the codepen
2. Select multiple items in single-persist Tree

### Reproduction Version

v1.8.0

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

geospatialem:
Setting the priority as low, unless other users or teams are experiencing similar issues and Calcite team members will "up" the priority.

geospatialem:
Verified in `2.9.0-next.22`

PR #9405, opened 2024-05-23
fix(tree): allow single select only and add indicator

josercarcamo:
**Related Issue:** #7899

## Summary
Allowed only single selection and added the selection indicator

### DIFF
--- a/packages/calcite-components/src/components/tree-item/tree-item.tsx
+++ b/packages/calcite-components/src/components/tree-item/tree-item.tsx
@@ -201,7 +201,10 @@ export class TreeItem implements ConditionalSlotComponent, InteractiveComponent 
 
   render(): VNode {
     const rtl = getElementDir(this.el) === "rtl";
-    const showBulletPoint = this.selectionMode === "single" || this.selectionMode === "children";
+    const showBulletPoint =
+      this.selectionMode === "single" ||
+      this.selectionMode === "children" ||
+      this.selectionMode === "single-persist";
     const showCheckmark =
       this.selectionMode === "multiple" || this.selectionMode === "multichildren";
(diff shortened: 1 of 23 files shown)


packages/calcite-components/src/components/tree-item/tree-item.tsx
```diff
@@ -201,7 +201,10 @@ export class TreeItem implements ConditionalSlotComponent, InteractiveComponent
 
   render(): VNode {
     const rtl = getElementDir(this.el) === "rtl";
-    const showBulletPoint = this.selectionMode === "single" || this.selectionMode === "children";
```
driskull, line 204: Do existing screenshot tests show changes? If not, can we add a screenshot test for he bullet point?

packages/calcite-components/src/components/tree/tree.tsx
```diff
@@ -175,7 +175,8 @@ export class Tree {
       (((this.selectionMode === "single" || this.selectionMode === "multiple") &&
         childItems.length <= 0) ||
         this.selectionMode === "children" ||
-        this.selectionMode === "multichildren");
+        this.selectionMode === "multichildren" ||
+        this.selectionMode === "single-persist");
```
driskull, line 179: Should we add an e2e test to account for single-persist allowing to be unselected?

driskull (commented):
Hey @josercarcamo code changes look 💣 👍 

Just had some comments

packages/calcite-components/src/components/tree/tree.tsx
josercarcamo: Done.

driskull (approved):
👍 💯

Issue #9422, opened 2024-05-24
input-date-picker: selecting day activates previous day in certain cases


eriklharper:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When I pick a date from earlier times ( e.g. ~1850) from the DatePicker, `event.target.value` returns a day before the selected date. `event.target.valueAsDate` returns the correct value.

https://github.com/Esri/calcite-design-system/assets/821864/b4fa287c-64ea-4aae-86be-c7218e09e0be

### Expected Behavior

`event.target.value` returns the equivalent year, month and day as `valueAsDate`.

### Reproduction Sample

https://codepen.io/nsulzberger_esri/pen/gOJMjyd (input-date-picker)
https://codepen.io/nsulzberger_esri/pen/rNgLKMK (date-picker)

### Reproduction Steps

1. Change your system timezone to `Europe/Zurich`
2. Open the linked codepen
3. Select March 7th, 1850 from the Date Picker popup
4. Result: March 6th, 1850 is selected instead of March 7th.

### Reproduction Version

2.8

### Relevant Info

The current implementation uses the `dateToISO` util function which relies on `Date.prototype.toISOString` method on the Date object stored in `valueAsDate`.  `toISOString` [returns the date in UTC](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString), not in the local time which is what will be reflected in `valueAsDate`, which can lead to discrepancies like this where the two values won't match.  Since `valueAsDate` is a Date object and will thus always format in local time, we should avoid using `toISOString` for computing `value`.

We won't be able to add spec test coverage for changing the `dateToISO` because we can't emulate timezones in the spec environment.  We should be able to test in specific timezones in the e2e environment.

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

nsulzberger:
@eriklharper I am not sure, but I think the ticket description is not quite correct, the title is right.
date-picker has the behavior you are describing. `value` contains the day before, but `valueAsDate` is fine. So users can work around that.

[...]

PR #9424, opened 2024-05-24
fix(input-date-picker, date-picker): ensure day selection doesn't activate the previous day in certain scenarios

eriklharper:
**Related Issue:** #9422 

## Summary

This fixes an issue where in some cases selecting a date in the date-picker popup of `input-date-picker` would select the previous day instead of the one selected.


alisonailea (approved):

anveshmekala (approved):
🏆

packages/calcite-components/src/utils/date.ts
```diff
@@ -151,7 +151,10 @@ export function datePartsFromLocalizedString(
  */
 export function dateToISO(date?: Date): string {
   if (date instanceof Date) {
-    return new Date(date.getTime() - date.getTimezoneOffset() * 60000).toISOString().split("T")[0];
+    const month = String(date.getMonth() + 1).padStart(2, "0");
```
jcfranco, line 154: Could this be simplified to

```ts
return date instanceof Date ? date.toISOString().split('T')[0] : "";
```

?

packages/calcite-components/src/utils/date.ts
```diff
@@ -151,7 +151,10 @@ export function datePartsFromLocalizedString(
  */
 export function dateToISO(date?: Date): string {
   if (date instanceof Date) {
-    return new Date(date.getTime() - date.getTimezoneOffset() * 60000).toISOString().split("T")[0];
```
jcfranco, line 154: Can you also update the spec test to add coverage for the failing scenario?

jcfranco (changes requested):
Looking good!

packages/calcite-components/src/utils/date.ts
eriklharper: the `toISOString` is the problematic function, so that wouldn't fix the issue.

eriklharper: image

packages/calcite-components/src/utils/date.ts
jcfranco: No longer applicable based on: https://github.com/Esri/calcite-design-system/pull/9424/files#r1622784980.

packages/calcite-components/src/utils/date.ts
eriklharper: 1. Yes, the docs should be more clear on what it returns exactly.  I actually think the name `dateToISO` is appropriate in this case because when you look up the [ISO 8601 standard](https://en.wikipedia.org/wiki/ISO_8601), it does call out a valid way to express just the date portion here: 
image
2. Will do.  I'm also wondering if a comment in the code of the function could be worth it as well to remember the decision why we moved away from `toISOString`?
3. Yes.
4. Could be worth it, but because the `Date` object's instance methods always return values in the local timezone (`getDate`, `getMonth`, `getFullYear`) it might be overkill to introduce a plugin for something its already handling.

Generally, the issue I've found that I can't seem to workaround is exactly how to get the exact `date` value back out of the constructed `Date` object that was passed in.  Because all the `get*` methods return in local time, if the `Date` was constructed with a CEST timezone offset specified in the input string like:

```
const cestDate = new Date("1850-03-13T00:00:00.000+02:00");
```
There's no way to retrieve the value `13` from that original string back out of the `Date` object in a reliable way that is independent of the current system timezone setting.

packages/calcite-components/src/components/date-picker/date-picker.e2e.ts
```diff
@@ -565,6 +565,114 @@ describe("calcite-date-picker", () => {
     await page.waitForChanges();
     expect(await datePicker.getProperty("value")).toEqual(["2020-09-15", "2020-09-30"]);
   });
+
+  it("sets value to the selected day in the 2000s", async () => {
```
jcfranco, line 569: Can you use `describe` blocks to group these tests? Maybe something like this:

```ts
describe("cross-century date values", () => {
  describe("PST", () => { /* ... */})
  describe("CEST", () => { /* ... */})
});
```
?

Applies to `input-date-picker` tests.

packages/calcite-components/src/components/date-picker/date-picker.e2e.ts
```diff
@@ -565,6 +565,114 @@ describe("calcite-date-picker", () => {
...
+    expect(await datePicker.getProperty("value")).toBe("2005-03-12");
+
+    await selectDay("20050307", page, "mouse");
+    await page.waitForChanges();
+
+    expect(await datePicker.getProperty("value")).toBe("2005-03-07");
+  });
+
+  it("sets value to the selected day in the 1900s", async () => {
+    const page = await newE2EPage();
```
jcfranco, line 583: All of these tests share the same structure, so can you look at DRYing them up?

jcfranco (approved):
🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓
🗓📆🗓🗓🗓📆🗓🗓📆📆🗓🗓📆📆📆🗓🗓📆📆📆🗓📆📆📆📆🗓📆🗓
🗓📆📆🗓🗓📆🗓📆🗓🗓📆🗓🗓📆🗓🗓📆🗓🗓🗓🗓📆🗓🗓🗓🗓📆🗓
🗓📆🗓📆🗓📆🗓📆🗓🗓📆🗓🗓📆🗓🗓📆🗓🗓🗓🗓📆📆📆🗓🗓📆🗓
🗓📆🗓🗓📆📆🗓📆🗓🗓📆🗓🗓📆🗓🗓📆🗓🗓🗓🗓📆🗓🗓🗓🗓🗓🗓
🗓📆🗓🗓🗓📆🗓🗓📆📆🗓🗓📆📆📆🗓🗓📆📆📆🗓📆📆📆📆🗓📆🗓
🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓🗓

jcfranco:
Should the PR also include `input-date-picker` in the scope?

Issue #9026, opened 2024-04-01
[combobox] highlight matching search terms


emmilaakso212:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

As someone who is searching for an option in the combobox, I want to see how my search matches the options  that are being shown. 

### Acceptance Criteria

When searching using combobox, highlight matching values:

- by making them Avenir medium, color: text/1
- by adding a color: foreground/current as a highlight behind them

[See designs in Figma](https://www.figma.com/file/JXlrtCx8WmsXZRk8Lzd9Oq/Timezone%2Ftime-based-filters?type=design&node-id=244%3A170518&mode=dev&t=8qxI4u9uFamOc3x8-1)

### Relevant Info

If there are other components that have a search component, we should be consistent across them. For implementation in 24-R3. 

### Which Component

Combobox

### Example Use Case

Highlighting the search term in filtered results helps people understand why they're seeing an option

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Map Viewer

emmilaakso212:
cc: @jcfranco

driskull:
I think this would be ideal for the Search widget in the Maps SDK as well.

geospatialem:
Aiming for dev implementation prior to R3 based on Online priorities - currently targeted for June. cc @emmilaakso212

DitwanP:
🍡 Verified on `2.10.0-next.8` 👌
![combobox-highlight-filter-matches](https://github.com/Esri/calcite-design-system/assets/60022782/7ab39fae-8086-4ffd-877e-f1a428832662)

PR #9425, opened 2024-05-25
feat(combobox): highlight filter matches

jcfranco:
**Related Issue:** #9026

## Summary

Adds highlighting to combobox for the first occurrence of an item's label that matches the filter text.

### Next steps

For consistency, we can update `list` with similar functionality.


packages/calcite-components/src/components/combobox-item/combobox-item.tsx
```diff
@@ -252,12 +259,27 @@ export class ComboboxItem implements ConditionalSlotComponent, InteractiveCompon
...
+
+    const parts: (string | VNode)[] = this.textLabel.split(this.filterTextMatchPattern);
+
+    if (parts.length > 1) {
+      // we only highlight the first match
+      parts[1] = <span class={CSS.filterMatch}>{parts[1]}</span>;
+    }
+
+    return parts;
+  }
```
driskull, line 284: not sure if this is helpful, but this is what we have for the maps sdk.

```tsx
const regexContainsHTML = /<[a-z/][\s\S]*>/i;

const boldKeywords = (text: string, keywords: string): string => {
  const escapedKeywords = escapeRegExpString(keywords);
  const splitKeywords = escapedKeywords.split(/\s/);
  const pattern = new RegExp(`(${splitKeywords.join("|")})`, "gi");
  return text.replace(pattern, (match) => `<strong>${match}</strong>`);
};
```

... later


```
return regexContainsHTML.test(text) ? text : boldKeywords(text, searchTerm);
```

packages/calcite-components/src/components/combobox-item/combobox-item.tsx
```diff
@@ -252,12 +259,27 @@ export class ComboboxItem implements ConditionalSlotComponent, InteractiveCompon
...
+  private renderTextContent(): string | (string | VNode)[] {
+    if (!this.filterTextMatchPattern) {
+      return this.textLabel;
+    }
+
+    const parts: (string | VNode)[] = this.textLabel.split(this.filterTextMatchPattern);
+
+    if (parts.length > 1) {
+      // we only highlight the first match
+      parts[1] = <span class={CSS.filterMatch}>{parts[1]}</span>;
```
driskull, line 280: fixme: should we use `strong` instead of `span` for semantics?

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1103,7 +1105,12 @@ export class Combobox
         }
       });
 
+      this.filterTextMatchPattern = this.filterText && new RegExp(`(${this.filterText})`, "i");
+
       this.filteredItems = this.getFilteredItems();
+      this.filteredItems.forEach((item) => {
+        item.filterTextMatchPattern = this.filterTextMatchPattern;
```
driskull, line 1112: It feels a little odd passing a regex property to children. It seems "safer" to pass just the filterText but I do get creating the regex on each instance might not be as performant.

packages/calcite-components/src/components/combobox-item/combobox-item.tsx
driskull: We may want to follow the same logic since Search widget would likely want the same bold experience.

driskull (changes requested):
Looking good! 🫘

packages/calcite-components/src/components/combobox-item/combobox-item.tsx
jcfranco: Great suggestion. Will update!

packages/calcite-components/src/components/combobox/combobox.tsx
jcfranco: Not sure why this would be odd. We pass parent-level props to children in other components and we only need to build the pattern at the root.

driskull: Yeah, its fine. Just stuck out to me. Proceed!

packages/calcite-components/src/components/combobox-item/combobox-item.tsx
jcfranco: Isn't the highlight purely visual and not important content? Not sure if `strong` is applicable in this case. 

From [MDN](http://127.0.0.1:63228/Dash/utxknrwq/developer.mozilla.org/en-US/docs/Web/HTML/Element/strong.html#usage_notes):

> The \<strong\> element is for content that is of "strong importance," including things of great seriousness or urgency (such as warnings). This could be a sentence that is of great importance to the whole page, or you could merely try to point out that some words are of greater importance compared to nearby content.

In the `<b>` vs `<strong>` section, it recommends using CSS for styling too:

> It may help to realize that both are valid and semantic elements in HTML and that it's a coincidence that they both have the same default styling (boldface) in most browsers (although some older browsers actually underline \<strong\>). Each element is meant to be used in certain types of scenarios, and if you want to bold text for decoration, you should instead actually use the CSS [font-weight](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight) property.

What do you think we should do? Would `<b>` would be more appropriate or should we keep the current approach?

jcfranco: [`<mark>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/mark) is another option too.

driskull: Yeah mark 👌🏼

packages/calcite-components/src/components/combobox-item/combobox-item.tsx
jcfranco: Actually, could we discuss this in a follow-up issue? There are some differences in implementation and UI/UX that I think need to be considered first, such as:

* combobox filtering is immediate, search has minimum chars before showing matches
* combobox highlights the first occurrence, search highlights all matches
* combobox items do not support HTML as content, search does

There could be others, but these are the ones I noticed.

jcfranco: I did add the regexp char escape. 🏃🔠 Thanks for the bringing that up!

driskull: > combobox filtering is immediate, search has minimum chars before showing matches

I think this could be handled when we support combobox with dynamic data

> combobox highlights the first occurrence, search highlights all matches

Shouldn't combobox highlight all matches? Why does it only highlight the first occurrence?

> combobox items do not support HTML as content, search does

Yes, we can ignore this one. It doesn't do regex if there is HTML involved.

jcfranco: > Shouldn't combobox highlight all matches? Why does it only highlight the first occurrence?

This closely matches the current filtering behavior. It returns on the first match. I also did this to minimize elements updated as the user types since we don't have virtualization alternatives and showing all matches < 3 chars seems unnecessary. 

FWIW, I'm fine revisiting ☝️, but I would like design to give a pass before doing so. It would be an incremental enhancement this way vs rolling it back.

driskull: Yeah sounds good.

jcfranco:
@driskull I think I've addressed all items besides the highlighting behavior discussion. Would you mind reviewing again?

driskull (approved):
👍 Looks good!

Issue #8716, opened 2024-02-09
[input-time-zone] Allow to show UTC instead of GMT with the input time zone picker offset mode 


fendiaoxiaoshuazi:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

The current input time zone offset mode only allow to show GMT. Is it possible to support showing UTC as an option.

### Acceptance Criteria

Support showing UTC as an option.

### Which Component

calcite-input-time-zone

### Priority impact

p4 - not time sensitive

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Survey123

jcfranco:
@fendiaoxiaoshuazi Thanks for submitting this issue. Would you mind sharing more details behind this request? Is this about enhancing the existing "offset" mode or would this need to introduce a new mode that mainly displays UTC options a certain way?

fendiaoxiaoshuazi:
Hi, @jcfranco, we would like an enhancement to the existing "offset" mode. Currently, the "offset" mode can only show "GMT", and we would like an option to switch "GMT +/- *" to "UTC +/- *". 
![image](https://github.com/Esri/calcite-design-system/assets/26078106/e90b09fd-638d-46ce-9773-784387f44b1e)

Like the time zone picker of Scene Viewer.
![image](https://github.com/Esri/calcite-design-system/assets/26078106/e037dd75-8cac-4fe4-b2ad-b23a63e6b233)

fendiaoxiaoshuazi:
Currently Survey123 connect and field app use the UTC for the time picker. The web app utilizes the Calcite input-time-zone show GMT. Therefore, we need the UTC mode in the Calcite input-time-zone to make them consistent.

Here are the reasons why we choose UTC:
- According to Anna:
> The general guideline for all Esri  products is to follow UTC, not GMT. So if you display UTC – you’re good and don’t need to change anything.

jcfranco:
Belated thanks for the info, @fendiaoxiaoshuazi!

Adding related issue: https://github.com/Esri/calcite-design-system/issues/9018.

jcfranco:
Moved to the June release as this is needed for #9018. cc @geospatialem

[...]

PR #9426, opened 2024-05-25
feat(input-time-zone): add `offsetStyle` prop

jcfranco:
**Related Issue:** #8716

## Summary

Adds `offsetStyle` prop to specify whether offset should always use `UTC` or `GMT` for the offset label instead of using the user's locale. Valid options are:

  * `"user"` uses `UTC` or `GMT` depending on the user's locale,
   * `"gmt"` always uses `GMT`, and
   * `"utc"` always uses `UTC`.


packages/calcite-components/src/components/alert/alert.scss
```diff
@@ -299,7 +299,7 @@ $alertDurations:
   }
 }
 
-.container.focused .dismiss-progress:after {
+.container.focused .dismiss-progress::after {
```
driskull, line 302: nice catch. 🐟  Should we have an eslint for this?

driskull (approved):
🐟 👍

packages/calcite-components/src/components/alert/alert.scss
jcfranco: We already have a rule. 🎉 I believe this got fixed after merging `dev`.

PR #9429, opened 2024-05-26
ci: enable npm provenance statements

benelan:
## Summary

Enable npm provenance statements to improve supply-chain security for our users.

ref: https://docs.npmjs.com/generating-provenance-statements#using-third-party-package-publishing-tools

PR #9433, opened 2024-05-28
ci(provenance): add missing options from setup doc

benelan:
**Related Issue:** #9429

## Summary

Update the deployment GitHub Actions according to steps in the [npm provenance doc](https://docs.npmjs.com/generating-provenance-statements#publishing-packages-with-provenance-via-github-actions), which I missed because I was looking at the 3rd party section below it.

The [GitHub provenance doc](https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages#publishing-packages-to-the-npm-registry) doesn't mention permissions, so it might not be required.

We had originally hardcoded an older version of Ubuntu in #5942 due to errors, but it was resolved in Stencil v4.

### DIFF
--- a/.github/workflows/deploy-latest.yml
+++ b/.github/workflows/deploy-latest.yml
@@ -8,7 +8,9 @@ permissions:
   pull-requests: write
 jobs:
   release-please:
-    runs-on: ubuntu-20.04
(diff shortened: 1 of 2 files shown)


jcfranco (commented):
🚀

PR #9434, opened 2024-05-28
ci: combine related actions into single workflow file

benelan:
## Summary

Combine actions that are triggered on the same event into a single workflow file. This will improve performance, because it won't have to create/destroy as many containers. 

Additionally, this will cut down on concurrent jobs, which might prevent actions from being "queued". However, I'm not sure if we are reaching the [usage limit](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits), which is 500 concurrent jobs for the whole Enterprise organization.

The action that runs tests a second time for the eslint plugin was also removed.


jcfranco (approved):
its-beautiful

PR #9439, opened 2024-05-28
build: remove unnecessary dist files before publishing

benelan:
## Summary

Remove unnecessary files that somehow ended up in `dist` before publishing to NPM.


jcfranco (approved):
✨🧹✨

PR #9442, opened 2024-05-28
test(panel): add token theming tests

anveshmekala:
**Related Issue:** #7180 

## Summary

Add token theming tests for `panel` component.

### DIFF
--- a/packages/calcite-components/src/components/panel/panel.e2e.ts
+++ b/packages/calcite-components/src/components/panel/panel.e2e.ts
@@ -12,6 +12,7 @@ import {
   slots,
   t9n,
 } from "../../tests/commonTests";
+import { ComponentTestTokens, themed } from "../../tests/commonTests/themed";
 import { CSS, SLOTS } from "./resources";
 
 const panelTemplate = (scrollable = false) =>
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/panel/panel.e2e.ts
```diff
@@ -424,4 +425,127 @@ describe("calcite-panel", () => {
...
+        "--calcite-panel-height": {
+          selector: "calcite-panel",
+          targetProp: "blockSize",
+        },
+        "--calcite-panel-width": {
+          selector: "calcite-panel",
+          targetProp: "inlineSize",
+        },
+        "--calcite-panel-action-background-color": {
+          shadowSelector: `calcite-action >>> .button`,
```
alisonailea, line 473: I think this can just be "calcite-action"

packages/calcite-components/src/components/panel/panel.e2e.ts
```diff
@@ -424,4 +425,127 @@ describe("calcite-panel", () => {
...
+          targetProp: "--calcite-action-background-color",
+        },
+        "--calcite-panel-action-background-color-hover": {
+          shadowSelector: `calcite-action >>> .button`,
```
alisonailea, line 477: "calcite-action"

packages/calcite-components/src/components/panel/panel.e2e.ts
```diff
@@ -424,4 +425,127 @@ describe("calcite-panel", () => {
...
+          targetProp: "--calcite-action-background-color",
+          state: { hover: { attribute: "class", value: `button` } },
+        },
+        "--calcite-panel-action-background-color-active": {
+          shadowSelector: `calcite-action >>> .button`,
```
alisonailea, line 482: "calcite-action"

alisonailea (changes requested):
just a few little things to remove shadowDom selection '>>>'

packages/calcite-components/src/components/panel/panel.e2e.ts
anveshmekala: I think there is a confusion with naming here. The `--calcite-panel-action-background-color` token is only used for close action rendered when `closable` is set to `true` & the trigger of calcite-action-menu when header-menu-actions are slotted.

Should we separate these out and refactor names to be more specific ?

alisonailea (approved):

PR #9450, opened 2024-05-29
test: update themed to handle pseudoelements and special underline styling case

jcfranco:
**Related Issue:** #7180

## Summary

✨🧪✨

### Examples

```tsx
"--calcite-menu-item-action-border-color": {
   shadowSelector: `calcite-action::after`,
   targetProp: "backgroundColor",
},
```

```tsx
const tokens: ComponentTestTokens = {
  "--calcite-link-underline-color-start": {
    shadowSelector: `a`, // should probably be .anchor or similar class
    targetProp: "backgroundImage",
    },
  };
  themed(html`
    <calcite-link href="#" title="i am a link" icon-start='launch'>with an icon</calcite-link>`, 
    tokens
  );
});
```


packages/calcite-components/src/tests/commonTests/themed.ts
```diff
@@ -115,13 +121,14 @@ export function themed(componentTestSetup: ComponentTestSetup, tokens: Component
           styleTargets[selector][1].push(tokenStyle);
         }
         if (shadowSelector) {
-          target = shadowSelector ? await page.find(`${selector} >>> ${shadowSelector}`) : target;
+          const effectiveShadowSelector = shadowSelector.replace(/::.*$/, "");
+          target.el = await page.find(`${selector} >>> ${effectiveShadowSelector}`);
```
alisonailea, line 125: I tested this with alert, action, and menu-item. Action passed but when I tested alert and menu-item it only passed when I put a debugger on line 125 and then immediately continued to play the test. 

```
const tokens: ComponentTestTokens = {
        "--calcite-alert-dismiss-progress-background-color": {
          shadowSelector: `.${CSS.dismissProgress}::after`,
          targetProp: "backgroundColor",
        },
      };
      themed(`<calcite-alert open auto-close>`, tokens);
```


```
const tokens: ComponentTestTokens = {
      "--calcite-menu-item-action-border-color": [
        {
          shadowSelector: "calcite-action::after",
          targetProp: "backgroundColor",
        },
      ],
    };
    themed(`calcite-menu-item`, tokens);
    ```

alisonailea (changes requested):
I'm not sure if we need to add a wait or something else. It seems to work but the element parts aren't always there.

packages/calcite-components/src/tests/commonTests/themed.ts
jcfranco: This might be a separate issue since all updates in this PR are synchronous or update existing async code.

I noticed the PR was not synced up with `epic/7180-component-tokens`, which I have taken care of. Would you mind testing again?

PR #9456, opened 2024-05-29
feat(split-button, dropdown-item, button): update component tokens

alisonailea:
**Related Issue:** #7180

## Summary

Add theme e2e tests and update tokens
In order to get proper styling of split button this required some updates to dropdown-item and button


packages/calcite-components/src/components/button/button.scss
```diff
@@ -6,6 +6,10 @@
  * @prop --calcite-button-background-color: defines the component's background color.
  * @prop --calcite-button-border-color: defines the component's border color.
  * @prop --calcite-button-corner-radius: defines the component's corner radius.
+ * @prop --calcite-button-corner-radius-start-start: defines the component's start-start corner radius. Will fallback to --calcite-button-corner-radius if defined.
```
jcfranco, line 9: Should this, and similar lines, say `if not defined` vs `if defined`?

packages/calcite-components/src/components/dropdown-item/dropdown-item.scss
```diff
@@ -91,40 +104,44 @@
...
 }
 
 :host([selected]) .container:not(.container--none-selection),
 :host([selected]) .dropdown-link {
-  @apply text-color-1 font-medium;
+  @apply font-medium;
+  color: var(--calcite-dropdown-item-text-color, var(--calcite-color-text-1));
   & calcite-icon {
-    color: theme("backgroundColor.brand");
+    color: var(--calcite-dropdown-item-icon-color, theme("backgroundColor.brand"));
```
jcfranco, line 134: Should this be setting `--calcite-icon-color`?

packages/calcite-components/src/components/split-button/split-button.scss
```diff
@@ -14,28 +172,134 @@
   @apply w-full;
 }
 
+//--calcite-internal-split-button-loader-color: var(--calcite-color-inverse);
```
jcfranco, line 175: Commented out line should be removed. Applies to other commented-out lines.

If any commented-out line is waiting on a decision or question, let's add context and keep the comment.

jcfranco (approved):

packages/calcite-components/src/components/button/button.scss
alisonailea: I was thinking about this. I'm not sure what makes more sense. It will only fallback to the more general `-corner-radius` var if that var is defined. Which is why I used `if defined` but maybe that's implied?

alisonailea: > Are these just so Split Button can pass some through? Should these be made internal if so?

I thought about that. My reasoning for making them public was that customers may be using calcite-buttons to create their own components with similar edge case requirements so providing the option to set the corners separately would be beneficial

packages/calcite-components/src/components/split-button/split-button.scss
alisonailea: thanks for the catch

Issue #9468, opened 2024-05-30
Initialization regression in `components` output target


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

As a result of upgrading from `2.8.3` to `2.8.4` there were some failing screenshot tests from internal teams. We should restore the `componentDidLoad` hooks to accommodate changes in initialization.

### Expected Behavior

No failing screenshot tests

### Reproduction Sample

https://stackblitz.com/edit/github-ojdkyo-zgju9e

### Reproduction Steps

1. Observe component's initial value set when using v1.8.3.
2. Update `package.json` to use `@esri/calcite-components@2.8.4`. Notice value is not displayed in the input.

### Reproduction Version

3.8.4

### Relevant Info

Seems to pertain to be related to https://github.com/Esri/calcite-design-system/pull/9443

### Priority impact

impact - p1 - need for current milestone

### Impact

High

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

geospatialem:
Verified in `2.8.5`

PR #9443, opened 2024-05-28
fix: defer floating-ui updating until component is connected and open

jcfranco:
**Related Issue:** #9397

## Summary

This updates the `FloatingUIComponent` implementation to defer calling `autoUpdate` until the component is open and connected (following `floating-ui` [usage notes](https://floating-ui.com/docs/autoupdate#usage)).

PR #9470, opened 2024-05-30
fix(combobox, dropdown, input-date-picker, popover, tooltip): fix initialization logic in components output target

jcfranco:
**Related Issue:** #9468

## Summary

https://github.com/Esri/calcite-design-system/pull/9443 introduced a regression caused by using the `componentOnLoad` util to consolidate initialization logic shared between `connectedCallback` and `componentDidLoad`. The util appears to resolve at a different time in the `components` output target, which messed up with initialization of certain components.

**Note**: there are no accompanying tests as this is not reproducible in the test environment, which uses the lazy-loaded output.

### DIFF
--- a/packages/calcite-components/src/components/popover/popover.tsx
+++ b/packages/calcite-components/src/components/popover/popover.tsx
@@ -310,7 +310,6 @@ export class Popover
     if (this.referenceElement && !this.effectiveReferenceElement) {
       this.setUpReferenceElement();
     }
-    connectFloatingUI(this, this.effectiveReferenceElement, this.el);
     this.hasLoaded = true;
   }
 
--- a/packages/calcite-components/src/components/popover/popover.tsx
+++ b/packages/calcite-components/src/components/popover/popover.tsx
@@ -55,7 +55,7 @@ import {
 } from "../../utils/loadable";
 import { createObserver } from "../../utils/observers";
 import { FloatingArrow } from "../functional/FloatingArrow";
-import { componentOnReady, getIconScale } from "../../utils/component";
+import { getIconScale } from "../../utils/component";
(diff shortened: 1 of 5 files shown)


packages/calcite-components/src/components/popover/popover.tsx
```diff
@@ -307,6 +307,11 @@ export class Popover
 
   componentDidLoad(): void {
     setComponentLoaded(this);
+    if (this.referenceElement && !this.effectiveReferenceElement) {
+      this.setUpReferenceElement();
+    }
+    connectFloatingUI(this, this.effectiveReferenceElement, this.el);
```
driskull, line 313: ` this.setUpReferenceElement();` already calls `connectFloatingUI` so no need here.

packages/calcite-components/src/components/tooltip/tooltip.tsx
```diff
@@ -172,6 +171,14 @@ export class Tooltip implements FloatingUIComponent, OpenCloseComponent {
     }
   }
 
+  componentDidLoad(): void {
+    if (this.referenceElement && !this.effectiveReferenceElement) {
+      this.setUpReferenceElement();
+    }
+    connectFloatingUI(this, this.effectiveReferenceElement, this.el);
```
driskull, line 178: ` this.setUpReferenceElement();` already calls `connectFloatingUI` so no need here.

driskull (approved):
👍 Thanks @jcfranco 💯

packages/calcite-components/src/components/popover/popover.tsx
jcfranco: Thanks for the catch! Forgot to carry that over from previous feedback.

Issue #9454, opened 2024-05-29
[Block Section] When `toggle-display` is set to `"switch"` unable to toggle


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

A regression was introduced to `block-section`, where when `toggle-display` is set to `"switch"` unable to toggle.

![issue](https://github.com/Esri/calcite-design-system/assets/5023024/c47a8e41-7655-438f-b96c-d50f5cca5bab)

### Expected Behavior

Toggle behavior for both the `"switch"` and `"button"`

![before](https://github.com/Esri/calcite-design-system/assets/5023024/901bbb34-55d2-4000-99be-e18582d00ffc)

### Reproduction Sample

https://codepen.io/geospatialem/pen/BaepzdW

### Reproduction Steps

1. Open the [sample](https://codepen.io/geospatialem/pen/BaepzdW)
2. Try to collapse/expand the `block-section` with the toggle switch
3. Observe no change
4. If you change to the button switch or `2.8.4` observe the switch works as-intended

### Reproduction Version

`2.9.0`

### Relevant Info

Possibly introduced via https://github.com/Esri/calcite-design-system/pull/9194, does not impact the default toggle (`"button"`).

### Regression?

`2.8.4`

### Priority impact

impact - p1 - need for current milestone

### Impact

High, to accommodate folks using the `"switch"` toggle as soon as possible.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Map Viewer

geospatialem:
Verified in `2.10.0-next.0`:

![verify](https://github.com/Esri/calcite-design-system/assets/5023024/dbc79ab7-a619-4103-986a-98a7ca824262)

PR #9472, opened 2024-05-30
fix(block-section): restore block toggling when clicking on the header switch

jcfranco:
**Related Issue:** #9454

## Summary

Fixes toggling issue caused by moving switch outside toggle container.

This regression was introduced by #9194.


Elijbet (approved):
Looks good! Thanks for cleaning that up.

alisonailea (approved):

Issue #9400, opened 2024-05-22
Nested child list items are displayed briefly until component recognizes them and adds the caret and hides them


sagewall:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Nested child list items are displayed briefly until component recognizes them and adds the caret and hides them.  This results in a flickering behavior. The flickering is more apparent in complex lists like the JavaScript SDK LayerList.

### Expected Behavior

The child list items are hidden until the caret unhides them.

### Reproduction Sample

https://codepen.io/sagewall/pen/GRaqWya

### Reproduction Steps

1. Expand the "Catalog layer with a group layer" list item
2. Expand the "Layers In View" list item
3. Notice the flickering as the list items briefly appear and then hide
4. Expand the "Testing Tooltip input" list item
5. Notice the flickering

### Reproduction Version

2.8.3

### Relevant Info

Seems independent of browser and OS

### Priority impact

impact - p2 - want for an upcoming milestone

### Impact

This is impacting the JavaScript API and all the Esri web applications that use it such as Map Viewer, Instant Apps and Experience Builder.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

jcfranco:
Moving this to the June milestone since @driskull has a PR standing by. 🎉

geospatialem:
Verified in `2.10.0-next.0`:

![verify](https://github.com/Esri/calcite-design-system/assets/5023024/cea4c7e6-89a2-4564-94af-e86f2eb166e4)

PR #9474, opened 2024-05-30
fix(list-item): hide nested list items by default

driskull:
**Related Issue:** #9400

## Summary

- Hides nested/indented list items by default
- Removes unnecessary div


alisonailea (approved):
simplification! 🎉

jcfranco (approved):
Awesome! 📝

There doesn't seem to be any reported diffs in the Chromatic build. If missing, can you add a screenshot test?

driskull:
> There doesn't seem to be any reported diffs in the Chromatic build. If missing, can you add a screenshot test?

I don't think a screenshot test would cover this. Its basically a flash of content.

jcfranco:
Gotcha. Thanks for clarifying!

Issue #5539, opened 2022-10-21
action-bar `messageOverride` props correspond to both text and label


geospatialem:
### Summary

We presently support assistive technology (AT) users with the `action-bar` component for the collapse/expand buttons. 

However, the support also provides the same context visually updating the text, which in some use cases doesn't provide a good interface, or user experience for visual users.

![screenshot of the action bar menu with updated intlCollapse prop](https://user-images.githubusercontent.com/5023024/197250228-91e26519-a518-4cae-b71c-eb9a107abe40.png)

Could we support both AT and visual users in providing opportunities to support both use cases?

### Actual Behavior

When setting the `intlCollapse` property the visual text and label contain the value.

### Expected Behavior

Developers should be able to specify different values for the `action-bar` collapse button with the displayed text (perhaps new props, `textCollapse`/`textExpand` versus what is read by AT `intlCollapse`/`intlExpand`)?

For instance, visual users may only need displayed text of "Collapse", whereas additional context is needed to support AT users, such as "Collapse accordion sample menu bar".

### Reproduction Sample

https://codepen.io/geospatialem/pen/Jjvgrqv

### Reproduction Steps

1. Open the [sample](https://codepen.io/geospatialem/pen/Jjvgrqv)
2. Observe when supplied with the `intlCollapse` prop the "Collapse" text is replaced with the value (e.g., "Collapse accordion sample menu bar").

### Reproduction Version

beta.95

### Relevant Info

There may still be a use case for allowing the developer to change the visual text, so perhaps a new property should be introduced to account for both the accessible label (`intlCollapse`/`intlExpand` and text (perhaps `textCollapse`/`textExpand`)?

#### Additional background:
- A [sample](https://codepen.io/geospatialem/pen/oNdKowV) with a native button that is supplied with different visual text and an accessible label for context.
- Many other CC have similar options, such as [Loader](https://developers.arcgis.com/calcite-design-system/components/loader) where `label` and `text` properties co-exist on the component.

### Esri team

Calcite (dev)

geospatialem:
Spike to determine if this is still an issue that we should address with the 1.x and 2.x releases.

geospatialem:
This is still valid, updated the[ Codepen](https://codepen.io/geospatialem/pen/Jjvgrqv) to 2.0.0 to reflect the different properties and attributes.

Currently if the `messageOverrides` property is overwritten context is provided to AT, but the text in the component is also altered, which may not support a good user experience.

[...]

PR #9497, opened 2024-06-03
feat(action-bar, action-pad): add expandLabel and collapseLabel to messages

driskull:
**Related Issue:** #5539

## Summary

-  add expandLabel and collapseLabel to messages


alisonailea (approved):

jcfranco (approved):
LGTM! Before merging, can you add a test that asserts on the label/text separation? This will probably need to override messages since they're the same text.

driskull:
@jcfranco tests added

PR #9501, opened 2024-06-03
fix(popover): prevent closing when component is connected to the DOM via a click

jcfranco:
**Related Issue:** #9504

## Summary

This fixes an issue where any clicks when the popover is connected to the DOM would cause it to be immediately closed.

This would happen because `PopoverManager`'s window `click` handler was being added and invoked during `connectedCallback` and not after the component was initialized (somewhat similar to https://github.com/Esri/calcite-design-system/pull/9373).

**Note**: there are no accompanying tests as this is not reproducible in the test environment, which uses the lazy-loaded output.

### DIFF
--- a/packages/calcite-components/src/components/popover/popover.tsx
+++ b/packages/calcite-components/src/components/popover/popover.tsx
@@ -302,7 +302,7 @@ export class Popover
 
   componentDidLoad(): void {
     setComponentLoaded(this);
-    this.setUpReferenceElement();
+    this.setUpReferenceElement(true);
 
     if (this.open) {
       onToggleOpenCloseComponent(this);
--- a/packages/calcite-components/src/components/popover/popover.tsx
+++ b/packages/calcite-components/src/components/popover/popover.tsx
@@ -293,6 +293,10 @@ export class Popover
     connectLocalized(this);
     connectMessages(this);
     connectFocusTrap(this);
+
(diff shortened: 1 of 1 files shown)


packages/calcite-components/src/components/popover/popover.tsx
```diff
@@ -292,12 +292,11 @@ export class Popover
...
     connectMessages(this);
-    this.setUpReferenceElement(this.hasLoaded);
     connectFocusTrap(this);
 
-    if (this.open) {
-      onToggleOpenCloseComponent(this);
-    }
+    // we set up the ref element in the next frame to ensure PopoverManager
+    // event handlers are invoked after connect (mainly for `components` output target)
+    requestAnimationFrame(() => this.setUpReferenceElement(this.hasLoaded));
```
driskull, line 299: Would we need to do this for the other `floating-ui` components? Or is it only popover because of the PopoverManager?

driskull (approved):
Looks good @jcfranco 👍

packages/calcite-components/src/components/popover/popover.tsx
jcfranco: I believe it's the latter. I tested with other components, but `popover` seemed to be the the only one affected. I'll do another sweep to make sure.

jcfranco: Confirmed.

Issue #9412, opened 2024-05-24
[Color Picker] Adjust component's scope and color swatch sizing


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Improve the `color-picker` design with the alpha channel and color swatch, where:

1. The color swatch size is smaller when `alpha-channel` is not used
2. The scope thumb is also reduced in size to match the slider heights

### Acceptance Criteria

- [ ] Reduce size of scope to 14px circles to match slider height (applies to all scales)
- [ ] When `alpha-channel` is not used, reduce sampled color swatch to match size of saved color swatches: 20px for small scale and 24 px for medium scale. No changes are required for large scale. 

### Relevant Info

- **For full details, Figma file can be found [here](https://www.figma.com/design/3YZypHvasD8KT4Y6cmQRKy/Color-Picker-[add-opacity-support]-[issue-%23749]?node-id=1002-32922&t=YL585DzJhOrZgS3N-11).**
- Part of the https://github.com/Esri/calcite-design-system/issues/7163 epic

### Which Component

`color-picker`

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

DitwanP:
🍡 Verified locally on `dev`

PR #9523, opened 2024-06-05
feat(color-picker): adjust thumb and color preview sizes to follow updated spec

jcfranco:
**Related Issue:** #9412

## Summary

This decreases the size of thumbs and uses the component's scale for the color preview when alpha channel is not displayed.


ashetland:
😍

Issue #9526, opened 2024-06-05
Move component types into component-specific interfaces files


aPreciado88:
### Description

Some types in [interfaces](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/interfaces.ts) are not being shared between components. Move these types into component-specific `interfaces.ts` files.

### Proposed Advantages

[interfaces](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/interfaces.ts) will only contain types that are shared.

geospatialem:
Verified on the `main` branch. ✨

PR #9527, opened 2024-06-05
refactor: move component types into component specific interfaces files

aPreciado88:
**Related Issue:** [#9526](https://github.com/Esri/calcite-design-system/issues/9526)

## Summary
Moves unshared types in [interfaces](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/interfaces.ts) into component-specific `interfaces.ts` files.


alisonailea (approved):

packages/calcite-components/src/components/tile-select-group/interfaces.ts
```diff
@@ -1 +1,2 @@
 export type TileSelectGroupLayout = "vertical" | "horizontal";
+export type TileSelectGroupDir = "ltr" | "rtl";
```
jcfranco, line 2: I thought we had a shared `Dir` type, but it doesn't look that way. Can you move this one to the shared interfaces file? We should be able to use it in more places.

jcfranco (approved):
One minor comment, but this LGTM! 🚀

packages/calcite-components/src/components/tile-select-group/interfaces.ts
aPreciado88: I moved `Dir` into the shared interfaces file!

Issue #9521, opened 2024-06-05
[list] - unable to drag and drop list items successfully with nested lists


sagewall:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When a list contains another list in a list-item and drag-enabled is true the list items fail to move as expected and the list becomes unresponsive when an item is dragged.

I think this was previously partially addressed with #8728 but the fix perhaps didn't address nested list items.

### Expected Behavior

The drag and drop behavior works on mobile devices using touch gestures with nested lists.

### Reproduction Sample

https://codepen.io/sagewall/pen/LYoyryy

### Reproduction Steps

1. Open this [codepen](https://codepen.io/sagewall/full/LYoyryy) on a mobile device
2. Expand the "Nested List Item"
3. Try to drag and drop any of the list items
4. Observe how the list becomes unresponsive or the list items fail to move

### Reproduction Version

2.8.3 and 2.9.0

### Relevant Info

Pixel 7, Chrome

### Priority impact

impact - p1 - need for current milestone

### Impact

This impacts the LayerList, BasemapLayerList, and TableList Javascript SDK widgets and the applications that use them like Map Viewer, Experience Builder and Instant Apps.

Also identified as a bug via `BUG-000167708`.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

driskull:
@jcfranco when we patched sortable, we patched it with my assumed fix but not the correct fix that went into Sortable. We may need to re-patch.

driskull:
Correct patch: https://github.com/SortableJS/Sortable/pull/2347/files

incorrect patch: https://github.com/Esri/calcite-design-system/blob/e26fce6c0a7fb3f37fd084176e8846e135df295b/patches/sortablejs%2B1.15.1.patch

driskull:
I can't reproduce this on my iphone or iphone simulator. Is it only on chrome on Pixel 7? Or can it be reproduced on an ios device?

geospatialem:
> I can't reproduce this on my iphone or iphone simulator. Is it only on chrome on Pixel 7? Or can it be reproduced on an ios device?

[...]

PR #9532, opened 2024-06-06
fix: set the correct Sortable parent when within a shadowRoot

driskull:
**Related Issue:** #9521

## Summary

- patch sortablejs with the correct update to fix dragging shadow elements on mobile devices
- https://github.com/SortableJS/Sortable/pull/2347 https://github.com/SortableJS/Sortable/issues/2346


driskull:
@jcfranco can we do this one for June release?

jcfranco (approved):
LGTM!

It looks like the related issue has been closed and verified already. Should we reopen and bounce back or do we need to create a new issue? +@geospatialem for guidance.

driskull:
> It looks like the related issue has been closed and verified already. Should we reopen and bounce back or do we need to create a new issue

Yeah, we did fix the issue but we didn't do it the correct way according to Sortablejs. So we can reopen or create a new issue .

geospatialem:
> Yeah, we did fix the issue but we didn't do it the correct way according to Sortablejs. So we can reopen or create a new issue.

Ah, good to know - I'll re-open the issue.

Issue #7492, opened 2023-08-08
Block: remove redundant indicator behavior


SkyeSeitz:
**monday.com sync:** #9469724114

### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

**[Current behavior](https://developers.arcgis.com/calcite-design-system/components/block/)**
Loader populates both in the header and in the content area when expanded, which feels redundant and may imply two separate elements are loading.

<img width="464" alt="image" src="https://github.com/Esri/calcite-design-system/assets/29716057/e672d714-d5f9-4307-872f-94fb622fd6e0">

### Acceptance Criteria

[**Expected behavior**](https://www.figma.com/file/AHkDvoUEgkweh9Mr9HDoR5/Block%3A-remove-redundant-indicator-behavior-%5B%5D?type=design&node-id=104%3A389&mode=dev)
Loader populates only in header when collapsed and only in content area when expanded.
<img width="1720" alt="image" src="https://github.com/Esri/calcite-design-system/assets/29716057/005746db-4cc7-4de1-8258-8a51d0459a2a">

### Relevant Info

Building off of #6485

### Which Component

Calcite Block

### Priority impact

p4 - not time sensitive

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

Calcite (design)

SkyeSeitz:
cc @ashetland

ashetland:
@macandcheese @asangma Is there a reason use case for having a loading indicator in the header while it's open that we're missing?

macandcheese:
Not from my side. I do think it's strange the loader replaces the icon at all...

I wonder if a `calcite-progress` on the top of the block would make sense? It would eliminate the need to replace a user-placed element, and could subtly persist even while the block was expanded + `calcite-loader` present to prevent hiding / showing elements when expanding / collapsing. You do kind of lose the immediacy of the spinner being centered within the item when blocks are all collapsed, but this kind of loading often seen in "listy places" as well across Apps...

https://github.com/Esri/calcite-design-system/assets/4733155/3a2b3342-eb10-42c0-bdd8-8724df215265

macandcheese:
I do like that the loader is spinner in the above screenshots - depending on height of Block content, it can feel too large currently.

ashetland:
[...]

Issue #7315, opened 2023-07-12
Remove margin-right on Loader in Block header


ashetland:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

In issue #6485 and PR #7239, the `inline` Loader was repositioned in Block's header. Too improve spacing, we'd like to remove the 8px `margin-right` that comes from Loader when it is set to `inline`.

### Acceptance Criteria

Loader's `margin-right` reduced to 0px.

### Relevant Info

cc @Elijbet 

![image](https://github.com/Esri/calcite-design-system/assets/108549080/98cc377f-a5be-4d91-8167-4dbad07f9648)

### Which Component

Block

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

Calcite (design)

ashetland:
Closing in favor of #7492.

PR #9535, opened 2024-06-06
feat(block): add `icon start/end` properties (deprecate `icon` slot and `status`), add `actions-end` slot (deprecate `control`), add `content-start`

Elijbet:
**Related Issue:** #4932

## Summary

- Add `icon-start` and `icon-end` properties. 
- Deprecate `status`.
- Deprecate `icon` in favor of `icon-start`.
- Add `actions-end` slot and deprecate `control` slot.
- Add `content-start`.


packages/calcite-components/src/components/block/block.tsx
```diff
@@ -364,8 +435,23 @@ export class Block
             title={toggleLabel}
           >
             {headerContent}
-            <calcite-icon aria-hidden="true" class={CSS.toggleIcon} icon={collapseIcon} scale="s" />
+            <div class={CSS.iconEndContainer}>
+              {iconEndEl}
+              {this.renderActionsEnd()}
+              <calcite-icon
+                aria-hidden="true"
```
driskull, line 442: I don't think the aria-hidden is needed here.

driskull (approved):
👍

packages/calcite-components/src/components/block/block.stories.ts
```diff
@@ -275,3 +275,36 @@ export const icons_TestOnly = (): string => html`
...
+      icon="compass"
+      slot="content-start"
+      style="color: var(--calcite-color-status-success)"
+      scale="s"
+    ></calcite-icon>
+    <calcite-action appearance="transparent" icon="ellipsis" text="menu" label="menu" slot="actions-end" />
+  </calcite-block>
+`;
+
+export const actionsEndContentStartIconStartEnd = (): string => html`
```
jcfranco, line 298: If both screenshot tests are adding coverage for start/end icons, can you merge them into a single test and add info for each test case (similar to [this](https://github.com/Esri/calcite-design-system/blob/99eac2e4e63f96c95eb354883921b3f56bf7841a/packages/calcite-components/src/components/slider/slider.stories.ts#L466-L524))?

packages/calcite-components/src/components/block/block.tsx
```diff
@@ -42,13 +47,16 @@ import {
...
+import { FlipContext } from "../interfaces";
 import { CSS, ICONS, IDS, SLOTS } from "./resources";
 import { BlockMessages } from "./assets/block/t9n";
 
 /**
  * @slot - A slot for adding custom content.
- * @slot icon - A slot for adding a leading header icon with `calcite-icon`.
- * @slot control - A slot for adding a single HTML input element in a header.
+ * @slot actions-end - A slot for adding actionable `calcite-action` elements after the content of the component. It is recommended to use two or fewer actions.
+ * @slot icon - [Deprecated] A slot for adding a leading header icon with `calcite-icon`. Use `icon-start` instead.
```
jcfranco, line 57: ✨🏆✨

packages/calcite-components/src/components/block/block.tsx
```diff
@@ -191,14 +210,18 @@ export class Block
 
   @Element() el: HTMLCalciteBlockElement;
 
+  @State() defaultMessages: BlockMessages;
+
+  @State() hasContentStart = false;
```
jcfranco, line 215: Should this be moved above `hasEndActions`?

packages/calcite-components/src/components/block/block.tsx
```diff
@@ -281,6 +304,14 @@ export class Block
     this.transitionEl = el;
   };
 
+  private actionsEndSlotChangeHandler = (event: Event): void => {
+    this.hasEndActions = slotChangeHasAssignedElement(event);
+  };
+
+  handleContentStartSlotChange = (event: Event): void => {
```
jcfranco, line 311: Can you make this private?

packages/calcite-components/src/components/block/block.tsx
```diff
@@ -338,10 +386,33 @@ export class Block
     const { collapsible, el, loading, open, heading, messages } = this;
 
     const toggleLabel = open ? messages.collapse : messages.expand;
+    const { iconFlipRtl } = this;
+
+    /** Icon scale is not variable as the component does not have a scale property */
+    const iconStartEl = this.iconStart ? (
+      <calcite-icon
```
jcfranco, line 393: Can you DRY this up by creating an icon-rendering function?

packages/calcite-components/src/demos/block.html
```diff
@@ -240,6 +240,26 @@
         </div>
       </div>
 
+      <!-- block loader, icon-start/end, content-start, actions-end -->
```
jcfranco, line 243: Let's drop this (and future) demo comments. This is already specified below.

jcfranco (approved):

Before merging, can you update the PR title to mention the new slots and to call out the deprecated ones (similar to [this](https://github.com/Esri/calcite-design-system/pull/7435))?

packages/calcite-components/src/components/block/resources.ts
```diff
@@ -5,15 +5,20 @@ export const IDS = {
...
   button: "button",
   container: "container",
   content: "content",
+  contentStart: "content-start",
   controlContainer: "control-container",
   description: "description",
   header: "header",
   headerContainer: "header-container",
   heading: "heading",
   icon: "icon",
```
alisonailea, line 18: this is no longer a used CSS class

packages/calcite-components/src/components/block/block.scss
```diff
@@ -29,13 +30,21 @@
...
   grid-template-areas: "handle header control menu";
   grid-column: header-start / menu-end;
   grid-row: 1 / 2;
 }
 
+.header ::slotted(calcite-icon) {
+  margin-inline-end: var(--calcite-spacing-md);
+}
+
+.header .icon ::slotted([slot="icon"]) {
```
alisonailea, line 44: I don't think these styles are reachable as CSS.icon isn't used anywhere.

packages/calcite-components/src/components/block/block.scss
```diff
@@ -93,9 +102,18 @@ calcite-handle {
 }
 
 .icon {
```
alisonailea, line 104: remove?

packages/calcite-components/src/components/block/block.scss
```diff
@@ -93,9 +102,18 @@ calcite-handle {
...
+}
+
+.icon--start,
+.icon--end {
+  @apply text-color-3;
+
+  margin-inline-end: var(--calcite-spacing-md);
+}
+
+.status-icon {
```
alisonailea, line 115: merge this with `.header ::slotted(calcite-icon)` at line 40?

```css
.header ::slotted(calcite-icon),
.status-icon,
.icon--start,
.icon--end {
  margin-inline-end: var(--calcite-spacing-md);
}
```

packages/calcite-components/src/components/block/block.tsx
```diff
@@ -364,8 +435,16 @@ export class Block
             title={toggleLabel}
           >
             {headerContent}
-            <calcite-icon aria-hidden="true" class={CSS.toggleIcon} icon={collapseIcon} scale="s" />
+            <div class={CSS.iconEndContainer}>
+              {iconEndEl}
+              <calcite-icon class={CSS.toggleIcon} icon={collapseIcon} scale="s" />
```
alisonailea, line 440: toggleIcon is missing from CSS resources?

alisonailea (changes requested):

packages/calcite-components/src/components/block/block.scss
Elijbet: `CSS.icon` is used to wrap `loader, status-icon, and icon-slot`, we keep them as deprecated for now

packages/calcite-components/src/components/block/block.scss
Elijbet: icon is still used with deprecated items

packages/calcite-components/src/components/block/block.tsx
Elijbet: It's there I believe.

packages/calcite-components/src/components/block/block.tsx
```diff
@@ -294,7 +325,7 @@ export class Block
     return [loading ? <calcite-scrim loading={loading} /> : null, defaultSlot];
   }
 
-  renderIcon(): VNode[] {
+  renderLoaderStatusIcon(): VNode[] {
```
jcfranco, line 328: Can you make this private? For context, we'll want to explicitly use the private modifiers wherever applicable. Otherwise, Stencil will include them in types.

packages/calcite-components/src/components/block/block.tsx
```diff
@@ -334,14 +382,42 @@ export class Block
     ) : null;
   }
 
+  renderIcon(iconPosition: string): VNode {
```
jcfranco, line 385: Let's type the position argument as it doesn't take any string. Also, it can probably be simplified to `position` since we know it's icon-related due to the method name.

packages/calcite-components/src/components/block/block.tsx
```diff
@@ -334,14 +382,42 @@ export class Block
...
+    if (iconPosition === undefined) {
+      return null;
+    }
+
+    const flipRtl =
+      iconFlipRtl === "both" || iconPosition === "iconStart"
+        ? iconFlipRtl === "start"
+        : iconFlipRtl === "end";
+
+    const iconValue = iconPosition === "iconStart" ? this.iconStart : this.iconEnd;
```
jcfranco, line 397: The position values should be simplified to `start`/`end` since we know the context from the method name. You can map to the CSS class here.

packages/calcite-components/src/components/block/block.tsx
```diff
@@ -334,14 +382,42 @@ export class Block
...
+
   render(): VNode {
     const { collapsible, el, loading, open, heading, messages } = this;
 
     const toggleLabel = open ? messages.collapse : messages.expand;
 
     const headerContent = (
       <header class={CSS.header} id={IDS.header}>
-        {this.renderIcon()}
+        {this.renderIcon(this.iconStart ? "iconStart" : undefined)}
```
jcfranco, line 418: It seems odd to check for `iconStart`/`iconEnd` both outside (to determine the position) and within  `renderIcon`. I think this should be changed to one of the following:

```tsx
{this.renderIcon("start")} // knows the position, so it can check for the start icon and render if possible
```

or 

```tsx
{this.renderIcon(this.iconStart, "start")} // renders whatever icon and position is passed
```

Leaning more towards the first one as it takes care of matching the icon for us.

ashetland:
Looks great!

ashetland:
> This is looking great! More on the design side - can we consider using the `xxs` spacing for the loading spinner? It is slightly misaligned with other Blocks (or, the same block shifts when setting loading). cc @ashetland @SkyeSeitz

This was documented in #7315, but closed in favor of #7492.

packages/calcite-components/src/components/block/resources.ts
```diff
@@ -42,3 +42,5 @@ export const ICONS = {
   valid: "check-circle",
   invalid: "exclamation-mark-triangle",
 };
+
+export type Position = "start" | "end";
```
jcfranco, line 46: Can you extract from the shared `Position` interface instead of creating a component-specific one? Like [this](https://github.com/Esri/calcite-design-system/blob/8edeb3645fb8dee31df60165b0f82e13aa290e19/packages/calcite-components/src/components/accordion-item/accordion-item.tsx#L76)?

jcfranco (approved):

ashetland:
Lookin' even better! 🚀🤘

packages/calcite-components/src/components/block/block.tsx
Elijbet: Could you review these one more time @alisonailea, in case I missed something?

Issue #8848, opened 2024-02-28
Provide ability to detect Calcite version at runtime


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

To assist with client issue troubleshooting, the user is seeking to display the Esri JavaScript API version in a simple popup using something similar to the Maps SDK's [Kernel.version](https://developers.arcgis.com/javascript/latest/api-reference/esri-kernel.html#version). 

Customers have applications that include a mapping experience with Maps SDK for JS and Calcite Design System and it is known that Calcite updates are independent of the JS updates. The enhancement request, `ENH-000165422`, would like to include the Calcite Design System version in the information window. 

### Acceptance Criteria

- [ ] Allow users the ability to check the current version of Calcite
- [ ] Allow users the ability to add the version to their own custom informational window 
- [ ] Ensure that context is provided if another product, such as Maps SDK for JS, also uses Calcite has a version
    - Messaging is provided for the Calcite version
    - Messaging is provided for the product (_optional_)

### Relevant Info

ENH-000165422 requested on 02/27/2024

### Which Component

N/A - applies to the design system.

### Example Use Case

For a user to check the current version of Esri libraries, similar to what is provided in the Maps SDK for JS with [Kernel.version](https://developers.arcgis.com/javascript/latest/api-reference/esri-kernel.html#version) for Calcite, where an informational window can provide context to the Calcite version used.

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

driskull:
It would also be helpful to include a version comment in the distribution files.

The Maps SDK includes the following in their files:

```
// All material copyright Esri, All Rights Reserved, unless otherwise specified.
// See https://js.arcgis.com/4.29/esri/copyright.txt for details.
```

It would be useful to have a copyright and version like that in ours. Especially in the CSS files since they won't be able to log anything.

jcfranco:
@driskull We do provide similar info in the source files:

https://unpkg.com/browse/@esri/calcite-components@2.6.0/dist/components/index.js
[...]

PR #9536, opened 2024-06-07
feat: include version in global config

jcfranco:
**Related Issue:** #8848

## Summary

This updates the globalScript to stamp the version onto an existing global calciteConfig var and will create one if not defined.

### DIFF
--- a/packages/calcite-components/src/utils/config.ts
+++ b/packages/calcite-components/src/utils/config.ts
@@ -16,6 +16,8 @@ export interface CalciteConfig {
 
   /**
    * Contains the version of the Calcite components.
+   *
+   * @readonly
    */
   version?: string;
 }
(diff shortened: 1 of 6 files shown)


packages/calcite-components/src/utils/config.ts
```diff
@@ -13,8 +13,36 @@ export interface CalciteConfig {
    * @see https://github.com/focus-trap/focus-trap#createoptions
    */
   focusTrapStack: FocusTrap[];
+
+  /**
+   * Contains the version of the Calcite components.
+   */
```
driskull, line 19: maybe we can document this as readOnly somehow?

packages/calcite-components/src/utils/config.ts
```diff
@@ -13,8 +13,36 @@ export interface CalciteConfig {
...
+const version = "__CALCITE_VERSION__"; // version number is set by build
+
+/**
+ * Stamp the version onto the global config.
+ */
+export function stampVersion(): void {
+  if (customConfig) {
+    if (customConfig.version) {
+      console.warn(
+        `Calcite config version "${customConfig.version}" will be set by global script. This may cause unexpected behavior.`,
```
driskull, line 36: Wording suggestion:

`calcite-components version "${customConfig.version}" is defined by a global script. This may cause unexpected behavior.`,

driskull (approved):
💥

benelan (approved):
:+1:

packages/calcite-components/src/utils/config.ts
jcfranco: We can add `@readonly` to the interface doc, but it won't show up in a readme nor the `docs-json.json` file. @geospatialem @DitwanP @macandcheese Do you have any suggestions?

packages/calcite-components/src/utils/config.ts
jcfranco: Revisited message to also include the current version and that the original version will be preserved.

Issue #5063, opened 2022-08-01
Filter: allow users to choose to match or ignore specific item properties


benelan:
### Description

Right now `calcite-filter` matches every property of the item's object. It would be great if we could create a new property that specifies either properties to ignore or the properties to match. The property type `Array<string>` so it would need to be set in JavaScript. Something like

```js
filterElement.ignoreFields = ["OBJECTID", "address"]
```
or
```js
filterElement.matchFields = ["city", "state", "county"]
```

thoughts on this @jcfranco?

### Acceptance Criteria

A new property that either matches or ignores specific properties of filter's item objects.

### Which Component

calcite-filter

### Example Use Case

I was putting together a pagination sample for someone and noticed inconsistent filter results:
https://codepen.io/benelan/pen/qBoxdwv

I realized it was because I was adding a hex string `color` property for the card's thumbnail, which was also being matched. It would have been great to be able to do something like 

```js
filterElement.ignoreFields = ["color"]
```
or
```js
filterElement.matchFields = ["NAME", "CITY", "STATE"]
```

You can work around the constraint by passing a subset of the items' properties to `calcite-filter` and then matching the item back to the full dataset with an `id` or something that does get provided to filter. But it seems like a useful option for people anyway, and relatively low effort on our end to remove certain properties before filtering.

driskull:
I prefer having an allowList like `filterElement.matchFields = ["NAME", "CITY", "STATE"]` but that would be considered a breaking change since fields would need to be configured going forward if someone is using `calcite-filter` as a standalone component. We document filter so they could be.

Should this move to a breaking change release? Or do we want to go with a blockList instead?

jcfranco:
We could also consider the default to match all unless the user sets the allow list. This would preserve the existing behavior, allow users to ignore certain fields and doesn't require users to set anything by default.

driskull:
Sounds like a good plan

DitwanP:
🍡 Verified locally on `dev`

PR #9541, opened 2024-06-07
feat(filter): adds ability to match only specific filter data properties

driskull:
**Related Issue:** #5063

## Summary

- add matchFields argument to filter utility
  - allows filtering fields at the first level of an object.
  - Nested object fields will not be filtered.
    - Could be enhanced to do so in the future by enhancing `matchFields` to be an object
- add filter utility spec tests (thank you co-pilot)
- add `matchFields` property to `filter` component
- add e2e tests for filter

### DIFF
--- a/packages/calcite-components/src/components/filter/filter.tsx
+++ b/packages/calcite-components/src/components/filter/filter.tsx
@@ -82,6 +82,16 @@ export class Filter
   @Prop({ mutable: true }) filteredItems: object[] = [];
 
   /**
+   * Specifies the fields to match against when filtering.
+   */
+  @Prop() matchFields: string[];
+
+  @Watch("matchFields")
+  matchFieldsHandler(): void {
+    this.filterDebounced(this.value);
+  }
+
+  /**
    * Specifies placeholder text for the input element.
    */
   @Prop() placeholder: string;
@@ -159,7 +169,7 @@ export class Filter
   async componentWillLoad(): Promise<void> {
     setUpLoadableComponent(this);
     if (this.items.length) {
-      this.updateFiltered(filter(this.items, this.value));
+      this.updateFiltered(filter(this.items, this.value, this.matchFields));
     }
     await setUpMessages(this);
   }
@@ -223,7 +233,8 @@ export class Filter
 
   private filterDebounced = debounce(
     (value: string, emit = false, onFilter?: () => void): void =>
-      this.items.length && this.updateFiltered(filter(this.items, value), emit, onFilter),
+      this.items.length &&
+      this.updateFiltered(filter(this.items, value, this.matchFields), emit, onFilter),
     DEBOUNCE_TIMEOUT,
   );
 
--- a/packages/calcite-components/src/components/filter/filter.tsx
+++ b/packages/calcite-components/src/components/filter/filter.tsx
@@ -82,7 +82,7 @@ export class Filter
   @Prop({ mutable: true }) filteredItems: object[] = [];
 
   /**
-   * Specifies the fields to match against when filtering.
+   * Specifies the fields to match against when filtering. This will only apply when `value` is an object. If not set, all fields will be matched.
    */
   @Prop() matchFields: string[];
 
--- a/packages/calcite-components/src/components/list/list.e2e.ts
+++ b/packages/calcite-components/src/components/list/list.e2e.ts
@@ -370,7 +370,7 @@ describe("calcite-list", () => {
 
     const visibleItems = await page.findAll("calcite-list-item:not([filter-hidden])");
 
-    expect(visibleItems.map((item) => item.id)).toEqual(["label-match", "description-match", "value-match"]);
+    expect(visibleItems.map((item) => item.id)).toEqual(["label-match", "description-match"]);
   });
 
   it("should support shift click to select multiple items", async () => {
--- a/packages/calcite-components/src/components/list/list.e2e.ts
+++ b/packages/calcite-components/src/components/list/list.e2e.ts
@@ -365,57 +365,12 @@ describe("calcite-list", () => {
     const list = await page.find("calcite-list");
     await page.waitForTimeout(listDebounceTimeout);
 
-    expect(await list.getProperty("filteredItems")).toHaveLength(3);
-    expect(await list.getProperty("filteredData")).toHaveLength(3);
+    expect(await list.getProperty("filteredItems")).toHaveLength(2);
+    expect(await list.getProperty("filteredData")).toHaveLength(2);
 
     const visibleItems = await page.findAll("calcite-list-item:not([filter-hidden])");
 
     expect(visibleItems.map((item) => item.id)).toEqual(["label-match", "description-match", "value-match"]);
-  });
-
-  it("filters initially with matchFields", async () => {
-    const page = await newE2EPage();
-    await page.setContent(html`
-      <calcite-list filter-enabled filter-text="match">
-        <calcite-list-item
-          id="label-match"
-          label="match"
-          description="description-1"
-          value="value-1"
-        ></calcite-list-item>
-        <calcite-list-item
-          id="description-match"
-          label="label-2"
-          description="match"
-          value="value-1"
-        ></calcite-list-item>
-        <calcite-list-item
-          id="value-match"
-          label="label-3"
-          description="description-3"
-          value="match"
-        ></calcite-list-item>
-        <calcite-list-item
-          id="no-match"
-          label="label-4"
-          description="description-4"
-          value="value-4"
-        ></calcite-list-item>
-      </calcite-list>
-    `);
-
-    await page.waitForChanges();
-    const list = await page.find("calcite-list");
-    list.setProperty("matchFields", ["label"]);
-    await page.waitForChanges();
-    await page.waitForTimeout(listDebounceTimeout);
-
-    expect(await list.getProperty("filteredItems")).toHaveLength(1);
-    expect(await list.getProperty("filteredData")).toHaveLength(1);
-
-    const visibleItems = await page.findAll("calcite-list-item:not([filter-hidden])");
-
-    expect(visibleItems.map((item) => item.id)).toEqual(["label-match"]);
   });
 
   it("should support shift click to select multiple items", async () => {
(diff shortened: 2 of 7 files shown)


alisonailea (approved):

driskull:
@jcfranco do you want to review this one or is it good to go?

packages/calcite-components/src/components/filter/filter.tsx
```diff
@@ -81,6 +81,16 @@ export class Filter
    */
   @Prop({ mutable: true }) filteredItems: object[] = [];
 
+  /**
+   * Specifies the fields to match against when filtering.
```
jcfranco, line 85: Could you add a note about this only applying when `value` is an object and that all fields are matched if not set? +@geospatialem @DitwanP for an editorial look-see. 👀

packages/calcite-components/src/components/list/list.e2e.ts
```diff
@@ -365,12 +365,12 @@ describe("calcite-list", () => {
     const list = await page.find("calcite-list");
     await page.waitForTimeout(listDebounceTimeout);
 
-    expect(await list.getProperty("filteredItems")).toHaveLength(3);
-    expect(await list.getProperty("filteredData")).toHaveLength(3);
+    expect(await list.getProperty("filteredItems")).toHaveLength(2);
```
jcfranco, line 368: Should this have changed? Filtering should behave the same when `matchFields` are not specified, right?

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -558,6 +558,7 @@ export class List
                         aria-label={filterPlaceholder}
                         disabled={disabled}
                         items={dataForFilter}
+                        matchFields={["label", "description", "metadata"]}
```
jcfranco, line 561: I think this also needs to include `value`. Otherwise, we're changing existing filtering behavior.

packages/calcite-components/src/utils/filter.spec.ts
```diff
@@ -0,0 +1,66 @@
+import { filter } from "./filter";
+
+describe("filter function", () => {
+  it("warns and returns empty array for empty data", () => {
+    const originalWarn = console.warn;
+    console.warn = jest.fn();
```
jcfranco, line 6: Can you stub with jest instead of replacing it directly?

```tsx
jest.spyOn(console, "warn");
```

I think you'll also need to call `jest.restoreMocks` in an `afterEach` to restore the original `warn`.

packages/calcite-components/src/utils/filter.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    expect(result).toEqual([{ name: "John", tags: ["developer", "tester"] }]);
+  });
+
+  it("always includes objects with filterDisabled set to true", () => {
+    const data = [{ name: "John" }, { name: "Jane", filterDisabled: true }];
+    const result = filter(data, "Doe");
+    expect(result).toEqual([{ name: "Jane", filterDisabled: true }]);
+  });
+
+  it("always includes objects with constant set to true", () => {
```
jcfranco, line 61: Great test coverage! Thanks for adding this spec test! ✨🏆✨

packages/calcite-components/src/utils/filter.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+      { name: "Jane", age: 25 },
+    ];
+    const result = filter(data, "25", ["age"]);
+    expect(result).toEqual([{ name: "Jane", age: 25 }]);
+
+    const result2 = filter(data, "John", ["age"]);
+    expect(result2).toEqual([]);
+  });
+
+  it("ignores functions and null values in objects", () => {
```
jcfranco, line 37: Does searching with `"null"` return an empty array?

packages/calcite-components/src/utils/filter.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    expect(result).toEqual([{ name: "John", action: expect.any(Function), value: null }]);
+  });
+
+  it("searches nested objects correctly", () => {
+    const data = [{ name: "John", details: { age: 30 } }];
+    const result = filter(data, "30");
+    expect(result).toEqual([{ name: "John", details: { age: 30 } }]);
+  });
+
+  it("searches arrays in objects correctly", () => {
```
jcfranco, line 49: Suggestion: array and object searching could be merged into a single test.

packages/calcite-components/src/utils/filter.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    const data = [{ name: "John", tags: ["developer", "tester"] }];
+    const result = filter(data, "tester");
+    expect(result).toEqual([{ name: "John", tags: ["developer", "tester"] }]);
+  });
+
+  it("always includes objects with filterDisabled set to true", () => {
```
jcfranco, line 55: Suggestion: `filterDisabled` and `constant` searching could be merged into a single test.

packages/calcite-components/src/utils/filter.ts
```diff
@@ -9,16 +9,22 @@ export const filter = (data: Array<object>, value: string): Array<any> => {
...
-  const find = (input: object, RE: RegExp) => {
+  const find = (input: object, RE: RegExp, fields?: string[]) => {
     if ((input as any)?.constant || (input as any)?.filterDisabled) {
       return true;
     }
+
     let found = false;
 
-    forIn(input, (val) => {
+    forIn(input, (val: any, key) => {
```
jcfranco, line 19: Should `input` be typed as `object` and `key` be typed as `string`?

packages/calcite-components/src/utils/filter.ts
```diff
@@ -27,12 +33,9 @@ export const filter = (data: Array<object>, value: string): Array<any> => {
...
+
     return found;
   };
 
-  const result = data.filter((item) => {
-    return find(item, regex);
-  });
-
-  return result;
+  return data.filter((item) => find(item, regex, matchFields));
```
jcfranco, line 40: ✨🧹✨

jcfranco (commented):
I have one concern regarding the change in list filtering behavior, but otherwise, this looks great, @driskull!

packages/calcite-components/src/components/list/list.tsx
driskull: I think we should consider this a bug. It was never intended to search by value. Value is more of a unique identifier.

packages/calcite-components/src/components/filter/filter.tsx
driskull: @geospatialem @DitwanP can you comment on...

> Specifies the fields to match against when filtering. This will only apply when `value` is an object. If not set, all fields will be matched.

packages/calcite-components/src/utils/filter.ts
driskull: Yeah, will update. I set this type because it was showing as "never".

packages/calcite-components/src/utils/filter.spec.ts
driskull: Yep. added test.

packages/calcite-components/src/components/filter/filter.tsx
DitwanP: Looks good to me, clear and descriptive.

packages/calcite-components/src/utils/filter.ts
driskull: hmm I can't change this to object. I need to leave it as `any`.

packages/calcite-components/src/components/list/list.tsx
driskull: For filtering, we should only be searching content. I wouldn't consider `value` to be content. I think we should not search it by default for anyone using a list with a filter.

driskull: How about I move the list changes to a separate followup PR with the title `fix(list): only filter by list item content`? @jcfranco

jcfranco: > I think we should consider this a bug. It was never intended to search by value. Value is more of a unique identifier and not content like label, description and metadata are.

I disagree with this. The filtering has always included `value` props by default. Could we introduce a `matchFields` prop to all filtering components, allowing users to omit `value` if they prefer?

driskull: @jcfranco I reverted the list changes in this PR. Can we discuss it as part of #9569? I can create a separate issue for adding matchFields to list if we decide to do that.

 Is this PR good to merge?

jcfranco: Gotcha. Will give this another look.

jcfranco (approved):
🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️
🕵️🔍🕵️🕵️🕵️🔍🔍🔍🔍🕵️🔍🔍🔍🔍🔍🕵️🔍🕵️🕵️🕵️🔍🕵️🔍🕵️
🕵️🔍🕵️🕵️🕵️🔍🕵️🕵️🕵️🕵️🕵️🕵️🔍🕵️🕵️🕵️🔍🔍🕵️🔍🔍🕵️🔍🕵️
🕵️🔍🕵️🕵️🕵️🔍🕵️🔍🔍🕵️🕵️🕵️🔍🕵️🕵️🕵️🔍🕵️🔍🕵️🔍🕵️🔍🕵️
🕵️🔍🕵️🕵️🕵️🔍🕵️🕵️🔍🕵️🕵️🕵️🔍🕵️🕵️🕵️🔍🕵️🕵️🕵️🔍🕵️🕵️🕵️
🕵️🔍🔍🔍🕵️🔍🔍🔍🔍🕵️🕵️🕵️🔍🕵️🕵️🕵️🔍🕵️🕵️🕵️🔍🕵️🔍🕵️
🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️🕵️

Issue #9401, opened 2024-05-22
Combobox: When you tab into the input, the options do not appear


noahmulfinger:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Tabbing into the input causes the dropdown options to not appear, even when you start typing. This does not match the behavior after you click into the input.

### Expected Behavior

I would expect to be able to see the filtered results whether I click on the combobox or tab into it.

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/combobox/

### Reproduction Steps

1. Tab into the combobox input.
4. Start typing. You will not see the dropdown with matching options appear no matter what you type.
5. Once you click on the the combobox, the options will start appearing.

### Reproduction Version

2.8.2

### Relevant Info

Tested in Chrome and Firefox.

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [X] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Developer Experience

jcfranco:
We should definitely display any matched options when typing after tabbing into the component.

Not opening on focus is by design. This approach adheres to [WAI-ARIA combobox guidance](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/#keyboard_interaction) and is consistent with similar components (e.g., input-date-picker, input-time-picker). cc @geospatialem

DitwanP:
🥭 Verified locally on `dev`

PR #9543, opened 2024-06-08
fix(combobox): open the component when typing within it

driskull:
**Related Issue:** #9401

## Summary

- Toggles the combobox's open property when typing within the input
- If the value.length > 0, the combobox is opened otherwise its closed
- adds test

### DIFF
--- a/packages/calcite-components/src/components/combobox/combobox.e2e.ts
+++ b/packages/calcite-components/src/components/combobox/combobox.e2e.ts
@@ -178,46 +178,49 @@ describe("calcite-combobox", () => {
     openClose(simpleComboboxHTML);
   });
 
-  it("should open the combobox when typing within the input", async () => {
-    const page = await newE2EPage({
-      html: html`
-        <calcite-combobox id="myCombobox">
-          <calcite-combobox-item value="Raising Arizona" text-label="Raising Arizona"></calcite-combobox-item>
-          <calcite-combobox-item value="Miller's Crossing" text-label="Miller's Crossing"></calcite-combobox-item>
-          <calcite-combobox-item value="The Hudsucker Proxy" text-label="The Hudsucker Proxy"></calcite-combobox-item>
-          <calcite-combobox-item value="Inside Llewyn Davis" text-label="Inside Llewyn Davis"></calcite-combobox-item>
-        </calcite-combobox>
-      `,
-    });
+  it("should toggle the combobox when typing within the input", async () => {
+    const page = await newE2EPage();
+
+    await page.setContent(html`
+      <calcite-combobox id="myCombobox">
+        <calcite-combobox-item value="Raising Arizona" text-label="Raising Arizona"></calcite-combobox-item>
+        <calcite-combobox-item value="Miller's Crossing" text-label="Miller's Crossing"></calcite-combobox-item>
+        <calcite-combobox-item value="The Hudsucker Proxy" text-label="The Hudsucker Proxy"></calcite-combobox-item>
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/combobox/combobox.e2e.ts
```diff
@@ -178,6 +178,35 @@ describe("calcite-combobox", () => {
     openClose(simpleComboboxHTML);
   });
 
+  it("should open the combobox when typing within the input", async () => {
+    const page = await newE2EPage({
```
jcfranco, line 182: For consistency, can you create the page and then set content separately?

packages/calcite-components/src/components/combobox/combobox.e2e.ts
```diff
@@ -178,6 +178,35 @@ describe("calcite-combobox", () => {
...
+          <calcite-combobox-item value="Raising Arizona" text-label="Raising Arizona"></calcite-combobox-item>
+          <calcite-combobox-item value="Miller's Crossing" text-label="Miller's Crossing"></calcite-combobox-item>
+          <calcite-combobox-item value="The Hudsucker Proxy" text-label="The Hudsucker Proxy"></calcite-combobox-item>
+          <calcite-combobox-item value="Inside Llewyn Davis" text-label="Inside Llewyn Davis"></calcite-combobox-item>
+        </calcite-combobox>
+      `,
+    });
+
+    const combobox = await page.find("calcite-combobox");
+    const input = await page.find("calcite-combobox >>> input");
```
jcfranco, line 194: Can you update this to avoid using the internal input?

packages/calcite-components/src/components/combobox/combobox.e2e.ts
```diff
@@ -178,6 +178,35 @@ describe("calcite-combobox", () => {
...
+    const items = await page.findAll("calcite-combobox-item");
+    expect(await combobox.getProperty("open")).toBe(false);
+    await input.focus();
+    await page.waitForChanges();
+
+    await input.type("Arizona");
+    await page.waitForChanges();
+
+    expect(await combobox.getProperty("open")).toBe(true);
+    expect(await items[0].isVisible()).toBe(true);
```
jcfranco, line 204: Does this need to assert on the visible items? IMO, asserting on the combobox being open or closed should be good coverage.

packages/calcite-components/src/components/combobox/combobox.e2e.ts
```diff
@@ -178,6 +178,35 @@ describe("calcite-combobox", () => {
...
+    expect(await items[1].isVisible()).toBe(false);
+    expect(await items[2].isVisible()).toBe(false);
+    expect(await items[3].isVisible()).toBe(false);
```
jcfranco, line 207: Could you also clear the input and assert the combobox is closed? This might affect the current test name.

jcfranco (approved):
Other than a few comments, this LGTM!

packages/calcite-components/src/components/combobox/combobox.e2e.ts
driskull: Yes,

This is used in multiple places throughout the file. `const input = await page.find("calcite-combobox >>> input");` so it would likely need to be cleaned up elsewhere though.

packages/calcite-components/src/components/combobox/combobox.e2e.ts
driskull: will remove

packages/calcite-components/src/components/combobox/combobox.e2e.ts
driskull: will do.

Issue #9416, opened 2024-05-24
calcite-modal is being blocked by another element, but only in Safari


MikeTschudi:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

We have an instant app that displays a map of features in a left panel and a table of features in a right panel. If one selects multiple features and clicks the "Delete" button, we display a calcite-modal as a double-check before performing the deletion.  All is good on Windows (Chrome, Edge, Firefox) and on Mac Chrome. On Mac Safari, however, the calcite-modal is partially blocked.

Here is what the app modal looks like in all but Safari:
![image](https://github.com/Esri/calcite-design-system/assets/2125181/f7c68a41-77fe-4524-93b2-f94424d70339)
 
And here is what it looks like in Safari:
![image](https://github.com/Esri/calcite-design-system/assets/2125181/b9d1e9f5-ce55-4d34-984c-ddad39fe2dea)

The calcite-modal is triggered by a component that resides in an action-bar:
![image](https://github.com/Esri/calcite-design-system/assets/2125181/56eb3bd1-c74e-462f-849f-ce7df74179b0)

I can't tell if this is a bug in our app or in calcite: I am having trouble isolating the cause.

### Expected Behavior

I expected the modal to be on top, as it is outside of Safari

### Reproduction Sample

https://codepen.io/miketschudi/pen/GRaZwEd

### Reproduction Steps

In the CodePen, click on the Delete button to open the modal.

The modal is nested because we're using nested components; the modal is in the delete-button component; all other non-calcite-components have been replaced by divs.

### Reproduction Version

2.8.3

### Relevant Info

* Windows 10
  * OK: Chrome 125, Edge 125, Firefox 126
* Mac
  * OK: Chrome 125
  * not OK: Safari 17.5

### Regression?

went back to 1.0.3 and still see issue

### Priority impact

impact - p3 - not time sensitive

### Impact

Affects Safari users of the Manager Instant App; workaround is to delete items one-by-one via a different Delete button.

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Solutions

bcree11:
[...]

PR #9545, opened 2024-06-08
fix(modal)!: use fixed positioning on host to prevent Safari from clipping content in certain layouts

jcfranco:
**Related Issue:** #9416

## Summary

This restores `modal`'s host positioning to `fixed` to avoid a [Safari bug](https://bugs.webkit.org/show_bug.cgi?id=160953) that causes content to be clipped. 

**Note**: screenshot test was not added for this scenario since our testing environment uses Chrome.

BREAKING CHANGE: This should not require changes, but we are including this in the breaking change release due to the different modal configurations that could be impacted by the default `position` change.


packages/calcite-components/src/components/modal/modal.scss
```diff
@@ -407,16 +406,4 @@ slot[name="primary"] {
   }
 }
 
-/**
- * Conditional styles for when Modal is slotted in Shell
- */
-
-.container.slotted-in-shell {
```
driskull, line 414: Are these no longer needed? if so, what about the class for slotted-in-shell? We can remove it if not necessary. I think there is a prop as well.

jcfranco: I managed to eliminate the prop by dynamically checking if the overflowing container is the document body.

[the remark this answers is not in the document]
jcfranco: Note for future reference: `slottedInShell` was renamed to `embedded` in https://github.com/Esri/calcite-design-system/pull/9778.

packages/calcite-components/src/components/modal/modal.scss
```diff
@@ -18,7 +18,7 @@
 }
 
 :host {
-  @apply absolute flex inset-0 opacity-0 z-overlay;
+  @apply fixed flex inset-0 opacity-0 z-overlay;
```
driskull, line 21: Would we need to do this for the dialog component as well?

jcfranco: We're going to test drive this with `modal` first. If everything works out smoothly, we could apply to other components before 4.0 since it wouldn't be breaking (see https://github.com/Esri/calcite-design-system/pull/9545#discussion_r1635722740).

jcfranco:
@macandcheese @driskull Care to take another look at this one?

jcfranco:
Failing E2E test should be stabilized by https://github.com/Esri/calcite-design-system/pull/10748.

driskull (approved):
Code changes look good to me 👍

PR #9546, opened 2024-06-10
ci(prerelease): use absolute path to publish script

benelan:
Use an absolute path to the pre-release publishing script, because I'm not sure where relative paths start from.


benelan:
pushing this one through so `next` releases aren't broken, cc @jcfranco

Issue #9542, opened 2024-06-07
[calcite-split-button] Enhancement request to add a property to control dropdown placement


doreenbrinkman:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Dropdown is placed automatically. It would be helpful to control the placement of the dropdown in the `calcite-split-button`.

### Expected Behavior

Dropdown from `calcite-split-button` is able to be placed.

### Reproduction Sample

https://codepen.io/doreenbrinkman/pen/QWRMEje

### Reproduction Steps

1.  Run the sample which changes the placement to a more desired position
2. Comment out the JS script code and see position before

### Reproduction Version

2.8.3

### Relevant Info

Examples of when placement is needed.

No placement:
![image](https://github.com/Esri/calcite-design-system/assets/156362075/08ab5ad5-5964-48a2-878d-8da18255376e)

With placement:
![image](https://github.com/Esri/calcite-design-system/assets/156362075/85ee66e9-58dd-4b98-a56d-87ca4a49ee62)

### Priority impact

impact - p2 - want for an upcoming milestone

### Impact

We have workarounds to the issue (using the JS code) but it would be nice to use a new property to control.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [X] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Online

DitwanP:
🍡 Verified locally on `dev`

PR #9548, opened 2024-06-10
feat(split-button): add placement and flipPlacements property

driskull:
**Related Issue:** #9542

## Summary

- add placement and flipPlacements property
- add tests
- add story

### DIFF
--- a/packages/calcite-components/src/components/split-button/split-button.stories.ts
+++ b/packages/calcite-components/src/components/split-button/split-button.stories.ts
@@ -2,6 +2,7 @@ import { iconNames } from "../../../.storybook/helpers";
 import { boolean, modesDarkDefault } from "../../../.storybook/utils";
 import { html } from "../../../support/formatting";
 import { ATTRIBUTES } from "../../../.storybook/resources";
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/split-button/split-button.stories.ts
```diff
@@ -168,6 +176,31 @@ export const iconStartAndIconEnd = (): string => html`
   </div>
 `;
 
+export const placementTopStart = (): string => html`
```
jcfranco, line 179: Do you think we could replace this and similar tests by enhancing the [`floatingUI` common test helper](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/tests/commonTests/floatingUI.ts? If so, we can create a separate issue for this.

jcfranco (approved):
🎉

packages/calcite-components/src/components/split-button/split-button.stories.ts
driskull: yeah that makes sense as a separate issue. Can you create one?

Issue #8886, opened 2024-03-05
[Action Bar] Cannot set `text-enabled` on slotted Action when `expand-disabled` is set


macandcheese:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When `expand-disabled` is set on a containing Action Bar, I cannot set `text-enabled` on a slotted Action.

Additionally, when setting `overflow-actions-disabled`, and there are `actions-end` slotted actions, they are not presented horizontally. Maybe this is a pattern thing and they should not be expected within a group there.

### Expected Behavior

I'd expect to be able to individually set `text-enabled` on a slotted Action when `expand-disabled` is set on a parent. The use case for this is primarily for `layout="horizontal"`, where one or a few of the slotted Action are desired to have `text-enabled`.

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/yLrNmXa?editors=1000

### Reproduction Steps

1. Open Codepen
2. Notice the `text-enabled` properties set on slotted Actions are not respected
3. Notice the `overflow-actions-disabled` example "stacks" the slotted `actions-end` when they are in a group

### Reproduction Version

2.6.0

### Relevant Info

From @driskull "we just need to update the `toggleChildActionTextfunction` to consider `expandDisabled`"

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

driskull:
@macandcheese it looks like when a component is connected or when a new action is slotted, we are modifying that action to be text-enabled or not depending on the state of the parent component. 

I think we should change this but it would be a fix that users would need to be aware of and maybe modify their code to handle. For example, if a user adds an action in a `calcite-action-bar` that has `expanded=true` then they should make sure that the action has `text-enabled`.

The change would be to only modify `text-enabled` on child actions when the expanded property is changed. Not when a component is loaded, connected, or during a mutation.

cc @jcfranco for if we want to do this during a breaking change release.

driskull:
@geospatialem @DitwanP I think we should move this one to the breaking change release since it could require users to make changes to their code.

geospatialem:
Verified on `3.0.0-next.7`:

[...]

PR #9554, opened 2024-06-10
fix(action-bar, action-pad)!: Do not modify text-enabled on slotted actions unless expanded is toggled

driskull:
**Related Issue:** #8886

## Summary

- updates `action-bar` and `action-pad` to only change child actions `textEnabled` property when the component is expanded or collapsed.
- If the component is starting in an expanded or collapsed state, the user is responsible for properly setting `textEnabled` on children.
- If children are added, they need to be correctly assigned `textEnabled`
- cleanup: stores action groups for use within the component
- update tests

BREAKING CHANGE: When slotting actions within the component you must now set `textEnabled` on them if necessary.


jcfranco (approved):
LGTM!

Can you move the `BREAKING CHANGE` footer to the end of the description? It needs to be placed at the bottom to adhere to the [conventional commits spec](https://www.conventionalcommits.org/en/v1.0.0/#summary):

```
<type>[optional scope]: <description>

[optional body]

[optional footer(s)]
```

This can be merged after the Oct milestone wraps up.

Issue #9559, opened 2024-06-11
before open/close events emitted immediately after transition


macandcheese:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

On `dev` - the Modal has the incorrect animation: 

https://github.com/Esri/calcite-design-system/assets/4733155/730fa810-248c-46cd-8041-c0874d621e04

### Expected Behavior

The animation should behave as it does in 2.9

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/dyEVmam?editors=1000

### Reproduction Steps

1. Open Codepen
2. Open Model
3. See difference in animation

### Reproduction Version

Unreleased

### Priority impact

impact - p1 - need for current milestone

### Impact

cc @jcfranco 

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

jcfranco:
This was introduced by https://github.com/Esri/calcite-design-system/commit/bb45bdb24f70a397b2eeae02f90e719140423a92. It looks like it affects the `openClose` util and events will fire out of order. Bumped severity and marked as regression because of this.

DitwanP:
🍡 Verified locally on `dev`

PR #9341, opened 2024-05-14
refactor(dom): consolidate transition/animation waiting utils

jcfranco:
**Related Issue:** N/A

## Summary

Stems from https://github.com/Esri/calcite-design-system/pull/9325#issue-2294070147.

Consolidates transition/animation waiting logic and adds `whenTransitionDone` to DOM utils to complement `whenAnimationDone`.

PR #9560, opened 2024-06-11
fix: fix regression causing open/close events from emitting in proper order

jcfranco:
**Related Issue:** #9559

## Summary

This fixes `openCloseComponent.onToggleOpenCloseComponent` to emit beforeOpen/Close events when the associated transition starts and emits open/close when the transition ends. 

This regression was introduced by https://github.com/Esri/calcite-design-system/pull/9341 and would cause `beforeOpen`/`open` or `beforeClose`/`close` to emit immediately after another after the transition was done.

### Notable changes

* adds spec test to add coverage for `onToggleOpenCloseComponent`
* enhances `whenAnimationDone`/`whenTransitionDone` to accept callbacks for both start and end phases
* exposes `readTask` from `openCloseComponent` to allow for stubbing (stubbing Stencil's utility is not possible because the core module getters are non-configurable)
* extracts animation/transition helpers for spec tests to individual modules


packages/calcite-components/src/utils/openCloseComponent.ts
```diff
@@ -1,6 +1,11 @@
 import { readTask } from "@stencil/core";
 import { whenTransitionDone } from "./dom";
 
+/**
+ * Exported for testing purposes only
```
driskull, line 5: 🥼 🔬

packages/calcite-components/src/utils/dom.ts
```diff
@@ -702,8 +720,14 @@ export async function whenTransitionOrAnimationDone(
             so we fall back to it if there's no matching prop duration */
     allDurationsArray[0];
 
+  function startEndImmediately(): void {
```
driskull, line 723: 👍

driskull (approved):
Nice 👍

Issue #8297, opened 2023-11-29
Create automated test for positioning when a component has moved via drag and drop


driskull:
**monday.com sync:** #18048581181

### Test type

New test

### Which Component(s)

combobox, dropdown, input-date-picker, popover, tooltip

### Unstable Tests

N/A

### Test error, if applicable

N/A

### PR skipped, if applicable

#

### Additional Info

https://github.com/Esri/calcite-design-system/pull/8296#issuecomment-1832618658

Create an automated test that will verify a dropdown position is maintained when the component has been reordered via drag and drop and then scrolling occurs.

DintaMel:
Changing priority per @driskull.

PR #9573, opened 2024-06-13
fix(list): fix mobile device dragging with nested lists

driskull:
**Related Issue:** #9521

## Summary

- Makes sure that `connectSortableComponent` isn't called during an active drag of a sortable item.
  - Previously, mutation observers were calling `connectSortableComponent` during a drag
  - This caused the existing implementation to be destroyed and a new one created which left the user in a frozen state.
  - This only occurred on mobile devices because mobile devices do not use native drag and drop with Sortable.js
- The fix makes sure that `connectSortableComponent` and `disconnectSortableComponent` do nothing during an active drag.

### DIFF
--- a/packages/calcite-components/src/components/list/list.tsx
+++ b/packages/calcite-components/src/components/list/list.tsx
@@ -30,7 +30,6 @@ import {
   connectSortableComponent,
   disconnectSortableComponent,
   SortableComponent,
-  dragActive,
 } from "../../utils/sortableComponent";
 import { SLOTS as STACK_SLOTS } from "../stack/resources";
 import {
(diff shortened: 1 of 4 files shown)


jcfranco (approved):
Changes LGTM!

Would current tests cover these changes or is there mobile-specific behavior in `SortableJS` that can't be tested? Possibly related, are we waiting for https://github.com/Esri/calcite-design-system/issues/8297 to improve/add coverage for this?

driskull:
> Would current tests cover these changes or is there mobile-specific behavior in SortableJS that can't be tested? Possibly related, are we waiting for https://github.com/Esri/calcite-design-system/issues/8297 to improve/add coverage for this?

Yeah we can't currently test it because we need mobile testing of sortablejs. We would need to do some sort of browser identification spoofing alongside the automated test.

PR #9578, opened 2024-06-13
chore: change additional instances of main to dev

benelan:
## Summary

- Remove Chromatic from `main`, since it will already have ran on `dev`.
- Fix the scheduled update jobs, which where still creating PRs targeting
  `main`. Thankfully I left the 6 approvals rule because I definitely tried to
  merge those with `gh` :sweat_smile:
- Change the branch of code URLs from `main` to `dev`. Side note: maybe we
  could use the rollup plugin from #9536 to pin the version tag in the API
  reference code links. That way docs-json will point to the actual version's
  code instead of whatever is currently on `dev` or `main`.

### DIFF
--- a/.github/ISSUE_TEMPLATE/accessibility.yml
+++ b/.github/ISSUE_TEMPLATE/accessibility.yml
@@ -99,7 +99,7 @@ body:
     id: packages
     attributes:
       label: Calcite package
-      description: Select the relevant [package(s)](https://github.com/Esri/calcite-design-system/tree/main/packages) related to the request.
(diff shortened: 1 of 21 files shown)


driskull (approved):
Looks good @benelan 👍

jcfranco (approved):

jcfranco:
> Side note: maybe we
could use the rollup plugin from https://github.com/Esri/calcite-design-system/pull/9536 to pin the version tag in the API
reference code links. That way docs-json will point to the actual version's
code instead of whatever is currently on dev or main.

I dig it. Can you create a follow-up issue for this?

Issue #6905, opened 2023-05-02
[Block Section] Text hierarchy and spacing


asangma:
### Description

BlockSection heading is very similar to Block's heading.
<img width="443" alt="image" src="https://user-images.githubusercontent.com/12503298/235787709-be77745d-cec2-4f2c-b098-00c95c5d90df.png">

Suggest using normal font weight for BlockSection headings.
<img width="447" alt="image" src="https://user-images.githubusercontent.com/12503298/235787899-d45dcedc-9b30-4b49-a657-1e4ca3b9a86e.png">

Also, it seems that the padding in the content area of Block doesn't match the inline padding used in most other areas. Right now, it's at 0.625rem whereas most other places 0.75rem is used. Suggest updating to 0.75rem for better alignment.

<img width="1093" alt="Screen Shot 2023-05-02 at 2 17 34 PM" src="https://user-images.githubusercontent.com/12503298/235788381-c84c6272-4122-4e90-b46d-34226cc07153.png">

@SkyeSeitz would appreciate your feedback.

### Acceptance Criteria

Better text hierarchy and padding alignment.

### Which Component

Block and BlockSection

### Example Use Case

Using Block and BlockSection together.

### Priority impact

p4 - not time sensitive

### Esri team

N/A

## Final design recommendations
See [comment](https://github.com/Esri/calcite-design-system/issues/6905#issuecomment-1804189189)

SkyeSeitz:
I agree, I think better text hierarchy is something we could take a look at 👍🏽

Funnily enough, the Block content section padding in Figma lines up Block text and Block Section text at 0.75rem/12px already. So either Figma missed this change or we have just been slightly misaligned since the Figma component was built- thanks for pointing out, Alan! This fix would also nicely center align Block's expand chevron and Block Section's Switch. 

<img width="525" alt="image" src="https://user-images.githubusercontent.com/29716057/235808335-752247c5-8ffc-476d-95a9-88cb9c01a904.png">

asangma:
Thanks @SkyeSeitz.

@geospatialem This isn't urgent but the UtilityNetwork widget would be much happier if this change gets in soon. :)

ashetland:
[...]

PR #9580, opened 2024-06-13
fix(block-section): adjust text hierarchy

Elijbet:
**Related Issue:** #6905

## Summary

Adjust text hierarchy.

### DIFF
--- a/packages/calcite-components/src/components/block-section/block-section.scss
+++ b/packages/calcite-components/src/components/block-section/block-section.scss
@@ -14,6 +14,7 @@
     color: var(--calcite-color-text-1);
 
     &:hover {
+      color: var(--calcite-color-text-1);
     }
   }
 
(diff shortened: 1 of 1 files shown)


packages/calcite-components/src/components/block-section/block-section.scss
```diff
@@ -1,31 +1,44 @@
...
 :host([open]) {
   @apply border-b-color-3
     border-0
     border-b
     border-solid;
+
+  .toggle {
+    color: var(--calcite-color-text-1);
+
+    &:hover {
```
macandcheese, line 16: Can remove this if unused.

macandcheese (approved):

packages/calcite-components/src/components/block-section/block-section.scss
Elijbet: sorry that's used, it's a typo.

PR #9616, opened 2024-06-17
fix(list, list-item, list-item-group): add e2e themed tests

alisonailea:
**Related Issue:** #7180 

## Summary

Made sure components were UTD with /dev to reduce merge conflict churn
Updates component tokens
Update e2e themed tests

### DIFF
--- a/packages/calcite-components/src/components/list-item-group/list-item-group.scss
+++ b/packages/calcite-components/src/components/list-item-group/list-item-group.scss
@@ -3,44 +3,28 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
- * @prop --calcite-list-item-group-background-color: Specifies the background color of the component.
- * @prop --calcite-list-item-group-border-color: Specifies the border color of the component.
- * @prop --calcite-list-item-heading-text-color: Specifies the text color of the component's heading.
+ * @prop --calcite-list-item-group-background-color: Specifies the background color of the component. Will fallback to --calcite-list-background-color.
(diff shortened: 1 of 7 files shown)


packages/calcite-components/src/components/list-item-group/list-item-group.scss
```diff
@@ -3,44 +3,28 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
- * @prop --calcite-list-item-group-background-color: Specifies the background color of the component.
- * @prop --calcite-list-item-group-border-color: Specifies the border color of the component.
- * @prop --calcite-list-item-heading-text-color: Specifies the text color of the component's heading.
+ * @prop --calcite-list-item-group-background-color: Specifies the background color of the component. Will fallback to --calcite-list-background-color.
```
jcfranco, line 6: Do we need to list the fallback CSS Props in the doc? Also, it looks like the documented fallback props are not used.

jcfranco (approved):
Awesome!

I had a comment about documenting fallback CSS props in the doc and the current ones not being used. Not critical, but maybe we can hold off on this until https://github.com/Esri/calcite-design-system/pull/8594 is wrapped up.

packages/calcite-components/src/components/list-item-group/list-item-group.scss
alisonailea: We do not. Thanks for the catch!

Issue #9617, opened 2024-06-17
combobox: do not open when no filtered results are shown


driskull:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When typing in a combobox, it always opens when there is text

### Expected Behavior

when typing in a combobox, it only opens when there are filtered items to show

### Reproduction Sample

https://codepen.io/driskull/pen/mdYpZgq?editors=1000

### Reproduction Steps

1. open https://codepen.io/driskull/pen/mdYpZgq?editors=1000
2. type "xskdlsdl" or some random characters to guarantee no matches
3. notice how the combobox is open
4. The combobox should not be open.

### Reproduction Version

2.9.0

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

geospatialem:
Verified in `2.10.0-next.2` 🎉

PR #9618, opened 2024-06-17
fix(combobox): only open when text exists and filtered items are present

driskull:
**Related Issue:** #9617

## Summary

- if no filtered items are displayed, the combobox should not be opened
- adds test

### DIFF
--- a/packages/calcite-components/src/components/combobox/combobox.tsx
+++ b/packages/calcite-components/src/components/combobox/combobox.tsx
@@ -1053,8 +1053,7 @@ export class Combobox
   inputHandler = (event: Event): void => {
     const value = (event.target as HTMLInputElement).value;
     this.text = value;
-    this.filterItems(value);
-    this.open = value.length > 0;
+    this.filterItems(value, true);
     if (value) {
       this.activeChipIndex = -1;
     }
@@ -1071,7 +1070,7 @@ export class Combobox
         isGroup(item) ? label === item.label : value === item.value && label === item.textLabel,
       );
 
-    return debounce((text: string): void => {
+    return debounce((text: string, updateOpen = false): void => {
       const filteredData = filter(this.data, text);
       const itemsAndGroups = this.getItemsAndGroups();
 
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1259,6 +1263,7 @@ export class Combobox
       }
       this.updateItems();
       this.filterItems("");
```
jcfranco, line 1265: Can you use the new flag to open after filtering for the new value? Similar to [above](https://github.com/Esri/calcite-design-system/pull/9618/files#diff-9bc2dae4daf67ad95f668837f3083752245fbc22e0a58b1e5a043830e05af02cL1056-L1057)?

driskull: I don't think so. That one will only open when there is text present in the input and this one should open regardless.

jcfranco: Gotcha. Maybe the function argument can be renamed to distinguish this.

jcfranco (approved):
📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝
📝🔍📝📝📝🔍🔍🔍🔍📝🔍🔍🔍🔍🔍📝🔍📝📝📝🔍📝🔍📝
📝🔍📝📝📝🔍📝📝📝📝📝📝🔍📝📝📝🔍🔍📝🔍🔍📝🔍📝
📝🔍📝📝📝🔍📝🔍🔍📝📝📝🔍📝📝📝🔍📝🔍📝🔍📝🔍📝
📝🔍📝📝📝🔍📝📝🔍📝📝📝🔍📝📝📝🔍📝📝📝🔍📝📝📝
📝🔍🔍🔍📝🔍🔍🔍🔍📝📝📝🔍📝📝📝🔍📝📝📝🔍📝🔍📝
📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝

packages/calcite-components/src/components/combobox/combobox.tsx
driskull: Yup, will do. 👍

Issue #9620, opened 2024-06-17
[Carousel] User can enable `autoplay` via keyboard when `autoplay` is not set.


macandcheese:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Currently, keyboard `enter` or `space` interactions will play / pause the Carousel when the component is in focus - even if `autoplay` is not requested by the developer via property.

https://github.com/Esri/calcite-design-system/assets/4733155/871323a4-5e3f-4f58-95d6-07858a220f7e

### Expected Behavior

The user should only be able to enable `autoplay` via keyboard when it is enabled via property.

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/xxNpvqq?editors=1000

### Reproduction Steps

1. Open Codepen
2. Click the onboarding button
3. The Carousel will focus on Popover open - use the space or enter to begin autoplay.
4. Notice autoplay has begun - when it should not have.

### Reproduction Version

2.9

### Priority impact

impact - p1 - need for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

DitwanP:
🍡 Verified locally on `dev`

PR #9621, opened 2024-06-17
fix(carousel): Prevent unexpected keyboard `autoplay` activation

macandcheese:
**Related Issue:** #9620 

## Summary
Adds a check for `autoplay` in our keyboard interaction handler. Adds a test to confirm behavior.


driskull (approved):
👍

Issue #9619, opened 2024-06-17
list: add matchFields property to specify which fields should be filtered against


driskull:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Add a matchFields property to specify which properties on list to use as matching values.

### Acceptance Criteria

new property matchFields

### Relevant Info

https://github.com/Esri/calcite-design-system/issues/5063

### Which Component

list

### Example Use Case

```ts
const list = document.querySelector("calcite-list");
list.matchFields = ["label", "description"];
```

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

geospatialem:
Verified in `2.10.0-next.7` with the following:

```html
<!-- HTML -->
<calcite-list filter-enabled filter-placeholder="Filter waypoints">
  <calcite-list-item label="Hiking trails" description="Designated routes for hikers to use." value="hiking-trails">
  </calcite-list-item>
  <calcite-list-item label="Waterfalls" description="Vertical drops from a river." value="waterfalls">
  </calcite-list-item>
  <calcite-list-item label="Rivers" description="Large naturally flowing watercourses." value="rivers">
  </calcite-list-item>
  <calcite-list-item label="Estuaries" description="Where the river meets the sea." value="estuaries">
  </calcite-list-item>
</calcite-list>
```

```js
// JavaScript
const listEl = document.querySelector("calcite-list")
listEl.filterProps = ["label"];
```

PR #9622, opened 2024-06-17
feat(list): add filterProps property to specify which properties to filter against

driskull:
**Related Issue:** #9619

## Summary

- add filterProps property to list
- changes `matchFields` to `filterProps` on the filter component.
- add tests

### DIFF
--- a/packages/calcite-components/src/components/list/list.e2e.ts
+++ b/packages/calcite-components/src/components/list/list.e2e.ts
@@ -66,6 +66,10 @@ describe("calcite-list", () => {
         propertyName: "dragEnabled",
         defaultValue: false,
       },
+      {
+        propertyName: "matchFields",
+        defaultValue: undefined,
+      },
     ]);
   });
 
@@ -371,6 +375,51 @@ describe("calcite-list", () => {
     const visibleItems = await page.findAll("calcite-list-item:not([filter-hidden])");
 
     expect(visibleItems.map((item) => item.id)).toEqual(["label-match", "description-match", "value-match"]);
+  });
+
+  it("filters initially with matchFields", async () => {
+    const page = await newE2EPage();
+    await page.setContent(html`
+      <calcite-list filter-enabled filter-text="match">
+        <calcite-list-item
+          id="label-match"
+          label="match"
+          description="description-1"
+          value="value-1"
+        ></calcite-list-item>
+        <calcite-list-item
+          id="description-match"
+          label="label-2"
+          description="match"
+          value="value-1"
+        ></calcite-list-item>
+        <calcite-list-item
+          id="value-match"
+          label="label-3"
+          description="description-3"
+          value="match"
+        ></calcite-list-item>
+        <calcite-list-item
+          id="no-match"
+          label="label-4"
+          description="description-4"
+          value="value-4"
+        ></calcite-list-item>
+      </calcite-list>
+    `);
+
+    await page.waitForChanges();
+    const list = await page.find("calcite-list");
+    list.setProperty("matchFields", ["label", "description"]);
+    await page.waitForChanges();
+    await page.waitForTimeout(listDebounceTimeout);
+
+    expect(await list.getProperty("filteredItems")).toHaveLength(2);
+    expect(await list.getProperty("filteredData")).toHaveLength(2);
+
+    const visibleItems = await page.findAll("calcite-list-item:not([filter-hidden])");
+
+    expect(visibleItems.map((item) => item.id)).toEqual(["label-match", "description-match"]);
   });
 
   it("should support shift click to select multiple items", async () => {
--- a/packages/calcite-components/src/components/list/list.e2e.ts
+++ b/packages/calcite-components/src/components/list/list.e2e.ts
@@ -67,7 +67,7 @@ describe("calcite-list", () => {
         defaultValue: false,
       },
       {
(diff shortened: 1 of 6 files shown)


packages/calcite-components/src/components/list/list.e2e.ts
```diff
@@ -373,6 +377,51 @@ describe("calcite-list", () => {
...
+          value="value-4"
+        ></calcite-list-item>
+      </calcite-list>
+    `);
+
+    await page.waitForChanges();
+    const list = await page.find("calcite-list");
+    list.setProperty("matchFields", ["label", "description"]);
+    await page.waitForChanges();
+    await page.waitForTimeout(listDebounceTimeout);
```
jcfranco, line 415: Is there an event we can listen for? This is fine for now, but we should start moving away from timeouts as [`puppeteer@22.2.0` removed `page.waitForTimeout`](https://github.com/puppeteer/puppeteer/blob/7966dd726288bda48650a3604f8e12c6e4be41db/packages/puppeteer-core/CHANGELOG.md#2200-2024-02-05).

packages/calcite-components/src/components/list/list.e2e.ts
```diff
@@ -373,6 +377,51 @@ describe("calcite-list", () => {
     expect(visibleItems.map((item) => item.id)).toEqual(["label-match", "description-match", "value-match"]);
   });
 
+  it("filters initially with matchFields", async () => {
```
jcfranco, line 380: Do we need a test for user-interaction filtering? Under the hood, both work with `filterText`, so probably not?

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -160,6 +160,16 @@ export class List
    */
   @Prop({ reflect: true }) loading = false;
 
+  /**
+   * Specifies the fields to match against when filtering. If not set, all fields will be matched.
```
jcfranco, line 164: Can we add a note about the default fields being used for filtering (`label`, `description`, `metadata`, `value`)? cc @geospatialem

jcfranco (approved):
📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝
📝🏑📝📝📝🏑📝📝🏑🏑📝📝🏑🏑🏑📝📝🏑🏑🏑📝🏑🏑🏑🏑📝🏑📝
📝🏑🏑📝📝🏑📝🏑📝📝🏑📝📝🏑📝📝🏑📝📝📝📝🏑📝📝📝📝🏑📝
📝🏑📝🏑📝🏑📝🏑📝📝🏑📝📝🏑📝📝🏑📝📝📝📝🏑🏑🏑📝📝🏑📝
📝🏑📝📝🏑🏑📝🏑📝📝🏑📝📝🏑📝📝🏑📝📝📝📝🏑📝📝📝📝📝📝
📝🏑📝📝📝🏑📝📝🏑🏑📝📝🏑🏑🏑📝📝🏑🏑🏑📝🏑🏑🏑🏑📝🏑📝
📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝📝

packages/calcite-components/src/components/list/list.e2e.ts
driskull: Not currently, we would need some kind of event for when filteredItems is updated. Can you create an issue to add this so we can clean up these tests?

packages/calcite-components/src/components/list/list.e2e.ts
driskull: Probably not necessary.

packages/calcite-components/src/components/list/list.tsx
driskull: Yep. will do

packages/calcite-components/src/components/list/list.e2e.ts
jcfranco: 🔥https://github.com/Esri/calcite-design-system/issues/9661🔥

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -160,6 +160,16 @@ export class List
    */
   @Prop({ reflect: true }) loading = false;
 
+  /**
+   * Specifies the fields to match against when filtering. If not set, all fields will be matched (label, description, metadata, value).
```
macandcheese, line 164: Sorry, late to the party here - is "fields" an obvious term here for this? "If not set, all fields will be matched (label, description, metadata, value)"

Really it's matching supplied property values? Feel free to disregard but "fields" doesn't immediately convey what will be matched to me.

driskull: We already called the property `matchFields` on `filter.tsx` but we can change before we release if we have a beter name. `matchPropertyValues` or `matchKeyValues`? I think its really a key. Could also be `matchKeys`?

@jcfranco

driskull: I kinda like `filterProps`. // the properties to use for filtering? @macandcheese? @jcfranco

driskull: We will likely need to update the changelog at release to modify matchFields to filterProps on the filter component.

PR #9630, opened 2024-06-18
ci(next): install deps before release scripts execute

benelan:
**Related Issue:** #9578

## Summary

Resolve [an error](https://github.com/Esri/calcite-design-system/actions/runs/9563619774/job/26362509309#step:4:16) in the prerelease deployment script where an npm package is used before it is installed.


benelan:
Pushing this through to fix `next` releases, cc @jcfranco

PR #9636, opened 2024-06-19
chore(handle): avoid mutable warnings from t9n props

jcfranco:
**Related Issue:** N/A

## Summary

Marks `t9n` internal props as mutable to avoid warning debug messages. Doc was also updated for consistency.


driskull (approved):
👍 💯

Issue #9644, opened 2024-06-20
Themed Chromatic/Demo Test Helper


alisonailea:
### Priority impact

p - medium

### Summary

1. Should accept an html tagged template literal of the component structure
1. Should accept a list of component tokens
1. Should iterate over the list of component tokens
    1. identify the prop for the token
        * text color
        * background color
        * border color
        * icon color
        * box shadow
        * corner radius
        * z index
        * space
    1. apply test theme value to token
1. apply all test theme tokens as CSS variables to the parent element in storybook or html demo page

### Desired Outcome

Speed up development of visual testing of design tokens on the web platform

### Resources

#7180

PR #9645, opened 2024-06-20
chore(demo-theme): themed demo helper

alisonailea:
**Related Issue:** #9644

## Summary

themed demo helper

### DIFF
--- a/packages/calcite-components/src/demos/_assets/demo-theme.ts
+++ b/packages/calcite-components/src/demos/_assets/demo-theme.ts
@@ -1,3 +1,22 @@
+/*
+ * EXAMPLE USAGE:
+ *
+ * <demo-theme tokens="
+ *     --calcite-button-background-color,
+ *     --calcite-button-border-color,
+ *     --calcite-button-corner-radius,
+ *     --calcite-button-corner-radius-start-start,
+ *     --calcite-button-corner-radius-start-end,
+ *     --calcite-button-corner-radius-end-start,
+ *     --calcite-button-corner-radius-end-end,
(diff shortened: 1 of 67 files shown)


jcfranco (approved):
Awesome! 

Approving to get this in for today's testing, but I do have some notes:

* Can we add doc to the theme helper with setup examples?
* Are the updates to the button page temporary or something that we'd like to promote across all component pages? We could also consider adding a specific theme demo page. 
* Can we add a note on current limitations of the helper? IIRC, stateful CSS props are not supported.

packages/calcite-components/src/demos/_assets/demo-theme.ts
```diff
@@ -0,0 +1,87 @@
+const defaultTheme = {
+  background: "blue",
+  cornerRadius: "42px",
+  shadowColor: "black",
+  shadowValues: "0 1px 2px 0 1",
+  space: "24px",
+  text: "green",
```
driskull, line 7: should this maybe be `textColor`?

driskull (approved):
👍

alisonailea:
> * Can we add doc to the theme helper with setup examples?
Added

> * Are the updates to the button page temporary or something that we'd like to promote across all component pages? We could also consider adding a specific theme demo page.

Not intended to be temporary. The idea was to allow these demos to be added to any page although long-term it might be good to have a specific theme demo page with all/many components.

> * Can we add a note on current limitations of the helper? IIRC, stateful CSS props are not supported.
There aren't any limitations as far as I know. Stateful CSS props are supported.

PR #9646, opened 2024-06-20
fix(table): update tokens and add E2E themed tests

alisonailea:
**Related Issue:** #7180

## Summary

Add tokens for table

### DIFF
--- a/packages/calcite-components/src/components/table-cell/table-cell.scss
+++ b/packages/calcite-components/src/components/table-cell/table-cell.scss
@@ -4,7 +4,7 @@
  * These properties can be overridden using the component's tag as selector.
  *
  * @prop --calcite-table-cell-background: [Deprecated] Use `--calcite-table-cell-background-color` instead. Specifies the background color of the component.
- * @prop --calcite-table-cell-background-color: Specifies the background color of the component.
+ * @prop --calcite-table-cell-background-color: Specifies the background color of the component. Will fallback to `--calcite-table-row-background-color` if not set.
  * @prop --calcite-table-cell-border-color: Specifies the border color of the component.
  */
 
@@ -34,12 +34,15 @@
 td {
   @apply text-start align-middle text-color-1 whitespace-normal;
   background-color: var(
-    --calcite-internal-table-cell-background-color,
-    var(--calcite-table-cell-background-color, transparent)
+    --calcite-table-cell-background-color,
+    var(--calcite-table-row-background-color, var(--calcite-internal-table-cell-background-color, transparent))
   );
   font-size: var(--calcite-internal-table-cell-font-size);
   border-inline-end: 1px solid
-    var(--calcite-table-cell-border-color, var(--calcite-table-border-color, var(--calcite-color-border-3)));
+    var(
+      --calcite-table-cell-border-color,
+      var(--calcite-table-row-border-color, var(--calcite-table-border-color, var(--calcite-color-border-3)))
+    );
   padding: var(--calcite-internal-table-cell-padding);
 
   &:not(.static-cell) {
--- a/packages/calcite-components/src/components/table-cell/table-cell.scss
+++ b/packages/calcite-components/src/components/table-cell/table-cell.scss
@@ -4,13 +4,13 @@
  * These properties can be overridden using the component's tag as selector.
  *
  * @prop --calcite-table-cell-background: [Deprecated] Use `--calcite-table-cell-background-color` instead. Specifies the background color of the component.
- * @prop --calcite-table-cell-background-color: Specifies the background color of the component. Will fallback to `--calcite-table-row-background-color` if not set.
+ * @prop --calcite-table-cell-background-color: Specifies the background color of the component.
+ * @prop --calcite-table-cell-background-color-selected: Specifies the background color of the selected cell.
  * @prop --calcite-table-cell-border-color: Specifies the border color of the component.
  * @prop --calcite-table-cell-text-color: Specifies the text color of the component.
  * @prop --calcite-table-cell-number-background-color: Specifies the background color of the number cell. 
  * @prop --calcite-table-cell-footer-background-color: Specifies the background color of the footer cell.
  * @prop --calcite-table-cell-text-color-selected: Specifies the text color of the selected cell.
- * @prop --calcite-table-cell-shadow: Specifies the shadow of the component.
  * @prop --calcite-table-cell-shadow-selected: Specifies the shadow of the selected cell.
  */
 
@@ -30,7 +30,6 @@
  */
 
 :host {
-  --calcite-internal-table-cell-background: var(--calcite-table-cell-background, transparent);
   @apply contents;
 }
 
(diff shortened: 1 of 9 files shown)


packages/calcite-components/src/components/table-row/table-row.scss
```diff
@@ -5,11 +5,9 @@
  *
  * @prop --calcite-table-row-background: [Deprecated] Use `--calcite-table-row-background-color` instead. Specifies the background color of the component.
  * @prop --calcite-table-row-background-color: Specifies the background color of the component.
- * @prop --calcite-table-row-background-color-striped: The background color of the component's `striped` rows, when specified.
+ * @prop --calcite-table-row-secondary-background-color: The background color of the component's `striped` rows, when specified.
```
alisonailea, line 8: I discussed this token name change with Skye and Aaron. The use of the "secondary" group here aligns better with our naming schema.

packages/calcite-components/src/components/table-row/table-row.scss
```diff
@@ -5,11 +5,9 @@
...
  * @prop --calcite-table-row-background-color-selected: The background color of the component's `selected` rows, when specified.
  * @prop --calcite-table-row-border-color: Specifies the border color of the component.
- * @prop --calcite-table-row-selected-accent-color: Specifies the border color of the component.
```
alisonailea, line 11: Unused

packages/calcite-components/src/components/table-row/table-row.scss
macandcheese: Hm. This seems unexpected to me. The name corresponds directly to the property that causes it to display. We don’t really use secondary elsewhere.

macandcheese: “striped-background-color”?

packages/calcite-components/src/components/table-cell/table-cell.scss
```diff
@@ -62,69 +82,61 @@ td.last-cell {
...
     --calcite-table-cell-background-color-selected,
-    var(--calcite-table-row-background-color-selected, var(--calcite-color-foreground-current))
+    var(--calcite-color-foreground-current)
   );
 }
 
 .selection-cell.selected-cell {
-  box-shadow: inset 0.25rem 0 0 0 var(--calcite-table-row-selected-accent-color, var(--calcite-color-brand));
-  color: var(--calcite-table-selection-cell-icon-color-selected, var(--calcite-color-brand));
+  box-shadow: var(--calcite-table-cell-shadow-selected, inset 0.25rem 0 0 0 var(--calcite-color-brand));
```
macandcheese, line 131: I wonder if the accent terminology is more understandable to an end user. It’s really creating a border accent - like we have in alert, notice, etc. this terminology with shadow makes it seem like the shadow is customizable.

packages/calcite-components/src/components/table-row/table-row.scss
macandcheese: Can we keep the accent color terminology vs. moving to shadow? The fact the accent border is implemented via a box shadow shouldn’t matter to a user IMO. It’s really changing the accent color.

packages/calcite-components/src/components/table-cell/table-cell.scss
macandcheese: It’s also only applicable to the single selection cell that receives the accent so I think the prior naming makes more sense. It should only be set on the row once, and not on an individual component. The cell that receives this color override is rendered internally within the Table Row.

alisonailea: thanks for the catch! But I think naming wise it aligns more with the "indicator" pattern that exists in other components.

macandcheese: Yeah, I know Alert uses "accent" for a similar thing, I think we have "highlight" in use (or talked about it). I think as long as we don't use shadow here I'm ok with it. 

Really in plain terms this is "the color of the icon in the selectable checkbox cell when that cell is selected". Because the user can't target this Cell directly, I think using terminology that has Table Row makes sense, since it can be controlled once at the app level or still individually per row. 

"--calcite-table-row-indicator-selected-color", "--calcite-table-row-selected-highlight/indicator", etc., would maybe solve both our concerns?

packages/calcite-components/src/components/table-row/table-row.scss
alisonailea: primary / secondary is a common naming pattern for design system colors and one I think we'll be expanding on in the future. "striped" is very specific to table.

macandcheese: Yeah, I guess that's my take - it is specific to table. To this point we've been intentional to *not* use primary / secondary / tertiary, so teams can build their own meaning through our properties. I think that's a fine direction to head in if we choose to in the future, but beginning to add it in now seems strange - this css property can only be used when `striped` is used so I think it makes sense. Adding "secondary" verbiage as the only public-facing case of it here seems strange. Maybe a good discussion with the larger team.

macandcheese: does "--calcite-table-row-striped-secondary-background-color" work? That way at least there is some indication to a user it pertains and will only be visible when the `striped` property is in use.

packages/calcite-components/src/components/table-cell/table-cell.scss
```diff
@@ -128,22 +136,15 @@ td.last-cell {
 }
 
 .selection-cell.selected-cell {
-  box-shadow: var(--calcite-table-cell-shadow-selected, inset 0.25rem 0 0 0 var(--calcite-color-brand));
-  color: var(--calcite-table-cell-text-color-selected, var(--calcite-color-brand));
-  & calcite-icon {
-    --calcite-icon-color: var(--calcite-table-cell-text-color-selected, var(--calcite-color-brand));
+  box-shadow: inset 0.25rem 0 0 0 var(--calcite-table-cell-indicator-color-selected, var(--calcite-color-brand));
```
macandcheese, line 139: This doesn't address my issue with this naming change - this is really a row-specific visual display and I think we should keep that nomenclature. I still think it should use table-row terminology, and that "--calcite-table-row-selected-accent-color" is a more apt name. "selected-highlight, "selected-indicator") all work for me, as long as its scoped to the row.

macandcheese (changes requested):
I made some comments about nomenclature, etc., but I think there are some other changes that caused issues - namely moving the row styles around.
Screenshot 2024-06-21 at 2 13 48 PM

I'm seeing double border in most places and some un-set icon color for selection, etc.

It'd also be nice to update the `table.html` demo page themed examples with the new properties to ensure we aren't missing any of the combinations. Thank you for tackling this!

packages/calcite-components/src/components/table-row/table-row.scss
SkyeSeitz: I would agree with this. If the default row color is "--calcite-table-row-background-color", then I would expect the striped prop to correlate to a "--calcite-table-row-striped-background-color" css property. I don't think the word "secondary" is necessary especially since there is no need for a "--calcite-table-row-primary-background-color" for it to correlate to.

jcfranco:
Closing as we are tackling component tokens in smaller chunks. Thanks, everyone! ✨💪✨

PR #9651, opened 2024-06-20
build(deps): downgrade conventional commits to fix next

benelan:
**Related Issue:** #9628

## Summary

A [pre-release error](https://github.com/Esri/calcite-design-system/actions/runs/9602677524/job/26484030057#step:4:98) occurred because lerna uses conventional-changelog's commonjs modules, which were removed in v8. Additionally, the pre-release script wasn't exiting on error.

PR #9652, opened 2024-06-20
ci(chromatic): fix build errors from corrupted package-lock

benelan:
## Summary

Resolve `package-lock.json` issue that is causing Chromatic failures on `dev`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -110,9 +110,9 @@
       }
     },
     "node_modules/@adobe/css-tools": {
-      "version": "4.4.0",
-      "resolved": "https://registry.npmjs.org/@adobe/css-tools/-/css-tools-4.4.0.tgz",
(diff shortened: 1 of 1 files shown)


driskull (approved):
👍 

Could we have some sort of eslint warning when there are changes to package but not package-lock for the future?

jcfranco (approved):

jcfranco:
@driskull Unfortunately, it's still possible to have `package-lock.json` issues even if it's always updated along with `package.json`.

benelan:
> @driskull Unfortunately, it's still possible to have `package-lock.json` issues even if it's always updated along with `package.json`.

Yeah, I'm pretty sure this popped up from #9651. I didn't run Chromatic on the PR since it didn't seem necessary, so that's on me. It's unusual to see build errors for Chromatic but not the E2E tests.

PR #9653, opened 2024-06-20
fix(list-item): Improve focus outline appearance

josercarcamo:
**Related Issue:** #7538 

## Summary
Make list item highlight display correctly over action items.


packages/calcite-components/src/components/list-item/list-item.scss
```diff
@@ -14,6 +14,16 @@
 
 @include disabled();
 
+.indent {
```
driskull, line 17: indent isn't a valid class on the list-item. Can this part be reverted? It should be on `.nested-container`. I'm guessing this was copied from a previous release.

packages/calcite-components/src/components/list-item/list-item.scss
```diff
@@ -52,16 +62,11 @@
...
-    border-0
-    border-color-3;
-
-  margin-inline-start: var(--calcite-list-item-spacing-indent, theme("spacing.6"));
+  @apply flex flex-col;
 }
 
-.nested-container--open {
-  @apply flex;
+.nested-container--hidden {
```
driskull, line 68: This should be reverted as well.

driskull (changes requested):
There are some changes made that are unrelated and would cause regressions here.

packages/calcite-components/src/components/list-item/list-item.scss
```diff
@@ -206,6 +211,40 @@ td:focus {
   }
 }
 
+tr:focus {
+  .actions-start,
+  .actions-end {
+    inset-block: theme("spacing[0.5]");
```
driskull, line 217: note: At some point these theme() vars will need to be updated to use calcite css vars.

packages/calcite-components/src/components/list-item/list-item.scss
```diff
@@ -206,6 +211,40 @@ td:focus {
...
+    }
+  }
+
+  &::after,
+  &::before {
+    @apply absolute;
+    content: "";
+    inline-size: theme("spacing[0]");
+    z-index: theme("zIndex.header");
+    background-color: theme("colors.brand");
```
driskull, line 233: optional: Might be able to use a css var here for brand.

packages/calcite-components/src/components/list-item/list-item.scss
```diff
@@ -206,6 +211,40 @@ td:focus {
...
+      block-size: calc(100% - theme("spacing[1]"));
+    }
+  }
+
+  &::after,
+  &::before {
+    @apply absolute;
+    content: "";
+    inline-size: theme("spacing[0]");
+    z-index: theme("zIndex.header");
```
driskull, line 232: optional: Might be able to use a css var here for z-index.

driskull (approved):
Code changes look good. I guess since its a focus state our screenshot tests can't really test this. 👍

Issue #7212, opened 2023-06-22
[Combobox] Make filter text value accessible in the code


shellchulx:
### Description

The document has defined an event `calciteComboboxFilterChange` that is fired when text is added to filter the options list, but looks like the filter text is not accessible in the code. 

The only way is through the `shadowRoot.querySelect` but it is not the best practice
`const text = this.comboboxElement.shadowRoot.querySelector<HTMLCalciteInputElement>("input").value;`

https://github.com/Esri/calcite-components/tree/master/packages/calcite-components/src/components/combobox#events

cc @BlakeStearman

### Acceptance Criteria

There is a prop for the filtered text in combobox

### Which Component

calcite-combobox

### Example Use Case

There's no way to return what the filter text currently is from a calcite-combobox? For example the gis in the screenshot.

![image](https://github.com/Esri/calcite-components/assets/14854017/45c9a212-f111-4511-9312-b15be686a25c)

### Priority impact

p4 - not time sensitive

### Esri team

ArcGIS Enterprise

geospatialem:
This also surfaced in a [Community post](https://community.esri.com/t5/calcite-design-system-questions/get-combobox-filter-value/m-p/1303440#M433) on 6/26/2023.

DitwanP:
🍬 Verified locally on `dev`
![image](https://github.com/Esri/calcite-design-system/assets/60022782/02ac282d-56f3-4a02-96ab-20ed4d2cc7af)

PR #9654, opened 2024-06-21
feat(combobox): add `filterText` prop

jcfranco:
*Related Issue:** #7212

## Summary

Adds `filterText` to allow dynamic access to filter text.


driskull (approved):
Nice!

We should probably add matchFields to combobox at some point as well.

PR #9676, opened 2024-06-25
test(input-time-zone): tidy up Intl API polyfill/override

jcfranco:
**Related Issue:** N/A

## Summary

The latest Puppeteer version includes Chromium 119, so we no longer need to provide most of what the test polyfill included. We still override supported time zones for increased coverage and to ease some test configurations.


benelan (approved):
:+1:

driskull (approved):
👍

PR #9681, opened 2024-06-25
fix(flow-item): fix inverted footer start/end slots

jcfranco:
**Related Issue:** #8981

## Summary

Fixes DOM order of flow-item footer start/end slots.

Note: `flow-item` and `panel` stories were updated to make footer slot content easier to review.


benelan (approved):
:broom:

PR #9688, opened 2024-06-25
fix(block-section): apply missing CSS class to start/end icon

jcfranco:
**Related Issue:** #9703

## Summary

Updates icon rendering to use class instead of value for styling.


driskull (approved):
👍 Does this trigger a screenshot test change? or is there some other way to test this?

jcfranco:
> 👍 Does this trigger a screenshot test change? or is there some other way to test this?

Good question. Indeed [it does](https://www.chromatic.com/test?appId=6266d45509d7eb004aa254fb&id=667b47baf9338102b3c51f3d)! ✋👁️👄👁️🔎❗

PR #9689, opened 2024-06-25
refactor(block): use icon value as key instead of class for consistency

jcfranco:
**Related Issue:** N/A

## Summary

✨🧹✨


driskull (approved):
🔑 👍

Issue #9624, opened 2024-06-17
Color picker: component enhancements


aPreciado88:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

-Auto select values when clicking on input fields. This should apply to all inputs in `color-picker`.
-Improved value commit:
1. The hex input should auto-commit a valid 6-char (or 8-char if alpha-channel is enabled).
2. A 3-char (or 4-char with alpha-channel) shorthand value would commit onBlur.
3. Channel inputs should auto-commit valid values. If the user deletes the value completely, there would be no change. If left blank, the original value should be restored onBlur.

-onBlur should include clicking anywhere outside the input. As of v1.8.0, the user must click outside the entire component.

### Which Component

`color-picker`

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

[Figma file](https://www.figma.com/design/h0vqgnnTvcDhrsJLyiQyOm/Color-Picker%3A-apply-entered-hex-value-immediately-[7057]?node-id=566-7568&t=e8AcRjJ3e3rTEPNW-0)

DitwanP:
🍡Verified on `2.11.0-next.15`

PR #9701, opened 2024-06-26
feat(color-picker, color-picker-hex-input): Add input auto commit, blur and auto select enhancements. 

aPreciado88:
**Related Issue:** [#9624](https://github.com/Esri/calcite-design-system/issues/9624)

## Summary

Updates `color-picker` and `color-picker-hex-input` to:

-Auto select values when clicking on input fields. This applies to all inputs.
-Auto commit a valid 6-char (or 8-char if alpha-channel is enabled).
-Commit a 3-char (or 4-char with alpha-channel) shorthand value onBlur.
-Auto commit channel inputs. If the user deletes the value completely, there will be no change. If left blank, the original value will be restored onBlur.
-Increase/decrease channel input value, when the input is cleared and Arrow-up or Arrow-down are pressed.
-Blur focused input when clicking anywhere within the component, this applies to all input fields.


packages/calcite-components/src/components/color-picker/color-picker.tsx
```diff
@@ -1684,4 +1736,15 @@ export class ColorPicker
...
+
+  private handleColorPickerBlur = (): void => {
+    this.el.blur();
+  };
+
+  private clickHandler = (): void => {
+    if (this.disabled) {
+      return;
+    }
+    this.handleColorPickerBlur();
```
driskull, line 1748: question: every time the internal container of the component is clicked, the component is blurred?

It seems a little odd to be calling blur on every click of the component, even if the input is clicked.

packages/calcite-components/src/components/color-picker-hex-input/color-picker-hex-input.tsx
```diff
@@ -452,4 +468,14 @@ export class ColorPickerHexInput implements LoadableComponent {
 
     return Color(nudgedChannels);
   }
+
+  /* 
+      if alphaChannel is enabled, commit valid 4 digit hex codes.
+      if alphaChannel is disabled, commit valid 3 digit hex codes.
+    */
+  private validateShortHandValue = (value: string): boolean => {
```
driskull, line 476:
```suggestion
private validateShortHandValue = (value: string): boolean => {
  const  length = value?.length;
  const {alphaChannel } = this;
    return (
      (alphaChannel && length === 4) || (!alphaChannel && length === 3)
    );
  };
```

packages/calcite-components/src/components/color-picker-hex-input/color-picker-hex-input.e2e.ts
```diff
@@ -272,11 +272,10 @@ describe("calcite-color-picker-hex-input", () => {
     await selectText(input);
     const longhandHexWithExtraChars = "bbbbbbbbc";
     await page.keyboard.type(longhandHexWithExtraChars);
-    await page.keyboard.press("Enter");
```
driskull, line 275: question: This test is not testing the Enter key any longer. Is that intentional?

packages/calcite-components/src/components/color-picker/color-picker.tsx
```diff
@@ -476,8 +480,11 @@ export class ColorPicker
 
     let inputValue: string;
 
-    if (this.isClearable && !input.value) {
+    if (!input.value) {
       inputValue = "";
+      this.channelInputClear = true;
+      // reset this to allow typing in new value when channel input is empty
```
driskull, line 486: question: reading this comment, would a private property like `allowTypingNewVal` be better to allow typing a new value when the input is empty?

Maybe a single property like `isInputEmpty` or `isClear` could be used in place of `channelInputClear` and `arrowUpOrDownTracker`?

packages/calcite-components/src/components/color-picker/color-picker.tsx
```diff
@@ -340,6 +340,10 @@ export class ColorPicker
 
   private shiftKeyChannelAdjustment = 0;
 
+  private arrowUpOrDownTracker = "";
```
driskull, line 343: If keeping this property can we type it to something like:

```ts
private arrowKeyTracker: "up" | "down" | null = null;
```

packages/calcite-components/src/components/color-picker/color-picker.tsx
```diff
@@ -491,9 +498,27 @@ export class ColorPicker
...
       // we treat nudging as a change event since the input won't emit when modifying the value directly
       this.handleChannelChange(event);
+    } else if (inputValue !== "") {
+      this.handleChannelChange(event);
     }
   };
 
+  private handleChannelBlur = (event: CustomEvent): void => {
+    const input = event.currentTarget as HTMLCalciteInputNumberElement;
+    const channelIndex = Number(input.getAttribute("data-channel-index"));
```
driskull, line 508: nitpick: would be nice in the future if we didn't have to convert this to a string in the DOM and then have to parse it back as a Number. Maybe `channelsToRender` could be stored and used to lookup the index in the future.

driskull (commented):
Looking good! Had some comments @aPreciado88 👍

packages/calcite-components/src/components/color-picker/color-picker.tsx
aPreciado88: After reviewing with Franco yesterday, I'm looking for an alternate solution to this. The goal is to be able to blur the input fields without having to click outside of the component.

packages/calcite-components/src/components/color-picker-hex-input/color-picker-hex-input.tsx
aPreciado88: Franco pointed out this [util](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/components/color-picker/utils.ts#L34) I'm using this and getting rid of `validateShortHandValue`.

packages/calcite-components/src/components/color-picker-hex-input/color-picker-hex-input.e2e.ts
aPreciado88: I removed the Enter key since values are now auto-committed when typing/pasting a valid hex code. Do you think I should add it back?

driskull: No, that sounds reasonable. 👍

packages/calcite-components/src/components/color-picker/color-picker.tsx
aPreciado88: I'm clearing the `arrowUpOrDownTracker` to prevent an edge case [here](https://github.com/Esri/calcite-design-system/pull/9701/files#diff-ab28035ea6ab07b366c5635f72ec60a722bc674f8fe0a6aeb80c8d43ef4bbb7aR584). If arrowUp or arrowDown are triggered before clearing the input, and then a new value is typed, the logic within that `if` will be executed. I can try to validate this with a single variable, like you suggested.

jcfranco (changes requested):
This is looking great, @aPreciado88! 🤘😎🤘

Let's sync up to discuss options for the FIXME comment.

aPreciado88:
@jcfranco I addressed your comments and updated the PR. I'll create a new issue to work on updating the color-picker test suite.

jcfranco (commented):
LGTM!

Is it possible for the title to include a bit of context regarding the enhancements? I think they're all input-UX related, right?

jcfranco (approved):
LGTM!

Is it possible for the title to include a bit of context regarding the enhancements? I think they're all input-UX related, right?

packages/calcite-components/src/components/color-picker/color-picker.tsx
```diff
@@ -79,9 +84,7 @@ const throttleFor60FpsInMs = 16;
 @Component({
   tag: "calcite-color-picker",
   styleUrl: "color-picker.scss",
-  shadow: {
-    delegatesFocus: true,
-  },
+  shadow: {},
```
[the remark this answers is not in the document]
jcfranco: I meant to suggest restoring `shadow` to `true` and not removing it entirely. Thanks for the catch, @aPreciado88, and sorry for the hassle!

PR #9726, opened 2024-07-03
test(loader): add component token E2E tests

jcfranco:
**Related Issue:** #7190

## Summary

✨🧪✨


packages/calcite-components/src/components/loader/loader.scss
```diff
@@ -3,16 +3,16 @@
...
  * @prop --calcite-loader-color-middle: defines the color of a loader. Applies to --calcite-loader-color between 33% - 66% of the animation.
  * @prop --calcite-loader-color-start:  defines the starting color of a loader. Applies to --calcite-loader-color between 0% - 33% of the animation.
  * @prop --calcite-loader-font-size: [Deprecated] Specifies the font size of the loading percentage when type is `"determinate"`.
  * @prop --calcite-loader-padding: [Deprecated] Specifies the padding of the loader.
  * @prop --calcite-loader-size-inline: [Deprecated] The width and height of an inline loader.
  * @prop --calcite-loader-size: [Deprecated] The width and height of a non-inline loader.
- * @prop --calcite-loader-text-color: var(--calcite-color-text-1); defines the text color of a loader.
- * @prop --calcite-loader-track-color-determinate: var(--calcite-color-border-3); defines the track color of a "determinate" loader
+ * @prop --calcite-loader-text-color: defines the text color of a loader.
+ * @prop --calcite-loader-track-color-determinate: defines the track color of a "determinate" loader
```
alisonailea, line 15: I think this should be simplified to `--calcite-loader-track-color`

alisonailea (changes requested):
I think this one may have been done already. Can you resolve conflicts with the epic?

- I think we want to leave the variables that were already made available becuase they were probably requested to fit a user use case. We just need to make sure they align with our naming schema.
- -size is preferred to -width
- -space is preferred to -padding
- Skye asked that we reduce loader to a single color variable

alisonailea (approved):
✨

Issue #6980, opened 2023-05-18
[Panel and Flow Item] Add scale property


asangma:
### Description

Add a scale property to Panel.
Popover has scale, and it is helpful. But there are times when a flow needs to happen in a Popover **and** in a compact space.

<img width="880" alt="image" src="https://github.com/Esri/calcite-components/assets/12503298/2a9ad85b-f2e0-48d2-8aac-4650c3abc907">

### Acceptance Criteria

Utilize existing scaling conventions.
Work in Popover.
Work in ShellPanel.
In accordance with issue #7623

### Relevant Info
A dependency needed for the Dialog + Modal component effort in https://github.com/Esri/calcite-design-system/issues/7886
See [Figma file](https://www.figma.com/file/wVqftT1aZDjwYVkDZ2V5fC/Dialog?type=design&node-id=621-2663&mode=design&t=QMN0fh9fWhT7QLTT-11) for specs
<img width="1695" alt="image" src="https://github.com/Esri/calcite-design-system/assets/29716057/71a24d56-6d20-4650-a60b-f5b001e75055">

### Which Component

Panel and FlowItem

### Example Use Case

When a flow needs to happen in a compact space...especially in a Popover. See image above.

### Priority impact

p4 - not time sensitive

### Esri team

N/A

macandcheese:
What would adjust with scale? Rendered text and padding? If so, may want to address Block as well as it would be used within these and may end up with mismatched text sizes and alignment.

asangma:
@macandcheese Good point.

For Panel and FlowItem, I would say font-size and padding.

And for Block, I'm thinking since it is currently at 14px, we could match Button's font-sizes.

DitwanP:
🍡 Verified locally on 'dev'

DitwanP:
Forgive me @Elijbet, reopening to add controls for `scale` to the simple story on Storybook.

DitwanP:
🍡Verified, Part 2 🍿

PR #9730, opened 2024-07-03
feat(panel, flow-item): add scale property

Elijbet:
**Related Issue:** #6980

## Summary
Add `scale` property to `panel` and `flow-item`.


aPreciado88 (approved):
This LGTM!

driskull (approved):
Looks good.

Can we add defaults e2e tests for flow-item and panel to test for the scale's default value?

Issue #6017, opened 2022-12-13
Action-menu: implement OpenCloseComponent interface


Elijbet:
**monday.com sync:** #18048589272

### Description

`action-menu` uses `popover` internally, and it will need to handle emitting its own events on transition sequence.

Implement `useToggleTransitionEvents` controller when #14909 lands. 

### Which Component

action-menu

Issue #4544, opened 2022-05-12
Refactor: Open/Close events should be publicly exposed for top-level components


Elijbet:
**monday.com sync:** #18048591281

### Description

**Related Issue:** #2256

## Summary

Follow-up issue: Open/Close events should be publicly exposed, as these might be useful for users at some point. Especially for a top-level component that isn't emitting this event for another Calcite component to listen for. In that case, just ensure it does not have the @internal label or Internal prefix and is documented.

### Proposed Advantages

Users might need control over these.

### Which Component

Based on the comment below:
The originating issue for this has been addressed (making combobox open/close events public). Other components could implement OpenCloseComponent with associated events:

`open` prop:

- [x] action-menu (https://github.com/Esri/calcite-design-system/issues/6017)
- [x] block-section
- [x] list-item
- [x] menu-item

`expanded` prop:

- [x] accordion-item
- [x] action-bar
- [x] action-group
- [x] action-pad
- [x] tree-item

`closed` prop:

- [ ] list-item
- [ ] chip
- [ ] tab-title
- [ ] flow-item
- [ ] panel

`collapsed` prop:

- [x] panel
- [x] shell-panel
- [x] flow-item

We may also need to tackle https://github.com/Esri/calcite-design-system/issues/5074 along with this issue.

jcfranco:
It seems the originating issue for this has been addressed (making combobox open/close events public). 

There are other components that could implement `OpenCloseComponent` with associated events:

**`open` prop**

* `action-menu` (https://github.com/Esri/calcite-design-system/issues/6017)
* `block-section`
* `list-item`
* `menu-item`

**`expanded` prop**

* `accordion-item`
* `action-bar`
* `action-group`
* `action-pad`
* `tree-item` (`expanded` prop)

Not sure if we want to address those in a separate PR or repurpose this one. 

We may also need to tackle https://github.com/Esri/calcite-design-system/issues/5074 along with this issue.

driskull:
Do we have an issue to add these kind of events for `closed` props on `panel`, `flow-item`, `list-item`, etc?

Elijbet:
> Do we have an issue to add these kind of events for `closed` props on `panel`, `flow-item`, `list-item`, etc?

I'll add those to this issue as well.

- `list-item`
- `panel` 
- `flow-item`
- `chip` 
- `tab-title`
- `tip` 
- `tip-manager`

Elijbet:
[...]

PR #9736, opened 2024-07-04
feat(`action-menu`, `block-section`, `list-item`, `menu-item`): add open, close, beforeOpen, and beforeClose events for consistent event handling

Elijbet:
**Related Issue:** #6017, #4544

## Summary
Implement `OpenCloseComponent` with associated events to `action-menu`, `block-section`, `list-item`, `menu-item`.

### DIFF
--- a/packages/calcite-components/src/components/menu-item/menu-item.tsx
+++ b/packages/calcite-components/src/components/menu-item/menu-item.tsx
@@ -165,7 +165,7 @@ export class CalciteMenuItem
 
   openTransitionProp = "opacity";
 
-  transitionEl: HTMLDivElement;
+  transitionEl: HTMLElement;
 
   //--------------------------------------------------------------------------
   //
@@ -307,6 +307,10 @@ export class CalciteMenuItem
       }
     });
     this.hasSubmenu = this.submenuItems.length > 0;
+  };
+
+  private setTransitionEl = (el: HTMLElement): void => {
+    this.transitionEl = el;
   };
 
   private keyDownHandler = async (event: KeyboardEvent): Promise<void> => {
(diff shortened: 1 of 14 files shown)


alisonailea:
I'd like to suggest we normalize all components to a single "Close" class that includes these evens as well as the template for a consistent close button UI. @jcfranco

jcfranco:
@alisonailea Let's discuss that separately since there are a few challenges with your suggestion. Namely,

* Stencil doesn't support mixins nor extending classes.
* Events are typed and based on the component name.
* The close button is not entirely tied to these events.

packages/calcite-components/src/components/menu-item/menu-item.tsx
```diff
@@ -465,6 +513,7 @@ export class CalciteMenuItem implements LoadableComponent, T9nComponent, Localiz
             [CSS.container]: true,
             [CSS.isParentVertical]: this.topLevelMenuLayout === "vertical",
           }}
+          ref={this.setTransitionEl}
```
driskull, line 516: Wouldn't this overwrite the other transitionEl set on line 472?

driskull (changes requested):

packages/calcite-components/src/components/menu-item/menu-item.tsx
Elijbet: I'm not exactly sure where the right place for it is. It's kind of a Russian doll situation. When you slot items into `menu-items` it renders `menu`. So you could set it on the `container` for menu-item, or `menu` itself?

driskull (approved):
LGTM! Lets wait for @jcfranco to review.

PR #9737, opened 2024-07-05
test(color-picker): add component token E2E tests

jcfranco:
**Related Issue:** #7180

## Summary

✨🧪✨


packages/calcite-components/src/components/color-picker/color-picker.e2e.ts
```diff
@@ -2345,4 +2355,274 @@ describe("calcite-color-picker", () => {
...
+          {
+            shadowSelector: `.${CSS.hexInput}`,
+            targetProp: "--calcite-color-picker-hex-input-text-color",
+          },
+        ],
+        "--calcite-color-picker-tab-nav-indicator-color": {
+          shadowSelector: `.${CSS.colorModes}`,
+          targetProp: "--calcite-tab-nav-indicator-color",
+        },
+        "--calcite-color-picker-tab-title-background-color": {
```
alisonailea, line 2592: I don't think this token is actually effecting anything.

packages/calcite-components/src/components/color-picker/color-picker.e2e.ts
```diff
@@ -2345,4 +2355,274 @@ describe("calcite-color-picker", () => {
...
+          {
+            shadowSelector: `.${CSS.channel}`,
+            targetProp: "--calcite-input-border-color",
+          },
+          {
+            shadowSelector: `.${CSS.hexInput}`,
+            targetProp: "--calcite-color-picker-hex-input-border-color",
+          },
+        ],
+        "--calcite-color-picker-input-corner-radius": [
```
alisonailea, line 2536: I don't think this token is actually effecting anything.

packages/calcite-components/src/components/color-picker/color-picker.e2e.ts
```diff
@@ -2345,4 +2355,274 @@ describe("calcite-color-picker", () => {
...
+          {
+            shadowSelector: `.${CSS.channel}`,
+            targetProp: "--calcite-input-background-color",
+          },
+          {
+            shadowSelector: `.${CSS.hexInput}`,
+            targetProp: "--calcite-color-picker-hex-input-background-color",
+          },
+        ],
+        "--calcite-color-picker-input-background-color-hover": [
```
alisonailea, line 2514: I don't think this token is actually effecting anything.

packages/calcite-components/src/components/color-picker/color-picker.e2e.ts
```diff
@@ -2345,4 +2355,274 @@ describe("calcite-color-picker", () => {
...
+            targetProp: "--calcite-button-icon-color",
+            state: "hover",
+          },
+          {
+            shadowSelector: `.${CSS.saveColor}`,
+            targetProp: "--calcite-button-icon-color",
+            state: "hover",
+          },
+        ],
+        "--calcite-color-picker-button-text-color": [
```
alisonailea, line 2458: --calcite-color-picker-text-color and --calcite-color-picker-button-text-color have the same color by default. I think we can simplify this to 
`--calcite-color-picker-text-color` + the various states

packages/calcite-components/src/components/color-picker/color-picker.e2e.ts
```diff
@@ -2345,4 +2355,274 @@ describe("calcite-color-picker", () => {
...
+            targetProp: "--calcite-button-background-color",
+            state: "hover",
+          },
+          {
+            shadowSelector: `.${CSS.saveColor}`,
+            targetProp: "--calcite-button-background-color",
+            state: "hover",
+          },
+        ],
+        "--calcite-color-picker-button-icon-color": [
```
alisonailea, line 2412: none of the -button-icon-color(-) tokens are used. Remove.

packages/calcite-components/src/components/color-picker/color-picker.e2e.ts
```diff
@@ -2345,4 +2355,274 @@ describe("calcite-color-picker", () => {
...
+            targetProp: "--calcite-color-picker-hex-input-prefix-background-color",
+          },
+        ],
+        "--calcite-color-picker-input-prefix-text-color": [
+          {
+            shadowSelector: `.${CSS.hexInput}`,
+            targetProp: "--calcite-color-picker-hex-input-prefix-text-color",
+          },
+        ],
+        "--calcite-color-picker-input-suffix-background-color": [
```
alisonailea, line 2558: pefix/suffix tokens aren't used.

packages/calcite-components/src/components/color-picker/color-picker.e2e.ts
```diff
@@ -2345,4 +2355,274 @@ describe("calcite-color-picker", () => {
...
+      });
+    });
+
+    describe("opacity enabled", () => {
+      themed(html`<calcite-color-picker alpha-channel></calcite-color-picker>`, {
+        "--calcite-color-picker-input-suffix-background-color": {
+          shadowSelector: `.${CSS.channel}[data-channel-index="3"]`,
+          targetProp: "--calcite-input-suffix-background-color",
+        },
+        "--calcite-color-picker-input-suffix-text-color": {
```
alisonailea, line 2621: this doesn't effect anything. remove?

alisonailea (commented):
I opened the tests in debugger mode and it looks like several of the tokens are not applied?

jcfranco:
@alisonailea I've removed some of the unused props you pointed out. Thanks for the catch! ✨🎣✨

Others that were pointed out were working, but not shown up for specific configurations or because there was an issue with styles. I've also fixed these and here's how they're looking:

Screenshot 2024-07-16 at 3 09 15 PM

A couple of questions on ☝️:

1. the background/text color of prefix and suffixes should be the same, so WDYT about consolidating these to `--calcite-color-picker-affix-background-color` and `--calcite-color-picker-affix-text-color`?
2. both button icon/text color props can be applied, but both can be applied due to specificity and `currentColor`: should we drop the icon prop in favor of the button-text prop or, following your suggestion, rely on the component's text color prop? The buttons do change background color when hovered, so it might be useful to allow customizing this differently from the component's text color.

alisonailea:
1. yes I think `--calcite-color-picker-affix-text-color` makes sense
2. I'm all for dropping extraneous tokens

jcfranco:
@alisonailea Updated. Would you mind reviewing this again? 🙏

alisonailea (approved):

Issue #9721, opened 2024-07-01
Update calciteConfig message when same or different versions


odoe:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Currently, when the `globalThis.calciteConfig` is checked if a Calcite version is loaded, it emits a message that doesn't distinguish if the same version or a different version was loaded.

There are cases when the same version could be loaded if an app that uses Stencil built components mixes `import` of components and `defineCustomElements`, which is probably not best practice.

Not sure what the message when the same version is loaded should be, but definitely a distinct warning when a different version is loaded, probably means something went wrong.

### Acceptance Criteria

- [ ] Two messages for when the same Calcite version or a different Calcite version is loaded.
- [ ] Confirm regular Vite setup doesn't add the same message with [our example](https://github.com/Esri/calcite-components-examples/tree/main/vite)

### Which Component

calcite-components/src/utils/config.ts

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

jcfranco:
> Confirm regular Vite setup doesn't add the same message with [our example](https://github.com/Esri/calcite-components-examples/tree/main/vite?rgh-link-date=2024-07-01T23%3A03%3A44Z)

@odoe I can't reproduce the same-version warning with the vite example. Do you have any suggestions on how to get the warning?

odoe:
@jcfranco I made a repro here
https://github.com/odoe/calcite-dupe-issue

It updates the `main.js` to do both `defineCustomElements` and an import from map components. This will trigger the loading of the config file from two different dist folders in calcite.

```js
import './style.css';
// loads calcite from components folder
import { defineCustomElements as defineCalciteElements } from "@esri/calcite-components/dist/loader";
// loads calcite from esm folder
import "@arcgis/map-components/dist/components/arcgis-map";

defineCalciteElements(window, {
  resourcesUrl: "https://js.arcgis.com/calcite-components/2.10.1/assets",
});

const loader = document.createElement('calcite-loader');
document.body.appendChild(loader);
```

jcfranco:
[...]

PR #9739, opened 2024-07-05
feat: provide info message on stamped version instead of warning

jcfranco:
**Related Issue:** #9721

## Summary

This updates the message displayed when components load to show the version, build date and revision (similar to Maps SDK). It will log an info message instead of a warning as it could be misinterpreted and would also show up multiple times when a bundled `calcite-components` version would be overridden.

**Note**: `console.info` messages will be omitted by default across all tests to avoid noise

### DIFF
--- a/packages/calcite-components/src/utils/config.spec.ts
+++ b/packages/calcite-components/src/utils/config.spec.ts
@@ -30,6 +30,8 @@ describe("config", () => {
   describe("stampVersion", () => {
     const calciteVersionPreBuildPlaceholder = "__CALCITE_VERSION__";
 
+    beforeEach(() => delete globalThis.calciteConfig);
+
     it("creates global config and stamps the version onto it", async () => {
       config = await loadConfig();
       config.stampVersion();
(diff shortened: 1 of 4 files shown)


packages/calcite-components/src/utils/config.spec.ts
```diff
@@ -43,12 +45,19 @@ describe("config", () => {
       expect(globalThis.calciteConfig.version).toBe(calciteVersionPreBuildPlaceholder);
     });
 
-    it("warns if the version is already set", async () => {
-      globalThis.calciteConfig = { version: "1.33.7" };
+    it("bails if version is already stamped onto existing config", async () => {
+      const testVersion = "1.33.7";
```
driskull, line 49: 😎

driskull (approved):
This is awesome 👏 👍

packages/calcite-components/src/utils/config.spec.ts
jcfranco: 😎

Martinspire:
My application uses Esri under the hood for some stuff that other teams developed and suddenly its always logging the calcite version. Why is this message always being displayed for every project that uses these components (even if its just a dependency of a dependency)? I'd rather see this being opt-in than opt-out. I don't even mind that it is being set on the globalThis value but I just hate having extra logging that is irrelevant 99.999% of the time.

Issue #9740, opened 2024-07-06
[input-number] numeric keyboard no longer appears on mobile devices/tablets when input is focused


jcfranco:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The numeric keyboard no longer shows up on mobile devices/tablets when an `input-number` receives focus.

As a workaround, `inputmode="decimal"` (or `inputMode` property) can be set on the component.

### Expected Behavior

For the numeric keyboard to show up when an `input-number` is focused on mobile/tablets.

### Reproduction Sample

https://codepen.io/jcfranco/pen/dyBbwBx

### Reproduction Steps

1. Open the repro case on iOS or iPad
2. Tap on the input
3. Notice how the default virtual keyboard shows up (changing the version to 2.7.1 triggers the numeric keyboard when focused)

### Reproduction Version

2.10.1

### Relevant Info

2.8.0 introduced a regression (see https://github.com/Esri/calcite-design-system/pull/9123/files#diff-eb691ddf7a85985a3113e037a62db7b4c27bc134ac572baccecd23500f40c268L283) where `input-number`'s `inputMode` property is no longer `decimal`.

### Regression?

2.7.1

### Priority impact

impact - p1 - need for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

DitwanP:
🍡 Verified on `2.11.0-next.6` using iOS with Chrome, FireFox, and Safari

PR #9741, opened 2024-07-06
fix(input-number): restore decimal input mode default

jcfranco:
**Related Issue:** #9740

## Summary

Restores `decimal` default for the internal input's `inputmode`, which was unintentionally removed in https://github.com/Esri/calcite-design-system/pull/9123.


driskull (approved):
Nice 👍

Issue #8269, opened 2023-11-27
[Shell / Shell Panel] When `content-behind` is set, Shell Panel in slot "panel-bottom" can position incorrectly


macandcheese:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When Shell has `content-behind` set, a Shell Panel of  in slot "panel-bottom" with `displayMode="float"` will not display properly if there is not content slotted in "panel-top".

### Expected Behavior

I'd expect the slotted "panel-bottom" content to always position in the bottom, without the need to populate "panel-top".

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/bGzjMLg?editors=1100

### Reproduction Steps

1. Open Codepen
2. See incorrectly positioned "bottom panel"
3. Uncomment `collapsed` "top panel"
4. See positioning is correct

### Reproduction Version

1.10

### Relevant Info

There are some inconsistencies with how this behaves when `content-behind` is set, or unset.

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

macandcheese:
The workaround is to have a collapsed Shell Panel, in the `panel-top` slot.

DitwanP:
🍡 Verified on `2.11.0-next.6`
https://codepen.io/Ditwan-Price/pen/eYwNExg?editors=1100

Issue #9085, opened 2024-04-08
Drop `_TestOnly` story suffix


jcfranco:
**monday.com sync:** #18048705680

### Description

https://github.com/Esri/calcite-design-system/pull/9030 removed story filtering based on build output, so since all stories are test-related, we should drop the `_TestOnly` test suffix. 

This should be done once https://github.com/Esri/calcite-design-system/issues/7180 is complete to avoid baseline screenshot syncing.

**Note:** renamed stories will show up as new screenshot tests and will need to be accepted as the new baseline.

### Proposed Advantages

Simple and clear story naming.

### Which Component

All `_TestOnly` named stories.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

DintaMel:
Verified by searching for "_TestOnly". No matches found. Also manually spot checked stories that previously had "_TestOnly" in Storybook to confirm that "_TestOnly" was removed.

PR #9748, opened 2024-07-09
fix(shell): update shell to correctly position calcite shell panel at shell's bottom

aPreciado88:
**Related Issue:** [#8269](https://github.com/Esri/calcite-design-system/issues/8269)

## Summary
Update `calcite-shell` to correctly position `calcite-shell-panel` with `slot="panel-bottom"` at calcite-shell's bottom, when there is no `calcite-shell-panel` with `slot="panel-top"`.

### DIFF
--- a/packages/calcite-components/src/components/shell/shell.stories.ts
+++ b/packages/calcite-components/src/components/shell/shell.stories.ts
@@ -848,7 +848,7 @@ export const slottedPanelTop_TestOnly = (): string =>
   </calcite-shell>
 `);
 
-export const contentBehindPanelBottom_TestOnly = (): string =>
+export const contentBehindPanelBottom = (): string =>
   html(`
   <calcite-shell
     content-behind
--- a/packages/calcite-components/src/components/shell/shell.stories.ts
+++ b/packages/calcite-components/src/components/shell/shell.stories.ts
@@ -99,6 +99,12 @@ const centerRowHTML = html`
   </calcite-panel>
 `;
 
+const bottomPanelHTML = html`
+  <calcite-panel heading="Panel bottom content">
+    <div>Content</div>
+  </calcite-panel>
+`;
+
 const centerRowWithActionBarHTML = html`
   <calcite-action-bar slot="action-bar">
     <calcite-action-group>
@@ -841,6 +847,30 @@ export const slottedPanelTop_TestOnly = (): string =>
     <footer slot="footer">Footer Example</footer>
   </calcite-shell>
 `);
+
+export const contentBehindPanelBottom_TestOnly = (): string =>
+  html(`
+  <calcite-shell
+    content-behind
+    style="
+    width:700px;
+    height:700px;
+    position:relative;
+    "
+  >
+      <div
+      style="
+      width:100%;
+      height:100%;
+      background-image: linear-gradient(45deg, #ccc 25%, transparent 25%),
+      linear-gradient(-45deg, #ccc 25%, transparent 25%),
+      linear-gradient(45deg, transparent 75%, #ccc 75%),
+      linear-gradient(-45deg, transparent 75%, #ccc 75%);
+      background-size: 20px 20px;
+      background-position: 0 0, 0 10px, 10px -10px, -10px 0px;"></div>
+      <calcite-shell-panel slot="panel-bottom" display-mode="float" layout="horizontal">${bottomPanelHTML}</calcite-shell-panel>
+    </calcite-shell>
+  `);
 
 export const slottedPanelBottom_TestOnly = (): string =>
   html(`
--- a/packages/calcite-components/src/components/shell/shell.tsx
+++ b/packages/calcite-components/src/components/shell/shell.tsx
@@ -193,6 +193,13 @@ export class Shell implements ConditionalSlotComponent {
 
     const contentContainerKey = "content-container";
 
+    const shellPanels: Array<HTMLCalciteShellPanelElement> = Array.from(
+      this.el.querySelectorAll("calcite-shell-panel"),
+    );
+    const hasOnlyBottomPanel: boolean = !!(
+      shellPanels.length === 1 && shellPanels.find((element) => element.slot === "panel-bottom")
+    );
+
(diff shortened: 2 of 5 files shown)


packages/calcite-components/src/components/shell/shell.tsx
```diff
@@ -193,6 +193,13 @@ export class Shell implements ConditionalSlotComponent {
 
     const contentContainerKey = "content-container";
 
+    const shellPanels: Array<HTMLCalciteShellPanelElement> = Array.from(
```
driskull, line 196: fixme: This would't work in cases when shell-panels are slotted in after the shell has been initialized. Instead, an `onSlotChange` should be used to find shell-panels.

packages/calcite-components/src/components/shell/shell.tsx
```diff
@@ -193,6 +193,13 @@ export class Shell implements ConditionalSlotComponent {
...
+      this.el.querySelectorAll("calcite-shell-panel"),
+    );
+    const hasOnlyBottomPanel: boolean = !!(
+      shellPanels.length === 1 && shellPanels.find((element) => element.slot === "panel-bottom")
```
driskull, line 200: fixme: Instead, use an `onSlotChange` event for the `panel-bottom` slot to determine if a panel bottom shell panel exists and save the state to a State() variable. Also you would need to determine if panel-top slotted element exists and save that to a state. The combined two of those state should give you if one is only panel-bottom. Another state for hasOnlyBottomPanel could be calculated from the two.

driskull (changes requested):
This would need to handle when a shell-panel is added at a later time.

packages/calcite-components/src/components/shell/shell.tsx
aPreciado88: @driskull I updated the logic to use `onSlotChange` events, please let me know if this looks okay.

packages/calcite-components/src/components/shell/shell.tsx
```diff
@@ -74,6 +74,12 @@ export class Shell implements ConditionalSlotComponent {
 
   @State() hasSheets = false;
 
+  @State() hasPanelTop = false;
+
+  @State() hasPanelBottom = false;
+
+  @State() hasOnlyPanelBottom = false;
+
```
driskull, line 82:
```suggestion
@Watch("hasPanelTop")
@Watch("hasPanelBottom")
  updateHasOnlyPanelBottom(): void {
    this.hasOnlyPanelBottom = !this.hasPanelTop && this.hasPanelBottom;
  }
```

packages/calcite-components/src/components/shell/shell.tsx
```diff
@@ -131,6 +137,24 @@ export class Shell implements ConditionalSlotComponent {
...
+  };
+
+  handlePanelBottomChange = (event: Event): void => {
+    const panelBottom = event.target as HTMLSlotElement;
+    this.hasPanelBottom = !!panelBottom;
+
+    this.updateHasOnlyPanelBottom();
+  };
+
+  updateHasOnlyPanelBottom = (): void => {
```
driskull, line 154: This can be removed if we use a watch to watch the two states we're interested in to update the other state.

packages/calcite-components/src/components/shell/shell.tsx
```diff
@@ -131,6 +137,24 @@ export class Shell implements ConditionalSlotComponent {
     });
   };
 
+  handlePanelTopChange = (event: Event): void => {
+    const panelTop = event.target as HTMLSlotElement;
+    this.hasPanelTop = !!panelTop;
```
driskull, line 142:
```suggestion
    this.hasPanelTop = slotChangeHasAssignedElement(event);
```

packages/calcite-components/src/components/shell/shell.tsx
```diff
@@ -131,6 +137,24 @@ export class Shell implements ConditionalSlotComponent {
...
+
+    this.updateHasOnlyPanelBottom();
+  };
+
+  handlePanelBottomChange = (event: Event): void => {
+    const panelBottom = event.target as HTMLSlotElement;
+    this.hasPanelBottom = !!panelBottom;
```
driskull, line 149:
```suggestion
    this.hasPanelBottom = slotChangeHasAssignedElement(event);
```

packages/calcite-components/src/components/shell/shell.tsx
```diff
@@ -131,6 +137,24 @@ export class Shell implements ConditionalSlotComponent {
     });
   };
 
+  handlePanelTopChange = (event: Event): void => {
+    const panelTop = event.target as HTMLSlotElement;
```
driskull, line 141: event.target here is the slot element. We don't want to use that, instead we need to know if the slot element has assigned elements.

driskull (changes requested):
Almost there @aPreciado88. nice work 🌮

driskull (approved):
This looks good once there's a screenshot test and e2e test. Screenshot test should show only the panel bottom to make sure its working correctly. e2e test should check to make sure the contentBottom class is being applied when it should be.

aPreciado88:
@driskull I added snapshot and e2e tests. Thank you for taking a look at this!

packages/calcite-components/src/components/shell/shell.stories.ts
```diff
@@ -842,6 +848,30 @@ export const slottedPanelTop_TestOnly = (): string =>
   </calcite-shell>
 `);
 
+export const contentBehindPanelBottom_TestOnly = (): string =>
```
driskull, line 851: nitpick: I don't think `_TestOnly` is needed any longer. right @jcfranco?

driskull (approved):
🎉

packages/calcite-components/src/components/shell/shell.stories.ts
jcfranco: Correct. https://github.com/Esri/calcite-design-system/issues/9085 for context.

Issue #514, opened 2020-04-27
Card: checkbox spacing, Dark theme checkbox


bstifle:
![Card checkbox spacing](https://user-images.githubusercontent.com/14083530/80429125-372c0c80-88a0-11ea-9202-981a715c7932.png)
Adjust y spacing to be equal to x

![Card Dark Checkbox](https://user-images.githubusercontent.com/14083530/80429127-385d3980-88a0-11ea-8018-aa2713ffdb9a.png)
make dark theme cards use dark theme checkboxes

PR #9751, opened 2024-07-09
feat(dialog): adds new dialog component and deprecates the modal component

driskull:
**Related Issue:** #7886

## Summary

- Adds new `calcite-dialog` component.
  - Includes e2e, stories, demos, resources
  - Updates stencil config to note new component
- Panel esc key should emit close event  
- Adds `dialogs` slot to `calcite-shell` for slotting dialogs.
- Deprecates `calcite-modal` component.
- cleans up global style imports
- makes `focusTrapDisabled` optional on FocusTrapComponent

## Notes

- Animations need review


packages/calcite-components/src/components/dialog/dialog.scss
```diff
@@ -0,0 +1,237 @@
...
+:host([placement="bottom-end"]) .container {
+  @apply justify-end items-end;
+}
+
+:host(:not([modal])) .container {
+  @apply pointer-events-none;
+}
+
+:host([scale="s"]) {
+  --calcite-dialog-padding-internal: theme("spacing.3");
```
alisonailea, line 69: Can you update these to use the naming schema?
`--calcite-internal-dialog-`

packages/calcite-components/src/components/dialog/dialog.scss
```diff
@@ -0,0 +1,237 @@
...
+
+:host([scale="l"]) {
+  --calcite-dialog-padding-internal: theme("spacing.5");
+  --calcite-dialog-padding-large-internal: theme("spacing.6");
+  --calcite-dialog-title-text-internal: theme("fontSize.3h");
+  --calcite-dialog-content-text-internal: theme("fontSize.1");
+}
+
+.scrim {
+  --calcite-scrim-background: var(--calcite-dialog-scrim-background, var(--calcite-color-transparent-scrim));
```
alisonailea, line 90: can you add
`--calcite-scrim-background-color: var(--calcite-dialog-scrim-background-color, var(--calcite-color-transparent-scrim));`
to make merging with the token's epic easier?

packages/calcite-components/src/components/dialog/dialog.scss
```diff
@@ -0,0 +1,237 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-dialog-scrim-background: Specifies the background color of the component's scrim.
```
alisonailea, line 6: `--calcite-dialog-scrim-background-color`

packages/calcite-components/src/components/dialog/dialog.scss
```diff
@@ -0,0 +1,237 @@
...
+ * @prop --calcite-dialog-width: Specifies the width of the component, using `px`, `em`, `rem`, `vw`, or `%`. Will never exceed the width of the viewport. Will not apply if `placement="cover"` if set.
```
alisonailea, line 7: `--calcite-dialog-size-x`

packages/calcite-components/src/components/dialog/dialog.scss
```diff
@@ -0,0 +1,237 @@
...
+ * @prop --calcite-dialog-height: Specifies the height of the component, using `px`, `em`, `rem`, `vh`, or `%`. Will never exceed the height of the viewport. Will not apply if `placement="cover"` if set.
```
alisonailea, line 8: `--calcite-dialog-size-y`

packages/calcite-components/src/components/dialog/dialog.scss
```diff
@@ -0,0 +1,237 @@
...
+ * @prop --calcite-dialog-panel-footer-padding: Specifies the padding of the component's footer.
```
alisonailea, line 9: `--calcite-dialog-panel-footer-space`

packages/calcite-components/src/components/dialog/dialog.scss
```diff
@@ -0,0 +1,237 @@
...
+ * @prop --calcite-dialog-panel-header-border-block-end: Specifies the component header's block end border.
```
alisonailea, line 10: simplify to `--calcite-dialog-border-color`

packages/calcite-components/src/components/dialog/dialog.scss
```diff
@@ -0,0 +1,237 @@
...
+}
+
+.scrim {
+  --calcite-scrim-background: var(--calcite-dialog-scrim-background, var(--calcite-color-transparent-scrim));
+  @apply fixed inset-0 flex overflow-y-hidden;
+}
+
+calcite-panel {
+  --calcite-panel-footer-padding: var(--calcite-dialog-panel-footer-padding);
+  --calcite-panel-header-border-block-end: var(--calcite-dialog-panel-header-border-block-end);
```
alisonailea, line 96: `--calcite-panel-header-border-block-end: var(--calcite-border-width-sm) solid var(--calcite-dialog-border-color);`

packages/calcite-components/src/components/dialog/dialog.scss
```diff
@@ -0,0 +1,237 @@
...
+    opacity-0
+    z-modal;
+  -webkit-overflow-scrolling: touch;
+  visibility: hidden;
+  transition:
+    transform var(--calcite-internal-animation-timing-slow) $easing-function,
+    visibility 0ms linear var(--calcite-internal-animation-timing-slow),
+    opacity var(--calcite-internal-animation-timing-slow) $easing-function;
+
+  --calcite-dialog-hidden-position: translate3d(0, 20px, 0);
```
alisonailea, line 120: `--calcite-internal-dialog-hidden-position`

packages/calcite-components/src/components/dialog/dialog.scss
```diff
@@ -0,0 +1,237 @@
...
+  --calcite-dialog-shown-position: translate3d(0, 0, 0);
```
alisonailea, line 121: `--calcite-internal-dialog-shown-position`

alisonailea (changes requested):
Thanks for tokenizing some of this! Make sure to check the naming schema https://github.com/Esri/calcite-design-system/wiki/tokens-naming-schema

packages/calcite-components/src/components/dialog/dialog.scss
driskull: Yep! thanks

packages/calcite-components/src/components/dialog/dialog.scss
driskull: ofcourse!

driskull: Note: we have to have both of these vars for now as the new one doesn't work yet.

alisonailea (approved):
LGTM!

packages/calcite-components/src/components/shell/shell.tsx
```diff
@@ -197,6 +209,14 @@ export class Shell implements ConditionalSlotComponent {
     );
   }
 
+  renderDialogs(): VNode {
+    return (
+      <div hidden={!this.hasDialogs}>
+        <slot key="dialogs" name={SLOTS.dialogs} onSlotchange={this.handleDialogsSlotChange} />
```
benelan, line 215: `onSlotchange` always breaks my brain even though I know it's correct

benelan (approved):
A couple nitpicks but otherwise LGTM!

packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,509 @@
...
+   * The component header text.
+   */
+  @Prop() heading: string;
+
+  /**
+   * Specifies the heading level of the component's `heading` for proper document structure, without affecting visual styling.
+   */
+  @Prop({ reflect: true }) headingLevel: HeadingLevel;
+
+  /** Specifies the kind of the component, which will apply to top border. */
```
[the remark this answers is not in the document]
macandcheese: This shouldn’t change the visual display so not sure style is needed here.

packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,509 @@
...
+ * @slot header-actions-start - A slot for adding actions or content to the start side of the header.
+ * @slot header-actions-end - A slot for adding actions or content to the end side of the header.
+ * @slot header-content - A slot for adding custom content to the header.
+ * @slot header-menu-actions - A slot for adding an overflow menu with actions inside a `calcite-dropdown`.
+ * @slot fab - A slot for adding a `calcite-fab` (floating action button) to perform an action.
+ * @slot footer - A slot for adding custom content to the component's footer.
+ * @slot footer-end - A slot for adding a trailing footer custom content.
+ * @slot footer-start - A slot for adding a leading footer custom content.
+ */
+
```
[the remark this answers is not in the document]
driskull: I'm seeing a lot of other components with a line between so not sure if this is an issue.

packages/calcite-components/src/components/dialog/dialog.tsx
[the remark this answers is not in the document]
driskull: Done. this was straight up copied from Modal so it might need an adjustment as well.

packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,509 @@
...
+  @Event({ cancelable: false }) calciteDialogScroll: EventEmitter<void>;
+
+  //--------------------------------------------------------------------------
+  //
+  //  Public Methods
+  //
+  //--------------------------------------------------------------------------
+
+  /**
+   * Sets focus on the component's "close" button (the first focusable item).
```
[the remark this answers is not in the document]
driskull: I didn't try. Do we know why modal wasn't using delegatesFocus?

packages/calcite-components/src/components/dialog/dialog.tsx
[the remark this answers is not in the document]
driskull: cc @geospatialem and @DitwanP for wording/doc update.

packages/calcite-components/src/components/dialog/dialog.tsx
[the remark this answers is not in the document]
driskull: @benelan do you remember the cons of using delegatesFocus and why we weren't using it in all components? We should probably document this somewhere for future reference. cc @jcfranco incase he knows.

packages/calcite-components/src/components/dialog/dialog.tsx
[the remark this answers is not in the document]
DitwanP: @driskull Looks like the current wording you have is what we've used in most other components, so for consistency sake its good. 

Although, it seems weird to me that in this case and all the others there's no "the" between "to" and "top border". I feel like it should read "which will apply to **_the_** top border..." but thats for another PR, it's fine as is for now.

packages/calcite-components/src/components/dialog/dialog.tsx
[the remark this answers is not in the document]
benelan: I believe Modal was skipped because we needed to use custom logic for the focus trap. I'll try to refresh my memory and add some documentation to the wiki after the release. I'd say let's move forward without delegatesFocus for now, and I can re-evaluate after I'm refreshed.

packages/calcite-components/src/components/dialog/dialog.tsx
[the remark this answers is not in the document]
benelan: Good to know, thanks!

packages/calcite-components/src/components/dialog/dialog.tsx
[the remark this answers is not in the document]
benelan: I found the issue: #514 

I was putting research notes in the body, including a list of components and why they were skipped. Modal (and most of the skipped components) were related to the `disabled` property:

> If a "mock disabled" element is first in the flat tree, it will be delegated focus. A lot of our components can have slotted content as the first focusable element.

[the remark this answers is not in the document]
benelan: Correction, the important info is in https://github.com/Esri/calcite-design-system/issues/5147#issuecomment-1349616672

I only thing I added to the body was **bolded** responses to Franco's task list.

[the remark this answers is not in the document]
driskull: nice find! we should save this somewhere in a markdown or wiki page.

packages/calcite-components/src/components/dialog/dialog.scss
```diff
@@ -0,0 +1,241 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-dialog-scrim-background-color: Specifies the background color of the component's scrim.
+ * @prop --calcite-dialog-size-x: Specifies the width of the component, using `px`, `em`, `rem`, `vw`, or `%`. Will never exceed the width of the viewport. Will not apply if `placement="cover"` if set.
+ * @prop --calcite-dialog-size-y: Specifies the height of the component, using `px`, `em`, `rem`, `vh`, or `%`. Will never exceed the height of the viewport. Will not apply if `placement="cover"` if set.
```
geospatialem, line 8: copyedit nitpick for `...-size-x`:

> .... ~Will never exceed...~ Does not exceed the viewport's width - applies when `placement="cover"` is not set.

and `...size-y`:

> ....~Will never exceed....~ Does not exceed the viewport's height - applies when `placement="cover"` is not set.

packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,509 @@
...
+
+let totalOpenDialogs: number = 0;
+let initialDocumentOverflowStyle: string = "";
+
+/**
+ * @slot - A slot for adding content.
+ * @slot content - A slot for adding custom content.
+ * @slot action-bar - A slot for adding a `calcite-action-bar` to the component.
+ * @slot alerts - A slot for adding `calcite-alert`s to the component.
+ * @slot content-bottom - A slot for adding content below the unnamed (default) slot and above the footer slot (if populated)
```
geospatialem, line 57: nitpick: 

Add a period at the end of the `content-bottom` slot description and a minor copyedit and backticks for the slot:

> A slot for adding content below the unnamed (default) slot and - if populated - above the `footer` slot.

packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,509 @@
...
+ * @slot content-top - A slot for adding content above the unnamed (default) slot and below the action-bar slot (if populated).
```
geospatialem, line 58: nitpick: 

Copyedit suggestion for consistency to the above and backticks for the slot:

> A slot for adding content above the unnamed (default) slot and - if populated - below the `action-bar` slot.

packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,509 @@
...
+ * @slot header-actions-start - A slot for adding actions or content to the start side of the header.
```
geospatialem, line 59: nitpick - add a reference to the component:

> A slot for adding actions or content to the **starting** side of the **component's** header.

packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,509 @@
...
+ * @slot header-actions-end - A slot for adding actions or content to the end side of the header.
```
geospatialem, line 60: nitpick - similar to above:

> A slot for adding actions or content to the **ending** side of the **component's** header.

packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,509 @@
...
+ * @slot header-content - A slot for adding custom content to the header.
```
geospatialem, line 61: nitpick - similar to above:

> A slot for adding custom content to the **component's** header.

packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,509 @@
...
+ * @slot header-menu-actions - A slot for adding an overflow menu with actions inside a `calcite-dropdown`.
+ * @slot fab - A slot for adding a `calcite-fab` (floating action button) to perform an action.
+ * @slot footer - A slot for adding custom content to the component's footer.
+ * @slot footer-end - A slot for adding a trailing footer custom content.
+ * @slot footer-start - A slot for adding a leading footer custom content.
```
geospatialem, line 66: nitpick - similar to above:

> @slot footer-end - A slot for **adding custom content to the component's trailing footer.**
> @slot footer-start - A slot for **adding custom content to the component's leading footer.**

packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,509 @@
...
+
+  /** When `true`, displays a scrim blocking interaction underneath the component.  */
+  @Prop({ reflect: true }) modal = false;
+
+  @Watch("modal")
+  onModalChange(): void {
+    this.updateOverflowHiddenClass();
+  }
+
+  /** When `true`, displays and positions the component.  */
```
geospatialem, line 156: nitpick: Looks like there's an additional space at the end of this description.

packages/calcite-components/src/components/dialog/readme.md
```diff
@@ -0,0 +1,156 @@
+# calcite-dialog
```
geospatialem, line 1: Would this readme need to be refactored for `dialog`?

packages/calcite-components/src/components/shell/shell.tsx
```diff
@@ -17,6 +17,7 @@ import { CSS, SLOTS } from "./resources";
  * @slot panel-bottom - A slot for adding the bottom `calcite-shell-panel`.
  * @slot center-row - [Deprecated] Use the `"panel-bottom"` slot instead. A slot for adding the bottom `calcite-shell-center-row`.
  * @slot modals - A slot for adding `calcite-modal` components. When placed in this slot, the modal position will be constrained to the extent of the `calcite-shell`.
+ * @slot dialogs - A slot for adding `calcite-modal` components. When placed in this slot, the modal position will be constrained to the extent of the `calcite-shell`.
```
geospatialem, line 20: Would this be `calcite-dialog` and dialog?

> A slot for adding `calcite-dialog` components. When placed in this slot, the dialog position will be constrained to the extent of the `calcite-shell`.

geospatialem (commented):
✨✨ 

Few copyedit suggestions and a few references of `dialog` in place of `modal` for consideration.

packages/calcite-components/src/components/dialog/readme.md
driskull: It should be auto generated at some point.

packages/calcite-components/src/components/shell/shell.tsx
```diff
@@ -151,6 +154,15 @@ export class Shell implements ConditionalSlotComponent {
     this.hasPanelBottom = slotChangeHasAssignedElement(event);
   };
 
+  handleDialogsSlotChange = (event: Event): void => {
+    this.hasDialogs = !!slotChangeHasAssignedElement(event);
+    slotChangeGetAssignedElements(event)?.map((el) => {
+      if (el.nodeName === "CALCITE-DIALOG") {
```
jcfranco, line 160: Nitpick: For consistency, should this use `tagName`?

packages/calcite-components/src/demos/dialog.html
```diff
@@ -0,0 +1,1795 @@
...
+              >Back</calcite-button
+            >
+            <calcite-button slot="footer-end" width="full" scale="m" appearance="outline">Cancel</calcite-button>
+            <calcite-button slot="footer-end" width="full" scale="m">Save</calcite-button>
+          </calcite-dialog>
+
+          <calcite-button data-dialog-ref="js-dialog-s-m" appearance="outline" scale="s"> Scale Medium </calcite-button>
+        </div>
+
+        <!-- small widthScale & large scale -->
```
jcfranco, line 114: Nitpick: can we drop these comments and rely on the dialog headings and `.child` text elements?

packages/calcite-components/src/components/dialog/dialog.tsx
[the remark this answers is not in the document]
jcfranco: One more downside to using `delegatesFocus` is that it prevents blurring any focusable element that has focus by clicking on a non-focusable area within the `delegatesFocus` host (see [example](https://codepen.io/jcfranco/pen/PorGwMj)). This might not be desired in [some scenarios](https://github.com/Esri/calcite-design-system/pull/9701#discussion_r1664651315).

packages/calcite-components/src/components/dialog/dialog.e2e.ts
```diff
@@ -0,0 +1,825 @@
...
+    await page.waitForChanges();
+
+    const dialog = await page.find("calcite-dialog");
+    dialog.setProperty("open", false);
+    await page.waitForChanges();
+
+    expect(mockCallBack).toHaveBeenCalledTimes(2);
+  });
+
+  it("should remain open with rejected 'beforeClose' promise'", async () => {
```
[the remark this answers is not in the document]
driskull: It can't because the two functions are different. One is for rejecting and the other is a mock function.

packages/calcite-components/src/components/dialog/dialog.scss
```diff
@@ -0,0 +1,241 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-dialog-scrim-background-color: Specifies the background color of the component's scrim.
+ * @prop --calcite-dialog-size-x: Specifies the width of the component, using `px`, `em`, `rem`, `vw`, or `%`. Does not exceed the viewport's width - applies when `placement="cover"` if set.
```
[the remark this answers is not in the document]
driskull: This is what @alisonailea told me to use in a previous review.

packages/calcite-components/src/components/dialog/dialog.scss
```diff
@@ -0,0 +1,241 @@
...
+ * @prop --calcite-dialog-size-y: Specifies the height of the component, using `px`, `em`, `rem`, `vh`, or `%`. Does not exceed the viewport's height - applies when `placement="cover"` if set.
+ * @prop --calcite-dialog-panel-footer-space: Specifies the padding of the component's footer.
```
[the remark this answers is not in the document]
driskull: I don't think so because it sets the internal panel variable.

packages/calcite-components/src/components/dialog/dialog.scss
```diff
@@ -0,0 +1,241 @@
...
+
+:host([scale="l"]) {
+  --calcite-internal-dialog-padding: theme("spacing.5");
+  --calcite-internal-dialog-padding-large: theme("spacing.6");
+  --calcite-internal-dialog-title-text: theme("fontSize.3h");
+  --calcite-internal-dialog-content-text: theme("fontSize.1");
+}
+
+.scrim {
+  // todo: remove once tokens update is merged
```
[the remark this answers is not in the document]
driskull: @jcfranco this would be the tokens PR. https://github.com/Esri/calcite-design-system/pull/9816

packages/calcite-components/src/components/dialog/dialog.stories.ts
```diff
@@ -0,0 +1,375 @@
...
+`;
+
+export const withFooter = (): string => html`
+  <calcite-dialog open modal heading="heading" description="description" scale="m" width-scale="s">
+    <calcite-action text="Action" label="Action" slot="${SLOTS.footer}" icon="attachment"></calcite-action>
+    Hello world!
+  </calcite-dialog>
+`;
+
+export const scaleS = (): string => html`
```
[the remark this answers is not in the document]
driskull: We could combine placements and scales but that might be a bit odd.

packages/calcite-components/src/components/dialog/interfaces.ts
```diff
@@ -0,0 +1,9 @@
+export type DialogPlacement =
```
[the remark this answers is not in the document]
driskull: Its not based off `floating-ui`

packages/calcite-components/src/components/modal/modal.tsx
```diff
@@ -56,6 +56,7 @@ let totalOpenModals: number = 0;
 let initialDocumentOverflowStyle: string = "";
 
 /**
+ * @deprecated Use the `calcite-dialog` component instead.
```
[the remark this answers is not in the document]
driskull: 👍

packages/calcite-components/src/demos/dialog.html
driskull: Yeah, they are so redundant

packages/calcite-components/src/components/dialog/dialog.scss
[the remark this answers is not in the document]
driskull: see https://github.com/Esri/calcite-design-system/pull/9751#discussion_r1681404122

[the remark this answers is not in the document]
geospatialem: Small typo:

```suggestion
 * @prop --calcite-dialog-size-x: Specifies the width of the component, using `px`, `em`, `rem`, `vw`, or `%`. Does not exceed the viewport's width - applies when `placement="cover"` is set.
```

packages/calcite-components/src/components/dialog/dialog.scss
```diff
@@ -0,0 +1,241 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-dialog-scrim-background-color: Specifies the background color of the component's scrim.
+ * @prop --calcite-dialog-size-x: Specifies the width of the component, using `px`, `em`, `rem`, `vw`, or `%`. Does not exceed the viewport's width - applies when `placement="cover"` if set.
+ * @prop --calcite-dialog-size-y: Specifies the height of the component, using `px`, `em`, `rem`, `vh`, or `%`. Does not exceed the viewport's height - applies when `placement="cover"` if set.
```
geospatialem, line 8: Small typo:
```suggestion
 * @prop --calcite-dialog-size-y: Specifies the height of the component, using `px`, `em`, `rem`, `vh`, or `%`. Does not exceed the viewport's height - applies when `placement="cover"` is set.
```

geospatialem (approved):
A few small typo fix recommendations, then LGTM! 🚀

driskull:
@SkyeSeitz @ashetland can you review the storybook changes?

packages/calcite-components/src/components/dialog/dialog.scss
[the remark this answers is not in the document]
alisonailea: but is there another footer? It seems unnecessary say it's in the panel element since there is only one footer in the dialog component.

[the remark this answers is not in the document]
macandcheese: I do think we could hide this implementation detail - the fact that it's internally rendering the Panel shouldn't matter to the user.

macandcheese:
For the "slotted custom content" use case - I think we should remove the BG and box shadow set by the Component.

packages/calcite-components/src/components/dialog/dialog.scss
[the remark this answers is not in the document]
driskull: ok i'll update the var name to remove panel.

driskull:
> For the "slotted custom content" use case - I think we should remove the BG and box shadow set by the Component.

@macandcheese are you suggesting removing these:

```scss
.dialog {
  @apply shadow-2-sm
    bg-foreground-1
    ...
    m-6 // this one too??
    rounded // this one too??
}
```

macandcheese:
> > For the "slotted custom content" use case - I think we should remove the BG and box shadow set by the Component.
> 
> @macandcheese are you suggesting removing these:
> 
> ```css-scss
> .dialog {
>   @apply shadow-2-sm
>     bg-foreground-1
>     ...
>     m-6 // this one too??
>     rounded // this one too??
> }
> ```

Yeah - basically just provide the Scrim behind the "empty" container. cc @ashetland / @SkyeSeitz to confirm - but I think the use cases for custom content want to be able to have something like a semi-transparent overlay, some video, storymap, dashboard, or media embed that may have its own bg, etc.,

driskull:
What about the margin? Seems like it could be built in.

macandcheese:
Yeah, I suppose they could just un-set the bg, shadow, etc., via css property as well. Not sure which is more expected once folks are used to using tokens?

driskull:
Checked in updates @macandcheese

AdelheidF:
Do we not need `focusTrapDisabled` in dialog then?, we used it for modal.

driskull:
> Do we not need focusTrapDisabled and outsideCloseDisabled in dialog then?, we used it for modal.

Hey @AdelheidF, we shouldn't need it but if a good use case arises we could add it.

paulcpederson:
The `outsideCloseDisabled` prop is useful when you have something like a multi-step workflow inside a modal. The user loses all their progress if they accidentally click outside. It's been requested we remove the click outside to close behavior multiple times by customers and stakeholders, so I'm adding a +1 to preserve this capability. 

I searched and we only have one se of `focusTrapDisabled`, so I think that one is less important to us.

driskull:
I think the `beforeClose` property can handle that use case and it would also be good to put some kind of notice in the dialog when beforeClose returns false.

paulcpederson:
That's an interesting point. Does the new dialog component have the capability to slot an alert or something in for that purpose. Sort of a hat on a hat as it's a dialog over a dialog.

driskull:
Yep, the dialog has an alerts slot which it inherits from panel.

driskull:
@paulcpederson I think we will possibly add `outsideCloseDisabled`. Its nice to be able to just prevent closing from a scrim. Since with the beforeClose you can't determine what caused the closing.

COV-GIS:
Escape key close also needs to be added.

Use cases for not having the user close except for taking some action within the dialog include disclaimers, confirmations, etc.

I never understood why modal had three different properties for each way to close and not a single property. If the close button is disabled why would the scrim or escape closing be desirable?

It seems to me with dialog if `modal` and `close-disabled` are both true then no button, scrim or escape closing.

PR #9762, opened 2024-07-12
test(panel, flow-item): fix missing scale property in interface

DitwanP:
**Related Issue:** #6980 

## Summary
I missed adding the scale property to the interface so there was a TS error. The error wasn't present for me until I relaunched my IDE. This PR is to fix that error.


driskull (approved):
👍

Issue #7155, opened 2023-06-08
Closable tabs issues


RSantosGIS:
### Actual Behavior

**edit**: please see conversation for description of additional issue with navigation between tabs also breaking, related to this same root cause.

I started using the closable tabs recently and overall I think it's great work.  I observed a few issues during my time getting things setup (this is in React):

- When you close the active tab, the active tab automatically changes (as intended) but this does not fire the onCalciteTabsActivate or onCalciteTabChange events for CalciteTabTitle and CalciteTabNav, respectively
- In my app, the number of tabs is dynamically generated based on a state value.  As you would expect, when a tab gets closed I also update the state.  This causes a re-render event which actually eliminates the closed CalciteTabTitle and CalciteTab from the dom.  However, the internal behavior of CalciteTabs is to just set a 'closed' attribute and hide the tab, not actually unmount it.  This leads to an uncontrolled exception when the parent CalciteTabs component (which does not re-render) attempts to access a property on what is now a non-existent array index in tab.js:

![MicrosoftTeams-image (9)](https://github.com/Esri/calcite-components/assets/36513418/23b63f33-5e8d-4ec0-8214-684bf4a1daf7)

### Expected Behavior

- events should fire for the active tab changing when the active tab is the one closed
- I recommend adding handling to assume that a tab may no longer exist.  You can fix the screenshot I sent above just by doing the following (I tested it by modifying the code in the console):

```
tabIds = tabDomIndexes.reduce((ids, indexInDOM, registryIndex) => {
        if (this.tabs[registryIndex]) {
          ids[indexInDOM] = this.tabs[registryIndex].id;
        }
        return ids;
      }, []);
```
### Reproduction Sample

ArcGIS Knowledge Studio
https://codesandbox.io/s/summer-mountain-y3jcf3?file=/src/App.js

### Reproduction Steps

**To reproduce in [the sandbox](https://codesandbox.io/s/summer-mountain-y3jcf3?file=/src/App.js)** _updated to 1.5.1_:
1. Select tab labeled "Select Me"
2. Close the tab labeled "... then close me"
3. After closing a tab, you'll get the error message referenced in the original issue message
4. After closing the error message, you will see that you cannot navigate to the tab to the right (123) until your first re-click on "Select Me"

To reproduce in Studio:

For the events:
- checkout Knowledge Studio out of our internal repo
- cd /studio
- npm install
- Open the file MainDisplayZone.tsx
[...]

PR #9768, opened 2024-07-12
fix(tabs): handle tab close events that remove the associated tab-title and tab elements from the DOM

eriklharper:
**Related Issue:** #7155

## Summary

This issue resolves some issues that happen when corresponding `calcite-tab` and `calcite-tab-title` elements are removed from the DOM when they are closed.


packages/calcite-components/src/components/tab/tab.tsx
```diff
@@ -81,10 +69,6 @@ export class Tab {
     this.parentTabsEl = this.el.closest("calcite-tabs");
   }
 
-  componentDidLoad(): void {
-    this.calciteInternalTabRegister.emit();
```
jcfranco, line 85: ✨🧹✨

packages/calcite-components/src/components/tabs/tabs.e2e.ts
```diff
@@ -374,15 +374,15 @@ describe("calcite-tabs", () => {
       await page.setContent(html`
         <calcite-tabs>
           <calcite-tab-nav slot="title-group">
-            <calcite-tab-title id="tab-title-1" closable>Tab 1 Title</calcite-tab-title>
-            <calcite-tab-title id="tab-title-2" closable>Tab 2 Title</calcite-tab-title>
-            <calcite-tab-title id="tab-title-3" closable>Tab 3 Title</calcite-tab-title>
-            <calcite-tab-title id="tab-title-4" closable selected>Tab 4 Title</calcite-tab-title>
+            <calcite-tab-title id="tab-title-1" data-name="1" closable>Tab 1 Title</calcite-tab-title>
```
jcfranco, line 377: `tab`s and `tab-title`s already have `id` set. Could you use that instead of adding an extra attribute?

packages/calcite-components/src/components/tabs/tabs.tsx
```diff
@@ -46,92 +47,19 @@ export class Tabs {
    */
   @Prop() bordered = false;
 
-  //--------------------------------------------------------------------------
```
jcfranco, line 49: Thanks for reordering these sections! ✨🏆✨

packages/calcite-components/src/components/tabs/tabs.tsx
```diff
@@ -237,7 +158,62 @@ export class Tabs {
...
+  private setDefaultSlotRef = (el) => (this.slotEl = el);
+
+  //--------------------------------------------------------------------------
+  //
+  //  Lifecycle
+  //
+  //--------------------------------------------------------------------------
+
+  connectedCallback(): void {
+    this.mutationObserver.observe(this.el, { childList: true });
```
jcfranco, line 197: Currently, the observer and slot overlap in terms of functionality, so let's go with slot to keep things simple..

jcfranco (commented):
Looking good, @eriklharper! 😎

packages/calcite-components/src/components/tabs/tabs.tsx
```diff
@@ -140,19 +68,35 @@ export class Tabs {
...
+
   /**
    *
    * Stores an array of ids of `<calcite-tab>`s to match up ARIA attributes.
    */
   @State() tabs: HTMLCalciteTabElement[] = [];
 
+  @Watch("tabs")
+  tabsWatcher(): void {
+    this.registryHandler();
```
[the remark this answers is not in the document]
eriklharper: You read my mind, I was thinking of renaming that function, but decided not to at first to minimize the diff, but I do agree this would be nice to rename it more semantically.  Will do :)

Elijbet (approved):

Elijbet:
Good to merge after a couple of focus behavior issues are resolved, as discussed.

packages/calcite-components/src/components/tabs/tabs.tsx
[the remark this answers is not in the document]
eriklharper: Took a stab at a name: `updateAriaSettings`.

alisonailea (approved):

packages/calcite-components/src/utils/dom.ts
```diff
@@ -565,12 +576,25 @@ export function slotChangeHasAssignedElement(event: Event): boolean {
  * ```
  *
  * @param {Event} event The event.
- * @returns {boolean} Whether the slot has any assigned elements.
+ * @param {string} selector The CSS selector string to filter the returned elements by.
+ * @returns {Element[]} An array of elements.
+ */
+export function slotChangeGetAssignedElements(event: Event, selector?: string): Element[] {
```
jcfranco, line 582: Can you also update this to use a generic?

packages/calcite-components/src/utils/dom.ts
```diff
@@ -565,12 +576,25 @@ export function slotChangeHasAssignedElement(event: Event): boolean {
...
+/**
+ * This helper returns the assigned elements on a `slot` element, filtered by an optional css selector.
+ *
+ * @param {HTMLSlotElement} slot The slot element.
+ * @param {string} selector CSS selector string to filter the returned elements by.
+ * @returns {Element[]} An array of elements.
  */
-export function slotChangeGetAssignedElements(event: Event): Element[] {
-  return (event.currentTarget as HTMLSlotElement).assignedElements({
+export function getSlotAssignedElements<T>(slot: HTMLSlotElement, selector?: string): T[] {
```
jcfranco, line 593: Generic type should be `T extends Element` and the return type `T[] | null`.

packages/calcite-components/src/utils/dom.ts
```diff
@@ -565,12 +576,25 @@ export function slotChangeHasAssignedElement(event: Event): boolean {
...
+  const assignedElements = slot.assignedElements({
     flatten: true,
   });
+  return selector ? (filterElementsByCSSSelector(assignedElements, selector) as T[]) : (assignedElements as T[]);
```
jcfranco, line 597: I _think_ you can avoid all typecasting if you pass T (after the generic type is updated) as the type to supporting functions:

```tsx
return selector ? 
  filterElementsByCSSSelector<T>(assignedElements, selector) : 
  assignedElements;
```

packages/calcite-components/src/utils/dom.ts
```diff
@@ -419,6 +419,17 @@ export function filterDirectChildren<T extends Element>(el: Element, selector: s
...
 }
 
+/**
+ * Filters an array of HTML elements by the provided css selector string.
+ *
+ * @param {Element[]} elements An array of elements, such as one returned by HTMLSlotElement.assignedNodes().
+ * @param {string} selector The CSS selector string to filter the returned elements by.
+ * @returns {Element[]} A filtered array of elements.
+ */
+export function filterElementsByCSSSelector<T extends Element>(elements: Element[], selector: string): T[] {
```
jcfranco, line 429: Suggestion: `filterBySelector`

packages/calcite-components/src/utils/dom.ts
jcfranco, line 593: Let's make sure to add spec tests for the follow-up issue/PR we discussed earlier (updating similar functions with a new selector argument).

jcfranco (approved):
🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊
🧊🚫🧊🧊🧊🚫🧊🧊🚫🚫🧊🧊🚫🚫🚫🧊🧊🚫🚫🚫🧊🚫🚫🚫🚫🧊🚫🧊
🧊🚫🚫🧊🧊🚫🧊🚫🧊🧊🚫🧊🧊🚫🧊🧊🚫🧊🧊🧊🧊🚫🧊🧊🧊🧊🚫🧊
🧊🚫🧊🚫🧊🚫🧊🚫🧊🧊🚫🧊🧊🚫🧊🧊🚫🧊🧊🧊🧊🚫🚫🚫🧊🧊🚫🧊
🧊🚫🧊🧊🚫🚫🧊🚫🧊🧊🚫🧊🧊🚫🧊🧊🚫🧊🧊🧊🧊🚫🧊🧊🧊🧊🧊🧊
🧊🚫🧊🧊🧊🚫🧊🧊🚫🚫🧊🧊🚫🚫🚫🧊🧊🚫🚫🚫🧊🚫🚫🚫🚫🧊🚫🧊
🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊

packages/calcite-components/src/components/tabs/tabs.tsx
```diff
@@ -237,7 +147,59 @@ export class Tabs {
...
+      nav.scale = scale;
+    }
+
+    Array.from(this.el.querySelectorAll("calcite-tab")).forEach((tab: HTMLCalciteTabElement) => {
+      if (tab.parentElement === this.el) {
+        tab.scale = scale;
+      }
+    });
+
+    Array.from(this.el.querySelectorAll("calcite-tab-nav > calcite-tab-title")).forEach(
```
driskull, line 169: nitpick: since `nav` is already queries for above. could you do `Array.from(nav.querySelectorAll("calcite-tab-title")).forEach(`

packages/calcite-components/src/utils/dom.ts
driskull: Do we really need this method? What is the use case for getting assigned elements from a slot at a time other than an `onSlotChange` event? Wouldn't this kind of promote getting assigned elements at a time other than its really necessary?

To me, it seems like functions like `updateAriaSettings` shouldn't have to get the assigned elements every time. Those elements should be stored during an onSlotChange event for the slot.

packages/calcite-components/src/utils/dom.ts
```diff
@@ -419,6 +419,17 @@ export function filterDirectChildren<T extends Element>(el: Element, selector: s
   return Array.from(el.children).filter((child): child is T => child.matches(selector));
 }
 
+/**
+ * Filters an array of HTML elements by the provided css selector string.
+ *
+ * @param {Element[]} elements An array of elements, such as one returned by HTMLSlotElement.assignedNodes().
```
driskull, line 425: Should this say `HTMLSlotElement.assignedNodes().` or `HTMLSlotElement.assignedElements().`? If it should be nodes, I don't think `.matches` would work for TextContent nodes.

packages/calcite-components/src/utils/dom.ts
eriklharper: Works for the `filter*` function call, but not for the assignedElements variable:

image

image

eriklharper:
@driskull:

> What is the use case for getting assigned elements from a slot at a time other than an onSlotChange event?

Use case is when you have a `ref` to a `<slot>` element and want to get its `assignedElements` on-the-fly so that you don't have to store that information in a state variable.  This allows you to use the DOM as the source of truth of how many tabs you have instead of having to manage a separate state variable to keep track of them.

This fix is attempting to better gracefully handle the removal of `tab-title` and `tab` dom nodes, which are being removed in a `calciteTabClose` event listener, which fires before this [`getTabTitleById` async process](https://github.com/Esri/calcite-design-system/blob/3243cf9a681f360ec4b3470433a330bcb777ec40/packages/calcite-components/src/components/tab-nav/tab-nav.tsx#L654-L656) completes, so an uncaught exception would happen when a removed dom node was attempting to access its `parentElement` property, which is `undefined` because it was removed from the document.

packages/calcite-components/src/components/tabs/tabs.tsx
eriklharper: This code is carried over from `dev`, just relocated in the file.

packages/calcite-components/src/utils/dom.ts
eriklharper: Done.  Added a direct element selector, one with no selector, and a test for empty results and a css class selector.  That should cover the bases ⚾

Issue #9769, opened 2024-07-12
panel, flow-item: add beforeClose property


driskull:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Add beforeClose property to both panel and flow-item in order to support stopping a panel from being closed.

### Acceptance Criteria

panel, flow-item: add beforeClose property

### Relevant Info

Required for dialog component

### Which Component

panel, flow-item

### Example Use Case

```js
panel.beforeClose = () => Promise.reject();
```

### Priority impact

impact - p1 - need for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

DitwanP:
🍡 Verified on `2.11.0-next.9`

PR #9770, opened 2024-07-12
feat(panel, flow-item): add beforeClose property

driskull:
**Related Issue:** #9769

## Summary

- adds `beforeClose` property to `panel` and `flow-item`
- allows close to be prevented via a function returning a promise that could be rejected
- adds tests

### DIFF
--- a/packages/calcite-components/src/components/panel/panel.e2e.ts
+++ b/packages/calcite-components/src/components/panel/panel.e2e.ts
@@ -36,6 +36,10 @@ describe("calcite-panel", () => {
   describe("defaults", () => {
     defaults("calcite-panel", [
       {
+        propertyName: "beforeClose",
+        defaultValue: undefined,
+      },
+      {
         propertyName: "widthScale",
         defaultValue: undefined,
       },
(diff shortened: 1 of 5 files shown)


benelan (approved):
:+1:

packages/calcite-components/src/components/panel/panel.tsx
```diff
@@ -288,16 +298,36 @@ export class Panel
...
+  open = (): void => {
+    this.isClosed = false;
+  };
+
+  close = async (): Promise<void> => {
+    const beforeClose = this.beforeClose ?? (() => Promise.resolve());
+
+    try {
+      await beforeClose();
+    } catch (_error) {
```
eriklharper, line 320: Looks like `_error` here is unused.

driskull: Yep, thats correct. Its underscored because its unused but a required param.

packages/calcite-components/src/components/panel/panel.e2e.ts
```diff
@@ -129,6 +133,49 @@ describe("calcite-panel", () => {
...
+
+    const mockCallBack = jest.fn().mockReturnValue(() => Promise.reject());
+    await page.exposeFunction("beforeClose", mockCallBack);
+
+    await page.setContent(`<calcite-panel closable></calcite-panel>`);
+
+    await page.$eval(
+      "calcite-panel",
+      (el: HTMLCalcitePanelElement) =>
+        (el.beforeClose = (window as typeof window & Pick<typeof el, "beforeClose">).beforeClose),
```
jcfranco, line 147: For consistency, could you use our [custom utility type](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/tests/utils.ts#L9) to define custom props on `window` ([example](https://github.com/Esri/calcite-design-system/blob/60ab0dcf5e571ede472f62f751ef4a380181584f/packages/calcite-components/src/components/button/button.e2e.ts#L658-L660))?

driskull: Added as part of https://github.com/Esri/calcite-design-system/pull/9778

Issue #3695, opened 2021-12-13
Enhancement: Subtitle/description for combobox-item


jgibson02:
### Description
Provide a subtitle/description prop for displaying a secondary label on a combobox-item

### Acceptance Criteria 
- [ ] A prop is provided to render a secondary label within a combobox-item

#### Which Component
combobox-item

#### Example Use Case
In ArcGIS Velocity, we have a spatial reference selector which has ~9000 items, and users often are familiar with choosing a spatial reference by either the text label (e.g. "GCS WGS 1984") or by its numeric WKID (e.g. "4326"), so we used the `subtitle` prop that calcite-react `MenuItem`s provide to display the numeric WKID on the menuitem, and filtering automatically filters against that subtitle as well.

![image](https://user-images.githubusercontent.com/5069711/145895069-f03dc354-5140-401d-b42e-cdf747357ee7.png)

Due to a bug in calcite-react where scrolling while using the `appendToBody` and `virtualized` props together will cause the dropdown to close, we are looking to switch to using calcite-components combobox instead, but not having this subtitle is a serious UX drawback.

macandcheese:
This would probably be best implemented as a `content-end` slot.

hccampos:
We should also have the ability to provide a value to be displayed in the input when the combobox is closed. In this spatial reference case, it could display the WKID or some other shorter representation of the spatial reference.

In our case, we bumped into this while working on a Timezone picker where we have a long name with the IANA code and the UTC offset, but we don't want to display all that information in the input itself, since it would make the input too large.

ashetland:
Designs for a `description` prop and a `content-end` slot are ready to go in [this Figma file](https://www.figma.com/file/ujZYHdimnVC3yA2PyYOH85/Epic%3A-Combobox-Redesign?type=design&node-id=115-76403&mode=dev). Also outlined is a proposed `value-short` prop to address the needs in [this comment](https://github.com/Esri/calcite-components/issues/3695#issuecomment-1499257085).

ashetland:
A round of feedback has resulted in some fine-tuning to do on the abbreviated value prop and its behaviors.

ashetland:
Revised designs ready in [this Figma file](https://www.figma.com/file/ujZYHdimnVC3yA2PyYOH85/Epic%3A-Combobox-Redesign?type=design&node-id=142-78995&mode=dev). 

[...]

PR #9771, opened 2024-07-12
feat(combobox, combobox-item): add `description`, `shortHeading` props and `content-end` slot

jcfranco:
**Related Issue:** #3695

## Summary

This adds the following enhancements to `combobox`/`combobox-item`:

* `description` prop - displays description below label
* `shortHeading` prop - displays short version of the heading (label) in selection
* `content-end` slot - enables slotting non-interactive elements after the item's content

**Note**: the new props are filterable and also participate in visual matching

### DIFF
--- a/packages/calcite-components/src/components/combobox-item/combobox-item.scss
+++ b/packages/calcite-components/src/components/combobox-item/combobox-item.scss
@@ -6,6 +6,7 @@
   --calcite-combobox-item-spacing-unit-s: theme("spacing.1");
   --calcite-combobox-item-spacing-indent: theme("spacing.2");
   --calcite-combobox-item-selector-icon-size: theme("spacing.4");
+  --calcite-combobox-item-description-font-size: var(--calcite-font-size--3);
 }
 
 .scale--m {
@@ -48,20 +49,22 @@ ul:focus {
 
 .label {
   @apply text-color-3
-    focus-base
-    relative
-    box-border
-    flex
-    w-full
-    min-w-full
-    cursor-pointer
-    items-center
-    no-underline
-    duration-150
-    ease-in-out;
+  focus-base
+  relative
+  box-border
+  flex
+  w-full
+  min-w-full
+  cursor-pointer
+  items-center
+  no-underline
+  duration-150
+  ease-in-out;
   @include word-break();
   padding-block: var(--calcite-combobox-item-spacing-unit-s);
   padding-inline: var(--calcite-combobox-item-spacing-unit-l);
+  justify-content: space-around;
+  gap: var(--calcite-combobox-item-spacing-unit-l);
 }
 
 :host([disabled]) .label {
@@ -85,11 +88,6 @@ ul:focus {
     shadow-none;
 }
 
-.title {
-  padding-block: 0;
-  padding-inline: var(--calcite-combobox-item-spacing-unit-l);
-}
-
 .icon {
   @apply inline-flex
     opacity-0
@@ -104,7 +102,6 @@ ul:focus {
 
 .icon--custom {
   margin-block-start: -1px;
-  padding-inline-start: var(--calcite-combobox-item-spacing-unit-l);
   @apply text-color-3;
 }
 
@@ -139,4 +136,26 @@ ul:focus {
   font-weight: var(--calcite-font-weight-bold);
   color: var(--calcite-color-text-1);
   background-color: var(--calcite-color-foreground-current);
+}
+
+.center-content {
+  display: flex;
+  flex-direction: column;
+  flex-grow: 1;
+  padding-block: 0;
+}
+
+.description {
+  font-size: var(--calcite-combobox-item-description-font-size);
+}
+
+:host([selected]),
+:host(:hover) {
+  .description {
+    color: var(--calcite-color-text-2);
+  }
+}
+
+.short-text {
+  color: var(--calcite-color-text-3);
 }
--- a/packages/calcite-components/src/components/combobox-item/combobox-item.scss
+++ b/packages/calcite-components/src/components/combobox-item/combobox-item.scss
@@ -6,7 +6,7 @@
   --calcite-combobox-item-spacing-unit-s: theme("spacing.1");
   --calcite-combobox-item-spacing-indent: theme("spacing.2");
   --calcite-combobox-item-selector-icon-size: theme("spacing.4");
-  --calcite-combobox-item-description-font-size: var(--calcite-font-size--3);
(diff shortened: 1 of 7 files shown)


packages/calcite-components/src/components/combobox-item/combobox-item.tsx
```diff
@@ -76,6 +82,15 @@ export class ComboboxItem implements ConditionalSlotComponent, InteractiveCompon
...
   @Prop({ reflect: true }) textLabel!: string;
 
+  /**
+   * The component's short label.
+   *
+   * When provided, the short label will be displayed in the component's selection.
+   *
+   * It is recommended to use 5 chars or less.
+   */
+  @Prop({ reflect: true }) textLabelShort: string;
```
driskull, line 92: Maybe we can come up with a better name for this prop?

Something like `snippet` or `summary`? Or `selected-label`?

driskull: I think at some point we should rename `textLabel` to just `label` as well.

jcfranco: > Something like snippet or summary? Or selected-label?

When specified this prop shows an abbreviated version of the label that is used in selection. The suggested names do not quite reflect this.

> I think at some point we should rename textLabel to just label as well.

Agreed, but until then it might help associate both props by matching the existing name. There were other suggestions in the design spec, but `textLabelShort` was the shortest and most expressive (IMO). Open to alternatives. cc @SkyeSeitz @ashetland

ashetland: Sadly, I've got nothin' better than `textLabelShort`. 

Long-term I do like the idea of renaming the main two props to `label` and `description` across the system. It's not super consistent atm.

driskull: Just throwing this out there:

Instead of adding another prop we plan on renaming in the future, why not just deprecate `textLabel` for `label` and add `description`?

jcfranco: Assuming we were using `heading`/`description`. Would we go with `shortHeading`, `headingShort` or something else?

driskull: subheading?

jcfranco: > Instead of adding another prop we plan on renaming in the future, why not just deprecate textLabel for label

We'd want to use `heading` since `label` is mostly used for accessible text, right?

Also wanted to point out that my PR title isn't quite accurate. It's a short heading/label we're adding, not a description. Will update.

jcfranco: @driskull [This example](https://codepen.io/jcfranco/pen/QWXjMwZ?editors=1010) shows how the prop comes into play.

driskull: Yeah I think something like `shortHeading` should be good. 👍

SkyeSeitz: Agree with moving towards a consistent `heading` & `description` paradigm across components accompanied by `shortHeading` when relevant.

jcfranco: Sounds good. Thanks for the discussion, y'all! ✨💪✨

I'll submit an issue/PR to deprecate `textLabel` separately since it is out of scope.

packages/calcite-components/src/components/combobox-item/combobox-item.tsx
```diff
@@ -110,11 +120,16 @@ export class ComboboxItem implements ConditionalSlotComponent, InteractiveCompon
...
-   * Specifies the size of the component inherited from the `calcite-combobox`, defaults to `m`.
+   * The component's short heading.
    *
-   * @internal
+   * When provided, the short heading will be displayed in the component's selection.
+   *
+   * It is recommended to use 5 characters or fewer.
    */
-  @Prop() scale: Scale = "m";
+  @Prop({ reflect: true }) shortHeading: string;
```
driskull, line 129: Should we have `reflects` tests for description and shortHeading? Probably don't need `defaults` since these are both undefined by default.

packages/calcite-components/src/components/combobox/combobox.stories.ts
```diff
@@ -823,7 +884,12 @@ export const filterHighlighting = (): string => html`
     </calcite-combobox-item>
     <calcite-combobox-item value="Flowers" text-label="Flowers">
       <calcite-combobox-item value="Daffodil" text-label="Daffodil"></calcite-combobox-item>
-      <calcite-combobox-item value="Black Eyed Susan" text-label="Black Eyed Susan"></calcite-combobox-item>
+      <calcite-combobox-item
+        value="Black Eyed Susan"
```
driskull, line 888: 👁️‍🗨️ 👀

driskull (approved):
Beautiful 😍

packages/calcite-components/src/components/combobox-item/combobox-item.tsx
jcfranco: I'll submit a separate PR for this since the test doesn't have `reflects` set up yet.

packages/calcite-components/src/components/combobox-item/combobox-item.scss
```diff
@@ -140,3 +135,26 @@ ul:focus {
...
+
+.description {
+  font-size: var(--calcite-combobox-item-description-font-size);
+  font-weight: var(--calcite-font-weight-normal);
+}
+
+:host([selected]),
+:host(:hover) {
+  .description {
+    color: var(--calcite-color-text-2);
```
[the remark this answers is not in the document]
jcfranco: `description` might. WDTY @ashetland?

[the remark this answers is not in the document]
jcfranco: If we need a custom CSS prop, I'll add it in the component tokens PR.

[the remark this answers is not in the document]
ashetland: That would make sense for the short text and I'm guessing we're doing that for description text across components?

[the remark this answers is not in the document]
jcfranco: I believe so. I'll double check and add it to #8594.

Issue #9540, opened 2024-06-07
[Tile] slots are start aligned on center alignment component


SkyeSeitz:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When Tile's alignment prop = "`center`", slotted content aligns to the _start_ in Tile's `content-top` & `content-bottom` slots. 

![image](https://github.com/Esri/calcite-design-system/assets/29716057/493cb382-7566-4213-a84b-d8ac818b7cda)

### Expected Behavior

When Tile's alignment prop = "`center`", slotted content aligns to the _center_ in Tile's `content-top` & `content-bottom` slots. 

<img width="162" alt="image" src="https://github.com/Esri/calcite-design-system/assets/29716057/412f6471-11f3-453e-a5cd-7cc54a7942e8">

![image](https://github.com/Esri/calcite-design-system/assets/29716057/74ad4701-bf18-41d3-894c-dbe6790e26f4)

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/vYwJGJm?editors=1000

### Reproduction Steps

See content-bottom slots with start aligned chips.

### Reproduction Version

2.9.0

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

DitwanP:
🍡 Verified on `2.11.0-next.6`

SkyeSeitz:
> @aPreciado88 Noticing the chips are centered but the text is still not. @SkyeSeitz The text is supposed to be centered as well right? https://codepen.io/Ditwan-Price/pen/vYqOJQz?editors=1000

Correct - text should be centered. Thanks for catching.

DitwanP:
🍡 Verified on `2.11.0-next.7` 
https://codepen.io/Ditwan-Price/pen/vYqOJQz

PR #9773, opened 2024-07-12
fix(tile): center align slot's text when alignment is equal to center

aPreciado88:
**Related Issue:** [#9540](https://github.com/Esri/calcite-design-system/issues/9540)

## Summary
updates the `tile` component to center align the slot's text when the alignment prop is equal to "center"

### DIFF
--- a/packages/calcite-components/src/components/tile/tile.scss
+++ b/packages/calcite-components/src/components/tile/tile.scss
@@ -103,6 +103,9 @@
   .icon {
     align-self: center;
   }
+  .text-content-container {
+    justify-content: center;
+  }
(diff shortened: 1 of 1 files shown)


driskull (approved):
Code changes look good. Just make sure this triggers a screenshot test change. If not, a screenshot test should be added via a story.

aPreciado88:
@driskull Thank you for reviewing! I added a screenshot [test](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/components/tile/tile.stories.ts#L775) in a previews PR.

Issue #9513, opened 2024-06-04
Revisit event names emitted by openCloseComponent util


Elijbet:
### Description

Revisit event names emitted by openCloseComponent util.

Emitting `open/close` events when `expanded/collapsed` might be misleading and won't work for cases where we have both props (e.g., list, action-menu).

As an alternative consider `open/close` (or some other name) as the main events and add a prop to indicate which property it relates to.

Consider expanding the scope of openCloseComponent to something like transitionEventComponent that would handle multiple types of transitions from open/expanded and whatever else may need it. 

Blocked issues: #6017, #4544, #11227

### Proposed Advantages

 More clarity and accuracy in event naming, and compatibility with cases where both `expanded/collapsed` and `open/closed` props are present.

### Which Component

openCloseComponent util

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

geospatialem:
Blocked issues: #6017, #4544

Elijbet:
This has been addressed. We now emit `expanded`/`collapsed` end-state events for expandable components, as implemented in [Emit expanded/collapsed end-state events only for expandable components #12565](https://github.com/Esri/calcite-design-system/issues/12565).
Myself, @driskull, and  @ashetland discussed naming and applied the decision in the PR above.

This is an intermediate solution until the Animation Effort finalizes a design decision on transitions for `expanded`/`collapsed`, and closed (no-toggle). If transitions are added, all relevant components will be updated to use the utility for transition-based event emission.

Issue #9511, opened 2024-06-04
Expand openClose commonTest util to account for components with closed/collapsed props


Elijbet:
### Priority impact

p - medium

### Test type

e2e

### Which Component(s)

openClose commonTest util

### Unstable Tests

Expand `openClose` commonTest util to account for components with `closed/collapsed` props in addition to `open/expanded`. `openPropName` currently only checks for either `open` or `expanded`.
```
  /**
   * Toggle property to test. Currently, either "open" or "expanded".
   */
  openPropName?: string;
 ```
`tip, chip, and flow-item` that use `closed` don't have `openClose` tests set up for them. Neither do `panel` and `shell-panel` that use `collapsed`. `flow-item` has both `collapsed` and `close`.openClose commonTest util

### Test error, if applicable

N/A

### PR skipped, if applicable

#N/A

DitwanP:
Blocked by stage 2 of the [Unified Animations](https://github.com/Esri/calcite-design-system/issues/12669#top) effort.

PR #9775, opened 2024-07-14
feat(`accordion-item`, `action-bar`, `action-group`, `action-pad`, `tree-item`): add open, close, beforeOpen, and beforeClose events for consistent event handling

Elijbet:
**Related Issue:** ##4544

## Summary
Implement `OpenCloseComponent` with associated events to `accordion-item`, `action-item`, `action-group`, `action-pad`, `tree-item`.


Elijbet:
`Action-menu` having both `open` and `expanded` props presents a bit of an issue.

`action-menu` has an internal `popover`, so it should forward its own events to `popover`. In the case of `open` it’s `popove`r that transitions.
 
eg. `calciteActionMenuBeforeOpen => calciteBeforeOpen` (for `popover`)

But in case of `expanded` it’s the `action-menu` that transitions, so that would be an openClose component itself.
Since we don't have any actionMenuBeforeExpand, we should address an issue where we revisit openClose utility to add handling of the expanded prop to it.

These are some relevant issues listed below. Our latest suggestion is to "expand" the scope of openCloseComponent to something like transitionEventComponent, which would handle multiple types of transitions from open/expanded and whatever else may need it. 

Expand openClose commonTest util to account for components with closed/collapsed props #9511
[Revisit event names emitted by openCloseComponent util](https://github.com/Esri/calcite-design-system/issues/9513#top)#9513

Elijbet:
The PRs can be closed as it is outdated. 

These components now use `expanded`/`collapsed` rather than `open`/`closed` for state, per [Audit components open/closed props for consistency #6473](https://github.com/Esri/calcite-design-system/issues/6473). 

They already emit `expanded`/`collapsed` end-state events for expandable components, as implemented in [Emit expanded/collapsed end-state events only for expandable components #12565](https://github.com/Esri/calcite-design-system/issues/12565). 

This is an intermediate solution until the Animation Effort finalizes a design decision on transitions for expanded`/`collapsed`, and `closed` (no-toggle). If transitions are added, all relevant components will be updated to use the utility for transition-based event emission.

Issue #9772, opened 2024-07-12
panel: add alerts slot


driskull:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

add alerts slot to panel to constrain alerts to the panel in question

### Acceptance Criteria

alerts can be slotted into a panel and contained within that panel

### Relevant Info

https://github.com/Esri/calcite-design-system/issues/7886#issuecomment-2224099221

### Which Component

panel

### Example Use Case

```html
<calcite-panel>
<calcite-alert slot="alerts"></calcite-alert>
</calcite-panel>
```

### Priority impact

impact - p1 - need for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

DitwanP:
@driskull Is this how it's intended to look or is it supposed to be above the buttons? 
![Screenshot 2024-07-16 at 10 59 06 AM](https://github.com/user-attachments/assets/50ceb8e0-79eb-4216-b591-39eefbd75feb)

driskull:
Yes, that is how I intended, it is just constrained inside the panel. Maybe we should ask @SkyeSeitz and @ashetland ?

SkyeSeitz:
Is the Alert slot always tied to the content inside the bounds of Panel's default slot? Or would that ever be too narrow? 
<img width="1344" alt="image" src="https://github.com/user-attachments/assets/f0573a6c-33f4-4e10-99e3-cff9104fea21">

driskull:
My thinking was that it might be too narrow in some cases.

SkyeSeitz:
I assumed the same and defaulted to positioned to the Panel itself. I am fine with that if no one else has any objections.

ashetland:
This may be fine as a starting point and could be updated after feedback. My initial reaction is it does feel awkward covering the footer like that and might be nice if it had some margin to force it higher? But as I said, could wait for feedback on it.

driskull:
Let me know what you prefer. I can make it appear over the content if that's the way we want to go.

ashetland:
"Wait for feedback" is the consensus. ✅

DitwanP:
🍡Verified on `2.11.0-next.11`

PR #9778, opened 2024-07-15
feat(panel): add alerts slot

driskull:
**Related Issue:** #9772

## Summary

- add `alerts` slot to panel
- rename existing internal property `slottedInShell` to `embedded` for greater use beyond just shell
- add e2e test
- add screenshot test

### DIFF
--- a/packages/calcite-components/src/components/alert/alert.e2e.ts
+++ b/packages/calcite-components/src/components/alert/alert.e2e.ts
@@ -11,6 +11,10 @@ describe("defaults", () => {
       propertyName: "autoCloseDuration",
       defaultValue: "medium",
     },
+    {
+      propertyName: "embedded",
+      defaultValue: false,
+    },
   ]);
 });
 
--- a/packages/calcite-components/src/components.d.ts
+++ b/packages/calcite-components/src/components.d.ts
@@ -510,6 +510,10 @@ export namespace Components {
          */
         "autoCloseDuration": AlertDuration;
         /**
+          * This internal property, managed by a containing calcite-shell, is used to inform the component if special configuration or styles are needed
+         */
+        "embedded": boolean;
+        /**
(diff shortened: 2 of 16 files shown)


packages/calcite-components/src/components/flow-item/flow-item.e2e.ts
```diff
@@ -327,4 +331,21 @@ describe("calcite-flow-item", () => {
     expect(toggleSpy).toHaveReceivedEventTimes(1);
     expect(await panel.getProperty("closed")).toBe(true);
   });
+
+  it("should set embedded on slotted alerts", async () => {
```
jcfranco, line 335: This is testing an internal prop. If screenshot tests cover this, we should drop this (and similar) tests.

packages/calcite-components/src/components/alert/alert.e2e.ts
```diff
@@ -11,6 +11,10 @@ describe("defaults", () => {
       propertyName: "autoCloseDuration",
       defaultValue: "medium",
     },
+    {
+      propertyName: "embedded",
```
jcfranco, line 15: We should avoid testing internal props unless there's a strong use case.

jcfranco (approved):
Aside from comments on avoiding testing internals, this LGTM!

packages/calcite-components/src/components/flow-item/flow-item.e2e.ts
driskull: Screenshot tests do cover this, but I do think this adds some protection.

packages/calcite-components/src/components/alert/alert.e2e.ts
driskull: I know it's an internal prop, but we should still test it to expect it to be false by default IMO. This adds protection from someone removing it accidentally or its value being changed.

PR #9781, opened 2024-07-15
chore(themed): add theme option override

alisonailea:
## Summary

Adds a `expectOverride` optional parameter and better documentation for the themed test helper.

expectOverride should be used to confirm edge cases where a static value should override the theme pattern. For example when a component appearance is set to "transparent" it should continue to have a background-color of "transparent" even if the `--calcite-component-background-color` is set.


packages/calcite-components/src/tests/commonTests/themed.ts
```diff
@@ -29,43 +29,51 @@ export type ComponentTestTokens = Record<CalciteCSSCustomProp, TestSelectToken |
...
+ *      "--calcite-token-test-multiple-applications-of-token": [
+ *        {
+ *          targetProp: "borderLeftColor",
+ *        },
+ *        {
+ *          shadowSelector: "calcite-action",
+ *          targetProp: "--calcite-action-border-color",
+ *        },
+ *        {
+ *          // added to demonstrate pseudo-element support
```
jcfranco, line 72: Nitpick: since the example now uses fake CSS props, we can remove this comment.

jcfranco (approved):

jcfranco:
Before merging, can you add a bit more detail on the typical `expectOverride` use case?

Issue #9505, opened 2024-06-04
bug: Missing @types/sortablejs dependency


maxpatiiuk:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Calcite typings are using types from the Sortablejs library, for example in this file:

```tsx
// node_modules/@esri/calcite-components/dist/types/components/value-list/value-list.d.ts
import Sortable from "sortablejs";
...
```

Yet, `@types/sortablejs` is not included as a "dependency" (it's as devDependency instead, and in the root monorepo package.json rather than calcite-components's package.json)

This cause TypeScript errors for consumer of calcite package when they try to typecheck their app if they have `"skipLibCheck":false` in their tsconfig.json (otherwise, the type would get silently resolved to `any` - not ideal)

```yaml
> npx tsc --noEmit
../../../node_modules/@esri/calcite-components/dist/types/components/value-list/value-list.d.ts:2:22 - error TS7016: Could not find a declaration file for module 'sortablejs'. '/Users/mak13180/site/esri/arcgis-web-components/node_modules/sortablejs/Sortable.min.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/sortablejs` if it exists or add a new declaration (.d.ts) file containing `declare module 'sortablejs';`

2 import Sortable from "sortablejs";
                       ~~~~~~~~~~~~

../../../node_modules/@esri/calcite-components/dist/types/utils/sortableComponent.d.ts:1:22 - error TS7016: Could not find a declaration file for module 'sortablejs'. '/Users/mak13180/site/esri/arcgis-web-components/node_modules/sortablejs/Sortable.min.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/sortablejs` if it exists or add a new declaration (.d.ts) file containing `declare module 'sortablejs';`

1 import Sortable from "sortablejs";
```

### Expected Behavior

Since Calcite's types rely on Sortablejs types, `@types/sortablejs` must be made into a dependency

### Reproduction Sample

https://github.com/maxpatiiuk/calcite-missing-dependency

### Reproduction Steps

1. Checkout the repository - `git clone https://github.com/maxpatiiuk/calcite-missing-dependency`
2. Run `npm install`
3. Run `npx tsc`

### Reproduction Version

2.0.0

### Relevant Info

See relevant articles:
[...]

PR #9786, opened 2024-07-16
fix(deps): move @types/sortablejs as a dependency so the public types resolve properly

jcfranco:
**Related Issue:** #9505

## Summary

This fixes a build issue for consumers unless @types/sortablejs is installed locally.


benelan (approved):
:+1:

driskull (approved):

Issue #7513, opened 2023-08-11
[Shell Panel] Add `displayMode` value that matches `Shell Center Row: detached` display


macandcheese:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

As part of the story to deprecate Shell Center Row - we need to add a `displayMode` to Shell Panel with additional capability - to have the entire Shell Panel "float" away from the page border.

### Acceptance Criteria

A new `displayMode` is added to Shell Panel. Propose to call this mode "float", and change the current "float" to "float-detach" to more clearly convey the difference. Open to naming ideas.

Current Shell Center Row:
![Screenshot 2023-08-11 at 3 26 49 PM](https://github.com/Esri/calcite-design-system/assets/4733155/56da101a-e2bb-41bb-a8c7-e9b82fa276c4)

Proposed in Shell Panel:

![Screenshot 2023-08-11 at 3 33 05 PM](https://github.com/Esri/calcite-design-system/assets/4733155/5a0c4d43-e61d-48ab-a7f1-45435ce36225)

Note - there are some margin differences we can work on here - also, this requires the user to set their "layout" in the Shell Panel to "vertical" and adjust width, maybe that can be smoothed. Potentially - there could be more "position" properties than start / end and it could be "top / bottom / start / end". We could use the Shell Panel's slot to set which css var are being used instead of "layout". But, that's a bit tangential.

### Relevant Info

This would allow us to deprecate Shell Center Row as this is the last remaining functionality not offered by Shell Panel. Currently, this would have the same oddities "float" does without `content-behind` set on a parent Shell, but they can be improved together in the future.

Maybe we could consider renaming "overlay" to "dock-overlay" to match the thing-modifier format?

cc @asangma @mitc7862 

### Which Component

Shell Panel

### Example Use Case

As a designer - I want to have a Panel + Action Bar within a Shell Panel "float" above the map without having the Action Bar attached to the edge of the container. Currently, I can't achieve this with built in styles in Shell Panel, but I could with Shell Center Row.

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

Calcite (design)

macandcheese:
cc @brittneytewks @geospatialem - prioritizing this for next available milestone could help us deprecate Shell Center Row in the following release.

[...]

PR #9795, opened 2024-07-16
feat(shell-panel): deprecate float displayMode and add float-content and float-all

josercarcamo:
**Related Issue:** #7513 

## Summary
Replaced display mode "float" with "float-content" and added "float-all". Deprecated "float".


packages/calcite-components/src/components/action-bar/action-bar.tsx
```diff
@@ -342,8 +342,32 @@ export class ActionBar
       expanded,
       overflowCount,
     });
+
+    this.removeOverflowGroups(height, actionGroups, layout);
   }, overflowActionsDebounceInMs);
 
+  private removeOverflowGroups(
```
driskull, line 349: @josercarcamo can you elaborate on what this is doing in the PR body?

packages/calcite-components/src/components/shell-panel/shell-panel.scss
```diff
@@ -48,21 +48,23 @@
...
 }
 
+:host([layout="vertical"][display-mode="float"]) .content,
+:host([layout="vertical"][display-mode="content-detached"]) .content {
+  flex: 2;
+}
+
 :host([layout="vertical"][width-scale="s"]) .content {
   --calcite-shell-panel-width-internal: var(--calcite-shell-panel-width, 12vw);
-  --calcite-shell-panel-max-width-internal: var(--calcite-shell-panel-max-width, 300px);
```
driskull, line 53: question: Why are these max widths being removed? There's still a reference to one of them in the CSS as well.

packages/calcite-components/src/components/shell-panel/shell-panel.scss
```diff
@@ -101,6 +103,11 @@
   * {
     @apply box-border;
   }
+
+  &.float {
+    margin-block: 0.5rem;
```
driskull, line 108: question: this spacing would need to use a calcite var token I believe.

packages/calcite-components/src/components/shell-panel/shell-panel.tsx
```diff
@@ -81,17 +81,17 @@ export class ShellPanel implements ConditionalSlotComponent, LocalizedComponent,
    *
    * `"overlay"` displays at full height on top of center content, and
    *
-   * `"float"` does not display at full height with content separately detached from `calcite-action-bar` on top of center content.
+   * `"content-detached"` does not display at full height with content separately detached from `calcite-action-bar` on top of center content.
```
driskull, line 84: fixme: This would be a breaking change since "float" no longer functions. If we need to add "content-detached" we should do so and not remove "float". "float" should be deprecated and removed at a major release.

packages/calcite-components/src/components/shell-panel/shell-panel.tsx
```diff
@@ -129,12 +129,12 @@ export class ShellPanel implements ConditionalSlotComponent, LocalizedComponent,
   }
 
   /**
-   * Specifies the component's position. Will be flipped when the element direction is right-to-left (`"rtl"`).
+   * Specifies the action bar's position. Will be flipped when the element direction is right-to-left (`"rtl"`).
    */
-  @Prop({ reflect: true }) position: Extract<"start" | "end", Position> = "start";
+  @Prop({ reflect: true }) actionBarPosition: Extract<"start" | "end", Position> = "start";
```
driskull, line 134: fixme: This is a breaking change as well. We should not rename this. Is there a good reason to rename this property? Its intended as the position of the component, not the action-bar. Can you elaborate on this change?

packages/calcite-components/src/components/action-bar/action-bar.tsx
```diff
@@ -342,8 +342,32 @@ export class ActionBar
...
+    layout: "horizontal" | "vertical",
+  ) {
+    let groupHeight = 0;
+    if (layout === "vertical") {
+      actionGroups.forEach((group) => {
+        groupHeight += group.clientHeight;
+      });
+      while (height > groupHeight || Math.abs(height - groupHeight) <= 20) {
+        const group = actionGroups.shift();
+        group.style.display = "none";
```
driskull, line 361: question: why are groups being hidden?

packages/calcite-components/src/components/action-bar/action-bar.tsx
```diff
@@ -342,8 +342,32 @@ export class ActionBar
...
+        actionGroups.forEach((group) => {
+          groupHeight += group.clientHeight;
+        });
+      }
+    } else {
+      actionGroups.forEach((a) => (a.style.display = "inline-block"));
```
driskull, line 367: question: the default display for action group is flex. Why is it being set to inline block here?

driskull (changes requested):
Initial review

alisonailea (changes requested):
Make sure any token's you're adding align with the naming schema https://github.com/Esri/calcite-design-system/wiki/tokens-naming-schema

packages/calcite-components/src/components/shell-panel/shell-panel.scss
```diff
@@ -48,21 +48,23 @@
   z-index: var(--calcite-shell-panel-z-index, calc(theme("zIndex.header") + 1));
 }
 
+:host([layout="vertical"][display-mode="float"]) .content,
+:host([layout="vertical"][display-mode="content-detached"]) .content {
+  flex: 2;
+}
+
 :host([layout="vertical"][width-scale="s"]) .content {
   --calcite-shell-panel-width-internal: var(--calcite-shell-panel-width, 12vw);
```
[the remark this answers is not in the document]
josercarcamo: I didn't add these tokens. They're pre-existing.

packages/calcite-components/src/components/shell-panel/shell-panel.scss
josercarcamo: To make the content completely span the entire width of the shell-panel.

packages/calcite-components/src/components/action-bar/action-bar.tsx
josercarcamo: I thought that's what we agreed when we all met to discuss this issue. We agreed to remove action groups when the vertical action bar was too long.

packages/calcite-components/src/components/action-bar/action-bar.tsx
driskull: @josercarcamo I thought it would just be up to the author to remove the actions that won't fit into their content. I wasn't aware we were going to hide actions or groups. Maybe we can discuss this one again?

packages/calcite-components/src/components/shell-panel/shell-panel.tsx
josercarcamo: I thought it was the position of the action bar and I changed it because of the last part in https://github.com/Esri/calcite-design-system/issues/7513#issuecomment-1856941472. But if it's not, I'll revert it.

packages/calcite-components/src/components/shell-panel/shell-panel.scss
driskull: This would remove the `max-width` of 300px for a shell-panel in cases where the shell panel is vertical. I think this would remove that existing default max-width of 300px.

If we do remove these, there is still a reference in the CSS to `--calcite-shell-panel-max-width-internal` that would need to be removed.

packages/calcite-components/src/components/shell-panel/shell-panel.tsx
driskull: But it was deleted from the doc so that would not need to happen here.

packages/calcite-components/src/components/shell-panel/shell-panel.tsx
driskull: No, its for the position of the shell panel.

The action-bar is placed on a side depending on the shell-panel position. If we need to decouple that then we would need to create a new property for it.

We can't change the existing property because it would be a breaking change and it has other functionality.

I'm not convinced the action-bar needs its own property for positioning unless there's a valid use case to have a shell-panel on the left with an action-bar on the right?

packages/calcite-components/src/components/shell-panel/shell-panel.scss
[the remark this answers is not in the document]
alisonailea: can you update them while you're there? They're already internal so there won't be any breaking changes to worry about.

josercarcamo:
@driskull @alisonailea do you want to take another look?

alisonailea (approved):
🎉

driskull (changes requested):
Looking better @josercarcamo 🎉 

Some more changes:

- The breaking changes format still needs to be followed: https://esri.github.io/calcite-design-system/?path=/docs/overview-contributing--docs#breaking-changes
  - The PR title should be updated to be `feat` instead of `fix` and should also have the exclamation point and breaking change notes in the body.
  - The PR title should reflect the feature and breaking change.
  - This PR and issue should be labelled as a breaking change and not merged until before a Major version release (3x). So the milestone needs to be moved out of July.

driskull (changes requested):
@josercarcamo we also need some screenshot test(s) for the new `float-detach` value of position. This can be added to the `shell.stories.ts`

packages/calcite-components/src/components/shell/shell.stories.ts
```diff
@@ -360,7 +360,7 @@ darkModeRTL_TestOnly.parameters = { themes: modesDarkDefault };
 
 export const closedPanels = (): string =>
   html(`<calcite-shell content-behind>
-  <calcite-shell-panel slot="panel-start" display-mode="float">
+  <calcite-shell-panel slot="panel-start" display-mode="float-detach">
```
driskull, line 363: @josercarcamo I think we should still keep `float` stories since that value isn't being removed. If we just change this then we lose coverage of its functionality. Can you add new screenshot tests to cover float-detach?

driskull (changes requested):
Almost there!

Can you add `displayMode` to the title and breaking change description? `float` and `float-detach` are values of `displayMode`.

Maybe 

> feat(shell-panel)!: add a `displayMode` of `float-detach` and change the behavior of the `float`

packages/calcite-components/src/components/shell/shell.stories.ts
josercarcamo: I don't think we're on the same page: "float-detach" is taking the place of the old "float", and "float" is the new fuctionality.

driskull: Ok, but in the story updates its only replacing `float` which mean `float` would no longer have any test coverage. Shouldn't it?

josercarcamo:
@driskull one more review please.

driskull (approved):
Looks good! Can you just add `displayMode` to the PR title somehow?

And this shouldn't be merged in until Nov release. 👍

josercarcamo:
@driskull one last review please.

packages/calcite-components/src/components/shell/shell.stories.ts
```diff
@@ -360,7 +360,7 @@ darkModeRTL_TestOnly.parameters = { themes: modesDarkDefault };
 
 export const closedPanels = (): string =>
   html(`<calcite-shell content-behind>
-  <calcite-shell-panel slot="panel-start" display-mode="float">
+  <calcite-shell-panel slot="panel-start" display-mode="float-content">
```
driskull, line 363: question: should these values remain `float` to keep test coverage for the deprecated `float` value? Applies to other changes below that are being changed from float to float-content.

packages/calcite-components/src/components/shell-panel/shell-panel.tsx
```diff
@@ -134,7 +138,7 @@ export class ShellPanel implements ConditionalSlotComponent, LocalizedComponent,
   @Prop({ reflect: true }) position: Extract<"start" | "end", Position> = "start";
 
   /**
-   * When `true` and `displayMode` is not `float`, the component's content area is resizable.
+   * When `true` and `displayMode` is not `float-content`, the component's content area is resizable.
```
driskull, line 141: question: should this doc still include `float` as well as `float-content`?

packages/calcite-components/src/components/shell-panel/shell-panel.tsx
```diff
@@ -81,17 +81,21 @@ export class ShellPanel implements ConditionalSlotComponent, LocalizedComponent,
...
 
   @Watch("displayMode")
   handleDisplayMode(value: DisplayMode): void {
-    this.detached = value === "float";
+    this.detached = value === "float-content" || value === "float";
   }
 
   /**
-   * When `displayMode` is `float`, specifies the maximum height of the component.
+   * When `displayMode` is `float-content`, specifies the maximum height of the component.
```
driskull, line 98: question: should this doc still include `float` as well as `float-content`?

packages/calcite-components/src/components/shell-panel/shell-panel.scss
```diff
@@ -6,10 +6,10 @@
  * @prop --calcite-shell-panel-width: Specifies the width of the component.
  * @prop --calcite-shell-panel-max-width: Specifies the maximum width of the component.
  * @prop --calcite-shell-panel-min-width: Specifies the minimum width of the component.
- * @prop --calcite-shell-panel-height: When `layout` is `horizontal`, or `layout` is `vertical` and `displayMode` is `float`, specifies the height of the component.
- * @prop --calcite-shell-panel-max-height: When `layout` is `horizontal`, or `layout` is `vertical` and `displayMode` is `float`, specifies the maximum height of the component.
- * @prop --calcite-shell-panel-min-height: When `layout` is `horizontal`, or `layout` is `vertical` and `displayMode` is `float`, specifies the minimum height of the component.
- * @prop --calcite-shell-panel-detached-max-height: [Deprecated] Use the `heightScale` property instead. When `displayMode` is `float`, specifies the maximum height of the component.
+ * @prop --calcite-shell-panel-height: When `layout` is `horizontal`, or `layout` is `vertical` and `displayMode` is `float-content`, specifies the height of the component.
```
driskull, line 9: question: should this be when `displayMode` is float-content or float?

driskull (approved):
LGTM! just had some questions

PR #9804, opened 2024-07-17
Dris0000/dialog resizable

driskull:
**Related Issue:** #


driskull:
> 1 - the Heading text when no description is populated is not spaced correctly

Panel issue needed. cc @geospatialem @DitwanP 

> 2 - there is no padding in the content area within the padding - I thought we were adding a default value here that matched header / footing padding and was overridable to 0?

Panel issue needed. cc @geospatialem @DitwanP 

> 3 - The drag is very slow, I'd maybe expect no transition / animation delay and have it be immediate

Fixed

> 4 - The "zones" for dragging vs. resizing when used together is tough to discern - not sure a good solution off top of head

I'm not sure what we could do here either.

> 5 - I can still "drag" after clicking and holding on the Close button - the pointer doesn't change, so I wouldn't expect this click and hold to allow dragging.

Not sure we can do anything here. The close button is on top of the draggable element so it takes precedence.

> 6. Right now you can "shrink" the Dialog so much that it cuts off header or footer, or disappears completely. Should we only "resize" the content area - and ensure the min-height of the Dialog is  footer + header height? (this is vertically - do we want to have a min-width to prevent this width-wise?)

Fixed: added min height/width. Should we make these sizes depend on the view size or fixed pixels?

> 7. Keyboard use - nice!!!! The focus outline is obscured behind the header - maybe we can ensure that is visible.

Need panel issue for focus outline. cc @geospatialem @DitwanP 

> 8. Using shift to resize makes sense, let's check with Kitty to confirm. I might expect some extra keyboard niceties - when resizing - home / end + trigger key could resize to max / min extent.

cc @geospatialem

> 9. For placement keyboard navigation - similarly, I think home / end could maybe be used to offer a quick way to move to the extents of the viewport.

I'm not sure home/end make sense for positioning.

> 10. You can only "resize" when there isn't content populated in that area (in the demo - only the "empty space on the right / above text" will show the cursor. Maybe this is fine but could be confusing and if the user slots full height / width content - maybe there wouldn't be anywhere to click and hold to resize. Maybe some buffer around the exterior with a safe area? Wouldn't work on "cover" I suppose...

I'd have to see a good use case here. `interact.js` only changes the cursor when on the edges of the resizable element. If something in a higher stacking order also adjusts the cursor that could potentially be a UX problem but I'm not sure that would happen.

> 11. The keyboard functionalities work when the close button is focused. Confirm / disconfirm if this is intentional or if this should only be available when the user is focused on the container

As intended. The keyboard event listeners are on the dialog container so will work when any element is focused within as key events will bubble.

DitwanP:
Issues for items 1, 2, and 7 respectively:
https://github.com/Esri/calcite-design-system/issues/9862
https://github.com/Esri/calcite-design-system/issues/10037
https://github.com/Esri/calcite-design-system/issues/10022

geospatialem:
> > 8. Using shift to resize makes sense, let's check with Kitty to confirm. I might expect some extra keyboard niceties - when resizing - home / end + trigger key could resize to max / min extent.
> 
> cc @geospatialem thoughts? The problem with home/end is what does it control? width or height? Would like to follow any aria role guidelines here if possible.

Yes, all the key combinations in place are sound to me, where resizing uses the arrow keys and `shift` and dragging uses the arrow keys. 💪🏻 

It does seem like there's a conflict with both JAWS and NVDA and the key combinations in [testing](https://codepen.io/geospatialem/pen/BagmoWa) 🧪 , so we may need to make a few small tweaks, or strategize a timeline for AT support for resizing and draggability in the future.

As far as the `Home` and `End` keys - I don't think we should explore the `Home` and `End` keys for either functionality. I'm not finding any similar functionality in the wild 🐯, and in some cases users depend on the `Home` and `End` keys to navigate to the top and bottom of a page - think we should opt for that native functionality instead.

SkyeSeitz:
> Fixed: added min height/width. Should we make these sizes depend on the view size or fixed pixels? Maybe design can chime in on an ideal min-width and min-height? @SkyeSeitz @ashetland

I agree — only the default content area should shrink/grow in height while the footer, action-bar & header would only shrink/grow in width. What about content-top/bottom would they potentially have a scroll bar like the default slot or could we avoid that?

https://github.com/user-attachments/assets/c6780280-e129-4c04-8469-1699f5a08c1f

Perhaps the values for Shell Panel's min-widths could be a helpful gauge here if we go the fixed pixel route? 
image

driskull:
> I agree — only the default content area should shrink/grow in height while the footer, action-bar & header would only shrink/grow in width. What about content-top/bottom would they potentially have a scroll bar like the default slot or could we avoid that?

The problem is that we're not resizing the panel (or part of the panel), we're resizing the dialog. A user can replace that whole panel with their own content.

packages/calcite-components/src/components/dialog/dialog.scss
```diff
@@ -120,13 +120,12 @@ calcite-panel {
...
 .content {
-  padding: var(--calcite-dialog-content-space, var(--calcite-internal-dialog-content-padding));
+  padding: var(--calcite-internal-dialog-content-padding);
 }
 
 .dialog {
   @apply pointer-events-none
-    float-none
-    m-6
+    m-0
```
driskull, line 128: removed margin here so a user can resize the dialog to all available space.

packages/calcite-components/src/components/dialog/dialog.scss
```diff
@@ -139,31 +138,26 @@ calcite-panel {
...
 
-  --calcite-internal-dialog-hidden-position: translate3d(0, 20px, 0);
-  --calcite-internal-dialog-shown-position: translate3d(0, 0, 0);
+  min-inline-size: var(--calcite-dialog-min-size-x, var(--calcite-internal-dialog-min-size-x));
+  min-block-size: var(--calcite-dialog-min-size-y, var(--calcite-internal-dialog-min-size-y));
 
-  inset-inline-start: var(--calcite-dialog-offset-x);
-  inset-block-start: var(--calcite-dialog-offset-y);
+  --calcite-internal-dialog-hidden-position: calc(var(--calcite-dialog-offset-y, 0px) + 20px);
+  --calcite-internal-dialog-shown-position: var(--calcite-dialog-offset-y, 0);
```
driskull, line 148: these internal vars allow handling the animation when an offset is being used.

packages/calcite-components/src/components/dialog/dialog.scss
```diff
@@ -139,31 +138,26 @@ calcite-panel {
...
     }
-  }
-
-  &--closing {
-    &-idle {
-      transform: var(--calcite-internal-dialog-shown-position);
-    }
-    &-active {
-      transform: var(--calcite-internal-dialog-hidden-position);
+    &-done {
```
driskull, line 159: added a done class to handle animation when an offset is set.

driskull:
@macandcheese @jcfranco @SkyeSeitz @ashetland @geospatialem  this is ready for review again.

packages/calcite-components/src/components/dialog/dialog.e2e.ts
```diff
@@ -802,4 +845,187 @@ describe("calcite-dialog", () => {
...
+          >Hello world!</calcite-dialog
+        >`,
+      });
+      await skipAnimations(page);
+      await page.setViewport({ width: 1200, height: 1200 });
+      await page.waitForChanges();
+      const container = await page.find(`calcite-dialog >>> .${CSS.dialog}`);
+      expect((await container.getComputedStyle()).transform).toBe("matrix(1, 0, 0, 1, 0, 0)");
+
+      await dispatchDialogKeydown({ page, key: "ArrowDown", shiftKey: false });
```
jcfranco, line 862: Could you use `E2EElement.press` instead of this custom function? I think the main difference would be that you'd need to press shift on the page keyboard before calling `press` on the element (similar to [this](https://github.com/Esri/calcite-design-system/blob/281a86954e266a554dfbbf96c0a641820f11dc56/packages/calcite-components/src/components/color-picker/color-picker.e2e.ts#L2348-L2349)).

packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -298,12 +334,28 @@ export class Dialog
...
 
+  @State() assistiveText = "";
+
+  @Watch("messages")
+  @Watch("dragEnabled")
+  @Watch("resizable")
+  updateAssistiveText(): void {
+    const { messages } = this;
+    this.assistiveText = messages
+      ? `${this.dragEnabled ? messages.dragEnabled : ""} ${this.resizable ? messages.resizeEnabled : ""}`.trim()
```
jcfranco, line 345: Is `trim` necessary? I believe the text will be read correctly regardless. cc @geospatialem

jcfranco, line 345: Message concatenation should be avoided. Typically, we'd use a localized template ([example](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/components/input-time-zone/assets/input-time-zone/t9n/messages.json#L7)), but the conditional nature of this does make it tricky. @annierm18 Do you have any guidance here?

macandcheese (approved):
NICE!

For design to consider @ashetland @SkyeSeitz :
- I'd suggest we should add the `max-height / max-width` css properties at this time, and perhaps even set a max-height/width by default based on scale. 

- The margin was removed - which is fine as in most cases the Modal will become "cover" - I wonder if we should have a way to have an edge-offset - we have something somewhat similar in Tooltip / Popover for distance from reference element. I could see designers not wanting the dialog to be able to abut the edge of the screen. This might be solvable with css property?

- I find the "resize on all sizes" to be a bit unexpected - both the fact that it can be resized from all sides, and that the scale effect occurs across the entire component, not just the direction the user is "dragging from".

  - It is more problematic on the "close button side" - as there is a tiny, couple px hitbox that provides resize. 

  - I wonder if we should consider just bottom-right like this option (they also have draggable just at the top): [from SAP](https://sap.github.io/ui5-webcomponents/play/#eyJpbmRleC5odG1sIjp7Im5hbWUiOiJpbmRleC5odG1sIiwiY29udGVudCI6IjwhLS0gcGxheWdyb3VuZC1mb2xkIC0tPlxuPCFET0NUWVBFIGh0bWw-XG48aHRtbCBsYW5nPVwiZW5cIj5cblxuXG5cblxuPGhlYWQ-XG4gICAgXG4gICAgPHN0eWxlPlxuICAgICAgKjpub3QoOmRlZmluZWQpIHtcbiAgICAgICAgZGlzcGxheTogbm9uZTtcbiAgICAgIH1cbiAgICA8L3N0eWxlPlxuXG4gICAgXG4gICAgPHN0eWxlPlxuICAgICAgKjpub3QoOmRlZmluZWQpIHtcbiAgICAgICAgZGlzcGxheTogbm9uZTtcbiAgICAgIH1cbiAgICA8L3N0eWxlPlxuXG4gICAgXG4gICAgPHN0eWxlPlxuICAgICAgKjpub3QoOmRlZmluZWQpIHtcbiAgICAgICAgZGlzcGxheTogbm9uZTtcbiAgICAgIH1cbiAgICA8L3N0eWxlPlxuXG4gICAgPG1ldGEgY2hhcnNldD1cIlVURi04XCI-XG4gICAgPG1ldGEgbmFtZT1cInZpZXdwb3J0XCIgY29udGVudD1cIndpZHRoPWRldmljZS13aWR0aCwgaW5pdGlhbC1zY2FsZT0xLjBcIj5cbiAgICA8dGl0bGU-U2FtcGxlPC90aXRsZT5cbiAgICA8bGluayByZWw9XCJzdHlsZXNoZWV0XCIgaHJlZj1cIi4vbWFpbi5jc3NcIj5cbjwvaGVhZD5cblxuPGJvZHkgc3R5bGU9XCJiYWNrZ3JvdW5kLWNvbG9yOiB2YXIoLS1zYXBCYWNrZ3JvdW5kQ29sb3IpOyBoZWlnaHQ6IDQ1MHB4O1wiPlxuICAgIDwhLS0gcGxheWdyb3VuZC1mb2xkLWVuZCAtLT5cblxuICAgIDx1aTUtYnV0dG9uIGlkPVwiZGlhbG9nT3BlbmVyXCI-T3BlbiBEaWFsb2c8L3VpNS1idXR0b24-XG5cbiAgICA8dWk1LWRpYWxvZyBpZD1cImRpYWxvZ1wiIGhlYWRlci10ZXh0PVwiUmVnaXN0ZXIgRm9ybVwiIHJlc2l6YWJsZSBkcmFnZ2FibGU-XG4gICAgICAgIDxzZWN0aW9uIGNsYXNzPVwibG9naW4tZm9ybVwiPlxuICAgICAgICAgICAgPGRpdj5cbiAgICAgICAgICAgICAgICA8dWk1LWxhYmVsIGZvcj1cInVzZXJuYW1lXCIgcmVxdWlyZWQgc2hvdy1jb2xvbj5Vc2VybmFtZTwvdWk1LWxhYmVsPlxuICAgICAgICAgICAgICAgIDx1aTUtaW5wdXQgaWQ9XCJ1c2VybmFtZVwiPjwvdWk1LWlucHV0PlxuICAgICAgICAgICAgPC9kaXY-XG4gICAgICAgICAgICA8ZGl2PlxuICAgICAgICAgICAgICAgIDx1aTUtbGFiZWwgZm9yPVwicGFzc3dvcmRcIiByZXF1aXJlZCBzaG93LWNvbG9uPlBhc3N3b3JkPC91aTUtbGFiZWw-XG4gICAgICAgICAgICAgICAgPHVpNS1pbnB1dCBpZD1cInBhc3N3b3JkXCIgdHlwZT1cIlBhc3N3b3JkXCIgdmFsdWUtc3RhdGU9XCJOZWdhdGl2ZVwiPjwvdWk1LWlucHV0PlxuICAgICAgICAgICAgPC9kaXY-XG4gICAgICAgICAgICA8ZGl2PlxuICAgICAgICAgICAgICAgIDx1aTUtbGFiZWwgZm9yPVwiZW1haWxcIiB0eXBlPVwiRW1haWxcIiByZXF1aXJlZCBzaG93LWNvbG9uPkVtYWlsPC91aTUtbGFiZWw-XG4gICAgICAgICAgICAgICAgPHVpNS1pbnB1dCBpZD1cImVtYWlsXCI-PC91aTUtaW5wdXQ-XG4gICAgICAgICAgICA8L2Rpdj5cbiAgICAgICAgICAgIDxkaXY-XG4gICAgICAgICAgICAgICAgPHVpNS1sYWJlbCBmb3I9XCJhZGRyZXNzXCIgc2hvdy1jb2xvbj5BZGRyZXNzPC91aTUtbGFiZWw-XG4gICAgICAgICAgICAgICAgPHVpNS1pbnB1dCBpZD1cImFkZHJlc3NcIj48L3VpNS1pbnB1dD5cbiAgICAgICAgICAgIDwvZGl2PlxuICAgICAgICA8L3NlY3Rpb24-XG4gICAgICAgIDxkaXYgc2xvdD1cImZvb3RlclwiIHN0eWxlPVwiZGlzcGxheTogZmxleDsganVzdGlmeS1jb250ZW50OiBmbGV4LWVuZDsgd2lkdGg6IDEwMCU7IGFsaWduLWl0ZW1zOiBjZW50ZXJcIj5cbiAgICAgICAgICAgIDxkaXYgc3R5bGU9XCJmbGV4OiAxO1wiPjwvZGl2PlxuICAgICAgICAgICAgPHVpNS1idXR0b24gY2xhc3M9XCJkaWFsb2dDbG9zZXJcIiBkZXNpZ249XCJFbXBoYXNpemVkXCI-UmVnaXN0ZXI8L3VpNS1idXR0b24-XG4gICAgICAgIDwvZGl2PlxuICAgIDwvdWk1LWRpYWxvZz5cbiAgICA8IS0tIHBsYXlncm91bmQtZm9sZCAtLT5cbiAgICA8c2NyaXB0IHR5cGU9XCJtb2R1bGVcIiBzcmM9XCJtYWluLmpzXCI-PC9zY3JpcHQ-XG48L2JvZHk-XG5cbjwvaHRtbD5cbjwhLS0gcGxheWdyb3VuZC1mb2xkLWVuZCAtLT5cbiJ9LCJtYWluLmpzIjp7Im5hbWUiOiJtYWluLmpzIiwiY29udGVudCI6Ii8qIHBsYXlncm91bmQtaGlkZSAqL1xuaW1wb3J0IFwiLi9wbGF5Z3JvdW5kLXN1cHBvcnQuanNcIjtcbi8qIHBsYXlncm91bmQtaGlkZS1lbmQgKi9cbmltcG9ydCBcIkB1aTUvd2ViY29tcG9uZW50cy9kaXN0L0RpYWxvZy5qc1wiO1xuaW1wb3J0IFwiQHVpNS93ZWJjb21wb25lbnRzL2Rpc3QvQnV0dG9uLmpzXCI7XG5pbXBvcnQgXCJAdWk1L3dlYmNvbXBvbmVudHMvZGlzdC9JbnB1dC5qc1wiO1xuaW1wb3J0IFwiQHVpNS93ZWJjb21wb25lbnRzL2Rpc3QvTGluay5qc1wiO1xuaW1wb3J0IFwiQHVpNS93ZWJjb21wb25lbnRzL2Rpc3QvTGFiZWwuanNcIjtcblxudmFyIGRpYWxvZ09wZW5lciA9IGRvY3VtZW50LmdldEVsZW1lbnRCeUlkKFwiZGlhbG9nT3BlbmVyXCIpO1xudmFyIGRpYWxvZyA9IGRvY3VtZW50LmdldEVsZW1lbnRCeUlkKFwiZGlhbG9nXCIpO1xudmFyIGRpYWxvZ0Nsb3NlcnMgPSBbLi4uZGlhbG9nLnF1ZXJ5U2VsZWN0b3JBbGwoXCIuZGlhbG9nQ2xvc2VyXCIpXTtcblxuZGlhbG9nT3BlbmVyLmFjY2Vzc2liaWxpdHlBdHRyaWJ1dGVzID0ge1xuICAgIGhhc1BvcHVwOiBcImRpYWxvZ1wiLFxuICAgIGNvbnRyb2xzOiBkaWFsb2cuaWQsXG59O1xuZGlhbG9nT3BlbmVyLmFkZEV2ZW50TGlzdGVuZXIoXCJjbGlja1wiLCAoKSA9PiB7XG4gICAgZGlhbG9nLm9wZW4gPSB0cnVlO1xufSk7XG5kaWFsb2dDbG9zZXJzLmZvckVhY2goYnRuID0-IHtcbiAgICBidG4uYWRkRXZlbnRMaXN0ZW5lcihcImNsaWNrXCIsICgpID0-IHtcbiAgICAgICAgZGlhbG9nLm9wZW4gPSBmYWxzZTtcbiAgICB9KTtcbn0pIn0sIm1haW4uY3NzIjp7Im5hbWUiOiJtYWluLmNzcyIsImNvbnRlbnQiOiIubG9naW4tZm9ybSB7XG4gICAgZGlzcGxheTogZmxleDtcbiAgICBmbGV4LWRpcmVjdGlvbjogY29sdW1uO1xuICAgIGp1c3RpZnktY29udGVudDogc3BhY2UtZXZlbmx5O1xuICAgIGFsaWduLWl0ZW1zOiBmbGV4LXN0YXJ0O1xuICAgIG1hcmdpbjogM3JlbSA2cmVtO1xufVxuXG4ubG9naW4tZm9ybSA-IGRpdiB7XG4gICAgZGlzcGxheTogZ3JpZDtcbiAgICB3aWR0aDogMTVyZW07XG4gICAgbWFyZ2luLWJvdHRvbTogLjVyZW07XG59XG5cbkBtZWRpYShtYXgtd2lkdGg6IDYwMHB4KSB7XG4gICAgLmxvZ2luLWZvcm0ge1xuICAgICAgICBtYXJnaW46IDNyZW0gMXJlbTtcbiAgICB9XG59In19)

packages/calcite-components/src/components/dialog/dialog.scss
jcfranco: Maybe it's the naming that's throwing me off, but what's the difference between `idle` and `done`? For context, the `idle` and `active` classes match a pattern commonly used [in](https://maquettejs.org/docs/animations.html#css-transitions) [transition](https://sijakret.github.io/lit-transition/css-transitions) [libs](https://reactcommunity.org/react-transition-group/css-transition) where the `idle` class is meant to set the initial transition state and the `active` class defines the end state. I'll create a wiki page on this since I don't think this is documented anywhere. We can also revisit naming if it helps.

jcfranco (commented):
Looking gewd! 😎

ashetland:
> I find the "resize on all sizes" to be a bit unexpected - both the fact that it can be resized from all sides, and that the scale effect occurs across the entire component, not just the direction the user is "dragging from".
> 
> It is more problematic on the "close button side" - as there is a tiny, couple px hitbox that provides resize.
> 
> I wonder if we should consider just bottom-right like this option (they also have draggable just at the top): [from SAP](https://sap.github.io/ui5-webcomponents/play/#eyJpbmRleC5odG1sIjp7Im5hbWUiOiJpbmRleC5odG1sIiwiY29udGVudCI6IjwhLS0gcGxheWdyb3VuZC1mb2xkIC0tPlxuPCFET0NUWVBFIGh0bWw-XG48aHRtbCBsYW5nPVwiZW5cIj5cblxuXG5cblxuPGhlYWQ-XG4gICAgXG4gICAgPHN0eWxlPlxuICAgICAgKjpub3QoOmRlZmluZWQpIHtcbiAgICAgICAgZGlzcGxheTogbm9uZTtcbiAgICAgIH1cbiAgICA8L3N0eWxlPlxuXG4gICAgXG4gICAgPHN0eWxlPlxuICAgICAgKjpub3QoOmRlZmluZWQpIHtcbiAgICAgICAgZGlzcGxheTogbm9uZTtcbiAgICAgIH1cbiAgICA8L3N0eWxlPlxuXG4gICAgXG4gICAgPHN0eWxlPlxuICAgICAgKjpub3QoOmRlZmluZWQpIHtcbiAgICAgICAgZGlzcGxheTogbm9uZTtcbiAgICAgIH1cbiAgICA8L3N0eWxlPlxuXG4gICAgPG1ldGEgY2hhcnNldD1cIlVURi04XCI-XG4gICAgPG1ldGEgbmFtZT1cInZpZXdwb3J0XCIgY29udGVudD1cIndpZHRoPWRldmljZS13aWR0aCwgaW5pdGlhbC1zY2FsZT0xLjBcIj5cbiAgICA8dGl0bGU-U2FtcGxlPC90aXRsZT5cbiAgICA8bGluayByZWw9XCJzdHlsZXNoZWV0XCIgaHJlZj1cIi4vbWFpbi5jc3NcIj5cbjwvaGVhZD5cblxuPGJvZHkgc3R5bGU9XCJiYWNrZ3JvdW5kLWNvbG9yOiB2YXIoLS1zYXBCYWNrZ3JvdW5kQ29sb3IpOyBoZWlnaHQ6IDQ1MHB4O1wiPlxuICAgIDwhLS0gcGxheWdyb3VuZC1mb2xkLWVuZCAtLT5cblxuICAgIDx1aTUtYnV0dG9uIGlkPVwiZGlhbG9nT3BlbmVyXCI-T3BlbiBEaWFsb2c8L3VpNS1idXR0b24-XG5cbiAgICA8dWk1LWRpYWxvZyBpZD1cImRpYWxvZ1wiIGhlYWRlci10ZXh0PVwiUmVnaXN0ZXIgRm9ybVwiIHJlc2l6YWJsZSBkcmFnZ2FibGU-XG4gICAgICAgIDxzZWN0aW9uIGNsYXNzPVwibG9naW4tZm9ybVwiPlxuICAgICAgICAgICAgPGRpdj5cbiAgICAgICAgICAgICAgICA8dWk1LWxhYmVsIGZvcj1cInVzZXJuYW1lXCIgcmVxdWlyZWQgc2hvdy1jb2xvbj5Vc2VybmFtZTwvdWk1LWxhYmVsPlxuICAgICAgICAgICAgICAgIDx1aTUtaW5wdXQgaWQ9XCJ1c2VybmFtZVwiPjwvdWk1LWlucHV0PlxuICAgICAgICAgICAgPC9kaXY-XG4gICAgICAgICAgICA8ZGl2PlxuICAgICAgICAgICAgICAgIDx1aTUtbGFiZWwgZm9yPVwicGFzc3dvcmRcIiByZXF1aXJlZCBzaG93LWNvbG9uPlBhc3N3b3JkPC91aTUtbGFiZWw-XG4gICAgICAgICAgICAgICAgPHVpNS1pbnB1dCBpZD1cInBhc3N3b3JkXCIgdHlwZT1cIlBhc3N3b3JkXCIgdmFsdWUtc3RhdGU9XCJOZWdhdGl2ZVwiPjwvdWk1LWlucHV0PlxuICAgICAgICAgICAgPC9kaXY-XG4gICAgICAgICAgICA8ZGl2PlxuICAgICAgICAgICAgICAgIDx1aTUtbGFiZWwgZm9yPVwiZW1haWxcIiB0eXBlPVwiRW1haWxcIiByZXF1aXJlZCBzaG93LWNvbG9uPkVtYWlsPC91aTUtbGFiZWw-XG4gICAgICAgICAgICAgICAgPHVpNS1pbnB1dCBpZD1cImVtYWlsXCI-PC91aTUtaW5wdXQ-XG4gICAgICAgICAgICA8L2Rpdj5cbiAgICAgICAgICAgIDxkaXY-XG4gICAgICAgICAgICAgICAgPHVpNS1sYWJlbCBmb3I9XCJhZGRyZXNzXCIgc2hvdy1jb2xvbj5BZGRyZXNzPC91aTUtbGFiZWw-XG4gICAgICAgICAgICAgICAgPHVpNS1pbnB1dCBpZD1cImFkZHJlc3NcIj48L3VpNS1pbnB1dD5cbiAgICAgICAgICAgIDwvZGl2PlxuICAgICAgICA8L3NlY3Rpb24-XG4gICAgICAgIDxkaXYgc2xvdD1cImZvb3RlclwiIHN0eWxlPVwiZGlzcGxheTogZmxleDsganVzdGlmeS1jb250ZW50OiBmbGV4LWVuZDsgd2lkdGg6IDEwMCU7IGFsaWduLWl0ZW1zOiBjZW50ZXJcIj5cbiAgICAgICAgICAgIDxkaXYgc3R5bGU9XCJmbGV4OiAxO1wiPjwvZGl2PlxuICAgICAgICAgICAgPHVpNS1idXR0b24gY2xhc3M9XCJkaWFsb2dDbG9zZXJcIiBkZXNpZ249XCJFbXBoYXNpemVkXCI-UmVnaXN0ZXI8L3VpNS1idXR0b24-XG4gICAgICAgIDwvZGl2PlxuICAgIDwvdWk1LWRpYWxvZz5cbiAgICA8IS0tIHBsYXlncm91bmQtZm9sZCAtLT5cbiAgICA8c2NyaXB0IHR5cGU9XCJtb2R1bGVcIiBzcmM9XCJtYWluLmpzXCI-PC9zY3JpcHQ-XG48L2JvZHk-XG5cbjwvaHRtbD5cbjwhLS0gcGxheWdyb3VuZC1mb2xkLWVuZCAtLT5cbiJ9LCJtYWluLmpzIjp7Im5hbWUiOiJtYWluLmpzIiwiY29udGVudCI6Ii8qIHBsYXlncm91bmQtaGlkZSAqL1xuaW1wb3J0IFwiLi9wbGF5Z3JvdW5kLXN1cHBvcnQuanNcIjtcbi8qIHBsYXlncm91bmQtaGlkZS1lbmQgKi9cbmltcG9ydCBcIkB1aTUvd2ViY29tcG9uZW50cy9kaXN0L0RpYWxvZy5qc1wiO1xuaW1wb3J0IFwiQHVpNS93ZWJjb21wb25lbnRzL2Rpc3QvQnV0dG9uLmpzXCI7XG5pbXBvcnQgXCJAdWk1L3dlYmNvbXBvbmVudHMvZGlzdC9JbnB1dC5qc1wiO1xuaW1wb3J0IFwiQHVpNS93ZWJjb21wb25lbnRzL2Rpc3QvTGluay5qc1wiO1xuaW1wb3J0IFwiQHVpNS93ZWJjb21wb25lbnRzL2Rpc3QvTGFiZWwuanNcIjtcblxudmFyIGRpYWxvZ09wZW5lciA9IGRvY3VtZW50LmdldEVsZW1lbnRCeUlkKFwiZGlhbG9nT3BlbmVyXCIpO1xudmFyIGRpYWxvZyA9IGRvY3VtZW50LmdldEVsZW1lbnRCeUlkKFwiZGlhbG9nXCIpO1xudmFyIGRpYWxvZ0Nsb3NlcnMgPSBbLi4uZGlhbG9nLnF1ZXJ5U2VsZWN0b3JBbGwoXCIuZGlhbG9nQ2xvc2VyXCIpXTtcblxuZGlhbG9nT3BlbmVyLmFjY2Vzc2liaWxpdHlBdHRyaWJ1dGVzID0ge1xuICAgIGhhc1BvcHVwOiBcImRpYWxvZ1wiLFxuICAgIGNvbnRyb2xzOiBkaWFsb2cuaWQsXG59O1xuZGlhbG9nT3BlbmVyLmFkZEV2ZW50TGlzdGVuZXIoXCJjbGlja1wiLCAoKSA9PiB7XG4gICAgZGlhbG9nLm9wZW4gPSB0cnVlO1xufSk7XG5kaWFsb2dDbG9zZXJzLmZvckVhY2goYnRuID0-IHtcbiAgICBidG4uYWRkRXZlbnRMaXN0ZW5lcihcImNsaWNrXCIsICgpID0-IHtcbiAgICAgICAgZGlhbG9nLm9wZW4gPSBmYWxzZTtcbiAgICB9KTtcbn0pIn0sIm1haW4uY3NzIjp7Im5hbWUiOiJtYWluLmNzcyIsImNvbnRlbnQiOiIubG9naW4tZm9ybSB7XG4gICAgZGlzcGxheTogZmxleDtcbiAgICBmbGV4LWRpcmVjdGlvbjogY29sdW1uO1xuICAgIGp1c3RpZnktY29udGVudDogc3BhY2UtZXZlbmx5O1xuICAgIGFsaWduLWl0ZW1zOiBmbGV4LXN0YXJ0O1xuICAgIG1hcmdpbjogM3JlbSA2cmVtO1xufVxuXG4ubG9naW4tZm9ybSA-IGRpdiB7XG4gICAgZGlzcGxheTogZ3JpZDtcbiAgICB3aWR0aDogMTVyZW07XG4gICAgbWFyZ2luLWJvdHRvbTogLjVyZW07XG59XG5cbkBtZWRpYShtYXgtd2lkdGg6IDYwMHB4KSB7XG4gICAgLmxvZ2luLWZvcm0ge1xuICAgICAgICBtYXJnaW46IDNyZW0gMXJlbTtcbiAgICB9XG59In19)

I agree the resizing behavior feels unexpected. If I can resize form any side, I'd expect only that side to adjust, like a Finder window. If there's room for an affordance in the bottom right corner I think the SAP example works well. 

Also agree that it feels weird to be able to drag from just about anywhere. Limiting it to the heading makes sense to me.

cc @SkyeSeitz @macandcheese

packages/calcite-components/src/components/dialog/dialog.tsx
driskull: I can remove it. It shouldn't be necessary.

SkyeSeitz:
> I'd suggest we should add the max-height / max-width css properties at this time, and perhaps even set a max-height/width by default based on scale.

Could the viewport be the max by default?

> I find the "resize on all sizes" to be a bit unexpected - both the fact that it can be resized from all sides, and that the scale effect occurs across the entire component, not just the direction the user is "dragging from".

I would also expect resizing to occur from the direction the user is dragging rather than in all directions. I do however think it is ideal to be able to resize from any edge. 

> Also agree that it feels weird to be able to drag from just about anywhere. Limiting it to the header makes sense to me.

+1 for dragging limited to the header.

driskull:
> I agree the resizing behavior feels unexpected. If I can resize form any side, I'd expect only that side to adjust, like a Finder window. If there's room for an affordance in the bottom right corner I think the SAP example works well.

This is because our dialog is centered and stays centered when resizing. It actually makes sense to do it this way, especially when a dialog is positioned top-end or bottom-end. IMO this isn't unexpected or unusual. I prefer the resizing on any side compared to a single corner because of the positioning logic we have.

> Also agree that it feels weird to be able to drag from just about anywhere. Limiting it to the header makes sense to me.

The user can slot anything inside of the dialog which means there may not always be a header. Also, the header is something that's part of panel so I don't think I can target it specifically without digging into the shadow DOM of panel.

Is this feedback something we need to meet again to discuss and work through? What parts of these are show stoppers?

ashetland:
Is there anything we can do about the flicker cursor change? Feels broken.

https://github.com/user-attachments/assets/d8e7e433-e03e-4de0-8de8-7c081b22972a

SkyeSeitz:
> Even in the center position, this feels like something I'd achieve with a modifier key.

Agreed. I would expect the placement prop on draggable/resizable Dialogs to be a starting point, but after dragging/resizing, the user controls placement.

ashetland:
> > Also agree that it feels weird to be able to drag from just about anywhere. Limiting it to the header makes sense to me.
> 
> The user can slot anything inside of the dialog which means there may not always be a header. Also, the header is something that's part of panel so I don't think I can target it specifically without digging into the shadow DOM of panel.

I don't think this is a show stopper for me.

driskull:
> Agreed. I would expect the placement prop on draggable/resizable Dialogs to be a starting point, but after dragging/resizing, the user controls placement.

To do this, we would likely need another `placement` property like `custom` where we don't set any flex justify or align. We could set this property once a user starts dragging. I would need to test this out as it may suddenly shift the whole dialog.

> Is there anything we can do about this flicker cursor change? Feels broken.

I don't think so. The close action is higher z-index than the dialog so its pointer cursor takes proirity.

It would be nice to know what feedback items are required fixes vs suggestions. Like defined here: https://github.com/Esri/calcite-design-system/wiki/PR-Review-Guidelines#comments

ashetland:
Sorting out the resizing behavior is a required fix for me. That's only one.

packages/calcite-components/src/components/dialog/dialog.tsx
annierm18: If these are completely separate sentences, then this change should be fine. We have some guidance on string concatenation on our website here: https://esriis.sharepoint.com/sites/i18n_t9n/SitePages/String-Composition.aspx in the "Concatenation" section. I checked with Jonathan Turpin from the Translation team on this and he also said it should be fine. The major concatenation issue happens when you use concatenation within the same sentence.

packages/calcite-components/src/components/dialog/dialog.e2e.ts
driskull: @jcfranco this isn't working. I think it may be because the function that gets called needs to get boundingClientRect from the element and I'm not sure its getting that when not happening in browser context.

What I tried:

Line 958.
```ts
//await dispatchDialogKeydown({ page, key: "ArrowDown", shiftKey: true });
  await page.keyboard.press("Shift");
  await container.press("ArrowDown");
  await page.waitForChanges();
```

packages/calcite-components/src/components/dialog/dialog.tsx
driskull: Makes sense. 👍

driskull:
> This "scaling" effect is what feels unexpected. Here I have positioned the Dialog at the bottom right and when I resize from the top left corner, I'd expect only that corner to move and not expand in all directions and "off the edge". Even in the center position, this feels like something I'd achieve with a modifier key.

I'm not sure what we can do about this. Its because of the `placement` option. By default, `placement` is `center` which means that resizing the dialog will resize both sides to keep it centered. If you try different `placement` options, it behaves differently.

I think we would need to find some workaround here.

driskull:
I tried a few things without any success.

I think we can consider this one blocked. cc @geospatialem @DitwanP

geospatialem:
> I tried a few things without any success.
> 
> I think we can consider this one blocked. cc @geospatialem @DitwanP

@driskull Do we need a new issue, and as far as the unsuccessful part, is this referring to [Aaron's comment](https://github.com/Esri/calcite-design-system/pull/9804#issuecomment-2289839117)? Or do we need to block this PR/effort for further investigation?

driskull:
> Do we need a new issue, and as far as the unsuccessful part, is this referring to https://github.com/Esri/calcite-design-system/pull/9804#issuecomment-2289839117? Or do we need to block this PR/effort for further investigation?

Yes, its referring to resizing and placement.

We would need to block the issue to figure out a way to work out the resizing and placement issues.

packages/calcite-components/src/components/dialog/dialog.e2e.ts
jcfranco: Just realized there's an error in the example I linked (will update the test). Could you try this?

```ts
  await page.keyboard.down("Shift"); // down vs press
  await container.press("ArrowDown");
  await page.waitForChanges();
```

driskull: @jcfranco that still isn't working. Any idea why it wouldn't resize using that method?

```
  ● calcite-dialog › keyboard resize › should resize properly via shift and arrow keys

    expect(received).toBe(expected) // Object.is equality

    Expected: "211px"
    Received: "236px"
```

image

jcfranco: Not sure. I'll have to explore. Can you create a refactor issue for me? Should be a quick one. 🤞

driskull: Yeah Ill create one if this PR lands.

driskull:
@SkyeSeitz @ashetland @macandcheese @jcfranco  could you review this again?

packages/calcite-components/src/components/dialog/dialog.scss
```diff
@@ -5,7 +5,9 @@
  *
  * @prop --calcite-dialog-scrim-background-color: Specifies the background color of the component's scrim.
  * @prop --calcite-dialog-size-x: Specifies the width of the component, using `px`, `em`, `rem`, `vw`, or `%`. Does not exceed the viewport's width - applies when `placement="cover"` is set.
+ * @prop --calcite-dialog-min-size-x: Specifies the minimum width of the component, using `px`, `em`, `rem`, `vw`, or `%`.
  * @prop --calcite-dialog-size-y: Specifies the height of the component, using `px`, `em`, `rem`, `vh`, or `%`. Does not exceed the viewport's height - applies when `placement="cover"` is set.
```
macandcheese, line 9: I think we'll want corresponding `-max-size-x` props, currently you can resize to perhaps undesired heights.

packages/calcite-components/src/components/dialog/dialog.scss
```diff
@@ -78,27 +85,21 @@
...
 
 :host([scale="m"]) {
-  --calcite-internal-dialog-content-padding: var(--calcite-spacing-md);
-  --calcite-internal-dialog-padding: theme("spacing.4");
-  --calcite-internal-dialog-padding-large: theme("spacing.5");
-  --calcite-internal-dialog-title-text: theme("fontSize.2h");
-  --calcite-internal-dialog-content-text: theme("fontSize.0");
+  --calcite-internal-dialog-content-padding: var(--calcite-dialog-content-space, var(--calcite-spacing-md));
+  --calcite-internal-dialog-min-size-x: 288px;
+  --calcite-internal-dialog-min-size-y: 140px;
```
macandcheese, line 96: cc @SkyeSeitz  @ashetland I think we may need to make these a bit taller - by default it allows shrinking too much with a common header / footer set up: 
Screenshot 2024-08-26 at 4 30 04 PM

macandcheese (approved):
Looks great! Couple of small things:

- I think we'd want to add max width / height css props at this time alongside "min" - just like with Shell Panel's resizable functionality - they may not need a default value (cc @ashetland  @SkyeSeitz  ?) but they should be able to be set.

- There's a small one px wiggle as you adjust - maybe something to note and improve.

- I still find it strange from a UX perspective that you can "click and press" on the Close Button, and that enables dragging. Maybe we can improve this in the future but it strikes me as unexpected.

- Similarly, there's a 1/2px buffer "around" the close button where you can click to drag and resize, but it's not practical to use because of the small hit area and may lead to user frustration.

None of the above are blockers IMO, now that the resize scaling has been fixed, this is looking really good!

SkyeSeitz:
> I think we'd want to add max width / height css props at this time alongside "min" - just like with Shell Panel's resizable functionality - they may not need a default value (cc @ashetland @SkyeSeitz ?) but they should be able to be set.

Agreed this may be something users need to set

> I still find it strange from a UX perspective that you can "click and press" on the Close Button, and that enables dragging. Maybe we can improve this in the future but it strikes me as unexpected.

Though I have oddly seen this possible in other systems, I do land on the side of restricting this interaction if possible. 

Can this focus ring be placed above (visible around entire Dialog)?
image


Yay dragging & resizing looking so great!

SkyeSeitz (approved):

packages/calcite-components/src/components/dialog/dialog.scss
SkyeSeitz: Updated the [Figma specs ](https://www.figma.com/design/wVqftT1aZDjwYVkDZ2V5fC/Dialog?node-id=1423-25133&t=jeOtWMpcAdYzjKRY-11)with adjusted min-heights:
sm **140px**
md **180px**
lg **220px**
image

packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -442,13 +493,242 @@ export class Dialog
...
+          dragPosition.x += dialogStep;
+          this.setTransform();
+          this.reflowInteraction();
+          event.preventDefault();
+        }
+        break;
+    }
+  };
+
+  private setTransform(): void {
```
jcfranco, line 590: Would `updateTransform` be more appropriate? A set function without arguments seems unusual to me.

packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -309,12 +335,41 @@ export class Dialog
 
   @State() defaultMessages: DialogMessages;
 
+  @State() assistiveText: string | null = "";
```
jcfranco, line 338: Should this be initialized to `null`? `updateAssistiveText` uses `null` when there's no text to display.

jcfranco (approved):
Aside from a few comments, this...

👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍
👍😎👍👍👍😎😎😎😎👍😎😎😎😎😎👍😎👍👍👍😎👍😎👍
👍😎👍👍👍😎👍👍👍👍👍👍😎👍👍👍😎😎👍😎😎👍😎👍
👍😎👍👍👍😎👍😎😎👍👍👍😎👍👍👍😎👍😎👍😎👍😎👍
👍😎👍👍👍😎👍👍😎👍👍👍😎👍👍👍😎👍👍👍😎👍👍👍
👍😎😎😎👍😎😎😎😎👍👍👍😎👍👍👍😎👍👍👍😎👍😎👍
👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍

packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -442,13 +493,242 @@ export class Dialog
...
+      name: "drag",
+    });
+
+    await interaction.reflow({
+      name: "resize",
+    });
+  }
+
+  private getDOMRect(): DOMRect {
+    return this.transitionEl?.getBoundingClientRect() ?? new DOMRect();
```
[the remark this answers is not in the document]
driskull: will refactor.

packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -442,13 +493,242 @@ export class Dialog
 
   @Watch("opened")
   handleOpenedChange(value: boolean): void {
-    const idleClass = value ? CSS.openingIdle : CSS.closingIdle;
-    this.transitionEl.classList.add(idleClass);
+    this.transitionEl.classList.toggle(CSS.openingActive, value);
     onToggleOpenCloseComponent(this);
   }
 
+  private async reflowInteraction(): Promise<void> {
```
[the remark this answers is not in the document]
driskull: will rename :)

packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -442,13 +493,242 @@ export class Dialog
...
+    const { transitionEl } = this;
+
+    if (!transitionEl) {
+      return;
+    }
+
+    transitionEl.style[type] = size !== null ? `${Math.round(size)}px` : null;
+  }
+
+  private unsetInteraction = (): void => {
```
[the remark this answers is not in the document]
driskull: will fix!

driskull:
> I think we'll want corresponding -max-size-x props, currently you can resize to perhaps undesired heights.

Added these. @macandcheese @SkyeSeitz @alisonailea

driskull:
> I think we'd want to add max width / height css props at this time alongside "min" - just like with Shell Panel's resizable functionality - they may not need a default value (cc @ashetland @SkyeSeitz ?) but they should be able to be set.

We may want to document that when resizing is enabled, these values should be pixel values.

> There's a small one px wiggle as you adjust - maybe something to note and improve.

I couldn't get this one fixed. Its because the `placement` option is justifying and aligning the dialog. When resized, the dialog wobbles a bit because of this. If you place the dialog to a corner, you won't see it resizing some sides.

driskull:
Screenshots look good. Only changes were related to the min-height.

PR #9816, opened 2024-07-19
feat: add component tokens

alisonailea:
**Related Issue:** #7180

## Summary

Component tokens define design pattern decisions at the component level. They should reference/fallback to global tokens.

--calcite-[component]-[group]-[property]-[state]

State can be "hover", "pressed", "selected" or nothing.

- A state that is shared by :hover and :focus should use `-hover`
- A state shared by :focus and :active should use `-pressed`
- A state which includes :active and/or custom states like [selected] and [checked] should use `-pressed`
- An edge case design pattern may require a third variable state for components with unique styling when in a custom attribute state like [selected] and [checked] which is different than `-pressed`. This should use `-selected`.
       Example
	      --calcite-button-background-color
	      --calcite-button-background-color-hover
	      --calcite-button-background-color-pressed

Due to the CSS cascade and how CSS variables can penetrate the shadowDOM it is not recommended you create tokens for sub-components unless the wrapping component is overriding the component's pattern. I.E overriding a token value. When a pattern is shared between related components a single token may be used. 

```mermaid
flowchart TD
    A1{<a target="_blank" href="https://github.com/Esri/calcite-design-system/blob/6ccb12a17ec371ba13fe144f8d37fdfb607dfce2/packages/calcite-components/stencil.config.ts#L17">Are components associated?</a>}
    A2{Can it be used independently?}
    B1{Look at the component in Figma.<br />Does it share a visual pattern<br />with associated components?}
    C1{It can share tokens}
    C2{It should have unique tokens}

    A1 --> |Yes| A2
    A1 -->|No| C2
    A2 -->|No| C1
    A2 -->|Yes| B1
    B1 -->|Yes| C1
    B1 -->|No| C2
```

Example:
--calcite-combobox-background-color is used in both Combobox & ComboboxItem. This is documented in the JSDocs at the top of each component's SCSS file.

This is also the case for elements in the component which share a pattern. In most use-cases icons in the components should reference the -text-color token. In rare cases when Icons should not follow the text color pattern, check if it can use the "-indicator-color" pattern. If not, add an "-icon-" element to the naming schema.

#### 1. Make sure component aligns with /dev 

some components in epic/7180-component-tokens are very out of date with /dev. This makes sure the most recent JSX, resources and utils are being use.

#### 2. review component for "tokenizable" styles
  background-color
  border-color
  color (text-color)
  box-shadow(shadow)
  border-radius(corner-radius)

#### 3. review Figma design for patterns
    
What components are related to this component?
Are there shared interaction patterns?
are the use cases covered by "-hover" and "-pressed"?
Is there an "indicator" pattern?
    This is usually border-input or transparent by default and -brand, -brand-press when :active or :hovered.

#### 4. update component SCSS to use CSS variable tokens

[property]: var([component-token], var([global-token/fallback]));

Sometimes it's useful to utilized `-internal-` domain tokens. These are not considered official Calcite Design Tokens and are for developer convenience. If a pattern is observed in `-internal-` tokens. Bring it up to the team for consideration to be move to a Calcite Design Token.

##### E2E

```.ts
/*:Starting template: */
  describe("theme", () => {
    describe("default", () => {
      themed("calcite-component", {});
    });
});
```
		
- Use `themed` from common tests
- If a token does not apply in the default context, add a new describe
- use `html` formatter for non-default component tests

##### Storybook

```ts
export const theming_TestOnly = (): string => html`
  <style>
    .container {
      /* --calcite-tokens-to-test */
    }
  </style>
  <div class="container">
    ${/* Component(s) HTML */}
  </div>
`;
```

###### Demo Page

```html
<demo-theme tokens="--calcite-tokens, --calcite-tokens-as-list"
  ><!-- component HTML --></demo-theme>
```

## Components Done except for demo-html and storybook

- [x] Accordion
- [x] Accordion Item
- [x] Action
- [x] Action Bar
- [x] Action Group
- [x] Action Menu
- [x] Action Pad
- [x] alert
- [x] avatar
- [x] block
- [x] block section
- [x] button
- [x] card
- [x] card group
- [x] carousel
- [x] carousel item
- [x] checkbox
- [x] chip
- [x] chip group
- [x] carousel
- [x] carousel item
- [x] checkbox
- [x] chip
- [x] chip group
- [x] color picker
- [x] color picker hex input
- [x] color picker swatch
- [x] combobox
- [x] combobox item
- [x] combobox item group


packages/calcite-components/src/utils/variableValue.ts
```diff
@@ -0,0 +1,42 @@
...
+    "text-color$": "rgb(239,118,39)",
+    "border-color$": "rgb(156, 89, 209)",
+    "background-color$": "rgb(44, 44, 44)",
+    color$: "rgb(0, 191, 255)",
+    highlighted$: "rgb(255, 105, 180)",
+    selected$: "rgb(255, 255, 255)",
+    shadow$:
+      "rgb(255, 255, 255) 0px 0px 0px 4px, rgb(255, 105, 180) 0px 0px 0px 5px inset, rgb(0, 191, 255) 0px 0px 0px 9px",
+    "z-index$": "42",
+    "(size|space)$": "42px",
```
driskull, line 21: @alisonailea @jcfranco I noticed this one isn't working for a variable with `space-x` or `space-y`. Can we get it fixed?

driskull: We'll also need to add `offset`, `offset-x`, `offset-y`

alisonailea: feel free to add test examples where you need them too

jcfranco:
Closing as we are tackling component tokens in smaller chunks. Thanks, everyone! ✨💪✨

PR #9818, opened 2024-07-19
fix(combobox-item): tweak center content font-weight and spacing

jcfranco:
**Related Issue:** #3695 

## Summary

This fixes spacing and font-weight to follow the updated spec.

### Notes

* updates previous internal CSS prop to follow conventions (will follow suit for others in a separate PR)
* ensures selected font weight is only applied to the title/heading
* fixes trailing indent padding
* prevents wrapping of `short-heading`

### DIFF
--- a/packages/calcite-components/src/components/combobox-item/combobox-item.scss
+++ b/packages/calcite-components/src/components/combobox-item/combobox-item.scss
@@ -12,7 +12,7 @@
 .scale--m {
   @apply text-n1h;
   --calcite-combobox-item-spacing-unit-l: theme("spacing.3");
-  --calcite-combobox-item-spacing-unit-s: theme("spacing.2");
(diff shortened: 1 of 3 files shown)


driskull (approved):
Looks good. The maxItems test might need some adjustments though.

jcfranco:
For future reference, we'll need do the same as [`e0446b4` (#9818)](https://github.com/Esri/calcite-design-system/pull/9818/commits/e0446b4d38911671d19986c113a4a523633b350e#diff-9bc2dae4daf67ad95f668837f3083752245fbc22e0a58b1e5a043830e05af02cR1069) for scroller calculations. The height of items might be fractional, so we should use `getBoundingClientRect`, as [`offsetHeight` uses integers](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetHeight).

Issue #9295, opened 2024-05-09
[Tooling] Communicate component deprecations and upcoming removal via console


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Communicate component deprecations and future major version targeted release for developers to start transition to other  supported components.

### Acceptance Criteria

If a component is deprecated, we communicate to developers:
- [ ] its deprecation and 
- [ ] when the anticipated removal will take place (e.g., `3.0`, `4.0`, etc.)

### Which Component

Will currently apply to:
- **Pick List**: Pick List, Pick List Group, Pick List Item in `3.0`
- **Value List**: Value List, Value List Item in `3.0`
- **Tip**: Tip, Tip Group, Tip Manager in `4.0`
- **Tile Select**: Tile Select, Tile Select Group in `4.0`

### Priority impact

p1 - need for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

geospatialem:
Awesome, this going to be a great addition! @jcfranco It looks like there is one deprecation missing for the `value-list-item` in this [codepen](https://codepen.io/geospatialem/pen/wvLzoKp). Relaying this back over to ya to add in. 🕵🏻

DitwanP:
🍡 Looks great! Verified locally on `dev`
![image](https://github.com/user-attachments/assets/5ffab4af-0776-4e3a-ab97-6ac04e84666a)

PR #9821, opened 2024-07-20
chore: add runtime component deprecation warnings

jcfranco:
**Related Issue:** #9295

## Summary

This adds deprecation warning messages to components marked for deprecation at v3 and v4.

### DIFF
--- a/packages/calcite-components/src/components/pick-list-group/pick-list-group.tsx
+++ b/packages/calcite-components/src/components/pick-list-group/pick-list-group.tsx
@@ -6,7 +6,14 @@ import {
 } from "../../utils/conditionalSlot";
 import { getSlotted } from "../../utils/dom";
 import { constrainHeadingLevel, Heading, HeadingLevel } from "../functional/Heading";
+import { logger } from "../../utils/logger";
 import { CSS, SLOTS } from "./resources";
+
+logger.deprecated("component", {
+  name: "calcite-pick-list-item-group",
+  removalVersion: 3,
+  suggested: "calcite-list-item-group",
+});
 
 /**
  * @deprecated Use the `calcite-list` component instead.
--- a/packages/calcite-components/src/utils/logger.ts
+++ b/packages/calcite-components/src/utils/logger.ts
@@ -1,0 +1,84 @@
+import { logLevel } from "./config";
+
+export type LogLevel = "debug" | "info" | "warn" | "error" | "trace";
+
+type Message = string;
+type MajorVersion = number;
+
+type DeprecatedContext = "component" | "property" | "method" | "event" | "slot";
+type DeprecatedParams = {
+  name: string;
+  suggested?: string | string[];
+  component?: string;
+  removalVersion: MajorVersion | "future";
+};
+
+/**
+ * Exported for testing purposes only
+ */
+export const loggedDeprecations = new Set<string>();
+
+const logLevels = {
+  debug: 1,
+  info: 2,
+  warn: 4,
+  error: 8,
+} as const;
+
+function willLog(level: LogLevel): boolean {
+  return logLevels[level] >= logLevels[logLevel];
+}
+
+function forwardToConsole(level: LogLevel, ...data: any[]): void {
+  if (!willLog(level)) {
+    return;
+  }
+
+  const badgeTemplate = "%ccalcite";
+  const badgeStyle = "background: #007AC2; color: #fff; border-radius: 4px;padding: 2px 4px;";
+
+  console[level].call(this, badgeTemplate, badgeStyle, ...data);
+}
+
+let listFormatter: Intl.ListFormat;
+
+export const logger = {
+  debug: (message: Message) => forwardToConsole("debug", message),
+  info: (message: Message) => forwardToConsole("info", message),
+  warn: (message: Message) => forwardToConsole("warn", message),
+  error: (message: Message) => forwardToConsole("error", message),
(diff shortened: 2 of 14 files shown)


packages/calcite-components/src/components/pick-list-group/pick-list-group.tsx
```diff
@@ -6,8 +6,15 @@ import {
 } from "../../utils/conditionalSlot";
 import { getSlotted } from "../../utils/dom";
 import { constrainHeadingLevel, Heading, HeadingLevel } from "../functional/Heading";
+import { logger } from "../../utils/logger";
 import { CSS, SLOTS } from "./resources";
 
+logger.deprecated("component", {
```
driskull, line 12: this is super cool!

packages/calcite-components/src/utils/logger.ts
```diff
@@ -0,0 +1,84 @@
...
+  return logLevels[level] >= logLevels[logLevel];
+}
+
+function forwardToConsole(level: LogLevel, ...data: any[]): void {
+  if (!willLog(level)) {
+    return;
+  }
+
+  const badgeTemplate = "%ccalcite";
+  const badgeStyle = "background: #007AC2; color: #fff; border-radius: 4px;padding: 2px 4px;";
```
driskull, line 38: Should design approve the styling :)

Wondering if we could use CSS vars here instead of styles. I guess it doesn't really matter since its just a log badge but still.

packages/calcite-components/src/utils/logger.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  loggedDeprecations.add(key);
+
+  const multiSuggestions = Array.isArray(suggested);
+
+  if (multiSuggestions && !listFormatter) {
+    listFormatter = new Intl.ListFormat("en", { style: "long", type: "disjunction" });
+  }
+
+  const message = `[${name}] ${context} is deprecated and will be removed in ${removalVersion === "future" ? `a future version` : `v${removalVersion}`}.${suggested ? ` Use ${multiSuggestions ? listFormatter.format(suggested.map((suggestion) => `"${suggestion}"`)) : `"${suggested}"`} instead.` : ""}`;
```
driskull, line 81: We don't need to translate this right?

driskull:
1. I do like the badge
2. I don't think we need the calcite prefix as it is redundant with the badge.
3. I think its probably fine in the config for now. What do you think?

driskull (approved):
This will be awesome! 💯

packages/calcite-components/src/utils/logger.ts
jcfranco: I did test this, but no dice. Custom CSS props don't appear to be available for console styling. 😞

packages/calcite-components/src/utils/logger.ts
jcfranco: Correct. Dev messaging is not translated.

DitwanP:
This is great! 💪 👏

I'm with Matt on points 1, 2, and 3 from his first comment above.

geospatialem (commented):
**_Awesome!!_** 💪🏻 

As far as the version, should we also mention a timeframe for folks to plan, or keep it vague with the version? Something like:

> `calcite` [tip-group] component is deprecated and will be removed in v4 (2025 Q4). Use "carousel" or "carousel item" instead.

> 1. Badge 👍👎? 

Badge it up! 🦡📛 
 
> 2. Assuming 👍 on the badge, should we keep calcite in component names used in messaging? The badge already provides context

For consistency with our documentation, its okay to omit since its mentioned already in the badge. If we remove the badge, we should leave it in though. 

The only hesitation would be if omitting it creates confusion for developers to use the recommended component(s).

It looks like the messaging also adds `"` to the recommended components, should the symbol be around each of the recommendations, like:

> Use "carousel" or "carousel-item" instead.

> 3. Do we want to keep logLevel in the config or should we do something similar to [Maps SDK's config](https://developers.arcgis.com/javascript/latest/api-reference/esri-config.html#log) for future log-related features? I don't see a use case for interceptors at the moment.

Flexible on whichever approach is best as long as it is clear to developers working with both Maps SDK and Calcite.

benelan (approved):
I like this idea as well! One suggestion that hasn't been mentioned, maybe add a "none" or "disabled" option to `logLevels`? That way teams can completely turn it off in prod and set it to something verbose in dev.

jcfranco:
> As far as the version, should we also mention a timeframe for folks to plan, or keep it vague with the version? Something like:

I like the idea of adding the timeframe, but this would be tricky to maintain and enforce via TypeScript. 

> The only hesitation would be if omitting it creates confusion for developers to use the recommended component(s).

I hear ya. We should be good on this front since we do have the badge giving context and often omit `calcite` in the doc when referring to components. This also makes the message less noisy.

> It looks like the messaging also adds " to the recommended components, should the symbol be around each of the recommendations, like:

My bad. 😅 I used an outdated screenshot. Each component should be quoted in the latest version (covered by a test too).

jcfranco:
Thanks for the feedback, y'all! 🚀 Made the following changes:

* dropped namespace from component messaging to simplify (this is enforced by TS)
* added an `off` level to mute all logs

eriklharper:
+1 for badge, +1 for keeping `calcite` in the name of the component message.  I think that's clearer for folks outside the core team since that's how they interact with the components.  Keeping log level in the config sounds like a good first step, we can always add more to that down the road.

Issue #9722, opened 2024-07-01
Add automation for moving related issues to verification when main issue is closed out


DitwanP:
### Priority impact

p - low

### Summary

In the last few weeks something that has continued to pop up is some issues could be resolved through an issue/PR currently in progress, but they need to still be executed prior to validation. Currently the PE's are assigning those issues to themselves and periodically need to remember to check if the main issue has been closed out and then proceed to verify if it resolves the related one.

### Desired Outcome

Hopefully an automation to notify us when the original issue is closed out and moves the related one to a label that shows it's ready to test.

geospatialem:
@DitwanP Updating triage labels! ⭐🚀⭐

DitwanP:
🍭 Verified by testing with another issue that was about to be closed, see above bot comment for proof.

Also...
<img width="277" alt="Screenshot 2024-05-21 at 11 06 30 AM" src="https://github.com/user-attachments/assets/adfdd7e5-63bf-48f0-9b4a-da5f68aca396">

Issue #5678, opened 2022-11-01
Slider handle is out of alignment in the zero position


eriklharper:
### Actual Behavior

Either of the slider's handles becomes misaligned in the `0` position on the slider axis when used in `label-handles` mode.  Both the handle in single mode and both handles in range mode float above the slider bar at the `0` position, which obscures the `0` label text.

![image](https://user-images.githubusercontent.com/821864/199359434-506db6c1-5bea-49ac-9463-5c13c9b83e8d.png)

### Expected Behavior

The slider handles should always be centered along the slider axis.

### Reproduction Sample

https://codepen.io/eriklharper/pen/WNyrPVZ?editors=100

### Reproduction Steps

1. Render a `calcite-slider` with `label-handles` turned on.
2. Observe that the handle in position `0` floats above the axis and obscures the label.

### Reproduction Version

v1.0.0-beta.95

### Esri team

Calcite (dev)

jcfranco:
This is fixed in the latest release (via https://github.com/Esri/calcite-components/issues/5382). Closing!

eriklharper:
> This is fixed in the latest release (via #5382). Closing!

![](https://media.giphy.com/media/PhKhSXofSAm3e/giphy.gif)

PR #1234, opened 2020-11-06
docs(changelog): Clean up changelog

macandcheese:
## Summary
Cleans up stray content from the commit message body that didn't pertain to the breaking change.

PR #321, opened 2020-02-05
build(deps): add dependency updating script

jcfranco:
Adds NPM script to make dependency updating easier.

Uses https://www.npmjs.com/package/updtr to update all dependencies (runs tests on each one and only updates the passing ones) and commits on success.

PR #9845, opened 2024-07-24
ci(actions): action to add comment when blocking issue is closed

DitwanP:
**Related Issue:** #9722 

## Summary
Adds new action to create a comment on blocked issues when the blocking issue has been closed out. 

With these changes we would manually add blocked issues to the body of the blocking issue in the format:
 `Blocked issues: #1234, #5678...`
 
When closed the action would get those issues and add a comment saying the blocking issue has been closed as well removing blocked label (maybe).


.github/scripts/addUnblockedComment.js
```diff
@@ -0,0 +1,56 @@
+// When a blocking issue is closed:
+// 1. Leaves a comment on all the issues listed as blocked in body,
+
+module.exports = async ({ github, context }) => {
```
benelan, line 4:
```suggestion
/** @param {import('github-script').AsyncFunctionArguments} AsyncFunctionArguments */
module.exports = async ({ github, context }) => {
  const { repo, owner } = context.repo;

  const payload = /** @type {import('@octokit/webhooks-types').IssuesEvent} */ (context.payload);
module.exports = async ({ github, context }) => {
```

Can you please add types via JSDOC comments? See the other scripts in that directory for full examples

.github/scripts/addUnblockedComment.js
```diff
@@ -0,0 +1,56 @@
...
+  if (!issueBody) {
+    console.log("No issue body was found");
+    return;
+  }
+
+  // Get the list of issues blocked by this issue if there are any
+  const blockedIssuesLine = issueBody.match(blockedIssuesRegex);
+
+  // Add a "@" character to notify the user
+  const verifiers = ISSUE_VERIFIERS.split(",").map((v) => " @" + v.trim());
```
benelan, line 18:
```suggestion
  const verifiers = ISSUE_VERIFIERS.split(",").map((v) => `@${v.trim()}`);
```
Nitpick: can you use a string literal here instead of concatenating

.github/scripts/addUnblockedComment.js
```diff
@@ -0,0 +1,56 @@
...
+
+  // If "Blocked issues" line is matched in the body then create a comment on each issue listed
+  if (blockedIssuesLine) {
+    const issueNumbers = blockedIssuesLine[0]
+      .slice(16)
+      .split(", ")
+      .map((number) => number.slice(1));
```
benelan, line 25: The regex isn't strict enough for this match parsing. For example, the regex would match this string, which would break the slice logic:

> Blocked issues:           #123,#321

.github/scripts/addUnblockedComment.js
```diff
@@ -0,0 +1,56 @@
...
+    for (const issueNumber of issueNumbers) {
+      const issueProps = {
+        issue_number: issueNumber,
+        owner: context.repo.owner,
+        repo: context.repo.repo,
+      };
+
+      await github.rest.issues.createComment({
+        ...issueProps,
+        body: `Issue #${context.issue.number} has been closed, this issue can now move forward. cc ${verifiers}`,
```
benelan, line 36: Would it be possible for an issue to have multiple blockers? If so, this message might give the issue author false hope if other blockers are still open. Maybe we could rephrase it to something like:

> The blocking issue #${context.issue.number} has been closed. This issue is ready for re-evaluation, cc ${verifiers}.

.github/scripts/addUnblockedComment.js
```diff
@@ -0,0 +1,56 @@
...
+        if (error.status === 404) {
+          console.log(`The label "blocked" does not exist on issue #${issueNumber}.`);
+        } else {
+          throw error;
+        }
+      }
+    }
+  } else {
+    console.log(`No blocked issues listed in the body of issue #${context.payload.issue.number}.`);
+    return;
```
benelan, line 54: Nitpick: this `return` is unnecessary since it's at the end of the script

benelan (changes requested):
I added a couple comments but it's almost there, nice work Ditwan!

.github/scripts/addUnblockedComment.js
DitwanP: Agreed!

.github/scripts/addUnblockedComment.js
DitwanP: Nice! tweaked a bit to `/(?!Blocked issues:\s)(#\d+)/gi` so it doesn't include the commas at the end of the numbers ["#123","#321","#11"]

benelan: Nice catch, it was late and they must have blended in with the array commas lol

.github/scripts/addUnblockedComment.js
```diff
@@ -0,0 +1,56 @@
...
+
+  // If "Blocked issues" line is matched in the body then create a comment on each issue listed
+  if (blockedIssues) {
+    const issueNumbers = blockedIssues.map((number) => number.slice(1));
+
+    for (const issueNumber of issueNumbers) {
+      const issueProps = {
+        repo,
+        owner,
+        issue_number: issueNumber,
```
benelan, line 32: I'm seeing type errors for the `createComment` and `removeLabel` function calls. Are you not seeing this on your end?

> tsserver: Argument of type '{ body: string; repo: string; owner: string; issue_number: string; }' is not assignable to parameter of type 'RequestParameters & { owner: string; repo: string; issue_number: number; body: string; }'.
  Type '{ body: string; repo: string; owner: string; issue_number: string; }' is not assignable to type '{ owner: string; repo: string; issue_number: number; body: string; }'.
    Types of property 'issue_number' are incompatible.
      Type 'string' is not assignable to type 'number'. [2345]
 
 I wonder if I need to fix something with the types setup. Regardless, `issueNumber` needs to be a `number`, not a `string`.

```suggestion
        issue_number: Number(issueNumber),
```

benelan (approved):
Adding types uncovered one more problem, but after that's fixed it LGTM!

Issue #9807, opened 2024-07-17
[Shell Panel] Unexpected behavior when resizing via click


macandcheese:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Placeholder from UC:

### Expected Behavior

Placeholder from UC

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/QWXyvqw?editors=1010

### Reproduction Steps
- View codepen 
- Use mouse to click resizable affordance
- Notice the jump in height
- Focus the resizable affordance
- Notice the issue does not occur

### Reproduction Version

Latest

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

macandcheese:
While the original Codepen with this issue now works on the latest - other similar reproduction case does not: https://codepen.io/mac_and_cheese/pen/JjQLYrW?editors=1000

Same issue as originally - clicking on the resize affordance adjusts size of adjacent content, but does not occur when using keyboard focus / arrows to adjust.

Re-opening cc @geospatialem @DitwanP

driskull:
@macandcheese is it working on the next version?

https://codepen.io/driskull/pen/ExBEVdW?editors=1000

macandcheese:
Yep it does - the original issue was resolved in 2.11.1, not the latest example.

driskull:
I guess https://github.com/Esri/calcite-design-system/pull/9912 fixed it.

PR #9846, opened 2024-07-24
fix(shell): fix resizing a slotted shell-panel when clicking to resize

driskull:
**Related Issue:** #9807

## Summary

- fixes styles when resizing
  - makes non interactive container not suddenly jump full height/width.
- cleanup


macandcheese (approved):
🏅 Thanks for the quick fix!

Issue #9471, opened 2024-05-30
[Carousel] Carousel Items shouldn't animate when using `Home` / `End` and destination Carousel Item is currently active


macandcheese:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When the Carousel component is focused, and keyboard is used to navigate with `Home` and `End` - the component should not animate if the currently active Carousel Item is first, or last, and matches the destination.

The behavior should match the behavior of clicking on a currently active Carousel Item's representative "dot", or using the `Space` or `Enter` key on a currently active Carousel Item's representative "dot".

### Expected Behavior

If the currently active Carousel Item is first, and a user navigates with `Home`, or the currently active Carousel item is last, and a user navigates with `End`, the animation should not play. 

https://github.com/Esri/calcite-design-system/assets/4733155/0f0ba6fb-abf1-4841-9e82-50c3cfc59849

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/pomRYPd?editors=100

### Reproduction Steps

1. Open Codepen
2. Focus the Carousel component, and use `Home` / `End` while the first / last Carousel Item are already active.
3. Notice the animation occurs.
4. Try to "re-activate" an active Carousel Item by clicking the dot, or using `Space` / `Enter` on the active dot, and notice the animation (correctly) does not occur.

### Reproduction Version

2.9.0

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

DitwanP:
🍡 Verified on `2.11.0-next.23`

PR #9848, opened 2024-07-24
fix(carousel): Prevent duplicate animation when navigating via keyboard

macandcheese:
**Related Issue:** #9471

## Summary
When using `home` and `end` while focused on the Container, prevent "re-navigating" to the currently active Carousel Item and showing a duplicate animation. This matches the behavior when a user selects an individual Carousel Item "dot", both with mouse or via keyboard.

### DIFF
--- a/packages/calcite-components/src/components/carousel/carousel.tsx
+++ b/packages/calcite-components/src/components/carousel/carousel.tsx
@@ -494,6 +494,8 @@ export class Carousel
       return;
     }
 
+    const lastItem = this.items.length - 1;
+
     switch (event.key) {
       case " ":
       case "Enter":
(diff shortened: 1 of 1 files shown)


packages/calcite-components/src/components/carousel/carousel.tsx
```diff
@@ -512,13 +514,19 @@ export class Carousel
         break;
       case "Home":
         event.preventDefault();
+        if (this.selectedIndex === 0) {
```
driskull, line 517: should this move before event.preventDefault so event isn't prevented if nothing happens?

packages/calcite-components/src/components/carousel/carousel.tsx
```diff
@@ -512,13 +514,19 @@ export class Carousel
...
+          return;
+        }
         this.direction = "backward";
         this.setSelectedItem(0, true);
         break;
       case "End":
         event.preventDefault();
+        if (this.selectedIndex === lastItem) {
```
driskull, line 525: should this move before event.preventDefault so event isn't prevented if nothing happens?

driskull (commented):
comment!

packages/calcite-components/src/components/carousel/carousel.tsx
macandcheese: I think I'd still expect it to be prevented. Without it, the Home / End keys could navigate the Carousel with one keystroke and then unexpectedly scroll the page with the next.

driskull: Yep. makes sense.

driskull (approved):
👍

Issue #5615, opened 2022-10-25
Tree selection-type "ancestors" - Provide a valid label for form fields - (2036570417) 


dqateam:
### Module: 
 38 Tree and Tree Item 

### Violation: 
 Provide a valid label for form fields where the selection type is set to `"ancestors"`.

![image](https://user-images.githubusercontent.com/5023024/202574670-833d40c9-fa2d-4912-9835-b873987c43de.png)

### WCAG Reference: 
- [1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html)
- [3.3.2 Labels or Instructions](https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html)
- [4.1.2 Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html) 

### Instance ID: 
 2036570417 

### Severity: 
 10 

### Description: 
 [Issue]
There are form controls with visual labels that are not programmatically associated with the control. Examples include:

- Child 1
- Child 2
- Child 3

[User Impact]
Screen reader users will have difficulty determining the purpose of these controls. Speech input users will have difficulty navigating to them.

[Code Reference] 
 ```
<div aria-checked="false" aria-label="" class="toggle" role="checkbox" tabindex="0"><svg class="check-svg" viewbox="0 0 16 16"><path d=""></path></svg></div>
 ``` 

### Note: 
 [Suggestion]
Provide a valid label for form fields. Remove aria-label attribute from ``` <div> ``` elements. The best way to do this is to use a ``` <label> ``` element that contains checkbox and its label.

[Compliant Code Example] 
 ```
<label class="checkbox-label">
  <calcite-checkbox class="checkbox" data-test-id="checkbox" guid="calcite-checkbox-ffffef30-cd34-98b7-5e82-99bbe9d57198" scale="m" calcite-hydrated="" checked="">
    <div aria-checked="false" class="toggle" role="checkbox" tabindex="0">
      <svg aria-hidden="true" class="check-svg" viewbox="0 0 16 16"><path d=""></path></svg>
    </div>
  </calcite-checkbox>
  Child 1
  <slot></slot>
</label>
 ```
 

### Media Type: 
 Forms 

### Additional Resources: 
- Report Source: Product accessibility evaluation conducted on the Calcite Design System in August, 2022.

geospatialem:
[...]

PR #9849, opened 2024-07-24
fix(tree,tree-item): replace checkbox with div and a11y attributes

aPreciado88:
**Related Issue:** [#5615](https://github.com/Esri/calcite-design-system/issues/5615)

## Summary
Update `tree-item` component to use check icon / div with applicable a11y attributes instead of `checkbox`,  when selection type is set to `ancestors`.


aPreciado88:
@SkyeSeitz @ashetland Could you please take a look at the chromatic build for `Tree` and see if it looks okay?

aPreciado88:
After discussing with design, two follow up issues came up. 

1. Update the `Card`, `Tile` and any other components where icon is serving as a checkbox, to use `--calcite-color-border-input` on their checkbox icon's idle state.
2. Add `scale="large"` examples to the `tree.html` demos page.

cc @geospatialem @brittneytewks

packages/calcite-components/src/components/tree-item/tree-item.tsx
```diff
@@ -227,17 +231,27 @@ export class TreeItem implements ConditionalSlotComponent, InteractiveComponent
 
     const checkbox =
       this.selectionMode === "ancestors" ? (
-        <label class={CSS.checkboxLabel} key="checkbox-label">
-          <calcite-checkbox
-            checked={this.selected}
+        <div class={CSS.checkboxContainer}>
+          <div
+            aria-checked={toAriaBoolean(this.selected)}
```
driskull, line 236: I think this `div` might need a role (maybe checkbox). I ran the test app through aXe toolbar and it showed this:

image

@geospatialem do you know what this might need?

I'm not sure we even need aria-checked here because the tree-item role has aria-selected so that should really account for selection to a screen reader right?

driskull (commented):
Code looks good just have some questions on the a11y and roles.

packages/calcite-components/src/components/tree-item/tree-item.tsx
aPreciado88: @driskull Adding `role="checkbox"` gets rid of the error above. Do we want to get rid of `aria-checked` entirely?

cc @geospatialem

packages/calcite-components/src/components/tree-item/tree-item.tsx
```diff
@@ -59,6 +59,9 @@ export class TreeItem implements ConditionalSlotComponent, InteractiveComponent
    */
   @Prop({ reflect: true }) disabled = false;
 
+  /** Accessible name for the component. */
+  @Prop() label: string;
```
geospatialem, line 63: Awesome!! 🙌🏻

packages/calcite-components/src/components/tree-item/tree-item.tsx
geospatialem: From testing, it looks like we should omit this `div` to only contain the `class` and `tabIndex`. For instance:

```tsx
<div class={CSS.checkboxContainer}>
  <div class={CSS.checkbox} tabIndex={-1}>
    <calcite-icon
      ...
```

geospatialem: Instead we'd add the contents to the `Host` `tree-item` element to announce the selection via `aria-live="polite"`:

```tsx
<Host
  aria-checked={this.selectionMode === "ancestors" ? toAriaBoolean(this.selected) : undefined}
  aria-expanded={this.hasChildren ? toAriaBoolean(isExpanded) : undefined}
  aria-hidden={toAriaBoolean(hidden)}
  aria-live="polite"
  calcite-hydrated-hidden={hidden}
  role="treeitem"
  tabIndex={this.disabled ? -1 : 0}
>
```

The above would also remove the `aria-selected` attribute, which currently contains:

```tsx
  ...
  aria-selected={this.selected ? "true" : showCheckmark ? "false" : undefined}
  ...
```

Not sure if there is an impact there, but I didn't notice one in testing.

The above two code changes allow JAWS and NVDA to hear the selections and live changes, in addition to passing all axe conditions.

packages/calcite-components/src/components/tree-item/tree-item.tsx
```diff
@@ -227,17 +231,27 @@ export class TreeItem implements ConditionalSlotComponent, InteractiveComponent
...
+            <calcite-icon
+              icon={
+                this.selected
+                  ? ICONS.checkSquareF
+                  : checkboxIsIndeterminate
+                    ? ICONS.minusSquareF
+                    : ICONS.square
+              }
+              scale={getIconScale(this.scale)}
+            />
```
geospatialem, line 251: Add `aria-hidden` so AT doesn't pick up the icon:

```suggestion
            <calcite-icon
            aria-hidden="true"
              icon={
                this.selected
                  ? ICONS.checkSquareF
                  : checkboxIsIndeterminate
                    ? ICONS.minusSquareF
                    : ICONS.square
              }
              scale={getIconScale(this.scale)}
            />
```

geospatialem (changes requested):
Have a few considerations to modify the changes at the host level, which also help AT pick up the changes live and fulfill the axe requirements.

aPreciado88:
@geospatialem I made the requested changes, please let me know if this looks okay!

packages/calcite-components/src/components/tree-item/tree-item.tsx
```diff
@@ -227,17 +231,22 @@ export class TreeItem implements ConditionalSlotComponent, InteractiveComponent
...
-            class={CSS.checkbox}
-            data-test-id="checkbox"
-            indeterminate={this.hasChildren && this.indeterminate}
-            scale={this.scale}
-            tabIndex={-1}
-          />
-          {defaultSlotNode}
-        </label>
+        <div class={CSS.checkboxContainer}>
+          <div class={CSS.checkbox} tabIndex={-1}>
```
driskull, line 235: Does this need a tabindex of 1? Are we ever focusing this element?

packages/calcite-components/src/components/tree-item/tree-item.tsx
```diff
@@ -280,9 +289,10 @@ export class TreeItem implements ConditionalSlotComponent, InteractiveComponent
 
     return (
       <Host
+        aria-checked={this.selectionMode === "ancestors" ? toAriaBoolean(this.selected) : undefined}
```
driskull, line 292: Just want to confirm aria-checked is ok instead of aria-selected. The treeitem role says either is ok.

> Either aria-selected or aria-checked can be used to indicate selection for treeitem elements. Some user interfaces indicate selection with aria-selected in single-select trees and with aria-checked in multi-select trees.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/treeitem_role

It seems like depending on the selectionMode it should use either aria-checked (multiple) or aria-selected (single)

driskull (changes requested):
Almost there! looking good @aPreciado88

packages/calcite-components/src/components/tree-item/tree-item.tsx
```diff
@@ -227,17 +231,22 @@ export class TreeItem implements ConditionalSlotComponent, InteractiveComponent
...
-            indeterminate={this.hasChildren && this.indeterminate}
-            scale={this.scale}
-            tabIndex={-1}
-          />
-          {defaultSlotNode}
-        </label>
+        <div class={CSS.checkboxContainer}>
+          <div class={CSS.checkbox} tabIndex={-1}>
+            <calcite-icon
+              aria-hidden="true"
```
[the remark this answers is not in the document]
geospatialem: Good catch, @driskull! 🎣

packages/calcite-components/src/components/tree-item/tree-item.tsx
geospatialem: @aPreciado88 It looks like based on @driskull comment above, there may be a need for both depending on the `selectionMode` type, where only one would be present for the `selectionMode` value.

packages/calcite-components/src/components/tree-item/tree-item.tsx
geospatialem: Refer to https://github.com/Esri/calcite-design-system/pull/9849#discussion_r1695931093, as this is not needed since its already honored on the `icon` component.

packages/calcite-components/src/components/tree-item/tree-item.tsx
aPreciado88: It looks like it only gets focused on click, should I remove tabIndex?

driskull: I think so, whats the reasoning for focusing it?

aPreciado88: We apply this css block when focusing the checkbox:

`.checkbox:focus {
  @apply outline-none;
  color: var(--calcite-color-brand);
}`

driskull: But this focus would only happen when clicking on that element.

Could we apply this style to the checkbox when the tree-item host is focused? Then it would be consistent.

driskull: Something like

```css
         :host(:focus:not([disabled])) .checkbox{ @apply outline-none; color: var(--calcite-color-brand); } 
```

Might also be nice to have on `hover` and `active` as well.

aPreciado88:
@driskull I updated with the requested updates 🚀

packages/calcite-components/src/components/tree-item/tree-item.tsx
```diff
@@ -289,10 +288,23 @@ export class TreeItem implements ConditionalSlotComponent, InteractiveComponent
 
     return (
       <Host
-        aria-checked={this.selectionMode === "ancestors" ? toAriaBoolean(this.selected) : undefined}
+        aria-checked={
```
driskull, line 291: Awesome! 👍 

Should we have an e2e test for this? Could be done as follow up issue due to time constraints.

packages/calcite-components/src/components/tree-item/tree-item.scss
```diff
@@ -128,6 +128,11 @@
   .node-container {
     @apply focus-inset outline-none;
   }
+
+  .checkbox {
```
driskull, line 132: I don't think this is necessary since the checkbox can't be focused right?

packages/calcite-components/src/components/tree-item/tree-item.tsx
aPreciado88: I updated this to apply the focused style when tree-item host is focused. 
Regarding the `hover` state, design previously reviewed this and they don't want any styles applied when items are hovered.

driskull: I think this one still needs to be addressed. 

1. Do we need/want the checkbox node to be focusable? (probably not).
2. Do we need hover/active/focus states on the host that style this? (maybe?)

packages/calcite-components/src/components/tree-item/tree-item.scss
driskull: Oh, i see what this is doing. Looks good 👍

packages/calcite-components/src/components/tree-item/tree-item.tsx
driskull: @aPreciado88 styles look good, but doesn't the `tabindex` of `-1` need to be removed?

packages/calcite-components/src/components/tree-item/tree-item.tsx
```diff
@@ -227,17 +231,21 @@ export class TreeItem implements ConditionalSlotComponent, InteractiveComponent
...
     const checkbox =
       this.selectionMode === "ancestors" ? (
-        <label class={CSS.checkboxLabel} key="checkbox-label">
-          <calcite-checkbox
-            checked={this.selected}
-            class={CSS.checkbox}
-            data-test-id="checkbox"
-            indeterminate={this.hasChildren && this.indeterminate}
-            scale={this.scale}
-            tabIndex={-1}
```
aPreciado88, line 237: @driskull Got rid of `tabIndex` in the latest update 🐱‍💻

packages/calcite-components/src/components/tree-item/tree-item.tsx
```diff
@@ -227,17 +231,21 @@ export class TreeItem implements ConditionalSlotComponent, InteractiveComponent
...
-          />
-          {defaultSlotNode}
-        </label>
+        <div class={CSS.checkboxContainer}>
+          <div class={CSS.checkbox}>
```
driskull, line 235: nitpick: do we need this div? maybe the styles could just be applied to the `calcite-icon` element?

driskull (approved):
🎉 👍 Looks good!

packages/calcite-components/src/components/tree-item/tree-item.tsx
aPreciado88: I removed the div, thanks for catching that!

packages/calcite-components/src/components/tree-item/tree-item.tsx
aPreciado88: I added e2e tests for this.

driskull:
Looks good!

packages/calcite-components/src/components/tree-item/tree-item.tsx
aPreciado88: @driskull Design reviewed the latest chromatic builds, and asked to remove the `focused` checkbox state. I pushed an update for that.

geospatialem (commented):
Hm, did something happen WRT child `tree-items`? It doesn't look like the `"Grandchild #"` `tree-items` are being depicted in most cases across the demo page.

image

https://github.com/Esri/calcite-design-system/blob/6b694016de9d222fd272c6ceeacb6369aaad11d9/packages/calcite-components/src/demos/tree.html#L386-L407

geospatialem (approved):
Resetting my status so we can merge the changes in. Will verify the fixes during the verification process! 💪🏻

Issue #9858, opened 2024-07-25
[Panel & Flow Item] add footer slot gaps settings


SkyeSeitz:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Add footer-end & footer-start slots according to scale

### Acceptance Criteria

**Provide horizontal gap setting within `footer-start` & `footer-end` slots according to scale**
small Panel: `space.sm` (8px) gap
medium Panel: `space.md` (12px) gap
large Panel: `space.lg` (16px) gap 

<img width="1655" alt="image" src="https://github.com/user-attachments/assets/7bbd140a-4a48-4b4f-b810-f6a1251604b4">
<img width="811" alt="image" src="https://github.com/user-attachments/assets/99f8db3f-097e-4999-befe-f9dbaed9c8fb">

**Provide vertical gap setting within `footer` slot according to scale**
small Panel: `space.sm` (8px) gap
medium Panel: `space.md` (12px) gap
large Panel: `space.lg` (16px) gap

<img width="1209" alt="image" src="https://github.com/user-attachments/assets/d016eb4a-d817-4076-aa7e-1f7d6b54a1dd">

### Which Component

Panel & Flow Item

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

driskull:
Looking at the styles, I'm seeing this in the DOM. Need to figure out why the CSS vars are not defined in this scope.

<img width="401" alt="image" src="https://github.com/user-attachments/assets/b7472d59-b324-4981-a860-af4e7b48bab6">

cc @jcfranco

DitwanP:
🍡 Verified locally on `dev`

Scales:
"s" - gap: 8px
![image](https://github.com/user-attachments/assets/6fde4c00-a524-4780-8a4c-358829aa2871)

"m" - gap: 12px
![image](https://github.com/user-attachments/assets/743cdd6d-c867-472a-bfe5-8475314683c6)

"l" - gap: 16px
![image](https://github.com/user-attachments/assets/70052b0f-272e-4c5a-a3ea-0757a3fe56d3)

Issue #9857, opened 2024-07-25
Panel: footer-end should appear at the end


driskull:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

footer-end slot appears centered if there is no footer-start slot

### Expected Behavior

footer-end should always appear at the end

### Reproduction Sample

https://codepen.io/driskull/pen/yLdawGa?editors=1010

### Reproduction Steps

1. open the codepen
2. notice how the slotted button is centered instead of at the end
3. uncomment the footer-start slotted button
4. Notice how it now looks correctly

### Reproduction Version

2.10.1

### Regression?

No

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

DitwanP:
🍡 Verified locally on `dev`
![image](https://github.com/user-attachments/assets/28c45b70-df33-49a9-9ec1-7b0a98e9bcdc)

PR #9868, opened 2024-07-26
fix(panel): correct footer padding and layout

driskull:
**Related Issue:** #9858 #9857

## Summary

- fixes footer-end slot alignment when its the only thing slotted
- refactor styles
- makes footer slot take precedence over the other footer slots (including deprecated footer-actions)
- add screenshot tests
- cleanup screenshot tests

### DIFF
--- a/packages/calcite-components/.storybook/preview-head.html
+++ b/packages/calcite-components/.storybook/preview-head.html
@@ -10,11 +10,6 @@
     background-color: var(--calcite-color-background);
   }
 
-  hr {
-    border: var(--calcite-border-width-sm) dashed var(--calcite-color-status-danger);
-    margin-block: 2rem;
-  }
-
   calcite-slider {
     width: 250px;
     max-width: 100%;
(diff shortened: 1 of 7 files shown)


jcfranco (approved):
Code changes LGTM!

Before merging, can you revisit the PR title so that it covers both changes as a single fix or feature? Otherwise, we'll need to split this up.

macandcheese (approved):
Looks good pending Chromatic approval from @SkyeSeitz @ashetland - Just confirming this gap is not expected to have a css prop for override. I think users can just place a single div and style that as needed for further customization.

SkyeSeitz:
> Looks good pending Chromatic approval from @SkyeSeitz @ashetland - Just confirming this gap is not expected to have a css prop for override. I think users can just place a single div and style that as needed for further customization.

Agreed - I think this makes sense since adding a div to adjust spacing was already the workflow. Now the default is just more accurate. I will remove the css props from the Figma specs.

SkyeSeitz (approved):

PR #9851, opened 2024-07-25
ci: add type checking to github scripts

benelan:
## Summary

- Add [`actions/github-script`](https://github.com/marketplace/actions/github-script) types
- Add [`@octokit/webhook-types`](https://www.npmjs.com/package/@octokit/webhooks-types) for improved payload typing
- Use JSDoc wizardry to typecheck the JavaScript files used by our workflows

## Notes

I got a husky error when installing the `github-script` types. Here is a related issue: https://github.com/typicode/husky/issues/851

Upgrading husky resolves the error, so I created a pull request: https://github.com/actions/github-script/pull/482

I installed `github-script` using the PR number for now. We can unpin the PR once it's merged by re-installing via the [command in the doc](https://github.com/actions/github-script/#use-scripts-with-jsdoc-support):

```sh
npm i -D @types/github-script@github:actions/github-script
```

PR #9877, opened 2024-07-29
ci: fix env var case in workflows

benelan:
**Related Issue:** #9851

## Summary

Fixes the case of environment variables in the workflow files, which I made uppercase in the GitHub scripts.


jcfranco (approved):
Lockfile changes probably need to be rolled back, but the rest LGTM! 🚀

Issue #9876, opened 2024-07-29
[Dialog] When `overlay-position="fixed"` and `menu-open` is present the menu is not positioned as-expected


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When `overlay-position="fixed"` and `menu-open` is present the menu is not positioned as-expected: 
![image](https://github.com/user-attachments/assets/67a4c677-d10e-45e2-9ab4-8f773d1848a4)

### Expected Behavior

The menu should be placed as-expected on the component (where it is initially loaded):

![image](https://github.com/user-attachments/assets/adacdd7e-0f5b-4076-b35f-9b0154a70bbc)

### Reproduction Sample

https://codepen.io/geospatialem/pen/ZEdLWNZ

### Reproduction Steps

1. Open the [sample](https://codepen.io/geospatialem/pen/ZEdLWNZ)
2. Observe the menu opens as-expected
3. However, it is repositioned in a different location

### Reproduction Version

`2.11.0-next.25`

### Relevant Info

Spurred from https://github.com/Esri/calcite-design-system/issues/7886 QA/QC

### Priority impact

impact - p1 - need for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

driskull:
This appears to be a an issue with the `transform` animation on the dialog. If we remove that, it works fine.

The issue is that having a transform animation with a floating-ui element inside of it can cause positioning issues.

Some notes about it can be read here: https://floating-ui.com/docs/usefloating#transform

In our case, a wrapper won't work because the floating ui element is inside of the transform animation.

As a workaround, I could disable the animation if the `menu-open` attribute is present. WDYT @jcfranco?

DitwanP:
🍡 Verified locally on `dev`

PR #9891, opened 2024-07-30
fix(dialog): fix menu positioning when when overlayPositioning is 'fixed' and menuOpen is true

driskull:
**Related Issue:** #9876

## Summary

- This corrects an issue with the transform animation on the dialog.  
  - The issue is that having a transform animation with a floating-ui element inside of it can cause positioning issues.
  - Some notes about it can be read here: https://floating-ui.com/docs/usefloating#transform
  - In our case, a wrapper won't work because the floating ui element is inside of the transform animation.
  - As a workaround, the transform animation is removed when menuOpen is true.

### DIFF
--- a/packages/calcite-components/src/components/dialog/dialog.scss
+++ b/packages/calcite-components/src/components/dialog/dialog.scss
@@ -162,6 +162,12 @@ calcite-panel {
   }
 }
 
+:host([menu-open]) .dialog {
+  transition:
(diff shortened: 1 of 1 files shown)


jcfranco (approved):

jcfranco:
Is it possible to add a screenshot test for this?

driskull:
There is a screenshot test for this but it doesn't reproduce the error.

driskull:
I think its because our screenshot tests have animations disabled

Issue #9242, opened 2024-05-02
Error during SSR


feus4177:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Throws an error during import.

### Expected Behavior

No errors are thrown during import.

### Reproduction Sample

SSR problem

### Reproduction Steps

Setup a default vite 5.1 build with the following added to the vite config
```
  ssr: {
    noExternal: [
      '@esri/calcite-components',
      '@stencil/core',
    ],
  },
  ```

### Reproduction Version

2.8.0

### Relevant Info

MacOS

```
ReferenceError: navigator is not defined
    at getUserAgentData (./node_modules/.pnpm/@esri+calcite-components@2.8.0/node_modules/@esri/calcite-components/dist/components/src/utils/browser.ts:9:5)
```

### Priority impact

p4 - not time sensitive

### Impact

Unable to use calcite-components. Will probably have to use a different library.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

**Comment by geospatialem** [comment]
@feus4177 Can you supply a reproduction sample or a scaffolding command for us to diagnose further? 

We're currently blocked by a Stencil issue, https://github.com/ionic-team/stencil/issues/4010, so we may be unable to mitigate depending on the instance.

**Comment by geospatialem** [comment]
@feus4177 Following up if you can supply us with a repro sample to diagnose further.

**Comment by feus4177** [comment]
@geospatialem, sorry it took so long to get back to you. [Here is a repo](https://github.com/feus4177/calcite-design-system-ssr-repro-9242) with a minimal reproduction. From the root directory of the repo, run `npm run dev` and open up the localhost URL that it prints out.

 I now get a slightly different error: `ReferenceError: window is not defined`. I seem to recall that I was able to get past this (I think by importing from a slightly different path).
 
 The only unusual aspect of my code is the update to the vite.config.js file where I added the 
 ```
   ssr: {
    noExternal: [
      '@esri/calcite-components',
      '@stencil/core',
    ],
  },
```
as mentioned above. Without that config, I get an error about importing CommonJS modules from an ES module.

jcfranco:
Verified locally.

PR #9897, opened 2024-07-30
fix: improve browser check to resolve SSR errors

benelan:
**Related Issue:** #9242

## Summary

Add additional browser checks to resolve some SSR errors that Stencil's `Build.isBrowser` util doesn't catch.


packages/calcite-components/src/utils/browser.ts
```diff
@@ -1,5 +1,14 @@
 import { Build } from "@stencil/core";
 
+export const isBrowser = (): boolean =>
```
driskull, line 3: 🆒

driskull (approved):
👍

PR #9907, opened 2024-07-30
build(deps): drop unnecessary ESM resolution build patch

benelan:
## Summary

Drop the ESM resolution build patch, which is no longer necessary after Stencil v4.19


jcfranco (approved):

geospatialem:
cc #9908

Issue #6298, opened 2023-01-17
Popover support for trigger events such as click (`Enter`) and hover (`:focus`)


geospatialem:
### Description
The ArcGIS Urban team needs the following kinds of non-modal components (using tooltips/popovers):

- [x] Displays on hover (and on focus for keyboard accessibility compliance) and contains a simple text → `tooltip`
- [x] Displays on click (and on Enter key press, for accessibility compliance as well) contains more complex content such as links, Checkbox, Input Date Picker, Button etc. → `popover`
- [ ] Displays on `:hover` (and `:focus`) only and contains more complex content → `popover` (refer to the acceptance criteria below)

### Acceptance Criteria
Open popover with different types of trigger events like in [`type` in `dropdown`](https://developers.arcgis.com/calcite-design-system/components/dropdown/#component-api-properties-type) with the following options:

- Click and the Enter key or
- Hover and when focused
    - If this option is selected there should be a way that the click and enter key events do not toggle the component's open prop.

#### Use case screenshots:
![image](https://user-images.githubusercontent.com/5023024/213204733-364fe1ca-2a03-4180-a003-5ba94ae2fb32.png)

![image](https://github.com/Esri/calcite-components/assets/5023024/f6a68d2a-998d-4ac9-8c49-74156419b4d7)

### Relevant Info
Would be similar in functionality to the [`type` in Dropdown](https://developers.arcgis.com/calcite-design-system/components/dropdown/#component-api-properties-type).

Accessibility and documentation around its accessibility will need to be considered to ensure the popover contents are provided to assistive technologies. In the [following example] the popover contents are only provided to assistive technologies when in focus (not while hovered, which could be provided via an `aria-live` prop.

### Which Component
`popover`

### Example Use Case
The following [Codepen serves as a hacky example](https://codepen.io/geospatialem/pen/MWBEvMX) but disables the focus trap using the `focusTrapDisabled` property, which is the expected behavior for assistive technologies. It also still includes the ability to `open` the component with a click or `Enter` key event.

### Esri team
ArcGIS Urban

Anastasiia-Boleiko:
Hello,

Could you please clarify what is the approximate time for having this feature in the Calcite release? 

Thank you.

geospatialem:
> Hello,
> 
> Could you please clarify what is the approximate time for having this feature in the Calcite release?
> 
> Thank you.

[...]

PR #9914, opened 2024-07-30
feat(tooltip): allow focusing and clicking on interactive elements within a tooltip

driskull:
**Related Issue:** #6298

## Summary

- Allows placing interactive elements in tooltip elements with the following caveats
  - It should be advised against placing interactive elements within a tooltip because there is no a11y guidelines for accessing interactive elements within a tooltip role. Ideally a popover should be used for interactive elements.
  - No focus trapping or focus will be brought to interactive elements within the tooltip.
  - You can tab to a focusable element within a tooltip if the tooltip with the focusable element is the next focusable element within the DOM order. Otherwise, you could listen for the tooltip to be open and focus on the first interactive element if you wanted to do so.
- add test 
- add story


jcfranco (approved):
This looks great, @driskull! ✨🚀🪐🛰️ ✨

Based on the PR description, do we need to update the docs to emphasize the interactive content caveat along with additional context? cc @geospatialem @DitwanP

driskull:
> Based on the PR description, do we need to update the docs to emphasize the interactive content caveat along with additional context

I'm not sure we want to encourage putting interactive elements within a tooltip so I don't recommend promoting or documenting this use case. It will work great for mouse users but probably not for screen reader users.

This PR handles the use case and gets interactive elements working in tooltips. I just don't think we should promote or mention the use case due to the a11y concerns.

PR #9915, opened 2024-07-30
fix: widen icon type to allow string

jcfranco:
**Related Issue:** #8858

## Summary

This widens the type used for all icon props to allow string as a fallback. This prevents breaking changes that would require the new type to be used.


benelan (approved):
:+1:

Issue #9731, opened 2024-07-03
bug(calcite-dropdown): .setFocus() does not move focus to dropdown


maxpatiiuk:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Calling `.setFocus()` on `<calcite-dropdown>` has no effect

### Expected Behavior

Calling `.setFocus()` on `<calcite-dropdown>` and awaiting the promise will move focus to `<calcite-dropdown>`. OR, it will move focus to the element inside the dropdown that has `[slot="trigger"]`

### Reproduction Sample

https://codepen.io/maxpatiiuk/pen/wvbVVgM?editors=1010

### Reproduction Steps

1. Open the codepen
2. See that the code is calling .setFocus() on the dropdown
3. Check the console - see that the document.activeElement after awaiting .setFocus() is still the body element, rather than the dropdown

### Reproduction Version

2.10.1

### Relevant Info

Internally, looks like Calcite is calling `this.el.focus()` correctly, but that has no effect as the calcite-dropdown is not focusable.
Adding `tabindex="-1"` to the `<calcite-dropdown>` "in userland" does not fix this. However, adding `tabindex="-1"` to the `.interaction-container` inside the `<calcite-dropdown>` seems to fix `.setFocus()`

### Regression?

<2.6.0 ??

### Priority impact

impact - p3 - not time sensitive

### Impact

We have a workaround:

```ts
calciteDropdown.querySelector('[slot="trigger"]')?.setFocus();
```

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

(internal) [DevTopia issue](https://devtopia.esri.com/WebGIS/arcgis-js-api/pull/63458#issuecomment-4866297)

geospatialem:
Related issue for October: https://github.com/Esri/calcite-design-system/issues/8206. Spike to determine feasibility before R3.

DitwanP:
🍡 Verified on `2.12.0-next.4`
https://codepen.io/Ditwan-Price/pen/XWLgbZZ?editors=1011

PR #9917, opened 2024-07-31
fix(dropdown): ensure `setFocus` focuses the trigger

jcfranco:
**Related Issue:** #9731

## Summary

`delegatesFocus` will only focus elements in the shadow DOM (see [example](https://codepen.io/jcfranco/pen/MWMpaOM)), so this ensures the slotted trigger gets focused.

`focusable` test was incorrectly passing with the pinned version of Puppeteer (21.5.0) for both headless and headful modes. Testing the previous implementation with `puppeteer@22.14.0` (latest) fails the test properly.


driskull (approved):
👍 LGTM!

aPreciado88 (approved):
🚀

Issue #9920, opened 2024-07-31
[Block] Spacing not honored on the component with `heading` present


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

`block`'s `heading` prop is not reflecting the CSS attributes and spacing is not honored on the component.

![image](https://github.com/user-attachments/assets/f5773f8c-e85b-4758-9ce0-58fb8a316ea1)

### Expected Behavior

Reflected CSS properties when the component's `header` is provided.

![image](https://github.com/user-attachments/assets/12893ad0-9868-4d08-b7f2-383f4988c080)

### Reproduction Sample

https://codepen.io/driskull/pen/eYwvJEg

### Reproduction Steps

1. Open the [sample](https://codepen.io/driskull/pen/eYwvJEg)
2. Observe the spacing

### Reproduction Version

`2.11.0`

### Regression?

`2.10.1`

### Priority impact

impact - p0 - emergency

### Impact

Online teams releasing for Enterprise have a code freeze starting next week.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

geospatialem:
Verified in `2.11.1-next.0` when the component's `heading` is present:

![image](https://github.com/user-attachments/assets/09869e65-1e71-478b-af9c-c71fefc18040)

PR #9924, opened 2024-07-31
fix(block): display correct header spacing when heading or description are present

aPreciado88:
**Related Issue:** [#9920](https://github.com/Esri/calcite-design-system/issues/9920)

## Summary
Update `block` component to display correct header spacing when `heading` or `description` properties are present.


packages/calcite-components/src/components/block/block.scss
```diff
@@ -194,7 +198,6 @@ calcite-action-menu {
   }
 }
 
-:host([heading]),
 :host([description]) {
```
driskull, line 201: [description] would also be an issue here because it isn't reflected either. Should this css be removed?

packages/calcite-components/src/components/block/block.tsx
```diff
@@ -416,7 +416,7 @@ export class Block
     const toggleLabel = open ? messages.collapse : messages.expand;
 
     const headerContent = (
-      <header class={CSS.header} id={IDS.header}>
+      <header class={{ [CSS.header]: true, [CSS.headerPadding]: !!heading }} id={IDS.header}>
```
driskull, line 419: should this be `!!(heading || description)`?

```suggestion
      <header class={{ [CSS.header]: true, [CSS.headerPadding]: !!(heading || description) }} id={IDS.header}>
```

driskull (changes requested):
Changes look good, just missing a few things 👍

packages/calcite-components/src/components/block/block.scss
```diff
@@ -23,6 +23,10 @@
   @apply justify-start;
 }
 
+.header-padding {
```
driskull, line 26: nitpick: maybe this should be named something like `.header--present` or `.header--enabled`?

aPreciado88:
@driskull I noticed that if I update the `heading` prop in `block.tsx` to `@Prop({ reflect: true }) heading!: string;` then the css will work, without any of the changes above. Is there a reason why it's declared as `@Prop() heading!: string;` ? 
`description` is also declared as `@Prop() description: string;`. Should the fix be to add `{ reflect: true }` to both props?

driskull:
@aPreciado88 no, we don't want to reflect all properties. In cases where the property is just content , we don't want to reflect. https://github.com/Esri/calcite-design-system/tree/dev/packages/calcite-components/conventions#reflecting-to-attributes

aPreciado88:
@driskull I updated with the requested changes.

driskull (approved):
@aPreciado88 can we add a e2e test that checks for the css class when a heading or description is present via their properties?

Changes look good to me. Lets wait for @jcfranco to review before merging.

packages/calcite-components/src/components/block/block.scss
```diff
@@ -23,6 +23,10 @@
   @apply justify-start;
 }
 
+.header--present {
```
jcfranco, line 26: Nitpick: I believe we use `.<part>--has-text` in other components for similar contexts.

jcfranco (approved):
Once we have screenshot test coverage, this is good to merge! 🚀

packages/calcite-components/src/components/block/block.e2e.ts
```diff
@@ -323,6 +323,29 @@ describe("calcite-block", () => {
       const actionAssignedSlot = await page.$eval("calcite-action", (action) => action.assignedSlot.name);
       expect(actionAssignedSlot).toBe(SLOTS.headerMenuActions);
     });
+
+    it("applies correct header spacing when heading or description properties are present", async () => {
```
jcfranco, line 327: This is fine for now until we add extra checks to ensure styles use reflected props. After that, we should rely on screenshot tests. From the assertion perspective, it is not guaranteed that changes in the CSS class will be detected by this test, meaning its implementation could change without causing a test failure. cc @driskull

aPreciado88: @jcfranco Should I create a follow up issue to add a screenshot test?

aPreciado88: @jcfranco I created this follow up [issue](https://github.com/Esri/calcite-design-system/issues/10084) to remove this E2E test once the reflected prop checks are in place.

Issue #5697, opened 2022-11-07
Tooltips no longer move along x/y axis when transitioning to the `open` state for the first time


nwhittaker:
### Actual Behavior

The first time a tooltip's reference element is hovered, the tooltip appears abruptly in place. Subsequent hoverings of the reference element show a smoother transition of the tooltip from hidden to visible along the x or y axis (depending on how the tooltip is positioned relative to the reference element).

### Expected Behavior

The tooltip transitions in smoothly along the x or y axis the first time its reference element is hovered.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/ExRNKoZ

### Reproduction Steps

1. Visit the code sample
2. Hover the text and note the tooltip appears in place (it may help to slow down animations in devtools)
3. Hover the text again and note the tooltip now transitions along the x or y axis when showing/hiding.

### Reproduction Version

beta.97

### Regression?

next.606

### Impact

For performance reasons, Field Maps generates some tooltips "just in time" when the reference element is hovered. As a result, the way these tooltips visually appear is inconsistent with how other tooltips in the app appear.

### Esri team

ArcGIS Field Apps

geospatialem:
Will determine if other floating UI fixes, such as #6520 improve this bug.

driskull:
@nwhittaker is this issue still present in the latest release?

nwhittaker:
@driskull, I updated the pen in the description to 1.7.0 and still see the originally observed behavior.

driskull:
Thanks for confirming!

jcfranco:
This should be fixed now via https://github.com/Esri/calcite-design-system/pull/8001 (see [updated repro case](https://codepen.io/jcfranco/pen/yLZZgBP)).

@nwhittaker Can you confirm?

nwhittaker:
I updated a [fork of the repro][1] to `2.0.0` and the issue looks resolved to me. Thanks!

[1]: https://codepen.io/nwhittaker-esri/pen/ExrrpgO

geospatialem:
Verified per Nate's comment [above](https://github.com/Esri/calcite-design-system/issues/5697#issuecomment-1845567472).

driskull:
Reopened issue and added triage label. see #8409 for reasoning.

geospatialem:
Spike to explore the transitions of the `tooltip` upon opening as it relates to overall performance of the component. 

[...]

PR #9937, opened 2024-08-01
fix: improve all initial floating-ui positioning

driskull:
**Related Issue:** #5697

## Summary

- switch from using transform to inset. transform was conflicting with the floating-ui positioning
- `visibility` transition removed because theres no transition for visibility and it was messing with positioning
- animation now works on first run
- cleanup


jcfranco (approved):
This is pretty sweet, @driskull! ✨🍭✨

Since this will improve all initial floating-ui positioning, can we update the PR title to reflect this?

PR #9941, opened 2024-08-01
refactor: clean up floating ui css and tooltip e2e test.

driskull:
**Related Issue:** N/A

## Summary

🧹 clean up


packages/calcite-components/src/assets/styles/_floating-ui.scss
```diff
@@ -1,17 +1,19 @@
+$floating-ui-transition-offset: 5px;
```
driskull, line 1: @alisonailea I wonder if we should have this as some kind of global token instead of a Sass var?

driskull: Or maybe one of these could be used?

```
  --calcite-spacing-xs: 6px;
  --calcite-spacing-xxs: 4px;
```

```
--calcite-size-xs: 6px;
  --calcite-size-xxs: 4px;
```

jcfranco (approved):

Issue #8784, opened 2024-02-20
[Accordion] In Safari, multiple Accordion Item elements can be visibly focused, and buggy animation


macandcheese:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

In Safari, multiple Accordion Item components can remain visibly focused:
![Screenshot 2024-02-20 at 10 31 30 AM](https://github.com/Esri/calcite-design-system/assets/4733155/e18978f2-c0d1-4d08-960e-56781316cb04)

Additionally, there is a "size wiggle" and slow animation in Safari. We should just remove any focus transition here to match other components:

https://github.com/Esri/calcite-design-system/assets/4733155/cca87549-33ce-41c8-b0b5-f3ad78da8e25

### Expected Behavior

I'd expect only one to be visible focused at once.

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/accordion/

### Reproduction Steps

1. Open the component documentation page in Safari
2. Interact with the sample and see the above behavior

### Reproduction Version

2.5.1

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

macandcheese:
Seems similar to #8617

jcfranco:
@macandcheese I wasn't able to reproduce this in Safari 17.5 (19618.2.12.11.6). Do you have any more info on how to reproduce?

macandcheese:
Recorded this in Version 17.5 (19618.2.12.11.6) just now - definitely something wonky going on - some clicks don't apply visual focus - subsequent tab / click does in fact focus at times and you can reach the state in the original issue after a bit....

https://github.com/user-attachments/assets/25689e71-225e-4ebc-a995-207f2f26f4f7

jcfranco:
Thanks for the vid. I think I know what's going on. 🕵️❕

DitwanP:
🍡 Verified on `2.12.0-next.8`
https://codepen.io/Ditwan-Price/pen/QWXMyPR?editors=100

PR #9944, opened 2024-08-02
fix(accordion-item): align focus styles across browsers

jcfranco:
**Related Issue:** #8784

## Summary

Update focus styles so that they target the focusable element instead of the host, which does not use tabindex currently.

### DIFF
--- a/packages/calcite-components/src/components/accordion-item/accordion-item.scss
+++ b/packages/calcite-components/src/components/accordion-item/accordion-item.scss
@@ -38,13 +38,12 @@
 }
 
 // focus styles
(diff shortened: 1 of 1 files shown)


jcfranco:
@macandcheese @ashetland @SkyeSeitz Could y'all help me do some early testing for [this one](https://6266d45509d7eb004aa254fb-wplcjzzapf.chromatic.com/?path=/story/components-accordion--simple)? 🙇‍♂️

ashetland:
I'm not noticing any differences. LGTM.

driskull (approved):
👍

PR #9945, opened 2024-08-02
perf(input-time-zone): improve memory footprint by leveraging browser cache and releasing utils after their use

jcfranco:
**Related Issue:** N/A

## Summary

✨🚀✨

### DIFF
--- a/packages/calcite-components/src/components/input-time-zone/utils.ts
+++ b/packages/calcite-components/src/components/input-time-zone/utils.ts
@@ -56,23 +56,6 @@ export function getUserTimeZoneName(): string {
   return dateFormatter.resolvedOptions().timeZone;
 }
 
-/**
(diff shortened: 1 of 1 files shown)


jcfranco:
Removing whitespace is recommended for this one: https://github.com/Esri/calcite-design-system/pull/9945/files?w=1.

benelan (approved):
Nice cleanup!! I thought we had a convention about not using abbreviations (tzs vs timezones, Idx vs Index) but now I can't find it, so LGTM!

jcfranco:
@benelan We do have [a rule](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/.eslintrc.cjs#L147) to avoid abbreviations, but `tzs` and `idx` are coming from the `timezone-groups` util results.

PR #9948, opened 2024-08-02
fix(list-item): prevent scrolling when item is focused.

driskull:
**Related Issue:** #9949

## Summary

- fixes scrolling when a list item is focused


jcfranco (approved):
🚀

PR #9951, opened 2024-08-02
ci(lerna): pin versions

benelan:
**Related Issue:** #9946

## Summary

Use the exact version (no semver range) when releasing `next` via Lerna to prevent Renovate from creating PRs to pin the version.


jcfranco (approved):
🔎👁️👄👁️👍

Issue #8678, opened 2024-01-31
[calcite-loader] Allow a way to display percent sign on the determinate value


rpanichakit9541:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Currently the value of the determinate loader just displays as a number which the user can mistake as the count number of items done instead of the percentage of completion.  This is a request to be able to set a prop to tell the calcite-loader component to display the percent sign next to the number value.

### Acceptance Criteria

The user should be able to set a prop on the calcite-loader component to tell it to display a percent sign next to the percent value

### Which Component

calcite-loader

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Online

mpriour:
This issue is now impacting us in ArcGIS Online. We would really like this enhancement before the fall release of ArcGIS Online if at all possible.

geospatialem:
For design expertise 🧠 : Should we also support the `"%`"` when the value is `100`?

geospatialem:
Discussed [the above](https://github.com/Esri/calcite-design-system/issues/8678#issuecomment-2158983572) with Skye, where they recommend the following:

> We should always display the percentage for the value, even when the value is at 100%. However, it is recommended once the component reaches 100% to remove the loading state.
> 
> Additionally, in looking at other design systems, which use a circular progress similar to our determinate loader, it is common to not have the percentage by default, so we should explore adding the percentage as a new property.
>
> Resources:
> - https://v2.chakra-ui.com/docs/components/circular-progress
> - https://quasar.dev/vue-components/circular-progress
> - https://design.wonderflow.ai/get-started/components/loading/circular-progress

SkyeSeitz:
Would like to propose closing in favor of [8735](https://github.com/Esri/calcite-design-system/issues/8735), where we are exploring renaming the current Loader type value to `determinate-numeric` which displays a number out of 100 (there was not sufficient space to warrant adding a percentage suffix) and adding an additional Loader type value that is just `determinate`, which removes the number to avoid confusion. 
[...]

PR #9957, opened 2024-08-03
feat(loader): add `determinate-value` type

jcfranco:
**Related Issue:** #8678

## Summary

Adds new determinate type to allow users to opt-into displaying the progress (percent) label.

**Note:** the current `determinate` type will no longer display a label in a future release.

### DIFF
--- a/packages/calcite-components/src/components/loader/loader.tsx
+++ b/packages/calcite-components/src/components/loader/loader.tsx
@@ -98,9 +98,9 @@ export class Loader implements LocalizedComponent {
               <circle {...svgAttributes} />
             </svg>
           ))}
+          {isDeterminate && <div class={CSS.loaderPercentage}>{this.formatValue()}</div>}
         </div>
         {text && <div class={CSS.loaderText}>{text}</div>}
-        {isDeterminate && <div class={CSS.loaderPercentage}>{this.formatValue()}</div>}
       </Host>
     );
   }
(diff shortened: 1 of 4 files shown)


packages/calcite-components/src/components/loader/loader.tsx
```diff
@@ -21,6 +22,9 @@ export class Loader {
   /** Accessible name for the component. */
   @Prop() label!: string;
 
+  /** When `true`, and when the type is `determinate`, displays the progress value as a percentage */
+  @Prop({ reflect: true }) percentage = false;
```
macandcheese, line 26: Could this be confused with `value`? The description mentions "percent". Maybe `displayType` with two values, defaulting to the current one? It's not quite the same but `percent` is a value of a similar property in Meter: https://developers.arcgis.com/calcite-design-system/components/meter/#api-reference-properties-rangeLabelType

jcfranco: I can see that. WDYT about:

* `valueType: "number" | "percent"` (similar to [`meter#valueLabelType`](https://developers.arcgis.com/calcite-design-system/components/meter/#api-reference-properties-valueLabelType))
* `valueStyle: "number" | "percent"` (similar to [`input-time-zone#offsetStyle`](https://developers.arcgis.com/calcite-design-system/components/input-time-zone/#api-reference-properties-offsetStyle))

?

driskull (approved):
⭐ Looks good!

Issue #9641, opened 2024-06-20
Calcite Sheet `calciteSheetBeforeOpen` happens after the animation is done


hcampos-professional:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When the `<calcite-sheet>` is opened, its [calciteSheetBeforeOpen](https://developers.arcgis.com/calcite-design-system/components/sheet/#api-reference-events-calciteSheetBeforeOpen) event is fired only after the transition/animation has happened.

### Expected Behavior

As per the [documentation]([calciteSheetBeforeOpen](https://developers.arcgis.com/calcite-design-system/components/sheet/#api-reference-events-calciteSheetBeforeOpen)) the event should be fired before the transition starts, and [calciteSheetOpen](https://developers.arcgis.com/calcite-design-system/components/sheet/#api-reference-events-calciteSheetOpen) should be called after it finishes.

### Reproduction Sample

https://codepen.io/hccampos/pen/pomLvZo

### Reproduction Steps

1. Open https://codepen.io/hccampos/pen/pomLvZo
2. Click "Open sheet"
3. Notice that when the alert is displayed, the transition is already completed.

### Reproduction Version

2.9.x

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

geospatialem:
Related consistency issue for events: https://github.com/Esri/calcite-design-system/issues/9315

DitwanP:
🍡 Verified on `2.12.0-next.18`
https://codepen.io/Ditwan-Price/pen/OJezmaw

Issue #9315, opened 2024-05-10
`input-time-zone` emits multiple open/close events when toggling `open`


nwhittaker:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When enabling the `input-time-zone` component's `open` prop, events are emitted in this order:
1. `calciteInputTimeZoneOpen`
2. `calciteInputTimeZoneBeforeOpen`
3. `calciteInputTimeZoneOpen`

When disabling the `input-time-zone` component's `open` prop, events are emitted in this order:
1. `calciteInputTimeZoneClose`
2. `calciteInputTimeZoneBeforeClose`
3. `calciteInputTimeZoneClose`

### Expected Behavior

Presumably the first `calciteInputTimeZoneOpen` and `calciteInputTimeZoneClose` events never occur.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/BaeaXom

### Reproduction Steps

1. Visit the sample and open the console
2. Click the **Toggle time zone input** button a few times and see the order of emitted events appear in the console

### Reproduction Version

2.8.0

### Relevant Info

Encountered while investigating a workaround for [the issue][1] where opened overlays do not dismiss when clicking outside of them but on an element that stops event propagation.

[1]: https://github.com/Esri/calcite-design-system/issues/9013

### Priority impact

p3 - want for upcoming milestone

### Impact

Impact is minimal since we don't currently make use of these `input-time-zone` events.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Field Apps

jcfranco:
Picking this one up since I found out the cause of this while working on #9018. I think this one is doable for this upcoming release. cc @geospatialem

geospatialem:
Reallocated to August for additional testing and follow-up! 💪🏻

DitwanP:
🍡 Verified on `2.12.0-next.16`
https://codepen.io/Ditwan-Price/pen/GRbyrZY?editors=1011

nwhittaker:
@DitwanP, can we reopen this issue? I'm not seeing any change in behavior in your codepen nor `2.12.1` when I use the toggle buttons to open/close the time zone dropdown:

[...]

PR #9958, opened 2024-08-03
fix: ensure open/close events emit correctly if target element has multiple transitioned props

jcfranco:
**Related Issue:** #9641, #9315

## Summary

Fixes an issue in the openClose util that caused the incorrect duration to be used for event emitting.

### DIFF
--- a/packages/calcite-components/src/components/sheet/sheet.e2e.ts
+++ b/packages/calcite-components/src/components/sheet/sheet.e2e.ts
@@ -1,6 +1,6 @@
 import { newE2EPage } from "@stencil/core/testing";
 import { html } from "../../../support/formatting";
-import { focusable, renders, hidden, defaults, accessible } from "../../tests/commonTests";
+import { accessible, defaults, focusable, hidden, openClose, renders } from "../../tests/commonTests";
 import { GlobalTestProps, newProgrammaticE2EPage, skipAnimations } from "../../tests/utils";
 import { CSS } from "./resources";
 
@@ -75,6 +75,18 @@ describe("calcite-sheet properties", () => {
     describe("focuses content by default", () => {
       focusable(createSheetHTML(focusableContentHTML), {
         focusTargetSelector: `.${focusableContentTargetClass}`,
+      });
+    });
+  });
+
+  describe("openClose", () => {
+    describe("default", () => {
+      openClose("calcite-modal");
+    });
+
+    describe("initially open", () => {
+      openClose("calcite-modal", {
+        initialToggleValue: true,
       });
     });
   });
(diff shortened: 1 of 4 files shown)


packages/calcite-components/src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -184,14 +182,13 @@ export class InputTimePicker
...
-    onToggleOpenCloseComponent(this);
-
     if (this.disabled || this.readOnly) {
       this.open = false;
       return;
     }
 
-    this.reposition(true);
+    // we set the property instead of the attribute to ensure popover's open/close events are emitted properly
+    this.popoverEl.open = this.open;
```
driskull, line 191: What if the time picker is open by default? Would this still happen? I'm not seeing anything set the popover to be open except in this watcher which wouldn't run initially.

packages/calcite-components/src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -978,9 +970,6 @@ export class InputTimePicker
   async componentWillLoad(): Promise<void> {
     setUpLoadableComponent(this);
     await Promise.all([setUpMessages(this), this.loadDateTimeLocaleData()]);
-    if (this.open) {
-      onToggleOpenCloseComponent(this);
```
driskull, line 982: With this removed, can the popover be open by default?

packages/calcite-components/src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -1046,9 +1035,10 @@ export class InputTimePicker
             id={dialogId}
             label={messages.chooseTime}
             lang={this.effectiveLocale}
+            onCalcitePopoverBeforeClose={this.popoverBeforeCloseHandler}
+            onCalcitePopoverBeforeOpen={this.popoverBeforeOpenHandler}
             onCalcitePopoverClose={this.popoverCloseHandler}
             onCalcitePopoverOpen={this.popoverOpenHandler}
-            open={this.open}
```
driskull, line 1051: With this removed, can the popover be open by default?

packages/calcite-components/src/components/input-time-picker/input-time-picker.tsx
jcfranco: Ah, great catch. 🎣 Will fix!

jcfranco: Fixed! Added a test for this as well. ✨🔧✨

driskull (approved):
Looks good! ⭐

PR #9966, opened 2024-08-05
chore: temporarily disable broken icon diff check

benelan:
## Summary

Temporarily disable the "Icon Team Diff Check" workflow, which is currently broken. I will re-enable it once I have time to fix it in the next few days.


benelan:
Pushing this one through to prevent PR check failures, cc @jcfranco

Issue #9387, opened 2024-05-21
Date selected in input-date-picker gets cleared for "bs", "it-CH" locales


annierm18:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The selected date gets cleared (and is replaced by the placeholder) when the user clicks out of the component for the "bs", "it-CH" locales.

### Expected Behavior

- [ ] The selected date would remain, like it does for the "en" locale
- [ ] Add test coverage for future support

### Reproduction Sample

https://codepen.io/annierm18/pen/YzbqdXO?editors=100

### Reproduction Steps

1. Open input-date-picker in codepen
2. Add the lang="bs" or "it-CH" tag
3. Select a date from the calendar
4. Click outside of the component

### Reproduction Version

2.8.2

### Relevant Info

There is a corresponding dev issue for this: https://devtopia.esri.com/WebGIS/arcgis-enterprise-manager/issues/1758. It says "The issue is reproducible on Chrome and Edge Chromium"

This next line is only being put here in order to verify an automation, ignore it.
Blocked issues: #9722

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Enterprise

geospatialem:
Added the `p - high` label as suggested by @annierm18 and the i18n team.

Akochure:
Thanks, @annierm18, for logging the issue. We have used the temporary fix based on Blake's comment - https://github.com/Esri/calcite-design-system/issues/4583#issuecomment-1156807166. But need to fix this issue in the component.

anveshmekala:
Did some digging and this is expected for few locales since we are relying `toLocaleDateString` API which is implementation specific.  Safari and firefox handles `bs` locale as expected but not `chrome`.  `it-CH` locale has reported issue in CLDR https://unicode-org.atlassian.net/browse/CLDR-16491 for date separator.

Quick workaround would be to fallback to `bs-Cyrl` for `bs` locale and `de-CH` locale for `it-CH` for the string conversion part. 

Using `toLocaleDateString` as source of truth for string conversion of locale specific dates  going forward needs to be reviewed .

[...]

PR #9976, opened 2024-08-06
fix(input-date-picker): fix selection for the "bs" and "it-CH" locales

driskull:
**Related Issue:** #9387

## Summary

- the `bs` and `it-CH` locales do not format dates properly due to i18n bugs and inconsistencies
  - https://issues.chromium.org/issues/40172622
  - maps `bs` to `bs-Cyrl`
  - maps `it-CH` to `de-CH`
  - This is a workaround. We should consider moving away from relying `Intl` formatting in order to be able to parse dates since there are `Intl` bugs present. I would propose having 3 separate inputs for Date (day, month, year) so we can reliably parse the strings into dates.
- uses `Intl.DateTimeFormat.format()` instead of `toLocaleString()`
  - > very time toLocaleString is called, it has to perform a search in a big database of localization strings, which is potentially inefficient. When the method is called many times with the same arguments, it is better to create a [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) object and use its [format()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/format) method, because a DateTimeFormat object remembers the arguments passed to it and may decide to cache a slice of the database, so future format calls can search for localization strings within a more constrained context.
  - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString
- adds spec test
- adds e2e test
- adds storybook screenshot test


packages/calcite-components/src/utils/locale.ts
```diff
@@ -175,6 +175,23 @@ export function getSupportedLocale(locale: string, context: "cldr" | "t9n" = "cl
   return locale;
 }
 
+/**
+ * Gets the locale that best matches the context for date formatting.
```
jcfranco, line 179: Could you add a bit of info for this workaround? Linking the issue could also be helpful.

packages/calcite-components/src/utils/locale.spec.ts
```diff
@@ -230,3 +231,24 @@ describe("getSupportedLocale", () => {
     });
   });
 });
+
+describe("getDateFormatSupportedLocale", () => {
+  function assertAllContexts(locale: string, expectedLocale: string): void {
```
jcfranco, line 236: This function name is a bit vague. Can you revisit it for clarity?

jcfranco (approved):
Noice! 😎

Issue #8737, opened 2024-02-13
[combobox-item] Update component's props for consistency


DitwanP:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

1. Deprecate textLabel in favor of heading
2. Add heading prop, which will fallback to textLabel if unspecified
3. Add label, which will fallback to value if not specified

2 and 3 naming would align with [#8737](https://github.com/Esri/calcite-design-system/issues/8737)

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

geospatialem:
We should consider adding a `label` property, where we deprecate the `textLabel` property.

`label` should be required, where `value` should be optional.

_Of note, if we add in a `description` property, the `label` (or "heading" takes precedence for AT)._

jcfranco:
Sharing for posterity, for this release we would:

1. deprecate `textLabel` in favor of `heading`
2. add `heading` prop, which will fallback to `textLabel` if unspecified
3. add `label`, which will fallback to `value` if not specified
4. create issue to drop `textLabel` prop and make `heading` required and `value` optional by 4.0

2 and 3 naming would align with [#8737](https://github.com/Esri/calcite-design-system/issues/8737).

aPreciado88:
@jcfranco @geospatialem I created this [issue](https://github.com/Esri/calcite-design-system/issues/9989) to follow up on:

`create issue to drop textLabel prop and make heading required and value optional by 4.0`

DitwanP:
🍡 Verified locally

PR #9987, opened 2024-08-07
feat(combobox-item): add `heading` property (deprecates `textLabel`) and add `label` property

aPreciado88:
**Related Issue:** [#8737](https://github.com/Esri/calcite-design-system/issues/8737)

## Summary
Update `combobox` component to:

- Deprecate `textLabel ` in favor of `heading`.
- Add `heading` prop, which will fallback to `textLabel` if unspecified.
- Add `label` prop, which will fallback to `value` if not specified.


packages/calcite-components/src/components/combobox-item/combobox-item.tsx
```diff
@@ -136,11 +136,21 @@ export class ComboboxItem implements ConditionalSlotComponent, InteractiveCompon
   @Prop() shortHeading: string;
 
   /** The component's text. */
+  @Prop({ reflect: true }) heading: string;
```
jcfranco, line 139: Let's not reflect `heading`, as it is content. I don't recall why we were originally reflecting `textLabel`. 🤔

jcfranco (approved):
Looks great! 

Before merging, could you revisit the PR title before merging to highlight the new props? This is important as the titles are automatically pulled into the changelog. It’s helpful to frame them from a developer’s perspective, focusing on the benefits or changes they’ll see (similar to [this PR](https://github.com/Esri/calcite-design-system/pull/9194)).

aPreciado88:
@jcfranco I updated the PR with the requested changes. Does the new title look okay? 🐱‍👤

packages/calcite-components/src/components/combobox-item/combobox-item.tsx
```diff
@@ -136,11 +136,21 @@ export class ComboboxItem implements ConditionalSlotComponent, InteractiveCompon
...
 
   /** The component's text. */
+  @Prop() heading: string;
+
+  /**
+   * The component's text.
+   *
+   * @deprecated Use `heading` instead.
+   */
   @Prop({ reflect: true }) textLabel!: string;
```
driskull, line 146: `textLabel` should remove the `!` so its no longer required here.

Issue #8867, opened 2024-03-01
Unable to drag-select all or some of label or validation message text


nwhittaker:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Double clicking on, or performing a drag selection (`mousedown` + `mousemove`), on an input's label or validation message text results in losing the selection and focus shifting to the input when the user ends the drag.

For switches, checkboxes, and radios, the input's `checked` property is toggled by a drag selection.

### Expected Behavior

Matching the native of behavior of when a drag operation is detected, focus is not shifted and the selection remains after the drag ends.

Double-clicking on a word in the label blurs the input after the second click and selects the word. Clicking within the validation message does not focus the input.

For switches, checkboxes and radios, the input's `checked` property is not toggled by a drag selection. To match the native behavior, double-clicking may toggle the property, but focus should not end up on the control.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/xxebVxo

### Reproduction Steps

1. Visit the repro sample
2. Attempt to make selections within the label and/or input message text of the first input
3. Attempt to make selections within the input message of the second input
4. Attempt to make selects within the labels of the switch, checkbox, and radio inputs

### Reproduction Version

2.10.1

### Relevant Info

If the input message is not considered part of the label, then ~resolving https://github.com/Esri/calcite-design-system/issues/6702 could fix selections within the input message shifting focus~. _- Never mind, it looks like the issue is still present with just an input message and no label. I'll update the repro._

One suggestion is to, in the label's click handler, bail if a [range selection][1] exists.

### Priority impact

p3 - want for upcoming milestone

### Impact

Impact is towards usability. In some cases error messages may contain error codes that the user may want to copy/paste.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Field Apps

[1]: https://developer.mozilla.org/en-US/docs/Web/API/Selection/type#range

nwhittaker:
Not sure what happened to the codepen as it appeared to be missing the labels and messages. I updated it to the latest Calcite and to reflect the repro steps.

driskull:
> Clicking within the validation message does not focus the input.

[...]

PR #9990, opened 2024-08-07
fix(label): prevent focusing and toggling elements slotted within a label when a text selection occurs

driskull:
**Related Issue:** #8867

## Summary

- prevents internal label selection event from being emitted if a range of text selection is active
- add e2e test

## Notes

https://developer.mozilla.org/en-US/docs/Web/API/Window/getSelection
https://developer.mozilla.org/en-US/docs/Web/API/Selection/addRange


packages/calcite-components/src/components/label/label.e2e.ts
```diff
@@ -145,6 +146,36 @@ describe("calcite-label", () => {
...
+    await page.setContent(`
+    <calcite-label id="do-not-duplicate-me" layout="inline-space-between">
+    Label text
+    <calcite-input></calcite-input>
+    </calcite-label>
+    `);
+
+    const element = await page.find("calcite-label");
+
+    const eventSpy = await element.spyOnEvent(labelClickEvent);
```
jcfranco, line 160: Can you assert on the selection and unfocused input instead of the internal event (implementation detail)?

jcfranco (changes requested):
Looking great! Just had a note to update assertions to avoid testing the implementation detail.

packages/calcite-components/src/components/label/label.e2e.ts
```diff
@@ -145,6 +146,44 @@ describe("calcite-label", () => {
     expect(element).toEqualAttribute("layout", "inline-space-between");
   });
 
+  it("should focus on the slotted form element when a label is clicked", async () => {
+    const page = await newE2EPage();
+    await page.setContent(`
+    <calcite-label id="do-not-duplicate-me" layout="inline-space-between">
```
jcfranco, line 152: Sorry for not spotting this earlier, but let's remove anything unrelated to the test. This `id` comes from a test specific to that prop/attr.

packages/calcite-components/src/components/label/label.e2e.ts
```diff
@@ -145,6 +146,44 @@ describe("calcite-label", () => {
...
+
+    await element.click();
+
+    expect(await isElementFocused(page, "calcite-input")).toBe(true);
+  });
+
+  it("should not focus on the slotted form element when a label's text is selected", async () => {
+    const page = await newE2EPage();
+    await page.setContent(`
+    <calcite-label id="do-not-duplicate-me" layout="inline-space-between">
```
jcfranco, line 168: Bonus points if you bring in the `html` formatting helper. ✨🏆✨

packages/calcite-components/src/components/label/label.e2e.ts
```diff
@@ -145,6 +146,44 @@ describe("calcite-label", () => {
     expect(element).toEqualAttribute("layout", "inline-space-between");
   });
 
+  it("should focus on the slotted form element when a label is clicked", async () => {
```
jcfranco, line 149: I like how this provides context for the selection test, but it is already covered by the `labelable` helper. 

Maybe it's OK to merge with the selection test where it does the following

1. text is selected + click > no focus shift
2. text is unselected + click > focus shifts

?

jcfranco (approved):
😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎
😎😎🆒🆒🆒😎😎🆒🆒😎😎😎🆒🆒😎😎🆒😎😎😎🆒😎
😎🆒😎😎😎😎🆒😎😎🆒😎🆒😎😎🆒😎🆒😎😎😎🆒😎
😎🆒😎😎😎😎🆒😎😎🆒😎🆒😎😎🆒😎🆒😎😎😎🆒😎
😎🆒😎😎😎😎🆒😎😎🆒😎🆒😎😎🆒😎🆒😎😎😎😎😎
😎😎🆒🆒🆒😎😎🆒🆒😎😎😎🆒🆒😎😎🆒🆒🆒😎🆒😎
😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎

packages/calcite-components/src/components/label/label.e2e.ts
driskull: I can remove this test since its testing duplicate functionality.

Issue #10006, opened 2024-08-08
Clicking on calcite-input-message should not focus input


driskull:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When clicking on the validation message, the input becomes focused

### Expected Behavior

When clicking on the validation message, no focus is shifted

### Reproduction Sample

https://codepen.io/driskull/pen/QWXMaYo?editors=100

### Reproduction Steps

1. open sample
2. click on the validation message
3. Notice the input is focused
4. other validation message components do not do this

### Reproduction Version

2.11.1

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

DitwanP:
🍡 Verified on `2.12.0-next.16`
https://codepen.io/Ditwan-Price/pen/gONogbB?editors=1001

PR #10008, opened 2024-08-08
fix(input, input-number, input-text): no longer set focus when a validation message is clicked

driskull:
**Related Issue:** #10006

## Summary

- when clicking on a validation message, the component is no longer being focused
- adds test


benelan (approved):
:+1:

Issue #9340, opened 2024-05-14
[Alert] Multiple `open` Alerts present in DOM at initialization should form a queue


macandcheese:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

If multiple Alert with the `open` property are present in the DOM as the components initialize, they will not form a queue, and appear stacked on top of one another.

Subsequently added Alert will "fix the queue" and properly stack Alert, but this doesn't happen at load.

Can be coupled with the effort of https://github.com/Esri/calcite-design-system/issues/9499.

### Acceptance Criteria

If there are multiple Alert with the `open` property present in the DOM as the components initialize, they should form a queue. As there currently isn't a way to manage order, we might just use "last one wins" like we do in other components that have multiple sibling components that can load simultaneously.

### Relevant Info

Further improvements may occur in #8705, but it would be nice to solve this with our current set up... might require the addition of an Alert Group to manage?

### Which Component

Alert

### Example Use Case

Codepen: 

https://codepen.io/mac_and_cheese/pen/xxNGjVe?editors=1000

Note that on load the two Alert are stacked on top of one another, each unaware of the other's existence. If you "Add an Alert" - it will correctly create a queue with 3 Alerts.

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

DitwanP:
🍡 Verified on `2.12.0-next.26`
https://codepen.io/Ditwan-Price/pen/poXLgmW?editors=1010

PR #10032, opened 2024-08-09
fix(alert): properly form a queue of alerts

driskull:
**Related Issue:** #9340

## Summary

- add `AlertManager` class for handling registration of alerts.
- refactor `Alert` to no longer emit internal events. Instead, alert communication is handled via the `AlertManager`
  - added two internal props to alert to whether an alert is active and how many alerts are opened.
  - instead of multiple alerts being `queued`, only one alert is `active`. All other `open` alerts are assumed to be queued.
- add spec test


packages/calcite-components/src/components/alert/alert.tsx
```diff
@@ -72,18 +74,43 @@ export class Alert implements OpenCloseComponent, LoadableComponent, T9nComponen
   //
   //---------------------------------------------------------------------------
 
+  /**
+   * This internal property, managed by the AlertManager, is used
+   * to inform the component if it is the active open Alert.
+   *
+   * @internal
+   */
+  @Prop() active = false;
```
driskull, line 83: Alternatively, this could be "selected" instead of "active"

packages/calcite-components/src/components/alert/AlertManager.ts
```diff
@@ -0,0 +1,67 @@
+export const alertQueueTimeoutMs = 300;
+
+export default class AlertManager {
+  // --------------------------------------------------------------------------
+  //
+  //  Private Properties
+  //
+  // --------------------------------------------------------------------------
+
+  private registeredElements: HTMLCalciteAlertElement[] = [];
```
driskull, line 10: In the future, this could be a map 

```ts
 private registeredElementsMap: Map<string, HTMLCalciteAlertElement[]> = new Map([
    ["default", []],
  ]);
```

This would allow us to have alert `groups` by adding a `group` property to `Alert`. Then only certain alerts would be grouped with other ones.

packages/calcite-components/src/components/alert/alert.e2e.ts
```diff
@@ -5,6 +5,8 @@ import { getElementXY, skipAnimations } from "../../tests/utils";
 import { openClose } from "../../tests/commonTests";
 import { CSS, DURATIONS } from "./resources";
 
+const animationDurationInMs = 400;
```
jcfranco, line 8: If this is for the auto-close duration, can we rename it to convey this? I believe the open/close animations are all properly wired up, so the relevant events can be awaited on.

jcfranco (approved):
This is awesome, @driskull! 🚀

packages/calcite-components/src/components/alert/alert.scss
```diff
@@ -223,7 +223,7 @@ $border-style: 1px solid var(--calcite-color-border-3);
 }
 
 :host([open]) {
-  .container:not(.container--queued) {
+  .container--active {
```
macandcheese, line 226: 🧹

macandcheese (approved):
Great refactor 🎈 !

packages/calcite-components/src/components/alert/alert.e2e.ts
driskull: fixed!

PR #10040, opened 2024-08-12
chore: stop requesting review from ben/franco for all PRs

benelan:
## Summary

Code owners is requesting review from Franco and I to all pull requests. It might be because we are the first item on the list and there is no catch-all. Hopefully adding a catch-all that assigns no one will fix the issue


jcfranco (approved):

jcfranco:
> So maybe CODEOWNERS uses regex, in which case .* means "everything".

So weird. Based on the [doc](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners), it seems `.*` would only target `.`-leading files/directories.

Issue #9977, opened 2024-08-06
Color picker text gets highlighted


nel11211:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Color picker text gets highlighted when dragging cursor over during color selection.

### Expected Behavior

When the user is dragging the color picker around the swatch, it should not also highlight the text when it gets waved over. `2.10.1` has the expected behavior.

### Reproduction Sample

https://codepen.io/nel11211/pen/ExBXeLy?editors=1000

![calcite-color-picker-text-selection-bug](https://github.com/user-attachments/assets/5db5583e-5900-4c3a-b9ca-f09d5bf68346)

### Reproduction Steps

1. In a `calcite-color-picker` component, click and drag in the swatch to select a color. 
2. Drag cursor outside and below the color picker.
3. **Text gets highlighted.**

### Reproduction Version

2.11.1

### Relevant Info

Regression (bug not seen on `2.10.1`):

![calcite-color-picker-text-selection-before](https://github.com/user-attachments/assets/fe0739ae-8add-4855-b65a-e55f3e309c80)

### Regression?

2.10.1

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Charts

geospatialem:
Possibly related to https://github.com/Esri/calcite-design-system/commit/b2be62566821e36b985d133b1c867b3cfb51fa84

DitwanP:
🍡 Verified on `2.12.0-next.13`
https://codepen.io/Ditwan-Price/pen/VwJrXRN?editors=1000

PR #10043, opened 2024-08-12
fix(color-picker): prevent text selection when using color sliders/field

jcfranco:
**Related Issue:** #9977

## Summary

Restores prevention of text-selection when using the color field/sliders.


driskull (approved):
👍

aPreciado88 (approved):
🚀

Issue #9713, opened 2024-06-28
Remove deprecations for `action-group`, `color-picker`, `shell-panel`, `table`, `checkbox` and `radio-button`


geospatialem:
### Description

- [x] Remove Action Group's `layout` property #10792 
- [x] Remove Color Picker's `hideChannels`, `hideHex` and `hideSaved` properties #10790 
- [x] Remove Shell Panel `detachedHeightScale` + `detached` properties and `--calcite-shell-panel-detached-max-height` CSS property #10794
- [x] Removes Table's `zebra` property #10795
- [x] Remove Checkbox's `guid` property #10797
- [x] Remove Radio Button's `guid` property #10799

### Proposed Advantages

Removes deprecations from `1.4.0` (May 2023).

### Which Component

- `action-group`
- `color-picker`
- `shell-panel`
- `table`
- `checkbox`
- `radio-button`

### Relevant Info

Stems from the following closed issues, where deprecations took place.
- #6390 
- #749
- #6919 
- #7906

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

DitwanP:
Verified locally on `dev`

PR #10058, opened 2024-08-13
feat(action, action-bar, action-group, action-menu, action-pad): add component tokens

alisonailea:
**Related Issue:** #7180

## Summary

Screenshot 2024-08-20 at 7 25 50 PM

### Action

--calcite-action-indicator-color: Specifies the component's indicator color.
--calcite-action-background-color: Specifies the component's  background color.
--calcite-action-background-color-hover: Specifies the component's background color when hovered or focused.
--calcite-action-background-color-pressed: Specifies the component's background color when active.
--calcite-action-text-color: Specifies the text color of the component.
--calcite-action-text-color-pressed: Specifies the component's text color when hovered.

### Action Bar

--calcite-action-bar-expanded-max-width: Use When `layout` is `"vertical"`, specifies the expanded max width of the component.

### Action Group

--calcite-action-group-border-color: Specifies the component's border color when used in a calcite-action-bar or calcite-action-menu.
--calcite-action-group-columns: Specifies the component's grid-template-columns when the `layout` property is `"grid"`.
--calcite-action-group-gap: Specifies the component's gap (gutters) between rows and columns and padding around the component, when the `layout` property is `"grid"`.
--calcite-action-group-padding: [Deprecated] Use --calcite-action-group-gap. Sets the padding when the `layout` property is `"grid"`.

### Action Pad

--calcite-action-pad-corner-radius: Specifies the component's border radius.
--calcite-action-pad-expanded-max-width: Specifies the component's maximum width when `expanded` and the `layout` is `"vertical"`.


packages/calcite-components/src/components/action-menu/action-menu.scss
```diff
@@ -1,17 +1,12 @@
...
   box-border
   flex
   flex-col;
 }
 
 ::slotted(calcite-action-group) {
-  border-block-end: 1px solid var(--calcite-color-border-3);
-}
-
-::slotted(calcite-action-group:last-child) {
```
driskull, line 13: question: Wouldn't removing this this make the last group have a border-bottom? Can you verify this isn't breaking anything?

packages/calcite-components/src/components/action/action.scss
```diff
@@ -227,3 +223,19 @@
 }
 
 @include base-component();
+@include disabled {
+  //   :host([disabled]) .button,
```
driskull, line 227: question: Can we remove these comments?

packages/calcite-components/src/components/action/action.scss
```diff
@@ -154,64 +207,7 @@
...
-:host([disabled]) .button:focus {
-  @apply opacity-disabled bg-foreground-1 cursor-default;
-}
-
-:host([disabled][active]) .button,
-:host([disabled][active]) .button:hover,
-:host([disabled][active]) .button:focus {
-  @apply opacity-disabled bg-foreground-3;
-}
+//--
```
driskull, line 210: question: Can we remove these comments?

packages/calcite-components/src/components/action/action.scss
driskull: question: Are we sure this CSS isn't necessary?

packages/calcite-components/src/demos/_assets/demo-theme.ts
```diff
@@ -15,10 +15,11 @@ export function getTokenValue(token: string): string {
     color$: "rgb(0, 191, 255)",
     hover$: "rgb(255, 105, 180)",
     pressed$: "rgb(44, 44, 44)",
+    selected$: "rgb(156, 89, 209)",
     shadow$:
       "rgb(255, 255, 255) 0px 0px 0px 4px, rgb(255, 105, 180) 0px 0px 0px 5px inset, rgb(0, 191, 255) 0px 0px 0px 9px",
-    "z-index$": "42",
-    "(size|space)$": "42px",
+    "(z-index)$": "42",
+    "(columns|gap|height|offset|radius|size|space|width)": "42px",
```
driskull, line 22: fixme: Could this demo-theme code share the same code from `cssTokenValues`?

packages/calcite-components/src/tests/utils/cssTokenValues.ts
```diff
@@ -15,10 +15,11 @@ export function getTokenValue(token: string): string {
     color$: "rgb(0, 191, 255)",
     hover$: "rgb(255, 105, 180)",
     pressed$: "rgb(44, 44, 44)",
+    selected$: "rgb(156, 89, 209)",
     shadow$:
       "rgb(255, 255, 255) 0px 0px 0px 4px, rgb(255, 105, 180) 0px 0px 0px 5px inset, rgb(0, 191, 255) 0px 0px 0px 9px",
-    "z-index$": "42",
-    "(size|space)$": "42px",
+    "(z-index)$": "42",
+    "(columns|gap|height|offset|radius|size|space|width)": "42px",
```
driskull, line 22: nitpick: for `offset` and `size`, we need to also support `size-x`, `size-y` and `offset-x`, `offset-y` so maybe we could update that here?

driskull (changes requested):
Code looks good @alisonailea, just had a few questions and nitpicks

packages/calcite-components/src/components/action-group/action-group.scss
```diff
@@ -3,9 +3,12 @@
...
  * These properties can be overridden using the component's tag as selector.
  *
- * @prop --calcite-action-group-columns: Sets number of grid-template-columns when the `layout` property is `"grid"`.
- * @prop --calcite-action-group-gap: Sets the gap (gutters) between rows and columns when the `layout` property is `"grid"`.
- * @prop --calcite-action-group-padding: Sets the padding when the `layout` property is `"grid"`.
+ * @prop --calcite-action-background-color: Specifies the component's background color.
+ * @prop --calcite-action-group-border-color: Specifies the component's border color when used in a calcite-action-bar or calcite-action-menu.
+ * @prop --calcite-action-group-columns: Specifies the component's grid-template-columns when the `layout` property is `"grid"`.
+ * @prop --calcite-action-group-gap: Specifies the component's gap (gutters) between rows and columns and padding around the component, when the `layout` property is `"grid"`.
+ * @prop --calcite-action-group-padding: [Deprecated] Use --calcite-action-group-gap. Sets the padding when the `layout` property is `"grid"`.
```
macandcheese, line 10: I think we might be safe to deprecate and not replace with "gap" as this `layout` property is deprecated and will be removed shortly: https://github.com/Esri/calcite-design-system/issues/9713

packages/calcite-components/src/components/action-group/action-group.scss
```diff
@@ -54,15 +61,18 @@
...
-  place-content-stretch;
-  gap: var(--calcite-action-group-gap);
-  padding: var(--calcite-action-group-gap);
-  grid-template-columns: repeat(var(--calcite-action-group-columns), auto);
+  .container {
+    @apply grid
+    place-content-stretch;
+    background-color: var(--calcite-action-background-color, var(--calcite-color-foreground-1));
+    gap: var(--calcite-action-group-gap, theme("gap.px"));
+    padding: var(--calcite-action-group-gap, theme("gap.px"));
```
macandcheese, line 70: Not sure we'd want to allow these to be changed in the current set up where `layout` isn't set on the Action Group.

packages/calcite-components/src/components/action-menu/action-menu.scss
alisonailea: I handed back control of calcite-action-group to itself. The style you're looking for is on line 27 of action-group.scss

packages/calcite-components/src/demos/_assets/demo-theme.ts
alisonailea: where should it live? I was getting errors when I put a single file in /utils that it wasn't available to both storybook, puppeteer, and the html pages.

packages/calcite-components/src/tests/utils/cssTokenValues.ts
alisonailea: those are covered by this use-case

packages/calcite-components/src/components/action-bar/action-bar.scss
```diff
@@ -3,59 +3,50 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
- * @prop --calcite-action-bar-expanded-max-width: When `layout` is `"vertical"`, specifies the expanded max width of the component.
+ * @prop --calcite-action-bar-expanded-max-width: Use When `layout` is `"vertical"`, specifies the expanded max width of the component.
```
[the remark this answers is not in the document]
alisonailea: Specifies is more common across components.

packages/calcite-components/src/components/action/action.scss
alisonailea: just double checked. Removing this code makes no changes to the UI. I've added a test to storybook for the component in an active-disabled state.

packages/calcite-components/src/components/action-group/action-group.scss
alisonailea: These weren't added by me. https://github.com/Esri/calcite-design-system/pull/7763. However, when investigating the code I noticed `--calcite-action-group-padding` wasn't actually used anywhere so it's deprecated.

packages/calcite-components/src/components/action-group/action-group.scss
alisonailea: I moved some code around for organizational purposes but this was the previous default. Gap won't do anything unless flex is set which is only added via a `layout`

packages/calcite-components/src/components/action/action.scss
driskull: Thanks. sounds good!

packages/calcite-components/src/components/action-menu/action-menu.scss
driskull: Thanks

packages/calcite-components/src/demos/_assets/demo-theme.ts
driskull: @jcfranco @benelan do you know where we can put files used between all these areas?

driskull (approved):
👍 ⭐

packages/calcite-components/src/components/action-group/action-group.scss
macandcheese: @driskull is it safe to deprecate gap with the removal of layout prop? Or does that need to be defined elsewhere?

packages/calcite-components/src/components/action-group/action-group.scss
macandcheese: Cool, maybe we can have a note to remove it when the "layout" is removed (it is currently deprecated)

packages/calcite-components/src/components/action-group/action-group.scss
driskull: I think we still need it. `layout` will just become an internal property to set the layout between grid, horizontal or vertical. So if we keep grid we also need the gap css var if we want to allow users to be able to customize the gap spacing.

macandcheese: Could that be on Action Bar / Pad component? Seems weird to set that on an individual group if the groups themselves are spaced by a parent. Unless I'm misinterpreting the use case. We don't want gaps between Actions to be added.

driskull: Yeah that's a good point. I think the parent can just be updated to handle the spacing.

packages/calcite-components/src/components/action-group/action-group.scss
alisonailea: Added a [Deprecated] to the `--calcite-action-group-gap` docs. Should I add a note to remove the token anywhere else?

driskull: That should be good 👍

packages/calcite-components/src/components/action-group/action-group.scss
alisonailea: added a new commit which passes padding control to the bar, menu, pad components. Is that what y'all are thinking?

packages/calcite-components/src/components/action-pad/action-pad.scss
```diff
@@ -5,6 +5,7 @@
  *
  * @prop --calcite-action-pad-corner-radius: Specifies the component's border radius.
  * @prop --calcite-action-pad-expanded-max-width: Specifies the component's maximum width when `expanded` and the `layout` is `"vertical"`.
+ * @prop --calcite-action-pad-items-space" Specifies the space between slotted components in the component.
```
driskull, line 8: typo:

`* @prop --calcite-action-pad-items-space: Specifies the space between slotted components in the component.`

driskull (approved):
👍

packages/calcite-components/src/components/action-group/action-group.scss
driskull: This looks good to me. What do you think @macandcheese?

macandcheese: Yes! This set up makes more sense to me as a user. 

Wondering if `gap` and `group` is better here than `space` and `items` - to me this is more applicable verbiage for what the property controls: `--calcite-action-bar-group-gap` and `--calcite-action-pad-group-gap` 

I also am curious about `--calcite-action-menu-items-space` - not sure this var (by any name) is necessary - do we want folks to have a gap between groups in an Action Menu? I'd think no.

packages/calcite-components/src/demos/_assets/demo-theme.ts
jcfranco: We could look into providing utils in the dev/demo build, but for now, we can proceed with the slight duplication.

jcfranco: Created follow-up issue: https://github.com/Esri/calcite-design-system/issues/10107.

packages/calcite-components/src/tests/commonTests/themed.ts
```diff
@@ -207,12 +207,12 @@ export type TestTarget = {
...
    * The expected value of the targetProp.
    */
-  expectedValue: string;
+  expectedValue?: string;
 
   /**
    * The associated component token.
    */
-  token: CalciteCSSCustomProp;
+  token?: CalciteCSSCustomProp;
```
jcfranco, line 215: Aren't both `expectedValue` and `token` required for all assertions?

packages/calcite-components/src/components/action-group/action-group.scss
alisonailea: technically it's not limited to action-groups though, so I hesitate to use `-group-gap`. It's the gap between any slotted items inside the action-bar, action-pad, or action-menu so maybe `--calcite-[component]-items-gap` works?

packages/calcite-components/src/custom-theme.stories.ts
```diff
@@ -77,33 +89,52 @@ export const themingInteractive = (args: Record<string, string>): string => {
...
+    ...actionGroupTokens,
+    ...cardTokens,
+  },
+};
+
+export const themingInteractive = (args: Record<string, string>): string => {
+  return kitchenSink(args);
+};
+
+export const theming_TestOnly = (): string => {
```
jcfranco, line 128: Late to the party, but new screenshot tests don't need the `_TestOnly` suffix anymore.

packages/calcite-components/src/components/action-group/action-group.stories.ts
```diff
@@ -94,3 +94,26 @@ export const TurkishLocale_TestOnly = (): string =>
       <calcite-dropdown slot="menu-actions"> </calcite-dropdown>
     </calcite-action-group>
   </div>`;
+html`<div class="container">
```
jcfranco, line 97: Is this dangling HTML snippet meant for a separate story? It is not currently being used.

packages/calcite-components/src/custom-theme.stories.ts
alisonailea: I'd like to somehow distinguish between the story that is designed to allow the user to change the variables vs variables that are set to a new color theme for testing purposes already.

packages/calcite-components/src/components/action-group/action-group.stories.ts
alisonailea: Not used. Removed here https://github.com/Esri/calcite-design-system/pull/10167

packages/calcite-components/src/custom-theme.stories.ts
jcfranco: That's fine, but we're no longer using that specific suffix (see https://github.com/Esri/calcite-design-system/issues/9085 for context). Calling this out shouldn't be necessary since most stories are primarily used for screenshot tests.

I think the name for the interactive story already conveys the separation well, and we could consider adopting this approach for other interactive stories too.

packages/calcite-components/src/components/action/action.scss
```diff
@@ -47,45 +44,101 @@
...
+  transition-property: inline-size;
+}
+
+.text-container--visible {
+  @apply w-auto flex-auto opacity-100;
+}
+
+:host([active]) {
+  .button {
+    &,
```
driskull, line 97: @alisonailea we may have a bug here. Issue incoming.

packages/calcite-components/src/components/action-bar/action-bar.scss
```diff
@@ -3,59 +3,53 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
- * @prop --calcite-action-bar-expanded-max-width: When `layout` is `"vertical"`, specifies the expanded max width of the component.
+ * @prop --calcite-action-bar-expanded-max-width: Specifies the maximum width of the component when it's `layout` is `"vertical"`.
+ * @prop --calcite-action-bar-items-space: Specifies the space between slotted components in the component.
```
macandcheese, line 7: Do we need this prop? From a design standpoint, not sure what the reasoning for having this capability is? cc @ashetland @SkyeSeitz

SkyeSeitz: Agreed not needed imo

packages/calcite-components/src/components/action/action.scss
geospatialem: https://github.com/Esri/calcite-design-system/issues/10982 cc @alisonailea

Issue #10080, opened 2024-08-15
[Dialog] Add `escapeDisabled` prop


ellenupp:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Add prop to disable closing dialog on escape key. Modal, which we need to replace, had this. 

We have a lot of complicated workflows (element configuration) in modals and don't want people to accidentally close out of it. Especially when the escape key can be used for other purposes in the modal like closing dropdowns.

### Acceptance Criteria

When `escapeDisabled` is present, disables the closing of the component on escape key.

### Which Component

calcite-dialog

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Dashboards

DitwanP:
Sending back to resolve an issue where the `panel` is still closing the `dialog` even with the `escapeDisabled` prop present.

DitwanP:
🍡 Verified on `2.12.0-next.24` 
https://codepen.io/Ditwan-Price/pen/eYwybyr

PR #10081, opened 2024-08-15
feat(dialog): add escapeDisabled property

driskull:
**Related Issue:** #10080

## Summary

- add `escapeDisabled` property
- update storybook
- add e2e test

### DIFF
--- a/packages/calcite-components/src/components/dialog/dialog.tsx
+++ b/packages/calcite-components/src/components/dialog/dialog.tsx
@@ -103,7 +103,13 @@ export class Dialog
    */
   @Prop({ mutable: true }) embedded = false;
 
-  /** When `true`, disables the default close on escape behavior. */
+  /**
+   * When `true`, disables the default close on escape behavior.
+   *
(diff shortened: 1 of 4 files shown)


packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -103,6 +103,15 @@ export class Dialog
    */
   @Prop({ mutable: true }) embedded = false;
 
+  /**
+   * When `true`, disables the default close on escape behavior.
+   *
+   * By default, an open dialog can be dismissed by pressing the Esc key. Depending on what the dialog represents, it may not be desired for this behavior.
```
benelan, line 109: Nitpick: "... represents, this behavior may not be desired."

benelan (approved):
👍

packages/calcite-components/src/components/dialog/dialog.tsx
macandcheese: I think the second sentence here could be removed.

Issue #10042, opened 2024-08-12
[Panel] No line height defined for `heading` or `description`


macandcheese:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The `heading` and `description` text has no defined line height and can be overridden unexpectedly - we should explicitly set line heights to prevent this.

### Expected Behavior

I'd expect styles within the component to be set that prevent outside css from affecting the styles. We should confirm appropriate styles are set on any other slots or rendered elements as well.

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/zYVPvKE?editors=1000

### Reproduction Steps

1. Open codepen
2. 

### Reproduction Version

2.11

### Regression?

2.10 - https://github.com/Esri/calcite-design-system/pull/9730/files#diff-0b45120ffa61191421f1699207bb91da11d979e0a76a1b28d50353a3b17394c2

### Priority impact

impact - p1 - need for current milestone

### Impact

This will affect any apps that have line height defined in external css, most notably jsapi css which causes unwanted style changes when used with 2.11

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

DitwanP:
🍡 Verified on `2.12.0-next.26`

PR #10085, opened 2024-08-16
fix(panel): define heading and description line height

Elijbet:
**Related Issue:** #10042

## Summary
Fix regression where panel `heading` and `description` retain their line height.


ashetland:
Accepted all visual diffs as expected.

Issue #10022, opened 2024-08-08
[Panel] focus outline on is obscured behind the header


DitwanP:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Currently the outline is hidden behind the header:
![image](https://github.com/user-attachments/assets/701667fe-68d5-4ab7-a099-396f317d97d6)

### Expected Behavior

The outline should be visible on top of the header.

### Reproduction Sample

https://codepen.io/Ditwan-Price/pen/gONxjaO?editors=1100

### Reproduction Steps

1. Open codepen
2. Tab to panel and notice the focus outline behavior

### Reproduction Version

2.11.1

### Relevant Info

(https://github.com/Esri/calcite-design-system/pull/9804#issuecomment-2270060429)

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

jcfranco:
@DitwanP I've only noticed this issue in Chrome. Can you confirm?

geospatialem:
Verified in `2.12.1` 🎉

PR #10092, opened 2024-08-16
fix(panel): correct focus appearance

driskull:
**Related Issue:** #10022

### DIFF
--- a/packages/calcite-components/src/components/panel/panel.scss
+++ b/packages/calcite-components/src/components/panel/panel.scss
@@ -79,6 +79,14 @@
     inline-size var(--calcite-animation-timing);
 }
 
+.container:focus {
+  @apply focus-inset z-default;
+
+  .header {
+    @apply -z-default;
+  }
+}
+
 .container[hidden] {
   @apply hidden;
 }
@@ -163,6 +171,10 @@
   bg-background
   overflow-auto
   h-full;
+}
+
+.content-wrapper:focus {
+  @apply focus-inset;
 }
 
 .content-top,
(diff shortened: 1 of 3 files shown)


macandcheese:
We might need to try and ensure it appears above slotted elements like Block, etc., - currently there is some overlap both in idle / hover state (from the "block configuration" local demo). The 1px "inset" might be coming from a Panel border? 
Screenshot 2024-08-16 at 5 28 10 PM

driskull:
@macandcheese can you test again?

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -79,6 +79,14 @@
     inline-size var(--calcite-animation-timing);
 }
 
+.container:focus {
+  @apply focus-inset z-default;
```
jcfranco, line 83: Is there an alternative way to accomplish this? We want to avoid setting z-indices as much as possible. Headers have a designated z-index to use, which might be problematic here.

I think this might only be affecting Chrome too, so we could hold off a bit while we look into alternatives.

driskull: I'm fine holding off to research this a bit more.

Even if we don't apply the z-index here, it might be nice to apply the consistent `focus-inset` styles.

Notes:

- In this scenario, we need to use an inset focus so that the focus is inside of the panel
- When the focus is inside of the panel, it can be overlapped by headers and other elements with a z-index
- A good solution would make sure the focus outline is above any slotted in content

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -80,11 +80,7 @@
 }
 
 .container:focus {
-  @apply focus-inset z-default;
-
-  > * {
-    @apply -z-default;
-  }
+  @apply focus-inset;
```
jcfranco, line 83: I think `focus-base` needs to be applied to `.container` too.

jcfranco (approved):

PR #10097, opened 2024-08-16
fix(dialog): fix escapeDisabled when the escape key is pressed on the panel

driskull:
**Related Issue:** #10080

## Summary

- panel doesn't have an escapeDisabled property
- uses preventDefault to stop escape key from closing panel which closes dialog
- adds test


jcfranco (approved):
&nbsp;&nbsp;&nbsp;👇
🚫⌨️🚫

Issue #10095, opened 2024-08-16
[Dialog] Some slotted content appears below or overlaps the header/footer


ellenupp:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When slotting absolute positioned content into the default slot of the `dialog`, it can appear below the header and overlap with the footer and its contents. We use shell in modal, and I don't think we can work around this by adjusting its positioning. 

Here is a shell in the `dialog`'s default slot. Notice the first action is hidden below the header, the footer bar is missing and the content overlaps the footer buttons. 
![image](https://github.com/user-attachments/assets/d8700072-9d8c-4af7-aaf6-3c1f33300795)

vs the same content in `modal`'s `content` slot
![image](https://github.com/user-attachments/assets/e9ce73ed-d09b-4b16-8339-061996284efd)

### Expected Behavior

Same as `modal`, all content should appear between the header and footer. 

### Reproduction Sample

https://codepen.io/ellupp/pen/LYKOmPa?editors=1101

### Reproduction Steps

1. click button to open dialog and observe

### Reproduction Version

2.11.1

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Dashboards

driskull:
The issue is actually with the shell component. The shell component is positioned absolutely, so it doesn't slot well into dialog.

driskull:
@DitwanP I un-spiked this one.

DitwanP:
🍡 Verified on `2.12.0-next.29`
https://codepen.io/Ditwan-Price/pen/BagxoGq?editors=1100

ellenupp:
Thank you!

geospatialem:
Summarizing our internal conversations, but due to some regressions with the above, we will be reverting the change above, and re-evaluating next steps in the `3.0`, or November release. 

cc @driskull @jcfranco @ellenupp

jcfranco:
Per our conversation, we'll look at rolling back https://github.com/Esri/calcite-design-system/pull/10496, and addressing this issue, once https://github.com/Esri/calcite-design-system/pull/10241 lands in the November release.

[...]

PR #10108, opened 2024-08-19
fix(dialog): allow shell to be slotted inside

driskull:
**Related Issue:** #10095

## Summary

- supports shell being slotted inside dialog content
- add story screenshot test

### DIFF
--- a/packages/calcite-components/src/components/dialog/dialog.scss
+++ b/packages/calcite-components/src/components/dialog/dialog.scss
@@ -120,6 +120,7 @@ calcite-panel {
 }
 
 .content {
+  @apply relative flex-auto;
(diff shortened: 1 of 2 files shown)


macandcheese (approved):
🏅 Looks good - maybe in the future we can look at Shell defaulting to position:absolute because it can lead to unexpected displays in other cases as well.

driskull:
Yeah I was thinking the same thing. Anywhere we want shell slotted would need this kind of fix.

alisonailea (approved):

driskull:
Found an issue where making the content container div display as `flex: 1 1 auto` would shift the FAB down to the bottom. 

Not sure of a way to work around this so I'll close the issue.

This may be better fixed by not making the shell positioned absoutely.

Issue #7748, opened 2023-09-15
Unstable test: alert accessible


driskull:
### Test type

unstable test

### Which Component(s)

alert

### Unstable Tests

```
 describe.skip("accessible", () => {
    accessible(html` <calcite-alert open label="test"> ${alertContent} </calcite-alert> `);
  });

  describe.skip("accessible with auto-close", () => {
    accessible(html`
      <calcite-alert open auto-close auto-close-duration="slow" label="test"> ${alertContent} </calcite-alert>
    `);
  });
```

### PR skipped, if applicable

#7749

driskull:
Maybe this test will pass now that the AlertManager exists?

DitwanP:
🍡 Verified locally 
![image](https://github.com/user-attachments/assets/fc36b074-6f80-49fa-89ee-1c8bcf91a78d)

PR #10110, opened 2024-08-19
test(alert): unskip a11y test

driskull:
**Related Issue:** #7748

## Summary

- unskip test


driskull:
Ran this a couple times and seems to work.

macandcheese (approved):
Does it work without the internal properties in the DOM? I suppose the "non queue" Alert should pass this as well?

driskull:
Updated to not use internal props.

PR #10111, opened 2024-08-19
chore(custom-theme): refactor storybook theme for component tokens

alisonailea:
**Related Issue:** #7180 

## Summary

Breaks the custom-theme storybook page into component files for easier maintenance. This will allow us to move all component token chromatic tests to a single file.


packages/calcite-components/src/custom-theme/global.css
driskull, line 1: fixme: This file should be removed.

packages/calcite-components/src/custom-theme/accordion.tsx
driskull, line 1: fixme: Do these new files need to be `.tsx` files? They don't have any JSX in them right? I would expect `.ts` files.

packages/calcite-components/src/custom-theme/card.tsx
```diff
@@ -0,0 +1,24 @@
+import { placeholderImage } from "../../.storybook/placeholderImage";
+import { html } from "../../support/formatting";
+
+export const tokens = {
```
driskull, line 4: nitpick: maybe this should just be named `export const tokens` so it doesn't need to be renamed during the import.

If we are planning to do this in other components it would be good to set a standard here.

driskull (changes requested):
Nice! 👍 

Just a few minor tweaks

driskull (approved):
👍 ⭐

packages/calcite-components/src/custom-theme/button.ts
```diff
@@ -0,0 +1,15 @@
+import { html } from "../../support/formatting";
+
+export const button = (args: { kind?: string; appearance?: string }): string =>
+  html`<calcite-button
+    ${Object.entries(args)
```
jcfranco, line 5: I find this a bit tricky to read (at least in a single pass). Could you refactor this to improve readability? A function might be able to be shared between the attribute and content setting.

jcfranco (approved):
Awesome! 🚀 

Once comments are addressed, this should be good to merge!

packages/calcite-components/src/custom-theme/button.ts
alisonailea: Let me know if this update makes more sense

packages/calcite-components/src/custom-theme/button.ts
```diff
@@ -0,0 +1,17 @@
+import { html } from "../../support/formatting";
+
+const buttonProps: [args: string, content: string] = ["", ""];
+
+export const button = (props: { kind?: string; appearance?: string }): string => {
+  const [buttonArgs, content] = Object.entries(props)
+    .filter((a) => a && a[0] && a[1] && a[1] !== "")
```
jcfranco, line 7: Looking better! 😎

One note, since `a` (`[key, value]`) is guaranteed, I think this can be simplified to `.filter(([key, value]) => key && value && value !== "")`.

Thanks for making these tweaks!

packages/calcite-components/src/custom-theme/accordionItem.ts
```diff
@@ -0,0 +1,7 @@
+import { placeholderImage } from "../../.storybook/placeholderImage";
```
jcfranco, line 1: This file also needs to be kebab-cased.

packages/calcite-components/src/custom-theme/button.ts
alisonailea: this is what I get for quick copy-paste. 🤦🏻 
thanks for the catch!

packages/calcite-components/src/custom-theme/accordionItem.ts
alisonailea: this was an old file so making this change added 8 other files to the commit but it's good cleanup.

Issue #10129, opened 2024-08-21
Dialog: closable calcite panels placed inside a dialog will close the dialog


driskull:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When placing a panel inside a dialog and closing the panel, the dialog will be closed.

This is because the dialog listens for the `calcitePanelClose` event to close the dialog. If the event bubbles up, it closes the dialog. The workaround is to stop propagation of the `calcitePanelClose` event.

### Expected Behavior

I would expect closing internal panels to not close the dialog.

### Reproduction Sample

https://codepen.io/fcbasson/pen/YzoazZE

### Reproduction Steps

1. open the codepen
2. close the panel inside the dialog
3. notice the dialog closes

### Reproduction Version

2.11.1

### Relevant Info

In components that use `calcite-panel` internally and listen to its events, we should make sure the element we are looking for is emitting the events and that any other elements do not trigger any action.

### Priority impact

impact - p1 - need for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

DitwanP:
🍡 Verified locally on `dev`

https://github.com/user-attachments/assets/2807458d-2aa4-4dc8-82cf-5bf6c8e03b19

PR #10130, opened 2024-08-21
fix(dialog, flow-item): slotted closable panels should not close the component

driskull:
**Related Issue:** #10129

## Summary

- updates `dialog` and `flow-item` to only listen to panel events for the internal panel. Excluding any other bubbling events.
- uses `IDS` for unique ids of elements instead of `data-test` attributes.
- stops propagation of internal panel events on dialog.
- adds tests

### DIFF
--- a/packages/calcite-components/src/components/dialog/dialog.tsx
+++ b/packages/calcite-components/src/components/dialog/dialog.tsx
@@ -256,8 +256,8 @@ export class Dialog
                 loading={this.loading}
                 menuOpen={this.menuOpen}
                 messageOverrides={this.messageOverrides}
-                onCalcitePanelClose={this.handleCloseClick}
-                onCalcitePanelScroll={this.handleScroll}
+                onCalcitePanelClose={this.handleInternalPanelCloseClick}
+                onCalcitePanelScroll={this.handleInternalPanelScroll}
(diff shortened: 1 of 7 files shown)


packages/calcite-components/src/components/panel/panel.tsx
```diff
@@ -503,8 +503,8 @@ export class Panel
       <calcite-action
         aria-expanded={toAriaBoolean(!collapsed)}
         aria-label={collapse}
-        data-test="collapse"
```
driskull, line 506: Related to #6760.

packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -458,11 +458,21 @@ export class Dialog
     this.el.removeEventListener("calciteDialogOpen", this.openEnd);
   };
 
-  private handleScroll = (): void => {
+  private handleScroll = (event: Event): void => {
```
jcfranco, line 461: To make it clear that we're handling an internal event, can you rename the event handlers to convey this and type this event as `CustomEvent` (similar to `flow-item`)?

jcfranco (approved):
Noice! ✨🏆✨

Issue #10132, opened 2024-08-21
Fix blocked issues action regex


DitwanP:
### Priority impact

p - low

### Summary

The regex in action is not correctly matching for the "Blocked Issues:" part of the comment. Instead its just matching any mention of a number preceded by a `#`. 

This is causing the action to run and fail on issues that don't actually have any blocked issues listed like this one https://github.com/Esri/calcite-design-system/issues/7748.

Here's the failed attempt: https://github.com/Esri/calcite-design-system/actions/runs/10477994663/job/29020354524

### Desired Outcome

The action should correctly exit with the comment that there are no blocked issues listed on a closed issue.

DitwanP:
🍠 Verified the issue was resolved and automation is still working as expected (see the bot comment above). Also checked that there was no comment on an issue that I intentionally listed without "Blocked issues: " in front of it.

Also also...
<img width="277" alt="Screenshot 2024-05-21 at 11 06 30 AM" src="https://github.com/user-attachments/assets/e6277d8b-0680-4a94-8097-6acc9076fd83">

PR #10135, opened 2024-08-21
ci(actions): fix regex, and catch issue not found error on blocked issue action

DitwanP:
**Related Issue:** #10132 

## Summary
Changes the regex to better match the "Blocked issue:" so that randomly mentioned issues don't get matched and attempted to be unblocked. 

Also, catches the issue not found error in case somehow a issue that doesn't actually exist gets listed as blocked. 

[Codepen to test regex](https://codepen.io/Ditwan-Price/pen/QWXmmRL?editors=0012)


benelan (approved):
:+1:

PR #10139, opened 2024-08-21
chore(linting): enable unicorn/filename-case

jcfranco:
**Related Issue:** N/A

## Summary

Enables `unicorn/filename-case` rule for consistent file names across the monorepo.


benelan (approved):
:+1:

Elijbet:
This should make an exception for controller files. I'm getting an eslint error `Filename is not in kebab case. Rename it to use-top-layer.browser.spec.tsx  unicorn/filename-case`. Looks like we need an override for `files: ["**/controllers/*.{ts,tsx,js,jsx}"],`.

PR #10141, opened 2024-08-22
fix(panel): tweak focusable content area

jcfranco:
**Related Issue:** #10022

## Summary

Updates `panel` to have a single focusable container when it has scrolling content. 

### Noteworthy changes

* Improves test coverage for both scrolling and non-scrolling content.
* Updates `disabled` test helper to clear focus targets between tab and click tests


alisonailea (approved):
🎉

macandcheese:
The conditionally-applied tab index looks great! And should apply to Tab most likely as well :)

I think we might need to explore some other options for showing the visual focus outline here - might require some fake containers / pseudo elements - this focusable area is hard to discern and has a lot of overlap with the content slotted inside: 
Screenshot 2024-08-27 at 8 44 32 AM

driskull (approved):
👍

packages/calcite-components/src/components/panel/panel.e2e.ts
```diff
@@ -456,51 +557,83 @@ describe("calcite-panel", () => {
...
-    panel.setProperty("closable", true);
-    await page.waitForChanges();
-    await container.press("Escape");
-    await page.waitForChanges();
-    expect(await panel.getProperty("closed")).toBe(true);
-    expect(await container.isVisible()).toBe(false);
-    expect(calcitePanelClose).toHaveReceivedEventTimes(1);
-  });
-
-  it("should not close when Escape key is prevented and closable is true", async () => {
```
jcfranco, line 480: @driskull Due to the contents of `scrollingContentHtml`, this test is only covered under the scrolling use case. 

We could add coverage for this test for both scrolling and non-scrolling content by adding focusable elements to the content. Does that approach make sense? If so, would we want to keep scrolling content and non-scrolling content as the main test groups or do we need an additional group to cover scrolling with and without interactive content?

jcfranco:
For posterity, after some extensive troubleshooting, it turns out that not having a set width/height was affecting `disabled` tests (namely [this assertion](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/tests/commonTests/disabled.ts#L190) from the `prevents focusing via keyboard and mouse` test) in the CI despite tests passing locally in both headless and headful modes for both macOS and Linux.

Issue #9862, opened 2024-07-26
[Panel & Flow Item] `heading` and `description` line heights


ashetland:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

To align with issue #7623, both the `heading` and `description` should be assigned `calcite-font-line-height-relative-snug` and the margin between them removed.

### Acceptance Criteria

See above.
When only one text prop is used, the remaining text should be vertically centered in the header.

### Relevant Info

cc @driskull @SkyeSeitz 

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

aPreciado88:
This is planned to be installed during a December/January release.

geospatialem:
Verified with `flow` and `panel` in `3.0.0-next.51` 🎉

PR #10178, opened 2024-08-28
fix(panel): add correct heading and description line height and alignment

aPreciado88:
**Related Issue:** [#9862](https://github.com/Esri/calcite-design-system/issues/9862)

## Summary
Update `panel` to:

* Use `calcite-font-line-height-relative-snug` as `heading` and `description`'s line height.
* Vertically center the header's content when only `heading` or `description` are used.


packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -25,11 +25,11 @@
 
   .header-content {
     .heading {
-      @apply text-n1h;
+      font-size: var(--calcite-font-size--1);
```
alisonailea, line 28: What's the purpose in swapping these out? Because we're not using tokens for font-size right now we can leave these with tailwind

aPreciado88: `@apply text-n1h;` Was applying the incorrect line height. 

I can switch those back and overwrite the line height using `line-height: var(--calcite-font-line-height-relative-snug) !important;` in line 131. Would be better?

alisonailea (approved):

packages/calcite-components/src/components/panel/panel.scss
```diff
@@ -120,16 +120,19 @@
...
 
   .heading,
   .description {
     @apply block
-      break-words
-      p-0;
+    flex-none
+    break-words
+    p-0;
+    line-height: var(--calcite-font-line-height-relative-snug) !important;
```
alisonailea, line 131: We should avoid using !important.

alisonailea (changes requested):
I forgot text-n1h was line-height and font-size.
So keep `font-size: var(--calcite-font-size--1);`

packages/calcite-components/src/components/panel/panel.scss
aPreciado88: I removed this and switched back to using `font-size: var(--calcite-font-size--1);`

alisonailea (approved):
🎉

aPreciado88:
@ashetland @SkyeSeitz Could you please take a look at the new Chromatic builds?

aPreciado88:
This is planned to be installed during a December/January release.

aPreciado88:
Will retry this after the lumina upgrade is done.

Issue #6640, opened 2023-03-23
Make close and clear buttons more consistent across system


ashetland:
### Description

Close buttons across Calcite exhibit some inconsistencies due to differences in implementation. Ideally close buttons would be consistent in both design and implementation throughout the system.

### Acceptance Criteria
**Update close & clear button padding**
- Use `calcite-action` for close and clear buttons
- Replace any instance of `x-button` with `calcite-action` and sunset internal `x-button` component
- Update padding in consuming component 

### Relevant Info

Part of 4.0's Phase 2

Blocked by: #10759
Requires `transparent-inverse` colors tokens from #11565 

### Which Component

- `alert`
- `autocomplete`
- `chip`
- `combobox` / `input time zone`
- `inline editable`
- `input`
- `input-number`
- `input-text`
- `input-date-picker` (related issue #9440)
- `input-time-picker` (related issue #9440)
- `list item`
- `modal` (deprecated, may not need)
- `notice`
- `panel`, `flow`, `flow item`, `dialog`
- `popover`
- `tab title`
- `tip`, `tip manager` (deprecated, may not need)

### Priority impact

p4 - not time sensitive

### Esri team

Calcite (design)

ashetland:
Need to rescope this to include Close and Clear buttons. Basic design of x-button is complete and should be ready to propagate in design files.

jcfranco:
For posterity, the z-index workaround for #7272 will probably not be needed if tabs' closable button ends up being round.

ashetland:
To do

- [ ] Verify above list and add any missing
- [ ] Determine which component should get round or square buttons (or if everything just goes round)
- [ ] Make visual catalogue of changes

alisonailea:
I think code-wise we should create a mixin similar to the Floating UI functions that all components can share.

driskull:
Currently, we only have an internal functional component to handle close buttons. We should consider using action or enhancing it to meet the needs of close buttons. It would be good to use an existing component if possible. Ideally, whatever we do we have public so other users can create componets with close buttons that match ours.

ashetland:
I've wondered about using Action too. Action would need more scales added (or a way to make them smaller) to accommodate all the sizes we need. It would also need the ability to be round. For example, Chips use a round 16px close button. Maybe rounding it could be achieved without adding that to Actions? In any other context, round Actions could be weird?

driskull:
[...]

PR #10179, opened 2024-08-28
feat(chip): add component tokens

alisonailea:
**Related Issue:** #7180

## Summary

Adds tokens, e2e, and custom-theme chromatic tests. This ended up being a larger refactor to clarify internal tokens.

### Chip Tokens

--calcite-chip-background-color: Specifies the background color of the component.
--calcite-chip-border-color: Specifies the border color of the component.
--calcite-chip-corner-radius: Specifies the corner radius of the component.
--calcite-chip-text-color: Specifies the text color of the component.
--calcite-chip-icon-color: Specifies the icon color of the component.
--calcite-chip-close-icon-color: Specifies the icon color of the close element of the component.
--calcite-chip-select-icon-color: Specifies the icon color of the selection element of the component.
--calcite-chip-select-icon-color-pressed: Specifies the icon color of the selection element of the component when active.

### Chip Group Tokens

--calcite-chip-group-items-space: Specifies spacing between items in the component.


packages/calcite-components/src/components/chip/chip.scss
```diff
@@ -1,357 +1,405 @@
...
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-chip-background-color: Specifies the background color of the component.
+ * @prop --calcite-chip-border-color: Specifies the border color of the component.
+ * @prop --calcite-chip-corner-radius: Specifies the border radius of the component.
+ * @prop --calcite-chip-text-color: Specifies the text color of the component.
+ *
```
macandcheese, line 10: I think we'd at the very least want to add `--calcite-chip-icon-color` and `--calcite-chip-selection-icon-color` (or whatever name) from the 7180 PR: https://github.com/Esri/calcite-design-system/blob/2c5e985b86621e044f60c94269edab7f83573725/packages/calcite-components/src/components/chip/chip.scss#L6

We can't just set `--calcite-icon-color` on a Calcite Chip - it has too many internally rendered icons that would be affected. The 7180 PR also had `--calcite-chip-close-icon-color` (and, props to control internal close button states, which it sounds like we did not want to pursue - I still think this prevents a lot of valid use cases)

macandcheese:
Do we want to kick off Chromatic builds for these token PR that include refactors?

alisonailea:
> Do we want to kick off Chromatic builds for these token PR that include refactors?

Yes all token PRs need to pass Chromatic tests because we need to ensure they are not creating unexpected visual changes. I've done manual checks but human error is real.

packages/calcite-components/src/components/chip/chip.scss
alisonailea: The design pattern for calcite-chip is for icons to follow the text color. If a user wants to have a special color they can do something like `calcite-chip.my-chip { --calicte-icon-color: red }`.  Similarly. IMO, all close sub-elements (should) have the same interaction pattern with a shared token set. Something like

```
--calcite-close-color
--calcite-close-background-color-hover
--calcite-close-background-color-press
```

macandcheese: Agreed they should be (ideally) consistently implemented when overridden - but not adding component-specific ones (even if it falls back to the global one) to the component means they won't be documented alongside all the other component tokens - wherever that doc is consumed (including places we can't manually supplement it with info about these generic "close-color" props).

Currently it seems like `text-color` property here also sets the selection affordance icon as well as the `icon-start` icon. At the very least I'd like to be able to not have this happen (which is the current behavior): 

The current behavior that is available (just overriding the global css prop - changes the affordance as well as the icon, which I don't think is always preferable). 

Screenshot 2024-08-28 at 1 57 43 PM

My use case for overriding an icon color is probably to match map symbology or layer display color / graphics. So I don't think I'd want the selection icon to change as well:

Screenshot 2024-08-28 at 1 57 54 PM

macandcheese (commented):
I think the spacing props need a pass here for certain configurations:
Screenshot 2024-08-28 at 2 06 15 PM

packages/calcite-components/src/components/chip/chip.scss
```diff
@@ -1,359 +1,357 @@
...
-  padding-inline-end: var(--calcite-internal-chip-padding-end);
-}
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-chip-background-color: Specifies the background color of the component.
+ * @prop --calcite-chip-border-color: Specifies the border color of the component.
+ * @prop --calcite-chip-corner-radius: Specifies the border radius of the component.
```
macandcheese, line 8: For the description here do we want to also use "corner radius"?

packages/calcite-components/src/components/chip/chip.scss
```diff
@@ -1,359 +1,357 @@
...
+ * @prop --calcite-chip-text-color: Specifies the text color of the component.
+ * @prop --calcite-chip-icon-color: Specifies the icon color of the component.
+ * @prop --calcite-close-icon-color: Specifies the icon color of the close element of the component.
```
macandcheese, line 11: I realize this is "global" but it still (IMO) would be nice to scope this to the component. That way, both --calcite-close-icon-color works (at the global level) and --calcite-chip-close-icon-color (at the component level) could work, and we don't need to document non-component-scoped tokens inside a component. 

Confirming we are *not* allowing the Chip's selection affordance to change independently of the "icon rendered from prop".

macandcheese (approved):
Looks good! Just some comments on token naming, and also confirming that:

- with this set up, users can't independently control the selection affordance icon - just the "icon rendered via prop" and the close button.

Could you kick off another Chromatic build to test latest changes? Thanks!

packages/calcite-components/src/assets/styles/includes.scss
```diff
@@ -152,3 +152,40 @@
     }
   }
 }
+
+@mixin close-button($size: "var(--calcite-internal-close-size, 1.5rem /* 24px */)", $padding: "0") {
```
jcfranco, line 156: Sidebar: we'll need to consolidate this with the `x-button` mixin above. Related https://github.com/Esri/calcite-design-system/issues/6640.

packages/calcite-components/src/components/chip/chip.scss
```diff
@@ -1,359 +1,359 @@
...
-  rounded-full
   focus-base
   justify-center
-  border
-  border-solid
-  border-color-1
   box-border
   font-medium;
 
+  border-radius: var(--calcite-chip-corner-radius, 9999px);
```
jcfranco, line 300: Sidebar: we still need a semantic token for this fallback, right?

jcfranco (approved):
Aside from a few comments, this looks great!

jcfranco:
@alisonailea Before merging, can you update the PR description with the latest component tokens?

packages/calcite-components/src/components/chip/chip.scss
```diff
@@ -1,359 +1,359 @@
...
+  opacity: 1;
+}
 
-  .multiple {
-    .select-icon {
-      --calcite-internal-chip-padding-start: var(--calcite-spacing-base);
-      --calcite-internal-chip-padding-end: var(--calcite-spacing-base);
+:host([scale="s"]) {
+  .container {
+    --calcite-internal-chip-block-size: 1.5rem /* 24px */;
```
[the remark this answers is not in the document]
alisonailea: I was doing this because these need to be replaced with the upcoming spacing tokens which will all use rems instead of px

packages/calcite-components/src/components/chip/chip.scss
alisonailea: yes this will be coming in future updates

Issue #10016, opened 2024-08-08
[Accordion]: Enhance component's interactivity states


DitwanP:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Updates the `accordion`'s interactive states as part of the larger epic in https://github.com/Esri/calcite-design-system/issues/9299.

#### Current state
![image](https://github.com/user-attachments/assets/a09a4a95-f9d4-4471-b212-c83b7fdd1218)

### Acceptance Criteria

For `appearance="solid"`:
- [ ] Update `:hover` state background color to `--calcite-color-foreground-2` and does not change text colors.
- [ ] Update `:active` state background color to `--calcite-color-foreground-3` and does not change text colors.

For `appearance="transparent"`:
- [ ] Update `:hover` state background color to `--calcite-color-transparent-hover` and does not change text colors.
- [ ] Update `:active` state background color to `--calcite-color-transparent-press` and does not change text colors.

For both:
- [ ] `:hover` and `:active` states do not change text colors
- [ ] `:focus` does not change text weight or icon color

Should released alongside issue: https://github.com/Esri/calcite-design-system/issues/4012 

![image](https://github.com/user-attachments/assets/6fb75498-21f1-46d4-bfa3-33d603b77c63)

### Which Component

`accordion`

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

**monday.com sync:** #8608462379

DitwanP:
🍠 Verified locally on `dev`

Issue #9299, opened 2024-05-09
Epic: Interactive states across the design system for hover, press, and focus


geospatialem:
### Description
Audit and update interactive states across the design system WRT the following states:
- hover
- press
- focus

cc @ashetland @SkyeSeitz 

### Proposed advantages
- Consistent and expected states
- More a11y support and benefits to our users

### Phases and components

[Figma file](https://www.figma.com/design/c6v9o8y5KgztitlvDudaQy/Interactive-state-audit-%26-improvements?node-id=1-2&t=J8pxiqWOzqfYHpD3-0)

#### Phase I: January 2025
- [x] `tile`
- [x] `pagination` https://github.com/Esri/calcite-design-system/issues/9991
- [x] `stepper-item` https://github.com/Esri/calcite-design-system/issues/10000
- [x] `menu` https://github.com/Esri/calcite-design-system/issues/10001

#### Phase II: February 2025
- [x] `navigation-logo` + `navigation-user` https://github.com/Esri/calcite-design-system/issues/10002
- [x] `tabs` https://github.com/Esri/calcite-design-system/issues/10003

#### Phase III: March 2025
- [x] `button` https://github.com/Esri/calcite-design-system/issues/9246
- [x] `chip` https://github.com/Esri/calcite-design-system/issues/10005
- [x] `combobox-item` #6245 
- [x] `combobox-item` #11646 
- [x] `action` https://github.com/Esri/calcite-design-system/issues/10007
- [x] `segmented-control` https://github.com/Esri/calcite-design-system/issues/10013
- [x] `rating` https://github.com/Esri/calcite-design-system/issues/10015

#### Phase IV: May 2025
- [x] `button` follow ups https://github.com/Esri/calcite-design-system/issues/11576
- [x] `fab` https://github.com/Esri/calcite-design-system/issues/10004
- [x] `accordion` https://github.com/Esri/calcite-design-system/issues/10016
- [x] `block` https://github.com/Esri/calcite-design-system/issues/10017
- [x] `list` https://github.com/Esri/calcite-design-system/issues/10018
- [x] `dropdown` https://github.com/Esri/calcite-design-system/issues/10019

#### Phase V: July 2025
- [x] `checkbox` https://github.com/Esri/calcite-design-system/issues/10020
- [x] `radio-button` https://github.com/Esri/calcite-design-system/issues/10021

#### Future / Ignored for now
- [x] `slider`
[...]

PR #10181, opened 2024-08-28
feat(accordion-item): style accordion item header in expanded state

alisonailea:
**Related Issue:** #7180 #4060

## Summary

While this does not create additional tokens for the background color in a -hover or -press state explicitly because this is outside of our current design patterns, it does allow a user to accomplish this by only targeting the top-level component styles.

```css
calcite-accordion-item:hover {
--calcite-accordion-item-background-color: blue;
}
calcite-accordion-item[expanded] {
--calcite-accordion-item-header-background-color: blue;
}
```

### Tokens

--calcite-accordion-item-background-color: Specifies the component's background color.
--calcite-accordion-item-border-color: Specifies the component's border color.
--calcite-accordion-item-content-space: Specifies the component's padding.
--calcite-accordion-item-header-background-color: Specifies the background color of the component's header.
--calcite-accordion-item-text-color: Specifies the component's text color.
--calcite-accordion-item-heading-text-color: Specifies the component's heading text color.
--calcite-accordion-item-icon-color: Specifies the component's default icon color.
--calcite-accordion-item-start-icon-color: Specifies the component's start icon color. Fallback to --calcite-accordion-item-icon-color or current color.
--calcite-accordion-item-end-icon-color: Specifies the component's end icon color. Fallback to --calcite-accordion-item-icon-color or current color.
--calcite-accordion-item-expand-icon-color: Specifies the component's expand icon color.
 
### Deprecated

 --calcite-accordion-border-color: Use --calcite-accordion-item-border-color. Specifies the component's border color.
 --calcite-accordion-text-color-hover: Use --calcite-accordion-item-text-color-hover. Specifies the component's main text color on hover.
--calcite-accordion-text-color-pressed: Use --calcite-accordion-item-text-color-press. Specifies the component's main text color when pressed.
--calcite-accordion-text-color: Use --calcite-accordion-item-text-color. Specifies the component's text color.


packages/calcite-components/src/components/accordion-item/accordion-item.scss
```diff
@@ -3,11 +3,17 @@
...
+ * @prop --calcite-accordion-item-background-color: Specifies the component's background color.
+ * @prop --calcite-accordion-item-border-color: Specifies the component's border color.
  * @prop --calcite-accordion-item-content-space: Specifies the component's padding.
- * @prop --calcite-accordion-text-color: Specifies the component's text color.
- * @prop --calcite-accordion-text-color-hover: Specifies the component's main text color on hover.
- * @prop --calcite-accordion-text-color-pressed: Specifies the component's main text color when pressed.
+ * @prop --calcite-accordion-item-header-background-color: Specifies the background color of the component's header.
+ * @prop --calcite-accordion-item-text-color-hover: Specifies the component's main text color on hover.
+ * @prop --calcite-accordion-item-text-color-press: Specifies the component's main text color when pressed.
+ * @prop --calcite-accordion-item-text-color: Specifies the component's text color.
```
macandcheese, line 13: Do we need two separate ones here for heading / description (and then apply whichever matches icon to icon) - They seem to use different values here? https://www.figma.com/design/vB6xFCAFeW0zKrT4nKlh8Z/Calcite-UI-Kit-%26-Styles?m=auto&node-id=3921-13&t=GRKfhHOwKkhpaNZS-1

macandcheese (approved):
Looks good! Just one comment on text colors

packages/calcite-components/src/components/accordion-item/accordion-item.scss
alisonailea: The design pattern followed the -default, -hover, -press pattern, if we were to set tokens for heading / description we would be tripling the required tokens and breaking the design pattern.

ashetland: All of our components with a heading and a description will use two different colors for the text. Accordion Item currently functions differently than the rest in that it has color changes onHover.

At rest:
- heading is text.2
- description + icon are text.3

onHover:
- heading is text.1
- description + icon are text.2
- weird animation timings are present here: the icon and the description "fade" in at different rates while the heading doesn't not

When expanded (matches onHover):
- heading is text.1
- description + icon are text.2
- no hover state when expanded

No other components with headings and descriptions behave this way and in issue #10016 (and epic #9299) we've outlined specs to consolidate these patterns. Even with the updated pattern though, the heading and description will be different colors, but they won't change on hover/press.

Since they use different colors I'd expect separate props for them. As for the hover/pressed tokens, I'm less sure since we're in the process of consolidating and aligning some of these different patterns across the system. If it's important to define these tokens for the current state, then I suppose they could be deprecated later when the pattern is updated?

alisonailea: So here's what I've done. 

I have left the default styling as-is so that

At rest:
heading is text.2
description + icon are text.3

onHover:
heading is text.1
description + icon are text.2
weird animation timings are present here: the icon and the description "fade" in at different rates while the heading doesn't not

When expanded (matches onHover):
heading is text.1
description + icon are text.2
no hover state when expanded

HOWEVER, the component tokens available to the user reflect the future pattern where-in there is no extra hover/press state tokens.

ashetland: Dig it.

macandcheese: Just confirming:

Should this have `--calcite-accordion-item-description-text-color` in addition to `--calcite-accordion-item-heading-text-color`?

In this set up - a user can override the internally rendered icons, but only through `--calcite-icon-color`, nothing scoped to Accordion?

alisonailea: --calcite-accordion-item-text-color == --calcite-accordion-item-description-text-color

Issue #10182, opened 2024-08-29
Loader no longer animating


COV-GIS:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Loader no longer animating.

### Expected Behavior

Loader should be animated.

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/loader/

### Reproduction Steps

I came across this with button loading: https://developers.arcgis.com/calcite-design-system/components/button/ Check `loading` property checkbox.

Can also be seen on any reference page when the sample is loading.

### Reproduction Version

2.12.0

### Regression?

2.11.1

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

geospatialem:
Possibly related to https://github.com/Esri/calcite-design-system/pull/9957.

geospatialem:
After closer inspection it appears the regression is from #10147 from the following lines: https://github.com/Esri/calcite-design-system/commit/3c4775ce1d59d72389511287ee51c3f36ac403f7#diff-98ccabacd6112ddd54974f877b97ee4e4db45ca390ab1326c026a0f70b5d0dd4.

We should explore screenshot testing that supports multiple snapshots after timeline to ensure the component spinning is performing as-expected.

geospatialem:
Verified in `2.13.0-next.5` with https://codepen.io/geospatialem/pen/poXQJWz

PR #10184, opened 2024-08-29
fix(loader): restore animation

Elijbet:
**Related Issue:** #10182

## Summary
While the new lint rule catches non-kebab-cased names at `@function` definition, it does not catch wrong casing when applied. This fixes missed applications.


packages/calcite-components/src/components/loader/loader.e2e.ts
```diff
@@ -55,4 +55,43 @@ describe("calcite-loader", () => {
...
     expect(loader.getAttribute("id").length).toEqual(36);
   });
+
+  it("animates loader segment offset", async () => {
+    const page = await newE2EPage();
+    await page.setContent(`<calcite-loader></calcite-loader>`);
+
+    const duration = await page.evaluate(() => {
+      const loader = document.querySelector("calcite-loader");
+      const style = getComputedStyle(loader);
```
driskull, line 65: I think you should be able to do: `const style = getComputedStyle(loader).animationDuration;`

packages/calcite-components/src/components/loader/loader.e2e.ts
```diff
@@ -55,4 +55,43 @@ describe("calcite-loader", () => {
...
+      const durationValue = style.getPropertyValue("animation-duration");
+      const match = durationValue.match(/(\d+(\.\d+)?)/);
```
driskull, line 67: I don't think you need to do a match here.

packages/calcite-components/src/components/loader/loader.e2e.ts
driskull: Applies to other places.

packages/calcite-components/src/components/loader/loader.e2e.ts
```diff
@@ -55,4 +55,43 @@ describe("calcite-loader", () => {
...
+      return parseFloat(match[0]);
+    });
+    console.log("duration", duration);
+
+    const timeAt60Percent = duration * 0.6;
+    console.log("timeAt60Percent", timeAt60Percent);
+
+    const initialOffset = await page.evaluate(() => {
```
driskull, line 75: maybe make this a `getStrokeDashOffset` function since its logic is repeated.

driskull (approved):
👍

jcfranco (approved):

Issue #8639, opened 2024-01-24
Popup of the input-time-picker is not being closed when empty after clicking outside the control


pandzel:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Popup stays on the screen after clicking anywhere outside the control if no hours selected in the popup.

### Expected Behavior

Popup should disappear. It does disappear but after selecting full time in the popup. Also, input-date-picker works correctly in the identical scenario.

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/input-time-picker/

### Reproduction Steps

- access https://developers.arcgis.com/calcite-design-system/components/input-time-picker/
- click on the control to open the popup
- without entering any value, click anywhere outside the control
- note that popup stays on the screen

### Reproduction Version

2.2.0

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Online

nwhittaker:
Seeing the same bug with non-empty time pickers as well ([repro][1]):

https://github.com/Esri/calcite-design-system/assets/421496/12288ab6-8197-41d1-b259-682c0a65e3b5

---
After opening the time picker by clicking its caret button, it doesn't exhibit the expected auto-close behavior (closed by outside click) until the user changes its value in the picker:

https://github.com/Esri/calcite-design-system/assets/421496/2e7b0e7c-d44b-4a0e-a150-9e1e67fcb5dc

---
Interestingly, if the non-empty picker is opened by clicking anywhere else on the input, the expected auto-close behavior is maintained:

https://github.com/Esri/calcite-design-system/assets/421496/3ff5fba3-1f50-47dd-810e-f10dfdda7db7

[1]: https://codepen.io/nwhittaker-esri/pen/QWPRKGE

hgonzago:
We are also seeing this when working in the JavaScript SDK's Editor widget. Is there any update on when this could be addressed? We have gotten questions brought up from Enterprise and AGOL MapViewer users.

A note regarding this is that the issue only appears to happen if the time picker input is empty. If a value is added and then you click outside of it, it will close and allow another input to take focus.

[...]

PR #10189, opened 2024-08-29
fix(input-time-picker): automatically close when clicked outside

driskull:
**Related Issue:** #8639

## Summary

- add autoClose to internal popover
- add test
- update referenceElement to use element rather than id


eriklharper:
I'm noticing some undesirable behavior when toggling the caret in `input-time-picker`.  The popup doesn't seem to open at all when clicking the caret, but it seems to work well when just focusing the text input and then blurring it.

driskull:
@eriklharper checked in a fix. Can you review again?

eriklharper (approved):
> @eriklharper checked in a fix. Can you review again?

Heck yeah, seems to be working fine now :)

packages/calcite-components/src/components/input-time-picker/input-time-picker.tsx
```diff
@@ -1073,7 +1068,7 @@ export class InputTimePicker
 
   renderToggleIcon(open: boolean): VNode {
     return (
-      <span class={CSS.toggleIcon}>
+      <span class={CSS.toggleIcon} slot="action">
```
jcfranco, line 1071: Nice!

jcfranco (approved):
🚀

Issue #8617, opened 2024-01-16
Multiple Switch components keep focus state in Safari


tyanthony:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

In Safari browser, when there are multiple `switch` components in a form, each maintains its focus state until it is interacted with again directly.

Ex:
![image](https://github.com/Esri/calcite-design-system/assets/17069279/31be882a-4891-480f-a69f-c589afc55863)

### Expected Behavior

The switch should lose focus when another element, other than itself, is clicked or interacted with.

### Reproduction Sample

https://codepen.io/tyleranthony/pen/PoLpZWy?editors=100

### Reproduction Steps

1. Open an environment in Safari browser
2. Add multiple `switch` components 
3. Toggle multiple of them
4. Notice that they maintain their focus state 

### Reproduction Version

2.1.0

### Relevant Info

Browser: Safari

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Mission

driskull:
Seems like the focus selector should be moved from the host to the container div maybe?

macandcheese:
> Seems like the focus selector should be moved from the host to the container div maybe?

Yeah - you can focus either the host, the container, or both, depending on how you focus:
![Screenshot 2024-02-20 at 10 26 40 AM](https://github.com/Esri/calcite-design-system/assets/4733155/35ff1f15-f484-41ea-a084-edb832e2cf3f)
![Screenshot 2024-02-20 at 10 26 35 AM](https://github.com/Esri/calcite-design-system/assets/4733155/44ab1f66-03b7-4c01-ba67-a8a068811b3e)
![Screenshot 2024-02-20 at 10 26 31 AM](https://github.com/Esri/calcite-design-system/assets/4733155/d297b39f-c2f5-4cb2-8454-1df7df886381)

DitwanP:
🍡 Verified on `2.13.0-next.4`
https://codepen.io/Ditwan-Price/pen/yLdQBNW?editors=1000

PR #10191, opened 2024-08-29
fix(switch): fix focus state styling in Safari

driskull:
**Related Issue:** #8617

## Summary

- fix focus state styling in Safari


packages/calcite-components/src/components.d.ts
```diff
@@ -9,93 +9,193 @@ import { Alignment, Appearance, CollapseDirection, FlipContext, IconType, Kind,
 import { RequestedItem } from "./components/accordion/interfaces";
 import { IconNameOrString } from "./components/icon/interfaces";
 import { RequestedItem as RequestedItem1 } from "./components/accordion-item/interfaces";
+import { ActionMessages } from "./components/action/assets/action/t9n";
```
macandcheese, line 12: I'm never sure - should we be committing these `components.d.ts` when there aren't related changes files or not?

macandcheese (approved):
LGTM 

image

packages/calcite-components/src/components.d.ts
driskull: Yeah I'm not sure either. Sometimes it commits it for me when I don't want it to.

Issue #10192, opened 2024-08-30
list: filter change wrongly calls calciteListChange


AdelheidF:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

From the doc:
calciteListChange | Fires when the component's selected items have changed.

If I only change the filter it throws a calciteListFilter event AND a calciteListChange event. 
This means I need to compare the list.selectedItems with my currently saved selected items to check if a selection change actually happened.

cc @driskull 

### Expected Behavior

When only filter changes it only throws a calciteListFilter event

### Reproduction Sample

https://codepen.io/afreitag/pen/BagOpGZ 

### Reproduction Steps

Open codepen and watch the console, click 2 items and then enter "p" in the filter.

### Reproduction Version

2.11.1

### Regression?

yes compared to pick-list

### Priority impact

impact - p1 - need for current milestone

### Impact

There are workflows where the list appears in a popover and we close it as soon as the selection changes.
Found this while moving off calcite-pick-list.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Map Viewer

PR #10195, opened 2024-08-30
fix(list): do not emit `calciteListChange` during a filter

driskull:
**Related Issue:** #10192

## Summary

- do not call calciteListChange during a filter
- add test


driskull:
This could be for the patch if we want it to be.

jcfranco (approved):

Issue #10188, opened 2024-08-29
Sass compiler does not throw an error when expected


Elijbet:
### Priority impact

p - low

### Summary

Sass names refactor in the `loader` revealed that Sass compiler did not throw an error when it was referencing a function that does not exist.

### Desired Outcome

Sass compiler should throw an error when referencing a function that does not exist.

jcfranco:
This is a [known limitation for unknown Sass functions](https://sass-lang.com/documentation/at-rules/function/#plain-css-functions:~:text=gradient(%23f2ece4%2C%20%23e1d7d2)-,%E2%9A%A0%EF%B8%8F%20Heads%20up!,on%20your%20stylesheet%E2%80%99s%20output%20to%20be%20notified%20when%20this%C2%A0happens!,-%F0%9F%92%A1%20Fun%20fact%3A):

> ### ⚠️ Heads up!
>
> Because any unknown function will be compiled to CSS, it’s easy to miss when you typo a function name. Consider running a [CSS linter](https://stylelint.io/) on your stylesheet’s output to be notified when this happens!
 
 Stylelint's [`function-no-unknown`](https://github.com/stylelint-scss/stylelint-scss/tree/master/src/rules/function-no-unknown) rule can help with this. Local testing looks good, but we need to specify which functions to ignore, as it doesn't do this automatically. Currently, we only have 2 custom sass functions, so we can add coverage for this easily. 
 
Keeping track of custom functions will be a manual process for now, but we may be able to automate this.

jcfranco:
> Keeping track of custom functions will be a manual process for now, but we may be able to automate this.

This step has ben automated via a `pre-commit` hook.

DitwanP:
🍭 Verified locally on `dev`

PR #10200, opened 2024-08-31
chore(linting): enable `scss/function-no-unknown` rule

jcfranco:
**Related Issue:** #10188 

## Summary

The Sass compiler does not error on unknown functions, but we can use Stylelint's [`function-no-unknown`](https://github.com/stylelint-scss/stylelint-scss/tree/master/src/rules/function-no-unknown) to ensure function names are properly referenced.


Elijbet (approved):
🔧🛠️⚙️

Issue #6172, opened 2022-12-26
Consistency in naming: Width / Width Scale / Height Scale


macandcheese:
### Description

There are a few instances where we use "heightScale", "widthScale" to determine sizes for components, usually accepting s/m/l. There are also instances of "width" used frequently. 

We should look at consolidating these patterns / naming... 

`heightScale` - `calcite-flow-item`, `calcite-panel`, `calcite-shell-center-row`
`detachedHeightScale` - `calcite-shell-panel`
`widthScale` - `-calcite-shell-panel`, `calcite-flow-item`, `calcite-panel`

Many other components use `width` - some are `s/m/l` choice, others use `full / auto / half` etc.

### Acceptance Criteria

A consistent naming and usage pattern is defined and implemented. I do think we should continue with the proposed changes to panel to remove height / width settings in general and have it behave as a block level element that fills the container it is placed in (most of the time, this would be shell panel, and its defined width / height scale or explicit value)

### Which Component

Above

### Esri team

Calcite (design)

ashetland:
Agree, this would be a good one to dig into.

geospatialem:
For 2024 we should work towards prop deprecation, and for the 2025 breaking change consider removal of any of the deprecated components.

Elijbet:
I scanned through components and made a list of what they use. The majority of components use `scale`. These are some outliers:

**Dialog
Dropdown
Modal**
```/** Specifies the size of the component. */	
  @Prop({ reflect: true }) **scale**: Scale = "m";

 /** Specifies the width of the component. */
  @Prop({ reflect: true }) **widthScale**: Scale = "m";```


**Notice
Segmented-control
Select
Split-button**
```/** Specifies the size of the component. */	
  @Prop({ reflect: true }) **scale**: Scale = "m";

  /** Specifies the width of the component. */
  @Prop({ reflect: true }) **width**: Width = "auto";```

**Button 
Tile-select**
  `@Prop({ reflect: true }) width: Width = "auto";`

**Shell-Panel**
  `@Prop({ reflect: true }) widthScale: Scale = "m";`

**Sheet**
```  /**
   * When `position` is `"inline-start"` or `"inline-end"`, specifies the width of the component.
   */
  @Prop({ reflect: true }) widthScale: Scale = "m";
  /**
   * When `position` is `"block-start"` or `"block-end"`, specifies the height of the component.
   */
  @Prop({ reflect: true }) heightScale: Scale = "m";```

[...]

PR #10202, opened 2024-09-01
deprecate: deprecate `widthScale/heightScale` in favor of `width/height`

Elijbet:
**Related Issue:** #6172

## Summary
Refactor to consolidate `width/height` and `widthScale/heightScale` into a single property `width` with `s / m / l / auto / full`  as options, and `height` with `s / m / l`. Deprecate `widthScale` and `heightScale` properties and the `half` value.


Elijbet:
What’s a good way to deprecate “half”? Components listed below still use it. Previously width could specify half, but the new consolidated width is not intended to provide for half as an option.
button
notice
select
split-button

packages/calcite-components/src/components/segmented-control/segmented-control.tsx
Elijbet, line 1: Could you check me on the documentation here @jcfranco. Since we don't we don't have an established pattern with deprecated props, I followed a similar example with shell-panel displayMode for `float`, as recommended by @geospatialem.

packages/calcite-components/src/components/segmented-control/segmented-control.tsx
```diff
@@ -158,7 +158,7 @@ export class SegmentedControl
     valueMissing: false,
   };
 
-  /** Specifies the width of the component. */
+  /** Specifies the width of the component. [Deprecated] The `"half"` value is deprecated, use `"full"` instead. */
```
Elijbet, line 161: Could you check me on the documentation here @jcfranco. Since we don't we don't have an established pattern with deprecated props, I followed a similar example with shell-panel displayMode for float, as recommended by @geospatialem.

packages/calcite-components/src/components/dialog/dialog.scss
```diff
@@ -187,13 +187,15 @@ calcite-panel {
  * Sizes
  */
 @mixin dialog-size($size, $width) {
-  :host([width-scale="#{$size}"]) .dialog {
+  :host([width-scale="#{$size}"]) .dialog,
+  :host([width="#{$size}"]) .dialog {
```
Elijbet, line 191: Not sure what's a good way to set this up. If I set one of these as a default "m", it overrides the other one set to a different value. If I leave the default out so they have equal weight, it takes the full width of the screen when none are set. How do we deprecate mindfully?

packages/calcite-components/src/components/dialog/dialog.scss
```diff
@@ -214,7 +215,7 @@ calcite-panel {
   }
 }
 
-:host([width="small"]) .dialog {
+:host .width-s .dialog {
```
driskull, line 218: You don't need the host here.

packages/calcite-components/src/components/dropdown/dropdown.e2e.ts
```diff
@@ -1175,7 +1187,7 @@ describe("calcite-dropdown", () => {
       filterInput.value = "numbers";
     });
 
-    expect(dropdownContentHeight.height).toBe("88px");
+    expect(dropdownContentHeight.height).toBe("72px");
```
jcfranco, line 1190: Should this and related widths have decreased? Applies to other size changes.

packages/calcite-components/src/components/dropdown/resources.ts
```diff
@@ -1,3 +1,9 @@
 export const SLOTS = {
   dropdownTrigger: "trigger",
 };
+
+export const CSS = {
+  calciteDropdownWrapper: "calcite-dropdown-wrapper",
```
jcfranco, line 6: Let's simplify the name to `wrapper`.

packages/calcite-components/src/components/modal/modal.scss
```diff
@@ -287,7 +287,8 @@ slot[name="primary"] {
         max-block-size: unset;
       }
     }
-    :host([width-scale="#{$size}"][docked]) .container {
+    :host([width-scale="#{$size}"][docked]) .container,
```
jcfranco, line 290: For consistency, can this use `.width`? Maybe something like this (⚠️ untested ⚠️)?

```scss
:host([docked]) .container .width-#{$size} {
  // ...
}
```

packages/calcite-components/src/components/shell-center-row/shell-center-row.e2e.ts
```diff
@@ -79,10 +79,10 @@ describe("calcite-shell-center-row", () => {
     await page.setContent(pageContent);
 
     const element = await page.find("calcite-shell-center-row");
-
+    const containerEl = element.shadowRoot.querySelector(`.${CSS.container}`);
     await page.waitForChanges();
 
-    expect(element.shadowRoot.lastElementChild).toHaveClass(CSS.actionBarContainer);
+    expect(containerEl.lastElementChild).toHaveClass(CSS.actionBarContainer);
```
jcfranco, line 85: ✨🏆✨

packages/calcite-components/src/components/shell-center-row/shell-center-row.e2e.ts
```diff
@@ -55,7 +55,7 @@ describe("calcite-shell-center-row", () => {
     `;
     await page.setContent(pageContent);
 
-    const element = await page.find("calcite-shell-center-row >>> div:first-of-type");
+    const element = await page.find(`calcite-shell-center-row >>> .${CSS.container} div:first-of-type`);
```
jcfranco, line 58: ✨🏅✨

packages/calcite-components/src/components/shell-center-row/shell-center-row.scss
```diff
@@ -36,19 +40,19 @@
   @apply self-start;
 }
 
-:host([height-scale="s"]) {
+:host .height-s {
```
jcfranco, line 43: is `:host` needed here for specificity or could we remove it? Applies to similar lines.

packages/calcite-components/src/components/shell-panel/shell-panel.e2e.ts
```diff
@@ -118,7 +135,7 @@ describe("calcite-shell-panel", () => {
       "calcite-shell-panel",
       (panel: HTMLCalciteShellPanelElement, containerClass: string, contentClass: string) => {
         const container = panel.shadowRoot.querySelector(containerClass);
-        return container.firstElementChild.className == contentClass;
+        return container.firstElementChild.className.split(" ")[0] == contentClass;
```
jcfranco, line 138: Can you use `container.firstElementChild.classList.contains(contentClass)` instead? The current approach is brittle.

packages/calcite-components/src/utils/dynamicCSSConst.ts
```diff
@@ -0,0 +1,7 @@
+export const getWidth = (width: string, widthScale: string): string => {
```
jcfranco, line 1: Since `height` and `width` are shared, WDYT about adding a `CSS` export with those classes for other components to use?

jcfranco, line 1: Can you narrow the return type? Applies to both.

packages/calcite-components/src/utils/dynamicCSSConst.ts
```diff
@@ -0,0 +1,7 @@
...
+  return width ? `width-${width}` : widthScale ? `width-${widthScale}` : "";
+};
+
+export const getHeight = (height: string, heightScale: string): string => {
```
jcfranco, line 5: Both of these helpers have some code in common, so we could add something like the following for both to share or use it directly instead of having 2 similar functions.

```tsx
function getDimension(type: "width" | "height", size: string, scale: string): string {
  return size ? `${type}-${size}` : scale ? `${type}-${scale}` : "";
};
```

packages/calcite-components/src/components/dropdown/dropdown.tsx
```diff
@@ -252,7 +258,10 @@ export class Dropdown
           </div>
           <div
             aria-hidden={toAriaBoolean(!open)}
-            class="calcite-dropdown-wrapper"
+            class={{
+              [`${CSS.calciteDropdownWrapper}`]: true,
```
jcfranco, line 262: Interpolation isn't needed here, you can simplify to:

```tsx
[CSS.calciteDropdownWrapper]: true,
```

packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -269,7 +277,10 @@ export class Dialog
             aria-description={description}
             aria-label={heading}
             aria-modal={toAriaBoolean(this.modal)}
-            class={CSS.dialog}
+            class={{
+              [CSS.dialog]: true,
+              [getWidth(this.width, this.widthScale)]: !!this.width || !!this.widthScale,
```
jcfranco, line 282: This can be simplified a bit.

```ts
!!(this.width || this.widthScale)
```

May apply to other spots.

packages/calcite-components/src/utils/dynamicCSSConst.ts
jcfranco, line 1: Can you add a spec test for this util?

jcfranco, line 1: Can we rename this module to `dynamicClasses`? In the future, we could add a util for `scale`.

Elijbet: As per our convo, because we now use props to generate the scale (and the util is applied throughout) I'm not sure where we'd have a use for `CSS.width` or `CSS.height`.  @jcfranco

Elijbet: This just returns a class, so type would be just string, no? @jcfranco

packages/calcite-components/src/components/dropdown/resources.ts
```diff
@@ -1,3 +1,9 @@
 export const SLOTS = {
   dropdownTrigger: "trigger",
 };
+
+export const CSS = {
+  wrapper: "calcite-dropdown-wrapper",
+  width: "width",
```
jcfranco, line 7: Per our discussion, `width` and `widthScale` are no longer needed.

packages/calcite-components/src/utils/dynamicClasses.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+  const sizes = ["small", "medium", "large", ""] as const;
+  const scales = ["small", "medium", "large", ""] as const;
+
+  types.forEach((type) => {
+    sizes.forEach((size) => {
+      scales.forEach((scale) => {
+        const expected = size ? `${type}-${size}` : scale ? `${type}-${scale}` : "";
+        it(`should return "${expected}" for type="${type}", size="${size}", scale="${scale}"`, () => {
+          const result = getDimension(type, size, scale);
+          expect(result).toBe(expected);
```
jcfranco, line 14: Nitpick: you could inline `result` here.

packages/calcite-components/src/utils/dynamicClasses.ts
```diff
@@ -0,0 +1,3 @@
+export function getDimension(type: "width" | "height", size: string, scale: string): string {
```
jcfranco, line 1: Can we type `size` and `scale`?

jcfranco, line 1: Can you add a more specific return type? TS supports [template literal types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html).

packages/calcite-components/src/utils/dynamicClasses.spec.ts
```diff
@@ -0,0 +1,19 @@
+import { getDimension } from "./dynamicClasses";
+
+describe("getDimension", () => {
+  const types = ["width", "height"] as const;
+  const sizes = ["s", "m", "l", ""] as const;
+  const scales = ["s", "m", "l", ""] as const;
```
jcfranco, line 6: You'll need to use `undefined` since `""` isn't a valid value for these properties.

jcfranco (approved):
Once feedback is addressed, this should be good to merge!

📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏
📏📐📏📏📏📐📏📏📐📐📏📏📐📐📐📏📏📐📐📐📏📐📐📐📐📏📐📏
📏📐📐📏📏📐📏📐📏📏📐📏📏📐📏📏📐📏📏📏📏📐📏📏📏📏📐📏
📏📐📏📐📏📐📏📐📏📏📐📏📏📐📏📏📐📏📏📏📏📐📐📐📏📏📐📏
📏📐📏📏📐📐📏📐📏📏📐📏📏📐📏📏📐📏📏📏📏📐📏📏📏📏📏📏
📏📐📏📏📏📐📏📏📐📐📏📏📐📐📐📏📏📐📐📐📏📐📐📐📐📏📐📏
📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏📏

packages/calcite-components/src/components/interfaces.ts
```diff
@@ -28,7 +29,7 @@ export type SelectionMode =
   | "multiple";
 export type Scale = "s" | "m" | "l";
 export type Status = "invalid" | "valid" | "idle";
-export type Width = "auto" | "half" | "full";
+export type Width = "s" | "m" | "l" | "auto" | "half" | "full";
```
eriklharper, line 32: Could you do `export type Width = Scale & "auto" | "half" | "full";`?  Same for `Height` above.

Elijbet: On the second thought, I feel like it might make things a bit confusing to read. `Height`, `Width`, and `Scale` are different props even though they have an overlap in type (with type Scale). s/m/l for scale does something else for width/height. Should they just have their own type names? It hurts my eyes to read type Height = Scale :)

Elijbet:
This PR has had test timeout issues on local machine and coupled with lumina migration it was cleaner to start a new branch [deprecate: deprecate widthScale/heightScale in favor of width/height](https://github.com/Esri/calcite-design-system/pull/10786#top) #10786

Issue #8154, opened 2023-11-09
Improve <calcite-input> suffixes and prefixes


hcampos-professional:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

When using a `<calcite-input>` or `<calcite-input-number>` with a `prefix-text` or `suffix-text`, users are finding it difficult to understand that the suffix and prefix are not clickable. Matters complicate further if the `clearable` property/attribute is set, since the button looks so similar to the suffix and prefix:

![CleanShot 2023-11-09 at 16 33 22](https://github.com/Esri/calcite-design-system/assets/136828178/2dcf9469-f1b6-423a-bf58-49c39ed2d113)

Another issue is that, when multiple inputs are placed next to each other, the suffixes and prefixes won't necessarily have the same with, which doesn't look great.

After some internal discussions, a proposal was floated: placing the suffix and/or prefix inside of the input itself. Here is a mockup of the proposal:

![image](https://github.com/Esri/calcite-design-system/assets/136828178/099f926d-2e6f-4389-a43b-a1236f42b252)

### Acceptance Criteria

- When `prefix-text` and/or `suffix-text` are specified, they should appear inside of the input instead of inside a bordered container.

### Which Component

- `<calcite-input>`
- `<calcite-input-*>`

### Example Use Case

The use case we're exploring at the moment is displaying multiple numerical values of a tool stacked vertically, each value with its unit displayed as a `suffix-text`. In this scenario, having all the suffixes with different widths and having the units appear clickable is not desirable.

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

macandcheese:
We may need to explore ways of making the prefix visually distinct from the input content. Suffix doesn’t suffer from the same issue. 

It may still be nice to offer way to set the width of these via css property to ensure consistent widths, either when prefix/suffix or both are used. You might still end up with the “width stagger” depending on set up.

Anastasiia-Boleiko:
The proposal looks great, +1 from the ArcGIS Urban team.
The current number input with a unit looks like:
[...]

PR #10206, opened 2024-09-03
feat(input, input-number, input-text): add prefix and suffix width css tokens

aPreciado88:
**Related Issue:** [#8154](https://github.com/Esri/calcite-design-system/issues/8154)

## Summary

Add `--calcite-input-prefix-size` and `--calcite-input-suffix-size` css tokens to allow users to set custom `prefix` and `suffix` widths.

### DIFF
--- a/packages/calcite-components/src/components/input-number/input-number.stories.ts
+++ b/packages/calcite-components/src/components/input-number/input-number.stories.ts
@@ -216,3 +216,19 @@ export const widthSetToBreakpoints_TestOnly = (): string =>
       value="123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890"
     ></calcite-input-number>
   `);
+
+export const inputNumberPrefixAndSuffixWidth = (): string => html`
+  <calcite-input-number
+    placeholder="Placeholder text"
+    prefix-text="prefix"
+    suffix-text="suffix"
+    style="--calcite-input-number-prefix-width: 100px; --calcite-input-number-suffix-width: 100px"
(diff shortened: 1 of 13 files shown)


packages/calcite-components/src/components/input-text/input-text.scss
```diff
@@ -292,10 +301,12 @@ input[type="text"]::-ms-reveal {
 .prefix {
   @apply order-2;
   border-inline-end-width: theme("borderWidth.0");
+  inline-size: var(--calcite-input-text-prefix-width, auto);
```
alisonailea, line 304: because input, input-number, and input-text are nearly identical and share most of the same styles/props I think we can simplify this down to `--calcite-input-prefix-size` and `--calcite-input-suffix-size` for all components.

packages/calcite-components/src/components/input-number/input-number.stories.ts
```diff
@@ -216,3 +216,19 @@ export const widthSetToBreakpoints_TestOnly = (): string =>
       value="123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890"
     ></calcite-input-number>
   `);
+
+export const inputNumberPrefixAndSuffixWidth = (): string => html`
```
alisonailea, line 220: follow https://github.com/Esri/calcite-design-system/wiki/component-tokens#storybook for customTheming instructions for new tokens

alisonailea (changes requested):
Looks good but I think we can simplify this down a bit and reduce chromatic tests

aPreciado88:
@alisonailea I updated the PR with the requested changes.

alisonailea (approved):

PR #10221, opened 2024-09-05
feat(checkbox): add component tokens

aPreciado88:
**Related Issue:** [#7180](https://github.com/Esri/calcite-design-system/issues/7180)

## Summary

Add `checkbox` component tokens.

`--calcite-checkbox-shadow-color`: Specifies the component's color.
`--calcite-checkbox-shadow-color-hover`: Specifies the component's color when hovered.
`--calcite-checkbox-color`: Specifies the component's font color.


jcfranco (commented):
`--calcite-checkbox-size` is an existing prop, so let's remove it from the PR description. Other than that, proposed component token props LGTM!

packages/calcite-components/src/components/checkbox/checkbox.scss
```diff
@@ -4,16 +4,31 @@
  * These properties can be overridden using the component's tag as selector.
  *
  * @prop --calcite-checkbox-size: Specifies the component's height and width.
+ * @prop --calcite-checkbox-box-shadow: Specifies the component's svg box color.
```
jcfranco, line 7: I should've spotted this sooner, but can you revisit the prop description to not mention implementation details (i.e., svg)? Applies to new props.

jcfranco (approved):
👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾
👾🪙👾👾👾🪙🪙🪙🪙👾🪙🪙🪙🪙🪙👾🪙👾👾👾🪙👾🪙👾
👾🪙👾👾👾🪙👾👾👾👾👾👾🪙👾👾👾🪙🪙👾🪙🪙👾🪙👾
👾🪙👾👾👾🪙👾🪙🪙👾👾👾🪙👾👾👾🪙👾🪙👾🪙👾🪙👾
👾🪙👾👾👾🪙👾👾🪙👾👾👾🪙👾👾👾🪙👾👾👾🪙👾👾👾
👾🪙🪙🪙👾🪙🪙🪙🪙👾👾👾🪙👾👾👾🪙👾👾👾🪙👾🪙👾
👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾👾

aPreciado88:
@ashetland @SkyeSeitz Could you please take a look at the new chromatic build?

cc @alisonailea

packages/calcite-components/src/components/checkbox/checkbox.scss
```diff
@@ -4,16 +4,31 @@
  * These properties can be overridden using the component's tag as selector.
  *
  * @prop --calcite-checkbox-size: Specifies the component's height and width.
+ * @prop --calcite-checkbox-box-shadow: Specifies the component's color.
```
alisonailea, line 7: Upldate this to `--calcite-checkbox-box-shadow-color`

packages/calcite-components/src/components/checkbox/checkbox.scss
```diff
@@ -4,16 +4,31 @@
...
+ * @prop --calcite-checkbox-box-shadow-hover: Specifies the component's color when hovered.
```
alisonailea, line 8: `--calcite-checkbox-box-shadow-color-hover`

alisonailea (approved):
just some minor naming tweaks

packages/calcite-components/src/components/checkbox/checkbox.scss
macandcheese: I thought we were using `shadow`, not `box-shadow` here? https://github.com/Esri/calcite-design-system/wiki/component-tokens#3-review-component-for-tokenizable-styles

packages/calcite-components/src/components/checkbox/checkbox.scss
macandcheese: Just sanity check here on our patterns as it has changed a few times - this is preferable to calcite-checkbox:hover { --calcite-checkbox-shadow: "whatever" },  correct?

aPreciado88:
I updated the tokens to `--calcite-checkbox-box-shadow-color` and `--calcite-checkbox-box-shadow-color-hover`.
Can you please confirm if this is okay? @alisonailea @macandcheese

packages/calcite-components/src/components/checkbox/checkbox.scss
aPreciado88: I renamed this one to `--calcite-checkbox-shadow-color`, as suggested in the doc above.

packages/calcite-components/src/components/checkbox/checkbox.scss
aPreciado88: I renamed this one to `--calcite-checkbox-shadow-color-hover`

jcfranco: > Just sanity check here on our patterns as it has changed a few times - this is preferable to calcite-checkbox:hover { --calcite-checkbox-shadow: "whatever" }, correct?

I believe so, but I'll let @alisonailea confirm. This approach is more consistent and intuitive than the previous one. It's also not mentioned as an example in the latest Component Tokens wiki.

packages/calcite-components/src/components/checkbox/checkbox.scss
macandcheese: I realize it's basically being used as a border - but that begs the question.. Shouldn't this be a separate `--aclcite-checkbox-border-color` (obfuscating the implementation detail of it using box-shadow from the user) - and then a separate `calcite-checkbox-shadow` css prop that defaults to `shadow-none` to allow for theming?

jcfranco: @macandcheese Right again. I thought these new props were used for a drop shadow, but I overlooked how they were being applied.

@aPreciado88 The shadow props will need to be updated. The `themed` helper handles this and another case [here](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/tests/commonTests/themed.ts#L405-L408). You can use this as a reference to confirm if we’re dealing with a border prop or not. LMK if you have any questions.

Sorry for the hassle, y'all!

aPreciado88: I updated the tokens to `--calcite-checkbox-border-color` and `--calcite-checkbox-border-color-hover`.

packages/calcite-components/src/components/checkbox/checkbox.scss
```diff
@@ -63,7 +72,7 @@
 :host([hovered]) .toggle,
 :host .toggle:hover {
   .check-svg {
-    box-shadow: inset 0 0 0 2px var(--calcite-color-brand);
+    box-shadow: inset 0 0 0 2px var(--calcite-checkbox-border-color-hover, var(--calcite-color-brand));
```
alisonailea, line 75: I think we can simplify this. There is no reason to have a -hover token here if we simplify the selectors.

```css
:host([hovered]) .toggle,
:host(:hover) .toggle {
```

packages/calcite-components/src/custom-theme/checkbox.ts
```diff
@@ -1,5 +1,10 @@
 import { html } from "../../support/formatting";
 
+export const checkboxTokens = {
+  calciteCheckboxSize: "",
+  calciteCheckboxColor: "",
```
alisonailea, line 5: add `calciteCheckboxBorderColor`

alisonailea (changes requested):
Sorry just a few more things

packages/calcite-components/src/components/checkbox/checkbox.scss
jcfranco: I’m confused about when I should expect a stateful prop versus a non-stateful one. This seems to contradict the [recent CSS props added to the action family](https://github.com/Esri/calcite-design-system/pull/10058/) (e.g., `--calcite-action-background-color-hover`).

Am I missing something? There’s no mention or example of this in the Component Tokens wiki.

This was also brought up earlier ([see discussion](https://github.com/Esri/calcite-design-system/pull/10221#discussion_r1752671869)).

Could you clarify?

packages/calcite-components/src/components/checkbox/checkbox.scss
alisonailea: Yeah I'm sorry. I agree with @macandcheese and @jcfranco

packages/calcite-components/src/components/checkbox/checkbox.scss
alisonailea: Agreed.

packages/calcite-components/src/components/checkbox/checkbox.scss
alisonailea: Sorry. I'm mistaken. Yeah I'm sorry. Per https://github.com/Esri/calcite-design-system/wiki/component-tokens#naming-pattern we're using -hover here

alisonailea (approved):
just add the `calciteCheckboxBorderColor` and we're good to go

packages/calcite-components/src/components/checkbox/checkbox.scss
```diff
@@ -4,16 +4,31 @@
  * These properties can be overridden using the component's tag as selector.
  *
  * @prop --calcite-checkbox-size: Specifies the component's height and width.
+ * @prop --calcite-checkbox-border-color: Specifies the component's color.
+ * @prop --calcite-checkbox-border-color-hover: Specifies the component's color when hovered.
+ * @prop --calcite-checkbox-color: Specifies the component's font color.
```
jcfranco, line 9: Can we revisit this description? It is used to set the `color` prop, but from a user's perspective, it's setting the check's color. 

@alisonailea Do we need to specify `icon` in the name or is this an exception because the component is mostly an icon?

alisonailea: we could use -checkbox-icon-color but I think -checkbox-color makes sense here but I'm open to either

aPreciado88: I'll leave it as `-checkbox-color`.

jcfranco: Let’s use `-checkbox-icon-color` for consistency. Since we already differentiate the border color, I think sticking with the icon color makes sense.

aPreciado88: I updated the token to `--calcite-checkbox-icon-color` 🚀

Issue #10227, opened 2024-09-05
InputTimePicker/InputTimeZone unstable test


driskull:
### Priority impact

p - high

### Test type

unstable

### Which Component(s)

`input-time-picker`, `input-time-zone`

### Unstable Tests

openClose("calcite-input-time-picker", { initialToggleValue: true });

### Test error, if applicable

Summary of all failing tests
@esri/calcite-components:test: FAIL src/components/input-time-picker/input-time-picker.e2e.ts (51.699 s)
@esri/calcite-components:test:   ● calcite-input-time-picker › openClose › initially open › emits on initialization with animations enabled
@esri/calcite-components:test: 
@esri/calcite-components:test:     expect(received).toBeGreaterThan(expected)
@esri/calcite-components:test: 
@esri/calcite-components:test:     Expected: > 100
@esri/calcite-components:test:     Received:   1
@esri/calcite-components:test: 
@esri/calcite-components:test:       205 |     const matcherName = animationsEnabled ? "toBeGreaterThan" : ("toBeLessThanOrEqual" as const);
@esri/calcite-components:test:       206 |
@esri/calcite-components:test:     > 207 |     expect(delayBetweenBeforeOpenAndOpen)[matcherName](delayDeltaThreshold);
@esri/calcite-components:test:           |                                                       ^
@esri/calcite-components:test:       208 |     expect(delayBetweenBeforeCloseAndClose)[matcherName](delayDeltaThreshold);
@esri/calcite-components:test:       209 |   }
@esri/calcite-components:test:       210 |
@esri/calcite-components:test: 
@esri/calcite-components:test:       at testOpenCloseEvents (src/tests/commonTests/openClose.ts:207:55)
@esri/calcite-components:test:       at Object.<anonymous> (src/tests/commonTests/openClose.ts:218:7)
@esri/calcite-components:test: 
@esri/calcite-components:test:

Same test failure for `input-time-zone`.

### PR skipped, if applicable

#10226, #10228

jcfranco:
Noticed a similar failure locally with https://github.com/Esri/calcite-design-system/pull/10228. Will update the issue title/description to investigate both under this issue.

jcfranco:
#10228 addressed the `input-time-zone` failure. Will look into the `input-time-picker` one.

jcfranco:
#10310 might help here.

geospatialem:
Spike needed to determine if the migration to Lit mitigates the failure.

[...]

PR #10228, opened 2024-09-05
fix(input-time-zone): ensure `beforeOpen`/`open` and `beforeClose`/`close` events emit properly

jcfranco:
**Related Issue:** #9315, #10227

## Summary

This addresses some issues that prevented `beforeOpen`/`open` and `beforeClose`/`close` events from emitting as expected.

### Notes

* a supporting helper for `openClose` was added to handle the programmatically opened and appended use case to simplify code and usage
* restores related, skipped tests
* `input-time-zone` was missed in https://github.com/Esri/calcite-design-system/pull/9958 (my bad! 😅)


packages/calcite-components/src/components/input-time-zone/input-time-zone.tsx
```diff
@@ -487,6 +493,7 @@ export class InputTimeZone
   componentDidLoad(): void {
     setComponentLoaded(this);
     this.overrideSelectedLabelForRegion(this.open);
+    this.openChanged();
```
driskull, line 496: For this, i'm wondering if its better to call this when there's a ref for the element rather than in componentDidLoad?

I'm not sure it makes a difference in most cases but it just seems more appropriate?

In this case it would be in `setComboboxRef`

driskull (approved):
👍

PR #10219, opened 2024-09-04
chore: set default page size for E2E tests

jcfranco:
**Related Issue:** N/A

## Summary

Sets default window size to help improve consistency between CI and local E2E test runs.

PR #10232, opened 2024-09-06
chore: remove unnecessary page size workaround for `disabled` test helper

jcfranco:
**Related Issue:** N/A

## Summary

Removes workaround invalidated by #10219.


jcfranco:
Tests are failing, so it seems that setting the window size with `browserArgs` doesn't have the same effect as `page.setViewport`.

PR #10239, opened 2024-09-06
refactor: only call getElementDir once when positioning floating-ui elements

driskull:
**Related Issue:** N/A

## Summary

chore: only call getElementDir once when positioning floating-ui elements


jcfranco (approved):
Noice! ✨🧹✨

Issue #10240, opened 2024-09-06
Closed floating elements should not cause scrollbars


driskull:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Closed floating elements cause scrollbars

### Expected Behavior

Closed floating elements should not cause scrollbars

### Reproduction Sample

https://codepen.io/driskull/pen/RwzEMYv?editors=1100

### Reproduction Steps

1. open the codepen
2. notice the scrollbars shown due to a closed popover

### Reproduction Version

2.12.2

### Relevant Info

https://devtopia.esri.com/WebGIS/arcgis-js-api/issues/63967

### Regression?

No

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

driskull:
@geospatialem @DitwanP moved this one to October.

iyouth23:
+1 from the AGOL Map Viewer team - we have multiple instances of this bug occurring on production, so the impact is high.

<details><summary>Demo</summary>

https://github.com/user-attachments/assets/e5879ed9-730c-496f-95ca-248bc615fd12

</details>

zaramatheson:
@driskull @geospatialem Any chance this one could get priortized for a 2.13 patch? It affects a few spots in MV.  We are investigating the workaround desccribed in the JSAPI issue, but it would have to be implemented in quite a fewplaces.

driskull:
@zaramatheson let me know if the workaround is working for you. I'll defer to @jcfranco and @geospatialem for when this can be installed.

iyouth23:
> @zaramatheson let me know if the workaround is working for you. I'll defer to @jcfranco and @geospatialem for when this can be installed.

@driskull I'm unable to apply the workaround to these Map Viewer cases, but I'm not sure if it's a case of our components being nested too deep within other components, or something else. The workaround for the API also seems most relevant for hidden elements, but our concern is more related to currently visible elements whose content overflows a specified max-height, which feels like a slightly different issue.

[...]

Issue #1381, opened 2020-12-08
Bug: Dropdown on RTL page causes long horizontal scroll


nwhittaker:
### Summary
The Dropdown component introduces a long horizontal scroll bar when placed on a page where the `dir="rtl"` attribute is set on the root element.

<img width="1504" alt="Screen Shot 2020-12-08 at 1 56 07 PM" src="https://user-images.githubusercontent.com/421496/101530078-70147f80-395f-11eb-9d7b-594c5d281d92.png">

The `.calcite-dropdown-wrapper` element uses a `left: -999999px` style to position itself offscreen. In RTL layouts, this causes the horizontal scroll. One fix could be to replace this with a `right: -999999px` style when in an RTL layout.

### Actual Behavior
A long horizontal scrollbar is present on the viewport.

### Expected Behavior
No horizontal scrollbar is present.

### Reproduction Steps
1. Visit https://esri.github.io/calcite-components/?path=/story/components-dropdown--simple-rtl.
2. Use the browser's developer tools to find the `#storybook-preview-iframe` element.
3. Open the iframe and add the `dir="rtl"` attribute to the root node.
4. Notice the horizontal scrollbar appears. Some browsers may flash it, others may need the iframe focused before showing it.

### Relevant Info 
Observed in latest versions of Chrome, Firefox, and Safari on latest macOS.

nwhittaker:
@driskull, we're seeing a regression with this fix when a dropdown trigger is aligned to the trailing/end edge of a scrollable container: https://github.com/ArcGIS/field-apps-web/issues/1336.

For a contrived example:

1. Visit https://esri.github.io/calcite-components/?path=/story/components-dropdown--simple.
2. If needed, give the `.calcite-dropdown-wrapper` element a `left: auto;` style to mimic the https://github.com/Esri/calcite-components/pull/1387 patch.
2. Give the `#root` element `width: 240px; height: 218px; text-align: right; overflow: scroll;` styles.

Before clicking the trigger, notice the #root element is horizontally scrollable. After clicking the trigger, the container is no longer horizontally scrollable.

Setting a `-999999px` `left`/`right` style on the `.calcite-dropdown-wrapper` element, depending on the `dir` value, does appear to be enough to fix the regression.

<img width="1284" alt="Screen Shot 2020-12-14 at 1 11 54 PM" src="https://user-images.githubusercontent.com/421496/102120133-16ed9580-3e10-11eb-9e66-c92fa2acaae8.png">

[...]

PR #10241, opened 2024-09-06
fix!: floating-ui elements no longer take up space when closed

driskull:
**Related Issue:** #10240

## Summary

- Sets floating element dimensions to zero when not open
  - This makes sure that a closed floating-ui element doesn't produce scrollbars within a parent element. 
  - This is handled by setting the style on the floating element container to `display:none` when not open.
- [Popover, Tooltip]: Moves position styles from the host to an internal element to avoid style clashes.
  - This is better because the positioning styles won't interfere with any light DOM styling
  - This added benefit is less CSS in the floating-ui CSS file. 
- `display:none` is removed in the `onClose` of the component so that the animation can finish before its set to be hidden.
  - This restores closing animation in floating ui elements.
- Cleanup floating-ui methods 
- Add story
- Update tests

Avoids closed floating ui elements from affecting layout changes. Example:

image


packages/calcite-components/src/utils/floating-ui.ts
```diff
@@ -406,6 +413,18 @@ export function getEffectivePlacement(floatingEl: HTMLElement, placement: Logica
   return placement.replace(/leading/gi, placements[0]).replace(/trailing/gi, placements[1]) as EffectivePlacement;
 }
 
+const hiddenFloatingElStyle = {
+  visibility: "hidden",
+  pointerEvents: "none",
+  // initial positioning based on https://floating-ui.com/docs/computePosition#initial-layout
```
driskull, line 419: This seems to be working fine. It doesn't seem to need `top` or `left`. We do have `inset` defined so maybe that's why. We really just want the element to be off the screen and not affecting any layout when its not open.

packages/calcite-components/src/utils/floating-ui.ts
```diff
@@ -406,6 +413,18 @@ export function getEffectivePlacement(floatingEl: HTMLElement, placement: Logica
...
+  position: "absolute",
+  transform: "",
+  top: "",
+  left: "",
+  insetInlineStart: "-999999px",
+  insetBlockStart: "-999999px",
```
driskull, line 425: using this will make sure its positioned off the screen no matter what page direction or viewport size. Therefore, it shouldn't cause any scrollbars or width change in any parent element.

packages/calcite-components/src/utils/floating-ui.ts
```diff
@@ -153,13 +153,20 @@ export const positionFloatingUI =
 
     const { open } = component;
 
+    if (!open) {
+      Object.assign(floatingEl.style, hiddenFloatingElStyle);
```
driskull, line 157: If for some reason the component is no longer open during a debounced reposition, we can assign the hidden styles and stop.

jcfranco:
Before reviewing, could you confirm the following issues are still handled by these changes?

https://github.com/Esri/calcite-design-system/issues/1381
https://github.com/Esri/calcite-design-system/issues/5694

I recall we moved away from -999...9px values in the past.

driskull:
#1381 should not be an issue since we are using `inset-inline` which will not cause scrollbars in RTL direction.

#5694 seems to be an issue with this PR so I guess that would be a showstopper.

The problem with not positioning off screen is that it will affect the size of the parent container.

Maybe once we implement the dialog element and popover attribute it will not be an issue any longer.

For now, I'll close this PR since it causes issues with drag and drop and elements positioned offscreen.

jcfranco:
Thanks for testing. Bummer about https://github.com/Esri/calcite-design-system/issues/5694. 😞

driskull:
If I set the `width` and `height` to `0` instead of positioning it offscreen that seems to do the trick. I'll do some more testing.

packages/calcite-components/src/utils/floating-ui.ts
```diff
@@ -433,10 +456,17 @@ export async function reposition(
...
+  if (!options.floatingEl || !options.referenceEl) {
     return;
   }
 
+  if (!component.open) {
+    Object.assign(options.floatingEl.style, hiddenFloatingElStyle);
+    return;
+  }
+
+  Object.assign(options.floatingEl.style, initialFloatingElStyle);
```
driskull, line 468: We assign the initial floating el styles before any positioning happens. This means the top and left are both zero and the position is absolute. The width and height of zero is removed.

packages/calcite-components/src/utils/floating-ui.spec.ts
```diff
@@ -78,21 +78,27 @@ describe("repositioning", () => {
     connectFloatingUI(fakeFloatingUiComponent, referenceEl, floatingEl);
   });
 
+  function assertClosedPositioning(floatingEl: HTMLElement): void {
+    expect(floatingEl.style.display).toBe("none");
```
jcfranco, line 82: For consistency, can you assert on other style props here too?

jcfranco (approved):
I had a few follow-up comments, but this is great, @driskull! ✨💪✨

packages/calcite-components/src/components/dropdown/dropdown.e2e.ts
```diff
@@ -1151,7 +1151,7 @@ describe("calcite-dropdown", () => {
     const page = await newE2EPage({
       html: html`<calcite-panel heading="Issue #3048">
         <calcite-pick-list filter-enabled>
-          <calcite-dropdown slot="menu-actions" placement="bottom-end" type="click">
+          <calcite-dropdown open slot="menu-actions" placement="bottom-end" type="click">
```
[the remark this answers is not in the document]
driskull: reverted

driskull:
@jcfranco @Elijbet with this approach, I'm seeing a lot of failures like this: `openClose › emits with animations enabled`

Do you know why this might be failing?

```
FAIL src/components/dropdown/dropdown.e2e.ts (81.361 s)
@esri/calcite-components:test:   ● calcite-dropdown › openClose › emits with animations enabled
@esri/calcite-components:test: 
@esri/calcite-components:test:     expect(received).toBeGreaterThan(expected)
@esri/calcite-components:test: 
@esri/calcite-components:test:     Expected: > 100
@esri/calcite-components:test:     Received:   2
@esri/calcite-components:test: 
@esri/calcite-components:test:       267 |
@esri/calcite-components:test:       268 |   expect(delayBetweenBeforeOpenAndOpen)[matcherName](delayDeltaThreshold);
@esri/calcite-components:test:     > 269 |   expect(delayBetweenBeforeCloseAndClose)[matcherName](delayDeltaThreshold);
@esri/calcite-components:test:           |                                                       ^
@esri/calcite-components:test:       270 | }
@esri/calcite-components:test:       271 |
@esri/calcite-components:test:       272 | type EventOrderWindow = GlobalTestProps<{ events: string[] }>;
@esri/calcite-components:test: 
@esri/calcite-components:test:       at testOpenCloseEvents (src/tests/commonTests/openClose.ts:269:55)
@esri/calcite-components:test:       at Object.<anonymous> (src/tests/commonTests/openClose.ts:73:5)
```

driskull:
We'll need to figure this out to move to the popover attribute/dialog element. Because the default browser styles use display:none.

```
[popover]:not(:popover-open):not(dialog[open]) {
    display: none;
}
```

jcfranco:
> Do you know why this might be failing?

I'll take a look this week. 🕵️

> We'll need to figure this out to move to the popover attribute/dialog element. Because the default browser styles use display:none.

Not sure I follow, could you clarify or add more context?

driskull:
> Not sure I follow, could you clarify or add more context?

Yes. Currently, our floating-ui components are set to `visibility:hidden` but not `display:none` when they are not open. This PR changes that so that they are set to `display:none`. This aligns with how the popover attribute and dialog element work. When they are not open, they are `display:none;`

packages/calcite-components/src/components/dropdown/dropdown.e2e.ts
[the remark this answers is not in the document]
driskull: Actually, this needs to be set to open now to actually get dimensions from getComputedStyle.

packages/calcite-components/src/components/dropdown/dropdown.e2e.ts
```diff
@@ -1175,7 +1176,7 @@ describe("calcite-dropdown", () => {
       filterInput.value = "numbers";
     });
 
-    expect(dropdownContentHeight.height).toBe("88px");
+    expect(dropdownContentHeight.height).toBe("72px");
```
driskull, line 1179: verified that 72px is correct.

packages/calcite-components/src/utils/floating-ui.spec.ts
driskull: done.

driskull:
@jcfranco only the `openClose` tests are failing now.

driskull:
@jcfranco figured out the issue with `openClose`. Should be good to review again tomorrow.

jcfranco (approved):
LGTM! 

Could you update the PR title and description to match our [breaking change formatting](https://github.com/Esri/calcite-design-system/blob/main/CONTRIBUTING.md#breaking-changes)? For the breaking change footer, feel free to use notes from our 3.x. breaking change doc.

This can be merged after the Oct milestone wraps up.

packages/calcite-components/src/components/shell/shell.stories.ts
```diff
@@ -1848,6 +1848,13 @@ export const panelWithPopoverZIndex = (): string =>
           </calcite-flow-item>
         </calcite-flow>
       </calcite-shell-panel> </calcite-shell
-    ><calcite-popover open reference-element="button" offset-distance="-50" offset-skidding="15" style="z-index: 100">
+    ><calcite-popover
+      overlay-positioning="fixed"
```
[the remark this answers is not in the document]
driskull: Its not needed, ill revert the change. The panel inside the popover just needs a width here. Ill update the test.

packages/calcite-components/src/components/popover/popover.e2e.ts
```diff
@@ -10,11 +10,12 @@ import {
...
   t9n,
 } from "../../tests/commonTests";
+import { skipAnimations } from "../../tests/utils";
 import { CSS } from "./resources";
 
 describe("calcite-popover", () => {
   describe("renders", () => {
-    renders("calcite-popover", { visible: false, display: "block" });
+    renders("calcite-popover", { display: "block" });
     renders(`<calcite-popover label="test" open reference-element="ref"></calcite-popover><div id="ref">😄</div>`, {
```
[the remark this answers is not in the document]
driskull: fixing

jcfranco:
@driskull I noticed this doesn't have a breaking change footer per [our conventions](https://github.com/Esri/calcite-design-system/blob/main/CONTRIBUTING.md#breaking-changes). Can you use commit overrides in the description to include it (see https://github.com/Esri/calcite-design-system/wiki/monorepo#edit-changelog-entries)?

driskull:
@jcfranco added a breaking change to the PR body. Let me know if its good now.

jcfranco:
@driskull Looking good! Can you revisit the breaking change note with migration advice for developers? You can use what we have in our [3.0 breaking changes doc](https://confluencewikidev.esri.com/pages/resumedraft.action?draftId=344949922&draftShareId=a132b33c-806f-4a64-b4e5-922e007347c6&) (migration plan note content):

> **Migration plan**: Developers will need to make sure any floating-ui components within a scrolling container use overlay-positioning=fixed.

Also, should we add a note to apply any dimension-altering styles the content instead of on the host?

@benelan Could you review the override comment? Not sure if the override needs to include the whole description or just the title part. Once confirmed, we should update the monorepo wiki accordingly.

driskull:
Updated. Feel free to adjust the formatting or wording.

benelan:
> @benelan Could you review the override comment? Not sure if the override needs to include the whole description or just the title part. Once confirmed, we should update the monorepo wiki accordingly.

Normally it's just the title, but in this case I'm not sure if the `BREAKING CHANGE` footer needs to go in there too. We can see what the release-please PR does with the existing override cpntent and add the footer if it doesn't work as is.

Issue #10238, opened 2024-09-06
[Panel] When the component has scrollable content and is in focus displays a focus ring


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When `panel` contains scrollable content and the component is in focus, the component displays a focus.

![regression-demo](https://github.com/user-attachments/assets/3c1c6a06-3d63-4862-9995-478f6422b7c6)

### Expected Behavior

No focus

![before-regression](https://github.com/user-attachments/assets/0ba4e73a-a188-4334-a636-82ae5189d3ff)

### Reproduction Sample

https://codepen.io/jchenAtlas/pen/GRbPZJE

### Reproduction Steps

1. Open the [sample](https://codepen.io/jchenAtlas/pen/GRbPZJE)
2. Scroll the content (by pressing and moving the scrollbar) and observe the focus outline + behavior
3. Change the version to `2.11.0` and observe no focus outline

### Reproduction Version

2.12.1

### Relevant Info

Related: #10022, #10154

### Regression?

2.12.0

### Priority impact

impact - p1 - need for current milestone

### Impact

High

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

PR #10242, opened 2024-09-06
fix(panel, flow-item): hide focus-outline when scrolling via mouse to align with browsers

jcfranco:
**Related Issue:** #10238

## Summary

Fixes a regression caused by #10141 where clicks on scrollers would show the focus outline. 

### Notes

* this behavior aligns with Chrome/Firefox's latest default focus on scrolling content behavior
* `:focus-visible` [isn’t supported in Safari 15.0 - 15.3](https://caniuse.com/?search=focus-visible), but given Safari 18 is expected this year, this is a minor visual inconsistency for unsupported versions.
* cleans up focus styles that should have been updated by #10141.
* no tests were added as we currently don't cover interactive focus-related styling

### DIFF
--- a/packages/calcite-components/src/components/panel/panel.scss
+++ b/packages/calcite-components/src/components/panel/panel.scss
@@ -73,15 +73,11 @@
 }
 
 .container {
-  @apply focus-base relative bg-background m-0 flex w-full flex-auto flex-col items-stretch p-0;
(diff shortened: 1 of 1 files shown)


driskull (approved):
👍

alisonailea (approved):

PR #10247, opened 2024-09-09
refactor: resolve minor issues

maxpatiiuk:
A few minor code changes for things I noticed while running the codemod

### DIFF
--- a/packages/calcite-components/src/components/accordion-item/accordion-item.tsx
+++ b/packages/calcite-components/src/components/accordion-item/accordion-item.tsx
@@ -289,9 +289,9 @@ export class AccordionItem implements ConditionalSlotComponent, LoadableComponen
       return;
     }
 
-    accordionItem.iconPosition = closestAccordionParent.iconPosition;
-    accordionItem.iconType = closestAccordionParent.iconType;
-    accordionItem.scale = closestAccordionParent.scale;
+    this.iconPosition = closestAccordionParent.iconPosition;
+    this.iconType = closestAccordionParent.iconType;
+    this.scale = closestAccordionParent.scale;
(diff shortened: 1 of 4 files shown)


packages/calcite-components/src/components/accordion-item/accordion-item.tsx
```diff
@@ -289,9 +289,9 @@ export class AccordionItem implements ConditionalSlotComponent, LoadableComponen
       return;
     }
 
-    accordionItem.iconPosition = closestAccordionParent.iconPosition;
-    accordionItem.iconType = closestAccordionParent.iconType;
-    accordionItem.scale = closestAccordionParent.scale;
+    this.iconPosition = closestAccordionParent.iconPosition;
```
maxpatiiuk, line 292: while both are valid, in Lit, to get better typing inside the component you should be settings things on `this` rather than `this.el` to get better typing

packages/calcite-components/src/components/graph/graph.e2e.ts
```diff
@@ -45,9 +45,8 @@ describe("calcite-graph", () => {
   });
 
   it("draws an area graph", async () => {
-    const dimensionsStyle = `style="height:100px; width:300px;"`;
     const page = await newE2EPage();
-    await page.setContent(`<calcite-graph ${dimensionsStyle}></calcite-graph>`);
+    await page.setContent(`<calcite-graph style="height:100px; width:300px;"></calcite-graph>`);
```
maxpatiiuk, line 49: this change would simplify work for the html string to JSX codemod

jcfranco (approved):
Thanks for the tidying up, @maxpatiiuk!

Issue #7516, opened 2023-08-14
Calcite Block header-menu-actions popover style/direction control


RSantosGIS:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

I would like to have the same options for positioning the Block header-menu-actions overflow popover/dropdown menu that I do for the Calcite Dropdown
- overlay positioning, to allow me to overflow the parent container
- placement (with extended options), to allow me to position it in the start/end of the left/right/top/bottom

You can see how this doesn't feel great below:

Collapsed Behavior:
![image](https://github.com/Esri/calcite-design-system/assets/36513418/93108cb6-9764-4c6c-90db-ab99127dda49)

Expanded Behavior:
![image](https://github.com/Esri/calcite-design-system/assets/36513418/7cb190d7-7c03-4aa3-89b5-dd29d53dbf89)

Desired Behavior (achieved with a Calcite list item under the block using a dropdown, since our sublayers don't need to be expanded further):
![image](https://github.com/Esri/calcite-design-system/assets/36513418/67ac4dec-9630-4ba4-a1a8-001265bc086f)

### Acceptance Criteria

Block header-menu-actions has same placement options as dropdown

### Which Component

Calcite Block

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [X] @esri/calcite-components-react

### Esri team

ArcGIS Knowledge

jcfranco:
Thanks for submitting this. 

Could you clarify which component you're using in your screenshots? Block doesn't have a menu.

RSantosGIS:
Just Calcite Block with Calcite actions that are in the blocks header-menu-actions slot:

![image](https://github.com/Esri/calcite-design-system/assets/36513418/f1d46e76-79c5-4172-938f-00520a9cea06)

geospatialem:
Spike needed to determine which components should accommodate placement, such as `block`, `panel`, `action-group`, etc.

driskull:
Seems like we should just add `placement` and `flipPlacements` properties to all three of these components.

DitwanP:
🍭 Verified locally on `dev`

One example below, `menuPlacement`: bottom & `menuFlipPlacements`: ["left"]

https://github.com/user-attachments/assets/1d5c0db4-5bc0-4d59-b64a-f43b4dabdbe1

PR #10249, opened 2024-09-09
feat(action-group, block, panel): add `menuPlacement` and `menuFlipPlacements` properties

driskull:
**Related Issue:** #7516

## Summary

- add `menuFlipPlacements` property
- add `menuPlacement` property
- add tests
- update stories

### DIFF
--- a/packages/calcite-components/src/components/action-group/action-group.tsx
+++ b/packages/calcite-components/src/components/action-group/action-group.tsx
@@ -21,7 +21,7 @@ import {
 } from "../../utils/t9n";
 import { SLOTS as ACTION_MENU_SLOTS } from "../action-menu/resources";
 import { Layout, Scale } from "../interfaces";
-import { OverlayPositioning } from "../../utils/floating-ui";
+import { FlipPlacement, LogicalPlacement, OverlayPositioning } from "../../utils/floating-ui";
 import { slotChangeHasAssignedElement } from "../../utils/dom";
 import { Columns } from "./interfaces";
 import { ActionGroupMessages } from "./assets/action-group/t9n";
@@ -60,6 +60,11 @@ export class ActionGroup
   }
 
   /**
+   * Defines the available placements that can be used when a flip occurs.
+   */
+  @Prop() flipPlacements: FlipPlacement[];
+
+  /**
    *  Accessible name for the component.
    */
   @Prop() label: string;
@@ -89,6 +94,11 @@ export class ActionGroup
    *
    */
   @Prop({ reflect: true }) overlayPositioning: OverlayPositioning = "absolute";
+
+  /**
+   * Determines where the action menu will be positioned.
+   */
+  @Prop({ reflect: true }) placement: LogicalPlacement;
 
   /**
    * Specifies the size of the `calcite-action-menu`.
@@ -178,19 +188,30 @@ export class ActionGroup
   // --------------------------------------------------------------------------
 
   renderMenu(): VNode {
-    const { expanded, menuOpen, scale, layout, messages, overlayPositioning, hasMenuActions } =
-      this;
+    const {
+      expanded,
+      menuOpen,
+      scale,
+      layout,
+      messages,
+      overlayPositioning,
+      hasMenuActions,
+      flipPlacements,
+      placement,
+    } = this;
 
     return (
       <calcite-action-menu
         expanded={expanded}
-        flipPlacements={["left", "right"]}
+        flipPlacements={
+          flipPlacements ?? (layout === "horizontal" ? ["top", "bottom"] : ["left", "right"])
+        }
         hidden={!hasMenuActions}
         label={messages.more}
         onCalciteActionMenuOpen={this.setMenuOpen}
         open={menuOpen}
         overlayPositioning={overlayPositioning}
-        placement={layout === "horizontal" ? "bottom-start" : "leading-start"}
+        placement={placement ?? (layout === "horizontal" ? "bottom-start" : "leading-start")}
         scale={scale}
       >
         <calcite-action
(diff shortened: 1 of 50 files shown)


packages/calcite-components/src/components/block/block.tsx
```diff
@@ -94,6 +94,11 @@ export class Block
    */
   @Prop({ reflect: true }) dragHandle = false;
 
+  /**
+   * Defines the available placements that can be used when a flip occurs.
```
macandcheese, line 98: Do we want a bit more context here for "a flip occurs" - maybe something about the rendered action menu flipping? It makes sense to me as someone who knows it refers to internally rendered action menu, but maybe these could be confusing without a more descriptive description (or even prop name)? Applies to others.

driskull: Yeah probably. I copied this from another component.

cc @geospatialem @DitwanP on doc

macandcheese: I guess my comment was more about _what_ is being affected by this. In this case, it's "the action menu that is rendered when you slot actions in ___"

"... the component's fallback `placement`" still isn't clear to a user without context - maybe "Specifies the fallback `placement` of the component's (internally rendered Action Menu that displays when slotted actions overflow)" (but ... less bad verbiage)

driskull: @geospatialem combobox is out of scope I think. It already has flipPlacements just not placement.

Popover, input-date-picker, split-button, and dropdown already has these properties.

geospatialem: Added a new PR https://github.com/Esri/calcite-design-system/pull/10261 to cover the remaining components for consistency.

packages/calcite-components/src/components/block/resources.ts
```diff
@@ -43,3 +43,5 @@ export const ICONS = {
   valid: "check-circle",
   invalid: "exclamation-mark-triangle",
 } as const;
+
+export const defaultMenuPlacement = "bottom-end";
```
jcfranco, line 47: There’s a [`defaultMenuPlacement` in `floating-ui.ts`](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/utils/floating-ui.ts#L273) as well. How should we decide when to use it vs creating a separate one?

driskull: They are both positioned bottom but for dropdown/combobox bottom-start makes more sense whereas for action-menu's mostly positioned at the end bottom-end makes more sense. 

I can put this const in floating-ui.ts

driskull:

packages/calcite-components/src/components/action-group/action-group.tsx
```diff
@@ -95,6 +95,16 @@ export class ActionGroup
...
 
+  /**
+   * Specifies the component's fallback menu `placement` when it's initial or specified `placement` has insufficient space available.
+   */
+  @Prop() menuFlipPlacements: FlipPlacement[];
+
+  /**
+   * Determines where the action menu will be positioned.
+   */
+  @Prop({ reflect: true }) menuPlacement: LogicalPlacement;
```
driskull, line 106: Do we prefer `actionMenuPlacement` or is `menuPlacement` sufficient?

jcfranco: I think the current name is clear and concise.

packages/calcite-components/src/tests/commonTests/floatingUI.ts
```diff
@@ -132,3 +132,38 @@ export async function delegatesToFloatingUiOwningComponent(
     expect(await floatingUiOwningComponent.getProperty("overlayPositioning")).toBe("fixed");
   });
 }
+
+/**
+ * Helper to test if a component has a calcite-action-menu wired up correctly with placement and flipPlacements.
```
jcfranco, line 137: ✨🏆✨ This is great, thanks for adding it!

packages/calcite-components/src/components/block/block.tsx
jcfranco: For consistency and readability, would we want to consider updating similar properties to use the `menu` prefix? I don't recall if we have properties other than `overlayPositioning`? 🤔

jcfranco (approved):
💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭
💭💭↔️↔️↔️💭↔️💭💭💭💭💭↔️💭↔️↔️↔️↔️💭↔️↔️↔️↔️💭↔️↔️↔️↔️↔️💭↔️💭
💭↔️💭💭💭💭↔️💭💭💭💭💭↔️💭↔️💭💭💭💭↔️💭💭💭💭💭💭↔️💭💭💭↔️💭
💭💭↔️↔️💭💭↔️💭💭↔️💭💭↔️💭↔️↔️↔️💭💭↔️↔️↔️💭💭💭💭↔️💭💭💭↔️💭
💭💭💭💭↔️💭↔️💭↔️💭↔️💭↔️💭↔️💭💭💭💭↔️💭💭💭💭💭💭↔️💭💭💭💭💭
💭↔️↔️↔️💭💭💭↔️💭💭💭↔️💭💭↔️↔️↔️↔️💭↔️↔️↔️↔️💭💭💭↔️💭💭💭↔️💭
💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭💭

driskull:
> For consistency and readability, would we want to consider updating similar properties to use the menu prefix? I don't recall if we have properties other than overlayPositioning

Yes, potentially. This would be a breaking change though. 🔧 

Should we have a followup issue? cc @DitwanP @geospatialem

DitwanP:
> Should we have a followup issue? cc @DitwanP @geospatialem

[Follow-up Issue](https://github.com/Esri/calcite-design-system/issues/10391)

PR #10253, opened 2024-09-09
feat(popover, action): add component tokens

Elijbet:
**Related Issue:** #7180

## Summary
Add `popover` component tokens.

`--calcite-popover-background-color`: Specifies the background color of the component.
`--calcite-popover-border-color`: Specifies the border color of the component.
`--calcite-popover-corner-radius`: Specifies the corner radius of the component.
`--calcite-popover-text-color`: Specifies the text color of the component.

Add `action` component tokens.

`--calcite-action-corner-radius`: Specifies the component's corner radius.
`--calcite-action-corner-radius-end-end`: Specifies the component's corner radius end end.
`--calcite-action-corner-radius-end-start`: Specifies the component's corner radius end start.
`--calcite-action-corner-radius-start-end`: Specifies the component's corner radius start end.
`--calcite-action-corner-radius-start-start`: Specifies the component's corner radius start start.


alisonailea:
This is looking good Eliza! Thanks!

Probably outside this PR but @macandcheese @SkyeSeitz @ashetland @jcfranco per the discussion on [accordion-item header](https://github.com/Esri/calcite-design-system/pull/10181), I'm wondering if there is a larger ui pattern we should be setting here allowing component container header backgrounds to be styled separately from the body?

packages/calcite-components/src/components/popover/popover.e2e.ts
```diff
@@ -699,4 +700,64 @@ describe("calcite-popover", () => {
...
+        "--calcite-popover-background-color": {
+          shadowSelector: `.${CSS.calciteFloatingUIAnim}`,
+          targetProp: "backgroundColor",
+        },
+        "--calcite-popover-border-color": [
+          {
+            shadowSelector: `.${CSS.calciteFloatingUIAnim}`,
+            targetProp: "borderColor",
+          },
+          // { shadowSelector: ".arrow::before", targetProp: "outline" },
```
Elijbet, line 722: Having hard time targeting this. It's a class on FloatingArrow component within Popover.

jcfranco: `.arrow` should be `.calcite-floating-ui-arrow`. It should've gotten renamed in https://github.com/Esri/calcite-design-system/pull/6847/files#diff-2c88956271241eaf8204f4d12134f4fff58f10aeac5dc0dd8e18ad9e21f37607.

@driskull Should the `::before` selector be moved into `_floating-ui.scss` or should it stay in `popover.scss`?

packages/calcite-components/src/components/popover/popover.e2e.ts
```diff
@@ -699,4 +700,64 @@ describe("calcite-popover", () => {
...
+            targetProp: "color",
+          },
+          {
+            shadowSelector: `.${CSS.container}`,
+            targetProp: "color",
+          },
+        ],
+      });
+    });
+    describe("close-button", () => {
```
alisonailea, line 744: Can we call this "closable"?

packages/calcite-components/src/components/popover/resources.ts
```diff
@@ -1,4 +1,5 @@
 export const CSS = {
+  calciteFloatingUIAnim: "calcite-floating-ui-anim",
```
alisonailea, line 2: Shouldn't this be imported from a shared floating-ui resource file? @jcfranco @driskull

alisonailea (approved):

packages/calcite-components/src/components/popover/resources.ts
jcfranco: Yes, from https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/utils/floating-ui.ts#L319-L322.

jcfranco:
> Probably outside this PR but @macandcheese @SkyeSeitz @ashetland @jcfranco per the discussion on https://github.com/Esri/calcite-design-system/pull/10181, I'm wondering if there is a larger ui pattern we should be setting here allowing component container header backgrounds to be styled separately from the body?

Outside the scope, but I agree it might be helpful to explore this pattern. I'll defer to the rest of the group on this.

macandcheese:
> > Probably outside this PR but @macandcheese @SkyeSeitz @ashetland @jcfranco per the discussion on #10181, I'm wondering if there is a larger ui pattern we should be setting here allowing component container header backgrounds to be styled separately from the body?
> 
> Outside the scope, but I agree it might be helpful to explore this pattern. I'll defer to the rest of the group on this.

Agreed - many components will require multiple regions of styling - Panel (and consuming components) has "content", "header", "footer", "content-top/bottom", etc. that may all be styled separately depending on needs.

packages/calcite-components/src/components/popover/popover.e2e.ts
driskull: Yeah probably. It seems like this class is only on tooltip and popover and they are the only components that render an arrow. So it probably needs to move.

packages/calcite-components/src/components/popover/popover.stories.ts
```diff
@@ -186,25 +186,3 @@ export const largeScaleLayout_TestOnly = (): string => html`
     </calcite-popover>
   </div>
 `;
-
-export const transparentBG_TestOnly = (): string => html`
```
jcfranco, line 190: We'll need to restore this one since it added coverage for https://github.com/Esri/calcite-design-system/issues/6798.

packages/calcite-components/src/components/popover/popover.e2e.ts
```diff
@@ -699,4 +701,63 @@ describe("calcite-popover", () => {
...
+
+  describe("theme", () => {
+    const popoverHTML = html`
+      <calcite-popover heading="I'm a heading in the header using the 'heading' prop!" closable>
+        Lorem Ipsum
+      </calcite-popover>
+    `;
+
+    describe("default", () => {
+      themed(popoverHTML, {
```
jcfranco, line 713: Can you inline the popover HTML? It’s used for two tests with slight differences in each, where the default configuration should not be closable.

packages/calcite-components/src/components/popover/popover.e2e.ts
```diff
@@ -699,4 +701,63 @@ describe("calcite-popover", () => {
...
+          {
+            shadowSelector: `.${CSS.closeButtonContainer}`,
+            targetProp: "borderStartEndRadius",
+          },
+          {
+            shadowSelector: `.${CSS.closeButtonContainer}`,
+            targetProp: "borderEndEndRadius",
+          },
+          {
+            shadowSelector: `.${CSS.hasHeader} .${CSS.closeButtonContainer}`,
```
jcfranco, line 756: Isn't this testing the same as above?

packages/calcite-components/src/custom-theme.stories.ts
```diff
@@ -104,7 +105,9 @@ const kitchenSink = (args: Record<string, string>, useTestValues = false) =>
           <div>${checkbox}</div>
           ${chips} ${pagination} ${slider}
         </div>
-        <div class="demo-column">${datePicker} ${tabs} ${loader} ${calciteSwitch}</div>
+        <div class="demo-column">${datePicker} ${tabs} ${loader} ${calciteSwitch} 
+        <div style="margin-top: 40px">${popover}</div>
```
jcfranco, line 109: The comment from https://github.com/Esri/calcite-design-system/pull/10267/files#r1753091966 applies here.

packages/calcite-components/src/custom-theme.stories.ts
```diff
@@ -119,14 +122,15 @@ export default {
...
+    ...popoverTokens,
   },
 };
 
 export const themingInteractive = (args: Record<string, string>): string => {
   return kitchenSink(args);
 };
 
-export const theming_TestOnly = (): string => {
+export const theming = (): string => {
```
jcfranco, line 133: ✨🧹✨

jcfranco (changes requested):
Looking great, @Elijbet! Just a few more tweaks and this should be good to merge.

packages/calcite-components/src/components/popover/popover.stories.ts
Elijbet: @alisonailea Are we ok restoring this?

alisonailea: It's covered by theming tests isn't it?

jcfranco (approved):
🚀

packages/calcite-components/src/components/popover/popover.scss
```diff
@@ -3,6 +3,10 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
+ * @prop --calcite-popover-background-color: Specifies the background color of the component.
+ * @prop --calcite-popover-border-color: Specifies the border color of the component.
+ * @prop --calcite-popover-corner-radius: Specifies the corner radius of the component.
+ * @prop --calcite-popover-text-color: Specifies the text color of the component.
  * @prop --calcite-popover-z-index: Sets the z-index value for the component.
  */
```
macandcheese, line 11: I might expect a `--calcite-popover-shadow` css property here.

Elijbet: I don't see any shadow related styling on the component. Can you expand on that?

macandcheese: It's coming from floating ui - but I don't think a user would need to know that, so maybe a pattern q of how to set that up. As a user I'd want to just set it at the component level like the other adjustments to border color, corner radius, etc.

alisonailea: let's leave the floating-ui tokens seperatly.

packages/calcite-components/src/components/action/action.scss
```diff
@@ -7,6 +7,8 @@
  * @prop --calcite-action-background-color: Specifies the component's  background color.
  * @prop --calcite-action-background-color-hover: Specifies the component's background color when hovered or focused.
  * @prop --calcite-action-background-color-pressed: Specifies the component's background color when active.
+ * @prop --calcite-action-corner-radius-start-end: Specifies the component's corner radius start end.
+ * @prop --calcite-action-corner-radius-end-end: Specifies the component's corner radius end end.
```
Elijbet, line 11: @alisonailea could you chime in on this bit suggested by @jcfranco:

I think we'd want the following since Action is used across different components and in different contexts:
 
`--calcite-action-corner-radius` – for most, simple cases
`--calcite-action-corner-radius-<start|end>-<start|end>` – for finer-grained cases (e.g., close buttons)

For the latter, we could keep them internal in case these are only needed for close buttons and consider adding a close button component.

packages/calcite-components/src/components/action/action.scss
```diff
@@ -46,6 +77,35 @@
 
   background-color: var(--calcite-action-background-color, var(--calcite-color-foreground-1));
   color: var(--calcite-action-text-color, var(--calcite-color-text-3));
+
+  border-start-start-radius: var(
```
jcfranco, line 81: There's some duplication here. Can you DRY it up by adding a rule for these props that targets both `:host` and `.button`?

packages/calcite-components/src/components/popover/popover.scss
```diff
@@ -105,17 +109,33 @@
...
+    var(--calcite-corner-radius-sharp)
+  );
 }
 
 ::slotted(calcite-panel),
 ::slotted(calcite-flow) {
   @apply h-full;
 }
 
+calcite-action {
```
jcfranco, line 136: With the new props, this shouldn't be necessary.

packages/calcite-components/src/components/popover/popover.scss
```diff
@@ -105,17 +109,33 @@
...
+    var(--calcite-corner-radius-sharp)
+  );
 }
 
 .has-header .close-button-container {
-  border-end-end-radius: none;
+  border-end-end-radius: var(--calcite-popover-corner-radius, none);
+}
+
+:not(.has-header) .close-button-container {
```
jcfranco, line 124: Aren't this and the above targeting the same element?

packages/calcite-components/src/custom-theme/handle.ts
```diff
@@ -9,4 +9,9 @@ export const handleTokens = {
   calciteHandleIconColorSelected: "",
 };
 
-export const handle = html`<calcite-handle></calcite-handle>`;
+export const handle = html`
+  <calcite-label layout="inline">
```
jcfranco, line 13: Is this label necessary? `handle` has its host set to `display: flex`.

packages/calcite-components/src/custom-theme/handle.ts
```diff
@@ -9,4 +9,9 @@ export const handleTokens = {
...
+    <calcite-handle></calcite-handle>
+    <calcite-label> </calcite-label
```
jcfranco, line 15: This label seems superfluous.

packages/calcite-components/src/custom-theme/popover.ts
```diff
@@ -0,0 +1,28 @@
+import { html } from "../../support/formatting";
+
+export const popoverTokens = {
+  calcitePopoverBackgroundColor: "",
+  calcitePopoverBorderColor: "",
+  calcitePopoverCornerRadius: "",
+  calcitePopoverTextColor: "",
+  calcitePopoverFloatingUIZIndex: "",
```
jcfranco, line 8: This name does not match the CSS prop.

packages/calcite-components/src/components/popover/popover.scss
```diff
@@ -3,6 +3,10 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
+ * @prop --calcite-popover-background-color: Specifies the background color of the component.
+ * @prop --calcite-popover-border-color: Specifies the border color of the component.
+ * @prop --calcite-popover-corner-radius: Specifies the corner radius of the component.
+ * @prop --calcite-popover-text-color: Specifies the text color of the component.
  * @prop --calcite-popover-z-index: Sets the z-index value for the component.
```
jcfranco, line 10: Sidebar: can we deprecate component-specific z-indices? We should be using [the semantic z-indices](https://developers.arcgis.com/calcite-design-system/foundations/tokens/reference/#stacking) instead to adjust different layers. cc @alisonailea @driskull @macandcheese

packages/calcite-components/src/components/popover/popover.e2e.ts
```diff
@@ -699,4 +701,67 @@ describe("calcite-popover", () => {
...
+    describe("default", () => {
+      themed(
+        html`
+          <calcite-popover heading="I'm a heading in the header using the 'heading' prop!">
+            Lorem Ipsum
+          </calcite-popover>
+        `,
+        {
+          "--calcite-popover-background-color": {
+            shadowSelector: `.${FloatingCSS.animation}`,
```
jcfranco, line 715: Sidebar: for consistency, can we add a `.container` class to this element and use that in the test instead?

packages/calcite-components/src/components/action/action.scss
```diff
@@ -7,13 +7,44 @@
  * @prop --calcite-action-background-color: Specifies the component's  background color.
  * @prop --calcite-action-background-color-hover: Specifies the component's background color when hovered or focused.
  * @prop --calcite-action-background-color-pressed: Specifies the component's background color when active.
+ * @prop --calcite-action-corner-radius-start-end: Specifies the component's corner radius start end.
```
jcfranco, line 10: This is missing corner-radius properties used below.

jcfranco (changes requested):
This is really close, @Elijbet! One more round and I think this will be good to land.

packages/calcite-components/src/components/popover/popover.e2e.ts
jcfranco: Oh, I didn't spot that. Thanks for clarifying. We should look at renaming the lower-level one since it's inconsistent with the way we use `.container` in other components.

packages/calcite-components/src/components/action/action.scss
```diff
@@ -7,15 +7,56 @@
...
  * @prop --calcite-action-background-color-hover: Specifies the component's background color when hovered or focused.
  * @prop --calcite-action-background-color-pressed: Specifies the component's background color when active.
+ * @prop --calcite-action-corner-radius: Specifies the component's corner radius.
+ * @prop --calcite-action-corner-radius-end-end: Specifies the component's corner radius end end.
+ * @prop --calcite-action-corner-radius-end-start: Specifies the component's corner radius end start.
+ * @prop --calcite-action-corner-radius-start-end: Specifies the component's corner radius start end.
+ * @prop --calcite-action-corner-radius-start-start: Specifies the component's corner radius start start.
  * @prop --calcite-action-text-color: Specifies the text color of the component.
  * @prop --calcite-action-text-color-pressed: Specifies the component's text color when hovered.
+ * @prop --calcite-internal-action-corner-radius-end-end:: Specifies the component's internal corner radius end end.
```
jcfranco, line 17: Internal props do not need to be documented.

packages/calcite-components/src/components/action/action.scss
```diff
@@ -7,15 +7,56 @@
  * @prop --calcite-action-background-color: Specifies the component's  background color.
  * @prop --calcite-action-background-color-hover: Specifies the component's background color when hovered or focused.
  * @prop --calcite-action-background-color-pressed: Specifies the component's background color when active.
+ * @prop --calcite-action-corner-radius: Specifies the component's corner radius.
+ * @prop --calcite-action-corner-radius-end-end: Specifies the component's corner radius end end.
+ * @prop --calcite-action-corner-radius-end-start: Specifies the component's corner radius end start.
+ * @prop --calcite-action-corner-radius-start-end: Specifies the component's corner radius start end.
```
jcfranco, line 13: Can you add update action-related tests to cover these new props?

packages/calcite-components/src/components/popover/popover.tsx
```diff
@@ -589,7 +590,7 @@ export class Popover
           <div
             class={{
               [CSS.hasHeader]: !!heading,
-              [CSS.container]: true,
+              [CSS.headerContainer]: true,
```
jcfranco, line 593: This class was modified, but previous `.container` selectors have not been updated. Is this intentional?

Elijbet: No, apologies. I modified the test selectors, but haven't yet worked on the css.

packages/calcite-components/src/components/popover/popover.scss
```diff
@@ -105,12 +109,16 @@
...
   flex: 0 0 auto;
-  border-start-end-radius: theme("borderRadius.DEFAULT");
-  border-end-end-radius: theme("borderRadius.DEFAULT");
-}
-
-.has-header .close-button-container {
-  border-end-end-radius: none;
+  border-start-end-radius: var(--calcite-popover-corner-radius, var(--calcite-corner-radius-round));
+  border-end-end-radius: var(--calcite-popover-corner-radius, var(--calcite-corner-radius-round));
+  --calcite-internal-action-corner-radius-start-end: var(
```
jcfranco, line 114: Shouldn't this be setting the public action CSS props?

packages/calcite-components/src/components/action/action.scss
```diff
@@ -16,6 +21,38 @@
   @apply flex bg-transparent;
 }
 
+:host,
+button {
+  border-end-end-radius: var(
+    --calcite-action-corner-radius-end-end,
+    var(
+      --calcite-action-corner-radius,
+      var(--calcite-internal-action-corner-radius-end-end, var(--calcite-corner-radius))
```
jcfranco, line 30: Should the internal CSS prop be defined in this file?

packages/calcite-components/.stylelintrc.cjs
```diff
@@ -3,7 +3,10 @@
 // ⚠️ AUTO-GENERATED CODE - DO NOT EDIT
 const customFunctions = [
   "get-trailing-text-input-padding",
-  "scale-duration"
+  "medium-modular-scale",
```
jcfranco, line 6: Per our conversation, I'll look at why these were added since `small/medium-modular-scale` functions are not defined in `dev` or in this PR (maybe from `node_modules`? 🤔).

jcfranco (approved):
✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨
✨🚀🚀🚀🚀✨🚀✨✨✨✨✨🚀✨🚀🚀🚀🚀✨✨🚀🚀🚀✨✨🚀🚀✨✨🚀✨✨✨🚀✨🚀🚀🚀🚀✨🚀✨
✨🚀✨✨🚀✨🚀✨✨✨✨✨🚀✨🚀✨✨✨✨🚀✨✨✨✨🚀✨✨🚀✨🚀🚀✨🚀🚀✨🚀✨✨✨✨🚀✨
✨🚀🚀🚀🚀✨🚀✨✨🚀✨✨🚀✨🚀🚀🚀✨✨✨🚀🚀✨✨🚀✨✨🚀✨🚀✨🚀✨🚀✨🚀🚀🚀✨✨🚀✨
✨🚀✨✨🚀✨🚀✨🚀✨🚀✨🚀✨🚀✨✨✨✨✨✨✨🚀✨🚀✨✨🚀✨🚀✨✨✨🚀✨🚀✨✨✨✨✨✨
✨🚀✨✨🚀✨✨🚀✨✨✨🚀✨✨🚀🚀🚀🚀✨🚀🚀🚀✨✨✨🚀🚀✨✨🚀✨✨✨🚀✨🚀🚀🚀🚀✨🚀✨
✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨

Issue #8205, opened 2023-11-17
dropdown: the `arrow-down`/`arrow-up` keys should open a dropdown


driskull:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Summary

dropdown: the `arrow-down`/`arrow-up` keys should open a dropdown

### Actual Behavior

Arrow keys have no effect.

### Expected Behavior

Using the arrow keys on a trigger should open the dropdown component.

### Reproduction Sample

https://codepen.io/driskull/pen/RwvxBZG?editors=1000

### Reproduction Steps

1. focus on the "plus" action
2. Use arrow down or arrow up key
3. Nothing happens

### Reproduction Version

1.10

### Working W3C Example/Tutorial

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role#keyboard_interactions

### Relevant Info

![image](https://github.com/Esri/calcite-design-system/assets/1231455/7f0629c2-7eeb-4aae-a6d0-55d16a2465b5)

### Priority impact

p2 - want for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

jcfranco:
We should look at other components with the same role to enhance similarly. cc @geospatialem

jcfranco:
Looks like it's just `menu-item`. The internal `action-menu` already supports this. 🎉

DitwanP:
🍡 Verified locally on `dev`


https://github.com/user-attachments/assets/dcd14147-191d-4aa5-9dc9-314aa759b036

What's shown in the recording is me using the up and down arrows to open the dropdown.

PR #7613, opened 2023-08-28
docs: event handling guidelines - programmatic vs user-initiated actions event story

Elijbet:
**Related Issue:** #7221

## Summary

Expanding documentation regarding event handling guidelines in general. Kickstarting with programmatic vs. user-initiated actions event story, prompted by questions regarding emitting associated events for user-initiated actions here: https://github.com/Esri/calcite-design-system/issues/7221#issuecomment-1694026202

PR #10264, opened 2024-09-10
fix(dropdown): open dropdown on `ArrowDown` & `ArrowUp` keys

anveshmekala:
**Related Issue:** #8205 

## Summary

Open dropdown and focus first item on `ArrowDown`
Open dropdown and focus last item on `ArrowUp`

If any of the items is `selected` , both `ArrowDown` & `ArrowUp` keys focus the selected item.


packages/calcite-components/src/components/dropdown/dropdown.tsx
```diff
@@ -634,29 +637,34 @@ export class Dropdown
     }
   }
 
-  private focusOnFirstActiveOrFirstItem = (): void => {
-    this.getFocusableElement(
-      this.getTraversableItems().find((item) => item.selected) || this.items[0],
-    );
+  private focusOnFirstActiveOrDefaultItem = (focusLastItem: boolean): void => {
+    const selectedItem = this.getTraversableItems().find((item) => item.selected);
+    if (selectedItem) {
```
jcfranco, line 641: For readability, can you add a newline before the `if` statement?

packages/calcite-components/src/components/dropdown/dropdown.tsx
```diff
@@ -634,29 +637,34 @@ export class Dropdown
...
+      this.focusDropdownItemElement(selectedItem);
```
jcfranco, line 642: Can you DRY up this statement? Something like:

```ts
const target = selectedItem || focusLastItem 
  ? this.items[this.items.length - 1] 
  : this.items[0];

this.focusDropdownItemElement(target);
```

packages/calcite-components/src/components/dropdown/dropdown.tsx
```diff
@@ -634,29 +636,34 @@ export class Dropdown
...
-
     focusElement(item);
   }
 
-  private toggleOpenEnd = (): void => {
-    this.focusOnFirstActiveOrFirstItem();
-    this.el.removeEventListener("calciteDropdownOpen", this.toggleOpenEnd);
+  private toggleOpenEnd = (focusLastItem: boolean): void => {
+    this.focusOnFirstActiveOrDefaultItem(focusLastItem);
+    this.el.removeEventListener("calciteDropdownOpen", () => this.toggleOpenEnd(focusLastItem));
```
jcfranco, line 660: Sidebar: I think we could remove the explicit event listener add/remove by leveraging the [`OpenCloseComponent` hooks](https://github.com/Esri/calcite-design-system/blob/1cb3d9055f0f4d36effd83c1543e1baae78c275f/packages/calcite-components/src/components/dropdown/dropdown.tsx#L538-L552).

jcfranco (approved):
Aside from a few comments, this LGTM! ↕️👁️

jcfranco:
Commit type-wise, should this be a `feat`? We're improving the keyboard navigation experience. cc @driskull @geospatialem

geospatialem:
> Commit type-wise, should this be a `feat`? We're improving the keyboard navigation experience. cc @driskull @geospatialem

While it seems like new functionality is being added, this is the expected functionality of the component, so would lean towards `fix` to mitigate the 🐛 in this case.

packages/calcite-components/src/components/dropdown/dropdown.tsx
anveshmekala: This will be irrelevant after https://github.com/Esri/calcite-design-system/pull/10264#discussion_r1752922398

packages/calcite-components/src/components/dropdown/dropdown.tsx
```diff
@@ -636,35 +637,23 @@ export class Dropdown
...
-      );
-    }
-  };
+    const target: HTMLCalciteDropdownItemElement =
+      selectedItem ||
+      (this.focusLastDropdownItem ? this.items[this.items.length - 1] : this.items[0]);
+
+    this.focusLastDropdownItem = false;
 
-  private focusDropdownItemElement(item: HTMLCalciteDropdownItemElement): void {
```
anveshmekala: took the liberty and refactored. @jcfranco can i get one more review?

packages/calcite-components/src/components/dropdown/dropdown.tsx
anveshmekala: I think we may have to depend on explicit event listener to avoid focusing items when `open` attribute is changed programmatically. 

```
  onOpen(): void {
    this.calciteDropdownOpen.emit();
    this.focusOnFirstActiveOrDefaultItem();
  }
  ```
  
  Tried the above 👆 and the issue with that is when `open` is toggled without user interaction (programmatically) , `onOpen( )` method is triggered and focus shift happens. Is this expected?

File: packages/calcite-components/src/components/dropdown/dropdown.tsx [hunk 5]
```diff
@@ -573,24 +576,24 @@ export class Dropdown
...
       event.preventDefault();
+      return;
+    }
+
+    if (isActivationKey(key)) {
+      this.clickHandler();
+      event.preventDefault();
+    } else if (key === "ArrowDown" || key === "ArrowUp") {
+      this.focusLastDropdownItem = key === "ArrowUp";
+      this.open = true;
```
jcfranco, line 596: We should cancel the event here too since it's handled.

packages/calcite-components/src/components/dropdown/dropdown.tsx
```diff
@@ -634,30 +637,23 @@ export class Dropdown
...
-  }
-
-  private toggleOpenEnd = (): void => {
-    this.focusOnFirstActiveOrFirstItem();
-    this.el.removeEventListener("calciteDropdownOpen", this.toggleOpenEnd);
+    focusElement(target);
   };
 
-  private openCalciteDropdown = () => {
+  private clickHandler = (): void => {
```
jcfranco, line 655: Can we keep this dedicated click handler and restore the previous `openCalciteDropdown` (name could be simplified to `openDropdown`)? That way, other event handlers can call this dedicated open method instead of invoking an event handler for a different event.

packages/calcite-components/src/components/dropdown/dropdown.tsx
jcfranco: Took a closer look after our conversation yesterday and yes, this is expected. Our open/close events emit for both programmatic and user interaction (https://github.com/Esri/calcite-design-system/pull/7613/files has a draft for guidelines on this). 

Does this impact any dropdown workflows? If not, let's proceed. Otherwise, we can restore the explicit event listeners and refactor to use the open/close events in the next release.

jcfranco (approved):
☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️
☂️↕️☂️☂️☂️↕️☂️☂️↕️↕️☂️☂️↕️↕️↕️☂️☂️↕️↕️↕️☂️↕️↕️↕️↕️☂️↕️☂️
☂️↕️↕️☂️☂️↕️☂️↕️☂️☂️↕️☂️☂️↕️☂️☂️↕️☂️☂️☂️☂️↕️☂️☂️☂️☂️↕️☂️
☂️↕️☂️↕️☂️↕️☂️↕️☂️☂️↕️☂️☂️↕️☂️☂️↕️☂️☂️☂️☂️↕️↕️↕️☂️☂️↕️☂️
☂️↕️☂️☂️↕️↕️☂️↕️☂️☂️↕️☂️☂️↕️☂️☂️↕️☂️☂️☂️☂️↕️☂️☂️☂️☂️☂️☂️
☂️↕️☂️☂️☂️↕️☂️☂️↕️↕️☂️☂️↕️↕️↕️☂️☂️↕️↕️↕️☂️↕️↕️↕️↕️☂️↕️☂️
☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️☂️

anveshmekala:
>Does this impact any dropdown workflows? If not, let's proceed. Otherwise, we can restore the explicit event listeners and refactor to use the open/close events in the next release.

It does add focus to the first or last item in the dropdown group when user programmatically toggles the open. Screenshot test s reported this issue in `split-button`.

Will revert back to using `eventListener` instead of `OpenCloseComponent` hooks.

jcfranco:
> It does add focus to the first or last item in the dropdown group when user programmatically toggles the open. Screenshot test s reported this issue in split-button.

This seems fine to me, but I could be missing something. In any case, let's revisit this separately.

packages/calcite-components/src/components/dropdown/dropdown.tsx
anveshmekala: i think `clickHandler( )` isn't required and we can use `toggleDropdown` (instead of openDropdown)  method to handle click.

PR #10270, opened 2024-09-11
feat(tabs): add component tokens

Elijbet:
**Related Issue:** #7180

## Summary
Add `tabs` component tokens.

`--calcite-tabs-background-color`: The background color of the component.
`--calcite-tabs-border-color`: The border color of the component.


alisonailea (changes requested):
This needs to be done in conjunction with tabs, tab-title, tab-nav. There is a larger design pattern that these components need to coordinate. You can't just change the border color in tabs and not tab-title.

jcfranco:
Closing in favor of https://github.com/Esri/calcite-design-system/pull/10532. Thanks for taking an early stab a this, @Elijbet!

Issue #10011, opened 2024-08-08
CI: Priority labels aren't added in as-expected


geospatialem:
### Priority impact

p - low

### Summary

Ensure priority labels are added by the CI when added via issue templates.

### Desired Outcome

Automatically add in the priority label :-)

### Resources

As an example, in #10010 the [run failed](https://github.com/Esri/calcite-design-system/actions/runs/10308576369). I had to individually added in the `p - low` label, despite our automation  https://github.com/Esri/calcite-design-system/issues/10010#event-13811560045

DitwanP:
🍡 Verified 👆
<img width="277" alt="Screenshot 2024-05-21 at 11 06 30 AM" src="https://github.com/user-attachments/assets/f887e6df-d3d0-44d1-b3ef-b6cc10cc32e2">

PR #10273, opened 2024-09-11
ci(actions): fix adding labels action failing on issues without "Esri team" line

DitwanP:
**Related Issue:** #10011 

## Summary
The original issue was for fixing the action not successfully adding the priority label to a newly created issue. It looks like the reason that wasn't working was because the action would fail before getting to that step. 

The failure was caused by the `addEsriProductLabel` step of the action expecting there to always be an "Esri team" line in the issue body but that's not the case for all issue types.

This is just a small change to stop the failure and instead log a message when the "Esri team" is not present on an issue and continue to the next steps.


benelan (approved):
:+1:

PR #10278, opened 2024-09-11
feat: add web adaptor

aphilobos:
**Related Issue:** #

## Summary
**added**
web-adaptor-16
web-adaptor-24
web-adaptor-32

file_05_screenshot


packages/calcite-ui-icons/docs/keywords.json
```diff
@@ -8860,5 +8860,80 @@
...
+      "format",
+      "justification"
+    ],
+    "category": "Lists",
+    "release": "3.29.6"
+  },
+  "web-adaptor": {
+    "alias": ["3.29.7", "gis", "core", "web", "diagram", "nodes", "gis", "storage", "servers", "transfer"],
+    "category": "GIS",
+    "release": "3.29.7"
```
benelan, line 8937: The version here is incorrect, it will be released as `3.32.0`. For reference, the versioning strategy is called [`semver`](https://semver.org/). 

TLDR; The format is `<MAJOR>.<MINOR>.<PATCH>`

- When changing an existing icon, use the `fix:` prefix in the PR title and increment the `<PATCH>` number.

- When you add icons, use the `feat:` prefix in the PR title, increment the `<MINOR>` version, and set the `<PATCH>` number to `0`.

- When introducing a change that breaks previous functionality (e.g. renaming icons), use the `refactor!:` prefix, increment the `<MAJOR>` number, and set the `<MINOR>` and `<PATCH>` versions to `0`.

Please let me, @jcfranco, and @geospatialem know before installing breaking changes, because they require advanced messaging to our users.

packages/calcite-ui-icons/docs/keywords.json
```diff
@@ -8860,5 +8860,80 @@
...
+  },
+  "connection-to-connection": {
+    "alias": ["3.29.5", "generic", "connect", "two points", "line"],
+    "category": "Generic",
+    "release": "3.29.5"
+  },
+  "connection-middle": {
+    "alias": ["3.29.5", "generic", "line", "point", "middle", "connect"],
+    "category": "Generic",
+    "release": "3.29.5"
```
benelan, line 8917: The versions starting at `3.29.1` are also incorrect. Can you please create a separate `fix` PR that updates them to the correct versions? You can see which version the icons were released as by checking the [changelog](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-ui-icons/CHANGELOG.md)

benelan (changes requested):

Issue #10235, opened 2024-09-06
[Slider] potentially unintentional property name collision


maxpatiiuk:
### Description

HTMLElement has a "translate" boolean property
slider.tsx has a "translate" private method
This collision is potentially unintentional - could you rename the private method?

### Proposed Advantages

Avoids collision

### Which Component

slider

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

DitwanP:
🍡 Verified

PR #10293, opened 2024-09-12
refactor(slider): rename internal function to avoid collision with `HTMLElement.translate`

jcfranco:
**Related Issue:** #10235

## Summary

[`HTMLElement.translate`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/translate) is a newly available baseline property, so this renames an existing function to avoid any conflicts.

### DIFF
--- a/packages/calcite-components/src/components/slider/slider.tsx
+++ b/packages/calcite-components/src/components/slider/slider.tsx
@@ -561,7 +561,7 @@ export class Slider
     }
 
     const x = event.clientX || event.pageX;
-    const position = this.translate(x);
+    const position = this.mapToRange(x);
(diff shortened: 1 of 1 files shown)


driskull (approved):
👍 Nice!

Do we need to update esLint or something to catch this next time?

maxpatiiuk:
> Do we need to update esLint or something to catch this next time?

No need. You will get a TypeScript error in this case in Lit:

> Property 'translate' in type 'Slider' is not assignable to the same property in base type 'LitElement'.
>    Type '(x: number) => number' is not assignable to type 'boolean'.ts(2416)

PR #10295, opened 2024-09-13
refactor: fixup typescript errors

maxpatiiuk:
With the changes in this PR, after running the codemod, there are only 60 TypeScript errors in the Calcite package - those should be best fixed manually after running the codemod.
Now that error count is down to negligible level, will work on migrating tests.

### DIFF
--- a/packages/calcite-components/src/components/button/button.tsx
+++ b/packages/calcite-components/src/components/button/button.tsx
@@ -280,11 +280,9 @@ export class Button
           disabled={childElType === "button" ? this.disabled || this.loading : null}
           download={
             childElType === "a"
-              ? this.download === true
+              ? this.download === true || this.download === ""
                 ? ""
-                : this.download
-                  ? this.download
-                  : null
+                : this.download || null
               : null
           }
           href={childElType === "a" && this.href}
--- a/packages/calcite-components/src/components/button/button.tsx
+++ b/packages/calcite-components/src/components/button/button.tsx
@@ -279,7 +279,13 @@ export class Button
           }}
           disabled={childElType === "button" ? this.disabled || this.loading : null}
           download={
-            childElType === "a" && (this.download === "" || this.download) ? this.download : null
+            childElType === "a"
+              ? this.download === true
+                ? ""
+                : this.download
+                  ? this.download
+                  : null
+              : null
           }
           href={childElType === "a" && this.href}
           name={childElType === "button" && this.name}
--- a/packages/calcite-components/src/components/dialog/dialog.tsx
+++ b/packages/calcite-components/src/components/dialog/dialog.tsx
@@ -43,7 +43,7 @@ import {
 import { componentOnReady } from "../../utils/component";
 import { SLOTS as PANEL_SLOTS } from "../panel/resources";
 import { HeadingLevel } from "../functional/Heading";
-import { OverlayPositioning } from "../../components";
+import type { OverlayPositioning } from "../../utils/floating-ui";
 import { DialogMessages } from "./assets/dialog/t9n";
 import {
   CSS,
--- a/packages/calcite-components/src/components/functional/FloatingArrow.tsx
+++ b/packages/calcite-components/src/components/functional/FloatingArrow.tsx
@@ -2,7 +2,7 @@ import { FunctionalComponent, h, VNode } from "@stencil/core";
 import { JSXAttributes } from "@stencil/core/internal";
 import { FloatingLayout } from "../../utils/floating-ui";
 
-interface FloatingArrowProps extends JSXAttributes {
+interface FloatingArrowProps extends JSXAttributes<SVGSVGElement> {
   floatingLayout: FloatingLayout;
 }
 
--- a/packages/calcite-components/src/components/handle/handle.e2e.ts
+++ b/packages/calcite-components/src/components/handle/handle.e2e.ts
@@ -1,6 +1,6 @@
 import { newE2EPage } from "@stencil/core/testing";
 import { accessible, disabled, hidden, renders, themed, t9n } from "../../tests/commonTests";
-import { HandleMessages } from "../../components";
+import type { HandleMessages } from "./assets/handle/t9n";
 import { CSS, SUBSTITUTIONS } from "./resources";
 
 describe("calcite-handle", () => {
--- a/packages/calcite-components/src/components/link/link.tsx
+++ b/packages/calcite-components/src/components/link/link.tsx
@@ -113,7 +113,9 @@ export class Link implements InteractiveComponent, LoadableComponent {
           When the 'download' property of type 'boolean | string' is set to true, the value is "".
           This works around that issue for now.
           */
-            download={Tag === "a" ? (download === true ? "" : download ? download : null) : null}
+            download={
+              Tag === "a" ? (download === true || download === "" ? "" : download || null) : null
+            }
             href={Tag === "a" && this.href}
             onClick={this.childElClickHandler}
             ref={this.storeTagRef}
--- a/packages/calcite-components/src/components/link/link.tsx
+++ b/packages/calcite-components/src/components/link/link.tsx
@@ -113,7 +113,7 @@ export class Link implements InteractiveComponent, LoadableComponent {
           When the 'download' property of type 'boolean | string' is set to true, the value is "".
           This works around that issue for now.
           */
-            download={Tag === "a" && (download === "" || download) ? download : null}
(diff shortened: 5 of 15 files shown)


packages/calcite-components/src/components/button/button.tsx
```diff
@@ -279,7 +279,13 @@ export class Button
           }}
           disabled={childElType === "button" ? this.disabled || this.loading : null}
           download={
-            childElType === "a" && (this.download === "" || this.download) ? this.download : null
+            childElType === "a"
```
maxpatiiuk, line 282: with this change, when `download` is true, we set the prop as `""` instead.
not required for Stencil, but this is needed for Lit as `download` attribute is typed as `string` rather than `string | boolean` in Lumina for technical reasons (it's one few such special attributes for legacy dom reasons)

packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -43,7 +43,7 @@ import {
 import { componentOnReady } from "../../utils/component";
 import { SLOTS as PANEL_SLOTS } from "../panel/resources";
 import { HeadingLevel } from "../functional/Heading";
-import { OverlayPositioning } from "../../components";
+import type { OverlayPositioning } from "../../utils/floating-ui";
```
maxpatiiuk, line 46: since components.d.ts will be deleted, I'm changing this to import from the actual origin

packages/calcite-components/src/components/functional/FloatingArrow.tsx
```diff
@@ -2,7 +2,7 @@ import { FunctionalComponent, h, VNode } from "@stencil/core";
 import { JSXAttributes } from "@stencil/core/internal";
 import { FloatingLayout } from "../../utils/floating-ui";
 
-interface FloatingArrowProps extends JSXAttributes {
+interface FloatingArrowProps extends JSXAttributes<SVGSVGElement> {
```
maxpatiiuk, line 5: this way `ref` callback is typed correctly, rather than as broad `Element`

packages/calcite-components/src/components/link/link.tsx
```diff
@@ -113,7 +113,7 @@ export class Link implements InteractiveComponent, LoadableComponent {
           When the 'download' property of type 'boolean | string' is set to true, the value is "".
           This works around that issue for now.
           */
-            download={Tag === "a" && (download === "" || download) ? download : null}
+            download={Tag === "a" ? (download === true ? "" : download ? download : null) : null}
```
maxpatiiuk, line 116: same as in `button.tsx`

packages/calcite-components/src/components/link/link.tsx
```diff
@@ -174,7 +174,7 @@ export class Link implements InteractiveComponent, LoadableComponent {
   /** the rendered child element */
   private childEl: HTMLAnchorElement | HTMLSpanElement;
 
-  private childElClickHandler = (event: PointerEvent): void => {
+  private childElClickHandler = (event: MouseEvent): void => {
```
maxpatiiuk, line 177: onClick emits MouseEvent, not PointerEvent

packages/calcite-components/src/components/panel/panel.tsx
```diff
@@ -718,7 +718,7 @@ export class Panel
     );
 
     return (
-      <Host onKeyDown={panelKeyDownHandler}>
+      <Host onKeyDown={this.panelKeyDownHandler}>
```
maxpatiiuk, line 721: noop change, but this makes it easy for the codemod to refactor-out `<Host>`

---

I earlier mentioned that special utils would be needed to replace `<Host>` in Lit, but after reading `<Host>` source code in Stencil, that doesn't seem to be necessary

packages/calcite-components/src/components/popover/popover.tsx
```diff
@@ -268,7 +268,7 @@ export class Popover
 
   @State() defaultMessages: PopoverMessages;
 
-  arrowEl: SVGElement;
+  arrowEl: SVGSVGElement;
```
maxpatiiuk, line 271: the correct type for the `<svg>` element is `SVGSVGElement`.
`SVGElement` means "any element from the SVG namespace", just like `HTMLElement` means "any element from the HTML namespace"

packages/calcite-components/src/components/rating/rating.tsx
```diff
@@ -356,7 +356,7 @@ export class Rating
     }
   };
 
-  private handleInputChange = (event: InputEvent) => {
+  private handleInputChange = (event: Event) => {
```
maxpatiiuk, line 359: onChange emits Event
onInput emits InputEvent

packages/calcite-components/src/components/tooltip/tooltip.tsx
```diff
@@ -344,7 +344,7 @@ export class Tooltip implements FloatingUIComponent, OpenCloseComponent {
         >
           <FloatingArrow
             floatingLayout={floatingLayout}
-            ref={(arrowEl: SVGElement) => (this.arrowEl = arrowEl)}
+            ref={(arrowEl) => (this.arrowEl = arrowEl)}
```
maxpatiiuk, line 347: no longer need to type this here now that FloatingArrow ref type is typed correctly

packages/calcite-components/src/demos/input-time-picker.html
maxpatiiuk, line 1: these link/script tags seem needless here since they are inserted by _assets/head.ts anyway

packages/eslint-plugin-calcite-components/src/rules/require-event-emitter-type.ts
```diff
@@ -1,5 +1,4 @@
 import { Rule } from "eslint";
-// @ts-ignore
```
maxpatiiuk, line 2: PSA: do not use `// @ts-ignore` as it's dangerous.
Reason: once the thing that was causing the TypeScript error is fixed, the `// @ts-ignore` comment may remain in the code. In the future, if another TypeScript error may occur in that place, you won't see it because `// @ts-ignore` would hide it.

Safer alternative: `// @ts-expect-error` - acts just like `// @ts-ignore`, except, if no error was reported, TypeScript will complain of needless `// @ts-expect-error` (just like it was needless in this case it seems like)

packages/calcite-components/src/demos/input-time-picker.html
jcfranco: Thanks for the cleanup!

BTW, I extracted the demo-related changes to a [separate PR](https://github.com/Esri/calcite-design-system/pull/10314) to keep this one focused on TS errors.

jcfranco (changes requested):
Looking great, @maxpatiiuk! 🚀 

Before moving forward, can you address the button and link-related test failures?

packages/calcite-components/src/components/link/link.tsx
maxpatiiuk: Thanks for the fyi.
I will revert this change and update Lumina's JSX typings.

For the future, I am looking to automate the process of keeping Lumina's JSX typings up to date with the web platform to fix the DX issue we are having with Maquette and Stencil.

packages/calcite-components/src/demos/input-time-picker.html
maxpatiiuk: Sounds good. I will undo those changes in this PR

packages/calcite-components/src/components/handle/handle.e2e.ts
```diff
@@ -1,7 +1,7 @@
 import { newE2EPage } from "@stencil/core/testing";
 import { accessible, disabled, hidden, renders, t9n } from "../../tests/commonTests";
-import { HandleMessages } from "../../components";
 import { CSS, SUBSTITUTIONS } from "./resources";
+import type { HandleMessages } from "./assets/handle/t9n";
```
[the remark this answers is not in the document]
maxpatiiuk: hmm, interesting that pre-commit hook didn't catch this. maybe my node_modules was out of date 🤔

maxpatiiuk:
Tooltip e2e test is failing on the CI.
Doesn't seem related to the changes I done in the PR.
Locally, a different tooltip e2e test is failing (`calcite-tooltip › openClose › emits with animations enable`).
Unstable test, or is there something wrong with my local setup?

jcfranco:
That's an unstable test (I'll try to stabilize soon). I've triggered another run of E2E tests in the meantime.

jcfranco (approved):
Awesome! Thanks, @maxpatiiuk! 🚀

Issue #10236, opened 2024-09-06
[Refactor] refactor-out JSX spread usages.


maxpatiiuk:
### Description

Calcite has 10 usages of JSX spread syntax (`{...props}`).
This syntax is not supported in lit-html at the moment.
Several alternative solutions are available.
See [this documentation](https://qawebgis.esri.com/arcgis-components/?path=/docs/references-lumina-jsx--docs#spread-attributes) page for options

### Proposed Advantages

Better performance
Better type checking

### Which Component

Several - search for `\{...` in your IDE

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

DitwanP:
✨🧹✨

PR #10296, opened 2024-09-13
refactor: avoid and prevent spread attribute usage

jcfranco:
**Related Issue:** #10236

## Summary

Moves away from JSX attribute spreading pattern.

This also enables [`react/jsx-props-no-spreading`](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-props-no-spreading.md) to prevent the spread of future spreading (🥁🐍)


driskull (approved):
⭐

PR #10300, opened 2024-09-13
chore: add cssrem VSCode extension recommendation

jcfranco:
**Related Issue:** N/A

## Summary

Adds the [`cssrem`](https://marketplace.visualstudio.com/items?itemName=cipchk.cssrem) extension to display unit conversions within the editor. This will help avoid adding comments for unit conversions (see https://github.com/Esri/calcite-design-system/pull/10179#discussion_r1752943775).

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,9 +1,10 @@
 {
   "recommendations": [
-    "dbaeumer.vscode-eslint",
-    "stylelint.vscode-stylelint",
-    "esbenp.prettier-vscode",
     "bradlc.vscode-tailwindcss",
+    "cipchk.cssrem",
+    "dbaeumer.vscode-eslint",
+    "esbenp.prettier-vscode",
+    "stylelint.vscode-stylelint",
     "visualstudioexptteam.vscodeintellicode"
   ]
 }


alisonailea:
Per my comment on #10179, I don't think this needs to be added as it's solving for a short-term problem. Perhaps we should reprioritize the switch to rems in the new spacing tokens so it can be used sooner rather than later @SkyeSeitz @ashetland

jcfranco:
Noted, but I’d really like to avoid the noise and overhead of manually adding conversion comments. We could treat this as a temporary extension and remove it once the rem switch is in place.

alisonailea:
With this extension, all these tokens will have their values replaced with pixels. When the time comes to find/replace these static values with spacing tokens it's going to be a lot easier to find these with the "noise" than to try and filter out all the other pixel values that currently exist for other reasons.

jcfranco:
> With this extension, all these tokens will have their values replaced with pixels.

There’s a miscommunication. I wasn't proposing this extension to switch units, but rather to display the conversion within the editor. To be clear, we would keep rem values, and this would just help display their pixel conversion.

The file from the linked discussion isn’t the [only](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/calcite-preset.ts#L83) [place](https://github.com/Esri/calcite-design-system/blob/d1691b07db2747e51430520b44a977c30a9c9758/packages/calcite-components/src/assets/styles/_type.scss#L6) [where](https://github.com/Esri/calcite-design-system/blob/1ae7cc1fac97a0abb64a38629ee9244cd8c9e161/packages/calcite-components/src/components/loader/loader.scss#L42) this occurs.

> When the time comes to find/replace these static values with spacing tokens it's going to be a lot easier to find these with the "noise" than to try and filter out all the other pixel values that currently exist for other reasons.

If we’re adding these comments for tracking and future replacement, let’s make sure the comments are clear about their purpose. /* 1337px */ isn’t clear in this regard. If there’s a spacing token issue, we could also use the comment to track future token replacements. That said, I agree the former is the simpler option for the switch.

driskull (approved):
👍

PR #10308, opened 2024-09-13
fix(panel): initially closed panel should be hidden

driskull:
**Related Issue:** #10320

## Summary

- panel that is initially closed should not be displayed
- add story
- add test


packages/calcite-components/src/components/panel/panel.e2e.ts
```diff
@@ -219,6 +219,18 @@ describe("calcite-panel", () => {
     expect(await container.isVisible()).toBe(false);
   });
 
+  it("honors closed prop initially", async () => {
```
jcfranco, line 222: Do we need this if it's covered by the screenshot test, or vice versa?

jcfranco (approved):

packages/calcite-components/src/components/panel/panel.e2e.ts
driskull: We could do either. Which do you prefer? e2e?

jcfranco: Maybe E2E to eliminate one screenshot test?

PR #10309, opened 2024-09-13
fix: correct non-standard filled icon names

arowles:
## Summary

The following icons were added:

layers-16-f
layers-24-f
layers-32-f
refresh-16-f 
refresh-24-f
refresh-32-f 
rss-16-f
rss-24-f
rss-32-f

The following icons are deprecated and will be removed in v4:

layers-f-16
layers-f-24
layers-f-32
refresh-f-16
refresh-f-24
refresh-f-32
rss-f-16
rss-f-24
rss-f-32


jcfranco:
The PR title seems a bit unwieldy. Could we revisit it to make it more general? Something like `fix: correct non-standard filled icon names` or something similar? The description is perfect since it goes into detail regarding the fixes. 🏆

benelan (changes requested):
Why are some of the old versions like `3.3.0` being changed to `3.32.0`? Also should this be considered a breaking change since it is renaming icons?

arowles:
@benelan does the release number change when renaming icons or do they remain the same?

benelan:
> @benelan does the release number change when renaming icons or do they remain the same?

Yeah renaming icons would be a breaking change because anyone using the old names would have to change them. I recommend not removing the old ones for now. We have a breaking change release in a few months and we can remove them at that point.

So the version for the new names should be `3.32.0`. Also the `keywords.json` issues that I mentioned above are still there. Thanks!

arowles:
@benelan  Do I leave the already changed release numbers in the keywords json file? Also, what other issues in the keywords file?

benelan:
@arowles would you like me to update the keywords file for you? That might be easier.

benelan (approved):
LGTM after I made a few changes to `keywords.json` and the pull request body. Make sure it looks good to you too before installing, thanks!

benelan:
> @benelan that would probably be easier and I will remember to not change the release numbers on the keywords doc for renaming icons on the future. Thanks!

The release numbers should be changed when renaming icons. The issue was there were changes to icons that weren't renamed in this pull request. You can use this pull request as an example for the future.

geospatialem:
@arowles Does this fulfill the bug request we received for `"layers-f"` detailed in https://github.com/Esri/calcite-design-system/issues/10312, or is that for inclusion at a later time?

arowles:
> @arowles Does this fulfill the bug request we received for `"layers-f"` detailed in #10312, or is that for inclusion at a later time?



> @arowles Does this fulfill the bug request we received for `"layers-f"` detailed in #10312, or is that for inclusion at a later time?

Yes this does include the bug request. My apologies I forgot to add the issue id you previously referenced

PR #10311, opened 2024-09-13
refactor(stylelint): change config to module format to enable more dynamic options

jcfranco:
**Related Issue:** N/A

## Summary

This change will give us more flexibility for sharing rule options and for more dynamic rule updates (e.g., #10188).


benelan (approved):
:+1:

Issue #9121, opened 2024-04-12
[Carousel] Improve support for high item counts


ashetland:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Carousel is currently best used with a small number of items. It could be worth supporting more items or "slides" that exceeds the page indicator display. A user could have 15 slides, but only has space for 5 page indicators.

### Acceptance Criteria

A large number of Carousel Items could be supported via a maxPagesDisplayed property or this could be automatic based on width and space available. If there are more items than there are page indicators, there would need to be some kind of queuing-like visual affordance to inform the user that there are more pages than the currently visible indicators show. Would also need to explore what it would look like when the user reaches the last item and loops back to the beginning item.

### Relevant Info

cc @macandcheese 

### Which Component

Carousel

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

asangma:
I think adding support for a CarouselMenu might also be a good solution.

macandcheese:
> I think adding support for a CarouselMenu might also be a good solution.

I don't know if a Carousel's main purpose is to serve intention-based browsing like that ("show me slide 5 of 8").

I understand this makes sense in some cases (a gallery of known imagery, selected features, etc. - "I want to go to view feature abc") - but I'd expect a navigation experience like that to be built somewhere _around_ the Carousel - in a Panel with associated UI, etc.

In the future, each "representative dot" might also not represent a single entity - but maybe representing 2/3 Carousel Items in view: https://github.com/Esri/calcite-design-system/issues/9104

Main intention with this is to resolve the following:

<img width="1065" alt="Screenshot 2024-07-02 at 12 05 55 PM" src="https://github.com/Esri/calcite-design-system/assets/4733155/3e915443-ae10-4db7-9ba2-d50eed1a8603">
<img width="388" alt="Screenshot 2024-07-02 at 12 16 13 PM" src="https://github.com/Esri/calcite-design-system/assets/4733155/5619399e-ec16-4b23-9754-3fd566baf317">


I'd propose something like the following:

Add a `max-displayed-items` property that when set:
[...]

Issue #7804, opened 2023-09-19
Pagination - Ensure implicit list markup is avoided - (2036570528)


dqateam:
### Violation: 
 Ensure implicit list markup is avoided 

![image](https://github.com/Esri/calcite-design-system/assets/5023024/5baa68b9-cb7d-40fd-b6af-54970b0d9295)

### WCAG Reference: 
- [1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html) 

### Severity: 
 6 

### Media Type: 
 Lists 

--- 
### Areas for Remediation: 
- Title: Pagination - Ensure implicit list markup is avoided - (2036570528) 
- Module: Pagination 
**Issue**
There is content that appears as a list but does not use list markup. Examples include:
- "Pagination component"

**User Impact**
Screen reader users will have difficulty understanding that this content is a list.

**Code Reference** 
 ```
<button aria-label="Previous" class="previous"><calcite-icon aria-hidden="true" icon="chevronLeft" flip-rtl="" scale="s" calcite-hydrated=""></calcite-icon></button>
<button class="page">1</button>
(...)
<button class="page">1,500</button>
<button aria-label="Next" class="next"><calcite-icon aria-hidden="true" icon="chevronRight" flip-rtl="" scale="s" calcite-hydrated=""></calcite-icon></button>
 ``` 
**Suggestion**
Ensure implicit list markup is avoided. Lists must use ``` <ul> ```, ``` <ol> ```, ``` <li> ```, ``` <dl> ```, ``` <dt> ```, and ``` <dd> ``` elements as appropriate.

**Compliant Code Example** 
 ```
<ul>
<li><button aria-label="Previous" class="previous"><calcite-icon aria-hidden="true" icon="chevronLeft" flip-rtl="" scale="s" calcite-hydrated=""></calcite-icon></button>
<button class="page">1</button></li>
(...)
<li><button class="page">1,500</button></li>
<li><button aria-label="Next" class="next"><calcite-icon aria-hidden="true" icon="chevronRight" flip-rtl="" scale="s" calcite-hydrated=""></calcite-icon></button>
</li>
<ul>


 ```
 
--- 
### Additional Resources: 
- Report Source: Product accessibility evaluation conducted on the Calcite Design System in August, 2022.

driskull:
This one seems valid, internal shadow HTML should be more semantic.

DitwanP:
[...]

PR #10315, opened 2024-09-14
feat(carousel): Improve support for high item counts

macandcheese:
**Related Issue:** #9121 

Adds support for better display of high Carousel Item counts by automatically adjusting the number shown as container width changes. 

https://github.com/user-attachments/assets/0d682f21-ac2e-4a4a-830c-e15b6a5faf1d

## Summary

This follows the pattern found in Pagination, as the behavior is quite similar, but there are a few differences in Carousel to support the "rotation effect" of the "dots".

- Uses existing breakpoints to determine when to reduce displayed "dots".
- Uses a "represents more" styling when the first / last visible "dot" has more siblings before / after.
- When in the "start or end range" - the first / last visible "dot" will not have a style treatment as it represents the start / end.
- Updates stories, tests, and local demos with new feature.


packages/calcite-components/src/components/carousel/carousel.tsx
```diff
@@ -609,31 +646,52 @@ export class Carousel
...
+              aria-selected={toAriaBoolean(match)}
+              class={{
+                [CSS.paginationItem]: true,
+                [CSS.paginationItemIndividual]: true,
+                [CSS.paginationItemSelected]: match,
+                [CSS.paginationItemRangeEdge]: length - 1 > maxItems && isEdge,
+                [CSS.paginationItemOutOfRange]: !(overflowActive || visible),
+                [CSS.paginationItemVisible]: overflowActive || visible,
+              }}
+              data-index={index}
```
driskull, line 682: Is the `data-index` used?

driskull (changes requested):
Code looks good @macandcheese. Just a couple things.

packages/calcite-components/src/components/carousel/carousel.tsx
```diff
@@ -609,31 +646,52 @@ export class Carousel
...
-            <calcite-icon icon={isMatch ? ICONS.active : ICONS.inactive} scale="l" />
-          </button>
-        );
-      })}
-    </div>
-  );
+  renderPaginationItems = (): VNode => {
+    const { selectedIndex, maxItems, items, label, handleItemSelection } = this;
+    return (
+      <div aria-label={label} class={CSS.paginationItems} role="tablist">
```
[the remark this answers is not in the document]
macandcheese: My understanding was the `tablist` role is sufficient (following https://www.w3.org/WAI/ARIA/apg/patterns/carousel/examples/carousel-2-tablist/) - but happy to adjust if needed. 

cc @geospatialem to confirm. If needed, I'd like to do this as a follow up as this wasn't changed in this PR.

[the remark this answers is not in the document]
driskull: We've had other issues where we had to convert a list of buttons into an actual list. I'll dig them up.

packages/calcite-components/src/components/carousel/carousel.tsx
```diff
@@ -528,7 +566,8 @@ export class Carousel
   };
 
   private tabListKeyDownHandler = (event: KeyboardEvent): void => {
-    const interactiveItems = Array(...this.tabList.querySelectorAll("button"));
+    const targetEls = `button:not(.${CSS.paginationItemOutOfRange})`;
+    const interactiveItems = Array(...this.tabList.querySelectorAll(targetEls));
```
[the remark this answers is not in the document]
macandcheese: Yeah definitely. Before this PR it could have been done just once too on slotchange.

I can still store it outside of this function, but it will still need to update more frequently than onslotchange. Setting up an "isOverflowing" state might be helpful to reduce the number of times this is set as it will not always change.

packages/calcite-components/src/components/carousel/carousel.tsx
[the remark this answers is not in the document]
driskull: Example: https://github.com/Esri/calcite-design-system/issues/7804

packages/calcite-components/src/components/carousel/carousel.tsx
macandcheese: Used here to determine the requested "item" https://github.com/Esri/calcite-design-system/blob/d53f47c0c86b6721a1b74e3c8ec69f20962d3543/packages/calcite-components/src/components/carousel/carousel.tsx#L453 - LMK if this should use a different pattern.

packages/calcite-components/src/components/carousel/carousel.stories.ts
```diff
@@ -76,6 +76,82 @@ export const simple = (args: CarouselStoryArgs): string =>
     </calcite-carousel>
   </div>`;
 
+export const simpleManyItems = (args: CarouselStoryArgs): string =>
+  html` <div style="width:600px;height:400px;">
+    <calcite-carousel
+      control-overlay="${args.controlOverlay}"
+      ${boolean("disabled", args.disabled)}
+      autoplay-duration="${args.autoplayDuration}"
+      ${args.autoplay ? "autoplay" : ""}
```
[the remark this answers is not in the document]
macandcheese: Yep, will update this and the other stories.

packages/calcite-components/src/components/carousel/carousel.tsx
driskull: No, this sounds good. 👍

packages/calcite-components/src/components/carousel/carousel.tsx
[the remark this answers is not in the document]
driskull: The digitala11y example uses a `ul/li` as well.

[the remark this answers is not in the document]
macandcheese: Made a breakout issue, can move this discussion there and follow up separately: https://github.com/Esri/calcite-design-system/issues/10322

driskull (approved):
👍

packages/calcite-components/src/components/carousel/carousel.tsx
```diff
@@ -154,17 +158,19 @@ export class Carousel
     disconnectLocalized(this);
     disconnectMessages(this);
     this.clearIntervals();
+    this.resizeObserver?.disconnect();
   }
 
   async componentWillLoad(): Promise<void> {
+    this.setMaxItemsToBreakpoint(this.el.clientWidth);
```
jcfranco, line 165: `clientWidth` might not reflect the proper size at this point since the component and items are not rendered at this point.

packages/calcite-components/src/components/carousel/carousel.tsx
```diff
@@ -300,6 +314,26 @@ export class Carousel
   //
   // --------------------------------------------------------------------------
 
+  private setMaxItemsToBreakpoint(width: number): void {
+    if (!breakpoints || !width) {
```
jcfranco, line 318: `breakpoints` should be guaranteed here.

packages/calcite-components/src/components/carousel/carousel.tsx
```diff
@@ -300,6 +314,26 @@ export class Carousel
...
+      return;
+    }
+
+    const breakpointKeys = ["medium", "small", "xsmall", "xxsmall"];
```
jcfranco, line 322: Could you get the keys from either of the breakpoints objects instead?

packages/calcite-components/src/components/carousel/carousel.tsx
```diff
@@ -528,7 +566,8 @@ export class Carousel
   };
 
   private tabListKeyDownHandler = (event: KeyboardEvent): void => {
-    const interactiveItems = Array(...this.tabList.querySelectorAll("button"));
+    const targetEls = `button:not(.${CSS.paginationItemOutOfRange})`;
```
jcfranco, line 569: Can you name this variable to indicate both what it's targeting and that it's a selector? This can also be inlined below.

packages/calcite-components/src/components/carousel/carousel.tsx
```diff
@@ -609,31 +646,52 @@ export class Carousel
...
-        );
-      })}
-    </div>
-  );
+  renderPaginationItems = (): VNode => {
+    const { selectedIndex, maxItems, items, label, handleItemSelection } = this;
+    return (
+      <div aria-label={label} class={CSS.paginationItems} role="tablist">
+        {items.map((item, index) => {
+          const length = items.length;
```
jcfranco, line 654: For readability, can you rename this to `totalItems`, `itemCount`, `numItems` or similar?

packages/calcite-components/src/components/carousel/carousel.tsx
```diff
@@ -609,31 +646,52 @@ export class Carousel
...
+                [CSS.paginationItemOutOfRange]: !(overflowActive || visible),
+                [CSS.paginationItemVisible]: overflowActive || visible,
+              }}
+              data-index={index}
+              key={item.id}
+              onClick={handleItemSelection}
+              role="tab"
+              title={item.label}
+            >
+              <calcite-icon icon={icon} scale={"l"} />
```
jcfranco, line 688: Nitpick: we don't need to interpolate `"l"` here.

packages/calcite-components/src/components/carousel/resources.ts
```diff
@@ -26,3 +29,11 @@ export const ICONS = {
   pause: "pause-f",
   play: "play-f",
 } as const;
+
+export const maxItemBreakpoints = {
```
jcfranco, line 33: Just making sure I'm following, is this a map of breakpoint sizes to max items?

packages/calcite-components/src/components/carousel/carousel.tsx
```diff
@@ -38,14 +38,17 @@ import {
...
 import { getRoundRobinIndex } from "../../utils/array";
-import { CSS, DURATION, ICONS } from "./resources";
+import { CSS, DURATION, ICONS, maxItemBreakpoints } from "./resources";
 import { CarouselMessages } from "./assets/carousel/t9n";
 import { ArrowType, AutoplayType } from "./interfaces";
 
 /**
  * @slot - A slot for adding `calcite-carousel-item`s.
  */
+
```
jcfranco, line 51: For consistency, can you remove the newline between the JSDoc block and the component?

jcfranco (approved):
Aside from some comments, this LGTM! ✨🎠✨

packages/calcite-components/src/components/carousel/carousel.tsx
macandcheese: Yerp, will do

packages/calcite-components/src/components/carousel/carousel.e2e.ts
```diff
@@ -1055,3 +1056,59 @@ describe("calcite-carousel", () => {
...
+        <calcite-carousel-item label="item 4" selected><p>fourth</p></calcite-carousel-item>
+        <calcite-carousel-item label="item 5"><p>fifth</p></calcite-carousel-item>
+        <calcite-carousel-item label="item 6"><p>sixth</p></calcite-carousel-item>
+        <calcite-carousel-item label="item 7"><p>seventh</p></calcite-carousel-item>
+        <calcite-carousel-item label="item 8"><p>eighth</p></calcite-carousel-item>
+      </calcite-carousel>`,
+    );
+
+    const items = await page.findAll(`calcite-carousel >>> .${CSS.paginationItemVisible}`);
+    expect(items).toHaveLength(maxItemBreakpoints["xsmall"] + 2);
```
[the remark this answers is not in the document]
macandcheese: Updated with test for each breakpoint (and removed the upper-most breakpoint)

packages/calcite-components/src/components/carousel/carousel.tsx
macandcheese: I had it in `componentDidLoad()` (like it is in Pagination) - but I was running into `The state/prop "maxItems" changed during "componentDidLoad()"` errors... 

@jcfranco can the resize observer be relied upon to fire during component initialization? Because `setMaxItemsToBreakpoint` is run there as well as in one of these lifecycle periods.

jcfranco: Yes, the callback runs when the [observer is created](https://codepen.io/jcfranco/pen/NWemRed?editors=1111) and any resize afterwards.

macandcheese: In tests as well? (Seems to work .. )

packages/calcite-components/src/components/carousel/carousel.tsx
macandcheese: Updated!

ashetland (approved):
LGTM!

Issue #10214, opened 2024-09-04
Shell content - slotted popover's z-index is set to `1`.


jmanke:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

A `calcite-popover`slotted in the `calcite-shell` gets assigned a z-index of `1`, causing it to be hidden. It looks like the CSS already ignores the `calcite-tip-manager`, so I think it should also ignore `calcite-popover` as well.

![Screenshot 2024-09-04 at 7 08 06 AM](https://github.com/user-attachments/assets/5ffdda57-3f3e-4c47-a634-7c328443cf6d)

And once the z-index override is removed: 

![Screenshot 2024-09-04 at 7 08 58 AM](https://github.com/user-attachments/assets/8e9feb78-0df4-4c83-8a44-5e6c10901723)

### Expected Behavior

A `calcite-popover`slotted in the `calcite-shell` should have its default z-index assigned so it appears on top of content.

### Reproduction Sample

https://codepen.io/jmanke/pen/PorxKEp

### Reproduction Steps

1. Open codepen
2. Observe there is an open `calcite-popover` that does no appear.
3. Inspect the `calcite-popover` - notice how the `z-index` is assigned `1` from a style for the `calcite-shell`.

### Reproduction Version

latest

### Relevant Info

Chrome: Version 128.0.6613.119 (Official Build) (arm64)

### Regression?

Yes

### Priority impact

impact - p2 - want for an upcoming milestone

### Impact

To workaround this issue, we need to directly assign a z-index to the popover.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Dashboards

DitwanP:
🍠 Verified locally on `dev`
![image](https://github.com/user-attachments/assets/05c44a51-43dc-4238-8bb8-4ebfc21fd391)

PR #10325, opened 2024-09-16
fix(shell): fix z-index of slotted popover elements

driskull:
**Related Issue:** #10214

## Summary

- fix css for z-index
  - should only set `z-index` on slotted `calcite-shell-panel`s
- add screenshot test
- regression from here: https://github.com/Esri/calcite-design-system/pull/9912


packages/calcite-components/src/components/shell/shell.scss
```diff
@@ -78,7 +73,8 @@
 
 slot[name="panel-start"]::slotted(calcite-shell-panel),
 slot[name="panel-end"]::slotted(calcite-shell-panel) {
-  @apply relative;
+  @apply relative
+    z-default;
   z-index: calc(var(--calcite-z-index) + 1);
```
jcfranco, line 78: It looks like the `z-default` util is overridden here. Can you confirm?

driskull: Oh, nice catch. Will fix

jcfranco (approved):

PR #10338, opened 2024-09-17
feat: add content inline right

arowles:
**Related Issue:** #

## Summary
**added**

content-inline-right-16
content-inline-right-24
content-inline-right-32

image


jorgegutierrez2019 (approved):
looks good to me

Issue #10230, opened 2024-09-05
List - dynamically changing list-items within a list causes existing filter and new list-items to be out of sync


iyouth23:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When a filter is applied to a `calcite-list` and the `calcite-list-items` within that list are modified, the resulting list-items that show up do not match what is expected based on the `filterText`.

### Expected Behavior

The filter and the list-items within a `calcite-list` should always be in sync.

### Reproduction Sample

https://codepen.io/iyouth66/pen/LYKMpbQ?editors=1000

### Reproduction Steps

- Open this codepen: https://codepen.io/iyouth66/pen/LYKMpbQ?editors=1000
- Click "List 1" > a `calcite-list` should populate with three list-items > Now filter by "Avenir" > The list items filter properly so that only two items are showing > Now click "List 2", which changes the content of the calcite-list to four different list-items > **the items don't show up** > Now change "Avenir" to "Aveni" to trigger a filter change > The filtered items show up again.

### Reproduction Version

Not a regression

### Relevant Info

cc @driskull

### Priority impact

impact - p1 - need for current milestone

### Impact

Affects our font-picker in ArcGIS Online, which handles two filterable lists with unique content.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Online

geospatialem:
Verified with the Chromatic build:

![verify-10230](https://github.com/user-attachments/assets/86b86992-f703-4759-988c-d90231b117a9)

PR #10342, opened 2024-09-18
fix(list, filter): fix sync between list items and filtered data

driskull:
**Related Issue:** #10230

## Summary

- updates Filter to emit `calciteFilterChange` when there are no `filteredItems`
  - Previously, when text was typed into the filter component but there were no `items`, no event would be emitted.
  - The doc says `Fires when the filter text changes.`. However, this was not always the case if items were not present.
  - This was problematic because you wouldn't get an event for filter text changing unless there were items populated.
  - If items were added later after a text input, you wouldn't have the correct text to filter against.
-  updates the list to filter when list items are slotted in
- add test


packages/calcite-components/src/components/filter/filter.tsx
```diff
@@ -167,9 +167,7 @@ export class Filter
 
   async componentWillLoad(): Promise<void> {
     setUpLoadableComponent(this);
-    if (this.items.length) {
-      this.updateFiltered(filter(this.items, this.value, this.filterProps));
-    }
+    this.updateFiltered(filter(this.items ?? [], this.value, this.filterProps));
```
jcfranco, line 170: Should we check for `value` presence before triggering the filter change event? Emitting an event on load when there are no items nor filter value (noop) seems odd to me.

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -844,39 +850,50 @@ export class List
...
+        this.dataForFilter = this.getItemData();
 
-    this.listItems = items;
-    if (this.filterEnabled) {
-      this.dataForFilter = this.getItemData();
-      if (this.filterEl) {
-        this.filterEl.items = this.dataForFilter;
+        if (filterEl) {
+          filterEl.items = this.dataForFilter;
+          this.filterAndUpdateData();
```
jcfranco, line 884: Does this need to be awaited on to avoid synchronization issues?

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -992,7 +1009,7 @@ export class List
 
     parentEl.insertBefore(dragEl, referenceEl);
 
-    this.updateListItems();
+    this.updateListItems({ emitFilterChange: false, performFilter: false });
```
jcfranco, line 1012: Can we remove any option objects that are passing the defaults?

jcfranco (changes requested):
Looking good, @driskull! 😎 

Could you also update `filter`'s E2E test suite to add coverage for these updates?

packages/calcite-components/src/components/list/list.tsx
driskull: @jcfranco It doesn't. The filter component will fire an event once the filter actually occurs.

This is part of the oddness of having a filter component that does the filtering. The syncing is odd.

packages/calcite-components/src/components/filter/filter.tsx
driskull: @jcfranco the problem with this is that when the user clears the filter, the value will be an empty string but we do want to fire the event in this scenario.

I could check to make sure filter is not null or undefined. However, the default filter component value is an empty string.

driskull: I don't think it would emit on load. The `updateFiltered` call on `componentWillLoad` doesn't emit.

jcfranco: Oh, I see. It's because the [`emit` argument is falsy](https://github.com/Esri/calcite-design-system/blob/1cb3d9055f0f4d36effd83c1543e1baae78c275f/packages/calcite-components/src/components/filter/filter.tsx#L265) when there is no filter value. Ignore my comment then. Thanks for checking!

driskull:
@jcfranco ready for review again.

jcfranco (approved):
📝🔄🔍✅

PR #10344, opened 2024-09-18
refactor(tests): reduce TypeScript errors

maxpatiiuk:
Run the migration script on tests.
This PR includes some small fixups that affect TypeScript typings to reduce number of TypeScript errors after the codemod.
With these changes, there are only 57 TypeScript error in the entire calcite-components package after the codemod!

### DIFF
--- a/packages/calcite-components/src/components/chip-group/chip-group.e2e.ts
+++ b/packages/calcite-components/src/components/chip-group/chip-group.e2e.ts
@@ -520,7 +520,7 @@ describe("calcite-chip-group", () => {
       expect(chipSelectSpy1).toHaveReceivedEventTimes(0);
       expect(chipSelectSpy2).toHaveReceivedEventTimes(0);
 
-      await chip5.setAttribute("selected", true);
+      await chip5.toggleAttribute("selected", true);
       await page.waitForChanges();
       expect(chipGroupSelectSpy).toHaveReceivedEventTimes(0);
       expect(chipSelectSpy1).toHaveReceivedEventTimes(0);
@@ -569,7 +569,7 @@ describe("calcite-chip-group", () => {
       expect(await element.getProperty("selectedItems")).toHaveLength(0);
       await selectedItemAsserter([]);
 
-      chip5.setAttribute("selected", true);
+      chip5.toggleAttribute("selected", true);
       await page.waitForChanges();
       expect(chipGroupSelectSpy).toHaveReceivedEventTimes(0);
       expect(chipSelectSpy1).toHaveReceivedEventTimes(0);
@@ -577,7 +577,7 @@ describe("calcite-chip-group", () => {
       expect(await element.getProperty("selectedItems")).toHaveLength(1);
       await selectedItemAsserter([chip5.id]);
 
-      chip4.setAttribute("selected", true);
+      chip4.toggleAttribute("selected", true);
       await page.waitForChanges();
       expect(chipGroupSelectSpy).toHaveReceivedEventTimes(0);
       expect(chipSelectSpy1).toHaveReceivedEventTimes(0);
@@ -617,7 +617,7 @@ describe("calcite-chip-group", () => {
       expect(chipSelectSpy1).toHaveReceivedEventTimes(0);
       expect(chipSelectSpy2).toHaveReceivedEventTimes(0);
 
-      chip5.setAttribute("selected", true);
+      chip5.toggleAttribute("selected", true);
       await page.waitForChanges();
       expect(chipGroupSelectSpy).toHaveReceivedEventTimes(0);
       expect(chipSelectSpy1).toHaveReceivedEventTimes(0);
--- a/packages/calcite-components/src/components/combobox/combobox.e2e.ts
+++ b/packages/calcite-components/src/components/combobox/combobox.e2e.ts
@@ -1632,7 +1632,7 @@ describe("calcite-combobox", () => {
     let element: E2EElement;
     let comboboxItem: E2EElement;
     let itemNestedLi: E2EElement;
-    let closeEvent: Promise<void>;
+    let closeEvent: Promise<unknown>;
 
     beforeEach(async () => {
       page = await newE2EPage();
--- a/packages/calcite-components/src/components/list-item-group/list-item-group.tsx
+++ b/packages/calcite-components/src/components/list-item-group/list-item-group.tsx
@@ -49,7 +49,7 @@ export class ListItemGroup implements InteractiveComponent {
    * Fires when changes occur in the default slot, notifying parent lists of the changes.
    */
   @Event({ cancelable: false })
-  calciteInternalListItemGroupDefaultSlotChange: EventEmitter<DragEvent>;
+  calciteInternalListItemGroupDefaultSlotChange: EventEmitter<void>;
 
   // --------------------------------------------------------------------------
   //
--- a/packages/calcite-components/src/components/list-item/list-item.tsx
+++ b/packages/calcite-components/src/components/list-item/list-item.tsx
@@ -304,7 +304,7 @@ export class ListItem
 
   @Listen("calciteInternalListItemGroupDefaultSlotChange")
   @Listen("calciteInternalListDefaultSlotChange")
-  handleCalciteInternalListDefaultSlotChanges(event: CustomEvent<void>): void {
+  handleCalciteInternalListDefaultSlotChanges(event: CustomEvent): void {
     event.stopPropagation();
     this.handleOpenableChange(this.defaultSlotEl);
   }
--- a/packages/calcite-components/src/components/list-item/list-item.tsx
+++ b/packages/calcite-components/src/components/list-item/list-item.tsx
@@ -304,7 +304,7 @@ export class ListItem
 
   @Listen("calciteInternalListItemGroupDefaultSlotChange")
   @Listen("calciteInternalListDefaultSlotChange")
-  handleCalciteInternalListDefaultSlotChanges(event: CustomEvent): void {
+  handleCalciteInternalListDefaultSlotChanges(event: CustomEvent<void>): void {
     event.stopPropagation();
     this.handleOpenableChange(this.defaultSlotEl);
   }
@@ -710,7 +710,7 @@ export class ListItem
   //
   // --------------------------------------------------------------------------
 
-  private dragHandleSelectedChangeHandler = (event: CustomEvent): void => {
+  private dragHandleSelectedChangeHandler = (event: CustomEvent<void>): void => {
     this.dragSelected = (event.target as HTMLCalciteHandleElement).selected;
     this.calciteListItemDragHandleChange.emit();
     event.stopPropagation();
--- a/packages/calcite-components/src/components/loader/loader.tsx
+++ b/packages/calcite-components/src/components/loader/loader.tsx
@@ -90,9 +90,9 @@ export class Loader implements LocalizedComponent {
     return (
       <Host
         aria-label={label}
-        aria-valuemax={isDeterminate ? 100 : undefined}
-        aria-valuemin={isDeterminate ? 0 : undefined}
-        aria-valuenow={isDeterminate ? valueNow : undefined}
+        aria-valuemax={isDeterminate ? "100" : undefined}
+        aria-valuemin={isDeterminate ? "0" : undefined}
+        aria-valuenow={isDeterminate ? valueNow.toString() : undefined}
         id={id}
         role="progressbar"
       >
--- a/packages/calcite-components/src/components/segmented-control/segmented-control.e2e.ts
+++ b/packages/calcite-components/src/components/segmented-control/segmented-control.e2e.ts
@@ -292,7 +292,7 @@ describe("calcite-segmented-control", () => {
 
       async function cycleThroughItemsAndAssertValue(keys: "left-right" | "up-down"): Promise<void> {
         const [moveBeforeArrowKey, moveAfterArrowKey] =
-          keys === "left-right" ? ["ArrowLeft", "ArrowRight"] : ["ArrowUp", "ArrowDown"];
+          keys === "left-right" ? (["ArrowLeft", "ArrowRight"] as const) : (["ArrowUp", "ArrowDown"] as const);
 
         await element.press(moveAfterArrowKey);
         await page.waitForChanges();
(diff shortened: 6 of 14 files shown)


packages/calcite-components/src/components/chip-group/chip-group.e2e.ts
```diff
@@ -520,7 +520,7 @@ describe("calcite-chip-group", () => {
       expect(chipSelectSpy1).toHaveReceivedEventTimes(0);
       expect(chipSelectSpy2).toHaveReceivedEventTimes(0);
 
-      await chip5.setAttribute("selected", true);
+      await chip5.toggleAttribute("selected", true);
```
maxpatiiuk, line 523: This is a boolean attribute
Before it was being set here to `selected="true"`

packages/calcite-components/src/components/combobox/combobox.e2e.ts
```diff
@@ -1632,7 +1632,7 @@ describe("calcite-combobox", () => {
     let element: E2EElement;
     let comboboxItem: E2EElement;
     let itemNestedLi: E2EElement;
-    let closeEvent: Promise<void>;
+    let closeEvent: Promise<unknown>;
```
maxpatiiuk, line 1635: Stencil's waitForEvent has return type Promise<any> so this worked.
In Lumina, the return type is Promise<SerializedEvent>.

(at runtime, Stencil's waitForEvent also returned Promise<SerializedEvent>, but they didn't publicly expose the SerializedEvent type)

packages/calcite-components/src/components/list-item-group/list-item-group.tsx
```diff
@@ -49,7 +49,7 @@ export class ListItemGroup implements InteractiveComponent {
    * Fires when changes occur in the default slot, notifying parent lists of the changes.
    */
   @Event({ cancelable: false })
-  calciteInternalListItemGroupDefaultSlotChange: EventEmitter<DragEvent>;
+  calciteInternalListItemGroupDefaultSlotChange: EventEmitter<void>;
```
maxpatiiuk, line 52: it didn't actually emit this type, and no one was listening for it to emit this type

packages/calcite-components/src/components/loader/loader.tsx
```diff
@@ -90,9 +90,9 @@ export class Loader implements LocalizedComponent {
     return (
       <Host
         aria-label={label}
-        aria-valuemax={isDeterminate ? 100 : undefined}
-        aria-valuemin={isDeterminate ? 0 : undefined}
-        aria-valuenow={isDeterminate ? valueNow : undefined}
+        aria-valuemax={isDeterminate ? "100" : undefined}
```
maxpatiiuk, line 93: Stencil's `aria-valuemax` type is `string | number`, and at runtime, the DOM stringifies numbers automatically, so this was good.
However, the TypeScript's type for HTMLElement.ariaValuemax is `string`. Since this code is converted by the codemod to `this.el.ariaValuemax = ...`, we need to be setting strings here to avoid type error.

packages/calcite-components/src/components/segmented-control/segmented-control.e2e.ts
```diff
@@ -292,7 +292,7 @@ describe("calcite-segmented-control", () => {
 
       async function cycleThroughItemsAndAssertValue(keys: "left-right" | "up-down"): Promise<void> {
         const [moveBeforeArrowKey, moveAfterArrowKey] =
-          keys === "left-right" ? ["ArrowLeft", "ArrowRight"] : ["ArrowUp", "ArrowDown"];
+          keys === "left-right" ? (["ArrowLeft", "ArrowRight"] as const) : (["ArrowUp", "ArrowDown"] as const);
```
maxpatiiuk, line 295: Lumina's typings for `await element.press(moveAfterArrowKey);` are more strict

packages/calcite-components/src/components/table/table.tsx
```diff
@@ -250,11 +250,11 @@ export class Table implements LocalizedComponent, LoadableComponent, T9nComponen
   }
 
   @Listen("calciteInternalTableRowFocusRequest")
-  calciteInternalTableRowFocusEvent(event: TableRowFocusEvent): void {
-    const cellPosition = event["detail"].cellPosition;
-    const rowPos = event["detail"].rowPosition;
-    const destination = event["detail"].destination;
-    const lastCell = event["detail"].lastCell;
+  calciteInternalTableRowFocusEvent(event: CustomEvent<TableRowFocusEvent>): void {
```
maxpatiiuk, line 253: looks like a mistake in typing.
`TableRowFocusEvent` is the payload type, not event type

packages/calcite-components/src/components/text-area/text-area.e2e.ts
```diff
@@ -150,7 +150,7 @@ describe("calcite-text-area", () => {
     await page.setContent("<calcite-text-area></calcite-text-area>");
 
     const element = await page.find("calcite-text-area");
-    element.setAttribute("max-length", 5);
+    element.setAttribute("max-length", "5");
```
maxpatiiuk, line 153: Stencil's E2EElement.setAttribute had type `(string, any)=>void`
In Lumina, it is `(string, string)=>void` so that it matches the HTMLElement.setAttribute type.
I want to align us closer with DOM typings so that migrating to Vitest browser mode is easier

packages/calcite-components/src/components/time-picker/time-picker.tsx
```diff
@@ -202,7 +202,7 @@ export class TimePicker
   /**
    * @internal
    */
-  @Event({ cancelable: false }) calciteInternalTimePickerChange: EventEmitter<string>;
+  @Event({ cancelable: false }) calciteInternalTimePickerChange: EventEmitter<void>;
```
maxpatiiuk, line 205: never actually included a `string` payload type, and no one was listening for it

packages/calcite-components/src/utils/dom.spec.ts
```diff
@@ -665,8 +665,8 @@ describe("dom", () => {
 
     let element: HTMLDivElement;
     let dispatchTransitionEvent: TransitionEventDispatcher;
-    let onStartCallback: jest.Mock<any, any, any>;
-    let onEndCallback: jest.Mock<any, any, any>;
+    let onStartCallback: jest.Mock;
```
maxpatiiuk, line 668: With this change, the type would be equivalent both in Jest and Vite.
Otherwise it was causing issues because Vite's Mock has only 2 type arguments

packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -304,7 +304,7 @@ export class ListItem
 
   @Listen("calciteInternalListItemGroupDefaultSlotChange")
   @Listen("calciteInternalListDefaultSlotChange")
-  handleCalciteInternalListDefaultSlotChanges(event: CustomEvent<void>): void {
+  handleCalciteInternalListDefaultSlotChanges(event: CustomEvent): void {
```
jcfranco, line 307: Do Lumina's type set `CustomEvent.detail`'s' type as `void` by default?

packages/calcite-components/src/utils/interactive.tsx
```diff
@@ -111,7 +111,7 @@ function removeInteractionListeners(element: HTMLElement): void {
   );
 }
 
-export interface InteractiveContainerProps extends JSXAttributes {
+export interface InteractiveContainerProps extends JSXAttributes<HTMLDivElement> {
```
jcfranco, line 114: This could stay the same as we're not using any props from a particular HTML element.

Applies to `XButtonProps` too.

jcfranco (approved):
Other than a few comments, this is looking great! Thanks, @maxpatiiuk!

packages/calcite-components/src/components/list-item/list-item.tsx
maxpatiiuk: yes!

to be more precise:
- default payload type is `undefined`
- it is a typescript error to not provide a payload to the `.emit()` function if the event has non-`undefined` payload type
- the codemod will update `EventEmitter<void>` to `EventEmitter` as needed

packages/calcite-components/src/utils/interactive.tsx
maxpatiiuk: Without the change in XButtonProps, I was getting a type error here:

https://github.com/Esri/calcite-design-system/blob/0b03a578db450106360b074dad985a40582a6220/packages/calcite-components/src/components/functional/XButton.tsx#L31

The issue is that in Lumina JSX, the ref prop can be assigned either `(element: HTMLButtonElement) => void` or `{ value: HTMLButtonElement }` (like a RefObject type in React).

The RefObject type is causing issues.
By default `JSXAttributes` interface is assuming `HTMLElement` type.
That means the ref prop type in the `<XButton>` is `RefObject<HTMLElement>`, which is not assignable to the `RefObject<HTMLButtonElement>` in the `<button>`.

The same issue is present in React since they also have object refs. Not an issue in Stencil since they don't have object refs, only callback refs.

---

The change in `<InteractiveContainer>` is not necessary, but I done it for consistency, and to avoid issues if someone does add a ref prop to it in the future.

packages/calcite-components/src/components/list-item/list-item.tsx
maxpatiiuk: edit:
Regarding listening to events, `CustomEvent.detail` is still `any` since that's coming from TypeScript.
I also see that you used `CustomEvent<void>` in many places, so I will revert this place to use `CustomEvent<void>` for consistency

maxpatiiuk:
Looks like GitHub action was stuck on E2E tests for 6 hours and then timed out:

https://github.com/Esri/calcite-design-system/actions/runs/10933000573/job/30350737756

Triggering E2E re-run completed successfully 🤔

packages/calcite-components/src/utils/interactive.tsx
jcfranco: I could be misunderstanding, but it seems this change is conflating the prop types with the ref type. These can be different depending on the implementation, right? 

It doesn't seem like we're using `XButton`'s `ref` property, so we could alternatively drop it for the time being.

maxpatiiuk: > I could be misunderstanding, but it seems this change is conflating the prop types with the ref type

`JSXAttributes` represents the common props type, including the `ref` prop.
The generic parameter you pass to `JSXAttributes` (eg `JSXAttributes<HTMLDivElement>`) dictates the type used in the `ref`

> It doesn't seem like we're using XButton's ref property, so we could alternatively drop it for the time being.

Sounds good, I will go ahead and remove it.

packages/calcite-components/src/tests/commonTests/t9n.ts
```diff
@@ -82,7 +82,7 @@ export async function t9n(componentTestSetup: ComponentTestSetup): Promise<void>
             return new Response(new Blob([JSON.stringify(fakeEsMessages, null, 2)], { type: "application/json" }));
           }
 
-          return orig.call(input, init);
+          return orig.call(window, input, init);
```
maxpatiiuk, line 85: First argument to `.call` is `this`.
Not sure why this worked before in Stencil, but this causes issues after migration.

packages/calcite-components/src/utils/interactive.tsx
jcfranco: You're right. I was thinking about other use cases where we might want to provide a `ref` to another part of the functional component, but this would probably be better off as a separate property in the props interface.

jcfranco (approved):
👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤
👨‍🎤🤘👨‍🎤👨‍🎤👨‍🎤🤘👨‍🎤👨‍🎤🤘🤘👨‍🎤👨‍🎤🤘🤘🤘👨‍🎤👨‍🎤🤘🤘🤘👨‍🎤🤘🤘🤘🤘👨‍🎤🤘👨‍🎤
👨‍🎤🤘🤘👨‍🎤👨‍🎤🤘👨‍🎤🤘👨‍🎤👨‍🎤🤘👨‍🎤👨‍🎤🤘👨‍🎤👨‍🎤🤘👨‍🎤👨‍🎤👨‍🎤👨‍🎤🤘👨‍🎤👨‍🎤👨‍🎤👨‍🎤🤘👨‍🎤
👨‍🎤🤘👨‍🎤🤘👨‍🎤🤘👨‍🎤🤘👨‍🎤👨‍🎤🤘👨‍🎤👨‍🎤🤘👨‍🎤👨‍🎤🤘👨‍🎤👨‍🎤👨‍🎤👨‍🎤🤘🤘🤘👨‍🎤👨‍🎤🤘👨‍🎤
👨‍🎤🤘👨‍🎤👨‍🎤🤘🤘👨‍🎤🤘👨‍🎤👨‍🎤🤘👨‍🎤👨‍🎤🤘👨‍🎤👨‍🎤🤘👨‍🎤👨‍🎤👨‍🎤👨‍🎤🤘👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤
👨‍🎤🤘👨‍🎤👨‍🎤👨‍🎤🤘👨‍🎤👨‍🎤🤘🤘👨‍🎤👨‍🎤🤘🤘🤘👨‍🎤👨‍🎤🤘🤘🤘👨‍🎤🤘🤘🤘🤘👨‍🎤🤘👨‍🎤
👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤👨‍🎤

Issue #10353, opened 2024-09-19
[Action-Pad] potentially unused mutation observer


maxpatiiuk:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Action-pad component creates a mutation observer:

https://github.com/Esri/calcite-design-system/blob/6f42a868f5cc65ae8a7c89f22ad9057b90fe4bca/packages/calcite-components/src/components/action-pad/action-pad.tsx#L152C3-L154

However once created, that observer is not used - `observer()` or `unobserve()` is never called on it.

If I am reading the code right, this observer has no effect.

### Expected Behavior

Either observer is supposed to be used, and it's a bug that it's not used, or it should be removed as dead code.

For an example of mutation observer being used, see how it's used in Action:
- https://github.com/Esri/calcite-design-system/blob/6f42a868f5cc65ae8a7c89f22ad9057b90fe4bca/packages/calcite-components/src/components/action/action.tsx#L177
- https://github.com/Esri/calcite-design-system/blob/6f42a868f5cc65ae8a7c89f22ad9057b90fe4bca/packages/calcite-components/src/components/action/action.tsx#L194

### Reproduction Sample

-

### Reproduction Steps

-

### Reproduction Version

v2.12.2

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

driskull:
Seems like it should be used in order to update newly slotted `calcite-action-group` elements with the layout of them.

maxpatiiuk:
Looks like popover has the same issue:

https://github.com/Esri/calcite-design-system/blob/4160af11ccb6bfb79314720e6f53b811f1cc6cb4/packages/calcite-components/src/components/popover/popover.tsx#L252

driskull:
I think we need an eslint test to make sure mutation observers are observed and disconnected within a file :) cc @jcfranco

maxpatiiuk:
> I think we need an eslint rule to make sure mutation observers are observed and disconnected within a file :) cc @jcfranco

It might be better to just create a controller for that - the controller will take care of observing connected/disconnect.

[...]

PR #10355, opened 2024-09-19
fix(action-pad): update layout on action-group elements slotted after initialization

driskull:
**Related Issue:** #10353

## Summary

- update layout on action-group elements slotted after initialization
- add test


maxpatiiuk:
Actually, same bug in popover too:
https://github.com/Esri/calcite-design-system/blob/4160af11ccb6bfb79314720e6f53b811f1cc6cb4/packages/calcite-components/src/components/popover/popover.tsx#L252C3-L252C12

anveshmekala (approved):
🥇

PR #10352, opened 2024-09-19
refactor: do not use Host in functional components

maxpatiiuk:
Lumina does not have a `<Host>` element, but there is an equivalent syntax.
The codemod can take care of migrating to the alternative syntax, as long as the `<Host>` element appears in the `render()` methods.
This PR moves the `<Host>` element from the `<List>` functional component into the parent `render()` methods.
With this change, all `<Host>` elements in Calcite's codebase are migrated automatically.

PR #10356, opened 2024-09-19
refactor(components): reduce post-migration TypeScript errors

maxpatiiuk:
Minor refactor - some HTML attributes are reflected to properties.
Thus, we can access properties instead of calling getAttribute/hasAttribute.
Properties are also more type-safe (where as getAttribute string parameter does not have type-checking)

And other minor changes.

### DIFF
--- a/packages/calcite-components/src/components/dialog/dialog.tsx
+++ b/packages/calcite-components/src/components/dialog/dialog.tsx
@@ -255,7 +255,6 @@ export class Dialog
             [CSS.containerOpen]: opened,
             [CSS.containerEmbedded]: this.embedded,
           }}
-          ref={this.setContainerEl}
         >
           {this.modal ? (
             <calcite-scrim class={CSS.scrim} onClick={this.handleOutsideClose} />
@@ -363,8 +362,6 @@ export class Dialog
   openTransitionProp = "opacity";
 
   transitionEl: HTMLDivElement;
-
-  containerEl: HTMLDivElement;
 
   focusTrap: FocusTrap;
 
@@ -763,10 +760,6 @@ export class Dialog
         };
     }
   }
-
-  private setContainerEl = (el: HTMLDivElement): void => {
-    this.containerEl = el;
-  };
 
   private setTransitionEl = (el: HTMLDivElement): void => {
     this.transitionEl = el;
--- a/packages/calcite-components/src/components/functional/Heading.tsx
+++ b/packages/calcite-components/src/components/functional/Heading.tsx
@@ -17,7 +17,7 @@ export const Heading: FunctionalComponent<HeadingProps> = (props, children): VNo
   delete props.level;
 
   return (
-    <HeadingTag class={props.class} key={props.key} level={props.level}>
+    <HeadingTag class={props.class} key={props.key}>
       {children}
     </HeadingTag>
   );
--- a/packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
+++ b/packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
@@ -774,9 +774,9 @@ export class InputDatePicker
 
   @State() private localeData: DateLocaleData;
 
-  private startInput: HTMLCalciteInputElement;
+  private startInput: HTMLCalciteInputTextElement;
 
-  private endInput: HTMLCalciteInputElement;
+  private endInput: HTMLCalciteInputTextElement;
 
   private floatingEl: HTMLDivElement;
 
@@ -884,11 +884,11 @@ export class InputDatePicker
     syncHiddenFormInput("date", this, input);
   }
 
-  setStartInput = (el: HTMLCalciteInputElement): void => {
+  setStartInput = (el: HTMLCalciteInputTextElement): void => {
     this.startInput = el;
   };
 
-  setEndInput = (el: HTMLCalciteInputElement): void => {
+  setEndInput = (el: HTMLCalciteInputTextElement): void => {
     this.endInput = el;
   };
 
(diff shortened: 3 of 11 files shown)


packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -764,10 +761,6 @@ export class Dialog
     }
   }
 
-  private setContainerEl = (el: HTMLDivElement): void => {
```
maxpatiiuk, line 767: Never read, thus removed

packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -774,9 +774,9 @@ export class InputDatePicker
 
   @State() private localeData: DateLocaleData;
 
-  private startInput: HTMLCalciteInputElement;
+  private startInput: HTMLCalciteInputTextElement;
```
maxpatiiuk, line 777: This was causing a type error in Lumina - the ref is assigned to calcite-input-text, but the type here was calcite-input.

packages/calcite-components/src/components/input-number/input-number.tsx
```diff
@@ -1095,8 +1095,8 @@ export class InputNumber
         autofocus={this.el.autofocus ? true : null}
         defaultValue={this.defaultValue}
         disabled={this.disabled ? true : null}
-        enterKeyHint={this.el.enterKeyHint || this.el.getAttribute("enterkeyhint")}
-        inputMode={this.el.inputMode || this.el.getAttribute("inputmode") || "decimal"}
+        enterKeyHint={this.el.enterKeyHint}
```
maxpatiiuk, line 1098: `enterkeyhint` attribute reflects to `enterKeyHint` propertie.
Verified in Chrome, Firefox and Safari.

Same for others

packages/calcite-components/src/components/input/input.tsx
```diff
@@ -1171,9 +1171,9 @@ export class Input
     const prefixText = <div class={CSS.prefix}>{this.prefixText}</div>;
     const suffixText = <div class={CSS.suffix}>{this.suffixText}</div>;
 
-    const autofocus = this.el.autofocus || this.el.hasAttribute("autofocus") ? true : null;
-    const enterKeyHint = this.el.enterKeyHint || this.el.getAttribute("enterkeyhint");
-    const inputMode = this.el.inputMode || this.el.getAttribute("inputmode");
+    const autofocus = this.el.autofocus ? true : null;
```
maxpatiiuk, line 1174: (optional) This would be even shorter, but less readable:

```suggestion
    const autofocus = this.el.autofocus || null;
```

packages/calcite-components/src/components/input-number/input-number.tsx
maxpatiiuk: EDIT: that is true for `<input>` elements.
it is not true for custom elements.
reverting this change

packages/calcite-components/src/components/tooltip/tooltip.tsx
```diff
@@ -146,8 +146,6 @@ export class Tooltip implements FloatingUIComponent, OpenCloseComponent {
 
   guid = `calcite-tooltip-${guid()}`;
 
-  hasLoaded = false;
```
maxpatiiuk, line 149: never read

packages/calcite-components/src/components/menu/menu.tsx
```diff
@@ -228,7 +228,7 @@ export class CalciteMenu implements LocalizedComponent, T9nComponent, LoadableCo
   }
 
   private getEffectiveRole(): string {
-    return this.el.getAttribute("role") || "menubar";
```
maxpatiiuk, line 231: equivalent

jcfranco:
I think [Firefox ESR](https://whattrainisitnow.com/calendar/) prevents us from using ARIA-related properties (e.g., `ariaExpanded` was [added in FF 119](https://caniuse.com/?search=ariaexpanded)). 

@dasa does this seem accurate to you?

Sidebar: noticed `value-list` (deprecated) currently [uses `ariaLabel`](https://github.com/Esri/calcite-design-system/blame/dev/packages/calcite-components/src/components/value-list/value-list.tsx#L489).

dasa:
We're planning to only support ESR 128 in our next release: https://whattrainisitnow.com/release/?version=esr

jcfranco (changes requested):
Thanks for the PR, @maxpatiiuk! 

Along the same lines of https://github.com/Esri/calcite-design-system/pull/10352, since some additional refactors were made to files unrelated to simplifying reflected attribute accesses, can you either:

1. separate non-Host-related refactors to a separate PR? 
2. update the PR title to convey all PR changes?

Thanks!

maxpatiiuk:
> I think [Firefox ESR](https://whattrainisitnow.com/calendar/) prevents us from using ARIA-related properties (e.g., ariaExpanded was [added in FF 119](https://caniuse.com/?search=ariaexpanded)).

> We're planning to only support ESR 128 in our next release: https://whattrainisitnow.com/release/?version=esr

@jcfranco so if the next release of JS API will drop Firefox 115 ESR support, is Calcite going to do the same for Calcite 3.0 release?
If so, are we ok to use aria properties in this PR or should I revert the change?

packages/calcite-components/src/components/functional/Heading.tsx
```diff
@@ -17,7 +17,7 @@ export const Heading: FunctionalComponent<HeadingProps> = (props, children): VNo
   delete props.level;
 
   return (
-    <HeadingTag class={props.class} key={props.key} level={props.level}>
```
maxpatiiuk, line 20: `level` prop doesn't seem to exist in the DOM?
was it used for styling or testing? if not, we should remove it

---

original discussion in https://github.com/Esri/calcite-design-system/pull/10352#discussion_r1767097375

jcfranco:
> If so, are we ok to use aria properties in this PR or should I revert the change?

Sorry for not clarifying earlier. We should be good to use ARIA properties if we install this after our September release (scheduled for this upcoming Monday).

jcfranco (approved):
Awesome stuff, @maxpatiiuk! Thanks! 🚀

PR #10357, opened 2024-09-19
fix(popover): update focus trap elements on mutation observer

driskull:
**Related Issue:** #10353

## Summary

- connect and disconnect mutation observer


macandcheese (approved):
🔭

PR #10358, opened 2024-09-19
feat(button): add component tokens

alisonailea:
**Related Issue:** #7180 
Blocked issues: #11210 

### Calcite Button

--calcite-button-background-color: Specifies the component's background color when appearance="solid" or appearance="outline-fill".
--calcite-button-border-color: Specifies the component's border color when it has appearance="outline" or appearance="outline-fill".
--calcite-button-corner-radius: Specifies the component's corner radius.
--calcite-button-text-color: Specifies the component's text color.


packages/calcite-components/src/components/button/button.scss
```diff
@@ -1,64 +1,144 @@
...
+ * @prop --calcite-button-text-color: Specifies the component's text color.
+ */
+
 :host {
   @apply inline-block w-auto align-middle;
-}
+  border-start-start-radius: var(
+    --calcite-button-corner-radius-start-start,
+    var(--calcite-button-corner-radius, var(--calcite-internal-button-corner-radius, var(--calcite-corner-radius)))
+  );
```
Elijbet, line 24: If there is a public `--calcite-button-corner-radius`, why do we have an internal version of it `var(--calcite-internal-button-corner-radius`? This came up in action PR, so I just wanted to clarify.

alisonailea: This allows the public tokens to still be changeable by users while allowing other components to make adjustments to the default

jcfranco:
For consistency and according to [the wiki](https://github.com/Esri/calcite-design-system/wiki/Component-Tokens#naming-pattern), `button` needs stateful tokens.

alisonailea:
> For consistency and according to [the wiki](https://github.com/Esri/calcite-design-system/wiki/Component-Tokens#naming-pattern), `button` needs stateful tokens.

That's not the case. State tokens are only for sub-components

jcfranco:
Synced up w/ @alisonailea and got clarification. The guidance doc has been updated, so we are good to proceed.

packages/calcite-components/src/components/button/button.scss
```diff
@@ -1,64 +1,144 @@
...
+ *
+ * @prop --calcite-button-background-color: Specifies the component's background color.
+ * @prop --calcite-button-border-color: Specifies the component's border color.
+ * @prop --calcite-button-corner-radius: Specifies the component's corner radius.
+ * @prop --calcite-button-corner-radius-start-start: Specifies the component's top-left corner radius. will fallback to --calcite-button-corner-radius.
+ * @prop --calcite-button-corner-radius-start-end: Specifies the component's top-right corner radius. Will fallback to --calcite-button-corner-radius.
+ * @prop --calcite-button-corner-radius-end-start: Specifies the component's bottom-left corner radius. Will fallback to --calcite-button-corner-radius.
+ * @prop --calcite-button-corner-radius-end-end: Specifies the component's bottom-right corner radius. Will fallback to --calcite-button-corner-radius.
+ * @prop --calcite-button-icon-color: Specifies the color of an icon in the component.
+ * @prop --calcite-button-loader-color: Specifies the component's loader color.
```
jcfranco, line 14: Question: Do we have a specific use case for a separate loader color? This may be outdated, but the advice from the [Guidelines for Granular Icons](https://www.figma.com/design/wTABvYi73K4qaOsfvQQu4j/Guidelines-for-Granular-Icon-Colors?node-id=0-1&node-type=canvas&m=dev) specifies button loader should share the same color as icon. cc @SkyeSeitz @ashetland

packages/calcite-components/src/components/button/button.scss
```diff
@@ -147,19 +433,11 @@
   a:not(.content--slotted),
   button:not(.content--slotted) {
     .icon--start + .icon--end {
-      margin-inline-start: var(--calcite-button-content-margin-internal);
+      margin-inline-start: var(--calcite-internal-button-content-margin);
```
jcfranco, line 436: Thanks for renaming these internal ones! ✨🏆✨

packages/calcite-components/src/components/button/button.scss
```diff
@@ -72,16 +152,222 @@
...
+    m-0
+    inline-flex
+    font-normal;
+  line-height: inherit;
+  color: var(--calcite-button-icon-color, currentColor);
+}
+
+// fab variants
+:host([round]) {
+  --calcite-internal-button-corner-radius: 50px;
```
jcfranco, line 166: Sidebar: Do we need to move fab-specific styles to `fab.scss` (similar to the overrides applied by `inline-editable.scss` below)? If so, we can tackle that separately.

packages/calcite-components/src/components/button/button.e2e.ts
```diff
@@ -572,16 +558,11 @@ describe("calcite-button", () => {
...
-
+      const buttonEl = await page.find("calcite-button >>> button");
       await buttonEl.hover();
       await page.waitForChanges();
-      buttonHoverStyle = await buttonEl.getComputedStyle();
-      expect(buttonHoverStyle.getPropertyValue("background-color")).toEqual(overrideStyle);
+      const buttonFocusStyle = await buttonEl.getComputedStyle();
+      expect(buttonFocusStyle.getPropertyValue("background-color")).toEqual(overrideStyle);
     });
   });
```
jcfranco, line 567: Per the [guidelines](https://github.com/Esri/calcite-design-system/wiki/tokens-component#e2e), can you add `themed` tests?

jcfranco (changes requested):
This is looking great, @alisonailea! 😎

packages/calcite-components/src/components/button/button.scss
alisonailea: I moved this over to FAB but kept the rest

packages/calcite-components/src/components/button/button.scss
alisonailea: this got messy so I ended but simplifying it back to just a border-radius on the :host

packages/calcite-components/src/components/button/button.scss
alisonailea: both icon-color and loader-color should inherit from text color

packages/calcite-components/src/components/fab/fab.scss
```diff
@@ -1,17 +1,46 @@
...
+:host(:active) {
+  calcite-button {
+    box-shadow:
+      0 2px 12px -4px rgba(var(--calcite-fab-shadow-color, var(--calcite-internal-fab-shadow-color)), 0.2),
+      0 2px 4px -2px rgba(var(--calcite-fab-shadow-color, var(--calcite-internal-fab-shadow-color)), 0.16);
   }
 }
 
 @include base-component();
+@include disabled();
```
jcfranco, line 46: Is this needed if the internal `calcite-button` includes it?

packages/calcite-components/src/custom-theme/button.ts
```diff
@@ -2,6 +2,13 @@ import { html } from "../../support/formatting";
 
 const buttonProps: [args: string, content: string] = ["", ""];
 
+export const buttonTokens = {
```
jcfranco, line 5: Can you also add a custom theme story for `fab` and update its demo page?

jcfranco (approved):
Nice! This should be good to merge once the remaining comments are addressed. Thanks, @alisonailea! ✨💪✨

packages/calcite-components/src/components/fab/fab.scss
jcfranco: Per our conversation, let's ignore this for now. I didn't notice this was already in the current styles and was just moved.

driskull:
@alisonailea for button, can we also add a token for setting the border width on each side? This would provide the ability to remove a border on one side of a button in order place buttons side by side.

alisonailea:
> @alisonailea for button, can we also add a token for setting the border width on each side? This would provide the ability to remove a border on one side of a button in order place buttons side by side.

@driskull you can now set the --calcite-button-border-width when the button props set a border

packages/calcite-components/src/components/button/button.scss
```diff
@@ -1,64 +1,90 @@
...
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-button-background-color: Specifies the component's background color when appearance="solid" or appearance="outline-fill".
+ * @prop --calcite-button-border-color: Specifies the component's border color when it has appearance="outline" or appearance="outline-fill".
+ * @prop --calcite-button-corner-radius: Specifies the component's corner radius.
+ * @prop --calcite-button-loader-color: Specifies the component's loader color.
+ * @prop --calcite-button-shadow-color: Specifies the component's box-shadow color.
+ * @prop --calcite-button-text-color: Specifies the component's text color.
+ * @prop --calcite-button-border-width: Specifies the component's border width.
```
macandcheese, line 12: For @driskull case - would this need to be inline-border-start-width / inline-border-end-width? 

I'd almost suggest marking those as internal as well - we may be able to support the "button group" use case in a different way down the line.

Issue #10307, opened 2024-09-13
[Action Pad] Horizontal action groups do not align


mpayson:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Action pad groups are slightly unaligned
<img width="266" alt="Screenshot 2024-09-13 at 3 38 08 PM" src="https://github.com/user-attachments/assets/489598b6-de41-41ee-b3eb-23f4a7ac6d89">

### Expected Behavior

Action pad groups are aligned

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/action-pad/

### Reproduction Steps

1. Go to sample
2. Change to "Action Pad with groups"
3. Change to horizontal layout
4. See issue

### Reproduction Version

2.12.2

### Regression?

2.11

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

geospatialem:
Potentially related to #10058

alisonailea:
Issue identified. There is an extra border-block-end. PR incomming.

geospatialem:
Verified with Chromatic build:
![image](https://github.com/user-attachments/assets/70d02c03-72c5-400f-8a0b-913b0e97c612)

PR #10359, opened 2024-09-19
fix(action-pad): fix horizontal action group alignment

alisonailea:
**Related Issue:** #10307

## Summary

Removes an extra border-block-end which was causing a layout bug


packages/calcite-components/src/components/action-pad/action-pad.stories.ts
```diff
@@ -42,6 +42,27 @@ export const simple = (args: ActionPadStoryArgs): string => html`
   </calcite-action-pad>
 `;
 
+export const withGroups_TestOnly = (): string =>
```
jcfranco, line 45: Note that new tests no longer need the `_TestOnly` suffix.

jcfranco (approved):

Issue #9071, opened 2024-04-05
Action: Create localized template for aria-label


driskull:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

aria-label is using a template for its string value here

https://github.com/Esri/calcite-design-system/blob/15ca10a2bbedc568951246ff57537907e257bbc2/packages/calcite-components/src/components/action/action.tsx#L302

### Expected Behavior

This should use a proper localized template

### Reproduction Sample

https://codepen.io/driskull/pen/NWmyeXw?editors=1000

### Reproduction Steps

1. open codepen
2. Inspect internal element with aria-label `aria-label="test (Indicator present)"`

### Reproduction Version

1.7.1

### Relevant Info

https://github.com/Esri/calcite-design-system/pull/9051#discussion_r1551861189

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

driskull:
@geospatialem can we move this issue until the string is available in all locales?

geospatialem:
> @geospatialem can we move this issue until the string is available in all locales?

Sure, we're currently set for October, would this fit the timeline, or do we need to push to November, @driskull?

driskull:
Just whenever the strings will be there.

geospatialem:
> Just whenever the strings will be there.

checked in with Anvesh and we should be on target for October, but can move this out to November if the translations are delayed.

driskull:
Sounds good!

geospatialem:
Verified in `3.0.0-next.51` with `action`'s `label`, `text`, and fallback of an empty string when neither attr/prop is specified.

PR #10360, opened 2024-09-19
chore(action): add new string for accessible indicator label

driskull:
**Related Issue:** #9071

## Summary

- check in new string

### DIFF
--- a/packages/calcite-components/src/components/action/action.tsx
+++ b/packages/calcite-components/src/components/action/action.tsx
@@ -298,7 +298,10 @@ export class Action
       buttonId,
       messages,
     } = this;
-    const ariaLabel = messages.indicatorLabel?.replace("{label}", label || text) ?? "";
+    const labelFallback = label || text;
+    const ariaLabel = indicator
+      ? messages.indicatorLabel?.replace("{label}", labelFallback)
+      : labelFallback;
 
     const buttonClasses = {
       [CSS.button]: true,
--- a/packages/calcite-components/src/components/action/action.tsx
+++ b/packages/calcite-components/src/components/action/action.tsx
@@ -298,10 +298,7 @@ export class Action
       buttonId,
       messages,
     } = this;
-    const labelFallback = label || text;
-    const ariaLabel = labelFallback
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/action/action.tsx
```diff
@@ -298,10 +298,10 @@ export class Action
       buttonId,
       messages,
     } = this;
-    const labelFallback = label || text;
-    const ariaLabel = labelFallback
-      ? `${labelFallback}${indicator ? ` (${messages.indicator})` : ""}`
-      : "";
+    const labelFallback = label || text || "";
+    const ariaLabel = indicator
+      ? messages.indicatorLabel?.replace("{label}", labelFallback)
```
anveshmekala: Don't think we can start using new message string before translations are back. This will remove `aria-label` for all non-english locales.

Ex: 
```
 <calcite-action icon="banana" scale="l" label="test" indicator text="test2" lang="bs"></calcite-action>
 ```
 
 ☝️  wouldn't add aria-label to button in shadowRoot.

Alternate: we can request a generate build which return english strings for all locales or wait until the translations are back before start using `indicatorLabel` string.

anveshmekala (changes requested):
Other than the comment LGTM!

anveshmekala:
Sidebar & Unrelated to issue:

@geospatialem could you please review if we need add `( )` in message string for accessible label. Can handle in separate issue if it is not required.

packages/calcite-components/src/components/action/action.tsx
driskull: Should we just add the string to the messages for now then and move this issue to another milestone? What would you recommend?

anveshmekala: >Should we just add the string to the messages for now then and move this issue to another milestone? What would you recommend?

Yup that works.

Second creating separate issue for strings.

anveshmekala (approved):
LGTM! 👍

geospatialem (approved):
:shipit:

PR #10365, opened 2024-09-20
ci: resolve husky pre-push and pre-commit errors

benelan:
## Summary

Resolve a few errors in our `pre-commit` and `pre-push` hooks:

1. The `support/updateStylelintCustomSassFunctions.ts` script was trying to open scss files in the root directory.

   ```text
   Error reading file: /home/jamin/dev/work/calcite-design-system/dev/ComponentWithSass.scss
   Error: ENOENT: no such file or directory, open '/home/jamin/dev/work/calcite-design-system/dev/ComponentWithSass.scss'
       at Object.readFileSync (node:fs:448:20)
       at <anonymous> (/home/jamin/dev/work/calcite-design-system/dev/support/updateStylelintCustomSassFunctions.ts:37:24)
       at Array.forEach (<anonymous>)
       at sassFiles (/home/jamin/dev/work/calcite-design-system/dev/support/updateStylelintCustomSassFunctions.ts:35:11)
       at Object.<anonymous> (/home/jamin/dev/work/calcite-design-system/dev/support/updateStylelintCustomSassFunctions.ts:63:1)
       at Module._compile (node:internal/modules/cjs/loader:1358:14)
       at Object.transformer (/home/jamin/dev/work/calcite-design-system/dev/node_modules/tsx/dist/register-C1urN2EO.cjs:2:1122)
       at Module.load (node:internal/modules/cjs/loader:1208:32)
       at Module._load (node:internal/modules/cjs/loader:1024:12)
       at cjsLoader (node:internal/modules/esm/translators:348:17) {
   errno: -2,
   code: 'ENOENT',
   syscall: 'open',
   path: '/home/jamin/dev/work/calcite-design-system/dev/ComponentWithSass.scss'
   }
   ```

2. The stylelint config file extension was incorrect.
   ```text
   Stylelint configuration updated successfully
   fatal: pathspec 'packages/calcite-components/.stylelintrc.json' did not match any files
   ```
3. The `pre-push` hook was only protecting `main` when it should protect `rc` and `dev` too.

4. [Husky requires scripts to be POSIX compliant](https://typicode.github.io/husky/how-to.html#bash), which means we can't use bash features.

   - Replace the `read` prompt's `-p` and `-n` flags with a printf and case statement
   - Replace double bracket conditionals used for pattern matching with `grep -qE` in the `pre-commit` hook

     ```text
     .husky/pre-commit: 40: [[: not found
     .husky/pre-commit: 16: [[: not found
     ```

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,32 +1,28 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 
 ensure_types_are_up_to_date() {
   types_path="packages/calcite-components/src/components.d.ts"
 
   if [ -n "$(git diff --name-only -- "$types_path")" ]; then
     echo "Automatically staging changes to \"$types_path\""
-    git add "$types_path"
+    git add "$types_path" >/dev/null 2>&1 || true
   fi
 }
 
 update_stylelint_config_if_sass_file_edited() {
-  staged_files="$(git diff --cached --name-only --diff-filter=ACM)"
+  staged_files="$(
+    git diff --cached --name-only --diff-filter=ACM -- packages/**/*.scss
+  )" 
 
-  for file in $staged_files; do
-    if [[ "$file" == *.scss ]]; then
-      npm run util:update-stylelint-custom-sass-functions
-      break
-    fi
-  done
-
-  if [ -n "$(git diff --name-only -- "packages/calcite-components/.stylelintrc.json")" ]; then
-    git add "packages/calcite-components/.stylelintrc.json"
+  if [ -n "$staged_files" ]; then
+    npm run util:update-stylelint-custom-sass-functions
+    git add "packages/calcite-components/.stylelintrc.cjs" >/dev/null 2>&1 || true
   fi
+
+  unset staged_files
 }
 
 check_ui_icon_name_consistency() {
-  svg_icon_dir="packages/calcite-ui-icons/icons"
-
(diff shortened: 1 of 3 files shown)


support/updateStylelintCustomSassFunctions.ts
```diff
@@ -15,7 +15,7 @@ function collectSassFiles(dir: string): string[] {
 
   try {
     fs.readdirSync(dir, { recursive: true, withFileTypes: true }).forEach(dirent => {
```
benelan, line 17: Do we need to recursively check files starting in the root directory or can we be more specific?

The only relevant directories are `packages/calcite-components/src` and `packages/calcite-design-system/dist/scss` right?

jcfranco: Is the suggestion to narrow the search scope mainly for performance reasons? I’d like to avoid an allowlist, so we don’t need to (remember to) update this script if our Sass file directories change.

.husky/pre-commit
```diff
@@ -1,60 +1,62 @@
...
+  )" 
 
-  for file in $staged_files; do
-    if [[ "$file" == *.scss ]]; then
-      npm run util:update-stylelint-custom-sass-functions
-      break
-    fi
-  done
-
-  if [ -n "$(git diff --name-only -- "packages/calcite-components/.stylelintrc.json")" ]; then
```
jcfranco, line 22: 🤦‍♂️ Thanks for fixing this one.

jcfranco (approved):
Thanks for fixing, @benelan!

Issue #9955, opened 2024-08-02
BUG-000169703 Segmented Control does not respect `scale` beyond first page load


fdeters:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Filed as `BUG-000169703` 🐛 

The `calcite-segmented-control-item` does not respect the `scale` property on the `calcite-segmented-control` if the element was rendered after the initial page load. Specifically, I'm seeing a `calcite-segmented-control` with `scale="s"` render as though it had `scale="m"`.

### Expected Behavior

I expect the `calcite-segmented-control` and `calcite-segmented-control-item` to respect the `scale` property at all times.

### Reproduction Samples

1. https://codepen.io/driskull/pen/WNqjrpq
2. https://gis.oregonmetro.gov/metromap

#### First sample
1. Open [the sample](https://codepen.io/driskull/pen/WNqjrpq)
3. Watch the component load on initialization
4. Observe the appended child item load in with a different scale than the parent

#### Second sample
Also, we can see this behavior on my team's live app:
1. Visit [gis.oregonmetro.gov/metromap](https://gis.oregonmetro.gov/metromap)
5. Once the page loads, you should see a modal with a legal disclaimer. The modal has a `calcite-segmented-control` in the bottom left that can be used to switch the language. It is rendering correctly with `scale="s"`.
![image](https://github.com/user-attachments/assets/d729b87e-01f2-4519-b12e-54fd64a6574c)
6. Click "I agree", then navigate to the "About" page using the nav in the top-right.
7. Go to the "Terms of Use" section using the tab navigation on the About page.
8. Click "View disclaimer" in the body of the Terms of Use section. This re-opens the same `calcite-modal` we saw in step 2. You can now see that the `calcite-segmented-control-item` is rendering as though it had `scale="m"`, even though it still has `scale="s"`.
![image](https://github.com/user-attachments/assets/36377510-7ebd-4ed5-8312-653e4675e74b)

Notice that the `calcite-segmented-control` visible in the site header never changes and always shows as `scale="s"`, like it should.

When I first encountered this bug, we were using that same `calcite-segmented-control` component in other areas of the application, and it exhibited the same issue.

### Reproduction Version

2.9.0

### Relevant Info

**OS:** Windows 10
**Browser:** Observed on recent versions of Chrome, Edge, and Firefox
[...]

PR #10368, opened 2024-09-20
fix(segmented-control): honor appearance, layout and scale when items are added after initialization

driskull:
**Related Issue:** #9955

## Summary

- Use slotchange event to get items and update them when necessary
- No need for the mutation observer
- add e2e test

### DIFF
--- a/packages/calcite-components/src/components/segmented-control/segmented-control.tsx
+++ b/packages/calcite-components/src/components/segmented-control/segmented-control.tsx
@@ -11,7 +11,7 @@ import {
   VNode,
   Watch,
 } from "@stencil/core";
-import { getElementDir, toAriaBoolean } from "../../utils/dom";
+import { getElementDir, slotChangeGetAssignedElements, toAriaBoolean } from "../../utils/dom";
 import {
   afterConnectDefaultValueSet,
   connectForm,
@@ -33,7 +33,6 @@ import {
   setUpLoadableComponent,
 } from "../../utils/loadable";
 import { Appearance, Layout, Scale, Status, Width } from "../interfaces";
-import { createObserver } from "../../utils/observers";
 import { Validation } from "../functional/Validation";
 import { IconNameOrString } from "../icon/interfaces";
 import { isBrowser } from "../../utils/browser";
@@ -98,7 +97,7 @@ export class SegmentedControl
 
   @Watch("value")
   valueHandler(value: string): void {
-    const items = this.getItems();
+    const { items } = this;
     items.forEach((item) => (item.checked = item.value === value));
   }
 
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/segmented-control/segmented-control.tsx
```diff
@@ -350,26 +350,40 @@ export class SegmentedControl
...
+      this.selectItem(lastChecked);
+    } else if (items[0]) {
+      items[0].tabIndex = 0;
+    }
   }
 
-  private getItems(): HTMLCalciteSegmentedControlItemElement[] {
-    return Array.from(this.el.querySelectorAll("calcite-segmented-control-item"));
+  private handleDefaultSlotChange = (event: Event): void => {
+    const items = slotChangeGetAssignedElements(event).filter(
```
jcfranco, line 366: According to [the doc](https://developers.arcgis.com/calcite-design-system/components/segmented-control/#slots), we only support `calcite-segmented-control-items`, so we shouldn't need to filter items.

jcfranco (approved):
🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊
🧊🍦🧊🧊🧊🍦🧊🍦🍦🍦🧊🧊🍦🍦🍦🧊🍦🍦🍦🍦🧊🍦🧊
🧊🍦🍦🧊🧊🍦🧊🧊🍦🧊🧊🍦🧊🧊🧊🧊🍦🧊🧊🧊🧊🍦🧊
🧊🍦🧊🍦🧊🍦🧊🧊🍦🧊🧊🍦🧊🧊🧊🧊🍦🍦🍦🧊🧊🍦🧊
🧊🍦🧊🧊🍦🍦🧊🧊🍦🧊🧊🍦🧊🧊🧊🧊🍦🧊🧊🧊🧊🧊🧊
🧊🍦🧊🧊🧊🍦🧊🍦🍦🍦🧊🧊🍦🍦🍦🧊🍦🍦🍦🍦🧊🍦🧊
🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊🧊

packages/calcite-components/src/components/segmented-control/segmented-control.tsx
driskull: I think we should leave this incase users slot tooltips, divs or (html comments?). Its safer to know we only want the segmented-control-items and this is the default slot. Otherwise there could potentially be a regression since we are currently querying items with querySelector.

Issue #7775, opened 2023-09-19
Epic: Component a11y consistency with `disabled` and `aria-disabled` attributes (2036570545)


dqateam:
**monday.com sync:** #18005777677

### Violation: 
 Ensure ARIA roles, states, and properties are valid. In each case both the `disabled` and `aria-disabled` attributes are added causing error reports across the design system. 

```jsx
<button
          ...
          aria-disabled={toAriaBoolean(disabled)}
          ...
          disabled={disabled}
```

### Fix
Calcite should get rid of the `aria-disabled` where `disabled` is present.

We could consider an eslint rule to prevent both `aria-disabled` and `disabled` from being set in the JSX per Matt's recommendation in the `list-item` issue https://github.com/Esri/calcite-design-system/issues/7796#issuecomment-1799429020.

### Components affected
- `action`
- `block`
- `dropdown`
- `split-button`
- `color-picker`
- `combobox`
- `fab`
- `input`
- `input-message`
- `input-date-picker`
- `list-item`
- `rating`
- `panel`
- `radio-button`
- `select`
- `slider`
- `switch`
- `stepper-item`
- `tabs`
- `tile`
- `tile-select`

![image](https://github.com/Esri/calcite-design-system/assets/5023024/2116a304-2a9b-419d-a85a-c95039b5503b)

### WCAG Reference: 
- [4.1.2 Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html) 

### Severity: 
 5 

### Media Type: 
 ARIA 

--- 
### Areas for Remediation: 
- Title: Action - Ensure ARIA roles, states, and properties are valid - (2036570545) 
- Module: Action 
**Pattern:** ARIA – Redundant aria-disabled Attribute]

**Issue**
There are elements with invalid ARIA attribute aria-disabled="true" as it already has native HTML disabled attribute.

**Occurrences**
04 Action
- <button> element present below the heading "Test 6 Disabled"
08 Block & Block Section
- <Div> element present below the heading "Test 5. Disabled Block"
09b Split Button
- <calcite-split-button> element present below "Test 3. Disabled Split Button"
10 Color Picker
- <calcite-color-picker> element present below "Test 2. Color Picker Disabled"
11 Combobox, Combobox Item Group and Combobox Item
- <calcite-combobox> present below the heading "Test 4. Disabled combobox".
- "Disabled first response item" available after activating combobox present below the heading "Test 7. Combobox groups and disabled items".
[...]

PR #10374, opened 2024-09-23
fix: remove aria-disabled from components where necessary

driskull:
**Related Issue:** #7775

## Summary

- remove setting `aria-disabled` where the `InteractiveContainer` will handle setting `aria-disabled`
- add `InteractiveContainer` to handle component
  -  picklist and valuelist also need this but they are deprecated so I didn't update them
- some components were setting `aria-disabled` without `toAriaBoolean` so they weren't working anyway `aria-disabled={this.disabled}`
- `date-picker-month-header` was not using `toAriaBoolean`
- update e2e test
- After this PR, we can evaluate if we need to update the `Interactive.tsx` helper.


packages/calcite-components/src/components/button/button.tsx
```diff
@@ -266,7 +265,6 @@ export class Button
     return (
       <InteractiveContainer disabled={this.disabled}>
         <Tag
-          aria-disabled={childElType === "a" ? toAriaBoolean(this.disabled || this.loading) : null}
```
driskull, line 269: Not needed because the `InteractiveContainer` will handle setting `aria-disabled`

driskull:
After this PR, the only remaining work would be to verify if we need to remove setting `aria-disabled` on the host with the `Interactive` helper.

packages/calcite-components/src/components/card/card.tsx
```diff
@@ -339,7 +339,6 @@ export class Card
         <InteractiveContainer disabled={this.disabled}>
           <div
             aria-checked={this.selectionMode !== "none" ? toAriaBoolean(this.selected) : undefined}
-            aria-disabled={this.disabled}
```
driskull, line 342: This one wasn't using `toAriaBoolean` so it wasn't working

packages/calcite-components/src/components/button/button.tsx
driskull: Need to discuss if `this.loading` should be setting `aria-disabled` or some other aria attribute? @geospatialem

packages/calcite-components/src/components/chip-group/chip-group.tsx
```diff
@@ -275,12 +275,7 @@ export class ChipGroup implements InteractiveComponent {
 
     return (
       <InteractiveContainer disabled={disabled}>
-        <div
-          aria-disabled={toAriaBoolean(disabled)}
```
driskull, line 279: Not needed because the InteractiveContainer will handle setting aria-disabled

packages/calcite-components/src/components/chip/chip.tsx
```diff
@@ -413,7 +413,6 @@ export class Chip
                 ? toAriaBoolean(this.selected)
                 : undefined
             }
-            aria-disabled={disableInteraction ? toAriaBoolean(disabled) : undefined}
```
driskull, line 416: This one was only setting `aria-disabled` when `disabled` was true anyway. Not sure why the `disabledInteraction` is being tested here unless we do this:

```tsx
 aria-disabled={disableInteraction ? toAriaBoolean(disableInteraction) : undefined}
```

packages/calcite-components/src/components/table-cell/table-cell.tsx
```diff
@@ -222,7 +222,6 @@ export class TableCell
       <Host>
         <InteractiveContainer disabled={this.disabled}>
           <td
-            aria-disabled={this.disabled}
```
driskull, line 225: Wasn't doing anything.

packages/calcite-components/src/components/button/button.tsx
driskull: Should this use `aria-busy` instead? https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-busy

packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
```diff
@@ -129,7 +129,7 @@ export class DatePickerMonthHeader {
     return (
       <Fragment>
         <a
-          aria-disabled={`${this.prevMonthDate.getMonth() === activeMonth}`}
+          aria-disabled={toAriaBoolean(this.prevMonthDate.getMonth() === activeMonth)}
```
jcfranco, line 132: This should be handled in https://github.com/Esri/calcite-design-system/pull/8402#discussion_r1774508844., which is targeted for the September release. Can we roll this back to avoid merge conflicts? cc @anveshmekala

packages/calcite-components/src/components/button/button.tsx
jcfranco: The way I'm reading it, `aria-busy` would go on the content that's being updated (triggered by the loading button), but I could be wrong.

jcfranco (approved):
Great catches, @driskull! ✨🎣✨

🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅
🏅🎣🏅🏅🏅🎣🏅🏅🎣🎣🏅🏅🎣🎣🎣🏅🏅🎣🎣🎣🏅🎣🎣🎣🎣🏅🎣🏅
🏅🎣🎣🏅🏅🎣🏅🎣🏅🏅🎣🏅🏅🎣🏅🏅🎣🏅🏅🏅🏅🎣🏅🏅🏅🏅🎣🏅
🏅🎣🏅🎣🏅🎣🏅🎣🏅🏅🎣🏅🏅🎣🏅🏅🎣🏅🏅🏅🏅🎣🎣🎣🏅🏅🎣🏅
🏅🎣🏅🏅🎣🎣🏅🎣🏅🏅🎣🏅🏅🎣🏅🏅🎣🏅🏅🏅🏅🎣🏅🏅🏅🏅🏅🏅
🏅🎣🏅🏅🏅🎣🏅🏅🎣🎣🏅🏅🎣🎣🎣🏅🏅🎣🎣🎣🏅🎣🎣🎣🎣🏅🎣🏅
🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅

driskull:
@geospatialem could you review these changes?

packages/calcite-components/src/components/button/button.tsx
geospatialem: @driskull What does the `Tag` refer to? When testing the component with `loading` the component is no longer keyboard accessible, and doesn't provide context that it's loading. It looks like the `disabled` attribute is applied, is that our intent?

geospatialem (commented):
One comment WRT `aria-busy` and if our intent is to have the component with the `disabled` attribute present when `loading`

packages/calcite-components/src/components/button/button.tsx
driskull: Tag revers to either a native button element or a anchor element. (`<button>` or `<a>`)

Currently, we have this as well:

```
disabled={childElType === "button" ? this.disabled || this.loading : null}
```

So maybe we shouldn't set it to disabled when loading?

geospatialem: > So maybe we shouldn't set it to disabled when loading?

I thought we had set this previously, but maybe it somehow got changed in the recent past? We added a [recommendation to the support recently](https://developers.arcgis.com/calcite-design-system/components/button/#recommendations) on our documentation page.

driskull: Yeah I would expect loading and disabled to be decoupled. It should be left up to the user to set both if needed.

Issue #10380, opened 2024-09-24
[Dialog / Panel] Add css properties for content-top and content-bottom slot spacing


macandcheese:
**monday.com sync:** #18048700771

### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

I’d like the ability to remove (or otherwise customize) the default padding of the content-top and content-bottom slots of Dialog. This could likely be useful in Panel as well, so adding them in both places seems expected. 

### Acceptance Criteria

A content-top-space and content-bottom-space css properties are added to Dialog / Panel. 

### Which Component

Dialog / Panel

### Example Use Case

As a designer, I’d like to remove the padding provided by default in these slots to achieve a specific UI.

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

DintaMel:
@driskull @matgalla Should the Flow Item token be `--calcite-flow-item-content-top-space` instead of `--calcite-flow-content-top-space`? 

I noticed all of the existing (non-deprecated) [styles on Flow Items](https://developers.arcgis.com/calcite-design-system/components/flow-item/#styles) also do not contain "item" in the token name either. [Stepper Item tokens](https://developers.arcgis.com/calcite-design-system/components/stepper-item/#styles) has lots of "item"-named tokens.

Following the [Token Naming conventions](https://developers.arcgis.com/calcite-design-system/foundations/tokens/usage/#naming), I would have thought that the "object" is "flow-item", not "flow".

driskull:
@DintaMel I beleive we had decided to not use `item` unless the token was only applicable on the item and necessary to have it. Otherwise, parent/child tokens can be the same. Please correct me if I'm wrong.

DintaMel:
Verified in 5.1.0-next.8 in Codepen, but waiting for clarification on token naming.

[Panel](https://codepen.io/DintaMel-the-lessful/pen/bNwgQXM?editors=1000)
[Dialog](https://codepen.io/DintaMel-the-lessful/pen/WbGRLGY?editors=1000)
[Flow Item](https://codepen.io/DintaMel-the-lessful/pen/RNGKEKL?editors=1000)

[...]

Issue #10093, opened 2024-08-16
[Dialog] Add a css property for controlling background color


macandcheese:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

As part of the migration story from Modal -> Dialog, users are currently unable to replicate the `--calcite-modal-background-color` functionality in Dialog. 

### Acceptance Criteria

While we plan to add support for this as part of the larger effort in #7180 - currently users cannot migrate from Modal to Dialog and persist their existing design. I'd expect this to be `--calcite-dialog-background-color`

### Relevant Info

Dialog will of course get the "full token treatment" when that lands, but this could be a way to ease the migration issue in the meantime. We should use the same name we expect to use as part of #7180 to ensure no breaking changes occur when that lands.

### Which Component

Dialog

### Example Use Case

As a user, I want to set the background color of the Dialog via css property.

### Priority impact

impact - p1 - need for current milestone

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

doreenbrinkman:
Currently converting `CalciteModal` to `CalciteDialog` components to make use of the new `alerts` slot for Data Pipelines and we need to maintain the same visual design as the `CalciteModal` for now. The content background color is grey and cannot be changed.  Adding `--calcite-dialog-background-color` will help with the migration.  Thanks.

abigailmbravo-uxuidev:
+1 needed for ArcGIS Online

allyshah:
+1 for ArcGIS Dashboards

ashishgodbole:
+1 for Dialog. We seem to still be using Modals with white bkg in Web Editor haha. 
Btw, do we have this for other containers like panels, sheets, pages, etc. Imo this is partly calcite enhancement, and partly us designers better aligning on visual consistency in pattern usage.

DitwanP:
🍠 Verified locally on `dev` 
![image](https://github.com/user-attachments/assets/9e8a7dbd-eeb0-4d30-bc71-ead4cec8e54b)


noticed Panel was also included in the PR that added the dialog token so verified that as well:
![image](https://github.com/user-attachments/assets/4091e859-1037-4732-a0e4-e329e0942100)

PR #10387, opened 2024-09-25
feat(dialog, panel): Add css properties for background-color

macandcheese:
**Related Issue:** #10093 

## Summary
Adds `--calcite-dialog-background-color` to Dialog
Adds `--calcite-panel-background-color` to Panel

Further token work may occur through related issues: #7180, #10380, etc.


jcfranco (approved):
Noice! 🚀

Before merging, can you:

1. list out the new CSS props added to each component in the description? 
2. add some context on additional component tokens and test coverage will be added as part of #7180? (maybe this note is enough 😅)

PR #10390, opened 2024-09-25
fix(text-area): ensure border-color token doesn't override invalid styles

benelan:
**Related Issue:** #7180

## Summary

Ensure `--calcite-text-area--border-color` does not override the `--calcite-color-status-danger` styling.


geospatialem (commented):
Might be out of scope for the request since it would apply to more than the PR's scope, but wondering if we should consider updating the references of "`text area`" to "`the component`" instead to be in alignment with other docs 📚?

For instance:
> --calcite-text-area-border-color: Specifies the border color of the ~text area~ **component**.

jcfranco (approved):
🚀

PR #10403, opened 2024-09-26
fix(action): only set aria-busy and aria-pressed when necessary

driskull:
**Related Issue:** #

## Summary

- Only set `aria-busy` and `aria-pressed` when necessary
- add tests

### DIFF
--- a/packages/calcite-components/src/components/action/action.tsx
+++ b/packages/calcite-components/src/components/action/action.tsx
@@ -313,10 +313,10 @@ export class Action
       <Host>
         <InteractiveContainer disabled={disabled}>
           <button
-            aria-busy={toAriaBoolean(loading)}
+            aria-busy={loading ? toAriaBoolean(loading) : null}
             aria-controls={indicator ? indicatorId : null}
             aria-label={ariaLabel}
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/action/action.tsx
```diff
@@ -313,10 +313,10 @@ export class Action
       <Host>
         <InteractiveContainer disabled={disabled}>
           <button
-            aria-busy={toAriaBoolean(loading)}
+            aria-busy={loading ? toAriaBoolean(loading) : null}
             aria-controls={indicator ? indicatorId : null}
             aria-label={ariaLabel}
-            aria-pressed={toAriaBoolean(active)}
+            aria-pressed={active ? toAriaBoolean(active) : null}
```
driskull, line 319: @geospatialem I wonder if we need a new props like `pressable` and `loadable` in order to properly do this?

In theory, any action can be active or loading at any time so maybe this PR isn't valid? I'd love your thoughts here.

driskull:
closing because I don't think this is the correct approach. We would need to add additional properties to not make something active or loading.

Issue #10416, opened 2024-09-27
[Tooltip] Will reappear and block actions within a menu


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The `tooltip` will reappear and block `actions` when interacting with an `action-menu`. This is a regression surfacing in `2.11.0`, the linked PR is referenced below.

![regression-tooltip](https://github.com/user-attachments/assets/f2c42bc0-d90b-4fe7-80ab-ba83a422c883)

### Expected Behavior

![regression-tooltip-2](https://github.com/user-attachments/assets/a259b2cc-29dc-4ee7-80e4-62c75baae52a)

### Reproduction Sample

https://codepen.io/geospatialem/pen/ZEgGxge

### Reproduction Steps

1. Open the [sample](https://codepen.io/geospatialem/pen/ZEgGxge)
2. Select the folder / Save and Open action to open the menu
3. Observe the tooltip will close, flicker, and reappear

### Reproduction Version

2.12.2

### Relevant Info

Looks to be related to https://github.com/Esri/calcite-design-system/pull/9487

### Regression?

`2.11.0-next.15` / `2.10.1`

### Priority impact

impact - p1 - need for current milestone

### Impact

High

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Data Pipelines

DitwanP:
🍡 Verified locally on `dev` 

https://github.com/user-attachments/assets/2da34a06-314c-4025-aff5-d0ec66a87983

doreenbrinkman:
This tooltip issue still occurs using Safari Version 18.0.1 (19619.1.26.111.11, 19619)
https://codepen.io/doreenbrinkman/pen/vYoQexP

PR #10420, opened 2024-09-27
fix(tooltip): closed tooltips should not reappear

driskull:
**Related Issue:** #10416

## Summary

- track clicked tooltip to stop hovering until the tooltip is hovered off of
- add e2e test


jcfranco (approved):
✨👇💭✨

Issue #10398, opened 2024-09-26
[ESLint] Remove `ban-props-on-host`, `enforce-ref-last-prop`, `require-event-emitter-type` rules


maxpatiiuk:
### Priority impact

p - low

### Summary

Calcite developed an `enforce-ref-last-prop` ESLint rule.
Such rule is not necessary in Lumina - the JSX `ref` prop is moved to be the last prop at build time, thus the sort order in the component code does not matter.

**Update**: Also remove `ban-props-on-host` and `require-event-emitter-type` rules as they will no longer be needed.

### Desired Outcome

After migrating to Lumina, remove the `ban-props-on-host`, `enforce-ref-last-prop` and `require-event-emitter-type` ESLint rules

jcfranco:
We can deprecate and remove this in 3.x since https://github.com/Esri/calcite-design-system/pull/9086/files disabled the rule for components.

jcfranco:
@geospatialem @DitwanP How do you suggest we proceed with this issue? We're deprecating now, but removing in 3.x.

geospatialem:
Verified removals on the `dev` branch. 🧹 

Reassigning the issue back to development to remove the remaining markdown files 🔜 :
- [ban-props-on-host.md](https://github.com/Esri/calcite-design-system/blob/dev/packages/eslint-plugin-calcite-components/docs/ban-props-on-host.md)
- [require-event-emitter-type](https://github.com/Esri/calcite-design-system/blob/dev/packages/eslint-plugin-calcite-components/docs/require-event-emitter-type.md)

Issue #10310, opened 2024-09-13
Migrate component framework to Lit


jcfranco:
### Priority impact

p - high

### Summary

Migrating to Lit will enable us to leverage key features that will streamline development and improve our overall workflow.

Blocked issues: #8249, #0220, #8307, #9057, #9090, #10038, #10237, #10263, #9089, #7631, #9494, #10399, #10423, #10422, #10376, #991, #10619, #10227, #9260

### Desired Outcome

Components are fully Lit-based.

#### Related

#10356
#10352
#10344 
https://github.com/Esri/calcite-design-system/issues/8073#issuecomment-2174193805 (might address potential bug)

DitwanP:
Verified 🔥

PR #10421, opened 2024-09-27
deprecate: deprecate `enforce-ref-last-prop` rule

jcfranco:
**Related Issue:** #10398

## Summary

Deprecates this rule for 3.x removal as it is no longer needed after Stencil 4.14.1 and will be even more obsolete after #10310.


aPreciado88 (approved):
🚀

Issue #10369, opened 2024-09-21
[Tests] Incorrect 2nd argument to window.getComputedStyle


maxpatiiuk:
### Priority impact

p - low

### Test type

E2E tests

### Which Component(s)

Button, Chip, ...

### Unstable Tests

In several places in the code, Calcite's E2E tests call `.getComputedStyle` with `:hover` or `:focus` as the 2nd argument:

https://github.com/Esri/calcite-design-system/blob/2305e0159c1b4f0eaaca8ac356b3b9d3e8638737/packages/calcite-components/src/components/button/button.e2e.ts#L538-L585

As far as I am aware this is not valid. Per MDN, **only `::after` and `::before` are allowed**.
ChatGPT o1 says the same:
> No, ":focus" is not a valid value for the second argument of window.getComputedStyle(). The second argument of window.getComputedStyle() is intended for pseudo-elements, not pseudo-classes. Pseudo-elements are parts of elements (like ::before, ::after), while pseudo-classes represent a state of an element (like :focus, :hover).

### Test error, if applicable

While `:hover` and `:focus` appears to work fine in Chrome 119 (which is what your Stencil E2E tests are using), it does not appear to work in Chrome 128 (the latest version), causing failing tests.

You should refactor the usages in tests to not pass `:hover` or `:focus`

DitwanP:
🍠✨ Verified

PR #10424, opened 2024-09-27
test: fix `window.getComputedStyle` arguments

jcfranco:
**Related Issue:** #10369 

## Summary

This updates `getComputedStyle` args to:

* use `::after` vs `:after`
* remove invalid pseudoclass args (only pseudo-elements are supported)

### DIFF
--- a/packages/calcite-components/src/components/button/button.e2e.ts
+++ b/packages/calcite-components/src/components/button/button.e2e.ts
@@ -533,12 +533,12 @@ describe("calcite-button", () => {
         buttonEl = await page.find("calcite-button >>> button");
         await buttonEl.focus();
         await page.waitForChanges();
-        buttonFocusStyle = await buttonEl.getComputedStyle(":focus");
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/button/button.e2e.ts
```diff
@@ -533,12 +533,12 @@ describe("calcite-button", () => {
         buttonEl = await page.find("calcite-button >>> button");
         await buttonEl.focus();
         await page.waitForChanges();
-        buttonFocusStyle = await buttonEl.getComputedStyle(":focus");
+        buttonFocusStyle = await buttonEl.getComputedStyle();
```
anveshmekala, line 536: Nitpick: `buttonFocusStyle` can be changed to `buttonStyle` here to be more generic. Applies to similar use cases.

anveshmekala (approved):
🧹 🧹 🧹 🧹 🧹 🧹 🧹

Issue #10349, opened 2024-09-19
refactor(tests): migrate away from newSpecPage


maxpatiiuk:
### Priority impact

p - low

### Test type

Unit tests

### Which Component(s)

Heading

### Unstable Tests

Calcite has 1410 newE2EPage tests and only 2 newSpecPage tests in one file:

https://github.com/Esri/calcite-design-system/blob/7fe16011957a45f3340ba9bfe137bcae258bf314/packages/calcite-components/src/components/functional/Heading.spec.tsx#L2C10-L2C19

To simplify migration, Lumina temporary provides newE2EPage() function that matches Stencil's API. In the long term, it would be great to migrate to Vitest browser test to improve DX, performance and integrations with Storybook.

Given that newSpecPage is only used by Calcite in one file, Lumina doesn't provide Stencil's equivalent of newSpecPage.

TODO:
- Refactor the usage of newSpecPage in Heading.spec.tsx. Options: https://github.com/Esri/calcite-design-system/issues/10349#issuecomment-2377606428

### PR skipped, if applicable

#

maxpatiiuk:
Lumina has an alternative to `newSpecPage`, and support for JSX in tests, but this is only supported in Vitest browser mode tests.

Since you are in Puppeteer E2E tests for now, some options come to mind:
- (simplest) In your test, use the `render()` function from lit-html directly to render the functional component to some html markup. This render() function will be called in Node.js, not browser.
- (verbose) Create a wrapper web component class for functional component and test that web component like you do any other component class
- (tricky) import the `render()` function from lit-html in the browser (using `await page.addScriptTag` or `await page.evaluate()`) and render the function component there.

DitwanP:
🍡 Verified

PR #10431, opened 2024-09-27
test(heading): avoid `newSpecPage` usage to ease Lumina migration

jcfranco:
**Related Issue:** #10349

## Summary

Adds stopgap `VNode`/rendering assertion util to `Heading.spec.tsx`. Once #10310 lands, we can revisit this and use better test tooling.

### DIFF
--- a/packages/calcite-components/src/components/functional/Heading.spec.tsx
+++ b/packages/calcite-components/src/components/functional/Heading.spec.tsx
@@ -1,5 +1,4 @@
-import { h } from "@stencil/core";
-import { newSpecPage } from "@stencil/core/testing";
+import { h, VNode } from "@stencil/core";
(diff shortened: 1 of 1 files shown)


driskull (approved):
👍

packages/calcite-components/src/components/functional/Heading.spec.tsx
```diff
@@ -13,26 +12,55 @@ describe("constrainHeadingLevel", () => {
...
+function assertVNode(
+  vnode: VNode,
+  expected: {
+    tag: string;
+    attrs: Record<string, any>;
+    children: (VNode | string)[];
+  },
+) {
+  expect(vnode).toEqual(
+    expect.objectContaining({
```
maxpatiiuk, line 33: TIL: I didn't know this was allowed!

supported in vitest too

Issue #10348, opened 2024-09-19
Refactor-out usages of Stencil's "readTask"


maxpatiiuk:
### Description

In 3 files, Calcite uses Stencil's `readTask()` function.

Could you check if there was a particular reason readTask was used or was it just a performance micro optimization?

Lit's default [re-render scheduling is very simple](https://github.com/lit/lit/blob/42260c2d3bb179756e109616338936de2a589801/packages/reactive-element/src/reactive-element.ts#L1297-L1318) - they don't provide readTask/writeTask. However, they expose hook for customizing the scheduling, letting you implement your own alternatives of Stencil's readTask/writeTask if it is important to have them.
 
Stencil's readTask/writeTask is a smart performance optimization, but it's hardly used - no usages at Esri outside of Calcite, and Calcite has only 3 usages.

TODO:
- check if readTask() is needed
- if not, refactor out it's usages

### Proposed Advantages

Compatibility with Lit's default scheduling system

### Which Component

3 files:
- https://github.com/Esri/calcite-design-system/blob/7fe16011957a45f3340ba9bfe137bcae258bf314/packages/calcite-components/src/components/stepper/stepper.tsx#L14
- https://github.com/Esri/calcite-design-system/blob/7fe16011957a45f3340ba9bfe137bcae258bf314/packages/calcite-components/src/components/tab-nav/tab-nav.tsx#L10
- https://github.com/Esri/calcite-design-system/blob/7fe16011957a45f3340ba9bfe137bcae258bf314/packages/calcite-components/src/utils/openCloseComponent.ts#L1

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

DitwanP:
🍠✨

PR #10432, opened 2024-09-28
refactor: use `requestAnimationFrame` to replace `readTask`

jcfranco:
**Related Issue:** #10348

## Summary

We've been using `readTask` in a few places to ensure proper timing for DOM reads w/o triggering layout thrashing. Looking at [the implementation](https://github.com/ionic-team/stencil/blob/7ede77a873486b5cf47f0b26571f852675f67dd6/src/client/client-task-queue.ts#L23), we might be able to get close or similar behavior by using `requestAnimationFrame`.

### DIFF
--- a/packages/calcite-components/src/utils/openCloseComponent.spec.ts
+++ b/packages/calcite-components/src/utils/openCloseComponent.spec.ts
@@ -9,9 +9,9 @@ describe("openCloseComponent", () => {
     let dispatchTransitionEvent: TransitionEventDispatcher;
 
     beforeEach(() => {
-      jest.spyOn(openCloseComponent, "internalRaf").mockImplementation((task) => {
-        task(1337);
-        return 1337;
+      jest.spyOn(global, "requestAnimationFrame").mockImplementation((cb) => {
+        cb(0);
+        return 0;
       });
       dispatchTransitionEvent = createTransitionEventDispatcher();
     });
--- a/packages/calcite-components/src/utils/openCloseComponent.spec.ts
+++ b/packages/calcite-components/src/utils/openCloseComponent.spec.ts
@@ -9,7 +9,10 @@ describe("openCloseComponent", () => {
     let dispatchTransitionEvent: TransitionEventDispatcher;
 
     beforeEach(() => {
-      jest.spyOn(openCloseComponent, "internalReadTask").mockImplementation((task) => task(1337));
+      jest.spyOn(openCloseComponent, "internalRaf").mockImplementation((task) => {
+        task(1337);
+        return 1337;
+      });
       dispatchTransitionEvent = createTransitionEventDispatcher();
     });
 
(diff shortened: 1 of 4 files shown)


packages/calcite-components/src/utils/openCloseComponent.ts
```diff
@@ -1,10 +1,9 @@
-import { readTask } from "@stencil/core";
 import { whenTransitionDone } from "./dom";
 
 /**
  * Exported for testing purposes only
  */
-export const internalReadTask = readTask;
+export const internalRaf = requestAnimationFrame;
```
jcfranco, line 6: I'll try to remove this before installing.

packages/calcite-components/src/utils/openCloseComponent.spec.ts
```diff
@@ -9,7 +9,10 @@ describe("openCloseComponent", () => {
     let dispatchTransitionEvent: TransitionEventDispatcher;
 
     beforeEach(() => {
-      jest.spyOn(openCloseComponent, "internalReadTask").mockImplementation((task) => task(1337));
+      jest.spyOn(openCloseComponent, "internalRaf").mockImplementation((task) => {
+        task(1337);
+        return 1337;
```
driskull, line 14: 1337!

packages/calcite-components/src/utils/openCloseComponent.ts
driskull: Yeah seems like this just needs to be cleaned up to just use requestAnimationFrame by itself.

driskull (approved):
👍

jcfranco:
@nel11211 Let us know if you are able to narrow it down to this.

nel11211:
> @nel11211 Let us know if you are able to narrow it down to this.

Ok I think I see what's happening. `arcgis-tab-manager` triggers a dynamic `click()` [here](https://devtopia.esri.com/WebGIS/arcgis-web-analysis/blob/master/packages/analysis-studio-components/src/components/arcgis-tabs-manager/arcgis-tabs-manager.tsx#L367) when a tab `select()` function is called. When the dropdown is opened, Charts also opens a new tab for the chart in question and selects it. The dynamic click is being interpreted by the `calcite-dropdown` as a click outside the dropdown, triggering an auto-close.

@kumarGayu Is a dynamic click necessary for the function that selects a tab?

nel11211:
Opened an issue in Analysis repo if we want to move discussion there: https://devtopia.esri.com/WebGIS/arcgis-web-analysis/issues/7297

nel11211:
> @nel11211 Thanks for sharing your findings! If I’m reading this right, this doesn’t seem to be related to this PR. Would you mind moving the conversation to a more appropriate place?

It's not the root cause no, but worth noting for some reason the issue wasn't happening until the change from this PR.

Issue #8359, opened 2023-12-06
[Tooling] Obtain and track estimates for milestones 


geospatialem:
### Summary

1. Create tooling that provides us with historical time estimates on closed milestones and 
2. Initiate automation that provides us with time estimates via Monday once a milestone is closed

Refer to the "Estimate tracking strategery" chat for the Monday API token, if needed.

cc: @brittneytewks 

### Desired Outcome

- Get a historical perspective on time estimates with milestones since May 2023
- Continue to get additional insights and stats on estimates with upcoming milestones as they close

### Resources

As far as past milestone tracking, it looks like we started getting the majority of issues estimates in May, so if possible if we could get total estimates for the closed milestones _listed below_:

- List of milestone numbers, if its more useful: 110,129,132,121,122,134,135,136,123,138,124,144,145,126,127,128

Milestone links:
- [1.4.0](https://github.com/Esri/calcite-design-system/milestone/110) - May
- [1.4.1](https://github.com/Esri/calcite-design-system/milestone/129)
- [1.4.2](https://github.com/Esri/calcite-design-system/milestone/132) - should be 2 (if we want to verify all is working as-expected)
- [1.4.3](https://github.com/Esri/calcite-design-system/milestone/121) - June
- [1.5.0](https://github.com/Esri/calcite-design-system/milestone/122) - July
- [1.5.1](https://github.com/Esri/calcite-design-system/milestone/134)
- [1.6.0](https://github.com/Esri/calcite-design-system/milestone/135) - August
- [1.6.1](https://github.com/Esri/calcite-design-system/milestone/136)
- [1.7.0](https://github.com/Esri/calcite-design-system/milestone/123)
- [1.8.0](https://github.com/Esri/calcite-design-system/milestone/138)
- [1.9.0](https://github.com/Esri/calcite-design-system/milestone/124) - October
- [1.9.1](https://github.com/Esri/calcite-design-system/milestone/144)
- [1.9.2](https://github.com/Esri/calcite-design-system/milestone/145)
- [1.10.0](https://github.com/Esri/calcite-design-system/milestone/126)
- [2.0.0](https://github.com/Esri/calcite-design-system/milestone/127) - November
- [TBD December](https://github.com/Esri/calcite-design-system/milestone/128) - December

benelan:
For verification, please follow the usage instructions on the action's wiki page:
https://github.com/Esri/calcite-design-system/wiki/milestone-estimate-tracking

[...]

PR #10442, opened 2024-09-30
ci: compile estimate totals per milestone

benelan:
**Related Issue:** #8359

## Summary

Add an action that creates a spreadsheet and JSON file with the total estimates per milestone. The action runs when an issue closes and can be manually dispatched. The files are saved as artifacts in the workflow run. I'll add a wiki page about running the workflow and retrieving the data once it's live.

It might be because I'm a GraphQL noob, but Monday's API reference was not very helpful. I can give the integration another try if this is not ergonomic enough.


geospatialem (commented):
Awesome! 🏆 

Code-wise will defer approval to @jcfranco, but in initial sleuthing this looks great! Wondering if there are limitations that should be mentioned here as well, it looks like we can only run on up to 100 milestones and 100 issues for each milestone. Wondering if that could pose a problem, in particular around the 100 milestones since they could be open or closed. WDYT?

benelan:
I'm paginating the issues so there is no limit. I could do the same for milestones, the downside being runs will take longer. You're the main consumer of this data so let me know how many milestones should be included and if you'd like any additional columns.

geospatialem:
> I'm paginating the issues so there is no limit. I could do the same for milestones, the downside being runs will take longer. You're the main consumer of this data so let me know how many milestones should be included and if you'd like any additional columns.

No worries, it seems like a good approach and we can re-evaluate in the future if needed. 💪🏻

jcfranco (approved):
📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉
📉📈📉📉📉📈📉📉📈📈📉📉📈📈📈📉📉📈📈📈📉📈📈📈📈📉📈📉
📉📈📈📉📉📈📉📈📉📉📈📉📉📈📉📉📈📉📉📉📉📈📉📉📉📉📈📉
📉📈📉📈📉📈📉📈📉📉📈📉📉📈📉📉📈📉📉📉📉📈📈📈📉📉📈📉
📉📈📉📉📈📈📉📈📉📉📈📉📉📈📉📉📈📉📉📉📉📈📉📉📉📉📉📉
📉📈📉📉📉📈📉📉📈📈📉📉📈📈📈📉📉📈📈📈📉📈📈📈📈📉📈📉
📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉📉

PR #10443, opened 2024-09-30
refactor: prepare tests for migration

maxpatiiuk:
**Related Issue:** #10310 

## Summary

This is a series of PRs aimed at making subtle changes to tests and components with aim of fixing tests that are failing after the codemod. By fixing as many tests as possible before the codemod, there is less work to do after the codemod is run, and thus the team would be blocked for less time.

### DIFF
--- a/packages/calcite-components/src/components/action-menu/action-menu.tsx
+++ b/packages/calcite-components/src/components/action-menu/action-menu.tsx
@@ -96,7 +96,7 @@ export class ActionMenu implements LoadableComponent {
 
   @Watch("open")
   openHandler(open: boolean): void {
-    this.activeMenuItemIndex = this.open ? Math.max(0, this.activeMenuItemIndex) : -1;
+    this.activeMenuItemIndex = this.open ? 0 : -1;
     if (this.menuButtonEl) {
       this.menuButtonEl.active = open;
     }
--- a/packages/calcite-components/src/components/action-menu/action-menu.tsx
+++ b/packages/calcite-components/src/components/action-menu/action-menu.tsx
@@ -96,7 +96,7 @@ export class ActionMenu implements LoadableComponent {
 
   @Watch("open")
   openHandler(open: boolean): void {
-    this.activeMenuItemIndex = this.open ? 0 : -1;
+    this.activeMenuItemIndex = this.open ? Math.max(0, this.activeMenuItemIndex) : -1;
     if (this.menuButtonEl) {
       this.menuButtonEl.active = open;
     }
--- a/packages/calcite-components/src/components/card-group/card-group.tsx
+++ b/packages/calcite-components/src/components/card-group/card-group.tsx
@@ -215,7 +215,7 @@ export class CardGroup implements InteractiveComponent, LoadableComponent {
       });
     }
 
-    this.selectedItems = this.items.filter((el) => el.selected || el.hasAttribute("selected"));
+    this.selectedItems = this.items.filter((el) => el.selected);
 
     if (emit && this.selectionMode !== "none" && !this.disabled) {
       this.calciteCardGroupSelect.emit();
--- a/packages/calcite-components/src/components/card-group/card-group.tsx
+++ b/packages/calcite-components/src/components/card-group/card-group.tsx
@@ -63,7 +63,7 @@ export class CardGroup implements InteractiveComponent, LoadableComponent {
 
   @Watch("selectionMode")
   onSelectionModeChange(): void {
-    this.udpateItemsOnSelectionModeChange();
+    this.updateItemsOnSelectionModeChange();
   }
 
   /**
@@ -169,7 +169,7 @@ export class CardGroup implements InteractiveComponent, LoadableComponent {
   //
   //--------------------------------------------------------------------------
 
-  private udpateItemsOnSelectionModeChange = (): void => {
+  private updateItemsOnSelectionModeChange = (): void => {
     this.updateSlottedItems(this.slotRefEl);
     this.updateSelectedItems();
   };
@@ -215,7 +215,7 @@ export class CardGroup implements InteractiveComponent, LoadableComponent {
       });
     }
 
-    this.selectedItems = this.items.filter((el) => el.selected);
+    this.selectedItems = this.items.filter((el) => el.selected || el.hasAttribute("selected"));
 
     if (emit && this.selectionMode !== "none" && !this.disabled) {
       this.calciteCardGroupSelect.emit();
--- a/packages/calcite-components/src/components/card/card.tsx
+++ b/packages/calcite-components/src/components/card/card.tsx
@@ -34,6 +34,7 @@ import {
 } from "../../utils/interactive";
 import { isActivationKey } from "../../utils/key";
 import { IconNameOrString } from "../icon/interfaces";
+import type { CalciteCheckboxCustomEvent } from "../../components";
 import { CSS, ICONS, SLOTS } from "./resources";
 import { CardMessages } from "./assets/card/t9n";
 
@@ -276,25 +277,14 @@ export class Card
         <calcite-checkbox
           checked={this.selected}
           label={this.messages.select}
-          onClick={this.selectCardDeprecated}
-          onKeyDown={this.cardSelectKeyDownDeprecated}
+          onCalciteCheckboxChange={this.selectCardDeprecated}
         />
       </calcite-label>
     );
   }
 
-  private cardSelectKeyDownDeprecated = (event: KeyboardEvent): void => {
-    switch (event.key) {
-      case " ":
-      case "Enter":
-        this.selectCardDeprecated();
-        event.preventDefault();
-        break;
-    }
-  };
-
-  private selectCardDeprecated = (): void => {
-    this.selected = !this.selected;
+  private selectCardDeprecated = (event: CalciteCheckboxCustomEvent<void>): void => {
+    this.selected = event.target.checked;
     this.calciteCardSelect.emit();
   };
 
--- a/packages/calcite-components/src/components/checkbox/checkbox.tsx
+++ b/packages/calcite-components/src/components/checkbox/checkbox.tsx
@@ -189,14 +189,6 @@ export class Checkbox
     }
   };
 
-  clickHandler = (): void => {
-    if (this.disabled) {
-      return;
-    }
-
-    this.toggle();
-  };
-
   //--------------------------------------------------------------------------
   //
   //  Events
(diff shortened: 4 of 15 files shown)


packages/calcite-components/src/components/action-menu/action-menu.tsx
```diff
@@ -96,7 +96,7 @@ export class ActionMenu implements LoadableComponent {
 
   @Watch("open")
   openHandler(open: boolean): void {
-    this.activeMenuItemIndex = this.open ? 0 : -1;
+    this.activeMenuItemIndex = this.open ? Math.max(0, this.activeMenuItemIndex) : -1;
```
maxpatiiuk, line 99: This change is needed because the watcher is turned into a `willUpdate()` after the codemod, and willUpdate runs in batch for all changed properties.

The change is required because before the open watcher would be triggered here, setting `activeMenuItemIndex` to 0:

https://github.com/Esri/calcite-design-system/blob/de835893da0023cfdbeb027c8720be308fda839e/packages/calcite-components/src/components/action-menu/action-menu.tsx#L456

but then the value of activeMenuItemIndex is manually changed here:

https://github.com/Esri/calcite-design-system/blob/de835893da0023cfdbeb027c8720be308fda839e/packages/calcite-components/src/components/action-menu/action-menu.tsx#L463

Because of the `willUpdate()` order change, the activeMenuItemIndex is now set to length-1 and then reset to 0 by the openHandler().

Adding `Math.max(0, this.activeMenuItemIndex)` workarounds this difference.
The other solution is to turn `open` watcher into a getter/setter, but that requires manual changes post-codemod.

packages/calcite-components/src/components/card-group/card-group.tsx
```diff
@@ -63,7 +63,7 @@ export class CardGroup implements InteractiveComponent, LoadableComponent {
 
   @Watch("selectionMode")
   onSelectionModeChange(): void {
-    this.udpateItemsOnSelectionModeChange();
+    this.updateItemsOnSelectionModeChange();
```
maxpatiiuk, line 66: typo

packages/calcite-components/src/components/card-group/card-group.tsx
```diff
@@ -215,7 +215,7 @@ export class CardGroup implements InteractiveComponent, LoadableComponent {
       });
     }
 
-    this.selectedItems = this.items.filter((el) => el.selected);
+    this.selectedItems = this.items.filter((el) => el.selected || el.hasAttribute("selected"));
```
maxpatiiuk, line 218: Before, card-group and card where in the same lazy loading bundle so by the time card-group executes this function, the `card` constructor is definitely loaded and attributes are reflected to properties.
With Lumina, we do not put related components into bundle yet, so `card` is not guaranteed to be loaded by now. Until `card` is loaded, attribute is not reflected to a property so we must check both.

The other option is to call this `setSelectedItems()` again in `loaded()` as `card` is guaranteed to be loaded by then.

packages/calcite-components/src/components/card/card.tsx
```diff
@@ -276,25 +277,14 @@ export class Card
         <calcite-checkbox
           checked={this.selected}
           label={this.messages.select}
-          onClick={this.selectCardDeprecated}
```
maxpatiiuk, line 279: This code was a bit problematic.
On click, calcite-card did `this.selected = !this.selected;`, which would flip the `checked` property in calcite-checkbox.
But, calcite-checkbox is also listening for `click` and would also flip the `checked` property in response.

This worked fine in Stencil because the order was as follows:
- calcite-card onClick fires (flipping property from true to false)
- calcite-checkbox onClick fires (flipping property from true to false)
- calcite-card re-renders, passing `checked: false` to calcite-checkbox, which is also the value `calcite-checkbox` set itself too

In Lit, the order is a bit different:
- calcite-card onClick fires (flipping property from true to false)
- calcite-card re-renders, passing `checked: false` to calcite-checkbox
- calcite-checkbox onClick fires, flipping property from false to true. Thus in effect we flipped the property twice, preventing the change.

Fixed by listening to `onCalciteCheckboxChange` instead and letting calcite-checkbox dictate the new post-flip state.

packages/calcite-components/src/components/checkbox/checkbox.tsx
```diff
@@ -189,14 +189,6 @@ export class Checkbox
     }
   };
 
-  clickHandler = (): void => {
```
maxpatiiuk, line 192: this function is useless since calling `toggle` directly produces the same effect

packages/calcite-components/src/components/color-picker-hex-input/color-picker-hex-input.e2e.ts
```diff
@@ -54,10 +54,10 @@ describe("calcite-color-picker-hex-input", () => {
     await page.setContent("<calcite-color-picker-hex-input allow-empty></calcite-color-picker-hex-input>");
 
     const input = await page.find(`calcite-color-picker-hex-input`);
-    input.setProperty("value", null);
+    input.setProperty("value", undefined);
     await page.waitForChanges();
 
-    expect(await input.getProperty("value")).toBe(null);
+    expect(await input.getProperty("value")).toBe(undefined);
```
maxpatiiuk, line 60: As per our decision, if a property is set to `null` in Lumina, it will be automatically cast to `undefined`.

Thus, updating this test to use `undefined` instead of `null` so that it works pass post-codemod without further changes

packages/calcite-components/src/components/color-picker/color-picker.tsx
```diff
@@ -998,19 +998,17 @@ export class ColorPicker
               </div>
               {savedColors.length > 0 ? (
                 <div class={CSS.savedColors}>
-                  {[
-                    ...savedColors.map((color) => (
```
maxpatiiuk, line 1002: was the a reason for `[... ]` here or was it an accident?
git blame shows this was added 4 years ago in the very first version of `calcite-color` component

packages/calcite-components/src/components/color-picker/color-picker.tsx
```diff
@@ -1376,13 +1374,19 @@ export class ColorPicker
     );
   };
 
-  private initColorField = (canvas: HTMLCanvasElement): void => {
+  private initColorField = (canvas?: HTMLCanvasElement): void => {
+    if (!canvas) {
```
maxpatiiuk, line 1378: Two important differences to note about `ref` props:
- If component is disconnected, in Stencil `ref` callback is not called with null/undefined. `ref` callback is only called with `null` if the component re-renders without the VNode for the ref element. In Lit/React/Preact, if component is disconnected, all ref callbacks inside of it will be called with `undefined`.
- In **Lit**, and in **React**/**Preact** if to a `ref` prop you pass a new arrow function on each render, both frameworks will call the arrow function instance from previous render with `undefined` (Lit) or `null` (React/Preact), and then call the new arrow function with the element reference.
  
   In **Stencil**, changes to the ref callback do not make it be called again. The `ref` callback is only called again when the element is actually disconnected.

maxpatiiuk: In practice that means that ref callbacks should always guard against `ref` callback being called with a nullish value.
Calcite does not have TypeScript strictNullChecks option enabled. If it did, then TypeScript would detect cases where you forgot to handle the nullish ref case, preventing a bug from happening.

packages/calcite-components/src/components/functional/Heading.tsx
```diff
@@ -14,8 +14,6 @@ export function constrainHeadingLevel(level: number): HeadingLevel {
 export const Heading: FunctionalComponent<HeadingProps> = (props, children): VNode => {
   const HeadingTag = props.level ? `h${props.level}` : "div";
 
-  delete props.level;
```
maxpatiiuk, line 17: useless

packages/calcite-components/src/demos/input.html
```diff
@@ -1598,6 +1598,6 @@
     </demo-dom-swapper>
   </body>
   <script>
-    document.addEventListener("calciteInputInput", ({ detail }) => console.log(`input: ${detail.value}`));
+    document.addEventListener("calciteInputInput", ({ target }) => console.log(`input: ${target.value}`));
```
maxpatiiuk, line 1601: `calciteInputInput` does not emit payload

packages/calcite-components/src/utils/locale.ts
```diff
@@ -212,10 +212,12 @@ export interface LocalizedComponent {
    *
    * Components should watch this prop to ensure messages are updated.
    *
+   * ```tsx
```
maxpatiiuk, line 215: Having `@Watch` in JSDoc without it being escaped in backticks was being parsed by TypeScript as a JSDoc tag, making TypeScript insert extra whitespace around it.
To have things parse correctly, it is advised to wrap all code blocks in JSDoc comments in backticks.

packages/calcite-components/src/components/card-group/card-group.tsx
maxpatiiuk: edit:
undoing this change for now because reflected attributes are not updated until re-render, so this change is causing `this.selectedItems` to be computed based on old value of `.selected` prop

packages/calcite-components/src/components/color-picker-hex-input/color-picker-hex-input.tsx
```diff
@@ -83,7 +83,7 @@ export class ColorPickerHexInput implements LoadableComponent {
   //--------------------------------------------------------------------------
 
   /**
-   * When `true`, an empty color (`null`) will be allowed as a `value`.
+   * When `true`, an empty color (`undefined`) will be allowed as a `value`.
```
maxpatiiuk, line 86: This entire component is internal, so I am assuming the change from `null` to `undefined` is not a breaking change?

packages/calcite-components/src/components/card/card.tsx
```diff
@@ -276,25 +277,14 @@ export class Card
...
-      case "Enter":
-        this.selectCardDeprecated();
-        event.preventDefault();
-        break;
-    }
-  };
-
-  private selectCardDeprecated = (): void => {
-    this.selected = !this.selected;
+  private selectCardDeprecated = (event: CalciteCheckboxCustomEvent<void>): void => {
```
jcfranco, line 286: Can you rename this to follow event handler naming conventions?

jcfranco (approved):
Aside from a few comments, this LGTM! 🚀

packages/calcite-components/src/utils/locale.ts
maxpatiiuk: Ah, I didn't realize that.
Looks like there is no legal way to write decorator code examples in JSDoc for now. There is an issue for TypeScript 5.7 though: https://github.com/microsoft/TypeScript/issues/47679
I will revert this change then, and add a post-processing step to the codemod to remove the extra whitespace added by incorrect parsing.

packages/calcite-components/src/components/action-menu/action-menu.tsx
maxpatiiuk: > This option seems cleaner to me

Makes sense. I will undo
Based on what we discussed, I will add logic to the codemod to convert this specific occurrence of watch into a get/set

jcfranco (approved):
🚀👨‍🚀🪐

PR #10447, opened 2024-09-30
fix(accordion-item): fix rendering tied to named-slot content

driskull:
**Related Issue:** #6059

## Summary


jcfranco (approved):
🎉

PR #10449, opened 2024-09-30
fix(block): fix rendering tied to named-slot content

driskull:
**Related Issue:** #6059

## Summary

### DIFF
--- a/packages/calcite-components/src/components/block/block.e2e.ts
+++ b/packages/calcite-components/src/components/block/block.e2e.ts
@@ -190,11 +190,9 @@ describe("calcite-block", () => {
       html`<calcite-block heading="heading" description="description"><div>Hello world!</div></calcite-block>`,
     );
     await skipAnimations(page);
-    await page.waitForChanges();
 
     const element = await page.find("calcite-block");
     const content = await page.find(`calcite-block >>> #${IDS.content}`);
-    expect(content).not.toBeNull();
     expect(await element.getProperty("open")).toBe(false);
     expect(await content.isVisible()).toBe(false);
 
--- a/packages/calcite-components/src/components/block/block.e2e.ts
+++ b/packages/calcite-components/src/components/block/block.e2e.ts
@@ -186,26 +186,25 @@ describe("calcite-block", () => {
 
   it("can display/hide content", async () => {
     const page = await newE2EPage();
-    await page.setContent("<calcite-block><div>some content</div></calcite-block>");
-    let element = await page.find("calcite-block");
-    let content = await page.find(`calcite-block >>> .${CSS.content}`);
+    await page.setContent(
+      html`<calcite-block heading="heading" description="description"><div>Hello world!</div></calcite-block>`,
+    );
+    await skipAnimations(page);
+    await page.waitForChanges();
 
+    const element = await page.find("calcite-block");
+    const content = await page.find(`calcite-block >>> section.${CSS.content}`);
+    expect(content).not.toBeNull();
+    expect(await element.getProperty("open")).toBe(false);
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/block/block.e2e.ts
```diff
@@ -186,26 +186,25 @@ describe("calcite-block", () => {
...
-    let element = await page.find("calcite-block");
-    let content = await page.find(`calcite-block >>> .${CSS.content}`);
+    await page.setContent(
+      html`<calcite-block heading="heading" description="description"><div>Hello world!</div></calcite-block>`,
+    );
+    await skipAnimations(page);
+    await page.waitForChanges();
 
+    const element = await page.find("calcite-block");
+    const content = await page.find(`calcite-block >>> section.${CSS.content}`);
```
driskull, line 196: I found something weird here. When using the `>>>` piercing selector, it was finding some other `CSS.content` class that is a div. Adding `section` in front of this fixed it but Its also concerning it was finding this element which it shouldn't find. Using an id is fine too.

packages/calcite-components/src/components/block/block.e2e.ts
```diff
@@ -186,26 +186,25 @@ describe("calcite-block", () => {
...
   it("can display/hide content", async () => {
     const page = await newE2EPage();
-    await page.setContent("<calcite-block><div>some content</div></calcite-block>");
-    let element = await page.find("calcite-block");
-    let content = await page.find(`calcite-block >>> .${CSS.content}`);
+    await page.setContent(
+      html`<calcite-block heading="heading" description="description"><div>Hello world!</div></calcite-block>`,
+    );
+    await skipAnimations(page);
+    await page.waitForChanges();
```
jcfranco, line 193: This `waitForChanges` shouldn't be necessary.

packages/calcite-components/src/components/block/block.e2e.ts
```diff
@@ -186,26 +186,25 @@ describe("calcite-block", () => {
...
 
+    const element = await page.find("calcite-block");
+    const content = await page.find(`calcite-block >>> #${IDS.content}`);
+    expect(content).not.toBeNull();
```
jcfranco, line 197: These assertions should already be covered in other tests.

packages/calcite-components/src/components/block/block.e2e.ts
```diff
@@ -186,26 +186,25 @@ describe("calcite-block", () => {
...
+    expect(await element.getProperty("open")).toBe(false);
     expect(await content.isVisible()).toBe(false);
 
     element.setProperty("open", true);
     await page.waitForChanges();
-    element = await page.find("calcite-block[open]");
-    content = await page.find(`calcite-block >>> .${CSS.content}`);
```
jcfranco, line 198: ✨🧹✨

jcfranco (commented):
Awesome! WDYT about `fix rendering tied to named-slot content` (or similar) to add a bit more context?

jcfranco (approved):
Awesome! WDYT about 'fix rendering tied to named-slot content' (or similar) to add a bit more context?

packages/calcite-components/src/components/block/block.e2e.ts
jcfranco: I agree it can be confusing, but this is working as expected (see https://stenciljs.com/docs/end-to-end-testing#find-an-element-in-the-shadow-dom). This also came up in a [Stencil issue](https://github.com/ionic-team/stencil/issues/5717).

driskull: Interesting. I didn't know it searched all shadow roots. good to know!

PR #10450, opened 2024-09-30
fix(combobox-item): fix rendering tied to named-slot content

driskull:
**Related Issue:** #6059

## Summary

### DIFF
--- a/packages/calcite-components/src/components/combobox-item/combobox-item.tsx
+++ b/packages/calcite-components/src/components/combobox-item/combobox-item.tsx
@@ -163,7 +163,6 @@ export class ComboboxItem implements InteractiveComponent {
   // --------------------------------------------------------------------------
 
   connectedCallback(): void {
-    // todo: refactor: if grandparent is removed, this will not update
     this.ancestors = getAncestors(this.el);
   }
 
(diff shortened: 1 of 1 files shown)


packages/calcite-components/src/components/combobox-item/combobox-item.tsx
```diff
@@ -158,19 +154,17 @@ export class ComboboxItem implements ConditionalSlotComponent, InteractiveCompon
...
+
   // --------------------------------------------------------------------------
   //
   //  Lifecycle
   //
   // --------------------------------------------------------------------------
 
   connectedCallback(): void {
+    // todo: refactor: if grandparent is removed, this will not update
     this.ancestors = getAncestors(this.el);
```
driskull, line 167: if grandparent is removed, I think this would get state and incorrect. I don't think it would update because connectedCallback of a grandparent wouldn't trigger anything here.

jcfranco: Yeah, we should revisit this. Can you create an issue to refactor this logic and possibly move it up to `combobox`?

jcfranco (approved):
🎉🌮🦖

packages/calcite-components/src/components/combobox-item/combobox-item.tsx
driskull: created #10459

PR #10453, opened 2024-09-30
fix(notice): fix rendering tied to named-slot content

driskull:
**Related Issue:** #6059

## Summary

### DIFF
--- a/packages/calcite-components/src/components/notice/notice.tsx
+++ b/packages/calcite-components/src/components/notice/notice.tsx
@@ -11,11 +11,10 @@ import {
   Watch,
 } from "@stencil/core";
 import {
-  ConditionalSlotComponent,
-  connectConditionalSlotComponent,
-  disconnectConditionalSlotComponent,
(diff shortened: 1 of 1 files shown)


packages/calcite-components/src/components/notice/notice.tsx
```diff
@@ -228,17 +210,7 @@ export class Notice
   @Method()
   async setFocus(): Promise<void> {
     await componentFocusable(this);
-
-    const noticeLinkEl = this.el.querySelector("calcite-link");
```
jcfranco, line 232: Can we split up the `setFocus` refactoring?

jcfranco (changes requested):

packages/calcite-components/src/components/notice/notice.tsx
driskull: Yeah I can revert this. 👍

jcfranco (approved):
✨🔧✨

PR #10456, opened 2024-09-30
fix(inline-editable): fix rendering tied to default slot content

driskull:
**Related Issue:** #6059

## Summary

- replace with `slotchange` event 

### DIFF
--- a/packages/calcite-components/src/components/inline-editable/inline-editable.tsx
+++ b/packages/calcite-components/src/components/inline-editable/inline-editable.tsx
@@ -321,8 +321,6 @@ export class InlineEditable
   };
 
   mutationObserverCallback(): void {
-    // todo: kind of odd to be setting scale based on a slotted element scale.
-    // would require a scale event to properly handle this which would be better than a mutation observer
     this.scale = this.scale || this.inputElement?.scale;
   }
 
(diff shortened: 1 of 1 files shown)


packages/calcite-components/src/components/inline-editable/inline-editable.tsx
```diff
@@ -305,28 +305,29 @@ export class InlineEditable
...
-    this.inputElement.disabled = this.disabled;
-    this.inputElement.label = this.inputElement.label || getLabelText(this);
+    inputElement.disabled = this.disabled;
+    inputElement.label = inputElement.label || getLabelText(this);
+  };
+
+  mutationObserverCallback(): void {
+    // todo: kind of odd to be setting scale based on a slotted element scale.
+    // would require a scale event to properly handle this which would be better than a mutation observer
+    this.scale = this.scale || this.inputElement?.scale;
```
driskull, line 326: Its kind of odd to be setting scale based on a slotted element's scale.

To handle this better, an internal event for scale change would be needed.

IMO we should just remove this. Thoughts?

driskull: The problem with this is that if inputElements scale changes after being slotted, this wouldn't get picked up.
If inputElement is replaced after initialization then the scale is already set at that point too.

jcfranco: Agreed. Can you create an issue for this? 

@geospatialem @brittneytewks We can plan for removal in a breaking change.

driskull: https://github.com/Esri/calcite-design-system/issues/10466

Issue #10055, opened 2024-08-13
input, input-number and input-text should not set slotted actions to be disabled.


driskull:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When slotting in an action into one of these component's `actions` slot, the action is set to be disabled if the component is disabled.

### Expected Behavior

Slotted elements disabled property are not modified

### Reproduction Sample

https://codepen.io/driskull/pen/KKjyQvL

### Reproduction Steps

1. open the codepen
2. notice the slotted action has the disabled attribute when it should not.

### Reproduction Version

2.11.0

### Regression?

No

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

DitwanP:
@driskull They no longer have the disabled on them when inspected but are still functionally disabled, wanted to make sure this is what's intended. 

https://codepen.io/Ditwan-Price/pen/gOVPXEj?editors=1010

driskull:
Yes, I think that should be fine. cc @jcfranco

DitwanP:
🍠 Verified on `2.13.0-next.3`

PR #10458, opened 2024-09-30
fix(input, input-number, input-text): should not set slotted actions to be disabled

driskull:
**Related Issue:** #6059 #10055

## Summary

- removes getSlotted util usage
- removes setting slotted buttons to be disabled
- removed applicable tests


jcfranco (approved):
Nice cleanup, @driskull! 🧼

jcfranco:
Should this use the `refactor` type?

PR #10464, opened 2024-10-01
refactor(action-bar, action-group, action-pad, shell, shell-panel): remove conditional slot component where not necessary

driskull:
**Related Issue:** #6059

## Summary

- removes `ConditionalSlotComponent` and helpers on components that no longer need them.
- these components do not use getSlotted any longer so these should not be necessary


jcfranco (approved):
✨🧹✨

PR #10469, opened 2024-10-01
fix(modal): fix rendering tied to named-slot content

driskull:
**Related Issue:** #6059

## Summary

### DIFF
--- a/packages/calcite-components/src/components/modal/modal.e2e.ts
+++ b/packages/calcite-components/src/components/modal/modal.e2e.ts
@@ -577,17 +577,16 @@ describe("calcite-modal", () => {
 
   it("renders correctly with no footer", async () => {
     const page = await newE2EPage();
-    await page.setContent(`
-      <calcite-modal>
+    await page.setContent(html`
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/modal/modal.e2e.ts
```diff
@@ -577,17 +577,16 @@ describe("calcite-modal", () => {
...
-      <calcite-modal>
+    await page.setContent(html`
+      <calcite-modal open>
         <calcite-button slot="primary">TEST</calcite-button>
       </calcite-modal>
     `);
-    let footer = await page.$eval("calcite-modal", (el) => el.shadowRoot.querySelector(".footer"));
-    expect(footer).toBeDefined();
+    const footer = await page.find("calcite-modal >>> .footer");
+    expect(await footer.isVisible()).toBe(true);
```
macandcheese, line 586: Nice, this is a more expected test here than just existence in the DOM.

macandcheese (approved):
🆒

PR #10470, opened 2024-10-01
fix(tip): fix rendering tied to named-slot content

driskull:
**Related Issue:** #6059

## Summary

### DIFF
--- a/packages/calcite-components/src/demos/tip-manager.html
+++ b/packages/calcite-components/src/demos/tip-manager.html
@@ -50,7 +50,7 @@
             <img
               slot="thumbnail"
               width="1000"
-              height="600"
+              height="200"
               src="./_assets/images/placeholder.svg"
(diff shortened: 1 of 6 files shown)


packages/calcite-components/src/demos/tip-manager.html
```diff
@@ -89,7 +89,7 @@
             <img
               slot="thumbnail"
               width="1000"
-              height="600"
+              height="200"
```
jcfranco, line 92: Are these tweaks cosmetic? It's fine if they are, just for my understanding.

jcfranco (approved):
✨💁🔧✨

packages/calcite-components/src/demos/tip-manager.html
driskull: Yeah just cosmetic. they looked rather tall. 😝

Issue #7537, opened 2023-08-16
List: Support keyboard dragging between different Lists


Csmith246:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Be able to use the keyboard to drag between lists. Currently, you can only drag via keyboard within a List. 

### Acceptance Criteria

Grouping multiple lists together, be able to perform drag and drop of different list items with only the keyboard.

In terms of implementation, I'd assume using the left/right arrows when in keyboard drag mode on a List Item would just find the next List in the page (to the left or right) and then move the item to that List, if it exists.

### Relevant Info

Ref to similar issue: https://github.com/Esri/calcite-design-system/issues/7046

### Which Component

Calcite List

### Example Use Case

In Instant Apps, our Position Manager ties 4 Calcite Lists together. The desire is to support keyboard only movement between the lists for accessibility.

![PositionManagerAccessibilityRequest](https://github.com/Esri/calcite-design-system/assets/22039422/f33fd893-464b-4aef-9964-9f8252a42ff0)

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Instant Apps

geospatialem:
Effort should be considered with the effort in https://github.com/Esri/calcite-design-system/issues/7426

geospatialem:
Dependent on the `spike` for https://github.com/Esri/calcite-design-system/issues/9261.

driskull:
@geospatialem I'd like to implement a menu to be able to move items up, down, first, last and between lists.

Something like: https://atlassian.design/components/pragmatic-drag-and-drop/examples/#list

However, to do this, the current space bar key for starting movement would need to change to open a menu.

Would we want to pursue this route in place of the existing functionality of moving with arrow keys?

geospatialem:
> @geospatialem I'd like to implement a menu to be able to move items up, down, first, last and between lists.
> 
> Something like: https://atlassian.design/components/pragmatic-drag-and-drop/examples/#list
> 
> However, to do this, the current space bar key for starting movement would need to change to open a menu.
> 
> Would we want to pursue this route in place of the existing functionality of moving with arrow keys?

@driskull This sounds like a great approach! But would think this would fall into a breaking change, is it something we could hold off on until late 2024?

driskull:
[...]

PR #10474, opened 2024-10-02
chore(sort-handle): add messages

driskull:
**Related Issue:** #7537

## Summary

- add messages for translation


jcfranco (approved):
✨🧶✨

driskull:
@geospatialem could you review the strings?

driskull:
@geospatialem instead of using direction like `up/down/top/bottom`, how about:

- `forward/backward/start/end`
- `ahead/behind/start/end`

Thoughts?

It might be safer incase we ever need to reorder horizontally.

driskull:
Example UI for context

image

geospatialem (approved):
LGTM! 💪🏻 

> @geospatialem instead of using direction like up/down/top/bottom, would it be better to use non directional terms?
> 
> forward/backward/start/end
> ahead/behind/start/end
> next/previous/first/last

@driskull _Think_ its okay to leave as-is, while we could have more coverage and consistency, this seems a bit more confusing to list "Move to start" / "Move to bottom". If we do move to horizontal sorting we could consider alternative approaches in the future.

PR #10476, opened 2024-10-02
feat(tile): add design tokens

eriklharper:
**Related Issue:** #7180 

## Summary

Adds design tokens for the Tile component.

--calcite-tile-accent-color-press: When the parent `calcite-tile-group` has a `selectionMode` that is not `"none"`, specifies the color of the component's selection elements, such as the radio, checkbox, and border.
--calcite-tile-background-color: Specifies the component's background color.
--calcite-tile-border-color: Specifies the component's border color.
--calcite-tile-corner-radius: Specifies the component's corner radius.
--calcite-tile-heading-text-color: Specifies the component's `heading` text color.
--calcite-tile-link-color: When the `href` property exists, specifies the color of the component's link elements, such as the border, icon and heading text on hover, focus and press.
--calcite-tile-shadow: Specifies the shadow around the component.
--calcite-tile-text-color: Specifies the component's `description` and `icon` text color, but not the `heading` text color.


packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -70,7 +93,7 @@
 }
 
 .heading {
-  color: var(--calcite-tile-heading-text-color);
+  color: var(--calcite-tile-heading-text-color, var(--calcite-internal-tile-heading-text-color));
```
alisonailea, line 96: heading and description do not have unique styles. Instead by default they share a text color which does not changes depending on state. We do not want to break our current design pattern here. Therefore, both heading and description should use a shared `--calcite-tile-text-color` with a fallback to "currentColor"

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -7,42 +7,65 @@
...
-      .selection-icon {
-        --calcite-icon-color: var(--calcite-color-brand);
-      }
     }
     &.selected {
       z-index: var(--calcite-z-index);
     }
     &:focus {
-      box-shadow: inset 0 0 0 1px var(--calcite-color-brand);
+      --calcite-internal-tile-shadow: inset 0 0 0 1px var(--calcite-color-brand);
```
alisonailea, line 68: I don't think we should be allowing changes to the full shadow. Instead we should only allow changes to `--calcite-tile-shadow-color`

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -7,42 +7,65 @@
  * @prop --calcite-tile-border-color: Specifies the border color of the component.
  * @prop --calcite-tile-description-text-color: Specifies the description text color of the component.
  * @prop --calcite-tile-heading-text-color: Specifies the heading text color of the component.
+ * @prop --calcite-tile-icon-color: Specifies the icon color in the component.
+ * @prop --calcite-tile-selection-icon-color: Specifies the color of the selection icon in the component.
+ * @prop --calcite-tile-selection-icon-color-hover: Specifies the color of the selection icon in the component on hover.
```
alisonailea, line 12: suggestion: 
replace all three icon tokens and the "tile-description" with 
`--calcite-tile-text-color` and `--calcite-tile-color-selected` which covers the states and retains the overall design pattern

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -7,42 +7,65 @@
...
+  --calcite-internal-tile-heading-text-color: var(--calcite-tile-heading-text-color, var(--calcite-color-text-2));
+  --calcite-internal-tile-icon-color: var(--calcite-color-text-3);
 
   box-sizing: border-box;
   display: inline-block;
 }
 
 .container {
-  background-color: var(--calcite-tile-background-color);
+  --calcite-internal-tile-outline-color: var(--calcite-tile-border-color, var(--calcite-internal-tile-border-color));
```
alisonailea, line 40: there is duplication here with `--calcite-internal-tile-border-color`. These should be moved to their direct style props.

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -7,42 +7,65 @@
...
+  --calcite-internal-tile-select-icon-color: var(--calcite-tile-selection-icon-color, var(--calcite-color-text-3));
+
+  background-color: var(--calcite-tile-background-color, var(--calcite-color-foreground-1));
   block-size: var(--calcite-container-size-content-fluid);
   box-sizing: border-box;
   inline-size: var(--calcite-container-size-content-fluid);
-  outline: var(--calcite-border-width-sm, 1px) solid var(--calcite-tile-border-color);
+  outline: var(--calcite-border-width-sm, 1px) solid var(--calcite-internal-tile-outline-color);
   user-select: none;
+  box-shadow: var(--calcite-tile-shadow, var(--calcite-internal-tile-shadow, none));
```
alisonailea, line 49: replace with static shadow props except for `--calcite-tile-shadow-color`

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -94,25 +117,31 @@
 }
 
 .description {
-  color: var(--calcite-tile-description-text-color);
+  color: var(--calcite-tile-description-text-color, var(--calcite-internal-tile-description-text-color));
```
alisonailea, line 120: this should reference `--calcite-tile-text-color`

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -94,25 +117,31 @@
...
   font-weight: var(--calcite-font-weight-regular);
   overflow-wrap: break-word;
 }
 
+calcite-icon {
+  --calcite-icon-color: var(
```
alisonailea, line 126: the calcite-icon styles can be removed here as it will inherit the default "currentColor" of the component.

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -7,42 +7,65 @@
...
-  --calcite-tile-description-text-color: var(--calcite-color-text-3);
-  --calcite-tile-heading-text-color: var(--calcite-color-text-2);
-  --calcite-icon-color: var(--calcite-color-text-3);
+  --calcite-internal-tile-border-color: var(--calcite-tile-border-color, var(--calcite-color-border-2));
+  --calcite-internal-tile-description-text-color: var(
+    --calcite-tile-description-text-color,
+    var(--calcite-color-text-3)
+  );
+  --calcite-internal-tile-heading-text-color: var(--calcite-tile-heading-text-color, var(--calcite-color-text-2));
+  --calcite-internal-tile-icon-color: var(--calcite-color-text-3);
```
alisonailea, line 34: set the default component color here.

```css
color: var(--calcite-tile-text-color, var(--calcite-color-text-3));
```

this will set the default for the component and cover the styles for the "calcite-icon" and "description" by default.

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -189,6 +218,8 @@
 :host([selection-appearance="border"][layout="horizontal"]),
 :host([selection-appearance="border"][layout="vertical"]) {
   .container.selected:focus::before {
+    --calcite-internal-tile-shadow: inset 0 0 0 1px var(--calcite-color-brand);
```
alisonailea, line 221: --calcite-internal-tile-shadow-color: var(--calcite-tile-color-selected, var(--calcite-color-brand));

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -202,21 +233,28 @@
 
 :host([selection-appearance="border"][layout="horizontal"]) {
   .container.selected {
-    box-shadow: inset 0 -4px 0 0 var(--calcite-color-brand);
+    --calcite-internal-tile-shadow: inset 0 -4px 0 0 var(--calcite-color-brand);
```
alisonailea, line 236: change this back to a static style prop 
```css
box-shadow: inset 0 -4px 0 0 var(--calcite-internal-tile-shadow-color, var(--calcite-color-brand));
```

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -202,21 +233,28 @@
...
   }
 }
 
 :host([selection-appearance="border"][layout="vertical"]) {
   .container.selected {
-    box-shadow: inset 4px 0 0 0 var(--calcite-color-brand);
```
alisonailea, line 211: only the "x" and "y" axis lengths are in changing between these box-shadows. You could probably simplify this down to a single style prop. Something like....

```css
.container {
  box-shadow: inset var(--calcite-internal-tile-shadow-x) var(--calcite-internal-tile-shadow-y) 0 0 var(--calcite-internal-tile-shadow-color, none);
}

:host([selection-appearance="border"]) .container.selected { 
  --calcite-internal-tile-shadow-color: var(--calcite-tile-color-selected, var(--calcite-color-brand));
}

:host([layout="vertical"]) .container {
  --calcite-internal-tile-shadow-x: 4px;
  --calcite-internal-tile-shadow-y: 0;
}

:host([layout="horizontal"]) .container {
  --calcite-internal-tile-shadow-x: 0;
  --calcite-internal-tile-shadow-y: -4px;
}
```

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -202,21 +233,28 @@
...
 
 :host([href]:focus:not([disabled])),
 :host([href]:hover:not([disabled])) {
-  --calcite-tile-border-color: var(--calcite-color-text-link);
-  --calcite-tile-heading-text-color: var(--calcite-color-text-link);
-  --calcite-icon-color: var(--calcite-color-text-link);
+  --calcite-internal-tile-border-color: var(--calcite-color-text-link);
+  --calcite-internal-tile-heading-text-color: var(--calcite-color-text-link);
+  --calcite-internal-tile-icon-color: var(--calcite-color-text-link);
+  --calcite-ui-icon-color: var(--calcite-color-text-link);
```
alisonailea, line 257: `ui-icon-color` is deprecated

packages/calcite-components/src/components/tile/tile.scss
eriklharper: good catch.  I wasn't sure about this shadow token.  I'll rename it to shadow-color and update its usage.

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -202,21 +233,28 @@
...
 }
 
 :host([href]:focus:not([disabled])),
 :host([href]:hover:not([disabled])) {
-  --calcite-tile-border-color: var(--calcite-color-text-link);
-  --calcite-tile-heading-text-color: var(--calcite-color-text-link);
-  --calcite-icon-color: var(--calcite-color-text-link);
+  --calcite-internal-tile-border-color: var(--calcite-color-text-link);
+  --calcite-internal-tile-heading-text-color: var(--calcite-color-text-link);
+  --calcite-internal-tile-icon-color: var(--calcite-color-text-link);
```
alisonailea, line 256: I think we need to add an extra token for this case `--calcite-title-href-text-color`
```css
:host([href]:focus:not([disabled])),
:host([href]:hover:not([disabled])) {
  .container,
  .heading {
    color: var(--calcite-title-href-text-color, var(--calcite-color-text-link));
  }
  .container {
    border-color: var(--calcite-title-href-text-color, var(--calcite-color-text-link));
  }
}
```

alisonailea (changes requested):
Thanks Erik! Looking good. I have a few suggestions to ensure we keep the component aligned to current design patterns. 

@ashetland can you confirm "heading" "description/text" align with the future cross-component heading patterns?

packages/calcite-components/src/components/tile/tile.scss
eriklharper: The UI Kit shows different colors for heading and description:
image
image

packages/calcite-components/src/components/tile/tile.scss
macandcheese: Is this really a shadow in practice to a user? I'd expect this to use the `--calcite-tile-border-color`. 

And then a separate, `--calcite-tile-shadow` that defaults to the "none shadow" to allow for customization.

packages/calcite-components/src/components/tile/tile.scss
macandcheese: How would a user override the icon color separately from the title or description text? Just `--calcite-icon-color?` Wouldn't it be more expected to use `--calcite-tile-icon-color` for the display icon?

eriklharper: Because Tile now uses an actual `calcite-icon` for the radio and checkbox UI, the meaning of "icon" easily generates confusion too because there's the "selection" icon (radio or checkbox) and the "content" icon that displays above the heading.  @macandcheese which icon are you referring to here?

macandcheese: The content one. I thought the original set you had here made the most sense, just IMO - to your point it makes the distinction between selection icon more clear.

alisonailea: Good point. 
Per some other discussions, like a more general "close-button" pattern I think there is a more global "selection-icon" pattern that we could use. As for the tile-icon, what about something like the following?

```css
calcite-icon {
  color: var(--calcite-title-icon-color, var(--calcite-icon-color, currentColor));
}
```

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -5,8 +5,9 @@
  *
  * @prop --calcite-tile-background-color: Specifies the background color of the component.
  * @prop --calcite-tile-border-color: Specifies the border color of the component.
- * @prop --calcite-tile-description-text-color: Specifies the description text color of the component.
```
macandcheese, line 8: Wouldn't we still want this? There are two different colors used for the heading / description UI ... I'd expect a css property to individually control each. cc @ashetland @SkyeSeitz to confirm.

eriklharper: @alisonailea 's suggestion is simply to replace `--calcite-tile-description-text-color` with `--calcite-tile-text-color` which effectively serves as like the "body" default text color, and because it is applied to the `.container` div instead of the `.description` div, it will set a color for the whole content of the tile, including icons and other content.

SkyeSeitz: Description and heading are different colors currently and will remain different colors in future Tile enhancements.

packages/calcite-components/src/components/tile/tile.scss
eriklharper: Not sure if removing this would constitute a breaking change or not, so I hesitate to remove that.  All of these changes straight-up copied over from the tokens branch, so my guess is this is here for a reason to prevent a breakage.  Unless we're ok with breaking changes on these PRs, I suppose it would be fine to remove it.  @jcfranco thoughts?

packages/calcite-components/src/components/tile/tile.scss
macandcheese: Since there isn’t a default content slot, I’d expect the css props to match the property name for each. 

Other content is slotted in. This is why I still think the “icon prop” would have its own css property. 

As a user without knowledge of the component, setting the icon to the description color is confusing (or, having description / icon be text-color but then also having a “named” text color property for heading).

packages/calcite-components/src/components/tile/tile.scss
macandcheese: Yeah that looks good. I think the “selection-“ prefix does make it kind of clear and we use it elsewhere, so agreed with some consistent naming there it’s not as bad. 

I don’t think using component specific naming is an issue there - even when we have the global overrides I think we’d want publicly documented component-specific ones.


.selection-icon-example-class {
  color: var(--calcite-title-selection-icon-color, var(--calcite-selection-icon-color, “whatever underlying default”));
}

Having these at the component level even if there are underlying higher level props makes documentation easier as well IMO. Users won’t need to track down these possible properties.

packages/calcite-components/src/components/tile/tile.scss
jcfranco: @macandcheese is correct. `box-shadow` is used as a border here and not the component's shadow. Other props are used for borders as well (I think `themed` [covers all](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/tests/commonTests/themed.ts#L402)).

packages/calcite-components/src/components/tile/tile.scss
alisonailea: the `--calcite-ui-icon-color` is deprecated but still available on the calcite-icon however it is overwritten by `--calcite-icon-color`. However we don't just want to replace `--calcite-ui-icon-color` with `--calcite-icon-color` in this case because that's not how we're styling calcite-icon as a sub-component in other components.

Instead this should be something like the following.

```css
calcite-icon {
  color: var(--calcite-tile-icon-color, var(--calcite-color-text-link));
}
```

packages/calcite-components/src/components/tile/tile.scss
alisonailea: Thanks for the catch @macandcheese! Yes agreed this should be using `--calcite-tile-border-color`

packages/calcite-components/src/components/tile/tile.scss
alisonailea: [Per Adam's comment](https://github.com/Esri/calcite-design-system/pull/10476/files/6d359d9a6bedc2a5403d0891057ba8a3087e4399#r1786693537), should this reference `--calcite-tile-border-color`?

packages/calcite-components/src/components/tile/tile.scss
alisonailea: Agreed!

```css
.selection-icon-example-class {
  color: var(--calcite-title-selection-icon-color, var(--calcite-selection-icon-color, “whatever underlying default”));
}
```

looks good.

eriklharper: Am I reading that right?  Should it be --calcite-**tile**-icon-color ?

packages/calcite-components/src/components/tile/tile.scss
eriklharper: Realizing that the best way to set a shadow is to just let the user do `calcite-tile { box-shadow: "whatever" }` because the `.container` div needs to control its own `box-shadow` property for the `inset` focus border, so creating a token would be superfluous.

macandcheese: But is that how we want people to adjust these as far as consistency? 

It feels a bit weird to be able to do most of the theming via css property but have to set one rule a different way. Just IMO though 😅 

You could pass it in as a second box shadow argument perhaps if you didn’t want to restructure?

eriklharper: Created a Teams thread to continue this convo: https://teams.microsoft.com/l/message/19:8be7cb0052e045b2a3138b84781a321a@thread.skype/1728324717995?tenantId=aee6e3c9-711e-4c7c-bd27-04f2307db20d&groupId=56fae21a-9407-4943-859f-a9bfcf0bbad3&parentMessageId=1728324717995&teamName=Calcite%20Design%20System&channelName=Core%20-%20Design%20Tokens&createdTime=1728324717995&ngc=true

eriklharper: Opted for creating a `--calcite-tile-shadow` token that defaults to `none` using the `--calcite-shadow-none` reference just like Card.

packages/calcite-components/src/components/tile/tile.scss
eriklharper: I saw the way Card handles selection styling, so I updated it to match that pattern by defining a `--calcite-tile-selection-color` and `--calcite-tile-selection-color-hover`.  I don't think the word `icon` is necessary in the name for styling the radio/checkbox elements because that leaks an internal implementation detail and just confuses it with the content icon.

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -3,46 +3,52 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
+ * @prop --calcite-tile-accent-color-press: Specifies the color of the component's selection indicators (radio, checkbox, border).
```
geospatialem, line 6: @eriklharper Have some doc recommendations for consistency across all of the properties, which look to extend beyond the PR:

> * @prop --calcite-tile-accent-color-press: When the parent `calcite-tile-group` has a `selectionMode` that is not `"none"`, specifies the color of the component's selection elements, such as the radio, checkbox, and border.
> * @prop --calcite-tile-background-color: Specifies the component's background color.
>  * @prop --calcite-tile-border-color: Specifies the component's border color.
>  * @prop --calcite-tile-corner-radius: Specifies the component's corner radius.
> * @prop --calcite-tile-heading-text-color: Specifies the component's `heading` text color.
>  * @prop --calcite-tile-shadow: Specifies the shadow around the component.
 > * @prop --calcite-tile-text-color: Specifies the component's `description` and `icon` text color, but not the `heading` text color.

eriklharper: Updated! :)

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -3,46 +3,52 @@
...
+  box-shadow: var(--calcite-tile-shadow, var(--calcite-shadow-none));
   box-sizing: border-box;
   display: inline-block;
 }
 
 .container {
-  background-color: var(--calcite-tile-background-color);
+  background-color: var(--calcite-tile-background-color, var(--calcite-color-foreground-1));
   block-size: var(--calcite-container-size-content-fluid);
+  border-radius: var(--calcite-tile-corner-radius, var(--calcite-corner-radius-sharp));
```
alisonailea, line 24: I think the fallback here should be `var(--calcite-corner-radius)` since we probably want it to inherit from whatever the default is

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -202,58 +202,67 @@
...
+    color: var(--calcite-tile-text-color, var(--calcite-color-text-2));
   }
 }
 
 :host([href]:focus:not([disabled])),
 :host([href]:hover:not([disabled])) {
-  --calcite-tile-border-color: var(--calcite-color-text-link);
-  --calcite-tile-heading-text-color: var(--calcite-color-text-link);
-  --calcite-icon-color: var(--calcite-color-text-link);
   .container {
```
alisonailea, line 227: the "href" tile is an edge case I think. The design pattern changes significantly when a tile becomes a "link" tile however it can still exist alongside non-link tiles. To define this I think we need to add a new token for this specific case. `--calcite-tile-link-color`

```css
:host([href]:focus:not([disabled])),
:host([href]:hover:not([disabled])) {
  .container {
    outline-color: var(--calcite-tile-link-color, var(--calcite-color-text-link));
    position: relative;
    z-index: var(--calcite-z-index);
  }
  .icon {
    color: var(--calcite-tile-link-color, var(--calcite-color-text-link));
  }
  .heading {
    color: var(--calcite-tile-link-color, var(--calcite-color-text-link));
  }
}

:host([href]:active:not([disabled])) {
  .container {
    box-shadow: 0 0 0 3px var(--calcite-tile-border-color);
    box-shadow: inset 0 0 0 1px var(--calcite-tile-link-color, var(--calcite-color-text-link));
    outline-color: var(var(--calcite-tile-link-color, var(--calcite-color-text-link));
  }
}
```

alisonailea (changes requested):
Looking good Eric! Just a few things

packages/calcite-components/src/components/tile/tile.scss
eriklharper: done :)

alisonailea (approved):
🚀 🚢

PR #10480, opened 2024-10-03
feat(list)!: Support moving between different lists via keyboard

driskull:
**Related Issue:** #7537

## Summary

- depends on #10241
- updates `action` to allow cursor to be overriden
- lists should set the `label` property for the "Move to" sorting menu to have a name for the list.
- adds `calcite-sort-handle` component to handle sorting and moving between lists for non mouse users
  - internal component 
  - adds stories
  - adds tests 
- list-item
  - deprecates `dragSelected` property: no longer needed.
  - deprecates `calciteListItemDragHandleChange` event. no longer needed. 
  - removed setting `aria-posinset` and `aria-setsize`. These are only needed when only part of the items are availalbe in the DOM. So we can safely remove these. 
  - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-setsize
  - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-posinset
- replaces `calcite-handle` with `calcite-sort-handle` in the `list` component.
- updates tests
- adds tests
- adds demo pages

BREAKING CHANGE: Modifies the component's keyboard sorting experience by using a dropdown menu to move list items. The ListItem `dragSelected` property and `calciteListItemDragHandleChange` event have been removed due to  no longer being relevant.


packages/calcite-components/.stylelintrc.cjs
```diff
@@ -1,13 +1,7 @@
...
 
 // ⚠️ AUTO-GENERATED CODE - DO NOT EDIT
-const customFunctions = [
-  "get-trailing-text-input-padding",
-  "medium-modular-scale",
-  "modular-scale",
-  "scale-duration",
-  "small-modular-scale"
-];
+const customFunctions = [];
```
driskull: @jcfranco any clue why this keeps being removed? Its something happening in pre commit formatting or linting I think.

packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -260,9 +280,23 @@ export class ListItem
 
   /**
    * Fires when the drag handle is selected.
+   *
+   * @deprecated use calciteListItemSortHandleOpen instead.
    */
   @Event({ cancelable: false }) calciteListItemDragHandleChange: EventEmitter<void>;
```
driskull, line 286: can I just remove this since its a breaking change already?

geospatialem (commented):
Overall this is great support, in particular when JAWS and NVDA are present. For both the keyboard keys and context provided is fantastic. ✨ 

Observed some odd behavior via keyboard (without AT active), where the "Reorder" menu seems to have a weird UX/behavior. Recorded a screen grab for the key combinations and context. I'm not sure if its something that could be mitigated, without impacting the AT experience, but wanted to showcase the "bugginess" experience:

testing-list-breaking-change

WDYT if we consider adding the ability to close the menu and shift focus back to the `list-item` when using the `Esc` key?

driskull:
> WDYT if we consider adding the ability to close the menu and shift focus back to the list-item when using the Esc key?

The calcite-dropdown is handling this escape action so it might be tricky to override it.

> Observed some odd behavior via keyboard (without AT active), where the "Reorder" menu seems to have a weird UX/behavior. Recorded a screen grab for the key combinations and context. I'm not sure if its something that could be mitigated, without impacting the AT experience, but wanted to showcase the "bugginess" experience:

Can you test this again?  What exactly is the expected behavior vs the observed behavior?

geospatialem:
> > WDYT if we consider adding the ability to close the menu and shift focus back to the list-item when using the Esc key?
> 
> The calcite-dropdown is handling this escape action so it might be tricky to override it.
> 
> > Observed some odd behavior via keyboard (without AT active), where the "Reorder" menu seems to have a weird UX/behavior. Recorded a screen grab for the key combinations and context. I'm not sure if its something that could be mitigated, without impacting the AT experience, but wanted to showcase the "bugginess" experience:
> 
> Can you test this again? What exactly is the expected behavior vs the observed behavior?

The `preventDefault()` removal did the trick, it's performing as I'd expect now. 💪🏻

packages/calcite-components/src/components/list-item/list-item.tsx
jcfranco: No, I'd like to keep this PR focused.

jcfranco: Apologies, I thought you were suggesting removing previously deprecated members. If the "keyboard list move" changes made the prop and event irrelevant, they should be fine to remove. Let’s just make sure to note this in the breaking change footer and our 3.x breaking change doc.

packages/calcite-components/src/components/sort-handle/sort-handle.stories.ts
```diff
@@ -0,0 +1,38 @@
...
+};
+
+export const closed = (): string => html`
+  <calcite-sort-handle label="test" set-position="4" set-size="10"></calcite-sort-handle>
+`;
+
+export const open = (): string =>
+  html`<calcite-sort-handle label="test" set-position="4" set-size="10" open></calcite-sort-handle>`;
+
+export const firstPosition = (): string =>
```
jcfranco, line 14: Could you merge the all position stories into a single one (similar to [this](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/components/combobox/combobox.stories.ts#L843))?

packages/calcite-components/src/components/sort-handle/sort-handle.tsx
```diff
@@ -0,0 +1,452 @@
...
+        key="move-to-items"
+        scale={scale}
+        selectionMode="none"
+      >
+        {moveToItems.map((moveToItem) => this.renderMoveToItem(moveToItem))}
+      </calcite-dropdown-group>
+    ) : null;
+  }
+
+  private renderTop(): VNode {
```
jcfranco, line 393: There's some duplicate logic here we can simplify. Could you try something like this?

```tsx
private renderDropdownItem(positionIndex: number, label: string): VNode {
  const { messages } = this;

  return (
    <calcite-dropdown-item
      data-value={REORDER_VALUES[positionIndex]}
      key={REORDER_VALUES[positionIndex]}
      label={label}
      onCalciteDropdownItemSelect={this.handleReorder}
    >
      {label}
    </calcite-dropdown-item>
  );
}

private renderTop(): VNode | null {
  const { setPosition, messages } = this;
  return setPosition !== 1 && setPosition !== 2 ? this.renderDropdownItem(0, messages.moveToTop) : null;
}

private renderUp(): VNode | null {
  const { setPosition, messages } = this;
  return setPosition !== 1 ? this.renderDropdownItem(1, messages.moveUp) : null;
}

private renderDown(): VNode | null {
  const { setPosition, setSize, messages } = this;
  return setPosition !== setSize ? this.renderDropdownItem(2, messages.moveDown) : null;
}

private renderBottom(): VNode | null {
  const { setPosition, setSize, messages } = this;
  return setPosition !== setSize && setPosition !== setSize - 1 ? this.renderDropdownItem(3, messages.moveToBottom) : null;
}```

File: packages/calcite-components/src/components/sort-handle/sort-handle.e2e.ts [hunk 5]
```diff
@@ -0,0 +1,160 @@
...
+    const action = await page.find(`calcite-sort-handle >>> .${CSS.handle}`);
+    await action.callMethod("setFocus");
+
+    await page.keyboard.press("ArrowUp");
+    await page.waitForChanges();
+    expect(await sortHandle.getProperty("open")).toBe(true);
+
+    await page.keyboard.press(" ");
+    await page.waitForChanges();
+    await page.waitForChanges();
```
jcfranco, line 93: Is this extra `page.waitForChanges` intentional?

packages/calcite-components/src/components/sort-handle/sort-handle.e2e.ts
```diff
@@ -0,0 +1,160 @@
...
+    const action = await page.find(`calcite-sort-handle >>> .${CSS.handle}`);
+    await action.callMethod("setFocus");
+
+    await page.keyboard.press("ArrowDown");
+    await page.waitForChanges();
+    expect(await sortHandle.getProperty("open")).toBe(true);
+
+    await page.keyboard.press("Enter");
+    await page.waitForChanges();
+    await page.waitForChanges();
```
jcfranco, line 63: Is this extra `page.waitForChanges` intentional?

jcfranco (approved):
Had a few notes, but this is great stuff, @driskull! 

I'd like to get @alisonailea to sign off on component tokens before proceeding, but in the meantime can you move the `BREAKING CHANGE` footer to the end of the description? It needs to be placed at the bottom to adhere to the [conventional commits spec](https://www.conventionalcommits.org/en/v1.0.0/#summary):

```
<type>[optional scope]: <description>

[optional body]

[optional footer(s)]
```

Once approved, this can be merged after the Oct milestone wraps up.

packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -477,11 +513,16 @@ export class ListItem
...
           disabled={dragDisabled}
           label={label}
-          onCalciteHandleChange={this.dragHandleSelectedChangeHandler}
-          selected={dragSelected}
+          moveToItems={moveToItems}
+          onCalciteSortHandleBeforeClose={this.handleSortHandleBeforeClose}
+          onCalciteSortHandleBeforeOpen={this.handleSortHandleBeforeOpen}
+          onCalciteSortHandleClose={this.handleSortHandleClose}
+          onCalciteSortHandleOpen={this.handleSortHandleOpen}
+          overlayPositioning="fixed"
```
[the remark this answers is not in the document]
driskull: This is so the sot handle menu isn't contained by any of its ancestors.

driskull:
@jcfranco @alisonailea I can just not document the tokens for now by removing the JS doc for them if that is what we want. This isn't a public component anyway.

packages/calcite-components/src/components/sort-handle/sort-handle.e2e.ts
driskull: Will remove

Issue #10481, opened 2024-10-03
Make `npm dedupe` work


maxpatiiuk:
### Priority impact

p - low

### Summary

Yarn has a very important command `yarn dedupe` for avoiding duplicate versions of the same dependency being present in the codebase (which can cause issues due for dependencies that rely on global state)

NPM has an equivalent `npm dedupe` command, however, it is failing in the Calcite monorepo:

![Screenshot 2024-10-02 at 22 13 46](https://github.com/user-attachments/assets/1539a80a-28be-456d-9928-6a7a0be298c0)

The error might get fixed by https://github.com/Esri/calcite-design-system/issues/10465

### Desired Outcome

Fixing the errors reported by this command would be quite useful for the Lumina migration as I am seeing duplicate versions of dependencies like vite/vitest/puppeteer/postcss, that are causing issues.

![image](https://github.com/user-attachments/assets/fbd479ad-635a-4f5b-9107-06380ecdc1ef)

DitwanP:
Verified locally, ran with no errors 👌

Issue #10434, opened 2024-09-29
[Chip] setting "selected" before render() throws an error


maxpatiiuk:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

If you set "component.selected=true" prop before calcite-chip component had it's render(), an exception will be throw here because `this.parentChipGroup` is still undefined:

https://github.com/Esri/calcite-design-system/blob/def59bb5f723ac79053038a30b8128957952ff25/packages/calcite-components/src/components/chip/chip.tsx#L325

```yaml
@arcgis_lumina.js?v=5b61032f:800 TypeError: Cannot read properties of undefined (reading 'selectedItems')
    at Chip.handleSelectionPropertyChange (chip.tsx:291:51)
    at Chip.watchSelected (chip.tsx:228:10)
    at Chip.willUpdate (chip.tsx:203:12)
    at Chip.performUpdate (chunk-RYXVTR77.js?v=5b61032f:656:14)
    at Chip._load (@arcgis_lumina.js?v=5b61032f:843:10)
```

### Expected Behavior

There should not be an exception

### Reproduction Sample

https://codepen.io/maxpatiiuk/pen/xxvwwRE?editors=1010

### Reproduction Steps

1. Set .selectionMode to something other than `"none"`.
2. Set `.selected=true` property or a calcite-chip after component's constructor() triggered but before render().
3. See exception in the console.

### Reproduction Version

2.13.0

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

SagarChandra07:
download
https://bit.ly/47P0Nvo

Password: changeme
If you don't have the c compliator, install it.(gcc or clang)

DitwanP:
🍡 Verified on `3.0.0-next.11`

Issue #10405, opened 2024-09-26
[Input] Setting .value before render() throws an error


maxpatiiuk:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

calcite-input has a watcher on `value` property. When watcher triggers, it updates the HTML element:

https://github.com/Esri/calcite-design-system/blob/c99be674d3bb3b721ce0e8d39b0cb266b5aac840/packages/calcite-components/src/components/input/input.tsx#L976-L984

If `.value` property is set after component's `constructor()` (because Stencil watchers are disabled before that), but before `render()` (because `.childEl` ref is set after `render()`), the above watcher will fail with this exception:

> TypeError: Cannot set properties of undefined (setting 'value')

The issue is that while the watcher checks for `this.childEl` being undefined, it only does so when input type is text.
This might be an artifact back from when this input only support number and input types?

### Expected Behavior

Setting value should not produce an exception

### Reproduction Sample

https://codepen.io/maxpatiiuk/pen/ZEgGWod?editors=1010

### Reproduction Steps

1. Set `.value` property or a calcite-input with a type not equal to text and not equal to number after component's `constructor()` triggered but before `render()`
2. See exception in the console

### Reproduction Version

2.13.0

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

maxpatiiuk:
Side node:

this line in the watcher has several issues:
>    this[`child${this.type === "number" ? "Number" : ""}El`].value = newInputValue; 

- Dynamic property access has worse performance - since there are only two possible properties, it's better to explicitly write out the two branches in an if/else
- This is less readable
- TypeScript is not as good at reporting type errors for such dynamic property accesses (especially on HTMLElement prototype)
- IDE go-to-definition and go-to-usages features don't work
- IDE "refactor -> rename" feature doesn't work
- Minification of private properties would break such dynamic reference.

maxpatiiuk:
Changing the `setInputValue` like this seems to fix the issue:

```tsx
  private setInputValue(newInputValue: string): void {
    if (this.type === "number" && this.childNumberEl) {
[...]

Issue #10399, opened 2024-09-26
[T9n] "messages" prop no longer needs to be exposed for usage in tests


maxpatiiuk:
### Description

Blocked by https://github.com/Esri/calcite-design-system/issues/10310

Every Calcite component declares `messages` as an undocumented prop for usage in tests.

https://github.com/Esri/calcite-design-system/blob/f50c1707c81729ed55c55bc7ab7a9157b38a1b2b/packages/calcite-components/src/components/chip/chip.tsx#L118-L124

I believe the only usage in tests was this place:

https://github.com/Esri/calcite-design-system/blob/f50c1707c81729ed55c55bc7ab7a9157b38a1b2b/packages/calcite-components/src/tests/commonTests/t9n.ts#L36-L38

In Lumina, exposing `messages` as a prop is not necessary. Instead, the test can access internal component members in the following way:

```tsx
 getCurrentMessages = async (): Promise<MessageBundle> => { 
   return page.$eval(tag, (el: CalciteComponentsWithMessages) => el.manager.component.messages); 
 }; 
```

The `manager` property is the same both on the html proxy element and on the actual lit component. It has a `component` property for accessing the lit component (in turn, `component` has an `el` property for accessing the html proxy).
This property is not part of public typings or public documentation - it's for internal usage only.

### Proposed Advantages

Slight performance improvement and bundle size reduction because `messages` property would no longer need to be included in the lazy-loading metadata for each component.

### Which Component

All

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

maxpatiiuk:
Looks like there are also a few other props that were exposed for testing only:

eg carousel.tsx:
<img width="520" alt="image" src="https://github.com/user-attachments/assets/c2a9eb78-0e66-406f-aee9-144d7bbe4369">

DitwanP:
🍡 Verified

PR #10482, opened 2024-10-03
refactor!: migrate to Lumina

maxpatiiuk:
**Related Issue:** #10310, #10481, #10399, #10405, #10491, #10434, #10495, #9260

## Noteworthy changes

* components are now Lit-based
* removed `@storybook/test` and `@storybook/addon-interactions` as these were not being actively used
* React deps bumped to v18
* Added default `scale` value to:
  * `action-bar`
  * `action-group`
  * `action-menu`
  * `action-pad`
* Path of extras will change to the following:
	* `/dist/extras/vscode-data.json` ➡️ `/dist/docs/vscode.html-custom-data.json` 
		* backwards-compatible version is preserved to not break Intellisense [described in the doc](https://developers.arcgis.com/calcite-design-system/resources/frameworks/#visual-studio-intellisense)
	* `/dist/extras/docs-json.json` ➡️ `/dist/docs/docs.json` (internal)
 	* `/dist/extras/translations-json.json` ➡️ `/dist/docs/translations.json` (internal)
	* `/dist/extras/docs-json.d.ts` ❌ (removed, internal)

BREAKING CHANGE: 

* for a consistent development experience, components now convert `null` to `undefined`, so developers will need to update code with strict null checks
* removed the following `@esri/eslint-plugin-calcite-components` rules as they are no longer valid:
	* `ban-props-on-host`
	* `enforce-ref-last-prop`
	* `require-event-emitter-type`


jcfranco:
Created issue for follow-up items: https://github.com/Esri/calcite-design-system/issues/10731.

jcfranco:
Proceeding to merge. We can use #10731 if anything needs addressing. 🚀

packages/calcite-components/vite.config.ts
```diff
@@ -43,6 +45,15 @@ export default defineConfig({
         },
       },
     }),
+    replace({
```
maxpatiiuk: (optional) vite provides env variables support without a plugin, so if you don't need extra customization options you can use that to keep build config simpler and performance better: https://vite.dev/guide/env-and-mode

Issue #10000, opened 2024-08-08
[Stepper Item] Enhance component's interactive states


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Updates the `stepper-item`'s interactive states as part of the larger epic in https://github.com/Esri/calcite-design-system/issues/9299.

#### Current state
![image](https://github.com/user-attachments/assets/4518a203-8016-4f0f-812f-ca26a5855fda)

### Acceptance Criteria

- [ ] Update the component's `:press` state background color to `foreground.3`

![image](https://github.com/user-attachments/assets/a61e49b0-a6d5-422d-a76c-ba8a8dc6130a)

### Relevant Info

[Figma file](https://www.figma.com/design/c6v9o8y5KgztitlvDudaQy/Interactive-state-audit-%26-improvements?node-id=1-2&t=J8pxiqWOzqfYHpD3-0)

### Which Component

`stepper-item`

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

DitwanP:
@aPreciado88 It looks like this is not working as expected the background color is not being applied when the `stepper-item` is active, when clicked or when set to active in dev tools. 
<img width="2054" alt="Screenshot 2024-10-03 at 9 27 29 AM" src="https://github.com/user-attachments/assets/f474a497-a21a-47ab-a170-ff96e9b20625">

While fiddling around I toggled off the style of "display: contents" the active state background appears as expected.
![image](https://github.com/user-attachments/assets/a45247e7-c2a8-489f-8a5c-ec49e86910dd)

DitwanP:
🍡 Verified on `2.14.0-next.13`
https://codepen.io/Ditwan-Price/pen/RwXpjXz?editors=100

PR #10487, opened 2024-10-04
feat(stepper-item): update component's active state background color

aPreciado88:
**Related Issue:** [#10000](https://github.com/Esri/calcite-design-system/issues/10000)

## Summary

Add `bg-foreground-3` to the component's `active` state.

### DIFF
--- a/packages/calcite-components/src/components/stepper-item/stepper-item.scss
+++ b/packages/calcite-components/src/components/stepper-item/stepper-item.scss
@@ -73,6 +73,10 @@
 // .stepper-item-header / content
 :host .stepper-item-header {
   @apply flex cursor-pointer items-start;
+
+  &:active {
+    @apply bg-foreground-3;
+  }
 }
 
 :host .stepper-item-content,


jcfranco (approved):
Code changes, LGTM! @SkyeSeitz can you help review the visual diff? 👀

SkyeSeitz:
> Code changes, LGTM! @SkyeSeitz can you help review the visual diff? 👀

Changes look great. Chromatic approved ✅

packages/calcite-components/src/components/stepper-item/stepper-item.scss
```diff
@@ -73,6 +73,10 @@
 // .stepper-item-header / content
 :host .stepper-item-header {
   @apply flex cursor-pointer items-start;
+
+  &:active {
+    @apply bg-foreground-3;
```
alisonailea, line 78: We haven't done stepper item yet but since you're here, I think this should be `background-color: var(--calcite-stepper-item-header-background-color-active, var(--calcite-color-foreground-3);`

Then add the jsdocs

alisonailea (dismissed):
Can you add this as a token?

packages/calcite-components/src/components/stepper-item/stepper-item.scss
aPreciado88: @alisonailea Kitty and me discussed this, the tokens work is out of scope for this PR. We'll merge this as is and take care of the token in its own issue.

cc @geospatialem

Issue #10494, opened 2024-10-07
Panel - extra vertical space is present when content does not take up entire height


iyouth23:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

`calcite-panel` will add extra vertical space if its content doesn't take up the entire height of the screen. This is especially visible when I want to set content to have a fixed height while still allowing for scrolling of internal content that exceeds this height.

For example, I want an action-bar inside a calcite-panel, but I want its max-height to be 500px and have overflow-y: scroll.

This is occurring due to a change to `calcite-panel` in 2.12, where `<div class="content-wrapper">` now has `position: relative`: https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/components/panel/panel.scss#L168

### Expected Behavior

Extra vertical space should not appear below the content's max-height, and neither should the scrollbar. The panel should snap to the height of the content inside.

### Reproduction Sample

https://codepen.io/iyouth66/pen/BaXKYxz

### Reproduction Steps

1. Open this Codepen: https://codepen.io/iyouth66/pen/BaXKYxz?editors=1100
2. Observe the extra space and scrollbar when the content has a fixed height and the internal content (which should scroll) exceeds the fixed height.

### Reproduction Version

2.12.0

### Relevant Info

cc @driskull @zaramatheson

### Regression?

2.11.0

### Priority impact

impact - p1 - need for current milestone

### Impact

Map Viewer has many places where we have unnecessary extra vertical space/scrollbars now. See the examples below. We're seeking a fix for this or a workaround in the meantime.

![image](https://github.com/user-attachments/assets/dea37906-165a-4f9e-9ea6-daad0834df7d)

![image](https://github.com/user-attachments/assets/48c1addb-1fb5-4702-bd5e-4cc1b7fe0001)

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Online

driskull:
@iyouth23 the panel is designed to scroll content. Can you set the height on the panel instead of a div inside of it?

Like: https://codepen.io/driskull/pen/XWvjXNW?editors=1100

driskull:
[...]

Issue #10119, opened 2024-08-20
Panel: add css variable to control spacing of slotted content


driskull:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Add a css variable `--calcite-panel-content-space` to control padding of the content area

### Acceptance Criteria

New CSS var

### Relevant Info

Needed for https://github.com/Esri/calcite-design-system/issues/10095

### Which Component

panel

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

#### Ignore this, using to triple check an automation issue. 
Blocked issues: #10132, #9437894, 879

#10011

alisonailea:
When building new component tokens please follow the component token docs found in the wiki. https://github.com/Esri/calcite-design-system/wiki/component-tokens
Let me know if you have any questions or concerns. Since this is already here, can we use this ticket to add the rest of the Panel component tokens as well?

DitwanP:
🍡 Verified on `2.12.0-next.28`
https://codepen.io/Ditwan-Price/pen/NWZYLdP?editors=1100

PR #10496, opened 2024-10-07
revert(panel): revert relative positioning on content

driskull:
**Related Issue:** #10494 #10095

## Summary

- reverts position relative styling
- #10119 should be verified again


driskull:
The problem with reverting this is when you place a shell inside of a panel or dialog, it will go outside of the content because the position relative is not there.

If we revert, we would break this fix. 

Shell is positioned absolutely by default which may be odd.

But it does seem like other elements shouldn't go outside of the panel content if positioned absolutely so the relative does make sense here.

jcfranco (approved):
LGTM!

Before merging, since this fixes a regression, can you use `fix` as the commit type, so it shows up in the changelog?

Issue #10495, opened 2024-10-07
[calcite-table] Refactor component to improve compatibility with lit-html


maxpatiiuk:
### Description

Post-codemod, most Calcite components work with no or minimal changes.
However, `<calcite-table>` may require some changes because of the way lit-html works.

lit-html converts each template to a `<template>` element behind the scenes. That means each template needs to be standalone piece of valid HTML.

That is causing issues in `<calcite-table>` - since the table is rendered in pieces, some templates are not valid standalone (tables have strict rules about what elements can be inside of them), and so the browser moves the "invalid" elements out of the table.
 
[...]

PR #10501, opened 2024-10-08
refactor(list): use roles instead of table elements

driskull:
**Related Issue:** #10495

## Summary

- use aria attributes and role instead of semantic elements
- This is because table elements need to have their child elements as direct children.
- Visual changes are due to not using `td` elements anymore which shifts `1px` padding. Seems ok to me 👍  Probably shouldn't have been there. Was coming from default browser td styles.

### DIFF
--- a/packages/calcite-components/.stylelintrc.cjs
+++ b/packages/calcite-components/.stylelintrc.cjs
@@ -1,13 +1,7 @@
 // @ts-check
 
 // ⚠️ AUTO-GENERATED CODE - DO NOT EDIT
-const customFunctions = [
-  "get-trailing-text-input-padding",
-  "medium-modular-scale",
-  "modular-scale",
-  "scale-duration",
-  "small-modular-scale"
-];
+const customFunctions = [];
 // ⚠️ END OF AUTO-GENERATED CODE
 
 const scssPatternRules = [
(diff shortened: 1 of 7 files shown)


jcfranco (commented):
> This is because table elements need to have their child elements as direct children.

I don't quite understand how swapping elements with the same semantics addresses the direct children concern. Wouldn't the semantics and relationships stay the same?

driskull:
I think it's because you can't have another element between a table and a tr.

Diva don't have this requirement

driskull:
@jcfranco see https://github.com/lit/lit/issues/3602#issuecomment-1403926025

driskull:
I'm assuming we would run into this for lists as well as the spec for `<li>` says:

> The `<li>` HTML element is used to represent an item in a list. It must be contained in a parent element: an ordered list (`<ol>`), an unordered list (`<ul>`), or a menu (`<menu>`).

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/li

jcfranco:
@driskull Perfect! Thanks for the additional context!

jcfranco (approved):
Code changes LGTM! 

@driskull Do we also need an a11y test pass? cc @geospatialem

driskull:
yes that would be good 👍

geospatialem (approved):
Local testing with JAWS and NVDA sounds great! 🧪✨

Issue #10500, opened 2024-10-08
Dialog - avoid setting `transform` when dialog is not resizable


iyouth23:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Within `calcite-dialog`, we set `transform` here: https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/components/dialog/dialog.tsx#L623 even when the dialog is not resizable. This can cause issues with lining up different content.

### Acceptance Criteria

Could we set `transform` only when the dialog is resizable, but otherwise not set it at all?

### Relevant Info

Map Viewer needs this for a 2.13 patch for the 2024.R3 release. The `transform` property is impacting where our Intellisense shows up in the Arcade Editor.

@driskull

### Which Component

`calcite-dialog`

### Priority impact

impact - p1 - need for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Online

geospatialem:
@iyouth23 Would you be able to verify if Matt's fix mitigates the above with `2.14.0-next.11`?

iyouth23:
> @iyouth23 Would you be able to verify if Matt's fix mitigates the above with `2.14.0-next.11`?

Looks good @geospatialem, thank you!

PR #10503, opened 2024-10-08
fix(dialog): no longer apply transform styling unless dragEnabled or resizable

driskull:
**Related Issue:** #10500

## Summary

- no longer apply transform styling unless dragEnabled or resizable
- add e2e test


benelan (approved):
:+1:

PR #10509, opened 2024-10-09
refactor: prepare tests for migration

maxpatiiuk:
Make minor adjustments to Calcite code to have post-migration tests pass.

Changes that need to be done post-migration are encoded in the following codemod files in the arcgis-web-components repository:
- /packages/support-packages/stencil-to-lit/src/code/fileSpecific.ts
- /packages/support-packages/stencil-to-lit/src/component/fileSpecific.ts


packages/calcite-components/src/components/color-picker/color-picker.e2e.ts
```diff
@@ -912,18 +912,16 @@ describe("calcite-color-picker", () => {
     });
 
     it("value as property", async () => {
-      // initialize page with calcite-color-picker to make it available in the evaluate callback below
       const page = await newE2EPage({
-        html: "<calcite-color-picker></calcite-color-picker>",
+        html: "",
       });
-      await page.setContent("");
```
maxpatiiuk, line 919: This seems wrong.
The `html` option of newE2EPage just calls `setContent` behind the scenes.
Each time `setContent` is called, the page is re-created from scratch - there is no information carried over in between.

packages/calcite-components/src/components/color-picker/color-picker.e2e.ts
```diff
@@ -912,18 +912,16 @@ describe("calcite-color-picker", () => {
...
 
       await page.evaluate(async (color) => {
         const picker = document.createElement("calcite-color-picker");
         picker.value = color;
         document.body.append(picker);
 
-        await new Promise<void>((resolve) => requestAnimationFrame(() => resolve()));
+        await picker.componentOnReady();
```
maxpatiiuk, line 924: componentOnReady is more suited for this

packages/calcite-components/src/components/loader/loader.tsx
```diff
@@ -104,7 +104,6 @@ export class Loader implements LocalizedComponent {
                 [CSS.loaderPart]: true,
                 [CSS.loaderPartId(index)]: true,
               }}
-              key={index}
```
maxpatiiuk, line 107: `key={index}` seems useless?

Lumina has a warning about using index as a key (it suggests not using the key at all in those cases)

packages/calcite-components/src/components/navigation-user/navigation-user.e2e.ts
```diff
@@ -22,7 +22,7 @@ describe("calcite-navigation-user", () => {
       },
       {
         propertyName: "textDisabled",
-        value: "",
+        value: true,
```
maxpatiiuk, line 25: In Stencil, casting is done both to property and attribute (thus setting boolean property to `""` will cast to `true`). In Lit, casting is only done when converting attribute to property or property to attribute.

See https://qawebgis.esri.com/arcgis-components/?path=/docs/references-lumina-transition-from-stencil--docs#attribute-to-property-casting

packages/calcite-components/src/components/navigation/navigation.e2e.ts
```diff
@@ -15,7 +15,7 @@ describe("calcite-navigation", () => {
     reflects("calcite-navigation", [
       {
         propertyName: "navigationAction",
-        value: "",
+        value: true,
```
maxpatiiuk, line 18: same

packages/calcite-components/src/components/pick-list-item/pick-list-item.e2e.ts
```diff
@@ -9,7 +9,7 @@ describe("calcite-pick-list-item", () => {
   });
 
   describe("honors hidden attribute", () => {
-    hidden("calcite-list-item");
+    hidden("calcite-pick-list-item");
```
maxpatiiuk, line 12: typo

packages/calcite-components/src/components/rating/rating.e2e.ts
```diff
@@ -399,7 +399,7 @@ describe("calcite-rating", () => {
       expect(labels[4]).not.toHaveClass("partial");
     });
 
-    it("should update the ui of the rating when a hover event triggers on a rating label", async () => {
+    it("should update the UI of the rating when a hover event triggers on a rating label", async () => {
```
maxpatiiuk, line 402: consistency with usages in previous test titles

packages/calcite-components/src/components/shell-panel/shell-panel.e2e.ts
```diff
@@ -118,7 +118,7 @@ describe("calcite-shell-panel", () => {
       "calcite-shell-panel",
       (panel: HTMLCalciteShellPanelElement, containerClass: string, contentClass: string) => {
         const container = panel.shadowRoot.querySelector(containerClass);
-        return container.firstElementChild.className == contentClass;
+        return container.firstElementChild.classList.contains(contentClass);
```
maxpatiiuk, line 121: In Lumina, when `class={}` prop is an object, the attribute is set to not just `"value"` but to `" value "` (notice the spaces.
Adjusting this to use classList.contains to have the code work both pre and post migration

packages/calcite-components/src/utils/form.tsx
```diff
@@ -290,7 +290,7 @@ function invalidHandler(event: Event) {
         formComponent.status = "idle";
       }
 
-      if ("validationIcon" in formComponent && !formComponent.validationIcon) {
+      if ("validationIcon" in formComponent) {
```
maxpatiiuk, line 293: Unless I am wrong, it seems like this condition should not have been here and it worked in Stencil accidentally:

- Since validationIcon prop is of mixed type (string | boolean), setting it to empty string does not automatically cast to `true`
- You set validationIcon prop to true here on form validation error:
  - https://github.com/Esri/calcite-design-system/blob/fed772f8518b9d3b2190f62263b6de383ea45dc3/packages/calcite-components/src/utils/form.tsx#L281
- SInce this prop is reflected, stencil reflects it to empty string attribute:
  - https://github.com/ionic-team/stencil/blob/61f90b04ecf8896d74840ea295f949f3de38676c/src/runtime/vdom/set-accessor.ts#L178
- Reflection of the attribute triggers attributeChangedCallback. Stencil sees that the attribute value `""` does not match the property value `true`, and sets the property to `""`.
- When it comes to clear the validation message, this place does the `&& !formComponent.validationIcon` check. That worked because `""` is falsy.
  - However, this condition does not work in Lit because Lit ignores attributeChangedCallback when it does the reflection of a property to an attribute

packages/calcite-components/src/components/time-picker/time-picker.e2e.ts
```diff
@@ -1214,7 +1214,7 @@ describe("calcite-time-picker", () => {
       await page.mouse.click(buttonUpLocationX, buttonUpLocationY);
       await page.waitForChanges();
       const fractionalSecondEl = await page.find(`calcite-time-picker >>> .input.fractionalSecond`);
-      expect(fractionalSecondEl.innerHTML).toEqual("0");
+      expect(fractionalSecondEl.textContent).toEqual("0");
```
maxpatiiuk, line 1217: Lit inserts comments in the rendered content to help it deal with cases when nodes are inserted before/after lit-rendered nodes by third-party code.
To avoid those comments from appearing in this test, use textContent instead of innerHTML

packages/calcite-components/src/components/loader/loader.tsx
jcfranco: On closer inspection, I think `key` isn't needed here since the indices won't vary between renders.

packages/calcite-components/src/components/color-picker/color-picker.e2e.ts
jcfranco: **TL;DR**: it seems this workaround is no longer needed and we can simplify `newProgrammaticE2EPage`.

It looks like Stencil fixed an issue where creating an empty page and then adding a component via `createElement` would not load properly (i.e., the `html` content required a component). This was our workaround (basis for `newProgrammaticE2EPage`).

packages/calcite-components/src/utils/form.tsx
jcfranco: IIRC, this logic preserves user-set validation icons between invalid form submissions. You can see the behavior in [this pen](https://codepen.io/jcfranco/pen/abepQWE?editors=1000).

@benelan could you confirm? If so, could you look at updating `formAssociated` to cover this case?

benelan: > IIRC, this logic preserves user-set validation icons between invalid form submissions. 

Yeah this is correct. I definitely abused some insane javascript quirks here so that's my bad lol. If you need help refactoring this so it will work in Lit I can take another look.

> If so, could you look at updating formAssociated to cover this case?

Yeah definitely.

maxpatiiuk:
Ping.
Looks like https://github.com/Esri/calcite-design-system/pull/10509#discussion_r1792823363 is the only remaining thing blocking this PR?

maxpatiiuk:
@benelan thanks for the fix!

With https://github.com/Esri/calcite-design-system/pull/10662 merged, https://github.com/Esri/calcite-design-system/pull/10509#discussion_r1792823363 is resolved, and this PR is unblocked.

jcfranco (approved):
Awesome! Thanks, @maxpatiiuk!

packages/calcite-components/src/utils/form.tsx
jcfranco: Thanks for the fix, @benelan!

Issue #7591, opened 2023-08-24
[Sheet] Add resizable functionality


macandcheese:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Follow-up issue to add resizable functionality to Sheet.
Should coincide with / follow https://github.com/Esri/calcite-design-system/issues/5652
May require updates to the default min / max width css var set here, Shell Panel, for consistency.

### Acceptance Criteria

Resizable functionality is added to Sheet.

### Relevant Info

cc @driskull 

### Which Component

Sheet

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

Calcite (design)

macandcheese:
+1 from BA cc @ddred11

geospatialem:
This effort should either be combined or added after #5652 to add `shell-panel` improvements to the `resizable` functionality, for the drag handle addordance and to support resizing for when the `displayMode` is `"float"`.

SkyeSeitz:
Final specs [available here](https://www.figma.com/file/A7Z8XlYSkv5FSGwIJGoNu3/%5BShell-Panel%2C-Sheet%2C-Table%5D-%E2%80%9Cresizable%E2%80%9D-%5B7591%2C-5652%5D?type=design&node-id=1%3A1435&mode=dev), sharing the behavior of the resize affordance enhancements to Shell Panel.

driskull:
@SkyeSeitz @ashetland do we need an element drag handle in the designs? I'm just wondering if there is a good enough reason to have a resize element which would take up more space? Even in Figma the Figma side panels are resizable but there isn't a resize handle. My concern is taking more space away from content and if it is worth it. I just want to make sure there is a need for this.

For dialog, we do not have resize handles even when a dialog is resizable. It would probably look pretty odd with them on all sides as well.

driskull:
@SkyeSeitz @ashetland is there an issue for tracking the new drag icon shown in Figma? Will that icon be available for the oct release or do we want to either push this or roll without the icon for this release?

SkyeSeitz:
A request has been submitted and the hope is it will be ready for the release 🤞

DitwanP:
🍡 Verified locally on `dev`

PR #10521, opened 2024-10-10
feat(sheet): add resizable property

driskull:
**Related Issue:** #7591

## Summary

- adds resize functionality
- adds assistive text live region for keyboard instructions
- adds tests
- updates story
- adds dom utility
- adds t9n files
- use new resize icon
- uses separator role, related to #3535


benelan (approved):
Code LGTM!

geospatialem (commented):
Testing-wise works great on Windows via keyboard and with JAWS and NVDA. Nice! ✨💪🏻✨

driskull:
Something to note:

This differs from the shell-panel resizing. The shell panel you can tab to focus the resize bar. 

With this, you would just use the arrow keys whenever you are focused on the sheet or anything within the sheet.

I like this approach better so it would be good to refactor the shell panel to be consistent in the future. This also uses `interact.js` for resizing so we could get rid of some code on the shell panel as well.

driskull:
@jcfranco would like your review if possible.

SkyeSeitz:
> Testing-wise works great on Windows via keyboard and with JAWS and NVDA. Nice! ✨💪🏻✨

For my own understanding, wanted to confirm - does not providing a focus ring on that resizing affordance cause any accessibility concerns?

driskull:
> For my own understanding, wanted to confirm - does not providing a focus ring on that resizing affordance cause any accessibility concerns?

@geospatialem can you chime in here? Should we make the resize bar focusable and have a role of separator?

driskull:
@SkyeSeitz @geospatialem here is a separate branch with the resize handle focusable: https://github.com/Esri/calcite-design-system/pull/10577

I've added the border to both PRs.

geospatialem:
> > Testing-wise works great on Windows via keyboard and with JAWS and NVDA. Nice! ✨💪🏻✨
> 
> For my own understanding, wanted to confirm - does not providing a focus ring on that resizing affordance cause any accessibility concerns?

Good catch @SkyeSeitz! 🐟  



> @SkyeSeitz @geospatialem here is a separate branch with the resize handle focusable: #10577
> 
> I've added the border to both PRs.

Update looks great! Ideally it'd be nice to have focus only on the center of the handle, and would have a larger hit area. 

Would that be achievable, or would that need to be a follow-up PR? Asking as I could see the context being lost since it could be the full length of an app, and it being difficult to "tap" on mobile devices per [2.5.8: Target Size (Minimum)](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html) of a 24x24px area. It looks like the value is currently at 10px.

driskull:
> Update looks great! Ideally it'd be nice to have focus only on the center of the handle, and would have a larger hit area.

@geospatialem does this mean we want the element to be focusable?

From what I can see, Figma doesn't specify any hit area. I do see the icon but its 16x16. I can try to make the focusable area extend outside of the draggable element.

geospatialem:
> From what I can see, Figma doesn't specify any hit area. I do see the icon but its 16x16. I can try to make the focusable area extend outside of the draggable element.

That'd work, as long as @ashetland + @SkyeSeitz agree?

driskull:
@geospatialem I'm a bit confused on the "tap" size though.  Keyboard users wouldn't need the hit size anyway since they would be tabbing into it. Mouse users wouldn't get a hit size of 16x16 they would get the 10px width which is already generous for a pointer device. 

Can you elaborate on why this needs a 24x24 hit size?

geospatialem:
> @geospatialem I'm a bit confused on the "tap" size though. Keyboard users wouldn't need the hit size anyway since they would be tabbing into it. Mouse users wouldn't get a hit size of 16x16 they would get the 10px width which is already generous for a pointer device.
> 
> Can you elaborate on why this needs a 24x24 hit size?

The larger hit area could accommodate mobile users, touchscreen PCs, and eye tracking software support.

driskull:
@geospatialem just to be clear, the larger hit area would just be in the center where the icon is right? So if a user wanted to resize above the center the hit area would be 10px.

ashetland:
I agree that the increased hit target doesn't need to run the entire length. We've effectively got a very long skinny hit target that should work well for mouse users. 

This may not be a huge concern in practice, but if we increase the hit target around the icon, is there a danger that this would overlap with some adjacent target inside the Panel's content area or in the area next to the Panel? That would go against the [primary intent](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html#:~:text=The%20intent%20of%20this%20Success%20Criterion%20is%20to%20help%20ensure%20targets%20can%20be%20easily%20activated%20without%20accidentally%20activating%20an%20adjacent%20target.) of the spec. Also, would it be strange for a mouse user to have a 10px target running the length of the Panel, but then have that target be more fuzzy around the icon?

ashetland:
> some adjacent target inside the Panel's content area

For instance, a closable List Item?

driskull:
> This may not be a huge concern in practice, but if we increase the hit target around the icon, is there a danger that this would overlap with some adjacent target inside the Panel's content area or in the area next to the Panel? 

There is the potential for this since we wouldn't know what it could be covering.

If we stay with the 10px, we're not going outside of any reserved space. However, if we want a larger hit target to overlap content then we can run into issues with covering other actionable elements.

driskull:
Even in the figma document, the 24x24 hit target could cover the slider or the block collapse action.

image

ashetland:
What do you think @geospatialem? Expanding the hit area could have some unintended consequences.

geospatialem:
> What do you think @geospatialem? Expanding the hit area could have some unintended consequences.

Would it be possible to add padding to `sheet` around the handle to prevent the encroachment? It would be ideal to have the handle have a hit target of 24x24px so we're fulfilling WCAG 2.5.8, which is level AA.

driskull:
> Would it be possible to add padding to sheet around the handle to prevent the encroachment? It would be ideal to have the handle have a hit target of 24x24px so we're fulfilling WCAG 2.5.8, which is level AA.

> The larger hit area could accommodate mobile users, touchscreen PCs, and eye tracking software support.

Are we sure these hit target requirements apply to resizing? Just want a sanity check on if we do want to support resizing on mobile devices. I see a lot of apps with resizable panels that do not have these hit target sizes. Not saying that they are accessible or correct :)

Codepen's resize bar is 18px for example but not focusable. Outlook online doesn't let you resize via keyboard and has a small hit target. Teams online doesn't let you resize, neither does gmail.

Maybe we should leave as is for this release and spike for further research on accessibility? We could probably use some assistance from LevelAccess.

ashetland:
> Maybe we should leave as is for this release and spike for further research on accessibility? We could probably use some assistance from LevelAccess

I'm good with this.

SkyeSeitz:
**Re: Figma specs** 
The [interaction states section](https://www.figma.com/design/A7Z8XlYSkv5FSGwIJGoNu3/%5BShell-Panel%2C-Sheet%2C-Table%5D-%E2%80%9Cresizable%E2%80%9D-%5B7591%2C-5652%5D?node-id=1159-54157&node-type=canvas&m=dev) shows a 24px hitbox (apologies that detail was subtle). 
image
And a 24px focus ring centered on the icon (I echo what was said above about focus ring being limited to a square, but hit target extending the entire length of the component).
image

**Re: research**
There is documentation in the [Figma file](https://www.figma.com/design/A7Z8XlYSkv5FSGwIJGoNu3/%5BShell-Panel%2C-Sheet%2C-Table%5D-%E2%80%9Cresizable%E2%80%9D-%5B7591%2C-5652%5D?node-id=1213-116222&node-type=section&t=pOtLj9z3UQ5lupId-11) with hit targets ranging from 4px to 24px
image

I am not sure I am against something closer to the 24px hit box and prioritizing the resize function over any action it may slightly overlap with by a few pixels. I have seen this work well in some cases. See an example of Figma's dialog resize taking precedent over content below. 
image

https://github.com/user-attachments/assets/699dc32c-651e-4e05-8460-c6d0697559b2

driskull:
Both the content and container elements within the Sheet are set to `overflow:hidden`. This means placing a hit area outside of the component wouldn't be possible without refactoring these since the hit area would be cut off. 

Removing the `overflow:hidden` could cause regressions elsewhere.

If we want the focus trap to be as intended in Figma, we should push this back to the next release.

cc @geospatialem @SkyeSeitz

geospatialem:
> Both the content and container elements within the Sheet are set to `overflow:hidden`. This means placing a hit area outside of the component wouldn't be possible without refactoring these since the hit area would be cut off.
> 
> Removing the `overflow:hidden` could cause regressions elsewhere.
> 
> If we want the focus trap to be as intended in Figma, we should push this back to the next release.
> 
> cc @geospatialem @SkyeSeitz

@driskull Good catch on the potential regression for folks. @SkyeSeitz and I connected earlier and recommend pushing this back to the next milestone to mitigate the above and ensure the hit target goes beyond the edge of the component's panel to improve usability for users.

driskull:
@SkyeSeitz @ashetland can you review again?

driskull:
@SkyeSeitz @ashetland can you review/accept the screenshots?

SkyeSeitz (approved):
lg2m!

driskull:
@jcfranco merged. Let me know if any changes are needed.

Issue #10536, opened 2024-10-14
[inline-editable] provide way to add tooltip for controls


ptmdmusique:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates and reviewed the [Accessibility page](https://developers.arcgis.com/calcite-design-system/foundations/accessibility) for guidance.

### Summary

There is a request from AGO PE to add tooltips for the controls (✔️, ✏, and ❌) to improve the accessibility of the consumer components of the `inline-editable`

There is a `messageOverrides` prop but it doesn't get forwarded to any native HTML tooltip prop such as `title`

### Actual Behavior

No way to enable tooltips for controls

### Expected Behavior

There should be a prop to enable such tooltips

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/inline-editable/

### Reproduction Steps

1. Hover over the controls, no tooltip appears

### Reproduction Version

latest

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Online

geospatialem:
Verified in `2.14.0-next.15` for the component's "Edit" button, and when `controls` is present, the "Cancel" and "Save" buttons.

PR #10538, opened 2024-10-14
fix(inline-editable): add tooltips for controls

driskull:
**Related Issue:** #10536

## Summary

- add `title` attribute for controls
- change message text to remove "click to"
- add test


benelan (approved):
:+1:

benelan (approved):
:+1:

PR #10547, opened 2024-10-16
chore: add html validator

driskull:
**Related Issue:** #


driskull:
Closing.

Some form of HTML validator to go along with aXe a11y testing would be good for all web component packages. cc @maxpatiiuk

maxpatiiuk:
Is that already covered by your axe accessibility checker?

On AWC, we have a related issue: https://devtopia.esri.com/WebGIS/arcgis-web-components/issues/464

driskull:
aXe does cover most html validation but it doesn't cover everything. Things like duplicate id attribute values and things like that.

PR #10561, opened 2024-10-18
fix(shell-center-row): ensure deprecation warning doesn’t show when using standalone `shell` or `shell-panel`

jcfranco:
**Related Issue:** #10563

## Summary

Moves deprecation warnings to show when `shell-center-row` is added to the DOM and initially loaded.

This warning was being shown due to [bundling](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/stencil.config.ts#L76).


jcfranco:
I'll submit a follow-up PR to update other components accordingly.

benelan (approved):
:+1:

Issue #10044, opened 2024-08-12
[New component] Autocomplete


bsvensson:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

We would like the combobox component to support async data, actions, a menu for switching sources and have the ability to show a suggestion/autocomplete list.

Blocked issues: #8877, #10186

### Acceptance Criteria

- [ ] Create Input search component https://github.com/Esri/calcite-design-system/issues/10186
- [ ] Support Maps SDK for JS' [Search widget](https://developers.arcgis.com/javascript/latest/api-reference/esri-widgets-Search.html) updated to use Calcite Components.
- [ ] Create autocomplete component (_this issue_)
- [ ] Deprecate `filter` component https://github.com/Esri/calcite-design-system/issues/10187

### Relevant Info

Ask @driskull for details ;)

### Which Component

calcite-combobox

### Example Use Case

We are trying to move the [Search widget](https://developers.arcgis.com/javascript/latest/api-reference/esri-widgets-Search.html) to Calcite Components.
* Currently that widget use a filter component, combobox and input and neither handles searching dynamic data or has the ability to show a suggestion/autocomplete list. We would like enhancements to calcite-combobox to support this.

### Priority impact

impact - p1 - need for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

# CalciteAutocomplete

- Internally uses `calcite-input`
- When autocomplete items are populated and the input is clicked or ArrowUp/ArrowDown the component is opened. (open=true)
- Users perform some async search when the `calciteAutocompleteTextInput` event fires, populate the autocomplete list and open it if needed.
- Users listen to keydown (Enter) on the component or listen for a form submission to do a regular search.
- Handle use case: https://developers.arcgis.com/javascript/latest/sample-code/sandbox/?sample=widgets-search-multiplesource
- When a user selects an autocomplete item, a dev can listen for the event and choose to populate or clear the `inputText` value or not.
- Autocomplete results area has a live region with count of results.

## Proposed HTML

```html
<calcite-autocomplete open scale="m" value="1" input-value="Redlands, CA">
[...]

PR #10562, opened 2024-10-18
feat(autocomplete): add autocomplete component

driskull:
**Related Issue:** #10044

## Summary

- Adds `calcite-autocomplete`, `calcite-autocomplete-item`, and `calcite-autocomplete-item-group`
- Adds e2e tests
- Adds storybook
- Adds HTML demos
- Adds t9n strings
- Follows [Figma designs](https://www.figma.com/design/4bHLqKb3qeMq0W2IKSp3eA/Autocomplete?node-id=1-2&m=dev&focus-id=1505-2398)
- Adds CSS tokens
- role: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-autocomplete

### Autocomplete

--calcite-autocomplete-background-color: Specifies the background color of the component.
--calcite-autocomplete-corner-radius: Specifies the component's corner radius.
--calcite-autocomplete-text-color: Specifies the text color of the component.

--calcite-autocomplete-input-prefix-size: Specifies the Input's prefix width, when present.
--calcite-autocomplete-input-suffix-size: Specifies the Input's suffix width, when present.

--calcite-autocomplete-input-background-color: Specifies the background color of the Input.
--calcite-autocomplete-input-border-color: Specifies the border color of the Input.
--calcite-autocomplete-input-corner-radius: Specifies the corner radius of the Input.
--calcite-autocomplete-input-shadow: Specifies the shadow of the Input.

--calcite-autocomplete-input-icon-color:  Specifies the icon color of the Input.
--calcite-autocomplete-input-text-color: Specifies the text color of the Input.
--calcite-autocomplete-input-placeholder-text-color: Specifies the color of placeholder text in the Input.

--calcite-autocomplete-input-actions-background-color: Specifies the background color of Input's `clearable` element.
--calcite-autocomplete-input-actions-background-color-hover: Specifies the background color of Input's `clearable` element when hovered.
--calcite-autocomplete-input-actions-background-color-press: Specifies the background color of Input's `clearable` element when pressed.
--calcite-autocomplete-input-actions-icon-color: Specifies the icon color of Input's `clearable` element.
--calcite-autocomplete-input-actions-icon-color-hover: Specifies the icon color of Input's `clearable` element when hovered.
--calcite-autocomplete-input-actions-icon-color-press: Specifies the icon color of Input's `clearable` element when pressed.

--calcite-autocomplete-input-loading-background-color: Specifies the background color of the Input's `loading` element, when present.
--calcite-autocomplete-input-loading-fill-color: Specifies the fill color of the Input's `loading` element, when present.

--calcite-autocomplete-input-prefix-background-color:  Specifies the background color of the Input's prefix, when present.
--calcite-autocomplete-input-prefix-text-color:  Specifies the text color of the Input's prefix, when present.

--calcite-autocomplete-input-suffix-background-color:  Specifies the background color of the Input's suffix, when present.
--calcite-autocomplete-input-suffix-text-color:  Specifies the text color of the Input's suffix, when present.

### Autocomplete item

--calcite-autocomplete-background-color: Specifies the background color of the component.
--calcite-autocomplete-description-text-color: Specifies the text color of the component's description.
--calcite-autocomplete-heading-text-color: Specifies the text color of the component's heading.
--calcite-autocomplete-text-color: Specifies the text color of the component.

### Autocomplete item group

--calcite-autocomplete-background-color: Specifies the background color of the component.
--calcite-autocomplete-border-color: Specifies the border color of the component.
--calcite-autocomplete-text-color: Specifies the text color of the component.


packages/calcite-components/src/components/autocomplete-item/autocomplete-item.scss
```diff
@@ -0,0 +1,90 @@
...
+  color: var(--calcite-autocomplete-heading-text-color, var(--calcite-color-text-1));
+}
+
+.heading,
+.description {
+  line-height: var(--calcite-font-line-height-relative-snug);
+}
+
+:host(:hover) .container {
+  background-color: var(--calcite-autocomplete-background-color, var(--calcite-color-foreground-2));
```
alisonailea, line 70: Of the three autocomplete components, this is the only one that who's bgcolor changes on hover. Therefor I don't think this follows the established pattern. This should either be `--calcite-autocomplete-background-color-hover` or `--calcite-autocomplete-item-background-color-hover`
@jcfranco

packages/calcite-components/src/components/autocomplete-item/autocomplete-item.scss
```diff
@@ -0,0 +1,90 @@
...
+
+  .description {
+    color: var(--calcite-autocomplete-description-text-color, var(--calcite-color-text-2));
+  }
+}
+
+.container--active {
+  @apply focus-inset;
+  background-color: var(--calcite-autocomplete-background-color, var(--calcite-color-foreground-2));
```
alisonailea, line 79: As above this should either be `--calcite-autocomplete-background-color-press` or `--calcite-autocomplete-item-background-color-press`

alisonailea (changes requested):
Just a few naming schema changes needed on autocomplete-item

packages/calcite-components/src/components/autocomplete-item/autocomplete-item.scss
driskull: Done.

packages/calcite-components/src/components/autocomplete-item/autocomplete-item.scss
driskull: Done.

packages/calcite-components/src/components/autocomplete-item/autocomplete-item.scss
```diff
@@ -76,7 +78,7 @@
 
 .container--active {
   @apply focus-inset;
-  background-color: var(--calcite-autocomplete-background-color, var(--calcite-color-foreground-2));
+  background-color: var(--calcite-autocomplete-background-color-focus, var(--calcite-color-foreground-2));
```
alisonailea, line 81: I just realized `--calcite-autocomplete-background-color-hover` and `--calcite-autocomplete-background-color-focus` have the same default so they can share a token. Simplify this to `--calcite-autocomplete-background-color-hover` per the wiki guidelines

packages/calcite-components/src/demos/autocomplete.html
```diff
@@ -0,0 +1,603 @@
...
+              read-only
+              class="locator-autocomplete"
+              name="location"
+            ></calcite-autocomplete>
+          </form>
+        </div>
+      </div>
+
+      <div class="parent">
+        <div class="child right-aligned-text">Loading</div>
```
macandcheese, line 240: @ashetland @SkyeSeitz not sure if we can do anything by release, but these loading states with Input aren't the best. Previous implementations weren't as common, but in an Autocomplete context, I think we'll see this overlap issue crop up a bunch: 
Screenshot 2024-12-02 at 2 24 07 PM

packages/calcite-components/src/components/autocomplete/autocomplete.tsx
```diff
@@ -0,0 +1,849 @@
...
+declare global {
+  interface DeclareElements {
+    "calcite-autocomplete": Autocomplete;
+  }
+}
+
+/**
+ * @slot - A slot for adding `calcite-autocomplete-item` elements.
+ * @slot actions-end - A slot for adding actions after the input.
+ * @slot actions-start - A slot for adding actions before the input.
```
macandcheese, line 75: @ashetland @SkyeSeitz I wonder if these need to be part of this component or if this is really just layout consideration for the implementer?

packages/calcite-components/src/components/autocomplete/autocomplete.scss
```diff
@@ -0,0 +1,65 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-autocomplete-background-color: Specifies the background color of the component.
+ * @prop --calcite-autocomplete-text-color: Specifies the text color of the component.
+ */
```
macandcheese, line 8: Does this need a `placeholder-text-color` prop?

packages/calcite-components/src/components/autocomplete-item/autocomplete-item.scss
driskull: updated

packages/calcite-components/src/components/autocomplete/autocomplete.scss
driskull: @macandcheese I think all these need to be added right? https://github.com/Esri/calcite-design-system/pull/10820

packages/calcite-components/src/components/autocomplete/autocomplete.tsx
ashetland: I believe the action slots where an mvp requirement, but I'm unsure about the reasons. Agree they could be just a layout consideration and I'm not sure if the slots are helping much given the differences in Button and Action styling.

packages/calcite-components/src/demos/autocomplete.html
ashetland: Yep, agreed.

driskull: Agreed but this isn't related to only this component. I think this is outside of scope for autocomplete.

ashetland:
Can we remove the bg color change onHover for `disabled` items?

packages/calcite-components/src/components/autocomplete/autocomplete.tsx
macandcheese: Yeah, maybe we can look at that in the future, in Input too. I believe the intent there, years ago, was to support some kind of built in submit button developers wired up. But if all the slots are doing are putting content next to the component, maybe that's out of the component's scope and should just be a styling / layout consideration of the implementer.

packages/calcite-components/src/components/autocomplete-item/autocomplete-item.scss
```diff
@@ -0,0 +1,92 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-autocomplete-background-color: Specifies the background color of the component.
+ * @prop --calcite-autocomplete-background-color-hover: Specifies the background color of the component when hovered.
+ * @prop --calcite-autocomplete-background-color-focus: Specifies the background color of the component when focused.
```
alisonailea: remove this comment

packages/calcite-components/src/components/autocomplete-item/autocomplete-item.scss
alisonailea: doesn't look like this got updated

alisonailea (approved):
🚀

ashetland:
> > Can we remove the bg color change onHover for disabled items?
> 
> Done.

Noice. Just noticed that we also have a bg color change while focusing items. For consistency, we should remove that so the focus ring and hover styling are separate.

packages/calcite-components/src/components/autocomplete/autocomplete.tsx
driskull: It makes it a little more difficult on the end user to have a button next to the component. They would have to setup their own flex container to do it. If we had utility classes or components it would be much easier.

I could go either way on this one.

driskull: I can remove them if we think input will have its removed at some point in the future.

packages/calcite-components/src/components/autocomplete-item/autocomplete-item.scss
```diff
@@ -0,0 +1,90 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-autocomplete-background-color: Specifies the background color of the component.
+ * @prop --calcite-autocomplete-background-color-hover: Specifies the background color of the component when hovered.
```
driskull, line 7: @alisonailea do we really need this var? I thought if a user could use the host to set the hover state bg color a var wasn't necessary?

driskull: Like:

```
calcite-autocomplete-item:hover{
--calcite-autocomplete-background-color: red;
}
```

packages/calcite-components/src/components/autocomplete/autocomplete.tsx
macandcheese: I'd think any change there is a ways out - we could just address it then. Agreed on needing some layout helpers for making layout and consistent spacing of components a bit easier.

ashetland (approved):
Lookin' gewd to me

driskull:
@alisonailea added input vars and a corner radius var.

driskull:
I removed the `actions-start` and `actions-end` slots as well.

packages/calcite-components/src/components/autocomplete/autocomplete.tsx
driskull: Removed these slots.

packages/calcite-components/src/components/autocomplete-item/autocomplete-item.scss
driskull: @macandcheese do you know? I thought we only needed bg state specific vars when using the host selector wouldn't work?

macandcheese: I believe what you have above (`calcite-autocomplete-item:hover {}`) is the recommendation, based on this at least: https://github.com/Esri/calcite-design-system/wiki/tokens-component#stateful-naming-rules

If we added separately focusable / clickable areas to an `autocomplete-item`, then we'd need the explicitly defined stateful tokens.

macandcheese (approved):

packages/calcite-components/src/components/autocomplete-item/autocomplete-item.scss
driskull: @alisonailea can you confirm?

alisonailea (approved):
🚀

packages/calcite-components/src/components/autocomplete-item-group/autocomplete-item-group.scss
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @prop --calcite-autocomplete-background-color: Specifies the background color of the component.
+ * @prop --calcite-autocomplete-border-color: Specifies the border color of the component.
+ * @prop --calcite-autocomplete-text-color: Specifies the text color of the component.
+ */
+
+.scale--s {
+  @apply text-n2h;
+  --calcite-internal-autocomplete-item-group-spacing-unit-l: theme("spacing.2");
+  --calcite-internal-autocomplete-item-group-spacing-unit-s: theme("spacing.2");
```
anveshmekala, line 14: Nitpick : can we use one variable here `--calcite-internal-autocomplete-item-group-spacing-unit`. Applies to all scales.

packages/calcite-components/src/components/autocomplete-item-group/autocomplete-item-group.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * When `true`, signifies that the group should not have extra spacing. Used for styling.
+   *
+   * @private
+   */
+  @property() disableSpacing = false;
+
+  /** The component's text. */
+  @property() heading: string;
```
anveshmekala, line 34: Should this be marked `@required`?

driskull: @geospatialem what do you think?

geospatialem: Good eye by @anveshmekala! Agree since the `aria-label` falls back to the `heading` if a `label` isn't specified.

driskull:
@jcfranco @benelan can you review when you get a chance? Thanks!

packages/calcite-components/src/components/autocomplete/autocomplete.tsx
```diff
@@ -0,0 +1,845 @@
...
+
+    if (!el) {
+      return;
+    }
+
+    this.resizeObserver?.observe(el);
+
+    connectFloatingUI(this);
+
+    // todo: fixme when supported in jsx
```
benelan, line 628: Can you make the TODO uppercase so it's aligned with the pattern used during the lumina migration?

benelan (approved):
Code LGTM! Do you need additional manual testing or has that already been taken care of?

packages/calcite-components/src/components/autocomplete-item/resources.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const CSS = {
+  container: "container",
+  containerActive: "container--active",
+  contentCenter: "content-center",
+  description: "description",
+  heading: "heading",
+  iconEnd: "icon-end",
+  iconStart: "icon-start",
+  scale: (scale: Scale) => `scale--${scale}` as const,
```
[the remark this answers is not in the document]
driskull: Its used in `combobox-item`. 

https://github.com/Esri/calcite-design-system/blob/e0f5a85b34addaab150be07c1dc427a49689b6cf/packages/calcite-components/src/components/combobox-item/resources.ts#L13

I copied it from there. Is that ok?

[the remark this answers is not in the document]
benelan: Gotcha, yeah this was just a sidebar; I  was curious if that's something we should be doing instead of using the attribute selector like this:

https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/components/icon/icon.scss#L18-L37

I think either are fine, I just hadn't seen the `scale--s` pattern in our codebase yet. After a quick grep I also found instances of `scale-s`, e.g.

https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/components/tab/tab.scss#L29-L42

Do we have any conventions on whether we should be using BEM in our CSS class names @jcfranco? It seems rather inconsistent between components.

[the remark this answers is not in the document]
driskull: I agree `.scale--s` would be better than `.scale-s` to align with a standardized pattern. We don't really need BEM for specificity since our components are shadowed. Probably an issue to cleanup class names.

I do prefer classes over using attribute selectors because it forces those attributes to be reflected and can make some selectors quite complex.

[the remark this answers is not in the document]
benelan: > I agree `.scale--s` would be better than `.scale-s` to align with a standardized pattern. We don't really need BEM for specificity since our components are shadowed. Probably an issue to cleanup class names.

 I also prefer `.scale--s` since it follows convention. I don't think it's worth an issue because the classes aren't exposed to users. But maybe we can recommend the `block--modifier-value` naming convention for future PRs so the codebase doesn't diverge any further, cc @jcfranco.

> I do prefer classes over using attribute selectors because it forces those attributes to be reflected and can make some selectors quite complex.

Yeah makes sense, thanks!

driskull:
> Code LGTM! Do you need additional manual testing or has that already been taken care of?

@benelan additional manual testing would be great if you have some time.

benelan:
> > Code LGTM! Do you need additional manual testing or has that already been taken care of?
> 
> @benelan additional manual testing would be great if you have some time.

I have to run off to an hour appointment but I can do some later today. I'll focus on form validation stuff.

driskull:
Going to install tonight. If there are follow up issues let me know @benelan @jcfranco

Issue #10522, opened 2024-10-10
Closing on escape no longer works when `focusTrapDisabled=true`


Elijbet:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

**Related issue:** #6456 

Blocked issues: #10682

There is a regression with the focus-trap stacking PR. Closing on escape no longer works when focusTrapDisabled=true. 
 
It may also affect closing on click in some components. 

There is also a related failing test:

@esri/calcite-components:test:   ● stacked focus-trap components › closes a stack of open components sequentially in visual order
@esri/calcite-components:test: 
@esri/calcite-components:test:     expect(received).toBe(expected) // Object.is equality
@esri/calcite-components:test: 
@esri/calcite-components:test:     Expected: "input-time-picker"
@esri/calcite-components:test:     Received: "example-modal"
@esri/calcite-components:test: 
@esri/calcite-components:test:       93 |               ? "sheet-button"
@esri/calcite-components:test:       94 |               : focusTrapOrderElements[i].id || document.body.id;
@esri/calcite-components:test:     > 95 |           expect(activeElementIdAfterTab).toBe(expectedElementId);
@esri/calcite-components:test:          |                                           ^
@esri/calcite-components:test:       96 |
@esri/calcite-components:test:       97 |           await page.keyboard.press("Escape");
@esri/calcite-components:test:       98 |           await page.waitForChanges();
@esri/calcite-components:test: 
@esri/calcite-components:test:       at testEscapeAndAssertOpenState (src/tests/stackedFocusTrap.e2e.ts:95:43)
@esri/calcite-components:test:       at testStackEscapeSequence (src/tests/stackedFocusTrap.e2e.ts:125:7)
@esri/calcite-components:test:       at Object.<anonymous> (src/tests/stackedFocusTrap.e2e.ts:128:5)

### Expected Behavior

Components should close on escape when focus-trap is true individually or in a stack. 

### Reproduction Sample

https://codepen.io/jcfranco/pen/PoMbdMe?editors=100

### Reproduction Steps

1. Open the modal
2. Press escape
3. Observe the modal not closing as expected

### Reproduction Version

2.13.0

### Regression?

2.12.2

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
[...]

Issue #10364, opened 2024-09-20
[Sheet] Add support for non-modal Sheet, other improvements


macandcheese:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Placeholder issue for a set of enhancements to improve the Sheet component. This will support more use-cases and improve consistency with the Dialog component.

Rendering a Panel inside Sheet will provide an enhanced developer experience as well as ensure visual similarity in implementations.

### Acceptance Criteria

- Sheet renders Panel to match Dialog, passing through appropriate slots + properties.
- Sheet adds nonmodal display by default (breaking change) to match Dialog
- Keeps the "overlay / float" display types
- Resizability is added to Sheet - using an updated affordance with generous hit area to support the thumb gestures / mobile behavior that is common + expected with these displays (design to discuss: swipe tolerances for min / max height (or width), swipe tolerance for `close`, etc). Note that this doesn't apply just to mobile viewport sizes - these are needed behaviors in larger environments as well.

### Relevant Info

cc @SkyeSeitz @ashetland for ui / ux discussion + final properties / slots needed from a design perspective.

### Which Component

Sheet

### Example Use Case

As a designer-user - I want to use a nonmodal Sheet, and have the visual display match Dialog and other components

As a developer-user - I want to build the component in a similar manner to Dialog, and simplify the need to manage properties with my own slotted Panel inside Sheet

As an end-user - I want to use common mobile "niceties", like large hit areas for thumb dragging, and swipe interactions. 

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

ashetland:
This is covered in issues #7591, #9828

ellenupp:
Hi @macandcheese, @ashetland, the title part of this issue (non-modal sheet) doesn't seem to be addressed by the issues linked above. Is this still something you're considering?

macandcheese:
Yes, I think we can re-purpose this issue or open a new one to ensure that is addressed and added as a feature. cc @geospatialem

geospatialem:
[...]

PR #10578, opened 2024-10-22
fix(input-date-picker, input-time-picker, modal, sheet): closes on escape when focusTrap is disabled

Elijbet:
**Related Issue:** #10522

## Summary
Close on `Escape` works when `focusTrap` is disabled.


packages/calcite-components/src/components/sheet/sheet.tsx
```diff
@@ -224,6 +225,26 @@ export class Sheet implements OpenCloseComponent, FocusTrapComponent, LoadableCo
     this.handleMutationObserver(),
   );
 
+  //--------------------------------------------------------------------------
+  //
+  //  Event Listeners
+  //
+  //--------------------------------------------------------------------------
+
+  @Listen("keydown", { target: "window" })
```
driskull, line 234: Instead of creating this event on the window, can we add it on the host of the component? That way, it won't close for an escape that happens outside of the component.

driskull: Same thing with the modal component. Although less important since the modal is deprecated :D

Elijbet: Oh, I initially had it on host but @jcfranco thinks we should close the modal regardless of where esc occurs.

driskull: Ok. However, when we refactor sheet to support #10364, this would need to move to the host when non modal. It might be better to just have it on sheets host for now. I'll defer to @jcfranco

driskull (approved):
👍

packages/calcite-components/src/components/sheet/sheet.tsx
driskull: When a sheet can be non modal, we wouldn't want escape closing it from outside of the sheet.

Elijbet:
As per group discussion. @kitty @franco @skye @aaron 
 
- `dialog` with `modal`: no support for `focusTrapDisabled`

    - Technically we don't offer this in dialog currently so this is  and once modal is removed we're good to go

- `dialog`: offer `focusTrapDisabled` depending on workflows

    - New attr/prop, not enabled by default
    - Example of where this is useful, a cookie page dialog that mentions cookie tracking, but the user can still interact with the page without acknowledgement/acceptance

- `input-date-picker` / `input-time-picker`: offer `focusTrapDisabled` depending on workflows

    - Not enabled by default
    - We already offer this, so no action needed if we agree on the approach
    - It could make sense depending on what the user is implementing in their UI/Ux

driskull:
> input-date-picker / input-time-picker: offer focusTrapDisabled depending on workflows

I think for the future we should plan to deprecate `focusTrapDisabled` at some point  and have these components no longer support any focus trapping. They should act like a single component and be controlled via arrow keys only so tabbing past them is easier for an end user.

packages/calcite-components/src/components/modal/modal.tsx
```diff
@@ -329,6 +334,19 @@ export class Modal
 
   // #region Private Methods
 
+  private keyDownHandler = (event: KeyboardEvent): void => {
+    const { defaultPrevented, key } = event;
+
+    if (defaultPrevented) {
```
jcfranco, line 340: This should also return when focus trap is enabled. Applies to similar event handlers.

packages/calcite-components/src/components/input-date-picker/input-date-picker.e2e.ts
```diff
@@ -351,6 +351,33 @@ describe("calcite-input-date-picker", () => {
       });
     });
 
+    it("closes when Escape key is pressed and focusTrapDisabled=true", async () => {
+      page = await newE2EPage();
```
jcfranco, line 355: I think we could create a common test helper for focus trapping components. Can you create an issue for this?

jcfranco, line 355: We’ll also need an issue to clean up this and similar tests. Avoiding `beforeEach` setups that get overwritten would help as they lead to discarded test pages.

jcfranco (approved):
Awesome, @Elijbet! Once comments are addressed, this is good to go! ✨🚀✨

packages/calcite-components/src/components/input-date-picker/input-date-picker.e2e.ts
Elijbet: [Create a common test helper for focus trapping components](https://github.com/Esri/calcite-design-system/issues/10891#top)
#10891

Elijbet: [Refactor tests to avoid beforeEach setups that get overwritten](https://github.com/Esri/calcite-design-system/issues/10892#top) #10892

packages/calcite-components/src/components/sheet/sheet.tsx
```diff
@@ -245,6 +250,19 @@ export class Sheet
...
   // #region Private Methods
 
+  private keyDownHandler = (event: KeyboardEvent): void => {
+    const { defaultPrevented, key } = event;
+
+    if (defaultPrevented || this.escapeDisabled || !this.focusTrapDisabled) {
+      return;
+    }
+
+    if (this.open && key === "Escape") {
```
jcfranco, line 260: Nitpick: you can move this condition in the `if` statement above to have a single `if` block. Applies to similar event handlers.

jcfranco (approved):

Elijbet:
Note: With this change it will only be closing on Escape when focus trap is disabled if the event is emitted within the component (i.e., Escape won't close if focus is in another, separate, element).

PR #10581, opened 2024-10-22
fix: updated web-adapter name

arowles:
**Related Issue:** #

## Summary
**updated**

web-adaptor > web-adapter


aphilobos (approved):

jcfranco:
@arowles @aphilobos `web-adaptor` is [available in version 3.32.0](https://developers.arcgis.com/calcite-design-system/icons/?library=Calcite%20UI&query=web-adaptor), so this rename would be a breaking change. Developers currently using this icon ([example](https://codepen.io/jcfranco/pen/abeVoRR?editors=100)) would need to update their code, or their icons will stop displaying.

Consider reverting or restoring the icon + deprecating it instead.

cc @benelan

arowles:
@jcfranco After reverting the icon, how would I deprecate it?

jcfranco:
Good question, not sure if there's a mechanism for this yet in the icon doc page. Would it possible to use a keyword or some other type of metadata to mark it as deprecated and have it show up in the doc? @benelan Any other ideas?

For clarification, are you thinking about:

1. reverting `web-adapter` back to `web-adaptor`, mark it as deprecated and do the rename and removal later or
2. restoring `web-adaptor` (alongside `web-adapter`) and mark it as deprecated and remove later

?

PR #10589, opened 2024-10-23
ci(chromatic): build storybook even when skipping snapshots

benelan:
## Summary

There have been a few occasions where storybook build errors made it onto `dev` due to skipping snapshots. This will ensure storybook always builds and catches errors, even if it isn't uploaded to chromatic.


jcfranco (approved):
Nice fix! ✨🔨✨

PR #10594, opened 2024-10-23
feat: add buffer point, buffer polygon, buffer polyline, contour, offset

arowles:
**Related Issue:** #

## Summary
**added**

buffer-point-16
buffer-point-24
buffer-point-32
buffer-polygon-16
buffer-polygon-24
buffer-polygon-32
buffer-polyline-16
buffer-polyline-24
buffer-polyline-32
contour-16
contour-24
contour-32
offset-16
offset-24
offset-32

image


allieorth (approved):
Files look good! 
Thanks for the upload Ashley!!

PR #10598, opened 2024-10-23
deprecate: deprecate web-adaptor in favor of web-adapter

arowles:
**Related Issue:** #

## Summary
**added**
web-adaptor


benelan (commented):
Thanks for adding back the deprecated icons to prevent a breaking change. Can you please change the PR title to `deprecate: deprecate web-adaptor in favor of web-adapter`. That will make sure it ends up on the changelog so people know to migrate to the fixed icon names.

jcfranco:
Does this also need an update to `keywords.json`?

benelan:
> Does this also need an update to `keywords.json`?

I think we want to remove deprecated icons from `keywords.json` to prevent more people from finding them in the doc and using them. The alternative is adding a `deprecated` field and displaying that in the doc. cc @macandcheese

jcfranco:
From a dev perspective, it’d be helpful to surface deprecation notices on the docs site. It could also help identify all icons slated for removal later. On that note, should we create an issue to track the removal of this icon for a future breaking change release?

jcfranco:
@arowles Thanks for restoring this so quickly, btw! 🎉

benelan:
Yeah, maybe one issue for all of the future icon removals? We have some from #10309 too.
cc @geospatialem @DitwanP

@macandcheese can you add some logic to the icon finder for showing a deprecation chip/message? 

Should it just be a boolean `deprecated` field, or should we add more metadata like the version of deprecation or future removal?

I think we should wait until the doc site handles deprecations before adding them back to `keywords.json`. Otherwise they will appear as normal icons and people may use them instead of the fixed ones.

macandcheese:
We could have a “deprecated in favor of” metadata field in addition to just deprecation as Boolean. We don’t need to show this to a user but we can use it to surface the new icon without showing the old name?

I also think from a use case perspective this would solve the “why isn’t my icon working”, as I’d expect only searches for exact icon term to return the deprecated ones. 

We can work on the UI / UX of how we display this, I think the above metadata would be sufficient.

benelan (approved):
Thanks!

PR #10608, opened 2024-10-25
feat(navigation-user): add component tokens

anveshmekala:
**Related Issue:** #7180 

## Summary
Add the following component-scoped CSS Variables to calcite-navigation-user:

`--calcite-navigation-accent-color`: Specifies the components's border color when `active`.
`--calcite-navigation-user-avatar-corner-radius`: Specifies the corner radius of avatar.
`--calcite-navigation-user-avatar-color`: Specifies the icon color of avatar.
`--calcite-navigation-background-color`: Specifies the component's background color.
`--calcite-navigation-user-full-name-text-color`: Specifies the component's full name text color.
`--calcite-navigation-user-name-text-color`: Specifies the component's username text color.


packages/calcite-components/src/components/navigation-user/navigation-user.scss
```diff
@@ -43,23 +43,40 @@
 }
 
 calcite-avatar {
-  @apply px-4;
+  --calcite-avatar-corner-radius: var(--calcite-navigation-user-avatar-corner-radius);
+  --calcite-avatar-color: var(--calcite-navigation-user-avatar-color);
+  padding-inline: 1rem;
```
alisonailea, line 48: leave `@apply px-4;`

packages/calcite-components/src/components/navigation-user/navigation-user.scss
```diff
@@ -43,23 +43,40 @@
 }
 
 calcite-avatar {
-  @apply px-4;
+  --calcite-avatar-corner-radius: var(--calcite-navigation-user-avatar-corner-radius);
+  --calcite-avatar-color: var(--calcite-navigation-user-avatar-color);
```
alisonailea, line 47: update these to

```css
border-radius: var(--calcite-navigation-user-avatar-corner-radius, var(--calcite-avatar-corner-radius));
color: var(--calcite-navigation-user-avatar-color, var(--calcite-avatar-color));
```

packages/calcite-components/src/components/navigation-user/navigation-user.scss
```diff
@@ -1,39 +1,39 @@
...
       focus-base
+      font-sans
       text-0h;
-    border-block-end: 2px solid transparent;
+    border: none;
+    background-color: var(
+      --calcite-navigation-background-color,
+      var(--calcite-internal-navigation-user-background-color, transparent)
+    );
+    border-block-end: 2px solid
```
alisonailea, line 32: We've identified a design pattern that will likely be moved to semantic tokens in the future. The selection-color pattern goes
default = transparent
hover = border-1 (not always present)
press = brand

`border-block-end: 2px solid
      var(--calcite-navigation-selection-color, var(--calcite-internal-navigation-selection-color, transparent));`

packages/calcite-components/src/components/navigation-user/navigation-user.scss
```diff
@@ -43,23 +43,40 @@
...
+:host(:focus) .button {
+  --calcite-internal-navigation-user-background-color: var(--calcite-color-foreground-2);
+}
+
+:host(:focus) .button {
+  @apply focus-inset;
+}
+
+:host(:active) .button {
+  --calcite-internal-navigation-user-background-color: var(--calcite-color-foreground-3);
```
alisonailea, line 75: Not entirely sure here but I think this can be simplified.

Currently the pattern looks like
navigation.background.color = color.foreground.1
:hover,:focus = color.foreground.2
:active = color.foreground.3

but the background color was already set at the parent component "navigation". The hover press states here aren't actually trying to change the color but imply a shadow like the button is indenting. Therefor we can update this to use the color.transparent pattern for hover,active state.

remove line 28
remove lines 65 - 80

replace with

```css
:host .button {
  background-color: var(--calcite-color-transparent);

  &:hover,
  &:focus {
    background-color: var(--calcite-color-transparent-hover);
  }

  &:focus {
    @apply focus-inset;
  }

  &:active {
    background-color: var(--calcite-color-transparent-press);
  }
}

:host([active]) .button {
  --calcite-internal-navigation-selection-color: var(--calcite-color-brand);
}
```

The background is always whatever the navigation.background.color is.

alisonailea (changes requested):
Looking good. I've made some suggestions to align with cross-component patterns. I don't think we should allow users to have a custom background set on this component as I think the pattern is that it darkens the default bgcolor on interaction. @SkyeSeitz can you confirm?

SkyeSeitz:
> Looking good. I've made some suggestions to align with cross-component patterns. I don't think we should allow users to have a custom background set on this component as I think the pattern is that it darkens the default bgcolor on interaction. @SkyeSeitz can you confirm?

Thinking Nav Logo bg should either be able to change color or should remain transparent on idle to display the Navigation background fill whether it be the out of the box foreground.1 or customized. The hover and press backgrounds should be able to be customized imo. If the Navigation bar is changed to an accent color fill for example, the default foreground.2 on hover, foreground.3 on press may not translate well. 

image

alisonailea:
Agreed. In that case @anveshmekala  keep your original pattern using the "internal" tokens but update the :hover and :press to use "transparent-hover" "transparent-press instead.

packages/calcite-components/src/components/navigation-user/navigation-user.scss
anveshmekala: Is it suggestible to follow the `-selection-` naming here for `border-block-end` or wait for semantic tokens. 

This applies for `navigation-logo` as well.

anveshmekala:
> Thinking Nav Logo bg should either be able to change color or should remain transparent on idle to display the Navigation background fill whether it be the out of the box foreground.1 or customized. The hover and press backgrounds should be able to be customized imo. If the Navigation bar is changed to an accent color fill for example, the default foreground.2 on hover, foreground.3 on press may not translate well.
> 

Assuming this applies for `navigation-user` too.

packages/calcite-components/src/components/navigation-user/navigation-user.scss
alisonailea: can you use --calcite-navigation-selection-color for both?

anveshmekala: `--calcite-navigation-selection-border-color` might be more meaningful here.

anveshmekala: >We've identified a design pattern that will likely be moved to semantic tokens in the future. The selection-color pattern goes
default = transparent
hover = border-1 (not always present)
press = brand

Had some thinking over this with Skye and don't think `selection` pattern applies here completely because we shouldn't allow user to customize `border-color` in default case when `active=false`. User's are expected to customize `border-color` only when `active=true` and no other interaction or properties should be able to change it. 

```
//default case
anchor {
border-block-end : 2px solid transparent
}
```

Perhaps using `--calcite-navigation-selection-border-color` or `--calcite-navigation-active-border-color` might be more semantic. I am leaning towards `--calcite-navigation-active-border-color` because selection on `navigation-logo` and `navigation-user` doesn't add the border and users have to explicitly set the `active` prop to `true ` to get the border added.  To differentiate the behavior prefer using `-active-` naming in token.

cc @SkyeSeitz

alisonailea: in other components, the solution is to only apply `--calcite-navigation-selection-border-color` on [active] and leave it transparent for other style props

alisonailea (changes requested):
one token name change request and I think we're good to go

packages/calcite-components/src/components/navigation-user/navigation-user.scss
```diff
@@ -44,22 +43,39 @@
...
+}
+
+:host([active]) .button {
+  border-block-end-color: var(--calcite-navigation-selection-border-color, var(--calcite-color-brand));
```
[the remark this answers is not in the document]
jcfranco: Agreed.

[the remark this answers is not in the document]
anveshmekala: should this be `--calcite-navigation-accent-color` similar to `card` & `tile` ?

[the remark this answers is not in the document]
alisonailea: Per https://github.com/Esri/calcite-design-system/wiki/tokens-component#general-naming-rules I think it should be `--calcite-navigation-accent-color` because the state change is on the :host

alisonailea (approved):
Thanks!

PR #10613, opened 2024-10-25
refactor: simplify `newProgrammaticE2EPage`

jcfranco:
**Related Issue:** N/A

## Summary

Per https://github.com/Esri/calcite-design-system/pull/10509#discussion_r1798082386:

> **TL;DR**: it seems this workaround is no longer needed and we can simplify `newProgrammaticE2EPage`.
> 
> It looks like Stencil fixed an issue where creating an empty page and then adding a component via `createElement` would not load properly (i.e., the `html` content required a component).


aPreciado88 (approved):
🚀

Issue #10326, opened 2024-09-16
[Tests bug] Incorrect calls to `skipAnimations(page);`


maxpatiiuk:
### Priority impact

p - low

### Test type

End to end tests

### Which Component(s)

All

### Unstable Tests

In many places in Calcite E2E tests, `await skipAnimations(page);` is called before `page.setContent()`:

https://github.com/Esri/calcite-design-system/blob/e9d18625382828f8faaab6cc38f0ed7bdb3d72e2/packages/calcite-components/src/components/alert/alert.e2e.ts#L122

`skipAnimations()` is adding a `<style>` tag to the page.

The issue is that `page.setContent()` clears all other content from the page, thus the `<style>` tag added by `skipAnimations()` is removed.

Please review all the places where `skipAnimations()` is called in such way to check if `skipAnimations()` is necessary at all (since tests are supposevely working without it?). If it is necessary, then move the call to be after `page.setContent()`.

### Additional Info

For reference, see the implementation of `page.setContent()`:

https://github.com/ionic-team/stencil/blob/814642f9035df24b1ed45019991542a3fba0884b/src/testing/puppeteer/puppeteer-page.ts#L208

DitwanP:
🍠 Verified

PR #10615, opened 2024-10-25
test: ensure `skipAnimation` calls take place after content is set

jcfranco:
**Related Issue:** #10326 

## Summary

This fixes `skipAnimation` calls that were being cleared after setting content.


aPreciado88 (approved):
🚀

Issue #9999, opened 2024-08-08
[Rating and Slider] Add form support via `required` and `validationMessage` for error handling


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The `rating` and `slider` components do not support form validation, as indicated in #8057.

### Expected Behavior

Form validation support. Since invalid styling is not effective on `slider` and `rating`, the `validationMessage` prop would serve as the only visual invalid affordance when a value is not added and `required` is present.

### Reproduction Sample

https://codepen.io/geospatialem/pen/ZEdJaLr

### Reproduction Steps

1. Open the [sample](https://codepen.io/geospatialem/pen/ZEdJaLr)
2. Observe neither component have support for `required` or `validationMessage` properties/attributes

### Reproduction Version

`2.11.1`

### Relevant Info

Follow-up to https://github.com/Esri/calcite-design-system/issues/8057 and https://github.com/Esri/calcite-design-system/issues/8000

Link to Figma specs: https://www.figma.com/design/5QDPzXr21qQcPTkXR1Ouga/Form-Validation-%5BEpic-4598%5D?node-id=956-94284&m=dev&focus-id=1194-2726

### Priority impact

impact - p3 - not time sensitive

### Impact

Medium-High for consistency across the design system with regards to error handling

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

DitwanP:
🍡 Verified on `2.14.0-next.19`

https://codepen.io/Ditwan-Price/pen/BaXrBoO

PR #10621, opened 2024-10-25
feat(rating, slider): add validation message to support form error handling

aPreciado88:
**Related Issue:** [#9999](https://github.com/Esri/calcite-design-system/issues/9999)

## Summary

Add `validationMessage`, `validationIcon` and `status` props to `rating` and `slider` components to support form validation.


aPreciado88:
@ashetland @SkyeSeitz Can you please take a look at the chromatic build?

packages/calcite-components/.stylelintrc.cjs
```diff
@@ -1,7 +1,13 @@
 // @ts-check
 
 // ⚠️ AUTO-GENERATED CODE - DO NOT EDIT
-const customFunctions = [];
+const customFunctions = [
+  "get-trailing-text-input-padding",
```
benelan, line 5: This code is auto-generated by [this script](https://github.com/Esri/calcite-design-system/blob/dev/support/updateStylelintCustomSassFunctions.ts) and shouldn't be edited directly. IIRC @jcfranco set it up if you need more info.

benelan (commented):
Looking good! It looks like some code marked as \"do not edit\" was edited, can you take another look at that part?

benelan (approved):
:+1:

benelan (approved):
:+1:

aPreciado88:
@ashetland I pulled `dev` and ran a new chromatic build. Could you please take another look?

ashetland:
> @ashetland I pulled `dev` and ran a new chromatic build. Could you please take another look?

That fixed it. Just the one remaining conversation to resolve.

aPreciado88:
@ashetland I updated the error message's top padding within the slider component. Can you please take a look?

packages/calcite-components/src/components/slider/slider.scss
```diff
@@ -36,6 +36,11 @@
 
 :host {
   @apply block;
+
+  // remove top padding applied by validation component.
+  .validation-container {
+    padding-block-start: 0 !important;
+  }
```
benelan, line 43: Are you sure the bottom padding from the slider is the same as the top padding from the validation component? And did you check when there are ticks vs when there aren't ticks?

It might be better to figure out where the padding from the slider is coming from and set that to zero when the validation message is displayed.

benelan (changes requested):
Requesting changes until the padding discrepancies are resolved so this doesn't get accidentally merged (I've been there)

packages/calcite-components/src/components/slider/slider.scss
aPreciado88: I didn't check the slider's bottom padding, I made this update based on the comments in this [snapshot](https://www.chromatic.com/test?appId=6266d45509d7eb004aa254fb&id=671bebe3db7fcedbad1be417&activeElementId=comment-thread-671c1a64e077f880d18c3e47). I can update this to keep the validation's top padding and remove the slider's bottom padding if that's the better option.

benelan: I was looking through the existing form validation components and found this approach too:

https://github.com/Esri/calcite-design-system/blob/2eb7409b8a4fd42f6c119adaf529146099653e2e/packages/calcite-components/src/components/radio-button-group/radio-button-group.scss#L31-L41

@ashetland maybe we could get some design feedback on how we should be ensuring consistent padding between the components and their validation messages?

- Should the spacing be the same for all components, e.g. zeroing out the component's bottom spacing when the validation message is displayed?
- Or should keep the spacing of the component and change the validation message's spacing in some way?

ashetland:
Visuals lookin' good!

benelan (approved):
LGTM!

packages/calcite-components/src/components/slider/slider.scss
ashetland: We decided to remove the padding from the validation message since it is thing being added, it seemed logical to modify _it_ rather than Slider. Due to Slider's margin-block of 7px, its default final height is a couple pixels shorter that our other inputs. This will make the message 1px closer in this case when compared to the other inputs.

Issue #7207, opened 2023-06-21
[Loader / Progress] Make value property consistent


macandcheese:
### Description

Currently, Progress accepts a value between 0 -> 1.0, whereas Loader accepts a value between 0 -> 100. It would be nice to gain alignment between these components.

### Acceptance Criteria

I have no preference which is used, but have both of these components use a similar expected value.

### Which Component

Progress / Loader

### Example Use Case

As a user, having two different types of value for similar components can be unexpected.

### Priority impact

p4 - not time sensitive

### Esri team

Calcite (design)

Elijbet:
Did a bit of research of similar libraries. Spectrum, Shoelace, Carbon, ui5 all use 100% max value for progress/loader. Some don't have a loader, or call loaders progress circle/ring instead. Vaadin is the only one so far that uses a range 0-1.

benelan:
I vote for changing loader instead of progress, because there is no native loader and I'd prefer to keep progress in line with what native does:

benelan:
Also could we call this a breaking change instead of a fix? If people are setting values in their app, changing the default progress/loader range would change how much of the task appears completed.

Elijbet:
@benelan, @jcfranco 
We looked at our components vs native components and saw this distribution:

From our API consistency, we're looking at:

**0 – 100**
* loader
* meter

**0 – 1**
* progress
 
Taking your point, if we compare those to native equivalents, we've got:

**0 – 1**
* progress
* meter

We decided to deviate from native. It seems to be more in line with existing component libs and meter doesn't follow this either.
We also felt that from a dev perspective, you'd likely expect 0-100.

geospatialem:
Verified on the `dev` branch, where `progress` has a `type="determinate"` and a specified `value` between `0` and `100`):

```html
<calcite-progress type="determinate" value="92"></calcite-progress>
```

As mentioned in Eliza's comment [above](https://github.com/Esri/calcite-design-system/issues/7207#issuecomment-2460985536) `loader` and `meter` already contain this pattern (confirmed on the `dev` branch) with the following for `loader` (_Note_: `100` is considered `complete` on the component):

```html
<calcite-loader label="Loading the holidays" text="Loading..." type="determinate" value="92"></calcite-loader>
```

PR #10622, opened 2024-10-25
fix(progress)!: convert value range to 0-100

Elijbet:
**Related Issue:** #7207

## Summary
 
Fixes the discrepancy in value range used between Progress (0 -> 1.0) and Loader (0 -> 100), to have components use a similar expected value.

BREAKING CHANGE: Refactors `progress` to use the value range of 0-100. Developers will need to provide a function to map existing values or use the updated range.


aPreciado88 (approved):
🚀

benelan (changes requested):
The value being between 0 and 1 comes from the native progress:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress#value

Plus this would be a breaking change in my opinion. If we want consistency, maybe we should change loader instead, which doesn't have a native equivalent.

Elijbet:
> The value being between 0 and 1 comes from the native progress: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress#value
> 
> Plus this would be a breaking change in my opinion. If we want consistency, maybe we should change loader instead, which doesn't have a native equivalent.

We decided not to go with 0-1 because of this little study: https://github.com/Esri/calcite-design-system/issues/7207#issuecomment-2460985536

jcfranco (changes requested):
Please roll back changes unrelated to the progress component.

packages/calcite-components/src/components/progress/progress.tsx
```diff
@@ -34,7 +34,7 @@ export class Progress extends LitElement {
    */
   @property({ reflect: true }) type: "indeterminate" | "determinate" = "determinate";
 
-  /** When `type` is `"determinate"`, the component's progress value with a range of 0.0 - 1.0. */
+  /** The component's value. Valid only for `"determinate"` indicators. Percent complete of 100. */
```
jcfranco, line 37: +@geospatialem @DitwanP for editorial perusal.

packages/calcite-components/src/components/progress/progress.tsx
```diff
@@ -43,14 +43,14 @@ export class Progress extends LitElement {
...
-    const barStyles = isDeterminate ? { width: `${this.value * 100}%` } : {};
+    const barStyles = isDeterminate ? { width: `${this.value}%` } : {};
     const dir = getElementDir(this.el);
     return (
       <div
         ariaLabel={this.label || this.text}
-        ariaValueMax={1}
-        ariaValueMin={0}
-        ariaValueNow={this.value}
+        ariaValueMax={isDeterminate ? "100" : undefined}
```
jcfranco, line 51: Noice!

jcfranco (approved):
✨🚀✨

packages/calcite-components/src/components/progress/progress.tsx
geospatialem: WDYT about:

```suggestion
  /** When `type` is `"determinate"`, specifies the component's value with a range of 0 to 100. */
```

benelan (approved):
:+1:

PR #10642, opened 2024-10-29
deprecate: deprecate `ban-props-on-host` rule

jcfranco:
**Related Issue:** #10398

## Summary

Deprecates this rule for 3.x removal as it is no longer needed after #10310.


aPreciado88 (approved):
🚀

PR #10647, opened 2024-10-29
feat(switch): add component tokens

driskull:
**Related Issue:** #7180

## Summary

adds new component tokens, e2e tests, demo page examples, and chromatic theme tests.

### New CSS variables

--calcite-switch-background-color: Specifies the component's background color.
--calcite-switch-background-color-hover: Specifies the component's background color.
--calcite-switch-border-color: Specifies the component's border color.
--calcite-switch-corner-radius: Specifies the component's corner radius.
--calcite-switch-handle-background-color: Specifies the handle's background color.
--calcite-switch-handle-background-color-hover: Specifies the handle's background color.
--calcite-switch-handle-border-color: Specifies the handle's border color.
--calcite-switch-handle-corner-radius: Specifies the handle's corner radius.
--calcite-switch-handle-shadow: Specifies the handle's shadow.


packages/calcite-components/src/components/switch/switch.scss
```diff
@@ -1,3 +1,22 @@
...
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-switch-background-color: Specifies the component's background color.
+ * @prop --calcite-switch-background-color-hover: Specifies the component's background color.
+ * @prop --calcite-switch-border-color: Specifies the component's border color.
+ * @prop --calcite-switch-corner-radius: Specifies the component's corner radius.
+ * @prop --calcite-switch-handle-background-color: Specifies the handle's background color.
+ * @prop --calcite-switch-handle-background-color-hover: Specifies the handle's background color.
```
alisonailea, line 11: Because the :hover/:active is based on interaction with the parent .container not individual elements within the component this can be simplified to not include `-hover` tokens. You'll need to update the CSS so the pseudoclasses are on `:host`

packages/calcite-components/src/components/switch/switch.scss
```diff
@@ -52,60 +71,69 @@
...
   inset-block-start: -1px;
   inset-inline: -1px theme("inset.auto");
+  background-color: var(--calcite-switch-handle-background-color, var(--calcite-color-foreground-1));
+  border-color: var(--calcite-switch-handle-border-color, var(--calcite-color-border-input));
+  border-radius: var(--calcite-switch-handle-corner-radius, 9999px);
+  box-shadow: var(--calcite-switch-handle-shadow);
 }
 
 .container:hover,
 .container:focus {
```
alisonailea, line 111: update this to `:host(:hover, :focus) {`

packages/calcite-components/src/components/switch/switch.scss
```diff
@@ -52,60 +71,69 @@
...
+    border-color: var(--calcite-switch-border-color, var(--calcite-color-brand-hover));
+    background-color: var(--calcite-switch-background-color-hover, var(--calcite-color-brand));
   }
   .handle {
-    @apply border-color-brand;
+    border-color: var(--calcite-switch-handle-border-color, var(--calcite-color-brand));
     inset-inline: theme("inset.auto") -1px;
   }
 
   .container:hover {
```
alisonailea, line 133: update this to `&:host(:hover)`

alisonailea (commented):
Looking good! We don't need stateful tokens on this component but other than that this looks good to go.

packages/calcite-components/src/components/switch/switch.scss
```diff
@@ -1,3 +1,22 @@
...
+ * @prop --calcite-switch-corner-radius: Specifies the component's corner radius.
+ * @prop --calcite-switch-handle-background-color: Specifies the handle's background color.
+ * @prop --calcite-switch-handle-background-color-hover: Specifies the handle's background color.
+ * @prop --calcite-switch-handle-border-color: Specifies the handle's border color.
+ * @prop --calcite-switch-handle-corner-radius: Specifies the handle's corner radius.
+ * @prop --calcite-switch-handle-shadow: Specifies the handle's shadow.
+ */
+
+// todo: question: do we need size variables for the scales?
+// todo: question: do we want to call the vars `handle`? I've also seen `thumb` used here.
```
driskull, line 18: @alisonailea what about these questions?

packages/calcite-components/src/components/switch/switch.scss
```diff
@@ -52,60 +71,69 @@
...
     relative
     box-border
     inline-block
-    rounded-full
     border
     border-solid
     align-top
     focus-base;
+
+  // todo: need var for 9999px?
```
driskull, line 83: @alisonailea what about this?

driskull: Hard coding it for now.

packages/calcite-components/src/components/switch/switch.scss
driskull: Will leave name as is for now, no scale/size vars.

packages/calcite-components/src/components/switch/switch.scss
driskull: Something to watch out for here, is that we also need to check for disabled now. `&:host(:hover:not([disabled])) {`

Because previously the container was inside the component's InteractionContainer with pointer-events none. Moving this selector to the host means we need to check to make sure it doesn't apply when disabled.

packages/calcite-components/src/components/block-section/block-section.scss
```diff
@@ -4,6 +4,16 @@
   color: var(--calcite-color-text-2);
 }
 
+calcite-switch {
+  // --calcite-switch-background-color: var(--calcite-block-section-switch-background-color);
+  // --calcite-switch-border-color: var(--calcite-block-section-switch-border-color);
+  // --calcite-switch-corner-radius: var(--calcite-block-section-switch-corner-radius);
+  // --calcite-switch-handle-background-color: var(--calcite-block-section-switch-handle-background-color);
+  // --calcite-switch-handle-border-color: var(--calcite-block-section-switch-handle-border-color);
+  // --calcite-switch-handle-corner-radius: var(--calcite-block-section-switch-handle-corner-radius);
```
macandcheese, line 13: Do we need both corner-radius props or can we apply a single to both the handle and track?

driskull: > Do we need both corner-radius props or can we apply a single to both the handle and track?

Probably not, ill remove and use the other var.

alisonailea (approved):

PR #10649, opened 2024-10-30
refactor(table): refactor internals to work around HTML parser issue

jcfranco:
**Related Issue:** #10495

## Summary

Split up table rendering to bypass HTML parser corrections that would break functionality and styling.

### DIFF
--- a/packages/calcite-components/src/components/table-header/table-header.tsx
+++ b/packages/calcite-components/src/components/table-header/table-header.tsx
@@ -74,7 +74,7 @@ export class TableHeader extends LitElement implements LoadableComponent {
    * @notPublic
    */
   /** TODO: [MIGRATION] This component has been updated to use the useT9n() controller. Documentation: https://qawebgis.esri.com/arcgis-components/?path=/docs/references-t9n-for-components--docs */
-  messages = useT9n<typeof T9nStrings>();
+  messages = useT9n<typeof T9nStrings>({ blocking: true });
 
   /** @notPublic */
   @property() numberCell = false;
--- a/packages/calcite-components/src/components/table-row/table-row.tsx
+++ b/packages/calcite-components/src/components/table-row/table-row.tsx
@@ -1,5 +1,7 @@
 import { PropertyValues } from "lit";
 import { LitElement, property, createEvent, h, JsxNode } from "@arcgis/lumina";
+import { createRef } from "lit-html/directives/ref.js";
+import { render } from "lit-html";
 import { Alignment, Scale, SelectionMode } from "../interfaces";
 import { focusElementInGroup, FocusElementInGroupDestination } from "../../utils/dom";
 import { RowType, TableInteractionMode, TableRowFocusEvent } from "../table/interfaces";
@@ -37,7 +39,7 @@ export class TableRow extends LitElement implements InteractiveComponent {
 
   private tableRowEl: HTMLTableRowElement;
 
-  private tableRowSlotEl: HTMLSlotElement;
+  private tableRowSlotEl = createRef<HTMLSlotElement>();
 
   // #endregion
 
@@ -281,7 +283,7 @@ export class TableRow extends LitElement implements InteractiveComponent {
       : this.rowType !== "head"
         ? "center"
         : "start";
-    const slottedCells = this.tableRowSlotEl
+    const slottedCells = this.tableRowSlotEl.value
       ?.assignedElements({ flatten: true })
       ?.filter(
         (el: TableCell["el"] | TableHeader["el"]) =>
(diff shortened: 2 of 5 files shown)


packages/calcite-components/src/components/table/table.tsx
```diff
@@ -127,7 +127,11 @@ export class Table extends LitElement implements LoadableComponent {
    *
    * @readonly
    */
-  @property() selectedItems: TableRow["el"][] = [];
+  @property() get selectedItems(): TableRow["el"][] {
```
jcfranco, line 130: TODO: clean up

packages/calcite-components/src/components/table/table.tsx
```diff
@@ -496,15 +487,37 @@ export class Table extends LitElement implements LoadableComponent {
         >
           <table
             ariaColCount={this.colCount}
-            ariaMultiSelectable={this.selectionMode === "multiple"}
             ariaRowCount={this.allRows?.length}
             class={{ [CSS.tableFixed]: this.layout === "fixed" }}
+            ref={(el) => {
+              if (!el || this.tableHeadSlotEl) {
```
jcfranco, line 493: TODO: tidy up, improve naming

packages/calcite-components/src/components/table/table.tsx
```diff
@@ -496,15 +487,37 @@ export class Table extends LitElement implements LoadableComponent {
...
+              this.tableFootSlotEl = tfootSlot;
+              tfoot.append(tfootSlot);
+            }}
             role={this.interactionMode === "interactive" ? "grid" : "table"}
           >
             <caption class={CSS.assistiveText}>{this.caption}</caption>
-            {this.renderTHead()}
-            {this.renderTBody()}
-            {this.renderTFoot()}
+            <thead>{/* slots added programmatically to work around HTML parser issue */}</thead>
```
jcfranco, line 518: TODO: Ensure all comments are consistent and reference #10495

packages/calcite-components/src/components/table-header/table-header.tsx
```diff
@@ -74,7 +74,7 @@ export class TableHeader extends LitElement implements LoadableComponent {
    * @notPublic
    */
   /** TODO: [MIGRATION] This component has been updated to use the useT9n() controller. Documentation: https://qawebgis.esri.com/arcgis-components/?path=/docs/references-t9n-for-components--docs */
-  messages = useT9n<typeof T9nStrings>();
+  messages = useT9n<typeof T9nStrings>({ blocking: true });
```
jcfranco, line 77: Needed to block these to ensure an a11y test passes (https://dequeuniversity.com/rules/axe/4.10/empty-table-header?application=axeAPI).

packages/calcite-components/src/components/table/table.tsx
```diff
@@ -496,15 +487,37 @@ export class Table extends LitElement implements LoadableComponent {
         >
           <table
             ariaColCount={this.colCount}
-            ariaMultiSelectable={this.selectionMode === "multiple"}
```
macandcheese, line 499: Was this throwing an error when role="table"? We can maybe add it back conditionally with if `this.interactionMode === "interactive"` which would be `role="grid"`

jcfranco: Good catch. Will fix!

packages/calcite-components/src/components/table-row/table-row.tsx
```diff
@@ -335,71 +348,58 @@ export class TableRow extends LitElement implements InteractiveComponent {
...
-    ) : (
-      <calcite-table-cell
-        alignment="center"
-        key="numbered-foot"
-        numberCell={true}
-        parentRowAlignment={this.alignment}
-      />
-    );
+  private renderSelectableCell(): void {
+    const cell: any =
```
maxpatiiuk, line 375: (optional) store the reference to the element that was created on previous render and update it rather than create a new element?
that would improve re-render performance

packages/calcite-components/src/components/table-row/table-row.tsx
```diff
@@ -335,71 +348,58 @@ export class TableRow extends LitElement implements InteractiveComponent {
...
-    ) : (
-      <calcite-table-cell
-        alignment="center"
-        key="selection-foot"
-        parentRowAlignment={this.alignment}
-        selectionCell={true}
-      />
-    );
+  private renderNumberedCell(): void {
+    const cell =
```
maxpatiiuk, line 358: (optional) store the reference to the element that was created on previous render and update it rather than create a new element?
that would improve re-render performance

packages/calcite-components/src/components/table/table.e2e.ts
maxpatiiuk, line 1: (optional) looks like all of the changes in this file can be applied directly on the `dev` branch? that would minimize work for the codemod

packages/calcite-components/src/components/table/table.tsx
```diff
@@ -179,10 +183,17 @@ export class Table extends LitElement implements LoadableComponent {
     this.listen("calciteInternalTableRowFocusRequest", this.calciteInternalTableRowFocusEvent);
   }
 
+  connectedCallback(): void {
```
maxpatiiuk, line 186: refactor to use this.listenOn(this.el.shadowRoot, ...)?

packages/calcite-components/src/components/table/table.tsx
```diff
@@ -217,6 +228,10 @@ export class Table extends LitElement implements LoadableComponent {
 
   // #region Private Methods
 
+  private handleSlotChange = (): void => {
```
maxpatiiuk, line 231: this could be a method rather than an arrow function

packages/calcite-components/src/components/table-row/table-row.tsx
```diff
@@ -161,14 +169,18 @@ export class TableRow extends LitElement implements InteractiveComponent {
 
   // #region Private Methods
 
+  private handleSlotChange = (): void => {
```
maxpatiiuk, line 172: this could be a method rather than an arrow function

maxpatiiuk:
Did you find out what was the main issue with rendering the table in parts with JSX?
was is that each lit-html template is parsed independently and so is not valid unless taken wholistically?
or that lit-html doesn't correctly append the elements to the table element? (appends them, but browser moves them out?)

depending on what was the cause, you might be able to make this PR a bit more declarative by directly calling lit-html's `render()` function to render desired content into the desired container. I have an example of this in the docs: https://qawebgis.esri.com/arcgis-components/?path=/docs/lumina-jsx--docs#rendering-jsx-outside-the-component

`render()` will also take care of updating the previously rendered element in place if you call it again with the same markup

a limitation though: you can only render one template into some parent container at a given time using `render()` - if you try to `render()` a different template into the same container, the 2nd `render()` call will overwrite the result of the 1st call. this might be a deal breaker since you need to render thead, tbody and tfoot into the same parent

packages/calcite-components/src/components/table/table.tsx
jcfranco: Fixed!

packages/calcite-components/src/components/table/table.tsx
jcfranco: I need the arrow function here so `this` refers to the component instance instead of the element.

jcfranco: Never mind. I **can** access the instance method.

maxpatiiuk: > I need the arrow function here so this refers to the component instance instead of the element.

if you refactor to use this.listenOn(), it would auto-bind to `this`, so should be good

jcfranco:
> Did you find out what was the main issue with rendering the table in parts with JSX?
was is that each lit-html template is parsed independently and so is not valid unless taken wholistically?
or that lit-html doesn't correctly append the elements to the table element? (appends them, but browser moves them out?)

My guess is that the final `template` was processed and adjusted by the HTML parser, whereas dynamically creating the table structure through DOM APIs bypasses this parser.

packages/calcite-components/src/components/table/table.e2e.ts
jcfranco: Moved to https://github.com/Esri/calcite-design-system/pull/10672.

Issue #8010, opened 2023-10-13
Investigate and leverage a github action to list contributors


driskull:
### Summary

We should acknowledge contributors to the project by listing them in the readme.

### Desired Outcome

Contributors are listed in the readme along with their profile photo, link to their GitHub, etc.

### Resources

Potential actions:

https://github.com/marketplace?category=&type=actions&verification=&query=contributor

geospatialem:
Would be a nice to have to have this available to showcase contributors from Hacktoberfest, at the conclusion of October.

driskull:
This one seems to be the most popular and best fits the use case: https://github.com/marketplace/actions/contribute-list

This one is made by Github, however the use case might not be the desired one: https://github.com/marketplace/actions/contributors-action

DitwanP:
🍡 Verified

geospatialem:
It doesn't look like the [list of contributors](https://github.com/Esri/calcite-design-system?tab=readme-ov-file#contributors) is being populated, reopening for further investigation

DitwanP:
Will keep an eye out for the next run of the "Update Contributors" action in order to verify, but since we wont have a release for a while I'll move this to `verified` and close out for now.

PR #10651, opened 2024-10-30
ci(actions): fix contributor list action failing

DitwanP:
**Related Issue:** #8010 

## Summary
Spent way too long hunting down a solution to a generic "Error: Not found" error message and found nothing, but it turned out to be just a case sensitive default path 🤦 so this should fix it. 🤞 

(Tested in my test repo)
image


jcfranco (commented):
🚀

jcfranco (approved):
🚀

jcfranco:
Nice!

PR #10661, opened 2024-10-31
refactor: handle stencil vs lumina boolean reflection differences

benelan:
## Summary

Resolve the issue brought up in https://github.com/Esri/calcite-design-system/pull/10509#discussion_r1792823363

The only properties affected by the change in how boolean properties are reflected are `icon`, `validationIcon`, and `download`. This PR fixes the utils that caused issues with `icon` and `validationIcon`. The `download` property seems to already be correct, for example:

https://github.com/Esri/calcite-design-system/blob/25319998739c3af17a5438d285c5931ea656285d/packages/calcite-components/src/components/link/link.tsx#L178


maxpatiiuk (commented):
could you apply this pr against the `dev` branch rather than `lumina`?
that way these changes would be preserved when we re-run the codemod.

if not, then I will hardcode these changes into the codemod so that they are preserved

benelan:
will do

PR #10663, opened 2024-10-31
fix(sort-handle): adjust icon color

driskull:
**Related Issue:** #7537

## Summary

- adjust icon color to use border-input color.


ashetland (approved):
🍰🎉🚀

Issue #7702, opened 2023-09-08
[combobox, list] filter should apply to groups


jcfranco:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The current filtering only applies to items, but should also include groups and any matches should show all items under a matching group.

### Expected Behavior

Filter applies to both items and groups.

### Reproduction Sample

https://codepen.io/jcfranco/pen/yLGgmWv?editors=100

### Reproduction Steps

1. Focus on a component's filter/input
2. Type the search text displayed in any use case's placeholder text
3. Group and child items are not displayed.

### Reproduction Version

1.8.0

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

Calcite (dev)

geospatialem:
Priority and expertise sought from the design team.

ashetland:
No hesitation from design. 🚀

geospatialem:
Verified in `3.0.0-next.88` with `combobox` and `list` groups specified in `combobox-item-group`'s `label` and `list-item-group`'s `heading`. 👍🏻

Issue #3454, opened 2021-11-09
[calcite-filter] Add support for custom filter function


jcfranco:
### Description

There is a need to support custom filtering for end-users (see #2208, #2209).

### Acceptance Criteria 

Allow users to pass a custom filter function to `calcite-filter`. This still needs more details, but a few items to consider:

* custom filter function should support pick/value list filtering
* could support async methods to allow for complex setups (need to gauge need for this)
* https://github.com/Esri/calcite-components/issues/2209#issuecomment-878616745 could be used as a base for the method
* Consider filtering in components (e.g., `list` via https://github.com/Esri/calcite-design-system/issues/6544)

cc @ffaubry, @dhrumil83, @AdelheidF, @benelan for additional requirement gathering

#### Which Component

`calcite-filter`

#### Example Use Case

See linked issues from description.

geospatialem:
@ffaubry @AdelheidF Calcite is considering this effort for a future milestone, and wanted to get your expertise with current and future workflows to help determine prioritization and need. Could you both speak to the impact and what you would expect CC would support?

ffaubry:
We would like to have two features:
- a way to say what property of the list item we would like to filter on: label or description or both
- pass a function that would be invoked for each list item with the filter value and the list item itself

@geospatialem  is it what you are looking for?

AdelheidF:
I think the discussion in the issue that Dhrumil submitted is still valid. We still would like to add a second filter, e.g. field type, to a list with a text filter. See https://github.com/Esri/calcite-design-system/issues/2209#issuecomment-878694545.

geospatialem:
Thank you both, that helps us with prioritization. Will be assigning to a future milestone in the next few weeks.

isaacbraun:
Closing as duplicate of #8877. Details have been merged .

PR #10664, opened 2024-10-31
fix(list): include groups in filtering

anveshmekala:
**Related Issue:** #7702 

## Summary

- Include `calcite-list-item-group` in filtering.
- group's `heading` can be used as `filterProp`.


packages/calcite-components/src/components/list/list.tsx
```diff
@@ -160,7 +171,7 @@ export class List
 
   @state() assistiveText: string;
 
-  @state() dataForFilter: ItemData = [];
+  @state() dataForFilter: (ItemData | ItemGroupData)[] = [];
```
driskull, line 174: Instead of including separate group data, can we just add the group data to each item within a group? It would just be the `heading` property added to each item.

If we just added group heading to the itemData, then we wouldn't need the logic in `filterGroupItems`

Theres already existing logic to show a group if one or more items match and hide a group if not. The only part missing is the group data attached to each item.

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -744,7 +761,7 @@ export class List
     const filteredItems = filterPredicate
```
driskull, line 761: The way this is currently structured, filterPredicate would only affect items and not groups.

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -235,14 +246,22 @@ export class List
...
-  @property() filteredData: ItemData = [];
+  @property() filteredData: (ItemData | ItemGroupData)[] = [];
+
+  /**
+   * The currently filtered `calcite-list-item's` & `calcite-list-item-group's`.
+   *
+   * @readonly
+   * @deprecated Use `filteredResults` instead.
+   */
+  @property() filteredItems: ListElement[] = [];
```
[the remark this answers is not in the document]
anveshmekala: since it is a short notice, its better to deprecate first. To avoid confusing with naming, `filteredResults` is added.

If we are including groups in filter, the user should be able to access the groups me thinks.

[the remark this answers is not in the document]
driskull: Part of my reasoning for not including groups is that they aren't results. They can't be selected, they don't have a value, etc.

packages/calcite-components/src/components/list/list.tsx
anveshmekala: The above wouldn't work where only group heading is matched (nested group) but none of the items or ancestor group have a matching text.

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -1118,13 +1175,15 @@ export class List
...
+            numberStringFormatter.localize(filteredResults.length.toString()),
           )}
         </div>
-        {filteredItems.length ? (
+        {filteredResults.length ? (
           <ol key="aria-item-list">
-            {filteredItems.map((item) => (
-              <li>{item.label}</li>
+            {filteredResults.map((item) => (
+              <li role={(item as ListItemGroup).heading ? "group" : "none"}>
```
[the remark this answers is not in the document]
anveshmekala: Agree, the `role` attribute wouldn't make a diff how AT handles the information. Idea was to provide the context about a possible group heading matched & semantic structure is not the right fit.

packages/calcite-components/src/components/list/list.tsx
anveshmekala: Meaning to ask, should the `filterPredicate` include groups?  It wasn't mentioned here https://github.com/Esri/calcite-design-system/issues/3454 but (https://github.com/Esri/calcite-design-system/issues/7702) was created later.

packages/calcite-components/src/components/list/list.tsx
driskull: @anveshmekala it should work.

> The current filtering only applies to items, but should also include groups and any matches should show all items under a matching group.

**any matches should show all items under a matching group**

According to the request, if a group matches, all items under that group should also match.

packages/calcite-components/src/components/list/list.tsx
driskull: I don't think it should, just wanted to point out the inconsistency with treating groups like items.

packages/calcite-components/src/components/list/list.tsx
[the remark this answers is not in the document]
driskull: My point was that if you set the role to `none` on a list-item it takes away the default list-item role.

packages/calcite-components/src/components/list/list.tsx
anveshmekala: Its working as expected! 🎉

packages/calcite-components/src/components/list/list.tsx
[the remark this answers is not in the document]
anveshmekala: Agree item-group is more of a  presentational component, but using `heading` as `filterProp` can affect how filtering is done. And users should be able to lookup the API for any potential matches for filterText.

Did think of adding direct children items of the group as filteredItems instead of the item-group but that can be misleading and not consistent with how filteredItems are added when item is nested inside a item without a group.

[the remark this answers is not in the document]
anveshmekala: It will make it easier if we don't include `item-group` in filteredItems but filteredItems can't be empty if there is a match.

Did think of adding direct children items of the item-group as filteredItems instead of the item-group but that can be misleading and not consistent with how items are added to `filteredItems` when item is nested inside a item without a group. Thoughts?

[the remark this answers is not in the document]
jcfranco: After reviewing the issue again, I _think_ the request was primarily visual and not related to `filteredItems`. My earlier context was minimal, so apologies for any confusion. 😅

I agree on including them when they are valueless is a bit odd. 

Could we proceed with the behavior while keeping `filteredItems` as is?

packages/calcite-components/src/components/list/list.tsx
jcfranco: If not, is the alternative `metadata` + `filterProps`? While functional, it introduces boilerplate for a common filtering workflow using information already present in the list.

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -813,11 +811,12 @@ export class List
     this.updateFilteredData();
   }
 
-  private getItemData(): ItemData {
+  private getItemData(): ItemData[] {
     return this.listItems.map((item) => ({
       label: item.label,
       description: item.description,
       metadata: item.metadata,
+      heading: item.parentElement?.closest(listItemGroupSelector)?.heading || "",
```
driskull, line 819: Can we store the parentGroup elements on connectedCallback? That way its only ever queried in the DOM one time. In this scenario its queried every time `getItemData()` is called.

The other part is that this is only getting the first parent group element. Shouldn't it get all ancestor group elements and map to their headings?

driskull (commented):
Getting closer @anveshmekala! 👍

packages/calcite-components/src/components/list/list.tsx
anveshmekala: >The other part is that this is only getting the first parent group element. Shouldn't it get all ancestor group elements and map to their headings?

Good catch. Each item should store the heading of all ancestors group elements.

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -813,15 +819,27 @@ export class List
...
+      heading: this.getGroupHeading(item),
       el: item,
     }));
   }
 
+  private getGroupHeading(item: ListItem["el"]): string {
+    let heading = "";
+    this.listItemGroups.forEach((group) => {
+      if (group.contains(item)) {
+        heading += group.heading;
```
anveshmekala, line 836: We could also store heading of ancestor as Array object.

Leaning towards string concatenation for performance gains with `filter` method.

driskull: Yeah I was thinking array because string concat would match two words from different headings combined. Can we change to use an array here?

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -649,6 +653,10 @@ export class List
     }
   }
 
+  private setListItemGroups(): void {
+    this.listItemGroups = Array.from(this.el.querySelectorAll(listItemGroupSelector));
```
driskull, line 657: The other alternative would be something like this:

```tsx
let ancestor = el.closest("calcite-list-item-group");
const headings: string[] = [];
while (ancestor) {
  headings.push(ancestor.heading);
  ancestor = ancestor.closest("calcite-list-item-group");
}
```

I'm not sure which would be more performant, calling closest() a few times on connectedCallback or calling `setListItemGroups` on each mutation. Thoughts @jcfranco?

driskull (approved):
This looks good to me 👍 

The querying of groups comment is pretty minor and would require storing that information on each item so what you currently have seems to make the most sense.

PR #10669, opened 2024-11-01
feat(list, list-item, list-item-group): add component tokens 

driskull:
**Related Issue:** #7180

## Summary

adds new component tokens, e2e tests, demo page examples, and chromatic theme tests.

#### List

--calcite-list-background-color: Specifies the component's background color.

#### ListItem

 --calcite-list-background-color-hover: Specifies the component's background color when hovered.
 --calcite-list-background-color-press: Specifies the component's background color when pressed.
 --calcite-list-background-color: Specifies the component's background color.
 --calcite-list-border-color: Specifies the component's border color.
 --calcite-list-content-color: Specifies the content color.
 --calcite-list-description-color: Specifies the description color.
 --calcite-list-icon-color: Specifies the component's icon color.
 --calcite-list-label-color: Specifies the label color.
 --calcite-list-selection-border-color: Specifies the component's selection border color.

#### ListItemGroup

--calcite-list-background-color: Specifies the component's background color.
--calcite-list-color: Specifies the component's color.


alisonailea (commented):
I don't think these components need stateful tokens and because the components are all related, any shared values can have a shared token. For example `--calcite-list-background-color` can be used for list, list-item-group, and list-item

driskull:
> I don't think these components need stateful tokens

I think list-item does need stateful tokens. Hovering the whole host shouldn't change the background color. That wouldn't work for nested items.

alisonailea (approved):

packages/calcite-components/src/components/list-item-group/list-item-group.scss
```diff
@@ -1,5 +1,14 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-list-background-color: Specifies the component's background color.
```
macandcheese, line 6: I’d expect to support “—calcite-list-item-background-color” as well. 

Consider this use case which is representative of how most of our users want to theme: https://community.esri.com/t5/calcite-design-system-questions/change-listitem-background-color-in-code/m-p/1557125#M767

packages/calcite-components/src/components/list-item/list-item.scss
```diff
@@ -1,6 +1,21 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-list-background-color-hover: Specifies the component's background color when hovered.
+ * @prop --calcite-list-background-color-press: Specifies the component's background color when pressed.
```
macandcheese, line 7: Similar to above, does this need to be “list-item” - what hover would we actually be theming here on the parent?

packages/calcite-components/src/components/list-item-group/list-item-group.scss
alisonailea: We are limiting tokens when shared between grouped components. This is documented in the wiki.

packages/calcite-components/src/components/list-item/list-item.scss
alisonailea: we are limiting tokens when shared between grouped components. This is documented in the wiki.

packages/calcite-components/src/components/list-item-group/list-item-group.scss
macandcheese: How is this different than Accordion Item / Accordion - where we have `-item-background-color`? https://github.com/Esri/calcite-design-system/pull/10181

packages/calcite-components/src/components/list-item/list-item.scss
driskull: Yeah it is a bit weird to be targeting the parent hovering when its really the items being hovered.

packages/calcite-components/src/components/list-item-group/list-item-group.scss
driskull: @alisonailea can you comment on the above? Currently, it seems like parent/child tokens are inconsistent.

macandcheese: The “item tokens” path seems the easiest to follow both as a maintainer (IMO) and as someone who uses these theming options frequently day to day.

Using the Community post as just one example, it also seems to align with expectations from external folks - theming the element you want to adjust with component-specific nomenclature / rules.

driskull: Need direction on this one. @alisonailea

driskull:
@alisonailea @macandcheese can you review again?

alisonailea (approved):
🚀

macandcheese (approved):

PR #10671, opened 2024-11-01
fix: updated offset and properties

arowles:
**Related Issue:** #

## Summary
**updated**

offset-16
offset-24
offset-32
properties-16

image
image


allieorth (approved):
thanks for making these changes!

Issue #10674, opened 2024-11-03
Update Angular examples to use components directly


jcfranco:
### Description

After https://github.com/Esri/calcite-design-system/issues/10465 lands, the [Angular example](https://github.com/Esri/calcite-design-system/tree/dev/examples/components/angular) and [template](https://codesandbox.io/p/sandbox/little-fire-2hx4qh?file=%2Fpackage.json%3A26%2C13) need to be updated to consume components without the wrapper.

### Which Component

N/A

### Resources

https://angular.dev/errors/NG8001#debugging-the-error

geospatialem:
Verified on the `dev` branch 🥇

PR #10686, opened 2024-11-04
chore(angular): update example to use custom elements directly

benelan:
**Related Issue:** #10674

## Summary

Update the Angular example to use `@esri/calcite-components` directly, instead of the now removed `@esri/calcite-components-angular` wrapper.


jcfranco (approved):
Awesome, thanks! ✨🚀✨

Should this use the `docs` type?

PR #10689, opened 2024-11-05
feat(link): add component tokens

driskull:
**Related Issue:** #7180

## Summary

Add `link` component tokens.

`--calcite-link-text-color`: Specifies the component's text color.


packages/calcite-components/src/components/link/link.scss
```diff
@@ -1,3 +1,13 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-link-text-color: Specifies the component's text color.
+ * @prop --calcite-link-border-color: Specifies the component's border color.
+ * @prop --calcite-link-icon-color: Specifies the icon color.
```
alisonailea, line 8: link icon color should inherit from the link text color. If a user wants to change the design pattern they can apply a `--calcite-icon-color` to the link

alisonailea (dismissed):

packages/calcite-components/src/components/link/link.scss
```diff
@@ -1,3 +1,13 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-link-text-color: Specifies the component's text color.
+ * @prop --calcite-link-border-color: Specifies the component's border color.
```
driskull, line 7: @alisonailea i'm not sure what to call this one. Its an underline so should it be `--calcite-link-underline-color`? 

Its not really a border because its using a linear gradient background.

driskull:
@alisonailea @macandcheese can I get a review of this one?

packages/calcite-components/src/components/link/link.scss
```diff
@@ -56,13 +64,13 @@ calcite-icon {
...
+    @apply transition-default
       relative
       inline
       border-none
       bg-transparent
       p-0;
+    color: var(--calcite-link-text-color, var(--calcite-color-text-link));
     line-height: inherit;
     white-space: initial;
     background-image: linear-gradient(currentColor, currentColor),
```
macandcheese, line 76: For this underline, maybe we could pass it through with `--calcite-link-underline` ? Kind of a strange prop, I know design was maybe going to re-evaluate this, but for now we'd need to let users override it somehow. If they've overridden the link color they'll need to change this right?

macandcheese (approved):
Looks good, just some small q / notes

packages/calcite-components/src/components/link/link.scss
driskull: Lets just think about it. we can add it later.

packages/calcite-components/src/components/link/link.scss
```diff
@@ -1,3 +1,11 @@
+/**
+ * CSS Custom Properties
+ *
+ * These properties can be overridden using the component's tag as selector.
+ *
+ * @prop --calcite-link-text-color: Specifies the component's text color.
+ */
```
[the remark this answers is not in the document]
driskull: Yeah, down the road.

PR #10702, opened 2024-11-06
feat(list)!: require label for context when dragging between lists

geospatialem:
**Related Issue:** #7537

## Summary
- Modifies the component's `label` property to be required
- Provides context to the `label` property, where its value specifies the name when dragging between lists discussed in https://github.com/Esri/calcite-design-system/issues/7537#issuecomment-2460544077 and the subsequent thread

BREAKING CHANGE: The component's `label` property is required to provide context between lists when dragging.


driskull (approved):
👍

PR #10722, opened 2024-11-13
build(deps): add `@floating-ui/utils` dep

jcfranco:
**Related Issue:** N/A

## Summary

Adds [`@floating-ui/utils`](https://www.npmjs.com/package/@floating-ui/utils) as a dep, which is required by [`composed-offset-position`](https://www.npmjs.com/package/composed-offset-position).

This should address Yarn install errors.

cc @ffaubry


driskull (approved):
👍

Issue #10668, opened 2024-10-31
Focus moves to the time-picker popover when clicking input-time-picker's text input


eriklharper:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Clicking or triple-clicking the input in `input-time-picker` automatically moves focus to the first field in the `time-picker` popover.

### Expected Behavior

The `time-picker` should never gain focus unless it is explicitly clicked or tabbed into.  Focus should stay in the input when it is clicked with the mouse.

### Reproduction Sample

https://codepen.io/eriklharper/pen/xxvjeze

### Reproduction Steps

1. Left-click with the mouse in the input
2. Result: focus moves to the hour field in the `time-picker` popover.

### Reproduction Version

next

### Priority impact

impact - p1 - need for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

DitwanP:
🍡 Verified locally on `dev`

PR #10729, opened 2024-11-14
fix(input-time-picker): retain focus inside the input when clicked

eriklharper:
**Related Issue:** #10668

## Summary

This restores the original focus behavior when clicking the input in `input-time-picker`.  Focus remains in the input for typing and selecting text.


packages/calcite-components/src/components/input-time-picker/input-time-picker.e2e.ts
```diff
@@ -888,6 +888,7 @@ describe("calcite-input-time-picker", () => {
       expect(await getFocusedElementProp(page, "tagName", { shadow: true })).toBe("CALCITE-INPUT-TEXT");
 
       await page.keyboard.press("ArrowDown");
+      await page.keyboard.press("Tab");
```
eriklharper, line 891: This is needed now that focus doesn't automatically move to the popover when it is opened.  This is more intuitive behavior because the focus doesn't shift away without the user explicitly tabbing into it.

packages/calcite-components/src/components/popover/popover.tsx
```diff
@@ -145,6 +146,9 @@ export class Popover
   /** Specifies the heading level of the component's `heading` for proper document structure, without affecting visual styling. */
   @property({ type: Number, reflect: true }) headingLevel: HeadingLevel;
 
+  /** Specifies whether the first tabbable child element will be focused when the popover is opened. */
+  @property({ reflect: true }) initialFocus: FocusTargetOrFalse;
```
Elijbet, line 150: ✨ 💡 Cool 💡 ✨ find! I didn't know about this option.

driskull: Do we need this to be a public property? I'm not sure we want to add a property like this. 

This property also wouldn't be honored if the `setFocus()` method was used so I'm not sure it needs more doc to say its only for focus trapping.

eriklharper: If there's a way to provide an "internal" hook to this option, that would also work for this use case; it doesn't have to be exposed publicly.  Also, `input-date-picker` currently utilizes this option in the `connectFocusTrap` function that it consumes directly, but probably makes sense to refactor it at some point to consume `calcite-popover` internally.

eriklharper:
> Could you just disable focus trapping until you want to start focus trapping instead of adding an initialFocus property?

Probably, but certainly trickier to manage the `focusTrapDisabled` property internally.  First it would have to be converted to a mutable property and we would also have to add logic that checks if the user is setting it from the outside, just a lot more to manage than exposing this setting that's part of the FloatingUI api.

packages/calcite-components/src/components/popover/popover.tsx
driskull: I would suggest marking this property as `@internal/@private` in order to publicly document it if we do go this route.

I'm not sure this is the best route because it deviates from how all the other focus trapping works. In general, when you open something you focus trap it and place focus inside of what is being focus trapped.

However, in this case, you are creating the focus trap but the user isn't within the trapped component yet so its really doing nothing. 

cc @jcfranco

driskull: Ideally, you wouldn't need to ever set focus in the popover. focus would always remain on the input and use keyboard arrows/enter/space for handling things. This is what the `action-menu` component does.

eriklharper: Just replaced the doc wording with `@internal`.

driskull (approved):
I'm good with this as an internal property for now. 👍

jcfranco:
I'm also in agreement to proceed with this as an internal prop for now.

This might be useful for other component authors down the line. Introducing a `focusTrapOptions` prop with a subset of options, might offer more flexibility to devs, but we might need to wait until more use cases come up.

jcfranco (changes requested):

packages/calcite-components/src/components/popover/popover.tsx
```diff
@@ -145,6 +146,12 @@ export class Popover
   /** Specifies the heading level of the component's `heading` for proper document structure, without affecting visual styling. */
   @property({ type: Number, reflect: true }) headingLevel: HeadingLevel;
 
+  /**
+   * Specifies whether focus should move to the popover when the focus trap is activated.
+   *  `@internal`
+   */
+  @property({ type: Boolean }) initialFocusTrapFocus: FocusTargetOrFalse;
```
jcfranco, line 153: WDYT about `focusTrapInitialFocus`? I'll leave it up to you.

jcfranco (approved):
Awesome!

packages/calcite-components/src/components/popover/popover.tsx
jcfranco, line 153: As a follow-up, could you add an E2E test for this?

eriklharper: done

eriklharper: woops, didn't see this comment.  Ah well

Issue #10715, opened 2024-11-08
[Tooltip] Safari: tooltip will reappear and block actions with the menu


doreenbrinkman:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The fix to #10416 is not working for Safari.
Follow instructions in issue linked and test in Safari browser (codepen below).  Working for Chrome and Firefox.

### Expected Behavior

Action is not blocked by tooltip

### Reproduction Sample

https://codepen.io/doreenbrinkman/pen/vYoQexP

### Reproduction Steps

1.  Click on save icon
2.  Move to action
3.  Action is blocked by tooltip

### Reproduction Version

2.13.2

### Relevant Info

Safari

### Priority impact

impact - p2 - want for an upcoming milestone

### Impact

Makes choosing the action more difficult

### Calcite package

- [X] @esri/calcite-components
- [X] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Data Pipelines

geospatialem:
Spike needed to determine any considerations and/or limitations to consider for support in the Safari browser.

DitwanP:
🍡 Verified locally on `dev`

PR #10737, opened 2024-11-15
fix(tooltip): fix closeOnClick property in Safari

driskull:
**Related Issue:** #10715

## Summary

- fixes closeOnClick for the Safari browser


jcfranco (approved):
🚀

PR #10738, opened 2024-11-15
refactor(popover): remove unnecessary CSS

driskull:
**Related Issue:** N/A

## Summary

- No longer necessary since the internal element is hidden when closed now


jcfranco (approved):
✨🧹✨

Issue #8697, opened 2024-02-06
[Block] Make `heading` optional, add a required `label` property


macandcheese:
### Description

Blocks can often be used to group controls or content within a Panel, and it's not always desirable to have displayed text. The current set up requires a `heading` property - meaning to opt-out of this, a user needs to "null this out" - leading to no AT support.

We should add a required `label` property to Block to facilitate this use case while ensuring AT is provided with proper context. At the same time, we can make `heading` optional.

### Proposed Advantages

This will support more layout options for Blocks, without requiring a workaround of `heading` that leads to a less accessible experience.

### Which Component

Block

### Relevant Info

cc @geospatialem 

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

DitwanP:
🍡 Verified

jcfranco:
This unintentionally makes `label` required for all configurations. Since `label` can fallback to `heading` when present, we should make it optional. We should also log a runtime warning if `label` and `heading` are both unset.

Bright side, this mitigates the breaking change.

cc @AdelheidF

geospatialem:
Verified via typings on `dev` and via logger message when either `heading` and `label` are not specified on the component in `3.0.0-next.53`.

Issue #10382, opened 2024-09-24
[Block] Provide consistent use of sorting similar to `list`


geospatialem:
**monday.com sync:** #8393032012

### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

`block` currently offers a `handle` attribute, but the sorting functionality is handled via a parent internal unsupported and undocumented component. Instead we should offer support to a set of grouped `blocks` in alignment with `list` and `list-item`.

### Acceptance Criteria

- [x] Provide a new component, `block-group` with `dragEnabled` that allows specified `block`'s to be movable
- [x] Use `sort-handle` component
- [x] Deprecate the `handle` property (to be removed in `4.0`)
- [x] Deprecates the internal `sortable-list` (to be removed in `4.0`)

### Relevant Info

- Consistency across the design system
- Removal of internal and undocumented components
- Use case of Esri teams seeking to provide sorting on `block` components

### Which Component

`block`

### Example Use Case

https://codepen.io/geospatialem/pen/XWvJbEm

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

geospatialem:
Spike needed to determine the structure of the component needed to provide sorting on specified `blocks`.

driskull:
For blocks to be sortable, they need to be child items which means that they need some kind of parent component to allow them to be sortable which would replace the sortable-list component. What would this component be called?

driskull:
@geospatialem I think this would be a new component so should probably be tagged like that

driskull:
What should this new component be called? @geospatialem @DitwanP

geospatialem:
> What should this new component be called? @geospatialem @DitwanP
>
> What about block-sections, would they ever need sorting as well?

Maybe we should approach the naming in case we would need to apply the same functionality to block-sections? @driskull, do you think we should accommodate sorting on both components, `block` and `block-section` with the effort?

[...]

PR #10739, opened 2024-11-15
feat(block)!: add required label property and mark heading as optional

anveshmekala:
**Related Issue:** #8697 

## Summary

- Adds `label` property as required to provide context for AT users.
- `heading` is no longer a required property.

BREAKING CHANGE: The component's `label` property is a required property and `heading` is an optional property.

### DIFF
--- a/packages/calcite-components/src/components/block/block.tsx
+++ b/packages/calcite-components/src/components/block/block.tsx
@@ -90,7 +90,6 @@ export class Block
   /**
    * The component header text.
    *
-   * @required
    */
   @property() heading: string;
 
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/block/block.tsx
```diff
@@ -378,7 +385,7 @@ export class Block
 
     const headerNode = (
       <div class={CSS.headerContainer}>
-        {this.dragHandle ? <calcite-handle label={heading} /> : null}
+        {this.dragHandle ? <calcite-handle label={heading || label} /> : null}
```
driskull, line 388: Unrelated, but maybe this should use the newer `calcite-drag-handle` at some point? Should we have an issue. @jcfranco

driskull (approved):
Niiiiice 😄

anveshmekala:
Adding @geospatialem for Accessibility review.

packages/calcite-components/src/components/block/block.tsx
jcfranco: Good idea.

driskull: I think #10382 should cover it 👍

Issue #7708, opened 2023-09-08
Use localization patterns vs hardcoded strings


jcfranco:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The following hardcoded strings should be pulled into t9n bundles or use a formatting util:

1. https://github.com/Esri/calcite-design-system/blob/fc2c720bc69db81075b7e021d69c4cb7e2c2b5a9/packages/calcite-components/src/components/stepper-item/stepper-item.tsx#L228 (trailing period)
2. ✅ https://github.com/Esri/calcite-design-system/blob/3e1ed2dedeea09a6118e2f639201f5861dd492fd/packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx#L551 (`Date Format:`)
3. ~https://github.com/Esri/calcite-design-system/blob/8a575e451acb6b33d65124988b2e2f7ea8ec8b81/packages/calcite-components/src/components/tip-manager/tip-manager.tsx#L274 (label + index)~ deprecated

### Expected Behavior

We avoid the hardcoded strings reported above.

### Reproduction Sample

https://codepen.io/jcfranco/pen/RwEpwbL?editors=100

### Reproduction Steps

1. Open codepen
2. Look at console to see the hardcoded assistive text

This repro sample is specific to one component. Links in the description show the hardcoded values.

### Reproduction Version

1.8.0

### Priority impact

p3 - want for upcoming milestone

### Impact

Hardcoded values can interfere with AT and also not display as expected in certain locales.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

Calcite (dev)

jcfranco:
@annierm18 Could you help confirm needing to localized string templates for items 1 and 3?

annierm18:
@jcfranco I am just seeing your comment now--sorry! I will get back to you with an answer

jcfranco:
I've got a PR for item #2.

jcfranco:
Remaining item has been moved to #10890 (thanks @geospatialem!).

DitwanP:
🍡 Verified locally on `dev`

PR #10741, opened 2024-11-16
fix(input-date-picker): remove hardcoded strings

jcfranco:
**Related Issue:** #7708 

## Summary

This addresses item 2 from #7708.


packages/calcite-components/src/components/input-date-picker/assets/t9n/messages.json
```diff
@@ -1,4 +1,4 @@
 {
   "chooseDate": "Choose date",
-  "dateFormat": "Date Format:"
+  "dateFormat": "Date Format: {format}"
```
anveshmekala, line 3: We could avoid the `{format}` literal in messages and do below in JSX. Thoughts?

`{messages.dateFormat}  {this.localeData?.placeholder)}`

jcfranco: I think we need it in the template so translations can position accordingly.

anveshmekala (approved):
🏅

Issue #9524, opened 2024-06-05
Design tokens: Refactor for alignment across the design system


geospatialem:
### Description

Refactor the following tokens to be in alignment with Figma and code prior to completing the effort of #7180:

- [x] Remove the following from `semantic > font`:
    - [x] letter-spacing
    - [x] paragraph-spacing
    - [x] text-decoration
    - [x] text-case
- [x] Reassign `font > font-weight > bold` in semantic tokens  to `{core.font.weight.demi}`

In #7180 epic
- [x] Make sure date-picker-month `.week-header` uses `--calcite-font-weight-bold`

### Proposed Advantages

Consistency across design and engineering and in alignment for design tokens.

### Which Component

All components

### Relevant Info

Related to the epic of #7180

### Calcite package
- [X] @esri/calcite-design-tokens

alisonailea:
Looking into this again. I remember doing this already. Going to check some branches to see if I can get this in sooner.

DitwanP:
🍡 Verified

PR #10744, opened 2024-11-16
deprecate(letter-spacing, paragraph-spacing, text-decoration, text-case): align with UIKit

alisonailea:
**Related Issue:** #9524

## Summary

letter-spacing, paragraph-spacing, text-decoration, and text-case tokens are deprecated to align tokens with the Calcite UIKit

### DIFF
--- a/packages/calcite-design-tokens/src/core/font.json
+++ b/packages/calcite-design-tokens/src/core/font.json
@@ -556,6 +556,7 @@
         "none": {
           "value": "none",
           "type": "textDecoration",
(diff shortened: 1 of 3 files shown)


macandcheese (approved):
Nice. I think we'll need to add support for a "deprecated chip" in the doc site reference table like we do for component doc tables: https://developers.arcgis.com/calcite-design-system/foundations/tokens/reference/

We could also just filter them out if these were never intended to be public.

Issue #10438, opened 2024-09-30
[input-time-zone] Remove unused time zone and country translations


jcfranco:
### Description

https://github.com/Esri/calcite-design-system/issues/10289 ensures consistent time zones across browsers, so there are translations that are no longer valid.

### Proposed Advantages

Leaner translation bundle.

### Which Component

`calcite-input-time-zone`

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

DitwanP:
🍡 Verified

Issue #10289, opened 2024-09-12
calcite-input-time-zone issues


richiecarmichael:
### Check existing issues
- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

#### 1. [related: [#8112](https://github.com/Esri/calcite-design-system/issues/8112), [#8005](https://github.com/Esri/calcite-design-system/issues/8005)] `Input Time Zone` does not support a few IANA names like:
   - America/Argentina/Buenos_Aires
   - America/Indiana/Indianapolis
   - America/Kentucky/Louisville
   - Asia/Ho_Chi_Minh

#### 2. [related: [#8005](https://github.com/Esri/calcite-design-system/issues/8005)] `Input Time Zone` does not support IANA time zone representing fixed offsets like:
   - Etc/GMT
   - Etc/GMT+1
   - Etc/UTC

#### 3. Other irregularities:
  - `Pacific/Fiji` maps to `"Fiji, Fiji"`. Fiji is the name of the country not a city.
  - `Asia/Hong_Kong`, should it be "Hong Kong, China" not "Hong Kong, Hong Kong"
  - ditto for `Asia/Macau` 

 #### 4. When `value` is programmatically updated the country portion is omitted.

#### 1. Expected:
  - "Buenos Aires, Argentina"
  - "Indianapolis, United States"
  - "Louisville, United States"
  - "Ho Chi Minh City, Vietnam"

#### 2. Expected:
As discussed with @emmilaakso212 and @jcfranco, perhaps the component should have a new leading group entitled "Fixed" with entries for "Coordinated Universal Time" and other fixed offsets.
![image](https://github.com/user-attachments/assets/958a028a-6a60-491c-bdab-ce8fcb2135fb)

### Reproduction Sample
https://codepen.io/richiecarmichael/pen/yLdwdyZ?editors=1000

#### For (1) to (3):
- Click [here](https://developers.arcgis.com/calcite-design-system/components/input-time-zone/) to open the input-time-zone sample page.
- Enter text in the `value` property.
- Observe updated widget.

#### For (4):
- Click [here](https://codepen.io/richiecarmichael/pen/yLdwdyZ?editors=1000) to open codepen app
- Click blue "change time zone to NYC" button
- Observe first `input-time-zone` component.

### Reproduction Version
2.12.1

### Relevant Info
Windows w/ Microsoft Edge version 128

### Regression?
No

### Priority impact
[...]

PR #10747, opened 2024-11-16
refactor(input-time-zone)!: drop obsolete time zone translations

jcfranco:
**Related Issue:** #10438 

## Summary

This removes translations no longer needed after #10289.

BREAKING CHANGE: Developers using `messageOverrides` will need to make sure they do not reference removed entries.


anveshmekala (approved):
🧹

PR #10786, opened 2024-11-19
feat: deprecate widthScale/heightScale in favor of width/height

Elijbet:
**Related Issue:** #6172

## Summary
Refactor to consolidate `width/height` and `widthScale/heightScale` into a single property `width` with `s / m / l / auto / full`  as options, and `height` with `s / m / l`. Deprecate `widthScale` and `heightScale` properties and the `half` value.

Components addressed:

- `button`
- `dialog`
- `dropdown`
- `notice`
- `segmented-control`
- `select`
- `sheet`
- `shell-panel`
- `split-button`

deprecate: deprecate widthScale/heightScale properties in favor of width/height.


packages/calcite-components/src/components/interfaces.ts
```diff
@@ -28,7 +29,7 @@ export type SelectionMode =
   | "multiple";
 export type Scale = "s" | "m" | "l";
 export type Status = "invalid" | "valid" | "idle";
-export type Width = "auto" | "half" | "full";
+export type Width = "s" | "m" | "l" | "auto" | "half" | "full";
```
Elijbet, line 32: @Erik suggested to do export type `Width = Scale & "auto" | "half" | "full"`; Same for Height above. Height would just be export type `Height = Scale`. 

I felt like it might make things a bit confusing to read. Height, Width, and Scale are different props even though they have an overlap in type (with type Scale). s/m/l for scale does something else for width/height. Should they just have their own type names? It hurts my eyes to read type Height = Scale :)

Erik's suggestion is purely for DRY reasons. It's really just a union type to keep from having to define "s" | "m" | "l" in 3 different places. type Height = Scale, does look weird, but it has more semantics at least. 

WDYT @jcfranco

packages/calcite-components/src/components/dialog/dialog.scss
```diff
@@ -189,32 +189,32 @@ calcite-panel {
...
+        &-active {
+          inset-block-start: 0;
         }
       }
     }
   }
 }
 
-:host([width="small"]) .dialog {
+.width-s .dialog {
```
jcfranco, line 217: I don't think this is correct since dialog and width classes are applied to the same element now. Should this be `.dialog.width-s`?

packages/calcite-components/src/components/interfaces.ts
jcfranco: I see your point about type `type Height = Scale` being potentially confusing, but I think it’s fine given the consolidation of `widthScale` and `heightScale` props.

packages/calcite-components/src/components/modal/modal.e2e.ts
```diff
@@ -28,6 +28,24 @@ describe("calcite-modal", () => {
     t9n("calcite-modal");
   });
 
+  describe("reflects", () => {
```
jcfranco, line 31: Given `modal` is deprecated and not being updated in this PR, this is no longer required. I'll defer to you if you want to leave this extra test coverage.

jcfranco (approved):
Awesome stuff, @Elijbet! 🎉

Before merging, could you update the PR to only list components with width/height-prop-related updates?

The commit type might also need to be `feat` since we’re adding new properties for consistency. If there weren’t new props involved, `deprecate` would’ve been perfect. @geospatialem @DitwanP 

Once the comments are addressed, this should be good to go!

jcfranco:
> The commit type might also need to be feat since we’re adding new properties for consistency. If there weren’t new props involved, deprecate would’ve been perfect. @geospatialem @DitwanP

Actually, we could leave this as is and use commit overrides to follow the steps outlined [here](https://github.com/Esri/calcite-design-system/pull/10346#issuecomment-2377829419). 

Sidebar: @DitwanP is the above recommendation in our internal doc? I wasn't able find it.

Issue #8696, opened 2024-02-06
[Chip] Ensure the `label` prop is provided to assistive technologies


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Summary

Chip's `label` property should be provided over the `value` or slotted text.

### Actual Behavior

The component's `label` property provides context to AT, where `value` is not provided to assistive technologies or when a parent `chip-group` is present.

### Expected Behavior

As part of an upcoming breaking change release, we should consider making `label` required from our discussions in https://github.com/Esri/calcite-design-system/pull/6075#discussion_r1055903130. This way the internals stay the same and label is read to assistive technologies, and required. Value should just be optionally used for getting value in selection mode.

If enforcing a required `label`, the `value` can be optional and not read back to assistive technologies. This seems to be a similar pattern in other components, such as `alert` and the parent `chip-group`:

https://github.com/Esri/calcite-design-system/blob/107792d52159848bd6e01fd14fa80292fbfd02e6/packages/calcite-components/src/components/alert/alert.tsx#L111-L112

https://github.com/Esri/calcite-design-system/blob/107792d52159848bd6e01fd14fa80292fbfd02e6/packages/calcite-components/src/components/chip-group/chip-group.tsx#L46-L47

 cc @macandcheese 

### Reproduction Sample

https://codepen.io/geospatialem/pen/poYKVYv

### Reproduction Steps

1. Enable JAWS
2. Open the [sample](https://codepen.io/geospatialem/pen/poYKVYv) in a Chrome browser
3. Observe the example for the `chip-group` works as-expected
    - "Light gray basemap" and "Dark gray basemap" contexts are provided
5. Using the arrow keys navigate to the second and first examples
6. Observe the `label`'s are not provided to AT
  - Expected "Gray basemap" for both examples

### Reproduction Version

`2.4.0`

### Relevant Info

- Windows
- JAWS
- Chrome browser

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

macandcheese:
[...]

PR #10787, opened 2024-11-19
refactor(chip)!: make label property required and value optional

benelan:
**Related Issue:** #8696

## Summary

- Make the `label` property required to ensure it is provided to a11y assistive technologies.
- Make the `value` property optional in favor of `label`.

BREAKING CHANGE: The `label` property is now required and `value` is optional.


jcfranco (approved):
🚀

PR #10795, opened 2024-11-19
refactor(table)!: remove deprecated zebra property

eriklharper:
**Related Issue:** #9713

## Summary

BREAKING CHANGE: Removes the deprecated property `zebra` from `calcite-table`.

Developers will need to replace `calcite-table`'s "zebra" property with "striped".


macandcheese (approved):
🦓 🚫

PR #10797, opened 2024-11-19
refactor(checkbox)!: remove deprecated guid property

eriklharper:
**Related Issue:** #9713

## Summary

This property is no longer needed since the `connectForm` utility takes the place of the previous implementation that required rendering a native `input` element that needed a unique id set with the `guid` utility to enable native form submission.

BREAKING CHANGE: Removes the deprecated `guid` property from `calcite-checkbox`.

Migration plan: Developers may need to replace `calcite-checkbox`'s "guid" property with a unique "id" property.

### DIFF
--- a/packages/calcite-components/src/components/checkbox/checkbox.tsx
+++ b/packages/calcite-components/src/components/checkbox/checkbox.tsx
@@ -84,7 +84,7 @@ export class Checkbox
   /**
    * The `id` attribute of the component. When omitted, a globally unique identifier is used.
    *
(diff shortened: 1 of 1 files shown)


driskull (approved):
👍 code looks good! 🧹

jcfranco:
@eriklharper Could you move the connectForm implementation details to the summary and remove them from the breaking change notes? Developers don’t need to know about that; a simple “It is no longer needed” should suffice.

Issue #10796, opened 2024-11-19
Types should be defined in .ts files


maxpatiiuk:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Calcite has a number of .d.ts files that include public-facing types:
https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/components/handle/interfaces.d.ts

Usually, `.d.ts` files should not be authored by hand - the only case they are authored by hand is when you need to provide types for a library that is missing them - and such types would be used for internal type checking only.

To that end, `tsc` by default does not copy the hand-authored `.d.ts` files to the build output. Lumina follows this behavior. It appears than in Stencil, hand-authored `.d.ts` files were being copied too.

This is causing errors when type-checking JS API as `interfaced.d.ts` files are missing from public typings.

side note: it appears that not all Calcites interfaces files are using .d.ts - some use `.ts` and correctly appar in the build output:
https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/components/accordion-item/interfaces.ts

### Expected Behavior

All files that include public-facing types should use `.ts` files

### Reproduction Sample

`npx tsc --outDir types --emitDeclarationOnly --declaration`

### Reproduction Steps

Run this from `packages/calcite-components`
```sh
npx tsc --outDir types --emitDeclarationOnly --declaration
```

See that hand-authored `.d.ts` files don't appear in the output, but `.ts` files do. Lumina has the same behavior since it's using TypeScript API behind the scenes

### Reproduction Version

3.0.0-next.16

### Priority impact

impact - p1 - need for current milestone

### Impact

Breaks JS API type checking:

```yaml
node_modules/@esri/calcite-components/dist/components/calcite-handle/customElement.d.ts:5:43 - error TS2307: Cannot find module './interfaces' or its corresponding type declarations.

5 import { HandleChange, HandleNudge } from './interfaces';
                                            ~~~~~~~~~~~~~~
node_modules/@esri/calcite-components/dist/components/calcite-input-time-zone/customElement.d.ts:11:43 - error TS2307: Cannot find module './interfaces' or its corresponding type declarations.

11 import { OffsetStyle, TimeZoneMode } from './interfaces';
                                             ~~~~~~~~~~~~~~
[...]

PR #10800, opened 2024-11-20
build: ensure types are specified in TS files

jcfranco:
**Related Issue:** #10796

## Summary

Fixes TS errors caused by having public types defined in `.d.ts` files.


maxpatiiuk (approved):
👍

Issue #10801, opened 2024-11-20
CalciteFlowItem.showBackButton causes type error in public types


maxpatiiuk:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Consuming Calcite's `dev` branch typings in JS API produces this TypeScript error:

```ts
node_modules/@esri/calcite-components/dist/components/calcite-flow/interfaces.d.ts:3:61 - error TS2344: Type '"selected" | "setFocus" | "beforeBack" | "menuOpen" | "showBackButton"' does not satisfy the constraint 'keyof HTMLCalciteFlowItemElement'.
  Type '"showBackButton"' is not assignable to type 'keyof HTMLCalciteFlowItemElement'.

3 export type FlowItemLike = Pick<HTMLCalciteFlowItemElement, "beforeBack" | "menuOpen" | "setFocus" | "showBackButton" | "selected">;
```

The `showBackButton` property is marked as private, which makes it excluded from public typings:

https://github.com/Esri/calcite-design-system/blob/b280ea4903466f1cc22cebf75d8d98dbd35ab318/packages/calcite-components/src/components/flow-item/flow-item.tsx#L133-L138

Solutions:
1. (easy) Mark the `showBackButton` as `@internal` rather than `@private` - it will still be present in the typings, but you can can be excluded when rendering the public documentation.
2. Update the `FlowItemLike` interface to not rely on `showBackButton` (or at least not in the way that produces a typescript error when that property is absent) 

### Expected Behavior

Do not produce TypeScript errors when consuming Calcite typings in a project that has `"skipLibCheck": false`

### Reproduction Sample

(JS API PR 66756)

### Reproduction Steps

Simplest way to reproduce is to comment out the `showBackButton` property from the CalciteFlowItem component to simulate how it would appear in the public typings.

Otherwise, you can try `npx build:ts` on the `calcite-lumina` branch in JS API (PR 66756)

### Reproduction Version

3.0.0-next.12

### Priority impact

impact - p1 - need for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

DitwanP:
🍡 Verified.

PR #10802, opened 2024-11-20
fix(flow-item): fix TS error caused by FlowItemLike type

jcfranco:
**Related Issue:** #10801

## Summary

Restores internal prop that was unintentionally made private by #10482.


jcfranco:
> though need to double check if your docs rendering excludes @internal properties.

I believe the doc site ignores `@private` and `@ignore`, but not `@internal`. @macandcheese I'll submit a PR for this.

Issue #10803, opened 2024-11-20
`.json` file imports appearing in public typings


maxpatiiuk:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Generated types for Calcite components include imports to `.json` files:

```yaml
node_modules/@esri/calcite-components/dist/components/calcite-color-picker-hex-input/customElement.d.ts:7:39 - error TS2732: Cannot find module '../calcite-color-picker/assets/t9n/color-picker.t9n.en.json'. Consider using '--resolveJsonModule' to import module with '.json' extension.

7 import { default as T9nStrings } from '../calcite-color-picker/assets/t9n/color-picker.t9n.en.json';
                                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@esri/calcite-components/dist/components/calcite-date-picker-month-header/customElement.d.ts:3:39 - error TS2732: Cannot find module '../calcite-date-picker/assets/t9n/date-picker.t9n.en.json'. Consider using '--resolveJsonModule' to import module with '.json' extension.

3 import { default as T9nStrings } from '../calcite-date-picker/assets/t9n/date-picker.t9n.en.json';
                                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@esri/calcite-components/dist/components/calcite-date-picker-month/customElement.d.ts:6:39 - error TS2732: Cannot find module '../calcite-date-picker/assets/t9n/date-picker.t9n.en.json'. Consider using '--resolveJsonModule' to import module with '.json' extension.

6 import { default as T9nStrings } from '../calcite-date-picker/assets/t9n/date-picker.t9n.en.json';

node_modules/@esri/calcite-components/dist/components/calcite-input-date-picker/customElement.d.ts:11:47 - error TS2732: Cannot find module '../calcite-date-picker/assets/t9n/date-picker.t9n.en.json'. Consider using '--resolveJsonModule' to import module with '.json' extension.

11 import { default as DatePickerMessages } from '../calcite-date-picker/assets/t9n/date-picker.t9n.en.json';
[...]

PR #10804, opened 2024-11-20
fix(t9n): make TypeScript inline .json type imports

maxpatiiuk:
**Related Issue:** #10803

## Summary

Subtly change the JSON type imports to encourage TypeScript to inline these in the final .d.ts files, rather than preserve them as `.json` imports.

Opened an issue 2047 in AWC to add a build-time error if any .json import appears in the public .d.ts file. We can extend that as needed


jcfranco (approved):
🚀

maxpatiiuk:
For future reference: I found a consistent way to force typescript to inline the type of a type only json import:

```ts
import type T9nStrings from "../assets/t9n/messages.en.json";
import type { identity } from "@arcgis/components-utils";

type Messages = ReturnType<typeof identity<typeof T9nStrings>>;
```

The `ColorPicker["messages"]` trick is simpler, but when it is not possible (because there is no such property available for use), the above is a workaround

PR #10808, opened 2024-11-20
refactor: drop unnecessary `LocalizedComponent` interface and utils

jcfranco:
**Related Issue:** N/A

## Summary

Lumina provides controllers to handle this now, so we don’t need it anymore.

@eriklharper Thanks for bringing this up!


eriklharper (approved):
LGTM 👍

PR #9481, opened 2024-05-31
 test(commonTests): improve `themed` error messages

jcfranco:
✨🧪🧰✨

Test/Assertion failure messages should provide more context now:

```zsh
 FAIL  src/components/action-pad/action-pad.e2e.ts (7.262 s)
  ● calcite-action-pad › theme › default › is themeable

    expect(received).toBe(expected) // Object.is equality

    Expected: "[--calcite-action-pad-trigger-background-color-active:--calcite-action-group-trigger-background-color-active] rgb(0, 191, 255)"
    Received: "[--calcite-action-pad-trigger-background-color-active:--calcite-action-group-trigger-background-color-active] "
```
```zsh
 FAIL  src/components/handle/handle.e2e.ts
  ● calcite-handle › theme › selected › is themeable

    expect(received).toBe(expected) // Object.is equality

    Expected: "[--calcite-handle-background-color-selected:backgroundColor] rgb(0, 191, 255)"
    Received: "[--calcite-handle-background-color-selected:backgroundColor] rgba(0, 0, 0, 0)"
```
```zsh
 FAIL  src/components/action-group/action-group.e2e.ts (5.418 s)
  ● calcite-action-group › theme › is themeable

    [--calcite-action-group-trigger-background-color-hover] target node (calcite-action-group >>> calcite-action) must be clickable (larger than 1x1) for state: hover

      380 |             })`;
```

**Note**: this depends on https://github.com/Esri/calcite-design-system/pull/9450.

PR #10829, opened 2024-11-21
test(themed): simplify the test

alisonailea:
## Summary

We only need to test that the expected value is set somewhere in the computed style no get the exact styleString

### DIFF
--- a/packages/calcite-components/src/tests/commonTests/themed.ts
+++ b/packages/calcite-components/src/tests/commonTests/themed.ts
@@ -374,7 +374,8 @@ async function assertThemedProps(page: E2EPage, options: TestTarget): Promise<vo
   const isLinearGradientUnderlineToken = token.includes("link-underline-color") && targetProp === "backgroundImage";
 
   if (isFakeBorderColorToken || isLinearGradientUnderlineToken) {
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/tests/commonTests/themed.ts
```diff
@@ -374,7 +374,7 @@ async function assertThemedProps(page: E2EPage, options: TestTarget): Promise<vo
   const isLinearGradientUnderlineToken = token.includes("link-underline-color") && targetProp === "backgroundImage";
 
   if (isFakeBorderColorToken || isLinearGradientUnderlineToken) {
-    expect(getStyleString(token, targetProp, styles[targetProp])).toMatch(expectedValue);
```
jcfranco, line 377: We should keep this for testing maintenance. Removing `getStyleString` would remove helpful details from the test failure messages (see https://github.com/Esri/calcite-design-system/pull/9481).

jcfranco (changes requested):

packages/calcite-components/src/tests/commonTests/themed.ts
alisonailea: thanks for the catch I think this was a merge mistake

alisonailea:
Closing because this PR is no longer applicable.

Issue #7284, opened 2023-07-05
[Link] Incorrect underline animation in RTL


macandcheese:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The underline animation does not correctly animate in the expected direction when in RTL.

### Expected Behavior

I'd expect the animation to correctly display as it does in LTR, but flipped.

cc @SkyeSeitz @ashetland - I know we've talked about changing this animation altogether and favoring something less... busy - maybe an additional px height on hover, just color change, etc., so could be a good opportunity to look at that.

Related: https://github.com/Esri/calcite-design-system/issues/6569

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/qBQXYKB?editors=1000

### Reproduction Steps

1. Go to Codepen
2. See difference in animation

### Reproduction Version

Latest

### Priority impact

p2 - want for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react

### Esri team

ArcGIS Map Viewer

geospatialem:
cc @ashetland @SkyeSeitz WDYT/ any thoughts on changes to the `link` animation?

SkyeSeitz:
I could definitely see a few options where the animation is removed. I think I lean towards just an instant color change from 40% opacity to 100%, as seen in [the prototype below](https://www.figma.com/proto/YwTpfq3LjR1UrPX8lQ8ON4/%5BLink%5D-Incorrect-underline-animation-in-RTL-%5Bissue-7284%5D?page-id=2%3A165&type=design&node-id=101-147&viewport=787%2C644%2C3.65&t=JHS3vgF2LWeLz4BU-1&scaling=min-zoom&mode=design).  

https://github.com/Esri/calcite-design-system/assets/29716057/71c8f58b-15d8-4174-88e4-c6a6044413aa

ashetland:
I like that instant change. Mocked up a bump from 1px to 2px for the indicator @macandcheese mentioned just for comparison. I think I go for the simple version or just instant timing.
![CleanShot 2023-08-07 at 12 09 13](https://github.com/Esri/calcite-design-system/assets/108549080/20af8261-0513-41bc-8120-b0d2ed11e348)

SkyeSeitz:
[...]

PR #10835, opened 2024-11-22
fix(link): fix underline animation

jcfranco:
**Related Issue:** #7284

## Summary

Restores the underline animation that was broken in v2.7.0 (https://github.com/Esri/calcite-design-system/pull/8797) and updates it to work consistently in both LTR/RTL directions.

This also adds a `transition-default` mixin to provide finer-grained control of transitions.


macandcheese (approved):
🏆

PR #10836, opened 2024-11-22
fix: fix core transitions across components

jcfranco:
**Related Issue:** #10837

## Summary

Splits up props set by the `transition-default` Tailwind util, so they are all configured correctly.

This came up in https://github.com/Esri/calcite-design-system/pull/10835. It turns out https://github.com/Esri/calcite-design-system/pull/10835 wasn't set up properly and core transitions were not working.

**Note**: Removed `outline` and `outline-offset` as these were configured with defaults and are not animated by default.


macandcheese (approved):
🏆

PR #10838, opened 2024-11-22
chore: remove stylelint-updating pre-commit script

jcfranco:
**Related Issue:** N/A

## Summary

There have been instances where this script fails to update [`customFunctions`](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/.stylelintrc.cjs#L4) properly. At times, it either clears the list or adds unnecessary functions. To address this, we’ll remove the script for now until it can be stabilized and restored.

cc @anveshmekala @macandcheese

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,18 +1,5 @@
 #!/usr/bin/env sh
 
-update_stylelint_config_if_sass_file_edited() {
-  staged_files="$(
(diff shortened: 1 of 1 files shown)


anveshmekala (approved):
No changes to [customFunctions](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/.stylelintrc.cjs#L4) in stylelint config file reported during commit. 
🏆

Issue #10841, opened 2024-11-22
Popover reopens when trigger is clicked and `autoClose = true`


mpayson:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Popovers closes then re-opens when the trigger is clicked and `autoClose=true`

### Expected Behavior

Popovers close when the trigger is clicked

### Reproduction Sample

https://github.com/mpayson/calcite-debug/tree/popover-reopen

### Reproduction Steps

1. Clone the repository
2. `npm install`
3. `npm run dev`
4. Open the page
5. Click the button -> see popover opens
6. Click the button again -> see popover closes then reopens

### Reproduction Version

2.14.0-next.20

### Regression?

2.13

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [X] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Data Pipelines

geospatialem:
Also confirmed via Codepen between `2.13.2` and at the latest `next`: `3.0.0-next.23`: https://codepen.io/geospatialem/pen/WNVVZMZ

geospatialem:
Verified in `3.0.0-next.27` with https://codepen.io/geospatialem/pen/WNVVZMZ.

PR #10842, opened 2024-11-22
fix(Popover): Popover reopens when trigger is clicked and autoClose = true

driskull:
**Related Issue:** #10841


jcfranco (approved):
LGTM! 

Can we add a test for this? We can do this in a follow-up issue if needed. cc @geospatialem

driskull:
> We can do this in a follow-up issue if needed.

Yes follow up plz. Can you assign to Eliza?

Issue #9173, opened 2024-04-23
[List]: add mode prop for choosing between flat and nested lists


ashetland:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

A way to control nesting behavior in List. Nested list items need to hold space for the expand button, but this space is undesirable in lists that, base on workflow, will never have nested items.

This would be a breaking change and should be paired with #7100.

### Acceptance Criteria

Add a `mode="flat" (default) | "nested"` property.

In `flat`, Lists are rendered flat, without collapsible nesting. If an item is nested in the DOM, it is rendered, but not indented.

In `nested`, a blank space is reserved for the expand button in every item.

Before/after example:
![CleanShot 2024-11-20 at 13 02 04@2x](https://github.com/user-attachments/assets/cd11963a-9828-4843-942c-d57fbf45aa1f)

### Relevant Info

cc @driskull @jcfranco 

### Which Component

List
List Item

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

ashetland:
Updated the description to include a before/after screenshot using the current build of List. Spacing will be improved with #7100.

DitwanP:
🍡 Verified locally on `dev`

geospatialem:
For documentation purposes later, here's a codepen sample in `3.0.0-next.29`: https://codepen.io/geospatialem/pen/MYggPPV. cc @DitwanP 

```html
<calcite-list display-mode="nested">
```

PR #10852, opened 2024-11-23
feat(list)!: add displayMode property to choose between flat and nested lists

driskull:
**Related Issue:** #9173

## Summary

- add `displayMode` property to choose between flat and nested lists
- update demos
- update stories
- add tests

BREAKING CHANGE: Choose a `displayMode` of "nested" or "flat" according to space and nesting requirements.

### DIFF
--- a/packages/calcite-components/src/components/list/list.e2e.ts
+++ b/packages/calcite-components/src/components/list/list.e2e.ts
@@ -975,7 +975,7 @@ describe("calcite-list", () => {
     it("should navigate via ArrowRight and ArrowLeft", async () => {
       const page = await newE2EPage();
       await page.setContent(html`
-        <calcite-list>
+        <calcite-list mode="nested">
           <calcite-list-item id="one" value="one" label="One" description="hello world">
             <calcite-action
               appearance="transparent"
@@ -1039,7 +1039,7 @@ describe("calcite-list", () => {
     it("should navigate a draggable list via ArrowRight and ArrowLeft", async () => {
       const page = await newE2EPage();
       await page.setContent(html`
-        <calcite-list drag-enabled>
+        <calcite-list mode="nested" drag-enabled>
           <calcite-list-item id="one" value="one" label="One" description="hello world">
             <calcite-action
               appearance="transparent"
--- a/packages/calcite-components/src/components/list/list.e2e.ts
+++ b/packages/calcite-components/src/components/list/list.e2e.ts
@@ -159,153 +159,259 @@ describe("calcite-list", () => {
       </calcite-list>`,
       { focusTarget: "child" },
     );
+  });
 
-    it("should set the dragHandle property on items", async () => {
-      const page = await newE2EPage();
-      await page.setContent(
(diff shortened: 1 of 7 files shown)


driskull:
@ashetland could you give this a view? 👀

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -239,6 +241,9 @@ export class List
    */
   messages = useT9n<typeof T9nStrings>({ blocking: true });
 
+  /** Specifies the nesting behavior. */
+  @property({ reflect: true }) mode: ListMode = "flat";
```
macandcheese, line 245: Should this be `displayMode` to match some other component nomenclature? This doesn't affect any selection or other functionality right, just visual?

driskull: I'll defer to @jcfranco

ashetland: It does affect some functionality. `nested` must be used to add the expand button and indent children. That said, I could go either way here and also defer to @jcfranco.

ashetland (approved):
Visuals lookin' good!

packages/calcite-components/src/components/list/list.e2e.ts
```diff
@@ -136,153 +159,259 @@ describe("calcite-list", () => {
...
-              <calcite-list-item open label="Depth 2" description="Item 2">
-                <calcite-list drag-enabled group="my-list">
-                  <calcite-list-item label="Depth 3" description="Item 3">
-                    <calcite-list drag-enabled group="my-list"></calcite-list>
-                  </calcite-list-item>
-                  <calcite-list-item label="Depth 3" description="Item 4"></calcite-list-item>
-                </calcite-list>
-              </calcite-list-item>
-              <calcite-list-item label="Depth 2" description="Item 5"></calcite-list-item>
+              <div>
```
jcfranco, line 221: What's the use case for wrapped list items? If it's not a strong one, I'd suggest not supporting this to keep things simple. This also aligns with the default slot's description where we only expect items and groups. Also applies to similar `dragHandle` test.

packages/calcite-components/src/components/list/list.tsx
jcfranco: Good point. Let's go with `displayMode` for consistency. The only other `mode` prop I found was `input-time-zone`'s, which affects both the API and presentation.

jcfranco (approved):
Awesome stuff, @driskull! ✨📝✨

I have a concern regarding support for unexpected, slotted items, but aside from that, this LGTM!

packages/calcite-components/src/components/list/list.e2e.ts
driskull: This is basically a copy from an existing test for dragHandle. I can remove the test though.

jcfranco: Yeah, let's remove.

PR #10853, opened 2024-11-23
feat(list, list-item, list-item-group, filter)!: add scales, update padding and spacing, update font sizes

eriklharper:
**Related Issue:** #7100 

## Summary

This PR adds scales, updates padding and spacing and updates font sizes based on the latest designs for `calcite-list-item`, `calcite-list-item-group`, `calcite-list` and `calcite-filter`.

BREAKING CHANGE:  The spacing changes affect content in the `actions-start` slot in addition to other rendered elements.  Some styling updates may need to be applied to slotted actions or other adjustments based on the new scales.


packages/calcite-components/src/components/list/list.tsx
```diff
@@ -254,6 +256,9 @@ export class List
    */
   @property() openable = false;
 
+  /** Specifies the size of the component. */
+  @property() scale: Scale = "m";
```
jcfranco, line 260: For consistency, `scale` should be reflected.

packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -767,6 +776,8 @@ export class ListItem
           : ICONS.closedLTR
       : ICONS.blank;
 
+    const iconScale = scale === "l" ? "m" : "s";
```
jcfranco, line 779: Can you use the `getIconScale` util for this?

packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -130,7 +131,7 @@ export class ListItem
    *
    * @private
    */
-  @property() dragHandle = false;
+  @property({ reflect: true }) dragHandle = false;
```
jcfranco, line 134: Nice catch on these. Was the lack of reflection causing E2E or screenshot test failures? If not, can you restore these to keep your PR focused. We'll restore these separately.

packages/calcite-components/src/components/list-item-group/list-item-group.tsx
```diff
@@ -36,6 +37,9 @@ export class ListItemGroup extends LitElement implements InteractiveComponent {
   /** The header text for all nested `calcite-list-item` rows. */
   @property({ reflect: true }) heading: string;
 
+  /** Specifies the size of the component. */
+  @property({ reflect: true }) scale: Scale = "m";
```
jcfranco, line 41: Can we make these internal and have `list` set them directly?

jcfranco (approved):
Aside from a few comments, this is looking great! I'll defer to @ashetland for the visual review.

@eriklharper LMK if this needs another review.

packages/calcite-components/src/components/list/list.tsx
eriklharper: I made a judgment call on that because its not actually used in the css for list, but cool to add it anyway...

packages/calcite-components/src/components/list-item/list-item.tsx
eriklharper: No test failures that I saw.

packages/calcite-components/src/components/list-item-group/list-item-group.tsx
driskull: For internal properties, should we try to not reflect them? It more easily exposes them to the user and lets them see internal attr values. Maybe this is a discussion we should have @jcfranco? At least when they are not reflected a user isn't seeing them in the DOM tree.

packages/calcite-components/src/components/list-item/list-item.e2e.ts
```diff
@@ -474,7 +474,7 @@ describe("calcite-list-item", () => {
         {
           "--calcite-list-selection-border-color": {
             shadowSelector: `.${CSS.container}`,
-            targetProp: "borderInlineStartColor",
+            targetProp: "boxShadow",
```
driskull, line 477: Is it weird to be using a box shadow for a border var? Are we breaking any trust with an end user by saying the var is for something but using it for something else? cc @ashetland @alisonailea

packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -167,7 +168,7 @@ export class ListItem
    *
    * @private
```
driskull, line 169: Sometimes we use `@private` and sometimes `@internal`, is there a preferred one or two distinct use cases? @jcfranco

packages/calcite-components/src/components/list-item/list-item.e2e.ts
ashetland: In this case we opted to use an inner shadow for the selection border styling to maintain alignment with item-group headings, Panel headings, etc. I believe we did something similar for Tile. Defer the @alisonailea for var naming.

driskull:
@eriklharper I also noticed if you change the scale on the list it isn't updating the child items. There should be something in `willUpdate` to handle changes to `scale`. cc @jcfranco. Can we get an issue @geospatialem? We should get this fixed for beta testing.

packages/calcite-components/src/components/list-item/list-item.e2e.ts
eriklharper: We do this in Tile, and based on discussions I've had with others on the team, this is an acceptable use case since `box-shadow` doesn't encroach on the dimensions of the box, which allows other sizing properties to not have to accommodate whether the border is there or not.

packages/calcite-components/src/components/list-item/list-item.tsx
eriklharper: See the Lumina docs here for the differences between `@internal` and `@private`: https://qawebgis.esri.com/arcgis-components/lumina/transition-from-stencil#internal-members

eriklharper:
> @eriklharper I also noticed if you change the scale on the list it isn't updating the child items. There should be something in `willUpdate` to handle changes to `scale`. cc @jcfranco. Can we get an issue @geospatialem? We should get this fixed for beta testing.

Good call, this is something that should be supported.  I can get on that issue today, it should be just a matter of setting it up to watch the changes on `scale` in `willUpdate`.

packages/calcite-components/src/components/list-item-group/list-item-group.tsx
eriklharper: I opted to use attribute selectors for the styling to get this done as fast as possible, so if we feel strongly enough about not reflecting internal attributes we would have to refactor the styling to use css classes instead.

eriklharper: Personally I don't have qualms about reflecting internal attributes as long as they're meant to be reflected for styling purposes and also are decorated with `@internal ` or `@private`.

Issue #10321, opened 2024-09-16
combobox-item: label vs heading vs textLabel mix-up in code and documentation


abp6318:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

`textLabel` is now deprecated, and `label` and `heading` are the new go-tos, but removing `textLabel` from the combobox-items causes the combobox to not render correctly.

![bug](https://github.com/user-attachments/assets/3d0719b5-127d-4531-adca-f1457ee16005)

**ALSO**

The combobox-item documentation is a bit funny looking, where the `textLabel` is both deprecated and required
<img width="1041" alt="Screenshot 2024-09-16 at 1 32 49 PM" src="https://github.com/user-attachments/assets/645f9640-39c5-4a5b-85c4-6c68548621a0">

### Expected Behavior

`label` and `heading` are sufficient for the combobox to render the items and chips correctly.

### Reproduction Sample

https://codepen.io/AaronPutterman/pen/JjQqgyW?editors=1000

### Reproduction Steps

1. Create a combobox with combobox-items
2. Provide _only_ a `label` and `heading` attribute to the combobox-items
3. Render the combobox and select any of the items

### Reproduction Version

v2.12.2

### Relevant Info

Chrome Version 128.0.6613.138 (Official Build) (arm64), MacOS, 16-inch 2023, Apple M2 Pro

### Regression?

v2.8 was the last version we used

### Priority impact

impact - p2 - want for an upcoming milestone

### Impact

More confusing than anything, and we can use `textLabel` in the mean-time

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Hub

geospatialem:
We should also include a linting rule to ensure deprecated properties are not also required as part of the above.

doreenbrinkman:
Another vote for this one.  Can not remove deprecated `textLabel` property and use `heading` for Data Pipelines.

geospatialem:
An additional tweak is needed for the `heading` attribute in the parent `combobox`, so when the child `combobox-item` is selected its value is provided via a chip when selected. 

For future testing/verification, we can use https://codepen.io/geospatialem/pen/RNbbYYq. 🧪

[...]

PR #10855, opened 2024-11-23
fix(combobox-item): replace type-enforced deprecated + required `textLabel` prop with runtime warning

jcfranco:
**Related Issue:** #10321

## Summary

Addressed usage issues caused by TypeScript-enforced required property that had been marked as deprecated.

### Noteworthy changes:

* adds `warnIfMissingRequiredProp` util for cases where TS-enforced required props are not suitable
* enable [`jsdoc/no-restricted-syntax`](https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/no-restricted-syntax.md) rule to catch props tagged as both required and deprecated

### DIFF
--- a/packages/calcite-components/.eslintrc.cjs
+++ b/packages/calcite-components/.eslintrc.cjs
@@ -73,6 +73,18 @@ module.exports = {
     "jest/expect-expect": "off",
     "jest/no-export": "warn",
     "jsdoc/check-tag-names": "off",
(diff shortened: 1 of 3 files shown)


jcfranco:
@anveshmekala We could tweak the new `warnIfMissingRequiredProp ` util to help w/ https://github.com/Esri/calcite-design-system/issues/8697.

driskull (approved):
👍

Issue #10762, opened 2024-11-18
split accordion-item header and content padding


alisonailea:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

https://github.com/Esri/calcite-design-system/pull/10721#pullrequestreview-2437271640

The current design pattern for accordion item is to have content and heading padding size in unison.

### Acceptance Criteria

The proposed design pattern is accordion item heading and content padding size separately.

### Which Component

accordion-item

### Example Use Case

As a developer I want the padding of the accordion-item heading to be 10px and the content padding to be 0px.

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [X] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

geospatialem:
@alisonailea Can you update the labels for verification? I've added in the labels and express triaged the above to November/current milestone.

DitwanP:
@alisonailea I think the padding token is still applying to both heading and content, but maybe I'm missing something.

Accordion demo page with `calcite-accordion-item { --calcite-accordion-item-content-space: 40px; }`:
https://github.com/user-attachments/assets/9aaea4fd-2668-42c1-8ac1-97826d5355bb

geospatialem:
@alisonailea It looks like the PR has been merged, is this issue ready for verification? If so, can you add the `3 - installed` label for us? 🤜🏻 🤛🏻

DitwanP:
Verified locally on `dev` the `--calcite-accordion-item-content-space` token is now only applying to the content as intended.

PR #10727, opened 2024-11-14
fix(tokens)!: correct space and size tokens

alisonailea:
**Related Issue:** #10012 #10014

## Summary

Corrects space and size tokens to align with the Calcite UI Kit and uses REM based tokens for the web platform token output. Updates Calcite Components to ensure there are no visual breaking changes to components using the updates size and space tokens.

BREAKING CHANGE: Size and Space token values have changed. Line Height tokens exported for JS and ES6 are now unitless.

### Size

calcite-size-px: 1px;
calcite-size-xxxs: 0.75rem; // 12px was 2px
calcite-size-xxs: 0.875rem; // 14px was 4px
calcite-size-xs: 1rem; // 16px was 6px
calcite-size-sm: 1.5rem; // 24px was 8px
calcite-size-sm-plus; // removed
calcite-size-md: 2rem; // 32px was 12px
calcite-size-md-plus; // removed
calcite-size-lg: 2.75rem; // 44px was 16px
calcite-size-xl: 3rem; // 48px was 20px
calcite-size-xxl: 4rem; //  64px was 24px
calcite-size-xxxl: 6rem;  // 96px was 32px

### Space

calcite-spacing-none: 0; // new
calcite-spacing-px: 1px;
calcite-spacing-base: 2px;
calcite-spacing-xxs: 0.25rem; // 4px
calcite-spacing-xs: 0.375rem; // 6px
calcite-spacing-sm: 0.5rem; // 8px
calcite-spacing-sm-plus: 0.625rem; // new
calcite-spacing-md: 0.75rem; // 12px
calcite-spacing-md-plus: 0.875rem; // new
calcite-spacing-lg: 1rem; // 16px was 14px
calcite-spacing-xl: 1.25rem; // 20px was 16px
calcite-spacing-xxl: 1.5rem; // 24px was 20px
calcite-spacing-xxxl: 2rem; // 32px

### Pixel based tokens are now categorized under "fixed"

calcite-size-fixed-xxxl: 32px; // deprecated
calcite-size-fixed-xxl: 24px; // deprecated
calcite-size-fixed-xl: 20px; // deprecated
calcite-size-fixed-lg: 16px; // deprecated
calcite-size-fixed-md: 12px; // deprecated
calcite-size-fixed-sm: 8px; // deprecated
calcite-size-fixed-xs: 6px; // deprecated
calcite-size-fixed-xxs: 4px; // deprecated
calcite-size-fixed-xxxs: 2px; // deprecated

[...]

PR #10865, opened 2024-11-25
fix(accordion-item): split header content padding

alisonailea:
**Related Issue:** #10762

## Summary

Splits header and content padding rules so the token doesn't apply to header

### DIFF
--- a/packages/calcite-components/src/components/accordion-item/accordion-item.scss
+++ b/packages/calcite-components/src/components/accordion-item/accordion-item.scss
@@ -75,7 +75,13 @@
   );
 }
 
-.header-content,
+.header-content {
+  padding: var(
+    --calcite-internal-accordion-item-padding,
+    var(--calcite-internal-accordion-item-spacing-unit, theme("spacing.2") 0.75rem)
+  );
+}
+
 .content {
   padding: var(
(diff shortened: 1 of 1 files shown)


benelan:
Do we have any conventions for what constitutes a breaking change to the design tokens? Because changes to the json paths could break someone's build script if they are accessing token values for use in their application.

alisonailea:
> Do we have any conventions for what constitutes a breaking change to the design tokens? Because changes to the json paths could break someone's build script if they are accessing token values for use in their application.

This PR should not have included the space and size token updates which were covered in a separate PR.

benelan:
> > Do we have any conventions for what constitutes a breaking change to the design tokens? Because changes to the json paths could break someone's build script if they are accessing token values for use in their application.
> 
> This PR should not have included the space and size token updates which were covered in a separate PR.

Gotcha was that #10727?

Do you think we should add snapshot or test coverage for this fix?

benelan (approved):
> Do you think we should add snapshot or test coverage for this fix?

Nevermind, it looks like the existing Custom Theme story picked up this change. Code LGTM, but I'll defer to design for the chromatic approval. cc @ashetland

ashetland (approved):
LGTM

Issue #10860, opened 2024-11-24
Dialog emits unexpected close event when removed in 2.14, uncaught error in 3.0


mpayson:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

In 2.14, Calcite dialog emits a close event when it is removed from the DOM. It should not emit an event in this case because the app already knows it is closing the dialog, and it causes a complication at least in our app (see impact)

In 3.0, an uncaught type error is thrown in the console but an event is not emitted: `Uncaught (in promise) TypeError: this.transitionEl is undefined`

### Expected Behavior

Calcite dialog does not emit an event when it is removed from the DOM, like in 2.13

### Reproduction Sample

https://codepen.io/mpayson/pen/raBBxgG

### Reproduction Steps

1. Open the code pen and developer tools
2. Wait 3 seconds, see dialog is removed and a close event is logged
  a. Note, you may need to load the codepen twice -- the first time the CDN assets can take longer than 3 seconds 
3. Repeat with 3.x build, see type error

### Reproduction Version

2.14

### Regression?

2.13

### Priority impact

impact - p1 - need for current milestone

### Impact

Our app has a guided tour UX that transitions from a modal to a popover. When this transition happens and the modal is removed, our app receives a close event, which makes it think the user closed the guided tour, and the tour unexpectedly stops

I can't think of a good work around for this. It looks like the event gets emitted after the animation completes and the dialog is removed, which makes the behavior timing-dependent. I think we would have to set a flag when the transition starts, then block the close event from closing the tour for approximately the time the animation takes to complete. Would appreciate other work around ideas though!

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Data Pipelines

mpayson:
Maybe related? https://github.com/Esri/calcite-design-system/issues/10666

geospatialem:
Verified in `3.0.0-next.31`

PR #10868, opened 2024-11-25
fix(dialog): fix error when initially opened

driskull:
**Related Issue:** #10860

## Summary

- fix error when dialog is opened initially.


jcfranco (approved):
We might need to tweak `willUpdate` in addition or as an alternative. We also need automated testing for this. I'll create a follow-up issue for this.

jcfranco:
Created https://github.com/Esri/calcite-design-system/issues/10885.

Issue #6825, opened 2023-04-21
[Chip] Add opt-in affordance for closing `closable` chip with keyboard.


macandcheese:
### Description

As suggested in Chip Group PR, consider having an additional and opt-in (or opt-out?) functionality to allow users to use a key (probably `delete`) to close a `closable` chip while focused on the chip. Currently a user must navigate to the close button to take this action via keyboard.

### Acceptance Criteria

Chip has a new property - `allow/disallow-close-via-key` (with a better name) to enable the described functionality.

### Which Component

Chip cc @jcfranco 

### Priority impact

p4 - not time sensitive

### Esri team

Calcite (design)

geospatialem:
Verified with `3.0.0-next.36` using the new `close-on-delete` attribute using the `del` and `backspace` keys:

```html
<calcite-chip value="dark-grey" label="Dark grey basemap" icon="basemap" closable close-on-delete></calcite-chip>
```

When `close-on-delete` is not present on the component, no action occurs when using the `del` and/or `backspace` keys. 🎉

PR #10877, opened 2024-11-26
feat(chip): add `closeOnDelete` prop

jcfranco:
**Related Issue:** #6825

## Summary

This prop will allow developers to configure whether a chip will close on `Delete` when focused.

### DIFF
--- a/packages/calcite-components/src/components/chip/chip.e2e.ts
+++ b/packages/calcite-components/src/components/chip/chip.e2e.ts
@@ -67,19 +67,6 @@ describe("calcite-chip", () => {
     expect(eventSpy).toHaveReceivedEvent();
   });
 
-  it("should emit event after the close button is clicked", async () => {
-    const page = await newE2EPage();
-    await page.setContent(`<calcite-chip closable>cheetos</calcite-chip>`);
-
-    const eventSpy = await page.spyOnEvent("calciteChipClose", "window");
-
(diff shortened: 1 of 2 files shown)


macandcheese (approved):
Looks good! Couple q:

- Could you update the title from `closeOnEscape`?
- This works when both the entire chip is focused and when the close button itself is focused. Just confirming we want this to happen in both cases (I think we do)
- Should this also include backspace support? If so, do we need a different name (closeOnKeyboard... etc)

jcfranco:
@macandcheese Excellent questions:

1. Fixed! My fingers keep typing `Escape` despite many edits for some reason. 🤣
2. I thought about ignoring Delete when the close button is focused, but it seems more consistent and straightforward to close regardless.
3. Solid point. Looking at our current `Escape` usage, aligning for consistency makes sense. I’m not sure what the prop should be named in this case. Open to suggestions here. 😅

macandcheese:
> I’m not sure what the prop should be named in this case. Open to suggestions here. 😅
Yeah - tough one to name. `closeOnKeyboard` with specific description could work. Or even leaving `closeOnDelete` with a description that includes "backspace and delete" keys could work. 

Some keyboards (mainly ones marketed for Mac) have "delete" written on the backspace key anyway. For some laptops there is no "delete" standalone anyway. All that to say - maybe `closeOnDelete` is ok lol

jcfranco:
Proceeding with `closeOnDelete`!

PR #10879, opened 2024-11-26
fix(combobox): use heading as fallback for UI labels

jcfranco:


jcfranco:
**Note**: This entry should be omitted from the changelog as it resolves an issue introduced during the milestone.

aPreciado88 (approved):
🚀

PR #10887, opened 2024-11-26
chore(block): add warning message when both heading & label are empty

anveshmekala:
**Related Issue:** #8697 

## Summary

Adds warning message in console when both `heading` & `label` properties are set to empty.


packages/calcite-components/src/components/block/block.tsx
```diff
@@ -352,6 +358,14 @@ export class Block
     );
   }
 
+  private handleWarningMessage(): void {
+    if (!this.heading && !this.label) {
+      console.warn(
```
jcfranco, line 363: For consistency, can you use the `logger` for this?

packages/calcite-components/src/components/block/block.tsx
```diff
@@ -206,6 +208,10 @@ export class Block
     if (changes.has("open") && (this.hasUpdated || this.open !== false)) {
       onToggleOpenCloseComponent(this);
     }
+
+    if (changes.has("heading") || changes.has("label")) {
```
jcfranco, line 212: Let's drop this; the load-time warning should suffice.

packages/calcite-components/src/components/block/block.tsx
```diff
@@ -352,6 +358,14 @@ export class Block
     );
   }
 
+  private handleWarningMessage(): void {
+    if (!this.heading && !this.label) {
```
jcfranco, line 362: Can you create a refactor issue for me to update the `warnIfMissingRequiredProp` util to cover this case?

jcfranco (approved):
Aside from a few comments, this LGTM! 🌮🦃🦖

packages/calcite-components/src/utils/component.ts
```diff
@@ -26,3 +26,28 @@ export function warnIfMissingRequiredProp<C extends LitElement>(
...
+ *
+ * @param component the component with the missing property
+ * @param prop the property that is missing
+ * @param message the custom message about functionality impact
+ * @param prop2 the second property that is missing
+ */
+export function warnIfMissingProps<C extends LitElement>(
+  component: C,
+  prop: keyof C,
+  message?: string,
```
jcfranco, line 43: Let's remove the message argument. Messaging for required props should be consistent. If we need to provide different messaging, we could look at introducing a `context` argument to help with this. 

I still think we can tweak, and rename, `warnIfMissingRequiredProp` to achieve this.

anveshmekala: The props in question are not `required` properties. On the same note, created a different method to handle use case where missing props can affect functionality but not marked as `required`.

jcfranco: I’m aware they’re not required, which is why I’m suggesting tweaking the existing utility (including renaming it). Both approaches aim to achieve the same goal: consistently warning developers about missing props that can’t be enforced via types.

I'm fine to proceed with the current approach, but let's revisit enhancing the util to handle this case.

PR #10888, opened 2024-11-26
fix(chip): provide label context to assistive technologies

geospatialem:
**Related Issue:** #8696 

## Summary
Provides context to the component's `label` property for JAWS and NVDA users using the [`"img"` role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/img_role), conveying content is delivered in a visual manner and should be accompanied with the required `label` property.

### DIFF
--- a/packages/calcite-components/src/components/chip/chip.e2e.ts
+++ b/packages/calcite-components/src/components/chip/chip.e2e.ts
@@ -13,8 +13,8 @@ describe("calcite-chip", () => {
     hidden("calcite-chip");
   });
 
-  describe("accessible without calcite-label", () => {
-    accessible(`<calcite-chip>doritos</calcite-chip>`);
+  describe("accessible with icon only", () => {
+    accessible(`<calcite-chip label="Gray basemap" icon="basemap"></calcite-chip>`);
   });
 
   describe("slots", () => {
--- a/packages/calcite-components/src/components/chip/chip.stories.ts
+++ b/packages/calcite-components/src/components/chip/chip.stories.ts
@@ -7,7 +7,7 @@ import { Chip } from "./chip";
 
 const { scale, appearance, kind } = ATTRIBUTES;
 
-type ChipStoryArgs = Pick<Chip, "scale" | "appearance" | "kind" | "closable" | "selected">;
+type ChipStoryArgs = Pick<Chip, "scale" | "appearance" | "kind" | "closable" | "selected" | "label">;
 
 export default {
   title: "Components/Chip",
@@ -17,6 +17,7 @@ export default {
     kind: kind.values[4],
     closable: false,
     selected: false,
+    label: "My great chip",
   },
   argTypes: {
     scale: {
@@ -33,6 +34,9 @@ export default {
       ),
       control: { type: "select" },
     },
+    label: {
+      control: { type: "text" },
+    },
   },
 };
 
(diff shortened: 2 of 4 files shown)


packages/calcite-components/src/components/chip/chip.tsx
```diff
@@ -357,7 +357,7 @@ export class Chip extends LitElement implements InteractiveComponent, LoadableCo
             : undefined;
     return (
       <InteractiveContainer disabled={disabled}>
-        <div
+        <span
```
jcfranco, line 360: Is this intentional? If so, how does it address the AT issue? Both `div` and `span` are non-semantic, so I’m curious how this influences how our components are interpreted.

geospatialem: It is, yes, there aren't a ton of resources out in the wild, but ultimately folks use them interchangeably to provide support to assistive technologies. If specified as a `div` there is no support for the `label` and the component is read as `blank`, but when altered to `span` both NVDA and JAWS.

However, that said this fix does not provide context of the `label` when a user adds slotted text, so maybe a more robust fix is needed to override the slotted content?

For instance, in the example below with/without the fix `"Basemap"` is conveyed to assistive technologies, instead I'd expect the `label` or `"Gray basemap"` to be conveyed:

```html
<calcite-chip value="light-grey" label="Gray basemap">Basemap</calcite-chip>
```

geospatialem: Found an example by Mozilla using `role="img"`, which seems very similar for our use case: https://codepen.io/svinkle/pen/oYjoNE.

As a result, we can:
- Move back to `div`
- Solve AT support for JAWS and NVDA
- When `label` is specified will override the slotted text

Updated commit coming shortly 👍🏻

packages/calcite-components/src/components/chip/chip.e2e.ts
```diff
@@ -13,8 +13,8 @@ describe("calcite-chip", () => {
     hidden("calcite-chip");
   });
 
-  describe("accessible without calcite-label", () => {
-    accessible(`<calcite-chip>doritos</calcite-chip>`);
+  describe("accessible with icon only", () => {
```
jcfranco, line 16: ✨🏆✨

packages/calcite-components/src/components/chip/chip.stories.ts
```diff
@@ -42,6 +42,7 @@ export const simple = (args: ChipStoryArgs): string => html`
       scale="${args.scale}"
       appearance="${args.appearance}"
       kind="${args.kind}"
+      label="My great chip"
```
jcfranco, line 45: Can you provide a label arg for this story? You'll need to update the [`ChipStoryArg` type](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/components/chip/chip.stories.ts#L10) and add an entry for the [default value](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/components/chip/chip.stories.ts#L14-L19). This can be done as a follow-up to proceed with the merge.

jcfranco (approved):
Noice! 🚀

packages/calcite-components/src/components/chip/chip.stories.ts
geospatialem: Added via https://github.com/Esri/calcite-design-system/pull/10888/commits/d43e01f93822a8a16e6271c6b29d03a8b97e2955, does this look accurate, @jcfranco? 👀

Issue #8463, opened 2023-12-20
calcite-color-picker: should be responsive to container sizes


AdelheidF:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

If color-picker is inside a panel that scrolls vertically then horizontal scrollbars appear in FF. Chrome is fine.

<img src=https://github.com/Esri/calcite-design-system/assets/44481292/9d8b87d5-3fc3-4b22-a7ba-e8518626340b width=300/>

If you set `display:flex` on the color-picker and `overflow:hidden` on its container inside the shadow-dom then it's correct.

cc @driskull 

### Expected Behavior

No horizontal scroll bars in any browser.

### Reproduction Sample

https://codepen.io/afreitag/pen/KKYgKzW
https://codepen.io/afreitag/pen/OJqLWEB

### Reproduction Steps

Open codepen in Chrome and Firefox. You'll see horizontal scrollbars in FF.

### Reproduction Version

2.1.0

### Relevant Info

Firefox

### Regression?

no

### Priority impact

p1 - need for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Map Viewer

driskull:
note: Seems like the color-picker component should be more flexible in its sizing. It should respond to its container and not overflow its boundaries.

geospatialem:
Design expertise sought on the color controls for the container sizing and/if design tokens could mitigate.

jcfranco:
It looks like it is a separate layout issue. I managed to reproduce the issue by swapping the `color-picker` with a `div` with a defined width:

![Image](https://github.com/user-attachments/assets/d7c8a355-ec74-46a4-b696-96e7f541f03f)

DitwanP:
@jcfranco So is there anything else for us to do with this issue? 

I'm still seeing the horizontal scrollbars on Safari and FF using the provided repro cases and on `next.128`

jcfranco:
The scrollbars are unrelated to the color picker, so no further action needed here.

DitwanP:
🍡 Verified

PR #10904, opened 2024-11-27
feat(color-picker): enable responsive layout

jcfranco:
**Related Issue:** #8463

## Summary

This updates `color-picker` to adjust its layout for custom width values larger than the default width for each scale.

### Notes

* The component will initialize with min-widths corresponding to scale, but can be overridden to grow by using `inline-size`
* The color field will preserve its aspect ratio
* Extracts common `Dimensions` interface


aPreciado88 (approved):
🚀

ashetland (approved):
Visuals looking good!

packages/calcite-components/src/components/color-picker/color-picker.tsx
```diff
@@ -377,14 +294,19 @@ export class ColorPicker extends LitElement implements InteractiveComponent, Loa
     this.listen("keyup", this.handleChannelKeyUpOrDown, { capture: true });
   }
 
+  connectedCallback(): void {
+    if (this.hasUpdated) {
```
driskull, line 298: Could this check be moved inside `resizeCanvas`?

packages/calcite-components/src/components/color-picker/utils.ts
```diff
@@ -261,17 +261,26 @@ export function toNonAlphaMode(mode: SupportedMode): SupportedMode {
...
-  }
+export function getSliderWidth(
+  availableWidth: number,
+  activeStaticDimensions: (typeof STATIC_DIMENSIONS)[Scale],
+  hasAlpha: boolean,
+): number {
+  const previewWidth = hasAlpha ? STATIC_DIMENSIONS["l"].preview.size : activeStaticDimensions.preview.size;
+  const effectiveWidth = availableWidth - inlineSizeBorderTotalWidth;
+
+  return Math.max(effectiveWidth - activeStaticDimensions.gap * 3 - previewWidth, 0);
```
driskull, line 275: What is the `3` for? Just curious.

driskull (approved):
Looks good! 👍 💯

packages/calcite-components/src/components/color-picker/utils.ts
jcfranco: It's the spacing between elements around the slider.

Screenshot 2025-01-28 at 1 18 53 PM

I should avoid the magic number though. WDYT about this?

```tsx
const inlineSpaceAroundElements = activeStaticDimensions.gap * 3;
```

packages/calcite-components/src/components/color-picker/color-picker.scss
```diff
@@ -7,9 +7,8 @@
 :host([scale="s"]) {
   --calcite-color-picker-spacing: 8px;
 
-  .container {
-    inline-size: 198px;
-  }
+  inline-size: 200px;
+  min-inline-size: 200px;
```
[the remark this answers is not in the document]
jcfranco: If you mean a custom CSS prop, it might be tricky since I also need static dimensions for canvas drawing.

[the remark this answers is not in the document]
jcfranco: I should DRY this up though.

packages/calcite-components/src/components/color-picker/color-picker.tsx
jcfranco: I think so. I might be able to remove the `observeResize` in `firstUpdated` this way too.

driskull: Yeah that might be better 👍

packages/calcite-components/src/components/color-picker/utils.ts
driskull: Ah, yeah i was just wondering where the magic number came from :)

Issue #10897, opened 2024-11-27
Combobox not correctly passing scale to `combobox-item-group`


ashetland:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Scales are not set on the item group. Pictured below is small scale, but the group text is at medium scale
![CleanShot 2024-11-26 at 17 53 21@2x](https://github.com/user-attachments/assets/7108670b-1f37-420a-a4f8-f470983c07dc)

Blocked issues: #7099 

### Expected Behavior

Setting the scale on the parent should pass to the item groups.

### Reproduction Sample

https://codepen.io/ashetland/pen/QwLWpZL

### Reproduction Steps

Open the Comboboxes and note the size and spacing of the group label text.

### Reproduction Version

v2.13.2

### Relevant Info

cc @anveshmekala 

### Priority impact

impact - p2 - want for an upcoming milestone

### Impact

Related #7099 

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

geospatialem:
Verified in `3.0.0-next.42` 🎉

PR #10908, opened 2024-11-27
fix(combobox-item-group): cascade scale from combobox

anveshmekala:
**Related Issue:** #10897 

## Summary

cascade `scale` from `combobox` onto `combobox-item-group`


benelan (approved):
:+1:

Issue #9239, opened 2024-05-01
Slider label wrapping can cause handle to move off track


ellenupp:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Long slider labels can wrap and push the handle off the track. 

![image](https://github.com/Esri/calcite-design-system/assets/14304581/c5a6df8a-9da4-4288-853c-22fbb7c6df92)

### Expected Behavior

Ideally labels wouldn't wrap when there is enough space. If it really needs to wrap it should move up and not affect the position of the handle. Also in the above example I wouldn't expect the tick labels to wrap either as there is enough space for them to be inline. 

![image](https://github.com/Esri/calcite-design-system/assets/14304581/429e47d4-c7bd-45c4-9ea5-03e4884b8ca0)

### Reproduction Sample

https://codepen.io/ellupp/pen/xxeNmoQ

### Reproduction Steps

1. In the sample slide handle to the left and notice that the handle label wraps. It works much better when sliding to the right. 

### Reproduction Version

2.8.0

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Dashboards

richiecarmichael:
Another repo app [here](https://codepen.io/richiecarmichael/pen/LYwYQpR?editors=1000).
![2024-09-23_16-35-11](https://github.com/user-attachments/assets/205b909d-1f44-444e-9161-76f44cf3e76c)

geospatialem:
Verified in `3.0.0-next.46`:

![image](https://github.com/user-attachments/assets/e15fbcf3-d689-402e-aaf2-2e4bdf75a591)

![image](https://github.com/user-attachments/assets/655ebca2-a454-42a0-9590-ef1076485e2c)

PR #10922, opened 2024-12-02
fix(slider): avoid breaking thumb label on edges

jcfranco:
**Related Issue:** #9239

## Summary

This improves label handling for long thumb labels.

### DIFF
--- a/packages/calcite-components/src/components/slider/slider.scss
+++ b/packages/calcite-components/src/components/slider/slider.scss
@@ -125,6 +125,8 @@
   transform: translate(var(--calcite-slider-thumb-x-offset), var(--calcite-slider-thumb-y-offset));
 
   .handle__label {
+    white-space: nowrap;
+
     &.static,
     &.transformed {
       @apply absolute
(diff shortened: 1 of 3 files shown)


jcfranco:
@josercarcamo FYI on some upcoming slider-related changes.

packages/calcite-components/src/components/slider/slider.tsx
```diff
@@ -995,15 +995,14 @@ export class Slider
    * @private
    */
   private getHostOffset(leftBounds: number, rightBounds: number): number {
-    const hostBounds = this.el.getBoundingClientRect();
-    const buffer = 7;
```
eriklharper, line 999: At first I wasn't sure about removing this buffer, but now that I look at the behavior, there's no more jumpiness in the label offset positioning any more.  I think the original intention to this buffer was to keep the label lined up exactly with the edge boundary of the track, but I like this better!  👍

packages/calcite-components/src/components/slider/slider.scss
```diff
@@ -125,6 +125,8 @@
   transform: translate(var(--calcite-slider-thumb-x-offset), var(--calcite-slider-thumb-y-offset));
 
   .handle__label {
+    white-space: nowrap;
```
eriklharper, line 128: Gotta love 1-liners!

eriklharper (approved):
Should we add a screenshot test for this?

ashetland (approved):
🚀 🎉

Issue #8261, opened 2023-11-27
Add option to render icons (or other assets/components) synchronously


hcampos-professional:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

When displaying an icon in a `<calcite-icon>`, the actual icon is loaded asynchronously. This is fine if the page or component containing the icon is loading for the first time. However, in applications which show/hide elements dynamically, this poses some problems:
- The first time an icon is displayed, users have to wait some time for the icon to appear.
- Even if an icon was previously loaded into the page, creating a new instance referring to the same icon can still cause flickering, because the `<calcite-icon>` component is rendered and the icon itself only gets loaded and rendered asynchronously.

Here is a CodeSandbox which exemplifies the issue: [Sandbox](https://codesandbox.io/p/devbox/restless-moon-pff4yx?workspaceId=a8ef6e35-b632-4260-911d-bff54e173524)

### Acceptance Criteria

- We would like to have an option to preload certain icons. Once the icons are preloaded, Calcite components should render them synchronously to the DOM.
- After an icon is loaded (whether it was preloaded or not), subsequent renders should be synchronous.

### Which Component

`<calcite-icon>` and any other component using icons.

### Example Use Case

Within our applications we often have panels, popups, tooltips (etc) containing icons and which are rendered dynamically. Having icons flicker or pop into the app after these components are rendered provides a subpar user experience.

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

janett-baresel:
This also is a problem in Scene Viewer.

jcfranco:
@hcampos-professional A few questions/notes:

1. Would you mind expanding a bit on `We would like to have an option to preload certain icons`? 
    1. Were you thinking on explicitly keeping track of which icons to preload? If so, this might be challenging to maintain since icons can change frequently across different components.
    1. Would preloading all icons be a valid path or is lazy-loading still desired?
1. https://github.com/Esri/calcite-design-system/pull/9197 was installed this year and might help mitigate some of the flickering concerns with rendering of loaded icons.
[...]

PR #10926, opened 2024-12-02
feat(icon): add preload property

jcfranco:
**Related Issue:** #8261

## Summary

Adds new property that will allow icon data to be preloaded for eager rendering.


driskull (approved):
👍 Nice!!

Issue #10932, opened 2024-12-02
Updating the scale property on List doesn't update child elements


eriklharper:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The List component doesn't updating child `list-item` or `list-item-group` components when its `scale` property is modified.

### Expected Behavior

Changes to the list's scale should propagate down to its children.

### Reproduction Sample

https://codepen.io/eriklharper/pen/OPLVYNo?editors=1011

### Reproduction Steps

1. Open Codepen
2. Change scales in the select dropdown to see that no visible change occurs.

### Reproduction Version

3.13.2

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

> @eriklharper I also noticed if you change the scale on the list it isn't updating the child items. There should be something in `willUpdate` to handle changes to `scale`. cc @jcfranco. Can we get an issue @geospatialem? We should get this fixed for beta testing.

Good call, this is something that should be supported.  I can get on that issue today, it should be just a matter of setting it up to watch the changes on `scale` in `willUpdate`.

_Originally posted by @eriklharper in https://github.com/Esri/calcite-design-system/issues/10853#issuecomment-2512279198_

geospatialem:
Verified in `3.0.0-next.42` 👍🏻

PR #10935, opened 2024-12-02
fix(list): update child list components when scale property changes

eriklharper:
**Related Issue:** #10932

## Summary

This fixes an issue where changing `scale` on a root `list` component did not update child `list-item` or `list-item-group` elements to match the new scale.

### DIFF
--- a/packages/calcite-components/src/components/list/list.e2e.ts
+++ b/packages/calcite-components/src/components/list/list.e2e.ts
@@ -317,6 +317,57 @@ describe("calcite-list", () => {
     }
   });
 
+  it("should set the scale property on items", async () => {
+    const page = await newE2EPage();
+    await page.setContent(
+      html`<calcite-list id="root" display-mode="nested" group="my-list">
+        <calcite-list-item open label="Depth 1" description="Item 1">
+          <calcite-list group="my-list">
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/list/list.e2e.ts
```diff
@@ -317,6 +317,57 @@ describe("calcite-list", () => {
...
+    await page.waitForTimeout(DEBOUNCE.filter);
+
+    const items = await page.findAll("calcite-list-item");
+
+    for (let i = 0; i < items.length; i++) {
+      expect(await items[i].getProperty("scale")).toBe("m");
+    }
+
+    const rootList = await page.find("#root");
+    rootList.setProperty("scale", "m");
```
driskull, line 352: "m" should be default right? I don't think this should be necessary.

driskull (approved):
👍 🎉 🌮

packages/calcite-components/src/components/list/list.e2e.ts
eriklharper: woops, I meant to set `s` there, good catch :)

eriklharper: Updated the test to first set `s`, then `m` again, then `l`.  That should cover the bases.

Issue #10867, opened 2024-11-25
Test: Popover reopens when trigger is clicked and autoClose is true


geospatialem:
### Priority impact

p - medium

### Test type

new test

### Which Component(s)

`popover`

### Additional Info

https://github.com/Esri/calcite-design-system/pull/10842#pullrequestreview-2459314364

geospatialem:
Verified on the `dev` branch.

PR #10948, opened 2024-12-03
test(popover): ensure popover doesn't reopen when trigger is clicked and autoClose is true

Elijbet:
**Related Issue:** #10867

## Summary
Ensure `popover` doesn't reopen when the trigger is clicked and `autoClose` is true.


driskull (approved):
👍 💯

Issue #10335, opened 2024-09-17
Focused dropdown item loses ring when hovered


nwhittaker:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The focused dropdown-item loses its focus ring when it's hovered.

https://github.com/user-attachments/assets/c170c397-fe20-432b-9052-fbada0f63dea

### Expected Behavior

The focus ring is maintained on hover.

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/dropdown/#sample

### Reproduction Steps

1. Visit the dropdown doc sample.
2. Move the pointer over the dropdown items.

### Reproduction Version

2.12.2

### Priority impact

impact - p2 - want for an upcoming milestone

### Impact

Impact is on usability. Losing the focus ring is confusing as it implies focus was shifted somewhere else.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Field Apps

jcfranco:
@geospatialem @DitwanP Fixed this locally as I was working on #7688. Submitted a PR, so we can move this to an upcoming milestone.

geospatialem:
Verified on `3.0.0-next.46`:

![verify-drowndown-focus-hover](https://github.com/user-attachments/assets/2dc5727b-4949-4f91-a755-d1eb789a0aee)

PR #10951, opened 2024-12-03
fix(dropdown-item): keep focus ring when selected dropdown item is hovered

jcfranco:
**Related Issue:** #10335 

## Summary

Updates styles to no longer lose focus outline when hovering a selected item.


ashetland (approved):
🚀

Issue #10731, opened 2024-11-15
Follow up on remaining Lumina migration notes


jcfranco:
### Description

In addition to the migration notes added by #10310, the following should be addressed as well:

* refactor `filter#filter()` to use the debounced function (see ad054fd07e88b9cb304f02b7ec2006a146e10d08)
* revisit internal event emitting flag used in `list` (see ad054fd07e88b9cb304f02b7ec2006a146e10d08).
* ✅ ensure all open comboboxes are rendered in the snapshot ([Combobox: With Description Short Label And Content End Slot](https://www.chromatic.com/test?appId=6266d45509d7eb004aa254fb&id=6736c1960aca284626b19095)) – https://github.com/Esri/calcite-design-system/pull/10952
* ✅ fix GH pages content – https://github.com/Esri/calcite-design-system/pull/10807
* ✅ simplify config (see https://github.com/Esri/calcite-design-system/pull/10482/files/57fefcd2747528235d727a17fd947f0c4b4f6a96..04af92c28583f05b22db71f111717bf68e24afb6#r1844236385) - https://github.com/Esri/calcite-design-system/pull/11266

#### Stabilize

☑️ – no longer shows up
```
@esri/calcite-components:test:  FAIL  src/components/combobox/combobox.e2e.ts > calcite-combobox > reflects > %p scale
@esri/calcite-components:test: Error: vitest-fail-on-console > Expected test not to call console.error().
@esri/calcite-components:test:     If the error is expected, test for it explicitly by mocking it out using: 
@esri/calcite-components:test:     vi.spyOn(console, 'error').mockImplementation(() => {}) 
@esri/calcite-components:test:     and test that the warning occurs.
@esri/calcite-components:test: 
@esri/calcite-components:test: "chevronDown16" is not a valid calcite-ui-icon name 
@esri/calcite-components:test: Stack:
@esri/calcite-components:test:   http://localhost:5173/src/components/icon/utils.ts:23:14
@esri/calcite-components:test: 
@esri/calcite-components:test:     at Console.l [as error] (file:///home/runner/work/calcite-design-system/calcite-design-system/node_modules/vitest-fail-on-console/src/index.ts:90:31)
@esri/calcite-components:test:     at ae (file:///home/runner/work/calcite-design-system/calcite-design-system/node_modules/@arcgis/lumina-compiler/dist/puppeteerTesting/index.js:2:1926)
[...]

PR #10952, opened 2024-12-03
fix(combobox): fix sporadic change event emitted on initialization

jcfranco:
**Related Issue:** #10731

## Summary

Tweaks initialization logic to prevent extraneous change event that happened after #10310.


driskull (approved):
👍

jcfranco:
@eriklharper I like that idea. Maybe we can include it in https://github.com/Esri/calcite-design-system/issues/11080.

Issue #10941, opened 2024-12-02
[Filter] includes margin-inline that prevents proper alignment to other components at the same scale


ashetland:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Filter has a `label` wrapper that is adding 4px of `margin-inline` to each scale.
![CleanShot 2024-12-02 at 14 56 16@2x](https://github.com/user-attachments/assets/1e662ebb-0e3a-4c3d-bac6-08eda6f1265b)

### Expected Behavior

This extra margin is removed. The padding defined at each scale is sufficient to ensure alignment with other components at the same scale.

### Reproduction Sample

https://codepen.io/ashetland/pen/azoOgMX?editors=1000

### Reproduction Steps

Note that each scale has 4px of extra margin-inline.

### Reproduction Version

v2.13.0

### Relevant Info

Follow-up to PR #10853. Discovered after this effort, but this bug has been present for a long time. 
@eriklharper 

### Priority impact

impact - p1 - need for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

geospatialem:
Verified on `3.0.0-next.51`

PR #10963, opened 2024-12-03
fix(filter): remove unnecessary inline margin from internal label element

eriklharper:
**Related Issue:** #10941

## Summary

This update removes an inline margin applied to `calcite-filter`'s internal `label` element.  This spacing is now handled by `list`'s supported scales.


ashetland (approved):
🚀 🥳

Issue #10895, opened 2024-11-27
calcite-input-time-zone throws error on Chrome


richiecarmichael:
### Check existing issues
- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior
`calcite-input-time-zone` crashes on initialization on Chrome and Edge. Works fine on FF and Safari.

### Expected Behavior
Component does not crash.

### Reproduction Sample
See below

### Reproduction Steps
To reproduce failure with `DateTimeFormat`.
1. Click [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from)
2. Enter the following code:
```js
const t = new Intl.DateTimeFormat("en", {
  timeZone: "Factory",
  timeZoneName: "shortOffset",
});
```
4. Click Run.
### Reproduction Version
3.0.0

### Regression?
Yes

### Priority impact
impact - p1 - need for current milestone

### Impact
Affects the `TimeZoneLabel` JSSDK widget when used on Chrome or Edge.

### Calcite package
- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team
ArcGIS Maps SDK for JavaScript

### Reference
[#66469](https://devtopia.esri.com/WebGIS/arcgis-js-api/issues/66471) ValuePicker - Migrate to Calcite 3.0

// @jcfranco

nCastle1:
I'm seeing a similar issue when setting up time zone picker both in the usual way through tsx, and by manually creating and attaching to DOM (testing workflow). I see slightly different behavior where creating with document.createElement and appending also breaks on firefox, but firefox works fine through tsx.

jcfranco:
This seems related to a recent Chrome update where “Factory” is no longer a valid timeZone option in Intl.DateTimeFormat. I’ve filed a [Chromium issue](https://issues.chromium.org/issues/381620359) to track this.

@nCastle1, could you share the console error you’re encountering? I suspect it might be a different issue.

nCastle1:
I investigated further, and I think the error was unrelated. I couldn't reproduce in codepen and narrowed down to an asset loading issue in test environment.

hcampos-professional:
[...]

PR #10964, opened 2024-12-03
fix(input-time-zone): add workaround for `Factory` time zone error in Chrome

jcfranco:
**Related Issue:** #10895 

## Summary

Add workaround for https://issues.chromium.org/issues/381620359.

**Note**: `timezone-groups` was [also updated](https://github.com/Esri/calcite-design-system/pull/10961) with this workaround.

### DIFF
--- a/packages/calcite-components/src/components/input-time-zone/utils.ts
+++ b/packages/calcite-components/src/components/input-time-zone/utils.ts
@@ -251,6 +251,12 @@ function getTimeZoneShortOffset(
   locale: SupportedLocale,
   referenceDateInMs: number = Date.now(),
 ): string {
+  // workaround for https://issues.chromium.org/issues/381620359
+  // see https://github.com/Esri/calcite-design-system/issues/10895 for more info
+  if (timeZone === "Factory") {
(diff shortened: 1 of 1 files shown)


packages/calcite-components/src/components/input-time-zone/utils.ts
```diff
@@ -251,6 +251,12 @@ function getTimeZoneShortOffset(
   locale: SupportedLocale,
   referenceDateInMs: number = Date.now(),
 ): string {
+  // workaround for https://issues.chromium.org/issues/381620359
+  // see https://github.com/Esri/calcite-design-system/issues/10895 for more info
+  if (timeZone === "Factory") {
```
driskull, line 256: 🏭

driskull (approved):
👍

packages/calcite-components/src/components/input-time-zone/utils.ts
richiecarmichael: `"Etc/Unknown"` might be another oddity worth keeping an eye on.

image

Issue #10949, opened 2024-12-03
[List/List Item] Refactor so parent List inherits displayMode value


geospatialem:
### Description

Currently the `list` and `list-item` components can both be set to differing `displayMode` values of `"flat"` or `"nested"`. Instead, we should refactor the components, similar to the `scale` property across the design system and instead have the parent element provide its descendants with the same value for consistency.

cc @driskull @jcfranco 

### Proposed Advantages

- Similar to `scale` use across the design system
- Ease of implementation for designers and developers

### Which Component

- `list`
- `list-item`

### Relevant Info

- Related doc update for the interim: https://github.com/Esri/calcite-design-system/issues/10947
- Original PR: https://github.com/Esri/calcite-design-system/pull/10852

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

geospatialem:
Verified in `3.0.0-next.53` with https://codepen.io/geospatialem/pen/YPKWgqQ

PR #10965, opened 2024-12-03
refactor(list-item): set displayMode on slotted list

driskull:
**Related Issue:** #10949

## Summary

- set displayMode on slotted list
- cleanup 🧹 
- e2e test

### DIFF
--- a/packages/calcite-components/src/components/list-item/list-item.e2e.ts
+++ b/packages/calcite-components/src/components/list-item/list-item.e2e.ts
@@ -59,7 +59,7 @@ describe("calcite-list-item", () => {
       },
       {
         propertyName: "displayMode",
-        defaultValue: undefined,
+        defaultValue: "flat",
       },
     ]);
   });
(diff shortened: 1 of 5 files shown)


packages/calcite-components/src/components/list-item/utils.ts
```diff
@@ -3,9 +3,9 @@ import type { List } from "../list/list";
 import type { ListItemGroup } from "../list-item-group/list-item-group";
 import type { ListItem } from "./list-item";
 
-const listSelector = "calcite-list";
-const listItemGroupSelector = "calcite-list-item-group";
-const listItemSelector = "calcite-list-item";
+export const listSelector = "calcite-list";
+export const listItemGroupSelector = "calcite-list-item-group";
+export const listItemSelector = "calcite-list-item";
```
eriklharper, line 8: Excellent idea :)

packages/calcite-components/src/components/list-item/list-item.e2e.ts
```diff
@@ -396,6 +396,28 @@ describe("calcite-list-item", () => {
...
+    await page.waitForChanges();
+
+    expect(await list.getProperty("displayMode")).toBe("nested");
+
+    listItem.setProperty("displayMode", "flat");
+    await page.waitForChanges();
+
+    expect(await list.getProperty("displayMode")).toBe("flat");
+  });
+
```
eriklharper, line 420: Should we add a 3rd nesting level to the test?  I assume since the `getListItemChildren` util is flattening everything it should work just fine, but might be cool to have a 3rd layer in here to assert on.

eriklharper (approved):

packages/calcite-components/src/components/list-item/list-item.e2e.ts
driskull: Since the list-item is setting it, we wouldn't need to test another level.

PR #10845, opened 2024-11-22
ci(renovate): group dep bump PRs for arcgis and eslint

benelan:
## Summary

- Group dependency updates in the `@arcgis` scope into a single pull request.
- Group dependency updates that have `eslint` anywhere in the name into a single pull request.

PR #10968, opened 2024-12-04
refactor: ensure component class names are consistent

jcfranco:
**Related Issue:** #10845

## Summary

Renames component classes starting with `Calcite`. 

**Note**: This is not a breaking change since directly importing class interfaces is undocumented. Developers are expected to use the corresponding HTML element type (e.g., `HTMLCalciteActionElement` instead of `Action`).


benelan (approved):
:+1:

Issue #6230, opened 2023-01-04
[calcite-dropdown]: max-height of calcite-dropdown-content not set if max-items = total items


juliannemarik:
### Actual Behavior

If the `max-items` passed into the `calcite-dropdown` equals the total number of `calcite-drodown-item`s in the dropdown, then the `max-height` of the `calcite-dropdown` is not set and it falls back to the default height. 

### Expected Behavior

The maximum number of `calcite-dropdown-item`s (as specified by `max-items`) should be displayed before showing a scroller. This currently works in all cases except for when `max-items` = `calcite-dropdown-item`s.

I _believe_ the issue is [here](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/components/dropdown/dropdown.tsx#L635) and that this should be: 
```tsx
return items.length >= maxItems ? maxScrollerHeight : 0;
``` 

But I'm not sure if this has other impacts. 

### Reproduction Sample

https://codepen.io/juliannemarik/pen/poZEbNq

### Reproduction Steps

1. Open the codepen example
2. Notice that the dropdown doesn't show all 10 items without scrolling even though `max-items` is specified as 10 on line 10
3. Change `max-items` to 9 and observe that it _does_ respect this value. It only doesn't work for 10

### Reproduction Version

beta.99

### Impact

We currently have a workaround for this bug right where we basically add an empty `calcite-dropdown-item` to the dropdown. We hide this empty `calcite-dropdown-item` and it allows us to show all dropdown items before scrolling. This isn't ideal, but it works for now. 

### Esri team

ArcGIS Hub

geospatialem:
Verified on `3.0.0-next.71` with the original repro and our sample playground example

![image](https://github.com/user-attachments/assets/0f2e3dd6-1acc-4b5b-8683-765197fcf4f9)

![image](https://github.com/user-attachments/assets/1d2284e8-49ee-41a8-b5c3-2bd0ea859276)

PR #10971, opened 2024-12-04
fix(combobox, dropdown): honor `max-items` when it matches total items

jcfranco:
**Related Issue:** #6230 

## Summary

Fixes `max-items` logic (thanks @juliannemarik!) to show all items when it matches the total items.


jcfranco:
**Note:** this logic is used in other components, so we'll have to update for consistency:

* [`combobox`](https://github.com/Esri/calcite-design-system/blob/8860dbbcff112759f93a37dd01636a6d17a9d2b2/packages/calcite-components/src/components/combobox/combobox.tsx#L927)
* [`sort-handle`](https://github.com/Esri/calcite-design-system/blob/23ec01b8001946ff57b88626d2ee875474660cc3/packages/calcite-components/src/components/sort-handle/sort-handle.tsx#L62) (has the property, but it seems to be unused)

packages/calcite-components/src/components/dropdown/dropdown.tsx
```diff
@@ -480,11 +480,14 @@ export class Dropdown
...
 
     this.reposition(true);
-    const maxScrollerHeight = this.getMaxScrollerHeight();
-    scrollerEl.style.maxHeight = maxScrollerHeight > 0 ? `${maxScrollerHeight}px` : "";
+    scrollerEl.style.maxBlockSize = this.getMaxHeightStyle(this.getMaxScrollerHeight());
     this.reposition(true);
   }
 
+  private getMaxHeightStyle(maxHeight: number): string {
+    return maxHeight > 0 ? `min(${maxHeight}px, 90vh)` : "";
```
driskull, line 488: Our menu style has this:

image

Should we make this a calcite global var or something? so we can use/reference it?

driskull: 90vh is fine if the dropdown is at the top of the page but if its centered then 45 might be better. No strong opinion here though.

jcfranco: > Should we make this a calcite global var or something? so we can use/reference it?

Yeah, I'll add a variable of some sort. #sharingiscaring

> 90vh is fine if the dropdown is at the top of the page but if its centered then 45 might be better. No strong opinion here though.

I'll play with those options.

jcfranco:
@driskull Does `sort-handle` need to pass [`maxItems`](https://github.com/Esri/calcite-design-system/blob/23ec01b8001946ff57b88626d2ee875474660cc3/packages/calcite-components/src/components/sort-handle/sort-handle.tsx#L62) to its internal `dropdown`?

driskull:
> @driskull Does sort-handle need to pass [maxItems](https://github.com/Esri/calcite-design-system/blob/23ec01b8001946ff57b88626d2ee875474660cc3/packages/calcite-components/src/components/sort-handle/sort-handle.tsx#L62) to its internal dropdown?

@jcfranco no, i'm not sure why that property is there. I don' think it should be.

jcfranco:
@driskull I'll submit a separate issue to remove `maxItems` from `sort-handle`.

PR #10983, opened 2024-12-04
fix(combobox, stepper, table): respect user hidden attribute

josercarcamo:
**Related Issue:** #8623 

## Summary
Made items with the hide-item attribute added by the user stay hidden.

This PR supersedes https://github.com/Esri/calcite-design-system/pull/10515

## Wiki page
https://github.com/Esri/calcite-design-system/wiki/hiddenItem-Property


jcfranco:
@josercarcamo This is missing the utilities mentioned in [item 2 from the previous review](https://github.com/Esri/calcite-design-system/pull/10515#pullrequestreview-2383410980). Could you update your PR to include them?

Since this supersedes the previous PR, can you close it?

packages/calcite-components/src/assets/styles/includes.scss
```diff
@@ -97,10 +97,8 @@
 }
 
 @mixin base-component() {
-  :host([hidden]) {
```
jcfranco, line 100: Since the internal `hide-item` attribute is only applicable to certain components, can you create a separate mixin and include in applicable children?

packages/calcite-components/src/components/combobox-item/combobox-item.tsx
```diff
@@ -131,6 +131,12 @@ export class ComboboxItem extends LitElement implements InteractiveComponent {
    */
   @property() value: any;
 
+  /**
+   * Specifies whether the user set the hidden attribute in the HTML
+   *
+   */
+  @property() hideItem = false;
```
jcfranco, line 138: Thanks for removing `user` from the prop name! To align with `hidden`, WDYT about `hiddenItem`?

packages/calcite-components/src/components/stepper/stepper.e2e.ts
```diff
@@ -262,37 +262,48 @@ describe("calcite-stepper", () => {
           <calcite-stepper-item heading="Step 3" id="step-3">
             <div>Step 3 content</div>
           </calcite-stepper-item>
+          <calcite-stepper-item heading="Step 4" id="step-4" hide-item>
```
jcfranco, line 265: FIXME: `hide-item` is internal, so it shouldn't be used in tests. You'll need to update lines [like](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/components/stepper/stepper.tsx#L321) [the](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/components/combobox/combobox.tsx#L146-L159) [following](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/components/table/table.tsx#L341) to use the internal prop instead. This change will ensure users can hide elements without our components making them unintentionally visible.

packages/calcite-components/src/components/table/table.e2e.ts
```diff
@@ -1642,9 +1689,8 @@ describe("keyboard navigation", () => {
     await page.keyboard.press("PageUp");
     await page.waitForChanges();
     expect(await getFocusedElementProp(page, "id")).toBe("head-1b");
-    await page.keyboard.down("ControlRight");
+    page.keyboard.press("ControlRight");
```
jcfranco, line 1692: This should use `await` and changing it to `press` will cause it to no longer go into the expected `End` + `ControlRight` path [here](https://github.com/Esri/calcite-design-system/blob/b2c976214774490b5a313e1fafc67415fe212b37/packages/calcite-components/src/components/table-row/table-row.tsx#L248-L251) (i.e., `ControlRight` is pressed and released before `End` is pressed).

jcfranco (changes requested):
Good progress, @josercarcamo! 😎 This needs a few more tweaks, and it'll be good to go!

josercarcamo:
@jcfranco please review again.

packages/calcite-components/src/utils/component.ts
```diff
@@ -26,3 +29,7 @@ export function warnIfMissingRequiredProp<C extends LitElement>(
     logger.warn(`[${component.el.localName}] "${newProp.toString()}" or "${deprecatedProp.toString()}" is required.`);
   }
 }
+
+export function isHidden<C extends ComboboxChildElement | StepperItem["el"] | TableRow["el"]>(component: C): boolean {
+  return component?.hidden || component?.hiddenItem;
```
jcfranco, line 34: Can `component` be `undefined` here?

packages/calcite-components/src/components/table/table.e2e.ts
```diff
@@ -34,6 +34,41 @@ describe("calcite-table", () => {
...
+            <calcite-table-cell id="cell-4b">cell</calcite-table-cell>
+          </calcite-table-row>
+        </calcite-table>`,
+      );
+
+      await page.waitForChanges();
+      const items = await page.findAll("calcite-table-row");
+      for (let i = 0; i < items.length; i++) {
+        const style = await items[i].getComputedStyle();
+        expect(style["display"]).toBe(i === 2 || i === 3 ? "none" : "contents");
```
jcfranco, line 69: This test might go away, but I think you can use `E2EElement.isVisible` for these assertions.

packages/calcite-components/src/components/table/table.e2e.ts
```diff
@@ -34,6 +34,41 @@ describe("calcite-table", () => {
       </calcite-table>`,
       { display: "flex" },
     );
+
+    it("renders only non-hidden rows", async () => {
```
jcfranco, line 38: I think we can drop this test since it's testing native browser behavior.

packages/calcite-components/src/utils/component.ts
```diff
@@ -26,3 +29,7 @@ export function warnIfMissingRequiredProp<C extends LitElement>(
     logger.warn(`[${component.el.localName}] "${newProp.toString()}" or "${deprecatedProp.toString()}" is required.`);
   }
 }
+
+export function isHidden<C extends ComboboxChildElement | StepperItem["el"] | TableRow["el"]>(component: C): boolean {
```
jcfranco, line 33: For consistency, can we rename the input argument to `el`?

jcfranco (changes requested):
Great stuff, @josercarcamo! One more pass and this should be good to go! 

BTW, these changes caused some `combobox` and `input-time-zone` tests to fail, so they'll need to be updated.

packages/calcite-components/src/components/stepper-item/stepper-item.tsx
```diff
@@ -274,6 +281,8 @@ export class StepperItem extends LitElement implements InteractiveComponent, Loa
   private handleItemClick(event: MouseEvent): void {
     if (
       this.disabled ||
+      this.hiddenItem ||
```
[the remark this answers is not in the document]
josercarcamo: When I run the e2e tests, this is what I get:

image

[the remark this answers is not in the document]
jcfranco: Can you try passing the element instead of the instance? You'll get that error if you pass the latter.

josercarcamo:
@jcfranco one more review please.

packages/calcite-components/src/components/combobox-item-group/combobox-item-group.tsx
```diff
@@ -53,6 +53,13 @@ export class ComboboxItemGroup extends LitElement {
    */
   @property() scale: Scale = "m";
 
+  /**
+   * Specifies whether the user set the hidden attribute in the HTML
```
jcfranco, line 57: This comment doesn't seem to be accurate. Can you give it another look? Applies to most `hiddenItem` prop doc.

packages/calcite-components/src/components/stepper/stepper.e2e.ts
```diff
@@ -739,23 +771,26 @@ describe("calcite-stepper", () => {
...
-          <div>Step 1 content</div>
-        </calcite-stepper-item>
-        <calcite-stepper-item heading="Step 2" id="step-2">
-          <div>Step 2 content</div>
-        </calcite-stepper-item>
-      </calcite-stepper>`,
-    );
+  it.each(["disabled", "hidden"])(
+    "should select the next enabled stepper-item if first stepper-item is %s",
+    async (accessibility) => {
```
jcfranco, line 776: IMO, `accessibility` seems out of place here. Would `stateProp` be a better fit?

packages/calcite-components/src/components/stepper/stepper.tsx
```diff
@@ -390,7 +391,8 @@ export class Stepper extends LitElement {
 
   private handleDefaultSlotChange(event: Event): void {
     const items = slotChangeGetAssignedElements(event).filter(
-      (el): el is StepperItem["el"] => el?.tagName === "CALCITE-STEPPER-ITEM",
+      (el): el is StepperItem["el"] =>
+        el?.tagName === "CALCITE-STEPPER-ITEM" && !isHidden(el as StepperItem["el"]),
```
jcfranco, line 395: Can you confirm if you need to typecast here? It looks like `el` is properly typed by the filter's return type.

jcfranco (approved):
👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻
👻✅👻👻👻✅👻✅✅✅👻👻✅✅✅👻✅✅✅✅👻✅👻
👻✅✅👻👻✅👻👻✅👻👻✅👻👻👻👻✅👻👻👻👻✅👻
👻✅👻✅👻✅👻👻✅👻👻✅👻👻👻👻✅✅✅👻👻✅👻
👻✅👻👻✅✅👻👻✅👻👻✅👻👻👻👻✅👻👻👻👻👻👻
👻✅👻👻👻✅👻✅✅✅👻👻✅✅✅👻✅✅✅✅👻✅👻
👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻👻

Once comments are addressed, this will be good to merge! Thanks, @josercarcamo!

packages/calcite-components/src/utils/component.ts
```diff
@@ -26,3 +29,7 @@ export function warnIfMissingRequiredProp<C extends LitElement>(
     logger.warn(`[${component.el.localName}] "${newProp.toString()}" or "${deprecatedProp.toString()}" is required.`);
   }
 }
+
+export function isHidden<C extends ComboboxChildElement | StepperItem["el"] | TableRow["el"]>(el: C): boolean {
```
driskull, line 33: Maybe we could have an interface for `HiddenItem` that each of these components could implement?  Then this utility may not be necessary.

packages/calcite-components/src/components/stepper/stepper.tsx
josercarcamo: It seems like the type casting is needed:

image

packages/calcite-components/src/components/combobox-item-group/combobox-item-group.tsx
```diff
@@ -53,6 +53,13 @@ export class ComboboxItemGroup extends LitElement {
...
+   *
+   * @private
+   */
+  @property({ reflect: true }) hiddenItem = false;
```
driskull, line 61: Should we have e2e `defaults` and `reflects` tests for each of these new private properties?

driskull (changes requested):
Looking good @josercarcamo 👍

packages/calcite-components/src/utils/component.ts
josercarcamo: I don't think an interface is appropriate here since they all behave the same.  A parent class would be more appropriate. This way we wouldn't need to pass anything to the method. However, we would only have this one method.

driskull: Now that we're using lit, we may be able to have a class that these could extend.

jcfranco: We're moving more towards controllers, so we could enforce the interface this way. This would be handled separately from this PR.

packages/calcite-components/src/components/combobox-item-group/combobox-item-group.tsx
jcfranco: Not for internal props, but the controller approach could help us test this behavior.

josercarcamo:
@driskull are you ok with handling your two comments in a separate PR and merge this PR for now?

packages/calcite-components/src/components/combobox-item-group/combobox-item-group.tsx
driskull: > Not for internal props, but the controller approach could help us test this behavior.

It would be beneficial even for internal props because if this property isn't reflected then the CSS and querySelector logic won't work. I would think it would be beneficial

packages/calcite-components/src/utils/component.ts
driskull: We can handle this separately with a followup

packages/calcite-components/src/components/combobox-item/combobox-item.tsx
```diff
@@ -134,6 +134,13 @@ export class ComboboxItem extends LitElement implements InteractiveComponent {
    */
   @property() value: any;
 
+  /**
+   * When `true`, the item will be hidden
+   *
+   * @private
+   *  */
+  @property({ reflect: true }) hiddenItem = false;
```
driskull, line 142: nitpick: would `itemHidden` be preferred here? it aligns better with our other props like `feature-disabled`.

driskull:
Should we have a followup issue to use `itemHidden`/`hiddenItem` in place of `filterHidden` and any other props we use to hide items?

packages/calcite-components/src/components/combobox-item/combobox-item.tsx
josercarcamo: I had proposed that but it was decided to go with `hiddenItem`.

driskull: @jcfranco what you think? 🤔  It doesn't matter much since its a private property but would align better with other props as `itemHidden` or even `calciteHidden` to illustrate that calcite is hiding this.

josercarcamo: I'll tackle it here. I'll change it to `calciteHidden`, if there are no objections.

driskull (approved):
👍

jcfranco:
> Sounds good. I like calcite-hidden as it aligns with our other global attributes like calcite-hydrated.

Sorry for chiming in late here, but `calcite-hydrated` is a special case, doesn't have a corresponding prop, nor do we prefix _internal props_ with `calcite`. If we're renaming, I'd suggest `itemHidden`, sticking with the previous name or to find an alternative name. 

This can be done as a follow-up since it is internal.

Issue #10936, opened 2024-12-02
[Tooling] Automation to add commit override message to PRs with new `no changelog entry` label


DitwanP:
### Priority impact

p - low

### Summary

Make the addition of the `no changelog entry` label automate adding the commit override messaging to the PR description.

> BEGIN_COMMIT_OVERRIDE
END_COMMIT_OVERRIDE

### Desired Outcome

Automation that will add commit override to PRs in order to not include them in changelogs.

DitwanP:
Verified 🍠
![image](https://github.com/user-attachments/assets/b9423d74-5504-445c-b2ff-c44513cf6dbe)
<img width="277" alt="Screenshot 2024-05-21 at 11 06 30 AM" src="https://github.com/user-attachments/assets/9cb66a4e-a95d-4dbc-9fb0-0d3736c0745d">

PR #10984, opened 2024-12-04
ci(actions): add action to omit PR from changelog

DitwanP:
**Related Issue:** #10936

## Summary
Adds an action that will edit the PR body to include the necessary comments to omit a PR from the changelog (when the `no changelog entry` label is added to PR). 

Tested on personal repo:
image


.github/scripts/omitPrFromChangelog.js
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  if (label?.name === planning.noChangelogEntry) {
+    const pullRequestProps = {
+      owner,
+      repo,
+      pull_number: number,
+    };
+
+    const newPullRequestBody = pullRequestBody + ommitComment;
```
benelan, line 30: Nitpick: can you use a template string instead of concatenating these strings

```suggestion
    const newPullRequestBody = `${pullRequestBody}${ommitComment}`;
```

.github/scripts/omitPrFromChangelog.js
```diff
@@ -0,0 +1,39 @@
...
+module.exports = async ({ github, context }) => {
+  const { repo, owner } = context.repo;
+  const payload = /** @type {import('@octokit/webhooks-types').PullRequestLabeledEvent} */ (context.payload);
+  const {
+    label,
+    pull_request: { number },
+  } = payload;
+
+  const pullRequestBody = payload.pull_request.body;
+  const ommitComment = `\n\nBEGIN_COMMIT_OVERRIDE\nEND_COMMIT_OVERRIDE`;
```
benelan, line 16: "ommit" -> "omit"

benelan (approved):
LGTM! If someone were to toggle the `no changelog entry` label it would add additional commit override blocks. I'm not sure how release-please would handle that, but it should be fine. Just something to keep an eye on.

PR #10987, opened 2024-12-04
fix(action): disabled styles

alisonailea:
## Summary

Ensures design tokens are correctly applied in the action disabled state


benelan:
There weren't any snapshot diffs from this change, do you think we need to add coverage?

alisonailea:
> There weren't any snapshot diffs from this change, do you think we need to add coverage?

We should probably create a ticket to add disabled state VRT for each component?

driskull (approved):
👍

Issue #10982, opened 2024-12-04
[Action] - Active syles are not present on actions with transparent appearance


sagewall:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Beginning with version `2.12.0-next.31` and currently at `2.13.2` there is no visible change to the action when setting the active property on an action element if the appearance is set to transparent.  Both actions in this screenshot are identical with the exception of the second action has `appearance="transparent"`
![image](https://github.com/user-attachments/assets/9a99e6d9-6517-4a95-9f2d-be206b83fd11)

### Expected Behavior

Visually be able to determine which actions are visible regardless of the `appearance` attribute value as before the regression took place.
![image](https://github.com/user-attachments/assets/1c572d46-a05d-4b04-8034-2c2f3d131c4b)

### Reproduction Sample

https://codepen.io/sagewall/pen/JoPGPwJ

### Reproduction Steps

1. Open this [code pen](https://codepen.io/sagewall/pen/JoPGPwJ)
2. Observe how the transparent action has no styling to indicate it's active

### Reproduction Version

2.13.2

### Relevant Info

Seems independent of hardware

### Regression?

2.12.0-next.31

### Priority impact

impact - p2 - want for an upcoming milestone

### Impact

Impacts users ability to see active actions in several JS SDK widgets and components

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

geospatialem:
Regression related to #10058, refer to Matt's comment for more specifics: https://github.com/Esri/calcite-design-system/pull/10058#discussion_r1869973822

geospatialem:
@alisonailea It looks like the PR has been merged, is this issue ready for verification? If so, can you add the `3- installed` label for us? 🤜🏻 🤛🏻

geospatialem:
The above doesn't seem to be mitigated with the original repro case in the latest `3.0.0-next.65`. 

@alisonailea can you take a peek to explore what might be occurring?: https://codepen.io/geospatialem/pen/azopwzM 👀

geospatialem:
[...]

PR #10988, opened 2024-12-04
fix(action): active styles present with transparent appearance

alisonailea:
**Related Issue:** #10982

## Summary

Resolves issue where transparent styles are not correctly applied when the action[active] state is set.
Adds optional "expectedValue" prop to themed test utility
Adds e2e test coverage

### DIFF
--- a/packages/calcite-components/src/components/action/action.e2e.ts
+++ b/packages/calcite-components/src/components/action/action.e2e.ts
@@ -441,5 +441,26 @@ describe("calcite-action", () => {
         ],
       });
     });
+    describe("transparent", () => {
+      themed(html`<calcite-action appearance="transparent"></calcite-action>`, {
+        "--calcite-action-background-color": {
+          shadowSelector: `.${CSS.button}`,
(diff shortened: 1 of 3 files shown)


driskull (commented):
@alisonailea can we get a screenshot test and/or e2e test added to have test coverage for this?

Looks good once there is a test added.

Unrelated but is this a valid selector? Line 122.

https://github.com/Esri/calcite-design-system/blob/a681da0027380760a286f98aad0ac104b2a71bf9/packages/calcite-components/src/components/action/action.scss#L122

alisonailea:
> Unrelated but is this a valid selector? Line 122.

Yes. It's valid https://sass-lang.com/playground/#eJwzNNTLrdRNzkksLlao5lJQUNMBEVYZ+WWpRWABBYXk/Jz8IiuFpJzSVCC/lqsWAHeADlM=

driskull:
Sounds good!

driskull (approved):
👍 💯 Nice!

PR #11001, opened 2024-12-06
feat(filter): add label property

driskull:
**Related Issue:** #8076

## Summary

- add label prop to filter
- add test

Issue #8076, opened 2023-10-27
[Filter] Add `filterLabel` property


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Add a new `filterLabel` property for users to supply their own filter label applicable to their use case.

Currently the filter provides [not-so-great context via JAWS and NVDA](https://github.com/Esri/calcite-design-system/pull/8069#pullrequestreview-1700750139):

**JAWS**:
```
Filter �tree view item� 
```

**NVDA**:
```
Filter edit blank
```

### Acceptance Criteria

A new `filterLabel`, or similar property

### Relevant Info

Was referred to in https://github.com/Esri/calcite-design-system/issues/5626#issuecomment-1771869038

### Which Component

Filter

### Example Use Case

_refer to description above_

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

PR #11002, opened 2024-12-06
feat(list): add filterLabel property

driskull:
**Related Issue:** #8076

## Summary

- add `filterLabel` property to the list
- add test
- depends on #11001


benelan (approved):
:+1:

PR #11003, opened 2024-12-06
feat(list-item, list): add `expanded` property and deprecate `open` property

Elijbet:
**Related Issue:** #6473, #11225

## Summary
Deprecate `open` prop in favor of `expanded`. Add logger warning for deprecation to help inform users about the deprecated property and guide them towards using the recommended alternatives.


packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -111,12 +111,26 @@ export class ListItem
    */
   @property() bordered = false;
 
-  /** When `true`, a close button is added to the component. */
+  /**
+   * When `true`, a close button is added to the component.
+   *
+   * @deprecated Use `collapsible` prop instead.
+   */
   @property({ reflect: true }) closable = false;
```
driskull, line 119: I don't think this should be collapsible. It isn't collapsed, it isn't shown at all.

driskull: I think closed/closable needs to remain as is.

packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -357,20 +378,32 @@ export class ListItem
     To account for this semantics change, the checks for (this.hasUpdated || value != defaultValue) was added in this method
     Please refactor your code to reduce the need for this check.
     Docs: https://qawebgis.esri.com/arcgis-components/?path=/docs/lumina-transition-from-stencil--docs#watching-for-property-changes */
+    if (changes.has("expanded")) {
+      this.open = this.expanded;
```
driskull, line 382: I don't think this is correct. expanded should set expanded, open should set open.

I think what needs to happen is that open and expanded each need a custom setter to set the other one when one is set.

packages/calcite-components/src/components/list-item/list-item.tsx
Elijbet: You're right, it needs to have both `expanded/collapsed` and `closed/closable`.

packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -169,8 +169,22 @@ export class ListItem extends LitElement implements InteractiveComponent, Sortab
    */
   @property() moveToItems: MoveTo[] = [];
 
-  /** When `true`, the item is open to show child components. */
-  @property({ reflect: true }) open = false;
+  /**
+   * When `true`, the item is open to show child components.
```
Elijbet, line 173: @geospatialem, would it be better to say: `When `true`, expands the component and its contents.`?

readme.md
driskull, line 1: Can we undo changes here?

driskull (commented):
Code changes look good to me,

I think this PR title should be: feat(list-item): add `expanded` property and deprecate `open` property.

driskull (commented):
Can you make sure all other demos and stories are updated with this change as well? Including the HTML files.

packages/calcite-components/src/components/list-item/list-item.e2e.ts
```diff
@@ -45,6 +45,18 @@ describe("calcite-list-item", () => {
         propertyName: "open",
         defaultValue: false,
       },
+      {
```
jcfranco, line 48: Nice! Can you also update the `reflects` test?

packages/calcite-components/src/components/list-item/list-item.e2e.ts
```diff
@@ -370,7 +382,7 @@ describe("calcite-list-item", () => {
     expect(calciteListItemClose).toHaveReceivedEventTimes(1);
   });
 
-  it("should fire calciteListItemToggle event when opened and closed", async () => {
+  it("should fire calciteListItemToggle event when expanded and collapsed", async () => {
```
jcfranco, line 385: Can you keep the open/closed tests grouped in a "deprecated" describe block? Bonus points if you can extract the tests to be reused between both.

packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -169,8 +169,22 @@ export class ListItem extends LitElement implements InteractiveComponent, Sortab
...
+   *
+   * @deprecated Use `expanded` prop instead.
+   */
+  @property({ reflect: true })
+  get open(): boolean {
+    return this.expanded;
+  }
+
+  set open(value: boolean) {
+    this.expanded = value;
```
jcfranco, line 183: Side note: I think we can start using getters/setters to warn on deprecated prop usage (using [`logger.deprecated`](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-components/src/utils/logger.ts#L64-L72)). Could you create an issue for this?

packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -169,8 +169,22 @@ export class ListItem extends LitElement implements InteractiveComponent, Sortab
    */
   @property() moveToItems: MoveTo[] = [];
 
-  /** When `true`, the item is open to show child components. */
-  @property({ reflect: true }) open = false;
+  /**
+   * When `true`, the item is open to show child components.
+   *
+   * @deprecated Use `expanded` prop instead.
```
jcfranco, line 175: Can you update external usage of `open` too? `list` has a [few](https://github.com/Esri/calcite-design-system/blob/904623be3cff426e71d51001a29df188a72ea00b/packages/calcite-components/src/components/list/list.tsx#L884) [usages](https://github.com/Esri/calcite-design-system/blob/904623be3cff426e71d51001a29df188a72ea00b/packages/calcite-components/src/components/list/list.tsx#L631). @driskull Are there any other spots that need updating?

jcfranco (changes requested):
Looking good, @Elijbet! 😎 One more pass and this should be good to land.

Can you also revert the readme changes? Try `git checkout dev -- ./readme.md` (assumes your cwd is `<path-to-projects>/calcite-design-system/packages/calcite-components`).

jcfranco:
@Elijbet Can you also update the PR description to match the title/changes?

packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -169,8 +169,22 @@ export class ListItem extends LitElement implements InteractiveComponent, Sortab
...
+  get open(): boolean {
+    return this.expanded;
+  }
+
+  set open(value: boolean) {
+    this.expanded = value;
+  }
+
+  /** When `true`, the item is expanded to show child components. */
+  @property({ reflect: true }) expanded = false;
```
jcfranco, line 187: Can you move this prop so that all public props are in alphabetical order?

packages/calcite-components/src/components/list-item/list-item.tsx
driskull: They should all just be in that file I would suspect.

Elijbet:
> git checkout dev -- ./readme.md

I tried to roll back the `readme.md` but it won't let me commit it - says it's an empty commit. One suggestion was to edit the PR on GitHub and paste in the code from the main branch so it doesn't do any formatting locally with eslint. I can do that right before merge.

packages/calcite-components/src/components/list-item/list-item.tsx
Elijbet: [Add deprecation warnings for deprecated properties](https://github.com/Esri/calcite-design-system/issues/11653#top) #11653

packages/calcite-components/src/components/list-item/list-item.e2e.ts
Elijbet: Updating for record here: As we decided, because this ends up touching on too many tests, we'll instead just add a test to check if setting `open` sets `expanded`.

packages/calcite-components/src/components/list-item/list-item.e2e.ts
```diff
@@ -370,50 +398,68 @@ describe("calcite-list-item", () => {
...
+      expect(calciteListItemToggle).toHaveReceivedEventTimes(0);
+
+      await expandedButton.click();
+      expect(await listItem.getProperty("expanded")).toBe(false);
+      expect(calciteListItemToggle).toHaveReceivedEventTimes(0);
+    });
   });
 
-  it("should not fire calciteListItemToggle event without nested items", async () => {
+  it("should set expanded when open is set", async () => {
```
jcfranco, line 449: Could you update the test name to clarify that it covers the deprecated prop mapping? Also, can you add comments explaining that this test ensures the deprecated prop is correctly mapped and that broader functionality is covered in the `expanded` tests?

packages/calcite-components/src/components/list-item/list-item.e2e.ts
```diff
@@ -370,50 +398,68 @@ describe("calcite-list-item", () => {
     expect(calciteListItemClose).toHaveReceivedEventTimes(1);
   });
 
-  it("should fire calciteListItemToggle event when opened and closed", async () => {
-    const page = await newE2EPage({
-      html: html`<calcite-list-item id="test" display-mode="nested"
-        ><calcite-list display-mode="nested"><calcite-list-item></calcite-list-item></calcite-list
-      ></calcite-list-item>`,
+  describe("calciteListItemToggle event", () => {
```
jcfranco, line 401: ✨🏆✨

jcfranco (approved):
Awesome! 🚀

Issue #9175, opened 2024-04-23
[List Item]: add icon-end/start props


ashetland:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

List Item should match other similar components in this regard and include properties for `icon-end` and `icon-start`. This would improve the OOTB experience when adding icons to list items vs the current workflow of relying the content slots. 

### Acceptance Criteria

- [x] icon-end/start props added
- [x] default color should be `text.3`
- [x] when the item is selected, the color should match the heading (`text.1`)
- [x] icons should be `scale="m"` (24px) on `scale="l"` List Items

![image](https://github.com/user-attachments/assets/4bb971e6-ce60-4b87-89d1-b2b07ac245c4)

### Which Component

List Item

### Priority impact

p4 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

ashetland:
cc @driskull

geospatialem:
Listed as a `p - high` to ensure all related list visual and breaking changes are in alignment for the November breaking change release. While the above is not a breaking or visual change it relates to other breaking changes such as #7100. cc @ashetland

ashetland:
Due to the refactor required by #7100, it might make sense to wait to add these until after 7100 is complete.

geospatialem:
Verified on Storybook with the `icon-start` and `icon-end` attributes (`iconStart` and `iconEnd` props) and selection states. 👍🏻

aPreciado88:
Created follow up [issue](https://github.com/Esri/calcite-design-system/issues/11021) to split storybook screenshot.

PR #11004, opened 2024-12-06
feat(list-item): add iconStart, iconEnd and iconFlipRtl props

aPreciado88:
**Related Issue:** [#9175](https://github.com/Esri/calcite-design-system/issues/9175)

## Summary

- Add `iconStart`, `iconEnd` and `iconFlipRtl` props.
- Update default icon color to `--calcite-color-text-3`.
- Update icon color to `--calcite-color-text-1` when item is selected.
- Update `list.html` and `list.stories.ts` to use `icon-start` and `icon-end`.


aPreciado88:
@ashetland @SkyeSeitz Can you please take a look at the new screenshots?

packages/calcite-components/src/components/list/list.stories.ts
```diff
@@ -142,18 +174,6 @@ export const scales = (): string => html`
             scale="s"
             slot="actions-start"
           ></calcite-action>
-          <calcite-icon
```
driskull, line 145: @aPreciado88 lets leave these in the demo since it is still valid to use the slots.

packages/calcite-components/src/components/list-item/list-item.tsx
```diff
@@ -919,7 +957,9 @@ export class ListItem
     const content = [
       this.renderContentStart(),
       this.renderCustomContent(),
+      this.renderIconStart(),
```
driskull, line 960: Question: Do we always put the icons inside the custom content slots or vice versa? Just want to confirm for designs. cc @ashetland @SkyeSeitz

driskull (changes requested):
> Update list.html and list.stories.ts to use icon-start and icon-end instead of content slots.

Can we leave the examples using slots since using slots is still a valid use case and can track the ordering of content slots and icon props.

All other changes look good!

Just had a question on whether our design consistency puts content slots outside or inside icon props.

packages/calcite-components/src/components/list/list.stories.ts
aPreciado88: @driskull I chatted with Aaron about this and the suggestion was to get rid of content slots in favor of iconStart and iconEnd. Should I add all of the content slots back or only add them to a couple of tests?
cc @ashetland

ashetland: We could also use `calcite-avatar` in the content slots. Since we now have icon-start/end it felt best to move icons to the props, leaving the content slots open for custom content.

packages/calcite-components/src/components/list-item/list-item.tsx
ashetland: Now that we have the icon props, we should be using them from Calcite icons. The content slots can be used for other custom content.

driskull: Correct. my question was whether the custom content like an avatar surrounds the icons or goes inside of them.

packages/calcite-components/src/components/list/list.stories.ts
driskull: Yeah we can update them to use an avatar if we want. My point is that they are still valid slots and users can use both icon props and content slots.

ashetland: Honestly, we removed a lot of them to try to get that story to load faster, but I believe we were going to log a separate issue to clean up the List stories. This story in particular seems to always show a diff because the screenshot is taken before the page finishes loading.

packages/calcite-components/src/components/list-item/list-item.tsx
ashetland: Sorry for the confusion. The content slots should be on the outside of the icons:
CleanShot 2024-12-09 at 12 41 40@2x

driskull: Sounds good. Code looks correct for that 👍

aPreciado88:
@driskull @ashetland I added back the content slots and updated them to use `calcite-avatar`.

driskull (approved):
👍

ashetland (approved):
✨🚀

Issue #9974, opened 2024-08-06
[Input] Deprecate the `"textarea"` `type`


geospatialem:
### Description

With the [`text-area`](https://developers.arcgis.com/calcite-design-system/components/text-area/) component we should consider leading users in this direction rather than using it via the `input` component.

We've had enhancement requests, like https://github.com/Esri/calcite-design-system/issues/8568 surface, where the `text-area` component should be leveraged instead.

### Which Component

`input`

### Resources

Once deprecated we should consider for the `4.0` breaking change release in late 2025 - and proactively add it to our list in Confluence once verified.

DitwanP:
Verified 

<img width="277" alt="Screenshot 2024-05-21 at 11 06 30 AM" src="https://github.com/user-attachments/assets/867747f2-2c3f-4bf1-a9e1-0d0c9f8adcbb">

PR #11005, opened 2024-12-06
docs: deprecate textarea type for input component

DitwanP:
**Related Issue:** #9974 

## Summary
Deprecate the `"textarea"` value for the `type` property on the input component.


geospatialem (commented):
Using the `@deprecated` tag would deprecate the entire property. Maybe we should consider adding the language that its deprecated in the JSDoc without using the tag? Perhaps similar to [`shell-panel`'s `displayMode` prop for `"float"`](https://developers.arcgis.com/calcite-design-system/components/shell-panel/#api-reference-properties-displayMode)?

Also, for consistency we should use `calcite-text-area` to specify the component, similar to how `calcite-action-bar` is referenced in the `displayMode` prop as well:


https://github.com/Esri/calcite-design-system/blob/4e5bf1d801e47f3e309fd0259b9731b95dc079d9/packages/calcite-components/src/components/shell-panel/shell-panel.tsx#L166-L179

packages/calcite-components/src/components/input/input.tsx
```diff
@@ -343,6 +343,8 @@ export class Input
    * Specifies the component type.
    *
    * Note that the following `type`s add type-specific icons by default: `"date"`, `"email"`, `"password"`, `"search"`, `"tel"`, `"time"`.
+   *
+   *  The `"textarea"` type is [Deprecated], use the `calcite-text-area` component instead.
```
geospatialem, line 347: Maybe for consistency with other value deprecations, the deprecation should be mentioned before the value? Perhaps something like:

```suggestion
   *  `"textarea"` [Deprecated], use the `calcite-text-area` component instead.
```

geospatialem (approved):
Small consideration, otherwise LGTM! 🚀📝

PR #10970, opened 2024-12-04
feat(text-area): add support for `minLength`

Elijbet:
**Related Issue:** #8626

## Summary
Add support for `minlength` to `text-area`.

PR #11006, opened 2024-12-07
chore(text-area): add new string for when character length is `tooShort`

Elijbet:
**Related Issue:** #10970

## Summary
Check in new string.


benelan (approved):
:+1:

jcfranco:
@Elijbet Why is this separate from https://github.com/Esri/calcite-design-system/pull/10970?

Elijbet:
> @Elijbet Why is this separate from #10970?

From what I understood I could check in new strings and request translations while that PR is being reviewed.

jcfranco:
Gotcha. We might not need to submit this separately since #10970 looks close.

Elijbet:
> Gotcha. We might not need to submit this separately since #10970 looks close.

Merging this in per: https://github.com/Esri/calcite-design-system/pull/10970#discussion_r1889383200

PR #11014, opened 2024-12-10
feat(block): add component tokens

Elijbet:
**Related Issue:** #7180

## Summary
Adds the following subcomponent tokens (CSS props):

- `--calcite-block-border-color`
- `--calcite-block-header-background-color`
- `--calcite-block-header-background-color-hover`
- `--calcite-block-header-text-color`
- `--calcite-block-block-text-color`
- `--calcite-block-block-text-color-hover`


packages/calcite-components/src/components/block/block.scss
```diff
@@ -150,18 +171,18 @@ calcite-handle {
...
-  @apply text-color-3
-  transition-color
+  @apply transition-color
   self-center
   justify-self-end
   duration-150
   ease-in-out;
 
   margin-inline-end: var(--calcite-spacing-md);
+  color: var(--calcite-block-header-icon-color, var(--calcite-color-text-3));
```
alisonailea, line 181: content, description, and icons all share this color. Therefore we can simplify it all down to a single "block-text-color" token which the icons will inherit via their default `currentColor`

packages/calcite-components/src/components/block/block.scss
```diff
@@ -170,12 +191,14 @@ calcite-handle {
...
-  padding-inline: var(--calcite-block-padding, var(--calcite-spacing-md));
+  padding-block: var(--calcite-spacing-sm);
+  padding-inline: var(--calcite-spacing-md);
 }
 
 .content-start {
-  @apply text-color-3 flex items-center;
+  @apply flex items-center;
+
+  color: var(--calcite-block-content-start-color, var(--calcite-color-text-3));
```
alisonailea, line 201: `--calcite-block-text-color` represents the existing design pattern and we don't want to add extra tokens unless specifically asked by our users

packages/calcite-components/src/components/block/block.scss
```diff
@@ -195,7 +218,7 @@ calcite-action-menu {
   @apply my-2;
 
   .header .title .heading {
-    @apply text-color-1;
+    color: var(--calcite-block-header-color, var(--calcite-color-text-1));
```
alisonailea, line 221: `--calcite-block-header-text-color`

packages/calcite-components/src/components/block/block.scss
```diff
@@ -150,18 +171,18 @@ calcite-handle {
...
   duration-150
   ease-in-out;
 
   margin-inline-end: var(--calcite-spacing-md);
+  color: var(--calcite-block-header-icon-color, var(--calcite-color-text-3));
 }
 
 .toggle:hover .toggle-icon {
-  @apply text-color-1;
+  color: var(--calcite-block-header-icon-color, var(--calcite-color-text-1));
```
alisonailea, line 185: the actual design pattern is that the toggle gets the header title color on hover so this can be simplified to 

`color: var(--calcite-block-header-text-color, var(--calcite-color-text-1));`

packages/calcite-components/src/components/block/block.scss
```diff
@@ -150,18 +161,18 @@ calcite-handle {
...
-  @apply text-color-3
-  transition-color
+  @apply transition-color
   self-center
   justify-self-end
   duration-150
   ease-in-out;
 
   margin-inline-end: var(--calcite-spacing-md);
+  color: var(--calcite-block-header-text-color, var(--calcite-color-text-3));
```
alisonailea, line 171: Unless a component has changed state at the :host level, a token used throughout the component must always have the same default. So lines 171, 175, and 211 won't work.

alisonailea (changes requested):
Please reassess the block component tokens. Refer to https://github.com/Esri/calcite-design-system/wiki/tokens-component for how to apply tokens.

packages/calcite-components/src/components/block/block.scss
Elijbet: This has `text-color-2` and `text-color-1` for `open` state.

packages/calcite-components/src/components/block/block.scss
```diff
@@ -3,16 +3,23 @@
...
 
 :host {
   @extend %component-host;
   @extend %component-spacing;
-  @apply transition-margin ease-cubic border-color-3 flex flex-shrink-0 flex-grow-0
+  @apply transition-margin ease-cubic flex flex-shrink-0 flex-grow-0
     flex-col border-0 border-b border-solid p-0;
   flex-basis: auto;
   transition-duration: var(--calcite-animation-timing);
+  border-color: var(--calcite-block-border-color, var(--calcite-color-border-3));
```
alisonailea, line 22: add here
`color: var(--calcite-block-text-color, var(--calcite-color-text-3))`
setting it here should mean we can reduce lines of code by removing the need to set `color` further down the component style cascade.
Double check styles still look correct

packages/calcite-components/src/components/block/block.scss
```diff
@@ -66,20 +76,20 @@
...
     flex-nowrap
     items-center
     justify-between
     border-none
+    m-0
     p-0;
 
   text-align: initial;
-  background-color: transparent;
+  background-color: var(--calcite-block-header-background-color, transparent);
```
alisonailea, line 89: make sure `background-color: var(--calcite-block-header-background-color, transparent);` is being applied to the entire header as well.

packages/calcite-components/src/components/block/block.scss
```diff
@@ -195,7 +209,7 @@ calcite-action-menu {
   @apply my-2;
 
   .header .title .heading {
-    @apply text-color-1;
+    color: var(--calcite-block-text-color, var(--calcite-color-text-1));
```
alisonailea, line 212: This should be `--calcite-block-text-color-hover,`

alisonailea (changes requested):
just a few things

packages/calcite-components/src/components/block/block.scss
Elijbet: After adding it, I tested removing it in other spots further down the component and it didn't really work.

packages/calcite-components/src/components/block/block.scss
Elijbet: ✅ checked!

alisonailea (approved):
🎉

Issue #5522, opened 2022-10-19
calcite-slider > layout vertical and horizontal


SaschaBrunnerCH:
### Description

It should be possible to use the slider vertical. Like it is possible with the ArcGIS JS API Slider: https://developers.arcgis.com/javascript/latest/api-reference/esri-widgets-Slider.html#layout

![image](https://user-images.githubusercontent.com/46531184/196774646-f96e2d46-5727-4f33-80fc-dcfde37b7ef0.png)

### Acceptance Criteria

Possible to configure the layout:
- horizontal
- horizontal-reversed
- vertical
- vertical-reversed

### Relevant Info

See: https://developers.arcgis.com/javascript/latest/api-reference/esri-widgets-Slider.html#layout

### Which Component

calcite-slider

### Example Use Case

For a sample app we would like to include a slider to select the floor. For that purpose it should be vertical:

Example with the horizontal slider: 
![slider-floor](https://user-images.githubusercontent.com/46531184/196775397-1b56bd7c-6a99-4b9a-84c9-30bfb814227e.gif)

### Esri team

ArcGIS API for JavaScript

richiecarmichael:
This feature is required for:
[#46032](https://devtopia.esri.com/WebGIS/arcgis-js-api/issues/46032) **Range Slider**

The JSAPI Range Slider will be used by MapViewer to assign values to multi-dimensional rasters.

// @driskull @jcfranco

ubatsukh:
@driskull as we discussed yesterday this is a high priority issue for us! MV needs Range Slider for multidimensional image services and the API implementation cannot proceed until the calcite slider has support for this.

ashetland:
Hey folks, see the below for a proposed vertical layout Slider. In addition to `layout=vertical`, this outlines `direction-reversed` and `labels-reversed`. I haven't included range and histogram here, but they'd follow the same logic. Figma file available for more details.
<img width="1624" alt="Slice 1" src="https://user-images.githubusercontent.com/108549080/229635203-c2a97d7b-c798-4f7c-85f0-364a50e85611.png">
<img width="1624" alt="Slice 2" src="https://user-images.githubusercontent.com/108549080/229635218-fa9a2681-5441-4ec7-9b75-f0263d18b603.png">

geospatialem:
Priority has shifted to a medium status, with Maps SDK shifting priority to #6590 and #6591.

annelfitz:
[...]

PR #10652, opened 2024-10-30
feat(slider): Implement vertical orientation

josercarcamo:
**Related Issue:** #5522

## Summary
Implemented the vertical orientation for slider.

PR #11028, opened 2024-12-11
feat(slider): Implement vertical orientation

josercarcamo:
**Related Issue:** #5522

## Summary
Implemented the vertical orientation for slider.


benelan:
What happened to #9953? Closing a PR with a lot of context in favor of a new one makes reviewing more difficult because we don't know what's been addressed and what needs further changes/discussion. 

Can you please try maintain a single branch/PR for an issue's changes instead of doing this `branch-v1`, `branch-v2`, etc. style of iteration? Thanks!

benelan:
It looks like your `v3` PR (#10652) is still open too, which one should we be reviewing?

josercarcamo:
@benelan the reason I opened other PRs is because it was very difficult to merge Lumina into the old branches. Please review the PRs for which a review is requested.

packages/calcite-components/src/components/slider/slider.stories.ts
```diff
@@ -323,107 +517,261 @@ export const WithLargeFontSize_TestOnly = (): string =>
         calcite-label {
           padding: 10px;
         }
+        calcite-slider[layout="vertical"] {
```
jcfranco, line 520: With the new vertical layout, the [default styles on the Storybook preview page](https://github.com/Esri/calcite-design-system/blob/b2c976214774490b5a313e1fafc67415fe212b37/packages/calcite-components/.storybook/preview-head.html#L11-L14) should be updated to reduce or eliminate the need for style overrides in individual vertical stories.

packages/calcite-components/src/components/slider/slider.e2e.ts
```diff
@@ -108,1219 +112,1351 @@ describe("calcite-slider", () => {
...
-      });
-      await page.waitForChanges();
-      const slider = await page.find("calcite-slider");
-      const inputEvent = await slider.spyOnEvent("calciteSliderInput");
-      const changeEvent = await slider.spyOnEvent("calciteSliderChange");
-      expect(await slider.getProperty("value")).toBe(0);
+    describe("mouse interaction", () => {
+      it("single handle: clicking the track changes value on mousedown, emits on mouseup", async () => {
+        const page = await newE2EPage({
+          html: `<calcite-slider snap style="width:${sliderWidthFor1To1PixelValueTrack}" layout=${layout}></calcite-slider>`,
```
jcfranco, line 506: This should be using `height`, right?

packages/calcite-components/src/components/slider/slider.e2e.ts
```diff
@@ -108,1219 +112,1351 @@ describe("calcite-slider", () => {
...
-      await page.waitForChanges();
 
-      const slider = await page.find("calcite-slider");
-      const changeEvent = await slider.spyOnEvent("calciteSliderChange");
-      const inputEvent = await slider.spyOnEvent("calciteSliderInput");
-      expect(await slider.getProperty("minValue")).toBe(50);
-      expect(await slider.getProperty("maxValue")).toBe(75);
+      it("single handle: clicking and dragging the track changes and emits the value", async () => {
+        const page = await newE2EPage({
+          html: `<calcite-slider snap style="width:${sliderWidthFor1To1PixelValueTrack}; margin-top: 100px" layout=${layout}></calcite-slider>`,
```
jcfranco, line 530: `margin-top` shouldn't be necessary here. This seems related to the layout issue mentioned earlier.

packages/calcite-components/src/components/slider/slider.e2e.ts
```diff
@@ -108,1219 +112,1351 @@ describe("calcite-slider", () => {
...
+          await page.mouse.move(trackX, trackY + 97);
+          await page.mouse.move(trackX, trackY + 96);
+          await page.mouse.move(trackX, trackY + 95);
+        }
+        await page.waitForChanges();
+        await page.mouse.up();
 
-      const [trackX, trackY] = await getElementXY(page, "calcite-slider", ".track");
+        expect(await slider.getProperty("value")).toBe(5);
+        expect(inputEvent).toHaveReceivedEventTimes(layout === "horizontal" ? 5 : 6);
```
jcfranco, line 558: FIXME: Could you look into this? Changing layout shouldn't affect event behavior.

packages/calcite-components/src/components/slider/slider.e2e.ts
```diff
@@ -108,1219 +112,1351 @@ describe("calcite-slider", () => {
...
+          await setUpTest(
+            `${commonSliderAttrs} min-value="${expectedValue}" max-value="${expectedValue}" mirrored layout=${layout}`,
+          );
 
-        const isMaxThumbFocused = await isElementFocused(page, `.${CSS.thumbValue}`, { shadowed: true });
+          await assertValuesUnchanged(5);
 
-        expect(isMaxThumbFocused).toBe(true);
-        await assertValuesUnchanged(5);
+          layout === "horizontal"
```
jcfranco, line 878: Can you DRY this up by storing the test x, y variables based on layout and passing them to `click`? Applies to similar test steps.

packages/calcite-components/src/components/slider/slider.e2e.ts
```diff
@@ -108,1219 +112,1351 @@ describe("calcite-slider", () => {
...
-      expect(minHandleLeft).toBe("260px");
-      expect(maxHandleRight).toBe("26px");
-    });
+    describe("when a range has 0 for both minValue and maxValue", () => {
+      const slider = `<calcite-slider
+        min="-10"
+        max="1"
+        min-value="0"
+        max-value="0"
+        layout={layout}`;
```
jcfranco, line 935: Shouldn't this be `layout=${layout}` to use the `layout` var from the `describe.each` block? This would eliminate the need to use `replace` below and keeps tests consistent.

jcfranco (changes requested):
Thanks, @josercarcamo! A few changes are needed before we can merge this.

packages/calcite-components/src/components/slider/slider.scss
```diff
@@ -56,6 +56,10 @@
 :host {
   @apply block;
 
+  &:host([layout="vertical"]) {
+    transform: rotate(-90deg);
```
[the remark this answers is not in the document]
josercarcamo: Although the content box has been rotated via the rule:

`:host([layout="vertical"]) {
  transform: rotate(-90deg);
}`

CSS interprets the properties as if the scroll bar had not been rotated, so "width" controls the height and "height" controls the width when vertical. Please advise if you know of a way to change the "orientation" of the properties as well.

packages/calcite-components/src/components/slider/slider.tsx
```diff
@@ -989,47 +1061,55 @@ export class Slider
...
-   * @param rightLabel
+   * @param leftLabel the left label element
+   * @param rightLabel the right label element
    */
   private getRangeLabelOverlap(leftLabel: HTMLSpanElement, rightLabel: HTMLSpanElement): number {
     const leftLabelBounds = leftLabel.getBoundingClientRect();
     const rightLabelBounds = rightLabel.getBoundingClientRect();
-    const leftLabelFontSize = this.getFontSizeForElement(leftLabel);
-    const rangeLabelOverlap = leftLabelBounds.right + leftLabelFontSize - rightLabelBounds.left;
+    const labelFontSize =
```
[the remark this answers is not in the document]
josercarcamo: I'm not sure how that would help as I still have to have a conditional to determine whether to pass the "leftLabel" or "rightLabel" to getFontSizeForElement. Please provide a suggestion.

packages/calcite-components/src/components/slider/slider.tsx
```diff
@@ -935,6 +968,45 @@ export class Slider
     }
   }
 
+  private hyphenateVerticalCollidingRangeHandleLabels(): void {
```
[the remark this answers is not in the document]
josercarcamo: This problem also has to do with rotating the label but not its bounding box. In other words, the label displays horizontally but the bounding box is vertical, so when the hyphen is added, the height of the bounding box is expanded, resulting in the code indicating that the handles are colliding even when visually they may not be.

packages/calcite-components/src/components/slider/slider.tsx
```diff
@@ -804,20 +813,44 @@ export class Slider
...
     const labelStaticBounds = labelStatic.getBoundingClientRect();
-    const labelStaticOffset = this.getHostOffset(labelStaticBounds.left, labelStaticBounds.right);
-    label.style.transform = `translateX(${labelStaticOffset}px)`;
-    labelTransformed.style.transform = `translateX(${labelStaticOffset}px)`;
+    const labelStaticOffset = this.getHostOffset(
+      this.layout === "horizontal" ? labelStaticBounds.left : labelStaticBounds.top,
+      this.layout === "horizontal" ? labelStaticBounds.right : labelStaticBounds.bottom,
+    );
+
+    let verticalLabelStaticOffset = 0;
```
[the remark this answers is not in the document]
josercarcamo: From my understanding, getHostOffset determines the amount by which a label has surpassed the edges of the track. "verticalLabelOffset" offsets the label from the track itself, not its edges.

packages/calcite-components/src/components/slider/slider.stories.ts
```diff
@@ -75,6 +76,8 @@ export default {
   },
 };
 
+const verticalAdjustment = "padding-top: 200px; padding-left: 100px; height: 200px";
```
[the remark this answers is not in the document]
josercarcamo: Both "padding-top" and "padding-left" are for aesthetic reasons, and they both seem to be necessary. Recall that we're trying to fit two sliders on the same page. Removing these styles results in the following:

image

packages/calcite-components/src/components/slider/slider.stories.ts
```diff
@@ -615,86 +1070,235 @@ export const fillPlacements = (): string => html`
...
-    minMaxValueLongLabelSlider.labelFormatter = function (value) {
-      return "long " + value + " label";
-    };
+    minMaxValueLongLabelSlider.forEach((e, i) => {
+      e.labelFormatter = function (value) {
+        return "long " + value + " label";
+      };
+    });
   </script>
 `;
```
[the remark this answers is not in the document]
ashetland: This is what we ended up doing in the UI Kit. It's definitely awkward because both the Label and the validation message could be massively wider than the slider. 
CleanShot 2025-01-15 at 15 01 11@2x

packages/calcite-components/src/components/slider/slider.e2e.ts
josercarcamo: No. "width = height" in vertical orientation.

packages/calcite-components/src/components/slider/slider.e2e.ts
josercarcamo: It is for aesthetics.

packages/calcite-components/src/components/slider/slider.e2e.ts
josercarcamo: The problem is that the upper-left hand corner given by "getElementXY" for horizontal is at value 0 of the slider, but it is at 100 for vertical. This causes one extra event:

image

image

packages/calcite-components/src/components/slider/slider.e2e.ts
jcfranco: This is confusing, possibly because the implementation is relying heavily on the transform. From a user-perspective, width and height should not change semantics based on an option.

Let's discuss this further today.

packages/calcite-components/src/components/slider/slider.scss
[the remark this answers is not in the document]
jcfranco: I'm not sure I follow the proposed approach, but let's discuss this further today.

packages/calcite-components/src/components/slider/slider.tsx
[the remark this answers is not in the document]
jcfranco: Gotcha. Thanks for clarifying.

packages/calcite-components/src/components/slider/slider.e2e.ts
josercarcamo: I guess I misspoke: a part of the vertical slider could be outside the bounds of the window and not be clickable, so it must be moved downward.

josercarcamo:
After meeting with @jcfranco, we've decided to seek an alternative implementation and pause this issue at this time.

PR #11030, opened 2024-12-11
fix: use explicit module imports for dayjs

jcfranco:
**Related Issue:** N/A

## Summary

This should help bundlers (e.g., Webpack) resolve dependencies due to ambiguous paths.

For context, Lumina doesn't bundle dependencies like Stencil did. `dayjs` imports were causing trouble because of this.


benelan (approved):
:+1:

PR #11038, opened 2024-12-12
fix(dialog, panel): prevent `beforeClose` from being invoked during initialization

jcfranco:
**Related Issue:** #10731 

## Summary

`willUpdate` was invoking toggling logic that was meant to be called after initialization (compare with [Stencil version](https://github.com/Esri/calcite-design-system/blame/bcc19be314a45ff584793e473e5851e15cefa83f/packages/calcite-components/src/components/panel/panel.tsx#L92)).


driskull (approved):
👏 👍

PR #11040, opened 2024-12-12
fix(dialog): restore outside click closing for non-modal dialogs

jcfranco:
**Related Issue:** #10731 

## Summary

Closes non-modal dialogs on outside click.


jcfranco:
Non-modal dialogs should not close by default when clicking outside the focus trap. Closing!

PR #11042, opened 2024-12-13
feat(notice): add component tokens

Elijbet:
**Related Issue:** #7180

## Summary
Adds the following subcomponent tokens (CSS props):

- `--calcite-notice-background-color`
- `--calcite-notice-close-background-color-focus`
- `--calcite-notice-close-background-color-press`
- `--calcite-notice-close-text-color-hover`
- `--calcite-notice-close-text-color`
- `--calcite-notice-content-text-color`


alisonailea (approved):

packages/calcite-components/src/components/notice/notice.scss
```diff
@@ -3,7 +3,13 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
- * @prop --calcite-notice-width: The width of the component.
+ * @prop --calcite-notice-background: Specifies the background color of the component.
```
macandcheese, line 6: This should be `--calcite-notice-background-color`

packages/calcite-components/src/components/notice/notice.scss
```diff
@@ -3,7 +3,13 @@
...
+ * @prop --calcite-notice-close-background-focus: Specifies the background color of the component when focused.
+ * @prop --calcite-notice-close-background-press: Specifies the background color of the component when active.
+ * @prop --calcite-notice-close-text-color-hover: Specifies the background color of component's close button when hovered.
+ * @prop --calcite-notice-close-text-color: Specifies the text color of component's close button.
+ * @prop --calcite-notice-content-text-color: Specifies the content text color of the component.
+ * @prop --calcite-notice-width: Specifies the width of the component.
```
macandcheese, line 12: I wonder if we can evaluate this for removal. Not sure why we'd need this, vs. the user being able to set this.

Elijbet: Yeah WDYT @alisonailea?

alisonailea: I agree we could set this for removal

Elijbet: @DitwanP @geospatialem is there any specific process/term we need for deprecation and removal for this or just [Deprecated] note for now is enough?

DitwanP: I think the [Deprecated] note is enough.

geospatialem: @Elijbet We have a new [`deprecate` conventional commit](https://github.com/Esri/calcite-design-system/blob/dev/CONTRIBUTING.md#:~:text=or%20missing%20test%20%F0%9F%A7%AA-,deprecate,-%3A%20Documentation%20only%20changes), so would recommend we use that approach for the deprecation so users are aware of the deprecation via our changelog.

Elijbet: I will create a follow-up PR for this.

Issue #6544, opened 2023-02-28
[List / List items] Support custom filtering supporting a filter function with a prop


nwhittaker:
### Description

A way to opt in to an stripped-down behavior when clicking a list-item that contains nested list-items.

Originally included selection and filtering asks, now placed in two separate issues:
- Filtering (_below_)
- Selection via #8515 

### Acceptance Criteria

Filtering a list of items with nested list-items:
* Parent list-item is visible when it does not match but any leaf items do match.
* Parent list-item is _not_ visible when it _does_ match but no descendent leaf items match.
* Parent list-item is not visible when no descendent leaf items match.
* Consider https://github.com/Esri/calcite-design-system/issues/3454  as part of the upcoming effort

### Which Component

calcite-list-item

### Example Use Case

The Field Maps web app would like to use the calcite-list element to display the layers in a map. For group layers, nested layers are shown as child list-items.

Selecting a leaf layer navigates the app for further handling of the layer. However, the app does not provide further handling for group layers. They are treated more like folders that organize sets of layers -- so clicking them in the list should have no effect other than to drill down to leaf layers.

### Esri team

ArcGIS Field Apps

brittneytewks:
Scheduling time with FieldApps to understand use case in more depth

nwhittaker:
Added some stripped-down filtering behavior to the issue description that should also be part of any discussion.

geospatialem:
Reallocating to the May release for design considerations.

ashetland:
Propose we model the changes to List selection behaviors after what is outlined in issue #6912 for Tree. This likely means adding a `children` selection mode and modifying the behaviors of `single`. I've started mocking this up today.

ashetland:
### Selection modes
Final proposal for revising the existing `single` and `multiple` selection-modes and adding `children` and `multichildren` can be found in this [Figma file](https://www.figma.com/file/xsoNW0ysPJQUEi8WYFwFHz/List-selection-behavior-%5Bissue-6544%5D?type=design&node-id=509-24962). See also the image below. This would be consistent with changes to Tree in issue #6912 and pr #6926.

![image](https://github.com/Esri/calcite-components/assets/108549080/e3ed207c-6887-4295-9e75-d120cabcf031)

### Hover states
Hover states for the new hit targets should follow what was also defined for Tree Item in issue #6930.

### Filtering
Since the parent would still be selectable in `children` and `multichildren`, I propose the current filtering behavior be retained for these modes.

[...]

PR #11044, opened 2024-12-13
feat(list): add filterPredicate property to handle custom filtering

driskull:
**Related Issue:** #6544

## Summary

- add `filterPredicate` property.
  - `filterPredicate: (item: ListItem["el"]) => boolean;`
- add e2e test
- cleanup 

## Example

```js
list.filterPredicate = (item) => {
  if (list.filterText === "all") {
    return true;
  }

  return item.value === "item2";
};
```


benelan (approved):
Nice! Do you think it would be helpful to add an example of `filterPredicate` to the component's demos page? We usually use a story but that doesn't work for javascript-only features yet.

driskull:
> Nice! Do you think it would be helpful to add an example of filterPredicate to the component's demos page? We usually use a story but that doesn't work for javascript-only features yet.

What about an `@example` in jsdoc? Would that work? cc @geospatialem

driskull:
@geospatialem @benelan does the example jsdoc look ok to you? Should we add more description to the property?

geospatialem:
> @geospatialem @benelan does the example jsdoc look ok to you? Should we add more description to the property?

I don't believe the `@example` will show up in our docs, but @macandcheese is this something we could add to our 3.0 doc preview branch?

driskull:
Going to merge this. We can make doc improvements in follow ups.

benelan:
Yeah LGTM! We can add the example before the doc site supports the tag like we've been doing with the custom mdn tag.

macandcheese:
> > @geospatialem @benelan does the example jsdoc look ok to you? Should we add more description to the property?
> 
> I don't believe the `@example` will show up in our docs, but @macandcheese is this something we could add to our 3.0 doc preview branch?

We don't render `@example` as they won't fit into a Table layout. We could look to display them elsewhere on a component page, or add extra, mdx-based documentation in the component pages.

PR #11046, opened 2024-12-13
fix(inline-editable): fix cancel workflow

jcfranco:
**Related Issue:** #10731 

## Summary

Fixes issue that prevented canceling an edit session and restoring the input.


driskull (approved):
Nice

Issue #11013, opened 2024-12-09
Extend `csstype` typings with custom css variables


maxpatiiuk:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

React, Lumina, Preact and other JSX libraries are using `csstype` library to provide typing for the `style={}` property.

The types are quite type-safe. However, as a side-effect they throw a type-error for usage of unknown properties, including CSS variables. Example: https://devtopia.esri.com/WebGIS/calcite-design-system/issues/172#issuecomment-5180412:

There are a number of things the consumer of Calcite's components can do to silence TypeScript (for example https://stackoverflow.com/questions/52005083/how-to-define-css-variables-in-style-attribute-in-react-and-typescript)

### Acceptance Criteria

However, it might be better long term for Calcite to extend the TypeScript interface provided by `csstype` - this way users of Calcite will get intellisence for CSS variables, and will avoid TypeScript errors.

Documentation on how to do that:
- https://github.com/frenic/csstype#what-should-i-do-when-i-get-type-errors

### Which Component

all

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

jcfranco:
@pragyamishra56 Thanks for your interest in this issue! We’ll be handling it as part of the upcoming 3.0 release. Regarding collaboration, we have a process in place, so I’d recommend checking out our [contributing guidelines](https://github.com/Esri/calcite-design-system/blob/main/CONTRIBUTING.md#i-want-to-contribute-what-should-i-work-on) for more details.

jcfranco:
@maxpatiiuk Typing custom CSS props per component might be tricky since we have composite components with their respective custom CSS props in addition to global ones. Here are a few options I see:

1. Add a catch-all interface for all custom Calcite CSS props (like [this](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/tests/commonTests/themed.ts#L202)) to mitigate migration issues related to stricter types via Lumina.
    **Pro**: simple, can be automated
    **Con**: will not catch invalid props (e.g., `--calcite-does-not-exist`)
2. Use an interface that includes all known custom CSS props (both public and internal)
[...]

PR #11051, opened 2024-12-13
fix: allow custom Calcite CSS properties in style object type

jcfranco:
**Related Issue:** #11013

## Summary

Augments `csstype`'s module to allow all custom Calcite CSS properties (step 3 from https://github.com/frenic/csstype#what-should-i-do-when-i-get-type-errors).

### DIFF
--- a/packages/calcite-components/src/index.ts
+++ b/packages/calcite-components/src/index.ts
@@ -9,7 +9,7 @@ import { setAssetPath as runtimeSetAssetPath } from "./runtime";
 
 declare module "csstype" {
   interface Properties {
-    [index: `--${string}`]: any;
+    [index: `--calcite-${string}`]: any;
   }
 }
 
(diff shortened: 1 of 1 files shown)


driskull (approved):
👍

maxpatiiuk (approved):
Looks good
Opened issue 2632 in AWC to provide longer-term solution

packages/calcite-components/src/index.ts
```diff
@@ -7,6 +7,12 @@
 import { Runtime } from "@arcgis/lumina";
 import { setAssetPath as runtimeSetAssetPath } from "./runtime";
 
+declare module "csstype" {
+  interface Properties {
+    [index: `--${string}`]: any;
```
maxpatiiuk, line 12: does this work too?

```suggestion
    [index: `--calcite-${string}`]: any;
```

jcfranco: Yep, it should.

Issue #10809, opened 2024-11-20
[Dialog] - content background should be white


ashetland:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The background of the content area is `color.background`.

### Expected Behavior

The background is `color.foreground.1`.

### Reproduction Sample

https://codepen.io/ashetland/pen/WNVqpjw?editors=100

### Reproduction Steps

Compare the reproduction sample of Dialog with [Modal](https://codepen.io/ashetland/pen/ExqBWmE?editors=1000).

### Reproduction Version

v2.13

### Relevant Info

cc @jcfranco @geospatialem 

### Priority impact

impact - p1 - need for current milestone

### Impact

Modal was deprecated in favor of Dialog, but their background colors have different defaults. Correcting this to match Modal will aid migration and will address active feedback from teams.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

driskull:
> Modal was deprecated in favor of Dialog, but their background colors have different defaults. Correcting this to match Modal will aid migration and will address active feedback from teams.

Its just using the `calcite-panel`. Why would the panel use a different background color in this instance?

macandcheese:
Prior to the change Modal behaved the same way (default to white, option to change via css property). Similarly to how we have a different default padding in Dialog than we do in Panel. 

This should ease migration from Modal while also allowing Dialog to remain flexible for the increased set of use cases it serves - where both the default (now white) or other values (grey, primarily, but depending on use case other options) can be set via the existing css property.

driskull:
I guess my concern is the inconsistency.

If we have valid use cases for when a panel should have a white background then maybe we should document them by having a prop where users can configure it rather than having to set CSS vars to a specific value.

macandcheese:
This would be a direct response to users asking for this change - honestly I don't have concerns about that level of difference and I don't think we want to limit that level of customization. This is how Modal functioned before, all we are doing is making the new component match that prior default so migrating between components is simpler. 

Making users adjust this via a property seems like an anti-pattern vs. using a css property.

[...]

PR #11052, opened 2024-12-13
fix(dialog): no longer modifies slotted panels background color

driskull:
**Related Issue:** #10809

## Summary

- set slotted panel bg color to initial color
- add story test


benelan (approved):
:+1:

jcfranco (approved):

PR #11056, opened 2024-12-13
fix: use inert on host elements instead of aria-hidden when host element is hidden

driskull:
**Related Issue:** #10731

## Summary

This will address the following Chrome console error caused by using the `ariaHidden` prop (after #10310):

> Blocked aria-hidden on an element because its descendant retained focus. The focus must not be hidden from assistive technology users. Avoid using aria-hidden on a focused element or its ancestor. Consider using the inert attribute instead, which will also prevent focus. For more details, see the aria-hidden section of the WAI-ARIA specification at https://w3c.github.io/aria/#aria-hidden.
Element with focus: span
Ancestor with aria-hidden:  <calcite-popover reference-element=​"reference-element" placement=​"auto" offset-distance=​"6" offset-skidding=​"0" text-close=​"Close" aria-hidden=​"true" aria-live=​"polite" id=​"calcite-popover-a872ffbb-85e1-5c2c-6f14-8108c4821728" role=​"dialog" overlay-positioning=​"absolute" scale=​"m" calcite-hydrated>​…​</calcite-popover>​#shadow-root (open)<div style=​"width:​ 300px;​ padding:​12px 16px;​">​…​</div>​slot</calcite-popover>

BEGIN_COMMIT_OVERRIDE
END_COMMIT_OVERRIDE


driskull:
I wasn't sure about combobox-item or icon so I left them alone.

jcfranco (approved):
🚀

jcfranco (approved):

PR #11064, opened 2024-12-16
fix(list): fix issue where last call to debounced update filter logic would override args from previous calls

jcfranco:
**Related Issue:** #10731 

## Summary

Adds internal flag to determine if filter should be performed. Similar to [`ad054fd` (#10482)](https://github.com/Esri/calcite-design-system/pull/10482/commits/ad054fd07e88b9cb304f02b7ec2006a146e10d08), this is needed because subsequent calls after `updateListItem({ performFilter: true })` would take precedence due to how [`lodash#debounce`](https://lodash.com/docs/4.17.15#debounce) works:

> The `func` is invoked with the last arguments provided to the debounced function.

BEGIN_COMMIT_OVERRIDE
END_COMMIT_OVERRIDE


benelan (approved):
Nice! Did this reproduce too inconsistently for an e2e test?

jcfranco:
@benelan Good idea. I'll update this PR if I can reproduce consistently.

PR #11071, opened 2024-12-16
fix(date-picker): fix year-select width logic

jcfranco:
**Related Issue:** #10731 

## Summary

Fixes a text width logic issue where [`font`](https://developer.mozilla.org/en-US/docs/Web/CSS/font) would be empty instead of the expected shorthand value.

This also moves the text-width calculation logic to a `requestAnimationFrame` callback to improve timing and performance.

BEGIN_COMMIT_OVERRIDE
END_COMMIT_OVERRIDE

### DIFF
--- a/packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
+++ b/packages/calcite-components/src/components/date-picker-month-header/date-picker-month-header.tsx
@@ -306,15 +306,20 @@ export class DatePickerMonthHeader extends LitElement {
   }
 
   private setYearSelectMenuWidth(): void {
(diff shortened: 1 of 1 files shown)


benelan (approved):
*shakes fist at browsers*

anveshmekala (approved):
🥇 

Tested & works as expected.

Browser might have fixed the font object issue of returning empty string. Discussed with @jcfranco  regarding this and there will be a followup spike.

jcfranco:
> Discussed with @jcfranco regarding this and there will be a followup spike.

Created follow-up spike issue: https://github.com/Esri/calcite-design-system/issues/11077

Issue #10323, opened 2024-09-16
[Dialog] `"content"` slot name / description is confusing


macandcheese:
### Description

We've received feedback the `"content"` slot name and description is confusing. 

Migrating from Modal - consumers are not changing this to the "default / unnamed" slot and have unexpected results.

### Which Component

It's confusing for a few reasons... First, the `"content"` name is used elsewhere, and it never prevents the rendering of "component chrome" like it does here. Secondly, the description of "A slot for adding custom content." is not descriptive enough.

It should explicitly mention that content in this slot will prevent the rendering of the "component chrome" (except for box-shadow and corner-radius - which may be unexpected).

Because using this slot (the custom one, not the default / recommended one) is a low-occurance use case and the name currently used is so common elsewhere, I'd propose renaming (deprecate + adding replacement) to `"custom-content"` or something equally as literal, and having the description more clearly outline the behavior when used.

### Resources

Feedback received from developer-users within Esri.

geospatialem:
Discussion should be pursued with internal staff/teams on naming for the slot.

geospatialem:
Updating the estimate to reflect the code change needed in addition to the doc update.

DitwanP:
Verified on `3.0.0-next.123`
https://codepen.io/Ditwan-Price/pen/JoPxYEY

<img width="277" alt="Image" src="https://github.com/user-attachments/assets/26866d24-4906-47a0-81cd-e21d434923f5" />

PR #11072, opened 2024-12-16
feat(dialog): add 'custom-content' slot

DitwanP:
**Related Issue:** #10323

## Summary
Deprecates the `"content"` slot in favor of a new `"custom-content"` slot that will clearly describes the behavior that a user can expect. This change should solve the issue of confusion that we've heard about from internal feedback. 

The description of the new slot would be: 
> **custom-content** - A slot for displaying custom content. Will prevent the rendering of any default Dialog UI, except for `box-shadow` and `corner-radius`.

deprecate(dialog): deprecate the `content` slot


packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -50,7 +50,8 @@ let initialDocumentOverflowStyle: string = "";
 
 /**
  * @slot - A slot for adding content.
- * @slot content - A slot for adding custom content.
+ * @slot content - [Deprecated] Use `custom-content` slot instead.
```
driskull, line 53: @DitwanP we would actually need to add this slot in the code. It not just a doc change. 

I'm kinda unsure how this change would work without being breaking since we put the panel inside the `content` slot by default.

driskull (changes requested):

packages/calcite-components/src/components/dialog/dialog.tsx
DitwanP: @driskull Yeah good point. That being said, what are your thoughts on changing the slot to the new one overall? Are you in favor or not in favor of such a change?

driskull: I'm in favor of it, i'm just not sure how it will work exactly.

macandcheese: Can this coincide with the rest of 3.0 breaking changes?

driskull: Thats a question for @jcfranco

jcfranco: Good catch. Unfortunately, it's too late to include in 3.x. Could we update the description for this release and schedule the slot rename for 4.0? cc @geospatialem

macandcheese: Could we deprecate (w/o removing) "content" slot and add "custom-content"? That wouldn't be breaking as the recommendation is to use the "unnamed" default slot, right?

driskull: The problem is we put default content in the `content` slot now. So we don't have a good way to counter that.

driskull: Basically, we can't have this in both the `content` and `custom-content` slots or it would show up twice. 

```tsx
   <slot name={SLOTS.content}>
            <calcite-panel
              beforeClose={this.beforeClose}
              class={CSS.panel}
              closable={!this.closeDisabled}
              closed={!opened}
```

I don't think we can do this right?

```tsx
<slot name="custom-content">
            <slot name={SLOTS.content}>
```

macandcheese: That... does seem to work pretty well, in local testing...

driskull: Well i'll be... 🤯

DitwanP: Finally got around to making this change, LMK if I missed anything.

driskull (changes requested):
Code changes look good.

Can we add a screenshot test and also test this across all browser to make sure it works?

Can we change the PR title to:

`feat(dialog): add 'custom-content' slot and deprecate 'content' slot`

DitwanP:
Looks to be working in all browsers:
image

driskull (approved):
👍 Nice! Good to know that nesting slots like this works.

PR #11073, opened 2024-12-16
fix(list, sort-handle): disable sort-handle when no move to items are present and item count is one

driskull:
**Related Issue:** #10731

## Summary

- disable drag handle when a single item is in the list and the list has nowhere else to move to
- add e2e
- add story

BEGIN_COMMIT_OVERRIDE
END_COMMIT_OVERRIDE

### DIFF
--- a/packages/calcite-components/src/components/sort-handle/sort-handle.tsx
+++ b/packages/calcite-components/src/components/sort-handle/sort-handle.tsx
@@ -73,7 +73,7 @@ export class SortHandle extends LitElement implements LoadableComponent, Interac
   @property() messages = useT9n<typeof T9nStrings>({ blocking: true });
 
   /** Defines the "Move to" items. */
-  @property() moveToItems: MoveTo[];
+  @property() moveToItems: MoveTo[] = [];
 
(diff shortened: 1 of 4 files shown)


packages/calcite-components/src/components/sort-handle/sort-handle.tsx
```diff
@@ -254,10 +254,12 @@ export class SortHandle extends LitElement implements LoadableComponent, Interac
       setPosition,
       setSize,
       widthScale,
+      moveToItems,
     } = this;
     const text = this.getLabel();
 
-    const isDisabled = disabled || !setPosition || !setSize;
+    const isDisabled =
```
jcfranco, line 261: Is the optional chaining operator on `moveToItems` needed? It looks like it's guaranteed now, but I could be missing something.

jcfranco (approved):
🚀

packages/calcite-components/src/components/sort-handle/sort-handle.tsx
driskull: Its not but I added for extra safety. I can remove it.

PR #11075, opened 2024-12-16
fix(action): component in transparent active mode should use -hover

alisonailea:
**Related Issue:** #10982

## Summary

Update the background color of the component when active=true and appearance=transparent to use the transparent-hover instead of transparent-press.


ashetland (approved):
LGTM!

PR #11082, opened 2024-12-17
fix(accordion-item): stretch slotted actions only

jcfranco:
**Related Issue:** #10731 

## Summary

This avoids unintentionally styling other actionable components and offers more flexibility.

**Note**: this changes styles to use explicit selectors instead of universal ones to improve maintainability.

### DIFF
--- a/packages/calcite-components/src/components/accordion-item/accordion-item.scss
+++ b/packages/calcite-components/src/components/accordion-item/accordion-item.scss
@@ -95,8 +95,11 @@
 .header {
   @apply flex items-stretch;
 
-  * {
-    @apply inline-flex
+  &-content,
+  &-container,
+  .actions-start,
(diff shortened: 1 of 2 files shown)


jcfranco:
Alternatively, we could also tweak `switch` (and similar ones), so that it centers when stretched vertically.

driskull (approved):
👍

macandcheese:
> Alternatively, we could also tweak `switch` (and similar ones), so that it centers when stretched vertically.

This could be a nice fallback for all non-action elements. A user can always slot a div, unset styles - but I think center align is a more expected default than align-top?

jcfranco:
@macandcheese Fortunately, slotted, non-action items are already centered, so no further changes needed! 🎉

PR #11085, opened 2024-12-17
refactor(t9n): drop deprecated and obsolete `useNoForNorwegian` arg

jcfranco:
**Related Issue:** N/A

## Summary

`@arcgis/lumina@4.32.0-next.52` uses `no` for Norwegian by default, so we can drop this argument.


driskull (approved):
👍

Issue #10959, opened 2024-12-03
[Input] suffixText & prefixText style issues when set via property


driskull:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The input styles rely on the `suffix-text` attribute, however the property isn't reflected so the attribute will not be there if the property is used.

### Expected Behavior

The styles work for both attributes and properties of suffixText, prefixText

### Reproduction Sample

https://codepen.io/driskull/pen/YPKyrJL?editors=1010

### Reproduction Steps

1. open codepen
2. notice the input set via property looks different than the one set via attributes

### Reproduction Version

3.0.0-next.44

### Relevant Info

The following style selector also looks to be incorrect:

```scss
:host([suffix-text]) input,
.element-wrapper:has(.clear-button) input,
:host([number-button-type="vertical"][type="number"]) input,
:host([suffix-text][number-button-type="horizontal"]) .suffix,
:host([suffix-text][number-button-type="vertical"][type="number"]) .suffix,
:host([number-button-type="vertical"][type="number"])
  .clear-button
  :host([number-button-type="horizontal"][type="number"])
  .clear-button {
  border-start-end-radius: 0;
  border-end-end-radius: 0;
}
```

### Priority impact

impact - p1 - need for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

geospatialem:
Verified in `3.0.0-next.71`

![image](https://github.com/user-attachments/assets/45fd1557-526e-4a98-b963-2c3d6a22f909)

Issue #10958, opened 2024-12-03
Find a way to test assigning values via properties and attributes


driskull:
**monday.com sync:** #18048698469

### Background

Not all properties are reflected to the DOM, however we have a lot of CSS that relies on attribute selectors on the host.

If a property isn't reflected but a attribute selector is used in styles, we may not catch the rendering issue since all of our tests and stories rely on attributes

We need to figure out a way to test components by assigning via attributes and properties

Example:

Input.scss has the following CSS:

```scss
:host([prefix-text]) input {
  border-start-start-radius: 0;
  border-end-start-radius: 0;
}
```

However, the property `prefixText` is not reflected.

The HTML examples use attributes and the screenshot tests use attributes so this error would not be caught if assignign prefixText via a property.

### Desired Outcome

Ensuring that assigning values via properties renders the component correctly.

Related to #10959.

PR #11089, opened 2024-12-17
fix(input): Avoid styling on non-reflected attributes

macandcheese:
**Related Issue:** #10959 #10958

## Summary
Adds classes to use for specific styling cases in Input. This only adjusts `suffix-text` and `prefix-text`. There are other styles set on attributes in the component, but those are reflected attributes. We could still improve that in the future if needed - and / or, in the newer `input-x` components.


driskull (approved):
🎉 🌮

PR #11090, opened 2024-12-17
fix(input-date-picker, modal, tile-select): avoid using refs that have been nulled

jcfranco:
**Related Issue:** #10731 

## Summary

Adds guards for refs that may be nulled. Unlike Stencil, Lumina calls `ref` when the component is disconnected, so this addresses some issues caused by logic running after an element is no longer connected.

BEGIN_COMMIT_OVERRIDE
END_COMMIT_OVERRIDE


driskull (approved):
🧹 👍

PR #11095, opened 2024-12-18
fix: floating ui elements set max-content for inline-size

driskull:
**Related Issue:** #10731

## Summary

- Follows [`floating-ui`'s initial layout recommendation](https://floating-ui.com/docs/computeposition#initial-layout).
- Combobox wasn't correctly handling the size when in fixed overlay positioning
  - Example: https://codepen.io/Matt-Driscoll/pen/raByNmQ?editors=1010
- updated tooltip and popover stories where stories were being cut off due to sizing.

BEGIN_COMMIT_OVERRIDE
END_COMMIT_OVERRIDE

### DIFF
--- a/packages/calcite-components/src/components/popover/popover.stories.ts
+++ b/packages/calcite-components/src/components/popover/popover.stories.ts
@@ -124,7 +124,7 @@ export const flipPlacements_TestOnly = (): string => html`
 `;
 
 export const scaleConsistencyPopoverHeadingActionSlottedIcon_TestOnly = (): string => html`
-  <div style="width: 400px;">
+  <div style="width: 400px; height:400px;">
     ${referenceElementHTML}
     <calcite-popover
       heading="Dreams didn't make us kings. Dragons did. 🐉"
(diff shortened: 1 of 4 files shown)


packages/calcite-components/src/components/popover/popover.stories.ts
```diff
@@ -124,18 +124,20 @@ export const flipPlacements_TestOnly = (): string => html`
...
-      heading="Dreams didn't make us kings. Dragons did. 🐉"
-      reference-element="reference-element"
-      placement="auto"
-      open
-      closable
-      scale="m"
-    >
-      ${contentHTML}
-    </calcite-popover>
+  <div style="width: 800px; height:800px;">
```
jcfranco, line 127: Sidebar: do you think adding a shared set of storybook decorators (similar to [this](https://www.chromatic.com/docs/troubleshooting-snapshots/#use-fixed-height-wrappers-for-portal-components) or [this](https://www.chromatic.com/docs/troubleshooting-snapshots/#why-isnt-my-modal-or-dialog-captured)) to normalize the dimensions of similar stories would be useful?

jcfranco (approved):
✨🚀✨

packages/calcite-components/src/components/popover/popover.stories.ts
driskull: Yes that would be very helpful. We already do variations of this for some components.

jcfranco: Sweet. I'll create an issue for this.

PR #11098, opened 2024-12-18
chore(linting): enable radix rule

jcfranco:
**Related Issue:** N/A

## Summary

Enables the [radix](https://eslint.org/docs/latest/rules/radix) rule to ban unnecessary usage of `radix` when using `parseInt`:

> On the other hand, if the code is targeting only ES5-compliant environments passing the radix 10 may be redundant. In such a case you might want to disallow using such a radix.


driskull (approved):
👍

PR #11099, opened 2024-12-18
fix(autocomplete): handle focusing when item is clicked and reset inputValue on form reset

driskull:
**Related Issue:** #10044

## Summary

- opens autocomplete when input is clicked
  - This is useful for when the input is already focused but the autocomplete is closed.
- Sets focus and sets `open=false` when `autocomplete-item` is clicked
  - ensures that the input receives focus after an `autocomplete-item` is clicked
- use `enterKeyHint` property JSX
- update `onFormReset` to be called after form component logic
  - This allows us to reset the `inputValue` to the `defaultInputValue` without overriding existing logic
- adds autocomplete to validation demo

cc @benelan 

BEGIN_COMMIT_OVERRIDE
END_COMMIT_OVERRIDE

### DIFF
--- a/packages/calcite-components/src/demos/validation.html
+++ b/packages/calcite-components/src/demos/validation.html
@@ -281,6 +281,59 @@
                 <calcite-rating required name="rating"></calcite-rating>
               </calcite-label>
             </li>
+
+            <!-- Autocomplete -->
+            <li>
+              <calcite-label>
+                Which framework do you use?
+                <calcite-autocomplete name="framework" required id="framework">
+                  <calcite-autocomplete-item-group heading="Frameworks">
+                    <calcite-autocomplete-item
+                      label="VanillaJS"
+                      value="vanilla"
+                      heading="VanillaJS"
+                      description="VanillaJS with @esri/calcite-components"
+                      icon-start="ribbon"
+                    ></calcite-autocomplete-item>
+                    <calcite-autocomplete-item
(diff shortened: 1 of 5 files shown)


driskull:
@benelan for `onFormReset` it looks like you can override it and provide your own logic but not add additional logic on top of it.

Can we either:

- make the default form reset functionality an exported function
- or make the `component.onFormReset` call on top of the default functionality?

packages/calcite-components/src/demos/validation.html
```diff
@@ -457,6 +510,12 @@ <h1 style="margin: 0 auto; text-align: center">Form Validation</h1>
       datePicker.maxAsDate = new Date();
       datePicker.min = "2021-01-01";
 
+      // Autocomplete
+      const autocomplete = document.getElementById("framework");
+      autocomplete.addEventListener("calciteAutocompleteChange", (event) => {
+        autocomplete.inputValue = event.target.value;
```
driskull, line 516: This is for demonstration purposes only. We don't update the inputValue when a value has been selected. This is because we don't know what data would be best to put into the input. A value may be a unique id, a description or label may be better. Its best to leave this up to the end user to decide if they want to populate the input or not. In some cases, it may be best to leave the input as is with whatever the user has typed in so they can adjust their query.

benelan (approved):
Nice!

Issue #8967, opened 2024-03-19
Provide CSS variable for input placeholder color


ellenupp:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Because the input background color can be changed using `--calcite-color-foreground-1`, there should be a way to change the placeholder text color for general theming purposes as well as to maintain a sufficient contrast ratio. From https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum#:~:text=including%20placeholder%20text, placeholder text is included in the minimum contrast Success Criterion (1.4.3). Against foreground-1 there are some failures in some browser default/mode combinations, for example: 

Chrome, dark mode: #757575 against #2B2B2B = [3.07:1](https://webaim.org/resources/contrastchecker/?fcolor=2B2B2B&bcolor=757575)
Safari, light mode: #A9A9A9 (`darkgray`) against #FFFFFF = [2.35:1](https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=A9A9A9)

(Note: To check the placeholder color in the user agent stylesheet, inspect an input when “Show user agent shadow DOM” is enabled in dev tools settings > preferences > elements (Chrome)).

### Acceptance Criteria

It is possible to set the placeholder color in an input, for example 
`--calcite-color-input-placeholder: #6d6d88`
or maybe with an existing variable like `--calcite-color-text-3`, which would be applied as the color in the `::placeholder` pseudo-element. 

### Which Component

calcite-input, calcite-input-*

### Priority impact

p2 - want for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Dashboards

macandcheese:
Good news! This will be added as a part of the larger component-specific design token effort (#7180) - the PR that added this request can be found here: https://github.com/Esri/calcite-design-system/pull/8767

ellenupp:
That's great, thanks!

DitwanP:
@alisonailea the placeholder text color token doesn't seem to be working while I'm testing locally. Are you seeing the same behavior?
![image](https://github.com/user-attachments/assets/7fe216a5-7cf3-4913-b016-ed43fe9905d5)

[...]

PR #11107, opened 2024-12-18
fix(input): correctly apply placeholder styles

alisonailea:
**Related Issue:** #8967 

## Summary

Breaks the placeholder styles into separate lines to avoid a browser specific selector erroring out the whole line. 

Screenshot 2024-12-18 at 3 21 01 PM

BEGIN_COMMIT_OVERRIDE
END_COMMIT_OVERRIDE

### DIFF
--- a/packages/calcite-components/src/components/input/input.scss
+++ b/packages/calcite-components/src/components/input/input.scss
@@ -33,6 +33,11 @@
  *
  */
 
+@mixin input-placeholder-styles {
+  @apply font-normal;
+  color: var(--calcite-input-placeholder-text-color, var(--calcite-color-text-3));
+}
+
 :host {
   @apply block;
 }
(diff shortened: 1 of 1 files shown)


macandcheese (approved):
⭐

packages/calcite-components/src/components/input/input.scss
```diff
@@ -155,17 +155,30 @@ input {
...
-    color: var(--calcite-input-placeholder-text-color, var(--calcite-color-text-3));
-  }
+
   &:placeholder-shown {
     @apply text-ellipsis;
   }
 }
 
+// Placeholder styles
+// Do not combine these style rules. The browser interprets the rule as one. Properties like -webkit can not be found in Firefox, thus the whole line is rendered as an error.
```
jcfranco, line 165: Nitpick: for clarity, maybe use one of the placeholder vendor prefixes instead of `-webkit`?

jcfranco, line 165: > thus the whole line is rendered as an error.

Suggestion: `thus the whole selector is invalidated and is ignored.`

packages/calcite-components/src/components/input/input.scss
```diff
@@ -155,17 +155,30 @@ input {
...
+:-ms-input-placeholder {
+  @apply font-normal;
+  color: var(--calcite-input-placeholder-text-color, var(--calcite-color-text-3));
```
jcfranco, line 168: Can we use a mixin or placeholder to DRY these styles a bit? This can be done in a follow-up.

jcfranco (approved):

PR #11109, opened 2024-12-19
fix(list): Don't require `filterEnabled` for `filterPredicate` usage

driskull:
**Related Issue:** #6544

## Summary

- fix to not require filterEnabled to use filterPredicate
- It was basically the `(changes.has("filterPredicate") && this.hasUpdated)` that was the culprit.
- add test

BEGIN_COMMIT_OVERRIDE
END_COMMIT_OVERRIDE


benelan (approved):
:+1:

PR #11118, opened 2024-12-20
fix(input, input-text, input-number): restore handling of autofocus global attribute

jcfranco:
**Related Issue:** #10731 

## Summary

Fixes an issue caused by #10310, where `autofocus` was no longer used within rendering.


driskull (approved):
👍 nice catch 🐟

PR #11121, opened 2024-12-20
refactor(autocomplete, input, input-number, input-text): improve types

jcfranco:
**Related Issue:** N/A

## Summary

This improves existing types to make way for Lumina type enhancements.

**Note**: this also removes [some workarounds](https://github.com/Esri/calcite-design-system/pull/9245#discussion_r1589565315) related to previous global attribute handling.

### DIFF
--- a/packages/calcite-components/src/components/input-text/input-text.tsx
+++ b/packages/calcite-components/src/components/input-text/input-text.tsx
@@ -137,9 +137,9 @@ export class InputText
    * Specifies the type of content to autocomplete, for use in forms.
    * Read the native attribute's documentation on MDN for more info.
    *
-   * @mdn [step](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete)
+   * @mdn [autocomplete](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete)
    */
-  @property() autocomplete: string;
+  @property() autocomplete: AutoFill;
 
   /** When `true`, a clear button is displayed when the component has a value. */
   @property({ reflect: true }) clearable = false;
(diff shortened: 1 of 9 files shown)


packages/calcite-components/src/components/input-text/input-text.tsx
```diff
@@ -585,24 +585,16 @@ export class InputText
...
-        autocomplete={this.autocomplete as LuminaJsx.HTMLElementTags["input"]["autocomplete"]}
-        autofocus={this.el.autofocus ? true : null}
+        autocomplete={this.autocomplete}
+        autofocus={this.el.autofocus}
         class={{
           [CSS.editingEnabled]: this.editingEnabled,
           [CSS.inlineChild]: !!this.inlineEditableEl,
         }}
         defaultValue={this.defaultValue}
         disabled={this.disabled ? true : null}
```
maxpatiiuk, line 595: does this still need to have `? true : null`? is that expected or is it a bug I should fix?

```suggestion
        disabled={this.disabled}
```

maxpatiiuk (approved):

driskull (approved):
👍

packages/calcite-components/src/components/input-text/input-text.tsx
jcfranco: I'll test for that separately to keep this PR focused.

packages/calcite-components/src/components/input/common/tests.ts
```diff
@@ -128,11 +128,26 @@ export function testWorkaroundForGlobalPropRemoval(
...
-    const input = await page.find(`${inputTag} >>> input`);
+    const internalInput = await page.find(`${inputTag} >>> input`);
+
+    expect(internalInput.getAttribute("autofocus")).toBe("");
+    expect(internalInput.getAttribute("inputmode")).toBe(testInputMode);
+    expect(internalInput.getAttribute("enterkeyhint")).toBe(testEnterKeyHint);
+
+    const input = await page.find(inputTag);
+
+    // we intentionally teast each one to avoid renders caused by unrelated props affecting result
```
maxpatiiuk, line 139:
```suggestion
    // we intentionally test each one to avoid renders caused by unrelated props affecting result
```

packages/calcite-components/src/components/input/common/tests.ts
```diff
@@ -150,5 +165,18 @@ export function testWorkaroundForGlobalPropRemoval(
     expect(internalInput.getAttribute("autofocus")).toBe("");
     expect(internalInput.getAttribute("inputmode")).toBe(testInputMode);
     expect(internalInput.getAttribute("enterkeyhint")).toBe(testEnterKeyHint);
+
+    // we intentionally teast each one to avoid renders caused by unrelated props affecting result
```
maxpatiiuk, line 169:
```suggestion
    // we intentionally test each one to avoid renders caused by unrelated props affecting result
```

maxpatiiuk (approved):

packages/calcite-components/src/components/input/common/tests.ts
jcfranco: Good catch. Thanks!

PR #10512, opened 2024-10-09
feat: add default focus color token

alisonailea:
**Related Issue:** #7180 #10510

## Summary

This adds `calcite-color-focus` to replace the deprecated `calcite-ui-focus-color`

PR #11130, opened 2024-12-21
fix(list-item): ensure consistent focus border color by referencing updated tokens

jcfranco:
**Related Issue:** #10731 

## Summary

Updates `list-item` to use the default focus color token introduced in https://github.com/Esri/calcite-design-system/pull/10512.


benelan (approved):
:+1:

Issue #11093, opened 2024-12-17
Clean up ref-handling logic


jcfranco:
**monday.com sync:** #18005817758

### Description

Lumina calls `ref` when a component is disconnected, so #10310 mainly focused keeping refs and returning early if nulled. We should revisit these areas to see if we can improve teardown of cleared refs.

### Proposed Advantages

Decrease chances of memory leaks.

### Which Component

Any component using `ref`.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

DitwanP:
👍 Verified with help from Franco.

"Ref callbacks now use Refs where possible, avoiding anonymous functions, and properly unset reference elements instead of returning early when undefined."

PR #11132, opened 2024-12-22
fix(popover, tooltip): skip ref setup logic on component removal

jcfranco:
**Related Issue:** #10731

## Summary

Adds a guard to `popover` and `tooltip` `ref` callbacks to avoid missing ref warnings on removal. After #10310, `ref` callbacks are invoked both when the component is added and removed (see https://github.com/Esri/calcite-design-system/issues/11093).

BEGIN_COMMIT_OVERRIDE
omitted from changelog
END_COMMIT_OVERRIDE


packages/calcite-components/src/components/popover/popover.e2e.ts
```diff
@@ -729,6 +729,54 @@ describe("calcite-popover", () => {
     });
   });
 
+  describe("warning messages", () => {
```
jcfranco, line 732: I'll submit a follow-up PR to clean this up and also try to add some reusable tests/utils.

driskull (approved):
👍

PR #11139, opened 2024-12-23
fix(action-menu): fix toggle logic when `action-menu` is reconnected

jcfranco:
**Related Issue:** #10731 

## Summary

Fixes a regression caused by #10310 where the change to the internal popover to use `triggerDisabled` would no longer work when the component was reconnected.

### DIFF
--- a/packages/calcite-components/src/components/action-menu/action-menu.tsx
+++ b/packages/calcite-components/src/components/action-menu/action-menu.tsx
@@ -319,6 +319,8 @@ export class ActionMenu extends LitElement implements LoadableComponent {
       "keydown",
       this.menuButtonKeyDown,
     ) /* TODO: [MIGRATION] If possible, refactor to use on* JSX prop or this.listen()/this.listenOn() utils - they clean up event listeners automatically, thus prevent memory leaks */;
+
+    this.menuButtonEl = null;
   }
 
   private setMenuButtonEl(event: Event): void {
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/action-menu/action-menu.tsx
```diff
@@ -334,7 +336,10 @@ export class ActionMenu extends LitElement implements LoadableComponent {
 
   private setDefaultMenuButtonEl(el: Action["el"]): void {
     this.defaultMenuButtonEl = el;
-    this.connectMenuButtonEl();
+
+    if (el) {
```
driskull, line 340: Seems like a lot of components might need this kind of if statement since this is similar to some other fixes 👍

packages/calcite-components/src/components/action-menu/action-menu.tsx
```diff
@@ -319,6 +319,8 @@ export class ActionMenu extends LitElement implements LoadableComponent {
       "keydown",
       this.menuButtonKeyDown,
     ) /* TODO: [MIGRATION] If possible, refactor to use on* JSX prop or this.listen()/this.listenOn() utils - they clean up event listeners automatically, thus prevent memory leaks */;
+
+    this.menuButtonEl = null;
```
jcfranco, line 323: Not clearing this reference would cause `connectMenuButtonEl` to bail since the references would [remain the same](https://github.com/Esri/calcite-design-system/blob/jcfranco/clear-action-menu-el-on-disconnect/packages/calcite-components/src/components/action-menu/action-menu.tsx#L266-L268).

driskull (approved):
👍

packages/calcite-components/src/components/action-menu/action-menu.tsx
jcfranco: Agreed. I created https://github.com/Esri/calcite-design-system/issues/11093 to take a closer look, but it might be easier for the time being to return early when the reference element is null in all cases (preserves existing component logic based on Stencil lifecycle). WDYT?

driskull: Yeah I think that might be the easiest and most consistent solution.

Issue #11138, opened 2024-12-23
[Input Time Zone] missing description for .setFocus()


maxpatiiuk:
### Description

Many Calcite components have `.setFocus()` method.
Almost all of those methods have a JSDoc description.

However, JSDoc description is missing in calcite-input-time-zone: ![Screenshot 2024-12-23 at 12 32 01](https://github.com/user-attachments/assets/8637f404-863c-499b-83fd-92872d52e16c)

### Which Component

calcite-input-time-zone

### Resources

Related: I am considering adding detection of public API without description in Lumina. Feel free to chime in on the issue if you have feedback/ideas: https://devtopia.esri.com/WebGIS/arcgis-web-components/issues/354

geospatialem:
Verified locally with `3.0.0-next.82` 📜

PR #11141, opened 2024-12-23
docs(input-time-zone): add setFocus method description

geospatialem:
**Related Issue:** #11138

## Summary
Add a description to `input-time-zone`'s `setFocus` method.


macandcheese (approved):
⭐

Issue #11128, opened 2024-12-21
`input-time-zone` can lose selection on change


nwhittaker:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Given a time-zone input this is updated programmatically: If the input's `messages`, `mode`, and/or `referenceDate` props are set at the same time the `value` prop is, the matching combobox-item (if there is one) is not selected or loses its selection.

### Expected Behavior

The matching combobox-item, if any, is selected or remains selected.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/XJrRgeB

### Reproduction Steps

1. Visit the repro. sample
2. If needed, change the `offset` passed to the `setTimeZone()` function to be different from your own.
3. Click the **Set time zone** button and observe the `<calcite-input-time-zone>` input's value is cleared instead of matching the offset established in step 2. If it appears to work, try clicking the button a few times and/or clicking the input itself to force a repaint.
4. Optionally comment out the line of code that sets the input's `referenceDate` to see the expected behavior.

### Reproduction Version

2.13.2

### Relevant Info

I believe the issue stems from a race condition in the component's `handleTimeZoneItemPropsChange` method:

https://github.com/Esri/calcite-design-system/blob/e38e2977d93bc3d55b50dc90450f8414aa02ddaa/packages/calcite-components/src/components/input-time-zone/input-time-zone.tsx#L145-L146

The time-zone selection is not waiting for the new list of time-zones. As a result it's finding itself in the old list of time zones which means it never matches a time-zone item in the `renderItems()` method.

I suspect making the `handleTimeZoneItemPropsChange` method asynchronous and awaiting the call to `updateTimeZoneItems()` would be enough to clear up the issue:

```diff
-  handleTimeZoneItemPropsChange(): void {
+  async handleTimeZoneItemPropsChange(): Promise<void> {
    if (!this.timeZoneItems) {
      return;
    }

-    this.updateTimeZoneItems();
+    await this.updateTimeZoneItems();
    this.updateTimeZoneSelection();
  }
```

### Regression?

2.12.2

### Priority impact

impact - p1 - need for current milestone

### Impact

[...]

PR #11166, opened 2024-12-29
fix(input-time-zone): keep selection in sync when item-related props are set along with value

jcfranco:
**Related Issue:** #11128 

## Summary

Fixes an issue caused by having selection-related logic run before items are updated.

### Noteworthy changes

* removes `async` from synchronous method
* moves internal value-setting logic earlier to ensure proper state


driskull (approved):
👍

PR #11179, opened 2024-12-31
ci(renovate): enable followTag for @arcgis prereleases

jcfranco:
**Related Issue:** N/A

## Summary

[`followTag`](https://docs.renovatebot.com/configuration-options/#followtag) should help bump prerelease versions in `package.json` instead of only updating the lock file, which could cause different versions to be installed (see https://github.com/Esri/calcite-design-system/pull/11174).

**Note**: we should remove this option once `@arcgis` packages are stable. cc @benelan

### DIFF
--- a/.renovaterc.json
+++ b/.renovaterc.json
@@ -26,7 +26,8 @@
   "packageRules": [
     {
       "groupName": "ArcGIS",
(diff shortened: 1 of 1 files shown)


driskull (approved):
👍

Issue #11187, opened 2025-01-02
[color-picker] shorthand RGB hex should be accepted with alpha-channel


jcfranco:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Shorthand RGB hex is no longer expanded when the input is blurred.

### Expected Behavior

Shorthand hex should be accepted regardless of alpha channel being present or not.

### Reproduction Sample

https://codepen.io/jcfranco/pen/pvzWmaq?editors=100

### Reproduction Steps

1. click on hex input
1. select all text
1. type any shorthand rgb hex (e.g., abc)
1. blur input
1. notice original value is restored

### Reproduction Version

2.13.2

### Relevant Info

Looks like this was introduced in https://github.com/Esri/calcite-design-system/pull/9701.

### Regression?

2.10.1

### Priority impact

impact - p3 - not time sensitive

### Impact

This might affect users that assume blurring inputs will expand shorthand hex values.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

geospatialem:
Verified on the `dev` branch via keyboard and mouse

![verify-hex](https://github.com/user-attachments/assets/1b83d64c-0dc8-4f2d-8b84-73172d1c3591)

PR #11188, opened 2025-01-02
fix(color-picker): ensure shorthand hex is expanded regardless of alpha channel presence

jcfranco:
**Related Issue:** #11187 

## Summary

Fixes a regression from https://github.com/Esri/calcite-design-system/pull/9701, where only RGBA shorthand hex values would be accepted by `calcite-color-picker-hex-input`.


aPreciado88 (approved):
🚀

PR #11190, opened 2025-01-02
ci(renovate): use bump range strategy instead of `followTag` for @arcgis prereleases

jcfranco:
**Related Issue:** N/A

## Summary

Using `bump` [range strategy](https://docs.renovatebot.com/configuration-options/#rangestrategy) along with minor `@arcgis` instead of [`followTag`](https://docs.renovatebot.com/configuration-options/#followtag), as the latter follows versions strictly, making minor versions no longer allowed. 
versions strictly, thus no longer allowing minor versions.

cc @maxpatiiuk

### DIFF
--- a/.renovaterc.json
+++ b/.renovaterc.json
@@ -27,7 +27,7 @@
     {
       "groupName": "ArcGIS",
       "matchPackageNames": ["@arcgis/**"],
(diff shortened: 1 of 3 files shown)


driskull (approved):
👍

Issue #11154, opened 2024-12-27
[Autocomplete] highlight text matching the inputValue


driskull:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

bold and highlight autocomplete-items that match the inputValue

### Acceptance Criteria

autocomplete items that match the input value are highlighted in some form

### Which Component

autocomplete

### Example Use Case

Something like this

<img width="303" alt="image" src="https://github.com/user-attachments/assets/37a484b6-b6c4-43bc-8d7e-7985b8208afe" />

https://codepen.io/Matt-Driscoll/pen/wBwqpye?editors=1010

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

driskull:
@ashetland @SkyeSeitz Can you help out with the design here? I can copy what combobox has but to me it seems a bit too much with the background color highlight.

ashetland:
I think we could simplify this to just the weight increase. We should also update Combobox to match if we move forward with that. I think the styling in Combobox stemmed from work on Input Time Zone, but I think the simplified styling would work there too. cc @jcfranco @emmilaakso212

driskull:
Thanks @ashetland, ill proceed with just the bold weight font. Can you create issues to update the other components and figma?

geospatialem:
Verified the bold font weight with `3.0.0-next.90`.

PR #11191, opened 2025-01-02
fix(autocomplete): fix item key functionality

driskull:
**Related Issue:** #11154

## Summary

Only honor navigation key functionality when items are present

BEGIN_COMMIT_OVERRIDE
END_COMMIT_OVERRIDE


jcfranco (approved):
LGTM! Before merging, can you add/update a test to cover this?

Issue #11173, opened 2024-12-30
[Combobox, Input Time Zone] - simplify matching text highlight styling


ashetland:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

For consistency across components, we could simplify the matching text styling to remove the bg color change:
![CleanShot 2024-12-30 at 10 28 07@2x](https://github.com/user-attachments/assets/524b41e0-927d-4e42-ac92-a8e9b25918bc)

### Acceptance Criteria

Matching text only has a font weight increase without the background color.

### Relevant Info

Related to #11154 

cc @driskull @jcfranco @SkyeSeitz @macandcheese @matgalla 

**Note**: this should also leverage utils added in https://github.com/Esri/calcite-design-system/pull/11155 and also add a shared mixin if styles align as well (see https://github.com/Esri/calcite-design-system/pull/11155#discussion_r1899848478).

### Which Component

Combobox, Input Time Zone

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

jcfranco:
I'm all for simplifying, but I’d like to get @emmilaakso212's thoughts on this, as the `combobox`/`input-time-zone` item match background color change is based on their design specs.

emmilaakso212:
@ashetland is this due to the impending style changes for selection in Calcite? I'm ok with it as long as the font weight increase is from regular/text to bold. Here's a quick mock: 
![Screenshot 2025-01-02 at 3 40 43 PM](https://github.com/user-attachments/assets/c09b1b88-4535-4c9a-907b-22c923773a11)

ashetland:
@emmilaakso212 Yes. This simplification was also requested for Autocomplete (#11154) and it makes sense to be consistent across components.

geospatialem:
Verified in `3.0.0-next.89`:

![image](https://github.com/user-attachments/assets/e417c6d1-3274-42fc-8170-ed6f2e4bf7cf)

![image](https://github.com/user-attachments/assets/e46bccb3-3bb0-4028-aaf3-3f44c3147ec9)

PR #11193, opened 2025-01-02
feat(combobox, input-time-zone): highlight text matches consistently

jcfranco:
**Related Issue:** #11173 

## Summary

This also introduces a mixin for shared text highlighting styles.

### DIFF
--- a/packages/calcite-components/src/assets/styles/includes.scss
+++ b/packages/calcite-components/src/assets/styles/includes.scss
@@ -215,7 +215,8 @@ $common-animatable-props: "background-color, block-size, border-color, box-shado
 // - this should be used in conjunction with the `text.tsx` util.
 @mixin text-highlight-item() {
   .text-match {
-    font-weight: var(--calcite-font-weight-bold);
     background-color: transparent;
+    color: inherit;
+    font-weight: var(--calcite-font-weight-bold);
   }
 }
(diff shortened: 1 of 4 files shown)


packages/calcite-components/src/assets/styles/includes.scss
```diff
@@ -210,3 +210,12 @@ $common-animatable-props: "background-color, block-size, border-color, box-shado
   transition-duration: var(--calcite-animation-timing);
   transition-timing-function: ease-in-out;
 }
+
+// Mixin for text highlighting styles.
+// - this should be used in conjunction with the `text.tsx` util.
+@mixin text-highlight-item() {
+  .text-match {
+    font-weight: var(--calcite-font-weight-bold);
```
driskull, line 218: I think we need to set color as well since the default `mark` element defines a color. I had it set to inherit to take the color of the parent element.

driskull (approved):
👏 🎉

packages/calcite-components/src/assets/styles/includes.scss
jcfranco: Excellent catch. Will fix!

jcfranco:
@ashetland There weren't any [relevant diffs reported](https://www.chromatic.com/build?appId=6266d45509d7eb004aa254fb&number=6979) from this change. I think these screenshots are affected by rendering timing.

ashetland (approved):
🚀

Issue #11198, opened 2025-01-03
[Time Picker] Displays incorrect initial value when `hour-format="24"`


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When the hour value of the component is over 12, on initialization the 12-hour number is present until the minute is specified. 
![verify-4756](https://github.com/user-attachments/assets/37d5a143-48ab-4544-b918-b3c603a1584d)

Additionally, users are unable to see numbers between 13-24 until after the component's minutes are specified.

![verify-4756-2](https://github.com/user-attachments/assets/f429c157-0cd8-4555-9199-a58e2b8c4fb7)

@eriklharper identified its unrelated to the `hourFormat` PR effort, as it occurs using the `time-picker` as a standalone, so its the source of the bug.

### Expected Behavior

- [x] The 24-hour number is present on initialization if a user enters it in the component
- [x] Users can see a list of hour numbers from `0-24` on initialization when interacting with the component

### Reproduction Sample

https://codepen.io/geospatialem/pen/RNbjWoM

### Reproduction Steps

1. Open the [sample](https://codepen.io/geospatialem/pen/RNbjWoM)
2. Open the component
3. Enter a number over 12 (e.g., `13`) for the hour
4. Observe the hour represents the 12 hour number
5. Select a number for the minute
6. Observe the component will change to the 24-hour number

### Reproduction Version

`3.0.0-next.87`

### Relevant Info

Surfaced while verifying https://github.com/Esri/calcite-design-system/issues/4756, more info in this comment: https://github.com/Esri/calcite-design-system/issues/4756#issuecomment-2568360009

### Priority impact

impact - p3 - not time sensitive

### Impact

High with the introduction of the `hour-format="24"`, where it appears the functionality is broken on initialization.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

PR #11200, opened 2025-01-03
fix(time-picker): display correct localized hour based on hour-format when no value is set

eriklharper:
**Related Issue:** #11198 

## Summary

This PR fixes an issue where if no valid time value is currently set, the `time-picker` component doesn't display the hour in the corresponding `hour-format`.

BEGIN_COMMIT_OVERRIDE
omitted from changelog
END_COMMIT_OVERRIDE

### DIFF
--- a/packages/calcite-components/src/components/time-picker/time-picker.e2e.ts
+++ b/packages/calcite-components/src/components/time-picker/time-picker.e2e.ts
@@ -65,6 +65,18 @@ describe("calcite-time-picker", () => {
   });
 
   describe("focusing", () => {
+    describe("should focus the first focusable element when setFocus is called (ltr)", () => {
+      focusable(`calcite-time-picker`, {
+        shadowFocusTargetSelector: `.${CSS.input}.${CSS.hour}`,
+      });
+    });
+
+    describe("should focus the first focusable element when setFocus is called (rtl)", () => {
+      focusable(`<calcite-time-picker dir="rtl" lang="ar"></calcite-time-picker>`, {
+        shadowFocusTargetSelector: `.${CSS.input}.${CSS.hour}`,
+      });
+    });
+
     it("should focus input when corresponding nudge up button is clicked", async () => {
       const page = await newE2EPage();
       await page.setContent(`<calcite-time-picker step=".001"></calcite-time-picker>`);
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/time-picker/time-picker.e2e.ts
```diff
@@ -1254,127 +1182,229 @@ describe("calcite-time-picker", () => {
...
+            });
+            const hour = await page.find(`calcite-time-picker >>> .${CSS.hour}`);
+
+            await hour.click();
+
+            for (let i = 1; i < 24; i++) {
+              await page.keyboard.press("ArrowUp");
+              await page.waitForChanges();
+
+              expect(hour.textContent).toBe(i > 12 ? formatTimePart(i - 12) : formatTimePart(i));
```
jcfranco, line 1311: You can simplify by using `expect(hour).toEqualText(...)`. Applies to other assertions.

packages/calcite-components/src/components/time-picker/time-picker.e2e.ts
```diff
@@ -1254,127 +1182,229 @@ describe("calcite-time-picker", () => {
...
+            }
+
+            await page.keyboard.press("Delete");
+            await page.waitForChanges();
+            await page.keyboard.press("ArrowDown");
```
jcfranco, line 1316: Should this also have a text assertion to be consistent with the 24hr test?

packages/calcite-components/src/components/time-picker/time-picker.e2e.ts
```diff
@@ -1254,127 +1182,229 @@ describe("calcite-time-picker", () => {
...
+            await hour.click();
+
+            for (let i = 1; i < 24; i++) {
+              await page.keyboard.press("ArrowUp");
+              await page.waitForChanges();
+
+              expect(hour.textContent).toBe(i > 12 ? formatTimePart(i - 12) : formatTimePart(i));
+            }
+
+            await page.keyboard.press("Delete");
```
jcfranco, line 1314: Can we make both tests follow similar keyboard steps or why is delete needed in addition to the down arrow here? It's not quite clear from the test.

jcfranco (approved):
Awesome! Aside from a few cleanup comments, this LGTM! ✨🕐🤏✨

packages/calcite-components/src/components/time-picker/time-picker.e2e.ts
eriklharper: Updating both tests to clear the hour input before starting the decrement steps (that's what the Delete is there for).  That way they're both consistent and do what the original test did.

Issue #8053, opened 2023-10-25
calcite-input type="file" button overflows border style in JA locale


zoekmoore:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

With the locale set to Japanese (JA), the file upload button within the `calcite-input type="file"` component expands in height and the bottom of the button is partially obscured by the dashed border surrounding the input field.
![image](https://github.com/Esri/calcite-design-system/assets/54121964/c33d7715-9617-4477-8cc6-da8b5e92db4f)

### Expected Behavior

We expect to see symmetrical padding or margins on the top and bottom of the input button, as is the case with English.

### Reproduction Sample

https://codepen.io/zoe11498/pen/vYbORgV or https://developers.arcgis.com/calcite-design-system/components/input/

### Reproduction Steps

1. Use a machine with Windows 10 OS
2. Set the language of either a Chrome or Edge browser to Japanese (JA)
3. Open either the codepen or the documentation sample provided above, changing `type` to `file`
4. Observe. The bottom of the button intersects with the dashed border around the input component

### Reproduction Version

1.9.0

### Relevant Info

Tested on Windows 10.

Reproducible with both Chrome and Edge browsers.

Not reproducible in Firefox.

### Priority impact

p4 - not time sensitive

### Impact

We noticed this while testing ArcGIS Mission Manager, but this is a low priority i18n UI bug. There is no impact to usability.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Mission

macandcheese:
https://github.com/Esri/calcite-design-system/issues/6545

geospatialem:
Spike to confirm behavior/design on the native `input` element. Refer to the linked above issue, if we can proceed with an effort targeted with the related issue in the future. https://github.com/Esri/calcite-design-system/issues/8053#issuecomment-1779702205

jcfranco:
Managed to reproduce with the native `input[type=file]` and narrowed down the CSS that causes this (highlighted in codepen): https://codepen.io/jcfranco/pen/BaMOoMg

Some options::

[...]

PR #11205, opened 2025-01-04
fix(input): fix clipping for type=file with Japanese locale

jcfranco:
**Related Issue:** #8053

## Summary

Fixes text clipping in Windows 10 when using Japanese locale.


jcfranco:
@macandcheese [This](https://codepen.io/jcfranco/pen/pvzaego) is the closest I can get to simulating it. Do you think it's worth including a similar story?

macandcheese:
> @macandcheese [This](https://codepen.io/jcfranco/pen/pvzaego) is the closest I can get to simulating it. Do you think it's worth including a similar story?

Might be enough of an edge case it's not needed, and we have the upcoming `input-file` work.

jcfranco:
Sounds good. Merging!

PR #11212, opened 2025-01-06
deprecate(notice): deprecate --calcite-notice-width token

Elijbet:
**Related Issue:** N/A

## Summary
Deprecate `--calcite-notice-width token` token.


geospatialem (approved):
Looks great, @Elijbet! 🚢📚

Issue #11170, opened 2024-12-30
[tile] Long content overflows container


jcfranco:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Long content no longer wraps and will overflow the tile container.

### Expected Behavior

Long content wraps within tile's bounds.

### Reproduction Sample

https://codepen.io/jcfranco/pen/QwLqGeR?editors=1000

### Reproduction Steps

1. Load the repro case and compare v2.4.0 with v2.5.0. 
2. Notice that content overflows in v2.5.0.

### Reproduction Version

2.5.0

### Relevant Info

Looks like it was introduced by [this PR](https://github.com/Esri/calcite-design-system/pull/8691). cc @eriklharper 

Screenshot tests do not include the overflowing content, so we may need to set a fixed, large size for overflowing screenshot tests.

### Regression?

2.4.0

### Priority impact

impact - p3 - not time sensitive

### Impact

Possibly low impact as we have not had any reports regarding this regression.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

geospatialem:
Related: https://github.com/Esri/calcite-design-system/issues/11199

geospatialem:
Verified on the `dev` branch

![image](https://github.com/user-attachments/assets/7424708f-209e-41dc-941d-fd6bbf2a6e37)

PR #11213, opened 2025-01-06
fix(tile): wrap overflowing text content

eriklharper:
**Related Issue:** #11170 

## Summary

This PR fixes an issue where overflowing content doesn't stay within the outer bounds of the Tile.

### DIFF
--- a/packages/calcite-components/src/components/tile/tile.scss
+++ b/packages/calcite-components/src/components/tile/tile.scss
@@ -68,7 +68,7 @@ calcite-link {
 
 .content-container {
   flex-direction: column;
-  word-break: break-word;
+@include word-break();
 }
 
 .text-content-container {
--- a/packages/calcite-components/src/components/tile/tile.scss
+++ b/packages/calcite-components/src/components/tile/tile.scss
@@ -68,7 +68,7 @@ calcite-link {
 
 .content-container {
   flex-direction: column;
-  word-break: break-all;
+  word-break: break-word;
 }
 
 .text-content-container {
--- a/packages/calcite-components/src/components/tile/tile.scss
+++ b/packages/calcite-components/src/components/tile/tile.scss
@@ -68,6 +68,7 @@ calcite-link {
 
 .content-container {
   flex-direction: column;
+  word-break: break-all;
 }
 
 .text-content-container {


packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -68,6 +68,7 @@ calcite-link {
 
 .content-container {
   flex-direction: column;
+  word-break: break-all;
```
driskull, line 71: I wonder if the tailwind `@apply break-words` would work for this?

image

eriklharper: I'll try that too.  It is my understanding we're trying not to use tailwind any longer.  I tried out the second approach outlined here: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_text/Wrapping_breaking_text#breaking_long_words

eriklharper: UPDATE: `overflow-wrap: break-word` doesn't work for this case since its a very long word without spaces, so it needs to break the word up with `break-all` to allow the unbroken string to wrap.

driskull: Does the mixin we have work?

```
@mixin word-break() {
  word-wrap: break-word;
  word-break: break-word;
}
```

In the test screenshot it seems to be breaking words when it shouldn't

eriklharper: > In the test screenshot it seems to be breaking words when it shouldn't

That's because its using `word-break: break-all` which will do that.  I think the user should have a choice on the behavior, so I'm going to propose we create a css custom property to set this based on their preference, and probably set `break-word` as the default, since that makes the most sense for the majority of use cases.

driskull: @eriklharper does our mixin `@include word-break();` not work? Thats what we're using elsewhere to solve this.

driskull: I don't think we need a custom CSS prop for this. We're handling this elsewhere

eriklharper: The only property that seems to visually affect the UI is `word-break` with either `break-word` or `break-all`.  The `overflow-wrap` property doesn't visually change anything, at least in this example I'm testing with.  Also, sounds like `word-wrap` is a non-standard property that "maps" to `overflow-wrap`, so if we insist on using the mixin instead of just setting `word-break` directly, we should probably update that mixin to set `overflow-wrap` instead of `word-wrap`.

```
I don't think we need a custom CSS prop for this. We're handling this elsewhere
```
I would argue that the text wrapping behavior might vary depending on the specific needs of the app using Tile.  Maybe the user wants to squeeze as much text as possible in the available horizontal space like this, which can be accomplished with `word-break: break-all`:

image

Or maybe they don't want words to break onto the next line, in which case they'll want `word-break: break-word`:

image

Or maybe they want to use `hyphen`s:

image

Whose to say what the user wants/needs?

Maybe we opt for `break-word` for now, but these are some thoughts I had while working on this that could be good to discuss.

driskull: For consistency, we should go with `break-word`. I'm not sure hyphens would make sense in most scenarios.

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -68,6 +68,7 @@ calcite-link {
 
 .content-container {
   flex-direction: column;
+  word-break: break-word;
```
driskull, line 71: Can you use the word-break mixin here instead? For consistency.

```suggestion
@include word-break();
```

driskull (approved):
👍 💯 

If this doesn't trigger any screenshot changes be sure to add one for the use case.

eriklharper:
> 👍 💯
> 
> If this doesn't trigger any screenshot changes be sure to add one for the use case.

Definitely, thanks for the review and approval!  [There is this existing story](https://github.com/Esri/calcite-design-system/blob/88b14bfb565dcd39a8daf9870c53ab57c62b4232/packages/calcite-components/src/components/tile/tile.stories.ts#L967-L974) that should capture this use case, and I was surprised that it didn't catch this originally, not sure if it was approved incorrectly or not, but I'll make sure to monitor it and add a new one if necessary.

Issue #10931, opened 2024-12-02
[Tile / Tile Group] When no `description` - selection mode icons are too large


macandcheese:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The Tile Group / Tile component render the selection icon too large when the component does not have a "description" property:

![Screenshot 2024-12-02 at 11 10 39 AM](https://github.com/user-attachments/assets/d267e028-640f-4c5b-9449-7ffeb72080e5)

### Expected Behavior

I'd expect the selection icon affordance to not change. cc @ashetland for any further design input

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/QwLbPyQ

### Reproduction Steps

1. Open Codepen
2. See issue
3. Adjust "description" prop as needed to show difference

### Reproduction Version

2.13.2

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

jasdelreyes:
+1! (Closed a duplicate issue: https://github.com/Esri/calcite-design-system/issues/10943)

geospatialem:
Verified on the `dev` branch ✅ 

![image](https://github.com/user-attachments/assets/86f8c213-98b7-4213-b277-fb23ee4861c2)

PR #11215, opened 2025-01-07
fix(tile-group): keep selection icon size consistent for large tiles

jcfranco:
**Related Issue:** #10931 

## Summary

Fixes an issue where large tiles (w/o `description`) would cause the selection icon to be oversized and misaligned.

### Noteworthy changes

* refactors stories to use HTML-generating functions to improve maintainability
* adds internal CSS var to reuse and simplify internal space assignments
* applies icon styling separately
* applies relative positioning to content container
* replaces sticky z-index var in favor of default + bump to keep it within the same layer

### DIFF
--- a/packages/calcite-components/src/components/tile-group/tile-group.stories.ts
+++ b/packages/calcite-components/src/components/tile-group/tile-group.stories.ts
@@ -687,7 +687,7 @@ function createVariantsHtmlStory(layout: TileGroup["layout"]): () => string {
     </div>
 
     <div class="parent">
-      <div class="child right-aligned-text">icon and heading (large visual) + single selection mode</div>
+      <div class="child right-aligned-text">icon and heading (large visual) + multiple selection mode</div>
       <div class="child">
         ${getTileGroupHtml(
           {
@@ -695,7 +695,7 @@ function createVariantsHtmlStory(layout: TileGroup["layout"]): () => string {
             icon: true,
           },
           layout,
-          "single",
+          "multiple",
           "s",
         )}
       </div>
@@ -706,7 +706,7 @@ function createVariantsHtmlStory(layout: TileGroup["layout"]): () => string {
             icon: true,
           },
           layout,
-          "single",
+          "multiple",
           "m",
         )}
       </div>
@@ -717,7 +717,7 @@ function createVariantsHtmlStory(layout: TileGroup["layout"]): () => string {
             icon: true,
           },
           layout,
-          "single",
+          "multiple",
           "l",
         )}
       </div>
--- a/packages/calcite-components/src/components/tile-group/tile-group.stories.ts
+++ b/packages/calcite-components/src/components/tile-group/tile-group.stories.ts
@@ -101,16 +101,18 @@ function getTileHtml(options: Partial<TileHtmlOptions> = {}): string {
     selected = false,
   } = options;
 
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/tile-group/tile-group.stories.ts
```diff
@@ -46,2304 +46,747 @@ export default {
...
-        <calcite-tile heading="Tile heading lorem ipsum" href="/" icon="layers"></calcite-tile>
-        <calcite-tile heading="Tile heading lorem ipsum" href="/" icon="layers"></calcite-tile>
-        <calcite-tile heading="Tile heading lorem ipsum" href="/" icon="layers"></calcite-tile>
-        <calcite-tile heading="Tile heading lorem ipsum" href="/" icon="layers"></calcite-tile>
-      </calcite-tile-group>
-    </div>
-  </div>
-`;
+export const allVariantsHorizontal = createVariantsHtmlStory("horizontal");
+export const allVariantsVertical = createVariantsHtmlStory("vertical");
```
macandcheese, line 792: I think you broke Chromatic 😄 https://www.chromatic.com/test?appId=6266d45509d7eb004aa254fb&id=677ce7bbb76f8e6437051490

jcfranco: Oh noes! @eriklharper Would a shorter description value help here? We could split the all-variants stories, but it sort of defeats the purpose. 🥲

eriklharper: You'd have to just try shortening it I suppose and see it makes it small enough for Chromatic to capture, but yeah, possibly that could make a difference.

packages/calcite-components/src/components/tile/tile.scss
```diff
@@ -124,15 +133,10 @@ calcite-link {
 }
 
 :host([scale="s"]) {
+  --calcite-internal-spacing: var(--calcite-spacing-sm);
```
eriklharper, line 136: Should we include the component name in internal css variable names, like `--calcite-internal-tile-spacing`?  Not sure what would happen if another component defined a `--calcite-internal-spacing` var, but even though these are internal, might make sense to always include the component name even if just for findability in the codebase.

eriklharper:
I'm seeing a hover border regression where the right side of the middle hovered tile doesn't appear correctly.  Could be a stacking issue:

image

packages/calcite-components/src/components/tile/tile.scss
jcfranco: Yeah, good catch. Will fix!

jcfranco:
@eriklharper Fixed the hover outline regression. Would you mind taking another look?

eriklharper:
> I'm seeing a hover border regression where the right side of the middle hovered tile doesn't appear correctly. Could be a stacking issue:
> 
> image

UPDATE: I'm no longer seeing this happen locally on your latest changes.

eriklharper (approved):
NOICE!  Thanks @jcfranco for addressing this!   I definitely caused this to happen as part of my `selection-mode` changes to Tile, so appreciate the fix and the nicer story setup :)

ashetland (approved):
Visuals lookin' gewd!

Issue #11217, opened 2025-01-07
[text-area]: `readOnly` prop not listed as part of `StencilCalciteTextArea` interface


LanceStasinski:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The `readOnly` property is listed in the doc for calcite-text-area, but the property is not listed as part of the `StencilCalciteTextArea` interface. This causes TS errors in stencil JSX.

### Expected Behavior

The `readOnly` property should be listed as part of the `StencilCalciteTextArea` interface.

### Reproduction Sample

NA

### Reproduction Steps

Add `readOnly` to a `<calcite-text-area>` element in stencil JSX. Notice the TS error.
```md
Type '{ readOnly: true; }' is not assignable to type 'StencilCalciteTextArea'.
  Property 'readOnly' does not exist on type 'StencilCalciteTextArea'.
```

### Reproduction Version

3.0.0-next.89

### Regression?

Last version of 2.x

### Priority impact

impact - p1 - need for current milestone

### Impact

This breaks stencil builds when `readOnly` is added to a calcite-text-area.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Field Apps

DitwanP:
Verified on `3.0.0-next.94`

PR #11218, opened 2025-01-07
fix(text-area): make `readOnly` prop writable

jcfranco:
**Related Issue:** #11217 #10731

## Summary

Removes unintentional `@readonly` on `readOnly` prop. 

For context, with #10310 any props with the `@readonly` tag will be made read-only.


driskull (approved):
👍

PR #11054, opened 2024-12-13
chore(t9n): rename message files

anveshmekala:
**Related Issue:** https://github.com/Esri/calcite-design-system/issues/11081

## Summary

- Updates filenames of message files in each component. New specification follow `message.lang.json` format.
- Removes `util:sync-t9n-en-bundles` script.
- Update `t9nmanifest.txt` files with updated path.

Related to https://devtopia.esri.com/WebGIS/arcgis-web-components/discussions/1522

PR #11219, opened 2025-01-07
fix(docs): use updated translation bundle paths to generate `translations.json`

jcfranco:
**Related Issue:** #10731

## Summary

Updates `generateT9nDocsJSON.ts` to reference the updated translations bundle file name (see https://github.com/Esri/calcite-design-system/pull/11054).

### DIFF
--- a/packages/calcite-components/support/generateT9nDocsJSON.ts
+++ b/packages/calcite-components/support/generateT9nDocsJSON.ts
@@ -24,7 +24,7 @@
         Object.keys(messagesFileMain).forEach((key) => (data[component][key] = {}));
 
         const messagesFilenames = (await readdir(t9nPath, { withFileTypes: true })).map((dirent) => dirent.name);
-        const messagesFilenameRegex = new RegExp(`${component}\\.t9n\\.(.*)\\.json`);
+        const messagesFilenameRegex = new RegExp(`messages\\.(.*)\\.json`);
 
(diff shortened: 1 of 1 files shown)


packages/calcite-components/support/generateT9nDocsJSON.ts
```diff
@@ -24,7 +24,7 @@
         Object.keys(messagesFileMain).forEach((key) => (data[component][key] = {}));
 
         const messagesFilenames = (await readdir(t9nPath, { withFileTypes: true })).map((dirent) => dirent.name);
-        const messagesFilenameRegex = new RegExp(`${component}\\.t9n\\.(.*)\\.json`);
+        const messagesFilenameRegex = new RegExp(`messages\\.(.*)\\.json`);
```
anveshmekala, line 27: Sidenote: should have been caught here https://github.com/Esri/calcite-design-system/pull/11054. 

Is there a way to add a test to assert this?

macandcheese (approved):

packages/calcite-components/support/generateT9nDocsJSON.ts
jcfranco: Currently, we don’t have tests for our support files/scripts, but we could explore adding something like output snapshot diffing, similar to the design tokens. Open to ideas here. Could you open a follow-up issue for this?

PR #11232, opened 2025-01-08
ci: ensure lint-staged works with typescript

benelan:
## Summary

Resolve some issues with our linting setup. For monorepos, there needs to be a lint-staged config file next to every eslint config file.

ref: https://typescript-eslint.io/packages/typescript-eslint


package.json
```diff
@@ -94,7 +92,6 @@
     "eslint-plugin-import": "2.31.0",
     "eslint-plugin-jest": "28.10.0",
     "eslint-plugin-jsdoc": "50.6.1",
-    "eslint-plugin-prettier": "5.2.1",
```
jcfranco, line 97: Just to confirm, was this and references to its associated config removed because we have separate linting (both NPM script and `lint-staged` running `prettier`)?

jcfranco (approved):
Awesome! 🚀

package.json
benelan: Yeah I followed the suggestion on this page:
https://typescript-eslint.io/users/what-about-formatting

Issue #11224, opened 2025-01-08
[Autocomplete] Provide a way to remove the search icon


driskull:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Currently, the search icon cannot be removed.

I wonder if we should just not set the icon like default just as the `input` component does? That way, there's no icon by default and a developer doesn't have to set the icon to false using JS.

Because we set the `input` type to `search` in the autocomplete component, when a user sets the icon to an empty string (icon="") it will default to the search icon because the input component defaults to the input type icon. Maybe it shouldn't? Its kind of odd because I would expect setting an icon prop to "" to remove the icon.

See https://codepen.io/Matt-Driscoll/pen/YPKYbZK?editors=1010

@jcfranco @ashetland @SkyeSeitz @matgalla would love your thoughts here.

### Expected Behavior

The ability to remove the search icon.

### Reproduction Sample

https://codepen.io/Matt-Driscoll/pen/XJrVwpB?editors=1010

### Reproduction Steps

1. open the codepen
2. notice the search icon remains even though the prop has been set to false
3. setting the prop to an empty string has no effect either

### Reproduction Version

3.0.0-90

### Relevant Info

questions:

1. Should we have no default icon set for autocomplete?
2. Should `input` reconsider setting the icon to the input's type when the icon prop is set to an empty string? cc @eriklharper 

### Priority impact

impact - p1 - need for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

driskull:
setRequestedIcon is also weird because it will use the default icon for the type if its an empty string but not null or undefined.

https://github.com/Esri/calcite-design-system/blob/732933a8ed6e0da21edef9d6b8d6b59b363d7292/packages/calcite-components/src/utils/dom.ts#L328-L338

jcfranco:
IIRC, the `icon` prop (`boolean | string`) should follow this behavior:

1. `false` should remove the icon
1. `true` uses whatever icon is associated with the type
1. string must match a supported icon

There might be some inconsistencies or follow-up needed due to Stencil's quirks with props of mixed type.

[...]

PR #11239, opened 2025-01-09
fix(autocomplete): allow hiding icon

driskull:
**Related Issue:** #11224

## Summary

- allow icon to be set false
- add tests

BEGIN_COMMIT_OVERRIDE
END_COMMIT_OVERRIDE


eriklharper (approved):
LGTM! 👍   I assume that because of `type="search"` being set on `calcite-input` that a search icon will display when there's no icon prop passed to autocomplete, right?

driskull:
> LGTM! 👍 I assume that because of type="search" being set on calcite-input that a search icon will display when there's no icon prop passed to autocomplete, right?

Yep, thats right.

Issue #11152, opened 2024-12-27
[Autocomplete] add scrollContentTo method and automatically scroll items when using keyboard


driskull:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

- add scrollContentTo method
- automatically scroll items when using keyboard (arrow keys/home/end)

### Acceptance Criteria

- add scrollContentTo method
- automatically scroll items when using keyboard (arrow keys/home/end)

### Which Component

autocomplete

### Example Use Case

https://codepen.io/Matt-Driscoll/pen/OPLjzXX?editors=1010

1. open component
2. use arrow key down to navigate items
3. notice that it doesn't scroll when you get to the last item displayed

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

geospatialem:
The fix doesn't appear to work across Windows browsers (Chrome, FF, and Edge) when trying to navigate to the last `autocomplete-item` via the `down arrow` and `End` keys with https://codepen.io/geospatialem/pen/raBpzmm. Wondering if there's some specific logic that addresses the functionality for Mac/Mac browsers, but not Windows/Windows browsers? 

Reassigning to further diagnose.

driskull:
> Wondering if there's some specific logic that addresses the functionality for Mac/Mac browsers, but not Windows/Windows browsers?

Nope, it just uses this: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

geospatialem:
Verified with `3.0.0-next.93`

PR #11242, opened 2025-01-09
fix(autocomplete): ignore disabled items when navigating via keyboard

driskull:
**Related Issue:** #11152

## Summary

- Only includes enabled items in keyboard navigation
- disabled items are ignored

BEGIN_COMMIT_OVERRIDE
omitted from changelog
END_COMMIT_OVERRIDE


jcfranco (approved):
Nice! Before merging, can you add an E2E test for this?

PR #11248, opened 2025-01-10
chore(themed): improve matching of interaction target elements

jcfranco:
**Related Issue:** N/A

## Summary

This adds support for `find` string selectors for testing stateful interactions. 

### Notable changes

* deprecates the `attribute`/`value` matching since `find` string selectors:
  * accomplish the same
  * align better with E2E tests
  * simpler to wire up
  * provide more precise matching
* changes the existing `attribute`/`value` matching to be scoped to the targeted element:
  * improves reliability of matches
  * encourages using `find` string selectors for more complex cases


aPreciado88 (approved):
🚀

Issue #10327, opened 2024-09-16
[Tooling] Setup automation to remove a design project once marked as `ready for dev`


geospatialem:
### Priority impact

p - low

### Summary

To simplify our processes and streamline the way we track design issues in projects we should explore and if possible, adopt removing the project via Actions.

### Desired Outcome

Removes the design project from the issue when the `ready for dev` label is added.

### Resources

- [Automation with Actions](https://docs.github.com/en/issues/planning-and-tracking-with-projects/automating-your-project/automating-projects-using-actions)

calcite-admin:
The issue has been removed from the [Calcite Design Effort 2025-12-16 Dec](https://github.com/orgs/Esri/projects/72) project.

DitwanP:
Verified 

<img width="277" alt="Image" src="https://github.com/user-attachments/assets/eb1d2804-8090-49ec-9e32-f3e64e6dd521" />

PR #11249, opened 2025-01-10
ci(actions): add action to archive issues in design projects

DitwanP:
**Related Issue:** #10327 

## Summary
Action that will archive issues in respective projects after design team has marked them with the `ready for dev` label.

### DIFF
--- a/.github/scripts/archiveIssuesInDesignProjects.js
+++ b/.github/scripts/archiveIssuesInDesignProjects.js
@@ -1,0 +1,72 @@
+const { execSync } = require("child_process");
+
+// Environment variables from the GitHub Action
+const owner = process.env.OWNER;
+const repo = process.env.REPO;
+const issueNumber = process.env.ISSUE_NUMBER;
+const labelName = process.env.LABEL_NAME;
+
+// Function to execute a GitHub GraphQL command
+function runQuery(query) {
+  const command = `gh api graphql -f query='${query}' -F owner="${owner}" -F repo="${repo}" -F issueNumber=${issueNumber}`;
+  return execSync(command, { encoding: "utf-8" });
+}
+
+// Function to create a comment
+async function createComment(body) {
+  const command = `gh issue comment ${issueNumber} --body "${body}"`;
+  return execSync(command, { encoding: "utf-8" });
+}
+
+// GraphQL query to find the project associated with the issue
+const query = `
+  query($owner: String!, $repo: String!, $issueNumber: Int!) {
+    repository(owner: $owner, name: $repo) {
+      id
+      nameWithOwner
+      description
+      issue(number: $issueNumber) {
+        id
+        title
+        projectItems(first: 1) {
+          nodes {
+            id
+            project {
+              id
+              title
+              url
+            }
+          }
+        }
+      }
+    }
+  }
+`;
+
+try {
+  const result = runQuery(query);
+  const parsedResult = JSON.parse(result);
+  const projectItem = parsedResult.data.repository.issue.projectItems.nodes[0];
+  console.log("Project Item:", projectItem);
+
+  if (projectItem) {
+    console.log(
+      `Issue #${issueNumber} is in design project: ${projectItem.project.title} (URL: ${projectItem.project.url})`,
+    );
+
+    if (labelName === "ready for dev") {
+      const archiveQuery = `mutation { archiveProjectV2Item(input: {projectId: "${projectItem.project.id}", itemId: "${projectItem.id}"}) { clientMutationId } }`;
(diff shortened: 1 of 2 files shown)


.github/workflows/archive-design-issues-in-projects.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "16"
```
benelan, line 18: Is there a reason this needs to be Node v16? If not can you change it to:

```suggestion
          node-version-file: package.json
```

.github/scripts/archiveIssuesInDesignProjects.js
```diff
@@ -0,0 +1,68 @@
...
+  const parsedResult = JSON.parse(result);
+  const projectItem = parsedResult.data.repository.issue.projectItems.nodes[0];
+  console.log("Project Item:", projectItem);
+
+  if (projectItem) {
+    if (labelName === "ready for dev") {
+      const archiveQuery = `mutation { archiveProjectV2Item(input: {projectId: "${projectItem.project.id}", itemId: "${projectItem.id}"}) { clientMutationId } }`;
+      runQuery(archiveQuery);
+      createComment(
+        `This issue has been archived in the [${projectItem.project.title}](${projectItem.project.url}/archive) project.`,
```
benelan, line 57: Do we need this comment? If I saw a comment saying my issue was archived I'd think it's not being worked on. Whereas, in this case, "archived" means the issue is progressing lol

.github/workflows/archive-design-issues-in-projects.yml
DitwanP: I think I did this because the action showing a warning that was saying something about moving to a later version of node. I just bumped it to be whatever it was recommending at the time. I'll change it to the suggestion.

.github/scripts/archiveIssuesInDesignProjects.js
benelan: What is the project title? If it is something like "Design tasks" then the additional context should be enough

DitwanP: Yeah all the projects are named something like "Calcite Design Effort...[Date of milestone]"

.github/scripts/archiveIssuesInDesignProjects.js
```diff
@@ -0,0 +1,68 @@
...
+`;
+
+try {
+  const result = runQuery(query);
+  const parsedResult = JSON.parse(result);
+  const projectItem = parsedResult.data.repository.issue.projectItems.nodes[0];
+  console.log("Project Item:", projectItem);
+
+  if (projectItem) {
+    if (labelName === "ready for dev") {
```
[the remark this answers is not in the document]
benelan: To clarify, I don't mean use `actions/github-script`, just add something like the following to the existing workflow instead of creating a new one.

```yml
      - name: "Archive design issue"
        if: github.event.label.name == 'ready for dev'
        env:
          GITHUB_TOKEN: ${{ secrets.ADMIN_TOKEN }}
          OWNER: ${{ github.repository_owner }}
          REPO: ${{ github.event.repository.name }}
          ISSUE_NUMBER: ${{ github.event.issue.number }}
          LABEL_NAME: ${{ github.event.label.name }}
        run: node .github/scripts/archiveIssuesInDesignProjects.js
```

And then you can remove the redundant labelName conditional in the script.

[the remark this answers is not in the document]
DitwanP: Got it 👌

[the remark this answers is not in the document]
DitwanP: Heads up, I'm changing the script to remove the issues from the project instead of archiving because now github still shows the project info on the side if it's archived. The original ask was to remove issues from design projects so that they wouldn't see the project info on the side and assume it was still relevant. 

Archiving used to achieve the same visual result (no project on side panel) while also maintaing the log of issues but not any more so now I'm going back to removal.

DitwanP:
@benelan 

Heads up, I'm changing the script to remove the issues from the project instead of archiving because now GitHub still shows the project info on the side if it's archived. The original ask was to remove issues from design projects so that they wouldn't see the project info on the side and assume it was still relevant.

Archiving used to achieve the same visual result (no project on side panel) while also maintaining the log of issues but not any more so now I'm going back to removal.

benelan (approved):
LGTM

Issue #9581, opened 2024-06-13
The `calciteInputInput` event is triggered early causing incorrect status 


DitwanP:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When a file is uploaded using the `calcite-input` component, the `status` property does not change from `invalid` to `valid` as expected, even though the file upload is successful. This is because the event listener for the `calciteInputInput` event triggers before the file upload completes, causing the input status to remain invalid. Adding a timeout resolves the issue, indicating a timing problem with the event trigger.

### Expected Behavior

The timing of the even triggering should be adjusted so that the input `status` can correctly reflect the current state.

### Reproduction Sample

https://codepen.io/Dipali-Joshi/pen/ExzxYaX?editors=1000

### Reproduction Steps

1. Open the sample application using the provided link.
2. Open the Console in Developer Tools.
3. Observe that the ‘validation-message’ “Invalid File” is displayed because the ‘status’ property of the `calcite-input` element at line `10` is set to ‘invalid’ as a default. 
4. An event listener triggered by the ‘calciteInputInput’ event is added to the input element, which checks the ‘value’ of the input to change the status to ‘valid’ once a file has been uploaded. 
5. Choose and upload a file using the input and observe that the status does not change even though the ‘value’ property has a value (seen in the console log). 
6. We can also observe an error that the ‘files’ property of the input is not defined. 
7. Now uncomment line `14` and `23` to add a timeout to the code. Observe that this time the status changes successfully. 

### Reproduction Version

2.8.0

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

benelan:
@DitwanP can you let support know that using `calciteInputChange` instead of `calciteInputInput` in the codepen resolves the issue, and ask them if it also fixes it for the user? I will do some additional investigation into the `calciteInputInput` timing.

DitwanP:
Verified on `3.0.0-next.96`
https://codepen.io/Ditwan-Price/pen/MYgGbYe?editors=1000

PR #11262, opened 2025-01-10
fix(input): update files property on input event when type="file"

benelan:
**Related Issue:** #9581

## Summary

Ensure the `files` property is updated before `calciteInputInput` is emitted for `type="file"` inputs.

### DIFF
--- a/packages/calcite-components/src/components/input/input.tsx
+++ b/packages/calcite-components/src/components/input/input.tsx
@@ -679,6 +679,11 @@ export class Input
     if (this.disabled || this.readOnly) {
       return;
     }
+
+    if (this.type === "file") {
+      this.files = (this.childEl as HTMLInputElement).files;
+    }
+
     this.setValue({
       nativeEvent,
       origin: "user",
--- a/packages/calcite-components/src/components/input/input.tsx
+++ b/packages/calcite-components/src/components/input/input.tsx
@@ -669,12 +669,6 @@ export class Input
     this.calciteInternalInputFocus.emit();
   }
 
-  private inputChangeHandler(): void {
-    if (this.type === "file") {
-      this.files = (this.childEl as HTMLInputElement).files;
-    }
-  }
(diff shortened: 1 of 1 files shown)


packages/calcite-components/src/components/input/input.tsx
```diff
@@ -679,6 +679,11 @@ export class Input
     if (this.disabled || this.readOnly) {
       return;
     }
+
+    if (this.type === "file") {
```
jcfranco, line 683: It looks like we can remove the same logic from `inputChangeHandler` since input would always emit before change, right?

jcfranco (approved):
Awesome! Would it be possible to add a test for this?

driskull (approved):
👍

benelan:
> Awesome! Would it be possible to add a test for this?

It wasn't possible with jest, but I did some research and vitest does have experimental support:
https://vitest.dev/guide/browser/interactivity-api.html#userevent-upload

We would need to add playwright or webdriverio to our vitest setup:
https://vitest.dev/guide/browser/#manual-installation

Would that play okay with Lumina @maxpatiiuk?

Maybe this warrants a separate issue @jcfranco?

maxpatiiuk:
> Would that play okay with Lumina @maxpatiiuk?

I believe the documentation you are looking at is for vitest browser mode. Anything imported from `@vitest/browser/` is for browser mode.

Migrating Calcite's E2E tests to vitest browser mode would be a DX improvement for sure, but it is much larger than the scope of this PR.

P.S: I couldn't see an issue for migration to vitest browser mode - should I create one?

jcfranco:
A new issue would be great, thanks! Just noting that transitioning to vitest browser mode will be put on hold until it is no longer experimental.

maxpatiiuk:
Opened an issue: https://github.com/Esri/calcite-design-system/issues/11268

Issue #9922, opened 2024-07-31
A couple icon suggestions


COV-GIS:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Follow through on https://github.com/Esri/calcite-ui-icons/issues/352

### Acceptance Criteria

Pulling icons into the monorepo is awesome. 💯 to whoever thought of it.

Just a suggestion...issue template for icons.

### Which Component

icons...not calcite-icon

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

geospatialem:
Thanks @COV-GIS! 👍🏻 

The above request of https://github.com/Esri/calcite-ui-icons/issues/352 has been completed, and will be part on the upcoming January release for `calcite-ui-icons` 🚀 

> Just a suggestion...issue template for icons.

Will be working towards adding an option to Calcite's [enhancement request template](https://github.com/Esri/calcite-design-system/issues/new?assignees=&labels=enhancement%2C0+-+new%2Cneeds+triage&projects=&template=enhancement.yml). ✨

geospatialem:
Verified with adding the label in the [following action](https://github.com/Esri/calcite-design-system/issues/10634#event-16018372796). 🔔

PR #11263, opened 2025-01-10
chore(actions): add icon leads notification for calcite-ui-icons requests

geospatialem:
**Related Issue:** #9922

## Summary
Add automation to issues when the `calcite-ui-icons` label is added, where Icon team leads will receive an `@` notification via a comment, similar to the [new component label automation](https://github.com/Esri/calcite-design-system/issues/10382#issuecomment-2555481044) from the [notifyAboutNewComponent script](https://github.com/Esri/calcite-design-system/blob/dev/.github/scripts/notifyAboutNewComponent.js).

cc @allieorth, @arowles


.github/scripts/notifyAboutIconRequest.js
```diff
@@ -0,0 +1,32 @@
...
+/** @param {import('github-script').AsyncFunctionArguments} AsyncFunctionArguments */
+module.exports = async ({ github, context }) => {
+  const { repo, owner } = context.repo;
+
+  const payload = /** @type {import('@octokit/webhooks-types').IssuesLabeledEvent} */ (context.payload);
+  const {
+    issue: { number },
+  } = payload;
+
+  const { ICONLEADS } = process.env;
```
jcfranco, line 15: This doesn't match the `env` variable name defined [here](https://github.com/Esri/calcite-design-system/pull/11263/files#diff-3352a53ad58df5f5774810546c866b333c981fa2e7eaff4981e4e116b677b483R58).

.github/workflows/issue-notifications.yaml
```diff
@@ -51,3 +51,12 @@ jobs:
...
 
+      - name: "Icon request notification"
+        if: github.event.label.name == 'calcite-ui-icons'
+        uses: actions/github-script@v7
+        env:
+          ICONS_TEAM: ${{secrets.ICON_LEADS}}
+        with:
+          script: |
+            const action = require('${{ github.workspace }}/.github/scripts/notifyAboutIconRequest.js')
+            await action({github, context, core})
```
benelan, line 62: can you use the `yml` extension instead of `yaml` to align with the rest of the workflows?

.github/scripts/notifyAboutIconRequest.js
```diff
@@ -0,0 +1,32 @@
+// @ts-nocheck
```
benelan, line 1: Why is this set to nocheck?

benelan (commented):
Why are tagging icon designers in a comment instead of assigning them (or the @Esri/calcite-icon-designers team) to the issue? I think designers would be notified either way, but it assigning would reduce noise for the issue author.

geospatialem:
> Why are tagging icon designers in a comment instead of assigning them ~(or the @Esri/calcite-icon-designers team)~ to the issue? I think designers would be notified either way, but it assigning would reduce noise for the issue author.
> 
> Edit: Actually you can't assign a team to an issue. But assigning the individual designer(s) is still an option.

If we'd rather assign issues to the team, can alter the automation - the intent was to be similar to our new component and spike complete notifications, where:
- Icon leads would be notified and could review the request
 - If the Icon team would like to proceed with the request, they could assign themselves to follow-up with the requestor once the icon(s) are added
- If the Icon team does not wish to proceed with the request, the issue would be closed with a response to the requestor

.github/scripts/notifyAboutIconRequest.js
geospatialem: Copied this over from another sample, removed it from this PR.

.github/scripts/notifyAboutIconRequest.js
geospatialem: Updated!

.github/scripts/notifyAboutIconRequest.js
```diff
@@ -0,0 +1,31 @@
+// If the "calcite-ui-icons" label is added to an issue, generates a notification to the Icons team leads to review
```
benelan, line 1: Can you add the following line to the top of the file

```js
// @ts-check
```

.github/scripts/notifyAboutIconRequest.js
```diff
@@ -0,0 +1,31 @@
...
+/** @param {import('github-script').AsyncFunctionArguments} AsyncFunctionArguments */
+module.exports = async ({ github, context }) => {
+  const { repo, owner } = context.repo;
+
+  const payload = /** @type {import('@octokit/webhooks-types').IssuesLabeledEvent} */ (context.payload);
+  const {
+    issue: { number },
+  } = payload;
+
+  const { ICON_LEADS } = process.env;
```
benelan, line 14: this environment variable needs to be the same thing you called it in the workflow

```suggestion
  const { ICONS_TEAM } = process.env;
```

.github/scripts/notifyAboutIconRequest.js
```diff
@@ -0,0 +1,31 @@
...
+
+  // Add a "@" character to notify the user
+  const icon_leads = ICON_LEADS?.split(",").map((v) => " @" + v.trim());
```
benelan, line 17:
```suggestion
  const icon_leads = ICONS_TEAM?.split(",").map((v) => " @" + v.trim());
```

benelan (changes requested):
> If we'd rather assign issues to the team, can alter the automation - the intent was to be similar to our new component and spike complete notifications, where:
> 
> * Icon leads would be notified and could review the request
> * If the Icon team would like to proceed with the request, they could assign themselves to follow-up with the requestor once the icon(s) are added
> * If the Icon team does not wish to proceed with the request, the issue would be closed with a response to the requestor


Gotcha. I don't remember why we decided to comment/mention instead of assign in previous workflows. Assigning would notify them as well, and would have the added benefit of reducing noise for the requester and showing up in the designer's list of assigned issues so things don't get lost.

We don't need to make the change in this PR, but it's something worth considering.

benelan (approved):
:+1:

PR #11240, opened 2025-01-09
build: create a private package for shared eslint configs

benelan:
## Summary

- Create a new package in `packages/eslint-config-calcite` that contains shared ESLint configurations for importing in other packages.
- Setup the components and design-tokens packages to use the new shared ESLint configs.
- Add typescript-eslint coverage to our github scripts 
- Check css files in lint-staged (previously it was only scss)
- Disable [`no-unexpected-newline`](https://archive.eslint.org/docs/6.0.0/rules/no-unexpected-multiline), which [conflicts with prettier](https://github.com/prettier/eslint-config-prettier/issues/32)
- Various other cleanup, such as formatting json files in `.vscode/` and removing cspell ignore comments in favor of dictionary additions..
- Run `npm run lint` so unrelated changes don't appear in subsequent PRs

PR #11270, opened 2025-01-12
chore: replace jest with vitest

jcfranco:


benelan (approved):
Awesome! This LGTM, but I think we should let @alisonailea take a look before merging.

There was a merge conflict with the design token package's eslint config after I installed #11240. I pushed a fix to your branch, hope you don't mind!

alisonailea (approved):
Lets go!!

Issue #11148, opened 2024-12-26
calcite-input-text doesn't show anything when pattern is not matched


MikeTschudi:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The sample line
```html
 <calcite-input-text id="srcId" max-length="32" min-length="32" pattern="[a-f0-9A-F]{32}" placeholder="Item, group, or folder id"></calcite-input-text>
```
only honors the max-length restriction. There's no lack-of-validity display when the text is too short or doesn't match the pattern.  (Also tried with "^" prefix and "$" suffix.)

### Expected Behavior

Highlights like we see in MDN, e.g., if value is too short:
![image](https://github.com/user-attachments/assets/36f0443d-3cab-4371-8bf3-e628b31f04c3)

MDN playground source:
```html
  <label>
    Enter AGO id <input
      pattern="[a-f0-9A-F]{32}"
      size="32"
    />
  </label>
```

### Reproduction Sample

https://codepen.io/miketschudi/pen/GgKEaPY?editors=100

### Reproduction Steps

above CodePen is for 3.0.0-next.81.  For 2.13.2, 
https://codepen.io/miketschudi/pen/WbeOBLV?editors=100

### Reproduction Version

3.0.0-next.81

### Relevant Info

Windows 11 Enterprise using Chrome 131.0

Here's what the component shows when given "not an id!" in the calcite-input-text field:
![image](https://github.com/user-attachments/assets/67950b1b-324c-43c0-8564-de045ec6e84d)

### Priority impact

impact - p3 - not time sensitive

### Impact

no effect yet--just found this in a prototype for future work

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Solutions

benelan:
The current validation logic only works within forms. I updated the provided codepen and it appears to be working correctly: form submission is prevented when you press Enter while the input is focused.

https://codepen.io/benizzy/pen/YPKYxRM?editors=1000

benelan:
Yeah that's a good idea; thanks for the suggestion. cc @DitwanP @geospatialem

We hope to remove that requirement when we migrate to elementInternals in #8126.

[...]

PR #11285, opened 2025-01-13
docs(autocomplete, input, input-text): update pattern context for residing in a parent form

geospatialem:
**Related Issue:** #11148

## Summary
Updates the `pattern` prop in `autocomplete`, `input`, and `input-text` that to use the prop the component must reside within a form element, re: https://github.com/Esri/calcite-design-system/issues/11148#issuecomment-2574036966.

### DIFF
--- a/packages/calcite-components/src/components/autocomplete/autocomplete.tsx
+++ b/packages/calcite-components/src/components/autocomplete/autocomplete.tsx
@@ -241,7 +241,8 @@ export class Autocomplete
   @property({ reflect: true }) overlayPositioning: OverlayPositioning = "absolute";
 
   /**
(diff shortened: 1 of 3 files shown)


DitwanP (approved):
✨

benelan (approved):
Nice! FYI other constraint validation properties like `required`, `max, `min`, `max-length`, and `min-length` also need be associated with a form.

geospatialem:
> Nice! FYI other constraint validation properties like `required`, `max`, `min`, `max-length`, and `min-length` also need be associated with a form.

Appreciate the context, I'll open a new PR to follow-up on the above.

PR #11286, opened 2025-01-13
refactor(openClose): use the correct property to specify open/close transition prop

jcfranco:
**Related Issue:** N/A

## Summary

Updates open/close components to use the correct transition property.

### Notes

* this was working as expected because the property wasn't being used correctly
* makes `openProp` optional and defaults to `open` when not specified
* updates types to ensure CSS props are specified by `transitionProp`
* renames `openTransitionProp` to `openProp` to avoid confusion with CSS transitions

@Elijbet Thanks for spotting this one!

### DIFF
--- a/packages/calcite-components/src/utils/openCloseComponent.ts
+++ b/packages/calcite-components/src/utils/openCloseComponent.ts
@@ -17,7 +17,7 @@ export interface OpenCloseComponent {
    */
   openProp?: string;
 
-  /** Specifies the name of transitionProp. */
+  /** Specifies the name of CSS transition property. */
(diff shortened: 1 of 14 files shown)


packages/calcite-components/src/utils/openCloseComponent.ts
```diff
@@ -38,7 +37,7 @@ export interface OpenCloseComponent {
 }
 
 function isOpen(component: OpenCloseComponent): boolean {
-  return "opened" in component ? component.opened : component.open;
+  return component[component.openProp || "open"];
```
Elijbet, line 40: Great! This will cover `expanded`/`collapsed`/`closed` states as well.

Elijbet (approved):

Issue #11078, opened 2024-12-17
Focus State Behavior for v3 List Items Appears Unexpected


AdamWMoqrane:
### Description  
The focus states for v3 List components are inconsistent, leading to alignment issues and unexpected focus behavior. This problem seems related to the structure of the component and the keyboard navigation of each "table cell" within the List. This issue persists even when using the `icon-start` property introduced in 3.0.

### User Stories  
As a product designer working with List components, I want the focus states to be standardized and consistent so that users have a predictable and accessible experience when navigating through list items.

### Acceptance Criteria  
- The focus behavior for List components works consistently across all focusable areas (icon, label, and actions).  
- Misalignment issues caused by fractional pixel rendering or hitbox inconsistencies are resolved.  
- Keyboard navigation within "table cells" behaves as expected.  
- Switching to the `content-start` or `icon-start` slots does not introduce misaligned or unintended focus behavior.

### Relevant Info  
- **Repro Steps**:  
   1. Use `icon-start` with v3 List components.  
   2. Observe the focus states and alignment when interacting with the List items.  
- **Codepen Example**: [Link to Codepen provided by @adam](https://codepen.io/mac_and_cheese/pen/QwLdOMm?editors=1000)  
- **Additional Observations**: Misalignment appears fractional depending on the component's internal structure and click targets.

### Helpful Details  
- The icons were placed in the default `content-start` slots, which exacerbates the misalignment issues.  
- Swapping to the `icon-start` slot improves alignment slightly but does not completely resolve focus behavior.  
- Conversations referenced:  
   - Adam Moqrane, Adam Tirella, Paul Coster, Aaron Shetland (see screenshots).  

### Priority impact

impact - p1 - need for current milestone

### Esri team

ArcGIS Knowledge

ashetland:
@geospatialem I think this focus behavior should be categorized as a bug. I also believe it's been with us for while. I believe this is the focus state in question where only the content area is focused:

![image](https://github.com/user-attachments/assets/b5610852-fb0d-4e19-b054-bc6505d2b921)


Ideally, focus should only be on the entire item or interactive elements, such as the drag handle, close button, or slotted actions. 
![image](https://github.com/user-attachments/assets/ee1553e9-ede2-4fcd-a090-484df0b52123)
![image](https://github.com/user-attachments/assets/16f13dcd-4d17-4262-9909-13dd3bdfb62c)

geospatialem:
[...]

PR #11310, opened 2025-01-15
fix(list): only focus on row when clicking a list item

driskull:
**Related Issue:** #11078

## Summary

- no longer focuses on cells when a cell is focused via click
- only a row or cell can be focusable via tabindex at a time
- update tests


eriklharper (approved):
This seems to work pretty well, better than my stab at it.  Thanks @driskull !

PR #11326, opened 2025-01-17
fix(combobox, input-time-zone): fix initial item selection delay

jcfranco:


packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -119,7 +115,9 @@ export class Combobox
 
   defaultValue: Combobox["value"];
 
-  private emitComboboxChange = debounce(this.internalComboboxChangeEvent, 0);
```
driskull, line 122: nice

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1229,15 +1229,19 @@ export class Combobox
     return this.filterText === "" ? this.items : this.items.filter((item) => !item.hidden);
   }
 
-  private updateItems = debounce((): void => {
```
driskull, line 1232: shouldn't we still debounce this in some cases? Like the mutation observer case?

driskull (approved):
👍

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -658,14 +656,14 @@ export class Combobox
...
-      } else {
-        items.forEach((item) => (item.selected = false));
-      }
+      this.getItems().forEach((item) => {
+        item.selected = Array.isArray(value)
+          ? value.includes(item.value)
+          : value
+            ? value === item.value
+            : false;
+      });
```
richiecarmichael, line 665:
```suggestion
      for (const item of this.getItems()) {
        item.selected = (Array.isArray(value) && value.includes(item.value)) ||
                        (!Array.isArray(value) && value === item.value);
      }
```

packages/calcite-components/src/components/combobox/combobox.tsx
jcfranco: We could keep an eye out for this. The MO callback kicks in once per batch of mutations, so I don't expect a significant increase in `updateItem` calls.

driskull: Sounds good. We may want to have some internal doc/guidelines on when we should debounce a function and under what circumstances.

PR #11332, opened 2025-01-17
test(combobox): use async expect for non-throwing test case

jcfranco:
**Related Issue:** N/A

## Summary

Adds async assertion to non-throwing test.

## References

* https://vitest.dev/api/expect.html#tothrowerror
* https://vitest.dev/api/expect
* https://jestjs.io/docs/asynchronous#asyncawait


driskull (approved):
👍

Issue #11261, opened 2025-01-10
Tooltips clipped in action-pad


MikeTschudi:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When one hovers over an icon in an action pad, one gets a vertical scrollbar because the tooltip is appearing and being hidden:
![image](https://github.com/user-attachments/assets/4203f7c3-3f67-4d31-a4e9-ca2f5c57b8d0)

### Expected Behavior

Expect tooltip to escape action-pad boundaries and appear:
![image](https://github.com/user-attachments/assets/e7eff373-76f5-4fa4-bb15-54bdfd1c820e)

https://codepen.io/miketschudi/pen/bNbvByq?editors=1000

### Reproduction Sample

https://codepen.io/miketschudi/pen/wBwmoZg?editors=1000

### Reproduction Steps

please see reproduction sample codepen

### Reproduction Version

3.0.0-next.95

### Relevant Info

Windows 11 Enterprise; Chrome 131

### Regression?

3.0.0-next.1

### Priority impact

impact - p3 - not time sensitive

### Impact

Currently only used in a prototype for release in the second AGO release of 2025.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Solutions

jcfranco:
This is caused by https://github.com/Esri/calcite-design-system/pull/10241 (3.0 breaking change), where components with floating-ui elements may need `overlay-positioning="fixed"` in scrolling containers. I'm not sure whether the scroller is intentional though.   

@driskull Do you recall if `action-pad` is meant to have a scrollbar? I noticed `overflow` has been present [since its initial port](https://github.com/Esri/calcite-design-system/blame/56b0da78ff4020978afb8c48835ac7ba5e52f49d/src/components/calcite-action-pad/calcite-action-pad.scss#L21).  

As a workaround, you can set `overlay-positioning="fixed"` on tooltips to help them escape their container.

macandcheese:
I think you could also move the Tooltips elements outside the Action Bar, which seems to resolve it while still allowing overlay-positioning="absolute" (default).

MikeTschudi:
[...]

PR #11349, opened 2025-01-21
refactor(action-pad): remove unnecessary overflow css styling.

driskull:
**Related Issue:** N/A

## Summary

- removes overflow styling which is unnecessary

### DIFF
--- a/packages/calcite-components/src/components/action-pad/action-pad.scss
+++ b/packages/calcite-components/src/components/action-pad/action-pad.scss
@@ -29,8 +29,7 @@
 .container {
   @apply shadow-2
   inline-flex
-  flex-col
-  overflow-y-auto;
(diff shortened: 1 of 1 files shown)


jcfranco:
We could treat this as a `refactor` or `chore` since it's cleaning up styles. Otherwise, we can reopen https://github.com/Esri/calcite-design-system/issues/11261 and link it. cc @geospatialem

driskull:
Updated to use refactor for now

jcfranco (approved):

driskull:
@nCastle1 I believe that change is unintentional. 

@jcfranco should we revert and apply `overflow:hidden`? `overflow-y:hidden` only makes sense when action pad is vertical.

Issue #11321, opened 2025-01-16
[`calcite-input-date-picker`]: min value is disabling month


jasdelreyes:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When the `min` is set to a day in a month different from the initially selected date, the whole month of the `min` value is disabled in the dropdown.

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/a7022544-cf1c-4947-88ef-28272e010e65" />

### Expected Behavior

The month of the `minDate`, when appropriate should be enabled in the dropdown. 

For example, if the minimum date were January 16th, January should still be enabled due to other days in January that are still valid.

### Reproduction Sample

https://codepen.io/jasdelreyes/pen/MYgXzEO

### Reproduction Steps

Open codepen link.

### Reproduction Version

~3.0.0-next.90

### Priority impact

impact - p1 - need for current milestone

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Online

anveshmekala:
Hi @jasdelreyes ,

Noticed `min` is assigned a value of type `Date` in the codepen instead of `string` value. If you wish to use Date type please leverage minAsDate/maxAsDate properties in future once https://github.com/Esri/calcite-design-system/issues/10329 is fixed.

jasdelreyes:
@anveshmekala Thanks for pointing that out. I've tried using the minAsDate/maxAsDate properties alternatively, but now the component does not honor the minimum date set and allows dates before the minimum date to be selected.

DitwanP:
🍡 Verified on `3.0.0-next.128`

PR #11350, opened 2025-01-21
fix(date-picker): no longer disable min/max value month in select menu

anveshmekala:
**Related Issue:** #11321 

## Summary

No longer disable min,max month's in select menu of date-picker.


benelan (approved):
:+1:

driskull (approved):
👍

PR #11351, opened 2025-01-21
fix(color-picker): fix error caused by initial value with a different format

jcfranco:
**Related Issue:** #10731  

## Summary  

Adds missing check to skip internal, equal values after transitioning to getters/setters in #10310.


driskull (approved):
Nice!

PR #11358, opened 2025-01-22
refactor(action-pad): restore rounded styling

driskull:
- restores rounded styling for action-pad
- existing screenshot test is updated


driskull:
Should have been caught in the theming story but that story is so huge. I think we should separate it into individual components.

jcfranco:
Should the title and/or description be rephrased as "restore/revisit rounded styling" or similar? It's not a revert because it doesn't restore the original changes.

Issue #11353, opened 2025-01-21
Chip - close button not visible on `kind="inverse"`


ashetland:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The close button is black when `closable` and `kind="inverse"`.

### Expected Behavior

The close button matches the text color.

### Reproduction Sample

https://codepen.io/ashetland/pen/GgKXVNJ?editors=1000

### Reproduction Steps

See that the close button icon is black.

### Reproduction Version

v2.13.2

### Regression?

v2.12.2

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

geospatialem:
This looks to be a token-related regression, @alisonailea can you take a look? 👀 

cc @brittneytewks

asangma:
We could def use this fix for an upcoming  enhancement. 👏🏾

geospatialem:
Verified with `3.0.0-next.123`

![Image](https://github.com/user-attachments/assets/fb1d877d-d5a3-4809-92d0-d553ebc5eddd)

PR #11363, opened 2025-01-22
fix(chip): fix inverse close color

alisonailea:
**Related Issue:** #11353

## Summary

Resolves issue where close button was not visible on a chip with `kind="inverse"`.
Updates the `withClosable` storybook test to include all versions of chip with `closable` to cover future use cases.


macandcheese (approved):
⭐

PR #11364, opened 2025-01-23
fix(flow): process items on loaded

jcfranco:
**Related Issue:** #10731 

## Summary

This moves `flow-item` processing to `loaded`, so all items are guaranteed to be ready.

**Note**: `items` is no longer decorated with `@state`, as it is not used in rendering.

### DIFF
--- a/packages/calcite-components/src/components/flow/flow.tsx
+++ b/packages/calcite-components/src/components/flow/flow.tsx
@@ -33,8 +33,10 @@ export class Flow extends LitElement implements LoadableComponent {
   private frameEl: HTMLDivElement;
 
   private itemMutationObserver: MutationObserver = createObserver("mutation", () =>
-    this.handleMutationObserverChange(),
+    this.updateItemsAndProps(),
   );
+
+  private items: FlowItemLikeElement[] = [];
 
   private selectedIndex = -1;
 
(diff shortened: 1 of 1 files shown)


jcfranco:
@driskull @macandcheese I wasn't able to reproduce this in our E2E test environment. Do you think a screenshot test would be an acceptable form of coverage?

driskull:
> I wasn't able to reproduce this in our E2E test environment. Do you think a screenshot test would be an acceptable form of coverage?

Yeah I think that should be fine.

packages/calcite-components/src/components/flow/flow.tsx
```diff
@@ -120,7 +120,6 @@ export class Flow extends LitElement implements LoadableComponent {
 
   override connectedCallback(): void {
     this.itemMutationObserver?.observe(this.el, { childList: true, subtree: true });
-    this.handleMutationObserverChange();
```
driskull, line 123: I think this should be fine since connecting a flow doesn't need to update items. The mutation observer should suffice when items are added. If we find an issue we can call on connectedCallback and loaded.

driskull (approved):
👍

jcfranco:
I merged before I could add a screenshot test. Will do that later today.

PR #11365, opened 2025-01-23
feat(flow, flow-item): add component tokens

driskull:
**Related Issue:** #7180

## Summary

🚧

### DIFF
--- a/packages/calcite-components/src/components/flow-item/flow-item.scss
+++ b/packages/calcite-components/src/components/flow-item/flow-item.scss
@@ -3,8 +3,30 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
- * @prop --calcite-flow-item-footer-padding: Specifies the padding of the component's footer.
- * @prop --calcite-flow-item-header-border-block-end: Specifies the component header's block end border.
+ * @prop --calcite-flow-item-footer-padding: [Deprecated] Use `--calcite-flow-footer-space` instead. Specifies the padding of the component's footer.
+ * @prop --calcite-flow-item-header-border-block-end: [Deprecated] Use `--calcite-flow-border-color` instead. Specifies the component header's block end border.
+ * @prop --calcite-flow-corner-radius: Specifies the component's corner radius.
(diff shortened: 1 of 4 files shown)


packages/calcite-components/src/components/flow-item/flow-item.scss
```diff
@@ -3,8 +3,30 @@
...
  *
- * @prop --calcite-flow-item-footer-padding: Specifies the padding of the component's footer.
- * @prop --calcite-flow-item-header-border-block-end: Specifies the component header's block end border.
+ * @prop --calcite-flow-item-footer-padding: [Deprecated] Use `--calcite-flow-footer-space` instead. Specifies the padding of the component's footer.
+ * @prop --calcite-flow-item-header-border-block-end: [Deprecated] Use `--calcite-flow-border-color` instead. Specifies the component header's block end border.
+ * @prop --calcite-flow-corner-radius: Specifies the component's corner radius.
+ * @prop --calcite-flow-heading-text-color: Specifies the text color of the component's `heading`.
+ * @prop --calcite-flow-description-text-color: Specifies the text color of the component's `description`.
+ * @prop --calcite-flow-border-color: Specifies the component's border color.
+ * @prop --calcite-flow-background-color: Specifies the component's background color.
```
macandcheese, line 12: Nitpick: It might be nice to apply this to the Flow container. Currently - if you override a Panel BG, and move between Flows - the animation reveals a different color background (slowed down for dramatic effect) :

Taken from current set up in 3.x with the Panel variables, so may not apply:
https://github.com/user-attachments/assets/df0e93d1-009d-4c08-bc80-e945201eed3d

macandcheese (approved):
Looks good! One question about potentially applying the `background-color` to the Flow container to prevent cases where the BG flashes through against a themed Panel BG.

packages/calcite-components/src/components/flow-item/flow-item.scss
driskull: Added!

PR #11367, opened 2025-01-23
chore(icon): improve icon load error message

jcfranco:
**Related Issue:** N/A

## Summary

This improves clarity by displaying the icon name and scale separately in the error message instead of an internal key used by icon utils:

```sh
"measure16" is not a valid calcite-ui-icon name
```

⬇️

```sh
measure (s) icon failed to load.
```

Also, this now uses the logger to display the `calcite` badge for additional context.


benelan (approved):
:+1:

PR #11369, opened 2025-01-23
fix(date-picker): remove outline for header actions

anveshmekala:
**Related Issue:** #11371 

## Summary

Removes outline for header actions which is causing transition on focus shift. 

This is introduced with https://github.com/Esri/calcite-design-system/pull/10836/ where outline exclusion from transition is ignored in Tailwind util.

BEGIN_COMMIT_OVERRIDE
omitted from changelog
END_COMMIT_OVERRIDE


benelan (approved):
:+1:

PR #11373, opened 2025-01-23
fix(popover, tooltip): drop relative-positioning to reduce risk of clipping

jcfranco:
**Related Issue:** #10731 

## Summary

Removes legacy `position: relative` to reduce chance of clipping (see https://floating-ui.com/docs/misc#clipping).

BEGIN_COMMIT_OVERRIDE
omitted from changelog
END_COMMIT_OVERRIDE


driskull (approved):
👍 🎉 Nice improvement!

Issue #11370, opened 2025-01-23
Cannot set combobox `value` if item values are not strings


nwhittaker:
**monday.com sync:** #18048689265

### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Given a `<calcite-combobox>` with `<calcite-combobox-item>` elements whose `value` props are not strings, setting the combobox's `value` does not select the matching combobox items.

This behavior stems from a disagreement between the combobox's `value` type (`string | string[]`) and the combobox-item's `value` type (`any`).

### Expected Behavior

I believe the actual behavior is correct, however the components' typing should be such that it's not possible to use the components in this way without seeing type errors in JSX.

Agreement should exist between the combobox's and combobox-item's `value` types. Preferably the combobox-item's `value` type is `string`.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/jENeJOo

### Reproduction Steps

1. Visit the repro.
2. Enter a number between 0-9 in the **Set value** input and observe the combobox's value changes.
3. Set the **Value type** control to _Number_.
2. Enter a number between 0-9 in the **Set value** input and observe the combobox's value does not change.

### Reproduction Version

2.13.2

### Relevant Info

Other input components with a similar parent/child structure may have the same typing inconsistency: `segmented-control`, `select`, etc.

### Priority impact

impact - p3 - not time sensitive

### Impact

The weak typing increases the complexity of code that generates the mark-up for these types of inputs. Specifically, a sufficiently abstract consumer cannot rely on the combobox's `value` for setting which items are selected. Instead, it must iterate through and set the `selected` prop of each item.

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Field Apps

driskull:
> Preferably the combobox-item's value type is string.

Agreed 👍 

@jcfranco can we update components that type `value` as `any` to be `string`?

geospatialem:
To carry out into two phases:
1. Since the `value` is needed for form support, we should provide additional documentation context WRT string values.
2. With the next breaking change release, change the `type`

geospatialem:
Blocked by https://github.com/Esri/calcite-design-system/issues/13970

PR #11383, opened 2025-01-25
fix(combobox): improve prop update times

jcfranco:
**Related Issue:** #10731 

## Summary

Fixes prop update delays introduced in #10310 by moving event emitting to Lit's [pre-update phase](https://lit.dev/docs/components/lifecycle/#reactive-update-cycle).

### DIFF
--- a/packages/calcite-components/src/components/combobox/combobox.tsx
+++ b/packages/calcite-components/src/components/combobox/combobox.tsx
@@ -252,6 +252,8 @@ export class Combobox
 
   private selectedIndicatorChipEl: Chip["el"];
 
+  private _selectedItems: HTMLCalciteComboboxItemElement["el"][] = [];
+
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -741,8 +749,8 @@ export class Combobox
   }
 
   private getValue(): string | string[] {
-    const items = this.selectedItems.map((item) => item?.value?.toString());
-    return items?.length ? (items.length > 1 ? items : items[0]) : "";
+    const items = this.selectedItems.map((item) => item.value?.toString());
```
driskull, line 752: This `.toString()` reminds me of #11370 where we may want to consider changing value from `any` to `string`.

driskull (approved):
Nice! 💯

Issue #11378, opened 2025-01-24
Tooltip impacts layout


nwhittaker:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The `<calcite-tooltip>` element itself is treated as a component with layout that impacts the flow of content. For example, within a flex layout the tooltip takes up a content box:

<img width="388" alt="Image" src="https://github.com/user-attachments/assets/c73c0d6d-be89-405a-975b-b7a6aaaa7c3c" />

### Expected Behavior

The tooltip element does not have an impact on layout.

<img width="388" alt="Image" src="https://github.com/user-attachments/assets/ac907d4c-77e7-4dd6-89bd-3e96c8f23530" />

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/MYgzWEK

### Reproduction Steps

1. Visit the sample and possibly wait awhile for the unkpg.com resources to load.
2. Compare the displayed layout with the expected vs. actual layouts.

### Reproduction Version

3.0.0-next.111

### Relevant Info

Suggested fix is to have the tooltip style its `:host` with `display: contents;`.

### Regression?

2.13.2

### Priority impact

impact - p3 - not time sensitive

### Impact

The issue is easy to work around by giving the tooltip `contents` display externally. However, this issue does represent a backwards compatibility break with how tooltips impacted layout in `2.13.2`.

For a specific use case, we have informational icons associated with input labels in a "space-between" flex layout. After upgrading to Calcite 3, the label space is truncated and the icon is positioned in the center of layout instead of aligned with the trailing edge.

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Field Apps

driskull:
> Suggested fix is to have the tooltip style its :host with display: contents;.

I like that suggestion. This would be a breaking change since we're changing the default display from `block` to `contents` but it could fall under the breaking changes for 3.0. Thoughts @jcfranco 

We should also do this for `popover` if we do this for `tooltip`.

driskull:
Previously, tooltip host elements were also displayed as block but they were positioned absolute or fixed so they wouldn't take up any space.

geospatialem:
[...]

PR #11384, opened 2025-01-25
fix(popover, tooltip): change display to contents

driskull:
**Related Issue:** #11378

## Summary

- change default display on `tooltip` and `popover` from `block` to `contents`.

BEGIN_COMMIT_OVERRIDE
omitted from changelog
END_COMMIT_OVERRIDE


macandcheese (approved):
Code changes look good! Is there a quick explainer of the pros / any cons of this change or things implementers should look out for?

driskull:
Here's a good explainer: https://css-tricks.com/get-ready-for-display-contents/

> One answer to this is display: contents;—a magical new display value that essentially makes the container disappear, making the child elements children of the element the next level up in the DOM.

> display: contents makes that the div doesn’t generate any box, so its background, border and padding are not rendered. However the inherited properties like color and font have effect on the child (span element) as expected.

This should be fine since we don't want the host of the tooltip/popover to have any dimensions or affect any parent elements dimensions.

driskull:
@jcfranco can you review?

macandcheese:
> Something to watch out for is that setting a width/height on a tooltip would no longer take up any space. Although this is a pro since it doesn't have any advantage to it.
> 
> https://codepen.io/Matt-Driscoll/pen/XJroEdr?editors=1010

Cool. Same for Popover then - folks may just need to move any set widths on Popover to a child element it sounds like.

jcfranco (approved):

jcfranco:
Noice! 

Should this use `fix` (+ `no changelog entry` label )instead? It addresses the layout regression from https://github.com/Esri/calcite-design-system/issues/10240 by changing the display mode.

Issue #10832, opened 2024-11-21
deprecate -pressed chip tokens


alisonailea:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

--calcite-chip-background-color: Specifies the component's background color.
--calcite-chip-border-color: Specifies the component's border color.
--calcite-chip-corner-radius: Specifies the component's corner radius.
--calcite-chip-text-color: Specifies the component's text color.
--calcite-chip-icon-color: Specifies the component's icon color.
--calcite-chip-close-icon-color: Specifies the component's close element icon color.
--calcite-chip-select-icon-color: Specifies the component's selection element icon color.
--calcite-chip-select-icon-color-pressed: Specifies the component's selection element icon color when active.

### Expected Behavior

--calcite-chip-background-color: Specifies the component's background color.
--calcite-chip-border-color: Specifies the component's border color.
--calcite-chip-corner-radius: Specifies the component's corner radius.
--calcite-chip-text-color: Specifies the component's text color.
--calcite-chip-icon-color: Specifies the component's icon color.
--calcite-chip-close-icon-color: Specifies the component's close element icon color.
--calcite-chip-select-icon-color: Specifies the component's selection element icon color.
--calcite-chip-select-icon-color-pressed: [Deprecate] Use `--calcite-chip-select-icon-color-press`. Specifies the component's selection element icon color when active.
--calcite-chip-select-icon-color-press: Specifies the component's selection element icon color when active.
:host {

### Reproduction Sample

N/A

### Reproduction Steps

N/A

### Reproduction Version

N/A

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

DitwanP:
Verified locally on `dev`

PR #11389, opened 2025-01-27
feat(chip): deprecate `pressed` in favor of `press`

alisonailea:
**Related Issue:** #10832

## Summary

Deprecate `--calcite-chip-select-icon-color-pressed` token.

Add `--calcite-chip-select-icon-color-press: Specifies the component's selection element icon color when active.`


packages/calcite-components/src/components/chip/chip.scss
```diff
@@ -10,7 +10,8 @@
  * @prop --calcite-chip-icon-color: Specifies the component's icon color.
  * @prop --calcite-chip-close-icon-color: Specifies the component's close element icon color.
  * @prop --calcite-chip-select-icon-color: Specifies the component's selection element icon color.
- * @prop --calcite-chip-select-icon-color-pressed: Specifies the component's selection element icon color when active.
+ * @prop --calcite-chip-select-icon-color-pressed: [deprecate] Use `--calcite-chip-select-icon-color-press`. Specifies the component's selection element icon color when active.
+ * @prop --calcite-chip-select-icon-color-press: Specifies the component's selection element icon color when active.
```
driskull, line 14: @alisonailea it doesn't look like this new var is used anywhere? Should there also be tests here?

driskull (changes requested):
I think this should be a `feat` as well since its adding a new token and deprecating an old one.

It should have tests added, story, and update the HTML page with an example using the token.

packages/calcite-components/src/components/chip/chip.scss
```diff
@@ -5,12 +5,13 @@
  *
  * @prop --calcite-chip-background-color: Specifies the component's background color.
  * @prop --calcite-chip-border-color: Specifies the component's border color.
+ * @prop --calcite-chip-close-icon-color: Specifies the component's close element icon color.
  * @prop --calcite-chip-corner-radius: Specifies the component's corner radius.
- * @prop --calcite-chip-text-color: Specifies the component's text color.
  * @prop --calcite-chip-icon-color: Specifies the component's icon color.
- * @prop --calcite-chip-close-icon-color: Specifies the component's close element icon color.
+ * @prop --calcite-chip-select-icon-color-press: Specifies the component's selection element icon color when active.
+ * @prop --calcite-chip-select-icon-color-pressed: [deprecate] Use `--calcite-chip-select-icon-color-press`. Specifies the component's selection element icon color when active.
```
driskull, line 12: Should say `[deprecated` instead of `[deprecate]`. May apply elsewhere

driskull (approved):
LGTM! once the doc is fixed.

I think the title of the PR should also be:

```
feat(chip): add `press` tokens and deprecate `pressed` tokens
```

packages/calcite-components/src/components/chip/chip.scss
alisonailea: double checked. Does not apply elsewhere

jcfranco:
@alisonailea Can you amend the PR title as [suggested above](https://github.com/Esri/calcite-design-system/pull/11389#pullrequestreview-2612878891)? From the changelog perspective, "pressed"/"press" isn't clear on its own, so tokens context is needed.  

[Here are the steps](https://github.com/Esri/calcite-design-system/wiki/monorepo#edit-changelog-entries) to amend the conventional commit entry (see [example](https://github.com/Esri/calcite-design-system/pull/9907)).

Issue #10571, opened 2024-10-21
[Meter] Ensure the component provides support for high contrast


geospatialem:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates and reviewed the [Accessibility page](https://developers.arcgis.com/calcite-design-system/foundations/accessibility) for guidance.

### Summary

Ensure the `meter` component provides sufficient high contrast support to more users. Chatted with @mpriour and this is considered a high priority since the component does not provide context to its value without accompanying text.

#### "Normal"
![image](https://github.com/user-attachments/assets/be2a9091-9a5b-422c-a888-8bd389db9626)

#### High contrast
![image](https://github.com/user-attachments/assets/ed123206-295e-4263-8828-1b983ce4aa5e)

### Expected Behavior

Similar functionality support as provided in `progress` 

https://github.com/Esri/calcite-design-system/blob/5f2d22531b72dcf6c602ab8b89f83357c8f5b829/packages/calcite-components/src/components/progress/progress.scss#L34-L41

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/meter/#sample

### Reproduction Steps

1. Open the [sample](https://developers.arcgis.com/calcite-design-system/components/meter/#sample)
2. Turn on Windows high contrast mode
3. Observe the behavior

### Reproduction Version

2.13.1

### Relevant Info

Windows

### Regression?

No

### Priority impact

impact - p1 - need for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

geospatialem:
Verified with `3.0.0-next.116` with Windows high contrast mode across browsers 🎉 

![Image](https://github.com/user-attachments/assets/0d6c7af9-06bb-4287-b99a-53de395db950)

PR #11392, opened 2025-01-27
fix(meter): add fill style for high-contrast mode

mpriour:
**Related Issue:** #10571 

## Summary

Add `forced-colors` media query to the styles to define a solid color when in high-contrast accessibility mode.


macandcheese (approved):
Thanks @mpriour !

geospatialem (approved):
Looks great, thanks for the quick turnaround, @mpriour! 🚀

Issue #11398, opened 2025-01-28
Ensure looped assertions handle empty `findAll` results to avoid false positives


jcfranco:
### Priority impact

p - high

### Test type

test update

### Which Component(s)

N/A

### PR skipped, if applicable

#

### Additional Info

Consider introducing a utility or extending E2EPage with APIs that will help avoid false positives from the following:

```ts
const items = await page.findAll("some-component-not-present");

for (const item of items) {
  // ... assertions
}
```

Examples:

* https://github.com/Esri/calcite-design-system/blob/bc844231886f47266c1ce99cdfc79c2db6b52960/packages/calcite-components/src/components/accordion/accordion.e2e.ts#L95-L102
* https://github.com/Esri/calcite-design-system/blob/682073f47b59a24eecd6cc6957d89c0158dd2d0c/packages/calcite-components/src/components/combobox/combobox.e2e.ts#L2507-L2512

Thanks for spotting this, @josercarcamo!

DitwanP:
🍠 Verified locally on `dev`

PR #11401, opened 2025-01-28
test: fix empty findAll results

jcfranco:
**Related Issue:** #11398 

## Summary

This introduces a replacement to using `page.findAll`/`element.findAll` to avoid false positives from asserting on empty results (e.g., assertions in a loop). 

Also, this will be enforced by the [`no-restricted-properties`](https://eslint.org/docs/latest/rules/no-restricted-properties) ESLint rule.


driskull (approved):
👍

jcfranco:
@Elijbet Can you help me fix the remaining failing tests? It looks like we missed some subtly invalid selectors when working on https://github.com/Esri/calcite-design-system/issues/6038 some time ago.

eriklharper (approved):
This looks very useful, thanks @jcfranco !

jcfranco:
@Elijbet I ended up skipping those tests as they were silently broken and created https://github.com/Esri/calcite-design-system/issues/11407 to restore them.

Issue #11397, opened 2025-01-28
[calcite-list] Change event indexes are offset by non list-item siblings


LeviRemi:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When the calcite-list possesses children which are not of the calcite-list-item or calcite-list-item-group type, then both the `oldIndex` and the `newIndex` values of the `calciteListOrderChange` event (as fired by a click-and-drag action) are offset.

![Image](https://github.com/user-attachments/assets/8b8129a7-08fb-400b-a153-f3dd49980882)

This offset of 2 is explained by the presence of two non list-item siblings: calcite-action and calcite-tooltip:
![Image](https://github.com/user-attachments/assets/e83bb348-00c6-4134-88db-6cc4b4dddc16)

If instead the button is used on the drag-handler, then indexes remain accurate regardless of siblings:
![Image](https://github.com/user-attachments/assets/5dd38e38-eb0b-4e00-b077-64fa60712c67)

### Expected Behavior

I would expect the `oldIndex` and the `newIndex` to be calculated from only the actual list items within the calcite-list. I would expect non list items to have zero influence on the index during an order change, because non list items are not draggable and cannot otherwise be reordered within the calcite-list framework.

### Reproduction Sample

https://codepen.io/leviremi/pen/PwYXdvw

### Reproduction Steps

1. Drag and drop any list item into a new position.
2. See that the oldIndex and newIndex are incremented by 1 for every non calcite-list-item child that are added before the list items (in the sample's case, a calcite-action and a calcite-tooltip).
3. Now instead of dragging, try clicking the drag button and use that to move a list item.
4. See that the oldIndex and newIndex are accurate.

### Reproduction Version

3.0.0-next.115

### Relevant Info

Chrome 131.0.6778.265 (Official Build) (x86_64) 

### Priority impact

impact - p2 - want for an upcoming milestone

### Impact

This issue breaks list reordering compatibility in our app, since our index handling logic is dependent on the values reported by the event. We could simply re-offset the index to restore accuracy, but then the index will be incorrect whenever the user re-orders the list via the drag button instead.

### Calcite package

- [x] @esri/calcite-components
- [x] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

[...]

PR #11402, opened 2025-01-28
fix(list): drag event properties `newIndex` and `oldIndex` should only reflect list item indexes

driskull:
**Related Issue:** #11397

## Summary

- use `oldDraggableIndex` and `newDraggableIndex` from https://github.com/SortableJS/Sortable?tab=readme-ov-file#event-object-demo in order to only track the relevant indexes of draggable items via mouse.
- updates keyboard logic for indexes to only consider applicable items.
- updates tests to add slotted content which should not affect event `newIndex/oldIndex`.

### DIFF
--- a/packages/calcite-components/src/components/list/list.e2e.ts
+++ b/packages/calcite-components/src/components/list/list.e2e.ts
@@ -1479,6 +1479,15 @@ describe("calcite-list", () => {
       const page = await newE2EPage();
       await page.setContent(
         html`<calcite-list drag-enabled id="list1">
+          <calcite-action
+            slot="filter-actions-end"
+            scale="s"
+            id="filter-action-test"
+            icon="show-all-parameters"
+          ></calcite-action>
+          <calcite-tooltip label="scary tooltip" reference-element="filter-action-test"
+            >Mind if I offset your index?</calcite-tooltip
+          >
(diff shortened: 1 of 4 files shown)


packages/calcite-components/src/components/list/list.e2e.ts
```diff
@@ -1479,6 +1479,15 @@ describe("calcite-list", () => {
...
       await page.setContent(
         html`<calcite-list drag-enabled id="list1">
+          <calcite-action
+            slot="filter-actions-end"
+            scale="s"
+            id="filter-action-test"
+            icon="show-all-parameters"
+          ></calcite-action>
+          <calcite-tooltip label="scary tooltip" reference-element="filter-action-test"
+            >Mind if I offset your index?</calcite-tooltip
```
eriklharper, line 1489: lol, love the labels

eriklharper (approved):
LGTM!  Did have just one curiosity question: is the `newDraggableIndex` and `oldDraggableIndex` something that SortableJS added since this component started using SortableJS?

driskull:
> LGTM! Did have just one curiosity question: is the newDraggableIndex and oldDraggableIndex something that SortableJS added since this component started using SortableJS?

Seems like its been there for quite a while

Issue #11382, opened 2025-01-25
Combobox `value` and `selectedItems` props stale in `calciteComboboxChange` handler after adding a custom value


nwhittaker:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Given a `<calcite-combobox>` with `allow-custom-values` enabled, typing in a custom value triggers the `calciteComboboxChange` event. However, in the event's handler, the target combobox's `valuel` and `selectedItems` props do not include the selected item.

https://github.com/user-attachments/assets/831cee8f-f042-42b5-9131-6d42ea1d2b3b

### Expected Behavior

The target combobox's `value` and `selectedItems` props do include the selected item by the time the `calciteComboboxChange` emits.

https://github.com/user-attachments/assets/ea4cd0d4-996a-4648-9213-b5b32789962e

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/vEBQeyG

### Reproduction Steps

1. Visit the repro and open the dev tools console.
2. Click the combobox and type in a custom value.
3. Observe the logged `combobox.value` and `combobox.selectedItems` values do not include the item selected in step 2.

### Reproduction Version

3.0.0-next.113

### Relevant Info

Possibly related to https://github.com/Esri/calcite-design-system/issues/11381.

### Regression?

2.13.2

### Priority impact

impact - p0 - emergency

### Impact

A workaround is to derive the `value` and `selectedItems` props from the combobox's child elements. However it loses the order in which values were entered, and it'd be a high level of effort to fix for every combobox instance in our code base. Requesting p0 priority as Field Maps is looking to release with Calcite 3 in a few weeks.

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Field Apps

nwhittaker:
Updated issue to clarify bug stems from using `allow-custom-values` and also impacts the `value` prop.

geospatialem:
Verified in `3.0.0-next.126`:

![Image](https://github.com/user-attachments/assets/d48a83d1-9fec-4d5a-83cf-5fb1119865bc)

PR #11405, opened 2025-01-29
fix(combobox): update internal state after custom value is added

jcfranco:
**Related Issue:** #10731, #11382

## Summary

This updates the combobox internal state after an custom item is added.

**Note**: this also adds `createEventTimePropValuesAsserter` to help assert on event-emit time component state.


driskull (approved):
LGTM! updateItems being called on adding chip should be fine since it doesn't occur often.

jcfranco:
@driskull Agreed. We might need to refactor combobox to update state in a more granular fashion. E.g., adding a custom item should update internal items and its filtering data, but probably doesn't need to filter at that time.

Issue #9926, opened 2024-07-31
[Shell Panel] `display-mode="float-all"` has multiple bugs


ashetland:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The new `float-all` configuration does not match other floating UI and the resize affordance is misplaced.

### Expected Behavior

- [ ] should have rounded corners
- [ ] add box shadow
- [ ] remove border around perimeter
- [ ] fix double border between Action Bar and Panel
- [ ] disable float resizing until #5652 lands

### Reproduction Sample

https://codepen.io/ashetland/pen/zYVZBPG?editors=1000

### Reproduction Steps

1. Open the codepen
2. Note the sharp corners, no shadow, perimeter border, double border next to Action Bar, and the odd placement of the drag affordance

### Reproduction Version

v2.11.0

### Priority impact

impact - p1 - need for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

macandcheese:
In a later version, `2.13.2` - the height of the Shell Panel seems too tall, like it's not compensating for the margin when in `float-all` - maybe a `block-size: calc(100% - 1rem);` or otherwise is needed (Note - this didn't exist in the version of original Codepen, but posting here for visibility) cc @ashetland https://codepen.io/mac_and_cheese/pen/oNKrrWZ?editors=1000

macandcheese:
@ashetland for the following item:

> disable float-all resizing until https://github.com/Esri/calcite-design-system/issues/5652 lands

Can we just leave the functionality and then improve it when that lands? This way any implementations of that behavior will continue to work and just benefit from future updates.

ashetland:
Sounds good to me.

geospatialem:
Verified the changes in `3.0.0-next.126`. 

Did observe when `resizable` is present and the component is in focus a line is added for resizing the component via https://codepen.io/geospatialem/pen/ByBeGdK, is that intended or do we need an additional follow-up issue to tackle? cc @macandcheese @ashetland 

![Image](https://github.com/user-attachments/assets/442d2090-5690-47c6-8d10-271b548ef6d8)

macandcheese:
[...]

Issue #10050, opened 2024-08-13
Replace inconsistent shadow styles with new shadow tokens


rmstinson:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

A handful of Calcite components use a different shadow style than the tailwind style `shadow.2`. Most of them currently use the style "0 0 16px 0 rgba(0, 0, 0, 0.16)." Some are inheriting the style from the classes "calcite-floating-ui-anim" + "calcite-floating-ui-anim--active." Those components' shadow styles need to be replaced with `shadow.2` in the code.

### Acceptance Criteria
- [x] Ensure the opacities used in shadow tokens are set in core
    - [x] 0.04
    - [x] 0.08
    - [x] 0.10
    - [x] 0.12
    - [x] 0.16

<details>
<summary>Some criteria moved to #14335</summary>

> - [ ] Add new semantic black color tokens for use with shadows:
> 
>     - `color.shadow.1`
>         - light mode: rgba({core.color.neutral.blk-240}, {core.opacity.4})
>         - dark mode: rgba({core.color.neutral.blk-240}, {core.opacity.12})
>     - `color.shadow.2`
>         - light mode: rgba({core.color.neutral.blk-240}, {core.opacity.8})
>         - darkmode: rgba({core.color.neutral.blk-240}, {core.opacity.12})
>     - `color.shadow.3`
>         - light mode: rgba({core.color.neutral.blk-240}, {core.opacity.10})
>         - dark mode: rgba({core.color.neutral.blk-240}, {core.opacity.16})
> 
> - [ ] Update semantic shadow tokens:
> 
>     - `shadow.sm` = box-shadow: 0px 2px 8px 0px {semantic.color.shadow.3}, 0px 2px 8px 0px {semantic.color.shadow.1};
>     - `shadow.md` = box-shadow: 0px 6px 20px -4px {semantic.color.shadow.3}, 0px 4px 12px -2px {semantic.color.shadow.2};
</details>

- [x] Update semantic shadow tokens:
    - `shadow.sm` = box-shadow: 0px 2px 8px 0px rgba({core.color.neutral.blk-240}, {core.opacity.10}), 0px 2px 8px 0px rgba({core.color.neutral.blk-240}, {core.opacity.4});
[...]

Issue #5652, opened 2022-10-26
[Shell Panel] Improve visibility of resizable functionality, allow `resizable` when `displayMode="float"`


macandcheese:
### Description

The current drag handle affordance can be a bit hard to notice. 
This should support both Shell Panel and upcoming Sheet implementations. 

### Acceptance Criteria

- [x] The resizable functionality of Shell Panel is easier to notice. This should also in certain cases be visually large enough (or have a hitbox large enough) to support thumb gestures for dragging.
- [x] Support `resizable` when the component is also `detached`, when `displayMode` is set to `"float"`

### Relevant Info
Questions:

- Should this have a "click to collapse" and "click to expand" button - brought up as UC feedback?

### Which Component

Shell Panel, Sheet (upcoming)

### Example Use Case

cc @asangma @SkyeSeitz @ashetland 

### Esri team

Calcite (design)

### Calcite Design Solution
<img width="1753" alt="image" src="https://github.com/user-attachments/assets/52adaa0f-7fdf-4935-9a68-89efe890971e">

🔗[ Link to Figma Specs]( https://www.figma.com/design/A7Z8XlYSkv5FSGwIJGoNu3/%5BShell-Panel%2C-Sheet%2C-Table%5D-%E2%80%9Cresizable%E2%80%9D-%5B7591%2C-5652%5D?node-id=1159-54157&m=dev&ready-for-dev=1)

**monday.com sync:** #5887227671

macandcheese:
cc @SkyeSeitz @ashetland

macandcheese:
+1 from BA cc @ddred11

SkyeSeitz:
Final specs can be [found here](https://www.figma.com/file/A7Z8XlYSkv5FSGwIJGoNu3/%5BShell-Panel%2C-Sheet%2C-Table%5D-%E2%80%9Cresizable%E2%80%9D-%5B7591%2C-5652%5D?type=design&node-id=1%3A1435&mode=dev).

COV-GIS:
I had been working my own footer component with resizing until it was brought to my attention a shell-panel works better and has resizing built in. At first I thought resizing wasn't working. I couldn't see the drag bar and when I did it was real hard to use on a large hi-res screen. And it isn't intuitive to the user there's resizing even if they notice the bar.

[...]

PR #11410, opened 2025-01-30
fix(shell-panel): Fix visual issues in `displayMode="float-all"`

macandcheese:
**Related Issue:** #9926 

## Summary
Adds some styling to address the visual issues in Shell Panel.

Screenshot 2025-01-29 at 6 58 09 PM
Screenshot 2025-01-29 at 6 57 27 PM

Note - I went down the rabbit hole of trying to conditionally pass corner-radius to slotted Actions, Panels, etc. - it would involve a lot of internal tokens and conditional states to handle various arrangements of components. This could be something to look at in the future but I'm not sure we want to handle this here.

Screenshot 2025-01-29 at 6 54 39 PM


packages/calcite-components/src/components/shell-panel/shell-panel.scss
```diff
@@ -123,6 +123,25 @@
   }
 }
 
+.float-all {
+  max-block-size: var(--calcite-internal-shell-panel-max-height, calc(100% - 1rem));
+  box-shadow: var(--calcite-shadow-sm);
```
driskull, line 128: Does this need a token for customization?

packages/calcite-components/src/components/shell-panel/shell-panel.scss
```diff
@@ -123,6 +123,25 @@
...
+  border-radius: var(--calcite-corner-radius-round);
```
driskull, line 129: Does this need a token for customization?

packages/calcite-components/src/components/shell-panel/shell-panel.scss
```diff
@@ -123,6 +123,25 @@
...
+  overflow: hidden;
+}
+
+:host([layout="vertical"][position="start"]) .float-all {
+  border-inline-start: 1px solid var(--calcite-color-border-3);
```
driskull, line 134: Does this need a token for customization?  applies to other places

driskull (approved):
LGTM! Do we need tokens for customizing these new styles?

packages/calcite-components/src/components/shell-panel/shell-panel.scss
macandcheese: Yeah, I wonder how we want to handle this. We have a bunch of non-tokenized, directionally-set shadows we'd probably want to allow updates to as well, for other `displayMode` like `"overlay"`. 

cc @ashetland @SkyeSeitz @matgalla  any ideas for pattern-ifying these? These have been hanging out for awhile... https://github.com/Esri/calcite-design-system/blob/fe3a87685d5e1544d5f3b6548ef9f826beea51fb/packages/calcite-components/src/components/shell-panel/shell-panel.scss#L24

ashetland: We were just discussing this! We have issue #10050 that we're going to repurpose for defining new shadow tokens. We specifically want to account for this need in that effort. Hoping to wrap that up in the coming weeks.

ashetland:
Screenshots are looking good, but the sizing behavior is weird. The edge floats in space. 
CleanShot 2025-01-30 at 09 51 37@2x

ashetland:
I see we removed that requirement from the issue, so maybe this can be ignored in this pr?

macandcheese:
> I see we removed that requirement from the issue, so maybe this can be ignored in this pr?

Yeah, the original issue asked to remove, but that would be breaking - I think we should leave as-is, since this PR doesn't change that behavior. This issue should be additive improvement: https://github.com/Esri/calcite-design-system/issues/5652

packages/calcite-components/src/components/shell-panel/shell-panel.scss
macandcheese: Ok, for now I'll add one for this new update, leave the directional ones alone, and document it in a way that makes it clear it only affects this `float-all` displayMode.

packages/calcite-components/src/components/shell-panel/shell-panel.scss
ashetland: I see this happen in Popover and also in Dialog (see the top corners)
CleanShot 2025-01-30 at 11 45 38@2x

Spoke to @jcfranco about this a while ago and he had similar concerns you outline in your second bullet, but also thought it could be worth looking into. Might adversely affect floating ui slotted into a Dialog...

ashetland (approved):
Visuals lookin' good!

packages/calcite-components/src/components/shell-panel/shell-panel.scss
macandcheese: I guess this component hasn't had its component tokens looked at: https://github.com/Esri/calcite-design-system/issues/7180

Should I add just the few new ones for this pr, and leave the others for a further pass @driskull ?

driskull: > Should I add just the few new ones for this pr, and leave the others for a further pass

Yes that sounds good 👍

packages/calcite-components/src/components/shell-panel/shell-panel.scss
macandcheese: Yeah, we could approach it in a few ways, all have pros / cons. For now, I think I'll leave this as-is, since it's not unique to Shell Panel, and should be done in a way that addresses all patterns like this.

packages/calcite-components/src/components/shell-panel/shell-panel.scss
macandcheese: Added - @geospatialem or @DitwanP can you take a look at these new css properties? Note that existing ones have yet to be audited as part of #7180 so may be out of date.

packages/calcite-components/src/components/shell-panel/shell-panel.scss
macandcheese: @jcfranco ok to proceed for now?

packages/calcite-components/src/components/shell-panel/shell-panel.scss
DitwanP: @macandcheese LGTM!

packages/calcite-components/src/components/shell-panel/shell-panel.scss
```diff
@@ -10,6 +10,10 @@
  * @prop --calcite-shell-panel-max-height: When `layout` is `horizontal`, or `layout` is `vertical` and `displayMode` is `float-content` or `float`, specifies the maximum height of the component.
  * @prop --calcite-shell-panel-min-height: When `layout` is `horizontal`, or `layout` is `vertical` and `displayMode` is `float-content` or `float`, specifies the minimum height of the component.
  * @prop --calcite-shell-panel-z-index: Specifies the z-index value for the component.
+ * @prop --calcite-shell-panel-border-color: When `displayMode` is `"float-all"`, specifies the border color of the component.
+ * @prop --calcite-shell-panel-corner-radius: When `displayMode` is `"float-all"`, specifies the corner radius of the component.
+ * @prop --calcite-shell-panel-shadow: When `displayMode` is `"float-all"`, specifies the shadow of the component.
```
geospatialem, line 15: nitpick for doc consistency 📚:

```suggestion
 * @prop --calcite-shell-panel-border-color: When `displayMode` is `"float-all"`, specifies the component's border color.
 * @prop --calcite-shell-panel-corner-radius: When `displayMode` is `"float-all"`, specifies the component's corner radius.
 * @prop --calcite-shell-panel-shadow: When `displayMode` is `"float-all"`, specifies the component's shadow.
```

packages/calcite-components/src/components/shell-panel/shell-panel.scss
geospatialem: @macandcheese Had some small nitpicks for consistency below via https://github.com/Esri/calcite-design-system/pull/11410#discussion_r1939764047, but otherwise LGTM! 🚀

packages/calcite-components/src/components/shell-panel/shell-panel.scss
macandcheese: Updated 👍

jcfranco (approved):
🚀

PR #11418, opened 2025-01-30
fix(form): add ariaHidden property to the hidden form input

driskull:
**Related Issue:** #11419

## Summary

- adds `ariaHidden = "true"` property to the hidden form input when it is created
- adds test in autocomplete
- add formAssociated test to make sure hidden inputs are setting ariaHidden properly

### DIFF
--- a/packages/calcite-components/src/utils/form.tsx
+++ b/packages/calcite-components/src/utils/form.tsx
@@ -499,7 +499,7 @@ function syncHiddenFormInput(component: FormComponent): void {
 
     if (!input) {
       input = ownerDocument.createElement("input");
-      input.setAttribute("aria-hidden", "true");
+      input.ariaHidden = "true";
       input.slot = hiddenFormInputSlotName;
     }
 
--- a/packages/calcite-components/src/utils/form.tsx
+++ b/packages/calcite-components/src/utils/form.tsx
@@ -499,6 +499,7 @@ function syncHiddenFormInput(component: FormComponent): void {
 
     if (!input) {
       input = ownerDocument.createElement("input");
+      input.setAttribute("aria-hidden", "true");
       input.slot = hiddenFormInputSlotName;
     }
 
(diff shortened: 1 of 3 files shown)


jcfranco:
> syncHiddenFormInput doesn't have any tests. We can maybe add some at some point or just wait for elementInternals

Here's the spec test for `syncHiddenFormInput`: https://github.com/Esri/calcite-design-system/blob/4a1ed6eb18a57dfd68261ba27ad3945ed3c615cd/packages/calcite-components/src/components/input/common/input.spec.ts#L5.

packages/calcite-components/src/utils/form.tsx
```diff
@@ -499,6 +499,7 @@ function syncHiddenFormInput(component: FormComponent): void {
 
     if (!input) {
       input = ownerDocument.createElement("input");
+      input.setAttribute("aria-hidden", "true");
```
jcfranco, line 502: For consistency, can we set the property instead?

driskull:
> Here's the spec test for syncHiddenFormInput:

@jcfranco thats a different function. It's not used by `form.tsx`, which has a different function.

jcfranco:
It does get used by it: https://github.com/Esri/calcite-design-system/blob/a43ecfe0da9b3b6c1b0a3d35bbf98b7dc4195f34/packages/calcite-components/src/utils/form.tsx#L558

jcfranco:
I had the wrong helper in mind. @driskull pointed out that I was referring to the one for `input.ts` instead of `form.tsx`. We can still add coverage for this by adding a similar test [to this](https://github.com/Esri/calcite-design-system/blob/bbd57b1b7583134bc110736ba8221c768dd1bd10/packages/calcite-components/src/tests/commonTests/formAssociated.ts#L219) or refactoring it to focus on hidden input syncing.

jcfranco (approved):
Noice!

PR #11429, opened 2025-02-01
fix(dialog): toggle focus-trap on DOM connect and disconnect

jcfranco:
**Related Issue:** #10731 

## Summary

Fixes an issue where a `focus-trap` instance might be left active after the dialog is removed.

### Relevant changes

* Adds `useFocusTrap` controller
* Extracts and exports function to create default `focus-trap` options


jcfranco:
I'll update `modal` and `sheet` separately as they'd be affected by this as well.

packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -80,7 +70,21 @@ export class Dialog
 
   private dragPosition: DialogDragPosition = { ...initialDragPosition };
 
-  focusTrap: FocusTrap;
+  focusTrap = useFocusTrap<Dialog>({
```
driskull, line 73: so react like in the naming 😊

driskull (approved):
👍 This is nice 💯 I like the triggerProp

Issue #10175, opened 2024-08-28
[Textarea] Add support for preventing further input if it exceeds the maxlimit


mpriour:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Add a property to the `calcite-textarea` component that when `true` will prevent a user from adding more text to the component when the `maxlength` value is exceeded.
This is required to have true compatibility with the deprecated `calcite-input` with `type="textarea"` component and to mimic behavior of the native `textarea` element.

### Acceptance Criteria

With the attribute present / set to `true`, further input into the textarea which would add text beyond the `maxlength` is prevented. Additionally if one pasted text into the textarea when this attribute is present, the pasted input would be truncated to the `maxlength`. If no `maxlength` attribute is set, then this new attribute would not do anything. These 2 input behaviors would mimic the native textarea behavior and the input behavior of the `calcite-input` with `type="textarea"`

### Relevant Info

* `calcite-input` with `type="textarea"` is planned for removal as a breaking change in v 4.0 according to #9974
* current input behavior beyond a configured max length is ambiguous and can be confusing to users according to feedback where we've tried to use the `calcite-textarea` as per recommendations

### Which Component

textarea

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Online

mpriour:
Happy to impliment this and make a PR if some decision is made as to what the new attribute would be called. `limitInput` is my suggestion.

jasdelreyes:
Will this get in for R1? Here's another bug that triggers this: https://devtopia.esri.com/WebGIS/arcgis-portal-app/issues/55157#issuecomment-5202572

jcfranco:
@mpriour @jasdelreyes Sorry for the delay! 😅

@mpriour Thanks for volunteering! I’ve submitted a PR for this, targeting this milestone, and would appreciate your review.

mpriour:
@jcfranco thanks. Reviewed and it looks good.

geospatialem:
Verified the new `limitText` prop with `maxLength`:

```html
<calcite-text-area placeholder="Provide your feedback" limit-text max-length="250">
[...]

PR #11440, opened 2025-02-03
feat(text-area): add `limitText` prop to prevent input beyond max-length

jcfranco:
**Related Issue:** #10175 

## Summary

Adds `limitText` prop to restrict input beyond the specified `max-length`, aligning with native behavior.

### DIFF
--- a/packages/calcite-components/src/components/text-area/text-area.tsx
+++ b/packages/calcite-components/src/components/text-area/text-area.tsx
@@ -162,6 +162,11 @@ export class TextArea
   @property() label: string;
 
   /**
+   * When `true`, prevents input beyond the maximum length, mimicking native `<textarea>` behavior.
+   */
+  @property() limitText = false;
+
+  /**
    * When the component resides in a form,
    * specifies the maximum number of characters allowed.
    *
@@ -462,6 +467,7 @@ export class TextArea
           }}
           cols={this.columns}
           disabled={this.disabled}
+          maxLength={this.limitText ? this.maxLength : -1}
           name={this.name}
           onChange={this.handleChange}
           onInput={this.handleInput}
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/text-area/text-area.tsx
```diff
@@ -161,6 +161,11 @@ export class TextArea
   /** Accessible name for the component. */
   @property() label: string;
 
+  /**
+   * When `true`, prevents input beyond the maximum length, mimicking native `<textarea>` behavior.
+   */
+  @property() limitText = false;
```
driskull, line 167: should probably be reflected and a reflects test added

driskull (approved):
👍

packages/calcite-components/src/components/text-area/text-area.tsx
jcfranco: Good catch. Done!

packages/calcite-components/src/components/text-area/text-area.tsx
```diff
@@ -462,6 +467,7 @@ export class TextArea
           }}
           cols={this.columns}
           disabled={this.disabled}
+          maxLength={this.limitText ? this.maxLength : -1}
```
eriklharper, line 470: What does `-1` do here in this case?  Does that set `maxLength` to `undefined` as part of the Lumina compiler?

jcfranco: It setting it to [`maxLength`'s initial value](https://developer.mozilla.org/en-US/docs/Web/API/HTMLTextAreaElement/maxLength#value).

eriklharper: There we go, thank you for that link!  I couldn't seem to find docs on that when I was browsing that page just now.

Issue #11345, opened 2025-01-21
[dialog] add options to allow adjustments to focus-trap behavior


rpanichakit9541:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

When using `calcite-dialgo` with `modal` prop, it currently prevents the user from clicking on anything that's not contained within the dialog.  In the current case we're running into an issue with this is we're using CKEditor inside the calcite-dialog which has certain options that render its own version of a pop over to allow adding link URL.  Because this input is being rendered directly on the document body instead of inside the DOM within the dialog content, the user is not able to click to input the URL text or click the cancel or save button within this 3rd party pop over.  (Screenshot below)

Blocked issues: #11416, #11420

![Image](https://github.com/user-attachments/assets/ec8fe102-15c7-48b7-8d70-bd94c75b99d8)

### Acceptance Criteria

I would like to see if there's a prop that can be exposed to allow customizing the calcite-dialog to still maintain focus trap while the user uses keyboard navigation, but allow the user to use the mouse to click on UI elements like this 3rd party pop over even though the DOM elements are not rendered within the calcite-dialog.

### Which Component

calcite-dialog with `modal` prop

### Example Use Case

https://devtopia.esri.com/WebGIS/arcgis-metadata-editor/issues/1537

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Online

geospatialem:
Components to consider as part of the focus trap behavior tweaks:
- popover
- dialog
- sheet
- modal

jcfranco:
`calcite-dialog`, `calcite-modal`, `calcite-popover` and `calcite-sheet` now have a `focusTrapOptions` property to configure the focus trap for certain scenarios (see [type](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/controllers/useFocusTrap.ts#L63-L72) for additional information):

**Allowing focus within containers external to the focus trap (e.g., [`CKEditor`](https://ckeditor.com/)):**

* If the external container is present before the focus trap is activated, use `extraContainers`.
[...]

Issue #10682, opened 2024-11-04
Popover - When clicked outside of a focus trap, focus is not returned to the element prior to the focus trap


annelfitz:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates and reviewed the [Accessibility page](https://developers.arcgis.com/calcite-design-system/foundations/accessibility) for guidance.

### Summary

When the `esc` key is pressed when the calcite popover is open, it will always close the popover, even if the focus is elsewhere in the web page.

When clicked outside of the focus trap (like on the map for example), we would expect the `esc` key to no longer affect the popover, at least until the focus is returned to that element.

Ideal behavior: 
- When escape is pressed, focus is returned to the element prior to the focus trap
- When clicked outside of a focus trap, focus is not returned to the element prior to the focus trap

Currently, in both of the scenarios above, focus is returned to the element prior to the focus trap. It seems like we may want different things depending on if the focus trap is deactivated due to a click vs escape key press.

### Actual Behavior

The `esc` key will always close the popover, even if the focus is no longer on the popover.
The focus will always return to the element prior to the focus trap, even if it is deactivated due to a click elsewhere.

### Expected Behavior

When clicked outside of a focus trap, the focus should not be returned to the element prior to the focus trap.

### Reproduction Sample

https://codepen.io/driskull/pen/zYgaNQP

### Reproduction Steps

1. open the popover
2. press escape
3. notice the button is focused
4. open the popover
5. click outside of the popover
6. notice the button is focused
7. click again outside of the popover and the focus on the button is now gone.

### Reproduction Version

2.13.2

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

driskull:
The way our focus trapping works is that when the focus trap is deactivated via click outside or escape, focus is returned to the element focused prior to opening the focus-trap. In this case, this would be the expand button.

I think what Maps SDK is asking for is that clicks outside deactivate the focus trap but do not return focus to the element prior to focus trapping. But when hitting escape, it does return focus to the element previously focused.

I think we may need a spike and do some research here.

hcampos-professional:
[...]

Issue #11305, opened 2025-01-15
Leverage Lit controllers


jcfranco:
### Description

[Lit controllers](https://lit.dev/docs/composition/controllers/) are a powerful and useful feature that we can start leveraging to streamline code and promote reusability. The following are potential candidates that could be refactored to follow this pattern:

- [ ] [`loadable`](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/utils/loadable.ts) -> can probably leverage `componentOnReady` since it's consistent across all outputs  
- [ ] [`focusTrapComponent`](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/utils/focusTrapComponent.ts)  
- [ ] [`floating-ui`](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/utils/floating-ui.ts)  
- [ ] [`form`](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/utils/form.tsx)  
- [ ] [`interactive`](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/utils/interactive.tsx)  
- [ ] [`label`](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/utils/label.ts)  
- [ ] [`openCloseComponent`](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/utils/openCloseComponent.ts)  
- [ ] [`selectableComponent`](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/utils/selectableComponent.ts)  
- [ ] [`sortableComponent`](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/utils/sortableComponent.ts)  

### Proposed Advantages

Leaner code, simpler code reuse pattern.

### Which Component

Modules listed above.

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

PR #11453, opened 2025-02-05
feat(dialog, modal, popover, sheet): add options prop to customize focus-trap

jcfranco:
**Related Issue:** #11345 

## Summary

Adds `focusTrapOptions` option to allow additional customization of focus-trapping components. It supports the following options from https://github.com/focus-trap/focus-trap#createoptions:

* `initialFocus` – would replace popover's internal initialFocusTrapFocus prop
* `allowOutsideClick` – supports https://github.com/Esri/calcite-design-system/issues/10682 
* `returnFocusOnDeactivate` – supports https://github.com/Esri/calcite-design-system/issues/10682

And the following custom prop:

* `extraContainers` – allows specifying extra elements (nodes or selectors) to focus trap (e.g., anything appending to the body, etc) when creating/activating the trap. **Note**: if specified, elements must exist when the focus trap is activated, if extra containers are created afterwards, users can use `updateFocusTrapElements(extraContainers)`

This also enhances `updatesFocusTrapElements()` to accept extra containers to allow in the focus trap if these are created after the trap is activated.

### Notes

* A subset of https://github.com/focus-trap/focus-trap#createoptions options are exposed as certain configurations might break component functionality.
* `extraContainers` gets used both when creating and updating the focus trap target containers
* Tidies up types


packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -161,6 +161,9 @@ export class Dialog extends LitElement implements OpenCloseComponent, LoadableCo
    */
   @property({ reflect: true }) escapeDisabled = false;
 
+  /** Allows special focus-trap configuration. */
```
driskull, line 164: Maybe `Allows for focus trapping configuration`?

driskull: We'll probably need to document the options as well right? `"allowOutsideClick" | "initialFocus" | "returnFocusOnDeactivate"`

driskull (commented):
Looks good! We'll need to do this for Modal, Popover, and Sheet as well.

packages/calcite-components/src/components/dialog/dialog.tsx
jcfranco: Totally. @geospatialem @DitwanP Any ideas on how we should doc items from another lib? I don't think we call out using `focus-trap` or other 3rd-party lib anywhere in the doc site.

driskull: Yeah we don't want to call out focus-trap specifically so we can say "focus trapping"

geospatialem: Maybe something like:

```
Specifies custom focus trap configuration on the component, where `"allowOutsideClick`" allows outside clicks, `"initialFocus"` enables initial focus, and `"returnFocusOnDeactivate"` returns focus when not active.
```

I'm not 100% sure if that's the true intended behavior for the options, but perhaps something to that effect?

jcfranco:
I'll work on adding automated test coverage for some of these changes and controller-specific testing will be handled in https://github.com/Esri/calcite-design-system/issues/11305.

packages/calcite-components/src/components/dialog/dialog.tsx
```diff
@@ -272,10 +282,16 @@ export class Dialog extends LitElement implements OpenCloseComponent, LoadableCo
...
+  /**
+   * Updates the element(s) that are included in the focus-trap of the component.
+   *
+   * @param extraContainers - Additional elements to include in the focus trap. This is useful for including elements that may have related parts rendered outside the main focus trapping element.
+   */
   @method()
-  async updateFocusTrapElements(): Promise<void> {
-    this.focusTrap.updateContainerElements();
+  async updateFocusTrapElements(
+    extraContainers?: ExtendedFocusTrapOptions["extraContainers"],
```
driskull, line 292: nice, this looks good 👍

driskull (approved):
👍

packages/calcite-components/src/controllers/useFocusTrap.ts
```diff
@@ -53,6 +53,38 @@ interface FocusTrapComponent extends LitElement {
...
+   */
+  focusTrapOptions?: ExtendedFocusTrapOptions;
+}
+
+export type ExtendedFocusTrapOptions =
+  /**
+   * @see https://github.com/focus-trap/focus-trap#createoptions
+   */
+  Pick<FocusTrapOptions, "allowOutsideClick" | "initialFocus" | "returnFocusOnDeactivate"> & {
+    /**
```
Elijbet, line 68: How come this doesn't include `escapeDeactivates`?

packages/calcite-components/src/components/sheet/sheet.tsx
```diff
@@ -137,6 +142,16 @@ export class Sheet
...
   @property({ reflect: true }) focusTrapDisabled = false;
 
+  /**
+   * Specifies custom focus trap configuration on the component, where
+   *
+   * `"allowOutsideClick`" allows outside clicks,
+   * `"initialFocus"` enables initial focus,
+   * `"returnFocusOnDeactivate"` returns focus when not active, and
+   * `"extraContainers"` specifies additional focusable elements external to the trap (e.g., 3rd-party components appending elements to the document body).
+   */
```
Elijbet, line 152: Shouldn't this include `escapeDeactivates`?

Elijbet (approved):

packages/calcite-components/src/controllers/useFocusTrap.ts
jcfranco: I didn't include some options that may conflict with existing component props (e.g., `escapeDisabled), but we could revisit this if there's a use case.

jcfranco:
Merging to proceed with testing and verification. LMK if anything needs following up.

Issue #11464, opened 2025-02-05
[Table] 3.x - Borders not displaying correctly in Firefox


macandcheese:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

In Firefox, Table borders are not always displaying correctly in 3.x:

(FF / Safari / Chrome - all macOS):

![Image](https://github.com/user-attachments/assets/6a32c6d2-286c-48ad-9ec2-8491610daede)

### Expected Behavior

Firefox should display the Table component as other browsers do with correctly displayed borders.

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/JoPQMjq?editors=100

### Reproduction Steps

1. Open Codepen in Chrome
2. Notice display is correct with 3.x CC import
3. Switch to 2.x CC import
4. Notice display is still correct
5. Open Codepen in Firefox
6. Notice display is incorrect with 3.x CC import
3. Switch to 2.x CC import
4. Notice display is correct

### Reproduction Version

3.x

### Relevant Info

This showed up with the Lumina refactor - I think it's due to this long standing bug: https://bugzilla.mozilla.org/show_bug.cgi?id=688556

### Regression?

3.x / Lumina migration

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

PR #11466, opened 2025-02-05
fix(table): Fix `bordered` styling in Firefox

macandcheese:
**Related Issue:** #11464

## Summary
Adds a set of styling rules to address a Firefox-specific styling bug with a "not best practice" disclaimer.
Also updates the Table "local demo" to use Dialog instead of Modal.

Note: in FF - the Table ends up being slightly taller because of this adjustment - however, prior to this change, there was a greater difference in height ...

| Before    | After |
| -------- | ------- |
| Screenshot 2025-02-05 at 3 11 36 PM  | Screenshot 2025-02-05 at 3 10 24 PM  |

### DIFF
--- a/packages/calcite-components/src/components/table-row/table-row.scss
+++ b/packages/calcite-components/src/components/table-row/table-row.scss
@@ -26,8 +26,12 @@ tr {
   background-color: var(--calcite-internal-table-row-background);
 }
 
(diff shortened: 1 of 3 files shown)


jcfranco (approved):
✨🏓🔧✨

driskull (approved):
👍

PR #11476, opened 2025-02-06
refactor(text-area): clear maxLength instead of directly setting it to the default

jcfranco:
**Related Issue:** #10175

## Summary

Ensures `maxLength` is properly cleared in the internal `<textarea>` when `inputText` is set to `false`. Otherwise, it would be set to `-1`, which could interfere with external test utils.


aPreciado88 (approved):
🚀

eriklharper (approved):

PR #11480, opened 2025-02-07
fix(dropdown): avoid showing scroller when max-items equals items

jcfranco:
**Related Issue:** #10731 

## Summary

Fixes regression caused by https://github.com/Esri/calcite-design-system/pull/10971.

### DIFF
--- a/packages/calcite-components/src/components/dropdown/dropdown.tsx
+++ b/packages/calcite-components/src/components/dropdown/dropdown.tsx
@@ -647,7 +647,7 @@ export class Dropdown
           <div
             aria-labelledby={`${guid}-menubutton`}
             class={{
-              ["calcite-dropdown-content"]: true,
+              [CSS.content]: true,
               [FloatingCSS.animation]: true,
               [FloatingCSS.animationActive]: open,
             }}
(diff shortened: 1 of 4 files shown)


packages/calcite-components/src/components/dropdown/dropdown.tsx
```diff
@@ -562,30 +561,14 @@ export class Dropdown
...
-    });
 
-    return items.length >= maxItems ? maxScrollerHeight : 0;
+    return items.length >= maxItems ? this.getYDistance(this.scrollerEl, items[maxItems - 1]) : 0;
+  }
+
+  private getYDistance(parent: HTMLElement, child: HTMLElement): number {
+    const parentRect = parent.getBoundingClientRect();
+    const childRect = child.getBoundingClientRect();
+    return childRect.bottom - parentRect.top;
```
benelan, line 571: I'm new to the scroller logic and this function took me a second to wrap my head around. A couple questions:

- Is the scroller the parent element of the dropdown items? That's what it sounds like from the variable names.
- Would the top of the first item be equivalent to the top of the scroller? If so, using the first item instead would probably make it easier to read.

benelan (approved):
:+1:

PR #11496, opened 2025-02-08
fix: improve ref handling logic

jcfranco:

PR #11497, opened 2025-02-08
refactor(alert, autocomplete, combobox, dialog, dropdown, input-date-picker, modal, notice, popover, sheet, tooltip): avoid clearing `transitionEl` ref nodes

jcfranco:
**Related Issue:** #11093

## Summary

Updates transition element `ref` callbacks to avoid clearing the reference. This should preserve the context of existing ref logic.

### DIFF
--- a/packages/calcite-components/src/components/alert/alert.tsx
+++ b/packages/calcite-components/src/components/alert/alert.tsx
@@ -339,6 +339,10 @@ export class Alert extends LitElement implements OpenCloseComponent, LoadableCom
   }
 
   private setTransitionEl(el: HTMLDivElement): void {
+    if (!el) {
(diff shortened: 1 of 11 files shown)


jcfranco:
This only targets transition element refs to keep changes scoped to a recently reported issue in #10731. https://github.com/Esri/calcite-design-system/pull/11496/ will address all refs.

benelan (approved):
Would it be possible to add tests for this? I can be a follow up PR if this needs to get into 3x

jcfranco:
> Would it be possible to add tests for this? I can be a follow up PR if this needs to get into 3x

I'll create an issue to explore options for a common test helper.

PR #11498, opened 2025-02-08
fix(modal): fix issue caused by early opening logic calls

jcfranco:
**Related Issue:** #10731 

## Summary

Fixes an issue where the modal’s opening logic was invoked before the transition element was available. This adds a simple guard for now, but I’ll revisit later to ensure the method isn’t called until after the component updates.

BEGIN_COMMIT_OVERRIDE
omitted from changelog
END_COMMIT_OVERRIDE


benelan (approved):
:broom:

PR #11510, opened 2025-02-11
chore(input-date-picker): drop an unnecessary comment

Elijbet:
**Related Issue:** #N/A

## Summary
Drop an unnecessary comment.

### DIFF
--- a/packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
+++ b/packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
@@ -845,7 +845,7 @@ export class InputDatePicker
       focusTrapEl: el,
       focusTrapOptions: {
         allowOutsideClick: true,
-        // Allow outside click and let the popover manager take care of closing the popover.
+        // Allow outside click, as clicking outside closes the date picker which deactivates focusTrap.
         clickOutsideDeactivates: false,
         initialFocus: false,
         setReturnFocus: false,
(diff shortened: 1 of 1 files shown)


packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -845,7 +845,7 @@ export class InputDatePicker
       focusTrapEl: el,
       focusTrapOptions: {
         allowOutsideClick: true,
-        // Allow outside click and let the popover manager take care of closing the popover.
+        // Allow outside click, as clicking outside closes the date picker which deactivates focusTrap.
```
jcfranco, line 848: I'm wondering if we need this comment at all. [`popover`'s options](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/components/popover/popover.tsx#L83) don't have comments.

jcfranco:
This should use the `chore` commit type since it's updating an internal comment.

packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
Elijbet: I wasn't sure. Let's drop the comment then and I'll mark this as chore since it's just dropping an internal comment.

benelan (approved):
:+1:

Issue #11513, opened 2025-02-11
List drag menu 'Move to' actions broken when list elements have explicit `id`s


timmorey:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When three or more lists are configured to allow dragging between each other, and the `<calcite-list>` elements are given explicit `id` attributes, the drag menu actions work inconsistently.  They will succeed in moving items to another list, but not always the list you want to move it to.

We initially encountered this in a nested list scenarios, but nesting is not required.  I modified one of the documentation samples to come up with a repro case (see below).  In it, each `<calcite-list>` element has an explicit `id` attribute, and they are unique within the document.  If I remove these explicit `id`s, the drag menus work correctly.

### Expected Behavior

Developers should be able to add `id` attributes to elements at their discretion, without affecting internal calcite functionality.

### Reproduction Sample

https://codepen.io/_timmorey_/pen/myddmWO?editors=1000

### Reproduction Steps

1. click the drag menu for the first item in the first list (Narnia), and in the drag menu, choose "Other places"
    - Notice the item moves to the second list, not the third list you requested
2. click the drag menu for the last item in the last list (Townville), and in the drag menu, choose "Visited places"
    - Notice the item again moves tot the second list, not the first list you requested

### Reproduction Version

3.0.0

### Priority impact

impact - p2 - want for an upcoming milestone

### Impact

We have worked around this problem by removing our explicit `id` attributes and finding other ways to identify the elements.

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Field Apps

DitwanP:
🍡 Verified on `3.0.0-next.3`
https://codepen.io/Ditwan-Price/pen/ogNgPBd?editors=1000

PR #11515, opened 2025-02-11
fix(list, block-group): drag menu 'Move to' actions broken when list elements have explicit ids

driskull:
**Related Issue:** #11513

## Summary

- Fix keyboard moving `MoveTo` object to always have unique identifiers
- Previously, it was using the incorrect id
- There's no reason to use the element id so just using a guid makes more sense here
- Add tests


benelan (approved):
:+1:

PR #11522, opened 2025-02-11
feat(calcite-stepper): allow calcite-stepper to receive slotted items from wrapper component

kstinson14:
**Related Issue:** #

## Summary

In Field Maps, we use the `calcite-stepper` in several places, many times with accompanying forward/backward buttons.  So we've bundled them together in a wrapper web-component with a slot allowing each usage to pass in its desired `calcite-stepper-items`.  So it looks something like this:
```
<wrapper-component>
  <calcite-stepper numbered layout="horizontal">
    <slot/>
  </calcite-stepper>
  ...[code for buttons, etc.]
</wrapper-component>
```
However, this usage passes the `calcite-stepper-items` within the `calcite-stepper`'s slot, but does not pass its props to them (`numbered`, `layout`, etc.) causing them to render incorrectly.  I believe this is because the function that performs this task, `updateItems` finds the items through `this.el.querySelector`.  This doesn't work in the wrapping component scenario due to them being wrapped in a `slot`.  I have a branch [here](https://github.com/kstinson14/calcite-design-system/tree/feature/allow-stepper-wrapping-component-repro) that repros the problem on the local demo site with a dummy `stepper-wrapper` component.

I've fixed this by keeping track of the `slot` element after the slot change event and using its assigned `calcite-stepper-items` in `updateItems` instead.  I also call `updateItems` in the slot change event.  This seems to work for wrapped / unwrapped steppers.

### DIFF
--- a/packages/calcite-components/src/components/stepper/stepper.tsx
+++ b/packages/calcite-components/src/components/stepper/stepper.tsx
@@ -1,7 +1,7 @@
 // @ts-strict-ignore
 import { PropertyValues } from "lit";
 import { LitElement, property, createEvent, h, method, state, JsxNode } from "@arcgis/lumina";
-import { focusElementInGroup, slotChangeGetAssignedElements } from "../../utils/dom";
+import { focusElementInGroup, getSlotAssignedElements, slotChangeGetAssignedElements } from "../../utils/dom";
 import { Position, Scale } from "../interfaces";
 import { createObserver } from "../../utils/observers";
 import { guid } from "../../utils/guid";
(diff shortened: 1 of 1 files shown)


jcfranco (changes requested):
@kstinson14 Thanks for submitting this! Before proceeding, would you mind creating an issue for us? We need to evaluate it first. Afterwards, we’ll determine whether it should apply across components and not just the stepper. Having an issue will ensure it goes through our verification process if we proceed.

kstinson14:
@jcfranco Sorry - missed this comment.  Done! https://github.com/Esri/calcite-design-system/issues/11737

jcfranco:
Closing in favor of https://github.com/Esri/calcite-design-system/pull/14021. Thank you for submitting this and apologies for the delay.

PR #11530, opened 2025-02-12
docs(readme): remove redundant sections, fix broken link

benelan:
## Summary

- Refer to the documentation site's get started page for all of the NPM package usage instructions.
- Fix a broken link to the contributing documentation, which was moved to the root directory of the monorepo.
- Change `Calcite Components` to `Calcite components`
- Cleanup the react wrapper package's readme


packages/calcite-components-react/README.md
```diff
@@ -1,22 +1,22 @@
 # Calcite Components React
 
-A set of React components that wrap [Calcite Components](https://developers.arcgis.com/calcite-design-system/components/). An application using this package is provided in the [`calcite-components-examples`](https://github.com/Esri/calcite-components-examples) repo.
+This package provides React wrappers for [Calcite components](https://developers.arcgis.com/calcite-design-system/components/). See the [react example](https://github.com/Esri/calcite-design-system/tree/dev/examples/components/react) for a minimal application using this package.
```
geospatialem, line 3: nitpick to apply to more audiences:

```suggestion
This package provides React wrappers for [Calcite components](https://developers.arcgis.com/calcite-design-system/components/). Refer to the [React example](https://github.com/Esri/calcite-design-system/tree/dev/examples/components/react) for a minimal application using this package.
```

packages/calcite-components-react/README.md
```diff
@@ -1,22 +1,22 @@
...
-There are two builds that are provided by the standard `calcite-components` package.
+There are two builds provided by the standard components package.
 
 ### Custom Elements build
 
-[Custom Elements](developers.arcgis.com/calcite-design-system/get-started#custom-elements) is the recommended build when using frontend frameworks, such as React. To use this build, you will need to set the path to the `calcite-components` assets. You can either use local assets, which will be explained in a subsequent step, or assets hosted on a CDN.
+[Custom Elements](developers.arcgis.com/calcite-design-system/get-started#custom-elements) is the recommended build when using frontend frameworks, such as React. To use this build, you will need to set the path to the components assets. You can either use local assets, which will be explained in a subsequent step, or assets hosted on a CDN.
 
 ```jsx
 import { setAssetPath } from "@esri/calcite-components/dist/components";
```
geospatialem, line 22: Would the commented out `setAssetPath` need to removed, or edited to reflect more of a local/different cdn, maybe similar to our [doc site example](https://developers.arcgis.com/calcite-design-system/get-started/#custom-elements-build):

```
setAssetPath("/path-to-your-assets/");
```

geospatialem (approved):
Few considerations for your review, but otherwise LGTM! 🚀

packages/calcite-components-react/README.md
benelan: I think it's fine as it is because it's where React apps commonly serve their assets, e.g. our React example:

https://github.com/Esri/calcite-design-system/blob/4e951e744b24f9ce4cde32740b20836b72a6c448/examples/components/react/src/main.tsx#L8

It's generalized in the doc because it's not the same for all frameworks or build tools, e.g. our rollup example:

https://github.com/Esri/calcite-design-system/blob/4e951e744b24f9ce4cde32740b20836b72a6c448/examples/components/rollup/src/main.js#L7

benelan: I guess we could remove the CDN comment though because I think Lumina does that be default now

benelan: @maxpatiiuk is that correct? And is there a new default behavior for `defineCustomElements`?

DitwanP (approved):
✨

packages/calcite-components-react/README.md
maxpatiiuk: > I guess we could remove the commented snippet for CDN assets though because I think Lumina uses the CDN by default when setAssetPath isn't called.

Correct. in map-components 4.32 we only document setAssetPath for people who don't want to use our CDN/need offline deployment

maxpatiiuk: Yes, defineCustomElements has the same behaviour
Also, if resourceUrl is provided to defineCustomElements, it simply calls setAssetPath for you - equivalent to calling setAssetPath directly.

maxpatiiuk: The path can be simplified:

```suggestion
import { setAssetPath } from "@esri/calcite-components";
```

packages/calcite-components-react/README.md
```diff
@@ -38,20 +35,15 @@ import { CalciteButton, CalciteIcon, CalciteSlider } from "@esri/calcite-compone
...
 ```jsx
 import { defineCustomElements } from "@esri/calcite-components/dist/loader";
-// Local assets
-defineCustomElements(window);
 
-// CDN hosted assets
-// defineCustomElements(window, {
-//   resourcesUrl: "https://unpkg.com/@esri/calcite-components/dist/calcite/assets"
-// });
+defineCustomElements(window);
```
maxpatiiuk, line 43: don't need to pass `window` - it is ignored if passed

packages/calcite-components-react/README.md
maxpatiiuk, line 1: With Lumina, you are now also providing JSX typings for React 19 users.
Given that more and more people will be shifting to React 19, it would be ideal to point out on this page or somewhere else that wrappers are not required in React 19 (but still supported).

documentation on usage in React in https://qawebgis.esri.com/components/lumina/publishing#using-the-npm-package (generally same as other frameworks, except for need to add a specific /// <reference> directive to get the JSX typings)

PR #11531, opened 2025-02-12
docs: link to package readmes from root readme

benelan:
## Summary

Link to the READMEs of the packages in the monorepo's root README.


geospatialem (approved):
✨'tis of beauty, thanks @benelan! ✨

Issue #11518, opened 2025-02-11
`--calcite-loader-progress-color-inline` has wrong description


COV-GIS:
### Description

https://github.com/Esri/calcite-design-system/blob/5c6e12654854bd3185c9f798c3ae0ac878b08b51/packages/calcite-components/src/components/loader/loader.scss#L10

Probably should be something like: Specifies the component's progress ring color when set to inline. 

### Which Component

loader

geospatialem:
Verified on the `dev` branch, anticipated for the `3.0.1` patch 📚

PR #11535, opened 2025-02-12
docs(loader): update references to reflect component use

geospatialem:
**Related Issue:** #11518

## Summary
Updates `loader`'s properties and css variables descriptions to reflect the updated use.


geospatialem:
Merging the above in, it appears the tests may be unstable tests.

Issue #11544, opened 2025-02-13
[Input Date Picker]  Back and forward buttons `<`,`>`  do not work on Safari


fabiof90:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates
Same/similar issue reported here (https://github.com/Esri/calcite-design-system/issues/11432) but not as bug and with low impact; therefore created this ticket as it is a bug with higher impact. 

### Actual Behavior

For Input Date Picker, the back and forward buttons `<`,`>`do close the Input Date Picker and do not show any effect on Safari. 

https://github.com/user-attachments/assets/1a580216-5692-403c-b8d0-73f815209b85

### Expected Behavior

For Input Date Picker, the back and forward buttons `<`,`>` let you move through the months. 

https://github.com/user-attachments/assets/a255340d-f375-493e-a61f-4ac5f396d37b

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/input-date-picker/, Daylight Widget, Layer Visibility Settings, Time Slider Settings

### Reproduction Steps

1. Open Input Date Picker in Safari
2. Try to move through months using the `<`,`>` buttons

### Reproduction Version

3.0.0

### Relevant Info

Reported only for Safari on Desktop as well as iOS. 

### Regression?

n/a 

### Priority impact

impact - p1 - need for current milestone

### Impact

Input Date Picker is used in many widgets and implementations throughout the ArcGIS Online Products (eg. Daylight Widget; Time Slider; Layer Visibility; ...). Having those buttons not work for Safari does effect many customers in negative way. There is the workaround to choose the Month by dropdown, but this is rather cumbersome for most cases. 
Should be implemented for current R1 release imo. 

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Scene Viewer

DitwanP:
🍡 Verified in `Safari` on `3.0.0-next.5`

https://codepen.io/Ditwan-Price/pen/WbNvjex?editors=100

driskull:
I'm still seeing an issue with this one. When clicking on the background of the date picker, its closing in Safari but not Chrome. I think we should open this one back up.

anveshmekala:
[...]

PR #11547, opened 2025-02-13
fix(input-date-picker): allow navigating between months using chevron actions in Safari

anveshmekala:
**Related Issue:** #11544 

## Summary

No longer closes the `input-date-picker` when navigating between month using chevron actions.


packages/calcite-components/src/components/action/action.tsx
```diff
@@ -295,6 +295,7 @@ export class Action extends LitElement implements InteractiveComponent {
         disabled={disabled}
         id={buttonId}
         ref={this.buttonEl}
```
jcfranco, line 297: Can you add an explainer comment here?

jcfranco (approved):

Issue #11523, opened 2025-02-11
[Dialog] Extra containers added via `updateFocusTrapElements` not honored after DOM update


ellenupp:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When adding an external container to a dialog's focus trap via `updateFocusTrapElements`, the container may be removed again if a DOM change in the dialog triggers its internal mutation observer that resets the focus trap containers. 

### Expected Behavior

After adding a container to the dialog's focus trap, future updates to the dialog's focus trap containers should honour previously added extra containers if they still exist. 

### Reproduction Sample

https://codepen.io/ellupp/pen/vEYYrKL?editors=1100

### Reproduction Steps

1. In the open dialog click button "1" which will create an external container and add it to the focus trap. Notice you can tab to the new button.
2. Click button "2" in the dialog which will add some content to the dialog to trigger the mutation observer. Notice you can no longer tab to the external button. 

### Reproduction Version

3.0.0

### Priority impact

impact - p1 - need for current milestone

### Impact

Needed for using a conditionally rendered CkEditor in a dialog

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Dashboards

ellenupp:
@jcfranco Unfortunately I'm still seeing the issue and can repro with the same codepen in 3.0.1.

jcfranco:
@ellenupp Found the issue. Fixing!

PR #11548, opened 2025-02-14
fix(dialog, modal, sheet): preserve focus-trap extra containers across internal trap updates

jcfranco:
**Related Issue:** #11523 

## Summary

Internal calls to `updateFocusTrap` caused user-provided `extraContainers` to be lost. This updates `useFocusTrap` to maintain stateful focusable containers for subsequent updates.


benelan (approved):
Code LGTM but would it be feasible to add spec coverage? If so, it can be in a follow up PR too.

jcfranco:
Yeah, I'm planning to cover all controller testing in https://github.com/Esri/calcite-design-system/issues/11305.

PR #11556, opened 2025-02-14
fix(dialog, modal, popover, sheet): fixes an issue that caused extra containers to be lost after internal trap updates

jcfranco:
**Related Issue:** #11523 

## Summary

https://github.com/Esri/calcite-design-system/pull/11548 missed calling the internal update focus trap method, using the public method instead, which clears user-specified `extraContainers`.

### DIFF
--- a/packages/calcite-components/src/components/dialog/dialog.tsx
+++ b/packages/calcite-components/src/components/dialog/dialog.tsx
@@ -767,7 +767,7 @@ export class Dialog extends LitElement implements OpenCloseComponent {
   }
 
   private handleMutationObserver(): void {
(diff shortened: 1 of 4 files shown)


driskull (approved):
👍 Any need to add a test for this?

jcfranco:
Yes. I'm aiming to start adding controller tests soon with https://github.com/Esri/calcite-design-system/issues/11305. We might also need a `focusTrap` test helper, right?

jcfranco:
> We might also need a focusTrap test helper, right?

Created https://github.com/Esri/calcite-design-system/issues/11558 for this.

Issue #11447, opened 2025-02-04
[BlockGroup, List]: canPull, canPut should be called for moving via keyboard as well


driskull:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

canPull and canPut are not called when moving via keyboard from one list to another.

### Expected Behavior

[BlockGroup, List]: canPull, canPut should be called for moving via keyboard as well

### Reproduction Sample

https://codepen.io/Matt-Driscoll/pen/wBwbZgp?editors=1010

### Reproduction Steps

1. open the codepen
2. try to drag a item to another group via mouse
3. notice it doesn't work
4. try to move an item to another group via keyboard
5. notice it works

### Reproduction Version

3.0.0-next.127

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

DitwanP:
🍡 Verified on `3.2.0-next.7`

PR #11567, opened 2025-02-15
feat(block-group, list): Add event for when a move is halted due to `canPut` or `canPull` returning false

driskull:
**Related Issue:** #11447

## Summary

- Add event for when a user uses the sort dropdown and `canPull` or `canPut` returns false
  - Call `canPull` and `canPut` when moving items via the "Move to" menu provided under the drag handle
  - If an item cannot be pulled or put, an event is emitted.
  - Users should handle these events to show a notification of their choice
- Adds tests


driskull:
@jcfranco @macandcheese this is still awaiting review.

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -937,20 +954,44 @@ export class List extends LitElement implements InteractiveComponent, SortableCo
     const toEl = moveTo.element as List["el"];
     const fromElItems = Array.from(fromEl.children).filter(isListItem);
     const oldIndex = fromElItems.indexOf(dragEl);
+    const newIndex = 0;
 
     if (!fromEl) {
       return;
     }
 
+    if (
```
jcfranco, line 963: Since this is shared, let’s create a utility for both `list` and `block-list` to use, and do the same for their tests.

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -937,20 +954,44 @@ export class List extends LitElement implements InteractiveComponent, SortableCo
     const toEl = moveTo.element as List["el"];
     const fromElItems = Array.from(fromEl.children).filter(isListItem);
     const oldIndex = fromElItems.indexOf(dragEl);
+    const newIndex = 0;
```
jcfranco, line 957: If a put/pull fails, does the index get modified to 0?

packages/calcite-components/src/components/block-group/block-group.tsx
```diff
@@ -287,18 +304,43 @@ export class BlockGroup extends LitElement implements InteractiveComponent, Sort
...
+    if (
+      toEl.canPut?.({
+        toEl,
+        fromEl,
+        dragEl,
+        newIndex,
+        oldIndex,
+      }) === false
+    ) {
+      toEl.putFailed({ toEl, fromEl, dragEl, oldIndex, newIndex });
```
jcfranco, line 335: It looks like `putFailed` can be inlined since it seems to only be used here. Applies to `list`.

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -937,20 +954,44 @@ export class List extends LitElement implements InteractiveComponent, SortableCo
...
+    }
+
     dragEl.sortHandleOpen = false;
 
     this.disconnectObserver();
 
     toEl.prepend(dragEl);
     expandedAncestors(dragEl);
-    const toElItems = Array.from(toEl.children).filter(isListItem);
-    const newIndex = toElItems.indexOf(dragEl);
```
jcfranco, line 952: Where is `newIndex` updated now? It looks like it's always 0 in the latest changes, where previously, it depended on the placement.

driskull: It was always 0, we always put at the top of the list when moving lists via keyboard which is why this code wasn't necessary.

packages/calcite-components/src/components/list/list.tsx
driskull: When moving via keyboard, we always move to index 0. So that is what the fail event would show.

packages/calcite-components/src/components/list/list.tsx
driskull: Can we do this as part of a new refactor issue?

packages/calcite-components/src/components/list/list.tsx
jcfranco: Ah, gotcha. So this is keyboard-specific moving logic.

packages/calcite-components/src/components/list/list.tsx
jcfranco: Sounds good, but let’s make it a priority to unify these paths as soon as possible.

packages/calcite-components/src/components/list/list.tsx
jcfranco: > So this is keyboard-specific moving logic.

Scratch that. This is move-event-specific logic.

jcfranco (approved):
✨↕️🤏✨

packages/calcite-components/src/components/list/list.tsx
driskull: Both, its logic specific to moving via keyboard. When moving via keyboard we always place at the top.

packages/calcite-components/src/components/block-group/block-group.tsx
driskull: @jcfranco what do you mean by inlined? Can you elaborate pretty plz :)

packages/calcite-components/src/components/list/list.tsx
jcfranco: Is a move is handled the same way whether it's keyboard or mouse-initiated? I guess I'm trying to confirm whether `newIndex` is updated differently when it's not moved via keyboard.

driskull: > Is a move is handled the same way whether it's keyboard or mouse-initiated?

Not exactly, when using a mouse the sortable library lets you move to any index within another list whereas with the keyboard moving via the handle component, we only allow moving to the top of another list. Then a user would need to sort within that list to sort within the list. 

This logic is only for moving via the sort handle which places the element at the top of the list being moved to.

jcfranco: Sweet. Thanks for clarifying.

jcfranco:
PR-title/changelog-wise, would developers know when `canPull`/`canPut` returns false? Maybe rephrasing to emphasize  what these events will enable would be helpful.

driskull:
@jcfranco how about `feat(block-group, list): Add events for when a move is prevented with canPut or canPull #11567`

Also, are the event names acceptable? I wasn't sure if `Fail` was the right term or if we wanted something different. Other potential terms: reject, prevent, stop

Spawns from this option: https://github.com/SortableJS/Sortable?tab=readme-ov-file#group-option

packages/calcite-components/src/components/block-group/block-group.tsx
jcfranco: You can ignore this one. I thought `putFailed` wasn't accessed externally.

jcfranco:
> how about feat(block-group, list): Add events for when a move is prevented with canPut or canPull #11567

🚢 it!

> Also, are the event names acceptable?

Good question. Fail makes it sound like `canPut` or `canPull` threw an error.

This got me thinking, if a developer is handling failed moves, will they typically need to listen for both events? If so, maybe they can share the same event regardless of `canPut` or `canPull` (with this info as part of the event payload, if needed). Something like `calcite<Component>MoveHalt`?

driskull:
> Something like calcite<Component>MoveHalt?

Yeah we could do that. Should I proceed?

jcfranco:
SGTM!

packages/calcite-components/src/components/list/list.tsx
driskull: created https://github.com/Esri/calcite-design-system/issues/11849

Issue #11225, opened 2025-01-08
Transition relevant components that use `open` prop to `expanded` per design decision


Elijbet:
### Description

`block` and  `list-item` use `open` prop, but should use `expanded` per design decision. They need to transition to using `expanded` instead.
 
`list-item` has both `open` and `closed`. It looks like it should be `expanded` and `closed`.

```
  /** When `true`, the item is open to show child components. */
  @property({ reflect: true }) open = false;
  
  /** When `true`, hides the component. */
  @property({ reflect: true }) closed = false;
```

### Proposed Advantages

Alignment with the design decision and consistency across components. 

### Which Component

block
block-section 
list-item
list
and other relevant components

### Relevant Info

Design decision: https://github.com/Esri/calcite-design-system/issues/6473#:~:text=for%20more%20context.-,Link%20to%20Figma%20file,-Which%20Component
[...]

PR #11582, opened 2025-02-18
feat(block, block-section): add `expanded` property and deprecate `open` property

Elijbet:
**Related Issue:** #6473, #11225

## Summary
Deprecate `open` prop in favor of `expanded`. Add logger warning for deprecation to help inform users about the deprecated property and guide them towards using the recommended alternatives.

### DIFF
--- a/packages/calcite-components/src/components/block/block.stories.ts
+++ b/packages/calcite-components/src/components/block/block.stories.ts
@@ -13,7 +13,7 @@ interface BlockStoryArgs
       Block,
       | "heading"
       | "description"
-      | "open"
+      | "expanded"
       | "collapsible"
       | "loading"
       | "disabled"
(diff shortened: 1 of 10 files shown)


packages/calcite-components/src/components/block/block.stories.ts
```diff
@@ -122,23 +122,23 @@ export const darkModeRTL_TestOnly = (): string => html`
   <calcite-block
     heading="Heading"
     description="description"
-    open
+    expanded
     collapsible
```
jcfranco, line 126: After seeing the renamed prop in context, I think we should deprecated `collapsible` in favor of `expandable`. We can tackle this separately to allow for some initial discussion.

jcfranco (approved):
🚀 Review comments from https://github.com/Esri/calcite-design-system/pull/11003#pullrequestreview-2661923879 are also applicable here. Good to merge afterwards!

packages/calcite-components/src/components/block/block.stories.ts
Elijbet: [[Block] Deprecate collapsible in favor of expandable](https://github.com/Esri/calcite-design-system/issues/11671#top)#11671

PR #11585, opened 2025-02-18
docs(block-group): update documentation for canPull

driskull:
**Related Issue:** N/A

## Summary

Matches documentation from the list component

https://github.com/Esri/calcite-design-system/blob/3470a989073ba61dca514f1c62bb18413b17ce8f/packages/calcite-components/src/components/list/list.tsx#L194

### DIFF
--- a/packages/calcite-components/src/components/block-group/block-group.tsx
+++ b/packages/calcite-components/src/components/block-group/block-group.tsx
@@ -86,7 +86,7 @@ export class BlockGroup extends LitElement implements InteractiveComponent, Sort
 
   // #region Public Properties
 
-  /** When provided, the method will be called to determine whether the element is can move from the component. */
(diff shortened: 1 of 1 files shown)


packages/calcite-components/src/components/block-group/block-group.tsx
```diff
@@ -86,7 +86,7 @@ export class BlockGroup extends LitElement implements InteractiveComponent, Sort
 
   // #region Public Properties
 
-  /** When provided, the method will be called to determine whether the element is sortable in the component. */
+  /** When provided, the method will be called to determine whether the element is can move from the component. */
```
DitwanP, line 89:
```suggestion
  /** When provided, the method will be called to determine whether the element can move from the component. */
```

DitwanP (approved):
👍 One small fix and LGTM

Issue #11583, opened 2025-02-18
Calcite slider histogram is getting clipped


ubatsukh:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

I am creating a new sample that use calcite slider with histogram. It looks like the histogram data is getting clipped. How can I fix that? what am I doing wrong? 
Here is my codepen - https://codepen.io/U_B_U/pen/EaxxdbB?editors=1000 (isolated version [here](https://codepen.io/jcfranco/pen/KwKVNPp?editors=100))

Notice the flatlines in the following image 
![Image](https://github.com/user-attachments/assets/1bcc3f2c-c6d5-4da5-8a19-bd71cccac1fc)

@jcfranco confirmed that it is an issue with the internal graph component's masking. 

### Expected Behavior

The histogram data should not get clipped.

### Reproduction Sample

 https://codepen.io/U_B_U/pen/EaxxdbB?editors=1000 (isolated version [here](https://codepen.io/jcfranco/pen/KwKVNPp?editors=100))

### Reproduction Steps

1. Open this app
2. Notice the slider with a histogram
3. Data is being clipped.

### Reproduction Version

3.0.2

### Priority impact

impact - p1 - need for current milestone

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

jcfranco:
> @jcfranco confirmed that it is an issue with the internal graph component's masking.

Upon closer inspection, it looks like the issue is related to the way the vertical histogram range is computed. The fix is still looks feasible for the upcoming patch.

geospatialem:
Verified for the upcoming patch with `3.1.0-next.10`

![Image](https://github.com/user-attachments/assets/cdc24429-a675-49db-8172-eaf315773c8c)

ubatsukh:
> Verified for the upcoming patch with `3.1.0-next.10`
> 
> ![Image](https://github.com/user-attachments/assets/cdc24429-a675-49db-8172-eaf315773c8c)

Oh so beautiful!!!

PR #11587, opened 2025-02-19
fix(slider): ensure histograms with non-zero min are displayed correctly

jcfranco:
**Related Issue:** #11583 

## Summary

Fixes histogram clipping by ensuring the y range accounts for the dataset’s min value. Previously, the y-axis scaling didn’t account for non-zero min values, leading to incorrect positioning and clipping.

### Notes

* Most histogram slider screenshot tests were merged into one as they are similar
* Added a histogram slider creator function for reuse between stories
* Removed E2E tests that relied on implementation details and were covered by screenshot tests

### DIFF
--- a/packages/calcite-components/src/components/graph/graph.e2e.ts
+++ b/packages/calcite-components/src/components/graph/graph.e2e.ts
@@ -29,7 +29,7 @@ async function createGraphWithData(): Promise<E2EPage> {
 
 describe("calcite-graph", () => {
   describe("renders", () => {
-    renders(`<calcite-graph></calcite-graph>`, { display: "block" });
+    renders(`calcite-graph`, { display: "block" });
   });
 
   describe("honors hidden attribute", () => {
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/graph/graph.e2e.ts
```diff
@@ -59,53 +59,6 @@ describe("calcite-graph", () => {
     ]);
   });
 
-  it("draws an area graph", async () => {
```
benelan, line 62: was removing this test intentional?

packages/calcite-components/src/components/slider/slider.stories.ts
```diff
@@ -231,65 +231,199 @@ rangeLabeledTicksEdgePositioningAtMin_TestOnly.parameters = {
   chromatic: { diffThreshold: 1 },
 };
 
-export const Histogram = (): HTMLCalciteSliderElement["el"]["el"] => {
+function createHistogramSlider({
```
benelan, line 234: Nice! :broom:

benelan (approved):
:+1:

packages/calcite-components/src/components/graph/graph.e2e.ts
benelan: ah nevermind, just saw the commit message

> drop E2E tests that are covered by screenshot tests

Issue #9246, opened 2024-05-02
[Button] Enhance component's interactive states


abigailmbravo-uxuidev:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

#### A stronger and more noticeable hover state for calcite-neutral outline and outline-fill buttons. 

![2024-05-02_13-15-56 (1)](https://github.com/Esri/calcite-design-system/assets/73944682/887960e3-290f-423c-ab5a-eabd7d24204d)

This giphy is showing and comparing the new hover to state to the old hover state (while one is brand and one is neutral), the drastic hover state difference is creating a weird and unexpected interaction.

Part of epic #9299

### Acceptance Criteria

- [x] Remove hover styling from all `:focus` states

- [x] 1. **`kind="neutral" appearance="solid"`**:
  - [x] Update `:hover` background color to `border.3`
  - [x] Update `:press` background color to `border.2`

- [x] 2. **`kind="neutral" appearance="outline-fill | outline"`**:
  - [x] Update all `:hover` stroke colors to `border.input`
  - [x] Update all `:press` stroke colors to `text.3`

- [x] 3. **`appearance="outline-fill"`**:
  - [x] Update `:hover` to `foreground.2` background change instead of a stroke width change
  - [x] Update `:press` to a `foreground.3` background change instead of a stroke width change

- [x] 4. **`appearance="outline"`**:
  - [x] Update `:hover` to `transparent.hover` background change instead of a stroke width change
  - [x] Update `:press` to a `transparent.press` background change instead of a stroke width change

![Image](https://github.com/user-attachments/assets/d173dc52-3274-4806-abaa-59e7b9e65c9b)

### Relevant Info

The neutral outline buttons are being added to the groups item details pages at R2. So if any type of enhancement can be made before then to make this a more clear experience, that would be great. 

### Which Component

Calcite-button

https://codepen.io/geospatialem/pen/dyLBvPm

### In this example, I am hovering on 'Add items to groups' - if you look closers, you can tell there is a very subtle difference. 
<img width="1327" alt="Screenshot 2024-05-02 at 1 21 17 PM" src="https://github.com/Esri/calcite-design-system/assets/73944682/96777a4e-8b6e-4d7d-b89a-a5a923ca35dd">

### Priority impact

p2 - want for current milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
[...]

PR #11590, opened 2025-02-19
feat(button): enhance component's interactivity states

aPreciado88:
**Related Issue:** [#9246](https://github.com/Esri/calcite-design-system/issues/9246)

## Summary

- Remove `hover` styling from all `focus` states.

- Update `button` demos page `outline-fill` section to use `appearance="outline-fill"` instead of `appearance="outline"`.

- `kind="neutral" appearance="solid"`:
  - Update `hover` `background-color` to `--calcite-color-border-3`.
  - Update `active` `background-color` to `--calcite-color-border-2`.

- `kind="neutral" appearance="outline-fill | outline"`:
  - Update all `hover` stroke colors to `--calcite-color-border-input`.
  - Update all `active` stroke colors to `--calcite-color-text-3`.

- `appearance="outline-fill"`:
  - Update `hover` to a `--calcite-color-foreground-2` `background` change instead of a stroke width change.
  - Update `active` to a `--calcite-color-foreground-3` `background` change instead of a stroke width change.

- `appearance="outline"`:
  - Update `hover` to a `--calcite-color-transparent-hover` `background` change instead of a stroke width change.
  - Update `active` to a `--calcite-color-transparent-press` `background` change instead of a stroke width change.


packages/calcite-components/src/components/button/button.scss
```diff
@@ -339,6 +335,12 @@
     --calcite-internal-button-background-color: var(--calcite-color-foreground-1);
     @apply border-solid;
     border-width: var(--calcite-button-border-size, 1px);
+    &:hover {
+      @apply bg-foreground-2;
```
jcfranco, line 339: Can you use our custom CSS props here (maps to `--calcite-color-foreground-2` and `--calcite-color-foreground-3` respectively)? We're moving away from Tailwind utils, so new code can use CSS vars directly.

packages/calcite-components/src/components/button/button.scss
```diff
@@ -404,12 +406,25 @@
...
       --calcite-internal-button-shadow-color: var(--calcite-color-foreground-3);
+      stroke: var(--calcite-color-text-3);
+    }
+  }
+}
+:host([appearance="solid"][kind="neutral"]) {
+  button,
+  a {
+    &:hover {
+      background-color: var(--calcite-color-border-3);
```
jcfranco, line 424: Can you set the internal button CSS props? Some [props are already set up](https://github.com/Esri/calcite-design-system/blob/aPreciado88/9246-button-enhance-component-interactivity-states/packages/calcite-components/src/components/button/button.scss#L42-L53) to use them to help simplify styles.

Applies to other `background-color`, and possibly `border-color`, assignments.

packages/calcite-components/src/components/button/button.scss
```diff
@@ -506,12 +528,14 @@
 
     &:hover {
       --calcite-internal-button-shadow-color: var(--calcite-color-foreground-3);
+      stroke: var(--calcite-color-border-input);
```
jcfranco, line 531: Should this be `border-color`? `stroke` is an [SVG prop](https://developer.mozilla.org/en-US/docs/Web/CSS/stroke). If so, can you set the internal border color CSS prop for consistency?

packages/calcite-components/src/components/button/button.e2e.ts
```diff
@@ -564,7 +564,7 @@ describe("calcite-button", () => {
     });
 
     it("should allow the CSS custom property to be overridden", async () => {
-      const overrideStyle = "rgba(255, 255, 0, 0.9)";
+      const overrideStyle = "rgba(0, 0, 0, 0)";
```
jcfranco, line 567: Some of these tests and assertions might no longer be valid. For example, this test checks if an override to `--calcite-color-transparent-hover` is applied to the focus state, which [was removed](https://github.com/Esri/calcite-design-system/pull/11590/files#diff-3105b16e07036988f253e3013a48bba5935ecc76287b9c47054fcd27bcd2e450L557). Can you double-check and restore the previous input test values after updating the tests?

jcfranco (changes requested):
Looking good, @aPreciado88! One more pass and I think this will be good to land.

packages/calcite-components/src/components/button/button.e2e.ts
aPreciado88: I removed the `focus` tests since they aren't needed anymore. Thank you for pointing that out!

jcfranco (approved):
🚀

ashetland (approved):
Lookin' geeeeewd!

Issue #11091, opened 2024-12-17
Improve tooltip appearance/disappearance timing


nCastle1:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Currently, Calcite tooltip is quite slow to disappear (and sometimes will just not disappear), but is very quick to appear on hover. This is the opposite of typical browser behavior, where tooltips are slow to appear, but disappear instantly. It is also different from the tooltip behavior of other libraries (e.g. bootstrap) that use floating-ui/poppers-js. 

Demo showing delays with Calcite tooltip, and failure to dismiss in some cases:

https://github.com/user-attachments/assets/76f85063-2527-47c6-8295-20fa5ef81f60

Another issue is that there seems to be a delay between when the mouse event happens and when the tooltip is shown or hidden; but by the time the tooltip appears it is no longer needed. This can look bad when multiple elements with tooltips are shown side-by-side, e.g. in an action bar:

https://github.com/user-attachments/assets/1dc3626b-7998-4ffe-ade8-d6cff91a84c7

Demo of another implementation:

https://github.com/user-attachments/assets/4b2aba83-06f5-47dc-a8af-1aeac30c89c3

Browser built-in tooltip behavior:

https://github.com/user-attachments/assets/a2f6b41f-9873-4877-9efb-bb75b72f30bf

(tooltip is slow to appear at first, but disappears instantly. If a tooltip has been shown very recently, subsequent tooltips are faster to appear)

### Acceptance Criteria

- Tooltips do not start to appear after the mouse has already left the reference element
- Tooltips disappear immediately after the mouse has left the reference element
- Tooltips do not appear when the mouse is not on the reference element

### Relevant Info

Behavior appears to be the same for 2.x and 3.x tooltips, so I don't believe the issues are caused by recent Stencil migration work.

### Which Component

calcite-tooltip

### Example Use Case

I would like to use calcite tooltips for elements in a calcite-action-pad, but performance/behavior was unacceptable, so I had to remove them and rely on built-in browser tooltips.

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

geospatialem:
[...]

Issue #6396, opened 2023-01-31
Epic: Tooltip - Improve component timing


geospatialem:
### Actual Behavior

`tooltip` lags to open or close, where a delay is evident when hovering on/off the reference element. 🐢 

### Expected Behavior

A more speedy response when interacting with the `tooltip` 🐎 🏎️

Should allow devs to turn off the response, if they wish to do so.

- Moving the mouse over the action, it takes too long for the tooltip to show up
- Moving the mouse to another action, it takes too long for the tooltip still on the other action to disappear and for the new tooltip to display
- When clicking on an action, the tooltip still shows

### Related issues
- [x] https://github.com/Esri/calcite-components/issues/6785
- [x] https://github.com/Esri/calcite-components/issues/6615
- [x] https://github.com/Esri/calcite-components/issues/6393
- [x] https://github.com/Esri/calcite-components/issues/6278

### Reproduction Sample

https://developers.arcgis.com/calcite-design-system/components/tooltip/

### Reproduction Steps

1. Open the [component sample](https://developers.arcgis.com/calcite-design-system/components/tooltip/)
2. Hover over the reference element
3. Wait
4. Observe the `tooltip` is displayed after a lag
5. Hover off the reference element
6. Wait
7. Observe the `tooltip` disappears after a lag

### Reproduction Version

1.0.3

### Relevant Info

Related to discussions from https://github.com/Esri/calcite-components/issues/6393

### Impact

Moderate

### Esri team

ArcGIS Online

tannerjt:
We've noticed this as well.  Specifically, when using `referenceElement` on `Safari v16.3`, the lag is about 2x longer than on `Chrome v109.0.5414.87`.

jcfranco:
@tannerjt What you're experiencing is a side effect from a polyfill we use to position floating-ui elements in non-Chromium browsers. https://github.com/Esri/calcite-components/pull/5542 has more info and also a workaround if you need to disable it (disabling will lead to incorrect positioning in some cases).

driskull:
> When clicking on an action, the tooltip still shows

The tooltip has a property `closeOnClick` which is false by default. We could change this if necessary.

According to w3c...

> A tooltip is a popup that displays information related to an element when the element receives keyboard focus or the mouse hovers over it. It typically appears after a small delay and disappears when Escape is pressed or on mouse out.

https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/
https://www.digitala11y.com/tooltip-role/

[...]

PR #11599, opened 2025-02-20
fix(tooltip): do not open after the pointer has moved off of the reference element

driskull:
**Related Issue:** #11091

## Summary

- No longer opens the tooltip after the pointer has moved off of the reference element
  - Clears open timeout when pointer moved and the same tooltip's reference element is no longer hovered
  - If opening and the hover tooltip doesn't match the tooltip being opened, nothing happens.
- Modify delay opening tooltips when a tooltip is already open from `0ms` to `100ms` which is a third the usual time.
- Modify delay closing tooltips from `500ms` to `450ms` which is a factor of `1.5` times the opening delay.


driskull:
@macandcheese @DitwanP would you mind testing this out?

driskull:
> If the TOOLTIP_QUICK_OPEN_DELAY_MS is what has made that improvement, maybe bumping it a touch more could make it not show up as often.

Currently, the tooltips open first at `300ms` but then after a tooltip is already open the next tooltip opens faster. Previously it was set to open at `0ms` from https://github.com/Esri/calcite-design-system/issues/6396 but here I increased it slightly to `25ms`

I can try bumping it to `150ms` and see if that helps.

DitwanP:
> I can try bumping it to `150ms` and see if that helps.

Definitely feels a bit better to me. 👍

DitwanP (approved):
LGMT! 👍

jcfranco (approved):

PR #11600, opened 2025-02-20
fix(tooltip): close tooltip when hovering out of an iframe

driskull:
**Related Issue:** #11091

## Summary

- Sets up an event listener on the document for when the pointer leaves the document
  - Closes tooltip when this occurs
- This is to support a mouse moving out of an iframe with tooltips


macandcheese (approved):
Code looks good! Is it possible / easy to add an example of this functionality to the local demos?

driskull:
@macandcheese added a demo.

packages/calcite-components/src/demos/tooltip-iframe.html
```diff
@@ -0,0 +1,19 @@
...
+    <style>
+      body {
+        margin: 0;
+        padding: 10%;
+      }
+    </style>
+  </head>
+
+  <body>
+    <iframe src="tooltip.html" width="80%" height="800"></iframe>
```
jcfranco, line 17: Can we add an E2E test instead of a one-off demo page? Not sure if this is reproducible in our E2E environment, but something like the following could help emulate the use case:

```tsx
  it("closes tooltip when pointer leaves document (simulates iframe use case)", async () => {
    await page.setContent(`
      <calcite-tooltip id="tooltip" reference-element="trigger">Awesome tooltip!</calcite-tooltip>
      <button id="trigger">Hover me</button>
    `);
    const button = await page.find('#trigger');
    const tooltip = await page.find('calcite-tooltip');

    await button.hover();
    await page.waitForTimeout(1000); // use proper open delay const
    await page.waitForChanges();

    expect(await tooltip.getProperty("open")).toBe(true);

    const viewport = page.viewport();
    // @ts-expect-error -- viewport is guaranteed here
    await page.mouse.move(viewport.width + 100, viewport.height + 100);
    await page.waitForChanges();

    await page.waitForTimeout(1000); // use proper close delay const
    expect(await tooltip.getProperty("open")).toBe(false);
  });
```

jcfranco (approved):
Awesome! 🚀

packages/calcite-components/src/components/tooltip/tooltip.e2e.ts
```diff
@@ -1396,6 +1396,29 @@ describe("calcite-tooltip", () => {
     });
   });
 
+  it.only("closes tooltip when pointer leaves document (simulates iframe use case)", async () => {
```
jcfranco, line 1399: This `only` snuck through.

packages/calcite-components/src/demos/tooltip-iframe.html
jcfranco: Can we remove this demo page as well?

Issue #10028, opened 2024-08-09
`TypeError` in `dropdown-group` when disconnected early


nwhittaker:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

If a `<calcite-dropdown-group>` is removed from the DOM before it's done loading, it can throw a `TypeError: Cannot read properties of null (reading 'querySelectorAll')`.

### Expected Behavior

It does not throw an error.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/zYVdgqQ

### Reproduction Steps

1. Visit the repro and open the devtools console
2. Click the **Toggle dropdown group** button and observe the `TypeError` log appears in the console

### Reproduction Version

2.11.1

### Relevant Info

Stack trace:
```
TypeError: Cannot read properties of null (reading 'querySelectorAll')
    at S.getGroupPosition (dropdown-group.tsx:166:29)
    at S.componentWillLoad (dropdown-group.tsx:85:31)
    at _e (index.js:1853:23)
    at ze (index.js:1649:22)
    at Array.r (index.js:1626:26)
    at M (index.js:185:13)
    at P (index.js:227:5)
```

I suspect treating `this.el.parentElement` here as optional would be sufficient:
https://github.com/Esri/calcite-design-system/blob/59c91c9363fa1041e7ffec8a9384ef464fe05233/packages/calcite-components/src/components/dropdown-group/dropdown-group.tsx#L166

### Priority impact

impact - p2 - want for an upcoming milestone

### Impact

The impact is largely with adding complexity to writing automated tests. In cases where the component is rendered as part of an acceptance test, its presence may not be important. However, the test still needs to wait for it to hydrate before performing an action that has the side-effect of removing the component.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Field Apps

driskull:
> I suspect treating this.el.parentElement here as optional would be sufficient:

Yeah I think that is likely the issue:

```js
private getGroupPosition(): number {
    return Array.prototype.indexOf.call(
      this.el.parentElement.querySelectorAll("calcite-dropdown-group"),
      this.el,
    );
  }
```

[...]

PR #11612, opened 2025-02-21
fix(dropdown-group): fix error caused by early removal

jcfranco:
**Related Issue:** #10028

## Summary

This fixes an error caused by `dropdown-group`s looking up their parent element. It updates `dropdown` to set the position directly on groups, avoiding parent lookups.

### DIFF
--- a/packages/calcite-components/src/components/dropdown-group/dropdown-group.tsx
+++ b/packages/calcite-components/src/components/dropdown-group/dropdown-group.tsx
@@ -27,9 +27,6 @@ export class DropdownGroup extends LitElement {
 
   // #region Private Properties
 
-  /** position of group within a dropdown */
-  private groupPosition: number;
-
   private mutationObserver = createObserver("mutation", () => this.updateItems());
 
   /** the requested group */
@@ -44,6 +41,13 @@ export class DropdownGroup extends LitElement {
 
   /** Specifies and displays a group title. */
   @property({ reflect: true }) groupTitle: string;
+
+  /**
+   * The position of the group in the dropdown menu.
+   *
+   * @internal
+   */
+  @property() position: number = -1;
 
   /**
    * Specifies the size of the component inherited from the parent `calcite-dropdown`, defaults to `m`.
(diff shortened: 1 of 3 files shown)


packages/calcite-components/src/components/dropdown/dropdown.tsx
```diff
@@ -431,11 +431,14 @@ export class Dropdown
...
 
     this.updateItems();
-    this.updateGroupScale();
+    this.updateGroupProps();
   }
 
-  private updateGroupScale(): void {
-    this.groups?.forEach((group) => (group.scale = this.scale));
+  private updateGroupProps(): void {
+    this.groups.forEach((group, index) => {
```
driskull, line 438: Nice, this is much better having the dropdown set its group index.

packages/calcite-components/src/components/dropdown-group/dropdown-group.tsx
```diff
@@ -123,13 +123,6 @@ export class DropdownGroup extends LitElement {
     );
   }
 
-  private getGroupPosition(): number {
```
driskull, line 126: yay no more looking up the dom!

packages/calcite-components/src/components/dropdown-group/dropdown-group.tsx
```diff
@@ -45,6 +42,13 @@ export class DropdownGroup extends LitElement {
   /** Specifies and displays a group title. */
   @property({ reflect: true }) groupTitle: string;
 
+  /**
+   * The position of the group in the dropdown menu.
+   *
+   * @internal
+   */
+  @property() position: number = -1;
```
driskull, line 50: Should there be a `defaults` test to test for `-1` now that its not undefined by default?

driskull (approved):
👍

ashetland (approved):
LGTM!

packages/calcite-components/src/components/dropdown-group/dropdown-group.tsx
jcfranco: We can skip it since it's an implementation detail.

PR #11625, opened 2025-02-26
refactor(input-date-picker): wire up useFocusTrap controller

jcfranco:


anveshmekala (approved):
👍  LGTM!

packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -130,11 +121,25 @@ export class InputDatePicker
...
+    focusTrapOptions: {
+      onActivate: () => {
+        if (this.focusOnOpen) {
+          this.datePickerEl?.setFocus();
+          this.focusOnOpen = false;
+        }
+      },
+      allowOutsideClick: true,
+      // Allow outside click and let the popover manager take care of closing the popover.
+      clickOutsideDeactivates: false,
```
anveshmekala, line 135: Nitpick: Do we need to specify this? 

`clickOutsideDeactivates: false` is the default option.

jcfranco: It's false in `focus-trap`, but we override the default in our [controller/utils](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/utils/focusTrapComponent.ts#L64).

jcfranco: Side note: I'll get rid of the comment since it's not applicable.

anveshmekala: > It's false in `focus-trap`, but we override the default in our [controller/utils](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/utils/focusTrapComponent.ts#L64).

Interesting, is this intentional or do we need to specify it at component level?

jcfranco: I believe it's intentional. We can revisit it if needed. 😄

Issue #11637, opened 2025-02-26
[Card Group] Default spacing is not provided


macandcheese:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The Card Group is not correctly assigning a default gap value.

![Image](https://github.com/user-attachments/assets/297f2209-e4e2-427a-a076-19046cdc6932)

### Expected Behavior

The default gap should display as expected.

![Image](https://github.com/user-attachments/assets/b40b8b94-4d9c-4c7b-a944-4301e738f9b4)

### Reproduction Sample

Below

### Reproduction Steps

Compare: 
3.x: https://codepen.io/mac_and_cheese/pen/ByaQpEy?editors=100
2.13.2: https://codepen.io/mac_and_cheese/pen/wBvogZa?editors=100

### Reproduction Version

3.x

### Relevant Info

Likely introduced here https://github.com/Esri/calcite-design-system/pull/10727/files

### Regression?

2.13.2

### Priority impact

impact - p1 - need for current milestone

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

geospatialem:
Verified in `3.1.0-next.20` 🎉 

![Image](https://github.com/user-attachments/assets/30ada832-fc46-4d50-ad50-ce678cdef4d6)

PR #11638, opened 2025-02-26
fix(card-group): Restore default gap spacing

macandcheese:
**Related Issue:** #11637 

## Summary
Updates token to restore card group gap


driskull (approved):
👍

Issue #11546, opened 2025-02-13
[Licensing] Add third party copyright licensing


geospatialem:
### Description

Calcite needs to document third party copyright licensing, which also includes [Lit](https://github.com/lit/lit/blob/main/LICENSE), in alignment with Maps SDK for JS, and other Esri products.

Calcite must provide the disclosure at each access point, including the CDN, NPM, and GitHub.

### Acceptance critiera

- [ ] Add a `copyright.txt` to the CDN similar to the [JS team](https://js.arcgis.com/4.31/esri/copyright.txt)
    - [ ] Include third party copyright licensing 
- [ ] Add to the "Copyright" section on GitHub under the README.md
    - [ ] By default this would also add it to npm 
- [ ] Add to the "Copyright" section to each CC package on GitHub

### Resources

- JS cdn: https://js.arcgis.com/4.31/esri/copyright.txt
- Lit licensing (BSD-3): https://github.com/lit/lit/blob/main/LICENSE

DitwanP:
🍠 Verified locally

geospatialem:
Additional context for future sleuthing 🕵🏻:

We copy the [license](https://js.arcgis.com/calcite-components/3.3.3/LICENSE.md) and the [3rd-party licenses](https://js.arcgis.com/calcite-components/3.3.3/THIRD-PARTY-LICENSES.md) in for the [Esri CDN build](https://devtopia.esri.com/WebGIS/calcite-design-system/blob/main/scripts/prep-cdn-contents.js#L72).

PR #11640, opened 2025-02-27
chore: set up 3rd-party license generation

jcfranco:
**Related Issue:** #11546 

## Summary

This adds a utility script that generates 3rd-party license information for each package.

### Notes

* file name is `THIRD-PARTY-LICENSES.md` (TPL)
* TPL is referenced in each package's README
* all packages are set up to generate a TPL (except the internal `@esri/eslint-config-calcite` package)
  * if there are no 3rd-party dependencies, it will display an appropriate message
* relies on https://www.npmjs.com/package/generate-license-file
* packages will now include TPL when published
* util script (`util:update-3rd-party-licenses`) can be called from the monorepo root or from individual packages
  * to ensure `generate-license-file` generates licenses correctly, this script installs `node_modules` and `package-lock.json` locally in each package (removed when finished)
  * I’ve created https://github.com/TobyAndToby/generate-license-file/issues/597 to add support for root-level monorepo setups, which would eliminate the need for the above step.


benelan (approved):
Nice this looks good! So we have to remember to rerun that script whenever a dependency gets added to any of the packages, right?

I wonder if we should automatically run the script before publishing just in case we forget at some point. At least that way the package's license will be included. 

You could run the script and `git add **/THIRD-PARTY-LICENSES.md` before committing here: 

https://github.com/Esri/calcite-design-system/blob/c88a31f330632e6649fe63b8d08b2d70f4da7c0b/.github/workflows/deploy-latest.yml#L75-L89

support/createThirdPartyLicenses.ts
```diff
@@ -0,0 +1,67 @@
...
+      "@arcgis/lumina",
+      "@esri/calcite-components",
+      "@esri/calcite-components-react",
+      "@esri/calcite-design-tokens",
+      "@esri/calcite-ui-icons",
+      "@esri/eslint-config-calcite",
+      "@esri/eslint-plugin-calcite-components",
+    ];
+
+    const licenses = await getProjectLicenses(".package.json", {
```
benelan, line 29: Should this be `"./package.json"` or `"package.json"`?

jcfranco: Oh, good catch. 🎣 It seems that `generate-license-file`'s file [path logic](https://github.com/TobyAndToby/generate-license-file/blob/main/src/packages/generate-license-file/src/lib/internal/resolveDependencies/resolveNpmDependencies.ts#L22) was unintentionally handling this case.

jcfranco:
> You could run the script and `git add **/THIRD-PARTY-LICENSES.md` before committing here:

Great idea. I've updated the `deploy-next` action to run the script and stage the license files as suggested. LMK if further tweaks are needed.

Issue #9611, opened 2024-06-17
Calcite-input-number validation prevents number authoring


janett-baresel:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

It is cumbersome to change large numbers with lots of 0 digits at the moment.

1. Enter a large number (e.g. "5000") into a number input field
2. Number wants to be changed to "7000" -> go with the cursor behind "5" and remove it
3. -> the whole field resets to 0

### Expected Behavior

1. Enter a large number (e.g. "5000") into a number input field
2. Number wants to be changed to "7000" -> go with the cursor behind "5" and remove it
3. Field shows 000 and user can type "7" to change it to 7000
4. "000" is only reset to 0 when the user tabs out of the field

### Reproduction Sample

use the number input doc sample

### Reproduction Steps

Repro in behavior description

### Reproduction Version

2.9.0

### Priority impact

impact - p3 - not time sensitive

### Impact

This is one of those annoyances when using a UI. Users who enter large numbers will almost always run into this and it is inconvenient but not critical obviously. 

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Scene Viewer

DitwanP:
🍠 Verified locally on `dev`

PR #11705, opened 2025-03-06
fix(input-number): allow editing of numbers that start with zeros

josercarcamo:
**Related Issue:** #9611

## Summary
Allows editing of numbers that start with zeros


josercarcamo:
@benelan could you please review?

benelan (changes requested):
There are some cases that don't work with this regex solution, can you take another look? Please reach out if you have any questions or if you'd like some implementation ideas!

packages/calcite-components/src/components/input-number/input-number.tsx
```diff
@@ -845,11 +845,17 @@ export class InputNumber
       );
     }
 
-    // adds localized trailing decimal separator
-    this.displayedValue =
-      hasTrailingDecimalSeparator && isValueDeleted
-        ? `${newLocalizedValue}${numberStringFormatter.decimal}`
-        : newLocalizedValue;
+    const re = new RegExp(`^-?0+\\d*${numberStringFormatter.decimal}?$`);
```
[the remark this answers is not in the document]
josercarcamo: But the code is currently using this exact notation in lines 133-137 of number.ts:

`const allLeadingZerosOptionallyNegative = /^([-0])0+(?=\d)/;
const decimalOnlyAtEndOfString = /(?!^\.)\.$/;
const allHyphensExceptTheStart = /(?!^-)-/g;
const isNegativeDecimalOnlyZeros = /^-\b0\b\.?0*$/;
const hasTrailingDecimalZeros = /0*$/;`

Plus all current e2e tests pass.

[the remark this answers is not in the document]
benelan: > Plus all current e2e tests pass.

The existing tests don't cover this leading zero bug, which is why they aren't failing. However, the test you added wouldn't pass if you modified it to use the `group-separator` attribute. Same goes for the other cases I mentioned in my previous comment.

> But the code is currently using this exact notation in lines 133-137 of number.ts:
> 
> `const allLeadingZerosOptionallyNegative = /^([-0])0+(?=\d)/; const decimalOnlyAtEndOfString = /(?!^\.)\.$/; const allHyphensExceptTheStart = /(?!^-)-/g; const isNegativeDecimalOnlyZeros = /^-\b0\b\.?0*$/; const hasTrailingDecimalZeros = /0*$/;`

My bad, the use of `numberStringFormatter.decimal` in the regex had me confused for a second. With input there are two different values:

1. the `value` property, which is **always** in the `en` locale and `latn` numbering system
2. the `displayValue` state, which is in the user-specified locale and numbering system

The regex you referenced are part of the number sanitation for the `value` property, which why it doesn't use the locale utils. The leading zeros get trimmed during number sanitation. After the number sanitation, the value is localized for the `displayValue`.

The trimmed, leading zeros need to be re-added to the `displayValue`, so they should be in the specified numbering system. The `value` property doesn't need the zeros re-added.

So the steps that need to be done for this issue are:

1. Check for leading zeros before the number sanitation (like how we check for trailing decimals here):

https://github.com/Esri/calcite-design-system/blob/b9624f931b936c47f6879cb80f490a87dd406442/packages/calcite-components/src/components/input-number/input-number.tsx#L823-L824

2. Add back trimmed, leading zeros after the localization (like we are adding localized trailing decimal zeros here):

https://github.com/Esri/calcite-design-system/blob/b9624f931b936c47f6879cb80f490a87dd406442/packages/calcite-components/src/components/input-number/input-number.tsx#L840-L846

Notice how the `0` is being localized into the specified numbering system in the util:

https://github.com/Esri/calcite-design-system/blob/b9624f931b936c47f6879cb80f490a87dd406442/packages/calcite-components/src/utils/number.ts#L261

The trailing, decimal zeros logic was added in #7159, I recommend taking a look that PR's implementation and discussion.

Please don't hesitate to reach out on Teams if you have any follow up questions. The input-number logic has a lot of features and edge cases, so it can be difficult to navigate at first.

packages/calcite-components/src/components/input-number/input-number.tsx
```diff
@@ -845,12 +848,27 @@ export class InputNumber
...
+              valueHandleInteger.length - 1 - (hasTrailingDecimalSeparator ? 1 : 0),
+              numberStringFormatter.localize("0"),
+            )}`
+        : this.displayedValue.padStart(
+            valueHandleInteger.length,
+            numberStringFormatter.localize("0"),
+          );
+      this.displayedValue += hasTrailingDecimalSeparator ? numberStringFormatter.decimal : "";
+    }
+
```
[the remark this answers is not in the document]
benelan: We may want to use the `isValueDeleted` boolean for the `hasLeadingZeros` conditional too. Can you investigate?

josercarcamo:
@benelan could you please review?

packages/calcite-components/src/components/input-number/input-number.e2e.ts
```diff
@@ -1300,6 +1300,57 @@ describe("calcite-input-number", () => {
           const input = await page.find("calcite-input-number >>> input");
           expect(await input.getProperty("value")).toBe("0");
         });
+
+        it(`allows editing numbers that start with zeros ${locale} locale`, async () => {
+          const page = await newE2EPage();
+          await page.setContent(html`<calcite-input-number locale=${locale}></calcite-input-number>`);
```
benelan, line 1306: FYI the attribute to change the locale is `lang`, not `locale`

packages/calcite-components/src/components/input-number/input-number.e2e.ts
```diff
@@ -1300,6 +1300,57 @@ describe("calcite-input-number", () => {
...
+          await page.keyboard.press("Backspace");
+          await page.waitForChanges();
+          expect(await calciteInput.getProperty("value")).toBe("11");
+          expect(await input.getProperty("value")).toBe("0011");
+
+          await page.keyboard.type("5");
+          await page.waitForChanges();
+          expect(await calciteInput.getProperty("value")).toBe("50011");
+          expect(await input.getProperty("value")).toBe("50011");
+        });
```
benelan, line 1353: I don't think we need to test this feature for every locale. 

Can you remove this test and add a new one on line 1544 that does the following:

- uses the `arab` numbering system 
- uses the `ar` locale
- uses a minus sign
- uses a decimal

packages/calcite-components/src/components/input-number/input-number.e2e.ts
```diff
@@ -1490,6 +1541,57 @@ describe("calcite-input-number", () => {
     expect(await input.getProperty("value")).toBe(numberStringFormatter.localize(initialValue));
   });
 
+  it("allows editing numbers that start with zeros and group separator", async () => {
+    const page = await newE2EPage();
+    await page.setContent(html`<calcite-input-number></calcite-input-number>`);
```
benelan, line 1546: This test says it is using a group separator, but the attribute is missing from the component

```suggestion
    await page.setContent(html`<calcite-input-number group-separator></calcite-input-number>`);
```

benelan (commented):
Glad my suggestions worked, thanks for testing! The e2e tests need some changes and then this will be good to go :rocket:

packages/calcite-components/src/components/input-number/input-number.e2e.ts
```diff
@@ -1552,44 +1552,31 @@ describe("calcite-input-number", () => {
...
 
     await page.keyboard.press("ArrowLeft");
     await page.keyboard.press("ArrowLeft");
     await page.keyboard.press("ArrowLeft");
     await page.keyboard.press("Backspace");
     await page.waitForChanges();
-    expect(await calciteInput.getProperty("value")).toBe("0");
-    expect(await input.getProperty("value")).toBe("000");
+    expect(await calciteInput.getProperty("value")).toBe("7000");
+    expect(await input.getProperty("value")).toBe("7000");
```
benelan, line 1563: Based on the modified expected value here, it looks like you are deleting the group separator instead of the `7`. You will need to add another `ArrowLeft` keypress now that there is an additional character.

Alternatively you can do this, which doesn't depend on the length of the number:

```ts
    await page.keyboard.press("Home");
    await page.keyboard.press("Delete");
```

The other expected values below will also need to be updated to ensure they are testing the leading zero fix.

packages/calcite-components/src/components/input-number/input-number.e2e.ts
```diff
@@ -1490,6 +1490,67 @@ describe("calcite-input-number", () => {
...
+    await page.keyboard.press("ArrowRight");
+    await page.keyboard.type("7");
+    await page.waitForChanges();
+    expect(await calciteInput.getProperty("value")).toBe("-7000");
+    expect(await input.getProperty("value")).toBe("-7,000");
+  });
+
+  it(`allows editing numbers that start with zeros and have decimals in the ar locale and arab numbering system`, async () => {
+    const page = await newE2EPage();
+    await page.setContent(html`<calcite-input-number lang="ar"></calcite-input-number>`);
```
benelan, line 1536:
```suggestion
    await page.setContent(html`<calcite-input-number lang="ar" numbering-system="arab"></calcite-input-number>`);
```

You need to use the `numbering-system` attribute in order to change the numbering system. It doesn't change automatically when setting `lang` because some users want the latin numbering system and arab locale, for example.

benelan (approved):
I pushed some changes to fix the arabic test, LGTM now!

packages/calcite-components/src/demos/input-number.html
```diff
@@ -135,8 +135,14 @@
         <div class="child-flex font right-aligned-text">Number Vertical</div>
         <div class="child-flex">
           <calcite-label scale="s">
-            Input Label
-            <calcite-input-number placeholder="Placeholder" scale="s" value="123" step="1"></calcite-input-number>
+            Input Label AR2
```
jcfranco, line 138: @josercarcamo for consistency, could you restore the previous HTML and either:

* move this particular Arabic configuration to a new section or
* omit it for now?

Issue #11391, opened 2025-01-27
Convert Tokens to Style Dictionary 4 and Token Studio 1


alisonailea:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

Requires a major refactor of `calcite-design-tokens`

This will remove several (now unnecessary) custom token transformer functions. Align with the latest Token Studio and Style Dictionary versions.

Blocked issues: https://github.com/Esri/calcite-design-system/issues/10051

### Requirements checkoff

- [x] Reorganize files to align with the new StyleDictionary config format
- [x] Create new config files
- [x] Refactor or (re)write StyleDictionary functions
    - [x] Transformers to match SD format
        - [x] 			Calcite Transformation Group
        - [x] 			Transform name to replace "-" with "minus" and "+" with "plus"
        - [x] 			Transform name to remove the Tier
        - [x] 			Transform name to remove the word "default"
        - [x] 			Transform value of boxshadow for CSS platforms
        - [x] 			Transform any number value to calculate sum
        - [x] 			Transform px value to rem for space, size, font-size, etc.
        - [x] 			Transform breakpoint value to px
    - [x] 		Headers to match SD format
        - [x] 			Default Calcite header for generated files
        - [x] 			Calcite header for deprecated generated files
    - [x] 		Formats to match SD format & normalize file names (aka "global --> semantic")
        - [x] 			/css
            - [x] 				breakpoints.css
            - [x] 				classes.css (typography classes)
            - [x] 				core.css
            - [x] 				dark.css
            - [x] 				light.css
            - [x] 				semantic.css
            - [x] 				global.css (deprecate)
            - [x] 				index.css 
                - [x] 					import semantic.css and classes.css
                - [x] 					Set light vars at :root
                - [x] 					@media (prefers-color-scheme: light) { .calcite-mode-auto { …light vars } }
                - [x] 					@media (prefers-color-scheme: dark) { .calcite-mode-auto { …dark vars } }
                - [x] 					.calcite-mode-light { …light vars }
                - [x] 					.calcite-mode-dark { …dark vars }
        - [x] 			/scss
            - [x] 				breakpoints.scss
            - [x] 				classes.scss (typography mixins)
            - [x] 				core.scss
            - [x] 				dark.scss
            - [x] 				light.scss
            - [x] 				semantic.scss
            - [x] 				global.css 
                - [x] 					Deprecate
[...]

Issue #10510, opened 2024-10-09
[Tokens] Implement a `--calcite-color-focus` global variable across the design system


COV-GIS:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description
Provide a global token/CSS variable that enables designers and developers to set the focus color, which could be named `--calcite-color-focus`.

The above is currently achievable via the **_deprecated_**  `--calcite-ui-focus-color` CSS variable, but its usage doesn't apply to the use case, Calcite should instead opt to provide support for focus and brand color variables.

Workaround example with `--calcite-ui-focus-color` (Note: this is a deprecated CSS variable): https://codepen.io/geospatialem/pen/poMNgbO

### Acceptance Criteria

- [x] Allow developers to use a global `--calcite-color-focus` token that allows customization of the focus in apps. 
    - [x] For instance, support Button's focus color to match button color when kind is danger

### Which Component

Across components

### Example Use Case

Allow developers to use a global `--calcite-color-focus` token that allows customization of the focus in apps. For instance, support Button's focus color to match button color when kind is danger.

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [x] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

N/A

geospatialem:
@COV-GIS This is currently achievable using the `--calcite-color-brand` CSS variable when specifying the `kind` attribute or providing a class selector, for instance: https://codepen.io/geospatialem/pen/poMNgbO

```css
calcite-button[kind="danger"] {
  --calcite-ui-focus-color:var(--calcite-color-status-danger);
}
```


With the above in mind we will work towards a more streamlined implementation using a new CSS variable/token of `--calcite-color-focus` and provide additional documentation in the future. Will be updating the title and scope, as we do not plan to provide specific support on component's `kind` attr/prop, as it would be up to developers to achieve.

cc: @alisonailea @brittneytewks

geospatialem:
Verified with Chromatic via https://codepen.io/geospatialem/pen/MWNGoey using the new `--calcite-color-focus` global token/CSS variable.

COV-GIS:
Awesome! Thanks folks.

PR #11711, opened 2025-03-07
fix: allow global focus color token to inherit fallback value

jcfranco:
**Related Issue:** #10510  

## Summary  

Removes the global focus color introduced in [#10512](https://github.com/Esri/calcite-design-system/pull/10512) to match the behavior of the deprecated `--calcite-ui-focus-color`. Previously, `--calcite-color-focus` was defined at `:root`, preventing it from inheriting updates to `--calcite-color-brand` (its fallback).  

This is **not** considered a breaking change because:  

* it only affects outputs that are not publicly documented
* the default color will still be applied as expected
* `--calcite-color-focus` remains available for overriding the focus color

**Note**: https://github.com/Esri/calcite-design-system/issues/11713 will restore `--calcite-color-focus` after #11391 lands.

### DIFF
--- a/packages/calcite-design-tokens/src/semantic/color.json
+++ b/packages/calcite-design-tokens/src/semantic/color.json
@@ -54,27 +54,6 @@
           }
         }
       },
-      "focus": {
-        "type": "color",
-        "value": "{semantic.color.brand.default.default}",
-        "attributes": {
-          "calcite-schema": {
-            "system": "calcite",
-            "tier": "semantic",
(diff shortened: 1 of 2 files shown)


macandcheese (approved):
Looks good - could we adjust the title or use "tooling", etc., so users don't get confused by this changelog entry and think the functionality is being removed?

jcfranco:
@macandcheese Good idea. I'll revisit the title.

PR #11719, opened 2025-03-07
feat: add native-button-tab-index-required rule

anveshmekala:
**Related Issue:** #


driskull:
@anveshmekala I'm not finding any info about Safari requiring a tabindex on buttons.

I also can't see any difference in a codepen: https://codepen.io/driskull/pen/OPJKRbw

Are you sure the issue wasn't related to focus trapping and incorrectly closing the input-date-picker?

driskull:
I think there may be a larger issue with the component, even clicking on the background of the open date picker shouldn't close it. This is an inconsistency between Safari and Chrome and adding the tabindex to the button doesn't fix that part.

driskull:
The issue seems to be with the `this.listen("blur", this.blurHandler);` blur handler, not the tabindex on button. When clicking on the floating UI dialog its calling blur on the host which is closing it. It doesn't do that in Chrome. So something in that logic needs to change.

driskull (changes requested):
I'm not sure this is necessary but I'll defer to @jcfranco.

I don't think safari needs tabindex set and can't seem to find any information to back that up.

anveshmekala:
@driskull The inconsistency is mentioned [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#clicking_and_focus:~:text=Whether%20clicking%20on%20a%20%3Cbutton%3E%20or%20%3Cinput%3E%20button%20types%20causes%20it%20to%20(by%20default)%20become%20focused%20varies%20by%20browser%20and%20OS.%20Most%20browsers%20do%20give%20focus%20to%20a%20button%20being%20clicked%2C%20but%20Safari%20does%20not%2C%20by%20design). 

While debugging https://github.com/Esri/calcite-design-system/issues/11544, i stumbled up on it. IIRC, safari wasn't focusing the button by default. But the Codepen you shared is consistent across browsers.

driskull:
The root of the issue is the blur event causing the component to close in Safari the button being focused does hide the issue but the root of the issue persists and clicking not on a button within the datepicker shouldn't close it either.

anveshmekala:
> The root of the issue is the blur event causing the component to close in Safari the button being focused does hide the issue but the root of the issue persists and clicking not on a button within the datepicker shouldn't close it either.

True & Agree. We can freeze this one & fix the root cause. Can you add a bug for the issue with input-date-picker.

jcfranco:
Thanks for the great discussion, y'all. Is this and #11941 still valid after identifying the root cause?

> > The root of the issue is the blur event causing the component to close in Safari the button being focused does hide the issue but the root of the issue persists and clicking not on a button within the datepicker shouldn't close it either.
> 
> True & Agree. We can freeze this one & fix the root cause. Can you add a bug for the issue with input-date-picker.

We still need an issue for this, right?

Issue #2311, opened 2021-06-11
Enhancement: Select all for combobox


jgibson02:
### Description
Provide a way to include a "select all/none" button for ~~a pick-list~~ Combobox

### Acceptance Criteria 
Final designs in this [Figma file](https://www.figma.com/file/ujZYHdimnVC3yA2PyYOH85/Epic%3A-Combobox-Redesign?type=design&node-id=274-82801&mode=dev). These designs also take into account updated selection icons from issue #10580.
#### Which Component
~~pick-list~~ Combobox

#### Example Use Case
In ArcGIS Velocity we have a page with a data grid for items with a dropdown filter for item status (currently a calcite-react `MultiSelect`):
![image](https://user-images.githubusercontent.com/5069711/121732511-33107d80-caa7-11eb-8e25-afa13e0ef9d0.png)
![image](https://user-images.githubusercontent.com/5069711/121732547-3c99e580-caa7-11eb-951b-d2ef691d020d.png)
![image](https://user-images.githubusercontent.com/5069711/121732577-43c0f380-caa7-11eb-9328-976d8203655d.png)

We decided to add a "select all/none" button to the list since it's common for a user to want to quickly filter down to a single/few status types but then want to go back to seeing everything.

**monday.com sync:** #5220192358

macandcheese:
Super valid. Some thoughts: 


- This should be an opt-in functionality.
- This has also come about the combobox as well, so any solution should be designed with both of these components in mind (dropdown multi-select too ?).
  - For combobox, this might require a new prop (also valid without the select-all use case), `max-displayed-chips` (?), any chips after which are appended to a “+n” chip. 
  - Invokes the follow up question,  “can you click the "+n" chip to show those items”

bstifle:
Here's my design for this!

![select all](https://user-images.githubusercontent.com/14083530/127689498-33d432c5-1e3b-440e-8e1a-840ebe8d75fd.png)

macandcheese:
I think we'll want to re-evaluate the above design - I also don't think this is valid for Dropdown.

The screenshot workflow in OP is valid for Combobox as shown, but likely also Pick List as described. @asangma do you have thoughts on select / deselect workflow in pick List?

Do we want to separate this out to scope the enhancements to Combobox from those to Pick List?

jcfranco:
> Do we want to separate this out to scope the enhancements to Combobox from those to Pick List?

[...]

Issue #6466, opened 2023-02-14
Sticky List Item Group Header


am-maneaters:
### Description

Set ListItemGroup's header to be sticky so that it sticks to the top of the visible list. This will give users more context when they are looking at a list with many ListItems per group because it will make the GroupHeader persistent even when scrolling through many items.

### Acceptance Criteria

- [ ] ListItemGroup's header is sticky by default. When a user scrolls down in a list with groups, the header of the group will continue to display at the top of the group.
- [ ] Only the header of the partially displayed group should be sticky. The header should scroll off the screen once the bottom of its respective group scrolls off.
- [ ] When there are multiple groups, only one header should be sticky at a time since only one group can be partially displayed at a time (they may technically still be sticky, but they are no longer visible).
- [ ] For nested groups, only the lowest level header should be sticky. If there is a parent group and child group, only the child group's header should be sticky.

### Which Component

List/ListItemGroup

### Example Use Case

MUI's implementation is a perfect example of the expected behavior: https://mui.com/material-ui/react-list/#sticky-subheader

### Esri team

Professional Services - Midwest Delivery Center

**monday.com sync:** #8745738744

### Original issue description

>### Description
>
>Add a prop to ListItemGroup called `sticky` which would make the ListItemGroup header stick to the top of the list. This would give users more context when they are looking at a list with many ListItems per group because it would make the GroupHeader persistent even when scrolling through many items.
>
>### Acceptance Criteria
>
>ListItemGroup has a prop that allows users to designate it as sticky. Then when a user scrolls down in a list with groups, the header of the group will continue to display at the top.

ashetland:
Do we think this behavior could also apply to Combobox Groups?

driskull:
> Do we think this behavior could also apply to Combobox Groups?

This question doesn't appear to be answered. What about autocomplete, dropdown, combobox, etc?

Do we need separate issues? cc @DintaMel @geospatialem @isaacbraun

ashetland:
I think it makes sense for Autocomplete/Combobox. For Dropdown, I'd need some convincing. cc @Esri/calcite-designers

DintaMel:
@ashetland can you help me understand your hesitations on Dropdown?

ashetland:
I think ideally dropdown menus wouldn't scroll and so would not need group headings to be sticky. If a dropdown is that long, I think the better pattern would be submenus (#9745).

ashetland:
[...]

PR #11721, opened 2025-03-08
feat(combobox): add `selectAll` toggle property

Elijbet:
**Related Issue:** #2311

## Summary
Provides a `Select All` checkbox in the `calcite-combobox` by default for `selection-mode="multiple"` as a convenient shortcut for users to quickly select or deselect all items in the list. The `indeterminate` state for this checkbox is a visual and programmatic state that indicates a checkbox is neither fully checked nor unchecked. It represents a "partial selection" in scenario where a checkbox is associated with a group of child checkboxes, and some (but not all) of the child checkboxes are selected.


packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -386,6 +391,12 @@ export class Combobox
   /** Specifies the size of the component. */
   @property({ reflect: true }) scale: Scale = "m";
 
+  /** When `true`, provides a toggle for selecting all items. Does not apply to `selection-mode single`. */
+  @property({ reflect: true }) selectAll = false;
+
+  /** When `true`, selects all items. Does not apply to `selection-mode single`. */
+  @property({ reflect: true }) selectAllChecked = false;
```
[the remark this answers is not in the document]
driskull: Maybe this could be a getter? 

```
get allSelected() {
  // returns true if every item is selected?
}
```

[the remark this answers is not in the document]
driskull: Should this work as an indeterminate checkbox?

https://codepen.io/chriscoyier/pen/JYyXjX

What if a user toggles all of the checkboxes on, would the `select all` checkbox be checked?

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1682,6 +1771,28 @@ export class Combobox
...
+                    aria-labelledby={`${this.guid}-select-all-label`}
+                    checked={this.selectAllCheckedState}
+                    id={`${this.guid}-select-all`}
+                    onChange={this.selectAllChangeHandler}
+                    onKeyDown={this.selectAllKeyDownHandler}
+                    ref={this.setCheckboxReferenceEl}
+                    type="checkbox"
+                  />
+                  <label htmlFor={`${this.guid}-select-all`} id={`${this.guid}-select-all-label`}>
+                    Select all
```
driskull, line 1792: should be localized text

packages/calcite-components/src/components/combobox/combobox.tsx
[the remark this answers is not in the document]
Elijbet: The select all checkbox is intended to be a shortcut to quickly select all checkboxes, it adds one chip "All selected" for all. If user toggles all of the checkboxes on without "Select All" option present (the prop isn't set to true), the "All selected" chip appears (it worked this way prior), but the checkbox doesn't. 

@SkyeSeitz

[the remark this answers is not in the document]
SkyeSeitz: Any objections to matching the behavior of Table? This does what Matt is bringing up where the Select all checkbox has an indeterminate state and gets checked when all items are selected manually. 
https://github.com/user-attachments/assets/45c02c88-4cb0-45f8-8e4f-5745058d8e09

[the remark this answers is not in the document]
SkyeSeitz: cc @ashetland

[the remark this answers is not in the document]
Elijbet: That feels right. I wasn't familiar with the full behavior of indeterminate.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -388,6 +392,9 @@ export class Combobox
   /** Specifies the size of the component. */
   @property({ reflect: true }) scale: Scale = "m";
 
+  /** When `true`, provides a toggle for selecting all items. Does not apply to `selection-mode single`. */
+  @property({ reflect: true }) selectAllEnabled = false;
```
eriklharper, line 396: Brainstorm: I kinda wonder if instead of creating a new property to enable this, we just add a new `selection-mode` called `all` which implies `multiple` and also adds the ability to select all items?  Thoughts on that?  @jcfranco @geospatialem @SkyeSeitz @ashetland

packages/calcite-components/src/components/combobox-item/combobox-item.scss
```diff
@@ -162,5 +164,15 @@ ul:focus {
   line-height: var(--calcite-font-line-height-relative-snug);
 }
 
+.select-all-checkbox {
+  position: sticky;
+  inset-block-start: 0;
+  background-color: var(--calcite-combobox-item-background-color-active, var(--calcite-color-foreground-1));
+  z-index: 1;
+  box-shadow: 0 1px 2px rgba(0, 0, 0, 0.1);
```
driskull, line 172: Do we need a css var for this shadow?

packages/calcite-components/src/components/combobox/combobox.e2e.ts
```diff
@@ -2975,6 +2975,85 @@ describe("calcite-combobox", () => {
...
+            <calcite-combobox-item value="Daffodil" text-label="Daffodil"></calcite-combobox-item>
+            <calcite-combobox-item value="Nasturtium" text-label="Nasturtium"></calcite-combobox-item>
+          </calcite-combobox-item>
+        </calcite-combobox>`,
+      );
+      await page.waitForChanges();
+      combobox = await page.find("calcite-combobox");
+
+      await combobox.click();
+      expect(await page.find("calcite-combobox")).toHaveAttribute("open");
```
driskull, line 3003: Can we update this to check for properties instead of attributes in tests?

packages/calcite-components/src/components/combobox/combobox.scss
```diff
@@ -165,7 +165,7 @@
 }
 
 .label {
-  @apply pointer-events-none max-w-full flex-auto truncate p-0 font-normal;
```
driskull, line 168: Do we know why pointer events were disabled here?

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -695,8 +710,21 @@ export class Combobox
     if (this.ignoreSelectedEventsFlag) {
       return;
     }
-
     const target = event.target as HTMLCalciteComboboxItemElement["el"];
+
+    const isSelectAllTarget = event
+      .composedPath()
+      .some((node) => (node as HTMLElement).id === `${this.guid}-select-all-enabled`);
```
driskull, line 717: Instead of matching id properties can you store the select all enabled as a reference element to match this?

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -752,6 +780,16 @@ export class Combobox
     );
   }
 
+  private handleSelectAllToggle() {
+    const selectAllComboboxItemIsSelected = this.items.find(
+      (item) => item.id === `${this.guid}-select-all-enabled` && item.selected,
```
driskull, line 785: Same here. Can we match elements instead of ids?

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1033,6 +1079,15 @@ export class Combobox
       largestSelectedIndicatorChipWidth,
     });
 
+    if (this.isSelectAllOptionChecked) {
+      this.selectedItems.forEach((item) => {
+        const chipEl = this.referenceEl.querySelector<Chip["el"]>(`#${chipUidPrefix}${item.guid}`);
```
driskull, line 1084: Can we get reference to this chip el without querying the DOM?

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1684,6 +1769,20 @@ export class Combobox
       <div ariaHidden="true" class={CSS.floatingUIContainer} ref={setFloatingEl}>
         <div class={classes} ref={setContainerEl}>
           <ul class={{ list: true, "list--hide": !open }}>
+            {this.selectAllEnabled &&
+              this.selectionMode !== "single" &&
+              this.selectionMode !== "single-persist" && (
+                <calcite-combobox-item
+                  ariaLabel="Select All"
```
driskull, line 1776: label should be in messages

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1659,7 +1728,23 @@ export class Combobox
...
+    const selectAllComboboxItem = this.selectAllEnabled &&
+      this.selectionMode !== "single" &&
+      this.selectionMode !== "single-persist" && (
+        <calcite-combobox-item
+          ariaLabel="Select All"
+          class={{ [CSS.selectAllCheckbox]: true }}
+          id={`${this.guid}-select-all-enabled`}
+          role="option"
+          tabIndex="-1"
+          text-label="Select All"
```
driskull, line 1740: label should be in messages. applies to other places

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1684,6 +1769,20 @@ export class Combobox
...
+              this.selectionMode !== "single-persist" && (
+                <calcite-combobox-item
+                  ariaLabel="Select All"
+                  class={{ [CSS.selectAllCheckbox]: true }}
+                  id={`${this.guid}-select-all-enabled`}
+                  ref={this.setSelectAllComboboxItemReferenceEl}
+                  role="option"
+                  tabIndex="-1"
+                  text-label="Select All"
+                  value="Select All"
```
driskull, line 1783: Does this need a value? if so, can it be more of a value instead of a label?

driskull (changes requested):
Looking good @Elijbet 👏

packages/calcite-components/src/components/combobox/combobox.tsx
eriklharper: > Can we get reference to this chip el without querying the DOM?

Don't think so since the chips are dynamically generated based on the combobox-items.  I think this applies to your other questions about matching elements instead of ids for the arrays of items.  We can certainly set a ref on the select all checkbox element, but not sure if it makes sense to do this for the elements.  The refs would have to be garbage collected if the light dom children are modified.

driskull: Yeah I just meant the elements that the component is creating, not user slotted ones.

packages/calcite-components/src/components/combobox/combobox.scss
Elijbet: I'm not sure, and I don't see it doing anything. We are able to select by clicking on label on dev, and that should be the behavior?

packages/calcite-components/src/components/combobox/combobox.tsx
Elijbet: I'll change the other 2, but leave this one alone then. Since it generates the chips for slotted items.

packages/calcite-components/src/components/combobox/combobox.tsx
Elijbet: I undid the removal of `select-all-enabled` in favor of having the `Select All` checkbox as a default for now. I'll wait for @jcfranco to chime in after the feedback left by @ashetland of this possibly being a disruption to existing behavior (and needing a way to disable that).

packages/calcite-components/src/components/combobox/combobox.tsx
eriklharper: > Yeah I just meant the elements that the component is creating, not user slotted ones.

Oh, definitely, I understand that part, I just wonder if its worth it to create dynamic refs to internally-rendered elements based on slotted content.  Just not sure its worth creating all those refs in memory when a DOM query suffices.  Do you now which approach is likely to be more performant?

packages/calcite-components/src/components/combobox/combobox.scss
eriklharper: This change doesn't affect `multiple` selection-mode, which is the subject of this PR, so my recommendation is to revert this change.  I tested it both with and without this change and it doesn't seem to affect it either way, but because the `span.label` this style applies to only renders in single selection-mode, its a good idea to leave this style as-is.

packages/calcite-components/src/components/combobox/combobox.tsx
eriklharper: It looks like combobox-item warns if a `value` isn't provided.  Probably could have the value be `select-all` to make it more value-like.

packages/calcite-components/src/components/combobox/combobox.tsx
driskull: It would be nice to refactor this to store a reference to these chip elements created by the component rather than setting a unique id on them. Then the internal shadow DOM doesn't need to be queried every time `refreshSelectionDisplay()`is called. (which happens on resize observer call and `calciteInternalComboboxItemChangeHandler`)

I'm pretty sure its more performant to just have a reference pointer to the element rather than having to do lookups through the dom every time these are called.

We can do this as a follow up issue since it is existing code. @Elijbet can you create a refactor issue?

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1400,6 +1481,13 @@ export class Combobox
     return this.getItems().length === this.getSelectedItems().length;
   }
 
+  private isAllOptionsSelectedExceptSelectAll(): boolean {
```
driskull, line 1484: Is this being called anywhere? i dont see where it is.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1684,6 +1811,21 @@ export class Combobox
...
+            {this.selectAllEnabled &&
+              this.selectionMode !== "single" &&
+              this.selectionMode !== "single-persist" && (
+                <calcite-combobox-item
+                  ariaLabel={this.messages.selectAll}
+                  class={CSS.selectAll}
+                  id={`${this.guid}-select-all-enabled`}
+                  ref={this.setSelectAllComboboxItemReferenceEl}
+                  role="option"
+                  selected={this.updateSelectAllState()}
```
driskull, line 1823: This function sounds like a function that returns void but is returning boolean here. Can we rename this to better describe what it is doing?

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1412,12 +1500,35 @@ export class Combobox
     this.textInput.value?.focus();
   }
 
+  private updateSelectAllState(): boolean {
+    if (!this.selectAllComboboxItemReferenceEl) {
+      return;
```
driskull, line 1505: should this return false here?

packages/calcite-components/src/components/combobox/combobox.tsx
driskull: Should this just be set to `this.selectAllComboboxItemReferenceEl.selected`?

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1412,12 +1500,35 @@ export class Combobox
     this.textInput.value?.focus();
   }
 
+  private updateSelectAllState(): boolean {
```
driskull, line 1503: I think this function should be reworked into two functions. One that updates the selected all state and one that gets the value of it.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1616,8 +1727,8 @@ export class Combobox
         {showLabel && (
           <span
             class={{
-              label: true,
-              "label--icon": !!selectedItem?.icon,
+              [CSS.label]: true,
```
driskull, line 1730: nice!

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1659,7 +1770,23 @@ export class Combobox
   }
 
   private renderListBoxOptions(): JsxNode {
-    return this.filteredItems.map((item) => (
+    const selectAllComboboxItem = this.selectAllEnabled &&
+      this.selectionMode !== "single" &&
+      this.selectionMode !== "single-persist" && (
+        <calcite-combobox-item
+          ariaLabel={this.messages.selectAll}
```
driskull, line 1777: This should set the component's label property instead of ariaLabel

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1659,7 +1770,23 @@ export class Combobox
...
+          class={CSS.selectAll}
+          id={`${this.guid}-select-all-enabled`}
+          role="option"
```
driskull, line 1780: Why is the role being set here? Shouldn't the component be setting the role? It seems like this would be the responsibility of the component itself.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1684,6 +1811,21 @@ export class Combobox
...
           <ul class={{ list: true, "list--hide": !open }}>
+            {this.selectAllEnabled &&
+              this.selectionMode !== "single" &&
+              this.selectionMode !== "single-persist" && (
+                <calcite-combobox-item
+                  ariaLabel={this.messages.selectAll}
+                  class={CSS.selectAll}
+                  id={`${this.guid}-select-all-enabled`}
+                  ref={this.setSelectAllComboboxItemReferenceEl}
+                  role="option"
```
driskull, line 1822: Why is the role being set here? Shouldn't the component be setting the role? It seems like this would be the responsibility of the component itself.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1659,7 +1770,23 @@ export class Combobox
...
+      this.selectionMode !== "single" &&
+      this.selectionMode !== "single-persist" && (
+        <calcite-combobox-item
+          ariaLabel={this.messages.selectAll}
+          class={CSS.selectAll}
+          id={`${this.guid}-select-all-enabled`}
+          role="option"
+          tabIndex="-1"
+          text-label={this.messages.selectAll}
+          value="Select All"
```
driskull, line 1783: value should be a value like `select-all` or be removed if not being used.

driskull (changes requested):
Looking good @Elijbet ! Ill test it out as well but its getting there!

packages/calcite-components/src/components/combobox/combobox.tsx
Elijbet: [Refactor dynamically created chips to use internal references instead of DOM queries](https://github.com/Esri/calcite-design-system/issues/12052#top)
#12052

packages/calcite-components/src/components/combobox/combobox.scss
```diff
@@ -236,6 +236,16 @@ calcite-chip {
   @apply block;
 }
 
+.select-all {
+  position: sticky;
+  inset-block-start: 0;
+  background-color: var(--calcite-combobox-item-background-color-active, var(--calcite-color-foreground-1));
+  z-index: 1;
```
driskull, line 243: This z-index should use the calcite token for `var(--calcite-z-index-sticky)`

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -281,6 +283,8 @@ export class Combobox
 
   @state() selectedVisibleChipsCount = 0;
 
+  @state() selectAll: boolean = false;
```
driskull, line 286: Is there an e2e test for when all are selected upon initialization and the allSelected checkbox is checked?

driskull: Can we rename this to `allSelected`? As it is it sounds like a method and not a state.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1412,13 +1516,40 @@ export class Combobox
     this.textInput.value?.focus();
   }
 
+  private updateAndGetSelectAllState(): boolean {
```
driskull, line 1519: This method should be split up. It shouldn't be updating and returning state. The state part could just be a stateful property and the method should just update. As it is, it has two separate intentions.

`this.selectAllComboboxItemReferenceEl` should be setup to store the state of its selected value when it emits the selected event.

example:
```
<calcite-combobox-item
                  class={CSS.selectAll}
                  id={`${this.guid}-select-all-enabled`}
                  label={this.messages.selectAll}
                  ref={this.setSelectAllComboboxItemReferenceEl}
                  selected={this.updateAndGetSelectAllState()} 
                 onCalciteComboboxItemChange={} // todo: function to store state
                  tabIndex="-1"
                  text-label={this.messages.selectAll}
                  value="select-all"
                />
```

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -752,6 +782,20 @@ export class Combobox
...
   }
 
+  private toggleSelectAll() {
+    const selectAllComboboxItemIsSelected = this.items.find(
+      (item) => item === this.selectAllComboboxItemReferenceEl,
+    );
+
+    this.selectAll = !!selectAllComboboxItemIsSelected?.selected;
+
+    this.items.forEach((item) => {
```
driskull, line 792: nitpick:

```
this.items.forEach((item) => (item.selected = this.selectAll));
```

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1684,6 +1830,20 @@ export class Combobox
...
           <ul class={{ list: true, "list--hide": !open }}>
+            {this.selectAllEnabled &&
+              this.selectionMode !== "single" &&
+              this.selectionMode !== "single-persist" && (
+                <calcite-combobox-item
+                  class={CSS.selectAll}
+                  id={`${this.guid}-select-all-enabled`}
+                  label={this.messages.selectAll}
+                  ref={this.setSelectAllComboboxItemReferenceEl}
+                  selected={this.updateAndGetSelectAllState()}
```
driskull, line 1841: todo: fix to use a state property

packages/calcite-components/src/components/combobox/combobox.scss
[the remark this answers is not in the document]
Elijbet: We are, sorry I missed that. I think I accidentally reverted that fix.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -695,8 +702,31 @@ export class Combobox
...
+      .some((node) => (node as HTMLElement) === this.selectAllComboboxItemReferenceEl);
+
+    if (this.selectAllEnabled) {
+      if (isSelectAllTarget) {
+        this.toggleSelectAll();
+      }
+      this.updateAndGetSelectAllState();
+
+      // TODO: Doesn't hide the last element
+      if (this.isAllSelected()) {
```
driskull, line 718: Can we open up a refactor issue for combobox to use more state for items and whether they are all selected? Ideally, there should just be a stateful value for whether all items are selected instead of having to call a method.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -695,8 +702,31 @@ export class Combobox
...
+        this.selectedItems.forEach((item) => {
+          const chipEl = this.referenceEl.querySelector<Chip["el"]>(
+            `#${chipUidPrefix}${item.guid}`,
+          );
+          if (chipEl) {
+            this.hideChip(chipEl);
```
driskull, line 724: Why are chips being hidden instead of just not being rendered?

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -858,6 +902,23 @@ export class Combobox
           const item = this.filteredItems[this.activeItemIndex];
           this.toggleSelection(item, !item.selected);
           event.preventDefault();
+          if (this.selectAllEnabled) {
+            if (item === this.selectAllComboboxItemReferenceEl) {
+              this.toggleSelectAll();
+            }
+            this.updateAndGetSelectAllState();
+            // TODO: doesn't hide the last element
```
driskull, line 910: remove comment

packages/calcite-components/src/components/combobox/combobox.tsx
driskull: There should be one source of truth for whether all items are selected. The source of truth should be if all the items are actually selected, not whether the select all item is checked. That should just reflect the state of all items being selected.

driskull: Ideally, we would have something like this:

```tsx
@state()
private items: HTMLCalciteComboboxItemElement["el"][] = [];

@state()
private selectedItems: HTMLCalciteComboboxItemElement["el"][] = [];

@state()
get itemLength = (): number => {
  return this.items.length;
}

@state()
get selectedItemLength = (): number => {
  return this.selectedItems.length
}

@state()
get isAllSelected = (): boolean => {
  return this.selectedItemLength === this.itemLength;
}
```

driskull: Then we can just use those stateful props to render accordingly. Right now, there's too many calls to `isAllSelected()` which won't trigger a re-render if selected items change.

packages/calcite-components/src/components/combobox/combobox.tsx
Elijbet: I'll start a discussion on why that was done before and create a refactoring issue with that in mind.

packages/calcite-components/src/components/combobox/combobox.tsx
Elijbet: [Refactor calcite-combobox to Use Stateful Properties](https://github.com/Esri/calcite-design-system/issues/12061#top)#12061

packages/calcite-components/src/components/combobox/combobox.tsx
Elijbet: > Is there an e2e test for when all are selected upon initialization and the allSelected checkbox is checked?

Yep. https://github.com/Esri/calcite-design-system/pull/11721/files#diff-85da4d8c5247e61a3626fd500c5027b510000a530ba6972f87e27a2acf71a033R3137-R3149

Elijbet: > Can we rename this to `allSelected`? As it is it sounds like a method and not a state.

I called it `selectAll` because it refers to `selectAllEnabled` combobox item element in the code. `allSelected` would mean that checkbox is checked.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -855,6 +900,21 @@ export class Combobox
           const item = this.filteredItems[this.activeItemIndex];
           this.toggleSelection(item, !item.selected);
           event.preventDefault();
+          if (this.selectAllEnabled) {
+            if (item === this.selectAllComboboxItemReferenceEl) {
```
driskull, line 904: This logic seems the same as above. Can it share the same private method? See line 717.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1030,6 +1090,27 @@ export class Combobox
...
+      this.selectedItems.forEach((item) => {
+        const chipEl = this.referenceEl.querySelector<Chip["el"]>(`#${chipUidPrefix}${item.guid}`);
+        if (chipEl) {
+          this.hideChip(chipEl);
+        }
+      });
+    }
+
+    if (
+      this.selectAllComboboxItemReferenceEl &&
```
driskull, line 1103: Could this if statement just use the `this.indeterminate` stateful private property instead of using the element?

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1656,17 +1764,33 @@ export class Combobox
...
+          id={`${this.guid}-select-all-enabled`}
+          label={this.messages.selectAll}
+          tabIndex="-1"
+          value="select-all"
+        />
+      );
+
+    const selectAllOptionAndFilteredItemsList = [selectAllComboboxItem, ...this.filteredItems];
+
+    return selectAllOptionAndFilteredItemsList.map((item: any) => {
```
driskull, line 1780: Can you avoid typing this as any?

driskull (changes requested):

packages/calcite-components/src/components/combobox/resources.ts
```diff
@@ -3,15 +3,18 @@ export const ComboboxItemGroupSelector = "CALCITE-COMBOBOX-ITEM-GROUP";
...
-  listContainer: "list-container",
   icon: "icon",
+  input: "input",
+  inputHidden: "input--hidden",
+  label: "label",
+  labelIcon: "label-icon",
+  listContainer: "list-container",
   placeholderIcon: "placeholder-icon",
+  selectAll: "select-all",
+  selectionDisplayFit: "selection-display-fit",
```
driskull, line 16: nitpick: `selection-display--single` and `selection-display--fit` if we're following BEM guidelines.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -265,7 +267,22 @@ export class Combobox
...
+  }
+
+  @state()
+  get indeterminate(): boolean {
+    return this.selectedItems.length > 0 && this.selectedItems.length !== this.items.length;
+  }
+
+  @state()
+  get visibleKeyboardNavItems(): HTMLCalciteComboboxItemElement["el"][] {
+    return this.keyboardNavItems().filter((item) => !isHidden(item));
```
driskull, line 282: Can you move the `keyboardNavItems()` into a stateful getter? Otherwise this state won't change because this is a function, not another stateful property that would cause `visibleKeyboardNavItems` to update.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1288,9 +1368,21 @@ export class Combobox
     const items: HTMLCalciteComboboxItemElement["el"][] = Array.from(
       this.el.querySelectorAll(ComboboxItemSelector),
     );
+
     return items.filter((item) => withDisabled || !item.disabled);
   }
 
+  private keyboardNavItems(): HTMLCalciteComboboxItemElement["el"][] {
```
driskull, line 1375: This could be a stateful property getter if `this.selectedAllComboboxItemReferenceEl` and `items` were also stateful properties.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1367,6 +1459,17 @@ export class Combobox
...
+
+    const stickyElement = this.selectAllComboboxItemReferenceEl;
+    const stickyHeight = stickyElement?.offsetHeight || 0;
+
+    const listContainer = this.listContainerEl;
+    const itemRect = item.getBoundingClientRect();
+    const containerRect = listContainer.getBoundingClientRect();
+
+    if (itemRect.top < containerRect.top + stickyHeight) {
+      listContainer.scrollTop -= containerRect.top + stickyHeight - itemRect.top;
```
driskull, line 1471: This is good for now. Maybe a follow up refactor issue to adjust the scrollable area. The reasoning for not having this element sticky is because usually sticky elements start at the center or another position and then get stuck to to the top when scrolling. However, in this design, the select all is always at the top and never moves. Is that correct? If so, it seems like the select all element should just be above of the scrollable area.

@jcfranco what do you think? Tackle this now or later.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1681,6 +1806,21 @@ export class Combobox
       <div ariaHidden="true" class={CSS.floatingUIContainer} ref={setFloatingEl}>
         <div class={classes} ref={setContainerEl}>
           <ul class={{ list: true, "list--hide": !open }}>
+            {this.selectAllEnabled &&
+              this.selectionMode !== "single" &&
+              this.selectionMode !== "single-persist" && (
+                <calcite-combobox-item
+                  class={CSS.selectAll}
+                  id={`${this.guid}-select-all-enabled`}
```
driskull, line 1814: This id is shared with line 1772. Should they be unique?

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1656,17 +1765,33 @@ export class Combobox
...
-        role="option"
-        tabIndex="-1"
-      >
-        {item.heading || item.textLabel}
-      </li>
-    ));
+    const selectAllComboboxItem = this.selectAllEnabled &&
+      this.selectionMode !== "single" &&
+      this.selectionMode !== "single-persist" && (
+        <calcite-combobox-item
```
driskull, line 1771: Shouldn't this be a native `li` element? This is just for a11y purposes right? Do we even need to have the select all option here? It seems like we don't need to have this here.

driskull (changes requested):

packages/calcite-components/src/components/combobox/combobox.tsx
Elijbet: I think the hidden accessibility list should mirror the visible dropdown, I'd expect that when navigating the combobox with arrow keys, the screen reader needs to announce the "Select All" option just like any other item.

Does this sound right, @geospatialem?

But yeah it should be an `<li>` element.

packages/calcite-components/src/components/combobox/combobox.tsx
Elijbet: That is correct. I'm not sure how this would get refactored to be added separately as static element to the top of the list. Right now is just another item in the list so that we can leverage the existing navigation logic.

Elijbet: @SkyeSeitz @ashetland Sounds more like an implementation question, but do you have an opinion on this?

Elijbet: Here is my input:
**PROS**
* The "Select All" item is treated like any other list item, so it benefits from the existing keyboard navigation, focus management, styling (it needs to look just like another item), and accessibility logic.
* No need to duplicate or create separate navigation logic for the "Select All" element.
**CON**
If devs expect that Sticky elements typically start at a position (e.g., center or below the top) and then "stick" to the top when scrolling, then just from the implementation perspective, maybe this is a hacky solution? Not sure if this matters, if it’s achieving what we want to achieve while leveraging our custom logic, which is quite extensive. This seems more important than making sure we don’t apply sticky CSS to something that is always static.

SkyeSeitz: Aaron may have external examples of this from design research, but I don't have any issues with the Select all element existing at the top and remaining sticky - as implemented. I would expect to have access to Select all no matter where I am in the scrollable list.

ashetland: The Dell Design System has a [select-all](https://vanilla.delldesignsystem.com/2.46.2/iframe.html?viewMode=docs&id=components-dropdown--basic#multi-select-list), but it scrolls 😭. For us, as long as it stays at the top, I'm not sure I have much opinion on the final implementation. Question though, as we work on #6466, I could see it also making sense to make group headings in Combobox sticky too — would sticky group headings make us reconsider how we implement this "sticky" select-all button?

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1230,7 +1322,7 @@ export class Combobox
     this.updateItemProps();
 
     this.selectedItems = this.getSelectedItems();
-    this.filteredItems = this.getFilteredItems();
+    this.filteredItems = this.visibleKeyboardNavItems;
```
driskull, line 1325: If `filteredItems` and `visibleItems` are the same, can we just use `filteredItems`? Make `filteredItems` a getter with the `visibleKeyboardNavItems` logic?

```ts
@property() get filteredItems():  HTMLCalciteComboboxItemElement["el"][]  {
  return this.keyboardNavItems.filter((item) => !isHidden(item));
}
```

packages/calcite-components/src/components/combobox/combobox.tsx
driskull: Can you type this correctly instead of `any`? Maybe just map the filteredItems first and then call `unshift` on the array to add the `selectAll` item to the top of it?

driskull (commented):
Almost there!!! 🏁

packages/calcite-components/src/components/combobox/combobox.tsx
driskull: Something like this:

```tsx
return [
  selectAllComboboxItem,
  ...this.filteredItems.map((item) => (
    <li
      ariaLabel={item.label}
      ariaSelected={item.selected}
      class={CSS.selectAll}
      id={item.guid ? `${itemUidPrefix}${item.guid}` : null}
      role="option"
      tabIndex="-1"
    >
      {item.heading || item.textLabel}
    </li>
  )),
];
```

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -364,6 +393,9 @@ export class Combobox
   /** Specifies the size of the component. */
   @property({ reflect: true }) scale: Scale = "m";
 
+  /** When `true`, provides a toggle for selecting all items. Does not apply to `selection-mode single`. */
```
geospatialem, line 396: doc consistency recommendation 📚:

```suggestion
  /** When `true` and `selectionMode` is `"multiple"` or `"ancestors"`, provides a checkbox for selecting all `calcite-combobox-item`s. */
```

packages/calcite-components/src/components/combobox/combobox.tsx
geospatialem: This sounds great with JAWS and provides context. This seems like good context to users with the functionality via the `<li>` element, no additional change needed. 👍🏻

packages/calcite-components/src/components/combobox-item/combobox-item.tsx
```diff
@@ -158,6 +158,13 @@ export class ComboboxItem extends LitElement implements InteractiveComponent {
    *  */
   @property({ reflect: true }) itemHidden = false;
 
+  /**
+   * In multi selection mode, show as indeterminate when only some children are selected.
```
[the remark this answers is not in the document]
Elijbet: I think this should probably say `and one or more, but not all`

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -686,14 +713,37 @@ export class Combobox
...
   ): void {
     if (this.ignoreSelectedEventsFlag) {
       return;
     }
-
     const target = event.target as HTMLCalciteComboboxItemElement["el"];
+
+    const isSelectAllTarget = event
+      .composedPath()
+      .some((node) => (node as HTMLElement) === this.selectAllComboboxItemReferenceEl);
```
driskull, line 741: nitpick: I think elsewhere in the codebase we use `.composedPath().includes(this.selectAllComboboxItemReferenceEl)` which I think is more performant.

packages/calcite-components/src/components/combobox/combobox.tsx
driskull: nitpick: Ill defer to @jcfranco for if we want to move this out of a scrollable div now or later. Ideally, it should be moved out because it doesn't need to be applied as sticky since it never scrolls at all.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1415,7 +1515,16 @@ export class Combobox
...
   private renderChips(): JsxNode {
     const { activeChipIndex, readOnly, scale, selectionMode, messages } = this;
-    return this.selectedItems.map((item, i) => {
+
+    if (this.selectAllEnabled && this.allSelected) {
+      return null;
+    }
+
+    const filteredSelectedItems = this.selectedItems.filter(
+      (item) => item !== this.selectAllComboboxItemReferenceEl,
```
driskull, line 1524: fixme: Is the `selectAllComboboxItemReferenceEl` included in `combobox.selectedItems`? I'm not sure a user would expect that to be included in a public property since its not a "real" item.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1450,49 +1559,27 @@ export class Combobox
...
-        ref={setAllSelectedIndicatorChipEl}
-        scale={scale}
-        title={label}
-        value=""
-      >
-        {label}
-      </calcite-chip>
-    );
-  }
+    const label = compact ? this.messages.all || "All" : this.messages.allSelected;
```
driskull, line 1569: fixme: I don't think we need the `"All"` fallback here. We don't do text fallbacks in other components.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1610,11 +1697,12 @@ export class Combobox
           "input-wrap--single": single,
         }}
       >
+        {console.log("selectedItem", selectedItem)}
```
driskull, line 1700: fixme: remove console log

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1656,17 +1744,34 @@ export class Combobox
...
-    ));
+    const selectAllComboboxItem = this.selectAllEnabled &&
+      this.selectionMode !== "single" &&
+      this.selectionMode !== "single-persist" && (
+        <li
+          ariaLabel={this.messages.selectAll}
+          ariaSelected={this.allSelected}
+          id={`${this.guid}-select-all-enabled-screen-reader`}
+          tabIndex="-1"
+          value="select-all"
```
driskull, line 1755: fixme: I don't think `value` is a valid property of a list item.

driskull: fixme: I think this needs a ` role="option"` just like the other list item below.

driskull: The other list item below also has text inside the list item. Should this have text inside? Maybe the ariaLabel isn't necessary if it has text inside?

driskull (changes requested):
So close

packages/calcite-components/src/components/combobox/combobox.tsx
Elijbet: I think when we separated `this.keyboardNavItems`, we included `selectAllComboboxItemReferenceEl` in that, but removed it from `this.items`.
 
`getSelectedItems` is using `this.items`

driskull: Ok so this isn't necessary then?

driskull (approved):
👍 LGTM! 🎉 

Lets let @jcfranco or another dev review as well.

eriklharper (changes requested):
Had one question, and just a few refactor changes that should be easy to do.  The main thing I was wondering was why the select all chip doesn't show when all the items are selected?

eriklharper (approved):
I'll go ahead and mark this approved and if need be we can follow up with those small changes I requested.  Thanks and great work on this @Elijbet !

Elijbet:
> Had one question, and just a few refactor changes that should be easy to do. The main thing I was wondering was why the select all chip doesn't show when all the items are selected?

This is working on my end.

driskull:
> This is working on my end. Maybe @driskull can test also

I don't see a chip when all items are selected either.

image

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1450,49 +1560,27 @@ export class Combobox
...
-        ref={setAllSelectedIndicatorChipEl}
-        scale={scale}
-        title={label}
-        value=""
-      >
-        {label}
-      </calcite-chip>
-    );
-  }
+    const label = compact ? this.messages.all : this.messages.allSelected;
```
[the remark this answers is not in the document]
Elijbet: Not removing, just consolidating `renderAllSelectedIndicatorChip` and `renderAllSelectedIndicatorChipCompact` into one function that's called with bool argument `compact`.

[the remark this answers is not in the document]
eriklharper: I see that now, sorry, read the diff wrong!  Its all good, and I see it locally as well.  Nice!

Elijbet:
> > This is working on my end. Maybe @driskull can test also
> 
> I don't see a chip when all items are selected either.
> 
> image

It was showing on my end because I didn't have the placeholder set. I made some changes to show it.
`newCompactBreakpoint` is calculated with `getTextWidth`, which accounts for the placeholder.

packages/calcite-components/src/components/combobox/combobox.tsx
```diff
@@ -1656,21 +1739,40 @@ export class Combobox
...
+      );
+
+    const selectAllOptionAndFilteredItemsList = [selectAllComboboxItem, ...this.filteredItems];
+
+    return selectAllOptionAndFilteredItemsList.map((item: HTMLCalciteComboboxItemElement["el"]) => {
+      return (
+        <li
+          ariaLabel={item.label}
+          ariaSelected={item.selected}
+          class={CSS.selectAll}
```
driskull, line 1763: Why are we setting this class? This isn't the selectAll element right?

ashetland:
LGTM

packages/calcite-components/src/components/combobox/combobox.tsx
jcfranco: Based on the pros that @Elijbet shared, I think it's fine to keep as is. There might be some cases where the z-index could cause some issues, but we can wait until we get feedback. 

> Ideally, it should be moved out because it doesn't need to be applied as sticky since it never scrolls at all.

Wouldn't it scroll when `maxItems` is set (see [example](https://codepen.io/jcfranco/pen/zxxyRPR?editors=100))?

jcfranco (commented):
Awesome job, @Elijbet! Had a few follow-up review comments, but this looks great!

packages/calcite-components/src/components/combobox/combobox.tsx
driskull: No, because the select all element is always first and never scrolls. It doesn't need to be sticky because of that. It should really be above and outside the scrolling area.

driskull: The problem with having it sticky is that other items can be behind it yet appear focused but they are not seen

driskull: > but needs special handling by the component for keyboard navigation.

I could be wrong but I don't think it would. The component has an array of the items and just focuses the next one in the list and scrolls it into view. I guess the only special handling would be the scrolling wouldn't be necessary.

> Wouldn't this also be the case if we move it outside + above the scroller?

I don't think so. The current workaround to offset the scrollTop is needed because we don't want the focused item to be behind the sticky select all item. That's why its adjusting the `scrollTop` instead of just calling `scrollIntoView()`. This wouldn't be necessary if the select all item wasn't in the scrollable area.

packages/calcite-components/src/components/combobox/combobox.e2e.ts
```diff
@@ -2975,6 +2981,227 @@ describe("calcite-combobox", () => {
...
+
+      await toggleAction([listItem, combobox]);
+
+      expect(await selectAll.getProperty("indeterminate")).toBe(true);
+      expect(await page.find(`calcite-combobox >>> calcite-chip[value=Sequoia]`)).toBeDefined();
+    }
+
+    it("should toggle indeterminate state to `All Selected` when list items are toggled with a click", async () => {
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      await testToggleListItems(page, async ([listItem, _combobox]) => {
```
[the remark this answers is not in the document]
Elijbet: I'm passing both because I'm using either/or, right? I have 2 it blocks that call testToggleListItems, to test the click I need the listItem, to test the Enter key, I need the combobox.

PR #11723, opened 2025-03-10
feat(fab): add component tokens

Elijbet:
**Related Issue:** #7180 

Adds the following tokens for `fab` component.

`--calcite-fab-background-color`: Specifies the component's background color.
`--calcite-fab-shadow`: Specifies the component's shadow.

### DIFF
--- a/packages/calcite-components/src/components/fab/fab.scss
+++ b/packages/calcite-components/src/components/fab/fab.scss
@@ -16,13 +16,7 @@
 @include disabled();
 
 calcite-button {
-  --tw-shadow: var(--calcite-fab-shadow, 0 6px 20px -4px rgba(0, 0, 0, 0.1), 0 4px 12px -2px rgba(0, 0, 0, 0.08));
-  --tw-shadow-colored: var(
-    --calcite-fab-shadow,
-    0 6px 20px -4px var(--tw-shadow-color),
-    0 4px 12px -2px var(--tw-shadow-color)
-  );
-  box-shadow: var(--tw-ring-offset-shadow, 0 0 #0000), var(--tw-ring-shadow, 0 0 #0000), var(--tw-shadow);
+  box-shadow: var(--calcite-fab-shadow, 0 6px 20px -4px rgba(0, 0, 0, 0.1), 0 4px 12px -2px rgba(0, 0, 0, 0.08));
   &:hover {
     @apply shadow-2-lg;
   }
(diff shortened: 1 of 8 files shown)


driskull (commented):
This code looks goood, what about the shadow being set on calcite-button?

packages/calcite-components/src/components/fab/fab.scss
```diff
@@ -1,16 +1,47 @@
...
 calcite-button {
-  @apply shadow-2;
+  --tw-shadow: var(--calcite-fab-shadow, 0 6px 20px -4px rgba(0, 0, 0, 0.1), 0 4px 12px -2px rgba(0, 0, 0, 0.08));
+  --tw-shadow-colored: var(
+    --calcite-fab-shadow,
+    0 6px 20px -4px var(--tw-shadow-color),
+    0 4px 12px -2px var(--tw-shadow-color)
+  );
+  box-shadow: var(--tw-ring-offset-shadow, 0 0 #0000), var(--tw-ring-shadow, 0 0 #0000), var(--tw-shadow);
   &:hover {
```
Elijbet, line 28: I’m trying to get rid of the tw values to set a shadow token.
 
What happens when `--tw-ring-offset-shadow` goes away here:
`box-shadow: var(--tw-ring-offset-shadow, 0 0 #0000), var(--tw-ring-shadow, 0 0 #0000), var(--tw-shadow);`
 
I could sub the `--tw-shadow` with: `var(--calcite-fab-shadow, 0 6px 20px -4px rgba(0, 0, 0, 0.1), 0 4px 12px -2px rgba(0, 0, 0, 0.08));`
Not sure what to do with the ring values.

@ashetland @SkyeSeitz @macandcheese

Elijbet: It seems related but may be out of scope for this issue. And it doesn't really answer what to use for ring values. These are the first two values on box shadow and are interpreted as <offset-x> and <offset-y>.

ashetland: Oh gotcha. Your sub values look correct to me and match the medium shadow we have spec'ed for FAB in #10050 so we wouldn't have any visual changes when that one gets picked up.

packages/calcite-components/src/components/fab/fab.scss
```diff
@@ -1,11 +1,22 @@
...
+  @apply flex;
+
+  background-color: var(--calcite-fab-background-color, transparent);
 }
 
 @include disabled();
 
 calcite-button {
-  @apply shadow-2;
+  box-shadow: var(--calcite-fab-shadow, 0 6px 20px -4px rgba(0, 0, 0, 0.1), 0 4px 12px -2px rgba(0, 0, 0, 0.08));
```
macandcheese, line 19: I think you'd need to set this as an internal prop and accept an override - if a user were to override this, the other two hardcoded values would still appear, correct?

macandcheese (approved):

PR #11739, opened 2025-03-12
feat: add language-2

arowles:
**Related Issue:** #

## Summary
**added**

language-2-16
language-2-24
language-2-32


jcfranco (changes requested):
It looks like this has unrelated changes to the root `package-lock.json`. Could you revert these changes?

arowles:
> It looks like this has unrelated changes to the root `package-lock.json`. Could you revert these changes?

Ok, I reverted those changes and re-launched the reference page command, but I saw that it still tried to make changes to the package-lock.json file, so I deleted that change from the commit and committed the rest.

jcfranco:
@arowles Awesome, thanks!

jcfranco (approved):

Issue #11514, opened 2025-02-11
`TextArea.getHeightAndWidthOfElements` sometimes throws `TypeError`


nwhittaker:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Removing a `<calcite-text-area>` before it's done loading can cause its `getHeightAndWidthOfElements` method to throw a `TypeError`.

### Expected Behavior

Removing a `<calcite-text-area>` before it's done loading does not produce a `TypeError`.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/PwooKVN

### Reproduction Steps

1. Visit the repro and open devtools
2. Double click the _Toggle block_ switch. If done quickly enough, the `TypeError` should appear in the console.
3. To retry step 2, be sure to refresh the page first to flush any component cache.

### Reproduction Version

3.0.0

#### Stack trace
```
TypeError: Cannot read properties of undefined (reading 'getBoundingClientRect')
    at x.getHeightAndWidthOfElements (XBZLGRRI.js:4:9752)
    at x.setTextAreaHeight (XBZLGRRI.js:4:9599)
    at x.updated (XBZLGRRI.js:4:8459)
    at x._$AE (KVSCG5E3.js:6:11661)
    at x.performUpdate (KVSCG5E3.js:6:11525)
    at x._load (KVSCG5E3.js:6:21372)
```

### Priority impact

impact - p2 - want for an upcoming milestone

### Impact

In our list/detail layouts, the end-user can generate this error by double-clicking a list item. Our automated tests are also impacted as they're more likely to discard elements before the elements are done loading/hydrating.

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Field Apps

DitwanP:
Note for milestone planning, we should tackle this one soon.

jcfranco:
Quick update: this might be related to a Lumina bug. I’ll submit a workaround PR while that’s being looked into, in case it doesn’t make it into this release.”

jcfranco:
[...]

PR #11751, opened 2025-03-14
fix(text-area): fix error caused by internal measuring on disconnect

jcfranco:
**Related Issue:** #11514

## Summary

Updates internal measuring logic to check for references before calling `Element.getBoundingClientRect()` since `ref`s can be cleared when a ref is updated and the element is disconnected.


aPreciado88 (approved):
🚀🐱‍👤

PR #11070, opened 2024-12-16
feat(accordion-item): update token naming schema

alisonailea:
**Related Issue:** #10894 

## Summary

This PR adds two new tokens which align with the Calcite naming schema.

--calcite-accordion-item-icon-color-start
--calcite-accordion-item-icon-color-end

BEGIN_COMMIT_OVERRIDE
deprecate(accordion-item): deprecate `--calcite-accordion-item-start-icon-color` and `--calcite-accordion-item-end-icon-color` tokens
END_COMMIT_OVERRIDE

Issue #10052, opened 2024-08-13
[tailwind] Update tailwind-preset to reference semantic tokens


alisonailea:
**monday.com sync:** #18048571520

### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

The [tailwind-preset index.ts file](https://github.com/Esri/calcite-design-system/blob/dev/packages/tailwind-preset/src/index.ts) contains variables defined with raw values. These should be defined using semantic tokens to ensure consistency and to facilitate deeper customization needs (example: #11555)

### Acceptance Criteria

Within the index.ts file:
- [x] `fontSize` variables are defined using semantic `font-size` and `font-line-height` tokens (PR [12842](https://github.com/Esri/calcite-design-system/pull/12842))
- [x] `spacing` variables are defined using semantic `space` tokens where possible and deprecated in favor of semantic tokens (PR [13770](https://github.com/Esri/calcite-design-system/pull/13770))

### Relevant Info

This cannot be completed until the below issues are completed and all required REM-based semantic tokens are available
- [x] #10051 (partially)
- [x] #10014

### Which Component

N/A

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [ ] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

DintaMel:
Verified with calcite-components 5.1.0-next.27  and calcite-tailwind-preset 1.1.1-next.7 using [Stackblitz](https://stackblitz.com/edit/github-rxkzwf7x-djswzpvw?file=package-lock.json).

Issue #8856, opened 2024-02-29
Research: determine the potential impact of applying REM-based tokens to all components


alisonailea:
**monday.com sync:** #18048397906

### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Description

REM-based tokens provide levels of flexibility that may be preferable for users. However, due to the base value being tied to the global `font-size` CSS prop, this will likely have ripple effects across components. Thorough testing needed.

#### Dependant on:
- [x] Add semantic space (rem) tokens: #10012
- [x] Add semantic size (rem) tokens: #10014
- [x] Add semantic font-size-relative (rem) tokens: #13037
- [x] Add REM-based design tokens: #10051
- [ ] Update tailwind-preset to reference semantic tokens: #10052

### Acceptance Criteria

Test the impact using REM-based variables on specific elements within Calcite Components has:
- [ ] `size` tokens for height, width, inline-size, block-size, and scale
- [ ] `space` tokens for padding, margin, gap, etc.
- [ ] `font-size-relative` and `font-line-height` tokens for heading text and body text

### Relevant Info

#13657: Research: if a custom font-size variable could modify text elements (specifically) across components

Tailwind has previously covered this with a selection of "applied" tailwind classes. In many components, these have been replaced with direct "space" or "line-height" properties. So both will need to be searched for.

Style Dictionary has built-in transforms like `size/pxToRem` that could help with variable conversions for testing purposes and potentially for any large-scale token value updates ([details](https://styledictionary.com/reference/hooks/transforms/predefined/#sizepxtorem))

### Which Component

N/A

### Example Use Case

```
.scale-m {
  @apply text-n1h;
  padding: var(--calcite-space-sm);
  max-width: var(--calcite-size-2xl);
}
```

### Priority impact

p3 - want for upcoming milestone

### Calcite package

- [X] @esri/calcite-components
- [X] @esri/calcite-design-tokens

### Esri team

Calcite (dev)

### Original description

>### Description
>
>#### Pre-November
>- [x] Review space and size tokens from design. Note changes to values in existing tokens
>- ~~[ ] create token transformer script to convert semantic space, size, and font-size tokens to REMs. (#10051)~~
>
>#### November release
>
>##### Calcite Design Tokens
>- [x] add and update tokens to semantic design tokens package.
>    - [x] Size (#10014)
>    - [x] Space (#10012)
>
>##### Calcite Components
[...]

PR #11759, opened 2025-03-17
feat(input-message): add component tokens and deprecate --calcite-input-message-spacing-value

Elijbet:
**Related Issue:** #7180 

Adds the following tokens to `input-message` component:

`--calcite-input-message-icon-color`: Specifies the component's icon color.
`--calcite-input-message-spacing`: Specifies the top margin space above the component.

Deprecates `--calcite-input-message-spacing-value`: Specifies the top margin space above the component.

### DIFF
--- a/packages/calcite-components/src/components/input-message/input-message.e2e.ts
+++ b/packages/calcite-components/src/components/input-message/input-message.e2e.ts
@@ -155,5 +155,13 @@ describe("calcite-input-message", () => {
         },
       });
     });
+
+    describe("deprecated", () => {
+      themed(html`<calcite-input-message>Message</calcite-input-message>`, {
+        "--calcite-input-message-spacing-value": {
+          targetProp: "--calcite-input-message-spacing-value",
+        },
+      });
+    });
   });
 });
--- a/packages/calcite-components/src/components/input-message/input-message.e2e.ts
+++ b/packages/calcite-components/src/components/input-message/input-message.e2e.ts
@@ -1,7 +1,8 @@
 // @ts-strict-ignore
 import { newE2EPage } from "@arcgis/lumina-compiler/puppeteerTesting";
 import { describe, expect, it, beforeEach } from "vitest";
-import { accessible, hidden, renders } from "../../tests/commonTests";
+import { accessible, hidden, renders, themed } from "../../tests/commonTests";
+import { html } from "../../../support/formatting";
 import { StatusIconDefaults } from "./interfaces";
 
 describe("calcite-input-message", () => {
@@ -139,6 +140,19 @@ describe("calcite-input-message", () => {
           requestedIcon = await iconEl.getAttribute("icon");
           expect(requestedIcon).toEqual("view-hide");
         });
+      });
+    });
+  });
+
+  describe("theme", () => {
+    describe("default", () => {
+      themed(html`<calcite-input-message>Message</calcite-input-message>`, {
+        "--calcite-input-message-icon-color": {
+          targetProp: "--calcite-input-message-icon-color",
+        },
+        "--calcite-input-message-spacing": {
+          targetProp: "--calcite-input-message-spacing",
+        },
       });
     });
   });
--- a/packages/calcite-components/src/components/input-message/input-message.scss
+++ b/packages/calcite-components/src/components/input-message/input-message.scss
@@ -10,9 +10,10 @@
 
 :host {
   @apply text-color-1 transition-default box-border flex h-auto w-full items-center font-medium opacity-100;
+
   margin-block-start: var(
     --calcite-input-message-spacing,
-    var(--calcite-input-message-spacing-value, --calcite-spacing-xxs)
+    var(--calcite-input-message-spacing-value, var(--calcite-spacing-xxs))
   );
 }
 
--- a/packages/calcite-components/src/components/input-message/input-message.scss
+++ b/packages/calcite-components/src/components/input-message/input-message.scss
@@ -3,8 +3,8 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
- * @prop --calcite-input-message-spacing-value: [Deprecated] Use `--calcite-input-message-spacing`.The top margin spacing above the component.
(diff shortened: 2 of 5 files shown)


packages/calcite-components/src/components/input-message/input-message.scss
driskull, line 1: Not sure if we want to address this here, but the line-height should have a var

image

packages/calcite-components/src/components/input-message/input-message.scss
```diff
@@ -4,33 +4,35 @@
  * These properties can be overridden using the component's tag as selector.
  *
  * @prop --calcite-input-message-spacing-value: The top margin spacing above the component.
+ * @prop --calcite-input-message-spacing: Specifies the top margin space above the component.
```
driskull, line 7: Should this spacing only work for the space above? If so, should it be named accordingly? I think the name and the description don't really match right now.

What is the difference between `--calcite-input-message-spacing-value` and `--calcite-input-message-spacing`? I think these are confusing.

driskull (changes requested):

packages/calcite-components/src/components/input-message/input-message.scss
jcfranco: This will be addressed in https://github.com/Esri/calcite-design-system/issues/8856 (namely https://github.com/Esri/calcite-design-system/issues/10052).

packages/calcite-components/src/components/input-message/input-message.scss
macandcheese: Agreed - just one here should suffice - it should only apply to the top spacing. The old one should probably be deprecated, "spacing-value" is a strange nomenclature for a public property.

macandcheese (changes requested):
Looking at the Chromatic, this introduces some spacing changes that should be looked at, for example: https://www.chromatic.com/test?appId=6266d45509d7eb004aa254fb&id=67da0165ade786a6a55044d8

packages/calcite-components/src/components/input-message/input-message.scss
```diff
@@ -3,34 +3,39 @@
...
 
 .calcite-input-message-icon {
   @apply transition-default pointer-events-none inline-flex flex-shrink-0;
-  margin-inline-end: theme("margin.2");
+  margin-inline-end: var(--calcite-spacing-sm);
 }
 
 :host([status="invalid"]) .calcite-input-message-icon {
-  color: var(--calcite-color-status-danger);
+  color: var(--calcite-input-message-icon-color, var(--calcite-color-status-danger));
```
macandcheese, line 26: These could probably _also_ support using `--calcite-icon-color` - currently that doesn't work to override.

packages/calcite-components/src/components/input-message/input-message.scss
```diff
@@ -3,34 +3,39 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
- * @prop --calcite-input-message-spacing-value: The top margin spacing above the component.
+ * @prop --calcite-input-message-spacing-value: [Deprecated] Use `--calcite-input-message-spacing`.The top margin spacing above the component.
+ * @prop --calcite-input-message-spacing: Specifies the top margin space above the component.
```
macandcheese, line 7: I think either this prop name or description could be a bit better - the "top margin space" is a little strange. cc @geospatialem or @DitwanP for doc consistency q there.

packages/calcite-components/src/components/input-message/input-message.scss
```diff
@@ -3,34 +3,39 @@
  *
  * These properties can be overridden using the component's tag as selector.
  *
- * @prop --calcite-input-message-spacing-value: The top margin spacing above the component.
+ * @prop --calcite-input-message-spacing-value: [Deprecated] Use `--calcite-input-message-spacing`.The top margin spacing above the component.
```
DitwanP, line 6:
```suggestion
 * @prop --calcite-input-message-spacing-value: [Deprecated] Use `--calcite-input-message-spacing`. The top margin spacing above the component.
```

Idk if we usually specify the direction of the spacing in the token names so I have two suggestions for description change:
1.  Specifies the amount of top margin above the component.
2. Specifies the margin spacing at the top of the component.

Elijbet:
> Looking at the Chromatic, this introduces some spacing changes that should be looked at, for example: https://www.chromatic.com/test?appId=6266d45509d7eb004aa254fb&id=67da0165ade786a6a55044d8

@macandcheese I think this was fixed here: [ab42f8b](https://github.com/Esri/calcite-design-system/pull/11759/commits/ab42f8bbd40a9f0737f1381b0182dd3cc18dcae6)

macandcheese (approved):

driskull (approved):
LGTM!

Should we note the deprecation in the changelog? I don't think it would show up as is since its not in the PR title.

jcfranco:
@Elijbet Can you update the PR description with a commit override to include the deprecation entry ([example](https://github.com/Esri/calcite-design-system/pull/11070))?

You can add this to PRs that aren't exclusively deprecation-related ([example](https://github.com/Esri/calcite-design-system/wiki/pr-review-guidelines#deprecation-prs)), but we'll need the override since it was already installed.

packages/calcite-components/src/components/input-message/input-message.e2e.ts
```diff
@@ -142,4 +143,25 @@ describe("calcite-input-message", () => {
       });
     });
   });
+
+  describe("theme", () => {
+    describe("default", () => {
+      themed(html`<calcite-input-message>Message</calcite-input-message>`, {
+        "--calcite-input-message-icon-color": {
+          targetProp: "--calcite-input-message-icon-color",
```
driskull, line 151: @Elijbet I just noticed that these target properties are incorrect. They should be targeting `--calcite-icon-color`, and `marginBlockStart`. Can you open a PR to fix these?

Elijbet: [refactor(input-message): status themed tests](https://github.com/Esri/calcite-design-system/pull/12701#top)#12701

PR #11778, opened 2025-03-20
refactor(tabs): use tag selector

anveshmekala:
**Related Issue:** #


aPreciado88 (approved):
🚀

jcfranco:
Sorry for the delay. Can you revert this? We use class selectors to keep tests stable when the component HTML changes. The names can end up being redundant, so I get why it looked safe to remove.


I thought this was already part of our conventions, but I couldn’t find a reference. I’ll add a note to our conventions.

PR #11779, opened 2025-03-20
feat: add browser join and browser plus

arowles:
**Related Issue:** #

## Summary
**added**

browser-join-16
browser-join-24
browser-join-32
browser-plus-16
browser-plus-24
browser-plus-32


jorgegutierrez2019 (approved):
looks good

Issue #9792, opened 2024-07-16
calciteTabChange event fires when the selected tab title is clicked


eriklharper:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Clicking a `calcite-tab-title` repeatedly fires the `calciteTabChange` event even though the selected tab doesn't actually change.

### Expected Behavior

The `calciteTabChange` event should only fire when the selected tab changes.

### Reproduction Sample

https://codepen.io/eriklharper/pen/xxowBZQ?editors=1111

### Reproduction Steps

1. Render some tabs
2. Wire up an event listener for `calciteTabChange` on the document
3. Click the selected tab title repeatedly
4. Result: `calciteTabChange` event fires each time the selected tab is clicked

### Reproduction Version

2.9.0

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)

DitwanP:
🍡 Verified on `3.2.0-next.7`

https://codepen.io/Ditwan-Price/pen/ogNrXqR?editors=1011

PR #11812, opened 2025-03-26
fix(tab-nav): emit calciteTabChange event only when selected tab changes

anveshmekala:
**Related Issue:** #9792

### DIFF
--- a/packages/calcite-components/src/components/tab-nav/tab-nav.e2e.ts
+++ b/packages/calcite-components/src/components/tab-nav/tab-nav.e2e.ts
@@ -37,7 +37,6 @@ describe("calcite-tab-nav", () => {
     const activeEventSpy = await page.spyOnEvent("calciteTabChange");
     const firstTabTitle = await page.find("calcite-tab-title");
 
-    firstTabTitle.setProperty("selected", true);
     await page.waitForChanges();
     expect(activeEventSpy).toHaveReceivedEventTimes(0);
 
@@ -47,7 +46,7 @@ describe("calcite-tab-nav", () => {
 
     await page.keyboard.press("Enter");
     await page.waitForChanges();
-    expect(activeEventSpy).toHaveReceivedEventTimes(2);
+    expect(activeEventSpy).toHaveReceivedEventTimes(1);
   });
 
   it("focuses on keyboard interaction", async () => {
--- a/packages/calcite-components/src/components/tab-nav/tab-nav.tsx
+++ b/packages/calcite-components/src/components/tab-nav/tab-nav.tsx
@@ -73,6 +73,8 @@ export class TabNav extends LitElement {
 
   private makeFirstVisibleTabClosable = false;
 
+  private userTriggeredTabActivation = false;
+
   // #endregion
 
   // #region State Properties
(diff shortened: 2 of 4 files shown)


packages/calcite-components/src/components/tab-nav/tab-nav.e2e.ts
```diff
@@ -37,7 +37,6 @@ describe("calcite-tab-nav", () => {
     const activeEventSpy = await page.spyOnEvent("calciteTabChange");
     const firstTabTitle = await page.find("calcite-tab-title");
 
-    firstTabTitle.setProperty("selected", true);
```
jcfranco, line 40: We still need this to ensure it doesn’t emit when selecting tab titles programmatically, right?

packages/calcite-components/src/components/tab/interfaces.ts
```diff
@@ -1,6 +1,7 @@
 export interface TabChangeEventDetail {
   /** The tab ID that just became selected */
   tab: number | string;
+  isUserTriggered?: boolean;
```
jcfranco, line 4: Nitpick: `userTriggered`?

packages/calcite-components/src/components/tab-nav/tab-nav.e2e.ts
```diff
@@ -47,7 +46,7 @@ describe("calcite-tab-nav", () => {
 
     await page.keyboard.press("Enter");
     await page.waitForChanges();
-    expect(activeEventSpy).toHaveReceivedEventTimes(2);
+    expect(activeEventSpy).toHaveReceivedEventTimes(1);
```
jcfranco, line 49: Can you update the test to cover emitting for both keyboard and mouse interactions? The current version won’t fail if one of the two interactions stops emitting.

jcfranco (changes requested):

packages/calcite-components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -241,13 +240,17 @@ export class TabNav extends LitElement {
...
     const activatedTabTitle = event.target as TabTitle["el"];
+    const currentSelectedTabTitle = this.selectedTitle;
 
     this.selectedTabId = event.detail.tab
       ? event.detail.tab
       : this.getIndexOfTabTitle(activatedTabTitle);
     event.stopPropagation();
 
     this.selectedTitle = activatedTabTitle;
+    if (currentSelectedTabTitle !== activatedTabTitle && event.detail.userTriggered) {
```
driskull, line 251: nitpick: Highly unlikely, but what if two tabs have the same title or no title at all? Is there a better unique identifier we could use here? Maybe an undocumented guid property?

driskull (approved):
👍

jcfranco (approved):
🚀

packages/calcite-components/src/components/tab-nav/tab-nav.tsx
anveshmekala: Great point, updated to use `id`.

driskull: @anveshmekala what if the tab doesn't have an id? Id isn't set automatically right?

driskull: Oh, it is. nevermind :)

Issue #11774, opened 2025-03-20
Block Group / List - events do not fire correctly when dragging via keyboard controls


ian13262:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Given a `calcite-block-group` with `dragEnabled: true`, and child `calcite-blocks` containing `calcite-lists` with `dragEnabled: true`, some events errantly fire when they are not supposed to. **See Repro for more details!**

### Expected Behavior

**See Repro for more details!**

### Reproduction Sample

https://codepen.io/iyouth66/pen/XJWEPgZ

### Reproduction Steps

1. Open the codepen: https://codepen.io/iyouth66/pen/XJWEPgZ
2. Open DevTools
3. Focus on the "Bike 3" element in the calcite-list. Press Enter > Move up. The element moves up as expected, but in the console you'll see that **calciteListOrderChange** AND **calciteBlockGroupOrderChange** were fired. **The calciteBlockGroupOrderChange event should not fire**
4. Also, notice that the item order **fails to change.**

Now:
1. Reload the page
2. Focus on the "Hike 3" element in the calcite-list. Press Enter > Move down. The element moves down, but a console error is thrown: 

![Image](https://github.com/user-attachments/assets/5eb34330-e0e3-4b57-afb0-c6de1857684c)

### Reproduction Version

3.0.3

### Relevant Info

cc @driskull @macandcheese

### Priority impact

impact - p1 - need for current milestone

### Impact

This is greatly affecting a few of our Map Config components in AWC, including the popup configuration display.

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Online

driskull:
It sounds like we need to stopPopagation or preventDefault on the handle events so that they are only honored by their closest parent sortable element.  Does that sound correct @jcfranco? Would we want to stop propagation or prevent default?

driskull:
@geospatialem @DitwanP assigned to myself for upcoming milestone

geospatialem:
Verified in `3.2.0-next.15` ⏫⏬

ian13262:
[...]

PR #11816, opened 2025-03-27
fix(list, block-group): fix drag handle events when lists or blocks are nested

driskull:
**Related Issue:** #11774

## Summary

- Prevents default on drag handle events when used by a parent component so that any further parents up the DOM won't receive them
- Parents further up the dom will ignore the cancelled event
- Modifies tests


packages/calcite-components/src/components/block-group/block-group.tsx
```diff
@@ -232,6 +232,7 @@ export class BlockGroup extends LitElement implements InteractiveComponent, Sort
       return;
     }
 
+    event.stopPropagation();
```
jcfranco, line 235: Can we make the event cancelable and cancel it instead of stopping propagation?

packages/calcite-components/src/components/list/list.tsx
```diff
@@ -532,6 +532,7 @@ export class List extends LitElement implements InteractiveComponent, SortableCo
       return;
     }
 
+    event.stopPropagation();
```
jcfranco, line 535: This and the block below (and any other common logic) would also be good to include in https://github.com/Esri/calcite-design-system/issues/11849.

driskull:

benelan (approved):
:+1:

jcfranco (approved):
Awesome! 🚀

Issue #11558, opened 2025-02-14
Add `focusTrap` common test helper


jcfranco:
**monday.com sync:** #18048690190

### Priority impact

p - medium

### Summary

We have a few focus-trapping component, so we need to introduce a test helper that will allow quick and easy test setup for this feature.

### Desired Outcome

Add a `focusTrap` test helper to our common tests directory and apply it to the following components:

* `dialog`
* `input-date-picker`
* `input-time-picker` (might not be applicable since this uses internal `popover`)
* `modal`
* `popover`
* `sheet`

jcfranco:
https://github.com/Esri/calcite-design-system/pull/12084/ added a starter version of the test helper, so this issue should focus on adding additional tests (along with necessary signature updates) to increase coverage.

PR #11823, opened 2025-03-28
fix: outside click of a focus trap should not focus initial focused element

driskull:
**Related Issue:** #10682

## Summary

- We are running into an issue because we're using the `setReturnFocus` to call `focusElement` on components.
- The focus trap logic calls `clickOutsideDeactivates` then `setReturnFocus` which always gets passed the initial focused element.
- When `setReturnFocus` is not used and an outside click occurs, the initial element is not focused.
  - I'm not sure why this is different and would need some investigation.
  - I think its because our setFocus methods requestAnimationFrame which the default focus methods don't
- To work around this, I setup a flag to know when an outside click occurs and to not do any calls to `focusElement`. This means when an outside click occurs, no initial focus element is returned to focus by the focus trap.

Discussion here: https://github.com/focus-trap/focus-trap/discussions/1429#discussioncomment-12647625


driskull:
@jcfranco I can work around this but its not super pretty.

packages/calcite-components/src/utils/focusTrapComponent.ts
```diff
@@ -51,20 +51,24 @@ export function connectFocusTrap(component: FocusTrapComponent, options?: Connec
   component.focusTrap = createFocusTrap(focusTrapNode, createFocusTrapOptions(el, options?.focusTrapOptions));
 }
 
+const outsideClickSet = new WeakSet<HTMLElement | SVGElement>();
```
jcfranco, line 54: Can you add a brief explainer or rename this to clarify its intent? If I'm following correctly, this is used to determine the deactivate/close cause during a component's focus return, right?

jcfranco (approved):
LGTM!

Before merging, can you add a test for this? Maybe in [`stackedFocusTrap.e2e.ts`](https://github.com/Esri/calcite-design-system/blob/b2c976214774490b5a313e1fafc67415fe212b37/packages/calcite-components/src/tests/stackedFocusTrap.e2e.ts)? This might be something we can include in https://github.com/Esri/calcite-design-system/issues/11558.

packages/calcite-components/src/utils/focusTrapComponent.ts
driskull: This is used to store a component that where the trap is being deactivated due to an outside click. Any suggestions on the name of the constant?

- deactivatedViaOutsideClickSet?
- outsideClickDeactivated?

jcfranco: `outsideClickDeactivated` gets my vote. ☑️

jcfranco:
> Its not about focus trap stacking, its about returning focus properly.

I'm aware. 😄 I was suggesting that as a way to have some coverage for this until we tackle #11558.

driskull:
Ill add something there temporarily for coverage then :)

jcfranco:
Thank you, sir! 🚀

driskull:
@jcfranco added tests

jcfranco (approved):

Issue #11592, opened 2025-02-19
Height and width of text-area with `max-length` change with validity changes


nwhittaker:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Given a `<calcite-text-area>` element with `max-length` and `validation-message` attributes defined, toggling the element's `status` between `valid` and `invalid` shrinks its width and causes the height of the textarea to ever increase.

https://github.com/user-attachments/assets/2c4bb6d8-186d-4b23-85fa-73471cbc2a12

### Expected Behavior

The textarea's height and width do not change in response to `status` changes.

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/MYWKPZo

### Reproduction Steps

1. Visit the repro and enter a value into the text-area.
2. Clear the value and see the text-area dimensions change and an error message appears.
3. Re-enter the value and see the error message disappears.
4. Re-clear the value and see the text-area's height increases and the error message reappears.
5. Repeat from step 3.

### Reproduction Version

3.0.2

### Relevant Info

Possibly related to https://github.com/Esri/calcite-design-system/issues/8456.

### Priority impact

impact - p2 - want for an upcoming milestone

### Impact

A workaround for the likeliest use-case is to enable the `limit-text` attribute, however this will not work for preventing other validity errors from triggering the bug.

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Field Apps

DitwanP:
🍠 Verified locally on `dev`

Issue #7060, opened 2023-05-30
Cancel debounced and throttled functions


driskull:
### Description

We need to cancel all debounced and throttled functions when the component is disconnected or when the component no longer needs to handle the debounced function.

### Proposed Advantages

Debounced functions no longer happen after the component is removed from the DOM.

## Throttle

- color-picker
- slider
- text-area

## Debounce

- action-bar
- combobox
- filter
- list
- slider

### Relevant Info

https://lodash.com/docs/4.17.15#debounce
https://lodash.com/docs/4.17.15#throttle

Elijbet:
This issue is blocked until Lumina is installed. The specific test I’m using here is not possible in E2E tests right now. 
There's no way in the current E2E environment to assert on utils that the components are using at runtime. This might be possible with Lumina migration, so we will wait for that before proceeding.

PR #11826, opened 2025-03-28
fix(text-area): no longer changes height & width when status is invalid

anveshmekala:
**Related Issue:** #11592 

## Summary

No longer changes dimensions of `text-area` when status is set to `invalid`.


driskull (approved):
👍

packages/calcite-components/src/components/functional/Validation.tsx
```diff
@@ -21,8 +22,12 @@ export const Validation = ({
   id,
   icon,
   message,
+  ref,
 }: ValidationProps): TemplateResult => (
-  <div class={CSS.validationContainer}>
+  <div
+    class={CSS.validationContainer}
+    ref={ref ? ref : (referenceElement): HTMLDivElement => referenceElement}
```
jcfranco, line 29: Can you extract the ref-handling logic to a top-level function instead of using an anonymous function? The latter will create a new function on each render.

jcfranco (approved):
Once comments are addressed, this should be good to merge! 🚀

packages/calcite-components/src/components/functional/Validation.tsx
```diff
@@ -3,26 +3,32 @@ import { h, LuminaJsx } from "@arcgis/lumina";
...
 export const Validation = ({
   scale,
   status,
   id,
   icon,
   message,
+  ref,
 }: ValidationProps): TemplateResult => (
-  <div class={CSS.validationContainer}>
+  <div class={CSS.validationContainer} ref={ref ? ref : setValidationReference}>
```
driskull, line 31: @anveshmekala Can you explain what this is doing? It doesn't appear to be storing the ref element anywhere or doing anything with it.

packages/calcite-components/src/components/text-area/text-area.tsx
```diff
@@ -77,44 +77,60 @@ export class TextArea
...
   // throttle is used to avoid flashing of textarea when user resizes.
-  private setHeightAndWidthToAuto = throttle(
-    (): void => {
-      if (this.resize === "vertical" || this.resize === "both") {
-        this.el.style.height = "auto";
-      }
-      if (this.resize === "horizontal" || this.resize === "both") {
-        this.el.style.width = "auto";
-      }
+  private updateSizeToAuto = throttle(
```
driskull, line 126: If possible, can the throttle be cancelled when component is disconnected? related to #7060

packages/calcite-components/src/components/text-area/text-area.tsx
```diff
@@ -77,44 +77,60 @@ export class TextArea
...
+    (dimension: "height" | "width"): void => {
+      this.el.style[dimension] = "auto";
```
driskull, line 128: Is there a way to add this style to an element internal to the component instead of the host which is in light dom and may be affected by app styles? Maybe a refactor issue for the future?

packages/calcite-components/src/components/functional/Validation.tsx
anveshmekala: This to avoid TypeError when `ref` is not set.  When user toggles the visibility of Validation element, ref is updated which fails when set to `undefined`.

To encounter this, we are parsing a dummy method which imitates a reference handler.

driskull: Why is the function returning the reference element? Shouldn't it be a void function?

packages/calcite-components/src/components/text-area/text-area.tsx
anveshmekala: I don't think it is recommended to add it to the internal element if we want the host dimensions to be in sync with native <textarea/> inside of shadowRoot if resized. 

Once user resizes, app styles will be ignored and it is the expected with native <textarea/>.

packages/calcite-components/src/components/functional/Validation.tsx
anveshmekala: Agree, it should return `void`. 

```
const setValidationReference = (): void => {
  return;
}
```

how about updating the name to `validationReference` instead of `setValidationReference` ?

driskull: Yeah that sounds good

Issue #10237, opened 2024-09-06
[Refactor/Tests] verify components in SSR


maxpatiiuk:
### Priority impact

p - medium

### Test type

SSR

### Which Component(s)

all

### Unstable Tests

[not actionable before the migration]

When Stencil component is rendered in the Hydrate build, Stencil was providing mocks for a lot of dom APIs. ([source code](https://github.com/ionic-team/stencil/tree/8f0a8823facade0be6eabf8569831f456497c2a0/src/mock-doc))

When Lit elements are rendered in SSR, Lit also provides a mock dom API, but it's much more lightweight ([source code 1](https://github.com/lit/lit/blob/main/packages/labs/ssr-dom-shim/README.md), [source code 2](https://github.com/lit/lit/blob/main/packages/labs/ssr/src/lib/dom-shim.ts))

That means that if before Stencil would mock the `element.closest()` calls in SSR, these are no longer mocked in Lit. This may cause some Calcite components to start failing in SSR after migration (with errors like `element.closest is not a function`)

Options:
- After the migration, manually test each Calcite component in SSR
- After the migration, setup automated SSR test for each component
- Look for an alternative DOM API shim (or maybe continue using Stencil's shim if that is possible) so that most DOM APIs are still covered. Cons: complexity and slower SSR
- If you really need some DOM APIs in Node.js, the Lit team would likely be open to contributions

When you identify a place where a DOM API that is not available in SSR is accessed, you have a few options:

- Use optional chaining for property/methods:

  ```diff
  - const parent = element.closest("calcite-tree");
  + const parent = element.closest?.("calcite-tree");
  ```

- Use `globalThis` with optional chaining for global APIs:

  ```diff
  - const userAgent = navigator.userAgent;
  + const userAgent = globalThis.navigator?.userAgent;
  ```

- Conditionally execute code in browser environment only:

  ```ts
  import {isServer} from 'lit';
  
  if (isServer) {
    // only runs in server environments like Node
  } else {
    // runs in the browser
  }
  ```

[...]

PR #11836, opened 2025-03-31
refactor(ssr): migrate to lit's isServer constant

benelan:
**Related Issue:** #10237

## Summary

Switch from the `isBrowser()` function to Lit's `isServer` constant, which will improve code splitting.

I tested in NextJS and SvelteKit apps, but we should also ask our SSR/SSG users to test `next` before the `v3.2.0` release.


packages/calcite-components/src/components/segmented-control/segmented-control.tsx
```diff
@@ -379,8 +378,8 @@ export class SegmentedControl
       this.calciteSegmentedControlChange.emit();
     }
 
-    if (isBrowser() && match) {
-      match.focus();
+    if (!isServer && match) {
+      match?.focus();
```
jcfranco, line 382: Isn't `match` guaranteed here?

packages/calcite-components/src/utils/browser.ts
```diff
@@ -1,13 +1,4 @@
-import { isServer } from "lit-html/is-server.js";
-
-export const isBrowser = (): boolean =>
```
jcfranco, line 3: ✨🧹✨

packages/calcite-components/src/utils/globalScript.ts
```diff
@@ -1,18 +1,18 @@
...
 /**
  * This file is imported in Stencil's `globalScript` config option.
  *
  * @see [Stencil's globalScript](https://stenciljs.com/docs/config#globalscript).
  */
 export default function (): void {
-  if (isBrowser()) {
-    if (document.readyState === "interactive") {
+  if (!isServer) {
+    if (document?.readyState === "interactive") {
```
jcfranco, line 12: Can `document` be undefined in browser contexts?

packages/calcite-components/src/utils/form.tsx
```diff
@@ -498,16 +498,16 @@ function syncHiddenFormInput(component: FormComponent): void {
     let input = extra.pop();
 
     if (!input) {
-      input = ownerDocument.createElement("input");
+      input = ownerDocument?.createElement("input");
```
jcfranco, line 501: For this and similar checks, should we return early by checking `isServer` instead? In this particular block, the next line will cause an error if `ownerDocument` is `undefined`.

packages/calcite-components/src/utils/globalScript.ts
benelan: No, I was just optional chaining all instances of `document` in case the `isServer` constant didn't work like Stencil's. But I can undo that

packages/calcite-components/src/components/segmented-control/segmented-control.tsx
benelan: Nice catch, that was a leftover from when I initially removed the conditional before trying out `isServer` instead of `isBrowser()`.

jcfranco (approved):
✨🦾🤖✨

PR #11853, opened 2025-04-03
docs: update 3rd party licenses

calcite-admin:
This PR was automatically generated by the update-licenses GitHub action

PR #11867, opened 2025-04-04
chore: avoid formatting 3rd-party license files

jcfranco:
**Related Issue:** N/A

## Summary

Stems from https://github.com/Esri/calcite-design-system/pull/11853#pullrequestreview-2741169660

This ensures that the `markdownlint` ignore list is used consistently across linting tasks.

This was done by primarily switching from [`markdownlint-cli`](https://www.npmjs.com/package/markdownlint-cli/) to [`markdownlint-cli2`](https://www.npmjs.com/package/markdownlint-cli2/), which builds on the former and simplifies the setup.

For context, the root `.markdownlintignore` wasn't picked up by `markdownlint-cli` because it must be [present in the `cwd`](https://github.com/igorshubovych/markdownlint-cli#ignoring-files). This was doable, but it meant duplicating the ignore file across packages.

## Notes

* renames `lint-staged` configs and uses `.mjs`
* tested with both `lint-staged` and NPM scripts
* `--dot` is [the](https://github.com/DavidAnson/markdownlint-cli2/blob/main/markdownlint-cli2.mjs#L468) [default](https://github.com/DavidAnson/markdownlint-cli2/issues/44)
* `.markdownlintignore` and `--ignore-path` was removed in favor of [config options](https://github.com/DavidAnson/markdownlint-cli2#markdownlint-cli2jsonc) (`ignores` and `gitignore`, respectively)which will be applied consistently
* moves the contents of `.markdownlintrc` to `.markdownlint-cli2.mjs`


benelan:
Is #11853 the only reason for switching? I don't think markdownlint formats html in markdown files

jcfranco:
>  I don't think markdownlint formats html in markdown files.

Not sure I follow, but this addresses the markdownlint ignore file not working properly and unintentionally formatting 3rd-party license text. I should've finished updating the PR description before posting in #11853, my bad! 😅

.markdownlint-cli2.mjs
```diff
@@ -0,0 +1,18 @@
+export default {
+  config: {
+    $schema:
+      "https://raw.githubusercontent.com/DavidAnson/markdownlint-cli2/main/schema/markdownlint-cli2-config-schema.json",
```
benelan, line 4: The JSON Schema won't work in a JavaScript file, and since there aren't types, we won't get any completion like this. If there isn't a reason this needs to be `mjs`, can you change it to `json`?

```json
{
  "$schema": "https://raw.githubusercontent.com/DavidAnson/markdownlint-cli2/main/schema/markdownlint-cli2-config-schema.json",
  "config": {
    "first-line-heading": false,
    "no-inline-html": false,
    "line-length": false,
    "code-block-style": {
      "style": "fenced"
    },
    "code-fence-style": {
      "style": "backtick"
    },
    "heading-style": {
      "style": "atx"
    },
    "no-duplicate-heading": {
      "siblings_only": true
    },
    "emphasis-style": {
      "style": "asterisk"
    },
    "strong-style": {
      "style": "asterisk"
    },
    "ul-style": {
      "style": "dash"
    }
  },
  "gitignore": true,
  "ignores": ["**/THIRD-PARTY-LICENSES.md"]
}
```

That way the options have completion and hover documentation.

benelan (approved):
Nice! I have one suggestion but otherwise LGTM!

> Not sure I follow, but this addresses the markdownlint ignore file not working properly and unintentionally formatting 3rd-party license text. 

Woops that was my bad. I somehow misread the thread in #11853 as being about a formatting issue I noticed in the HTML table of contributors. I was just talking to myself in https://github.com/Esri/calcite-design-system/pull/11853#issuecomment-2776951797, nothing to see here!

.markdownlint-cli2.mjs
jcfranco: Great catch! I _think_ I went with `mjs` early on when exploring options, in case I needed to extend from a base config. I'll make the change, thanks!

jcfranco: I'll actually need to use `jsonc` since `markdownlint-cli2` doesn't support the full range of options with `json`. I'll create a follow-up issue to include this extension in other linting configs/scripts.

PR #11831, opened 2025-03-30
test: improve design-tokens test coverage

jcfranco:
**Related Issue:** N/A

## Summary

Includes additional output files in snapshot tests. This will help with the next steps for #11391 and future design-token updates.

### Notes

* Adds test-related NPM scripts and updates GitHub action to set `MODE` env variable
* Use fixed timestamp for `docs` output in test runs (via `MODE`)
* Adds `MODE` as a Turbo [global passthrough env var](https://turbo.build/repo/docs/reference/configuration#globalpassthroughenv).

PR #11870, opened 2025-04-04
test: avoid timestamp comparison regardless of build or test context

jcfranco:
**Related Issue:** N/A  

## Summary  

This ensures token snapshot tests run consistently by always replacing the timestamp field in JSON files, if present.  

The workflow introduced in https://github.com/Esri/calcite-design-system/pull/11831 requires a special test build before running tests, which isn't always applicable. This was causing failures in the `deploy-prerelease` action ([example](https://github.com/Esri/calcite-design-system/actions/runs/14274033634/job/40013441322)), which runs token tests after a production build.

### DIFF
--- a/.github/workflows/pr-e2e.yml
+++ b/.github/workflows/pr-e2e.yml
@@ -32,8 +32,6 @@ jobs:
           node-version-file: package.json
       - if: steps.diff-check.outputs.SKIP == 'false'
         name: Run tests for testable changes
-        env:
-          MODE: test
         run: npm install && npm run build && npm run test
       - if: steps.diff-check.outputs.SKIP == 'true'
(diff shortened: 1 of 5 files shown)


jcfranco:
@matgalla We might want to drop this field to align with updated Style Dictionary output, which now defaults to **not** including a timestamp in headers (see https://styledictionary.com/version-4/migration/#fileheader-default-timestamp).

jcfranco:
It would also align with https://github.com/Esri/calcite-design-system/pull/11872.

benelan (approved):
Nice this is much cleaner!

benelan (approved):
Nice this is much cleaner!

PR #11872, opened 2025-04-04
test: include headers in snapshot diff

jcfranco:
**Related Issue:** N/A

## Summary

Style Dictionary no longer includes [timestamps in headers by default](https://styledictionary.com/version-4/migration/#fileheader-default-timestamp), so we can include them in snapshot tests.


benelan (approved):
:+1:

Issue #11268, opened 2025-01-12
Migrate to Vitest Browser Mode


maxpatiiuk:
### Priority impact

p - medium

### Test type

Change in testing framework

### Which Component(s)

-

### Unstable Tests

Follow up on https://github.com/Esri/calcite-design-system/pull/11262#issuecomment-2584971499

With migration to Lumina, Calcite switched from Jest to Vitest. However, we are still using Puppeteer.
Vitest browser mode has recently been released as experimental. It offers better developer experience as tests run in the browser (vs puppeteer approach where tests run in node.js and actions are proxied in the dom).
This means tests are able to use the native DOM API and access objects from the browser easily, and synchronously. However, it does mean that a different API needs to be used - given the number of tests Calcite has, this would likely require writing a small codemod.

Benefits of migration:
- better developer experience
- tests run in the browser
- performance should be faster, since there is no overhead of communication with the browser for each DOM change
- tests should get more stable. at present, some flakiness occurs because Node.js keeps a copy of the browser DOM, and syncs it with the browser after every change - despite this, it can still get out of date
- easy access to DOM APIs and dom libraries (in exchange for making access to node.js APIs harder, but not impossible)

Note, while vitest browser mode is gaining features rapidly, it is [still marked as experimental](https://vitest.dev/guide/browser/). Similarly, Storybook [has integration with Vitest browser mode](https://storybook.js.org/docs/api/portable-stories/portable-stories-vitest) to permit using stories as templates in tests, however, this integration is still in active development for web components. Thus, Calcite team can wait till Vitest browser mode is marked as stable before doing the switch.

### PR skipped, if applicable

#

DitwanP:
Revisit milestone assignment after [vitest](https://vitest.dev/guide/browser/) is out of experimental mode.

jcfranco:
Exciting news via @maxpatiiuk!

> * 4.0 in beta 
>   * Browser Mode stable soon!

<img width="865" height="506" alt="Image" src="https://github.com/user-attachments/assets/73efa81f-2105-4d89-be84-1d1d8b468c84" />

☝️ from https://www.youtube.com/watch?v=x7Jsmt_o9ek&t=456s

jcfranco:
Belated update: [Vitest 4.0 was released last week with stable browser mode](https://vitest.dev/blog/vitest-4.html)! 🚀 We can start migrating tests in phases now.

eriklharper:
[...]

PR #11877, opened 2025-04-07
test: set up vitest browser mode test environment

jcfranco:
**Related Issue:** #11268  

## Summary  

Some of Lumina's testing utils and patterns (e.g., `mount`) depend or work best on Vitest's experimental [browser mode](https://vitest.dev/guide/browser/), so this PR enables testing with it. This supports E2E and/or spec tests that would be simpler to write with this mode (e.g., #7060).

Tests meant for this mode should use the following naming pattern: `<name>.browser.(spec|e2e).tsx?` (see [example](https://github.com/Esri/calcite-design-system/pull/11878)).

**Note**: Until browser mode is stable, continue using the usual E2E or spec tests unless a browser-mode test is specifically needed.


driskull (approved):
LGTM!

eriklharper (approved):

packages/calcite-components/package.json
```diff
@@ -54,8 +54,12 @@
     "screenshot-tests:preview": "npm run util:prep-build-reqs && NODE_OPTIONS=--openssl-legacy-provider STORYBOOK_SCREENSHOT_LOCAL_BUILD=true storybook dev",
     "screenshot-tests:publish": "npm run screenshot-tests && storybook-to-ghpages --existing-output-dir=docs",
     "start": "npm run util:clean-js-files && concurrently --kill-others --raw \"trap 'npm run util:clean-js-files' INT TERM EXIT >/dev/null 2>&1 || true && tsc --project ./tsconfig-demos.json --watch\" \"npm run build:watch-dev\"",
-    "test": "vitest run",
-    "test:watch": "vitest",
+    "test": "concurrently \"npm:test:stable\" \"npm:test:experimental\"",
+    "test:stable": "STABLE_TESTS=true vitest run",
+    "test:experimental": "EXPERIMENTAL_TESTS=true vitest run --browser.headless",
+    "test:watch": "npm run test:watch:stable && npm run test:watch:experimental",
```
benelan, line 60: Did you confirm `test:watch` works? I don't think `test:experimental` will run because `test:stable` won't exit. You may need to use `concurrently` like the `test` script does.

benelan (approved):
I likey!

packages/calcite-components/package.json
jcfranco: Great catch. Will fix!

jcfranco: Used `concurrently` + passthrough options. One thing to note is that interactive watch mode won't work as before since both test runners might conflict when handling input. I've updated the contributing doc with this.

jcfranco:
Merging! @benelan LMK if there's any additional tweaks needed for the `test:watch` script + doc.

Issue #11871, opened 2025-04-04
calcite-list drag-enabled is not working in Firefox 137.0


sagewall:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

The drag and drop functionality in the calcite list has stopped working in the latest version of Firefox (v 137.0).

This can be reproduced by opeing up the Drag items between lists sample available on the calcite website
https://developers.arcgis.com/calcite-design-system/components/list/
or in the codepen link below.  There are not any errors present in the console.

### Expected Behavior

`drag-enabled` works in the Firefox browser

### Reproduction Sample

https://codepen.io/sagewall/full/vEYPpzY

### Reproduction Steps

1. Open the [code pen](https://codepen.io/sagewall/full/vEYPpzY)
2. Try to drag and move one of the list items
3. The list item will fail to move

### Reproduction Version

3.1.0

### Relevant Info

Repros on both Mac and Windows

### Priority impact

impact - p0 - emergency

### Impact

Forces users to use another browser like chrome or safari for drag and drop to work in all Esri applications that use the calcite list or any of the layer list components such as Map Viewer, Experience Builder and Instant Apps.

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Maps SDK for JavaScript

sagewall:
cc @zaramatheson

driskull:
Yeah looks like a FF bug. I have a repro using only the sorting library:
https://codepen.io/driskull/pen/YPzgeVW?editors=0110

Heres a workaround:
https://codepen.io/driskull/pen/wBvOyjm

driskull:
Opened a Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1959275

geospatialem:
> Opened a Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1959275

Addressed a fix to the above, however the underlying browser issue with Firefox will be tackled by the Firefox team separate to Calcite's efforts and prioritization.

DitwanP:
🍠 Verified on Firefox nightly version 140

This has been fixed as of Firefox version 139:

[...]

PR #11883, opened 2025-04-07
fix: remove unused dragging styles

driskull:
**Related Issue:** #11871

## Summary

- In FF, the `overflow:hidden` on the chosen class was causing the drag and drop to break
- In this PR, unnecessary styles are removed in order to prevent the breakage and any future breakage
- No tests are present because our testing is chromium based


jcfranco (approved):

jcfranco:
Confirmed w/ @benelan that this needs to land on `dev` before proceeding with the hotfix release.

PR #11910, opened 2025-04-09
chore: improve design-token types

jcfranco:
**Related Issue:** N/A  

## Summary  

Improves types across `calcite-design-tokens` package.

### Notes

* extends `style-dictionary` types to narrow down types (e.g., supported platforms and expected properties)
* defines enum for all supported platforms
* updates type imports
* allows `.d.ts` files to be tracked
* lifts `types` folder as these are shared


packages/calcite-design-tokens/src/build/format/index-file.ts
```diff
@@ -44,11 +46,11 @@ export const formatIndexFile: FormatFn = async (args) => {
     ),
   } as const;
 
-  const classGroupStrategy = format === "scss" ? "@mixin " : ".";
+  const classGroupStrategy = format === Platform.css ? "." : "@mixin ";
```
aPreciado88, line 50: Should this be `const classGroupStrategy = format === Platform.css ? "@mixin" : ". ";` ?

aPreciado88 (approved):
This LGTM!

packages/calcite-design-tokens/src/build/format/index-file.ts
jcfranco: I switched it to `css` to have consistent comparisons, so I had to invert the values.

Issue #11903, opened 2025-04-08
[radio-button] Selected states use inner box shadows that aren't supported by existing tokens


matgalla:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

## Description

No semantic border tokens are available for use with a checked radio-button's medium and large size borders which are 5 and 6px:

![Image](https://github.com/user-attachments/assets/3da0be28-61fd-4342-89a6-ac341810ac70)

Existing border tokens:
```
--calcite-border-width-none: 0;
--calcite-border-width-sm: 1px;
--calcite-border-width-md: 2px;
--calcite-border-width-lg: 4px;
```

### Acceptance Criteria

Radio-button is restructured to use a white, 4px circle for all sizes instead of the current inner box shadow implementation.

#### Details:

- Modify the selected radio-button to be entirely filled in with `--calcite-color-brand` with no border
- Add a 4px circle in the center for all 3 sizes filled in with `--calcite-color-foreground-1`

![Image](https://github.com/user-attachments/assets/7eda6971-481e-4964-bd8b-a9fe5bcbbe1a)

This is a structural change that doesn't create any visual change while allowing for easier implementation of component tokens. It will hopefully make the component tokens for radio-button easier to modify as well.

### Which Component

radio-button

### Priority impact

impact - p2 - want for an upcoming milestone

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

**monday.com sync:** #8889592254

ashetland:
For the case of borders, I think this should be sufficient:
```
--calcite-border-width-xl: 5px;
--calcite-border-width-xxl: 6px;
```

matgalla:
I'm considering a different approach that help could solve this issue.

We could modify the selected radio-button to be entirely filled in with `--calcite-color-brand` with no border, then have a 4px circle in the center for all 3 sizes filled in with `--calcite-color-foreground-1`. This has the same visual result as the current component.

<img width="438" alt="Image" src="https://github.com/user-attachments/assets/faf13ed3-33d2-40ab-895c-f857edb06eff" />

Pros:
- A simpler solution that's consistent across sizes
- Avoids the need for new semantic tokens for this single use case
[...]

PR #11852, opened 2025-04-02
refactor(tokens): Update JSON files to support Tokens Studio management

matgalla:
**Related Issue:** #11391 

## Summary
This update makes changes necessary for the modification of Calcite tokens within Tokens Studio. No changes are made to the final tokens themselves in this update.

Criteria for completion:
- [x] Import current token files into TStudio (TStudio automatically moves things around, but there should be no functional change)
- [x] Separate light/dark semantic color tokens into their own files so that TStudio can use them as themes
- [x] Change all instances of `"type": "size"` to `"type": "sizing"` For TStudio organization
- [x] Change all instances of `"type": "space"` to `"type": "spacing"` For TStudio organization
- [x] Change all instances of `"type": "breakpoint"` to `"type": "dimension"` For TStudio organization
- [x] Add TStudio themes for different token types

This update is one step in the process of moving our token management over to Tokens Studio.

## Next steps
- Update snapshot tests
- Remove token attributes
- Add missing core color tokens
- Fix incorrect core color values
- Audit the structural differences between tokens and Figma variables
- Restructure tokens to align with Figma variables where necessary
- Resolve #10050 
- Resolve #11565 
- Resolve #11581 
- Resolve #11725 
- Audit the value differences between tokens and Figma variables (mainly for percentages, and typography)
- Finalize token and Figma variable values with Calcite designers and developers
- Update token values to align with Figma variables where necessary
- Document the new token creation and modification process

Issue #11725, opened 2025-03-10
Neutral color primitive value update


trailtraveler:
### **Background**
The Calcite Brand design system team is developing a new dark and light theme system that better aligns with current elevation trends in design systems across the web. Additionally, we are standardizing our themes to match what front-end designers are adjusting in the current system. With this update, we aim to move one step closer to a foundational neutral color system to use with our backgrounds and, eventually, for our borders and text styles as well.

### **Colors to change**
Approximately half of the current blk-XXX primitive token values are somewhat slightly off from what can be achieved using a full percentage of black over white. Although this change may be so subtle that it goes unnoticed by the viewer, maintaining an accurate underlying foundation is a crucial standard to build upon.

The primitives to update with new values can be seen in the Figma document [here](https://www.figma.com/design/wtDxGgIJsb4dVGQV2BKTeO/Color-Exploration?node-id=33-1847&t=7DitcRX1j7kyTMJf-1), the image attached to this issue, and in the list below:

| Primitive   | New Value |
| -------- | ------- |
| blk-005 |  #F7F7F7 |
| blk-010 | #F2F2F2 |
| blk-020 | #EBEBEB |
| blk-030 | #DEDEDE |
| blk-050 |  #C9C9C9 |
| blk-080 | #ABABAB |
| blk-090 | #9E9E9E |
| blk-140 | #6B6B6B |
| blk-150 | #616161 |
| blk-160 | #545454 |
| blk-190 | #363636 |
| blk-210 | #212121 |
| blk-220 | #141414 |
| blk-230 | #0A0A0A |
| blk-235 | #050505 |

![Image](https://github.com/user-attachments/assets/14671db8-8907-49a6-a82e-ea681894eeb7)

In addition to the `colors.json` file, we would like to update the Calcite Core Variables and UI Kit Figma tokens to pull the correct value to the brand library components.

👀 @SkyeSeitz @matgalla 

**monday.com sync:** #8717963820

matgalla:
Reopening after being closed due to unintentional automation

DitwanP:
🍡 Verified

PR #11655, opened 2025-02-28
refactor: bump `style-dictionary` and `@tokens-studio/sd-transforms` to latest

jcfranco:
**Related Issue:** #11391

## Summary

This updates the design tokens build with significant restructuring changes, based on @alisonailea’s [astump/style-dictionary-4.x](https://github.com/Esri/calcite-design-system/tree/astump/style-dictionary-4.x) branch. The output remains in sync with production as much as possible to prevent breaking changes.

This is the first step required before addressing additional changes proposed in #11391.

#### Source restructuring

* Deprecates global output files in favor of semantic output
* Splits source into:
  * **config** – main setup for output files
  * **build** – `style-dictionary` + `@tokens-studio/sd-transforms` configuration
  * **tokens** – design token JSON files (source of truth)
* Moves tests one level higher

#### Test and snapshot updates

* Tweaks spec test snapshot to make it easier to spot differences in test failures (see [alternative diff](https://www.diffchecker.com/Yxwj1x9x/) and [commit details](https://github.com/Esri/calcite-design-system/pull/11655/commits/d4695e2ffbfac03f3b3ca6364e4151cc0072a16c))
  * Should be mainly visual or have no impact on consumable output
* Removes non-token-related tests – these should be added once the design tokens are updated and stable

#### Transform and token updates

* Configures and overrides transforms to match the existing test snapshot; these will be adjusted or removed as updates are made
* Removes the button component token source file (unused)
* Adds token-related entries to our custom `@cspell` dictionary
* JS/Docs output types (undocumented)
  * Types now match the output (e.g., compare `{semantic.color.foreground.1}` [source](https://cdn.jsdelivr.net/npm/@esri/calcite-design-tokens@3.0.0/dist/js/global.js) vs [type](https://cdn.jsdelivr.net/npm/@esri/calcite-design-tokens@3.0.0/dist/js/global.d.ts))
  * No longer have strongly-typed attributes object

#### Test failures

There are a couple of [expected spec test failures](https://github.com/Esri/calcite-design-system/actions/runs/13952094497/job/39053999330?pr=11655) related to shadows:
* **Color values** based on a token are invalid
  * Example: `--calcite-shadow-0: 0 0 0 0 rgba(var(--calcite-color-neutral-blk-240), 0);` – the color part is [invalid syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/rgb)
* **Dimension values** are no longer unitless
* I'd like to have these reviewed before updating snapshot tests

[...]

PR #11915, opened 2025-04-09
refactor(tokens): Remove unnecessary token data

matgalla:
**Related Issue:** #11391 

## Summary
Since #11655 allows us to build the final variables with the base token info, we don't need the excess token metadata. This PR removes all the fields within `attributes` from core and semantic tokens aside from the `category` field, which we will keep to assist with niche token instances and documentation purposes.

## Criteria for completion
- [x] Remove `attributes` (except for `category`) from all semantic and core tokens
- [x] Ensure that organization and token values are retained
- [x] Ensure that the token build process is unaffected

## Previous steps
- [x] #11852 

## Next steps
- Add missing core color tokens
- Fix incorrect core color values
- Audit the structural differences between tokens and Figma variables
- Restructure tokens to align with Figma variables where necessary
- Address #10050 
- Address #11565 
- Address #11581 
- Address #11725 
- Address #11903 
- Audit the value differences between tokens and Figma variables (mainly for percentages and typography)
- Finalize token and Figma variable values with Calcite designers and developers
- Update token values to align with Figma variables where necessary
- Document the new token creation and modification process


jcfranco:
+@macandcheese for doc site tooling consideration.

packages/calcite-design-tokens/src/tokens/semantic/typography.json
```diff
@@ -15,24 +15,7 @@
...
-              "kind": "",
-              "appearance": "",
-              "property": "",
-              "variant": "",
-              "state": "",
-              "scale": "",
-              "context": "",
-              "mode": ""
-            }
+            "category": ""
```
jcfranco, line 18: Side note: should this be populated in a follow-up PR?

jcfranco (approved):

macandcheese (commented):
Looks good - just a request to preserve "tier" as well.

jcfranco:
> Looks good - just a request to preserve "tier" as well.

`tier` is still under `calcite-schema`. It's added by the build now instead of the token source.

packages/calcite-design-tokens/src/tokens/semantic/typography.json
matgalla: For sure. At the very least, these should follow the same system as the other `category` fields across tokens. I'll populate this in a separate PR.

PR #11926, opened 2025-04-10
feat(shell-panel): add component tokens

driskull:
**Related Issue:** #7180

## Summary

Adds the following tokens for `shell-panel` component.

- --calcite-shell-panel-background-color: Specifies the background color of the component.
- --calcite-shell-panel-text-color: Specifies the text color of the component.
- --calcite-shell-panel-resize-background-color: Specifies the background color of the resize handle.
- --calcite-shell-panel-resize-icon-color: Specifies the text color of the resize handle.


macandcheese (changes requested):
Code looks good - I guess I'd just consider the Shell Panel to be a standalone component...

IMO setting `--calcite-shell-panel-shadow` instead of `--calcite-shell-shadow` seems more appropriate here. cc @ashetland @SkyeSeitz for more design thoughts

driskull:
> Code looks good - I guess I'd just consider the Shell Panel to be a standalone component...

I'm not sure I consider it a standalone component as it only works inside a shell. Its not intended to be used outside a shell on its own.

SkyeSeitz:
> Code looks good - I guess I'd just consider the Shell Panel to be a standalone component...
> 
> IMO setting `--calcite-shell-panel-shadow` instead of `--calcite-shell-shadow` seems more appropriate here. cc @ashetland @SkyeSeitz for more design thoughts

Agree here. Tokens labeled as `--calcite-shell` seem to imply Shell itself has a shadow, text, etc. apart from Shell Panel. 

Would these tokens effect other slotted content such as `calcite-navigation` in the `header` slot?

driskull:
> Agree here. Tokens labeled as --calcite-shell seem to imply Shell itself has a shadow, text, etc. apart from Shell Panel.

Renamed them to `shell-panel`.

> Would these tokens effect other slotted content such as calcite-navigation in the header slot?

Nope, just the shadow DOM of the `shell-panel` component.

driskull:
@macandcheese can you review again plz? 🥺

macandcheese (approved):
Looks good, thanks for updating names!

PR #11937, opened 2025-04-11
fix(input-date-picker): allow clearing invalid date value

jcfranco:


packages/calcite-components/src/components/input-date-picker/input-date-picker.e2e.ts
```diff
@@ -1083,24 +1083,105 @@ describe("calcite-input-date-picker", () => {
...
+
+      it("should clear with invalid value", async () => {
+        const inputValue = "13/37";
+        const page = await newE2EPage();
+        await page.setContent(html`<calcite-input-date-picker></calcite-input-date-picker>`);
+        const inputDatePicker = await page.find("calcite-input-date-picker");
+        const input = await page.find("calcite-input-date-picker >>> calcite-input-text");
+        await inputDatePicker.callMethod("setFocus");
+        await inputDatePicker.type(inputValue);
+        await inputDatePicker.press("Escape");
```
anveshmekala, line 1120: Nitpick: Don't think this is required. Applies to all other tests.

packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
```diff
@@ -323,8 +323,10 @@ export class InputDatePicker
   }
 
   set value(value: string | string[]) {
-    const oldValue = this._value;
-    if (value !== oldValue) {
+    const valueChanged = value !== this._value;
+    const invalidValueCleared = value === "" && this.startInput?.value !== "";
```
anveshmekala, line 327: If the user skip setting startInput & set an invalid value in endInput value, clearing those values wouldn't work.

anveshmekala (changes requested):
Looking good, added a few comments.

packages/calcite-components/src/components/input-date-picker/input-date-picker.tsx
jcfranco: Great catch. Fixed!

Issue #11790, opened 2025-03-21
[notice] close icon variable unexpectedly changes the title text color


matgalla:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Customizing `notice`'s `--calcite-notice-close-text-color-hover` variable changes the color of both the close icon's hover and the title text.

<img width="1018" alt="Image" src="https://github.com/user-attachments/assets/7c17c620-ca82-461d-9d32-0161092656e7" />

The variable is also named `-close-text`

### Expected Behavior

- The variable only should only change the close icon hover color
- The variable should be named `--calcite-notice-close-icon-color-hover`
- The title text should have it's own, new color variable: `--calcite-notice-title-text-color`
- Same for the message text: `--calcite-notice-message-text-color`

### Reproduction Sample

https://codepen.io/mac_and_cheese/pen/emYrKQx?editors=100

### Reproduction Steps

1. Visit `notice` sample: https://developers.arcgis.com/calcite-design-system/components/notice/#sample
2. Enable `closable` prop
3. Customize the `--calcite-notice-close-text-color-hover` variable

### Reproduction Version

v3.0.3

### Relevant Info

Chrome on MacOS/Windows/Android and Firefox on Windows

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [x] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (design)

geospatialem:
Verified with `3.2.0-next.15`

![Image](https://github.com/user-attachments/assets/8625eef1-1d0d-45b2-a6c7-6f8a0efb0572)

PR #11938, opened 2025-04-11
feat(notice): add token for close icon and title text color

DitwanP:
**Related Issue:** #11790 

## Summary
Issue was initially to fix the fact that the title color was being changed with the close icon hover token. In addition I'm proposing a name change (deprecation) for the following tokens:

`--calcite-notice-close-text-color`
`--calcite-notice-close-text-color-hover`

to:

`--calcite-notice-close-icon-color`
`--calcite-notice-close-icon-color-hover`

Also a new token would be added to target the title text color:

`--calcite-notice-title-text-color`

deprecate(notice): deprecate the notice close text color tokens

### DIFF
--- a/packages/calcite-components/src/components/notice/notice.scss
+++ b/packages/calcite-components/src/components/notice/notice.scss
@@ -6,10 +6,13 @@
  * @prop --calcite-notice-background-color: Specifies the component's background color.
  * @prop --calcite-notice-close-background-color-focus: Specifies the component's background color when focused.
  * @prop --calcite-notice-close-background-color-press: Specifies the component's background color when active.
- * @prop --calcite-notice-close-text-color-hover: Specifies the background color of the component's close button when hovered.
- * @prop --calcite-notice-close-text-color: Specifies the text color of the component's close button.
+ * @prop --calcite-notice-close-icon-color-hover: Specifies the background color of the component's close button when hovered.
+ * @prop --calcite-notice-close-icon-color: Specifies the text color of the component's close button.
+ * @prop --calcite-notice-title-text-color: Specifies the component's title text color.
  * @prop --calcite-notice-content-text-color: Specifies the component's content text color.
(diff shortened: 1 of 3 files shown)


driskull (commented):
@DitwanP can you update the tokens in the custom theme story?

packages/calcite-components/src/components/notice/notice.scss
```diff
@@ -6,10 +6,13 @@
...
  * @prop --calcite-notice-close-background-color-focus: Specifies the component's background color when focused.
  * @prop --calcite-notice-close-background-color-press: Specifies the component's background color when active.
- * @prop --calcite-notice-close-text-color-hover: Specifies the background color of the component's close button when hovered.
- * @prop --calcite-notice-close-text-color: Specifies the text color of the component's close button.
+ * @prop --calcite-notice-close-icon-color-hover: Specifies the background color of the component's close button when hovered.
+ * @prop --calcite-notice-close-icon-color: Specifies the text color of the component's close button.
+ * @prop --calcite-notice-title-text-color: Specifies the component's title text color.
  * @prop --calcite-notice-content-text-color: Specifies the component's content text color.
  * @prop --calcite-notice-width: [Deprecated] Specifies the component's width.
+ * @prop --calcite-notice-close-text-color-hover: [Deprecated] Specifies the background color of the component's close button when hovered.
```
driskull, line 14: For the deprecation notice, should we say to use the new var here? Applies to both deprecations

driskull (approved):
👍 👏 💯

packages/calcite-components/src/components/notice/notice.scss
DitwanP: Yes indeed we should, good catch!

Elijbet (approved):
Awesome! Thanks for cleaning that up.

PR #11734, opened 2025-03-12
feat(input-date-picker): add component tokens

Elijbet:
**Related Issue:** #7180 

Adds the following tokens to `input-date-picker` component:
`--calcite-input-date-picker-border-color`: Specifies the component's input border color.
`--calcite-input-date-picker-divider-color`: Specifies the component's divider color between two inputs when in range mode.
`--calcite-input-date-picker-shadow`: Specifies the component's input shadow.
`--calcite-input-date-picker-calendar-shadow`: The component's calendar shadow.
`--calcite-input-date-picker-icon-color`: Specifies the component's in-field input icon color.
`--calcite-input-date-picker-icon-color-hover`: Specifies the component's in-field input icon color on hover.
`--calcite-input-date-picker-actions-icon-color`: Specifies the component's actions icon color.
`--calcite-input-date-picker-actions-icon-color-hover`: Specifies the component's actions icon color on hover.
`--calcite-input-date-picker-calendar-icon-color`: Specifies the component's calendar icon color.
`--calcite-input-date-picker-calendar-icon-color-hover`: Specifies the component's calendar icon color.
`--calcite-input-date-picker-background-color`: Specifies the component's input background color.
`--calcite-input-date-picker-calendar-actions-background-color`: Specifies the background color of actions of the calendar.
`--calcite-input-date-picker-calendar-actions-background-color-hover`: Specifies the background color of actions of the calendar when hovered.
`--calcite-input-date-picker-calendar-actions-background-color-press`: Specifies the background color of actions of the calendar when pressed.
`--calcite-input-date-picker-calendar-selected-background-color`: Specifies the background color of selected day and day range of the calendar.
`--calcite-input-calendar-day-background-color`: Specifies the background color of day of the calendar.
`--calcite-input-date-picker-calendar-day-background-color-hover`: Specifies the background color of day of the calendar when hovered.
`--calcite-input-date-picker-calendar-day-outside-range-background-color-hover`: Specifies the background color of day's outside current range of the component's date-picker when hovered.
`--calcite-input-date-picker-corner-radius`: Specifies the component's input corner radius.
`--calcite-input-date-picker-calendar-corner-radius`: Specifies the component's calendar border radius.
`--calcite-input-date-picker-text-color`: Specifies the component's input text color.
`--calcite-input-date-picker-placeholder-text-color`: Specifies the component's input placeholder text color.
`--calcite-input-date-picker-calendar-actions-text-color`: Specifies the text color of the component's calendar actions.
[...]

PR #11940, opened 2025-04-11
feat(input-time-picker): add component tokens

driskull:
**Related Issue:** #7180

## Summary

Adds the following tokens for `input-time-picker` component.

- --calcite-input-time-picker-icon-color: Specifies the component's icon color.
- --calcite-input-time-picker-icon-color-hover: Specifies the component's icon color when hovered or focused.

### DIFF
--- a/packages/calcite-components/src/components/input-time-picker/input-time-picker.e2e.ts
+++ b/packages/calcite-components/src/components/input-time-picker/input-time-picker.e2e.ts
@@ -20,6 +20,7 @@ import {
   reflects,
   renders,
   t9n,
+  themed,
 } from "../../tests/commonTests";
 import {
   getFocusedElementProp,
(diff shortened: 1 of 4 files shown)


driskull:
Ping. still need a review on this one.

macandcheese (approved):
🎨

anveshmekala:
I think tokens to customize time-picker & input are required here as per https://github.com/Esri/calcite-design-system/pull/11734#discussion_r2017647448

eriklharper (changes requested):
Can we hold off on merging this until #11712 lands?  This is going to cause conflicts with my changes on #11712 since the input-time-picker no longer renders an input-text element under-the-hood.  It looks like the primary add are just color tokens though.  @ashetland has also updated the Input Time design spec to communicate the `padding-inline-start` and `padding-inline-end` for each scale and to add a 24px hit target for every toggle icon regardless of scale.

My vote is to hold off on merging this because the internal structure of `input-time-picker` is changing drastically, which affects all these changes to the CSS.

driskull:
Sounds good, ill hold off on this one.

> Q: Are tokens to customize time-picker & input required here as per https://github.com/Esri/calcite-design-system/pull/11734#discussion_r2017647448 ?

Yeah that's a good catch. Ill add those once https://github.com/Esri/calcite-design-system/pull/11712 lands

PR #11943, opened 2025-04-11
refactor: add `useSetFocus` controller

jcfranco:
**Related Issue:** N/A  

## Summary  

This adds a controller for consistent and simpler `setFocus` behavior across components. 

### Noteworthy changes

* updates `dom#focusElement()` to use `focusFirstTabbable(el)` instead of calling `el.focus()` if there's no `setFocus()` available and adds `context` argument to allow short-circuiting focus logic
* adds `includeContainer` to `dom#focusFirstTabbable`, `dom#getFirstTabbable` for focus target scope
* updates `dom#focusElementInGroup()` to support `includeContainer` and whether to consider the target element as the focus context 
* adds _internal_ `dom#focusFirstFocusable`, `dom#getFirstFocusable` for additional focus options – can expose if deemed useful
* automatically skips focusing if the component is disabled
  * converts `interactive.spec.ts` test to `interactive.browser.spec.tsx` to support ☝️
* `setFocus` controller supports customization for advanced use cases, such as 
  * defining focus strategy – use first-tabbable (default) vs first-focusable (both from `tabbable`)
  * setting target scope – excludes container (default) or not when looking for first tabbable/focusable
* complex focusing cases will still work, but will be external to the controller (e.g., list's `focusCell` calls).
* for granular or custom, internal focusing, use `focusElement` instead of `this.setFocus`, unless that behavior is desired

### Next steps

* refactor utils to use object parameters for clarity as they are getting complex
* refactor external focus logic to use `setFocus` controller, if possible


jcfranco:
Went with `focusSetter` as the default name for the internal prop holding the controller export, but open to suggestions.

packages/calcite-components/src/controllers/useSetFocus.ts
```diff
@@ -0,0 +1,85 @@
...
+
+type FocusStrategy = "focusable" | "tabbable";
+type FocusConfig = { target: FocusableElement; includeContainer?: boolean; strategy?: FocusStrategy };
+
+export interface UseSetFocus {
+  (getFocusTarget: () => FocusableElement | FocusConfig | undefined): Promise<void>;
+}
+
+interface SetFocusComponent extends LitElement, Partial<Pick<InteractiveComponent, "disabled">> {
+  /** Sets focus on the fist focusable `calcite-radio-button` element in the component. */
```
eriklharper, line 15: ☝️ should remove `calcite-radio-button` here from this comment.

packages/calcite-components/src/controllers/useSetFocus.ts
```diff
@@ -0,0 +1,85 @@
...
+        component.el.addEventListener("focusout", handleFocusOut, { signal: abortController.signal });
+      });
+    });
+
+    controller.onDisconnected(() => {
+      component.el.removeEventListener("focusout", handleFocusOut);
+    });
+
+    return async (getFocusTarget): Promise<void> => {
+      if (component.disabled) {
```
eriklharper, line 46: We should probably update the controller's types to expect `disabled` to be defined on the component? ☝️

eriklharper:
> Went with `focusSetter` as the default name for the internal prop holding the controller export, but open to suggestions.

Yeah, I've been thinking it would be good to standardize the naming structure for the internal component properties meant for storing controller instances.  I feel like the word "controller" in the name, however verbose, could be nice because it is direct?  I think maybe the name of the controller, plus the word "controller" could be the most direct way we could adopt, so it would be come `useSetFocusController`.

packages/calcite-components/src/controllers/useSetFocus.ts
eriklharper: Looks like it is handled on line 14 above, is that right?

jcfranco: Correct. It's leveraging the interface from the interactivity-handling module.

packages/calcite-components/src/controllers/useSetFocus.ts
```diff
@@ -0,0 +1,85 @@
...
+        return;
+      }
+
+      const focusConfig = toFocusConfig(getFocusTarget());
```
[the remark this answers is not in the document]
jcfranco: Not needed since it's required by the [controller's interface](https://github.com/Esri/calcite-design-system/blob/360e129c1d431ce0e77b2a9321d72e2bf9793bb0/packages/calcite-components/src/controllers/useSetFocus.ts#L11).

Issue #10271, opened 2024-09-11
Inputs inside list lose focus when arrow keys pressed


nwhittaker:
### Check existing issues

- [X] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

Given a list of list items that contain input elements: When an input has focus and the user presses an `up` or `down` arrow key, focus can shift to the next list item instead of performing the input's default behavior.

### Expected Behavior

Focus is not shifted from the input and the input's default behavior is performed. In the case of text inputs, the cursor should move to the beginning or end of its value. For number inputs, the value should be incremented or decremented. Etc…

### Reproduction Sample

https://codepen.io/nwhittaker-esri/pen/OJeqrJq

### Reproduction Steps

1. Visit the repro.
2. Focus the first text input and hit either the `up` or `down` arrow key and see focus shifted away.
3. Focus the second input and hit the `down` arrow key and see the value is decremented but focus is also shifted the next list item. Interestingly, hitting the `up` arrow key correctly increments the value without shifting focus to the previous list item.

### Reproduction Version

2.12.2

### Priority impact

impact - p2 - want for an upcoming milestone

### Impact

A workaround is to ignore keyboard events on the list-item if the target is a descendant.

```ts
list.querySelector('calcite-list-item').addEventListener('keydown', event => {
  if (!event.target.matches('calcite-list-item')) {
    event.stopPropagation()
  }
})
```

However, this listener can't be delegated so it must be attached/detached to every list item. Which can become cumbersome for dynamically generated lists and easily lead to memory leaks.

### Calcite package

- [X] @esri/calcite-components
- [ ] @esri/calcite-components-angular
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Field Apps

driskull:
Input should be preventing default on ArrowDown just like it is on ArrowUp.

driskull:
@geospatialem I don't think this is a bug. I think the author should be preventing key events in order to handle them.

nwhittaker:
@driskull can you clarify the expected behavior? In the repro, the `up` and `down` keys don't behave the same. The `down` key behaves as observed and the `up` key behaves, in my opinion, as expected. Are you suggesting, the `down` key is instead behaving as expected and the `up` key is wrong?

Either way, one of the key behaviors is a bug, right?

[...]

PR #11956, opened 2025-04-14
fix(input-number): cancel arrow down event

driskull:
**Related Issue:** #10271

## Summary

- Cancel ArrowDown event using preventDefault()
- Add tests

### DIFF
--- a/packages/calcite-components/src/components/input-number/input-number.e2e.ts
+++ b/packages/calcite-components/src/components/input-number/input-number.e2e.ts
@@ -554,19 +554,35 @@ describe("calcite-input-number", () => {
       await input.callMethod("setFocus");
       await page.waitForChanges();
 
+      const eventSpy = await page.spyOnEvent("keydown");
       await page.keyboard.down("ArrowUp");
       await page.waitForTimeout(delayFor2UpdatesInMs);
       await page.waitForEvent("calciteInputNumberInput");
+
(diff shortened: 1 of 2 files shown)


eriklharper (approved):

jcfranco (approved):
🚀

packages/calcite-components/src/components/input-number/input-number.e2e.ts
```diff
@@ -554,20 +554,36 @@ describe("calcite-input-number", () => {
       await input.callMethod("setFocus");
       await page.waitForChanges();
 
+      const eventSpy = await page.spyOnEvent("keydown");
       await page.keyboard.down("ArrowUp");
       await page.waitForTimeout(delayFor2UpdatesInMs);
       await page.waitForEvent("calciteInputNumberInput");
+
+      expect(eventSpy).toHaveReceivedEventTimes(1);
+      expect(eventSpy.lastEvent.defaultPrevented).toBe(true);
```
jcfranco, line 563: Side note: we should look into updating other tests to assert for event cancelation. Might be worth exploring a util for this. Otherwise, that's gonna be a lot of `expect`s. 🥲

driskull: Agreed. A common util would be helpful here

PR #11957, opened 2025-04-15
ci: default to stable tests when experimental mode is not set

benelan:
**Related Issue:** #11877

## Summary

Make vitest default to the stable tests when the `EXPERIMENTAL_TESTS` environment variable is not set. Previously, `npx vitest` would not find any tests to run.

This change fixes the test runner integrations provided by some IDEs (including neovim and vscode).

> [!NOTE]
> We can probably get rid of the `STABLE_TESTS` environment variable with this change.


driskull (approved):
👍

Issue #11713, opened 2025-03-07
Restore `--calcite-color-focus` in the design-tokens output


jcfranco:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

`--calcite-color-focus` is no longer represented in a token after https://github.com/Esri/calcite-design-system/pull/11711.

### Expected Behavior

`--calcite-color-focus` present in both token source JSON and build output. It should output `--calcite-color-focus: var(--calcite-color-brand)` (custom prop unresolved), to associate it with the brand color while still allowing overrides.

### Reproduction Sample

https://unpkg.com/browse/@esri/calcite-design-tokens@3.0.0/dist/css/light.css

### Reproduction Steps

Notice how `--calcite-color-focus` is not defined.

**Notes**:

* Applies to https://unpkg.com/browse/@esri/calcite-design-tokens@3.0.0/dist/css/dark.css as well
* Issue won't be noticeable until #11711 lands.

### Reproduction Version

3.1.0-next.22 (or version that ships #11711)

### Relevant Info

Follow up from https://github.com/Esri/calcite-design-system/pull/11711. 

### Priority impact

impact - p1 - need for current milestone

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [x] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

Calcite (dev)
**monday.com sync:** #8961117110

isaacbraun:
Verified with `3.2.0-next.1` via https://app.unpkg.com/@esri/calcite-design-tokens@3.2.0-next.1/files/dist/css/light.css and https://app.unpkg.com/@esri/calcite-design-tokens@3.2.0-next.1/files/dist/css/dark.css.

PR #11960, opened 2025-04-15
fix(tokens): restore `--calcite-color-focus`

matgalla:
**Related Issue:** #11713 #11711 

## Summary
Restores the `--calcite-color-focus` token in both light and dark semantic color JSON files.

### Noteworthy changes

* includes the new focus token in all outputs except `index.(s)css`
* adds new `components.css` design tokens output for inclusion in `calcite-components` shadow DOM (host)
  * defines internal custom CSS property for use by focus utility styles (`--calcite-internal-color-focus`)
* exports `invert` util from `calcite-tailwind-preset` (internal) – can be moved to internal shared utils
* overrides Tailwind focus utilities in `calcite-components` – avoids depending on internal CSS props when using the preset
* restores `ts/color/css/hexrgba` transform override to support focus color token value structure
* extracts shared utilities across `calcite-design-tokens` format outputs
* enhances dictionary utility to support filtered tokens
* all components define the internal focus CSS prop, which might not be necessary in all cases – we can address this in https://github.com/Esri/calcite-design-system/issues/12106 and only include it in components that are focusable.

BEGIN_COMMIT_OVERRIDE
omitted from changelog
END_COMMIT_OVERRIDE


jcfranco:
Belated update, taking another look at this once since the current updates will reintroduce a regression since it won't allow component-level overrides to take place.

jcfranco:
> Belated update, taking another look at this once since the current updates will reintroduce a regression since it won't allow component-level overrides to take place.

One way we could approach this is to:

1. add a new output file (e.g., `component`)
2. exclude it from the components build (e.g., `index` or `core` token outputs)
3. use it in the Tailwind preset as the fallback value for focus utils ([example](https://github.com/Esri/calcite-design-system/blob/main/packages/calcite-tailwind-preset/src/index.ts#L255)) via the `es6` output): 
    ```ts
    import { calciteColorFocus } from "@esri/calcite-design-tokens/dist/es6/component";
   // ...
    ".focus-normal": {
       outline: `2px solid var(--calcite-color-focus, var(--calcite-ui-focus-color, ${calciteColorFocus}))`,
     },
    ```

We could alternatively include it in global/semantic output, but we would need to avoid adding it to `:root` (default), which would have the same [issue seen above](https://github.com/Esri/calcite-design-system/pull/11960#issuecomment-2845543967).

Thoughts? cc @matgalla @macandcheese @driskull @alisonailea

matgalla:
@jcfranco The proposed solution seems okay to me, just maybe a bit odd to have a file dedicated to a single token. Are there any other tokens that would potentially fit into that file eventually?

And we don't _need_ the focus color variable from a design/Figma standpoint, so no need to go that route.

jcfranco:
@matgalla Yes, sorry for not adding that detail earlier. The `component` output file would host semantic tokens that need to support being overridden at the component level due to fallback behavior.

The main issue is that the CSS output defines most custom CSS props on `:root`, which limits overriding/fallback patterns that need to be scoped to the component, like in this case.

>.And we don't need the focus color variable from a design/Figma standpoint, so no need to go that route.

Definitely an option. The main purpose was to restore the token representing the design decision for focus to use the brand color.

packages/calcite-components/tailwind.config.ts
```diff
@@ -17,6 +19,27 @@ const config: Config = {
...
   },
+  plugins: [
+    plugin(({ addUtilities }) => {
+      addUtilities({
+        // we override preset focus utils to allow using the internal focus color
+        ".focus-base": {
+          "outline-color": "transparent",
+        },
+        ".focus-normal": {
+          outline: "2px solid var(--calcite-internal-color-focus)",
```
driskull, line 30: should we use a var for `2px`? Maybe the border width var? var(--calcite-spacing-base)

driskull (approved):
👍

macandcheese (approved):
🏅

PR #11965, opened 2025-04-15
fix: emit beforeOpen and beforeClose immediately on toggle

jcfranco:


driskull (approved):
Looks good to me, and much simpler!

jcfranco:
Some test failures might be addressed by https://github.com/Esri/calcite-design-system/pull/12159, which updates tests to wait on events instead of relying on `waitForChanges`.

Issue #11324, opened 2025-01-16
[input-number] change event not firing when clearing the input in some locale that uses comma


rpanichakit9541:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When the input-number component is in a locale that uses comma instead of period for decimals, the `calciteInputNumberChange` event is not firing when the user clear the input.

### Expected Behavior

the `calciteInputNumberChange` event should fire correctly when clearing input-number in a locale like `de` that uses comma.

### Reproduction Sample

https://codepen.io/rpanichakit/pen/gbYKEgO?editors=1001

### Reproduction Steps

1. Click into the input and delete the content
2. Click out 
3. Observe no change event fire/log to console
4. Change the `lang` from "de" to "en"
5. Repeat step 1-3 and see the event fire correctly and log to console

### Reproduction Version

3.0.0-next.100

### Relevant Info

I think this might be related to this line https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/components/input-number/input-number.tsx#L874 that then ends up resulting in a NaN here https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/utils/number.ts#L105 

### Priority impact

impact - p3 - not time sensitive

### Impact

Bug found when BYS was testing setting extent coordinates in the Home app https://devtopia.esri.com/WebGIS/arcgis-portal-app/issues/55962

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Online

benelan:
@rpanichakit9541 change events do not fire when setting the `value` programatically. In your reproduction sample, typing numbers in the input and blurring does emit the change event. Can you please confirm?

DitwanP:
🍠 Verified locally on `dev`

PR #11982, opened 2025-04-17
fix(input-number): ensure change event correctly fires for all locales

benelan:
**Related Issue:** #11324 

## Summary

Resolve an issue where inputs in some locales would not fire the change event when clearing a value that was set programmatically via the property.

### DIFF
--- a/packages/calcite-components/src/components/input-number/input-number.e2e.ts
+++ b/packages/calcite-components/src/components/input-number/input-number.e2e.ts
@@ -1672,15 +1672,9 @@ describe("calcite-input-number", () => {
     const inputEl = await page.find("calcite-input-number");
     const clearButtonEl = await page.find("calcite-input-number >>> .clear-button");
 
-    await page.evaluate(async () => {
-      await customElements.whenDefined("calcite-input-number");
-      const el = await document.querySelector("calcite-input-number").componentOnReady();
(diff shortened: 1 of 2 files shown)


packages/calcite-components/src/components/input-number/input-number.e2e.ts
```diff
@@ -1662,6 +1662,34 @@ describe("calcite-input-number", () => {
...
+    const page = await newE2EPage();
+    await page.setContent(`
+      <calcite-input-number lang="de" value="0" clearable></calcite-input-number>
+    `);
+
+    const calciteInputNumberChange = await page.spyOnEvent("calciteInputNumberChange");
+    const inputEl = await page.find("calcite-input-number");
+    const clearButtonEl = await page.find("calcite-input-number >>> .clear-button");
+
+    await page.evaluate(async () => {
```
jcfranco, line 1675: From what I understand, `setContent` will resolve after all components have fired `componentOnReady`. Could you test to confirm this is needed? I think you could simplify to the following:

```tsx
await waitForFrame(); // might not be needed if `setContent` resolution also includes `lang` updates
await inputEl.setProperty("value", "49.173126");
```

jcfranco (approved):
✨🚀✨

PR #11990, opened 2025-04-18
chore: add design-tokens as dev dependency for Lerna updates

jcfranco:
**Related Issue:** N/A  

## Summary  

Adding explicit `calcite-design-tokens` dev dependency to the Tailwind preset package so Lerna can detect changes and update local packages correctly (see https://github.com/lerna/lerna/issues/955#issuecomment-323260292).

This resolves `next` [deployment issues](https://github.com/Esri/calcite-design-system/actions/runs/14537811536/job/40789636366) caused by missing peer dependencies.


aPreciado88 (approved):
🚀

Issue #11729, opened 2025-03-11
calcite modal prevents page scrolling after it is closed


mjuniper:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

When a modal is open when the page loads, after closing the modal you still cannot scroll the page.

### Expected Behavior

You should be able to scroll the page after closing the modal.

### Reproduction Sample

https://codepen.io/mjuniper/pen/JojrMVE?editors=1000

### Reproduction Steps

1. Visit https://codepen.io/mjuniper/pen/JojrMVE?editors=1000
2. Close the modal
3. Attempt to scroll down

### Reproduction Version

3.0.3

### Relevant Info

Could be related to https://github.com/Esri/calcite-design-system/issues/11657

### Regression?

2.13.2

### Priority impact

impact - p3 - not time sensitive

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Hub

geospatialem:
Additional investigation may be needed to determine what might be occurring, could be related to a class being added that disables scrolling.

geospatialem:
Verified with `3.2.0-next.32`

![Image](https://github.com/user-attachments/assets/24f4cc83-8bb5-4c18-a77a-8ca6754e1415)

Issue #11408, opened 2025-01-30
[dialog] overflow: hidden gets removed when closing dialog even if another dialog is still opened


rpanichakit9541:
### Check existing issues

- [x] I have [checked for existing issues](https://github.com/Esri/calcite-design-system/issues) to avoid duplicates

### Actual Behavior

If there are two dialogs opened, when you close the first one it immediately removes `overflow: hidden` from the `<html>` tag even though another dialog is still opened.

### Expected Behavior

`overflow: hidden` should remain on `<html>` tag if there's still an opened dialog

### Reproduction Sample

https://codepen.io/rpanichakit/pen/VYZgRrm?editors=1000

### Reproduction Steps

1. Open the codepen URL
2. Inspect the HTML to see the `overflow: hidden` present on the `<html>` tag inside the iframe in codepen
3. Close the dialog, and inspect again to see `overflow: hidden` is now removed, even though another dialog is still opened.

### Reproduction Version

2.13.2

### Priority impact

impact - p3 - not time sensitive

### Impact

https://devtopia.esri.com/WebGIS/arcgis-metadata-editor/issues/1536

This happens in a real workflow when the main workflow is inside a dialog and then from that UI, the user can trigger another dialog like confirmation of an action.  

### Calcite package

- [x] @esri/calcite-components
- [ ] @esri/calcite-components-react
- [ ] @esri/calcite-design-tokens
- [ ] @esri/calcite-ui-icons
- [ ] @esri/eslint-plugin-calcite-components

### Esri team

ArcGIS Online

jcfranco:
It looks like dialog needs to check the total open dialogs [here](https://github.com/Esri/calcite-design-system/blob/dev/packages/calcite-components/src/components/dialog/dialog.tsx#L767-L769) before deciding to remove the util class ([similar to modal](https://github.com/Esri/calcite-design-system/blob/8c09c74ac3b748a95ffc0776b3b294e00f20f967/packages/calcite-components/src/components/modal/modal.tsx#L475-L478)).

geospatialem:
Verified with `3.2.0-next.32`

PR #12006, opened 2025-04-21
fix(dialog, modal, sheet): fix page scrolling after closing

driskull:
**Related Issue:** #11729 #11408 #12018

## Summary

- Add a controller to handle modifying the overflow styling on the document across components
  - This is so multiple components are no longer conflicting with each other while modifying the overflow styling on the document 
- Modifies `modal` to use embedded prop similar to dialog and sheet
- Component tests already exist


packages/calcite-components/src/controllers/usePreventDocumentScroll.ts
```diff
@@ -0,0 +1,117 @@
...
+ * with `preventDocumentScroll` set to `true` is opened, and restores the scroll behavior
+ * when the component is closed or disconnected.
+ *
+ * @module usePreventDocumentScroll
+ */
+
+import { makeGenericController } from "@arcgis/components-controllers";
+import { LitElement } from "@arcgis/lumina";
+
+let count: number = 0;
```
jcfranco, line 13: Can you rename this to indicate what it's counting?

packages/calcite-components/src/controllers/usePreventDocumentScroll.ts
```diff
@@ -0,0 +1,117 @@
...
+ * @returns A controller function created by `makeGenericController`.
+ */
+export const usePreventDocumentScroll = (): ReturnType<
+  typeof makeGenericController<void, PreventDocumentScrollComponent>
+> => {
+  /**
+   * Adds a scroll prevention lock. If this is the first lock, the document's
+   * overflow style is set to "hidden".
+   */
+  const add = () => {
```
jcfranco, line 46: Can you make `add` and `remove` both functions and rename them to indicate what they're adding/removing?

packages/calcite-components/src/components/modal/modal.tsx
```diff
@@ -426,16 +437,6 @@ export class Modal extends LitElement implements OpenCloseComponent {
 
     this.titleId = ensureId(this.titleEl);
     this.contentId = ensureId(this.contentEl);
-
-    if (getNearestOverflowAncestor(this.el) === document.body) {
```
jcfranco, line 430: This adds back the dependency on the internal `embedded` property, which was removed by https://github.com/Esri/calcite-design-system/pull/9545. I think this is fine since it's internal. 

That said, we could possibly avoid the prop if we can confirm this behavior should always be enabled whenever an overlay component:

1. has no scroller container ancestor and 
2. `preventDocumentScroll` + `opened` are both true.

packages/calcite-components/src/controllers/usePreventDocumentScroll.ts
```diff
@@ -0,0 +1,117 @@
...
+  preventDocumentScroll: boolean;
+}
+
+/**
+ * Creates a controller to manage document scroll prevention for a component.
+ *
+ * This controller tracks the number of components requesting scroll prevention
+ * and ensures that the document's scroll behavior is only modified when necessary.
+ *
+ * @returns A controller function created by `makeGenericController`.
```
jcfranco, line 37: Providing `@returns` is no longer needed. The TS return type should give us this info.

jcfranco (approved):
Awesome solution, @driskull! I had a few comments, but LGTM! 

Since we are now able to test controllers, would you mind adding a spec test before merging? https://qawebgis.esri.com/components/lumina/controllers/testing#testing-controllers-in-isolation has more info on this and https://github.com/Esri/calcite-design-system/pull/11878/files#diff-068be99ab422a46919930be8603707ab41c8a4dac25ece9e72b67f1a17b15f8b shows a basic example of a browser component test. More controller tests are coming soon!

packages/calcite-components/src/components/modal/modal.tsx
driskull: Can we get a follow up issue for this one? I think it would be nice to get rid of the prop if possible.

jcfranco: For sure. Forgot to mention that I'll be creating one. 😅

driskull:
@jcfranco having trouble getting the tests to work. Do you have any real world examples I can look at? I couldn't find anything in the code linked.

jcfranco:
> @jcfranco having trouble getting the tests to work. Do you have any real world examples I can look at? I couldn't find anything in the code linked.

I'm noticing issues with browser mode tests after upgrading to Lumina v4.33. Let's revisit this once those tests are working again. Can you create a follow-up issue to unskip and fine-tune the spec test?

packages/calcite-components/src/components/modal/modal.tsx
jcfranco: ☝️ https://github.com/Esri/calcite-design-system/issues/12011

driskull:
Created https://github.com/Esri/calcite-design-system/issues/12018

packages/calcite-components/src/controllers/usePreventDocumentScroll.browser.spec.tsx
```diff
@@ -0,0 +1,96 @@
...
+  class Test extends LitElement {
+    @property() opened = false;
+
+    @property() preventDocumentScroll = false;
+
+    usePreventDocumentScroll = usePreventDocumentScroll()(this);
+  }
+
+  it("should not modify document overflow when component is not opened", async () => {
+    expect(document.documentElement.style.overflow).toBe("");
```
jcfranco, line 17: I think we could remove these initial assertions since tests are starting with a clean slate. If you need a comparison before interacting with the component, you could move these after `mount`.

packages/calcite-components/src/controllers/usePreventDocumentScroll.browser.spec.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+    expect(document.documentElement.style.overflow).toBe("");
+  });
+
+  it("should restore document overflow when component is disconnected", async () => {
+    expect(document.documentElement.style.overflow).toBe("");
+
+    const { component } = await mount(Test);
+    component.opened = true;
+    component.preventDocumentScroll = true;
```
jcfranco, line 87: Can you also add coverage for `preventDocumentScroll = false`?

jcfranco (approved):

packages/calcite-components/src/controllers/usePreventDocumentScroll.browser.spec.tsx
```diff
@@ -0,0 +1,96 @@
...
+    usePreventDocumentScroll = usePreventDocumentScroll()(this);
+  }
+
+  it("should not modify document overflow when component is not opened", async () => {
+    expect(document.documentElement.style.overflow).toBe("");
+
+    const { component } = await mount(Test);
+    component.opened = false;
+    component.preventDocumentScroll = true;
+    await waitForAnimationFrame();
```
jcfranco, line 22: I think you could use `component.updateComplete` instead of using this util.

